PR #12317, opened 2023-06-28
refactor: use macro to remove duplication of workspace inheritable fields getters

weihanglo:

Use macros to avoid repeating the getter method definitions in workspace inheritable fields.

This also removes some unnecessary clones.

I am not sure if it is preferable. Feel free to reject and close.

r? @Muscraft

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -1595,7 +1595,43 @@ pub struct InheritableFields {
     ws_root: PathBuf,
 }
 
+/// Defines simple getter methods for inheritable fields.
+macro_rules! inheritable_field_getter {
+    ( $(($key:literal, $field:ident -> $ret:ty),)* ) => (
+        $(
+            #[doc = concat!("Gets the field `workspace.", $key, "`.")]
+            pub fn $field(&self) -> CargoResult<$ret> {
+                let Some(val) = &self.$field else  {
+                    bail!("`workspace.{}` was not defined", $key);
+                };
+                Ok(val.clone())
+            }
+        )*
+    )
+}
+
 impl InheritableFields {
+    inheritable_field_getter! {
+        // Please keep this list lexicographically ordered.
+        ("dependencies",          dependencies  -> BTreeMap<String, TomlDependency>),
+        ("lints",                 lints         -> TomlLints),
+        ("package.authors",       authors       -> Vec<String>),
+        ("package.badges",        badges        -> BTreeMap<String, BTreeMap<String, String>>),
+        ("package.categories",    categories    -> Vec<String>),
+        ("package.description",   description   -> String),
+        ("package.documentation", documentation -> String),
+        ("package.edition",       edition       -> String),
+        ("package.exclude",       exclude       -> Vec<String>),
+        ("package.homepage",      homepage      -> String),
+        ("package.include",       include       -> Vec<String>),
+        ("package.keywords",      keywords      -> Vec<String>),
+        ("package.license",       license       -> String),
+        ("package.publish",       publish       -> VecStringOrBool),
+        ("package.repository",    repository    -> String),
+        ("package.rust-version",  rust_version  -> String),
+        ("package.version",       version       -> semver::Version),
+    }
+
     pub fn update_deps(&mut self, deps: Option<BTreeMap<String, TomlDependency>>) {
         self.dependencies = deps;
     }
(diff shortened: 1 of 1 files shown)


weihanglo:
```
 thread 'registry_auth::token_not_logged' panicked at 'assertion failed: `(left == right)`
  left: `8`,
 right: `7`', tests/testsuite/registry_auth.rs:588:5


failures:
    registry_auth::token_not_logged
```

Not sure why but I had a hard time reproducing it on a Linux machine.

@bors try

src/cargo/util/toml/mod.rs
```diff
@@ -1595,7 +1595,42 @@ pub struct InheritableFields {
...
+/// Defines simple getter methods for inheritable fields.
+macro_rules! inheritable_field_getter {
+    ( $(($path:literal, $field:ident -> $ret:ty),)* ) => (
+        $(
+            #[doc = concat!("Gets the field `workspace.", $path, "`.")]
+            pub fn $field(&self) -> CargoResult<$ret> {
+                if let Some(val) = &self.$field {
+                    return Ok(val.clone())
+                }
+                bail!("`workspace.{}` was not defined", $path);
```
Muscraft, line 1607: Nit: I am not sure `path` is the correct name here. It makes me think of a file path more than something within `Cargo.toml`
```suggestion
    ( $(($key:literal, $field:ident -> $ret:ty),)* ) => (
        $(
            #[doc = concat!("Gets the field `workspace.", $key, "`.")]
            pub fn $field(&self) -> CargoResult<$ret> {
                if let Some(val) = &self.$field {
                    return Ok(val.clone())
                }
                bail!("`workspace.{}` was not defined", $key);
```

src/cargo/util/toml/mod.rs
```diff
@@ -1608,161 +1643,36 @@ impl InheritableFields {
...
-            |d| Ok(d),
-        )
-    }
-
-    pub fn documentation(&self) -> CargoResult<String> {
-        self.documentation.clone().map_or(
-            Err(anyhow!("`workspace.package.documentation` was not defined")),
-            |d| Ok(d),
-        )
+        let Some(deps) = &self.dependencies else {
```
Muscraft, line 1648: Nit: It feels odd to have the getters split up by the `update_*` functions; it might be better if all the getters were beside one another. 

Note: [`ws_root`](https://github.com/rust-lang/cargo/blob/5febbe5587b74108165f748e79a4f8badbdf5e0e/src/cargo/util/toml/mod.rs#L1768) could also be moved, but it is not the exact same as the other getters. I could go either way so I will leave the decision up to you.

src/cargo/util/toml/mod.rs
```diff
@@ -1608,161 +1643,36 @@ impl InheritableFields {
...
-            |d| Ok(d),
-        )
-    }
-
-    pub fn license(&self) -> CargoResult<String> {
-        self.license.clone().map_or(
-            Err(anyhow!("`workspace.package.license` was not defined")),
-            |d| Ok(d),
-        )
+        let Some(readme) = readme_for_package(self.ws_root.as_path(), self.readme.as_ref()) else {
```
Muscraft, line 1663: Ditto

src/cargo/util/toml/mod.rs
```diff
@@ -1608,161 +1643,36 @@ impl InheritableFields {
...
-            |d| Ok(d),
-        )
-    }
-
-    pub fn include(&self) -> CargoResult<Vec<String>> {
-        self.include.clone().map_or(
-            Err(anyhow!("`workspace.package.include` was not defined")),
-            |d| Ok(d),
-        )
+        let Some(license_file) = &self.license_file else {
```
Muscraft, line 1672: Ditto

src/cargo/util/toml/mod.rs
Muscraft, line 1607: Nit: A `let-else` might be better as it would eliminate the early `return`. It would also move `bail!` into an `else`, which feels more explicit than falling through when `None` is encountered.
```suggestion
               let Some(val) = &self.$field else  {
                    bail!("`workspace.{}` was not defined", $path);
                };
                Ok(val.clone())
```

src/cargo/util/toml/mod.rs
```diff
@@ -1595,7 +1595,42 @@ pub struct InheritableFields {
...
+                }
+                bail!("`workspace.{}` was not defined", $path);
+            }
+        )*
+    )
+}
+
 impl InheritableFields {
+    inheritable_field_getter! {
+        ("dependencies",          dependencies  -> BTreeMap<String, TomlDependency>),
```
Muscraft, line 1615: Nit: Add a comment about keeping this ordered, similar to [this comment](https://github.com/rust-lang/cargo/blob/5febbe5587b74108165f748e79a4f8badbdf5e0e/src/cargo/core/features.rs#L719)
```suggestion
        // Please keep this list lexicographically ordered.
        ("dependencies",          dependencies  -> BTreeMap<String, TomlDependency>),
```

Muscraft (commented):
Overall this is a good change. One of my biggest regrets with workspace inheritance was not finding a way to reduce the amount of duplicate code. This is an excellent step in that direction!

My only concern would be with the potential drift in the doc comments for the getters as some are [handwritten](https://github.com/rust-lang/cargo/blob/324af1d678b4a9fe85643e07af26f3841d738233/src/cargo/util/toml/mod.rs#L1661) while the rest are generated by the [macro](https://github.com/rust-lang/cargo/blob/324af1d678b4a9fe85643e07af26f3841d738233/src/cargo/util/toml/mod.rs#L1602). This is a very minor concern that doesn't need to be dealt with in currently, but it is something to note.

weihanglo:
Thanks for the feedback! It's ready for the next round of review :)

Muscraft (approved):
Thanks!

@bors r+

Muscraft:
Not sure why bors didn't work, so I'll approve it again

@bors r+

PR #12319, opened 2023-06-29
Add some more documentation for Source download functions.

ehuss:
This adds a little more of a description to the download functions of the `Source` trait, since the behavior wasn't clear to me (and I often forget how these work).


weihanglo:
@bors r=epage

This didn't get into bors queue.

PR #11993, opened 2023-04-18
fix: Allow win/mac credential managers to build on all platforms

epage:
### What does this PR try to resolve?

This is a step towards #11987 by making two of the platform-specific credential managers build on all platforms using `cfg`.

I haven't done `gnome-secret` yet because that is more of an oddball in that it isn't just platforms-specific but dependent on what is installed on that platform.

### How should we test and review this PR?

```console
$ cargo check --workspace --exclude cargo-credential-gnome-secret
```

Note that the commits are broken down so you can view the movements of code separate from the functionality being changed.

### Additional information

Other information you want to mention in this PR, such as prior arts,
future extensions, an unresolved problem, or a TODO list.
-->

PR #12321, opened 2023-06-29
Don't try to compile cargo-credential-gnome-secret on non-Linux platforms.

ehuss:
This is a followup on #11993 to put the same treatment for `cargo-credential-gnome-secret` so that it builds on non-Linux platforms, substituting the "unsupported" implementation in that case.

This will still fail on Linux platforms that don't have libsecret-1-dev installed. I'm not sure how to best treat that, since we do want an error to be generated if the user tries to run `cargo install cargo-credential-gnome-secret`, and they don't have libsecret installed.


weihanglo (approved):
It's good enough. Thank you!

weihanglo:
What the merge has yet been triggered?

@bors r-

PR #12329, opened 2023-07-04
embedded: use `_` instead of `-` at the start of package names

figsoda:
Related: #12207

`-Zscript` fails when the file name starts with a digit: e.g. `0.rs` because `-0` is not a valid package name. Changing the placeholder from `-` to `_` at the start of the package name fixes the issue.


epage:
Could you please open an issue for this?  I feel like this needs more solution-exploration / solution-design discussion which I find better in Issues as things can evolve and there can be more than one PR per Issue, while I try to leave PRs to just the implementation.

figsoda:
> Could you please open an issue for this?

#12330

PR #12201, opened 2023-05-30
Extend the build directive syntax with `cargo::`

hi-rustin:
### What does this PR try to resolve?

close https://github.com/rust-lang/cargo/issues/11461

Extend the build directive syntax with `cargo::` and update tests and docs.

### How should we test and review this PR?

Check out the unit tests.

Issue #11461, opened 2022-12-06
Cannot extend build directives within cargo


epage:
Because of `cargo:KEY=VALUE`, it isn't safe to add `cargo:error=` (or any other directive) as it can be interpreted as a KEY/VALUE, blocking #10159.

XaurDesu:
Could i take a look into this?  Currently learning rust and it is marked as 'help wanted'

epage:
Sure.  Feel free to message me on Zulip if you have questions

hi-rustin:
@rustbot claim

Since there was no progress for a long time, I wanted to try to implement it.

hi-rustin:
> * Change it on an edition boundary
>   
>   * A directive could opt-in to the style
>     
>     * We'd need "begin" and "end"s to allow composing with third-party dependencies
>   * Provide an env variable with the `package.edition` for what directive syntax is accepted by default
>     
>     * There are target-specific editions but the build script is tied to the package and not the target

@epage Could you explain a bit about `change it on an edition boundary`? Did you mean we use `cargo:` in edition 2021, and use `cargo::` in the next edition?

XaurDesu:
> @rustbot claim
> 
> Since there was no progress for a long time, I wanted to try to implement it.

Go ahead, i wanted to help here but college brought me down. As @epage told me, please take a look into #11461 , because it points out to a bunch of the places that must be changed for this request.

epage:
@hi-rustin sorry if this wasn't clear but all of those notes were a summary of the discussion.  We ended up deciding this will not need an edition boundary as we ended up preferring the "Hack in a new namespace, say cargo:: instead of cargo:" route (those were meant to be alternatives, not parts of the same solution).

epage:
Its not quite clear to me what we would want another issue for or what you mean by "added to the issue list".

This issue would be resolved by the `cargo:::`.  We could have an issue for directive versioning support but I personally wouldn't bother creating one.  If we actually run into that, we can reference this issue for context.  I see it as unlikely and adding noise to our already large backlog.

XaurDesu:
Alright then, makes sense.

hi-rustin:
> @hi-rustin sorry if this wasn't clear but all of those notes were a summary of the discussion. We ended up deciding this will not need an edition boundary as we ended up preferring the "Hack in a new namespace, say cargo:: instead of cargo:" route (those were meant to be alternatives, not parts of the same solution).

Got it. Thanks! I will give it a try recently.

PR #12332, opened 2023-07-07
Bail out an error when using cargo:: in custom build script

hi-rustin:

part of https://github.com/rust-lang/cargo/issues/11461

refs https://github.com/rust-lang/cargo/pull/12201

see more: https://github.com/rust-lang/cargo/pull/12201#discussion_r1244154504

So we decided to bail out an error for `cargo::` first. After around 6 months, we will merge #12201 and stable it.

See unit test.

### Additional information
None


src/cargo/core/compiler/custom_build.rs
```diff
@@ -690,7 +690,17 @@ impl BuildOutput {
...
-                Some(val) => val,
+                Some(val) => {
+                    if val.starts_with(":") {
+                        // Line started with `cargo::`.
+                        bail!("invalid output in {}: `{}`\n\
+                            Expected a line with `cargo:key=value`, \
+                            but found `cargo::` instead.\n\
+                            See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script \
+                            for more information about build script outputs.", whence, line);
+                    }
```
epage, line 701: Should we make a comment about `cargo::` is reserved for future use or some way to indicate that there is a version mismatch between the build script and cargo, whether its because no MSRV was specified or someone was reading too new of documentation?

hi-rustin: Good suggestion. Thank you.

hi-rustin:
@epage Friendly ping~

src/cargo/core/compiler/custom_build.rs
epage: Sorry, I meant to the end-user in the error message.

Its not that this is invalid syntax but that we've reserved the syntax for future use.  Explaining that to a user will help them realize how to resolve the error (upgrade or not use it) vs being left without a way to resolve it.

hi-rustin: Oh, got it. Updated.

src/cargo/core/compiler/custom_build.rs
```diff
@@ -690,7 +690,20 @@ impl BuildOutput {
...
+                    if val.starts_with(":") {
+                        // Line started with `cargo::`.
+                        bail!("invalid output in {}: `{}`\n\
+                            Expected a line with `cargo:key=value`, \
+                            but found `cargo::` instead.\n\
+                            `cargo::` is reserved for future use or some way to indicate that there is a version mismatch \
+                            between the build script and cargo, whether its because no MSRV was specified \
+                            or someone was reading too new of documentation. \
+                            See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script \
+                            for more information about build script outputs.", whence, line);
```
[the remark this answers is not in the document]
hi-rustin: Looks good. Updated.

epage (approved):

epage:
My understanding is that we verbally agreed to this in a prior cargo meeting but I want to do a straw poll just in case since this is a minor incompatibility.

This prevents `cargo:KEY=VALUE` link metadata build-directives from having a `KEY` that starts with `:`, reserving it so we can change the build-directive prefix from `cargo:` to `cargo::` so we can change the metadata from `cargo:KEY=VALUE` to `cargo::metadata=KEY=VALUE` so we can add new build directives in the future (exact name of the directive us TBD).  The likelihood of a link metadata `KEY` starting with `:` seemed low enough that we felt it ok to have this incompatible change.  See #11461 for a summary of that past conversation.

This PR reserves the syntax (errors) so we have a transition window before accepting the new syntax so if someone uses the new syntax on older Rusts, they are more likely to get an error so they can fix it rather than the build directive appearing to be ignored (when it really was treated as link metadata).  See #12201 for the discussion that led to this transition window.

@rfcbot fcp merge

epage:
My understanding is that we verbally agreed to this in a prior cargo meeting but I want to do a straw poll just in case since this is a minor incompatibility.

This prevents `cargo:KEY=VALUE` link metadata build-directives from having a `KEY` that starts with `:`, reserving it so we can change the build-directive prefix from `cargo:` to `cargo::` so we can change the metadata from `cargo:KEY=VALUE` to `cargo::metadata=KEY=VALUE` so we can add new build directives in the future (exact name of the directive us TBD).  The likelihood of a link metadata `KEY` starting with `:` seemed low enough that we felt it ok to have this incompatible change.  See #11461 for a summary of that past conversation.

This PR reserves the syntax (errors) so we have a transition window before accepting the new syntax so if someone uses the new syntax on older Rusts, they are more likely to get an error so they can fix it rather than the build directive appearing to be ignored (when it really was treated as link metadata).  See #12201 for the discussion that led to this transition window.

@rfcbot fcp merge

hi-rustin:
@weihanglo @Eh2406 @Muscraft @arlosi Friendly ping~

epage:
Added `relnotes` label as this is a minor incompatibility that we need to make sure gets called out.

joshtriplett:
FWIW, I think we don't need a full 6 months to catch potential existing uses, but :+1: for having a transition period.

PR #12310, opened 2023-06-25
feat(crates-io): expose HTTP headers and Error type

weihanglo:
### What does this PR try to resolve?

This is part of #11521.

[RFC 3231] mentions the authentication process could have an additional **challenge-response mechanism** to prevent potential replay attacks. The challenge usually comes from HTTP `www-authenticate` header as a opaque string. When a client gets a 401/403 response with such a challenge, it may attach the `challenge` to the payload and request again to anwser the challenge.

```
➡️ cargo  requests
⬅️ server responds with `www-authenticate` containing some opaque challenge string
➡️ cargo  automatically requests again without any user perception
⬅️ server responds ok
```

However, `crates-io` crate doesn't expose HTTP headers. There is no access to `www-authenticate` header. 

This PR make it expose HTTP headers and the custom `Error` type, so `cargo` can access and do further on the authentication process.

[RFC 3231]: https://rust-lang.github.io/rfcs/3231-cargo-asymmetric-tokens.html#the-authentication-process

### How should we test and review this PR?

By commit.

This removes the dependency `anyhow` and uses our own custom `Error` enum, so that `crates-io` consumer can access `Error::API::challenge` field. It optionally uses `thiserror` to reduce boilerplates but this part can be dropped if we don't want.

PR #10592, opened 2022-04-22
Implement RFC 3139: alternative registry authentication support

arlosi:
Allows registries to request Cargo to send the authentication token for all requests, rather than just publish/yank, implementing [RFC 3139](https://github.com/rust-lang/cargo/issues/10474).

### Items from the [tracking issue](https://github.com/rust-lang/cargo/issues/10474)

> Do registries need a more fine-grained switch for which API commands require authentication?

This PR uses the `auth_required` boolean as described in the RFC.

> The RFC mentions adding --token to additional commands like install and search

These flags are not added by this PR.

> Consider changing the name and form of the X- header

Changed to the `www-authenticate` header as suggested by the comments.

> Will there be any concerns with the interaction with https://github.com/rust-lang/rfcs/pull/3231

Not that I know of.

-------------

Adds a new field `"auth-required": true` to `config.json` that indicates Cargo should include the token in all requests to a registry.

For HTTP registries, Cargo first attempts an un-authenticated request, then if that fails with HTTP 401, an authenticated request is attempted. The registry server may include a `www-authenticate` header with the HTTP 401 to instruct Cargo with URL the user can visit to acquire a token (crates.io/me).

Since the API URL is not known (because it's stored in the index), the unstable credential provider feature is modified to key off the index url, and the registry name is no longer provided.

To handle the case where an alternative registry's name is not known (such as coming from a lock file, or via `--index`), Cargo can now look up the token in the configuration by matching on the index URL. This introduces a new error if two alternative registries are configured with the same index URL.

Several operations, such as `cargo install` could have had a `--token` argument added, however it appears that Cargo would like to move away from passing the token on the command line for security reasons. In this case, users would need to configure the registry via the config file (or environment variables) when using `cargo install --index ...` or similar.

PR #12334, opened 2023-07-07
credential provider implementation

arlosi:
The current credential process protocol only allows sending the credential without any additional information. This changes the protocol in two important ways: Cargo will tell the credential provider what the token is needed for, and the credential provider can tell Cargo how the token can be used.

Since the credential provider knows why Cargo needs a token (`publish` for example), it can produce a signed token specifically for that operation. This would enable a credential process to produce an asymmetric token, or a token with restricted scope such as PASETO or Biscuit.

The credential process can also indicate back to Cargo if the token can be cached in-memory for subsequent requests. For example, if a credential provider integrates with an SSO identity provider that provides short-lived tokens, Cargo will only continue to use the token while it is valid.

### Summary of changes
* Rename `credential-process` to `credential-provider` in config.
* Add a new line-oriented JSON protocol for communicating with external credential providers via stdin/stdout.
* Allow built-in credential providers to run in the Cargo process.
* Move support for asymmetric tokens (RFC3231) into a built-in credential provider (`cargo:paseto`).
* Change the unstable key for asymmetric tokens from `registry-auth` to `credential-process`
* Add a new built-in provider to represent the current config/token based system (`cargo:token`).
* Add a new built-in provider for the a "basic" provider that prints only the token on stdout (`cargo:basic`).
* Create a new config key for the fallback credential providers (`registry.credential-providers`) as a list.
* The provider for `crates.io` no longer also acts as a fallback for other registries.
* Adds a `[credential-alias]` table for defining aliases of credential providers.
* Collect all headers from `http_registry` requests, passing them through to the cred provider.

Everything remains unstable under the `-Zcredential-process` flag.

### How to review this:
I recommend starting with the changes in `unstable.md` for a more detailed description.

### Open questions
* [x] Should we pass all the HTTP headers rather than just `www-authenticate`


ehuss (commented):
This is looking great!

This is just a partial review. I'll try to finish the rest tomorrowish.

src/cargo/util/credential/process.rs
```diff
@@ -0,0 +1,79 @@
...
+    fn perform(
+        &self,
+        registry: &RegistryInfo<'_>,
+        action: &Action<'_>,
+        args: &[&str],
+    ) -> Result<CredentialResponse, cargo_credential::Error> {
+        let mut cmd = Command::new(&self.path);
+        cmd.stdout(Stdio::piped());
+        cmd.stdin(Stdio::piped());
+        cmd.arg("--cargo-plugin");
```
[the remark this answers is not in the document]
arlosi: Added to the documentation. The idea was that a credential provider executable might have functionality that could be exposed beyond what Cargo expects.

For example, you could have `cargo-credential-foo --clear-all-credentials` or some other custom command that would only apply to running the provider executable outside of Cargo. You could also make the provider give a better help message when run without arguments.

I don't have a particular use for it at the moment though, so if we want to remove it that would be fine.

ehuss:
@arlosi Your most recent commit seems to have introduced a bunch of formatting changes.

ehuss:
Oh, I think the let-else formatting changes just landed in nightly. Lemme look at that.

ehuss:
Posted https://github.com/rust-lang/cargo/pull/12351 to update the formatting.

credential/cargo-credential-macos-keychain/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+                    let token = String::from_utf8(pass.as_ref().to_vec())
+                        .map_err(|_| Error::Other("failed to convert token to UTF8".to_string()))?;
+                    Ok(CredentialResponse::Get {
+                        token: token.into(),
+                        cache: CacheControl::Session,
+                        operation_independent: true,
+                    })
+                }
+                Action::Login(options) => {
+                    let service_name = registry(reg.name.unwrap_or(reg.index_url));
```
ehuss, line 46: There is something wrong here, though I don't think it was introduced in this PR (I'm not sure, but looks like this regressed in #10592).  This is failing for me because `login` is using a different service name than `get`. Is this something you can fix? If not, I can look at it later. It's ok to do that later if needed.

src/cargo/util/auth/mod.rs
```diff
@@ -284,10 +197,8 @@ fn registry_credential_config_inner(
         ))
     };
     Ok(
-        match (token, credential_process, secret_key, secret_key_subject) {
-            (Some(_), Some(_), _, _) => return err_both("token", "credential-process"),
+        match (token, credential_provider, secret_key, secret_key_subject) {
             (Some(_), _, Some(_), _) => return err_both("token", "secret-key"),
-            (_, Some(_), Some(_), _) => return err_both("credential-process", "secret-key"),
```
ehuss, line 290: Just checking, why were these removed?  IIUC, if you have `credential-providers` and a `token` or `secret-key` that it no longer generates an error, and ignores credential-providers? I'm concerned that could be confusing, or make it difficult to start using credential-providers. Would it be possible to retain the previous behavior?

credential/cargo-credential-1password/src/lib.rs
```diff
@@ -262,53 +255,54 @@ impl OnePasswordKeychain {
...
+                    Err(Error::NotFound)
+                }
+            }
+            _ => Err(Error::OperationNotSupported),
         }
-        Ok(())
     }
 }
-
-fn main() {
```
ehuss, line 305: In the future, how hard would it be to translate the 1password provider back to a standalone executable? We haven't decided if we are going to support these providers directly. One option was to migrate these to a separate repo (perhaps under my name) if we decide that rust-lang isn't going to support these. The maintenance burden isn't too bad, so I'm not sure if it matters (and we can document that we can drop providers in the future if we can't support it).

src/cargo/util/auth/mod.rs
```diff
@@ -440,285 +448,95 @@ pub fn auth_token(
...
-    if let Some(cache_token_value) = cache.get(url) {
-        // Tokens for endpoints that do not involve a mutation can always be reused.
-        // If the value is put in the cache by the command line, then we reuse it without looking at the configuration.
-        if cache_token_value.from_commandline
-            || cache_token_value.independent_of_endpoint
-            || mutation.is_none()
+    if let Some(cached_token) = cache.get(url) {
+        if cached_token
+            .expiration
+            .map(|exp| OffsetDateTime::now_utc() < exp)
```
ehuss, line 460: Does this need a grace window to handle the case where it expires in like 1 second? 
(Probably not too necessary, just curious if that needs to be considered.)

ehuss (commented):
Review part 2. I think I've gone through everything except the tests, will try to get to those tomorrow.

credential/cargo-credential-1password/src/lib.rs
arlosi: It's very easy to translate back. Just need to change it back to a `bin` crate and add:
```
fn main() {
    cargo_credential::main(OnePasswordCredential);
}
```

I think I'd also prefer that 1password was a standalone provider that you could `cargo install`.

credential/cargo-credential-macos-keychain/src/lib.rs
arlosi: Yes, it looks like I broke it in #10592 😢. I'll fix it and try to borrow a Mac to validate it.

src/cargo/util/auth/mod.rs
arlosi: I intended it to prefer `credential-provider` if both were set. I'll fix that. More tests are needed.

The intent of the change here was to:
* Allow a `credential-provider` such as `cargo:token`, which would then read corresponding `token` field
* Use the `credential-provider` (for when running with nightly), but then still have the `token` fallback so stable continues to work.

src/cargo/util/auth/mod.rs
```diff
@@ -1,142 +1,106 @@
...
+        }) if allow_cred_proc => vec![resolve_credential_alias(config, provider)],
 
-impl<T: AsRef<str>> Secret<T> {
-    /// Checks if the contained value is empty.
-    pub fn is_empty(&self) -> bool {
-        self.inner.as_ref().is_empty()
-    }
-}
+        // If a token is configured, use the token provider.
+        Some(RegistryConfig { token: Some(_), .. }) => vec![vec!["cargo:token".to_string()]],
```
ehuss, line 72: One of my concerns about how this fallback works is the following scenario:

1. Someone has `registry.token` or `registries.NAME.token` saved in the credentials.toml.
2. They want to switch to using a credential provider, so they set `registry.credential-providers = ["…"]`
3. They run some commands, but the provider never works, it only uses the old token.

What do you think about doing the following? In this case, check if the fallback providers contains `cargo:token`. If it does, proceed using the fallback list of providers. If it does not...I'm not sure. I didn't like the old error message. Perhaps it could just ignore the token? Or generate a warning that the token is ignored, and indicate the file where you can delete it to remove the warning?

arlosi: Thanks for bringing this up. I agree with your suggestion of warning if the token is going to be ignored. I've added similar style warnings for other cases (such as `secret-key` and `token`) as well.

Here's the message I've added for this case.
```
warning: {sid} has a token configured in {} that will be ignored because the `cargo:token` credential 
provider is not listed in  `registry.global-credential-providers`
```

arlosi:
> Something weird happens if you have a configuration with `registries.NAME.credential-providers = ["..."]` (notice the `s` in the key). It seems to ignore that value, and doesn't generate a warning about an unused key. I would expect a warning about an unused key. It also seems to do some basic validation, since if you have `credential-providers = "..."`, it will generate an error about not being a list. In general I'm a little concerned that users will get confused about the subtle distinction of whether or not there is an `s` in the key.

There was a shared struct that was used for both `registries.NAME` and `registry`, with `registry` allowing more keys. I've separated it out so we get better warnings for unused keys.

I've also renamed the `credential-providers` to `global-credential-providers` (name subject to bikeshedding), to make it more distinct from `credential-provider`.

tests/testsuite/credential_process.rs
```diff
@@ -416,89 +237,233 @@ fn owner() {
...
+                description = "foo"
+                license = "MIT"
+                homepage = "https://example.com/"
+            "#,
+        )
         .file("src/lib.rs", "")
         .build();
 
-    p.cargo("publish --no-verify --registry alternative -Z credential-process")
+    let output = r#"[UPDATING] `alternative` index
```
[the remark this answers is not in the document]
arlosi: > If it wasn't expired, would the second request (for "publish") be missing?

Yes. I'll add a comment as well.

arlosi:
@ehuss I think this is ready for another round of review.

Given that #12310 was just merged, I've also made changes here to pass all the headers into the credential provider.

ehuss:
Looks great, thanks!  If there are thoughts about what needs followup, please file new issues, but I can't think of anything right now.

@bors r+

src/cargo/ops/registry/login.rs
```diff
@@ -47,114 +33,21 @@ pub fn registry_login(
...
-
-        if let Some(tok) = new_token.as_token() {
-            crates_io::check_token(tok.as_ref().expose())?;
+    let mut token_from_stdin = None;
+    if !std::io::stdin().is_terminal() {
+        let token = std::io::read_to_string(std::io::stdin()).unwrap_or_default();
+        if !token.is_empty() {
+            token_from_stdin = Some(token);
         }
     }
```
DrChat, line 43: Just a friendly heads up - this is causing our CI jobs to hang if we provide the token via the command-line instead of `stdin`.

weihanglo: Sorry for the inconvenience!

I didn't quite follow the PR at that time. Is that a regression? If yes, please file an issue with reproduction, and we can then arrange a fix!

DrChat: Yes, this is a regression.
Sounds good, thanks :)

DrChat: #12438

PR #12335, opened 2023-07-08
update re: multiple versions that differ only in the metadata tag

ahl:
Resolved by https://github.com/rust-lang/crates.io/issues/1059 a few weeks ago


epage:
Thanks!

@bors r+

PR #12339, opened 2023-07-09
docs: Use heading attributes to control the fragment.

ehuss:
mdBook recently gained the ability to use [heading attributes](https://rust-lang.github.io/mdBook/format/markdown.html#heading-attributes) which allows you to define the HTML id (and thus URL fragment) for a section. This is an improvement over using manual `id` tags since mdBook can now know which ID matches the section, and will automatically set up links and such.

Some of these pages were using `id` tags to handle backwards-compat for old URLs. For those, I switched them to use redirects instead.

One downside is that the links won't work in GitHub render previews. Overall I think the benefits are worth it, though.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -232,7 +232,7 @@ jobs:
     - name: Install mdbook
       run: |
         mkdir mdbook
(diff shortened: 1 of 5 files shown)


weihanglo (commented):
I guess we need to upgrade mdbook to at least `0.4.30`

https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-0430

ehuss:
Sorry, I was going to comment here. The rust-lang/rust repo is already updated, so there shouldn't be anything that is needed.

PR #12340, opened 2023-07-10
Minor: Use "number" instead of "digit" when explaining Cargo's use of semver

nicoburns:
Digit is technically incorrect here, as it would imply that Cargo treats 0.10.0 and 0.11.0 as semver compatible (because they have the same leftmost non-zero *digit*, even though they do not have the same leftmost non-zero *number*).


weihanglo (approved):
Looks reasonable. Thanks!

PR #12341, opened 2023-07-10
chore: Automatically update dependencies monthly

epage:

This will create PRs that will update `Cargo.toml` and `Cargo.lock` at the same time.  By default, RenovateBot will update all dependencies, so we pull out the compatible dependencies into a consolidated PR to reduce disruptions and CI load.

This doesn't just cover Rust dependencies but github workflows as well.

This starts off with the dependency dashboard and automerging being disabled but explicitly specified to call attention to the potential for this.  We can evaluate these later.

The frequency of monthly was chosen to consolidate disruptions.  The main risk is doing an update right before a beta branch but we can't have a cron schedule for once every 6 weeks (I assume) which would be ideal.  At least its not daily or instant.

I chose `json5` over `json` for comments and trailing commas, to make human-editing easier.

I pulled the config from https://github.com/gitext-rs/git-stack/blob/main/.github/renovate.json5
- [Dependency dashboard](https://github.com/gitext-rs/git-stack/issues/309)
- [Compatible PR](https://github.com/gitext-rs/git-stack/pull/255)
- [Incompatible PR](https://github.com/gitext-rs/git-stack/pull/310)

CC @Turbo87

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,35 @@
+{
+  schedule: [
+    'before 3am on the first day of the month',
+  ],
+  semanticCommits: 'enabled',
+  configMigration: true,
+  dependencyDashboard: false,
+  packageRules: [
+    // Goals:
+    // - Rollup safe upgrades to reduce CI runner load
+    // - Have lockfile and manifest in-sync (implicit rules)
+    {
+      matchManagers: [
+        'cargo',
+      ],
+      matchCurrentVersion: '>=0.1.0',
+      matchUpdateTypes: [
+        'patch',
+      ],
+      automerge: false,
+      groupName: 'compatible',
+    },
+    {
+      matchManagers: [
+        'cargo',
+      ],
+      matchCurrentVersion: '>=1.0.0',
+      matchUpdateTypes: [
+        'minor',
+      ],
+      automerge: false,
+      groupName: 'compatible',
+    },
+  ],
+}


Turbo87 (approved):
not quite what I would use, but this looks good for achieving your goals! :)

weihanglo:
Thanks for this! I was thinking about the same thing early today. rust-lang/rust did a similar thing from another angle:
https://github.com/rust-lang/rust/blob/05b82e551e781154ad2af8a72f34712e97206859/.github/workflows/dependencies.yml

epage:
> Thanks for this! I was thinking about the same thing early today. rust-lang/rust did a similar thing from another angle:

RenovateBot will provide us a bit better of an experience with giving us release notes.

Also a potential future change is for the semver-check job to be run using the latest version of rust with a CI job to update that version (thanks to RenovateBot).  This would isolate error change failures to a PR rather than affecting all PRs.

weihanglo (approved):

weihanglo:
Going to merge it now. If someone objects to this or it ends up too noisy, we can always disable it afterward.

@bors r+

ehuss:
Have we confirmed whether or not the renovatebot app is installed for all rust-lang repos, or just select ones?

epage:
Posted: https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/renovatebot.20for.20cargo

Issue #11987, opened 2023-04-17
Leverage the `--workspace` flag when running cargo command against cargo itself


weihanglo:
### Problem

After #11851 Cargo becomes a Cargo workspace. That means we can dogfood ourselves finally 🎉.

We left some TODOs in that pull request, one of them are making `--workspace` usable for cargo itself. That is, making `cargo build` and other subcommands happy with `--workspace` flag, instead of using `-p` flag everywhere in our CI workflow yaml.

Some workspace members are platform-dependant inside cargo workspace, i.e., `cargo-credential-gnome-secret` and friends. We don't really have a good programming way to get the list other than `cargo metadata | jq .workspace_members[] -r | cut -d' ' -f1`. As a consequence, we expands the list manually in a ugly way: https://github.com/rust-lang/cargo/blob/5861176b5607c67e44ea7d438954fbb5fa20b190/.github/workflows/main.yml#L117-L128

We want to make running command like `cargo test --workspace` as easy as possible to reduce the friction in CI also in contributor experience.

- [ ] ~~Fix clippy warnings and re-enable clippy check in CI for all workspace members.~~
  - Blocked on rust-lang/rfcs#3389 so we can more easily propagate our clippy settings
- [ ] Fix rustdoc warnings and re-enable rustdoc check in CI for all workspace members.
- [ ] Fix `linkchecker.sh` warnings in CI (https://github.com/rust-lang/cargo/pull/11851#discussion_r1147051184)
- [ ] Leverage workspace flag `--workspace` when running `cargo build` or `cargo test`, instead of using flag `-p`.
- [ ] Leverage glob syntax when probing members in `[workspace]` in Cargo.toml (i.e., `crates/*`).

### Proposed Solution

Could look into <https://github.com/rust-lang/cargo/issues/5220> and also <https://github.com/rust-lang/cargo/issues/11526>.

An alternative workaround would be making those subcrates and dependencies platform specific. For example, dependencies could be listed in `target."cfg(not(windows))".dependencies` section. And have a `lib.rs` doing nothing but delegating to `unix/lib.rs` or `windows/lib.rs`, and one of them also does nothing.

### Notes

I think at least the follow commands should be able to run with `--workspace` for cargo itself.

- [ ] `cargo build`/`cargo check`
- [ ] `cargo test`
- [ ] `cargo doc`
- [ ] `cargo clippy` (this is blocked on rust-lang/rfcs#3389, so we can more easily propagate our clippy settings)

[...]

PR #12342, opened 2023-07-10
chore(ci): Automatically test new packages by using `--workspace`

epage:

This is a part of #11987 that was unblocked by #12321 and #11993

Relying on CI to verify this

I'd prefer to move the `cargo test --workspace` step earlier but we run out of disk space if we do (we currently only have 661 MB of headroom of windows-gnu)

See the individual commits for smaller chunks and explanations for why changes were made.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -154,6 +154,7 @@ jobs:
     - run: cargo test -p cargo
     - name: Clear intermediate test output
       run: ci/clean-test-output.sh
+
     - name: gitoxide tests (all git-related tests)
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
One drawback for testing all members in a `cargo test` invocation is that the feature resolution is unified. If we aim to make each subcrate work more independently in the future, we may not want to unify features.

That being said, I have no issue merging them at this stage.

weihanglo:
Oh no. I didn't notice it was still a draft.

epage:
It was just a draft until it could successfully build

PR #12343, opened 2023-07-10
docs: Put cargo-add change under nightly

epage:
I'm assuming we'll want to beta-backport this.


epage:
This was carried over into the rust 1.71 release notes where I noticed it.  See also rust-lang/rust#113541.

ehuss:
Thanks. I don't think this necessarily needs a backport since the release notes and blog posts all link directly to the master branch here. I don't think anyone sees the CHANGELOG from the release branches. Also, I believe the 1.71 stable release has already left the station.

@bors r+

PR #12344, opened 2023-07-10
contrib docs: Add some more detail about how publishing works

ehuss:
This adds a little clarification on how cargo releases are made, and how things get published.

### DIFF
--- a/src/doc/contrib/src/process/release.md
+++ b/src/doc/contrib/src/process/release.md
@@ -132,15 +132,18 @@ to [crates.io] as part of the 6-week stable release process by the [Release
 team]. There is a [`publish.py` script] that is used by the Release team's
 automation scripts (see <https://github.com/rust-lang/simpleinfra/>) to handle
 determining which packages to publish. The test and build tool crates aren't
-published.
+published. This runs on the specific git commit associated with the cargo
+submodule in the `stable` branch in `rust-lang/rust` at the time of release.
 
 On very rare cases, the Cargo team may decide to manually publish a new
 release to [crates.io]. For example, this may be necessary if there is a
 problem with the current version that only affects API users, and does not
 affect the `cargo` binary shipped in the stable release. In this situation,
-the patch version should be bumped, and then someone with permissions
-(currently a subset of the Cargo team, or the Release team) should publish it
-manually.
+PRs should be merged to the associated stable release branch in the cargo repo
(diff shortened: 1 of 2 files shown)


src/doc/contrib/src/process/release.md
```diff
@@ -127,11 +127,35 @@ module].
...
 ## crates.io publishing
 
-Cargo's library is published to [crates.io] as part of the stable release
-process. This is handled by the [Release team] as part of their process. There
-is a [`publish.py` script] that in theory should help with this process. The
-test and build tool crates aren't published.
-
+Cargo's library and its related dependencies (like `cargo-util`) are published
+to [crates.io] as part of the 6-week stable release process by the [Release
+team]. There is a [`publish.py` script] that is used by the Release team's
```
epage, line 132: from the release branch, correct?

src/doc/contrib/src/process/release.md
```diff
@@ -127,11 +127,35 @@ module].
...
+determining which packages to publish. The test and build tool crates aren't
+published.
+
+On very rare cases, the Cargo team may decide to manually publish a new
+release to [crates.io]. For example, this may be necessary if there is a
+problem with the current version that only affects API users, and does not
+affect the `cargo` binary shipped in the stable release. In this situation,
+the patch version should be bumped, and then someone with permissions
+(currently a subset of the Cargo team, or the Release team) should publish it
+manually.
```
epage, line 143: Is this happening from `master`?

epage: If this is happening from master, couldn't we run into temporal issues where patches for old versions have features going into the new versions?

Seems like we'd need to cherry-pick relevant changes back into the release branches and release from those.

src/doc/contrib/src/process/release.md
weihanglo: I think yes. This is the same release process as `cargo`'s.

weihanglo (approved):
I am good with the current one, but also agree with epage on some places we can make it clearer

src/doc/contrib/src/process/release.md
ehuss: To be more specific, it is the `stable` branch in `rust-lang/rust`. I have added some text to try to clarify this.

epage:
Thanks for the clarifications!

PR #12349, opened 2023-07-11
fix(embedded): Always generate valid package names

### LINKED ISSUE #12330: cargo -Zscript uses invalid package names when the file name starts with a digit
### Problem

When the file name starts with a digit, e.g. `0.rs`, `cargo -Zscript 0.rs` defaults the package name to `-0` and fails

### Steps

```bash
echo 'fn main() {}' > 0.rs
cargo -Zscript 0.rs
```

```console
warning: `package.edition` is unspecifiead, defaulting to `2021`
error: failed to parse manifest at `<...>/0.rs`

Caused by:
  invalid character `-` in package name: `-0`, the first character must be a Unicode XID start character (most letters or `_`)
```

### Possible Solution(s)

When the first character of the file name is invalid, insert `_` instead of `-`, which makes the package name valid. So for `0.rs`, the package name would default to `_0`. This is #12329's approach

### Notes

Related: #12207, https://github.com/rust-lang/cargo/pull/12329#issuecomment-1620857167

### Version

```text
cargo 1.72.0-nightly (5b377cece 2023-06-30)
release: 1.72.0-nightly
commit-hash: 5b377cece0e0dd0af686cf53ce4637d5d85c2a10
commit-date: 2023-06-30
os: NixOS 23.11.0 [64-bit]
```

epage:

The sanitization logic uses a placeholder for the first character that isn't valid in the first character position.  #12329 took the approach of always using `_` which has the problem of mixing separators if the user used `-` or we had other placeholders to insert.  Instead, this takes the approach of stripping the leading invalid characters and using a placeholder name if nothing is left.

Fixes #12330

Per-commit.  The first adds tests so the change in behavior can be observed over each additional commit.

### Additional information

I was also hoping to make the binary name not use placeholders by setting `bin.name` to `file_stem` but then I got
```
   Compiling s-h-w-c- v0.0.0 (/home/epage/src/personal/cargo/target/tmp/cit/t133/foo)
error: invalid character `'.'` in crate name: `s_h.w§c!`

error: invalid character `'§'` in crate name: `s_h.w§c!`

error: invalid character `'!'` in crate name: `s_h.w§c!`

error: could not compile `s-h-w-c-` (bin "s-h.w§c!") due to 3 previous errors
```
I decided to not get into what are or aren't valid characters according to rustc.

### DIFF
--- a/src/cargo/util/restricted_names.rs
+++ b/src/cargo/util/restricted_names.rs
@@ -100,6 +100,9 @@ pub fn sanitize_package_name(name: &str, placeholder: char) -> String {
             slug.push(placeholder);
         }
     }
+    if slug.is_empty() {
+        slug.push_str("package");
+    }
     slug
 }
 
--- a/src/cargo/util/restricted_names.rs
+++ b/src/cargo/util/restricted_names.rs
@@ -87,17 +87,13 @@ pub fn validate_package_name(name: &str, what: &str, help: &str) -> CargoResult<
 pub fn sanitize_package_name(name: &str, placeholder: char) -> String {
     let mut slug = String::new();
     let mut chars = name.chars();
-    if let Some(ch) = chars.next() {
-        if ch.is_digit(10) {
-            slug.push(placeholder);
+    while let Some(ch) = chars.next() {
+        if (unicode_xid::UnicodeXID::is_xid_start(ch) || ch == '_') && !ch.is_digit(10) {
             slug.push(ch);
-        } else if unicode_xid::UnicodeXID::is_xid_start(ch) || ch == '_' {
-            slug.push(ch);
-        } else {
-            slug.push(placeholder);
+            break;
         }
     }
-    for ch in chars {
+    while let Some(ch) = chars.next() {
         if unicode_xid::UnicodeXID::is_xid_continue(ch) || ch == '-' {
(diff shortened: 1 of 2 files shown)


src/cargo/util/restricted_names.rs
```diff
@@ -87,23 +87,22 @@ pub fn validate_package_name(name: &str, what: &str, help: &str) -> CargoResult<
...
-    for ch in chars {
+    while let Some(ch) = chars.next() {
         if unicode_xid::UnicodeXID::is_xid_continue(ch) || ch == '-' {
             slug.push(ch);
         } else {
             slug.push(placeholder);
         }
     }
+    if slug.is_empty() {
+        slug.push_str("package");
```
weihanglo, line 104: I might prefer `script` but never mind it's not a big deal.

weihanglo (approved):
Looks reasonable. Thanks!

src/cargo/util/restricted_names.rs
epage: I'm avoiding "script" in user-facing contexts because I feel it can be misleading.

epage: I'm avoiding "script" in user-facing contexts because I feel it can be misleading.

PR #12350, opened 2023-07-11
fix(embedded): Error on intentionally unsupported commands 

epage:

We had talked about de-prioritizing some commands in the cargo team meeting today.  Looking through `cargo --list` and trying them out, the important ones to defer seemed to be:

- `cargo pkgid` is unsupported because we can't (yet) generate valid
  pkgids for embedded manifests.  Adding support for this would be a
  step towards workspace support
- `cargo package` and `cargo publish` are being deferred.  These would
  be more important for `[lib]` support.  `cargo install` for `[[bin]]`s
  is a small case and As single-file packages are fairly restrictive, a
  `[[bin]]` is a lower priority.

Per-commit

### DIFF
--- a/src/bin/cargo/commands/package.rs
+++ b/src/bin/cargo/commands/package.rs
@@ -40,6 +40,13 @@ pub fn cli() -> Command {
 
 pub fn exec(config: &mut Config, args: &ArgMatches) -> CliResult {
     let ws = args.workspace(config)?;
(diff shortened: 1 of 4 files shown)


weihanglo (approved):
Looks good! I am still surprised you always can file PRs immediately right after we made some decisions during the meeting 😆.

PR #12352, opened 2023-07-12
refactor: Clean up package metadata

epage:

Clean up workspace metadata
- Reduce noise by allowing fields to be inferred
  - I left documentation because that is runtime inferred on crates.io which has trade offs
- Default package fields at the workspace level

Align us on a single edition

Mostly relying on CI for this

### Additional information

I noticed this and decided to do this while I was considering the idea of setting an MSRV to "latest" by leveraging #12341 and whether that should just be for `cargo` or for all workspace members.  I'm leaning towards all workspace members as that is the only thing we test though strictly speaking people might be able to use lower versions; we just wouldn't officially support it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -423,7 +423,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-util"
-version = "0.2.4"
+version = "0.2.5"
 dependencies = [
  "anyhow",
  "core-foundation",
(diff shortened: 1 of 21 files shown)


weihanglo (approved):

weihanglo:
Thanks for this!

@bors r+

weihanglo:
@bors r-

version bump check is failing

epage:
@weihanglo this looks ready if we merge it now and don't sneeze to break it again...

weihanglo (approved):

weihanglo (approved):

weihanglo:
Nice! Thank you.

@bors r+

PR #12244, opened 2023-06-08
fix: fetch nested git submodules

krs98:
### What does this PR try to resolve?

Fixes #12151.

When recursing submodules, the url of the parent remote was being passed to `update_submodules` instead of the child remote url. This caused Cargo to try to add the wrong submodule. 

### How should we test and review this PR?

A test case is added in the first commit. The second one renames the `url` variable as suggested in the issue. The third includes the changes to fix the issue. The last one includes a minor refactor where a redundant `match` expr is removed.

PR #12359, opened 2023-07-14
fix(git): respect scp-like URL for nested submodules

### LINKED ISSUE #12295: Git SSH submodules that omit "ssh://" no longer work
### Problem

When using a Git submodule that uses the SSH protocol, Git allows the `ssh://` to be omitted and the URL to be provided in a form such as `git@github.com:...`. Previously this worked fine with Cargo, but since  some changes a couple of weeks ago related to handling submodules this seems to no longer work. I instead get an error such as the following:

```
Caused by:
  failed to load source for dependency `<redacted>`

Caused by:
  Unable to update ssh://git@github.com/<redacted>

Caused by:
  failed to update submodule `<redacted>`

Caused by:
  failed to parse url for submodule `<redacted>`: `git@github.com:<redacted>.git`

Caused by:
  relative URL without a base
```

As you can see it's failing to parse the `git@github.com:...` URL, it seems to think it's a relative URL, but it is not a relative URL, it is an absolute URL which doesn't specify the `ssh://` protocol, which is a valid Git submodule URL, so should be supported.

I am already using `git-fetch-with-cli = true` in my `.cargo/config.toml` file to fix other issues with using SSH authentication to private repositories, but it doesn't seem to fix this issue.

### Steps

1. Create a crate which contains a submodule which uses an SSH URL that omits the `ssh://` prefix.
2. Depend on that crate from another crate.
3. Try to build this latter crate.

### Possible Solution(s)

Any parsing of a Git submodule URL should allow the protocol to be omitted. It may be the call to `Url::parse()` added in the commit https://github.com/rust-lang/cargo/commit/2717a34b60af17282ecfa66e8f47fa83a63ebf0d which has triggered this, so possibly revert that particular part of the change to use the string verbatim without parsing, as was done before.

### Notes

_No response_

### Version

Fails using `nightly-2023-06-11` or onwards, `nightly-2023-06-10` does not exhibit this issue. The version output on the earliest failing version is:

```text
cargo 1.72.0-nightly (49b6d9e17 2023-06-09)
release: 1.72.0-nightly
commit-hash: 49b6d9e179a91cf7645142541c9563443f64bf2b
commit-date: 2023-06-09
libgit2: 1.6.4 (sys:0.17.1 vendored)
libcurl: 8.1.2-DEV (sys:0.4.63+curl-8.1.2 vendored ssl:OpenSSL/1.1.1t)
```

weihanglo:

`parent_remote_url` used to be `&str` before #12244. However, we changed
the type to `Url` and it started failing to parse scp-like URLs since
they are not compliant with WHATWG URL spec.

In this commit, we change it back to `&str` and construct the URL
manually. This should be safe since Cargo already checks if it is a
relative URL for that if branch.

This is current a draft because I had a hard time adding a test around it. 

* Scp-like URL (e.g. `git@github.com:rust-lang/cargo.git`) doesn't accept custom port. In our `container_test` we need to specify a custom port number.
* There seems to be no way to configure a git submodule without cloning the repo. Otherwise we could have configured it and perceived it fail when fetching instead of parsing URL.

To test it, I ended up manually creating a repo with a submodule with scp-like URL.

### Additional information

I feel like this worth a beta backport (we missed the nightly window).

I also feel like it is pretty safe to merge without adding a test if we cannot figure out how to.

Fixes #12295

### DIFF
--- a/tests/testsuite/git.rs
+++ b/tests/testsuite/git.rs
@@ -3633,7 +3633,7 @@ fn different_user_relative_submodules() {
             .file("Cargo.toml", &basic_lib_manifest("dep1"))
             .file("src/lib.rs", "")
     });
-    let user2_git_project2 = git::new("user2/dep2", |project| {
+    let _user2_git_project2 = git::new("user2/dep2", |project| {
         project
             .file("Cargo.toml", &basic_lib_manifest("dep1"))
             .file("src/lib.rs", "")
@@ -3673,14 +3673,14 @@ fn different_user_relative_submodules() {
             "\
 [UPDATING] git repository `{}`
 [UPDATING] git submodule `{}`
-[UPDATING] git submodule `{}`
+[UPDATING] git submodule `{}/../dep2`
 [COMPILING] dep1 v0.5.0 ({}#[..])
 [COMPILING] foo v0.5.0 ([CWD])
 [FINISHED] dev [unoptimized + debuginfo] target(s) in [..]
 ",
             path2url(&user1_git_project.root()),
             path2url(&user2_git_project.root()),
-            path2url(&user2_git_project2.root()),
+            path2url(&user2_git_project.root()),
             path2url(&user1_git_project.root()),
         ))
         .run();
--- a/src/cargo/sources/git/utils.rs
+++ b/src/cargo/sources/git/utils.rs
@@ -392,13 +392,13 @@ impl<'a> GitCheckout<'a> {
     ///
     /// [^1]: <https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt-none>
     fn update_submodules(&self, cargo_config: &Config) -> CargoResult<()> {
-        return update_submodules(&self.repo, cargo_config, self.remote_url());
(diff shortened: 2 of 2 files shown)


tests/testsuite/git.rs
```diff
@@ -3673,14 +3673,14 @@ fn different_user_relative_submodules() {
             "\
 [UPDATING] git repository `{}`
 [UPDATING] git submodule `{}`
-[UPDATING] git submodule `{}`
+[UPDATING] git submodule `{}/../dep2`
```
weihanglo, line 3676: The regression of URL display is one downside of it. However, since we drop `Url::parse` here so now we can have relative submodule inside a scp-like submodule. For example, this was not possible before this PR:

```
direct git dependency
  submodule url `git@github.com/user/repo`
     submodule url `../`
```

ehuss:
Looks good to me.

Yea, there isn't a particularly good way to test scp-style URLs since they don't support port numbers. My only idea would be to put cargo itself inside a docker container, and orchestrate the test around using port 22. But that sounds sufficiently annoying that I wouldn't want to bother with it yet. 

A beta backport sounds good to me, too.

@bors r+

PR #12362, opened 2023-07-15
Perform atomic write to avoid corrupting files

### LINKED ISSUE #11386: `cargo add` and `ctrl + c`
### Problem

I hit `ctrl+c` when cancelling a redundant `cargo add` command, and discovered that my Cargo.toml had been entirely deleted.

### Steps

1. `cargo add` a crate
2. `ctrl+c` before it finishes successfully

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
cargo version --verbose
cargo 1.65.0 (4bc8f24d3 2022-10-20)
release: 1.65.0
commit-hash: 4bc8f24d3e899462e43621aab981f6383a370365
commit-date: 2022-10-20
os: CentOS 7.9.2009 (Core) [64-bit]
```

marcospb19:
Fix #11386 with @epage's suggestion, atomic writes.

Instead of applying the fix just to `cargo-add`, I applied it to everything, we might want to review each specific case to tell if cases.

### DIFF
--- a/crates/cargo-util/src/paths.rs
+++ b/crates/cargo-util/src/paths.rs
@@ -153,18 +153,39 @@ pub fn read_bytes(path: &Path) -> Result<Vec<u8>> {
     fs::read(path).with_context(|| format!("failed to read `{}`", path.display()))
 }
 
-/// Writes a file to disk.
+/// Atomically writes a file to disk.
 ///
-/// Equivalent to [`std::fs::write`] with better error messages.
-pub fn write<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()> {
+/// This function writes to a temporary file and renames it to overwrite the previous one,
+/// this way, we prevent that interruptions (like SIGINT or SIGKILL) corrupt the file.
+///
+/// # Errors:
+///
+/// This function will return an error if:
+///
+/// 1. `fs::write` fails writing to the temporary file.
+/// 2. `fs::rename` fails when copying from temporary to destination file.
+pub fn atomic_write<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()> {
     let path = path.as_ref();
-    fs::write(path, contents.as_ref())
-        .with_context(|| format!("failed to write `{}`", path.display()))
+
+    let mut temporary_file = tempfile::Builder::new().suffix(path).tempfile()?;
+    temporary_file
+        .write_all(contents.as_ref())
+        .with_context(|| format!("failed to write to `{}`", temporary_file.path().display()))?;
+
+    fs::rename(temporary_file.path(), path)
+        .with_context(|| "failed to perform atomic write")
+        .with_context(|| {
+            format!(
+                "failed to rename `{}` to `{}`",
+                temporary_file.path().display(),
+                path.display()
+            )
+        })
 }
 
-/// Equivalent to [`write()`], but does not write anything if the file contents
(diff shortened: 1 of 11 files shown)


marcospb19:
So @epage I'm getting this error in local testing when trying to copy from /tmp to current folder:

```rust
`Result::unwrap()` on an `Err` value: Os { code: 18, kind: CrossesDevices, message: "Invalid cross-device link" }'
```

So I need a temporary file in the same filesystem? Where is a good place for that?

crates/cargo-util/src/paths.rs
```diff
@@ -153,18 +153,39 @@ pub fn read_bytes(path: &Path) -> Result<Vec<u8>> {
...
     let path = path.as_ref();
-    fs::write(path, contents.as_ref())
-        .with_context(|| format!("failed to write `{}`", path.display()))
+
+    let mut temporary_file = tempfile::Builder::new().suffix(path).tempfile()?;
+    temporary_file
+        .write_all(contents.as_ref())
+        .with_context(|| format!("failed to write to `{}`", temporary_file.path().display()))?;
+
+    fs::rename(temporary_file.path(), path)
```
epage, line 175: > So @epage I'm getting this error in local testing when trying to copy from `/tmp` (generated by `tempfile` crate) to current folder:
> 
> ```rust
> `Result::unwrap()` on an `Err` value: Os { code: 18, kind: CrossesDevices, message: "Invalid cross-device link" }'
> ```
> 
> So I need a temporary file in the same filesystem? Where is a good place for that?
> 
> EDIT: I guess we could pick `target/temporary_file` and we should be OK? File disappears after renaming anyways. EDIT2: nevermind, `target` doesn't exist after `cargo new` and `cargo add`.

Atomic renames requires doing so on the same filesystem.

crates/cargo-util/src/paths.rs
```diff
@@ -153,18 +153,39 @@ pub fn read_bytes(path: &Path) -> Result<Vec<u8>> {
...
+                path.display()
+            )
+        })
 }
 
-/// Equivalent to [`write()`], but does not write anything if the file contents
+/// Equivalent to [`atomic_write()`], but does not write anything if the file contents
 /// are identical to the given contents.
-pub fn write_if_changed<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()> {
+pub fn atomic_write_if_changed<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()> {
```
epage, line 188: This is a bit confusing of a name.  Its "'atomic write' if changed" but could be read as "atomic 'write if changed'". which can't be atomic.

crates/cargo-util/src/paths.rs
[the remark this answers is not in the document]
epage: The bug is "wiped my file clean on ctrl-c" (#11386).  I assume these "atomic writes" would instead be "I've got a random temp file on ctrl-c".  Only slightly better.  If that is the case, then I would expect this to only be applied to the cases where the empty file causes problems.

tompscanlan:
@marcospb19 do you plan on getting back to this?

marcospb19:
> @marcospb19 do you plan on getting back to this?

Not right now, feel free to pick it up if you want to.

tompscanlan:
Thanks, and I hope you're doing well.

marcospb19:
Ty! Closing in favor of https://github.com/rust-lang/cargo/pull/12744.

Issue #12087, opened 2023-05-05
`cargo config get` doesn't seem to pick up RUSTFLAGS or RUSTDOCFLAGS


LukeMathWalker:
### Problem

`cargo config get build.rustdocflags` and `cargo config get build.rustflags` do not seem to behave as I would have expected.
If the respective environment variables are set (i.e. `RUSTDOCFLAGS` and `RUSTFLAGS`) but no value is set in `.cargo/config.toml`, I'd expect the environment variable value to be returned.

Instead, the command fails with:

```text
error: config value `build.rustdocflags` is not set
```

### Steps

```bash
export RUSTDOCFLAGS="--cfg tokio_unstable"
cargo -Z unstable-options config get build.rustdocflags
```

and 

```bash
export RUSTFLAGS="--cfg tokio_unstable"
cargo -Z unstable-options config get build.rustflags
```

### Notes

Is this intended behaviour?

weihanglo:
I would say it is kinda intentional. The Cargo configuration system only cares about environment variables starting with `CARGO_`[^1][^2]. Or more precisely, those that listed in [the doc] are officially supported; those not may be subject to change. 

Try this instead:

```
CARGO_BUILD_RUSTFLAGS="--cfg tokio_unstable" cargo +nightly -Z unstable-options config get build.rustflags
```

However, we must admit that the story around rustflags management is a bit messy.

[^1]: https://github.com/rust-lang/cargo/blob/2d693e20ea2dd3a55af94c56372c960503a6c673/src/cargo/util/config/key.rs#L31
[^2]: https://doc.rust-lang.org/cargo/reference/config.html#environment-variables

[the doc]: https://doc.rust-lang.org/cargo/reference/config.html

LukeMathWalker:
I see, but the [docs](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags) do mention `RUSTFLAGS` as well:

```
build.rustflags

    Type: string or array of strings
    Default: none
    Environment: CARGO_BUILD_RUSTFLAGS or CARGO_ENCODED_RUSTFLAGS or RUSTFLAGS

Extra command-line flags to pass to rustc. The value may be an array of strings or a space-separated string.
```

LukeMathWalker:
The usecase here is being able to accurately predict what extra flags will be passed when invoking `rustc` or `rustdoc` without actually performing the invocation.
If `cargo config get` shows a value that differs from what `cargo` actually does, its value is greatly diminished.

[...]

PR #12367, opened 2023-07-17
Add clarifications to environment-variables.md that will help people not break their builds when targeting specific CPUs

alexpyattaev:
Added clarification for use of RUSTFLAGS variable (and some minor tweaks around it)

```RUSTFLAGS environment variable is too broad and should be avoided when doing any cross-compiling (or targeting specific CPU model). This is now clarified here.```

A confusing lack of information regarding the effect of *RUSTFLAGS environment variables. This is a doc change only.

### Additional information
In general, the only reason for this issue to be relevant is that Cargo profiles can not set rustflags (there is an open issue for that already). In the meantime, env variables are the way to go.

### DIFF
--- a/src/doc/src/reference/environment-variables.md
+++ b/src/doc/src/reference/environment-variables.md
@@ -54,10 +54,12 @@ system:
   see `CARGO_ENCODED_RUSTDOCFLAGS`.
 * `CARGO_ENCODED_RUSTDOCFLAGS` ---  A list of custom flags separated by `0x1f`
   (ASCII Unit Separator) to pass to all `rustdoc` invocations that Cargo performs.
-* `RUSTFLAGS` --- A space-separated list of custom flags to pass to all compiler
+* `RUSTFLAGS` --- A space-separated list of custom flags to pass to *all* compiler
   invocations that Cargo performs. In contrast with [`cargo rustc`], this is
   useful for passing a flag to *all* compiler instances. See
-  [`build.rustflags`] for some more ways to set flags. This string is
+  [`build.rustflags`] for some more ways to set flags which will only apply to the "production" code.
+  Setting RUSTFLAGS incorrectly may break your build if the flags supplied result in
+  platform-specific code that can not run on the build host. This string is
   split by whitespace; for a more robust encoding of multiple arguments,
   see `CARGO_ENCODED_RUSTFLAGS`.
 * `CARGO_ENCODED_RUSTFLAGS` --- A list of custom flags separated by `0x1f`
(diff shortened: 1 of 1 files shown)


weihanglo (commented):
Thanks for the effort!

I got some concerns around the change:

* What does “production code” mean here? Is it `--release` build or something else? 
* If I understand correctly, under some circumstances, proc macros and build scripts also get these rustflags. [`build.rustflags`][1] should have explained this already.
* In [`build.rustflags`][1] doc we already have some "Caution". We'd better consolidating rustflags warning in once place, and [`build.rustflags`][1] looks like a better place to me.
* Continue. Not only platform-specific flags could break the build. Any contradictory to what Cargo passes to rustc could potential break any build.

[1]: https://doc.rust-lang.org/nightly/cargo/reference/config.html#buildrustflags

weihanglo:
> In general, the only reason for this issue to be relevant is that Cargo profiles can not set rustflags (there is an open issue for that already). In the meantime, env variables are the way to go.

Yep but not only an open issue, it is an unstable feature [`-Zprofile-rustflags`](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#profile-rustflags-option) people can try today. Feel free to give feedback there :)

weihanglo:
As [aforementioned](https://github.com/rust-lang/cargo/pull/12367#pullrequestreview-1532478074), the doc of `build.rustflags` of `.cargo/config.toml` is a better place if we want everything to be in one place, and there is already a similar notice:

https://github.com/rust-lang/cargo/blob/5a9f611178f48cd8c11632aa3f18776e58a0e59d/src/doc/src/reference/config.md?plain=1#L479-L483

alexpyattaev:
After reading the doc for [build.rustflags](https://doc.rust-lang.org/nightly/cargo/reference/config.html#buildrustflags) (especially without reading the doc for the environment variables) a reader would be even more confused. 
The build.rustflags variable is said to be affected by environment variables "CARGO_BUILD_RUSTFLAGS or CARGO_ENCODED_RUSTFLAGS or RUSTFLAGS", but then 1 paragraph down from there it does not mention CARGO_BUILD_RUSTFLAGS at all (even though it clearly does affect compiler flags). I can go over that file and add patches to that too so it is consistent with the environment variables file. I suppose a better way would be to make environment.md file do nothing but contain pointers to the config.md, so that the actual descriptions are not duplicated (and thus are not out of sync w.r.t. each other.) What do you think?

weihanglo:
Yes. It is a bit confusing. [This paragraph](https://doc.rust-lang.org/nightly/cargo/reference/config.html#environment-variables) explains that `CARGO_BUILD_RUSTFLAGS` is actually equivalent to `build.rustflags` but in the disguise of an environment variable. The kind of mapping between env and config is the secret sauce of Cargo configuration system but something people do feel confused. See <https://github.com/rust-lang/cargo/issues/12087> for a similar discussion. 

Not sure how to write it down well without adding more complexity (more text usually mean more cognitive overhead). From tooling perspective, personally I do want a simpler way to inspect actual rustflags being passed in. (`cargo --verbose` does that in a "verbose` way though)

weihanglo:
I feel like we were a bit off rails. There were two issues in the discussion:

1. Incorrect `RUSTFLAGS` setting may break one's build, and there is no warning in the Environments chapter.
2. The confusion of the interaction and precedence of rustflags related env vars and configs.

For the first one, we have a **Caution** paragraph in `build.rustflags` so personally I don't think we need the duplication in environment doc.

For the second one, we also have `CARGO_TARGET_<triple>_RUSTFLAGS` not mentioned in `build.rustflags`. And sorry I currently have no idea how to reduce the confusion.

I am going to close this, and we can discuss any idea in #12087, like something this comment said <https://github.com/rust-lang/cargo/issues/12087#issuecomment-1537208641>.

Thank you for the pull request :)

alexpyattaev:
Fair enough. I suppose one day the proper targets configuration will become a thing and all this silly environment variable business can RIP.

PR #12368, opened 2023-07-17
Upgrade to indexmap v2

cuviper:
No code changes are required for the current uses of indexmap.

I also updated `toml_edit` to get its update to `indexmap 2`, and
`gix-hashtable` to align them all on `hashbrown 0.14`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -864,6 +864,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "equivalent"
+version = "1.0.1"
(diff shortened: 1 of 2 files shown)


cuviper:
Unfortunately, `imara-diff` is still on `hashbrown 0.12` -- see pascalkuthe/imara-diff#6 -- but that dependency duplication already existed. So this PR goes from `hashbrown` 0.12.3+1.13.2 to 0.12.3+0.14.0.

ehuss:
Thanks!

@bors r+

Issue #9455, opened 2021-05-03
Consider making the `src` cache read-only.


ehuss:
Registry dependencies get extracted into cargo home (in the `src` directory) with whatever metadata is in the tar file.  One issue with this is that the files are usually writeable. This can cause a problem if the user accidentally modifies these files, which breaks cargo's assumption that they are immutable and reusable. One way this can happen is that in some editors, when there is an error or warning, they may open those files to display the error/warning (particularly with macros).  The user may not realize that this is from a remote location, and may not understand the consequence of making changes.

We may want to consider making those files read-only when extracting them. This would help with confusing situations where the `src` cache is inadvertently corrupted.

This would not protect from general filesystem corruption, which is also an issue. This is also an issue with `git` dependencies, which may be more difficult to adjust permissions on.

There is some risk that this would introduce new problems. For example, if people are using tools to clean the src directory, and those tools have trouble with read-only files.

matthiaskrgr:
Oh that's an interesting idea!
While I fully agree that it makes sense to make the cache (or at least `.cargo/registry/src` and `.cargo/git/checkout`) write-only, it does indeed prevent cargo-cache from cleaning up anything;

````
$ cargo cache --autoclean
Clearing cache...

Warning: failed to recursively remove directory "/home/matthias/.cargo/registry/src".
error: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
Warning: failed to recursively remove directory "/home/matthias/.cargo/git/checkouts".
error: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
````

BUT I am already happy that it does not panic :D
If it turned out that `cargo-cache` does modify the cache in any way (anything that goes beyond just removing whole files) I would indeed consider this a bug.

Right now I am just running `remove_dir_all::remove_dir_all()` on directories without doing any checks whatsoever.
https://crates.io/crates/remove_dir_all
I never had a read-only cache in mind when writing `cargo-cache` but adding a little bit of extra logic which tries to change  file permission to read-write before removing them does not sound too difficult, assuming that we have a crate around that handles file permissions on linux, macos and windows.

eminence:
I'm bringing up a new rust user on my team, and I found that they ended up editing some files in $CARGO_HOME  (which is actually quite easy to do with VSCode+RA) to add some debugging messages.   A read-only src cache might have given them a sufficient hint that what they were doing was not the right approach.

weihanglo:
[...]

PR #12369, opened 2023-07-17
fix: only skip mtime check on `~/.cargo/{git,registry}`

### LINKED ISSUE #12090: `cargo run` and `cargo build` don't seem to reflect recent changes
### Problem

i've been using vscode and 
```
anand@anands-MacBook-Pro ~/.c/d/lorenz (main)> cargo -V
cargo 1.69.0 (6e9a83356 2023-04-12)
anand@anands-MacBook-Pro ~/.c/d/lorenz (main)> rustup -V
rustup 1.26.0 (5af9b9484 2023-04-05)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.69.0 (84c898d65 2023-04-16)`
```
and i have a simple lib. when i change some stuff then rerun `cargo run` it runs the old code. 
i found this which seems similar to my issue
https://stackoverflow.com/questions/57087386/cargo-run-does-not-reflect-any-code-changes-in-main-rs

but in my case the only way that i reliably found to get cargo to register the changes is by deleting the `target` folder and rerunning. 

happy to provide more information if needed! i am probably doing something silly (pretty new to rust) but im confused

### Steps

_No response_

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

_No response_

jo1gi:
Fixes #12090

Make cargo only skip mtime checks on `$CARGO_HOME/{git, registry}`.

### DIFF
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -1857,14 +1857,27 @@ where
         Err(..) => return Some(StaleItem::MissingFile(reference.to_path_buf())),
     };
 
+    let skipable_dirs = if let Ok(cargo_home) = home::cargo_home() {
+        let skipable_dirs: Vec<_> = ["git", "registry"]
+            .into_iter()
+            .map(|subfolder| cargo_home.join(subfolder))
+            .collect();
+        Some(skipable_dirs)
+    } else {
+        None
+    };
+
     for path in paths {
         let path = path.as_ref();
 
-        // Assuming anything in cargo_home is immutable (see also #9455 about marking it readonly)
-        // which avoids rebuilds when CI caches $CARGO_HOME/registry/{index, cache} and
-        // $CARGO_HOME/git/db across runs, keeping the content the same but changing the mtime.
-        if let Ok(true) = home::cargo_home().map(|home| path.starts_with(home)) {
-            continue;
+        // Assuming anything in cargo_home/{git, registry} is immutable
+        // (see also #9455 about marking the src directory readonly) which avoids rebuilds when CI
+        // caches $CARGO_HOME/registry/{index, cache} and $CARGO_HOME/git/db across runs, keeping
+        // the content the same but changing the mtime.
+        if let Some(ref skipable_dirs) = skipable_dirs {
+            if skipable_dirs.iter().any(|dir| path.starts_with(dir)) {
+                continue;
+            }
         }
         let path_mtime = match mtime_cache.entry(path.to_path_buf()) {
             Entry::Occupied(o) => *o.get(),
--- a/tests/testsuite/freshness.rs
+++ b/tests/testsuite/freshness.rs
@@ -14,7 +14,8 @@ use super::death;
 use cargo_test_support::paths::{self, CargoPathExt};
 use cargo_test_support::registry::Package;
 use cargo_test_support::{
-    basic_manifest, is_coarse_mtime, project, rustc_host, rustc_host_env, sleep_ms,
+    basic_lib_manifest, basic_manifest, is_coarse_mtime, project, rustc_host, rustc_host_env,
+    sleep_ms,
 };
 
 #[cargo_test]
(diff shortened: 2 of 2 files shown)


src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1860,11 +1860,18 @@ where
...
-            continue;
+        // Assuming anything in cargo_home/{git, registry} is immutable
+        // (see also #9455 about marking cargo_home readonly) which avoids rebuilds when CI caches
+        // $CARGO_HOME/registry/{index, cache} and $CARGO_HOME/git/db across runs, keeping the
+        // content the same but changing the mtime.
+        if let Ok(cargo_home) = home::cargo_home() {
+            let is_readonly = ["git", "registry"]
+                .into_iter()
+                .map(|subfolder| cargo_home.join(subfolder))
+                .any(|readonly_dir| path.starts_with(readonly_dir));
```
weihanglo, line 1871: * Maybe just a micro optimization. This `cargo_home.join` is in a hot loop over `paths`. Could we do the allocation out of the loop upfront?
* I am a big wary that we now duplicate the path construction logic. Do you feel it better if we reuse the code [here in `Config`](https://github.com/rust-lang/cargo/blob/20df9e40a4d41dd08478549915588395e55efb4c/src/cargo/util/config/mod.rs#L353-L361)?

weihanglo (commented):
Thanks for the contribution! Could you add a test to verify that only registry and git path are ignored? Maybe by ensuring mtime check won't skip `~/.cargo/bin`.

weihanglo:
Pardon my title editing 🙇🏾.

src/cargo/core/compiler/fingerprint/mod.rs
jo1gi: I will take a look at it. Should I combine the commits, when I'm done?

weihanglo: My personal preference is making each commit atomic. Given the size of this change is not huge, it makes sense to me just rebase and leave one or two commits.

More personally, I may have the first commit to have a test first asserting the "bug behavior". Then the second contains the actual patch and the fixup of the test.

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1860,11 +1860,18 @@ where
     for path in paths {
         let path = path.as_ref();
 
-        // Assuming anything in cargo_home is immutable (see also #9455 about marking it readonly)
-        // which avoids rebuilds when CI caches $CARGO_HOME/registry/{index, cache} and
-        // $CARGO_HOME/git/db across runs, keeping the content the same but changing the mtime.
-        if let Ok(true) = home::cargo_home().map(|home| path.starts_with(home)) {
-            continue;
+        // Assuming anything in cargo_home/{git, registry} is immutable
+        // (see also #9455 about marking cargo_home readonly) which avoids rebuilds when CI caches
```
tamird, line 1864: FWIW: #9455 is not about marking CARGO_HOME readonly, but rather only a subdirectory of it.

tests/testsuite/build.rs
```diff
@@ -5972,6 +5972,62 @@ fn build_with_relative_cargo_home_path() {
...
+    p.cargo("run")
+        .env("CARGO_HOME", &cargo_home)
+        .with_stdout("1")
+        .run();
+    // Run with cache
+    p.change_file("src/main.rs", r#"fn main() { println!("2"); }"#);
+    p.cargo("run")
+        .env("CARGO_HOME", &cargo_home)
+        .with_stdout("1")
+        .run();
```
weihanglo, line 6000: The test looks like. Just some minor details:

* We usually assert `stderr` instead of `stdout`, so that we can avoid `cargo build` and instead use `cargo check`. That makes testsuite run a bit faster.
* When asserting the output for compilation freshness, we usually use `--verbose/-v` to assert `[FRESH]` status unless there is a need of verifying the behavior of final artifacts.
* Changed the `lib.rs` content to some illegal syntax so we can observe whether the build fails or not.
* `env` accepts `AsRef<OsStr>` so `to_str()` is unnecessary.


```suggestion
    let p = project()
        .at("cargo_home/registry/foo")
        .file("Cargo.toml", &basic_lib_manifest("foo"))
        .file("src/lib.rs", "")
        .build();
    let project_root = p.root();
    let cargo_home = project_root
        .parent()
        .unwrap()
        .parent()
        .unwrap();
    p.cargo("check -v")
        .env("CARGO_HOME", &cargo_home)
        .with_stderr(
            "\
[CHECKING] foo v0.5.0 ([CWD])
[RUNNING] `rustc --crate-name foo src/lib.rs [..]
[FINISHED] dev [..]",
        )
        .run();
    // Run with cache
    p.change_file("src/lib.rs", "illegal syntax");
    p.cargo("check -v")
        .env("CARGO_HOME", &cargo_home)
        .with_stderr(
            "\
[FRESH] foo v0.5.0 ([CWD])
[FINISHED] dev [..]",
        )
        .run();
```

tests/testsuite/build.rs
```diff
@@ -5972,6 +5972,62 @@ fn build_with_relative_cargo_home_path() {
...
+        .file("src/main.rs", r#"fn main() { println!("1"); }"#)
+        .build();
+    let project_root = p.root();
+    let cargo_home = project_root
+        .parent()
+        .unwrap()
+        .parent()
+        .unwrap()
+        .to_str()
+        .unwrap();
```
weihanglo, line 6018: Should be only one level up, right?

```suggestion
        .unwrap();
```

tests/testsuite/build.rs
```diff
@@ -5972,6 +5972,62 @@ fn build_with_relative_cargo_home_path() {
...
+    // Run with cache
+    p.change_file("src/main.rs", r#"fn main() { println!("2"); }"#);
+    p.cargo("run")
+        .env("CARGO_HOME", &cargo_home)
+        .with_stdout("1")
+        .run();
+}
+
+#[cargo_test]
+fn use_mtime_cache_in_cargo_home() {
```
weihanglo, line 6004: Could you apply the same suggestion gave to `skip_mtime_check_in_selected_cargo_home_subdirs` here?

weihanglo (commented):
Thanks for adding tests! 

Could we move these two tests to `tests/testsuite/freshness.rs`. It makes more sense to put them there.

tests/testsuite/build.rs
jo1gi: Thank you for the help.

weihanglo (commented):
We're almost there! Could you run `cargo fmt --all` to address the format error?

weihanglo:
Thanks @jo1gi 

@bors r+

PR #12380, opened 2023-07-19
chore: Don't update test data

epage:
Inspired by #12378

There is a preset for ignores but I feel like we'd more likely want to be able to easily see whats actuazlly ignored.


weihanglo (approved):

weihanglo:
Thanks for this. And yes we can add more ignores when needed.

@bors r+

ehuss:
Thanks!

@bors r+

Issue #12347, opened 2023-07-11
version-bump check doesn't know a subcrate hasn't bumped on nightly if we've bumped it on beta


weihanglo:
### Problem

We've added `cargo unpublished` xtask and `ci/validate-version-bump.sh` to help the project to detect if a pull request has touched any subcrate and needs a version bump for it. This works well when there is no bump on beta branch. Bad thing happens when we already have a bump on the current beta but not yet published, `cargo published` would assume we've already bumped that crate.

In reality, we should consider crate version in beta branch as “published” and check the version bump against that instead of published one on crates.io. Every crate gets into beta branch of the 6 weeks release window should be considered as “published”.

### Steps

In https://github.com/rust-lang/cargo/pull/12310, the check thought `crates-io` crate has been bumped. Actually it is required to update to 0.37.0, since 0.36.1 will be published with 1.72.0, but #12310 is targeted at 1.73.0. or 1.74.0 if merged.

### Possible Solution(s)

It is quite tricky to solve this. Subcrates are not necessary to follow the train release model with Rust and Cargo. However, there are some requirements we need to pay attention to

* Subcrates are somehow loosely bound to Cargo. Sometimes a specific change is required by a certain version of Cargo, so it needs to be released when cargo gets released.
* We may not want to have a separate release cadence for each subcrate. It would be a maintenance burden.
* We may not want to release a subcrate when the cargo version it is bound to has yet hit stable, since everything could change before that.

I don't have a good answer for the release model of subcrates, but here is a proposal that may make it more explicit:

* A subcrate version can contain build metadata to show with which cargo version it is supposed to be released. That is, for `crates-io` 0.36.1 it would be `0.36.1+1.71.0`, so that is pretty clear it is targeted to be released with Rust 1.71.0.
* If we need to backport to a subcrate on beta, for example `crates-io@0.36.1` to `0.37.0`, then we just bump it to `0.37.0+1.71.0`. We then need a xtask automation to detect that on beta it has got a version bump so on nightly we need to adjust accordingly.

### Version

This is about project infra but let me record the point of time it happened.

```text
cargo 1.70.0 (ec8a8a0ca 2023-04-25)
release: 1.70.0
commit-hash: ec8a8a0cabb0e0cadef58902470f6c7ee7868bdc
commit-date: 2023-04-25
```

weihanglo:
I believe `cargo-semver-checks` could help this part.

> ## Does the crate I'm checking have to be published on crates.io?
[...]

PR #12381, opened 2023-07-19
fix: Set MSRV for internal packages

epage:

Correctly communicates the MSRV we support to our users.

For packages that are more generally mean to be used by other people, I've punted on for now (see ehuss' comment on this PR).  We'll likely need to figure this out for #12432 though

### Additional information

This is prep for a future change which will have us use a fixed rust version for the semver tests with a PR updating them.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -8,7 +8,33 @@
   ignorePaths: [
     "**/tests/**",
   ],
+  regexManagers: [
+    {
+      customType: 'regex',
+      fileMatch: [
+        '^Cargo.toml$',
+      ],
+      matchStrings: [
+        'rust-version.*?(?<currentValue>\\d+\\.\\d+(\\.\\d+)?)',
+      ],
+      depNameTemplate: 'latest-msrv',
+      packageNameTemplate: 'rust-lang/rust',
+      datasourceTemplate: 'github-releases',
+    },
+  ],
   packageRules: [
+    {
+      commitMessageTopic: 'Latest MSRV',
+      matchManagers: [
+        'regex',
+      ],
+      matchPackageNames: [
+        'latest-msrv',
+      ],
+      schedule: [
+        '* * * * *',
+      ],
+    },
     // Goals:
     // - Rollup safe upgrades to reduce CI runner load
     // - Have lockfile and manifest in-sync (implicit rules)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -427,7 +427,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-util"
-version = "0.2.6"
+version = "0.2.7"
 dependencies = [
  "anyhow",
  "core-foundation",
@@ -588,7 +588,7 @@ dependencies = [
 
 [[package]]
 name = "crates-io"
-version = "0.38.0"
+version = "0.38.1"
 dependencies = [
  "curl",
  "percent-encoding",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ exclude = [
 ]
 
 [workspace.package]
+rust-version = "1.71.0"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 
--- a/benches/benchsuite/Cargo.toml
+++ b/benches/benchsuite/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "benchsuite"
 version = "0.0.0"
+rust-version.workspace = true
 edition.workspace = true
 license.workspace = true
 homepage = "https://github.com/rust-lang/cargo"
--- a/benches/capture/Cargo.toml
+++ b/benches/capture/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "capture"
 version = "0.1.0"
+rust-version.workspace = true
 edition.workspace = true
 license.workspace = true
 description = "Tool for capturing a real-world workspace for benchmarking."
--- a/crates/cargo-test-macro/Cargo.toml
+++ b/crates/cargo-test-macro/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "cargo-test-macro"
 version = "0.1.0"
+rust-version.workspace = true
 edition.workspace = true
 license.workspace = true
 homepage = "https://github.com/rust-lang/cargo"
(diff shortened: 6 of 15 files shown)


weihanglo (commented):
It is technically true that Cargo-the-library is only guaranteed to be compatible with one corresponding Rust version. And this also gets rid off some burdens we would have on MSRV. However, it is a bit too strict for every crate to stick to the latest stable. Someone may only depends on a tiny part of cargo-the-library but with this MSRV bump that couldn't anymore.

weihanglo:
> This is prep for a future change which will have us use a fixed rust version for the semver tests with a PR updating them.

What will this look like? I didn't get the full picture maybe because of missing this?

epage:
>  Someone may only depends on a tiny part of cargo-the-library but with this MSRV bump that couldn't anymore.

Unless its behind a feature flag or relying on a behavior change, MSRV is all-or-nothing.

weihanglo:
Some crates like `crates-io` and `home` are pretty not directly coupled with `cargo`, feel like they shouldn't be set with the same MSRV.

(and perhaps `cargo-util` if we have a firmed definition of what utils are…)

epage:
> Some crates like crates-io and home are pretty not directly coupled with cargo, feel like they shouldn't be set with the same MSRV.

The choices I'm aware of
- Don't set MSRV and put it in users hands
  - Short term: users have to play around to find the version that works
  - Long term: with MSRV aware resolver, users will still have to play around to find the version that works
- Set to latest
  - Might be too high in some cases
  - Easy to verify
- Set different MSRVs
  - To ensure accuracy, we'd need to add MSRV check jobs
  - We'd need to hand-manage the list of which packages to verify with which MSRV or write a tool
- Manage multiple workspaces, one of the core of cargo and one for more external stuff
  -This might be of more interest if we start taking responsibility for Rust APIs for cargo (e.g. build scripts, metadata)

Considering
- `home` is primarily for cargo (#12386) (and `cargo-util` looks similar)
- So far, we've not been looking to take on the maintenance load of clients for our crates (making them more "buyer beware")
- We want our process to scale with adding more packages, including CI

To me, it seems like its a choice to no MSRV or latest (until / unless we decide to have multiple workspaces) and setting it would be useful so I would lean towards that.

Thoughts?

weihanglo:
BTW, version bump check is somehow broken. See <https://github.com/rust-lang/cargo/issues/12347>. You may need to manually check if it does need a bump for nightly.

(I'll try fixing that in the next few days)

ehuss:
I'm going to check my box, though there were some mild concerns that I think were already highlighted above, but I wanted to repeat back to make sure they are understood.

- There is a desire to make it easy to add new packages, but if we had independent MSRV for different packages, that might require some manual management to make sure they are set and tested in CI correctly.
- Adding new syntax to our workspace `Cargo.toml` that wasn't forwards compatible would force us to hold on making that change until that support hit the MSRV version. For example, if we used something like the recently added named-debug options, we would have to hold off on dogfooding it for a long while.
- Something like `build-rs` (#12432) would need to be mindful of which versions of cargo it supports. For example, if we had a package like that, and we switched it to use the new `cargo::` syntax, then we could only make that change once that support passes the MSRV.

Some possible alternate solutions:
- Multiple workspaces.
- Have an xtask to help manage things.
- Allow commands like `cargo check` to skip unsupported rust versions (either built-in to cargo, or an xtask like thing). This wouldn't help with manifest syntax changes.
- Don't set MSRV at all for now.
- Set MSRV, but not for every package (like the credential providers).

My concerns were mostly around the credential providers, cargo-platform, and home. For now, credential providers are unstable, so the exact MSRV isn't important. But I think once we head towards stabilization, we should try to figure out a solution. I think it will be important to make sure that people can `cargo install` providers with some MSRV older than latest stable.

`cargo-platform` almost never changes, and it will be stable-2 when it publishes, which is the MSRV I would suggest anyway. The only concern is if we need to do an out-of-band publish, which I don't remember if we've ever done, but would be very exceptional. That seems like something that we can improve later.

`home` is in a weird state, but I don't think we will be releasing a new version in the short term. (There are some unpublished version bumps, but I don't know what's going on there.) This also seems like something we can figure out, but we need to be mindful of.

I would have a slight preference towards not setting the MSRV in the short term for the things that might cause problems (like cargo-platform or home), just so that this doesn't introduce new pain points.

epage:
>  but I wanted to repeat back to make sure they are understood.
>
> - Adding new syntax to our workspace Cargo.toml that wasn't forwards compatible would force us to hold on making that change until that support hit the MSRV version.

There is also the issue that shared dependencies need to meet the lowest common denominator of the MSRVs.

epage:
I've gone ahead and limited this to more internal scripts, punting on how to handle MSRV testing for more public packages to #12432

epage:
@Turbo87 mind double checking my RenovateBot regex manager?

.github/renovate.json5
```diff
@@ -8,7 +8,32 @@
   ignorePaths: [
     "**/tests/**",
   ],
+  regexManagers: [
+    {
+      fileMatch: [
+        '^Cargo.toml$',
+      ],
+      matchStrings: [
+        'rust-version.*?(?<currentValue>\\d+\\.\\d+(\\.\\d+)?)',
```
Turbo87, line 17:
```suggestion
        'rust-version\\s*=\\s*"(?<currentValue>\\d+\\.\\d+(\\.\\d+)?)"',
```

I would probably go with this slightly more specific regex, but I think it should work either way :)

I'm using something roughly similar in https://github.com/Turbo87/renovate-config/blob/04f6fb54cca78973deacec582fb36b4bf5f2da9c/rust/updateToolchain.json#L8

weihanglo (approved):

.github/renovate.json5
```diff
@@ -8,7 +8,33 @@
...
+      customType: 'regex',
+      fileMatch: [
+        '^Cargo.toml$',
+      ],
+      matchStrings: [
+        'rust-version.*?(?<currentValue>\\d+\\.\\d+(\\.\\d+)?)',
+      ],
+      depNameTemplate: 'latest-msrv',
+      packageNameTemplate: 'rust-lang/rust',
+      datasourceTemplate: 'github-releases',
```
weihanglo, line 22: Just mind that `1.71.1` is not yet in [GitHub Release](https://github.com/rust-lang/rust/releases).

weihanglo (approved):

weihanglo:
Just realized, with this pull request and #12395, it is inevitable that every time a Rust release comes out, `cargo-util` and `crates-io` will be forced to release a version due to the bump of `package.rust-version`, even when there is no actual code change involved.

Is that something we've considered? It feels like a bit unnecessary.

Issue #9930, opened 2021-09-21
MSRV-dependent dependency version resolution


newpavlov:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/issues/9930)*

Based on [RFC 2495](https://github.com/rust-lang/rfcs/pull/2495) Rust now supports the [`rust-version`](https://doc.rust-lang.org/nightly/cargo/reference/manifest.html#the-rust-version-field) field, which currently works as:

>If the currently selected version of the Rust compiler is older than the stated version, cargo will exit with an error, telling the user what version is require

This behavior is not ideal. If a project depends on crate `foo v0.1.0` with MSRV 1.56, then releasing crate `foo v0.1.1` with MSRV 1.60 will break build of the project on older toolchains after simple `cargo update`. Ideally Cargo would select `foo v0.1.0` on older tolchains, thus preventing the breakage.

RFC 2495 has [described](https://github.com/rust-lang/rfcs/blob/master/text/2495-min-rust-version.md#influencing-version-resolution) MSRV-dependent version resolution as a potential future extension. This issue is intended for design discussions and tracking implementation progress of this feature.

Third-party support
- dependabot/dependabot-core#5423
- renovatebot/renovate#26060

---

This has been approved as [RFC #3537](https://rust-lang.github.io/rfcs/3537-msrv-resolver.html).

Implementation by stabilization milestone
- [x] Improve the `cargo install` error (#12798)
- [x] Improve documentation around `rust-version` (#13056)
- [x] cargo-add (#13608)
  - [x] `cargo add` support (#10653) 
  - [x] Infer users rust-version from `rustc -V` (#13516)
- [x] Attempt to report the MSRV error top-down, rather than bottom up (#13514)
- [x] `cargo update` keeps user informed when using old versions (#13372)
- [x] #13539
- [x] Resolver (#14639, #14754)
  - [x] Prototype (#12560)
  - [x] Switch resolver to "prefer" rather than "enforce" (#12950)
  - [x] De-prioritize dep versions without `rust-version` (#13066)
  - [x] Infer users rust-version from `rustc -V` (#13743)
  - [x] Existing `--ignore-rust-version` support (#13738)
  - [x] `--ignore-rust-version` support for  `cargo update` / `cargo-generate-lockfile` (#13742)
  - [x] #13540
  - [x] Add resolver v3 (#13776)
  - [x] Verify resolver 3 doesn't change cargo-install (#13769, #13790)
[...]

PR #12382, opened 2023-07-19
fix: Change the defaults to always check-in `Cargo.lock`

### LINKED ISSUE #8728: Check Cargo.lock in version control for libraries

[The official FAQ](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) and [various other places](https://github.com/rust-lang/cargo/issues/315) strongly suggest (almost mandate) that libraries do not check in Cargo.lock. This serves so that if the library is used as a dependency from another crate, the parent crate should get to decide what particular versions of dependencies to use for itself and all its transitive dependencies, respecting versioning restrictions indicated in the relevant Cargo.toml files.

- strongly suggest versioning Cargo.lock in version control even for libraries; without this, it is practically impossible to achieve reproducibility of builds, tests and CI pipelines: the same exact version (commit checksum) of a library may compile / pass at some point in time on someone's machine, but spuriously break / fail later on or on some other machine, just because some dependency published a new version on crates.io
- add an option to `cargo` to skip packaging the local Cargo.lock when publishing a library to crates.io (by default it will exclude it for libraries and include it for binaries)
  + perhaps it is sufficient to [manually exclude](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields) `Cargo.lock` from a package until this is implemented
- suggest relying on https://dependabot.com/rust/ or manually bumping dependencies in order to detect incompatibilities with new version of dependent crates in an automated way that is tracked under version control and maintains the hermeticity and reproducibility of builds, even for libraries

**Notes**

epage:

Having libraries leave `Cargo.lock` "on the float" has been serving the ecosystem well, including
- Encouraging people to validate that latest dependencies work
- Encouraging ecosystem-wide health by acting as a "distributed crater"

These benefits are limited though.  The policy is inconsistent between workspaces with or without `[[bin]]`s, reducing the affect of testing the latest.  This is also subject to when CI last ran; for passively maintained projects, there is little coverage of new dependencies.

There are also costs associated with this policy
- `git bisect` is using an unpredictable set of dependencies, affecting the ability to identify root cause
- This is another potential cause for Red CI / broken local development if version is yanked or a bug is introduced
  - Impacting the perceived level of quality for a project
  - Confusing to new contributors who might not recognize why CI failed and assume its their fault
  - Requiring context switching from maintainers to get fixes in

In particular, since this policy was decided, there has been an increased interest in supporting an MSRV (as recently as v1.56.0, cargo gained support for specifying a package's MSRV).   This has led to long discussions on *what* MSRV a package should use (e.g. rust-lang/libs-team#72,. time-rs/time#535).  Worst, there has been a growing trend for people to set an non-semver upper bound on dependencies, making it so packages can't work well with other packages (see #12323).  Tooling support would help with this (#9930) but the sooner we address this, the less entrenched bad practices will be.

On the positive side, since the policy was decided
- In general, CI became easier to setup and maintain with Github Actions compared to TravisCI
- Dependabot went GA on Github in 2021 (https://github.blog/changelog/2021-03-31-dependabot-version-updates-are-now-generally-available)
  - I believe Dependabot will post security update PRs even when Dependabot is not more generally enabled

So to get some of the benefit from not checking in `Cargo.lock`, we can recommend either automatically applying updates or having CI check the latest dependencies in a way to get this out of the critical path of PRs and releases.

Since there is no one right answer on how to solve all of these problems, we're documenting these trade offs so people can make the choice that is most appropriate for them.  However, we are changing the default to a consistent "always check it in" as the answer for those who don't want to think about it.

Prior art
- [Yarn (Javascript)](https://yarnpkg.com/getting-started/qa#should-lockfiles-be-committed-to-the-repository)
- [Poetry (Python)](https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control)
- [Bundler (Ruby)](https://bundler.io/guides/faq.html#using-gemfiles-inside-gems)

Fixes #8728

Please review per-commit.  I tried to minimize changes I made to the structure of the CI document

In #8728, I brought up having a CI reference page.  I keep going back and forth on whether this is guide-level material or reference-level material.  Obviously, right now I'm leaning towards it being guide-level.

### Additional information

This changes cargo from telling people what to do to giving them a starting point, or default, and giving them the information to make their own choice, if needed.

So the question for defaults is who are we targeting? For a default path in documentation, it would be for new to intermediate users. As for `cargo new`, we've been prioritizing new users over those that run it frequently (boiler plate comment, bin is default, etc).

See #8728 for the FCP on this policy change

### DIFF
--- a/src/doc/src/faq.md
+++ b/src/doc/src/faq.md
@@ -102,33 +102,45 @@ issue][cargo-issues].
 
 [cargo-issues]: https://github.com/rust-lang/cargo/issues
 
-### Why do binaries have `Cargo.lock` in version control, but not libraries?
+### Why have `Cargo.lock` in version control?
+
+Whether you do is dependent on the needs of your package.
 
 The purpose of a `Cargo.lock` lockfile is to describe the state of the world at
-the time of a successful build. Cargo uses the lockfile to provide
-deterministic builds on different times and different systems, by ensuring that
-the exact same dependencies and versions are used as when the `Cargo.lock` file
-was originally generated.
+the time of a successful build.
+Cargo uses the lockfile to provide deterministic builds at different times and
+on different systems,
+by ensuring that the exact same dependencies and versions are used as when the
+`Cargo.lock` file was originally generated.
 
-This property is most desirable from applications and packages which are at the
-very end of the dependency chain (binaries). As a result, it is recommended that
-all binaries check in their `Cargo.lock`.
+Deterministic builds help with
+- Running `git bisect` to find the root cause of a bug
+- Ensuring CI only fails due to new commits and not external factors
+- Reducing confusion when contributors see different behavior as compared to
+  other contributors or CI
 
-For libraries the situation is somewhat different. A library is not only used by
-the library developers, but also any downstream consumers of the library. Users
-dependent on the library will not inspect the library’s `Cargo.lock` (even if it
-exists). This is precisely because a library should **not** be deterministically
-recompiled for all users of the library.
+Having this snapshot of dependencies can also help when projects need to be
+verified against consistent versions of dependencies, like when
+- Verifying a minimum-supported Rust version (MSRV) that is less than the latest
+  version of a dependency supports
+- Verifying human readable output which won't have compatibility guarantees
+  (e.g. snapshot testing error messages to ensure they are "understandable", a
+  metric too fuzzy to automate)
 
-If a library ends up being used transitively by several dependencies, it’s
-likely that just a single copy of the library is desired (based on semver
-compatibility). If Cargo used all of the dependencies' `Cargo.lock` files,
-then multiple copies of the library could be used, and perhaps even a version
-conflict.
+However, this determinism can give a false sense of security because
+`Cargo.lock` does not affect the consumers of your package, only `Cargo.toml` does that.
+For example:
+- [`cargo install`] will select the latest dependencies unless `--locked` is
+  passed in.
+- New dependencies, like those added with [`cargo add`], will be locked to the latest version
 
-In other words, libraries specify SemVer requirements for their dependencies but
-cannot see the full picture. Only end products like binaries have a full
-picture to decide what versions of dependencies should be used.
+The lockfile can also be a source of merge conflicts.
+
+For strategies to verify newer versions of dependencies via CI,
+see [Verifying Latest Dependencies](guide/continuous-integration.md#verifying-latest-dependencies).
+
+[`cargo add`]: commands/cargo-add.md
+[`cargo install`]: commands/cargo-install.md
 
 ### Can libraries use `*` as a version for their dependencies?
 
--- a/src/doc/src/faq.md
+++ b/src/doc/src/faq.md
@@ -104,7 +104,8 @@ issue][cargo-issues].
 
 ### Why have `Cargo.lock` in version control?
 
-Whether you do is dependent on the needs of your package.
+While [`cargo new`] defaults to tracking `Cargo.lock` in version control,
+whether you do is dependent on the needs of your package.
 
 The purpose of a `Cargo.lock` lockfile is to describe the state of the world at
 the time of a successful build.
@@ -139,6 +140,7 @@ The lockfile can also be a source of merge conflicts.
 For strategies to verify newer versions of dependencies via CI,
 see [Verifying Latest Dependencies](guide/continuous-integration.md#verifying-latest-dependencies).
 
+[`cargo new`]: commands/cargo-new.md
 [`cargo add`]: commands/cargo-add.md
 [`cargo install`]: commands/cargo-install.md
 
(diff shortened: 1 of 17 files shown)


epage:
I've marked this as a draft until the FCP is closed.

weihanglo (commented):
Haven't really looked into it yet. Should we take care of source files traversal and `cargo package` as well?

* https://github.com/rust-lang/cargo/blob/1b15556767f4b78a64e868eedf4073c423f02b93/src/cargo/ops/cargo_package.rs#L268-L274
* https://github.com/rust-lang/cargo/blob/1b15556767f4b78a64e868eedf4073c423f02b93/src/cargo/sources/path.rs#L182-L186

epage:
I believe both of those cases are more about what your users have access to.  In theory, users of a lib should not be affected by a `Cargo.lock`, so leaving it out should still be fine.

src/doc/src/faq.md
```diff
@@ -102,33 +102,45 @@ issue][cargo-issues].
...
+
+Deterministic builds help with
+- Running `git bisect` to find the root cause of a bug
+- Ensuring CI only fails due to new commits and not external factors
+- Reducing confusion when contributors see different behavior as compared to
+  other contributors or CI
+
+Having this snapshot of dependencies can also help when projects need to be
+verified against consistent versions of dependencies, like when
+- Verifying a minimum-supported Rust version that is less than the what latest
```
joshtriplett, line 125:
```suggestion
- Verifying a minimum-supported Rust version (MSRV) that is less than the what latest
```

src/doc/src/faq.md
```diff
@@ -102,33 +102,45 @@ issue][cargo-issues].
...
+  metric too fuzzy to automate)
+
+However, this determinism can give a false sense of security because
+`Cargo.lock` does not affect the consumers of your package, only `Cargo.toml` does that.
+For example:
+- [`cargo install`] will select the latest dependencies unless `--locked` is
+  passed in.
+- New dependencies, like those added with [`cargo add`], will be locked to the latest version
+
+For strategies to verify newer dependencies,
```
joshtriplett, line 138:
```suggestion
For strategies to verify newer versions of dependencies via CI,
```

Turbo87 (approved):
I very much support everything in this PR! ❤️

src/doc/src/faq.md
```diff
@@ -102,33 +102,45 @@ issue][cargo-issues].
...
+Deterministic builds help with
+- Running `git bisect` to find the root cause of a bug
+- Ensuring CI only fails due to new commits and not external factors
+- Reducing confusion when contributors see different behavior as compared to
+  other contributors or CI
+
+Having this snapshot of dependencies can also help when projects need to be
+verified against consistent versions of dependencies, like when
+- Verifying a minimum-supported Rust version (MSRV) that is less than the what latest
+  version of a dependency supports
```
joshtriplett, line 126: This still has a grammar issue (which my suggestion was intended to fix): "the what latest version of a dependency supports".

You could delete "what", or swap "what" and "the".

epage: Sorry about missing that

epage:
Forgot to add, this is not just a draft until the FCP window closes but the cargo team was also wanting to wait on this until after the next beta branch so we get 12 weeks of feedback before this hits stable.

Nemo157:
> I believe both of those cases are more about what your users have access to. In theory, users of a lib should not be affected by a `Cargo.lock`, so leaving it out should still be fine.

There are some usecases that use packaged `Cargo.lock` when building a library, like docs.rs, we attempt to build with the provided lockfile as we assume that gives the best chance of success, then delete it and try again if it fails (for packages that contain both library and binary crates and so distribute their lockfile). I also semi-regularly download and extract packages source code to inspect issues before going to find a repo, having a lockfile there can be useful.

epage:
Looks like [rust-1.73.0](https://github.com/rust-lang/cargo/tree/rust-1.73.0) has been branched so I assume we are good to merge.

weihanglo:
Ed, thanks for your effort to this! Also thanks everyone for providing lots of useful practices and considerations. Hold on. We're now full speed ahead!

@bors r+

src/doc/src/guide/continuous-integration.md
```diff
@@ -1,25 +1,8 @@
 ## Continuous Integration
 
-### Travis CI
+### Getting Started
```
Marcono1234, line 3: Is this header here really needed? And also, in general maybe the header levels have to be adjusted to be semantically more useful?

Currently the header hierarchy looks like this:
- Continuous Integration
  - Getting Started
  - GitHub Actions
  - GitLab CI
  - builds.sr.ht
  - Verifying Latest Dependencies

Maybe something like this would be better?
- Continuous Integration
  - _\<no sure how to name this\>_
    - GitHub Actions
    - GitLab CI
    - builds.sr.ht
  - Verifying Latest Dependencies

Marcono1234 (commented):
Sorry for adding a review this late; I hope the comments are useful nonetheless.

epage:
Fixed these in https://github.com/rust-lang/cargo/pull/12630

PR #12383, opened 2023-07-20
fix compiler BuildOutput

cuishuang:
BuildOutput information repeatedly output script_out_dir, replace with script_dir


weihanglo (commented):
Without looking closer into the change, I want to request a clearer PR description to explain the intent (this repositry does have a PR template).

Could you explain how it was broken, and how this fixes the issue? Could you also have a test to verify the change if possible? Thanks!

weihanglo:
Closing as the repeat doesn't look like a thing needing a fix. Please file an issue instead before open a pull request.

Issue #12387, opened 2023-07-21
Features of `[target.'cfg(...)'.dependencies]` and `[target.'cfg(not(...))'.dependencies]` are merged when used in a workspace


oblique:
### Problem

I have project in a workspace that has something like the following `Cargo.toml`:

```toml
[package]
name = "bug"
version = "0.1.0"
edition = "2021"

[dependencies]

[target.'cfg(target_arch = "wasm32")'.dependencies]
tokio = { version = "1", features = ["sync"] }

[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
tokio = { version = "1", features = ["net"] }
```

When I  was running `cargo build --target wasm32-unknown-unknown` then Tokio's `net` feature got enabled too and I was getting compilation errors because of `mio` crate.

This is an issue *only* if the project is in a workspace.

### Steps

1. `git clone https://github.com/oblique/cargo-features-bug`
2. `cargo build --target wasm32-unknown-unknown`

### Notes

I created a minimal example: https://github.com/oblique/cargo-features-bug

### Version

```text
cargo 1.71.0 (cfd3bbd8f 2023-06-08)
release: 1.71.0
commit-hash: cfd3bbd8fe4fd92074dfad04b7eb9a923646839f
commit-date: 2023-06-08
host: x86_64-unknown-linux-gnu
libgit2: 1.6.4 (sys:0.17.1 vendored)
libcurl: 8.0.1-DEV (sys:0.4.61+curl-8.0.1 vendored ssl:OpenSSL/1.1.1t)
ssl: OpenSSL 1.1.1t  7 Feb 2023
os: Arch Linux Rolling Release [64-bit]
```

oblique:
Worked after adding the following:

```toml
[workspace]
resolver = "2"
```

I didn't know that resolver needs to be explicitly set in case of a workspace.
I'm closing this.

PR #11082, opened 2022-09-13
docs(ref): Clarify workspace settings

epage:
### What does this PR try to resolve?

In reviewing the status of #10625, I was reminded
- that we are having growing pains with the workspace documentation
- that `workspace.resolver` isn't documented

So I re-organized the workspace docs to have a high level intro / behavior description and then to focus on being a field reference, much like `manifest.md`.   I could see splitting it specifically into tutorial/reference like the overriding dependencies document does it.

When adding `workspace.resolver`, I remembered in the nested workspace discussion there were other workspace related sections that are not present.  We now link out to `profile`, `patch`, and `replace`.  In doing this, I realized that `patch` and `replace` do not specify their workspace behavior, so I do that.

### How should we test and review this PR?

Look at it commit by commit to get more digestible chunks.  Unfortunately, the first commit didn't split up so easily.

### Additional information

Other information you want to mention in this PR, such as prior arts,
future extensions, an unresolved problem, or a TODO list.
-->

PR #10910, opened 2022-07-28
Warn when an edition 2021 crate is in a virtual workspace with default resolver

Nemo157:
Edition 2021 updates the default resolver to version "2", but developers using virtual workspaces commonly don't get this update because the virtual workspace defaults to version "1". Warn when this situation occurs so those developers can explicitly configure their workspace and will be more likely to know that they will need to update it in the future.

Fixes #10112

PR #10625, opened 2022-05-03
Add "resolver = 2" option to workspace manifest examples

mousetail:
It seems like `resolver = 2` is just a basic requirement to build any package with dependencies. To make the examples work it's needed. Apparently there is work to make it default (https://github.com/rust-lang/rust/issues/90148) but right now it's required.

### What does this PR try to resolve?

The examples in the documentation about workspaces don't work in the latest edition of rust. This would have saved me many hours of debugging.

Ideally there would also be a description of what the option actually does but I don't really understand that myself.

### How should we test and review this PR?

It's a documentation change. Check if you think the information is useful.

### Additional information

Relevant links:
- https://github.com/rust-lang/rust/issues/90148
- https://github.com/rust-lang/cargo/issues/9996
- https://doc.rust-lang.org/nightly/edition-guide/rust-2021/default-cargo-resolver.html#details

PR #12388, opened 2023-07-21
docs: raise awareness of resolver used inside workspace

zvolin:
Workspaces by default use the resolver in version 1. It's been some time already since the 2021 edition which made version 2 a default for alone packages, but yet most of the projects that make a use of the workspaces still depends on  an old resolver.

By being explicit about the resolver tag inside the workspace we can lower the usage of older resolver for the new projects.

This can raise the awareness of this behavior and prevent issues like https://github.com/rust-lang/cargo/issues/12387. I also wasn't aware of this behavior before while not being so new to Rust, and we have the resolver 2 for good reasons, so I think we should be more explicit about it in the documentation.
When someone looks for the 'how to make cargo workspace' answers, he's unlikely to get to the `Dependency Resolution` section at the same time, he'll likely just copy paste the workspace example from the `Workspaces` and call it a day, yet extending the usage of an old resolver and not benefiting from the new one.

### DIFF
--- a/src/doc/src/reference/workspaces.md
+++ b/src/doc/src/reference/workspaces.md
@@ -94,7 +94,6 @@ the workspace:
 [workspace]
 members = ["member1", "path/to/member2", "crates/*"]
 exclude = ["crates/foo", "path/to/other"]
-resolver = "2"
 ```
 
 All [`path` dependencies] residing in the workspace directory automatically
@@ -136,7 +135,6 @@ used:
 [workspace]
 members = ["path/to/member1", "path/to/member2", "path/to/member3/*"]
 default-members = ["path/to/member2", "path/to/member3/foo"]
-resolver = "2"
 ```
 
 When specified, `default-members` must expand to a subset of `members`.
@@ -168,7 +166,6 @@ Example:
 # [PROJECT_DIR]/Cargo.toml
 [workspace]
 members = ["bar"]
-resolver = "2"
 
 [workspace.package]
 version = "1.2.3"
@@ -203,7 +200,6 @@ Example:
 # [PROJECT_DIR]/Cargo.toml
 [workspace]
 members = ["bar"]
-resolver = "2"
 
 [workspace.dependencies]
 cc = "1.0.73"
@@ -236,7 +232,6 @@ configuration in `Cargo.toml`. For example:
 ```toml
 [workspace]
 members = ["member1", "member2"]
-resolver = "2"
 
 [workspace.metadata.webcontents]
 root = "path/to/webproject"
--- a/src/doc/src/reference/workspaces.md
+++ b/src/doc/src/reference/workspaces.md
@@ -49,7 +49,6 @@ where the workspace's `Cargo.toml` is located.
 
 ```toml
 [workspace]
-resolver = "2"
 
 [package]
 name = "hello_world" # the name of the package
--- a/src/doc/src/reference/workspaces.md
+++ b/src/doc/src/reference/workspaces.md
@@ -39,9 +39,6 @@ To create a workspace, you add the `[workspace]` table to a `Cargo.toml`:
 
 At minimum, a workspace has to have a member, either with a root package or as
 a virtual manifest.
-It's also a good practice to specify the `resolver = "2"` unless it is necessary
-to rely on the old one. It is a default resolver for all packages in the `2021` edition
-but it has to be explicitely set under the workspace.
 
 #### Root package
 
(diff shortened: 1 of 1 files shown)


epage:
I feel like there was another PR that did something similar that I wanted to reference for comparing but so far I can't find it.

As for raising awareness of this, in 1.72 we'll have a warning for this (#10910)

weihanglo:
@epage are you looking for this <https://github.com/rust-lang/cargo/pull/10625>?

zvolin:
Thanks, wasn't aware of https://github.com/rust-lang/cargo/pull/10910. It's great that there will be a warning, however I'm not sure if it makes changes to the documentation unneeded. Anyone who just goes through the `Workspaces` section and tries to use the examples will now face the warning right after

epage:
#10625 is what I intended and the comments won't translate 100% because we did do some documentation updates in #11082.

src/doc/src/reference/workspaces.md
```diff
@@ -49,6 +52,7 @@ where the workspace's `Cargo.toml` is located.
 
 ```toml
 [workspace]
+resolver = "2"
```
epage, line 55: Note: for this case, it isn't needed because it will be inferred from `package.edition`

src/doc/src/reference/workspaces.md
```diff
@@ -127,6 +133,7 @@ used:
 [workspace]
 members = ["path/to/member1", "path/to/member2", "path/to/member3/*"]
 default-members = ["path/to/member2", "path/to/member3/foo"]
+resolver = "2"
```
epage, line 136: For me, adding `resolver = "2"` to all examples clutters them ,distracting from what we are trying to teach

src/doc/src/reference/workspaces.md
```diff
@@ -39,6 +39,9 @@ To create a workspace, you add the `[workspace]` table to a `Cargo.toml`:
 
 At minimum, a workspace has to have a member, either with a root package or as
 a virtual manifest.
+It's also a good practice to specify the `resolver = "2"` unless it is necessary
+to rely on the old one. It is a default resolver for all packages in the `2021` edition
+but it has to be explicitely set under the workspace.
```
epage, line 44: > I'm not sure if it makes changes to the documentation unneeded. Anyone who just goes through the Workspaces section and tries to use the examples will now face the warning right after

I agree that we should have some workspace reocommendations.  Its been something we've been talking about.  [This zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Version.20Requirements.20documentation) talks some about what we might want to cover and how to include it but that is much bigger than this change and I don't think you need to tackle it.

For how we are presenting this specific case, I feel like we are doing so without context.  Personally, I would add a "workspace recommendations" section at the end and have a link to that.

src/doc/src/reference/workspaces.md
```diff
@@ -39,6 +39,9 @@ To create a workspace, you add the `[workspace]` table to a `Cargo.toml`:
 
 At minimum, a workspace has to have a member, either with a root package or as
 a virtual manifest.
+It's also a good practice to specify the `resolver = "2"` unless it is necessary
```
epage, line 42: This should link out to the documentation for the `resolver` field (see above)

src/doc/src/reference/workspaces.md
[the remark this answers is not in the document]
zvolin: fixed the explanation

src/doc/src/reference/workspaces.md
zvolin: done

src/doc/src/reference/workspaces.md
zvolin: removed from all the examples but virtual workspace

src/doc/src/reference/workspaces.md
zvolin: noted :+1:  I actually mostly used the virtual ones

src/doc/src/reference/workspaces.md
zvolin: > For how we are presenting this specific case, I feel like we are doing so without context. Personally, I would add a "workspace recommendations" section at the end and have a link to that.

As it looks that this is specific to virtual workspaces, I moved the example and explanation under this section, and imo it looks good there after also applying your suggestions from the other comment. The "workspace recommendations" section would be great, but I think it should cover more than just this and then probably be separate PR

src/doc/src/reference/workspaces.md
epage: This entry with a non-virtual workspace seems to still be here

src/doc/src/reference/workspaces.md
epage: I agree, the new version works well enough.  Once we resolve the extra entry I noted, this should be good to go

src/doc/src/reference/workspaces.md
zvolin: sorry, no idea how I missed that, removed

epage:
Thanks!

PR #12200, opened 2023-05-29
[beta-1.71.0] detect the channel a PR wants to merge into

weihanglo:
### What does this PR try to resolve?

This detects which channel a pull request wants to merge into.

It is hard because bors runs in a different repo.
Bors knows nothing about branches or tag in this repo.
It only see one base commit and a merge commit AFAIK.

Basically the assumption and idea are

1. bors create a merge commit, so `HEAD~` should find the base branch.
2. Cargo maintains `rust-1.y.0` branch for each Rust minor version releases.
   Therefore, we can use the symbolic name of `origin/rust-1.x.0`
   to determine if it aims for stable, beta, or nightly channel.
3. After we know which channel it is targeted at,
   we can skip jobs that are likely to be broken.

### How should we test and review this PR?

This is like #12181 but targeted on 1.71.0.

r? ehuss

PR #12089, opened 2023-05-05
feat(xtask): 'changes' xtask

epage:
This will report what commits affect each publishable package and serves as an alternative to rust-lang/cargo#12085.

CI can run this as `cargo changes HEAD~ HEAD^2` and that will capture all of the commits within the PR (this is the range I used in `committed`).

There is still work needed to integrate this with an action; this is just a rough base-line implementation.

I originally tried to use `cargo package --list` to determine what files belong to what packages but that was taking too long for some reason (it works well in `cargo release` for my crates).  We can either look into that in the future or do our own heuristics to filter out content.

PR #12395, opened 2023-07-24
ci: rewrite bump check and respect semver 

### LINKED ISSUE #1234: #1159 is back
Same cause as last time; the warning messages changed.
(Issue #1159)

### LINKED ISSUE #12347: version-bump check doesn't know a subcrate hasn't bumped on nightly if we've bumped it on beta

weihanglo:

This ports `ci/validate-version-bump.sh` and #12200 to the new xtask `bump-check`. It also adds the ability to set the correct baseline revision when checking version bump. That is, it fixes #12347.

In addition, this integrates the community tool `cargo-semver-checks`. SemVer violation can now be detected earlier.

This PR extracts the registry query part from `xtask-unpublished` and removes other pars. I don't feel like we need it in the short term.

For how it works, please the check doc comment in each function.

One concern is that this check is still a required job for bors. I believe @obi1kenobi is quite responsive and willing to help if there is something wrong. So, waiting for an upstream fix won't be a problem for cargo. Also as a good citizen in the community, we can always contribute back.
(Take it easy @obi1kenobi, don't be stressed out 🙂)


ci/validate-version-bump.sh
```diff
@@ -19,43 +16,4 @@ head_sha=$(git rev-parse "${HEAD_SHA:-HEAD}")
...
-    exit 0
-fi
-
-echo "Detected changes in these crates but no version bump found:"
-echo "$crates_need_bump"
-echo
-echo "Please bump at least one patch version for each corresponding Cargo.toml:"
-echo 'Run "cargo unpublished" to read the publish status table for details.'
-exit 1
+cargo semver-checks --workspace --baseline-rev "$base_sha"
```
epage, line 19: One important distinction between the old script and the new script is that this will only ensure we bump the version if the API has changed.  Changes in the implementation will not be caught

obi1kenobi (commented):
This looks good. Two suggestions to make it even better:
- Consider using the [cargo-semver-checks GitHub Action](https://github.com/obi1kenobi/cargo-semver-checks-action), which automatically sets up proper rustdoc caching and other nice things like that.
- If not, then consider at least caching the `target/semver-checks/cache/` directory across runs, to enable reuse of rustdoc JSON when the baseline version hasn't changed. This would probably reduce semver-checking time by about half.

ci/validate-version-bump.sh
weihanglo: That's fair. Sounds like we inevitably requires a xtask/script to detect changes in member crates. Maybe we can keep lines 23-33 and extend them a bit?

(I am thinking just checking if any `.rs` `Cargo.toml` have changed)

weihanglo:
I understood the baseline rustdoc wrong. It should be fixed on latest beta branch for nightly cargo, and latest stable branch for beta cargo. Make it as a draft temporarily.

weihanglo:
Thanks @obi1kenobi! Some questions:

* The GitHub Action doesn't seem able to set `--baseline-rev`, or am I missing something?
* Regarding manual caching, is using baseline rev as cache key sufficient and safe?

ci/validate-version-bump.sh
weihanglo: Just rewrote it in Rust. Should be a bit more friendly :)

weihanglo:
Let's see how far it goes.

@bors try

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,338 @@
...
+    base_commit: &git2::Commit<'r>,
+    head: &git2::Commit<'r>,
+) -> CargoResult<HashMap<&'ws str, &'ws Package>> {
+    let member_dirs = ["crates", "credential", "benches"];
+    let ws_members: HashMap<&str, &Package> =
+        HashMap::from_iter(ws.members().map(|m| (m.name().as_str(), m)));
+    let base_tree = base_commit.as_object().peel_to_tree()?;
+    let head_tree = head.as_object().peel_to_tree()?;
+    let diff = repo
+        .diff_tree_to_tree(Some(&base_tree), Some(&head_tree), Default::default())
```
weihanglo, line 211: I am not sure which one is better

* Diff members between `base...head`
* Diff members between `referenced_commit..head`

`base` here refers to the commit this PR want to merge into. `referenced_commit` is what `fn referenced_commit` returns.

The former one focuses on the current PR and is more time/resource efficient, while the latter one can catches something like #12401, though in the future case like #12401 shouldn't appear after this PR gets merged.

weihanglo:
I see. Thanks for the info! The release of Cargo is included in Rust's 6-weeks train model. Every crate gets into beta is considered as “released”, though they won't actually be released until hitting stable.

The model I propose here is roughly like

* PR for master branch should check against beta branch in this repo (i.e, the latest `rust-1.*.0` branch).
* A beta backport PR should check against stable branch in this repo. Note that it's not the version on crates.io, since there might also be a stable-backport needed to take care of).
* A stable backport PR should check against the version on crates.io.

I don't feel like it is a common case. It should be a low priority task as well, as `cargo-semver-checks` doesn't seem to be the bottleneck of Cargo CI pipeline. I am fine with no caches at this moment :)

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,338 @@
...
+    Ok(())
+}
+
+/// Main entry of `xtask-bump-check`.
+///
+/// Assumption: version number are incremental. We never have point release for old versions.
+fn bump_check(args: &clap::ArgMatches, config: &mut cargo::util::Config) -> CargoResult<()> {
+    let ws = args.workspace(config)?;
+    let repo = git2::Repository::open(ws.root()).unwrap();
+    let base_commit = arg_to_commit(args, &repo, "baseline-rev")?;
```
epage, line 289: Can we make this optional and fall back to [upstream/master](https://github.com/rust-lang/cargo/pull/12089/files#diff-294d5621c148475db1f52332ac5a5d808b52516e931cbf48883beb6727915ea5R93)?

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,338 @@
...
+    let head_commit = arg_to_commit(args, &repo, "head-rev")?;
```
epage, line 290: Can we make this optional and fallback to [HEAD](https://github.com/rust-lang/cargo/pull/12089/files#diff-294d5621c148475db1f52332ac5a5d808b52516e931cbf48883beb6727915ea5R139)?

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,338 @@
...
+/// Lists all changed workspace members between two commits.
+///
+/// Assumption: Paths of workspace members. See `member_dirs`.
+fn changed<'r, 'ws>(
+    ws: &'ws Workspace<'_>,
+    repo: &'r git2::Repository,
+    base_commit: &git2::Commit<'r>,
+    head: &git2::Commit<'r>,
+) -> CargoResult<HashMap<&'ws str, &'ws Package>> {
+    let member_dirs = ["crates", "credential", "benches"];
```
epage, line 205: Can we make this smarter and [filter for whats actually packaged](https://github.com/rust-lang/cargo/pull/12089/files#diff-294d5621c148475db1f52332ac5a5d808b52516e931cbf48883beb6727915ea5R161)?

obi1kenobi:
Ah yes, I forgot about the release train. The model you propose makes sense and the cargo-semver-checks portion of this PR looks good to me :+1:

If caching isn't the bottleneck, let's skip it for now. There's a way to make it work, but it can be a bit of a footgun in the general case (i.e. outside of this use case) so I've avoided adding it thus far. If needed down the line, we can figure out a way to enable it for this repo without causing cache bloat or other issues.

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,338 @@
...
+    } else {
+        Some(beta_commit)
+    };
+
+    Ok(commit)
+}
+
+/// Lists all changed workspace members between two commits.
+///
+/// Assumption: Paths of workspace members. See `member_dirs`.
```
epage, line 198: The other assumption is that a package's leaf directory is named the same as the package name which shouldn't necessarily be needed as we can check which is the most specific workspace member

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,338 @@
...
+
+    for delta in diff.deltas() {
+        let mut insert_changed = |path: &Path| {
+            if !member_dirs.iter().any(|dir| path.starts_with(dir)) {
+                return;
+            }
+            let Some(member) = path.components().nth(1) else {
+                log::trace!("skipping {path:?}, not a valid member path");
+                return;
+            };
```
epage, line 224: Can we instead search for the entry where a `strip_prefix` works and then we take the next path element after that?

That would harden this code against someone adding a `member_dirs` that isn't just one level

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,338 @@
...
+            // Exact to avoid returning all for path/git
+            match registry.query_vec(&query, QueryKind::Exact) {
+                task::Poll::Ready(res) => {
+                    break res?;
+                }
+                task::Poll::Pending => registry.block_until_ready()?,
+            }
+        };
+        let max_version = possibilities.iter().map(|s| s.version()).max();
+        if max_version >= Some(current) {
```
epage, line 275: So we are only finding versions that are `<=` but we need a bump if the version is` >=`?

Shouldn't the version req be `>=` and then we bump if a match is found?

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,338 @@
...
+            let name = member.name().as_str();
+            let Some(changed) = changed_members.get(name) else {
+                log::trace!("skpping {name}, may be removed or not published");
+                continue;
+            };
+
+            if changed.version() <= member.version() {
+                needs_bump.push(*changed);
+            }
+        }
```
epage, line 313: Could we update the variables to clarify these are all reference ones?
- `commit` -> `referenced_commit`
- `member` -> `referenced_member`

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,338 @@
...
+    let stable_commit = stable.get().peel_to_commit()?;
+    let beta_commit = beta.get().peel_to_commit()?;
+
+    let commit = if rev_id == stable_commit.id() {
+        None
+    } else if rev_id == beta_commit.id() {
+        Some(stable_commit)
+    } else {
+        Some(beta_commit)
+    };
```
epage, line 191: Somewhere in here, should we have some kind of warning or error check if the `referenced_commit`s version is greater than one version off than the current commit?

This would help with running locally and someone hasn't done a `git fetch upstream` in a while

ci/validate-version-bump.sh
```diff
@@ -20,3 +20,4 @@ echo "Base branch  is $base_sha"
 echo "Current head is $head_sha"
 
 cargo bump-check --baseline-rev "$base_sha" --head-rev "$head_sha"
+cargo semver-checks --workspace --baseline-rev "$base_sha"
```
epage, line 23: Doesn't the `cargo semver-checks` also need the `base` -> `referenced` conversion done as well or else if the current version makes a breaking change but major is already bumped from a prior commit, this will fail when it shouldn't

weihanglo:
Thanks for the review. Most of the review comments are addressed. `cargo semver-checks check-release --workspace` seems broken at this moment. Filed an issue to them already <https://github.com/obi1kenobi/cargo-semver-checks/issues/511>.

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,393 @@
...
+    let mut changed_members = HashSet::new();
+
+    for delta in diff.deltas() {
+        let mut insert_changed = |path: &Path| {
+            for (pkg_root, pkg) in ws_members.iter() {
+                if path.starts_with(pkg_root) {
+                    changed_members.insert(*pkg);
+                    break;
+                }
+            }
```
epage, line 273: Won't this flag everything as a change in `cargo`?  I would assume we should look for the most specific pkg_root

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,393 @@
...
+
+    if let Some(referenced_commit) = referenced_commit.as_ref() {
+        let mut cmd = ProcessBuilder::new("cargo");
+        cmd.arg("semver-checks")
+            .arg("--workspace")
+            .arg("--baseline-rev")
+            .arg(referenced_commit.id().to_string());
+        config.shell().status("Running", &cmd)?;
+        cmd.exec()?;
+    }
```
epage, line 383: nit: a comment explaining why we do both?

crates/xtask-bump-check/src/xtask.rs
weihanglo: Fixed by filtering out root package.

epage: Cheater :) 

The failure mode if we do any further nesting is just extra packages being flag for being bumped which is better than fewer packages being flagged.

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,418 @@
...
+    let excludes = |cmd: &mut ProcessBuilder| {
+        cmd.arg("--exclude")
+            .arg("cargo-credential-1password")
+            .arg("--exclude")
+            .arg("cargo-credential-gnome-secret")
+            .arg("--exclude")
+            .arg("cargo-credential-macos-keychain")
+            .arg("--exclude")
+            .arg("cargo-credential-wincred");
+    };
```
epage, line 163: nit: couldn't his be a `let exclude_args = [...];` and then later do `cmd.args(exclude_args)`?

When I see closures being used for code reuse, I tend to look for other ways (which I generally find simpler)

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -0,0 +1,418 @@
...
+) -> CargoResult<HashSet<&'ws Package>> {
+    let root_pkg_name = ws.current()?.name(); // `cargo` crate.
+    let ws_members: HashMap<_, &Package> = HashMap::from_iter(
+        ws.members()
+            .filter(|pkg| pkg.name() != root_pkg_name) // Only take care of sub crates here.
+            .filter(|pkg| pkg.publish() != &Some(vec![])) // filter out `publish = false`
+            .map(|pkg| {
+                let relative_pkg_root = pkg.root().strip_prefix(ws.root()).unwrap();
+                (relative_pkg_root, pkg)
+            }),
```
epage, line 331: nit: Could you add a comment why?

epage (approved):
Feel free to r= me

weihanglo:
@bors r=epage

Thanks a lot for the review and your prior pull request!

crates/xtask-bump-check/src/xtask.rs
weihanglo: Okay I ended up removing all closures. No idea how I got obsessed by closures…

PR #11912, opened 2023-03-29
test(cli): Track --help output

epage:
### What does this PR try to resolve?

This makes it easier to evaluate the usability of PRs, like #11905

This follows the pattern of `cargo add` and `cargo remove` of putting these ui tests in `cargo_<cmd>/` directories.  `init` didn't follow this pattern, so it was renamed to `cargo_init/`.  `cargo_config.rs` was going to conflict with this, it was merged in.

We can evaluate other `<cmd>.rs` files at a later point and consolidate.

### How should we test and review this PR?

The main risks are
- Are all files linked together (`main.rs` -> `<cmd>/mod.rs` -> `<cmd>/<help>.rs`
- Are all `help/mod.rs`s pointing to the right command

PR #12398, opened 2023-07-24
refactor(test): Move cargo-config into a dir

epage:
This is split out of #11912 and is prep for adding more UI tests.

Generally our UI tests are in a directory named after the full cargo command (`cargo config`).  These tend to use `snapbox`.

Here we are tests for the `cargo config` command not written by `snapbox` in a `cargo_config.rs` file.  This conflicts with adding snapbox UI tests later in a `cargo_config/` folder.  Upon looking at this file, it appears to be UI tests, so I think it would make sense to move them into the `cargo_config/` folder.  Definitely wouldn't make sense to move them into `config.rs` since that is general config testing.


weihanglo (approved):
Make sense to me. Thanks!

Issue #12235, opened 2023-06-06
User control over cargo warnings


epage:
### Problem

Today, cargo maintainers are very cautious as to what warnings are created because they cannot be turned off by the user when they are intended.  The problem is we've not had a mechanism to control it.  With #12115, we'll now have a mechanism

Turning existing warnings into lints:
- #11782
- Publish size warnings (https://github.com/rust-lang/rfcs/pull/3397#discussion_r1213330487)
- #7380
- Control the warning from #10910
- #2568
- <https://github.com/rust-lang/cargo/issues/15964>

Porting clippy lints to cargo
- #4377 (see also [clippy::cargo_common_metadata](https://rust-lang.github.io/rust-clippy/master/index.html#cargo_common_metadata))
- #5340 (see also [clippy::wildcard_dependencies](https://rust-lang.github.io/rust-clippy/master/index.html#/wildcard_dependencies?groups=cargo), [deny::wildcard](https://embarkstudios.github.io/cargo-deny/checks/bans/diags.html#wildcard))
- #13899 (see also [clippy::multiple_crate_versions](https://rust-lang.github.io/rust-clippy/master/index.html#/multiple_crate_versions?groups=cargo), [deny::duplicate](https://embarkstudios.github.io/cargo-deny/checks/bans/diags.html#duplicate))
- #15581
- #15580

Examples of new lints:
- #4413 ([deny::detected-executable](https://embarkstudios.github.io/cargo-deny/checks/bans/diags.html#detected-executable), [deny::detected-executable-script](https://embarkstudios.github.io/cargo-deny/checks/bans/diags.html#detected-executable-script))
- #4300
- #9058
- Warn on old build-directive syntax (#11461)
- Warn in more cases than #10910
- Maybe #2489
- Disable warning from path dependencies #8546
- Warn on mixed case package names (#6043)
- #7081
- #2039 (see also rust-lang/rust-clippy#3867, [deny's licenses](https://embarkstudios.github.io/cargo-deny/checks/licenses/diags.html))
- #6003
- #4611
- #5890
- #7760
- #6827
- Redundant `package.homepage` (see #13293)
- #15578
- #15579
- non-latest edition
- #14390
- warn if two dep info files within the same package point to the same source file (#7730).
- deny missing metadata when running `cargo package` https://github.com/rust-lang/cargo/issues/15398#issuecomment-2781116158

### Proposed Solution

As package and workspace level lint control, turning existing warnings into diagnostics where possible

Tasks
[...]

PR #12399, opened 2023-07-24
fix(package): Recognize and normalize `cargo.toml` 

### LINKED ISSUE #12384: `cargo publish` with `cargo.toml` on windows doesn't normalize the content
### Problem

On case insensitive filesystems (which can include Windows and Mac), cargo allows `cargo.toml` to be used.  When publishing, the case doesn't get normalize and we miss it when normalizing the content of the file (`prepare_for_publish`).

Once published
- crates.io won't be able to process the file
- the crate is unusable on systems with case-sensitive filesystems
- behavior might be off or flat out broken without normalization (e.g. workspace inheritance)

### Steps

```console
$ cargo new foo
$ cd foo
$ cargo package --list --allow-dirty
Cargo.lock
Cargo.toml
Cargo.toml.orig
src/main.rs
$ move Cargo.toml t
$ move t cargo.toml
$ cargo package --list --allow-dirty
Cargo.lock
Cargo.toml
src/main.rs
```

Note the lack of `Cargo.toml.orig`

### Possible Solution(s)

Long term, I'd prefer we error, suggesting a rename if `cargo.toml` exists

To help people transition, we should warn first.

Alternatively
- we could normalize the file name
- we could detect this and error when loading manifests

### Notes

See also
- #11962
- #45
- https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/broken.20crate.20files
- https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/.60cargo.2Etoml.60.20on.20case.20insensitive.20filesystems

### Version

_No response_

epage:

This solution is a blend of conservative and easy
- Normalizes `cargo.toml` to `Cargo.toml` on publish
  - Ensuring we publish the `prepare_for_publish` version and include `Cargo.toml.orig`
  - Avoids dealing with trying to push existing users to `Cargo.toml`
- All other cases of `Cargo.toml` are warnings
  - We could either normalize or turn this into an error in the future
  - When helping users with case previously, we've only handle the `cargo.toml` case
  - We already should have a fallback in case a manifest isn't detected
  - I didn't want to put in too much effort to make the code more complicated to handle this

As a side effect, if a Linux user has `cargo.toml` and `Cargo.toml`, we'll only put one of them in the `.crate` file.  We can extend this out to also include a warning for portability for case insensitive filesystems but I left that for after rust-lang/cargo#12235.

A PR at a time will show how the behavior changed as the source was edited

This does add a direct dependency on `unicase` to help keep case-insensitive comparisons easy / clear and to avoid riskier areas for bugs like writing an appropriate `Hash` implementation.  `unicase` is an existing transitive dependency of cargo.

### Additional information

Fixes #12384

[Discussion on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/.60cargo.2Etoml.60.20on.20case.20insensitive.20filesystems)

### DIFF
--- a/tests/testsuite/package.rs
+++ b/tests/testsuite/package.rs
@@ -3041,3 +3041,59 @@ src/main.rs
         &[],
     );
 }
+#[cargo_test]
+#[cfg(windows)] // windows is the platform that is most consistently configured for case insensitive filesystems
+fn no_manifest_found() {
+    let p = project()
+        .file("src/main.rs", r#"fn main() { println!("hello"); }"#)
+        .file("src/bar.txt", "") // should be ignored when packaging
+        .build();
+    // Workaround `project()` making a `Cargo.toml` on our behalf
+    std::fs::remove_file(p.root().join("Cargo.toml")).unwrap();
+    std::fs::write(
+        p.root().join("CARGO.TOML"),
+        r#"
+                [package]
+                name = "foo"
+                version = "0.0.1"
+                authors = []
+                exclude = ["*.txt"]
+                license = "MIT"
+                description = "foo"
+            "#,
+    )
+    .unwrap();
+
+    p.cargo("package")
+        .with_stderr(
+            "\
+[WARNING] manifest has no documentation[..]
+See [..]
+[PACKAGING] foo v0.0.1 ([CWD])
+[VERIFYING] foo v0.0.1 ([CWD])
+[COMPILING] foo v0.0.1 ([CWD][..])
+[FINISHED] dev [unoptimized + debuginfo] target(s) in [..]
+[PACKAGED] 3 files, [..] ([..] compressed)
+",
+        )
+        .run();
+    assert!(p.root().join("target/package/foo-0.0.1.crate").is_file());
+    p.cargo("package -l")
+        .with_stdout(
+            "\
+CARGO.TOML
+Cargo.lock
+src/main.rs
+",
+        )
+        .run();
+    p.cargo("package").with_stdout("").run();
+
+    let f = File::open(&p.root().join("target/package/foo-0.0.1.crate")).unwrap();
+    validate_crate_contents(
+        f,
+        "foo-0.0.1.crate",
+        &["CARGO.TOML", "Cargo.lock", "src/main.rs"],
+        &[],
+    );
+}
--- a/tests/testsuite/package.rs
+++ b/tests/testsuite/package.rs
@@ -2983,3 +2983,60 @@ src/main.rs.bak
         ],
     );
 }
+
+#[cargo_test]
+#[cfg(windows)] // windows is the platform that is most consistently configured for case insensitive filesystems
+fn normalize_case() {
+    let p = project()
+        .file("src/main.rs", r#"fn main() { println!("hello"); }"#)
+        .file("src/bar.txt", "") // should be ignored when packaging
+        .build();
+    // Workaround `project()` making a `Cargo.toml` on our behalf
+    std::fs::remove_file(p.root().join("Cargo.toml")).unwrap();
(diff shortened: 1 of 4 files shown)


weihanglo (commented):
> Another case this does not address (though slightly changes) is when a Linux system has multiple cases of Cargo.toml

Will this cause problems for projects already having both `Cargo.toml` and `cargo.toml`? Should we examine existing packages on crates.io?

Apart from that, looks pretty good!

epage:
> Will this cause problems for projects already having both `Cargo.toml` and `cargo.toml`? Should we examine existing packages on crates.io?

To lazy to examine them so I took an approach for handling it.  Thoughts?

ChrisDenton:
Hm... unicase seems overkill no? IIRC, cargo's special filenames are all ascii.

epage:
> Hm... unicase seems overkill no? IIRC, cargo's special filenames are all ascii.

I wrote this with an eye for portability hazards in the future.  Unsure if filesystems are case insensitive by ascii or by unicode but thought I'd play it safe rather than people in the future building on the this work and not taking this into account.

ChrisDenton:
Sure but I mean, pure ascii filenames are the only ones that can be compared portably.

What version of the Unicode spec does a filesystem use to do case-insensitive comparisons? Trick question! It could be using any version or its own weird case table. So the only portable thing would be to ask the filesystem, unless you stick to ascii only.

epage:
@ChrisDenton to clarify, how much of your concern is with
- `unicase::UniCase`
- `unicase` as a dependency
- any dependency for this

ChrisDenton:
Well I do prefer having fewer dependencies but that's only a minor concern.

And I'm not against using `unicase::UniCase` exactly. It'll work for `Cargo.toml`, but then so would an ascii comparison. My concern is a bit more fundamental in that using any single case algorithm is wrong. With ascii comparisons the limitation is at least clear to the reader and the results correct within its domain.

The "correct" way to do this for arbitrary names outside of ascii is to try opening with the file then comparing the canonical name of the opened file with the expected name. But I realise this might be more involved then is wanted here, especially considering cross-platform concerns.

epage:
Thanks for the clarification.

I've updated to ASCII comparisons.  I kept `unicase` because it makes life easier in a couple of ways, in particular by choosing to use a `HashMap`, it makes it easy to get a correct `Hash` implementation.  It is already a transitive dependency of cargo.

tests/testsuite/package.rs
```diff
@@ -2983,3 +2983,115 @@ src/main.rs.bak
...
+                name = "foo"
+                version = "0.0.1"
+                authors = []
+                exclude = ["*.txt"]
+                license = "MIT"
+                description = "foo"
+            "#;
+    let p = project()
+        .file("Cargo.toml", manifest)
+        .file("cargo.toml", manifest)
```
riking, line 3059: Test should put a bogus manifest into the lowercase version and assert it's not the one chosen

riking: Ah, you stated above you aren't dealing with that in this change.

tests/testsuite/package.rs
```diff
@@ -2983,3 +2983,60 @@ src/main.rs.bak
...
+        p.root().join("cargo.toml"),
+        r#"
+                [package]
+                name = "foo"
+                version = "0.0.1"
+                authors = []
+                exclude = ["*.txt"]
+                license = "MIT"
+                description = "foo"
+            "#,
```
weihanglo, line 3006: nit: One less indentation?

weihanglo (approved):
Looks good to me, and thanks everybody here for helping review :)

PR #12407, opened 2023-07-27
Clarify `lto` setting passing `-Clinker-plugin-lto`

keith:
Previously this document implied that the `lto` setting only controlled `-C lto` but in some cases it can also pass `-C linker-plugin-lto`

https://github.com/rust-lang/cargo/blob/1b3eb297f946779b1ede257074afff62d854ff79/src/cargo/core/compiler/mod.rs#L1234

### DIFF
--- a/src/doc/src/reference/profiles.md
+++ b/src/doc/src/reference/profiles.md
@@ -155,9 +155,10 @@ The valid options are:
 
 #### lto
 
-The `lto` setting controls the [`-C lto` flag] which controls LLVM's [link
-time optimizations]. LTO can produce better optimized code, using
-whole-program analysis, at the cost of longer linking time.
+The `lto` setting controls `rustc`'s [`-C lto`], [`-C linker-plugin-lto`], and
+[`-C embed-bitcode`] options, which control LLVM's [link time optimizations].
+LTO can produce better optimized code, using whole-program analysis, at the cost
+of longer linking time.
 
 The valid options are:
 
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/profiles.md
```diff
@@ -171,7 +171,8 @@ The valid options are:
   similar to "fat".
 * `"off"`: Disables LTO.
 
-See also the [`-C linker-plugin-lto`] `rustc` flag for cross-language LTO.
+When enabled this setting can also potentially pass the [`-C linker-plugin-lto`]
+`rustc` flag for cross-language LTO.
```
ehuss, line 175: This sentence is more trying to provide a breadcrumb to the user for more information on *how* to do cross-language LTO. What do you think about something like the following:

```suggestion
See the [linker-plugin-lto chapter] if you are interested in cross-language LTO. This is not yet supported natively in Cargo, but can be performed via RUSTFLAGS.

[linker-plugin-lto chapter]: ../../rustc/linker-plugin-lto.html
```

And also update the first sentence of this section to something like:

> The `lto` setting controls rustc's [`-C lto`], [`-C linker-plugin-lto`], and [`-C embed-bitcode`] options, which controls LLVM's ...

(with links to each of those options).

ehuss (commented):
Thanks for the PR! I can see how this seems a little confusing.

src/doc/src/reference/profiles.md
keith: Thanks for the edits, updated!

ehuss:
Thank you!

@bors r+

PR #12408, opened 2023-07-28
doc: add missing reference to `CARGO_PROFILE_<name>_STRIP` in environment variables docs

poliorcetics:
The doc for the env var was added in b6d1979683 but was forgotten in env vars


weihanglo (approved):
Nice catch. Thank you!

PR #12411, opened 2023-07-29
fix: normalize relative git submodule urls with `ssh://`

### LINKED ISSUE #12404: Cargo does not normalize relative urls for submodules in ssh git dependencies
### Problem

Github does not like relative repository names if you use ssh to download repositories.
I.e. according to github the following is _not_ a valid url to fetch a repo: `ssh://git@github.com/owner/repo/../sub`
Cargo does not normalize such relative paths for submodules and therefore fails to fetch the submodule if using ssh.

### Steps

If you have a crate that depends on a git repo via ssh. E.g.
> Cargo.toml
> ```toml
> crate = { git = "ssh://git@github.com/owner/repo" }
> ```

And that repo has a submodule specified with a relative path. E.g.
> .gitmodules
> ```
> [submodule "sub"]
>     path = sub
>     url = ../sub
> ```

Cargo is not able to fetch the submodule because it doesn't normalize the path; github doesn't like that.
Cargo tries the following:
```
Updating git repository `ssh://git@github.com/owner/repo`
Updating git submodule `ssh://git@github.com/owner/repo/../sub`
```

Github then throws an error because `owner/repo/../sub` is not a valid repository name.

Btw, if you enter this manually in the commandline, i.e. `git clone ssh://git@github.com/owner/repo/../sub` it doesn't work either, but it does work with `http`, seems to be a github server-side thing, although I don't know if this is indended or not. Git itself seems to do this normalization so relative submodules using ssh work fine there.

### Possible Solution(s)

I guess normalize the repo path if using ssh and hope nothing breaks 😅

### Notes

_No response_

### Version

```text
cargo 1.73.0-nightly (7ac9416d8 2023-07-24)
release: 1.73.0-nightly
commit-hash: 7ac9416d82cd4fc5e707c9ec3574d22dff6466e5
commit-date: 2023-07-24
host: x86_64-unknown-linux-musl
os: Alpine Linux 3.17.4 [64-bit]
```

weihanglo:

Git only assumes a URL is a relative path if it starts with `./` or `../`.
See <https://git-scm.com/docs/git-submodule>. To fetch the current repo,
we need to construct the complete submodule URL.

At this moment it comes with some limitations:

* GitHub doesn't accept non-normalized URLs wth relative paths.
  (`ssh://git@github.com/rust-lang/cargo.git/relative/..` is invalid)
* `url` crate cannot parse SCP-like URLs.
  (`git@github.com:rust-lang/cargo.git` is not a valid WHATWG URL)

To overcome these, this patch always tries `url` first to normalize the
path. If it couldn't, append the relative path as the last resort and
pray the remote git service supports non-normalized URLs.

I did testing by hand. Not sure how to construct a proper test around it.

cc @ehuss if you can help with this. I am a bit lost.

## Additional information

See rust-lang/cargo#12404 and rust-lang/cargo#12295.

Fixes #12404.

I'd like to nominate this as a beta backport (again 😂).

### DIFF
--- a/src/cargo/sources/git/utils.rs
+++ b/src/cargo/sources/git/utils.rs
@@ -440,20 +440,7 @@ impl<'a> GitCheckout<'a> {
                 return Ok(());
             }
 
-            // Git only assumes a URL is a relative path if it starts with `./` or `../`.
-            // See [`git submodule add`] documentation.
-            //
-            // [`git submodule add`]: https://git-scm.com/docs/git-submodule
-            let child_remote_url = if ["./", "../"].iter().any(|p| child_url_str.starts_with(p)) {
-                let mut new_remote_url = parent_remote_url.to_string();
-                if !new_remote_url.ends_with('/') {
-                    new_remote_url.push('/');
-                }
-                new_remote_url.push_str(child_url_str);
-                Cow::from(new_remote_url)
-            } else {
-                Cow::from(child_url_str)
-            };
+            let child_remote_url = absolute_submodule_url(parent_remote_url, child_url_str)?;
 
             // A submodule which is listed in .gitmodules but not actually
             // checked out will not have a head id, so we should ignore it.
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks for taking a look at this! Can you add a unittest to validate the behavior of the path joining code? That is, move all the code that computes `child_remote_url` into a free function, and add a unittest that has a few asserts for the various permutations that would be possible? Like, what if it starts with `./` or `../` or does or does not end with `/` or contains multiple `..` or is a relative path, or if the child is a full url, etc. That should also help keep the `update_submodule` function a little shorter to help make it a little easier to follow.

weihanglo:
Updated. Thank you for the advice. I almost forgot we have unit tests 🤣.

ehuss:
Thanks!

I pushed a change to fix some typos in the comments.

@bors r+

PR #12416, opened 2023-07-30
test: relax help text assertion

weihanglo:

When running test in rust-lang/rust via `./x test src/tool/cargo`, there is no rustup integrated.

Update the submodule and run `./x test src/tool/cargo` in rust-lang/rust repo.


ehuss:
Thanks!

@bors r+

PR #12418, opened 2023-07-31
fix: align `cargo --help` text

weihanglo:
Too caseless of me. #12416 didn't fix them all.

The two help texts should be aligned after this PR.

To test this, change the value of `is_rustup` and run `cargo t --test testsuite cargo::help`.


ehuss:
😄 

@bors r+

PR #12420, opened 2023-07-31
Display crate version on timings graph

### LINKED ISSUE #7384: --timings: show crate version in the graph?
When different dependencies pull in multiple versions of the same crate, it would help to see the version of the crate that was compiled. (See syn when building rustfmt
syn
)

Angelin01:

Fixes #7384

The solution here is the simplest one: since duplicate crates with different versions can cause confusion, we always output the version.

Build any create using the `--timings` option:
```
cargo build --timings
```
and verify the resulting HTML file.

### Additional information

The formatting is consistent with how the crates are displayed in the terminal output for Cargo and also on the table below the graph. Initially, [this comment](https://github.com/rust-lang/cargo/issues/7384#issuecomment-1657102900) suggested said formatting for ease of searching, but I do believe the browser can't search in the graph itself.

Sample console output:
```
   Compiling lazycell v1.3.0
   Compiling unicode-xid v0.2.4
   Compiling unicode-width v0.1.10
   Compiling glob v0.3.1
   Compiling curl-sys v0.4.65+curl-8.2.1
   Compiling curl v0.4.44
   Compiling git2 v0.17.2
```

Sample rendered HTML output:
image

### Possible issues and future work

A possible issue in this solution is that the output becomes too noisy. Other possible solutions:
- Only display the version if there are two "units" with the same name. One possible implementation is to count the names, aggregate them in a simple map and look it up during rendering.
- Create a small selection to select the disambiguate method between "Never", "Always" and "Only duplicates". This may be overkill.

### DIFF
--- a/src/cargo/core/compiler/timings.js
+++ b/src/cargo/core/compiler/timings.js
@@ -75,6 +75,8 @@ function render_pipeline_graph() {
   ctx.translate(X_LINE, MARGIN);
 
   // Compute x,y coordinate of each block.
+  // We also populate a map with the count of each unit name to disambiguate if necessary
+  const unitCount = new Map();
   UNIT_COORDS = {};
   for (i=0; i<units.length; i++) {
     let unit = units[i];
@@ -86,6 +88,9 @@ function render_pipeline_graph() {
     }
     let width = Math.max(px_per_sec * unit.duration, 1.0);
     UNIT_COORDS[unit.i] = {x, y, width, rmeta_x};
+
+    const count = unitCount.get(unit.name) || 0;
+    unitCount.set(unit.name, count + 1);
   }
 
   // Draw the blocks.
@@ -111,7 +116,10 @@ function render_pipeline_graph() {
     ctx.textAlign = 'start';
     ctx.textBaseline = 'middle';
     ctx.font = '14px sans-serif';
-    const label = `${unit.name}${unit.target} ${unit.duration}s`;
+
+    const labelName = (unitCount.get(unit.name) || 0) > 1 ? `${unit.name} (v${unit.version})${unit.target}` : `${unit.name}${unit.target}`;
+    const label = `${labelName}: ${unit.duration}s`;
+
     const text_info = ctx.measureText(label);
     const label_x = Math.min(x + 5.0, canvas_width - text_info.width - X_LINE);
     ctx.fillText(label, label_x, y + BOX_HEIGHT / 2);


src/cargo/core/compiler/timings.js
```diff
@@ -111,7 +111,7 @@ function render_pipeline_graph() {
     ctx.textAlign = 'start';
     ctx.textBaseline = 'middle';
     ctx.font = '14px sans-serif';
-    const label = `${unit.name}${unit.target} ${unit.duration}s`;
+    const label = `${unit.name} v${unit.version}${unit.target} ${unit.duration}s`;
```
epage, line 114: > > image

For me, my biggest concern is how easily can I digest the timing numbers.  I feel that rendering like `quote v1.0.28 0.45s` makes this difficult

Ideas
- @weihanglo mentioned only showing versions on duplicates
- In #7384, parenthesis were used, adding noise but also making it easier to scan
- Tooltips might be a possibility

Any other ideas?

Something else I want to keep in mind is that some improvement is better than no improvement, so if we can't converge on a solution quickly enough, we probably should just move forward.

Angelin01: I have no strong opinion, much like @weihanglo. I opened the PR with the simple one line solution exactly for the same reason: displaying the version always is better than not showing at all.

I can implement "showing versions on duplicates" or change the format relatively easily, but tooltips on rendered graphs might be beyond my JS abilities.

epage: I can definitely understanding of limited JS abilities; I'm in a similar boat :).  As I said, some improvement is better than none so I didn't want to put too much of a burden on this PR.

Just checked in with @weihanglo and if you are ok with limiting this to dupes, let's go with that.  We can iterate on this further later.

weihanglo: If you are open to make versions show only on duplicates, I personally more lean towards that 😆.

weihanglo: > I do believe the browser can't search in the graph itself.

Oops. My bad. I forgot it is an SVG.

epage: Thanks for trying it out and sharing the results!

I am curious why `syn build script (run)` doesn't have a version because that should be a dupe.

What do you think of adding separator between `labelName` and `duration` so we get a little bit more visual distinction?  Maybe `:` or `-`?

> But I'm still getting some duplicate behavior, apparently we are compiling the same crate twice? 

I suspect this is because we compile for "host" for build scripts and "target" for everything else.  We try to de-duplicate these build steps.  Unsure what is going on that this de-duplication isn't working.  I wonder if its related to the debug info shrinking or something else.

Angelin01: > I am curious why syn build script (run) doesn't have a version because that should be a dupe.
Because the target is different. I explicitly used name + target for uniqueness, but that's easy enough to change.

> What do you think of adding separator between labelName and duration so we get a little bit more visual distinction? Maybe : or -?
I like `:`.

@epage @weihanglo
I'm willing to implement any chosen format as long as we decide which one. I personally still lean towards always showing the versions as I find it less confusing (someone looking at this might ask: "why is it showing the version for X and Y but not Z?") but that's just me.

If consistent formatting with the rest of the output is not necessary here, may I suggest one of these formats:
1. `serde_json v1.2.3 syn build script (run): 0.21s`
2. `serde_json (v1.2.3) syn build script (run): 0.21s`
3. `serde_json v1.2.3 (syn build script (run)): 0.21s`
4. `serde_json v1.2.3 syn build script (run) - 0.21s`
5. `serde_json (v1.2.3) syn build script (run) - 0.21s`
6. `serde_json (v1.2.3) (syn build script (run)) - 0.21s`

Bear in mind target can be empty string and, since we showing versions only on duplicate entries, so can it.

weihanglo: >  Unsure what is going on that this de-duplication isn't working. I wonder if its related to the debug info shrinking or something else.

Could be. I think we can move on and deal with it afterward.

Regarding the format, I'll leave the choice to you :)

BTW, could you help updates screenshots in that pages? Some of them could be in a better resolution.

weihanglo (approved):
Thanks! Let's merge it :)

weihanglo:
Thought you've changed the format but apparently not. 

Will wait for you replying back :)

@bors r-

Angelin01:
@weihanglo I've adjusted the format to number 2, only showing the version when there are duplicate crates and taking epage's suggestion to add a `:` separator and put the version between `()`. Here's a report so you can have a look yourself if you want.

[cargo-timing-20230801T120113Z.tar.gz](https://github.com/rust-lang/cargo/files/12228892/cargo-timing-20230801T120113Z.tar.gz)

> BTW, could you help updates screenshots in that pages? Some of them could be in a better resolution.

I'm sorry, I don't know which screenshots you mean, but I'm happy to help if you point me in the correct direction.

weihanglo:
Was talking about <https://doc.rust-lang.org/nightly/cargo/reference/timings.html>, though it is not necessary for this pull request. Anyway, thanks for the contribution!

@bors r+

weihanglo:
> But I'm still getting some duplicate behavior, apparently we are compiling the same crate twice? Or it's showing on the graph twice and I'm not sure why.

Anything different in the `Unit` can contribute to the duplicates. For examples, build script dependencies have different profile settings than normal runtime dependencies. And different "CompileMode" would also make it looks like a duplicate.

<https://github.com/rust-lang/cargo/blob/c91a693e7977e33a1064b63a5daf5fb689f01651/src/cargo/core/compiler/unit.rs#L37-L85>

epage:
Thanks for working with us on this!

Angelin01:
No problem! This was quite an easy process. Thanks for the reviews and iteration on the PR! I'll contribute again when I can.

Issue #6669, opened 2019-02-15
`cargo test --all-targets` does not run doc tests


thomaseizinger:
As per title `cargo test --all-targets` does not run the doc tests on a crate. This is very surprising behaviour. I found [this](https://github.com/rust-lang/cargo/pull/4592#issuecomment-338409476) and it looks like running doc tests is done by `rustdoc` which is apparently not invoked when doing `--all-targets`.

dwijnand:
An attempt to fix this was made in #6671, but there's a bit of a mismatch between `cargo test` and doc tests that makes the issue complicated, so the PR didn't land.

thomaseizinger:
Maybe it is worth to add some context to this issue.

We ran into this because we use `cargo test --all-targets` on Travis and usually `cargo test` locally.
A not-compiling doc test sneaked into a last-minute update of a PR but the build was fine since we use `--all-targets` on Travis.
Later on, someone complained that we have a broken `master` because of a non-compiling doc test.

I read some of the discussion in the PR and TBH, the whole `--doc` is a mode would be SUPER confusing.
As a user, I expect `cargo test` to test LESS than `cargo test --all-targets` since the latter has `all` in the name.
In addition, the documentation (`--help`) lists doc tests next to all the other targets.

glts:
This means the command to run absolutely all tests everywhere is `cargo test --workspace`, not `cargo test --workspace --all-targets`, right?

ehuss:
> This means the command to run absolutely all tests everywhere is `cargo test --workspace`, not `cargo test --workspace --all-targets`, right?

It depends on the structure of your packages, but in general there is currently no single command to "test everything".   `cargo test` is intended to be the usual command.  If you also want to test benchmarks and examples, you can either mark them with `test=true` in `Cargo.toml`, or run two commands (`cargo test --all-targets` and `cargo test --doc`).

Veetaha:
Oh god, I think we've been not running doc tests on our CI for a couple of years due to this...

ehuss:
[...]

PR #12422, opened 2023-07-31
Clarify in `--help` that `cargo test --all-targets` excludes doctests

obi1kenobi:
Per my proposal here: https://github.com/rust-lang/cargo/issues/6669#issuecomment-1658593167

I tried to keep the edit minimalistic to match the surrounding style.

If the maintainers are amenable to it, I think it could also be useful to do one or more of:
- Offer concrete guidance on what to do to run actually-all tests (`--all-targets` then separately `--doc`).
- Link to the issue at: https://github.com/rust-lang/cargo/issues/6669
- Mention that `cargo test` without `--all-targets` runs doctests by default, which seems not immediately obvious.

I'd be happy to attempt to add any of the above that the maintainers feel would be a good fit here.

### DIFF
--- a/src/bin/cargo/commands/test.rs
+++ b/src/bin/cargo/commands/test.rs
@@ -34,7 +34,7 @@ pub fn cli() -> Command {
             "Test all tests",
             "Test only the specified bench target",
             "Test all benches",
-            "Test all targets",
+            "Test all targets (does not include doctests)",
         )
         .arg(flag("doc", "Test only this library's documentation"))
         .arg(flag("no-run", "Compile, but don't run tests"))
--- a/tests/testsuite/cargo_test/help/stdout.log
+++ b/tests/testsuite/cargo_test/help/stdout.log
@@ -17,7 +17,7 @@ Options:
       --test [<NAME>]           Test only the specified test target
       --benches                 Test all benches
       --bench [<NAME>]          Test only the specified bench target
-      --all-targets             Test all targets
+      --all-targets             Test all targets (does not include doctests)
       --doc                     Test only this library's documentation
       --no-run                  Compile, but don't run tests
       --no-fail-fast            Run all tests regardless of failure


ehuss:
I left a comment over at https://github.com/rust-lang/cargo/issues/6669#issuecomment-1658829220. This isn't a review on this PR (I'll leave that up to epage), but just a passing comment that saying "except doctests" might imply that a doctest is a target (which it isn't). I would probably word this like `(does not include doctests)` or something like that, but I am splitting hairs and the exact wording isn't important to me, just a suggestion if you like it better.

I do agree it is helpful to clarify, so thank you for opening this!

epage:
> If the maintainers are amenable to it, I think it could also be useful to do one or more of:

I think at least clarifying the webpage / man page / `cargo help test` in the areas [I highlighted](https://github.com/rust-lang/cargo/issues/6669#issuecomment-1658861048) would be good.  We could fix the current miscommunication and handle this at the same time.

As for the ``-h` / --help`, we generally keep that brief (and I think `cargo add` is the only command with separate `-h` and `--help` on top of the man page)

src/bin/cargo/commands/test.rs
```diff
@@ -34,7 +34,7 @@ pub fn cli() -> Command {
             "Test all tests",
             "Test only the specified bench target",
             "Test all benches",
-            "Test all targets",
+            "Test all targets except doctests",
```
epage, line 37: > >  I would probably word this like (does not include doctests) or something like that, but I am splitting hairs and the exact wording isn't important to me, just a suggestion if you like it better.

I like the suggestion for it being paranthetical

I think I also prefer wording besides "except" makes it sound like we are actively excluding it
- `does not include` is ok
- `precludes` might also work

obi1kenobi: I can switch to `does not include` with a parenthetical.

Though, one thing that phrasing doesn't address or even hint at is that `cargo test` without this flag _does_ run doctests. From that perspective, adding this flag does have the effect of actively excluding doctests, even though I understand the difference in intent there.

obi1kenobi:
> I think at least clarifying the webpage / man page / cargo help test in the areas https://github.com/rust-lang/cargo/issues/6669#issuecomment-1658861048 would be good. We could fix the current miscommunication and handle this at the same time.

I don't feel like I understand the finer points of targets well enough to tackle the "docs text implies doctests are a target" issue in the links you shared, but I did open another PR to add a clarification for the `--all-targets` flag on that page: https://github.com/rust-lang/cargo/pull/12423

obi1kenobi:
I'm running into some linking-related problem where even though I have `pkg-config` and OpenSSL installed as recommended in the README, cargo is still failing to build.

I tried to edit the snapshot test by hand so as not to block this PR on the issues with my local setup.

weihanglo:
> I'm running into some linking-related problem where even though I have `pkg-config` and OpenSSL installed as recommended in the README, cargo is still failing to build.

Off topic but I am interested in this. If you are okay with it please share the detail on zulip or in a new issue :)

epage:
Thanks!

@bors r+

PR #12423, opened 2023-07-31
Clarify in documentation that `--all-targets` does not run doctests

obi1kenobi:
Based on discussion in https://github.com/rust-lang/cargo/issues/6669#issuecomment-1658593167

### DIFF
--- a/src/doc/src/commands/cargo-test.md
+++ b/src/doc/src/commands/cargo-test.md
@@ -241,7 +241,8 @@ manifest settings for the target.</dd>
 
 
 <dt class="option-term" id="option-cargo-test---all-targets"><a class="option-anchor" href="#option-cargo-test---all-targets"></a><code>--all-targets</code></dt>
-<dd class="option-desc">Test all targets. This is equivalent to specifying <code>--lib --bins --tests --benches --examples</code>.</dd>
+<dd class="option-desc">Test all targets. This is equivalent to specifying <code>--lib --bins --tests --benches --examples</code>.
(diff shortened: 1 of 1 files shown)


obi1kenobi:
ah I messed this up, sorry -- don't bother reviewing yet 😅

obi1kenobi:
I'm actually a bit stuck here. The `--all-targets` doc block is shared between a number of different pages, and as far as I understand the caveat about doctests only really applies to `cargo test` and not, for example, `cargo bench`.

As one possibility, I could move the `--all-targets` doc to a separate "include" markdown file `options-all-targets.md`, then write a custom `--all-targets` doc block for `cargo test`. This would accomplish the goal, but at the cost of future edits to `options-all-targets.md` not being reflected in the `cargo test` page's entry for that flag.

What do you think @epage? Perhaps there's a better way?

epage:
Bleh, includes are difficult

Options
- Don't both with this part of the change
- Duplicate the whole options-targets directly in `cargo test`
- Add a variable to customize the behavior of options-targets

Not thrilled with any of the options

obi1kenobi:
Agreed, don't love any of them.

If you have a preference, I'm happy to do that. If not, I'll default to closing this and leaving a note about it in #6669 so it doesn't slip through the cracks and is readily found once the testing sub-team is up and running.

epage:
For me, the more important part is everything but what is in the shared documentation.  Since you aren't feeling up for that, I lean towards closing.

PR #12424, opened 2023-07-31
Use thiserror for credential provider errors

arlosi:
Errors from credential providers currently must a single string. This leads to a lot of `.map_err(|e|cargo_credential::Error::Other(e.to_string())`, which loses the `source()` of these errors.

This changes the `cargo_credential::Error` to use `thiserror` and adds a custom serialization for `std::error::Error` that preserves the source error chain across serialization / deserialization.

A unit test is added to verify serialization / deserialization.

### DIFF
--- a/credential/cargo-credential/Cargo.toml
+++ b/credential/cargo-credential/Cargo.toml
@@ -7,6 +7,8 @@ repository = "https://github.com/rust-lang/cargo"
 description = "A library to assist writing Cargo credential helpers."
 
 [dependencies]
+anyhow.workspace = true
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
+thiserror.workspace = true
 time.workspace = true
--- a/credential/cargo-credential/src/error.rs
+++ b/credential/cargo-credential/src/error.rs
@@ -1,0 +1,193 @@
+use serde::{Deserialize, Serialize};
+use std::error::Error as StdError;
+use thiserror::Error as ThisError;
+
+/// Credential provider error type.
+///
+/// `UrlNotSupported` and `NotFound` errors both cause Cargo
+/// to attempt another provider, if one is available. The other
+/// variants are fatal.
+///
+/// Note: Do not add a tuple variant, as it cannot be serialized.
+#[derive(Serialize, Deserialize, ThisError, Debug)]
+#[serde(rename_all = "kebab-case", tag = "kind")]
+#[non_exhaustive]
+pub enum Error {
+    #[error("registry not supported")]
+    UrlNotSupported,
+    #[error("credential not found")]
+    NotFound,
+    #[error("requested operation not supported")]
+    OperationNotSupported,
+    #[error("protocol version {version} not supported")]
+    ProtocolNotSupported { version: u32 },
+    #[error(transparent)]
+    #[serde(with = "error_serialize")]
+    Other(Box<dyn StdError + Sync + Send>),
+}
+
+impl From<std::io::Error> for Error {
+    fn from(err: std::io::Error) -> Self {
+        Box::new(err).into()
+    }
+}
+
+impl From<serde_json::Error> for Error {
+    fn from(err: serde_json::Error) -> Self {
+        Box::new(err).into()
+    }
+}
+
+impl From<String> for Error {
+    fn from(err: String) -> Self {
+        Box::new(StringTypedError {
+            message: err.to_string(),
+            source: None,
+        })
+        .into()
+    }
+}
+
+impl From<&str> for Error {
+    fn from(err: &str) -> Self {
+        err.to_string().into()
+    }
+}
+
+impl From<anyhow::Error> for Error {
+    fn from(value: anyhow::Error) -> Self {
+        let mut prev = None;
+        for e in value.chain().rev() {
+            prev = Some(Box::new(StringTypedError {
+                message: e.to_string(),
+                source: prev,
+            }));
+        }
+        Error::Other(prev.unwrap())
+    }
+}
+
+impl<T: StdError + Send + Sync + 'static> From<Box<T>> for Error {
+    fn from(value: Box<T>) -> Self {
+        Error::Other(value)
+    }
+}
+
+/// String-based error type with an optional source
+#[derive(Debug)]
+struct StringTypedError {
+    message: String,
+    source: Option<Box<StringTypedError>>,
+}
+
+impl StdError for StringTypedError {
+    fn source(&self) -> Option<&(dyn StdError + 'static)> {
+        self.source.as_ref().map(|err| err as &dyn StdError)
+    }
+}
+
+impl std::fmt::Display for StringTypedError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.message.fmt(f)
+    }
+}
+
+/// Serializer / deserializer for any boxed error.
+/// The string representation of the error, and its `source` chain can roundtrip across
+/// the serialization. The actual types are lost (downcast will not work).
+mod error_serialize {
+    use std::error::Error as StdError;
+    use std::ops::Deref;
+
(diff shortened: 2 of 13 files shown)


credential/cargo-credential/Cargo.toml
```diff
@@ -7,6 +7,8 @@ repository = "https://github.com/rust-lang/cargo"
 description = "A library to assist writing Cargo credential helpers."
 
 [dependencies]
+anyhow.workspace = true
```
arlosi, line 10: `anyhow` could be made an optional dependency with a feature to enable it. It's here to enable a convenience conversion.

credential/cargo-credential/src/error.rs
```diff
@@ -0,0 +1,193 @@
...
+/// 
+/// `UrlNotSupported` and `NotFound` errors both cause Cargo
+/// to attempt another provider, if one is available. The other 
+/// variants are fatal.
+/// 
+/// Note: Do not add a tuple variant, as it cannot be serialized.
+#[derive(Serialize, Deserialize, ThisError, Debug)]
+#[serde(rename_all = "kebab-case", tag = "kind")]
+#[non_exhaustive]
+pub enum Error {
```
epage, line 15: For Deserialization, do we need to worry about older credential providers having more `Error` variants?

arlosi: I think we're OK making a breaking change by removing variants here since it's unstable and it was only added last week.

credential/cargo-credential/src/error.rs
```diff
@@ -0,0 +1,193 @@
...
+    UrlNotSupported,
+    #[error("credential not found")]
+    NotFound,
+    #[error("requested operation not supported")]
+    OperationNotSupported,
+    #[error("protocol version {version} not supported")]
+    ProtocolNotSupported { version: u32 },
+    #[error(transparent)]
+    #[serde(with = "error_serialize")]
+    Other(Box<dyn StdError + Sync + Send>),
```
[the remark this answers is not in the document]
epage: And whether messages and programmatic variants are actually mutually exclusive or not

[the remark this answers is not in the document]
epage: Oh, most of this moved so I guess this is out of scope for this PR though something for us to keep in mind

credential/cargo-credential/src/lib.rs
```diff
@@ -63,7 +65,7 @@ pub struct RegistryInfo<'a> {
     /// The crates.io registry will be `crates-io` (`CRATES_IO_REGISTRY`).
     pub name: Option<&'a str>,
     /// Headers from attempting to access a registry that resulted in a HTTP 401.
-    #[serde(skip_serializing_if = "Vec::is_empty")]
+    #[serde(skip_serializing_if = "Vec::is_empty", default)]
```
epage, line 68: Can we at least split this unrelated fix into a separate commit?

credential/cargo-credential/Cargo.toml
epage: Not the biggest fan of things like `anyhow` being in library APIs **but** the place we care most about compatibility is the wire protocol and `anyhow` doesn't matter there so ... meh.

credential/cargo-credential/src/error.rs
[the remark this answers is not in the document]
arlosi: I wanted to have variants where Cargo may want to do something differently based on the provider output. Everything else should go into the `Other` variant.

`UrlNotSupported` and `NotFound` both cause Cargo to try the next provider in the global credential provider list rather than erroring out.

I did reduce the number of variants in this change, but we could possibly reduce it further. I kept `ProtocolNotSupported` and `OperationNotSupported` in case we wanted to handle those differently in the future.

credential/cargo-credential/src/error.rs
```diff
@@ -0,0 +1,193 @@
+use serde::{Deserialize, Serialize};
+use std::error::Error as StdError;
+use thiserror::Error as ThisError;
+
+/// Credential provider error type.
+/// 
+/// `UrlNotSupported` and `NotFound` errors both cause Cargo
+/// to attempt another provider, if one is available. The other 
+/// variants are fatal.
```
epage, line 9: By making it easier to use, a part of me worries that we are making it harder to correctly report the right error variant.

The way I've handled this in the error codes for my CLI is to have a sloppy error type for the implementation of each function call in `main` and then have a more constrained error type with explicit conversions used by `main`.  This does require the structure of the code to align well with the error variants.

epage (commented):
Thanks!  Overall, I appreciate how much this cleans up the code

credential/cargo-credential/src/error.rs
epage: I meant for when this becomes stable, what constraints are we putting on evolution and how do we reduce them.

credential/cargo-credential/src/error.rs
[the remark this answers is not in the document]
epage: Makes sense.  The fact that we only document behavior differences with some variants is partly why I asked.

credential/cargo-credential/src/error.rs
arlosi: Thanks. I'm definitely open to feedback on improving this. 

Since it will be implemented by external crates, it's hard to know up front what all the variants should be. In the original implementation, I tried to anticipate the common ones and add variants, but in this change I decided to reduce the number of variants and have an `Other` case.

I'm not quite sure what you're suggesting here. Removing some of the `From` impls that convert to `Other` and requiring explicit conversions?

epage: > I'm not quite sure what you're suggesting here. Removing some of the From impls that convert to Other and requiring explicit conversions?

Not quite a suggestion but effectively yes.  Phrased differently: "I'm looking for input on whether we should reduce `From` impls to encourage explicit conversions so people are more likely to specify the correct error semantics".

arlosi: I made changes to remove the `From` impls for `serde::Error` and `io::Error` and added more error contexts.

credential/cargo-credential/src/error.rs
arlosi: Great point! I added a new `Unknown` variant marked with `serde(other)` to each of the enums that's serialized across the JSON protocol. That enables deserialization to succeed even if there are new variants added. Between this and marking the enums `non_exhaustive`, we're hopefully flexible enough.

credential/cargo-credential/src/error.rs
```diff
@@ -13,29 +13,32 @@ use thiserror::Error as ThisError;
...
 
-impl From<serde_json::Error> for Error {
-    fn from(err: serde_json::Error) -> Self {
-        Box::new(err).into()
-    }
+    /// A new variant was added to this enum since Cargo was built
+    #[error("unknown error kind; try updating Cargo?")]
+    #[serde(other)]
+    Unknown,
 }
```
epage: nit: was this meant to be part of the next commit?

credential/cargo-credential/src/error.rs
```diff
@@ -156,6 +159,16 @@ mod error_serialize {
...
 
+    #[test]
+    pub fn unknown_kind() {
+        let json = r#"{
+            "kind": "unexpected-kind",
+            "unexpected-content": "test"
+          }"#;
+        let e: Error = serde_json::from_str(&json).unwrap();
+        assert!(matches!(e, Error::Unknown));
+    }
```
epage, line 170: nit: was this meant to be part of the next commit?

credential/cargo-credential/src/error.rs
epage: To be clear, I wasn't saying it had to be done but was wanting to make sure we made a conscious decision.  If this is what you feel is best great.  If not, feel free to say so.

epage:
Overall, I'm fine with it as-is and leave it to arlosi if any of the remaining points are changed.

@bors r=arlosi

epage:
@bors cancel

I got the command wrong.  Feel free to r= me

Issue #12425, opened 2023-08-01
A way for users to bulk upgrade across incompatible versions


epage:
### Problem

A lot of incompatible upgrades have a small enough of breakages that a "upgrade all of my incompatible version requirements" would make it easier to go through this process.

This is currently exposed in the `cargo upgrade` command

### Proposed Solution

See https://internals.rust-lang.org/t/feedback-on-cargo-upgrade-to-prepare-it-for-merging/17101/141

epage:
Below is some background on what we are using to help come up with a design

## Care Abouts

Priorities
- Don't break behavior on `cargo update`
- Don't write out incompatible `Cargo.lock` and `Cargo.toml`
- Focus is on end-users solving common problems and not on being a general programmatic CLI that is meant to cover every case
- Be predictable and understandable
  - Can someone unfamiliar with Rust, reading a blog post, predict what different command invocations will do?
  - Preference for not having too similarly named commands
- When higher priorities allow, avoid errors that make users go "if you know exactly what I was asking for then why didn't you just do it?"; those are a sign of issues with the UX.
- Don't be hassle when dealing with intentionally held back dependencies

Primary use cases:
- Want to have simple workflow for "upgrade incompatible dependencies only".
- Want to have simple workflow for bulk lock to latest compatible (already exists as `cargo update`)
- (medium priority) Selective modify one dependency's version requirement to latest compatible, latest incompatible
- (lower priority) Want to have simple-ish workflow for bulk upgrade to latest compatible
- (lower priority) bulk upgrade all dependencies (could just be two command invocations) 

Secondary use cases are:
- Selective modify version requirement to user-provided value
- Upgrade explicitly pinned version requirements

Some open questions we had
- How do we tell when a renamed dependency like `tokio_03` is pinned or not?
  - We could just assume all renamed are pinned
  - We could add a dependency field but I'm a bit leery of adding that kind of bookkeeping to the manifest
  - We could force users to `--exclude` these dependencies but that might be a bit of a pain to always remember to do
  - We could only skip renamed if multiple dependencies exist that point to the same package

## Context

Currently, `cargo update` is focused solely on `Cargo.lock` editing
- Spans entire dependency tree
- Multiple versions of a package may exist, referenced by `name@version`
- Deals with exact versions and not version ranges
- Only affects you and not your dependents

Version requirement editing is different in that
- Workspace members only
- May want differences between members
- Supports alternative names for packages
- Affects dependents

And as a reminder of the CLI:
[...]

PR #12428, opened 2023-08-01
fix(update): Tweak CLI behavior

epage:

When looking at `cargo update` for #12425, I noticed that the two flags related to `--package` were not next to it or each other.  I figured grouping them like that would make things easier to browse.

When looking into that, I noticed that the two flags conflict and figured we'd provide a better error message if we did that through clap.

Looking per commit will help show the behavior changes.

### Additional information

I wanted to scope this to being simple, non-controversial, low effort, incremental improvements with this change so I did not look into the history of `--aggressive` not requiring  `--package` like `--precise` does and figure out if there is any consistency we can be working towards.

### DIFF
--- a/src/bin/cargo/commands/update.rs
+++ b/src/bin/cargo/commands/update.rs
@@ -9,10 +9,13 @@ pub fn cli() -> Command {
         .arg_quiet()
         .arg(flag("workspace", "Only update the workspace packages").short('w'))
         .arg_package_spec_simple("Package to update")
-        .arg(flag(
-            "aggressive",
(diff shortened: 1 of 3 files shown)


weihanglo (commented):
Should this be removed as well?

https://github.com/rust-lang/cargo/blob/c91a693e7977e33a1064b63a5daf5fb689f01651/src/cargo/ops/cargo_generate_lockfile.rs#L39-L41

BTW, it's great to see help text is tracked :)

weihanglo (approved):

weihanglo:
Make sense. Thank you!

@bors r+

PR #12429, opened 2023-08-01
`#[allow(internal_features)]` in RUSTC_BOOTSTRAP test

Nilstrieb:
This will be required in the future (where "the future" is rust-lang/rust#108955 which fails CI because of cargo here).

This does emit the unknown lints lint right now but that doesn't matter as it's just warn-by-default - internal_features is deny-by-default though, so it causes errors.

### DIFF
--- a/tests/testsuite/build_script_env.rs
+++ b/tests/testsuite/build_script_env.rs
@@ -117,7 +117,10 @@ fn rustc_bootstrap() {
     "#;
     let p = project()
         .file("Cargo.toml", &basic_manifest("has-dashes", "0.0.1"))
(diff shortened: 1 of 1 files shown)


Nilstrieb:
Hm, it doesn't actually seem to fix the issue...

Nilstrieb:
Ah, bootstrap was messing with my submodules. After messing with submodules a little more I was finally able to test it locally on my branch and tests passed.

weihanglo (approved):

weihanglo:
Tested. It works. Thank you :)

@bors r+

PR #12435, opened 2023-08-02
replace `master` branch by default branch in documentation

szabgab:


weihanglo (approved):
Thanks. I bet we can adopt this change, like what we did in <https://github.com/rust-lang/cargo/pull/12251>.

PR #12439, opened 2023-08-02
Fix printing multiple warning messages for unused fields in [registries] table

arlosi:
Cargo currently prints the same warning message multiple times for unexpected fields in the `[registries.<registry>]` or `[registry]` tables. This is because Cargo warns each time that the structure is deserialized from `Config`. Depending on which code path is taken I've seen the warning printed up to 6 times.

* A cache of deserialized registry configurations is added to the `Config` struct.
* Registry authentication is changed to directly read the config when searching for a registry name, rather than deserializing each registry configuration.

A test is added to ensure both `[registries]` and `[registry]` only warn once for unexpected fields.

### DIFF
--- a/src/cargo/util/auth/mod.rs
+++ b/src/cargo/util/auth/mod.rs
@@ -12,7 +12,6 @@ use cargo_credential::{
 
 use core::fmt;
 use serde::Deserialize;
-use std::collections::HashMap;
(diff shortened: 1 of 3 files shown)


ehuss:
Thanks!

@bors r+

PR #12444, opened 2023-08-03
Remove build metadata from libgit2-sys dependency

ehuss:
This removes the `+1.6.4` build metadata from the libgit2-sys dependency because it causes cargo to generate a warning every time it is used.


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #12445, opened 2023-08-03
Update pretty_env_logger to 0.5

ehuss:
This updates pretty_env_logger from 0.4.0 to 0.5.0. The primary motivation is to drop some dependencies like `atty`.

https://github.com/seanmonstar/pretty-env-logger/compare/v0.4.0...v0.5.0
I think the main change is updating to env_logger 0.7 to 0.10 which syncs with what is normally used by cargo.


weihanglo (approved):
Seems works on my machine 😆

BTW, I was just thinking about replacing with `tracing`. How about me kicking off with a pull request?

ehuss:
> BTW, I was just thinking about replacing with tracing. How about me kicking off with a pull request?

Using `tracing` sounds good to me. There's a variety of interesting things we could do with it.

Issue #10474, opened 2022-03-11
Tracking Issue for alternative registry authentication (RFC 3139)


ehuss:
### Summary

RFC: [#3139](https://github.com/rust-lang/rfcs/pull/3139) 
Implementation: https://github.com/rust-lang/cargo/pull/10592
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#registry-auth
Issue: https://github.com/rust-lang/cargo/labels/Z-registry-auth

This feature adds the ability to authenticate additional endpoints to a registry, including downloading crates.

### Unresolved Issues

* [x] Do registries need a more fine-grained switch for which API commands require authentication?
* [X] The RFC mentions adding `--token` to additional commands like `install` and `search`, but we are leaning away from allowing tokens from being passed in on the command-line due to the ease of leaking. Should the `--token` flag be added or no?  `--token` won't be added for now.
* [X] Consider changing the name and form of the `X-` header. See https://github.com/rust-lang/rfcs/pull/3139#discussion_r819834286 and https://github.com/rust-lang/rfcs/pull/3139#discussion_r823436983  Cargo now uses the www-authenticate header with the `Cargo` scheme and the `login_url` value, as in `WWW-Authenticate: Cargo login_url="https://test-registry-login/me`.
* [X] Will there be any concerns with the interaction with [RFC 3231](https://github.com/rust-lang/rfcs/pull/3231) (asymmetric tokens)?
* [ ] Require a `credential-provider` to be defined in order to use authenticated registries

Stabilization tracked in #8933

### Future Extensions

* Support authentication with git indexes.  Preferably, cargo will transition to [HTTP indexes](https://github.com/rust-lang/cargo/issues/9069) which will make this not necessary.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

tarcieri:
> The RFC mentions adding --token to additional commands like install and search, but we are leaning away from allowing tokens from being passed in on the command-line due to the ease of leaking. Should the --token flag be added or no?

I would recommend against providing any sort of secret as a command-line argument.

Command line arguments are exposed to the systemwide process environment, which is a risk to any machine running any sort of untrusted process.

[...]

PR #12446, opened 2023-08-04
feat: also try to auth for config if server returns status 403

Fishrock123:
This makes it so that the `registry-auth` unstable feature also attempts to make authorized requests to a registry's `config.json` (in sparse mode) if the server's response code is 403, not just 401.

These status codes have very similar meanings and some 3rd-party registry providers (e.g. Artifactory) prefer to return 403 for unauthorized requests.

Also updates the docs on this feature to reflect this.

Artifactory likes to return 403 for unauthorized requests.

The meanings of HTTP [`401 Unauthorized`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) and HTTP [`403 Forbidden`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) are very similar and largely overlapping.

In order to make Cargo's `registry-auth` feature play better with 3rd-party registries, I would recommend also supporting attempting authentication for 403s even if that is not the perfectly semantic meaning.

Ideally, talk to a 3rd-party registry such as artifactory with this change included and all the config set on the 3rd party registry.

### Additional information

Related tracking issue for the whole `registry-auth` feature: https://github.com/rust-lang/cargo/issues/10474

### DIFF
--- a/src/cargo/sources/registry/http_remote.rs
+++ b/src/cargo/sources/registry/http_remote.rs
@@ -163,6 +163,7 @@ enum StatusCode {
     NotModified,
     NotFound,
     Unauthorized,
+    Forbidden,
 }
 
 /// Represents a complete [`Download`] from an HTTP request.
@@ -315,6 +316,7 @@ impl<'cfg> HttpRegistry<'cfg> {
                     200 => StatusCode::Success,
                     304 => StatusCode::NotModified,
                     401 => StatusCode::Unauthorized,
+                    403 => StatusCode::Forbidden,
                     404 | 410 | 451 => StatusCode::NotFound,
                     _ => {
                         return Err(HttpNotSuccessful::new_from_handle(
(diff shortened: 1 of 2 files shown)


weihanglo:
r? @arlosi 

You're a better reviewer than me for this PR.

arlosi:
The intent of only sending the token for HTTP 401 is to minimize the possibility of sending a token to a server that's not expecting it. Additionally allowing HTTP 403 relaxes that slightly.

Is there a reason why Artifactory can't send HTTP 401 when Cargo is attempting to fetch `config.json`? Other 3rd party registries have implemented this.

Fishrock123:
> Is there a reason why Artifactory can't send HTTP 401 when Cargo is attempting to fetch config.json? Other 3rd party registries have implemented this.

I don’t work for JFrog / Artifactory so I cannot answer that. However it seems like 403 is generally what Artifactory returns for any unauthenticated request.

arlosi:
I'd rather not merge this unless we know we need to, since it makes Cargo slightly more likely to send a token to a URL where it shouldn't. Given that a registry already has to implement the rest of the Cargo protocol, it doesn't seem too difficult to send HTTP 401 for authentication failure.

I suspect that the biggest blocker for registries adding Cargo support likely that `registry-auth` is still unstable, rather than not allowing for authenticating on HTTP 403.

I'm happy to revisit this if we get a compelling request from a registry operator though. Thanks.

sjbeskur:
Just FYI..   
The auth error will also show up if if you have an older / expired CARGO_REGISTRY_TOKEN defined somewhere (~/.bashrc)
cargo login will prefer the env variable before looking in ~/.cargo/credentials.toml

You have to explicitly `unset CARGO_REGISTRY_TOKEN` in order to force publish to use the credentials file

At a minimum we should let the user know where the token was sourced if cargo publish fails.

Happy to open a PR if you agree

PR #12447, opened 2023-08-04
fix: preserve jobserver file descriptors on rustc invocation to get `TargetInfo`

belovdv:
Then cargo asks rustc for `TargetInfo`, cargo passes environment variables that describe jobserver file descriptors, but doesn't actually passed them, because `jobserver-rs` sets `FD_CLOEXEC` by default and expects users to explicitly call `Client::configure` to pass them further. Rustc respects make jobserver, so this causes error - env vars refers to nonexistent file descriptors, which is ignored now.

This PR makes cargo compatible with https://github.com/rust-lang/rust/pull/113730, which attempts to catch such situations.

### DIFF
--- a/src/cargo/core/compiler/build_context/target_info.rs
+++ b/src/cargo/core/compiler/build_context/target_info.rs
@@ -185,6 +185,12 @@ impl TargetInfo {
                 .args(&rustflags)
                 .env_remove("RUSTC_LOG");
 
+            // Removes `FD_CLOEXEC` set by `jobserver::Client` to pass jobserver
+            // as environment variables specify.
+            if let Some(client) = config.jobserver_from_env() {
+                process.inherit_jobserver(client);
+            }
+
             if let CompileKind::Target(target) = kind {
                 process.arg("--target").arg(target.rustc_target());
             }


petrochenkov:
What cargo does when it wants to *really* run rustc (for compilation, with jobserver passed through properly) - changes CLOEXEC flags to false? Creates jobserver-rs client in some other way?

bjorn3:
The `jobserver` crate has a method which forwards the jobserver to a given `Command`.

belovdv:
Method `Client::configure` takes `Command`, while `TargetInfo::new` uses `ProcessBuilder`, so removing env vars seems to be simpler solution.

weihanglo (commented):
Thanks for working on jobserver so far! 

The intent of this pull request is not immediately clear to me. What does this try to resolve in practice? People in the future may be confused and wonderwhy these envs were removed. I know there are links to discussions, but without a summary it's not easy to follow. Besides, assumed this quite important, I believe a test is pretty much wanted. Thus, I'd like to see something beyond code:

* A clear PR description of why this is bad and how it affects other programs.
* Why the solution is desired (remove envs) but not others (e.g. inherit jobserver).
* A test that verifies the “wrong” behavior, and one follow-up commit fixes it.

belovdv:
I've updated PR description.

> how it affects other programs

This PR affects only rustc invocation, so I'm not sure, what do you mean.

> Why the solution is desired but not others

I don't know, if this solution is better. It shouldn't change anything, since now jobserver isn't used and it won't be used with this PR. Also, according to @bjorn3 "That rustc call doesn't use jobserver", so passing jobserver won't change anything, so it's may be better to use simplest solution.

> A test that verifies the “wrong” behavior

My local test is makefile, called `make -j 2`, which contains `+ RUSTC=dummy cargo build`, there `dummy` - some binary, that responds to `dummy -vV`, and checks that make variables (if any) points to existing file descriptors.

One option to write such test is to use some wrapper over rustc, which would verify that passed jobserver-related env vars from cargo don't refer to nonexistent file descriptors, but imho it would be enough to have such check in rustc, which my PR to rustc is intended to add.

petrochenkov:
>The intent of this pull request is not immediately clear to me. What does this try to resolve in practice?

At high level:
- The situation when jobserver exists, but is not being inherited, is often erroneous and at least suspicious.
- https://github.com/rust-lang/rust/pull/113730 attempts to catch such situations.
- This PR makes cargo compatible with https://github.com/rust-lang/rust/pull/113730.

(Even if the check in https://github.com/rust-lang/rust/pull/113730 is later made optional or relaxed due to incompatibilities in third-party tools that we cannot fix, with cargo we can indeed fix it so it's better to make it compatible.)

I'd personally follow https://github.com/rust-lang/cargo/pull/12447#issuecomment-1665360918 and `Client::configure` this rustc run even if it's tiny, just for consistency.

weihanglo:
To fix CI, rebase onto master. See https://github.com/rust-lang/cargo/pull/12450

belovdv:
Branch rebased, thanks.

`ProcessBuilder` calls `Client::configure`, if `inherit_jobserver` provided one. Does this solution suit better (so I should update PR name and description)?

About test - should I add one, or it'd be enough to have check in rustc?

weihanglo (approved):
Thanks! Looks a bit more reasonable to me in terms of consistency.

While updating PR title and description, could you also add a comment of a line or two to explain why Cargo inherits jobserver here?

weihanglo:
Thanks!

@bors r+

Issue #12377, opened 2023-07-19
Determine story around JSON message compatibility


ehuss:
So far, we have treated the JSON messages emitted by `--message-format=json` ([docs](https://doc.rust-lang.org/cargo/reference/external-tools.html#json-messages)) to be somewhat stable (with the caveats of adding new fields). However, that stability has been broken at times (see https://github.com/oli-obk/cargo_metadata/issues/240 as an example). We don't have a strategy for how such changes could be made. We should probably come up with some story and documentation for how we manage the compatibility.

This issue is to collect feedback and ideas on how this should be handled and documented.

Note that this somewhat transitively applies to rustc's diagnostic structure. There should probably be a consistent story with both that and cargo's own messages, which may involve consulting with the compiler team.

----

My thoughts/ideas:

Versioning doesn't work particularly well since introducing a new version that tooling doesn't recognize would cause them to break (and in the debuginfo case, would cause them to unnecessarily to break in most cases since that change only affected the rare projects that would use the string-based types).

We could potentially indicate that *types* of fields can change between versions, but otherwise meanings of existing fields wouldn't change. But that could be a challenge for deserializers like `cargo_metadata` since the only choice to stay compatible would be to deserialize to something like `serde_json::Value`, losing all structure. Or those tools could just accept that type changes would break their tooling.

We could potentially just document that this format is not stable at all, and shed responsibility. An option for tools like `cargo_metadata` is that they could just do semver-breaking changes periodically.

Another option is to document that the format is stable except in the presence of new opt-in features (like the debuginfo string example). This could make it challenging or impossible to make broader structural changes.

ehuss:
The problem with using requested versions is trying to figure out how to map new concepts into the old version. Things like `Cargo.toml` evolve independently of the JSON interface. This issue was triggered by https://github.com/oli-obk/cargo_metadata/issues/240 where the type of a field changed. For example, let's say there was a project with:

```toml
[profile.dev]
debug = "line-tables-only"
```

Format version 1 specified that `debuginfo` was an integer. If cargo tried to keep compatibility with format version 1, what would it show in the `debuginfo` field? No integer value would be correct. Should it generate an error? That's not much different from just breaking the tool (and possibly worse depending on how the tool parses, since most tools don't care about this field, though an explicit error could be clearer for tools that just fail to parse).

ehuss:
> you'd be forced to bump the format version.

This is the part I'm not quite following. Can you clarify what that means?

My interpretation is: 
[...]

PR #12449, opened 2023-08-05
feat(metadata): Add profiles information to `cargo metadata`

### LINKED ISSUE #11228: Cargo metadata: Expose available build profiles
### Problem

Currently cargo metadata does not expose which custom build profiles are available.
For [corrosion](https://github.com/corrosion-rs/corrosion) which integrates cargo into CMake it would be nice to expose a list of available build profiles for users to choose from (e.g. in a graphical interface for configuring CMake cache variables).

### Proposed Solution

Add an additional field to the `package` section which contains an array of available profile names (`dev`, `release`, `test`, `bench` plus any custom profiles).

### Notes

_No response_

calavera:
Fixes #11228 

I do understand that the team is not accepting PRs for issues that have not been accepted. However, I believe these changes are scoped well enough to be considered for review. Nevertheless, feel free to close this PR without reviewing, I understand that I didn't follow the contribution rules. In honesty, I wrote the code before I saw that warning, and I didn't want to keep the code for myself 😅 

This change adds profile information to the output of `cargo metadata`. The output only changes if the user has included any specific profile settings in their `Cargo.toml` file, it doesn't expose default profiles that have not been modified. I opted for serializing this information only when there is information present because I believe it's a minimal incremental improvement that's easier to test, review and accept than if I tried to expose profile information that is not explicitly in `Cargo.toml` files.

I've added integration tests in the metadata test suite for these changes.

### DIFF
--- a/src/cargo/core/profiles.rs
+++ b/src/cargo/core/profiles.rs
@@ -356,7 +356,7 @@ fn create_profile_maker(
                 name
             );
         }
-        // Already inserted from `add_root_profiles`, no need to do anything.
+        // Already created in `root_profiles`, ignore them to prevent overwrites.
         return Ok(None);
     }
 


ehuss:
Sorry about the hiccup, I have posted #12450 to fix the CI issue.

src/cargo/core/package.rs
```diff
@@ -104,6 +105,8 @@ pub struct SerializedPackage {
     metabuild: Option<Vec<String>>,
     default_run: Option<String>,
     rust_version: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    profiles: Option<TomlProfiles>,
```
epage, line 109: > Cargo only looks at the profile settings in the Cargo.toml manifest at the root of the workspace. Profile settings defined in dependencies will be ignored.

imo we should only serialize profiles at the workspace level so people get something more akin to expected cargo behavior

src/cargo/ops/cargo_output_metadata.rs
```diff
@@ -65,6 +72,8 @@ pub struct ExportInfo {
     version: u32,
     workspace_root: PathBuf,
     metadata: Option<toml::Value>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    workspace_profiles: Option<TomlProfiles>,
```
epage, line 76: > The output only changes if the user has included any specific profile settings in their Cargo.toml file, it doesn't expose default profiles that have not been modified. 

For me, the question is what the intended use case is we are covering.  All callers will need to reinvent cargo's profile system to fully interpret this, including reading from config.  Is that ok?  Can we change this in the future if we find it doesn't work?

epage, line 76: Seems like this should be `profiles` instead of `workspace_profiles`

It looks like we only use the `workspace` prefix when qualifying things workspace root, workspace members) but otherwise we don't (e.g. metadata, resolve)

epage:
@rfcbot concern default-config

See https://github.com/rust-lang/cargo/pull/12449#discussion_r1310878610

src/cargo/ops/cargo_output_metadata.rs
epage: One specific use case doesn't tell us much though.  We are looking at adding a new feature and we need to consider other users and future users to make sure we are providing the right thing.

Do we understand enough to move forward either direction?

epage:
> I would like to check with you that that's the right direction

My question isn't so much about what needs to be changed but exploring the requirements for the change itself.  We need to understand that before we know what the right change is.

src/cargo/ops/cargo_output_metadata.rs
jschwe: In my usecase I would like to have the default profiles available. My usecase would be to provide CMake users an option to select one of the available cargo profiles. For this I would like to have
a) name of the profile
b) path of the profile (mainly because `dev` maps to `debug`)

This list should include the default profiles, although I'm not sure about `test` and `bench` though, since those profiles are weird and artifacts don't land at a predictable location due to the hash being added.

src/cargo/core/profiles.rs
```diff
@@ -370,13 +268,170 @@ impl Profiles {
...
+        (InternedString::new("bench"), InternedString::new("release")),
+    ]
+    .into()
+}
+
+/// Returns the "root" profiles (dev/release)
+fn root_profiles(
+    profiles: &BTreeMap<InternedString, TomlProfile>,
+) -> Vec<(InternedString, ProfileMaker)> {
+    vec![
```
weihanglo, line 287: We dont need a `Vec` here I guess? an `[(InternedString, ProfileMaker); 2]` should be sufficient.

src/cargo/core/profiles.rs
```diff
@@ -370,13 +268,170 @@ impl Profiles {
...
+            InternedString::new("release"),
+            ProfileMaker::new(Profile::default_release(), profiles.get("release").cloned()),
+        ),
+    ]
+}
+
+/// Returns the built-in profiles (not including dev/release, which are
+/// "root" profiles).
+fn predefined_profiles() -> Vec<(&'static str, TomlProfile)> {
+    vec![
```
weihanglo, line 302: ditto. `[(&'static str, TomlProfile); 3]` is good to go.

tests/testsuite/metadata.rs
```diff
@@ -6,6 +6,84 @@ use cargo_test_support::registry::Package;
 use cargo_test_support::{basic_bin_manifest, basic_lib_manifest, main_file, project, rustc_host};
 use serde_json::json;
 
+pub(crate) const DEFAULT_PROFILES: &str = r#"{
+  "bench": {
+    "codegen_backend": null,
```
weihanglo, line 11: Should `codegen_backend` be skipped, as the feature is not stable yet?

tests/testsuite/metadata.rs
```diff
@@ -6,6 +6,84 @@ use cargo_test_support::registry::Package;
...
+    "codegen_units": null,
+    "debug_assertions": false,
+    "debuginfo": 0,
+    "incremental": false,
+    "lto": "false",
```
weihanglo, line 16: it seems odd that `lto` false is a stringified `"false"`.

src/cargo/core/profiles.rs
```diff
@@ -870,6 +925,18 @@ impl fmt::Display for Strip {
...
 
+impl serde::ser::Serialize for Strip {
+    fn serialize<S>(&self, s: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::ser::Serializer,
+    {
+        match self {
+            Strip::None => "none".serialize(s),
+            Strip::Named(n) => n.serialize(s),
+        }
```
weihanglo, line 936:
```suggestion
        s.collect_str(self)
```

BTW it would be great if we can make commit atomic. Like, this change can fit it one commit of its own.

src/cargo/core/profiles.rs
```diff
@@ -62,196 +63,84 @@ pub struct Profiles {
...
+                makers.insert(name, maker);
+            }
+        }
+
+        for (name, profile) in &definitions {
+            let name = *name;
+            if let Some(maker) = create_profile_maker(name, profile, &makers, &definitions)? {
+                makers.insert(name, maker);
+            }
+        }
```
weihanglo, line 90: Duplicate?

```suggestion
```

src/cargo/core/profiles.rs
```diff
@@ -62,196 +63,84 @@ pub struct Profiles {
...
         // Verify that the requested profile is defined *somewhere*.
         // This simplifies the API (no need for CargoResult), and enforces
         // assumptions about how config profiles are loaded.
-        profile_makers.get_profile_maker(requested_profile)?;
+        get_profile_maker(&profile_makers.by_name, requested_profile)?;
         Ok(profile_makers)
     }
 
-    /// Returns the hard-coded directory names for built-in profiles.
-    fn predefined_dir_names() -> HashMap<InternedString, InternedString> {
```
weihanglo, line 112: This seems like a refactor in order to reuse code between `Profiles::new` and `Proflies::all`. I would suggest commit the refactor in a commit, and the other for adding the new feature. It helps both review and tracing Git history.

weihanglo:
@rfcbot concern "JSON message compatibility"


This PR insta-stabilizes all profiles settings in JSON format. However, we do have some issues around JSON message compatbility, especially when we want to add more types to a field later. The issue is tracked in <https://github.com/rust-lang/cargo/issues/12377>.

Do we want to resolve #12377 first then merge this PR? Should we look into each field and determine the serialization format one by one? Or both?

weihanglo:
Thank you @calavera for working hard on this from different angles!

PR #12454, opened 2023-08-06
Fix cargo remove incorrectly removing used patches

### LINKED ISSUE #12419: `cargo rm` unexpectedly removes patches
### Problem

When I run `cargo rm dependency`, I expect Cargo to remove the crate `dependency` from the current crate's \[dependencies\], and other related necessary changes, like removing `dependency` from the right-hand side of features that previously enabled an optional dependency on "dependency" or one of its features "dependency/&hellip;".

Cargo does these things, but it also wipes out my `patch.crates-io` section from the workspace's manifest. I could not figure out how this seems related in any way to removing a single dependency, so I think this is a bug.

### Steps

```toml
# Cargo.toml

[workspace]
members = ["serde", "serde_derive"]

[patch.crates-io]
serde = { path = "serde" }
```

```toml
# serde/Cargo.toml

[package]
name = "serde"
version = "1.0.0"

[dependencies]
serde_derive = { path = "../serde_derive" }
```

```toml
# serde_derive/Cargo.toml

[package]
name = "serde_derive"
version = "1.0.0"
```

+`serde/src/lib.rs` and `serde_derive/src/lib.rs` which can be empty files.

In the `serde` directory, run:

```console
$ cargo check  # sanity check--the manifests are valid

$ cargo rm serde_derive
```

Expected diff:

```diff
diff --git a/serde/Cargo.toml b/serde/Cargo.toml
index 759de5d..10027d7 100644
--- a/serde/Cargo.toml
+++ b/serde/Cargo.toml
@@ -3,6 +3,3 @@
 [package]
 name = "serde"
 version = "1.0.0"
-
-[dependencies]
-serde_derive = { path = "../serde_derive" }
```

Actual diff produced by Cargo:

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 583dd2f..60ffc4f 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,3 @@
 
 [workspace]
 members = ["serde", "serde_derive"]
-
-[patch.crates-io]
-serde = { path = "serde" }
diff --git a/serde/Cargo.toml b/serde/Cargo.toml
index 759de5d..10027d7 100644
--- a/serde/Cargo.toml
+++ b/serde/Cargo.toml
@@ -3,6 +3,3 @@
 [package]
 name = "serde"
 version = "1.0.0"
-
-[dependencies]
-serde_derive = { path = "../serde_derive" }
```

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
cargo 1.73.0-nightly (7ac9416d8 2023-07-24)
release: 1.73.0-nightly
commit-hash: 7ac9416d82cd4fc5e707c9ec3574d22dff6466e5
commit-date: 2023-07-24
```

alcolmenar:

Fixes an issue where patches are being removed when member dependencies don't explicitly contain the patched crate.

Fixes #12419 

- Created a test for the failing use case
- Verify passing test


tests/testsuite/cargo_remove/invalid_gc_patch/out/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+# Cargo.toml
+
+[workspace]
+members = ["serde", "serde_derive"]
+
+[patch.crates-io]
+serde = { path = "serde" }
+
```
epage, line 8: > [chore: add failing test](https://github.com/rust-lang/cargo/pull/12454/commits/9d3fcfab35ce0c063c49a8a5ab19fb16646a9f7b)

Please make the commit that adds the test pass.  Then when you have a commit that changes the behavior, this test will then need to be updated.  As a reviewer, I can then see how the behavior changes with the code change.  This also makes it nicer for bisections as the tests are always passing.

src/bin/cargo/commands/remove.rs
```diff
@@ -109,9 +110,14 @@ pub fn exec(config: &mut Config, args: &ArgMatches) -> CliResult {
...
         // Reload the workspace since we've changed dependencies
         let ws = args.workspace(config)?;
-        resolve_ws(&ws)?;
-    }
+        let (_, resolve) = resolve_ws(&ws)?;
 
+        // Attempt to gc unused patches and re-resolve if anything is removed
+        if gc_unused_patches(&workspace, &resolve)? {
+            let ws = args.workspace(config)?;
+            resolve_ws(&ws)?;
```
epage, line 118: I've not looked at the lockfile for one of these but I take it the lockfile needs to be updated when removing an unused patch?

Should we include the lockfile in the test so we can make sure this continues working?

tests/testsuite/cargo_remove/invalid_gc_patch/out/Cargo.toml
```diff
@@ -3,3 +3,6 @@
 [workspace]
 members = ["serde", "serde_derive"]
 
+[patch.crates-io]
```
epage, line 6: I assume this is tied into the previous commit. Could you squash this in so each commit stands on its own with passing tests?

tests/testsuite/cargo_remove/gc_patch/out/Cargo.toml
```diff
@@ -5,5 +5,8 @@ members = [ "my-member" ]
 name = "my-project"
 version = "0.1.0"
 
+[patch."[ROOTURL]/bar1"]
+bar = { git = "[ROOTURL]/bar2" }
+
 [patch.crates-io]
 bar = { git = "[ROOTURL]/bar2" }
```
epage, line 12: Something doesn't seem to be working right

This test has
- my-project depend on bar from git and does a `cargo remove` on it
- my-member depends on bar from crates.io

The crates-io patch should remain because that is untouched but the git patch should be removed.

tests/testsuite/cargo_remove/gc_patch/out/Cargo.toml
```diff
@@ -5,5 +5,8 @@ members = [ "my-member" ]
 name = "my-project"
 version = "0.1.0"
 
+[patch."[ROOTURL]/bar1"]
```
epage, line 8: btw like the last commit, I assume this is tied into the previous commit. Could you squash this in so each commit stands on its own with passing tests?

src/bin/cargo/commands/remove.rs
alcolmenar: Hmm this particular test case doesn't actually have unused patches populated. I can make the test so that it does though. Is it better for that to be a separate test?

tests/testsuite/cargo_remove/invalid_gc_patch/out/Cargo.toml
alcolmenar: Will do. Is the goal to make it so every commit has no failing tests? I actually don't understand why. I read about bisection but it's not clear how that comes into play on a branch. Does it have to do with how the branch is merged in?

tests/testsuite/cargo_remove/gc_patch/out/Cargo.toml
alcolmenar: I spent some time here trying to figure out why this was the case as well. I inspected the sandbox for the test and I had noticed that the patch wasn't being set as unused evidenced by the lack of emitted warning and the Cargo.lock file didn't have any `patch.unused` sections.

I'll look into this a bit more.

tests/testsuite/cargo_remove/gc_patch/out/Cargo.toml
alcolmenar: will do

tests/testsuite/cargo_remove/gc_patch/out/Cargo.toml
alcolmenar: Ok so the issue is that both patch dependency have the same PackageId. Since one of the patches is used and both patches from the same PackageId, neither of the patches show up as unused. There doesn't really appear to be an easy way to determine if a specific dependency was patched so it looks like to reliably know if a patch is unused, it'd have to be populated during graph generation.

Do we support this use case where multiple dependencies with different sources are overridden by the same patch dependency?

src/bin/cargo/commands/remove.rs
epage: We have an existing test for patch gc, so it could be in there

tests/testsuite/cargo_remove/invalid_gc_patch/out/Cargo.toml
epage: Ideally, we don't squash on merge but preserve the commits.

epage: Also, it makes it clearer to the review when they can see all of the related changes in one commit to know what is related

alcolmenar: Ah ok that makes sense now

tests/testsuite/cargo_remove/gc_patch/out/Cargo.toml
alcolmenar: Realized that I wasn't writing to the manifest so I updated it. Also, I'm taking the Dependency Url into account when looking through the list of unused_patches just in case there are multiple with the same name

tests/testsuite/cargo_remove/gc_patch/stderr.log
```diff
@@ -1,3 +1,8 @@
     Removing bar from dependencies
-    Updating git repository `[ROOTURL]/bar2`
+    Updating git repository `[ROOTURL]/bar[..]`
+    Updating git repository `[ROOTURL]/bar[..]`
     Updating `dummy-registry` index
+warning: Patch `bar v0.1.0 ([..])` was not used in the crate graph.
+Perhaps you misspelled the source URL being patched.
+Possible URLs for `[patch.<URL>]`:
+    [ROOT]/bar2
```
alcolmenar, line 8: This is a result of resolving twice. The first time unused patches are found so this is emitted. Later they are gc and re-resolve so we don't get the same warning twice

tests/testsuite/cargo_remove/mod.rs
```diff
@@ -8,6 +8,7 @@ mod gc_replace;
 mod help;
 mod invalid_arg;
 mod invalid_dep;
+mod invalid_gc_patch;
```
epage, line 11: Sorry, didn't pay enough attention before but this test name isn't accurate.  This isn't testing something invalid but it is verifying we GC preserves relevant patches.

I'm assuming "invalid" came from this showing the incorrect behavior of the GC but once the commit that fixes it is in, then this changes to preventing regressions of ithat.

tests/testsuite/cargo_remove/gc_patch/stderr.log
epage: This will be very confusing to users and should be resolved.

A quick hack would be to capture the verbosity, set it to quiet, resolve, then restore the verbosity.  We *might* be able to improve it with #12235 by pushing an "allow" on that warning and then popping it.

tests/testsuite/cargo_remove/mod.rs
alcolmenar: that's right. I created the test initially before realizing what the issue was or how to fix it. I'll update the name accordingly

tests/testsuite/cargo_remove/gc_patch/stderr.log
alcolmenar: Agreed. Is there any example of this in the code? I'm not quite sure how to do that

epage: Not that I can think of.  The relevant calls are
- https://doc.rust-lang.org/nightly/nightly-rustc/cargo/util/config/struct.Config.html#method.shell
- https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/shell/struct.Shell.html#method.set_verbosity
- https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/shell/struct.Shell.html#method.verbosity

alcolmenar: Nevermind I figured it out

alcolmenar: unfortunately, this is removing other status messages like 
```
```

epage: Could you post it so I can see?

alcolmenar: done

alcolmenar: Thank you for the guidance and the code snippet. The snippet helped me realize the bug here 👍

epage:
Thanks for all of the work you put into researching and fixing this!

@bors r+

alcolmenar:
> Thanks for all of the work you put into researching and fixing this!

Definitely was my pleasure! Thank you for your guidance. It was a good first experience on the project

PR #12458, opened 2023-08-07
refactor: migrate to `tracing`

weihanglo:

The migrates `CARGO_LOG` logging from `log` to `tracing`.

`tracing` is already widely used among the ecosystem and rustc adopted it for a while. We've chatted about this a while back in the weekly meeting.

Last week in the office hour, epage, Muscraft, and arlosi have no objection to this when I mentioned this upcoming PR. ehuss seems to looking forward to the migration <https://github.com/rust-lang/cargo/pull/12445#issuecomment-1664753017>. Hence posted this pull request.

By commits. And please check if `CARGO_LOG` works for you.

Largely, a search & replace worked just fine. Only in `util/network/http.rs` I did a manual fix.

The format is also a bit different. Should we make it compatible? (I don't want to though)

```apachelog
# before
[2023-08-07T10:53:48Z DEBUG cargo::sources::registry::http_remote] loading config
# after
2023-08-07T10:43:47.867106Z DEBUG cargo::sources::registry::http_remote: loading config
```

By default `log` enables all levels, whereas the built-in `fmt` subscriber in tracing-subscriber [enables up to `INFO` only](https://docs.rs/tracing-subscriber/0.3/tracing_subscriber/fmt/struct.Subscriber.html#associatedconstant.DEFAULT_MAX_LEVEL). Is that a thing we need to consider? I do feel like INFO is a better default.

### Additional information

```
-2 dependencies
+11 dependencies
```

Last year during 1.63.0 release cargo only had ~180 deps, but today it doubles the number to ~337 🥲.


epage:
> The format is also a bit different. Should we make it compatible? (I don't want to though)

Compatibility only matters if we intend people to parse it which I don't think we do.

The only other issue is usability which it is probably good enough

Cargo.toml
```diff
@@ -94,6 +91,8 @@ thiserror = "1.0.44"
 time = { version = "0.3", features = ["parsing", "formatting", "serde"] }
 toml = "0.7.6"
 toml_edit = "0.19.14"
+tracing = "0.1.37"
+tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
```
epage, line 95: Should we also use [tracing-log](https://crates.io/crates/tracing-log) for capturing / controlling any dependencies that still use `log`?

arlosi: Probably a good idea.  `git2` still uses `log`, and there may be others.

arlosi (approved):

Cargo.toml
weihanglo: The compatible already should be included already I believe. See [`tracing_subscriber::SubscriberInitExt::init`](https://docs.rs/tracing-subscriber/0.3/tracing_subscriber/util/trait.SubscriberInitExt.html#method.init).

<details><summary>Output of <code>cargo tree -i log</code></summary>
<p>

```
log v0.4.19
├── cargo-util v0.2.5 (/projects/cargo/crates/cargo-util)
│   ├── cargo v0.74.0 (/projects/cargo)
│   └── cargo-test-support v0.1.0 (/projects/cargo/crates/cargo-test-support)
│       [dev-dependencies]
│       └── cargo v0.74.0 (/projects/cargo)
├── git2 v0.17.2
│   ├── cargo v0.74.0 (/projects/cargo)
│   ├── cargo-test-support v0.1.0 (/projects/cargo/crates/cargo-test-support) (*)
│   └── git2-curl v0.18.0
│       └── cargo v0.74.0 (/projects/cargo)
├── git2-curl v0.18.0 (*)
├── gix v0.45.1
│   └── cargo v0.74.0 (/projects/cargo)
├── gix-attributes v0.13.1
│   ├── gix v0.45.1 (*)
│   └── gix-worktree v0.18.0
│       └── gix v0.45.1 (*)
├── gix-config v0.23.0
│   └── gix v0.45.1 (*)
├── globset v0.4.10
│   └── ignore v0.4.20
│       └── cargo v0.74.0 (/projects/cargo)
├── ignore v0.4.20 (*)
├── os_info v3.7.0
│   └── cargo v0.74.0 (/projects/cargo)
├── rustfix v0.6.1
│   └── cargo v0.74.0 (/projects/cargo)
└── tracing-log v0.1.3
    └── tracing-subscriber v0.3.17
        └── cargo v0.74.0 (/projects/cargo)
```

</p>
</details> 

Here is a patch you can apply and run `CARGO_LOG=cargo_util cargo update -p cargo` to test against.

<details><summary>patch</summary>
<p>

```
﻿diff --git a/Cargo.lock b/Cargo.lock
index b0c1047a5..8081aa6f4 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -431,6 +431,7 @@ dependencies = [
  "hex",
  "jobserver",
  "libc",
+ "log",
  "miow",
  "same-file",
  "sha2",
diff --git a/crates/cargo-util/Cargo.toml b/crates/cargo-util/Cargo.toml
index 36ca7bc17..0d7f25cb0 100644
--- a/crates/cargo-util/Cargo.toml
+++ b/crates/cargo-util/Cargo.toml
@@ -19,6 +19,7 @@ shell-escape.workspace = true
 tempfile.workspace = true
 tracing.workspace = true
 walkdir.workspace = true
+log = "0.4.19"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 core-foundation.workspace = true
diff --git a/crates/cargo-util/src/paths.rs b/crates/cargo-util/src/paths.rs
index ce6755859..400b412cf 100644
--- a/crates/cargo-util/src/paths.rs
+++ b/crates/cargo-util/src/paths.rs
@@ -140,6 +140,7 @@ pub fn resolve_executable(exec: &Path) -> Result<PathBuf> {
 ///
 /// Equivalent to [`std::fs::read_to_string`] with better error messages.
 pub fn read(path: &Path) -> Result<String> {
+    log::info!("I am from `log` crate in cargo_util::paths::read");
     match String::from_utf8(read_bytes(path)?) {
         Ok(s) => Ok(s),
         Err(_) => anyhow::bail!("path at `{}` was not valid utf-8", path.display()),
```

</p>
</details>

weihanglo:
I'd like to call out the other change. By default `log` enables all levels, whereas the built-in `fmt` subscriber in tracing-subscriber [enables up to `INFO` only](https://docs.rs/tracing-subscriber/0.3/tracing_subscriber/fmt/struct.Subscriber.html#associatedconstant.DEFAULT_MAX_LEVEL). Is that a thing we need to consider? I do feel like INFO is a better default.

Cargo.toml
```diff
@@ -94,6 +91,8 @@ thiserror = "1.0.44"
 time = { version = "0.3", features = ["parsing", "formatting", "serde"] }
 toml = "0.7.6"
 toml_edit = "0.19.14"
+tracing = "0.1.37"
```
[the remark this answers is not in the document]
weihanglo: That's exactly what I was thinking about! It would be better if we could migrate some logging to `#[instrument]`.

Cargo.toml
epage: Oh, didn't realize it had built-in support and it was enabled by default!

weihanglo:
Timed out again…

@bors retry

PR #12463, opened 2023-08-07
prompt the use of `--nocapture` flag if `cargo test` process is terminated via a signal.

### LINKED ISSUE #10855: Suggest `--nocapture` flag when a test fails with abort
### Problem

Consider the following test 

```rust
#[test]
fn rayon_panic() {
    rayon::spawn(|| panic!("o_O"));
    loop {}
}
```

`cargo test` for it produces the following output: 

```console
λ cargo t
   Compiling rt v0.1.0 (/home/matklad/tmp/rt)
    Finished test [unoptimized + debuginfo] target(s) in 0.17s
     Running unittests src/lib.rs (target/debug/deps/rt-fb003f4fc4a43299)

running 1 test
error: test failed, to rerun pass '--lib'

Caused by:
  process didn't exit successfully: `/home/matklad/tmp/rt/target/debug/deps/rt-fb003f4fc4a43299` (signal: 6, SIGABRT: process abort signal)
```

Note how it's super unclear what exactly failed and why. Passing `--nocapture` flags helps a bit, but only an expert user would realize that such a flag might be helpful!

```console
λ cargo t -- --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 0.01s
     Running unittests src/lib.rs (target/debug/deps/rt-fb003f4fc4a43299)

running 1 test
thread '<unnamed>' panicked at 'o_O', src/lib.rs:3:21
Rayon: detected unexpected panic; aborting
error: test failed, to rerun pass '--lib'

Caused by:
  process didn't exit successfully: `/home/matklad/tmp/rt/target/debug/deps/rt-fb003f4fc4a43299 --nocapture` (signal: 6, SIGABRT: process abort signal)
```

---

What happens here is that the test calls `std::process::abort`. Because, by default, `cargo test` captures stderr, and because abort doesn't unwind, libtest doesn't get a chance to flush stderr. This seems like a fundamental limitation of libtest.

### Proposed Solution

It'd be cool if cargo realised that `cargo test` process was killed with a singnal, and suggested re-running with `--nocapture` in this case. Note that this won't *always* help (it might be the case that a test died without printing anything to stderr), but it doesn't seem that the note like the following would hurt: 

```
error: test failed, to rerun pass '--lib'
 note: test was terminated by the signal, stderr might be truncated, pass `--nocapture` disable output buffering.
```

### Notes

_No response_

stupendoussuperpowers:
Fixes #10855 

As per the discussion on this issue, we want to prompt the user to use `--nocapture` if a test is terminated abnormally. The motivation for this change is described in the issue.

We check for 3 things before we display this flag. - 
- `!is_simple` (if the test ended with a non 101 status code)
- `harness` (if the standard test harness was used), and
- `!nocapture` (whether or not the `--nocapture` flag was already passed to the test)

There's further tests added to `test::nonzero_exit_status` that check that the `stderr` is correct for the various combinations possible when a test ends with a non-101 status code.

The new expected behavior is - 
- Display `--nocapture` note for only non-zero exit statuses, when the `--nocapture` flag is not passed. 
- Only display the note if we use a standard test harness since custom test harnesses do not implement the `--nocapture` flag.

To implement the check for the `--nocapture` flag, the function definition for `report_test_errors` was changed to add the `test_args: &[&str]` parameter. This parameter is passed from the immediate calling function. This private function is only called twice change and is not causing regression after making the appropriate changes to both the places it's called in.

### DIFF
--- a/src/cargo/ops/cargo_test.rs
+++ b/src/cargo/ops/cargo_test.rs
@@ -149,7 +149,7 @@ fn run_unit_tests(
                 unit: unit.clone(),
                 kind: test_kind,
             };
-            report_test_error(ws, &options.compile_opts, &unit_err, e);
+            report_test_error(ws, test_args, &options.compile_opts, &unit_err, e);
             errors.push(unit_err);
             if !options.no_fail_fast {
                 return Err(CliError::code(code));
@@ -275,7 +275,7 @@ fn run_doc_tests(
                 unit: unit.clone(),
                 kind: TestKind::Doctest,
             };
-            report_test_error(ws, &options.compile_opts, &unit_err, e);
+            report_test_error(ws, test_args, &options.compile_opts, &unit_err, e);
             errors.push(unit_err);
             if !options.no_fail_fast {
                 return Err(CliError::code(code));
@@ -407,6 +407,7 @@ fn no_fail_fast_err(
 /// Displays an error on the console about a test failure.
 fn report_test_error(
     ws: &Workspace<'_>,
+    test_args: &[&str],
     opts: &ops::CompileOptions,
     unit_err: &UnitTestError,
     test_error: anyhow::Error,
@@ -420,13 +421,23 @@ fn report_test_error(
     let mut err = format_err!("{}, to rerun pass `{}`", which, unit_err.cli_args(ws, opts));
     // Don't show "process didn't exit successfully" for simple errors.
     // libtest exits with 101 for normal errors.
-    let is_simple = test_error
+    let (is_simple, executed) = test_error
         .downcast_ref::<ProcessError>()
         .and_then(|proc_err| proc_err.code)
-        .map_or(false, |code| code == 101);
+        .map_or((false, false), |code| (code == 101, true));
+
     if !is_simple {
         err = test_error.context(err);
     }
 
     crate::display_error(&err, &mut ws.config().shell());
+
+    let harness: bool = unit_err.unit.target.harness();
+    let nocapture: bool = test_args.contains(&"--nocapture");
+
+    if !is_simple && executed && harness && !nocapture {
+        drop(ws.config().shell().note(
+            "test exited abnormally; to see the full output pass --nocapture to the harness.",
+        ));
+    }
 }
--- a/tests/testsuite/test.rs
+++ b/tests/testsuite/test.rs
@@ -4799,12 +4799,47 @@ Caused by:
         .with_status(4)
         .run();
 
+    p.cargo("test --test t2 -- --nocapture")
+        .with_stderr(
+            "\
+[FINISHED] test [..]
+[RUNNING] tests/t2.rs (target/debug/deps/t2[..])
+error: test failed, to rerun pass `--test t2`
+
+Caused by:
+  process didn't exit successfully: `[ROOT]/foo/target/debug/deps/t2[..]` (exit [..]: 4)
+",
+        )
+        .with_status(4)
+        .run();
+
     // no-fail-fast always uses 101
     p.cargo("test --no-fail-fast")
         .with_stderr(
             "\
 [FINISHED] test [..]
 [RUNNING] tests/t1.rs (target/debug/deps/t1[..])
+error: test failed, to rerun pass `--test t1`
+[RUNNING] tests/t2.rs (target/debug/deps/t2[..])
+error: test failed, to rerun pass `--test t2`
+
+Caused by:
+  process didn't exit successfully: `[ROOT]/foo/target/debug/deps/t2[..]` (exit [..]: 4)
+error: 2 targets failed:
+    `--test t1`
+    `--test t2`
+",
+        )
+        .with_status(101)
+        .run();
+
+    p.cargo("test --no-fail-fast -- --nocapture")
+        .with_stderr(
+            "\
+[FINISHED] test [..]
+[RUNNING] tests/t1.rs (target/debug/deps/t1[..])
+thread 't' panicked at 'this is a normal error', tests/t1[..]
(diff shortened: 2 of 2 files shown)


src/cargo/ops/cargo_test.rs
```diff
@@ -431,9 +431,6 @@ fn report_test_error(
 
     crate::display_error(&err, &mut ws.config().shell());
 
-    // println!("{:#?}", opts);
-    // println!("{:#?}", test_args);
```
epage: Could you squash this commit?

tests/testsuite/test.rs
```diff
@@ -4792,6 +4792,21 @@ error: test failed, to rerun pass `--test t1`
...
 
+Caused by:
+  process didn't exit successfully: `[ROOT]/foo/target/debug/deps/t2[..]` (exit [..]: 4)
+note: test was terminated via a signal. stderr might be truncated. use `--nocapture` flag to disable output buffering.
+",
+        )
+        .with_status(4)
+        .run();
+
+    p.cargo("test --test t2 -- --nocapture")
```
epage, line 4803: Can you split these new test cases out into a commit (first in the PR).  It should pass, showing the behavior of `master`.  The commit with all of the other changes should then update the test to still pass.  This will make it easier to see (1) the tests work and (2) what behavior changed

src/cargo/ops/cargo_test.rs
```diff
@@ -429,4 +430,13 @@ fn report_test_error(
     }
 
     crate::display_error(&err, &mut ws.config().shell());
+
+    let harness: bool = unit_err.unit.target.harness();
+    let nocapture: bool = test_args.contains(&"--nocapture");
+
+    if !is_simple && harness && !nocapture {
+        drop(ws.config().shell().note(
+            "test was terminated via a signal. stderr might be truncated. use `--nocapture` flag to disable output buffering."
```
epage, line 439: I feel like this is long enough that its going to lose users (and technically a test can exist from something besides a signal, like `std::process::exit`)

How about

> "test exited abnormally; to see the full output pass `--nocapture` to the harness"

stupendoussuperpowers:
I've split the changes into 2 commits as you asked, and changed the display string. 

The tests are failing right now because it is outputting the --nocapture note even if there's an exec failure. Looking into a fix for this.

ehuss:
r? epage

Hope you don't mind taking this.

tests/testsuite/test.rs
```diff
@@ -4792,6 +4792,35 @@ error: test failed, to rerun pass `--test t1`
...
 
+Caused by:
+  process didn't exit successfully: `[ROOT]/foo/target/debug/deps/t2[..]` (exit [..]: 4)
+note: test exited abnormally; to see the full output pass --nocapture to the harness.
+",
+        )
+        .with_status(4)
+        .run();
+
+    p.cargo("test --test t2 -- --nocapture")
```
epage, line 4803: Looks like there was a merge problem and the second commit re-adds the test cases the first commit added.

stupendoussuperpowers: Let me take a look at all these commits once, they've gotten a bit confusing. Will push with correct changes in a bit.

stupendoussuperpowers:
Hi @epage need some help with this. 

`cargo test -p cargo` for `linux x86_64 beta` is passing, but the test fails for `linux x86_64 nightly.` Looks like the `--no-fail-fast --nocapture` are outputting different things for these two. Not sure how to deal with this.

epage:
> Hi @epage need some help with this.
> 
> `cargo test -p cargo` for `linux x86_64 beta` is passing, but the test fails for `linux x86_64 nightly.` Looks like the `--no-fail-fast --nocapture` are outputting different things for these two. Not sure how to deal with this.

`[..]` are a type of wildcard.  You can adjust these wildcards to make it still show the relevant information (`t1.rs`) while hiding the information that is tied to a specific rust version

epage:
@stupendoussuperpowers could you rebase, instead of merge, so I can more easily browse the individual commits?

stupendoussuperpowers:
give me a day or so to work on this, really embarrassing but i'm running into multiple git issues on my local machine. will mark it ready once it's done.

epage:
btw I'm hosting office hours right now but also available at other times in case you want to do a call.  See https://github.com/rust-lang/cargo/wiki/Office-Hours

epage:
Hmm, forgot to look back at this.  @stupendoussuperpowers anything left or is this good to go?

epage:
@bors r+

Thanks for all the work you did on this!

weihanglo:
```
 thread 'git_auth::ssh_something_happens' panicked at '
test failed running `/home/runner/work/cargo/cargo/target/debug/cargo check -v`
error: expected to find:
[..]Connection [..] by [..]

did not find in output:
    Updating git repository `ssh://127.0.0.1:37377/foo/bar`
warning: spurious network error (3 tries remaining): failed to fill whole buffer
error: failed to get `bar` as a dependency of package `foo v0.0.1 (/home/runner/work/cargo/cargo/target/tmp/cit/t99/foo)`

Caused by:

Caused by:
  Unable to update ssh://127.0.0.1:37377/foo/bar

Caused by:
  failed to clone into: /home/runner/work/cargo/cargo/target/tmp/cit/t99/home/.cargo/git/db/bar-d4c725e0d0110fc9

Caused by:
  network failure seems to have happened
  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here

Caused by:
  An IO error occurred when talking to the server

Caused by:
  ssh: connect to host 127.0.0.1 port 37377: Connection refused
', tests/testsuite/git_auth.rs:301:10
```

Not sure what's going on. Could be the TCP connection accidentally closed.

Let's @bors retry and see.

PR #12466, opened 2023-08-08
chore(gh): Expand update window

epage:
I've been having some repos not updated lately.  RenovateBot project's theory is that its because RenovateBot is scheduled to run once every 3 hours and checks the schedule at that point while the schedules I use only offers an exact 3 hour window, making this a race condition.  The schedule originally came from RenovateBot but they changed their polling schedule.

See https://github.com/renovatebot/renovate/discussions/23642#discussioncomment-6618560

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,6 +1,6 @@
 {
   schedule: [
-    'before 3am on the first day of the month',
+    'before 5am on the first day of the month',
   ],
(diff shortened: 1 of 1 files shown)


hi-rustin (approved):
Looks good to me.

weihanglo (approved):

weihanglo:
Make sense 

@bors +

PR #12468, opened 2023-08-09
Fix panic when enabling http.debug for certain strings

### LINKED ISSUE #12459: Debug logging causes panic while looking for secrets to redact
A `cargo metadata` call spuriously causes a panic while downloading dependencies, possibly during a time of intermittent network issues on GitHub. The panic is in some cargo debug logging code that is attempting to redact secret information like cookies as headers are being logged.

Error:
```
thread 'main' panicked at 'byte index 14 is not a char boundary; it is inside 'ꓦ' (bytes 12..15) of `>iꓦ>iꓦ>iꓦ>iꓦ>iꓦ>i`', src/cargo/ops/registry.rs:715:17
```
The line in question: https://github.com/rust-lang/cargo/blob/c0eb6cc012af6d6555c761729f7fb5f5254b40ab/src/cargo/ops/registry.rs#L715

Log file:
[error.log](https://github.com/rust-lang/cargo/files/12280249/error.log)

Similar issues here:
https://github.com/libp2p/rust-libp2p/actions/runs/5782756485/job/15670482772#step:41:715
https://github.com/libp2p/rust-libp2p/actions/runs/5782750882/job/15670392407

Underlying cause of: https://github.com/obi1kenobi/cargo-semver-checks/issues/513

arlosi:
Fixes an issue where enabling HTTP debugging may attempt to slice a string across a character boundary resulting in a panic. This likely occurs when binary HTTP data happens to be valid UTF-8.

By interpreting the strings as `&[u8]` before slicing, (which is what `eq_ignore_ascii_case` did internally anyway), the panic is no longer possible.

Closes #12459

r? @weihanglo

### DIFF
--- a/src/cargo/util/network/http.rs
+++ b/src/cargo/util/network/http.rs
@@ -152,6 +152,8 @@ pub fn configure_http_handle(config: &Config, handle: &mut Easy) -> CargoResult<
                 _ => return,
             };
             let starts_with_ignore_case = |line: &str, text: &str| -> bool {
+                let line = line.as_bytes();
+                let text = text.as_bytes();
                 line[..line.len().min(text.len())].eq_ignore_ascii_case(text)
             };
             match str::from_utf8(data) {


weihanglo (approved):
Clever. Thanks!

PR #12480, opened 2023-08-11
Change order of registry.global-credential-providers

arlosi:

`registry.global-credential-providers` merges configuration from multiple files, with configuration coming from more-specific files being placed *earlier* in the list. The original intent was that these credential providers would be tried first, however the current implementation reverses the merged list, which results in more-specific providers being attempted later.

Configuration from the environment *does* get placed at the end of the list, which complicates this, since the value from the environment should be attempted *first*.

### Change summary
* Attempt providers from the list in forward order
* Explicitly insert the value from the environment at the beginning of the list if it exists.

* Adds a test to validate behavior when all credential providers fail
* Adds a test to validate behavior with configuration coming from multiple files and the environment

Edit: this PR blocked on the outcome of #12506

### DIFF
--- a/src/cargo/util/auth/mod.rs
+++ b/src/cargo/util/auth/mod.rs
@@ -22,7 +22,7 @@ use crate::util::credential::adaptor::BasicProcessCredential;
 use crate::util::credential::paseto::PasetoCredential;
 
 use super::{
-    config::{CredentialCacheValue, OptValue, PathAndArgs},
+    config::{CredentialCacheValue, Definition, OptValue, PathAndArgs},
     credential::process::CredentialProcessCredential,
     credential::token::TokenCredential,
 };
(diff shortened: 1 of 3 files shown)


src/cargo/util/auth/mod.rs
```diff
@@ -88,17 +88,29 @@ fn credential_provider(config: &Config, sid: &SourceId) -> CargoResult<Vec<Vec<S
...
-                .rev()
+            // Cargo normally merges configuration from environment variables at the end of a list,
+            // but here we want to the global credential provider attempted first.
+            let env_iter = config.get_env(&env_key).ok().map(|val| Value {
+                val,
+                definition: Definition::Environment(env_key),
+            });
+            let cfg_iter = p
+                .iter()
+                .filter(|p| !matches!(p.definition, Definition::Environment(_)));
```
weihanglo, line 105: This doesn't look correct, no?  `cargo --config` takes precedence over other stuff.

weihanglo (commented):
Thanks for this. I was too lazy to test it with `cargo config get` 😆. How does the merged array look like via `cargo +nightly -Zunstable-options config get registry.global-credential-providers`?

src/cargo/util/auth/mod.rs
arlosi: I'd forgotten about `--config`. After adding tests, I found the current behavior to be even more unexpected, so I filed #12506

This PR will probably need to wait until we resolve that.

weihanglo: Oh no. #12506 doesn't looks too nice…

Issue #12403, opened 2023-07-27
Forwarding a build.rs env var and running `cargo build` within the binary causes rebuilds


rukai:
### Problem

In some of my projects I do the following:
* forward the [env vars from build.rs](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts) up to my actual application
* invoke `cargo build` from a test or example, this is useful for certain integration tests.

However the combination of these things is causing my application to be rebuilt.

### Steps

build.rs:
```rust
fn main() {
    let profile = std::env::var("PROFILE").unwrap();
    println!("cargo:rustc-env=PROFILE={profile}");
    println!("cargo:rerun-if-env-changed=PROFILE");
}
```
main.rs:
```rust
fn main() {
    let output = std::process::Command::new(env!("CARGO"))
        .args(["build", "-vv"])
        .output()
        .unwrap();
    let stdout = String::from_utf8(output.stdout).unwrap();
    let stderr = String::from_utf8(output.stderr).unwrap();
    println!("stdout:\n{stdout}\nstderr:\n{stderr}");
}
```

Then run `cargo run -vv` to get the following output every time it reruns:
```
~2/Projects/Crates/foo_bin> cargo run -vv
       Dirty foo v0.1.0 (/home/rukai2/Projects/Crates/foo_bin): the env variable PROFILE changed
   Compiling foo v0.1.0 (/home/rukai2/Projects/Crates/foo_bin)
     Running `/home/rukai2/Projects/Crates/foo_bin/target/debug/build/foo-9cd20279bf15ecca/build-script-build`
[foo 0.1.0] cargo:rustc-env=PROFILE=debug
[foo 0.1.0] cargo:rerun-if-env-changed=PROFILE
[...]

PR #12482, opened 2023-08-12
docs: make the env var source of rerun-if-env-changed clearer

rukai:
closes https://github.com/rust-lang/cargo/issues/12403

The wording of the documentation on [rerun-if-env-changed](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-env-changedname) was misleading which led me to be confused by the behavior of cargo and then raise https://github.com/rust-lang/cargo/issues/12403
These changes should avoid users being confused by this in the future.


weihanglo (approved):

weihanglo:
Nice. Thank you for clarifying this tiny but confusing part!

@bors r+

PR #12487, opened 2023-08-12
Contrib: Add process for security responses.

ehuss:
This adds some documentation to give some guidance and checklists for how a security issue is handled.

### DIFF
--- a/src/doc/contrib/src/process/security.md
+++ b/src/doc/contrib/src/process/security.md
@@ -1,0 +1,136 @@
+# Security issues
+
+Issues involving reporting a security vulnerability in cargo usually start by following the [Rust security policy].
+The Security Response Working Group ("the WG") is responsible for running the process of handling the response to a security issue.
+Their process is documented at [Handling Reports].
+This document gives an overview of the process from a Cargo team member's perspective.
+
+The general order of events happens as follows:
+
+1. The "reporter" (even if it is a Cargo team member) reports an issue to <security@rust-lang.org>.
+1. The WG will evaluate if the report is credible, and manages responses to the reporter.
+1. The WG will start a private Zulip stream to coordinate discussion and plans for a fix.
+1. The WG will pull in one or more team members into the Zulip stream ("responders").
+    - Security vulnerabilities are **embargoed** until they are released publicly.
+      People who are brought into these discussions should **not** discuss the issue with *anyone* outside of the group, or with your employer, without first consulting The WG.
+1. A discussion then starts to evaluate the severity of the issue and what possible solutions should be considered.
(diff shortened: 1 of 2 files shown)


ehuss:
cc @rust-lang/security, would love if you have any feedback.

My brain turned to mush while writing this, so I imagine there are things missing or confusing.

src/doc/contrib/src/process/security.md
```diff
@@ -0,0 +1,136 @@
...
+This document gives an overview of the process from a Cargo team member's perspective.
+
+The general order of events happens as follows:
+
+1. The "reporter" (even if it is a Cargo team member) reports an issue to <security@rust-lang.org>.
+1. The WG will evaluate if the report is credible, and manages responses to the reporter.
+1. The WG will start a private Zulip stream to coordinate discussion and plans for a fix.
+1. The WG will pull in one or more team members into the Zulip stream ("responders").
+    - Security vulnerabilities are **embargoed** until they are released publicly.
+      People who are brought into these discussions should **not** discuss the issue with *anyone* outside of the group, or with your employer, without first consulting The WG.
```
epage, line 15: nit "or with your employer" -> "including your employer" since they are covered by the first statement, this is less an "or" and more a clarification

epage:
Thanks for writing this up!  From my limited perspective, it looked good to go

weihanglo (commented):
Thanks for this. I love the write-up!

I am a bit worried about it getting out of sync with the actual procedure of security WG.Should this be documented in security WG's readme or elsewhere? Do they have any public docs for this purpose? I would be better if we doc only the part relevant to cargo, and link to their own doc.

Manishearth (approved):

ehuss:
> I am a bit worried about it getting out of sync with the actual procedure of security WG.Should this be documented in security WG's readme or elsewhere? Do they have any public docs for this purpose? I would be better if we doc only the part relevant to cargo, and link to their own doc.

Their public docs are linked at the beginning of this doc (https://github.com/rust-lang/wg-security-response/blob/main/docs/handling-reports.md). Theirs is oriented towards WG members, and this was oriented specifically for cargo team members.  I agree there is risk that this information becomes stale. There are some unique properties of the cargo project that make this a bit complicated (for example, our branching and release process, version bumping, etc.). I tried to write this as much as "There is a rough outline of what the WG does. This fills in all the in-between bits that are specific to cargo."

weihanglo (approved):

weihanglo:
@bors try

`{{ github.sha }}` should be a merge commit that never touched `cargo-credential`. I have no idea how it ended up there.

weihanglo:
@bors retry

Going to merge this as Manishearth already approved

Issue #9012, opened 2020-12-23
Cargo doesn't respect `--color` ~~or `CARGO_TERM_COLOR`~~ for help text and errors from clap


mchernyavsky:
**Problem**

^title

I expect that with `--color=always` option (or `CARGO_TERM_COLOR=always` environment variable) the output will contain expected ANSI escape codes.

**Steps**
1. Run `cargo build --color=always --unknown &> out.txt`
2. The`out.txt` will contain the following output (without ANSI escape codes):

```
error: Found argument '--unknown' which wasn't expected, or isn't valid in this context

USAGE:
    cargo build --color <WHEN>

For more information try --help

```

**Notes**

Output of `cargo version`:

```
cargo 1.50.0-nightly (a3c2627fb 2020-12-14)
```

epage:
The main problem here is the circular nature of this.
- To read `--color=always`, we need to parse the CLI but the error is coming from our CLI parser
- To read `CARGO_TERM_COLOR`, we need the `Config` configured which requires the CLI to be parsed.

We can't workaround it for `--color=always` but we can for `CARGO_TERM_COLOR` by directly looking up the value and passing that to clap.  We'd probably want to extract the value parser out of `set_color_choice` into a `std::str::FromStr` impl on `ColorChoice`.

This would make it so `fn cli()` would have something like
```rust
    let color_choice = std::env::var_os("CARGO_TERM_COLOR");
    let color_choice = color_choice.and_then(|c| c.parse::<ColorChoice>().ok());
    let color_choice = match color_choice {
        Some(ColorChoice::CargoAuto) | None => clap::ColorChoice::Auto,
        Some(ColorChoice::Always) => clap::ColorChoice::Always,
        Some(ColorChoice::Never) => clap::ColorChoice::Never,
    };
```
This intentionally ignores the errors since it isn't worth reporting at this stage.

The main question is if this is worth it or not.

epage:
#13463 resolves the config side of this but leaves `--color` alone.

To handle `--color`, we'd need to be able to parse the command line to get those which requires knowing what level of styling to use.  We could say "parse the command line, take the result, and strip the message based on our own decision".  However, `--help` and errors short-circuits CLI parsing.  We'd have to parse, see if an error gets back, and then re-parse using the "ignore errors" setting which is a very brittle feature in clap.

I'd propose that we reject `--color` and close this as "resolved" with the merge of #13463

weihanglo:
[...]

PR #12488, opened 2023-08-13
fix(log): enable ansi color only in terminal

weihanglo:

In #12458 we introduced `tracing` but forgot to consult if it is in tty to enable color logs.

`cargo tree &> log` and check if `log` file contains any ansi color code.

### Additional information

I'll hold off `CARGO_LOG` respecting `--color=always` for now.

<https://github.com/rust-lang/cargo/issues/9012> is something can be resolved as a whole. [`tracing_subscriber::reload`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/reload/index.html) is also a thing to consider, so that we can configure logging color and output file after loading `config.toml`


weihanglo:
IIUC, this is aligned with how `env_logger` handles ASNI color codes: <https://github.com/rust-cli/env_logger/blob/bdc25caee0c27fbb1c9c66d6492a5910aaf3ac1d/src/fmt/writer/atty.rs>

weihanglo:
Currently blocked on <https://github.com/rust-lang/rust/issues/114804>.

While we can switch from main.rs to an empty lib.rs to bypass that, I do think this is a good test to catch this kind of subtle change.

```diff
diff --git a/tests/testsuite/future_incompat_report.rs b/tests/testsuite/future_incompat_report.rs
index 9f451a64c..4d2c66d17 100644
--- a/tests/testsuite/future_incompat_report.rs
+++ b/tests/testsuite/future_incompat_report.rs
@@ -164,7 +164,7 @@ fn test_multi_crate() {
                 second-dep = "*"
               "#,
         )
-        .file("src/main.rs", "fn main() {}")
+        .file("src/lib.rs", "")
         .build();
 
     for command in &["build", "check", "rustc", "test"] {
```

epage (approved):
r= me when this is able to go through

weihanglo:
Thank you for the review!

@bors r+

Issue #8693, opened 2020-09-09
rerun-if-env-changed=CARGO_MANIFEST_DIR doesn't trigger a rebuild on change


alecmocatta:
`CARGO_MANIFEST_DIR` changing doesn't trigger a rebuild, even if `rerun-if-env-changed=CARGO_MANIFEST_DIR` is specified in the `build.rs`.

Thus `env!("CARGO_MANIFEST_DIR")` and `env::var("CARGO_MANIFEST_DIR").unwrap()` can differ in binaries executed by `cargo run/test/bench`.

Mark-Simulacrum:
I personally am less certain that this should be considered a bug without the rerun-if-env-changed, as most crates don't want to rebuild when moving their directory around. Indeed, e.g., perf.rust-lang.org relies on that currently...

alecmocatta:
IMO the least surprising behaviour is triggering a rebuild if the value returned by any `env!` invocation in a crate would change.

> as most crates don't want to rebuild when moving their directory around

I'm envisaging this would only happen if said crate had `env!("CARGO_MANIFEST_DIR")`. Which seems reasonable to me? If someone doesn't want this behaviour they can change to `env::var("CARGO_MANIFEST_DIR").unwrap()`.

alexcrichton:
Cargo does explicilty not rebuild a crate when it's folder is renamed, but it's true that the compilation would otherwise differ due to environment variables like this. What we can do, however, is Cargo can avoid recompiling on a directry name if this env var isn't used, but it would recompile if it was used. I think that would keep everything working generally as expected for other users which currently rely on this behavior not recompiling.

Veetaha:
I've just been faced with a problem here. I think the issue should be named more clearly as "Build script isn't rebuild if CARGO_MANIFEST_DIR compile-time variable changes, i.e. the source dir was moved".

This bug reproduces in my build script depends on a crate that uses `env!("CARGO_MANIFEST_DIR")` to reference sources from its source folder (inside of `.cargo/registry/src`) with protobuf files. When I mount the sources under a different absolute path in a docker container and mount the target dir cache I see that build fails because the build script wasn't recompiled with a different `CARGO_MANIFEST_DIR` value for the library that the build script depends on, and thus the script can't find the protobuf files it needs.

weihanglo:
[...]

PR #12489, opened 2023-08-14
docs: further tweak rerun-if-env-changed docs

rukai:
Continuation of https://github.com/rust-lang/cargo/issues/12403 + https://github.com/rust-lang/cargo/pull/12482

I thought about this some more and I think its important we mention this as well.
Otherwise the user may think it is impossible to use the env vars set by cargo in the build.rs without incorrect behavior when those values change.

r? weihanglo

### DIFF
--- a/src/doc/src/reference/build-scripts.md
+++ b/src/doc/src/reference/build-scripts.md
@@ -349,8 +349,9 @@ Note that the environment variables here are intended for global environment
 variables like `CC` and such, it is not possible to use this for environment
 variables like `TARGET` that [Cargo sets for build scripts][build-env]. The
 environment variables in use are those received by `cargo` invocations, not
-those received by the executable of the build script.
-
+those received by the executable of the build script. Due to Cargo's
+internal cache invalidation Cargo will effectively re-run the build script
+every time one of the variables that Cargo sets changes.
 
 ### The `links` Manifest Key
 


src/doc/src/reference/build-scripts.md
```diff
@@ -350,7 +350,8 @@ variables like `CC` and such, it is not possible to use this for environment
 variables like `TARGET` that [Cargo sets for build scripts][build-env]. The
 environment variables in use are those received by `cargo` invocations, not
 those received by the executable of the build script.
-
+Instead cargo will automatically re-run the build script if the value of the environment
```
weihanglo, line 353: Thanks for the effort. However, Cargo internally doesn't really monitor those environment variables directly. Cargo has its own mechanism to handle those envs. See [Fingerprints and Metadata](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/compiler/fingerprint/index.html#fingerprints-and-metadata). I may avoid adding this, or find a better way to doc the actual stuff being tracked by Cargo in general elsewhere.

rukai: oh interesting, so its likely that for some of the environment variables that cargo sets we would need to do something hacky like write the contents to a file and then set rerun-if-changed for the file?

weihanglo: No. You usually don't need to do that — Cargo tracks them for you.

If you see anything should be tracked but isn't, that could be another recompile detection issue.

weihanglo: > Cargo internally doesn't really monitor those environment variables directly

By this I mean Cargo doesn't track what it has set directly, but if you set something directly to the invocation, like `PROFILE=something cargo build`, it'll be tracked as a plain old user-specified environment variable. There's nothing special.

rukai: I tweaked the wording to be a bit more accurate from your description.

My concern is that a user may read the docs as they currently are and think: "But if I cant use `rerun-if-env-changed` on variables set by cargo then I shouldnt use variables set by cargo since the cached build.rs results will go stale."
Thats why I think its important to mention that cargo handles this for us.

weihanglo: Feel like 9b89ff7563789ffa1d37655ca1110254b81e73c0 went way too far. It's my fault not get it clearer, but still felt the current patch complicates stuff.

> _Cargo will effectively re-run the build script every time one of the variables that Cargo sets changes._

After reading that statement, people may do this, observe nothing rerun and complain the doc is lying:

```shell
# User: hmm… cargo takes care of this for me.
# User: I dont need to add rerun-if-env-changed=NUM_JOBS
NUM_JOBS=1 cargo b # full build
NUM_JOBS=2 cargo b # no rebuild
```

Like I said earlier, “usually” you don't need to pay attention to those variables. Some of those are different, and `CARGO_MANIFEST_DIR` is one of them not triggering a rebuild (see #8693). I think people should generally trust the rebuild detection in Cargo. When they find an issue, they report one like #8693.

As in #12403, you discovered rebuild too eagerly, not in reverse. Personally I don't feel like it worth putting the statement here. Maybe there is a better explanation but I cannot think of one now.

rukai: Alright, it looks like you think a larger rework is required than just tweaks to this paragraph.
I'll go ahead and close this.
Thanks for your time.

Issue #4879, opened 2017-12-30
cargo run -v should show setting of LD_LIBRARY_PATH


joshtriplett:
I spoke with someone recently who had some trouble figuring out how to run Cargo-built binaries directly, until they realized that `cargo run` sets `LD_LIBRARY_PATH`, which is not documented. When running `cargo run -v`, cargo shows the binaries it runs, but not the setting of `LD_LIBRARY_PATH`. Please include that in the emitted command line invocations (e.g. `LD_LIBRARY_PATH=... target/release/...`) when using `-v`, to make it more obvious what `cargo run` does.

ehuss:
`LD_LIBRARY_PATH` behavior is now documented at https://doc.rust-lang.org/cargo/reference/environment-variables.html#dynamic-library-paths.

`cargo run -vv` will show all environment variables, including `LD_LIBRARY_PATH`.

I think it would be good to leave this open for a more programmatic way to get the appropriate paths (some kind of JSON output or something).

sfackler:
`cargo run -vv` only appears to print full command lines for calls to rustc, but the actual run command is still just `Running 'target/debug/foo'`. Did this regress?

ehuss:
Oh, I don't think it ever showed the env vars for the actual *running* of the executable.  I think I was just meaning that the calls to `rustc` will show it.  They are slightly different (between running the exe and building it), so it would probably be good to display it somehow.

Kobzol:
This makes it difficult to debug Rust binaries that use native dependencies and that require `LD_LIBRARY_PATH` to be set. Without `cargo run`, the libraries are not found, and I don't know to which directory in `target` should I point the environment variable. And trying to debug `cargo run` with `gdb` doesn't seem to work for me, the program is constantly being stopped on some SIGUSR1 signal (it can be [ignored](https://stackoverflow.com/questions/24999208/how-to-handle-all-signals-in-gdb), but still).

It would be nice if `cargo run -vv` actually emitted the environment flags. I'd like to try to implement it, if you consider it to be useful (and maybe suggest some mentoring tips :) ).

weihanglo:
I believe that is feasible to be done.

Basically, search for `"Running"` string and check if  it is really `config.extra_verbose()`, then set `ProcessBuilder::display_env_vars()` for the command builder around.

Guess not only `cargo run -vv` needs that, `cargo test -vv` could have the same mechanism as well.

---

[...]

PR #12498, opened 2023-08-14
Print environment variables for `cargo run` in extra verbose mode

Kobzol:

This PR changes `cargo run` so that it prints the environment variables that it sets for the executed binary in extra verbose mode (`cargo run -vv`). This is particularly useful when it sets `LD_LIBRARY_PATH` which is needed to actually run the binary. Because if users want to run the binary without `cargo`, they will need to know where should they point that variable. This is described in issue https://github.com/rust-lang/cargo/issues/4879.

Fixes: https://github.com/rust-lang/cargo/issues/4879

I tried to write a functional test, but I'm not sure how to check that `stderr` contains only a given substring, not a whole line.

### DIFF
--- a/tests/testsuite/run.rs
+++ b/tests/testsuite/run.rs
@@ -1,6 +1,8 @@
 //! Tests for the `cargo run` command.
 
-use cargo_test_support::{basic_bin_manifest, basic_lib_manifest, project, Project};
+use cargo_test_support::{
+    basic_bin_manifest, basic_lib_manifest, basic_manifest, project, Project,
+};
 use cargo_util::paths::dylib_path_envvar;
 
 #[cargo_test]
@@ -1414,6 +1416,24 @@ fn default_run_workspace() {
         .build();
 
     p.cargo("run").with_stdout("run-a").run();
+}
+
+#[cargo_test]
+fn print_env_verbose() {
+    let p = project()
+        .file("Cargo.toml", &basic_manifest("a", "0.0.1"))
+        .file("src/main.rs", r#"fn main() {println!("run-a");}"#)
+        .build();
+
+    p.cargo("run -vv")
+        .with_stderr(
+            "\
+[COMPILING] a v0.0.1 ([CWD])
+[RUNNING] `[..]CARGO_MANIFEST_DIR=[CWD][..] rustc --crate-name a[..]`
+[FINISHED] dev [unoptimized + debuginfo] target(s) in [..]
+[RUNNING] `[..]CARGO_MANIFEST_DIR=[CWD][..] target/debug/a[EXE]`",
+        )
+        .run();
 }
 
 #[cargo_test]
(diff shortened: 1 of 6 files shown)


tests/testsuite/run.rs
```diff
@@ -1416,6 +1416,23 @@ fn default_run_workspace() {
...
+            r#"
+                [package]
+                name = "a"
+                version = "0.0.1"
+            "#,
+        )
+        .file("src/main.rs", r#"fn main() {println!("run-a");}"#)
+        .build();
+
+    p.cargo("run -vv").with_stderr("LD_LIBRARY_PATH").run();
```
weihanglo, line 1433: This contains two shortcomings:

* [Different platforms have different library paths](https://doc.rust-lang.org/nightly/cargo/reference/environment-variables.html#dynamic-library-paths). This won't pass on Windows and macOS.
* This matches both `rustc` and the final binary invocations. We can't tell whether the env is from which.

What about this? `[..]` for wildcard matches, and we match one env vars to verify the behavior.

```suggestion
    p.cargo("run -vv")
        .with_stderr(
            "\
[COMPILING] a v0.0.1 ([CWD])
[RUNNING] `[..]CARGO_MANIFEST_DIR=[CWD][..] rustc --crate-name a[..]`
[FINISHED] dev [unoptimized + debuginfo] target(s) in [..]
[RUNNING] `[..]CARGO_MANIFEST_DIR=[CWD][..] target/debug/a[EXE]`",
        )
        .run();
```

weihanglo (commented):
Looks good to me. Do you want to deal with `cargo test` as well in this pull request, or a follow-up?

(Can also leave it there. I am fine with that 😁)

tests/testsuite/run.rs
arlosi, line 1433: `LD_LIBRARY_PATH` won't be set on Windows. I'd recommend a different one like `CARGO_PKG_NAME` for the test case. 

You can use `[..]` as a wildcard for matching.

```suggestion
    p.cargo("run -vv")
        .with_stderr_contains("[RUNNING] [..]CARGO_PKG_NAME=a[..]")
        .run();
```

arlosi (changes requested):

tests/testsuite/run.rs
arlosi: @weihanglo beat me to the review with the same comment 😊. I'll close this one.

Kobzol:
Thanks for the hints. I added support for `cargo test` too. I have to say, it would be really nice if `cargo` had auto-blessable tests with placeholders :laughing: Getting the text output right is not straightforward.

weihanglo:
I believe changing to `[RUNNING] `[..] rustdoc [..]--test [..]` would fix CI failures.

Kobzol:
Thanks for the hint! I fixed the two failing tests and used `[DOCTEST] foo` in the `test` test, as I noticed it was used in the failing tests. I also added a test for `bench`.

arlosi:
Thanks!

@bors r+

arlosi (approved):

Issue #8933, opened 2020-12-03
Tracking Issue for credential-process RFC 2730


ehuss:
**Summary**

RFC: https://github.com/rust-lang/rfcs/pull/2730
Implementation: #8934
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#credential-process
Issues: https://github.com/rust-lang/cargo/labels/Z-credential-process

This feature provides a configuration option to specify a process to fetch a token to authenticate with a registry.

**Unresolved issues**

* [X] Is this approach useful enough?  Things like macos keychain don't protect against being executed to extract tokens (and I don't see a way to require a password, or force the process to be untrusted).  The lack of signatures also cause issues (each update of the toolchain will cause it to become untrusted again).
  * Making the built-in providers like `cargo:macos-keychain` part of the cargo binary improves security here, since only the Cargo binary is accessing the Keychain. If we get Cargo signed by Apple, then it would be further improved.
* [X] Should the `login` API be changed? Providers that need to be interactive need to be able to read from stdin.
  * This is worked around by having the credential-provider re-open `stdin` from `/dev/tty` or `$CONIN`
* [X] Should the storage key off the registry name? [RFC 3139](https://github.com/rust-lang/rfcs/pull/3139) discusses problems with this. RESOLVED: credential helpers are now keying off the index url via #10592.
* [x] Should it be possible to pass additional arguments on the command line for `cargo login` that would be sent to the provider? E.g. `cargo login -- --extra-arg-for-provider` 
  * Implemented in #12499
* [X] Figure out the name for `cargo:basic`: `cargo:token-from-stdout`
* [X] Remove cargo-1password as a built-in (add the `fn main` so it can be published separately).
* [x] Create a new `-Z` flag for `cargo:paseto`.
  * [x] Publish `cargo-credential-1password` to crates.io
* [x] Figure out MSRV support. rust-lang/cargo#12381 is proposing to set MSRV for all packages in the repo. I think it will be important that the credential provider libraries have a relatively relaxed requirement (like stable-2 or something). I think this will be important to figure out before stabilization. This may be complicated with CI costs, so I'm wondering if `cargo-credential` could grow a few independent tests on its own so we could test just the library with an older rust? https://github.com/rust-lang/cargo/pull/12623
[...]

PR #12499, opened 2023-08-14
login: allow passing additional args to provider

arlosi:
As part of moving asymmetric token support to a credential provider in #12334, support for passing `--key-subject` to `cargo login` was removed.

This change allows passing additional arguments to credential providers when running `cargo login`. For example:
`cargo login -- --key-subject foo`.

The asymmetric token provider (`cargo:paseto`) is updated to take advantage of this and re-enables setting `--key-subject` from `cargo login`.

r? @Eh2406

cc #8933


Eh2406:
Looks good to me.

ehuss:
Seems reasonable to me.

@bors r=Eh2406

PR #12413, opened 2023-07-29
test: relax assertions of panic handler message format

weihanglo:
### What does this PR try to resolve?

Fixes test failure in https://github.com/rust-lang/rust/pull/112849

### How to test

Built std from <https://github.com/rust-lang/rust/pull/112849> and test against Cargo's test suite in this PR.

PR #12500, opened 2023-08-14
test: relax assertions of panic message (again)

weihanglo:

Fixes test failure in https://github.com/rust-lang/rust/pull/111071

Similar PR in the past: #12413

### How to test

Built std from <https://github.com/rust-lang/rust/pull/111071> and test against Cargo's test suite in this PR.

### DIFF
--- a/tests/testsuite/bench.rs
+++ b/tests/testsuite/bench.rs
@@ -314,9 +314,9 @@ fn cargo_bench_failing_test() {
 [RUNNING] [..] (target/release/deps/foo-[..][EXE])",
         )
         .with_stdout_contains("[..]thread '[..]' panicked at[..]")
-        .with_stdout_contains("[..]assertion failed[..]")
-        .with_stdout_contains("[..]left: `\"hello\"`[..]")
-        .with_stdout_contains("[..]right: `\"nope\"`[..]")
+        .with_stdout_contains("[..]assertion [..]failed[..]")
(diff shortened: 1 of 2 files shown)


tests/testsuite/bench.rs
```diff
@@ -314,9 +314,9 @@ fn cargo_bench_failing_test() {
 [RUNNING] [..] (target/release/deps/foo-[..][EXE])",
         )
         .with_stdout_contains("[..]thread '[..]' panicked at[..]")
-        .with_stdout_contains("[..]assertion failed[..]")
-        .with_stdout_contains("[..]left: `\"hello\"`[..]")
-        .with_stdout_contains("[..]right: `\"nope\"`[..]")
+        .with_stdout_contains("[..]assertion [..]failed[..]")
```
nyurik, line 317: thanks @weihanglo!  Does `[..]` match `.*` or `.+` (in regex terms)?

weihanglo: Yes, similar to `.*` for a line. Please see <https://doc.rust-lang.org/nightly/nightly-rustc/cargo_test_support/compare/index.html>.

I am doing this because it need to match both stable and nightly formats 🥲

epage:
I'd be tempted to say RenovateBot should update our nightly every night but that would get tedious to merge and would flood our commit history

PR #12501, opened 2023-08-14
fix(lints): Doctest extraction should respect `[lints]`

### LINKED ISSUE #12497: Lints table doesn't seem to work with doctests
### Problem

Lints defined in the lints table doesn't seem to work with doctests, whereas lints defined as an attribute do work. This seems inconsistent, as I would expect them to behave the same.

### Steps

Example rust file:

```rust
/// Test
///
/// [`Foo`]
///
/// ```
/// let s = "rust";
/// let ｓ_ｓ = "rust2";
/// ```
pub fn f() {}
pub const Ě: i32 = 1;
pub const Ĕ: i32 = 2;
```

With manifest:

```
[package]
name = "foo"
version = "0.1.0"
edition = "2021"

[dependencies]

[lints.rust]
confusable-idents = 'allow'
```

Running `cargo check -Zlints` produces no warnings:
```
    Checking foo v0.1.0 (/Users/eric/Temp/foo)
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
```

Running `cargo test --doc -Zlints` produces warnings:
```
   Compiling foo v0.1.0 (/Users/eric/Temp/foo)
    Finished test [unoptimized + debuginfo] target(s) in 0.02s
   Doc-tests foo
warning: found both `Ě` and `Ĕ` as identifiers, which look alike
  --> src/lib.rs:11:11
   |
10 | pub const Ě: i32 = 1;
   |           - other identifier used here
11 | pub const Ĕ: i32 = 2;
   |           ^ this identifier can be confused with `Ě`
   |
   = note: `#[warn(confusable_idents)]` on by default

warning: 1 warning emitted


running 1 test
test src/lib.rs - f (line 5) ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
```

If you put `#![allow(confusable_idents)]` at the top of the file, then no warning is generated.

### Possible Solution(s)

I think lint flags should be passed to rustdoc [here](https://github.com/rust-lang/cargo/blob/f13759445fbcb9cfea241f86690e6d833c2a8b34/src/cargo/ops/cargo_test.rs#L209) when building the rustdoc comand line. I'm not sure if there was a reason this wasn't added.

### Notes

_No response_

### Version

```text
cargo 1.73.0-nightly (d78bbf4bd 2023-08-03)
release: 1.73.0-nightly
commit-hash: d78bbf4bde3c6b95caca7512f537c6f9721426ff
commit-date: 2023-08-03
host: aarch64-apple-darwin
libcurl: 7.88.1 (sys:0.4.65+curl-8.2.1 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 13.4.1 [64-bit]
```

epage:
Note the first commit shows the bug and the second commit fixes it

Fixes #12497

### DIFF
--- a/src/cargo/ops/cargo_test.rs
+++ b/src/cargo/ops/cargo_test.rs
@@ -260,6 +260,8 @@ fn run_doc_tests(
             p.arg("--test-args").arg("--quiet");
         }
 
+        p.args(unit.pkg.manifest().lint_rustflags());
+
         p.args(args);
 
         if *unstable_opts {
--- a/tests/testsuite/lints.rs
+++ b/tests/testsuite/lints.rs
@@ -637,3 +637,70 @@ error: unresolved link to `bar`
         )
         .run();
 }
+
+#[cargo_test]
+fn doctest_respects_lints() {
+    let foo = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.0.1"
+                authors = []
+
+                [lints.rust]
+                confusable-idents = 'allow'
+            "#,
+        )
+        .file(
+            "src/lib.rs",
+            r#"
+/// Test
+///
+/// [`Foo`]
+///
+/// ```
+/// let s = "rust";
+/// let ｓ_ｓ = "rust2";
+/// ```
+pub fn f() {}
+pub const Ě: i32 = 1;
+pub const Ĕ: i32 = 2;
+"#,
+        )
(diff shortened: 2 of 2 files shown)


weihanglo (approved):
doctests always looks exotic as it constructs all rustflags in `ops` module.

Anyway, thank you all!

PR #12507, opened 2023-08-15
credential-providers: make 1password no longer built-in

arlosi:
Since 1password is just one of many potential CLI-based credential providers, it makes more sense for it to be installable as a plugin rather than built-in to Cargo.

This means that `cargo:1password` will no longer work as a credential provider. 

The replacement would be `cargo install cargo-credential-1password` and using `cargo-credential-1password` instead.

r? @ehuss

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -263,7 +263,6 @@ dependencies = [
  "base64",
  "bytesize",
  "cargo-credential",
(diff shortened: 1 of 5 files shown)


arlosi:
I don't have a 1password subscription, so this is difficult for me to test.

PR #12508, opened 2023-08-15
ci: use pull request head commit whenever possible

weihanglo:
The `github.sha` is a merge commit with the parents of latest master and the head of the pr. Trying to diff the changes from that merge SHA to base will show all changes that have been made in-between.

And that doesn't seem about right.

We switch to `github.event.pull_request.head.sha` if there is a pr.

See: https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/version-bump.20CI.20failing


ehuss:
Thanks!

@bors r+

Issue #12254, opened 2023-06-12
cargo vendor modifies the package cache without locking


Amanieu:
We are getting occasional, but infrequent, failures in our CI which uses `cargo vendor`. Looking into it, it seems that this could be due to https://github.com/alexcrichton/cargo-vendor/pull/131, where `cargo vendor` will delete downloaded crates in the package cache, even though those crates could be concurrently used by other build processes.

This is done without holding the package cache lock, but even if it did I'm not sure this is correct since other cargo commands will read from the package cache (e.g. `cargo build` will read the code from extracted crates) without holding the package cache lock.

weihanglo:
I think your observation is correct. This line does delete sources from cargo glocal package source cache.

https://github.com/rust-lang/cargo/blob/928b9561c16e73fca6206b42e4c464b80b0e9221/src/cargo/ops/vendor.rs#L138

> This is done without holding the package cache lock, but even if it did I'm not sure this is correct since other cargo commands will read from the package cache (e.g. `cargo build` will read the code from extracted crates) without holding the package cache lock.

I think the suggested approach is pretty much doable, and my impression is that during the dependency resolution `cargo build` also need to acquire package-cache lock. May be as easy as [putting a line above here](https://github.com/rust-lang/cargo/blob/56bb8c49db1b99fed671700b48ffa696d3a9655f/src/cargo/ops/vendor.rs#L116) to acquire the package cache lock. Just the heavy-hammer approach for removing-then-redownloading is still a bit uncomfortable to me in terms of performance.

---

triage:

@rustbot label +C-bug +Command-vendor +S-accepted +E-easy

weihanglo:
`cargo build` is already acquiring the package cache lock if I rememeber correctly. It isn't? 🤔

Amanieu:
Only for dependency resolution and downloading crates, not during the actual build itself. However it still needs to read the crate source code from the package cache during the build.

weihanglo:
> However it still needs to read the crate source code from the package cache during the build.

That's what I was saying. `cargo build` gets the lock for reading.

> I think the best way forward might be to just revert https://github.com/alexcrichton/cargo-vendor/pull/131.

[...]

Issue #5956, opened 2018-08-31
Cargo and/or other tools add Cargo.lock files and other build artifacts to registry


mykmelez:
As described in [bug 1461553, comment 2](https://bugzilla.mozilla.org/show_bug.cgi?id=1461553#c2) and later comments, Cargo and/or other tools that are using Cargo's global registry sometimes add Cargo.lock files to the crate directories (i.e. unpacked tarballs) in the registry. And they've been known to add other files as well, f.e. the [lalrpop](https://github.com/lalrpop/lalrpop) crate directory sometimes sprouts the file src/parser/lrgrammar.rs.

Cargo.lock and src/parser/lrgrammar.rs are both build artifacts, so presumably this is caused by a tool that is building the crate directly within the registry directory. Unfortunately, Cargo and tools like cargo-vendor assume that the registry is immutable, so these added files cause unwanted effects, like unexpected changes to already-vendored crates when vendoring unrelated ones (or even simply re-running cargo-vendor after adding no new crates).

Over in that bug, @alexcrichton notes a few options for resolving the issue:

> One change would be to have cargo-vendor work directly from the tarballs instead of the checked-out state, but such a change isn't entirely trivial to implement today. Otherwise Cargo could also do something like periodically execute a sanity check to make sure checked out crates look alright.
> …
> I'd personally prefer to dig in and discover what crates are modifying the registry. I suspect it's actually very few but perhaps heavily depended on ones that have consequences for the ecosystem.

In the meantime, a workaround is to delete the registry (`rm -rf ~/.cargo/registry`), which Cargo will then repopulate as needed.

epage:
The reasond packages may be doing this
- `build.rs` files that don't follow the rules of writing to `OUT_DIR`
- `build.rs` running cargo commands within the PACKAGE_DIR to do various forms of introspection.

A combination of the following would fix this
- Sandboxing (#5720)
- Never extracting `.crate` files (https://internals.rust-lang.org/t/jar-for-rust-single-file-crate-support-for-rustc/19576)

However, as the underlying need was resolved in cargo-vendor#131, I'm going to go ahead and close this.

weihanglo:
The other relevant issue is <https://github.com/rust-lang/cargo/issues/9455> for making unpacked tarballs and directories "read-only".

PR #12509, opened 2023-08-16
Hold the mutate exclusive lock when vendoring

hi-rustin:

close https://github.com/rust-lang/cargo/issues/12254

Held the mutate exclusive lock when vendoring. This would block all other processes which want to read or change the unpacked files.

See: https://github.com/rust-lang/cargo/pull/12509#issuecomment-1732415990

r? @weihanglo

### DIFF
--- a/src/cargo/ops/vendor.rs
+++ b/src/cargo/ops/vendor.rs
@@ -4,6 +4,7 @@ use crate::core::{GitReference, Package, Workspace};
 use crate::ops;
 use crate::sources::path::PathSource;
 use crate::sources::CRATES_IO_REGISTRY;
+use crate::util::cache_lock::CacheLockMode;
 use crate::util::{try_canonicalize, CargoResult, Config};
 use anyhow::{bail, Context as _};
 use cargo_util::{paths, Sha256};
@@ -31,6 +32,7 @@ pub fn vendor(ws: &Workspace<'_>, opts: &VendorOptions<'_>) -> CargoResult<()> {
         extra_workspaces.push(ws);
     }
     let workspaces = extra_workspaces.iter().chain(Some(ws)).collect::<Vec<_>>();
+    let _lock = config.acquire_package_cache_lock(CacheLockMode::MutateExclusive)?;
     let vendor_config = sync(config, &workspaces, opts).with_context(|| "failed to sync")?;
 
     if config.shell().verbosity() != Verbosity::Quiet {


ehuss:
I assume this is related to #12254?

weihanglo:
Thanks and sorry I didn't speak it more clearly. I would expect to see acquiring a lock there first. If somebody can working on digging into the issue deeper, we can perhaps remove this piece of code when we feel safe.

Thinking a bit more on the integrity of `cargo-vendor`, re-unpacking source looks like a safer solution before we can find a way to resolve <https://github.com/rust-lang/cargo/issues/9455>. We never know if some other tools changes unpacked registry sources, which is already broken when you run `cargo build` today, unfortunately 😢.

ehuss:
> I would expect to see acquiring a lock there first.

Just FYI, I don't think the existing locking would help here, since the problem is with the build process which does not hold a lock. I am currently reworking locking for the garbage collection support, which adds read locks to the build process. It might be usable for this situation.

I'm also curious, the other alternative from the original issue was to read from the `.crate` files instead of the src cache. I'm wondering how difficult that would be to support?

ehuss:
> Do you mean we need to unpack it again from .cargo/registry/cache/xxx/some-pkg.crate?

Yea, to have `cargo vendor` directly read from the `.crate` and extract it into the vendor folder instead of reading from the `src` cache.  I'm not sure what the API for that would look like, I'm just curious what kind of options there are or how difficult that would be to do.

hi-rustin:
> I'm not sure what the API for that would look like, I'm just curious what kind of options there are or how difficult that would be to do.

I think it can be done. The tricky part is that we have to calculate the checksum for each file we want to copy. So we may need another `unpack_package` function with filters and checksum calculation.

But there is an easy way to do it, we can unpack it to a temp folder first and then execute the original logic. But I think this has a big impact on performance.

hi-rustin:
> I think it can be done. The tricky part is that we have to calculate the checksum for each file we want to copy. So we may need another `unpack_package` function with filters and checksum calculation.

@ehuss Do you think this is a reasonable way to move forward? I am happy to give it a try.

ehuss:
I'm looking over the vendor code, and I think it looks like it would be quite challenging to directly extract from the `.crate` files. There would need to be two variants of `cp_sources`, one that reads from a source directory (like git dependencies) and one from `.crate` files. It also seems complex, since there are multiple workspaces and multiple sets of `Source`s (a potentially different set from each workspace). It could possibly build a union of all the sources, and then ask the appropriate `Source` to handle extracting into the target directory. That seems like it could be a pretty large change to me, though, and quite hard to avoid duplicating some of the code.

#12706 is introducing the cache locking support. That might be a much easier option, since it would be a roughly one line change to acquire an exclusive lock. It's still unpleasant, since this `remove_dir_all` stuff has a fairly significant performance penalty, and it would be nice to remove it.

Unfortunately I don't think I have the time to dig into exactly how to architect a design that would support extracting `.crate` files directly. Perhaps @weihanglo can help if you want to continue down that path? It would be nice (due to the significant performance improvement), but it seems like it would require a lot of changes.

weihanglo:
r? weihanglo

Would find a time looking into it and discuss with hi-rustin :)

hi-rustin:
> #12706 is introducing the cache locking support. That might be a much easier option, since it would be a roughly one line change to acquire an exclusive lock. It's still unpleasant, since this `remove_dir_all` stuff has a fairly significant performance penalty, and it would be nice to remove it.

I've added the exclusive lock now. But it is still very difficult to remove `remove_dir_all` stuff because we don't know the reason of https://github.com/rust-lang/cargo/issues/5956. I don't think it's safe to remove this code unless we force users not to change the unpacked file in the src directory.

@weihanglo @ehuss Should we move forward with this PR? Or do you have any other thoughts?

ehuss:
I'm fine with moving forward with this. I agree that it seems like removing `remove_dir_all` will be quite challenging.

weihanglo (approved):
This works as expected. Thank you.

Still not pleasant, as now cargo hold the mutate exclusive lock for the entire duration of `cargo vendor`.

I wonder if there is a way to “downgrade” from `MutateExclusive` to `DownloadExclusive` so that cargo doesn't need to hold the lock that long.

PR #12511, opened 2023-08-16
fix(crates.io): Decorate add_owners error for missing team responses

### LINKED ISSUE #11505: "cargo owner --add" doesn't seem to work with teams
### Problem

We've tried a few different ways of adding a team to grant the publishing rights to Nushell's repo, but haven't been able for this to work.

Here's what we're seeing:

```
〉 cargo owner --add github:nushell:publishing
error: failed to invite owners to crate `nu` on registry at https://crates.io

Caused by:
  the remote server responded with an error: could not find the github team nushell/publishing
```

### Steps

1. see above

### Possible Solution(s)

It'd be nice to perhaps give a better error to help add a team if there's an issue with adding the specific team. Otherwise, I think this should work based on the name of the org + team.

### Notes

_No response_

### Version

```text
cargo 1.66.0 (d65d197ad 2022-11-15)
release: 1.66.0
commit-hash: d65d197ad5c6c09234369f219f943e291d4f04b9
commit-date: 2022-11-15
host: aarch64-apple-darwin
libcurl: 7.84.0 (sys:0.4.55+curl-7.83.1 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 13.0.1 [64-bit]
```

calavera:

GitHub might return a generic error if the user has not configured their permissions correctly to access a team. In this case, the `add_owners` function decorates the response error to include a link to the documentation.

Fixes #11505

I didn't see any integration test that exercises this code, but let me know if I missed where the crates api is tested.
This is what I did to test the code:

1. Compile cargo
2. Try to add a team that doesn't exist in GitHub to a crate that you have access to.

### DIFF
--- a/crates/crates-io/lib.rs
+++ b/crates/crates-io/lib.rs
@@ -132,7 +132,7 @@ pub enum Error {
     #[error(transparent)]
     Curl(#[from] curl::Error),
 
-    /// Error from seriailzing the request payload and deserialzing the
+    /// Error from serializing the request payload and deserializing the
     /// response body (like response body didn't match expected structure).
     #[error(transparent)]
     Json(#[from] serde_json::Error),
@@ -181,6 +181,15 @@ pub enum Error {
          Total size was {0}."
     )]
     Timeout(u64),
+
+    /// Error from the API when the user cannot access a GitHub team.
+    #[error(
+        "{}. \
+        Make sure that you have the right permissions in GitHub: \
+        https://doc.rust-lang.org/cargo/reference/publishing.html#github-permissions",
+        errors.join(", ")
+    )]
+    MissingOwners { errors: Vec<String> },
 }
 
 impl Registry {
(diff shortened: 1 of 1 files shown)


arlosi:
Thanks for this! 

I wonder if this could be done on crates.io side instead. Then we wouldn't need to special-case this in Cargo, and older versions of Cargo would also get the improved error message.

[Here's where the error is coming from](https://github.com/rust-lang/crates.io/blob/7e376fc3311facfb89ffceb83d9f2d7f74d52a4a/src/models/team.rs#L146C30-L146C30) in crates.io.

weihanglo (commented):
Oh ya agree with arlosi on in that regard. I didn't though of that.

Actually in #11505 I was expecting to see documentation improvement with some screenshots, as JT described. The error message enhancement is just a nice-to-have.

Thanks for your contribution anyway!

arlosi:
> I don't know how older versions of Cargo display this. It might be fine, or it might break it. I would have to test with all versions of Cargo.

I agree, it would be a good idea verify that this didn't break old versions. I don't think we'd need to test every release, however.

> I don't know which other clients might consume this API. I don't want to break all expectations for other crates.io clients.

I feel like the change would be acceptable. It's just changing the error message text. Cargo is the only supported client of crates.io. Of course, we'd need to get the crates.io team approval. I don't speak for them.

> The documentation link is specific to Cargo. I don't know if there is a precedence of crates.io returning links to the Cargo docs, but it doesn't sound right to me.

Pretty much all of the crates.io documentation is in the Cargo docs. Go to https://crates.io and click "Getting Started" and you're taken to Cargo docs.

arlosi:
@calavera the implementation in the PR looks good, but I think I'd rather this error came from crates.io, rather than making a special case for it here in Cargo.

Would you be interested in making the documentation updates that @weihanglo suggested here? Or possibly even making a PR to crates.io to improve the error message there?

PR #12513, opened 2023-08-16
fix(xtask-bump-check): query by package name to detect changes

weihanglo:

versions and paths of a workspace members between the original and
a checked-out workspace are different, and shouldn't be included in
hash keys when querying packages.

The second commit demonstrates the detection.


epage (approved):
Feel free to r= me

PR #12515, opened 2023-08-16
config: merge lists in precedence order

### LINKED ISSUE #8128: Cargo concatenates `rustflags` from `.cargo/configs`s in reversed order
Consider this setup:
```
outer
├── .cargo
│   └── config
└── inner
    ├── .cargo
    │   └── config
    ├── Cargo.toml
    └── src
        └── main.rs
```
Less specific `outer/.cargo/config` contains
```toml
[build]
rustflags = ["-Ccodegen-units=1"]
```
More specific `outer/inner/.cargo/config` contains
```toml
[build]
rustflags = ["-Ccodegen-units=2"]
```

The compiler's command line in this case is
```sh
rustc --crate-name inner --edition=2018 'src\main.rs' ... -Ccodegen-units=2 -Ccodegen-units=1
```
`rustc` like many other tools prefers the last specified value for the given option to support hierarchical build systems.
With Cargo's flag concatenation behavior this means that less specific flags override more specific flags, which is not a correct behavior.

cc https://github.com/rust-lang/rust/pull/70729#issuecomment-610034489

### LINKED ISSUE #12506: Merged configuration lists have unexpected precedence ordering
### Problem

When merging TOML lists from configuration, the order does not follow the precedence order described in the documentation. The documentation does not explicitly specify the order for merging lists other than "Arrays will be joined together".

The [docs state](https://doc.rust-lang.org/cargo/reference/config.html)
> If a key is specified in multiple config files, the values will get merged together. Numbers, strings, and booleans will use the value in the deeper config directory taking precedence over ancestor directories, where the home directory is the lowest priority. Arrays will be joined together.

> The --config option may be specified multiple times, in which case the values are merged in left-to-right order, using the same merging logic that is used when multiple configuration files apply. Configuration values specified this way take precedence over environment variables, which take precedence over configuration files.

I expected a merged list would be (or the reverse of this):
* command line
* environment
* deeper config
* shallower config

Instead, the order appears to be:
* deeper config
* shallower config
* command line
* environment

### Steps

The following script illustrates the current situation.
```bash
mkdir -p experiment/.cargo/
echo "test=['file']" > experiment/.cargo/config.toml
mkdir -p experiment/nested/.cargo/
echo "test=['nested-file']" > experiment/nested/.cargo/config.toml

pushd experiment/nested
CARGO_TEST=env-var cargo -Zunstable-options --config 'test=["first --config"]' --config 'test=["second --config"]' config get test
popd
rm -r experiment
```

```
test = ["nested-file", "file", "first --config", "second --config", "env-var"]
```

If a non-list is used, the precedence of these values is:
```
second --config
first --config
env-var
nested-file
file
```

### Possible Solution(s)

Merge the lists with highest precedence first. The output of the above bash script should be:

```
test = ["second --config", "first --config", "env-var", "nested-file", "file"]
```

An alternative would be to merge them with lowest precedence first.

```
test = ["file", "nested-file", "env-var", "first --config", "second --config"]
```

### Notes

_No response_

### Version

```text
cargo 1.73.0-nightly (7e9de3f4e 2023-08-13)
release: 1.73.0-nightly
commit-hash: 7e9de3f4ec3708f500bec142317895b96131e47c
commit-date: 2023-08-13
libcurl: 8.2.1-DEV (sys:0.4.65+curl-8.2.1 vendored ssl:OpenSSL/1.1.1u)
```

arlosi:
When merging configuration lists, the current order does not match the expected precedence. This makes merged lists follow precedence order, with higher precedence items merged later in lists.

When a list in configuration exists in multiple places, Cargo merges the lists together. The ordering of this merging is unexpected and does not follow the precedence rules that non-list configuration uses.

The current merging order appears to be:
* project-specific `config.toml`
* global `config.toml`
* command-line (`--config`)
* environment variable (`CARGO_*`)

This PR changes the order to follow the precedence rules with higher precedence configuration merging later in the lists.
* global `config.toml`
* project-specific `config.toml`
* environment variable (`CARGO_*`)
* command-line (`--config`)

This aligns with config such as `build.rustflags` where later flags take precedence over earlier ones.

Since `--config` is relatively new, it's unlikely to cause too much breakage by making it come after environment variables.

Switching global and project-specific ordering is more likely to cause breakage, since it's been around longer (reported as an issue in #8128). Projects relying on global configuration flags (in `$CARGO_HOME\config.toml` or in `.cargo/config.toml` further from the project) being merged first in lists will be broken.

For most uses of merged lists (such as `build.rustflags`), if the flags do not conflict with each other, there will be no impact.

Fixes #12506
Fixes #8128

### DIFF
--- a/src/cargo/util/config/mod.rs
+++ b/src/cargo/util/config/mod.rs
@@ -938,6 +938,7 @@ impl Config {
                     .map(|s| (s.to_string(), def.clone())),
             );
         }
+        output.sort_by(|a, b| a.1.cmp(&b.1));
         Ok(())
     }
 
@@ -2106,8 +2107,8 @@ impl ConfigValue {
 
     /// Merge the given value into self.
     ///
-    /// If `force` is true, primitive (non-container) types will override existing values.
-    /// If false, the original will be kept and the new value ignored.
+    /// If `force` is true, primitive (non-container) types will override existing values
+    /// of equal priority. For arrays, incoming values of equal priority will be placed later.
     ///
     /// Container types (tables and arrays) are merged with existing values.
     ///
@@ -2115,7 +2116,13 @@ impl ConfigValue {
     fn merge(&mut self, from: ConfigValue, force: bool) -> CargoResult<()> {
         match (self, from) {
             (&mut CV::List(ref mut old, _), CV::List(ref mut new, _)) => {
-                old.extend(mem::take(new).into_iter());
+                if force {
+                    old.append(new);
+                } else {
+                    new.append(old);
+                    mem::swap(new, old);
+                }
+                old.sort_by(|a, b| a.1.cmp(&b.1));
             }
             (&mut CV::Table(ref mut old, _), CV::Table(ref mut new, _)) => {
                 for (key, value) in mem::take(new) {
--- a/src/cargo/util/config/value.rs
+++ b/src/cargo/util/config/value.rs
@@ -10,6 +10,7 @@
 
 use crate::util::config::Config;
 use serde::de;
+use std::cmp::Ordering;
 use std::fmt;
 use std::marker;
 use std::mem;
@@ -61,6 +62,24 @@ pub enum Definition {
     /// Passed in on the command line.
     /// A path is attached when the config value is a path to a config file.
     Cli(Option<PathBuf>),
+}
+
+impl PartialOrd for Definition {
+    fn partial_cmp(&self, other: &Definition) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for Definition {
+    fn cmp(&self, other: &Definition) -> Ordering {
+        if mem::discriminant(self) == mem::discriminant(other) {
+            Ordering::Equal
+        } else if self.is_higher_priority(other) {
+            Ordering::Greater
+        } else {
+            Ordering::Less
+        }
+    }
 }
 
 impl Definition {
--- a/src/doc/src/reference/config.md
+++ b/src/doc/src/reference/config.md
@@ -28,7 +28,8 @@ with a configuration file in your home directory.
 If a key is specified in multiple config files, the values will get merged
 together. Numbers, strings, and booleans will use the value in the deeper
 config directory taking precedence over ancestor directories, where the
-home directory is the lowest priority. Arrays will be joined together.
+home directory is the lowest priority. Arrays will be joined together
+with higher precedence items being placed later in the merged array.
 
 At present, when being invoked from a workspace, Cargo does not read config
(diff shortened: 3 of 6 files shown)


arlosi:
Since this is a bug fix breaking change, let's make sure everyone is OK with it.

@rfcbot fcp merge

weihanglo:
You need @rustbot label +T-cargo to do an FCP

arlosi:
Thanks @weihanglo. Let's try again.

@rfcbot fcp merge

epage:
Could the PR re-summarize the situation, especially who will be broken and the level of impact of that?

weihanglo:
Should we do a crater run for this?
I doubt it'll catch anything interesting with crater though. Usually this kind of configuration is deeply rooted in a user's or an enterprise environment.

ehuss:
I don't think crater would be able to run in situations were there would be separate configuration files to merge.

weihanglo:
I'll check my box here but want to raise the awareness to a wider audience. I knew some large cooperations do weird configurations around `build.rustflags`. We should have a way to solicit feedback before specifying an unspecified behavior. An FCP period sounds too short to me.

@rustbot reviewed
@rustbot concern raise-awareness

weihanglo:
I'll check my box here but want to raise the awareness to a wider audience. I knew some large cooperations do weird configurations around `build.rustflags`. We should have a way to solicit feedback before specifying an unspecified behavior. An FCP period sounds too short to me.

@rfcbot reviewed
@rfcbot concern raise-awareness

weihanglo:
We've talked about this in the Cargo weekly meeting today. We'll have a blog post for this incoming change, and solicit feedback from users. Since beta is just branched off, we'll make this change available on nightly as soon as possible.

@rustbot resolve raise-awareness


We would like to make sure this will get into release note as a compatibility note, so

@rustbot label +relnotes

weihanglo:
Oh no it's rfcbot

@rfcbot resolve raise-awareness

src/cargo/util/config/mod.rs
```diff
@@ -2106,16 +2107,22 @@ impl ConfigValue {
 
     /// Merge the given value into self.
     ///
-    /// If `force` is true, primitive (non-container) types will override existing values.
-    /// If false, the original will be kept and the new value ignored.
+    /// If `force` is true, primitive (non-container) types will override existing values
+    /// of equal priority. For arrays, incoming values of equal priority will be placed later.
```
weihanglo, line 2111: The code looks correct. Just what to make sure we have test to verify this behavior.

weihanglo (approved):
I'll merge this today and update the submodule. Thanks!

PR #12521, opened 2023-08-17
Make cargo-credential-gnome-secret built-in as cargo:libsecret

arlosi:
We previously couldn't have cargo-credential-gnome-secret built into Cargo because of its build-time dependency on `libsecret`. However, this limitation has now been lifted by #12518.

Adds a new built-in credential provider `cargo:libsecret`.

Adds `ISC` as an allowed license for `libloading`. 
* `rustc` already uses `libloading`
* ISC license is very similar to MIT

Renames the crate from `cargo-credential-gnome-secret` to `cargo-credential-libsecret` and changes the crate structure to more closely match `wincred` and `macos-keychain`.


credential/cargo-credential-gnome-secret/src/main.rs
```diff
@@ -1,12 +1,5 @@
 //! Cargo registry gnome libsecret credential process.
 
-#[cfg(target_os = "linux")]
-mod libsecret;
-#[cfg(not(target_os = "linux"))]
-use cargo_credential::UnsupportedCredential as GnomeSecret;
-#[cfg(target_os = "linux")]
-use libsecret::GnomeSecret;
-
 fn main() {
```
epage, line 3: Was `main.rs` meant to be removed?

Cargo.toml
```diff
@@ -123,6 +124,7 @@ base64.workspace = true
 bytesize.workspace = true
 cargo-platform.workspace = true
 cargo-credential.workspace = true
+cargo-credential-gnome-secret.workspace = true
```
epage, line 127: Should we rename it to match the built-in name?

PR #12527, opened 2023-08-18
refactor: put `Source` trait under `cargo::sources`

weihanglo:

It feels more natural that `Source` trait is under `cargo::sources`, as that module contains all built-in implementaions of `Source` trait.

This PR also flattens the module path of `SourceId`.

Should be safe since this doesn't include any behavior change.

### Additional information

This was a something in my git stash for a while when I trie extracting the real "core" types into a sane module that doesn't contain any source downloading logic.


epage:
I have mixed feelings about this.

Overall, our layering is a mess in that `core` and `sources` depend on each other.  I'd love to be findign ways to make the layering more strict as i feel that will help make the code more understandable in the short term and easier to split up into smaller, more understandable chunks in the long term.

Most likely this kind of layering would involve keeping the trait where it is and having a way to inject the implementations into core.

But maybe we are too far off from that to be worth designing towards?  And having all source related stuff in one place could be nice?  

What are your thoughts @weihanglo ?

weihanglo:
What I was trying to do is keep the "core" types being pure types, so that other modules importing them don't need to pull in unrelated code. For example, `SourceId` can be decoupled from `Source` trait, so that if we had a lockfile processing module, it doesn't need to depend on `Source` trait and its family.

`SourceId::load` is an outlier of that, though we can turn it a `SourceIdExt` it under `cargo::sources`. People want that extension can import it at will.

Not sure if it makes sense though.

epage:
Eh, my main concern is how this helps us get to long term goals but those are probably far enough away that its not worth blocking this on

@bors r+

PR #12531, opened 2023-08-19
refactor: Use `Config::emit_diagnostic` to emit warnings

Muscraft:
#12235 is tracking user control over warnings. Part of the effort means moving warnings from being emitted by `Shell` to `Config`. This is for a few reasons:
1. `Shell` should not know what a user wants to do with a given warning other than emit anything that is passed to it
2. `Config` should (eventually) know what a user wants to do for each warning `cargo` is attempting to display, be that `forbid`, `deny`, `warn`, or `allow`.

In both scenarios, it makes sense to move things to `Config`, which is what this PR does. It adds a new function `Config::emit_diagnostic`, which is a copy of  [`Shell::warn`](https://github.com/rust-lang/cargo/blob/80eca0e58fb2ff52c1e94fc191b55b37ed73e0e4/src/cargo/core/shell.rs#L262), and moves all calls from `Shell::warn` to this new function.

### DIFF
--- a/crates/xtask-bump-check/src/xtask.rs
+++ b/crates/xtask-bump-check/src/xtask.rs
@@ -219,7 +219,7 @@ fn get_base_commit<'a>(
             let upstream_ref = upstream_branches[0].get();
             if upstream_branches.len() > 1 {
                 let name = upstream_ref.name().expect("name is valid UTF-8");
-                let _ = config.shell().warn(format!(
+                let _ = config.emit_diagnostic(format!(
                     "multiple `{UPSTREAM_BRANCH}` found, picking {name}"
                 ));
             }
--- a/src/cargo/core/shell.rs
+++ b/src/cargo/core/shell.rs
@@ -2,7 +2,7 @@ use std::fmt;
 use std::io::prelude::*;
 use std::io::IsTerminal;
 
-use termcolor::Color::{Cyan, Green, Red, Yellow};
+use termcolor::Color::{Cyan, Green, Red};
 use termcolor::{self, Color, ColorSpec, StandardStream, WriteColor};
 
 use crate::util::errors::CargoResult;
@@ -126,7 +126,7 @@ impl Shell {
 
     /// Prints a message, where the status will have `color` color, and can be justified. The
     /// messages follows without color.
-    fn print(
+    pub(crate) fn print(
         &mut self,
         status: &dyn fmt::Display,
         message: Option<&dyn fmt::Display>,
@@ -256,14 +256,6 @@ impl Shell {
         }
         self.output
             .message_stderr(&"error", Some(&message), Red, false)
-    }
-
-    /// Prints an amber 'warning' message.
-    pub fn warn<T: fmt::Display>(&mut self, message: T) -> CargoResult<()> {
-        match self.verbosity {
-            Verbosity::Quiet => Ok(()),
-            _ => self.print(&"warning", Some(&message), Yellow, false),
-        }
     }
 
     /// Prints a cyan 'note' message.
--- a/src/cargo/ops/cargo_add/mod.rs
+++ b/src/cargo/ops/cargo_add/mod.rs
@@ -217,7 +217,9 @@ pub fn add(workspace: &Workspace<'_>, options: &AddOptions<'_>) -> CargoResult<(
     }
 
     if options.dry_run {
-        options.config.shell().warn("aborting add due to dry run")?;
+        options
+            .config
+            .emit_diagnostic("aborting add due to dry run")?;
     } else {
         manifest.write()?;
     }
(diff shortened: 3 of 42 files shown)


crates/xtask-bump-check/src/xtask.rs
```diff
@@ -219,7 +219,7 @@ fn get_base_commit<'a>(
             let upstream_ref = upstream_branches[0].get();
             if upstream_branches.len() > 1 {
                 let name = upstream_ref.name().expect("name is valid UTF-8");
-                let _ = config.shell().warn(format!(
+                let _ = config.emit_diagnostic(format!(
```
epage, line 222: Should xtasks still use `shell()`?

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -217,7 +217,9 @@ pub fn add(workspace: &Workspace<'_>, options: &AddOptions<'_>) -> CargoResult<(
     }
 
     if options.dry_run {
-        options.config.shell().warn("aborting add due to dry run")?;
+        options
+            .config
+            .emit_diagnostic("aborting add due to dry run")?;
```
epage, line 222: Should this use the diagnostic system or just be a warning?

src/cargo/ops/cargo_generate_lockfile.rs
```diff
@@ -159,8 +159,7 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
     }
     if opts.dry_run {
         opts.config
-            .shell()
-            .warn("not updating lockfile due to dry run")?;
+            .emit_diagnostic("not updating lockfile due to dry run")?;
```
epage, line 162: Ditto about diagnostics

src/cargo/ops/cargo_remove.rs
```diff
@@ -55,8 +55,7 @@ pub fn remove(options: &RemoveOptions<'_>) -> CargoResult<()> {
     if options.dry_run {
         options
             .config
-            .shell()
-            .warn("aborting remove due to dry run")?;
+            .emit_diagnostic("aborting remove due to dry run")?;
```
epage, line 58: another dry run

src/cargo/ops/registry/publish.rs
```diff
@@ -387,7 +387,7 @@ fn transmit(
 
     // Do not upload if performing a dry run
     if dry_run {
-        config.shell().warn("aborting upload due to dry run")?;
+        config.emit_diagnostic("aborting upload due to dry run")?;
```
epage, line 390: and another.  Maybe I should stop?

src/cargo/core/shell.rs
```diff
@@ -258,14 +258,6 @@ impl Shell {
             .message_stderr(&"error", Some(&message), Red, false)
     }
 
-    /// Prints an amber 'warning' message.
-    pub fn warn<T: fmt::Display>(&mut self, message: T) -> CargoResult<()> {
```
epage, line 262: nit: would have been nice for this to be removed in a separate commit

src/cargo/ops/cargo_package.rs
```diff
@@ -523,18 +523,15 @@ fn check_repo_state(
...
+                path.display(),
+                workdir.display()
+            ))?;
         }
     } else {
-        config.shell().verbose(|shell| {
-            shell.warn(format!("No (git) VCS found for `{}`", p.root().display()))
-        })?;
+        config
+            .emit_verbose_diagnostic(format!("No (git) VCS found for `{}`", p.root().display()))?;
```
epage, line 534: I believe these are the only `emit_verbose_diagnostic`.  

How should these be handled?
- In the future, put these under a `pedantic` group and remove `emit_verbose_diagnostic`
- Always display these
- Not make these diagnostics but more like logging messages

src/cargo/util/config/mod.rs
```diff
@@ -1949,6 +1949,33 @@ impl Config {
...
+                Some(&diagnostic),
+                termcolor::Color::Yellow,
+                false,
+            ),
+        }
+    }
+
+    /// Emits a diagnostic only if the verbosity level is `Verbose`. This uses
+    /// [Config::emit_diagnostic] internally.
+    pub fn emit_verbose_diagnostic<T: fmt::Display>(&self, diagnostic: T) -> CargoResult<()> {
```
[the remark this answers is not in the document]
weihanglo: This looks like a duplicate of `Shell::warn`. Should we remove `Shell::warn`, or at least do something to prevent it from being use again?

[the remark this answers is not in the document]
weihanglo: Oh just saw epage's comment <https://github.com/rust-lang/cargo/pull/12531#discussion_r1300456322> 😆.

Issue #12472, opened 2023-08-09
Platform-specific `linker` option ignored when specified using `cfg` syntax


tgnottingham:
### Problem

When the following is placed in `.cargo/config.toml`, the `linker` option is respected (on an x86_64 Linux system):

```
[target.x86_64-unknown-linux-gnu]
linker = "clang"
rustflags = ["-C", "link-arg=-fuse-ld=lld"]
```

However, when the `cfg` syntax is used, as in the following, the `linker` option seems to be ignored (`-C linker=clang` isn't visible in `cargo -vv` output):

```
[target.'cfg(unix)']
linker = "clang"
rustflags = ["-C", "link-arg=-fuse-ld=lld"]
```

In either case, the `rustflags` option is respected.

Is the `linker` option being ignored here a bug?

### Version

```text
cargo 1.71.1 (7f1d04c00 2023-07-29)
release: 1.71.1
commit-hash: 7f1d04c0053083b98fa50b69b6f56e339b0556a8
commit-date: 2023-07-29
host: x86_64-unknown-linux-gnu
libgit2: 1.6.4 (sys:0.17.1 vendored)
libcurl: 8.0.1-DEV (sys:0.4.61+curl-8.0.1 vendored ssl:OpenSSL/1.1.1t)
ssl: OpenSSL 1.1.1t  7 Feb 2023
os: Ubuntu 22.04 (jammy) [64-bit]
```

weihanglo:
[This section of `[target]`](https://doc.rust-lang.org/1.71.1/cargo/reference/config.html#target>) config lists all possible configurations and doesn't include `target.<cfg>'.linker` and `target.<cfg>.<links>`.

Delved into the history, it seems that just nobody requested for that.

* For rustflags — https://github.com/rust-lang/cargo/pull/3854 
* For runner — https://github.com/rust-lang/cargo/pull/5959

However, the mechanisms of plain target-triple and `cfg` expression are different. May need some efforts to implement those configs. My guess is that they'll behave the same as `target.<cfg>.runner`: `<triple>` takes precedence over them, and errors out when multiple `<cfg>` of them found.

@ehuss the wise, was there any reason not supporting `linker` and link overridden with cfg expression?

ehuss:
I don't think there is a specific reason other than nobody needed it. If we add it, I suppose it should follow the same logic as `runner`.

weihanglo:
Thanks for quick reply!

[...]

PR #12535, opened 2023-08-21
WIP: Add support for `target.'cfg(..)'.linker`

hi-rustin:

close https://github.com/rust-lang/cargo/issues/12472

Add support for `target.'cfg(..)'.linker`.

- [ ] unit test
-->


hi-rustin (commented):
Still working in progress. Not ready for review.

hi-rustin (commented):
Still working in progress. Not ready for review.

src/cargo/core/compiler/compilation.rs
```diff
@@ -441,3 +455,39 @@ fn target_runner(
         )
     }))
 }
+
+/// Gets the user-specified linker for a particular host or target from the configuration.
+fn target_linker(bcx: &BuildContext<'_, '_>, kind: CompileKind) -> CargoResult<Option<PathBuf>> {
```
weihanglo, line 460: So sad we cannot easily share the code with runner 😆.

tests/testsuite/tool_paths.rs
```diff
@@ -32,6 +32,91 @@ fn pathless_tools() {
...
+            linker = "nonexistent-linker2"
+            "#,
+        )
+        .build();
+
+    foo.cargo("build --verbose")
+        .with_status(101)
+        .with_stderr_contains(&format!(
+            "\
+[ERROR] several matching instances of `target.'cfg(..)'.linker` in configurations
```
weihanglo, line 114: should we match the entire error message here?

weihanglo (commented):
Thanks! Could you update the doc with a new entry for `target.<cfg>.runner`, which should look like the [runner](https://doc.rust-lang.org/nightly/cargo/reference/config.html#targetcfgrunner) one?

tests/testsuite/tool_paths.rs
hi-rustin: Updated! Good suggestion.

src/doc/src/reference/config.md
```diff
@@ -1098,6 +1098,17 @@ This option is deprecated and unused.
 Specifies the linker which is passed to `rustc` (via [`-C linker`]) when the
 [`<triple>`] is being compiled for. By default, the linker is not overridden.
 
+##### `target.<cfg>.linker`
+This is similar to the [target linker](#targettriplelinker), but using
+a [`cfg()` expression]. 
+
+The config precedence is:
+1. `host.<triple>.linker`, if [host-config](unstable.md#host-config) is enabled.
```
weihanglo, line 1106: I appreciate the effort, though not sure if it is worth a mention, since it may be a bit more confusing that people thought this is a stable feature. OTOH people will be aware of this feature and start testing it.

Therefore, either it is marked as unstable explicitly, or remove.


```suggestion
1. `host.<triple>.linker`, if the unstable feature [host-config](unstable.md#host-config) is enabled.
```

weihanglo (changes requested):
Thanks for the effort. I guess only one thing left: this build script environment `RUSTC_LINKER` should update accordingly.

https://github.com/rust-lang/cargo/blob/e3b3ed89115f96fe7f70d68188bf2907f3d6fdce/src/cargo/core/compiler/custom_build.rs#L302-L307

src/doc/src/reference/config.md
hi-rustin: I removed it. Thanks!

hi-rustin:
> I guess only one thing left: this build script environment `RUSTC_LINKER` should update accordingly.

Good catch! Updated. 
Thanks for your review!  💚 💙 💜 💛

hi-rustin:
r? @weihanglo 

weihanglo (commented):
Not sure why you removed the test for build script. Maybe you could try this?

```rust
#[cargo_test]
fn custom_build_env_var_rustc_linker_with_target_cfg() {
    let unused_target = cross_compile::unused();
    let p = project()
        .file(
            ".cargo/config",
            r#"
            # HACK: We know that unused target is wasm32
            [target.'cfg(target_arch = "wasm32")']
            linker = "/path/to/linker"
            "#,
        )
        .file(
            "build.rs",
            r#"
            use std::env;

            fn main() {
                assert!(env::var("RUSTC_LINKER").unwrap().ends_with("/path/to/linker"));
            }
            "#,
        )
        .file("src/lib.rs", "")
        .build();

    // no crate type set => linker never called => build succeeds if and
    // only if build.rs succeeds, despite linker binary not existing.
    p.cargo("build --target")
        .arg(&unused_target)
        .run();
}
```

hi-rustin:
> Not sure why you removed the test for build script. Maybe you could try this?

Thanks for your help! It works.

I removed it because I thought it was difficult to construct a workable test case for all OSs and targets.

weihanglo:
Hmm. Good catch. Maybe only run the test on x86 Linux. Do cross compile. And set `cfg(target_arch = "i686")` in the config?

tests/testsuite/build_script.rs
```diff
@@ -453,6 +453,41 @@ fn custom_build_env_var_rustc_linker() {
     p.cargo("build --target").arg(&target).run();
 }
 
+#[cargo_test]
+fn custom_build_env_var_rustc_linker_with_target_cfg() {
+    // Only run this test on aarch64, since it's difficult to construct
+    // a case suitable for all platforms.
+    if cross_compile::disabled() || cfg!(not(target_arch = "aarch64")) {
+        return;
+    }
```
weihanglo, line 462: I lean towards testing on Linux for such reasons

* It's faster running the test suite on Linux than on macOS and Windows.
* Only on Linux Cargo's CI pipeline covers beta toolchain.
* There is no CI job running on `aarch64` at this moment. This is skipped.

```suggestion
// Only run this test on aarch64, since it's difficult to construct a case suitable for all platforms.
#[cargo_test]
#[cfg(target_os = "linux")]
fn custom_build_env_var_rustc_linker_with_target_cfg() {
    if cross_compile::disabled() {
        return;
    }
```

And below change to `i686` I believe it would work.

hi-rustin: Oh, good catch! Thanks!

tests/testsuite/build_script.rs
```diff
@@ -453,6 +453,43 @@ fn custom_build_env_var_rustc_linker() {
...
+    if cross_compile::disabled() {
+        return;
+    }
+
+    let target = cross_compile::alternate();
+    let p = project()
+        .file(
+            ".cargo/config",
+            r#"
+            [target.'cfg(target_arch = "i686")']
```
weihanglo, line 471: My fault. `i686` is still x86 arch. Just the pointer is 32bits. Trust me one more time. This should work 🙇🏾.

```suggestion
            [target.'cfg(target_pointer_width = "32")']
```

src/cargo/core/compiler/compilation.rs
```diff
@@ -441,3 +455,39 @@ fn target_runner(
...
+    let matching_linker = cfgs.next();
+    if let Some((key, runner)) = cfgs.next() {
+        anyhow::bail!(
+            "several matching instances of `target.'cfg(..)'.linker` in configurations\n\
+             first match `{}` located in {}\n\
+             second match `{}` located in {}",
+            matching_linker.unwrap().0,
+            matching_linker.unwrap().1.definition,
+            key,
+            runner.definition
```
weihanglo, line 489:
```suggestion
    if let Some((key, linker)) = cfgs.next() {
        anyhow::bail!(
            "several matching instances of `target.'cfg(..)'.linker` in configurations\n\
             first match `{}` located in {}\n\
             second match `{}` located in {}",
            matching_linker.unwrap().0,
            matching_linker.unwrap().1.definition,
            key,
            linker.definition
```

weihanglo (commented):
Apologize I led you to a wrong `cfg` 😅.

weihanglo (approved):

weihanglo:
Thanks for working on this along with us!

@bors r+

tests/testsuite/build_script.rs
hi-rustin: Thanks!

PR #12542, opened 2023-08-23
fix(log): Use a more compact relative-time format

epage:
This changes logged messages from
```
2023-08-23T01:01:59.922018Z DEBUG cargo::core::compiler::fingerprint: filesystem up-to-date "/home/epage/src/personal/dump"
```
To
```
   0.041729583s DEBUG cargo::core::compiler::fingerprint: filesystem up-to-date "/home/epage/src/personal/dump"
```
Benefits
- Less horizontal space taken up in boilerplate
- Easier to compare within a run

Downsides
- Harder to correlate with other processes, like with crates.io server operations

This gives us up to 4 digits for seconds which should be sufficient for cargo build times.

We could make this more compact by dropping the digits of precision from 9 to 6 but that would require a custom Timer which might be a paint to keep in sync between packages.

### DIFF
--- a/crates/xtask-bump-check/src/main.rs
+++ b/crates/xtask-bump-check/src/main.rs
@@ -20,6 +20,7 @@ fn setup_logger() {
     let env = tracing_subscriber::EnvFilter::from_env("CARGO_LOG");
 
     tracing_subscriber::fmt()
+        .with_timer(tracing_subscriber::fmt::time::Uptime::default())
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
Looks great.

Have you considered what Arlo said, output the current time at the beginning?

https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/.60CARGO_LOG.60.20output

weihanglo:
Thanks!

@bors r+

PR #12544, opened 2023-08-23
fix(update): Clarify meaning of --aggressive as --recursive

epage:
When working on cargo-upgrade, I found the meaning of `--aggressive` confusing and named it `--recursive` there.

Renaming this in `cargo update` (with a backwards compatible alias) was referenced in #12425.

### DIFF
--- a/src/etc/cargo.bashcomp.sh
+++ b/src/etc/cargo.bashcomp.sh
@@ -87,7 +87,7 @@ _cargo()
 	local opt__t="$opt__test"
 	local opt__tree="$opt_common $opt_pkg_spec $opt_feat $opt_mani $opt_lock --target -i --invert --prefix --no-dedupe --duplicates -d --charset -f --format -e --edges"
 	local opt__uninstall="$opt_common $opt_lock $opt_pkg --bin --root"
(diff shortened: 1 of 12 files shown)


epage:

@rfcbot fcp merge

src/etc/cargo.bashcomp.sh
```diff
@@ -87,7 +87,7 @@ _cargo()
 	local opt__t="$opt__test"
 	local opt__tree="$opt_common $opt_pkg_spec $opt_feat $opt_mani $opt_lock --target -i --invert --prefix --no-dedupe --duplicates -d --charset -f --format -e --edges"
 	local opt__uninstall="$opt_common $opt_lock $opt_pkg --bin --root"
-	local opt__update="$opt_common $opt_mani $opt_lock $opt_pkg --aggressive --precise --dry-run"
+	local opt__update="$opt_common $opt_mani $opt_lock $opt_pkg --aggressive --recursive --precise --dry-run"
```
weihanglo, line 90: Should we demote `--aggressive` by removing it from shell completions?

weihanglo (commented):
Should we have a simple test to verify that `--aggressive` alias still works?

weihanglo (approved):

weihanglo:
Thanks for the update!

@bors r+

PR #12545, opened 2023-08-23
fix(update): Make `-p` more convenient by being positional

epage:
Generally, cargo avoids positional arguments.  Mostly for the commands that might forward arguments to another command, like `cargo test`. It also allows some flexibility in turning flags into options.

For `cargo add` and `cargo remove`, we decided to accept positionals because the motivations didn't seem to apply as much (similar to `cargo install`).

This applies the pattern to `cargo update` as well which is in the same category of commands as `cargo add` and `cargo remove`.

As for `--help` formatting, I'm mixed on whether `[SPEC]...` should be at the top like other positionals or should be relegated to "Package selection".  I went with the latter mostly to make it easier to visualize the less common choice.

Switching to a positional for `cargo update` (while keeping `-p` for backwards compatibility) was referenced in #12425.

### DIFF
--- a/src/etc/_cargo
+++ b/src/etc/_cargo
@@ -344,7 +344,8 @@ _cargo() {
                         '--aggressive=[force dependency update]' \
                         "--dry-run[don't actually write the lockfile]" \
                         '(-p --package)'{-p+,--package=}'[specify package to update]:package:_cargo_package_names' \
-                        '--precise=[update single dependency to precise release]:release'
+                        '--precise=[update single dependency to precise release]:release' \
+                        '*:package:_cargo_package_names'
                         ;;
 
                 verify-project)
--- a/src/bin/cargo/commands/update.rs
+++ b/src/bin/cargo/commands/update.rs
@@ -1,5 +1,6 @@
 use crate::command_prelude::*;
 
+use anyhow::anyhow;
 use cargo::ops::{self, UpdateOptions};
 use cargo::util::print_available_packages;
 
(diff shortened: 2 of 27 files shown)


epage:

@rfcbot fcp merge

epage:
when running `cargo tree gix-actor`, I realized we should probably do the same thing there

weihanglo:
> when running `cargo tree gix-actor`, I realized we should probably do the same thing there

I may hold off this now. `cargo tree` has `-i` and `-d` flags, which make things a bit complicated.

epage:
> I may hold off this now. cargo tree has -i and -d flags, which make things a bit complicated.

How so?

weihanglo:
Just checked what we have today in cargo-tree. It is already confusing.

```console
$ cargo tree -p resolver-tests -i tracing
tracing v0.1.37
├── cargo v0.75.0 (/projects/cargo)
│   └── resolver-tests v0.0.0 (/projects/cargo/crates/resolver-tests)
├── cargo-util v0.2.7 (/projects/cargo/crates/cargo-util)
│   ├── cargo v0.75.0 (/projects/cargo) (*)
│   └── resolver-tests v0.0.0 (/projects/cargo/crates/resolver-tests)
└── tracing-subscriber v0.3.17
    └── cargo v0.75.0 (/projects/cargo) (*)
```

This command prints an inverted tree with `tracing` as the root, and… without looking into code I can't even understand how to interpret the effect of `--package` in the graph.

They are completely different issues, so not a blocker :)

epage:
Oh man, I forgot `-i` could accept a package **or** use the `-p` package.  Yeah, thats a no-go

src/etc/_cargo
```diff
@@ -345,6 +345,7 @@ _cargo() {
                         "--dry-run[don't actually write the lockfile]" \
                         '(-p --package)'{-p+,--package=}'[specify package to update]:package:_cargo_package_names' \
                         '--precise=[update single dependency to precise release]:release'
```
weihanglo, line 347:
```suggestion
                        '--precise=[update single dependency to precise release]:release' \
```

weihanglo (commented):
Looks good. Just one trailing backslash is missing 😆.

src/etc/_cargo
weihanglo: r=me when this is fixed

PR #12548, opened 2023-08-24
use AND search when having multiple terms

tshepang:
This makes search more like like common search engines


weihanglo (approved):
Looks like people have used for a while: <https://github.com/search?q=org%3Arust-lang%20use-boolean-and&type=code>.

I am good with this 👍🏾.

PR #12551, opened 2023-08-24
Create dedicated unstable flag for asymmetric-token

arlosi:
Asymmetric tokens are gated by `-Zcredential-process`. Since we're considering stabilizing that soon, this moves asymmetric token support to have its own unstable flag.

It was previously gated by `-Zregistry-auth`, and some of the docs were not updated when it moved.

r? @Eh2406

### DIFF
--- a/src/cargo/util/auth/mod.rs
+++ b/src/cargo/util/auth/mod.rs
@@ -76,9 +76,8 @@ impl RegistryConfigExtended {
 /// Get the list of credential providers for a registry source.
 fn credential_provider(config: &Config, sid: &SourceId) -> CargoResult<Vec<Vec<String>>> {
     let cfg = registry_credential_config_raw(config, sid)?;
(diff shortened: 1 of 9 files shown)


src/cargo/util/auth/mod.rs
```diff
@@ -454,7 +453,10 @@ fn credential_action(
         tracing::debug!("attempting credential provider: {args:?}");
         let provider: Box<dyn Credential> = match process {
             "cargo:token" => Box::new(TokenCredential::new(config)),
-            "cargo:paseto" => Box::new(PasetoCredential::new(config)),
+            "cargo:paseto" if config.cli_unstable().asymmetric_token => {
+                Box::new(PasetoCredential::new(config))
+            }
+            "cargo:paseto" => bail!("cargo:paseto requires -Zasymmetric-token"),
```
arlosi, line 459: We could also decide to silently skip it rather than error here.

PR #12556, opened 2023-08-24
fix(toml): Improve parse errors

epage:

When we adopted `toml_edit`, we got TOML syntax errors that showed the context for where the error occurred.  However, the work was not done to extend this to semantic errors reported by serde.

This updates `Cargo.toml` and `Cargo.lock` code to provide that context on semantic errors.  `config.toml` is not done because the schema is decentralized.

In theory, this will also improve performance because we aren't having to allocate a lot of intermediate data to then throw away for every `Cargo.toml` we read.

Check by commit to see this change gradually.
- The `package.cargo-features` change was made to drop out dependence on `toml::Table` so we could do the direct deserialization

### DIFF
--- a/src/cargo/ops/lockfile.rs
+++ b/src/cargo/ops/lockfile.rs
@@ -2,7 +2,6 @@ use std::io::prelude::*;
 
 use crate::core::{resolver, Resolve, ResolveVersion, Workspace};
 use crate::util::errors::CargoResult;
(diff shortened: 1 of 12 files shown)


arlosi:
Looks great!

@bors r+

weihanglo (approved):
Love the change!

PR #12559, opened 2023-08-25
string leek is stable

Eh2406:

String leek is now stable so lets use it.

Code compiles and test run.

### DIFF
--- a/src/cargo/util/interning.rs
+++ b/src/cargo/util/interning.rs
@@ -12,10 +12,6 @@ use std::str;
 use std::sync::Mutex;
 use std::sync::OnceLock;
 
-fn leak(s: String) -> &'static str {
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
Looks like `Box::leak(s.into_boxed_str())` and `String::leak` are not 100% equivalent.

> It does not reallocate or shrink the String, so the leaked allocation may include unused capacity that is not part of the returned slice. If you don’t want that, call [into_boxed_str](https://doc.rust-lang.org/std/string/struct.String.html#method.into_boxed_str), and then [Box::leak](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.leak).

So we end up using slightly more memory and in return less re-allocations?

Anyway it shouldn't be a problem. Thanks for this.

PR #12560, opened 2023-08-25
feat(resolver): **Very** preliminary MSRV resolver support

epage:

A bare bones implementation of an MSRV resolver that is good enough for people running on nightly when they really need it but is not ready for general use.  

Current limitations
- Does not honor `--ignore-version`
- Gives terrible error messages
- Nothing is done yet regarding `cargo install`
- Doesn't inform the user when choosing non-latest

These will be  noted in #9930 on merge.

Implementation wise, this is yet another hack (sorry @Eh2406).  Our expectation to get this GA is to refactor the resolver to make the cargo/resolver boundary look a little more like the cargo/pubgrub boundary so we can better control policy without any of these hacks which will also make having all of the policy we need for this easier to maintain.

This is a part of #9930

Per commit

### DIFF
--- a/src/cargo/core/compiler/standard_lib.rs
+++ b/src/cargo/core/compiler/standard_lib.rs
@@ -145,6 +145,7 @@ pub fn resolve_std<'cfg>(
     let cli_features = CliFeatures::from_command_line(
         &features, /*all_features*/ false, /*uses_default_features*/ false,
     )?;
+    let max_rust_version = ws.rust_version();
     let resolve = ops::resolve_ws_with_opts(
         &std_ws,
         target_data,
@@ -153,6 +154,7 @@ pub fn resolve_std<'cfg>(
         &specs,
         HasDevUnits::No,
         crate::core::resolver::features::ForceAllTargets::No,
+        max_rust_version,
     )?;
     Ok((
         resolve.pkg_set,
--- a/src/cargo/core/resolver/mod.rs
+++ b/src/cargo/core/resolver/mod.rs
@@ -71,6 +71,7 @@ use crate::util::config::Config;
 use crate::util::errors::CargoResult;
 use crate::util::network::PollExt;
 use crate::util::profile;
+use crate::util::PartialVersion;
 
 use self::context::Context;
 use self::dep_cache::RegistryQueryer;
@@ -138,6 +139,7 @@ pub fn resolve(
     version_prefs: &VersionPreferences,
     config: Option<&Config>,
     check_public_visible_dependencies: bool,
+    mut max_rust_version: Option<PartialVersion>,
 ) -> CargoResult<Resolve> {
     let _p = profile::start("resolving");
     let minimal_versions = match config {
@@ -148,8 +150,19 @@ pub fn resolve(
         Some(config) => config.cli_unstable().direct_minimal_versions,
         None => false,
     };
-    let mut registry =
-        RegistryQueryer::new(registry, replacements, version_prefs, minimal_versions);
+    if !config
+        .map(|c| c.cli_unstable().msrv_policy)
+        .unwrap_or(false)
+    {
+        max_rust_version = None;
+    }
+    let mut registry = RegistryQueryer::new(
+        registry,
+        replacements,
+        version_prefs,
+        minimal_versions,
+        max_rust_version,
+    );
     let cx = loop {
         let cx = Context::new(check_public_visible_dependencies);
         let cx = activate_deps_loop(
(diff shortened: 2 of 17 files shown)


Eh2406:
🙈

src/cargo/core/compiler/standard_lib.rs
```diff
@@ -153,6 +154,7 @@ pub fn resolve_std<'cfg>(
         &specs,
         HasDevUnits::No,
         crate::core::resolver::features::ForceAllTargets::No,
+        max_rust_version,
```
epage, line 157: This is one case I'm not entirely sure of but since its another unstable feature, I figure doing a best effort isn't the end of the world

src/cargo/core/resolver/mod.rs
```diff
@@ -148,8 +150,19 @@ pub fn resolve(
...
         None => false,
     };
-    let mut registry =
-        RegistryQueryer::new(registry, replacements, version_prefs, minimal_versions);
+    if !config
+        .map(|c| c.cli_unstable().msrv_policy)
+        .unwrap_or(false)
+    {
+        max_rust_version = None;
+    }
```
epage, line 158: I centralized the feature check to this one spot

I am somewhat tempted to do something similar for `honor_rust_version` and have the config initialized with it and just check the config here.

tests/testsuite/cargo_add/rust_version_ignore/mod.rs
```diff
@@ -26,7 +26,7 @@ fn case() {
         .current_dir(cwd)
         .masquerade_as_nightly_cargo(&["msrv-policy"])
         .assert()
-        .success()
+        .code(101)
```
epage, line 29: This will be undone when we get `--ignore-rust-version` working

tests/testsuite/rust_version.rs
```diff
@@ -189,6 +189,189 @@ fn rust_version_dependency_fails() {
...
+        .with_stderr(
+            "\
+[UPDATING] `dummy-registry` index
+[ERROR] failed to select a version for the requirement `bar = \"^1.0.0\"`
+candidate versions found which didn't match: 1.6.0
+location searched: `dummy-registry` index (which is replacing registry `crates-io`)
+required by package `foo v0.0.1 ([CWD])`
+perhaps a crate was updated and forgotten to be re-vendored?
+",
+        )
```
epage, line 245: Example of the bad error message

tests/testsuite/rust_version.rs
```diff
@@ -189,6 +189,189 @@ fn rust_version_dependency_fails() {
...
+    p.cargo("check --ignore-rust-version")
+        .arg("-Zmsrv-policy")
+        .masquerade_as_nightly_cargo(&["msrv-policy"])
+        // This should pick 1.6.0
+        .with_stderr(
+            "\
+[UPDATING] `dummy-registry` index
+[DOWNLOADING] crates ...
+[DOWNLOADED] bar v1.5.0 (registry `dummy-registry`)
+[CHECKING] bar v1.5.0
```
epage, line 285: Another example of `--ignore-rust-version` not working yet

epage:
> 🙈

r? @Eh2406 

;)

Eh2406:
There was discussion about modeling this as a `dependency` so that the resolver did not explicitly need to know about this new kind of requirement. What happened with that idea?

With this being part of the resolver is it possible to or is there value in adding this to proptest test generation?

src/cargo/core/compiler/standard_lib.rs
epage: I'm assuming this is build-std and we do a separate resolve pass for that.  What should we consider `max_rust_version` to be in this case?

epage:
> There was discussion about modeling this as a `dependency` so that the resolver did not explicitly need to know about this new kind of requirement. What happened with that idea?

This is exclusively meant as an unstable hack to
- Unblock people from getting this feature even if its not to the quality of stable
- Allow us to iterate on other parts of the design while working on the resolver refactor

The resolver factor to get us to the state of this being a dependency is doing to be a bit of work and I figured this can allow us to decouple parts of the process for this feature.

> With this being part of the resolver is it possible to or is there value in adding this to proptest test generation?

As this implementation is only ever intended for unstable use, I would hold off but we can call this out in the issue as remaining work in case anything changes about the plan to ensure we have it before stabilization.

Eh2406:
As a reviewer I don't quite know what to say then. I can confirm that it is an unholy hack. But, doing things properly is not in scope for this PR. The way it's implemented it should not corrupt internal data structures or cause other similar problems. What other feedback would you find useful from your reviewer?

Is there other work you would like to do before this hack gets Merged?

epage:
I am fine with it being merged as-is but with this touching the resolver, I wanted you to both review the work and get your Ok on doing this unholy hack.

Issue #10496, opened 2022-03-22
Tracking issue for `--keep-going`


dtolnay:
### Summary

Implementation: #10383
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#keep-going

This flag makes Cargo build as many crates in the dependency graph as possible, rather than aborting the build at the first one that fails to build.

### Unresolved Issues

- [ ] Add `-k` short flag? According to @joshtriplett: *"similar to `make -k`; the nature of this option is such that it'd should be easy for people to invoke"*
- [x] Reimplement https://github.com/dtolnay/trybuild using this feature, since this was the original motivating use case. See "Motivation" section of https://github.com/rust-lang/cargo/pull/10383#issue-1133692422.

### Future Extensions

- Josh suggests: *"we may want to extend this option in the future to have more kinds of "keep going" behavior, such as deferring some compile-time errors to runtime panics so that the test suite can run"*

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

ehuss:
Just a small question, I was wondering if there might be confusion on the difference between `--no-fail-fast` and `--keep-going` for `cargo test`.  There seems to be a little overlap conceptually. I'm wondering how that compares to other tools like buck or bazel?

dtolnay:
> I'm wondering how that compares to other tools like buck or bazel?

@ehuss, at least for Buck, `--no-fail-fast` is the default behavior and there is no flag to invert it &mdash; that is not a concern of the build system, but of the infrastructure which runs the build system. In other words `buck test …` will always run all tests in the targets specified. There is also no separate *build* and *run* phase as in Cargo; each test becomes runnable as soon as its executable is done building, even if other tests are not built yet.

[...]

PR #12568, opened 2023-08-26
Stabilize `--keep-going`

dtolnay:
Tracking issue with completed FCP: https://github.com/rust-lang/cargo/issues/10496

### DIFF
--- a/src/doc/man/includes/options-keep-going.md
+++ b/src/doc/man/includes/options-keep-going.md
@@ -1,5 +1,10 @@
 {{#option "`--keep-going`"}}
 Build as many crates in the dependency graph as possible, rather than aborting
-the build on the first one that fails to build. Unstable, requires
-`-Zunstable-options`.
+the build on the first one that fails to build.
+
+For example if the current package depends on dependencies `fails` and `works`,
+one of which fails to build, `cargo {{command}} -j1` may or may not build the
+one that succeeds (depending on which one of the two builds Cargo picked to run
+first), whereas `cargo {{command}} -j1 --keep-going` would definitely run both
+builds, even if the one run first fails.
 {{/option}}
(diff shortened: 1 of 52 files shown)


src/doc/man/includes/options-keep-going.md
```diff
@@ -1,5 +1,10 @@
 {{#option "`--keep-going`"}}
 Build as many crates in the dependency graph as possible, rather than aborting
-the build on the first one that fails to build. Unstable, requires
-`-Zunstable-options`.
+the build on the first one that fails to build.
+
+For example if the current package depends on dependencies `fails` and `works`,
+one of which fails to build, `cargo check -j1` may or may not build the one that
```
dtolnay, line 6: I was looking for an interpolation to use here instead of "check". `cargo {{lower actionverb}} -j1` isn't quite right because it would render as "cargo document" instead of "cargo doc", for example. Do I need to add a new `{{*set` for this or is there one I missed?

src/doc/src/reference/unstable.md
```diff
@@ -429,36 +428,6 @@ like to stabilize it somehow!
...
-While `cargo test` and `cargo bench` commands involve compilation, they do not provide a `--keep-going` flag.
-Both commands already include a similar `--no-fail-fast` flag, allowing running as many tests as possible without stopping at the first failure.
-To "compile" as many tests as possible, use target selection flags like `--tests` to build test binaries separately.
-For example,
-
-```console
-cargo build --tests --keep-going -Zunstable-options
-cargo test --tests --no-fail-fast
-```
-
```
weihanglo, line 461: We usually move this section to the bottom of this file. Maybe something like

(feel free to make it better 😆)

```markdown
### keep-going

The `--keep-going` option has been stabilized in the 1.74 release.
See the [`--keep-going` flag](../commands/cargo-build.html#option-cargo-build---keep-going) in `cargo build` as an example for more details.

src/doc/man/includes/options-keep-going.md
weihanglo: Sounds reasonable to me. Maybe `{{*set command="check"}}` and it could possibley be reused elsewhere.

dtolnay: https://github.com/rust-lang/cargo/pull/12570

dtolnay:
Some screenshots of the rendered doc:

### Unstable Features

<table><td></table>

### cargo build

<table><td></table>

### cargo check

<table><td></table>

### cargo test

<table><td></table>

weihanglo (approved):
🎉 🎉 🎉 

Thanks for driving this toward stabilization!

PR #12569, opened 2023-08-26
Update serde

ehuss:
This updates serde to the latest version (essentially reverting #12528). The current lock is preventing publishing of credential crates because they were published with a newer version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2887,9 +2887,9 @@ dependencies = [
 
 [[package]]
 name = "serde"
-version = "1.0.171"
+version = "1.0.188"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

epage:
> The current lock is preventing publishing of credential crates because they were published with a newer version.

For my own underatanding, could you expand on what problem is being run into?

ehuss:
`cargo-credential` 0.3.0 was published with a dependency on serde 1.0.180 (a few weeks ago). When attempting to publish one of the helpers, cargo pulls in the workspace `Cargo.lock`, but then fails to regenerate the lock file because serde was locked at 171, but cargo-credential 0.3.0 has a minimum requirement of 180.

epage:
So it builds fine locally because its using the local `cargo-credential` but that is unpublished so when it switches to using the registry version, it fails.  Makes sense.  Thanks for the explanation.

PR #12574, opened 2023-08-27
Improve deserialization errors of untagged enums

dtolnay:

```toml
# .cargo/config.toml

[http]
ssl-version.min = false
```

**Before:**

```console
$ cargo check
error: data did not match any variant of untagged enum SslVersionConfig
```

**After:**

```console
$ cargo check
error: error in /path/to/.cargo/config.toml: could not load config key `http.ssl-version`

Caused by:
  error in /path/to/.cargo/config.toml: `http.ssl-version.min` expected a string, but found a boolean
```

The first commit adds tests showing the pre-existing error messages &mdash; mostly just _"data did not match any variant of untagged enum T"_ with no location information. The second commit replaces all `#[derive(Deserialize)] #[serde(untagged)]` with Deserialize impls based on https://docs.rs/serde-untagged/0.1, showing the effect on the error messages.

Tested with `cargo test`, and by handwriting some bad .cargo/config.toml files and looking at the error produced by `rust-lang/cargo/target/release/cargo check`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -307,6 +307,7 @@ dependencies = [
  "same-file",
  "semver",
  "serde",
+ "serde-untagged",
  "serde-value",
  "serde_ignored",
  "serde_json",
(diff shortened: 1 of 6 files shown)


epage:
Thanks!  I'm assuming this will also offer some (slight) performance improvements based on my own profiling with `untagged`.

@bors r+

Any thoughts on adding derive support for this?

dtolnay:
Yeah we could potentially add derive support by having the call site tag the type of each of the variants.

```rust
pub enum SslVersionConfig {
    #[serde(string)]
    Single(String),
    #[serde(map)]
    Range(SslVersionConfigRange),
}
```

PR #12575, opened 2023-08-28
cargo install: suggest --git when package name is url

### LINKED ISSUE #10485: Suggest cargo install url using --git
### Problem

Currently, someone may try `cargo install https://github.com/light4/cargo-info`, then I see.
```
    Updating crates.io index
error: could not find `https://github.com/light4/cargo-info` in registry `crates-io` with version `*`
```

### Proposed Solution

It would be nice to see a suggestion to ask the user to run the same command with `--git`.

### Notes

_No response_

cardoso:

Improve the error message when specifying a URL for a package name in `cargo install`.

Fixes #10485 

Just cargo test and trying a common case like `cargo install https://github.com/rust-lang/cargo`

### Additional information

I found this PR after finishing this one: #10522
But it seems have a larger scope to refactor some of the related code.

Perhaps this one would be easier to merge and that one could focus on the refactor, otherwise sorry for the noise and feel free to close.

### DIFF
--- a/src/bin/cargo/commands/install.rs
+++ b/src/bin/cargo/commands/install.rs
@@ -121,6 +121,14 @@ pub fn exec(config: &mut Config, args: &ArgMatches) -> CliResult {
             )
             .into());
         }
+
+        if let Ok(url) = crate_name.into_url() {
+            return Err(anyhow!(
+                "invalid package name: `{url}`
+    Use `cargo install --git {url}` if you meant to install from a git repository."
+            )
+            .into());
+        }
     }
 
     let mut from_cwd = false;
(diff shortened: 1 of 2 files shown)


src/bin/cargo/commands/install.rs
```diff
@@ -121,6 +121,14 @@ pub fn exec(config: &mut Config, args: &ArgMatches) -> CliResult {
             )
             .into());
         }
+
+        if let Ok(url) = crate_name.into_url() {
```
arlosi, line 125: I'd prefer we make this a bit more specific and match only `https` or `http` `scheme`s in the URL.

Otherwise, this looks good. We don't need to cover 100% of possible git URLs for a helpful suggestion, and `https` is the most common.

cardoso: Thanks @arlosi. That makes sense. I added the check.

arlosi:
While this doesn't cover all possible git URLs, it catches the most common case and provides a helpful suggestion. Thanks!

@bors r+

src/bin/cargo/commands/install.rs
weihanglo: I think we need to cover `ssh` as it is also widely used.

Issue #11211, opened 2022-10-11
Cargo display color codes on terminals without color support


amarao:
### Problem

Some terminals lack ability to display colors. If `TERM` is set to the terminal type without color support, cargo should not put random (other terminal codes) on the screen.

### Steps

1. `TERM=t10 cargo build`

Expected result: no color codes.
Actual result: color console codes in output.

### Possible Solution(s)

Unknown terminal types should switch cargo into 'no colors' mode.

### Notes

```
infocmp t10
#	Reconstructed via infocmp from file: /usr/share/terminfo/t/t10
t10|Teleray 10 special,
	km, xhp, xt,
	cols#80, it#8, lines#24, xmc#2,
	clear=\Ej$<30/>, cub1=^H, cud1=\EB, cuf1=\EC,
	cup=\EY%p1%' '%+%c%p2%' '%+%c, cuu1=\EA, dch1=\EQ,
	dl1=\EM, ed=\EJ, el=\EK, home=\EH, ht=^I, ich1=\EP, il1=\EL,
	ind=\Eq, pad=\0, ri=\Ep, rmso=\ER@, rmul=\ER@, smso=\ERD,
	smul=\ERH,
```

### Version

```text
cargo 1.63.0 (fd9c4297c 2022-07-01)
release: 1.63.0
commit-hash: fd9c4297ccbee36d39e9a79067edab0b614edb5a
commit-date: 2022-07-01
host: x86_64-unknown-linux-gnu
libgit2: 1.4.2 (sys:0.14.2 vendored)
libcurl: 7.83.1-DEV (sys:0.4.55+curl-7.83.1 vendored ssl:OpenSSL/1.1.1n)
os: Ubuntu 22.10 (kinetic) [64-bit]
```

weihanglo:
Cargo at this moment depends on `termcolor` and `atty` to detect color support. So far as I understand it, rules are:


1. If `atty` detects the stream source is a tty, hand over to `termcolor` to determine.[^1]
     1. if `TERM=dumb` or `NO_COLOR` is set, disable the color support.  [^2]
     2. Otherwise enable it.
3. Otherwise disable the color support.

I am not an expert on tty specifications, but this seems to be more an enhancement of upstreams.

As of now, environment variable `CARGO_TERM_COLOR=never` is at your service, as well as the configuration counterpart [`term.color`](https://doc.rust-lang.org/nightly/cargo/reference/config.html#termcolor).

[...]

Issue #2290, opened 2016-01-17
some way to disable bold colors?


oconnor663:
I use the [Solarized](http://ethanschoonover.com/solarized) colorscheme in my terminal, which means that bold colors end up looking like different shades of gray. I'd like to be able to tell Cargo to use non-bold colors, or maybe be able to specify the whole palette in a config somewhere. What would be the best way to do that? I'm new to Rust, but I'm happy to try to write a PR if someone could suggest the way they'd like this to work.

alexcrichton:
Currently we don't have configuration for this (beyond turning colors off entirely), but I'd be fine adding some options to `.cargo/config` for this!

retep998:
I think having a configuration for which colors to use would be very useful.

beckclib:
@alexcrichton Similar issue here, so I will no open a new one. How would one go about adding these options? Also, would you mind reminding me how to turn colours off entirely.

Thanks!

alexcrichton:
You should be able to turn colors off entirely with `cargo foo --color never`, and you could add this by probably changing how `MultiShell` is constructed/configured in Cargo itself. I'm not exactly sure what we'd want to call all the options here and how we could have them be flexible for future configuration, however.

beckclib:
Undertood and thanks for the kind response.

oconnor663:
@beckclib I finally got around to trying this myself, but I'm very new to Rust, so by all means put up your own PR if you like.

epage:
As of #12578, cargo now has styling in multiple places.  #12655 centralizes some of this and adds semantic names to it.  #12627 proposes switching from termcolor to anstream which will make the styling API ansi escape codes on all platforms.

Note that cargo's styling has been duplicated into [cargo fmt](https://github.com/rust-lang/rustfmt/pull/5908) and soon [cargo clippy](https://github.com/rust-lang/rustfmt/pull/5908).  We'd need to make sure a solution works across these tools.

Related to `anstream` is [anstyle-git](https://docs.rs/anstyle-git/latest/anstyle_git/) and [anstyle-ls](https://docs.rs/anstyle-ls/latest/anstyle_ls/) which parse some user configuration formats into `anstyle::Style`s.

With all of that said, I don't think styling is worth it.  We use basic ANSI styles which puts the user in control already by their theme choice / customization.  See also clap-rs/clap#4132

BurntSushi:
I'm not sure if this is the right place for this comment, but this issue seems related and is still open.

[...]

PR #12578, opened 2023-08-28
feat(help): Add styling to help output 

epage:

Try to make `--help` output easier to parse by using terminal styling

Screenshots:

Screenshot from 2023-09-06 09-57-11

Screenshot from 2023-09-06 09-57-21

Screenshot from 2023-09-06 09-57-36

*(`nargo` is my shell script wrapping `cargo run --manifest-path cargo/Cargo.toml`)*

At this time, the only styling snapshotting library I know of is a pain to use, so testing this requires manually running the commands which I did.  Screenshots are included for easier evaluation of the general idea.

Snapshotting of the plain text output ensures we don't have accidental formatting regressions from this change since the formatting isn't as obvious from looking at the code.

### Additional information

Traditionally, cargo has disabled clap's styled output.  My assumed
reason is that cargo mixes custom help output with auto-generated and
you couldn't previously make it all styled.
Clap 4.2 allowed users to pass in strings styled using ANSI escape
codes, allowing us to pass in styled text that matches clap, unblocking this.  In clap
4.4.1, clap gained the ability for the user to override the style.

In this PR, I decided to use the new 4.4.1 feature to style clap's
output to match the rest of cargo's output.  Alternatively, we could use
a more subdue style that clap uses by default.

I used the `color-print` crate to allow something almost html-like for styling `&static str`.  Alternatively, we could directly embed the ANSI escape codes harder to get write, harder to inspect), or we could do the styling at runtime and enable the `string` feature in clap.

I decided to *not* style `Arg::help` messages because
- It might be distracting to have the descriptions lit up like a
  christmas tree
- It is a lot more work

The one exception I made was for `--list` since it is for a
psuedo-command (`...`) and I wanted to intentionally draw attention to
it.

#12593 made styling of `cargo -h` cleaner imo.
#12592 and #12594 were improvements I noticed while doing this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -270,6 +270,7 @@ dependencies = [
  "cargo-test-support",
  "cargo-util",
  "clap",
+ "color-print",
  "crates-io",
  "curl",
  "curl-sys",
@@ -539,6 +540,27 @@ name = "clru"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b8191fa7302e03607ff0e237d4246cc043ff5b3cb9409d995172ba3bea16b807"
+
+[[package]]
+name = "color-print"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f2a5e6504ed8648554968650feecea00557a3476bc040d0ffc33080e66b646d0"
+dependencies = [
+ "color-print-proc-macro",
+]
+
+[[package]]
+name = "color-print-proc-macro"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d51beaa537d73d2d1ff34ee70bc095f170420ab2ec5d687ecd3ec2b0d092514b"
+dependencies = [
+ "nom",
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
 
 [[package]]
 name = "colorchoice"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ cargo-test-support = { path = "crates/cargo-test-support" }
 cargo-util = { version = "0.2.6", path = "crates/cargo-util" }
 cargo_metadata = "0.17.0"
 clap = "4.4.2"
+color-print = "0.3.4"
 core-foundation = { version = "0.9.3", features = ["mac_os_10_7_support"] }
 crates-io = { version = "0.39.0", path = "crates/crates-io" }
 criterion = { version = "0.5.1", features = ["html_reports"] }
@@ -129,6 +130,7 @@ cargo-credential-libsecret.workspace = true
 cargo-credential-macos-keychain.workspace = true
 cargo-credential-wincred.workspace = true
 cargo-util.workspace = true
+color-print.workspace = true
 clap = { workspace = true, features = ["wrap_help"] }
 crates-io.workspace = true
 curl = { workspace = true, features = ["http2"] }
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -513,10 +513,23 @@ impl GlobalArgs {
 
 pub fn cli() -> Command {
     let usage = if is_rustup() {
-        "cargo [+toolchain] [OPTIONS] [COMMAND]\n       cargo [+toolchain] [OPTIONS] -Zscript <MANIFEST_RS> [ARGS]..."
+        color_print::cstr!("<cyan,bold>cargo</> <cyan>[+toolchain] [OPTIONS] [COMMAND]</>\n       <cyan,bold>cargo</> <cyan>[+toolchain] [OPTIONS]</> <cyan,bold>-Zscript</> <cyan><<MANIFEST_RS>> [ARGS]...</>")
     } else {
-        "cargo [OPTIONS] [COMMAND]\n       cargo [OPTIONS] -Zscript <MANIFEST_RS> [ARGS]..."
+        color_print::cstr!("<cyan,bold>cargo</> <cyan>[OPTIONS] [COMMAND]</>\n       <cyan,bold>cargo</> <cyan>[OPTIONS]</> <cyan,bold>-Zscript</> <cyan><<MANIFEST_RS>> [ARGS]...</>")
     };
+
+    let styles = {
+        use clap::builder::styling::*;
+        Styles::styled()
+            .header(AnsiColor::Green.on_default() | Effects::BOLD)
+            .usage(AnsiColor::Green.on_default() | Effects::BOLD)
+            .literal(AnsiColor::Cyan.on_default() | Effects::BOLD)
+            .placeholder(AnsiColor::Cyan.on_default())
+            .error(AnsiColor::Red.on_default() | Effects::BOLD)
+            .valid(AnsiColor::Cyan.on_default() | Effects::BOLD)
+            .invalid(AnsiColor::Yellow.on_default() | Effects::BOLD)
+    };
+
     Command::new("cargo")
         // Subcommands all count their args' display order independently (from 0),
         // which makes their args interspersed with global args. This puts global args last.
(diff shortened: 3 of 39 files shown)


epage:
Before applying this to the rest of cargo's commands, I figured I'd post this draft for feedback from the cargo team on (1) if we want this and (2) if we should make any tweaks

epage:
The one thing I'm not too sure of is if "literal" (bold blue) and "placeholder" (cyan) will work together in a variety of themes.  With my theme, it works well.  One workaround is to make them the same color (blue or cyan) with different intensity (bold or not) since themes would ideally treat those similarly.

weihanglo:
[zuilp discussion](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Styling.20help.20output.3F) of this PR

weihanglo (commented):
I am not entirely sure the choice of color. Do you have any reference for terminal styling convention? Should we write down Cargo's convention somewhere?

Cargo.lock
```diff
@@ -541,6 +542,27 @@ version = "0.6.1"
...
+ "color-print-proc-macro",
+]
+
+[[package]]
+name = "color-print-proc-macro"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d51beaa537d73d2d1ff34ee70bc095f170420ab2ec5d687ecd3ec2b0d092514b"
+dependencies = [
+ "nom",
```
weihanglo, line 560: Just realized cargo depends on `nom` again (via `gix`).

weihanglo (approved):
This looks nice. One thing may need to address is that we now have `termcolor`, `color-print` and `clap` handling terminal styling in different ways in Cargo. Should we have a cleanup?

Cargo.lock
epage: I've removed `nom` from gix

epage:
> I am not entirely sure the choice of color. Do you have any reference for terminal styling convention? Should we write down Cargo's convention somewhere?

Do you mean for us internally?  In most cases, it doesn't matter as its centrally defined in the clap code / `Shell` and the `color_print` code is mostly to copy the `clap` code

Do you mean publicly for others to copy?  That could be interesting.  One thought I had was to expose the style in `clap-cargo`

> This looks nice. One thing may need to address is that we now have `termcolor`, `color-print` and `clap` handling terminal styling in different ways in Cargo. Should we have a cleanup?

Don't forget `fwdansi` :)

I'm working on a bug fix to make it really solid but my plan was to migrate `cargo` from `termcolor` to `anstream` which is what `clap` uses under the hood.  Its more like `fwdansi` with less overhead so you then use it with some terminal styling crate like `color-print` for html-like formatting, `owo-colors` for a fancy API or `anstyle` for barebones styling.

If you mean centralizing the definitions of our colors, that is something that we can do.

- clap uses `anstyle`
- `anstyle-termcolor` exists for us to take a central `anstyle` definition and convert it to `termcolor` in `Shell` (until we use `anstream`)
- We could switch away from `color-print` to do runtime string formatting which would allow us to centralize the color definitions completely
  - `anstyle` isn't the most ergonomic, so we'd probably want to use `anstyle-owo-colors` or something
  - We'd need to do `cargo add clap -F string` to support runtime generated strings.

epage:
Decided to try out a variety of themes

<details>

"GNOME"

Screenshot from 2023-09-05 11-57-48

Tango

Screenshot from 2023-09-05 11-59-03

"Linux console"

Screenshot from 2023-09-05 11-59-30

"XTerm"

Screenshot from 2023-09-05 11-59-53

"rxvt"

Screenshot from 2023-09-05 12-00-18

Solarized

Screenshot from 2023-09-05 12-00-37

I've also done a run through A-E of https://github.com/Gogh-Co/Gogh (thats a bit much to include screenshots for) and "XTerm" is likely the worst.

</details>

weihanglo:
Fantastic! You've thought way more thoroughly than I have 😂

I was initially thinking of writing down our colour palette for internal use, and then moving forward by centralizing the colour definitions. It would be great to have definitions for Cargo plugin developers to follow, even if we don't announce it in public.

I see most of the approaches you mentioned needing experiments. Perhaps we should have an issue tracking the entire progress.

---

Ooof xterm is pretty bad.  Is it one of the major themes? Might be a blocker to me if it is the case.

weihanglo:
(not a blocker)

The other issue is that this patch introduces one more dependency (and transitive proc-macro). This doesn't really affect Cargo development too much. However still some cargo plugins depend on cargo directly, so we still need to keep an eye on that.

epage:
> Ooof xterm is pretty bad. Is it one of the major themes? Might be a blocker to me if it is the case.

I switched it o Cyan + Bold.  In doing so I did a double check and realized that Yellow and Red were also supposed to be Bold, so I fixed that and updated the main screenshots

For other themes:

<details>

Gnome

Screenshot from 2023-09-06 10-01-30

Tango

Screenshot from 2023-09-06 10-01-37


Linux Console
Screenshot from 2023-09-06 10-02-06



Solarized


Screenshot from 2023-09-06 10-02-20

</details>

Cargo.toml
```diff
@@ -30,6 +30,7 @@ cargo-test-support = { path = "crates/cargo-test-support" }
 cargo-util = { version = "0.2.6", path = "crates/cargo-util" }
 cargo_metadata = "0.17.0"
 clap = "4.4.2"
+color-print = "0.3.4"
```
epage, line 33: > (not a blocker)
>
> The other issue is that this patch introduces one more dependency (and transitive proc-macro). This doesn't really affect Cargo development too much. However still some cargo plugins depend on cargo directly, so we still need to keep an eye on that.

syn, quote, and proc-macro2 will already be pulled in.  I can see if I can work with the maintainer to switch to syn 2 (a gix dep also pulls in syn v1) and winnow (gix pulls in nom for now).  That leaves just the macro itself which shouldn't be all that much.  While is only used in the bin for now, I'd like to work towards styled text in more places once we are using anstream.  The main issue with color-print itself is theming.  This is also balanced out with not needing to pull in the `String` feature into clap which affects binary size, compile times, and runtime performance.

So for me, I'm not too worried about it and I can easy see the situation evolving.  To what?  I don't yet know.

epage:
> I see most of the approaches you mentioned needing experiments. Perhaps we should have an issue tracking the entire progress.

For at least part of it, I've created #12627

joshtriplett:
It feels *mildly* unfortunate that we have the concept of what "cyan" and "bold" semantically represent embedded in multiple places. This should *not* stop trying this experiment, but long-term I'm wondering how easily we might be able to introduce some compile-time semantic names for these styles, such as in a common wrapper around `cstr!`.

joshtriplett:
@epage Could you please provide sample screenshots for this color scheme on a black background, to verify that it doesn't create any hard-to-read text?

epage:
@joshtriplett see my earlier comment: https://github.com/rust-lang/cargo/pull/12578#issuecomment-1708552335

In addition to those screenshots, I've done a run through to spot check A-E of https://github.com/Gogh-Co/Gogh.

epage:
To add, all of the colors used are colors already used in cargo *except* the Cyan used in placeholders (e.g. `<command>`).  That was intentional both to both match cargo's existing colors and with the assumption that cargo's existing colors seem to be working out for us.  The Cyan for placeholders is just a variant of Cyan+Bold we already use.

joshtriplett:
@epage LGTM! Thank you.

@rfcbot reviewed

weihanglo:
Let's merge it and see how people feel about it 🚀 

@bors r+

your-diary:
How can I disable this or change the used colors?

`--color never` didn't work: `cargo --color never build --help`

epage:
Could you go into why you want to disable it? It matches existing cargo colors so we thought it'd be innocuous. It'd be good to know what problems people are having with it.

As for disabling, the cli doesn't work because help gets processed before we can process `--color`. I can't remember if I hacked in `term.color` / `CARGO_TERM_COLOR` support but that'd be the next thing id recommend. Otherwise, as an implementation detail, we do support `NO_COLOR=1` (I think there is an underscore in there).

your-diary:
@epage 

> Could you go into why you want to disable it?

Because personally I don't like the colors used.
Additionally I believe there are not small number of people who want to opt-out the colorful output entirely, taking into account how many people in the world use Rust.
How colors look to one's eyes highly depends on the person.
As personal preferences of course significantly vary from person to person, I expect the command lets users customize its appearance as much as possible.
I believe this is ***one*** of the reasons why many CUI commands have `--color={auto|never|always}` option.
Some commands like `ls` even lets users customize which colorset is used (i.e. `LS_COLORS`).

> It matches existing cargo colors

I don't think so, partly. The first impression for the current output I had was "too bright".

Actually, as you say, the color cyan has been used in Cargo. But, as far as I know, the main color is green and cyan is just temporarily displayed for highlighting purposes:

Screenshot 2023-11-18 at 22 30 37

However, in the output of `--help`, cyan is used as one of the main colors:

Screenshot 2023-11-18 at 22 31 02

This looks too bright for me.

> As for disabling, the cli doesn't work because help gets processed before we can process --color.

I see. I suspect so. I think this is not a good UI/UX; it seems at least the doc should be fixed. The current doc is this (`cargo --help` and `man cargo` respectively):

```
      --color <WHEN>            Coloring: auto, always, never
```

```
       --color when
           Control when colored output is used. Valid values:

           •  auto (default): Automatically detect if color support is available on
               the terminal.

           •  always: Always display colors.

           •  never: Never display colors.

           May also be specified with the term.color config value
           <https://doc.rust-lang.org/cargo/reference/config.html>.
```

>  I can't remember if I hacked in term.color / CARGO_TERM_COLOR support but that'd be the next thing id recommend.

Didn't work either. I'd be great if the doc would be fixed for the same reason described right above.

> Otherwise, as an implementation detail, we do support NO_COLOR=1

Worked like a charm. Thank you.
```bash
$ NO_COLOR=1 cargo build --help
```

weihanglo:
Thanks for calling them out. Some relevant contexts here about why it is hacky to support the option in CLI options and `CARGO_TERM_COLOR`:

* https://github.com/rust-lang/cargo/issues/2290
* https://github.com/rust-lang/cargo/issues/9012
* https://github.com/rust-lang/cargo/issues/11211

That said, this is a thing I would like to see it fixed.

epage:
> @epage 
> 
> > Could you go into why you want to disable it?
> 
> Because personally I don't like the colors used.

Note that we are using terminal-configured colors, rather than something like RGB, so you do have full control over how this gets rendered in your terminal.

your-diary:
@epage 
Yes. But the setting is global; it cannot (or maybe difficult if not impossible) be set on a per-command basis.

Strictly speaking, the problem is not the color itself. I feel some inconsistency in how the color is used:

> Actually, as you say, the color cyan has been used in Cargo. But, as far as I know, the main color is green and cyan is just temporarily displayed for highlighting purposes:
> 
> However, in the output of --help, cyan is used as one of the main colors:

Anyway, the essential problem is *all* of the Rust users with different personal preferences are now forced to be exposed to the colorful output.
`cargo build --color never` works. `cargo build --help --color never` does not work.

epage:
> Strictly speaking, the problem is not the color itself. I feel some inconsistency in how the color is used:

The reason I chose those colors for the help is that Green is used as more of a primary header while Cyan is a lower header / note. 

> cargo build --color never works. cargo build --help --color never does not work.

Usually a command-line flag is for transient behavior.  If this is for user preference, a CLI flag wouldn't be appropriate.

As there are related issues for this., I'd recommend we shift the focus to the relevant ones as anything discussed here will be lost track of.

your-diary:
@epage 

> Usually a command-line flag is for transient behavior. If this is for user preference, a CLI flag wouldn't be appropriate.

I don't think so. This is also a personal preference for design, I think.
I've learned 500 or more UNIX commands and many of them use CLI flags for user preferences.
And..., at least, `cargo` does support CLI flag for user preference:
```bash
function cargo() {
    if [[ $1 == 'build' ]]; then
        command cargo build --color never "$@"
    else
        command cargo "$@"
    fi
}
```

whether or not this is the way how `--color` is intended to be used.

> As there are related issues for this., I'd recommend we shift the focus to the relevant ones as anything discussed here will be lost track of.

I do agree.
I just subscribed to #9012.
As #9012 is more essential, I will suspend my discussion and won't create a new issue.
For the time being, I'll use `NO_COLOR=1`.

Thank you for the series of the replies.

PR #12581, opened 2023-08-28
refactor: Use more serde_untagged 

epage:

This is a follow up to #12574 that replaces hand-implemented Visitors with `serde_untagged` because I felt it is cleaner code

Due to an error reporting limitation in `serde_untagged`, I'm not using
this for some `MaybeWorkspace` types because it makes the errors worse.  See dtolnay/serde-untagged#2

I also held off on some config visitors because they seemed more
complicated and I didn't want to risk that code.

The main caveat is we might not have tests for every single error case.

### DIFF
--- a/src/cargo/util/interning.rs
+++ b/src/cargo/util/interning.rs
@@ -1,4 +1,5 @@
 use serde::{Serialize, Serializer};
+use serde_untagged::UntaggedEnumVisitor;
 use std::borrow::Borrow;
 use std::cmp::Ordering;
 use std::collections::HashSet;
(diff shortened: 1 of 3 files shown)


tests/testsuite/inheritable_workspace_fields.rs
```diff
@@ -1234,11 +1234,13 @@ fn error_workspace_false() {
...
 Caused by:
-  TOML parse error at line 7, column 41
+  TOML parse error at line 7, column 27
     |
   7 |             description = { workspace = false }
-    |                                         ^^^^^
+    |                           ^^^^^^^^^^^^^^^^^^^^^
   `workspace` cannot be false
+  in `workspace`
+
```
epage, line 1243: @dtolnay In this commit, I switched some explicit visitors to `serde_untagged` and it seems the error handling isn't 1-1.  This error message is basically showing that `toml` deserialized `workspace = false` and generated an error that then got turned around and plugged into `Error::custom`, causing us to wrap that `toml` error in another `toml` error but with context this time. 
- The context is for a wider span
- We then show the redundant `in workspace` which came from the original error
- We get two newlines, one for each `toml` Error

dtolnay: I am working on a fix.

epage: I've split this change out of the PR

weihanglo (approved):
This looks pretty good!

Muscraft (approved):
Looks good to me!

PR #12290, opened 2023-06-20
refactor(registry): extract and rearrange items to their own modules

weihanglo:
### What does this PR try to resolve?

A refactor moving things to their own new modules.

### How should we test and review this PR?

Commit by commit.

The PR just moves things around. I know it's a bit painful to see this kinda of change but to have a better module hierarchy we need to do it now or later.

### Additional information

This was done during coping with todos of asymmetric tokens #10519.

PR #12582, opened 2023-08-28
Set tracing target for networking messages.

ehuss:
This changes the log messages for messages related to network traffic to use the "network" tracing target. This makes it easier to do `CARGO_LOG=network=trace CARGO_HTTP_DEBUG=true` instead of trying to figure out which modules to include (and to avoid `CARGO_LOG=trace` which can be too noisy). For example, #12290 moved the location of some log messages to a different module, which broke the documented workflow of using `CARGO_LOG=cargo::ops::registry=debug` to get networking information.

### DIFF
--- a/src/cargo/core/package.rs
+++ b/src/cargo/core/package.rs
@@ -712,7 +712,7 @@ impl<'a, 'cfg> Downloads<'a, 'cfg> {
         // happen during `wait_for_download`
         let token = self.next;
         self.next += 1;
-        debug!("downloading {} as {}", id, token);
+        debug!(target: "network", "downloading {} as {}", id, token);
         assert!(self.pending_ids.insert(id));
 
         let (mut handle, _timeout) = http_handle_and_timeout(self.set.config)?;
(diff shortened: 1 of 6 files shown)


ehuss:
I'm not sure if we want to start using tracing targets, but I'm intrigued by the idea and I think it would be something interesting to explore. For example, to get fingerprint logging, the logs from `cargo_util::paths` is extremely useful, but unfortunately people normally don't include it. I'd like to explore adding additional targets like "fingerprint" for things like that.

epage:
@bors r+

This seems like a good place for us to experiment with this feature.

weihanglo (commented):
If a dependency is set with the same tracing target, will `CARGO_LOG=network` enable it also? 

If this is the case, should we have a less generic target name?

ehuss:
> If a dependency is set with the same tracing target, will `CARGO_LOG=network` enable it also?
> 
> If this is the case, should we have a less generic target name?

I believe targets are a global namespace, so I think so. But if a dependency does that, presumably the message would be related to networking and we would want it included? The idea is to show everything networking-related.

Also, it seems to be pretty rare that I've seen anyone use the target feature. I would be surprised if any low-level library is doing that.

Issue #12115, opened 2023-05-09
Tracking Issue for `[lints]` table RFC 3389


ehuss:
### Summary

RFC: [#3389](https://github.com/rust-lang/rfcs/pull/3389)
Implementation: #12148, #12168, #12174, #12174
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#lints
Issues: https://github.com/rust-lang/cargo/labels/Z-lints-table

This feature adds a `[lints]` table to `Cargo.toml` to configure reporting levels for rustc and other tool lints.

See also #3591, #5034

### Testing instructions

Requires
- nightly: nightly-2023-05-25 or newer
- ~~beta~~
- ~~stable~~

Steps:

1. Switch from attributes to `[lints]` table per the [unstable reference](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#lints)
  - `-Dwarnings` should likely still be passed in explicitly so local experimentation doesn't get blocked on perfection
  - CI jobs passing in `-Adeprecations` likely should still do so since users should see deprecation warnings locally to decide whether to address them
2. Verify lints in CI with a supported compiler by adding the `-Zlints` flag

Debugging

- Build with `--verbose` to see what order `[lints]` and other `RUSTFLAGS` are being passed to the underlying tool

Example: #12178

### Updates Since RFC

- `cargo new` will automatically inherit the lints table (#12174)
- `profile.rustflags` precedence with `[lints]` was unspecified.  We have put made it a higher precedence.
- rustflags derived from `[lints]` table are not included in `rustc -C metadata` hash.
- `[lints]` should only apply to doctests as much as a diagnostic attributes (`#[warn(...)]`) in an `.rs` file applies to doctests
- Per-target or per-target-kind (bin vs lib) `[lints]` tables is deferred (not in original RFC)
  - Workaround: using diagnostic attributes (e.g. `#[warn]` in the target)
  - This is helpful when production code (a subset of `[lib]`, `[bin]`s in a workspace) want to be held to a higher standard than development code (especially tests)
    - e.g. clippy has an [unwrap_used](https://rust-lang.github.io/rust-clippy/master/index.html#/unwrap_used) lint but people might want that in tests.  clippy works around it today by having a `allow-unwrap-in-tests` setting
  - Since workspaces can have non-production libs and bins, it almost seems like this is more of user-defined lint settings groups and inheriting specific groups.  Or put another way, more design work is needed.
[...]

PR #12584, opened 2023-08-29
fix(lints): Fail when overriding inherited lints

epage:

Overriding of inherited lints was reserved for the future but as pointed out in https://github.com/rust-lang/cargo/issues/12115#issuecomment-1695293006, we aren't failing on these when we should but silently ignoring the overrides.

This turns it into a hard error.

In fixing this, I had to add a `#[serde(expecting)]` attribute to maintain behavior on an error case (otherwise it would say "expecting struct WorkspaceLints").  Since this drew the error message to my attention, I also tweaked it to make it more specific.

Commits are broken down by the relevant tests and fixes to make the intended behavior changes obvious.

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -3499,7 +3499,7 @@ impl fmt::Debug for PathValue {
 }
 
 #[derive(Deserialize, Serialize, Debug, Clone)]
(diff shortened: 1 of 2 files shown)


arlosi:
Thanks!

@bors r+

PR #12587, opened 2023-08-29
Improve logout message for asymmetric tokens

arlosi:
When doing `cargo logout` with an asymmetric token, it currently always succeeds with no message printed.

This changes the `cargo:paseto` provider to match the other providers and respond with `NotFound` if the user is not logged in. It also adds a message to indicate that the token was successfully removed (if it existed).


epage:
@bors r+

Thanks!

PR #2551, opened 2016-04-08
Add support for `cargo --explain`

alexcrichton:
The error messages in the compiler are being tweaked and will likely drop the
`rustc --explain` part of the error message in favor of `--explain`. In that
case you're expected to basically take whatever tool you're using and pass
`--explain` to it with an error code, so let's add it to Cargo as well!

PR #12592, opened 2023-08-29
fix(help): Explain --explain

epage:
In working on #12578, I'm focusing on each help string to decide how it should be handled and I noticed this.  It feels weird to explain something in terms of another command's CLI, so I took `rustc --help`s message and added `rustc` to clarify it.

Looking back, the flag was added in #2551 with the message we have today.  Nothing seems to really be said about it.

In reflecting on this, I'm not 100% convinced and am open to other opinions.


weihanglo (approved):

weihanglo:
@bors r+

It does looks better :)

epage:
@bors r=weihanglo 

Retrying after that bors failure

PR #12593, opened 2023-08-29
fix(help): Provide better commands heading for styling

epage:
In working on #12578, I felt it would be weird to style the entire statement about commands but it also felt weird to not style it.  So this change explores an alternatively way of communicating the information.


arlosi:
Both the current and the proposed change look fine to me. I'm curious where it's going with the changed styling. Screenshots may help.

epage:
#12578 has screenshots if you expand "details".  It doesn't include this change but I'm hopeful that makes the intent clear.

ehuss:
Seems fine to me, thanks!

@bors r+

PR #12594, opened 2023-08-29
fix(help): Remove redundant information from new/init

epage:
Auditing all of the `--help` in prep for #12578 and noticed that we list the VCS information twice, once on our end and once by clap.

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -260,8 +260,7 @@ pub trait CommandExt: Sized {
             opt(
                 "vcs",
                 "Initialize a new repository for the given version \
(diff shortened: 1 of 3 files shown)


weihanglo:
Can we do rollup in Cargo 😆?

@bors r+ rollup=always

weihanglo (approved):

epage:
> Can we do rollup in Cargo 😆?

Yeah, chose to do multiple PRs for this because so its easier to deal with it if there is push back against any of these ideas

weihanglo:
Actually I mean the [**rollup** in bors](https://forge.rust-lang.org/release/rollups.html). I think we are able to do that in rust-lang/cargo. Just never saw anyone doing it.

Issue #12599, opened 2023-08-30
Cargo update -p <localcrate> in a workspace has inconsistent behaviors


roblabla:
Cargo's handling of `cargo update -p <cratename>` is **extremely** inconsistent. Depending on how local crates are named or how up-to-date the Cargo.toml is compared to the Cargo.lock file, it will behave very differently with regards to git dependencies.

I have created a repository explaining the problems I found @ https://github.com/roblabla/cargo-wtf-why-are-you-drunk/

In that repository, you'll find two examples of inconsistencies in the behavior of `cargo update -p`:

- In the first one, we show that cargo seems to handle git dependencies differently depending on whether it's got "work" to do on local crates or not. In particular, in my local monorepo, if I bump the version of a local crate in Cargo.toml, then `cargo update -p <crate>` will update the local lockfile to bump the versions of `crate` and nothing else. However, if I *don't* bump the version and run the same `cargo update -p <crate>` command, cargo will update the git dependencies.

  This discrepancy is especially frustrating, as it complicates our release process by making bumping the version more complicated.

- In the second one, we show that whether or not cargo updates the git dependencies after bumping the version seem to depend on how your local crates in your workspace are named. This makes any call to `cargo update -p <crate>` inherently fragile, as what it does will depend on how your dependencies are named.

# Possible Solutions

IMO, `cargo update -p <crate>` should never update an unrelated git dependency, as it does not match what the user intended.

# Rust version:

rust 1.71.0 (8ede3aae2 2023-07-12)

epage:
So exhibit 2 involves running `cargo update -p <workspace-member>` and a git dependency gets updated.

I ran `CARGO_LOG=trace` and did a diff of the logs.

The sad case has a long list of 
```
ignoring any lock pointing directly
```
that when only 1 exists in the happy case.

epage:
Correction, I'm assuming that difference is intended to showcase the difference in behavior.

We have a check where we only avoid a path dependency if its changed.  That "if" i what is causing the problem here.

roblabla:
> Correction, I'm assuming that difference is intended to showcase the difference in behavior.

Yeah, the trigger.sh in the happy case changes the version in the manifest, while in the sad case it doesn't, which causes the unexpected difference in behavior.

epage:
[...]

PR #12602, opened 2023-08-30
fix(resolver): Make resolver behavior independent of package order

epage:
This address one of the problems mentioned in #12599

The intent behind the `path_pkg` check is to make sure we update
workspace members in the lockfile if their version number changed.
In this case, we don't need to recursively walk, because the change
doesn't affect dependencies.  However, we also shouldn't *prevent*
recursive walks which is what we are doing today, both for packages
marked to keep and for packages that have been "poisoned".  So we fix
this by moving that call after all recursive adds are complete so as to
not interfere with them.

This should not affect `Cargo.lock` at rest, so no upgrade compatibility concerns.

This just allows more packages to be considered available to change which can prevent unclear failures.

The main case I can think of that this does something "undesirable" is when wanting to prevent another "bug" from manifesting: the updating of git dependencies when updating workspace members (#12599).  I think I'm ok with that as that needs to be looked into separately.

### DIFF
--- a/src/cargo/ops/resolve.rs
+++ b/src/cargo/ops/resolve.rs
@@ -632,23 +632,11 @@ fn register_previous_locks(
     // however, nothing else in the dependency graph depends on `log` and the
     // newer version of `serde` requires a new version of `log` it'll get pulled
     // in (as we didn't accidentally lock it to an old version).
-    //
-    // Additionally, here we process all path dependencies listed in the previous
-    // resolve. They can not only have their dependencies change but also
-    // the versions of the package change as well. If this ends up happening
-    // then we want to make sure we don't lock a package ID node that doesn't
-    // actually exist. Note that we don't do transitive visits of all the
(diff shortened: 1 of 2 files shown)


Eh2406:
It's always a little disturbing when you can remove a line of code and the tests still pass. Especially when it's documented to fix a known bug.

It is in the pile of hacks for lockfiles. I will need to load this code back into my had before I can be shore this is good.

epage:
> It's always a little disturbing when you can remove a line of code and the tests still pass. Especially when it's documented to fix a known bug.

Except this doesn't remove any code, it just ones style of processing before the other.  It does concern me a little bit but less than trying to fix the other issue in #12599.  In that one, it is removal of a line or two and that does cause tests to fail.

epage:
@rfcbot ready

Eh2406:
🤞 

@bors r+

PR #12604, opened 2023-08-30
Fewer temporary needless strings

Eh2406:

I noticed a few places where we were using `InternedString::to_string()` in places where we just needed a hash key or to display it. In these situations we could avoid a memory allocation by using the `InternedString` directly. I used the clippy `disallowed-methods` to look through all of the places who called `to_string` on an `InternedString`, and fixed all the places where it was a straightforward change to remove it. I don't think any of them matter in practice. But doing less work can't hurt.

This was an internal re-factor, and the tests still pass.

### DIFF
--- a/credential/cargo-credential/src/error.rs
+++ b/credential/cargo-credential/src/error.rs
@@ -42,9 +42,9 @@ pub enum Error {
 }
 
 impl From<String> for Error {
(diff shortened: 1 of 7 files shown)


epage:
Thanks!

@bors r+

weihanglo (approved):

Issue #12585, opened 2023-08-29
Target tab completion invokes `rustup`


mkoncek:
### Problem

Invoking `cargo build --target <TAB>` invokes `rustup`. Some environments may not have it installed and I do not see why it should be invoked if the tab completion can run equivalent of `rustc --print target-list`.

### Steps

1. Set up a basic Rust project with cargo
2. Write `cargo build --target ` on terminal and then press the TAB key.
3. If `rustup` is not installed, terminal will print:
`cargo build --target bash: rustup: command not found`

### Possible Solution(s)

I suppose `cargo` could return the output of `rustc --print target-list`

### Notes

It is generally a good idea to depend only on the most basic tools, especially if they get the job done right, like `rustc` does.

### Version

```text
cargo 1.71.0
release: 1.71.0
host: x86_64-unknown-linux-gnu
libgit2: 1.6.4 (sys:0.17.1 system)
libcurl: 8.0.1 (sys:0.4.61+curl-8.0.1 system ssl:OpenSSL/3.0.9)
os: Fedora 38.0.0 [64-bit]
```
```

epage:
Interesting, the zsh completions are using `rustc --print target-list` but the bash ones aren't.

zsh support for targets was added in #8740

bash support was added in #6640

I don't see any references to why different strategies were taken, so it seems like we could consolidate down to just `rustc`

weihanglo:
Another idea is we can go with `rustup` way first and fall back to `rustc --print target-list`.

epage:
Is there a reason to prefer rustup?  I guess so you can check which targets are actually installed?

weihanglo:
Yes. Completing with installed ones. `rustc --print target-list` now prints 218 target triples which doesn't sounds too friendly.

Angelin01:
I can probably whip this up by tomorrow, already have the bash one ready, just need to do some testing with the ZSH one.

PR #12607, opened 2023-08-31
fix(cli): Help users know possible `--target` values

epage:

I was needing to do some more cross-compilation and forgot what the target triple was that I needed to run.  I realized i had to re-remember the command yet again.   Especially with #12585 in memory, I realized that `--target` isn't working like `--package` and other arguments that can report supported values.

In working on it, I realized I probably didn't want to report supported values yet out of concern for how big the list is (see also #12585), so I decided to just list the relevant commands for now.  We *might* be able to parse the rustup output to report those targets but I didn't receive a glowing endorsement from the rustup team about parsing the list (more of "yes, hyrums law and at least its interactive rather than CI").

Before:
```
error: a value is required for '--target <TRIPLE>' but none was supplied

```

After:
```
error: "--target" takes a target architecture as an argument.

Run `rustup target list` to see possible targets.
```
(quotes were used because the other "list available" use them, we should probably work to be uniform in how we quote)

First PR adds a test showing the existing output and then through the rest you can see how the output changed

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -440,8 +440,8 @@ pub trait ArgMatchesExt {
         self.maybe_flag("keep-going")
     }
 
-    fn targets(&self) -> Vec<String> {
-        self._values_of("target")
+    fn targets(&self) -> CargoResult<Vec<String>> {
+        Ok(self._values_of("target"))
     }
 
     fn get_profile_name(
@@ -590,7 +590,7 @@ pub trait ArgMatchesExt {
             config,
             self.jobs()?,
             self.keep_going(),
-            &self.targets(),
+            &self.targets()?,
             mode,
         )?;
         build_config.message_format = message_format.unwrap_or(MessageFormat::Human);
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -218,7 +218,10 @@ pub trait CommandExt: Sized {
     }
 
     fn arg_target_triple(self, target: &'static str) -> Self {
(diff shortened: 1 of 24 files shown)


tests/testsuite/cargo_build/help/stdout.log
```diff
@@ -42,7 +42,7 @@ Compilation Options:
       --profile <PROFILE-NAME>  Build artifacts with the specified profile
   -j, --jobs <N>                Number of parallel jobs, defaults to # of CPUs.
       --keep-going              Do not abort the build as soon as there is an error
-      --target <TRIPLE>         Build for the target triple
+      --target [<TRIPLE>]       Build for the target triple
```
weihanglo, line 45: I wish we could have a way to avoid this showing as optional 🥲.

(Never mind we already have this issue everywhere.)

src/cargo/util/command_prelude.rs
```diff
@@ -440,8 +444,20 @@ pub trait ArgMatchesExt {
...
     }
 
-    fn targets(&self) -> Vec<String> {
-        self._values_of("target")
+    fn targets(&self) -> CargoResult<Vec<String>> {
+        if self.is_present_with_zero_values("target") {
+            let cmd = if is_rustup() {
+                "rustup target list"
+            } else {
+                "ustc --print target-list"
```
weihanglo, line 452: Not a real objection. Just this recommendation could often be false when using `+channel` command line override.

Any typo

```suggestion
                "rustc --print target-list"
```

epage: Typo fixed.  I can go either way in how we handle the recommendation.

weihanglo (approved):

PR #12614, opened 2023-09-01
feat(pkgid): Allow incomplete versions when unambigious

epage:

This was proposed in #12425 to help sand off some of the rough edges around `cargo update` for its wider use it would be getting.  Its easy to accidentally get duplicate copies packages in a repo and a pain to have to specify the full version when `cargo update -p foo@1` is sufficient to describe it.

Other effects
- profile overrides also supports this since we already allow a spec to match multiple items
- `cargo clean -p foo@...` already ignored the version, so now we also parse and ignore the partial version
- `cargo tree --prune` will now accept partial versions and will match all of them

Parts not effected:
- Replacements
  - Two of the cases were found and we treat it as if the version isn't present which will error, so I think that is correct

This extracts `PartialVersion` from `RustVersion` where `RustVersion` is a more specialized variant, not allowing prerelease or build.

This works by adopting `PartialVersion` into `PackageIdSpec`.  For `PackageIdSpec::query`, this will "just work".

### DIFF
--- a/src/cargo/core/package_id_spec.rs
+++ b/src/cargo/core/package_id_spec.rs
@@ -10,7 +10,8 @@ use crate::core::PackageId;
 use crate::util::edit_distance;
 use crate::util::errors::CargoResult;
 use crate::util::interning::InternedString;
-use crate::util::{validate_package_name, IntoUrl, ToSemver};
+use crate::util::PartialVersion;
+use crate::util::{validate_package_name, IntoUrl};
 
 /// Some or all of the data required to identify a package:
 ///
@@ -24,7 +25,7 @@ use crate::util::{validate_package_name, IntoUrl, ToSemver};
 #[derive(Clone, PartialEq, Eq, Debug, Hash, Ord, PartialOrd)]
 pub struct PackageIdSpec {
     name: InternedString,
-    version: Option<Version>,
+    version: Option<PartialVersion>,
     url: Option<Url>,
 }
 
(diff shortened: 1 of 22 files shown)


epage:

The main deviation from the proposal in #12425 is that I made the change to `PackageIdSpec` and it applies everywhere that does.  I'd recommend checking the documentation and test changes in 53f80bf before checking your box to make sure you are on board with it.  The one part I'm least confident in is that the resolver uses `PackageIdSpec`, rather than `PackageId`, in some places.

@rfcbot fcp merge

weihanglo (approved):

weihanglo:
I think this is fairly a small but useful change not required to wait for the entire FCP. Let's merge it.

@bors r+

src/cargo/core/package_id_spec.rs
```diff
@@ -170,7 +176,8 @@ impl PackageIdSpec {
         }
 
         if let Some(ref v) = self.version {
-            if v != package_id.version() {
+            let req = v.exact_req();
+            if !req.matches(package_id.version()) {
```
epage, line 180: This causes us to no longer require the `build` field matches which causes problems.  Working on a fix which we'll need to backport.

src/cargo/util/semver_ext.rs
```diff
@@ -171,12 +171,12 @@ impl std::str::FromStr for PartialVersion {
...
             Ok(req) if value.chars().all(|c| c.is_ascii_digit() || c == '.') => req,
-            _ if value.contains('+') => {
-                anyhow::bail!("unexpected build field, expected a version like \"1.32\"")
-            }
             _ if value.contains('-') => {
                 anyhow::bail!("unexpected prerelease field, expected a version like \"1.32\"")
             }
+            _ if value.contains('+') => {
+                anyhow::bail!("unexpected build field, expected a version like \"1.32\"")
+            }
```
Eh2406, line 262: Do we need to check for this? `VersionReq` cannot parse requirements that have build metadata? And why are we parsing the string to determine what kind of requirement we got and not looking at the `req` we matched on?

src/cargo/core/package_id_spec.rs
epage: See #12806

PR #12615, opened 2023-09-01
use split_once for cleaner code

Eh2406:

Search the code base for `.splitn(2` and replace with `.split_once` where it was clearer. I don't think any of them matter in practice.

This was an internal re-factor, and the tests still pass.
The two methods have subtly different semantics, so please review carefully.

### DIFF
--- a/crates/mdman/src/main.rs
+++ b/crates/mdman/src/main.rs
@@ -98,15 +98,16 @@ fn process_args() -> Result<Options, Error> {
                 let man = args
                     .next()
                     .ok_or_else(|| format_err!("--man requires a value"))?;
-                let parts: Vec<_> = man.splitn(2, '=').collect();
-                let key_parts: Vec<_> = parts[0].splitn(2, ':').collect();
-                if parts.len() != 2 || key_parts.len() != 2 {
-                    bail!("--man expected value with form name:1=link");
-                }
-                let section: u8 = key_parts[1].parse().with_context(|| {
-                    format!("expected unsigned integer for section, got `{}`", parts[1])
+                let parts = man.split_once('=').ok_or_else(|| {
+                    anyhow::format_err!("--man expected value with form name:1=link")
                 })?;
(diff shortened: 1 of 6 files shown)


src/cargo/core/package_id_spec.rs
```diff
@@ -123,24 +123,20 @@ impl PackageIdSpec {
...
-                        }
-                        None => {
-                            if name_or_version.chars().next().unwrap().is_alphabetic() {
-                                (InternedString::new(name_or_version), None)
-                            } else {
-                                let version = name_or_version.to_semver()?;
-                                (InternedString::new(path_name), Some(version))
-                            }
+                Some(fragment) => match fragment.split_once([':', '@']) {
+                    Some((name_or_version, part)) => {
```
epage, line 127: I believe `name_or_version` is only `name` here

epage (approved):

epage:
Feel free to `r=` me whether you decide to resolve the nit or not

crates/mdman/src/main.rs
```diff
@@ -98,15 +98,16 @@ fn process_args() -> Result<Options, Error> {
                 let man = args
                     .next()
                     .ok_or_else(|| format_err!("--man requires a value"))?;
-                let parts: Vec<_> = man.splitn(2, '=').collect();
-                let key_parts: Vec<_> = parts[0].splitn(2, ':').collect();
-                if parts.len() != 2 || key_parts.len() != 2 {
-                    bail!("--man expected value with form name:1=link");
-                }
```
Eh2406, line 105: The old code errored on having too many `=` or `:` which the new code no longer does. Do we care?

src/cargo/core/package_id_spec.rs
Eh2406: Done.

crates/mdman/src/main.rs
weihanglo: I don't fully understand the format but we can keep the behavior the same for this refactor

weihanglo (approved):

PR #12622, opened 2023-09-05
cargo-credential: change serialization of cache expiration

arlosi:
Serde has [multiple options](https://serde.rs/enum-representations.html) for serialization of enum types. The default is to use the field name in a map (for variants with data), or a string (for variants without data). This causes forward compatibility problems when switching between these two cases. To avoid this, all `enum`s within `cargo-credential` used the "internally tagged" approach, but `CacheControl` did not.

* Changes `CacheControl` to be internally tagged and `flattened` within `CredentialResponse::Get`
* Adds forward compatibility tests to ensure adding additional fields will not break deserialization

Within a credential response, this is the change:
```diff
- "cache":{"expires":1684251794},
+ "cache":"expires", "expiration":1684251794,
```
Other variants, such as `"cache":"session"` remain the same.

## How to review
This PR contains two commits, one for the breaking change for `CacheControl` serialization, and one non-breaking change that makes several other fields skipped if none/empty.

r? @ehuss

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -333,7 +333,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-credential"
-version = "0.3.1"
+version = "0.4.0"
 dependencies = [
  "anyhow",
  "libc",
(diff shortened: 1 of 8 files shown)


ehuss:
Thanks!

@bors r+

PR #12623, opened 2023-09-05
Add MSRV validation GitHub Action for cargo-credential

arlosi:
`cargo-credential` should have a separate MSRV from the rest of Cargo that is more relaxed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,6 +20,7 @@ jobs:
     needs:
     - build_std
     - clippy
+    - credential_msrv
     - docs
     - lockfile
     - resolver
@@ -37,6 +38,7 @@ jobs:
     needs:
     - build_std
     - clippy
+    - credential_msrv
     - docs
     - lockfile
     - resolver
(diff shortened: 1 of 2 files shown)


epage:
FYI I'm working on features for `cargo-hack` so you can run `cargo hack check --rust-version` and it will verify MSRV for all crates individually to help the scaling problem.

epage:
(don't feel like you need to wait on this work for this PR)

.github/workflows/main.yml
```diff
@@ -246,3 +246,10 @@ jobs:
         cd target
         curl -sSLO https://raw.githubusercontent.com/rust-lang/rust/master/src/tools/linkchecker/linkcheck.sh
         sh linkcheck.sh --all --path ../src/doc cargo
+
+  credential_msrv:
```
weihanglo, line 250: Should we make it a required job?

(I am brain-fogged now sorry if this doesn't make sense)

arlosi: Yes, thanks!

weihanglo (approved):

weihanglo:
Looks good. Thanks!

@bors r+

PR #12625, opened 2023-09-06
feat: implement RFC 3127 `-Ztrim-paths`

weihanglo:

Implement [RFC 3127 trim paths](https://rust-lang.github.io/rfcs/3127-trim-paths.html) on Cargo side. The counterpart of the compiler is in <https://github.com/rust-lang/rust/pull/115214>. The tracking issue for Cargo is <https://github.com/rust-lang/rust/issues/111540>.

By commits. I would recommend reading the doc first to get the an overview on this.

#### Things I am uncertain

- [x] The correct path prefix for sysroot remap. See https://github.com/rust-lang/cargo/pull/12625#discussion_r1371044041.
    - @Urgau has pointed out the remapping of sysroot has already been done by rustc bootstrap [here](https://github.com/rust-lang/rust/blob/c2ef35161fc7477b38f2e556be2fd6d85d9f4905/src/bootstrap/src/lib.rs#L1107-L1119). Changing that may require `-Ztrim-paths` being stabilized, and may leak CI details into all pre-built std, core…. This is something requiring to collaborate with `T-bootstrap`. We'll postpone it for this pull request.
- [x] Do we want to verify the final artifact is actually sanitized? See https://github.com/rust-lang/cargo/pull/12625#discussion_r1371087973
    - Current we're satisfied with testing only on Linux.
    - For other platforms we verify Cargo passing in correct arguments to rustc.

#### Things to decide (but can postpone to follow-up PRs)

- Is remapping dependencies to `<pkg>-<version>` good enough for all source kinds?
    - How to tell if a package is from Git or registry? IDE may wants that information?
    - Should we use `SourceId` hash or something instead?
- The remap of workspace was not clear in the RFC. This PR choses to remap from workspace root instead. See https://github.com/rust-lang/cargo/pull/12625#discussion_r1371066017

### Additional information

Things to do:

- [x] Support `trim-paths = "all"` and `trim-paths = "none"` but not `trim-paths = ["all", "none"]`.
- [x] Support `trim-paths = true/false`.
- [x] Default value for `dev` and `release` profile.
- [ ] Set `CARGO_TRIM_PATHS` environment variable for build scripts. (plan to do in follow-up)
- [ ] Fingerprint should take trim-paths settings into account. (plan to do in follow-up)
- [ ] Make `--remap-path-prefix` works with `rustdoc` invocations (which `rustdoc` need an update)

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -94,6 +94,7 @@ use crate::util::errors::{CargoResult, VerboseError};
 use crate::util::interning::InternedString;
 use crate::util::machine_message::{self, Message};
 use crate::util::toml::TomlDebugInfo;
+use crate::util::toml::TomlTrimPaths;
 use crate::util::{add_path_args, internal, iter_join_onto, profile};
 use cargo_util::{paths, ProcessBuilder, ProcessError};
 use rustfix::diagnostics::Applicability;
@@ -950,6 +951,7 @@ fn build_base_args(cx: &Context<'_, '_>, cmd: &mut ProcessBuilder, unit: &Unit) 
         incremental,
         strip,
         rustflags: profile_rustflags,
+        trim_paths,
         ..
     } = unit.profile.clone();
     let test = unit.mode.is_any_test();
@@ -1026,6 +1028,10 @@ fn build_base_args(cx: &Context<'_, '_>, cmd: &mut ProcessBuilder, unit: &Unit) 
                 cmd.arg("-C").arg(format!("split-debuginfo={split}"));
             }
         }
+    }
+
+    if let Some(trim_paths) = trim_paths {
+        trim_paths_args(cmd, cx, unit, &trim_paths)?;
     }
 
     cmd.args(unit.pkg.manifest().lint_rustflags());
@@ -1160,6 +1166,74 @@ fn features_args(unit: &Unit) -> Vec<OsString> {
     }
 
     args
+}
+
+/// Generates the `--remap-path-scope` and `--remap-path-prefix` for [RFC 3127].
+/// See also unstable feature [`-Ztrim-paths`].
+///
+/// [RFC 3127]: https://rust-lang.github.io/rfcs/3127-trim-paths.html
+/// [`-Ztrim-paths`]: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#profile-trim-paths-option
+fn trim_paths_args(
+    cmd: &mut ProcessBuilder,
+    cx: &Context<'_, '_>,
+    unit: &Unit,
+    trim_paths: &TomlTrimPaths,
+) -> CargoResult<()> {
+    if trim_paths.is_none() {
+        return Ok(());
+    }
+
+    // feature gate was checked during mainfest/config parsing.
+    cmd.arg("-Zunstable-options");
+    cmd.arg(format!("-Zremap-path-scope={trim_paths}"));
+
+    let sysroot_remap = {
+        let sysroot = &cx.bcx.target_data.info(unit.kind).sysroot;
+        let mut remap = OsString::from("--remap-path-prefix=");
+        remap.push(sysroot);
+        remap.push("/lib/rustlib/src/rust"); // See also `detect_sysroot_src_path()`.
+        remap.push("=");
+        remap.push("/rustc/");
+        // This remap logic aligns with rustc:
+        // <https://github.com/rust-lang/rust/blob/c2ef3516/src/bootstrap/src/lib.rs#L1113-L1116>
+        if let Some(commit_hash) = cx.bcx.rustc().commit_hash.as_ref() {
+            remap.push(commit_hash);
+        } else {
+            remap.push(cx.bcx.rustc().version.to_string());
+        }
+        remap
+    };
+    cmd.arg(sysroot_remap);
+
+    let package_remap = {
+        let pkg_root = unit.pkg.root();
+        let ws_root = cx.bcx.ws.root();
+        let is_local = unit.pkg.package_id().source_id().is_path();
+        let mut remap = OsString::from("--remap-path-prefix=");
+        // Remapped to path relative to workspace root:
+        //
+        // * path dependencies under workspace root directory
+        //
+        // Remapped to `<pkg>-<version>`
+        //
+        // * registry dependencies
+        // * git dependencies
+        // * path dependencies outside workspace root directory
+        if is_local && pkg_root.strip_prefix(ws_root).is_ok() {
+            remap.push(ws_root);
+            remap.push("="); // empty to remap to relative paths.
+        } else {
+            remap.push(pkg_root);
+            remap.push("=");
+            remap.push(unit.pkg.name());
+            remap.push("-");
+            remap.push(unit.pkg.version().to_string());
+        }
+        remap
+    };
+    cmd.arg(package_remap);
+
+    Ok(())
 }
 
 /// Generates the `--check-cfg` arguments for the `unit`.
--- a/src/cargo/core/profiles.rs
+++ b/src/cargo/core/profiles.rs
@@ -24,9 +24,11 @@
 use crate::core::compiler::{CompileKind, CompileTarget, Unit};
 use crate::core::dependency::Artifact;
 use crate::core::resolver::features::FeaturesFor;
+use crate::core::Feature;
 use crate::core::{PackageId, PackageIdSpec, Resolve, Shell, Target, Workspace};
 use crate::util::interning::InternedString;
 use crate::util::toml::TomlTrimPaths;
+use crate::util::toml::TomlTrimPathsValue;
 use crate::util::toml::{
     ProfilePackageSpec, StringOrBool, TomlDebugInfo, TomlProfile, TomlProfiles,
 };
@@ -81,7 +83,9 @@ impl Profiles {
             rustc_host,
         };
 
-        Self::add_root_profiles(&mut profile_makers, &profiles);
+        let trim_paths_enabled = ws.unstable_features().is_enabled(Feature::trim_paths())
+            || config.cli_unstable().trim_paths;
+        Self::add_root_profiles(&mut profile_makers, &profiles, trim_paths_enabled);
 
         // Merge with predefined profiles.
         use std::collections::btree_map::Entry;
@@ -124,6 +128,7 @@ impl Profiles {
     fn add_root_profiles(
         profile_makers: &mut Profiles,
         profiles: &BTreeMap<InternedString, TomlProfile>,
+        trim_paths_enabled: bool,
     ) {
         profile_makers.by_name.insert(
             InternedString::new("dev"),
@@ -132,7 +137,10 @@ impl Profiles {
 
         profile_makers.by_name.insert(
             InternedString::new("release"),
-            ProfileMaker::new(Profile::default_release(), profiles.get("release").cloned()),
+            ProfileMaker::new(
+                Profile::default_release(trim_paths_enabled),
+                profiles.get("release").cloned(),
+            ),
         );
     }
 
@@ -636,7 +644,7 @@ compact_debug! {
         fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
             let (default, default_name) = match self.name.as_str() {
                 "dev" => (Profile::default_dev(), "default_dev()"),
-                "release" => (Profile::default_release(), "default_release()"),
+                "release" => (Profile::default_release(false), "default_release()"),
                 _ => (Profile::default(), "default()"),
             };
             [debug_the_fields(
@@ -697,11 +705,13 @@ impl Profile {
     }
 
     /// Returns a built-in `release` profile.
-    fn default_release() -> Profile {
+    fn default_release(trim_paths_enabled: bool) -> Profile {
+        let trim_paths = trim_paths_enabled.then(|| TomlTrimPathsValue::Object.into());
         Profile {
             name: InternedString::new("release"),
             root: ProfileRoot::Release,
             opt_level: InternedString::new("3"),
+            trim_paths,
             ..Profile::default()
         }
     }
--- a/src/cargo/core/profiles.rs
+++ b/src/cargo/core/profiles.rs
@@ -26,6 +26,7 @@ use crate::core::dependency::Artifact;
 use crate::core::resolver::features::FeaturesFor;
 use crate::core::{PackageId, PackageIdSpec, Resolve, Shell, Target, Workspace};
 use crate::util::interning::InternedString;
+use crate::util::toml::TomlTrimPaths;
 use crate::util::toml::{
     ProfilePackageSpec, StringOrBool, TomlDebugInfo, TomlProfile, TomlProfiles,
 };
@@ -556,6 +557,9 @@ fn merge_profile(profile: &mut Profile, toml: &TomlProfile) {
     if let Some(flags) = &toml.rustflags {
         profile.rustflags = flags.iter().map(InternedString::from).collect();
     }
+    if let Some(trim_paths) = &toml.trim_paths {
+        profile.trim_paths = Some(trim_paths.clone());
+    }
     profile.strip = match toml.strip {
         Some(StringOrBool::Bool(true)) => Strip::Named(InternedString::new("symbols")),
         None | Some(StringOrBool::Bool(false)) => Strip::None,
@@ -599,6 +603,9 @@ pub struct Profile {
     #[serde(skip_serializing_if = "Vec::is_empty")] // remove when `rustflags` is stablized
     // Note that `rustflags` is used for the cargo-feature `profile_rustflags`
     pub rustflags: Vec<InternedString>,
+    // remove when `-Ztrim-paths` is stablized
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub trim_paths: Option<TomlTrimPaths>,
 }
 
 impl Default for Profile {
@@ -619,6 +626,7 @@ impl Default for Profile {
             panic: PanicStrategy::Unwind,
             strip: Strip::None,
             rustflags: vec![],
+            trim_paths: None,
         }
     }
 }
@@ -647,6 +655,7 @@ compact_debug! {
                 panic
                 strip
                 rustflags
+                trim_paths
             )]
         }
     }
@@ -713,6 +722,7 @@ impl Profile {
             self.rpath,
             (self.incremental, self.panic, self.strip),
             &self.rustflags,
+            &self.trim_paths,
         )
     }
 }
--- a/src/cargo/util/rustc.rs
+++ b/src/cargo/util/rustc.rs
@@ -28,6 +28,8 @@ pub struct Rustc {
     pub version: semver::Version,
     /// The host triple (arch-platform-OS), this comes from verbose_version.
     pub host: InternedString,
+    /// The rustc full commit hash, this comes from `verbose_version`.
+    pub commit_hash: Option<String>,
     cache: Mutex<Cache>,
 }
 
@@ -80,6 +82,17 @@ impl Rustc {
                 verbose_version
             )
         })?;
+        let commit_hash = extract("commit-hash: ").ok().map(|hash| {
+            debug_assert!(
+                hash.chars().all(|ch| ch.is_ascii_hexdigit()),
+                "commit hash must be a hex string"
+            );
+            debug_assert!(
+                hash.len() == 40 || hash.len() == 64,
+                "hex string must be generated from sha1 or sha256"
+            );
+            hash.to_string()
+        });
 
         Ok(Rustc {
             path,
@@ -88,6 +101,7 @@ impl Rustc {
             verbose_version,
             version,
             host,
+            commit_hash,
             cache: Mutex::new(cache),
         })
     }
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -2692,6 +2692,8 @@ pub struct TomlProfile {
     // requires all non-tables to be listed first.
     pub package: Option<BTreeMap<ProfilePackageSpec, TomlProfile>>,
     pub build_override: Option<Box<TomlProfile>>,
+    /// Unstable feature `-Ztrim-paths`.
+    pub trim_paths: Option<TomlTrimPaths>,
 }
 
 impl TomlProfile {
@@ -2887,6 +2889,15 @@ impl TomlProfile {
             match (
                 features.require(Feature::profile_rustflags()),
                 cli_unstable.profile_rustflags,
+            ) {
+                (Err(e), false) => return Err(e),
+                _ => {}
+            }
+        }
+        if self.trim_paths.is_some() {
+            match (
+                features.require(Feature::trim_paths()),
+                cli_unstable.trim_paths,
             ) {
                 (Err(e), false) => return Err(e),
                 _ => {}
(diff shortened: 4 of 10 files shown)


src/cargo/core/compiler/mod.rs
```diff
@@ -1166,6 +1172,62 @@ fn features_args(unit: &Unit) -> Vec<OsString> {
...
+///
+/// [RFC 3127]: https://rust-lang.github.io/rfcs/3127-trim-paths.html
+/// [`-Ztrim-paths`]: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#profile-trim-paths-option
+fn trim_paths_args(
+    cmd: &mut ProcessBuilder,
+    cx: &Context<'_, '_>,
+    unit: &Unit,
+    scopes: &[ProfileTrimPaths],
+) -> CargoResult<()> {
+    if scopes.iter().all(|s| matches!(s, ProfileTrimPaths::None)) {
```
Urgau, line 1186: Shouldn't this be `any` (instead of `all`)?
Otherwise you will end-up with the invalid scope: `none` from `ProfileTrimPaths::as_str`

src/doc/src/reference/unstable.md
```diff
@@ -1665,6 +1666,96 @@ version = "0.1.0"
...
+It takes an array of the following values:
+
+- `"none"` and `false` --- disable path sanitisation
+- `"macro"` --- sanitise paths in the expansion of `std::file!()` macro.
+    This is where paths in embedded panic messages come from
+- `"diagnostics"` --- sanitise paths in printed compiler diagnostics
+- `"object"` --- sanitise paths in compiled executables or libraries
+- `"all"` and `true` --- sanitise paths in all possible locations
+
+It is defaulted to `none` for the `dev` profile, and `object` for the `release` profile.
```
Urgau, line 1715: I don't think this was implemented, right?

weihanglo: Thank you. It's a bit awkward to implement, as `dev` and `release profiles are kinda hard-coded. And we need to take care of both route of `Cargo.toml` and `.cargo/config.toml`. Anyway, this will be done before ready for review.

src/cargo/core/compiler/mod.rs
weihanglo: Nice catch! Fixed.

ehuss:
Do you want a review on this while it is in draft status?

weihanglo:
I would wait for rustc side being reviewed.

tests/testsuite/profile_trim_paths.rs
```diff
@@ -0,0 +1,406 @@
...
+        )
+        .with_stderr_contains(
+            &format!("\
+[RUNNING] [..]rustc [..]-Zremap-path-scope=diagnostics --remap-path-prefix=[..]/lib/rustlib/src/rust=/rustc/[..] --remap-path-prefix={pkg_remap} [..]",
+            ))
+        .with_stderr_contains("\
+[RUNNING] [..]rustc [..]-Zremap-path-scope=diagnostics --remap-path-prefix=[..]/lib/rustlib/src/rust=/rustc/[..] --remap-path-prefix=[CWD]= [..]",
+        )
+        .run();
+}
```
weihanglo, line 406: I'd like to set up a smoke test against debug symbol, so that we can make sure they are sanitized. I can follow [what rustc does](https://github.com/Urgau/rust/blob/8501f1c7ba440c97833ea9dec03965b4d5ef2fd7/tests/run-make/split-debuginfo/Makefile#L114-L121) by using `objdump`. Is there any better way to verify it? What about cross-platform tests?

weihanglo:
This is ready for review. Not 100% complete but good for shipping as a nightly feature.

src/cargo/core/compiler/mod.rs
```diff
@@ -1162,6 +1168,69 @@ fn features_args(unit: &Unit) -> Vec<OsString> {
...
+    // feature gate was checked during mainfest/config parsing.
+    cmd.arg("-Zunstable-options");
+    cmd.arg(format!("-Zremap-path-scope={trim_paths}"));
+
+    let sysroot_remap = {
+        let sysroot = &cx.bcx.target_data.info(unit.kind).sysroot_target_libdir;
+        let commit_hash = &cx.bcx.rustc().commit_hash;
+        let mut remap = OsString::from("--remap-path-prefix=");
+        // TODO(trim-paths): what is the correct prefix for sysroot? does rustup affect this?
+        remap.push(sysroot);
```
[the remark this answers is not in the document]
Urgau: > What's the correct remap prefix for sysroot

[`rustc --print=sysroot`](https://doc.rust-lang.org/rustc/command-line-arguments.html#--print-print-compiler-information)

> and how to setup a test case for it?

Well you could have a Rust `fn main() { panic!("something"); }` with `RUST_BACKTRACE=1` and see the resulting backtrace and assert that it doesn't contain the un-remapped sysroot.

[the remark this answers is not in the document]
weihanglo: Thanks for the tip!

Unfortunately I've tried `panic!` backtrace and it has already remapped even without `-Ztrim-paths` 😞.

```
   0: std::panicking::begin_panic
             at /rustc/cd674d61790607dfb6faa9d754bd3adfa13aea7c/library/std/src/panicking.rs:638:12
```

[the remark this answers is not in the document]
weihanglo: The desired behavior is mentioned in the RFC

* https://rust-lang.github.io/rfcs/3127-trim-paths.html#handling-sysroot-paths
* https://rust-lang.github.io/rfcs/3127-trim-paths.html#changing-handling-of-sysroot-path-in-rustc

To follow the RFC, yes, we might want to disable it in bootstrap.

tests/testsuite/profile_trim_paths.rs
weihanglo: Ended up using `readelf`. Now we have covered Linux for `object` option. I think that's pretty acceptable at this stage. We can add more tests for macOS and Windows later.

src/cargo/core/compiler/mod.rs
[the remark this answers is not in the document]
weihanglo: Me and Urgau had a chat on this. Urgau pointed out changing the bootstrap this may require `-Ztrim-paths` being stabilized, and may leak CI details into all pre-built std, core…. This is something requiring to collaborate with T-bootstrap. We'll postpone it for this pull request.

src/doc/src/reference/unstable.md
```diff
@@ -1289,6 +1290,96 @@ edition that may break your build.
...
+
+This will not affect any hard-coded paths in the source code, such as in strings.
+
+#### Environment variable
+
+*as a new entry of ["Environment variables Cargo sets for build scripts"](./environment-variables.md#environment-variables-cargo-sets-for-crates)*
+
+* `CARGO_TRIM_PATHS` --- The value of `trim-paths` profile option.
+    If the build script introduces absolute paths to built artefacts (such as by invoking a compiler),
+    the user may request them to be sanitised in different types of artefacts.
```
epage, line 1379: `artefacts` -> `artifacts`

src/doc/src/reference/unstable.md
```diff
@@ -1289,6 +1290,96 @@ edition that may break your build.
...
+(just like when it is selected for sanitisation).
+Paths to all other source files will not be affected.
+
+This will not affect any hard-coded paths in the source code, such as in strings.
+
+#### Environment variable
+
+*as a new entry of ["Environment variables Cargo sets for build scripts"](./environment-variables.md#environment-variables-cargo-sets-for-crates)*
+
+* `CARGO_TRIM_PATHS` --- The value of `trim-paths` profile option.
```
epage, line 1377: Is this comma separated?  When there are two settings that match (e.g. `"none"` and `false`), what happens?

src/cargo/util/toml/mod.rs
```diff
@@ -3146,6 +3157,120 @@ impl<'de> de::Deserialize<'de> for TomlDebugInfo {
...
+            .string(|value| {
+                TomlTrimPathsValue::deserialize(value.into_deserializer()).map(TomlTrimPaths::One)
+            })
+            .seq(|seq| {
+                let seq: Vec<String> = seq.deserialize()?;
+                let seq = seq
+                    .into_iter()
+                    .map(|s| TomlTrimPathsValue::deserialize(s.into_deserializer()))
+                    .collect::<Result<_, _>>()?;
+                Ok(TomlTrimPaths::Many(seq))
```
epage, line 3191: Can we not deserliaize directly into `Vec<TomlTrimPathsValue>`?

src/cargo/util/toml/mod.rs
```diff
@@ -3146,6 +3157,120 @@ impl<'de> de::Deserialize<'de> for TomlDebugInfo {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Ord, PartialOrd, Hash, Serialize)]
+#[serde(untagged)]
+pub enum TomlTrimPaths {
+    One(TomlTrimPathsValue),
+    Many(Vec<TomlTrimPathsValue>),
```
epage, line 3164: Do we need to preserve `One` vs `Many` or can we just have the deserialize always put it in a `Vec`?

src/doc/src/reference/unstable.md
```diff
@@ -1289,6 +1290,96 @@ edition that may break your build.
...
+```
+
+### Documentation updates
+
+#### trim-paths
+
+*as a new ["Profiles settings" entry](./profiles.html#profile-settings)*
+
+`trim-paths` is a profile setting which enables and controls the sanitisation of file paths in build outputs.
+It takes either string or an array of the following values:
```
epage, line 1330: Not seeing anywhere that we specify an empty list

src/cargo/util/toml/mod.rs
```diff
@@ -3146,6 +3157,120 @@ impl<'de> de::Deserialize<'de> for TomlDebugInfo {
...
+                Ok(())
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum TomlTrimPathsValue {
+    None,
```
epage, line 3249: Do we need a `None` or should we use the empty list?

tests/testsuite/config.rs
```diff
@@ -1729,3 +1731,64 @@ jobs = 2
...
     }
 }
+
+#[cargo_test]
+fn trim_paths_parsing() {
+    let config = ConfigBuilder::new().build();
+    let p: cargo_toml::TomlProfile = config.get("profile.dev").unwrap();
+    assert_eq!(p.trim_paths, None);
+
+    // .unstable_flag("advanced-env")
```
epage, line 1741: Whats this line about?

src/cargo/util/toml/mod.rs
```diff
@@ -3146,6 +3157,120 @@ impl<'de> de::Deserialize<'de> for TomlDebugInfo {
...
+
+    pub fn all() -> TomlTrimPaths {
+        TomlTrimPaths::One(TomlTrimPathsValue::All)
+    }
+
+    pub fn diagnostics() -> TomlTrimPaths {
+        TomlTrimPaths::One(TomlTrimPathsValue::Diagnostics)
+    }
+
+    pub fn r#macro() -> TomlTrimPaths {
```
epage, line 3210: Should we avoid the need for `r#`?

src/cargo/core/profiles.rs
```diff
@@ -634,8 +641,8 @@ compact_debug! {
     impl fmt::Debug for Profile {
         fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
             let (default, default_name) = match self.name.as_str() {
-                "dev" => (Profile::default_dev(), "default_dev()"),
-                "release" => (Profile::default_release(), "default_release()"),
+                "dev" => (self.clone(), "default_dev()"),
+                "release" => (self.clone(), "default_release()"),
```
epage, line 645: Looks like this is a general fix, in dependent of this commit?

tests/testsuite/profile_trim_paths.rs
```diff
@@ -57,3 +61,380 @@ Caused by:
...
+
+    p.cargo("build").run();
+
+    let bin_path = p.bin("foo");
+    assert!(bin_path.is_file());
+    let stdout = run_readelf(bin_path).stdout;
+    // TODO: re-enable this check when rustc bootstrap disables remapping
+    // <https://github.com/rust-lang/cargo/pull/12625#discussion_r1371714791>
+    // assert!(memchr::memmem::find(&stdout, rust_src).is_some());
+    assert!(memchr::memmem::find(&stdout, registry_src_bytes).is_some());
```
epage, line 499: Is there a reason we are using `memchr::memmem`?

src/cargo/util/rustc.rs
```diff
@@ -80,6 +82,17 @@ impl Rustc {
...
         })?;
+        let commit_hash = extract("commit-hash: ").ok().map(|hash| {
+            assert!(
+                hash.chars().all(|ch| ch.is_ascii_hexdigit()),
+                "commit hash must be a hex string"
+            );
+            assert!(
+                hash.len() == 40 || hash.len() == 64,
+                "hex string must be generated from sha1 or sha256"
+            );
```
epage, line 93: Should we make these `debug_assert`s?

Or should we have a general message in production asserts/panics about being an internal error and what they should collect to report the issue?

src/cargo/util/rustc.rs
```diff
@@ -28,6 +28,8 @@ pub struct Rustc {
     pub version: semver::Version,
     /// The host triple (arch-platform-OS), this comes from verbose_version.
     pub host: InternedString,
+    /// The rustc full commit hash, this comes from `verbose_version`.
+    pub commit_hash: Option<String>,
```
epage, line 32: Could you update the commit message for this to say why this is added?  It took me digging 4 commits in further and searching to find it.

src/cargo/core/compiler/mod.rs
```diff
@@ -1162,6 +1168,74 @@ fn features_args(unit: &Unit) -> Vec<OsString> {
...
+        let mut remap = OsString::from("--remap-path-prefix=");
+        remap.push(sysroot);
+        remap.push("/lib/rustlib/src/rust"); // See also `detect_sysroot_src_path()`.
+        remap.push("=");
+        remap.push("/rustc/");
+        if let Some(commit_hash) = cx.bcx.rustc().commit_hash.as_ref() {
+            remap.push(commit_hash);
+        } else {
+            // This fallback aligns with rustc:
+            // <https://github.com/rust-lang/rust/blob/c2ef3516/src/bootstrap/src/lib.rs#L1113-L1116>
```
epage, line 1201: Should this comment be on the whole if-statement?  I think this explains why we use the commit hash.

src/cargo/core/compiler/mod.rs
```diff
@@ -1162,6 +1168,69 @@ fn features_args(unit: &Unit) -> Vec<OsString> {
...
+    cmd.arg(sysroot_remap);
+
+    let package_remap = {
+        let pkg_root = unit.pkg.root();
+        let ws_root = cx.bcx.ws.root();
+        let is_local = unit.pkg.package_id().source_id().is_path();
+        let mut remap = OsString::from("--remap-path-prefix=");
+        // Remapped to path relative to workspace root:
+        //
+        // * path dependencies under workspace root directory
```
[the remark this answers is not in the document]
epage: For me, the main question of whether to be relative to workspace or package root is if there'd be ambiguity.  Since we have name+version, I'm guessing not.

src/cargo/util/toml/mod.rs
weihanglo: Could you say more on the modeling you're looking for, or is there any existing code I can take as a reference?

src/doc/src/reference/unstable.md
weihanglo: Will remove this paragraph. Plan to do this in a follow-up.

src/cargo/util/toml/mod.rs
epage: Started writing this up but realized I was thinking of the wrong stuff, so sharing it here if its relevant

`package.publish` does this kind of things though in post-processing (`to_real_manifest`).  Its a `VecOrBool` and the `false` gets turned into `Some(vec![])`, `true` gets mapped to `None` (ie all), and then `Some(vec![...])` is the enumerated items.

src/cargo/util/toml/mod.rs
epage: Can't you do
```rust
let seq: Vec<TomlTrimPathsValue> = seq.deserialize()?;
```
Whether you deserialzie the `Vec` to have `String` or `TomlTrimPathsValue` shouldn't make a difference to serde.

src/doc/src/reference/unstable.md
weihanglo: Actually there are only two combinations are necessary:

* `[diagnostics, macro]`
* `[diagnostics, object]`

Other combinations can be represented by just one value.

I wonder if we really want an array for it.

tests/testsuite/profile_trim_paths.rs
weihanglo: The output of `readelf` is huge and I am not sure about its encoding. A premature optimization maybe. Any straightforward way to do it with std?

epage: Oh, right, `std` doesn't have substring matching for bytes,. right?  Oh well, then its fine.

src/doc/src/reference/unstable.md
epage: So this comment was on the documentation for how the end-user configures things, so we'd likely to keep it as specified.  Assuming you meant to refer to the internal representation, I'd worry its over optimization.  Notice my idea would add more allocations (though less indirection).

weihanglo: Not really. I am thinking that the current way to configure trim-paths is a bit confusing. You can have `trim-paths = ["all", "diagnostics", "macro"]` and it is exactly the same as `trim-paths = "all". If we add one more `"none"` to the array it become `trim-paths = "none"`.

We can implement the way as RFC says, but we might want to think about it later.

weihanglo: It also tricky to determine the meaning of `trim-paths = []`. Slightly perfer `none` for empty array now.

epage: So `none` has higher precedence and wipes everything out?  Thats bizarre.

src/cargo/util/toml/mod.rs
weihanglo:
```
invalid type: string "diagnostics", expected enum TomlTrimPathsValue
```

That was what I got with `let seq: Vec<TomlTrimPathsValue> = seq.deserialize()?;`. Not yet figure out why. A bit odd.

src/doc/src/reference/unstable.md
weihanglo: To avoid this issue, I just disallow "all" and "none" presence in the array. However, it makes `serialize` and `deserialize` asymmetric and not match each other.

src/cargo/util/toml/mod.rs
weihanglo: My thought is modeling what we see in user-end, but might be overkill you're right.

src/cargo/util/toml/mod.rs
weihanglo: These methods are now removed.

tests/testsuite/config.rs
weihanglo: Removed

src/cargo/util/toml/mod.rs
weihanglo: Just pushed a change disallowing empty array during deserialization.

src/cargo/util/toml/mod.rs
epage: My thought is mostly around keeping the internals simple rather than having multiple ways of expressing the same thing

epage: Looking at this more, I think I would personally model this as
```rust
enum TomlTrimPaths {
    None,
    Values(Vec<TomlTrimPathValues>),
    All,
}

enum TomlTrimPathValues {
    Objects,
    Diagnostics,
    Macro,
}
```

src/cargo/util/toml/mod.rs
epage: How come, doesn't an empty list communicate the same thing?  The empty list would also parallel `package.publish`

weihanglo: It is not really straightforward. An empty array is not always falsy in programming world, and it seems better to reduce the size of possible options.

Anyway, given we have the existing pattern, I'll follow.

src/cargo/util/toml/mod.rs
weihanglo: I went with the second one you suggested. Thanks for the idea and review!

src/cargo/core/compiler/mod.rs
```diff
@@ -1028,6 +1030,10 @@ fn build_base_args(cx: &Context<'_, '_>, cmd: &mut ProcessBuilder, unit: &Unit)
         }
     }
 
+    if let Some(trim_paths) = trim_paths {
+        trim_paths_args(cmd, cx, unit, &trim_paths)?;
+    }
```
ehuss, line 1035: Should this also be included in rustdoc to handle the diagnostics stripping (and same with doctests)?

(Though I still don't 100% understand the purpose of diagnostic remapping.)

tests/testsuite/profile_trim_paths.rs
```diff
@@ -0,0 +1,517 @@
...
+        .masquerade_as_nightly_cargo(&["-Ztrim-paths"])
+        .with_stderr_contains(
+            "\
+warning: unused variable: `unused`
+ --> bar-0.0.1/src/lib.rs:1:18
+  |
+1 | pub fn f() { let unused = 0; }
+  |                  ^^^^^^ help: if this is intentional, prefix it with an underscore: `_unused`
+  |
+  = note: `#[warn(unused_variables)]` on by default",
```
ehuss, line 405: Tests can't match on the compiler's output exactly, since it changes too often, and makes it impossible to change (since it would require simultaneously changing both rust-lang/rust and rust-lang/cargo).

I think you can do something like `with_stderr_contains("[..]bar-0.0.1/src/lib.rs:1[..]")` to just check the important part.

src/cargo/core/compiler/mod.rs
weihanglo: I don't think there is such an option for `rustdoc`. Did I miss something? 
Maybe this could be put in "unresolved questions" in RFC 3127 tracking issue.

```console
$ rustdoc +nightly -vV
rustdoc 1.75.0-nightly (608e9682f 2023-10-29)
binary: rustdoc
commit-hash: 608e9682f0a6482903de8d3332770104a0ad943c
commit-date: 2023-10-29
host: aarch64-apple-darwin
release: 1.75.0-nightly
LLVM version: 17.0.3

$ rustc --remap-path-prefix
error: Argument to option 'remap-path-prefix' missing

$ rustdoc --remap-path-prefix
error: Unrecognized option: 'remap-path-prefix'
```

tests/testsuite/profile_trim_paths.rs
weihanglo: Thanks for calling this out, and replaced this with `with_stderr_line_without`.

src/cargo/util/toml/mod.rs
```diff
@@ -3167,6 +3178,122 @@ impl<'de> de::Deserialize<'de> for TomlDebugInfo {
...
+                    .into_iter()
+                    .map(|s| TomlTrimPathsValue::deserialize(s.into_deserializer()))
+                    .collect::<Result<_, _>>()?;
+                Ok(seq.into())
+            })
+            .deserialize(d)
+    }
+}
+
+impl TomlTrimPaths {
```
epage, line 3229: nit: when i reorganized this file, I put all non-trait impl blocks first

src/cargo/util/toml/mod.rs
epage: Baffled by this but not a reason to block being merged

epage (approved):
Besides anything Eric brings up, I'm good with merging this.  Thanks for moving this along!

src/cargo/util/toml/mod.rs
weihanglo: Oh how I missed that. Fixed!

is there any rustfmt or clippy rules to enforce this?

src/cargo/core/compiler/mod.rs
ehuss: Yea, leaving that as an open issue sounds good. The rustdoc argument parser would just need to be updated.

Though, before that I would like to better understand why the diagnostic option exists and when someone would want to turn it on. I can't really follow the discussion from https://github.com/rust-lang/rust/issues/87745 and how that relates to cargo. If we end up not stabilizing "diagnostic" as an option, then I don't have any motivation to do that.

Urgau: Regrading the `diagnostics` option I think it's related to build-systems that create VFS (Virtual File System) or temporary location when building, throwing the all diagnostics path to the wrong location.

ehuss:
I haven't done a thorough review, but overall looks good to me.

@bors r=epage

PR #5205, opened 2018-03-19
fix #2773 with new precise encode

gibix:
Changed the precise encode from <pkg>=<precise> to <pkg>=<present_version>-><future_version> in order to check the correct requirements.

cc @lu-zero

PR #12629, opened 2023-09-06
Read/write the encoded `cargo update --precise` in the same place

Eh2406:

There's a stringly typed interface between https://github.com/rust-lang/cargo/blob/de7537e63296ee13cb78611a988bcc9a6ac26134/src/cargo/ops/cargo_generate_lockfile.rs#L105 and https://github.com/rust-lang/cargo/blob/de7537e63296ee13cb78611a988bcc9a6ac26134/src/cargo/sources/registry/index.rs#L587, the only reason I found it with by finding the original commit https://github.com/rust-lang/cargo/pull/5205

As far as I can tell, anyone could just create this internally meaningful ~structure~ string by passing it on the command line.

This should get cleaned up, for now by moving the encoding and decoding in to the same file.

Internal refactor and test still pass.

### Additional information

I am hoping that in the redesign of `cargo update` we can come up with a better design for smuggling this data from the API all the way to querying the registry. It seems like locking the dependency to the selected version would be conceptually simpler, or using the patch system, or something.

### DIFF
--- a/src/cargo/core/source/source_id.rs
+++ b/src/cargo/core/source/source_id.rs
@@ -2,7 +2,8 @@ use crate::core::PackageId;
 use crate::sources::registry::CRATES_IO_HTTP_INDEX;
 use crate::sources::{DirectorySource, CRATES_IO_DOMAIN, CRATES_IO_INDEX, CRATES_IO_REGISTRY};
 use crate::sources::{GitSource, PathSource, RegistrySource};
-use crate::util::{config, CanonicalUrl, CargoResult, Config, IntoUrl};
+use crate::util::{config, CanonicalUrl, CargoResult, Config, IntoUrl, ToSemver};
+use anyhow::Context;
 use serde::de;
 use serde::ser;
 use std::cmp::{self, Ordering};
@@ -430,11 +431,6 @@ impl SourceId {
         }
     }
 
-    /// Gets the value of the precise field.
-    pub fn precise(self) -> Option<&'static str> {
-        self.inner.precise.as_deref()
-    }
-
     /// Gets the Git reference if this is a git source, otherwise `None`.
     pub fn git_reference(self) -> Option<&'static GitReference> {
         match self.inner.kind {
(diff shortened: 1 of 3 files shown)


src/cargo/core/source/source_id.rs
```diff
@@ -443,6 +439,32 @@ impl SourceId {
...
+    pub fn precise(self) -> Option<&'static str> {
+        self.inner.precise.as_deref()
+    }
+
+    /// Check if the precise data field stores information for this `name`
+    /// from a call to [with_precise_registry_version].
+    /// If so return the version currently in the lock file and the version to be updated to.
+    /// If specified, our own source will have a precise version listed of the form
+    // `<pkg>=<p_req>-><f_req>` where `<pkg>` is the name of a crate on
+    // this source, `<p_req>` is the version installed and `<f_req> is the
```
epage, line 452: `<f_req>` doesn't have a closing back tick

src/cargo/core/source/source_id.rs
```diff
@@ -443,6 +439,32 @@ impl SourceId {
...
     }
 
+    /// Gets the value of the precise field.
+    pub fn precise(self) -> Option<&'static str> {
+        self.inner.precise.as_deref()
+    }
+
+    /// Check if the precise data field stores information for this `name`
+    /// from a call to [with_precise_registry_version].
+    /// If so return the version currently in the lock file and the version to be updated to.
```
epage, line 449: nit: a newline between short description and the rest

src/cargo/core/source/source_id.rs
```diff
@@ -443,6 +439,32 @@ impl SourceId {
...
+    pub fn precise_registry_version(
+        self,
+        name: &str,
+    ) -> Option<(semver::Version, semver::Version)> {
+        self.inner
+            .precise
+            .as_deref()
+            .filter(|p| p.starts_with(name) && p[name.len()..].starts_with('='))
+            .map(|p| {
+                let (current, requested) = p[name.len() + 1..].split_once("->").unwrap();
```
epage, line 463: nit: this could be simplified
```rust
.filter_map(|p|p.strip_prefix(name)?.strip_prefix('='))`
```

epage (approved):
All minor so feel free to r= me when you feel its ready

PR #12631, opened 2023-09-06
Ues strip_prefix for cleaner code

Eh2406:

In https://github.com/rust-lang/cargo/pull/12629#pullrequestreview-1614154046 Ed pointed out how much cleaner the code can be using `strip_prefix`, so I found a bunch more places where we should be using it.

Internal refactor and test still pass.

### DIFF
--- a/src/cargo/core/resolver/encode.rs
+++ b/src/cargo/core/resolver/encode.rs
@@ -338,7 +338,7 @@ impl EncodableResolve {
         let mut to_remove = Vec::new();
         for (k, v) in metadata.iter().filter(|p| p.0.starts_with(prefix)) {
             to_remove.push(k.to_string());
-            let k = &k[prefix.len()..];
+            let k = k.strip_prefix(prefix).unwrap();
             let enc_id: EncodablePackageId = k
                 .parse()
                 .with_context(|| internal("invalid encoding of checksum in lockfile"))?;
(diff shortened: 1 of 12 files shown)


src/cargo/core/resolver/encode.rs
```diff
@@ -338,7 +338,7 @@ impl EncodableResolve {
         let mut to_remove = Vec::new();
         for (k, v) in metadata.iter().filter(|p| p.0.starts_with(prefix)) {
             to_remove.push(k.to_string());
-            let k = &k[prefix.len()..];
+            let k = k.strip_prefix(prefix).unwrap();
```
epage, line 341: nit: you could turn the `filter` into a `filter_map` but then each iteration would have two different keys so unsure if its worth it

src/cargo/util/rustc.rs
```diff
@@ -63,8 +63,8 @@ impl Rustc {
         let extract = |field: &str| -> CargoResult<&str> {
             verbose_version
                 .lines()
-                .find(|l| l.starts_with(field))
-                .map(|l| &l[field.len()..])
+                .filter_map(|l| l.strip_prefix(field))
+                .next()
```
epage, line 67: nit: [find_map](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map)?

epage (approved):
Looks good.  Feel free to r= me

Eh2406:
I found a lot more.

epage (approved):

Eh2406:
I don't understand the error report, I don't think I made a breaking change.

tests/testsuite/ssh.rs
```diff
@@ -184,10 +184,9 @@ fn known_host_works() {
     // Validate the fingerprint while we're here.
     let fingerprint = stderr
         .lines()
-        .find(|line| line.starts_with("  The ECDSA key fingerprint"))
+        .find_map(|line| line.strip_prefix("  The ECDSA key fingerprint"))
```
ehuss, line 187: This isn't quite equivalent, and is causing a test failure. I think it needs to be something like:

```suggestion
        .find_map(|line| line.strip_prefix("  The ECDSA key fingerprint is: "))
```

ehuss:
The check isn't for just semver-breaking changes, but for any change. If there aren't any semver-breaking changes, then you just need to bump the patch version.

Issue #12633, opened 2023-09-07
Tracking Issue for garbage collection


ehuss:
### Summary

Original proposal: https://hackmd.io/@rust-cargo-team/SJT-p_rL2
Implementation: #NNNN
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#gc

The `-Zgc` flag enable garbage collection for deleting old, unused files in cargo's cache.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

epage:
Quick scan of brew
- [autoremove](https://docs.brew.sh/Manpage#autoremove---dry-run)
- [cleanup](https://docs.brew.sh/Manpage#cleanup-options-formulacask-)
  - "Removes all downloads more than 120 days old. This can be adjusted with HOMEBREW_CLEANUP_MAX_AGE_DAYS."
- HOMEBREW_CLEANUP_PERIODIC_FULL_DAYS (default 30 days)

One complaint that came up was " "`brew cleanup` has not been run in 30 days, running now" ... and then proceeds to run an interminable process in the middle of you attempting to do something else." (mastadon](https://hachyderm.io/@nathanhammond@mastodon.social/111048254907768474)

epage:
From https://hachyderm.io/@nathanhammond@mastodon.social/111048319933010803

>  Also: global vs. local? Just keep all things local. Errant cache hits pulling from a non-scoped global cache are a disaster. Errant cache hits from the same local cache are a "whoops." For shared artifacts (e.g. a singular tar downloaded from the registry), global is fine, and should just grow unbounded. (Seriously: nobody cares.)
>
> pnpm, bun, npm, yarn: package cache is intended to be shared global immutable, unbounded growth.

I'm assuming our global package cache is to help with CI but we should probably explicitly document out priority use cases.

mcclure:
Is -Zgc the final planned CLI interface?

To my mind clearing the global cache is a "verb" so it being something like "cargo clean-cache" would be discoverable but "-Zgc" (to what command?) would be less so.

[...]

PR #11879, opened 2023-03-23
Support `cargo owner add`

heisen-li:
### What does this PR try to resolve?

User confusion over how to invoke it, including
- `cargo owner` doing nothing, rather than erroring and telling the user how to use it
- Users assuming the interface uses subcommands, rather than action flags

This is done by deprecating the old interface and replacing it with subcommands.

To do this
- `cargo owner` will now be synonymous with `cargo owner --help` except a bad error code will be returned
- `cargo owner --help` includes a usage for each subcommand (clap-rs/clap#1334 could do this for us)
- `--add`, `--list`, and `--remove` are deprecated in favor of `add`, `list`, and `remove`
  - Using a flag with a subcommand is a hard error
  - The old flags are hidden in `--help`

This new interface loses the ability to do multiple operations at once but that seems like a premature optimization.

### Additional information

**Breaking change:** `cargo owner` used to be a no-op (no actions specified) but will now be a hard error.  This change is being made assuming the the use case for running `cargo owner` without any actions in a programmatic setting is small enough that this is acceptable

Fixes #4352

Issue #6102, opened 2018-09-26
Hide unstable options from --help without -Z unstable-options


joshtriplett:
To simplify `--help`, we should hide unstable options unless you pass `-Z unstable-options --help`.

epage:
At this time, clap does not support dynamically generating help based on other flags being present.  This would be made even more complicated by the fact that `-Z` is also a global flag.

As an alternative, we could check the `channel` and `hide` all unstable flags based on that.  This would make it so they only show up when using a nightly build.

PR #12634, opened 2023-09-07
Add cache garbage collection

ehuss:

This introduces a new garbage collection system which can track the last time files were used in cargo's global cache, and delete old, unused files either automatically or manually.

This is broken up into a large number of commits, and each commit should have a short overview of what it does. I am breaking some of these out into separate PRs as well (unfortunately GitHub doesn't really support stacked pull requests). I expect to reduce the size of this PR if those other PRs are accepted. 

I would first review `unstable.md` to give you an idea of what the user side of this looks like. I would then skim over each commit message to give an overview of all the changes. The core change is the introduction of the `GlobalCacheTracker` which is an interface to a sqlite database which is used for tracking the timestamps.

### Additional information

I think the interface for this will almost certainly change over time. This is just a stab to create a starting point where we can start testing and discussing what actual user flags should be exposed. This is also intended to start the process of getting experience using sqlite, and getting some testing in real-world environments to see how things might fail. 

I'd like to ask for the review to not focus too much on bikeshedding flag names and options. I expect them to change, so this is by no means a concrete proposal for where it will end up. For example, the options are very granular, and I would like to have fewer options. However, it isn't clear how that might best work. The size-tracking options almost certainly need to change, but I do not know exactly what the use cases for size-tracking are, so that will need some discussion with people who are interested in that.

I decided to place the gc commands in cargo's `cargo clean` command because I would like to have a single place for users to go for deleting cache artifacts. It may be possible that they get moved to another command, however introducing new subcommands is quite difficult (due to shadowing existing third-party commands). Other options might be `cargo gc`, `cargo maintenance`, `cargo cache`, etc. But there are existing extensions that would interfere with.

There are also more directions to go in the future. For example, we could add a `cargo clean info` subcommand which could be used for querying cache information (like the sizes and such). There is also the rest of the steps in the original proposal at https://hackmd.io/U_k79wk7SkCQ8_dJgIXwJg for rolling out sqlite support.

See #12633 for the tracking issue


crates/xtask-bump-check/src/xtask.rs
```diff
@@ -361,7 +362,7 @@ fn check_crates_io<'a>(
 ) -> CargoResult<()> {
     let source_id = SourceId::crates_io(config)?;
     let mut registry = PackageRegistry::new(config)?;
-    let _lock = config.acquire_package_cache_lock()?;
+    let _lock = config.acquire_package_cache_lock(CacheLockMode::DownloadExclusive)?;
```
epage, line 365: Could this be split out into its own PR?  I feel like there is a lot to unpack with this
- Clearly documenting interactions with old cargo
- Working out how callers are supposed to know which lock type to use, especially when copy/pasting
- Strategies for correctness

src/bin/cargo/commands/clean.rs
```diff
@@ -17,29 +21,243 @@ pub fn cli() -> Command {
...
             )
             .short('n'),
         )
+
+        // NOTE: Not all of these options may get stabilized. Some of them are
+        // very low-level details, and may not be something typical users need.
+        .arg(
+            optional_opt(
+                "gc",
+                "Delete old and unused files (unstable) (comma separated): all, download, target, shared-target",
```
epage, line 34: Clap will list out the values for you because you used `PossibleValueParser`

(this is mostly a nit but don't want to forget this as we move forward)

src/bin/cargo/commands/clean.rs
```diff
@@ -17,29 +21,243 @@ pub fn cli() -> Command {
...
+            )
+            .hide(true)
+            .value_name("KINDS")
+            .value_parser(
+                PossibleValuesParser::new(["all", "download", "target", "shared-target"]).map(|x|
+                    match x.as_str() {
+                        "all" => AutoGcKind::All,
+                        "download" => AutoGcKind::Download,
+                        "target" => panic!("target is not yet implemented"),
+                        "shared-target" => panic!("shared-target is not yet implemented"),
```
epage, line 44: Should we just not mention the "not implemented" stuff?

There are other things not implemented as well (e.g. cargo script cache) and we don't really know what the CLI will look like yet

src/bin/cargo/commands/clean.rs
```diff
@@ -17,29 +21,243 @@ pub fn cli() -> Command {
...
+                        x => panic!("possible value out of sync with `{x}`"),
+                    }
+            ))
+            .require_equals(true),
```
epage, line 48: nit (since focus isn't on CLI yet): not normally a fan of requires_equals esp if we don't use it on everything

(looks like we only use it in one place so far, `--timings`)

src/bin/cargo/commands/clean.rs
```diff
@@ -17,29 +21,243 @@ pub fn cli() -> Command {
...
+    let unstable_cache_opt = |opt| -> CargoResult<Option<&str>> {
+        let arg = args.get_one::<String>(opt).map(String::as_str);
+        if arg.is_some() {
+            unstable_gc(opt)?;
+        }
+        Ok(arg)
+    };
+    let unstable_size_opt = |opt| -> CargoResult<Option<u64>> {
+        unstable_cache_opt(opt)?
+            .map(|s| parse_human_size(s))
```
epage, line 195: Should we just be putting the `parse_human_size`, etc, as `Arg::value_parser`?

src/bin/cargo/commands/clean.rs
```diff
@@ -17,29 +21,243 @@ pub fn cli() -> Command {
...
+            .hide(true)
+            .value_name("DURATION"),
+        )
+
         .after_help("Run `cargo help clean` for more detailed information.\n")
 }
 
 pub fn exec(config: &mut Config, args: &ArgMatches) -> CliResult {
-    let ws = args.workspace(config)?;
+    let ws = args.workspace(config);
```
epage, line 171: A mantra that I feel should be followed here is "the hackiest hack is the best hack" or rather "when throwing something together, be surgical rather than over-generalize".

The CLI is a placeholder atm so I'd expect it to be very surgical in how its implemented.  The parts of the GC we are working on for now are global, so I would expect a switch in `exec` for global vs non-global mode and not to also dive down deeper into the layers in the clean op.  Because this is trying to jump ahead to target stuff, we are making touching a lot more code than is appropriate for where this effort is atm

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
...
+pub struct Gc<'a, 'config> {
+    config: &'config Config,
+    global_cache_tracker: &'a mut GlobalCacheTracker,
+    /// A lock on the package cache.
+    ///
+    /// This is important to be held, since we don't want multiple cargos to
+    /// be allowed to write to the cache at the same time, or for others to
+    /// read while we are modifying the cache.
+    #[allow(dead_code)] // Held for drop.
+    lock: CacheLock<'config>,
```
epage, line 34: nit: normally I've done `_` prefixes for this instead of an `allow`

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
...
+    pub max_download_size: Option<u64>,
+
+    /// The `--max-target-age` CLI option (UNIMPLEMENTED).
+    pub max_target_age: Option<Duration>,
+    /// The `--max-shared-target-age CLI option (UNIMPLEMENTED).
+    pub max_shared_target_age: Option<Duration>,
+    /// The `--max-target-size` CLI option  (UNIMPLEMENTED).
+    pub max_target_size: Option<u64>,
+    /// The `--max-shared-target-size` CLI option (UNIMPLEMENTED).
+    pub max_shared_target_size: Option<u64>,
```
epage, line 89: The more I see of these, the more uncomfortable I get.  For how big of a PR this is, I don't think we should be trying to also demonstrate other things ,creating a lot of dead code baggage.

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
...
+    /// This returns immediately without doing work if garbage collection has
+    /// been performed recently (since `gc.auto.frequency`).
+    fn auto(&mut self, clean_ctx: &mut CleanContext<'config>) -> CargoResult<()> {
+        if !self.config.cli_unstable().gc {
+            return Ok(());
+        }
+        let auto_config = self
+            .config
+            .get::<Option<AutoConfig>>("gc.auto")?
+            .unwrap_or_default();
```
epage, line 255: Double check but I'm assuming I'm over-optimizing by thinking that for `gc.auto` we should only load as much of the config as is necessary for determining whether GC needs to run or no before loading everything else.

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
...
+        Ok(())
+    }
+
+    /// Performs garbage collection based on the given options.
+    pub fn gc(
+        &mut self,
+        clean_ctx: &mut CleanContext<'config>,
+        gc_opts: &GcOpts,
+    ) -> CargoResult<()> {
+        self.global_cache_tracker.clean(clean_ctx, gc_opts)?;
```
epage, line 279: From an API perceptive, this feels weird that `GlobalCacheTracker` sometimes represents dry-run and sometimes the caller has to take care of it

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
+//! Support for garbage collecting unused files from downloaded files or
```
epage, line 1: A part of me feels like this doesn't belong in `core` but I don't have a big enough picture yet to say for sure or where this should go but leaving this comment in case it inspires thoughts in others.

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
...
+             \"N seconds/minutes/days/weeks/months\", got: {span:?}"
+        )
+    })
+}
+
+/// Parses a time span string.
+///
+/// Returns None if the value is not valid. See [`parse_time_span`] if you
+/// need a variant that generates an error message.
+fn maybe_parse_time_span(span: &str) -> Option<Duration> {
```
epage, line 352: Thoughts on using [humantime::parse_duration](https://docs.rs/humantime/2.1.0/humantime/fn.parse_duration.html)?

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
...
+    let Some(right_i) = span.find(|c: char| !c.is_ascii_digit()) else {
+        return None;
+    };
+    let left = &span[..right_i];
+    let mut right = &span[right_i..];
```
epage, line 357: We could use [str::split_at](https://doc.rust-lang.org/std/string/struct.String.html#method.split_at)

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
...
+        right = &right[1..];
+    }
+    let count: u64 = left.parse().ok()?;
+    let factor = match right {
+        "second" | "seconds" => 1,
+        "minute" | "minutes" => 60,
+        "hour" | "hours" => 60 * 60,
+        "day" | "days" => 24 * 60 * 60,
+        "week" | "weeks" => 7 * 24 * 60 * 60,
+        "month" | "months" => 30 * 24 * 60 * 60,
```
epage, line 368: alternative idea: humantime uses 30.44 days (https://github.com/tailhook/humantime/blob/master/src/duration.rs#L136)

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
...
+        format_err!(
+            "expected a value of the form \
+             \"N seconds/minutes/days/weeks/months\", got: {span:?}"
+        )
+    })
+}
+
+/// Parses a file size using metric or IEC units.
+pub fn parse_human_size(input: &str) -> CargoResult<u64> {
+    let re = regex::Regex::new(r"(?i)^([0-9]+(\.[0-9])?) ?(b|kb|mb|gb|kib|mib|gib)?$").unwrap();
```
epage, line 386: There a reason we aren't using `bytesize` for this?

epage, line 386: How come the mix of regex (for bytes) and hand parsing (time)?

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
...
+/// It should be cheap to call this multiple times (subsequent calls are
+/// ignored), but try not to abuse that.
+pub fn auto_gc(config: &Config) {
+    if !config.cli_unstable().gc {
+        return;
+    }
+    if !config.network_allowed() {
+        // As a conservative choice, auto-gc is disabled when offline. If the
+        // user is indefinitely offline, we don't want to delete things they
+        // may later depend on.
```
epage, line 435: Should we trace?  And should we have a gc trace target?

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
...
+/// be performed automatically based on the config settings. The default
+/// behavior is to only clean once a day.
+///
+/// This should only be called in code paths for commands that are already
+/// doing a lot of work. It should only be called *after* crates are
+/// downloaded so that the last-use data is updated first.
+///
+/// It should be cheap to call this multiple times (subsequent calls are
+/// ignored), but try not to abuse that.
+pub fn auto_gc(config: &Config) {
```
epage, line 428: This file is organized by
1. types
2. impls
3. free functions
4. tests

I found this confusing in reading it and I expect people to have false starts with this API in using it.  It appears the primary entry point into this is `gc::auto_gc`.  I saw `Gc` first and would have immediately tried to use the functions on it.   For me, I would put `gc::auto_gc` first and have it serve a a Table of Content for the rest of the file (much like the idea of putting `main` first)

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
...
+    if !config.network_allowed() {
+        // As a conservative choice, auto-gc is disabled when offline. If the
+        // user is indefinitely offline, we don't want to delete things they
+        // may later depend on.
+        return;
+    }
+
+    if let Err(e) = auto_gc_inner(config) {
+        if global_cache_tracker::is_silent_error(&e)
+            && config.shell().verbosity() != Verbosity::Verbose
```
epage, line 441: Currently "extra verbose" is a property of `config` and not `shell`.  If we refactored that, this would break.

src/cargo/util/sqlite.rs
```diff
@@ -0,0 +1,118 @@
...
+
+/// Perform one-time SQL migrations.
+///
+/// See [`Migration`] for more information.
+pub fn migrate(conn: &mut Connection, migrations: &[Migration]) -> CargoResult<()> {
+    // EXCLUSIVE ensures that it starts with an exclusive write lock. No other
+    // readers will be allowed. This generally shouldn't be needed if there is
+    // a file lock, but might be helpful in cases where cargo's `FileLock`
+    // failed.
+    let tx = conn.transaction_with_behavior(TransactionBehavior::Exclusive)?;
```
Eh2406, line 88: Do we want to migrate our overall locking approach so that we lock a row in the table instead of creating a file lock? long-term it's something to think about, short-term obviously not.

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+//! file.
+//!
+//! While cargo is running, when it detects a use of a cache file, it adds a
+//! timestamp to [`DeferredGlobalLastUse`]. This batches up a set of changes
+//! that are then flushed to the database all at once (via
+//! [`DeferredGlobalLastUse::save`]). Ideally saving would only be done once
+//! for performance reasons, but that is not really possible due to the way
+//! cargo works, since there are different ways cargo can be used (like `cargo
+//! generate-lockfile`, `cargo fetch`, and `cargo build` are all very
+//! different ways the code is used).
```
epage, line 15: Could we put the `save` in the CLI layer?

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+//!
+//! All of the database interaction is done through the [`GlobalCacheTracker`]
+//! type.
+//!
+//! There is a single global [`GlobalCacheTracker`] and
+//! [`DeferredGlobalLastUse`] stored in [`Config`].
+//!
+//! ## Automatic gc
```
epage, line 23: A lot of this documentation I would have expected to find in the `gc` module as I would assume its the top level entry point for gc.

I feel like we should more explicitly say to check this module for more details or move the details over to there

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+//!
+//! Some commands (primarily the build commands) will trigger an automatic
+//! deletion of files that haven't been used in a while. The interface for
+//! this is in the [`crate::core::gc`] module. The database tracks the last
```
epage, line 27: imo we should link directly to `gc::auto_gc`

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+//! time an automatic gc was performed so that it is only done once per day
+//! for performance reasons.
+//!
+//! ## Manual gc
+//!
+//! The user can perform a manual garbage collection with the `cargo clean`
+//! command. That command has a variety of options to specify what to delete.
+//! Manual gc supports deleting based on age or size or both.
```
epage, line 35: imo we should link to where the top-level manual gc call (currently lives in the op) as that has all of the contract of what is expected and is equivalent of `gc::auto_gc`

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+//!
+//! ## Locking
+//!
+//! Usage of the database requires that the package cache is locked to prevent
+//! concurrent access. Although sqlite has built-in locking support, we want
+//! to use cargo's locking so that the "Blocking" message gets displayed, and
+//! so that locks can block indefinitely for long-running build commands.
+//! [`rusqlite`] has a default timeout of 5 seconds, though that is
+//! configurable.
```
epage, line 44: > so that locks can block indefinitely for long-running build commands.

This had me worried for a second that a `cargo build` will block on another `cargo build` due to gc but this is only true for manual gc and not auto-gc as we skip auto-gc if there is contention for the lock.

I don't know if we should go into that level of detail but I feel like some clarification is needed or else this might continue to raise eyebrows (from future me or others) who see this and get concerned.

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+//! version 3 had a format change, but we wouldn't use it anyway since it has
+//! shared-memory requirements cargo can't depend on due to things like
+//! network mounts).
+//!
+//! Schema changes must be managed through [`migrations`] by adding new
+//! entries that make a change to the database. Changes must not break older
+//! versions of cargo. Generally, adding columns should be fine (either with a
+//! default value, or NULL). Adding tables should also be fine. Just don't do
+//! destructive things like removing a column, or changing the semantics of an
+//! existing column.
```
epage, line 65: How are we expected to deal with adding new columns because we intend to use them but old cargo will ignore those columns?

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+//! compare different environments, or changes to the code here. Please try to
+//! keep performance in mind if making any major changes.
+//!
+//! Performance of `cargo clean` is not quite as important since it is not
+//! expected to be run often. However, it is still courteous to the user to
+//! try to not impact it too much. One part that has a performance concern is
+//! that the clean command will synchronize the database with whatever is on
+//! disk if needed (in case files were added by older versions of cargo that
+//! don't do cache tracking, or if the user manually deleted some files). This
+//! can potentially be very slow, especially if the two are very out of sync.
```
epage, line 83: More a placeholder for myself so I don't forget but is the state sync only on manual GC and not auto-gc?  This makes it sound like it but I wanted to go review that code to double check

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+//!
+//! ## Filesystems
+//!
+//! Everything here is sensitive to the kind of filesystem it is running on.
+//! People tend to run cargo in all sorts of strange environments that have
+//! limited capabilities, or on things like read-only mounts. The code here
+//! needs to gracefully handle as many situations as possible.
+//!
+//! The sections above about performance and locking are very relevant when
```
epage, line 92: nit: "about about" -> "about"

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+/// via the [`DeferredGlobalLastUse::save`] method.
+///
+/// This is here to improve performance.
+#[derive(Debug)]
+pub struct DeferredGlobalLastUse {
+    /// Cache of registry keys, used for faster fetching.
+    ///
+    /// The key is the registry name (which is its directory name) and the
+    /// value is the `id` in the `registry_index` table.
+    registry_keys: HashMap<InternedString, i64>,
```
epage, line 152: Can we use a new type for the table id's?  I feel like it'd make the intent of the code clearer and likely reduce the chance for bugs

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+    pub package_dir: InternedString,
+    // Total size of the src directory in bytes.
+    //
+    // This can be None when the size is unknown. For example, when the src
+    // directory already exists on disk, and we just want to update the
+    // last-use timestamp. We don't want to take the expense of computing disk
+    // usage unless necessary. `populate_untracked_src` will handle any actual
+    // NULL values in the database, which can happen when the src directory is
+    // created by an older version of cargo that did not track sizes.
+    pub size: Option<u64>,
```
epage, line 204: If we still wanted to scale this PR back further, I think cutting out disk size based GC operations would be a big help as it would reduce the du code.  It'd also remove the du runtime overhead so we can better evaluate what that parts impact is to overall cache tracking performance.

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+            // want to rely on cargo's own "Blocking" system (which can
+            // provide user feedback) rather than blocking inside sqlite
+            // (which by default has a short timeout).
+            let db_path =
+                config.assert_package_cache_locked(CacheLockMode::DownloadExclusive, &db_path);
+            Connection::open(db_path)?
+        } else {
+            // To simplify things (so there aren't checks everywhere for being
+            // enabled), just process everything in memory.
+            Connection::open_in_memory()?
```
epage, line 341: Should we still call `assert_package_cache_locked` in the "stable" code path so we're more likely to catch bugs when developing new features?  Or would we be concerned about that destablizing stable and just deal with it when we work to stabilize?

Eh2406:
I'm sure there are lots of details still to be discussed. In fact Ed found a lot of them while I was looking things over. Overall this looks really good and I'm excited to see this happen! Thank you!

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+            crate_dir: config.registry_cache_path().into_path_unlocked(),
+            src: config.registry_source_path().into_path_unlocked(),
+        };
+        let now = now();
+        trace!("cleaning {gc_opts:?}");
+        let tx = self.conn.transaction()?;
+        let mut delete_paths = Vec::new();
+        // This can be an expensive operation, so only perform it if necessary.
+        if gc_opts.is_download_cache_opt_set() {
+            // TODO: Investigate how slow this might be.
```
epage, line 555: How do we plan to track this TODO?

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+            Self::sync_db_with_files(
```
epage, line 556: `is_download_cache_opt_set` takes into account settings from auto-gc but I thought the sync isn't supposed to happen on auto-gc due to how slow it is?

See also https://github.com/rust-lang/cargo/pull/12634#discussion_r1320310008

src/cargo/core/global_cache_tracker.rs
epage: See also https://github.com/rust-lang/cargo/pull/12634/commits/e13110f67826e85b85aa5e4da815b1ee8e1eee3e#r1320328466

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+        ));
+        trace!("checking for untracked parent to add to {parent_table_name}");
+        let names = Self::names_from(base_path)?;
+
+        let mut stmt = conn.prepare_cached(&format!(
+            "INSERT INTO {parent_table_name} (name, timestamp)
+                VALUES (?1, ?2)
+                ON CONFLICT DO NOTHING",
+        ))?;
+        let now = now();
```
epage, line 795: `DeferredGlobalLastUse` had a single instance of `now()`.  I'm seeing a lot here as well.  Should the global use tracker have a single member?  Maybe we even have the config do a single `now` for the process and initialize both of these with it?

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+    /// not tracked (such as when they are downloaded by an older version of
+    /// cargo).
+    fn populate_untracked_crate(conn: &Connection, base_path: &Path) -> CargoResult<()> {
+        let _p = crate::util::profile::start("populate untracked crate");
+        trace!("populating untracked crate files");
+        let mut insert_stmt = conn.prepare_cached(
+            "INSERT INTO registry_crate (registry_id, name, size, timestamp)
+             VALUES (?1, ?2, ?3, ?4)
+             ON CONFLICT DO NOTHING",
+        )?;
```
epage, line 879: Would it help if we logged all of our DB mutations we perform?

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+                }
+            }
+        }
+        Ok(())
+    }
+
+    /// Updates the database to add any `.crate` files that are currently
+    /// not tracked (such as when they are downloaded by an older version of
+    /// cargo).
+    fn populate_untracked_crate(conn: &Connection, base_path: &Path) -> CargoResult<()> {
```
epage, line 872: Note to self (and others): this is about the point where my eyes are glazing over and not giving things as much attention

src/cargo/core/global_cache_tracker.rs
```diff
@@ -0,0 +1,1754 @@
...
+            return true;
+        }
+    }
+    false
+}
+
+fn du(path: &Path, table_name: &str) -> CargoResult<u64> {
+    // !.git is used because clones typically use hardlinks for the git
+    // contents. TODO: Verify behavior on Windows.
+    // TODO: Or even better, switch to worktrees, and remove this.
```
epage, line 1747: Whats the plan for tracking and resolving these TODOs?

src/cargo/ops/cargo_fetch.rs
```diff
@@ -76,6 +76,7 @@ pub fn fetch<'a>(
     }
 
     packages.get_many(to_download)?;
+    crate::core::gc::auto_gc(config);
```
epage, line 79: When compiling, we gc first.  When fetching, we gc after.  Which should we do generally and why the deviation?

src/cargo/sources/git/source.rs
```diff
@@ -294,8 +310,19 @@ impl<'cfg> Source for GitSource<'cfg> {
...
-        self.path_source.as_mut().unwrap().update()
+        self.path_source.as_mut().unwrap().update()?;
+
+        // Hopefully this shouldn't incur too much of a performance hit since
+        // most of this should already be in cache since it was just
+        // extracted.
+        //
+        // !.git is used because clones typically use hardlinks for the git
+        // contents. TODO: Verify behavior on Windows.
+        let size = cargo_util::du(&checkout_path, &["!.git"])?;
```
epage, line 323: Currently we duplicate the ignore patterns between the sync code and the fresh insert code.  Is there an alternative?

src/bin/cargo/commands/clean.rs
```diff
@@ -17,29 +21,243 @@ pub fn cli() -> Command {
         .arg(
             flag(
                 "dry-run",
-                "Display what would be deleted without deleting anything",
+                "Display what would be deleted without deleting anything (unstable)",
             )
```
epage: Why was this marked `unstable`?

tests/testsuite/clean.rs
```diff
@@ -676,6 +676,20 @@ fn clean_spec_reserved() {
         .run();
 }
 
+#[cargo_test]
+fn clean_dry_run_gated() {
```
epage, line 680: I thought the other PR is making it stable, why is this commit within the PR undoing that?

tests/testsuite/global_cache_tracker.rs
```diff
@@ -0,0 +1,1865 @@
...
+        .exists());
+    assert_eq!(get_index_names().len(), 1);
+}
+
+#[cargo_test]
+fn cache_clean_options_gated() {
+    // Checks that all cache clean options require -Zgc.
+    let p = project().build();
+    for opt in [
+        "--dry-run",
```
epage, line 179: Looks like the dry-run unstable test is duplicated

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,550 @@
...
+                )?;
+                self.max_git_co_age = newer_time_span_for_config(
+                    self.max_git_co_age,
+                    "gc.auto.max-git-co-age",
+                    auto_config.max_git_co_age.as_deref().unwrap_or("1 month"),
+                )?;
+                self.max_git_db_age = newer_time_span_for_config(
+                    self.max_git_db_age,
+                    "gc.auto.max-git-db-age",
+                    auto_config.max_git_db_age.as_deref().unwrap_or("3 months"),
```
epage, line 185: Should we use constants for the defaults.

Looks like they are divided between being a primary and secondary source

tests/testsuite/clean.rs
```diff
@@ -685,7 +685,7 @@ fn clean_dry_run_gated() {
         .with_stderr("\
 error: the `dry-run` flag is unstable, and only available on the nightly channel of Cargo, but this is the `stable` channel
 See https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.
-See https://github.com/rust-lang/cargo/issues/0 for more information about the `dry-run` flag.
+See https://github.com/rust-lang/cargo/issues/12633 for more information about the `dry-run` flag.
```
epage, line 688: This commit didn't update the tracking issue link in the unstable docs

src/doc/src/reference/unstable.md
```diff
@@ -1665,6 +1666,66 @@ version = "0.1.0"
...
+# Anything older than this duration will be deleted in the source cache.
+max-src-age = "1 month"
+# Anything older than this duration will be deleted in the compressed crate cache.
+max-crate-age = "3 months"
+# Any index older than this duration will be deleted from the index cache.
+max-index-age = "3 months"
+# Any git checkout older than this duration will be deleted from the checkout cache.
+max-git-co-age = "1 month"
+# Any git clone older than this duration will be deleted from the git cache.
+max-git-db-age = "3 months"
```
epage, line 1726: Can you put the table header for these?

epage:
That was a lot of work; thanks for doing it all Eric.  I am fine with deferring CLI conversations and likely other things to get this in.  My review comments were both low level and high level for the sake of not wanting to forget things.

mati865:
CC @matthiaskrgr, creator of https://github.com/matthiaskrgr/cargo-cache for awareness or maybe even some insights

src/doc/src/reference/unstable.md
```diff
@@ -1665,6 +1666,66 @@ version = "0.1.0"
...
+
+The `-Zgc` flag enables garbage-collection within cargo's global cache within the cargo home directory.
+This includes downloaded dependencies such as compressed `.crate` files, extracted `src` directories, registry index caches, and git dependencies.
+When `-Zgc` is present, cargo will track the last time any file was used, and then uses those timestamps to manually or automatically delete files that have not been used for a while.
+
+```sh
+cargo build -Zgc
+```
+
+Automatic deletion happens on commands that are already doing a significant amount of work,
```
matthiaskrgr, line 1681: >on

before or after?

src/doc/src/reference/unstable.md
```diff
@@ -1665,6 +1666,66 @@ version = "0.1.0"
...
+- `--max-crate-age=DURATION` --- Deletes crate cache files that have not been used since the given age.
+- `--max-index-age=DURATION` --- Deletes registry indexes that have not been used since then given age (including their `.crate` and `src` files).
+- `--max-git-co-age=DURATION` --- Deletes git dependency checkouts that have not been used since then given age.
+- `--max-git-db-age=DURATION` --- Deletes git dependency clones that have not been used since then given age.
+- `--max-download-age=DURATION` --- Deletes any downloaded cache data that has not been used since then given age.
+- `--max-src-size=SIZE` --- Deletes the oldest source cache files until the cache is under the given size.
+- `--max-crate-size=SIZE` --- Deletes the oldest crate cache files until the cache is under the given size.
+- `--max-git-size=SIZE` --- Deletes the oldest git dependency caches until the cache is under the given size.
+- `--max-download-size=SIZE` --- Deletes the oldest downloaded cache data until the cache is under the given size.
+
```
matthiaskrgr, line 1700: Kind of wondering what will happen, if I set a limit of 500mb but a rust projects needs a git db that exceeds this limit.
Will this break the build?

matthiaskrgr:
From the hackmd

````
This information will include timestamps of the last time the following were “used”:

    git/checkouts/<name>/<hash>/
    git/db/<name>
    registry/index/<name>
    registry/cache/<registry>/<name>.crate
    registry/src/<registry>/<name>

````

name/hash etc, there are what the code refers to as "file", right?
I find this a bit confusing because "delete files that have not been used for ..." sounds a bit like we could delete single of a folder containing `.rs` files.
Maybe "artifact" would be a better term?

````
The timestamp is updated whenever something is initially added, or something reads from it.
````


So can cargo only track changes that are done by itself?
I have seem some programs write into the cargo cache, sometimes even build files end up in there. Will this cause problems or lead to overeagerly deleted source artifacts?




```
Initial experiments showed about a 35ms performance hit with a HDD with a project with about 200 dependencies, though more investigation will be needed.
```
I'm curious what this looks like if you have a cache of lets say 10-20 GB :upside_down_face:  (not too unrealistic imo, my current cache is around 10 gb but I clean it from time to time)
````
Total:                               9.91 GB
  78 installed binaries:             1.17 GB
  Registry:                          6.11 GB
    4 registry indices:              1.39 GB
    23586 crate archives:            4.20 GB
    1075 crate source checkouts:   527.22 MB
  Git db:                            2.63 GB
    272 bare git repos:              2.57 GB
    9 git repo checkouts:           53.73 MB
````

Let's say I set it to to size limit == 20 gb, but gc=always/every-time, will this slow down common cargo commands significantly?

File: crates/xtask-bump-check/src/xtask.rs (hunk 1 above)
Comment (ehuss) [reply crates/xtask-bump-check/src/xtask.rs:365]: Sure, I have posted #12706 to separate out that change.  I can separate out others as desired as well, just let me know.

File: src/bin/cargo/commands/clean.rs (hunk 4 above)
Comment (ehuss) [reply src/bin/cargo/commands/clean.rs:48]: My thinking is that since it has an optional value, this helps ensure that there isn't a forwards compatibility hazard in case we decide if `cargo clean` should take standalone arguments. It currently doesn't, and I can't imagine it will, but I wanted to be cautious with it. 

I'm also uncertain how that would interact with subcommands (like maybe `cargo clean query`). I'm guessing the ambiguity there is also not important?

Mainly I wanted to be careful since optional arguments can be ambiguous.

I'm not sure if there is a better design here for the CLI option. We could change it to require an argument, or switch to separate CLI options, or maybe remove the options to simplify it. 🤷

File: src/cargo/core/gc.rs (hunk 9 above)
Comment (ehuss) [reply src/cargo/core/gc.rs:255]: I think the overhead of the `.get()` code is fairly heavy in comparison to the actual creation of the value here. If was separate, it would have to go through that mechanism twice in the common case, which I think would be more expensive (or at least a wash).

File: src/cargo/core/gc.rs (hunk 12 above)
Comment (ehuss) [reply src/cargo/core/gc.rs:352]: My preference is to try to keep the syntax supported as simple as possible, so there isn't a lot to document or explain, or to run into ambiguities or confusion with things like spaces, or deal with changes in dependencies that need to be reviewed. I would prefer that this can't be configured at the nanosecond level, or to have arbitrary combinations of units (like "1y 6 weeks").

File: src/cargo/core/gc.rs (hunk 15 above)
Comment (ehuss) [reply src/cargo/core/gc.rs:386]: The split on time was simpler (since it requires a unit, and doesn't allow fractions), so it seemed more straightforward to me to just split the string and parse the left/right sides. I could use a regex there, too, if you want, but I think overall it would be more lines of code. The reason the byte size uses a regex is because it was less code and more straightforward compared to trying to do it all with just str methods.

File: src/cargo/util/sqlite.rs (hunk 19 above)
Comment (ehuss) [reply src/cargo/util/sqlite.rs:88]: I think that is more difficult because we need to be able to tell the user that it is blocking on something. To do that, it needs access to `Config`. We could use a busy handler, but I don't know if that is possible to work due to lifetimes. Either the transaction stuff would need to be done in a separate thread, or cargo's printing system would need to be in a separate thread.

File: src/cargo/core/global_cache_tracker.rs (hunk 20 above)
Comment (ehuss) [reply src/cargo/core/global_cache_tracker.rs:15]: I do not think so, for a few reasons:
* Saving needs to happen before auto-gc runs, so that current information is in the database.
* That would prevent saving for cargo-the-library users.
* I would be nervous trying to ensure that every code path would handle calling `save` at the right time (especially around error handling).

File: src/cargo/core/global_cache_tracker.rs (hunk 25 above)
Comment (ehuss) [reply src/cargo/core/global_cache_tracker.rs:65]: For new columns, newer cargos will need to assume that the column might be filled in with the default (which can be NULL if that is how it is defined).  For new tables, it will just need to assume that the data is not managed by older cargos.

File: src/cargo/core/global_cache_tracker.rs (hunk 34 above)
Comment (ehuss) [reply src/cargo/core/global_cache_tracker.rs:879]: Just to clarify, are you suggesting that on each `execute()` or similar call, that there also be something like:


```rust
trace!("inserting untracked crate file {crate_name} {size} {now}");
```

?

My concern is that could be a pretty substantial amount of log data.

src/cargo/ops/cargo_fetch.rs
ehuss: I don't think there is really a before/after difference here. Gc always happens after downloading. The sequence is generally:

* `cargo fetch` — fetch index, resolve, download, gc
* `cargo build` — fetch index, resolve, download, gc, start compiling

It is not done after compiling to make it easier to deal with error handling (I felt like it was better to not co-mingle the two, or to require compiling to succeed, or to perform some action after compiling failed).

ehuss:
@matthiaskrgr:

>  Maybe "artifact" would be a better term?

"Artifact" is already used by cargo to refer to the compilation outputs.

I understand it can maybe be a little confusing to say "files" in the generic sense, when it may not be clear that cargo only tracks individual "src" directories as a unit (not the individual files within). I'm not sure if there is a better way here, though I can try to tweak some of the text.

> So can cargo only track changes that are done by itself?

Mostly. It has a synchronization step where it will detect if files are added or removed, such as by an older version of cargo.
(It doesn't track individual files in the "src" directory, just the directory as a whole.)

> I have seem some programs write into the cargo cache, sometimes even build files end up in there. Will this cause problems or lead to overeagerly deleted source artifacts?

Unless those programs are using cargo-the-library, then they probably shouldn't be messing with cargo's cache directories. :wink: I realize that `cargo-cache` is interacting with it, and that's fine, just be aware of the consequences of modifying cargo's internal files. The format and layout was never intended to be something stable that other programs should be interacting with.

Also, because the default automatic deletion is on the timespan of months, cargo shouldn't be too aggressive with deleting anything. Worst case, cargo will just need to re-download something.

> I'm curious what this looks like if you have a cache of lets say 10-20 GB

I've been doing most of my testing with a very large cache. I also do tests with dependency trees of various sizes (up to 500 dependencies).

```
Total:                                  19.09 GB
  46 installed binaries:               389.73 MB
  Registry: github.com                  11.12 GB
    Registry index:                    950.07 MB
    12711 crate archives:                1.66 GB
    12711 crate source checkouts:        8.50 GB
  Registry: github.com                 385.40 MB
    Registry index:                    385.40 MB
  Registry: github.com                  46.54 KB
    Registry index:                     46.54 KB
  Registry: index.crates.io              4.70 GB
    Registry index:                    118.58 MB
    4534 crate archives:               687.69 MB
    4534 crate source checkouts:         3.90 GB
  Git db:                                2.10 GB
    42 bare git repos:                 422.32 MB
    49 git repo checkouts:               1.68 GB
```

> Let's say I set it to to size limit == 20 gb, but gc=always/every-time, will this slow down common cargo commands significantly?

Just to be clear, size limits are only available for manual gc. Automatic currently only supports time-based.

src/cargo/core/global_cache_tracker.rs
ehuss: I added some more information about the opportunistic behavior of automatic gc.

src/cargo/core/global_cache_tracker.rs
ehuss: Synchronization happens on both auto and manual gc. The important difference is that size checking is only done if size restrictions are enabled, which is only possible with manual gc.

The comment on `sync_db_with_files` was out-of-date, since I had done some changes along the way.

src/cargo/core/global_cache_tracker.rs
ehuss: It isn't a duplicate, the first word is *above* with a `v`. It's a little awkward, though, so I have reworded this.

src/cargo/core/global_cache_tracker.rs
ehuss: I can add resolving those to the tracking issue.

FWIW, I did a little bit of testing on this particular TODO already, but I don't have the exact numbers anymore, so I will need to re-run them.

src/cargo/core/global_cache_tracker.rs
ehuss: Yea, the comment on `sync_db_with_files` was out-of-date. It always runs the sync, which is not too terribly slow. The slow part is the `size` scanning, which should only be done if requested in a manual gc.

src/cargo/core/global_cache_tracker.rs
ehuss: I can add resolving those to the tracking issue.

src/bin/cargo/commands/clean.rs
ehuss: This was written before #12638, where it was decided to go ahead and stabilize the dry-run option. After rebasing, this should be resolved.

tests/testsuite/clean.rs
ehuss: This was written before #12638, where it was decided to go ahead and stabilize the dry-run option. After rebasing, this should be resolved.

tests/testsuite/global_cache_tracker.rs
ehuss: Similar, this was just an issue with this being behind #12638.

src/doc/src/reference/unstable.md
ehuss: It runs just after resolution and packages are downloaded, but before compilation. I added a comment to that effect.

src/doc/src/reference/unstable.md
```diff
@@ -1665,6 +1666,66 @@ version = "0.1.0"
...
+cargo clean --max-git-size=0 --max-download-size=100MB
+```
+
+The automatic gc behavior can be specified via a cargo configuration setting.
+The settings available are:
+
+```toml
+# The maximum frequency that automatic garbage collection happens.
+# Can be "never" to disable automatic-gc, or "always" to run on every command.
+frequency = "1 day"
```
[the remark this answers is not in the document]
ehuss: Sorry, I accidentally forgot to add that. It is `[gc.auto]`.

ehuss:
I'm still working on addressing some of the comments, it may be a few more days.

@rustbot author

src/bin/cargo/commands/clean.rs
epage: Since the CLI is not being decided on at this time, we can pass over this.

src/cargo/core/gc.rs
epage: Fair enough

crates/cargo-test-support/src/paths.rs
```diff
@@ -155,6 +159,15 @@ impl CargoPathExt for Path {
             .unwrap_or_else(|e| panic!("failed to mkdir_p {}: {}", self.display(), e))
     }
 
+    fn ls_r(&self) -> Vec<PathBuf> {
+        let mut file_list: Vec<_> = walkdir::WalkDir::new(self)
+            .into_iter()
+            .filter_map(|e| e.map(|e| e.path().to_owned()).ok())
+            .collect();
+        file_list.sort();
```
epage, line 167: There a reason to manually sort rather than use the sorting built-in to `WalkDir`?

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,557 @@
...
+/// ignored), but try not to abuse that.
+pub fn auto_gc(config: &Config) {
+    if !config.cli_unstable().gc {
+        return;
+    }
+    if !config.network_allowed() {
+        // As a conservative choice, auto-gc is disabled when offline. If the
+        // user is indefinitely offline, we don't want to delete things they
+        // may later depend on.
+        tracing::trace!("auto gc disabled, running offline");
```
epage, line 51: nit: compared to how the other traces are structured, this should be `"running offline, auto gc disabled"`

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,557 @@
...
+    /// been performed recently (since `gc.auto.frequency`).
+    fn auto(&mut self, clean_ctx: &mut CleanContext<'config>) -> CargoResult<()> {
+        if !self.config.cli_unstable().gc {
+            return Ok(());
+        }
+        let auto_config = self
+            .config
+            .get::<Option<AutoConfig>>("gc.auto")?
+            .unwrap_or_default();
+        let Some(freq) = parse_frequency(auto_config.frequency.as_deref().unwrap_or("1 day"))?
```
epage, line 320: nit: I feel like this constant should be with the definition of `AutoConfig` or pulled out into a separate place with the others

src/cargo/core/gc.rs
epage: The asymmetry between
```rust
        if !clean_ctx.dry_run {
            self.global_cache_tracker.set_last_auto_gc()?;
        }
```
and
```rust
self.global_cache_tracker.clean(clean_ctx, gc_opts)?;
```
is what I'm calling out

If the API doesn't predictably handle dry_run, that will more likely lead to bugs.

src/cargo/core/gc.rs
```diff
@@ -0,0 +1,557 @@
...
+    let factor = match cap.get(3) {
+        Some(suffix) => match suffix.as_str().to_lowercase().as_str() {
+            "b" => 1.0,
+            "kb" => 1_000.0,
+            "mb" => 1_000_000.0,
+            "gb" => 1_000_000_000.0,
+            "kib" => 1024.0,
+            "mib" => 1024.0 * 1024.0,
+            "gib" => 1024.0 * 1024.0 * 1024.0,
+            s => panic!("suffix `{s}` out of sync with regex"),
```
epage, line 465: nit: `unreachable` rather than `panic` to communicate the intent

src/cargo/ops/cargo_fetch.rs
epage: I think my comment was future-looking to when we do `target/` GC but we can look into it further then

src/doc/src/reference/unstable.md
```diff
@@ -1665,6 +1666,66 @@ version = "0.1.0"
...
+Automatic deletion happens on commands that are already doing a significant amount of work,
+such as all of the build commands (`cargo build`, `cargo test`, `cargo check`, etc.), and `cargo fetch`.
+Automatic deletion is only done once per day.
+Automatic deletion is disabled if cargo is offline such as with `--offline` or `--frozen` to avoid deleting files that may need to be used if you are offline for a long period of time.
+
+Manual deletion can be done with the `cargo clean` command.
+Deletion of cache contents can be performed by passing one of the cache options:
+
+- `--gc` --- Performs the same garbage collection that is performed by the once-a-day automatic deletion.
+- `--max-src-age=DURATION` --- Deletes source cache files that have not been used since the given age.
```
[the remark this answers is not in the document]
epage: For myself, I'm leaning out all fine discussion over the CLI and focusing on the core of this.  As part of the path towards stabiliztion, we'll want to layout CLI options and evaluate what we want to go with

src/cargo/core/global_cache_tracker.rs
epage: Yes, that was what I intended.  As I'm not entirely sure how much it'll pay off, I'm fine holding off for now due to the traffic that'd cause

src/bin/cargo/commands/clean.rs
ehuss: I removed the shared-target stuff, since that was a little far-fetched. However, I would like to keep the "target" part to indicate how it is intended to be used, and I think helps indicate how the pieces are intended to fit together. Otherwise, various parts I think would not really be clear as to why it is done like it is. I also don't expect the "target" part to be as far out, and has a rough plan for getting there.

src/cargo/core/gc.rs
ehuss: Hm, I'm not sure I see a better way to particularly do that. The `dry_run` flag could be passed to `set_last_auto_gc` instead of checking it outside, but I don't quite see how that would help. Did you have any specific ideas on what could be different there?

src/cargo/core/global_cache_tracker.rs
ehuss: I don't think it is called often enough to be noticeable from a performance standpoint. Passing the timestamp around ends up adding more lines of code which I would prefer to avoid.

As for tracking something in `Config`, I considered that, but I'd be a bit uneasy with dealing with long-lived programs or handling before/after relationships (like different parts of the code not considering that all timestamps might be equal). I think it's worth considering, but I'm not sure this specific PR really justifies it balanced with the risks.

src/cargo/sources/git/source.rs
ehuss: Good call, I pulled that out into a separate function. Ideally it would not be necessary (as the TODO mentions), and is hopefully not too hard to fix if someone feels inclined.

src/cargo/core/gc.rs
ehuss: I pulled these into constants, though I'm not sure I feel entirely comfortable with that. My thinking is that we may want to tweak these defaults in ways that are unique to each setting (like maybe git db should be longer than `.crate`, since it can be much more expensive to download). But I also don't really know if that will actually happen, and the structure can be tweaked if necessary.

crates/cargo-test-support/src/paths.rs
ehuss: Oh, I just didn't know that existed, thanks!

src/cargo/core/gc.rs
ehuss: I moved it to a module-level constant.

I'm not particularly happy with the way the defaults are defined, but I don't know of a better way. It could use serde defaults like:

```rust
struct AutoConfig {
    /// The maximum frequency that automatic garbage collection happens.
    #[serde(default="default_frequency")]
    frequency: String,
    #[serde(default="default_max_src_age")]
    max_src_age: String,
    // etc....
}

fn default_frequency() -> String {
    "1 day".to_string()
}

fn default_max_src_age() -> String {
    "1 month".to_string()
}
```

but that seems worse somehow (lots of lines of code, lots of unnecessary string allocations, though that is minor). Unless I'm missing something about serde.

Or, it could have accessors like:

```rust
fn frequency(&self) -> &str {
    self.frequency.as_deref().unwrap_or("1 day")
}
```

But again, that adds 20-30 lines of code.  I could do that, but since all fields are accessed in exactly one place, I'm not sure if it helps.

I could make those changes if you want, or if you have other ideas on how to deal with config defaults.

ehuss:
I think I have addressed most of the comments.

@rustbot ready

src/bin/cargo/commands/clean.rs
epage: How do you expect to show these not-implemented parts?
- No behavior implemented, so interaction doesn't work
- They are hidden in help, so people can't browse it
- People can construct command-lines and see them panic but they can just construct them and see them fail with clap messages.

What if `clean::cli` checked the channel and only hid them if its stable?

src/cargo/core/gc.rs
epage: Thought of another angle of this: layering.  I assume whatever you get gets layered together, so if we get `gc.auto` that means if you override one field in a config, you mask all of the other fields.  Is that correct?

How much do we care though?

src/cargo/core/gc.rs
epage: I think I'd prefer a `dry_run` flag for `set_last_auto_gc` but it is a messy situation that my preference isn't strong enough to hold this back.

src/cargo/core/global_cache_tracker.rs
epage: I'm also a bit uneasy about the different `now`s used but I can't think of a concrete problem.

And cargo-as-a-library is already shot for long running applications because of `InternedString`.

However, this is likely not enough of a deal that we can pass on this.

src/cargo/core/gc.rs
epage: For what this is doing, the `const`s are fine.

epage:
Overall, looks good (excepting any discussion on the CLI).

I do have a couple of unreplied-to comments in the Hidden section.

src/bin/cargo/commands/clean.rs
epage: #6102 is somewhat related to this.

ehuss: I had either two intentions:
* Anything unimplemented would remain hidden when anything else is stabilized. The only people who would encounter it is someone reading the code.
* Unimplemented parts would get removed when moving towards stabilization if they are in the way or causing a hassle.

A big part of the path for this feature is figuring out what the CLI would look like. I am expecting the CLI to change, so I didn't want to put too much effort into polishing it right now.

src/bin/cargo/commands/clean.rs
ehuss: I'm not sure I understand what this comment is getting towards. Our general policy is that the command level is responsible for parsing the CLI flags, and then handing off all the logic to the ops level. Since there is some important logic for deciding what actions need to be taken based on how the options combine together, I think it is good to keep that away from the command level. For example, I wouldn't want it to be possible to pass in an invalid configuration to the `ops::clean()` function. If that logic for whether or not the workspace is loaded was moved to `clean::exec()`, it would have to be duplicated again to validate that the caller set up the options correctly.

src/cargo/core/gc.rs
ehuss: If by masking you mean they get set to `None` in the deserialized form, yea that's how tables get merged.

epage: Seems sub-optimal to merge on the table level, rather than the field level, but in working on #12889, I realized we already have precedence for this.

src/bin/cargo/commands/clean.rs
epage: Not quite what I was getting at.  The unimplemented flags are here to "to indicate how it is intended to be used".  I originally took this to mean at the CLI level but at the CLI level they aren't exposed.  Do you instead mean at `ops` level?  If so, I don't feel its all that needed.  While the CLI isn't finalized and so we can let everything go, I feel like adding a bunch of code muddies the waters and made this review much bigger than it should be (even if doing the work helped to work around design issues).

src/bin/cargo/commands/clean.rs
epage: > I'm not sure I understand what this comment is getting towards.

For being a placeholer, this feels fairly intrusive.  Seems like we could switch between different `ops` calls within the CLI.

> For example, I wouldn't want it to be possible to pass in an invalid configuration to the ops::clean() function. If that logic for whether or not the workspace is loaded was moved to clean::exec(), it would have to be duplicated again to validate that the caller set up the options correctly.

If we made the modes mutually exclusive (rather than additive) such that two separate ops can be decided between, than we can have something in and more easily talk through what the CLI should be, even if it ends up in `cargo clean`.  For example, I was just looking at the details of the `exec` and `ops` to check-in on what I was saying and noticed a subtle interaction that I'd want to talk about for the CLI that could easily be overlooked in the current semi-placeholder, semi-proposal route which would become obvious in a PR that takes a pure-placeholder and turns it into a proposal.

I have to say that I also have a bias in this thread coming from me not liking this being in `cargo clean` so to me this feels more like throwaway code and too much time can be spent on the details of it despite that (and later removing it)

src/bin/cargo/commands/clean.rs
ehuss: The "how it is intended to be used" was in regards to how the interface of `ops::clean()` works. So, for example, when cargo is instructed to clean caches, the interface of `ops::clean()` needs a variety of options to tell it what to clean. Those options include global cache, target cache, and potentially other caches as well. That's the purpose of the `AutoGcKind` enum. It is intended to show why `AutoGcKind` exists, how the detailed options interact, and how it all fit together, why it exists, how it is intended to evolve, etc.

I'm not sure I see it being a bunch of code. I see maybe 30 lines total in this file, which seems like a relatively minor part? Or maybe I'm not understanding or seeing what the concern is.

My concern is that removing the parts that this is intended to have will make it not as clear why it is structured the way it is, and the removed parts will just need to be re-added again. Keeping them here makes it easier to maintain in my opinion.

src/bin/cargo/commands/clean.rs
ehuss: Ah, so perhaps we had a misunderstanding here, or at least a disagreement. I'm not seeing this as a "placeholder". This is a proposed interface that is intended to work generally as it is, but that I expect to *evolve* the specifics. For example, perhaps individual flags will be removed (like being able to control src/git/index separately), or changed to be higher level (like "delete things that can be recreated without a network connection") or maybe there is only one extra flag (like `--gc`). But my intention is very much that there is a single command for controlling all caches, and that the CLI interface has some options to control which caches are cleaned (with some default if no flags are specified).

So for me, this is not throwaway code. This is how I intend it to work. The open part is just the specific details of flag names and how general or specific they are.

epage: At this time, I am strongly against stabilizing this in `cargo clean`.  Whichever way it goes, our different stances show we have fairly fundamental conversations to have around the UX which means anything can change.  As such I've been taking the approach of "focus on the core with the UX being a mechanism to exercise the core".

Sorry, I thought I had made it clear this clear at the outset.

src/bin/cargo/commands/clean.rs
epage: If this is meant to show the interface for `ops::clean`, I think this then comes back to our other conversation about fundamentals vs details in UX (https://github.com/rust-lang/cargo/pull/12634#discussion_r1378171958).  For me, since I view the UX under review to be a `yeet`-like CLI that exists for the sake of exercising the core, it seem weird for that CLI to go "deeper".

weihanglo:
What's the progress of this? @epage and @ehuss, do you want an extra pair of eyes helping review this?

epage:
We talked over this with a small list of action items.  Once those are taken care of, we were going to merge this.

src/bin/cargo/commands/clean.rs
```diff
@@ -19,223 +20,128 @@ pub fn cli() -> Command {
...
-            .value_name("SIZE")
-            .value_parser(parse_human_size)
-            .hide(true),
+        .args_conflicts_with_subcommands(true)
+        .subcommand(
+            subcommand("gc")
+                .about("Clean global caches")
+                .hide(true)
+                // FIXME: arg_quiet doesn't work because `config_configure`
+                // doesn't know about subcommands.
```
epage, line 29: Side note: what do we do about `cargo report`?  I assume this would also affect #11879?

epage:
Thanks!

@bors r+

src/bin/cargo/commands/clean.rs
ehuss: So far, `cargo report` doesn't have a need for the `-q` flag because its only purpose is to display something on the console, so there isn't any output that needs suppressing. But it is something we should consider moving forward.

I think it will be a problem for #11879. I don't really know the best way to fix this. I don't suppose clap has a way to say "if this arg appears anywhere"?  Perhaps another option is to have each subcommand manually handle setting quiet/verbose settings?  Do you have any ideas?

epage: clap has `global(true)` which propagates a flag definition down through the commands and then propagates the parsed value back up and is used by all of the flags in `cli.rs`'s `cli()`.

src/doc/src/reference/unstable.md
matthiaskrgr: Ah sorry, only now saw he reply (too much github things going on at a time :/ )

My assumption was that the the automatic gc would trigger every time, lets say after running "cargo build"

so cargo build would 
1) download the 1 Gb git repo
2) build the project
3) run auto gc
3.1) notice that 1 gb  exceeds the limit of 500 mb
3.2) delete the git repo from the cache

=> a subsequent "cargo build" would have to download the source again then?

src/bin/cargo/commands/clean.rs
ehuss: I opened https://github.com/rust-lang/cargo/issues/12957 to continue this conversation. We currently don't use `global(true)` because we want to be able to customize the help text per-command.

weihanglo:
Thank you @ehuss  and @epage for your efforts for this!

src/doc/src/reference/unstable.md
epage: For "cargo script", we'll need to GC their target directory, especially if the script no longer exists.

We could instead track all workspaces.  One way we could use this is we could look at all `Cargo.lock` files and pin the entries in the cache so they can't be GCed.  This should solve that problem.  Of course, I have no idea how fast/slow this might be so no idea if we'll actually do it.

hi-rustin:
Thanks for working on it! ❤️ :shipit:

src/doc/src/reference/unstable.md
matthiaskrgr: My question is not about the artifacts that are produced by a build but about the dependencies that are downloaded before a build :sweat_smile:

epage: Thats what I'm referring to: we scan all `Cargo.lock`s for registry entries in use and don't clean them up.

PR #12635, opened 2023-09-07
Add with_stdout_unordered.

ehuss:
This adds the `with_stdout_unordered` method to cargo's test system so that tests can use it to check stdout but ignoring the order of lines. Nothing in this PR actually uses this method, but it is added to support #12634. I also expect it could potentially be useful in other cases in the future.


weihanglo (approved):

weihanglo:
Thanks for the addition!

@bors r+

PR #12636, opened 2023-09-07
Add wrappers around std::fs::metadata

ehuss:
This adds wrappers around `std::fs::metadata` and `std::fs::symlink_metadata` which provide better error messages indicating the path that caused the error. This just helps clean up some duplicated code, and is also going to be used to assist with some code changes in #12634.

### DIFF
--- a/crates/cargo-util/src/paths.rs
+++ b/crates/cargo-util/src/paths.rs
@@ -4,7 +4,7 @@ use anyhow::{Context, Result};
 use filetime::FileTime;
 use std::env;
 use std::ffi::{OsStr, OsString};
-use std::fs::{self, File, OpenOptions};
+use std::fs::{self, File, Metadata, OpenOptions};
 use std::io;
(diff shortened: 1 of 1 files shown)


weihanglo (approved):

weihanglo:
Thanks for the incremental enhancments!

@bors r+

crates/cargo-util/src/paths.rs
```diff
@@ -136,6 +136,24 @@ pub fn resolve_executable(exec: &Path) -> Result<PathBuf> {
     }
 }
 
+/// Returns metadata for a file (follows symlinks).
+///
+/// Equivalent to [`std::fs::metadata`] with better error messages.
+pub fn metadata<P: AsRef<Path>>(path: P) -> Result<Metadata> {
```
epage, line 142: Any thoughts on using something like [fs-err](https://crates.io/crates/fs-err) instead?

Issue #8790, opened 2020-10-16
`cargo clean -p` does not clean documentation


jyn514:
**Problem**

`cargo clean -p package` should remove `target/doc/package`, but does not.

**Steps**

```
$ cargo new test-doc
     Created binary (application) `test-doc` package
$ cd test-doc
Cargo.toml  src
$ cargo doc
 Documenting test-doc v0.1.0 (/home/joshua/src/rust/test-doc)
    Finished dev [unoptimized + debuginfo] target(s) in 0.70s
$ ls target/doc
ayu.css                main.js                      source-files.js
brush.svg              normalize.css                source-script.js
COPYRIGHT.txt          noscript.css                 SourceSerifPro-Bold.ttf.woff
dark.css               rustdoc.css                  SourceSerifPro-It.ttf.woff
down-arrow.svg         rust-logo.png                SourceSerifPro-LICENSE.md
favicon.ico            search-index.js              SourceSerifPro-Regular.ttf.woff
FiraSans-LICENSE.txt   settings.css                 src
FiraSans-Medium.woff   settings.html                storage.js
FiraSans-Regular.woff  settings.js                  test_doc
LICENSE-APACHE.txt     SourceCodePro-LICENSE.txt    theme.js
LICENSE-MIT.txt        SourceCodePro-Regular.woff   wheel.svg
light.css              SourceCodePro-Semibold.woff
$ cargo clean -p test-doc
$ ls target/doc
ayu.css                main.js                      source-files.js
brush.svg              normalize.css                source-script.js
COPYRIGHT.txt          noscript.css                 SourceSerifPro-Bold.ttf.woff
dark.css               rustdoc.css                  SourceSerifPro-It.ttf.woff
down-arrow.svg         rust-logo.png                SourceSerifPro-LICENSE.md
favicon.ico            search-index.js              SourceSerifPro-Regular.ttf.woff
FiraSans-LICENSE.txt   settings.css                 src
FiraSans-Medium.woff   settings.html                storage.js
FiraSans-Regular.woff  settings.js                  test_doc
LICENSE-APACHE.txt     SourceCodePro-LICENSE.txt    theme.js
LICENSE-MIT.txt        SourceCodePro-Regular.woff   wheel.svg
light.css              SourceCodePro-Semibold.woff
$ cargo clean
$ ls target/doc
ls: cannot access 'target/doc': No such file or directory
```

**Possible Solution(s)**

**Notes**

Output of `cargo version`:

[...]

PR #12637, opened 2023-09-07
Error out if `cargo clean --doc` is mixed with `-p`.

ehuss:
This changes `cargo clean --doc -p foo` to generate an error instead of ignoring the `-p` flag. There is still an outstanding issue https://github.com/rust-lang/cargo/issues/8790 tracking this. It *should* support `-p`, but until it is supported, I think cargo should tell you that the flag is ignored. This is also in preparation for some code changes in #12634 which needs to handle any combination of various different clean flags.

### DIFF
--- a/src/cargo/ops/cargo_clean.rs
+++ b/src/cargo/ops/cargo_clean.rs
@@ -7,7 +7,7 @@ use crate::util::errors::CargoResult;
 use crate::util::interning::InternedString;
 use crate::util::{Config, Progress, ProgressStyle};
 
-use anyhow::Context as _;
+use anyhow::{bail, Context as _};
 use cargo_util::paths;
 use std::fs;
 use std::path::Path;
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
This may break somebody's script, but I believe that's rare and easy to fix.

src/cargo/ops/cargo_clean.rs
```diff
@@ -33,6 +33,15 @@ pub fn clean(ws: &Workspace<'_>, opts: &CleanOptions<'_>) -> CargoResult<()> {
 
     // If the doc option is set, we just want to delete the doc directory.
     if opts.doc {
+        if !opts.spec.is_empty() {
+            // FIXME: https://github.com/rust-lang/cargo/issues/8790
```
epage, line 37: This culd be done with clap but sicne this is just a fixme, it likely doesn't matter

PR #12638, opened 2023-09-07
Add some enhancements to `cargo clean`

ehuss:

This adds some enhancements to `cargo clean` that fell out as a result of some refactorings in #12634 for supporting an interface for cleaning from other places in cargo, and these were relatively easy to add and assist with testing in #12634. 

The changes are:
- Introduce some refactoring to offer a cleaning interface that can be used elsewhere in cargo.
- Adds a `--dry-run` CLI option which will print what `cargo clean` will delete without deleting it. **NOTE** This PR makes the flag insta-stable. I don't figure there is too much that can be learned about it keeping it unstable, though we could change that. #12634 has this flag gated with `-Zgc`.
- Adds a summary line at the end of the `cargo clean` operation that indicates how much was deleted.

Note that this PR also includes the changes from #12635 and #12637. Those commits can be dropped if those PRs are merged.

For the most part, this involves wrapping the cleaning operations in a `CleanContext` which provides an interface for performing cleaning operations. The dry run is just a flag that is checked at the deletion points. The summary data is also collected at those same points.

### DIFF
--- a/src/bin/cargo/commands/clean.rs
+++ b/src/bin/cargo/commands/clean.rs
@@ -14,6 +14,13 @@ pub fn cli() -> Command {
         .arg_target_triple("Target triple to clean output for")
         .arg_target_dir()
         .arg_manifest_path()
+        .arg(
+            flag(
+                "dry-run",
+                "Display what would be deleted without deleting anything",
+            )
+            .short('n'),
+        )
         .after_help(color_print::cstr!(
             "Run `<cyan,bold>cargo help clean</>` for more detailed information.\n"
         ))
@@ -33,6 +40,7 @@ pub fn exec(config: &mut Config, args: &ArgMatches) -> CliResult {
         requested_profile: args.get_profile_name(config, "dev", ProfileChecking::Custom)?,
         profile_specified: args.contains_id("profile") || args.flag("release"),
         doc: args.flag("doc"),
+        dry_run: args.flag("dry-run"),
     };
     ops::clean(&ws, &opts)?;
     Ok(())
--- a/src/bin/cargo/commands/clean.rs
+++ b/src/bin/cargo/commands/clean.rs
@@ -14,13 +14,7 @@ pub fn cli() -> Command {
         .arg_target_triple("Target triple to clean output for")
         .arg_target_dir()
         .arg_manifest_path()
-        .arg(
-            flag(
-                "dry-run",
-                "Display what would be deleted without deleting anything",
-            )
-            .short('n'),
-        )
+        .arg_dry_run("Display what would be deleted without deleting anything")
         .after_help(color_print::cstr!(
             "Run `<cyan,bold>cargo help clean</>` for more detailed information.\n"
         ))
@@ -40,7 +34,7 @@ pub fn exec(config: &mut Config, args: &ArgMatches) -> CliResult {
         requested_profile: args.get_profile_name(config, "dev", ProfileChecking::Custom)?,
         profile_specified: args.contains_id("profile") || args.flag("release"),
         doc: args.flag("doc"),
-        dry_run: args.flag("dry-run"),
+        dry_run: args.dry_run(),
     };
     ops::clean(&ws, &opts)?;
     Ok(())
--- a/src/cargo/ops/cargo_clean.rs
+++ b/src/cargo/ops/cargo_clean.rs
@@ -417,7 +417,13 @@ impl<'cfg> CleanContext<'cfg> {
         };
         self.config
             .shell()
-            .status(status, format!("{file_count}{byte_count}"))
+            .status(status, format!("{file_count}{byte_count}"))?;
+        if self.dry_run {
+            self.config
+                .shell()
+                .warn("no files deleted due to --dry-run")?;
+        }
+        Ok(())
     }
 
     /// Deletes all of the given paths, showing a progress bar as it proceeds.
--- a/src/cargo/ops/cargo_clean.rs
+++ b/src/cargo/ops/cargo_clean.rs
@@ -23,11 +23,14 @@ pub struct CleanOptions<'cfg> {
     pub requested_profile: InternedString,
     /// Whether to just clean the doc directory
     pub doc: bool,
+    /// If set, doesn't delete anything.
+    pub dry_run: bool,
 }
 
 pub struct CleanContext<'cfg> {
     pub config: &'cfg Config,
     progress: Box<dyn CleaningProgressBar + 'cfg>,
+    pub dry_run: bool,
 }
 
 /// Cleans various caches.
@@ -35,6 +38,7 @@ pub fn clean(ws: &Workspace<'_>, opts: &CleanOptions<'_>) -> CargoResult<()> {
     let mut target_dir = ws.target_dir();
     let config = opts.config;
     let mut ctx = CleanContext::new(config);
+    ctx.dry_run = opts.dry_run;
 
     if opts.doc {
         if !opts.spec.is_empty() {
@@ -262,6 +266,7 @@ impl<'cfg> CleanContext<'cfg> {
         CleanContext {
             config,
             progress: Box::new(progress),
+            dry_run: false,
         }
     }
 
@@ -322,26 +327,48 @@ impl<'cfg> CleanContext<'cfg> {
             }
         };
 
-        self.config
-            .shell()
-            .verbose(|shell| shell.status("Removing", path.display()))?;
+        if self.dry_run {
+            // Concise because if in verbose mode, the path will be written in
+            // the loop below.
+            self.config
+                .shell()
+                .concise(|shell| Ok(writeln!(shell.out(), "{}", path.display())?))?;
+        } else {
+            self.config
+                .shell()
+                .verbose(|shell| shell.status("Removing", path.display()))?;
+        }
         self.progress.display_now()?;
 
+        let rm_file = |path: &Path| {
+            if !self.dry_run {
+                paths::remove_file(path)?;
+            }
+            Ok(())
+        };
+
         if !meta.is_dir() {
-            return paths::remove_file(path);
+            return rm_file(path);
         }
 
         for entry in walkdir::WalkDir::new(path).contents_first(true) {
             let entry = entry?;
             self.progress.on_clean()?;
+            if self.dry_run {
+                self.config
+                    .shell()
+                    .verbose(|shell| Ok(writeln!(shell.out(), "{}", entry.path().display())?))?;
+            }
             if entry.file_type().is_dir() {
                 // The contents should have been removed by now, but sometimes a race condition is hit
                 // where other files have been added by the OS. `paths::remove_dir_all` also falls back
                 // to `std::fs::remove_dir_all`, which may be more reliable than a simple walk in
                 // platform-specific edge cases.
-                paths::remove_dir_all(entry.path())?;
+                if !self.dry_run {
+                    paths::remove_dir_all(entry.path())?;
+                }
             } else {
-                paths::remove_file(entry.path())?;
+                rm_file(entry.path())?;
             }
         }
 
--- a/src/cargo/ops/cargo_clean.rs
+++ b/src/cargo/ops/cargo_clean.rs
@@ -262,7 +262,7 @@ fn escape_glob_path(pattern: &Path) -> CargoResult<String> {
 }
 
 impl<'cfg> CleanContext<'cfg> {
-    pub fn new(config: &'cfg Config) -> CleanContext<'cfg> {
+    pub fn new(config: &'cfg Config) -> Self {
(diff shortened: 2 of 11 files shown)


src/bin/cargo/commands/clean.rs
```diff
@@ -14,6 +14,13 @@ pub fn cli() -> Command {
         .arg_target_triple("Target triple to clean output for")
         .arg_target_dir()
         .arg_manifest_path()
+        .arg(
+            flag(
+                "dry-run",
+                "Display what would be deleted without deleting anything",
+            )
+            .short('n'),
+        )
```
weihanglo, line 23: And maybe we should bikshed the short alias `-n`?

```suggestion
        .arg_dry_run("Display what would be deleted without deleting anything")
```

src/cargo/ops/cargo_clean.rs
```diff
@@ -243,92 +254,124 @@ fn escape_glob_path(pattern: &Path) -> CargoResult<String> {
...
-            .0;
-
-        if pkg_name != package {
-            continue;
+impl<'cfg> CleanContext<'cfg> {
+    pub fn new(config: &'cfg Config) -> CleanContext<'cfg> {
+        // This progress bar will get replaced, this is just here to avoid needing
+        // an Option until the actual bar is created.
+        let progress = CleaningFolderBar::new(config, 0);
+        CleanContext {
```
epage, line 269: nit: `Self` rather than `CleanContext`

src/bin/cargo/commands/clean.rs
```diff
@@ -31,6 +38,7 @@ pub fn exec(config: &mut Config, args: &ArgMatches) -> CliResult {
         requested_profile: args.get_profile_name(config, "dev", ProfileChecking::Custom)?,
         profile_specified: args.contains_id("profile") || args.flag("release"),
         doc: args.flag("doc"),
+        dry_run: args.flag("dry-run"),
```
weihanglo, line 41: ditto

```suggestion
        dry_run: args.dry_run()
```

weihanglo (commented):
Saw Ed is reviewing. I'll step back and take a rest 😁

src/cargo/ops/cargo_clean.rs
```diff
@@ -322,26 +327,48 @@ impl<'cfg> CleanContext<'cfg> {
...
+            // Concise because if in verbose mode, the path will be written in
+            // the loop below.
+            self.config
+                .shell()
+                .concise(|shell| Ok(writeln!(shell.out(), "{}", path.display())?))?;
+        } else {
+            self.config
+                .shell()
+                .verbose(|shell| shell.status("Removing", path.display()))?;
+        }
```
epage: Is there a reason `status` isn't being used?  Feels odd to dump paths to the screen without anything else

src/cargo/ops/cargo_clean.rs
```diff
@@ -23,18 +23,22 @@ pub struct CleanOptions<'cfg> {
     pub requested_profile: InternedString,
     /// Whether to just clean the doc directory
     pub doc: bool,
+    /// If set, doesn't delete anything.
+    pub dry_run: bool,
```
epage, line 27: Generally our commands with a `--dry-run` have a `shell().warn("aborting add due to dry run")?;`

epage:
> NOTE This PR makes the flag insta-stable. I don't figure there is too much that can be learned about it keeping it unstable, though we could change that. https://github.com/rust-lang/cargo/pull/12634 has this flag gated with -Zgc.

Agreed that there isn't too much to learn.  We also have a precedence for this and I'd love to see us extend that precedence further.

src/bin/cargo/commands/clean.rs
epage: I'd love to add `-n` as a short alias as there is a strong precedence for it but I agree that should be done separately an we should be using `arg_dry_run`

src/cargo/ops/cargo_clean.rs
```diff
@@ -368,13 +381,35 @@ impl<'cfg> CleanContext<'cfg> {
...
+                format!(", {}B total", self.total_bytes_removed)
+            } else {
+                let (bytes, unit) = human_readable_bytes(self.total_bytes_removed);
+                format!(", {bytes:.1}{unit} total")
+            }
+        };
+        self.config.shell().status(
+            status,
+            format!(
+                "{} files/directories{byte_count}",
```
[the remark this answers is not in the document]
ehuss: I included directories since I wanted it to communicate something if it only deleted directories (instead of saying "0 files"), since that wouldn't accurately reflect what it did. I agree the phrasing is a little awkward. It could possibly track them separately (like `N files, M directories, XMB total`). Or it could special case it so that it normally just says `N files, XMB total`, but if there are 0 files and nonzero directories, it could say `0 files, M directories`.

src/cargo/ops/cargo_clean.rs
ehuss: My intent was to have the "Summary" line communicate that it didn't actually delete anything (instead of saying "Removed"). I would prefer to avoid the term "aborting" since it isn't actually aborting anything. Would adding something like `(no files deleted due to --dry-run)` to the summary work?

src/cargo/ops/cargo_clean.rs
ehuss: The intent with `--dry-run` is to print a list of files that it would delete (that's all it does). I also didn't think there was a particular status prefix that would fit in this situation. I also presume that most paths are going to be long, and adding to that could make it harder to see. But I'm not completely opposed if there are some suggestions on what to print.

epage: I feel like we should be consistent in telling people what we are removing even if we then have additional dry-run specific messages listing all of the details of what was removed.

epage: Or to put it another way, maybe another enhancement is for the `status` to not report a path but report a description and then when we extend this to GC, it can be "Removing .crate files", "Removing git repos", etc.

src/cargo/ops/cargo_clean.rs
epage: For myself, I'd like us to have some level of consistency between the dry run messages.  For add/remove, I followed the pattern of publish of having a `shell.warn` and for it to follow the same pattern of listing out the operation that was not performed.

src/cargo/ops/cargo_clean.rs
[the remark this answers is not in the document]
epage: That makes sense.

I like the idea of listing out directories separately (or even only showing them if there are no files) but I'll leave it to you for whether that is a future improvement or to do it now.  The incremental improvement of that is not worth blocking this.

src/cargo/ops/cargo_clean.rs
ehuss: I like the idea of displaying high-level summaries for the different categories of things that could be deleted. I can maybe include something like that when rebasing this in #12634. However, the particulars seem a little complicated.

I changed the behavior here to not display paths with `--dry-run` without `--verbose`, since the individual files aren't usually particularly interesting. I did keep it with `--verbose` since I think it can potentially be useful, and is also useful for testing to see what it is doing. I was also wondering if maybe it would make sense to have the non-dry-run behave the same (that is, `cargo clean --verbose` should display all paths, not just the ones passed to rm_rf). In concept, behaving the same as Unix `rm -rv`. WDYT?

So, for example, show categories without `--verbose`, and show specific paths with `--verbose`. 

I'm debating with myself of whether or not to show "Removed xyz" after it is removed, or "Removing xyz" before it starts removing. With "Removed", it could display a file/byte summary, which could be interesting to the user ("oh wow, it just deleted 5GiB in git repos!"). With displaying "Removing" before it starts, that can provide useful information in case it fails at some point (though maybe the error message could be updated to explain which category it was trying to delete?).

Some sketches on different ideas:

non-verbose:
```
Removed target directory, 19 files, 19.6KiB
Removed 19 files, 19.7KiB total
```

(Maybe the final summary could be elided if there is only one category of thing that is removed? So the above would just be one line instead of repeating sizes twice.)

Or with `cargo clean -p foo`:
```
Removed target artifacts for package `foo`, 19 files, 19.7KiB
Removed 19 files, 19.7KiB total
```

non-verbose when garbage collecting:
```
Removed stale src caches, 23 files, 1.8MiB
Removed stale .crate caches, 4 files, 192KiB
Removed stale git repos, 192 files, 5.9MiB
Removed stale target artifacts, 36 files, 1.1GiB
Removed 255 files, 1.5GiB total
```

verbose:
```
Removing stale src caches
.... bunch of paths
Removed stale src caches, 23 files, 1.8MiB
Removing target directory
/path/to/target/.rustc_info.json
/path/to/target/CACHEDIR.TAG
/path/to/target/debug/.fingerprint/bar-402c80d9e18085ec/dep-lib-bar
/path/to/target/debug/.fingerprint/bar-402c80d9e18085ec/lib-bar.json
/path/to/target/debug/.fingerprint/bar-402c80d9e18085ec/invoked.timestamp
/path/to/target/debug/.fingerprint/bar-402c80d9e18085ec/lib-bar
/path/to/target/debug/.fingerprint/bar-402c80d9e18085ec
/path/to/target/debug/.fingerprint/foo-810d0583585bc9f9/lib-foo.json
/path/to/target/debug/.fingerprint/foo-810d0583585bc9f9/dep-lib-foo
/path/to/target/debug/.fingerprint/foo-810d0583585bc9f9/lib-foo
/path/to/target/debug/.fingerprint/foo-810d0583585bc9f9/invoked.timestamp
/path/to/target/debug/.fingerprint/foo-810d0583585bc9f9
/path/to/target/debug/.fingerprint
/path/to/target/debug/incremental/foo-hg2akmqki9nq/s-goltvoprn5-rijetd-bu4tjhm078p81fl04874cly2b/query-cache.bin
/path/to/target/debug/incremental/foo-hg2akmqki9nq/s-goltvoprn5-rijetd-bu4tjhm078p81fl04874cly2b/dep-graph.bin
/path/to/target/debug/incremental/foo-hg2akmqki9nq/s-goltvoprn5-rijetd-bu4tjhm078p81fl04874cly2b/work-products.bin
/path/to/target/debug/incremental/foo-hg2akmqki9nq/s-goltvoprn5-rijetd-bu4tjhm078p81fl04874cly2b
/path/to/target/debug/incremental/foo-hg2akmqki9nq/s-goltvoprn5-rijetd.lock
/path/to/target/debug/incremental/foo-hg2akmqki9nq
/path/to/target/debug/incremental
/path/to/target/debug/.cargo-lock
/path/to/target/debug/examples
/path/to/target/debug/deps/foo-810d0583585bc9f9.d
/path/to/target/debug/deps/libfoo-810d0583585bc9f9.rmeta
/path/to/target/debug/deps/bar-402c80d9e18085ec.d
/path/to/target/debug/deps/libbar-402c80d9e18085ec.rmeta
/path/to/target/debug/deps
/path/to/target/debug/build
/path/to/target/debug
/path/to/target
Removed target directory, 19 files, 19.7KiB
Removed 42 files, 1.9MiB total
```

src/cargo/ops/cargo_clean.rs
ehuss: Ok, I like having a warning. I'm still a little reluctant to use the word "aborting" since it isn't really aborting anything, and seems a little confusing to me, since it is still going through the motions. I went with the wording `warning: no files deleted due to --dry-run`. Does that sound ok?

src/cargo/ops/cargo_clean.rs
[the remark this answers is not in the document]
ehuss: I made a change so that it tracks files and directories separately. I decided to not display the number of directories since it feels kinda noisy to me, but it will display the directory count if no files were found. I suspect that will almost never happen, though, since cargo always places files in the `target` directory, and I don't suspect cache directories to ever be missing files.

src/cargo/ops/cargo_clean.rs
ehuss: To follow up on this, @epage and I chatted a little about different ideas here. I would like to add category summaries at the end when it is finished, though those aren't very meaningful without #12634. After reviewing the current behavior, I think I would like to move forward with the behavior in this PR as-is. I'm not entirely sure how the API might change to be able to track the different categories, and it's a little challenging to do that in isolation in this PR. Is there any objection to proceeding with the behavior here? And then perhaps following up on that later?

I'm still a little undecided on the `--verbose` behavior. Currently it is a little inconsistent about what it displays. For example, a bare `cargo clean -v` just prints `Removing /path/to/target` whereas `cargo clean -p ... -v` or `cargo clean --gc` will print `Removing` for individual files. That seems like an independent issue that could be revisited at some point in the future. I don't have much of a preference either way, and it is a relatively easy change to make.

epage (approved):

weihanglo (approved):

PR #12640, opened 2023-09-07
cargo: fix inconsistent dashes in lib.name

dvdhrm:
Change the cargo TOML parser to allow dashes in `lib.name`.

Cargo currently refuses explicit library names if they contain a dash. However, if no library name is specified, the package-name will be used, which itself can contain dashes. This leads to packages with `lib.name` containing dashes, if their package-name contains dashes.

Drop this requirement and allow explicit `lib.name` configurations to contain dashes.

The current ecosystem clearly allows `lib.name` to contain dashes, given that `cargo metadata` already reports such names for any package with a dash. It looks like an oversight that any explicit configuration is not allowed to include dashes.

All consumers of `lib.name` already use `Target.crate_name()` rather than `Target.name()` if a sanitized name is required. Therefore, they already replace dashes with underscores.

Last but not least, the current documentation is simply wrong in stating that `lib.name` defaults to the package-name with dashes replaced by underscores. Cargo never replaces dashes for `lib.name`, and `cargo metadata` and friends already happily show `lib.name` with dashes. Cargo merely replaces dashes with underscores to retrieve the crate-name of a target, which is then used as stem for library artifacts.

### DIFF
--- a/src/cargo/util/toml/targets.rs
+++ b/src/cargo/util/toml/targets.rs
@@ -152,18 +152,10 @@ fn clean_lib(
 ) -> CargoResult<Option<Target>> {
     let inferred = inferred_lib(package_root);
     let lib = match toml_lib {
-        Some(lib) => {
-            if let Some(ref name) = lib.name {
-                // XXX: other code paths dodge this validation
-                if name.contains('-') {
-                    anyhow::bail!("library target names cannot contain hyphens: {}", name)
-                }
-            }
-            Some(TomlTarget {
-                name: lib.name.clone().or_else(|| Some(package_name.to_owned())),
-                ..lib.clone()
-            })
-        }
+        Some(lib) => Some(TomlTarget {
+            name: lib.name.clone().or_else(|| Some(package_name.to_owned())),
+            ..lib.clone()
+        }),
         None => inferred.as_ref().map(|lib| TomlTarget {
             name: Some(package_name.to_string()),
             path: Some(PathValue(lib.clone())),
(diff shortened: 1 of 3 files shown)


dvdhrm:
(update: dropped the test that verifies build-failure for `lib.name` with explicit dashes (but not implicit ones))

hi-rustin:
Related to https://github.com/rust-lang/cargo/issues/2775

epage:
@dvdhrm how did you discover the discrepancy?

We discussed this in the cargo team meeting and didn't really see how this discrepancy would affect people and want to better understand the use cases where people are noticing this so we can better evaluate the compatibility aspect to this (we were leaning towards making the code match the docs).

ps As this is a weightier PR, affecting policy and/or compatibility, we recommend starting something like this off as an Issue and not moving forward until it is "needs mentor" (little help) or "accepted" (mentor available).  If nothing else, (1) we are much more likely to look for discussions like this among the issues and (2) its not uncommon for there to be multiple PRs for one "topic" and focusing on the issue keeps the discussion in one place, making it easier to follow.

dvdhrm:
> @dvdhrm how did you discover the discrepancy?
> 
> We discussed this in the cargo team meeting and didn't really see how this discrepancy would affect people and want to better understand the use cases where people are noticing this so we can better evaluate the compatibility aspect to this (we were leaning towards making the code match the docs).

I am trying to deduce the artifacts a crate produces from its metadata. I was looking at `cargo help metadata` and trying to understand what `packages[].targets[].name` actually contains. And then figure out how to deduce the artifact name from it.

We currently take the value, underscorify it, and then use it as stem for the library artifact. I don't think this would break if you decide to do that in cargo, since the 'underscorify' can be safely applied multiple times.

epage:
If `cargo metadata` is the only way this is exposed and to make it work with anything else, you have to do a conversion that will still work if we pre-apply it, it seems like we could change this behavior and make cargo match the docs.

epage:
Closing in favor of #12783

PR #12643, opened 2023-09-07
Index summary enum

Eh2406:

This is a tiny incremental part of allowing registries to return richer information about summaries that are not available for resolution. Eventually this should lead to better error messages, but for now it's just an internal re-factor. Discussion of the overall approach is at https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/better.20error.20messages.20for.20filtered.20versions.2E

Internal re-factor and tests still pass.

### DIFF
--- a/src/cargo/sources/registry/index.rs
+++ b/src/cargo/sources/registry/index.rs
@@ -430,7 +430,7 @@ impl<'cfg> RegistryIndex<'cfg> {
     ///
     /// Internally there's quite a few layer of caching to amortize this cost
     /// though since this method is called quite a lot on null builds in Cargo.
-    pub fn summaries<'a, 'b>(
+    fn summaries<'a, 'b>(
         &'a mut self,
         name: &str,
         req: &'b OptVersionReq,
--- a/src/cargo/sources/registry/index.rs
+++ b/src/cargo/sources/registry/index.rs
@@ -195,6 +195,8 @@ pub enum IndexSummary {
     Candidate(Summary),
     /// Yanked within its registry
     Yanked(Summary),
+    /// Not available as we are offline and create is not downloaded yet
+    Offline(Summary),
     /// From a newer schema version and is likely incomplete or inaccurate
     Unsupported(Summary, u32),
 }
@@ -205,6 +207,7 @@ impl IndexSummary {
         match self {
             IndexSummary::Candidate(sum)
             | IndexSummary::Yanked(sum)
+            | IndexSummary::Offline(sum)
             | IndexSummary::Unsupported(sum, _) => sum,
         }
     }
@@ -214,6 +217,7 @@ impl IndexSummary {
         match self {
             IndexSummary::Candidate(sum)
             | IndexSummary::Yanked(sum)
+            | IndexSummary::Offline(sum)
             | IndexSummary::Unsupported(sum, _) => sum.package_id(),
         }
     }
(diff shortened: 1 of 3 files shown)


arlosi:
Did you intend to include the commit that refactors to `strip_prefix`?

Eh2406:
Not really. I will rebase after it merges.

Eh2406:
rebased to remove the commit.

tests/testsuite/registry.rs
```diff
@@ -2604,7 +2604,9 @@ fn ignores_unknown_index_version_git() {
 fn ignores_unknown_index_version() {
     // If the version field is not understood, it is ignored.
     Package::new("bar", "1.0.0").publish();
-    Package::new("bar", "1.0.1").schema_version(9999).publish();
+    Package::new("bar", "1.0.1")
```
epage, line 2607: May we never need this change :)

src/cargo/sources/registry/index.rs
```diff
@@ -189,11 +189,34 @@ enum MaybeIndexSummary {
...
-    v: u32,
+#[derive(Clone, Debug)]
+pub enum IndexSummary {
+    /// Available for consideration
+    Candidate(Summary),
+    /// Yanked within its registry
+    Yanked(Summary),
+    /// From a newer schema version and is likely incomplete or inaccurate
+    Unsupported(Summary, u32),
+}
```
epage: Is there a reason you left off the error case from my branch?

src/cargo/sources/registry/index.rs
```diff
@@ -611,7 +634,7 @@ impl<'cfg> RegistryIndex<'cfg> {
 
         let mut count = 0;
         for summary in summaries {
-            f(summary);
+            f(summary.as_summary().clone());
```
epage: Should we add a `into_summary`?

src/cargo/sources/registry/index.rs
```diff
@@ -924,11 +954,20 @@ impl IndexSummary {
...
+            INDEX_V_MAX
+        };
+
+        if v_max < v {
+            return Ok(IndexSummary::Unsupported(summary, v));
+        }
+        if yanked.unwrap_or(false) {
+            return Ok(IndexSummary::Yanked(summary));
+        }
+        Ok(IndexSummary::Candidate(summary))
```
epage, line 998: nit: since these are mutually exclusive branches, I would use an if / else-if setup instead to better communicate that to readers

epage (approved):
Thanks for doing this clean up!

Just nits and questions about work that doesn't have to be in this, so feel free to `r=` me when you feel its ready, whether you apply my nits or not

src/cargo/sources/registry/index.rs
Eh2406: `?` can be used in functions that return `CargoResult<IndexSummary>`. It cannot be used in functions and return `IndexSummary`, even if it has an error variant. (try trait is not stable.) If `IndexSummary` has an error variant then `CargoResult<IndexSummary>` have two different states with the same semantic meaning.

Also it seems like the kind of thing that would be easy to do as a re-factor once all the infrastructure is set up. So it seems like kind decision that could easily be undone.

Eh2406:
we have too many projects half done, the least we can do is address nits before merging!

PR #12648, opened 2023-09-08
feat: Stabilize lints

### LINKED ISSUE #12115: Tracking Issue for `[lints]` table RFC 3389

epage:
Fixes #12115

### DIFF
--- a/src/doc/src/reference/manifest.md
+++ b/src/doc/src/reference/manifest.md
@@ -48,6 +48,7 @@ Every manifest file consists of the following sections:
   * [`[target]`](specifying-dependencies.md#platform-specific-dependencies) --- Platform-specific dependencies.
 * [`[badges]`](#the-badges-section) --- Badges to display on a registry.
 * [`[features]`](features.md) --- Conditional compilation features.
+* [`[lints]`](#the-lints-section) --- Configure linters for this package.
 * [`[patch]`](overriding-dependencies.md#the-patch-section) --- Override dependencies.
 * [`[replace]`](overriding-dependencies.md#the-replace-section) --- Override dependencies (deprecated).
 * [`[profile]`](profiles.md) --- Compiler settings and optimizations.
@@ -528,6 +529,46 @@ both `src/bin/a.rs` and `src/bin/b.rs`:
 ```toml
 [package]
 default-run = "a"
+```
+
+#### The `lints` section
+
+Override the default level of lints from different tools by assigning them to a new level in a
+table, for example:
+```toml
+[lints.rust]
+unsafe_code = "forbid"
+```
+
+This is short-hand for:
+```toml
+[lints.rust]
+unsafe_code = { level = "forbid", priority = 0 }
+```
+
+`level` corresponds to the lint levels in `rustc`:
+- `forbid`
+- `deny`
+- `warn`
+- `allow`
+
+`priority` is a signed integer that controls which lints or lint groups override other lint groups:
+- lower (particularly negative) numbers have lower priority, being overridden
+  by higher numbers, and show up first on the command-line to tools like
+  `rustc`
+
+To know which table under `[lints]` a particular lint belongs under, it is the part before `::` in the lint
+name.  If there isn't a `::`, then the tool is `rust`.  For example a warning
+about `unsafe_code` would be `lints.rust.unsafe_code` but a lint about
+`clippy::enum_glob_use` would be `lints.clippy.enum_glob_use`.
+
+For example:
+```toml
+[lints.rust]
+unsafe_code = "forbid"
+
+[lints.clippy]
+enum_glob_use = "deny"
 ```
 
 ## The `[badges]` section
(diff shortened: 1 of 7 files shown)


src/doc/src/reference/manifest.md
```diff
@@ -530,6 +531,46 @@ both `src/bin/a.rs` and `src/bin/b.rs`:
 default-run = "a"
 ```
 
+#### The `lints` section
+
+Override the default level of lints from different tools by assigning them to a new level in a
+table, for example:
+```toml
+[lints.rust]
+unsafe = "forbid"
```
taiki-e, line 540: I think `unsafe` is not a valid lint name, did you mean `unsafe_code`? Or is there an alias for a particular lint?

weihanglo (approved):
🚀 🚀 🚀

src/doc/src/reference/manifest.md
```diff
@@ -46,6 +46,7 @@ Every manifest file consists of the following sections:
   * [`[dev-dependencies]`](specifying-dependencies.md#development-dependencies) --- Dependencies for examples, tests, and benchmarks.
   * [`[build-dependencies]`](specifying-dependencies.md#build-dependencies) --- Dependencies for build scripts.
   * [`[target]`](specifying-dependencies.md#platform-specific-dependencies) --- Platform-specific dependencies.
+* [`[lints]`](#the-lints-section) --- Configure linters for this package.
```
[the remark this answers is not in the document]
epage: Did a slight tweak to where it is at.

My thought process is:
- User-facing entries are more important and should go first (like `features`)
- Of the internal stuff (`patch`, etc), this seemed like the more likely one to be used.

src/doc/src/reference/workspaces.md
```diff
@@ -222,6 +223,33 @@ cc.workspace = true
...
+
+Example:
+
+```toml
+# [PROJECT_DIR]/Cargo.toml
+[workspace]
+members = ["crates/*"]
+
+[workspace.lints.rust]
+unsafe_code = "forbid"
```
weihanglo, line 240: It seems that both `unsafe_code` and `unsafe-code` works. It may not be worth a mention in doc but the error message always shows dash version, which is a bit unfortunate. 

```
  = note: requested on the command line with `-D unsafe-code`
```

weihanglo (approved):

weihanglo:
Let's merge it. Thank you @epage and everyone working and testing on this!

@bors r+

PR #12649, opened 2023-09-08
feat: stabilize credential-process and registry-auth

### LINKED ISSUE #8933: Tracking Issue for credential-process RFC 2730

### LINKED ISSUE #10474: Tracking Issue for alternative registry authentication (RFC 3139)

arlosi:
Stabilization PR for `registry-auth` and `credential-process`.

Tracking approval of this stabilization is done in the via the FCP in [#8933](https://github.com/rust-lang/cargo/issues/8933#issuecomment-1711990123). This PR is here to help reviewers of the FCP.

* Stabilizes `registry-auth` and `credential-process`
* Makes authenticated registries require a credential provider
* Adds stable documentation for credential providers and authenticated registries

Closes #8933
Closes #10474

### DIFF
--- a/src/doc/src/reference/registry-authentication.md
+++ b/src/doc/src/reference/registry-authentication.md
@@ -1,0 +1,111 @@
+# Registry Authentication
+Cargo authenticates to registries with through credential providers. These
+credential providers are external executables or built-in providers that Cargo
+uses to store and retreive credentials.
+
+Using alternative registries with authentication *requires* a credential provider to be configured
+to avoid unknowningly storing unecrypted credentials on disk. For historical reasons, public
+(non-authenticated) registres do not require credential provider configuration and the `cargo:token`
+provider is used if no providers are configured.
+
+Cargo also includes platform-specific providers that use the operating system to securely store
+tokens. The `cargo:token` provider is also included which stores credentials in unencrypted plain
+text in the [credentials](config.md#credentials) file. 
+
+## Recommended configuration
+It's recommended to configure a global credential provider list in `$CARGO_HOME/config.toml`
+which defaults to:
+* Windows: `%USERPROFILE%\.cargo\config.toml`
+* Unix: `~/.cargo/config.toml`
+
+This recommended configuration uses the operating system provider, with a fallback to `cargo:token`
+to look in Cargo's [credentials](config.md#credentials) file or environment variables.
+
+Some private registries may also recommend a registry-specific credential-provider. Check your
+registry's documentation to see if this is the case.
+
+### macOS configuration
+```toml
+# ~/.cargo/config.toml
+[registry]
+global-credential-providers = ["cargo:token", "cargo:macos-keychain"]
+```
+
+### Linux (libsecret) configuration
+```toml
+# ~/.cargo/config.toml
+[registry]
+global-credential-providers = ["cargo:token", "cargo:libsecret"]
+```
+
+### Windows configuration
+```toml
+# %USERPROFILE%\.cargo\config.toml
+[registry]
+global-credential-providers = ["cargo:token", "cargo:wincred"]
+```
+
+See [`registry.global-credential-providers`](config.md#registryglobal-credential-providers)
+for more details.
+
+## Built-in providers
+Cargo includes several built-in credential providers. The available built-in providers
+may change in future Cargo releases (though there are currently no plans to do so).
+
+### `cargo:token`
+Uses Cargo's [credentials](config.md#credentials) file to store tokens unencrypted in plain text.
+When retreiving tokens, checks the `CARGO_REGISTRIES_<NAME>_TOKEN` environment variable.
+If this credential provider is not listed, then the `*_TOKEN` environment variables will not work.
+
+### `cargo:wincred`
+Uses the Windows Credential Manager to store tokens.
+
+The credentials are stored as `cargo-registry:<index-url>` in the Credential Manager
+under "Windows Credentials".
+
+### `cargo:macos-keychain`
+Uses the macOS Keychain to store tokens.
+
+The Keychain Access app can be used to view stored tokens.
+
+### `cargo:libsecret`
+Uses [libsecret](https://wiki.gnome.org/Projects/Libsecret) to store tokens.
+
+On GNOME, credentials can be viewed using [GNOME Keyring](https://wiki.gnome.org/Projects/GnomeKeyring)
+applications.
+
+### `cargo:token-from-stdout <command> <args>`
+Launch a subprocess that returns a token on stdout. Newlines will be trimmed.
+* The process inherits the user's stdin and stderr.
+* It should exit 0 on success, and nonzero on error.
+* [`cargo login`] and [`cargo logout`] are not supported and return an error if used.
+
+The following environment variables will be provided to the executed command:
+
+* `CARGO` --- Path to the `cargo` binary executing the command.
+* `CARGO_REGISTRY_INDEX_URL` --- The URL of the registry index.
+* `CARGO_REGISTRY_NAME_OPT` --- Optional name of the registry. Should not be used as a lookup key.
+
+Arguments will be passed on to the subcommand.
+
+[`cargo login`]: ../commands/cargo-login.md
+[`cargo logout`]: ../commands/cargo-logout.md
+
+## Credential plugins
+For credential provider plugins that follow Cargo's [credential provider protocol](credential-provider-protocol.md),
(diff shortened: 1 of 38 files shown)


src/doc/src/reference/registry-authentication.md
```diff
@@ -0,0 +1,103 @@
...
+* `CARGO` --- Path to the `cargo` binary executing the command.
+* `CARGO_REGISTRY_INDEX_URL` --- The URL of the registry index.
+* `CARGO_REGISTRY_NAME_OPT` --- Optional name of the registry. Should not be used as a lookup key.
+
+[`cargo login`]: ../commands/cargo-login.md
+[`cargo logout`]: ../commands/cargo-logout.md
+
+## External providers
+For external providers that follow Cargo's [credential provider protocol](credential-provider-protocol.md)
+use the path to the executable (or the executable name if on the `PATH`).
```
ehuss, line 92: This sentence was a little confusing on my first read. Does something like the following seem a little clearer?

```suggestion
For external providers that follow Cargo's [credential provider protocol](credential-provider-protocol.md),
the configuration value should be a string with the path to the executable (or the executable name if on the `PATH`).
```

ehuss:
Thanks so much for writing all the documentation, it looks great!

Just some minor notes about the documentation:

Can you update https://github.com/rust-lang/cargo/blob/HEAD/src/doc/src/reference/environment-variables.md to include `CARGO_CREDENTIAL_ALIAS_<name>` and `CARGO_REGISTRIES_<name>_CREDENTIAL_PROVIDER` and `CARGO_REGISTRY_GLOBAL_CREDENTIAL_PROVIDERS`?

----

One thing I'd like to chat about is documenting the stability of the built-in credential providers. I'm concerned that we may not be able to support them indefinitely, at which point we would need to remove one which would break cargo's general compatibility guarantees. For example, let's say Apple or Microsoft suddenly start requiring signed binaries in order to access the global credential stores, and we can't accommodate that. Or, let's say libsecret stops development and distros stop using it. I'm wondering it there should be a note in the auth or config chapter about that? I don't want a note that sounds scary, since it is unlikely that we would stop supporting them, and we would also try to provide a warning ahead of time if possible if we plan to remove it. But I do think it is useful to clearly communicate to users what our expectations around compatibility are.

An alternate idea would be to have a separate "compatibility contract" that would generally state how cargo treats backwards and forwards compatibility, and then exceptions like this (and other things like RUSTFLAGS or platform support) could be mentioned there. Right now, we only have https://doc.crates.io/contrib/design.html which spells that out, but that is not end-user documentation. Perhaps that could move to the cargo guide?

Any thoughts on that?

src/doc/src/reference/registry-authentication.md
arlosi: Yes, this is better. Thanks!

arlosi:
Good catch on the missing docs in `environment-variables.md`. Updated.

> I'm concerned that we may not be able to support them indefinitely, at which point we would need to remove one which would break cargo's general compatibility guarantees.

I've added a note in `registry-authentication.md` that says:
> Cargo includes several built-in credential providers. The available built-in providers may change in future Cargo releases.

Do you think that's sufficient?

ehuss:
> Do you think that's sufficient?

Seems good to me.

Eh2406:
> Cargo includes several built-in credential providers. The available built-in providers may change in future Cargo releases. Although, we do not have plans to do so in the near future. If plans change we will update this documentation.

Or sum such.

ehuss:
I forgot, can you also update [`publish.py`](https://github.com/rust-lang/cargo/blob/master/publish.py) so that it includes all of the credential crates? 1password doesn't *need* to be included, but I don't think it will cause any harm to include it.

weihanglo (approved):

weihanglo:
Your efforts on this are much appreciated, Arlo! Also thanks everyone for reviewing and testing on this feature :)

We're not going to wait for the FCP ended, as the beta branching-off window is approaching, and the FCP [was started 9 days ago](https://github.com/rust-lang/cargo/issues/8933#issuecomment-1711990145). If you encounter any issue with this stabilization, please report it here or open a new issue.

@bors r+

Issue #12432, opened 2023-08-01
Official API for build scripts


epage:
### Problem

Having a Rust API would help with several problems
- Help people migrate through build script best practices (e.g. #11461)
- Avoid people making type mistakes
- Bring documentation more inline with people's workflow, e.g. IDEs auto-completing and rendering doc comments
- Help improve discoverability of APIs
- Ensures a feedback loop with the cargo team, hoping it avoids some failure modes where compatibility is broken
- Reduce duplicated effort
  - [build_script](https://crates.io/crates/build_script)
  - [build-rs](https://crates.io/crates/build-rs)
  - [build-env](https://crates.io/crates/build-env)
  - [vergen](https://docs.rs/vergen/latest/vergen/)
  - [foreman <=0.4.0](https://docs.rs/foreman/0.4.0/foreman/) (package name was handed off and used for a different purpose after that version)
  - [system-deps](https://crates.io/crates/system-deps)
  - [rsconf](https://crates.io/crates/rsconf)

### Proposed Solution

A crate that provides an API for working with build scripts
- Low level API handling writing of directives, reading runtime env variables, etc (see also #12431)
- A higher level API to help with re-run-if-changed

### Notes

Things to work out
- Name
- If we co-maintain, how repo and publish permissions work out
- ~~How do we handle the MSRV because the cargo repo currently only handles "latest" (see also #12381)~~ see #12654

As for timing, it'd help if we had this to recommend before we fully resolve #11461 so we can tell people to migrate once, rather than twice

[FCP](https://github.com/rust-lang/cargo/issues/12432#issuecomment-2389527055)

epage:
FYI we now have an MSRV policy as of #12654

jmesmon:
I'm the person who wrote `build-env` (and also added the target specific env var naming pattern it uses to the `cc` crate originally):

The content of `build-env` was primarily written to allow build scripts to implement target specific env variable reading in a way that matches the `cc` crate's methodology. In practice though, this has been either reimplemented in other build scripts either in the same way (rarely) or in a manner closer but not the same as the style used by cargo's own target specific env variables.

A `build-rs`-like API for emitting the directives read by cargo or parsing variables set by cargo seems fine, and I expect this crate could use it easily (if I ever update it). I generally like the idea of having a real API instead of peeking into env variables and printing magic strings.

Some encouragement on increasing the scope though:

[...]

PR #12654, opened 2023-09-11
Set and verify all MSRVs in CI

epage:

Allow us to advertise an MSRV for all public packages, unblocking #12432.

Packages are broken down into two MSRV policies
- Internal packages: `rust-version=stable`
- Public packages: `rust-version=stable-2`

To support this
- RenovateBot will automatically update all MSRVs
- CI will verify all packages build according to their MSRV

Since this takes the "single workspace" approach, the downside is that common dependencies are subject to each package's MSRV.  We also can't rely on `-Zmsrv-policy` to help us generate a lockfile because it breaks down when trying to support multiple MSRVs in a workspace

Per commit

### Additional information

#12381 skipped setting some MSRVs because we weren't sure how to handle it.  For `cargo-credential`, we needed to do something and did one-off verification (#12623).  `cargo-hack` recently gained the ability to automatically select MSRVs for each package allowing us to scale this up to the entire workspace.

I don't know if we consciously chose an MSRV policy for `cargo-credential` but it looked like N-2 so that is what I stuck with and propagated out.
- Without an aggressive sliding MSRV, we discourage people from using newer features because they will feel like they need permission which means it needs to be justified
- Without an aggressive sliding MSRV, if the MSRV at one point in time works for someone, they tend to assume it will always work, leading to frustration at unmet expectations.

I switched the MSRV check to `cargo check` from `cargo test` because I didn't want to deal with the out-of-diskspace issues and `check` will catch 99% of MSRV issues.

Potential future improvements to `cargo-hack`
- Allow `--version-range ..stable` so we can verify all MSRVs that aren't stable which would bypass the diskspace issues and allow us to more easily use `cargo test` again
- Verify on a `cargo package` version of a crate (taiki-e/cargo-hack#216)

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -22,6 +22,19 @@
       packageNameTemplate: 'rust-lang/rust',
       datasourceTemplate: 'github-releases',
     },
+    {
+      customType: 'regex',
+      fileMatch: [
+        'Cargo.toml$',
+      ],
+      matchStrings: [
+        '\bMSRV:3\b.*?(?<currentValue>\\d+\\.\\d+(\\.\\d+)?)',
+        '(?<currentValue>\\d+\\.\\d+(\\.\\d+)?).*?\bMSRV:3\b',
+      ],
+      depNameTemplate: 'MSRV:3',  // Support 3 versions of rustc
+      packageNameTemplate: 'rust-lang/rust',
+      datasourceTemplate: 'github-releases',
+    },
   ],
   packageRules: [
     {
@@ -32,10 +45,26 @@
       matchPackageNames: [
         'MSRV:1',
       ],
+      schedule: [
+        '* * * * *',
+      ],
+      groupName: 'msrv',
+    },
+    {
+      commitMessageTopic: 'MSRV (3 versions)',
+      matchManagers: [
+        'regex',
+      ],
+      matchPackageNames: [
+        'MSRV:3',
+      ],
       "extractVersion": "^(?<version>\\d+\\.\\d+)",  // Drop the patch version
       schedule: [
         '* * * * *',
       ],
+      minimumReleaseAge: '85 days',  // 2 releases back * 6 weeks per release * 7 days per week + 1
+      internalChecksFilter: 'strict',
+      groupName: 'msrv',
     },
     // Goals:
     // - Rollup safe upgrades to reduce CI runner load
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -12,24 +12,25 @@
     {
       customType: 'regex',
       fileMatch: [
-        '^Cargo.toml$',
+        'Cargo.toml$',
       ],
       matchStrings: [
-        'rust-version.*?(?<currentValue>\\d+\\.\\d+(\\.\\d+)?)',
+        '\bMSRV:1\b.*?(?<currentValue>\\d+\\.\\d+(\\.\\d+)?)',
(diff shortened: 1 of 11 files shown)


credential/cargo-credential-1password/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "cargo-credential-1password"
 version = "0.4.0"
 edition.workspace = true
 license.workspace = true
+rust-version = "1.70.0"  # MSRV:3
```
epage, line 6: From https://github.com/rust-lang/cargo/pull/12381#issuecomment-1698283115

> Just realized, with this pull request and #12395, it is inevitable that every time a Rust release comes out, `cargo-util` and `crates-io` will be forced to release a version due to the bump of `package.rust-version`, even when there is no actual code change involved.
> 
> Is that something we've considered? It feels like a bit unnecessary.

What we could do is swap `workspace.rust-version` so it is `MSRV:3` and so these packages will be "untouched" despite getting MSRV bumps

.github/renovate.json5
```diff
@@ -29,12 +43,26 @@
...
+        '* * * * *',
+      ],
+    },
+    {
+      commitMessageTopic: 'MSRV',
+      matchManagers: [
+        'regex',
+      ],
+      matchPackageNames: [
+        'msrv:3',
```
weihanglo, line 58: What does this naming stand for?

weihanglo (commented):
Looks pretty good. Sorry for the long wait.

crates/cargo-platform/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "cargo-platform"
 version = "0.1.5"
 edition.workspace = true
 license.workspace = true
+rust-version = "1.70.0"  # MSRV:3
```
weihanglo, line 6: This comment seems like a magic and is prone to remove? Do we have a way to avoid a remove even when maintainers aren't familiar with the msrv CI job?

epage: The only thing I could think of is to add additional text after saying this is a renovatebot marker.

Thoughts?

credential/cargo-credential-1password/Cargo.toml
epage: @weihanglo any thoughts on whether `workspace.package.rust-version` should track the MSRV for internal packages (MSRV:1) or user-facing packages (MSRV:3)?

epage:
We talked about this in the last cargo team meeting.

We were in favor of a sliding, auto-updating MSRV window because
- It gives contributors permission to use new features, rather than requiring them to have the gumption to ask if they can bump MSRV, feeling like their case needs to be justified
- From past experience, I've at least seen people misinterpret a MSRV policy based what people assumed it was because of a lack of MSRV bumps which caused frustration when we didn't meet those expectations
- It is unclear the benefit gained to people by us extending past the minimum of the policy as MSRV bumps would be unpredictable and they wouldn't be able to reliably get value out of it.

We went with N-2 (3 versions total) policy as some projects have been doing that and we didn't really have a reason to do otherwise.  Something to keep in mind is that we have two MSRVs here (cargo internals, user facing packages) with one lockfile, so the lower the MSRV, the more constrained we are on what dependency versions we can pull in for our cargo internals.

weihanglo (approved):

weihanglo:
Thanks. I'll go ahead and merge this 👍🏾 

@bors r+

PR #12655, opened 2023-09-11
refactor: Consolidate clap/shell styles

epage:

This is a follow up to #12578 to reduce duplication of terminal styling.

This still leaves styling in `color_print::cstr!`.

This is also prep for copy/pasting into `clap-cargo` for others to use.  Another step might be to extract `cargo::util::style` into its own crate.

We have no styling tests so this can only be verified by inspection or running the commands

### Additional information

I chose `anstyle` for expressing these as its an API designed specifically for stable style definitions to put in public APIs.

### DIFF
--- a/src/cargo/ops/registry/search.rs
+++ b/src/cargo/ops/registry/search.rs
@@ -6,10 +6,9 @@ use std::cmp;
 use std::iter::repeat;
 
 use anyhow::Context as _;
-use termcolor::Color;
-use termcolor::ColorSpec;
 use url::Url;
 
+use crate::util::style;
 use crate::util::truncate_with_ellipsis;
 use crate::CargoResult;
 use crate::Config;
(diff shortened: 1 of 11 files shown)


src/cargo/ops/registry/search.rs
```diff
@@ -58,15 +57,12 @@ pub fn search(
...
         let mut fragments = line.split(query).peekable();
         while let Some(fragment) = fragments.next() {
-            let _ = config.shell().write_stdout(fragment, &ColorSpec::new());
+            let _ = config.shell().write_stdout(fragment, &style::NOP);
             if fragments.peek().is_some() {
-                let _ = config.shell().write_stdout(
-                    query,
-                    &ColorSpec::new().set_bold(true).set_fg(Some(Color::Green)),
-                );
+                let _ = config.shell().write_stdout(query, &style::GOOD);
```
weihanglo, line 62: Bold failed to apply here.

weihanglo (commented):
Some texts are not applied correctly with bold style. Filed an upstream fix for that: <https://github.com/rust-cli/anstyle/pull/124>.

Apart from that, it looks good :)

epage:
Updated with the new release of anstyle-termcolor

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #12657, opened 2023-09-11
libgit2 fixed upstream

Eh2406:

This "FIXME" points to https://github.com/libgit2/libgit2/issues/2514 witch has bean fixed for... 8 years. So maybe it is no longer needed.

By adding panic messages I was able to demonstrate that there were tests hitting this code path. Then I removed the "FIXME" and had the test still pass.


epage:
*shrug* tests pass

@bors r+

PR #12660, opened 2023-09-12
feat(cli): Add '-n' to dry-run

epage:
This came from #12638 and my many small frustrations from wanting to use `-n` and not being able to.

We do not have any existing `-n` flags for this to be confused with.

I would wager that `-n` is such an entrenched short flag in build tools that it would not make sense for us to use it with any other flag.

For a survey of where `-n` is used as a short, see https://www.gnu.org/prep/standards/html_node/Option-Table.html#Option-Table

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -291,7 +291,7 @@ pub trait CommandExt: Sized {
     }
 
     fn arg_dry_run(self, dry_run: &'static str) -> Self {
-        self._arg(flag("dry-run", dry_run))
+        self._arg(flag("dry-run", dry_run).short('n'))
     }
 
     fn arg_ignore_rust_version(self) -> Self {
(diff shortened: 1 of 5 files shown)


weihanglo:
Let's do a poll for this short alias

@rfcbot fcp merge

I might be biased but here is a quick survey of some prior arts

### Projects using `-n` as a `--dry-run` alias

* `rsync`
* GNU `make`
* `git add/rm/clean…`
* Dart `dart pub publish`
* Ruby `rake`
* Meson build system
* SQLite archive
* `mysqlhotcopy`

### Projects using other aliases

* `conda` — `-d`
* `gem cleanup` — both `-d` and `-n`
* `kubectl` – no alias
* `npm` — no alias
* `pip` — no alias
* `bun` — no alias

weihanglo:
Going to merge this promptly as it doesn't look too controversial within the Cargo team. Please call out if you have objection to the decision.

@bors r+

PR #12664, opened 2023-09-13
chore: set `publish=false` for built-in credential providers

weihanglo:

I could be wrong, but these packages no longer need to be published, right?

r? arlosi


ehuss:
These all still need to be published. Otherwise cargo-the-library would not be able to download them.

weihanglo:
Oh you're absolutely correct. I don't know what I was thinking…

PR #12666, opened 2023-09-13
Fix typos: `informations` -> `information`

hi-rustin:
We do not use information in the plural form.


epage (approved):
Feel free to suggest this for `typos`

PR #12677, opened 2023-09-15
refactor: use `RegistryOrIndex` enum to replace two booleans

weihanglo:

`fn registry()` has two booleans representing `--index` and `--registry` respectively. However, they are inherently mutually exclusive and better to use a single sum type `RegistryOrIndex` to model them.

This PR is expected not to bring any behavior change. The only change is that the cli arg conflict detection is now moved to command line parsing phase (clap).

A new test `publish_failed_with_index_and_only_allowed_registry` demonstrates a quirky message was emitted when 

* `package.publish = […]` contains the only one registry that trigger an implicit publish
* `--index` arg was provided and took precedence over the implicit registry publish

This behavior should be fixed IMO but let's leave it to another PR.

### DIFF
--- a/src/bin/cargo/commands/init.rs
+++ b/src/bin/cargo/commands/init.rs
@@ -7,7 +7,7 @@ pub fn cli() -> Command {
         .about("Create a new cargo package in an existing directory")
         .arg(Arg::new("path").action(ArgAction::Set).default_value("."))
         .arg_new_opts()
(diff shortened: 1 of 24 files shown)


hi-rustin (approved):
Looks good. Thanks!

ehuss:
Thanks!

@bors r=hi-rustin

PR #12678, opened 2023-09-15
Shortest path

Eh2406:

Currently error messages from the resolver are based a random path from a package that couldn't be resolved back to the root package. It was pointed out to me that the shortest route is likely to be more useful so the switches to using breadth first search to pick the path. There is also one re-factor to use let-else that I spotted while doing this work.

The shortest path is is a random path, so this is not technically a behavioral change. As such the tests still pass is good evidence for being reasonable.

### DIFF
--- a/crates/resolver-tests/src/lib.rs
+++ b/crates/resolver-tests/src/lib.rs
@@ -164,7 +164,7 @@ pub fn resolve_with_config_raw(
                 // we found a case that causes a panic and did not use all of the input.
                 // lets print the part of the input that was used for minimization.
                 eprintln!(
-                    "{:?}",
+                    "Part used befor drop: {:?}",
                     PrettyPrintRegistry(
                         self.list
                             .iter()
--- a/src/cargo/util/graph.rs
+++ b/src/cargo/util/graph.rs
@@ -106,6 +106,18 @@ impl<N: Eq + Ord + Clone, E: Default + Clone> Graph<N, E> {
             result.push(p);
             pkg = p.0;
         }
+        #[cfg(debug_assertions)]
+        {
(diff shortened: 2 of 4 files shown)


src/cargo/util/graph.rs
```diff
@@ -87,53 +87,74 @@ impl<N: Eq + Ord + Clone, E: Default + Clone> Graph<N, E> {
...
+        }
+
+        let mut result = Vec::new();
+        let mut next = bottom.unwrap();
+        while let Some((p, e)) = back_link.remove(&next) {
+            result.push((next, Some(e)));
+            next = p;
         }
+        result.push((next, None));
+        result.reverse();
```
epage, line 157: Are you able to make a test that shows the old behavior so it can be contrasted with the new behavior?

Eh2406: Good news, this code is covered by prop tests so it's easy to create examples. (Their ugly and require a lot of work to clean up, but easy to create in the first place.)

Bad news, the first one I worked on the new message appears to be completely incorrect.

Eh2406:
code fixed. History rewritten. And test added.

crates/resolver-tests/src/lib.rs
```diff
@@ -164,7 +164,7 @@ pub fn resolve_with_config_raw(
                 // we found a case that causes a panic and did not use all of the input.
                 // lets print the part of the input that was used for minimization.
                 eprintln!(
-                    "{:?}",
+                    "Part used befor drop: {:?}",
```
epage, line 167: nit: `before` rather than `befor`

epage (approved):
Feel free to r= when its ready

PR #12679, opened 2023-09-15
doc: clarify caret requirements

### LINKED ISSUE #12112: subtle documentation ambiguity
### Problem

The Cargo Book says:

> Caret requirements are an alternative syntax for the default strategy, `^1.2.3` is exactly equivalent to `1.2.3`.

This phrasing has an unfortunate interpretation for Rust newbies, who may take the statement to mean that caret means the exact version identifier.

A very careful reading reveals that caret is merely the default semantic, but is *not* exactly equivalent to any fixed version. Some users may get tripped up by this ambiguity.

### Possible Solution(s)

We can help the reader to disambiguate Cargo syntax and semantics. For example, adjust the section to say something like:

> The caret prefix is also the default version resolution strategy; it may be omitted. Both `^1.2.3` and `1.2.3` match the latest available, *Cargo* SemVer compatible release that Cargo can find.

tompscanlan:

clearing up caret version requirements

possibly fixes #12112

expert review...

### DIFF
--- a/src/doc/src/reference/specifying-dependencies.md
+++ b/src/doc/src/reference/specifying-dependencies.md
@@ -57,8 +57,17 @@ using special operators, though it shouldn't be necessary most of the time.
 
 ### Caret requirements
 
-**Caret requirements** are an alternative syntax for the default strategy,
-`^1.2.3` is exactly equivalent to `1.2.3`.
+**Caret requirements** are the default version requirement strategy. 
+This version strategy allows [SemVer] compatible updates.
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
This looks way better than I expected! Is this ready to merge?

tompscanlan:
Thanks. I was expecting some discussion. Go for it if it looks good.

weihanglo:
Thanks! You've done a great job so I don't have to change any word :)

@bors r+

PR #12681, opened 2023-09-16
feat(embedded): Hack in code fence support 

epage:

This is to allow us to get feedback on the design proposed
[on zulip](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/Embedding.20cargo.20manifests.20in.20rust.20source/near/391427092)
to verify we want to make an RFC for this syntax.

````rust
#!/usr/bin/env cargo
```cargo
[dependencies]
clap = { version = "4.2", features = ["derive"] }
```

use clap::Parser;

#[derive(Parser, Debug)]
#[clap(version)]
struct Args {
    #[clap(short, long, help = "Path to config")]
    config: Option<std::path::PathBuf>,
}

fn main() {
    let args = Args::parse();
    println!("{:?}", args);
}
````

The tests were updated in a separate commit to ensure there was no regression while then migrating to the new syntax to make sure it worked.

This involves some future work
- Removing doc comment support
- Getting the syntax approved and implemented
- Migrating to rustc support for the syntax

#12207 was updated to record these items so we don't lose track of them

### DIFF
--- a/src/cargo/util/toml/embedded.rs
+++ b/src/cargo/util/toml/embedded.rs
@@ -15,19 +15,72 @@ pub fn expand_manifest(
     path: &std::path::Path,
     config: &Config,
 ) -> CargoResult<String> {
-    let comment = extract_comment(content)?.unwrap_or_default();
-    let manifest = match extract_manifest(&comment)? {
-        Some(manifest) => Some(manifest),
-        None => {
-            tracing::trace!("failed to extract manifest");
-            None
+    let source = split_source(content)?;
+    if let Some(frontmatter) = source.frontmatter {
+        match source.info {
+            Some("cargo") => {}
+            None => {
+                anyhow::bail!("frontmatter is missing an infostring; specify `cargo` for embedding a manifest");
+            }
+            Some(other) => {
+                if let Some(remainder) = other.strip_prefix("cargo,") {
+                    anyhow::bail!("cargo does not support frontmatter infostring attributes like `{remainder}` at this time")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/cargo/util/toml/embedded.rs [hunk 1]
```diff
@@ -159,8 +205,74 @@ fn sanitize_name(name: &str) -> String {
...
+
+fn split_source(input: &str) -> CargoResult<Source<'_>> {
+    let mut source = Source {
+        shebang: None,
+        info: None,
+        frontmatter: None,
+        content: input,
+    };
+
+    if source.content.starts_with("#![") {
```
epage, line 223: I need to update this to match https://github.com/rust-lang/rust/blob/7d9bce327332fef22cf687b44de58f06d30daeb0/compiler/rustc_lexer/src/lib.rs#L229

src/cargo/util/toml/embedded.rs
```diff
@@ -15,26 +15,72 @@ pub fn expand_manifest(
     path: &std::path::Path,
     config: &Config,
 ) -> CargoResult<String> {
-    let comment = match extract_comment(content) {
-        Ok(comment) => Some(comment),
-        Err(err) => {
-            tracing::trace!("failed to extract doc comment: {err}");
-            None
+    let source = split_source(content)?;
```
epage, line 18: Docs need updating

Muscraft (approved):
Looks good to me!

PR #12682, opened 2023-09-18
util/toml: Remove duplicate `serde(rename)` attributes

Turbo87:
The struct is already using `#[serde(rename_all = "kebab-case")]`, so the rename attributes on the individual fields seems unnecessary.


weihanglo (approved):

weihanglo:
Nice catch. Thanks!

@bors r+

PR #12687, opened 2023-09-18
SemVer: Update documentation about removing optional dependencies

ehuss:
This updates the documentation cautioning against removing optional dependencies with more up-to-date information about using the `dep:` syntax in the features table. This documentation was written before `dep:`, and I just forgot to update these docs when it was stabilized.

### DIFF
--- a/src/doc/src/reference/semver.md
+++ b/src/doc/src/reference/semver.md
@@ -2135,8 +2135,15 @@ std = []
 
 #### Possibly-breaking: removing an optional dependency {#cargo-remove-opt-dep}
 
-Removing an optional dependency can break a project using your library because
+Removing an [optional dependency][opt-dep] can break a project using your library because
 another project may be enabling that dependency via [Cargo features].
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/semver.md
```diff
@@ -2152,7 +2159,33 @@ curl = { version = "0.4.31", optional = true }
...
+#
+# This example shows how to avoid breaking changes with optional dependencies.
+
+###########################################################
+# Before
+[dependencies]
+curl = { version = "0.4.31", optional = true }
+
+[features]
+networking = ["dep:curl"]
```
epage, line 2173: I think its fine for us to suggest this be treated as minor.  The main risk is rust-lang/rust#109005 as people might be seeing the internal names and using them.  I actually saw someone [mislead by this](https://www.youtube.com/watch?v=jVoEA7qmN8c&t=8s) when demoing the compiler error.

PR #5179, opened 2018-03-14
Copy `.pdb` files to `target` directory

matklad:
`.pdb` files are for windows debug info (unlike on linux, debug info is
in a separate file). Windows executable actually hard-code paths to
`.pdb` files, so debugging mvsc rust programs works even without this
patch. However, if you want to distribute the executable to other
machines, you'd better distribute both `foo.exe` and `foo.pdb`, because
absolute paths won't work on another machine. Having same-named .pdb
file alongside the binary would work though.

closes #4960

PR #12688, opened 2023-09-18
fix: copy PDBs for EFI targets

lf-:

EFI also uses the PE format with .pdb files, and rustc generates them, but Cargo does not copy them out of target/*/deps as it does on Windows. This is an oversight, so this PR fixes it.

See the test below; you can copy it into a new project and run `cargo build --target x86_64-unknown-uefi` and verify that the pdb is properly copied.

### Additional information

Related: https://github.com/rust-osdev/uefi-rs/issues/289
Related (originally adding the pdb stuff for windows): https://github.com/rust-lang/cargo/pull/5179

### DIFF
--- a/src/cargo/core/compiler/build_context/target_info.rs
+++ b/src/cargo/core/compiler/build_context/target_info.rs
@@ -465,7 +465,7 @@ impl TargetInfo {
                     // the names to match.
                     should_replace_hyphens: false,
                 })
-            } else if target_triple.ends_with("-msvc") {
+            } else if target_triple.ends_with("-msvc") || target_triple.ends_with("-uefi") {
                 ret.push(FileType {
                     suffix: ".pdb".to_string(),
                     prefix: prefix.clone(),


hi-rustin:
@lf- The CI failed because there is no `x86_64-unknown-uefi` target in the CI.

lf-:
> @lf- The CI failed because there is no `x86_64-unknown-uefi` target in the CI.

oh, that's very unsurprising. I was somewhat worried that would mess us up. How should we actually test this? -Zbuild-std would work, but I don't know if that's allowable in the cargo testsuite?

ehuss:
We generally don't have tests for non-tier-1 targets. I wouldn't worry about adding a test for this case.  Ping @dvdhrm and @nicholasbishop as uefi target maintainers, in case you want to verify this change.

lf-:
Cool, there is an easy solution to this, test has been removed :P

dvdhrm (approved):
Looks good! Thanks a lot!

Alternatively, we could check the `is_like_windows` key of the target, but not sure that would make it better.

arlosi:
Thanks!

@bors r+

PR #12691, opened 2023-09-18
Fix docs in profiles.md 

sansyrox:
I think we should be using `abort` instead of `unwind` here
Explain the motivation behind this change.
A clear overview along with an in-depth explanation are helpful.

You can use `Fixes #<issue number>` to associate this PR to an existing issue.

Simple docs PR. Should not require testing

If you don't know how to write and run your tests, please read the guide:
https://doc.crates.io/contrib/tests

### Additional information

Other information you want to mention in this PR, such as prior arts,
future extensions, an unresolved problem, or a TODO list.

### DIFF
--- a/src/doc/src/reference/profiles.md
+++ b/src/doc/src/reference/profiles.md
@@ -206,7 +206,7 @@ The `rustc` test harness currently requires `unwind` behavior. See the
 [`panic-abort-tests`] unstable flag which enables `abort` behavior.
 
 Additionally, when using the `abort` strategy and building a test, all of the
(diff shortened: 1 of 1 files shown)


ehuss:
The current docs are correct. When building tests, cargo will not use `abort` because that is incompatible with test harness, which requires the use of unwinding to determine when a test fails. This ties in to the previous paragraph which explains that `panic` is ignored for tests, and this sentence is further clarifying that *all* dependencies will use the `unwind` strategy.

weihanglo:
Closing as the doc correctly describes the current behavior and the limitation.

Issue #11702, opened 2023-02-11
Cargo should give helpful errors when unsupported flags are used.


jonhoo:
### Problem

There are several flags one might imagine users passing to Cargo that aren't supported _yet_. It'd be nice if we told users who try to use these what to use instead, and perhaps also linked to the issue (if any) proposing to add such a flag. Some examples:

- [x] There's no `-C` for running in a different directory. (blocked on #11957)
- [x] There's no `-w` or `-W` short for `--workspace` (except in `update`, blocked on #11554).
- [x] There's no `-x` for `--exclude` #12805
- [x] There's no `-t` for `--target` #12805
- [x] There's no `-s` for `--quiet` (`-s` is "silent" in `curl`)  https://github.com/rust-lang/cargo/pull/12723
- [ ] It's `--manifest-path`, not `--path` (though there _is_ `--path` for `cargo add`)
- [x] There's no `-c` for `--config` https://github.com/rust-lang/cargo/pull/12777
- [x] There's no `-z`, only `-Z` #12788
- [x] There's no `-O`, only `--out-dir` https://github.com/rust-lang/cargo/pull/12755
- [x] There's no `-n` for `--dry-run` https://github.com/rust-lang/cargo/pull/12660

weihanglo:
Does it make sense having this as a built-in mechanism in clap for suggesting an alternative flag?

epage:
@weihanglo I think that can make sense to seed clap with specific suggestions and not just rely on the automatic suggestions and that would help simplify things for application, like cargo.  Mind opening an issue?

weihanglo:
More examples:

- There's no `--debug` for `cargo build`. Only `--release`.
- There's no `--release` for `cargo install`. Only `--debug`.

These does confuse people a lot. See #12009.

weihanglo:
I am not sure if we should wait for upstream fix or make our own custom suggestion. Let's label it as waiting for upstream first. We still **seek a ad-hoc solution for each case** if somebody is interested in.

@rustbot label +S-blocked-external

weihanglo:
https://github.com/clap-rs/clap/pull/5075 just got landed with [clap@4.3.22](https://crates.io/crates/clap/4.3.22). People, we now can have basic suggestions for unsupported arguments!

I'll write down a list of flags of interest and instructions shortly.

@rustbot label -S-blocked-external +S-accepted +E-easy

weihanglo:
[...]

PR #12693, opened 2023-09-19
Better suggestion for redundant mode in build and install commands

hi-rustin:

See: https://github.com/rust-lang/cargo/issues/11702#issuecomment-1517563353

Added better suggestions for unsupported mode in build and install commands.

Check out the unit tests.

### Additional information

r? @weihanglo

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -525,7 +525,7 @@ Run `{cmd}` to see possible targets."
 
         let name = match (
             self.maybe_flag("release"),
-            self.flag("debug"),
+            self.maybe_flag("debug"),
             specified_profile,
         ) {
             (false, false, None) => default,
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -112,17 +112,15 @@ pub trait CommandExt: Sized {
         self._arg(flag("keep-going", "").value_parser(value_parser).hide(true))
     }
 
-    fn arg_unsupported_mode(
+    fn arg_redundant_default_mode(
         self,
-        want: &'static str,
+        default_mode: &'static str,
         command: &'static str,
-        actual: &'static str,
+        supported_mode: &'static str,
     ) -> Self {
-        let msg = format!(
-            "There is no `--{want}` for `cargo {command}`. Only `--{actual}` is supported."
-        );
+        let msg = format!("`--{default_mode}` is the default for `cargo {command}`; instead `--{supported_mode}` is supported");
         let value_parser = UnknownArgumentValueParser::suggest(msg);
-        self._arg(flag(want, "").value_parser(value_parser).hide(true))
+        self._arg(flag(default_mode, "").value_parser(value_parser).hide(true))
     }
 
     fn arg_targets_all(
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -112,6 +112,19 @@ pub trait CommandExt: Sized {
         self._arg(flag("keep-going", "").value_parser(value_parser).hide(true))
     }
 
+    fn arg_unsupported_mode(
+        self,
+        want: &'static str,
+        command: &'static str,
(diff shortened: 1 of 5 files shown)


src/cargo/util/command_prelude.rs
```diff
@@ -112,6 +112,19 @@ pub trait CommandExt: Sized {
...
     }
 
+    fn arg_unsupported_mode(
+        self,
+        want: &'static str,
+        command: &'static str,
+        actual: &'static str,
+    ) -> Self {
+        let msg = format!(
+            "There is no `--{want}` for `cargo {command}`. Only `--{actual}` is supported."
```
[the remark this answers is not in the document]
hi-rustin: Thanks for the suggestion!

I prefer the current one. If you use it we will get:
```diff
3        -  tip: There is no `--release` for `cargo install`. Only `--debug` is supported.
    3    +  tip: a similar argument exists: 'debug'
```

I think the current one is more makes sense and more clear.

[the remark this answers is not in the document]
weihanglo: No problem. Just a nit: the convention is using lower case for the first letter.

```suggestion
            "there is no `--{want}` for `cargo {command}`. Only `--{actual}` is supported."
```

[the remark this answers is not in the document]
hi-rustin: Updated! Good catch!

src/cargo/util/command_prelude.rs
```diff
@@ -119,7 +119,7 @@ pub trait CommandExt: Sized {
         actual: &'static str,
     ) -> Self {
         let msg = format!(
-            "There is no `--{want}` for `cargo {command}`. Only `--{actual}` is supported."
+            "there is no `--{want}` for `cargo {command}`. Only `--{actual}` is supported."
```
epage, line 122: nit: should that be a `,` or `;` before `Only`?

Overall, I'd probably simplify it to `cargo {command} supports `--{actual}` instead` since the main message already has "unexpected argument".

hi-rustin: Updated! Thanks!

epage: I'm sorry, there is way to much back and forth on this.

Reading back over this, I realized what was missing or isn't clear.

```
`{want}` is the default for `cargo {command}`; instead `--actual` is supported
```

The downside to this is that it is failure specialized.  If we go with this wording, we should either
- Not use the helper
- Clarify the naming of the helper that it is about being the default mode
- Leave the message as is and go ahead and merge this

hi-rustin: > `{want}` is the default for `cargo {command}`; instead `--actual` is supported

I don't really get this. Do you mean `{actual}` is the default for `cargo {command}`; instead `--actual` is supported `?



> * Not use the helper
> * Clarify the naming of the helper that it is about being the default mode

The current helper name is `arg_unsupported_mode`. It helps to catch the unsupported mode instead of indicating the default mode. So I guess its name is OK?

epage:
```
error: unexpected argument '--release' found

  tip: `cargo install` supports `--debug` instead.

Usage: cargo[EXE] install [OPTIONS] [crate]...

```
Here, `--release` is unexpected because the `release` profile is the default.  `--debug` isn't really an alternative to use instead but is instead what we expose for controlling the profile.

So this isn't about an unsupported mode in this case but that the flag is unexpected because its redundant.

Maybe in other cases, a `arg_unsupported_mode` function with a message like that will make sense but I feel like here we should be focusing on the fact that the flag is redundant.

hi-rustin: Oh, I got it. Make sense. You are right it is a bad name and the tips are confusing.

hi-rustin: Updated. Thanks for your help!

epage (approved):
@weihanglo since you were originally looking at this, any other thoughts?

weihanglo (approved):
This looks better. Thank you all for the help.

PR #12662, opened 2023-09-13
Limit cargo add feature print

Angelin01:
### Readyness checklist

- [x] Test validating activated features under limit, total over limit
- [x] Test validating activated features over limit
- [x] At most 49 features are displayed
- [x] "X more features" messages are displayed
- ~~[ ] The "X more features" messages should be hyperlinks~~

### What does this PR try to resolve?

Fixes #12617 

Limits the amount of printed features to 49. After that, it simply prints "... X more activated features" and "... Y more deactivated features".

### How should we test and review this PR?

Two new tests were added to validate the functionality, see `tests/testsuite/cargo_add/features_too_many_activated`
and tests/testsuite/cargo_add/features_too_many_few_activated.

The first commit shows the output before this PR, the second with the modifications.

To test it live, run cargo add with a crate with a ton of features, such as [web-sys](https://docs.rs/crate/web-sys/latest/features)

A few commands and sample outputs:

<details>
  <summary><b>Adding web-sys with no features activated</b></summary>

  Command:
```shell
cargo add --dry-run web-sys
```

  Output:
```shell
      Updating crates.io index
      Adding web-sys v0.3.64 to dependencies.
             Features:
             - AbortController
             - AbortSignal
             - AddEventListenerOptions
             - AesCbcParams
             - AesCtrParams
             - AesDerivedKeyParams
             - AesGcmParams
             - AesKeyAlgorithm
             - AesKeyGenParams
             - Algorithm
             - AlignSetting
             - AllowedBluetoothDevice
             - AllowedUsbDevice
             - AlphaOption
             - AnalyserNode
             - AnalyserOptions
             - AngleInstancedArrays
             - Animation
             - AnimationEffect
             - AnimationEvent
             - AnimationEventInit
             - AnimationPlayState
             - AnimationPlaybackEvent
             - AnimationPlaybackEventInit
             - AnimationPropertyDetails
             - AnimationPropertyValueDetails
             - AnimationTimeline
             - AssignedNodesOptions
             - AttestationConveyancePreference
             - Attr
             - AttributeNameValue
             - AudioBuffer
             - AudioBufferOptions
             - AudioBufferSourceNode
             - AudioBufferSourceOptions
             - AudioConfiguration
             - AudioContext
             - AudioContextOptions
             - AudioContextState
             - AudioData
             - AudioDataCopyToOptions
             - AudioDataInit
             - AudioDecoder
             - AudioDecoderConfig
             - AudioDecoderInit
             - AudioDecoderSupport
[...]

PR #12702, opened 2023-09-19
Cargo add displays either feature list or summarized count

### LINKED ISSUE #12617: Limit the number of crate features that are displayed as result of `cargo add`
### Problem

Some crates have way too many features (such as [web-sys](https://docs.rs/crate/web-sys/latest/features)). `cargo add` lists every feature when adding a crate and could potentially even overflow terminal buffer (it just did for me with web-sys using console)

### Proposed Solution

Limit the number of features that are displayed. The limit should preferably be configurable but a hard-coded one is also fine. 

If the feature list is truncated, cargo should link to docs.rs feature list for that crate

### Notes

_No response_

Angelin01:

Fixes #12617

Adjusts cargo add to display only the activated/deactivate feature list or the summary, not both.

Simply run `cargo add` with some features enabled. [The previous PR](https://github.com/rust-lang/cargo/pull/12662) also has a few example commands.

Tests were adjusted and have sample output as well.

### Additional Information

This first version does not consider a variable terminal size, as suggested in the issue.

There's also some minor code refactor, trying to improve readability. Please warn if the changes are undesirable


src/cargo/ops/cargo_add/mod.rs
```diff
@@ -943,56 +943,57 @@ fn print_dep_table_msg(shell: &mut Shell, dep: &DependencyUI) -> CargoResult<()>
...
-                let remaining = total_activated - activated_printed;
-                shell.write_stderr(
-                    format_args!("{prefix}... {remaining} more activated features\n"),
-                    &style::NOP,
-                )?;
-                break;
-            }
+    if activated.is_empty() && deactivated.is_empty() {
+        return Ok(());
+    }
```
epage, line 950: While I generally prefer guard statements / early returns,. I don't think its appropriate here as we could add more formatting after features and could easily miss his early return and miss it in the output that the extra information is dropped.

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -943,56 +943,57 @@ fn print_dep_table_msg(shell: &mut Shell, dep: &DependencyUI) -> CargoResult<()>
...
+            &style::NOP,
+        )?;
+        return Ok(());
+    }
+
+    for feat in deactivated {
+        shell.write_stderr(&prefix, &style::NOP)?;
+        shell.write_stderr('-', &style::ERROR)?;
+        shell.write_stderr(format_args!(" {feat}\n"), &style::NOP)?;
     }
```
epage, line 991: For me, early returns are for guard statements / boilerplate.  When it involves business logic, `else`s should be used.  This is on top of the problem of bugs if/when new features are added as pointed out in m,y other comment.

This can be as simple as:
```rust
if activated.len() <MAX_FEATURE_PRINTS {
    ... for loop
} else {
    .. summary
}
if total_activated + total_deactivated < MAX_FEATURE_PRINTS {
    ... for loop
} else {
    ... summary
}
```

Angelin01: Agreed for this case

src/cargo/ops/cargo_add/mod.rs
Angelin01: I'd argue in this case it serves a similar purpose as lines 943~945:

```rs
    if matches!(shell.verbosity(), crate::core::shell::Verbosity::Quiet) {
        return Ok(());
    }
```

To me, these lines are saying: "If we don't need to display anything, stop here and leave". What are the reasons we don't display anything?
- The shell verbosity is set to quiet
- There are no features to display

In this case, I'd really say this is the intended logic.

But then again, you could easily argue the opposite: "if there are features to display, display them". I'd still lean towards keeping this section as is, but I don't have a strong enough opinion about it to not change it if you prefer the late return.

> and could easily miss his early return and miss it in the output that the extra information is dropped.

I think the tests would catch these cases quite easily.

epage: > I'd argue in this case it serves a similar purpose as lines 943~945:

That block says "everything after this should be bypassed for quiet".  That is very different from what your code is doing which can early return before feature unrelated stuff.

I feel fairly strongly that an early return here is inappropriate and you are laying a trap for future cargo contributors.

> I think the tests would catch these cases quite easily.

It would require people writing new tests for there could that hit it.

Angelin01: That's understandable, I've reverted this change

epage:
Yes, terminal interactions don't get as much test coverage; thats understandable.  terminal_size also isn't a blocker for this.

Angelin01:
Thank you for the help and the reviews. Despite the confusion, I'll stand by what I said [here](https://github.com/rust-lang/cargo/pull/12662#issuecomment-1719444011), the whole process, from my end, has been quite painless and fruitful.

PR #12706, opened 2023-09-19
Add new package cache lock modes

ehuss:
The way locking worked before this PR is that only one cargo could write to the package cache at once (otherwise it could cause corruption). However, it allowed cargo's to read from the package cache while running a build under the assumption that writers are append-only and won't affect reading. This allows multiple builds to run concurrently, only blocking on the part where it is not possible to run concurrently (downloading to the cache).

This introduces a new package cache locking strategy to support the ability to safely modify existing cache entries while other cargos are potentially reading from the cache. It has different locking modes:

- `MutateExclusive` (new) — Held when cargo wants to modify existing cache entries (such as being introduced for garbage collection in #12634), and ensures only one cargo has access to the cache during that time.
- `DownloadExclusive`  (renamed) — This is a more specialized name for the lock that was before this PR.   A caller should acquire this when downloading into the cache and doing resolution.  It ensures that only one cargo can append to the cache, but allows other cargos to concurrently read from the cache.
- `Shared` (new) — This is to preserve the old concurrent build behavior by allowing multiple concurrent cargos to hold this while a build is running when it is reading from the cache

**Reviewing suggestions:**
There are a few commits needed to help with testing which are first. The main commit has the following:
- `src/cargo/util/cache_lock.rs` is an abstraction around package cache locks, and is the heart of the change. It should have comments and notes which should guide what it is doing. The `CacheLocker` is stored in `Config` along with all our other global stuff.
- Every call to `config.acquire_package_cache_lock()` has been changed to explicitly state which lock mode it wants to lock the package cache in.
- `Context::compile` is the key point where the `Shared` lock is acquired, ensuring that no mutation is done while the cache is being read.
- `MutateExclusive` is not used in this PR, but is being added in preparation for #12634.
- The non-blocking `try_acquire_package_cache_lock` API is not used in this PR, but is being added in preparation for #12634 to allow automatic gc to skip running if another cargo is already running (to avoid unnecessary blocking).
- `src/cargo/util/flock.rs` has been updated with some code cleanup (removing unused stuff), adds support for non-blocking locks, and renames some functions to make their operation clearer.
- `tests/testsuite/cache_lock.rs` contains tests for all the different permutations of ways of acquiring locks.

### DIFF
--- a/src/cargo/util/cache_lock.rs
+++ b/src/cargo/util/cache_lock.rs
@@ -136,19 +136,6 @@ pub enum CacheLockMode {
     MutateExclusive,
 }
 
-/// A locker that can be used to acquire locks.
-///
-/// See the [`crate::util::cache_lock`] module documentation for an overview
-/// of how cache locking works.
-#[derive(Debug)]
-pub struct CacheLocker {
-    /// The state of the locker.
-    ///
-    /// [`CacheLocker`] uses interior mutability because it is stuffed inside
-    /// the global `Config`, which does not allow mutation.
-    state: RefCell<CacheState>,
-}
-
 /// Whether or not a lock attempt should block.
 #[derive(Copy, Clone)]
 enum BlockingMode {
@@ -371,85 +358,6 @@ struct CacheState {
     mutate_lock: RecursiveLock,
 }
 
-/// A held lock guard.
-///
-/// When this is dropped, the lock will be released.
-#[must_use]
-pub struct CacheLock<'lock> {
-    mode: CacheLockMode,
-    locker: &'lock CacheLocker,
-}
-
-/// The filename for the [`CacheLockMode::DownloadExclusive`] lock.
-const CACHE_LOCK_NAME: &str = ".package-cache";
-/// The filename for the [`CacheLockMode::MutateExclusive`] and
-/// [`CacheLockMode::Shared`] lock.
-const MUTATE_NAME: &str = ".package-cache-mutate";
-
-// Descriptions that are displayed in the "Blocking" message shown to the user.
-const SHARED_DESCR: &str = "shared package cache";
-const DOWNLOAD_EXCLUSIVE_DESCR: &str = "package cache";
-const MUTATE_EXCLUSIVE_DESCR: &str = "package cache mutation";
-
-impl CacheLocker {
-    /// Creates a new `CacheLocker`.
-    pub fn new() -> CacheLocker {
-        CacheLocker {
-            state: RefCell::new(CacheState {
-                cache_lock: RecursiveLock::new(CACHE_LOCK_NAME),
-                mutate_lock: RecursiveLock::new(MUTATE_NAME),
-            }),
-        }
-    }
-
-    /// Acquires a lock with the given mode, possibly blocking if another
-    /// cargo is holding the lock.
-    pub fn lock(&self, config: &Config, mode: CacheLockMode) -> CargoResult<CacheLock<'_>> {
-        let mut state = self.state.borrow_mut();
-        let _ = state.lock(config, mode, Blocking)?;
-        Ok(CacheLock { mode, locker: self })
-    }
-
-    /// Acquires a lock with the given mode, returning `None` if another cargo
-    /// is holding the lock.
-    pub fn try_lock(
-        &self,
-        config: &Config,
-        mode: CacheLockMode,
-    ) -> CargoResult<Option<CacheLock<'_>>> {
-        let mut state = self.state.borrow_mut();
-        if state.lock(config, mode, NonBlocking)? == LockAcquired {
-            Ok(Some(CacheLock { mode, locker: self }))
-        } else {
-            Ok(None)
-        }
-    }
-
-    /// Returns whether or not a lock is held for the given mode in this locker.
-    ///
-    /// This does not tell you whether or not it is locked in some other
-    /// locker (such as in another process).
-    ///
-    /// Note that `Shared` will return true if a `MutateExclusive` lock is
-    /// held, since `MutateExclusive` is just an upgraded `Shared`. Likewise,
-    /// `DownlaodExclusive` will return true if a `MutateExclusive` lock is
-    /// held since they overlap.
-    pub fn is_locked(&self, mode: CacheLockMode) -> bool {
-        let state = self.state.borrow();
-        match (
-            mode,
-            state.cache_lock.count,
-            state.mutate_lock.count,
-            state.mutate_lock.is_exclusive,
-        ) {
-            (CacheLockMode::Shared, _, 1.., _) => true,
-            (CacheLockMode::MutateExclusive, _, 1.., true) => true,
-            (CacheLockMode::DownloadExclusive, 1.., _, _) => true,
-            _ => false,
-        }
-    }
-}
-
 impl CacheState {
     fn lock(
         &mut self,
@@ -513,20 +421,13 @@ impl CacheState {
     }
 }
 
-/// Returns whether or not the error appears to be from a read-only filesystem.
-fn maybe_readonly(err: &anyhow::Error) -> bool {
-    err.chain().any(|err| {
-        if let Some(io) = err.downcast_ref::<io::Error>() {
-            if io.kind() == io::ErrorKind::PermissionDenied {
-                return true;
-            }
-
-            #[cfg(unix)]
-            return io.raw_os_error() == Some(libc::EROFS);
-        }
-
-        false
-    })
+/// A held lock guard.
+///
+/// When this is dropped, the lock will be released.
+#[must_use]
+pub struct CacheLock<'lock> {
+    mode: CacheLockMode,
+    locker: &'lock CacheLocker,
 }
 
 impl Drop for CacheLock<'_> {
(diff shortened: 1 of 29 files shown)


ehuss:
This is split off from #12634.

epage:
Could you go into why these specific cache modes were chosen?  What I'm wanting to understand is which areas of performance are we concerned about and how much this helps.  I feel like the package cache locking in cargo was already confusing (being a un-Rust-like lock) and I'm concerned this will make things harder.

src/cargo/util/flock.rs
```diff
@@ -193,57 +195,78 @@ impl Filesystem {
...
-        )
+        let (path, f) = self.open(path.as_ref(), &OpenOptions::new().read(true), false)?;
+        self.lock(path, f, config, msg, false)
     }
 
-    fn open(
+    /// Opens shared access to a file, returning the locked version of a file.
+    ///
+    /// Compared to [`Filesystem::open_ro`], this will create the file (and
+    /// any directories in the parent) if the file does not already exist.
```
epage, line 205: Why this difference?

src/cargo/util/flock.rs
```diff
@@ -193,57 +195,78 @@ impl Filesystem {
...
+            acquire(config, msg, &path, &|| try_lock_shared(&f), &|| {
+                lock_shared(&f)
+            })?;
         }
-        Ok(FileLock {
-            f: Some(f),
-            path,
-            state,
-        })
+        Ok(FileLock { f: Some(f), path })
```
epage, line 269: Is this still worth it?

I felt this made the code more confusing due to the lack of parallels between the try and non-try variants.

src/cargo/util/cache_lock.rs
```diff
@@ -0,0 +1,463 @@
...
+        }
+    };
+}
+
+/// Macro to help with acquiring a lock.
+///
+/// `shared` and `exclusive` are inputs to the macro so that either the
+/// blocking or non-blocking implementations can be called based on what the
+/// caller wants.
+macro_rules! do_lock {
```
epage, line 249: Why all the macros?

src/cargo/util/cache_lock.rs
```diff
@@ -0,0 +1,463 @@
...
+//!   collection runs.
+//! * [`CacheLockMode::MutateExclusive`] -- This is an exclusive lock acquired
+//!   whenever needing to modify existing source files (for example, with
+//!   cache garbage collection). This is acquired to make sure that no other
+//!   cargo is reading from the cache.
+//!
+//! Importantly, a `DownloadExclusive` lock does *not* interfere with a
+//! `Shared` lock. The download process generally does not modify source
+//! files, so other cargos should be able to safely proceed in reading source
+//! files[^1].
```
epage, line 22: I feel like most of this should be on the enum because that is where people will look (for what might not already be) and we should avoid duplicating such detailed documentation.

epage: Looks like these descriptions are more user facing / policy driven while the enum has more technical details

src/cargo/util/cache_lock.rs
```diff
@@ -0,0 +1,463 @@
...
+    Shared,
+    /// A `DownloadExclusive` lock ensures that only one cargo is downloading
+    /// new packages.
+    ///
+    /// If another cargo has a `MutateExclusive` lock, then an attempt to get
+    /// a `DownloadExclusive` lock will block.
+    ///
+    /// If another cargo has a `Shared` lock, then both can operate
+    /// concurrently.
+    DownloadExclusive,
```
epage, line 90: Seems like this should be the common / default choice.  Maybe should put it first.

src/cargo/util/cache_lock.rs
```diff
@@ -0,0 +1,463 @@
...
+use super::FileLock;
+use crate::CargoResult;
+use crate::Config;
+use anyhow::Context;
+use std::cell::RefCell;
+use std::io;
+
+/// The style of lock to acquire.
+#[derive(Copy, Clone, Debug)]
+pub enum CacheLockMode {
```
epage, line 72: Whether on the enum or the variants, I feel like this could be a little more caller-focused in helping callers know what lock type they should choose.

src/cargo/util/flock.rs
ehuss: The first time locking the package cache, it needs to create the lock file in order to lock it.

The `open_ro` variant is used with locking `Cargo.lock`, where it doesn't want to create an empty file if `Cargo.lock` doesn't exist.

So, one is used for locking when there isn't anything relevant to lock against (like the package cache), and the other is for locking pre-existing files (like `Cargo.lock`).

src/cargo/util/flock.rs
ehuss: I'm not sure if I'm understanding. I assume this is asking if this should just be inlined? That seems fine to me. I don't like bool arguments very much, but I felt like the `acquire` call was complex enough to justify trying to share it.

src/cargo/util/cache_lock.rs
ehuss: Without the macros, the equivalent permutations would all need to be expanded as separate methods (in order to access the correct fields). It would maybe roughly double the code, duplicating parts that I think are quite subtle and important not to duplicate.

src/cargo/util/cache_lock.rs
ehuss: Can you say what on the enum seems unclear? Or is the concern that this lists the modes here? The intent is the module docs provide a global description of how everything in the module works and fits together. It seems hard to me to do that without first introducing what possible lock modes exist. This introduction is about as brief as possible needed to give a high-level understanding of what the modes are, and then the enum itself has detailed documentation on each mode. I could remove the list here and say "go read the `CacheLockMode` docs first, and then come back", but that seems a little awkward to me.

Or are you asking to move the entire module docs to the enum? That doesn't seem quite right to me because this documentation is covering more than just the modes, but instead how everything here fits together (modes, locker, files, invariants the locker needs to hold, etc.).

src/cargo/util/cache_lock.rs
ehuss: I'm not quite understanding what isn't clear or what could be added. If the caller is downloading packages to the cache, then it should have the `DownloadExclusive` lock. If the caller is reading from the cache (like when a build is running), then it should have the `Shared` lock. If the caller is modifying the cache, it should have `MutateExclusive`, as it mentions an example as garbage collection. I feel like each variant is saying how to choose it, so I'm not understanding what is missing.

ehuss:
> Could you go into why these specific cache modes were chosen? What I'm wanting to understand is which areas of performance are we concerned about and how much this helps.

The way locking worked before this PR is that only one cargo could write to the package cache at once (otherwise it could cause corruption). However, it allowed cargo's to read from the package cache while running a build under the assumption that writers won't modify anything that it is reading. This allows multiple builds to run concurrently, only blocking on the part where it is not possible to run concurrently (downloading to the cache).

This adds a "mutate" mode that ensures that no other cargo is accessing the cache in any way while it is running. This allows garbage collection to make any modifications it wants without worrying about stomping on something another cargo might be using.

The "share" mode is just a formalization of the way cargo used to run while it was doing a build. The reason it was added is to have something to block on the "mutate" lock, to ensure that no cargo is accessing the package cache while a mutate lock is held.

I'm not sure if that answers your question. The only performance concern is to ensure that multiple builds can run concurrently without blocking one another, and this preserves that behavior.

src/cargo/util/flock.rs
epage: If that is the main distinction, I feel like the name should convey that rather than trying to push the "shared" policy naming down to this level.

src/cargo/util/cache_lock.rs
epage: When I looked over the PR this was pulled out of, I was confused for a while as to why `cargo add` and `cargo remove` were `DownloadExclusive`.  I'm not downloading anything to the cache (though `cargo add` downloads to the index).  `cargo remove` only downloads in corner cases.  Looking over this PR, its clearer to me now but when someone comes p and writes code, they're likely to either pick an over-restrictive mode or just play error whack-a-mole until they get it right.

src/cargo/util/cache_lock.rs
epage: The following is in the mod comment and is clearer than what is on the enum

> acquired while downloading packages and doing resolution.

"Oh, resolution, thats why `cargo add` and `cargo remove` need it".

One challenge with this doc comment is its serving double-duty.  It covers both user-facing information and implementation details.  Looking at it, I could see also moving the "old cargo versions" part down to the enum as well.

Maybe a different angle: how much of these details should even be a doc  comment on the mod vs a doc or regular comment on `CacheState`?

src/cargo/util/cache_lock.rs
epage: I feel like at least some of them would be just as clear or clearer as functions that take `&mut` arguments, like `decrement` or `do_shared_lock`.

epage:
That makes sense.  I also realized the intention last night.  I've attempted to merge your description into the PR description; feel free to clear up anything I made confusing.

src/cargo/util/flock.rs
ehuss: Sorry, I'm not quite sure what else would convey that better, or I'm not understanding your concern. It is opening a file with a shared lock, creating it if it doesn't exist. The name seems to be describing exactly what it does, so I'm not clear what it could be instead?

Or are you referring to `open_ro`? I could imagine that being `open_shared_ro` to make it more explicit what it is doing (though read-only implies it is shared).

src/cargo/util/cache_lock.rs
ehuss: I have rewritten this to not use macros. It's a little longer to accommodate some extra types, but I think still retains the goal of avoiding too much duplication.

src/cargo/util/cache_lock.rs
ehuss: I think it is useful to have an overview of how everything fits together in one place. There are important aspects of the high-level picture that I think are useful to know, as well as some of the lower-level details (since these aren't perfect abstractions). If everything was split up, then one would need to go hunting through multiple places to get a understanding of how the pieces fit together and what the different inter-playing aspects are.  Things like the old cargo versions isn't relevant to choosing one of the enum variants, so I'm not sure I understand why it would be moved there. The intent is the enum docs are specifically for choosing which enum to use, not to explain how the system works overall (which I think would distract from the act of trying to pick a variant).

I have added the "and resolution" part to the enum. Otherwise, I think the enum docs currently explain how to choose which variant to use.

Is this an important concern, or a non-blocking nit?

src/cargo/util/flock.rs
epage: The lack of consistency in the naming was drawing my attention to the wrong things, making this harder to understand.

One idea
- `open_shared`
- `create_shared` or `create_or_open_shared` (some `create`s are idempotent while others aren't)
- `create_exclusive` or `create_or_open_exclusive`

src/cargo/util/flock.rs
epage: Thanks!

src/cargo/util/cache_lock.rs
epage: Thanks!

src/cargo/util/cache_lock.rs
epage: > I have added the "and resolution" part to the enum. Otherwise, I think the enum docs currently explain how to choose which variant to use.

Only `DownloadExclusive` has guidance on when to use it.  Ways to improve it (part of my "move things down").  The top-level docs talk about builds for `shared` and garbage collection for `garbage collection`.  Having those in would at least provide some context to what people are reading when reading the documentation for the variants.

> Is this an important concern, or a non-blocking nit?

With how complicated this is, I would consider having actionable documentation to be blocking.  Refinement past that would not be.

src/cargo/util/flock.rs
ehuss: I decided to go with the following:

| Before | After |
|--------|-------|
| open_rw | open_rw_exclusive_create |
| try_open_rw | try_open_rw_exclusive_create |
| open_ro | open_ro_shared |
| open_shared_create | open_ro_shared_create |
| try_open_shared_create | try_open_ro_shared_create |

The form is roughly (`try_`) `open_` *access_mode* *lock_mode* (`_create`).

The intent is to be explicit about exactly what the method is doing. I'm not a huge fan of wordy methods, but I think each word is important to understand what the method is doing. An alternate solution might be to use a builder (similar to `OpenOptions`), but with only 5 variants, I'm not sure that is necessary at this stage. If there were more, I would advocate for that.

src/cargo/util/cache_lock.rs
ehuss: I extended the documentation with more guidance on how to use the locks and the different modes.

src/cargo/util/cache_lock.rs
ehuss: Each variant now gives guidance on when to use it.

src/cargo/util/flock.rs
epage: Not the biggest fan but it works

src/cargo/util/cache_lock.rs
epage: Thanks!

epage:
Thanks for your patience in working with my feedback!

PR #12708, opened 2023-09-19
Ensure package IDs are unique when cleaning.

jfgoog:
get_many will panic if there are duplicate IDs.

### DIFF
--- a/src/cargo/ops/cargo_clean.rs
+++ b/src/cargo/ops/cargo_clean.rs
@@ -8,6 +8,7 @@ use crate::util::interning::InternedString;
 use crate::util::{human_readable_bytes, Config, Progress, ProgressStyle};
 use anyhow::bail;
 use cargo_util::paths;
+use indexmap::IndexSet;
 use std::fs;
 use std::path::{Path, PathBuf};
 
@@ -129,7 +130,7 @@ fn clean_specs(
     // Doc tests produce no output.
 
     // Get Packages for the specified specs.
-    let mut pkg_ids = Vec::new();
+    let mut pkg_ids = IndexSet::new();
     for spec_str in spec.iter() {
         // Translate the spec to a Package.
         let spec = PackageIdSpec::parse(spec_str)?;


src/cargo/ops/cargo_clean.rs
```diff
@@ -104,7 +105,7 @@ pub fn clean(ws: &Workspace<'_>, opts: &CleanOptions<'_>) -> CargoResult<()> {
     // Doc tests produce no output.
 
     // Get Packages for the specified specs.
-    let mut pkg_ids = Vec::new();
+    let mut pkg_ids = BTreeSet::new();
```
epage, line 108: Should we instead use `IndexSet` to preserve the users order?

epage, line 108: Can you add a test for this?  Ideally the first commit would have the test with a `#[should_panic]` and then the second commit would include this with the test updated.   Using these two commits will make it easier to ensure the test is testing the right thing and to show to reviewers what the intended behavior change is (especially since this is a PR without an issue)

jfgoog: An internal call to get_many() from download_accessible() in package.rs uses a BTreeSet, so that's what I used here. I don't know enough to say whether an IndexSet is warranted.

jfgoog: I took a quick look at this, and adding a test seems fairly complicated. There are not any tests in cargo_clean.rs, and most of it consists of a monolithic clean() function, in which this code change is buried. I would need to construct an appropriate Workspace object, which is fairly complicated.

It looks like there are some cursory integration tests, but adding something would require figuring out exactly how to trigger a duplicate ID, which might be tricky.

Overall, it doesn't look at all straightforward to test. :( I will keep poking at it, though, and see if I can figure anything out.

epage: Look in `tests/testsuite/clean.rs` for tests.

epage: I'd lean towards keeping ordering.

jfgoog: Done.

jfgoog:
After pulling in #12638 I am no longer getting a panic, so I am going to abandon this.

Issue #8531, opened 2020-07-23
cargo test --doc not load same shared libraries as cargo test --lib


adamcrume:
**Problem**
`cargo test --doc` doesn't appear to be loading the same shared libraries as `cargo test --lib` in 1.45

**Steps**
1. Clone https://github.com/tensorflow/rust
2. Set Rust to 1.44
3. Note that both `cargo test --doc` and `cargo test --lib` pass
4. Set Rust to 1.45
5. Note that `cargo test --lib` passes but `cargo test --doc` fails with `error while loading shared libraries: libtensorflow.so.1: cannot open shared object file: No such file or directory`

**Notes**

Output of `cargo version`:
cargo 1.44.0 (05d080faa 2020-05-06) or cargo 1.45.0 (744bd1fbb 2020-06-15)

Running on Ubuntu 20.04 LTS.

thoren-d:
For what it's worth, I'm seeing the same issue running doctests against my library that uses SDL2 (https://github.com/Rust-SDL2/rust-sdl2). `cargo +1.44.0 test --doc` works.

raskr:
Same problem with macOS + rust1.48+, but not with rust1.42.
`cargo test --doc` seems to be broken.

adamcrume:
Using `git bisect`, I narrowed down to 3fd28143de72d2d0e63186273d68aa890d8297e7 being the breaking commit.

pipehappy1:
This is still broken on 1.58

Trolldemorted:
This issue is known for almost 3 years, could you give us a timeline when a fix will come?

justsmth:
Is there a workaround for this issue?

Trolldemorted:
@adamcrume do you have any clue what might have caused this bug? Is @alexcrichton still around? Iirc he was stepping down from several Rust projects, but not cargo.

adamcrume:
I don't know what specifically in 3fd28143de72d2d0e63186273d68aa890d8297e7 caused this.  I'm not familiar with the cargo codebase.

weihanglo:
We appreciate Alex efforts on Cargo and Alex has already retired from the Cargo Team.[^1] 

There was a previous attempt <https://github.com/rust-lang/cargo/pull/10469> and concerns were listed <https://github.com/rust-lang/cargo/pull/10469#issuecomment-1079503079>. Feel free to pick it up and continue.

[...]

Issue #3366, opened 2016-12-04
`cargo run` exports bad DYLD_LIBRARY_PATH on macOS


casey:
I've been going down the rabbit hole for a while on this one. Thankfully though, it seems like it's actually hopefully a pretty simple fix. Or, if it isn't, at least maybe I'm reporting the issue in the right repo ^_^

I originally thought it was a git2-rs issue, and reported it in https://github.com/alexcrichton/git2-rs/issues/175. Then I thought it was a rustc issue, and reported it here: https://github.com/rust-lang/rust/issues/36250.

Finally though, I've gotten to the bottom of it, and it appears to be a cargo bug.

A few things need to happen to trigger the issue:

1. We are linking ina  dynamic library that is not provided by the system, for example something from homebrew in `/usr/local/lib`, or from macports in `/opt/local/lib`.

2. We are also linking in a system framework.

3. In the nonstandard location in 1., there is a dynamically linked library which the system framework from 2. needs. However, this library in 1. is not the version that the framework expects.

4. We run the binary with `cargo run`.

To make everything concrete, let's say that we're linking in `openssl` provided by macports in `/opt/local/lib`, the framework we're using is the `ApplicationServices` framework, and we also happen to have `libiconv` installed in `/opt/local/lib`.

Cargo will export `DYLD_LIBRARY_PATH=/opt/local/lib:...`, and we'll get the following error:

```
: cargo run
   Compiling rust-dynamic-linker-error v0.0.0 (file:///Users/rodarmor/src/rust-dynamic-linker-error)
    Finished debug [unoptimized + debuginfo] target(s) in 0.34 secs
     Running `target/debug/rust-dynamic-linker-error`
dyld: Symbol not found: _iconv
  Referenced from: /System/Library/PrivateFrameworks/LanguageModeling.framework/Versions/A/LanguageModeling
  Expected in: /opt/local/lib/libiconv.2.dylib
 in /System/Library/PrivateFrameworks/LanguageModeling.framework/Versions/A/LanguageModeling
```

What's happening is that the `LanguageModeling` framework, which is a transitive dependency of the `ApplicationServices` framework also depends on `libiconv`. Normally it finds it in `/usr/lib/libiconv.dylib`. However, because of the value of `DYLD_LIBRARY_PATH`, it finds it in `/usr/local/lib/libiconv.dylib`, and we get the error because it is the wrong version with a missing symbol.

If we run the binary directly, without the environment variable, it works:

```
[...]

PR #12710, opened 2023-09-20
test for doctest path problem #8531, and solution

tompscanlan:
* add test reproducing #8531
* re-use solution from #10469
* solution fixes test

Fixes issue #8531 , wherein we have doctests that fail with a shared library error.

The issue seems painful enough that folks are changing their code to avoid this bug.  We need a solution.

I added a tentative test that reproduces the problem, and reusing #10469, show that solution works.
The repair in question is small. Toggle it off and the test fails with shared library problems.

### DIFF
--- a/src/cargo/core/compiler/compilation.rs
+++ b/src/cargo/core/compiler/compilation.rs
@@ -274,7 +274,7 @@ impl<'cfg> Compilation<'cfg> {
         is_rustc_tool: bool,
     ) -> CargoResult<ProcessBuilder> {
         let mut search_path = Vec::new();
-        // this divides our path based on CompileKind: host vs target.
+        // this divides rustc/rustdoc invocations and other executions of build artifacts
         if is_rustc_tool {
             search_path.extend(super::filter_dynamic_search_path(
                 self.native_dirs.iter(),
--- a/src/cargo/core/compiler/compilation.rs
+++ b/src/cargo/core/compiler/compilation.rs
@@ -274,6 +274,7 @@ impl<'cfg> Compilation<'cfg> {
         is_rustc_tool: bool,
     ) -> CargoResult<ProcessBuilder> {
         let mut search_path = Vec::new();
+        // this divides our path based on CompileKind: host vs target.
         if is_rustc_tool {
             search_path.extend(super::filter_dynamic_search_path(
                 self.native_dirs.iter(),
--- a/src/cargo/core/compiler/compilation.rs
+++ b/src/cargo/core/compiler/compilation.rs
@@ -275,6 +275,10 @@ impl<'cfg> Compilation<'cfg> {
     ) -> CargoResult<ProcessBuilder> {
         let mut search_path = Vec::new();
         if is_rustc_tool {
+            search_path.extend(super::filter_dynamic_search_path(
+                self.native_dirs.iter(),
+                &self.root_output[&CompileKind::Host],
+            ));
             search_path.push(self.deps_output[&CompileKind::Host].clone());
             search_path.push(self.sysroot_host_libdir.clone());
         } else {
--- a/tests/testsuite/test.rs
+++ b/tests/testsuite/test.rs
@@ -2,7 +2,6 @@
 
 use cargo_test_support::paths::CargoPathExt;
 use cargo_test_support::registry::Package;
-use cargo_test_support::Project;
 use cargo_test_support::{
     basic_bin_manifest, basic_lib_manifest, basic_manifest, cargo_exe, project,
 };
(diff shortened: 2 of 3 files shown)


tests/testsuite/test.rs
```diff
@@ -48,6 +49,37 @@ fn cargo_test_simple() {
         .run();
 }
 
+#[cargo_test]
+// download a known broken doctests code base and run cargo test, should fail
+fn cargo_test_tensorflow_not_failing_due_to_shared_library() {
+    use git2::build::RepoBuilder;
+
+    // git repo for cargo project containing doctests that require shared libraries
+    let url = "https://github.com/tensorflow/rust";
```
epage, line 58: Note that we don't rely on network resources in the tests

tompscanlan: This was created to reproduce only. I'm hoping you can help me get it shaped into the cargo way.

tompscanlan: I'm trying to id a minimal test for this, but it's taking longer than I want, so this was a shortcut to open discussion.

tompscanlan:
I added a test that reproduces the error initially seen in the initial tensorflow issue. I'd like to reduce that to a few files that reproduce the error so I can include it into a unit test.  I'm not having luck, so if anyone wants to take a crack at it, let me know. I can get you details.

I then pulled in @sgrif change from #10469 on top, and agree that the change is making the reproduced error go away. Thanks @sgrif! 

Additionally, I verified that the following can work as a temporary work around if you've got the shared library installed  in your system.  `/usr/loca/lib` is where the library was installed in my system.

    LD_LIBRARY_PATH=/usr/local/lib  cargo test --doc

I feel like adding a build.rs should also be another path to do accomplishing the same thing, but haven't made it work yet.

I'm going to go read #3366 and think on this a little longer.  Next goal is slimming this test down. Anything else you want to see to close this out?

tompscanlan:
I'm spinning my wheels on making a tiny test that replicates the problem.  I removed the test that hits the network. We lose the exact error replication, but the test that remains ensures we have good results assuming you get libs in the right place,.

@epage or @weihanglo , what else should I cover to get this closed?

tests/testsuite/test.rs
tompscanlan: removing this. only needed for validation.

djc:
We're running into this for rustls. Seems like a strange bug (potentially a regression?) with potentially a straightforward fix? Cargo team, any hints on moving this forward? I could invest some time to help this along if necessary.

src/cargo/core/compiler/compilation.rs
```diff
@@ -274,7 +274,12 @@ impl<'cfg> Compilation<'cfg> {
         is_rustc_tool: bool,
     ) -> CargoResult<ProcessBuilder> {
         let mut search_path = Vec::new();
+        // this divides our path based on CompileKind: host vs target.
         if is_rustc_tool {
+            search_path.extend(super::filter_dynamic_search_path(
```
weihanglo, line 279: I believe you've read this comment <https://github.com/rust-lang/cargo/pull/10469#issuecomment-1079503079> and [this zulip topic](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/dynamic.20linkers.20.26.20mixed.20targets). 

If I understand correctly, the issue is like

1. `rustdoc --test` does two things: compiling the example and running it
2. a proc-macro dep uses a shared lib “foo.so” from system
3. A build script builds a shared lib “foo.so” from vendored sources
4. Name collision happens as proc-macro is supposed to use "foo.so" from system not vendored one.

This should be unlikely as ehuss stated out:

> In general, using shared libraries with proc-macros has some bugs, so hopefully people don't actually use them too often. For example, I just discovered that `cargo doc` does not set the search path for build script OUT directories, so any proc-macro that tries to load a shared library generated by a build script will fail.

That being said, we can propably have patch like this PR and document the truth that rustc/cargo don't support proc-macro loading shared libs at this moment.

weihanglo (commented):
Sorry for the delay. While I think a proper solution would rustdoc delegate doctests runnning to cargo (which should be tracked in <https://github.com/rust-lang/testing-devex-team/issues/5>), we should try moving this forwards, or at least have a conclusion what to do next. Will put this in the agenda of the triage meeting.

src/cargo/core/compiler/compilation.rs
```diff
@@ -274,7 +274,12 @@ impl<'cfg> Compilation<'cfg> {
         is_rustc_tool: bool,
     ) -> CargoResult<ProcessBuilder> {
         let mut search_path = Vec::new();
+        // this divides our path based on CompileKind: host vs target.
```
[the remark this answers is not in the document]
tompscanlan: comment updated

src/cargo/core/compiler/compilation.rs
tompscanlan: I'll defer to the team. I'm just carrying the change forward from #10469.

PR #12711, opened 2023-09-20
refactor: move cached crates.io SourceID to config module

weihanglo:

`SourceId` should know nothing about `config` caching.

This factors out `Config` from some core types like `SourceId` step-by-step.

Shouldn't have any behavior change.

### DIFF
--- a/src/cargo/core/source_id.rs
+++ b/src/cargo/core/source_id.rs
@@ -253,11 +253,7 @@ impl SourceId {
     /// This is the main cargo registry by default, but it can be overridden in
     /// a `.cargo/config.toml`.
     pub fn crates_io(config: &Config) -> CargoResult<SourceId> {
-        config.crates_io_source_id(|| {
-            config.check_registry_index_not_set()?;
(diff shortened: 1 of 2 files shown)


epage:
> This factors out Config from some core types like SourceId step-by-step.

is this the right direction for us to be going?

I'm trying to picture how we'd take advantage of this refactor.  The main reason is splitting things up but we'll at least need some level of `Config` support for at least nightly features, if not diagnostics and other features.

weihanglo:
> we'll at least need some level of `Config` support for at least nightly features, if not diagnostics and other features.

Agree, though this kind of "extension" seems better to be provided on demand behind some feature gates or trait imports. By making types being a bit purer, it's easier to do larger scales of refactor, and beneficial for lib users to use only a subset of the cargo lib. For example, a lib using the `SourceId` and `PackageId` as opaque unique IDs, without them being pure they need to pull the entire `Config` which doesn't make sense.

The roadmap in my mind is first doing some less controversial refactors, and then propose a greater one seeking consensus. That said, if this PR appears to be wayy too much, we can just drop the change.

PR #12720, opened 2023-09-21
Update target-arch-aware crates to support mips r6 targets

chenx97:

This PR gets rid of rustix@0.37 and linux-raw-sys@0.3 and updates libc in order to support MIPS R6 targets introduced by [#112374](https://github.com/rust-lang/rust/pull/112374).

Existing tests and CI should suffice.

### Additional information

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ cargo-test-macro = { path = "crates/cargo-test-macro" }
 cargo-test-support = { path = "crates/cargo-test-support" }
 cargo-util = { version = "0.2.6", path = "crates/cargo-util" }
 cargo_metadata = "0.17.0"
-clap = "4.4.2"
+clap = "4.4.4"
 color-print = "0.3.4"
 core-foundation = { version = "0.9.3", features = ["mac_os_10_7_support"] }
 crates-io = { version = "0.39.0", path = "crates/crates-io" }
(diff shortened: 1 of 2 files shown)


weihanglo (commented):
Thanks for the contribution :)

Could you share the command you had the lockfile updated? Also, we generally update `Cargo.toml` to guarantee versions cannot be downgraded. For example, `clap = "4.4.4"` seems to be the correct version requirement for this PR.

chenx97:
> Could you share the command you had the lockfile updated?

`cargo +nightly update gix-path gix-prompt clap is-terminal libc`

chenx97:
Cargo.toml updated.

weihanglo:
Thanks. For version bumps for `gix` related packages, I lean toward updating `gix` directly, though doing so requires human reviews.

chenx97:
> Thanks. For version bumps for `gix` related packages, I lean toward updating `gix` directly, though doing so requires human reviews.

Gix now uses symbolic refs for refs/remotes/origin/HEAD and it confuses cargo. What should I do?

weihanglo:
I am fine that we move forward without updating `gix`. Just note that it may have a chance of being downgraded accidentally. Does that sound good to you, @chenx97?

chenx97:
> I am fine that we move forward without updating `gix`. Just note that it may have a chance of being downgraded accidentally. Does that sound good to you, @chenx97?

We can upgrade gix to 0.50.1.

Cargo.toml
```diff
@@ -43,8 +43,8 @@ flate2 = { version = "1.0.27", default-features = false, features = ["zlib"] }
 fwdansi = "1.1.0"
 git2 = "0.18.0"
 git2-curl = "0.19.0"
-gix = { version = "0.45.1", default-features = false, features = ["blocking-http-transport-curl", "progress-tree"] }
-gix-features-for-configuration-only = { version = "0.30.0", package = "gix-features", features = [ "parallel" ] }
+gix = { version = "0.50.1", default-features = false, features = ["blocking-http-transport-curl", "progress-tree"] }
+gix-features-for-configuration-only = { version = "0.34.0", package = "gix-features", features = [ "parallel" ] }
```
weihanglo, line 47: Should be `0.32.1`?

```suggestion
gix-features-for-configuration-only = { version = "0.32.1", package = "gix-features", features = [ "parallel" ] }
```

weihanglo (commented):
Thanks!

Could we squash/rearrange them into fewer commits, so that the git history could be a bit clearer?

Cargo.toml
chenx97: nice catch!

weihanglo (approved):
Thanks for the back-and-forth updates!

The only thing not guaranteed in `Cargo.toml` is `is-terminal`, which once 1.73.0 is released, `criterion` package can use `IsTerminal` from std. Still, it's up to maintainers of the package.
Since `criterion` is a dev-dependency, it should be less likely to be downgraded. I am going to merge this.

PR #12723, opened 2023-09-22
Add better suggestion for the unsupported silent flag

hi-rustin:

Added a better suggestion for the unsupported silent flag.

Check out the unit tests.

### Additional information

r? @weihanglo

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -331,6 +331,18 @@ pub trait CommandExt: Sized {
     }
 
     fn arg_quiet(self) -> Self {
+        let unsupported_silent_arg = {
+            let value_parser = UnknownArgumentValueParser::suggest_arg("--quiet");
+            flag("silent", "")
+                .short('s')
+                .value_parser(value_parser)
+                .hide(true)
+        };
+        self._arg(flag("quiet", "Do not print cargo log messages").short('q'))
+            ._arg(unsupported_silent_arg)
+    }
+
+    fn arg_quiet_without_unknown_silent_arg_tip(self) -> Self {
         self._arg(flag("quiet", "Do not print cargo log messages").short('q'))
     }
 
--- a/src/bin/cargo/commands/vendor.rs
+++ b/src/bin/cargo/commands/vendor.rs
@@ -36,7 +36,7 @@ pub fn cli() -> Command {
         .arg(unsupported("relative-path"))
         .arg(unsupported("only-git-deps"))
         .arg(unsupported("disallow-duplicates"))
-        .arg_quiet()
+        .arg_quiet_without_unknown_silent_arg_tip()
         .arg_manifest_path()
         .after_help(color_print::cstr!(
             "Run `<cyan,bold>cargo help vendor</>` for more detailed information.\n"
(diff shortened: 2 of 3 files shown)


src/cargo/util/command_prelude.rs
```diff
@@ -331,7 +331,16 @@ pub trait CommandExt: Sized {
     }
 
     fn arg_quiet(self) -> Self {
+        let unsupported_silent_arg = || {
```
weihanglo, line 334: Why do we need a closure here?

src/cargo/util/command_prelude.rs
```diff
@@ -331,7 +331,16 @@ pub trait CommandExt: Sized {
...
+            let msg = "there is no `--silent` flag, use `--quiet` to suppress output";
+            let value_parser = UnknownArgumentValueParser::suggest(msg);
```
weihanglo, line 336: I feel like the default message is sufficient. We probably don't need/want to customize everything every time.

`tip: a similar argument exists: '--quiet'`

```suggestion
            let value_parser = UnknownArgumentValueParser::suggest_arg("--quiet");
```

epage: I'm also hoping we can make the rust-native completion engine complete `suggest_arg` flags with what should be used instead.

src/cargo/util/command_prelude.rs
hi-rustin: I want to add the `quiet` flag first, but I guess it doesn't matter.

src/cargo/util/command_prelude.rs
hi-rustin: Good point!

src/cargo/util/command_prelude.rs
```diff
@@ -331,7 +331,15 @@ pub trait CommandExt: Sized {
     }
 
     fn arg_quiet(self) -> Self {
+        let unsupported_silent_arg = {
+            let value_parser = UnknownArgumentValueParser::suggest_arg("--quiet");
+            flag("silent", "")
+                .short('s')
```
weihanglo, line 337: `cargo vendor` has a `-s` shorthand for `--sync`, which leads to a conflict.

```
thread 'main' panicked at /home/weihanglo/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.4.2/src/builder/debug_asserts.rs:112:1
7:
Command vendor: Short option names must be unique for each argument, but '-s' is in use by both 'tomls' and 'silent'
```

so maybe do this for `cargo vendor`?

```rust
    fn arg_quiet<const suggestion: bool = true>(self) -> Self {
        // ...
        let cmd = self._arg(flag("quiet", "Do not print cargo log messages").short('q'));
        if suggestion {
            cmd._arg(unsupported_silent_arg);
        }
        cmd
    }

hi-rustin: Yes. There is no good way to do it. What about we pass a normal parameter to it?

weihanglo: Sounds good to me!

hi-rustin: I added `arg_quiet_without_unknown_silent_arg_tip` to make it more explicit. Using a bool parameter will have too big effect on other commands.

weihanglo (approved):

weihanglo:
Looks nice. Thanks!

@bors r+

Issue #12639, opened 2023-09-07
registry_auth::token_not_logged is failing randomly


ehuss:
The test `registry_auth::token_not_logged` has been failing in CI randomly, and I don't immediately have thoughts why. 

* https://github.com/rust-lang/cargo/pull/12624#issuecomment-1707712539
* https://github.com/rust-lang/cargo/pull/12575#issuecomment-1697926451
* https://github.com/rust-lang/cargo/pull/12317#issuecomment-1611319929
* https://github.com/rust-lang/cargo/pull/12539#issuecomment-1688209161
* https://github.com/rust-lang/cargo/pull/12632#issuecomment-1709381422

```
---- registry_auth::token_not_logged stdout ----
req: Request {
    url: Url {
        scheme: "http",
        cannot_be_a_base: false,
        username: "",
        password: None,
        host: Some(
            Ipv4(
                127.0.0.1,
            ),
        ),
        port: Some(
            37477,
        ),
        path: "/index/config.json",
        query: None,
        fragment: None,
    },
    method: "get",
    authorization: None,
    if_modified_since: None,
    if_none_match: None,
}
req: Request {
    url: Url {
        scheme: "http",
        cannot_be_a_base: false,
        username: "",
        password: None,
        host: Some(
            Ipv4(
                127.0.0.1,
            ),
        ),
        port: Some(
            37477,
        ),
        path: "/index/config.json",
        query: None,
        fragment: None,
    },
    method: "get",
    authorization: Some(
        "a-unique_token",
    ),
    if_modified_since: None,
    if_none_match: None,
}
req: Request {
    url: Url {
        scheme: "http",
        cannot_be_a_base: false,
        username: "",
        password: None,
        host: Some(
            Ipv4(
                127.0.0.1,
            ),
        ),
        port: Some(
            37477,
        ),
        path: "/index/config.json",
        query: None,
        fragment: None,
    },
    method: "get",
    authorization: Some(
        "a-unique_token",
    ),
    if_modified_since: None,
    if_none_match: None,
}
req: Request {
    url: Url {
        scheme: "http",
        cannot_be_a_base: false,
        username: "",
        password: None,
        host: Some(
            Ipv4(
                127.0.0.1,
            ),
        ),
        port: Some(
            37477,
        ),
        path: "/index/3/b/bar",
        query: None,
        fragment: None,
    },
    method: "get",
    authorization: Some(
        "a-unique_token",
    ),
    if_modified_since: None,
[...]

PR #12726, opened 2023-09-22
Fix spurious errors with networking tests.

ehuss:
This fixes an issue where some networking tests could behave erratically. In particular, the `registry_auth::token_not_logged` has been failing somewhat often (see https://github.com/rust-lang/cargo/issues/12639). The issue is that curl can behave inconsistently based on whether or not it immediately detects that the connection has closed or not, which is not done consistently. HTTP 1.1 defaults to `Connection: open`, so this mini HTTP server was essentially not standards compliant. `Connection: close` tells curl to expect the connection to close ([ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection)).


arlosi:
Thanks! Glad to see this resolved.

@bors r+

PR #12727, opened 2023-09-22
Buffer console status messages.

ehuss:
This adds buffering to some of the console output. This can help with interleaved output, particularly with things like the log output interleaving with status messages. This fixes that interleaving by atomically writing the entire status message, which in turn, helps fix some spurious errors such as https://github.com/rust-lang/cargo/issues/12639.

I'm uncertain what the performance impact of this change might have. It might improve performance, since there should be a lot fewer syscalls, and I suspect terminals will be able to handle it more efficiently (and particularly across a network connection). However, I don't know if that will have a noticeable impact.

Only the "status" messages are buffered. Everything else is still unbuffered as before.

### DIFF
--- a/src/cargo/core/shell.rs
+++ b/src/cargo/core/shell.rs
@@ -4,7 +4,7 @@ use std::io::IsTerminal;
 
 use anstyle::Style;
 use anstyle_termcolor::to_termcolor_spec;
-use termcolor::{self, StandardStream, WriteColor};
+use termcolor::{self, BufferWriter, StandardStream, WriteColor};
 
 use crate::util::errors::CargoResult;
 use crate::util::style::*;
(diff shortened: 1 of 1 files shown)


src/cargo/core/shell.rs
```diff
@@ -81,9 +81,15 @@ enum ShellOut {
     /// A plain write object without color support
     Write(Box<dyn Write>),
     /// Color-enabled stdio, with information on whether color should be used
+    ///
+    /// The separate buffered fields are used for buffered writing to the
+    /// corresponding stream. The non-buffered fields should be used when you
+    /// do not want content to be buffered.
     Stream {
         stdout: StandardStream,
+        buffered_stdout: BufferWriter,
```
weihanglo, line 90: Is it possible for us to use [`termcolor::BufferedStandardStream`](https://docs.rs/termcolor/latest/termcolor/struct.BufferedStandardStream.html) here?

ehuss: I don't think so, at least not without auditing and modifying all users of the streams. The way BufferedStandardStream works is that all writes to it go to a buffer, and the buffer isn't flushed unless there is a manual call to flush, or the `Shell` gets dropped. I would like to ensure that all output is sent immediately in case cargo itself gets terminated (which would cause the output to be otherwise lost if it was buffered). I'm not confident we can have the necessary discipline to make sure things get flushed in all the right places (especially where callers access the stream directly like with `out()` and `err()`). I didn't see a way to get unbuffered access to a `BufferedStandardStream`. (In theory, we could write a wrapper that implements `Write` and always flushes, but that seems like too much complexity.)

weihanglo (approved):

PR #12731, opened 2023-09-24
upgrade gitoxide to v0.54

Byron:
This reduces the binary size and fixes an exploitable bug that could allow
code execution by injection arguments into hostnames of ssh URLs.

##### Binary Sizes (Release)

* `master`: 27930520 
* this branch: 27869304 - a whopping 61216B less. I assume this will get worse again once more functionality will be used in future PRs.

##### Possible Vulnerability

In versions prior to v0.54, running the following would cause the calculator app to be started on MacOS:

```
❯ gix clone 'ssh://-oProxyCommand=open$IFS-aCalculator/foo'
```

Now it prints `Error: Host name '-oProxyCommand=open-aCalculator' could be mistaken for a command-line argument`.

Given the nature of builds with `cargo` and the availability of build scripts, I think `cargo` isn't prone to this issue. However, I thought it was good to upgrade anyway.

Please note that a CVE doesn't exist yet, but I will check with Rustsec on how to proceed with this.
CC @Shnatsel

### Tasks

* [x] fix tests - the bug was introduced in https://github.com/Byron/gitoxide/commit/74ce8639e88db5107691e9279df2bbfd38d26de3 and it leads the local symref  `refs/remotes/origin/HEAD` to point to a non-existing branch.
    - This is a feature, but one with the shortcoming that it's allowed to point to a none-existing ref, and that ref is not automatically created, yet,  and it's not covered by a refspec. Previously it would just let the symbolic ref point to the peeled object that is known.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,18 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
-name = "ahash"
-version = "0.8.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
-dependencies = [
- "cfg-if",
- "getrandom",
- "once_cell",
- "version_check",
-]
-
-[[package]]
 name = "aho-corasick"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -234,6 +222,12 @@ name = "bytesize"
 version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a3e368af43e418a04d52505cf3dbc23dda4e3407ae2fa99fd0e4f308ce546acc"
+
+[[package]]
+name = "byteyarn"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a7534301c0ea17abb4db06d75efc7b4b0fa360fce8e175a4330d721c71c942ff"
 
 [[package]]
 name = "camino"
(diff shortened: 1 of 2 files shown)


Shnatsel:
At a glance it does deserve a security advisory, and I would be glad to help draft a RUSTSEC one. I believe something that clones untrusted URLs can end up with arbitrary code execution, is that correct?

Byron:
Yes, that's correct. Here is more details about how this can be abused: https://secure.phabricator.com/T12961 .

I also sent an email to Rust-Sec (security@rust-lang) directly BTW, in case it helps to avoid duplication of effort, otherwise your help would definitely be appreciated.

Shnatsel:
If possible, a point release for the previous series with the fix would also be great, so that affected API users could simply `cargo update` to fix the issue instead of requiring them to adapt to the API changes.

Byron:
> If possible, a point release for the previous series with the fix would also be great, so that affected API users could simply `cargo update` to fix the issue instead of requiring them to adapt to the API changes.

I think there is no API change, at least not for most, that's the way `gix` is being released that causes breaking changes upstream to indicate breaking changes downstream (within `gix-*`) for safety as it conservatively assumes these changes are exposed in some way. `gitoxide` users know that, and I'd hope a CVE helps to inform all of them.

In practice, I think 'only' `cargo-audit` and `tame-index` have a chance to run into this issue.

Shnatsel:
The presence of a point release defines how many packages will need to release new versions with the fix, and how many of these packages then need to be picked up by Linux distributions, etc. to get the issue fixed.

Is it possible to take the 0.53.1 release, apply *only* the patch for this one issue, and release it as 0.53.2? That would be ideal.

Byron:
Technically, `gix-transport` is the crate that saw the fix, but unfortunately it is legitimately breaking as it changes an Error struct to an enum. In theory, I could release that as patch probably rightfully assuming that nobody will be broken by it as nobody matched on that particular error.

If this is what you'd suggest doing, and I support that choice, I can make it happen and all users of `gix` 0.53 will get the upgrade more easily.

Shnatsel:
Could you release a minimum fix that does not properly report errors, but is semver-compatible? I believe that is warranted for a security patch. It is fine to leave proper error reporting to 0.54.

Byron:
Ok, a new version of `gix-transport`, v0.36.1, has just been released with a semver compatible backport. This will fix `gix` v0.53 as well.

weihanglo (approved):
This looks pretty good to me. Thank you all for taking care of security!

PR #12741, opened 2023-09-26
chore(ci): Update Renovate schema

epage:
We have `configMigration` enabled, so in theory a PR should show up at any time, but it will drop comments because they don't support format-preserving edits.

I did this based on clap-rs/clap#5142


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #12744, opened 2023-09-26
fix bug: corruption when cargo killed while writing

### LINKED ISSUE #11386: `cargo add` and `ctrl + c`

tompscanlan:

fix  #11386

Added unit test showing basic equivalency to existing `write(path, content)`. Full test suite should exercise write.

### Additional information

This uses tempfile::persist which was an existing dependency. atomicwrites crate, an alternative option for this fix, indicates `tempfile::persist` is the same thing.  Since we already use tempfile as a dep, I stuck with that.


ehuss:
@epage or @weihanglo, can either of you take assignment on this?

epage:
r? epage

tompscanlan:
If it helps. I can write a test that makes `write` fail and shows no failure under `write_atomic`.  Just let me know what you'd like to see.

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1689,14 +1689,14 @@ fn write_fingerprint(loc: &Path, fingerprint: &Fingerprint) -> CargoResult<()> {
...
-    paths::write(loc, util::to_hex(hash).as_bytes())?;
+    paths::write_atomic(loc, util::to_hex(hash).as_bytes())?;
 
     let json = serde_json::to_string(fingerprint).unwrap();
     if cfg!(debug_assertions) {
         let f: Fingerprint = serde_json::from_str(&json).unwrap();
         assert_eq!(f.hash_u64(), hash);
     }
-    paths::write(&loc.with_extension("json"), json.as_bytes())?;
+    paths::write_atomic(&loc.with_extension("json"), json.as_bytes())?;
```
marcospb19, line 1699: (Laying out my remaining doubt from the last PR.)

Citing @epage's [comment](https://github.com/rust-lang/cargo/pull/12362#discussion_r1269787624).
> The bug is "wiped my file clean on ctrl-c" (#11386). I assume these "atomic writes" would instead be "I've got a random temp file on ctrl-c". Only slightly better. If that is the case, then I would expect this to only be applied to the cases where the empty file causes problems.

I don't know how all `cargo` components work, but I imagine fingerprints is a place where empty files are treated correctly? If so, should these be reverted?

tompscanlan: With the existing write, I will occasionally get zero length Cargo.toml when being destructive. With larger config, I might expect to see corrupt nonzero writes, but I have not seen this.

I consider a command that unexpectedly moves my project from a working state to non-existent to have a bug.

This change should at least keep our output config in one of 2 states, the same as before the change, or the new config content.

I think we are in danger, though small, of this bug in any place we write Cargo.toml out.  I'm also not familiar with fingerprints code, but can't imagine leaving a zero length config accidentally is not a bug.

I think we'd want to keep this change.

What am I missing?



Sorry this is.from a phone, so edits are hard.

tompscanlan:
To make discussion more productive, I added a test that is failing around `write()` when we use `cargo remove dep`.  The same results should apply to any command using `write()`.   Replacing the `write()` with `write_atomic()` makes this failing test pass.

I recommend we switch to the `write_atomic()`, but if occasionally zeroing out a cargo.toml is acceptable, then we should close this issue as a non-bug.

src/cargo/core/compiler/fingerprint/mod.rs
marcospb19: > What am I missing?

I believe you're missing this: the changes in question here does not write a manifest file, but a fingerprint.

(same for the change in the `src/cargo/core/compiler/build_context/target_info.rs` file)

crates/cargo-util/src/paths.rs
```diff
@@ -176,10 +176,28 @@ pub fn read_bytes(path: &Path) -> Result<Vec<u8>> {
 /// Equivalent to [`std::fs::write`] with better error messages.
 pub fn write<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()> {
     let path = path.as_ref();
+    // println!("write: {}", path.display());
```
epage, line 179: Please remove this dead code

crates/cargo-util/src/paths.rs
```diff
@@ -176,10 +176,28 @@ pub fn read_bytes(path: &Path) -> Result<Vec<u8>> {
...
 
+/// Writes a file to disk atomically.
+///
+/// write_atomic uses tempfile::persist to accomplish atomic writes.
+pub fn write_atomic<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()> {
+    let path = path.as_ref();
+    // println!("write_atomic: {}", path.display());
+    // if contents.as_ref().eq(b"") {
+    // println!("write_atomic: contents started empty");
+    // }
```
epage, line 192: Please remove this dead code

crates/cargo-util/src/paths.rs
```diff
@@ -193,9 +211,7 @@ pub fn write_if_changed<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) ->
         let mut orig = Vec::new();
         f.read_to_end(&mut orig)?;
         if orig != contents {
-            f.set_len(0)?;
-            f.seek(io::SeekFrom::Start(0))?;
-            f.write_all(contents)?;
```
epage, line 198: Why is this changed?

tests/testsuite/death.rs
```diff
@@ -87,6 +87,93 @@ fn ctrl_c_kills_everyone() {
...
+    version = "0.0.1"
+    authors = []
+    build = "build.rs"
+    "#;
+
+    // This test depends on killing the cargo process at the right time to cause a failed write.
+    // Note that we're iterating and using the index as time in ms to sleep before killing the cargo process.
+    // Max test time ~10s testing timings from 1-100ms between start and kill.
+    // If it is working correctly, we never fail, but can't hang out here all day...
+    for i in 1..100 {
```
epage, line 116: I'm a bit concerned about having a test that runs for 10s

epage:
Would you be up for cleaning up the commits for how you'd want them communicating out this change?

tests/testsuite/death.rs
tompscanlan: I also do not like.  It's a timing issue, and I'm not sure how else to get the timing right.  Open to suggestions to tighten the loop.  

Is there a regression test suite, or something that run less frequently, but can handle some longer tests?

crates/cargo-util/src/paths.rs
tompscanlan: Poor self-control :) Will back out.

tompscanlan:
Thanks for the feedback, and yes, I'll  be cleaning and adding one more test for the "add" command.

crates/cargo-util/src/paths.rs
epage: I mean it might be reasonable (in its own commit); I'd just want to see some investigation as to why its written the way it is.  People don't write code like that out of nowhere.

tests/testsuite/death.rs
epage: I'm not aware of any stress test suite we can run this with.  I think I'd chalk it up with other ux issues and say "we've tested it as part of the PR".  The pattern is well understood so I don't think we're too concerned on that front.

src/cargo/core/compiler/fingerprint/mod.rs
tompscanlan: Thanks for hanging in there and helping me understand. 

Ok, turned it around. Only changing the `write()`s in the path for command "cargo add" and "cargo remove".  If other sections need to adopt this change, they can do so later.  

Thanks for helping me understand.

tests/testsuite/death.rs
tompscanlan: Do these make sense to keep at all?  I reduced to 3 attempts each, for less than half a second total. If not, I'll trim off entirely next.

tests/testsuite/death.rs
```diff
@@ -88,32 +88,111 @@ fn ctrl_c_kills_everyone() {
...
+        // kill the process unexpectedly
+        let kill_result = child.kill();
+        assert!(kill_result.is_ok());
+
+        // collect the process
+        let wait_result = child.wait();
+        assert!(wait_result.is_ok());
+
+        println!("cargo out {}: {:?}", i, cargo.output().unwrap());
+        // assert!(false);
```
epage, line 138: Dead code?

tests/testsuite/death.rs
```diff
@@ -88,32 +88,111 @@ fn ctrl_c_kills_everyone() {
...
+        // nap for a slightly random amount of time.
+        thread::sleep(time::Duration::from_millis(i));
+
+        // kill the process unexpectedly
+        let kill_result = child.kill();
+        assert!(kill_result.is_ok());
+
+        // collect the process
+        let wait_result = child.wait();
+        assert!(wait_result.is_ok());
```
epage, line 135: nit: these comments seep to be repeating exactly what the code says

tests/testsuite/death.rs
```diff
@@ -88,32 +88,111 @@ fn ctrl_c_kills_everyone() {
...
+my-package = "0.1.1"
+"#;
+    let without_dependency = r#"
+[package]
+name = "foo"
+version = "0.0.1"
+authors = []
+"#;
+
+   for i in [30, 60, 90] {
```
epage, line 111: nit: `i` could be more specific like `sleep_time_ms`

tests/testsuite/death.rs
```diff
@@ -88,32 +88,111 @@ fn ctrl_c_kills_everyone() {
...
+build = "build.rs"
+"#;
 
     // This test depends on killing the cargo process at the right time to cause a failed write.
     // Note that we're iterating and using the index as time in ms to sleep before killing the cargo process.
-    // Max test time ~10s testing timings from 1-100ms between start and kill.
     // If it is working correctly, we never fail, but can't hang out here all day...
-    for i in 1..100 {
+    // So we'll just run it a few times and hope for the best.
+    for i in [30, 60, 90] {
```
epage, line 195: ditto about `i`s ame

tests/testsuite/death.rs
epage: Unsure but the current sleep times might be sufficient.  We'll need to see how we're doing on test time at some point and work to improve it.  If we had proper fixture support, each sleep time could be run in parallel...

epage:
Thanks for continuing on this and sorry for the initial delay before I got to reviewing this!

tests/testsuite/death.rs
tompscanlan: I'm still new to rust and it helps me sometimes to be able to squint my eyes and just see the comments.  Agree it's redundant here. Tidy.

tompscanlan:
No worries. I know ya'll are swamped. Thanks for steering.  I may be away from keyboard for several days... will return to finish.

tompscanlan:
I think I'm done unless there's feedback on this change.  Let me know @epage , thanks!

epage:
@tompscanlan sounded like you planned to clean up the commits.  Were you still planning to do that?

tompscanlan:
I think I'm done @epage, unless we want to trim of the tests. I might take a few days to respond.

epage:
Thanks!

@bors r+

Issue #841, opened 2014-11-11
Git tags when packaging


steveklabnik:
Bundler does some cool git stuff when publishing a package[1]. Conceptually, with `cargo` commands instead, and with error handling and rollback elided, it does this:

``` bash
$ cargo build
$ git diff --exit-code == 0
$ git diff-index --quiet --cached HEAD == 0
$ git tag -a -m \"Version #{version}\" #{version_tag} 
$ git push
$ git push --tags
$ cargo publish
```

(`version` is like `1.2.3` and `version_tag` is like `v1.2.3`)

Having the tags automatically created at the same commit as was released, plus ensuring that what's on github/whatever is the same as what's on crates.io would be pretty great.

1: https://github.com/bundler/bundler/blob/master/lib/bundler/gem_helper.rb#L47-L50

shadowmint:
I don't have much to add to this except 1) yes please, and 2) version_tag should not be prefixed with v; this is a left over from pre-2.0 of semver and no longer part of the spec.



Also worth noting that this may require some

bluss:
`packagename-VERSION` may be a good tag name pattern. It works for those that have many cargo packages per repository (which is quite common, regex, quickcheck, rand, etc, many do).

sbeckeriv:
@steveklabnik how important is the push tag step? @bluss doesnt think it should push.

I have auto tagging configurable via cargo config. Its not clear to me yet on how to push tags via libgit2-rs. Also should the config live in the cargo toml so its project wide? I thought Cargo.toml was loaded in the config object but my test say otherwise. 

https://github.com/rust-lang/cargo/compare/master...sbeckeriv:git-auto-tag-841?expand=1

let me know your thoughts/concerns/comments.
thanks
Becker

Nemo157:
Just thought I'd mention that I use an inverse workflow, I tag (with PGP signature) and use that as a trigger that the CI service should publish that commit to crates.io. I feel it comes closer to ensuring reproducible builds, the exact build steps taken (modulo the setup of the CI container) are specified in the `.travis.yml`/`.gitlab-ci.yml`/... file with the code. Not sure how common this might be compared to publishing from your local dev machine though.

If this auto-tagging is configurable that would ensure this workflow still works, I'm not sure if there's any other way that this could interact better/worse with it.

matklad:
Automatic tagging and pushing to the repo do raise some design questions.

A possibly simpler solution is to check that a tag exists before publishing. This should be done over here:

[...]

PR #12745, opened 2023-09-26
docs(ref): Establish publish best practices

epage:
Inspired by #841 and conversations at RustConf.

At RustConf, someone was talking to me about the lack or inconsistent use of best practices within the crate ecosystem.
We have a challenge of needing to help users get up and going who can't research every last decision while not over-encouraging the status quo, leading to stagnation.

The list of tools included was based on a mixture of
- Download counts
- Word of mouth exposure
- Unique aspects covered
- Maturity

I do have a bias in this conversation as the maintainer of one of these tools which is why I decided to pick an ordering without judgement. I picked alphabetical ordering before sorting and realizing this puts my own crate first.

### DIFF
--- a/src/doc/src/reference/publishing.md
+++ b/src/doc/src/reference/publishing.md
@@ -136,6 +136,19 @@ In order to release a new version, change [the `version` value](manifest.md#the-
 Keep in mind [the SemVer rules](semver.md) which provide guidelines on what is a compatible change.
 Then run [`cargo publish`] as described above to upload the new version.
 
+> **Recommendation:** Consider the full release process and automate what you can.
+>
+> Each version should include:
+> - A changelog entry, preferrably [manually curated](https://keepachangelog.com/en/1.0.0/) though a generated one is better than nothing
+> - A [git tag](https://git-scm.com/book/en/v2/Git-Basics-Tagging) pointing to the published commit
+>
+> Examples of third-party tools that are representative of different workflows include (in alphabetical order):
+> - [cargo-release](https://crates.io/crates/cargo-release)
+> - [cargo-smart-release](https://crates.io/crates/cargo-smart-release)
+> - [release-plz](https://crates.io/crates/release-plz)
+>
+> For more, see [crates.io](https://crates.io/search?q=cargo%20release).
+
 ## Managing a crates.io-based crate
 
 Management of crates is primarily done through the command line `cargo` tool


hi-rustin (approved):
Looks good to me. Thanks!

src/doc/src/reference/publishing.md
```diff
@@ -136,6 +136,19 @@ In order to release a new version, change [the `version` value](manifest.md#the-
...
+> Each version should include:
+> - A changelog entry, preferrably [manually curated](https://keepachangelog.com/en/1.0.0/) though a generated one is better than nothing
+> - A [git tag](https://git-scm.com/book/en/v2/Git-Basics-Tagging) pointing to the published commit
+>
+> Examples of third-party tools that are representative of different workflows include (in alphabetical order):
+> - [cargo-release](https://crates.io/crates/cargo-release)
+> - [cargo-smart-release](https://crates.io/crates/cargo-smart-release)
+> - [release-plz](https://crates.io/crates/release-plz)
+>
+> For more, see [crates.io](https://crates.io/search?q=cargo%20release)
```
weihanglo, line 150:
```suggestion
> For more, see [crates.io](https://crates.io/search?q=cargo%20release).
```

weihanglo: I am not sure if "cargo release" is the correct keyword, but yet I can't see any better alternative 😆.

weihanglo (commented):
What's the criteria if somebody wants to put their release tool in the list? Should we make it clear in this PR thread, or we'll discuss it case by case?

Apart from that, looks nice!

src/doc/src/reference/publishing.md
epage: Typo fixed.

Unsure how to make the search term any better. It seems to cover a lot of cases for command-line focused releases though that leaves out tools like release-plz.  The `release` keyword isn't universally used and is too vague, covering a lot of other cases.

epage:
> What's the criteria if somebody wants to put their release tool in the list? Should we make it clear in this PR thread, or we'll discuss it case by case?

I've noted in the PR why I chose the crates I did.  I think we'll need to handle it on a case-by-case basis though I expect a lot of cases to be obvious (brand-new tool that barely does anything is likely not something worth including).

weihanglo (approved):

weihanglo:
Thanks!

 @bors r+

Issue #4225, opened 2017-06-26
Add flag to allow Cargo to create a lock file that depends on yanked crates


retep998:
Implementation: https://github.com/rust-lang/cargo/pull/13333

---

Currently if you have a `Cargo.lock` that depends on yanked crates you are able to build the project just fine as the yanked crates still exist. However there is currently no way to create a `Cargo.lock` if the only versions of some dependency that satisfies version constraints are yanked, aside from creating a `Cargo.lock` by hand which is a really poor solution.

I propose a new flag that is added to Cargo commands capable of creating a `Cargo.lock` which will allow Cargo to add dependencies on yanked crates if it has no other option.

Inspired by https://users.rust-lang.org/t/crates-io-disable-yanking-crates-that-have-dependant-crates/11492

carols10cents:
I'm mostly against this because it defeats the whole purpose of yanking (which could have been done because of malicious code or a security problem). I also think this could at least be prototyped in an external tool first, and we could get a better idea of how often this is actually needed and whether this is the solution to the problem that we want to bake into cargo. 

Leaving this open for others' thoughts, though.

pietroalbini:
Having a flag like this (even perma-unstable) would be really nice for Crater: it regenerates lockfiles every run, so if a crate depends on yanked ones it can't be built anymore.

sbditto85:
Speaking of _ring 0.12.x_ i have a bin that uses _biscuit 0.0.8_ which depends on _ring 0.12.x_ ... because of that i cannot add any other non-related crates (like hex which has no deps) because that would be creating a new lock file. We are currently working on updating deps etc, but it really leaves me high and dry trying to add a quick feature we need. It would be nice to be able to include _ring 0.12.x_ in my cargo.toml file with a flag that says "yeah i know its yanked but I accept any consequences until i can upgrade"

birkenfeld:
Another valid use case that came up [on Reddit](https://old.reddit.com/r/rust/comments/ctayew/rust_crypto_developers_please_stop_the_yanking/): allowing people to do `git bisect`.

dpc:
Another use case: I am writting a fuzzer. I want to temporarily run it against a affected crate version to see if it would catch the problem.

dtolnay:
Workaround for those landing on this thread needing to do this: I commented the following line:

[...]

PR #9847, opened 2021-08-27
Distinguish lockfile version req from normal dep in resolver error message

weihanglo:
Resolves #9840 

This PR adds a new variant `OptVersionReq::Locked` as #9840 described.
The new variant represents as a locked version requirement that contains
an exact locked version and the original version requirement.

Previously we use exact version req to synthesize locked version req.
Now we make those need a truly locked to be `OptVersionReq::Locked`,
including:

- `[patch]`: previous used exact version req to emulate locked version,
  and it only need to lock dep version but not source ID, so we make
  an extra method `Dependency::lock_version` for it.
- Dependencies from lock files: No need to change the call site.

PR #12749, opened 2023-09-28
move up looking at index summary enum

Eh2406:
This builds on the work from #12643 and the discussion of the overall approach is at https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/better.20error.20messages.20for.20filtered.20versions.2E

This uses the enum added in #12643 to delay making decisions about yanked and `–– precise` versions until higher in the stack.

This is an internal re-factoring and all the tests still pass. BUT, this is also tricky interleaved code where small changes may have important impacts to semantics.

### Additional information

I will try and call out potentially breaking changes as inline comments. If any of these are controversial enough to deserve separate discussion I'm happy split it into separate PR's.

### DIFF
--- a/src/cargo/sources/registry/index.rs
+++ b/src/cargo/sources/registry/index.rs
@@ -98,7 +98,6 @@ use cargo_util::{paths, registry::make_dep_path};
 use semver::Version;
 use serde::Deserialize;
 use std::borrow::Cow;
-use std::cmp::Ordering;
 use std::collections::BTreeMap;
 use std::collections::{HashMap, HashSet};
 use std::fs;
@@ -632,11 +631,7 @@ impl<'cfg> RegistryIndex<'cfg> {
         f: &mut dyn FnMut(IndexSummary),
         online: bool,
     ) -> Poll<CargoResult<()>> {
-        let source_id = self.source_id;
-
-        let summaries = ready!(self.summaries(name, req, load))?;
-
-        let summaries = summaries
+        ready!(self.summaries(name, &req, load))?
             // First filter summaries for `--offline`. If we're online then
             // everything is a candidate, otherwise if we're offline we're only
             // going to consider candidates which are actually present on disk.
@@ -657,35 +652,8 @@ impl<'cfg> RegistryIndex<'cfg> {
             // Next filter out all yanked packages. Some yanked packages may
             // leak through if they're in a whitelist (aka if they were
             // previously in `Cargo.lock`
-            .filter(|s| !s.is_yanked() || yanked_whitelist.contains(&s.package_id()));
-
-        // Handle `cargo update --precise` here.
-        let precise = source_id.precise_registry_version(name.as_str());
-        let summaries = summaries.filter(|s| match precise {
-            Some((current, requested)) => {
-                if req.matches(current) {
-                    // Unfortunately crates.io allows versions to differ only
-                    // by build metadata. This shouldn't be allowed, but since
-                    // it is, this will honor it if requested. However, if not
-                    // specified, then ignore it.
-                    let s_vers = s.package_id().version();
-                    match (s_vers.build.is_empty(), requested.build.is_empty()) {
-                        (true, true) => s_vers == requested,
-                        (true, false) => false,
-                        (false, true) => {
-                            // Compare disregarding the metadata.
-                            s_vers.cmp_precedence(requested) == Ordering::Equal
-                        }
-                        (false, false) => s_vers == requested,
-                    }
-                } else {
-                    true
-                }
-            }
-            None => true,
-        });
-
-        summaries.for_each(f);
+            .filter(|s| !s.is_yanked() || yanked_whitelist.contains(&s.package_id()))
+            .for_each(f);
         Poll::Ready(Ok(()))
     }
 
--- a/src/cargo/sources/registry/index.rs
+++ b/src/cargo/sources/registry/index.rs
@@ -99,7 +99,7 @@ use semver::Version;
 use serde::Deserialize;
 use std::borrow::Cow;
 use std::collections::BTreeMap;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::fs;
 use std::io::ErrorKind;
 use std::path::Path;
@@ -573,8 +573,7 @@ impl<'cfg> RegistryIndex<'cfg> {
         name: InternedString,
         req: &OptVersionReq,
         load: &mut dyn RegistryData,
-        yanked_whitelist: &HashSet<PackageId>,
-        f: &mut dyn FnMut(Summary),
+        f: &mut dyn FnMut(IndexSummary),
     ) -> Poll<CargoResult<()>> {
         if self.config.offline() {
             // This should only return `Poll::Ready(Ok(()))` if there is at least 1 match.
@@ -591,31 +590,15 @@ impl<'cfg> RegistryIndex<'cfg> {
             let callback = &mut |s: IndexSummary| {
                 if !s.is_offline() {
                     called = true;
-                    f(s.into_summary());
+                    f(s);
                 }
             };
-            ready!(self.query_inner_with_online(
-                name,
-                req,
-                load,
-                yanked_whitelist,
-                callback,
-                false
-            )?);
+            ready!(self.query_inner_with_online(name, req, load, callback, false)?);
             if called {
                 return Poll::Ready(Ok(()));
             }
         }
-        self.query_inner_with_online(
-            name,
-            req,
-            load,
-            yanked_whitelist,
-            &mut |s| {
-                f(s.into_summary());
-            },
-            true,
-        )
+        self.query_inner_with_online(name, req, load, f, true)
     }
 
     /// Inner implementation of [`Self::query_inner`]. Returns the number of
@@ -627,7 +610,6 @@ impl<'cfg> RegistryIndex<'cfg> {
         name: InternedString,
         req: &OptVersionReq,
         load: &mut dyn RegistryData,
-        yanked_whitelist: &HashSet<PackageId>,
         f: &mut dyn FnMut(IndexSummary),
         online: bool,
     ) -> Poll<CargoResult<()>> {
@@ -649,10 +631,6 @@ impl<'cfg> RegistryIndex<'cfg> {
                     IndexSummary::Offline(s.as_summary().clone())
                 }
             })
-            // Next filter out all yanked packages. Some yanked packages may
-            // leak through if they're in a whitelist (aka if they were
-            // previously in `Cargo.lock`
-            .filter(|s| !s.is_yanked() || yanked_whitelist.contains(&s.package_id()))
             .for_each(f);
         Poll::Ready(Ok(()))
     }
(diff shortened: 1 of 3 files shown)


src/cargo/sources/registry/index.rs
```diff
@@ -611,9 +603,8 @@ impl<'cfg> RegistryIndex<'cfg> {
             name,
             req,
             load,
-            yanked_whitelist,
             &mut |s| {
-                f(s.into_summary());
+                f(s);
```
Eh2406, line 607: oops this can just be `f` not `&mut |s| f(s)`

src/cargo/sources/registry/mod.rs
```diff
@@ -709,26 +709,33 @@ impl<'cfg> Source for RegistrySource<'cfg> {
...
-        // If this is a precise dependency, then it came from a lock file and in
+        let mut req = dep.version_req().clone();
+
+        // Handle `cargo update --precise` here.
+        if let Some((_, requested)) = self
+            .source_id
+            .precise_registry_version(dep.package_name().as_str())
+            .filter(|(c, _)| req.matches(c))
+        {
+            req.lock_to(&requested);
```
Eh2406, line 720: `lock_to` was changed have the same semantics as `--precise`, but it's possible we should have two different kind of locked requirements.

src/cargo/sources/registry/mod.rs
```diff
@@ -709,26 +709,33 @@ impl<'cfg> Source for RegistrySource<'cfg> {
...
+        }
+
+        // If this is a locked dependency, then it came from a lock file and in
         // theory the registry is known to contain this version. If, however, we
         // come back with no summaries, then our registry may need to be
         // updated, so we fall back to performing a lazy update.
-        if kind == QueryKind::Exact && dep.source_id().precise().is_some() && !self.ops.is_updated()
-        {
+        if kind == QueryKind::Exact && req.is_locked() && !self.ops.is_updated() {
```
Eh2406, line 727: I think `req.is_locked()` and `dep.source_id().precise().is_some()` have the same meaning. The same method sets both of them, and assert_eq pass all tests. So what is the change for clarity.

src/cargo/sources/registry/mod.rs
```diff
@@ -709,26 +709,33 @@ impl<'cfg> Source for RegistrySource<'cfg> {
...
-                dep.version_req(),
-                &mut *self.ops,
-                &self.yanked_whitelist,
-                &mut |s| {
-                    if dep.matches(&s) {
+            ready!(self
+                .index
+                .query_inner(dep.package_name(), &req, &mut *self.ops, &mut |s| {
+                    if dep.matches(s.as_summary()) {
+                        // We are looking for a package from a lock file so we do not care about yank
```
Eh2406, line 734: The to individually save changes above may mean that we can now `--precise` on a version that is locked. If that is a bug, and not a feature, we can filter out locked versions here as well. If it's a feature, we should add a test.

src/cargo/sources/registry/index.rs
```diff
@@ -592,17 +591,10 @@ impl<'cfg> RegistryIndex<'cfg> {
...
-            ready!(self.query_inner_with_online(
-                name,
-                req,
-                load,
-                yanked_whitelist,
-                callback,
-                false
-            )?);
+            ready!(self.query_inner_with_online(name, req, load, callback, false)?);
             if called {
```
Eh2406, line 598: by moving filtering of yanked versions from beneath this code to above this code there may be situations in which `called` is no different. I don't think this will break anything, tests still pass, but it needs more thinking than I gave it.

src/cargo/util/semver_ext.rs
```diff
@@ -84,10 +84,17 @@ impl OptVersionReq {
             OptVersionReq::Any => true,
             OptVersionReq::Req(req) => req.matches(version),
             OptVersionReq::Locked(v, _) => {
```
Eh2406, line 86: There are three semantics used for two versions being equal.
1. constructing a `=` requirement from `A` and seeing if the requirement matches `B`. This is what the old code does, for backwards compatibility with before we had `Locked`. (cc @weihanglo re https://github.com/rust-lang/cargo/pull/9847#discussion_r697887365)
2. directly checking if 'A == B`, requiring all fields to match. The comet argues that this should be correct, because we only set this to a value of a version from the lock file which should be fully specified version. All tests pass if we change to that behavior. but tests don't always consider all the weird things in the world like versions that only different metadata.
3. The comparison used for `--precise` witch allows the metadata fields to be anything if the past in version does not have a metadata field.

In this PR I make `Locked` more strict then 1, but only to the semantics of 3. This simplifies the handling of `--precise`. but maybe conflating unrelated issues. Perhaps it would be better to have two different variants for the two different semantics.

src/cargo/util/semver_ext.rs
```diff
@@ -84,10 +84,17 @@ impl OptVersionReq {
...
+                // Generally  we can, lockfiles record an exact version
+                // which is the only one we should use on repeat builds.
+                // However, this is also used for the `--precise` field of cargo update
+                // in that context we treat a requirement that does not have build metadata as allowing any metadata.
+                version.major == v.major
+                    && version.minor == v.minor
+                    && version.patch == v.patch
+                    && version.pre == v.pre
+                    && (version.build == v.build
+                        || (!version.build.is_empty() && v.build.is_empty()))
```
epage, line 97: nit: I'd put the "self" value on the left of these operators as that is how I'm used to writing expressions like this.  Had to think about that last expression a bit to decide whether it was valid.

epage, line 97: Can't this just be `|| v.build.is_empty()` since if they are both empty, that will be caught in the `==`.

I think I'd also find that easier to read.  For me, seeing just "v.build.is_empty()`, that makes it sounds like "this is a wild card" but seeing it `&&` with something else makes it seem like something more is going on

src/cargo/util/semver_ext.rs
epage: Can/should this change be a separate commit?

Also, why do we check `metadata` at all?  I thought it isn't "significant" in versions and in various places its left off and I could see problems with making sure we deal with mixed cases correctly (like here)

src/cargo/sources/registry/mod.rs
```diff
@@ -740,15 +750,11 @@ impl<'cfg> Source for RegistrySource<'cfg> {
...
+                &req,
                 &mut *self.ops,
                 &self.yanked_whitelist,
                 &mut |s| {
-                    let matched = match kind {
-                        QueryKind::Exact => dep.matches(&s),
-                        QueryKind::Fuzzy => true,
-                    };
-                    if matched {
+                    if dep.matches(&s) || kind == QueryKind::Fuzzy {
```
epage: The downside to this refactor (separate commit?) is that we always evaluate `dep.matches(&s)` even if we don't need to.

However, unsure how much that matters as I'm assuming `Exact` is the more common case.

The other downside is if we add new cases to `kind`, the compiler won't yell at us to update this code (which is why I used explicit matches when I did that refactor)

src/cargo/sources/registry/index.rs
epage: So we query offline and see if anything came up, if it did, we move on.  We then query online.

So I guess in theory if what the offline check found was yanked, then we might move on, filter it out, and then error instead of switching to on.  Whether that can actually happen and cause problems needs a wider view of this stack than I have at this time.  From how this is all behaving, I'm assuming it will be fine?

src/cargo/sources/registry/mod.rs
epage: I wanted to double check, is that typo and you mean "`--precise` on a version that is *yanked*"?

A related issue is #4225 and I feel like allowing people to select yanked versions (with a warning) is reasonable and aligns with #12425.

Note: I don't think this completely resolves #4225.

If we do make this change, we need to make sure its clearly communicated so we don't lose track of it at least for changelog generation.  In some cases like this, I've gone as far as having the refactor do extra work to maintain existing behavior and then had a separate commit that removed that just so its easier to discover that a change was made.

Also, if we do this and don't already have a test failing because of this, we should have a commit that adds a test showing the existing behavior and then the commit with the behavior change updates the test to show the new behavior.

src/cargo/sources/registry/mod.rs
```diff
@@ -781,13 +779,9 @@ impl<'cfg> Source for RegistrySource<'cfg> {
...
-                        .query_inner(
-                            name_permutation,
-                            &req,
-                            &mut *self.ops,
-                            &self.yanked_whitelist,
-                            f,
-                        )?
+                        .query_inner(name_permutation, &req, &mut *self.ops, &mut |s| {
+                            f(s.into_summary());
+                        })?
```
epage, line 784: Is there a reason this doesn't do filtering?

src/cargo/sources/registry/mod.rs
epage: I feel I'm missing something with what this comment is referring to.  Where is `dep.source_id().precise().is_some()` that this is referring to?

Eh2406: The code before the PR.

src/cargo/sources/registry/index.rs
Eh2406: That analysis sounds correct.

> (this should likely only happen with locked packages which will be allow-listed)

I think the problematic case is that there is a downloaded version that has been yanked and is not in the allowed list. The old behavior would filter out the version because it yanked  and not in the allow list, meaning that the first pass would see no candidates and so fallback to the second pass. Whereas, the new code will see the yanked version witch is available off-line so return it is the only candidate not falling back to the second pass.

src/cargo/sources/registry/mod.rs
Eh2406: Backed out.

src/cargo/sources/registry/mod.rs
Eh2406: This is in a `kind == QueryKind::Fuzzy` so it is only being used for looking for near misses for error messages. At this point, the more results we can return the better even if there yanked or similar.

src/cargo/sources/registry/mod.rs
epage: Ah, hidden scroll bars...

src/cargo/sources/registry/mod.rs
epage: Should we put that in as a comment?

src/cargo/util/semver_ext.rs
Eh2406: I had a lot to say in response to this, and decided it needed its own PR #12772.

src/cargo/sources/registry/mod.rs
Eh2406: code updated so as to not add this interaction.

Eh2406:
Rebased and updated to have a different variant of `OptVersionReq` for the different semantics.

src/cargo/sources/registry/index.rs
epage: So it looks like we are allowing yanked in that first pass, which is fine because its a lockfile, bypassing this problem.  Do I have that right?

epage (approved):

Eh2406:
I cannot reproduce the build failures locally.

PR #12751, opened 2023-09-28
refactor: Switch from termcolor to anstream

### LINKED ISSUE #12627: Migrate from termcolor to anstream
`anstream` provides the same feature set as `termcolor` (including legacy windows support) but with a less awkward API to work with.  Instead of hand-constructing data in a special container with each step requiring setting the style, you use whatever styling library you like (e.g. color-print like in #12578) and then write to the appropriate stream, and you are done.  The stream does all the rest, allowing styling to be decoupled from where it is being rendered.

So `anstream` would replace
- `termcolor`
- `fwdansi`
- `strip-ansi-escapes`

Benefits
- Lower boilerplate way to style text while leaving `Shell` in control (see `cargo add`s use of `Shell::write_stderr`)
- Open the door for styling within messages
- Lower overhead as it only intercepts when something needs changing.  This will be especially true in switching from `fwdansi` to `anstream`
- Strip API is significantly faster than `strip-ansi-escapes`
- Centralize color schemes between `Shell` (currently using termcolor) and `clap` (using `anstyle`)

epage:

`anstream` asks the question "what if you took `fwdansi` and removed `termcolor` underneath it.  It wraps output streams, adapting ANSI escape codes to what is needed
- Pass through if its supported
- Strip if its not
- Adapt to wincon API if needed

Benefits
- Lower boilerplate: we can use `write!` with styled text rather than the back-and-forth between colors and writing that termcolor needs
- Allows richer styling as `Shell` can accept styled messages and adapt them as needed

Side effects
- We'll now respect [NO_COLOR](https://no-color.org/), [CLICOLOR_FORCE](https://bixense.com/clicolors/), and [CLICOLOR](https://github.com/jhasse/clicolors/commit/3a22aaa5e8a39f2f0a7664a58a23294baf606450)

Fixes #12627

This is broken up by commits for easier browsing.

However, as there aren't really tests for colored output, this needs hand inspection to verify

### Additional information

This allowed us to remove the need for stripping ansi escape codes completely.  Even if it didn't, it exposes its internal stripping API for reuse, saving on a dependency and being significantly faster.

### DIFF
--- a/src/cargo/core/shell.rs
+++ b/src/cargo/core/shell.rs
@@ -2,9 +2,8 @@ use std::fmt;
 use std::io::prelude::*;
 use std::io::IsTerminal;
 
+use anstream::AutoStream;
 use anstyle::Style;
-use anstyle_termcolor::to_termcolor_spec;
-use termcolor::{self, BufferWriter, StandardStream, WriteColor};
 
 use crate::util::errors::CargoResult;
 use crate::util::style::*;
@@ -81,15 +80,9 @@ enum ShellOut {
     /// A plain write object without color support
     Write(Box<dyn Write>),
     /// Color-enabled stdio, with information on whether color should be used
-    ///
-    /// The separate buffered fields are used for buffered writing to the
-    /// corresponding stream. The non-buffered fields should be used when you
-    /// do not want content to be buffered.
     Stream {
-        stdout: StandardStream,
-        buffered_stdout: BufferWriter,
-        stderr: StandardStream,
-        buffered_stderr: BufferWriter,
+        stdout: AutoStream<std::io::Stdout>,
+        stderr: AutoStream<std::io::Stderr>,
         stderr_tty: bool,
         color_choice: ColorChoice,
     },
@@ -111,15 +104,13 @@ impl Shell {
     /// output.
     pub fn new() -> Shell {
         let auto_clr = ColorChoice::CargoAuto;
-        let stdout_choice = auto_clr.to_termcolor_color_choice(Stream::Stdout);
-        let stderr_choice = auto_clr.to_termcolor_color_choice(Stream::Stderr);
+        let stdout_choice = auto_clr.to_anstream_color_choice();
+        let stderr_choice = auto_clr.to_anstream_color_choice();
         Shell {
             output: ShellOut::Stream {
-                stdout: StandardStream::stdout(stdout_choice),
-                buffered_stdout: BufferWriter::stdout(stdout_choice),
-                stderr: StandardStream::stderr(stderr_choice),
-                buffered_stderr: BufferWriter::stderr(stderr_choice),
-                color_choice: ColorChoice::CargoAuto,
+                stdout: AutoStream::new(std::io::stdout(), stdout_choice),
+                stderr: AutoStream::new(std::io::stderr(), stderr_choice),
+                color_choice: auto_clr,
                 stderr_tty: std::io::stderr().is_terminal(),
             },
             verbosity: Verbosity::Verbose,
@@ -297,9 +288,7 @@ impl Shell {
     pub fn set_color_choice(&mut self, color: Option<&str>) -> CargoResult<()> {
         if let ShellOut::Stream {
             ref mut stdout,
-            ref mut buffered_stdout,
             ref mut stderr,
-            ref mut buffered_stderr,
             ref mut color_choice,
             ..
         } = self.output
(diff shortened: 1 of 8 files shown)


src/cargo/core/shell.rs
```diff
@@ -79,17 +78,15 @@ impl fmt::Debug for Shell {
...
     Stream {
-        stdout: StandardStream,
-        buffered_stdout: BufferWriter,
-        stderr: StandardStream,
-        buffered_stderr: BufferWriter,
+        stdout: AutoStream<std::io::Stdout>,
+        stderr: AutoStream<std::io::Stderr>,
         stderr_tty: bool,
         color_choice: ColorChoice,
     },
```
obi1kenobi, line 92: Seems like you removed the buffered fields here. The doc comment might need adjusting, since it's still referencing "buffered fields."

src/cargo/core/shell.rs
```diff
@@ -79,17 +78,15 @@ impl fmt::Debug for Shell {
 /// A `Write`able object, either with or without color support
 enum ShellOut {
     /// A plain write object without color support
-    Write(Box<dyn Write>),
+    Write(AutoStream<Box<dyn Write>>),
     /// Color-enabled stdio, with information on whether color should be used
     ///
     /// The separate buffered fields are used for buffered writing to the
     /// corresponding stream. The non-buffered fields should be used when you
     /// do not want content to be buffered.
```
arlosi: Update doc comments here please.

src/cargo/core/shell.rs
```diff
@@ -523,32 +457,21 @@ impl ShellOut {
 
 impl ColorChoice {
     /// Converts our color choice to termcolor's version.
```
arlosi, line 459:
```suggestion
    /// Converts our color choice to anstream's version.
```

src/cargo/core/shell.rs
```diff
@@ -130,7 +121,7 @@ impl Shell {
     /// Creates a shell from a plain writable object, with no color, and max verbosity.
     pub fn from_write(out: Box<dyn Write>) -> Shell {
         Shell {
-            output: ShellOut::Write(out),
+            output: ShellOut::Write(AutoStream::never(out)),
```
arlosi, line 124: I'm assuming the `AutoStream::never` here is to ensure we strip out formatting? Might be worth a comment.

arlosi (approved):
Thanks! I like the cleaner code fewer dependencies.

src/cargo/core/shell.rs
```diff
@@ -425,82 +398,43 @@ impl ShellOut {
...
+        if justified {
+            write!(&mut buffer, "{style}{status:>12}{reset}")?;
+        } else {
+            write!(&mut buffer, "{style}{status}{reset}{bold}:{reset}")?;
         }
+        match message {
+            Some(message) => writeln!(buffer, " {message}")?,
+            None => write!(buffer, " ")?,
+        }
+        self.stderr().write_all(&buffer)?;
```
folkertdev, line 415: obviously I'm just some outsider, but I'm surprised by how this PR adds many small allocations. The previous implementation seems to carefully avoid making those allocations.

It seems to me that some "code pain" in cargo is worth it given the huge number of daily cargo invocations.

In this particular case using a Cursor around an array seems like a simple alternative.

```suggestion
        let mut buffer = std::io::Cursor::new([0u8; 1024]);
        if justified {
            write!(&mut buffer, "{style}{status:>12}{reset}")?;
        } else {
            write!(&mut buffer, "{style}{status}{reset}{bold}:{reset}")?;
        }
        match message {
            Some(message) => writeln!(buffer, " {message}")?,
            None => write!(buffer, " ")?,
        }
        self.stderr()
            .write_all(&buffer.get_ref()[..buffer.position() as usize])?;
```

I'm not entirely sure why writing into a separate buffer before actually writing to stderr is required here. If it turns out that separate buffer is not needed, then this could work? 

```suggestion
        let stderr = self.stderr();

        if justified {
            write!(stderr, "{style}{status:>12}{reset} ")?;
        } else {
            write!(stderr, "{style}{status}{reset}{bold}:{reset} ")?;
        }

        if let Some(message) = message {
            write!(stderr, "{message}")?;
        }
```

obviously this is not the main point of this PR, it just surprised me. This is just one example, the same pattern seems to occur in a bunch of places and the allocations can be removed in similar ways.

epage: The original buffering was added in #12727 to workaround the fact that we aren't locking the stream (unsure why [lock](https://docs.rs/termcolor/latest/termcolor/struct.StandardStream.html#method.lock) wasn't used instead).

The difference between #12727  and this is that we aren't reusing the buffer.  

I expect allocations aren't our biggest concern here
- In most cases, we are writing to the screen which has its own performance concerns
- A good number of calls to these functions will already be having to do a `format!`.  I've found the biggest difference is between 'no allocation" and "allocation" code paths.  The difference between "3 allocations" and "4 allocations" is small.

If allocations were to be an issue, our options are
- Cache the `Vec` inside of `ShellOutput`
- Extend `anstream` to allow locking without transferring ownership
  - `AutoStream::lock` transfers ownership, which doesn't work here, because (1) `AutoStream` is stateful and it adds a lot of complexity in both the implementation and the caller (the locked stream needs a `&mut` back to the original stream) (2) most of the time none of this will matter but when it doesn, I found enough cases want access to an owned stream

arlosi: Since the number of logged status messages is relatively small, I don't think there's a going to be any measurable performance impact from this allocation. Using a fixed-length stack buffer would avoid the allocation, but also uses an unnecessary amount of stack space in most cases, and limits message length.

arlosi:
Thanks!

@bors r+

Issue #10714, opened 2022-05-28
Access target triplet of artifact from proc macros


bmcpt:
### Problem

I'm writing a proc macro that generates code based on the operating system for which the library/executable is being compiled. This macro generates OS specific code for windows/linux/macos/android. Currently there is no way of finding out which operating system we are compiling for (from a proc macro).

### Proposed Solution

Some way of accessing the target triplet from a proc macro.

### Notes

`cfg(target_os="windows")` returns true on windows even if compiling for android.

epage:
A complication to exposing the target triplet is that I believe we build proc macros and build scripts once for the host, independent of which target platform is currently selected.

ehuss:
I think if the environment variable is only set while building a crate (which uses a proc-macro), then the proc-macro would only have access to it at runtime, and thus the concern of using `env!` when building the proc-macro wouldn't be a problem.

I would recommend, if possible, to generate the code for all targets, and include `#[cfg(…)]` attributes in the generated output to select the appropriate target.

I don't think this is a decision that should be driven just by the Cargo team in isolation. This may benefit from a broader consideration from other teams like the libs, lang, or compiler teams, as this relates to the experience for how proc-macro authors write their code.

An example of an alternate approach would be to have this as part of the proc-macro API (similar to [`tracked_env`](https://doc.rust-lang.org/proc_macro/tracked_env/index.html) and [`tracked_path`](https://doc.rust-lang.org/proc_macro/tracked_path/index.html)) where `rustc` itself could provide this information (like `proc_macro::target() -> Target` or whatever).  `rustc` already knows the target information.

Another consideration is if it is a good idea to key of the target triple string. My opinion is that it is a bad idea, and code should instead use the target `cfg` values (family, os, arch, vendor, etc.). Triples can change over time or have unexpected naming conventions, and custom JSON targets have arbitrary names.  I think matching on the cfg-specific rules is more precise and reliable. That is also something the `proc_macro` API could provide (like returning a struct with all this information).

Lokathor:
I would absolutely love if the `proc_macro` crate provided the info about the target to the proc-macro in a structured, programatic way, rather than checking environment variables.

ogoffart:
Cargo sets the `TARGET` env variable when running the build script.
It would be nice if that env variable was also set when invoking rustc.
That way the proc macro code can do something depending on the result of `std::env::var("TARGET")`  just like build script can already do.
Is there a reason why "TARGET" is not defined when running rustc?

PR #12752, opened 2023-09-29
Set the `TARGET` env variable when running rustc

ogoffart:

The goal is to be able to do things differently in a proc macro depending on the target platform. It is already defined for build script and it would help to have it for macros as well.

Closes: #10714

### Additional information

I understand that the discussion in #10714 asks for a more structured way to get the target. But this change is really easy to make like this and is consistent with the way to do it in build.rs  (I always get confused as to what env variable is defined where, and consistency helps here.) So I thought I'd make a PR anyway.

### DIFF
--- a/src/cargo/core/compiler/compilation.rs
+++ b/src/cargo/core/compiler/compilation.rs
@@ -183,7 +183,7 @@ impl<'cfg> Compilation<'cfg> {
             self.rustc_process.clone()
         };
 
-        let cmd = fill_rustc_tool_env(rustc, unit);
+        let cmd = fill_rustc_tool_env(rustc, unit, &self.host);
         self.fill_env(cmd, &unit.pkg, None, unit.kind, true)
     }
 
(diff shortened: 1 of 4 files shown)


src/cargo/core/compiler/compilation.rs
```diff
@@ -383,6 +383,11 @@ fn fill_rustc_tool_env(mut cmd: ProcessBuilder, unit: &Unit) -> ProcessBuilder {
         cmd.env("CARGO_BIN_NAME", name);
     }
     cmd.env("CARGO_CRATE_NAME", unit.target.crate_name());
+    let target_triple = match &unit.kind {
+        CompileKind::Host => host_triple,
+        CompileKind::Target(t) => t.short_name(),
```
weihanglo, line 388: Should we use this instead so that local target JSON files are respected?

```suggestion
        CompileKind::Target(t) => t.rustc_target(),
```

weihanglo (commented):
Thanks for the PR, though I don't think this will merge at this moment. There are more considerations in <https://github.com/rust-lang/cargo/issues/10714#issuecomment-1142655393>. Proc macros is a thing multiple teams in Rust get involved. We should perhaps consult other teams for other possibilities before making a decision.

epage:
Since #10714 is still stuck in "needs design", should this be marked as a draft or closed until the issues are worked out?

weihanglo:
I will go ahead and close this. It doesn't mean it is not worthy to solve this problem. The issue at this time needs a driver to solicit feedback and thought from teams and community.

Your effort on this is much appreciated. Let us focus on the discussion first, and then we can revisit the implementation :)

PR #12765, opened 2023-10-02
chore: Specify all of toml_edit's features

epage:
This is preventing us from being able to update toml/toml_edit independent of each other since the new versions are both breaing changes, so by updating one, we are getting the features enabled for us by the wrong version.


hi-rustin (approved):
Tested locally. Works well. Thanks!

PR #12766, opened 2023-10-03
Warn about crate name's format when creating new crate

### LINKED ISSUE #2708: Enforce crate naming rules of rustc
## Actual behavior

Creating a new crate:

``` sh
cargo new --bin 'UPPERcase_name'
```

No issue.

Compiling it:

``` sh
 cargo build
```

```
warning: crate `UPPERcase_name` should have a snake case name such as `uppercase_name`, #[warn(non_snake_case)] on by default
```
## Expected behavior

Cargo should only create crates with names that raise `rustc` warnings if forced.
## Meta

```
  rustc --version --verbose
rustc 1.8.0
binary: rustc
commit-hash: unknown
commit-date: unknown
release: 1.8.0
```

```
  cargo --version
cargo 0.10.0 (10ddd7d 2016-04-08)
```

shnaramn:
Warns about a crate's name during creation (`crate new ...`) if it doesn't follow the preferred snake_case format.

Fixes #2708 

The warning message uses the language mentioned in [RFC 430](https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md#general-naming-conventions).

Verified existing tests succeeded with updates.  Added new tests to verify fix.

### Additional information
The link to [API naming guidelines](https://rust-lang.github.io/api-guidelines/naming.html) was not used since it still stays `unclear` for naming convention for crates.

The fix reuses [the method that checks for snake_case when building](https://github.com/rust-lang/rust/blob/9136560d32440cc7cc0ab78353054c90648c070c/compiler/rustc_lint/src/nonstandard_style.rs#L237).


sanmai-NL:
The API naming guidelines are not referenced but I think that is rightly so, since it has a lower status than an RFC.

src/cargo/ops/cargo_new.rs
```diff
@@ -258,10 +258,47 @@ fn check_name(
...
+            "the name `{}` does not follow the preferred snake case format\n\
+            See https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md#General-naming-conventions",
+            name
+        ))?;
+    }
 
     Ok(())
 }
 
+fn to_snake_case(mut str: &str) -> String {
```
sanmai-NL, line 272: Why not lowercase the whole string at once? It seems you assume significance of the upper case characters but is that worthwhile for this functionality? https://devdocs.io/rust/std/primitive.str#method.to_lowercase

sanmai-NL (changes requested):

ehuss:
Just some drive-by comments:

I'm reluctant to issue this warning on a binary crate, since that affects the executable name. I don't see a reason why that should be restricted, and I think it is probably wrong that rustc issues a warning for it.

This doesn't seem to handle `-` from what I can tell.

The rules that rustc uses for camel-case is fairly complicated. See [`to_camel_case`](https://github.com/rust-lang/rust/blob/eb0f3ed59c6508a37c6598bc9762987f053993a7/compiler/rustc_lint/src/nonstandard_style.rs#L98-L140). It is also kinda broken (see https://github.com/rust-lang/rust/issues/116336). I'm reluctant to try to duplicate what rustc is trying to do, since keeping them in sync will be challenging and cause confusion when they are out of sync. I'd like to be very careful of whether or not this is something that cargo really needs to do, since a warning will be evident on the very first build, which I think is a minor delay compared to immediately with `cargo new`. I'd like to hear from other cargo team members about this.

src/cargo/ops/cargo_new.rs
shnaramn: Wanted to be in sync with what the compiler did. But I see your point.  Will update.

tests/testsuite/new.rs
```diff
@@ -501,6 +503,19 @@ or change the name in Cargo.toml with:
         .run();
 }
 
+#[cargo_test]
+fn non_snake_case_name() {
+    cargo_process("new UPPERcase_name")
+        .with_stderr(
+            "\
```
epage, line 510: We should have a test that to ensure there isn't a warning with kebab case

src/cargo/ops/cargo_new.rs
```diff
@@ -258,6 +258,13 @@ fn check_name(
             name
         ))?;
     }
+    if name != name.to_lowercase() {
+        shell.warn(format!(
+            "the name `{}` does not follow the preferred snake case format\n\
+            See https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md#General-naming-conventions",
+            name
+        ))?;
+    }
```
epage, line 267: While this new style is simpler and allows for kebab case package names, the message makes it sound like a kebab case package name wouldn't be allowed.

src/cargo/ops/cargo_new.rs
```diff
@@ -258,6 +258,13 @@ fn check_name(
             name
         ))?;
     }
+    if name != name.to_lowercase() {
+        shell.warn(format!(
+            "the name `{}` does not follow the preferred snake case format\n\
+            See https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md#General-naming-conventions",
+            name
```
epage, line 265: side note: with #12235, I could see us having a warning for mixing `-` and `_`

src/cargo/ops/cargo_new.rs
```diff
@@ -258,6 +258,13 @@ fn check_name(
             name
         ))?;
     }
+    if name != name.to_lowercase() {
+        shell.warn(format!(
+            "the name `{}` does not follow the preferred snake case or kebab case format\n\
+            See https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md#General-naming-conventions",
```
epage, line 264: 1. Should we just say "lower case" since that is simpler and the only thing decided at this time?
2. Mirroring the message from rustc might be a good start "package `{name}` should have a lowercase name such as `{lower_name}`"
3. We likely shouldn't be linking to that RFC.  It refers to "crate" and not 'package".  Sometimes those get mixed up but the [current naming guidelines, which also conflate crate and package, say there is no consesus](https://rust-lang.github.io/api-guidelines/naming.html)

shnaramn: Wanted to keep the warning message similar to other warnings messages in the file.

shnaramn:
Friendly ping.  Do the latest changes look good?

src/cargo/ops/cargo_new.rs
```diff
@@ -258,6 +258,13 @@ fn check_name(
             name
         ))?;
     }
+    let name_in_lowercase = name.to_lowercase();
+    if name != name_in_lowercase {
+        shell.warn(format!(
+            "package `{}` should have a lowercase name such as `{}`",
```
epage, line 264: In the last change, the wording became strong, going from talking of "preferred" to "should"

src/cargo/ops/cargo_new.rs
```diff
@@ -258,6 +258,13 @@ fn check_name(
...
+            name, name_in_lowercase
```
epage, line 265: nit: saying this just because I already made another comment but feel free to reference these inside the `format!` string (`format!("package '{name}'...`)

shnaramn: There are few more places in the method where this applies.  I'll send a separate PR addressing these.

src/cargo/ops/cargo_new.rs
shnaramn: Is this a better option:

"package {name} could have a lowercase name (preferred) such as {lower_name}"

src/cargo/ops/cargo_new.rs
epage: Yup, not accountable for fixing everything but we can improve what new code we write :)

src/cargo/ops/cargo_new.rs
epage: "could" feels bit weird

For brainstorming ideas, here are some options:
```rust
// soften with "generally" and focusing on the principle rather than this specific package
"package names should generally be snake_case or kebab-case, like `{name_in_lowercase}`, but got `{name}`
// Switch to "recommended" and offering an option
"`{name}` is not snake_case or kebab case which is recommended for package names, consider {name_in_lowercase}"
```

This is a lot harder when we don't have a specific lint to point to which provides just enough context for the original message to work.

epage:
@ehuss 

> I'm reluctant to issue this warning on a binary crate, since that affects the executable name. I don't see a reason why that should be restricted, and I think it is probably wrong that rustc issues a warning for it.

Somehow I overlooked this originally. Bin names should be kebab case and lib package names should be kebab or snake case.  That just leaves GUI bins.  That seems like a small enough of a slice that they can choose to ignore it (and in the future `allow` it if we make this a user-controlled lint)

> I'm reluctant to try to duplicate what rustc is trying to do, since keeping them in sync will be challenging and cause confusion when they are out of sync. I'd like to be very careful of whether or not this is something that cargo really needs to do, since a warning will be evident on the very first build, which I think is a minor delay compared to immediately with cargo new. I'd like to hear from other cargo team members about this.

I think its worthwhile, especially with this simplified form.
- Since this is a scaffolding problem, it seems like it'd be nice to report it then rather than the user being surprised by a warning on first build.
- Also, if the API guidelines ever get updated to distinguish crate from package naming, we'd have something unique to check compared to rustc (since we convert `-` into `_`).

epage (approved):

epage:
@shnaramn could you clean up the commits for how you'd like them merged?

PR #12768, opened 2023-10-03
refactor(toml): Make manifest file layout more consitent

epage:

This is to make the file easier to browse (to me) as I work to split out the file format from the business logic to explore ideas for how we could split out a manifest crate.

This is just moving code around, so existing tests should be sufficient.

This does try to break up the refactors by theme to make it slightly easier.

If there are ideas for further improvements, that might be best left for future refactors rather than cramming every possible idea in this one.


weihanglo (approved):
Thanks for the refactor. Still brittle but better than before!

Hope someday we can split them into modules

PR #9588, opened 2021-06-15
Enable support for fix --edition for 2021.

ehuss:
This adds support for using `cargo fix --edition` to migrate to 2021.  

This also uses the new, currently unstable, `--force-warns` flag. This was added because there were a significant number of crates that were "allow"ing lints that are required for migrating the edition. This wasn't a problem for 2018, because its lints were new, and all "allow" to start.  For 2021, several older "warn" lints are becoming hard errors.  "allow"ing them would cause the migration to fail.

PR #12771, opened 2023-10-04
feat: Add `Edition2024`

Muscraft:
[RFC for `Edition2024`](https://github.com/rust-lang/rfcs/pull/3501).  While the RFC is not yet merged, this follows rustc which added the 2024 edition previously in rust-lang/rust#94461

This PR adds `Edition2024` as a possible value for `edition = "xxxx"`. I did this by following the [guide here](https://github.com/rust-lang/cargo/blob/ed0a7873107f48079a424da2920f7d434fd22fdc/src/cargo/core/features.rs#L163-L174).

### DIFF
--- a/src/cargo/core/features.rs
+++ b/src/cargo/core/features.rs
@@ -205,20 +205,22 @@ pub enum Edition {
     Edition2018,
     /// The 2021 edition
     Edition2021,
+    /// The 2024 edition
+    Edition2024,
 }
 
 impl Edition {
     /// The latest edition that is unstable.
     ///
     /// This is `None` if there is no next unstable edition.
-    pub const LATEST_UNSTABLE: Option<Edition> = None;
+    pub const LATEST_UNSTABLE: Option<Edition> = Some(Edition::Edition2024);
     /// The latest stable edition.
     pub const LATEST_STABLE: Edition = Edition::Edition2021;
     /// Possible values allowed for the `--edition` CLI flag.
     ///
     /// This requires a static value due to the way clap works, otherwise I
     /// would have built this dynamically.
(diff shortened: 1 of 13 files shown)


src/doc/src/reference/unstable.md
```diff
@@ -1271,6 +1272,32 @@ Differences between `cargo run --manifest-path <path>` and `cargo <path>`
 
 ### Documentation Updates
 
+## Edition 2024
+* Tracking Issue: (none created yet)
```
Muscraft, line 1276: No tracking issue has been created yet, so this will have to be updated to include a link when the [RFC](https://github.com/rust-lang/rfcs/pull/3501) gets merged and a tracking issue gets created. 

If there is a tracking issue for `cargo` and `rust`, both links should probably be included.

src/cargo/core/features.rs
```diff
@@ -286,6 +292,7 @@ impl Edition {
             Edition2015 => false,
             Edition2018 => true,
             Edition2021 => true,
+            Edition2024 => false,
```
Muscraft, line 295: While there currently is [`rust_2024_compatibility`](https://github.com/rust-lang/rust/blob/187b8131d4f760f856b214fce34534903276f2ef/compiler/rustc_span/src/edition.rs#L57) lint group, I set it to [`false`](https://github.com/Muscraft/cargo/blob/b3481b7bf36c0405238cd2404ae8bdd5eaeba65e/src/cargo/core/features.rs#L295) as there are no lints in the group currently (from what I can tell). This was how `Edition2021` originally was before it was [updated to be `true`](https://github.com/rust-lang/cargo/pull/9588/files#diff-7b081a836e71ae12b28f42ba8f09384036be9ae8ee479b4849f43a8522e94b37R219).

epage (approved):

PR #11447, opened 2022-12-01
Crate checksum lookup query should match on semver build metadata

arlosi:
Since crates.io allows crate versions to differ only by build metadata, a query using `OptVersionReq::exact` + `next()` can return nondeterministic results.

This change fixes the issue by adding an additional `filter` that ensures the version is equal (including build metadata).

It still feels somewhat wrong that a query using `exact` can match multiple crates, so an alternative fix would be to add a new variant of `OptVersionReq` that also matched on build metadata.

Fixes #11412

PR #12772, opened 2023-10-04
If there's a version in the lock file only use that exact version

Eh2406:

Generally, cargo is of the opinion that semver metadata should be ignored.
It's is relevant to dependency resolution as any other description of the version, just like the description field in cargo.toml.
If your registry has two versions that only differing metadata you get the bugs you deserve.
We implement functionality to make it easier for our users or for us to maintain.

~~So let's use `==` because it's less code to write and test.~~

We also believe that lock files should ensure reproducibility
 and protect against mutations from the registry.
 In this circumstance these two goals are in conflict, and this PR picks reproducibility.
 If the lock file tells us that there is a version called `1.0.0+bar` then
 we should not silently use `1.0.0+foo` even though they have the same version.

This is one of the alternatives/follow-ups discussed in #11447.
It was separated from #12749, to allow for separate discussion and because I was being too clever by half.

All tests still pass except for the ones that were removed. The removed tests were only added to verify the on intuitive behavior of the older implementation in #9847.

### DIFF
--- a/src/cargo/sources/registry/index.rs
+++ b/src/cargo/sources/registry/index.rs
@@ -439,11 +439,9 @@ impl<'cfg> RegistryIndex<'cfg> {
     /// checking the integrity of a downloaded package matching the checksum in
     /// the index file, aka [`IndexSummary`].
     pub fn hash(&mut self, pkg: PackageId, load: &mut dyn RegistryData) -> Poll<CargoResult<&str>> {
-        let req = OptVersionReq::exact(pkg.version());
+        let req = OptVersionReq::lock_to_exact(pkg.version());
         let summary = self.summaries(pkg.name(), &req, load)?;
-        let summary = ready!(summary)
-            .filter(|s| s.package_id().version() == pkg.version())
-            .next();
+        let summary = ready!(summary).next();
         Poll::Ready(Ok(summary
             .ok_or_else(|| internal(format!("no hash listed for {}", pkg)))?
             .as_summary()
@@ -697,10 +695,8 @@ impl<'cfg> RegistryIndex<'cfg> {
         pkg: PackageId,
         load: &mut dyn RegistryData,
     ) -> Poll<CargoResult<bool>> {
-        let req = OptVersionReq::exact(pkg.version());
-        let found = ready!(self.summaries(pkg.name(), &req, load))?
-            .filter(|s| s.package_id().version() == pkg.version())
-            .any(|s| s.is_yanked());
+        let req = OptVersionReq::lock_to_exact(pkg.version());
+        let found = ready!(self.summaries(pkg.name(), &req, load))?.any(|s| s.is_yanked());
         Poll::Ready(Ok(found))
     }
 }
--- a/src/cargo/util/semver_ext.rs
+++ b/src/cargo/util/semver_ext.rs
@@ -1,6 +1,5 @@
 use semver::{Comparator, Op, Version, VersionReq};
 use serde_untagged::UntaggedEnumVisitor;
-use std::cmp::Ordering;
 use std::fmt::{self, Display};
 
 #[derive(PartialEq, Eq, Hash, Clone, Debug)]
@@ -44,6 +43,13 @@ impl OptVersionReq {
         OptVersionReq::Req(VersionReq::exact(version))
     }
 
+    // Since some registries have allowed crate versions to differ only by build metadata,
+    // A query using OptVersionReq::exact return nondeterministic results.
+    // So we `lock_to` the exact version were interested in.
+    pub fn lock_to_exact(version: &Version) -> Self {
+        OptVersionReq::Locked(version.clone(), VersionReq::exact(version))
+    }
+
     pub fn is_exact(&self) -> bool {
         match self {
             OptVersionReq::Any => false,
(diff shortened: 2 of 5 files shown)


src/cargo/sources/registry/index.rs
```diff
@@ -437,11 +437,13 @@ impl<'cfg> RegistryIndex<'cfg> {
     /// checking the integrity of a downloaded package matching the checksum in
     /// the index file, aka [`IndexSummary`].
     pub fn hash(&mut self, pkg: PackageId, load: &mut dyn RegistryData) -> Poll<CargoResult<&str>> {
-        let req = OptVersionReq::exact(pkg.version());
+        let mut req = OptVersionReq::exact(pkg.version());
+        // Since crates.io allows crate versions to differ only by build metadata,
```
epage, line 441: Because of rust-lang/crates.io#6518, should we change this to "Since some registries may allow"?

Or should we go a step further and make this a requirement on all registries?  If not, should we at least document it as a best practice?

src/cargo/sources/registry/index.rs
```diff
@@ -698,10 +700,13 @@ impl<'cfg> RegistryIndex<'cfg> {
...
-        let req = OptVersionReq::exact(pkg.version());
-        let found = ready!(self.summaries(pkg.name(), &req, load))?
-            .filter(|s| s.package_id().version() == pkg.version())
-            .any(|s| s.is_yanked());
+        let mut req = OptVersionReq::exact(pkg.version());
+
+        // Since crates.io allows crate versions to differ only by build metadata,
+        // a query using OptVersionReq::exact + next() can return nondeterministic results.
+        // So we `lock_to` the exact version were interested in
+        req.lock_to(pkg.version());
```
epage, line 708: nit: `exact` + `lock_to` seems like a weird dance.  Should we have a new constructor for this purpose?

src/cargo/util/semver_ext.rs
```diff
@@ -311,40 +315,3 @@ fn is_req(value: &str) -> bool {
...
-mod tests {
-    use super::*;
-
-    #[test]
-    fn locked_has_the_same_with_exact() {
-        fn test_versions(target_ver: &str, vers: &[&str]) {
-            let ver = Version::parse(target_ver).unwrap();
-            let exact = OptVersionReq::exact(&ver);
-            let mut locked = exact.clone();
-            locked.lock_to(&ver);
```
epage, line 325: Should we keep this test but change it so it requires metadata to be the same?

src/cargo/sources/registry/index.rs
Eh2406: I am open to wordsmithing the comment.

I'm pretty sure we already documented it. If not we should.

src/cargo/sources/registry/index.rs
Eh2406: Yes. Sounds good.

src/cargo/util/semver_ext.rs
Eh2406: I think that's testing an implementation detail. If we want to add a test for this, we could extend the test added in #11447, to ensure that if we have a lock file we get the version specified in the lock file. I may give that a try.

src/cargo/sources/registry/index.rs
Eh2406: Found it! https://doc.rust-lang.org/cargo/reference/resolver.html?highlight=metadata#version-metadata

src/cargo/util/semver_ext.rs
Eh2406: I have expanded the test to make sure that the version from a lock file is used even if there other versions available from the registry. Confusingly, it passes on head. I am trying to determine what's going on.

Eh2406:
While attempting to implement the requested test I discovered that this PR does not change behavior. We apparently already had code that ensured that if a lock file was insufficient to uniquely identify a solution the resolver would prefer versions from the lockfile. [Here](https://github.com/rust-lang/cargo/blob/b48c41aedbd69ee3990d62a0e2006edbb506a480/src/cargo/ops/resolve.rs#L451). With that in place, a lock file containing semver build metadata will be respected through a somewhat circuitous route.

- Lock file code will notice that the dependency requirement can be satisfied by a version in the lock file. It will therefore modify the dependency "lock_to" the particular version.
- Despite having been "locked" multiple packages from the registry match the dependency requirement. (Because "locked" dependencies do not look at build metadata and there are more than one package  that only differ on build metadata.)
- The resolver is free to try/choose any matching package first. It happens to choose the one from the lock file so as to minimize the diff. And because package ID equality does depend on [build metadata](https://github.com/rust-lang/cargo/blob/b48c41aedbd69ee3990d62a0e2006edbb506a480/src/cargo/core/package_id.rs#L47).

So we seem to have redundant mechanisms to ensure lock files are respected. Specifically, we have a mechanism that tell the resolver to prefer the versions from lock file AND we have a mechanism that locks dependencies to the version from lock file. The former is more elegant and simpler, but is not (yet?) capable of enforcing `--locked` nor determining when to skip network updates. Moving to only using the former is a laudable project but not in scope at this time.

In the meantime this PR makes the implementation of "locking a dependency to a version" actually match its description, while not changing any user facing behavior.

src/cargo/sources/registry/index.rs
epage: In light of everything, maybe this should be "Since some registries may incorrectly allow ..."

epage: (plus the other comment)

Eh2406:
Do any of to anyone following along this is blocked on #12806.

Eh2406:
This is unblocked, rebased, and ready for review.

I dealt with the issue of the new "locked means locked" semantics are not the ones we want for "replace" by not "locking" the replace dependencies. The plane `=` constraint gives the exact semantics that "replace" wants.

PR #10511, opened 2022-03-26
Support inheriting jobserver fd for external subcommands

weihanglo:
### What does this PR try to resolve?

If cargo detects the existence of a jobserver, cargo should pass the
jobserver down to the external subcommand. Here are the reasons:

1. The existence of jobserver implies the user "expects" the amount of
   job is under control. However, before this commit, external
   subcommands cannnot benefit from the global view of the jobserver.
2. `cargo-clippy` as an external subcommand migth also love to respect
   the jobserver protocol.
3. There are several well-known external subcommands calling "cargo"
   interally (cargo-fuzz, cargo-tarpaulin, etc.)

Fixes #10477 

[zulip stream](https://zulip-archive.rust-lang.org/stream/246057-t-cargo/topic/external.20subcommand.20inherits.20jobserver.20fd.html)

### How should we test and review this PR?

A new test `jobserver::external_subcommand_inherits_jobserver` is added.

You can also execute cargo-clippy from `make -j2` to test the behaviour 
(I usually run [`procs -w rustc`] to observe the amount of rustc invocations):

```make
# Makefile
all:
	+./target/debug/cargo clippy --release
```

### Caveats

Job without special prefix `+` might still be considered as a sub-make
and would inherit the jobserver, though I don't see it as an issue.

According to GNU Make Manual ["13.1.1 POSIX Jobserver Interaction"],
if `--jobserver-auth` option is available in `MAKEFLAGS` but the file
descriptors are closed, it means that the calling `make` didn't consider
our tool awas a recursive `make` invocation. I make an assumption that
if those fds are still open, we are happy to use those jobserver tokens.

["13.1.1 POSIX Jobserver Interaction"]: https://www.gnu.org/software/make/manual/make.html#POSIX-Jobserver
[`procs -w rustc`]: https://github.com/dalance/procs

PR #12776, opened 2023-10-06
feat: inherit jobserver from env for all kinds of runner

### LINKED ISSUE #12597: Support inheriting jobserver when using `cargo run`
### Problem

Sometimes commands run through `cargo run` actually perform build actions for a larger build system.

For example in https://github.com/rust-lang/cargo/pull/10511#issuecomment-1092057057 the scenario is running something like `cargo run -p binding_generator` which generates headers to be used during the build later.

And in https://github.com/rust-lang/rust/pull/113730#discussion_r1308504892 cargo-miri is set as `target.runner` and runs as a part of `cargo run` (and calls build tools like cargo and rustc recursively).

### Proposed Solution

In all cases like these cargo shouldn't close its jobserver file descriptors (or other handles), but pass them to the tool being run instead.

I think doing this under an options would be fine.
The default for such option can be decided upon later (probably kept being false, which is equivalent to the current behavior).

### Notes

_No response_

weihanglo:

External subcommands are already able to inherit the jobserver from
env since #10511. However, users reported that they've expected
`cargo run` to behave the same as external subcommands.

A popular example "cargo-xtask" pattern is used as scripting to run
arbitrary tasks. Users may want to invoke `cargo run` from Make and
expect some parallelism. This PR provides such an ability to the
general `target.<...>.runner`, which affects `cargo test`,
`cargo bench`, and `cargo run`.

Note that this PR doesn't create any jobserver client if there is no
existing jobserver from the environment. Nor `-j`/`--jobs` would create
a new client. Reasons for this decision:

* There might be crates don't want the jobserver to pollute their
  file descriptors, although they might be rare
* Creating a jobsever driver with the new FIFO named pipe style is not
  yet supported as of `jobserver@0.1.26`. Once we can create a named
  pipe-based jobserver, it will be less risky and inheritance by
  default can be implemented.

* aa6793dc93afe3ecfc1c8c2f6f59669350ef2db0 demonstrates the current behavior — `cargo run` doesn't support inheriting jobserver from env.
* baf606af53fd551e158dfa0616bac97b6b749fd1 Cargo starts supporting such case.

### Additional informations

r? ehuss

cc @petrochenkov, does this meet your needs?

fixes #12597

### DIFF
--- a/src/cargo/core/compiler/compilation.rs
+++ b/src/cargo/core/compiler/compilation.rs
@@ -257,7 +257,13 @@ impl<'cfg> Compilation<'cfg> {
         } else {
             ProcessBuilder::new(cmd)
         };
-        self.fill_env(builder, pkg, script_meta, kind, false)
+        let mut builder = self.fill_env(builder, pkg, script_meta, kind, false)?;
+
+        if let Some(client) = self.config.jobserver_from_env() {
+            builder.inherit_jobserver(client);
+        }
+
+        Ok(builder)
     }
 
     /// Prepares a new process with an appropriate environment to run against
--- a/tests/testsuite/jobserver.rs
+++ b/tests/testsuite/jobserver.rs
@@ -192,33 +192,25 @@ test-runner:
         .env("CARGO", cargo_exe())
         .arg("run")
         .arg("-j2")
-        .with_status(2)
-        .with_stderr_contains("[..]no jobserver from env[..]")
         .run();
     p.process(make)
         .env("PATH", path)
         .env("CARGO", cargo_exe())
         .arg("run-runner")
         .arg("-j2")
-        .with_status(2)
         .with_stderr_contains("[..]this is a runner[..]")
-        .with_stderr_contains("[..]no jobserver from env[..]")
         .run();
     p.process(make)
         .env("CARGO", cargo_exe())
         .arg("test")
         .arg("-j2")
-        .with_status(2)
-        .with_stdout_contains("[..]no jobserver from env[..]")
         .run();
     p.process(make)
         .env("PATH", path)
         .env("CARGO", cargo_exe())
         .arg("test-runner")
         .arg("-j2")
-        .with_status(2)
         .with_stderr_contains("[..]this is a runner[..]")
-        .with_stdout_contains("[..]no jobserver from env[..]")
         .run();
 
     // but not from `-j` flag
--- a/tests/testsuite/jobserver.rs
+++ b/tests/testsuite/jobserver.rs
@@ -1,18 +1,23 @@
 //! Tests for the jobserver protocol.
 
 use cargo_util::is_ci;
+use std::env;
 use std::net::TcpListener;
 use std::process::Command;
 use std::thread;
 
-use cargo_test_support::install::{assert_has_installed_exe, cargo_home};
-use cargo_test_support::{cargo_exe, project};
+use cargo_test_support::basic_bin_manifest;
+use cargo_test_support::cargo_exe;
+use cargo_test_support::install::assert_has_installed_exe;
+use cargo_test_support::install::cargo_home;
+use cargo_test_support::project;
+use cargo_test_support::rustc_host;
 
 const EXE_CONTENT: &str = r#"
 use std::env;
 
 fn main() {
-    let var = env::var("CARGO_MAKEFLAGS").unwrap();
+    let var = env::var("CARGO_MAKEFLAGS").expect("no jobserver from env");
     let arg = var.split(' ')
                  .find(|p| p.starts_with("--jobserver"))
                 .unwrap();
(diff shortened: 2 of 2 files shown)


petrochenkov:
>cc @petrochenkov, does this meet your needs?

It's hard to say from looking at the changes because I'm not familiar with cargo code.
The added test looks like https://github.com/rust-lang/cargo/pull/10511#issuecomment-1092057057, so that use case seems covered.

Is the `target.runner` case covered as well?
Is it possible to test this change with https://github.com/rust-lang/rust/pull/113730 (after removing the Miri workaround)?

weihanglo:
> Is the `target.runner` case covered as well? Is it possible to test this change with [rust-lang/rust#113730](https://github.com/rust-lang/rust/pull/113730) (after removing the Miri workaround)?

I believe the answer is yes, as it calls `target_process`, which respects `target.runner` config.
https://github.com/rust-lang/cargo/blob/5a2ea1930909dde2bae22f3a1261970de22ea4a2/src/cargo/ops/cargo_run.rs#L95

> Is it possible to test this change with [rust-lang/rust#113730](https://github.com/rust-lang/rust/pull/113730) (after removing the Miri workaround)?

Looks like miri calls cargo an rustc from [`CARGO`](https://github.com/rust-lang/miri/blob/b02a8a93728809306ab5fe3da6bc6f532368e48e/cargo-miri/src/util.rs#L111-L113) and `RUSTC` environment variables, so it should works if you build cargo from source and set `RUSTC` to the actual compiler exe path test it.

petrochenkov:
> Looks like miri calls cargo an rustc from [`CARGO`](https://github.com/rust-lang/miri/blob/b02a8a93728809306ab5fe3da6bc6f532368e48e/cargo-miri/src/util.rs#L111-L113) and `RUSTC` environment variables, so it should works if you build cargo from source and set `RUSTC` to the actual compiler exe path test it.

I'll never have time to do it in observable future :D
I guess it's simple to merge this and then try to remove the Miri FIXME when the new cargo with this change reaches bootstrap.

src/cargo/ops/cargo_run.rs
```diff
@@ -104,6 +104,10 @@ pub fn run(
         process.display_env_vars();
     }
 
+    if let Some(client) = config.jobserver_from_env() {
+        process.inherit_jobserver(client);
```
epage, line 108: I'm assuming this works fine (based on the tests) when `exec_replace` both does a replace and doesn't.

epage, line 108: From the 99% case of `cargo run` these semantics feel a bit odd.

I guess to break this down to more concrete points to overcome the oddity
- Can this change negatively impact anyone?
- Would it be a breaking change if we felt we needed to remove this in the future?

weihanglo: > Can this change negatively impact anyone?

Unless you're using `make` with `-j` arg, otherwise there is no jobserver you can inherit from the environment. So 99% case of `cargo run` won't inherit this.

> Would it be a breaking change if we felt we needed to remove this in the future?

Yes for sure. for those going to depend on this change like cargo-miri.

weihanglo: Yes. That also what I understand about `Command::exec`. See <https://github.com/rust-lang/rust/blob/b5c050febf10c9bcc0459d41fe2a1e190ad30b8d/library/std/src/sys/unix/process/process_unix.rs#L249-L270>.

epage: Overall, this feels too specialized for me and makes me wonder if we are solving the right problem.

ehuss:
> > Is the `target.runner` case covered as well? Is it possible to test this change with [rust-lang/rust#113730](https://github.com/rust-lang/rust/pull/113730) (after removing the Miri workaround)?
> 
> I believe the answer is yes, as it calls `target_process`, which respects `target.runner` config.

I'm confused by this. I believe the answer is "no", since `target_process` does not inherit the jobserver?  

I'm fine with this PR as-is, I just wanted to make sure I understand. I think I would also be fine with inheriting in a runner as well. I'm skeptical that there is really any risk with including extra file descriptors open.

ehuss:
> Unless I misunderstood the question, doesn't this PR make the `target_process`, either test executable or target runner, inherit jobserver? 🤔

I'm not sure if we are looking at the same thing. This PR (as of 7ad590451626749802995dcd491a921fa6b7b409) modifies `src/cargo/ops/cargo_run.rs` to have `cargo run` inherit the jobserver. It does not appear to modify [`target_process`](https://github.com/rust-lang/cargo/blob/7ad590451626749802995dcd491a921fa6b7b409/src/cargo/core/compiler/compilation.rs#L238-L261) or anything in the chain that would affect the `target.*.runner` configuration variable for anything other than `cargo run` (specifically, `cargo test` or `cargo bench` aren't supported).

I assume the question (`Is the target.runner case covered as well?`) is referring to the general ability of commands like `cargo test` to also forward the jobserver to `target.*.runner` executables.

weihanglo:
I see. Thanks for the clarification. Yeah I misunderstood that 😕

ehuss:
@weihanglo Just checking, are you ok with moving `inherit_jobserver` to `target_process` to also cover runners as mentioned in #12597?

weihanglo:
Ended up making target runner inherit jobserver.

@rustbot ready

src/cargo/core/compiler/compilation.rs
```diff
@@ -248,6 +248,7 @@ impl<'cfg> Compilation<'cfg> {
         kind: CompileKind,
         pkg: &Package,
         script_meta: Option<Metadata>,
+        jobserver_client: Option<&jobserver::Client>,
```
ehuss, line 251: I'm not clear, why is this being added as a new parameter if it is always passed `config.jobserver_from_env()`? Couldn't `target_process` just call `self.config.jobserver_from_env()` itself?

weihanglo: You're absolutely correct. I didn't realize it has the access to `config` 🙇🏾‍♂️.

Addressed.

weihanglo:
It's hard to find a place to document which part of cargo gets access to jobserver. I'll do it in a follow-up when figuring out.

ehuss:
Thanks!

@bors r+

RalfJung:
Given that the Miri issue, as I understood it, was about the cargo-managed jobserver (that exists during build time) not being around at run time, I don't think this lets us remove the Miri workaround. It might be good for other usecases though. :)

PR #12777, opened 2023-10-06
Add test for unsupported short config flag

hi-rustin:

Added unsupported short alias suggestion for --config flag.

Check out the unit tests.

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -378,6 +378,21 @@ pub trait CommandExt: Sized {
         )
         ._arg(unsupported_short_arg)
     }
+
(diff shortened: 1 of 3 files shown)


src/cargo/util/command_prelude.rs
```diff
@@ -378,6 +378,21 @@ pub trait CommandExt: Sized {
         )
         ._arg(unsupported_short_arg)
     }
+
+    fn arg_config(self) -> Self {
```
weihanglo, line 382: Is it possible that we inline this function? It doesn't seem to be reused elsewhere.

weihanglo (approved):
Thanks!

Issue #6009, opened 2018-09-11
cargo should support glob syntax on workspace excludes


bjgill:
Currently, we can include crates in a workspace with glob syntax (see https://github.com/rust-lang/cargo/issues/3911). However, we can't exclude crates in the same way.

See https://github.com/rust-lang/cargo/blob/master/src/cargo/core/workspace.rs#L826, where it looks as if we're just doing `manifest_path.starts_with(self.root_dir.join(exclude))`.

We've hit this because we have a set of autogenerated crates with a common suffix that we cannot neatly exclude without glob support.

nrc:
I believe this has been implemented now.

ehuss:
I believe this is still an issue.  The `exclude` field is still using `starts_with`.

ehuss:
cc #6745, which is a similar issue.

The `exclude` key was added in #3837. There doesn't appear to be any discussion at the time about the format to use.  It was enhanced in #4297 to support overlapping.

I think it would be good to consider what format and behavior `members` and `exclude` should have. Would it make sense to have something more powerful like gitignore?  I think it may be confusing if one uses globs and the other uses gitignore.  Would we need a transition period since it may result in different behavior (similar to #4268)?  Also consider that package include/exclude are mutually exclusive, but workspace members/exclude is not.

See also #4593 for improving `members` matching.

behnam:
I was supposed to work on this while back, but didn't find the time to formalize the problem and proposed solution. Let me try to do that:

Now that the file include/exclude rules are on `gitignore`-style patterns, I think we all agree that having a similar feature for member packages makes the most sense.

The problem, however, is the fact that when working with the files under a package, we already have a *root* defined (package root is the directory where the manifest file is seating). That allows having the anchor `gitignore`-style rules need when writing them.

For members, however, we don't have such a well-defined *root* anchor, since the workspace package (like A), can be in a directory next to a member package (like B). Like this:

```
<project-root>
 |-- A
 |-- B
...
```

I was thinking to add another general rule to `gitignore` rule-set, which would allow getting outside of the current root (package root = dir of manifest) and point to parent and sibling directories.

This additional rule would be: "If the pattern starts with `../`, then it's a relative glob, ...". (We need to figure out the exact details of the rest of the rule to keep it consistent with the rest of gitignore/glob rules.

Having that, then we can have A include (or exclude) B (or a package under it) in the example above.

Hope you find this helpful. Would love to hear what you think!

[...]

PR #12779, opened 2023-10-06
Add new packages to [workspace.members] automatically

### LINKED ISSUE #6378: Feature Request: cargo new --in-workspace [directory] [name]
# The problem

Imagine you're working in a project called `foo` and it's a workspace. You want to create a new member, `bar`, as a subdirectory of `foo`. If you use `cargo new bar` (or `mkdir bar && cd bar && cargo init`) you'll get warnings like this:

```
warning: compiling this new crate may not work due to invalid workspace configuration

current package believes it's in a workspace when it's not:
current:   [new project's Cargo.toml]
workspace: [encompassing workspace's Cargo.toml]

this may be fixable by ensuring that this crate is depended on by the workspace root: 
[encompassing workspace's Cargo.toml]
```

Ideally, that manual work should be automated (if the user wants it).

There's more manual work if you don't want subdirectories. For example, if `foo` is in `code/foo` and you want to make `bar` in `code/bar`, then you have to change the `Cargo.toml` files in *both* directories after using `cargo new`

# The solution

In the case of creating `bar` as a subdirectory, you'd `cd` to the workspace root and use `cargo new --in-workspace . bar` and `foo/Cargo.toml` would be updated to include `bar` in its `[workspace]` section. In the case of creating `bar` outside of the root, you'd use `cargo new --in-workspace ../foo bar` as an example. Similar capabilities should also be given to `cargo init` for consistency

If the directory passed to `--in-workspace` is not a workspace, then there would be an error.

I've never contributed to cargo before (and am pretty new to Rust) but I don't think this is too complicated, and if the idea is accepted I'd love to try and implement it. The changes would take place in [this file](https://github.com/rust-lang/cargo/blob/485670b3983b52289a2f353d589c57fae2f60f82/src/cargo/ops/cargo_new.rs). Add a boolean to [NewOptions](https://github.com/rust-lang/cargo/blob/485670b3983b52289a2f353d589c57fae2f60f82/src/cargo/ops/cargo_new.rs#L27-L35), change [its constructor](https://github.com/rust-lang/cargo/blob/485670b3983b52289a2f353d589c57fae2f60f82/src/cargo/ops/cargo_new.rs#L75-L102), update [this behavior](https://github.com/rust-lang/cargo/blob/485670b3983b52289a2f353d589c57fae2f60f82/src/cargo/ops/cargo_new.rs#L607-L614) to change the `workspaces` section of the root's `Cargo.toml` (and the child's `package.workspace` if the child isn't a subdirectory of the root), and it'll be mostly done.

# Notes

This is *not* a breaking change. The new behavior only goes into effect when a new argument (`--in-workspace`) is specified.

Relevant reading: [the original RFC](https://github.com/rust-lang/rfcs/blob/master/text/1525-cargo-workspace.md)

Alternative ideas:
+ `cargo new --root [directory] [name]`
+ `cargo new --member [name]` - strictly for subdirectories (only addresses `foo/bar` example; does not address the issue of creating `code/bar` when your root is `code/foo`)

calavera:

If a new package is created in a workspace, this change adds the package's path to the workspace's members list automatically.

It doesn't add the package to the list if the path is in the workspace's exclude list, or if the members list doesn't exist already. I noticed that a `cargo_new` test broke if I create the members list when it doesn't exist. This is because the workspace's manifest can be used for package templating. I think it's better to not break that use case.

Fixes #6378

I've included tests in the `cargo_new` suite.

### DIFF
--- a/src/cargo/ops/cargo_new.rs
+++ b/src/cargo/ops/cargo_new.rs
@@ -1,10 +1,11 @@
 use crate::core::{Edition, Shell, Workspace};
 use crate::util::errors::CargoResult;
 use crate::util::important_paths::find_root_manifest_for_wd;
+use crate::util::toml_mut::is_sorted;
 use crate::util::{existing_vcs_repo, FossilRepo, GitRepo, HgRepo, PijulRepo};
 use crate::util::{restricted_names, Config};
-use anyhow::{anyhow, Context as _};
-use cargo_util::paths;
+use anyhow::{anyhow, Context};
+use cargo_util::paths::{self, write_atomic};
 use serde::de;
 use serde::Deserialize;
 use std::collections::BTreeMap;
@@ -13,6 +14,7 @@ use std::io::{BufRead, BufReader, ErrorKind};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::{fmt, slice};
+use toml_edit::{Array, Value};
 
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum VersionControl {
@@ -809,7 +811,7 @@ fn mk(config: &Config, opts: &MkOptions<'_>) -> CargoResult<()> {
         // Sometimes the root manifest is not a valid manifest, so we only try to parse it if it is.
         // This should not block the creation of the new project. It is only a best effort to
         // inherit the workspace package keys.
-        if let Ok(workspace_document) = root_manifest.parse::<toml_edit::Document>() {
+        if let Ok(mut workspace_document) = root_manifest.parse::<toml_edit::Document>() {
             if let Some(workspace_package_keys) = workspace_document
                 .get("workspace")
                 .and_then(|workspace| workspace.get("package"))
@@ -832,6 +834,13 @@ fn mk(config: &Config, opts: &MkOptions<'_>) -> CargoResult<()> {
                 table["workspace"] = toml_edit::value(true);
                 manifest["lints"] = toml_edit::Item::Table(table);
             }
+
+            // Try to add the new package to the workspace members.
+            update_manifest_with_new_member(
+                &root_manifest_path,
+                &mut workspace_document,
+                opts.path,
+            )?;
         }
     }
 
@@ -930,4 +939,96 @@ fn update_manifest_with_inherited_workspace_package_keys(
 
         try_remove_and_inherit_package_key(key, manifest);
     }
+}
+
+/// Adds the new package member to the [workspace.members] array.
+/// - It first checks if the name matches any element in [workspace.exclude],
+///  and it ignores the name if there is a match.
+/// - Then it check if the name matches any element already in [workspace.members],
+/// and it ignores the name if there is a match.
+/// - If [workspace.members] doesn't exist in the manifest, it will add a new section
+/// with the new package in it.
+fn update_manifest_with_new_member(
+    root_manifest_path: &Path,
+    workspace_document: &mut toml_edit::Document,
+    package_path: &Path,
+) -> CargoResult<()> {
+    // Find the relative path for the package from the workspace root directory.
+    let workspace_root = root_manifest_path.parent().with_context(|| {
+        format!(
+            "workspace root manifest doesn't have a parent directory `{}`",
+            root_manifest_path.display()
+        )
+    })?;
+    let relpath = pathdiff::diff_paths(package_path, workspace_root).with_context(|| {
+        format!(
+            "path comparison requires two absolute paths; package_path: `{}`, workspace_path: `{}`",
+            package_path.display(),
+            workspace_root.display()
+        )
+    })?;
+
+    let mut components = Vec::new();
+    for comp in relpath.iter() {
+        let comp = comp.to_str().with_context(|| {
+            format!("invalid unicode component in path `{}`", relpath.display())
+        })?;
+        components.push(comp);
+    }
+    let display_path = components.join("/");
+
+    // Don't add the new package to the workspace's members
+    // if there is an exclusion match for it.
+    if let Some(exclude) = workspace_document
+        .get("workspace")
+        .and_then(|workspace| workspace.get("exclude"))
+        .and_then(|exclude| exclude.as_array())
+    {
+        for member in exclude {
+            let pat = member
+                .as_str()
+                .with_context(|| format!("invalid non-string exclude path `{}`", member))?;
+            if pat == display_path {
+                return Ok(());
+            }
+        }
+    }
+
+    // If the members element already exist, check if one of the patterns
+    // in the array already includes the new package's relative path.
+    // - Add the relative path if the members don't match the new package's path.
+    // - Create a new members array if there are no members element in the workspace yet.
+    if let Some(members) = workspace_document
+        .get_mut("workspace")
+        .and_then(|workspace| workspace.get_mut("members"))
+        .and_then(|members| members.as_array_mut())
+    {
+        for member in members.iter() {
+            let pat = member
+                .as_str()
+                .with_context(|| format!("invalid non-string member `{}`", member))?;
+            let pattern = glob::Pattern::new(pat)
+                .with_context(|| format!("cannot build glob pattern from `{}`", pat))?;
+
+            if pattern.matches(&display_path) {
+                return Ok(());
+            }
+        }
+
+        let was_sorted = is_sorted(members.iter().map(Value::as_str));
+        members.push(&display_path);
+        if was_sorted {
+            members.sort_by(|lhs, rhs| lhs.as_str().cmp(&rhs.as_str()));
+        }
+    } else {
+        let mut array = Array::new();
+        array.push(&display_path);
+
+        workspace_document["workspace"]["members"] = toml_edit::value(array);
+    }
+
+    write_atomic(
+        &root_manifest_path,
+        workspace_document.to_string().to_string().as_bytes(),
+    )
 }
--- a/tests/testsuite/cargo_new/not_inherit_workspace_package_table_if_not_members/out/Cargo.toml
+++ b/tests/testsuite/cargo_new/not_inherit_workspace_package_table_if_not_members/out/Cargo.toml
@@ -1,3 +1,5 @@
+[workspace]
+members = ["foo"]
 [workspace.package]
 authors = ["Rustaceans"]
 description = "foo"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2962,9 +2962,9 @@ dependencies = [
 
 [[package]]
 name = "serde_spanned"
-version = "0.6.3"
+version = "0.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "96426c9936fd7a0124915f9185ea1d20aa9445cc9821142f0a73bc9207a2e186"
+checksum = "12022b835073e5b11e90a14f86838ceb1c8fb0325b72416845c487ac0fa95e80"
 dependencies = [
  "serde",
 ]
@@ -3270,18 +3270,18 @@ dependencies = [
 
 [[package]]
 name = "toml_datetime"
-version = "0.6.3"
+version = "0.6.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7cda73e2f1397b1262d6dfdcef8aafae14d1de7748d66822d3bfeeb6d03e5e4b"
+checksum = "3550f4e9685620ac18a50ed434eb3aec30db8ba93b0287467bca5826ea25baf1"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "toml_edit"
-version = "0.20.2"
+version = "0.20.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "396e4d48bbb2b7554c944bde63101b5ae446cff6ec4a24227428f15eb72ef338"
+checksum = "70f427fce4d84c72b5b732388bf4a9f4531b53f74e2887e3ecb2481f68f66d81"
 dependencies = [
  "indexmap",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,7 +92,7 @@ tempfile = "3.8.0"
 thiserror = "1.0.49"
 time = { version = "0.3", features = ["parsing", "formatting", "serde"] }
 toml = "0.8.2"
-toml_edit = { version = "0.20.2", features = ["serde"] }
+toml_edit = { version = "0.20.7", features = ["serde"] }
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 unicase = "2.7.0"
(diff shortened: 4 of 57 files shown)


src/cargo/ops/cargo_compile/packages.rs
```diff
@@ -213,8 +213,3 @@ fn match_patterns(pkg: &Package, patterns: &mut Vec<(glob::Pattern, bool)>) -> b
         is_matched
     })
 }
-
-/// Build [`glob::Pattern`] with informative context.
-pub fn build_glob(pat: &str) -> CargoResult<glob::Pattern> {
```
epage, line 218: Why was this moved to `restricted_names`?

src/cargo/ops/cargo_new.rs
```diff
@@ -925,3 +941,79 @@ fn update_manifest_with_inherited_workspace_package_keys(
...
+    Ok(())
+}
+
+fn matches_in_package_list<'a, P: Iterator<Item = &'a str>>(
+    members: P,
+    package_path: &str,
+) -> CargoResult<bool> {
+    for name in members {
+        let pattern = build_glob(name)?;
+        if pattern.matches(package_path) {
```
epage, line 996: `package_path` -> `relative_path` which is the path relative to the CWD which is different than the path relative to the workspace root.
1. Please add a test for this case
2. Please make the test path.  One option is to use `pathdiff` to find the path of the new package relative to the manifest root that was discovered.

src/cargo/ops/cargo_new.rs
```diff
@@ -925,3 +941,79 @@ fn update_manifest_with_inherited_workspace_package_keys(
...
 }
+
+/// Adds the new package member to the [workspace.members] array.
+/// - It first checks if the name matches any element in [workspace.exclude],
+///  and it ignores the name if there is a match.
+/// - Then it check if the name matches any element already in [workspace.members],
+/// and it ignores the name if there is a match.
+/// - If [workspace.members] doesn't exist in the manifest, the new package
+/// is not added to the members array. This prevents manifest files from breaking
+/// if they're only used for package templating.
```
epage, line 952: This isn't making sense to me for why we would do this; this is still allowing us to create packages that will fail otherwise.

There might be tests that operate on these failed packages but that was a stop gap until this issue was resolved.

src/cargo/ops/cargo_new.rs
```diff
@@ -925,3 +941,79 @@ fn update_manifest_with_inherited_workspace_package_keys(
...
+        .and_then(|workspace| workspace.get("members"))
+        .and_then(|members| members.as_array())
+    {
+        let members = members.iter().map(|s| s.as_str().unwrap());
+        if !matches_in_package_list(members, relative_path)? {
+            let Some(toml_members) = workspace_document["workspace"]["members"].as_array_mut()
+            else {
+                anyhow::bail!("[workspace.members] is not an array");
+            };
+            toml_members.push(relative_path);
```
epage, line 982: What formatting situations should we detect and preserve?

src/cargo/ops/cargo_new.rs
```diff
@@ -925,3 +941,79 @@ fn update_manifest_with_inherited_workspace_package_keys(
...
+) -> CargoResult<()> {
+    // Don't add the new package to the workspace's members
+    // if there is an exclusion match for it.
+    if let Some(members) = workspace_document
+        .get("workspace")
+        .and_then(|workspace| workspace.get("exclude"))
+        .and_then(|members| members.as_array())
+    {
+        let members = members.iter().map(|s| s.as_str().unwrap());
+        if matches_in_package_list(members, relative_path)? {
```
epage, line 966: I don't think we support globs for exclude, see #6009

tests/testsuite/cargo_new/add_member_to_workspace/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[workspace]
```
epage, line 1: Why is this file here?

tests/testsuite/cargo_new/add_member_to_workspace/empty_members/in/Cargo.toml
```diff
@@ -0,0 +1,3 @@
+[workspace]
```
epage, line 1: > tests/testsuite/cargo_new/add_member_to_workspace/empty_members/in/Cargo.toml

Let's not nest these cases the extra level

tests/testsuite/cargo_new/add_member_to_workspace/mod.rs
```diff
@@ -0,0 +1,36 @@
+use cargo_test_support::compare::assert_ui;
+use cargo_test_support::curr_dir;
+use cargo_test_support::CargoCommand;
+use cargo_test_support::Project;
+
+#[cargo_test]
```
epage, line 6: Please follow the pattern for these tests with a `mod` in the directory of the test case

src/cargo/ops/cargo_new.rs
epage: See https://github.com/rust-lang/cargo/pull/12779/files#diff-90bbae39fa79149d8bb0faceba6b32e93efbe939964f77c3ad5077a3d0a86942R810

src/cargo/ops/cargo_new.rs
epage: I can't think of any examples but to dig into `toml_edit`s API.

There are likely to main formats to work with
```toml
members = ["one", "two"]
members = [
    "one",   # amount of indentation may vary
    "two",  # use of trailing comma may vary
]
```

With `cargo add`, the only format preserving we do is to keep the list of dependencies sort if it isn't.  We are prescriptive on how to format dependencies rather than making it configurable.

Here, we might have some wiggle room if the people working on `Cargo.toml` style guide (@joshtriplett ?) want to be prescriptive on indentation or trailing commas, because we can just detect specific cases and preserve those and let the user deal with it otherwise.

A third route is a "we don't care" and leave it to a future version of `cargo fmt` that updates `Cargo.toml`.  This is basically the approach that `cargo fix` takes.

ehuss:
r? epage

tests/testsuite/cargo_new/add_member_to_workspace/empty_members/in/Cargo.toml
epage: Please remove the nesting.  So far we have these flat, even for related tests.

tests/testsuite/cargo_new/add_member_to_workspace/absolute_package_path/in/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+pub fn add(left: usize, right: usize) -> usize {
```
epage, line 1: Could you pare these tests down
- Remove the README
- Remove the `lib.rs` since the workspace is virtual anyways

(ditto for others)

tests/testsuite/cargo_new/not_inherit_workspace_package_table_if_not_members/out/Cargo.toml
```diff
@@ -1,3 +1,5 @@
+[workspace]
+members = ["foo"]
 [workspace.package]
```
epage, line 3: Again, the question we need to answer: how much formatting should we do vs defer to `cargo fmt` like `cargo fix`

In this case, we normally do newlines between tables, but that'd require you to detect when adding `members` affects that.
- In this case, we'd need `members` to have a newline in its decor suffix
- In other cases (like if `resolver = "2"` is present), we'll need to remove the decor suffix on the last field and add it to `members`

src/cargo/ops/cargo_new.rs
epage: ... that is strange.  I'm the maintainer of `toml_edit` and I'm not sure why that worked.

epage: With the idea that we will have `Cargo.toml` auto-fix in the near future, I think it'd be worthwhile doing what auto-fix does for Rust: punt to `cargo fmt`.  Going to check with others to see if we're ok moving forward with that.

epage: See https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Automatic.20edits.20to.20.60Cargo.2Ctoml.60

epage:
Could you show the debug representation for each?

epage:
At this point, it comes down to what we want out policy to be for formatting.

epage:
Let's go ahead and remove the auto-formatting but keep the relevant tests.  If nothing else, this is an improvement over what we already have and changing it isn't a breaking change.

src/cargo/ops/cargo_new.rs
```diff
@@ -925,3 +933,103 @@ fn update_manifest_with_inherited_workspace_package_keys(
...
+            if Path::new(pat) == relpath {
+                return Ok(());
+            }
+        }
+    }
+
+    let display_path = relpath
+        .to_str()
+        .map(|s| s.replace('\\', "/"))
+        .with_context(|| format!("invalid non-unicode path `{}`", relpath.display()))?;
```
epage, line 984: You likely want to do
```rust
let display_path: Result<String, _> = relpath.components().map(|os| os.to_str());
```
The key part being to use `components()` to split the path.  `\\` is a valid character in a path on Linux.

src/cargo/ops/cargo_new.rs
```diff
@@ -925,3 +933,103 @@ fn update_manifest_with_inherited_workspace_package_keys(
...
+            let pat = member
+                .as_str()
+                .with_context(|| format!("invalid members' element `{}`", member))?;
+            if Path::new(pat) == relpath {
+                return Ok(());
+            }
+        }
+    }
+
+    let display_path = relpath
```
epage, line 981: If we already need to do this, why not move it above and use it for exclude?

src/cargo/ops/cargo_new.rs
```diff
@@ -925,3 +933,103 @@ fn update_manifest_with_inherited_workspace_package_keys(
...
+    // in the array already includes the new package's relative path.
+    // - Add the relative path if the members don't match the new package's path.
+    // - Create a new members array if there are no members element in the workspace yet.
+    if let Some(members) = workspace_document
+        .get("workspace")
+        .and_then(|workspace| workspace.get("members"))
+        .and_then(|members| members.as_array())
+    {
+        for member in members {
+            let pat = member.as_str().unwrap();
```
epage, line 996: Can we handle this unwrap?

src/cargo/ops/cargo_new.rs
```diff
@@ -925,3 +933,103 @@ fn update_manifest_with_inherited_workspace_package_keys(
...
+            let pattern = glob::Pattern::new(pat)
+                .with_context(|| format!("cannot build glob pattern from `{}`", pat))?;
+
+            if pattern.matches_path(&relpath) {
+                return Ok(());
+            }
+        }
+
+        let Some(toml_members) = workspace_document["workspace"]["members"].as_array_mut() else {
+            anyhow::bail!("[workspace.members] is not an array");
```
epage, line 1006: Since `workspace.members` is not a table, can you remove the `[]`?  Also, let's use backticks around things like that in messages.

src/cargo/ops/cargo_new.rs
```diff
@@ -925,3 +933,103 @@ fn update_manifest_with_inherited_workspace_package_keys(
...
+        };
+
+        toml_members.push(&display_path);
```
epage, line 1009: Let's do a sort check on this.  If its sorted, then keep it sorted.  We do that for dependency tables.

src/cargo/ops/cargo_new.rs
```diff
@@ -925,3 +933,103 @@ fn update_manifest_with_inherited_workspace_package_keys(
...
+            .as_table_mut()
+            .with_context(|| format!("[workspace] is not a table"))?;
+
+        let mut array = Array::new();
+        array.push(&display_path);
+
+        // only add a trailing new line if the manifest doesn't
+        // include any elements that are not inline tables.
+        // This respects the previous document formatting if there was any already.
+        if !ws.get_values().iter().any(|(_, v)| !v.is_inline_table()) {
```
epage, line 1021: Why does something being an inline table matter to this check?  Would we also need to remove a previous trailing newline?

src/cargo/ops/cargo_new.rs
```diff
@@ -925,3 +933,103 @@ fn update_manifest_with_inherited_workspace_package_keys(
...
+        ws.insert("members", toml_edit::value(array));
+    } else {
+        let mut array = Array::new();
+        array.push(&display_path);
+        array.decor_mut().set_suffix("\n"); // always add a trailing new line
+
+        workspace_document["workspace"]["members"] = toml_edit::value(array);
+    }
+
+    paths::write(&root_manifest_path, workspace_document.to_string())
```
epage, line 1034: This should probably be `write_atomic`, see #12744

epage:
While at it, could you clean up the commits for how you'd want them merged into the repo?

epage:
> @epage I don't really have any preference about commit ordering. Do you want me to squash all of them together? What's the project's practice?

If you could clean them up to how you want them in history, that'd be appreciated.  Ideally, all commits would be passing tests, be atomic, and tell a story.  Sometimes, we add a test-only commit and then the commit with the feature/fix so the change in behavior is more obvious.

src/cargo/ops/cargo_new.rs
```diff
@@ -962,11 +962,17 @@ fn update_manifest_with_new_member(
...
-    let display_path = relpath
-        .iter()
-        .filter_map(|comp| comp.to_str())
-        .collect::<Vec<_>>()
-        .join("/");
+    let mut components = Vec::new();
+    for comp in relpath.iter() {
+        let comp = comp.to_str().with_context(|| {
+            format!(
+                "invalid non-ascii component in path `{}`",
```
epage, line 969: "invalid unicode component in path"

src/cargo/ops/cargo_new.rs
```diff
@@ -962,11 +962,17 @@ fn update_manifest_with_new_member(
...
     })?;
 
-    let display_path = relpath
-        .iter()
-        .filter_map(|comp| comp.to_str())
-        .collect::<Vec<_>>()
-        .join("/");
+    let mut components = Vec::new();
+    for comp in relpath.iter() {
+        let comp = comp.to_str().with_context(|| {
```
epage, line 967: nit: you can do this as `let components = relpath.iter().map(...).collect::<Result<_, _>>()?;` but its sometimes I pain to do it that way, so not required.

src/cargo/util/toml/mod.rs
```diff
@@ -3470,3 +3470,19 @@ impl<'de> de::Deserialize<'de> for PathValue {
         Ok(PathValue(String::deserialize(deserializer)?.into()))
     }
 }
+
+// Based on Iterator::is_sorted from nightly std; remove in favor of that when stabilized.
+pub fn is_sorted(mut it: impl Iterator<Item = impl PartialOrd>) -> bool {
```
epage, line 3475: nit: I think this would more belong in `toml_mut/mod.rs` since its related to toml modifications

epage:
Everything at this point is minor.  Feel free to clean up the commits.

epage:
Thanks!

@bors r+

PR #12782, opened 2023-10-06
crates-io: Add doc comment for `NewCrate` struct

Turbo87:
This PR adds a small doc comment to the `NewCrate` struct in the `crates-io` package, linking to https://doc.rust-lang.org/cargo/reference/registry-web-api.html#publish to give additional context.

I saw other links to https://doc.rust-lang.org in some of the doc comments, so I'm hoping this is correct vs. linking directly to the file with a relative path?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -582,7 +582,7 @@ dependencies = [
 
 [[package]]
 name = "crates-io"
-version = "0.39.0"
+version = "0.39.1"
 dependencies = [
  "curl",
  "percent-encoding",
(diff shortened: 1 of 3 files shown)


epage (approved):
Just need to bump the version of the crate so it gets a release with this change.

Turbo87:
probably no need to bump it just for this change. it was mostly helpful for me while browsing the cargo codebase :)

Turbo87:
oh wait, now that I see the failing CI... did you mean to say that I should bump the version myself? how are changelogs etc. handled in that case?

epage:
Yes, to fix CI the version needs to be bumped.  Its a crude check, making us release more often than we would otherwise, but we've had too many times where we forgot to release.

So far we don't have per-crate changelogs.  I'm researching options for that.

epage:
Thanks!

PR #12783, opened 2023-10-06
cargo: prevent dashes in lib.name

### LINKED ISSUE #12780: Inconsistent definitions for lib.name regarding hyphens
Currently, Cargo refuses values with hyphens or special characters in `lib.name`, but uses the package-name as default value, if unspecified. The package-name can contain hyphens, though. This leads to inconsistent behavior, where `lib.name` ends up with special characters, yet Cargo refuses specifying such values directly in the manifest.

Cargo documentation states that the default value is the package-name, but any special characters replaced by underscores. Unfortunately, Cargo never implemented this transformation, thus leading to the inconsistent behavior.

PR #12640 tried to change Cargo to allow any value as `lib.name` (aligning library targets with all other targets). After discussion, the recommended approach is to instead align Cargo with its documentation and actually transforming the package-name to a valid `lib.name`.

Related issue: #11259

dvdhrm:
The TOML parser of Cargo currently refuses `lib.name` entries that contain dashes. Unfortunately, it uses the package-name as default if no explicit `lib.name` entry is specified. This package-name, however, can contain dashes.

Cargo documentation states that the package name is converted first, yet this was never implemented by the code-base.

Fix this inconsistency and convert the package name to a suitable crate-name first.

This fixes #12780. It is an alternative to #12640.

### DIFF
--- a/tests/testsuite/artifact_dep.rs
+++ b/tests/testsuite/artifact_dep.rs
@@ -817,8 +817,10 @@ fn lib_with_selected_dashed_bin_artifact_and_lib_true() {
                 let _b = include_bytes!(env!("CARGO_BIN_FILE_BAR_BAZ_baz-suffix"));
                 let _b = include_bytes!(env!("CARGO_STATICLIB_FILE_BAR_BAZ"));
                 let _b = include_bytes!(env!("CARGO_STATICLIB_FILE_BAR_BAZ_bar-baz"));
+                let _b = include_bytes!(env!("CARGO_STATICLIB_FILE_BAR_BAZ_bar_baz"));
                 let _b = include_bytes!(env!("CARGO_CDYLIB_FILE_BAR_BAZ"));
                 let _b = include_bytes!(env!("CARGO_CDYLIB_FILE_BAR_BAZ_bar-baz"));
+                let _b = include_bytes!(env!("CARGO_CDYLIB_FILE_BAR_BAZ_bar_baz"));
             }
         "#,
         )
--- a/src/cargo/core/compiler/artifact.rs
+++ b/src/cargo/core/compiler/artifact.rs
@@ -29,15 +29,39 @@ pub fn get_env(
             let path = artifact_path.parent().expect("parent dir for artifacts");
             env.insert(var, path.to_owned().into());
 
-            let var = format!(
+            let var_file = format!(
                 "CARGO_{}_FILE_{}_{}",
                 artifact_type_upper,
                 dep_name_upper,
                 unit_dep.unit.target.name()
             );
-            env.insert(var, artifact_path.to_owned().into());
 
-            if unit_dep.unit.target.name() == dep_name.as_str() {
+            // In older releases, lib-targets defaulted to the name of the package. Newer releases
+            // use the same name as default, but with dashes replaced. Hence, if the name of the
+            // target was inferred by Cargo, we also set the env-var with the unconverted name for
+            // backwards compatibility.
+            let need_compat = unit_dep.unit.target.is_lib() && unit_dep.unit.target.name_inferred();
+            if need_compat {
+                let var_compat = format!(
+                    "CARGO_{}_FILE_{}_{}",
+                    artifact_type_upper,
+                    dep_name_upper,
+                    unit_dep.unit.pkg.name(),
+                );
+                if var_compat != var_file {
+                    env.insert(var_compat, artifact_path.to_owned().into());
+                }
+            }
+
+            env.insert(var_file, artifact_path.to_owned().into());
+
+            // If the name of the target matches the name of the dependency, we strip the
+            // repetition and provide the simpler env-var as well.
+            // For backwards-compatibility of inferred names, we compare against the name of the
+            // package as well, since that used to be the default for library targets.
+            if unit_dep.unit.target.name() == dep_name.as_str()
+                || (need_compat && unit_dep.unit.pkg.name() == dep_name.as_str())
+            {
                 let var = format!("CARGO_{}_FILE_{}", artifact_type_upper, dep_name_upper,);
                 env.insert(var, artifact_path.to_owned().into());
             }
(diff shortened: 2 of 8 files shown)


dvdhrm:
(update: fixed rustfmt complaints)

epage:
@rfcbot fcp merge

When auto-filling `lib.name`, cargo has a discrepancy
- documentation: says lib.name will use `_`
- implementation: any `-` -> `_` conversion happens later in the process, missing some areas like `cargo metadata` and environment variables set for artifact dependencies

We talked in the cargo team meeting about leaning towards making the code match the documentation.  This FCP is a strawpoll to confirm that.

In stable rust, this only affects the lib name in `cargo metadata`.  I've been looking for any other way this may manifest, including [our environment variables](https://doc.rust-lang.org/cargo/reference/environment-variables.html) and haven't seen any so far.

epage (approved):

ehuss:
Hm, I didn't think about the consequence of this impacting the artifact-dependency environment variables, which appears to be a breaking change. For example, I did a quick code search and found https://github.com/lf-/clipper/blob/dd1caf13191e0fa58afbc3680fb83782c20e63d2/fixtures/dlopen-openssl-fixture/src/main.rs#L9 which would break with this change.

I'm not sure what to do about that. Perhaps one option is to set both environment variables for backwards compatibility?

dvdhrm:
The RFC says the env-var is named after the "crate", so this is actually a fix. But that is more of a theoretical argument, it is still a breaking change.

epage:
> . Perhaps one option is to set both environment variables for backwards compatibility?

If that works, it seems reasonable.

dvdhrm:
(update: we now remember whether target-names were inferred, which in turn allows us to keep setting the previously used artifact env-vars)

dvdhrm:
The `bindeps` and `multidep` features of Cargo are unstable, but the new version tries to avoid breakage in the env-vars set by these features. Unfortunately, `dash->underscore` conversion is lossy, so we cannot reliably tell whether compatibility-mode is required. Therefore, I extended the Cargo-internal manifest state to remember whether a target-name was inferred, or whether it is explicit. Now we know whether a target-name used to be different in the past, and we can use this information to keep backwards-compatibility.

Note that I think the stripped env-var of `bindeps` is quite confusing, since changing the dep-name _locally_ will affect which env-var becomes the stripped one. I think I did my best to retain this behavior, so this should be fine.

dvdhrm:
(update: rebase to master to fix CI failures)

dvdhrm:
(Any update on this?)

tests/testsuite/artifact_dep.rs
```diff
@@ -817,8 +817,10 @@ fn lib_with_selected_dashed_bin_artifact_and_lib_true() {
                 let _b = include_bytes!(env!("CARGO_BIN_FILE_BAR_BAZ_baz-suffix"));
                 let _b = include_bytes!(env!("CARGO_STATICLIB_FILE_BAR_BAZ"));
                 let _b = include_bytes!(env!("CARGO_STATICLIB_FILE_BAR_BAZ_bar-baz"));
+                let _b = include_bytes!(env!("CARGO_STATICLIB_FILE_BAR_BAZ_bar_baz"));
                 let _b = include_bytes!(env!("CARGO_CDYLIB_FILE_BAR_BAZ"));
                 let _b = include_bytes!(env!("CARGO_CDYLIB_FILE_BAR_BAZ_bar-baz"));
+                let _b = include_bytes!(env!("CARGO_CDYLIB_FILE_BAR_BAZ_bar_baz"));
```
weihanglo, line 823: I am not sure, but from the [doc of artifact dependencies](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#artifact-dependencies-environment-variables).

> Note that `<NAME>` is not modified in any way from the `name` specified in the crate supplying the artifact, or the crate name if not specified; for instance, it may be in lowercase, or contain dashes.

So this is still a breaking change and the old behavior is more straightforward to me.

ehuss: I'm not quite following this comment. Can you clarify, what is the breaking change, and which behavior is more straightforward?

FWIW, I didn't really have an answer for the question about the change in behavior. Since this is an unstable feature, we should be fine to change it, I was just tossing it out as something I noticed that would be affected. Do you have a proposal on how this should behave?

weihanglo: Just feel like it introduces another inconsistency. Now we have a mix of env with uppercase and dash-replaced, but for the artifact name parts we keep original cases with dash-replaced. And the actual artifact is either dash-replaced or not depending on their crate-types.

I wish we could have no case toggle or replacement since the beginning. But yeah it is a thing we could definitely change and stick to a well-written rule.

dvdhrm: I am a bit lost what to do with this. I mean, yes, this entire PR changes the way how Cargo decides the default value of library targets. So yeah, it is a breaking change. However, I think we still expose compatible behavior by retaining the previous env-vars. So what exactly do you mean with "this is still a breaking change"? Can you show an example how a previously working package would no longer work with this patch?

Apart from that, with this patch applied, Cargo still uses `<NAME>` unmodified in any way from the `name` specified in the crate supplying the artifact, or the crate name if not specified.
I mean, if I understand correctly, `crate name` refers to dashes replaced with underscores in Cargo (unlike `package name`), so this PR makes the behavior match the RFC, unlike before, doesn't it?

weihanglo: My apologies. I missed the compatible part. Looks like it would be sufficient. Thank you for your clarification.

ehuss:
@Eh2406, @Muscraft, @joshtriplett do you think you can check the FCP [above](https://github.com/rust-lang/cargo/pull/12783#issuecomment-1750726185)? Do you have any questions or concerns? Anything that can help you understand this change?

Eh2406:
Sorry for the delay in ✔️

ehuss:
Thanks! Will merge now that the FCP has passed.

@bors r+

PR #12785, opened 2023-10-06
refactor(core): Clean up deserialization of PackageId

epage:
This came as part of another change that I'm abandoning but I felt this made things cleaner, so I decided to keep it.

### DIFF
--- a/src/cargo/core/package_id.rs
+++ b/src/cargo/core/package_id.rs
@@ -82,25 +82,29 @@ impl<'de> de::Deserialize<'de> for PackageId {
         D: de::Deserializer<'de>,
     {
         let string = String::deserialize(d)?;
-        let mut s = string.splitn(3, ' ');
-        let name = s.next().unwrap();
-        let name = InternedString::new(name);
-        let Some(version) = s.next() else {
-            return Err(de::Error::custom("invalid serialized PackageId"));
-        };
-        let version = version.to_semver().map_err(de::Error::custom)?;
-        let Some(url) = s.next() else {
-            return Err(de::Error::custom("invalid serialized PackageId"));
-        };
-        let url = if url.starts_with('(') && url.ends_with(')') {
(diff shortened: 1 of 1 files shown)


src/cargo/core/package_id.rs
```diff
@@ -82,27 +82,31 @@ impl<'de> de::Deserialize<'de> for PackageId {
...
-        };
-        let url = if url.starts_with('(') && url.ends_with(')') {
-            &url[1..url.len() - 1]
-        } else {
-            return Err(de::Error::custom("invalid serialized PackageId"));
-        };
+
+        let (field, rest) = string
+            .split_once(' ')
+            .ok_or_else(|| de::Error::custom("invalid serialized PackageId"))?;
```
hi-rustin, line 88: Should we declare the error message in advance? Then we don't need to write the same string third times.

hi-rustin (approved):
Looks good to me. Thanks!

weihanglo (approved):
Looks good to me.

`r=me` whenever you feel good or fix <https://github.com/rust-lang/cargo/pull/12785#discussion_r1349426876>

src/cargo/core/package_id.rs
epage: For me, DRY isn't about straight code de-duplication but about centralizing requirements.  Here there aren't those (its an error but nothing says the message should be the same) and in fact, in the ideal world we'd provide more details, rather than less. A focus on de-duplicating errors makes it so there is greater inertia to not change  them.

I also feel that it leads to readability because it lets you think more locally.  Too much abstraction can make code harder to follow (like with cache locality, I feel there is a human element to reading code that I refer to as code locality).

All of this is likely more thought than this off-hand comment was likely meant to convey.  I can see multiple people having valid points in different directions on this.  In that case, I lean towards status quo as this PR isn't making the situation worse but is already helping to improve other things.

Issue #6583, opened 2019-01-22
Possibility to set the package version on command line


vorner:
Hello

I'm building a rust application using internal teamcity CI job. I want the build number to be part of the version of the binary (so eg. `./app --version` knows from which build it came).

The only way I found so far is to let the build first edit the `Cargo.toml` (`sed -i -e 's/^version = .*/version = "%build.version%"/' Cargo.toml`), which seems ugly and fragile.

It would be great if I could somehow override the version from `Cargo.toml` through the command line ‒ either as a parameter or an environment variable.

Would something like that make sense? Are there plans to support it?

dwijnand:
I think this could be created as a cargo extension using https://github.com/ordian/toml_edit.  You could use https://github.com/killercup/cargo-edit as inspiration.

vorner:
Actually, I don't want to modify the `Cargo.toml` at all if possible, just use that different value of `version` during a single `cargo build`. So something like `cargo build --config package.version="0.1.2"`.

But yes, you're right, there are certainly less fragile ways to edit `Cargo.toml`.

dwijnand:
I think there's a ticket for being able to modify options (I guess I'm both Cargo.toml and .cargo/config) via a command, like git config, but I don't know if it extends to temporary overrides.

mixalturek:
It's very common use case to specify the version from command line while releasing on a CI server that tracks and auto increments the build number. GNU/make, Maven and Gradle natively support it, manual editing of a build-spec file is not practical and error prone.

```
make target VERSION=%build.version%
mvn release:prepare -DreleaseVersion=%build.version%
./gradlew build -Pversion=%build.version%
```

dwijnand:
`mvn release:prepare` is based on the maven's release plugin.  The equivalent is https://github.com/sunng87/cargo-release, and I don't know if it supports specifying the version on the command line.

dwijnand:
What does 

> the binary should be fed with a version different from the one in `Cargo.toml`

mean?  Is it just `env!("CARGO_PKG_VERSION")` (and variants, like `CARGO_PKG_VERSION_MAJOR`) usage?

vorner:
I'm not 100% sure there are no other places the version is passed into the compilation, but I guess so. I haven't tried yet (I might), but I guess cargo will set the env var unconditionally and overwrite it even if I pass it from outside.

dwijnand:
Maybe we could just change that: only set the environment var if unset.

[...]

Issue #6153, opened 2018-10-07
cargo publish should be opt-in


jmaargh:
Publishing, whether to crates.io or elsewhere, should be opt-in rather by on-by-default. Although rare, people [do sometimes publish accidentally](https://stackoverflow.com/questions/52639162/how-to-delete-a-published-crate-from-crates-io). Given that the default is currently to publish to crates.io publicly, this is potentially dangerous for using cargo in a corporate closed-source environment. My guess is that even most always-intended-to-be-open-source projects would rather keep things private until they're ready for some sort of release.

I see several options for making this work:

1. Have `cargo new` add `publish = false` or `publish = []` or equivalent in the default `Cargo.toml`. This is probably the most minor change and I can't imagine this will break any existing uses.

2. Change the behaviour so that `publish = true` in `Cargo.toml` is _required_ for `cargo publish` to work, with an error message explaining this. This might break workflow for some existing projects, but it would be a very easy fix.

3. Change the behaviour of `cargo publish` so that if there is no `publish = True` in `Cargo.toml` it interactively asks whether you're sure. Possibly with a `-y`/`--yes` option (a la `apt install`) to automatically say "yes".

Personally, I'd prefer both 1 and 2. I'd be happy to make a PR if people like this idea.

Originally raised as an idea on #6123, but was decided it belonged as a separate issue.

dwijnand:
Is this a distinct issue to that being discussed in #6123?

yonni12:
Yes, I posted a comment there and people asked me to create as separate issue. Sorry, should have linked.

withoutboats:
My initial inclination is against doing something like this, but I could be convinced.

There are a few barriers already against accidentally publishing:

1. The `cargo publish` command has no other purpose, and none of our other major commands are similar, making users less likely to accidentally run `cargo publish` when they meant to run something else.
2. The publication will fail if you aren't logged in, making it less likely for you to successfully publish something when you are new to Rust & haven't yet learned what `cargo publish` is for. That is, the first time you publish you have to actively try to by creating a crates.io account, so we avoid the problem of people running commands without understanding them.
3. The crate created by `cargo new` will be rejected by crates.io because it lacks a license and description field. You have to have set both of those up before you can publish.
4. `cargo publish` fails if your crate doesn't build or if your VCS repository is dirty.
5. If you're publishing a subsequent version of a crate, `cargo publish` will fail if you haven't updated the version number to the new version.

[...]

Issue #5979, opened 2018-09-05
Cargo.lock contains project's own version, unnecessarily


kornelski:
In `Cargo.lock` there's a `[[package]]` entry for the project itself, and like all entries, it contains a `version` field.

If I bump the version in `Cargo.toml`, the `Cargo.lock` version becomes out of sync, and that causes some annoyances for a project that keeps `Cargo.lock` under version control:

 * When I bump the project version, I have to remember to also run some lock-rewriting Cargo command to update the lockfile version, even if I don't intend to update any dependencies. This feels like an unnecessary step to be mindful of. 

 * OTOH, if I don't update the lockfile at the same time when bumping version, some future operation will. This causes an unwanted/unrelated change in later commits in the project.

Fortunately, it looks like the version in `Cargo.toml` is not actually used for anything important. Ideally, I'd like to completely remove the `version` field from packages in `Cargo.lock` that are workspace members, but for better backwards compatibility, the version could be locked to some constant value (e.g. `0.0.0-workspace`).

Technically it seems possible: in `EncodableResolve::into_resolve()`, when `EncodableDependency` is in the workspace, use the version from the `ws.members()` instead.

Does that make sense? Would that work? Would you accept such change?

PR #12786, opened 2023-10-06
feat(toml): Allow version-less manifests

### LINKED ISSUE #9829: Consider making the `version` field optional in Cargo.toml

Today, the `version` field is required. However, in many (most?) cases it doesn't make sense:

* when doing some casual rust hacking for fun or learning 
* inside "live at HEAD" company monorepo
* when working on a large Rust projects not distributed via `cargo install`

This creates specific issues:

* additional confusion -- there is a line in Cargo.toml which is there "just because", and not due to some specific reason
* diluted meaning of versions: semver compatibility is a lot to promise,  by having versions everywhere it's harder to notice where they matter and where you need to care about API compatibility
* For projects which mix published and non-published crates in the same source tree, it becomes harder to know at a glance if a particular crate is published. 

Make the `version` field in Cargo.toml optional. Do populate it by default, as it is today. Prevent publishing the crates which (transitively) depend on non-versioned crates. 

**Notes**

There is `publish = false`, to my mind that serves a different purpose -- preventing accidental publishing of the crates. It doesn't help with solving confusion problem -- the version is still there, and it still doesn't have a meaning. Although, if we were to re-design Cargo from scratch, I would argue that we shouldn't have `publish` field at all, and insead define it as `publish = version.is_some()`. 

Implementation wise, it seems like we can desugar absence of version as `version = "0.0.0" publish = false`. I personally use `version = "0.0.0"` as a way to explicitly signal that there's no meaningful version.

### LINKED ISSUE #12690: Change the `package.publish` default to `false` on an edition boundary
### Problem

For `Cargo.toml`, `package.publish` defaults to `true`.

For "cargo script", `package.publish` defaults to `false`.

I also feel like publishing should be more of an explicit choice. especially when working on a workspace where a `cargo release` (and eventually `cargo publish` in #1169) can release everything in the workspace.

### Proposed Solution

By changing the default, we close the documentation gap between these features.

### Notes

See also #6153

epage:

Expected behavior with this PR:
- `package.version` defaults to `0.0.0`
- `package.publish` is defaulted to `version.is_some()`

This also updates "cargo script" to rely on this new behavior.

My motivation is to find ways to close the gap between "cargo script" and `Cargo.toml`.  With "cargo script", we want to allow people to only write however much of a manifest is directly needed for the work they are doing (which includes having no manifest).  Each difference between "cargo script" and `Cargo.toml` is a cost we have to pay in our documentation and a hurdle in a users understanding of what is happening.

There has been other interest in this which I also find of interest (from #9829):
- Lower boilerplate, whether for [cargo xtasks](https://github.com/matklad/cargo-xtask), nested packages (rust-lang/rfcs#3452), etc
- Unmet expectations from users because this field is primarily targeted at registry operations when they want it for their marketing version (#6583).
- Make "unpublished" packages stand out

This then unblocks unifying `package.publish` by making the field's default based on the presence of a version as inspired by the proposal in #9829.  Without this change, we were trading one form of boilerplate (`version = "0.0.0"`) for another (`publish = false`).

Fixes #9829
Fixes #12690
See also #6153

The initial commit has test cases I thought would be relevant for this change and you can see how each commit affects those or existing test cases.  Would definitely be interested in hearing of other troubling cases to test

Implementation wise, I made `MaybeWorkspaceVersion` deserializer trim spaces so I could more easily handle the field being an `Option`.  This is in its own commit.

### Additional information

Alternatives considered
- Making the default version "stand out more" with it being something like `0.0.0+HEAD`.  The extra noise didn't seem worth it and people would contend over what the metadata field *should be*
- Make the default version the lowest version possible (`0.0.0-0`?).  Unsure if this will ever really matter especially since you can't publish
- Defer defaulting `package.publish` and instead error
  - Further unifying more fields made this too compelling for me :)
- Put this behind `-Zscript` and make it a part of rust-lang/rfcs#3502
  - Having an affect outside of that RFC, I wanted to make sure this got the attention it deserved rather than getting lost in the noise of a large RFC.
- Don't just default the version but make packages versionless
  - I extended the concept of versionless to `PackageId`'s internals and saw no observable difference
  - I then started to examine the idea of version being optional everywhere (via `PackageId`s API) and ... things got messy especially when starting to look at the resolver.  This would have also added a lot of error checks / asserts for "the version must be set here".  Overall, the gains seemed questionable and the cost high, so I held off.

### DIFF
--- a/src/cargo/ops/registry/publish.rs
+++ b/src/cargo/ops/registry/publish.rs
@@ -103,7 +103,7 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
         if allowed_registries.is_empty() {
             bail!(
                 "`{}` cannot be published.\n\
-                 `package.publish` is set to `false` or an empty list in Cargo.toml and prevents publishing.",
+                 `package.publish` must be set to `true` or a non-empty list in Cargo.toml to publish.",
                 pkg.name(),
             );
         } else if !allowed_registries.contains(&reg_name) {
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -1012,7 +1012,8 @@ impl TomlManifest {
         let publish = match publish {
             Some(VecStringOrBool::VecString(ref vecstring)) => Some(vecstring.clone()),
             Some(VecStringOrBool::Bool(false)) => Some(vec![]),
-            None | Some(VecStringOrBool::Bool(true)) => None,
+            Some(VecStringOrBool::Bool(true)) => None,
+            None => version.is_none().then_some(vec![]),
         };
 
         if version.is_none() && publish != Some(vec![]) {
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -550,11 +550,17 @@ impl TomlManifest {
         let version = package
             .version
             .clone()
-            .resolve("version", || inherit()?.version())?;
+            .map(|version| version.resolve("version", || inherit()?.version()))
+            .transpose()?;
 
-        package.version = MaybeWorkspace::Defined(version.clone());
+        package.version = version.clone().map(MaybeWorkspace::Defined);
 
-        let pkgid = package.to_package_id(source_id, version)?;
+        let pkgid = package.to_package_id(
+            source_id,
+            version
+                .clone()
+                .unwrap_or_else(|| semver::Version::new(0, 0, 0)),
+        )?;
 
         let edition = if let Some(edition) = package.edition.clone() {
             let edition: Edition = edition
@@ -1008,6 +1014,10 @@ impl TomlManifest {
             Some(VecStringOrBool::Bool(false)) => Some(vec![]),
             None | Some(VecStringOrBool::Bool(true)) => None,
         };
+
+        if version.is_none() && publish != Some(vec![]) {
+            bail!("`package.publish` requires `package.version` be specified");
+        }
 
         if summary.features().contains_key("default-features") {
             warnings.push(
@@ -1659,7 +1669,7 @@ pub struct TomlPackage {
     edition: Option<MaybeWorkspaceString>,
     rust_version: Option<MaybeWorkspaceRustVersion>,
     name: InternedString,
-    version: MaybeWorkspaceSemverVersion,
+    version: Option<MaybeWorkspaceSemverVersion>,
     authors: Option<MaybeWorkspaceVecString>,
     build: Option<StringOrBool>,
     metabuild: Option<StringOrVec>,
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -1659,7 +1659,6 @@ pub struct TomlPackage {
     edition: Option<MaybeWorkspaceString>,
     rust_version: Option<MaybeWorkspaceRustVersion>,
     name: InternedString,
-    #[serde(deserialize_with = "version_trim_whitespace")]
     version: MaybeWorkspaceSemverVersion,
     authors: Option<MaybeWorkspaceVecString>,
     build: Option<StringOrBool>,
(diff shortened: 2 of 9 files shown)


epage:
@rfcbot fcp merge

See the PR description for the change proposal

src/cargo/ops/registry/publish.rs
```diff
@@ -103,7 +103,7 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
         if allowed_registries.is_empty() {
             bail!(
                 "`{}` cannot be published.\n\
-                 `package.publish` is set to `false` or an empty list in Cargo.toml and prevents publishing.",
+                 `package.publish` is must be set to `true` or a non-empty list in Cargo.toml to publish.",
```
arlosi, line 106:
```suggestion
                 `package.publish` must be set to `true` or a non-empty list in Cargo.toml to publish.",
```

src/cargo/util/toml/mod.rs
```diff
@@ -1006,9 +1012,14 @@ impl TomlManifest {
         let publish = match publish {
             Some(VecStringOrBool::VecString(ref vecstring)) => Some(vecstring.clone()),
             Some(VecStringOrBool::Bool(false)) => Some(vec![]),
-            None | Some(VecStringOrBool::Bool(true)) => None,
+            Some(VecStringOrBool::Bool(true)) => None,
+            None => version.is_none().then_some(vec![]),
         };
 
+        if version.is_none() && publish != Some(vec![]) {
+            bail!("`package.version` must be specified if `package.publish = true`");
```
arlosi, line 1020: This error could be confusing if `publish = ['my-registry']`

epage: Hopefully the new message is a little better:

> `package.publish` requires `package.version` be specified

src/doc/src/reference/manifest.md
```diff
@@ -109,6 +109,8 @@ resolve dependencies, and for guidelines on setting your own version. See the
 [SemVer compatibility] chapter for more details on exactly what constitutes a
 breaking change.
 
+This field is optional and defaults to `0.0.0`.
```
weihanglo: Just a random idea. Wonder if we can have a more structural way to show the information of manifest key definition, such as how configuration doc currently looks like.

src/cargo/util/toml/mod.rs
```diff
@@ -1006,9 +1012,14 @@ impl TomlManifest {
...
             Some(VecStringOrBool::VecString(ref vecstring)) => Some(vecstring.clone()),
             Some(VecStringOrBool::Bool(false)) => Some(vec![]),
-            None | Some(VecStringOrBool::Bool(true)) => None,
+            Some(VecStringOrBool::Bool(true)) => None,
+            None => version.is_none().then_some(vec![]),
         };
 
+        if version.is_none() && publish != Some(vec![]) {
+            bail!("`package.publish` requires `package.version` be specified");
+        }
```
weihanglo, line 1021: This always bugs me. Now it looks like

* `publish = true` -> `None`
* not specified -> `[]`

which is a good thing as it exactly aligns to the logic here:

https://github.com/rust-lang/cargo/blob/87f4b1ba07f698e0a30d75ea0c9fbdb1e8655079/src/cargo/core/package.rs#L170-L175

I wonder if we could have a type wrapper to make the logic more clear with some methods.

weihanglo (commented):
Implementation-wise looks good to me.

weihanglo:
Not really on topic, but somehow relevant. <https://github.com/rust-lang/cargo/issues/5979> proposes to remove `package.version` filed in `Cargo.lock` for workspace members.

src/cargo/util/toml/mod.rs
epage: We likely should refactor this to use a custom enum, rather than `Option`.

src/doc/src/reference/manifest.md
epage: I feel like each approach works well for some cases and not for others.  Wish we could have a code fence with links.  I feel like that could allow the best of both worlds.

weihanglo (approved):
Reviewed again and looks good.

Feel free to `r=me` once FCP period ends.

src/doc/src/reference/manifest.md
```diff
@@ -470,23 +472,22 @@ if any of those files change.
...
-publish = false
+publish = ["some-registry-name"]
 ```
 
-The value may also be an array of strings which are registry names that are
-allowed to be published to.
-
+To prevent a package from being published to a registry (like crates.io) by mistake,
+for instance to keep a package private in a company,
+you can leave off the [`version`](#the-version-field) field.
```
WhyNotHugo, line 485: **Nitpick**: "leave off" sounds like you can set it to off/false.

```suggestion
you can omit the [`version`](#the-version-field) field.
```

weihanglo:
🎉 🎉 🎉 

@bors r+

PR #12788, opened 2023-10-07
Add unsupported lowercase `-z` flag suggestion for `-Z` flag 

hi-rustin:

Added unsupported lowercase `-z` flag suggestion for `-Z` flag.

Check out the unit test.

### Additional information

None

### DIFF
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -619,14 +619,7 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
                 .global(true),
         )
         .arg_config()
-        .arg(
-            Arg::new("unstable-features")
-                .help("Unstable (nightly-only) flags to Cargo, see 'cargo -Z help' for details")
-                .short('Z')
-                .value_name("FLAG")
-                .action(ArgAction::Append)
-                .global(true),
-        )
+        .arg_unstable_feature()
         .subcommands(commands::builtin())
 }
 
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -2,7 +2,7 @@ use anyhow::{anyhow, Context as _};
 use cargo::core::shell::Shell;
 use cargo::core::{features, CliUnstable};
 use cargo::{self, drop_print, drop_println, CargoResult, CliResult, Config};
-use clap::{Arg, ArgMatches};
+use clap::{builder::UnknownArgumentValueParser, Arg, ArgMatches};
 use itertools::Itertools;
 use std::collections::HashMap;
 use std::ffi::OsStr;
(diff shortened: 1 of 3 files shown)


hi-rustin (commented):
🔢 Self-check

tests/testsuite/build.rs
```diff
@@ -244,6 +244,33 @@ fn cargo_compile_directory_not_cwd() {
...
+    p.cargo("-zunstable-options -C foo build")
+        .masquerade_as_nightly_cargo(&["chdir"])
+        .cwd(p.root().parent().unwrap())
+        .with_stderr(
+            "\
+error: unexpected argument '-z' found
+
+  tip: a similar argument exists: '-Z'
+
+Usage: cargo [+toolchain] [OPTIONS] [COMMAND]
```
hi-rustin, line 264: Is displaying `cargo[EXE]` on Windows necessary?

src/cargo/util/command_prelude.rs
```diff
@@ -393,6 +393,28 @@ pub trait CommandExt: Sized {
         self._arg(unsupported_short_arg)
             ._arg(multi_opt("config", "KEY=VALUE", "Override a configuration value").global(true))
     }
+
+    fn arg_unstable_feature(self) -> Self {
```
epage, line 397: Why is this being moved to the prelude?  That is usually for when we need to share flag definitions but `-Z` is "global" meaning it being defined on the parent causes it to be inherited everywhere

tests/testsuite/build.rs
epage: Its necessary with clap when using the default generated usage because it uses `argv[0].file_name()`.  However, `cargo` overrides the usage to include the `[+toolchain]` and `<MANIFEST_RS>`

src/cargo/util/command_prelude.rs
hi-rustin: I want to put the real Arg and unsupported Arg together. 

Moved! Thanks!

tests/testsuite/build.rs
hi-rustin: I meant do we need to display it to get the same experience with clap?

epage: Likely not worth it.

src/bin/cargo/cli.rs
```diff
@@ -618,18 +618,31 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
...
+            .hide(true))
+        .arg(Arg::new("unstable-features")
+            .help("Unstable (nightly-only) flags to Cargo, see 'cargo -Z help' for details")
+            .short('Z')
+            .value_name("FLAG")
+            .action(ArgAction::Append)
+            .global(true))
         .subcommands(commands::builtin())
 }
-
```
epage, line 632: nit: why was this newline removed?

hi-rustin (commented):
Checked

src/bin/cargo/cli.rs
hi-rustin: Reverted.

Issue #10903, opened 2022-07-26
Respect `rust-version` when selecting a version for `cargo install`


epage:
When running `cargo install foo`, if my rust version is too old, I'll get compilation errors.  If the failing crate sets `rust-version`, then I'll get  a nice error message about it but no path forward for fixing it.  #10891 adds a path forward for `cargo install` when the failure is in a dependency but not when its in the crate itself that is being installed.

Ideally, `cargo install` will select an older version (with a warning) that is compatible with my toolchain.  As a step back from the ideal, I'd get an error with corrective steps (ideally, telling me what version to use).

epage:
This is blocked on `rust-version` being in the registry

8573:
While I generally very much support supporting older Rust versions, I worry about how this would (not silently but) quietly give users insecure versions of programs, in the absence of either integration with RustSec or a culture of yanking insecure versions (which I would hope would be blocked on having [a way to bypass yankedness](https://internals.rust-lang.org/t/cargo-allow-publishing-the-reason-for-yanking/16896/22) for e.g. `git bisect`).

epage:
Our choices are
- Error and let the user figure it out (current behavior)
- Error and tell the user how to resolve it
  - This is included as an alternative in this issue
  - This puts it in the users hands for deciding what is safe for them
  - We are still leaving a road hump along the users path to accomplishing their goal, souring their opinion of the ecosystem and possibly being the part that broke their will to continue experimenting with the solution (which will include the common refrain "why is it erroring instead of just doing it")
- Auto-select a compatible version
  - This is included as the primary proposed solution to this issue
  - The user might be surprised at the version used and lack of features
  - The crate version *might* have vulnerabilities that affect the user

Maybe I'm missing some use cases but I personally have little sympathy for the security side of this.  If the error just leads them to selecting an old version anyways, did we really help them with security?  How do we know the newest version is any more secure than the older version?  This also isn't too different than if someone had `cargo install`ed things previously.

cassaundra:
Thoughts I had from a discussion with Ed:

My inclination here is to have `cargo install` still error, but provide a more useful suggestion. I see a couple of problems with auto-selecting a compatible version:
[...]

PR #12798, opened 2023-10-10
fix(install): Suggest an alternative version on MSRV failure 

epage:

Moves users from a bad error message, suggesting `--locked` which won't do anything, to suggesting a version of a package to use instead.

The side benefit is errors get reported sooner
- Before downloading the `.crate`
- When installing multiple packages, before building the first

This comes at the cost of an extra `rustc` invocation.

Per-commit this builds it up, from tests to the final design.

### Additional information

This is also written in a way to align fairly well with how we'd likely implement #10903.
This improved error message will still be useful after that issue is resolved when the MSRV compatible version is outside of the version req.

### DIFF
--- a/src/cargo/ops/cargo_install.rs
+++ b/src/cargo/ops/cargo_install.rs
@@ -68,6 +68,7 @@ impl<'cfg> InstallablePackage<'cfg> {
         force: bool,
         no_track: bool,
         needs_update_if_source_is_index: bool,
+        current_rust_version: Option<&semver::Version>,
     ) -> CargoResult<Option<Self>> {
         if let Some(name) = krate {
             if name == "." {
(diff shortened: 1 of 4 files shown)


src/cargo/ops/cargo_install.rs
```diff
@@ -611,19 +620,46 @@ pub fn install(
...
+    let current_rust_version = if honor_rust_version {
+        let rustc = config.load_global_rustc(None)?;
+
+        // Remove any pre-release identifiers for easier comparison
+        let current_version = &rustc.version;
+        let untagged_version = semver::Version::new(
+            current_version.major,
+            current_version.minor,
+            current_version.patch,
+        );
```
epage, line 638: This was copied from `src/cargo/ops/cargo_compile/mod.rs`

src/bin/cargo/commands/install.rs
```diff
@@ -203,6 +203,7 @@ pub fn exec(config: &mut Config, args: &ArgMatches) -> CliResult {
             &compile_opts,
             args.flag("force"),
             args.flag("no-track"),
+            !args.flag("ignore-rust-version"),
```
ehuss, line 206: Is it possible to access `compile_opts.honor_rust_version` instead of adding another parameter?

ehuss:
Thanks!

@bors r+

PR #12799, opened 2023-10-10
feat: add package name and version to warning messages

### LINKED ISSUE #8018: Include package name/version in build script warnings.
In some situations, it is not clear where a warning is coming from when emitted from a build script. For example, a build script like this:

```
fn main() {
    println!("cargo:warning=this is a warning");
}
```

displays:

```
warning: this is a warning
```

I think it might be helpful to prefix the warning with the package name and version (similar to how `-vv` output is).  Perhaps something like this:

```
warning: [pkg_name 1.2.3] this is a warning
```

**Notes**
Warnings are emitted [here](https://github.com/rust-lang/cargo/blob/92d0ce80694fa888ff5b59e4323c1dfa6cdc2ba8/src/cargo/core/compiler/job_queue.rs#L852-L882). The package name/version should be accessible from the unit.

marchiore:

Hi, this PR Fixes #8018, where i add the package name and version on warning messages.
As this is my first contribution to the project, if anything is not in compliance in the PR let me know and I will be happy to correct it.

### DIFF
--- a/src/cargo/core/compiler/job_queue/mod.rs
+++ b/src/cargo/core/compiler/job_queue/mod.rs
@@ -952,7 +952,10 @@ impl<'cfg> DrainState<'cfg> {
                 }
 
                 for warning in output.warnings.iter() {
-                    bcx.config.shell().warn(warning)?;
+                    let warning_with_package =
+                        format!("{}@{}: {}", unit.pkg.name(), unit.pkg.version(), warning);
+
+                    bcx.config.shell().warn(warning_with_package)?;
                 }
 
                 if msg.is_some() {
--- a/tests/testsuite/build_script.rs
+++ b/tests/testsuite/build_script.rs
@@ -3778,8 +3778,8 @@ fn warnings_emitted() {
 [COMPILING] foo v0.5.0 ([..])
 [RUNNING] `rustc [..]`
 [RUNNING] `[..]`
-warning: foo
-warning: bar
+warning: foo@0.5.0: foo
+warning: foo@0.5.0: bar
 [RUNNING] `rustc [..]`
 [FINISHED] dev [unoptimized + debuginfo] target(s) in [..]
 ",
@@ -3816,7 +3816,7 @@ fn warnings_emitted_when_build_script_panics() {
     p.cargo("build")
         .with_status(101)
         .with_stdout("")
-        .with_stderr_contains("warning: foo\nwarning: bar")
+        .with_stderr_contains("warning: foo@0.5.0: foo\nwarning: foo@0.5.0: bar")
         .run();
 }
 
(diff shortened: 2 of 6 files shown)


src/cargo/core/compiler/job_queue/mod.rs
```diff
@@ -952,7 +952,9 @@ impl<'cfg> DrainState<'cfg> {
                 }
 
                 for warning in output.warnings.iter() {
-                    bcx.config.shell().warn(warning)?;
+                    let package = format!("[{} {}]", unit.pkg.name(), unit.pkg.version());
+                    let warning_with_package = format!("{} {}", package, warning);
```
hi-rustin, line 956: I think we can merge them together to get a better reading experience.

hi-rustin (commented):
Thanks!

src/cargo/core/compiler/job_queue/mod.rs
marchiore: done @hi-rustin

src/cargo/core/compiler/job_queue/mod.rs
```diff
@@ -952,7 +952,12 @@ impl<'cfg> DrainState<'cfg> {
                 }
 
                 for warning in output.warnings.iter() {
-                    bcx.config.shell().warn(warning)?;
+                    let warning_with_package = format!(
+                        "{} {}",
+                        format!("[{} {}]", unit.pkg.name(), unit.pkg.version()),
```
epage, line 957: nit: let's not embed `format!` within `format!`

epage, line 957: So this is my bad, I think I marked this as "Accepted" without giving this enough thought.

I had also considered whether this actually is needed since we generally don't show warnings for dependencies.  Should we instead cap these and then it becomes less ambiguous?  I don't know the historical context for this but most build scripts are for adapting code to the current system, so the warnings will be relevant to the end user, more than usual.

So assuming we go through with this, I feel like
- We should be brief to not take away from the warning message
- We should be unobtrusive to not take away from the warning message

For any dependency to warn, we likely can't skip on the version.  Maybe we could skip on both if its a workspace member but I'm unsure about the inconsistency.  So I think it can't really be more brief as-is

One way to make this less obtrusive is to do `{name}@{version}:` instead of `[{name} {version}]`.  The `[]` syntax  visually stops the eye.

src/cargo/core/compiler/job_queue/mod.rs
```diff
@@ -952,11 +952,9 @@ impl<'cfg> DrainState<'cfg> {
                 }
 
                 for warning in output.warnings.iter() {
-                    let warning_with_package = format!(
-                        "{} {}",
-                        format!("[{} {}]", unit.pkg.name(), unit.pkg.version()),
-                        warning
-                    );
+                    let warning_with_package =
+                        format!("{}@{} {}", unit.pkg.name(), unit.pkg.version(), warning);
```
epage, line 956: Could you put a `:` separator in there?  Seems like it'd help to distinguish this from the build scripts output.  I had it in my message but it can easily be lost in everything else I was saying :)

marchiore: I let this one slide, but now it's fixed :)

epage (approved):
Looks good.  Could you just clean up the commits for how you want them presented in the history and then I'll merge this.

marchiore:
> Looks good. Could you just clean up the commits for how you want them presented in the history and then I'll merge this.

done!

epage:
Thanks!

@bors r+

PR #12803, opened 2023-10-10
chore: Sort dependency tables

epage:
This will make `cargo add` put things in the right place.

Verified by doing `cargo add arcstr && git diff && git reset --hard`


weihanglo:
Got an impression that @Muscraft had a plan of writing a `tidy` for cargo.

@bors r+

weihanglo (approved):

PR #12806, opened 2023-10-11
fix(replace): Partial-version spec support

epage:

#12614 changed package ID specs to allow fields in the version number to be optional.  This earliest branch with this change is `rust-1.74.0` (beta).  While @Eh2406 was investigating version metadata issues in #12772, problems with the partial version change were found
- `replace`s that specify version metadata were ignored **(fixed with this PR)**
- `replace`s that don't specify version metadata when the `Cargo.lock` contained a version metadata, it would previously be ignored (with a warning) but now match **(unchanged with this PR)**
  - When the version metadata in `Cargo.lock` differed from the overriding `Cargo.toml`, cargo would panic **(now an error in this PR)**

The questions before are
- Do we revert #12614 in `master` and `rust-1.74.0` or merge this PR into `master`
- If we merge this PR into `master`, do we cherry-pick this into `rust-1.74.0` or revert #12614, giving ourselves more time to find problems

The initial commit adds tests that pass as of #12614.  Prior to #12614, these tests would have warned that the `replace` was unused and failed because `bar::bar` didn't exist.  Each commit then changes the behavior (or not) and updates the corresponding test.


Eh2406:
Thank you for the quick fix!

src/cargo/core/resolver/dep_cache.rs
```diff
@@ -173,11 +173,20 @@ impl<'a> RegistryQueryer<'a> {
...
-            assert_eq!(s.version(), summary.version());
-            assert_eq!(s.name(), summary.name());
+            assert_eq!(
+                s.name(),
+                summary.name(),
+                "dependency should be hard coded to have the same name"
+            );
+            if s.version() != summary.version() {
+                return Poll::Ready(Err(anyhow::anyhow!(
+                    "replacement specication `{}` matched {} and tried to override it with {}\n\
```
ehuss, line 183:
```suggestion
                    "replacement specification `{}` matched {} and tried to override it with {}\n\
```

src/cargo/core/resolver/dep_cache.rs
```diff
@@ -173,11 +173,20 @@ impl<'a> RegistryQueryer<'a> {
...
+                     avoid matching unrelated packaged by being more specific",
```
ehuss, line 184:
```suggestion
                     avoid matching unrelated packages by being more specific",
```

src/cargo/core/package_id_spec.rs
```diff
@@ -176,8 +176,7 @@ impl PackageIdSpec {
         }
 
         if let Some(ref v) = self.version {
-            let req = v.exact_req();
-            if !req.matches(package_id.version()) {
+            if !v.matches(package_id.version()) {
```
ehuss, line 179: Can you update the `matching` unittest below to include the new behavior?

src/cargo/util/semver_ext.rs
```diff
@@ -182,16 +182,16 @@ impl PartialVersion {
...
-        VersionReq {
-            comparators: vec![Comparator {
-                op: semver::Op::Exact,
-                major: self.major,
-                minor: self.minor,
-                patch: self.patch,
-                pre: self.pre.as_ref().cloned().unwrap_or_default(),
-            }],
-        }
+    pub fn matches(&self, version: &Version) -> bool {
```
ehuss, line 185: Can this method have a doc comment explaining that it matches build metadata, and that is a distinct decision that that caller needs to make whether or not that is what they want, since it is a subtle distinction?  (As compared to `Version::eq` versus `Version::cmp_precedence`, which is a subtle point that is easy to get wrong.)

ehuss (commented):
From what I can tell, this fixes much more than just `[replace]`, right?

For example, any command that takes a Package ID Spec (like `cargo build -p foo@1.2.3+123`), which on 1.73 (stable) will error if the metadata doesn't match, on beta it ignores it, and this PR fixes it to again error if it doesn't match, correct?

ehuss:
I'd be fine with backporting this to beta, since there is still about a month left. (Though reverting #12614 would also be fine, but I would lean towards just fixing it.)

epage:
> From what I can tell, this fixes much more than just `[replace]`, right?
> 
> For example, any command that takes a Package ID Spec (like `cargo build -p foo@1.2.3+123`), which on 1.73 (stable) will error if the metadata doesn't match, on beta it ignores it, and this PR fixes it to again error if it doesn't match, correct?

Yes, one of the bugs identifier and fixed in this (allowing different build metadata to still match) affects PackageIDSpecs everywhere.  I've updated the description to call this out.

In adding the unit tests, the fix for the above made it so `foo@1.2` will now match `foo@1.2.3-alpha.0`.  I'll call that out in the description as well so we can more explicitly decide if we want to special case pre-release or not.

epage:
We talked about this in the cargo team meeting.  We were fine with the change in
 behavior and with backporting this
- This required a cargo expert to detect the bug and was not found through users using nightly or beta
- With package id specs and replaces (deprecated) combined with build metadata and pre-release, the bast radius if a bug is fairly small and we can live with it broken in a release
- Though we were probably comfortable with `1.2.3` matching `1.2.3-alpha.0`, we decided to require opting into pre-releases, like version requirements, so we have more room for changing things if someone takes a broad view of pre-release and works to solve it.  d318ed4  has this change

Eh2406:
Looks good, and if we missed something then new PR.
@bors r+

Issue #12271, opened 2023-06-14
Custom registries are infectious to downstream dependents


w4:
### Problem

When depending on custom registries over Git, the consumer is currently expected to also have a reference to the same registry in their `.cargo/config.toml`.

### Steps

1. crate-a containing:
   ```
   $ cat Cargo.toml
   [package]
   name = "crate-a"
   version = "0.1.0"
   edition = "2021"

   [dependencies]
   some-crate = { version = "0.1", registry = "some-registry" }

   $ cat .cargo/config.toml
   [registries]
   some-registry = { index = "ssh://registry.x.com/path/to/some-registry" }
   ```
2. crate-b containing:
   ```
   $ cat Cargo.toml
   [package]
   name = "crate-b"
   version = "0.1.0"
   edition = "2021"

   [dependencies]
   crate-a = { version = "0.1", git = "https://git.x.com/path/to/crate-a" }

   $ cat .cargo/config.toml
   cat: .cargo/config.toml: No such file or directory
   ```
3. Compiling `crate-b` gives the cryptic error:
    ```
    error: no matching package named `crate-a` found
    location searched: https://git.x.com/path/to/crate-a
    ```
4.
    Which can be fixed by adding
    ```
    [registries]
    some-registry = { index = "ssh://registry.x.com/path/to/some-registry" }
    ```
    to crate-b's .cargo/config.toml

### Possible Solution(s)

Potentially a flag that can be set in the dependent's `.cargo/config.toml` to allow git dependencies  to use arbitrary registries?

w4:
That's an interesting revelation - thanks, I didn't actually test transient dependencies while going through the registry.

Although, now that I'm testing with that knowledge, that brings up a surprising behaviour that the crate-a's `registry = "some-registry"` definition leaks into crate-b and requires crate-a to define a registry with the exact name `some-registry` rather than either:

a. ignoring it because it's going to use the path dependency anyway (see #12272), or
b. using the remote crate's .cargo/config.toml `[registries]` definition for the build of that crate

Eh2406:
`.cargo/config.toml` are never read from dependencies. They are only read from the dir where cargo was invoked (and its parents). They are configuration for the build, not for the package. If you use `--manifest-path` you can find a different example of this https://github.com/rust-lang/cargo/issues/2930 Ferthermore, there is no guarantee that the maintainers of `crate-a` even checked in there config.toml. They may have put there `[registries]` section in `~/.cargo/config.toml` for example.

[...]

Issue #12272, opened 2023-06-14
Adding a `registry` and `version` to a local dependency causes dependents over Git to attempt to pull the registry by name


w4:
### Problem

When moving crates to registries that were previously depended on via Git, adding a `registry` definition to a local dependency causes dependents to attempt to pull the registry, rather than ignoring it and using the local path.

### Steps

1. crate-a containing:
   ```
   $ cat Cargo.toml
   [package]
   name = "crate-a"
   version = "0.1.0"
   edition = "2021"

   [dependencies]
   some-crate = { path = "../some-crate", registry = "x", version = "0.1" }
   ```
2. crate-b containing:
   ```
   $ cat Cargo.toml
   [package]
   name = "crate-b"
   version = "0.1.0"
   edition = "2021"

   [dependencies]
   crate-a = { git = "https://git.x.com/path/to/crate-a" }
   ```
3. Attempting to compile `crate-b` attempts to pull the `x` registry rather than ignoring it since it is overridden by the relative path.

When combined with #12271, this prevents any sort of slow migration to registries -- requiring a big bang update to all crates which is untenable for any mid to large sized organisation.

### Possible Solution(s)

When depending on a crate using `{ git = "https://git.x.com/path/to/repo" }`, the local dependency should be preferred over the one in the registry.

weihanglo:
I am sorry to hear this.

If you could provide more details or the exact steps for reproduction, it would be helpful, such as:

* How did you install rust/cargo on Debian 12?
* From which package repository you to install rust/cargo?
* What commands you ran after that?
* How did you remove rust/cargo?

Please let us know if you manage to reproduce it, and vice verse.

weihanglo:
Wrong thread. Just deleted the comment. Sorry for the noise.

weihanglo:
Copied from <https://github.com/rust-lang/cargo/issues/12271#issuecomment-1759935393>, since this is the right place for the comment. 

---

Note that currently if you have both `path` and `registry` declared in your direct dependency and the registry config is missing, Cargo will error out and stop the build.


There are several scenario of them, and we may only need to change the behavior for git and path dependencies if we really want it.

* ✅ direct dependency -> transitive dependency with both `path` and `registry`: shouldn't exist on crates.io. No change needed.
* ✅ direct `registry` dependency -> transitive dependency with both `path` and `registry`: Cargo will stick to registry dependencies transitively. No change needed.
[...]

PR #12807, opened 2023-10-11
Ignore errors when declaring a registry on path dependency

w4:
When depending on crate A from crate B, if crate A declares a `registry` on a path local dependency, cargo would attempt to resolve the registry name from crate B's config.toml, even though it would never be used and likely doesn't exist there.

It may be wishful thinking to open a PR for this in its current form, but it fixes the exact issue outlined in #12271. I'm happy to modify it if there's a more "correct" fix out there.

### Environment

```toml
CMX2XM in cargo-issue-repro
❯ \cat ws/Cargo.toml
[workspace]
resolver = "2"
members = ["ws1", "ws2"]

CMX2XM in cargo-issue-repro
❯ \cat ws/ws1/Cargo.toml
[package]
name = "ws1"
version = "0.1.0"
edition = "2021"

[dependencies]

CMX2XM in cargo-issue-repro
❯ \cat ws/ws2/Cargo.toml
[package]
name = "ws2"
version = "0.1.0"
edition = "2021"

[dependencies]
ws1 = { path = "../ws1", registry = "a" }

CMX2XM in cargo-issue-repro
❯ \cat leaf/Cargo.toml
[package]
name = "leaf"
version = "0.1.0"
edition = "2021"

[dependencies]
ws2 = { git = "file:///Users/jordan/Code/cargo-issue-repro/ws" }
```

### Before

```
❯ CARGO_LOG=trace cargo check
...
2023-10-11T20:47:59.716458Z  INFO cargo::ops::cargo_read_manifest: skipping malformed package found at `/Users/jordan/.cargo/git/checkouts/ws-a3d765f02ebb7e68/6b8487d/ws2`
...
error: no matching package found
searched package name: `ws2`
perhaps you meant:      ws1
location searched: file:///Users/jordan/Code/cargo-issue-repro/ws#6b8487d6
required by package `leaf v0.1.0 (/Users/jordan/Code/cargo-issue-repro/leaf)`
```

### After

```
❯ /Users/jordan/Code/cargo/target/debug/cargo check
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
```

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -2523,8 +2523,19 @@ impl<P: ResolveToPath + Clone> DetailedTomlDependency<P> {
             .set_optional(self.optional.unwrap_or(false))
             .set_platform(cx.platform.clone());
         if let Some(registry) = &self.registry {
-            let registry_id = SourceId::alt_registry(cx.config, registry)?;
-            dep.set_registry_id(registry_id);
+            let registry_id = SourceId::alt_registry(cx.config, registry);
+
+            match registry_id {
+                Ok(registry_id) => {
+                    dep.set_registry_id(registry_id);
+                }
+                Err(_) if cx.source_id.is_git() && self.path.is_some() => {
+                    // if the crate `cx` depended on us via git and we're
+                    // resolving a path dependency, ignore any errors for
+                    // the registry being unknown.
+                }
+                Err(e) => return Err(e.into()),
+            }
         }
         if let Some(registry_index) = &self.registry_index {
             let url = registry_index.into_url()?;


weihanglo:
Thanks for being interested in this.

For direct dependencies, if the registry configuration is missing, Cargo errors out. It seems a bit unfortunate that the behavior of transitive and direct dependencies can't be consistent after this changes.

I'll propose this to the Cargo team for discussion. Before that, if you have any more thoughts please comment on the issue instead.

w4:
Thanks @weihanglo. I do agree the inconsistency would preferably be avoided (though even now it's still there - direct dependencies against an unknown registry give a clear error, transitive dependencies silently discard the crate from the workspace). Maybe hiding this code behind a config flag would make this change more easily digestable?

epage:
btw you link to #12271 but I think you meant #12272

weihanglo:
Close as the current behaviour is expected. See https://github.com/rust-lang/cargo/issues/12271#issuecomment-1777640307 and https://github.com/rust-lang/cargo/issues/12272#issuecomment-1777906610 for more.

Do let us know if there is something impossible to be resolved by those approaches mentioned. Thank you.

weihanglo:
To move forward, we need some more tests. This is how people usually do :

* The first commit will have tests demonstrating the current bad behavior.
* The following commits will fix the bad behavior and the tests.

That make the intention clear when reviewing or reading the Git commit history.

You might want to add some new tests around this [`registry_incompatible_with_git` test](https://github.com/rust-lang/cargo/blob/9bf67a1026c023e2789e867d8b88dc6fc181f357/tests/testsuite/alt_registry.rs#L219-L247). Feel free to ping me if you have questions.

epage:
As this has been waiting-on-author for quiet a while, I'm going to go ahead and close this.  Feel free to resurrect these changes!

PR #12814, opened 2023-10-12
contrib docs: update rfc and roadmap links

tshepang:


weihanglo (commented):
Thanks.

I feel like it was misplaced under rust-lang org. It should be a repo-level project not org-level.
@ehuss, do you know why it became like this and what's your plan in mind?

epage:
I wonder if this is a old vs new projects thing.  When I migrated to the new projects system, i wasn't given any choice of what it did (granted, migration itself is still in preview)

weihanglo:
I see. I've prefixed all of them with "Cargo". Thank you.

Issue #12818, opened 2023-10-13
`cargo metadata`: expose checksum in packages


justahero:
### Problem

When creating a Software Bill of Materials (SBOMs) an important aspect is to verify the integrity of the downloaded packages. The process of fetching information using `cargo` to build SBOMs for Rust crates / dependencies is quite cumbersome & error prone at the moment. One important piece of information is the `checksum` of a downloaded crate to verify that the source has not been modified.

### Proposed Solution

The `cargo metadata` command should expose the `checksum` field in its JSON output for packages. To my understanding checksums are available for downloaded packages from the registry (e.g. crates.io). Compared to `cargo` as a library the output of `cargo metadata` is considered stable, it's also a recommended option for [3rd party tools](https://doc.rust-lang.org/nightly/cargo/reference/external-tools.html). Adding a new field is assumed to not break backwards compatibility.

### Notes

This issue is based on [this discussion](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/cargo.3A.20Best.20API.20to.20get.20resolved.20dependency.20tree) in the `t-cargo` Zulip stream. One current option to get the `checksum` value is to read the `Cargo.lock` file. Unfortunately to collect all information this requires mapping of information from separate sources, including the output of `cargo metadata`. Another option is to use `cargo` as a library, which is deemed unstable & is strongly discouraged.

I ran some initial checks to learn how the total size of the JSON output is affected. Exposing checksums increases the output size of `cargo metadata` by about 1.5%. For example the output for the `cargo` repository went from `1.559.278` (master) to `1.586.508` bytes (w/ checksums). Other projects have a comparable increase in size.

epage:
When talking with @lfrancke further, there were questions about *what* checksum format different SBOM formats would want.  Us providing *a* checksum format might not be as helpful.

The other challenge is if we might be wanting to adjust checksums, should this wait on that?
- #2526 is about supporting more compression formats and we currently checksum the compressed version, making us need to support more checksum formats
- #7212 is about providing a more stable, reproducible checksum.  The proposed solution would add one more checksum field to deal with for the rest of cargo but it would mean #2526 wouldn't add any further checksum fields.

justahero:
Thanks for your feedback.

[...]

PR #12819, opened 2023-10-13
Expose checksum in `cargo metadata` command

justahero:

This change exposes the `checksum` field in packages for the `cargo metadata` command.

See #12818 

Run `cargo metadata` to see that the `checksum` field is present. Tests have been updated with existing calculated checksums to reflect the change.

### Additional information

When running `cargo test` locally there are 2 tests failing, in particular

* [`alt_registry::alt_reg_metadata`](https://github.com/rust-lang/cargo/blob/master/tests/testsuite/alt_registry.rs#L810)
* [`alt_registry::unknown_registry`](https://github.com/rust-lang/cargo/blob/master/tests/testsuite/alt_registry.rs#L1122)

Running them on their own, e.g. `cargo test alt_registry::alt_reg_metadata` passes the test.

I will try to pinpoint the failing tests, maybe someone has an idea?

For example running test file `cargo test alt_registry` repeatedly will almost always fail the `unknown_registry` test. With a lower frequency it will additionally fail `alt_reg_metadata`.


epage (changes requested):
Going to post my comments on #12818 as this has more to do with "what should be done" than "how"

justahero:
Let's close this one until a solid approach to generate usable checksums is discussed.

PR #12829, opened 2023-10-15
Print environment variables for build script executions with `-vv`

Nilstrieb:

When debugging complicated builds (I was trying to figure out how  `cargo-miri` cross-compiles compiler_builtins without needing a C cross compiler), it's useful to see all the environment variables passed to the build script.

This is also consistent with other commands.

I tested it locally by creating a small crate with an empty `build.rs` and building it. Additionally, a test is included.


weihanglo (approved):
Looks reasonable. Thank you!

I am going to merge this. If people have concern with it, we can tweak or revert it then.

PR #12832, opened 2023-10-17
doc(cargo-login): mention args after `--` in manpage

weihanglo:

We failed to mention `-- args` in manpage. Also `--` should be quoted with backticks.


ehuss:
Thanks!

@bors r+

PR #12834, opened 2023-10-17
ci/contrib: do not fail on missing gh-pages

dvdhrm:
The current contrib deploy-hook fails if there is no `gh-pages` branch. Change the CI order to disregard the old `gh-pages` branch first.

The `contrib` deploy-hook always creates a fresh `gh-pages` commit and pushes it out. However, currently it relies on the old `gh-pages` branch to exist, since it does not ignore errors when pruning it. Fortunately, the code always creates a new orphan branch, since it does not want to keep history for deployments. Therefore, we can simply use:

    `git worktree --orphan -B <branch> <path>`

This will ensure to always create an orphan branch named `<branch>`, and override an existing branch if it exists (see `-b` vs `-B`). Hence, there is no need for us to prune the old branch, anymore.

Since we will recreate the branch on every push, we have to explicitly specify the remote to push to. We no longer set up branch tracking.

Note that running github-actions in a private fork is quite useful to avoid constantly pushing changes to a PR and triggering notifications. Unfortunately, private forks on github do not automatically clone the gh-pages branch. This PR fixes annoying CI failures when updating the master branch on private forks, since github does not automatically clone the gh-pages branch.

### DIFF
--- a/.github/workflows/contrib.yml
+++ b/.github/workflows/contrib.yml
@@ -25,14 +25,12 @@ jobs:
       run: |
         cd src/doc/contrib
         mdbook build
-        git worktree add gh-pages gh-pages
+        # Override previous ref to avoid keeping history.
+        git worktree add --orphan -B gh-pages gh-pages
         git config user.name "Deploy from CI"
         git config user.email ""
         cd gh-pages
-        # Delete the ref to avoid keeping history.
-        git update-ref -d refs/heads/gh-pages
-        rm -rf contrib
         mv ../book contrib
         git add contrib
         git commit -m "Deploy $GITHUB_SHA to gh-pages"
-        git push --force
+        git push origin +gh-pages


.github/workflows/contrib.yml
```diff
@@ -25,14 +25,12 @@ jobs:
       run: |
         cd src/doc/contrib
         mdbook build
-        git worktree add gh-pages gh-pages
+        # Override previous ref to avoid keeping history.
+        git worktree add --orphan -B gh-pages gh-pages
```
weihanglo, line 29: `--orphan` in `git worktree add` is a relatively new feature added in Git 2.42. I just confirmed that [GitHub hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#preinstalled-software) already pre-installed that version of Git. 👍🏾

weihanglo, line 29: Could you verify that `gh-pages` doesn't exist in this environment? I did a rudimentary test and got this error:


```console
$ git worktree add --orphan -B gh-pages gh-pages
Preparing worktree (resetting branch 'gh-pages'; was at 8eb8acb)
fatal: a branch named 'gh-pages' already exist
```

dvdhrm: `-B` (rather than `-b`) explicitly overrides existing branches. I tested this on github-actions on my private fork and it correctly overrides existing branches. Did you get this result on a local test or on github-actions?

weihanglo: > `-B` (rather than `-b`) explicitly overrides existing branches.

Yeah that's what I supposed. However, it didn't behave that way if my local repo has an existing `gh-pages`.

I'll merge this as is and see if it works. Thank you.

weihanglo (approved):

.github/workflows/contrib.yml
dvdhrm: Indeed. If you have a local branch, it fails. I guess the reason it worked in the workflow is because the `gh-pages` branch is not synced, so only exists as `origin/gh-pages`.

This might be a bug in git where `--orphan` and `-B` don't work together.

dvdhrm:
The post-merge workflow ran successfully:

https://github.com/rust-lang/cargo/actions/runs/6549199896

weihanglo:
Ok it does need an old gh-pages to continue. Until we find a better solution I'll revert this for now.

PR #12835, opened 2023-10-17
ci/contrib: use separate concurrency group

dvdhrm:
Run the contrib workflow in its own concurrency group to avoid multiple runs causing conflicting git pushes.

The contrib workflow pushes to the gh-pages branch, and thus is not safe to be run multiple times in parallel. Given that github-actions can stall workflow runs for quite some time, multiple executions can overtake each other. By using concurrency groups, all workflow runs will wait until previous runs completed.

Given that the workflow uses orphan-branches and forced pushes, conflicting workflows will not fail, but might produce outdated data if they overtake each other.

We explicitly disable cancellation to get better diagnostics and more easily find the first possible run that failed. Since these workflow runs are rather fast, and only run on master, cancellation seems not necessary.

### DIFF
--- a/.github/workflows/contrib.yml
+++ b/.github/workflows/contrib.yml
@@ -4,6 +4,10 @@ on:
     branches:
       - master
 
+concurrency:
+  cancel-in-progress: false
+  group: "gh-pages"
+
 permissions:
   contents: read
 


dvdhrm:
@weihanglo was already reviewing my other change to the same workflow. Maybe it makes sense to assign this to them as well? (see #12834)

weihanglo:
The best way to override assigned reviewers is using `r?`, i.e r? weihanglo

dvdhrm:
> I can't recall there was any conflict happened on master, since `bors` is served as a merge queue manager for `rust-lang/cargo` and protect from that angle. Could you say more about what you've experienced makes you do this change?

Yeah, this is again mostly about private forks. I sometimes push to master on private forks, which triggers this workflow (even in parallel, since there is no bors), and then I can share links to the generated docs of some proposed changes.

This is not crucial at all, but I thought kinda nice to have and untangles the workflow from bors.

dvdhrm:
I mean, we could argue that there are better ways to render the docs and share them. But the workflow is already there and works, so I thought why not fix it. No strong opinions on my side, though.

weihanglo (approved):
Valid. Thanks!

weihanglo:
It doesn't work <https://doc.crates.io/contrib/>. The tree has one more level of directory <https://github.com/rust-lang/cargo/tree/gh-pages>.

@dvdhrm could you help fix or revert it?

PR #12836, opened 2023-10-17
docs(contrib): Policy on manifest editing

epage:
Based on https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Automatic.20edits.20to.20.60Cargo.2Ctoml.60

PR #12837, opened 2023-10-17
fix(remove): Preserve feature comments

### LINKED ISSUE #11743: `cargo add` reformats the features lists
### Problem

`cargo add` does not preserve formatting of TOML lists unrelated to the `[dependencies]` table

### Steps

With the following `Cargo.toml`:

```toml
[package]
name = "xxx"
version = "0.1.0"
edition = "2021"

[dependencies]

[features]
default = [
  "a",
  "b",
  "c",
]
a = []
b = []
c = []
```

Run:

```console
$ cargo add directories
    Updating crates.io index
      Adding directories v4.0.1 to dependencies.
```

observe that the `[features]` table is reformatted:

```diff
diff --git i/Cargo.toml w/Cargo.toml
index ccae1b6..61c1383 100644
--- i/Cargo.toml
+++ w/Cargo.toml
@@ -6,13 +6,10 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

 [dependencies]
+directories = "4.0.1"

 [features]
-default = [
-  "a",
-  "b",
-  "c",
-]
+default = ["a", "b", "c"]
 a = []
 b = []
 c = []
```

### Possible Solution(s)

Preserve formatting of all TOML contents outside of the newly added dependency.

### Notes

_No response_

### Version

```text
cargo 1.67.1 (8ecd4f20a 2023-01-10)
release: 1.67.1
commit-hash: 8ecd4f20a9efb626975ac18a016d480dc7183d9b
commit-date: 2023-01-10
host: x86_64-apple-darwin
libgit2: 1.5.0 (sys:0.16.0 vendored)
libcurl: 7.86.0 (sys:0.4.59+curl-7.86.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 13.2.1 [64-bit]
```

epage:

We've been having a hard time balancing leaving the feature list in a good looking start and preserving formatting.  With our new formatting policy (#12836), we can just choose to preserve formatting instead.

Fixes #11743

The first commit copies an existing test.  The second is where the fun begins, customizing the test for some weird cases.  The follow up commits do the slow walk for improving it.

We ended up preserving some line-trailing comments because they come after the comma and toml_edit treats that as part of the prefix of the next item.  Tracking removal of that was going to require us to determine if the newline existed in the suffix or in the next item's prefix and edit accordingly and I decided to skip that to keep this initial implementation simpler.

### Additional information


src/cargo/util/toml_mut/manifest.rs
```diff
@@ -540,5 +540,42 @@ fn non_existent_dependency_err(
...
+        merged_lines.push_str(suffix_lines);
+        merged_lines.push('\n');
+    }
+
+    let next_index = index; // Since `index` was removed, that effectively auto-advances us
+    if let Some(next) = array.get_mut(next_index) {
+        let next_decor = next.decor_mut();
+        let next_prefix = next_decor
+            .prefix()
+            .map(|s| s.as_str().expect("spans removed"))
```
weihanglo, line 572: question for `toml_edit`. Since `RawString` is opaque to users, what does "span removed" mean to them (or me, I don't quite follow)?

tests/testsuite/cargo_remove/multiple_dev/out/Cargo.toml
```diff
@@ -17,4 +17,4 @@ toml = "0.1"
 clippy = "0.4"
 
 [features]
-std = ["semver/std"]
+std = [ "semver/std"]
```
weihanglo, line 20: This isn't addressed. Should we?

epage: This is the cost of this change and the new policy.  The user had a space before the entry, so we preserved that.

Every once in a while, I'm tempted to change `toml_edit` so it detects "default" formatting and preserves that notion so that when an edit happens, it continues to use the default.

src/cargo/util/toml_mut/manifest.rs
epage: When `toml_edit` does its parsing pass, `RawString`s are just spans (`Range<usize>`).  This allows `toml` to never allocate for decor as it just gets abandoned. We then replace all of these spans with the actual content when generating a `toml_edit::Document` so you can copy a `Value` out of one document and insert it into another without getting bounds issues.  Since its unused, we save on memory by tracking this in an enum. 

We wouldn't want to change out the internal string type when switching away from spans because that would require re-allocating everything around the string (the array, table, etc).

We wouldn't want to have a separate public/private API that makes these assumptions for the user because then we need to make sure the public isn't used internally.

So this gets exposed to users.  In short, `RawString` is too low level to know of the details of when its safe but the caller easily does.

tests/testsuite/cargo_remove/multiple_dev/out/Cargo.toml
weihanglo: Let me know when we get it in toml_edit.

weihanglo (approved):

PR #12838, opened 2023-10-17
fix(add): Preserve more comments

### LINKED ISSUE #10850: Comments aren't preserved when modifying an existing stanrdard toml table with `cargo add`
### Problem

`cargo add` can modify existing entries, like adding more features.  When using a standard table (as opposed to an inline table, see #10849), comments get deleted.

### Steps

*Note: steps are using this is a test case in cargo and these crates don't normally exist*

With `Cargo.toml`:
```toml
[workspace]

[package]
name = "cargo-list-test-fixture"
version = "0.0.0"

[dependencies.your-face]
version="99999.0.0"  # Hello world
features=["eyes"]  # Goodbye moon
```
run
```console
$ cargo add your-face -F features nose
```

and you'll get
```toml
[workspace]

[package]
name = "cargo-list-test-fixture"
version = "0.0.0"

[dependencies.your-face]
version = "99999.0.0"
features = ["eyes"]
```

### Possible Solution(s)

There are two problems
- We drop the existing formatting
- We explicitly call `table.fmt()` which deletes all decor

The reason for the latter is so the trailing space in inline tables is removed when an item is no longer the last.

### Notes

_No response_

### Version

_No response_

epage:

This was fixed in killercup/cargo-edit#725 and I forgot to carry it over to here.

I kept in some auto-formatting because there is little to preserve until the TOML 1.1 spec is out which will allow mult-line inline-tables which means there might be comments interspersed.  We'll see which comes first, `cargo fmt` support for `Cargo.toml` or the 1.1 spec.

Fixes #10850

First commit adds a test demonstrating the problem.  The second makes a lot of noise that the third cleans up, so its a mixed bag as to which commit to look at.

### Additional information

### DIFF
--- a/src/cargo/util/toml_mut/dependency.rs
+++ b/src/cargo/util/toml_mut/dependency.rs
@@ -464,7 +464,7 @@ impl Dependency {
         } else if let Some(table) = item.as_table_like_mut() {
             match &self.source {
                 Some(Source::Registry(src)) => {
-                    table.insert("version", toml_edit::value(src.version.as_str()));
+                    overwrite_value(table, "version", src.version.as_str());
 
                     for key in ["path", "git", "branch", "tag", "rev", "workspace"] {
                         table.remove(key);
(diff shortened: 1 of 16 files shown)


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #12840, opened 2023-10-17
fix(cli): Suggest cargo-search on bad commands

### LINKED ISSUE #4682: missing cargo command suggestions
Inspired by the various cargo subcommands available in the broader crates.io ecosystem: we should consider adding a mechanism to automatically search for Cargo subcommands if someone enters an unknown subcommand. For instance, `cargo tree` could suggest installing [cargo-tree](https://crates.io/crates/cargo-tree). This could make use of metadata provided by crates.io.

cc @carols10cents

epage:
This is a low-tech solution alternative to the options proposed in #4682
- Search `[[bin]]`s within all packages in the registry (which aren't tracked atm), suggesting to `cargo install` what is found
- Check if `cargo-<cmd>` is in the registry, suggesting `cargo install if it is

By suggesting `cargo search`, we are giving them a tool so they can verify if the package is what they want (a `cargo info` would help as a next step).

Is is needed?
- New users might not know of `cargo search` but they can search on crates.io
- New users might not be aware of the `cargo-<cmd>` naming pattern

Seems like this can still offer some benefit.

Fixes #4682

### DIFF
--- a/src/bin/cargo/main.rs
+++ b/src/bin/cargo/main.rs
@@ -192,10 +192,9 @@ fn execute_external_subcommand(config: &Config, cmd: &str, args: &[&OsStr]) -> C
                 let did_you_mean = closest_msg(cmd, suggestions.keys(), |c| c);
 
                 anyhow::format_err!(
-                    "no such command: `{}`{}\n\n\t\
-                    View all installed commands with `cargo --list`",
-                    cmd,
-                    did_you_mean
+                    "no such command: `{cmd}`{did_you_mean}\n\n\t\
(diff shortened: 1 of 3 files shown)


epage:
I did this as an experiment and lean towards it offering enough benefit to be worth merging but fine closing it if people think it is of questionable value.

hi-rustin (approved):
Looks good to me.

weihanglo (approved):

weihanglo:
Thanks!

@bors r=hi-rustin

Issue #4830, opened 2017-12-18
The --bin flag on the install subcommand doesn't work


RazrFalcon:
```bash
% cargo install --force --bin svgcleaner --git https://github.com/RazrFalcon/svgcleaner               
    Updating git repository `https://github.com/RazrFalcon/svgcleaner`
error: multiple packages with binaries found: docgen, files-testing, stats, svgcleaner, use_as_lib
```

ishanjain28:
All the packages are installed using [install](https://github.com/rust-lang/cargo/blob/83912a083ea3111fc2294dfea87bae5f4687a4a2/src/cargo/ops/cargo_install.rs#L55) function in `src/cargo/ops/cargo_install.rs`. This function takes a `krate: Vec<&str>` which is created from `options.arg_crate` in `execute` function in `src/bin/install.rs`.


In the command,

    cargo install --force --bin svgcleaner --git https://github.com/RazrFalcon/svgcleaner               

`options.arg_crate` is an empty vector. The value of `--bin` flag(`svgcleaner`) is used to generate `CompileFilter` but never used. So, You are getting this error. 

@RazrFalcon,
For your case, You can use `cargo install --force svgcleaner --git https://github.com/RazrFalcon/svgcleaner` and it'll work just fine. I'll try to send a PR to fix this before Monday.

RazrFalcon:
@ishanjain28 thanks, that worked.

kornelski:
I've ran into the same problem:

```sh
cargo install --bin cargo-deb --git https://github.com/mmstick/cargo-deb --force
    Updating git repository `https://github.com/mmstick/cargo-deb`
error: multiple packages with binaries found: cargo-deb, example
```

I see from this issue that skipping `--bin` solves the problem. However, I wouldn't have figured it out myself, because:

* I assumed `cargo install foo` refers to `foo` on crates.io *only*, and `--git foo` is a replacement for the package name, not an independent argument.
* Other `cargo` commands use `-p` to select the package, but not this one.
* `cargo install [OPTIONS] [--] [crate]...` says **crate**, but the error is about **packages**.
* `cargo install --help` has "--bin <NAME>...          Install only the specified binary" which sounded exactly like what I wanted.

Things that could help:

* Add help text to the error message suggesting adding crate name as the first arg
* When `--bin` is specified, but not the crate/package, either search all of them, or show an error explaining specifically why `--bin` was ignored.
* Support `-p, --package` to specify the package for consistency with other commands.

[...]

PR #12841, opened 2023-10-17
fix(help):Clarify install's positional 

epage:

- That a version is accepted
- That you are selecting from the source a package which led to part of
  the confusion in #4830

I wonder if we should rename our `CRATE` value names to `PKG`/`PACKAGE`

While doing this, I decided to fix the inconsistency in how we handle value names in help.

### Additional information


tests/testsuite/cargo_test/no_keep_going/stderr.log
```diff
@@ -2,6 +2,6 @@ error: unexpected argument '--keep-going' found
 
   tip: use `--no-fail-fast` to run as many tests as possible regardless of failure
 
-Usage: cargo[EXE] test [OPTIONS] [TESTNAME] [-- [args]...]
```
hi-rustin, line 5: This change looks weird. It breaks the CI.

hi-rustin (commented):
Thanks!

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #12842, opened 2023-10-18
docs: remove review capacity notice

weihanglo:
The Cargo project is in a better state than it was when maintainers added this notice. I believe it's time to take it down, as maintainers have got more time on it, and more people are willing to contribute to the project constantly.

Note that this doesn't mean Cargo start accepting arbitrary pull requests or features. The reviewers are still a small handful of people. The guideline never change — issue, discuss, then pull request.

### DIFF
--- a/src/doc/contrib/src/process/index.md
+++ b/src/doc/contrib/src/process/index.md
@@ -8,11 +8,6 @@ process.
 
 Please read the guidelines below before working on an issue or new feature.
 
-**Due to limited review capacity, the Cargo team is not accepting new features
-or major changes at this time. Please consult with the team before opening a
-new PR. Only issues that have been explicitly marked as accepted will be
-reviewed.**
-
 [Working on Cargo]: working-on-cargo.md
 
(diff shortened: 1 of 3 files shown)


src/doc/contrib/src/process/index.md
```diff
@@ -8,11 +8,6 @@ process.
 
 Please read the guidelines below before working on an issue or new feature.
 
-**Due to limited review capacity, the Cargo team is not accepting new features
-or major changes at this time. Please consult with the team before opening a
-new PR. Only issues that have been explicitly marked as accepted will be
-reviewed.**
```
epage, line 14: I feel like this part of the message is buried and indirect in the other sections.  Should we find a way to elevate it?

weihanglo: Updated with a link to the `S-accepted` label. What do you think?

epage: Thanks!

PR #12846, opened 2023-10-18
docs(contrib): generate redirection HTML pages in CI

weihanglo:

#12834 wiped out the old content of all link redirections. From my understanding it relies on the existing gh-pages hosting those files.

This PR pulls out the [`generate.py`](https://github.com/rust-lang/cargo/blob/cff41259837c912b44f2eba98777d71996740fe0/generate.py) from older `gh-pages` branch, and run it during the CI deploy job.

### How to review

It succeeds on my fork <https://github.com/weihanglo/cargo/tree/gh-pages> ([CI log](https://github.com/weihanglo/cargo/actions/runs/6564128361/job/17829881499#step:4:1)).

I don't think `.lock` or [`.nojekyll`](https://github.blog/2009-12-29-bypassing-jekyll-on-github-pages/) is needed. If it causes problem, we can add them back.

### DIFF
--- a/.github/workflows/contrib.yml
+++ b/.github/workflows/contrib.yml
@@ -27,8 +27,11 @@ jobs:
         echo `pwd`/mdbook >> $GITHUB_PATH
     - name: Deploy docs
       run: |
+        GENERATE_PY="$(pwd)/ci/generate.py"
+
         cd src/doc/contrib
         mdbook build
+
         # Override previous ref to avoid keeping history.
         git worktree add --orphan -B gh-pages gh-pages
         git config user.name "Deploy from CI"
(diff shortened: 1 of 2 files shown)


dvdhrm:
Just for record, this was the state of the last deployment that worked: https://github.com/rust-lang/cargo/tree/cff412598

This has a lot of statically-generated redirects in the root-directory. I cannot find any reference to those in the source tree, so I assume this was statically generated and pushed to the branch.

Please go ahead with the revert. I can try to pull these static redirects into the source-tree and copy them into `gh-pages` on each deploy. But it is late here, so can do this tomorrow at the earliest. Sorry for breaking things!

weihanglo:
Yeah. Maybe I should include the generate.py here and do all the works, instead of reverting this.

weihanglo:
It's ready for review. Thank you @dvdhrm for having a look at it :)

PR #12849, opened 2023-10-18
Make the precise field of a source an Enum

Eh2406:

The precise field of a source_id is a stringly typed untagged union for storing various kinds of data. This is a series of changes untangle what this field is used for. Eventually leading to it being stored as an Enum with dedicated helpers for setting or getting the values different use cases need.

This is an internal re-factor, and all the tests still pass.
This can be reviewed one PR at a time. The first three look at patterns and how precise is used the throughout the code base and make dedicated helpers for each common use case. The last PR switches to an Enum and make helpers for the one-off use cases.

### Additional information

I remember having an issue to do this work, but I cannot find it now.

It's not clear to me if I went overboard on making helpers, API design iteration is welcome.

It would be nice if the `precise` was entirely an internal of the source, this PR does not get us all the way there. This PR makes the representation of the field and internally detail, but its existence still matters to many other parts of the code. Suggestions are welcome.

### DIFF
--- a/src/cargo/core/source_id.rs
+++ b/src/cargo/core/source_id.rs
@@ -3,7 +3,8 @@ use crate::sources::registry::CRATES_IO_HTTP_INDEX;
 use crate::sources::source::Source;
 use crate::sources::{DirectorySource, CRATES_IO_DOMAIN, CRATES_IO_INDEX, CRATES_IO_REGISTRY};
 use crate::sources::{GitSource, PathSource, RegistrySource};
-use crate::util::{config, CanonicalUrl, CargoResult, Config, IntoUrl, ToSemver};
+use crate::util::interning::InternedString;
+use crate::util::{config, CanonicalUrl, CargoResult, Config, IntoUrl};
 use anyhow::Context;
 use serde::de;
 use serde::ser;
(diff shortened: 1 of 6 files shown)


src/cargo/core/source_id.rs
```diff
@@ -449,6 +449,16 @@ impl SourceId {
...
     }
 
+    /// Check if the precise data field has bean set
+    pub fn has_precise(self) -> bool {
+        self.inner.precise.is_some()
+    }
+
+    /// Check if the precise data field has bean set to "Locked"
+    pub fn has_locked_precise(self) -> bool {
+        self.inner.precise.as_deref() == Some("Locked")
```
epage: Here, `"Locked"` is uppercase but it is lowercase at 

```rust
if source_id.precise() != Some("locked") {
```

src/cargo/core/package_id.rs
```diff
@@ -160,14 +160,6 @@ impl PackageId {
         self.inner.source_id
     }
 
-    pub fn with_precise(self, precise: Option<String>) -> PackageId {
```
epage, line 163: Was this change meant to be in the next commit?

epage:
Thanks for doing this!  Last night I was thinking about this being needed for #12425

src/cargo/core/package_id.rs
Eh2406: It is dead code. Would you like me to split it out into a separate commit?

src/cargo/core/source_id.rs
```diff
@@ -50,14 +51,37 @@ struct SourceIdInner {
...
+    },
+    GitUrlFragment(String),
+}
+
+impl fmt::Display for Precise {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Precise::Locked => "locked".fmt(f),
+            Precise::Updated { name, from, to } => {
+                write!(f, "{}={}->{}", name, from, to)
```
weihanglo, line 78: nit

```suggestion
                write!(f, "{name}={from}->{to}")
```

weihanglo (approved):
Looks good :)

Eh2406:
Rebased and fixed.

PR #12844, opened 2023-10-18
docs: fix typo

shuoer86:

PR #12856, opened 2023-10-19
chore(triagebot): auto label when PR review state changes

weihanglo:

Saw this <https://github.com/rust-lang/cargo/pull/12844#issuecomment-1770971174> and realized that we can configure:

* https://forge.rust-lang.org/triagebot/review-requested.html
* https://forge.rust-lang.org/triagebot/review-submitted.html

People have already [done this](https://github.com/rust-lang/rust/blob/3fbcfd2b6f7030cb70328aa759107efc1516912c/triagebot.toml#L26-L36) in rust-lang/rust.

Merge it and use it.

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -36,6 +36,15 @@ warn_non_default_branch = true
 [assign.owners]
 "*" = ["@ehuss", "@epage", "@weihanglo"]
 
+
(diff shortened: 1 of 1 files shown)


shuoer86:
so efficient!

Issue #9172, opened 2021-02-14
Cargo vendor with patch section and local sources does not vendor the local files


Firstyear:
**Problem**
If a patch section contains a specification with a "path" override, cargo vendor does not vendor these content to the vendor directory. 

**Steps**
Add a patch statement such as:

```
[patch.crates-io]
mycrate = { path = "../mycrate" }
```

Then run `cargo vendor`

**Possible Solution(s)**
Cargo vendor should copy the content of the crate (excluding git artefacts) to the vendor directory. 

**Notes**

cargo 1.50.0 (f04e7fab7 2021-02-04)

ianrrees:
I think this is a more general problem actually, it seems `cargo vendor` doesn't vendor crates referred to by paths.

Firstyear:
Paths can be a tricky one as the source could be in the same workspace though. I think it would only be a path that refers to a crate outside the workspace that would be needed here?

ianrrees:
Yeah, and I guess the implementation would only copy files that would be left after a `cargo clean`.

Mainly, my comment was just meant to say that the patch statement in your original example may be expected to fail.  I found this issue after being surprised that `cargo vendor` didn't copy a dependency that's referred to by path.  It would be nice if there were at least an option to also vendor such dependencies.

stevenroose:
I proposed https://github.com/rust-lang/cargo/pull/12858 to solve this issue.

weihanglo:
I am not going to propose `cargo-vendor` to support vendoring all path dependencies. But for this `[patch]` use case specifically, should we reconsider to lift the limitation a bit?

[`[patch]` in config.toml](https://doc.rust-lang.org/cargo/reference/unstable.html#patch-in-config) is a thing since 1.56. This feature opens a door for people to have a central place to configure their patches. In an enterprise environment people might have a pre-configured `.cargo/config.toml` for their own internal patched dependencies. These patched dependencies might also be pre-downloaded somewhere in the local file system as target patched path dependencies. These all work fine until `cargo vendor` fails.

`[patch]` in config could be a system-wide setting that a user might not have control over, so it is better that `cargo vendor` includes those pointing to local path dependencies.

(This also applies to vendoring path dependencies in general. However, path dependencies are not system-wide configuration hence I am not going to justify that use case.)

weihanglo:
Proposed another proof-of-concept here in <https://github.com/rust-lang/cargo/pull/13347>.

Firstyear:
[...]

PR #10135, opened 2021-11-29
vendor: Add --skip-path-deps to skip path dependencies

stevenroose:
This changes the erroneous behavior to skip path dependencies and adds a
command option to keep the current behavior.

Fixes https://github.com/rust-lang/cargo/issues/10134 and https://github.com/rust-lang/cargo/issues/9172.

Issue #10134, opened 2021-11-29
Vendoring broken for path dependencies


stevenroose:
### Problem

I have a path dependency (`crate-name = { path = "/some/local/dir" }`) in my `Cargo.toml` and when I run `cargo vendor`, it doesn't show up in the vendor directory, so when I export the current directory, that crate is unaccounted for.

### Steps

1. create a crate `dep1` in `~/tmp/dep1`
2. create a crate `project1` in ~/tmp/project1`
3. add `dep1 = { path = "../dep1" }` to `~/tmp/project1/Cargo.toml`'s dependency section
4. go into `~/tmp/project1`
5. run `cargo vendor`

### Possible Solution(s)

Don't make a special exception for path dependencies when vendoring.

### Version

```text
cargo 1.49.0 (d00d64df9 2020-12-05)
```

weihanglo:
Similar issue: https://github.com/rust-lang/cargo/issues/9172

Seems that cargo-vendor skips path dependency at this moment.

https://github.com/rust-lang/cargo/blob/8032ac5ef09fe984b818561928793f963f2ffc2c/src/cargo/ops/vendor.rs#L146-L150

ehuss:
Can you share some more details about your use case?

Cargo does not support source replacement of path dependencies, so just including path dependencies in the vendor directory won't help.

hbina:
I am working on a project where I need to make a lot of little changes to many crates that I use. I also would like to test them out in various environemnt (Windows, Linux, MacOS). So it is desirable to be able to simply checkin all these changes.

I thought I was able to do this:

1. Work on crate A
2. Notice I need to make a change on crate B
3. Create a local copy of B
4. Vendor crate B, check in the code
5. Checkout the commit in other environments to test them out

I could have cloned each crates individually, checkout the specific branch, and coordinate all this.

I am not sure why cargo is making that choice on my behalf?

stevenroose:
I proposed https://github.com/rust-lang/cargo/pull/12858 to solve this issue.

weihanglo:
Hi. @hbina I don't quite get the case you share. Do you have a sample project or Cargo.toml to demo?

As I understand it, you want to vendor every change of you path dependencies, and share it across different machines to test it. If your project is under git control, is it possible to just create a git branch containing those change?

@stevenroose could you share the use case or workflow of why you want path dependencies also are exported s vendored source?

PR #12858, opened 2023-10-19
vendor: Add --include-path-deps to include path dependencies

stevenroose:
Fixes https://github.com/rust-lang/cargo/issues/9172, https://github.com/rust-lang/cargo/issues/10134.
Replacement of https://github.com/rust-lang/cargo/pull/10135.

I would just like to say that I personally think it's incorrect that path dependencies are skipped by default (especially without this being documented explicitly anywhere). I think #10135 is a more correct fix for this behavior that I consider a bug. However, since it seems that #10135 is not considered a bugfix by the cargo devs, I would like to propose this alternative solution to make a flag that explicitly includes path dependencies.

This means that using the flag will break with existing behavior, but using the flag will also fix the current documented behavior.

### DIFF
--- a/src/bin/cargo/commands/vendor.rs
+++ b/src/bin/cargo/commands/vendor.rs
@@ -32,6 +32,7 @@ pub fn cli() -> Command {
             "versioned-dirs",
             "Always include version in subdir name",
         ))
+        .arg(flag("include-path-deps", "Include path dependencies"))
         .arg(unsupported("no-merge-sources"))
         .arg(unsupported("relative-path"))
         .arg(unsupported("only-git-deps"))
@@ -75,6 +76,7 @@ pub fn exec(config: &mut Config, args: &ArgMatches) -> CliResult {
             no_delete: args.flag("no-delete"),
             destination: &path,
             versioned_dirs: args.flag("versioned-dirs"),
+            include_path_deps: args.flag("include-path-deps"),
             extra: args
                 .get_many::<PathBuf>("tomls")
                 .unwrap_or_default()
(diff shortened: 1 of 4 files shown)


weihanglo:
While the effort is pretty much appreciated, the current approach doesn't seem to work.

As ehuss [mentioned earlier](https://github.com/rust-lang/cargo/issues/10134#issuecomment-1005013624), source replacement doesn't support path dependencies. The current output vendor-config is like

```toml
[source."path+file:///home/user/cargo/target/tmp/cit/t0/foo/bar"]
directory = "file:///home/user/cargo/target/tmp/cit/t0/foo/bar"

[source.vendored-sources]
directory = "vendor"
```

It makes `bar` a [directory source](https://doc.rust-lang.org/nightly/cargo/reference/source-replacement.html#directory-sources) at `~/cargo/target/tmp/cit/t0/foo/bar`, but the `bar` is not really a directory source. If you add the config to your `.cargo/config.toml` and run `cargo build --verbose`, you'll see that cargo won't pick the vendored `bar` from `vendor/bar`.

We'd happy to discuss use cases, workflows and design first in <https://github.com/rust-lang/cargo/issues/10134>, and see which approach fits better with the community need and the current architecture of Cargo.

weihanglo:
I am going to close this and keep the alternative <https://github.com/rust-lang/cargo/pull/13347> open. <https://github.com/rust-lang/cargo/pull/13347> is a proof-of-concept that addresses the issue mentioned in <https://github.com/rust-lang/cargo/pull/12858#issuecomment-1784156230>. It supports path sources in source replacement so that Cargo can pick up correct sources.

Thanks for the pull request!

PR #5592, opened 2018-05-29
Print the full path to the generated documentation

esteve:
Resolves #5562

PR #12859, opened 2023-10-19
feat(doc): Print the generated docs links

### LINKED ISSUE #5562: Add command line option to print URL/path/link to generated documentation
I use `cargo doc --open` to generate/update docs and open them in a browser, but often I want to view them in a non-default browser.   Sometimes I want to use the viewer built into my IDE, sometimes I'm testing out a new browser, etc.

It would be nice if either `cargo doc` printed the URL/path/link to the newly generated docs when it was done (similar to how `cargo test` prints "Running target/debug/deps/foo"), or if a new option like `cargo doc --link`, `cargo doc --url`, etc could be added to do this.

If either of these sounds like a reasonable idea I could take a crack at putting a pull request together?

epage:
I've wanted something like this myself.  I dislike using `--open` because I tend to move up to re-run my `cargo doc` run but then have to edit it to remove `--open`.
Also makes it annoying when opening docs when `cargo doc` is wrapped by a tool like `make`.

This was previously attempted in #5592:
- Unlike the request in #5562, this aligns with #5592 in always printing rather than using a flag as this seems generally useful
- Unlike #5592, this prints as an alternative to "Opening" to keep things light
- Unlike #5592, this prints afterwards as the link is only valid then

Fixes #5562

### DIFF
--- a/crates/cargo-test-support/src/compare.rs
+++ b/crates/cargo-test-support/src/compare.rs
@@ -237,6 +237,7 @@ fn substitute_macros(input: &str) -> String {
         ("[WAITING]", "     Waiting"),
         ("[PUBLISHED]", "   Published"),
         ("[BLOCKING]", "    Blocking"),
+        ("[GENERATED]", "   Generated"),
     ];
     let mut result = input.to_owned();
     for &(pat, subst) in &macros {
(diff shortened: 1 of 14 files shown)


weihanglo (approved):
This looks great. Have you consider integrating something like [hyperlinks in terminal emulators](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)?


`ripgrep` just got a similar feature, though not by OSC escape sequence:

* <https://github.com/BurntSushi/ripgrep/blob/7099e174acbcbd940f57e4ab4913fee4040c826e/crates/printer/src/hyperlink_aliases.rs#L4-L22>
* <https://github.com/BurntSushi/ripgrep/discussions/2611>

epage:
We'd also want to do that for `--timings`.

How about we do that in a follow up?

epage:
Yeah, reading that further, seems like we should have a dedicated issue for planning and implementing it

weihanglo:
Fair. Thank you

@bors r+

PR #12861, opened 2023-10-20
feat: respect `rust-version` when generating lockfile 

weihanglo:

Respect `package.rust-version` when generating lockfile, so that
a package with an old MSRV will never get an incompatible lockfile,
even when using the latest Cargo.

Users are still able to edit the `version` field in the lockfile
manually, if they intend to switch to a specific lockfile version.

There is a test matrix of different combinations of msrv and lockfile, which demonstrates that only when missing a lockfile it generates a lockfile compatible with MSRV.

Unless people are doing weird things to lockfile, this is a "more correct" behavior for a project with `rust-version` set.

### Additional information

This is a request from <https://github.com/rust-lang/cargo/pull/12852#discussion_r1365593339>

Blocked on #12860.


src/cargo/core/resolver/resolve.rs
```diff
@@ -97,6 +98,21 @@ impl ResolveVersion {
     pub fn max_stable() -> ResolveVersion {
         ResolveVersion::V3
     }
+
+    /// Gets the default lockfile version for the given Rust version.
+    pub fn with_rust_version(rust_version: Option<&RustVersion>) -> Self {
+        let Some(rust_version) = rust_version else {
+            return ResolveVersion::default();
```
epage, line 105: Should we remove `Default` impl to help force people to use `with_rust_version`?

src/cargo/ops/lockfile.rs
```diff
@@ -64,9 +64,10 @@ pub fn write_pkg_lockfile(ws: &Workspace<'_>, resolve: &mut Resolve) -> CargoRes
     // out lock file updates as they're otherwise already updated, and changes
     // which don't touch dependencies won't seemingly spuriously update the lock
     // file.
-    let default_version = ResolveVersion::default();
+    let default_version = ResolveVersion::with_rust_version(ws.rust_version());
     let current_version = resolve.version();
     let next_lockfile_bump = ws.config().cli_unstable().next_lockfile_bump;
+    tracing::debug!("lockfile - current: {current_version:?}, default: {default_version:?}");
```
epage, line 70: I believe the callers of this have the rust version.  Should we use that?

The difference is whether `--ignore-rust-version` was used or not.

src/cargo/core/resolver/resolve.rs
```diff
@@ -97,6 +98,21 @@ impl ResolveVersion {
     pub fn max_stable() -> ResolveVersion {
         ResolveVersion::V3
     }
+
+    /// Gets the default lockfile version for the given Rust version.
+    pub fn with_rust_version(rust_version: Option<&RustVersion>) -> Self {
```
epage, line 103: Should we put this behind `-Zmsrv-policy` or insta-stable?

I'm fine with insta-stable.  The main policy questions would be around
- Whether this should respect `--ignore-rust-version`, where present
- A workspace has no `rust-version`, so this is determined by finding the minimum `package.rust-version` among all workspace members

src/cargo/core/resolver/resolve.rs
weihanglo: A even terrible idea: we rename it to `fn default(…)` so people never misuse 🙈.

src/cargo/ops/lockfile.rs
weihanglo: Did you mean the caller of `write_pkg_lockfile`? Yep we probably need to take care of that route.

src/cargo/core/resolver/resolve.rs
weihanglo: Don't feel like it's a thing `-Zmsrv-policy` wants to address. If we had got MSRV-aware resolver when we started versioning lockfile, I believe we would have added this.

Regarding `--ignore-rust-version`, I am pretty sold on respecting that, given the semantic is like "hey I want Cargo to work as if I never set `rust-version`".

epage: I'm fine with that.

weihanglo: After looking into the current codebase, I am now more unsure about supporting `--ignore-rust-version` for generating lockfile. If we aim to do that, we may need to expose `--ignore-rust-version` to every command that generates lockfiles.

* `cargo build` and other commands involving build
* `cargo fix`
* `cargo generate-lockfile`
* `cargo tree`
* `cargo metadata`
* `cargo update`

It's a cognitive overload for users to learn they have the control over lockfile versions for all these commands. The user experience would be better if we just provide it out-of-the-box. Lockfile versions are somehow opaque to users to me.

I cannot think of any situation people may opt-in `--ignore-rust-version` for controlling the lockfile version. It's incompatible when you want to support old versions but requesting for a new lockfile version. For people want old lockfiles, unless there are bugs in newer versions, otherwise we don't want people to continue using old versions. And they are alwasy able to set `rust-version` for older lockfiles.

Moreover, the workaround is always there — manually changing the `version` field in the lockfile.

Is there any use case I am not aware of for respecting `--ignore-rust-version` for lockfile generating?

weihanglo: We could start respecting `rust-version` from v4, so `rust-version` older than `1.75` won't affect lockfile versions. This would be less controversial.

src/cargo/core/resolver/resolve.rs
```diff
@@ -97,6 +98,21 @@ impl ResolveVersion {
...
+        let Some(rust_version) = rust_version else {
+            return ResolveVersion::default();
+        };
+
+        if rust_version >= &"1.53".parse::<RustVersion>().unwrap() {
+            ResolveVersion::V3
+        } else if rust_version >= &"1.41".parse::<RustVersion>().unwrap() {
```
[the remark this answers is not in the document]
weihanglo: Fixed but if you already have an alternative in mind, do let me know!

src/cargo/core/resolver/resolve.rs
epage: This is reminding me of an idea I had for MSRV-aware resolver of putting the MSRV in the lockfile and making it sticky.  Having to pass flags along for every call can be pretty frustrating.

ehuss:
r? @epage 
If you're ok with that.

epage:
note: this was also discussed at https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Cargo.2Elock.20and.20--ignore-rust-version

src/cargo/core/resolver/resolve.rs
weihanglo: addressed this in an awkward way but easier to refer to: <https://github.com/rust-lang/cargo/pull/12861/commits/991663ca614e1f7e48541d632d71766a17282490>. Let me know if there are other strategies to communicate this.

weihanglo:
@rustbot fcp merge

#12852 is just merged, and it's better, though not required, to have this merged hand-in-hand.

For the change, see the PR description.

I propose to the team to merge this without respecting `--ignore-rust-version`. Reasons:

* MSRV resolution is still in discussion. I would like to avoid overloading the meaning of `--ignore-rust-version` before it settles down.
* Lockfile version is not really tied to dependency resolution, at least for now.

The remaining question is, would it be an breaking change if we revert this pull request in the future?

weihanglo:
wrong bot

@rfcbot fcp merge

#12852 is just merged, and it's better, though not required, to have this merged hand-in-hand.

For the change, see the PR description.

I propose to the team to merge this without respecting `--ignore-rust-version`. Reasons:

* MSRV resolution is still in discussion. I would like to avoid overloading the meaning of `--ignore-rust-version` before it settles down.
* Lockfile version is not really tied to dependency resolution, at least for now.

The remaining question is, would it be an breaking change if we revert this pull request in the future?

Eh2406:
I'm confused. What does this PR do/stabilize? The description makes it sound like it's implementing the MSRV resolver, but that RFC has not yet been accepted. And if this PR is only changing the unstable behavior then why the checkboxes?

weihanglo:
Sorry @Eh2406. It was my fault. The description has been updated. This PR has nothing to do with MSRV resolver; it stabilizes lockfile generation respecting `package.rust-version`.

weihanglo:
> And if this PR is only changing the unstable behavior then why the checkboxes?

The PR is insta-stable, and affects lockfile generation back to v1, if a package has set `package.rust-version` with such an old version.

Eh2406:
So the version of the lock file encoding will be in a format that was stable/readable as of your MSRV. Got it. Yes, that makes sense.

ehuss:
@weihanglo you can `r=ehuss` if and when you are ready.

weihanglo:
@bors r=ehuss

Thanks!

Issue #12405, opened 2023-07-27
`cargo rm foo` on error should suggest `--dev` or `--build`, if appropriate


epage:
### Problem

I just ran
```console
$ cargo rm dunce
    Removing dunce from dependencies
error: the dependency `dunce` could not be found in `dependencies`.
```

### Proposed Solution

It'd be nice if the error had a note saying that a `--dev` dependency existed

epage:
If you accidentally run `cargo rm dunce` twice, it shouldn't act as if you did `cargo rm --dev dunce`.

While "if the tool knew what I intended, it should just do it" is a great principle, I feel like over doing it could lead to some very wrong behavior.

hi-rustin:
I am working on it.

PR #12865, opened 2023-10-22
Add better error message when it can not find the search section

hi-rustin:

close https://github.com/rust-lang/cargo/issues/12405

Try to give a better error message when it can not find the specific section.

I updated all broken tests. You can check out these tests.

### Additional information
None

### DIFF
--- a/src/cargo/util/toml_mut/manifest.rs
+++ b/src/cargo/util/toml_mut/manifest.rs
@@ -367,17 +367,31 @@ impl LocalManifest {
     pub fn remove_from_table(&mut self, table_path: &[String], name: &str) -> CargoResult<()> {
         let parent_table = self.get_table_mut(table_path)?;
 
-        let dep = parent_table
-            .get_mut(name)
-            .filter(|t| !t.is_none())
-            .ok_or_else(|| non_existent_dependency_err(name, table_path.join(".")))?;
+        match parent_table.get_mut(name).filter(|t| !t.is_none()) {
+            Some(dep) => {
+                // remove the dependency
+                *dep = toml_edit::Item::None;
 
-        // remove the dependency
-        *dep = toml_edit::Item::None;
+                // remove table if empty
+                if parent_table.as_table_like().unwrap().is_empty() {
+                    *parent_table = toml_edit::Item::None;
+                }
+            }
+            None => {
+                // Search in other tables.
+                let sections = self.get_sections();
+                let found_table_path = sections.iter().find_map(|(t, i)| {
+                    let table_path: Vec<String> =
+                        t.to_table().iter().map(|s| s.to_string()).collect();
+                    i.get(name).is_some().then(|| table_path.join("."))
+                });
 
-        // remove table if empty
-        if parent_table.as_table_like().unwrap().is_empty() {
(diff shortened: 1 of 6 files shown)


hi-rustin:
Hi @epage, do you have any suggestions here? It seems we use the API from the local manifest to modify the `Cargo.toml`. 

https://github.com/rust-lang/cargo/blob/a2a2cf09e5fb8ba8e405858c12a5f04da4b64c6f/src/cargo/ops/cargo_remove.rs#L49

So I am not sure does it makes sense to change the API to add `fallback_search_tables` here or if we have a better API design.

epage:
> I'm not entirely certain whether it would make sense to change the API to include fallback_search_tables or if there's a better API design available.

`remove_from_table` is already dependency-specific.  I would pass `non_existent_dependency_err` and the raw `table_path` to `non_existent_dependency_err` and have it walk all dependency tables (hard coded), skipping over `table_path`, checking if its present and include it in building up in the error.

The caller of the `op` shouldn't be responsible for generating `fallback_search_tables`.

hi-rustin:
@epage Friendly ping~

src/cargo/util/toml_mut/manifest.rs
```diff
@@ -537,9 +551,14 @@ fn non_existent_table_err(table: impl std::fmt::Display) -> anyhow::Error {
...
 fn non_existent_dependency_err(
     name: impl std::fmt::Display,
-    table: impl std::fmt::Display,
+    search_table: impl std::fmt::Display,
+    found_table: Option<impl std::fmt::Display>,
 ) -> anyhow::Error {
-    anyhow::format_err!("the dependency `{name}` could not be found in `{table}`.")
+    let mut msg = format!("the dependency `{name}` could not be found in `{search_table}`.");
+    if let Some(found_table) = found_table {
+        msg.push_str(&format!(" But it was found in `{found_table}`.",));
```
epage, line 559: nit: Starting with "But" feels a little abrupt.  Maybe
```
" It is present in `{found_table}`."
```

Maybe even remove the periods and change it to
```
"; it is present in `{found_table}`"
```

tests/testsuite/cargo_remove/invalid_section/stderr.log
```diff
@@ -1,2 +1,2 @@
     Removing docopt from build-dependencies
-error: the dependency `docopt` could not be found in `build-dependencies`.
+error: the dependency `docopt` could not be found in `build-dependencies`. But it was found in `dependencies`.
```
epage, line 2: Can you update it so that each commit that will be merged passes tests?

src/cargo/util/toml_mut/manifest.rs
hi-rustin: Updated.

tests/testsuite/cargo_remove/invalid_section/stderr.log
hi-rustin: I've reabsed them into one commit.

epage:
rust-lang/cargo#12921 addresses the nightly failure

epage:
Thanks!

@bors r+

Issue #8283, opened 2020-05-26
Provide Cargo messages as JSON messages


ehuss:
**Describe the problem you are trying to solve**
When integrating Cargo with an editor or IDE, it is common to use JSON to get structured messages from `rustc`. However, Cargo's own messages (such as errors, warnings, etc.) are sent as plain text, making it difficult for the development environment to know how to display these messages.

**Describe the solution you'd like**
It would be nice if Cargo's messages were sent as JSON when `--message-format=json` is used.

**Notes**
There are many decisions to be made on the behavior and exact structure to use.

* Is it OK to simply change Cargo to start sending JSON with `--message-format=json`? That could break or disrupt tools, though I'm not sure how likely that is, or what the consequences would be.  Cargo could include something like [rustc's `--json` flag](https://doc.rust-lang.org/nightly/rustc/command-line-arguments.html#option-json) to configure JSON behavior, but it would be nice to avoid if possible to keep things simple.
* Should Cargo re-use rustc's `"compiler-message"` structure for warnings and errors, or should it add a new one?  It might be difficult to shoe-horn Cargo's format (which uses a stack of causes) to rustc's format.  The drawback of using a separate structure is that then there would need to be multiple ways to handle "messages".
* What other activities and messages should be JSON messages? See https://github.com/rust-lang/cargo/pull/8165#issuecomment-630183516 for some examples. Cargo has many different things it prints (status messages, "downloading", etc.), and there are quite a few action points ("build started", "unit started", "unit finished", etc.) that could be useful to have as JSON.
* Should Cargo intercept non-JSON output from rustc and wrap it in JSON (#8179)?

mchernyavsky:
> Is it OK to simply change Cargo to start sending JSON with --message-format=json? That could break or disrupt tools, though I'm not sure how likely that is, or what the consequences would be. Cargo could include something like rustc's --json flag to configure JSON behavior, but it would be nice to avoid if possible to keep things simple.

Don't forget about other JSON-related formats for `--message-format` (e.g., `json-diagnostic-rendered-ansi`) in case if you choose to include a new flag.

> Should Cargo re-use rustc's "compiler-message" structure for warnings and errors, or should it add a new one? It might be difficult to shoe-horn Cargo's format (which uses a stack of causes) to rustc's format. The drawback of using a separate structure is that then there would need to be multiple ways to handle "messages".

Is it ok to mark build tool messages as messages from the compiler? (I'm talking about "compiler-message" reason)

[...]

PR #12867, opened 2023-10-22
add message when compile starts for crate

### LINKED ISSUE #12864: `cargo … --message-format json` should tell you when it starts compiling a cate.
### Problem

`cargo build --message-format json` will have output such as:
```jsonc
// stdout: (none)
// stderr:
Compiling png
// stdout:
{"reason":"compiler-artifact","package-id":"png 0.17.0"} // compilation finishes
```
There is no way to see when `png` starts compiling.
The only way to find out, is to *parse stderr*, which also doesnt give you the full version information + manifest + etc

### Proposed Solution

`cargo build --message-format json` will have output such as:
```json
{"reason":"begin-compile-crate","package_id":"png 0.17.0","target":"…","features":[],"executable":null,"profile":"…"}
{"reason":"compiler-artifact","package-id":"png 0.17.0"}
```

bend-n:
fixes #12864

adds json output such as:
```json
{"reason":"compile-started","package_id":"foo 0.0.1"}
```

- `git clone --depth 1 https://github.com/rust-lang/cargo cargo1`
- `git clone --depth 1 https://github.com/bend-n/cargo cargo2`
- `cd cargo2`
- `cargo build`
- `cd ../cargo1`
- `../cargo2/target/debug build --message-format json`

or just `cargo test -- json`

### Notes

should the message have more information?

### DIFF
--- a/src/cargo/core/compiler/job_queue/mod.rs
+++ b/src/cargo/core/compiler/job_queue/mod.rs
@@ -565,7 +565,13 @@ impl<'cfg> DrainState<'cfg> {
                 // NOTE: An error here will drop the job without starting it.
                 // That should be OK, since we want to exit as soon as
                 // possible during an error.
-                self.note_working_on(cx.bcx.config, cx.bcx.ws.root(), &unit, job.freshness())?;
+                self.note_working_on(
+                    cx.bcx.config,
+                    cx.bcx.ws.root(),
+                    &unit,
+                    job.freshness(),
+                    cx.bcx.build_config.emit_json(),
+                )?;
             }
             self.run(&unit, job, cx, scope);
         }
@@ -1096,6 +1102,7 @@ impl<'cfg> DrainState<'cfg> {
         ws_root: &Path,
         unit: &Unit,
         fresh: &Freshness,
+        json: bool,
     ) -> CargoResult<()> {
         if (self.compiled.contains(&unit.pkg.package_id())
             && !unit.mode.is_doc()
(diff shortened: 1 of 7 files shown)


src/cargo/core/compiler/job_queue/mod.rs
```diff
@@ -1129,6 +1136,13 @@ impl<'cfg> DrainState<'cfg> {
                     if unit.mode.is_check() {
                         config.shell().status("Checking", &unit.pkg)?;
                     } else {
+                        if json {
+                            let msg = machine_message::CompileStarted {
+                                package_id: unit.pkg.package_id(),
```
Nemo157, line 1141: This should also include the target and platform, to allow distinguishing the different crates of a package and builds for the host vs target (vs artifact-dependencies targets).

There's also other unit types that "start", like `run-custom-build`, which would be useful to have a message for too.

bend-n: > include the target and platform

The other cargo messages dont do this, it would be pointless for this one to, unless all of them did?

bend-n: Would `run-custom-build` be the start of a `build-script-executed`?

I think I can do that in another PR.

Nemo157: > The other cargo messages dont do this, it would be pointless for this one to, unless all of them did?

The others do include the `target`, IMO them not including the `platform` is a bug (https://github.com/rust-lang/cargo/issues/12869), but maybe best to see what the maintainers say.

> Would `run-custom-build` be the start of a `build-script-executed`?

Yep, exactly.

> I think I can do that in another PR.

:+1:

epage:
Considering the associated issue is closed (#12864), I'm closing this PR

please wait until an issue is discussed and accepted before posting PRs.

bend-n:
The associated issue was "moved" into https://github.com/rust-lang/cargo/issues/8283 though?

epage:
That was a consolidation; there is still not an accepted issue.

Issue #12866, opened 2023-10-22
`cargo install eza eza` installs eza twice


pysan3:
### Problem

I am new to the community so please kindly forgive me if I'm doing something wrong.

I don't think I had this behavior til last month or so, but when I specify the same package name to `cargo install` multiple times, that package gets compiled multiple times as well.
This happens not only with `eza` but with all packages.

I would like cargo to only compile the package once since compiling multiple times shouldn't change anything but take time.

I've put the output of the command in `# Notes`.

### Steps

1. `cargo uninstall eza` (optional)
2. `cargo install eza eza`
3. You will see `eza` installed two times.

TLDR;
```txt
    Updating crates.io index
  Installing eza v0.15.0
   Compiling libc v0.2.149
   ...
    Finished release [optimized] target(s) in 1m 06s
  Installing /home/xxx/.local/share/cargo/bin/eza
   Installed package `eza v0.15.0` (executable `eza`)
  Installing eza v0.15.0
   Compiling libc v0.2.149
   ...
    Finished release [optimized] target(s) in 2m 15s
   Replacing /home/xxx/.local/share/cargo/bin/eza
    Replaced package `eza v0.15.0` with `eza v0.15.0` (executable `eza`)
     Summary Successfully installed eza, eza!
```

### Notes

<details><summary>Command Output</summary>
<p>

```txt
$ cargo install eza eza
    Updating crates.io index
  Installing eza v0.15.0
   Compiling libc v0.2.149
   Compiling proc-macro2 v1.0.69
   Compiling autocfg v1.1.0
   Compiling unicode-ident v1.0.12
   Compiling pkg-config v0.3.27
   Compiling num-traits v0.2.17
   Compiling vcpkg v0.2.15
   Compiling jobserver v0.1.27
   Compiling quote v1.0.33
   Compiling tinyvec_macros v0.1.1
   Compiling cc v1.0.83
   Compiling syn v2.0.38
   Compiling tinyvec v1.6.0
   Compiling thiserror v1.0.50
   Compiling rand_core v0.6.4
   Compiling siphasher v0.3.11
   Compiling unicode-normalization v0.1.22
   Compiling phf_shared v0.11.2
   Compiling rand v0.8.5
   Compiling libz-sys v1.1.12
   Compiling libgit2-sys v0.16.1+1.7.1
   Compiling percent-encoding v2.3.0
   Compiling unicode-bidi v0.3.13
   Compiling iana-time-zone v0.1.58
[...]

PR #12868, opened 2023-10-23
Remove duplicate binaries during install

hi-rustin:

close https://github.com/rust-lang/cargo/issues/12866

Deduped the binaries by name.

See the unit test changes.

### Additional information

Should we pick up a bigger version here? For example: `cargo install foo@1 foo@2` Should we pick the last one here?

### DIFF
--- a/src/bin/cargo/commands/install.rs
+++ b/src/bin/cargo/commands/install.rs
@@ -9,6 +9,7 @@ use cargo::util::IntoUrl;
 use cargo::util::ToSemver;
 use cargo::util::VersionReqExt;
 use cargo::CargoResult;
+use itertools::Itertools;
 use semver::VersionReq;
 
 use cargo_util::paths;
@@ -119,6 +120,7 @@ pub fn exec(config: &mut Config, args: &ArgMatches) -> CliResult {
         .get_many::<CrateVersion>("crate")
         .unwrap_or_default()
         .cloned()
+        .dedup_by(|x, y| x == y)
         .map(|(krate, local_version)| resolve_crate(krate, local_version, version))
         .collect::<crate::CargoResult<Vec<_>>>()?;
 
--- a/tests/testsuite/install.rs
+++ b/tests/testsuite/install.rs
@@ -58,6 +58,34 @@ fn simple() {
 }
 
 #[cargo_test]
+fn install_the_same_version_twice() {
+    pkg("foo", "0.0.1");
+
+    cargo_process("install foo foo")
+        .with_stderr(
+            "\
+[UPDATING] `[..]` index
+[DOWNLOADING] crates ...
+[DOWNLOADED] foo v0.0.1 (registry [..])
+[INSTALLING] foo v0.0.1
+[COMPILING] foo v0.0.1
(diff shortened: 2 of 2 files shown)


src/bin/cargo/commands/install.rs
```diff
@@ -119,6 +120,7 @@ pub fn exec(config: &mut Config, args: &ArgMatches) -> CliResult {
         .get_many::<CrateVersion>("crate")
         .unwrap_or_default()
         .cloned()
+        .dedup_by(|x, y| x.0 == y.0)
```
hi-rustin, line 123: Should we pick up a bigger version here? For example: cargo install foo@1 foo@2 Should we pick the last one here?

weihanglo: What about only de-duping when they are going to install the same version, and giving an error (or warning?) when installing different versions of the same package?

weihanglo: Alternatively, don't de-dup. We only emit warnings at this time, so that we can avoid breaking stuff if people really do odd things like <https://github.com/rust-lang/cargo/issues/12866#issuecomment-1774101768> mentioned.

epage: FYI I think we need some more info from the author before we act on this.

hi-rustin: > when installing different versions of the same package?

It seems we don't have the ability to support different versions. So always pick the newer one is better?

> We only emit warnings at this time, so that we can avoid breaking stuff if people really do odd things like [#12866 (comment)](https://github.com/rust-lang/cargo/issues/12866#issuecomment-1774101768) mentioned.

This behavior is too odd. At least, for the same version, I don't think we break anything because we always get the same binary installed.

weihanglo: Say, running `cargo install mycompiler@2.0.0 mycompiler@3.0.0`. People rely on the old version to install a new version, done in their `build.rs` as a bootstrap process. This is rare but technically possible.

It could start as a warning. Once the warning hits stable, the transition period ends and turns it into picking the newer one as proposed.

What do you think? It might be I am overthinking.

epage: For myself, that is a corner case that I don't expect to be an issue.  However, having gotten more information from the author, I feel like the value gained by this change is below the cost of these conversations.  For myself, I'd say if there is little discuss and we can merge it, fine, otherwise close it.

weihanglo: Fine with either as well. If we want to merge it, yes taking the larger version sounds good to me.

weihanglo: Fine with either as well. If we want to merge it, yes taking the last one sounds good to me.

hi-rustin: @epage @weihanglo I updated it to only dedup the same version. I think it is ready to go. We don't really break anything for this situation.

weihanglo (approved):

weihanglo:
@bors r+

Thanks!

hi-rustin:

PR #12876, opened 2023-10-24
Fix unused_imports warning

ehuss:
https://github.com/rust-lang/rust/pull/116033 recently improved the fidelity of the unused_imports warning, causing this code to surface as something that is not being used (because the `context` module is not publicly exported, and nothing was referencing these imports). This fixes the warning by removing the unused imports.


weihanglo (approved):
Thanks. I've seen these warnings as well.

Issue #7164, opened 2019-07-23
Add documentation for docs.rs metadata


Stargateur:
A question pop on [`stackoverflow`](https://stackoverflow.com/questions/57158261/how-to-generate-the-documentation-of-an-ffi-crate-when-the-native-library-is-not#57158261) today asking about how one could build the doc when `build.rs` fail because there is not the lib needed to link. The lib should not be needed for a doc build so the question ask about how we could skip the build script somehow.

The solution exist but it's very poorly documented, in the [`about`](https://docs.rs/about) page ! And https://doc.rust-lang.org/cargo/commands/cargo-metadata.html or https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata doesn't do better. Would be nice if someone who write better English than me write some doc for it.

Note: I'm surprise this go behind metadata, in my opinion metadata should not affect build, well, this too could be clarify.

https://github.com/integer32llc/rust-playground/issues/192

epage:
What do people think of cross-linking [the metadata table docs](https://doc.rust-lang.org/cargo/reference/manifest.html#the-metadata-table) to https://docs.rs/about/metadata?

The main concern I can think of is the slippery slope of what all should be cross-linked.  I'd limit it to official Rust projects, like docs.rs, the line distinguishing the individual parts is blurred.  When more explicitly dealing with an external process, like `cargo-release`s `package.metadata.release`, its more obvious that where to look for the documentation.

Stargateur:
wow 3 years It feel it someone else that open this issue. I updated the SO question with what docs.io advice in 2022 so now maybe add a link on [build page](https://doc.rust-lang.org/cargo/reference/build-scripts.html) that link to https://docs.rs/about/builds is better.

But metadata page of cargo could also use a link to the metadata table of docs.io

ehuss:
I think it would be great to link to the documentation of docs.rs. Linking to other rust-lang projects is fine. 

I know for myself, it still takes me a while to find the docs.rs docs, and I've seen them many times. They do seem to be somewhat hidden.

I'm not sure exactly where to link it. As mentioned, there are multiple points of interaction (metadata table, `cargo doc`, build scripts, publishing, etc.), and I'd be cautious to emphasize it in too many places. Sorry that isn't too constructive. I'd suggest picking a couple different places where you think it seems like it would be likely to be discovered and found.

epage:
Probably the link we added in #12879 is sufficient.

PR #12879, opened 2023-10-25
docs(ref): Link to docs.rs metadata table

epage:
This gets us part of the cross-linking described in #7164

### DIFF
--- a/src/doc/src/reference/manifest.md
+++ b/src/doc/src/reference/manifest.md
@@ -511,6 +511,10 @@ package-name = "my-awesome-android-app"
 assets = "path/to/static"
 ```
 
+You'll need to look in the documentation for your tool to see how to use this field.
+For Rust Projects that use `package.metadata` tables, see:
+- [docs.rs](https://docs.rs/about/metadata)
+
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/manifest.md
```diff
@@ -511,6 +511,9 @@ package-name = "my-awesome-android-app"
 assets = "path/to/static"
 ```
 
+First-party `package.metadata` tables:
```
weihanglo, line 514: Feel a bit abrupt we suddenly mention this. Not a good one but something like below?

```suggestion
Here is a list of `package.metadata` definitions maintained by the Rust Project:
```

epage: Did a bit of a hybrid.

> You'll need to look in the documentation for your tool to see how to use this field.
> For Rust Projects that use `package.metadata` tables, see:
> - [docs.rs](https://docs.rs/about/metadata)

Thoughts?

weihanglo: Way better than mine!

weihanglo (approved):

PR #12880, opened 2023-10-25
refactor(shell): Write at once rather than in fragments

epage:

For `cargo add` and `cargo search`, rather than expose `termcolor`s API, we wrote a styled fragment at a time.

This is no longer needed as of #12751 because we switched from termcolor to anstream which decorates `stdout` and `stderr` with any of the logic we need to adapt to the configured terminal. .

Ran the commands locally to verify styled output.  Tests verify unstyled output.

### DIFF
--- a/src/cargo/core/shell.rs
+++ b/src/cargo/core/shell.rs
@@ -340,13 +340,6 @@ impl Shell {
         }
     }
 
-    /// Write a styled fragment
-    ///
-    /// Caller is responsible for deciding whether [`Shell::verbosity`] is affects output.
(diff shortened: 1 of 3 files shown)


epage:
@hi-rustin this is an example of how to style without `write_stderr` / `write_stdout`

weihanglo (approved):

weihanglo:
Looks good. Thanks

@bors r+

hi-rustin:
> @hi-rustin this is an example of how to style without `write_stderr` / `write_stdout`

Thank you! Got it!

Issue #12801, opened 2023-10-10
Official API for parsing `Cargo.toml`s schema


epage:
### Problem

Something is needed by crates.io for creating index entries, rather than reading the json blob which might be out of sync.  They switched to community developed crates but this creates hidden relationships when making manifest changes (rust-lang/crates.io#6914).

### Proposed Solution

Bigger requests, like #4614, are blocked on the fact that the code ties into everything within cargo.  So instead, let's get an incremental step by just putting out the schemas, with no business logic.

### Notes

Steps
- [x] #12654
- [x] #12768
- [x] #12881
- [x] #12902
- [x] #12911
- [x] #12926
- [x] #12940
- [x] #12948
- [x] #12954
- [x] #12960
- [x] #12971
- [x] #13000
- [x] #13080
- [x] #13097
- [x] #13128
- [x] #13154
- [x]  #13164
- [x] #13166
- [ ] #13178

Related work
- #12924

See also #2259

epage:
`InternedString` offers us two big advantages
- Cheap clone
- Cheap equality checks

My hope is that we can switch to a `Arc<str> | InlinedStr` type that can offer us the same benefits and use it everywhere.  The fallback is to use it just for the schema.

I've updated [string-rosetta-rs](https://github.com/rosetta-rs/string-rosetta-rs) to include equality checks and everything but `Arc<str>` and `arcstr` fails on this front.  I was hoping to use something with a bit more flexibility than `arcstr` as not having to switch between `String` and `OptimizedString` would be nice, like `ecow` or `hipstr` (`flexstr` doesn't have full mutability but does have a `format!`-like macro).  However, that is for a future date and the overall performance of `arcstr` seems really tempting, so I'm going to propose we go with that (it is also maintained by a libs team member)

The main alternative to all of this is to make our types generic over the string type which I'd rather avoid because that gets messy.

Eh2406:
> Cheap equality checks

The thing that makes the equality check so fast is that it does not need to dereference the pointer. Before `InternedString` was added the cash thrashing caused by the pointer dereferences significantly slow down the resolver. That being said, a lot else has improved since then so this may no longer be critical to performance.

> Next step, benchmarking cargo with the change from `InternedString` to `arcstr`

[...]

PR #12881, opened 2023-10-25
refactor(toml): Decouple parsing from interning system

epage:

To have a separate manifest API (#12801), we can't rely on interning because it might be used in longer-lifetime applications.

To keep this limited in scope, this only removes `InternedString` from manifest parsing.  Everything else still uses `InternedString`.

I had problems getting the cargo benchmarks running, so I did a quick and dirty benchmark that is end-to-end, covering fresh builds, clean builds, and resolution.  I ran these against a fresh clone of cargo's code base.  See [my comment](https://github.com/rust-lang/cargo/issues/12801#issuecomment-1778116720) for the script that managed the benchmarks.

Benchmarks:

<details>

```console
$ ../dump/cargo-12801-bench.rs run
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/cargo -Zscript -Zmsrv-policy ../dump/cargo-12801-bench.rs run`
warning: `package.edition` is unspecified, defaulting to `2021`
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `/home/epage/.cargo/target/0a/7f4c1ab500f045/debug/cargo-12801-bench run`
$ hyperfine "../cargo-old check" "../cargo-new check"
Benchmark 1: ../cargo-old check
  Time (mean ± σ):     119.3 ms ±   3.2 ms    [User: 98.6 ms, System: 20.3 ms]
  Range (min … max):   115.6 ms … 124.3 ms    24 runs

Benchmark 2: ../cargo-new check
  Time (mean ± σ):     119.4 ms ±   2.4 ms    [User: 98.0 ms, System: 21.1 ms]
  Range (min … max):   115.7 ms … 123.6 ms    24 runs

Summary
  ../cargo-old check ran
    1.00 ± 0.03 times faster than ../cargo-new check
$ hyperfine --prepare "cargo clean" "../cargo-old check" "../cargo-new check"
Benchmark 1: ../cargo-old check
  Time (mean ± σ):     20.249 s ±  0.392 s    [User: 157.719 s, System: 22.771 s]
  Range (min … max):   19.605 s … 21.123 s    10 runs

Benchmark 2: ../cargo-new check
  Time (mean ± σ):     20.123 s ±  0.212 s    [User: 156.156 s, System: 22.325 s]
  Range (min … max):   19.764 s … 20.420 s    10 runs

Summary
  ../cargo-new check ran
    1.01 ± 0.02 times faster than ../cargo-old check
$ hyperfine --prepare "cargo clean && rm -f Cargo.lock" "../cargo-old check" "../cargo-new check"
Benchmark 1: ../cargo-old check
  Time (mean ± σ):     21.105 s ±  0.465 s    [User: 156.482 s, System: 22.799 s]
  Range (min … max):   20.156 s … 22.010 s    10 runs

Benchmark 2: ../cargo-new check
  Time (mean ± σ):     21.358 s ±  0.538 s    [User: 156.187 s, System: 22.979 s]
  Range (min … max):   20.703 s … 22.462 s    10 runs

Summary
  ../cargo-old check ran
    1.01 ± 0.03 times faster than ../cargo-new check
```

</details>

### Additional information

I consulted https://github.com/rosetta-rs/string-rosetta-rs when deciding on what string type to use for performance.

Originally, I hoped to entirely replacing string interning.  For that, I was looking at `arcstr` as it had a fast equality operator.  However, that is only helpful so long as the two strings we are comparing came from the original source.  Unsure how likely that is to happen (and daunted by converting all of the `Copy`s into `Clone`s), I decided to scale back.

Concerned about all of the small allocations when parsing a manifest, I assumed I'd need a string type with small-string optimizations, like `hipstr`, `compact_str`, `flexstr`, and `ecow`.
The first three give us more wiggle room and `hipstr` was the fastest of them, so I went with that.

I then double checked macro benchmarks, and realized `hipstr` made no difference and switched to `String` to keep things simple / with lower dependencies.

When doing this, I had created a type alias (`TomlStr`) for the string type so I could more easily swap it out if needed
(and not have to always write out a lifetime).
With just using `String`, I went ahead and dropped that.

### DIFF
--- a/src/bin/cargo/commands/remove.rs
+++ b/src/bin/cargo/commands/remove.rs
@@ -286,7 +286,7 @@ fn spec_has_match(
     config: &Config,
 ) -> CargoResult<bool> {
     for dep in dependencies {
-        if spec.name().as_str() != &dep.name {
+        if spec.name() != &dep.name {
             continue;
         }
 
--- a/src/cargo/core/package_id_spec.rs
+++ b/src/cargo/core/package_id_spec.rs
@@ -9,7 +9,6 @@ use url::Url;
 use crate::core::PackageId;
 use crate::util::edit_distance;
 use crate::util::errors::CargoResult;
-use crate::util::interning::InternedString;
 use crate::util::PartialVersion;
 use crate::util::{validate_package_name, IntoUrl};
 
@@ -24,7 +23,7 @@ use crate::util::{validate_package_name, IntoUrl};
 /// sufficient to uniquely define a package ID.
 #[derive(Clone, PartialEq, Eq, Debug, Hash, Ord, PartialOrd)]
 pub struct PackageIdSpec {
-    name: InternedString,
+    name: String,
     version: Option<PartialVersion>,
     url: Option<Url>,
 }
@@ -76,7 +75,7 @@ impl PackageIdSpec {
         };
         validate_package_name(name, "pkgid", "")?;
         Ok(PackageIdSpec {
-            name: InternedString::new(name),
+            name: String::from(name),
             version,
             url: None,
         })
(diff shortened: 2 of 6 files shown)


ehuss:
Thanks!

@bors r+

PR #12884, opened 2023-10-27
Remove outdated option to `-Zcheck-cfg` warnings

Urgau:
Forgot to remove those in https://github.com/rust-lang/cargo/pull/12845.

Note: Would it make sense to have feature-gate tests for those? and if so what would be the recommended way?

### DIFF
--- a/src/cargo/core/compiler/custom_build.rs
+++ b/src/cargo/core/compiler/custom_build.rs
@@ -803,7 +803,7 @@ impl BuildOutput {
                     if extra_check_cfg {
                         check_cfgs.push(value.to_string());
                     } else {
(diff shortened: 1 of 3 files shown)


epage:
> Note: Would it make sense to have feature-gate tests for those? and if so what would be the recommended way?

We regularly have feature gate tests.  In `check_cfg.rs`, just add a test with a build script and without a feature flag that  verifies the warning is in the output.

epage:
Would you like to add the tests as part of this or should I merge it and you'll add them later?

Urgau:
I can do it in this PR, let's me quickly add them.

PR #12889, opened 2023-10-28
feat: Make browser links out of HTML file paths

### LINKED ISSUE #12888: Turn HTML output into links
### Problem

In most terminals, if a URL is printed it is turned into a link you can click on for easy opening.  This is less true for file paths.

This is less of a problem for `cargo doc --open` as cargo will open the page up for you.

Places where we are falling short include
- `cargo doc --target foo --target bar` (`--open` doesn't work)
- Wanting to a `cargo doc` invocation that you can run multiple times, using shell history, without editing in/out `--open`
- `cargo doc` calls wrapped in other tools, like `make`
- `cargo build --timings` (it'd be awkward to add a `--open` just for `--timings`)

### Proposed Solution

Turn the file paths into links using ANSI escape codes

### Notes

From https://github.com/rust-lang/cargo/pull/12859#pullrequestreview-1702451615

> This looks great. Have you consider integrating something like [hyperlinks in terminal emulators](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)?
> 
> `ripgrep` just got a similar feature, though not by using OSC escape sequence:
> 
> * https://github.com/BurntSushi/ripgrep/blob/7099e174acbcbd940f57e4ab4913fee4040c826e/crates/printer/src/hyperlink_aliases.rs#L4-L22
> 
> * [what are your thoughts on ripgrep's new support for hyperlinks? BurntSushi/ripgrep#2611](https://github.com/BurntSushi/ripgrep/discussions/2611)

epage:
This provides an alternative to `--open`, where supported.

Note: because we are relying on `supports-hyperlinks`, we are getting `FORCE_HYPERLINK` for "free".  Unsure whether it and the current policy (it gets overridden by `term.hyperlinks`) is something we want.  `FORCE_HYPERLINK` mirrors the npm package's behavior of the same name (see zkat/supports-hyperlinks#2) though though I also found reading of it in ohmyzsh, a go terminal library and many more places.  Similarly, #12751 added indirect, undocumented support for community environment variables.

Fixes #12888


src/cargo/util/hostname.rs
```diff
@@ -0,0 +1,85 @@
+use std::{ffi::OsString, io};
```
epage, line 1: Copied from [ripgrep](https://github.com/BurntSushi/ripgrep/blob/7099e174acbcbd940f57e4ab4913fee4040c826e/crates/cli/src/hostname.rs) because the `hostname` and `gethostname` packages had what looked like less robust implementations.

src/cargo/ops/cargo_doc.rs
```diff
@@ -47,7 +47,11 @@ pub fn doc(ws: &Workspace<'_>, options: &DocOptions) -> CargoResult<()> {
                     .join("index.html");
                 if path.exists() {
                     let mut shell = ws.config().shell();
-                    shell.status("Generated", path.display())?;
+                    let link = shell.err_file_hyperlink(&path);
+                    shell.status(
+                        "Generated",
```
epage, line 52: Need to add this to `--timings` as well

Might also add it to the above `Opening`

src/cargo/core/shell.rs
```diff
@@ -340,6 +356,63 @@ impl Shell {
...
+            Hyperlink { url: None }
+        }
+    }
+
+    pub fn out_file_hyperlink(&mut self, path: &std::path::Path) -> Hyperlink<url::Url> {
+        let url = self.file_hyperlink(path);
+        url.map(|u| self.out_hyperlink(u)).unwrap_or_default()
+    }
+
+    pub fn err_file_hyperlink(&mut self, path: &std::path::Path) -> Hyperlink<url::Url> {
```
epage, line 392: Need to test that this actually works (unsure if its a bug or my terminal is too old)

weihanglo (commented):
Works for me on Kitty with Zellij.

epage:
*shrug* no idea why wezterm isn't working for me, even when I do `CARGO_TERM_HYPERLINKS=true` and I instrument the code to show it is creating the link.  I even dropped the hostname in case it calculated the hostname differently than us.  I'm also on the latest wezterm release and the nightly changelogs dont mention this being new but its generally advertised as supported.

epage:
This adds a new, insta-stable `.cargo/config.toml` field `term.hyperlinks = <bool>` to allow overriding our auto-detection of terminal hyperlink support.  Technically, any terminal that supports ANSI escape codes should gracefully handle this if its unsupported but in practice some terminals are non-conforming.

This is used to turn file paths into links so people can open `--timings` reports or rustdoc directly without using a `--open` flag.  This doesn't explore further uses like (1) explicit HTTP links (terminals tend to support these implicitly) (2) finding other output for us to easily link, like linking to crates.io pages for `cargo search` results.

@rustbot fcp merge

epage:
This adds a new, insta-stable `.cargo/config.toml` field `term.hyperlinks = <bool>` to allow overriding our auto-detection of terminal hyperlink support.  Technically, any terminal that supports ANSI escape codes should gracefully handle this if its unsupported but in practice some terminals are non-conforming.

This is used to turn file paths into links so people can open `--timings` reports or rustdoc directly without using a `--open` flag.  This doesn't explore further uses like (1) explicit HTTP links (terminals tend to support these implicitly) (2) finding other output for us to easily link, like linking to crates.io pages for `cargo search` results.

@rfcbot fcp merge

ehuss:
I'm a little concerned about this making the interaction worse on some terminals. For example, on mine (iTerm2), it breaks it so that instead of opening the file, it opens a dialog box asking for a username to SCP the file. How important is it to include the hostname? If I was connected via ssh, how would having the hostname in the file path work? Is it possible to not include the hostname if not remotely connected?

> unsupported but in practice some terminals are non-conforming

Do we have a sense of which terminals those are, or how common this is?

epage:
> I'm a little concerned about this making the interaction worse on some terminals. For example, on mine (iTerm2), it breaks it so that instead of opening the file, it opens a dialog box asking for a username to SCP the file. How important is it to include the hostname? If I was connected via ssh, how would having the hostname in the file path work? Is it possible to not include the hostname if not remotely connected?

By "it breaks", do you mean you could before open file paths and now it has the pop up?

From the [reference gist](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#file-uris-and-the-hostname)

>  As such, we don't allow this sloppiness. Utilities that print hyperlinks are requested to fill out the hostname, and terminal emulators are requested to match it against the local hostname and refuse to open the file if the hostname doesn't match (or offer other possibilities, e.g. to download with scp as iTerm2 does).

That makes it sound like iTerm2 should be supported.  As for how it should handle the hostname, it says

> It's unreasonable for simple utilities to go into the business of hostname resolution. As such, we urge utilities to place the value from gethostname() there (shell scripts might go for $HOSTNAME).
>
> Terminal emulators should match the given value against the local hostname. They might accept multiple values, e.g. both short and fully qualified names, but they are free to go with just the value from gethostname(). They also must accept the string localhost or the empty string as local ones. If a different hostname is present, they must not open the local counterpart with the same filename.

> > unsupported but in practice some terminals are non-conforming
> 
> Do we have a sense of which terminals those are, or how common this is?

Again, from the [reference gist](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#backward-compatibility)

> At this moment, terminals known to be buggy (OSC 8 resulting in display corruption) are VTE versions up to 0.46.2 and 0.48.1, Windows Terminal up to 0.9, Emacs's built-in terminal, and [screen](https://savannah.gnu.org/bugs/index.php?57718) with 700+ character long URLs.

In hopes to avoid problems with those is why I pulled in `supports-hyperlinks` as it handles detection for us.

In both cases, in case there are problems, is why I added `term.hyperlinks`.  Someone can open an issue against us and then set that until the issue is fixed.

src/cargo/core/shell.rs
epage: Ah, apparently I missed [this note from burntsushi](https://github.com/BurntSushi/ripgrep/discussions/2611)

> (NOTE: If you're using tmux, its latest release doesn't support hyperlinks. You'll need to build tmux from source and then add set-option -sa terminal-features ",*:hyperlinks" to your tmux configuration to make it work.)

epage:
In the ripgrep thread, I found this sub-thread devoted to iTerm2.  Not seeing anyone mention the SCP problem

https://github.com/BurntSushi/ripgrep/discussions/2611#discussioncomment-7112252

ehuss:
> By "it breaks", do you mean you could before open file paths and now it has the pop up?

Correct.

Just a little more testing, Windows Terminal seems to work, which is nice since it doesn't seem to have native auto detection of paths (though it does auto-detect URLs). ConEmu seems to want to launch FAR, even though I don't have it installed. But it also tries to do that on bare file paths, too, so there is no change there.

Not working with Emacs seems a little worrying, though I think it depends on what it does wrong. I don't have emacs installed, so I can't really test it.

Unfortunately without a hostname, iTerm2 just opens Finder.

I don't know how widespread the issue might be. If out of the majority of terminals, iTerm2 is one of the few that has poor behavior, would it be possible to just disable it for that terminal (since it already works without this PR)?

epage:
I can't tell what we are doing differently from ripgrep (or what might be different about your machine) as I don't have a Mac to debug with.  For now, I've overridden `supports_hyperlinks` to say iTerm2 is not supported.  This can still be overridden with `term.hyperlinks`.  On merge, I'll open an issue to track fixing iTerm2.

src/cargo/core/shell.rs
epage: Hmm, just tried switching from tmux to Zellij.  It now highlights the links but they don't work.  With a raw terminal in Wezterm, the links work.

weihanglo (commented):
I have my vote here just because it works for me 😛.

One tiny concern around the env [`FORCE_HYPERLINK`](https://github.com/zkat/supports-hyperlinks/blob/77b5caadd6e26f431bf051abf3c96149c6be7a49/src/lib.rs#L13-L16) provided by supports-hyperlinks. Is it a thing we consider as part of the insta-stable?

epage:
Sorry, I should have called that out.

Added to the PR description:

> Note: because we are relying on `supports-hyperlinks`, we are getting `FORCE_HYPERLINK` for "free".  Unsure whether it and the current policy (it gets overridden by `term.hyperlinks`) is something we want.  `FORCE_HYPERLINK` mirrors the npm package's behavior of the same name (see zkat/supports-hyperlinks#2) though though I also found reading of it in ohmyzsh, a go terminal library, and many more places.

src/cargo/core/shell.rs
```diff
@@ -340,6 +356,63 @@ impl Shell {
...
+            ShellOut::Write(_) => false,
+            ShellOut::Stream {
+                stdout, hyperlinks, ..
+            } => stdout.current_choice() == anstream::ColorChoice::AlwaysAnsi && *hyperlinks,
+        };
+        if supports_hyperlinks {
+            Hyperlink { url: Some(url) }
+        } else {
+            Hyperlink { url: None }
+        }
```
weihanglo, line 370: nit

```suggestion
        Hyperlink {
            url: supports_hyperlinks.then_some(url),
        }
```

src/cargo/core/shell.rs
```diff
@@ -439,6 +512,44 @@ fn supports_color(choice: anstream::ColorChoice) -> bool {
...
 }
 
+fn supports_hyperlinks() -> bool {
+    #[allow(clippy::disallowed_methods)] // We are reading the state of the system, not config
+    if std::env::var_os("TERM_PROGRAM").as_deref() == Some(std::ffi::OsStr::new("iTerm.app")) {
+        // Override `supports_hyperlinks` as we have an unknown incompatibility with iTerm2
+        return false;
+    }
+
+    ::supports_hyperlinks::supports_hyperlinks()
```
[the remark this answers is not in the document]
epage: Huh, I assumed it was needed to disambiguate but it isn't.

src/cargo/util/hostname.rs
epage: Done

weihanglo:
I don't think this require to run 10 days for FCP, as it is only a cosmetic change for console output, and affect only `--timings` and `cargo doc`. If anyone doesn't agree, please call out. Otherwise I am going to merge this shortly.

@bors r+

weihanglo (approved):

PR #12890, opened 2023-10-28
Improve the margin calculation for the search command's UI

hi-rustin:

Improved the margin calculation for the search command's UI.
1. We don't need always to do the addition, we can only add it after we find the max value.
2. We can use [Formatted print](https://doc.rust-lang.org/rust-by-example/hello/print.html#formatted-print) to add the margin spaces.

I don't break any tests. So it would be OK.

### Additional information
None

### DIFF
--- a/src/cargo/ops/registry/search.rs
+++ b/src/cargo/ops/registry/search.rs
@@ -3,7 +3,6 @@
 //! [1]: https://doc.rust-lang.org/nightly/cargo/reference/registry-web-api.html#search
 
 use std::cmp;
-use std::iter::repeat;
 
 use anyhow::Context as _;
 use url::Url;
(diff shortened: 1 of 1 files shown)


src/cargo/ops/registry/search.rs
```diff
@@ -54,10 +54,12 @@ pub fn search(
...
-                    .take(description_margin - name.len())
-                    .collect::<String>();
-                name + &space + "# " + &desc
+                format!(
+                    "{name: <margin$}# {desc}",
+                    name = name,
+                    margin = description_margin,
+                    desc = desc
+                )
             }
```
weihanglo, line 63:
```suggestion
            Some(desc) => format!("{name: <description_margin$}# {desc}"),

```

weihanglo (commented):
Some nitpicks. Thanks for the contribution!

src/cargo/ops/registry/search.rs
hi-rustin: 👍🏿

hi-rustin:

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #12902, opened 2023-10-31
refactor(toml): Cleanup noticed on the way to #12801

epage:
In splitting up `toml/mod.rs` for #12801, I noticed some things to clean up.  I decided to split this up.

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -1709,7 +1709,7 @@ pub struct TomlPackage {
 
     metadata: Option<toml::Value>,
 
-    // Provide a helpful error message for a common user error.
+    /// Provide a helpful error message for a common user error.
     #[serde(rename = "cargo-features", skip_serializing)]
     _invalid_cargo_features: Option<InvalidCargoFeatures>,
 }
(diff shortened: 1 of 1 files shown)


src/cargo/util/toml/mod.rs
```diff
@@ -1530,25 +1530,23 @@ fn unique_build_targets(
...
     resolver: Option<String>,
+    metadata: Option<toml::Value>,
 
     // Properties that can be inherited by members.
     package: Option<InheritableFields>,
     dependencies: Option<BTreeMap<String, TomlDependency>>,
     lints: Option<TomlLints>,
-
-    // Note that this field must come last due to the way toml serialization
-    // works which requires tables to be emitted after all values.
```
weihanglo, line 1546: Seems like metadata isn't needed to be the last anymore. Do we have a test to verify the change?

epage: Looks like we have several tests in `testsuite/package.rs` that use `package.metadata` which means we'll serialize the prepared manifest and verify its output which would hit this situation.

weihanglo: This was <https://github.com/rust-lang/cargo/commit/6bc74556f9dbfe6d50a0e51a20591f420fb72c0a> but since we have tests and `toml` crate has changed largely. I am going to merge this. Thank you!

weihanglo (approved):

PR #12911, opened 2023-11-02
refactor(toml): Pull out the schema

epage:

On its own, this PR is a net negative for readability / complexity.  It moves all of the serde types related to manifest from `toml/mod.rs` to `toml/schema.rs`, leaving a lot of the functions and some trait implementations back in `toml/mod.rs` (some basic functions that made sense for the type on their own were also moved).

So why do this?  This is an ooch towards having the schema broken out into a separate package (#12801).  To do this, we need to
- Identify what all types need to be put in the package.  This refactor highlights the dependence on `RustVersion` and `PackageIdSpec`
- Highlights what functionality we need to find a new home for

Follow up PRs would
- Find better homes for the logic in `toml/mod.rs`, moving us away from having `impl schema::Type` blocks.
- Pull out a `src/cargo/util_semver` package to own `PartialVersion` (at least) as prep for making a `cargo-util-semver` package
- Move `RustVersion` to `manifest-toml/schema.rs`, deciding what functionality needs to move with the type
- Move or copy `PackageIdSpec` into `manfest-toml/schema.rs`, deciding what functionality remain in `core/` and what moves over
- Move `toml/schema.rs` to `src/cargo/util_schema`
- Actually make `cargo-util-semver` and `cargo-util-manifest-schema` packages

So why do this now? This is a big change!  By being incremental
- Reduce churn for me and others
- Be easier to review
- Collect feedback as we go on the whole plan to avoid more painful changes later

We *can* back this out if needed but the further we go, the more painful it will be.


ehuss:
Just to confirm, this just moves things, there aren't any other changes of substance?

If so, it sounds good to me. However, I'd like to defer to @weihanglo to see what they would like.

r? @weihanglo

weihanglo (commented):
The module name doesn't catch my eye. `toml::schema` looks pretty genernal that we might also want `config.toml` to be inside. I don't think that's the better direction, but is it in the plan in your mind?

src/cargo/util/toml/schema.rs
```diff
@@ -0,0 +1,1134 @@
...
+    pub license: Option<MaybeWorkspaceString>,
+    pub license_file: Option<MaybeWorkspaceString>,
+    pub repository: Option<MaybeWorkspaceString>,
+    pub resolver: Option<String>,
+
+    pub metadata: Option<toml::Value>,
+
+    /// Provide a helpful error message for a common user error.
+    #[serde(rename = "cargo-features", skip_serializing)]
+    pub _invalid_cargo_features: Option<InvalidCargoFeatures>,
```
weihanglo, line 148: What is the use case of making this public?

src/cargo/util/toml/schema.rs
```diff
@@ -0,0 +1,1134 @@
...
+    pub default_features: Option<bool>,
+    #[serde(rename = "default_features")]
+    pub default_features2: Option<bool>,
+    pub optional: Option<bool>,
+    pub public: Option<bool>,
+
+    /// This is here to provide a way to see the "unused manifest keys" when deserializing
+    #[serde(skip_serializing)]
+    #[serde(flatten)]
+    pub unused_keys: BTreeMap<String, toml::Value>,
```
weihanglo, line 462: This seems like for emitting a better warning only. Do we want to pub this?

weihanglo (approved):
While I left two comments about the visibility of seemingly internal only fields, I am going to merge this to unblock stuff.

weihanglo:
> Correct. The most visible changes are for cargo-as-a-library users
> 
>     * Types moved
> 
>     * More types are pub
> 
>     * Fields are pub

Worth noting that, making fields and types `pub` doesn't mean they are stable. There is still no strong stability guarantee for Cargo as a library, but this PR is a path forward that goal.

@bors r+

src/cargo/util/toml/schema.rs
epage: That depends on if we want this to be `#[non_exhaustive]` or directly constructable.  For now, I was going with directly constructable.

epage:
Blocked on #12921

PR #12913, opened 2023-11-03
Updating "features" documentation to add a note about the new limit on number of features

anupj:
Adding the following note to [features docs page](https://doc.rust-lang.org/cargo/reference/features.html)

> Note: New crates or versions published on [crates.io](http://crates.io/) are now limited to
> a maximum of 300 features. Exceptions are granted on a case-by-case basis.
> See this [blog post] for details. Participation in solution discussions is
> encouraged via the [crates.io](http://crates.io/) Zulip stream.

I don't think this change requires any additional testing beyond a general review.


epage:
Thanks!

@bors r+

Issue #7725, opened 2019-12-19
pkgid spec confusion


ehuss:
The documentation for the package id spec (https://doc.rust-lang.org/1.40.0/cargo/reference/pkgid-spec.html) does not seem to be correct in several ways.

* The proto is not optional.  Cargo defaults to `cargo://` which is not a scheme Cargo understands anywhere.
* `crates.io` as a host does not work.  It looks like you can only give the full path to the index, like `https://github.com/rust-lang/crates.io-index#bitflags:1.2.1`
* `host/name` doesn't work.  It looks like [the matching code](https://github.com/rust-lang/cargo/blame/d5e20a9b7f9e3888fd6144f606aadb62bbabece0/src/cargo/core/package_id_spec.rs#L160-L175) requires the entire scheme/host/path to match.

The question is, what should we do?  Should we just remove all the things that don't work from the documentation?  Or should we try to enhance the parsing/matching code to work better.  

* Should `crates.io` work as an alias for `https://github.com/rust-lang/crates.io-index`?
* Missing proto probably shouldn't default to anything, and should probably match any proto.
* Should `host/name` match `host/path#name`?

EDIT: #9188 and #9249 fixed the broken parts, but leaving this open to consider adding abbreviated forms.

alexcrichton:
Oh dear that didn't exactly age well...

I think though we've started using `crates-io` in other places such as `[source.crates-io]`, so supporting either a full host name or the "short name" of the registry seems like a good thing to support (we don't support this today afaik). 

As for the proto business I think it's fine to basically remove that and make the docs match the current implementation.

In any case, what I would propose:

* First make the docs match the current implementation
* Next support `registry-identifier/crate-name` where `registry-identifier` matches what we do for alternative registries.

epage:
I'm proposing we close this.  All that is left is considering more abbreviated forms.  In the 5 years this has been open, no one has piped in about wanting some.  Instead of tracking this as a speculated need, I'd recommend we wait until a user requests it.

ehuss:
Sounds good!

ayanwaar:
why my deposite not arrive how to i recive

PR #12914, opened 2023-11-03
fix(metadata): Stabilize id format as PackageIDSpec 

### LINKED ISSUE #7267: Package spec in build plan and metadata?

I'd like to selectively prebuild some or all of the dependencies of my crate.

I can do this with `cargo build --package log` which is super awesome.

If I have two version of `log`, I need to specify which one with a package spec. Cargo gives me the ability to do `cargo metadata` and `cargo build --build-plan`, it also gives me the option to get an ID with `cargo pkgid` (https://doc.rust-lang.org/cargo/reference/pkgid-spec.html)

Unfortunately, I can't make them play together. `metadata` and `--build-plan` don't provide package specs, so I end up forced to do manual parsing if I want to hope to make all these tools interact.

Include a package spec in the output of `cargo metadata` and/or `cargo build --build-plan`

**Notes**
Related to #2644 since that was essentially what I was trying to do.

epage:

For tools integrating with cargo, `cargo metadata` is the primary interface.  Limitations include:
-  There isn't an unambiguous way to map a package entry from `cargo metadata`  to a parameter to pass to other `cargo` commands.  An `id` field exists but it is documented as an opaque string, useful only for comparisons with other `id`s within the document.
- There isn't an unambiguous way of taking user parameters (`--package`) and mapping them to `cargo metadata` entries.  `cargo pkgid` could help but it returns a `PackageIdSpec` which doesn't exist within the `cargo metadata` output.

This attempts to solve these problems by switching the `id` field from `PackageId` to `PackageIdSpec` which is a [publicly documented format](https://doc.rust-lang.org/cargo/reference/pkgid-spec.html), can be generated by `cargo pkgid`, and is accepted by most commands via the `--package` flag.

As the `"id"` field is documented as opaque, this technically isn't a breaking change though people could be parsing it.

For `cargo_metadata` they do [use a new type that documents it as opaque but publicly expose the inner `String`](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.PackageId.html).  The `String` wasn't publicly exposed due to a request by users but instead their `PackageId` type replaced using `String`s in the API in oli-obk/cargo_metadata#59 with no indication given as to why the `String` was still exposed.  However, you'll note that before that PR, they had `WorkspaceMember` that parsed `PackageId`.  This was introduced in oli-obk/cargo_metadata#26 without a motivation given.

**Note that `PackageIdSpec` has multiple representation that might uniquely identify a package and we can return any one of them.**

Fixes #7267

### Additional information

cc @oli-obk

### DIFF
--- a/src/doc/man/cargo-metadata.md
+++ b/src/doc/man/cargo-metadata.md
@@ -56,7 +56,7 @@ The JSON output has the following format:
             /* The Package ID, an opaque and unique identifier for referring to the
                package. See "Compatibility" above for the stability guarantee.
             */
-            "id": "my-package 0.1.0 (path+file:///path/to/my-package)",
+            "id": "file:///path/to/my-package#0.1.0",
             /* The license value from the manifest, or null. */
             "license": "MIT/Apache-2.0",
             /* The license-file value from the manifest, or null. */
@@ -242,13 +242,13 @@ The JSON output has the following format:
        Each entry is the Package ID for the package.
     */
     "workspace_members": [
-        "my-package 0.1.0 (path+file:///path/to/my-package)",
+        "file:///path/to/my-package#0.1.0",
     ],
     /* Array of default members of the workspace.
        Each entry is the Package ID for the package.
     */
     "workspace_default_members": [
-        "my-package 0.1.0 (path+file:///path/to/my-package)",
+        "file:///path/to/my-package#0.1.0",
     ],
     // The resolved dependency graph for the entire workspace. The enabled
     // features are based on the enabled features for the "current" package.
@@ -266,10 +266,10 @@ The JSON output has the following format:
         "nodes": [
             {
                 /* The Package ID of this node. */
-                "id": "my-package 0.1.0 (path+file:///path/to/my-package)",
+                "id": "file:///path/to/my-package#0.1.0",
                 /* The dependencies of this package, an array of Package IDs. */
                 "dependencies": [
-                    "bitflags 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)"
+                    "https://github.com/rust-lang/crates.io-index#bitflags@1.0.4"
                 ],
                 /* The dependencies of this package. This is an alternative to
                    "dependencies" which contains additional information. In
@@ -283,7 +283,7 @@ The JSON output has the following format:
                         */
                         "name": "bitflags",
                         /* The Package ID of the dependency. */
-                        "pkg": "bitflags 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
+                        "pkg": "https://github.com/rust-lang/crates.io-index#bitflags@1.0.4"
                         /* Array of dependency kinds. Added in Cargo 1.40. */
                         "dep_kinds": [
                             {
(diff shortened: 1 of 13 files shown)


tests/testsuite/script.rs
```diff
@@ -1024,7 +1024,7 @@ fn cmd_metadata_with_embedded() {
                 "default_run": null,
                 "name": "script",
                 "version": "0.0.0",
-                "id": "script[..]",
+                "id": "file:[..]foo#script@0.0.0",
```
epage, line 1027: Not a blocker for this PR but for #12207 is that we need to make sure cargo script PackageIdSpec's are unambiguous.  They currently point to the directory, rather than the script.

tests/testsuite/git.rs
```diff
@@ -3332,28 +3332,28 @@ fn metadata_master_consistency() {
...
               ],
               "workspace_default_members": [
-                "foo 0.1.0 [..]"
+                "[..]foo#0.1.0"
               ],
               "resolve": {
                 "nodes": [
                   {
-                    "id": "bar 1.0.0 (__BAR_SOURCE__#__BAR_HASH__)",
+                    "id": "file:///[..]/bar#1.0.0",
```
epage, line 3343: I worry that `PackageIdSpec` is ambiguous and will need to be extended before we can move forward with this
- It loses the semantics on top of the raw protocol (PackageId exposed this as `git+file`)
- It tracks the version instead of the hash

Any other problems people can find?

epage: If we do make changes, we should keep #7725 in mind

src/doc/man/cargo-metadata.md
```diff
@@ -331,6 +331,11 @@ The JSON output has the following format:
 }
 ````
 
+Notes:
+- For `"id"` field syntax, see [Package ID Specifications] in the reference.
+
+[Package ID Specifications]: ../reference/pkgid-spec.md
```
ehuss, line 337: I think it was noted in the meeting that this generates a broken link in the `.txt` file. mdman does not do the same `.md` to `.html` translation that mdbook does, so all the links here should be `.html`.

```suggestion
[Package ID Specifications]: ../reference/pkgid-spec.html
```

tests/testsuite/git.rs
epage: We talked about this at the cargo team meeting and were generally ok with moving this forward.

weihanglo:
@rfcbot reviewed

BTW, do we have a plan for `SourceId`?

epage:
> BTW, do we have a plan for SourceId?

I do not as I don't see as much reason to stabilize it. The end-user benefit here is that you can run a cargo command against a package you came across in `cargo metadata`.  I don't see similar use cases at this time for SourceId.

src/doc/man/cargo-metadata.md
```diff
@@ -331,6 +331,11 @@ The JSON output has the following format:
 }
 ````
 
+Notes:
```
weihanglo, line 334: should we as well add this to "SEE ALSO" at the end of this file? Like

```
SEE ALSO
       cargo(1), cargo-pkgid(1), [Package ID Specifications](..reference/pkgid-spec.html)
```

weihanglo: https://github.com/rust-lang/cargo/pull/13298

weihanglo:
@bors r+

Thanks!

PR #12915, opened 2023-11-03
fix(cli): Clarify --test is for targets, not test functions

### LINKED ISSUE #7864: `cargo test --test` does not find any tests
Background: I'm improving fish's dynamic shell completions for `cargo` to do things like have `cargo run --bin <TAB>` give you a list of available binaries to choose from for tab completion.

It seems that cargo's dynamic test discovery is broken (or never implemented?). 

e.g. while `cargo bench --bench` returns a list of all benchmarks along with the error message:

```
$ cargo bench --bench
error: "--bench" takes one argument.
Available benches:
    bench
```

`cargo test --test` always returns "no tests available" incorrectly:

```
$ cargo test --test
error: "--test" takes one argument.
No tests available.
```

while actually running `cargo test` shows that cargo has no problem enumerating tests:

```
mqudsi@ZBOOK /m/c/U/m/g/unquote> cargo test
    Finished test [unoptimized + debuginfo] target(s) in 0.12s
     Running target/debug/deps/unquote-d4809e2bcea39ff6

running 18 tests
... <SNIP>
```

epage:
We already refer to test targets as "test targets" instead of "tests" in `--test` but not `--tests` or in the error output.  This makes it uniformly refer to them as "test targets", making it clearer that these aren't test functions.

Fixes #7864

### DIFF
--- a/src/bin/cargo/commands/bench.rs
+++ b/src/bin/cargo/commands/bench.rs
@@ -37,9 +37,9 @@ pub fn cli() -> Command {
             "Benchmark only the specified example",
             "Benchmark all examples",
             "Benchmark only the specified test target",
-            "Benchmark all tests",
+            "Benchmark all test targets",
             "Benchmark only the specified bench target",
-            "Benchmark all benches",
+            "Benchmark all bench targets",
             "Benchmark all targets",
         )
         .arg_features()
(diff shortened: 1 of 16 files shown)


weihanglo (approved):
Most hated word in Cargo: _target_

PR #12921, opened 2023-11-06
tests: Remove plugin tests

epage:
Compiler plugins were removed in rust-lang/rust#116412, so we don't need these tests.

As for the `plugin` field on build-targets, it appears to be [stable-but-deprecated](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-plugin-field), so I left it alone


weihanglo (approved):
Seems good. Thanks you.

PR #12924, opened 2023-11-06
fix: Report more detailed semver errors

epage:
For `cargo install` we'll now show a more specific parse error for semver, much like other parts of cargo.

This came out of my work on #12801.  I was looking at what might be appropriate to put in a `cargo-util-semver` crate and realized we have the `ToSemver` trait that exists but doesn't do much, so I dropped it.


weihanglo (approved):

weihanglo:
Look reasonable. Thank you!

@bors r+

PR #12925, opened 2023-11-06
refactor: Remove VersionExt

epage:
This came out of my work on #12801.
I was looking at what might be appropriate to put in a `cargo-util-semver` crate and realized we have the `VersionExt` trait that exists but doesn't do much, so I dropped it.


epage:
Decided to keep it for now as I'm refactoring something else related

PR #12928, opened 2023-11-07
Do not allow empty feature name

hi-rustin:

Found it when I was working on https://github.com/rust-lang/crates.io/pull/7379

See the unit tests and integration tests.

### Additional information

Do we have any special reasons to allow empty names? I am not sure. It seems there is no way to use an empty feature.

### DIFF
--- a/src/cargo/core/summary.rs
+++ b/src/cargo/core/summary.rs
@@ -431,6 +431,9 @@ impl fmt::Display for FeatureValue {
 pub type FeatureMap = BTreeMap<InternedString, Vec<FeatureValue>>;
 
 fn validate_feature_name(pkg_id: PackageId, name: &str) -> CargoResult<()> {
+    if name.is_empty() {
+        bail!("feature name cannot be empty");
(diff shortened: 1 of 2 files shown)


weihanglo (commented):
Looks great, but this stops packages with empty feature names from compiling, right? Do we know how  many of them are in crates.io index and do we want to make them unbuildable?

hi-rustin:
> Do we know how many of them are in crates.io index and do we want to make them unbuildable?

Crates.io doesn't allow empty feature names.

https://github.com/hi-rustin/crates.io/blob/7b34715de4d7e16903f02d70bc9cfd869e266bae/src/models/krate.rs#L227

ehuss:
There was an unintentional regression in 1.60 for this due to the switch to toml_edit which followed the spec more closely.

epage:
Oh, did `toml` 0.5 disallow empty keys?

Should we disallow any other empty keys, like dependency names (only relevant when using `package`)

PR #12930, opened 2023-11-07
refactor(resolver): Consolidate logic in `VersionPreferences`

epage:

This makes customizing the resolver less intrusive by putting the logic in `VersionPreferences` and making it easy to add new priority cases to it.

In particular, this is prep for tweaking the MSRV resolver to prefer compatible versions, rather than require them.
See https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Alternative.20MSRV-aware.20resolver.20approach.3F

Each step is broken down into its own commit for easier browsing

### DIFF
--- a/src/cargo/core/resolver/version_prefs.rs
+++ b/src/cargo/core/resolver/version_prefs.rs
@@ -6,6 +6,7 @@ use std::collections::{HashMap, HashSet};
 
 use crate::core::{Dependency, PackageId, Summary};
 use crate::util::interning::InternedString;
+use crate::util::RustVersion;
 
 /// A collection of preferences for particular package versions.
 ///
@@ -19,6 +20,7 @@ pub struct VersionPreferences {
     try_to_use: HashSet<PackageId>,
     prefer_patch_deps: HashMap<InternedString, HashSet<Dependency>>,
     version_ordering: VersionOrdering,
+    max_rust_version: Option<RustVersion>,
 }
 
 #[derive(Copy, Clone, Default, PartialEq, Eq, Hash, Debug)]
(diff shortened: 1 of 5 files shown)


src/cargo/core/resolver/version_prefs.rs
```diff
@@ -62,6 +68,9 @@ impl VersionPreferences {
                     .map(|deps| deps.iter().any(|d| d.matches_id(*pkg_id)))
                     .unwrap_or(false)
         };
+        if self.max_rust_version.is_some() {
+            summaries.retain(|s| s.rust_version() <= self.max_rust_version.as_ref());
```
Eh2406, line 72: Should this still be called `sort_summaries` if it can remove versions?

Eh2406:
Looked this over and LGTM

src/cargo/core/resolver/version_prefs.rs
epage: That name mismatch goes back to when I added `-Zdirect-minimal-versions`.

Have an idea for a better name?

Eh2406: Not realy. Sorry.

src/cargo/core/resolver/version_prefs.rs
```diff
@@ -39,14 +44,21 @@ impl VersionPreferences {
...
+
     /// Sort the given vector of summaries in-place, with all summaries presumed to be for
     /// the same package.  Preferred versions appear first in the result, sorted by
     /// `version_ordering`, followed by non-preferred versions sorted the same way.
     pub fn sort_summaries(
         &self,
         summaries: &mut Vec<Summary>,
-        version_ordering: VersionOrdering,
-        first_version: bool,
+        first_version: Option<VersionOrdering>,
```
ehuss, line 61: Can the doc comment be updated? It still refers to `version_ordering`. It might also be nice to add a little bit of text explaining that it can remove summaries, too.

Eh2406:
r = me when you are ready.

PR #12933, opened 2023-11-08
feat(spec): Extend PackageIdSpec with source kind + git ref for unambiguous specs

### LINKED ISSUE #10256: Multiple git dependencies with different refs confuses cargo
### Problem

If there are multiple git dependencies on the same repository with different refs, but those refs contain the same package version, then cargo gets confused.  There is no package spec that can be passed to `-p` to unambiguously refer to them. Thus commands like `cargo update -p my-git-dep` can't work.

### Steps

Example using cargo's testsuite:

```rust
#[cargo_test]
fn multiple_git_same_version() {
    // Test what happens if different packages refer to the same git repo with
    // different refs, and the package version is the same.
    let (xyz_project, xyz_repo) = git::new_repo("xyz", |project| {
        project
            .file("Cargo.toml", &basic_lib_manifest("xyz"))
            .file("src/lib.rs", "fn example() {}")
    });
    let rev1 = xyz_repo.revparse_single("HEAD").unwrap().id();
    xyz_project.change_file("src/lib.rs", "pub fn example() {}");
    git::add(&xyz_repo);
    let rev2 = git::commit(&xyz_repo);
    // Both rev1 and rev2 point to version 0.1.0.

    let p = project()
        .file(
            "Cargo.toml",
            &format!(
                r#"
                    [package]
                    name = "foo"
                    version = "0.1.0"

                    [dependencies]
                    bar = {{ path = "bar" }}
                    xyz = {{ git = "{}", rev = "{}" }}

                "#,
                xyz_project.url(),
                rev1
            ),
        )
        .file("src/lib.rs", "")
        .file(
            "bar/Cargo.toml",
            &format!(
                r#"
                    [package]
                    name = "bar"
                    version = "0.1.0"

                    [dependencies]
                    xyz = {{ git = "{}", rev = "{}" }}
                "#,
                xyz_project.url(),
                rev2
            ),
        )
        .file("bar/src/lib.rs", "")
        .build();

    p.cargo("check").run();
    p.cargo("tree")
        .with_stdout(&format!(
            "\
foo v0.1.0 ([..]/foo)
├── bar v0.1.0 ([..]/foo/bar)
│   └── xyz v0.5.0 (file://[..]/xyz?rev={}#{})
└── xyz v0.5.0 (file://[..]/xyz?rev={}#{})
",
            rev2,
            &rev2.to_string()[..8],
            rev1,
            &rev1.to_string()[..8]
        ))
        .run();
    // FIXME: This fails since xyz is ambiguous, but the
    // possible pkgids are also ambiguous.
    p.cargo("pkgid xyz")
        .with_status(101)
        .with_stderr(
            "\
error: There are multiple `xyz` packages in your project, and the specification `xyz` is ambiguous.
Please re-run this command with `-p <spec>` where `<spec>` is one of the following:
  file://[..]/xyz#0.5.0
  file://[..]/xyz#0.5.0
",
        )
        .run();
    // TODO, what should the `-p` value be here?
    //p.cargo("update -p")
}
```

### Possible Solution(s)

I'm not sure if this should be an allowed configuration, since @alexcrichton has mentioned a few times like [here](https://github.com/rust-lang/cargo/issues/8101#issuecomment-613037817) that is an invalid Cargo.lock configuration.  However, I think it probably should be supported, since otherwise it can make it hard to get all `Cargo.toml` files in sync.

I could image cargo could allow pkgid specs of the form `https://.../foo.git?rev=499e8c4bb3b75ae6c4dcc91a8bdd81947c577147` or something like that.

### Notes

There are some related issues like #8101 and #8293, though I'm not sure if those are duplicates.

### Version

```text
cargo 1.59.0-nightly (fcef61230 2021-12-17)
release: 1.59.0-nightly
commit-hash: fcef61230c3b6213b6b0d233a36ba4ebd1649ec3
commit-date: 2021-12-17
host: x86_64-apple-darwin
libcurl: 7.77.0 (sys:0.4.51+curl-7.80.0 system ssl:(SecureTransport) LibreSSL/2.8.3)
os: Mac OS 12.0.1 [64-bit]
```

epage:

This tries to add just enough information to Package ID Specs that we can be sure they are unambiguous.  On its own thats important but this will unblock #12914 so we can have a user-facing ID that can be used with cargo's CLI.

More specifically, this adds
- `git+`, etc prefixes to the scheme
  - These were previously stabilized for `cargo metadata`s source id urls
  - Like with `SourceID`, this will allow `PackageIDSpec` to generate `directory+` and `local-registry+` prefixes but not parse them.  I'm assuming that the layers where those are dealt with they won't appear here but I don't have enough experience with them
- git ref query strings for URLs

Things from `SourceID` that this does not include
- precise: this seems less related to matching (e.g. ignored for `impl Ord for SourceId`)
- canonical URL for git kinds: this could be nice for users but users aren't the target audience and we can switch to these later

Fixes #10256

Per-commit

### DIFF
--- a/src/cargo/core/package_id_spec.rs
+++ b/src/cargo/core/package_id_spec.rs
@@ -6,6 +6,7 @@ use semver::Version;
 use serde::{de, ser};
 use url::Url;
 
+use crate::core::GitReference;
 use crate::core::PackageId;
 use crate::core::SourceKind;
 use crate::util::edit_distance;
@@ -104,17 +105,47 @@ impl PackageIdSpec {
             name: String::from(package_id.name().as_str()),
             version: Some(package_id.version().clone().into()),
             url: Some(package_id.source_id().url().clone()),
-            kind: None,
+            kind: Some(package_id.source_id().kind().clone()),
         }
     }
 
     /// Tries to convert a valid `Url` to a `PackageIdSpec`.
     fn from_url(mut url: Url) -> CargoResult<PackageIdSpec> {
+        let mut kind = None;
+        if let Some((kind_str, scheme)) = url.scheme().split_once('+') {
+            match kind_str {
+                "git" => {
+                    let git_ref = GitReference::DefaultBranch;
+                    kind = Some(SourceKind::Git(git_ref));
+                    url = strip_url_protocol(&url);
+                }
+                "registry" => {
+                    kind = Some(SourceKind::Registry);
+                    url = strip_url_protocol(&url);
+                }
+                "sparse" => {
+                    kind = Some(SourceKind::SparseRegistry);
+                    // Leave `sparse` as part of URL, see `SourceId::new`
+                    // url = strip_url_protocol(&url);
+                }
+                "path" => {
+                    if scheme != "file" {
+                        anyhow::bail!("`path+{scheme}` is unsupported; `path+file` and `file` schemes are supported");
+                    }
+                    kind = Some(SourceKind::Path);
+                    url = strip_url_protocol(&url);
+                }
+                kind => anyhow::bail!("unsupported source protocol: {kind}"),
+            }
+        }
+
         if url.query().is_some() {
             bail!("cannot have a query string in a pkgid: {}", url)
         }
+
         let frag = url.fragment().map(|s| s.to_owned());
         url.set_fragment(None);
+
         let (name, version) = {
             let mut path = url
                 .path_segments()
@@ -148,7 +179,7 @@ impl PackageIdSpec {
             name,
             version,
             url: Some(url),
-            kind: None,
+            kind,
         })
     }
 
@@ -173,6 +204,14 @@ impl PackageIdSpec {
         self.url = Some(url);
     }
 
+    pub fn kind(&self) -> Option<&SourceKind> {
+        self.kind.as_ref()
+    }
+
+    pub fn set_kind(&mut self, kind: SourceKind) {
+        self.kind = Some(kind);
+    }
+
     /// Checks whether the given `PackageId` matches the `PackageIdSpec`.
     pub fn matches(&self, package_id: PackageId) -> bool {
         if self.name() != package_id.name().as_str() {
@@ -187,6 +226,12 @@ impl PackageIdSpec {
 
         if let Some(u) = &self.url {
             if u != package_id.source_id().url() {
+                return false;
+            }
+        }
+
+        if let Some(k) = &self.kind {
+            if k != package_id.source_id().kind() {
                 return false;
             }
         }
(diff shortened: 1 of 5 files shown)


epage:
@rfcbot fcp merge

src/doc/src/reference/pkgid-spec.md
```diff
@@ -38,28 +40,32 @@ that come from different sources such as different registries.
...
-| `https://github.com/rust-lang/crates.io-index#regex`        | `regex` | `*`     |
-| `https://github.com/rust-lang/crates.io-index#regex@1.4.3`  | `regex` | `1.4.3` |
+| Spec                                                              | Name    | Version |
+|:------------------------------------------------------------------|:-------:|:-------:|
+| `regex`                                                           | `regex` | `*`     |
+| `regex@1.4`                                                       | `regex` | `1.4.*` |
+| `regex@1.4.3`                                                     | `regex` | `1.4.3` |
+| `https://github.com/rust-lang/crates.io-index#regex`              | `regex` | `*`     |
+| `https://github.com/rust-lang/crates.io-index#regex@1.4.3`        | `regex` | `1.4.3` |
+| `sparse+https://github.com/rust-lang/crates.io-index#regex@1.4.3` | `regex` | `1.4.3` |
```
ehuss, line 50: To be pedantic, this won't work since that isn't a sparse URL (it is a git one).
I don't think there is a way to directly refer to index.crates.io, due to the way it works as a "replaced" source. My only thought is to use something like `sparse+https://example.com/rindex/#regex@1.4.3`, though that would no longer be on "crates.io". It could also just be `sparse+https://index.crates.io/#regex@1.4.3`, though that technically won't work.

ehuss:
There's a small bug, perhaps not directly related to this PR, where `?` interferes with package name globbing. 

`cargo doc -p git+https://example.com/?branch=dev#bar@1.2.3` doesn't work because cargo sees the `?` and thinks it is a glob pattern, and thus takes a different code path. I believe this stems from [`opt_patterns_and_names`](https://github.com/rust-lang/cargo/blob/b97cffe0699b81826dcfb0125550e5c0d089b86c/src/cargo/ops/cargo_compile/packages.rs#L190-L205), which probably needs to be a little smarter about what it determines is a glob pattern.

This generally only affects `cargo doc` because with resolver=2, you cannot use `-p` for non-workspace-members for all other commands.

I don't think this should be addressed in this PR, and seems like an extremely low priority, at least for now.

src/cargo/core/package_id_spec.rs
```diff
@@ -101,16 +105,57 @@ impl PackageIdSpec {
...
 
     /// Tries to convert a valid `Url` to a `PackageIdSpec`.
     fn from_url(mut url: Url) -> CargoResult<PackageIdSpec> {
-        if url.query().is_some() {
-            bail!("cannot have a query string in a pkgid: {}", url)
+        let mut kind = None;
+        if let Some((kind_str, scheme)) = url.scheme().split_once('+') {
+            match kind_str {
+                "git" => {
+                    let git_ref = GitReference::from_query(url.query_pairs());
```
weihanglo, line 118: > git ref query strings for URLs

We've got a fair amount of issue about git dependencies ambiguity because of not checking in the actual rev. This patch seems to fix <https://github.com/rust-lang/cargo/issues/10256>. Do you want to include that demo test case as well?

src/cargo/core/package_id_spec.rs
```diff
@@ -101,16 +105,57 @@ impl PackageIdSpec {
...
+                    kind = Some(SourceKind::Registry);
+                    url = strip_url_protocol(&url);
+                }
+                "sparse" => {
+                    if url.query().is_some() {
+                        bail!("cannot have a query string in a pkgid: {}", url)
+                    }
+                    kind = Some(SourceKind::SparseRegistry);
+                    // Leave `sparse` as part of URL
+                    // url = strip_url_protocol(&url);
```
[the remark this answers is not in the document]
epage: I've added notes referencing the the primary explanation

weihanglo (approved):

weihanglo:
Going to merge this. I don't think we need to wait for 10 day long.

If people think it is wrong, or have a better alternative, please call it out and we can reconsider.

@bors r+

PR #12938, opened 2023-11-08
refactor(source): Prepare for new PackageIDSpec syntax

epage:

This adds tests and refactors to prepare for #12933

### DIFF
--- a/src/cargo/core/source_id.rs
+++ b/src/cargo/core/source_id.rs
@@ -188,17 +188,7 @@ impl SourceId {
         match kind {
             "git" => {
                 let mut url = url.into_url()?;
-                let mut reference = GitReference::DefaultBranch;
-                for (k, v) in url.query_pairs() {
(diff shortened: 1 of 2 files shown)


src/cargo/core/source_id.rs
```diff
@@ -752,6 +752,20 @@ impl PartialEq for SourceIdInner {
     }
 }
 
+impl SourceKind {
+    pub fn protocol(&self) -> Option<&str> {
```
weihanglo, line 746: Do we need to `pub` this?

It look like a implementation details for `Display` impl, especially that `SparseRegistry` return nothing here.

epage: Huh, not sure why I made it `pub`.  Switched to `pub(crate)`.

weihanglo (approved):
All refactors look great now! Thank you :)

PR #12926, opened 2023-11-06
refactor(util): Prepare for splitting out semver logic

epage:
### What does this PR try to resolve?

Like #12924, this was cleanup I noticed as I was looking to pull out our reusable semver code for #12801

PR #12940, opened 2023-11-08
refactor(util): Pull out `mod util_semver`

epage:
This `mod` is a proposal for what a new package would look like. This needs to be split out so a future `util_manifest_schema` package can depend on it (#12801).

This doesn't address where `RustVersion` should live (along with `PackageIdSpec`).

### Additional information

This builds on the work from #12924 and #12926

### DIFF
--- a/src/cargo/util_semver.rs
+++ b/src/cargo/util_semver.rs
@@ -1,0 +1,195 @@
+use std::fmt::{self, Display};
+
+use semver::{Comparator, Op, Version, VersionReq};
+use serde_untagged::UntaggedEnumVisitor;
+
+pub trait VersionExt {
+    fn is_prerelease(&self) -> bool;
+
+    fn to_exact_req(&self) -> VersionReq;
+}
+
+impl VersionExt for Version {
+    fn is_prerelease(&self) -> bool {
+        !self.pre.is_empty()
+    }
+
+    fn to_exact_req(&self) -> VersionReq {
+        VersionReq {
+            comparators: vec![Comparator {
+                op: Op::Exact,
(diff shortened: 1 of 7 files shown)


src/cargo/util_semver.rs
```diff
@@ -0,0 +1,195 @@
...
+            minor: Some(ver.minor),
+            patch: Some(ver.patch),
+            pre,
+            build,
+        }
+    }
+}
+
+impl std::str::FromStr for PartialVersion {
+    type Err = anyhow::Error;
```
weihanglo, line 107: Not a blocker. Are we planning to use thiserror or custom error type instead of anyhow?

src/cargo/util_semver.rs
```diff
@@ -0,0 +1,195 @@
+use std::fmt::{self, Display};
```
weihanglo, line 1: Why is `OptVersionReq` left out in the party? Because it is not a dependency of `schema.rs`?

src/cargo/util_semver.rs
epage: Would be good to do at some point in this process.

src/cargo/util_semver.rs
epage: Not quite sure where it should live.  It isn't just a simple semver extension like `VersionExt` or `PartialVersion` but is tightly coupled to specific parts of cargo, like `RustVersion`.

weihanglo (approved):
This looks fine. Let's not block on the progress and merge.

For where to put `RustVersion` and `OptVersionReq`, we can decide later.

Issue #12790, opened 2023-10-07
Cargo publish excludes files called `target`


KarelPeeters:
### Problem

When publishing or packaging a crate that contains a file called `target` (different from the typical root `target` folder), the file is not included. This happens even though the `target` file is not ignored by either `.gitignore` or `.git/info/exclude`. It's also not a glob evaluation issue, this is still an issue if `.gitignore` is completely empty.

This also only happens if the `target` file is currently tracked in git. This is not that important by itself, but it might help figuring out what exactly causes the bug.

### Steps

I've made a repository to reproduce this bug: https://github.com/KarelPeeters/cargo_bug_package_target

The steps to follow:

* clone https://github.com/KarelPeeters/cargo_bug_package_target
* check the list of files that will be packaged
```
> cargo package --list --allow-dirty
.gitignore
Cargo.lock
Cargo.toml
Cargo.toml.orig
README.md
Readme.md
data/not_target
src/main.rs
```
note that `data/target` is missing, this is the main bug.
* ensure that target is not "tracked" any more by moving or deleting `.git` entirely
```
> mv .git .not_git
```
* check the list of files that will be packaged again
```
> cargo package --list --allow-dirty
Cargo.lock
Cargo.toml
Cargo.toml.orig
README.md
Readme.md
data\not_target
data\target
src\main.rs
```
Note that `data/target` is not missing anymore, this shows that the bug only happens for tracked files.

### Possible Solution(s)

I assume this is caused by some bad interaction in `build_ar_list` between the `VCSInfo` stuff and the hardcoded check for a filename of `"target"` here: https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/cargo_package.rs#L934

I'm not sure what much of the context there is trying to do, so I don't know what the right fix is.

One issue to keep in mind is that multiple crates that share a repository while not being a workspace can have multiple `target` folders. So a simple check like "only ignore the root `target` directory" is also not good enough. There should be existing infrastructure to figure out what the real `target` directory is though, since `cargo clean` (like many other subcommands) needs that.

### Notes

I noticed this when trying to vendor some cuda headers and associated files, which is neccesary to get docs.rs to build one of my crates correctly.

One of these files happens to be called `target`.

Here it is in my github repository: https://github.com/KarelPeeters/Kyanite/tree/1eb40f2dde70d3432f218bfbaa5c7d37eebbd394/kn-cuda-sys/doc_headers/cuda_include/nv
[...]

PR #12944, opened 2023-11-09
Only filter out target if its in the package root

linyihai:
Only filter out target if its in the package root. Fixed https://github.com/rust-lang/cargo/issues/12790

Add two testcase in tests/testsuite/package.rs for this PR.

- `include_files_called_target_project` testcase test the logic for none git repository. By the way, our PR was based on git repository,  so this testcase was a little irrelevant.
-  `include_files_called_target_git` testcase was made for git repository.  There are two cases here, one is the target in the uncommitted state, at this time should not be included, and one is the target in the committed state, at this time should be included

### Additional information

### DIFF
--- a/tests/testsuite/package.rs
+++ b/tests/testsuite/package.rs
@@ -3162,10 +3162,10 @@ src/main.rs
 fn include_files_called_target_git() {
     // https://github.com/rust-lang/cargo/issues/12790
     // files and folders called "target" should be included, unless they're the actual target directory
-    let (p, repo) = git::new_repo("target_uncommitted", |p| init_and_add_inner_target(p));
+    let (p, repo) = git::new_repo("foo", |p| init_and_add_inner_target(p));
     // add target folder but not committed.
-    let _ = fs::create_dir_all(&repo.workdir().unwrap().join("target/foo.txt")).unwrap();
-
+    _ = fs::create_dir(p.build_dir()).unwrap();
+    _ = fs::write(p.build_dir().join("foo.txt"), "").unwrap();
     p.cargo("package -l")
         .with_stdout(
             "\
@@ -3183,9 +3183,8 @@ src/main.rs
         .run();
 
     // if target is committed, it should be include.
-    let p = git::new("target_committed", |p| {
-        init_and_add_inner_target(p).file("target/foo.txt", "")
-    });
+    git::add(&repo);
+    git::commit(&repo);
     p.cargo("package -l")
         .with_stdout(
             "\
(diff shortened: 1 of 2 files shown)


weihanglo (changes requested):
Thanks for the contribution!

Could you add some tests for this patch? Such as one commit adding a test to demonstrate the current "bad" behavior, and the next commit to fix both bug and test.

KarelPeeters:
I wrote a unit test for this a while back, I've pushed it here: https://github.com/KarelPeeters/cargo/commits/package-target

Feel free to cherry pick or take inspiration!

linyihai:
> I wrote a unit test for this a while back, I've pushed it here: https://github.com/KarelPeeters/cargo/commits/package-target
> 
> Feel free to cherry pick or take inspiration!

That's awesome ! It's really a big help and can save a lot of work. Good Job.

tests/testsuite/package.rs
```diff
@@ -3132,3 +3133,94 @@ See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for
...
+derp/target/foo.txt
+src/main.rs
+",
+        )
+        .run();
+
+    // if target is committed, it should be include.
+    let p = git::new("target_committed", |p| {
+        init_and_add_inner_target(p).file("target/foo.txt", "")
+    });
```
weihanglo, line 3188: Would be better if we can reuse the git repo instead of creating a new one.

```suggestion
    // if target is committed, it should be include.
    let mut index = t!(repo.index());
    t!(index.add_path(Path::new("target/foo.txt")));
    t!(index.write());
    git::commit(&repo);

```

And you might need to change code on line 3167 accordling to create a real foo.txt file:

```rust
    // add target folder but not committed.
    _ = fs::create_dir_all(p.build_dir()).unwrap();
    _ = fs::write(p.build_dir().join("foo.txt"), "").unwrap();
```

weihanglo (commented):
Thanks. Just some nits in tests.

tests/testsuite/package.rs
weihanglo: Okay. I can see they are not equivalent, but feel that reusing the git repo make the intent a bit clear.

linyihai: > Okay. I can see they are not equivalent, but feel that reusing the git repo make the intent a bit clear.

I had reused the repo to commit the file, it looks more tidy, thanks.

weihanglo (approved):

weihanglo:
Looks awesome. Thank you for the contribution!

@bors r+

KarelPeeters:
@linyihai Thanks a lot for fixing this bug (and the tests)!

PR #12948, opened 2023-11-09
refactor(toml): Simplify code to make schema split easier

epage:

This is a follow up to #12911 and is prep for #12801.

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -2279,15 +2279,6 @@ impl schema::TomlProfile {
     }
 }
 
-impl schema::TomlTarget {
-    fn name(&self) -> String {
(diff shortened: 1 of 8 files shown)


src/cargo/util/toml/mod.rs
```diff
@@ -1538,7 +1537,7 @@ impl schema::InheritableFields {
     }
 
     /// Gets a workspace dependency with the `name`.
-    pub fn get_dependency(
+    fn get_dependency(
```
weihanglo, line 1488: While I understand and agree the purpose of making them private, I am not sure if other projects depends on these and might be unhappy.

weihanglo (approved):
Impressive! Thank you for the huge cleanup.

Issue #6179, opened 2018-10-16
Allow specifying a set of supported targets in Cargo.toml


luser:
Spun off from the specific proposal in https://github.com/rust-lang/cargo/issues/4544#issuecomment-357371733
"A random solution which may help solve this though is to perhaps list the valid targets for a project in a workspace root Cargo.toml. That way Cargo could be smarter and just vendor dependencies for those targets (and alter resolution so it won't include winapi in the lock file)"

This would help projects like Firefox and Fuchsia that vendor their dependencies by allowing them to avoid vendoring crates that are only necessary for platforms that the project does not support.

cc @cramertj

alexcrichton:
This seems plausible to me! Cargo has to execute rustc to learn about cfg information for targets, so that may be one roadblock to getting this working but otherwise I think it in theory shouldn't be too too hard to plumb through what we've got today

cramertj:
Similarly, it seems like you could also plumb through the set of features with which each library will be built, so that there's no need to pull in optional dependencies.

luser:
I poked around that a bit and that already seems to work. If you have a dependency on crate `A` and `A` has an optional dependency on `B`, you won't wind up with `B` in your Cargo.lock unless you enable that feature.

mankinskin:
Another idea is to declare something like `non-targets` to explicitly exclude targets from compatibility for a crate. This could for example be useful when you never intend a crate to target something special like Wasm.

kvark:
This would help a problem we have in Firefox, where some of the dependencies may have WASM-specific dependencies (https://github.com/grovesNL/glow), which are dragged into the 3rd party sources by `cargo vendor`.

flukejones:
I've just recently hit on a circumstance where this functionality would be very useful: preventing cargo (and vendoring by extension) from pulling in crates that will never be built on the intended target.

Or at least that would be my hope.

An example is that I build and maintain a particular project, and this project has many dependencies. Various of these dependencies have `winapi` as a dependency themselves. This in itself is not an issue really, as the `winapi` crate is behind a target.cfg flag, so it never gets built on Linux or Macos.

[...]

PR #12949, opened 2023-11-09
Filter `cargo-credential-*` dependencies by OS

### LINKED ISSUE #12945: non-target-specific / unconditional dependencies on potentially unusable credential providers
I see that different credentials providers were recently merged: https://github.com/rust-lang/cargo/pull/12334
Is there a reason why the dependencies for the various credentials providers are unconditional? This is kind of awkward, especially for Linux distributions, where we package the "cargo" crate (for example, as a dependency of "cargo-c"), and cannot package macOS- and Windows-specific crates like windows-sys or security-framework.

I would have thought that the libsecret provider could be specific to Linux/BSD, the macOS keychain provider specific to ... well, macOS, and the wincred provider specific to windows targets. Right cargo pulls in *all* of them unconditionally - should they be optional and / or target-scoped?

PS: The published crates for the new cargo-credentials* crates also don't contain the required MIT / Apache-2.0 license files. I can submit a PR for that, if it helps.

cuviper:

The `cargo-credential-*` crates have OS-specific functionality, with `cfg` for an "unsupported" fallback, and these are unconditional dependencies of `cargo`. In distros like Fedora that package dependencies individually, that means these crates still have to be packaged even when they are useless on Linux. (Or else patch them out as a downstream change.) Instead, we can filter those dependencies in `Cargo.toml` and add the fallback at the point of use.

Fixes #12945.

### Additional information

We could further *remove* the `cfg`-unsupported fallbacks from the individual crates, and just have them `#![cfg(..)]` themselves globally. I haven't done that yet, because it would look like a big change for mostly just whitespace removing the nested module. I'm happy to add that if desired though.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,9 +128,6 @@ anyhow.workspace = true
 base64.workspace = true
 bytesize.workspace = true
 cargo-credential.workspace = true
-cargo-credential-libsecret.workspace = true
-cargo-credential-macos-keychain.workspace = true
-cargo-credential-wincred.workspace = true
 cargo-platform.workspace = true
 cargo-util.workspace = true
 clap = { workspace = true, features = ["wrap_help"] }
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
I am good with this change. Thank you!

weihanglo:
> We could further remove the cfg-unsupported fallbacks from the individual crates,

This was a workaround to make cargo buildable on every supported platform, as well as those crates themselves, before any progress made in <https://github.com/rust-lang/cargo/issues/6179>.

arlosi:
@bors r+

Thanks!

PR #12950, opened 2023-11-10
fix(resolver): Prefer MSRV, rather than ignore incompatible

epage:

This is another experiment for #9930.

Comparing preferring over exclusively using MSRV compatible:

Benefits
- Better error messages
- `--ignore-rust-version` is implicitly sticky

Downsides
- Can't backtrack for MSRV compatible version
- Still requires workspace-wide MSRV (compared to our desired end state of declaring MSRV as yet another dependency)

### Additional information

Note: `--ignore-rust-version` is not yet implemented for the resolver.

This builds on #12930

### DIFF
--- a/src/cargo/core/resolver/version_prefs.rs
+++ b/src/cargo/core/resolver/version_prefs.rs
@@ -58,10 +58,10 @@ impl VersionPreferences {
     ///
     /// Sort order:
     /// 1. Preferred packages
-    /// 2. `first_version`, falling back to [`VersionPreferences::version_ordering`] when `None`
+    /// 2. [`VersionPreferences::max_rust_version`]
+    /// 3. `first_version`, falling back to [`VersionPreferences::version_ordering`] when `None`
     ///
     /// Filtering:
-    /// - [`VersionPreferences::max_rust_version`]
     /// - `first_version`
     pub fn sort_summaries(
         &self,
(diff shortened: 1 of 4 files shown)


Eh2406:
At a code level this looks fine, although it would be nice to have a test that correctly backtracked before and does not do so now.

How are we going to evaluate if this provides a better user experience for the people attempting to maintain MSRV?

epage:
> At a code level this looks fine, although it would be nice to have a test that correctly backtracked before and does not do so now.

I feel fairly weak in creating resolver test cases...

> How are we going to evaluate if this provides a better user experience for the people attempting to maintain MSRV?

I had considered trying to find ways to evaluate the two options until I realized
- We might be able to stabilize this approach relatively easily as the default approach
- We can always continue on the hard-error path later, even if this is stabilized

So I instead see the questions as
- Is my base assumption correct that this can be an on-by-default approach?
- Does this provide a sufficient quality of an experience?

Eh2406:
> I feel fairly weak in creating resolver test cases...

Root depends on `a`. `a` comes in two versions, the newer version depends on `b` and the older version has no dependencies. The only version of `b` has an incompatible MSRV requirement. If I've constructed this correctly... Before this PR you will get the older version of `a`. After this PR you will get `a` and `b` and a MSRV error.

> Is my base assumption correct that this can be an on-by-default approach?

I'm wondering if this should be an always on approach. We sort lock file versions first with no way to opt out of it. To be blunt, I don't want to have to write the description for the flag to turn off this functionality. :-P But then how does a library that has an MSRV policy test with the latest versions of everything. 🤔 

> Does this provide a sufficient quality of an experience?

I have never maintained a library with an MSRV policy, so I don't know that I'm the person to ask. My sense is that the hard part is dealing with a transitive dependency adding a new dependency on something that does not work with your MSRV. Which this PR does not help with. On the other hand, many real-world dependency trees are smaller and simpler than I imagine. And this seems pretty painless for the easy case.

epage:
> I'm wondering if this should be an always on approach. We sort lock file versions first with no way to opt out of it.

I agree, I'm hoping we can just make this the default with little fanfare / ceremony.

> To be blunt, I don't want to have to write the description for the flag to turn off this functionality. :-P But then how does a library that has an MSRV policy test with the latest versions of everything. 

I was going to leverage `--ignore-rust-version`; I just haven't plumbed it through yet.  It was a lower priority to me since the behavior is opt-in at the moment

> I have never maintained a library with an MSRV policy, so I don't know that I'm the person to ask. My sense is that the hard part is dealing with a transitive dependency adding a new dependency on something that does not work with your MSRV. Which this PR does not help with. On the other hand, many real-world dependency trees are smaller and simpler than I imagine. And this seems pretty painless for the easy case.

Good point to call out that the quality of this is predicated on "if you care about MSRV, your dependencies should at least declare an MSRV".  There are likely corner cases where that can still have problems from backtracking but it is unlikely to be a problem.

Eh2406:
> Good point to call out that the quality of this is predicated on "if you care about MSRV, your dependencies should at least declare an MSRV". There are likely corner cases where that can still have problems from backtracking but it is unlikely to be a problem.

If that's the assumption of this mechanism, should it consider a version that does not state and MSRV as being sorted with the noncompatible versions?

epage:
Alright, a backtracking test is now in

PR #12951, opened 2023-11-10
fix: preserve jobserver file descriptors on rustc invocation in `fix_exec_rustc`

belovdv:
Similar to #12447

Command `cargo fix` invokes `cargo rustc`. It sends environment variable which points to closed file descriptors.

This PR makes cargo compatible with https://github.com/rust-lang/rust/pull/113730.

It should be enough to have tests in rustc. It seems to be good to have more centralized way to pass jobserver.


weihanglo (approved):
Make sense. Thank you!

PR #7886, opened 2020-02-14
Link the licenses into crates/cargo-platform

cuviper:
The licenses should be included in the package published on crates.io

PR #12953, opened 2023-11-10
credential: include license files in all published crates

decathorpe:

It appears that the addition of license files was missed when the cargo-credential* crates were added to this cargo workspace. This PR adds symbolic links to the license files so that `cargo publish` picks them up and includes the files when publishing them to crates.io (which is a requirement for both the Apache-2.0 and the MIT license).

Running `cargo package` for the cargo-credential* crates should result in a copy of the LICENSE-APACHE and LICENSE-MIT files to be included in the "packaged" files.

### Additional information

Similar changes have been pushed for other workspace members in the past, for example https://github.com/rust-lang/cargo/pull/7886

### DIFF
--- a/credential/cargo-credential-1password/LICENSE-APACHE
+++ b/credential/cargo-credential-1password/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../../LICENSE-APACHE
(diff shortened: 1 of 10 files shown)


weihanglo (approved):

weihanglo:
Thanks for the contribution!

@bors r+

decathorpe:
Thanks!

PR #6358, opened 2018-11-28
Add --quiet option for `cargo test`

collin5:
Fixes #4325

PR #12959, opened 2023-11-12
Fix --quiet being used with nested subcommands.

### LINKED ISSUE #12957: Determine how to handle `--quiet` in commands with subcommands
Commands that have subcommands (like `cargo report future-incompat`, `cargo config get`, `cargo owner add`, `cargo clean gc`, etc.) currently don't support the `--quiet` flag. The problem is:

[`arg_quiet()`](https://github.com/rust-lang/cargo/blob/9a1b0924c6ea3ac306b663aa73347d3b6a59e7c3/src/cargo/util/command_prelude.rs#L366-L380) is specified on every command individually, instead of using `.global(true)` like `--verbose` does. This is done in order to customize the help string on some commands (like [`cargo test`](https://github.com/rust-lang/cargo/blob/9a1b0924c6ea3ac306b663aa73347d3b6a59e7c3/src/bin/cargo/commands/test.rs#L27-L33)). 

AFAIK, we can't have both `.global(true)` and a per-command customized help text.

This means that commands like `cargo owner add` don't support the `--quiet` flag. It doesn't work to add `.arg_quiet()` on the subcommand because [`config_configure`](https://github.com/rust-lang/cargo/blob/9a1b0924c6ea3ac306b663aa73347d3b6a59e7c3/src/bin/cargo/cli.rs#L375) only looks at the arguments of the top-level command, and the global arguments. It isn't capable of seeing subcommands.

I don't really know how to fix this. Some ideas:

- Change `config_configure` to somehow "see" subcommand arguments.
- Have each subcommand manually set the quiet argument on the config. This might have some awkwardness, since `config_configure` handles things like clashing arguments (I suspect clap's conflicts_with won't help with that).
- Drop support for per-command customized help text, and use `.global(true)`.

ehuss:
This fixes an issue where `--quiet` doesn't work with commands that have subcommands. This is because `config_configure` only looks at the global and top-level subcommand, and not deeper subcommands. The issue was that `--quiet` was not defined as a global flag. This was changed in https://github.com/rust-lang/cargo/pull/6358 in order to give a better help message for `cargo test --quiet`. I don't remember if clap just didn't support overriding at the time, or if we just didn't know how it worked. Anyways, it seems to work to override it now, so I think it should be fine to mark it as global.

This should bring in `--quiet` more in-line with how `--verbose` works. This means that `--quiet` is now accepted with `cargo report`, `cargo help`, and `cargo config`.

This also fixes `--quiet` with `cargo clean gc`.

This should also help with supporting `--quiet` with the new `cargo owner` subcommands being added in
https://github.com/rust-lang/cargo/pull/11879.

As for this diff, the help text changes because the `--quiet` flag is changing position (except for the 3 commands mentioned above where it is now added).

Fixes #12957

### DIFF
--- a/crates/xtask-bump-check/src/xtask.rs
+++ b/crates/xtask-bump-check/src/xtask.rs
@@ -41,7 +41,11 @@ pub fn cli() -> clap::Command {
             .action(ArgAction::Count)
             .global(true),
         )
-        .arg_quiet()
+        .arg(
+            flag("quiet", "Do not print cargo log messages")
+                .short('q')
+                .global(true),
+        )
         .arg(
             opt("color", "Coloring: auto, always, never")
                 .value_name("WHEN")
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -589,7 +589,7 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
             .action(ArgAction::Count)
             .global(true),
         )
-        .arg_quiet()
+        .arg(flag("quiet", "Do not print cargo log messages").short('q').global(true))
         .arg(
             opt("color", "Coloring: auto, always, never")
                 .value_name("WHEN")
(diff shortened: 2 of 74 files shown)


epage (approved):

PR #12954, opened 2023-11-10
refactor(toml): Improve consistency

epage:
### What does this PR try to resolve?

As I was working towards #12801, these inconsistencies bothered me.  Hopefully this will also make it easier to discover special cases in the schema and deal with them properly.

PR #12960, opened 2023-11-12
refactor(toml): Consistently lead with 'Toml' prefix

epage:

Like with #12954, I noticed this inconsistency while working towards #12801 and wanted to fix it.


weihanglo (approved):
Seems reasonable to me.

PR #12962, opened 2023-11-12
Improve about information of `cargo search`

hi-rustin:

Since it supports multiple registries, it's better to mention that it's not limited to just crates.io.

- [x] See the docs changes.

### Additional information

None

### DIFF
--- a/src/bin/cargo/commands/search.rs
+++ b/src/bin/cargo/commands/search.rs
@@ -6,7 +6,7 @@ use cargo::ops;
 
 pub fn cli() -> Command {
     subcommand("search")
-        .about("Search packages in crates.io")
+        .about("Search packages in the registry. Default registry is crates.io")
         .arg(Arg::new("query").value_name("QUERY").num_args(0..))
         .arg(
             opt(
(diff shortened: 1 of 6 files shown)


hi-rustin:
Wried. I supposed I needed to update the man page, but it seemed we didn't check it.

weihanglo (changes requested):
Thanks for the contribution!

> Wried. I supposed I needed to update the man page, but it seemed we didn't check it.

Guess we might want to update manpage here as well?
<https://github.com/rust-lang/cargo/blob/master/src/doc/man/cargo-search.md>

hi-rustin:
> Guess we might want to update manpage here as well?
> [`master`/src/doc/man/cargo-search.md](https://github.com/rust-lang/cargo/blob/master/src/doc/man/cargo-search.md?rgh-link-date=2023-11-12T16%3A13%3A46Z)

I see. We only check if we generate it from markdown correctly.

hi-rustin (commented):
🔢 Self-check

weihanglo (approved):

PR #12965, opened 2023-11-12
fix: do not panic when failed to parse rustc commit-hash

weihanglo:

In some situation the commit-hash in `rustc -vV` output is "unknown".
The debug assertion must not block any progress on others like miri,

See https://github.com/rust-lang/miri/issues/3160

### DIFF
--- a/src/cargo/util/rustc.rs
+++ b/src/cargo/util/rustc.rs
@@ -83,14 +83,20 @@ impl Rustc {
             )
         })?;
         let commit_hash = extract("commit-hash: ").ok().map(|hash| {
-            debug_assert!(
-                hash.chars().all(|ch| ch.is_ascii_hexdigit()),
-                "commit hash must be a hex string, got: {hash:?}"
-            );
-            debug_assert!(
-                hash.len() == 40 || hash.len() == 64,
-                "hex string must be generated from sha1 or sha256 (i.e., it must be 40 or 64 characters long)\ngot: {hash:?}"
-            );
+            // Possible commit-hash values from rustc are SHA hex string and "unknown". See:
+            // * https://github.com/rust-lang/rust/blob/531cb83fc/src/bootstrap/src/utils/channel.rs#L73
+            // * https://github.com/rust-lang/rust/blob/531cb83fc/compiler/rustc_driver_impl/src/lib.rs#L911-L913
+            #[cfg(debug_assertions)]
+            if hash != "unknown" {
+                debug_assert!(
+                    hash.chars().all(|ch| ch.is_ascii_hexdigit()),
+                    "commit hash must be a hex string, got: {hash:?}"
+                );
+                debug_assert!(
+                    hash.len() == 40 || hash.len() == 64,
+                    "hex string must be generated from sha1 or sha256 (i.e., it must be 40 or 64 characters long)\ngot: {hash:?}"
+                );
+            }
             hash.to_string()
         });
 


src/cargo/util/rustc.rs
```diff
@@ -82,17 +82,19 @@ impl Rustc {
...
-            );
-            debug_assert!(
-                hash.len() == 40 || hash.len() == 64,
-                "hex string must be generated from sha1 or sha256 (i.e., it must be 40 or 64 characters long)\ngot: {hash:?}"
-            );
-            hash.to_string()
-        });
+        let commit_hash = extract("commit-hash: ").ok()
+            .filter(|hash| {
+                if !hash.chars().all(|ch| ch.is_ascii_hexdigit()) {
```
RalfJung, line 87: Maybe add a comment explaining why the code does what it does now?

weihanglo: Thanks. Updated!

src/cargo/util/rustc.rs
```diff
@@ -82,17 +82,22 @@ impl Rustc {
...
+        let commit_hash = extract("commit-hash: ").ok()
+            .filter(|hash| {
+                // Ensure that commit hash is always a SHA hex string.
+                // If it is not, like `commit-hash: unknown`,
+                // then treat it as if the entire `commit-hash:` line never exist.
+                if !hash.chars().all(|ch| ch.is_ascii_hexdigit()) {
+                    debug!("commit hash must be a hex string, got: {hash:?}");
+                    return false;
+                }
+                if hash.len() != 40 && hash.len() != 64 {
```
epage, line 94: Do we have a way to verify that if rustc changes the [commit hash format](https://git-scm.com/docs/hash-function-transition), that we'll update these checks?

weihanglo: I cannot think of a way now 😞.

We've already covered both SHA1 (40 chars) and SHA2 (64 chars) hash algorithms used by Cargo. I guess rustc won't change to any other format when Git is still in use. If people are going to add a new hash format into Git, it will be breaking news than this fails.

Normal developers also don't care about the hash here, since they always get a real rustc version, and the hash is only used by `-Ztrim-paths` now.

epage: > I cannot think of a way now 😞.

Do we have a test that the commit hash remapping works?  Seems like that would cover this.

weihanglo: If I remember correctly, the assertion of hash part is omitted, as we don't know whether people run this with a rustc with or without git.

RalfJung: The alternative would be to specifically allow the hash to be `"unknown"`, but still assert in all the other cases.

weihanglo: Or should we just remove these checks, since they are not really important?

epage: The length checks?  sure.  Thats basically what Ralf suggested since our `is_ascii_hexdigit` check will filter out `unknown`.

Alternatively, we do `if hash.len() < 40` to avoid short garbage (random words).  Anything that looks like a hash that is longer is probably good enough.

Either way, I'm fine.

src/cargo/util/rustc.rs
```diff
@@ -83,14 +83,20 @@ impl Rustc {
...
-            );
-            debug_assert!(
-                hash.len() == 40 || hash.len() == 64,
-                "hex string must be generated from sha1 or sha256 (i.e., it must be 40 or 64 characters long)\ngot: {hash:?}"
-            );
+            // Possible commit-hash values from rustc are SHA hex string and "unknown". See:
+            // * https://github.com/rust-lang/rust/blob/531cb83fc/src/bootstrap/src/utils/channel.rs#L73
+            // * https://github.com/rust-lang/rust/blob/531cb83fc/compiler/rustc_driver_impl/src/lib.rs#L911-L913
+            #[cfg(debug_assertions)]
+            if hash != "unknown" {
```
epage, line 90: Oops, I thought we were going to filter out `unknown`.

Which approach does trim paths need?

weihanglo: Trim-paths needs merely a string as a fallback when rustc version is missing to construct `rustc/<commit-hash|whatever>`.

PR #12966, opened 2023-11-13
Bump to 0.77.0; update changelog

weihanglo:
[rendered](https://github.com/weihanglo/cargo/blob/version-bump/CHANGELOG.md)

PR #12968, opened 2023-11-13
Respect `rust-lang/rust`'s `omit-git-hash`

pietroalbini:
The `config.toml` file in `rust-lang/rust` has the `omit-git-hash` option, which prevents git information from being embedded into binaries. This works for most tools, as they rely on the git information provided by bootstrap through environment variables.

Cargo does its own git detection in its build script though, which didn't adhere to to that option. This changes that by skipping git detection whenever bootstrap signals the option is enabled.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -50,6 +50,14 @@ fn commit_info() {
     if !Path::new(".git").exists() {
         return;
     }
+
+    // Var set by bootstrap whenever omit-git-hash is enabled in rust-lang/rust's config.toml.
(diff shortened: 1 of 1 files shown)


pietroalbini:
Opened https://github.com/rust-lang/rust/pull/117868 to set that environment variable in the first place. There is no dependency between the two PRs, they can be landed in any order. They will just do nothing until both of them are merged in `rust-lang/rust`.

pietroalbini:
CI failure seems spurious?

epage:
CI failure should be addressed by #12966

weihanglo:
@bors r+

Thanks!

PR #12969, opened 2023-11-13
Fix book page specifying-dependencies.md

ajoklar:

The Cargo Book currently suggests a wrong syntax for specifying target OS specific dependencies, i.e.:

`[target.'cfg(unix)'.dependencies]`) 

That leads to many 'use of undeclared crate or module' errors, because it doesn't recognize this as a syntax error (could be an issue of it's own).

I replaced these occurrences in the book with the correct syntax (`[target.'cfg(target_os = "unix")'.dependencies]`).

### DIFF
--- a/src/doc/src/reference/specifying-dependencies.md
+++ b/src/doc/src/reference/specifying-dependencies.md
@@ -315,10 +315,10 @@ syntax](../../reference/conditional-compilation.html) will be used to define
 these sections:
 
 ```toml
-[target.'cfg(windows)'.dependencies]
+[target.'cfg(target_os = "windows")'.dependencies]
 winhttp = "0.4.0"
 
-[target.'cfg(unix)'.dependencies]
+[target.'cfg(target_os = "unix")'.dependencies]
 openssl = "1.0.1"
 
 [target.'cfg(target_arch = "x86")'.dependencies]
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/specifying-dependencies.md
```diff
@@ -315,10 +315,10 @@ syntax](../../reference/conditional-compilation.html) will be used to define
 these sections:
 
 ```toml
-[target.'cfg(windows)'.dependencies]
+[target.'cfg(target_os = "windows")'.dependencies]
```
epage, line 318: Hmm, rustc seems to support `cfg(windows)` and `cfg(unix)`

See https://doc.rust-lang.org/reference/conditional-compilation.html#unix-and-windows

This likely needs a closer look

ehuss:
Thanks for the PR!  The syntax seems correct to me, and works on my system. Can you open an issue with a complete reproduction, including the `Cargo.toml` files and the commands you ran?

ajoklar:
Hi, that's interesting.

I looked at it again and realized that `unix` and `windows` seem to be special cases that both work in that position. In my project I used `android` and `macos` which both fail without a `target_os = ""`. I never actually tried `unix` or `windows` because I didn't recognize them as something different.

When I add `#[cfg(android)]` in a Rust file, I get this clear error:
> operating system used in target family position
>
> did you mean `unix`?

So there is a differentiation between "target family" and operating system that was not apparent to me.

So my PR is wrong, I'm sorry!

It would be nice to get the same error above when trying to use i.e.`[target.'cfg(android)'.dependencies]` in Cargo.toml

ehuss:
Unfortunately cargo does not have access to the information to know the names of the operating systems, or what the set of valid cfg key names are, so I'm not sure it is feasible to provide a warning for that.

PR #12970, opened 2023-11-13
query{_vec} use IndexSummary

Eh2406:
This builds on the work from https://github.com/rust-lang/cargo/pull/12749 and the discussion of the overall approach is at https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/better.20error.20messages.20for.20filtered.20versions.2E

Changing the two traits `Registry` and `Source` to use the new `IndexSummary' involves a lot of changes all throughout the code base. This would be hard to review if it also included any logic changes. So this PR is just adding the type to the trait and immediately unwrapping every place it is used.

The hope is that reviewing changes to logic/ergonomics will be easier to review once the mechanical changes have been merged.

This is an internal re-factoring and all the tests still pass.


epage:
@bors r+

I guess this is like my work on #12801.  Too big to do in a single PR but unsure of the results.  So only one way to find out.

crates/xtask-bump-check/src/xtask.rs
```diff
@@ -377,6 +377,7 @@ fn check_crates_io<'a>(
                 "`{name}@{current}` needs a bump because its should have a version newer than crates.io: {:?}`",
                 possibilities
                     .iter()
+                    .map(|s| s.as_summary())
```
epage, line 380: Thinking on this more, should we have done these as `as_candidate` and `into_candidate`.  Currently, we should only be exposing candidates and this allows us to update the registry to expose more things without updating every caller.

epage: For now, aren't they filtered out?  I assumed part of the change with this is you can make more or fewer things candidates.

Eh2406: That may end up being a better plan. I will keep it in mined.

PR #12971, opened 2023-11-13
refactor(toml): Clean up workspace inheritance

epage:

The goal is to simplify the code so we have a better boundary between `toml/mod.rs` and `toml/schema.rs` for when we break `toml/schema.rs` into a separate package for #12801.

This let us
- Remove a trait used in some back and forth for error handling
- Move a lot of the inheritable bookkeeping and logic out of `schema.rs`

A lot of these changes were inspired by [`cargo_toml`](https://docs.rs/cargo_toml/latest/cargo_toml/).  This included some renaming that I felt made the code clearer.

I didn't go as far as `cargo_toml`, yet.
- They derive more `Deserializers`, producing worse errors
- Their equivalent of `InheritableField` has an `inherit` function on it.  They eagerly inherit things and then error if anything isn't inherited
  - I'm still toying with something like this but held off for now
  - One idea is `InheritableField` has an `inherit_with` function, like today, that only passes errors up but doesn't generate an error.  We then have a `get` function that errors if it isn't inherited.  We could encode the field names, for error reporting, into a type parameter for `InheritableField`
- They flatten `InheritableDependency` into `TomlDependency`
  - By splitting these up, `workspace.dependencies` and `.cargo/config.toml` code can directly use `TomlDependency` without extra error handling
  -If we went this rout, I think I'd merge`InheritableDependency::Inherit` with `DetailedDependency`, having callers handle the errors (much like `TomlManifest` is both a real and virtual)

Some things I'm trying to balance
- Complexity
- Quality of error messages
- Knowing what code needs touching when changes are made
  - Some more work is needed here

### DIFF
--- a/src/cargo/util/toml/schema.rs
+++ b/src/cargo/util/toml/schema.rs
@@ -83,7 +83,7 @@ pub struct TomlWorkspace {
     pub metadata: Option<toml::Value>,
 
     // Properties that can be inherited by members.
-    pub package: Option<InheritableFields>,
+    pub package: Option<InheritablePackage>,
     pub dependencies: Option<BTreeMap<String, TomlDependency>>,
     pub lints: Option<TomlLints>,
 }
@@ -91,14 +91,7 @@ pub struct TomlWorkspace {
 /// A group of fields that are inheritable by members of the workspace
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
 #[serde(rename_all = "kebab-case")]
-pub struct InheritableFields {
-    // We use skip here since it will never be present when deserializing
-    // and we don't want it present when serializing
-    #[serde(skip)]
-    pub dependencies: Option<BTreeMap<String, TomlDependency>>,
-    #[serde(skip)]
-    pub lints: Option<TomlLints>,
-
+pub struct InheritablePackage {
     pub version: Option<semver::Version>,
     pub authors: Option<Vec<String>>,
     pub description: Option<String>,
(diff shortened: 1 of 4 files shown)


src/cargo/util/toml/schema.rs
```diff
@@ -83,22 +83,15 @@ pub struct TomlWorkspace {
     pub metadata: Option<toml::Value>,
 
     // Properties that can be inherited by members.
```
weihanglo, line 85: Is this comment still meaningful in the context? People using `TomlWorkspace` might not need to worry about workspace inheritance.

weihanglo:
This seems pretty neat!

I'd like to hand this to @Muscraft to see if they have any feedback.

r? Mustcraft

weihanglo:
r? Muscraft

src/cargo/util/toml/schema.rs
```diff
@@ -83,22 +83,15 @@ pub struct TomlWorkspace {
...
 #[serde(rename_all = "kebab-case")]
-pub struct InheritableFields {
-    // We use skip here since it will never be present when deserializing
-    // and we don't want it present when serializing
-    #[serde(skip)]
-    pub dependencies: Option<BTreeMap<String, TomlDependency>>,
-    #[serde(skip)]
-    pub lints: Option<TomlLints>,
-
+pub struct PackageTemplate {
```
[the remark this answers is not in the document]
epage: Thanks, fixing it

src/cargo/util/toml/schema.rs
epage: I'm not too sure what the concern is.  From my perspective, the comment still seems relevant (these are `workspace.package`, `workspace.dependencies`, and `workspace.lints` tables).  For any of `TomlWorkspace`, the caller decides whether they care or not.

Muscraft (approved):
Thanks!

This makes it easier to understand and reason about what is going on!

PR #12973, opened 2023-11-14
fix error message for duplicate links

dimbleby:
Fixing an error message which is both misformatted and contains the wrong text.

Previously it said that only one crate in the tree should use `links ='linking-crate'`; now it says that only one crate should use `links = 'linked-library'`

### DIFF
--- a/src/cargo/core/resolver/errors.rs
+++ b/src/cargo/core/resolver/errors.rs
@@ -140,9 +140,9 @@ pub(super) fn activation_error(
                     msg.push_str("` as well:\n");
                     msg.push_str(&describe_path_in_context(cx, p));
                     msg.push_str("\nOnly one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. ");
-                    msg.push_str("Try to adjust your dependencies so that only one package uses the links ='");
-                    msg.push_str(&*dep.package_name());
-                    msg.push_str("' value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.");
+                    msg.push_str("Try to adjust your dependencies so that only one package uses the `links = \"");
+                    msg.push_str(link);
+                    msg.push_str("\"` value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.");
                 }
                 ConflictReason::MissingFeatures(features) => {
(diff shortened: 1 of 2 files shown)


src/cargo/core/resolver/errors.rs
```diff
@@ -140,8 +140,8 @@ pub(super) fn activation_error(
                     msg.push_str("` as well:\n");
                     msg.push_str(&describe_path_in_context(cx, p));
                     msg.push_str("\nOnly one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. ");
-                    msg.push_str("Try to adjust your dependencies so that only one package uses the links ='");
-                    msg.push_str(&*dep.package_name());
+                    msg.push_str("Try to adjust your dependencies so that only one package uses the links = '");
+                    msg.push_str(link);
```
epage, line 144: I would err on the side of removing the first space rather than adding the second.

With the formatting we have today,  having spaces puts more of a burden on the reader to tell when they are dealing with words vs syntax.  Adding backticks could help but only so much.

Looking at `util/toml/mod.rs`, it seems we do a mix of space+backticks and no-space+no-backticks.  A lot of those messages are a lot less dense then this one though.

dimbleby: but that's not how anyone actually formats their `Cargo.toml`, right?  we're all writing `links = 'whatever'`?

weihanglo: BTW, should we refactor these lines into `format!` macro? It'll marginally increase the readability.

epage: Correct but people need to be able to parse this message.

We can either
- Add backticks
- Remove spaces

I lean towards the latter but we should at least do one of these two.

dimbleby: actually we're probably all writing `links = "whatever"` with the double quote.

I'll do that and put backticks around this bit, that would be consistent with the rest of the message.  Sound ok?

epage: Looking at the messages in practice, the backticks at least help with direct reading and probably ok enough for scanning.

epage:
Thanks!

@bors r+

PR #12975, opened 2023-11-14
fix(resolver): Don't do git fetches when updating workspace members

### LINKED ISSUE #8821: `cargo update --package X` updates other git repositories
**Problem**

`cargo update -p` should only update the specify package. Instead, it updates git repositories as well.

**Steps**

1. `git clone https://github.com/rust-lang/crater/`
2. `cd crater`
3. `git checkout dcc54566d5ec82cabca435c0145cbd06f32d3602`
4. `cargo update -p crater`

```
$ cargo update -p crater
    Updating crates.io index
    Updating git repository `https://github.com/rust-lang/team`
    Updating rust_team_data v1.0.0 (https://github.com/rust-lang/team#8d8902d7) -> #a3c474be
```

Output of `cargo version`: cargo 1.47.0 (f3c7e066a 2020-08-28)

### LINKED ISSUE #12599: Cargo update -p <localcrate> in a workspace has inconsistent behaviors

epage:

Before, when running `cargo update <member>`, we'd not reuse the
previous resolve result and when the resolver started walking into the
dependencies, it would do a git fetch.

Now, we won't even try to resolve the workspace members and so we won't
look at those dependencies and do git fetch.

This will make `cargo update <workspace-member>`
match `cargo update --workspace`.

Fixes #12599
Fixes #8821

I considered whether there were other ways of handling this but I
figured aiming for consistency in approaches was the best way.
We can investigate improving those approaches separately.

There are other discrepancies in the different code paths (handling of
patches, adding sources) but I'm deferring looking over those.

I added a test to demonstrate the `--workspace` behavior to use as a base line to compare with.

### Additional information

Between this and #12602, this should finally resolve #12599.

### DIFF
--- a/src/cargo/ops/cargo_generate_lockfile.rs
+++ b/src/cargo/ops/cargo_generate_lockfile.rs
@@ -122,6 +122,26 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
             }
         }
 
+        // Mirror `--workspace` and never avoid workspace members.
+        // Filtering them out here so the above processes them normally
+        // so their dependencies can be updated as requested
+        to_avoid = to_avoid
+            .into_iter()
+            .filter(|id| {
+                for package in ws.members() {
+                    let member_id = package.package_id();
+                    // Skip checking the `version` because `previous_resolve` might have a stale
+                    // value.
+                    // When dealing with workspace members, the other fields should be a
+                    // sufficiently unique match.
+                    if id.name() == member_id.name() && id.source_id() == member_id.source_id() {
+                        return false;
+                    }
+                }
+                true
+            })
+            .collect();
+
         registry.add_sources(sources)?;
     }
 
--- a/tests/testsuite/update.rs
+++ b/tests/testsuite/update.rs
@@ -1190,3 +1190,91 @@ rustdns.workspace = true
         ))
         .run();
 }
+
+#[cargo_test]
+fn update_only_members_with_workspace() {
+    let git_project = git::new("rustdns", |project| {
+        project
+            .file("Cargo.toml", &basic_lib_manifest("rustdns"))
+            .file("src/lib.rs", "pub fn bar() {}")
+    });
+
+    let workspace_toml = format!(
+        r#"
+[workspace.package]
+version = "2.29.8"
+edition = "2021"
+publish = false
+
+[workspace]
+members = [
+    "crate2",
+    "crate1",
+]
+resolver = "2"
+
+[workspace.dependencies]
+# Internal crates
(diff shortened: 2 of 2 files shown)


src/cargo/ops/cargo_generate_lockfile.rs
```diff
@@ -122,6 +122,26 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
...
+        // Filtering them out here so the above processes them normally
+        // so their dependencies can be updated as requested
+        to_avoid = to_avoid
+            .into_iter()
+            .filter(|id| {
+                for package in ws.members() {
+                    let member_id = package.package_id();
+                    // Skip checking the `version` because `previous_resolve` might have a stale
+                    // value.
+                    // When dealing with workspace members, the other fields should be a
```
weihanglo, line 135: This fix looks fairly easy and straightforward, though I have a scenario in mind. If a member depends on its own old version on crates.io, but now seems it cannot update to the latest version?

epage: We should be protected against this with the source check. In fact, the source check should be sufficient, without a name check, because it points to the directory for the package.

weihanglo: Fair. I missed the source check part

weihanglo (approved):

weihanglo:
Thanks for the fix!

@bors r+

PR #12976, opened 2023-11-14
docs(ref): Find a place to comment on --cap-lints

### LINKED ISSUE #5998: Improve documentation on cap-lints
I'm looking at some docs for the `--cap-lints` and it's not clear what the expected behavior is.

In <https://doc.rust-lang.org/rustc/lints/levels.html>, we have:

> This feature is used heavily by Cargo; it will pass `--cap-lints` allow when compiling your dependencies, so that if they have any warnings, they do not pollute the output of your build.

In <https://github.com/rust-lang/rfcs/pull/1193> we have:

> Cargo will then pass `--cap-lints` allow to all upstream dependencies when compiling code.

And in <https://github.com/rust-lang/cargo/pull/1830> we have:

> This commit adds support to Cargo to pass `--cap-lints` allow to all upstream
> dependencies instead of `-A warings`. This should serve the same purpose of
> suppressing warnings in upstream dependencies as well as preventing widespread
> breakage whenever a change to a lint is introduced in the compiler.

And [the RFC](https://github.com/alexcrichton/rfcs/blob/cap-lints/text/0000-cap-lints.md) has similar wordings, with an emphasis at the end:

> Modifications to existing lints to emit new warnings will not get triggered, and new lints will also be entirely suppressed **only for upstream dependencies**.

----

Reading all the docs above, I'm not sure what "upstream dependencies" mean exactly. From the `rustc` doc, it seems that it's just any dependent crate.

But my guess is that Cargo doesn't want to disable such warnings when working in a *workspace*. So, probably "upstream" means any dependency out of current manifest's workspace.

But, setting up a simple test tells me neither of the above is right, and I'm guessing again that "upstream" would mean anything not fetched from the local machine. (Which is a bit more confusing, because I would expect the same behavior if using a remote git repo or using a local checkout of that the same repo--both being outside of my current workspace.)

----

This is the library:

`my-lint-error/src/lib.rs`
```rust
#[deny(missing_copy_implementations)]

pub struct Foo {
    pub field: i32
}
```

and this is the caller:

`my-caller-app/src/main.rs`
```rust
extern crate my_lint_error;

fn main() {
    println!("Hello, world!");
}
```

and the caller's manifest has:
```toml
[dependencies]
my-lint-error = { path="../my-lint-error/" }
```

The build fails with the lint error coming from the dependency:

```rust
$ cargo build --manifest-path my-caller-app/Cargo.toml -v
   Compiling my-lint-error v0.1.0 (file:///.../my-lint-error)
     Running `rustc --crate-name my_lint_error /.../my-lint-error/src/lib.rs --crate-type lib --emit=dep-info,link -C debuginfo=2 -C metadata=a5f75dbabcaae9e9 -C extra-filename=-a5f75dbabcaae9e9 --out-dir /.../my-caller-app/target/debug/deps -C incremental=/.../my-caller-app/target/debug/incremental -L dependency=/.../my-caller-app/target/debug/deps`
error: type could implement `Copy`; consider adding `impl Copy`
 --> /.../my-lint-error/src/lib.rs:3:1
  |
3 | / pub struct Foo {
4 | |     pub field: i32
5 | | }
  | |_^
  |
note: lint level defined here
 --> /.../my-lint-error/src/lib.rs:1:8
  |
1 | #[deny(missing_copy_implementations)]
  |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: Could not compile `my-lint-error`.

Caused by:
  process didn't exit successfully: `rustc --crate-name my_lint_error /.../my-lint-error/src/lib.rs --crate-type lib --emit=dep-info,link -C debuginfo=2 -C metadata=a5f75dbabcaae9e9 -C extra-filename=-a5f75dbabcaae9e9 --out-dir /.../my-caller-app/target/debug/deps -C incremental=/.../my-caller-app/target/debug/incremental -L dependency=/.../my-caller-app/target/debug/deps` (exit code: 1)
```

----

The test shows how a `deny` error from a library is breaking `cargo build` in a dependent crate. Looking at the `rustc` command, there's no `--cap-lints` passed in. Running the same command with `--cap-lints allow` builds the dependent without any issues. Therefore, I'm guessing there's a bug here in Cargo.

What do you think? Did I miss something in the docs?

PS. I'm on today's `nightly` and my `~/.cargo/config` only has a bunch of aliases, so I believe my personal configuration is not causing this.

epage:
Looking over the points of confusion highlighted in #5998, one is in rustc which is agnostic of any of this policy and the rest are in historical documents.

Inspired by previous comments, I figured we could fit this into a discussion of `[lints]` by talking about the scope of the feature.

Fixes #5998

### DIFF
--- a/src/doc/src/reference/manifest.md
+++ b/src/doc/src/reference/manifest.md
@@ -576,6 +576,11 @@ unsafe_code = "forbid"
 enum_glob_use = "deny"
 ```
 
+Generally, these will only affect local development of the current package.
+Cargo only applies these to the current package and not to dependencies.
+As for dependents, Cargo suppresses lints from non-path dependencies with features like
+[`--cap-lints`](../../rustc/lints/levels.html#capping-lints).
+
 ## The `[badges]` section
 
 The `[badges]` section is for specifying status badges that can be displayed


ehuss:
Thanks!

@bors r+

Issue #12941, opened 2023-11-09
panic on cyclical dev-dependency with missing feature


ehuss:
### Problem

A cyclical dev-dependency with a missing feature causes a panic:

```
thread 'main' panicked at src/cargo/util/graph.rs:149:20:
the only path was a cycle, no dependency graph has this shape
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::option::expect_failed
   3: <cargo::util::graph::Graph<cargo::core::package_id::PackageId, im_rc::hash::set::HashSet<cargo::core::dependency::Dependency>>>::path_to::<<cargo::util::graph::Graph<cargo::core::package_id::PackageId, im_rc::hash::set::HashSet<cargo::core::dependency::Dependency>>>::path_to_bottom::{closure#0}, core::iter::adapters::flatten::FlatMap<core::option::IntoIter<&im_rc::ord::map::OrdMap<cargo::core::package_id::PackageId, im_rc::hash::set::HashSet<cargo::core::dependency::Dependency>>>, im_rc::ord::map::Iter<cargo::core::package_id::PackageId, im_rc::hash::set::HashSet<cargo::core::dependency::Dependency>>, <cargo::util::graph::Graph<cargo::core::package_id::PackageId, im_rc::hash::set::HashSet<cargo::core::dependency::Dependency>>>::edges::{closure#0}>>
   4: cargo::core::resolver::errors::describe_path_in_context
   5: cargo::core::resolver::errors::activation_error
   6: cargo::core::resolver::activate_deps_loop
   7: cargo::core::resolver::resolve
   8: cargo::ops::resolve::resolve_with_previous
   9: cargo::ops::cargo_generate_lockfile::generate_lockfile
  10: cargo::commands::generate_lockfile::exec
  11: cargo::cli::main
  12: cargo::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

### Steps

A test using cargo's test infrastructure:

```rust
#[cargo_test]
fn cyclical_dep_with_missing_feature() {
    // Checks for error handling when a cyclical dev-dependency specify a
    // feature that doesn't exist.
    let p = project()
        .file(
            "Cargo.toml",
            r#"
                [package]
                name = "foo"
                version = "0.1.0"

                [dev-dependencies]
                foo = { path = ".", features = ["missing"] }
            "#,
        )
        .file("src/lib.rs", "")
        .build();
[...]

PR #12977, opened 2023-11-14
If the only path is a loop then counted as the shortest path.

Eh2406:
This is a fix for #12941

This graph data structure is used to store dependency DAGs. Where each edge represents a dependency from a package to the package that fulfilled the dependency. Different parts of the resolver store this data in opposite directions, sometimes packages point at the things that depend on them other times packages point to the parents that required them. Error messages often need to report on why a package is in the graph, either by walking up toward parents or down toward children depending on how this graph is stored. #12678 unified the two different walking implementations, and replace them with a breadth first search so as to find the shortest path. This code ignored when edge pointed at a package that had already been reached, because that generally describes a longer path to an existing package.

Unfortunately, when I said this was a DAG that was a simplification. There can be cycles introduced as dev-dependencies. The existing code would reasonably ignore the cycles figuring that if we continue searching we would eventually find the root package (a package that nothing depended on). Missing the possibility that the root package created the cycle.

Now we search through the entire graph looking for a root package. If we do not find a root package we report the path to the last package we processed.

### DIFF
--- a/src/cargo/util/graph.rs
+++ b/src/cargo/util/graph.rs
@@ -128,30 +128,32 @@ impl<'s, N: Eq + Ord + Clone + 's, E: Default + Clone + 's> Graph<N, E> {
     {
         let mut back_link = BTreeMap::new();
         let mut queue = VecDeque::from([pkg]);
-        let mut bottom = None;
+        let mut last = pkg;
 
         while let Some(p) = queue.pop_front() {
-            bottom = Some(p);
+            last = p;
+            let mut out_edges = true;
             for (child, edge) in fn_edge(&self, p) {
-                bottom = None;
+                out_edges = false;
                 back_link.entry(child).or_insert_with(|| {
                     queue.push_back(child);
                     (p, edge)
                 });
             }
-            if bottom.is_some() {
+            if out_edges {
                 break;
             }
         }
 
         let mut result = Vec::new();
-        let mut next =
-            bottom.expect("the only path was a cycle, no dependency graph has this shape");
+        let mut next = last;
         while let Some((p, e)) = back_link.remove(&next) {
             result.push((next, Some(e)));
             next = p;
         }
-        result.push((next, None));
+        if result.iter().all(|(n, _)| n != &next) {
+            result.push((next, None));
+        }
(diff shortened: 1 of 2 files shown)


src/cargo/util/graph.rs
```diff
@@ -165,8 +165,12 @@ impl<'s, N: Eq + Ord + Clone + 's, E: Default + Clone + 's> Graph<N, E> {
                 ));
             }
             let last = result.last().unwrap().0;
-            // fixme: this may sometimes be wrong when there are cycles.
```
epage, line 168: Oh the fun we leave for ourselves :)

src/cargo/util/graph.rs
```diff
@@ -165,8 +165,12 @@ impl<'s, N: Eq + Ord + Clone + 's, E: Default + Clone + 's> Graph<N, E> {
...
-            if !fn_edge(&self, last).next().is_none() {
+            let set: Vec<_> = result.iter().map(|(k, _)| k).collect();
+            if !fn_edge(&self, last)
```
epage, line 171: Were these changes meant to be in the commit for adding a test?

weihanglo:
Thank you. It is pretty straightforward, but could we have a bit more description so that won't lose track of the patch in the future?

Eh2406:
I can do either... Are you asking me to better comment the code or improve the PR description?

src/cargo/util/graph.rs
Eh2406: It seems more like correcting the test than fixing the implementation, despite which function it was living in. If you would like me to move it and I need to re-base, I will fix it.

weihanglo:
PR description or commit message, which will show in git log so that people don't need to click the issue link and open a browser. It's pretty minor and not a blocker :)

Eh2406:
Edited the PR description. let me know if that's better, or is there somewhere else you want it.

weihanglo (approved):
Thanks!

src/cargo/util/graph.rs
epage: Ah, I missed the debug assertion part.  ick.

PR #12981, opened 2023-11-15
Switch from AtomicU64 to Mutex.

ehuss:
Not all platforms support AtomicU64, so this swaps the usage with Mutex. The difference in performance should be imperceptible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -416,7 +416,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-util"
-version = "0.2.8"
+version = "0.2.9"
 dependencies = [
  "anyhow",
  "core-foundation",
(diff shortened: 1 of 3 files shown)


epage (approved):
r= me when its ready

weihanglo:
Going to merge this now as it blocks submodule update in rust-lang/rust

@bors r+

Eh2406:
I think there's a user adjustable clippy lint to ban individual types, do we want to ban the reintroduction of AtomicU##?

ehuss:
@bors retry

Looks like these jobs are updating `Cargo.lock` to the latest versions. There appears to have been a bad release of `primeorder` which have been yanked, so I think it should pass.

(I don't know why those jobs are updating `Cargo.lock`, though.)

ehuss:
> I think there's a user adjustable clippy lint to ban individual types, do we want to ban the reintroduction of AtomicU##?

If someone wants to add that, it's probably ok. However, I don't know how that would interact if it is made conditional on `target_has_atomic`, in case someone wanted to have different variants for different platforms. I think 32 and smaller is available on all our platforms, it's just 64 and bigger.

PR #12982, opened 2023-11-15
Fix some test output validation.

ehuss:
A bunch of these tests weren't testing what they claimed to. For some reason, a bunch of them were checking for cargo's output with `.with_stdout("")`, but since cargo almost never prints to stdout, that doesn't actually validate the test. Most of these are checking that cargo doesn't recompile when run a second time, and in those cases it should be checking that there is a single line of output ("Finished"). I don't know the history here, many of these tests were before my time, perhaps cargo did use to print something to stdout? Or perhaps it was a long run of bad copy-and-paste.

There were a few other tests that were checking `with_stdout("")` for no apparent reason. I switched those to check the stderr output.


epage (approved):
r= me when its ready

PR #12985, opened 2023-11-16
cargo-credential-1password: Add missing `--account` argument to `op signin` command

Turbo87:

Without this the account chooser is shown by the `op signin` command, even though the user has already specified an account via the `--account` command line argument to the `cargo-credential-1password` CLI.

Note that the `--vault` in this case does not need to be forwarded to `op`, since it is irrelevant for the `op signin` command.

- Have a 1password installation with multiple accounts
- Use `global-credential-providers = ["cargo-credential-1password --account my.1password.com"]` in the cargo config file
- Run e.g. `cargo publish`
- Notice how you are seeing an account switcher even though `--account` was used
- Apply this patch and notice that the account switcher is no longer there and the correct account is selected automatically

### Additional information

see https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/1password.20credentials.20provider

### DIFF
--- a/credential/cargo-credential-1password/src/main.rs
+++ b/credential/cargo-credential-1password/src/main.rs
@@ -79,6 +79,10 @@ impl OnePasswordKeychain {
         }
         let mut cmd = Command::new("op");
         cmd.args(["signin", "--raw"]);
+        if let Some(account) = &self.account {
+            cmd.arg("--account");
+            cmd.arg(account);
+        }
         cmd.stdout(Stdio::piped());
         let mut child = cmd
             .spawn()


weihanglo:
@Muscraft already got an 1password account so they might be more helpful than me :)

r? Muscraft

ehuss:
I can take this.

r? ehuss

Muscraft:
I don't have multiple accounts, so I couldn't verify it would go to the account chooser before this change, but I was able to verify I could sign in after the change.

Note:  `op signin` states this is a way to pick a specific account.
```If you add multiple 1Password accounts to the 1Password app, 1Password CLI
determines which to use in this order:

1. The account specified with the '--account' flag.
````

ehuss:
Thanks!

@bors r+

Muscraft:
The failure was caused by a crates.io deployment. It's being rolled back, [see this Zulip comment](https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/deployments/near/402536867).

PR #12986, opened 2023-11-16
cargo-credential-1password: Fix README

Turbo87:

The README.md file of the cargo-credential-1password package does not match the implemented reality (anymore?).

- `--sign-in-address` and `--email` are not actually supported by the implementation
- manually calling `op signin` appears to be unnecessary

This PR also adds configuration instructions directly to the README.md so that they show up on crates.io.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -335,7 +335,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-credential-1password"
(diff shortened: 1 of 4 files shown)


ehuss:
Thanks, I went ahead and bumped the version to fix the check and get the docs updated.

@bors r+

PR #12987, opened 2023-11-16
Fix a rustflags test using a wrong buildfile name

MoSal:
The test seem to have always had the wrong buildfile `build.rs.rs`.


epage:
Thanks!

@bors r+

PR #12988, opened 2023-11-16
chore(ci): Catch naive use of AtomicU64 early

epage:
AtomicU64 is not portable and use of it, without `target_has_atomic`, will fail CI for rust-lang/rust when the cargo submodule gets updated (see #12981).

If there is a reason to use it, we could always `allow` in that one case.


weihanglo (approved):

weihanglo:
@bors r+ 

Thanks!

PR #12991, opened 2023-11-16
refactor: log when loading config from file

weihanglo:
When I was tracing some nasty bug and wanted to point at config probding, I wish there were some logs standing out and telling me I was wrong.

Would be great if every config file loading is logged.

### DIFF
--- a/src/cargo/util/config/mod.rs
+++ b/src/cargo/util/config/mod.rs
@@ -1200,6 +1200,8 @@ impl Config {
                 path.display()
             );
         }
+        tracing::debug!(?path, ?why_load, includes, "load config from file");
+
         let contents = fs::read_to_string(path)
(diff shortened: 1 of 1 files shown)


src/cargo/util/config/mod.rs
```diff
@@ -1200,6 +1200,8 @@ impl Config {
                 path.display()
             );
         }
+        tracing::event!(tracing::Level::DEBUG, ?path, ?why_load, includes, "load config from file");
```
epage, line 1203: nit: can't we do `tracing::debug!(` and skip referencing the constant?

epage (approved):
Feel free to r= me whichever way you go

src/cargo/util/config/mod.rs
weihanglo: Nice catch!

weihanglo:
@bors r+

Thanks for the review!

PR #12992, opened 2023-11-16
feat(cli): add color output for `cargo --list`

weihanglo:

This is a work based on #12578.

Add color output for `cargo --list`

image

with `--verbose`:

image

This shouldn't impact existent programmatic parsing of the output of `cargo --list`. For example, 

```
cargo --list 2>/dev/null | awk 'NR>1 {print $1}'
```

Each case should be covered

- [x] cargo-fmt and cargo-clippy
- [x] built-in commands
- [x] external commands
- [x] aliases

### DIFF
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -134,9 +134,14 @@ Run with 'cargo -Z [FLAG] [COMMAND]'",
                 "Formats all bin and lib files of the current crate using rustfmt.",
             ),
         ]);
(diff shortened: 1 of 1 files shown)


ehuss:
Sounds good to add some color, but I'm going to pass the review since I have not quite caught up with the ansi work.

r? epage

epage:
@bors r+

Thanks!

epage:
Sorry, forgot to double check that we were writing to an `anstream` stream

@bors r+

PR #12993, opened 2023-11-17
docs: Recommend a wider selection of libsecret-compatible password managers

cyqsimon:

Previously the only password manager recommended was Gnome Keyring, which not everyone will find ideal for a variety of reasons. A few common ones:
- GTK. The user may not want to have to install the entire toolkit.
- Look and feel. The design of Gnome software is pretty opinionated.
- They already use something else.

So I think it makes sense to recommend a few more widely-used alternatives. All three are GPL.

Documentation changes only.


weihanglo (approved):
Sounds reasonable. Thank you :)

ehuss:
Thanks! I'm not too familiar with libsecret (other than writing the helper), so I can't really confirm that it works with all of these, but I'll assume this is correct, and it matches my vague recollection of how it works.

@bors r+

Issue #11383, opened 2022-11-16
`cargo package` doesn't include build script if this has a custom path


cdecompilador:
### Problem

One package can have an entry on the manifest indicating a custom path for the build script, for example:

```toml
[package]
build = "../folder/build.rs"
```

### Steps

Create a multiple packages that share a single build script and for programmability reasons keep just a single build script

### Possible Solution(s)

The expected behaviour in this case would be to ship that build script on the root, and a naive solution would be to modify the behavior of the `build` attribute to be optional, so in case that the path is not found, default to the root.

The perfect solution would be to modify that entry also on the package, but I don't know if modification of the manifest is on the philosophy of `cargo package`

I put the example of `../folder/build.rs` because is the worst-case scenario in which the naive solution is clearly naive, since that dir might exist with another context.

### Version

```text
cargo 1.65.0 (4bc8f24d3 2022-10-20)
release: 1.65.0
commit-hash: 4bc8f24d3e899462e43621aab981f6383a370365
commit-date: 2022-10-20
host: x86_64-pc-windows-msvc
libgit2: 1.5.0 (sys:0.15.0 vendored)
libcurl: 7.83.1-DEV (sys:0.4.55+curl-7.83.1 vendored ssl:Schannel)
os: Windows 10.0.19044 (Windows 10 Enterprise LTSC 2021) [64-bit]
```

weihanglo:
Oh this doesn't look nice. Thank you for reporting this!

If you have no time to wait for a fix but still want to reuse the shared build script, a workaround would be: Inside your package, create a symlink to the shared file. Then modify the value `package.build` to the symlink. `cargo package` generally follows symlinks and copies file over, so build script should be included automatically.

Besides, I believe `cargo package` without `--no-verify` will fail as Cargo cannot find build script.

weihanglo:
### Instructions to fix this

I noticed that [`package.readme`] and [`package.license-file`] fall into the same trap. Thankfully, they've got addressed, and the solution is pretty close to what @cdecompilador suggested: modifying the manifest file during packaging.

So, if anyone wants to help with this, here are some rough instructions in my mind:

1. When Cargo creates a list of archive files, it takes extra processes on `readme` and `license-file` [here]. We could do the same thing on `build`.
2. Consider refactor [this piece of code][here] into [`check_for_file_and_add`] as there will be 3 fields under a similar process.
[...]

PR #12995, opened 2023-11-17
Exited with hard error when custom build file no existence or not in package 

linyihai:
Fixed https://github.com/rust-lang/cargo/issues/11383

Add test `build_script_outside_pkg_root`,  this will check `custom_build.rs` existence and whether in the package root, if not then exited with a hard error

## Additional information
The code just handle the `custom build` target that i know how to test it.  Other target type is skipped.

### DIFF
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -439,7 +439,7 @@ fn warn_custom_build_file_not_in_package(
         "the source file of {:?} target `{}` doesn't appear to be a path inside of the package.\n\
         It is at {}, whereas the root the package is {}.\n\
         This may cause issue during packaging, as modules resolution and resources included via macros are often relative to the path of source files.\n\
-        Please update the `build` setting in the manifest at `{}` and point to a path inside the root of the package.\n",
+        Please update the `build` setting in the manifest at `{}` and point to a path inside the root of the package.",
         target.kind(), target.name(), path.display(), pkg.root().display(),  pkg.manifest_path().display()
     );
 
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -7,6 +7,7 @@ use std::sync::Arc;
 use std::task::Poll;
 
 use crate::core::compiler::{BuildConfig, CompileMode, DefaultExecutor, Executor};
+use crate::core::manifest::Target;
 use crate::core::resolver::CliFeatures;
 use crate::core::{registry::PackageRegistry, resolver::HasDevUnits};
 use crate::core::{Feature, Shell, Verbosity, Workspace};
@@ -331,6 +332,29 @@ fn build_ar_list(
             warn_on_nonexistent_file(&pkg, &readme_path, "readme", &ws)?;
         }
     }
+
+    for t in pkg
+        .manifest()
+        .targets()
+        .iter()
+        .filter(|t| t.is_custom_build())
+    {
+        if let Some(custome_build_path) = t.src_path().path() {
+            let abs_custome_build_path =
+                paths::normalize_path(&pkg.root().join(custome_build_path));
+            if abs_custome_build_path.is_file() {
+                if !abs_custome_build_path
+                    .ancestors()
+                    .any(|ancestor| ancestor == pkg.root())
+                {
+                    warn_custom_build_file_not_in_package(pkg, &abs_custome_build_path, t, &ws)?
+                }
+            } else {
+                warn_on_nonexistent_file(&pkg, &custome_build_path, "build", &ws)?
+            }
+        }
+    }
+
     result.sort_unstable_by(|a, b| a.rel_path.cmp(&b.rel_path));
 
     Ok(result)
@@ -403,6 +427,23 @@ fn warn_on_nonexistent_file(
         rel_msg,
         pkg.manifest_path().display()
     ))
+}
+
+fn warn_custom_build_file_not_in_package(
+    pkg: &Package,
+    path: &Path,
+    target: &Target,
+    ws: &Workspace<'_>,
+) -> CargoResult<()> {
+    let msg = format!(
+        "the source file of {:?} target `{}` doesn't appear to be a path inside of the package.\n\
+        It is at {}, whereas the root the package is {}.\n\
+        This may cause issue during packaging, as modules resolution and resources included via macros are often relative to the path of source files.\n\
+        Please update the `build` setting in the manifest at `{}` and point to a path inside the root of the package.\n",
+        target.kind(), target.name(), path.display(), pkg.root().display(),  pkg.manifest_path().display()
+    );
+
+    ws.config().shell().warn(&msg)
 }
 
 /// Construct `Cargo.lock` for the package to be published.
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -342,15 +342,9 @@ fn build_ar_list(
         if let Some(custome_build_path) = t.src_path().path() {
             let abs_custome_build_path =
                 paths::normalize_path(&pkg.root().join(custome_build_path));
-            if abs_custome_build_path.is_file() {
-                if !abs_custome_build_path
-                    .ancestors()
-                    .any(|ancestor| ancestor == pkg.root())
-                {
-                    warn_custom_build_file_not_in_package(pkg, &abs_custome_build_path, t, &ws)?
-                }
-            } else {
-                warn_on_nonexistent_file(&pkg, &custome_build_path, "build", &ws)?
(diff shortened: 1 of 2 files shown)


src/cargo/ops/cargo_package.rs
```diff
@@ -331,6 +332,29 @@ fn build_ar_list(
             warn_on_nonexistent_file(&pkg, &readme_path, "readme", &ws)?;
         }
     }
+
+    for t in pkg
+        .manifest()
+        .targets()
+        .iter()
+        .filter(|t| t.is_custom_build())
```
linyihai, line 340: should i handle other target type and how can i test on them?

weihanglo: I would say yes,, as not every `target.path` are required to be packaged. We don't want to hassle everybody with their excluded examples.

But maybe leave that for follow-ups. We can always iterate on it later on :)

src/cargo/ops/cargo_package.rs
```diff
@@ -331,6 +332,29 @@ fn build_ar_list(
...
+    {
+        if let Some(custome_build_path) = t.src_path().path() {
+            let abs_custome_build_path =
+                paths::normalize_path(&pkg.root().join(custome_build_path));
+            if abs_custome_build_path.is_file() {
+                if !abs_custome_build_path
+                    .ancestors()
+                    .any(|ancestor| ancestor == pkg.root())
```
weihanglo, line 348: Guess we can use `Path::starts_with` instead?

src/cargo/ops/cargo_package.rs
```diff
@@ -405,6 +429,23 @@ fn warn_on_nonexistent_file(
     ))
 }
 
+fn warn_custom_build_file_not_in_package(
+    pkg: &Package,
+    path: &Path,
+    target: &Target,
+    ws: &Workspace<'_>,
+) -> CargoResult<()> {
+    let msg = format!(
```
weihanglo, line 438: I would lean toward a hard error. Without build script being packaged this package cannnot even compile.

tests/testsuite/package.rs
```diff
@@ -3304,3 +3304,57 @@ fn init_and_add_inner_target(p: ProjectBuilder) -> ProjectBuilder {
     .file("derp/target/foo.txt", "")
     .file("derp/not_target/foo.txt", "")
 }
+
+#[cargo_test]
+fn custom_build_warning() {
```
weihanglo, line 3309: nitpick: this test name could be better to tell what it tests against, something like `build_script_outside_pkg_root`?

tests/testsuite/package.rs
```diff
@@ -3304,3 +3304,57 @@ fn init_and_add_inner_target(p: ProjectBuilder) -> ProjectBuilder {
...
+            "\
+warning: manifest has no documentation, homepage or repository.
+See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for more info.
+warning: build `{}/../t_custom_build/custom_build.rs` does not appear to exist.
+Please update the build setting in the manifest at `{}/Cargo.toml`
+This may become a hard error in the future.
+",
+            p.root().display(),
+            p.root().display()
+        ))
```
weihanglo, line 3336: You can use the handy `[CWD]` placeholder here:

```suggestion
        .with_stderr(
            "\
[WARNING] manifest has no documentation, homepage or repository.
See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for more info.
[WARNING] build `[CWD]/../t_custom_build/custom_build.rs` does not appear to exist.
Please update the build setting in the manifest at `[CWD]/Cargo.toml`
This may become a hard error in the future.
",
        )
```

More placeholders can be found in https://github.com/rust-lang/cargo/blob/43fbbc9b36614360eeddefa2df165ecad9ba0e50/crates/cargo-test-support/src/compare.rs#L189-L241

src/cargo/ops/cargo_package.rs
```diff
@@ -405,6 +429,23 @@ fn warn_on_nonexistent_file(
...
+        "the source file of {:?} target `{}` doesn't appear to be a path inside of the package.\n\
+        It is at {}, whereas the root the package is {}.\n\
```
weihanglo, line 440: for cosistency, could we add backticks here
```suggestion
        It is at `{}`, whereas the root the package is `{}`.\n\
```

weihanglo (commented):
Thanks for the PR. I am not sure if this is ready for review but I took a look.

src/cargo/ops/cargo_package.rs
linyihai: Thanks, this time i will focus on the `custom build target`.

tests/testsuite/package.rs
linyihai: nice suggestion.

src/cargo/ops/cargo_package.rs
linyihai: Ok. I will used `shell().error()` instead.

src/cargo/ops/cargo_package.rs
```diff
@@ -405,6 +423,35 @@ fn warn_on_nonexistent_file(
...
+            format!("the source file of {:?} target `{}` doesn't appear to be a path inside of the package.\n\
+            It is at `{}`, whereas the root the package is `{}`.\n",
+            target.kind(), target.name(), path.display(), pkg.root().display()
+            )
+        } else {
+            format!(
+                "the source file of {:?} target `{}` doesn't appear to exist.\n",
+                target.kind(),
+                target.name()
+            )
```
weihanglo, line 443: The error message shows "the source file of \"custom-build\" target `build-script-custom_build`", which does look correct to me. Maybe this?

```suggestion
            format!(
                "the source file of doesn't appear to exist.\n",
                target.description_named(),
            )
```

BTW I would suggest extract `target.description_named()` as a variable so that each branch can just use identifier in the format args position.

src/cargo/ops/cargo_package.rs
weihanglo: By hard error I meant Cargo should stop building with an exit code other than 0. I don't think `shell().error()` doing so. Returning an `Result::Err` from `anyhow` should do that automatically.

tests/testsuite/package.rs
```diff
@@ -3304,3 +3304,47 @@ fn init_and_add_inner_target(p: ProjectBuilder) -> ProjectBuilder {
...
+        "\
+warning: manifest has no documentation, homepage or repository.
+See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for more info.
+error: the source file of \"custom-build\" target `build-script-custom_build` doesn't appear to be a path inside of the package.
+It is at `{}/t_custom_build/custom_build.rs`, whereas the root the package is `[CWD]`.
+This may cause issue during packaging, as modules resolution and resources included via macros are often relative to the path of source files.
+Please update the `build` setting in the manifest at `[CWD]/Cargo.toml` and point to a path inside the root of the package.
+", p.root().parent().unwrap().display());
+    p.cargo("package -l").with_stderr(&expect_msg).run();
+    _ = fs::remove_dir_all(&custom_build_root).unwrap();
```
weihanglo, line 3349: I should have called this out earlier.

We can access the root of this test case via `cargo_test_suppor::paths::root()`. So in line 3336 we can have a path like like `path::roots().join("t_custom_build")`, and this folder doesn't need to cleanup as it should be sufficiently indenpendant.

```suggestion
```

src/cargo/ops/cargo_package.rs
linyihai: Oh, i misunderstood. :joy:

tests/testsuite/package.rs
linyihai: Good handy tool functions.

src/cargo/ops/cargo_package.rs
```diff
@@ -405,6 +423,31 @@ fn warn_on_nonexistent_file(
...
 
+fn error_custom_build_file_not_in_package(
+    pkg: &Package,
+    path: &Path,
+    target: &Target,
+) -> CargoResult<Vec<ArchiveFile>> {
+    let tip = {
+        let description_name = target.description_named();
+        if path.is_file() {
+            format!("the source file of `{description_name}` doesn't appear to be a path inside of the package.\n\
```
weihanglo, line 434: Just one tiny suggestion and we're there!

It seems odd to have backticks here. Could we remove those?

```suggestion
            format!("the source file of {description_name} doesn't appear to be a path inside of the package.\n\
```

linyihai: It seems good to me to remove the backticks.

weihanglo (approved):

weihanglo:
Thank you. Sometimes I would force push and clean up the commit history if it help reasoning the patch. But that's fine I'll go ahead and merge it.

@bors r+

PR #12158, opened 2023-05-19
Add CARGO_WORKSPACE_DIR env var for integration tests and benchmarks

yerke:
### What does this PR try to resolve?

This is step towards #3946

Add CARGO_WORKSPACE_DIR env var for integration tests and benchmarks

### How should we test and review this PR?

Added tests for projects with and without workspace.

### Additional information

Big thank you to @epage for helping with this PR.

PR #12996, opened 2023-11-17
feat: Add `CARGO_RUSTC_CURRENT_DIR` (unstable)

epage:

This is an alternative to #12158's `CARGO_WORKSPACE_DIR` that was
implementing the solution to #3946 that previously discussed in the
cargo team meeting.

`CARGO_WORKSPACE_DIR` is a bit awkward to document / describe because
its the effective workspace directory of the thing being built.
If the thing being built doesn't have a workspace, it falls back to
`CARGO_MANIFEST_DIR`.

It would also be hard to take into account what the
`CARGO_WORKSPACE_DIR` would be for path dependencies into foreign
workspaces *and* it wouldn't solve the problem the user is having.

What the user really wants is the CWD of rustc when it is invoked.
This is much simpler to describe and is accurate when using a path
dependency to a foreign package.

Because the CWD is a much simpler mechanism to talk about, I figured we
could diverge from our prior consensus and make it always present,
rather than limiting it to tests.

The preparatory refactor commits have explanation for why they were to help

### Additional information

Remaining work for #3946: get this stabilized

### DIFF
--- a/src/doc/src/reference/environment-variables.md
+++ b/src/doc/src/reference/environment-variables.md
@@ -265,6 +265,7 @@ corresponding environment variable is set to the empty string, `""`.
   where integration tests or benchmarks are free to put any data needed by
   the tests/benches. Cargo initially creates this directory but doesn't
   manage its content in any way, this is the responsibility of the test code.
+* `CARGO_RUSTC_CURRENT_DIR` --- This is a path that `rustc` is invoked from **(nightly only)**.
 
 [Cargo target]: cargo-targets.md
 [binaries]: cargo-targets.md#binaries
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -662,6 +662,15 @@ fn prepare_rustc(cx: &Context<'_, '_>, unit: &Unit) -> CargoResult<ProcessBuilde
     let mut base = cx
         .compilation
         .rustc_process(unit, is_primary, is_workspace)?;
+    build_base_args(cx, &mut base, unit)?;
+
+    base.inherit_jobserver(&cx.jobserver);
+    build_deps_args(&mut base, cx, unit)?;
+    add_cap_lints(cx.bcx, unit, &mut base);
+    base.args(cx.bcx.rustflags_args(unit));
+    if cx.bcx.config.cli_unstable().binary_dep_depinfo {
+        base.arg("-Z").arg("binary-dep-depinfo");
+    }
 
     if is_primary {
         base.env("CARGO_PRIMARY_PACKAGE", "1");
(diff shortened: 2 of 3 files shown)


weihanglo (commented):
> What the user really wants is the CWD of rustc when it is invoked.
This is much simpler to describe and is accurate when using a path
dependency to a foreign package.

I don't quite agree. People might just want to get the workspace root to do some tricks like generating code for other members. Granted this might seem like a "misuse" of build script but that is already a thing.

The proposed `CARGO_RUSTC_CURRENT_DIR` sounds harder to find and educate as well.

epage:
> I don't quite agree. People might just want to get the workspace root to do some tricks like generating code for other members. Granted this might seem like a "misuse" of build script but that is already a thing.

A previous iteration of the cargo team rejected that use case.  When  we met to discuss the problem, we intentionally designed the solution to exclude that use case, only allowing it for benches and tests.  We also made it relative to communicate its wibbly wobbly nature.

> The proposed `CARGO_RUSTC_CURRENT_DIR` sounds harder to find and educate as well.

Note that I'm not talking about documenting a fully generic `CARGO_WORKSPACE_DIR` but specifically what we discussed to stabilize: https://github.com/rust-lang/cargo/issues/3946#issuecomment-1402464796
- Only for test and bin (does that include unit tests?  doc tests)
- If a member of the primary workspace, we use the workspace path.  For registry dependencies, it is the manifesr dir. 
  - But when it comes to path dependencies on foreign workspaces...
- Its a relative path

But really what we are trying to assist the user with is resolving `file!`.  They are assuming the workspace dir is what they need but what they actually care about is the `current_dir` for rustc.

src/doc/src/reference/environment-variables.md
```diff
@@ -265,6 +265,7 @@ corresponding environment variable is set to the empty string, `""`.
   where integration tests or benchmarks are free to put any data needed by
   the tests/benches. Cargo initially creates this directory but doesn't
   manage its content in any way, this is the responsibility of the test code.
+* `CARGO_RUSTC_CURRENT_DIR` --- This is a path that `rustc` is invoked from **(nightly only)**.
```
weihanglo, line 268: Should we mention one or two use cases here?

weihanglo (commented):
Thanks for the reply. Now I recalled the discussion. This is a good step forward. 

I don't think an FCP is needed, but the env name might worth bikdshedding? don't have a better idea though. `CARGO_RUSTC_INVOKED_DIR` sounds even more confusing.

epage:
> I don't think an FCP is needed, but the env name might worth bikdshedding? don't have a better idea though. CARGO_RUSTC_INVOKED_DIR sounds even more confusing.

I figure we'd do an FCP on stabilization and can worry about the name then.  I'd be curious to get this in a nightly for feedback from requesters.

src/doc/src/reference/environment-variables.md
epage: When we talked about this, Eric was hesitant about locking down the behavior of `file!` so I left out that use case.  It should work and I can't imagine rustc doing anything but absolute or relative to `std::env::current_dir` but I'm leaving that to people to make the choice that they want to make those assumptions, rather than us advertising it.

weihanglo:
Thank you. All make sense as this is an nightly-only feature.

@bors r+

PR #12997, opened 2023-11-18
cargo-test-support: Add features to the default Cargo.toml file

hi-rustin:

Found in https://github.com/hi-rustin/cargo-information/pull/32#issuecomment-1815625144.

I think we need to add features to the default `Cargo.toml` file.

- [x] Tested in https://github.com/hi-rustin/cargo-information/pull/32
image

### Additional information

Do you which Cargo command can cover this case? I want to add a unit test for it.
It seems I can use `cargo metadata` to cover it. I can add a dep to the package and try to check the output JSON.

### DIFF
--- a/crates/cargo-test-support/src/registry.rs
+++ b/crates/cargo-test-support/src/registry.rs
@@ -1523,6 +1523,30 @@ impl Package {
             manifest.push_str(&format!("rust-version = \"{}\"", version));
         }
 
+        if !self.features.is_empty() {
+            let features: Vec<String> = self
+                .features
+                .iter()
+                .map(|(feature, features)| {
+                    if features.is_empty() {
+                        format!("{} = []", feature)
+                    } else {
+                        format!(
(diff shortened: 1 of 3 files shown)


tests/testsuite/metadata.rs
```diff
@@ -86,6 +86,185 @@ fn cargo_metadata_simple() {
         .run();
 }
 
+#[cargo_test]
+fn cargo_metadata_with_registered_dep() {
```
hi-rustin, line 90: I am not sure if this test case is necessary. But we don't have a test for feature names. So I added it. Please let me know if we have a better way to cover it.

hi-rustin (commented):
🔢 Self-check

tests/testsuite/metadata.rs
hi-rustin: I dropped it. It is not a test for the real case, so it might not be worth adding it.

PR #13000, opened 2023-11-19
refactor(toml): Further clean up inheritance

epage:

This is a follow up to #12971 that was found as I continued working towards #12801.

The first is a more general purpose API cleanup.  I was bothered by the idea that a caller could create a `field.workspace = false` when that is disallowed, so I modified the API to prevent that.

The second is part of needing to find a home for everything in `toml/mod.rs`.  I figured `IneheritableField::as_value` is reasonable in the API, so I carried that forward.  It would be reasonable to add other methods, from an API perspective, but I left that for future exploration.

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -1568,13 +1568,6 @@ impl<T> schema::InheritableField<T> {
             }),
         }
     }
-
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
Both make sense. Thank you

PR #13005, opened 2023-11-19
Migrate rustfix to the cargo repo

ehuss:
This migrates the `rustfix` crate from https://github.com/rust-lang/rustfix/ to the cargo repo. The cargo team has been responsible for the client-side of `cargo fix`, and it can make it easier to maintain with all our tooling and tracking here. This crate is used by some external parties (like the compiler), so it will need to be maintained like an "ecosystem" package, but hopefully there shouldn't be any outside requirements (I haven't seen any in several years).

After merging, I'll follow up with some things to address in the future, such as:
- Migrating issues from the other repo.
- Opening new issues for some cleanup tasks, such as adding documentation, fixing the `#[ignore]` annotations, fixing testing on windows, maybe migrating the test code to use different dependencies, various code cleanup.
- Archiving the repo.


weihanglo (commented):
Besides the PR description, do we have any other reference or context for why this decision was made? Words from outside the Cargo team would be good.

ehuss:
> Do we have any other reference or context for why this decision was made?

Yea. The rustfix working group was retired some time ago (after being inactive for much longer) in https://github.com/rust-lang/team/pull/1048 and https://rust-lang.zulipchat.com/#narrow/stream/301329-t-devtools/topic/Retire.20wg-rustfix. 

Cargo is about the only primary user of this crate (with the compiler using it for testing), and that means the cargo team has been mostly responsible for it for the past few years. However, having it in a separate repo has meant that it has been a little more difficult to manage. 

Additionally, people are sometimes confused about what "rustfix" is, and assume the rustfix repo is where `cargo fix` is implemented, but it is not. Issues are being filed there when they really should be tracked in this repo (or https://github.com/rust-lang/rust). 

Also, due to the legacy of wg-rustfix, there is a perception among some that all "fixing of diagnostics" is the responsibility in the "rustfix" repo, but that is not the case. Implementing correct diagnostic replacements is now the responsibility of the compiler team, and the `cargo fix` implementation is the responsibility of the cargo team.

The Cargo team discussed this transition in the 2023-10-17 meeting and agreed to migrate it over.

ehuss:
BTW, for the version bump, I think a patch bump should be fine. I don't see any necessarily semver-breaking changes (though I think one could argue switching from log to tracing could be breaking, I'm not concerned about that). But please double-check my work.

These are the changes from the repo since the last publish: https://github.com/rust-lang/rustfix/compare/v0.6.1...HEAD
These are the changes in this PR: https://github.com/rust-lang/cargo/compare/3ca00ad8cdc5150634c434e2e01df57813d887a6...f99a494c31d6d28ea5ede539f7f5058e006e7b03

weihanglo:
mysterious error. Let @bors try again.

crates/rustfix/Readme.md
```diff
@@ -5,31 +5,12 @@
...
+If you are looking for the [`cargo fix`] implementation, the core of it is located in [`cargo::ops::fix`].
 
-- `rustfix`, a library for consuming and applying suggestions in the format that `rustc` outputs (this crate)
-- `cargo-fix`, a binary that works as cargo subcommand and that end users will use to fix their code (maintained in the [cargo](https://github.com/rust-lang/cargo/blob/master/src/bin/cargo/commands/fix.rs) repo).
-
-
-The library (and therefore this repo) is considered largely feature-complete. This is because:
-* There is no compiler or even rust-specific logic here
-* New lints and suggestions come from the Rust compiler (and external lints, like [clippy]).
-* `rustfix` doesn't touch the filesystem to implement fixes, or read from disk
```
weihanglo: [not a blocker]
This can be put somehere either in rustfix's lib.rs or somewhere in design architecture.

weihanglo (commented):
Not blockers just call them out.

* Do we want a `A-rustfix` label or just accommendate them in `Command-fix`?
* We would like to update `[autolabel]` in `triagebot.toml` as well.
* Update places mentioning `rusfix` in Cargo Contributor Guide
    * https://doc.crates.io/contrib/issues.html#filing-issues

weihanglo:
This command is failing:

```
cargo semver-checks -p rustfix --baseline-rev 1d8b05cdd1287c64467306cf3ca2c8ac60c11eb0
```

Seems like there is already a report in `cargo-semver-checks`: https://github.com/obi1kenobi/cargo-semver-checks/issues/481

The scenario is a newly added package missing in the baseline version.

weihanglo:
We might want to exclude `rustfix` from baseline check for now. Once rustfix-in-cargo hit stable (or beta if we want robustness), we can add it back.

```patch
diff --git a/crates/xtask-bump-check/src/xtask.rs b/crates/xtask-bump-check/src/xtask.rs
index cba420f051b..634ac596e82 100644
--- a/crates/xtask-bump-check/src/xtask.rs
+++ b/crates/xtask-bump-check/src/xtask.rs
@@ -166,6 +166,7 @@ fn bump_check(args: &clap::ArgMatches, config: &cargo::util::Config) -> CargoRes
         let mut cmd = ProcessBuilder::new("cargo");
         cmd.arg("semver-checks")
             .arg("--workspace")
+            .arg("--exclude rustfix")
             .arg("--baseline-rev")
             .arg(referenced_commit.id().to_string());
         config.shell().status("Running", &cmd)?;
```

crates/rustfix/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[dependencies]
+anyhow.workspace = true
+serde = { workspace = true, features = ["derive"] }
+serde_json.workspace = true
+tracing.workspace = true
+
+[dev-dependencies]
+proptest.workspace = true
+similar = "0.4.0"
```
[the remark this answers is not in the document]
ehuss: I have deduplicated it.

crates/rustfix/Readme.md
ehuss: I put the beginning of a doc-comment in the library. My intention is to defer doing documentation work after this merged. There is essentially no documentation, and it can be a bit much to review updating all of that independently.

ehuss:
Thanks for catching all the things I missed!

> Do we want a `A-rustfix` label or just accommendate them in `Command-fix`?

I think for now I would prefer to just keep one label (`Command-fix`). I think in practice it is unlikely we will want to try to distinguish between the two, and it would make it a little more awkward to find issues because you would need to include both in your query.

> We would like to update `[autolabel]` in `triagebot.toml` as well.

Added.

> Update places mentioning `rusfix` in Cargo Contributor Guide

Updated.

> We might want to exclude `rustfix` from baseline check for now. Once rustfix-in-cargo hit stable (or beta if we want robustness), we can add it back.

I have applied your patch.

weihanglo (approved):

weihanglo:
Thank you ehuss for the migration!

@bors r+

PR #13006, opened 2023-11-19
contrib docs: Update now that credential crates are published.

ehuss:
The credential crates are now auto-published, so this updates the documentation to reflect that.

### DIFF
--- a/src/doc/contrib/src/process/release.md
+++ b/src/doc/contrib/src/process/release.md
@@ -146,10 +146,9 @@ Cargo team, or the Release team) should publish it manually using `cargo
 publish`.
 
 Some packages are not published automatically because they are not part of the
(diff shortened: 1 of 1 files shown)


weihanglo (commented):
https://github.com/rust-lang/cargo/blob/d794e1ff268f55456c0c270abc688a6994850880/publish.py#L22

do we want to remove this line from `publish.py`?

ehuss:
> do we want to remove this line from `publish.py`?

Oops, I got that mixed up. I instead updated the docs to not mention it. We can remove it if/when it is removed from the repo.

weihanglo (approved):

weihanglo:
Thank you!

@bors r+

PR #13008, opened 2023-11-19
Add more resources to the contrib docs.

ehuss:
This adds some information about HackMD and Zulip used by the Cargo team.


weihanglo (approved):
Thanks!

Issue #10554, opened 2022-04-11
Tracking Issue for rustc `--check-cfg` integration


weihanglo:
### Summary

RFC: [#3013]
rustc tracking issue: rust-lang/rust#82450

Documentation: 
  - https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#check-cfg
  - https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#cargorustc-check-cfgcheck_cfg

This tracks the integration of rustc `--check-cfg` flag ([RFC #3013][#3013]), which enables complete or partial checking of conditional compilation configuration at compile time.

[#3013]: https://github.com/rust-lang/rfcs/pull/3013

### Unresolved Issues

- [ ] Should these flags be enabled by default? What's the unintrusive interface to provide for users to opt-out the checking? 
- [x] If we feel the diagnostic on using a feature when no features are defined is insufficient, then do we feel it would not be incongruous with the design to fix this in the future? (see #13011).

### Post-RFC decisions

- Support `cargo:rustc-check-cfg` despite `::` being introduced in 1.77 and this being released after that.  This was done to allow for wider MSRVs while using this feature

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

### Implementation history:

- https://github.com/rust-lang/cargo/pull/10408
- https://github.com/rust-lang/cargo/pull/10428
- https://github.com/rust-lang/cargo/pull/10486
- https://github.com/rust-lang/cargo/pull/10566
- https://github.com/rust-lang/cargo/pull/10539
- https://github.com/rust-lang/cargo/pull/12845
- https://github.com/rust-lang/cargo/pull/12884
- https://github.com/rust-lang/cargo/pull/13011
- https://github.com/rust-lang/cargo/pull/13012
- https://github.com/rust-lang/cargo/pull/13316

est31:
Over in https://github.com/rust-lang/rust/pull/93628 I've discovered two things:

* First, a papercut: apparently there are empty `values()` calls when the feature is enabled, even if there is another `values(...)` call with content present already. In that case, `values()` is not needed, right?
[...]

PR #13011, opened 2023-11-19
Fix `--check-cfg` invocations with zero features

Urgau:
When generating the `--check-cfg` arguments for `-Zcheck-cfg` we currently generate `cfg(feature, values())` when there is 0 features. This is wrong since a empty `values()` would mean that it's possible to have `#[cfg(feature)]` without a feature name which is impossible.

We replace this by a simple `cfg()` to still enable well known names and values.

----

Note that currently `rustc` defines `feature` as a well known name with ANY values if it's not overridden by Cargo. I plan on submitting a PR to `rustc` to remove `feature` from being a well known name so that Cargo is the only source of truth.

*This doesn't block this PR from being merged*

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1243,24 +1243,31 @@ fn trim_paths_args(
 fn check_cfg_args(cx: &Context<'_, '_>, unit: &Unit) -> Vec<OsString> {
     if cx.bcx.config.cli_unstable().check_cfg {
         // This generate something like this:
-        //  - cfg(feature, values())
+        //  - cfg()
         //  - cfg(feature, values("foo", "bar"))
         //
         // NOTE: Despite only explicitly specifying `feature`, well known names and values
         // are implicitly enabled when one or more `--check-cfg` argument is passed.
+        // NOTE: Never generate a empty `values()` since it would mean that it's possible
+        // to have `cfg(feature)` without a feature name which is impossible.
 
         let gross_cap_estimation = unit.pkg.summary().features().len() * 7 + 25;
         let mut arg_feature = OsString::with_capacity(gross_cap_estimation);
-        arg_feature.push("cfg(feature, values(");
-        for (i, feature) in unit.pkg.summary().features().keys().enumerate() {
-            if i != 0 {
-                arg_feature.push(", ");
+
+        arg_feature.push("cfg(");
+        if !unit.pkg.summary().features().is_empty() {
+            arg_feature.push("feature, values(");
+            for (i, feature) in unit.pkg.summary().features().keys().enumerate() {
+                if i != 0 {
+                    arg_feature.push(", ");
+                }
+                arg_feature.push("\"");
+                arg_feature.push(feature);
+                arg_feature.push("\"");
             }
-            arg_feature.push("\"");
-            arg_feature.push(feature);
-            arg_feature.push("\"");
+            arg_feature.push(")");
         }
-        arg_feature.push("))");
+        arg_feature.push(")");
 
         vec![
             OsString::from("-Zunstable-options"),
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
Sounds pretty reasonable to me. Thank you!

epage:
> When generating the `--check-cfg` arguments for `-Zcheck-cfg` we currently generate `cfg(feature, values())` when there is 0 features. This is wrong since a empty `values()` would mean that it's possible to have `#[cfg(feature)]` without a feature name which is impossible.

To me, this reads as a bug in the design of the syntax if `values()` is special cased like this while the semantics are different.

Urgau:
`values()` (without any specified values) isn't special cased. It means no values are expected.
`values("foo", "bar")` means that the values "foo" and "bar" are expected.

The issue here lies before that, `cfg(feature, values())` means: a config named "feature" exists and takes no values; but this isn't what we want, we want that no config named "feature" with whatever values exists, which is simply achieved by not defining "feature" in the first place.

epage:
> values() (without any specified values) isn't special cased. It means no values are expected.
values("foo", "bar") means that the values "foo" and "bar" are expected.

To put this in Rust types, does `values()` make it a unit type or an empty array?
```rust
struct EmptyValues;
const EMPTY_VALUES = [];
```

If its like a unit type, then that does feel like special casing.

> The issue here lies before that, cfg(feature, values()) means: a config named "feature" exists and takes no values; but this isn't what we want, we want that no config named "feature" with whatever values exists, which is simply achieved by not defining "feature" in the first place.

Is that what we want?

Whether there even is a way to define a cfg as an empty array, I suspect being able to do so for `features` would offer the best usability.

As a user, if I get an diagnostic saying that `features` doesn't exist, I assume I fundamentally didn't understand how this worked.  However, if I get a diagnostic saying that `"foo"` doesn't exist for `features`, then that is much clearer what the next steps are.

Urgau:
> To put this in Rust types, does `values()` make it a unit type or an empty array?
> 
> ```rust
> struct EmptyValues;
> const EMPTY_VALUES = [];
> ```
> 
> If its like a unit type, then that does feel like special casing.

It's more like `const EMPTY_VALUES = [];`.

*Internally it would be more like `const EMPTY_VALUES = [None];` where `None` represent the absence of value since we need to differentiate from `""` (empty string).*
*This follows what is done for `--cfg`:*
 - *`--cfg foo` => `("foo", None)`*
 - *`--cfg foo=""` => `("foo", Some(""))`*

> > The issue here lies before that, cfg(feature, values()) means: a config named "feature" exists and takes no values; but this isn't what we want, we want that no config named "feature" with whatever values exists, which is simply achieved by not defining "feature" in the first place.
> 
> Is that what we want?

I think so, from the perceptive of `--check-cfg` when config is not to be ever expected, since cargo will never pass `--cfg feature=...` to `rustc`, we should not declare `feature` as expected.

> Whether there even is a way to define a cfg as an empty array, I suspect being able to do so for `features` would offer the best usability.

Being able to say: declare a config as expected with all values as unexpected, is not something that is currently possible.

> As a user, if I get an diagnostic saying that `features` doesn't exist, I assume I fundamentally didn't understand how this worked. However, if I get a diagnostic saying that `"foo"` doesn't exist for `features`, then that is much clearer what the next steps are.

I agree, but `feature` is really unexpected here, making it awkward to say it exist but really  no it doesn't.

Saying that `feature` is unexpected is not wrong but is not 100% true either, same for saying that `feature` is expected, it is not 100% right but it's not 100M wrong either.

I don't know how to fundamentally improve the situation here.

In https://github.com/rust-lang/rust/pull/118071 I propose to add a diagnostic for this case: `= help: consider defining some features in Cargo.toml`, which should help, but it's not incredible.

epage:
I've noted this in #10554 for us to discuss during stabillization

PR #13004, opened 2023-11-19
[stable-1.74] chore: bump `cargo-credential-*` crates as e58b84d broke stuff

weihanglo:
### What does this PR try to resolve?

e58b84d3 changed the shape of response of cargo credential protocol trait,
so credential plugins crates effectively depend on `cargo-credential@0.4.0`.
However, `cargo@0.74.0` still depends on`cargo-credential@0.3.0`.
They must depends on the same major version of `cargo-credential`
otherwise incompatible.

Addresses <https://github.com/rust-lang/cargo/issues/13002>

In order to make CI pass, the following PRs are also cherry-picked:
- 3dcf93f20a0cfd2773499fe73afe0d793397ae96 from #12876
- c12c4c3f8d89f41534e51f9ee7641c7a410edf04 from #12921
- Bump `cargo-semver-checks` to 0.24.0 for supporting rustdoc json v27 (the version staring from Rust 1.74.0)

### How should we test and review this PR?

To-dos **after PRs against each branch get merged**:

* Under `rust-1.74.0` branch
    * Publish `cargo-credential-wincred@0.4.0`
    * Publish `cargo-credential-macos-keychain@0.4.0`
    * Publish `cargo-credential-li@0.4.0`
* Under `rust-1.75.0` branch
    * Bump the version to `cargo-credential-wincred@0.4.1`
    * Bump the version to `cargo-credential-macos-keychain@0.4.1`
    * Bump the version to `cargo-credential-li@0.4.1`
* Under `master` branch`
    * Bump the version to `cargo-credential-wincred@0.4.2`
    * Bump the version to `cargo-credential-macos-keychain@0.4.2`
    * Bump the version to `cargo-credential-li@0.4.2`
* Yank
    * Yank `cargo-credential-wincred@0.3.1`
    * Yank `cargo-credential-macos-keychain@0.3.1`
    * Yank `cargo-credential-libsecret@0.3.2`

Not sure if it is necessary to publish a new version of `0.3.x` crates for `cargo-credential-*` before yanking, so I didn't put it in this plan.

PR #13015, opened 2023-11-20
[stable 1.74] Bump cargo patch version

ehuss:
https://github.com/rust-lang/cargo/pull/13004 bumped the credential crate versions to deal with a semver-incompatible change, but I believe we also need to bump and publish cargo.

My thinking is that:
* cargo 0.75.0 currently depends on cargo-credential-libsecret="0.3.1"
    * The `Cargo.lock` is locked to **0.3.2**
* We plan to yank cargo-credential-libsecret 0.3.2
* Users who depend on cargo 0.75.0 as a library will be unable to do so since after yanking it will pick up cargo-credential-libsecret="0.3.1", but this won't work.
    * cargo 0.75.0 -> cargo-credential 0.4.0
    * cargo-credential-libsecret 0.3.1 -> cargo-credential **0.3.0**

(Users using a `Cargo.lock`, like doing rust builds, shouldn't be affected by the yank.)

So, after publishing the new credential crates, I think we'll also need to publish a new 0.75.1 which depends on the correct versions so that users can use 0.75. It also probably makes sense to yank 0.75.0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -234,7 +234,7 @@ dependencies = [
 
 [[package]]
 name = "cargo"
-version = "0.75.0"
+version = "0.75.1"
 dependencies = [
  "anstream",
  "anstyle",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -105,7 +105,7 @@ windows-sys = "0.48"
 
 [package]
 name = "cargo"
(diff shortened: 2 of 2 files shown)


weihanglo (approved):
Thanks!

I forgot this part 😅

PR #13016, opened 2023-11-20
Handle $message_type in JSON diagnostics

dtolnay:

Unblocks https://github.com/rust-lang/rust/pull/115691.

Without this change, Cargo's testsuite fails in `doc::doc_message_format` and `metabuild::metabuild_failed_build_json`.

Tested with and without https://github.com/rust-lang/rust/pull/115691.

In Cargo repo: `cargo test --test testsuite`
In Rust repo: `x.py test src/tools/cargo` (separately on master and $message_type PR)

### DIFF
--- a/crates/cargo-test-support/src/compare.rs
+++ b/crates/cargo-test-support/src/compare.rs
@@ -591,15 +591,36 @@ fn find_json_mismatch_r<'a>(
                 .next()
         }
         (&Object(ref l), &Object(ref r)) => {
-            let same_keys = l.len() == r.len() && l.keys().all(|k| r.contains_key(k));
-            if !same_keys {
-                return Some((expected, actual));
-            }
+            let mut expected_entries = l.iter();
+            let mut actual_entries = r.iter();
 
-            l.values()
-                .zip(r.values())
-                .filter_map(|(l, r)| find_json_mismatch_r(l, r, cwd))
-                .next()
+            // Compilers older than 1.76 do not produce $message_type.
+            // Treat it as optional for now.
+            let mut expected_entries_without_message_type;
+            let expected_entries: &mut dyn Iterator<Item = _> =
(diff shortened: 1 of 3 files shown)


ehuss:
Thanks!

@bors r+

crates/cargo-test-support/src/compare.rs
```diff
@@ -591,15 +591,36 @@ fn find_json_mismatch_r<'a>(
...
-            if !same_keys {
-                return Some((expected, actual));
+            let mut expected_entries = l.iter();
+            let mut actual_entries = r.iter();
+
+            // Compilers older than 1.76 do not produce $message_type.
+            // Treat it as optional for now.
+            let mut expected_entries_without_message_type;
+            let expected_entries: &mut dyn Iterator<Item = _> =
+                if l.contains_key("$message_type") && !r.contains_key("$message_type") {
```
epage, line 601: Specializing this for the compiler runs counter to a (slow moving) effort I'm working on to use a generic, third party library rather than us using bespoke testing tools.

ehuss:
@bors retry

issue with the ubuntu azure mirror.

crates/cargo-test-support/src/compare.rs
dtolnay: We can delete this code as soon as Rust 1.76 is stable (February 8) and Cargo's MSRV is raised. From https://github.com/rust-lang/cargo/pull/12775 it looks like MSRV update is immediate upon stabilization. I have filed #13017 to follow up.

If the generic third-party library adoption proceeds faster than that, you would need to figure out how to accommodate this situation in a principled way. This will not be the last time that rustc's JSON is tweaked.

epage: imo this should have been done principled before merging rather than punting that onto others.

PR #13021, opened 2023-11-22
refactor(toml): Better abstract inheritance details

epage:
There is no reason that the caller for inheriting needs to know the details of the inner types doing inheritance.


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #6980, opened 2019-05-23
Test the Resolver against the varisat Library

Eh2406:
Resolution can be reduced to the SAT problem. So this is an alternative implementation of the resolver that uses a SAT library for the hard work. This is intended to be easy to read, as compared to the real resolver, and run as part of the test sweet to make sure the real resolver works as expected. Part of #6120.

Some notes on performance:
The initial version did not support public & private deps:
~64 loc, `O(nln(n))` vars, `O(nln(n) + n*d)` clauses, 0.5x slower on `prop_passes_validation`
The final version:
~163 loc, `O(dn^2`) vars, `O(dn^3)`  clauses, 1.5x slower on `prop_passes_validation`

That comparison makes me feel better about spending months trying to get public & private deps to be fast enough for stabilization.

PR #13036, opened 2023-11-22
fix(resolver): Remove unused public-deps error handling

epage:
To implement rust-lang/rfcs#3516, we need to decouple the resolver's behavior from the unstable flag.  Since the code path is now dead, I went ahead and removed it.


weihanglo (commented):
std in `rust-lang/rust` makes use of public-private dependency. Does it affect building std there?

Eh2406:
This looks reasonable. There is more to clean up in `resolver-tests` that needs to be done. The trickiest part of that will be https://github.com/rust-lang/cargo/blob/master/crates/resolver-tests/src/lib.rs#L328C92-L328C92

There may also be follow-up work where some data structures no longer need to be `RC`/`im`, but feel free to leave that for later.

epage:
Noticed resolver-tests and am getting a failure in the aforementioned function,, so going to need to do some more digging.

epage:
> std in `rust-lang/rust` makes use of public-private dependency. Does it affect building std there?

This is used for validation with maybe some backtracking to make things work better.  This shouldn't break anyone immediately and at worst it will make people need to do some more careful `cargo update`s.

Eh2406:
I did a quick review, and it seems like there is still more to be excised. Specifically the code in
SatResolve https://github.com/rust-lang/cargo/blob/master/crates/resolver-tests/src/lib.rs#L328C92-L328C92 and registry_strategy in https://github.com/rust-lang/cargo/blob/master/crates/resolver-tests/src/lib.rs#L812

epage:
The test in question was added in #6980 with public dependency support added in 6454efdd67dfbc426b275a7a1cfc45dd6fc0f392

Eh2406:
Looking good so far! Next to remove https://github.com/rust-lang/cargo/blob/master/crates/resolver-tests/src/lib.rs#L812

Issue #13037, opened 2023-11-22
Add `--public` support for `cargo add`


epage:
### Problem

rust-lang/rfcs#3516  calls for controlling the new `public` field in `cargo add`

### Proposed Solution

Add `--public` and `--no-public` flags.  This will mirror `--optional` / `--no-optional`

Test cases
- Ability to add a new dep that is a public dependency
- Able to turn an existing private dep into a public dep
- Able to turn an existing public dep into a private dep

See the test cases for `--optional` for patterns in naming, etc.

### Notes

The RFC called for `--public <bool>`.  ... I don't remember why I did that.  I think it might have been due to some aspects of a previous design of the RFC and I forgot to update it.  We can note this in the tracking issue for us to approve this deviation.

epage:
@hi-rustin if you are ok handing this off, I think @linyihai was interested in contributing to public-private dependencies and I made this as a starting point for them.

linyihai:
Never mind, if this issue you are instersted in then i will pick another https://github.com/rust-lang/cargo/issues/13038, i will be ok for me :)

hi-rustin:
Oh, I didn't know that. Please feel free to take it.

hi-rustin:
> Oh, I didn't know that.

I found it on the zuilp :(

Please take it and give it a shot! Have a nice hacking time!

weihanglo:
Resolved by #13046. Thank people for your inputs!

PR #12189, opened 2023-05-26
 refactor(tests): Reduce cargo-add setup load

epage:
This just gets rid of irrelevant packages in the registry.  Looking into which versions aren't needed would require a deeper pass, so I held off on that for now.

Before, the tests were in the 300-500ms range and now they take 100-300ms.

This did call to my attention that `unrelated` is misspelled as `unrelateed` but holding off on fixing that to reduce conflicts.

PR #13046, opened 2023-11-25
Add `--public` for `cargo add`

linyihai:
Complete https://github.com/rust-lang/cargo/issues/13037

This PR want to add `--public/--no public` flag for `cargo add`

Note: this assumes we'll remove workspace inheritance support for `public` as it sounds like we'll be reverting it https://github.com/rust-lang/rust/issues/44663#issuecomment-1826474620.  If we decide to keep workspace inheritance, we'll need to come back and update this.

Most of Code were reference `cargo add --optional`, So can reviewed the new code based on the part of  `optional` code.

The new testcases were origin from the `cargo add --optional` part.
- `public` testcase：there is no dependencies and will be add `public` dependencies.
- `no_public` testcase:  there is no dependencies  and will be add `no_public` dependencies.
- `overwrite_public` testcase: the dependencies already exists but will be overwrite with `public`.
- `overwrite_no_public` testcase: the dependencies already exists but will be overwrite with `no_public`.
- `overwrite_public_with_no_public` testcase: the dependencies already marked as `no_public` and will be overwrite with `public`.
- `overwrite_no_public_with_public` testcase: the dependencies already marked as `public` and will be overwrite with `no_public`.


src/bin/cargo/commands/add.rs
```diff
@@ -76,6 +76,18 @@ Example uses:
                 "ignore-rust-version",
                 "Ignore `rust-version` specification in packages (unstable)"
             ),
+            flag("public", "Mark the dependency as public")
```
epage, line 79: Please move this up next to optional

src/bin/cargo/commands/add.rs
```diff
@@ -76,6 +76,18 @@ Example uses:
...
+                .long_help("Mark the dependency as public
+
+The dependnecy will be visible in both of your crate and outside."),
+            flag("no-public", "Mark the dependency as private")
+                .conflicts_with("dev")
```
epage, line 84: no-public has this twice

We need it on public and no-public as well as a conflict for `build`

tests/testsuite/cargo_add/no_public/mod.rs
```diff
@@ -0,0 +1,39 @@
+use cargo_test_support::compare::assert_ui;
+use cargo_test_support::prelude::*;
+use cargo_test_support::Project;
+
+use cargo_test_support::curr_dir;
+
+#[cargo_test]
+fn case() {
+    cargo_test_support::registry::init();
+    for name in ["my-package1", "my-package2"] {
```
epage, line 10: As these are new tests, lets make sure we only add the minimum dep/version combinations needed as these slow down test running and move us closer to running out of space in CI (a very rough pass was previously done to minimize these when the impact was realized but more cleanup is possible)

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -366,6 +377,7 @@ impl Dependency {
             crate_root.display()
         );
         let table: toml_edit::Item = match (
```
epage, line 379: nit: we should be consistent in the order we list public/optional

epage (commented):
This is just a quick pass. I know that I at Ieast need to look at tests closer.

tests/testsuite/cargo_add/overwrite_public_with_no_public/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+
+    let project = Project::from_template(curr_dir!().join("in"));
+    let project_root = project.root();
+    let cwd = &project_root;
+
+    snapbox::cmd::Command::cargo_ui()
+        .arg("add")
+        .arg_line("my-package1 my-package2@0.4.1 --no-public")
```
epage, line 30: I don't think these tests are getting anything by adding two packages.  How about we drop it to one?

Yeah, another area where we could clean up the optional tests.

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -758,6 +761,13 @@ fn populate_dependency(mut dependency: Dependency, arg: &DepOp) -> Dependency {
             dependency.optional = None;
         }
     }
+    if let Some(value) = arg.public {
+        if value {
+            dependency.public = Some(true);
+        } else {
+            dependency.public = None;
+        }
+    }
```
epage, line 770: As currently implemented, when you inherit a public dependency, you can make it non-public.  I'm double checking those semantics at https://github.com/rust-lang/rust/issues/44663#issuecomment-1826474620

If we keep those semantics, we'll want to allow setting `dependency.public = Some(false);` if the dependency is inherited and it is public.  If we support that, we should have a test case for it.

tests/testsuite/cargo_add/mod.rs
```diff
@@ -65,6 +65,7 @@ mod namever;
 mod no_args;
```
epage, line 65: > detect_workspace_inherit_optional

We likely should have a version of that for public

tests/testsuite/cargo_add/no_public/mod.rs
linyihai: See https://github.com/rust-lang/cargo/pull/13046/commits/07ba57dd8ba2a4c4f0e4aac8198a3fc69b48e07b , Only use one package one version for `cargo add`

tests/testsuite/cargo_add/overwrite_public_with_no_public/mod.rs
linyihai: same solution : https://github.com/rust-lang/cargo/pull/13046#discussion_r1405740955

tests/testsuite/cargo_add/mod.rs
linyihai: It's the same reason: https://github.com/rust-lang/cargo/pull/13046#discussion_r1405784748

tests/testsuite/cargo_add/overwrite_public_with_no_public/mod.rs
epage: > By the way, I can make another PR to remove the unnecessary packages and versions for the optional tests. just like https://github.com/rust-lang/cargo/commit/07ba57dd8ba2a4c4f0e4aac8198a3fc69b48e07b

Yes, this would be a great follow up to the previous clean up work in #12189.

In terms of multiple packages ,some tests will want it to verify behavior but we don't need it on every test.  Similar for the many versions.

src/bin/cargo/commands/add.rs
epage: This is still missing from `public`

tests/testsuite/cargo_add/no_public/mod.rs
```diff
@@ -7,27 +7,14 @@ use cargo_test_support::curr_dir;
...
-
+    cargo_test_support::registry::Package::new("my-package", "0.1.0").publish();
     let project = Project::from_template(curr_dir!().join("in"));
     let project_root = project.root();
     let cwd = &project_root;
 
     snapbox::cmd::Command::cargo_ui()
         .arg("add")
-        .arg_line("my-package1 my-package2@0.4.1 --no-public")
+        .arg_line("my-package@0.1.0 --no-public")
```
epage, line 17: Let's drop the `@<ver>` as well

src/bin/cargo/commands/add.rs
linyihai: Sorry by left out it twice.

tests/testsuite/cargo_add/no_public/in/Cargo.toml
```diff
@@ -1,6 +1,9 @@
 cargo-features = ["public-dependency"]
 [workspace]
 
+[lib]
+path = "../../add_basic.in/src/lib.rs"
```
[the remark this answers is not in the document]
linyihai: By specifing `path = "../../add_basic.in/src/lib.rs"`,  i suppose this could reduce a lot of same file.  
  
the inspiration was from `no_optional` and `optional`, they made a symbolic link to the `add-basic.in`
```
ll optional/in no_optional/in
lrwxrwxrwx 1 root root 15 Oct  9 15:52 no_optional/in -> ../add-basic.in/
lrwxrwxrwx 1 root root 15 Oct  9 15:52 optional/in -> ../add-basic.in/
```

If there is a need to keep each test case separate, I can create a separate lib.rs for each test case in next commit.

tests/testsuite/cargo_add/no_public/mod.rs
linyihai: Ok.

tests/testsuite/cargo_add/mod.rs
epage: This would be testing the many-valued `match` statement to make sure  we correctly decide whether to use short-hand (`dep.workspace = true`) or not (`dep = { workspace = true, public = true }`).

src/cargo/ops/cargo_add/mod.rs
epage: Let's go ahead and assume we won't be inheriting `public` but that assumption should be noted in the PR description and tracking issue to make sure we revisit this when a decision is made.

src/bin/cargo/commands/add.rs
```diff
@@ -62,6 +62,19 @@ The package name will be exposed as feature of your crate.")
 The package will be removed from your features.")
                 .conflicts_with("dev")
                 .overrides_with("optional"),
+            flag("public", "Mark the dependency as public")
+                .conflicts_with("dev")
+                .conflicts_with("build")
+                .long_help("Mark the dependency as public
+
+The dependnecy will be visible in both of your crate and outside."),
```
epage, line 70: > The dependency can be referenced in your library's public API

src/bin/cargo/commands/add.rs
```diff
@@ -62,6 +62,19 @@ The package name will be exposed as feature of your crate.")
...
+            flag("no-public", "Mark the dependency as private")
+                .conflicts_with("dev")
+                .conflicts_with("build")
+                .overrides_with("public")
+                .long_help("Mark the dependency as private
+
+The dependnecy will be only visible in your crate other than outside."),
```
epage, line 77: > While you can use the crate in your implementation, it cannot be referenced in your public API

tests/testsuite/cargo_add/mod.rs
linyihai: I had added the new case 48c998fec7ff84f3aec3547e6ba32ca40eae32cd firstly. 
If this addition is not correct, please let me know why, thank you.

epage:
Looks great, thanks!

@bors r+

Issue #13034, opened 2023-11-22
`rustfix`: Fix ignored tests


ehuss:
For some reason, several tests in rustfix are ignored:

* `edition` tests: https://github.com/rust-lang/cargo/blob/65d0eb536dd4a7ae6ddc50ca14556896e450c4ff/crates/rustfix/tests/parse_and_replace.rs#L242C1-L247. There don't appear to be any edition tests, so I think all of that should just be removed.
* These two proptests: https://github.com/rust-lang/cargo/blob/65d0eb536dd4a7ae6ddc50ca14556896e450c4ff/crates/rustfix/src/replace.rs#L313-L335.  I'm guessing that these are ignored because they were either slow, or had errors. However, in my testing they seem to run pretty fast, so I'm guessing we could just remove the `#[ignore]`. This needs a little investigation, just to make sure these tests aren't flaky or anything.

PR #13047, opened 2023-11-25
review and remove ignored tests in rustfix

tompscanlan:
review ignored tests in rustfix crate per #13034.

CI testing

### Additional information

* Removed unproductive test in `parse_and_replace`
* un-ignore proptests, and reduce runtime from ~2s to ~<.25s

### DIFF
--- a/crates/rustfix/src/replace.rs
+++ b/crates/rustfix/src/replace.rs
@@ -312,18 +312,16 @@ mod tests {
 
     proptest! {
         #[test]
-        #[ignore]
         fn new_to_vec_roundtrip(ref s in "\\PC*") {
             assert_eq!(s.as_bytes(), Data::new(s.as_bytes()).to_vec().as_slice());
         }
 
         #[test]
-        #[ignore]
         fn replace_random_chunks(
             ref data in "\\PC*",
             ref replacements in prop::collection::vec(
                 (any::<::std::ops::Range<usize>>(), any::<Vec<u8>>()),
-                1..1337,
+                1..100,
             )
         ) {
             let mut d = Data::new(data.as_bytes());
--- a/crates/rustfix/tests/parse_and_replace.rs
+++ b/crates/rustfix/tests/parse_and_replace.rs
@@ -14,7 +14,6 @@ use tracing::{debug, info, warn};
 
 mod fixmode {
     pub const EVERYTHING: &str = "yolo";
-    pub const EDITION: &str = "edition";
 }
 
 mod settings {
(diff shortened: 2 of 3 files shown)


tompscanlan:
I'm not familiar with proptests. Is running them here beneficial, or should we trim them? I lean towards trimming.

Why:
1. I don't think regressions will be captured back into git, so I think if we hit a bug, we'd fail to record it as is
2. functions tested are covered by other tests, so this is somewhat redundant

Thoughts from the team: Should I trim the proptests off?

crates/rustfix/tests/parse_and_replace.rs
```diff
@@ -238,10 +238,3 @@ fn everything() {
     tracing_subscriber::fmt::init();
     assert_fixtures("./tests/everything", fixmode::EVERYTHING);
 }
-
-#[test]
-#[ignore = "Requires custom rustc build"]
-fn edition() {
```
weihanglo, line 244: Guess you might want to remove all references to `fixmode::EDITION`, and the `./test/edition` directory as well.

https://github.com/rust-lang/cargo/blob/35ed69cb1b32fcb66c77032cc5480965fb519558/crates/rustfix/tests/parse_and_replace.rs#L16

crates/rustfix/src/replace.rs
```diff
@@ -312,18 +312,16 @@ mod tests {
...
         }
 
         #[test]
-        #[ignore]
         fn replace_random_chunks(
             ref data in "\\PC*",
             ref replacements in prop::collection::vec(
                 (any::<::std::ops::Range<usize>>(), any::<Vec<u8>>()),
-                1..1337,
+                1..100,
```
weihanglo, line 324: Agree that these can be un-ignored. See <https://github.com/rust-lang/rustfix/pull/67#discussion_r186228464>

Would there be an impact on test correctness for this change?
1337 took me 4s to run, and time spent seems marginal. I don't know how the nubmer is chosen. Maybe because of [this](https://en.wikipedia.org/wiki/Leet)?

tompscanlan: This change should not affect correctness, just the number of times we iterate on generated samples for this test.

This number is the number replacements that are generated for testing.  You got the reference to 133t, which is arbitrary.  I reduced to 100 to shorten time in test, but to still exercise it.  This reduces time for testing from 4s to <.25s.

crates/rustfix/tests/parse_and_replace.rs
tompscanlan: Doing so leaves a useless param to calls to compile.  Will unwind that next unless you see a reason to keep.

weihanglo (approved):
Thanks. Let move on and merge this :)

Issue #10160, opened 2021-12-04
Change usage of anyhow to more specific error enums


kornelski:
### Problem

Most of Cargo's internal helper functions return `anyhow::Error`. This makes it difficult for callers of such functions to implement custom, specific error handling other than by downcasting. Fallible downcasting of anyhow `Error` to specific error types seems fragile: the error types are not checked at compile time, so if a helper method changes its implementation, error handling in callers can break at runtime.

Places where a specific error type could be useful:
* handling of SSH auth failures. Customized error message exists, but it depends on a downcast.
* more readable and tailored output of build script error. Currently `ProcessError` is used for multiple purposes, and its generic verbose output has a high noise to signal ratio for build script failures. #10159
* better handling of `cargo new` in workspaces https://github.com/rust-lang/cargo/issues/6378#issuecomment-972879170

### Proposed Solution

I suggest gradually moving away from use of `anyhow::Error`, and use concrete error types (function or module-specific structs or enums). For a start this could be limited to internal APIs.

alexcrichton:
Personally I don't really know what your goal here is but you've come to a pretty large Rust project and by saying:

> I suggest gradually moving away from use of anyhow::Error

You've basically said "please redo your entire error handling strategy". That's a very large change to a project like Cargo and naturally isn't really one we'd just sort of off-hand agree with and start accepting PRs for. Cargo's error handling was designed long ago and hasn't really changed since. That's not to say that it's perfect but I don't think it's necessarily the most productive route to not ask for or try to understand historical context and simply propose complete 180 turnaround changes from Cargo's error handling strategy today.

Cargo's error-handling strategy is designed to not be a maintenance burden on us while also enabling more sophisticated handling of errors where necessary. It's not really clear to me why Cargo's error handling today cannot be used to solve the issues you've mentioned here. I think it would be much better to reuse what Cargo does today instead of trying to design something new out-of-the-blue and completely change Cargo's internal idioms because of a few issues.

kornelski:
Thanks for the response! It may sound like an ask for a rewrite, but I don't think it is that major. It's more of an ask for permission for to remove an unnecessary layer of abstraction in a few places where it gets in the way of better error handling.

[...]

PR #13050, opened 2023-11-26
refactor: use custom error instead of anyhow

weihanglo:

This defines custom error types for `rustfix` crate, instead of `anyhow`.

Some error messages were slightly tweaked but shouldn't affect the behavior.

I am not particularly sure about the SemVer compatibility. Some functions' return type has changed from `Result<_, anyhow::Error>` to `Result<_, rustfix::Error>`. People already consume `anyhow::Error` should be able to use `rustfix::Error` without updating their code I think.


hi-rustin:
Can I ask why do we need to switch to using custom errors? Do we need it for use in other repos to check the error type?

weihanglo:
> Can I ask why do we need to switch to using custom errors? Do we need it for use in other repos to check the error type?

If we are on the way to make cargo crates more user-friendly, a better error management could be a good route to go. Otherwise people might need downcasting things to get the root error cause. 

Some more contexts:

* https://github.com/rust-lang/cargo/issues/10160
* https://github.com/rust-lang/cargo/issues/12235
* https://github.com/rust-lang/cargo/pull/12424
* https://github.com/rust-lang/cargo/pull/12310

epage (approved):
> I am not particularly sure about the SemVer compatibility. Some functions' return type has changed from Result<_, anyhow::Error> to Result<_, rustfix::Error>. People already consume anyhow::Error should be able to use rustfix::Error without updating their code I think.

Only if they are always using `?` / `map_err(Into::into)` which might not always be the case
- Someone making a call to rustfix in a trailing statement of a function
- Someone capturing a function pointer

We should likely go ahead and treat this as a breaking change.

I'm unsure how you might want to proceed, so I'm only approving and not merging, whether
- you want to hold off for other breaking changes
- need to bump the version in the manifest
- something else

Issue #3364, opened 2016-12-03
`cargo uninstall` tries to delete executables that are running, which fails on Windows


FaultyRAM:
`cargo uninstall` doesn't take into account that the crate being uninstalled may in fact be running, and that attempting to delete such an executable will fail on Windows. Because removing the binary is the last step in the uninstall process, this leads to a phenonemon I dub "Schrödinger's crate", in which a crate is both installed and uninstalled; the executable still exists, so `cargo install` fails, but all installation metadata is gone, so `cargo uninstall` fails!

One way to cause this is via the Rusty Code plugin for VS Code; it depends on Racer, which can be installed via `cargo install` and is active while Rusty Code is active. Trying to `cargo uninstall` Racer in this situation leads to breakage as described above.

You can work around this by deleting the leftover executable after making sure it isn't running; `cargo install --force` also works.

Tested with the latest nightly Rust, installed via `rustup-rs`:

```
rustc 1.15.0-nightly (908dba0c9 2016-12-01)
binary: rustc
commit-hash: 908dba0c9477b7dd022a236cb1514ddfca9369f2
commit-date: 2016-12-01
host: x86_64-pc-windows-msvc
release: 1.15.0-nightly
LLVM version: 3.9
```

```
cargo 0.16.0-nightly (3568be9 2016-11-26)
```

alexcrichton:
Hm so `cargo install` in theory is transactional in the sense that if it fails Cargo's still not in a "corrupt" state. Have you found though that after a failed `cargo uninstall` it's essentially in a corrupt state?

FaultyRAM:
In this case no, at least as far as I can tell. The crate isn't listed in .crates.toml, and once the leftover executable is dealt with everything works as normal. It really does seem to be just that the executables get left behind.

alexcrichton:
Sounds bad! If `cargo uninstall` fails Cargo should always leave everything in a consistent state.

weihanglo:
If we focus on making the installed trakcer consistent, I believe this is not too hard to fix. A rough solution may look like

* For `paths::remove_file()` and `tracker.remove()` should go hand in hand.
* `tracker.save()` should be called at the end of `cargo uninstall`, so that at least removed files are still tracked on disk. 
* The behavior happens here: https://github.com/rust-lang/cargo/blob/20df9e40a4d41dd08478549915588395e55efb4c/src/cargo/ops/cargo_uninstall.rs#L139-L152

[...]

PR #13053, opened 2023-11-28
Fixed uninstall a running binary failed on Windows

linyihai:
 Fixes https://github.com/rust-lang/cargo/issues/3364

The problem reproduce when you try to uninstall a running binary and it will failed on Windows,  this is because that cargo had already remove the installed binary tracker information in disk,  and next to remove the running binary but it is not allowed in Windows.  And to the worst,  you can not uninstall the binary already and only to reinstall it by the `--force` flag.

### How to solve the issue?
This PR try to make the uninstall a binary more reasonable that  only after remove the binary sucessfully then remove the tracker information on binary and make the remove binary one by one.

Add testcase https://github.com/rust-lang/cargo/pull/13053/commits/0fd4fd357b440d4155661c3b770ed8ddee43d13e
- install the `foo`
- run `foo` in another thread.
- try to uninstall running `foo` and only failed in Windows.
- wait the `foo` finish, uninstall `foo`
- install the `foo`

### DIFF
--- a/src/cargo/ops/cargo_uninstall.rs
+++ b/src/cargo/ops/cargo_uninstall.rs
@@ -147,12 +147,14 @@ fn uninstall_pkgid(
         tracker.remove(pkgid, &bins);
     }
 
+    // This should have been handled last, but now restore it to reproduce the problem on the Windows platform.
+    // See issue #3364.
+    tracker.save()?;
+
     for bin in to_remove {
         config.shell().status("Removing", bin.display())?;
         paths::remove_file(bin)?;
     }
 
-    // Only Save the tracker when remove Bin successfully.
-    tracker.save()?;
     Ok(())
 }
--- a/src/cargo/ops/cargo_uninstall.rs
+++ b/src/cargo/ops/cargo_uninstall.rs
@@ -135,18 +135,18 @@ fn uninstall_pkgid(
         }
     }
 
-    let to_remove: Vec<&String> = {
+    let to_remove = {
         if bins.is_empty() {
-            installed.iter().collect()
+            installed
         } else {
-            bins.iter().collect()
+            bins
         }
     };
 
     for bin in to_remove {
-        let bin_path = dst.join(bin);
+        let bin_path = dst.join(&bin);
         config.shell().status("Removing", bin_path.display())?;
-        tracker.remove_bin_then_save(pkgid, bin, &bin_path)?;
+        tracker.remove_bin_then_save(pkgid, &bin, &bin_path)?;
     }
 
     Ok(())
--- a/src/cargo/ops/cargo_uninstall.rs
+++ b/src/cargo/ops/cargo_uninstall.rs
@@ -6,7 +6,6 @@ use crate::util::errors::CargoResult;
 use crate::util::Config;
 use crate::util::Filesystem;
 use anyhow::bail;
-use cargo_util::paths;
 use std::collections::BTreeSet;
 use std::env;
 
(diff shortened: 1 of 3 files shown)


src/cargo/ops/cargo_uninstall.rs
```diff
@@ -146,11 +146,13 @@ fn uninstall_pkgid(
...
     }
-    tracker.save()?;
+
     for bin in to_remove {
         config.shell().status("Removing", bin.display())?;
         paths::remove_file(bin)?;
     }
 
+    // Only Save the tracker when remove Bin successfully.
+    tracker.save()?;
```
epage, line 156: This misses the following part of the suggested fix

> For paths::remove_file() and tracker.remove() should go hand in hand.

Or in other words
- The old behavior was to stop tracking binaries, leaking them
- The new approach is to remove binaries but we still track them

The proposed behavior in the issue is to keep the filesystem and the tracker in sync.

linyihai: > The new approach is to remove binaries but we still track them

If I'm not mistaken: the `the approach` is means what [this commit](09be0eef54a1e66d38012f1c1e7fd3db424e9938) had done.

linyihai: > The proposed behavior in the issue is to keep the filesystem and the tracker in sync.
I try to solving the problem in next commit

tests/testsuite/install.rs
```diff
@@ -2507,3 +2507,81 @@ fn install_incompat_msrv() {
...
+        .run();
+    assert_has_installed_exe(cargo_home(), "foo");
+
+    use cargo_util::ProcessBuilder;
+    use std::thread;
+    let foo_bin = cargo_test_support::install::cargo_home()
+        .join("bin")
+        .join(cargo_test_support::install::exe("foo"));
+
+    let t = thread::spawn(|| ProcessBuilder::new(foo_bin).exec().unwrap());
```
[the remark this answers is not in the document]
linyihai: it seems `cargo_test_support::process` do not work in a thread,  i take your suggestion, replace the `ProcessBuilder::new` with `cargo_test_support::process`, then the complier complains: 

```
running `C:\workspace\cargo\target\debug\cargo.exe install foo`
thread '<unnamed>' panicked at crates\cargo-test-support\src\paths.rs:95:20:
Tests must use the `#[cargo_test]` attribute in order to be able to use the crate root.
running `C:\workspace\cargo\target\debug\cargo.exe uninstall foo`
```

tests/testsuite/install.rs
```diff
@@ -2507,3 +2508,112 @@ fn install_incompat_msrv() {
...
+    cargo_home().join(".crates2.json")
+}
+
+fn load_crates1() -> toml::Value {
+    toml::from_str(&fs::read_to_string(v1_path()).unwrap()).unwrap()
+}
+
+fn load_crates2() -> serde_json::Value {
+    serde_json::from_str(&fs::read_to_string(v2_path()).unwrap()).unwrap()
+}
```
weihanglo, line 2526: nit: Could we inline these functions in `assert_tracker_noexistence`? They doesn't seem to be shared elsewhere and scattering a single logic to multiple places make it a bit harder to read.

tests/testsuite/install.rs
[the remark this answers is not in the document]
weihanglo: Good catch! `global_root` is in use in `process` so you can do that.

linyihai:
> We're about to merge this. Let me know if you want to make more change or I'll merge it as-is.

let me make some change according you suggestion.

weihanglo (approved):
Thank you for the contribution!

Issue #5656, opened 2018-06-26
Add "CI best practicies" section to the guide


matklad:
Things to talk about:

  - what directories should be cached (probably `~/.cargo`, `./target`, but keep `~/.cargo/credentials` in mind)
  - Rust channels (testing on current stable, beta and the oldest supported rustc)
  - using `--minimal-versions` to verify dependencies in Cargo.toml
  - setting-up rustfmt/clippy

Ideally, the section should contain copy-pastable config files for various CI providers.

matklad:
A blog post about this feature: https://blog.illicitonion.com/2018/06/rust-minimum-versions-semver-is-lie.html

ehuss:
I assume this means to expand https://github.com/rust-lang/cargo/blob/master/src/doc/src/guide/continuous-integration.md?  Should it maybe leverage or at least point to [japaric/trust](https://github.com/japaric/trust)?

matklad:
Ah, I haven't realized that https://github.com/rust-lang/cargo/blob/master/src/doc/src/guide/continuous-integration.md exists. 

> Should it maybe leverage or at least point to japaric/trust?

Yep, we should definitely mention it!

Eh2406:
@ehuss That is a good place for it to go, thanks for the reminder! 

The impetus for this is that when we stabilize `--minimal-versions` we want to craft a message about how to use it. Specifically as a small part of a thorough CI setup. Witch will just lead everyone to ask, "What showed the big parts of the CI setup be?" So we thought we should write up our opinion before we stabilize `--minimal-versions`.

withoutboats:
Relevant: rust-lang/rfcs#2483

Eh2406:
That is an excellent read. Thank you for writing this up so articulately. I especially liked the paragraph of:
> In the long run, it could even make sense to combine the two approaches, allowing crates to state their toolchain requirements (and have that influence resolution), but encourage core crates to state “LTS” as their requirement.
 
That sounds like the tradeoff bending I know and love from the rust ethic. I.E. a careful and pedantic system to ensure you are correct but with a large pit of success to make it usable. It also involves the most design work, so it may not be feasible.

In our earlier discussion today you suggested that "equality constraints are rare in the ecosystem" and I resisted being pedantic and pointing out lock files. So I was glad to see them addressed in your blog post. But I do have one nit with:
> Similarly, when dependencies are subsequently adjusted, Cargo will “unlock” the affected dependencies and again choose the maximum version.

[...]

PR #13056, opened 2023-11-28
docs: Provide pointers for MSRV

epage:

In today's cargo team meeting, we discussed the Pre-RFC for MSRV-aware resolver for #9930.
In that discussion, the question of recommending a policy came up. While we didn't feel the ecosystem has coalesced enough to set one (and we hope MSRV-aware resolver will avoid the need),
it became clear that some we can provide some basic help to the user, including
- Raising awareness of tools to find the actual MSRV
- The policy that they should verify it with examples on how to do so

While this recommends some specific third-party tools, I'm not aware of other tools within this for us to worry about at this time for us to create any guidelines on which we should include.

Explanations are given for the example CI job to discourage cargo culting and instead give people the information they need in making decisions relevant to their project.

### Additional information

I'd love to provide information to help users create their own MSRV policy but only if there was an automated way of collecting and reporting some of the data, like crates.io providing a dashboard of MSRVs set or rust-versions inferred from user-agents.

Without that, I felt it not worth getting into other policy discussions like reactive vs proactive updating of MSRV, automated MSRV updates, etc.  These can always be added later.

This also helps towards #5656.

### DIFF
--- a/src/doc/src/reference/manifest.md
+++ b/src/doc/src/reference/manifest.md
@@ -185,6 +185,10 @@ The `rust-version` may be ignored using the `--ignore-rust-version` option.
 Setting the `rust-version` key in `[package]` will affect all targets/crates in
 the package, including test suites, benchmarks, binaries, examples, etc.
 
+To find the minimum `rust-version` compatible with your project, you can use third-party tools like [`cargo-msrv`](https://crates.io/crates/cargo-msrv).
+
+When used on packages that get published, we recommend [verifying the `rust-version`](../guide/continuous-integration.md#verifying-rust-version).
+
 ### The `description` field
 
 The description is a short blurb about the package. [crates.io] will display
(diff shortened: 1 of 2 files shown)


src/doc/src/reference/manifest.md
```diff
@@ -185,6 +185,10 @@ The `rust-version` may be ignored using the `--ignore-rust-version` option.
 Setting the `rust-version` key in `[package]` will affect all targets/crates in
 the package, including test suites, benchmarks, binaries, examples, etc.
 
+To find the minimum `rust-version` compatible with your project, you can use third-party tools like [`cargo-msrv`](https://crates.io/crates/cargo-msrv).
+
+When used on packages that get published, we recommend [Verifying `rust-version`](../guide/continuous-integration.md#verifying-rust-version).
```
weihanglo, line 190: nit: Should it not be capitalized?

```suggestion
When used on packages that get published, we recommend [verifying `rust-version`](../guide/continuous-integration.md#verifying-rust-version).
```

weihanglo (commented):
That's exactly what I was talking about! Thanks for this.

weihanglo (approved):

PR #13065, opened 2023-11-28
fix: reorder `--remap-path-prefix` flags for `-Zbuild-std`

weihanglo:

Order of `--remap-path-prefix` flags is important for `-Zbuild-std`.
We want to show `/rustc/<hash>/library/std` instead of `std-0.0.0`.

Fixes rust-lang/rust#117839

Follow the steps in rust-lang/rust#117839, or run

```
CARGO_RUN_BUILD_STD_TESTS=true cargo +nightly t --test build-std
```

to verify.

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1214,8 +1214,6 @@ fn trim_paths_args(
         }
         remap
     };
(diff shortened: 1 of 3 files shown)


ehuss:
Thanks!

@bors r+

PR #13066, opened 2023-11-28
fix(resolver): De-prioritize no-rust-version in MSRV resolver

epage:

This is a corner case without a good answer.
As such, this change leans on some happy-path entries existing and
preferring those.

### Additional information

This was originally discussed around the time of #12950 but was held off.

When working on this, I was considering other heuristics like
- If a future version has an MSRV, assume that it applies also to the current version
  - This can be added in the future
  - We likely would want to consider an alternative value, like inferring the rust-version from the manifest or the rust-version used from publish
- Sort no-MSRV versions of a package by minimal versions
  - The lower the version, the more likely it is to be compatible
  - This likely could apply to incompatible MSRVs (or we could reverse-sort those by rust-version) but those will error anyways without `--ignore-rust-version`, so I decided against these
  - I realized this was a backdoor to minimal versions for dependencies without a MSRV and that the community support isn't there for that yet to be a high enough quality of an experience

### DIFF
--- a/src/cargo/core/resolver/version_prefs.rs
+++ b/src/cargo/core/resolver/version_prefs.rs
@@ -232,8 +232,11 @@ mod test {
         vp.max_rust_version(Some("1.50".parse().unwrap()));
 
         let mut summaries = vec![
-            summ("foo", "1.2.4", Some("1.60")),
-            summ("foo", "1.2.3", Some("1.50")),
+            summ("foo", "1.2.4", None),
+            summ("foo", "1.2.3", Some("1.60")),
+            summ("foo", "1.2.2", None),
+            summ("foo", "1.2.1", Some("1.50")),
+            summ("foo", "1.2.0", None),
             summ("foo", "1.1.0", Some("1.40")),
             summ("foo", "1.0.9", None),
         ];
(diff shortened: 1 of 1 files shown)


src/cargo/core/resolver/version_prefs.rs
```diff
@@ -84,7 +84,38 @@ impl VersionPreferences {
...
+                        if a <= max_rust_version {
+                            return Ordering::Less;
+                        } else {
+                            return Ordering::Greater;
+                        }
+                    }
+                }
                 let msrv_a = a.rust_version() <= self.max_rust_version.as_ref();
                 let msrv_b = b.rust_version() <= self.max_rust_version.as_ref();
                 let msrv_cmp = msrv_a.cmp(&msrv_b).reverse();
```
Eh2406, line 121: Do we still need these comparisons?

src/cargo/core/resolver/version_prefs.rs
```diff
@@ -84,7 +84,38 @@ impl VersionPreferences {
                 return previous_cmp;
             }
 
-            if self.max_rust_version.is_some() {
+            if let Some(max_rust_version) = &self.max_rust_version {
+                match (a.rust_version(), b.rust_version()) {
+                    // Fallback
+                    (None, None) => {}
+                    (Some(a), Some(b)) if a == b => {}
```
Eh2406, line 91: Can be combined with the previous as `(a, b) if a == b =>`

epage: The compiler doesn't recognize that `None == None` and complains about the `match` being non-exhaustive

PR #13069, opened 2023-11-29
doc: clarify different target has different set of `CARGO_CFG_*` values

### LINKED ISSUE #12728: CARGO_CFG_TARGET_FAMILY is not defined for build-script in no_std
### Problem

There is an issue reported [downstream](https://github.com/lukaslueg/built/issues/59#issuecomment-1727271559) where the `CARGO_CFG_TARGET_FAMILY` environment variable is not defined when executing a build-script for a `no_std` binary. The said environment variable seems to be indeed not defined at all. AFAICS this is contrary to what is documented about the environment variables set by cargo, e.g. [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html) or [here](https://github.com/rust-lang/cargo/issues/4302).

### Steps

* Set `main.rs` to be `[no_std]`
* Add a `build.rs`
* Try to observe `CARGO_CFG_TARGET_FAMILY` in `build.rs` while building the `no_std`-binary.
* While most variables are defined as documented, `CARGO_CFG_TARGET_FAMILY` is not present in the build-script's environment.

### Possible Solution(s)

The way I read the documentation, the listed environment variables (including `CARGO_CFG_TARGET_FAMILY`, even in a `no_std` setting) should _always_ be defined. If they have no value, possibly because they have no meaning in the given setting, they could be set to the empty string / `NULL`. The exact behavior of subprocesses launched by a build-script (if any) seems to vary with respect to empty environment variables, though.

If the full set of environment variables mentioned in the documentation can't be made present in all settings, this should be documented.

### Notes

Notice that `CARGO_CFG_TARGET_OS: none` **is** defined in the downstream bug report, while `CARGO_CFG_TARGET_FAMILY` is not.

### Version

```text
release: 1.72.0
commit-date: 2023-08-15
host: x86_64-apple-darwin
libcurl: 8.1.2 (sys:0.4.63+curl-8.1.2 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 13.5.2 [64-bit]
```

weihanglo:

Clarify that available `CARGO_CFG_*` variables are not universally the same for each target.

Resolves #12728

Chose to use block quote. Might be too much.

Screenshot:

image


src/doc/src/reference/environment-variables.md
```diff
@@ -353,6 +353,8 @@ let out_dir = env::var("OUT_DIR").unwrap();
     * `CARGO_CFG_TARGET_POINTER_WIDTH=64` --- The CPU [pointer width].
     * `CARGO_CFG_TARGET_ENDIAN=little` --- The CPU [target endianness].
     * `CARGO_CFG_TARGET_FEATURE=mmx,sse` --- List of CPU [target features] enabled.
+  > Note that different [target triple][Target Triple] has a different set of `cfg` values,
```
ehuss, line 356: Thanks, just a minor tweak of the wording:

```suggestion
  > Note that different [target triples][Target Triple] have different sets of `cfg` values,
```

weihanglo: Updated. Thanks!

Issue #11010, opened 2022-08-20
Have `cargo add --optional <dep>` create a `<dep> = "dep:<dep>` feature


epage:
### Problem

This is inspired by my running `cargo add` and seeing a lot of features that might be there only because people have not updated to the new `dep:` syntax to disable unintended implicit features.  For my crates, I'm aware of the syntax and plan to transition my crates.  For others, they might not be aware the feature exists and by us `cargo add` turning it explicit, it would help raise visibility on these features to help lead more people to using it.

In addition,  https://github.com/rust-lang/cargo/issues/10556 mentions the idea of `cargo fix` doing

> possibly define a feature named "package" if it doesn't already exist and dep:package isn't already used

ie by convention moving away from implicit feature declarations.

### Proposed Solution

`cargo add --optional <dep>` would create a `<dep> = "dep:<dep>` feature iff
- `rust-version` is unset or is new enough for the syntax
- `dep:<dep>` doesn't already exist

Bina1791:
> ### Problem
> 
> This is inspired by my running `cargo add` and seeing a lot of features that might be there only because people have not updated to the new `dep:` syntax to disable unintended implicit features.  For my crates, I'm aware of the syntax and plan to transition my crates.  For others, they might not be aware the feature exists and by us `cargo add` turning it explicit, it would help raise visibility on these features to help lead more people to using it.
> 
> In addition,  https://github.com/rust-lang/cargo/issues/10556 mentions the idea of `cargo fix` doing
> 
> > possibly define a feature named "package" if it doesn't already exist and dep:package isn't already used
> 
> ie by convention moving away from implicit feature declarations.
> 
> ### Proposed Solution
> 
> `cargo add --optional <dep>` would create a `<dep> = "dep:<dep>` feature iff
> - `rust-version` is unset or is new enough for the syntax
> - `dep:<dep>` doesn't already exist

epage:
This will be needed for #12826 once we no longer allow optional dependencies to be unreferenced.  I'd propose we go ahead and implement it now to help raise awareness of `dep:` and prepare for that future.

epage:
The contrib guide will soon have some guidance on how we should be handling formatting.  See #12836

LuuuXXX:
Looks like nobody's handling it. I'd like to try it. @rustbot claim

PR #13071, opened 2023-11-29
Have cargo add --optional <dep> create a <dep> = "dep:<dep> feature

LuuuXXX:

`cargo add --optional <dep>` would create a `<dep> = "dep:<dep>` feature iff

- `rust-version` is unset or is new enough for the syntax
- `dep:<dep>` doesn't already exist

Fixes https://github.com/rust-lang/cargo/issues/11010

As the `dep:` syntax is only available starting with Rust 1.60. this pr maintains the previous usage convention in the earlier version.

run

```shell
cargo add --optional <dep>
```

with different rust-version to verify.

### DIFF
--- a/src/cargo/ops/cargo_add/mod.rs
+++ b/src/cargo/ops/cargo_add/mod.rs
@@ -7,6 +7,7 @@ use std::collections::BTreeSet;
 use std::collections::VecDeque;
 use std::fmt::Write;
 use std::path::Path;
+use std::str::FromStr;
 
 use anyhow::Context as _;
 use cargo_util::paths;
@@ -196,6 +197,20 @@ pub fn add(workspace: &Workspace<'_>, options: &AddOptions<'_>) -> CargoResult<(
         print_dep_table_msg(&mut options.config.shell(), &dep)?;
 
         manifest.insert_into_table(&dep_table, &dep)?;
+        if dep.optional == Some(true) {
+            let is_namespaced_features_supported =
+                check_rust_version_for_optional_dependency(options.spec.rust_version())?;
+            if is_namespaced_features_supported {
+                let dep_key = dep.toml_key();
+                if !manifest.is_explicit_dep_activation(dep_key) {
+                    let table = manifest.get_table_mut(&[String::from("features")])?;
+                    let dep_name = dep.rename.as_deref().unwrap_or(&dep.name);
+                    let new_feature: toml_edit::Value =
+                        [format!("dep:{dep_name}")].iter().collect();
+                    table[dep_key] = toml_edit::value(new_feature);
+                }
+            }
+        }
         manifest.gc_dep(dep.toml_key());
     }
 
@@ -467,6 +482,26 @@ fn check_invalid_ws_keys(toml_key: &str, arg: &DepOp) -> CargoResult<()> {
         anyhow::bail!("{}", err_msg(toml_key, "--rename", "package"))
     }
     Ok(())
+}
+
+/// When the `--optional` option is added using `cargo add`, we need to
+/// check the current rust-version. As the `dep:` syntax is only avaliable
+/// starting with Rust 1.60.0
+///
+/// `true` means that the rust-version is None or the rust-version is higher
+/// than the version needed.
+///
+/// Note: Previous versions can only use the implicit feature name.
+fn check_rust_version_for_optional_dependency(
+    rust_version: Option<&RustVersion>,
+) -> CargoResult<bool> {
+    match rust_version {
+        Some(version) => {
+            let syntax_support_version = RustVersion::from_str("1.60.0")?;
+            Ok(&syntax_support_version <= version)
+        }
+        None => Ok(true),
+    }
 }
 
 /// Provide the existing dependency for the target table
--- a/src/cargo/util/toml_mut/manifest.rs
+++ b/src/cargo/util/toml_mut/manifest.rs
@@ -420,7 +420,7 @@ impl LocalManifest {
         }
     }
 
-    fn is_explicit_dep_activation(&self, dep_key: &str) -> bool {
+    pub fn is_explicit_dep_activation(&self, dep_key: &str) -> bool {
         if let Some(toml_edit::Item::Table(feature_table)) = self.data.as_table().get("features") {
             for values in feature_table
                 .iter()
--- a/tests/testsuite/cargo_add/change_rename_target/out/Cargo.toml
+++ b/tests/testsuite/cargo_add/change_rename_target/out/Cargo.toml
@@ -6,3 +6,6 @@ version = "0.0.0"
 
 [dependencies]
 some-package = { package = "my-package2", version = "99999.0.0", optional = true }
+
+[features]
+some-package = ["dep:some-package"]
--- a/tests/testsuite/cargo_add/overwrite_optional_with_optional/mod.rs
+++ b/tests/testsuite/cargo_add/overwrite_optional_with_optional/mod.rs
@@ -1,0 +1,36 @@
+use cargo_test_support::compare::assert_ui;
+use cargo_test_support::prelude::*;
+use cargo_test_support::Project;
+
+use cargo_test_support::curr_dir;
+
+#[cargo_test]
+fn case() {
+    cargo_test_support::registry::init();
+    for ver in [
+        "0.1.1+my-package",
+        "0.2.0+my-package",
+        "0.2.3+my-package",
+        "0.4.1+my-package",
+        "20.0.0+my-package",
+        "99999.0.0+my-package",
+        "99999.0.0-alpha.1+my-package",
+    ] {
+        cargo_test_support::registry::Package::new("my-package1", ver).publish();
+    }
+
+    let project = Project::from_template(curr_dir!().join("in"));
+    let project_root = project.root();
+    let cwd = &project_root;
+
+    snapbox::cmd::Command::cargo_ui()
+        .arg("add")
+        .arg_line("my-package1 --optional")
+        .current_dir(cwd)
+        .assert()
+        .success()
+        .stdout_matches_path(curr_dir!().join("stdout.log"))
+        .stderr_matches_path(curr_dir!().join("stderr.log"));
+
+    assert_ui().subset_matches(curr_dir!().join("out"), &project_root);
+}
--- a/tests/testsuite/cargo_add/detect_workspace_inherit_optional/out/primary/Cargo.toml
+++ b/tests/testsuite/cargo_add/detect_workspace_inherit_optional/out/primary/Cargo.toml
@@ -4,3 +4,6 @@ version = "0.0.0"
 
 [dependencies]
 foo = { workspace = true, optional = true }
+
+[features]
+foo = ["dep:foo"]
--- a/tests/testsuite/cargo_add/optional/out/Cargo.toml
+++ b/tests/testsuite/cargo_add/optional/out/Cargo.toml
@@ -7,3 +7,7 @@ version = "0.0.0"
 [dependencies]
 my-package1 = { version = "99999.0.0", optional = true }
 my-package2 = { version = "0.4.1", optional = true }
+
+[features]
+my-package1 = ["dep:my-package1"]
+my-package2 = ["dep:my-package2"]
--- a/tests/testsuite/cargo_add/overwrite_git_with_path/out/primary/Cargo.toml
+++ b/tests/testsuite/cargo_add/overwrite_git_with_path/out/primary/Cargo.toml
@@ -6,3 +6,6 @@ version = "0.0.0"
 
 [dependencies]
 cargo-list-test-fixture-dependency = { optional = true, path = "../dependency", version = "0.0.0" }
+
+[features]
+cargo-list-test-fixture-dependency = ["dep:cargo-list-test-fixture-dependency"]
--- a/tests/testsuite/cargo_add/overwrite_inherit_optional_noop/out/primary/Cargo.toml
+++ b/tests/testsuite/cargo_add/overwrite_inherit_optional_noop/out/primary/Cargo.toml
@@ -4,3 +4,6 @@ version = "0.0.0"
 
 [dependencies]
 foo = { workspace = true, optional = true }
+
+[features]
+foo = ["dep:foo"]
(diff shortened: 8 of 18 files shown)


src/cargo/ops/cargo_add/mod.rs
```diff
@@ -196,6 +196,14 @@ pub fn add(workspace: &Workspace<'_>, options: &AddOptions<'_>) -> CargoResult<(
         print_dep_table_msg(&mut options.config.shell(), &dep)?;
 
         manifest.insert_into_table(&dep_table, &dep)?;
+        if let Some(option) = dep.optional {
+            let rust_version_check =
+                check_rust_version_for_optional_dependency(workspace.rust_version())?;
```
epage, line 201: We should be checking the rust-version of the package being modified, rather than the lowest common denominator of the workspace.

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -469,6 +477,26 @@ fn check_invalid_ws_keys(toml_key: &str, arg: &DepOp) -> CargoResult<()> {
...
+/// than the version needed.
+///
+/// Note: Previous versions can only use the implicit feature name.
+fn check_rust_version_for_optional_dependency(
+    rust_version: Option<&RustVersion>,
+) -> CargoResult<bool> {
+    match rust_version {
+        Some(version) => {
+            let syntax_support_version = String::from("1.60.0");
+            Ok(version.to_string().cmp(&syntax_support_version).is_ge())
```
epage, line 494: string comparisons aren't accurate.  I'd recommend directly constructing a `RustVersion` and using that for the comparison

src/cargo/util/toml_mut/manifest.rs
```diff
@@ -363,6 +363,30 @@ impl LocalManifest {
         Ok(())
     }
 
+    /// Add feature entry to a Cargo.toml.
+    pub fn insert_into_feature_table(
+        &mut self,
+        table_path: &[String],
```
epage, line 369: The name of this function says what table we are editing, so we shouldn't need to accept `table_path`

src/cargo/util/toml_mut/manifest.rs
```diff
@@ -363,6 +363,30 @@ impl LocalManifest {
...
+        dep: &Dependency,
+    ) -> CargoResult<()> {
```
epage, line 371: This function has a fairly specific policy (handling for this feature) where the name is fairly generic.

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -921,6 +928,76 @@ impl Display for WorkspaceSource {
     }
 }
 
+pub struct Feature {
```
epage, line 931: This is confusing, a "`Feature`" is both the feature name + activations and the feature activation?

These two roles should be decoupled.

I also suspect this is a bit over-engineered.  We most likely don't need all of these types but can do the editing more directly.

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -921,6 +928,76 @@ impl Display for WorkspaceSource {
...
+        toml_edit::value(features)
+    }
+
+    /// Check whether `dep:<dep>` is defined in the features section.
+    ///
+    /// `true`  if there is a define `dep:<dep>`
+    /// `false` if `dep:<dep>` is not used ever.
+    ///
+    /// Make sure that `dep:<dep>` is included in one of features in the features table.
+    pub fn check_duplicate_feature(&self, item: &mut toml_edit::Item) -> bool {
```
epage, line 981: The name of this function does not match what it does.

I would expect this to either
- Check for two features of the same name (which can't happen)
- Check for duplicate activations within a single feature but we need to check for an existing activation of a feature value among all features

tests/testsuite/cargo_add/change_rename_target/out/Cargo.toml
```diff
@@ -6,3 +6,6 @@ version = "0.0.0"
 
 [dependencies]
 some-package = { package = "my-package2", version = "99999.0.0", optional = true }
+
+[features]
+some-package = ["dep:some-package"]
```
epage, line 11: Can you double check that we have an `overwrite` test that will make sure we don't add `some-package` if `dep:some-package` is already in use elsewhere?

If so, I'd recommend making the test its own commit and make it the first in this PR and it should pass.  Then the behavior-change commit wouldn't change the test as nothing would be affected, making it clear to reviewers that it worked.

epage (commented):
btw feel free to take advantage of https://github.com/rust-lang/cargo/wiki/Office-Hours

tests/testsuite/cargo_add/change_rename_target/out/Cargo.toml
LuuuXXX: It seems no `overwrite` test for that. I added one in this PR：tests/testsuite/cargo_add/overwrite_optional_with_optional

src/cargo/ops/cargo_add/mod.rs
LuuuXXX: thanks and solved.

src/cargo/util/toml_mut/manifest.rs
LuuuXXX: thanks and solved.

src/cargo/ops/cargo_add/mod.rs
LuuuXXX: thanks and solved.

src/cargo/util/toml_mut/dependency.rs
LuuuXXX: thanks and Modified to `is_active_as_feature`

src/cargo/util/toml_mut/dependency.rs
LuuuXXX: Looks like the `Feature` is redundant. I modified the implementation.

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -196,6 +197,13 @@ pub fn add(workspace: &Workspace<'_>, options: &AddOptions<'_>) -> CargoResult<(
         print_dep_table_msg(&mut options.config.shell(), &dep)?;
 
         manifest.insert_into_table(&dep_table, &dep)?;
+        if let Some(option) = dep.optional {
+            let rust_version_check =
+                check_rust_version_for_optional_dependency(options.spec.rust_version())?;
+            if option && rust_version_check {
```
epage, line 203: Can't we simplify the first if to `if dep.optional == Some(true)`?

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -196,6 +197,13 @@ pub fn add(workspace: &Workspace<'_>, options: &AddOptions<'_>) -> CargoResult<(
         print_dep_table_msg(&mut options.config.shell(), &dep)?;
 
         manifest.insert_into_table(&dep_table, &dep)?;
+        if let Some(option) = dep.optional {
+            let rust_version_check =
```
epage, line 201: nit: can we rename this to `is_namespaced_features_supported`?

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -469,6 +477,26 @@ fn check_invalid_ws_keys(toml_key: &str, arg: &DepOp) -> CargoResult<()> {
...
+/// than the version needed.
+///
+/// Note: Previous versions can only use the implicit feature name.
+fn check_rust_version_for_optional_dependency(
+    rust_version: Option<&RustVersion>,
+) -> CargoResult<bool> {
+    match rust_version {
+        Some(version) => {
+            let syntax_support_version = RustVersion::from_str("1.60.0")?;
+            Ok(version.cmp(&syntax_support_version).is_ge())
```
epage, line 494: Why `.cmp(...).is_get()` rather than `syntax_support_version <= version`?

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -921,6 +921,47 @@ impl Display for WorkspaceSource {
...
+        }
+    }
+
+    /// Check whether `dep:<dep>` is defined in the features section.
+    ///
+    /// `true`  if there is a define `dep:<dep>`
+    /// `false` if `dep:<dep>` is not used ever.
+    ///
+    /// Make sure that `dep:<dep>` is included in one of features in the features table.
+    pub fn is_active_as_feature(&self, item: &mut toml_edit::Item) -> bool {
```
epage, line 951: Why `&mut`?

epage, line 951: Can we give `item` a more specific name?

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -921,6 +921,47 @@ impl Display for WorkspaceSource {
...
+        item.as_table_like_mut().unwrap().iter().any(|(_, values)| {
+            if let Some(values) = &values.as_array() {
```
epage, line 953: `values` -> `activations`?

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -921,6 +921,47 @@ impl Display for WorkspaceSource {
...
+                values.iter().any(|value| match value.as_str() {
```
epage, line 954: `value` -> `activation`?

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -921,6 +921,47 @@ impl Display for WorkspaceSource {
...
+                    Some(val) => val.to_string().eq(&self.as_feature()),
```
epage, line 955: - Why do we need `to_string`?
- Why `.eq` rather than `==`?

src/cargo/util/toml_mut/dependency.rs
epage, line 951: By making this `pub` on `Dependency`, this over-generalizes some fairly specific logic.

Seems like this would be better as a free function near the lone caller

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -921,6 +921,47 @@ impl Display for WorkspaceSource {
...
+        assert!(
+            crate_root.is_absolute(),
+            "Absolute path needed, got: {}",
+            crate_root.display()
+        );
+        let features: toml_edit::Value = vec![self.as_feature()].iter().collect();
+        toml_edit::value(features)
+    }
+
+    pub fn as_feature(&self) -> String {
```
epage, line 937: Maybe `as_namespaced_feature`?

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -921,6 +921,47 @@ impl Display for WorkspaceSource {
     }
 }
 
+impl Dependency {
+    /// Convert denpency as feature to TOML.
+    /// Panics if the path is relative
+    pub fn to_toml_as_feature(&self, crate_root: &Path) -> toml_edit::Item {
```
epage, line 927: By putting this here it and `pub`, it over-generalizes this especially when we already have `as_feature`.  I think it'd be cleaner and easier to read if this was inlined into the caller

src/cargo/util/toml_mut/manifest.rs
epage: FYI the name is still fairly generic

tests/testsuite/cargo_add/overwrite_optional_with_optional/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+        }
+    }
+
+    let project = Project::from_template(curr_dir!().join("in"));
+    let project_root = project.root();
+    let cwd = &project_root;
+
+    snapbox::cmd::Command::cargo_ui()
+        .arg("add")
+        .arg_line("my-package1 my-package2@0.4.1 --optional")
```
epage, line 30: We likely don't need two packages and all of those versions of the packages for what this test is doing.

Yes, you copied it from other tests but we want to work to minimize those.  Updating everything can be slow but we can at least not add more to that backlog by ensuring new tests don't do it.

epage:
Thanks!  This is heading in a very good direction!

src/cargo/util/toml_mut/manifest.rs
LuuuXXX: what about `explicit_activate_optional_dependency_as_feature`

src/cargo/ops/cargo_add/mod.rs
LuuuXXX: Thanks for the advice and solved.

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -921,6 +921,47 @@ impl Display for WorkspaceSource {
...
+        assert!(
+            crate_root.is_absolute(),
+            "Absolute path needed, got: {}",
+            crate_root.display()
+        );
```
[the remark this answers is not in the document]
LuuuXXX: Thanks and has been removed

src/cargo/util/toml_mut/dependency.rs
LuuuXXX: Resolved and use `Manifest::is_explicit_dep_activation` instead

src/cargo/util/toml_mut/dependency.rs
LuuuXXX: Resolved and use `Manifest::is_explicit_dep_activation` instead

src/cargo/util/toml_mut/dependency.rs
LuuuXXX: Resolved and use `Manifest::is_explicit_dep_activation` instead

src/cargo/util/toml_mut/dependency.rs
LuuuXXX: Resolved and use `Manifest::is_explicit_dep_activation` instead

src/cargo/util/toml_mut/dependency.rs
LuuuXXX: Resolved and use `Manifest::is_explicit_dep_activation` instead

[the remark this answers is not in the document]
LuuuXXX: Thanks and updated

tests/testsuite/cargo_add/overwrite_optional_with_optional/mod.rs
LuuuXXX: Yes, redundant packages are deleted.

epage:
Thanks!

@bors r+

Issue #13074, opened 2023-11-29
`-Zrustdoc-scrape-examples` not working if there is no workspace


GuillaumeGomez:
### Problem

In https://github.com/killercup/test-rustdoc-example-scraping-across-workspace-on-docrs/issues/1 and in https://github.com/rust-lang/docs.rs/issues/2305, we were able to find out that if you re-export non-workspace members, examples would not be scraped for them

### Steps

```console
$ git clone https://github.com/killercup/test-rustdoc-example-scraping-across-workspace-on-docrs.git
$ cargo +nightly doc -Zunstable-options -Zrustdoc-scrape-examples
$ ... open in browser and see no examples
$ git checkout HEAD~
$ cargo +nightly doc -Zunstable-options -Zrustdoc-scrape-examples
$ ... open in browser and see examples
```

### Possible Solution(s)

When scraping examples from a package, cargo only asks for examples that use APIs from specific crates, presumably for performance reasons

```rust
        // Only scrape example for items from crates in the workspace, to reduce generated file size
        for pkg in cx.bcx.ws.members() {
            let names = pkg
                .targets()
                .iter()
                .map(|target| target.crate_name())
                .collect::<HashSet<_>>();
            for name in names {
                rustdoc.arg("--scrape-examples-target-crate").arg(name);
            }
        }
```

This is filtered to workspace members.  Really what is needed is filtering for public dependencies which has recently been revived (rust-lang/rfcs#3516).

If the performance hit is not noticeable enough, we likely can remove this completely.

If the performance his is noticeable enough, the path forward:
- Long term: switch this to filtering on public normal dependencies (recursively)
- Short term: decide whether to
  - keep filtering by workspace members
  - take the perf hit and do no filtering
  - filter for all normal dependencies, recursively

epage:
@willcrichton Wanted to double check on your motivation for doing the workspace member filtering.  Happen to remember if this was benchmarked?

epage:
If we filter by public dependency, because of the previous public dependency RFC, most of the infrastructure is in place. You'd just need to check if the unstable feature is active to make the behavior conditioned on the short term vs long term solution and then check the public dependencies in the long term solution.

The one piece of plumbing that might be missing is the taking the public dependency information at the package/dependency level and using it at the Unit level. I'm assuming its in place though.

Regardless of the short term solution, I'd recommend we include the long term solution of public dependencies because that is on a path for being stable for 2024 edition.

willcrichton:
[...]

PR #13077, opened 2023-11-30
Don't filter on workspace members when scraping doc examples

GuillaumeGomez:
Fixes https://github.com/rust-lang/cargo/issues/13074.

I confirmed locally that it fixed the issue in docs.rs.

cc @willcrichton 
r? @epage

### DIFF
--- a/tests/testsuite/docscrape.rs
+++ b/tests/testsuite/docscrape.rs
@@ -48,9 +48,8 @@ fn basic() {
     assert!(p.build_dir().join("doc/src/ex/ex.rs.html").exists());
 }
 
-// This test ensures that if there is no `[workspace]` in the top-level `Cargo.toml` file, the
-// dependencies will not get their examples scraped and that they appear in the generated
-// documentation.
+// This test ensures that even if there is no `[workspace]` in the top-level `Cargo.toml` file, the
+// dependencies will get their examples scraped and that they appear in the generated documentation.
 #[cargo_test(nightly, reason = "-Zrustdoc-scrape-examples is unstable")]
 fn scrape_examples_for_non_workspace_reexports() {
     let p = project()
@@ -124,7 +123,7 @@ impl Foo {
         .run();
 
     let doc_html = p.read_file("target/doc/foo/struct.Foo.html");
-    assert!(!doc_html.contains("Examples found in repository"));
+    assert!(doc_html.contains("Examples found in repository"));
 }
 
 #[cargo_test(nightly, reason = "rustdoc scrape examples flags are unstable")]
--- a/tests/testsuite/docscrape.rs
+++ b/tests/testsuite/docscrape.rs
@@ -48,6 +48,85 @@ fn basic() {
     assert!(p.build_dir().join("doc/src/ex/ex.rs.html").exists());
 }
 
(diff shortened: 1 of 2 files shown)


killercup (approved):
Thanks so much for seeing this through! Love how small the fix ended up being :)

GuillaumeGomez:
I'm working on adding a test based on the repository @killercup nicely made. :)

tests/testsuite/docscrape.rs
```diff
@@ -48,6 +48,127 @@ fn basic() {
     assert!(p.build_dir().join("doc/src/ex/ex.rs.html").exists());
 }
 
+// This test ensures that even if there is no `[workspace]` in the top-level `Cargo.toml` file, the
+// dependencies will get their examples scraped and that they appear in the generated documentation.
+#[cargo_test(nightly, reason = "rustdoc scrape examples flags are unstable")]
```
epage, line 53: Please name the specific Z flag.  It makes it easier for us to find and update these

e.g.
```
tests/testsuite/check_cfg.rs
32:#[cargo_test(nightly, reason = "--check-cfg is unstable")]
```

tests/testsuite/docscrape.rs
```diff
@@ -48,6 +48,127 @@ fn basic() {
...
+        .run();
+
+    let doc_html = p.read_file("target/doc/foo/struct.Foo.html");
+    assert!(doc_html.contains("Examples found in repository"));
+    // `Foo` is used in the two examples so since there are more than one example, we should have
+    // this sentence.
+    assert!(doc_html.contains("More examples"));
+    let doc_html = p.read_file("target/doc/foo/struct.Bar.html");
+    assert!(doc_html.contains("Examples found in repository"));
+    // There is only only usage of `Bar` so we shouldn't have this sentence in the docs.
```
epage, line 168: nit: "only only"

tests/testsuite/docscrape.rs
```diff
@@ -48,6 +48,127 @@ fn basic() {
...
+    assert!(!doc_html.contains("More examples"));
+}
```
epage, line 170: As we just need to make sure examples are found, can we simplify this?
- Only one dependency
- Only one example

I'm most interested in minimizing the number of assumptions we make in this test on rustdoc's output so its less likely to fail on upgrade and its easier to update when it does fail.

tests/testsuite/docscrape.rs
```diff
@@ -48,6 +48,127 @@ fn basic() {
     assert!(p.build_dir().join("doc/src/ex/ex.rs.html").exists());
 }
 
+// This test ensures that even if there is no `[workspace]` in the top-level `Cargo.toml` file, the
+// dependencies will get their examples scraped and that they appear in the generated documentation.
+#[cargo_test(nightly, reason = "rustdoc scrape examples flags are unstable")]
+fn without_workspace() {
```
epage, line 54: nit: can we make the name more descriptive, like `scrape_examples_for_non_workspace_reexports`?

GuillaumeGomez:
I updated the test name, simplified the test, rewrote the git history, fixed the typo and updated the nightly reason.

epage:
Thanks!

@bors r+

PR #13080, opened 2023-11-30
refactor(toml): Decouple logic from schema

epage:

This tries to decouple cargo logic from the schemas so we can split the schemas out into a separate crate for #12801

Profile layering was one of the few pieces of logic I kept in the schema, assuming its general enough / decoupled enough from cargo policies.

Each step is broken down into its own commit for easier analysis


src/cargo/util/toml/mod.rs
```diff
@@ -508,11 +508,12 @@ impl schema::TomlManifest {
 
         package.version = version.clone().map(schema::InheritableField::Value);
 
-        let pkgid = package.to_package_id(
-            source_id,
+        let pkgid = PackageId::pure(
```
weihanglo: I would love to see eithe rename this function or have a doc comment on it. It always puzzle me what `pure` means.

src/cargo/util/toml/schema.rs
```diff
@@ -670,6 +670,98 @@ pub struct TomlProfile {
     pub trim_paths: Option<TomlTrimPaths>,
 }
 
+impl TomlProfile {
+    /// Overwrite self's values with the given profile.
+    pub fn merge(&mut self, profile: &Self) {
```
weihanglo, line 675: > Profile layering was one of the few pieces of logic I kept in the schema, assuming its general enough / decoupled enough from cargo policies.

How is thing general enough? Do we have any possible use case outside cargo?

weihanglo (approved):
Mostly LGTM. Thank you!

weihanglo:
One thing worth calling out is that free functions sometimes are hard to find. By grouping methods under an item I can get a better auto-completion experience whey typing `things.<tab>`.

src/cargo/util/toml/mod.rs
epage: How about I handle this in a follow up PR (branch already exists locally) as this is unrelated to the current PR?

src/cargo/util/toml/schema.rs
epage: From what I understand of profiles, the merging is inherent to their definition, so if someone wants to do process the manifest, they'll need it.

I know its a bit dubious about why this when other things don't get included.  The biggest motivation is that this is tightly coupled to the struct definition and keeping them close is important to avoid forgetting to merge new fields.  The reason I felt this as justifiable was that it is very direct without special logic.

epage:
> One thing worth calling out is that free functions sometimes are hard to find. By grouping methods under an item I can get a better auto-completion experience whey typing `things.<tab>`.

This is a given if we still want to do #12801.  A lot of this logic is too tightly coupled to the internals of cargo to be able to split out without a lot of other work first (if ever)

weihanglo:
@bors r+

Thank you again for the hard work!

PR #13085, opened 2023-12-01
Add more options to registry test support.

ehuss:
This adds a few things that are missing in the registry test support code. These aren't immediately necessary, since no tests rely on them. However, I wrote it for something else that I ended up not needing, but I think it is still helpful for future work.

The additions are:

- Ability to specify `default-features=false` for a registry dependency.
- Include binary dependencies in the index for the `cargo publish` HTTTP server (current `cargo publish` tests use `file:///` upload, and don't verify the index).
- Include `rust-version` in the `cargo publish` HTTP server (current `cargo publish` tests don't verify the index).
- Include the `features=[…]` field for dependencies in the published `Cargo.toml` (cargo doesn't read features from `Cargo.toml`, it only uses the registry, but it is probably best to keep it in sync).
- Include the `package="…"` field for dependencies (for renamed dependencies) in the published `Cargo.toml` (similarly, cargo only uses the index, but it is probably good to keep in sync).

### DIFF
--- a/crates/cargo-test-support/src/registry.rs
+++ b/crates/cargo-test-support/src/registry.rs
@@ -557,6 +557,7 @@ pub struct Dependency {
     registry: Option<String>,
     package: Option<String>,
     optional: bool,
+    default_features: bool,
 }
 
 /// Entry with data that corresponds to [`tar::EntryType`].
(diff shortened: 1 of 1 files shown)


epage:
Thanks!

Even if its not immediately needed, having these gaps can be a real headache for someone to figure out why their test isn't working as expected.  hi-rustin recently ran into that and made #12997

PR #13097, opened 2023-12-01
refactor(schemas): Pull out mod for proposed schemas package

epage:
Originally for #12801 we talked about a `cargo-util-manifest-schema` package
- `util` in the name to not clash with plugins
- manifest specific to keep the scope down

The problem is we have types that aren't manifest specific, like
- `PartialVersion` (currently slated for `cargo-util-semverext`)
- `RustVersion`
- `PackageIdSpec`
- `SourceKind` (soon)

Things get messy if we try to break things down into common packages. Instead, I think it'd be useful to have a schemas package that has mods for each type of schema, re-exporting what is needed.

Normally, componentizing your package by the layer in the stack is a recipe for pain.
I don't think that'll apply here because these are meant to be so low level.

The other big concern could be compile times.  My hope is it won't be too bad.

So this moves the `util/toml` types into the module and we can add more in the future.


src/cargo/util_schemas/mood.rs
```diff
@@ -0,0 +1,3 @@
+//! Low-level Cargo format schemas
```
weihanglo, line 1: I am in a good mood to see this typo 😆

src/cargo/util_schemas/mod.rs
```diff
@@ -0,0 +1,3 @@
+//! Low-level Cargo format schemas
```
weihanglo, line 1: So there will be more types in `util_schemas`. I guess types for `.carg/config.toml` and `Cargo.toml` will all end up there? Are we going to have feature flags for different types?

It's also not clear what low-level actually means here, from a contributor's perspective.

weihanglo (commented):
Overall looks good!

src/cargo/util_schemas/mood.rs
epage: I had thought I had already made a file ...

src/cargo/util_schemas/mod.rs
epage: Added a bit more of an explanation.

My assumption is these will be light enough to not need feature flags.

weihanglo (approved):

weihanglo:
Thanks for the review and the update, people.

Let's go and merge this!

@bors r=ehuss

PR #13098, opened 2023-12-01
test(mdman): Switch to snapbox

epage:
Working with #13090 is the motivation

Benefits
- We use it elsewhere
- We don't have to bake in our own snapshotting solution
- It is more obvious how to update the snapshots

### DIFF
--- a/crates/mdman/Cargo.toml
+++ b/crates/mdman/Cargo.toml
@@ -16,7 +16,7 @@ serde_json.workspace = true
 url.workspace = true
 
 [dev-dependencies]
-pretty_assertions.workspace = true
+snapbox.workspace = true
 
 [lints]
 workspace = true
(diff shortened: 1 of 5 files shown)


crates/mdman/tests/compare.rs
```diff
@@ -1,13 +1,10 @@
 //! Compares input to expected output.
-//!
-//! Use the MDMAN_BLESS environment variable to automatically update the
-//! expected output.
 
 #![allow(clippy::disallowed_methods)]
```
weihanglo, line 3: Maybe?

```suggestion
```

crates/mdman/Cargo.toml
```diff
@@ -16,7 +16,7 @@ serde_json.workspace = true
 url.workspace = true
 
 [dev-dependencies]
-pretty_assertions.workspace = true
+snapbox.workspace = true
```
weihanglo, line 19: Nobody uses `pretty_assertions` so it can also be removed from `workspace.dependencies` I believe.

epage: :)  Sometimes its easier to delete a line than to hand write the parameters (since I don't have completions configured)

weihanglo (approved):

PR #13099, opened 2023-12-02
test(install): use TCP connection instead of thread sleep

weihanglo:

The test `install::uninstall_running_binary` is [flaky in CI job](https://github.com/rust-lang/cargo/actions/runs/7063691717/job/19230203118).

```console
running `/Users/runner/work/cargo/cargo/target/debug/cargo install foo`
running `/Users/runner/work/cargo/cargo/target/debug/cargo uninstall foo`
thread '<unnamed>' panicked at tests/testsuite/install.rs:2568:66:
called `Result::unwrap()` on an `Err` value: process didn't exit successfully: `/Users/runner/work/cargo/cargo/target/tmp/cit/t1651/home/.cargo/bin/foo` (signal: 9, SIGKILL: kill)
thread 'install::uninstall_running_binary' panicked at tests/testsuite/install.rs:2588:18:
called `Result::unwrap()` on an `Err` value: Any { .. }
```

TCP connection is more robust than `thread::sleep`, ensuring

* foo is running before the first `cargo uninstall` call
* foo is stopped before the second `cargo uninstall` call

### DIFF
--- a/tests/testsuite/install.rs
+++ b/tests/testsuite/install.rs
@@ -2522,28 +2522,32 @@ fn assert_tracker_noexistence(key: &str) {
 
 #[cargo_test]
 fn uninstall_running_binary() {
+    use std::io::Write;
+
     Package::new("foo", "0.0.1")
-        .file("src/lib.rs", "")
         .file(
             "Cargo.toml",
             r#"
-    [package]
-    name = "foo"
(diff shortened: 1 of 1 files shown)


weihanglo:
This may block cargo submodule sync in rust-lang/rust.

PR #13102, opened 2023-12-02
docs(book): make old title anchorable

weihanglo:

An alternative way to fix <https://github.com/rust-lang/blog.rust-lang.org/pull/1172>

I've tried [custom heading attribute](https://rust-lang.github.io/mdBook/format/markdown.html#heading-attributes) but that didn't work as you can only have one `id` per element.

If there is a preferred solution, do let me know.

### Additional information

r? ehuss


ehuss:
Thanks! The long-term solution to this is to support anchors in redirects. I thought there was an mdbook issue for that, but I can't find one, so I'll file a new one.

@bors r+

PR #13110, opened 2023-12-05
fix: only examine base URL when matching package

hi-rustin:

close https://github.com/hi-rustin/cargo-information/issues/57

We should not directly compare the package spec URL with the source ID URL because the package spec URL contains the package's name in the URL.

For example:

`https://github.com/rust-lang/crates.io-index/clap`

1. spec URL: `Url { scheme: "https", cannot_be_a_base: false, username: ", password: None, host: Some(Domain("github.com")), port: None, path: "/rust-lang/crates.io-index/clap", query: None, fragment: None }`
2. source id URL: `Url { scheme: "https", cannot_be_a_base: false, username: ", password: None, host: Some(Domain("github.com")), port: None, path: "/rust-lang/crates.io-index", query: None, fragment: None }`

Could you check the unit tests? Before the fix, it could fail.

### Additional information

Do we need to change the URL itself instead of the matching process?

### DIFF
--- a/src/cargo/core/package_id_spec.rs
+++ b/src/cargo/core/package_id_spec.rs
@@ -551,6 +551,13 @@ mod tests {
         assert!(PackageIdSpec::parse("https://example.com#foo@1.2")
             .unwrap()
             .matches(foo));
+        // FIXME: The two tests below need to be corrected or adjusted - they should be passing.
+        assert!(!PackageIdSpec::parse("https://example.com/foo")
+            .unwrap()
+            .matches(foo));
+        assert!(!PackageIdSpec::parse("https://example.com/foo#1.2.3")
+            .unwrap()
+            .matches(foo));
         assert!(!PackageIdSpec::parse("https://bob.com#foo@1.2")
             .unwrap()
             .matches(foo));
--- a/src/cargo/core/package_id_spec.rs
+++ b/src/cargo/core/package_id_spec.rs
@@ -181,7 +181,22 @@ impl PackageIdSpec {
         }
 
         if let Some(u) = &self.url {
-            if u != package_id.source_id().url() {
+            let package_base_url = format!(
+                "{}://{}",
+                u.scheme(),
+                u.host_str().expect("package spec url should have a host")
+            );
+            let source_id = package_id.source_id();
+            let source_id_url = source_id.url();
+            let source_id_base_url = format!(
+                "{}://{}",
+                source_id_url.scheme(),
+                source_id_url
+                    .host_str()
+                    .expect("source id url should have a host")
+            );
+            // Examine only the base URL, as the package spec URL might include a package name within its path.
+            if package_base_url != source_id_base_url {
                 return false;
             }
         }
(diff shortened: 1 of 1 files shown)


src/cargo/core/package_id_spec.rs
```diff
@@ -181,7 +181,22 @@ impl PackageIdSpec {
...
+            let source_id_url = source_id.url();
+            let source_id_base_url = format!(
+                "{}://{}",
+                source_id_url.scheme(),
+                source_id_url
+                    .host_str()
+                    .expect("source id url should have a host")
+            );
+            // Examine only the base URL, as the package spec URL might include a package name within its path.
+            if package_base_url != source_id_base_url {
```
epage, line 199: I don't think its correct to completely ignore paths within the URLs, like with git sources.

src/cargo/core/package_id_spec.rs
```diff
@@ -551,11 +566,10 @@ mod tests {
...
             .unwrap()
             .matches(foo));
-        // FIXME: The two tests below need to be corrected or adjusted - they should be passing.
-        assert!(!PackageIdSpec::parse("https://example.com/foo")
+        assert!(PackageIdSpec::parse("https://example.com/foo")
             .unwrap()
             .matches(foo));
-        assert!(!PackageIdSpec::parse("https://example.com/foo#1.2.3")
+        assert!(PackageIdSpec::parse("https://example.com/foo#1.2.3")
             .unwrap()
```
epage, line 573: What situation are you running into these package id specs existing that they should match the given package id?

From looking at the code, the implicit name is mostly to avoid redundancy, like if you depend on a package in the root of a git repo by the same name.

src/cargo/core/package_id_spec.rs
hi-rustin: Yeah. I found it from the failed CI. We also support file URLs.

src/cargo/core/package_id_spec.rs
hi-rustin: In this issue:
https://github.com/hi-rustin/cargo-information/issues/57

```console
cargo info https://github.com/rust-lang/crates.io-index/clap
    Updating crates.io index
error: could not find `https://github.com/rust-lang/crates.io-index/clap` in registry `https://github.com/rust-lang/crates.io-index`
```

As you can see, it told me it could not find the clap crate from the crates.io.

I think this was a bug because the source ID doesn't have the package name in its URL. We shouldn't directly compare it.

epage: Where did you get that package id spec from?

```rust
$ cargo pkgid clap
https://github.com/rust-lang/crates.io-index#clap@4.4.10
```

hi-rustin: Oh. I copied it from here:
https://github.com/rust-lang/cargo/blob/a5fa676731c8e570bb9be9c50d2202acea84b6b6/src/cargo/core/package_id_spec.rs#L41

hi-rustin: Can I ask what is the difference between them? When to use the `"https://crates.io/foo"`?

epage: I wonder how representative that example is of real world spec ids...

hi-rustin: I will take a look at the details of the `pkgid` command. Thank you!

PR #13112, opened 2023-12-05
feat: make bundled SQLite optional 

weihanglo:

This helps cargo-the-library consumers easier to manipulate their
dependency graph if they don't want to bundle SQLite3. For example,
Linux distro maintainers want full control of the dependency closure.

Make sure 

* `cargo build` fails when without SQLite3 development headers
* `cargo build` succeeds when with SQLite3 development headers
* `cargo build -F vendored-sqlite` always works

One drawback of this is that people now need to take care of one extra system library.

### Additional information

See also <https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/cache.20cleaning.20blog.20post/near/405596136>.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,7 +73,7 @@ proptest = "1.4.0"
 pulldown-cmark = { version = "0.9.3", default-features = false }
 rand = "0.8.5"
 regex = "1.10.2"
-rusqlite = { version = "0.30.0", features = ["bundled"] }
+rusqlite = "0.30.0"
 rustfix = { version = "0.7.0", path = "crates/rustfix" }
 same-file = "1.0.6"
 security-framework = "2.9.2"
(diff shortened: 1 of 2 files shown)


ehuss:
I'm concerned about this change, as I would prefer to not require users to install a separate dependency. Is there an issue with having distros use `LIBSQLITE3_SYS_USE_PKG_CONFIG`?

ehuss:
Another concern is that we cannot control the minimum sqlite version when using a packaged one. Thus, we would not be able to provide reliable build-time errors when the version is too old, or to fall back to the bundled version like other bindings do.

weihanglo:
> Another concern is that we cannot control the minimum sqlite version when using a packaged one.

I don't feel like that's a real issue. We should know the minimum version of SQLite Cargo depends on. But yep communication of it is a headache.

> Is there an issue with having distros use `LIBSQLITE3_SYS_USE_PKG_CONFIG`?

No. It is always an option. Just a pity that we cannot use built-in cargo feature mechanism to control it.

Agree on that it is a bit too painful to get around with dependency on each platform. That increases frictions for new contributors. So closing!

PR #13114, opened 2023-12-05
fix: explicitly remap current dir by using `.`

weihanglo:

In https://github.com/rust-lang/rust/blob/87e1447aa/compiler/rustc_codegen_llvm/src/debuginfo/metadata.rs#L856
when the remap result of `work_dir` is an empty string,
LLVM won't generate some symbols for root debuginfo node.
For example, `N_SO` and `N_OSO` on macOS,
or `DW_AT_comp_dir` and `DW_AT_GNU_dwo_name` when debuginfo is splitted.
Precisely, it is observed that when the `DIFile` of a root DI node of a
compile unit `DIFile` was provied with an empty `Directory` string,
LLVM would not emit those symbols for the root DI node.

This behavior is not desired,
resulting in corrupted debuginfo and degrading debugging experience.

This is might not be a bug of `--remap-path-prefix` in rustc,
since `-fdebug-prefix-map` in clang 16 could have the same result
(`DW_AT_comp_dir` is gone when `work_dir` is remapped to an empty string).
However, in gcc 12 `fdebug-prefix-map` will return an absolute work_dir
when an empty string occurs.

To not bother whether this needs to be fixed in rustc or not,
let's fix it by always appending an explicit `.`
when `--remap-path-prefix` remaps to relative workspace root
a.k.a. where rustc is invoking.

#### Build rustc

Without a possible fix in rustc like <https://github.com/rust-lang/rust/pull/118518>, the debuginfo is not corrputed even we dont remap to `.` explicitly. Therefore, to test it manually you need to build rustc from <https://github.com/rust-lang/rust/pull/118518>. Generally,

```
# in rust-lang/rust
./x b library --stage 1
rustup toolchain link stage1 build/host/stage1
```

Then cargo +stage1 build is available, and you can build package via

```
CARGO_PROFILE_DEV_TRIM_PATHS=object cargo +stage1 -Ztrim-paths b
```

After build a package, I would recommend playing with debuggers, either gdb or lldb.
After remapping, supposedly you need to launch the debugger from the same location the build kicked off, usually the workspace root.

Please help verify that source files still show, and breakpoints can pause/resume at either local or registry packages.

#### I don't want to build rustc

You can use either `objdump -Wi` or `readelf -wi` or `dsymutil -s` to inspect debug symbols, though it doesn't help much.

### Additional information

It would be awesome if somebody can help verify the behavior on Windows 😆.

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1230,7 +1230,7 @@ fn trim_paths_args(
         // * path dependencies outside workspace root directory
         if is_local && pkg_root.strip_prefix(ws_root).is_ok() {
             remap.push(ws_root);
-            remap.push("="); // empty to remap to relative paths.
+            remap.push("=."); // remap to relative rustc work dir explicitly
         } else {
             remap.push(pkg_root);
             remap.push("=");
--- a/tests/testsuite/profile_trim_paths.rs
+++ b/tests/testsuite/profile_trim_paths.rs
@@ -453,33 +453,65 @@ fn diagnostics_works() {
 }
 
 #[cfg(target_os = "macos")]
-#[cargo_test(requires_nm, nightly, reason = "-Zremap-path-scope is unstable")]
-fn object_works() {
-    object_works_helper(|path| {
+mod object_works {
+    use super::*;
+
+    fn inspect_debuginfo(path: &std::path::Path) -> Vec<u8> {
         std::process::Command::new("nm")
             .arg("-pa")
             .arg(path)
             .output()
             .expect("nm works")
             .stdout
-    })
+    }
+
+    #[cargo_test(requires_nm, nightly, reason = "-Zremap-path-scope is unstable")]
+    fn with_split_debuginfo_off() {
+        object_works_helper("off", inspect_debuginfo);
+    }
+
+    #[cargo_test(requires_nm, nightly, reason = "-Zremap-path-scope is unstable")]
+    fn with_split_debuginfo_packed() {
+        object_works_helper("packed", inspect_debuginfo);
+    }
+
+    #[cargo_test(requires_nm, nightly, reason = "-Zremap-path-scope is unstable")]
+    fn with_split_debuginfo_unpacked() {
+        object_works_helper("unpacked", inspect_debuginfo);
+    }
 }
 
 #[cfg(target_os = "linux")]
-#[cargo_test(requires_readelf, nightly, reason = "-Zremap-path-scope is unstable")]
-fn object_works() {
-    object_works_helper(|path| {
+mod object_works {
+    use super::*;
+
+    fn inspect_debuginfo(path: &std::path::Path) -> Vec<u8> {
         std::process::Command::new("readelf")
             .arg("-wi")
             .arg(path)
             .output()
             .expect("readelf works")
             .stdout
-    })
+    }
+
+    #[cargo_test(requires_readelf, nightly, reason = "-Zremap-path-scope is unstable")]
+    fn with_split_debuginfo_off() {
+        object_works_helper("off", inspect_debuginfo);
+    }
+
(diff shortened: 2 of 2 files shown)


src/cargo/core/compiler/mod.rs
```diff
@@ -1230,7 +1230,7 @@ fn trim_paths_args(
         // * path dependencies outside workspace root directory
         if is_local && pkg_root.strip_prefix(ws_root).is_ok() {
             remap.push(ws_root);
-            remap.push("="); // empty to remap to relative paths.
+            remap.push("=."); // explicitly remap to cwd (rustc working dir).
```
weihanglo, line 1233: Actually I am not sure if we want to fix this in cargo or in rustc…

weihanglo: * Run `clang main.c -gdwarf -fdebug-prefix-map="$PWD"= -gsplit-dwarf`
    * on Linux `DW_AT_comp_dir` is gone, as same as the result from rustc.
    * on macOS `N_SO` and `N_OSO` are gone, as same as the result from rustc.
* Run `gcc main.c -gdwarf -fdebug-prefix-map="$PWD"= -gsplit-dwarf`
    * on Linux it generates empty `DW_AT_comp_dir`.
    * on macOS it generates `N_SO` with value `/`.

Without bothering whether this should be fixed in rustc or cargo, I would propose we add an `.` for remapping the current working directory.

For more on gcc/clang remap options, see: https://reproducible-builds.org/docs/build-path/

weihanglo:
This is ready for review, but somehow a bit convoluted.

tests/testsuite/profile_trim_paths.rs
```diff
@@ -453,33 +453,65 @@ fn diagnostics_works() {
...
+
+    #[cargo_test(requires_nm, nightly, reason = "-Zremap-path-scope is unstable")]
+    fn with_split_debuginfo_packed() {
+        object_works_helper("packed", inspect_debuginfo);
+    }
+
+    #[cargo_test(requires_nm, nightly, reason = "-Zremap-path-scope is unstable")]
+    fn with_split_debuginfo_unpacked() {
+        object_works_helper("packed", inspect_debuginfo);
+    }
```
epage, line 481: I feel like I'm missing something as these two look the same

tests/testsuite/profile_trim_paths.rs
```diff
@@ -453,33 +453,65 @@ fn diagnostics_works() {
...
+
+    #[cargo_test(requires_readelf, nightly, reason = "-Zremap-path-scope is unstable")]
+    fn with_split_debuginfo_packed() {
+        object_works_helper("packed", inspect_debuginfo);
+    }
+
+    #[cargo_test(requires_readelf, nightly, reason = "-Zremap-path-scope is unstable")]
+    fn with_split_debuginfo_unpacked() {
+        object_works_helper("packed", inspect_debuginfo);
+    }
```
epage, line 510: ditto

tests/testsuite/profile_trim_paths.rs
```diff
@@ -530,14 +567,14 @@ fn object_works_helper(run: impl Fn(&std::path::Path) -> Vec<u8>) {
...
+[RUNNING] `rustc [..]-C split-debuginfo={split_debuginfo} [..]\
     -Zremap-path-scope=object \
     --remap-path-prefix={pkg_remap} \
     --remap-path-prefix=[..]/lib/rustlib/src/rust=/rustc/[..]
 [COMPILING] foo v0.0.1 ([CWD])
-[RUNNING] `rustc [..]\
+[RUNNING] `rustc [..]-C split-debuginfo={split_debuginfo} [..]\
     -Zremap-path-scope=object \
-    --remap-path-prefix=[CWD]= \
+    --remap-path-prefix=[CWD]=. \
```
epage, line 577: This looks like a result I'd expect from the follow up commit...

weihanglo: fixed 😆

Issue #11821, opened 2023-03-09
gitoxide auth tests sporadically failing


ehuss:
Some of the gitoxide tests seem to be randomly failing in CI:

* https://github.com/rust-lang/cargo/pull/11799#issuecomment-1454386556
* https://github.com/rust-lang/cargo/pull/11820#issuecomment-1462703407

The errors look like:

```
---- git_auth::http_auth_offered stdout ----
running `/Users/runner/work/cargo/cargo/target/debug/cargo build -v`
running `/Users/runner/work/cargo/cargo/target/debug/cargo check`
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `{}`,
 right: `{"GET /foo/bar/info/refs?service=git-upload-pack HTTP/1.1", "Authorization: Basic Zm9vOmJhcg==", "Accept: */*"}`', tests/testsuite/git_auth.rs:63:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'git_auth::http_auth_offered' panicked at '
test failed running `/Users/runner/work/cargo/cargo/target/debug/cargo check`
error: expected to find:
    Updating git repository `http://127.0.0.1:50629/foo/bar`
error: failed to get `bar` as a dependency of package `foo v0.0.1 [..]`

Caused by:
  failed to load source for dependency `bar`

Caused by:
  Unable to update http://127.0.0.1:50629/foo/bar

Caused by:
  failed to clone into: [..]

Caused by:
  failed to authenticate when downloading repository

  * attempted to find username/password via `credential.helper`, but [..]

  if the git CLI succeeds then `net.git-fetch-with-cli` may help here
  https://[..]

Caused by:


did not find in output:
    Updating git repository `http://127.0.0.1:50629/foo/bar`
error: failed to get `bar` as a dependency of package `foo v0.0.1 (/Users/runner/work/cargo/cargo/target/tmp/cit/testsuite/git_auth/http_auth_offered/foo)`

Caused by:
  failed to load source for dependency `bar`

Caused by:
  Unable to update http://127.0.0.1:50629/foo/bar

Caused by:
  failed to clone into: /Users/runner/work/cargo/cargo/target/tmp/cit/testsuite/git_auth/http_auth_offered/home/.cargo/git/db/bar-71575e16952ccc70

Caused by:
  failed to authenticate when downloading repository

  * attempted to find username/password via git's `credential.helper` support, but failed

[...]

PR #13117, opened 2023-12-05
re-enable previously disabled tests with Windows-specific fix

Byron:
Related to #11821 for which this is a fix. 
However, it's probably not yet the optimal solution, depending on how `stderr` of subprocesses should be handled.

### Tasks

* [x] try to fix the issue with an env var.
    - Failure, as one warning remains that seems to originate from a C# HTTP client
* [x] figure out if `stderr` should be on or off by default - on by default like before, but now one can control it.
* [x] create a new `gix` release and use it here

### Review Notes

* Personally, I think `cargo` should keep `stderr` to be inherited so users can see potentially relevant warnings or errors provided by credential helpers. Thus this is still the default, but the tests that need it explicitly disabled `stderr` of credential helpers.

----

On Windows, `gix` will call the `git-credential-manager, but with
`stderr` set to `inherit` which makes any errors visible to the user,
just like `git` does.

```
1   1         Updating git repository `https://foo.bar/foo/bar`
    2    +warning: auto-detection of host provider took too long (>2000ms)
    3    +warning: see https://aka.ms/gcm/autodetect for more information.
    4    +fatal: A task was canceled.
    5    +warning: auto-detection of host provider took too long (>2000ms)
    6    +warning: see https://aka.ms/gcm/autodetect for more information.`
````

This, however, isn't what's desirable in tests sometimes, nor may
it be desirable in Cargo.

For now, it seems easiest to disable this particular feature that
issues the warning messages, even though a future `gix` update
should allow to control what to do with `stderr`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,13 +10,14 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.8.3"
+version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
+checksum = "91429305e9f0a25f6205c5b8e0d2db09e0708a7a6df0f42212bb56c32c8ac97a"
 dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
+ "zerocopy",
 ]
 
 [[package]]
@@ -62,30 +63,30 @@ checksum = "7079075b41f533b8c61d2a4d073c4676e1f8b249ff94a393b0595db304e0dd87"
 
 [[package]]
 name = "anstyle-parse"
-version = "0.2.0"
+version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e765fd216e48e067936442276d1d57399e37bce53c264d6fefbe298080cb57ee"
+checksum = "c75ac65da39e5fe5ab759307499ddad880d724eed2f6ce5b5e8a26f4f387928c"
 dependencies = [
  "utf8parse",
 ]
 
 [[package]]
 name = "anstyle-query"
-version = "1.0.0"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5ca11d4be1bab0c8bc8734a9aa7bf4ee8316d462a08c6ac5052f888fef5b494b"
+checksum = "a3a318f1f38d2418400f8209655bfd825785afd25aa30bb7ba6cc792e4596748"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.1"
+version = "3.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f0699d10d2f4d628a98ee7b57b289abbc98ff3bad977cb3152709d4bf2330628"
+checksum = "1cd54b81ec8d6180e24654d0b371ad22fc3dd083b6ff8ba325b72e00c87660a7"
 dependencies = [
  "anstyle",
- "windows-sys 0.48.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Byron** [comment]
@weihanglo This PR is ready for review. Pleas also take a look at the Review Notes, as this issue can definitely be solved in various ways, even though I kept everything as is, while fixing the two tests specifically. Thanks for taking a look.

**Review by weihanglo — APPROVED** [review]
It's unbelievably fast! Thanks for the new release :)

**Comment by Byron** [comment]
Thanks for letting me know.


### Experiment: use the new environment variables to see if they work

I did the validation that I probably should have had before, and here is the outcome.

Stderr of external filter programs can now be swallowed as expected:

```git-config
# .git/config
[filter "noop"]
	clean = "echo clean 1>&2; cat"
	smudge = "echo smudge 1>&2; cat"
```

```git-attributes
# .gitattributes
*.filter filter=noop
```

```
delme ( main)
❯ GIX_EXTERNAL_COMMAND_STDERR=false gix rev parse -c :a.filter -b worktree
a

delme ( main)
❯ GIX_EXTERNAL_COMMAND_STDERR=1 gix rev parse -c :a.filter -b worktree
smudge
a
```

But what matters really is the credential helper:

```git-config
# .git/config
[credential]
	helper = "!echo error 1>&2"
[remote "origin"]
	url = https://github.com/user/private
	fetch = +refs/heads/*:refs/remotes/origin/*
```

```
delme ( main)
❯ GIX_CREDENTIALS_HELPER_STDERR=0 gix fetch
+refs/heads/*:refs/remotes/origin/*
        ed3c7a05308b8201cfa2917a7145acac4f663abb refs/heads/master -> refs/remotes/origin/master [up-to-date]
no negotiation was necessary

delme ( main)
❯ GIX_CREDENTIALS_HELPER_STDERR=1 gix fetch
error store
+refs/heads/*:refs/remotes/origin/*
        ed3c7a05308b8201cfa2917a7145acac4f663abb refs/heads/master -> refs/remotes/origin/master [up-to-date]
no negotiation was necessary
```

The `error store` message is being suppressed when the environment variable disables stderr, just as expected.

The wiring actually works, and I double-checked that `cargo` opens a repository in a way that allows environment variables to be read. So from that point of view it *should* work.

As the test-suite captures `stdout` and `stderr`, this means that the credentials manager also outputs its messages to one of these channels. As `stdout` is always piped, this means `git-credential-manager` must send its messages to `stderr`, which *should* be disabled. But it is not, which probably means that the envrionment variable isn't going through.

But I validated that by tracing, and any `gix` Repo that is opened has the correct configuration:

```git-config
[src/cargo/sources/git/oxide.rs:271] repo.config_snapshot() = [credential]
<tab>helper = /Users/byron/dev/github.com/rust-lang/cargo/target/tmp/cit/t0/script/target/debug/script
[core]
<tab>bare = true
<tab>repositoryformatversion = 0
<tab>filemode = true
<tab>ignorecase = true
<tab>precomposeunicode = true
[gitoxide "http"]
<tab>connectTimeout = 30000
[http]
<tab>lowSpeedLimit = 10
<tab>lowSpeedTime = 30
<tab>userAgent = cargo 1.76.0 (930d6536e 2023-12-06)
[core]
<tab>logAllRefUpdates = false
[gitoxide "credentials"]
<tab>helperStderr = 0 # from GIX_CREDENTIALS_HELPER_STDERR
```

I even validated that the credential-helper value is false:

```
[/Users/byron/dev/github.com/Byron/gitoxide/gix/src/config/snapshot/credential_helpers.rs:165] self.try_boolean(Credentials::HELPER_STDERR.logical_name().as_str()).map(|val|
                        Credentials::HELPER_STDERR.enrich_error(val)).transpose().with_leniency(self.repo.options.lenient_config)?.unwrap_or(true) = false
```

So from all I can tell, it's working as it's supposed to, yet the system-wide credential helper still manages to write to stderr.

Maybe,  it writes to the terminal directly, and *somehow* that ends up being picked up by the test-suite.

### Conclusion

So I looked further and decided that the system-wide credential helper shouldn't be picked up in the first place. This is a feature, but the test-suite should be much more isolated.

There are some ways to achieve this, and I think the most practical one is to assure that `GIT_CONFIG_NOSYSTEM` doesn't pick up the installation-configuration either.

It turns out that [`GIT_CONFIG_NO_SYSTEM` is already set](https://github.com/rust-lang/cargo/blob/2eac6f5c0c0942f907d6e71813920a3c32af46ca/crates/cargo-test-support/src/lib.rs#L1267), which means that `gix-config` needs an adjustment to not include the installation configuration when it is set.

PR #13126, opened 2023-12-06
fix bash completion in directory with spaces

lukas-code:

This PR fixes bash completion for `cargo run --bin <TAB>` when the path to the project contains a space.

1. create a project in a directory with a space, for example `mkdir '/tmp/has space' && cd '/tmp/has space' && cargo new foo && cd foo`.
2. add a binary to Cargo.toml, e.g.
   ```
   [[bin]]
   name = "bar"
   ```
3. on bash, type `cargo run --bin b` and press tab
4. observe error message `bash: $manifest: ambiguous redirect`, command is not completed
5. apply the fix to `$(rustc --print sysroot)/etc/bash_completion.d/cargo`
6. repeat step 3 and observe that the command is completed successfully

### DIFF
--- a/src/etc/cargo.bashcomp.sh
+++ b/src/etc/cargo.bashcomp.sh
@@ -209,12 +209,12 @@ _get_names_from_array()
 				line=${line##*=}
 				line=${line%%\"}
 				line=${line##*\"}
-				names+=($line)
+				names+=("$line")
 			fi
 		fi
 
 		last_line=$line
-	done < $manifest
+	done < "$manifest"
 	echo "${names[@]}"
 }
 


weihanglo:
Looks like we need to ignore those tests… again.

---

Instead of error out, now it completes nothing on my machine.

* macbook M1
* Bash 5.2.15
* bash-completion 2.11

I don't know what could go run.

lukas-code:
> Instead of error out, now it completes nothing on my machine.

I just tested it with bash 5.2.15 + bash-completion 2.11 in a docker image and it seems to work for me.

Did you add a `[[bin]]` target to Cargo.toml? The script currently only works for manually defined `[[bin]]`/`[[test]]`/`[[bench]]` targets and only auto-discovered targets for examples, and this PR does not change that.

weihanglo (approved):
Thanks for the step to review. I don't know how I missed that part.

Now it seems working again :)

PR #13128, opened 2023-12-06
refactor: Pull PackageIdSpec into schema

epage:

This removes one of the remaining ties `util_schemas` has on the rest of cargo as part of #12801.

This is broken up into small commits


src/cargo/util_schemas/mod.rs
```diff
@@ -5,4 +5,5 @@
 //! Any logic for getting final semantics from these will likely need other tools to process, like
 //! `cargo metadata`.
 
+pub mod core;
```
weihanglo, line 8: Although it is just moving things around, `core` is not too meaningful as a name.

src/cargo/core/package_id_spec.rs
```diff
@@ -1,92 +1,30 @@
...
-    /// Roughly equivalent to `PackageIdSpec::parse(spec)?.query(i)`
-    pub fn query_str<I>(spec: &str, i: I) -> CargoResult<PackageId>
+    /// Checks a list of `PackageId`s to find 1 that matches this `PackageIdSpec`. If 0, 2, or
+    /// more are found, then this returns an error.
+    fn query<I>(&self, i: I) -> CargoResult<PackageId>
+    where
+        I: IntoIterator<Item = PackageId>;
+}
+
+impl PackageIdSpecQuery for PackageIdSpec {
```
weihanglo, line 26: I like the name, though naming is hard, and cargo doesn't have a naming convention for this "extension" like trait. Do you have a better idea?

weihanglo, line 26: Should we have a simple doc for this? For example

* [`std::os::unix::ffi::OsStrExt`](https://doc.rust-lang.org/stable/std/os/unix/ffi/trait.OsStrExt.html#)
* [`anyhow::Context`](https://docs.rs/anyhow/1.0.75/anyhow/trait.Context.html)
* [`itertools::Itertools`](https://docs.rs/itertools/0.12.0/itertools/trait.Itertools.html) (not exactly an extension trait)

weihanglo, line 26: Do we want to "seal" this trait?

weihanglo (commented):
Looks good to me. Just some questions that we should deal with later.

src/cargo/util_schemas/mod.rs
epage: Got a better idea?
- I considered `common` and `misc` but figured those were the same or worse
- I considered the root but then that gets messy
- I considered `manifest` but these are general beyond that

src/cargo/core/package_id_spec.rs
epage: Most extension traits have are `<type>Ext` which is weird if there might be multiple (e.g. `ResultExt`) so I tend to use more specific names.

src/cargo/util_schemas/mod.rs
weihanglo: Or remove the entire `core` and expose every type to root level?

epage: Thats the "messy" part.  I feel like it'd more help users reading the docs to not have a large dumping ground in the root of types that are secondary to what they are doing.

PR #13130, opened 2023-12-07
re-enable flaky tests thanks to update to `gix-config`. (#11821)

Byron:
Related to #11821.

With `gix-config` now being fixed, it will properly respect `GIT_CONFIG_NOSYSTEM` both for system-wide configuration as well as for the git installation configuration.

That way, credential helpers provided by the git installation won't be called anymore, which prevents them from 'somehow' emitting information to stderr.

The latter was previously disabled for credential helpers, and despite [everything looking like it should work][1], it simply didn't.

[1]: https://github.com/rust-lang/cargo/pull/13117#issuecomment-1844881287

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1173,9 +1173,9 @@ dependencies = [
 
 [[package]]
 name = "gix-config"
-version = "0.32.0"
+version = "0.32.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ada0e0b904b17a3f2636b70a33e2c8b075b8eb947db80f6c6e94549f2d5b78d1"
+checksum = "0341471d55d8676e98b88e121d7065dfa4c9c5acea4b6d6ecdd2846e85cce0c3"
 dependencies = [
  "bstr",
  "gix-config-value",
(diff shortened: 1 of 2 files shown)


weihanglo (approved):

weihanglo:
I didn't mean to rush you fixing those tests but really appreciate your efforts! Dont be too hard to yourself :)

It seems [libgit2 respects `GIT_CONFIG_NO_SYSTEM`](https://github.com/libgit2/libgit2/blob/d9475611fec95cacec30fbd2c334b270e5265d3b/src/libgit2/repository.c#L1304-L1304) as well so that should be fine.

Let's go again!
@bors r+

Byron:
> I didn't mean to rush you fixing those tests but really appreciate your efforts! Dont be too hard to yourself :)

For the foreseeable future, I will be working on finishing the `cargo` port, which also means I will definitely rush to fix issues so nobody gets to think it's stalling despite sometimes it takes longer to get to the next integration PR as some features take a while.

So I appreciate fixes actually, as they are relatively quick to do.

> It seems [libgit2 respects `GIT_CONFIG_NO_SYSTEM`](https://github.com/libgit2/libgit2/blob/d9475611fec95cacec30fbd2c334b270e5265d3b/src/libgit2/repository.c#L1304-L1304) as well so that should be fine.

This made me aware that my implementation is still incorrect as it doesn't parse the variable content as bool - it seems I never checked the `git` code for that which is definitely not how it should be done.

I will fix this next, even though that one won't need another update here :D.

Byron:
Great minds think alike ? 😁 

https://github.com/Byron/gitoxide/commit/6738955949d82a66c070dcb65e368652898bd1d2#diff-6b512e8fdaec7ec06cf4d49b5440f5ab1157012258c8c85ddba6a426abf375b8L70

PR #13132, opened 2023-12-07
Avoid writing CACHEDIR.TAG if it already exists

arlosi:
Cargo currently unconditionally writes `CACHEDIR.TAG` files even if they already exist. 

This practice causes problems for build systems that disallow multiple writes to the same file.


epage:
Assuming we have no reason to "evolve" this file, I have no problem with this.

@bors r+

PR #13134, opened 2023-12-08
`all-static` feature should include `vendored-libgit2`

arlosi:
Cargo has an `all-static` feature which is supposed to statically link dependencies for distributing.

However, it doesn't include `libgit2`. If the system has a compatible version of `libgit2` and `pkg-config` installed, then the `all-static` build will still use the system `libgit2`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -246,7 +246,7 @@ doc = false
 vendored-openssl = ["openssl/vendored"]
 vendored-libgit2 = ["libgit2-sys/vendored"]
 # This is primarily used by rust-lang/rust distributing cargo the executable.
-all-static = ['vendored-openssl', 'curl/static-curl', 'curl/force-system-lib-on-osx']
+all-static = ['vendored-openssl', 'curl/static-curl', 'curl/force-system-lib-on-osx', 'vendored-libgit2']
 
 [lints]
 workspace = true


Cargo.toml
```diff
@@ -246,7 +246,7 @@ doc = false
 vendored-openssl = ["openssl/vendored"]
 vendored-libgit2 = ["libgit2-sys/vendored"]
 # This is primarily used by rust-lang/rust distributing cargo the executable.
-all-static = ['vendored-openssl', 'curl/static-curl', 'curl/force-system-lib-on-osx']
+all-static = ['vendored-openssl', 'curl/static-curl', 'curl/force-system-lib-on-osx', 'vendored-libgit2']
```
weihanglo, line 249: > If the system has a compatible version of `libgit2` and `pkg-config` installed, then the `all-static` build will still use the system `libgit2`.

Is this true?

It seems to use vendored source when `vendored` feature is activated: https://github.com/rust-lang/git2-rs/blob/8136881b60c6953e8260151bbfc4a79eb4326f72/libgit2-sys/build.rs#L35-L56

arlosi: The `vendored` feature *isn't* activated with `all-static` though. That's the problem.

weihanglo: I misunderstood. Now I see what you meant.

@ehuss, do you know the history why libgit2 wasn't in the list?

weihanglo (approved):

Cargo.toml
ehuss: I suspect it is because nobody asked for it. I think it should be fine to add.

PR #13140, opened 2023-12-08
fix: Print rustc messages colored on wincon

### LINKED ISSUE #13104: Error message colouring broken (on Windows) with 1.74
After upgrading to Rust 1.74.0 some of the message colouring and highlighting disappears when cargo or rustc are used in a Windows 10 console session (Take Command, cmd.exe and Powershell all exhibit the same behaviour):

image

As the above image demonstrates this was working fine with 1.73.0

epage:

The real fix is over in rust-cli/anstyle#150; this just upgrades to it

Fixes #13104

I hacked vt console support off, ran a build before and rustc messages weren't colored.  Now with this change, rustc messages are colored.

### Additional information

This still doesn't address why wincon is being used on their system

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,9 +43,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.4"
+version = "0.6.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
I have no sufficient background to approve this. Given this is a patch release and we eventually will update to that version, let's update to it and see if user satisfies.

StevePitts:
> let's update to it and see if user satisfies.

I noticed this fix in the change log for v1.76 and can confirm, as the original reporter of https://github.com/rust-lang/cargo/issues/13104, that the fix has the desired effect when 'Use legacy console' is in effect for consoles.

PR #13143, opened 2023-12-08
Don't rely on mtime to test changes

### LINKED ISSUE #13139: testsuite: explictly update the mtime when needed
Currently some tests use [`Project::change_file`](https://github.com/rust-lang/cargo/blob/749654c5fdb26e81f543e76806aedde4327c6cd1/crates/cargo-test-support/src/lib.rs#L398-L400) with an empty content string just to update the modified time of the file e.g.:

https://github.com/rust-lang/cargo/blob/bdef274df29a4f521a2c4f59f7ed286f1b720ee0/tests/testsuite/freshness.rs#L558

However, updating a zero length file to be zero length isn't guaranteed to change the modified time (it might be a no-op). Cargo's tests should explicitly set the modified time using the new [`File::set_times`](https://doc.rust-lang.org/beta/std/fs/struct.File.html#method.set_times) API.

There are at least a couple of ways this could be done. An explicit method could be added and used in place of calling `change_file` with an empty string. Or else `change_file` (or the underlying `FileBuilder`) could be updated to always set the modified time.

ChrisDenton:

Fixes #13139 by making sure tests aren't relying on changing the mtime alone 

The pattern to watch out for is when a file is created with empty contents, e.g. `.file("foo", "")`, then later the same file updated with empty contents, e.g. `.change_file("foo", "")`. Tests should be making an actual change to the contents.

### DIFF
--- a/tests/testsuite/build_script.rs
+++ b/tests/testsuite/build_script.rs
@@ -3471,7 +3471,7 @@ fn rebuild_only_on_explicit_paths() {
 
     // random other files do not affect freshness
     println!("run baz");
-    p.change_file("baz", "");
+    p.change_file("baz", "// modified");
     p.cargo("build -v")
         .with_stderr(
             "\
@@ -3483,7 +3483,7 @@ fn rebuild_only_on_explicit_paths() {
 
     // but changing dependent files does
     println!("run foo change");
-    p.change_file("foo", "");
+    p.change_file("foo", "// modified");
     p.cargo("build -v")
         .with_stderr(
             "\
--- a/tests/testsuite/build_script_env.rs
+++ b/tests/testsuite/build_script_env.rs
@@ -96,7 +96,7 @@ fn rerun_if_env_or_file_changes() {
         .with_stderr("[FINISHED] [..]")
         .run();
     sleep_ms(1000);
-    p.change_file("foo", "");
+    p.change_file("foo", "// modified");
     p.cargo("check")
         .env("FOO", "bar")
         .with_stderr(
--- a/tests/testsuite/freshness.rs
+++ b/tests/testsuite/freshness.rs
@@ -1522,7 +1522,7 @@ fn bust_patched_dep() {
         sleep_ms(1000);
     }
 
-    p.change_file("reg1new/src/lib.rs", "");
+    p.change_file("reg1new/src/lib.rs", "// modified");
     if is_coarse_mtime() {
         sleep_ms(1000);
     }


weihanglo:
Seems unrelated, but I wonder what it is 🤔 

```
---- features::no_feature_doesnt_build stdout ----
running `D:\a\cargo\cargo\target\debug\cargo.exe build`
running `D:\a\cargo\cargo\target\tmp\cit\t1334\foo\target\debug\foo.exe`
running `D:\a\cargo\cargo\target\debug\cargo.exe build --features bar -v`
thread 'features::no_feature_doesnt_build' panicked at tests\testsuite\features.rs:525:10:

test failed running `D:\a\cargo\cargo\target\debug\cargo.exe build --features bar -v`
error: process exited with code 101 (expected 0)
--- stdout

--- stderr
   Compiling bar v0.0.1 (D:\a\cargo\cargo\target\tmp\cit\t1334\foo\bar)
     Running `rustc --crate-name bar bar\src\lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no -C debuginfo=2 -C metadata=66d2c9045771087b -C extra-filename=-66d2c9045771087b --out-dir D:\a\cargo\cargo\target\tmp\cit\t1334\foo\target\debug\deps -L dependency=D:\a\cargo\cargo\target\tmp\cit\t1334\foo\target\debug\deps`
   Compiling foo v0.0.1 (D:\a\cargo\cargo\target\tmp\cit\t1334\foo)
     Running `rustc --crate-name foo src\main.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type bin --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 --cfg "feature=\"bar\"" -C metadata=7fe94a9cbff53163 -C extra-filename=-7fe94a9cbff53163 --out-dir D:\a\cargo\cargo\target\tmp\cit\t1334\foo\target\debug\deps -L dependency=D:\a\cargo\cargo\target\tmp\cit\t1334\foo\target\debug\deps --extern bar=D:\a\cargo\cargo\target\tmp\cit\t1334\foo\target\debug\deps\libbar-66d2c9045771087b.rlib`
error: couldn't create a temp dir: Access is denied. (os error 5) at path "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\rustcXRSPPA"

error: could not compile `foo` (bin "foo") due to 1 previous error

Caused by:
  process didn't exit successfully: `rustc --crate-name foo src\main.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type bin --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 --cfg "feature=\"bar\"" -C metadata=7fe94a9cbff53163 -C extra-filename=-7fe94a9cbff53163 --out-dir D:\a\cargo\cargo\target\tmp\cit\t1334\foo\target\debug\deps -L dependency=D:\a\cargo\cargo\target\tmp\cit\t1334\foo\target\debug\deps --extern bar=D:\a\cargo\cargo\target\tmp\cit\t1334\foo\target\debug\deps\libbar-66d2c9045771087b.rlib` (exit code: 1)
  ```

ChrisDenton:
That seems weird but yeah, I'm not seeing how it could be related to this PR unless tests are tripping over each other (which they shouldn't be, obv). But I'm also not seeing it affecting other PRs. Maybe spurious?

An access denied error can occur for a few different reasons, including permissions errors or if the directory exists and a delete is pending. But I'm not sure any apply to a temp dir.

weihanglo (approved):

PR #13152, opened 2023-12-11
Do not allow empty name in package ID spec

hi-rustin:

Do not allow empty names in package ID spec.

See: https://github.com/hi-rustin/cargo-information/pull/63

Check unit tests.

### DIFF
--- a/src/cargo/util_schemas/core/package_id_spec.rs
+++ b/src/cargo/util_schemas/core/package_id_spec.rs
@@ -602,5 +602,6 @@ mod tests {
         .is_err());
         assert!(PackageIdSpec::parse("@1.2.3").is_err());
         assert!(PackageIdSpec::parse("registry+https://github.com").is_ok());
+        assert!(PackageIdSpec::parse("https://crates.io/1foo#1.2.3").is_ok())
     }
 }
--- a/src/cargo/util_schemas/core/package_id_spec.rs
+++ b/src/cargo/util_schemas/core/package_id_spec.rs
@@ -601,5 +601,6 @@ mod tests {
         )
         .is_err());
         assert!(PackageIdSpec::parse("@1.2.3").is_err());
+        assert!(PackageIdSpec::parse("registry+https://github.com").is_ok());
     }
 }
--- a/src/cargo/util_schemas/core/package_id_spec.rs
+++ b/src/cargo/util_schemas/core/package_id_spec.rs
@@ -197,6 +197,10 @@ impl PackageIdSpec {
                 None => (String::from(path_name), None),
             }
         };
+        if name.is_empty() {
+            bail!("package ID specification must have a name: `{url}`");
+        }
+        validate_package_name(name.as_str(), "pkgid", "")?;
         Ok(PackageIdSpec {
             name,
             version,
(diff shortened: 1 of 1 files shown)


src/cargo/core/package_id_spec.rs
```diff
@@ -76,6 +76,9 @@ impl PackageIdSpec {
             Some(version) => Some(version.parse::<PartialVersion>()?),
             None => None,
         };
+        if name.is_empty() {
+            bail!("package ID specification must have a name: `{}`", spec);
```
weihanglo, line 80: nit
```suggestion
            bail!("package ID specification must have a name: `{spec}`");
```

hi-rustin: Thanks!

src/cargo/util_schemas/core/package_id_spec.rs
```diff
@@ -98,6 +98,9 @@ impl PackageIdSpec {
             Some(version) => Some(version.parse::<PartialVersion>()?),
             None => None,
         };
+        if name.is_empty() {
+            bail!("package ID specification must have a name: `{spec}`");
+        }
         validate_package_name(name, "pkgid", "")?;
```
hi-rustin, line 104: Maybe we should put the empty check into this function?

hi-rustin (commented):
🔢 Self-check

src/cargo/util_schemas/core/package_id_spec.rs
epage: That seems like where it should belong but we already have that check be separate for manifests
```rust
    let package_name = package.name.trim();
    if package_name.is_empty() {
        bail!("package name cannot be an empty string")
    }

    validate_package_name(package_name, "package name", "")?;
```

This makes me think some more investigation is needed.


Note: empty feature names were disallowed as of #12928 but that was because we relied on the behavior of our TOML parser to disallow it which wasn't conformant behavior.

hi-rustin: I have checked all the references for this function and it seems okay to include the check in the validate_package_name. However, I have also discovered some bugs in those places. For example, if you attempt to type `cargo add @1.2.3`, you will receive a panic instead of an error. Therefore, I suggest we move forward with this pull request first. Once that's done, I can work on moving it into the validate_package_name and also add tests to cover these problematic cases. This PR is more focused on fixing bugs from package ID spec.

epage: As a heads up, I'm looking at touching the name validation code as well, so it'll be good to coordinate so we don't cause each other a lot of pain

hi-rustin: Got it. I will prioritize ensuring that empty names are not allowed. I will make sure I only change one thing each time and do not bring any off-topic changes. Hope that can help.

hi-rustin:
The CI is broken by https://github.com/rust-lang/rust/commit/f481596ee4e761182f20eaf1b67d4431cd2d5556

hi-rustin:
@epage Could you please merge it again? Thanks!

PR #6276, opened 2018-11-07
Warn/err on some unused manifest keys in workspaces.

ehuss:
- Err on dependencies, features, target, badges in virtual root.
- Warn on replace/patch in members.

Closes #6248, Closes #4870, Closes #4947

PR #13155, opened 2023-12-11
fix(toml)!: Disallow `[lints]` in virtual workspaces

epage:
This was missed with the initial `[lints]` implementation.

While this is a breaking change, this is aligned with ones we've done in the past.  A lot of times, we warn first.  My hope is that isn't needed this time because
- It only exists virtual workspaces so they aren't published
- It is a nop to have this which is likely to be caught
- This is so new that the number of people using it, and likely running into this case, is quite low.

### DIFF
--- a/src/cargo/util_schemas/manifest.rs
+++ b/src/cargo/util_schemas/manifest.rs
@@ -45,6 +45,7 @@ pub struct TomlManifest {
     pub workspace: Option<TomlWorkspace>,
     pub badges: Option<InheritableBtreeMap>,
     pub lints: Option<InheritableLints>,
+    // when adding new fields, be sure to check whether `to_virtual_manifest` should disallow them
 }
 
 impl TomlManifest {
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -1122,6 +1122,9 @@ fn to_virtual_manifest(
     if me.badges.is_some() {
         bail!("this virtual manifest specifies a [badges] section, which is not allowed");
     }
+    if me.lints.is_some() {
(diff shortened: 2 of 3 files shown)


src/cargo/util_schemas/manifest.rs
```diff
@@ -45,6 +45,7 @@ pub struct TomlManifest {
     pub workspace: Option<TomlWorkspace>,
     pub badges: Option<InheritableBtreeMap>,
     pub lints: Option<InheritableLints>,
+    // when adding new fields, be sure to check whether `to_virtual_manifest` should disallow them
```
weihanglo, line 48: Wonder we have a way to catch this automatically.

weihanglo (approved):
Agree that we make this a hard error asap. The old PR did it as well <https://github.com/rust-lang/cargo/pull/6276>.

weihanglo:
What does `!` mean in `fix(toml)!:`?

epage:
> What does `!` mean in `fix(toml)!:`?

Breaking change

jdahlstrom:
I'm probably missing something, but what's the rationale for this? Surely it may be useful to also harmonize lints in a virtual workspace?

weihanglo:
> I'm probably missing something, but what's the rationale for this? Surely it may be useful to also harmonize lints in a virtual workspace?

Then one should use `[workspace.lints]` instead of `[lints]`. This PR prohibits the latter.

GeeWee:
Ah, that distinction also confused me. I think it might be worth emphasizing it in the release notes if they're not immutable :)

epage:
I posted #13425

PR #6771, opened 2019-03-21
Always nicely show errors from crates.io if possible

sgrif:
Currently if Cargo ever gets a non-200 response, it will either not show
the error at all (if it was a 403 or 404), or spit out the entire
response body. Historically crates.io has served a 200 for most errors
to work around this, but we've stopped doing this as it causes problems
for other clients.

Additionally, we're starting to server more errors that have semantic
meaning (429 for rate limiting, 503 when we're in read only mode). If
the request specifies "Accept: application/json", we should ideally
return the errors formatted nicely. This isn't always true, but it's
what we'd like to do going forward.

While the output that Cargo puts out at least contains the actual
message, it's buried under a ton of useless info. This changes the
behavior so that if the response was valid JSON in the format that Cargo
expects, it just shows that (along with a description of the response
status), and only falls back to spitting out everything if it can't
parse the response body.

I'd love to add some more tests for this, but I've had trouble finding
anywhere in the test suite that exercises these paths.

Issue #3995, opened 2017-05-04
Handle non-200 response codes from registry nicely


Nemo157:
Currently crates.io always returns a 200 status code unless cargo's token is invalid/nonexistent (or there is an internal error), any errors for cargo are then returned in a json object (e.g. validation of the Cargo.toml fields during a publish).

I attempted to change crates.io to return a 400 status code, but cargo does not handle these nicely

```
→ cargo publish --allow-dirty --host http://localhost:8888/git/index
    Updating registry `http://localhost:8888/git/index`
warning: manifest has no description, license, license-file, documentation, homepage or repository.
See http://doc.crates.io/manifest.html#package-metadata for more info.
   Packaging blah v0.1.0 (file:///Users/Nemo157/sources/blah)
   Verifying blah v0.1.0 (file:///Users/Nemo157/sources/blah)
   Compiling blah v0.1.0 (file:///Users/Nemo157/sources/blah/target/package/blah-0.1.0)
    Finished dev [unoptimized + debuginfo] target(s) in 0.12 secs
   Uploading blah v0.1.0 (file:///Users/Nemo157/sources/blah)
error: failed to get a 200 OK response, got 400
headers:
    HTTP/1.1 400 Bad Request

    Set-Cookie: cargo_session=--M9k+4zBMnWwFxkGyd/n7KZ2H/dU=; HttpOnly; Path=/

    Content-Type: application/json; charset=utf-8

    Content-Length: 171



body:
{"errors":[{"detail":"missing or empty metadata fields: description, license. Please see http://doc.crates.io/manifest.html#package-metadata for how to upload metadata"}]}
```

I have opened a ticket at https://github.com/rust-lang/crates.io/issues/712 as well, it's probably worth discussing whether this is a change that is wanted at all there, before discussing any kind of implementation details here.

alexcrichton:
Yeah definitely! Cargo's handling of responses from the registry is... not the best unfortunately. We should at the very least try to extract the `errors` coming back and package that up in a nice fashion to send to the user.

If you'd be interested in implementing this I'd be more than willing to mentor!

ehuss:
This has been updated in #6771 to always show the JSON errors in a nicer way, regardless of the error code.  I'm going to close this for now, as I think the spirit of the issue has been addressed. If there are ways to further clarify the error messages, feel free to open a new issue.

PR #13158, opened 2023-12-12
crates-io: Add support for other 2xx HTTP status codes

Turbo87:
Replying with `201 Created` of `202 Accepted` should not result in showing errors.

Related:

- https://github.com/rust-lang/cargo/issues/3995
- https://github.com/rust-lang/cargo/pull/6771

### DIFF
--- a/crates/crates-io/lib.rs
+++ b/crates/crates-io/lib.rs
@@ -436,7 +436,8 @@ impl Registry {
             .map(|s| s.errors.into_iter().map(|s| s.detail).collect::<Vec<_>>());
 
         match (self.handle.response_code()?, errors) {
-            (0, None) | (200, None) => Ok(body),
+            (0, None) => Ok(body),
+            (code, None) if is_success(code) => Ok(body),
             (code, Some(errors)) => Err(Error::Api {
                 code,
                 headers,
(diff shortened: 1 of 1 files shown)


ehuss:
I'm a little concerned about this change. We currently document that only 200 is a valid return code. Can we follow up with updating the documentation to make that clearer?

Also, just to verify, are all 2xx responses really success? I see some of them are kinda weird, like 204 No Content. I suppose that one is fine since cargo seems to mostly ignore the response. I'm just looking at wikipedia, and I'm not sure if there are certain kinds of responses that aren't really success.

Turbo87:
> Also, just to verify, are all 2xx responses really success?

that is my understanding and https://docs.rs/http/latest/src/http/status.rs.html#177-181 seems to confirm it

> We currently document that only 200 is a valid return code. Can we follow up with updating the documentation to make that clearer?

sure! can you link me to the docs that you're referring to? :)

epage:
See https://doc.rust-lang.org/cargo/reference/registry-web-api.html#web-api

weihanglo:
> Also, just to verify, are all 2xx responses really success?

Yes. See <https://datatracker.ietf.org/doc/html/rfc7231#section-6.3> and <https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses>.

PR #13164, opened 2023-12-13
fix: Fill in more empty name holes

epage:

This is a follow up to #13152 and expands on work done in #12928.

This is also part of #12801 as I am refactoring how we do validation so that it parts of it can move into the schema, removing the last dependency the schema has on the rest of cargo.

This prevents empty registry names which should be fine for the same reason as preventing empty feature names in #12928, that this was a regression due to changes in toml parsers


weihanglo (approved):

weihanglo:
Thank you.

@bors r+

Issue #12740, opened 2023-09-26
Cargo and rustc are using different tones of red for errors on the terminal


RalfJung:
### Problem

Cargo and rustc are not consistent in terms of how they are coloring errors on the terminal.

### Steps

1. Run `cargo check` on a project with build failures.

![image](https://github.com/rust-lang/cargo/assets/330628/4f90d615-9244-4a66-8788-addcdb842f23)

As you can see, the two "error" are using two different shades of red. In terms of the `colorize` crate, rustc is using `bright_red` and cargo is using `red`.

### Possible Solution(s)

One of cargo or rustc should switch to the other tone of red. I have a slight preference for "bright red", that's why I reported this as a cargo issue.

### Version

```text
$ cargo version --verbose
cargo 1.72.0 (103a7ff2e 2023-08-15)
release: 1.72.0
commit-hash: 103a7ff2ee7678d34f34d778614c5eb2525ae9de
commit-date: 2023-08-15
host: x86_64-unknown-linux-gnu
libgit2: 1.6.4 (sys:0.17.2 vendored)
libcurl: 8.1.2-DEV (sys:0.4.63+curl-8.1.2 vendored ssl:OpenSSL/1.1.1u)
ssl: OpenSSL 1.1.1u  30 May 2023
os: Debian n/a (trixie) [64-bit]
```

weihanglo:
Thanks for the report!

We're aware of this and have <https://github.com/rust-lang/cargo/issues/12627> for tracking one of the approach. See also <https://github.com/rust-lang/cargo/issues/2290#issuecomment-1728318644> for a short summary.

weihanglo:
Without investigating it by myself, do you know since which version this happened, or it has been there for a while?

RalfJung:
I have no idea. I only just realized this when working on the messages for ui_test. It might have been like this for a while, it might have been hidden by me using a different terminal/configuration, it might be recent.

RalfJung:
I went back through time a bit -- this is definitely a very long-standing issue. I see the difference even with Rust 1.40.

epage:
Glad to see this wasn't something I broke recently as I've been making changes to how we do styling :)

#12578 added styling to clap and #12655 made us use shared definitions to ensure we'd stay in sync which also made our style choices easier to audit.

Cargo uses bold+red and a lot of has likely haven't noticed the discrepancy because bold+red gets colored the same as bright_red in a lot of themes.

epage:
[...]

PR #13172, opened 2023-12-14
feat: Add `rustc` style errors for manifest parsing

Muscraft:
#12235 is tracking user control over warnings. One part of that is making `cargo`'s diagnostic system output messages in the style of `rustc`. To make it so that `cargo` doesn't have to manage a formatter and renderer, I decided to use [`annotate-snippets`](https://crates.io/crates/annotate-snippets), which matches `rustc`'s style well (at one time it was meant to be the formatted for `rustc`).

To get this work kicked off, it was suggested to me that we should start with styling manifest parsing errors, and that is what his PR does.

What manifest parsing errors look like after this change:
image

---

Note: `cargo` does not currently match `rustc` in color (#12740), `rustc`  specifies their colors [here](https://github.com/rust-lang/rust/blob/740cea81d6e34fc03c4495890e29f1c5ecb40b96/compiler/rustc_errors/src/lib.rs#L1755-L1768) and [here](https://github.com/rust-lang/rust/blob/740cea81d6e34fc03c4495890e29f1c5ecb40b96/compiler/rustc_errors/src/emitter.rs#L2689-L2723). I used `annotate-snippets` default colors which match what `rustc` currently uses for colors. I did this as a stopgap while I work to unify the colors used between `rustc` and `cargo`.

---

Note: the error messages come directly from `toml` and can be quite long. It would be nice if we could put some of the message below the highlight but this would require changes to `toml`.
Example:
```
error: invalid type: integer `3`
 --> Cargo.toml:7:7
  |
7 | bar = 3
  |       ^ expected a version string like "0.9.8" or a detailed dependency like { version = "0.9.8" }
  |
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ homepage = "https://github.com/rust-lang/cargo"
 repository = "https://github.com/rust-lang/cargo"
 
 [workspace.dependencies]
+annotate-snippets = "0.10.1"
 anstream = "0.6.5"
 anstyle = "1.0.4"
 anyhow = "1.0.79"
@@ -139,6 +140,7 @@ name = "cargo"
 path = "src/cargo/lib.rs"
 
 [dependencies]
+annotate-snippets.workspace = true
 anstream.workspace = true
 anstyle.workspace = true
 anyhow.workspace = true
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -1,9 +1,11 @@
+use annotate_snippets::{Annotation, AnnotationType, Renderer, Slice, Snippet, SourceAnnotation};
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
 use std::str::{self, FromStr};
 
+use crate::AlreadyPrintedError;
 use anyhow::{anyhow, bail, Context as _};
 use cargo_platform::Platform;
 use cargo_util::paths;
@@ -11,6 +13,7 @@ use cargo_util_schemas::manifest;
 use cargo_util_schemas::manifest::RustVersion;
 use itertools::Itertools;
 use lazycell::LazyCell;
+use pathdiff::diff_paths;
 use tracing::{debug, trace};
 use url::Url;
 
@@ -43,7 +46,7 @@ pub fn read_manifest(
     path: &Path,
     source_id: SourceId,
     config: &Config,
-) -> Result<(EitherManifest, Vec<PathBuf>), ManifestError> {
+) -> CargoResult<(EitherManifest, Vec<PathBuf>)> {
     trace!(
         "read_manifest; path={}; source-id={}",
         path.display(),
@@ -56,15 +59,23 @@ pub fn read_manifest(
             return Err(ManifestError::new(
                 anyhow::anyhow!("parsing `{}` requires `-Zscript`", path.display()),
                 path.into(),
-            ));
+            ))?;
         }
         contents = embedded::expand_manifest(&contents, path, config)
             .map_err(|err| ManifestError::new(err, path.into()))?;
     }
 
-    read_manifest_from_str(&contents, path, embedded, source_id, config)
-        .with_context(|| format!("failed to parse manifest at `{}`", path.display()))
-        .map_err(|err| ManifestError::new(err, path.into()))
+    read_manifest_from_str(&contents, path, embedded, source_id, config).map_err(|err| {
+        if err.is::<AlreadyPrintedError>() {
+            err
+        } else {
+            ManifestError::new(
+                err.context(format!("failed to parse manifest at `{}`", path.display())),
+                path.into(),
+            )
+            .into()
+        }
+    })
 }
 
 /// See also `bin/cargo/commands/run.rs`s `is_manifest_command`
@@ -94,11 +105,61 @@ fn read_manifest_from_str(
 
     let mut unused = BTreeSet::new();
     let deserializer = toml::de::Deserializer::new(contents);
-    let manifest: manifest::TomlManifest = serde_ignored::deserialize(deserializer, |path| {
+    let manifest: manifest::TomlManifest = match serde_ignored::deserialize(deserializer, |path| {
         let mut key = String::new();
         stringify(&mut key, &path);
         unused.insert(key);
-    })?;
+    }) {
+        Ok(manifest) => manifest,
+        Err(e) => {
+            let Some(span) = e.span() else {
+                return Err(e.into());
+            };
+
+            let (line_num, column) = translate_position(&contents, span.start);
+            let source_start = contents[0..span.start]
+                .rfind('\n')
+                .map(|s| s + 1)
+                .unwrap_or(0);
+            let source_end = contents[span.end - 1..]
+                .find('\n')
+                .map(|s| s + span.end)
+                .unwrap_or(contents.len());
+            let source = &contents[source_start..source_end];
+            // Make sure we don't try to highlight past the end of the line,
+            // but also make sure we are highlighting at least one character
+            let highlight_end = (column + contents[span].chars().count())
+                .min(source.len())
+                .max(column + 1);
+            // Get the path to the manifest, relative to the cwd
+            let manifest_path = diff_paths(manifest_file, config.cwd())
+                .unwrap_or_else(|| manifest_file.to_path_buf())
+                .display()
+                .to_string();
+            let snippet = Snippet {
+                title: Some(Annotation {
+                    id: None,
+                    label: Some(e.message()),
+                    annotation_type: AnnotationType::Error,
+                }),
+                footer: vec![],
+                slices: vec![Slice {
+                    source: &source,
+                    line_start: line_num + 1,
+                    origin: Some(manifest_path.as_str()),
+                    annotations: vec![SourceAnnotation {
+                        range: (column, highlight_end),
+                        label: "",
+                        annotation_type: AnnotationType::Error,
+                    }],
+                    fold: false,
+                }],
+            };
+            let renderer = Renderer::styled();
+            writeln!(config.shell().err(), "{}", renderer.render(snippet))?;
+            return Err(AlreadyPrintedError::new(e.into()).into());
+        }
+    };
     let add_unused = |warnings: &mut Warnings| {
         for key in unused {
             warnings.add_warning(format!("unused manifest key: {}", key));
@@ -2112,4 +2173,34 @@ impl ResolveToPath for ConfigRelativePath {
     fn resolve(&self, c: &Config) -> PathBuf {
         self.resolve_path(c)
     }
+}
+
+fn translate_position(input: &str, index: usize) -> (usize, usize) {
+    if input.is_empty() {
+        return (0, index);
+    }
+
+    let safe_index = index.min(input.len() - 1);
+    let column_offset = index - safe_index;
+
+    let nl = input[0..safe_index]
+        .as_bytes()
+        .iter()
+        .rev()
+        .enumerate()
+        .find(|(_, b)| **b == b'\n')
+        .map(|(nl, _)| safe_index - nl - 1);
+    let line_start = match nl {
+        Some(nl) => nl + 1,
+        None => 0,
+    };
+    let line = input[0..line_start]
+        .as_bytes()
+        .iter()
+        .filter(|c| **c == b'\n')
+        .count();
+    let column = input[line_start..=safe_index].chars().count() - 1;
+    let column = column + column_offset;
+
+    (line, column)
 }
--- a/tests/testsuite/alt_registry.rs
+++ b/tests/testsuite/alt_registry.rs
@@ -712,16 +712,17 @@ fn bad_registry_name() {
         .with_status(101)
         .with_stderr(
             "\
-[ERROR] failed to parse manifest at `[CWD]/Cargo.toml`
-
-Caused by:
-  TOML parse error at line 7, column 17
-    |
-  7 |                 [dependencies.bar]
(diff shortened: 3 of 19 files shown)


src/cargo/util/toml/mod.rs
```diff
@@ -97,11 +109,69 @@ fn read_manifest_from_str(
...
+                    .warning(WARN)
+                    .info(NOTE)
+                    .note(NOTE)
+                    .help(NOTE)
+                    .line_no(NOTE);
+                writeln!(config.shell().err(), "{}", renderer.render(snippet))?;
+                Err(AlreadyPrintedError::new(e.into()).into())
+            } else {
+                Err(e.into())
+            };
```
epage, line 172: Try to make structure match the business logic.  In this case, the indentation is for guard statements and early return / let-else are tools that should instead be used
```rust
let Some(span) = e.span() else {
    return Err(e.into());
};
```

src/cargo/util/toml/mod.rs
```diff
@@ -97,11 +109,69 @@ fn read_manifest_from_str(
...
+                let line_count = lines.clone().count();
+                let mut source = lines.skip(line).next().unwrap_or_default().to_string();
+                // Add back the new line that was stripped by `.lines()`
+                if line_count > line + 1 {
+                    source.push('\n');
+                // If we are trying to highlight past the end of the file, add a
+                // placeholder character to the end of the line.
+                } else if span.end >= source.len() - 1 {
+                    source.push('∅');
+                }
```
epage, line 131: 1. This should be in annotate-snippets,  not cargo
2. The `'\n'` should be `"\\n"` so it gets rendered
3. These should be Dimmed to indicate they aren't literal text

src/cargo/util/toml/mod.rs
```diff
@@ -2161,3 +2231,33 @@ impl ResolveToPath for ConfigRelativePath {
...
+        .find(|(_, b)| **b == b'\n')
+        .map(|(nl, _)| index - nl - 1);
+    let line_start = match nl {
+        Some(nl) => nl + 1,
+        None => 0,
+    };
+    let line = input[0..line_start].iter().filter(|b| **b == b'\n').count();
+    let line = line;
+
+    let column = std::str::from_utf8(&input[line_start..=index])
```
epage, line 2257: Why are we moving from `&str` -> `&[u8]` -> `&str`?

Cargo.toml
```diff
@@ -16,6 +16,7 @@ edition = "2021"
 license = "MIT OR Apache-2.0"
 
 [workspace.dependencies]
+annotate-snippets = "0.10.0"
```
weihanglo, line 19: Apparently `rustfmt`, `miri`, and `rustc_errors` haven't caught up `annotate-snippets@0.10.0`. Read through its [CHANGELOG.md](https://github.com/rust-lang/annotate-snippets-rs/blob/master/CHANGELOG.md#0100---december-12-2023), I assume there is nothing changing the rendered result.

So, just want to make sure that we don't need to update them all at once to be consistent in style?

src/cargo/util/toml/mod.rs
ehuss: I'm also not clear why the `∅` is being added, can you say more about the motivation there? I'm a little concerned about any potential confusion if users are not familiar with what that symbol means. I don't think rustc typically displays errors like that, and I think it is reasonably clear without a special character. For example:

```
error: this file contains an unclosed delimiter
 --> src/main.rs:4:10
  |
4 | fn x() {
  |        - ^
  |        |
  |        unclosed delimiter
```

or

```
error: expected one of `->`, `where`, or `{`, found `<eof>`
 --> src/main.rs:4:9
  |
4 | fn main()
  |         ^ expected one of `->`, `where`, or `{`
```

ehuss:
This PR probably isn't the best place to have a broader discussion, but I was wondering about what the longer-term higher-level design approach will be with this. Some questions I have:

* Is the error-chain approach cargo currently uses incompatible with the annotated diagnostics? Is there some way the two can coexist?
* Is it possible to have an annotated diagnostic encoded in a type that impls `Error` so that it can be returned like a normal error, and have cargo's top-level error handler (`display_error`) display the diagnostics correctly? 
    * I'm a little concerned about what looks like a special-cased `AlreadyPrintedError`.
* If there is an annotated structured `Error` type, is it possible to also have that work with JSON output (#8283).
* If it isn't possible to have an annotated `Error` type, would it make sense to have something like rustc's `emit_err` and keep track of whether or not errors have been encountered in a centralized place?

I'm happy to have a sync discussion sometime if that would be a better way to discuss the design ideas.

epage:
For myself, I see this PR as picking the smallest subsection of cargo to allow us to start introducing rustc-style diagnostics and most larger discussions (how to spread this pattern to other parts of the code, json output) would be handled in incremental steps on top.  I'm willing to accept some level of hacks (`AlreadyPrintedError`) to help keep the scope of this experiment small.

The rough roadmap I see is
1. Render TOML errors like rustc
2. Try to turn all `toml/mod.rs` errors into rustc-style errors
3. Create a new warning (implicit features) that leverages a new `emit_diagnostic` (intentionally unstable feature so its not shown to users while we work out the kinks)
4. Identify `shell().warn` candidates to migrate to `emit_diagnostic`
5. Stabilize `[lints.cargo]` (doesn't have to block on all warnings being turned into diagnostics but we'll likely want some)


At this point, depending on priorities, we can better evaluate
- `emit_err`: after `emit_diagnostic` as that will need to handle `deny` and isn't as critical to normal operations.  We'll need to discover how we want to handle "exit immediate" vs "fail on exit" (ie error recovery) in both cases.
- json output as this isn't in the critical path towards the goal we're working towards but having more structured information will help!

src/cargo/util/toml/mod.rs
epage: I'm trying to find the conversation around invisible characters but it is a bit buried.

I *think* an aspect is that TOML is newline sensitive, unlike Rust, which elevates the importance of invisible characters and makes the chance of an error more frequent.

Cargo.toml
Muscraft: We shouldn't need to update them all at once as there is no change to how things are actually rendered

src/cargo/util/toml/mod.rs
Muscraft: I removed `cargo`'s ability to render invisible characters, as it is something that should be discussed in another issue

tests/testsuite/bad_config.rs
```diff
@@ -1685,16 +1675,14 @@ fn bad_trim_paths() {
...
-    |
-  7 |                 trim-paths = "split-debuginfo"
-    |                              ^^^^^^^^^^^^^^^^^
-  expected a boolean, "none", "diagnostics", "macro", "object", "all", or an array with these options
-"#,
+        .with_stderr("\
+[ERROR] expected a boolean, \"none\", \"diagnostics\", \"macro\", \"object\", \"all\", or an array with these options
+ --> Cargo.toml:7:30
+  |
+7 |                 trim-paths = \"split-debuginfo\"
```
epage, line 1682: Why is it escaping the quote?

Huh, seems the old and new do it inconstistently?

tests/testsuite/alt_registry.rs
```diff
@@ -712,16 +712,14 @@ fn bad_registry_name() {
...
-  7 |                 [dependencies.bar]
-    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-  invalid character ` ` in registry name: `bad name`, [..]
+[ERROR] invalid character ` ` in registry name: `bad name`, characters must be Unicode XID characters (numbers, `-`, `_`, or most letters)
 
 
+ --> Cargo.toml:7:17
+  |
+7 |                 [dependencies.bar]
+  |_________________^
```
epage, line 721: Whats with the leading underscore?

Muscraft: Looks like I forgot to add my change for multi-line highlights. I pushed a fix.

tests/testsuite/bad_config.rs
Muscraft: The previous test used `r#"` and I switched it to `"`, as I find structuring the output the "same" as it would be emitted is clearer. In the actual output displayed to a user, the escaping won't be there.

Cargo.toml
epage: Weren't colors tweaked?

epage: But I agree that we don't need to change them as we just made the cargo diagnostics match rustc's diagnostics.  The others can catch up when they can

Muscraft: Whoops, you are correct, the colors did change. I am not as worried about that change being made in sync across the ecosystem.

src/cargo/util/toml/mod.rs
```diff
@@ -2154,3 +2225,28 @@ impl ResolveToPath for ConfigRelativePath {
...
+        .chars()
+        .rev()
+        .enumerate()
+        .find(|(_, c)| c == &'\n')
+        .map(|(nl, _)| safe_index - nl - 1);
+    let line_start = match nl {
+        Some(nl) => nl + 1,
+        None => 0,
+    };
+    let line = input[0..line_start].chars().filter(|c| c == &'\n').count();
```
epage, line 2247: nit: If you can't use something like `.lines()` then I'd recommend doing `as_bytes()` instead of `chars()` as newline counting can work on the byte level and `chars()` is slow.

src/cargo/util/toml/mod.rs
```diff
@@ -94,11 +105,71 @@ fn read_manifest_from_str(
...
+        Err(e) => {
+            let Some(span) = e.span() else {
+                return Err(e.into());
+            };
+
+            let (line_num, column) = translate_position(contents, span.start);
+            let num_new_lines = contents[span.start..span.end]
+                .chars()
+                .filter(|c| c == &'\n')
+                .count()
```
epage, line 123: nit: ditto about `as_bytes()` instead of `chars()`

src/cargo/util/toml/mod.rs
```diff
@@ -94,11 +105,71 @@ fn read_manifest_from_str(
...
+                - (contents[span.end - 1..span.end].ends_with('\n') as usize);
```
epage, line 124: I feel like this math is non-obvious enough that something should be clarified

src/cargo/util/toml/mod.rs
```diff
@@ -94,11 +105,71 @@ fn read_manifest_from_str(
...
+                    // Add back the new line that was stripped by `.lines()
+                    if num_new_lines > i {
+                        line.push('\n');
+                    }
+                    line
+                })
+                .collect();
+            // Make sure we don't try to highlight past the end of the line,
+            // but also make sure we are highlighting at least one character
+            let highlight_end = (column + (span.end - span.start))
```
epage, line 141: bug: `column` was calculated with `chars()` but the `span` math is using bytes

src/cargo/util/toml/mod.rs
```diff
@@ -94,11 +105,71 @@ fn read_manifest_from_str(
...
+                .enumerate()
+                .map(|(i, line)| {
+                    let mut line = line.to_string();
+                    // Add back the new line that was stripped by `.lines()
+                    if num_new_lines > i {
+                        line.push('\n');
+                    }
+                    line
+                })
+                .collect();
```
epage, line 138: Can't you just:
```rust
let source_start = contents[0..span.start].rfind('\n').map(|s| s + 1).unwrap_or(0);
let source_end = contents[span.end..].find('\n').unwrap_or(contents.len());
let source = &contents[source_start..source_end];
```

src/cargo/util/toml/mod.rs
```diff
@@ -94,11 +105,71 @@ fn read_manifest_from_str(
...
+            // Make sure we don't try to highlight past the end of the line,
+            // but also make sure we are highlighting at least one character
+            let highlight_end = (column + (span.end - span.start))
+                .min(source.len())
+                .max(column + 1);
+            // Get the path to the manifest, relative to the cwd
+            let manifest_path = diff_paths(manifest_file, config.cwd())
+                .unwrap_or(manifest_file.to_path_buf())
```
epage, line 146: nit: `unwrap_or_else` to only do the allocation when needed

src/cargo/util/toml/mod.rs
epage: imo `annotate-snippets` should be taking a byte span

src/cargo/util/toml/mod.rs
epage: Oops, I had a bug.  `source_end` needs `span.end` added to it.

src/cargo/util/toml/mod.rs
Muscraft: `rustc` uses char positions when rendering their diagnostics. They have a method of getting a `CharPos` from a byte. For consistency, we should probably do the same.

epage: From an API design perspective, are we intending the caller to describe the visual range to be highlighted of the slice within the content that should be highlighted?

While this might require a breaking change, imo the caller shouldn't be concerned with the exact rendering policy like that and it should be left up to `annotate-snippets`.

src/cargo/util/toml/mod.rs
Muscraft: Thanks for this it works much better

src/cargo/util/toml/mod.rs
Muscraft: I will open an issue for this in `annotate-snippets` to switch to bytes long-term. In the meantime, it shouldn't cause too many problems within `cargo`.

epage: My immediate concern with this code is that we are mixing units: `column` is `char` based but the span math is `u8` based.

Whatever way we go long term, we should at least pass in the data for the units `annotate-snippets` wants today and not mixed units.

Muscraft: The most recent change makes it so we are not mixing units

ehuss:
@bors retry

macos never started

PR #13178, opened 2023-12-15
refactor(schemas): Pull out as `cargo-util-schemas`

### LINKED ISSUE #12801: Official API for parsing `Cargo.toml`s schema

epage:

Fixes #12801

See the individual commits for further justifications on the changes


crates/cargo-util-schemas/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "cargo-util-schemas"
+version = "0.1.0"
+rust-version.workspace = true
+edition.workspace = true
+license.workspace = true
+homepage = "https://github.com/rust-lang/cargo"
```
hi-rustin, line 7: Off-topic question: Why don't we put the homepage and repository in the `workspace.package`?

hi-rustin (approved):
Looks good to me. Thanks! 👍

But it seems the third commit message contains a typo. I guess it should be `refactor(sources): Don't impl on schema types`.

src/cargo/sources/git/utils.rs
```diff
@@ -207,56 +207,54 @@ impl GitDatabase {
...
-            GitReference::Branch(s) => {
-                let name = format!("origin/{}", s);
-                let b = repo
-                    .find_branch(&name, git2::BranchType::Remote)
-                    .with_context(|| format!("failed to find branch `{}`", s))?;
-                b.get()
-                    .target()
-                    .ok_or_else(|| anyhow::format_err!("branch `{}` did not have a target", s))?
-            }
+/// Resolves self to an object ID with objects the `repo` currently has.
```
weihanglo, line 216:
```suggestion
/// Resolves an [`GitReference`] to an object ID with objects the `repo` currently has.
```

Though personally I prefer an extension trait for grouping these methods.

weihanglo:
> We can always rename and I expect we will when packages-as-namespaces is stabilized

Good point. Naming is a minor issue that might not worth blocking this. I was just trying to avoid anchoring bias. I mean once this is merged, people might be less likely to have opinions on minor details. @epage, do you suggest cancelling this FCP and instead I just r+ this? Totally okay with that.

src/cargo/sources/git/utils.rs
epage: For `PackageIdSpec`, the type is widely used and so having a single trait to pull in and access the functionality seems reasonable.  For this, its pretty limited in where its used (with one exception) which makes me lean the other way.  Or to put this another way, if I were writing this code from scratch, this is likely how I would have done it.

weihanglo (approved):

weihanglo:
People generally agree on the idea so going to merge this now.

@bors r+

PR #13183, opened 2023-12-18
test(priv_dep): add test for  `verify public is respected recursively`

linyihai:
This PR want to verify `public is respected recursively` by rustc.  
The original idea came from here: https://github.com/rust-lang/rust/issues/44663#issuecomment-1851227083.

you can review and test on per commit.

### Additional information
test case wait for add:
- [x] Add test case from https://github.com/rust-lang/rust/issues/119428

### DIFF
--- a/crates/cargo-test-support/src/registry.rs
+++ b/crates/cargo-test-support/src/registry.rs
@@ -558,6 +558,7 @@ pub struct Dependency {
     package: Option<String>,
     optional: bool,
     default_features: bool,
+    public: bool,
 }
 
 /// Entry with data that corresponds to [`tar::EntryType`].
@@ -1428,6 +1429,7 @@ impl Package {
                     "kind": dep.kind,
                     "registry": registry_url,
                     "package": dep.package,
+                    "public": dep.public
                 })
             })
             .collect::<Vec<_>>();
@@ -1678,6 +1680,7 @@ impl Dependency {
             optional: false,
             registry: None,
             default_features: true,
+            public: false,
         }
     }
 
@@ -1728,6 +1731,12 @@ impl Dependency {
     /// Changes this to an optional dependency.
     pub fn optional(&mut self, optional: bool) -> &mut Self {
         self.optional = optional;
+        self
+    }
+
+    /// Changes this to an public dependency.
+    pub fn public(&mut self, public: bool) -> &mut Self {
+        self.public = public;
         self
     }
 
--- a/tests/testsuite/pub_priv.rs
+++ b/tests/testsuite/pub_priv.rs
@@ -1,7 +1,7 @@
 //! Tests for public/private dependencies.
 
 use cargo_test_support::project;
-use cargo_test_support::registry::Package;
+use cargo_test_support::registry::{Dependency, Package};
 
 #[cargo_test(nightly, reason = "exported_private_dependencies lint is unstable")]
 fn exported_priv_warning() {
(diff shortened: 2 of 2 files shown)


tests/testsuite/pub_priv.rs
```diff
@@ -479,3 +479,63 @@ fn allow_priv_in_custom_build() {
...
+            pub use self::pub_bar::*;
+        ",
+        )
+        .file(
+            "src/grandparent_bar/parent_bar/pub_bar/mod.rs",
+            "
+            extern crate priv_dep;
+            pub fn use_priv(_: priv_dep::FromPriv) {}
+        ",
+        )
```
epage, line 530: This seems to be showing how things work with children modules, rather than children packages.

We also need to be focusing on verifying that the recursively public API does not cause warnings.  One caution about it is that rustc does not correctly lint for re-exports.

linyihai: > One caution about it is that rustc does not correctly lint for re-exports.

`recursive_package_pub_no_warning` and `recursive_module_pub_no_warning`  had added newly,  I thinks this two case can be prove `rustc` handle the lint correctly.

linyihai: > One caution about it is that rustc does not correctly lint for re-exports.

`recursive_package_pub_no_warning` and `recursive_package_priv_warning` had added newly.

And if my testcase write correctly, the `recursive_package_pub_no_warning` warning by rustc is definitely wrong.

linyihai: `recursive_package_priv_warning` is similar to this issue: https://github.com/rust-lang/cargo/issues/13193

epage:
Thanks for doing this investigation!

Any thoughts on what test cases are worth merging now vs waiting on #13193?

linyihai:
> Thanks for doing this investigation!
> 
> Any thoughts on what test cases are worth merging now vs waiting on #13193?

I think both test cases make sense.  IMO,  the test case `recursive_package_pub_no_warning` verifies `the recursively public API does not cause warnings`.  `recursive_package_priv_warning`  It's also a use case and a reminder of our current processing logic. It's ok to be updated `recursive_package_priv_warning` when this issue https://github.com/rust-lang/cargo/issues/13193 has any progress.

And now I have no plans to add more use cases.  If there is nothing else missing, I suggest a combined submission

linyihai:
In the commit https://github.com/rust-lang/cargo/pull/13183/commits/a5647da61de2f02ff107eb5f55e701eb99e251dc I add a `public` field when publish a crate, but this is only for test.

We should also support upload the `public` field when publish crates to `cratess-io` and so on.

At least one `public` needs to be added here:  https://github.com/rust-lang/cargo/blob/master/crates/crates-io/lib.rs#L67,
and `optional: dep.is_public()` in  https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/registry/publish.rs#L353C45-L353C45

ehuss:
r? epage

crates/cargo-test-support/src/registry.rs
```diff
@@ -558,6 +558,7 @@ pub struct Dependency {
     package: Option<String>,
     optional: bool,
     default_features: bool,
+    public: bool,
```
epage, line 561: Feel free to break this out into its own PR for us to merge immiedately

tests/testsuite/pub_priv.rs
```diff
@@ -479,3 +479,220 @@ fn allow_priv_in_custom_build() {
         )
         .run()
 }
+
+// A indirectly add D as private dependency.
```
epage, line 483: These tests don't make it too clear that they are not expected behavior but that some help show existing bugs.  We should call that out, linking to the issue in question

epage, line 483: I've been unsure whether these should be merged before the related fix.  Thinking on it more, and with rust-lang/rust#119428 being moved to the rust repo, I suspect we should hold off on tests that demonstrate bad behavior.  The problem is that the fix would be merged into the rust repo but they can't update the tests in the same commit, adding more frustration to the process of fixing the bugs.

Sorry that I hadn't thought of it before you had put more time into polishing this up.

crates/cargo-test-support/src/registry.rs
linyihai: Ok

tests/testsuite/pub_priv.rs
linyihai: It is ok to close this PR or add WIP to the title until the issue fixed in rust repo.

epage:
> and what do you think this? Maybe this is not what you need to do now, because this will add a 'public' field to the user's local file

Sorry, had overlooked that message.  Yes, we need to add publish support.  Looking at bindeps and artifacts, it looks like we do that for unstable features today.

We will also need to make sure the crates.io side is implemented.

linyihai:
> Sorry, had overlooked that message. Yes, we need to add publish support. Looking at bindeps and artifacts, it looks like we do that for unstable features today.
> 
> We will also need to make sure the crates.io side is implemented.

After talking in [t-crates-io channel](https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/publish.20crate.20with.20.60public.60.20field), I've gathered the information as below:
- the crates-io  currently not suport `public` field.
- if they plan to support that they may ignore the `public` in metadata carried with the publish api.

So is it deserves to make a PR  to add `public` field when publish crates?

linyihai:
This PR is outdated and there is no further progress to be made.

PR #13184, opened 2023-12-18
refactor: clean up package metadata

weihanglo:
Like #12352 but for `homepage` and `repository`.

Inspired by <https://github.com/rust-lang/cargo/pull/13178#discussion_r1429394873>.


Cargo.toml
```diff
@@ -126,8 +128,8 @@ version = "0.77.0"
 edition.workspace = true
 license.workspace = true
 rust-version.workspace = true
-homepage = "https://crates.io"
-repository = "https://github.com/rust-lang/cargo"
+homepage.workspace = true
```
weihanglo, line 131: Changed from <https://crates.io> to https://github.com/rust-lang/cargo>. Should be acceptable?

weihanglo: Good idea. Are we going to link to The Cargo Book for all subcrates or just `cargo` main crate?

epage: I think that only makes sense for `cargo`

PR #13187, opened 2023-12-20
refactor: centralize git checkouts and db paths

weihanglo:

This moves the logic of getting `$CARGO_HOME/git/checkouts` and `$CARGO_HOME/.cargo/git/db` to a central place.

There is no functional change.

### DIFF
--- a/src/cargo/util/config/mod.rs
+++ b/src/cargo/util/config/mod.rs
@@ -368,6 +368,18 @@ impl Config {
         self.home_path.join("git")
     }
 
+    /// Gets the directory of code sources Cargo checkouts from Git bare repos
+    /// (`<cargo_home>/git/checkouts`).
+    pub fn git_checkouts_path(&self) -> Filesystem {
+        self.git_path().join("checkouts")
+    }
+
+    /// Gets the directory for all Git bare repos Cargo clones
+    /// (`<cargo_home>/git/db`).
+    pub fn git_db_path(&self) -> Filesystem {
+        self.git_path().join("db")
+    }
+
     /// Gets the Cargo base directory for all registry information (`<cargo_home>/registry`).
     pub fn registry_base_path(&self) -> Filesystem {
         self.home_path.join("registry")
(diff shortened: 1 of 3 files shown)


src/cargo/util/config/mod.rs
```diff
@@ -368,6 +368,18 @@ impl Config {
...
+    /// (`<cargo_home>/git/checkouts`).
+    pub fn git_checkouts_path(&self) -> Filesystem {
+        self.git_path().join("checkouts")
+    }
+
+    /// Gets the directory for all Git bare repos Cargo clones
+    /// (`<cargo_home>/git/db`).
+    pub fn git_db_path(&self) -> Filesystem {
+        self.git_path().join("db")
+    }
```
epage, line 381: How appropriate is it to `join` on a `Filesystem` and pass that around?

From what little I've interacted with `Filesystem`, it seems like its meant to be the root for (e.g. its whats lockable) which makes it feel strange to creating `Filesystem`s for directories we don't lock (instead we lock the parent)

weihanglo: I see it like “hey this path might be locked so be careful when manipulate stuff under the path.” `registry_index_path` is constructed in the same way, though this doesn't justify it is odd.

ehuss:
Thanks!

@bors r+

weihanglo:
Rust version was updated to 1.77 hours ago <https://github.com/rust-lang/rust/pull/119181>. I'll bump the version of Cargo today.

weihanglo:
Blocked on #13192

src/cargo/util/config/mod.rs
epage: Which lock is being held for git checkouts and git dbs?

If its not the individual db's and checkouts, then this seems like the wrong API in the first place as we are saying "this is lockable" when, in reality, locking it would be the wrong decision.  One idea for improving this (if that is the case) is to "lock project" and allow `Filesystem.join` to carry-forward the path to the lockable resource.

Issue #9487, opened 2021-05-15
LGPL-2.1 license mentioned in LICENSE-THIRD-PARTY


hyandell:
Hi Cargo folk!

**Problem**
Reviewing the LICENSE-THIRD-PARTY file, I noticed there was mention of:

    The regex library (deps/regex/) is licensed under the GNU LGPL

https://github.com/rust-lang/cargo/blob/master/LICENSE-THIRD-PARTY#L533

I didn't see an obvious LGPL'd deps/regex; and if it's referring to the https://github.com/rust-lang/regex/ project that's often used in examples, the license of that project appears to be the Rust community standard "MIT OR Apache-2.0". It appears to be in the original 2014 commit. 

Is this now inaccurate attribution?

ehuss:
Thanks for the heads up, indeed the license file has become very out-of-date.

There are some issues tracking the general problem of license management and attribution at https://github.com/rust-lang/rust/issues/39897 and https://github.com/rust-lang/core-team/issues/8.  

I'm not sure what to do here.  One option is to just remove the file, though that doesn't seem ideal.  We could do another pass of the *current* licenses, but that will get out-of-date again over time.  There are some external tools for managing and collecting licenses (like [`cargo-lichking`](https://github.com/Nemo157/cargo-lichking)). But preferably we would have something that would integrate and automate all rust projects.

joshtriplett:
@hyandell It looks like libgit2's upstream COPYING file is out of date, and still lists `deps/regex`, even though they no longer vendor regex (they switched to pcre). Could you get libgit2 to update their COPYING file, and then we can update to a fresh verbatim copy of that file?

ehuss:
The current general meta-tracking issue for "what to do about licenses" is at https://github.com/rust-lang/leadership-council/issues/24.

PR #13196, opened 2023-12-23
fix typos in `LICENSE-THIRD-PARTY` file

Takashiidobe:
This fixes some typos in the license third party file. (Is it ok to fix typos in this file?)

### DIFF
--- a/LICENSE-THIRD-PARTY
+++ b/LICENSE-THIRD-PARTY
@@ -66,7 +66,7 @@ distributions of Cargo:
     The implementation was written so as to conform with Netscapes SSL.
 
     This library is free for commercial and non-commercial use as long as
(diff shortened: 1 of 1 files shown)


weihanglo (commented):
Thanks for the patch. However I don't think this is something could be fixed here. We should keep it as-is since it's copied from dependencies
<https://www.openssl.org/source/license-openssl-ssleay.txt>.

weihanglo:
BTW, the file is pretty outdated and is not really a source of truth for third party licenses. See <https://github.com/rust-lang/cargo/issues/9487#issuecomment-841715408>.

Takashiidobe:
Makes sense to me, closing.

PR #13197, opened 2023-12-23
feat: support `target.<triple>.rustdocflags` officially 

### LINKED ISSUE #13189: `target.<triple>.rustdocflags` is undocumented and buggy
### Problem

It is possible to pass target-specific rustdocflags via config-environment-variables:

```console
> CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTDOCFLAGS=--bar cargo doc
 Documenting foo v0.1.0 (/tmp/scratch.rust-unwrapped.2023-12-21T12-52.BcbaxB/foo)
error: Unrecognized option: 'bar'

error: could not document `foo`

Caused by:
  process didn't exit successfully: `rustdoc --edition=2021 --crate-type bin --crate-name foo src/main.rs -o /run/user/1000/cargo-home/target/shared/doc --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --diagnostic-width=201 --document-private-items '-Arustdoc::private-intra-doc-links' -C metadata=7eced6d5673aa97d -L dependency=/run/user/1000/cargo-home/target/shared/debug/deps --foo --crate-version 0.1.0` (exit status: 1)
```

Trying to pass it via `--config` doesn't work, it seems to believe it must simultaneously be both a table and a string/array-of-strings:

```console
> cargo doc --config 'target.x86_64-unknown-linux-gnu.rustdocflags="--some-flag"'
error: expected a table, but found a string for `target.x86_64-unknown-linux-gnu.rustdocflags` in --config cli option

> cargo doc --config 'target.x86_64-unknown-linux-gnu.rustdocflags.foo="--some-flag"'
error: invalid configuration for key `target.x86_64-unknown-linux-gnu.rustdocflags`
expected a string or array of strings, but found a table for `target.x86_64-unknown-linux-gnu.rustdocflags` in --config cli option
```

It is also not mentioned at all under https://doc.rust-lang.org/nightly/cargo/reference/config.html#target

### Steps

_No response_

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
cargo 1.76.0-nightly (978722961 2023-12-06)
release: 1.76.0-nightly
commit-hash: 9787229614b27854cf73d57ffae430d7c1e6caa4
commit-date: 2023-12-06
os: NixOS 23.11.0 [64-bit]
```

weihanglo:

`CARGO_TARGET_<TRIPLE>_RUSTDOCFLAGS` was accidentally accepted somehow,
so support both config and env officially,
given that removing env support would be a breaking change.

Fixes #13189

Commit by commit.

### Additional information

I leave `target.<cfg>.rustdocflags` for future possibilities,
as it might incur the same convergence issue `target.<cfg>.rustflags` has.

### DIFF
--- a/src/cargo/core/compiler/build_context/target_info.rs
+++ b/src/cargo/core/compiler/build_context/target_info.rs
@@ -698,6 +698,8 @@ impl Flags {
 /// For those artifacts, _only_ `host.*.rustflags` is respected, and no other configuration
 /// sources, _regardless of the value of `target-applies-to-host`_. This is counterintuitive, but
 /// necessary to retain backwards compatibility with older versions of Cargo.
+///
+/// Rules above also applies to rustdoc. Just the key would be `rustdocflags`/`RUSTDOCFLAGS`.
 fn extra_args(
     gctx: &GlobalContext,
     requested_kinds: &[CompileKind],
--- a/src/cargo/core/compiler/build_context/target_info.rs
+++ b/src/cargo/core/compiler/build_context/target_info.rs
@@ -800,7 +800,6 @@ fn rustflags_from_target(
                         .as_ref()
                         .map(|rustflags| (key, &rustflags.val)),
                     // `target.cfg(…).rustdocflags` is currently not supported.
-                    // In fact, neither is `target.<triple>.rustdocflags`.
                     Flags::Rustdoc => None,
                 }
             })
--- a/src/cargo/util/config/target.rs
+++ b/src/cargo/util/config/target.rs
@@ -21,7 +21,7 @@ pub struct TargetCfgConfig {
 }
 
 /// Config definition of a `[target]` table or `[host]`.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub struct TargetConfig {
     /// Process to run as a wrapper for `cargo run`, `test`, and `bench` commands.
     pub runner: OptValue<PathAndArgs>,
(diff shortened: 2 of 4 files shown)


ehuss:
Can you add a test for `cargo test --doc`, too, since it goes through a very different code path?

weihanglo:
Added a new test `target_triple_rustdocflags_works_through_cargo_test`.

Let me know if there is anything missing.

Nemo157:
When combined with `-Zhost-config` does this also support `host.rustdocflags`? (I've just found out we actually need to use this for docs.rs to fix issues with combining `-Zhost-config` and `--config build.rustdocflags`).

weihanglo:
I didn't include that in this PR, but can probably do that if there is a use case.

epage:
@bors r+

Seems reasonable.  Fleshing this out isn't the required way to go (could also deprecate it) but I think its a reasonable to do.  This does slightly expand our compatibility guarantees but its in such alignment with what we already do that I don't think its worth holding a straw vote on this.  If someone disagrees, we can revert before March 21 and re-evaluate.

weihanglo:
> When combined with `-Zhost-config` does this also support `host.rustdocflags`? (I've just found out we actually need to use this for docs.rs to fix issues with combining `-Zhost-config` and `--config build.rustdocflags`).

@Nemo157 could you elaborate on this more and create a new issue, so we won't lose track of it?

PR #13202, opened 2023-12-25
refactor: give some better examples for package ID spec

hi-rustin:

The old examples are so confusing because Cargo does not use those formats. For instance: We use `https://github.com/rust-lang/crates.io-index#foo` instead of `https://github.com/rust-lang/crates.io-index/foo`. You can verify it by using the `cargo pkgid` command.

It should covered by rustdoc tests.

### Additional information

r? @epage  

We discussed it one time during Cargo office hours.


weihanglo (approved):
Thanks!

Issue #13194, opened 2023-12-22
document external cargo subcommand PATH precedence


eval-exec:
### Problem

I have encountered an issue with the `cargo nextest` subcommand where it does not select the expected subcommand's version. 

### Steps

Here are the details of the problem:

1. I have `cargo-nextest` in my PATH, located at `/nix/store/3yl7vgm2qkp7skr5gskbfd457f6g43cg-cargo-nextest-0.9.66/bin/cargo-nextest`, and its version is 0.9.66.
```bash
bash-5.2$ which cargo-nextest
/nix/store/3yl7vgm2qkp7skr5gskbfd457f6g43cg-cargo-nextest-0.9.66/bin/cargo-nextest
```

3. Additionally, I have `cargo-nextest` installed in `~/.cargo/bin`, and its version is 0.9.59.
```bash
bash-5.2$ ~/.cargo/bin/cargo-nextest --version
cargo-nextest 0.9.59
```

4. Running `~/.cargo/bin/cargo-nextest --version` confirms that the version installed in `~/.cargo/bin` is indeed 0.9.59.

5. However, when I execute `cargo nextest --version`, it incorrectly selects version 0.9.59 from `~/.cargo/bin`, instead of the desired version 0.9.66 from `/nix/store/3yl7vgm2qkp7skr5gskbfd457f6g43cg-cargo-nextest-0.9.66/bin/cargo-nextest`.
```bash
bash-5.2$ cargo nextest --version
cargo-nextest-nextest 0.9.59
```

This behavior seems inconsistent and unexpected. It would be helpful if the `cargo` command could prioritize the version in the path `/nix/store/3yl7vgm2qkp7skr5gskbfd457f6g43cg-cargo-nextest-0.9.66/bin` when invoking the `cargo nextest` subcommand.

Thank you for your attention to this issue, and please let me know if you need any further information to reproduce or investigate the problem.

### Version

```text
cargo 1.72.1 (103a7ff2e 2023-08-15)
release: 1.72.1
commit-hash: 103a7ff2ee7678d34f34d778614c5eb2525ae9de
commit-date: 2023-08-15
host: x86_64-unknown-linux-gnu
libgit2: 1.6.4 (sys:0.17.2 vendored)
libcurl: 8.1.2-DEV (sys:0.4.63+curl-8.1.2 vendored ssl:OpenSSL/1.1.1u)
ssl: OpenSSL 1.1.1u  30 May 2023
[...]

PR #13203, opened 2023-12-25
chore(doc): doc for custom subcommands look up.

linyihai:

as the https://github.com/rust-lang/cargo/issues/13194 metions, the lookup rules for custom subcommands are only reflected in comments inside the code, and it is time to inform users of this behavior through documentation.

### Additional information

### DIFF
--- a/src/doc/src/reference/external-tools.md
+++ b/src/doc/src/reference/external-tools.md
@@ -265,6 +265,10 @@ cargo `(?<command>[^ ]+)` into an invocation of an external tool
 `cargo-${command}`. The external tool must be present in one of the user's
 `$PATH` directories.
 
+> **Note**: Cargo defaults to prioritizing external tools in `$CARGO_HOME/bin`
+> over `$PATH`. Users can override this precedence by adding `$CARGO_HOME/bin`
+> to `$PATH`.
+
 When Cargo invokes a custom subcommand, the first argument to the subcommand
 will be the filename of the custom subcommand, as usual. The second argument
 will be the subcommand name itself. For example, the second argument would be


src/doc/src/reference/external-tools.md
```diff
@@ -265,6 +265,12 @@ cargo `(?<command>[^ ]+)` into an invocation of an external tool
 `cargo-${command}`. The external tool must be present in one of the user's
 `$PATH` directories.
 
+> Note: Cargo looks for custom subcommands in `$PATH `and then `$CARGO_HOME/bin`,
```
weihanglo, line 268: This seems a bit incorrect. I may emphasize two things:

* By default, tools in `$CARGO_HOME/bin` take precedence over tools in `$PATH`.
* Users can put `$CARGO_HOME/bin` in `$PATH` to control the precedence.

linyihai: > By default, tools in $CARGO_HOME/bin take precedence over tools in $PATH.

Yes, Indeed.

linyihai: I've polished the description to the best of my ability, hoping to get it right.

src/doc/src/reference/external-tools.md
```diff
@@ -265,6 +265,10 @@ cargo `(?<command>[^ ]+)` into an invocation of an external tool
 `cargo-${command}`. The external tool must be present in one of the user's
 `$PATH` directories.
 
+> Note: By default, Cargo invokes custom subcommand in `$CARGO_HOME/bin`
+> instead of `$PATH`. If you want to change this behavior, you can put
+> `$CARGO_HOME/bin` in `$PATH` to control the precedence.
```
weihanglo, line 270: "instead of" doesn't convey the precedence. How about this?

```suggestion
> **Note**: Cargo defaults to prioritizing external tools in `$CARGO_HOME/bin`
> over `$PATH`. Users can override this precedence by adding `$CARGO_HOME/bin`
> to `$PATH`.
```

linyihai: I've used what you suggested and it looks pretty good without any grammar problems

weihanglo (approved):

weihanglo:
Thank you :)

@bors r+

Issue #13127, opened 2023-12-06
`OUT_DIR` env var Behaviour discrepancy between testing a whole crate and individual test


ShiqiHe000:
### Problem

Hi! I want to run the integration tests of a lib crate (`my_crate_2`) in my rust project. The crate has such structure:
```
my_project/
├── Cargo.toml
└── src/
    └── main.rs
my_crate_2/
├── Cargo.toml
|-- build.rs
└── src/
    └── lib.rs
|__ tests/ 
    |__my_crate_2.rs
```

In `tests/my_crate_2.rs`, I need to extract artifacts output path of the package, i.e., the [`OUT_DIR` env var](https://doc.rust-lang.org/cargo/reference/environment-variables.html), which should be set by `Cargo`. 

[...]

PR #13204, opened 2023-12-25
fix: set OUT_DIR for all units with build scripts

hi-rustin:

close https://github.com/rust-lang/cargo/issues/13127

- Added a minimal case to cover this behavior.
- Set OUT_DIR for all units with build scripts.

Check out the unit test.

### Additional information
None

### DIFF
--- a/src/cargo/core/compiler/context/mod.rs
+++ b/src/cargo/core/compiler/context/mod.rs
@@ -12,6 +12,7 @@ use crate::util::errors::CargoResult;
 use crate::util::profile;
 use anyhow::{bail, Context as _};
 use filetime::FileTime;
+use itertools::Itertools;
 use jobserver::Client;
 
 use super::build_plan::BuildPlan;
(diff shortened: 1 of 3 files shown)


tests/testsuite/test.rs
```diff
@@ -4897,3 +4897,51 @@ fn cargo_test_print_env_verbose() {
...
+                        assert!(std::env::var("OUT_DIR").is_ok());
+                        assert_eq!(foo::add(2, 5), 7);
+                    }
+                }
+            "#,
+        )
+        .build();
+
+    p.cargo("test").run();
+    p.cargo("test --package foo --test case -- tests::test_add --exact --nocapture")
```
hi-rustin, line 4943: It seems that this test is affected by this bug itself. When run independently, it passes. However, if run with 'cargo test -p cargo', it fails.

hi-rustin: Fixed by [`509c75e` (#13204)](https://github.com/rust-lang/cargo/pull/13204/commits/509c75e93126b14cf5cbe28a84afa60778377bef)

hi-rustin (commented):
🔢 Self-check

tests/testsuite/test.rs
```diff
@@ -4941,6 +4941,7 @@ fn cargo_test_set_out_dir_env_var() {
 
     p.cargo("test").run();
     p.cargo("test --package foo --test case -- tests::test_add --exact --nocapture")
+        .env_remove("OUT_DIR")
```
epage, line 4944: Should we do this instead in https://github.com/rust-lang/cargo/blob/master/crates/cargo-test-support/src/lib.rs#L1249 ?

tests/testsuite/test.rs
```diff
@@ -4897,3 +4897,51 @@ fn cargo_test_print_env_verbose() {
         )
         .run();
 }
+
+#[cargo_test]
```
epage, line 4901: Can you modify the commits so this commit passes tests?

src/cargo/core/compiler/context/mod.rs
```diff
@@ -185,6 +186,32 @@ impl<'a, 'cfg> Context<'a, 'cfg> {
...
+                        .to_string();
+                    let script_meta = self.get_run_build_script_metadata(&dep.unit);
+                    self.compilation
+                        .extra_env
+                        .entry(script_meta)
+                        .or_insert_with(Vec::new)
+                        .push(("OUT_DIR".to_string(), out_dir));
+                }
+            }
+        }
```
epage, line 213: If I'm understanding this, we only really need to set `OUT_DIR` once per package and the previous code did that be assuming the lib unit was sufficient (but some packages don't have that).

I'm assuming there isn't a more direct way to iterate over packages to find out the necessary information and set it, and instead we have to walk all units and de-dupe down to individual packages?

tests/testsuite/test.rs
hi-rustin: Updated. Good suggestion. ❤️

tests/testsuite/test.rs
hi-rustin: Updated.

src/cargo/core/compiler/context/mod.rs
hi-rustin: Yes. I think your understanding is correct. Before https://github.com/rust-lang/cargo/pull/3310/files, we set it for every unit. 
image

ehuss:
Thanks!

@bors r+

Issue #12358, opened 2023-07-13
Cargo panic in `ResolvedFeatures::activated_features_int` with unstable `bindeps` feature


phlip9:
### Problem

While experimenting with using the unstable `bindeps` feature, I stumbled upon a panic. Here's a repo with a minimized failing case: <https://github.com/phlip9/cargo-bindep-panic>. Simply clone it and run a cargo command like `cargo build` or `cargo tree` to see the panic.

In plain english, we have a crate, `will-get-bindeped`, that's targetting some other arch like `wasm` or `x86_64-fortanix-unknown-sgx` in our case. At the root is a crate, `top-level-that-depends-on-bindep`, that wants to depend on the built binary of `will-get-bindeped` (hence the `bindeps` feature).

This binary, `will-get-bindeped`, depends on `dep-that-depends-on-build-only-dep` that has a build-time dependency. This `build-only-dep` represents something like `protoc-rust` that needs to run on the host and compile some .proto's or w/e.

`build-only-dep` has a `cfg(unix)` (this seems important!) on `some-leaf-dep`. If `some-leaf-dep` is just an unconditional dependency or even a `cfg(windows)`, the panic no longer triggers.

Visually, the crate graph looks like:

```
top-level-that-depends-on-bindep
  |
  | [dependencies] artifact = "bin", target = "x86_64-fortanix-unknown-sgx"
  |     - or -
  | [build-dependencies] artifact = "bin", target = "x86_64-fortanix-unknown-sgx"
  V
will-get-bindeped
  |
  | [dependencies]
  V
dep-that-depends-on-build-only-dep
  |
  | [build-dependencies]
  V
build-only-dep
  |
  | [target.'cfg(unix)'.dependencies]
  V
some-leaf-dep
```

### Steps

```bash
$ git clone https://github.com/phlip9/cargo-bindep-panic
$ cd cargo-bindep-panic
$ RUST_BACKTRACE=1 cargo check

thread 'main' panicked at 'activated_features for invalid package: features did not find PackageId { name: "some-leaf-dep", version: "0.1.0", source: "/Users/phlip9/dev/cargo-bindep-panic/some-leaf-dep" } ArtifactDep(CompileTarget { name: "x86_64-fortanix-unknown-sgx" })

Stack backtrace:
   0: std::backtrace::Backtrace::create
   1: <anyhow::Error>::msg::<alloc::string::String>
   2: <cargo::core::resolver::features::ResolvedFeatures>::activated_features_int
[...]

Issue #10593, opened 2022-04-22
`cargo tree` panics with "activated_features for invalid package: features did not find PackageId" when a binary dependency is specified with an explicit `target` key


bstrie:
### Problem

For convenience I have also uploaded the failing code here: https://github.com/bstrie/bindeperror4

For the following code:

```
# Cargo.toml
[package]
name = "mycrate"
version = "0.0.0"
edition = "2021"

[dependencies]
mybindep = { path = "mybindep", artifact = "bin", target = "x86_64-unknown-linux-gnu" }
```

```rust
// src/main.rs
fn main() {
    env!("CARGO_BIN_FILE_MYBINDEP");
}
```

```
# mybindep/Cargo.toml
[package]
name = "mybindep"
version = "0.0.0"
edition = "2021"
```

```rust
// mybindep/src/main.rs
fn main() {}
```

Attempting to use `cargo tree` produces the following:

```
thread 'main' panicked at 'activated_features for invalid package: features did not find PackageId { name: "mybindep", version: "0.0.0", source: "/home/ben/code/scrap/bindeperror4/mybindep" } NormalOrDevOrArtifactTarget(None)', src/tools/cargo/src/cargo/core/resolver/features.rs:318:14
```

<details>
  <summary>Expand for full `cargo tree` backtrace.</summary>
  
```
thread 'main' panicked at 'activated_features for invalid package: features did not find PackageId { name: "mybindep", version: "0.0.0", source: "/home/ben/code/scrap/bindeperror4/mybindep" } NormalOrDevOrArtifactTarget(None)

Stack backtrace:
   0: <anyhow::Error>::msg::<alloc::string::String>
   1: <cargo::core::resolver::features::ResolvedFeatures>::activated_features_int
   2: cargo::ops::tree::graph::add_pkg
   3: cargo::ops::tree::graph::add_pkg
   4: cargo::ops::tree::graph::build
   5: cargo::ops::tree::build_and_print
   6: cargo::commands::tree::exec
   7: cargo::cli::main
   8: cargo::main
   9: std::sys_common::backtrace::__rust_begin_short_backtrace::<fn(), ()>
  10: std::rt::lang_start::<()>::{closure#0}
  11: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once
             at /rustc/de1bc0008be096cf7ed67b93402250d3b3e480d0/library/core/src/ops/function.rs:280:13
  12: std::panicking::try::do_call
[...]

PR #13207, opened 2023-12-26
Fix panic when running `cargo tree` on a package with a cross compiled bindep

rukai:

This PR fixes the `cargo tree` panic described in https://github.com/rust-lang/cargo/issues/12358 and https://github.com/rust-lang/cargo/issues/10593

The new integration test is sufficient.

### Additional information

[There was discussion](https://github.com/rust-lang/cargo/issues/10593#issuecomment-1317759526) of holding off on this until a full design of how to present bindeps in `cargo tree` is arrived at.
However I think it makes sense to land this PR first as:
* It introduces a test to catch the simple case.
* The fix itself is small and (afaik) correct.
* Provides genuine value to users by allowing usage of `cargo tree` in many more bindeps use cases.

So this PR is a good first step towards full support in `cargo tree`.

The changes to the `activated_features` methods are not related to the fix but just the improved error reporting I needed to fully diagnose the issue.
I moved `activated_features_int` into `activated_features` and `activated_features_unverified` as I was concerned of the performance cost of generating the full error when its not a fatal error and may occur many times.
I think this change will bring value in the future but I am happy to remove it if it is undesired.

I actually wrote up the test + fix independently before discovering https://github.com/rust-lang/cargo/issues/10593#issuecomment-1317759526 , once I discovered it I copied in some specific parts to improve comments + correctness

I did however leave out the `None if features_for != FeaturesFor::default() => features_for,` because that seems wrong to me. proc macro and build dependencies of a bindep need to be compiled for the host, regardless of the bindeps target, otherwise the host cant run them to build the bindep.

### DIFF
--- a/src/cargo/core/resolver/features.rs
+++ b/src/cargo/core/resolver/features.rs
@@ -319,8 +319,32 @@ impl ResolvedFeatures {
         pkg_id: PackageId,
         features_for: FeaturesFor,
     ) -> Vec<InternedString> {
-        self.activated_features_int(pkg_id, features_for)
-            .expect("activated_features for invalid package")
+        let fk = features_for.apply_opts(&self.opts);
+        let key = (pkg_id, fk);
+        if let Some(fs) = self.activated_features.get(&key) {
+            fs.iter().cloned().collect()
+        } else {
+            panic!(
+                "did not find features for {key:?} within activated_features:\n{:#?}",
+                self.activated_features
+            )
+        }
+    }
+
+    /// Variant of `activated_features` that returns `None` if this is
+    /// not a valid pkg_id/is_build combination. Used in places which do
+    /// not know which packages are activated (like `cargo clean`).
+    pub fn activated_features_unverified(
+        &self,
+        pkg_id: PackageId,
+        features_for: FeaturesFor,
+    ) -> Option<Vec<InternedString>> {
+        let fk = features_for.apply_opts(&self.opts);
+        if let Some(fs) = self.activated_features.get(&(pkg_id, fk)) {
+            Some(fs.iter().cloned().collect())
+        } else {
+            None
+        }
     }
 
     /// Returns if the given dependency should be included.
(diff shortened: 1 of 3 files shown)


rukai:
probably makes sense for
r? weihanglo

src/cargo/core/resolver/features.rs
```diff
@@ -319,8 +319,32 @@ impl ResolvedFeatures {
         pkg_id: PackageId,
         features_for: FeaturesFor,
     ) -> Vec<InternedString> {
-        self.activated_features_int(pkg_id, features_for)
-            .expect("activated_features for invalid package")
+        let fk = features_for.apply_opts(&self.opts);
```
weihanglo, line 322: Would like to see this refactor being in its own commit, to following atomic commit principle.

src/cargo/ops/tree/graph.rs
```diff
@@ -391,10 +391,20 @@ fn add_pkg(
...
-                FeaturesFor::HostDep
-            } else {
-                features_for
+            let dep_features_for = match dep
+                .artifact()
+                .and_then(|artifact| artifact.target())
+                .and_then(|target| target.to_resolved_compile_target(requested_kind))
+            {
+                // Dependency has a `{ …, target = <triple> }`
+                Some(target) => FeaturesFor::ArtifactDep(target),
```
weihanglo, line 400: I might have a rotted memory of artifact dependencies. However, even when `target` is present the dependency might still be depended as a normal `lib` (with `lib = true`). The fix doesn't look 100% correct to me for this reason.

tests/testsuite/artifact_dep.rs
```diff
@@ -1445,6 +1445,55 @@ foo v0.0.0 ([CWD])
         )
         .run();
 }
+
+#[cargo_test]
+fn artifact_dep_target_specified() {
```
weihanglo, line 1450: This test passes even without the patch in `ops::tree::graph`. We should probably first create a test verifying the current panic behavior in a commit, followed by the other commit fixing both the behavior and the test.

weihanglo (commented):
Sorry for the way too long delay of the review 😞.

src/cargo/ops/tree/graph.rs
weihanglo: Okay this looks like an adaption of my own patch... Need to refresh my memory 😅.

src/cargo/ops/tree/graph.rs
weihanglo, line 1: >  I did however leave out the `None if features_for != FeaturesFor::default() => features_for`, because that seems wrong to me. proc macro and build dependencies of a bindep need to be compiled for the host, regardless of the bindeps target, otherwise the host cant run them to build the bindep.

See <https://rust-lang.github.io/rfcs/3028-cargo-binary-dependencies.html#reference-level-explanation>, so arguably some build deps could be built for a target platform.

> By default, `build-dependencies` are built for the host, while  `dependencies` and `dev-dependencies` are built for the target. You can specify the `target` attribute to build for a specific target, such as `target = "wasm32-wasi"`; a literal `target = "target"` will build for the target even if specifying a build dependency. (If the target is not available, this will result in an error at build time, just as if building the specified crate with a `--target` option for an unavailable target.)

weihanglo:
@rukai  are you still interested in fixing this?

rukai:
I cant afford the time to work on this any time soon.
Anyone else is free to take this on.

elchukc:
We can probably close this

weihanglo:
Yes absolutely. Thank you!

Issue #13171, opened 2023-12-14
trim-paths: Remap rules for different dependency kinds


weihanglo:
As of 6982b443cf438732f6652c798bff9244d77f753c, the remap of `-Ztrim-paths` is relatively simple. It follows what is described in [RFC 3127](https://rust-lang.github.io/rfcs/3127-trim-paths.html#trim-paths-implementation-in-cargo):

> For the the current package (where the current working directory is in), from the the absolute path of the package root to empty string. For other packages, from the absolute path of the package root to `[package name]-[package version]`.

In Cargo, that is: https://github.com/rust-lang/cargo/blob/6982b443cf438732f6652c798bff9244d77f753c/src/cargo/core/compiler/mod.rs#L1222-L1230

However, this is not ideal for debugger. Several questions emerge in <https://github.com/rust-lang/rust/issues/111540#issuecomment-1852027930>. 

* Impossible to manually configure remap for debugger — People need to configure one remap rules per per dependency. That is not possible for people to do it manually for each `<pkg>-<version>/` back to absolute paths. It would require a dedicated tool to restore all the paths.
* The same issue also affects path dependency remap. For path dependencies outside the current workspace root, we remap paths to `<pkg>-<version>` as well.
* There is no way to distinguish relative path from different Cargo workspace — For path dependencies inside a workspace Cargo currently remaps to paths relative to workspace root. Imagine you have multiple Cargo workspaces and there is one non-Rust root project compiles them together. Now all your debuginfo files are messed up since they were remapped to relative path like `src/some/file.rs`.

### Registry and git dependencies

In <<https://github.com/rust-lang/rust/issues/111540#issuecomment-1852027930>, @michaelwoerister brought up a smarter remap strategy. We could remap paths from `$CARGO_HOME/registry/src` and `$CARGO_HOME/git/checkouts` to a fixed string for registry and git dependencies respectively, so that there is a finite set of rule (two exactly) to configure in debugger. That is to say, cargo could unconditionally pass these to the compiler:

* `--remap-path-prefix=$CARGO_HOME/.cargo/registry/src=registry@`
* `--remap-path-prefix=$CARGO_HOME/.cargo/git/checkouts=git@`

(or conditionally if we want to reduce number of args passed in to rustc)

### Path dependencies

[...]

PR #13210, opened 2023-12-26
fix(trim-paths): remap common prefix only

weihanglo:

New `-Ztrim-paths` remap rules for git/registry dependencies:

* Git dependencies: remove ~/.cargo/git/checkouts prefix.
* Registry dependencies: remove ~/.cargo/registry/src prefix.

This make the remap rules fixed in a finite number,
minimizing the burden for users to configure debuggers.

Part of <https://github.com/rust-lang/cargo/issues/13171>.

The new rule is a small deviation of RFC 3127.
Please review if it makes sense.

This doesn't tackle relative path dependencies described in <https://github.com/rust-lang/cargo/issues/13171>,
as git/registry should be less controversial and easier to be landed.

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1217,18 +1217,22 @@ fn trim_paths_args(
     let package_remap = {
         let pkg_root = unit.pkg.root();
         let ws_root = cx.bcx.ws.root();
-        let is_local = unit.pkg.package_id().source_id().is_path();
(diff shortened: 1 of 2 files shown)


epage:
Had some more thoughts on the underlying proposal that I posted over at https://github.com/rust-lang/cargo/issues/13171#issuecomment-1870491266

epage:
I have reservations but I've been slow to work them out and this is all unstable, so there isn't really a reason to hold this up.

@bors r+

PR #13213, opened 2023-12-28
`cargo add` - fix for adding features from repository with multiple packages.

### LINKED ISSUE #13121: Adding a dependency from a github repository with features seems to be failing/disambiguation command is not properly parsed
### Problem

Adding a workspace member from a github repository as a dependency via `cargo add` specifying existing features triggers an issue

### Steps

1. cargo add --git https://github.com/zama-ai/tfhe-rs.git tfhe --features=x86_64-unix

output:
```console
    Updating git repository `https://github.com/zama-ai/tfhe-rs.git`
      Adding tfhe (git) to dependencies.
error: unrecognized feature for crate tfhe: x86_64-unix
no features available for crate tfhe
```

As one can see here the feature does exist in the tfhe workspace member Cargo.toml:

https://github.com/zama-ai/tfhe-rs/blob/ad41fdf5a5060c0a981cd0c35bf998feafe68e02/tfhe/Cargo.toml#L138

### Possible Solution(s)

Unsure, I don't know the cargo code base, but my guess would be to parse the workspace member Cargo.toml to get some metadata/information about the crate

### Notes

this is probably an edge case but can prove interesting for automated workflows that may be doing some automatic api breaks/checks by adding a work in progress version of a project to a test/regression project and run tests from that project

### Version

```text
cargo 1.76.0-nightly (9b13310ca 2023-11-24)
release: 1.76.0-nightly
commit-hash: 9b13310ca596020a737aaa47daa4ed9ff8898a2f
commit-date: 2023-11-24
```

stupendoussuperpowers:
Fixes #13121 

As discussed in the issue, when using `cargo add` to add a package with features from a git repository from one of it's members, the command might fail due to improper target for querying for said features. 

This PR adds a test for this edge-case where we expect it to fail with current code. It also adds a fix for this, and updates the test to expect success.

While populating available features, the current code does a `Fuzzy` search which might lead to searching for features in a wrong member package. If we change it to an `Exact` query, we get back the proper member to search within.

### DIFF
--- a/src/cargo/ops/cargo_add/mod.rs
+++ b/src/cargo/ops/cargo_add/mod.rs
@@ -934,7 +934,7 @@ fn populate_available_features(
     }
 
     let possibilities = loop {
-        match registry.query_vec(&query, QueryKind::Fuzzy) {
+        match registry.query_vec(&query, QueryKind::Exact) {
             std::task::Poll::Ready(res) => {
                 break res?;
             }
--- a/tests/testsuite/cargo_add/git_multiple_packages_features/mod.rs
+++ b/tests/testsuite/cargo_add/git_multiple_packages_features/mod.rs
@@ -1,0 +1,63 @@
+use cargo_test_support::compare::assert_ui;
+use cargo_test_support::prelude::*;
+use cargo_test_support::Project;
+
+use cargo_test_support::curr_dir;
+
+#[cargo_test]
+fn case() {
+    cargo_test_support::registry::init();
+
+    let main_manifest = r#"
+    [package]
+    name = "main-package"
+    version = "0.1.1+main-package"
+    authors = []
+
+    [workspace]
+    members = ["package-wo-feature", "package-with-feature"]
+    "#;
+
+    let manifest_feature = r#"
+    [package]
+    name = "package-with-feature"
+    version = "0.1.3+package-with-feature"
+    [features]
+    target_feature = []
+    "#;
+
+    let project = Project::from_template(curr_dir!().join("in"));
+    let project_root = project.root();
+    let cwd = &project_root;
+    let git_dep = cargo_test_support::git::new("git-package", |project| {
+        project
+            .file("Cargo.toml", &main_manifest)
+            .file(
+                "package-wo-feature/Cargo.toml",
+                &cargo_test_support::basic_manifest(
+                    "package-wo-feature",
+                    "0.1.1+package-wo-feature",
+                ),
+            )
+            .file("package-wo-feature/src/lib.rs", "")
+            .file("package-with-feature/Cargo.toml", &manifest_feature)
+            .file("package-with-feature/src/lib.rs", "")
+    });
+    let git_url = git_dep.url().to_string();
+
+    snapbox::cmd::Command::cargo_ui()
+        .arg("add")
+        .args([
+            "--git",
+            &git_url,
+            "package-with-feature",
+            "--features=target_feature",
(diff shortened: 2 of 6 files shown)


tests/testsuite/cargo_add/git_multiple_packages_features/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+use cargo_test_support::Project;
+
+use cargo_test_support::curr_dir;
+
+#[cargo_test]
+fn case() {
+    cargo_test_support::registry::init();
+
+    let mut members_manifest = cargo_test_support::basic_manifest("packmain", "0.1.1+mainpack");
+    members_manifest.push_str("members = [pack1, pack2]");
```
epage, line 12: I feel like this is brittle, assuming that a `[workspace]` is placed last.

Personally, I'd just go ahead and write out the desired manifest.

tests/testsuite/cargo_add/git_multiple_packages_features/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+    let git_dep = cargo_test_support::git::new("git-package", |project| {
+        project
+            .file("Cargo.toml", &members_manifest)
+            .file(
+                "pack1/Cargo.toml",
+                &cargo_test_support::basic_manifest("pack1", "0.1.1+pack1"),
+            )
+            .file("pack1/src/lib.rs", "")
+            .file("pack2/Cargo.toml", &features_manifest)
+            .file("pack2/src/lib.rs", "")
```
epage, line 29: nit: Could you give these package names a more descriptive name so its easier to inspect the test and tell what is intended

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -934,7 +934,7 @@ fn populate_available_features(
     }
 
     let possibilities = loop {
-        match registry.query_vec(&query, QueryKind::Fuzzy) {
+        match registry.query_vec(&query, QueryKind::Exact) {
```
epage, line 937: The test in the first commit should be passing on its own and this commit should update the test so its still passing.

By the commit title for the first test, I'm assuming it isn't passing.  The intention of this pattern is
- Tests always pass, making bisection easier
- This shows to you and reviewers that the test does what it intended
- This shows to your reviewers what behavior change this PR causes

epage:
Thanks for picking this up and moving it along!

src/cargo/ops/cargo_add/mod.rs
stupendoussuperpowers: For this, should that initial test just contain one package within it so that we don't hit this issue?

epage: When I do this, I have the test verify we are hitting the failure mode.

For example for #13164
- https://github.com/rust-lang/cargo/pull/13164/commits/7200646a9794468d466a65267d1a33ce872cfed6 adds a lot of tests that pass
- https://github.com/rust-lang/cargo/pull/13164/commits/179135368991f442dc7770bc0a0a0ba044368444 changed the behavior.  In this case, it is showing the error changing from one message to another but in your case it would be an error changing to success.

stupendoussuperpowers:
have made the changes. 

first commit - add a test that expects failure from said command (test passes)
second commit - fixes issue, updates test to expect success + appropriate output (test passes)

epage:
Thanks! 

@bors r+

PR #13216, opened 2023-12-29
Remove repetitive words

cuishuang:

### DIFF
--- a/credential/cargo-credential/examples/stdout-redirected.rs
+++ b/credential/cargo-credential/examples/stdout-redirected.rs
@@ -15,7 +15,7 @@ impl Credential for MyCredential {
         _args: &[&str],
     ) -> Result<CredentialResponse, Error> {
         // Informational messages should be sent on stderr.
-        eprintln!("message on stderr should be sent the the parent process");
+        eprintln!("message on stderr should be sent to the parent process");
 
         // Reading from stdin and writing to stdout will go to the attached console (tty).
         println!("message from test credential provider");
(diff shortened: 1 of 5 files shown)


credential/cargo-credential/examples/stdout-redirected.rs
```diff
@@ -15,7 +15,7 @@ impl Credential for MyCredential {
         _args: &[&str],
     ) -> Result<CredentialResponse, Error> {
         // Informational messages should be sent on stderr.
-        eprintln!("message on stderr should be sent the the parent process");
+        eprintln!("message on stderr should be sent the parent process");
```
ehuss, line 18: I believe this is intended to be "to the parent process".

```suggestion
        eprintln!("message on stderr should be sent to the parent process");
```

credential/cargo-credential/tests/examples.rs
```diff
@@ -15,7 +15,7 @@ fn stdout_redirected() {
         .arg("--cargo-plugin")
         .assert()
         .stdout_eq(format!("{hello}\n{err_not_supported}\n"))
-        .stderr_eq("message on stderr should be sent the the parent process\n")
+        .stderr_eq("message on stderr should be sent the parent process\n")
```
ehuss, line 18: Same here.

```suggestion
        .stderr_eq("message on stderr should be sent to the parent process\n")
```

credential/cargo-credential/examples/stdout-redirected.rs
cuishuang: Thanks. Done

credential/cargo-credential/tests/examples.rs
cuishuang: Thanks.  Done.

PR #13218, opened 2023-12-29
Ignore test `fix_in_dependency` to make a change in rustc

petrochenkov:
This test relies on some specific diagnostics being produced in rustc, but those diagnostics are currently going through changes (https://github.com/rust-lang/rust/pull/119204 and there will be some more follow up PRs).

### DIFF
--- a/tests/testsuite/fix.rs
+++ b/tests/testsuite/fix.rs
@@ -1859,6 +1859,8 @@ fn non_edition_lint_migration() {
 }
 
 // For rust-lang/cargo#9857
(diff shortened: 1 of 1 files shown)


petrochenkov:
The CI failure looks unrelated.

ehuss:
I can post a PR to fix it to not rely on the compiler.

ehuss:
Posted #13220 as an alternative.

ehuss:
Closing as this should be resolved by https://github.com/rust-lang/cargo/pull/13220. Thanks!

PR #9938, opened 2021-09-22
Skip all `cargo fix` that tends to write to registry cache. 

weihanglo:
Skip all `cargo fix` that tends to write to registry cache. 

This is a temporary hack for #9857. The real fix may need to touch rustc.

PR #13220, opened 2023-12-29
Fix fix::fix_in_dependency to not rely on rustc

ehuss:
This changes the `fix::fix_in_dependency` test so that it does not rely on the behavior of rustc.

This test is checking the behavior when rustc includes a suggestion diagnostic that modifies a file in CARGO_HOME from a dependency. rustc should not be emitting suggestions that point outside of the current crate, but there are some known bugs where it does this. #9938 added a workaround for this to avoid writing to CARGO_HOME.

However, the current test was relying on one of those bugs in rustc to exhibit its behavior. https://github.com/rust-lang/rust/pull/119204 is fixing that particular behavior. Instead of relying on issues in rustc, this PR changes the test so that the test creates a fake `rustc` executable that will emit a hard-coded JSON message that tells `cargo fix` to modify a file in CARGO_HOME. This should be stable as it is no longer relying on rustc.

Testing or validating this change is a little tricky. You have to essentially comment out [these three lines](https://github.com/rust-lang/cargo/blob/4f70d1781a2b2d3611087e1ee6e8096903c9b73a/src/cargo/ops/fix.rs#L654-L656) from the original #9938 change, and verify that the test fails (it says "Fixed" in the output). With those three lines in place, it should pass.

### DIFF
--- a/tests/testsuite/fix.rs
+++ b/tests/testsuite/fix.rs
@@ -1858,9 +1858,22 @@ fn non_edition_lint_migration() {
     assert!(contents.contains("from_utf8(crate::foo::FOO)"));
 }
 
-// For rust-lang/cargo#9857
 #[cargo_test]
 fn fix_in_dependency() {
+    // Tests what happens if rustc emits a suggestion to modify a file from a
+    // dependency in cargo's home directory. This should never happen, and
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
Thanks. This looks good to me.

PR #7450, opened 2019-09-27
Stabilize cache-messages

ehuss:
This stabilizes the -Zcache-messages feature, making it always enabled.

## What is stabilized?

This feature is intended to redisplay previous warnings on a "fresh" build instead of displaying no output. 

Users have occasionally indicated frustration when they know there are warnings, but no output is displayed when the build is fresh. This also improves the interaction between `cargo check` and `cargo clippy-preview`. This also simplifies the code, and opens more possibilities for `rustc` to send side-channel messages to Cargo.

Cargo will now use JSON output from `rustc` and `rustdoc` 100% of the time (`rustdoc --test` does not use JSON). Previously Cargo would only use JSON for pipelined crates.

Cargo will save the JSON output into a file named `output` in the `.fingerprint` directory. This file is only created when the compiler outputs a diagnostic message.

If a crate is being recompiled, and Cargo considers it to be "fresh", it will replay the output file to the console.

## Notable changes in this PR

- Fixed a bug where replays were erroneously including pipeline rmeta artifact json messages.
- clippy-preview is now included in the metadata hash, to force its artifacts to be separate from `cargo check`.
- clippy-preview is no longer force-enabled, under the assumption that caching and fingerprinting is accurate, and the cached messages will be replayed.
- clippy-preview's arguments are included in the fingerprint now that it is not force-enabled.
- Rustdoc colors and short messages were fixed when pipelining was stabilized, so updated tests.

Closes #6986
Closes #6848
Closes #6664
Closes #2458

## Concerns

The only notable issue with this is that switching between short and long human messages only replays the format from the first invocation.  That is, if you do `cargo build` and it generates warnings, then running again with `--message-format=short` will still show the full length human messages. I'm personally fine with that behavior, even though it is not ideal. I think this feature overall improves the situation (where before *no* output was displayed). Being able to re-render between short/long is a very difficult problem, and unlikely to be fixable in the foreseeable future.

There was some concern expressed about being able to disable this. I think that would only be necessary if a severe bug is discovered. I do not feel that this change is risky enough to warrant a configurable option. If it does cause a problem, it can be quickly reverted with a one-line change to set `OutputOptions::cache_cell` to `None`. Since pipelining has been using JSON output for a while now without complaints, I feel pretty confident in it.

PR #13224, opened 2023-12-30
cleanup: Remove error-format special-case in `cargo fix`

ehuss:
`cargo fix` had some special handling to deal with ensuring that the `--error-format=json` flag was passed to rustc. However, cargo was changed in https://github.com/rust-lang/cargo/pull/7450 to always pass that flag. This special handling is no longer necessary, so this PR removes it.


epage:
@bors r+

If there is a conflict with the other PR, feel free to r= me when it is resolved.

PR #13225, opened 2023-12-30
`cargo fix`: always inherit the jobserver

ehuss:
https://github.com/rust-lang/cargo/pull/12951 changed `cargo fix` to ensure that `rustc` has access to the jobserver. However, it only did that for the final "verification" step. It was not set for any of the previous calls to rustc (which is up to ~5 calls).

I'm not sure if this was done intentionally, I did not see any discussion of this in #12951.

This isn't too important in the grand scheme of things, because `rustc` is not doing codegen, so the parallel behavior is currently not used. However, this removes an extraneous `warning: failed to connect to jobserver from environment variable` warning that is printed in cargo's log every time it runs rustc. It also might be relevant in the future if rustc enables the parallel frontend.


weihanglo (approved):
Make sense. Thank you!

PR #13226, opened 2023-12-31
rustfix: Support inserting new lines.

ehuss:
If rustfix received a suggestion which inserts new lines without replacing existing lines, it would ignore the suggestion. This is because `parse_snippet` would immediately return if the `lines` to replace was empty.

The solution here is to just drop the code which messes with the original text line. `cargo fix` (and compiletest) currently do not use this. This was originally added back in the days when rustfix supported an interactive UI which showed color highlighting of what it looks like with the replacement. My feeling is that when we add something like this back in, I would prefer to instead use a real diff library and display instead of trying to do various text manipulation for display. This particular code has generally been buggy, and has been a problem several times.

The included test fails without this fix because the changes do not apply, and the code cannot compile.

This also includes a little bit of cleanup for the `parse_and_replace` test. My feeling is that this test could use further improvements.


crates/rustfix/tests/parse_and_replace.rs
```diff
@@ -20,6 +42,7 @@ mod settings {
     pub const CHECK_JSON: &str = "RUSTFIX_TEST_CHECK_JSON";
     pub const RECORD_JSON: &str = "RUSTFIX_TEST_RECORD_JSON";
     pub const RECORD_FIXED_RUST: &str = "RUSTFIX_TEST_RECORD_FIXED_RUST";
+    pub const BLESS: &str = "RUSTFIX_TEST_BLESS";
```
weihanglo, line 45: This makes me wonder if we should use `snapbox` to minimize the custom logic here.

weihanglo (approved):

weihanglo:
Thank you. Going to merge this soon.
We can do snapbox migration later on if we want to.

@bors r+

Issue #10936, opened 2022-08-04
`cargo test --tests` also runs unit tests -- target filters are confusing


RalfJung:
### Problem

`cargo test` has flags `--libs`, `--bins`, `--tests`, `--doc`, which I thought (based on their documentation) would control whether the lib crate unit tests / bin crate unit tests / integration tests / doc tests, respectively, are being run. But that does not seem to be what happens, so either the documentation is wrong or it is very confusing.

### Steps

1. Run `cargo test --tests`.
2. Notice how it runs not just the integration tests, but also the unit tests!
3. However, `cargo test --bins` runs only the binary crate unit tests.

This seems clearly inconsistent. I would expect `cargo test --tests` to be the same as `cargo test --test a --test b` (manually listing all tests), which is how `cargo test --bins` seems to behave.

### Notes

Generally the `cargo test` documentation for all these many flags that control which tests are being run, is unclear on whether the flag *adds* tests to the default test set, or *remove* the default test set to only run a small select set of tests. (This also affects other commands like `cargo check`.) For instance, `cargo test --benches` tests more than the default test set, but `cargo test --bins` tests less than the default test set, but the documentation for these two flags reads exactly the same. Some flags use the word 'only', which is maybe meant to indicate "reduce default test set", but `--bins` does not use that word and anyway this is not at all clear.

The behavior I would have expected is along the lines of "if any of these flags is present, then only what is specified by the flags is being run, no default test set", but that does not match implemented behavior. This then makes me wonder how I would even say "run only integration tests" (`cargo test --tests` doesn't do that), or "check only the benchmarks" (`cargo check --benches` doesn't do that).

### Version

```text
cargo 1.64.0-nightly (85b500cca 2022-07-24)
release: 1.64.0-nightly
commit-hash: 85b500ccad8cd0b63995fd94a03ddd4b83f7905b
commit-date: 2022-07-24
host: x86_64-unknown-linux-gnu
libgit2: 1.4.2 (sys:0.14.2 vendored)
libcurl: 7.83.1-DEV (sys:0.4.55+curl-7.83.1 vendored ssl:OpenSSL/1.1.1q)
os: Debian testing (bookworm) [64-bit]
```

weihanglo:
Totally agree it is confusing. People usually invoke `cargo test --help` and see it say `Test all tests`. Then guessing like _"I know cargo tests. **Tests** here must refer to **all test targets**"_. And it turns out to be all targets that has default test flag set.

[...]

PR #13236, opened 2024-01-02
Clarify the function of the test options

heisen-li:

Make the description of test options clearer.

from #10936

### DIFF
--- a/src/bin/cargo/commands/test.rs
+++ b/src/bin/cargo/commands/test.rs
@@ -38,12 +38,12 @@ pub fn cli() -> Command {
         )
         .arg_targets_all(
             "Test only this package's library unit tests",
-            "Test only the specified binary",
-            "Test all binaries",
+            "Test only the specified binary's unit tests",
+            "Test all binaries's unit tests",
             "Test only the specified example",
             "Test all examples",
-            "Test only the specified test target",
-            "Test all test targets",
+            "Test only the specified target's integration tests",
+            "Test all test targets's integration tests and library unit tests",
             "Test only the specified bench target",
             "Test all bench targets",
             "Test all targets (does not include doctests)",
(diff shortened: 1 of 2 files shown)


src/bin/cargo/commands/test.rs
```diff
@@ -38,12 +38,12 @@ pub fn cli() -> Command {
         )
         .arg_targets_all(
             "Test only this package's library unit tests",
-            "Test only the specified binary",
-            "Test all binaries",
+            "Test only the specified binary's unit tests",
```
epage, line 41: For myself, I would prefer removing references to unit tests, rather than focusing more on unit/integration test terminology.

Its a bit of a misnomer to refer to them as unit/integration tests.  The mechanism allows whitebox / blackbox testing and the user can put unit or integration tests in either location.

src/bin/cargo/commands/test.rs
```diff
@@ -38,12 +38,12 @@ pub fn cli() -> Command {
...
+            "Test all binaries's unit tests",
             "Test only the specified example",
             "Test all examples",
-            "Test only the specified test target",
-            "Test all test targets",
+            "Test only the specified target's integration tests",
+            "Test all test targets's integration tests and library unit tests",
```
epage, line 46: This seems odd so I would want us to confirm it and whether it is expected behavior updating documenation

src/bin/cargo/commands/test.rs
weihanglo: Agree on Ed's argument. Mentioning unit tests is still confusing. “What does it mean by binary unit tests?”

heisen-li:
@weihanglo I understand that epage should mean to remove unit tests from the option description. Would you take a look?

weihanglo (approved):
I am okay with the change. @epage, do you have any other opinion?

PR #13237, opened 2024-01-02
Contrib: Fix team HackMD links

ehuss:
This adjusts the links in the contrib guide to HackMD to use links which are publicly visible. I did not intend to use the other links which AFAIK are private to team members.


weihanglo (approved):
Confirmed that all links are accessible to anonymous users.

PR #13239, opened 2024-01-02
chore: tracing be compat with rustc_log

weihanglo:

The current version of rustfix cannot be used in rust-lang/rust due to the dependency on `tracing`. `tracing-core` is currently locked to 0.1.30, which is too old for the version in the cargo repo (it looks like they are trying to avoid a known deadlock).

See: https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/rustfix.20is.20incompatible.20with.20rust-lang.2Frust

### Additional information

I guess we need a beta-backport for this?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,7 +98,7 @@ thiserror = "1.0.56"
 time = { version = "0.3", features = ["parsing", "formatting", "serde"] }
 toml = "0.8.8"
 toml_edit = { version = "0.21.0", features = ["serde"] }
-tracing = "0.1.40"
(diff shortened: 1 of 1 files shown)


epage (approved):
Thoughts on also holding back the version in `Cargo.toml` to keep the version we test and use in production the same as rustc?

Either way, feel free to r= me

weihanglo:
I am not sure if the deadlock is universal or rustc specific. Let's merge this, and if it is too bad we can pin the version then.

@bors r=epage

PR #13241, opened 2024-01-02
feat(embedded): Add multiple experimental manifest syntaxes

epage:

As syntax discussions for "cargo script" are on-going, this allows us to experiment with a couple of them so we can see how they work in practice.

This is missing the line-prefix syntax as we decide how we want to separate blocks for it.

While doing this, I removed the previous doc-comment syntax.  This was left in for transition purposes.  With where discussions are going, its unlikely we'll go back to that syntax.

### Additional information

rust-lang/rfcs#3503

https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/Syntax.20for.20embedded.20tooling.20metadata


src/cargo/util/toml/embedded.rs
```diff
@@ -229,16 +217,28 @@ fn split_source(input: &str) -> CargoResult<Source<'_>> {
...
     let (fence_pattern, rest) = match tick_end {
         0 => {
             return Ok(source);
         }
         1 | 2 => {
+            if tick_char == '#' {
+                // Attribute
+                return Ok(source);
+            }
             anyhow::bail!("found {tick_end} backticks in rust frontmatter, expected at least 3")
```
Muscraft, line 242: I think this comment needs updated
```suggestion
            anyhow::bail!("found {tick_end} `{tick_char}` in rust frontmatter, expected at least 3")
```

src/cargo/util/toml/embedded.rs
```diff
@@ -229,16 +217,28 @@ fn split_source(input: &str) -> CargoResult<Source<'_>> {
...
         .unwrap_or(source.content.len());
     let (fence_pattern, rest) = match tick_end {
         0 => {
             return Ok(source);
         }
         1 | 2 => {
+            if tick_char == '#' {
+                // Attribute
+                return Ok(source);
+            }
```
Muscraft, line 241: Should this explicitly check for `#![]` or `#[]`?

src/cargo/util/toml/embedded.rs
```diff
@@ -66,17 +63,8 @@ pub(super) fn expand_manifest(
         let manifest = toml::to_string_pretty(&manifest)?;
         Ok(manifest)
     } else {
-        // Legacy doc-comment support; here only for transitional purposes
-        let comment = extract_comment(content)?.unwrap_or_default();
```
Muscraft, line 70: Should there be a message saying we no longer support comment-based manifests if we encounter one?

epage: This would require still parsing the doc comment and I feel like its not worth keeping around longer for an unstable feature.

src/cargo/util/toml/embedded.rs
epage: I'd rather not get into the complexity of matching rust syntax.  If we go this route, we can further evaluate what the exact behavior should be.

Muscraft (approved):
Looks good to me! As this is an eRFC, I am happy to approve this.

weihanglo:
Should we update the unstable doc accordingly?

epage:
I think I'd rather wait until we have something more definitive to say about what direction we are going.

PR #13243, opened 2024-01-02
`cargo fix`: Call rustc fewer times.

### LINKED ISSUE #13215: `cargo fix` checks packages twice, even if nothing was fixed
### Problem

Running `cargo fix` (and worse, `cargo clippy --fix` on https://github.com/crate-ci/typos is very slow.  For clippy, its over 2m.

### Proposed Solution

Only re-check the package (to show errors/warnings) if there was output from the first run.

### Notes

See also https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Performance.20of.20.60cargo.20fix.60

ehuss:
This is an improvement of `cargo fix` so that it calls `rustc` one fewer times per target. The original code an extra step of calling `rustc` to display final diagnostics to the user. Part of the reason is that in the past, cargo did not always use JSON, and so `cargo fix` was forced to call `rustc` with and without JSON. Now that cargo uses JSON all the time, that is not necessary. This avoids the final call to `rustc` by remembering the original output from `rustc`.

This needs to keep track of both the first and last output from `rustc`. This is to handle the situation where `cargo fix` fails to apply some suggestion (either because the verification fails, or `rustfix` fails). The `cargo fix` output includes both the error, and the original diagnostics before the error.

The first commit is a little test framework to exercise the various edge cases around how fix works. The comments should explain how it works, but it essentially has a `rustc` replacement that emits various different diagnostics and counts how often it is called.

The subsequent commit includes the change to keep track of the output from `rustc` and to avoid the final call.

Fixes #13215


epage:
Thanks!

@bors r+

PR #13244, opened 2024-01-02
chore: update auto labels

weihanglo:
See

* `A-errors` has been renamed to `A-diagnostics`
* Some manifest related modules were moved to `cargo-util-schemas`.

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -120,9 +120,6 @@ trigger_files = [
   "src/cargo/util/config/environment.rs",
 ]
 
-[autolabel."A-errors"]
-trigger_files = ["src/cargo/util/diagnostic_server.rs"]
-
 [autolabel."A-features2"]
 trigger_files = ["src/cargo/core/resolver/features.rs"]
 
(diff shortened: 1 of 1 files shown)


triagebot.toml
```diff
@@ -120,8 +120,7 @@ trigger_files = [
   "src/cargo/util/config/environment.rs",
 ]
 
-[autolabel."A-errors"]
-trigger_files = ["src/cargo/util/diagnostic_server.rs"]
```
ehuss, line 124: This looks like it deleted the trigger, was that intentional?

weihanglo: I feel like that `diagnostic_server.rs` should belong to `Command-fix`, no?

ehuss: Sorry, I was not clear. I was confused that this is adding this empty table:

```
[autolabel."A-diagnostics"]
```

for `A-diagnostics` without any keys. If that table is not needed, I'm not clear why it is being added (I wasn't sure if there was a `trigger_files` intended to be added here).

Feel free to `r=ehuss` either way.

weihanglo:
`A-errors` auto-label has been removed.

@bors r=ehuss

Issue #13246, opened 2024-01-03
Remove support for rustc plugins


bjorn3:
### Problem

Cargo still supports building and depending on rustc plugins despite this having been removed from rustc a while ago.

### Proposed Solution

Remove support for building rustc plugins.

### Notes

Support was removed in https://github.com/rust-lang/rust/pull/116412

ehuss:
I think it would be good, but I think we need to be a little cautious since I believe there are people abusing (or misunderstanding) the plugin setting to force building a dylib. We may need to do a deprecation notice for a little while.

bjorn3:
Opened https://github.com/rust-lang/cargo/pull/13248 for a deprecation warning.

PR #13248, opened 2024-01-04
Deprecate rustc plugin support in cargo

bjorn3:

Support for rustc plugins has been removed entirely from rustc itself already. Deprecate support on the cargo side to allow removing support from cargo in the future.

### Additional information

Part of https://github.com/rust-lang/cargo/issues/13246


weihanglo (approved):
Thanks!

PR #13253, opened 2024-01-05
Use atomic instead of mutex in `du` function

BD103:

This optimizes the disk usage (`du()`) function in cargo-util to use an atomic to track the total size instead of a mutex.

The `du` function recursively checks the amount of bytes a folder. It uses a parallel iterator to read the size of multiple files at once. The variable `total` is a `u64` that tracks this size. It was a mutex to prevent data races in the parallel iterator and make sure that the total size is accurately counted.

Because `total` is a simple number and is only being incremented in parallel, it can be turned into an [`AtomicU64`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html). Atomics take up less bytes than mutexes. It also uses builtin CPU instructions to instrument locking, instead of using a manual mutex. This means it is generally faster, but results can vary.

Atomics require an [`Ordering`](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html) to specify synchronization. After reading [Rust Atomics and Locks](https://marabos.nl/atomics/memory-ordering.html), I chose `Ordering::Relaxed`. The code only needs to synchronize one atomic. By the time the `walker.run` returns, all the threads will be closed and the atomic's value will be final. **If someone sees an issue with using `Relaxed` ordering, please say so now. The general goal is avoid data races. :)**

Data races are tricky to reproduce, which is why I would like the logic behind my change to be verified. Additionally, I noticed that `du` was being used in multiple other tests, like the global cache tracker, so I ran those as well.

I couldn't find any specific test that just tested `du`. I'd be willing to write one if it would be useful.

### Additional information

Most of my knowledge about atomics and parallelism come from [Rust Atomics and Locks](https://marabos.nl/atomics/). If recommend reading it if you want learn more about atomics and memory ordering.

### DIFF
--- a/crates/cargo-util/src/du.rs
+++ b/crates/cargo-util/src/du.rs
@@ -4,6 +4,7 @@ use anyhow::{Context, Result};
 use ignore::overrides::OverrideBuilder;
 use ignore::{WalkBuilder, WalkState};
 use std::path::Path;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 
 /// Determines the disk usage of all files in the given directory.
@@ -39,7 +40,7 @@ fn du_inner(path: &Path, patterns: &[&str]) -> Result<u64> {
         .git_ignore(false)
         .git_exclude(false);
     let walker = builder.build_parallel();
-    let total = Arc::new(Mutex::new(0u64));
+    let total = AtomicU64::new(0);
     // A slot used to indicate there was an error while walking.
     //
     // It is possible that more than one error happens (such as in different
@@ -51,8 +52,8 @@ fn du_inner(path: &Path, patterns: &[&str]) -> Result<u64> {
                 Ok(entry) => match entry.metadata() {
                     Ok(meta) => {
                         if meta.is_file() {
-                            let mut lock = total.lock().unwrap();
-                            *lock += meta.len();
+                            // Note that fetch_add may wrap the u64.
+                            total.fetch_add(meta.len(), Ordering::Relaxed);
                         }
                     }
                     Err(e) => {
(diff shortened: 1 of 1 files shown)


weihanglo (commented):
This was an intentional change from `Atomic64` for compatiblity reason. See <https://github.com/rust-lang/cargo/pull/12981> and <https://github.com/rust-lang/rust/pull/117916#issuecomment-1812635848>.

BD103:
My apologies, I did not realize it was originally an atomic. Would you like me to make the atomic conditional as mentioned in the related comment?

weihanglo:
> Would you like me to make the atomic conditional as mentioned in the related comment?

If the shim doesn't bring too much code and is easy to understand, it could work I guess :)

crates/cargo-util/src/du.rs
```diff
@@ -74,5 +94,9 @@ fn du_inner(path: &Path, patterns: &[&str]) -> Result<u64> {
         return Err(e);
     }
 
-    Ok(total.load(Ordering::Relaxed))
+    #[cfg(target_has_atomic = "64")]
+    return Ok(total.load(Ordering::Relaxed));
```
BD103, line 98: I needed to use a return here so it would compile. Rust doesn't know that the two `cfg` blocks are mutually exclusive.

ehuss:
Thanks for the PR!  However, I'm having trouble understanding why we would make this change. It appears to be quite a lot more code from before, which adds maintenance burden and complexity. In my performance testing, I was not able to determine a noticeable difference. Are there some benchmarks that indicate that this is an issue?

BD103:
> Thanks for the PR! However, I'm having trouble understanding why we would make this change. It appears to be quite a lot more code from before, which adds maintenance burden and complexity. In my performance testing, I was not able to determine a noticeable difference. Are there some benchmarks that indicate that this is an issue?

No, I just happened to be browsing the codebase and noticed that the `Arc<Mutex<u64>>` could be replaced with an atomic. I don't have any real-world numbers to justify this change. I was so excited to find some way I could contribute back to Cargo that I didn't try looking further into why the code used a mutex, nor how much it affected performance.

I'm going to close this, since the additional code complexity does not deem the small performance gain. I apologize for wasting your time.

Issue #4122, opened 2017-06-05
Enable strip in release mode by default


SuperCuber:
I'd expect `build --release` to apply all optimisations, and one of them would be file size.

I tried two programs, a trivial "hello rust" and a trivial "hello panic". Both of their sizes were 3.8M which is pretty crazy for something that is supposed to be optimised... So I tried `strip --strip-all`ing the executables.

Hello rust went to 372K
Hello panic went to 364K

That's a pretty huge change... The only difference in functionality I found is that `RUST_BACKTRACE=1` shows a lot of `<unknown>`s on `strip`ed files which is to be expected, but I think that's not a real issue because the whole point of `--release` is that it's not supposed to be debugged, there is the normal `build` for that...

For now I will be `strip --strip-all`ing all of my released executables but I really don't see a reason why `build --release` doesn't do it. I'd be happy to be enlightened if I missed anything though!

cedenday:
I think this should only change if `panic = 'abort'` to avoid the backtrace getting wiped.

SuperCuber:
@cedenday I don't see an issue with backtraces being wiped in `--release` if it's used in the way it needs to be used though. Also it gives the basic information of where the panic was (file and line number) after `strip`ing so it should give enough information to rebuild the project in debug mode and reproduce + backtrace the issue.

alexcrichton:
Thanks for the report! I think this is mostly due to the addition of debuginfo into the standard library a few versions ago, so the lion's share of `strip` is stripping out that debugging information. This does make sense to me to do by default! I'm not sure what the best place for that would be, though. We probably can't literally run `strip` because it's not very cross-platform so this would probably go into rustc assuming LLVM can handle it, but that'd likely involve a good amount of work.

mati865:
@alexcrichton maybe LLVM's `opt --strip-debug` could be handy?
https://llvm.org/docs/CommandGuide/opt.html

For ld/lld `--strip-all` can be used and it gives the best results.
In my tests on Arch Linux (fairly recent toolchains) `RUSTFLAGS="-C link-arg=-s" cargo build --release` gave ~3% better results than running `strip -s` on release binary.

matklad:
@mati865 we have a nightly rustc flag for that: https://github.com/rust-lang/rust/pull/49212. I think the idea is, eventually, to enable stripping for `--release` by default and flip the flag's meaning to *disable strip*.

jcbritobr:
Hello. When this will be solved?

[...]

Issue #11641, opened 2023-01-28
cargo failed to run custom build command if strip in m1 mac


Ylarod:
### Problem

cargo faild to run custom build command if 
```
[profile.release]
strip = true
```

command output
```
$ cargo build --release
   Compiling libc v0.2.139
error: failed to run custom build command for `libc v0.2.139`

Caused by:
  process didn't exit successfully: `/private/tmp/tet/target/release/build/libc-c9b37293d5307b87/build-script-build` (signal: 9, SIGKILL: kill)
```

### Steps

Cargo.toml
```toml
[package]
name = "tet"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
libc = "0.2"

[profile.release]
strip = true
```

main.rs
```rust
fn main() {
    println!("Hello, world!");
    unsafe{
        libc::printf("Hello, world!\n".as_ptr() as *const libc::c_char);
    }
}
```

then just run `cargo build --release`

### Notes

```
$ file $(which cargo)
Mach-O 64-bit executable arm64

$ file $(which rustc)
Mach-O 64-bit executable arm64

$ file $(which rustup)
Mach-O 64-bit executable arm64

$ rustup -V
rustup 1.25.1 (bb60b1e89 2022-07-12)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.67.0 (fc594f156 2023-01-24)`

$ rustc -V
rustc 1.67.0 (fc594f156 2023-01-24)

$ uname -a
Darwin YlarodMBP 22.1.0 Darwin Kernel Version 22.1.0: Sun Oct  9 20:15:09 PDT 2022; root:xnu-8792.41.9~2/RELEASE_ARM64_T6000 arm64

$ sw_vers
ProductName:		macOS
ProductVersion:		13.0.1
BuildVersion:		22A400
```

### Version

```text
cargo 1.67.0 (8ecd4f20a 2023-01-10)
release: 1.67.0
commit-hash: 8ecd4f20a9efb626975ac18a016d480dc7183d9b
commit-date: 2023-01-10
host: aarch64-apple-darwin
libgit2: 1.5.0 (sys:0.16.0 vendored)
libcurl: 7.84.0 (sys:0.4.59+curl-7.86.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
[...]

PR #13257, opened 2024-01-06
Strip debuginfo when debuginfo is not requested

Kobzol:

This PR implements [this proposal](https://github.com/rust-lang/cargo/issues/4122#issuecomment-1868318860). It contains a detailed description of the change.

As a summary, this PR modifies Cargo so that if the user doesn't set `strip` explicitly, and debuginfo is not enabled for any package being compiled, Cargo will implicitly set `strip = "debuginfo"`, to strip pre-existing debuginfo coming from the standard library. This reduces the default size of release binaries considerably (~4.5 MiB => ~450 KiB for helloworld on Linux x64).

Perhaps we could only add the `-Cstrip` option for the leaf/root target (i.e., a binary), but cargo already passes `-Cstrip` to libraries if it's set by `[profile.<...>.package.<lib>]`, so it seems harmless.

Fixes: https://github.com/rust-lang/cargo/issues/4122

Best reviewed commit by commit. There is one commit that fixes an existing related test that was using wrong assertion.

### Additional information

The implementation of the deferred option was inspired by `DebugInfo`, which already uses a similar concept.

### Unresolved questions
Should we also do this on macOS by default? It [seems](https://github.com/rust-lang/cargo/issues/11641) that there can be some issues with `strip` there. If it doesn't work, it basically inhibits compilation in release mode, with no easy way to opt out (unless the user explicitly requests debuginfo, but that's not the same as the previous state).

### DIFF
--- a/src/cargo/core/profiles.rs
+++ b/src/cargo/core/profiles.rs
@@ -643,7 +643,7 @@ impl Default for Profile {
             rpath: false,
             incremental: false,
             panic: PanicStrategy::Unwind,
-            strip: Strip::Resolved(StripInner::None),
+            strip: Strip::Deferred(StripInner::None),
             rustflags: vec![],
             trim_paths: None,
         }
@@ -925,6 +925,15 @@ impl Strip {
         match self {
             Strip::Resolved(v) | Strip::Deferred(v) => v,
         }
+    }
+
+    pub(crate) fn is_deferred(&self) -> bool {
+        matches!(self, Strip::Deferred(_))
+    }
+
+    /// Reset to stripping debuginfo.
+    pub(crate) fn strip_debuginfo(self) -> Self {
+        Strip::Resolved(StripInner::Named("debuginfo".into()))
     }
 }
 
--- a/src/cargo/core/profiles.rs
+++ b/src/cargo/core/profiles.rs
@@ -573,10 +573,17 @@ fn merge_profile(profile: &mut Profile, toml: &TomlProfile) {
         profile.trim_paths = Some(trim_paths.clone());
     }
     profile.strip = match toml.strip {
-        Some(StringOrBool::Bool(true)) => Strip::Named(InternedString::new("symbols")),
-        None | Some(StringOrBool::Bool(false)) => Strip::None,
-        Some(StringOrBool::String(ref n)) if n.as_str() == "none" => Strip::None,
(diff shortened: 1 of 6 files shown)


src/cargo/core/profiles.rs
```diff
@@ -873,28 +880,87 @@ impl fmt::Display for PanicStrategy {
...
+///
+/// Internally, it's used to model a strip option whose value can be deferred
+/// for optimization purposes: when no package being compiled requires debuginfo,
+/// then we can strip debuginfo to remove pre-existing debug symbols from the
+/// standard library.
+#[derive(Clone, Copy, Debug, Eq, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum Strip {
+    /// A strip option that is fixed and will not change.
+    Resolved(StripInner),
```
Kobzol, line 916: Maybe it would make sense to introduce some kind of generic `Resolved/Deferred` wrapper and use it both here and in `DebugInfo`?

Kobzol:
The FCP has been [accepted](https://github.com/rust-lang/cargo/issues/4122#issuecomment-1891039373).

ehuss:
Thanks!

@bors r+

tests/testsuite/profiles.rs
```diff
@@ -607,7 +607,82 @@ fn strip_accepts_false_to_disable_strip() {
...
+                debug = 1
+            "#,
+        )
+        .file("src/main.rs", "fn main() {}")
+        .file(
+            "bar/Cargo.toml",
+            r#"
+                [package]
+                name = "bar"
+                verison = "0.1.0"
```
afraca, line 678: This has a typo by the way. 
I'm not sure if you ever see this comment

Kobzol: Oh, good catch! I thought that this would be an error, but it's just a warning. Fixed in https://github.com/rust-lang/cargo/pull/13342.

utkarshgupta137:
I think this is worth mentioning in the docs too?

Kobzol:
I would say that the docs already mentioned this (`debug = 0` claimed that there will be no debuginfo in the target binary). It just wasn't upheld before, and it is now. OTOH, we could maybe document the exact mechanism that is used to achieve this.

PR #13260, opened 2024-01-08
docs(changelog): Call out cargo-new lockfile change

epage:
I was looking for what release this happened in but we didn't have it listed.
We do list the documentation change.
This was likely from the PR focusing on the entire policy change which made it easy to overlook each aspect of the policy change.


ehuss:
Thanks!

@bors r+

Issue #13235, opened 2024-01-02
cargo init or cargo new Cargo.toml don't respect `workspace exclude`


kzhui125:
### Problem

when I have a workspace, exclude `data`:

```toml
[workspace]
resolver = "2"
members = ["crates/*"]
exclude = ["data"]
```

then I create a new crate in data folder:

The Cargo.toml will be:

```toml
version.workspace = true
edition.workspace = true
```

instead of

```toml
version = "0.1.0"
edition = "2021"
```

I have to change it manually.

hi-rustin:
Could you please provide your Cargo version?

I can not reproduce this issue. My Cargo version:
```console
❯ cargo -Vv
cargo 1.77.0-nightly (ac6bbb332 2023-12-26)
release: 1.77.0-nightly
commit-hash: ac6bbb33293d8d424c17ecdb42af3aac25fb7295
commit-date: 2023-12-26
host: aarch64-apple-darwin
libgit2: 1.7.1 (sys:0.18.1 vendored)
libcurl: 8.4.0 (sys:0.4.70+curl-8.5.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Mac OS 14.2.1 [64-bit]
```

kzhui125:
maybe nightly cargo fixed this issue?

```
cargo 1.75.0 (1d8b05cdd 2023-11-20)
release: 1.75.0
commit-hash: 1d8b05cdd1287c64467306cf3ca2c8ac60c11eb0
commit-date: 2023-11-20
host: aarch64-apple-darwin
libgit2: 1.7.1 (sys:0.18.1 vendored)
libcurl: 8.4.0 (sys:0.4.68+curl-8.4.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1u  30 May 2023
os: Mac OS 14.2.1 [64-bit]
```

kzhui125:
It seems another issue.

cargo init will add `members = ["data/a1"]` to workspace Cargo.toml

I have to remove the `members = ["data/a1"]`

```toml
[workspace]
resolver = "2"
exclude = ["data"]
members = ["data/a1"]
```

hi-rustin:
I changed my Cargo version:

```console
❯ cargo -Vv
cargo 1.75.0 (1d8b05cdd 2023-11-20)
release: 1.75.0
[...]

PR #13261, opened 2024-01-08
fix: only inherit workspace package table if the new package is a member

hi-rustin:

part of https://github.com/rust-lang/cargo/issues/13235

We should not inherit from the workspace if the new package is in the exclude list.

- [x] Updated the test case to cover it.
- [x] Do not inherit the workspace package table if it is not a member of the workspace.

Check out the unit tests.

### Additional information

None

### DIFF
--- a/src/cargo/ops/cargo_new.rs
+++ b/src/cargo/ops/cargo_new.rs
@@ -816,11 +816,8 @@ fn mk(config: &Config, opts: &MkOptions<'_>) -> CargoResult<()> {
         // This should not block the creation of the new project. It is only a best effort to
         // inherit the workspace package keys.
         if let Ok(mut workspace_document) = root_manifest.parse::<toml_edit::Document>() {
-            let (display_path, can_be_a_member) = get_display_path_and_check_membership(
-                &root_manifest_path,
-                &path,
-                &workspace_document,
-            )?;
+            let display_path = get_display_path(&root_manifest_path, &path)?;
+            let can_be_a_member = can_be_workspace_member(&display_path, &workspace_document)?;
             // Only try to inherit the workspace stuff if the new package can be a member of the workspace.
             if can_be_a_member {
                 if let Some(workspace_package_keys) = workspace_document
@@ -1003,11 +1000,7 @@ fn update_manifest_with_new_member(
     )
 }
 
-fn get_display_path_and_check_membership(
-    root_manifest_path: &Path,
-    package_path: &Path,
-    workspace_document: &toml_edit::Document,
-) -> CargoResult<(String, bool)> {
+fn get_display_path(root_manifest_path: &Path, package_path: &Path) -> CargoResult<String> {
     // Find the relative path for the package from the workspace root directory.
     let workspace_root = root_manifest_path.parent().with_context(|| {
         format!(
@@ -1031,7 +1024,14 @@ fn get_display_path_and_check_membership(
         components.push(comp);
     }
     let display_path = components.join("/");
+    Ok(display_path)
+}
 
+// Check if the package can be a member of the workspace.
+fn can_be_workspace_member(
+    display_path: &str,
+    workspace_document: &toml_edit::Document,
+) -> CargoResult<bool> {
     if let Some(exclude) = workspace_document
         .get("workspace")
         .and_then(|workspace| workspace.get("exclude"))
@@ -1042,10 +1042,9 @@ fn get_display_path_and_check_membership(
                 .as_str()
                 .with_context(|| format!("invalid non-string exclude path `{}`", member))?;
             if pat == display_path {
-                return Ok((display_path, false));
+                return Ok(false);
             }
         }
     }
-
-    Ok((display_path, true))
+    Ok(true)
 }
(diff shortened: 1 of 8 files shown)


hi-rustin (commented):
🔢 Self-check

src/cargo/ops/cargo_new.rs
```diff
@@ -955,8 +963,52 @@ fn update_manifest_with_inherited_workspace_package_keys(
...
+        workspace_document["workspace"]["members"] = toml_edit::value(array);
+    }
+
+    write_atomic(
+        &root_manifest_path,
+        workspace_document.to_string().to_string().as_bytes(),
+    )
+}
+
+fn get_display_path_and_check_exclusion(
```
epage, line 1007: `and`s in function names generally imply too much is being coupled together.  What would it look like if we made this two separate functions?

And if you are making changes, can you put the refactor into a separate commit?

src/cargo/ops/cargo_new.rs
```diff
@@ -955,8 +963,52 @@ fn update_manifest_with_inherited_workspace_package_keys(
 fn update_manifest_with_new_member(
     root_manifest_path: &Path,
     workspace_document: &mut toml_edit::Document,
-    package_path: &Path,
+    display_path: String,
```
epage, line 966: nit: to make the caller more refactor friendly, this should probably be a `&str`

src/cargo/ops/cargo_new.rs
```diff
@@ -816,18 +816,24 @@ fn mk(config: &Config, opts: &MkOptions<'_>) -> CargoResult<()> {
...
                 .get("workspace")
                 .and_then(|workspace| workspace.get("package"))
                 .and_then(|package| package.as_table())
             {
-                update_manifest_with_inherited_workspace_package_keys(
-                    opts,
-                    &mut manifest,
-                    workspace_package_keys,
-                )
+                if !exclude {
```
epage, line 829: Double-negatives aren't ideal.  Could we change the return value to `is_workspace_member`?

epage, line 829: Why is the `!exclude` check inside of the `if`?  Seems like from the perspective of trying to describe this code, we shouldn't try to look stuff up in the workspace if this is a workspace member.

Putting it on the outside will also make it easier to avoid bugs as we add new sections to inherit

src/cargo/ops/cargo_new.rs
```diff
@@ -839,12 +845,14 @@ fn mk(config: &Config, opts: &MkOptions<'_>) -> CargoResult<()> {
                 manifest["lints"] = toml_edit::Item::Table(table);
```
epage, line 845: We are not excluding `workspace.lints`

epage:
> part of #13235

If this doesn't close that issue, then what is left after this?

src/cargo/ops/cargo_new.rs
hi-rustin: Moved out.

hi-rustin:
GitHub is down.

src/cargo/ops/cargo_new.rs
hi-rustin: Separated

hi-rustin:
GitHub is down.

> If this doesn't close that issue, then what is left after this?

This is another bug in that issue. 

```console
❯ tree . --gitignore
.
├── Cargo.toml
├── crates
│   └── crate1
│       ├── Cargo.toml
│       └── src
│           └── main.rs
├── data
│   └── test2
│       ├── Cargo.toml
│       └── src
│           └── main.rs
└── src
    └── main.rs

8 directories, 6 files
```

The root workspace:

```toml
[package]
name = "test1"
version = "0.1.0"
edition = "2021"

[workspace]
resolver = "2"
members = ["crates/*", "data/test2"]
exclude = ["data/**"]


[workspace.package]
version = "0.1.0"

[dependencies]

```

If you create a package inside the data directory, it will inherit the version from the workspace and add the package into the members list, even if it should be ignored by the exclude glob.

src/cargo/ops/cargo_new.rs
```diff
@@ -993,46 +1042,9 @@ fn update_manifest_with_new_member(
                 .as_str()
                 .with_context(|| format!("invalid non-string exclude path `{}`", member))?;
             if pat == display_path {
```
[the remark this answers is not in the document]
epage: According to #6009, we don't support globs on exclude...

epage:
@bors r+

This fixes a definitive problem and we can work on the question of glob excludes separate from this

src/cargo/ops/cargo_new.rs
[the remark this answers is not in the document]
epage: Is the problem with globs or that `exclude` is a path prefix?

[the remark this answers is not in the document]
hi-rustin: > According to #6009, we don't support globs on exclude...

Oh, I forgot that 🤣 



> Is the problem with globs or that `exclude` is a path prefix?

No. It works well.

PR #13262, opened 2024-01-08
fix(manifest): Provide unused key warnings for lints table 

### LINKED ISSUE #12917: Warn on unknown keys in lint table entries
### Problem

In https://github.com/rust-lang/rust-clippy/issues/11751 a user had

```toml
[lints.clippy]
doc_markdown = { level = "deny", doc_valid_idents = ["GHz", ".."] }
```

Expecting the `doc_valid_idents` to set the clippy configuration of the same name, however it is silently ignored by cargo

### Steps

_No response_

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

_No response_

epage:

The use of `flatten` was getting in the way of `serde_ignored`.
A common workaround is to add our own `unused` tracking but that would
cause duplicates with `workspace.lints` (or we'd just ignore it).

Since the manual deserializer was relatively simple, I went that route.

Fixes #12917

Per commit

A test was added for the issue.  I then was worried about regressions in `workspace = false` errors (and I was right) so I added a test for that.  To get `workspace = false` to work nicely, I made it share code with other `workspace: bool` fields.

### Additional information

### DIFF
--- a/crates/cargo-util-schemas/src/manifest.rs
+++ b/crates/cargo-util-schemas/src/manifest.rs
@@ -410,11 +410,7 @@ impl<'de> de::Deserialize<'de> for InheritableBtreeMap {
         if let Ok(w) = TomlInheritedField::deserialize(
             serde_value::ValueDeserializer::<D::Error>::new(value.clone()),
         ) {
-            return if w.workspace {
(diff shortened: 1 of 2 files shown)


ehuss:
Nice!

@bors r+

Issue #13258, opened 2024-01-07
Running 'cargo new' with 'workspace.package.name' set creates an invalid manifest


TheNeikos:
### Problem

I have a workspace that has `workspace.package.name` set. I am aware that it is not a supported key.

When creating a new crate with `cargo new` it will set `package.name.workspace = true`. This is not supported, and will fail.

### Steps

1. Workspace with `workspace.package.name` set to a string
2. `cargo new`
3. Observe it fail to build

### Possible Solution(s)

Special case 'name' to not be inherited, even if it is set in the workspace?

### Version

```text
cargo 1.75.0 (1d8b05cdd 2023-11-20)
release: 1.75.0
commit-hash: 1d8b05cdd1287c64467306cf3ca2c8ac60c11eb0
commit-date: 2023-11-20
host: x86_64-unknown-linux-gnu
libgit2: 1.7.1 (sys:0.18.1 vendored)
libcurl: 8.4.0-DEV (sys:0.4.68+curl-8.4.0 vendored ssl:OpenSSL/1.1.1u)
ssl: OpenSSL 1.1.1u  30 May 2023
os: NixOS 24.5.0 [64-bit]
```

TheNeikos:
Oh I forgot to mention, but this may be a separate issue, but the error is cryptic to what the problem actually is:

```
error: failed to load manifest for workspace member `/home/neikos/projects/test/crates/builder`

Caused by:
  failed to parse manifest at `/home/neikos/projects/test/crates/builder/Cargo.toml`

Caused by:
  TOML parse error at line 1, column 1
    |
  1 | [package]
    | ^^^^^^^^^
  invalid type: map, expected a string
  ```

Muscraft:
The reason this is happening is that `cargo new`  [copies `[workspace.package]` if it exists](https://github.com/rust-lang/cargo/blob/13e97927cc98dd7c325fd5fb21bf527fee2bce3d/src/cargo/ops/cargo_new.rs#L819-L822) and then [tries to add each key to `[package]`](https://github.com/rust-lang/cargo/blob/13e97927cc98dd7c325fd5fb21bf527fee2bce3d/src/cargo/ops/cargo_new.rs#L936-L946). In this case, `workspace.package.name` is being copied as it is set within `workspace.package`. Fixing this would require `cargo` to copy only certain keys, but keeping that list in sync with what can be inherited could be tough.

[...]

PR #13263, opened 2024-01-08
test(manifest): Verify we warn on unused workspace.package fields

epage:
I assumed from #13258 that we didn't warn but apparently we do.  Figured it'd still be good to keep the test around.


ehuss:
Thanks!

@bors r+

PR #13264, opened 2024-01-08
crates-io: Set `Content-Type: application/json` only for requests with a body payload

Turbo87:

The `Content-Type` **request** header is only supposed to be used if the request comes with a body payload. `cargo` is currently sending it unconditionally, even for `GET` requests that typically don't have a payload attached to them.

This PR changes the implementation to only attach the header if the request has a body payload.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -613,7 +613,7 @@ dependencies = [
 
 [[package]]
 name = "crates-io"
-version = "0.39.2"
+version = "0.40.0"
 dependencies = [
  "curl",
  "percent-encoding",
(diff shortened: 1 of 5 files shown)


weihanglo (approved):
Could you bump a patch version for `crates-io` package?

I am okay with this change and would be surprised if it breaks other registries, though seems like the RFC doesn't say anything about whether an empty body can have a `Content-Type` header.

Turbo87:
> Could you bump a patch version for `crates-io` package?

since it could be seen as a breaking change I opted for a minor version bump. let me know if that's okay with you or if you want me to change it :)

weihanglo (approved):

PR #13267, opened 2024-01-08
refactor(toml): Make it more obvious to update package-dependent fields

epage:
Inspired by my having forgotten to add `[lints]` to the if sequence. Previously, we added a comment to suggest this but the further the code is, the harder it is to track.

I considered a custom `Deserialize` impl, possibly through a new type, that would error.
This would be the more "pure" solution.
Unfortunately, this would also have worse errors because the errors would be reported to the `Deserializer` at the document-level, rather than directly on the individual fields.
Well, we don't do on individual fields now but it is something we will soon be exploring.


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

Issue #10310, opened 2022-01-20
cargo vendor can't handle duplicates.


YellowOnion:
### Problem

```
Caused by:
  found duplicate version of package `xi-rope v0.3.0` vendored from two sources:

        source 1: registry `crates-io`
        source 2: https://github.com/lapce/xi-editor#5578c725
```

### Possible Solution(s)

reimplement `--no-merge-sources` from `cargo-vendor`

### Notes

Makes building packages on NixOS a pain to maintain:
https://github.com/NixOS/nixpkgs/issues/30742

Eh2406:
What were you doing when you got this error? How did you have things set up?
Can you construct a minimum complete and reproducible example for us too get the same error, if not why not?

weihanglo:
This repo [terminal-discord/weechat-discord](https://github.com/terminal-discord/weechat-discord/tree/d435c276dbbfe4fbfa6cc10d559e336d951b32f2) is reproducible.

And this is a reproducible test case:

```rust
    registry::alt_init();
    let p = project()
        .file(
            "Cargo.toml",
            r#"
                [project]
                name = "foo"
                version = "0.0.1"
                authors = []


                [dependencies]
                bar = "0.0.1"

                [dependencies.foo]
                package = "bar"
                version = "0.0.1"
                registry = "alternative"


            "#,
        )
        .file("src/main.rs", "fn main() {}")
        .build();

    Package::new("bar", "0.0.1").publish();
    Package::new("bar", "0.0.1").alternative(true).publish();

    p.cargo("build").run();
    p.cargo("vendor --respect-source-config")
        .with_stderr(
            "\
[ERROR] failed to sync

Caused by:
  found duplicate version of package `bar v0.0.1` vendored from two sources:

  <tab>source 1: registry `alternative`
  <tab>source 2: registry `crates-io`
",
        )
        .with_status(101)
        .run()
```

weihanglo:
Unfortunately, `cargo-vendor vendor --no-merge-sources` is no longer a fallback. It can't recognize packages with either 2021 edition or lockfile v2/v3.

When running under a project with 2018 edition and lockfile v1, `cargo-vendor vendor` got

```
[...]

PR #10344, opened 2022-01-31
Auto-switch to multi-sources vendor directory layout

weihanglo:
### What does this PR try to resolve?

`cargo-vendor` has been broken for a while, there is no a clear alternative way to solve the situation described in #10310. This PR tries to do what `--no-merge-sources` does but automatically switching to multi-sources layout instead of introducing the flag. (see https://github.com/rust-lang/cargo/pull/10344#pullrequestreview-868113795 and https://github.com/rust-lang/cargo/pull/10344#issuecomment-1032994506)

Fixes #10310

### How should we test and review this PR?

Part of the logic is copied from https://github.com/alexcrichton/cargo-vendor.

Several tests are updated and added:

- 🆕 `vendor::duplicate_version_from_multiple_sources`: Auto-switch between non-merged and merged sources.
- ❌ `vendor::git_duplicate`: Removed. cargo-vendor now can auto-switch.
- ❌ `vendor::vendor_sample_config`: Merged into `vendor::vendor_simple`.

### Additional information

There are somethings I am uncertain:

- [x] When switching between non-merged and merged sources. `cargo-vendor` removes the entire vendor directory. Should cargo emit a warning or just error out and tell user the incompatibility between merged and non-merged? Generally removing the vendor dir should not be a destructive operation but thing not always goes as we thought 😆 
**Solved**: See https://github.com/rust-lang/cargo/pull/10344#discussion_r795852617
- [x] The `cargo::util::short_hash` is not compatible with [the one in cargo-vendor]. I personally prefer to use `util::short_hash` instead, but if the compatibility is more important I am also ok to copy it over.
**Solved**: See https://github.com/rust-lang/cargo/pull/10344#discussion_r796996824

[the one in cargo-vendor]: https://github.com/alexcrichton/cargo-vendor/blob/07570e23/src/main.rs#L568-L575

PR #13271, opened 2024-01-09
Fix: cargo vendor can't handle duplicates.

junjihashimoto:

part of #10310
Add `--no-merge-sources` options for cargo-vendor to handle duplicates.

Check out the unit tests.

### Additional information

None.


weihanglo (commented):
Could you brief the design in the PR description? such as how the layout of vendor directory look like, and the difference between with and without `--no-merge-sources`?

Note that the flag was proposed but at that time the Cargo team prefered to do it automatically without a flag: <https://github.com/rust-lang/cargo/pull/10344#pullrequestreview-868113795>

junjihashimoto:
In the case of automation, the use cases may be different and are not compatible.

junjihashimoto:
If it is better to use https://github.com/rust-lang/cargo/pull/10344 as a base, I will consider that.

junjihashimoto:
@weihanglo Do you actually have a schedule to fix this issue?

weihanglo:
For myself, I would wait for FCP in <https://github.com/rust-lang/rfcs/pull/3243#issuecomment-1879610048> and see if we can avoid adding an extra flag. We can perhaps think ahead how the layout should be like with namespaced package feature. 

That being said, if there is any proposal with better UX, feel free to share.

(General design discussions are expected to happen in the issue #10310 and leave PR for specific implementation btw)

junjihashimoto:
@weihanglo Thank you for pointing it out.
https://github.com/rust-lang/rfcs/pull/3243#issuecomment-1879610048 looks like the individual packages themselves need to be modified with using namespaces when using the feature.
If such a fix is possible, patches like this PR are not needed.
This PR is effective when the dependencies of the package we are using are complex and it is difficult to adjust the version.

haraldh:
@junjihashimoto Thank you very much! This has been a "live" saver.

ehuss:
I'm not quite clear, why does this need a new flag? Why can't it just use separate directories when there is a conflict?

junjihashimoto:
I think it's due to different use cases.
With the current default, if a user wants to find a conflicting package, it can be detected as an error.
Automatic collision avoidance will not find conflicting packages.

I would like to be able to build with automatic collision avoidance.
Should I change to that?

weihanglo:
> With the current default, if a user wants to find a conflicting package, it can be detected as an error.
Automatic collision avoidance will not find conflicting packages.

In this case they should use tools like cargo-deny instead of cargo vendor.

im-0:
If anyone still needs it, here is the rebased commit: https://github.com/im-0/cargo/commit/c8524ebd42621131a6b776a69286c52f81f27c19

wiiznokes:
> In this case they should use tools like cargo-deny instead of cargo vendor.

i agree, if `cargo build` handle it by default, `cargo vendor` should do the same.

PR #13272, opened 2024-01-09
doc: add `public` info in `cargo-add` man page.

linyihai:
follow up https://github.com/rust-lang/cargo/pull/13046
add `public/private` explanation for `cargo-add` in man page.

### Additional information

The help info would be like this

- `cargo help add`
```
        --public
           Mark the dependency as public.

           The dependency can be referenced in your library’s public API.

           Unstable (nightly-only) <https://doc.rust-lang.org/cargo/reference/unstable.html#public-dependency>

       --no-public
           Mark the dependency as private.

           While you can use the crate in your implementation, it cannot be referenced in your public API.

           Unstable (nightly-only) <https://doc.rust-lang.org/cargo/reference/unstable.html#public-dependency>
```
- `cargo add -h`
```
      --public               Mark the dependency as public (unstable)
      --no-public            Mark the dependency as private (unstable)
```

- `cargo add --help`
```
      --public
          Mark the dependency as public (unstable)
          
          The dependency can be referenced in your library's public API.

      --no-public
          Mark the dependency as private (unstable)
          
          While you can use the crate in your implementation, it cannot be referenced in your public API.
```

### DIFF
--- a/src/bin/cargo/commands/add.rs
+++ b/src/bin/cargo/commands/add.rs
@@ -62,17 +62,17 @@ The package name will be exposed as feature of your crate.")
 The package will be removed from your features.")
                 .conflicts_with("dev")
                 .overrides_with("optional"),
-            flag("public", "Mark the dependency as public")
+            flag("public", "Mark the dependency as public (unstable)")
                 .conflicts_with("dev")
                 .conflicts_with("build")
-                .long_help("Mark the dependency as public
+                .long_help("Mark the dependency as public (unstable)
 
 The dependency can be referenced in your library's public API."),
-            flag("no-public", "Mark the dependency as private")
+            flag("no-public", "Mark the dependency as private (unstable)")
                 .conflicts_with("dev")
                 .conflicts_with("build")
                 .overrides_with("public")
-                .long_help("Mark the dependency as private
+                .long_help("Mark the dependency as private (unstable)
 
 While you can use the crate in your implementation, it cannot be referenced in your public API."),
             clap::Arg::new("rename")
--- a/src/doc/man/generated_txt/cargo-add.txt
+++ b/src/doc/man/generated_txt/cargo-add.txt
@@ -96,6 +96,23 @@ OPTIONS
            Mark the dependency as required
            <https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies>.
 
+       --public
+           Mark the dependency as public.
+
+           The dependency can be referenced in your library’s public API.
+
+           Unstable (nightly-only)
+           <https://doc.rust-lang.org/cargo/reference/unstable.html#public-dependency>
+
+       --no-public
+           Mark the dependency as private.
+
+           While you can use the crate in your implementation, it cannot be
+           referenced in your public API.
+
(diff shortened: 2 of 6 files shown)


src/doc/man/generated_txt/cargo-add.txt
```diff
@@ -96,6 +96,12 @@ OPTIONS
            Mark the dependency as required
            <https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies>.
 
+       --public
+           Mark the dependency as public.
```
weihanglo, line 100: Should we flag this as "nightly only", and have a link to unstable doc?

(same thing also applies to `cargo add -h/--help`?)

epage: Somehow I thought of this and forgot it all in the time it took for me to look at this and approve it...

linyihai: Ok，I will add the "nightly only" and the link latter.

src/bin/cargo/commands/add.rs
```diff
@@ -67,14 +67,18 @@ The package will be removed from your features.")
                 .conflicts_with("build")
                 .long_help("Mark the dependency as public
 
-The dependency can be referenced in your library's public API."),
+The dependency can be referenced in your library's public API.
+
+Unstable (nightly-only) <https://doc.rust-lang.org/cargo/reference/unstable.html#public-dependency>"),
```
epage, line 72: The short help should have a trailing ` (unstable)`

Personally, I wouldn't add more than that to the long help.  Being an unstable feature, we likely don't want to draw more attention to it which an extra long help section would do

linyihai: > The short help should have a trailing  (unstable)
I think it's better this way.

src/bin/cargo/commands/add.rs
```diff
@@ -65,14 +65,14 @@ The package will be removed from your features.")
             flag("public", "Mark the dependency as public")
```
weihanglo, line 65: `(unstable)` for short help as well? Same as `ignore-rust-version`

src/bin/cargo/commands/add.rs
linyihai: Thanks for you remind. 

(I totaly misunderstand the `short help`)

linyihai: Is anything need to tweak?

The finall help output is:

- cargo add --help 
```
      --public
          Mark the dependency as public (unstable)
          
          The dependency can be referenced in your library's public API.

      --no-public
          Mark the dependency as private (unstable)
          
          While you can use the crate in your implementation, it cannot be referenced in your public API.
```

- cargo help add

```
       --public
           Mark the dependency as public.

           Unstable (nightly-only) <https://doc.rust-lang.org/cargo/reference/unstable.html#public-dependency>

       --no-public
           Mark the dependency as private.

           Unstable (nightly-only) <https://doc.rust-lang.org/cargo/reference/unstable.html#public-dependency>

```

epage:
Thanks!

@bors r+

PR #13273, opened 2024-01-09
Document why `du` function uses mutex

BD103:

After closing #13253, it [was suggested](https://github.com/rust-lang/cargo/pull/13253#issuecomment-1883286035) to document why the `du` function uses a `Mutex` instead of an `AtomicU64`. This will prevent others from making the same mistake I did. :)

N/A

### Additional information

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -414,7 +414,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-util"
-version = "0.2.9"
+version = "0.2.10"
 dependencies = [
  "anyhow",
  "core-foundation",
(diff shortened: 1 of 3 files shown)


weihanglo:
The version bump check failed. Could you bump cargo-util to 0.2.10 here? Thank you.

https://github.com/rust-lang/cargo/blob/529130fe54d0194ab1315ce11d4e3ac537d8c4d8/crates/cargo-util/Cargo.toml#L3

BD103:
I had some confusion with the CI, but it should pass now. :)

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

Issue #4614, opened 2017-10-13
enhacement - expose manifest.rs as standalone crate


sagiegurari:
The manifest.rs has the structs and enums that define most of the Cargo.toml
That could be really helpful for other build/dev tools like [cargo-make](https://sagiegurari.github.io/cargo-make/)
For now, I also have internally mapped the relevant items in an internal data structs and enums, but would be happier if I would have just had a standard crate to use to pull those definitions from (naturally with serde support).
The manifest.rs seems like exactly the thing I was looking for but I would prefer a standalone lightweight crate only holding definitions instead of depending on entire cargo crate.
I would also help making your codebase more modular and reusable.

alexcrichton:
Sounds like a good idea to me!

carols10cents:
cc @joshtriplett, is this the toml crate you were talking about in the cargo meeting today or is that different?

joshtriplett:
@carols10cents This is different, but also very useful. I'd love to have this available for all sorts of metabuild scripts that want to parse `Cargo.toml`.

sagiegurari:
bump... just wondering if this is on any todo list...

epage:
We've now closed #12801 by providing [cargo_util_schemas::manifest::TomlManifest](https://docs.rs/cargo-util-schemas/latest/cargo_util_schemas/manifest/struct.TomlManifest.html).

This does not do any high level processing, like inheriting from a workspace, reading defaults, or complex validation.  This Issue is remaining open for that work.  The challenge is a lot of that logic is more tightly coupled to the internals of cargo (see #12801 for how much work it was just to do serde types).  Personally, I don't see this Issue being resolved but I'm open to being surprised.

weihanglo:
## Problem

We have `cargo-util-schemas` for low-level type definitions in Cargo, and [`cargo_util_schemas::manifest::TomlManifest`](https://doc.rust-lang.org/nightly/nightly-rustc/cargo_util_schemas/manifest/struct.TomlManifest.html) is the core one for serialization and deserialzation for `Cargo.toml`. However, since it is only a type definition, there is no high-level Cargo business logic in the library, which people still need to copy from or reimplement Cargo.

There are multi-layer needs people like to see in a `cargo-util-manifest` package. Here is a list of these layer: 

* Populates the default value for each field (requiring workspace-awareness)
* Populates the values of fields that depend on filesystem probing.
[...]

PR #13278, opened 2024-01-10
refactor: factor out `Config` from `core::features`

weihanglo:

To decouple manifest logic from `config.toml` behavior, a trait
`UnstableFeatureContext` is created for `Features` to query any
necessary information from the outer context.

This is a path toward a new package for high-level manifest processing
logic. The goal is to break out behavior for `Config` into several
small contexts, which can be imported when needed.

There is no functional change.

The plan is to add more context impls into `util::config::context`.

### Additional information

Related to #4614

### DIFF
--- a/src/cargo/core/features.rs
+++ b/src/cargo/core/features.rs
@@ -148,7 +148,6 @@ use serde::{Deserialize, Serialize};
 use crate::core::resolver::ResolveBehavior;
 use crate::util::errors::CargoResult;
 use crate::util::indented_lines;
-use crate::Config;
 
 pub const SEE_CHANNELS: &str =
     "See https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information \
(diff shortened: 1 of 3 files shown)


epage:
Could you lay out your motivation and plan in #4614?  I'm a bit concerned that #4614 is a large, invasive project that will affect maintainability with limited gains.

Example concerns
- Manifest dependent on `core` types
- The parsing and prepare-for-publish logic seem highly related but the abstraction for the latter is very leaky
- More contexts being required by manifests, like for diagnostics (can't `+` traits with `dyn` but can with `impl`)
- Filesystem interactions
- Layering the packages so a manifest parsing package has access to the relevant context traits and the contexts have access to their needed types

src/cargo/core/features.rs
```diff
@@ -598,7 +598,7 @@ impl Features {
                 see_docs()
             ),
             Status::Unstable => {
-                if let Some(allow) = &config.cli_unstable().allow_features {
+                if let Some(allow) = allow_features {
```
epage, line 601: imo this change loses some important context of which set of allowed features this is.  This is specifically the `-Z` / CLI features.  Maybe `cli_allow_features` for the trait method and variable??

src/cargo/util/config/contexts.rs
```diff
@@ -0,0 +1,13 @@
+use crate::core::features::AllowFeatures;
+use crate::core::features::UnstableFeatureContext;
+use crate::Config;
+
+impl UnstableFeatureContext for Config {
```
epage, line 5: I'm not usually a fan of trait impls being separate from either the trait or the struct definition.

epage: (`annotate-snippets` took this to an extreme)

weihanglo:
Close as this is not relevant at this moment. See <https://github.com/rust-lang/cargo/issues/4614#issuecomment-1912797097> for possible directions forward.

Issue #10729, opened 2022-06-04
`cargo add` refuses to add new non-vendored packages due to source replacement


k3d3:
### Problem

When a vendored directory is established, `cargo add` no longer adds new packages. Instead, it tries to translate a package name into a package that already exists in the vendored directory.

In this case, I've added `xml-rs`, then proceeded to vendor my dependencies. Now, no matter what I try to `cargo add`, it always tries to add `xml-rs`.

I've even tried to run `cargo +beta add cbindgen --registry crates-io, and it still gives me the same behaviour. The only way I can get `cargo add` to work is to remove `.cargo/config` and place it back afterwards.

![image](https://user-images.githubusercontent.com/1247908/171969035-42f5fbbe-bd31-42a9-8487-a173331d09bb.png)

### Steps

```sh
cargo new example
cd example
cargo +beta add xml-rs  # The same behaviour is exhibited in +nightly as well
cargo vendor ./vendor

# (add the output to .cargo/config.toml)

cargo +beta add --build cbindgen  # Doesn't need to be a build dependency, and "cbindgen" can be anything, even crates that don't exist
cargo +beta add cbindgen --registry crates-io  # This also does not work
```

### Possible Solution(s)

I assume the reason for this is because what's added in `.cargo/config.toml` is source replacement, meaning any dependencies that try to access `crates-io` will access the vendored directory instead. I think I understand *why* `cargo add` does what it does here, however I feel it's fairly unintuitive combined with vendoring.

I haven't figured out any way to work around this (at least, in a way that I can continue to use `cargo add`). I've tried to add to config.toml:

```toml
[source.actual-crates]
registry = "https://github.com/rust-lang/crates.io-index"
```
(I've also tried git in place of https, and adding .git to the end of the URL)

and if I try to use `cargo add cbindgen --registry actual-crates`, I get:

```
error: source `actual-crates` defines source registry `crates-io`, but that source is already defined by `crates-io`
note: Sources are not allowed to be defined multiple times.
```

If I put the following into cargo.toml:
```toml
[registries]
actual-crates = { index = "https://github.com/rust-lang/crates.io-index" }
```

and run the same `cargo add` command, I get the same behaviour:
```
warning: translating `cbindgen` to `xml-rs`
      Adding xml-rs v0.8.4 to dependencies.
```

For now, I can easily add my dependencies by modifying `Cargo.toml` myself, however the `cargo add` command is very convenient, especially when introducing Rust to a team that's new to the language.

### Notes

[...]

PR #13281, opened 2024-01-11
fix(add): Improve error when adding registry packages while vendored

LuuuXXX:
### **What does this PR try to resolve?**

When a vendored directory is established, cargo add no longer adds new packages. Instead, it tries to translate a package name into a package that already exists in the vendored directory.
[More details](https://github.com/rust-lang/cargo/issues/10729#issue-1260548746)

Since @epage has done most of the work, here I do the rest of the finishing work.

Improves the error from #10729

### **How should we test and review this PR?**

The implementation procedure is as follows:
https://github.com/rust-lang/cargo/issues/10729#issuecomment-1191633351

Test steps：
1. Try to get an arbitrary crate and execute `cargo vendor` command.
2. Configure the vendor directory in .cargo/config.toml.
3. Add `alter-registry` to the config.toml file.
```
[registries]
alter-registry= { index = "XXX" }
```
4. run the same `cargo add` command.
```
cargo add another-crate --registry alter-registry
```


epage:
I recommend re-arranging the commits so the first commit shows the current behavior via a test and then the commit with the fix updates the test.  This helps make clear to the reviewers what the behavior change was and ensures we "test the test" (assuming you ran the tests).

I tried to do that with your branch and when i cherry-picked the test onto `master`, it passed.

LuuuXXX:
>I recommend re-arranging the commits so the first commit shows the current behavior via a test 
Okay, revised the commit order

LuuuXXX:
> Its not just about the commit order but that every commit should pass cargo test.
No problem, I've resubmitted the commits.

epage:
The test is still the last commit.

The test should be the first commit **and** it should pass as-is.  The follow up changes should change the tests as the behavior changes which will show that the behavior did change in the intended way and that the test tests the right thing.  Last I looked, it appeared that the current tests do not test the right thing.

LuuuXXX:
>The test should be the first commit and it should pass as-is. 

To facilitate review, I reordered commit.

>Last I looked, it appeared that the current tests do not test the right thing.

I've added two new tests,
One is to ensure that no package can be obtained from the vendor directory when a non-vendor package is added. (Throwing error in my opinion is the right thing to do.)
The other is to ensure that non-vendor packages can be obtained through alter registry.

tests/testsuite/cargo_add/add_no_vendored_package_with_alter_registry/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+            r#"[workspace]
+
+[package]
+name = "cargo-list-test-fixture"
+version = "0.0.0"
+"#,
+        )
+        .build();
+    let project_root = project.root();
+    let cwd = &project_root;
```
epage, line 44: Why does this use both the `in/` and a `ProjectBuilder`?

epage: (same for other tests)

tests/testsuite/cargo_add/add_no_vendored_package_with_vendor/mod.rs
```diff
@@ -1,44 +1,12 @@
...
-        )
-        .file("src/lib.rs", "")
-        .file(
-            "Cargo.toml",
-            r#"[workspace]
-
-[package]
-name = "cargo-list-test-fixture"
-version = "0.0.0"
-"#,
```
epage: Was it intentional that `aa` wasn't carried over?  Seems like its needed to get the original error message.

LuuuXXX: This imitated the `in/` format to create `aa` (in the vendor directory) instead of ProjectBuilder.

epage:
@LuuuXXX it looks like you need to rebase against `master` and update your tests for the `snapbox` upgrade from #13441

epage:
Can you squash that in so all of the commits pass?

LuuuXXX:
Thanks, it's been squashed

epage:
> Thanks, it's been squashed

I was asking for the last commit to be squashed, not all of them.

LuuuXXX:
my bad, I rebuilt the commit history with latest modifies

LuuuXXX:
yes yes, recreated again

epage:
Thanks for all of the work you put into this and patience through the back and forth!

@bors r+

PR #13287, opened 2024-01-12
fix(resolver): do not panic when sorting empty summaries

weihanglo:

This is very unlikely to misuse… but I encountered that today.
At our best we should prevent this from happening.

(I didn't hit this when we had `first_version` and `version_ordering` separately)

Commit by commit.


Eh2406:
This seemed reasonable. But for some context how did this come up?

weihanglo:
> This seemed reasonable. But for some context how did this come up?

I use `cargo` as a lib dependency.

* Create a source manually.
* Query a non-existing package
* Sorting with `Some(VersionOrdering::MaximumVersionsFirst)` <-- It's my fault that didn't check API change carefully.

PR #13292, opened 2024-01-14
Delete sentence about parentheses being unsupported in license

dtolnay:
Parentheses have been supported by crates.io since 2 years ago.

- https://github.com/rust-lang/crates.io/pull/4257

Their functionality is tested by this test: https://github.com/rust-lang/crates.io/blob/3acd63c1f3625d83322e4819d76296ae5ef291c5/tests/utils/license-test.js#L68-L79.

I think a separate test in Cargo is most likely not valuable because Cargo does not parse these license strings, they are just treated as `Option<String>`.

Here is an example of an extremely widely used package (147 million downloads) with parentheses in its license: https://crates.io/crates/unicode-ident.


weihanglo (approved):

weihanglo:
Thanks!

(wonder how did you find this 😆)

@bors r+

PR #13301, opened 2024-01-15
Update ahash dependency to 0.8.7

Amanieu:
Previous versions use the `stdsimd` nightly feature if it detects that it is available (on a nightly toolchain). This will stop working once rust-lang/rust#117372 is merged.


epage:
@bors r+

> Previous versions use the stdsimd nightly feature if it detects that it is available (on a nightly toolchain). 

*grumble* imo dependents should be given the choice of nightly features being enabled, rather than doing it for them automatically.  This burned me a lot with dtolnay's packages already, making me extra grumbly.

PR #13317, opened 2024-01-18
Improve GitHub Actions CI config

Urgau:
This pull request improves Cargo CI by:
 1. Explicitly listing the branches the main CI should be run on (to avoid running on unwanted branches, particularly in forks)
 2. By adding a [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) config so that only one "CI" runs for each "group" (one by PR), very inspired by rust-lang/rust [one](https://github.com/rust-lang/rust/blob/c485ee71477a29041895c47cc441b364670f3772/.github/workflows/ci.yml#L35)
 3. ~~By making all the jobs run only if triggered from this repository (to avoid having forks unnecessarily run CI jobs)~~
 4. ~~By making all the checkouts take `fetch-depth: 2` like done in [rust-lang/rust](https://github.com/rust-lang/rust/blob/c485ee71477a29041895c47cc441b364670f3772/.github/workflows/ci.yml#L452-L455) (to avoid unnecessary work, saves bandwidth and maybe a bit of time)~~
 5. By including `refs/heads/try` in the `success` and `failure` bors job gates (to fix `bors try` from timing out)

The main motivation for this PR was (at first) to make the CI jobs not run on forks since it's a waste of CI resources as well as always failing (like [this](https://github.com/Urgau/cargo/actions/runs/7567435677/job/20606586124)); but after looking at the `main.yml` config I took the opportunity to improve the situation by using rust-lang/rust [ci.yml](https://github.com/rust-lang/rust/blob/master/.github/workflows/ci.yml) as reference.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,7 +36,7 @@ jobs:
     - test
     - test_gitoxide
     runs-on: ubuntu-latest
-    if: "success() && github.event_name == 'push' && github.ref == 'refs/heads/auto-cargo'"
+    if: "success() && github.event_name == 'push' && (github.ref == 'refs/heads/auto-cargo' || github.ref == 'refs/heads/try') && github.repository == 'rust-lang/cargo'"
     steps:
     - run: echo ok
   failure:
@@ -54,7 +54,7 @@ jobs:
     - test
     - test_gitoxide
     runs-on: ubuntu-latest
-    if: "!success() && github.event_name == 'push' && github.ref == 'refs/heads/auto-cargo'"
+    if: "!success() && github.event_name == 'push' && (github.ref == 'refs/heads/auto-cargo' || github.ref == 'refs/heads/try') && github.repository == 'rust-lang/cargo'"
     steps:
     - run: exit 1
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,9 +1,13 @@
 name: CI
 on:
   push:
-    branches-ignore: [master]
+    branches:
+      - auto-cargo
+      - try
+      - automation/bors/try
   pull_request:
-    branches: ['*']
+    branches:
+      - "**"
 
 defaults:
   run:
@@ -11,6 +15,10 @@ defaults:
 
 permissions:
   contents: read
+
+concurrency:
+  group: "${{ github.workflow }}-${{ (github.ref == 'refs/heads/try' && github.sha) || github.ref }}"
+  cancel-in-progress: true
 
 jobs:
   success:


.github/workflows/main.yml
```diff
@@ -53,17 +62,23 @@ jobs:
   # Check Code style quickly by running `rustfmt` over all code
   rustfmt:
     runs-on: ubuntu-latest
+    if: "github.repository == 'rust-lang/cargo'"
     steps:
     - uses: actions/checkout@v4
+      with:
+        fetch-depth: 2
```
epage, line 69: I'm assuming depth of 2 is merge commit plus top commit from users branch? If we only need to merge commit, we should probably just use that.

.github/workflows/main.yml
```diff
@@ -1,9 +1,14 @@
...
-    branches-ignore: [master]
+    branches:
+      - auto-cargo
+      - try
+      - automation/bors/try
+      - "rust-**"
   pull_request:
-    branches: ['*']
+    branches:
+      - "**"
```
epage, line 11: Whats the double glob do differently?

I think you can also not have the arry and it works? I think thats what my repos do

.github/workflows/main.yml
```diff
@@ -53,17 +62,23 @@ jobs:
...
+      with:
+        fetch-depth: 2
     - run: rustup update stable && rustup default stable
     - run: rustup component add rustfmt
     - run: cargo fmt --all --check
 
   # Ensure there are no clippy warnings
   clippy:
     runs-on: ubuntu-latest
+    if: "github.repository == 'rust-lang/cargo'"
```
epage, line 77: How much do we need this if we block push branches?

This can be annoying for people intentionally forking, like a company recently did for hack week.

epage (commented):
Thanks!

.github/workflows/main.yml
Urgau: According to [GitHub Workflow syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet):

- `*`: Matches zero or more characters, but does not match the `/` character. For example, `Octo*` matches `Octocat`.
- `**`: Matches zero or more of any character.

So I think we should use the double glob, unless we have a good reason to exclude branch name with `/` included in them.

.github/workflows/main.yml
Urgau: Strictly speaking, it's not required. I mainly followed here what rust-lang/rust does.

But if you think this will actively hurt people, I'm open to removing it.

.github/workflows/main.yml
```diff
@@ -1,9 +1,14 @@
 name: CI
 on:
   push:
-    branches-ignore: [master]
+    branches:
+      - auto-cargo
+      - try
+      - automation/bors/try
+      - "rust-**"
```
ehuss, line 8: Why add rust-** here?

.github/workflows/main.yml
ehuss: I frequently run CI in my fork to test things. This would make it more difficult since I would need to comment this out.

.github/workflows/main.yml
Urgau: This is apparently useful for [checking `HEAD^`](https://github.com/actions/checkout#checkout-head). But looking more closely at the `checkout` action the default is already at `1`, so I'm gonna revert those.

.github/workflows/main.yml
Urgau: CI is currently run on those branches and I didn't wanted to change that.

weihanglo:
Regardless, let's see if it works with bors

@bors try

.github/workflows/main.yml
Urgau: Done reverted.

.github/workflows/main.yml
Urgau: *I looked at your fork and didn't see any relevant PRs, so I'm assuming you don't use PRs to your own fork.*

Wouldn't you be also affected by the allowlist of branches? Or are you planning on making PRs to your own fork?

.github/workflows/main.yml
ehuss: I don't think it is needed. Those are beta branches, and those go through bors like a normal `master` merge (and thus are tested on the `auto-cargo` branch).

Urgau: Sure, but don't you want the CI checks to be run before doing `@bors r+` ?

Otherwise without this rule I don't think there will be any CI checks when targeting the beta branches.

.github/workflows/main.yml
epage: The company never opened a PR against our repo. They all worked against a specific fork.

Urgau: > The company never opened a PR against our repo. They all worked against a specific fork.

Okay, then removing those conditions seems better. Removed.

Urgau:
Hum, I have been informed @weihanglo that apparently the `bors try` kinda works but always timeout.

I believe this is due to an overly constrained condition in the `success` and `failure` bors gates. I've made some changes that I think will solve this. Let's try?

weihanglo:
Sure. Do it again.

@bors try

.github/workflows/main.yml
ehuss: Shouldn't that be handled by the `pull_request` trigger?

Urgau: Oh! You're probably right, I thought we were in the `pull_request` section, not the `push` one.

Urgau: Removed.

epage:
Looks like everything is good?

@bors r+

PR #13324, opened 2024-01-18
chore: Make MSRV=N-2 the workspace default

epage:

Since we update the MSRV every release, `cargo bump-check` forces the changed packages to be released (#13266).
A caveat is that `cargo bump-check` only checks the directory and doesn't do a `.crate` diff, so it misses out of changes to `workspace.package`.
Because of our current workspace default, this means every 6 weeks we ship our user-facing packages but not the internal ones.

This shifts that so we ship our internal packages every 6 weeks and our internal only when changed. 

### Additional information

 If `rust-version = "auto"` (rust-lang/rfcs#3537) gets approved, then we will only ship packages that actually change


weihanglo:
Thanks!

@bors r+

PR #13329, opened 2024-01-19
Fix static_mut_ref warning.

ehuss:
The nightly compiler has recently added the [`static_mut_ref`](https://doc.rust-lang.org/nightly/rustc/lints/listing/warn-by-default.html#static-mut-ref) lint, which warns about taking a reference to a `mut static`. This fixes the issue by copying the value instead of taking a reference. I don't recall what I was thinking when I wrote this, but it was probably a bad reflex against making copies.


weihanglo (approved):

weihanglo:
@bors r+

Thanks!

Issue #11666, opened 2023-02-01
Provide a json output format of `cargo package --list` 


messense:
### Problem

Currently `cargo package --list` outputs filenames line by line, it works fine normally. But when `Cargo.toml` contains metadata that uses relative path to point to parent directory, there is no way to get that information. For example `readme = "../README.md`:

```toml
[package]
readme = "../README.md"
```

`cargo package --list` outputs

```
README.md
```

instead of

```
../README.md
```

It makes sense because it's the filesystem layout in the `.crate` file, but for downstream tooling like [maturin](https://github.com/PyO3/maturin) it doesn't know that the `READMD.md` is actually coming from `../README.md`, which causes https://github.com/PyO3/maturin/issues/1431

### Proposed Solution

Add a json output format, for example `cargo package --list=json` outputs something like

```json
{
  "README.md": "../README.md",
  "src/lib.rs": "src/lib.rs"
}
```

as proposed by @konstin in https://github.com/PyO3/maturin/issues/1442#issuecomment-1410018083

messense:
Related to https://github.com/rust-lang/cargo/issues/11597, but `CARGO_PKG_README` isn't usable outside of Cargo build process.

cc @weihanglo

weihanglo:
Thanks for the proposal! 

Although this doesn't sound too hard to implement, we should still make sure that we have thought of different workarounds and designs  before moving on.

For workarounds, off the top of my head there are at least two:

* `cargo metadata | jq '.packages[] | select(.name == "<package-name>") | .readme'` to get the path.
* Make the `README.md` file a symlink. Cargo takes care of symlinks and copies over.

The second one is definitely a feasible option I feel like.

For the design of this proposal, I'd like to know apart from the issue from `maturin`, is there any other issue that this proposal can resolve? One bug could probably be fixed with this proposal: `cargo package --list` doesn't escape or quote filenames. If a path contains newlines it becomes hard to parse by machines. The other question is that, do people want to know more on filesize, file permission, and many more? What information should be included in that JSON output? We should also make sure that we don't push the responsibility of `cargo package` too far, as it is first served as “listing files being packaged”.

konstin:
[...]

Issue #11405, opened 2022-11-22
Make workspace member inclusion/exclusion more useful, consistent, and intuitive


weihanglo:
### Problem

As of now, Cargo supports glob syntax for [`workspace.member`] to include crates as workspace members. However, for `workspace.exclude` it is implemented by [a plain `starts_with`] comparing with manifest path of certain crates. This confuses people as they need to figure out which rules overrides the other. It is also inconsistent with [`package.include`] and [`package.exclude`], as they choose gitignore syntax.  

Similar issues have been raised from time to time:

- #4593
- #6745
- #6009
- #11362

[`workspace.member`]: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields
[a plain `starts_with`]: https://github.com/rust-lang/cargo/blob/65535890dbaa2b186633aedc11e550c3990fc5e9/src/cargo/core/workspace.rs#L1589-L1593
[`package.include`]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields
[`package.exclude`]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields

### Proposed Solution

There are several possible ways to "fix" the situation (to some extent).

* The first is implementing glob support for `workspace.exclude`. #11374 tries to do that. This is a relatively low effort approach but might confuse people still, and it won't fix #6745.
* The other way is using gitignore syntax, as ehuss mentioned earlier <https://github.com/rust-lang/cargo/issues/6009#issuecomment-519576849>. This would be a breaking change and might need a transition period and provide migration steps.
* In <https://github.com/rust-lang/cargo/issues/11362#issuecomment-1311811774>, we talked about using a single `workspace.member` and drop `workspace.exclude`. However, the re-include is not possible due to the limit of gitignore itself[^1].
* We might want to avoid `starts_with` and instead check the existence of `Cargo.toml` to include/exclude a package.

There might be other possible solution I am not aware of here.

[^1]: An optional prefix "!" which negates the pattern; any matching file excluded by a previous pattern will become included again. It is not possible to re-include a file if a parent directory of that file is excluded.
See https://git-scm.com/docs/gitignore#_pattern_format

### Notes

I would keep this as the main issue and close some. If this is inappropriate, please say it out load, and we can reopen the original issues.

hi-rustin:
I would like to find a solution to this issue. 
@rustbot claim

Issue #8407, opened 2020-06-25
cargo-package: list of dirty files contains ignored files


dotdash:
**Problem**
When running "cargo package" it complains about some files containing uncommited change, even though they're ignored by my .gitignore rules.

This seems to only happen when the file is in a directory that contains a mix of ignored and non-ignored files, but no tracked files at all.

```bash
$ git status
On branch master
nothing to commit, working tree clean

$ mkdir new_dir
$ touch new_dir/file
$ touch new_dir/.file.swp

$ git status
On branch master
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        new_dir/

nothing added to commit but untracked files present (use "git add" to track)

$ git status new_dir/
On branch master
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        new_dir/file

nothing added to commit but untracked files present (use "git add" to track)

$ cargo package
error: 2 files in the working directory contain changes that were not yet committed into git:

new_dir/.file.swp
new_dir/file
```

I expect to only see the entry for `new_dir/file` there, not `new_dir/.file.swp`.

**Notes**

Output of `cargo version`:

```bash
$ cargo --version
cargo 1.44.1 (88ba85757 2020-06-11)
```

ehuss:
What are your gitignore rules?

dotdash:
Nothing special, just:

```
/target
*.rs.bk
Cargo.lock
.*.swp
```

I originally had .*.swp in my `.config/git/ignore` but added it to the local .gitignore as well, just to make sure that that is not causing it. Other files like foo.rs.bk also show up, so it's not the leading dot either.

ehuss:
Ah, I see what is happening.  Git reports the entire directory as untracked if there is at least one non-ignored file.  Cargo [walks the entire directory](https://github.com/rust-lang/cargo/blob/3b142db791ac023a8e8e856c1b18c1ea9367a81a/src/cargo/sources/path.rs#L326) without further consulting git.  Git has an option to [recurse into untracked dirs](https://github.com/rust-lang/git2-rs/blob/31c151556d50783c9af48b2cff4e3289cd880b56/src/status.rs#L145-L151), but then Cargo would need to apply additional filtering, like skipping the `target` directory.  That isn't particularly easy since git is file-based, so it is tricky to detect if it has recursed into the target directory.

dotdash:
[...]

PR #13330, opened 2024-01-20
Add `include` and `exclude` information to metadata

GuillaumeGomez:
I'll need this information to fix https://github.com/rust-lang/rust-clippy/issues/11677. I'll also send a PR to `cargo_metadata` once merged.

Of course, if there is another way to retrieve this information without needing this PR, I'd love to hear it. :)

### DIFF
--- a/src/cargo/core/package.rs
+++ b/src/cargo/core/package.rs
@@ -106,6 +106,8 @@ pub struct SerializedPackage {
     metabuild: Option<Vec<String>>,
     default_run: Option<String>,
     rust_version: Option<RustVersion>,
+    include: Vec<String>,
+    exclude: Vec<String>,
 }
 
 impl Package {
@@ -262,6 +264,8 @@ impl Package {
             publish: self.publish().as_ref().cloned(),
             default_run: self.manifest().default_run().map(|s| s.to_owned()),
             rust_version: self.rust_version().cloned(),
+            include: self.manifest().include().to_vec(),
+            exclude: self.manifest().exclude().to_vec(),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/cargo/core/package.rs
```diff
@@ -106,6 +106,8 @@ pub struct SerializedPackage {
     metabuild: Option<Vec<String>>,
     default_run: Option<String>,
     rust_version: Option<RustVersion>,
+    include: Vec<String>,
```
weihanglo, line 109: While it might be useful to have `include` and `exclude` in `cargo metadata` output, it doesn't seem to be a proper way to fix rust-lang/rust-clippy#11677. The solution might end up reimplement the logic [Cargo find and package files](https://github.com/rust-lang/cargo/blob/7bb7b539558dc88bea44cee4168b6269bf8177b0/src/cargo/sources/path.rs#L133-L200), which includes `.gitignore`, shell glob syntax, symlinks handling, git dirty status, and other details (see #11405). `cargo package --list` might be more reliable, but not 100% accurate (see #8407 and #11666).

I would suggest starting the discussion from an issue, and jump back to implementation when the design is settled.

GuillaumeGomez: Yes that's a good point. I'll open an issue so it can be discussed.

GuillaumeGomez: I opened https://github.com/rust-lang/cargo/issues/13331.

weihanglo:
I'll close this in favor of high level discussions in #13331. When we have a settled design we can always come back and revisit this implementation. Thank you :)

Issue #13310, opened 2024-01-17
`cargo run` error: package(s) `...` not found in workspace


kzhui125:
### Problem

With `cargo run --package time-core:10.0.0 --bin time-core ` in workspace folder, I get error `error: package(s) ... not found in workspace`.

<img width="633" alt="image" src="https://github.com/rust-lang/cargo/assets/11925432/c5beb398-301b-421c-b998-0e55b255e2c2">

[workspace_0117.zip](https://github.com/rust-lang/cargo/files/13960582/workspace_0117.zip)

```
cargo run --package time-core:10.0.0 --bin time-core 
```

expected: no error.

### Version

```text
cargo 1.75.0 (1d8b05cdd 2023-11-20)
release: 1.75.0
commit-hash: 1d8b05cdd1287c64467306cf3ca2c8ac60c11eb0
commit-date: 2023-11-20
host: aarch64-apple-darwin
libgit2: 1.7.1 (sys:0.18.1 vendored)
libcurl: 8.4.0 (sys:0.4.68+curl-8.4.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1u  30 May 2023
os: Mac OS 14.2.1 [64-bit]
```

kzhui125:
If cargo don't support the same name for `my crate` and `dependency crate`. The error message should change and vscode rust extension should give error.

If cargo support the same name for  `my crate` and `dependency crate`.  cargo run should work.

What if `my crate` and `dependency crate` have the same name and same version?

hi-rustin:
It seems to work if we don't use a package name. For example:
```console
❯ cargo run                           
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/time-core`
Hello world
```

kzhui125:
VSCode extension will use `cargo run --package time-core:10.0.0 --bin time-core`, and it will fail.

epage:
It looks like we are passing the `spec` unchanged as a package name.  In my quick glance, not seeing why `cargo check` translates the spec into a package id but `cargo run` doesn't.  `cargo run` does do somethings differently.

hi-rustin:
@rustbot claim

Let me try to fix it.

hi-rustin:
After some debugging:

I think the problem is here:
[...]

PR #13335, opened 2024-01-22
fix: use spec id instead of name to match package

hi-rustin:
close https://github.com/rust-lang/cargo/issues/13310

Users tried `cargo run -p foo@0.5.0`.

The bug came from here:

1. We store the spec as the name in this function: [hi-rustin/cargo@aa64447/src/cargo/ops/cargo_compile/packages.rs#L201](https://github.com/hi-rustin/cargo/blob/aa6444734d0049b3534fa1ae623c55dfbef63412/src/cargo/ops/cargo_compile/packages.rs#L201)
2. But we use it here to match only the name of the package: [hi-rustin/cargo@aa64447/src/cargo/ops/cargo_compile/packages.rs#L128](https://github.com/hi-rustin/cargo/blob/aa6444734d0049b3534fa1ae623c55dfbef63412/src/cargo/ops/cargo_compile/packages.rs#L128)

So `foo@0.5.0` didn't match with `foo`. It said there is no `foo@0.5.0` in the workspace.

In order to resolve this issue, we need to utilize the package specification ID directly. If we only extract the name and drop the version specification, it could potentially lead to another bug. For instance, if you have a package `foo@0.1.1 in your workspace and you intend to use `foo@0.2.2`, extracting only the name would result in the selection of `foo@0.1.1` instead.

So in this PR I parsed the package spec ID and used it to match the package's spec ID.

Check out the unit test.

### Additional information

None

### DIFF
--- a/src/cargo/ops/cargo_compile/packages.rs
+++ b/src/cargo/ops/cargo_compile/packages.rs
@@ -2,7 +2,7 @@
 
 use std::collections::BTreeSet;
 
-use crate::core::Package;
+use crate::core::{Package, PackageIdSpecQuery};
 use crate::core::{PackageIdSpec, Workspace};
 use crate::util::restricted_names::is_glob_pattern;
 use crate::util::CargoResult;
@@ -50,16 +50,24 @@ impl Packages {
                 .map(|id| id.to_spec())
                 .collect(),
             Packages::OptOut(opt_out) => {
-                let (mut patterns, mut names) = opt_patterns_and_names(opt_out)?;
+                let (mut patterns, mut ids) = opt_patterns_and_ids(opt_out)?;
                 let specs = ws
                     .members()
                     .filter(|pkg| {
-                        !names.remove(pkg.name().as_str()) && !match_patterns(pkg, &mut patterns)
+                        let id = ids.iter().find(|id| id.matches(pkg.package_id())).cloned();
+                        if let Some(id) = &id {
+                            ids.remove(id);
+                        }
+                        !id.is_some() && !match_patterns(pkg, &mut patterns)
                     })
                     .map(Package::package_id)
                     .map(|id| id.to_spec())
                     .collect();
                 let warn = |e| ws.config().shell().warn(e);
+                let names = ids
+                    .into_iter()
+                    .map(|id| id.to_string())
+                    .collect::<BTreeSet<_>>();
                 emit_package_not_found(ws, names, true).or_else(warn)?;
                 emit_pattern_not_found(ws, patterns, true).or_else(warn)?;
                 specs
@@ -68,11 +76,7 @@ impl Packages {
                 vec![ws.current()?.package_id().to_spec()]
             }
             Packages::Packages(opt_in) => {
-                let (mut patterns, packages) = opt_patterns_and_names(opt_in)?;
-                let mut specs = packages
-                    .iter()
-                    .map(|p| PackageIdSpec::parse(p))
-                    .collect::<Result<Vec<_>, _>>()?;
+                let (mut patterns, mut specs) = opt_patterns_and_ids(opt_in)?;
                 if !patterns.is_empty() {
                     let matched_pkgs = ws
                         .members()
@@ -82,7 +86,7 @@ impl Packages {
                     specs.extend(matched_pkgs);
                 }
                 emit_pattern_not_found(ws, patterns, false)?;
-                specs
+                specs.into_iter().collect()
             }
             Packages::Default => ws
                 .default_members()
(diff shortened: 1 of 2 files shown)


tests/testsuite/run.rs
```diff
@@ -1566,3 +1566,47 @@ fn run_link_system_path_macos() {
...
+                [lib]
+                name = "foo"
+                path = "foo.rs"
+            "#,
+        )
+        .file("foo/foo.rs", "")
+        .build();
+
+    p.cargo("run").run();
+    p.cargo("run -p foo@0.5.0")
```
epage, line 1608: FYI I think the problem has less to do with sharing a name with a dependency and more to do with `cargo run` not accepting a package spec but a workspace member name.

src/cargo/ops/cargo_compile/packages.rs
```diff
@@ -50,16 +50,24 @@ impl Packages {
                 .map(|id| id.to_spec())
                 .collect(),
             Packages::OptOut(opt_out) => {
-                let (mut patterns, mut names) = opt_patterns_and_names(opt_out)?;
+                let (mut patterns, mut ids) = opt_patterns_and_ids(opt_out)?;
                 let specs = ws
                     .members()
                     .filter(|pkg| {
-                        !names.remove(pkg.name().as_str()) && !match_patterns(pkg, &mut patterns)
+                        let id = ids.iter().find(|id| id.matches(pkg.package_id())).cloned();
```
weihanglo, line 57: The interesting part is that none of these needs a full pkgid sped match. Comparing member name with spec.name is quite sufficient I guess?

weihanglo: That is not allowed. You cannot run a dependency

hi-rustin: I mean the member is `foo@0.1.0`, and users want to run `foo@0.2.0`. If we only check the name, we will pick `foo@0.1.1` and run it. But I think it is an invalid package spec, we should say `foo@0.2.0` does not exist in your workspace.

weihanglo: Yeah that is a valid pkgid spec not existing in the workspace. We can go further by suggesting closet member to them, maybe.

So far as I can tell, only in [this branch](https://github.com/rust-lang/cargo/blob/b2e1d3b6235c07221dd0fcac54a7b0c754ef8b11/src/cargo/ops/cargo_compile/packages.rs#L70) the given specs fall through, which makes `cargo build -p <dependency>` works. Under other branches one can never get specs beyond workspace members.

hi-rustin: It seems the bug from [this branch](https://github.com/rust-lang/cargo/blob/b2e1d3b6235c07221dd0fcac54a7b0c754ef8b11/src/cargo/ops/cargo_compile/packages.rs#L128). When running `cargo run -p mypackage@0.1.0`, if we compare `mypackage@0.1.0` with `mypackage`, the error message "no `mypckage@0.1.0 in this workspace" is shown, even though the package is present.
So the fix is we use spec directly instead of only comparing the name.

weihanglo: Oh I see. We still need a full pkgid spec matches, as we need to point out that pkgid spec, say `foo@0.2.0` doesn't exist.

hi-rustin: We call [this branch](https://github.com/rust-lang/cargo/blob/b2e1d3b6235c07221dd0fcac54a7b0c754ef8b11/src/cargo/ops/cargo_compile/packages.rs#L128) from [here](https://github.com/hi-rustin/cargo/blob/3c3fe21fd26aded05fb69124fba16ec0d47d5133/src/bin/cargo/commands/run.rs#L71).

weihanglo: I understand the patch, and didn't mean it's incorrect. I was trying to find a way to avoid full spec match, but maybe not worthy as it may get things complicated.

src/cargo/ops/cargo_compile/packages.rs
```diff
@@ -109,25 +113,41 @@ impl Packages {
...
+                let (mut patterns, mut ids) = opt_patterns_and_ids(opt_out)?;
                 let packages = ws
                     .members()
                     .filter(|pkg| {
-                        !names.remove(pkg.name().as_str()) && !match_patterns(pkg, &mut patterns)
+                        let id = ids.iter().find(|id| id.matches(pkg.package_id())).cloned();
+                        if let Some(id) = &id {
+                            ids.remove(id);
+                        }
+                        !id.is_some() && !match_patterns(pkg, &mut patterns)
```
hi-rustin, line 124: I am not sure if there is any better way to do it.

hi-rustin (commented):
🔢 Self-check

src/cargo/ops/cargo_compile/packages.rs
```diff
@@ -189,19 +209,20 @@ fn emit_pattern_not_found(
 
 /// Given a list opt-in or opt-out package selection strings, generates two
 /// collections that represent glob patterns and package names respectively.
```
weihanglo, line 211:
```suggestion
/// collections that represent glob patterns and package id specs respectively.
```

src/cargo/ops/cargo_compile/packages.rs
```diff
@@ -189,19 +209,20 @@ fn emit_pattern_not_found(
...
-    let mut opt_names = BTreeSet::new();
+    let mut opt_ids = BTreeSet::new();
     for x in opt.iter() {
         if PackageIdSpec::parse(x).is_err() && is_glob_pattern(x) {
             opt_patterns.push((build_glob(x)?, false));
         } else {
-            opt_names.insert(String::as_str(x));
+            let id = PackageIdSpec::parse(x)?;
+            opt_ids.insert(id);
         }
```
weihanglo, line 223: Something like this to avoid parsing twice?

```suggestion
        match PackageIdSpec::parse(x) {
            Ok(spec) => {
                opt_ids.insert(spec);
            }
            Err(_) if is_glob_pattern(x) => opt_patterns.push((build_glob(x)?, false)),
            Err(e) => return Err(e.into()),
        }
```

weihanglo (commented):
Other parts look great. Thanks for the fix!

tests/testsuite/run.rs
```diff
@@ -1606,4 +1606,9 @@ fn run_binary_with_same_name_as_dependency() {
     p.cargo("run").run();
     p.cargo("check -p foo@0.5.0").run();
     p.cargo("run -p foo@0.5.0").run();
+    p.cargo("run -p foo@0.5").run();
+    p.cargo("run -p foo@0.4")
+        .with_status(101)
+        .with_stderr("[ERROR] package(s) `foo@0.4` not found in workspace `[..]`")
```
epage, line 1612: Thanks for adding this test!
I was just looking to see if we had a test to cover this case

src/cargo/ops/cargo_compile/packages.rs
hi-rustin: Good suggestion!

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #13337, opened 2024-01-22
feat(tree): Control `--charset` via auto-detecting config value

epage:

This tries to generalize `cargo tree --charset` so any part of cargo's output can use it.  For example,
- `cargo search` could use this for fancier tables
- `cargo add` could use this for fancier feature lists
- #12235 could use this for fancy rendering like miette does (CC @Muscraft )
- Progress bars could use fancier characters <-- this is what I'm personally working towards

This builds on the idea from #12889 that we can use fancier terminal features so long as we have good auto-detection and provide users an escape hatch until the auto-detection is improved or in case they disagree.

As a side benefit
- `cargo tree` will auto-detect when the prior default of `--charset utf8` is a good choice
- Users can control `cargo tree --charset` via `.cargo/config.toml`

This is gradually introduced through the individual commits.

### DIFF
--- a/src/cargo/core/shell.rs
+++ b/src/cargo/core/shell.rs
@@ -251,6 +251,19 @@ impl Shell {
         Ok(())
     }
 
+    pub fn set_unicode(&mut self, yes: bool) -> CargoResult<()> {
+        if let ShellOut::Stream {
+            ref mut stdout_unicode,
+            ref mut stderr_unicode,
+            ..
+        } = self.output
+        {
+            *stdout_unicode = yes;
+            *stderr_unicode = yes;
+        }
+        Ok(())
+    }
+
     pub fn set_hyperlinks(&mut self, yes: bool) -> CargoResult<()> {
         if let ShellOut::Stream {
             ref mut hyperlinks, ..
--- a/src/cargo/core/shell.rs
+++ b/src/cargo/core/shell.rs
@@ -53,6 +53,8 @@ impl Shell {
                 color_choice: auto_clr,
                 hyperlinks: supports_hyperlinks(),
                 stderr_tty: std::io::stderr().is_terminal(),
+                stdout_unicode: supports_unicode(&std::io::stdout()),
+                stderr_unicode: supports_unicode(&std::io::stderr()),
             },
             verbosity: Verbosity::Verbose,
             needs_clear: false,
@@ -259,6 +261,20 @@ impl Shell {
         Ok(())
     }
 
+    pub fn out_unicode(&self) -> bool {
+        match &self.output {
+            ShellOut::Write(_) => true,
+            ShellOut::Stream { stdout_unicode, .. } => *stdout_unicode,
+        }
+    }
+
+    pub fn err_unicode(&self) -> bool {
+        match &self.output {
+            ShellOut::Write(_) => true,
+            ShellOut::Stream { stderr_unicode, .. } => *stderr_unicode,
+        }
+    }
+
     /// Gets the current color choice.
     ///
     /// If we are not using a color stream, this will always return `Never`, even if the color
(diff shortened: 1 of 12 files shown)


src/doc/src/reference/config.md
```diff
@@ -1279,6 +1280,16 @@ Can be overridden with the `--color` command-line option.
...
 
+#### `term.charset`
+* Type: string
+* Default: auto-detect
+* Environment: `CARGO_TERM_CHARSET`
+
+Sets what characters can be used when rendering cargo output.  Possible values:
+
+* `utf8`: Use the full set of UTF-8 characters for drawing
+* `ascii`: Only use 7-bit ASCII characters for drawing
```
epage, line 1291: My main concern is the future possibilities.  If we later add [emoji](https://crates.io/crates/terminal-supports-emoji) or [devicon fonts](https://devicon.dev/), how would we represent this?  A separate field, hierarchical values, or a list?

weihanglo: One possible way: default to "auto" (I sympathize this word after cargo meeting today), and remove the `utf8` option. This could maximize our future compatibility.

weihanglo (commented):
Implementation-wise looks good to me 👍🏾.

src/doc/src/reference/config.md
epage: The downside is that doesn't allow forcing of `utf8` if  our `auto` is being overly selective.  Granted, maybe thats not too bad for now?  It would mean we'd need separate enums for config from internal / CLI to support `cargo tree --charset utf8`.

Thoughts?

weihanglo: Do we have any incoming feature going to use Unicode tables? If not, I feel like stabilizing fewer things buys us more time to deal with future changes.

BTW, I was about to convince myself an “auto” option is more aligned to other `term` configs, until I saw `term.hyperlinks`.

>  It would mean we'd need separate enums for config from internal / CLI to support cargo tree --charset utf8.

Yeah. Sounds like it 😞.

epage: Yes, there isn't something using this yet.  Overall, this is part of a slow roll out of features to unblock user control over more complex progress bars so we have cleaner command output.

I also don't see us evaluating devicons any time soon.  Emoji is hit or miss. I don't see an immediate use of it but personally not a big emoji person.


Maybe we can at least hold off on this until I get a little further into the progress bar implementation research?  That way if emoji somehow looks useful in that time we re-evaluate?

epage: I just re-worked this PR to instead have a `term.unicode` field.  This would allow us to add more fields in the future like `term.emoji` without worrying about the specifics of what we might need in the future.

Thoughts?

src/cargo/core/shell.rs
```diff
@@ -519,6 +550,14 @@ fn supports_color(choice: anstream::ColorChoice) -> bool {
     }
 }
 
+fn supports_unicode(stream: &dyn IsTerminal) -> bool {
+    if !stream.is_terminal() || supports_unicode::supports_unicode() {
+        true
+    } else {
+        false
+    }
```
weihanglo, line 558: nit: While I know if-else is more readable, should we collapse it?

```suggestion
    !stream.is_terminal() || supports_unicode::supports_unicode()
```

weihanglo (approved):
Looks good to me. `r=weihanglo` whenever you feel ready to merge.

epage:
@rfcbot fcp merge

This adds a new config field, `term.unicode = <bool>`.  When unset, it is auto-detected via [supports-unicode](https://crates.io/crates/supports-unicode).  As of this PR, this just controls the default for  `cargo tree --charset` but this allows us to leverage this anywhere.  Potential ideas include
- `cargo info` feature trees
- Diagnostics like [miette](https://crates.io/crates/miette)
- Fancier looking progress updates

weihanglo:
I don't think this needs a full 10-day FCP. Merge this now.

@bors r+

Issue #12897, opened 2023-10-30
Cargo always uses LF line endings for Cargo.lock even if CRLF were present


jder:
### Problem

This [PR](https://github.com/rust-lang/cargo/pull/2092/files) fixed #2076 but it appears that the functionality has regressed. When cargo writes a lock file, it always uses LF line endings rather than matching the existing line endings as I would expect. 

The test for the above PR writes out a file with CRLFs and then runs generate-lockfile and expects the content on disk to be unchanged. Unfortunately, generate-lockfile tests to see if the [lines are equal](https://github.com/rust-lang/cargo/blob/b988daaf7fa63f10edf419c56340b0d7a4772e53/src/cargo/ops/lockfile.rs#L209) (which is line-ending agnostic) before writing anything out and so doesn't actually re-write the file. If you [edit the test to defeat this optimization](https://github.com/rust-lang/cargo/compare/master...jder:cargo:revive-issue-2076?expand=1#diff-2d3e72798f9e683ac610e8d9b1e316e4e4ca0ae58e5256e95e53660eb9fdb03a) then it fails because the lock file is written out with LFs only.

### Steps

1. Check out a rust project on Windows with the (default) git setting of autocrlf = true. 
2. Note that line endings of your cargo.lock file are CRLF. (`git ls-files --eol` is a convenient way to do this)
3. Run `cargo add $SOMETHING`
4. Note that line endings have changed to LF. 
5. Try committing the lock file; this will fail if you have the (default) git setting of safecrlf = true

### Possible Solution(s)

We could re-establish a check in `write_pkg_lockfile` to detect and match line endings, as was done in the original PR linked above. I'd be happy to take a crack at this if that sounds good.

### Notes

If helpful, git version says: `git version 2.41.0.windows.3`. 

### Version

```text
cargo 1.73.0
release: 1.73.0
host: x86_64-pc-windows-msvc
libgit2: 1.6.4 (sys:0.17.2 vendored)
libcurl: 8.2.1-DEV (sys:0.4.65+curl-8.2.1 vendored ssl:Schannel)
os: Windows 10.0.22621 (Windows 11 Enterprise) [64-bit]
```

weihanglo:
[...]

PR #13339, opened 2024-01-23
fix: Cargo.lock generation matches LF/CRLF line endings

JarlEvanson:
This PR attempts to resolve issue #12897. If a Cargo.toml is present, it detects if a CRLF line ending ends the first line, and then outputs the corresponding line ending when generating the lockfile.

The additional test should test for the CRLF adherence, just as the `preserve_line_endings_issue_2076` test used to do before we started testing for equality based on lines.

### Additional Notes
If this approach proves successful, the `Emitter` struct could be used to ensure compliance in other auto-generated files that contain line endings that should maintain the previous line ending style, such as the `cargo add` bug discussed in the issue #12897 thread.


epage (changes requested):
Maybe I misunderstood but this goes in a different direction than what is discussed in #12897 and I would recommend proposing it there and for us to resolve that before moving forward with an implementation review.

Issue #13308, opened 2024-01-16
Switch public/private dependencies to a non-blocking feature gate


epage:
#13307 updates our documentation to distinguish blocking and non-blocking feature gates.  Currently, public dependencies use a blocking feature gate and should be switched to a non-blocking feature gate.

When discussing this as a team, we tried to balance
- Avoiding MSRV bumps for using the feature, allowing earlier adoption
- How long this feature has until it is stabilized to make much of a difference for MSRVs

Due to some outstanding rustc bugs, the timeline for stabilization is out of our hands and so we feel moving forward with switching to non-blocking gating is worth it.

Parts of this change include
- Switching the `cargo-features` to a `-Z`
- Warning if `public` is used without `-Z` and setting it to false
  - We should also warn if the data type changes but that is less likely to happen and could possibly be skipped
- Ensuring the published version of the package does not have `public`

epage:
@linyihai in case you didn't see this.  If not interested at this time, thats fine.

linyihai:
> @linyihai in case you didn't see this. If not interested at this time, thats fine.

I didn't know I can make a hand. Thanks for you mention.

@rustbot claim

linyihai:
before I deep into this issue, I had glanced the code roughly. I still have some question.

> Switching the cargo-features to a -Z

Firstly, I can refer to PR https://github.com/rust-lang/cargo/pull/13296 to add the `-Zpublic-dependency`. 

The question is whether I replace  `cargo-features` with `-Zpublic-depenedncy` or not,  that means I will
-  delete the code `Feature::public_dependency()` and where use it.
-  some test case need to delete  `cargo-features = ["public-dependency"]` in their Cargo.toml

> Ensuring the published version of the package does not have public

This is a little vague to me. Do you mean check the code in `cargo-publish`  subcommand?

> https://github.com/rust-lang/cargo/pull/13307 updates our documentation to distinguish blocking and non-blocking feature gates.

Besides, it would be great if there were examples like this in the past.

@epage  Please give me some guidance, thanks you a lot.

linyihai:
> Warning if public is used without -Z and setting it to false

I think you intention is only warning when run `cargo add --public` and `cargo add --no-public` without the `-Zpublic-dependency`. 
if no `-Zpublic`,  we should warning people and ingore `--public` and `--no-public`.

epage:
> The question is whether I replace cargo-features with -Zpublic-depenedncy or not, that means I will

[...]

PR #13340, opened 2024-01-23
feat: Add "-Zpublic-dependency" for public-dependency feature.

linyihai:
Part of https://github.com/rust-lang/cargo/issues/13308, include:
- Switching the cargo-features to a -Z
- Warning if public is used without -Z and setting it to false

These had not done yet:
- We should also warn if the data type changes but that is less likely to happen and could possibly be skipped
- Ensuring the published version of the package does not have public

All test case should be pass.

### Additional information

r? @epage


src/cargo/core/features.rs
```diff
@@ -764,6 +764,7 @@ unstable_cli_options!(
     panic_abort_tests: bool = ("Enable support to run tests with -Cpanic=abort"),
     precise_pre_release: bool = ("Enable pre-release versions to be selected with `update --precise`"),
     profile_rustflags: bool = ("Enable the `rustflags` option in profiles in .cargo/config.toml file"),
+    public_dependency: bool = ("Enable public-dependency to allow marking dependencies as public/private"),
```
epage, line 767: nit
```suggestion
    public_dependency: bool = ("Respect a dependency's `public` field in Cargo.toml to control public/private dependencies"),
```

src/cargo/ops/cargo_package.rs
```diff
@@ -911,12 +911,7 @@ fn run_verify(
...
 
-    let rustc_args = if pkg
-        .manifest()
-        .unstable_features()
-        .require(Feature::public_dependency())
-        .is_ok()
-    {
+    let rustc_args = if ws.config().cli_unstable().public_dependency {
         // FIXME: Turn this on at some point in the future
         //Some(vec!["-D exported_private_dependencies".to_string()])
```
epage, line 916: We need to make sure we're tracking this

src/cargo/util/toml/mod.rs
```diff
@@ -1944,8 +1964,6 @@ fn detailed_dep_to_dependency<P: ResolveToPath + Clone>(
     }
 
     if let Some(p) = orig.public {
-        cx.features.require(Feature::public_dependency())?;
-
         if dep.kind() != DepKind::Normal {
             bail!("'public' specifier can only be used on regular dependencies, not {:?} dependencies", dep.kind());
```
epage, line 1968: Is this before or after we set `public = false`?

If before, we should probably check the unstable status and report this as a warning that this will fail when stabilized.  Or in other words, without the `-Z`, users should get the experience they have before any work was done on this but be warned what will break

src/cargo/util/toml/mod.rs
```diff
@@ -105,61 +105,62 @@ fn read_manifest_from_str(
...
-        let mut key = String::new();
-        stringify(&mut key, &path);
-        unused.insert(key);
-    }) {
-        Ok(manifest) => manifest,
-        Err(e) => {
-            let Some(span) = e.span() else {
-                return Err(e.into());
-            };
+    let mut manifest: manifest::TomlManifest =
```
epage, line 108: nit: ... so all of this was changed just from adding `mut`?  Ouch.

Might be worth a `let mut manifest = manifest` or a `let manifest = {... do changes; manifest};` to reduce churn

src/cargo/util/toml/mod.rs
```diff
@@ -183,10 +184,33 @@ fn read_manifest_from_str(
...
+                }
+            }
+        }
+    }
+    let public_warn = |warnings: &mut Warnings| {
+        for name in public_deps_without_z {
+            warnings.add_warning(format!(
+                "{name} is public, pass `-Zpublic-dependency` to enable support for it",
+            ))
+        }
```
epage, line 206: Why is this split into two separate passes, rather than adding the warnings while we make the changes?

src/cargo/util/toml/mod.rs
```diff
@@ -183,10 +184,33 @@ fn read_manifest_from_str(
...
         }
     }
+
+    let mut public_deps_without_z = BTreeSet::new();
+    if let Some(deps) = &mut manifest.dependencies {
+        for (name, dep) in deps {
+            if let InheritableDependency::Value(toml_dep) = dep {
+                if toml_dep.is_public() && !config.cli_unstable().public_dependency {
+                    if let TomlDependency::Detailed(d) = toml_dep {
+                        d.public = Some(false)
```
epage, line 194: Should we set it to `Some(false)` or should we unconditionally set `d.public = None;`?

epage (commented):

ehuss:
@weihanglo, just a heads up, I believe this will cause problems in the rust-lang/rust workspace. Can you coordinate with the release team to let them know, and perhaps help with the transition? This will need to be done during the next beta branch when bootstrap is updated. There are potentially several places where this `-Z` flag (or environment variable) will need to be added (anywhere in the bootstrap code that runs cargo, like vendoring, etc.). Worst case it will generate warnings, but I think it would be good to avoid those.

In general I'm concerned that this will still generate a bunch of warnings (such as running cargo manually). I'm curious if it is feasible to still honor `cargo-features`? I realize that might make the code more complicated, just checking how hard that would be.

Is it intentional that `public=false` does not generate a warning?

src/cargo/util/toml/mod.rs
epage: Note that tracking `.is_some()` and switching to `None` would also fix the problem ehuss pointed out

> Is it intentional that public=false does not generate a warning?

epage:
> @weihanglo, just a heads up, I believe this will cause problems in the rust-lang/rust workspace. Can you coordinate with the release team to let them know, and perhaps help with the transition? This will need to be done during the next beta branch when bootstrap is updated

If its a problem, we could instead do a transition where we temporarily support both or, as you mention elsewhere, generally support both.  We had talked about a transition but were unsure what the impact would be.

If there are projects that use this enough that they are fine with the `cargo-features` then I'm game.  Our main interest was to avoid an MSRV bump when using this.

src/cargo/util/toml/mod.rs
linyihai: The `public_warn`  is a closure function which need a `Manifest.warnings` as a input param.
The `public_warn` will invoke in latter, because the `Manifest.warnings` was produced by `to_real_manifest` or `to_virtual_manifest`.

Besides,  may be this need a `BTreeSet` to deduplicate.

The `public_warn` is similar to `add_unused`, I can combine them into one.

weihanglo:
Created a zulip topic here: <https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/-Zpublic-dependency.20syntax.20change>.

Supposedly editors tools like intellij-rust, rust-analyzer, and neovim plugins might rely on `cargo metadata` to analyze libstd from sysroot, we may want to have a short transition period to avoid too many warnings. One or two versions is sufficient I guess.

src/cargo/util/toml/mod.rs
linyihai: I had migrated the  code to here, because `detailed_dep_to_dependency` will invoke by `read_manifest_from_str`.

src/cargo/util/toml/mod.rs
linyihai: > I had migrated the code to here, because detailed_dep_to_dependency will invoke by read_manifest_from_str.
These code had deleted.

src/cargo/util/toml/mod.rs
linyihai: > I had migrated the code to here, because detailed_dep_to_dependency will invoke by read_manifest_from_str.

These code had deleted.

src/cargo/util/toml/mod.rs
linyihai: > I had migrated the code to here, because detailed_dep_to_dependency will invoke by read_manifest_from_str.

These code had deleted.

linyihai:
> Is it intentional that public=false does not generate a warning?

I used to think so, :-） but in fact i am not correct.  

 I think the warning shoud be emit in the situation
- people don't pass -Zpublic-dependency to Cargo when they use public specifer

src/cargo/util/toml/mod.rs
```diff
@@ -1944,13 +1939,18 @@ fn detailed_dep_to_dependency<P: ResolveToPath + Clone>(
...
-
         if dep.kind() != DepKind::Normal {
             bail!("'public' specifier can only be used on regular dependencies, not {:?} dependencies", dep.kind());
         }
 
-        dep.set_public(p);
+        if !cx.config.cli_unstable().public_dependency {
+            cx.warnings.push(format!(
+                "{name} uses 'public' specifier, pass `-Zpublic-dependency` to enable support for it",
+                name = dep.name_in_toml(),
```
linyihai, line 1949: Is nessary to set the dep.public as Nonek or false ?

src/cargo/util/toml/mod.rs
```diff
@@ -1944,13 +1939,18 @@ fn detailed_dep_to_dependency<P: ResolveToPath + Clone>(
     }
 
     if let Some(p) = orig.public {
-        cx.features.require(Feature::public_dependency())?;
-
         if dep.kind() != DepKind::Normal {
             bail!("'public' specifier can only be used on regular dependencies, not {:?} dependencies", dep.kind());
```
epage, line 1943: Generally, this should be a warning and not an error when the unstable feature is inactive.  Unsure if its worth it though in this case.

src/cargo/util/toml/mod.rs
epage: Leaving `dep.set_public` alone is likely the right call as that is what was done before without this unstable feature.

src/cargo/util/toml/mod.rs
epage: > If before, we should probably check the unstable status and report this as a warning that this will fail when stabilized. 

We still aren't downgrading this `bail` to a warning for stable.

src/cargo/util/toml/mod.rs
epage: We are now never setting `public = None` in the manifest.  This means that `public` will leak out on `cargo package` / `cargo publish`.

linyihai: Your are right, should set `public = None` here

src/cargo/util/toml/mod.rs
linyihai: > Why is this split into two separate passes, rather than adding the warnings while we make the changes?

This is similar to `add_unused`.  It can't warning in the same passes, because we can't get `Warning` instance now.

src/cargo/util/toml/mod.rs
linyihai: > CARGO_RUN_BUILD_STD_TESTS=true cargo test -p cargo --test build-std

the `build-std` will fail, because they don't pass `-Zpublic-dependency`, detail :https://github.com/rust-lang/cargo/actions/runs/7665093719/job/20890409198?pr=13340

src/cargo/ops/cargo_package.rs
linyihai: How about add it to the tracking issue or make a new issue for it?

src/cargo/util/toml/mod.rs
```diff
@@ -1944,7 +1965,17 @@ fn detailed_dep_to_dependency<P: ResolveToPath + Clone>(
     }
 
     if let Some(p) = orig.public {
-        cx.features.require(Feature::public_dependency())?;
```
weihanglo, line 1947: Brought back [feedback from t-release](https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/-Zpublic-dependency.20syntax.20change/near/418323741) for <https://github.com/rust-lang/cargo/pull/13340#issuecomment-1907284634>. We might want to support `cargo-features = ["public-dependency"]` for at least one release cycle, so it won't break self-rebuild. Hence this should be kept I believe (and probably other places `cli_unstable().public_dependency` gate are used as well).

Would be good if we have a test verifying it won't bail out or warn in this case.

linyihai: I had addded a new method named `is_public_dependency_enabled` to support ``support cargo-features = ["public-dependency"].

New test cases had addded to `pub_priv.rs`.  (there is one test for the mixture of -Z and cargo-features.)

src/cargo/core/features.rs
```diff
@@ -623,6 +623,12 @@ impl Features {
         &self.activated
     }
 
+    /// Temporary workaround for supporting `cargo-features = ["public-dependency"]`.
+    /// FIXME: Need to remove after at least on release cycle.
+    pub fn is_public_dependency_enabled(&self) -> bool {
+        Feature::public_dependency().is_enabled(self)
+    }
```
epage, line 630: Its not clear to me why this function is needed vs just doing what we did before

epage: (generally, when something is short-term, I find it best to be surgical, touching as little as possible)

src/cargo/util/toml/mod.rs
```diff
@@ -168,6 +196,13 @@ fn read_manifest_from_str(
             }
         }
     };
+    let public_warn = |warnings: &mut Warnings| {
+        for name in public_deps_without_z {
+            warnings.add_warning(format!(
+                "{name} uses 'public' specifier, pass `-Zpublic-dependency` to enable support for it",
```
epage, line 202: nit:
```suggestion
                "Ignoring `public` on dependency {name}.  Pass `-Zpublic-dependency` to enable support for it",
```

src/cargo/util/toml/mod.rs
```diff
@@ -198,6 +234,7 @@ fn read_manifest_from_str(
     } else {
         let (mut m, paths) = to_virtual_manifest(manifest, source_id, package_root, config)?;
         add_unused(m.warnings_mut());
+        public_warn(m.warnings_mut());
```
epage, line 237: virtual manifests can't have `[dependencies]`

src/cargo/util/toml/mod.rs
```diff
@@ -1944,7 +1981,19 @@ fn detailed_dep_to_dependency<P: ResolveToPath + Clone>(
     }
 
     if let Some(p) = orig.public {
-        cx.features.require(Feature::public_dependency())?;
+        if !cx.config.nightly_features_allowed {
+            bail!(
+                "Consider trying a newer version of Cargo (this may require the nightly release)."
+            );
+        }
```
epage, line 1988: What is this error check for?  Not seeing any tests for it.

src/cargo/core/features.rs
linyihai: My original idea was that this function would be used to check whether the public-dependency feature was enabled. If not, we can warn it and encourage people to use `-Zpublic-dependency`.

Supporting both approaches seems to be a tricky problem...

src/cargo/util/toml/mod.rs
```diff
@@ -104,13 +104,41 @@ fn read_manifest_from_str(
...
     let deserializer = toml::de::Deserializer::new(contents);
     let manifest: manifest::TomlManifest = match serde_ignored::deserialize(deserializer, |path| {
         let mut key = String::new();
         stringify(&mut key, &path);
         unused.insert(key);
     }) {
-        Ok(manifest) => manifest,
+        Ok(manifest) => {
+            let mut manifest: manifest::TomlManifest = manifest;
+            let is_public_dependency_enabled = {
```
linyihai, line 117: If no `public-dependency` and `-Zpublic-dependency`,  set public to None.
Then will warn people and encourage people to `-Zpublic-dependency`.

This behavior had test case to verify.

src/cargo/util/toml/mod.rs
```diff
@@ -1944,7 +1980,13 @@ fn detailed_dep_to_dependency<P: ResolveToPath + Clone>(
     }
 
     if let Some(p) = orig.public {
-        cx.features.require(Feature::public_dependency())?;
+        if !cx.config.cli_unstable().public_dependency
+            && !cx.features.require(Feature::public_dependency()).is_ok()
+        {
+            bail!(format!(
+                "Ignoring `public` on dependency {name}.  Pass `-Zpublic-dependency` to enable support for it", name = dep.name_in_toml()
```
linyihai, line 1987: `read_manifest_from_str` will not step into here, because had set orig.public = None before.

There are calls from other places, but I'm not sure

src/cargo/util/toml/mod.rs
linyihai: Had addressed.

epage: I'm not seeing how.  Could you clarify?  Depending on order, we either fire off the error or we are silent.  We should probably have a test for this.

linyihai: For stable, `bail_public_on_stable` can verify only emit error on stable.
For nightly,  `verify_z_public_dependnecy` can verify emit  warn if  no `-Zpublic-dependency` and keep silenc if pass `-Zpublic-dependency` on nightly.

I think the only thing that's different now is that the nightly version displays a warning instead of a direct error.

epage: Can you add a variant of the `pub_dev_dependency` test that runs without the unstable feature being enabled?

linyihai: Thanks for the heads-up. Similar tests have been added.

(`pub_dev_dependency_without_feature_in_stable` and `pub_dev_dependency_without_feature_in_nightly`)

epage: So those tests show that we are erroring on this unstable feature when we should instead warn that this will error out in the future when its stabilized.

I will say its a separate question of whether its worth fixing (depends on how intrusive the fix is).  The general philisophy of this approach to unstable features is they should act like you specified the keys back in <1.30.0 but warn people if there will be a breakage on stabilization.

linyihai: em, it seems to be blocking here.
If you look at the logic here again, both the stable and nightly versions are throwing errors.
Can we keep throwing errors on nightly if no -Z or features enabled?

epage: Generally, code activates with the feature being enabled, not just if nightly is used.

src/cargo/util/toml/mod.rs
```diff
@@ -716,6 +716,21 @@ pub fn to_real_manifest(
...
+                        if !cx.features.require(Feature::public_dependency()).is_ok()
+                            && !cx.config.cli_unstable().public_dependency
+                        {
+                            cx.warnings.push(format!(
+                                "Ignoring `public` on dependency {name}.  Pass `-Zpublic-dependency` to enable support for it", name = &dep.name_in_toml()
+                            ))
+                        }
+                    }
+                }
+            }
```
linyihai, line 733: the warn message emit here and disable dependency public  in the same time.

src/cargo/util/toml/mod.rs
linyihai: Please see https://github.com/rust-lang/cargo/pull/13340/files#diff-3d8669050f6f1c07884ef5a4e9d344a98c7270f73f4683125a00e8a3cc6381e6R1963.

tests/testsuite/pub_priv.rs
```diff
@@ -198,6 +193,47 @@ Caused by:
...
+        .with_stderr(
+            "\
+[ERROR] failed to parse manifest at `[..]`
+
+Caused by:
+  'public' specifier can only be used on regular dependencies, not Development dependencies
+",
+        )
+        .run()
+}
```
epage, line 235: Now that we have a test to more concretely talk about this, I'm going to continue on [this comment thread](https://github.com/rust-lang/cargo/pull/13340#discussion_r1463618205) here.

src/cargo/util/toml/mod.rs
epage: Now that we have a test, I'm moving the discussion to https://github.com/rust-lang/cargo/pull/13340/files#r1498114047

tests/testsuite/pub_priv.rs
epage: > ... we should probably check the unstable status and report this as a warning that this will fail when stabilized. Or in other words, without the -Z, users should get the experience they have before any work was done on this but be warned what will break

Our options are
1. Validate and error like we do here
2. Validate but warn on stable
3. Do nothing, relying only on the "this isn't stable / unused" warning

This takes (1).  I had thought `[lints]` went with (2) but it in looking at [the code](https://github.com/rust-lang/cargo/blob/0.74.0/src/cargo/util/toml/mod.rs#L2833), it went with (3).  Unsure why I changed my mind.

We should make a conscious choice about what route we go.
- The downside to errors (1) is that changes in unstable behavior affect users on stable.
- Warning (2) can help people on mixed stable/nightly detect changes earlier
- For stable-only people with an MSRV, the warning (2) will be unactionable and annoying

linyihai:
> ☔ The latest upstream changes (presumably #13409) made this pull request unmergeable. Please [resolve the merge conflicts](https://rustc-dev-guide.rust-lang.org/git.html#rebasing-and-conflicts).

Conficts had been addressed.

tests/testsuite/pub_priv.rs
linyihai: `that changes in unstable behavior affect users on stable.`  
Not quite sure how this happened, and could it have been avoided?

linyihai: > So those tests show that we are erroring on this unstable feature when we should instead warn that this will error out in the future when its stabilized.

```
  if dep.kind() != DepKind::Normal {
            bail!("'public' specifier can only be used on regular dependencies, not {:?} dependencies", dep.kind());
```
Change this to a warning?

epage: > Not quite sure how this happened, and could it have been avoided?

If someone is using an unstable feature on a specific nightly but wants it to work generally on stable, we need to be careful.  However, that is more about introducing new errors.  Removing an error (the main risk here) wouldn't be breaking.

> Change this to a warning?

When the feature isn't enabled?  That is option (2)

linyihai: I had made a new commit.
if a dependency isn't normal dependency,  warn `public` used in dev-dependency when no feature enabled in nightly.   we should to skip set `dep.set_public(p)`, and set `orig.public = None`.

epage:
@bors r+

Thanks for your work on this and patience as we iterated

src/cargo/util/toml/mod.rs
```diff
@@ -1992,15 +2013,23 @@ fn detailed_dep_to_dependency<P: ResolveToPath + Clone>(
...
+        let with_public_feature = public_feature.is_ok();
+        if !with_public_feature && (!with_z_public && !manifest_ctx.gctx.nightly_features_allowed) {
+            public_feature?;
+        }
 
         if dep.kind() != DepKind::Normal {
-            bail!("'public' specifier can only be used on regular dependencies, not {:?} dependencies", dep.kind());
+            match (with_public_feature, with_z_public) {
+                (true, _) => bail!("'public' specifier can only be used on regular dependencies, not {:?} dependencies", dep.kind()),
+                (_, true) => bail!("'public' specifier can only be used on regular dependencies, not {:?} dependencies", dep.kind()),
```
weihanglo, line 2026: nit: de-duplicate error messages

src/cargo/util/toml/mod.rs
```diff
@@ -727,6 +727,27 @@ pub fn to_real_manifest(
...
+                    if matches!(dep.kind(), DepKind::Normal) {
+                        if !manifest_ctx
+                            .features
+                            .require(Feature::public_dependency())
+                            .is_ok()
+                            && !manifest_ctx.gctx.cli_unstable().public_dependency
+                        {
+                            d.public = None;
+                            manifest_ctx.warnings.push(format!(
+                            "Ignoring `public` on dependency {name}.  Pass `-Zpublic-dependency` to enable support for it", name = &dep.name_in_toml()
```
weihanglo, line 742: nit: convention is starting with a lowercase letter

tests/testsuite/pub_priv.rs
```diff
@@ -198,6 +193,45 @@ Caused by:
...
+            pub fn use_pub(_: pub_dep::FromPub) {}
+        ",
+        )
+        .build();
+
+    p.cargo("check --message-format=short")
+        .masquerade_as_nightly_cargo(&["public-dependency"])
+        .with_stderr(
+            "\
+[WARNING] 'public' specifier can only be used on regular dependencies, not Development dependencies
```
weihanglo, line 227: This was not introduced by this PR, but capitalized "Development dependencies" doesn't seem ideal.

epage:
Most of these messages will be removed with stabilization, so at least to me, DRYing is not a priority and can be harmful.  I at least also care a little less about the quality of the messages.

weihanglo:
In case I forget, please remind me to update <https://github.com/rust-lang/rust/pull/121710> when this PR gets into beta release :)

src/cargo/util/toml/mod.rs
linyihai: It reminds me that I can do this：
`(true, _) | (_, true) => bail!("'public' specifier can only be used on regular dependencies, not {:?} dependencies", dep.kind()),`

PR #13347, opened 2024-01-25
[WIP] feat(cargo-vendor): vendor path dep if it is not in any given workspaces

### LINKED ISSUE #9172: Cargo vendor with patch section and local sources does not vendor the local files

weihanglo:

feat(cargo-vendor): vendor path dep if it is not in any given workspaces

Generally cargo don't vendor path dependencies.
This seems quiet reasonable path dependencies are "local" comparing
to git or registry dependencies, and usually under the user's control.
However, it is not always the case.

A workspace might contain

* any `[patch]` to local path dependencies
* a set of shared path dependencies outside the current workspace

These use cases demonstrate that users might not have controls or
permissions to those dependencies. When they want to create a
reproducible tarball for their own workspace, `cargo vendor` is not a
tool helping them achieve the goal.

There is one workaround: Have a `[patch]` to a local git repository
instead of a lcoal path dependency. This is not ergonomic and adds
overhead of setting git repositories.

This PR proposes that Cargo vendors path dependencies if they are
not belong to any given workspaces.

As a side effect, this exposes a new  `[source]` kind `path`:

```toml
[source."path+file:///path/to/package"]
path = "/path/to/package"
```

This is a proof-of-concept, not ready for serious code review.

### Additional information

An alternative to #12858
Fixes #9172
Possibly also #10134, but I am not sure if they intend to vendor workspace members.

### DIFF
--- a/src/cargo/ops/vendor.rs
+++ b/src/cargo/ops/vendor.rs
@@ -56,14 +56,13 @@ struct VendorConfig {
 }
 
 #[derive(Serialize)]
-#[serde(rename_all = "lowercase", untagged)]
+#[serde(rename_all = "kebab-case", rename_all_fields = "kebab-case", untagged)]
 enum VendorSource {
     Directory {
         directory: String,
     },
     Registry {
         registry: Option<String>,
-        #[serde(rename = "replace-with")]
         replace_with: String,
     },
     Git {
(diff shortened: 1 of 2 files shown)


weihanglo:
This is just a proof-of-concept so going to close. Further discussion should happen in #9172 or #10134.

PR #7295, opened 2019-08-24
Allow using 'config.toml' instead of just 'config' files.

zachlute:
Fixes #7273 

Note that this change only makes 'config.toml' optional to use instead of 'config'. If both exist, we will print a warning and prefer 'config', since that would be the existing behavior if both existed today.

We should also consider a separate change to make config.toml the default and update docs, etc.

PR #13349, opened 2024-01-26
fix(config): Deprecate non-extension files 

epage:

In #7295 (released in 1.39), we said we'd want to warn on use of
`.cargo/config` after about 6 months.  Over 4 years later, we are now
getting that warning.

This is important for addressing user confusion, like in
https://www.reddit.com/r/rust/comments/19fd5q2/cargoconfig/

It'll be important to look at the individual commits as one updates tests from using `.cargo/config` to `.cargo/config.toml` which touches a lot of code.

I added a test for `.cargo/config` in a separate commit so you can see how the output changes.

### Additional information

Discussed on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Deprecating.20credential.20provider.20default.3F)

### DIFF
--- a/crates/cargo-test-support/src/registry.rs
+++ b/crates/cargo-test-support/src/registry.rs
@@ -279,7 +279,7 @@ impl RegistryBuilder {
     /// Initializes the registry.
     #[must_use]
     pub fn build(self) -> TestRegistry {
-        let config_path = paths::home().join(".cargo/config");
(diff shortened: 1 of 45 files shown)


weihanglo (approved):
Looks good to me. Hope I didn't overlook something.

PR #13351, opened 2024-01-26
docs(ref): Try to improve reg auth docs

epage:

My end-goal *was* to discourage the use of `cargo:token` but it immediately failed for me and I need to look into that further.

So this is the smaller steps I made along the way.

### Additional information

For config.md, my expectation is to warn users about `cargo:token` right after the discussion on keeping tokens secret.

I want to find a way to talk about this on the "your first publish" page but that needs to wait until we smooth things out.


weihanglo (approved):
Looks reasonable. Thanks!

weihanglo:
Job cancelled with no reason?

@bors retry

PR #13361, opened 2024-01-28
refactor: remove unnecessary Option in `Freshness::Dirty`

weihanglo:

This encodes `Freshness::Dirty(None)` better with a new variant `DirtyReason::FreshBuild`.

The only behavior change in this PR is that source verification is
also performed for `DirtyReason::Forced` rebuild.

### DIFF
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -1750,7 +1750,25 @@ fn compare_old_fingerprint(
     }
 
     let compare = _compare_old_fingerprint(old_hash_path, new_fingerprint);
-    log_compare(unit, &compare);
+
+    match compare.as_ref() {
+        Ok(None) => {}
+        Ok(Some(reason)) => {
+            info!(
+                "fingerprint dirty for {}/{:?}/{:?}",
+                unit.pkg, unit.mode, unit.target,
+            );
+            info!("    dirty: {reason:?}");
+        }
+        Err(e) => {
+            info!(
+                "fingerprint error for {}/{:?}/{:?}",
+                unit.pkg, unit.mode, unit.target,
+            );
+            info!("    err: {e:?}");
+        }
+    }
+
     match compare {
         Ok(None) if forced => Some(DirtyReason::Forced),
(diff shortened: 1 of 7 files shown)


src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1755,10 +1740,15 @@ fn compare_old_fingerprint(
...
-) -> CargoResult<Option<DirtyReason>> {
+    forced: bool,
+) -> Option<DirtyReason> {
     let compare = _compare_old_fingerprint(old_hash_path, new_fingerprint, mtime_on_use);
     log_compare(unit, &compare);
-    compare
+    match compare {
+        Ok(None) if forced => Some(DirtyReason::Forced),
+        Ok(reason) => reason,
+        Err(_) => Some(DirtyReason::FreshBuild),
```
epage, line 1787: Should `FreshBuild` contain the error?

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1741,35 +1744,67 @@ fn compare_old_fingerprint(
     new_fingerprint: &Fingerprint,
     mtime_on_use: bool,
     forced: bool,
+    verbosity: Verbosity,
 ) -> Option<DirtyReason> {
-    let compare = _compare_old_fingerprint(old_hash_path, new_fingerprint, mtime_on_use);
-    log_compare(unit, &compare);
+    if mtime_on_use {
```
epage, line 1749: The other commits seem nice and could be merged on their own.  I feel like this adds complexity that, without a motivating reason (showing a noticeable difference on end-to-end benchmarks), seems of questionable value.

weihanglo: Sound reasonable. I was refactoring unnecessary Options and found it could be skipped. Will remove this part later.

src/cargo/core/compiler/fingerprint/mod.rs
weihanglo: This could be done in theory, but `anyhow::Error` doesn't implement `Clone`….

PR #13362, opened 2024-01-28
test: data layout fix for `x86_64-unknown-none-gnu`

weihanglo:
Tests are broken due to LLVM target layout check in rustc. See <https://github.com/rust-lang/rust/pull/120062>.


ehuss:
Thanks!

@bors r+

PR #13364, opened 2024-01-29
doc: Hide `cargo-fetch` description in offline man page

### LINKED ISSUE #13232: `cargo fetch --offline` documentation directs user to use `cargo fetch --offline`
### Problem

The documentation for `cargo fetch --offline` (the manual page) includes the words "See the cargo-fetch(1) command to download dependencies before going offline."

This seems needlessly self-recursive. In fact, I'm not sure why `cargo fetch` offers an offline mode, but maybe it's useful in some case.

### Steps

_No response_

### Possible Solution(s)

Drop the sentence directing the user to `cargo fetch --offline` from the description of `cargo fetch --offline`. I realize this is boilerplate shared with all `cargo` subcommands that have an `--offline` option, and I don't know if the template system used allows a way to exclude parts for specific commands.

### Notes

This is really not a big issue. I'd be happy for you to ignore this until all other issues have been resolved.

### Version

```text
release: 1.75.0
commit-date: 2023-11-20
os: Debian 12 (bookworm) [64-bit]
```

linyihai:
Fixes #13232

Suggest review commit by commit.

### Additional information
r? @weihanglo

### DIFF
--- a/src/doc/man/generated_txt/cargo-fetch.txt
+++ b/src/doc/man/generated_txt/cargo-fetch.txt
@@ -91,8 +91,7 @@ OPTIONS
            Beware that this may result in different dependency resolution than
            online mode. Cargo will restrict itself to crates that are
            downloaded locally, even if there might be a newer version as
-           indicated in the local copy of the index. See the cargo-fetch(1)
-           command to download dependencies before going offline.
+           indicated in the local copy of the index.
 
            May also be specified with the net.offline config value
            <https://doc.rust-lang.org/cargo/reference/config.html>.
--- a/src/doc/man/includes/options-locked.md
+++ b/src/doc/man/includes/options-locked.md
@@ -18,8 +18,10 @@ proceed without the network if possible.
 Beware that this may result in different dependency resolution than online
 mode. Cargo will restrict itself to crates that are downloaded locally, even
 if there might be a newer version as indicated in the local copy of the index.
+{{#if (ne command "fetch")}}
 See the {{man "cargo-fetch" 1}} command to download dependencies before going
 offline.
+{{/if}}
 
 May also be specified with the `net.offline` [config value](../reference/config.html).
 {{/option}}
--- a/src/doc/src/commands/cargo-fetch.md
+++ b/src/doc/src/commands/cargo-fetch.md
@@ -97,9 +97,7 @@ the network is not available. With this flag, Cargo will attempt to
 proceed without the network if possible.</p>
 <p>Beware that this may result in different dependency resolution than online
 mode. Cargo will restrict itself to crates that are downloaded locally, even
-if there might be a newer version as indicated in the local copy of the index.
-See the <a href="cargo-fetch.html">cargo-fetch(1)</a> command to download dependencies before going
-offline.</p>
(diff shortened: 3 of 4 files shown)


src/doc/man/cargo-fetch.md
```diff
@@ -2,6 +2,7 @@
 {{~*set command="fetch"}}
```
linyihai, line 2: > For each command there should be a command variable available, so guess we can check if it is fetch and exclude the self-referential part.

Some place not set command, like cargo-vendor.md, cargo-verify-project.md.

src/bin/cargo/commands/fetch.rs
```diff
@@ -9,6 +9,7 @@ pub fn cli() -> Command {
         .arg_silent_suggestion()
         .arg_target_triple("Fetch dependencies for the target triple")
         .arg_manifest_path()
+        .arg(flag("offline", "Run without accessing the network").hide(true))
```
epage, line 12: This *could* be used to test if all packages are fetched. Unsure if its worth showing for that reason or not.

If we hide it, should we also have the command say its deprecated when used?

src/doc/man/generated_txt/cargo-fetch.txt
```diff
@@ -82,21 +82,6 @@ OPTIONS
            Cargo.lock file is up-to-date (such as a CI build) or want to avoid
            network access.
 
-       --offline
-           Prevents Cargo from accessing the network for any reason. Without
-           this flag, Cargo will stop with an error if it needs to access the
-           network and the network is not available. With this flag, Cargo will
-           attempt to proceed without the network if possible.
-
-           Beware that this may result in different dependency resolution than
```
weihanglo, line 91: > Beware that this may result in different dependency resolution than
Comment online mode

Haven't tested it. If this statement is true, we may want to keep this, as it makes a huge difference between with and without `--offline`.

linyihai: > See the cargo-fetch(1)
           command to download dependencies before going offline.
           
I may have taken too big a step, and the original problem may have just been to remove this

src/bin/cargo/commands/fetch.rs
linyihai: Although my first impression is that this option is really not used, I think I'm solving the wrong problem.

src/doc/man/includes/options-locked.md
```diff
@@ -18,8 +18,10 @@ proceed without the network if possible.
 Beware that this may result in different dependency resolution than online
 mode. Cargo will restrict itself to crates that are downloaded locally, even
 if there might be a newer version as indicated in the local copy of the index.
+{{#if (ne no-fetch-in-offline true)}}
```
weihanglo, line 21: Would this work?

```suggestion
{{#if (ne command "fetch")}}
```

linyihai: Yes, it works.   I should have tried this before, but it didn't work.

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #13367, opened 2024-01-29
fix(new): Print a 'Creating', rather than 'Created' status

epage:

This has bothered me about `cargo new` and `cargo init` for a while that
the output is read backwards, for example:
```diff
--- i/tests/testsuite/cargo_init/path_contains_separator/stderr.log
+++ w/tests/testsuite/cargo_init/path_contains_separator/stderr.log
@@ -1,3 +1,3 @@
+    Creating binary (application) package
 warning: the path `[ROOT]/case/test:ing/.` contains invalid PATH characters (usually `:`, `;`, or `"`)
 It is recommended to use a different name to avoid problems.
-     Created binary (application) package
```


weihanglo (approved):
This seems reasonable, though not sure if we need a “Created” status to indicate the success.

Issue #5151, opened 2018-03-08
Tracking issue: Cargo templates


carols10cents:
Some history: There was an effort to do this that got reverted pending an RFC:

- https://github.com/rust-lang/cargo/issues/396
- https://github.com/rust-lang/cargo/pull/3004
- https://github.com/rust-lang/cargo/pull/3878

This is a new tracking issue.

Related: #3506

dwijnand:
Also relevant: #5114, initialising an examples directory for lib crates.

sanmai-NL:
See https://github.com/rust-lang/rfcs/pull/2549#issuecomment-433712561.

weihanglo:
See also #12769 proposing adding educational comments to `Cargo.toml` generated by `cargo new/init`.

Kobzol:
I created an interactive configurator for Cargo projects (focused on performance aspects of configuration), might be a bit relevant: https://github.com/Kobzol/cargo-wizard.

epage:
Crazy idea for a dead-simple template: allow the user to select an example from a package.

The main question would be dealing how to deal with multi-file examples.  Maybe if it is named `main.rs` or `lib.rs`, we pull the whole directory over while `topic.rs` gets written as `src/main.rs`?

dradetsky:
@epage To me, the obvious way to do this is: just copy the template package as written into the newly-created package. The newly created package has a file named `src/main.rs` if and only if the template package has such a file too. Do absolutely no renaming of files unless somehow specified by the user (assuming the template system provides a facility to rename template files, which it might not want to).

This way, it's easy (or easier) for me to specify any kind of package template I want: I just do the same thing I do normally to make a package by hand.

The argument against this approach would be: Given that users would probably want some kind of functionality to replace variables in the templated package, would the above approach interact badly with variable templating? I don't actually know since I haven't really thought about it. I don't immediately see why though.

boozook:
I'm not sure if templates are necessary _in general_, 
but personally, I really need a __working__ integration with another (non-git) VCS (like Pijul, jj, etc..). By "working" I mean something like how `cargo new`/`init`, `cargo fix`, `cargo publish` and `cargo update`(for git deps) currently work with Git.

Perhaps it would be enough to be able to specify all the necessary shell-commands in the config toml — for any VCS or service.

epage:
@boozook we have some dedicated issues for this like #12102, #9729.

[...]

PR #13371, opened 2024-01-30
fix(new): Print a note, rather than a comment, for more information

### LINKED ISSUE #12210: `cargo init` add option to remove default comment
### Problem

While initializing a new project using `cargo init`, a comment is automatically inserted at the top of the `Cargo.toml` file:

`# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html`

Although this is undoubtedly a helpful resource for newcomers and those unfamiliar with Cargo, for seasoned Rust developers this is not necessary and it adds an additional manual step of removing it every time a new project is initialized. Especially for developers working with automated systems or setting up multiple projects, the comment can be slightly annoying.

### Proposed Solution

I propose adding an option to the `cargo init` command, let's say `--no-comment`, that when used will not include the standard commented URL at the top of the `Cargo.toml` file. The command would look like this:

`cargo init --no-comment`

By providing this option, developers have the flexibility to opt-in or opt-out of having the comment in their `Cargo.toml` file as per their requirements or preferences. It wouldn't impact beginners as the default behavior would still be to include the comment, unless explicitly opted out using the `--no-comment` flag.

### Notes

I believe this minor enhancement would improve the developer experience by reducing manual interventions, particularly for those creating a large number of Rust projects or using automated tools to initialize Rust projects. This change should not affect the functionality of `cargo init`, and it would be backward compatible since it just introduces a new optional flag.

Please consider this request and let me know if you need any additional information. I am willing to try to implement it. Thanks!

epage:

For myself, I find it frustrating when tools add boilerplate that I have to remove after every use.  A `note:` on the command-line is more ephemeral, helping people who need it while not interfering with those that don't.

Fixes #12210

### DIFF
--- a/tests/testsuite/workspaces.rs
+++ b/tests/testsuite/workspaces.rs
@@ -1063,7 +1063,10 @@ fn new_creates_members_list() {
     let p = p.build();
 
     p.cargo("new --lib bar")
-        .with_stderr("    Creating library `bar` package")
+        .with_stderr("\
+[CREATING] library `bar` package
+[NOTE] see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+")
         .run();
 }
 
@@ -1082,6 +1085,7 @@ fn new_warning_with_corrupt_ws() {
   |
 [WARNING] compiling this new package may not work due to invalid workspace configuration
 
+[NOTE] see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 ",
         )
         .run();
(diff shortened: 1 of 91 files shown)


tests/testsuite/workspaces.rs
```diff
@@ -1074,6 +1077,7 @@ fn new_warning_with_corrupt_ws() {
         .with_stderr(
             "\
 [CREATING] binary (application) `bar` package
+[NOTE] see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
weihanglo, line 1080: Could this be printed only when Cargo succeeds to create a new package?

epage: In this case it does succeed, its just warning about the state of the workspace.

But I can move it later in case a spurious error happens through the process.

weihanglo (approved):

weihanglo:
Though I still think the comment is helpful, I am under the impression that we're on the way towards #5151 and avoid adding more comments or Rust files for `cargo new`. 

Hence @bors r+

Issue #7167, opened 2019-07-23
`cargo update` should print warnings for incompatible candidates


ehuss:
Currently, if there is a new release of a package that removes a required feature, then `cargo update` will silently do nothing.  This can be a confusing experience because if the user looks at crates.io and sees there is a newer version, it can be difficult to understand why it is not working.

I believe that removing a feature is a breaking semver change, but not everyone knows this or is diligent about managing correct version bumps.

I think `cargo update` should display a warning if the "newest" candidate is not used because it is missing a feature.  

It should probably ignore intermediate candidates, for example starting from version 0.1.0, publish bad release 0.1.1, publish fixed release 0.1.2, then updating from 0.1.0 to 0.1.2 should not produce any warnings.

There are other edge cases to handle (like being able to update from 0.1.0 to 0.1.1, but not 0.1.2).

The following is a test to demonstrate the scenario.  The warning at the bottom is just a suggestion, the actual warning will likely be a little different.

```rust
#[cargo_test]
fn update_with_missing_feature() {
    // Attempting to update a package to a version with a missing feature
    // should produce a warning.
    Package::new("bar", "0.1.0").feature("feat1", &[]).publish();

    let p = project()
        .file(
            "Cargo.toml",
            r#"
            [package]
            name = "foo"
            version = "0.1.0"

            [dependencies]
            bar = {version="0.1", features=["feat1"]}
            "#,
        )
        .file("src/lib.rs", "")
        .build();
    p.cargo("generate-lockfile").run();

    // Publish an update that is missing the feature.
    Package::new("bar", "0.1.1").publish();

    p.cargo("update")
        .with_stderr(
            "\
[WARNING] a newer version of package `bar` is available (`0.1.1`), \
but it does not include the required feature `feat1`, so it will not be updated
",
        )
        .run();

    // Publish a fixed version, should not warn.
    Package::new("bar", "0.1.2").feature("feat1", &[]).publish();
    p.cargo("update")
        .with_stderr(
            "\
[UPDATING] `[..]` index
[UPDATING] bar v0.1.0 -> v0.1.2
",
        )
        .run();
}
```

niedzielski:
Extending the "incompatible candidates" criteria, `cargo update` only considers dependencies that are permitted by the semantic version specified (often in the Cargo.toml). This means that newer forbidden dependencies are not reported and the caller is never made aware that a more involved upgrade is necessary to use the latest and greatest. E.g., consider

```lang=toml
[dependencies]
[...]

Issue #4309, opened 2017-07-20
Cargo features to help keeping dependencies up-to-date


behnam:
In Rust ecosystem, one main concern for library developers is to not have semver breaking updates to their package, because of the fear that majority of the dependent packages not getting the routing updates that would follow a breaking update.

There are two general solutions to this problem:

1. The old-school back-porting of updates, so that dependents on any API version would get all the routine updates. This solution can easily become a huge maintenance cost on library developers and is not a common practice in the Rust community at the moment.

2. Help owner of dependent packages to update their dependency versions and apply the changes needed, if any. This solution can be much less time consuming, and distributes the workload throughout the dependency chain. It also means that only the versions being used get the maintenance work needed, but not all theoretically-possible versions. So, in rustacean words: *you pay for what you use*.

Based on this, I want to propose to add new features to help owners with keeping their dependencies up-to-date. Most of the functionality can be exposed via the `cargo update` command.

## Proposed Features

* **Warn on dependencies stuck in older versions.** This would be a plain warning on `cargo update`. The feature would be enabled by default, but there will be an option to disable it; and there can be one release version of Cargo as transition period, where the option is disabled by-default, but there's a notice about the upcoming change, if it applies.

* **Commands to auto-update some/all dependency versions to latest.** This option would *update* the relevant manifests to use a latest version of a single or multiple packages. This can also be an option to `cargo update`, similar to `--precise`, but instead of updating the lock file, it would also update the manifest with the latest version, or some user-provided version.

  This is a tool to update Cargo Manifest files. Currently, `cargo update` only works with the Cargo Lock file, so it may be a good idea to have a separate sub-command for the manifest *manipulation* features. But since it's a core functionality, it's better have this functionality in Cargo proper.

* **Auto-update versions to a successful build.** This is something that we can aim for in longer term, to have cargo update dependencies, and in case of failures, bisect and find the most-recent version that does not break the builds and/or tests. Best place for this may be a third-party cargo command and not cargo itself.

What do you think?

behnam:
I'm sure there are many examples where these can be helpful. Here's one that inspired me to write up this proposal: https://github.com/servo/unicode-bidi/issues/46#issuecomment-316778436

/cc @alexcrichton, @Manishearth, @anowell

alexcrichton:
[...]

PR #13372, opened 2024-01-31
feat(update): Tell users when they are still behind

epage:

Part of this is an offshoot of #12425 which is about pulling  some of `cargo upgrade`s behavior into `cargo update`.  One of the "'Potential related `cargo update` improvements" is informing the user when they are behind.

Part of this is to help close the gap of users being behind on their dependencies unaware.  This is commonly raised when discussing an MSRV-aware resolver (see rust-lang/rfcs#3537) but breaking changes are just as big of a deal so I'm starting this now.

See also #7167, #4309

Compared to `cargo upgrade` / `cargo outdated`, I'm taking a fairly conservative approach and tweaking the existing output as a starting point / MVP.  We can experiment with a richer or easier-to-consume way of expressing this over time.

I view us telling people they aren't on the latest as a warning, so I made that text yellow.

`clap $ cargo update --dry-run` 
image

`clap $ cargo update --dry-run --verbose`
image

This sets up the minimal implementation and slowly adds bits at a time, with a test first that demonstrates it.

### Additional information

I'm expecting that the `cargo upgrade` integration will extend the notes to say something like "X dependencies may be updated with `--breaking`"

### DIFF
--- a/src/cargo/ops/cargo_generate_lockfile.rs
+++ b/src/cargo/ops/cargo_generate_lockfile.rs
@@ -1,8 +1,10 @@
 use crate::core::registry::PackageRegistry;
 use crate::core::resolver::features::{CliFeatures, HasDevUnits};
+use crate::core::Registry as _;
 use crate::core::{PackageId, PackageIdSpec, PackageIdSpecQuery};
 use crate::core::{Resolve, SourceId, Workspace};
 use crate::ops;
+use crate::sources::source::QueryKind;
 use crate::util::cache_lock::CacheLockMode;
 use crate::util::config::Config;
 use crate::util::style;
@@ -162,6 +164,35 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
         opts.config.shell().status_with_color(status, msg, color)
     };
     for (removed, added) in compare_dependency_graphs(&previous_resolve, &resolve) {
+        let latest = if let Some(added) = added
+            .iter()
+            .rev()
+            .next()
+            .filter(|p| p.source_id().is_registry())
+        {
+            let query =
+                crate::core::dependency::Dependency::parse(added.name(), None, added.source_id())?;
+            let possibilities = loop {
+                match registry.query_vec(&query, QueryKind::Exact) {
+                    std::task::Poll::Ready(res) => {
+                        break res?;
+                    }
+                    std::task::Poll::Pending => registry.block_until_ready()?,
+                }
+            };
+            let warn = style::WARN;
+            possibilities
+                .iter()
+                .map(|s| s.as_summary())
+                .map(|s| s.version().clone())
+                .max()
+                .filter(|v| added.version() < v)
+                .map(|v| format!(" {warn}(latest: v{v}){warn:#}"))
+        } else {
+            None
+        }
+        .unwrap_or_default();
+
         if removed.len() == 1 && added.len() == 1 {
             let msg = if removed[0].source_id().is_git() {
                 format!(
@@ -170,7 +201,7 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
                     &added[0].source_id().precise_git_fragment().unwrap()[..8],
                 )
             } else {
-                format!("{} -> v{}", removed[0], added[0].version())
+                format!("{} -> v{}{latest}", removed[0], added[0].version())
             };
 
             // If versions differ only in build metadata, we call it an "update"
@@ -184,10 +215,10 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
             }
         } else {
             for package in removed.iter() {
-                print_change("Removing", format!("{}", package), &style::ERROR)?;
+                print_change("Removing", format!("{package}"), &style::ERROR)?;
             }
             for package in added.iter() {
-                print_change("Adding", format!("{}", package), &style::NOTE)?;
+                print_change("Adding", format!("{package}{latest}"), &style::NOTE)?;
             }
         }
     }
--- a/src/cargo/ops/cargo_generate_lockfile.rs
+++ b/src/cargo/ops/cargo_generate_lockfile.rs
@@ -281,12 +281,12 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
     }
     if opts.config.shell().verbosity() == Verbosity::Verbose {
         opts.config.shell().note(
-            "To see how you depend on a package, run `cargo tree --invert --package <dep>@<ver>`",
+            "to see how you depend on a package, run `cargo tree --invert --package <dep>@<ver>`",
         )?;
     } else {
         if 0 < unchanged_behind {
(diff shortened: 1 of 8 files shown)


epage:
CC @djc since we had conversations about related to this for `cargo upgrade` in killercup/cargo-edit#812

src/cargo/ops/cargo_generate_lockfile.rs
```diff
@@ -161,7 +164,49 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
...
+                    s_version.pre.is_empty()
+                        // Only match pre-release if major.minor.patch are the same
+                        || (s_version.major == present_version.major
+                            && s_version.minor == present_version.minor
+                            && s_version.patch == present_version.patch)
+                })
+                .map(|s| s.version().clone())
+                .max()
+                .filter(|v| present.version() < v)
+                .map(|v| format!(" {warn}(latest: v{v}){warn:#}"))
```
djc, line 204: Style suggestion: personally I find this kind of code easier to follow if you combine closures for `map().filter().map()` and `filter().map()` into one each.

src/cargo/ops/cargo_generate_lockfile.rs
```diff
@@ -161,7 +164,49 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
...
         opts.config.shell().status_with_color(status, msg, color)
     };
-    for (removed, added) in compare_dependency_graphs(&previous_resolve, &resolve) {
+    let mut unchanged_behind = 0;
+    for (removed, added, unchanged) in compare_dependency_graphs(&previous_resolve, &resolve) {
+        let highest_present = [added.iter().rev().next(), unchanged.iter().rev().next()]
+            .into_iter()
+            .flatten()
+            .max_by_key(|s| s.version());
+        let latest = if let Some(present) = highest_present.filter(|p| p.source_id().is_registry())
```
djc, line 173: Maybe use a `filter_map()` for this too, to avoid the dangling `else { None }` branch at the end? Could use early return.

Also given that `latest` is prefiltered to be distinct from the current actual, I find the name not that obvious. Maybe `newer`?

src/cargo/ops/cargo_generate_lockfile.rs
```diff
@@ -184,11 +229,34 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
...
             }
+            unchanged_behind += unchanged.len();
+        }
+    }
+    if opts.config.shell().verbosity() == Verbosity::Verbose {
+        opts.config.shell().note(
+            "To see how you depend on a package, run `cargo tree --invert --package <dep>@<ver>`",
+        )?;
+    } else {
+        if 0 < unchanged_behind {
```
djc, line 256: Suggestion: I think writing `unchanged_behind > 0` would be more idiomatic (and easier to parse for me, at least)?

And even then, it's a little unclear to me what this condition is about.

djc (commented):
I really like this approach! I think showing this data in the `update` output relieves the pressure on having `upgrade` (or `--breaking` or whatever we spell it as).

Suggested some stylistic improvements, feel free to ignore of course.

src/cargo/ops/cargo_generate_lockfile.rs
```diff
@@ -170,7 +215,7 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
                     &added[0].source_id().precise_git_fragment().unwrap()[..8],
                 )
             } else {
-                format!("{} -> v{}", removed[0], added[0].version())
+                format!("{} -> v{}{latest}", removed[0], added[0].version())
```
epage, line 218: Note that we calculate `latest` once, so if you have depend on different versions of a package, we won't show anything if your highest version is the latest.

Thoughts?

src/cargo/ops/cargo_generate_lockfile.rs
epage: For me, its the opposite.  Each one is serving a unique purpose and having them represented separately better shows that and is easier to refactor (e.g. I should turn the `filter` into a `take_while`).  I also find `filter` for booleans easier to read than requiring converting bools to `Option` in a `filter_map`, whether through an explicit `if` or  through `.then`.

I use `filter_map` more when I'm working with `Option`s or the steps are more coupled.

src/cargo/ops/cargo_generate_lockfile.rs
epage: While there is yoda speak (`constant == variable`) and more general speak (`variable == constant`) for equality, I see comparison operators differently and prefer to write them in number line order as I find that easier to read.

src/cargo/ops/cargo_generate_lockfile.rs
epage: In my experience, the right balance of combiantors vs procedural logic is a matter of whether its bookkeeping and size.  In this case, this expression is too large in my opinion to shove in an iterator combinator.

epage:
> I would personally appreciate showing unchanged behind dependencies even without -v but that might be too big a step?

I think it is.  People might not want to play dependency whack-a-mole with transitive dependencies and so overwhelming them with all of the unchanged-but-behind dependencies, especially in a large workspace, would be frustrating especially as it might hide the changes the user cares about.

Or put another way, the primary purpose of the command is updating and the majority of the output should be devoted to that and not with "help".

src/cargo/ops/cargo_generate_lockfile.rs
weihanglo: I am on the same side of djc regarding the order, but not really a blocker.

djc:
> > I would personally appreciate showing unchanged behind dependencies even without -v but that might be too big a step?
> 
> I think it is. People might not want to play dependency whack-a-mole with transitive dependencies and so overwhelming them with all of the unchanged-but-behind dependencies, especially in a large workspace, would be frustrating especially as it might hide the changes the user cares about.

How about only showing them for direct dependencies, that is, dependency edges that originate in the current workspace?

epage:
In a large workspace, there can still be a lot of direct dependencies.

src/cargo/ops/cargo_generate_lockfile.rs
djc: I mean, the unfortunate version of this is if your direct dependencies require an older version whereas there is some indirect edge to the newest version, that is, it's not showing something that would be actionable.

src/cargo/ops/cargo_generate_lockfile.rs
```diff
@@ -218,47 +317,49 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
     fn compare_dependency_graphs(
```
weihanglo, line 317: nit: could we have a comment or return a struct so that people can look at it without read through the entire implementation to understand what it returns?

weihanglo (approved):
Thanks. Only some nits not blockers. `r=weihanglo` whenever you feel ready to go :)

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #13377, opened 2024-01-31
chore(ci): enable m1 runner

weihanglo:

Enable M1 runner in Cargo's CI pipeline.

In the foreseeable future, aarch64-apple-darwin would be a tier-1 platform. Regardless, it is a major platform people use right now.

* [GitHub Actions: Introducing the new M1 macOS runner available to open source!](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/)
* https://github.com/rust-lang/rust/pull/120509

Is there any other concern?

* Should we pay attention to billing issue?
* Reach out to t-infra?
* Too small the machine is ([7GB RAM right now](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-private-repositories))

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -139,6 +139,10 @@ jobs:
           os: macos-latest
           rust: nightly
           other: x86_64-apple-ios
+        - name: macOS aarch64 stable
+          os: macos-14
+          rust: stable
+          other: x86_64-apple-darwin
         - name: Windows x86_64 MSVC stable
           os: windows-latest
           rust: stable-msvc


.github/workflows/main.yml
```diff
@@ -139,6 +139,10 @@ jobs:
           os: macos-latest
           rust: nightly
           other: x86_64-apple-ios
+        - name: macOS aarch64 stable
+          os: macos-14
+          rust: stable
+          other: x86_64-apple-darwin
```
Muscraft, line 145: Shouldn't this be `aarch64-apple-darwin`?

Muscraft: Disregard this comment. 

`other` is for cross-compile tests, and or `arch64-apple-darwin` its `other` should be [`x86_64-apple-darwin`](https://github.com/rust-lang/cargo/blob/54281b804b4014e4238a227d65f5e8e14571912b/crates/cargo-test-support/src/cross_compile.rs#L212-L213) so this is correct.

Muscraft:
I was mistaken in [in my comment here](https://github.com/rust-lang/cargo/pull/13377#discussion_r1473268054), which led to an `r-,` so I will `r+` it again.

@bors r+

weihanglo:
The CI result shows that M1 runner is the fastest among all platforms, around ~12 mins. For comparsion, x86_64 Linux stable took 13 mins and x86_64 darwin took 40 mins.

PR #13382, opened 2024-02-01
Change tests to support changes to suggestion

estebank:
`rustc` will start marking the suggestions for prefacing unused bindings with underscores as "maybe incorrect", which makes them no longer auto applicable by `rustfix`.

Change done at https://github.com/rust-lang/rust/pull/120470.


weihanglo (approved):
Make sense. Thanks!

weihanglo:
Will update rust-lang/rust#120508 when this gets merged.

PR #13387, opened 2024-02-02
Don't print rustdoc command lines on failure by default

### LINKED ISSUE #13386: give a better error when rustdoc exits with a non-zero exit code
### Problem

currently, the error message when rustdoc errors looks like this:
```
; RUSTDOCFLAGS=-Dwarnings cargo doc --no-deps --document-private-items --message-format short -q
src/main.rs:1:11: error: unclosed HTML tag `usize`
error: could not document `example`

Caused by:
  process didn't exit successfully: `/Users/jyn/.local/lib/rustup/toolchains/nightly-aarch64-apple-darwin/bin/rustdoc --edition=2021 --crate-type bin --crate-name example src/main.rs -o /Users/jyn/.local/lib/cargo/target/doc --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat,diagnostic-short --diagnostic-width=186 --document-private-items '-Arustdoc::private-intra-doc-links' -C metadata=a59e1929dfc569b0 -L dependency=/Users/jyn/.local/lib/cargo/target/debug/deps --extern bitflags=/Users/jyn/.local/lib/cargo/target/debug/deps/libbitflags-d34f162dfbd7ed32.rmeta --extern inner=/Users/jyn/.local/lib/cargo/target/debug/deps/libinner-a5a332b9f54b66cc.rmeta -Dwarnings --crate-version 0.1.0` (exit status: 1)
```

this is already kind of long, and it gets much worse as you add more dependencies.

### Steps

1. add some docs that will warn, e.g. `//! Option<usize>` at the root of the crate
2. `RUSTDOCFLAGS=-Dwarnings cargo doc`

### Possible Solution(s)

it would be nice to make it consistent with the error for rustc:
```
; cargo c -q --message-format short
src/main.rs:10:12: error[E0038]: the trait `Bar` cannot be made into an object
src/main.rs:10:32: error[E0038]: the trait `Bar` cannot be made into an object
error: could not compile `example` (bin "example") due to 2 previous errors
```

### Notes

_No response_

### Version

```text
cargo 1.77.0-nightly (ac6bbb332 2023-12-26)
release: 1.77.0-nightly
commit-hash: ac6bbb33293d8d424c17ecdb42af3aac25fb7295
commit-date: 2023-12-26
host: aarch64-apple-darwin
libcurl: 8.4.0 (sys:0.4.70+curl-8.5.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 14.3.0 [64-bit]
```

alexcrichton:
This commit lifts a helper function from invoking rustc to additionally being used for invoking rustdoc. This enables hiding the command line invocation of `rustdoc` by default when it fails, although it's still available to see with `--verbose`. The intention here is to match the behavior of `cargo build` for rustc failures here and was inspired by recently running `cargo doc` and not being able to see the actual failure as the command line ended up taking the whole screen (afterwards I made the screen bigger and that helped too).

Fixes #13386

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -376,19 +376,6 @@ fn rustc(cx: &mut Context<'_, '_>, unit: &Unit, exec: &Arc<dyn Executor>) -> Car
             }
         }
 
-        fn verbose_if_simple_exit_code(err: Error) -> Error {
-            // If a signal on unix (`code == None`) or an abnormal termination
-            // on Windows (codes like `0xC0000409`), don't hide the error details.
-            match err
-                .downcast_ref::<ProcessError>()
-                .as_ref()
-                .and_then(|perr| perr.code)
-            {
-                Some(n) if cargo_util::is_simple_exit_code(n) => VerboseError::new(err).into(),
-                _ => err,
-            }
-        }
-
         state.running(&rustc);
         let timestamp = paths::set_invocation_time(&fingerprint_dir)?;
         if build_plan {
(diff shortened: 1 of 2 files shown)


epage:
Double checking but this fixes #13386, right?

epage (approved):

alexcrichton:
Oh wow it definitely does and this was not coordinated at all so turns out we just both ran into it at the same time!

Issue #13345, opened 2024-01-25
cargo-new: `workspace` is added as an inline table


weihanglo:
### Problem

In <https://github.com/rust-lang/cargo/pull/12779>, cargo adds new packages to [workspace.members] automatically. This is great when `workspace.member` array alreay exists, but the format doesn't look good when `workspace` array is missing.

### Steps

1. `cargo new foo` && cd foo
2. `cargo add bar --lib`

And you'll get the resulting TOML

```toml
workspace = { members = ["bar"] }
[package]
name = "foo"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
```

### Possible Solution(s)

Don't add the new package to `workspace.members` if there is no existing `workspace` in Cargo.toml.

### Version

```text
cargo 1.77.0-nightly (1ae631085 2024-01-17)
release: 1.77.0-nightly
commit-hash: 1ae631085f01c1a72d05df1ec81f3759a8360042
commit-date: 2024-01-17
```

epage:
... is the error instead that we shouldn't be doing this if `foo` doesn't have a `[workspace]` at all?

> If the existing toml contains an empty inline workspace table like workspace = {}, should we reformat it to [workspace]?

So far, we've taken the approach of preserving what the user already has rather than cleaning it up.  That is rustfmt's job.

weihanglo:
> ... is the error instead that we shouldn't be doing this if `foo` doesn't have a `[workspace]` at all?

This sounds better 👍🏾.

linyihai:
@rustbot claim

I'll try.

PR #13391, opened 2024-02-04
Don't add the new package to workspace.members if there is no existing workspace in Cargo.toml.

linyihai:
Fixed https://github.com/rust-lang/cargo/issues/13345

If the current package has no workspace table in Cargo.toml,  then if you run `cargo add  foo`, don't create the workspace inline item and don't add `foo` into the workspace.members.

Reviewed by commit by commit.


tests/testsuite/cargo_new/add_lib_into_non_workspace/in/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
```
weihanglo, line 1: What about another case: a package with an empty `[workspace]` section?

weihanglo (commented):
Thanks for working on this!

tests/testsuite/cargo_new/add_lib_into_non_workspace/in/Cargo.toml
linyihai: See `add_members_to_workspace_without_members`

src/cargo/ops/cargo_new.rs
```diff
@@ -970,33 +970,34 @@ fn update_manifest_with_new_member(
...
-        }
+        } else {
+            let mut array = Array::new();
+            array.push(display_path);
 
-        let was_sorted = is_sorted(members.iter().map(Value::as_str));
-        members.push(display_path);
-        if was_sorted {
-            members.sort_by(|lhs, rhs| lhs.as_str().cmp(&rhs.as_str()));
+            workspace_document["workspace"]["members"] = toml_edit::value(array);
```
weihanglo, line 999: nit: one less access

```suggestion
            workspace["members"] = toml_edit::value(array);
```

weihanglo (approved):
Looks pretty good!

src/cargo/ops/cargo_new.rs
linyihai: That's a good point.

weihanglo (approved):

weihanglo:
Thanks for the patch!

@bors r+

PR #13393, opened 2024-02-04
[beta-1.77.0] Fix panic on empty spans when parsing Cargo.toml

weihanglo:
Beta backports:

- <https://github.com/rust-lang/cargo/pull/13375>
- <https://github.com/rust-lang/cargo/pull/13376>

In order to make CI pass, the following PRs are also cherry-picked:

- <https://github.com/rust-lang/cargo/pull/13362>

PR #13396, opened 2024-02-05
docs(changelog): Slight cleanup

epage:

For strip, I felt the nuance of where this change applies isn't as obvious and might give people the wrong impression, so I tried to make that front and center.

For the panic, the fix was backported to 1.77 (#13393) which is when it got introduced (#13172).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,9 +21,6 @@
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


weihanglo (approved):

weihanglo:
Nice! Thank you.

@bors r+

PR #13399, opened 2024-02-05
feat(cli): Allow logging to chrome traces

epage:

> The time from executing cargo to executing rustc is 230 ms. I wonder if there’s scope for caching whatever expensive computations cargo is doing here.

*Source: https://davidlattimore.github.io/working-on-rust-iteration-time.html*

This made me curious where the time was going.  I've been meaning to try out `tracing-chrome` for a while and this gave me the opportunity.

This adds `CARGO_LOG_PROFILE=<bool>` and `CARGO_LOG_PROFILE_CAPTURE_ARGS=<bool>` for enabling and controlling these trace files.  These are perma-unstable env variables.

Traces can be viewed at https://ui.perfetto.dev or `chrome://tracing`.

Example: no-op `cargo check` run on `cargo-nextest`:

image

TODO
- [ ] Find a place to document this, see https://doc.crates.io/contrib/tests/profiling.html#internal-profiler

I looked in `config/` and didn't see a env variable parser to reuse (`get_cv_with_env` is too specialized).

### Additional information

In past projects, I've been able to use features like this to better understand code bases, bugs, etc.  Hopefully this can evolve to help on-board people one day.

Most of the implementation was inspired by https://github.com/arxanas/git-branchless/blob/2923924dfbff07faa61d50771b6604cb18e64bc8/git-branchless-invoke/src/lib.rs#L55

`tracing-chrome` hasn't received updates recently, but it is also fairly quiet on Issues and PRs, so this might just be "maturity".  It does have over a million downloads.  It also is only enabled if explicitly opted into.

rustup added trace visualization in rust-lang/rustup#3287 using [open telemetry](https://crates.io/crates/tracing-opentelemetry) but it is behind a feature flag and and requires running a server to receive the traces.


epage:
> What's the difference between [tracing-chrome](https://crates.io/crates/tracing-chrome) and [samply](https://github.com/mstange/samply/), and maybe other old school dtrace flamgraph approaches?

This builds on `tracing`
- Limited to what you trace
- But doesn't require any other special binary setup
- But can show argument values and other information
- The new instrument calls would hopefully improve the overall quality of our tracing

Overall, I'd say this is a lot more approachable solution for users asking the question "where is my build time going?", like those using `--timings` or `Ztime-passes`

arxanas:
Incidentally, `cargo` was a hotspot when I measured it a while back, as per https://lobste.rs/s/ktyp2q/improving_incremental_test_times_rust#c_ukhulp

src/bin/cargo/main.rs
```diff
@@ -35,16 +35,43 @@ fn main() {
...
+    let fmt_layer = tracing_subscriber::fmt::layer()
         .with_timer(tracing_subscriber::fmt::time::Uptime::default())
         .with_ansi(std::io::IsTerminal::is_terminal(&std::io::stderr()))
         .with_writer(std::io::stderr)
-        .with_env_filter(env)
-        .init();
+        .with_filter(env);
+
+    let (profile_layer, profile_guard) =
+        if env_to_bool(std::env::var_os("_CARGO_LOG_PROFILE").as_deref()) {
```
weihanglo, line 51: While not enforced, there is a convention in this repo of prefixing with double underscores `__CARGO`?

src/bin/cargo/main.rs
```diff
@@ -35,16 +35,43 @@ fn main() {
...
+            let capture_args =
+                env_to_bool(std::env::var_os("_CARGO_LOG_PROFILE_CAPTURE_ARGS").as_deref());
```
weihanglo, line 53: Idea: use the same env var and accept different values.

src/bin/cargo/main.rs
```diff
@@ -18,7 +18,7 @@ mod commands;
 use crate::command_prelude::*;
 
 fn main() {
-    setup_logger();
+    let _guard = setup_logger();
```
weihanglo, line 21: This makes me wonder, should we consider removing and replacing `util::profile` with this? I know @ehuss is using it in `-Zgc` though I never learnt how to use.

There is also an existing [`CARGO_PROFILE`](https://doc.crates.io/contrib/tests/profiling.html#internal-profiler) environment variable documented. We should expand the doc in this PR or a follow-up.

epage: @ehuss what are your thoughts on the existing `CARGO_PROFILE` / `util::profile` vs leveraging `tracing` to do profiling?

For myself, having a non-bespoke system to maintain / use and leveraging the same instrumentation for both logging and profiling, seems like a win.

src/bin/cargo/main.rs
epage: I was modeling this off of `git-branchless` where their equivalent of `_CARGO_LOG_PROFILE` accepted file names. I just held off on that for now.  Being able to set the file name is useful for predictability and for when nesting processes in each other.

We also have the possibility of doing logger-like stuff of filtering what gets captured.  This is remind me that profiling is just one role for this and this is also a programmatic format for logged output that allows graphical browsing of the relationship of events.

Unsure if that changes your thoughts here on merging them.

weihanglo: Given we claim it as perma-unstable, everything sounds good to me 👍🏾.

src/bin/cargo/main.rs
```diff
@@ -277,6 +304,7 @@ fn search_directories(config: &Config) -> Vec<PathBuf> {
 }
 
 /// Initialize libgit2.
+#[tracing::instrument(skip_all)]
 fn init_git(config: &Config) {
```
epage, line 308: Aside: @Byron do you know if there should be or will be an equivalent to this in gitoxide?  In the PR description is a trace that shows that this operation is 20% of the no-op `cargo check` time.  Wondering if this is something that is worth hacking around or if its a "this will go away with git2".

Byron: This is an interesting case as its unclear to me how much 20% actually represent in absolute terms of time.
With that said, I looked at what `init_git` is doing and I was surprised it takes time at all - it *felt* like it's loading some code dynamically (which is probably wrong).

`gitoxide` isn't comparable with `git2` as there are no globals, and everything is lazy, so I'd guess that the timing is going to be more favourable. For one, there is no equivalent to `init_git` - configuration is done only when opening the repository.

I hope that helps.

src/bin/cargo/main.rs
epage: I talked with ehuss and he was fine with replacing `CARGO_PROFILE`.  The latest push does this.

src/doc/contrib/src/tests/profiling.md
```diff
@@ -2,15 +2,24 @@
...
-profile stack to print results for.
+Cargo leverages [tracing](https://crates.io/crates/tracing)
+as a basic, hierarchical built-in profiler.
 
-```sh
-# Output first three levels of profiling info
-CARGO_PROFILE=3 cargo generate-lockfile
+Environment variables:
+- `CARGO_LOG_PROFILE=<true|1>`: log tracing events to a file in the current working directory
+- `CARGO_LOG_PROFILE_CAPTURE_ARGS=<true|1>`: include arguments in the events
```
epage, line 10: Inspired by us previously having `CARGO_PROFILE` but not really expecting compatibility on it, I've removed the `__` prefixes but added a note below.

I kept the name as `CARGO_LOG_PROFILE` rather than switching to `CARGO_PROFILE` to associate this with `CARGO_LOG` in people's minds as that is the source for our profiling information.  Really, this is for `CARGO_LOG` visualization, and not just profiling, but I feel like current name is "good enough" rather than finding a "perfect" name that captures the uses for this.

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -405,11 +405,6 @@ pub fn prepare_target(
     unit: &Unit,
     force: bool,
 ) -> CargoResult<Job> {
-    let _p = profile::start(format!(
```
weihanglo, line 408: The old profiler records `PackageId` and target name, while the new profiler misses that information as it skips the entire `unit` arg. Should we re-include that?

src/cargo/ops/resolve.rs
```diff
@@ -147,8 +147,6 @@ pub fn resolve_ws_with_opts<'gctx>(
         // Second, resolve with precisely what we're doing. Filter out
         // transitive dependencies if necessary, specify features, handle
         // overrides, etc.
-        let _p = profile::start("resolving with overrides...");
```
weihanglo, line 150: This is dropped and no profiling for the code path. Just pointing out but not too worried about the change.

src/cargo/core/global_cache_tracker.rs
```diff
@@ -863,9 +857,6 @@ impl GlobalCacheTracker {
         child_base_paths: &[&Path],
         delete_paths: &mut Vec<PathBuf>,
     ) -> CargoResult<()> {
-        let _p = crate::util::profile::start(format!(
```
weihanglo, line 866: `parent_table_name` is missing in instrument, and ditto in global_cache_tracker.rs

src/doc/contrib/src/tests/profiling.md
weihanglo: Sounds good. If we have `[log]` table in config in the future, we can just rename this.

src/cargo/ops/resolve.rs
epage: `add_overrides` has an `instrument` attribute.   The change in coverage is for the warning which I assume wasn't going to be too important.

src/cargo/core/compiler/fingerprint/mod.rs
epage: See #13538

src/cargo/core/global_cache_tracker.rs
epage: See #13538

weihanglo (approved):

weihanglo:
@bors r+

Thanks and really looking forward to this!

PR #13400, opened 2024-02-05
Fix markdown line break in cargo-add

ehuss:
This fixes a backslash line break in the cargo-add documentation. A future version of mdbook will adhere to the commonmark spec more closely, and this will end up showing as a literal backslash.


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #13401, opened 2024-02-05
Remove build metadata from curl-sys version.

ehuss:
#13379 added the build metadata to the curl-sys version, but cargo spits a warning that you shouldn't do that. It also didn't add the metadata correctly, as the version is `+curl-8.6.0`. This removes that, since it isn't needed.


epage:
@bors r+

Reported this at https://github.com/renovatebot/renovate/discussions/27086

weihanglo (approved):
Thanks!

We should find a way telling RenovateBot to stop adding build-metadata.

epage:
btw looks like crates.io also has this issue, created rust-lang/crates.io#8073

PR #13404, opened 2024-02-06
test: Verify build-std final crate graph against library lock file. 

c272:

Adds a test to the build-std test suite which checks whether the resolved standard library unit graph accurately describes a subset of the dependencies listed in the standard library's `Cargo.lock`, as mentioned in an issue from wg-cargo-std-aware:

> It might be a good idea to do the equivalent of --locked for the standard library to ensure that the Cargo.lock isn't modified, and that it is correct.
Source: https://github.com/rust-lang/wg-cargo-std-aware/issues/38

This PR implements this as part of the build-std test suite, as suggested by @adamgemmell in the above linked issue, however I also have a version of this patch implemented which can perform this check at runtime, if that would be a more appropriate place to perform the check.

Fixes https://github.com/rust-lang/wg-cargo-std-aware/issues/38

The test can be run as follows:
```
CARGO_RUN_BUILD_STD_TESTS=1 cargo test validate_std_crate_graph
```

### Additional information
N/A

### DIFF
--- a/src/cargo/core/resolver/encode.rs
+++ b/src/cargo/core/resolver/encode.rs
@@ -406,6 +406,10 @@ impl EncodableResolve {
             HashMap::new(),
         ))
     }
+
+    pub fn package(&self) -> Option<&Vec<EncodableDependency>> {
+        self.package.as_ref()
+    }
 }
 
 fn build_path_deps(ws: &Workspace<'_>) -> CargoResult<HashMap<String, SourceId>> {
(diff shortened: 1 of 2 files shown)


adamgemmell:
Bump - maybe @ehuss is interested in this patch?

ehuss:
Sorry for the long delay.

> which can perform this check at runtime, if that would be a more appropriate place to perform the check.

Yea, I think it would probably be the preferred route to go to verify it is a subset at runtime. Would you be able to update the PR to include that?

weihanglo:
Going to close this. See <https://github.com/rust-lang/wg-cargo-std-aware/issues/38#issuecomment-2271758701>.

We are still looking for closing out <https://github.com/rust-lang/wg-cargo-std-aware/issues/38>.

Issue #13407, opened 2024-02-06
Issue with the Code of Conduct in Readme.md redirecting to a `422: Unprocessable Entity`


gSayak:
### Problem

Right now the **code-of-conduct** link in the `Readme.md` in the `master` branch redirects to:
 `https://www.rust-lang.org/conduct.html` 

which results in an:
``` 
**422: Unprocessable Entity
The request was well-formed but was unable to be followed due to semantic errors.**
```
It should have redirected to: `https://www.rust-lang.org/policies/code-of-conduct`

Fixed this issue in the Readme.md by making a new branch and will be linking the PR to this issue

### Steps

1. Go to `https://github.com/rust-lang/cargo`
2. Check the readme.md and click on Code of Conduct
3. Click on the hyperlink of `can be found online`
4. This would redirect to a **422: Unprocessable Entity**

### Possible Solution(s)

Can fix the issue by changing the link to [Code-of-Conduct](https://www.rust-lang.org/policies/code-of-conduct)

gSayak:
Solved the issue and sent the PR #13408

weihanglo:
This has been fixed in <https://github.com/rust-lang/www.rust-lang.org/pull/1924>.

Redirections in staging website work now: <https://www-staging.rust-lang.org/conduct.html>. When the next deployment of the production website is done, this should be resolved. Thank you for the report!

PR #13408, opened 2024-02-06
fixed the issue with the link redirecting to a 422: Unprocessable Entity

gSayak:

This PR solves issue #13407 

There is an issue with the readme.md where if you check the Code-of-conduct tab in the readme and click on the `can be found online` link, it redirects to a **422: Unprocessable Entity**.  Fixed this by changing the link to the actual code-of-conduct link.

To test this PR, go to the readme.md file in the PR and click on the link of `can be found online`, it would redirect to the correct webpage.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,3 +1,3 @@
 # The Rust Code of Conduct
 
-The Code of Conduct for this repository [can be found online](https://www.rust-lang.org/conduct.html).
+The Code of Conduct for this repository [can be found online](https://www.rust-lang.org/policies/code-of-conduct).


weihanglo (commented):
Thanks for the fix!

While this does resolve the issue, I am not sure since when the link is broken. We might want to reach out to <https://github.com/rust-lang/www.rust-lang.org/> to have a redirection.

gSayak:
I see, I thought the Code of Conduct link was moved to a different link and it was not updated in the readme. 
Will we need to fix it in the www.rust-lang.org repo?

weihanglo:
The old link was used in pre 2018 website: <https://prev.rust-lang.org/conduct.html>. And here are [redirection rules](https://github.com/rust-lang/www.rust-lang.org/blob/0b933a84955ad879f79d5a489bbf54b41244bad9/src/redirect.rs#L10). I am not sure why it didn't work. Maybe something wrong when upgrading Rocket to v0.5? We can clone the repo and debug there.

Since this link is used in a couple of different repository (rust-lang/rust, rust-lang/rust-clippy, …), I lean towards fixing in the redirection. If we cannot find the root cause, then  contact Manishearth as they are [the only member on the website team](https://github.com/rust-lang/team/blob/master/teams/website.toml).

gSayak:
Okay then, I will try to fix this redirection issue and give an update here

weihanglo:
Close as fixed in <https://github.com/rust-lang/www.rust-lang.org/pull/1924>. Regardless, thanks for your help :)

PR #13409, opened 2024-02-06
chore: Rename `Config` to `GlobalContext`

Muscraft:
This PR :
- renames `Config` to `GlobalContext`
  - it also renames its references to `gctx` (including lifetimes)
  - This was done to reflect better what it really is at this point
- renames (and moves) `core::compiler::context::Context` to `core::compiler::build_runner::BuildRunner` 
  - [Suggested here](https://github.com/rust-lang/cargo/pull/13409#issuecomment-1930958847)

I believe I got all references to `Config` removed, Everything is on `GlobalContext`, but I could've missed renaming a variable or lifetime somewhere. I tried to find all references to `config: &GlobalContext` and rename them and I think I did so successfully. I also renamed all `'cfg`  to `'gctx`.

Note: I explicitly did not rename any files or paths as I was unsure about the best way to do this.

### How to Review this PR
Take your time (and many breaks)! 

I am sorry.

I am also sorry for the very brief description, looking at words and thinking about them has become hard... I need to not look at words for a while...

<hr>

As a complete side note, I honestly don't know if `config`, `Config`, `ctx`, or `Context` are really words at this point.


weihanglo:
Bikeshedding: GlobalContext or Cargo Context which might be better and good for text search?

ehuss:
Another bit of bikeshedding, I'm a little concerned that `BuildContext` and `CompileContext` might be confusing since in my mind those two words are essentially synonyms of one another. I'm not sure if anyone has other ideas? Maybe `Context` could be `BuildRunner`, to emphasize that it handles the actual process of running the build?

Muscraft:
> Bikeshedding: GlobalContext or Cargo Context which might be better and good for text search?

I am fine with `GlobalContext` and making the variable `gctx` if everyone else is.

> Another bit of bikeshedding, I'm a little concerned that `BuildContext` and `CompileContext` might be confusing since in my mind those two words are essentially synonyms of one another. I'm not sure if anyone has other ideas? Maybe `Context` could be `BuildRunner`, to emphasize that it handles the actual process of running the build?

I honestly don't know enough about `BuildContext` or `CompileContext` to say what they should be named, but I did think it was odd how similar their names were. I am happy to name things whatever.

weihanglo:
It might be a bit odd if `cargo tree` requires `BuildRunner` but doesn't really build stuff. I am not sure though.

ehuss:
> It might be a bit odd if `cargo tree` requires `BuildRunner` but doesn't really build stuff. I am not sure though.

Maybe I'm a little confused, I don't think `cargo tree` uses either of the contexts?

Or maybe my previous message wasn't clear. I was referring to the old `Context`, not the old `Config`.

weihanglo:
Thanks for the clarification, Eric. Then `BuildRunner` makes sense to me!

src/cargo/util/toml/mod.rs
```diff
@@ -672,19 +671,19 @@ pub fn to_real_manifest(
 
     let mut deps = Vec::new();
 
-    let mut cx = Context {
+    let mut manifest_gctx = ManifestContext {
```
epage, line 674: Why does this have a `g` in the name?

tests/testsuite/config.rs
```diff
@@ -115,17 +117,17 @@ impl ConfigBuilder {
...
             &self.config_args,
         )?;
-        Ok(config)
+        Ok(gctx)
     }
 }
 
-fn new_config() -> Config {
-    ConfigBuilder::new().build()
+fn gctx() -> GlobalContext {
```
epage, line 124: I think I like `new` being in the name as it conveys a new instance is being created, rather than accessing an existing instance (as most people will be looking at callers and might not notice this moves the value)

src/cargo/core/features.rs
```diff
@@ -79,7 +79,7 @@
 //!    [`unstable_cli_options!`]. Flags can take an optional value if you want.
 //! 2. Update the [`CliUnstable::add`] function to parse the flag.
 //! 3. Wherever the new functionality is implemented, call
-//!    [`Config::cli_unstable`] to get an instance of [`CliUnstable`]
+//!    [`Context::cli_unstable`] to get an instance of [`CliUnstable`]
```
weihanglo, line 82:
```suggestion
//!    [`GlobalContext::cli_unstable`] to get an instance of [`CliUnstable`]
```

src/cargo/core/features.rs
```diff
@@ -112,7 +112,7 @@
 //!    and summarize it similar to the other entries. Update the rest of the
 //!    documentation to add the new feature.
 //!
-//! [`Config::cli_unstable`]: crate::util::config::Config::cli_unstable
+//! [`Context::cli_unstable`]: crate::util::config::GlobalContext::cli_unstable
```
weihanglo, line 115:
```suggestion
//! [`GlobalContext::cli_unstable`]: crate::util::config::GlobalContext::cli_unstable
```

src/cargo/core/resolver/mod.rs
```diff
@@ -125,39 +125,40 @@ pub fn resolve(
...
         }
         _ => None,
     };
     let mut registry = RegistryQueryer::new(registry, replacements, version_prefs);
-    let cx = loop {
-        let cx = Context::new();
-        let cx = activate_deps_loop(cx, &mut registry, summaries, first_version, config)?;
+    let resolver_gctx = loop {
+        let resolver_gctx = ResolverContext::new();
+        let resolver_gctx =
```
weihanglo, line 140: and ditto

```suggestion
    let resolver_ctx = loop {
        let resolver_ctx = ResolverContext::new();
        let resolver_ctx =
```

weihanglo (commented):
Finished the review! Thanks for the hard work 👍🏾

There might be some other places need updates, but let's get this merged asap to avoid conflicts.

epage:
If there is anything else, we can patch it up over time

@bors r+

PR #13410, opened 2024-02-07
fix: Switch more notes/warnings to lowercase

epage:
See https://doc.crates.io/contrib/implementation/console.html#style

By fixing existing cases, we make it more likely people will copy a case they should.

I left out multi-sentence cases because I was unsure how to handle those


src/cargo/ops/registry/publish.rs
```diff
@@ -228,7 +228,7 @@ fn wait_for_publish(
         format!("{short_pkg_description} to {source_description}"),
     )?;
     config.shell().note(format!(
-        "Waiting for `{short_pkg_description}` to be available at {source_description}.\n\
+        "waiting for `{short_pkg_description}` to be available at {source_description}.\n\
```
weihanglo, line 231: can use multiple `note` invocation for multi-lines message maybe?

weihanglo (approved):

Issue #13403, opened 2024-02-06
Print a note when `cargo new` is editing `workspace.members`


weihanglo:
### Problem

Starting from #12779, `cargo new` automatically adds the new package to `workspace.members` when invoked inside from a directory of a workspace. This is a great feature, however sometimes a bit less noticeable.

### Proposed Solution

Print a notice when a workspace is found and cargo new are editing `workspace.members` of it. For example,

```diff
      Creating binary (application) `foo` package
+ note: adding `foo` as a workspace member of workspace `bar` at `/home/projects/bar`
  note: see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```

### Notes

version

```
cargo 1.77.0-beta.1 (7bb7b5395 2024-01-20)
release: 1.77.0-beta.1
commit-hash: 7bb7b539558dc88bea44cee4168b6269bf8177b0
commit-date: 2024-01-20
```

weihanglo:
cc @linyihai 

This may have conflicted with #13391, or you can take this after #13391 merges :)

linyihai:
@rustbot claim

Thank you for ping me.  I'd like to addressed this after https://github.com/rust-lang/cargo/pull/13391

epage:
I wonder if this could work as a status

```
Adding foo as member of workspace /path
```

PR #13411, opened 2024-02-07
feat: Add hint for adding members to workspace

linyihai:

Fixes https://github.com/rust-lang/cargo/issues/13403

Reviewd commit by commit

### Additional information

### DIFF
--- a/src/cargo/ops/cargo_new.rs
+++ b/src/cargo/ops/cargo_new.rs
@@ -844,11 +844,24 @@ fn mk(config: &Config, opts: &MkOptions<'_>) -> CargoResult<()> {
                 }
 
                 // Try to add the new package to the workspace members.
-                update_manifest_with_new_member(
+                if update_manifest_with_new_member(
                     &root_manifest_path,
                     &mut workspace_document,
                     &display_path,
-                )?;
+                )? {
+                    config.shell().status(
+                        "Adding",
+                        format!(
+                            "`{}` as member of workspace at `{}`",
+                            PathBuf::from(&display_path)
+                                .file_name()
+                                .unwrap()
+                                .to_str()
+                                .unwrap(),
+                            root_manifest_path.parent().unwrap().display()
+                        ),
+                    )?
+                }
             }
         }
     }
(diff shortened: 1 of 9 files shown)


src/cargo/ops/cargo_new.rs
```diff
@@ -992,11 +994,25 @@ fn update_manifest_with_new_member(
...
+            )?
         } else {
             let mut array = Array::new();
             array.push(display_path);
 
             workspace["members"] = toml_edit::value(array);
+            config.shell().status(
+                "Adding",
+                format!(
+                    "`{display_path}` as member of workspace at `{}`",
```
epage, line 1012: This is printing the literal entry being added to the workspace.  I don't think this will be as meaningful (and may even be confusing as they are more likely to see it as a relative path from where they are at) and would suggest the package name.

src/cargo/ops/cargo_new.rs
```diff
@@ -992,11 +994,25 @@ fn update_manifest_with_new_member(
             if was_sorted {
                 members.sort_by(|lhs, rhs| lhs.as_str().cmp(&rhs.as_str()));
             }
+            config.shell().status(
```
epage, line 997: Instead of two implementations (and putting it in this function),could we instead have this function return `bool` and put it in the caller?

src/cargo/ops/cargo_new.rs
```diff
@@ -844,11 +844,24 @@ fn mk(config: &Config, opts: &MkOptions<'_>) -> CargoResult<()> {
...
+                        format!(
+                            "`{}` as member of workspace at `{}`",
+                            PathBuf::from(&display_path)
+                                .file_name()
+                                .unwrap()
+                                .to_str()
+                                .unwrap(),
+                            root_manifest_path.parent().unwrap().display()
+                        ),
+                    )?
```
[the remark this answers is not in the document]
epage: I don't think so.  They expressed an intent already with a glob (or a speculative edit) so I don't think we should print anymore.

epage:
Thanks!

@bors r+

PR #13412, opened 2024-02-07
Update git2

ehuss:
This updates git2 primarily to pull in the update for libgit2 1.7.2 which fixes three security issues. @weihanglo did some investigation, and it looks like cargo may be susceptible to one of them with rev parsing. I am uncertain of the severity, but the CVE seems to imply that it is mainly a denial-of-service with an infinite loop from a well-crafted spec.

See https://github.com/libgit2/libgit2/releases/tag/v1.7.2 for more information.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1037,9 +1037,9 @@ dependencies = [
 
 [[package]]
 name = "git2"
-version = "0.18.1"
+version = "0.18.2"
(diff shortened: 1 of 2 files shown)


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #13417, opened 2024-02-07
[Beta-1.77] Update libgit2

ehuss:
This is a beta packport of #13412 to update libgit2 to fix the security issues mentioned in https://github.com/libgit2/libgit2/releases/tag/v1.7.2. From what I can tell, the threat to cargo is very small. The best I can come up with is a carefully crafted `rev` field in a `Cargo.toml` will cause cargo to hang. However, I would feel safer having this backported.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1037,9 +1037,9 @@ dependencies = [
 
 [[package]]
 name = "git2"
-version = "0.18.1"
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
Thanks. I'll include this in beta backport in rust-lang/rust later.

PR #13418, opened 2024-02-07
Publish test crates

### LINKED ISSUE #10147: release cargo test helper crate to crates-io
The cargo test helper crate would be a very much appreciated relief for all `cargo-$tool` plugins, mostly speaking for `cargo-unleash` and `cargo-spellcheck` .

(This didn't seem to match any category, hence the free form issue).

PaulDance:

This is an attempt to close #10147. In short, publish the `cargo-test-(macro|support)` crates to crates.io.

Main parts:
* Updating the concerned manifests to make the packages publishable.
* Adding them to `publish.py`: not sure if you want this to be done. I put them after all other packages and in reverse-dependency order to ensure things will work the best, hopefully.
* Updating the contribution documentation to reflect the change. I didn't see any other place where the crates' publication was mentioned.

Sadly, the effective testing could not be done by me as it would require publishing the crates, which would block future publications by Cargo team members and therefore render the whole point of this impossible or cumbersome. Hopefully, the PR is small and readable enough that mistakes can be seen easily. The real testing would need to be done manually by some team member either before by checking the source branch out or after merging.

The only testing I was able to perform was to run `cargo publish --dry-run -p cargo-test-(macro|support)`. The first worked just fine. The second failed on the `crates-io` package not being available on crates.io in version `0.40.0`. However, the package is in this version in the current sources. I therefore suspect that a run of `publish.py` should work since the packages have been added there after `crates-io`.

### Additional information

This is very much tentative and open to discussion. I tried my best to update things as I understood them. Please correct me on anything I would have done wrong or simply not thought of.

### DIFF
--- a/crates/cargo-test-macro/Cargo.toml
+++ b/crates/cargo-test-macro/Cargo.toml
@@ -2,12 +2,11 @@
 name = "cargo-test-macro"
 version = "0.1.0"
 edition.workspace = true
+rust-version.workspace = true
 license.workspace = true
 homepage.workspace = true
 repository.workspace = true
-documentation = "https://github.com/rust-lang/cargo"
 description = "Helper proc-macro for Cargo's testsuite."
-publish = false
 
 [lib]
 proc-macro = true
--- a/crates/cargo-test-macro/Cargo.toml
+++ b/crates/cargo-test-macro/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-test-macro"
-version = "0.1.0"
+version = "0.2.0"
 edition.workspace = true
 rust-version.workspace = true
 license.workspace = true
--- a/crates/cargo-test-macro/Cargo.toml
+++ b/crates/cargo-test-macro/Cargo.toml
@@ -2,7 +2,7 @@
 name = "cargo-test-macro"
 version = "0.2.0"
 edition.workspace = true
-rust-version.workspace = true
+rust-version = "1.77"  # MSRV:1
 license.workspace = true
 homepage.workspace = true
 repository.workspace = true
--- a/crates/cargo-test-support/Cargo.toml
+++ b/crates/cargo-test-support/Cargo.toml
@@ -2,7 +2,7 @@
 name = "cargo-test-support"
 version = "0.2.0"
 edition.workspace = true
-rust-version.workspace = true
+rust-version = "1.77"  # MSRV:1
 license.workspace = true
 homepage.workspace = true
 repository.workspace = true
--- a/crates/cargo-test-support/Cargo.toml
+++ b/crates/cargo-test-support/Cargo.toml
@@ -1,9 +1,12 @@
 [package]
 name = "cargo-test-support"
 version = "0.1.0"
-license.workspace = true
 edition.workspace = true
-publish = false
+rust-version.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+description = "Testing framework for Cargo's testsuite."
 
 [lib]
 doctest = false
--- a/crates/cargo-test-support/Cargo.toml
+++ b/crates/cargo-test-support/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-test-support"
-version = "0.1.0"
+version = "0.2.0"
 edition.workspace = true
 rust-version.workspace = true
 license.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -408,11 +408,11 @@ dependencies = [
 
 [[package]]
 name = "cargo-test-macro"
-version = "0.1.0"
+version = "0.2.0"
 
 [[package]]
 name = "cargo-test-support"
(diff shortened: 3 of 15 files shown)


PaulDance:
Added a fix to make `ci/validate-version-bump.sh` work: bump the crate versions.

However, the test still fails on: `Error: cargo-test-support not found in registry (crates.io)`. This is correct, but temporary. It should work as soon as the crates have been published at least once. How should we proceed about this? Have someone manually publish them?

crates/cargo-test-macro/Cargo.toml
```diff
@@ -1,13 +1,20 @@
...
 edition.workspace = true
+rust-version = "1.76.0"  # MSRV:1
 license.workspace = true
 homepage.workspace = true
 repository.workspace = true
-documentation = "https://github.com/rust-lang/cargo"
+documentation = "https://docs.rs/cargo-test-macro"
 description = "Helper proc-macro for Cargo's testsuite."
-publish = false
+include = [
```
weihanglo, line 11: Guess we can just publish everything?

crates/cargo-test-support/Cargo.toml
```diff
@@ -1,9 +1,22 @@
...
+version = "0.1.1"
 edition.workspace = true
-publish = false
+rust-version = "1.76.0"  # MSRV:1
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+documentation = "https://docs.rs/cargo-test-support"
+description = "Testing framework for Cargo's testsuite."
+include = [
```
weihanglo, line 11: ditto

crates/cargo-test-support/Cargo.toml
```diff
@@ -1,9 +1,22 @@
...
-version = "0.1.0"
-license.workspace = true
+version = "0.1.1"
 edition.workspace = true
-publish = false
+rust-version = "1.76.0"  # MSRV:1
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+documentation = "https://docs.rs/cargo-test-support"
```
weihanglo, line 9: ditto

crates/cargo-test-macro/Cargo.toml
```diff
@@ -1,13 +1,20 @@
 [package]
 name = "cargo-test-macro"
-version = "0.1.0"
+version = "0.1.1"
```
weihanglo, line 3: Why bumping to `0.1.1`?

weihanglo (commented):
> However, the test still fails on: `Error: cargo-test-support not found in registry (crates.io)`

We might want to do something like this, until we publish a version

https://github.com/rust-lang/cargo/pull/13178/commits/633929d3aa142730466aeb1c7c44001edcc92d57#diff-017fb78a9378b299832fb9eadf0e7b8c60ecca480f2415a168e1b48a5b6f3486

crates/cargo-test-macro/Cargo.toml
PaulDance: I simply made these crates have a manifest similar to the other ones, in this aspect also. I'll remove the `include` if you want though.

crates/cargo-test-macro/Cargo.toml
PaulDance: This was to make `validate-version-bump.sh` happy.

crates/cargo-test-macro/Cargo.toml
weihanglo: `home` is the only one there because it was imported from an external crate. Lean towards removing :)

crates/cargo-test-macro/Cargo.toml
weihanglo: Thanks. Got it. Let's bump it to 0.2.0 so we don't need to bump it again.

BTW, I just published them to crates.io and am now waiting for infra team to accept ownership transfer.

* https://crates.io/crates/cargo-test-support
* https://crates.io/crates/cargo-test-macro

weihanglo (approved):

weihanglo:
Cool. We're all set. Thanks Paul!

@bors r+

PaulDance:
@weihanglo Thank you!

PaulDance:
@weihanglo Do you have any news about this? The [version currently uploaded](https://docs.rs/cargo-test-support/0.1.0/cargo_test_support/index.html) is still `0.1.0` and empty while the ownership transfer seems to have been accepted. Isn't running `publish.py` part of the usual Cargo team workflow? Does it actually not work for `cargo-test-*`?

weihanglo:
> [rustbot](https://github.com/rustbot) added this to the [1.79.0](https://github.com/rust-lang/cargo/milestone/75) milestone

member crates have the same release schedule as the main Cargo binary / library.

See https://doc.crates.io/contrib/process/release.html#cratesio-publishing

PaulDance:
Got it, thanks!

PaulDance:
Just popping in to say that I've now switched to it and it works great, thanks again!

PR #13422, opened 2024-02-08
fix(compiler): Clarify Finished message

epage:

As highlighted on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/cargo.20build.20default.20profile/near/418316528), many users think "Rust is slow" because of the `dev` profile.

While a perfect solution is still being worked out, this attempts what will hopefully be smaller, incremental step that hopefully maintains balance of the different needs.

We are changing the message from:
```
Finished dev [unoptimized + debuginfo] target(s) in [..]s
```
to
```
Finished `dev` profile [unoptimized + debuginfo] target(s) in [..]s
```
where `dev profile` is a link to [the Cargo book](https://doc.rust-lang.org/cargo/reference/profiles.html#profiles).

The intent is
- Clarify what `dev` even means
  - Add `profile` to give it context
  - Quote it to highlight this is something mutable
- Make the profile content stand out on hover by being a link
- Help people learn more by following the link

For now, this leaves the profile description alone.

### DIFF
--- a/src/cargo/core/compiler/job_queue/mod.rs
+++ b/src/cargo/core/compiler/job_queue/mod.rs
@@ -806,8 +806,14 @@ impl<'cfg> DrainState<'cfg> {
             // `display_error` inside `handle_error`.
             Some(anyhow::Error::new(AlreadyPrintedError::new(error)))
         } else if self.queue.is_empty() && self.pending_queue.is_empty() {
-            let message =
-                format!("`{profile_name}` profile [{opt_type}] target(s) in {time_elapsed}",);
+            let profile_link = cx.bcx.config.shell().err_hyperlink(
+                "https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles",
+            );
+            let message = format!(
+                "{}`{profile_name}` profile [{opt_type}]{} target(s) in {time_elapsed}",
+                profile_link.open(),
+                profile_link.close()
+            );
             if !cx.bcx.build_config.build_plan {
                 // It doesn't really matter if this fails.
                 let _ = cx.bcx.config.shell().status("Finished", message);
--- a/src/cargo/core/compiler/job_queue/mod.rs
+++ b/src/cargo/core/compiler/job_queue/mod.rs
@@ -806,7 +806,8 @@ impl<'cfg> DrainState<'cfg> {
             // `display_error` inside `handle_error`.
             Some(anyhow::Error::new(AlreadyPrintedError::new(error)))
         } else if self.queue.is_empty() && self.pending_queue.is_empty() {
-            let message = format!("{profile_name} [{opt_type}] target(s) in {time_elapsed}",);
+            let message =
+                format!("`{profile_name}` profile [{opt_type}] target(s) in {time_elapsed}",);
             if !cx.bcx.build_config.build_plan {
                 // It doesn't really matter if this fails.
                 let _ = cx.bcx.config.shell().status("Finished", message);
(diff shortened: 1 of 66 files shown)


src/cargo/core/compiler/job_queue/mod.rs
```diff
@@ -806,9 +806,14 @@ impl<'cfg> DrainState<'cfg> {
...
             Some(anyhow::Error::new(AlreadyPrintedError::new(error)))
         } else if self.queue.is_empty() && self.pending_queue.is_empty() {
+            let profile_link =
+                cx.bcx.config.shell().err_hyperlink(
+                    "https://doc.rust-lang.org/cargo/reference/profiles.html#profiles",
+                );
             let message = format!(
-                "{} [{}] target(s) in {}",
-                profile_name, opt_type, time_elapsed
+                "{}`{profile_name}` profile{} [{opt_type}] target(s) in {time_elapsed}",
```
weihanglo, line 814: Not on topic, but we might way to make Hyperlink fancier like anstyle for open/close.

epage: I had considered that but didn't want to include it in this PR

weihanglo, line 814: What about making the entire `dev profile [unoptimized + debuginfo]` clickable? They are all profile related.

weihanglo (approved):
I am good with this change. Not sure if we want a quick poll.

src/cargo/core/compiler/job_queue/mod.rs
epage: Good idea.

epage:
I also changed the link to point directly to the built-in profile definitions so they can see what `dev` means and then browse further for what those fields mean.

weihanglo:
> I also changed the link to point directly to the built-in profile definitions so they can see what `dev` means and then browse further for what those fields mean.

Sounds like we can make it smarter by linking to each profile?

weihanglo:
Anyway, this is a two-way door. Going to merge this.

@bors r+

bend-n:
I dont like this change-- its rather annoying, and not very useful.
Im pretty sure a total of two people have clicked on the link intentionally (not to mention low terminal support for non XTerm feature), and the people who click on it intentionally are simply annoyed, as theres no need for this guide. Also, the pr description says that `\`dev\` profile` is a link, but the entirety of `\`dev\` profile [unoptimized + debuginfo]` is a link.

epage:
@bend-n could you clarify why you find this change to be annoying?  We expect to be adding a lot more links to Cargo and Rustc's outputto better connect the commands to their documentation.

> as theres no need for this guide

This runs a bit counter to our experience talking to users as we found they didn't know what `dev` meant.  Granted, also clarified it is a "profile" which also helps.

bend-n:
its disruptive, as you can accidentally click on it, and in certain terminals has pop ups that are annoying. however, i have since solved my problems:
```fish
set -x FORCE_HYPERLINK 0
```

epage:
Note that `FORCE_HYPERLINK` is not officially support and is a side effect of the implementation.

You should probably be using [`CARGO_TERM_HYPERLINKS=false`](https://doc.rust-lang.org/cargo/reference/config.html#termhyperlinks)

bend-n:
oh, thats cool. thanks!

epage:
You set
```toml
[term]
hyperlinks = false
```
and it didn't work?

bend-n:
well, i set
```toml
[build]
rustflags = ["-Ctarget-cpu=native"]
target-dir = "/mnt/HDD500/.target"

[unstable]
gc = true

[term]
hyperlinks = true
unicode = true

```
and it didntt work.

workingjubilee:
Would this be improved by shrinking the range of the hyperlink?

epage:
The point of the range of the hyperlink was to cover everything profile related.  The most likely alternative (what I wish I had done but hadn't felt motivated enough to implement after the PR) was to link directly to the documentation for the referenced fields, instead of one monolithic link.

Again, as I said, this also isn't expected to be isolated.  Both Cargo and Rustc are looking to take advantage of more links in the output to help bridge the tool / documentation divide.  Shrinking in a one off case will have little impact if we have a lot more links.  Its more of needing to decide if the general vision for links needs to be adjusted and, if so, to what.

PR #13439, opened 2024-02-13
Remove `rustc-check-cfg=` from the reserved directive prefixes

Urgau:
While `rustc-check-cfg=` is still nightly-only we can still remove it from the old build script reserved prefixes. I think we should do it, so let's remove it (while we still can).

r? @epage

### DIFF
--- a/src/cargo/core/compiler/custom_build.rs
+++ b/src/cargo/core/compiler/custom_build.rs
@@ -716,7 +716,6 @@ impl BuildOutput {
             "rustc-link-arg-examples=",
             "rustc-link-arg=",
             "rustc-cfg=",
-            "rustc-check-cfg=",
             "rustc-env=",
             "warning=",
             "rerun-if-changed=",


epage:
Should we have a transition period (warn, error)?

We at least need to make sure this is updated in rust-lang, see https://github.com/rust-lang/rust/blob/a84bb95a1f65bfe25038f188763a18e096a86ab2/compiler/rustc_llvm/build.rs#L105

There are also other users of it, see https://github.com/search?q=rustc-check-cfg&type=code

Urgau:
Unfortunately yes, since as noted `rust-lang/rust` will not pass; but it will also probably not pass with a deprecation warning either.

I will send PR for `rustc_llvm` and `libc`, and then update this PR with a deprecation warning.

Urgau:
After some chatting with @epage, we no longer think this is a good idea, since this will interfere with MSRV, as the double colon `cargo::` was only stabilized in 1.77.

PR #13441, opened 2024-02-15
chore: Update snapbox

epage:


hi-rustin (approved):
Looks good to me. Thanks! 👍

Muscraft:
This looks good to me! I am a little worried about this working in `rust-lang/rust` but I think your solution should work.

PR #13442, opened 2024-02-15
chore(ci): Drop MSRV:1 patch field

epage:
This was missed when I did MSRV:3

This make it less annoying for people to keep up on every patch with their own MSRVs.


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #13443, opened 2024-02-15
feat: Expost host and target sysroot to build scripts.

### LINKED ISSUE #7501: Rust sysroot is not exposed to build scripts
When building with wrappers like xargo or the new `-Z build std` flag, rustc invocations get a `--sysroot [something]` flag passed to them. Any build scripts that rely on running a new rustc invocation that matches the current build configuration as closely as possible (such as [autocfg](https://crates.io/crates/autocfg) is out of luck, because replicating the sysroot needs to be done manually. If the sysroot path were exposed through a `RUST_SYSROOT` environment variable to build scripts then this would be easier to deal with.

c272:
Currently, when using `-Zbuild-std`, Cargo passes a `--sysroot` argument down to the relevant `rustc` invocations. However, the value of this argument is not exposed to build scripts, so if a build script wishes to mirror the `rustc` configuration of the host, they must somehow discover the sysroot themselves.

This patch exposes the host (`RUSTC_HOST_SYSROOT`) and target (`RUSTC_TARGET_SYSROOT`) sysroots to build scripts through additional environment variables.

Fixes #7501
Related to https://github.com/rust-lang/wg-cargo-std-aware/issues/50

An example build script displaying these environment variables is as follows:
```rust
use std::env;

fn main() {
    println!("host: {}", env::var("RUSTC_HOST_SYSROOT").unwrap());
    println!("target: {}", env::var("RUSTC_TARGET_SYSROOT").unwrap());
}
```

### Additional information
N/A

### DIFF
--- a/src/cargo/core/compiler/custom_build.rs
+++ b/src/cargo/core/compiler/custom_build.rs
@@ -32,10 +32,10 @@
 //! [instructions]: https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script
 
 use super::{fingerprint, BuildRunner, Job, Unit, Work};
-use crate::core::compiler::artifact;
 use crate::core::compiler::build_runner::Metadata;
 use crate::core::compiler::fingerprint::DirtyReason;
 use crate::core::compiler::job_queue::JobState;
+use crate::core::compiler::{artifact, CompileKind};
 use crate::core::{profiles::ProfileRoot, PackageId, Target};
 use crate::util::errors::CargoResult;
 use crate::util::machine_message::{self, Message};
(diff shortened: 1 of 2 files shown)


weihanglo:
Sorry I don't quite follow. I read through the relevant issues but haven't found any design decision made. Do we have discussions in other places like Zulip I missed?

weihanglo:
Asking because we would like to have design discussions in issue threads, and PRs is for discussing each implementation separately. That would make things a bit easier to follow :)

adamgemmell:
Hi, sorry about that! I advised c272 to share the implementation first the best way to move the discussion forwards. We should have followed the contributor's guide a bit more closely - hope it didn't seem too pushy.

Writing the patch fleshed this idea out a bit more in the process, so one of us will share that in a separate feature request issue.

adamgemmell:
This PR, along with https://github.com/rust-lang/cargo/issues/13473, can be closed because https://github.com/rust-lang/wg-cargo-std-aware/issues/50 has been closed

PR #13444, opened 2024-02-15
Update tests for changes in latest nightly

ehuss:
There were two changes that has caused nightly tests to start failing:
* LLVM was updated, and that caused a change in the expected data layout for the target. (https://github.com/rust-lang/rust/pull/120055 update to LLVM 18)
* https://github.com/rust-lang/rust/pull/121049 Do not point at `#[allow(_)]` as the reason for compat lint triggering. The test was looking for the `unused_variables` note, which is no longer printed.


weihanglo:
build-std passed so should be good

@bors r+

PR #13448, opened 2024-02-16
fix(test): Suggest `--` for libtest arguments

### LINKED ISSUE #12494: Make `libtest` flags more discoverable
### Problem

The `libtest` unit test runner has a lot of flags. Cargo allows passing these through with `cargo test --lib -- `. This is not super easy to discover - `cargo test --help` *does* say "Run `cargo test -- --help` for test binary options" at the end, but if you don't know that libtest and cargo have different flags, the error messages aren't great:
```
; cargo +nightly test --ignored
error: unexpected argument '--ignored' found

  tip: a similar argument exists: '--ignore-rust-version'

Usage: cargo test --ignore-rust-version [TESTNAME] [-- [args]...]

```

### Proposed Solution

Ideally, cargo would know which flags libtest supports and suggest passing those through:
```
error: unexpected argument '--ignored' found

  tip: libtest supports the `--ignored` argument; try `cargo test -- --ignored`
  tip: a similar argument exists: '--ignore-rust-version'

Usage: cargo test --ignore-rust-version [TESTNAME] [-- [args]...]
```

### Notes

Right now, it's kind of hard to parse the `--help` output of libtest programmatically, so cargo would either have to hack a parser together or hardcode the options, neither of which is particularly maintainable. Ideally `--help --format json` would output a structured list that cargo can easily parse (I can open an upstream issue in rust-lang/rust if you're interested in pursuing this) and then idk embed it at build time or something like that.

My motivation for opening this issue is that I want to suggest `--ignored` in libtest if 1 or more tests are ignored, but right now that will be kind of confusing for people using cargo because it's a libtest flag and not a `cargo test` flag.

epage:
We already do this so long as the argument doesn't look like a `cargo test` argument (e.g. `--show-output`)
but `--ignored` looks like `--ignore-rust-version` do the the suggestion algorithms prefix checks.

Before
```
error: unexpected argument '--ignored' found

  tip: a similar argument exists: '--ignore-rust-version'

Usage: cargo test --ignore-rust-version [TESTNAME] [-- [ARGS]...]

```
After
```
error: unexpected argument '--ignored' found

  tip: a similar argument exists: '--ignore-rust-version'
  tip: to pass '--ignored' as a value, use '-- --ignored'

Usage: cargo test --ignore-rust-version [TESTNAME] [-- [ARGS]...]

```

This was fixed in clap-rs/clap#5356 and we just need to update to take advantage of it.

Fixes #12494

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -523,18 +523,18 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.4.18"
+version = "4.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e578d6ec4194633722ccf9544794b71b1385c3c027efe0c55db226fc880865c"
+checksum = "c918d541ef2913577a0f9566e9ce27cb35b6df072075769e0b26cb5a554520da"
 dependencies = [
  "clap_builder",
 ]
 
 [[package]]
 name = "clap_builder"
-version = "4.4.18"
+version = "4.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4df4df40ec50c46000231c914968278b1eb05098cf8f1b3a518a95030e71d1c7"
+checksum = "9f3e7391dad68afb0c2ede1bf619f579a3dc9c2ec67f089baa397123a2f3d1eb"
 dependencies = [
  "anstream",
  "anstyle",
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
Although I don't quite think "pass as a value" is straightforward for people to understand, it does give an example to follow. Thanks for the fix!

epage:
Agreed that its not the greatest but its also what we are already displaying in other cases.  This just fixes an inconsistency about displaying it.

PR #13461, opened 2024-02-19
test(cli): Verify terminal styling

epage:

This uses a new feature from snapbox that let's us render terminal styling in SVG files.  This let's us see / visualize ANSI escape codes, including in github's UI (will render images, including side-by-side images for diffs).

Potential concerns
- assert-rs/trycmd#257: If the snapshot file doesn't exist, you need to `SNAPSHOTS=overwrite` twice
- #9012: because we narrow the enabling of colors to cargo via `CARGO_TERM_COLOR`, clap doesn't know about it and those are rendered without color
- File size: `du -ch tests` reported 13MB before and after
- Loss of information: this doesn't yet include links but we already weren't snapshotting these
- Unconditionally turning on styling and snapshotting for it for all "ui" tests
- Too many SVGs in one PR can negatively affect github's UI


weihanglo (commented):
image

This is wild 🤯, but pretty neat!

Do you have an example of side-by-side image diff for this new feature?

epage:
> Do you have an example of side-by-side image diff for this new feature?

See https://github.com/rust-cli/anstyle/pull/174/commits/d5111b019fd2cf450a18233b807b670a814c48f0

weihanglo:
One pretty minor concern is that GitHub Mobile renders SVG as text. (Who else review PRs on mobile? 🤷🏾‍♂️)

epage:
> One pretty minor concern is that GitHub Mobile renders SVG as text. (Who else review PRs on mobile? 🤷🏾‍♂️)

How viewable do you feel that is to look at the raw SVG?  I tried to keep the overhead down to make comparing them in text form easier (and to keep the file sizes down).

weihanglo:
> > One pretty minor concern is that GitHub Mobile renders SVG as text. (Who else review PRs on mobile? 🤷🏾‍♂️)
> 
> How viewable do you feel that is to look at the raw SVG? I tried to keep the overhead down to make comparing them in text form easier (and to keep the file sizes down).

At the point that I won't review them on mobile. Keeping the size down is good, and I don't think we will go back to text diff too often. Even we do, git word diff should be sufficient to help review.

Specific to this PR, the "Files changed" tab is janky for loading images that I cannot even review them, and some of them stop rendering.

image

tests/testsuite/cargo_add/git/stderr.term.svg
```diff
@@ -0,0 +1,30 @@
+<svg width="831.6px" height="72px" xmlns="http://www.w3.org/2000/svg">
```
epage, line 1:
```
 ---- expected: tests/testsuite/cargo_remove/invalid_package/stderr.term.svg
++++ actual:   stderr
   1      - <svg width="890.4000000000001px" height="36px" xmlns="http://www.w3.org/2000/svg">
        1 + <svg width="882px" height="36px" xmlns="http://www.w3.org/2000/svg">
```
The presence of variable substitution is throwing off the `width` calculations

epage:
```
 ---- cargo_add::target::case stdout ----
thread 'cargo_add::target::case' panicked at tests/testsuite/cargo_add/target/mod.rs:34:10:
tests/testsuite/cargo_add/target/stdout.term.svg (expected):
Failed to read /Users/runner/work/cargo/cargo/tests/testsuite/cargo_add/target/stdout.term.svg: No such file or directory (os error 2)
```
A very generic test name combined with a very broad `.gitignore` entry caused me to not add these files.

Unsure if we should rename the test or rename the ignore to `/target`

tests/testsuite/cargo_add/git/stderr.term.svg
epage: Tracking in assert-rs/trycmd#258

epage:
We should instead not generate snapshots for empty content

epage:
We need to rename the `target` tests so they don't run into issues with our gitignore file.

weihanglo (approved):
> File size: du -ch tests reported 13MB before and after

This is 8.4 to 8.6MB on my machine, which seems pretty acceptable.

BTW, we should update contributor guide <https://doc.crates.io/contrib/tests/writing.html> after this change.

weihanglo:
I only reviewed 70 of all files changes, but assuming most of the changes are generated, I am going to check my box. If people disagree on it later, we can always revert.

@bors r+

PR #11029, opened 2022-08-29
refactor(cli): Lazy load config

epage:
This is trying to clarify `-C` support when it is implemented in #10952 

Cargo currently has two initialization states for Config, `Config::default` (process start) and `config.configure` (after parsing args).  The most help we provide for a developer touching this code is a giant `CAUTION` comment in one of the relevant functions.

Currently, #10952 adds another configuration state in the middle where the `current_dir` has been set.

The goal of this PR is to remove that third configuration state by
- Lazy loading `Config::default` so it can be called after parsing `-C`
- Allowing `-C` support to assert that the config isn't loaded yet to catch bugs with it

The hope is this will make the intent of the code clearer and reduce the chance for bugs.

In doing this, there are two intermediate refactorings
- Make help behave like other subcommands
  - Before, we had hacks to intercept raw arguments and to intercept clap errors and assume what their intention was to be able to implement our help system.
  - This flips it around and makes help like any other subcommand,
simplifying cargo initialization.
  - We had to upgrade clap because this exposed a bug where `Command::print_help` wasn't respecting `disable_colored_help(true)`
- Delay fix's access to config

Personally, I also find both changes make the intent of the code clearer.

To review this, I recommend looking at the individual commits.   As this is just refactors, this has no impact on testing.

PR #13463, opened 2024-02-20
fix(cli): Control clap colors through config

### LINKED ISSUE #9012: Cargo doesn't respect `--color` ~~or `CARGO_TERM_COLOR`~~ for help text and errors from clap

epage:

Part of #9012

To accomplish this, I pivoted in how we handle `-C`.  In #11029, I made the config lazily loaded.  Instead, we are now reloading the config for `-C` like we do for "cargo script" and `cargo install`.  If there is any regression, it will be felt by those commands as well and we can fix all together.  As this is unstable, if there is a regression, the impact is less.  This allowed us access to the full config for getting the color setting, rather than taking more limited approaches like checking only `CARGO_TERM_CONFIG`.

### DIFF
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -1,5 +1,4 @@
 use anyhow::{anyhow, Context as _};
-use cargo::core::shell::Shell;
 use cargo::core::{features, CliUnstable};
 use cargo::{drop_print, drop_println, CargoResult};
 use clap::builder::UnknownArgumentValueParser;
@@ -15,7 +14,11 @@ use crate::command_prelude::*;
 use crate::util::is_rustup;
 use cargo::util::style;
 
-pub fn main(lazy_gctx: &mut LazyContext) -> CliResult {
+pub fn main(gctx: &mut GlobalContext) -> CliResult {
+    // CAUTION: Be careful with using `config` until it is configured below.
+    // In general, try to avoid loading config values unless necessary (like
+    // the [alias] table).
+
     let args = cli().try_get_matches()?;
 
     // Update the process-level notion of cwd
@@ -38,13 +41,8 @@ pub fn main(lazy_gctx: &mut LazyContext) -> CliResult {
             .into());
         }
         std::env::set_current_dir(&new_cwd).context("could not change to requested directory")?;
-        lazy_gctx.get_mut().reload_cwd()?;
+        gctx.reload_cwd()?;
     }
-
-    // CAUTION: Be careful with using `config` until it is configured below.
-    // In general, try to avoid loading config values unless necessary (like
-    // the [alias] table).
-    let gctx = lazy_gctx.get_mut();
 
     let (expanded_args, global_args) = expand_aliases(gctx, args, vec![])?;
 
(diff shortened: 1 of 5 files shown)


src/bin/cargo/cli.rs
```diff
@@ -40,6 +38,7 @@ pub fn main(lazy_gctx: &mut LazyContext) -> CliResult {
             .into());
         }
         std::env::set_current_dir(&new_cwd).context("could not change to requested directory")?;
+        lazy_gctx.get_mut().reload_cwd()?;
```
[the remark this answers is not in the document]
weihanglo: This is not a blocker btw, given `-C` is unstable.

weihanglo (commented):
With this patch, the color control works for `CARGO_TERM_COLOR` environment variable and `config.toml`. However, it doesn't work with either `--color never` or `--config 'term.color="never"'`, so I don't think #9012 is resolved.

Not sure if we should ship a partial solution or hold on until `--color` and `--config` get addressed.

src/bin/cargo/cli.rs
```diff
@@ -514,7 +513,19 @@ impl GlobalArgs {
     }
 }
 
-pub fn cli() -> Command {
+pub fn cli(gctx: &GlobalContext) -> Command {
+    // Don't let config errors get in the way of parsing arguments
+    let term = gctx.get::<TermConfig>("term").unwrap_or_default();
```
weihanglo, line 518: The issue is that we sidestep without calling `GlobalContext::merge_cli_args`, which `GlobalContext::configure` does. 

There might also be a discrepancy when we call `reload_cwd()`, but that is minimal as the code path eventually calls `GlobalContext::configure`.

epage:
> Not sure if we should ship a partial solution or hold on until --color and --config get addressed.

I don't see a reasonable way for addressing that part.  We need to be able to parse the command line to get those which requires knowing what level of styling to use.  We could say "parse the command line, take the result, and strip the message based on our own decision".  However, `--help` and errors short-circuits CLI parsing.  We'd have to parse, see if an error gets back, and then re-parse using the "ignore errors" setting which is a very brittle feature in clap.

src/bin/cargo/cli.rs
epage: I'm not too sure what the concern is here.

Is it that we aren't respecting `--color` which you raised elsewhere and you are just calling out the relevant code?

Or is it that we aren't respecting the config from `-C` to control the colors from CLI parsing?  If so, that is a similar problem to `--color` support (and more difficult / brittle to support)

weihanglo: I was just making sure we're aware of this and also #9012 is not completely resolved. Should we reword #9012 to track `--config` and `--color` or open a new one?

I am good with merge this PR as-is btw. Feel free to `r=weihanglo`.

weihanglo (approved):

weihanglo:
Seems like we can leverage <https://github.com/rust-lang/cargo/pull/13461> to test colors for this PR, right?

weihanglo:
Okay I didn't notice you've linked the original issue in that PR.

PR #13466, opened 2024-02-20
Fix more redundant imports.

ehuss:
The latest nightly has started warning about redundant imports. This removes those warnings.

Presumably these were casualties of merging between #13464 and #13409.


weihanglo:
@bors r+

Thanks!

PR #13467, opened 2024-02-20
Add global_cache_tracker stability tests.

ehuss:
This adds some tests to ensure that the database used in the global cache tracker stays compatible across versions. These tests work by using rustup to run both the current cargo and the stable cargo, and verifying that when switching versions, everything works as expected.

These tests will be ignored on developer environments if they don't have rustup, or don't have the stable toolchain installed. It does assume that "stable" is at least 1.76. It is required for the tests to run in CI, but will be disabled in rust-lang/rust since it does not have rustup.

I'm not expecting too much trouble with these tests, but if they become too fiddly or broken, they can always be changed or removed.

The support code for running "cargo +stable" is very basic right now. If we expand to add similar tests in the future, then I think we could consider adding support functions (such as [`tc_process`](https://github.com/rust-lang/cargo/blob/64ccff290fe20e2aa7c04b9c71460a7fd962ea61/tests/testsuite/old_cargos.rs#L21-L36)) to make it easier or more reliable.

### DIFF
--- a/crates/cargo-test-macro/src/lib.rs
+++ b/crates/cargo-test-macro/src/lib.rs
@@ -248,6 +248,13 @@ fn has_rustup_stable() -> bool {
         // This cannot run on rust-lang/rust CI due to the lack of rustup.
         return false;
     }
+    if cfg!(windows) && !is_ci() && option_env!("RUSTUP_WINDOWS_PATH_ADD_BIN").is_none() {
+        // There is an issue with rustup that doesn't allow recursive cargo
+        // invocations. Disable this on developer machines if the environment
+        // variable is not enabled. This can be removed once
+        // https://github.com/rust-lang/rustup/issues/3036 is resolved.
+        return false;
+    }
     check_command("cargo", &["+stable", "--version"])
 }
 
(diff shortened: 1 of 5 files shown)


crates/cargo-test-macro/src/lib.rs
```diff
@@ -255,7 +257,16 @@ fn has_rustup_stable() -> bool {
         // https://github.com/rust-lang/rustup/issues/3036 is resolved.
         return false;
     }
-    check_command("cargo", &["+stable", "--version"])
+    // Cargo mucks with PATH on Windows, adding sysroot host libdir, which is
+    // "bin", which circumvents the rustup wrapper. Use the path directly from
+    // CARGO_HOME.
```
weihanglo, line 262: Would <https://github.com/rust-lang/cargo/pull/13468> fix this issue?

ehuss: Maybe! It will need to wait until that hits stable, though.

weihanglo: Oh, yeah. We need it in stable :)

weihanglo (approved):

weihanglo:
Thanks for this!

@bors r+

PR #7798, opened 2020-01-14
Fix wrong directories in host_libdir.

jnbr:
This fixes a regression from #7482 where the sysroot_target_libdir leaks into the host libdir. This can cause problems when the dynamic linker does not ignore the target libraries but tries to load them instead. This happens for example when building on x86_64-musl for aarch64-musl.

PR #4006, opened 2017-05-08
fix `cargo test` of dylib projects for end user runs too

mcgoo:
Fixes running `cargo test` and `cargo test --target <target>` for dylib projects.

Moves the logic just landed in https://github.com/rust-lang/cargo/pull/3996 into cargo itself, so cargo sets the dylib path for anyone running `cargo test` or `cargo run`. Current master sets the dylib path only for  `cargo test` on cargo itself.

This PR pins to rustup 1.2.0 for the purposes of testing. If https://github.com/rust-lang-nursery/rustup.rs/pull/1093 ends up working out, then this PR would only be important for non-rustup users and people doing cross testing, `cargo test --target <target>`.

Arguably https://github.com/mcgoo/cargo/blob/ed273851f8bc76f726eda4a2e2a7bb470c3718bc/src/cargo/ops/cargo_rustc/context.rs#L249-L253 should point to lib/rustlib/\<host triple\>/lib instead of sysroot/lib, because I think if the libs are different, you will never be able to compile a working plugin anyway, and for the host=target case you get the  lib/rustlib/\<host triple\>/lib anyhow. Is there ever a case where the lib/rustlib/\<host triple\>/lib and sysroot/lib versions of the libs would be expected to differ?

This is not a huge deal for me one way or the other - it doesn't impact my workflow at all. I nearly dropped it when I saw @alexcrichton had made it all work in 3996, but I think it's worth doing because it removes a surprise. It certainly would have saved me a couple of days of confusion. Either way, thanks for looking it over.

PR #13468, opened 2024-02-20
fix: remove unused `sysroot_host_libdir`

weihanglo:

Copy from <https://github.com/rust-lang/cargo/pull/10469#issuecomment-1079503079>:

> I've never been entirely clear why it does this. #4006 didn't really
> explain why it added the corresponding host_dylib_path. I can't envision
> a scenario where it matters. I think compiler plugins and proc-macros
> should load just fine, since libstd.so should already be loaded by the
> compiler. Also, rustc uses rpath these days, and on Windows libstd.so is
> placed in the bin directory which will be searched first anyways.
>
> On balance, I think it should be safe to just remove sysroot_host_libdir.
> I can't come up with a scenario where it matters, at least on
> windows/macos/linux. One issue is that this is most likely to affect
> plugins, but those are deprecated and I think only Servo was the real
> holdout. A concern is that nobody is going to test this use case before
> it hits stable.

Also,

* compiler plugins were removed https://github.com/rust-lang/rust/pull/116412
* servo has moved off from plugins: https://github.com/servo/servo/pull/30508

So should generally be fine.

There might be some edge case like <https://github.com/rust-lang/cargo/pull/7798> passing `--sysroot`, though I would also expect rustc to load sysroot from provided path. Hence safe?

### DIFF
--- a/src/cargo/core/compiler/build_context/target_info.rs
+++ b/src/cargo/core/compiler/build_context/target_info.rs
@@ -48,9 +48,6 @@ pub struct TargetInfo {
     support_split_debuginfo: Vec<String>,
     /// Path to the sysroot.
     pub sysroot: PathBuf,
-    /// Path to the "lib" or "bin" directory that rustc uses for its dynamic
-    /// libraries.
-    pub sysroot_host_libdir: PathBuf,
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks!

@bors r+

PR #13469, opened 2024-02-21
Fix unused imports on Windows.

ehuss:
These imports were causing a warning on Windows because they are only used in a section that is conditional for Unix.


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #13479, opened 2024-02-22
fix(cli): Respect CARGO_TERM_COLOR in '--list' and '-Zhelp'

epage:

Similar to #9012, we aren't respecting `CARGO_TERM_COLOR` for `-Zhelp` and other places.  This corrects that.

#9012 was about initialization order to get the value.  Here, the problem is that we don't update `Shell` with `CARGO_TERM_COLOR`.

In doing this, I was concerned about keeping track of where it is safe to call `config_configure` without running it twice.  To this end, I refactored `main` to make it clear that each call to `config_configure` is in a mutually exclusive path that exists immediately.

### Additional information

Found this with the test for `-Zhelp` in #13461.

### DIFF
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -48,46 +48,38 @@ pub fn main(gctx: &mut GlobalContext) -> CliResult {
 
     let (expanded_args, global_args) = expand_aliases(gctx, args, vec![])?;
 
+    let is_verbose = expanded_args.verbose() > 0;
+
     if expanded_args
         .get_one::<String>("unstable-features")
         .map(String::as_str)
         == Some("help")
     {
         print_zhelp(gctx);
-        return Ok(());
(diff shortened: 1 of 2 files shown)


weihanglo (commented):
The question is that how resilient to broken `.cargo/config.toml` these commands should be. Calling `config_configure` implies they fail if `.cargo/config.toml` is ill-formatted.

epage:
> The question is that how resilient to broken .cargo/config.toml these commands should be. Calling config_configure implies they fail if .cargo/config.toml is ill-formatted.

This was caught in test failures and I've hardened it to make it work
- We make sure we initialize the Shell before erroring out
- We ignore the errors

weihanglo (approved):

weihanglo:
Thanks. This looks great 👍🏾 

@bors r+

Issue #13394, opened 2024-02-04
Include location of workspace manifest in error loading workspace


fawq:
### Problem

I've tried to create new project inside project which was previously created by cargo init and removed. Now when I want to create new project inside, I've got an error that there is a missing Cargo.toml in parent directory, but parent directory doesn't need to have Cargo.toml because it should be simple directory contains all rust projects.

Steps:
1. Create directory e.g. "/home/ubuntu/Projects/Rust/"
2. Go inside this directory and type `cargo init`
3. Change mind and remove all files `rm -r *`
4. Create new project inside this directory e.g. `cargo new test_rust_wsl`
5. New directory is created "/home/ubuntu/Projects/Rust/test_rust_wsl"
6. Go inside new directory and try build `cargo build`

Get the error: 
> error: failed to load manifest for workspace member /home/ubuntu/Projects/Rust
> 
> Caused by:
>   failed to read /home/ubuntu/Projects/Rust/Cargo.toml
> 
> Caused by:
>   No such file or directory (os error 2)

### Version

```text
cargo 1.75.0 (1d8b05cdd 2023-11-20)
release: 1.75.0
commit-hash: 1d8b05cdd1287c64467306cf3ca2c8ac60c11eb0
commit-date: 2023-11-20
host: x86_64-unknown-linux-gnu
libgit2: 1.7.1 (sys:0.18.1 vendored)
libcurl: 8.4.0-DEV (sys:0.4.68+curl-8.4.0 vendored ssl:OpenSSL/1.1.1u)
ssl: OpenSSL 1.1.1u  30 May 2023
os: Ubuntu 22.04 (jammy) [64-bit]
```

ehuss:
Hm, it seems like there is a possible step missing from the reproduction. Can you check if you have a `Cargo.toml` located in any parent directory of `/home/ubuntu/Projects/Rust`? Or perhaps there is a workspace table in `/home/ubuntu/Projects/Rust/test_rust_wsl/Cargo.toml`?

fawq:
![image](https://github.com/rust-lang/cargo/assets/30419862/c719c7c9-8246-4425-a57e-da370340510a)
![image](https://github.com/rust-lang/cargo/assets/30419862/97c4d1ef-cb04-4fdd-9c0b-68a5b4131d11)
As you can see, there is no other Cargo.toml in a whole directory (except the one which I need and cannot build)

ehuss:
[...]

PR #13480, opened 2024-02-22
Error messages when collecting workspace members now mention the workspace root location

Cerber-Ursi:
Fixes https://github.com/rust-lang/cargo/issues/13394

This is a small tweak in error messages for workspaces, with intention to make it easier for users who accidentally created some unwanted workspace and then was surprised by the error coming seemingly out of nowhere. The exact inspiration for the change was [this comment](https://users.rust-lang.org/t/cargo-using-wrong-directory/107126/9?u=cerber-ursi) in discussion on URLO.

This is a simple change in error messages, so the existing test suite should probably be enough. Requests for changing the text further are welcome, of course.

### DIFF
--- a/src/cargo/core/workspace.rs
+++ b/src/cargo/core/workspace.rs
@@ -724,8 +724,8 @@ impl<'gctx> Workspace<'gctx> {
                         continue;
                     }
                     bail!(
-                        "package `{}` is listed in default-members\n\
-                        for workspace at {} but is not a member.",
+                        "package `{}` is listed in default-members but is not a member\n\
+                        for workspace at {}.",
                         path.display(),
                         root_manifest_path.display()
                     )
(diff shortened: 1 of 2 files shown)


weihanglo:
This also fixes <https://github.com/rust-lang/cargo/issues/13394>, right?

src/cargo/core/workspace.rs
```diff
@@ -722,9 +724,10 @@ impl<'gctx> Workspace<'gctx> {
                         continue;
                     }
                     bail!(
-                        "package `{}` is listed in workspace’s default-members \
-                         but is not a member.",
-                        path.display()
+                        "package `{}` is listed in default-members\n\
+                        for workspace at {} but is not a member.",
```
epage, line 728:
```suggestion
                        "package `{}` is listed in default-members but is not a member\n\
                        for workspace at {}.",
```
This keeps the need-to-parse part of the message up front and people can more easily glance at the rest of the message

Cerber-Ursi:
> This also fixes #13394, right?

It does indeed, added this fact to the OP.

Cerber-Ursi:
Sorry for the breakage - changed the error message as advised in review, but forgot about tests. Now they should be green again.

PR #13481, opened 2024-02-22
Add all unit's children recursively for `doc.extern-map` option

GuillaumeGomez:
Fixes https://github.com/rust-lang/rust/issues/120993.

It allows to fix link generation of reexport of foreign items when deps documentation is not generated. Sometimes, we need to have information about not only direct dependencies but also of dependencies of these direct dependencies and so on.

### DIFF
--- a/src/cargo/core/compiler/rustdoc.rs
+++ b/src/cargo/core/compiler/rustdoc.rs
@@ -102,6 +102,63 @@ impl hash::Hash for RustdocExternMap {
     }
 }
 
+/// Recursively generate html root url for all units and their children.
+///
+/// This is needed because in case there is a reexport of foreign reexport, you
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Nice, there were already tests for this with a comment in both mentioning that it was currently broken. Well, seems like everything's fixed now. Yeay \o/

weihanglo (approved):
It seems docs.rs does use [`doc.extern-map`](https://github.com/rust-lang/docs.rs/blob/02af83ebc3032ff47dac0ff14400df39f55f46df/src/docbuilder/rustwide_builder.rs#L812), so I believe this is a correct fix. Let's merge this and see if it works :)

GuillaumeGomez:
Yes it's primarily meant for docs.rs. Sorry, should have precised that. Thanks for the review!

PR #13490, opened 2024-02-26
fix(doctest): search native libs in build script outputs

weihanglo:

`rustdoc --test` not only compiles but executes doctests.
Ideally only execution phase should have search paths appended,
so the executions can find native libs just like other tests.
However, there is no way to separate these two phase, so this
hack is added for both phases.

There are some concerns for this fix but is pretty rare. 
* rustdoc --test does two things: compile and run tests. One proc-macro may load wrong shared libs when compiling doctest.
* Dynamic linkers may load dylibs built for a different platform and fail.

However, for cargo doc proc-macro already has bugs when loading shared libs, so this use case is perhaps unlikely to happen. The Cargo team had a discussion and decided to proceed to a fix before tools are able to separate build/run phases.

### Additional information

Fixes <https://github.com/rust-lang/cargo/issues/8531>.

See <https://github.com/rust-lang/cargo/pull/10469#issuecomment-1079503079> and [this Zulip topic](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/dynamic.20linkers.20.26.20mixed.20targets). 

Also prior pull requests <https://github.com/rust-lang/cargo/pull/12710> and <https://github.com/rust-lang/cargo/pull/10469>.

### DIFF
--- a/src/cargo/core/compiler/compilation.rs
+++ b/src/cargo/core/compiler/compilation.rs
@@ -293,6 +293,18 @@ impl<'gctx> Compilation<'gctx> {
     ) -> CargoResult<ProcessBuilder> {
         let mut search_path = Vec::new();
         if tool_kind.is_rustc_tool() {
+            if matches!(tool_kind, ToolKind::Rustdoc) {
+                // HACK: `rustdoc --test` not only compiles but executes doctests.
+                // Ideally only execution phase should have search paths appended,
+                // so the executions can find native libs just like other tests.
+                // However, there is no way to separate these two phase, so this
+                // hack is added for both phases.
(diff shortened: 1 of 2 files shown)


weihanglo:
Blockd on <https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/gix-ref.20CI.20error> and <https://github.com/Byron/gitoxide/issues/1309>.

epage (approved):
r= me when CI is ready

weihanglo:
Thanks!

bors try shows it passed, so merge.

@bors r=epage

Issue #9931, opened 2021-09-21
Don't do a full scan on first build


ehuss:
During the first build with a build script (or running `cargo doc`) cargo does a full scan of the package to determine the file with the newest mtime. This full scan can be expensive and can cause some issues. For example, in #9545, there is no way to tell cargo to avoid scanning certain paths.  This is also a lot of needless work, since for the purpose of determining if something needs to rebuild, during the first build it should know for certain that it needs to be built.

This is difficult to fix because the fingerprinting code is fairly tricky and subtle.  One idea I had was to add a new `FirstRun` variant to [`LocalFingerprint`](https://github.com/rust-lang/cargo/blob/e79ba78087c000be59fea2c56f00bd4aa3893301/src/cargo/core/compiler/fingerprint.rs#L643) that will be used if this is a build script that has never been run.  The `LocalFingerprint` should then get adjusted to the correct value [here](https://github.com/rust-lang/cargo/blob/e79ba78087c000be59fea2c56f00bd4aa3893301/src/cargo/core/compiler/fingerprint.rs#L459-L461).  However, I vaguely recall trying this and running into problems.

Fixing this for `cargo doc` isn't feasible unless `rustdoc` adds the ability to emit a dep-info file somehow.

weihanglo:
The `cargo doc` part of this full scan issue is expected to get fixed via <https://github.com/rust-lang/cargo/pull/15359>. It is now tracked in <https://github.com/rust-lang/cargo/issues/15370>.

epage:
By the tags, I'm assuming this is tied to build scripts.  If `rerun-if-changed` could be exhaustively specified statically in `Cargo.toml` (#6689), would those packages doing so be able to drop this full scan?

epage:
Is the problem detecting a first run or the ambiguity between first run and legacy mode? With #4587, we'd need to start tracking legacy mode which I wonder if that would allow us to fix this. I also wonder if this would address #6779.

Issue #4587, opened 2017-10-06
Rerun-if-changed without disabling other heuristics as a side-effect


pornel:
Currently any use rerun-if-changed opts in entire build into a completely "manual" tracking, disabling all built-in rules as a side effect. This means it's not possible to safely just add one file/env var check to the build without understanding all the dependencies of the entire build script.

In order to use rerun-if-changed features in build-time libraries without potentially breaking rest of the build, it's necessary to be able to use rerun-if-changed-like functionality without disabling the built-in heuristic (i.e. rerun if a var/file has changed *in addition* to all the other/default conditions that cause the build to be rebuilt).

https://github.com/alexcrichton/pkg-config-rs/pull/45#issuecomment-334552552

bikeshed syntax: `cargo:rerun-if-also-changed=file`, `cargo:rerun-if-changed+=file`, `cargo:rerun-if-also-changed;keep-defaults=file`

lukaslueg:
The current semantics of `rerun-if-changed` and `rerun-if-changed-env` are confusing and a new key `rerun-if-changed+` might just add to this confusion. Notice that

> rerun-if-changed+=foo
> rerun-if-changed+=bar
> rerun-if-changed=foobar

is effectively the same as without the `+`-key.

> rerun-if-changed+=foo
> rerun-if-changed+=bar
> rerun-if-change-env=CC

causes Cargo to still ignore the package sources due to a reference to an env-var and use `foo, bar, [CC]` as the only fingerprints.

IMHO `rerun-if-changed+` in addition to the current system is inferior to changing `rerun-if-changed` and `rerun-if-changed-env` to add fingerprints to the then already-considered package sources.

The current implementation calls for the build-script to `walkdir()` it's sources, including those that might be out-of-tree. The rationale is that the user has already opted in to do dependency selection himself and a `walkdir('./src')` is easily implemented.
Another approach might be to selectively opt-out of package-sources and let `rerun-if-changed` work as `rerun-if-changed+` would. The user can then escalate an "also consider this" to "just consider this".

pornel:
I'm also OK with changing the existing behavior to be additive, and adding another explicit option like `cargo:disable-built-in-heuristics`.

kornelski:
Can disabling of built-in heuristics as a side effect be fixed (i.e. side-effect removed)?

Does building more than necessary count as a breaking change? If so, can it be done as part of 2018 edition?

alexcrichton:
[...]

PR #13498, opened 2024-02-27
fix: reduce failing to open a git repo to a tracing warning

### LINKED ISSUE #10150: Cargo build crash in project with git Split Index
### Problem

Cargo build is breaking when a build.rs file is present in a project with a git repo using [split indexes](https://github.com/git/git/blob/b46013950aff31b6626af96ccbf2c48469e36c66/Documentation/git-update-index.txt#L393-L416).

```
$ cargo build

error: failed to determine package fingerprint for build script for split_index_bug v0.1.0 (/.../split_index_bug)

Caused by:
  failed to determine the most recently modified file in /.../split_index_bug

Caused by:
  failed to determine list of files in /.../split_index_bug

Caused by:
  failed to open git index at /.../split_index_bug/.git/

Caused by:
  unsupported mandatory extension: 'link'; class=Index (10)
 ```

### Steps to Reproduce

I have made a small repo where you can reproduce the issue https://github.com/Samangan/cargo-split-index-bug

### Version

```text
cargo 1.56.0 (4ed5d137b 2021-10-04)
release: 1.56.0
commit-hash: 4ed5d137baff5eccf1bae5a7b2ae4b57efad4a7d
commit-date: 2021-10-04
```

arlosi:
When listing files in a path source, Cargo attempts to open the path source as a git repository with a fallback to walking the filesystem if path source isn't a git repository or fails for various reasons.

If the path source *is* part of a git repository, but that git repository uses features that `cargo` (via `libgit2`) doesn't understand, Cargo will error out. This PR changes the error to tracing warning, similar to other failures on this path.

Fixes #10150

Manually validated on the repro repo linked in the issue.

### DIFF
--- a/src/cargo/sources/path.rs
+++ b/src/cargo/sources/path.rs
@@ -217,9 +217,13 @@ impl<'gctx> PathSource<'gctx> {
                 return Ok(None);
             }
         };
-        let index = repo
-            .index()
-            .with_context(|| format!("failed to open git index at {}", repo.path().display()))?;
+        let index = match repo.index() {
+            Ok(index) => index,
+            Err(e) => {
+                tracing::warn!("failed to open git index at {}: {e}", repo.path().display());
+                return Ok(None);
+            }
+        };
         let repo_root = repo.workdir().ok_or_else(|| {
             anyhow::format_err!(
                 "did not expect repo at {} to be bare",


weihanglo (commented):
Would there be a possilbility that

* people accidentally set split-index, and then
* `cargo publish` went with heuristic file inclusion rule instead of git tree, and
* unexpected files got uploaded?

weihanglo:
I might be overthinking though

arlosi:
> Would there be a possilbility that
> 
> * people accidentally set split-index, and then
> * `cargo publish` went with heuristic file inclusion rule instead of git tree, and
> * unexpected files got uploaded?

This is a possibility, but I'd expect it to be rare.

The current situation is that it's not possible to *build* a crate with a `build.rs` that's contained anywhere in a git repo that uses an unsupported git feature. For example, if I vendor a dependency in a repo that's using split index, then the dependency can no longer build in the checked-in vendor directory.

One alternative fix would be to only attempt to open the git repo if we detect that the `Cargo.toml` is next to the `.git` directory. Based on this doc comment that would be correct.
```
    /// Returns `Some(git2::Repository)` if found sibling `Cargo.toml` and `.git`
    /// directory; otherwise, caller should fall back on full file list.
```

The comment may be out of date based on the implementation though, which appears to support `Cargo.toml` nested within the git repo.

weihanglo:
The doc explicitly states that [Cargo respects gitignore files when packaging](https://doc.rust-lang.org/nightly/cargo/reference/manifest.html#the-exclude-and-include-fields)

> If `include` is not specified, then the following files will be excluded:
> 
> * If the package is not in a git repository, all “hidden” files starting with a dot will be skipped.
> * If the package is in a git repository, any files that are ignored by the [gitignore](https://git-scm.com/docs/gitignore) rules of the repository and global git configuration will be skipped.

I have no idea how strong Cargo follows those rules. Some people do [rely on the heuristics](https://github.com/rust-lang/cargo/issues/4587).

An alternative might be <https://github.com/rust-lang/cargo/issues/9931>, which disables the heuristics and sets `rerun-if-changed=build.rs` by default on an edition boundary. We do have [this item](https://github.com/orgs/rust-lang/projects/30?pane=issue&itemId=20027376) in Cargo editio 2024 planning 😆.

Should we do a quick poll and see if people accept this seemly low risk?

PR #13500, opened 2024-02-28
[beta-1.77] Backport ignore `cargo::rustc-check-cfg`

ehuss:
This is a beta backport of https://github.com/rust-lang/cargo/pull/13438#event-11953691715 to disable the warning when `cargo::rustc-check-cfg` is used in a build script on stable in order to provide a broader window where we can stabilize check-cfg, and allow developers to avoid the warning when they use a range of different Rust versions (such as stable and nightly).

Also backports to fix CI:
* https://github.com/rust-lang/cargo/pull/13455 — Remove unnecessary use statement in metabuild


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

Issue #13397, opened 2024-02-05
Want cargo publish --idempotent


ijackson:
### Problem

Sometimes a cargo publish run of multiple packages will fail.  It is then necessary to restart it.

It would be nice if we could run the whole publication from the top, and simply skip crates that were already published.  There are probably other uses.

### Proposed Solution

`cargo publish --idempotent`

If the current version of the crate has already been published, skips the crate and exits with a success status.

(Ideally it would download the crate and check it's identical to what is proposed to publish now, but I think that crates.io does unpredictable-to-cargo mangling of the uploaded data which makes such a check infeasible.)

ijackson:
> Could you clarify the use case? Did you mean one `cargo publish` invocation to publish multiple packages? `cargo publish` doesn't support that at the moment. Workspace publish is a feature tracked [here](https://github.com/rust-lang/cargo/issues/1169).

Yes.  Because cargo can't publish a workspace in one go, we all have add-on tools of varying sophistication to publish our multi-crate workspaces.  The add-on tools iterate over the crates in a workspace, calling cargo publish for each one.

I think #1169 looks quite hard.  It has been open for a very long time and although there's work on it, a good solution there will be a little while.

One of the ways that these add-on tools are suboptimal is that after a failure of any kind, the workspace can be left "half-published".  Recovering from such a situation involves either complicated predictive logic in the tool, or manual tweaking of intermediate publication lists, etc.

So in the meantime, while #1169 is worked on, our add-on tools would be simpler and more reliable of cargo offered an idempotent publish operation.

And, I think any reasonable solution to #1169 needs to cope with failures halfway through, and for that situation to be recoverable.  That would mean that to implement #1169, cargo needs an idempotent publish operation for its own use anyway.  Ie, the underlying functionality in my request here is a necessary *part* of #1169.

But it is also separately useful, so it's worth exposing as an operation to external callers.  The notion is fairly well-defined and, I think, not that hard to implement.

> > If the current version of the crate has already been published, skips the crate and exits with a success status.
> 
> There is a similar request #3662. I wonder if we should keep that and close this one.

#3662 requests early *failure*.  I'm requesting treating "this one is already published" as a *success*.  Similar underlying logic might be involved with both, but the overall semantics are rather different.

[...]

PR #13501, opened 2024-02-29
Bail publish job before packaging and upload

### LINKED ISSUE #3662: Fail fast when package version is already published

stupendoussuperpowers:
Fixes #3662 

This PR adds a check for existing version in the registry before we package and upload the crate.

With this PR we get an error pointing out that the same version exists in registry, and we abort the publish job.

Also adds tests for the same.

### DIFF
--- a/src/cargo/ops/registry/publish.rs
+++ b/src/cargo/ops/registry/publish.rs
@@ -28,6 +28,7 @@ use crate::ops;
 use crate::ops::PackageOpts;
 use crate::ops::Packages;
 use crate::sources::source::QueryKind;
+use crate::sources::source::Source;
 use crate::sources::SourceConfigMap;
 use crate::sources::CRATES_IO_REGISTRY;
 use crate::util::auth;
@@ -129,7 +130,7 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
         val => val,
     };
     let source_ids = super::get_source_id(opts.gctx, reg_or_index.as_ref())?;
-    let mut registry = super::registry(
+    let (mut registry, mut source) = super::registry(
         opts.gctx,
         &source_ids,
         opts.token.as_ref().map(Secret::as_deref),
@@ -138,6 +139,34 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
         Some(operation).filter(|_| !opts.dry_run),
     )?;
     verify_dependencies(pkg, &registry, source_ids.original)?;
+
+    // Bail before packaging and uploading if same version already exists in the registry
+
+    let query = Dependency::parse(pkg.name(), Some(&ver), source_ids.replacement)?;
+
+    let _lock = opts
+        .gctx
+        .acquire_package_cache_lock(CacheLockMode::DownloadExclusive)?;
+
+    let duplicate_query = loop {
+        match source.query_vec(&query, QueryKind::Exact) {
+            std::task::Poll::Ready(res) => {
+                break res?;
+            }
+            std::task::Poll::Pending => source.block_until_ready()?,
+        }
+    };
+
+    drop(_lock);
+
+    if !duplicate_query.is_empty() {
+        bail!(
+            "{}@{} already exists on {}",
+            pkg.name(),
+            pkg.version(),
+            source.describe()
+        );
+    }
 
     // Prepare a tarball, with a non-suppressible warning if metadata
     // is missing since this is being put online.
@@ -411,7 +440,7 @@ pub(crate) fn prepare_transmit(
         }
     }
 
-    let string_features = match manifest.resolved_toml().features() {
+    let string_features = match manifest.normalized_toml().features() {
         Some(features) => features
             .iter()
             .map(|(feat, values)| {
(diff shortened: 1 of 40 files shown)


stupendoussuperpowers:
Some tests with alt_registry are failing. 

When we query the index for duplicate version, it prints out an extra "Updating [..] index" line. Which is missing in those test cases. 

Should I be modifying the failing tests to add that line, or is there a different way to run the query where this log line is not outputted?

weihanglo:
If the fix does change the behaviour of those test cases, then assertions of printing status seems reasonable to me

src/cargo/ops/registry/publish.rs
```diff
@@ -136,6 +136,35 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
...
+            }
+            std::task::Poll::Pending => source.block_until_ready()?,
+        }
+    };
+
+    drop(_lock);
+
+    if !duplicate_query.is_empty() {
+        bail!(
+            "crate {} already has version {}. Aborting publish.",
```
epage, line 162: nit:
`"crate {name}@{version} is already present in {registry}"`
*(have not looked into what "registery" variable would be appropriate to lit here)*

- Specifying the registry might be helpful in case someone is publishing to multiple as that gives them a chance to think "Oh whoops, I meant this other one"
- We likely don't need an "Aborting publish" as thats what an error is
- The current phrasing is a little awkward (saying a crate has a version reads a little funny to me)

epage:
> When we query the index for duplicate version, it prints out an extra "Updating [..] index" line. Which is missing in those test cases.
>
> Should I be modifying the failing tests to add that line, or is there a different way to run the query where this log line is not outputted?

Not looked into the details of when this message will show up or not but we likely don't need a new instance of `SourceConfigMap` and if we can reuse the way to look up registeries with other code, that'd likely help.

wait-for-publish has to go through some extra hoops because it needs to avoid reusing cached results.

stupendoussuperpowers:
The easy fix I found was to use `source.set_quiet(true)` - tests are passing with this.

I am also not able to find any other `source` present in this flow that we can re-use.

`start_fetch()` is supposed to be called only once per `HttpRemote` instance, I am unsure why creating a new source here is causing it to create a different instance for alternate registries but not for crates-io.

The first update in both cases happens when we create the registry object here - [publish.rs#L130](https://github.com/rust-lang/cargo/blob/6994f624109039de1679d6b0a7118218581eec9c/src/cargo/ops/registry/publish.rs#L130)

For alternate registry upon querying it causes `start_fetch()` to be triggered again.

Any insight into how alternate registry queries work would be helpful. For now we can also continue with just adding `set_quiet(true)`

src/cargo/ops/registry/publish.rs
stupendoussuperpowers: registry variable here can be `source.describe()` which would be "crates.io index" in this case.

stupendoussuperpowers:
Hey, did anyone get a chance to check this?

In summary - 
1. We can use `source.set_quiet(true)` which makes the alternate publish pass
2. I'll need some guidance into why there's different outputs for crates-io and an alternate registry

weihanglo:
> I'll need some guidance into why there's different outputs for crates-io and an alternate registry

Could you show us the difference? Let us know when this PR is ready for review :)

---

Somehow I remembered this issue <https://github.com/rust-lang/cargo/issues/13397>. While this PR just make it fail fast, it might be something to consider in the future.

stupendoussuperpowers:
> start_fetch() is supposed to be called only once per HttpRemote instance, I am unsure why creating a new source here is causing it to create a different instance for alternate registries but not for crates-io.
> 
> The first update in both cases happens when we create the registry object here - [publish.rs#L130](https://github.com/rust-lang/cargo/blob/6994f624109039de1679d6b0a7118218581eec9c/src/cargo/ops/registry/publish.rs#L130)
> 
> For alternate registry upon querying it causes start_fetch() to be triggered again.
> 
> Any insight into how alternate registry queries work would be helpful. For now we can also continue with just adding set_quiet(true)

The above comment is the depth to which I was able to analyze the issue. 

The symptom for this is - 

When I added the code to check for existing version of the package, it logged an extra `"Updating [..] index"` line IF used with an alternate registry. It did not do this with crates-io. This causes certain alternate registry tests to fail since the output doesn't expect `"Updating [..] index"` to be present twice.

torhovland:
The multiple Updating messages do make sense in a way. The first one is notifying about fetching `/index/config.json`, and the second one about `/index/3/f/foo`. If these fetches were running in the same `RegistrySource` there would only be one message, but they aren't. One is made within `super::registry()` and one is made just after in this PR in `let mut source = SourceConfigMap::empty(opts.gctx)?.load(reg_ids.original, &HashSet::new())?;`.

I tried reusing the first one, but that broke the `publish::duplicate_version` test.

torhovland:
But that was fixed by changing

```
    let query = Dependency::parse(pkg.name(), Some(&ver), reg_ids.original)?;
```

to

```
    let query = Dependency::parse(pkg.name(), Some(&ver), reg_ids.replacement)?;
```

Here's a full fix:

[diff.txt](https://github.com/user-attachments/files/15988323/diff.txt)

@stupendoussuperpowers Will you try to apply this?

stupendoussuperpowers:
thanks @torhovland for the help, let me try and apply this patch. will submit a patched PR.

torhovland:
Hi! I already applied the fix and submitted it in #14338, but it turns out there's yet another test that demands some work. You can certainly look into that if you want.

torhovland:
Superseded by #14338.

PR #13499, opened 2024-02-28
fix(rustc): Always pass --edition to rustc

epage:
On [Internals](https://internals.rust-lang.org/t/idea-rustc-cargo-should-warn-on-unspecified-edition/20309), the idea came up for warning on unset Edition.
I am working on the cargo warning but if rustc ever wants to do so, they'd be blocked on cargo ensuring `--edition` is always set. Hence this change.

Issue #10954, opened 2022-08-09
Manifests with a resolver field will be published even if the MSRV is less than 1.51


taiki-e:
### Problem

On the recent nightly, it seems that cargo package and cargo publish always generate a resolver field, regardless of MSRV or edition.

This makes it impossible to compile on rustc versions such as 1.50, where the resolver field was unstable, even if the MSRV of the published crate is smaller than 1.51.

```
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/async-channel-1.7.0/Cargo.toml`

Caused by:
  feature `resolver` is required

  this Cargo does not support nightly features, but if you
  switch to nightly channel you can add
  `cargo-features = ["resolver"]` to enable this feature
```

Probably related to https://github.com/rust-lang/cargo/pull/10911
cc @Nemo157

### Steps

To reproduce:
```console
$ cargo new --lib repro 
$ cd repro
$ sed -i 's/edition = "2021"/edition = "2018"/' Cargo.toml
$ cargo package --allow-dirty
$ cat target/package/repro-0.1.0/Cargo.toml | grep resolver
resolver = "1"
```

### Version

```text
cargo 1.65.0-nightly (4fd148c47 2022-08-03)
release: 1.65.0-nightly
commit-hash: 4fd148c47e733770c537efac5220744945d572ef
commit-date: 2022-08-03
host: aarch64-apple-darwin
libgit2: 1.4.2 (sys:0.14.2 vendored)
libcurl: 7.79.1 (sys:0.4.55+curl-7.83.1 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 12.5.0 [64-bit]
```

epage:
@Nemo157 would you be willing to follow up on this as its a bug in #10911 (that we all overlooked)?

Our thought is to skip `resolve = "1"` if that is what will be inferred anyways.

Nemo157:
Yep, I think I still have the commit that did that in the reflog somewhere, I'll try and find it.

PR #13505, opened 2024-02-29
feat(toml): Warn on unset Edition

epage:

On [Internals](https://internals.rust-lang.org/t/idea-rustc-cargo-should-warn-on-unspecified-edition/20309), the idea came up for warning on unset Edition.

Besides helping people who forget to set the Edition, this creates symmetry between `Cargo.toml` and cargo scripts (#12207).  While the default is different in each case, we are making the default obvious and guiding people away from it.

There are separate commits for adding tests (and refactors) so the changes in behavior will be more obvious

### Additional information

This builds on
- #13499
- #13504

### DIFF
--- a/tests/testsuite/edition.rs
+++ b/tests/testsuite/edition.rs
@@ -124,6 +124,33 @@ fn edition_unstable() {
 }
 
 #[cargo_test]
+fn unset_edition_works_with_no_newer_compatible_edition() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = 'foo'
+                version = '0.1.0'
+                rust-version = "1.0"
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("check -v")
+        .with_stderr(
+            "\
+[WARNING] no edition set: defaulting to the 2015 edition
+[CHECKING] foo [..]
+[RUNNING] `rustc [..] --edition=2015 [..]`
+[FINISHED] [..]
+",
+        )
+        .run();
+}
+
+#[cargo_test]
 fn unset_edition_works_on_old_msrv() {
     let p = project()
         .file(
@@ -141,7 +168,7 @@ fn unset_edition_works_on_old_msrv() {
     p.cargo("check -v")
         .with_stderr(
             "\
-[WARNING] no edition set: defaulting to the 2015 edition while the latest is 2021
+[WARNING] no edition set: defaulting to the 2015 edition while 2018 is compatible with `rust-version`
 [CHECKING] foo [..]
 [RUNNING] `rustc [..] --edition=2015 [..]`
 [FINISHED] [..]
--- a/tests/testsuite/edition.rs
+++ b/tests/testsuite/edition.rs
@@ -122,3 +122,29 @@ fn edition_unstable() {
         )
         .run();
 }
+
+#[cargo_test]
+fn unset_edition_works_on_old_msrv() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = 'foo'
+                version = '0.1.0'
+                rust-version = "1.50"  # contains 2018 edition
+            "#,
+        )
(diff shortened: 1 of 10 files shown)


weihanglo (approved):

tests/testsuite/edition.rs
```diff
@@ -122,3 +122,57 @@ fn edition_unstable() {
...
+                rust-version = "1.50"  # contains 2018 edition
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("check -v")
+        .with_stderr(
+            "\
+[WARNING] no edition set: defaulting to the 2015 edition while 2018 is compatible with `rust-version`
```
weihanglo, line 171: Should we call out which package it is? There might be a workspace containing tons of members.

epage: Looks like its done for us:
```
                        // In a workspace, it can be confusing where a warning
                        // originated, so include the path.
                        format!("{}: {}", path.display(), warning.message)
```

The other warnings I looked at do not mention the package which is why I dug into it to find that code.  The test case must be only for a single package so it gets elided.

weihanglo:
Given edition 2015 is quite old, I assume most people won't be flooded with this error messages. If it turns out to be a real problem, we can consider revert.

@bors r+

taiki-e:
Is it possible to silence this warning to the workspace contains a crate that have rust-version < 1.31?

I don't think this warning is useful at all for crates with MSRV lower than 1.31, as adding the edition will break the MSRV.

weihanglo:
> I don't think this warning is useful at all for crates with MSRV lower than 1.31, as adding the edition will break the MSRV.

That sounds reasonable. However the `edition` field predates `rust-version`. How can Cargo determine MSRV when there is no `rust-version` set?

taiki-e:
> However the `edition` field predates `rust-version`. How can Cargo determine MSRV when there is no `rust-version` set?

rust-version field is just ignored on old rustc. This means a crate with old MSRV can have rust-version field. For example: https://github.com/serde-rs/serde/blob/v1.0.179/serde/Cargo.toml#L14

weihanglo:
> rust-version field is just ignored on old cargo.

Isn't this the same as setting `edition = "2015"` for those package on old caroges? Also, does it make sense that a package supports old MSRV than its edition settings? For example the author only uses Cargo-specific edition features but not Rust compiler's.


Yet I understand that Cargo should be smarter to infer `edition` for those on old MSRV. Need to think more on it.

taiki-e:
> Isn't this the same as setting `edition = "2015"` for those package on old caroges?

`edition = "2015"` causes an error on some old versions (e.g., 1.27). See also an example in "Caveats" section in https://github.com/rust-lang/rust-clippy/issues/6324.

```console
$ cat Cargo.toml 
[package]
name = "repro"
version = "0.1.0"
edition = "2015"

$ cargo +1.27 build
error: failed to parse manifest at `/Users/taiki/projects/tmp/repro/Cargo.toml`

Caused by:
  editions are unstable

Caused by:
  feature `edition` is required

this Cargo does not support nightly features, but if you
switch to nightly channel you can add
`cargo-features = ["edition"]` to enable this feature
```

Most previously unstable fields have similar properties (e.g., https://github.com/rust-lang/cargo/issues/10954), but the `rust-version` field is special and is implemented in a way that it does not cause such errors in any version.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

> The first version of Cargo that supports this field was released with Rust 1.56.0. In older releases, the field will be ignored, and Cargo will display a warning.

---

> does it make sense that a package supports old MSRV than its edition settings?

(Assuming you are talking about a crate like: MSRV is 1.20 but the edition is 2018.)
I don't think it makes sense because the only time it doesn't cause an error is when building with a very old cargo (pre-1.27), and it is just ignored.

Also, specifying MSRV older than edition in rust-version causes an error.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

> The rust-version must be equal to or newer than the version that first introduced the configured edition.

Issue #5805, opened 2018-07-26
Vendored crates should be checked for changes


glandium:
I was building Firefox, and tried to do some changes to some rust crate. Doing so made the build fail with something like
```
error: the listed checksum of ... has changed:
expected: ...
actual: ...

directory sources are not intended to be edited, if modifications are required then it is recommended that [replace] is used with a forked copy of the source
```

That's an IMO annoying feature of cargo, but this is not what this issue is about. The problem is the following: if I actually build the crate first, and **then** make the changes, rebuilding just goes on acting as if I had made no change, saying `Fresh $crate_name`.

It seems to me cargo should fail with the exact same error as if I hadn't already built.

Cc: @alexcrichton

alexcrichton:
This is currently expected behavior in that Cargo treats vendored crates like it does crates.io, it doesn't check any mtimes as it assumes it's all read-only. It is a performance optimization to not even look at the filesystem for these crates on incremental builds.

The correct way to modify a vendored crate is to use `[patch]` or a `path` dependency, and Cargo will then correctly handle the crate on incremental rebuilds as it knows that it's not a readonly dependency.

ctrlcctrlv:
> The correct way to modify a vendored crate is to use [patch] or a path dependency, and Cargo will then correctly handle the crate on incremental rebuilds as it knows that it's not a readonly dependency.

Fortunately GNU Guix can bypass this pointless check, as I today discovered.

https://fredrickbrennan.medium.com/how-to-bypass-rust-cargo-error-directory-sources-are-not-intended-to-be-edited-ab2f257b89da

k3d3:
When I vendor code into my repository, it changes line endings (and nothing else) which triggers this check. Is there any way to disable or at least mitigate the error in this situation?

ctrlcctrlv:
Does revendor.scm not work for you?

epage:
As this is expected behavior and is important for performance reasons, I'm going to propose to the cargo team that we close this.

weihanglo:
Instead of closing this now, I would like to see an improvement in the [cargo-vendor](https://doc.rust-lang.org/nightly/cargo/commands/cargo-vendor.html) doc. Could document that vendored sources are not intended to be modified, and `[patch]` is the correct way to do it.

briansmith:
[...]

PR #13512, opened 2024-03-01
[docs]: Clarify vendored sources as read-only and way to modify

LuuuXXX:
What does this PR try to resolve?

To document that vendored sources are not intended to be modified, and `[patch]` or a `path` dependency is the correct way to modify a vendored crate.

https://github.com/rust-lang/cargo/issues/5805#issuecomment-1868049928

### DIFF
--- a/src/doc/man/cargo-vendor.md
+++ b/src/doc/man/cargo-vendor.md
@@ -21,6 +21,12 @@ stdout after `cargo vendor` completes the vendoring process.
 You will need to add or redirect it to your Cargo configuration file,
 which is usually `.cargo/config.toml` locally for the current package.
 
+Cargo treats vendored sources as read-only as it does to registry and git sources.
+If you intend to modify a crate from a remote source,
+use `[patch]` or a `path` dependency pointing to a local copy of that crate.
+Cargo will then correctly handle the crate on incremental rebuilds,
+as it knowns that it is no longer a read-only dependency.
+
 ## OPTIONS
 
 ### Vendor Options
--- a/src/doc/man/generated_txt/cargo-vendor.txt
+++ b/src/doc/man/generated_txt/cargo-vendor.txt
@@ -18,6 +18,12 @@ DESCRIPTION
        need to add or redirect it to your Cargo configuration file, which is
        usually .cargo/config.toml locally for the current package.
 
+       Cargo treats vendored sources as read-only as it does to registry and
+       git sources. If you intend to modify a crate from a remote source, use
+       [patch] or a path dependency pointing to a local copy of that crate.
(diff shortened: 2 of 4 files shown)


src/doc/man/cargo-vendor.md
```diff
@@ -21,6 +21,11 @@ stdout after `cargo vendor` completes the vendoring process.
 You will need to add or redirect it to your Cargo configuration file,
 which is usually `.cargo/config.toml` locally for the current package.
 
+By default, Cargo treats vendored crates as read-only as it does crates.io.
```
weihanglo, line 24: “By default” sounds like there is a way to tweak it, though there is no way to change the read-only property.

```suggestion
Cargo treats vendored sources as read-only as it does to registry and git sources.
```

src/doc/man/cargo-vendor.md
```diff
@@ -21,6 +21,11 @@ stdout after `cargo vendor` completes the vendoring process.
...
+To modify a vendored crate the correct way is to use `[patch]` or a `path` dependency,
+and Cargo will then correctly handle the crate on incremental rebuilds as it knowns
+that it is not a read-only dependency.
```
weihanglo, line 27: A few tweaks. What do you think?

```suggestion
If you intend to modify a crate from a remote source,
use `[patch]` or a `path` dependency pointing to a local copy of that crate.
Cargo will then correctly handle the crate on incremental rebuilds,
as it knowns that it is no longer a read-only dependency.
```

weihanglo (commented):
Thanks! Added some suggestions :)

src/doc/man/cargo-vendor.md
LuuuXXX: Thanks, this looks more accurate.

src/doc/man/cargo-vendor.md
```diff
@@ -21,6 +21,12 @@ stdout after `cargo vendor` completes the vendoring process.
 You will need to add or redirect it to your Cargo configuration file,
 which is usually `.cargo/config.toml` locally for the current package.
 
+Cargo treats vendored sources as read-only as it does to registry and git sources.
```
weihanglo, line 24: Just notice there are some trailing spaces. Not sure if it is intentionally done but they do render to a linebreak here. Otherwise looks good to me.

```suggestion
Cargo treats vendored sources as read-only as it does to registry and git sources.
```

weihanglo (approved):

src/doc/man/cargo-vendor.md
LuuuXXX: More thanks, modified

weihanglo:
@bors r+ 

Thanks for the polish!

PR #13514, opened 2024-03-01
fix(msrv): Report all incompatible packages, not just a random one

epage:

In #9930, it recommended improving the error for incompatible packages
so people can better get to the root of the problem.
For example, you might get an error about `clap_lex` but resolving the
error for the higher level `clap` could make the problem with `clap_lex`
go away.

Because I generally saw earlier packages in the graph reported, I
assumed we were reporting these errors bottom up.
It turns out, we are reporting them in the `UnitGraph`s order, which is
non-deterministic because it is built on a `HashMap`.

So this adds determinism and shows all incompatible dependencies
(not just the bottom or the root).

This is a first step.
We might find that we still want to only include the shallowest units,
rather than all.
At that point, we can add the complexity to address this by walking the
unit graph.

We could also further improve this by querying the index to suggest
compatible versions of packages.

The multi-package test wasn't split out into its own commit because hat would have required fixing the non-determinism one way to then just revert it when we fix it this way.

### DIFF
--- a/tests/testsuite/rust_version.rs
+++ b/tests/testsuite/rust_version.rs
@@ -180,7 +180,8 @@ fn rust_version_too_high() {
         .with_status(101)
         .with_stderr(
             "\
-[ERROR] package `foo v0.0.1 ([..])` cannot be built because it requires rustc 1.9876.0 or newer, while the currently active rustc version is [..]
+[ERROR] rustc [..] is not supported by the following package:
+  foo@0.0.1 requires rustc 1.9876.0
 
 ",
         )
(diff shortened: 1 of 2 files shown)


tests/testsuite/rust_version.rs
```diff
@@ -218,10 +219,62 @@ fn dependency_rust_version_newer_than_rustc() {
...
+        )
+        .run();
+    p.cargo("check --ignore-rust-version").run();
+}
+
+#[cargo_test]
+fn dependency_tree_rust_version_newer_than_rustc() {
+    Package::new("baz", "0.0.1")
+        .dep("bar", "0.0.1")
+        .rust_version("1.2345.0")
```
weihanglo, line 238: nit: choose a larger version. I believe Rust will last over 300 years!

weihanglo (approved):

tests/testsuite/rust_version.rs
weihanglo: Just kidding 🤪

weihanglo:
@bors r+

Thanks!

Issue #6908, opened 2019-05-06
Optimizes Cargo's registry cache format for fewer files


alexcrichton:
First implemented in https://github.com/rust-lang/cargo/pull/6880 Cargo now has an on-disk cache for the registry index which avoids loading from git and parsing extraneous JSON which doesn't end up getting used. 

This format isn't fantastic for Windows, however, since it has an extra file-per-crate in the index. @Eh2406 recommended a [different implementation strategy](https://github.com/rust-lang/cargo/pull/6880#issuecomment-488442985) which would reduce the number of files in play, and we should likely implement that!

Eh2406:
A easier way to see how big a difference this makes may be to do a single file cache using sqlite or sled. Then when we know if it works we can have the discussion of if the dependency is worth it, or if we should implement our own format.

weihanglo:
I've benchmark for SQLite and sled in naive implementations and the result seems not better than the current in-house cache mechanism.

Both of the implementations store almost the same format, "package name -> content blob" key-value pairs, inside databases. The only slight difference is that I name the database files as `<cache-version>-<index-version>-<sha>.db` and remove the prefix from each blob contents.

Though the benchmark result seems not that greatI think there may still be some room to improve from SQLite side. For sled I have no idea how to tune it. I would love to do more investigation if needed. 

Any thought or direction are welcome!

---

The patches are listed as below:

- SQLite: weihanglo/cargo@4e277122b
- sled: weihanglo/cargo@789e3f3cb
- master: b1684e284

The benchmark are run against cargo project itself by [hyperfine](https://github.com/sharkdp/hyperfine).

| Command              | master                      | SQLite                      | sled                           |
| ------------------ | -------------------- | -------------------  | -------------------- |
| update                   | 🥇295.0 ms ±   9.9 ms | 307.7 ms ±   8.7 ms | 339.4 ms ±  13.3 ms |
| update -p git2       | 🥇317.6 ms ±  12.7 ms | 329.8 ms ±   9.2 ms | 362.4 ms ±  17.6 ms |
| generate-lockfile  | 🥇304.6 ms ±  11.7 ms | 320.5 ms ±   8.0 ms | 348.5 ms ±  13.7 ms |
| build                      | 🥇132.3 ms ±   3.4 ms |  140.6 ms ±   3.1 ms | 180.8 ms ±   9.1 ms |

<details>
<summary>Raw benchmark data from hyperfine</summary>

```
hyperfine -w 5 -m 20 \
[...]

PR #13515, opened 2024-03-01
refactor: abstract `std::fs` away from on-disk index cache

weihanglo:

Part of https://github.com/rust-lang/cargo/issues/6908

This abstracts `std::fs` so that we can integrate SQLite into on-disk cache in follow-ups.

* Make sure the documentation still render well.
* Shouldn't have any behavior change. Only a bit of extra allocations added.


src/cargo/sources/registry/index/cache.rs
```diff
@@ -212,3 +218,48 @@ impl<'a> SummariesCache<'a> {
         contents
     }
 }
+
+/// Manages the on-disk index caches.
+pub struct CacheManager<'gctx> {
+    /// [`GlobalContext`] reference for convenience.
+    pub gctx: &'gctx GlobalContext,
```
epage, line 228: Feels weird to me that callers access the context through this

weihanglo: Fixed!

epage: Passing in individual unstable features seems a bit higher churn as we evolve but whichever way.

Issue #10653, opened 2022-05-11
`cargo add` takes `package.rust-version` into account when no version-req is specified


epage:
- Implementation: #12078 
- Unstable feature: `-Zmsrv-policy`

---

### Problem

When running `cargo add foo`, we create a version-req out of the latest version in the index.  If this has a higher `rust-version`, then that can break for people.

### Proposed Solution

Constrain the candidate versions in the index by their `rust-version` being lesser or equal to the current crate's `rust-version`.

The main question is what to do when there are no matches due to `rust-version`.
- Add anyways
- Error with newest version's `rust-version`

epage:
When `rust-version` gets added to the registry, we should inform Dependabot people on https://github.com/dependabot/dependabot-core/issues/5423

cassaundra:
Implemented in #12078. Maybe close?

epage:
Implemented but not stable so this works as a place to track stabilization. I should update the issue to reflect that though.

PR #13516, opened 2024-03-01
feat(add): Fallback to `rustc -v` when no MSRV is set

epage:

#10653 made version-requirement selection respect MSRV as part of #9930.
This updates the implementation for the now-approved RFC specifies that we should respect `rustc -v` if there is no MSRV.

The messages also get a little bit of polish.

Tests are added in separate commits for easier viewing of behavior changes.


ehuss:
Thanks!

@bors r+

Issue #12225, opened 2023-06-02
Cargo should strip features of dev dependencies


Morganamilo:
### Problem

https://github.com/rust-lang/cargo/pull/7333 added stripping of dependencies, but did not strip features pointing to those dependencies.

This causes a rather unintuitive error where cargo says the dependency does not exist. Which one would only understand if they knew about the stripping that cargo does.

### Proposed Solution

Cargo should also strip these features from the manifest before publishing.

If `foo` is a dev-dependency being stripping then `a = ["foo/a"]` should be transformed to `a = []`.

eval-exec:
Hello, I submitted a PR to fix this。 Would you kindly review my PR and provide your feedback? :heart_decoration:

linyihai:
@rustbot claim

I'm going to try, and if I haven't moved for a long time, feel free to pick it up and continue.

PR #12230, opened 2023-06-04
Strip features which is owned by dev dependencies only

eval-exec:
### What does this PR try to resolve?
This PR want to close #12225

### How should we test and review this PR?
I added a unit test.

PR #13518, opened 2024-03-02
fix: strip feature dep when dep is dev dep

baby230211:
This change aims to strip features dependencies without a version key to be published. 
If a dev-dependency is missing the version, it will be stripped from the packaged manifest.

The features table may contains the deps in following places.
- Target
  - normal
  - dev
  - build
  - normal-and-dev
- normal
- dev
- build
- normal-and-dev

See the initial commit, it shows current behavior that will cause error when feature has deps that point to dev_dep and doesn't have a version specified.

Title | orignal toml | published toml |
---- | ---- | ---- |
Before |   feature = ["dev-dep/feature"]  <br/> [dev-dependencies] <br/> dev-dep = { .., features: ["feature"] }   |   feature = ["dev-dep/feature"] <br/> [dev-dependencies] <br/> dev-dep = { .., features: ["feature"] } |
After |  feature = ["dev-dep/feature"]  <br/> [dev-dependencies] <br/> dev-dep = { .., features: ["feature"] }  | feature = [] <br/> [dev-dependencies] ```

Fix: #12225

### DIFF
--- a/src/cargo/ops/registry/publish.rs
+++ b/src/cargo/ops/registry/publish.rs
@@ -3,6 +3,7 @@
 //! [1]: https://doc.rust-lang.org/nightly/cargo/reference/registry-web-api.html#publish
 
 use std::collections::BTreeMap;
+use std::collections::BTreeSet;
 use std::collections::HashSet;
 use std::fs::File;
 use std::time::Duration;
@@ -20,6 +21,7 @@ use crate::core::dependency::DepKind;
 use crate::core::manifest::ManifestMetadata;
 use crate::core::resolver::CliFeatures;
 use crate::core::Dependency;
+use crate::core::FeatureValue;
 use crate::core::Package;
 use crate::core::PackageIdSpecQuery;
 use crate::core::SourceId;
@@ -33,6 +35,7 @@ use crate::sources::CRATES_IO_REGISTRY;
 use crate::util::auth;
 use crate::util::cache_lock::CacheLockMode;
 use crate::util::context::JobsConfig;
+use crate::util::interning::InternedString;
 use crate::util::Progress;
 use crate::util::ProgressStyle;
 use crate::CargoResult;
@@ -412,13 +415,31 @@ fn transmit(
         return Ok(());
     }
 
+    let deps_set = deps
+        .iter()
+        .map(|dep| dep.name.clone())
+        .collect::<BTreeSet<String>>();
+
     let string_features = match manifest.original().features() {
         Some(features) => features
             .iter()
             .map(|(feat, values)| {
                 (
                     feat.to_string(),
-                    values.iter().map(|fv| fv.to_string()).collect(),
+                    values
+                        .iter()
+                        .filter(|fv| {
+                            let feature_value = FeatureValue::new(InternedString::new(fv));
+                            match feature_value {
+                                FeatureValue::Dep { dep_name }
+                                | FeatureValue::DepFeature { dep_name, .. } => {
+                                    deps_set.contains(&dep_name.to_string())
+                                }
+                                _ => true,
+                            }
+                        })
+                        .map(|fv| fv.to_string())
+                        .collect(),
                 )
             })
             .collect::<BTreeMap<String, Vec<String>>>(),
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -9,8 +9,8 @@ use crate::AlreadyPrintedError;
 use anyhow::{anyhow, bail, Context as _};
 use cargo_platform::Platform;
 use cargo_util::paths;
-use cargo_util_schemas::manifest;
 use cargo_util_schemas::manifest::RustVersion;
+use cargo_util_schemas::manifest::{self, TomlManifest};
 use itertools::Itertools;
 use lazycell::LazyCell;
 use pathdiff::diff_paths;
@@ -21,7 +21,7 @@ use crate::core::compiler::{CompileKind, CompileTarget};
 use crate::core::dependency::{Artifact, ArtifactTarget, DepKind};
 use crate::core::manifest::{ManifestMetadata, TargetSourcePath, Warnings};
 use crate::core::resolver::ResolveBehavior;
-use crate::core::{find_workspace_root, resolve_relative_path, CliUnstable};
+use crate::core::{find_workspace_root, resolve_relative_path, CliUnstable, FeatureValue};
 use crate::core::{Dependency, Manifest, PackageId, Summary, Target};
 use crate::core::{Edition, EitherManifest, Feature, Features, VirtualManifest, Workspace};
 use crate::core::{GitReference, PackageIdSpec, SourceId, WorkspaceConfig, WorkspaceRootConfig};
@@ -316,7 +316,7 @@ pub fn prepare_for_publish(
         }
     }
     let all = |_d: &manifest::TomlDependency| true;
-    return Ok(manifest::TomlManifest {
+    let mut manifest = manifest::TomlManifest {
         package: Some(package),
         project: None,
         profile: me.profile.clone(),
@@ -366,7 +366,52 @@ pub fn prepare_for_publish(
         badges: me.badges.clone(),
         cargo_features: me.cargo_features.clone(),
         lints: me.lints.clone(),
-    });
+    };
+    strip_features(&mut manifest);
+    return Ok(manifest);
+
+    fn strip_features(manifest: &mut TomlManifest) {
+        fn insert_dep_name(
+            dep_name_set: &mut BTreeSet<manifest::PackageName>,
+            deps: Option<&BTreeMap<manifest::PackageName, manifest::InheritableDependency>>,
+        ) {
+            let Some(deps) = deps else {
+                return;
+            };
+            deps.iter().for_each(|(k, _v)| {
+                dep_name_set.insert(k.clone());
+            });
+        }
+        let mut dep_name_set = BTreeSet::new();
+        insert_dep_name(&mut dep_name_set, manifest.dependencies.as_ref());
+        insert_dep_name(&mut dep_name_set, manifest.dev_dependencies());
+        insert_dep_name(&mut dep_name_set, manifest.build_dependencies());
+        if let Some(target_map) = manifest.target.as_ref() {
+            target_map.iter().for_each(|(_k, v)| {
+                insert_dep_name(&mut dep_name_set, v.dependencies.as_ref());
+                insert_dep_name(&mut dep_name_set, v.dev_dependencies());
+                insert_dep_name(&mut dep_name_set, v.build_dependencies());
+            });
+        }
+        let features = manifest.features.as_mut();
+
+        let Some(features) = features else {
+            return;
+        };
+
+        features.values_mut().for_each(|feature_deps| {
+            feature_deps.retain(|feature_dep| {
+                let feature_value = FeatureValue::new(InternedString::new(feature_dep));
+                match feature_value {
+                    FeatureValue::Dep { dep_name } | FeatureValue::DepFeature { dep_name, .. } => {
+                        let k = &manifest::PackageName::new(dep_name.to_string()).unwrap();
+                        dep_name_set.contains(k)
+                    }
+                    _ => true,
+                }
+            });
+        });
+    }
 
     fn map_deps(
         gctx: &GlobalContext,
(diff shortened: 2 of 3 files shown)


weihanglo:
Thanks for the contribution!

Not sure if @linyihai has started on this, though I believe they are willing to help :)
One reminder is that even this is not ready for review, it would be lovely to have a PR description for reviewers to understand what's going on.

Since this PR is marked as draft, so assuming you have some questions. Feel free to ask anything here or on Zulip.

weihanglo:
Oh you've just updated the PR description 👍🏾

src/cargo/util/toml/mod.rs
```diff
@@ -368,6 +368,42 @@ pub fn prepare_for_publish(
...
+                    .filter(|feature_dep| {
+                        if let Some(feature_pkg_name) = feature_dep.split('/').next() {
+                            let Some(deps) = deps else { return false };
+                            let k =
+                                &manifest::PackageName::new(feature_pkg_name.to_string()).unwrap();
+                            match deps.get(k) {
+                                Some(manifest::InheritableDependency::Value(def)) => {
+                                    def.is_version_specified()
+                                }
+                                _ => false,
```
weihanglo, line 392: Here we've got the same symptom mentioned in this comment <https://github.com/rust-lang/cargo/pull/12230#issuecomment-1575658799>:

* [Platform specific dependencies](https://doc.rust-lang.org/nightly/cargo/reference/specifying-dependencies.html#platform-specific-dependencies) (`target.<…>.dev-dependencies`) should be handled.
* A dependency might appear in both `[dev-dependencies]` other kind of dependency table, and we don't want to remove those appear in `[dependencies]` or [`build-dependencies`] because they are essential.

A good way to ensure that we don't introduce any bug is to write some tests to verify and record the current “incorrect” behavior in the first commit. Then the following commits fix both the issue and tests. This also help reviewer to follow the workflow.

src/cargo/util/toml/mod.rs
```diff
@@ -334,7 +334,7 @@ pub fn prepare_for_publish(
         dev_dependencies2: None,
```
weihanglo, line 334: W8 might move every `map_deps` up to have their own local variables, so that

* In `map_feature` we won't repeat ourselves filtering dev-deps which we've done to `dev_dependencies` here.
* We can create a map of all dependencies-to-publish so `map_feature` knows which is safe to strip.

src/cargo/util/toml/mod.rs
```diff
@@ -368,6 +368,42 @@ pub fn prepare_for_publish(
...
+        let Some(features) = features else {
+            return Ok(None);
+        };
+        let features = features
+            .iter()
+            .map(|(name, feature_deps)| {
+                let feature_deps = feature_deps
+                    .iter()
+                    .filter(|feature_dep| {
+                        if let Some(feature_pkg_name) = feature_dep.split('/').next() {
```
weihanglo, line 384: We might want to use [`FeatureValue`](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/summary/enum.FeatureValue.html) instead of parsing it again.

linyihai:
> Thanks for the contribution!
> 
> Not sure if @linyihai has started on this, though I believe they are willing to help :) One reminder is that even this is not ready for review, it would be lovely to have a PR description for reviewers to understand what's going on.
> 
> Since this PR is marked as draft, so assuming you have some questions. Feel free to ask anything here or on Zulip.

Don't mind. I haven't started this yet. I haven't fully figured out some of the implementation details yet, but I can look at them if I need to

epage:
When this is ready for review, I'd recommend reworking the commits
- An initial commit should exist that has passing tests, showing the current behavior
- A follow up commit should change the behavior and update the tests so they still pass
  - This helps show we are testing the right thing and makes the change in behavior very clear
- Any refactorings along the way should be separate commits

baby230211:
> When this is ready for review, I'd recommend reworking the commits
> 
> * An initial commit should exist that has passing tests, showing the current behavior
> * A follow up commit should change the behavior and update the tests so they still pass
>   
>   * This helps show we are testing the right thing and makes the change in behavior very clear
> * Any refactorings along the way should be separate commits

Sure.  i'll start working on commits.

src/cargo/ops/registry/publish.rs
```diff
@@ -418,7 +420,33 @@ fn transmit(
             .map(|(feat, values)| {
                 (
                     feat.to_string(),
-                    values.iter().map(|fv| fv.to_string()).collect(),
+                    values
```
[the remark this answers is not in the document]
baby230211: @epage , it sounds like it may be more better to strip the feature here??
https://github.com/rust-lang/cargo/blob/9e6288efe0bbccc53d95080070524e4640769339/src/cargo/util/toml/mod.rs#L939-L955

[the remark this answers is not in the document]
epage: That still requires stripping features twice
- Once for the Summary
- Once for `Cargo.toml`

tests/testsuite/publish.rs
```diff
@@ -1686,6 +1686,125 @@ repository = "foo"
...
+            homepage = "foo"
+            repository = "foo"
+
+
+            [features]
+            foo_feature = ["bar/cat"]
+
+            [dev-dependencies.bar]
+            path = "../bar"
+            features = ["cat"]
```
epage, line 1714:
```suggestion
            [features]
            foo_feature = [
                "normal-only/cat",
                "build-only/cat",
                "dev-only/cat",
                "target-normal-only/cat",
                "target-build-only/cat",
                "target-dev-only/cat",
                "normal-and-dev/cat",
            ]
```

tests/testsuite/publish.rs
```diff
@@ -1686,6 +1686,410 @@ repository = "foo"
...
+]
+"#,
+                cargo::core::package::MANIFEST_PREAMBLE
+            ),
+        )],
+    );
+}
+
+#[cargo_test]
+fn publish_with_feature_point_to_target_dev_dep() {
```
epage, line 1894: Do we need two separate tests or could just one do?

tests/testsuite/publish.rs
```diff
@@ -1686,6 +1686,410 @@ repository = "foo"
...
+            baz = { version = "1.0", features = ["cat"] }
+
+
+            [dev-dependencies.bat]
+            version = "1.0"
+            features = ["cat"]
+
+            [dev-dependencies.bar]
+            path = "../bar"
+            features = ["cat"]
```
epage, line 1730: Could you name the dependencies relative to what role they fulfill in the test to make the intent more clear at a glance?

tests/testsuite/publish.rs
baby230211: Fixed. Thanks~

tests/testsuite/publish.rs
baby230211: i'll update to one test. 
Fixed.

tests/testsuite/publish.rs
baby230211: Updated more clear name.

epage: Could you clean up the commits?

baby230211: Sure, updated.

src/cargo/util/toml/mod.rs
```diff
@@ -366,7 +366,55 @@ pub fn prepare_for_publish(
...
+            Some(features) => {
+                for (_name, feature_deps) in features.iter_mut() {
+                    feature_deps.retain(|feature_dep| {
+                        let feature_value = FeatureValue::new(InternedString::new(feature_dep));
+                        match feature_value {
+                            FeatureValue::Dep { dep_name }
+                            | FeatureValue::DepFeature { dep_name, .. } => {
+                                let k = &manifest::PackageName::new(dep_name.to_string()).unwrap();
+                                let include = is_include_dep(k, manifest.dependencies.as_ref())
+                                    || is_include_dep(k, manifest.dev_dependencies.as_ref())
```
epage, line 391:
```suggestion
                                    || is_include_dep(k, manifest.dev_dependencies().as_ref())
```

src/cargo/util/toml/mod.rs
```diff
@@ -366,7 +366,55 @@ pub fn prepare_for_publish(
...
+                                    || is_include_dep(k, manifest.build_dependencies.as_ref())
```
epage, line 392:
```suggestion
                                    || is_include_dep(k, manifest.build_dependencies().as_ref())
```

src/cargo/util/toml/mod.rs
```diff
@@ -366,7 +366,55 @@ pub fn prepare_for_publish(
...
+                                    || manifest.target.as_ref().iter().any(|target_map| {
+                                        target_map.iter().any(|(_k, v)| {
+                                            let include =
+                                                is_include_dep(k, v.dependencies.as_ref())
+                                                    || is_include_dep(
+                                                        k,
+                                                        v.dev_dependencies.as_ref(),
```
epage, line 399:
```suggestion
                                                        v.dev_dependencies().as_ref(),
```

src/cargo/util/toml/mod.rs
```diff
@@ -987,18 +987,39 @@ pub fn to_real_manifest(
         .transpose()?
         .unwrap_or_default();
     let empty_features = BTreeMap::new();
+    let deps_map = deps
+        .iter()
+        .filter(|dep| dep.is_transitive() || dep.specified_req())
+        .map(|dep| (dep.name_in_toml(), dep))
+        .collect::<BTreeMap<InternedString, &Dependency>>();
```
epage, line 994: A red flag for this commit is that no tests changed.

`to_real_manifest` doesn't strip any dependencies so features don't need to be stripped.

src/cargo/util/toml/mod.rs
epage: Tests now cover every combination I can think of

src/cargo/util/toml/mod.rs
epage: A `strip_features` was added and is performed on the `TomlManifest` after the mapping, having the same effect

src/cargo/ops/registry/publish.rs
```diff
@@ -412,13 +414,36 @@ fn transmit(
...
+                        .filter(|fv| {
+                            let feature_value = FeatureValue::new(InternedString::new(fv));
+                            match feature_value {
+                                FeatureValue::Dep { dep_name }
+                                | FeatureValue::DepFeature { dep_name, .. } => {
+                                    if let Some(dep) = deps_map.get(&dep_name) {
+                                        dep.is_transitive() || dep.specified_req()
+                                    } else {
+                                        true
+                                    }
```
epage, line 440: Like we did with the manifest, we should refactor this so we have one way to strip dependencies from a summary and then reuse those results to strip the feature activations, instead of re-implementing the dependency stripping logic in the feature activation code.

src/cargo/util/toml/mod.rs
baby230211: Thanks @epage , i'll then drop this commit later.

src/cargo/util/toml/mod.rs
```diff
@@ -366,7 +366,52 @@ pub fn prepare_for_publish(
...
+                insert_dep_name(&mut dep_name_set, v.dev_dependencies());
+                insert_dep_name(&mut dep_name_set, v.build_dependencies());
+            });
+        }
+
+        let features = manifest.features.as_mut();
+        let Some(features) = features else {
+            return;
+        };
+
```
epage, line 401: This should probably be done before we gather dependencies

baby230211: Good idea, let me update with this.

baby230211: Here's a problem if we move this before we gather dependencies, there will be a borrow checker problem( 
we cannot have mutable reference and immutable reference at the same time.)

src/cargo/util/toml/mod.rs
baby230211: Updated. Thanks

src/cargo/util/toml/mod.rs
baby230211: Updated. Thanks

src/cargo/util/toml/mod.rs
```diff
@@ -366,7 +366,55 @@ pub fn prepare_for_publish(
...
+                                        target_map.iter().any(|(_k, v)| {
+                                            let include =
+                                                is_include_dep(k, v.dependencies.as_ref())
+                                                    || is_include_dep(
+                                                        k,
+                                                        v.dev_dependencies.as_ref(),
+                                                    )
+                                                    || is_include_dep(
+                                                        k,
+                                                        v.build_dependencies.as_ref(),
```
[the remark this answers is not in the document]
baby230211: Updated. Thanks

epage:
Thanks!

@bors r+

PR #13520, opened 2024-03-02
feat: Use consistent colors when testing

Muscraft:
In cargo#13461, it was [noted that `annotate-snippets`](https://github.com/rust-lang/cargo/pull/13461#discussion_r1501136340) made testing hard as it outputs platform specific colors. To fix this I created [annotate-snippets-rs#82](https://github.com/rust-lang/annotate-snippets-rs/pull/82), which added a feature to make the output platform agnostic when enabled. This makes it so glob syntax does not need to be used when matching colored output.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -229,6 +229,7 @@ features = [
 ]
 
 [dev-dependencies]
+annotate-snippets = { workspace = true, features = ["testing-colors"] }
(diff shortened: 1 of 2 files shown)


weihanglo (approved):

weihanglo:
Great! Thank you!


@bors r+

PR #13522, opened 2024-03-03
Clarify warning for using `features` or `default-features` in `patch`

joshtriplett:
When attempting to substitute a local version of a dependency, since the
`patch` section uses syntax similar to a dependency (and the
documentation even says "The syntax is similar to the `[dependencies]`
section", it's natural to assume that other things from `[dependencies]`
also work, such as `features` or `default-features`. Attempting to use
them produces this warning:

> patch for `crate-name-here` uses the features mechanism.
> default-features and features will not take effect because the patch dependency does not support this mechanism

The phrasing "the patch dependency does not support this mechanism"
makes it seem at first glance like `patch` just doesn't support this at
all. But the actual problem is that the user needs to move the
`features`/`default-features` keys to an entry in `dependencies`
instead.

Modify the message, to make this more obvious.

Modify the corresponding warning for `replace` as well.

Update tests accordingly.

-----

Discovered while trying to help @benwis debug an issue.

### DIFF
--- a/src/cargo/core/registry.rs
+++ b/src/cargo/core/registry.rs
@@ -312,8 +312,8 @@ impl<'gctx> PackageRegistry<'gctx> {
 
                 if dep.features().len() != 0 || !dep.uses_default_features() {
                     self.source_config.gctx().shell().warn(format!(
-                        "patch for `{}` uses the features mechanism. \
-                        default-features and features will not take effect because the patch dependency does not support this mechanism",
+                        "patch for `{}` attempts to declare features. \
+                        The `features` and `default-features` keys need to appear in a `dependencies` entry, not the `patch` entry.",
                         dep.package_name()
                     ))?;
                 }
--- a/src/cargo/ops/resolve.rs
+++ b/src/cargo/ops/resolve.rs
@@ -165,8 +165,8 @@ pub fn resolve_ws_with_opts<'gctx>(
                 ws.gctx()
                 .shell()
                 .warn(format!(
-                    "replacement for `{}` uses the features mechanism. \
-                    default-features and features will not take effect because the replacement dependency does not support this mechanism",
+                    "replacement for `{}` attempts to declare features. \
+                    The `features` and `default-features` keys need to appear in a `dependencies` entry, not the `replace` entry.",
                     dep.package_name()
                 ))?
             }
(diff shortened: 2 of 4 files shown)


src/cargo/core/registry.rs
```diff
@@ -312,8 +312,8 @@ impl<'gctx> PackageRegistry<'gctx> {
 
                 if dep.features().len() != 0 || !dep.uses_default_features() {
                     self.source_config.gctx().shell().warn(format!(
-                        "patch for `{}` uses the features mechanism. \
-                        default-features and features will not take effect because the patch dependency does not support this mechanism",
+                        "patch for `{}` attempts to declare features. \
+                        The `features` and `default-features` keys need to appear in a `dependencies` entry, not the `patch` entry.",
```
epage, line 316: I think a key piece of information is still missing: that the features will be ignored (if I understand correctly).

What do you think of:
```suggestion
                        "ignoring features in patch for `{}`. \
                        Patch-specific `features` and `default-features` are unsupported; instead set these in the `dependencies` entry.",
```

epage: imo I think I'd also do this as:
```
shell().warn("ignoring features in patch for `{}`");
shell().note("patch-specific `features` and `default-features` are unsupported; instead set these in the `dependencies` entry");
```

joshtriplett: @epage I'd like to put the recommendation first, if possible. What about:

```suggestion
                        "patch for `{}` attempts to declare features. \
                        The `features` and `default-features` keys need to appear in a `dependencies` entry; they will be ignored in a `patch` entry.",
```

epage: That feels odd to me as this is a warning and should lead with what the warning is about.  I see that being like rustc saying 
```
error: consider introducing a named lifetime parameter

note: missing lifetime specifier
```

I do want to change my suggestion though after thinking more about why these are ignored:
```rust
shell().warn("unused fields in patch for `{}`: {}");
shell().note("configure {} in the `dependencies` entry");
```

joshtriplett: The warning that this leads with is ```patch for `{}` attempts to declare features```.

But :+1: for the warn+note version you just gave:

```rust
shell().warn("unused fields in patch for `{}`: {}");
shell().note("configure {} in the `dependencies` entry");
```

ehuss:
@joshtriplett Sorry for the delay. It looks like you maybe wanted to change it to the `warn()`/`note()` suggested above? Is that something you can do?

joshtriplett:
@ehuss Yes, though I expect it'll be a little while before I get to updating it.

PR #13523, opened 2024-03-03
chore(ci): Ensure lockfile is respected during MSRV testing

epage:
As a hack in cargo-hack, it doesn't respect lockfiles when doing MSRV testing unless `--locked` is passed in.
This adds that so we make sure we don't run into problems with newer, MSRV-imcompatible dependencies come out that break our build.

See
- https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/gix-ref.20CI.20error/near/423319798
- https://github.com/taiki-e/cargo-hack/issues/234
- https://github.com/taiki-e/cargo-hack/pull/236

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -267,4 +267,4 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - uses: taiki-e/install-action@cargo-hack
-    - run: cargo hack check --all-targets --rust-version --workspace --ignore-private
+    - run: cargo hack check --all-targets --rust-version --workspace --ignore-private --locked


taiki-e:
https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/gix-ref.20CI.20error/near/423319798

> I didn't find out until recently that when doing MSRV testing, they blow away the lockfile *because they assume you aren't committing your lockfile and they want to make sure they use a lockfile format that is compatible with your MSRV*.

If this were the *only* reason, we could have changed that behavior without any problem, but unfortunately the fact that cargo has made incompatible lockfile changes in the past complicates problems.

Especially when running locally on a project with updated/uncommitted lockfiles, without this functionality, calling --rust-version/--version-range after a previous build has been generated by the latest cargo will cause the build to fail. Such a build fail is a very annoying behavior from a QoL perspective.

Ideally, I would like to determine on the cargo-hack side which older lockfile formats cause compatibility issues and call generate-lockfile only when necessary, but I'm not sure how much more complicated that would be.

ehuss:
Thanks!

@bors r+

epage:
> https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/gix-ref.20CI.20error/near/423319798
> 
> > I didn't find out until recently that when doing MSRV testing, they blow away the lockfile _because they assume you aren't committing your lockfile and they want to make sure they use a lockfile format that is compatible with your MSRV_.
> 
> If this were the _only_ reason, we could have changed that behavior without any problem, but unfortunately the fact that cargo has made incompatible lockfile changes in the past complicates problems.

I don't see how what we said was different

> Especially when running locally on a project with updated/uncommitted lockfiles, without this functionality, calling --rust-version/--version-range after a `cargo update` or a previous build by the latest cargo will cause the build to fail. Such a build fail is a very annoying behavior from a QoL perspective.
> 
> Ideally, I would like to determine on the cargo-hack side which older lockfile formats cause compatibility issues and call generate-lockfile only when necessary, but I'm not sure how much more complicated that would be.

FYI #12861 was merged which makes lockfile generation respect MSRV.

imo if working with a version older than that
- If the lockfile exists.  Only generate the lockfile if it isn't present
- Maybe add a `--generate-lockfile` flag to opt-in to the current behavior

PR #13533, opened 2024-03-04
fix(toml): Don't warn on unset Edition if only 2015 is compatible

epage:

The warning doesn't help towards the stated goal if the only MSRV-compatible Edition is 2015 (plus, if you're MSRV is that old, you likely know better).

This also fixes a problem in #13505 where we were suggesting to set a field that might require an MSRV bump which may be unactionable to silence and we avoid warnings without an actionable way of silencing until #12235 gives us a general means.

### Additional information

This was discussed in
- #13505
- https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/warn.20on.20missing.20.60edition.60.20field.20in.20Cargo.2Etoml

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -616,16 +616,20 @@ pub fn to_real_manifest(
         let default_edition = Edition::default();
         let latest_edition = Edition::LATEST_STABLE;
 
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
Thanks!

PR #13537, opened 2024-03-05
fix: Consistently compare MSRVs 

epage:

Currently, we use several strategies to evaluate an MSRV
- Relying in `impl Ord for RustVersion` (`dep_msrv <= pkg_msrv`)
- Converting to version requirements
  - Decrementing a version

This consolidates around one strategy: `RustVersion::is_compatible_with`
- Ensure the comparisons have the same behavior
- Centralize knowledge of how to handle pre-release rustc
- Losslessly allow comparing with either rustc or workspace msrv

Refactors are split out

I didn't go through and verify if or how the different approaches varied in behavior, instead consolidating on the one, so only unit tests around the consolidated behavior were added rather than trying to hit all of the corner cases within the various ways `RustVersion` is used.


Eh2406:
Looks like a good simplification.

PR #13552, opened 2024-03-06
feat(toml): accept opt-levels 0-3 as strings too

GrigorenkoPV:
Pretty self-explanatory.

Was playing around with different `opt-level`s and found it annoying that I have to add quotation marks for `s` & `z`, but remove them for `0`, `1`, `2`, & `3`.


epage:
As this is a change in the compatibility surface of Cargo, i would recommend first broaching the subject in an issue or on [Internals](https://internals.rust-lang.org/) to feel out the interest and collect feedback.  For example, we just recently had a long back and forth over several forums over whether to change a string-number to be a number.  These aren't always straightforward decisions.

As such, I think I'm close this and we can revisit when a decision has been made.

PR #13556, opened 2024-03-07
doc: Add doc for -Zpublic-dependency

linyihai:

- add doc for -Zpublic-dependency
- optimize log display about -Zpublic-dependency

### Additional information

### DIFF
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -296,7 +296,21 @@ The 'public-dependency' feature allows marking dependencies as 'public'
 or 'private'. When this feature is enabled, additional information is passed to rustc to allow
 the 'exported_private_dependencies' lint to function properly.
 
-This requires the appropriate key to be set in `cargo-features`:
+To enable this feature, you can either use `-Zpublic-dependency`
+
+```sh
+cargo +nightly run -Zpublic-dependency
+```
+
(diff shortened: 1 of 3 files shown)


src/doc/src/reference/unstable.md
```diff
@@ -296,7 +296,21 @@ The 'public-dependency' feature allows marking dependencies as 'public'
...
+```toml
+# .cargo/config.toml
+[unstable]
+public-dependency = true
+```
+
+Another way not recommended is set the appropriate key in `cargo-features`,
 
 ```toml
 cargo-features = ["public-dependency"]
```
weihanglo, line 316: This is going to be phased out, right? Should we either remove this, or make it clear on that deprecation plan? Like

You could also

```suggestion
`public-dependency` could also be enabled in `cargo-features`, though this is deprecated and will be removed soon.

```toml
# in Cargo.toml
cargo-features = ["public-dependency"]
```

linyihai: Yeah, I was gonna write that it's gonna be deprecated soon. 

It is recommended that you update the documentation when you remove the code.

weihanglo (approved):

weihanglo:
Thanks for pushing the feature forward!

@bors r+

PR #13557, opened 2024-03-07
fix(doc): Collapse down Generated statuses without --verbose 

### LINKED ISSUE #13336: Ability to suppress "Generated" lines from `cargo doc`
### Problem

Since Rust 1.75, when `cargo doc` is run without `--open`, it prints "`Generated <path>`" for each documented crate in the workspace (#5562). While this is useful when users wish to open the documentation by some other means than `--open`, it's clutter (~doubling the height of the output) in a large workspace when the user already has the documentation open or has other means to access the path the documentation will be generated at.

### Proposed Solution

Add a configuration option which suppresses “Generated” lines.

Alternatively, make `cargo doc` emit only one such line for the entire workspace, in the same way that `cargo doc --open` picks one path; this could be an option or might even be acceptable as the default behavior.

### Notes

`--quiet` is not an adequate solution since it hides all output, including progress when re-documenting dependencies from clean.

The reason I personally care about this is that running `cargo doc` is part of my normal development cycle — if I don't run it, I won't discover I've broken an intra-doc link, which happens very often. So, I want the output to be short if nothing is wrong, but still let me see progress. (If `cargo check` reported broken intra-doc links and other doc lints, that would be an even better alternative, but I don't expect that to be feasible.)

epage:

This is trying to balance
- Overwhelming the user with a lot of paths
- Clarity to the user that the one path is representative of the rest

clap before:
```
    Finished dev [unoptimized + debuginfo] target(s) in 3.81s
   Generated /home/epage/src/personal/clap/target/doc/clap/index.html
   Generated /home/epage/src/personal/clap/target/doc/stdio_fixture/index.html
   Generated /home/epage/src/personal/clap/target/doc/clap_bench/index.html
   Generated /home/epage/src/personal/clap/target/doc/clap_builder/index.html
   Generated /home/epage/src/personal/clap/target/doc/clap_complete/index.html
   Generated /home/epage/src/personal/clap/target/doc/clap_complete_fig/index.html
   Generated /home/epage/src/personal/clap/target/doc/clap_complete_nushell/index.html
   Generated /home/epage/src/personal/clap/target/doc/clap_derive/index.html
   Generated /home/epage/src/personal/clap/target/doc/clap_lex/index.html
   Generated /home/epage/src/personal/clap/target/doc/clap_mangen/index.html
```
clap after:
```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.44s
   Generated /home/epage/src/personal/clap/target/doc/clap/index.html and 9 other files
```

Fixes #13336

Tests were updated in earlier commits to help show the behavior change

### DIFF
--- a/src/cargo/ops/cargo_doc.rs
+++ b/src/cargo/ops/cargo_doc.rs
@@ -1,5 +1,5 @@
 use crate::core::compiler::{Compilation, CompileKind};
-use crate::core::{Shell, Workspace};
+use crate::core::{shell::Verbosity, Shell, Workspace};
 use crate::ops;
 use crate::util::context::{GlobalContext, PathAndArgs};
 use crate::util::CargoResult;
@@ -77,7 +77,7 @@ pub fn doc(ws: &Workspace<'_>, options: &DocOptions) -> CargoResult<()> {
             )?;
             open_docs(&path, &mut shell, config_browser, ws.gctx())?;
         }
-    } else {
+    } else if ws.gctx().shell().verbosity() == Verbosity::Verbose {
         for name in &compilation.root_crate_names {
             for kind in &options.compile_opts.build_config.requested_kinds {
                 let path =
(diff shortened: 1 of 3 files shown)


src/cargo/ops/cargo_doc.rs
weihanglo, line 1: Same behavior foir `cargo test --no-run`? 
I believe people is doing evil things like parsing the executable paths…

src/cargo/ops/cargo_doc.rs
```diff
@@ -92,6 +92,36 @@ pub fn doc(ws: &Workspace<'_>, options: &DocOptions) -> CargoResult<()> {
...
+                n => format!(" and {n} other files"),
+            };
+
+            let mut shell = ws.gctx().shell();
+            let link = shell.err_file_hyperlink(&first_path);
+            shell.status(
+                "Generated",
+                format!(
+                    "{}{}{}{remaining}",
+                    link.open(),
```
weihanglo, line 119: Please implement the fancy alternate form for links 😁.

weihanglo (approved):

src/cargo/ops/cargo_doc.rs
epage: There are a lot of problems with `cargo test` and I'd like to look at all of them holistically

weihanglo:
@bors r+

Thanks!

Issue #13343, opened 2024-01-24
Deprecate use of `cargo:token` credential provider by default for crates.io


davidcorrigan714:
### Problem

Fetching dependencies from an authenticate provider using credentials stored in credentials.toml isn't working unless I explicitly add:
```
[registry]
global-credential-providers = ["cargo:token"]
```

Though [the documentation](https://doc.rust-lang.org/cargo/reference/registry-authentication.html) says that should be the default: "the cargo:token provider is used if no providers are configured."

I don't see any CARGO_ env variables set in my environment so it doesn't seem like I have conflicting configurations anywhere. My full config.toml looks like this now:
```
[registry]
global-credential-providers = ["cargo:token"]

[registries.ni]
index = "sparse+https://provider..../"

[registries.ni_pre]
index = "sparse+https://provider..../"

```

### Version

```text
cargo 1.75.0 (1d8b05cdd 2023-11-20)
release: 1.75.0
commit-hash: 1d8b05cdd1287c64467306cf3ca2c8ac60c11eb0
commit-date: 2023-11-20
host: x86_64-pc-windows-msvc
libgit2: 1.7.1 (sys:0.18.1 vendored)
libcurl: 8.4.0-DEV (sys:0.4.68+curl-8.4.0 vendored ssl:Schannel)
os: Windows 10.0.19044 (Windows 10 Enterprise) [64-bit]
```

Eh2406:
This is intentional, if you want to store you tokens in plain text you need to opted into it. Better documentation is welcome!

cc @arlosi

davidcorrigan714:
It's just blatantly wrong at the moment. The [default value specified here](https://doc.rust-lang.org/cargo/reference/config.html#registryglobal-credential-providers) should be updated too.

Eh2406:
Yes. That needs to be clearer! Thanks for the issues!

From the first sentence of the paragraph you quoted:

> Using alternative registries with authentication requires a credential provider to be configured to avoid unknowingly storing unencrypted credentials on disk.

davidcorrigan714:
> Using alternative registries with authentication requires a credential provider to be configured to avoid unknowingly storing unencrypted credentials on disk.

Which is directly contradictory to `the cargo:token provider is used if no providers are configured.` Cause I have no provider configured, and the "cargo:token" provider certainly wasn't used or treated as the default value for the configuration option.

[...]

PR #13558, opened 2024-03-07
fix(credential)!: Fallback when an auth method isn't available on the current machine

epage:

This changes cargo-credential-libsecret to report `UrlNotSupported` errors when `libsecret` can't be loaded.  This was also extended to `token-from-stdout` when the process can't be spawned.

The goal with this change is to make it easier to support a
cross-platform, cross-machine config file.
Before, someone couldn't enable `libsecret` for the machines that
supported it and then fallback to something else on machines that don't.

After this change, we should be able to fallback to other providers.

To help with debugability, we preserve the "`libsecret` can't be loaded"
message by reporting the first "rich" `UrlNotSupported` error message.
This is a newly invented construct as part of this PR.

This was discussed on
[zulip](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/reg.20auth.20and.20libsecret)
and in the cargo team meeting.

This is to improve the cross-platform config support in an effort to
deprecate having config be unset as part of #13343

### Additional information

Compatibility
- **Breaking:** `cargo-credential`'s `Error` went from an `enum` to an opaque `struct` with a separate `ErrorKind`
- `cargo-credential` json messages can now contain more, optional fields
- `cargo-credential-libsecret` will no longer hard error if `libsecret` can't be loaded


src/cargo/util/auth/mod.rs
```diff
@@ -552,20 +553,37 @@ fn credential_action(
...
-                })
-            }
+            Err(e) => match e.kind() {
+                cargo_credential::ErrorKind::UrlNotSupported => {
+                    if e.as_inner().is_some() {
+                        custom_not_supported.get_or_insert_with(|| {
+                            Err::<(), _>(e)
+                                .with_context(|| {
+                                    format!(
+                                        "credential provider `{}` could not handle the request",
```
arlosi, line 563: Could you please add a UI test so we can see the output when there are multiple failed providers?

src/cargo/util/auth/mod.rs
```diff
@@ -552,20 +553,37 @@ fn credential_action(
...
-            e => {
-                return e.with_context(|| {
-                    format!(
-                        "credential provider `{}` failed action `{action}`",
-                        args.join(" ")
-                    )
-                })
-            }
+            Err(e) => match e.kind() {
+                cargo_credential::ErrorKind::UrlNotSupported => {
```
arlosi, line 557: I'm not sure that the `UrlNotSupported` error makes sense for "non-fatal" errors. It was intended to indicate that the provider only supports some registry urls (like a provider only for a specific type of registry). It then grew to also be used for the platform-specific providers on the wrong platform - and that still made some sense since they are providers that don't work with any registry.

This use case of non-fatal (continuable) error is much further from that intent. However, I'm not sure what the right solution is.

Here are some other options:

1. Make all errors "non-fatal" and continue trying providers for all error types
2. Keep the existing behavior, but add an option in the global credential providers config to mark a provider as "optional"
3. Add a new error type of `OtherContinuable` that that works as you've implemented it here.

src/doc/src/reference/credential-provider-protocol.md
```diff
@@ -177,6 +181,10 @@ available.
 {"Err":{
     // Error: The credential could not be found in the provider.
     "kind":"not-found"
+    // (Optional) Error message string to be displayed
+    "message": "free form string error message",
```
arlosi, line 185: Do these messages from `NotFound` errors show anywhere?

src/cargo/util/auth/mod.rs
epage: Talked about this in office hours

Add an alternative to `Other` that is `Backtrack`
- `Backtrack` will need an MSRV note in docs
- Maybe add an alias for `Other` that is `Critical`
  - Be careful about forcing MSRV bump

As for whether every error could have a message, this was a bit unclear.
- `url-not-supported`: doesn't need it
- `not-found`: doesn't need it
- `operation-not-supported`: could benefit from it because a login failure on a provider that doesn't support it could give a better message saying why.

In the end, we decided that since the code was written, we should just move forward with a message on each one.

epage: For message printing:
- On error (everything backtracked):
  - Print every provider that backtracked and the message
- On success (a provider matched)
  - In extra verbose, print the providers that backtracked and the message

epage: Since we're doing a breaking change, we should also do #13615

epage: @arlosi will do that breaking change after this is merged but within the same release

epage:
My problem was actually #13927 and with that fixed, this isn't needed as much

Issue #13003, opened 2023-11-19
Build script not rerun when target rustflags change


Nemo157:
### Problem

Some crates such as `anyhow` utilize the `CARGO_ENCODED_RUSTFLAGS` to customize their build. When a build is performed with only the target rustflags changed the build script does not get rerun, so the build customization from it is inconsistent with the build environment for the library itself.

### Steps

```console
> cargo dl -e anyhow@=1.0.75 && cd anyhow-1.0.75
                          anyhow@=1.0.75   extracted anyhow 1.0.75 to anyhow-1.0.75

> cargo build --target x86_64-unknown-linux-gnu -v
   Compiling anyhow v1.0.75 (/tmp/tmp.7WdQVxA9Y6/anyhow-1.0.75)
     Running `rustc --crate-name build_script_build --edition=2018 build.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --diagnostic-width=201 --crate-type bin --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 --cfg 'feature="default"' --cfg 'feature="std"' -C metadata=ec4d899bd2351612 -C extra-filename=-ec4d899bd2351612 --out-dir /run/user/1000/cargo-home/target/shared/debug/build/anyhow-ec4d899bd2351612 -C incremental=/run/user/1000/cargo-home/target/shared/debug/incremental -L dependency=/run/user/1000/cargo-home/target/shared/debug/deps`
     Running `/run/user/1000/cargo-home/target/shared/debug/build/anyhow-ec4d899bd2351612/build-script-build`
[...]

PR #13560, opened 2024-03-08
[WIP][fix]:Build script not rerun when target rustflags change

heisen-li:

Fixes https://github.com/rust-lang/cargo/issues/13003

### DIFF
--- a/tests/testsuite/build_script.rs
+++ b/tests/testsuite/build_script.rs
@@ -5527,3 +5527,80 @@ fn test_old_syntax_with_old_msrv() {
     p.cargo("build -v").run();
     p.cargo("run -v").with_stdout("foo\n").run();
 }
+
+#[cargo_test]
+fn build_script_rerun_when_target_rustflags_change() {
+    let p = project()
+        .file(
+            "src/lib.rs",
+            "#![cfg_attr(error_generic_member_access, feature(error_generic_member_access))]",
+        )
+        .file(
+            "build.rs",
+            r#"
+            use std::env;
+            use std::process::{Command, Stdio};
+            fn main() {
+                let rustc = env::var_os("RUSTC").unwrap();
+                let mut cmd = Command::new(rustc);
+                cmd.stderr(Stdio::null())
+                    .arg("--crate-type=lib")
+                    .arg("--out-dir").arg(env::var_os("OUT_DIR").unwrap())
+                    .arg("build/probe.rs");
+                cmd.arg("--target").arg(env::var_os("TARGET").unwrap());
+                if let Ok(rustflags) = env::var("CARGO_ENCODED_RUSTFLAGS") {
+                    if !rustflags.is_empty() {
+                        for arg in rustflags.split('\x1f') {
+                            cmd.arg(arg);
+                        }
+                    }
+                }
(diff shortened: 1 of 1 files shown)


heisen-li:
@weihanglo @epage I haven't found a better way to limit environment variables to what the build script requires. This seems to require first confirming which environment variables the build script cares about, but I haven't figured out how to confirm this.I will keep looking and if you have any suggestions it would save a lot of time, thank you.

The test cases are from anyhow, I hope it will be helpful for understanding.

heisen-li:
> Note that we'd also need tests for this. Ideally a test is by itself in the first commit, showing the problem (by passing). The second commit fixes the problem and updates the test. The change in the test helps show that we are testing the right thing and what the intended change was.

If we understand it correctly，
`first commit, showing the problem (by passing)` is https://github.com/rust-lang/cargo/pull/13560/commits/dc88e341f734c0fc2883b2a599811a7f0499853a

`second commit fixes the problem and updates the test` is https://github.com/rust-lang/cargo/pull/13560/commits/b3c42f6839466e4790e544ec7586adb6e6db8ee0

heisen-li:
If you want to limit the environment variables to those required by the build script, then you need to determine which environment variables are required by the build script.

I haven't come up with a better solution other than progressive scanning (and I don't think this is the right kind of solution), and it's harder than I thought it would be if I wanted to do it.

Also, more information about whether unnecessary reconstruction occurred might have been better.

Honorable @weihanglo @epage

heisen-li:
The current revision will collect in target extra flags to pass to `rustc`.

Thanks for the discussion in [issue](https://github.com/rust-lang/cargo/issues/13003), it has been very beneficial for me. Is there still a revision I need to make?

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1526,11 +1526,14 @@ See https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-change
             .collect::<CargoResult<Vec<_>>>()?
     };
 
+    let rustflags = build_runner.bcx.rustflags_args(unit).to_vec();
```
weihanglo, line 1529: If we end up doing this, should we also take rustdocflags into account?

weihanglo (commented):
Could you clean up commits to two commits? Assumely we want to have first commit for the test, and last for the fix.

src/cargo/core/compiler/fingerprint/mod.rs
weihanglo: Never mind. we didn't set docflags for running build script.

tests/testsuite/build_script.rs
```diff
@@ -5527,3 +5527,80 @@ fn test_old_syntax_with_old_msrv() {
     p.cargo("build -v").run();
     p.cargo("run -v").with_stdout("foo\n").run();
 }
+
+#[cargo_test]
+fn build_script_rerun_when_target_rustflags_change() {
```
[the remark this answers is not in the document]
heisen-li: Sorry, I may have wasted some computing resources due to my oversight.

The revised test may make it easier to understand the fix.

PR #13562, opened 2024-03-08
refactor(shell): Use alternate to close links

epage:

Simplifies the code for includign hyperlinks, especially the fact that this allows for including the variable in the format string.

Hand tested each case as we don't have SVG support for hyperlinks and we don't have SVG tests for these cases


weihanglo:
Tested. Thanks!

@bors r+

Issue #13555, opened 2024-03-07
`object not found` when fetching git dependency


banool:
### Problem

When building a crate that depends on other crates via git, sometimes you get an error like this:
```
Caused by:
  failed to load source for dependency `processor`

Caused by:
  Unable to update https://github.com/aptos-labs/aptos-indexer-processors.git?rev=d44b2d209f57872ac593299c34751a5531b51352#d44b2d20
  
Caused by:
  object not found - no match for id (d44b2d209f57872ac593299c34751a5531b51352); class=Odb (9); code=NotFound (-3)
```

You can see an instance of this failure here: https://github.com/Homebrew/homebrew-core/pull/165260/files. Which comes from here in case the error fails to show up on the first link: https://github.com/Homebrew/homebrew-core/actions/runs/8180427193/job/22368538918?pr=165260.

This only happens sometimes. It seems to happen more often in CI environments, I'm not sure I've seen it happen locally.

### Steps

Repro is challenging because it occurs only sometimes. If you look at any of the past brew version bumps for the `aptos` formula you'll see that one of the CI steps fails at least once every time due to this issue: https://github.com/Homebrew/homebrew-core/pulls?q=is%3Apr+aptos+is%3Aclosed.

The dep in question comes from here: https://github.com/aptos-labs/aptos-indexer-processors. That repo itself has git submodules, so that could be a factor.

### Possible Solution(s)

I'm not sure how to fix this, mostly for now we've just been adding retries to get around it.

### Version

```text
This is the cargo version we're using for the binary in question.

cargo 1.75.0 (1d8b05cdd 2023-11-20)
release: 1.75.0
commit-hash: 1d8b05cdd1287c64467306cf3ca2c8ac60c11eb0
commit-date: 2023-11-20
host: aarch64-apple-darwin
libgit2: 1.7.1 (sys:0.18.1 vendored)
libcurl: 8.4.0 (sys:0.4.68+curl-8.4.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1u  30 May 2023
os: Mac OS 14.3.1 [64-bit]
```

banool:
I've noticed this happening in GitHub Actions, but notably on a variety of different runners, workflows, etc.

[...]

PR #13563, opened 2024-03-08
Use GITHUB_TOKEN in github_fast_path if available.

ehuss:
This changes the GitHub fast-path to include an Authorization header if the GITHUB_TOKEN environment variable is set. This is intended to help with API rate limits, which is not too hard to hit in CI. If it hits the rate limit, then the fetch falls back to fetching all branches which is significantly slower for some repositories. 

This is also a partial solution for #13555, where the user has a `rev="<commit-hash>"` where that commit hash points to a PR. The default refspec of `["+refs/heads/*:refs/remotes/origin/*", "+HEAD:refs/remotes/origin/HEAD"]` will not fetch commits from PRs (unless they are reachable from some branch).

There is some risk that if the user has a GITHUB_TOKEN that is invalid or expired, then this will cause `github_fast_path` to fail where it previously wouldn't. I think that is probably unlikely to happen, though.

This also adds redirect support. Apparently GitHub now issues a redirect for `repos/{org}/{repo}/commits/{commit}` to `/repositories/{id}/commits/{commit}`.

The test `https::github_works` exercises this code path, and will use the token on CI.

### DIFF
--- a/src/cargo/sources/git/utils.rs
+++ b/src/cargo/sources/git/utils.rs
@@ -1413,6 +1413,7 @@ fn github_fast_path(
                 // the branch has moved.
                 if let Some(local_object) = local_object {
                     if is_short_hash_of(rev, local_object) {
+                        debug!("github fast path already has {local_object}");
                         return Ok(FastPathRev::UpToDate);
                     }
                 }
@@ -1452,11 +1453,18 @@ fn github_fast_path(
     handle.get(true)?;
     handle.url(&url)?;
     handle.useragent("cargo")?;
+    handle.follow_location(true)?; // follow redirects
     handle.http_headers({
         let mut headers = List::new();
         headers.append("Accept: application/vnd.github.3.sha")?;
         if let Some(local_object) = local_object {
             headers.append(&format!("If-None-Match: \"{}\"", local_object))?;
+        }
+        if let Ok(token) = gctx.get_env("GITHUB_TOKEN") {
+            if !token.is_empty() {
+                // Avoid API rate limits if possible.
+                headers.append(&format!("Authorization: Bearer {token}"))?;
+            }
         }
         headers
     })?;
@@ -1472,14 +1480,17 @@ fn github_fast_path(
 
     let response_code = handle.response_code()?;
     if response_code == 304 {
+        debug!("github fast path up-to-date");
         Ok(FastPathRev::UpToDate)
     } else if response_code == 200 {
         let oid_to_fetch = str::from_utf8(&response_body)?.parse::<Oid>()?;
+        debug!("github fast path fetch {oid_to_fetch}");
         Ok(FastPathRev::NeedsFetch(oid_to_fetch))
     } else {
         // Usually response_code == 404 if the repository does not exist, and
         // response_code == 422 if exists but GitHub is unable to resolve the
         // requested rev.
+        debug!("github fast path bad response code {response_code}");
         Ok(FastPathRev::Indeterminate)
     }
 }
--- a/tests/testsuite/https.rs
+++ b/tests/testsuite/https.rs
@@ -4,6 +4,7 @@
 //! or CARGO_PUBLIC_NETWORK_TESTS.
 
 use cargo_test_support::containers::Container;
+use cargo_test_support::paths::{self, CargoPathExt};
 use cargo_test_support::project;
 
 #[cargo_test(container_test)]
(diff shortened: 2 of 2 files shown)


src/cargo/sources/git/utils.rs
```diff
@@ -1458,6 +1458,10 @@ fn github_fast_path(
         if let Some(local_object) = local_object {
             headers.append(&format!("If-None-Match: \"{}\"", local_object))?;
         }
+        if let Ok(token) = std::env::var("GITHUB_TOKEN") {
```
weihanglo, line 1461: Should we check if it is not empty?

src/cargo/sources/git/utils.rs
```diff
@@ -1458,6 +1458,10 @@ fn github_fast_path(
...
+            // Avoid API rate limits if possible.
+            headers.append(&format!("Authorization: token {token}"))?;
```
weihanglo, line 1463: According to [the GitHub REST API doc](https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28#about-authentication), both `token` and `Bearer` scheme work. The [example for GitHub Actions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-2-calling-the-rest-api) uses `Bearer`. Should we use that instead?

> Note: In most cases, you can use Authorization: Bearer or Authorization: token to pass a token. However, if you are passing a JSON web token (JWT), you must use Authorization: Bearer.

weihanglo: I think it depends on which scenario we want to address. If this is for CI, then bearer seems more reasonable?

src/cargo/sources/git/utils.rs
ehuss: Sure!

src/cargo/sources/git/utils.rs
ehuss: Sure, went ahead and changed it.

ehuss:
I decided to go ahead and update the github test to actually verify this behavior, testing the various code paths for different reference kinds. I also added redirect support, since it now seems to be required.

ehuss:
`github fast path bad response code 403`

Hm, I don't yet know why that is happening. I had tested tokens with essentially no permissions. I wonder if the Actions token is different somehow.

weihanglo:
> Hm, I don't yet know why that is happening. I had tested tokens with essentially no permissions. I wonder if the Actions token is different somehow.

But tests passed on Linux and M1 mac. Let's @bors try again

Mark-Simulacrum:
> There is some risk that if the user has a GITHUB_TOKEN that is invalid or expired, then this will cause github_fast_path to fail where it previously wouldn't. I think that is probably unlikely to happen, though.

How worried are we about exhausting rate limit?

It can be pretty easy to hit the GitHub rate limit, so I wonder if the number of queries we do will potentially make users unhappy if we exhaust the limit on their token... if this is only for CI, that's probably fine, but locally I could see it being pretty painful.

What about enabling "hidden" usage of credentials with this? For example, if someone has a GitHub token that can access private repositories, just a cargo fetch would "reveal" that here, right? (My sense is that the threat model for cargo wouldn't really be affected here and so this isn't concerning).

ehuss:
I posted #13718 to fix the immediate problem with the redirect.

ehuss:
The error in CI is:

```
---- https::github_works stdout ----
test , tag = "1.3.2"
thread 'https::github_works' panicked at tests/testsuite/https.rs:189:14:

test failed running `/Users/runner/work/cargo/cargo/target/debug/cargo fetch`
error: expected to find:
[..]github fast path fetch ed185cfb1c447c1b4bd6ac021c9ec3bb02c9e2f2

did not find in output:
    Updating git repository `https://github.com/rust-lang/bitflags.git`
   0.074247515s DEBUG cargo::sources::git::utils: attempting GitHub fast path for https://api.github.com/repos/rust-lang/bitflags/commits/1.3.2
   0.476275986s DEBUG cargo::sources::git::utils: github fast path bad response code 403
   0.476517613s DEBUG cargo::sources::git::utils: skipping gc as there's only 0 pack files
   0.477348019s DEBUG cargo::sources::git::utils: doing a fetch for https://github.com/rust-lang/bitflags.git
   0.477795554s DEBUG cargo::sources::git::utils: initiating fetch of ["+refs/tags/1.3.2:refs/remotes/origin/tags/1.3.2"] from https://github.com/rust-lang/bitflags.git
   1.604590687s  INFO cargo::sources::git::utils: reset /Users/runner/work/cargo/cargo/target/tmp/cit/t1575/home/.cargo/git/checkouts/bitflags-2d5483ada0800237/ed185cf/.git/ to ed185cfb1c447c1b4bd6ac021c9ec3bb02c9e2f2
   1.606428247s DEBUG cargo::sources::git::utils: doing reset
   1.663015905s DEBUG cargo::sources::git::utils: reset done
   1.664236545s DEBUG cargo::sources::git::utils: update submodules for: "/Users/runner/work/cargo/cargo/target/tmp/cit/t1575/home/.cargo/git/checkouts/bitflags-2d5483ada0800237/ed185cf/"
```

I've been unable to reproduce that in my own account, though. 😦

ehuss:
Oh, I'm a dummy. The `GITHUB_TOKEN` env var needs to be set for the test to actually work. I'm somewhat reluctant to do that, though, since the exposure surface would be so large. 😦

ehuss:
> How worried are we about exhausting rate limit?

I was originally not concerned, but I'm a little uncertain now. The Actions token has a limit of 1,000 per hour (or 15,000 for GHEC). I was originally thinking that limit is *per run*. But now I understand it is *per repository* and the limit is shared across all runs. That could be a problem, for example, if you have a large number of workflows, and a hefty number of git dependencies, then I could imagine that chewing through hundreds of points relatively quickly.

> What about enabling "hidden" usage of credentials with this?

I've been trying to think of ways this could be a security issue, but I haven't been able to come up with something concrete. I do have a fair sense of unease with it, though, since this could be a sensitive thing.

----

Since the user would need to set `GITHUB_TOKEN` when running cargo, I'm thinking almost nobody will do that, and thus this PR is probably not worth moving forward with. For some reason, I was originally thinking the env var was always set, but it needs to be done manually. It sucks since the fast path can make a fairly substantial difference in performance. 

I'm going to go ahead and close since I don't want to put any more time into it. This might be something we could consider doing for people who really want the fast-path performance, but I'm not sure it is worth it.

PR #13564, opened 2024-03-08
refactor(lockfile): Make diffing/printing more reusable

epage:

This is prep for #13561 so we can tailor the printing of lockfile changes to each use without a bunch of flags.

### DIFF
--- a/src/cargo/ops/cargo_generate_lockfile.rs
+++ b/src/cargo/ops/cargo_generate_lockfile.rs
@@ -176,11 +176,11 @@ fn print_lockfile_update(
         gctx.shell().status_with_color(status, msg, color)
     };
     let mut unchanged_behind = 0;
-    for ResolvedPackageVersions {
+    for PackageDiff {
         removed,
         added,
         unchanged,
-    } in compare_dependency_graphs(&previous_resolve, &resolve)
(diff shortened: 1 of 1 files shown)


src/cargo/ops/cargo_generate_lockfile.rs
```diff
@@ -405,4 +399,25 @@ impl PackageDiff {
 
         changes.into_iter().map(|(_, v)| v).collect()
     }
+
+    /// For querying [`PackageRegistry`] for alternative versions to report to the user
+    pub fn alternatives_query(&self) -> Option<crate::core::dependency::Dependency> {
+        let package_id = [
+            self.added.iter(),
+            self.unchanged.iter(),
+            self.removed.iter(),
```
weihanglo, line 418: `removed` is also included here. I don't get the reason reporting latest availables for removed crates.

epage: This function is more generic than that and exists just to find alternative versions for the current `PackageDiff`.  If a new design has a reason to look up alternatives for remove-only then I'd rather not surprise them by this not working,

weihanglo (approved):

weihanglo:
Thanks. Looks good.

@bors r+

PR #13566, opened 2024-03-08
fix: address RUSTSEC-2023-0072

rex4539:
https://rustsec.org/advisories/RUSTSEC-2023-0072


ehuss:
Thanks for the PR! Unfortunately we are currently forced to use an older version. See #13550 and #13179. Closing as we need to wait for upstream changes.

epage:
We just downgraded openssl because of problems, see #13550

Issue #7876, opened 2020-02-07
Cargo dies when in worktree of bare repository


maurer:
**Problem**
When building rustc from a downloaded tarball (in this case, [1.40.0](https://static.rust-lang.org/dist/rust-1.40.0-x86_64-unknown-linux-gnu.tar.gz)) that is inside a worktree of a bare git repo, Cargo will notice the `.git` directory and attempt to list files and [die while trying](https://github.com/rust-lang/cargo/blob/master/src/cargo/sources/path.rs#L211).

**Steps**

1. Download a Rust [source tarball](https://static.rust-lang.org/dist/rust-1.40.0-x86_64-unknown-linux-gnu.tar.gz)
2. Unpack it and check it into a git repository
3. Create a bare clone of that git repository
4. Use "git worktree" to create a checkout of the bare git repository.
5. Attempt to build rustc (e.g. `./x.py install`)

(This obviously isn't minimized, but is the original environment in which I encountered the problem. I expect the problem would occur when trying to use a path source in any worktree.)

**Possible Solution(s)**

* Use `list-tree` based on the current worktree rather than `list-files`
* Fall back to non-git base path sources in the event of a bare repo

**Notes**
This is related to #4131 but actually caused our CI to start failing because the backend switched to bare repos from normal clones to save disk.

I created a separate issue because it is not restricted to `cargo publish` and is not just a workflow issue.

Output of `cargo version`: `cargo 1.40.0-dev`

ehuss:
I can't seem to figure out how to reproduce this.  Can you put together an exact series of commands to run to repro it?  I tried:

```
cargo new foo
cd foo
# Create build.rs so it will list git files.
echo "fn main() {}" > build.rs
git add * .gitignore
git commit -m initial
cd ..
git clone --bare foo foo.bare
cd foo.bare
git worktree add master
cd master
cargo build
cargo package
```

I also tried something similar with the rustc-src tarball.

jynnantonix:
I am hitting this as well.  Why on earth does a build system care about the version control I'm using?!?!

chachi:
This also seems to fail when doing a sparse checkout with recent git (used to cut down cloned repo size).

Here's a repro with this repo itself:

```
git clone --sparse --filter=blob:none https://github.com/rust-lang/cargo/
cd cargo
git sparse-checkout add src crates
cargo +nightly check
```

I get the output

[...]

Issue #11165, opened 2022-09-30
Add support for sparse git checkouts


sdroege:
### Problem

Currently if a git repository is listed as a dependency in `Cargo.toml` then the whole git repository is cloned, which might contain a lot of other unnecessary things.

### Proposed Solution

It would be nice if cargo supported sparse git checkouts so that only a specific subdirectory of the repository is cloned. This subdirectory would have to be specified in `Cargo.toml` together with the dependency.

cargo would then use this subdirectory as the root and assume it to be either a plain crate or a workspace, like it now does for the actual repository root.

weihanglo:
Thank you for the suggestion. I learnt new stuff today!

As I understand it, with [sparse-checkout], git still needs to clone the whole repository index. Only if pairing with [partial-clone] can git avoids full clone. That means Cargo needs the supports of both features from libgit2, which Cargo depends on for all git-related operations at this time being. Those two features in libgit2 are still under consideration/development, and you can track them from [here](https://github.com/libgit2/libgit2/issues/2263) and [here](https://github.com/libgit2/libgit2/issues/5564). Without libgit2 supports them natively, Cargos hardly helps.

In the meanwhile, the Cargo team plans to do experiments on replacing some of [git2](https://crates.io/crates/git2) functionality in Cargo with [gitoxide](https://crates.io/crates/gitoxide). The short-term goals of gitoxide doesn't seem to include either sparse-index and partial-clone. You might want to kindly ask them about their opinions on these features.

[sparse-checkout]: https://git-scm.com/docs/sparse-index
[partial-clone]: https://git-scm.com/docs/partial-clone

Byron:
As an update, sparse-indices are supported in the sense that they can be read, and every index interaction does consider them. It's a bit of a fringe feature right now as well, but it's on the radar.

And indeed, partial checkouts would only help so-and-so much if it wasn't accompanied with a way of reducing the initial download size. Today, a shallow clone, i.e. receiving only the data needed for the most recent commit, can already help (presumably, it also takes more time on the remote to generate). 

As a next step, I imagine doing a partial clone with blob filter, so only a single commit (or maybe even the whole history) without any blob is downloaded. Despite being custom-generated, it should be fast as blobs should be the most costly here. Finally, `gitoxide` (partial) checkouts would have to be partial-repository aware, collect the missing blobs, and download them separately as part of the checkout. That pack would only be the subset of blobs actually needed, which should be good for a speed-boost on all sides.

[...]

Issue #4131, opened 2017-06-06
cargo publish doesn't work well with git worktrees


seanmonstar:
It seems that cargo works fine if a worktree hasn't been changed around much, since testing with a brand new worktree worked fine. However, these steps cause problems:

```
git checkout -b cargo-worktree
git worktree add ../foo-worktree cargo-worktree
cd ../
mv foo-worktree foo-cargo-worktree
cd foo-cargo-worktree
cargo pacakge -l
```

As for why you'd possibly do this: I have a worktree to track the 'previous version branch' of hyper. It started as a worktree of 0.8.x, and eventually its become 0.10.x, and I updated the name of the directory. All git commands work fine, but I cannot `cargo publish`.

This works fine with `cargo 0.18.0 (fe7b0cdcf 2017-04-24)`, but fails with `cargo 0.20.0-nightly (38ca9b702 2017-05-14)`.

alexcrichton:
Could you gist the error message you're seeing? This is probably just bubbling up an error when it should likely be ignoring it.

seanmonstar:
With `RUST_LOG=trace`:

```
TRACE:cargo::ops::cargo_read_manifest: read_package; path=/home/sean/code/hyper-0.10.x/Cargo.toml; source-id=file:///home/sean/code/hyper-0.10.x
DEBUG:cargo::core::workspace: find_root - trying /home/sean/code/Cargo.toml
DEBUG:cargo::core::workspace: find_root - trying /home/sean/Cargo.toml
DEBUG:cargo::core::workspace: find_root - trying /home/Cargo.toml
DEBUG:cargo::core::workspace: find_root - trying /Cargo.toml
DEBUG:cargo::core::workspace: find_members - only me as a member
TRACE:cargo::ops::cargo_read_manifest: read_package; path=/home/sean/code/hyper-0.10.x/Cargo.toml; source-id=file:///home/sean/code/hyper-0.10.x
TRACE:cargo::ops::cargo_read_manifest: read_package; path=/home/sean/code/hyper-0.10.x/Cargo.toml; source-id=file:///home/sean/code/hyper-0.10.x
DEBUG:cargo::ops::cargo_package: found a git repo at "./", checking if index present
   Packaging hyper v0.10.12 (file:///home/sean/code/hyper-0.10.x)
WARN:cargo::sources::path: list_files_git hyper v0.10.12 (file:///home/sean/code/hyper-0.10.x)
[...]

PR #13567, opened 2024-03-09
test: Add tests for using worktrees and sparse checkouts

LuuuXXX:

Based on @eminence's [comment](https://github.com/rust-lang/cargo/issues/7876#issuecomment-1325974682), Add tests for using worktrees or spase checkouts.

Checkout and run tests:
```
cargo test --package cargo --test testsuite -- git::git_worktree_with_original_repo_renamed
```
```
cargo test --package cargo --test testsuite -- git::git_worktree_with_bare_original_repo
```

### DIFF
--- a/tests/testsuite/git.rs
+++ b/tests/testsuite/git.rs
@@ -3760,3 +3760,135 @@ fn different_user_relative_submodules() {
 
     assert!(project.bin("foo").is_file());
 }
+
+#[cargo_test]
+fn git_worktree_with_original_repo_renamed() {
+    let project = project().build();
+    let git_project = git::new("foo", |project| {
+        project
+            .file(
+                "Cargo.toml",
+                r#"
+                    [package]
+                    name = "foo"
+                    version = "0.5.0"
+                    edition = "2015"
+                    authors = []
+                    license = "MIR OR Apache-2.0"
+                    description = "A test!"
+                    homepage = "https://example.org"
+                    documentation = ""
+                    repository = "https://example.org"
+                    readme = "./README.md"
+                "#,
+            )
+            .file("src/lib.rs", "")
+            .file("README.md", "")
+    });
+
+    let repo = git2::Repository::open(&git_project.root()).unwrap();
+    let repo_root = repo.workdir().unwrap().parent().unwrap();
+    let opts = git2::WorktreeAddOptions::new();
+    let _ = repo
+        .worktree("bar", &repo_root.join("bar"), Some(&opts))
+        .unwrap();
+
+    // Rename the original repository
+    let new = repo_root.join("foo2");
+    fs::rename(&git_project.root(), &new).unwrap();
+
+    project
+        .cargo("package --list")
+        .cwd(&new)
+        .with_stdout(
+            "\
+.cargo_vcs_info.json
+Cargo.toml
+Cargo.toml.orig
+README.md
+src/lib.rs
+",
+        )
+        .run();
+
+    project
+        .cargo("check")
+        .cwd(&new)
+        .with_stderr(
+            "\
+[CHECKING] foo v0.5.0 ([CWD])
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [..]
+",
+        )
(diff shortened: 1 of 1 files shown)


tests/testsuite/git.rs
```diff
@@ -3760,3 +3762,89 @@ fn different_user_relative_submodules() {
...
+Cargo.toml.orig
+README.md
+src/lib.rs
+",
+        )
+        .run();
+}
+
+#[cargo_test(requires_git)]
+fn cargo_check_with_sparse_checkouts() {
```
weihanglo, line 3812: Could you remove this test?

libgit2 hasn't supported sparse checkout (see <https://github.com/rust-lang/cargo/issues/11165>), hence cargo doesn't. I believe this test work just because the repo already has everything checked out.

tests/testsuite/git.rs
```diff
@@ -3760,3 +3762,89 @@ fn different_user_relative_submodules() {
 
     assert!(project.bin("foo").is_file());
 }
+
+#[cargo_test(requires_git)]
+fn cargo_package_with_new_worktree() {
```
weihanglo, line 3767: This test demonstrates "`cargo package` in a worktree from a non-bare Git repoistory" works. However, it doesn't verify the exact steps described in #4131, which is renaming the original repository.

Could you add a extra rename step and see if `cargo package` still works?

Also, it would be better if we could also verify `cargo check` works in the test.

tests/testsuite/git.rs
weihanglo, line 1: I'd like to see the exact test reproducing the old failing case:

* Create a worktree from a "bare" Git repository, and 
* check if both `cargo check` and `cargo package` work.

tests/testsuite/git.rs
LuuuXXX: deleted。 🆗

tests/testsuite/git.rs
LuuuXXX: >Could you add a extra rename step and see if cargo package still works?

modified.

tests/testsuite/git.rs
LuuuXXX: new test was added.

tests/testsuite/git.rs
```diff
@@ -3760,3 +3762,133 @@ fn different_user_relative_submodules() {
...
+        .cwd(&root)
+        .exec()
+        .unwrap();
+
+    let repo = git2::Repository::open(&root).unwrap();
+    assert!(repo.is_bare());
+    let repo_root = root.parent().unwrap();
+    let opts = git2::WorktreeAddOptions::new();
+    let wt = repo
+        .worktree("bar", &repo_root.join("bar"), Some(&opts))
```
weihanglo, line 3868: Not sure if this is better, but at least doesn't depend on external `git` CLI, and guaranteed to be actually `bare`. 

The current patch turns the non-bare repo to bare manually, but still have some checked out file leftover. I am not sure if it would cause any behavior out of expectation.

```suggestion
    let repo = {
        let mut repo_builder = git2::build::RepoBuilder::new();
        repo_builder
            .bare(true)
            .clone_local(git2::build::CloneLocal::Local)
            .clone(
                path2url(git_project.root()).as_str(),
                &paths::root().join("foo-bare"),
            )
            .unwrap()
    };
    assert!(repo.is_bare());
    let opts = git2::WorktreeAddOptions::new();
    let wt = repo
        .worktree("bar", &paths::root().join("bar"), Some(&opts))
```

LuuuXXX: >The current patch turns the non-bare repo to bare manually, but still have some checked out file leftover. I am not sure if it would cause any behavior out of expectation.

Indeed, thanks for your suggestion,Changes submitted.

weihanglo (approved):

weihanglo:
Looks great. Thanks!

@bors r+

codewriter3000:
I still have this issue running 1.83.0-nightly.

weihanglo:
@codewriter3000 could you share more details and a minimal reproduction?

Issue #13405, opened 2024-02-06
multiple path dependencies with the same name confuses the resolver


emilio:
I recently published a new rust-bindgen version (0.69.4), and that caused some Firefox CI jobs to fail (https://bugzilla.mozilla.org/show_bug.cgi?id=1878575), because some CI checks that run `cargo metadata` ended up touching the lockfile and updating bindgen.

The relevant bits that seem to be needed to reproduce this is a local crate for an older version of the crate.

To reproduce:

 * `git clone https://github.com/mozilla/gecko-dev`
 * `cd gecko-dev`
 * `git reset --hard 9d8d867dac5d35156a51ef21fe7b46f573aaabef`
 * `cargo metadata`

Somehow, the lockfile changes to `bindgen 0.69.4` even though the lockfile specifies 0.69.2

emilio:
(Apologies for not having something more reduced, can try to do so if needed, though might take a bit)

ehuss:
Thanks for the report! I believe I have a partial understanding of the problem. The following is a test for cargo's testsuite that demonstrates the issue:

```rust
#[cargo_test]
fn patched_reexport_stays_locked() {
    // Patch example where you emulate a semver-incompatible patch via a re-export.
    // Testing an issue where the lockfile does not stay locked after a new version is published.
    Package::new("bar", "1.0.0").publish();
    Package::new("bar", "2.0.0").publish();
    Package::new("bar", "3.0.0").publish();

    let p = project()
        .file(
            "Cargo.toml",
            r#"
                [workspace]


                [package]
                name = "foo"

                [dependencies]
                bar1 = {package="bar", version="1.0.0"}
                bar2 = {package="bar", version="2.0.0"}

                [patch.crates-io]
                bar1 = { package = "bar", path = "bar-1-as-3" }
                bar2 = { package = "bar", path = "bar-2-as-3" }
            "#,
        )
        .file("src/lib.rs", "")
        .file(
            "bar-1-as-3/Cargo.toml",
            r#"
                [package]
                name = "bar"
                version = "1.0.999"

                [dependencies]
                bar = "3.0.0"
            "#,
        )
        .file("bar-1-as-3/src/lib.rs", "")
        .file(
            "bar-2-as-3/Cargo.toml",
            r#"
                [package]
                name = "bar"
                version = "2.0.999"

                [dependencies]
                bar = "3.0.0"
            "#,
        )
        .file("bar-2-as-3/src/lib.rs", "")
        .build();

[...]

PR #13572, opened 2024-03-11
Fix:  Make path dependencies with the same name stays locked

linyihai:
Fixes: https://github.com/rust-lang/cargo/issues/13405

This is a workround based on https://github.com/rust-lang/cargo/issues/13405#issuecomment-1930496807

first commit will pass, second commit fixed it and update test.

### DIFF
--- a/src/cargo/core/resolver/encode.rs
+++ b/src/cargo/core/resolver/encode.rs
@@ -154,7 +154,7 @@ impl EncodableResolve {
     /// primary uses is to be used with `resolve_with_previous` to guide the
     /// resolver to create a complete Resolve.
     pub fn into_resolve(self, original: &str, ws: &Workspace<'_>) -> CargoResult<Resolve> {
-        let path_deps = build_path_deps(ws)?;
+        let path_deps: HashMap<String, HashMap<semver::Version, SourceId>> = build_path_deps(ws)?;
         let mut checksums = HashMap::new();
 
         let mut version = match self.version {
@@ -202,7 +202,11 @@ impl EncodableResolve {
                 if !all_pkgs.insert(enc_id.clone()) {
                     anyhow::bail!("package `{}` is specified twice in the lockfile", pkg.name);
                 }
-                let id = match pkg.source.as_deref().or_else(|| path_deps.get(&pkg.name)) {
+                let id = match pkg
+                    .source
+                    .as_deref()
+                    .or_else(|| get_source_id(&path_deps, pkg))
+                {
                     // We failed to find a local package in the workspace.
                     // It must have been removed and should be ignored.
                     None => {
@@ -364,7 +368,11 @@ impl EncodableResolve {
 
         let mut unused_patches = Vec::new();
         for pkg in self.patch.unused {
-            let id = match pkg.source.as_deref().or_else(|| path_deps.get(&pkg.name)) {
+            let id = match pkg
+                .source
+                .as_deref()
+                .or_else(|| get_source_id(&path_deps, &pkg))
+            {
                 Some(&src) => PackageId::try_new(&pkg.name, &pkg.version, src)?,
                 None => continue,
             };
@@ -395,7 +403,7 @@ impl EncodableResolve {
             version = ResolveVersion::V2;
         }
 
-        Ok(Resolve::new(
+        return Ok(Resolve::new(
             g,
             replacements,
             HashMap::new(),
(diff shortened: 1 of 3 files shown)


linyihai:
@ehuss any chance for some feedback?  I'm not sure I'm doing the right thing.

src/cargo/core/resolver/encode.rs
```diff
@@ -404,11 +412,42 @@ impl EncodableResolve {
...
             version,
             HashMap::new(),
-        ))
+        ));
+
+        fn get_source_id<'a>(
+            path_deps: &'a HashMap<String, HashMap<semver::Version, SourceId>>,
+            pkg: &'a EncodableDependency,
+        ) -> Option<&'a SourceId> {
+            {
```
ehuss, line 421: This looks like it has an extraneous block here. Can that be removed?

src/cargo/core/resolver/encode.rs
```diff
@@ -404,11 +412,42 @@ impl EncodableResolve {
...
+                }) {
+                    Some(source_id)
+                } else {
+                    // Next look for cases where the name is the same
+                    path_deps.iter().find_map(|(name, version_source)| {
+                        if name == &pkg.name {
+                            return version_source.values().next();
+                        }
+                        None
+                    })
```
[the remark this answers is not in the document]
linyihai: > Can you say more about why this was chosen, and maybe investigate if there is a better way to treat it?|

Please forget my previous writing, I submitted a more reasonable version, please check again sometime.

src/cargo/core/resolver/encode.rs
```diff
@@ -404,11 +412,33 @@ impl EncodableResolve {
...
+        ) -> Option<&'a SourceId> {
+            path_deps.iter().find_map(|(name, version_source)| {
+                if name == &pkg.name {
+                    // If there are multiple candidates for the same name, it needs to be determined by combining versions (See #13405).
+                    if version_source.len() > 1 {
+                        if let Ok(pkg_version) = pkg.version.parse::<semver::Version>() {
+                            if let Some(source_id) = version_source.get(&pkg_version) {
+                                return Some(source_id);
+                            }
+                        }
```
ehuss, line 429: I'm still a little concerned about the behavior here when there are multiple versions, but none of them match. There are two concerns:

* This randomly chooses one of the packages, which is not great for consistency.
* Depending on what changed and which random package it picks, you can end up with two packages with the same SourceId.

I'm still wondering if it is appropriate to return `None` here if there are multiple versions, but none of them match. For example, something like:

```suggestion
                        }
                        
                        return None;
```

Does that make sense?

ehuss:
Here is a test that illustrates the problem with having multiple versions:

```rust
#[cargo_test]
fn same_name_version_changed() {
    let p = project()
        .file(
            "Cargo.toml",
            r#"
                [package]
                name = "foo"
                version = "1.0.0"
                edition = "2021"

                [dependencies]
                foo2 = { path = "foo2", package = "foo" }
            "#,
        )
        .file("src/lib.rs", "")
        .file(
            "foo2/Cargo.toml",
            r#"
                [package]
                name = "foo"
                version = "2.0.0"
                edition = "2021"
            "#,
        )
        .file("foo2/src/lib.rs", "")
        .build();

    p.cargo("tree")
        .with_stderr("[LOCKING] 2 packages to latest compatible versions")
        .with_stdout(
            "\
foo v1.0.0 ([ROOT]/foo)
└── foo v2.0.0 ([ROOT]/foo/foo2)
",
        )
        .run();

    p.change_file(
        "foo2/Cargo.toml",
        r#"
            [package]
            name = "foo"
            version = "2.0.1"
            edition = "2021"
        "#,
    );
    p.cargo("tree")
        .with_stderr(
            "\
[ADDING] foo v2.0.1 ([ROOT]/foo/foo2)
",
        )
        .with_stdout(
            "\
foo v1.0.0 ([ROOT]/foo)
└── foo v2.0.1 ([ROOT]/foo/foo2)
",
        )
        .run();
}
```

With this PR as-is, this test will randomly fail.

I'd recommend adding this test to the `path.rs` testsuite module.

linyihai:
Oh on, the ci faild due to 
```


failures:
    bench::many_similar_names

test result: FAILED. 3326 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out; finished in 390.01s

error: test failed, to rerun pass `-p cargo --test testsuite`
```

I guess [this pr ](https://github.com/rust-lang/cargo/pull/13901) merge can solve this

ehuss:
Thanks! I moved the suggested test to `path.rs` since it is unrelated to patching.  I'm still not 100% confident with that final `None` return, but I can't think of any real alternatives.

@bors r+

PR #13573, opened 2024-03-11
fix(doc): syntax of `rustc-check-cfg` instruction

birkenfeld:
Remove the extra colon, and change the example to something that is actually valid.

### DIFF
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -1141,7 +1141,7 @@ You can use the flag like this:
 cargo check -Z unstable-options -Z check-cfg
 ```
 
-### `cargo::rustc-check-cfg=CHECK_CFG`
+### `cargo:rustc-check-cfg=CHECK_CFG`
 
 The `rustc-check-cfg` instruction tells Cargo to pass the given value to the
 `--check-cfg` flag to the compiler. This may be used for compile-time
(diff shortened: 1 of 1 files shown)


weihanglo:
The new two-colons syntax is valid in the current beta and nightly, and will be available in 1.77.0.

The old one-colon syntax is still supported for compatibility reasons.

See <https://github.com/rust-lang/cargo/pull/12201>

birkenfeld:
Oh I see! Sorry. Do you want to keep the other change?

weihanglo:
The stabilization PR of it is <https://github.com/rust-lang/cargo/pull/13571>. I would recommend adding a comment there for doc improvements.

src/doc/src/reference/unstable.md
```diff
@@ -1157,7 +1157,8 @@ You can use the instruction like this:
 
 ```rust,no_run
 // build.rs
-println!("cargo::rustc-check-cfg=cfg(foo, bar)");
```
Urgau, line 1160: Btw, this is valid; this defines two different cfgs `foo` and `bar`, not one `foo="bar"`.

But I agree this isn't the best example.

weihanglo:
Going to close this as the PR author of #13571 is aware of the example improvement. Thank you all :)

Issue #12692, opened 2023-09-18
CI version bump check can get confused when PR is behind master


ehuss:
The [`check-version-bump`](https://github.com/rust-lang/cargo/blob/819fa73b2dd86e32bf01a7317aec0fa0a447c871/.github/workflows/main.yml#L88-L104) check generates an error in the following case:

1. Contributor creates a branch.
2. Some PR merges that modifies a crate, for example bumping cargo-credential's version number.
3. Contributor submits their PR, where its base is before step 2 is merged.

This causes an error like:

```
   BumpCheck base commit `819fa73b2dd86e32bf01a7317aec0fa0a447c871`
   BumpCheck head commit `c9262bd58021cd3bcca101f68bc3a9efbb57cb70`
   BumpCheck compare against `crates-io`
    Updating crates.io index
   BumpCheck compare against `9c4383fb55986096b414d98125421ab87b5fd642`
error: Detected changes in these crates but no version bump found:
  cargo-credential@0.4.0
  cargo-credential-1password@0.4.0
```

I think the issue is that the diff is not using a "symmetric difference" between base and head. In the example above, base is the latest master (with the changes to cargo-credential) and head is the tip of my branch (which is missing those changes). The check is doing the equivalent of `git diff 819fa73b2dd86e32bf01a7317aec0fa0a447c871 c9262bd58021cd3bcca101f68bc3a9efbb57cb70`, which pulls in all changes that have happened on master since the branch was created. I think a solution is to use a symmetric diff (with git's `...` syntax), but AFAIK libgit2 does not support that. So I'm not sure how to solve this.

weihanglo:
How about using `git` when it is available, and libgit2 as a fallback?

ehuss:
I think it would be fine to unconditionally use the `git` CLI. I think it is safe to say anyone running this will have it installed.

PR #13581, opened 2024-03-14
fix(ci): bump check error when PR is behind master

LuuuXXX:

Fixes: https://github.com/rust-lang/cargo/issues/12692

### DIFF
--- a/crates/xtask-bump-check/src/xtask.rs
+++ b/crates/xtask-bump-check/src/xtask.rs
@@ -333,10 +333,8 @@ fn changed<'r, 'ws>(
             (relative_pkg_root, pkg)
         })
         .collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


weihanglo (commented):
Thanks! This is pretty good. Didn't know git2 has `merge_base` API

I've tested against the example in <https://github.com/rust-lang/cargo/issues/12692> with my suggestion, and it works 👍🏾

Specifically, you should see 

```
git merge-base 819fa73b2dd86e32bf01a7317aec0fa0a447c871 c9262bd58021cd3bcca101f68bc3a9efbb57cb70
```

yields the same common ancestor `41cef471d03e3ad18ee3e1a9e228fe33a89dba82`

weihanglo (approved):
Thanks!

Issue #13456, opened 2024-02-18
cargo publish has bad error message when explicit `[[bench]]` isn't in `package.include`


dhardy:
### Problem

`cargo publish` fails

### Steps

Using the [latest rand commit](https://github.com/rust-random/rand/commit/b1427c6447ce6f2aaeda967f777d6daebed18164):
```sh
$ cargo publish --dry-run
    Updating crates.io index
   Packaging rand v0.9.0-alpha.0 (/home/dhardy/projects/rand/rand)
    Updating crates.io index
   Verifying rand v0.9.0-alpha.0 (/home/dhardy/projects/rand/rand)
error: failed to verify package tarball

Caused by:
  failed to parse manifest at `/home/dhardy/projects/rand/rand/target/package/rand-0.9.0-alpha.0/Cargo.toml`

Caused by:
  can't find `uniform` bench at `benches/uniform.rs` or `benches/uniform/main.rs`. Please specify bench.path if you want to use a non-default path.
```

### Notes

The relevant part of the `Cargo.toml`:
```toml
[[bench]]
name = "uniform"
harness = false
```
This *could* have a `path` specified... but as far as I can tell, `cargo publish` is the only thing not happy to discover the default path (`benches/uniform.rs`).

Despite this, the benchmark works fine:
```sh
$ ls -l benches/uniform.rs 
-rw-r--r--. 1 dhardy dhardy 2696 Feb 13 11:39 benches/uniform.rs
$ cargo bench --bench uniform --features small_rng
   Compiling rand v0.9.0-alpha.0 (/home/dhardy/projects/rand/rand)
    Finished `bench` profile [optimized] target(s) in 2.28s
     Running benches/uniform.rs (target/release/deps/uniform-38f8da4b7c78a479)
samplei8/SmallRng/single
                        time:   [1.5141 ns 1.5143 ns 1.5145 ns]
[snip]
```

### Version

```text
cargo 1.78.0-nightly (fc1d58fd0 2024-02-09)
release: 1.78.0-nightly
commit-hash: fc1d58fd0531a57a6b942a14cdcdbcb82ece16f3
commit-date: 2024-02-09
host: x86_64-unknown-linux-gnu
libgit2: 1.7.2 (sys:0.18.2 vendored)
libcurl: 8.6.0-DEV (sys:0.4.71+curl-8.6.0 vendored ssl:OpenSSL/1.1.1w)
ssl: OpenSSL 1.1.1w  11 Sep 2023
[...]

PR #13589, opened 2024-03-15
refactor(toml): Flatten manifest parsing

epage:

This is just a clean up but the goals are
- Support diagnostics that show source by tracking `&str`, `ImDocument`, etc in `Manifest` by making each accessible in the creation of a `Manifest`
- Defer warning analysis until we know what is a local vs non-local workspace by refactoring warnings out into a dedicated step
- Centralize the logic for `cargo publish` stripping of dev-dependencies and their feature activations by allowing a `Summary` to be created from any "resolved" `TomlManifest`
- Enumerate all build targets in the "resolved" `TomlManifest` so they get included in `cargo publish`, reducing the work done on registry dependencies and resolving problems like #13456

Along the way, this fixed a bug where we were not reporting warnings from virtual manifests

### DIFF
--- a/crates/cargo-util-schemas/src/manifest/mod.rs
+++ b/crates/cargo-util-schemas/src/manifest/mod.rs
@@ -6,6 +6,7 @@
 //! - Keys that exist for bookkeeping but don't correspond to the schema have a `_` prefix
 
 use std::collections::BTreeMap;
+use std::collections::BTreeSet;
 use std::fmt::{self, Display, Write};
 use std::path::PathBuf;
 use std::str;
@@ -28,6 +29,7 @@ pub use rust_version::RustVersionError;
 #[derive(Debug, Deserialize, Serialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct TomlManifest {
+    // when adding new fields, be sure to check whether `requires_package` should disallow them
     pub cargo_features: Option<Vec<String>>,
     pub package: Option<Box<TomlPackage>>,
     pub project: Option<Box<TomlPackage>>,
@@ -51,7 +53,11 @@ pub struct TomlManifest {
     pub workspace: Option<TomlWorkspace>,
     pub badges: Option<InheritableBtreeMap>,
     pub lints: Option<InheritableLints>,
-    // when adding new fields, be sure to check whether `requires_package` should disallow them
+
+    /// Report unused keys (see also nested `_unused_keys`)
+    /// Note: this is populated by the caller, rather than automatically
+    #[serde(skip)]
+    pub _unused_keys: BTreeSet<String>,
 }
 
 impl TomlManifest {
--- a/src/cargo/core/manifest.rs
+++ b/src/cargo/core/manifest.rs
@@ -42,7 +42,7 @@ impl EitherManifest {
 #[derive(Clone, Debug)]
 pub struct Manifest {
     // alternate forms of manifests:
-    original: Rc<TomlManifest>,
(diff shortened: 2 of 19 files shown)


src/cargo/core/manifest.rs
```diff
@@ -41,7 +41,9 @@ impl EitherManifest {
 /// This is deserialized using the [`TomlManifest`] type.
 #[derive(Clone, Debug)]
 pub struct Manifest {
+    original: Rc<TomlManifest>,
     summary: Summary,
+
```
Muscraft: Nit: Adding a comment for the two sections explaining what they are and why they are split up, could be helpful

src/cargo/util/toml/mod.rs
```diff
@@ -209,7 +209,7 @@ fn convert_toml(
             }
         }
     }
-    return if manifest.project.is_some() || manifest.package.is_some() {
+    return if manifest.package().is_some() {
```
Muscraft: I believe both are checked here as we technically still support `[project]`

crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -6,6 +6,7 @@
 //! - Keys that exist for bookkeeping but don't correspond to the schema have a `_` prefix
 
 use std::collections::BTreeMap;
+use std::collections::BTreeSet;
```
Muscraft, line 9: Nit: this could be condensed
```suggestion
use std::collections::{BTreeMap, BTreeSet};
```

src/cargo/util/toml/mod.rs
```diff
@@ -515,6 +501,21 @@ pub fn to_real_manifest(
         );
     };
 
+    if let Some(deps) = me
```
Muscraft, line 487: This might have been better to keep together as we will now have to update two places

Muscraft (commented):
The cleanup looks nice!

crates/cargo-util-schemas/src/manifest/mod.rs
epage: For myself, I prefer not combining them.  I've dealt with many complex `use`s that are annoying to read, edit, and cause conflicts and much prefer each lint being independent.

src/cargo/util/toml/mod.rs
epage: Yes, that is the intention of this function, along with `dev_dependencies()`, `build_dependencies()`, etc

src/cargo/util/toml/mod.rs
epage: This isn't the only thing duplicated and my hope is that future refactors will allow us to merge most of the logic back together.

Muscraft (approved):

src/cargo/util/toml/mod.rs
Muscraft: I somehow missed the you were calling the function. disregard this comment

src/cargo/core/manifest.rs
```diff
@@ -42,7 +42,7 @@ impl EitherManifest {
 #[derive(Clone, Debug)]
 pub struct Manifest {
     // alternate forms of manifests:
-    original: Rc<TomlManifest>,
+    resolved_toml: Rc<TomlManifest>,
```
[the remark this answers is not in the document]
epage: Many more upcoming refactors.

At least `resolved` is more accurate than `original` :)

Issue #13576, opened 2024-03-12
Tracking Issue for Packages as (optional) namespaces


epage:
### Summary

RFC: [#3243](https://github.com/rust-lang/rfcs/pull/3243) 
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#my-feature
rustc tracking issue: rust-lang/rust#122349

Allow a parent package to open its API namespace for other packages to participate

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

hydroperfox:
I have created this simple package `sunform::package1` with the `open-namespaces`  feature enabled.

```toml
cargo-features = ["open-namespaces"]

[package]
...
```

[sunform-package1.zip](https://github.com/user-attachments/files/16323749/sunform-package1.zip)

When I initialize a Git repository in the local package, commit files, and run

```batch
cargo publish --allow-dirty
```

I get

```plain
    Updating crates.io index
error: failed to open: C:\Users\hydro\Downloads\Repositories\sunform-package1\target\package\.sunform::package1-0.1.0.crate

Caused by:
  A sintaxe do nome do arquivo, do nome do diretório ou do rótulo do volume está incorreta. (os error 123)
```

So for Windows these don't work yet?

epage:
@hydroperfox `publish` support is not done yet, see this task from the task list

> [ ] Publish support including index file system layout

hydroperfox:
I've remembered of binaries, where the default binary's name is derived from the package's name, in which case it requires translating `::` to `-` when writing the binary.

PR #13591, opened 2024-03-15
feat: Add 'open-namespaces' feature

epage:

This is a step towards #13576

### DIFF
--- a/crates/cargo-util-schemas/src/restricted_names.rs
+++ b/crates/cargo-util-schemas/src/restricted_names.rs
@@ -33,13 +33,15 @@ enum ErrorKind {
     FeatureNameStartsWithDepColon(String),
 }
 
-/// Check the base requirements for a package name.
-///
-/// This can be used for other things than package names, to enforce some
-/// level of sanity. Note that package names have other restrictions
-/// elsewhere. `cargo new` has a few restrictions, such as checking for
-/// reserved names. crates.io has even more restrictions.
-pub(crate) fn validate_package_name(name: &str, what: &'static str) -> Result<()> {
+pub(crate) fn validate_package_name(name: &str) -> Result<()> {
+    validate_name(name, "package name")
+}
+
+pub(crate) fn validate_registry_name(name: &str) -> Result<()> {
+    validate_name(name, "registry name")
+}
+
+pub(crate) fn validate_name(name: &str, what: &'static str) -> Result<()> {
     if name.is_empty() {
         return Err(ErrorKind::Empty(what).into());
     }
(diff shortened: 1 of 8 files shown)


tests/testsuite/open_namespaces.rs
```diff
@@ -0,0 +1,357 @@
...
+                [package]
+                name = "foo::bar"
+                version = "0.0.1"
+                edition = "2015"
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("read-manifest")
```
weihanglo, line 53: Why chose `cargo read-manifest` over `cargo metadata`?

weihanglo (approved):
Thanks. Just one question about `read-manifest`.

tests/testsuite/open_namespaces.rs
epage: Less output to deal with.

`read-manfiest` is a short term command.  I expect to replace this with `check` as we work to get compilation working.

crates/cargo-util-schemas/src/restricted_names.rs
```diff
@@ -84,6 +89,17 @@ pub(crate) fn validate_package_name(name: &str, what: &'static str) -> Result<()
 
 /// Ensure a package name is [valid][validate_package_name]
 pub(crate) fn sanitize_package_name(name: &str, placeholder: char) -> String {
+    let mut slug = String::new();
```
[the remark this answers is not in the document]
epage: `-Zscript` is also unstable :)

tests/testsuite/open_namespaces.rs
```diff
@@ -0,0 +1,358 @@
...
+      "crate_types": [
+        "lib"
+      ],
+      "doc": true,
+      "doctest": true,
+      "edition": "2015",
+      "kind": [
+        "lib"
+      ],
+      "name": "foo::bar",
```
Nemo157, line 90: My understanding from the RFC was that the library name was meant to just be `bar`, that would avoid the rustc issue you noted

> The autogenerated `lib.name` key for such a crate will just be bar, the leaf crate name, and the expectation is that to use such crates one must use `--extern foo::bar=bar.rlib` syntax. There may be some better things possible here, perhaps `foo_bar` can be used here.
> —<https://rust-lang.github.io/rfcs/3243-packages-as-optional-namespaces.html#reference-level-explanation>

Nemo157: (I'd assume this also applies to the autogenerated `bin.name` for `src/main.rs`)

tisonkun:
@epage @weihanglo Is there an easy way I enable this feature in a workspace? Now it seems I need to add `cargo-features = ["open-namespaces"]` to each crates in the workspace and I have more than 20 crates in the workspace ..

weihanglo:
No.

Just note that at this moment this feature is not useful at all.

epage:
To add, only enough has been implemented to unblock compiler work.

tisonkun:
Thank you! Yes I see https://github.com/rust-lang/rust/issues/122349#issuecomment-2002208868 and hit exactly the same now.

PR #13592, opened 2024-03-15
Use `gitoxide` for `list_files_git`

Byron:
Related to #10150.

### Tasks

* [x] update `gix` to v0.60
* [x] assure this is tested (currently only git-tests run with `git2` and `gitoxide`)
* [x] allow `list_files_git` to use `gitoxide` if it is enabled as feature.
* [x] use dirwalk iterator
* [x] use new release of `gix` with necessary updates

### Review Notes

As this PR has come a long way, I decided to keep a few of the steps leading up to the final state, showing the PR's evolution in the hope it helps the review.

* Would it be better to simply use `gitoxide` for this without a switch? I don't think
  it will cause more trouble than `git2`, and if there is an issue I will fix it with priority.
* In one test, the walk resolves a symlink to a submodule to individual files, including the `.git/*` folder contained in the submodule which is ignored by the walk, i.e. `submodule/*` does not contain it, but `submodule-link/*` does. This is fixed in the gitoxide version, and the `git2` version.
* I noticed that symlinks are resolved for packaging *and* are allowed to point to anywhere, even outside of package root. I left it, but felt that maybe this should be reconsidered.

### Remarks

* I love the test-suite! It's incredibly exhaustive to the point where it uncovers shortcomings in `gitoxide`, which I greatly appreciate.
* I also love `git2` as it's API for many things leads to pretty idiomatic code, and sometimes I really have to work to match it. The example here is the initial `dirwalk()` method which requires a delegate as it doesn't just collect into a `Vec` like `git2` does (for good reason). Turning that into an iterator via `dirwalk_iter()` makes it far more usable, and will definitely be good for performance as the dirwalk work is offloaded into its own thread.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -72,7 +72,6 @@
         'cargo',
       ],
       matchPackageNames: [
-        'gix-features-for-configuration-only',
         'gix',
       ],
       groupName: 'gix',
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -67,15 +67,6 @@
       internalChecksFilter: 'strict',
       groupName: 'msrv',
     },
-    {
-      matchManagers: [
-        'cargo',
-      ],
-      matchPackageNames: [
-        'gix',
-      ],
-      groupName: 'gix',
-    },
     // Goals:
     // - Rollup safe upgrades to reduce CI runner load
     // - Have lockfile and manifest in-sync (implicit rules)
--- a/src/cargo/sources/path.rs
+++ b/src/cargo/sources/path.rs
@@ -263,8 +263,12 @@ impl<'gctx> PathSource<'gctx> {
     /// Returns [`Some(gix::Repository)`](gix::Repository) if there is a sibling `Cargo.toml` and `.git`
     /// directory; otherwise, the caller should fall back on full file list.
     fn discover_gix_repo(&self, root: &Path) -> CargoResult<Option<gix::Repository>> {
-        let repo = match gix::discover(root) {
-            Ok(repo) => repo,
+        let mut mapping = gix::sec::trust::Mapping::default();
+        mapping.full = gix::open::Options::isolated();
+        mapping.reduced = gix::open::Options::isolated();
+        let repo = match gix::ThreadSafeRepository::discover_opts(root, Default::default(), mapping)
+        {
+            Ok(repo) => repo.to_thread_local(),
             Err(e) => {
                 tracing::debug!(
                     "could not discover git repo at or above {}: {}",
--- a/src/cargo/sources/path.rs
+++ b/src/cargo/sources/path.rs
@@ -533,26 +533,30 @@ impl<'gctx> PathSource<'gctx> {
             &mut delegate,
         )?;
 
-        let mut files = delegate.into_result()?;
+        let (mut files, mut subpackages_found) = delegate.into_result()?;
         // Append all normal files that might be tracked in `target/`.
-        files.extend(
-            index
-                .prefixed_entries(target_prefix.as_ref())
-                .unwrap_or_default()
-                .iter()
-                .filter(|entry| {
-                    entry.stage() == 0
-                        && entry
-                            .mode
-                            .to_tree_entry_mode()
-                            .map(|mode| mode.kind())
-                            .map_or(false, |kind| {
-                                use gix::object::tree::EntryKind::*;
-                                matches!(kind, Blob | BlobExecutable | Link)
-                            })
-                })
-                .map(|entry| root.join(gix::path::from_bstr(entry.path(&index)))),
-        );
+        for entry in index
+            .prefixed_entries(target_prefix.as_ref())
+            .unwrap_or_default()
+            .iter()
+            // probably not needed as conflicts prevent this to run, but let's be explicit.
+            .filter(|entry| entry.stage() == 0)
+        {
+            handle_path(
+                entry.path(&index),
+                // Do not trust what's recorded in the index, enforce checking the disk.
+                // This traversal is not part of a `status()`, and tracking things in `target/`
+                // is rare.
+                None,
+                root,
+                pkg,
+                pkg_path,
+                self,
+                &mut files,
+                &mut subpackages_found,
+                filter,
+            )?
+        }
         return Ok(files);
 
         struct Delegate<'a, 'gctx> {
@@ -602,10 +606,10 @@ impl<'gctx> PathSource<'gctx> {
                 })
             }
 
-            fn into_result(self) -> CargoResult<Vec<PathBuf>> {
+            fn into_result(self) -> CargoResult<(Vec<PathBuf>, Vec<PathBuf>)> {
                 match self.err {
                     None => {
-                        return Ok(self.paths);
+                        return Ok((self.paths, self.subpackages_found));
                     }
                     Some(e) => return Err(e),
                 }
@@ -615,65 +619,80 @@ impl<'gctx> PathSource<'gctx> {
                 if entry.status == Status::Untracked && entry.rela_path.as_ref() == "Cargo.lock" {
                     return Ok(());
                 }
-                let file_path = self
-                    .root
-                    .join(gix::path::from_bstr(entry.rela_path.as_ref()));
-                if file_path.file_name().and_then(|name| name.to_str()) == Some("Cargo.toml") {
-                    // Keep track of all sub-packages found and also strip out all
-                    // matches we've found so far. Note, though, that if we find
-                    // our own `Cargo.toml`, we keep going.
-                    let path = file_path.parent().unwrap();
-                    if path != self.pkg_path {
-                        warn!("subpackage found: {}", path.display());
-                        self.paths.retain(|p| !p.starts_with(path));
-                        self.subpackages_found.push(path.to_path_buf());
-                        return Ok(());
-                    }
-                }
+                handle_path(
+                    entry.rela_path.as_ref(),
+                    entry.disk_kind,
+                    self.root,
+                    self.pkg,
+                    self.pkg_path,
+                    self.parent,
+                    &mut self.paths,
+                    &mut self.subpackages_found,
+                    self.filter,
+                )
+            }
+        }
 
-                // If this file is part of any other sub-package we've found so far,
-                // skip it.
-                if self
-                    .subpackages_found
-                    .iter()
-                    .any(|p| file_path.starts_with(p))
-                {
+        fn handle_path(
+            rela_path: &gix::bstr::BStr,
+            kind: Option<gix::dir::entry::Kind>,
+            root: &Path,
+            pkg: &Package,
+            pkg_path: &Path,
+            source: &PathSource<'_>,
+            paths: &mut Vec<PathBuf>,
+            subpackages_found: &mut Vec<PathBuf>,
+            filter: &dyn Fn(&Path, bool) -> bool,
+        ) -> CargoResult<()> {
+            let file_path = root.join(gix::path::from_bstr(rela_path));
+            if file_path.file_name().and_then(|name| name.to_str()) == Some("Cargo.toml") {
+                // Keep track of all sub-packages found and also strip out all
+                // matches we've found so far. Note, though, that if we find
+                // our own `Cargo.toml`, we keep going.
+                let path = file_path.parent().unwrap();
+                if path != pkg_path {
+                    warn!("subpackage found: {}", path.display());
+                    paths.retain(|p| !p.starts_with(path));
+                    subpackages_found.push(path.to_path_buf());
                     return Ok(());
                 }
-
-                let is_dir = entry.disk_kind.map_or(false, |kind| {
-                    if kind == gix::dir::entry::Kind::Symlink {
-                        // Symlinks must be checked to see if they point to a directory
-                        // we should traverse.
-                        file_path.is_dir()
-                    } else {
-                        kind.is_dir()
-                    }
-                });
-                if is_dir {
-                    // This could be a submodule, or a sub-repository. In any case, we prefer to walk
-                    // it with git-support to leverage ignored files and to avoid pulling in entire
-                    // .git repositories.
-                    match gix::open_opts(&file_path, gix::open::Options::isolated()) {
-                        Ok(sub_repo) => {
-                            let files =
-                                self.parent
-                                    .list_files_gix(self.pkg, &sub_repo, self.filter)?;
-                            self.paths.extend(files);
-                        }
-                        _ => {
-                            self.parent
-                                .walk(&file_path, &mut self.paths, false, self.filter)?;
-                        }
-                    }
-                } else if (self.filter)(&file_path, is_dir) {
-                    assert!(!is_dir);
-                    // We found a file!
-                    warn!("  found {}", file_path.display());
-                    self.paths.push(file_path);
-                }
-                Ok(())
             }
+
+            // If this file is part of any other sub-package we've found so far,
+            // skip it.
+            if subpackages_found.iter().any(|p| file_path.starts_with(p)) {
+                return Ok(());
+            }
+
+            let is_dir = kind.map_or(false, |kind| {
+                if kind == gix::dir::entry::Kind::Symlink {
+                    // Symlinks must be checked to see if they point to a directory
+                    // we should traverse.
+                    file_path.is_dir()
+                } else {
+                    kind.is_dir()
+                }
+            });
+            if is_dir {
+                // This could be a submodule, or a sub-repository. In any case, we prefer to walk
+                // it with git-support to leverage ignored files and to avoid pulling in entire
+                // .git repositories.
+                match gix::open_opts(&file_path, gix::open::Options::isolated()) {
+                    Ok(sub_repo) => {
+                        let files = source.list_files_gix(pkg, &sub_repo, filter)?;
+                        paths.extend(files);
+                    }
+                    _ => {
+                        source.walk(&file_path, paths, false, filter)?;
+                    }
+                }
+            } else if (filter)(&file_path, is_dir) {
+                assert!(!is_dir);
+                // We found a file!
+                warn!("  found {}", file_path.display());
+                paths.push(file_path);
+            }
+            Ok(())
         }
     }
 
--- a/src/cargo/sources/path.rs
+++ b/src/cargo/sources/path.rs
@@ -13,7 +13,7 @@ use crate::util::{internal, CargoResult, GlobalContext};
 use anyhow::Context as _;
 use cargo_util::paths;
 use filetime::FileTime;
-use gix::bstr::{ByteSlice, ByteVec};
+use gix::bstr::ByteVec;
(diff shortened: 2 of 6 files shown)


src/cargo/sources/path.rs
```diff
@@ -365,15 +425,8 @@ impl<'gctx> PathSource<'gctx> {
...
-                let rel = file_path.strip_prefix(root)?;
-                let rel = rel.to_str().ok_or_else(|| {
-                    anyhow::format_err!("invalid utf-8 filename: {}", rel.display())
-                })?;
-                // Git submodules are currently only named through `/` path
-                // separators, explicitly not `\` which windows uses. Who knew?
-                let rel = rel.replace(r"\", "/");
-                match repo.find_submodule(&rel).and_then(|s| s.open()) {
+                warn!("  found directory {}", file_path.display());
+                match git2::Repository::open(&file_path) {
```
Byron, line 429: As the intent is to add the directory, submodule or not, it's better to open it as Repository to avoid picking up `.git` folders. This fix I applied after noticing this in the `gitoxide` implementation.

src/cargo/sources/path.rs
```diff
@@ -411,6 +464,151 @@ impl<'gctx> PathSource<'gctx> {
...
+        filter: &dyn Fn(&Path, bool) -> bool,
+    ) -> CargoResult<Vec<PathBuf>> {
+        warn!("list_files_gix {}", pkg.package_id());
+        let options = repo
+            .dirwalk_options()?
+            .emit_untracked(gix::dir::walk::EmissionMode::Matching)
+            .emit_ignored(None)
+            .emit_tracked(true)
+            .recurse_repositories(false)
+            .symlinks_to_directories_are_ignored_like_directories(true)
```
Byron, line 486: This option I added specifically to support a difference in behaviour when comparing `libgit2` to Git. Symlinks to directories are treated as directories by the `.gitignore` machinery (on Unix only in `libgit2`, all platforms in `gitoxide`). Git itself always treats symlinks as files.

src/cargo/sources/path.rs
```diff
@@ -411,6 +464,151 @@ impl<'gctx> PathSource<'gctx> {
...
+            } else {
+                "/target/"
+            });
+
+            // The `target` directory is never included if it is in the package root.
+            let exclude = {
+                let mut pattern = include.clone();
+                pattern
+                    .to_mut()
+                    .insert_str(0, if include.is_empty() { ":!" } else { ":!/" });
```
Byron, line 518: This pathspec prefix is short for  'exclude from top', with the 'top' being required to avoid the CWD to affect the pathspec.

src/cargo/sources/path.rs
```diff
@@ -411,6 +464,151 @@ impl<'gctx> PathSource<'gctx> {
...
+                        // probably not needed as conflicts prevent this to run, but let's be explicit.
+                        entry.stage() == 0
+                    })
+                    .map(|entry| {
+                        (
+                            entry.path(&index).to_owned(),
+                            // Do not trust what's recorded in the index, enforce checking the disk.
+                            // This traversal is not part of a `status()`, and tracking things in `target/`
+                            // is rare.
+                            None,
```
Byron, line 556: This is a possible source of failures in the `git2` version. The index can generally not be trusted to represent the working tree unless a `status` call has set the `UPTODATE` flag on the entry. This isn't currently happening, and fixing this would be more complicated than it's worth.
Maybe another argument to switch this part over to `gix`.

src/cargo/sources/path.rs
```diff
@@ -411,6 +464,151 @@ impl<'gctx> PathSource<'gctx> {
...
+                    file_path.is_dir()
+                } else {
+                    kind.is_dir()
+                }
+            });
+            if is_dir {
+                // This could be a submodule, or a sub-repository. In any case, we prefer to walk
+                // it with git-support to leverage ignored files and to avoid pulling in entire
+                // .git repositories.
+                match gix::open_opts(&file_path, gix::open::Options::isolated()) {
```
Byron, line 596: I opted for 'isolated' open (to avoid being affected by user or system settings) as these shouldn't be relevant for what we are doing. This makes sure of that, and is faster.

tests/testsuite/package.rs
```diff
@@ -689,6 +689,7 @@ fn package_symlink_to_submodule() {
     project
         .cargo("package --no-verify -v")
         .with_stderr_contains("[ARCHIVING] submodule/Makefile")
+        .with_stderr_does_not_contain("[ARCHIVING] submodule-link/.git/config")
```
Byron, line 692: This was to assure that in sub-repositories, we never pick up `.git` directories.

tests/testsuite/package.rs
```diff
@@ -3410,6 +3411,29 @@ derp/not_target/foo.txt
...
 target/foo.txt
+",
+        )
+        .run();
+
+    // Untracked files shouldn't be included, if they are also ignored.
+    _ = fs::write(repo.workdir().unwrap().join(".gitignore"), "target/").unwrap();
+    git::add(&repo);
+    git::commit(&repo);
+    _ = fs::write(p.build_dir().join("untracked.txt"), "").unwrap();
```
Byron, line 3422: This test was added to further stress the `gitoxide` implementation, which I could otherwise have implemented wrongly. But even if that wasn't the case, it's good to see that only tracked files in `target` will be picked up.

Byron (commented):
Once CI has passed, I think this is ready for review. I left plenty of comments in this PR, to associate some of my 'Review Notes' with code.

Overall, I am really excited about this PR and can say that it greatly improved `gitoxide` in the process. It's quite amazing to see what Cargo does to list files as well :).

.github/renovate.json5
```diff
@@ -72,7 +72,6 @@
         'cargo',
       ],
       matchPackageNames: [
-        'gix-features-for-configuration-only',
         'gix',
       ],
       groupName: 'gix',
```
epage, line 77: We can likely get rid of this group then.  This was just to make sure all gix packages were updated in the same PR

src/cargo/sources/path.rs
```diff
@@ -246,6 +259,52 @@ impl<'gctx> PathSource<'gctx> {
         Ok(None)
     }
 
+    /// Returns [`Some(gix::Repository)`](gix::Repository) if there is a sibling `Cargo.toml` and `.git`
+    /// directory; otherwise, the caller should fall back on full file list.
+    fn discover_gix_repo(&self, root: &Path) -> CargoResult<Option<gix::Repository>> {
+        let repo = match gix::discover(root) {
```
epage, line 266: Do we need something like https://github.com/rust-lang/cargo/blob/69726309bf12443a618643cc86125ef60dcbecfb/src/bin/cargo/main.rs#L345-L365 ?

src/cargo/sources/path.rs
```diff
@@ -411,6 +470,228 @@ impl<'gctx> PathSource<'gctx> {
...
+    /// Symlinks to directories will also be resolved, which may include `.git` folders
+    /// in entirety if the link points to a submodule.
+    fn list_files_gix(
+        &self,
+        pkg: &Package,
+        repo: &gix::Repository,
+        filter: &dyn Fn(&Path, bool) -> bool,
+    ) -> CargoResult<Vec<PathBuf>> {
+        use gix::dir::entry::Status;
+        warn!("list_files_gix {}", pkg.package_id());
```
epage, line 479: Why is this a `warn`?  Should we instead `instrument`?

src/cargo/sources/path.rs
epage: What is the `:` syntax and where can I read more about it?

epage: Oh, is it a path separator?  Why are we using strings with separators rather than `Vec`?

src/cargo/sources/path.rs
```diff
@@ -411,6 +470,228 @@ impl<'gctx> PathSource<'gctx> {
...
+
+        let mut files = delegate.into_result()?;
+        // Append all normal files that might be tracked in `target/`.
+        files.extend(
+            index
+                .prefixed_entries(target_prefix.as_ref())
+                .unwrap_or_default()
+                .iter()
+                .filter(|entry| {
+                    entry.stage() == 0
```
epage: What does `stage() == 0` mean?

src/cargo/sources/path.rs
Byron: The `pathspec` syntax is document in [the Git glossary](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec) of all places.

With a top-level `Cargo.toml` file, this code produces two pathspecs: `:/` and `:!/target/`, with a sub-package it would look like `:/sub-package` and `:!/sub-package/target/`.

src/cargo/sources/path.rs
Byron: A great catch! 

And I think it makes sense to at least open it so that it's isolated and ignores user and system configuration, just so that it aligns with the `gix::open_opts()` when opening submodules.

`gix` implements its security mechanism differently as it keeps track of the trust-level of each key. Code that wants to deal with paths/scripts will not see values that come from untrusted files. Thus one can interact with such repositories ordinarily, but without risking to execute untrusted programs.

src/cargo/sources/path.rs
Byron: It's a line that was copied from the `git2` version of the method. I am happy to change it in both once you confirm.

src/cargo/sources/path.rs
Byron: In a later commit this receives a comment that makes it a bit clearer.

```
// probably not needed as conflicts prevent this to run, but let's be explicit.
```

Stage zero essentially excludes entries that are related to conflicts, and it's something the code calling this would prevent as it checks for changes first. However, if that would be bypassed for any reason, it's better not to risk adding entries that point to the same path, but at different stages.

src/cargo/sources/path.rs
```diff
@@ -411,6 +470,228 @@ impl<'gctx> PathSource<'gctx> {
...
+                        }
+                        _ => {
+                            self.parent
+                                .walk(&file_path, &mut self.paths, false, self.filter)?;
+                        }
+                    }
+                } else if (self.filter)(&file_path, is_dir) {
+                    assert!(!is_dir);
+                    // We found a file!
+                    warn!("  found {}", file_path.display());
```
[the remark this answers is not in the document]
Byron: This is also copied from the `git2` implementation, and in a more recent commit I removed the comment at least. If you'd like both instances of this `warn!` to change to `instrument!` maybe, please let me know.

src/cargo/sources/path.rs
epage: Why is a Rust API using stringly-typed values based on a command-line?

src/cargo/sources/path.rs
epage: Huh, thought I checked for it but must have missed it.  No need to "innovate" here.

src/cargo/sources/path.rs
epage: Why are numeric codes being used for this?

src/cargo/sources/path.rs
Byron: Now I see what you are getting at.

I didn't think of innovating here, following the API used by `libgit2` and `git2` respectively. Having strings for pathspecs runs deep, and they are treated more like paths or regex or globs rather than a data structure. [The datastructure itself](https://github.com/Byron/gitoxide/blob/f8ce3d0721b6a53713a9392f2451874f520bc44c/gix-pathspec/src/lib.rs#L84) also isn't particularly well suited to be constructed by hand, as git settings affect it (like case-sensitivity). It's quite complex, but I can imagine to tackle this by adding a trait like `TryIntoPathspec` which then works for strings as well as for hand-created pathspecs.

I see how being able to not have to go through the parser here would be beneficial though, so I have [put this into the roadmap](https://github.com/Byron/gitoxide/commit/3fd668454acece9d218dc7b5200836dce938efbc).

(*Since it's not worse than `git2` which is also stringly typed, I hope this won't have to block this PR.*)

epage: No, thank you for the explanation!

For me, it is more of trying to make the intent of the code clearer.  I'll mark this as resolved but have some other thoughts for improving it with what we have now.

epage: Hmm, was tempted to encourage `pathspec` being added to the variables but the leading `let pathspec =` helps.  Still easy to miss the structure these variables are intended to uphold.  I feel like I could go either way.

src/cargo/sources/path.rs
Byron: I see what you mean now, and investigated this. 
First of all, the stage as it is here is already quite accessible compared to [decoding it by hand from `flags`](https://docs.rs/git2/latest/git2/struct.IndexEntry.html#structfield.flags) as is done in `libgit2`. But then I was wondering why it's not an enum and [found the `stage()` code](https://github.com/Byron/gitoxide/blob/27471e7f421446cc8f6d4543bdf85fe4e753b944/gix-index/src/entry/flags.rs#L64-L66) which initially looked like this bitshift could yield many different values. Digging deeper, it shows this was wrong and the mask is truly 2 bits in size, so an `enum` with 4 variants would be much better.

I pushed the (breaking) changes [into this pending PR](https://github.com/Byron/gitoxide/pull/1317) so that when that version releases and Cargo is updated, it will see the new enums and use them.

Thus, this can't change immediately (to not cause avoidable version-churn), but it will trickle through for sure.

src/cargo/sources/path.rs
Byron: It's definitely quite complex code, with a bunch of variables and inter-dependencies to make it all line up. This could already be improved a lit if `format!()` could be used, but since paths are involved, that's not possible :/.
Maybe the pathspec-code can be refactored into something better nonetheless.

Byron:
Just for clarity, from my point of view all issues brought up so far have been addressed. Is this PR blocked on me nonetheless?

epage:
For myself, I don't fully feel confident in answering a lot of the questions you raised (or identifying issues in this code).  I can say "this is unstable, just merge" but I worry we might lose track of a lot of this.

Would appreciate a look for someone else, especially with a lot of experience with this code.

src/cargo/sources/path.rs
```diff
@@ -139,7 +141,16 @@ impl<'gctx> PathSource<'gctx> {
         let root = pkg.root();
         let no_include_option = pkg.manifest().include().is_empty();
         let git_repo = if no_include_option {
-            self.discover_git_repo(root)?
+            if self
+                .gctx
+                .cli_unstable()
+                .gitoxide
+                .map_or(false, |features| features.fetch)
```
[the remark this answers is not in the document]
epage: A separate flag might be of use if we want to stabilize piece-meal.

weihanglo:
I'll find a time reviewing this.

src/cargo/sources/path.rs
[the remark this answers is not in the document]
Byron: Thanks! This is implemented now, and I called the flag 'list-files'.

src/cargo/sources/path.rs
arlosi: Does this mean that the user/system level gitignore configuration is ignored? I believe that would be a change from the current model.

Some people put editor-specific ignore configuration in their user-level git configuration. Cargo should respect that when publishing.

Byron: That's a great catch, I don't know why I wasn't seeing this 🤦‍♂️.

It's fixed now, which also simplifies the code quite a bit as one can now discover and open the repository without altering the default configuration.

arlosi:
I confirmed this PR does fix the issue I was having related to #10150 for missing the `sdir` extension, as well as the `link` extension mentioned in the issue.

src/cargo/sources/path.rs
```diff
@@ -411,6 +464,153 @@ impl<'gctx> PathSource<'gctx> {
...
+            .work_dir()
+            .ok_or_else(|| anyhow::format_err!("can't list files on a bare repository"))?;
+        assert!(
+            root.is_absolute(),
+            "BUG: paths used internally are absolute, and the repo inherits that"
+        );
+
+        let pkg_path = pkg.root();
+        let mut target_prefix;
+        let repo_relative_pkg_path = pkg_path.strip_prefix(root).ok().unwrap_or(Path::new(""));
```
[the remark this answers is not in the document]
Byron: Thanks for pointing this out. I totally see how that pattern is unnecessary and I will try to avoid this pattern in future.

src/cargo/sources/path.rs
```diff
@@ -246,6 +260,52 @@ impl<'gctx> PathSource<'gctx> {
         Ok(None)
     }
 
+    /// Returns [`Some(gix::Repository)`](gix::Repository) if there is a sibling `Cargo.toml` and `.git`
```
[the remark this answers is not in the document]
Byron: It's indeed copied, but it could be clearer as to what it really does. I have adjusted it (in the `gitoxide` version only for now) and is says:

```Rust
    /// Returns [`Some(gix::Repository)`](gix::Repository) if the discovered repository
    /// (searched upwards from `root`) contains a tracked `<root>/Cargo.toml`.
    /// Otherwise, the caller should fall back on full file list.
```

src/cargo/sources/path.rs
```diff
@@ -411,6 +465,153 @@ impl<'gctx> PathSource<'gctx> {
...
+                pattern
+                    .to_mut()
+                    .insert_str(0, if include.is_empty() { ":!" } else { ":!/" });
+                pattern.to_mut().push_str(b"/target/");
+                pattern
+            };
+            include.to_mut().insert_str(0, ":/");
+
+            vec![include.into_owned(), exclude.into_owned()]
+        };
```
arlosi, line 526: I've attempted to simplify the pathspec code to make it easier to understand. If this is still correct, then I think it's easier to follow (at least for me).
 
```suggestion
        let pkg_path = pkg.root();
        let repo_relative_pkg_path = pkg_path.strip_prefix(root).unwrap_or(Path::new(""));
        let target_prefix = gix::path::to_unix_separators_on_windows(gix::path::into_bstr(
            repo_relative_pkg_path.join("target"),
        ));
        let package_prefix =
            gix::path::to_unix_separators_on_windows(gix::path::into_bstr(repo_relative_pkg_path));

        let pathspec = {
            // Include the package root.
            let mut include = BString::from(":/");
            include.push_str(package_prefix.as_ref());

            // Exclude the target directory.
            let mut exclude = BString::from(":!");
            exclude.push_str(target_prefix.as_ref());

            vec![include, exclude]
        };
```

Byron: Thanks! Much easier to read!

I will commit the suggestion and then work on assuring it's truly the same, I *think* there are slash-related issues right now that alter the meaning of the pathspecs.

Byron: Thanks a lot for your help with this!

Leveraging `Path::join` and building BStrings front to back really make all the difference here.

arlosi:
Thanks! Looking forward to making this the default.

@bors r+

Issue #13280, opened 2024-01-11
Cargo fails to detect environment variable is newly set and rebuild if the environment variable is set in `config.toml`


usamoi:
### Problem

The problem can be only reproduced if the environment variable is "not set" in the first run, and "set" in the second run.

### Incorrect behaivor

.cargo/config.toml:
```toml
[env]
Z = "0"
```

src/main.rs:
```rust
fn main() {
    println!("{:?}", option_env!("Z"));
}
```

run these commands:
```bash
cargo run     # first build, output Some("0")
Z=1 cargo run # cargo does not notice the environment variable is set and runs it, output Some("0")
```

### Correct behaivor if we do not use a `config.toml`

src/main.rs:
```rust
fn main() {
    println!("{:?}", option_env!("Z"));
}
```

run these commands:
```bash
cargo run     # first build, output None
Z=1 cargo run # cargo does know the environment variable is set and rebuilds it, output Some("1")
```

### Correct behaivor if we change the environment variable from a value to another value

.cargo/config.toml:
```toml
[env]
Z = "0"
```

src/main.rs:
```rust
fn main() {
    println!("{:?}", option_env!("Z"));
}
```

run these commands:
```bash
Z=2 cargo run # first build, output Some("2")
Z=1 cargo run # cargo does know the environment variable changes and rebuilds it, output Some("1")
```

### Version

```text
cargo 1.77.0-nightly (ac6bbb332 2023-12-26)
cargo 1.75.0 (1d8b05cdd 2023-11-20)
```

heisen-li:
By the way, it may be related to question #10358.

weihanglo:
> By the way, it may be related to question #10358 & #12434.

Hmm… I don't quite think so. It is related to cargo's own dep-info file failed to track env vars from `[env]` table. Cargo removes those here:

https://github.com/rust-lang/cargo/blob/0c98d6ec3a3fc72683fe00963b8a55ffaa204d2a/src/cargo/core/compiler/fingerprint/mod.rs#L2017-L2019

env vars from `[env]` were [applied to `rustc_cmd` in `fill_env()`](https://github.com/rust-lang/cargo/blob/0c98d6ec3a3fc72683fe00963b8a55ffaa204d2a/src/cargo/core/compiler/compilation.rs#L368) earlier before this call.

weihanglo:
[...]

PR #8421, opened 2020-06-26
Parse `# env-dep` directives in dep-info files

alexcrichton:
This commit updates Cargo's parsing of rustc's dep-info files to account
for changes made upstream in rust-lang/rust#71858. This means that if
`env!` or `option_env!` is used in crate files Cargo will correctly
rebuild the crate if the env var changes.

Closes #8417

PR #13596, opened 2024-03-16
[WIP] Fix: Cargo fails to detect environment variable

LuuuXXX:

Fixes: https://github.com/rust-lang/cargo/issues/13280

Cargo fails to detect and rebuild while using environment variable reset environment variable which is set in `config.toml`.

examples:
.cargo/config.toml:
```
[env]
Z = "0"
```
src/main.rs:
```
fn main() {
    println!("{:?}", option_env!("Z"));
}
```
run these commands:
```
cargo run     # first build, output None
Z=1 cargo run # cargo does know the environment variable is set and rebuilds it, output Some("1")
```

checkout and run test:

```
cargo test --package cargo --test testsuite -- cargo_env_config::env_reset
```

### Additional information
Fingerprint code is not able to detect environment variables cause they were removed from Cargo's dep-info. See https://github.com/rust-lang/cargo/issues/13280#issuecomment-1887225691.

### DIFF
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -420,8 +420,14 @@ pub fn prepare_target(
     let mtime_on_use = build_runner.bcx.gctx.cli_unstable().mtime_on_use;
     let dirty_reason = compare_old_fingerprint(unit, &loc, &*fingerprint, mtime_on_use, force);
 
-    let Some(dirty_reason) = dirty_reason else {
-        return Ok(Job::new_fresh());
+    let dirty_reason = match dirty_reason {
+        Some(dr) => dr,
+        None => {
+            let Some(dr) = env_config_modified(bcx.gctx) else {
+                return Ok(Job::new_fresh());
+            };
+            dr
+        }
     };
 
     // We're going to rebuild, so ensure the source of the crate passes all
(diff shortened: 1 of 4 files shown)


ehuss:
I would expect this detection to be in the fingerprint code. There is code there already for managing changes in environment variables. Can you say why the current fingerprinting code is not catching this case?

weihanglo:
@ehuss
Environment variables were removed from Cargo's dep-info if they were in rustc ProcessBuilder. See <https://github.com/rust-lang/cargo/issues/13280#issuecomment-1887225691>.

weihanglo:
And agree with Eric on this being fixed in fingerprint.

LuuuXXX:
Thanks to @weihanglo for the additional explanation, it has been added to `Additional information`.

src/cargo/core/compiler/fingerprint/mod.rs
weihanglo, line 1: Wonder some alternatives may be like:

* Have a set of environment variables Cargo internally sets for rustc, and removes them from Cargo's dep info [here](https://github.com/rust-lang/cargo/pull/8421). This might need to take `cargo::rustc-env` build script directive into account, and I feel it will be pretty error-prone as we might miss something when adding more envs.
* Leverage the unstable [`--env-set`](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/env-set.html) rustc flag from [the env sandboxing RFC](https://github.com/rust-lang/rfcs/pull/2794). However, it is still unstable so Cargo can't use it.

weihanglo:
@LuuuXXX are you still interested in moving this forward?

LuuuXXX:
Sorry for missing message，I apologize for not having the time to work on it in the coming month，a few things are tripping me up.

weihanglo:
Going to close due to inactivity. We are still interested in some solutions here. Feel free to reopen if you have a chance to work on it again.

Issue #11123, opened 2022-09-22
cargo install --dry-run


epage:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/issues/11123)*

Implementation: https://github.com/rust-lang/cargo/pull/14280

`--dry-run` is an important feature for commands with side effects to help
- Give users confidence in what the commands will do (e.g. what binaries will be installed)
- Offer a tool for users and developers to debug subsets of the functionality
- Allow scripts to check the status of the system

Testing instructions: https://github.com/rust-lang/cargo/issues/11123#issuecomment-2460846663

weihanglo:
This is a feature to pursue. We can make it as an unstable feature behind `-Zunstale-options`. However, it is not really in a high priority, so label it as `S-needs-mentor` for now.

weihanglo:
Hey @PaulDance. Thanks for being interested in this.

Thinking backwards from what users want, here is my idea that the initial version of `--dry-run` could have:

* How: Is it an update (replace), a fresh install, or no-op?
* What: which binary from which source is going to be installed
* Where: to where the binary will be installed

There might be somethings hard to do in the first PR, like programming interface for scripting, or checking if yanked versions will block the installation. That's fine, since we plan to put it behind `-Zunstable-options` (the process of adding it can be found [here](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/features/index.html#-z-unstable-options)). It is benefical for maintainers to review the implementation, if we can break down a feature a into several small PRs. It also increases the chance to merge them sooner.


For the implementation, it might not be as easy as it looks like at first glance. It depends on how acurate we want for `cargo install --dry-run`. The best option we have is collecting the info after Cargo computing all the necessary build tasks.

[...]

PR #13598, opened 2024-03-17
Basic implementation for `cargo install --dry-run`

PaulDance:

This is an attempt at providing a starting implementation for `cargo install --dry-run` discussed in #11123. The precise objective is more to have something that works now rather than a much more complete version requiring time and complexity-facing skills I don't currently have.

Indeed, https://github.com/rust-lang/cargo/issues/11123#issuecomment-1910469992 suggested to descend into `InstallablePackage::install_one` and its inner guts. That would enable getting much more accurate information relatively to the usual non-dry-run operation. However, it looked a bit too complicated for a reviewable first step and also not too orthogonal with the `build-plan` and `unit-graph` features. The plan I adopted in this attempt is therefore to:
 * only handle the CLI flag in `cargo_install::install`;
 * basically only use the information provided by `InstallablePackage::new` and other easily-retrievable ones;
 * put itself between that and the start of the compilation to cut it entirely short;
 * display status messages indicating which very broad actions would be performed, but not much more.

As a quick preview:
```console
$ cargo run --bin cargo -- -Z unstable-options install --dry-run cargo-liner gitoxide bat git-delta
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/cargo -Z unstable-options install --dry-run cargo-liner gitoxide bat git-delta`
    Updating crates.io index
     Ignored package `cargo-liner v0.6.0` is already installed, use --force to override
    Updating crates.io index
  Downloaded gitoxide v0.33.0
  Downloaded 1 crate (75.5 KB) in 0.18s
     Ignored package `bat v0.24.0` is already installed, use --force to override
     Summary dry-run
 Destination /home/pauldance/.cargo/bin
     Install gitoxide: to 0.33.0 using registry `crates-io`
     Upgrade git-delta: from 0.16.5 to 0.17.0 using registry `crates-io`
  Up-to-date cargo-liner
  Up-to-date bat
```

Several tests have been added in order to check that the usual combinations of the flag with `--git`, `--path` and installation states behave as expected and print the desired outputs. See the end of `tests/testsuite/install.rs` for more details about this.

Review should be made easier using the few provided commits:
 * CLI change.
 * Implementation.
 * Tests.
 * Docs.

In particular, I might not be too fond of the wording I chose here and there. The way the information is displayed in a split manner using several new status kinds might not be the best either considering the rest of Cargo's output. Feel free to correct me on any of this.

### Additional information

This is very much a first implementation that could evolve in the near future depending on the precision of the emulation that is desirable. As-is, most of the operation is entirely avoided and only informational messages are printed instead. This feels to me enough for a first step. Being more precise would require emulating the compilation jobs or even their inner operations, at the cost of additional complexity.

Obviously, any feedback is welcome, especially if it is about the next steps potentially requiring the mentioned accuracy or if these should be done now instead.


PaulDance:
I also don't know exactly what to do about `check-version-bump` this time: should I simply bump the `cargo` library minor version?

tests/testsuite/install.rs
```diff
@@ -2623,3 +2623,598 @@ fn uninstall_running_binary() {
         )
         .run();
 }
+
+#[cargo_test(nightly, reason = "--dry-run is unstable")]
```
epage, line 2627: These do not need nightly, only to masquerade as nightly

tests/testsuite/install.rs
```diff
@@ -2623,3 +2623,598 @@ fn uninstall_running_binary() {
...
+fn dry_run_freshinstall_one() {
```
epage, line 2628: Please add these tests in their own commit at the start, without the dry-run flag, so we can compare what the dry-run flag does to the output

epage, line 2628: `freshinstall` is two words

crates/cargo-test-support/src/compare.rs
```diff
@@ -222,6 +222,12 @@ fn substitute_macros(input: &str) -> String {
         ("[REPLACING]", "   Replacing"),
         ("[UNPACKING]", "   Unpacking"),
         ("[SUMMARY]", "     Summary"),
+        ("[DESTINATION]", " Destination"),
+        ("[INSTALL]", "     Install"),
+        ("[UPGRADE]", "     Upgrade"),
+        ("[DOWNGRADE]", "   Downgrade"),
+        ("[UPTODATE]", "  Up-to-date"),
+        ("[REINSTALL]", "  Re-install"),
```
epage, line 230: imo the dry-run should generally match the regular behavior.

If we want to improve the `cargo install` output, that is a change on its own to be evaluated.

src/cargo/ops/cargo_install.rs
```diff
@@ -651,11 +653,59 @@ pub fn install(
             true,
             current_rust_version.as_ref(),
         )?;
-        let mut installed_anything = true;
-        if let Some(installable_pkg) = installable_pkg {
-            installed_anything = installable_pkg.install_one()?;
+        if dry_run {
+            match installable_pkg {
+                Some(installable_pkg) => dry_run_report(
```
epage, line 658: See https://github.com/rust-lang/cargo/issues/11123#issuecomment-2004139099

src/bin/cargo/commands/install.rs
```diff
@@ -74,6 +74,13 @@ pub fn cli() -> Command {
             "list",
             "List all installed packages and their versions",
         ))
+        .arg(
+            flag(
+                "dry-run",
+                "Display what would be installed without actually performing anything (unstable)",
+            )
+            .short('n'),
+        )
```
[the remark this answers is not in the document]
PaulDance: Ah yes, didn't see that one, thanks!

tests/testsuite/install.rs
PaulDance: Just to be clear, you want a copy of each test where only the flag is removed and the output adapted in consequence, right? Also, when you say "add these tests in their own commit at the start", are you referring to the already-written tests, the "copies" that need writing or all of them?

epage: Each test that exists in this PR should be added in their own commit, the first one.  The `--dry-run` flag would be missing.  These tests would pass.

The next commit in the PR would update the tests to add `--dry-run` and update the output accordingly.

PaulDance: OK, I see.

src/cargo/ops/cargo_install.rs
PaulDance: If this is indeed required, I'll see what I can do to rework this and go into the more fine-grained compilation steps to have a proper dry-run. It might take some time though, as it is almost starting from scratch.

@rustbot author

PR #13605, opened 2024-03-19
core/compiler: report executable paths for all binary crates

dvdhrm:
Ensure the `compiler-artifact` message reports `executable: "..."` for all builds where the crate-type is set to `bin`.

Currently, only crates with a target-type of `bin` will report the executable path. This is unfortunate, since several other builds produce executables that are otherwise hard to distinguish from other artifacts due to the lack of file extension.

In particular, the following builds now properly report their executables:

 - Library targets built via `cargo rustc --crate-type bin`
 - Build scripts

Before they would report `executable: null` and thus make it very hard for callers to decide whether the produced file is an executable.

### DIFF
--- a/src/cargo/core/compiler/build_runner/mod.rs
+++ b/src/cargo/core/compiler/build_runner/mod.rs
@@ -312,11 +312,13 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
 
     /// Returns the executable for the specified unit (if any).
     pub fn get_executable(&mut self, unit: &Unit) -> CargoResult<Option<PathBuf>> {
-        let is_binary = unit.target.is_executable();
+        let crate_types = unit.target.rustc_crate_types();
+        let is_binary = crate_types.contains(&compiler::CrateType::Bin);
         let is_test = unit.mode.is_any_test();
         if !unit.mode.generates_executable() || !(is_binary || is_test) {
             return Ok(None);
         }
+
         Ok(self
             .outputs(unit)?
             .iter()
--- a/tests/testsuite/build.rs
+++ b/tests/testsuite/build.rs
@@ -4266,7 +4266,7 @@ fn compiler_json_error_format() {
                     "overflow_checks": true,
                     "test": false
                 },
-                "executable": null,
+                "executable": "[..]/build-script-build[EXE]",
                 "features": [],
                 "filenames": "{...}",
                 "fresh": $FRESH
--- a/tests/testsuite/metabuild.rs
+++ b/tests/testsuite/metabuild.rs
@@ -709,7 +709,7 @@ fn metabuild_json_artifact() {
         .with_json_contains_unordered(
             r#"
             {
-              "executable": null,
+              "executable": "[..]/metabuild-foo[EXE]",
               "features": [],
               "filenames": "{...}",
               "fresh": false,
--- a/tests/testsuite/pkgid.rs
+++ b/tests/testsuite/pkgid.rs
@@ -318,7 +318,7 @@ fn pkgid_json_message_metadata_consistency() {
   "profile": "{...}",
   "features": [],
   "filenames": "{...}",
-  "executable": null,
+  "executable": "[..]",
   "fresh": false
 }
 
--- a/tests/testsuite/rustc.rs
+++ b/tests/testsuite/rustc.rs
@@ -270,6 +270,40 @@ fn build_with_crate_types_for_foo() {
 }
 
 #[cargo_test]
+fn build_lib_with_crate_type_bin() {
+    let p = project().file("src/lib.rs", "fn main() {}").build();
+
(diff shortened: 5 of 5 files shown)


epage:
Generally, I'd recommend creating an issue first for us to focus the discussion on the problem you are facing and what the potential solutions are.

Could you elaborate on why this is needed?

dvdhrm:
> Could you elaborate on why this is needed?

We build application bundles for a range of operating systems, including `macOS Bundles`, `iOS Bundles`, `Windows msix`, `Android APK/Bundles`, `...`. We post-process the artifacts generated by Cargo. We use `cargo metadata` to assemble the dependency tree that makes up a build, and then filter the artifacts reported by `cargo build --message-format=json` based on this tree. The artifacts are then post-processed (if needed) and assembled in the final bundle.

This setup keeps the Cargo package as the center piece and using as much metadata from the unmodified Cargo setup as possible, with minor annotations required as package-metadata.

By optionally using the unstable Cargo `bindeps` feature, we can even build multiple shared libraries from a single Cargo package, thus requiring no non-standard configuration to get exactly the artifacts needed for the respective bundle.

Unfortunately, we face 2 issues which we have to work around with rather ugly heuristics:

1) Some OSs require the entry-point as shared object, others as executable. A common solution is to add a simple `src/bin/runner.rs`, which invokes the application. The build-system then either builds the executable or the library, depending on which type is needed for the OS. Unfortunately, this requires implementing some dummy for systems that don't need it (you cannot guard the binary based on CFGs), and will build unneeded artifacts, which might have to be filtered and identified as dummies.
   
   To avoid all this, we instead always build via `cargo rustc --lib` but pass `--crate-type bin/cdylib` depending on which output is needed for the target. This merely requires putting an entry-point `fn main()` into the library, which can be nicely guarded by `cfgs` and implemented exactly for the targets that need it.
   
   Unfortunately, Cargo does not report such executables as `executable`, which this PR tries to fix, so our build-system can easily identify the artifact. Otherwise, we have to guess which of the artifacts is the executable (which we currently do, but would like to avoid).

2) Cargo can report a wide range of build-artifacts without any classification other than the `executable` key. We currently have to filter out files like `.rmeta`, `.d`, `.a`, `...` to ensure we copy the right artifacts to the correct destination location. Ideally, we would tell Cargo which artifacts we want, and it would only produce those. But our current heuristic simply filters based on target-type/crate-type and file-extension. It works fine, but feels like it guesses information that Cargo already has.
   
   Unfortunately, executables are quite hard to filter, given they lack suitable file-extensions on many platforms (or might even have random file-extensions if the `filename=` override is used). Hence, we rely on the `executable` key to tell us about them. Without this, we have to guess it based on lack of known file-extension for each target-type/crate-type.
   
   Note that we already filter `custom-build` or `proc-macro` targets, so we would never copy their artifacts, even if `executable` is set. I just thought it would be a nice addition to report those as well.

Given the documentation of the `executable` key, I thought it was an oversight that it did not report executables in all scenarios. Hence, I filed this mostly as a bug-fix.

What exactly are your concerns? Is this about confusing existing tools which expect a single build to only ever report a single executable? I am not entirely sure how the key is used by external tools. We always parse the entire metadata before the build, so we can filter based on target-types, but I guess if you do not do this, it might be confusing to see multiple `executable` keys in the output.

Alternatively, I could reduce the PR to only report executables for `--lib --crate-type=bin` builds?

epage:
I feel there is a problem you are having that we need to dig into and better understand.  Would you be willing to move this conversation to an Issue?  I've found Issues are much better for discussions of problems and proposed solutions while PRs are better for the implementation.  Part of this stems from PRs mixing problem/solution/implementation and part from the fact that there can be multiple PRs for one Problem and that can split the problem/solution discussion, making it harder to track.

dvdhrm:
Sure! Filed as #13612.

weihanglo:
Converted to draft for now until we have something from #13612.

epage:
The right place at the moment for this is #13612.

epage:
Closing, pending resolution of the discussion at #13612.

dvdhrm:
#13612 has not gotten any serious consideration from cargo developers since its creation almost 2 years ago. I find it particularly rude to then also close suggested solutions to the problem. How can public opinion be formed if maintainers do not respond to issues, and on top close suggested solutions to the issue?

epage:
I can understand not seeing progress and seeing this closed can be frustrating.  We've slowly been working through our ["needs team input" issues](https://github.com/rust-lang/cargo/issues?q=state%3Aopen%20label%3A%22S-needs-team-input%22) which takes time.  There are currently only 97.  We have 233 issues still in triage.

However, I'm not seeing what good leaving a PR ignored for a year does for anyone while waiting on discussing #13612.  We have had a [long standing policy](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo) discouraging PRs without an agreed-to design.

Issue #11259, opened 2022-10-19
cargo init with existing lib source file with hyphen in name makes bad Cargo.toml


BartMassey:
### Problem

(That is an *ugly* title. Improvements welcome.)

#### Background

`cargo init` has an obscure but convenient feature for small programs. Consider the following:

```sh
$ mkdir small
$ cd small
$ touch small.rs
$ cargo init --lib
```

This will produce a `Cargo.toml` that looks like

```toml
[package]
name = "small"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]

[lib]
name = "small"
path = "small.rs"
```

It will *not* produce `src/lib.rs` in this case: everything is self-contained in a single directory. I find this quite convenient.

(Note that for this to work, the directory name and the source file name have to be spelled the same.)

#### Hyphens

OK, let's try this with a different name.

```sh 
$ mkdir dashed-name
$ cd dashed-name
$ touch dashed-name.rs
$ cargo init --lib
```

We get a warning message this time:

```
warning: compiling this new package may not work due to invalid workspace configuration

failed to parse manifest at `/tmp/dashed-name/./Cargo.toml`

Caused by:
  library target names cannot contain hyphens: dashed-name
     Created library package
```

Sure enough, the generated `Cargo.toml` now looks like this.

```toml
[package]
name = "dashed-name"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]

[lib]
name = "dashed-name"
path = "dashed-name.rs"
```

And sure enough, trying to build using this will fail.

```
error: failed to parse manifest at `/tmp/dashed-name/Cargo.toml`

Caused by:
  library target names cannot contain hyphens: dashed-name
```

### Steps

See above.

### Possible Solution(s)

In order of preference:

1. Omit `lib.name` from the generated declaration. It is not needed for libs anyhow. This will cause the hyphens to be defaulted to underscores when deriving `lib.name` from `package.name`.

2. Replace hyphens with underscores when generating `lib.name`.

3. Just get rid of this whole defaulted-source feature of `cargo init`. I really like it, and would prefer that it stay, but it may be more trouble than it's worth.

### Notes

Vaguely relevant issues: #2775, #6827, #9333 (fixed, closed).

[...]

PR #13606, opened 2024-03-19
cargo/init: avoid target.name assignments if possible

dvdhrm:
Make `cargo init` skip `target.name` assignments if the default value is suitable.

Currently, all paths of `cargo init` will set target-names to the package-name, ensuring that a suitable target-name is set. This is required for all targets but libraries. Unfortunately, library-names have more restrictions than other targets. For example, creating a library with dashes will lead to errors:

    mkdir foo-bar
    cd foo-bar
    touch foo-bar.rs
    cargo init --lib

Fortunately, target-names for libraries are inferred from the package-name. Hence, by omitting the target-name, a valid configuration will be produced.

Instead of adjusting `SourceFileInformation::target_name` to use `Option<String>`, this commit strips the field completely, since all callers set it to the package-name.

Fixes: #11259

### DIFF
--- a/src/cargo/ops/cargo_new.rs
+++ b/src/cargo/ops/cargo_new.rs
@@ -87,7 +87,6 @@ impl fmt::Display for NewProjectKind {
 
 struct SourceFileInformation {
     relative_path: String,
-    target_name: String,
     bin: bool,
 }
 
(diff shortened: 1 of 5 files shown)


weihanglo (approved):
Make sense. Thanks!

weihanglo:
Note for the change:

While I understand people may use `cargo new`/`cargo init` in automation processes, the output Cargo.toml is not intended to consume in that way. So this is not considered as a breaking change.

Issue #3537, opened 2017-01-12
Published crates should include LICENSE file


raphlinus:
I'd like for published crates to include the text of the LICENSE along with the code. For our use case (making a repo of vendored crates), the cargo-vendor tool is populating the vendor directory with lots of packages that are missing LICENSE. There are basically two reasons for this:

1. The LICENSE file is present in the source repository, in the same directory as the Cargo.toml config file, but the config file doesn't include the LICENSE in the list of files to be uploaded (because of an explicit "include" keyword). Examples include [dtoa](https://crates.io/crates/dtoa) and [serde](https://crates.io/crates/serde).

2. The LICENSE file is present in the source repo (typically in the repo root), but the crate is published from a subdirectory. Examples include [num-traits](https://crates.io/crates/num-traits) and [xi-core-lib](https://crates.io/crates/xi-core-lib).

There are (at least) two things "cargo publish" could do to mitigate this problem:

1. The publish tool could do a lint check that an appropriate license file is present. There's a small amount of client-side sanity-checking now (checking that a license file exists if specified), and I believe the crates.io server also checks the spdx string.

2. Automatically including LICENSE files from the repo root if publishing from a subdirectory.

I'll get around this for the time being by making my vendoring script do the latter at vendoring time, but I think it would be better if every stage in the chain precisely matched the exact contents.

joshtriplett:
There's a third possibility that needs to be covered: the crate source cites a well-known license and *intentionally* does not include a license file. That's not necessarily a bug.

If the crate *does* include a license file, I'm entirely in favor of having cargo make sure it gets included.

kelseasy:
And some licenses can be found in a `COPYING` file too. There are also `PATENTS` files that can change the meaning of a license to something else entirely. :(

joshtriplett:
This issue still exists.

Would someone be interested in contributing to this? In particular, making `cargo package` check for a file named "LICENSE" or "COPYING" (or a few other patterns) and automatically including it in the `.crate` file?

richardwhiuk:
cargo publish does do this doesn't it - but cargo package doesn't?

PR #13608, opened 2024-03-20
feat(add): Stabilize MSRV-aware version req selection

epage:

This is part of #9930 for rust-lang/rfcs#3537

This will make it easier to maintain an MSRV-compliant `Cargo.toml` but leaves validation up to the user as the resolver will pick newer versions.
This helps the MSRV-aware workflows enumerated in
rust-lang/rfcs#3537

As for determining if this is ready for stabilization:

By stabilizing this without the MSRV-aware resolver, this could be confusing to the workflow with an MSRV-compatible lockfile.
PR #13561 at least raises awareness of that discrepancy.
In general there was interest in the RFC discussions to stabilize this ASAP, regardless of what resolver direction we went.

There is an unresolved question on differences in the resolver vs `cargo add` for dealing with an unset `rust-version` (noted in the tracking issue). However, we are only stabilizing the `cargo add` side which is a very light two-way door as this is a UX-focused command and not a programmatic command. 

This hasn't gotten much end-user acceptance testing but, as its UX focused, that seems fine (light, two way door)

As such, this seems like it is ready to stabilize.

### DIFF
--- a/src/bin/cargo/commands/add.rs
+++ b/src/bin/cargo/commands/add.rs
@@ -87,7 +87,7 @@ Example uses:
 - Depend on crates with the same name from different registries"),
             flag(
                 "ignore-rust-version",
-                "Ignore `rust-version` specification in packages (unstable)"
+                "Ignore `rust-version` specification in packages"
             ),
         ])
         .arg_manifest_path_without_unsupported_path_tip()
@@ -206,14 +206,6 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
     let dependencies = parse_dependencies(gctx, args)?;
 
     let ignore_rust_version = args.flag("ignore-rust-version");
-    if ignore_rust_version && !gctx.cli_unstable().msrv_policy {
-        return Err(CliError::new(
-            anyhow::format_err!(
-                "`--ignore-rust-version` is unstable; pass `-Zmsrv-policy` to enable support for it"
-            ),
-            101,
-        ));
-    }
     let honor_rust_version = !ignore_rust_version;
 
     let options = AddOptions {
--- a/src/cargo/ops/cargo_add/mod.rs
+++ b/src/cargo/ops/cargo_add/mod.rs
@@ -608,7 +608,7 @@ fn get_latest_dependency(
                 )
             })?;
 
-            if gctx.cli_unstable().msrv_policy && honor_rust_version {
+            if honor_rust_version {
                 let (req_msrv, is_msrv) = spec
                     .rust_version()
                     .cloned()
--- a/src/doc/man/cargo-add.md
+++ b/src/doc/man/cargo-add.md
@@ -141,11 +141,6 @@ which enables all specified features.
 
 {{#option "`--ignore-rust-version`" }}
 Ignore `rust-version` specification in packages.
-
-This option is unstable and available only on the
-[nightly channel](https://doc.rust-lang.org/book/appendix-07-nightly-rust.html)
-and requires the `-Z unstable-options` flag to enable.
-See <https://github.com/rust-lang/cargo/issues/5579> for more information.
 {{/option}}
 
 {{/options}}
--- a/src/doc/man/generated_txt/cargo-add.txt
+++ b/src/doc/man/generated_txt/cargo-add.txt
@@ -131,12 +131,6 @@ OPTIONS
        --ignore-rust-version
            Ignore rust-version specification in packages.
 
-           This option is unstable and available only on the nightly channel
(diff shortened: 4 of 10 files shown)


epage:
```
Err:10 https://ppa.launchpadcontent.net/ubuntu-toolchain-r/test/ubuntu jammy InRelease
  503  Service Unavailable [IP: 185.125.190.80 443]
```

epage:
While the details of this still need a review, I figure we could do the high level "should we stabilize" conversation in parallel.

Before, when a user ran `cargo add foo` (no version req specified), we'd always make a version req out of the latest version.

With this change, we still give priority to the version req the user selected.  However, if they did not specify one, instead of always picking the latest version to make a version req from, instead we'll filter out versions incompatible with `package.rust-version` (falling back to `rustc -V` if unspecified).  If we select a non-latest version, we'll warn the user.  If we can't select a version, we will error.  The user can force the `add` to work by specifying the version req or passing `--ignore-rust-version`.

This does not include `--update-rust-version`.  I was planning on implementing that for both the resolver and cargo-add at the same time.

Differences with the resolver
- This gives a hard error, rather than falling back to an incompatible version.  For an interactive command, I think this workflow is likely better. 
- This treats dependencies without a `package.rust-version` as compatible.  Being interactive, I think this is likely the best choice.  This is also a light two way door and we might want to consider changing the resolver to this behavior.  This discrepancy is being tracked in #9930.

This is part of [RFC #3537](https://rust-lang.github.io/rfcs/3537-msrv-resolver.html). being tracked in #9930.  The intention with the RFC is that we could stabilize each individual piece, rather than waiting for everything to be ready.

@rfcbot fcp merge

weihanglo (approved):
Implementation looks good.

---

To whom wants to test the change, simply run:

zsh

```zsh
fpath+=$PWD/src/etc
autoload -Uz compinit
compinit

cargo add --i<tab>
cargo add --<tab> # to review the flag help text
```

bash

```bash

. ./src/etc/cargo.bashcomp.sh

cargo add --i<tab>
```

weihanglo:
Tiny UX question. Should users be aware of an older version of a dependency was chosen because of MSRV? Do we want to print something for that situation?

src/bin/cargo/commands/add.rs
weihanglo, line 1: Tiny UX question. Should users be aware of an older version of a dependency was chosen because of MSRV? Do we want to print something for that situation?

epage: We print
```
    warning: ignoring rust-version-user@0.2.1 (which requires rustc 1.72) to maintain cargo-list-test-fixture's rust-version of 1.70
          Adding rust-version-user v0.1.0 to dependencies
         Locking 2 packages
          Adding rust-version-user v0.1.0 (latest: v0.2.1)
```
See https://raw.githubusercontent.com/rust-lang/cargo/master/tests/testsuite/cargo_add/rust_version_older/stderr.term.svg

So you get a message for both the version req and when locking.

weihanglo: Thanks! Didn't see that. I must be reading the wrong commit…

ehuss:
I'm not quite clear on how this works with the resolver. What happens if:
- `foo` 1.0.0 is compatible with my msrv
- `foo` 1.0.1 is not
- run `cargo add foo`

Does it add `foo = "1.0.0"`? What happens with `Cargo.lock`, is it 1.0.0 or 1.0.1?

epage:
This is independent of the resolver.  The version requirement will be MSRV-compatible (`foo = "1.0.0"`) while `Cargo.lock` will be set to 1.0.1.

For users who don't track MSRV-comptible dependencies in their lockfile, this is what they expect and things will be fine (until we change the resolver).

For users who do track MSRV-compatible dependencies, this removes one extra step for them.  They will have a valid `Cargo.toml` and will just need to work on their `Cargo.lock`.  They already know what to lock to for the first dependency.  They then only need to search the docs for any transitive dependencies.  They can also choose to use nightly to resolve as they want (either `cargo +nightly -Zmsrv-policy add ...` or `git checkout HEAD -- Cargo.lock && cargo +nightly check -Zmsrv-policy`).

So while this isn't a complete solution for everyone, until the resolver is stable, it does improvements for everyone.

epage:
@ehuss this does get me thinking that maybe we could update the 
```
Locking 3 packages
```
message to
```
Locking 3 packages to latest versions
Locking 3 packages to latest rust-version compatible versions
Locking 3 packages to lowest versions
```
Depending on which resolve behavior is being invoked.

weihanglo:
The FCP is complete. Thank you to everyone involved.

@bors r+

Issue #6198, opened 2018-10-21
"Package collision in lockfile" using UNC/root local device/etc paths


rpjohnst:
**Problem**
When invoking Cargo with the current directory or manifest path given as one of Windows' more "exotic" kinds of paths, such as a `\\?\` "root local device path," internal path comparisons can fail and the build aborts. Ideally, Cargo would be able to handle these paths.

**Steps**
One way to reproduce the bug is with a directory structure like this:
```
bug/
  main/
    Cargo.toml
    src/lib.rs
  dep_a/
    Cargo.toml
    src/lib.rs
  dep_b/
    Cargo.toml
    src/lib.rs
```

`main` depends, via `path = "../dep_a"` and `path = "../dep_b"`, on both `dep_a` and `dep_b`. `dep_a` also depends, via `path = "../dep_b"`, on `dep_b`.

Here is a zip file containing the structure described above: [path-bug.zip](https://github.com/rust-lang/cargo/files/2502476/path-bug.zip)

Then, the bug can be triggered by running this command:

```cmd
cargo build --manifest-path \\?\C:\full\path\to\bug\main\Cargo.toml
```

It produces an error message like this:

```cmd
error: package collision in the lockfile: packages dep_b v0.1.0 (C:\full\path\to\bug\main\../dep_b) and dep_b v0.1.0 (C:\full\path\to\bug\dep_b) are different, but only one can be written to lockfile unambigiously
```

`wasm-pack` hits this issue because it sets the current directory to a canonicalized `.` when invoking Cargo. See https://github.com/rustwasm/wasm-pack/issues/380, https://github.com/rustwasm/wasm-pack/issues/413, and an attempt to work around this problem in https://github.com/rustwasm/wasm-pack/pull/389.

**Notes**
I most recently reproduced this with Cargo version `cargo 1.31.0-nightly (5dbac9888 2018-10-08)`, on Windows 10.

rpjohnst:
It is a big laborious to recreate, isn't it? 😅

Here you go: [path-bug.zip](https://github.com/rust-lang/cargo/files/2502476/path-bug.zip)

Eh2406:
No I am just feeling brain dead from Rust Belt Rust, and wanted a supportive response that did not involve me thinking for a while. Sorry.

[...]

PR #13610, opened 2024-03-20
refactor(vendor): tiny not important refactors

weihanglo:

Did some tiny refactors in cargo vendor I found when working on other stuff.

~~I believe `std::path::Path` should be quite cross-platform compatible, as [it calls into something like [`GetFullPathNameW`](
https://github.com/rust-lang/rust/blob/e760daa6a729b3d52a38804e9766f7d89dc27357/library/std/src/sys/path/windows.rs#L280-L287) eventually when being used. I might be wrong though.~~

See https://github.com/rust-lang/cargo/pull/13610#discussion_r1532741537

### DIFF
--- a/src/cargo/ops/vendor.rs
+++ b/src/cargo/ops/vendor.rs
@@ -133,7 +133,7 @@ fn sync(
                 continue;
             }
             if let Ok(pkg) = packages.get_one(pkg) {
-                drop(fs::remove_dir_all(pkg.manifest_path().parent().unwrap()));
+                drop(fs::remove_dir_all(pkg.root()));
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/cargo/ops/vendor.rs
```diff
@@ -342,13 +342,7 @@ fn cp_sources(
...
-        // Join pathname components individually to make sure that the joined
-        // path uses the correct directory separators everywhere, since
-        // `relative` may use Unix-style and `dst` may require Windows-style
-        // backslashes.
-        let dst = relative
-            .iter()
-            .fold(dst.to_owned(), |acc, component| acc.join(&component));
+        let dst = dst.join(relative);
 
         paths::create_dir_all(dst.parent().unwrap())?;
```
epage, line 347: I assume this comment is relevant here

> I believe std::path::Path should be quite cross-platform compatible, as [it calls into something like [GetFullPathNameW](https://github.com/rust-lang/rust/blob/e760daa6a729b3d52a38804e9766f7d89dc27357/library/std/src/sys/path/windows.rs#L280-L287) eventually when being used. I might be wrong though.

I assume this was done for a reason though so I'm a bit cautious and would like input from others.

weihanglo: Seems like https://github.com/rust-lang/cargo/issues/6198 is relevant?

https://github.com/rust-lang/rust/blob/e760daa6a729b3d52a38804e9766f7d89dc27357/library/std/src/sys/path/windows.rs#L253

If it already contains `\\?\` then `GetFullPathNameW` won't be called. I'll drop the commit instead.

weihanglo: This evil behavior is deeply rooted in Cargo 😞.

PR #13614, opened 2024-03-21
Fix publish script due to crates.io CDN change

ehuss:
Now that crates.io goes directly to the CDN, this has changed the HTTP response code when a crate is not found (from 404 to 403). This updates the publish script to handle more error codes to assume something isn't published. This also updates it to go directly to the endpoint instead of through the crates.io redirect.

### DIFF
--- a/publish.py
+++ b/publish.py
@@ -31,11 +31,15 @@ TO_PUBLISH = [
 
 
 def already_published(name, version):
+    url = f'https://static.crates.io/crates/{name}/{version}/download'
     try:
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
Thanks.

Could we add some print for each crate to help future debugging?

weihanglo:
`r=weihanglo` whenever you want to proceed or add some print.

ehuss:
Good idea, I added a print.

@bors r=weihanglo

Issue #12826, opened 2023-10-14
Tracking Issue for remove-implicit-feature RFC 3491


ehuss:
### Summary

RFC: [#3491](https://github.com/rust-lang/rfcs/pull/3491)
Implementation: TODO
Documentation: TODO

Removes the implicit feature created for optional dependencies, requiring them to be explicitly stated in the `[features]` table with `dep:foo` syntax. This will be a hard error in the 2024 edition, and a warning in prior editions.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

ehuss:
@epage Do you want to take on implementation of this, or know someone who will? Or, if not, do you want to write mentoring instructions? I can probably do that if you don't.

epage:
@Muscraft is taking this on atm as his test case for #12235.  We should pick a time to re-evaluate and, if that isn't ready in time, we pivot and hard code all of this where it is needed.

epage:
@Muscraft and I have been talking about how to implement this.

Instead of erroring in the 2024 Edition that an implicit feature was created, we are looking at treating it as an "unused optional dependency" warning.  However, to truly be unused, no implicit feature should be created.

So how do we avoid that implicit feature?

### Option 1: We strip unused optional dependencies before generating a `Summary` / before publish

This will "just work".  All past and future versions of Cargo will not be able to access an implicit feature.

This would be an impediment to running other lints on unused optional dependencies because we'd be throwing out the information we'd use for linting.

This means that unused optional dependencies would not affect dependency resolution and not be included in `Cargo.lock`, like `workspace.dependencies` and unlike unused transitive dependencies from weak features (#10801).

### Option 2: We include the `edition` in the summary

We then use `edition` to decide whether optional dependencies should have implicit features created.

Maybe other future work could benefit from this?

This would be quite a bit of work, requiring crates.io changes and communicating out changes required for third-party registries to support Edition 2024.  But we only have 2 weeks left and coordinating across teams within that time span might be difficult.

### Option 3: We punt and have this warning be an error

Nuff said

### Option 4: When generating the Summary / published `Cargo.toml`, we create `foo = ["dep:foo"]` features for pre-2024 Edition

[...]

Issue #3491, opened 2017-01-04
workspaces: All building all members of the workspace


sdroege:
Something like "cargo build" in the root of the workspace would seem useful, which would then build all members of the workspace. Similar to how "cargo clean" currently cleans the workspace.

steveklabnik:
This should be `--all`.

alexcrichton:
Yeah we've got `cargo test --all` now on Nightly so it should be easy enough to add `cargo build --all`!

sdroege:
Basically something similar to 62c0497912f6c424a4a68d7af504f7943fad9864 then, just doing the same for build? Are there any other commands that could use some --all'ification?
I might take a look into that later, would be nice to replace my Makefile doing just that with a cargo command

alexcrichton:
Now that all of the support was landed with `cargo test --all` it may just involve [copying these lines](https://github.com/rust-lang/cargo/blob/afe53bdf7043c2ea8ed3cd0fb8c778ff11dccf4a/src/bin/test.rs#L119-L123) into [`build.rs`](https://github.com/rust-lang/cargo/blob/afe53bdf7043c2ea8ed3cd0fb8c778ff11dccf4a/src/bin/build.rs#L87). We could add `cargo doc` as well, although I'd want to just tread purposefully here to ensure we don't expand it *too* quickly

sdroege:
Thanks, it was almost just that (I've submitted a pull request). For "cargo doc" a little bit more is needed, but if you think that's useful to have (it seems useful to me), I can prepare a pull request for that too.

alexcrichton:
`cargo doc --all` sounds great to me! I'm sure it'll be requested sooner rather than later :)

sdroege:
Ok, I'll get to that once "build --all" is ready to be merged

frewsxcv:
Can this be closed now?

frewsxcv:
Should the default Travis CI Rust script use `--all` by default once this lands in stable?

sdroege:
I would wait until https://github.com/rust-lang/cargo/issues/3515 landed before closing

PR #13621, opened 2024-03-22
feat: Add a basic linting system

Muscraft:
This PR adds a basic linting system, the first step towards [User control over cargo warnings](https://github.com/rust-lang/cargo/issues/12235). To demonstrate the system, a lint for #12826 is added. It supports controlling cargo lints via the `[lints.cargo]` table. Lints can be controlled either by a group or individually.

This is meant to lay down some fundamental infrastructure for the whole linting system and is not meant to be fully featured. Over time, features will be added that will help us reach a much more solid state.

### DIFF
--- a/src/cargo/core/workspace.rs
+++ b/src/cargo/core/workspace.rs
@@ -24,10 +24,12 @@ use crate::sources::{PathSource, CRATES_IO_INDEX, CRATES_IO_REGISTRY};
 use crate::util::edit_distance;
 use crate::util::errors::{CargoResult, ManifestError};
 use crate::util::interning::InternedString;
+use crate::util::lints::check_implicit_features;
 use crate::util::toml::{read_manifest, InheritableFields};
 use crate::util::{context::ConfigRelativePath, Filesystem, GlobalContext, IntoUrl};
 use cargo_util::paths;
 use cargo_util::paths::normalize_path;
+use cargo_util_schemas::manifest;
 use cargo_util_schemas::manifest::RustVersion;
 use cargo_util_schemas::manifest::{TomlDependency, TomlProfiles};
 use pathdiff::diff_paths;
@@ -1095,11 +1097,14 @@ impl<'gctx> Workspace<'gctx> {
 
     pub fn emit_warnings(&self) -> CargoResult<()> {
         for (path, maybe_pkg) in &self.packages.packages {
+            let path = path.join("Cargo.toml");
+            if let MaybePackage::Package(pkg) = maybe_pkg {
+                self.emit_lints(pkg, &path)?
+            }
             let warnings = match maybe_pkg {
                 MaybePackage::Package(pkg) => pkg.manifest().warnings().warnings(),
                 MaybePackage::Virtual(vm) => vm.warnings().warnings(),
             };
-            let path = path.join("Cargo.toml");
             for warning in warnings {
                 if warning.is_critical {
                     let err = anyhow::format_err!("{}", warning.message);
@@ -1119,6 +1124,30 @@ impl<'gctx> Workspace<'gctx> {
             }
         }
         Ok(())
+    }
+
+    pub fn emit_lints(&self, pkg: &Package, path: &Path) -> CargoResult<()> {
+        let mut error_count = 0;
+        let lints = pkg
+            .manifest()
+            .resolved_toml()
+            .lints
+            .clone()
+            .map(|lints| lints.lints)
+            .unwrap_or(manifest::TomlLints::default())
+            .get("cargo")
+            .cloned()
+            .unwrap_or(manifest::TomlToolLints::default());
+
+        check_implicit_features(pkg, &path, &lints, &mut error_count, self.gctx)?;
+        if error_count > 0 {
+            Err(crate::util::errors::AlreadyPrintedError::new(anyhow!(
+                "encountered {error_count} errors(s) while running lints"
+            ))
+            .into())
+        } else {
+            Ok(())
+        }
     }
 
     pub fn set_target_dir(&mut self, target_dir: Filesystem) {
--- a/src/cargo/util/lints.rs
+++ b/src/cargo/util/lints.rs
@@ -1,0 +1,226 @@
+use crate::core::FeatureValue::Dep;
+use crate::core::{Edition, FeatureValue, Package};
+use crate::util::interning::InternedString;
+use crate::{CargoResult, GlobalContext};
+use annotate_snippets::{Level, Renderer, Snippet};
+use cargo_util_schemas::manifest::{TomlLintLevel, TomlToolLints};
+use pathdiff::diff_paths;
+use std::collections::HashSet;
+use std::ops::Range;
+use std::path::Path;
+use toml_edit::ImDocument;
+
+fn get_span(document: &ImDocument<String>, path: &[&str], get_value: bool) -> Option<Range<usize>> {
+    let mut table = document.as_item().as_table_like().unwrap();
+    let mut iter = path.into_iter().peekable();
+    while let Some(key) = iter.next() {
+        let (key, item) = table.get_key_value(key).unwrap();
+        if iter.peek().is_none() {
+            return if get_value {
+                item.span()
+            } else {
+                let leaf_decor = key.dotted_decor();
+                let leaf_prefix_span = leaf_decor.prefix().and_then(|p| p.span());
+                let leaf_suffix_span = leaf_decor.suffix().and_then(|s| s.span());
+                if let (Some(leaf_prefix_span), Some(leaf_suffix_span)) =
+                    (leaf_prefix_span, leaf_suffix_span)
+                {
+                    Some(leaf_prefix_span.start..leaf_suffix_span.end)
+                } else {
+                    key.span()
+                }
+            };
+        }
+        if item.is_table_like() {
+            table = item.as_table_like().unwrap();
+        }
+        if item.is_array() && iter.peek().is_some() {
+            let array = item.as_array().unwrap();
+            let next = iter.next().unwrap();
+            return array.iter().find_map(|item| {
+                if next == &item.to_string() {
+                    item.span()
+                } else {
+                    None
+                }
+            });
+        }
+    }
+    None
+}
+
+/// Gets the relative path to a manifest from the current working directory, or
+/// the absolute path of the manifest if a relative path cannot be constructed
+fn rel_cwd_manifest_path(path: &Path, gctx: &GlobalContext) -> String {
+    diff_paths(path, gctx.cwd())
+        .unwrap_or_else(|| path.to_path_buf())
+        .display()
+        .to_string()
+}
+
+#[derive(Copy, Clone, Debug)]
+pub struct LintGroup {
+    pub name: &'static str,
+    pub default_level: LintLevel,
+    pub desc: &'static str,
+    pub edition_lint_opts: Option<(Edition, LintLevel)>,
+}
+
+const RUST_2024_COMPATIBILITY: LintGroup = LintGroup {
+    name: "rust-2024-compatibility",
+    default_level: LintLevel::Allow,
+    desc: "warn about compatibility with Rust 2024",
+    edition_lint_opts: Some((Edition::Edition2024, LintLevel::Deny)),
+};
+
+#[derive(Copy, Clone, Debug)]
+pub struct Lint {
+    pub name: &'static str,
+    pub desc: &'static str,
+    pub groups: &'static [LintGroup],
+    pub default_level: LintLevel,
+    pub edition_lint_opts: Option<(Edition, LintLevel)>,
+}
+
+impl Lint {
+    pub fn level(&self, lints: &TomlToolLints, edition: Edition) -> LintLevel {
+        let level = self
+            .groups
+            .iter()
+            .map(|g| g.name)
+            .chain(std::iter::once(self.name))
+            .filter_map(|n| lints.get(n).map(|l| (n, l)))
+            .max_by_key(|(n, l)| (l.priority(), std::cmp::Reverse(*n)));
+
+        match level {
+            Some((_, toml_lint)) => toml_lint.level().into(),
+            None => {
+                if let Some((lint_edition, lint_level)) = self.edition_lint_opts {
+                    if edition >= lint_edition {
+                        return lint_level;
+                    }
+                }
+                self.default_level
+            }
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum LintLevel {
+    Allow,
+    Warn,
+    Deny,
+    Forbid,
+}
+
+impl LintLevel {
+    pub fn to_diagnostic_level(self) -> Level {
+        match self {
+            LintLevel::Allow => Level::Note,
+            LintLevel::Warn => Level::Warning,
+            LintLevel::Deny => Level::Error,
+            LintLevel::Forbid => Level::Error,
+        }
+    }
+}
+
+impl From<TomlLintLevel> for LintLevel {
+    fn from(toml_lint_level: TomlLintLevel) -> LintLevel {
+        match toml_lint_level {
+            TomlLintLevel::Allow => LintLevel::Allow,
+            TomlLintLevel::Warn => LintLevel::Warn,
+            TomlLintLevel::Deny => LintLevel::Deny,
+            TomlLintLevel::Forbid => LintLevel::Forbid,
+        }
+    }
+}
+
+/// By default, cargo will treat any optional dependency as a [feature]. As of
+/// cargo 1.60, these can be disabled by declaring a feature that activates the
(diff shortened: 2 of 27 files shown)


src/cargo/util/lints.rs
```diff
@@ -0,0 +1,241 @@
...
+    // Add implicit features for optional dependencies if they weren't
+    // explicitly listed anywhere.
+    let explicitly_listed: HashSet<_> = map
+        .values()
+        .flatten()
+        .filter_map(|fv| match fv {
+            Dep { dep_name } => Some(*dep_name),
+            _ => None,
+        })
+        .collect();
```
weihanglo, line 198: Two hash sets are good enough?

```suggestion
    let manifest = pkg.manifest();
    let user_defined_features = manifest.resolved_toml().features();
    let features = user_defined_features.map_or(HashSet::new(), |f| {
        f.keys().map(|k| InternedString::new(&k)).collect()
    });
    // Add implicit features for optional dependencies if they weren't
    // explicitly listed anywhere.
    let explicitly_listed = user_defined_features.map_or(HashSet::new(), |f| {
        f.values()
            .flatten()
            .filter_map(|v| match FeatureValue::new(v.into()) {
                Dep { dep_name } => Some(dep_name),
                _ => None,
            })
            .collect()
    });
```

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,241 @@
...
+            || features.contains_key(&dep_name_in_toml)
+            || explicitly_listed.contains(&dep_name_in_toml)
+        {
+            continue;
+        }
+
+        if features.contains_key(&dep_name_in_toml) || explicitly_listed.contains(&dep_name_in_toml)
+        {
+            continue;
+        }
```
weihanglo, line 211: Redundant

```suggestion
        if features.contains_key(&dep_name_in_toml) || explicitly_listed.contains(&dep_name_in_toml)
        {
            continue;
        }
```

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,241 @@
...
+                } else {
+                    None
+                }
+            });
+        }
+    }
+    None
+}
+
+fn manifest_path(path: &Path, gctx: &GlobalContext) -> String {
```
weihanglo, line 56: Understand what you mean, but this should be path relative to cwd, not exactly manifest_path?

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,241 @@
...
+        match self {
+            TomlLintLevel::Allow => LintLevel::Allow,
+            TomlLintLevel::Warn => LintLevel::Warn,
+            TomlLintLevel::Deny => LintLevel::Deny,
+            TomlLintLevel::Forbid => LintLevel::Forbid,
+        }
+    }
+}
+
+const IMPLICIT_FEATURES: Lint = Lint {
```
weihanglo, line 144: Could we have a brief doc comment with examples explaining this lint?
(And later we can parse it and generate a website 🎉)

tests/testsuite/lints.rs
```diff
@@ -749,3 +749,163 @@ pub const Ĕ: i32 = 2;
...
+                [lints.cargo]
+                implicit-features = "allow"
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("check")
+        .masquerade_as_nightly_cargo(&["edition2024"])
+        .with_status(0)
```
weihanglo, line 821: default is zero.

```suggestion
```

src/cargo/util/toml/mod.rs
```diff
@@ -1454,7 +1454,7 @@ fn verify_lints(lints: Option<&manifest::TomlLints>) -> CargoResult<()> {
     };
 
     for (tool, lints) in lints {
-        let supported = ["rust", "clippy", "rustdoc"];
+        let supported = ["rust", "clippy", "rustdoc", "cargo"];
```
weihanglo, line 1457: Should we put lint system behind zflag or nightly-only in case of accidentally stabilization some lints? Though personally I want it out ASAP 😂

Alternatively we can have a field on `Lint` indicating stable/unstable or Rust version.

tests/testsuite/lints.rs
```diff
@@ -749,3 +749,163 @@ pub const Ĕ: i32 = 2;
         )
         .run();
 }
+
+#[cargo_test(nightly, reason = "edition2024 is not stable")]
```
weihanglo, line 753: If possible, I lean toward UI test for specific lints.

(I know this is more like testing the system)

weihanglo (commented):
🎉 🎉 🎉

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,241 @@
...
+pub struct Lint {
+    pub name: &'static str,
+    pub desc: &'static str,
+    pub groups: &'static [LintGroup],
+    pub default_level: LintLevel,
+    pub edition_lint_opts: Option<(Edition, LintLevel)>,
+}
+
+impl Lint {
+    pub fn level(&self, lints: &TomlToolLints, edition: Edition) -> LintLevel {
```
weihanglo, line 88: One thing worth thinking hard is that how to avoid future build failure if people set something like `-D cargo::all` and we introduce new cargo lints.

A possible way is having every new lint capped max to warning in the current edition, and relax in the next.

Muscraft: One way to potentially do this is to have each lint have a `version` field similar to [`Feature`](https://github.com/rust-lang/cargo/blob/master/src/cargo/core/features.rs#L518-L518), that is the version they are stabilized in. From there, we add a "dynamic" group `cargo:new-lints` that all lints stabilized in that version are apart of.  We then add documentation saying if a user doesn't want to break CI with `-D cargo::all`, they should have the command be:
```
cargo <cmd> -- -W cargo:new-lints -D cargo:all
```

I don't think this is perfect but it would work

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,241 @@
...
+                } else {
+                    IMPLICIT_FEATURES.default_level
+                }
+            }
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum LintLevel {
```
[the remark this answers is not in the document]
Muscraft: We are slightly different only missing [`force-warn`](https://doc.rust-lang.org/rustc/lints/levels.html#force-warn) according to their [docs about lint levels](https://doc.rust-lang.org/rustc/lints/levels.html). If you look at the [source code for `Level`](https://github.com/rust-lang/rust/blob/1447f9d38ca388ca178a544534b3cff72945fa1e/compiler/rustc_lint_defs/src/lib.rs#L179-L214), we are also missing `Expect`. The idea for `expect` was added in [RFC 2383](https://rust-lang.github.io/rfcs/2383-lint-reasons.html#expect-lint-attribute), and I am not sure that we need it.

tests/testsuite/features.rs
```diff
@@ -2246,3 +2246,39 @@ fn default_features_conflicting_warning() {
         )
         .run();
 }
+
+#[cargo_test(nightly, reason = "edition2024 is not stable")]
```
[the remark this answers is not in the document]
Muscraft: This would make the tests fail on stable but pass on nightly unless I am missing something

src/cargo/util/lints.rs
Muscraft: Please let me know if you like the doc comment. I pulled it from the summary of [RFC #3491](https://rust-lang.github.io/rfcs/3491-remove-implicit-features.html).

tests/testsuite/lints.rs
Muscraft: I switched to UI tests, let me know if you like the structure

src/cargo/util/lints.rs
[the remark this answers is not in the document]
Muscraft: I am unsure where is best to put this to be honest. I am not happy about it currently, but I think it is okay. I am also worried about splitting it into multiple files too early.

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,226 @@
...
+use crate::{CargoResult, GlobalContext};
+use annotate_snippets::{Level, Renderer, Snippet};
+use cargo_util_schemas::manifest::{TomlLintLevel, TomlToolLints};
+use pathdiff::diff_paths;
+use std::collections::HashSet;
+use std::ops::Range;
+use std::path::Path;
+use toml_edit::ImDocument;
+
+fn get_span(document: &ImDocument<String>, path: &[&str], get_value: bool) -> Option<Range<usize>> {
```
weihanglo, line 13: Forgot to review this part during last round...

Not sure if this kind of access still have the correct span, but at least it passes all new tests :)

```rust
fn get_span(
    document: &ImDocument<String>,
    paths: &[&str],
    get_value: bool,
) -> Option<Range<usize>> {
    let (last, paths) = paths.split_last()?;
    let mut container = document.as_item();
    for path in paths {
        container = container.get(path)?;
    }
    if get_value {
        container.get(last)?.span()
    } else {
        container.as_table()?.key(last)?.span()
    }
}
```

weihanglo:
As this is for linting system, the only blocker is an unstable flag for `[lints.cargo]`. Let me know if you are willing to work on it in this PR or later.

src/cargo/util/lints.rs
Muscraft: Sadly, I don't think this works for `Value::Array` as `item.get()` only works on [`Item::Table` and `Value::InlineTable`](https://github.com/toml-rs/toml/blob/bb92dc30c2a3786d583b6d4a27c345b31c232892/crates/toml_edit/src/index.rs#L36-L51) (I think)

weihanglo (approved):
Thanks! Now looks pretty great and we're ready to merge this 👍🏾

weihanglo:
This linting system is behind a new unstable flag `-Z cargo-lints`, with only a warning if not on nightly (non-blocking flag). Since the Cargo team is aware of the linting system Scott has been working on, I think we could merge it as-is without an FCP. Let me know if something goes wrong then.

@bors r+

src/cargo/util/lints.rs
epage: I don't think rustc or clippy maintain compatibility with `all`

src/cargo/util/lints.rs
epage, line 13: nit: please put the focus of the file (what comes first) on the API and not the implementation details

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,226 @@
...
+/// [RFC #3491]: https://rust-lang.github.io/rfcs/3491-remove-implicit-features.html
+const IMPLICIT_FEATURES: Lint = Lint {
+    name: "implicit-features",
+    desc: "warn about the use of unstable features",
+    groups: &[RUST_2024_COMPATIBILITY],
+    default_level: LintLevel::Allow,
+    edition_lint_opts: Some((Edition::Edition2024, LintLevel::Deny)),
+};
+
+pub fn check_implicit_features(
```
epage, line 160: imo its a bit messy to shove lint implementations in here just because they are lints.  This body should live where its called.

src/cargo/util/lints.rs
epage, line 13: These needs documentation, especially around how arrays are handled

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,226 @@
...
+                }
+            });
+        }
+    }
+    None
+}
+
+/// Gets the relative path to a manifest from the current working directory, or
+/// the absolute path of the manifest if a relative path cannot be constructed
+fn rel_cwd_manifest_path(path: &Path, gctx: &GlobalContext) -> String {
```
epage, line 54: ditto, this belongs at the end of the file

epage, line 54: `String` in this is suspect.  Should this have `display` in the name?

Or should we try to switch to `camino`?

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,226 @@
...
+            None => {
+                if let Some((lint_edition, lint_level)) = self.edition_lint_opts {
+                    if edition >= lint_edition {
+                        return lint_level;
+                    }
+                }
+                self.default_level
+            }
+        }
+    }
```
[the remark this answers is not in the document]
epage: Note: even if we don't support the group in `[lints]`, we'd need to be able to support it in the future with #8424.

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    for dep in manifest.dependencies() {
+        let dep_name_in_toml = dep.name_in_toml();
+        if !dep.is_optional()
+            || features.contains(&dep_name_in_toml)
+            || explicitly_listed.contains(&dep_name_in_toml)
+        {
+            continue;
+        }
+        if lint_level == LintLevel::Forbid || lint_level == LintLevel::Deny {
```
epage, line 197: Should we have an `is_error` function?

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,226 @@
...
+            *error_count += 1;
+        }
+        let level = lint_level.to_diagnostic_level();
+        let manifest_path = rel_cwd_manifest_path(path, gctx);
+        let message = level.title("unused optional dependency").snippet(
```
epage, line 202: This message doesn't make sense pre-2024

tests/testsuite/lints/implicit_features/edition_2021/mod.rs
```diff
@@ -0,0 +1,34 @@
+use cargo_test_support::prelude::*;
```
epage, line 1: None of our other UI tests nest this deeply

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,226 @@
...
+    pub default_level: LintLevel,
+    pub desc: &'static str,
+    pub edition_lint_opts: Option<(Edition, LintLevel)>,
+}
+
+const RUST_2024_COMPATIBILITY: LintGroup = LintGroup {
+    name: "rust-2024-compatibility",
+    default_level: LintLevel::Allow,
+    desc: "warn about compatibility with Rust 2024",
+    edition_lint_opts: Some((Edition::Edition2024, LintLevel::Deny)),
```
epage, line 73: We're only making this `Deny` for 2024?  That means people can override this to allow it.  If they do, we should make sure that the dependency truly is unused and doesn't create a feature.

src/cargo/util/lints.rs
```diff
@@ -0,0 +1,226 @@
...
+pub struct Lint {
+    pub name: &'static str,
+    pub desc: &'static str,
+    pub groups: &'static [LintGroup],
+    pub default_level: LintLevel,
+    pub edition_lint_opts: Option<(Edition, LintLevel)>,
+}
+
+impl Lint {
+    pub fn level(&self, lints: &TomlToolLints, edition: Edition) -> LintLevel {
```
epage, line 86: We should verify this handles `Forbid` correctly.

src/cargo/util/lints.rs
epage, line 202: Should we have a transitional note that implicit features are no longer supported?

epage, line 202: rustc/clippy lints display the lint name (and where the level was set) on first instance of it being reported

src/cargo/core/workspace.rs
```diff
@@ -1095,11 +1097,14 @@ impl<'gctx> Workspace<'gctx> {
 
     pub fn emit_warnings(&self) -> CargoResult<()> {
         for (path, maybe_pkg) in &self.packages.packages {
+            let path = path.join("Cargo.toml");
+            if let MaybePackage::Package(pkg) = maybe_pkg {
+                self.emit_lints(pkg, &path)?
+            }
```
epage, line 1103: We need a test to ensure this is subject to cap-lints

src/cargo/util/toml/mod.rs
```diff
@@ -1448,17 +1448,24 @@ struct ManifestContext<'a, 'b> {
...
     for (tool, lints) in lints {
-        let supported = ["rust", "clippy", "rustdoc"];
+        let supported = ["cargo", "clippy", "rust", "rustdoc"];
         if !supported.contains(&tool.as_str()) {
             let supported = supported.join(", ");
             anyhow::bail!("unsupported `{tool}` in `[lints]`, must be one of {supported}")
         }
+        if tool == "cargo" && !gctx.cli_unstable().cargo_lints {
+            warn_for_cargo_lint_feature(gctx, warnings);
+        }
```
epage, line 1468: Do we need to strip `lints.cargo` before we publish?

ehuss:
Would it be possible to update `unstable.md` with documentation about this feature?

Also, is #12235 the main tracking issue for what needs to be done next? Is there an outline of what steps need to be taken?

Issue #13290, opened 2024-01-13
Tracking Issue for precise pre-release update


ehuss:
### Summary

RFC: [#3493](https://github.com/rust-lang/rfcs/pull/3493)
Implementation: TODO
Documentation: TODO

Extends `cargo update --precise` to allow pre-release versions, even if the dependency declarations do not specify a pre-release.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

linyihai:
Since https://github.com/rust-lang/cargo/pull/13296 had merged, i think i can lend a hand to continue.

Just in case I forgot, I was gonna @rustbot claim.

epage:
I believe @eopb was planning on working on this.  I'd recommend coordinating with them

eopb:
Yes, I'm still intending to work on this. If you'd like to help @linyihai, please reach out to me on zulip (under the same username as here).

I'm sure we can work together. The last thing I'd want to do is to slow things down by trying to hog this tracking issue.

linyihai:
FYI，I recently tried to implement an MVP based on RFC content.

weihanglo:
<https://github.com/rust-lang/cargo/pull/14013> shows the current implementation is a bit buggy, and there are upper bound semantic issues not yet resolved. See these test cases:

<https://github.com/rust-lang/cargo/blob/050f88c3316178a064619b79ef9f56f0583488cc/src/cargo/util/semver_ext.rs#L204-L233>

It is technically possible to fix them inside cargo. However to do that we'll need to duplicate the logic in `semver` crate. I am in the team of upstreaming things to `semver`.

linyihai:
Yes, The `semver` crate only supports to compare prerelease with prerelease if I remember correct，
https://docs.rs/semver/latest/src/semver/eval.rs.html#14-21

```
 // If a version has a prerelease tag (for example, 1.2.3-alpha.3) then it
    // will only be allowed to satisfy req if at least one comparator with the
    // same major.minor.patch also has a prerelease tag.
    for cmp in &req.comparators {
        if pre_is_compatible(cmp, ver) {
            return true;
        }
    }
```

[...]

PR #13626, opened 2024-03-22
Allow precise update to prerelease.

linyihai:

This is a feature that attempts to support updates to pre-release versions via `cargo update --precise`.

when `precise-pre-release` used, the prerelase version will be taking consider as compatible version.  That said, we can update to any compatible pre-release version. The logic of checking the compatibility of pre-release versions is currently tentative and does not take many conditions into account, this part of the logic makes more sense when implemented in semver.

Use `-Zunstable-options`  instead of  `-Zprecise-pre-release`.

### Additional information
Part of https://github.com/rust-lang/cargo/issues/13290


Eh2406:
I don't think I will be able to give this the attention it deserves until well after Rust Nation. But from a quick perusal it looks like a really good start. Thank you!

src/cargo/sources/registry/mod.rs
```diff
@@ -267,6 +267,7 @@ pub struct RegistrySource<'gctx> {
     /// warning twice, with the assumption of (`dep.package_name()` + `--precise`
     /// version) being sufficient to uniquely identify the same query result.
     selected_precise_yanked: HashSet<(InternedString, semver::Version)>,
+    selected_precise_prerelease: HashSet<(InternedString, semver::Version)>,
```
weihanglo, line 270: A similar comment as yanked one can be added for this field, if we're going to emit that warning

src/cargo/sources/registry/mod.rs
```diff
@@ -827,6 +847,22 @@ impl<'gctx> Source for RegistrySource<'gctx> {
...
+                    .expect("--precise <prerelease-version> in use");
+                if self
+                    .selected_precise_prerelease
+                    .insert((name, version.clone()))
+                {
+                    let mut shell = self.gctx.shell();
+                    shell.warn(format_args!(
+                        "selected prerelease package `{name}@{version}`"
+                    ))?;
+                    shell.note("if possible, try a compatible non-prerelease version")?;
```
weihanglo, line 863: Not sure if we want a message for prereleases. A yanked version implies a "don't use me" message from the crate author. Prereleases are in a different story like "use me at your own risk but thanks for being a tester!" Also, a prerelease version is more visible than a yanked non-prerelease version. People using it may already notice.

weihanglo (commented):
Thanks for this patch!

weihanglo:
Do we have tests for checking the feature gate work and doesn't leak on stable? Like prerelease would not be selected without the flag

weihanglo:
This `--precise <prerelease>` is pretty much similar to `--precise <yanked>`. This make me wonder if we want just overload `-Zunstable-options` for these two purpose instead of a separate `-Zprecise-pre-release` unstable flag.

src/cargo/sources/registry/mod.rs
linyihai: I agree. We can add it back if needed.

linyihai:
> This `--precise <prerelease>` is pretty much similar to `--precise <yanked>`. This make me wonder if we want just overload `-Zunstable-options` for these two purpose instead of a separate `-Zprecise-pre-release` unstable flag.

I already changed it by the way.

tests/testsuite/precise_pre_release.rs
```diff
@@ -41,20 +41,84 @@ perhaps a crate was updated and forgotten to be re-vendored?"#,
...
             name = "package"
+            [dependencies]
+            my-dependency = "0.1.1"
             "#,
         )
         .file("src/lib.rs", "")
         .build();
 
-    p.cargo("-Zprecise-pre-release update")
+    p.cargo("update -p my-dependency --precise 0.1.2-pre.0 -Zunstable-options")
```
weihanglo, line 64: nit: `-p` is no longer needed.

```suggestion
    p.cargo("update my-dependency --precise 0.1.2-pre.0 -Zunstable-options")
```

tests/testsuite/precise_pre_release.rs
```diff
@@ -41,20 +41,84 @@ perhaps a crate was updated and forgotten to be re-vendored?"#,
...
-        .run()
+        // This error is suffering from #12579 but still demonstrates that updating to
+        // a pre-release does not work on stable
+        .with_stderr(
+            r#"[UPDATING] `dummy-registry` index
+[UPDATING] my-dependency v0.1.1 -> v0.1.2-pre.0
+"#,
+        )
+        .run();
+    // Use yanked version.
```
weihanglo, line 74:
```suggestion
    // Use pre-release version.
```

tests/testsuite/precise_pre_release.rs
```diff
@@ -41,20 +41,84 @@ perhaps a crate was updated and forgotten to be re-vendored?"#,
...
         .file("src/lib.rs", "")
         .build();
 
-    p.cargo("-Zprecise-pre-release update")
+    p.cargo("update -p my-dependency --precise 0.1.2-pre.0 -Zunstable-options")
         .masquerade_as_nightly_cargo(&["precise-pre-release"])
-        .with_stderr("")
-        .run()
+        // This error is suffering from #12579 but still demonstrates that updating to
+        // a pre-release does not work on stable
```
weihanglo, line 67:
```suggestion
```

And ditto in `update_pre_release_differ`

src/cargo/sources/registry/mod.rs
```diff
@@ -752,7 +752,13 @@ impl<'gctx> Source for RegistrySource<'gctx> {
         if let Some((_, requested)) = self
             .source_id
             .precise_registry_version(dep.package_name().as_str())
-            .filter(|(c, _)| req.matches(c))
+            .filter(|(c, _)| {
+                if self.gctx.cli_unstable().unstable_options {
```
weihanglo, line 756: I don't know how `match_prerelease` will evolve, so it's better to have a check here maybe?

```suggestion
            .filter(|(c, to)| {
                if to.is_prerelease() && self.gctx.cli_unstable().unstable_options {
```

src/cargo/sources/registry/mod.rs
```diff
@@ -790,7 +796,13 @@ impl<'gctx> Source for RegistrySource<'gctx> {
                 .index
                 .query_inner(dep.package_name(), &req, &mut *self.ops, &mut |s| {
                     let matched = match kind {
-                        QueryKind::Exact => dep.matches(s.as_summary()),
+                        QueryKind::Exact => {
+                            if self.gctx.cli_unstable().unstable_options {
```
weihanglo, line 800: Same here. We might want this check to prevent any regression in the future, if some underlying functions change.

```suggestion
                            if req.is_precise() && self.gctx.cli_unstable().unstable_options {
```

weihanglo (commented):
Thanks for this!

Could we also have [a doc update](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#precise-pre-release) as well?

src/cargo/sources/registry/mod.rs
linyihai: Your concern makes sense.

src/doc/man/cargo-update.md
```diff
@@ -46,6 +46,9 @@ revision (such as a SHA hash or tag).
 While not recommended, you can specify a yanked version of a package (nightly only).
 When possible, try other non-yanked SemVer-compatible versions or seek help
 from the maintainers of the package.
+
+It's allows you choose any compatible `pre-release` version (nightly only) even when a pre-release is not specified by a projects `Cargo.toml`.
```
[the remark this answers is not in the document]
linyihai: Good annotations have always been my weakness, thank you for correcting this description.

tests/testsuite/precise_pre_release.rs
linyihai: Sounds gread, no need to pass -p anymore.

weihanglo (approved):
Thanks. The only missing piece is updating the unstable doc here to `-Zunstable-options`:

https://github.com/rust-lang/cargo/blob/f4ea1d16e9f7547c7dd4d45916142a244dc32332/src/doc/src/reference/unstable.md?plain=1#L350

weihanglo:
Thanks!

@bors r+

PR #13630, opened 2024-03-23
Do not strip debuginfo by default for MSVC

Kobzol:
This PR disables the logic which enables debuginfo stripping by default in release mode (https://github.com/rust-lang/cargo/pull/13257) for MSVC, since it causes problems there (https://github.com/rust-lang/rust/issues/122857).

I'm not sure if this is the correct way to gate the logic on a specific target triple.

The root of the issue is that `-Cstrip=debuginfo` currently behaves like `-Cstrip=symbols` on MSVC, which causes the original optimization to break backtraces on Windows. Ultimately, this should probably be fixed in `rustc`, but that might take some further design work, therefore a faster solution would be to just special case this in Cargo for now, and remove the special case later, once it gets fixed on the `rustc` side.

Related issue: https://github.com/rust-lang/rust/issues/122857

### DIFF
--- a/src/cargo/ops/cargo_compile/mod.rs
+++ b/src/cargo/ops/cargo_compile/mod.rs
@@ -35,6 +35,7 @@
 //! [`drain_the_queue`]: crate::core::compiler::job_queue
 //! ["Cargo Target"]: https://doc.rust-lang.org/nightly/cargo/reference/cargo-targets.html
 
+use cargo_platform::Cfg;
 use std::collections::{HashMap, HashSet};
 use std::hash::{Hash, Hasher};
 use std::sync::Arc;
@@ -436,6 +437,7 @@ pub fn create_bcx<'a, 'gctx>(
         &units,
         &scrape_units,
         host_kind_requested.then_some(explicit_host_kind),
+        &target_data,
     );
 
     let mut extra_compiler_args = HashMap::new();
@@ -575,6 +577,7 @@ fn rebuild_unit_graph_shared(
     roots: &[Unit],
     scrape_units: &[Unit],
     to_host: Option<CompileKind>,
+    target_data: &RustcTargetData<'_>,
 ) -> (Vec<Unit>, Vec<Unit>, UnitGraph) {
     let mut result = UnitGraph::new();
     // Map of the old unit to the new unit, used to avoid recursing into units
@@ -591,6 +594,7 @@ fn rebuild_unit_graph_shared(
                 root,
                 false,
                 to_host,
+                target_data,
             )
         })
         .collect();
@@ -617,6 +621,7 @@ fn traverse_and_share(
     unit: &Unit,
     unit_is_for_host: bool,
     to_host: Option<CompileKind>,
+    target_data: &RustcTargetData<'_>,
 ) -> Unit {
     if let Some(new_unit) = memo.get(unit) {
         // Already computed, no need to recompute.
@@ -634,6 +639,7 @@ fn traverse_and_share(
                 &dep.unit,
                 dep.unit_for.is_for_host(),
                 to_host,
+                target_data,
             );
             new_dep_unit.hash(&mut dep_hash);
             UnitDep {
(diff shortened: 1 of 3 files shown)


weihanglo (commented):
If this is intended to be a stable backport, you need to file a PR against `rust-1.77.0` branch.

Kobzol:
This should also be fixed/changed in nightly, I'm not sure if a different fix will appear *that soon*. So I thought that this should first go into `master`, and only then be backported into stable.

ehuss:
Is this intended to be temporary? It's hard to follow from https://github.com/rust-lang/rust/issues/122857 if this is actually an issue in rustc.

Kobzol:
I modified the PR description to add more context.

workingjubilee:
> Is this intended to be temporary? It's hard to follow from [rust-lang/rust#122857](https://github.com/rust-lang/rust/issues/122857) if this is actually an issue in rustc.

This is an overeager cargo behavior that, when combined with an inconsistent rustc behavior, leads to compounded incorrectness.

workingjubilee:
To be clear, it is my opinion, speaking mostly from my (brief) tenure as backtrace-rs maint:

- Kobzol's implementation for cargo *should have* been fine.
- rustc's behavior verges on simply incoherent, currently.
- sorting this out will likely take more than a week or two.
- I can't realistically test release-build backtrace-rs on Windows nightly right now because Windows release-build backtraces are effectively dead on nightly so whatever workaround I use won't reflect anything anyone except a scant handful of rustc/cargo devs know to do.

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -657,8 +662,13 @@ fn traverse_and_share(
         _ => unit.kind,
     };
 
+    let is_target_windows_msvc = target_data
+        .short_name(&unit.kind)
+        .ends_with("-pc-windows-msvc");
```
workingjubilee, line 667: Backtraces still live on windows-gnu, and (probably?) also work on -gnullvm, but are almost certainly **also** extremely dead on windows7-msvc. While both are tier 3 targets we don't promise any level of support for, I believe the most correct implementation would currently be:
```suggestion
    let is_target_windows_msvc = {
        let name = target_data.short_name(&unit.kind);
        // to be correct re: windows7, gnullvm, and whatever other silliness happens
        name.contains("windows") && name.contains("msvc")
    };
```

ehuss: Instead of scanning target names, I think ideally cargo should be looking at `cfg` values to avoid ambiguity. Perhaps something roughly like:

```rust
let cfg = target_data.cfg(unit.kind);
cfg.contains(Cfg::Name("windows")) && cfg.contains(Cfg::KeyPair("target_env", "msvc"))
```

workingjubilee: oh neat, I didn't realize that exists.

Kobzol: Cool! I modified it to use `Cfg`.

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -657,8 +663,13 @@ fn traverse_and_share(
...
     };
 
+    let cfg = target_data.cfg(unit.kind);
+    let is_target_windows_msvc = cfg.contains(&Cfg::Name("windows".to_string()))
+        && cfg.contains(&Cfg::KeyPair("target_env".to_string(), "msvc".to_string()));
     let mut profile = unit.profile.clone();
-    if profile.strip.is_deferred() {
+    // For MSVC, rustc currently treats -Cstrip=debuginfo same as -Cstrip=symbols, which causes
+    // this optimization to also remove symbols and thus break backtraces.
+    if profile.strip.is_deferred() && !is_target_windows_msvc {
```
weihanglo, line 672: Should we have a simple test for this on Windows MSVC to ensure the patch really fixes it?

Kobzol: We already have a test that this applies on Windows, since CI fails :) I can add another one that checks that backtraces seem to contain some actual symbols. Is there a preferred way of writing platform specific tests?

workingjubilee: It might take some fiddling but you can reference this test-crate for the test that should've worked with *some* reasonable configuration of the Cargo.toml: https://github.com/rust-lang/backtrace-rs/tree/6fa4b85b9962c3e1be8c2e5cc605cd078134152b/crates/without_debuginfo

tests/testsuite/profiles.rs
```diff
@@ -656,7 +658,32 @@ fn strip_debuginfo_in_release() {
         .run();
 }
 
+// Using -Cstrip=debuginfo in release mode by default is temporarily disabled on Windows due to
+// https://github.com/rust-lang/rust/issues/122857
 #[cargo_test]
+#[cfg(windows)]
```
weihanglo, line 664: Is this test expected to run also on `windows-gnu`?

```suggestion
#[cfg(all(target_os = "windows", target_env = "msvc"))]
```

Kobzol: Thank you :)

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #6194, opened 2018-10-20
Remove the remove TODO on Compilation.native_dirs

dwijnand:
It's still used, see tests:

* build_script::doctest_receives_build_link_args
* build_script::rename_with_link_search_path
* run::run_with_library_paths
* run::library_paths_sorted_alphabetically

Originally I was trying to action https://github.com/rust-lang/cargo/pull/792#discussion_r226826431 with this PR

PR #13631, opened 2024-03-23
docs: remove untrue TODO for `native_dirs`

weihanglo:
`native_dirs` is still in use.


ehuss:
Thanks! Looks like this was missed in #6194.

@bors r+

PR #13632, opened 2024-03-23
fix: Warn on -Zlints

Muscraft:
When the `[lints]` table was stabilized in #12648, it appears that making `-Zlints` into a warning when specified was missed, this PR fixes that.


ehuss (commented):
I think the field also needs to be removed from the `unstable_cli_options`.

ehuss:
Thanks!

@bors r+

Issue #344, opened 2014-08-07
Latest cargo nightly is segfaulting in toml


carllerche:
I installed the latest Cargo nightly on a clean ubuntu box:

```
$ uname -a
Linux vagrant-ubuntu-trusty-64 3.13.0-32-generic #57-Ubuntu SMP Tue Jul 15 03:51:08 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux

$ gdb cargo-build
Program received signal SIGSEGV, Segmentation fault.
0x000000000046774c in util::toml::to_manifest::h027d07b5ab34a4f4lto ()
(gdb) bt
#0  0x000000000046774c in util::toml::to_manifest::h027d07b5ab34a4f4lto ()
#1  0x0000000000464330 in ops::cargo_read_manifest::read_manifest::h58a159c4a720cfaeAHh ()
#2  0x000000000046ce3c in ops::cargo_read_manifest::read_package::haa2e9798e90a9eb97Hh ()
#3  0x0000000000470cc0 in ops::cargo_read_manifest::read_nested_packages::hccd53a40c03ffabbn0h ()
#4  0x000000000046f2fe in ops::cargo_read_manifest::process_possible_package::h8a4bd8fddd7ab3edYYh ()
#5  0x000000000046eae5 in ops::cargo_read_manifest::read_packages::h800d7d3df07bcc60WLh ()
#6  0x000000000041d25d in sources::path::PathSource.Source::update::h1adac7d7b084e48fUOl ()
#7  0x0000000000446c5d in ops::cargo_compile::compile::h5070006c8fe8bb4c0fh ()
#8  0x000000000040f1a3 in execute::h747a9f9db13b36eaqia ()
#9  0x000000000040dce4 in call_main_without_stdin::h12507149885516441373 ()
#10 0x000000000040a544 in main::h09e8cb9e8a4fd2faiia ()
#11 0x0000000000555473 in start::closure.8371 ()
#12 0x0000000000664b7c in rust_try ()
#13 0x0000000000661ed7 in unwind::try::h9e31b701f4e2b54d0Wd ()
#14 0x0000000000661cbf in task::Task::run::h2bd486ed91d9e52aC4c ()
#15 0x000000000055528d in start::h4e9a0bf33351433aTpe ()
#16 0x00000000005550c9 in lang_start::h9d7a12ba8e6cb946dpe ()
[...]

PR #13635, opened 2024-03-24
refactor: Make lint names snake_case

Muscraft:
When working on #13621, I somehow missed that lint names should be `snake_case` according to the [`rustc-dev-guide`](https://rustc-dev-guide.rust-lang.org/diagnostics.html#lint-naming) as well as [`RFC #344`](https://rust-lang.github.io/rfcs/0344-conventions-galore.html#lints). 

This PR renames: 
- `implicit-features` => `implicit_featires`
- `rust-2024-compatibility` => `rust_2024_compatibility`. 

<hr>

Note: We should probably have some tooling to enforce this, but I was unsure if it belonged to this PR or another one. One solution would be to use a macro to create the `const LINT_NAME: Lint = {...}`, where `LINT_NAME` would be the `ident` as well as the `name: &'static str` and then have a method on `Lint` to make it lowercase as needed. This is what `rustc` does, and it could work well here. It would ensure snake case as `const` names need to be [`SCREAMING_SNAKE_CASE`](https://rust-lang.github.io/rfcs/0430-finalizing-naming-conventions.html#general-naming-conventions), or a warning is shown.

### DIFF
--- a/src/cargo/core/workspace.rs
+++ b/src/cargo/core/workspace.rs
@@ -1128,18 +1128,23 @@ impl<'gctx> Workspace<'gctx> {
 
     pub fn emit_lints(&self, pkg: &Package, path: &Path) -> CargoResult<()> {
         let mut error_count = 0;
-        let lints = pkg
+        let toml_lints = pkg
             .manifest()
             .resolved_toml()
             .lints
             .clone()
             .map(|lints| lints.lints)
-            .unwrap_or(manifest::TomlLints::default())
+            .unwrap_or(manifest::TomlLints::default());
+        let cargo_lints = toml_lints
             .get("cargo")
             .cloned()
             .unwrap_or(manifest::TomlToolLints::default());
+        let normalized_lints = cargo_lints
+            .into_iter()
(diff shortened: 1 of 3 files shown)


epage:
This is strange because every lint and group in `rustc -Whelp` uses kebab-case, including `rust-2024-compatibility` which you are mirroring.

src/cargo/core/workspace.rs
```diff
@@ -1128,18 +1128,23 @@ impl<'gctx> Workspace<'gctx> {
...
-            .unwrap_or(manifest::TomlLints::default())
+            .unwrap_or(manifest::TomlLints::default());
+        let cargo_lints = toml_lints
             .get("cargo")
             .cloned()
             .unwrap_or(manifest::TomlToolLints::default());
+        let normalized_lints = cargo_lints
+            .into_iter()
+            .map(|(name, lint)| (name.replace('-', "_"), lint))
+            .collect();
```
epage, line 1145: non-blocking: this could live in `toml/mod.rs` when we resolve the manifest.

Please don't do that as I have probably 20 commits locally changing that code :)

Issue #3419, opened 2016-12-16
`cargo check` fails oddly on binaries


alexcrichton:
Not entirely sure what's going on here....

```
// Cargo.toml
[package]
name = "lib"
version = "0.1.0"
authors = []

[dependencies]
rustc-serialize = "*"

// src/lib.rs
extern crate rustc_serialize;

use rustc_serialize::Decodable;

pub fn take<T: Decodable>() {}

// src/main.rs
extern crate rustc_serialize;

extern crate lib;

#[derive(RustcDecodable)]
pub struct Foo;

fn main() {
    lib::take::<Foo>();
}
```

With that structure:

```
$ cargo check
    Updating registry `https://github.com/rust-lang/crates.io-index`
   Compiling rustc-serialize v0.3.22
   Compiling lib v0.1.0 (file:///home/alex/code/lol)
error[E0277]: the trait bound `Foo: rustc_serialize::serialize::Decodable` is not satisfied
 --> src/main.rs:9:5
  |
9 |     lib::take::<Foo>();
  |     ^^^^^^^^^^^^^^^^ the trait `rustc_serialize::serialize::Decodable` is not implemented for `Foo`
  |
  = note: required by `lib::take`

error: aborting due to previous error

error: Could not compile `lib`.

To learn more, run the command again with --verbose.
```

@nrc mind taking a look?

nrc:
looking into this...

nrc:
This appears to be fixed by #3425. Will submit a PR with a test...

PR #3296, opened 2016-11-16
cargo check

nrc:
~~This is not finished - the big omission is no tests, and it needs some more testing too. It also requires https://github.com/rust-lang/rust/pull/37681.~~

However, this is my first non-trivial Cargo patch, so I'd like to get some early feedback.

r? @alexcrichton 

and cc @rust-lang/tools since this adds a new Cargo sub-command (although we have previously discussed and approved the idea in a tools meeting).

PR #13637, opened 2024-03-25
Remove unnecessary test

jhpratt:
This removes a test that is blocking rust-lang/rust#116016 from landing. @dtolnay suggested removing the test rather than allowing the relevant lints ([comment](https://github.com/rust-lang/rust/pull/116016#issuecomment-2016649022)).

> The failure is in https://github.com/rust-lang/cargo/blob/77506e57392d94450bb3ed52cf75e3263bbb2792/tests/testsuite/check.rs#L222-L285 which is a 7 year old test for [rust-lang/cargo#3419](https://github.com/rust-lang/cargo/issues/3419), which is from 2 days after the initial implementation of `cargo check` landed in nightly Cargo in [rust-lang/cargo#3296](https://github.com/rust-lang/cargo/pull/3296).
> 
> I would recommend just deleting the test from Cargo. The implementation of `cargo check` has matured enough since then and I don't see anything useful in that test.


weihanglo (approved):
Agree that we might not need the test. The original fix was like handling cargo check to correctly emit rmeta, which should be covered by other `cargo check` test nowadays.

PR #13638, opened 2024-03-25
Update book: Rust 1.77 changed the default behaviour on debug symbol stripping

alpitol:
[Rust 1.77](https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html#enable-strip-in-release-profiles-by-default) changed the default behaviour on debug symbols stripping. That change seems not to be yet updated to the book, which this PR fixes.

### DIFF
--- a/src/doc/src/reference/profiles.md
+++ b/src/doc/src/reference/profiles.md
@@ -113,11 +113,12 @@ strip either symbols or debuginfo from a binary. This can be enabled like so:
 # ...
 
 [profile.release]
-strip = "debuginfo"
+strip = "symbols"
 ```
 
 Possible string values of `strip` are `"none"`, `"debuginfo"`, and `"symbols"`.
-The default is `"none"`.
+For debug profiles the default is `"none"`, and for profiles with `debug = false` 
+the default is `"debuginfo"`.
 
 You can also configure this option with the boolean values `true` or `false`.
 `strip = true` is equivalent to `strip = "symbols"`. `strip = false` is
@@ -293,7 +294,7 @@ The default settings for the `release` profile are:
 opt-level = 3
 debug = false
 split-debuginfo = '...'  # Platform-specific.
-strip = "none"
+strip = "debuginfo" 
 debug-assertions = false
 overflow-checks = false
 lto = false


src/doc/src/reference/profiles.md
```diff
@@ -113,11 +113,12 @@ strip either symbols or debuginfo from a binary. This can be enabled like so:
...
 
 [profile.release]
-strip = "debuginfo"
+strip = "symbols"
 ```
 
 Possible string values of `strip` are `"none"`, `"debuginfo"`, and `"symbols"`.
-The default is `"none"`.
+For debug profiles the default is `"none"`, and for profiles with `debug = false` 
+the default is `"debuginfo"`.
```
epage, line 122: > [Rust 1.77](https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html#enable-strip-in-release-profiles-by-default) changed the default behaviour on debug symbols stripping. That change seems not to be yet updated to the book, which this PR fixes.

I wonder if we need to update the documentation.  The idea behind this is to emulate `std` respecting `debug`.

src/doc/src/reference/profiles.md
```diff
@@ -293,7 +294,7 @@ The default settings for the `release` profile are:
 opt-level = 3
 debug = false
 split-debuginfo = '...'  # Platform-specific.
-strip = "none"
+strip = "debuginfo"
```
weihanglo, line 297: This seems incorrect to me. The default value of `strip` for release profile is still `none`.

alpitol:
Ah, ok. I don't really know how things work in cargo, just thought that this was a small slip of outdated information.
You know it best here, this PR can be closed if this documentation change is not wanted.

weihanglo:
That is not anyone's fault. The behavior is a bit hard to explain in one sentence. The default is still `strip=none` if you turn on `debug` for release profile. `strip=debuginfo` only happens for release profile when debug is off. One idea behind the change is let Cargo do what it promises: _"`0`, `false`, or `"none"`: no debug info at all"._ Without this change you will still get debuginfo from std even with `debug=0`. See <https://github.com/rust-lang/cargo/issues/4122#issuecomment-1868318860>.

Thanks for your patch as always. I'll going to close this and hope the explanation help you :)

ia0:
> `strip=debuginfo` only happens for release profile when debug is off.

The situation is actually more complicated than this, because "debug is off" is from cargo's perspective. When using `RUSTFLAGS='-C debuginfo=2' cargo build --release`, debuginfo will still be stripped. One has to explicitly override cargo's behavior with `RUSTFLAGS='-C debuginfo=2 -C strip=none' cargo build --release`.

That's not an issue, but just why the situation is even more complicated and difficult to debug.

weihanglo:
RUSTFLAGS is a lower level mechanism. The general idea is if a configuration is provided by Cargo, avoid using RUSTFLAGS to do it. Cargo doesn't parse nor track the content of RUSTFLAGS.

Such situation also applies to `RUSTFLAGS=--target xxx`

ia0:
I was going to write the following:

> The difference with `--target` is that it's a cargo flag, not a configuration in the `Cargo.toml`, so it's easy to pass on the command line. For `debug`, one would need to generate a new profile on the spot.

But then I realized there's the `--config` flag. So actually, I can simply `--config=profile.release.debug=2` instead of `RUSTFLAGS='-C debuginfo=2 -C strip=none'`.

Thanks for putting me in the right direction!

weihanglo:
Probably a better (?) way to share a set of compiler flags: make it a [custom profile](https://doc.rust-lang.org/nightly/cargo/reference/profiles.html#custom-profiles) in Cargo.toml, and you can specify via `--profile myrelease`.

src/doc/src/reference/profiles.md
mhnap: Is `strip` by default will be equal to `none` for the release profile if `debug` is set to `false`? I read #13677, #13257, #13763, and #13638 but still didn't get it. Thanks in advance!

weihanglo: The behavior of auto-stripping debuginfo is kinda dynamic. The default release profile settings is still static and `strip` is none when you want to inherit it.

Share these pieces of code for clarity.

https://github.com/rust-lang/cargo/blob/94aa7fb1321545bbe922a87cb11f5f4559e3be63/src/cargo/core/profiles.rs#L719-L728

https://github.com/rust-lang/cargo/blob/94aa7fb1321545bbe922a87cb11f5f4559e3be63/src/cargo/core/profiles.rs#L630-L651

mhnap: > The default release profile settings is still static and strip is none when you want to inherit it.

Hmm.
According to [Rust Blog](https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html#enable-strip-in-release-profiles-by-default):
> Cargo [profiles](https://doc.rust-lang.org/stable/cargo/reference/profiles.html) which do not enable [debuginfo](https://doc.rust-lang.org/stable/cargo/reference/profiles.html#debug) in outputs (e.g., debug = 0) will enable strip = "debuginfo" by default.

And according to [Cargo Book](https://doc.rust-lang.org/cargo/reference/profiles.html#release):
> [profile.release]
...
debug = false

I made a small test on 1.78.0 and release profile seems indeed have set `strip = "debuginfo"` by default.
Not sure that behaviour should be, but it seems for me also that documentation is not updated and #13677 make sense.
Or am I missing something here?
Thanks!

weihanglo: See also <https://github.com/rust-lang/cargo/pull/13638#issuecomment-2018594383> and <https://github.com/rust-lang/cargo/pull/13677#discussion_r1545924873>.

weihanglo: The settings are static. The actual flags passed to rustc are deferred later when Cargo see if it can safely strip std or not.

mhnap: > The default is still strip=none if you turn on debug for release profile. strip=debuginfo only happens for release profile when debug is off.

But `debug` is off by default for the release profile, so `strip="debuginfo"` will take place, no? 🤔
Not sure I get it correctly, but maybe it would be great to add more context to `The Cargo Book`.
Thanks for the explanation!

PR #13639, opened 2024-03-25
Add `unknown-lints` lint

Muscraft:
This PR adds a lint for `unknown-lints` similar [`rustc::unknown-lints`](https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#unknown-lints). This will help users know when they are passing something to `cargo` that `cargo` doesn't understand.

### DIFF
--- a/src/cargo/util/lints.rs
+++ b/src/cargo/util/lints.rs
@@ -1,5 +1,5 @@
 use crate::core::FeatureValue::Dep;
-use crate::core::{Edition, FeatureValue, Package};
+use crate::core::{Edition, FeatureValue, MaybePackage};
 use crate::util::interning::InternedString;
 use crate::{CargoResult, GlobalContext};
 use annotate_snippets::{Level, Renderer, Snippet};
@@ -57,6 +57,8 @@ fn rel_cwd_manifest_path(path: &Path, gctx: &GlobalContext) -> String {
         .display()
         .to_string()
 }
+
+const LINT_GROUPS: &[LintGroup] = &[RUST_2024_COMPATIBILITY];
 
 #[derive(Copy, Clone, Debug)]
 pub struct LintGroup {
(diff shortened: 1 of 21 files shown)


src/cargo/util/lints.rs
```diff
@@ -136,6 +138,121 @@ impl From<TomlLintLevel> for LintLevel {
...
+        .collect::<HashSet<_>>();
+
+    if let Some(lints) = original_toml
+        .workspace
+        .as_ref()
+        .map_or(&None, |ws| &ws.lints)
+    {
+        if let Some(cargo_lints) = lints.get("cargo") {
+            for (name, _) in cargo_lints.iter() {
+                let normalized_name = name.replace("-", "_");
```
epage, line 188: This is having me lean more towards normalizing earlier in the process

src/cargo/util/lints.rs
```diff
@@ -136,6 +138,121 @@ impl From<TomlLintLevel> for LintLevel {
...
+    let all_names = LINTS
+        .iter()
+        .map(|lint| lint.name)
+        .chain(LINT_GROUPS.iter().map(|group| group.name))
+        .collect::<HashSet<_>>();
+
+    if let Some(lints) = original_toml
+        .workspace
+        .as_ref()
+        .map_or(&None, |ws| &ws.lints)
```
epage, line 184: imo this should be treated as a separate "workspace" lint that is only subject to `workspace.lints` so we have consistency between virtual and real workspace packages.

tests/testsuite/lints_table.rs
```diff
@@ -763,7 +763,6 @@ fn cargo_lints_nightly_required() {
                 authors = []
 
                 [lints.cargo]
-                "unused-features" = "deny"
```
epage, line 766: Why were these here and why are they being removed?

ehuss:
r? epage
if you don't mind.

Muscraft:
Going to close in favor of #13837, as it contains the addition of an `unknown_lints`
lint

Issue #13484, opened 2024-02-23
Providing path to originally selected manifest file


wmmc88:
### Problem

When running cargo commands like `cargo build`, packages are automatically selected based off of the cargo manifest selected (either based off of the current working directory of the invocation or the `--manifest-path` argument). AFAIU, a dependency graph is generated based off of this manifest and then the build is executed. By this logic, before any build-script runs, cargo already knows all of the packages that need to be built and has access to all of its manifest data. I want to access that data in one of the dependency's build.rs.

I have 2 crates- crate A and crate B. crate B has a path dependency on crate A. I would like to be able to read custom metadata in crate B's manifest by using `cargo_metadata` from within crate A's `build.rs`. This metadata is something like a globally configured feature proposed in https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618/38. When running `cargo metadata` cli command from crate B's directory, I can see the metadata in the crate B's manifest, but when getting metadata via crate A's build.rs, I cannot. If both crates existed in the same workspace, it would work, but I intend for crate A to be able to be consumed by others to use and configure via metadata in user's dependent crates.

If the originally selected manifest path was provided to build scripts, I could point to that manifest when using `cargo_metadata` in crate A's build.rs. As it stands, I need to use `OUT_DIR` to derive where the selected/root manifest of the build is, but this is fragile and does not work when the default target directory is changed by things like `--target-dir`(similar issue to #9661 ).

### Proposed Solution

A environment variable provided to builds scripts that exposes to originally selected manifest path.

### Notes

seems like a similar feature request to #8148

weihanglo:
> ... cargo already knows all of the packages that need to be built and has access to all of its manifest data. I want to access that data in one of the dependency's build.rs.

This kind of inverse relation that a dependency knows the full picture of a build sounds a bit odd. Could you elaborate more on what you actually want to achieve?

BTW, there is a relevant issue <https://github.com/rust-lang/cargo/issues/3946> about finding the root workspace, and the current experiment of that is an unstable env var [`CARGO_RUSTC_CURRENT_DIR`](https://doc.rust-lang.org/nightly/cargo/reference/environment-variables.html?highlight=CARGO_RUSTC_CURRENT_DIR#environment-variables-cargo-sets-for-crates) you might be interested in.

wmmc88:
> This creates an inverse relationship in packages

[...]

PR #13644, opened 2024-03-26
feat: Stablize `CARGO_RUSTC_CURRENT_DIR`

### LINKED ISSUE #3946: Environment variable for Cargo Workspace

epage:
This provides what cargo sets as the `current_dir` for the `rustc` process.
While `std::file!` is unspecified in what it is relative to, it is relatively safe, it is generally relative to `rustc`s `current_dir`.

This can be useful for snapshot testing.
For example, `snapbox` has been using this macro on nightly since assert-rs/trycmd#247, falling back to finding a parent of `CARGO_MANIFEST_DIR`, if present.
This has been in use in Cargo since rust-lang/cargo#13441.

This was added in #12996.
Relevant points discussed in that issue:
- This diverged from the original proposal from the Cargo team of having a `CARGO_WORKSPACE_DIR` that is the "workspace" of the package being built (ie registry packages would map to `CARGO_MANIFEST_DIR`). In looking at the `std::file!` use case, `CARGO_MANIFEST_DIR`, no matter how we defined it, would only sort of work because no sane definition of that maps to `rustc`'s `current_dir`.a This instead focuses on the mechanism currently being used.
- Using "current dir" in the name is meant to be consistent with `std::env::current_dir`.
- I can go either way on `CARGO_RUSTC` vs `RUSTC`.  Existing related variables:
  - `RUSTC`
  - `RUSTC_WRAPPER`
  - `RUSTC_WORKSPACE_WRAPPER`
  - `RUSTFLAGS` (no `C`)
  - `CARGO_CACHE_RUSTC_INFO`

Note that #3946 was overly broad and covered many use cases. One of those was for packages to look up information on their dependents.
Issue #13484 is being left open to track that.

Fixes #3946

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -696,16 +696,14 @@ fn prepare_rustc(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> CargoResult
         let tmp = build_runner.files().layout(unit.kind).prepare_tmp()?;
         base.env("CARGO_TARGET_TMPDIR", tmp.display().to_string());
     }
-    if build_runner.bcx.gctx.nightly_features_allowed {
-        // This must come after `build_base_args` (which calls `add_path_args`) so that the `cwd`
-        // is set correctly.
-        base.env(
-            "CARGO_RUSTC_CURRENT_DIR",
-            base.get_cwd()
-                .map(|c| c.display().to_string())
-                .unwrap_or(String::new()),
-        );
-    }
+    // This must come after `build_base_args` (which calls `add_path_args`) so that the `cwd`
+    // is set correctly.
+    base.env(
+        "CARGO_RUSTC_CURRENT_DIR",
+        base.get_cwd()
+            .map(|c| c.display().to_string())
+            .unwrap_or(String::new()),
+    );
 
     Ok(base)
 }
--- a/src/doc/src/reference/environment-variables.md
+++ b/src/doc/src/reference/environment-variables.md
@@ -267,7 +267,7 @@ corresponding environment variable is set to the empty string, `""`.
   where integration tests or benchmarks are free to put any data needed by
   the tests/benches. Cargo initially creates this directory but doesn't
   manage its content in any way, this is the responsibility of the test code.
-* `CARGO_RUSTC_CURRENT_DIR` --- This is a path that `rustc` is invoked from **(nightly only)**.
+* `CARGO_RUSTC_CURRENT_DIR` --- This is a path that `rustc` is invoked from
 
 [Cargo target]: cargo-targets.md
 [binaries]: cargo-targets.md#binaries
--- a/tests/testsuite/build.rs
+++ b/tests/testsuite/build.rs
@@ -1748,9 +1748,7 @@ fn crate_env_vars() {
     };
 
     println!("build");
-    p.cargo("build -v")
-        .masquerade_as_nightly_cargo(&["CARGO_RUSTC_CURRENT_DIR"])
-        .run();
+    p.cargo("build -v").run();
 
     println!("bin");
     p.process(&p.bin("foo-bar"))
@@ -1758,20 +1756,14 @@ fn crate_env_vars() {
         .run();
 
     println!("example");
-    p.cargo("run --example ex-env-vars -v")
-        .masquerade_as_nightly_cargo(&["CARGO_RUSTC_CURRENT_DIR"])
-        .run();
+    p.cargo("run --example ex-env-vars -v").run();
 
     println!("test");
-    p.cargo("test -v")
-        .masquerade_as_nightly_cargo(&["CARGO_RUSTC_CURRENT_DIR"])
-        .run();
+    p.cargo("test -v").run();
 
     if is_nightly() {
         println!("bench");
-        p.cargo("bench -v")
-            .masquerade_as_nightly_cargo(&["CARGO_RUSTC_CURRENT_DIR"])
-            .run();
+        p.cargo("bench -v").run();
     }
 }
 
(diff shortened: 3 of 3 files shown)


src/cargo/core/compiler/mod.rs
```diff
@@ -696,16 +696,14 @@ fn prepare_rustc(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> CargoResult
...
-                .unwrap_or(String::new()),
-        );
-    }
+    // This must come after `build_base_args` (which calls `add_path_args`) so that the `cwd`
+    // is set correctly.
+    base.env(
+        "CARGO_RUSTC_CURRENT_DIR",
+        base.get_cwd()
+            .map(|c| c.display().to_string())
+            .unwrap_or(String::new()),
```
weihanglo, line 705: Wonder if it could just be.

```suggestion
        base.get_cwd().unwrap_or_default(),
```

(and ditto for `CARGO_TARGET_TMPDIR`? though that is not relevant to this PR)

src/doc/src/reference/environment-variables.md
```diff
@@ -265,7 +265,7 @@ corresponding environment variable is set to the empty string, `""`.
   where integration tests or benchmarks are free to put any data needed by
   the tests/benches. Cargo initially creates this directory but doesn't
   manage its content in any way, this is the responsibility of the test code.
-* `CARGO_RUSTC_CURRENT_DIR` --- This is a path that `rustc` is invoked from **(nightly only)**.
+* `CARGO_RUSTC_CURRENT_DIR` --- This is a path that `rustc` is invoked from
```
weihanglo, line 268: Still feel like people might overlook this, as it isn't an obvious thing easy to relate to their problems. That said, I am fine with the previous explanation. Just forward it here if people have the same question: <https://github.com/rust-lang/cargo/pull/12996#discussion_r1404504001>.

> When we talked about this, Eric was hesitant about locking down the behavior of `file!` so I left out that use case. It should work and I can't imagine rustc doing anything but absolute or relative to `std::env::current_dir` but I'm leaving that to people to make the choice that they want to make those assumptions, rather than us advertising it.

weihanglo (approved):

epage:
> Is there some information about the use cases for this? How is it expected to be used?

One of the requests in #3946 is a way to be able to resolve `std::file!()` so that the paths can be looked up at runtime.  A common use for this is in snapshot testing.  

For example, [snapbox::path::current_dir!](https://github.com/assert-rs/trycmd/blob/28de73bb57bac5fc5478946dce8f14b9a8214218/crates/snapbox/src/macros.rs#L6-L8)) is useful for
- Finding snapshot files relative to the test mod
- Finding the test mod for updating of inline snapshots

Possible workarounds
- `CARGO_MANIFEST_DIR`: Cargo initially used this for UI tests but ran into problems when being inside of the rust workspace (when we used to do that)
  - Like in snapbox, this can be augmented by a manual path walk that mostly works
- Defining a blank path for the `CARGO_WORKSPACE_DIR` in `[env]` config.  This does find the workspace dir but that isn't always the current-dir used for rustc which is what `std::file!` is determined by
- Combination of the above

As you mentioned the last time the Cargo team discussed this, `std::file!`s behavior is not specified. 

For a summary of the team's last discussion on this, see https://github.com/rust-lang/cargo/issues/3946#issuecomment-1402464796

ehuss:
I am concerned that if the primary (or only?) way to use this environment variable is with `file!`, and `file!` isn't defined and primarily intended for debugging purposes that it is changing the meaning and relationship of `file!`. There might also be concerns that if something like trim-paths changes the behavior of `file!`. It also seems to provide a way to circumvent trim-paths, which seems a bit concerning.

Were there other considerations made, such as adding something like `abs_file!()`?

Have there been any discussions with the libs team, and what their thoughts are on making `file!` be used for something other than debugging?

> This is another benefit to starting life as test/bench only as it lowers the barrier for abusing file!.

Is this limited to only test/bench?

bukowa:
I'm not sure if this is the appropriate place and topic—if not, feel free to delete this message.

I'm wondering how to obtain the full path to the current file, regardless of whether I'm on Linux, Windows, or Mac. I would like it to be absolute so that I can start testing my library—specifically, file operations that are located in the `tests/fixtures` folder within my test library path. Therefore, obtaining this path is very crucial for the proper functioning of the tests in my program.

Now imagine that I am working with a workspace, or I am working with a workspace where one package is dependent on another package, and they have different tests... so how can I get the absolute path to the current file? So that I can always move the `test_fixture.rs` file along with the `fixtures` folder - not worrying about implementation details?

RalfJung:
How does this interact with caching, specifically https://github.com/rust-lang/cargo/pull/4788? A bunch of work was done (causing quite [painful side-effects](https://github.com/rust-lang/cargo/issues/9887)) to ensure that a project can be moved to a different location in the file system without invalidating the cache. But if the current absolute dir is an input to the build then that can't work, can it?

epage:
This would only break caching if you use it and only for those build targets, just like with `CARGO_MANIFEST_DIR`.

For trim-paths, I see this having a similar interaction as `CARGO_MANIFEST_DIR`.

For `file!`s intended purpose, I talked with libs-api about a counter-proposal for this to be handled in the standard library and they pushed back as well.  They seemed fine with us solving it *but* its been a few months and I do not remember if we explicitly called out this re-purposing of `file!`.  I will note that the documentation for this doesn't call out `file!` and it would be left for the users to decide to do like they are doing today.  If we change `file!`, we'll have to decide if we are ok with breaking people who assumed more than they should.  The situation here doesn't change too much.  We do streamline it but most users likely already have `file!` abstracted away from them in a snapshotting API and so doing so is unlikely to affect adoption.

RalfJung:
Never mind, I misunderstood what this feature does.

epage:
Discussed this with @Amanieu at RustConf about this.

In stepping through the use cases, he is now leaning towards `file_absolute!()`.  This potentially would warn or error if trim paths is enabled.

epage:
Created the ACP for `file_absolute!` at rust-lang/libs-team#478

PR #13649, opened 2024-03-26
test: Add asserts to catch BorrowMutError's

epage:
### What does this PR try to resolve?

This intentionally borrows from `RefCell`s before conditional code to try to prevent regressions like #13646 without exhaustively covering every case that could hit these `BorrowMutError`s with complicated tests.

### How should we test and review this PR?

I tested this by ensuring the registry code path panicked before rebasing on top of #13647.
Once I rebased, the panic went away.

PR #13647, opened 2024-03-26
fix: do not borrow shell across registry query

weihanglo:
### What does this PR try to resolve?

Fixes #13646

### How should we test and review this PR?

```
$ git clone https://github.com/taiki-e/easytime
$ cd easytime
$ cargo generate-lockfile
$ cargo update -Z minimal-versions --verbose
```

### Additional information

A better and safer way is always calling `gctx.shell()` instead.

Issue #13646, opened 2024-03-26
Panic with "already borrowed: BorrowMutError" since nightly-2024-03-26


taiki-e:
### Problem

Since nightly-2024-03-26, -Z minimal-versions and -Z direct-minimal-versions panic on some my projects when lockfile is already exists.

EDIT: this seems to be reproducible without -Z minimal-versions/-Z direct-minimal-versions: https://github.com/rust-lang/cargo/issues/13646#issuecomment-2019285121

### Steps

```console
$ git clone https://github.com/taiki-e/easytime
$ cd easytime
$ cargo generate-lockfile
$ cargo update -Z minimal-versions
    Updating crates.io index
 Downgrading aho-corasick v1.1.3 -> v0.6.0 (latest: v1.1.3)
 Downgrading anyhow v1.0.81 -> v1.0.0 (latest: v1.0.81)
thread 'main' panicked at src/cargo/util/context/mod.rs:412:20:
already borrowed: BorrowMutError
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::cell::panic_already_borrowed
   3: <cargo::util::progress::State>::tick
   4: <cargo::sources::registry::http_remote::HttpRegistry as cargo::sources::registry::RegistryData>::block_until_ready
   5: <cargo::sources::registry::RegistrySource as cargo::sources::source::Source>::block_until_ready
   6: <cargo::sources::replaced::ReplacedSource as cargo::sources::source::Source>::block_until_ready
   7: <cargo::core::registry::PackageRegistry as cargo::core::registry::Registry>::block_until_ready
   8: cargo::ops::cargo_generate_lockfile::print_lockfile_updates
   9: cargo::ops::cargo_generate_lockfile::update_lockfile
  10: cargo::commands::update::exec
  11: <cargo::cli::Exec>::exec
  12: cargo::cli::main
  13: cargo::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

It can be work around by removing Cargo.lock before calling `-Z minimal-versions`.

It can also be reproduced with https://github.com/taiki-e/pin-project-lite.

It can also be reproduced with -Z direct-minimal-versions.

```console
$ cargo minimal-versions build --workspace --no-private --detach-path-deps=skip-exact --all-features --direct
info: running `cargo update -Z direct-minimal-versions`
thread 'main' panicked at src/cargo/util/context/mod.rs:412:20:
[...]

PR #13650, opened 2024-03-26
test: Add asserts to catch BorrowMutError's

epage:

This intentionally borrows from `RefCell`s before conditional code to try to prevent regressions like #13646 without exhaustively covering every case that could hit these `BorrowMutError`s with complicated tests.

I tested this by ensuring the registry code path panicked before rebasing on top of #13647.
Once I rebased, the panic went away.

### Additional information

Split off from #13649 to try to avoid appveyor

### DIFF
--- a/src/cargo/core/registry.rs
+++ b/src/cargo/core/registry.rs
@@ -739,6 +739,11 @@ impl<'gctx> Registry for PackageRegistry<'gctx> {
     }
 
     fn block_until_ready(&mut self) -> CargoResult<()> {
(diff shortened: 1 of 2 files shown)


epage:
Nope, still hitting it.

weihanglo:
That git commit is broken and got cached?

This amends the commit and seems work: https://github.com/rust-lang/cargo/pull/13651

weihanglo:
Close in favor of #13651.

PR #13654, opened 2024-03-26
[stable-1.77] Do not strip debuginfo by default for MSVC

weihanglo:
Stable backports:

- https://github.com/rust-lang/cargo/pull/13630

-

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -252,7 +252,7 @@ dependencies = [
 
 [[package]]
 name = "cargo"
-version = "0.78.0"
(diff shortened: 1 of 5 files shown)


weihanglo:
Assumed we're going to release the new version of cargo crate, I bumped it to 0.78.1. Let me know if we don't want that.

weihanglo:
check-version-bump job failed because of rustdoc.json version incompatibility. I am not too worried about that as we don't really have any semver-major change in this backport.

Issue #13645, opened 2024-03-26
`cargo add` deletes comments in TOML unexpectedly under certain conditions


cyqsimon:
### Problem

Running `cargo add <EXISTING_DEP>` causes the preceding comments to be deleted in `Cargo.toml`.

### Steps

1. Create a new project with `cargo new`
2. Add the following to `Cargo.toml`:
```toml
[dependencies]
# Duis aliqua ullamco incididunt cupidatat Lorem est consequat nostrud mollit veniam fugiat.
# Magna elit anim officia reprehenderit tempor exercitation in do voluptate cillum sunt deserunt duis est.
chrono = "0.4.35"
# Enim dolor veniam ipsum pariatur ullamco laborum veniam cillum dolore aliqua in officia.
# Proident esse cillum mollit adipisicing culpa ad labore labore.
serde = "1.0.197"
# Tempor sunt duis labore dolore consequat labore ad.
# Ipsum eu do amet excepteur sint nisi officia ullamco occaecat eu commodo amet veniam.
```
3. Run `cargo add chrono/serde` (in practise the likely reason you do this is to specify additional features, but that's not necessary for a MRE)
4. Observe that the two lines of preceding comments are deleted

### Version

```text
cargo 1.77.0 (3fe68eabf 2024-02-29)
release: 1.77.0
commit-hash: 3fe68eabf93cbf3772bbcad09a9206c783e2de3f
commit-date: 2024-02-29
host: x86_64-unknown-linux-gnu
libgit2: 1.7.2 (sys:0.18.2 vendored)
libcurl: 8.5.0-DEV (sys:0.4.70+curl-8.5.0 vendored ssl:OpenSSL/1.1.1w)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: EndeavourOS Rolling Release (rolling) [64-bit]
```

epage:
Can you try this on nightly? i believe this was fixed in #13402

cyqsimon:
Nope. Unless `cargo +nightly` isn't actually using nightly which I doubt is the case.

```bash
$ cargo +nightly -Vv
cargo 1.79.0-nightly (d438c80c4 2024-03-19)
release: 1.79.0-nightly
commit-hash: d438c80c45c24be676ef5867edc79d0a14910efe
commit-date: 2024-03-19
host: x86_64-unknown-linux-gnu
libgit2: 1.7.2 (sys:0.18.2 vendored)
[...]

PR #13655, opened 2024-03-26
fix(add): Preserve comments when updating simple deps 

epage:

Fixes https://github.com/rust-lang/cargo/issues/13645

A case the tests showed but isn't covered here is when a `[features]`
table is created, the dependencies-end comment gets attached to that,
e.g. see cargo_add/overwrite_optional


weihanglo (approved):
>  table is created, the dependencies-end comment gets attached to that,

I am fine with this. Suffix comments aren't common, and it's hard to guess their intent.

weihanglo:
@bors r+

Thanks for the patch!

Issue #10885, opened 2022-07-21
RUSTC_WRAPPER is ignored for `rustc -vV` call


RalfJung:
### Problem

We are setting RUSTC_WRAPPER in order to intercept every invocation of rustc from cargo. However, it looks like cargo will call `rustc -vV` in any case, completely bypassing the RUSTC_WRAPPER env var. Only the RUSTC env var seems to be honored for the `-vV` call. This is causing us a bunch of pain right now since we need to figure out how to best intercept this `-vV` call.

The [docs](https://doc.rust-lang.org/cargo/reference/environment-variables.html) describe both RUSTC and RUSTC_WRAPPER in very similar terms, so I think this is clearly a bug -- both env vars should be effective for the same invocations.

We prefer to use RUSTC_WRAPPER over RUSTC because we need to intercept *all* things that cargo calls (including rustdoc, and running the actual binaries), and RUSTC_WRAPPER makes it easier to figure out what cargo called us for -- if we are invoked with the first argument being `rustc`, we know it was rustc that cargo wanted. (This is in the `cargo-miri` wrapper.)

RalfJung:
The current behavior is annoying for both Miri and bootstrap, and probably other tools -- everyone has to overwrite *both* RUSTC_WRAPPER and RUSTC, which is quite surprising. At least in the case of Miri it is also fragile: what I'd like to do is set RUSTC to `/cargo-miri-phase-rustc/if-you-see-this-some-build-script-forgot-to-apply-the-rustc-wrapper` or something like that, so that when we get invoked as the wrapper we can easily tell that this is what happens. But of course that doesn't work because cargo then calls the non-existing wrapper. This means we have to use a less reliable technique to detect when we are the wrapper. It also means we can't outright *error* when something invokes RUSTC without the wrapper, even though that is almost always a bug -- it is absolutely crucial that the wrapper is applied so that it can adjust the command-line arguments accordingly.

@ehuss this was previously attempted in https://github.com/rust-lang/cargo/pull/10887, but unfortunately rejected because more testing was requested. So it seems like passing the cargo test suite is not sufficient. I have never used sccache, so I don't even know what "using sccache" looks like, let alone testing it sufficiently. Is there documentation of what you think should be done? Is it sufficient to pass rustc bootstrap, or does that not use sccache? sccache is mentioned in `src/ci`, for whatever that is worth.

[...]

PR #10887, opened 2022-07-21
Make sure to also wrap the initial `-vV` invocation

oli-obk:
fixes #10885

This one is fairly annoying to test, as it happens *before* the usual cargo logic (which is also the reason it was forgotten to get wrapped).

PR #13659, opened 2024-03-27
Make sure to also wrap the initial `-vV` invocation

RalfJung:
Fixes https://github.com/rust-lang/cargo/issues/10885 and therefore helps unblock https://github.com/rust-lang/miri/issues/3422.

This ensures that the version info actually matches the compiler that will later be doing the builds.

### DIFF
--- a/tests/testsuite/cargo_env_config.rs
+++ b/tests/testsuite/cargo_env_config.rs
@@ -192,10 +192,13 @@ fn env_applied_to_target_info_discovery_rustc() {
             "src/main.rs",
             r#"
             fn main() {
-                let mut args = std::env::args().skip(1);
+                let mut cmd = std::env::args().skip(1).collect::<Vec<_>>();
+                // This will be invoked twice (with `-vV` and with all the `--print`),
+                // make sure the environment variable exists each time.
                 let env_test = std::env::var("ENV_TEST").unwrap();
                 eprintln!("WRAPPER ENV_TEST:{env_test}");
-                let status = std::process::Command::new(&args.next().unwrap())
+                let (prog, args) = cmd.split_first().unwrap();
+                let status = std::process::Command::new(prog)
                     .args(args).status().unwrap();
                 std::process::exit(status.code().unwrap_or(1));
             }
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1907,7 +1907,10 @@ fn descriptive_pkg_name(name: &str, target: &Target, mode: &CompileMode) -> Stri
 }
 
 /// Applies environment variables from config `[env]` to [`ProcessBuilder`].
-fn apply_env_config(gctx: &crate::GlobalContext, cmd: &mut ProcessBuilder) -> CargoResult<()> {
+pub(crate) fn apply_env_config(
+    gctx: &crate::GlobalContext,
+    cmd: &mut ProcessBuilder,
+) -> CargoResult<()> {
     for (key, value) in gctx.env_config()?.iter() {
         // never override a value that has already been set by cargo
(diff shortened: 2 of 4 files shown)


RalfJung:
All right now in https://github.com/rust-lang/rust/pull/123124 we have a toolchain with this cargo. CI also still works (including Miri).

So now that's ready for experimentation with sccache. Does someone happen to have a setup ready for this? :D

tests/testsuite/cargo_env_config.rs
```diff
@@ -192,10 +192,15 @@ fn env_applied_to_target_info_discovery_rustc() {
             "src/main.rs",
             r#"
             fn main() {
-                let mut args = std::env::args().skip(1);
-                let env_test = std::env::var("ENV_TEST").unwrap();
-                eprintln!("WRAPPER ENV_TEST:{env_test}");
-                let status = std::process::Command::new(&args.next().unwrap())
+                let mut cmd = std::env::args().skip(1).collect::<Vec<_>>();
+                if cmd.get(1).map(|s| &**s) == Some("-vV") {
+                    // This is the version query, not the target info query, so skip this.
```
RalfJung, line 197: I guess this raises the question, should the env stuff also be set for the `-vV` query?

RalfJung:
I have installed sccache (`apt install sccache`), set the env var (`export RUSTC_WRAPPER=$(which sccache)`), and then tried building a few crates with the toolchain from https://github.com/rust-lang/rust/pull/123124:
```
rustup-toolchain-install-master -c cargo -- c656380a167bb05da121130f7364bfe5785f1c8b
cargo +c656380a167bb05da121130f7364bfe5785f1c8b build
```
That seems to work fine. I can see via `sccache --show-stats` that sccache was invoked. I can see via `RUSTC_WRAPPER=/bin/false cargo +c656380a167bb05da121130f7364bfe5785f1c8b build` that the `-vV` call *is* being wrapped:
```
error: process didn't exit successfully: `/bin/false /home/r/.rustup/toolchains/c656380a167bb05da121130f7364bfe5785f1c8b/bin/rustc -vV` (exit status: 1)
```

So... I think we're all good? @ehuss anything else you think should be checked?

(We also still need a decision about applying the `env` keys though, see above.)

weihanglo:
r? @ehuss

since you seems to involve in the discussion already.

tests/testsuite/cargo_env_config.rs
heisen-li: Sorry, did you mean to add an environment variable similar to RUSTC_VERSION? Is there any benefit to this?

RalfJung: This is about env vars configured with `[env]` in `config.toml`.

ehuss: Yea, it probably makes sense to also set the env vars. Can you add that? It looks like `apply_env_config` might need to be `pub(crate)` or something to make that work.

ehuss:
Thanks! There looks like there might be a few more things to get this ready to merge:

* Add an explicit test for this. There are a few ways to go about this, either what is suggested in https://github.com/rust-lang/cargo/pull/10887#issuecomment-1192046603, or just have a separate test that has a custom wrapper that checks for the appropriate behavior, or do something similar to the `rustc_info_cache` test which sets `CARGO_LOG=cargo::util::rustc=debug` and check for the `[..]running [..]my_wrapper rustc -Vv[..]` or similar. (That last option might be the easiest if it works.)
* Update the PR description to remove the "experimentation" description and include an explanation of why the change is being made.
* Squash the commits.

RalfJung:
All right, should be all done.

ehuss:
Thanks!

@bors r+

Issue #11777, opened 2023-02-27
Documentation for `documentation` field is misleading


RalfJung:
The [docs for the `documentation` field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-documentation-field) say

> If no URL is specified in the manifest file, [crates.io](https://crates.io/) will automatically link your crate to the corresponding [docs.rs](https://docs.rs/) page.

This sounds as if setting no URL is equivalent to setting the docs.rs URL. That is however not the case: in the crates.io [search result list](https://crates.io/search?q=mycrate), the "Documentation" link will only show up if `documentation` is explicitly set in the toml file.

I am not sure if this is a bug in the cargo docs or the crates.io frontend, but it is certainly surprising.

epage:
Seeing as the individual crate pages will list documentation, I would personally chalk this up as a crates.io bug.

From #11685, my suspicion is that this is either an optimization or an oversight.  Apparently, adding the docs.rs link to the crate's page is done via Javascript and I imagine doing that for every crate on a search page might be a bit much.

RalfJung:
This should be reopened since https://github.com/rust-lang/cargo/pull/14562 got reverted.

PR #13660, opened 2024-03-27
clarify default behavior of `documentation` field

RalfJung:
Cc https://github.com/rust-lang/cargo/issues/11777, https://github.com/rust-lang/crates.io/issues/1484

### DIFF
--- a/src/doc/src/reference/manifest.md
+++ b/src/doc/src/reference/manifest.md
@@ -205,9 +205,12 @@ description = "A short description of my package"
 ### The `documentation` field
 
 The `documentation` field specifies a URL to a website hosting the crate's
-documentation. If no URL is specified in the manifest file, [crates.io] will
+documentation.
+
+If no URL is specified in the manifest file, [crates.io] will
 automatically link your crate to the corresponding [docs.rs] page when the
 documentation has been built and is available (see [docs.rs queue]).
+However, this automatic link appears only on the crate page, not in search results.
 
 ```toml
 [package]


epage:
For myself, I feel like its a bug in the system that we have to add so much nuance here. If that bug isn't being fixed, should we just not reference this and encourage people to set the links?

ehuss:
Yea, I would agree this seems like weird bug to document. I'm also a little confused, as it seems like the accepted solution should be relatively trivial (something like `documentation.or_else(|| format!("https://docs.rs/{name}/latest/{name}"))` in the search code).

RalfJung:
Yeah, [agreed](https://github.com/rust-lang/crates.io/issues/1484#issuecomment-2022690205).

weihanglo:
If we agree on seeing a fix on crates.io or other services, should this be closed?

weihanglo:
Close as it is more like a buggy behavior in crates.io that we don't want to document. Thanks everyone for the discussion.

RalfJung:
Does the cargo team define whether this is a bug in crates.io? That seems like a crates.io decision to me. https://github.com/rust-lang/crates.io/issues/1484 is marked as "enhancement".

Meanwhile, we are doing our users a disservice by having documentation that actively contradicts reality.

RalfJung:
Even if this *is* a bug in crates.io, it is clearly a long-standing issue and hard to fix. So telling users in the cargo docs that they can just leave the docs field empty if they use docs.rs (which is effectively what the docs are doing) is not helping.

I have to say I am quite disappointed with how this has been handled -- we have docs that actively confuse users and lead them in the wrong direction, and instead of trying to fix this we're squabbling about whether this is a bug or not?

epage:
> Meanwhile, we are doing our users a disservice by having documentation that actively contradicts reality.

Its fairly close to reality

> Even if this is a bug in crates.io, it is clearly a long-standing issue and hard to fix. So telling users in the cargo docs that they can just leave the docs field empty if they use docs.rs (which is effectively what the docs are doing) is not helping.

While its long standing, there has been recent progress to fixing it.  iiuc the blocker for fixing it is on docs.rs's side and the PR for that was started this summer (rust-lang/docs.rs#2533).

I mentioned that I was fine with removal if this was going to the status quo but that doesn't appear to be the case, so I lean towards keeping the documentation.

RalfJung:
It's far enough from reality to cause actual problems: many crates have no "Documentation" link in the search results, adding an entirely unnecessary extra click for the quite common flow of "go search some crate to find its docs".

epage:
If the concern is about browsing crates.io, at this point, it would be a major undertaking to get people to notice the documentation change to specify `package.documentation` and for people to act on it.  It doesn't seem worthwhile to make that shift when we'll just switch back again.

PR #13665, opened 2024-03-28
docs: clarify `--locked` ensures Cargo uses dependency versions in lockfile

### LINKED ISSUE #11143: `cargo build --locked` description is confusing
### Problem

**In short:**
`cargo build --locked` description claims it requires the Cargo.lock to be up to date, but building works on an outdated file anyway.

**Long version:**
The description in `cargo build --help` about `--locked` parameter says:
> --locked                    Require Cargo.lock is up to date

And the information in [Cargo documentation](https://doc.rust-lang.org/cargo/commands/cargo-build.html#manifest-options) says:
> --frozen
> --locked
> Either of these flags requires that the Cargo.lock file is up-to-date. If the lock file is missing, or it needs to be updated, Cargo will exit with an error. (...)

When running `cargo outdated` on [my project](https://github.com/ArekPiekarz/struct-formatter) I get information about outdated packages:
```
$ cargo outdated
Name                                Project  Compat  Latest  Kind    Platform
----                                -------  ------  ------  ----    --------
clap_derive->proc-macro2            1.0.43   1.0.44  1.0.44  Normal  ---
proc-macro-error->proc-macro2       1.0.43   1.0.44  1.0.44  Normal  ---
proc-macro-error-attr->proc-macro2  1.0.43   1.0.44  1.0.44  Normal  ---
quote->proc-macro2                  1.0.43   1.0.44  1.0.44  Normal  ---
syn->proc-macro2                    1.0.43   1.0.44  1.0.44  Normal  ---
```

But when I try to build it with the `--locked` option, it works anyway, using outdated dependency:
```
$ cargo build --locked
(...)
Compiling proc-macro2 v1.0.43
(...)
Finished dev [unoptimized + debuginfo] target(s) in 11.50s
```
BTW building without that option uses the same version of proc-macro2.

### Steps

1. git clone https://github.com/ArekPiekarz/struct-formatter.git
2. cd struct-formatter
3. It should be at commit 071febcdb1f0b875d030b40094d02260f4469c16. If not, git checkout \<revision\>.
4. Confirm with `cargo outdated` that at least one dependency is outdated.
5. cargo build --locked

**Expected:**
According to the description, the build should fail.

**Actual:**
The build succeeds, using outdated dependency.

### Possible Solution(s)

Perhaps the description should say what does it mean that a package is "up to date". Or perhaps it should read "use versions from Cargo.lock or exit with error if it doesn't exist".

### Notes

_No response_

### Version

```text
cargo 1.64.0 (387270bc7 2022-09-16)
release: 1.64.0
commit-hash: 387270bc7f446d17869c7f208207c73231d6a252
commit-date: 2022-09-16
os: Manjaro 22.0.0 (Sikaris) [64-bit]
```

weihanglo:

Fixes #11143

This is an attempt to remove the `up-to-date` mentioned in `--locked`.
Up-to-date is quite confusing, especially when MSRV resolution is out,
dependency version may lag behind more often.

`--frozen` is now documented as an equivalent of `--locked` + `--offline`.

```
cargo run -- help build 
# and read the man page

cargo build --help
# and check the help text
```

### DIFF
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -640,17 +640,17 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
                 .value_parser(clap::builder::ValueParser::path_buf()),
         )
         .arg(
-            flag("frozen", "Require Cargo.lock and cache to be up-to-date")
-                .help_heading(heading::MANIFEST_OPTIONS)
-                .global(true),
-        )
-        .arg(
             flag("locked", "Require Cargo.lock to be up-to-date")
                 .help_heading(heading::MANIFEST_OPTIONS)
                 .global(true),
         )
         .arg(
             flag("offline", "Run without accessing the network")
+                .help_heading(heading::MANIFEST_OPTIONS)
+                .global(true),
+        )
+        .arg(
+            flag("frozen", "Equivalent to specifying both --locked and --offline")
                 .help_heading(heading::MANIFEST_OPTIONS)
                 .global(true),
         )
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -640,7 +640,7 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
                 .value_parser(clap::builder::ValueParser::path_buf()),
         )
         .arg(
-            flag("locked", "Require Cargo.lock to be up-to-date")
+            flag("locked", "Assert that `Cargo.lock` will remain unchanged")
                 .help_heading(heading::MANIFEST_OPTIONS)
                 .global(true),
         )
--- a/src/doc/man/generated_txt/cargo-add.txt
+++ b/src/doc/man/generated_txt/cargo-add.txt
@@ -171,16 +171,14 @@ OPTIONS
        -p spec, --package spec
            Add dependencies to only the specified package.
 
-       --frozen, --locked
-           Either of these flags requires that the Cargo.lock file be
-           up-to-date. If the lock file is missing, or it needs to be updated,
-           Cargo will exit with an error. The --frozen flag also prevents Cargo
-           from attempting to access the network to determine if it is
-           out-of-date.
+       --locked
+           Requires the Cargo.lock file be up-to-date. If the lock file is
+           missing, or it needs to be updated due to changes in the Cargo.toml
+           file, for example a new dependency is added, Cargo will exit with an
+           error.
 
-           These may be used in environments where you want to assert that the
-           Cargo.lock file is up-to-date (such as a CI build) or want to avoid
-           network access.
+           It may be used in environments where you want to assert that the
+           Cargo.lock file is up-to-date (such as a CI build).
 
        --offline
            Prevents Cargo from accessing the network for any reason. Without
@@ -196,6 +194,9 @@ OPTIONS
 
            May also be specified with the net.offline config value
            <https://doc.rust-lang.org/cargo/reference/config.html>.
+
+       --frozen
+           Equivalent to specifying both --locked and --offline.
 
    Common Options
        +toolchain
--- a/src/doc/man/generated_txt/cargo-add.txt
+++ b/src/doc/man/generated_txt/cargo-add.txt
@@ -172,13 +172,18 @@ OPTIONS
            Add dependencies to only the specified package.
 
        --locked
-           Requires the Cargo.lock file be up-to-date. If the lock file is
-           missing, or it needs to be updated due to changes in the Cargo.toml
-           file, for example a new dependency is added, Cargo will exit with an
-           error.
+           Asserts that the exact same dependencies and versions are used as
+           when the existing Cargo.lock file was originally generated. Cargo
+           will exit with an error when either of the following scenarios
+           arises:
 
-           It may be used in environments where you want to assert that the
-           Cargo.lock file is up-to-date (such as a CI build).
+           o  The lock file is missing.
+
+           o  Cargo attempted to change the lock file due to a different
+              dependency resolution.
+
+           It may be used in environments where deterministic builds are
+           desired, such as in CI pipelines.
 
        --offline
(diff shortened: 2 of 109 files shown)


src/doc/man/generated_txt/cargo-add.txt
```diff
@@ -171,16 +171,19 @@ OPTIONS
...
            Add dependencies to only the specified package.
 
-       --frozen, --locked
-           Either of these flags requires that the Cargo.lock file be
-           up-to-date. If the lock file is missing, or it needs to be updated,
-           Cargo will exit with an error. The --frozen flag also prevents Cargo
-           from attempting to access the network to determine if it is
-           out-of-date.
+       --locked
+           Ensures that Cargo uses the exact version of every dependency
```
weihanglo, line 175: Still not too happy with "Cargo **uses** the exact version of every dependency", which optional dependencies are not used. Suggestions?

weihanglo:
```
---- ssh::bundled_github_works stdout ----
thread 'ssh::bundled_github_works' panicked at tests/testsuite/ssh.rs:522:10:

test failed running `/Users/runner/work/cargo/cargo/target/debug/cargo fetch`
error: stderr did not match:
1   1         Updating git repository `ssh://git@github.com/rust-lang/bitflags.git`
    2    +warning: spurious network error (3 tries remaining): failed to connect to github.com: Operation timed out; class=Os (2)
    3    +warning: spurious network error (2 tries remaining): failed to connect to github.com: Operation timed out; class=Os (2)
2   4     error: failed to get `bitflags` as a dependency of package `foo v0.1.0 (/Users/runner/work/cargo/cargo/target/tmp/cit/t2886/foo)`
3   5     
4   6     Caused by:
5   7       failed to load source for dependency `bitflags`
6   8     
7   9     Caused by:
8   10      Unable to update ssh://git@github.com/rust-lang/bitflags.git?tag=1.3.2
9   11    
10  12    Caused by:
11  13      failed to clone into: /Users/runner/work/cargo/cargo/target/tmp/cit/t2886/home/.cargo/git/db/bitflags-[..]
12  14    
13  15    Caused by:
14  16      failed to authenticate when downloading repository
15  17    
16  18      * attempted ssh-agent authentication, but no usernames succeeded: `git`
17  19    
18  20      if the git CLI succeeds then `net.git-fetch-with-cli` may help here
19  21      https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli
20  22    
21  23    Caused by:
22  24      no authentication methods succeeded
```

This seems to be a spurious network error.

@bors try

src/doc/man/generated_txt/cargo-add.txt
weihanglo: Not sure but "locking" sounds like Cargo will lock something but this is Cargo being locked to Cargo.lock

src/bin/cargo/cli.rs
```diff
@@ -640,17 +640,17 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
...
         )
         .arg(
-            flag("locked", "Require Cargo.lock to be up-to-date")
+            flag("offline", "Run without accessing the network")
                 .help_heading(heading::MANIFEST_OPTIONS)
                 .global(true),
         )
         .arg(
-            flag("offline", "Run without accessing the network")
+            flag("frozen", "Equivalent to specifying both --locked and --offline")
```
epage, line 653: > This is an attempt to remove the up-to-date mentioned in --locked.
Up-to-date is quite confusing, especially when MSRV resolution is out,
dependency version may lag behind more often.

I actually prefer the original wording for `--locked`.  Its not saying its up-to-date as in `cargo update` but that it is synchronized with `Cargo.toml`.  

With the new wording, it isn't clear that adding a new dependency will fail with `--locked` (there isn't a version in `Cargo.lock`, so its undefined according to this text(

Also, `require` makes it sound like it will fail while `ensure` makes it sound like it could fail or it could correct things (ie the opposite of `--locked`).

epage:
> This seems to be a spurious network error.

Had a similar spurious error yesterday.

btw in the github actions view, you can tell it to re-run failed jobs.

src/bin/cargo/cli.rs
weihanglo: > Its not saying its up-to-date as in cargo update but that it is synchronized with Cargo.toml.

That's the problem. Up-to-date can be interpreted differently. I was there and also confused until I learnt how `--locked` works internally in Cargo. Maybe something like

"Enforce only use of dependency versions captured in Cargo.lock"

Still not too happy with it.

epage: Some ideas for brainstorming
```
Require Cargo.lock be up-to-date with Cargo.toml
Require all dependency versions to be specified in Cargo.lock
Require Cargo.lock to specify every dependency version
Fail if Cargo.lock would change
```

weihanglo: > Fail if Cargo.lock would change


Neat! Like this one. Updated this and the man pages.

src/doc/man/generated_txt/cargo-add.txt
```diff
@@ -171,16 +171,19 @@ OPTIONS
...
            Add dependencies to only the specified package.
 
-       --frozen, --locked
-           Either of these flags requires that the Cargo.lock file be
-           up-to-date. If the lock file is missing, or it needs to be updated,
-           Cargo will exit with an error. The --frozen flag also prevents Cargo
-           from attempting to access the network to determine if it is
-           out-of-date.
+       --locked
+           Asserts that the exact same dependencies and versions are used as
```
weihanglo, line 175: shamelessly copied from 

https://github.com/rust-lang/cargo/blob/07253b7ea640e8466408790bb6cad4440eb9531f/src/doc/src/faq.md?plain=1#L114-L115

src/bin/cargo/cli.rs
weihanglo: Sounds good. Updated per suggestion

Assert is also a common and strong enough word for this purpose.

PR #13666, opened 2024-03-28
refactor(package): Simplify getting of published Manifest

epage:

This is a parallel effort to #13664 in an effort to #13456

This abstracts away the logic for getting the published version of `Cargo.toml` so we can more easily change the APIs that were previously exposed


src/cargo/util/toml/mod.rs
```diff
@@ -218,10 +218,32 @@ fn warn_on_unused(unused: &BTreeSet<String>, warnings: &mut Vec<String>) {
     }
 }
 
+pub fn prepare_for_publish(me: &Package, ws: &Workspace<'_>) -> CargoResult<Package> {
```
weihanglo, line 221: While not necessary, I feel like `prepare_for_publish` and friends could be moved to a dedicated module.

weihanglo (approved):

weihanglo:
The comment is not a blocker, so

@bors r+

Thanks for the refactor!

src/cargo/util/toml/mod.rs
epage: I've been going back and forth on it, so I've just left it for now.

The main benefit for keeping it here is to keep awareness of it when adding new features to the manifest in the hope that we remember that they might impact this so we update it.

PR #13669, opened 2024-03-30
Protect .cargo-ok on case-insensitive file systems

kornelski:
Tar decompression is trying to forbid `.cargo-ok` from being expanded from the archive, but it's performing an exact string check, which is effective only on case-sensitive file systems. On macOS and Windows a `.CARGO-OK` (or some other combination of upper case letters) can create a file that has the same path as `.cargo-ok`.


weihanglo (commented):
I believe [this line](
https://github.com/rust-lang/cargo/blob/3b43aa875fa262f53bdcc5ad4c0f677802630186/src/cargo/sources/registry/mod.rs#L690
) already protects them from overwrites, though your patch seems also reasonable. Could you add a test to prevent future regression?

epage:
non-blocking: I'm curious how someone gets in this situation since this is cargo managed and those are case preserving systems.

epage:
When adding the test, please make it its own commit first and having passing, showing the  bad behavior. Then in the commit with your fix, it will be updated to show the change in behavior.

PR #13677, opened 2024-04-01
Document variance in strip

workingjubilee:
Two variances in strip will remain, even after MSVC shenanigans settle:
- Its default based on profile.
- That it can sometimes make backtraces nonsensical.

In my experience, people will complain to me if backtraces are fucked up, but sometimes it was their own decision to make the backtrace really bad. Inform them of this. Hopefully this also encourages people to report if their backtraces regress in some platform-dependent way without enabling `strip = "symbols"`, which we would want to know about.

### DIFF
--- a/src/doc/src/reference/profiles.md
+++ b/src/doc/src/reference/profiles.md
@@ -117,7 +117,9 @@ strip = "debuginfo"
 ```
 
 Possible string values of `strip` are `"none"`, `"debuginfo"`, and `"symbols"`.
-The default is `"none"`.
+The default is `"none"` by default, and `"debuginfo"` in release profile.
+Be aware that stripping `"symbols"` may make backtraces incomprehensible,
+depending on the platform.
 
 You can also configure this option with the boolean values `true` or `false`.
 `strip = true` is equivalent to `strip = "symbols"`. `strip = false` is
@@ -293,7 +295,7 @@ The default settings for the `release` profile are:
 opt-level = 3
 debug = false
 split-debuginfo = '...'  # Platform-specific.
-strip = "none"
+strip = "debuginfo"
 debug-assertions = false
 overflow-checks = false
 lto = false


src/doc/src/reference/profiles.md
```diff
@@ -117,7 +117,9 @@ strip = "debuginfo"
 ```
 
 Possible string values of `strip` are `"none"`, `"debuginfo"`, and `"symbols"`.
-The default is `"none"`.
+The default is `"none"` by default, and `"debuginfo"` in release profile.
```
ehuss, line 120: Similar to #13638, I don't think this is entirely correct. The default strip behavior depends on if debug is set (including overrides). 

In #13257, we somewhat intentionally did not document this, as it *should* be a completely internal matter that should be transparent to the user. It is only running `strip` to deal with the particular way the standard library is distributed.

I'm not yet entirely convinced this should be documented, since it is a relatively complicated thing to explain to the user (something they shouldn't even need to know about) which could just contribute to confusion.

src/doc/src/reference/profiles.md
```diff
@@ -117,7 +117,9 @@ strip = "debuginfo"
...
+Be aware that stripping `"symbols"` may make backtraces incomprehensible,
+depending on the platform.
```
ehuss, line 122: For the most part, the cargo documentation doesn't dive into the particulars of explaining the different options and their consequences. That documentation should generally live in the rustc book (so we don't have to duplicate it in two remote places). Adding warnings about the consequences of stripping symbols would probably be appropriate over there (and seems good to include).

workingjubilee: It is my experience that people often do not actually click through, simply put.

src/doc/src/reference/profiles.md
workingjubilee: ...Frankly, I don't know why there are even two options if one affects how the other is interpreted.

workingjubilee: In any case, I do not believe these behaviors are actually completely internal matters that are transparent to the user, as the effective behavior seems to change wildly based on the target, host, or available binaries, e.g.
- https://github.com/rust-lang/rust/issues/123114
- https://github.com/rust-lang/rust/issues/122902
- https://github.com/oxidecomputer/helios/issues/147

Perhaps we can make it one with https://github.com/rust-lang/rust/issues/123151 but it isn't now.

orlp: @ehuss The current default lying about being `"none"` made it much harder today to debug a problem where the compiler is stripping all **all** symbols on MacOS if `debug = false` is specified: https://github.com/rust-lang/rust/issues/135028.

workingjubilee: @ehuss What should the behavior be when we replace a strip binary with Rust library code?

workingjubilee: There isn't actually anything "transparent" here when they can set these options, examine a backtrace, and then set these options again, and examine a different backtrace.

weihanglo:
> Would it may be better to just mention that by default cargo will choose the value based on the other options without getting into the details? That makes it clear that when a specific value is needed it should be set explicitly and otherwise using the default is fine.

I am not sure how to accurately document that. There are two things: The default value for profile inheritance. This is static. And the final value passing to rustc at runtime.

0x9fff00:
I've made a new attempt at documenting this in #15680

epage:
We generally encourage Issues before PRs.  One reason is to centralize the conversation.  We've now had 3 PRs and 1 Issue open for this.  I'm going to go ahead and close this in favor of the issue, #16670.

PR #13682, opened 2024-04-01
Maintain sorting of dependency features

### LINKED ISSUE #13578: `cargo add` should maintain sorting of features
### Problem

When you run `cargo add <package> -F <feature>`  when the dependency already has features that are sorted alphabetically, the new features get added to the back instead of in alphabetical order. This is unexpected and different from what happens when you add a dependency to a sorted dependency list, where it respects being sorted ([cargo code](https://github.com/rust-lang/cargo/blob/7065f0ef4aa267a7455e1c478b5ccacb7baea59c/src/cargo/ops/cargo_add/mod.rs#L106)).

[Example](https://github.com/rust-lang/annotate-snippets-rs/commit/c8d8c65232cf988caa1a36d6cda2ee674548bc0b#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R37)

Reproduction:
1. Add below to `[dependencies]`
    ```toml
    snapbox = { version = "0.5.8", features = ["diff", "harness", "path", "term-svg"] }
    ```
2. Run `cargo add snapbox -F cmd`
3. See that `cmd` is not added to the front of the list

### Proposed Solution

Cargo should keep features sorted alphabetically when encountered.

### Notes

_No response_

Flowrey:
Fixes #13578


epage:
Could you squash that last commit into the one before so it passes tests?

PR #13686, opened 2024-04-01
fix: bash completion fallback in `nounset` mode

scop:
With the shell in `nounset` mode, the intended fallback to filename completion provokes an error:

```
$ cargo foo <TAB>bash: !opt_var: unbound variable
```


ehuss:
Thanks!

@bors r+

Issue #13285, opened 2024-01-11
Tracking Issue for `-Zgit`


weihanglo:
### Summary

Original issue: #1171. See also #11813
Implementation: #13252
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#my-feature

`-Zgit` is a collection of new Git features in Cargo. Currently we have experiments on

* Shallow clones for git dependencies
* Shallow clones for registry indexes

### Unresolved Issues

* [ ] No native support of local `file` shallow clones for either gitoxide or libgit2
  * https://github.com/Byron/gitoxide/issues/734
  * https://github.com/libgit2/libgit2/issues/6634
* [ ] Add tests for git2 shallow clones. This is blocked on the support of `file:` protocol.
* [ ] Is `*-shallow` directory layout optimal? Can we avoid that but remain compatible?

### Future Extensions

See "Future Extensions" in #11813. Most of them depend on upstream features in libgit2.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

NoelJacob:
My `~/.cargo/config.toml` looks like this:
```
git.shallow-index = true
git.shallow-deps = true
gitoxide.fetch = true
gitoxide.checkout = true
```
But on running `cargo help` I get :
```
>  cargo help
error: missing field `internal_use_git2`

>  cargo help
warning: unused config key `unstable.git.shallow-index` in `/home/noel/.cargo/config.toml`
warning: unused config key `unstable.git.shallow-deps` in `/home/noel/.cargo/config.toml`
error: missing field `shallow_index`
```
Keep repeating `cargo <command>` and the error seems to repeat between the two. Passing `-Zgit -Zgitoxide` seems to not give error.

weihanglo:
Under `~/.cargo/git`, check if the git clone contains a `-shallow` suffix.

```console
~/.cargo/git $ tree -L 2
./
├── checkouts/
│  └── empty-library-0161899a0cd3bb43-shallow/
└── db/
   └── empty-library-0161899a0cd3bb43-shallow/
```

As of this writing, you need this in your `.cargo/config.toml`:

```toml
[unstable.git]
shallow_index = true
shallow_deps = true
[unstable.gitoxide]
[...]

Issue #11813, opened 2023-03-09
Tracking Issue for `-Zgitoxide`


Byron:
### Summary

RFC: None
Original issue: None
Implementation: #11448
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#gitoxide

With the ‘gitoxide’ unstable feature, all or the the specified git operations will be performed by the gitoxide crate instead of git2.

### Unresolved Issues

* [ ] support for the `file://` protocol without relying on the `git` binary
* [ ] support of the `ssh://` protocol without relying on the `ssh` binary
* [ ] double-check and validate that authentication works similarly to `git2` or if unfeasible, assure users have a migration path. How to deal with this correctly is still to be determined.

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

epage:
> transitive dependency tempfile is pinned to ~3.3.0 temporarily (see [gix-tempfile@4.1.1](https://github.com/Byron/gitoxide/releases/tag/gix-tempfile-v4.1.1))

This surprises me and the links don't quite make the motivation for this clear.

From my experience
- Using anything but `^` for general dependencies like `tempfile` can be [ecosystem-breaking](https://www.reddit.com/r/rust/comments/p8clcx/how_to_fix_cargo_dependency_issue/)
- Generally workarounds like this should be handled in `Cargo.lock`

weihanglo:
Hmm… agreed! `gix-tempile` previously depends on `3.4.0`, so alternatively it could just depend on `^3.3.0` and we handle it in rust-lang/rust. I cannot guarantee but believe bootstrap team will fix it in the near future :).

ehuss:
I would agree that using `^3.3.0` would be the more appropriate requirement. I would generally avoid `~` in most situations.

>  I cannot guarantee but believe bootstrap team will fix it in the near future

I do not think anyone is looking at fixing this particular issue. The fix from https://github.com/rust-lang/rust/pull/106610 didn't seem to work for me, though I'm not sure why.

If cargo moves to a separate workspace, then this particular problem might not be an issue.

Byron:
[...]

PR #13687, opened 2024-04-02
Allows the default git/gitoxide configuration to be obtained from the ENV and config

linyihai:
The default git/gitoxide feautures config can be obtained througt `-Zgit` and `-Zgitoxide`.  
However, it cannot be obtained from environment variables or configurations.
It's not very ergonomic.

The previous commit explained the staus quo, the next commit addressed the problem.

### Additional information
Inspired by https://github.com/rust-lang/cargo/issues/11813#issuecomment-1817517629
See also https://github.com/rust-lang/cargo/issues/13285#issuecomment-2016875459

### Change of usage

1. Mirror Zgit/Zgitoxide when they parsed as string

Specify the feature you like
```
CARGO_UNSABLE_GIT='shallow-deps' cargo fetch
cargo fetch --config "unstable.git='shallow-deps'"
cargo fetch -Zgit="shallow-deps"
```

2. Specify partial fields when parsed as table

```
CARGO_UNSTABLE_GITOXIDE_FETCH=true cargo fetch

```

The rest fields will use Rust default value. That said, checkout and internal_use_git2 will be false.

Besides, you can pass true to the whole feature to specify the pre-defined features.

```
CARGO_UNSTABLE_GITOXIDE=true cargo fetch

```

### DIFF
--- a/src/cargo/core/features.rs
+++ b/src/cargo/core/features.rs
@@ -764,7 +764,9 @@ unstable_cli_options!(
     dual_proc_macros: bool = ("Build proc-macros for both the host and the target"),
     features: Option<Vec<String>>,
     gc: bool = ("Track cache usage and \"garbage collect\" unused files"),
+    #[serde(deserialize_with = "deserialize_git_features")]
     git: Option<GitFeatures> = ("Enable support for shallow git fetch operations"),
+    #[serde(deserialize_with = "deserialize_gitoxide_features")]
     gitoxide: Option<GitoxideFeatures> = ("Use gitoxide for the given git interactions, or all of them if no argument is given"),
     host_config: bool = ("Enable the `[host]` section in the .cargo/config.toml file"),
     minimal_versions: bool = ("Resolve minimal dependency versions instead of maximum"),
@@ -874,7 +876,8 @@ where
     ))
 }
 
-#[derive(Debug, Copy, Clone, Default, Deserialize)]
+#[derive(Debug, Copy, Clone, Default, Deserialize, Ord, PartialOrd, Eq, PartialEq)]
+#[serde(default)]
 pub struct GitFeatures {
     /// When cloning the index, perform a shallow clone. Maintain shallowness upon subsequent fetches.
     pub shallow_index: bool,
@@ -883,12 +886,71 @@ pub struct GitFeatures {
 }
 
 impl GitFeatures {
-    fn all() -> Self {
+    pub fn all() -> Self {
         GitFeatures {
             shallow_index: true,
             shallow_deps: true,
         }
     }
+
+    fn expecting() -> String {
+        let fields = vec!["`shallow-index`", "`shallow-deps`"];
+        format!(
+            "unstable 'git' only takes {} as valid inputs",
+            fields.join(" and ")
+        )
+    }
+}
+
+fn deserialize_git_features<'de, D>(deserializer: D) -> Result<Option<GitFeatures>, D::Error>
+where
+    D: serde::de::Deserializer<'de>,
+{
+    struct GitFeaturesVisitor;
+
+    impl<'de> serde::de::Visitor<'de> for GitFeaturesVisitor {
+        type Value = Option<GitFeatures>;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+            formatter.write_str(&GitFeatures::expecting())
+        }
+
+        fn visit_bool<E>(self, v: bool) -> Result<Self::Value, E>
+        where
+            E: serde::de::Error,
+        {
+            if v {
+                Ok(Some(GitFeatures::all()))
+            } else {
+                Ok(None)
+            }
+        }
+
+        fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
+        where
+            E: serde::de::Error,
+        {
+            Ok(parse_git(s.split(",")).map_err(serde::de::Error::custom)?)
+        }
+
+        fn visit_some<D>(self, deserializer: D) -> Result<Self::Value, D::Error>
+        where
+            D: serde::de::Deserializer<'de>,
+        {
+            let git = GitFeatures::deserialize(deserializer)?;
+            Ok(Some(git))
+        }
+
+        fn visit_map<V>(self, map: V) -> Result<Self::Value, V::Error>
+        where
+            V: serde::de::MapAccess<'de>,
+        {
+            let mvd = serde::de::value::MapAccessDeserializer::new(map);
+            Ok(Some(GitFeatures::deserialize(mvd)?))
+        }
+    }
+
+    deserializer.deserialize_any(GitFeaturesVisitor)
 }
 
 fn parse_git(it: impl Iterator<Item = impl AsRef<str>>) -> CargoResult<Option<GitFeatures>> {
@@ -903,16 +965,15 @@ fn parse_git(it: impl Iterator<Item = impl AsRef<str>>) -> CargoResult<Option<Gi
             "shallow-index" => *shallow_index = true,
             "shallow-deps" => *shallow_deps = true,
             _ => {
-                bail!(
-                    "unstable 'git' only takes 'shallow-index' and 'shallow-deps' as valid inputs"
-                )
+                bail!(GitFeatures::expecting())
             }
         }
     }
     Ok(Some(out))
 }
 
-#[derive(Debug, Copy, Clone, Default, Deserialize)]
+#[derive(Debug, Copy, Clone, Default, Deserialize, Ord, PartialOrd, Eq, PartialEq)]
+#[serde(default)]
 pub struct GitoxideFeatures {
     /// All fetches are done with `gitoxide`, which includes git dependencies as well as the crates index.
     pub fetch: bool,
@@ -926,7 +987,7 @@ pub struct GitoxideFeatures {
 }
 
 impl GitoxideFeatures {
-    fn all() -> Self {
+    pub fn all() -> Self {
         GitoxideFeatures {
             fetch: true,
             checkout: true,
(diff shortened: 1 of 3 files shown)


epage:
I assume this is meant to resolve the immediate part of #13688 (leaving aside the more general problem).  Is that right?

src/cargo/core/features.rs
```diff
@@ -908,15 +914,19 @@ fn parse_git(it: impl Iterator<Item = impl AsRef<str>>) -> CargoResult<Option<Gi
 #[derive(Debug, Copy, Clone, Default, Deserialize)]
 pub struct GitoxideFeatures {
     /// All fetches are done with `gitoxide`, which includes git dependencies as well as the crates index.
+    #[serde(default = "default_true")]
```
epage, line 917: I assume these were chosen to align with `GitoxideFeatures::safe()` and to mirror if you do `-Zgitoxide`

However, this breaks down if you do `CARGO_UNSTABLE_GITOXIDE_FETCH=true`, then almost everything else becomes enabled (we should have a test to show this case).

Some other challenges with these defaults
- Unclear what the motivation is (could have comment pointing to `GitoxideFeature::safe`)
- Challenge in keeping in sync with `GitoxideFeature::safe`

Should we instead just default everything to false?

Nemo157: What I would love to just work is `CARGO_UNSTABLE_GITOXIDE=fetch,list_files` mirroring the CLI flag (which would also imply you could have `unstable.gitoxide = "fetch,list_files"` in the config file).

weihanglo: Yeah maybe we can make it a comma separated list to reduce the complexity. 

Wonder do people really use individual gitoxide features? I always assume that people would just enable them as a whole, though gitoxide author recommends keeping them separately: <https://github.com/rust-lang/cargo/pull/13252#discussion_r1447868466>.

linyihai: > What I would love to just work is CARGO_UNSTABLE_GITOXIDE=fetch,list_files and CARGO_UNSTABLE_GITOXIDE=true mirroring the CLI flag (which would also imply you could have unstable.gitoxide = "fetch,list_files" in the config file).

So what we defined as table is going to be replaced with a string? It makes sense,  I have no objection to this, but it is not clear that it will break the configuration of the previous version？

Nemo157: It's possible to support both, via `#[serde(deserialize_with = ...)]` and a custom deserializer that handles a string/list or a struct, like 

https://github.com/rust-lang/cargo/blob/88d55a89e0cb596a4bda53ce194b6753c08a3798/src/cargo/util/context/mod.rs#L2673

ehuss:
r? weihanglo

linyihai:
In my new commit, a string can be parsed as git/gitoxide features in ENV and Config.
You can pass 'all' to enable predefined features.

```
CARGO_UNSABLE_GIT='all' cargo fetch
CARGO_UNSABLE_GIT='shallow-deps' cargo fetch
cargo fetch --config "unstable.git='shallow-deps'"
cargo fetch --config "unstable.git='all'"
```

In Config, the git/gitoxide can be defined as 

```
# as string
[unstable]
git = 'all'

# still can be defined as table, but all fields needed
[unstable.git]
shallow_deps = true
shallow_index = true

```

But the defect is that the fields as ENV will be ignored.
```
CARGO_UNSABLE_GIT_SHALLOW_DEPS CARGO_UNSABLE_GIT_SHALLOW_INDEX cargo fetch
```
In ENV, it seems that a struct cann't be parser from string and table in the sametime.

linyihai:
> I assume this is meant to resolve the immediate part of #13688 (leaving aside the more general problem). Is that right?

Yes，this tend to mirror the -Zgit/-Zgitoxide as possible.

src/cargo/util/context/de.rs
```diff
@@ -62,7 +62,12 @@ impl<'de, 'gctx> de::Deserializer<'de> for Deserializer<'gctx> {
             let (res, def) = res;
             return res.map_err(|e| e.with_key_context(&self.key, def));
         }
-        Err(ConfigError::missing(&self.key))
+
+        // Let's assume that `unstable.git` had defined its `deserialize_any`` method,
+        // then run `CARGO_UNSTABLE_GIT_SHALLOW_INDEX cargo fetch`, here will return `missing config key unstable.git`.
+        // It seems that anything that starts with CARGO_UNSTABLE_GIT, even like CARGO_UNSTABLE_GITOXIDE can trigger this.
+        // This is a workaround for now, but should be fixed in the future.
+        visitor.visit_none()
```
weihanglo, line 70: While this seems fine for now, I am a bit worried about the change.

What about having an enum deriving serde?

```
#[derive(Debug, Copy, Clone, Default, Deserialize)]
#[serde(untagged)]
pub enum Gitoxide {
    #[default]
    All,
    Some(GitoxideFeatures),
}
```

so that I assumed it would work well with

```toml
[unstable]
gitoxide = true

# or
[unstable.gitoxide]
checkout = true
```

As well as enviroment variables?

```
CARGO_UNSTABLE_GITOXIDE=true
CARGO_UNSTABLE_GITOXIDE_CHECKOUT=true
```

weihanglo (commented):
Sorry for the late review 🙇🏾‍♂️

src/cargo/util/context/de.rs
linyihai: It doesn't seem to be working.  see `gitoxide_features_as_table` in the lastest commit.

linyihai: I had repaced `vistor.visit_none()` with below,
```
        // The effect here is the same as in [`deserialize_option`].
        if self.gctx.has_key(&self.key, self.env_prefix_ok)? {
            return visitor.visit_some(self);
        }
```

So we can still specify a inner field of a struct in the ENV, like `CARGO_UNSTABLE_GITOXIDE_FETCH`, even that `CARGO_UNSTABLE_GIOXIDE` doesn't exist.


But that's not enough, It will still warning `unstable.git` missing key,  Why is this happening?  After some groping,  Problems gradually surfaced, see https://github.com/rust-lang/cargo/blob/master/src/cargo/util/context/de.rs#L268-L269
```
                if env_key.starts_with(field_key.as_env_key()) {
                    fields.insert(KeyKind::Normal(field.to_string()));
```
here will treat `CARGO_UNSTABLE_GIT` as prefix of `CARGO_UNSTABLE_GIOXIDE_FETCH`,  and the `unstable.git` and `unsable.gitoxide` will be pushed into fields. This is a problem,  when the unstable fields is a struct and the fields implement their `deserialize_any` method.  So any environment variable that starts with the XX character will have missing XX keys.

src/cargo/util/context/de.rs
```diff
@@ -265,8 +271,14 @@ impl<'gctx> ConfigMapAccess<'gctx> {
             let mut field_key = de.key.clone();
             field_key.push(field);
             for env_key in de.gctx.env_keys() {
-                if env_key.starts_with(field_key.as_env_key()) {
```
linyihai, line 268: See https://github.com/rust-lang/cargo/pull/13687#discussion_r1607775037

I wonder if this is worth fixing on separately PR.  Because it's only a problem in my situation

weihanglo: Probably. I'll find a time reviewing this.
(it's always a headache for me to read config deserialization code 🤯)

weihanglo: Just reviewed it again. Yeah I believe this can pull out into a separate PR.

Also, we improve it a bit to avoid extra allocation and string comparsion:

```
            for env_key in de.gctx.env_keys() {
                let Some(nested_field) = env_key.strip_prefix(field_key.as_env_key()) else {
                    continue;
                };
                // This distinguishes fields that share the same prefix.
                // For example, when env_key is UNSTABLE_GITOXIDE_FETCH
                // and field_key is UNSTABLE_GIT, the field shouldn't be 
                // added because `unstable.gitoxide.fetch` doesn't
                // belong to `unstable.git` struct.
                if nested_field.is_empty() || nested_field.starts_with('_') {
                    fields.insert(KeyKind::Normal(field.to_string()));
                }
            }
```

src/cargo/core/features.rs
```diff
@@ -875,12 +878,71 @@ pub struct GitFeatures {
...
-    fn all() -> Self {
+    pub fn all() -> Self {
         GitFeatures {
             shallow_index: true,
             shallow_deps: true,
         }
     }
+
+    fn expecting() -> String {
+        let fields = vec!["'all'", "'shallow-index'", "'shallow-deps'"];
```
weihanglo, line 889:
```suggestion
        let fields = ["'all'", "'shallow-index'", "'shallow-deps'"];
```

src/cargo/core/features.rs
```diff
@@ -892,19 +954,19 @@ fn parse_git(it: impl Iterator<Item = impl AsRef<str>>) -> CargoResult<Option<Gi
 
     for e in it {
         match e.as_ref() {
+            "all" => return Ok(Some(GitFeatures::all())),
```
weihanglo, line 957: I don't think we need to support string `"all"`. `-Zgit` or `--config unstable.git=true` should be sufficient?

src/cargo/core/features.rs
```diff
@@ -935,6 +997,67 @@ impl GitoxideFeatures {
             internal_use_git2: false,
         }
     }
+
+    fn expecting() -> String {
+        let fields = vec!["'all'", "'fetch'", "'checkout'", "'internal-use-git2'"];
```
weihanglo, line 1002: ditto. And `internal-use-git2` shouldn't be suggested I feel like.

```suggestion
        let fields = ["'fetch'", "'checkout'"];
```

src/cargo/core/features.rs
```diff
@@ -935,6 +997,67 @@ impl GitoxideFeatures {
...
+        format!(
+            "unstable 'gitoxide' only takes {} as valid inputs, your can use 'all' to turn out all gitoxide features, for shallow fetches see `shallow-index,shallow-deps`",
```
weihanglo, line 1004: could we use backtick "`" to quote options?

weihanglo (commented):
Looks reasonable. Just some minor suggestions.

src/cargo/core/features.rs
linyihai: Ok, it is a convention

src/cargo/core/features.rs
```diff
@@ -943,6 +1004,67 @@ impl GitoxideFeatures {
             internal_use_git2: false,
         }
     }
+
+    fn expecting() -> String {
+        let fields = vec!["`fetch`", "`checkout`", "`internal-use-git2`"];
+        format!(
+            "unstable 'gitoxide' only takes {} as valid inputs, for shallow fetches see `shallow-index,shallow-deps`",
```
weihanglo, line 1011:
```suggestion
            "unstable 'gitoxide' only takes {} as valid inputs, for shallow fetches see `-Zgit=shallow-index,shallow-deps`",
```

weihanglo (commented):
One minor error message regression left and we're ready to go

src/cargo/core/features.rs
linyihai: Shouldn't it be `-Zgitoxide`

linyihai: Looks like the old tip was wrong, too. See https://doc.rust-lang.org/cargo/reference/unstable.html, let me  correct it to `-Zgitoxide` also

linyihai: Sorry, I misunderstand

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #7768, opened 2020-01-06
Search for root manifest with ephemeral workspaces

chrisduerr:
Fixes #5495.

This seems like it's too simple to just work like this, but after trying a few different things, this was the only solution which worked reliably for me.

I've verified that no `/target` is present in the actual checkout location, the target directory used is actually the one created in `/tmp`.

I've also verified that both workspaces and "normal" packages still install through git and that a normal `cargo install --path` works too (though that doesn't use ephemeral workspaces anyways).

PR #5685, opened 2018-07-05
Fix `cargo install` using a workspace target dir

alexcrichton:
Closes #5662

Issue #5662, opened 2018-06-28
cargo install doesn't follow workspace target


xzn:
I'm following this book here: [https://doc.rust-lang.org/book/second-edition/ch14-03-cargo-workspaces.html](https://doc.rust-lang.org/book/second-edition/ch14-03-cargo-workspaces.html)

and `cargo build` works as described. However `cargo install` doesn't work in the workspace level, and at crate level it tries to recompile the binary to a `target` folder within the crate folder instead of using the workspace level `target` folder.

`cargo build` from a workspace level also sometimes compiles things differently from `cargo install` from the crate. I was trying to compile [pijul](https://pijul.com/releases/pijul-0.10.0.tar.gz) on Windows earlier and while workspace level `cargo build` completes successfully, `cargo install pijul` (as well as `cargo install` from pijul crate) results in a bunch missing trait implementation errors.

In comparison `cargo run` doesn't have this problem as it uses the executables from workspace level target folder. (Although it'd be nice if `cargo run` can be used from workspace root directly instead of at crate folder)

alexcrichton:
Hm can you describe a bit more precisely what you're executing with `cargo install`? The meaning of `cargo install` and `cargo install pijul` are very different which may be why you're seeing issues.

xzn:
Sorry could've been more clear.

Using [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen) as an example,

Executing `cargo build` at workspace root works as expected and binaries from `crates/cli` directory are compiled to workspace root target folder. Now let's say I want to install the binaries with `cargo install`, at workspace root level this doesn't work so I `cd` to `crates/cli` and run `cargo install` from there. The problem is that instead of using the the workspace root target folder, it creates and uses a target folder from `crates/cli` instead.

The expectation is that `cargo install` works like `cargo build` and `cargo run` etc and uses the target folder at workspace root level.

`cargo install pijul` is prolly a separate issue: basically running `cargo build` at workspace root of pijul source tree succeeds with binary being placed in workspace root target folder. Running `cargo install pijul` directly fails with some compilation error.

ehuss:
This is caused by the change in #5350.  @matklad, would it make sense to revert that?  It's not clear to me why it changed.

PR #5350, opened 2018-04-12
Simplify install

matklad:
Just a mechanical simplification.

PR #13689, opened 2024-04-02
fix: create ephemeral workspace for git source

### LINKED ISSUE #13259: `cargo install --git` not recompiling different revs when using shared target-dir
### Problem

when using `cargo install` to install a different commit from a git repo, the executable doesn't gets replaced to target version

### Steps

1. cargo install --git https://github.com/BillGoldenWater/Tools-rs --rev b6c330a video_in_waveform
2. get the hash of .cargo/bin/video_in_waveform
3. cargo install --git https://github.com/BillGoldenWater/Tools-rs --rev e25e95d video_in_waveform
4. compare the hash of .cargo/bin/video_in_waveform to previously, it's the same 

### Possible Solution(s)

_No response_

### Notes

maybe related to workspace

tried `--force`, `+nightly` and uninstall then install

after run `cargo clean`, it will install the proper version

the nightly version:
```
cargo 1.77.0-nightly (2ce45605d 2024-01-04)
release: 1.77.0-nightly
commit-hash: 2ce45605d9db521b5fd6c1211ce8de6055fdb24e
commit-date: 2024-01-04
host: aarch64-apple-darwin
libcurl: 8.4.0 (sys:0.4.70+curl-8.5.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 14.2.1 [64-bit]
```

### Version

```text
release: 1.75.0
commit-date: 2023-11-20
host: aarch64-apple-darwin
libcurl: 8.4.0 (sys:0.4.68+curl-8.4.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 14.2.1 [64-bit]
```

0xPoe:

close https://github.com/rust-lang/cargo/issues/13259

#### Root cause
The root cause of this problem is that cargo install creates a non-ephemeral workspace for git sources, that would try finding all member packages in that git repo, and add them as path dependencies.

However, when computing fingerprints to determine if a rebuild is needed, Cargo doesn't hash git revision into fingerprint directly. Instead, it checks if source file path has changed. For git source, it is an absolute path to ~/.cargo/git/checkouts/<pkg-hash>/<short-commit-sha>, so commit SHA is always there. For path dependency, it is relative to workspace root, so nothing change and Cargo doesn't rebuild even it is from a different git revision.

#### Sloution
If any time we create an ephemeral workspace, then we will update the git source every time and that would help us use the right code and rebuild the binary correctly.

There are some historical PRs related to this change:
1. 4 years ago, we tried to all use ephemeral workspace in https://github.com/rust-lang/cargo/pull/5350. But it caused https://github.com/rust-lang/cargo/issues/5662 so we revert that change in https://github.com/rust-lang/cargo/pull/5685.
2. https://github.com/rust-lang/cargo/pull/7768 In this PR, we try to search the root manifest by using `Workspace::new`, but it also caused the https://github.com/rust-lang/cargo/issues/13259.

So in my fix, I just rollback to use ephemera workspace but using `find_root` to achieve the same goal of searching the root manifest. You can find more in the test: `git_install_reads_workspace_manifest`

- [x] manually test https://github.com/rust-lang/cargo/pull/13689#issuecomment-2053473575
- [x] unit test

### Additional information

None

### DIFF
--- a/src/cargo/core/workspace.rs
+++ b/src/cargo/core/workspace.rs
@@ -283,6 +283,10 @@ impl<'gctx> Workspace<'gctx> {
         ws.member_ids.insert(id);
         ws.default_members.push(ws.current_manifest.clone());
         ws.set_resolve_behavior();
+        // The find_root function is used here to traverse the directory tree and locate the root of the workspace.
+        // Despite being ephemeral, we still need to validate all the manifests in the workspace,
+        // which is what `find_root` helps us achieve here.
+        ws.find_root(ws.current_manifest.clone().as_path())?;
         Ok(ws)
     }
 
--- a/tests/testsuite/install.rs
+++ b/tests/testsuite/install.rs
@@ -2065,6 +2065,75 @@ fn git_install_reads_workspace_manifest() {
 }
 
 #[cargo_test]
+fn git_install_the_same_bin_twice_with_different_rev() {
+    let project = git::new("foo", |project| {
+        project
+            .file(
+                "Cargo.toml",
+                r#"
+            [workspace]
+            members = ["bin1"]
+            "#,
+            )
+            .file("bin1/Cargo.toml", &basic_manifest("bin1", "0.1.0"))
+            .file(
+                "bin1/src/main.rs",
+                r#"fn main() { println!("Hello, world!"); }"#,
+            )
+    });
+    let repository = git2::Repository::open(&project.root()).unwrap();
+    let first_rev = repository.revparse_single("HEAD").unwrap().id().to_string();
+
+    // Change the main.rs.
+    fs::write(
+        project.root().join("bin1/src/main.rs"),
+        r#"fn main() { println!("Hello, world! 2"); }"#,
+    )
+    .expect("failed to write file");
+    git::commit(&repository);
+    let second_rev = repository.revparse_single("HEAD").unwrap().id().to_string();
+
+    // Set up a temp target directory.
+    let temp_dir = paths::root().join("temp-target");
+    cargo_process(&format!(
+        "install --git {} --rev {} bin1 --target-dir {}",
+        project.url().to_string(),
+        second_rev,
+        temp_dir.display()
+    ))
+    .with_stderr(
+        "\
+[UPDATING] git repository [..]
+[INSTALLING] bin1 v0.1.0 [..]
+[COMPILING] bin1 v0.1.0 [..]
+[FINISHED] [..]
+[INSTALLING] [..]home/.cargo/bin/bin1[..]
+[INSTALLED] package `bin1 [..]
+[WARNING] be sure to add [..]
+",
+    )
+    .run();
+
+    cargo_process(&format!(
+        "install --git {} --rev {} bin1 --target-dir {}",
+        project.url().to_string(),
+        first_rev,
+        temp_dir.display()
+    ))
+    .with_stderr(
+        "\
+[UPDATING] git repository [..]
+[INSTALLING] bin1 v0.1.0 [..]
+[FINISHED] [..]
+[REPLACING] [..]home/.cargo/bin/bin1[..]
+[REPLACED] package `bin1 [..]
+[WARNING] be sure to add [..]
+",
+    )
+    .run();
+}
+
+#[cargo_test]
 fn install_git_with_symlink_home() {
     // Ensure that `cargo install` with a git repo is OK when CARGO_HOME is a
     // symlink, and uses an build script.
--- a/tests/testsuite/install.rs
+++ b/tests/testsuite/install.rs
@@ -2124,6 +2124,7 @@ fn git_install_the_same_bin_twice_with_different_rev() {
         "\
 [UPDATING] git repository [..]
 [INSTALLING] bin1 v0.1.0 [..]
+[COMPILING] bin1 v0.1.0 [..]
 [FINISHED] [..]
 [REPLACING] [..]home/.cargo/bin/bin1[..]
 [REPLACED] package `bin1 [..]
--- a/src/cargo/ops/cargo_install.rs
+++ b/src/cargo/ops/cargo_install.rs
@@ -816,7 +816,7 @@ fn make_ws_rustc_target<'gctx>(
     source_id: &SourceId,
     pkg: Package,
 ) -> CargoResult<(Workspace<'gctx>, Rustc, String)> {
-    let mut ws = if source_id.is_git() || source_id.is_path() {
+    let mut ws = if source_id.is_path() {
         Workspace::new(pkg.manifest_path(), gctx)?
     } else {
         Workspace::ephemeral(pkg, gctx, None, false)?


hi-rustin:
Debugging note:
1. The failed case was caused by my change.
2. The reason is that we don't check the package content in `Workspace::ephemera` function, but in `Workspace::new` we will try to load the package again. This way, it can bail out the correct error from the manifest.
3. Perhaps we should explicitly check the manifest in the ephemera function or find a way to abort the process before `make_ws_rustc_target`.

hi-rustin:
Tested locally:
1. Add a new config
```toml
[build] 
target-dir = "/Volumes/t7/code/cargo/target/tmp1" 
```
2. Install the video_in_waveform with new cargo build:
```console
./target/debug/cargo install --git https://github.com/hi-rustin/cargo-information --rev 5fd64bb6afe37092ee1dd7c518f4fc4865175c22
   Compiling bytesize v1.3.0
   Compiling cargo-information v0.4.2 (https://github.com/hi-rustin/cargo-information?rev=5fd64bb6afe37092ee1dd7c518f4fc4865175c22#5fd64bb6)
    Finished `release` profile [optimized] target(s) in 1m 26s
  Installing /Users/hi-rustin/.cargo/bin/cargo-info
   Installed package `cargo-information v0.4.2 (https://github.com/hi-rustin/cargo-information?rev=5fd64bb6afe37092ee1dd7c518f4fc4865175c22#5fd64bb6)` (executable `cargo-info`)
```
3. Calculate the hash:
```console
md5 ~/.cargo/bin/cargo-info
MD5 (/Users/hi-rustin/.cargo/bin/cargo-info) = 3a7ccae20527098730fda436790ddf72
```
4. Install it again with a different version:
```console
./target/debug/cargo install --git https://github.com/hi-rustin/cargo-information --rev e8b3419bc033d3252ad19081f516946c94e0c02e
    Updating git repository `https://github.com/hi-rustin/cargo-information`
    Updating crates.io index
   Compiling cargo-information v0.4.2 (https://github.com/hi-rustin/cargo-information?rev=e8b3419bc033d3252ad19081f516946c94e0c02e#e8b3419b)
    Finished `release` profile [optimized] target(s) in 7.85s
   Replacing /Users/hi-rustin/.cargo/bin/cargo-info
    Replaced package `cargo-information v0.4.2 (https://github.com/hi-rustin/cargo-information?rev=5fd64bb6afe37092ee1dd7c518f4fc4865175c22#5fd64bb6)` with `cargo-information v0.4.2 (https://github.com/hi-rustin/cargo-information?rev=e8b3419bc033d3252ad19081f516946c94e0c02e#e8b3419b)` (executable `cargo-info`)
```
5. Calculate the hash:
```console
md5 ~/.cargo/bin/cargo-info
MD5 (/Users/hi-rustin/.cargo/bin/cargo-info) = abd82eed7d6f3f88cf7a5371bf75d469
```

hi-rustin:
@weihanglo 
Could you please take a look? Thanks! 💚 💙 💜 💛 ❤️

src/cargo/core/workspace.rs
```diff
@@ -283,6 +283,10 @@ impl<'gctx> Workspace<'gctx> {
         ws.member_ids.insert(id);
         ws.default_members.push(ws.current_manifest.clone());
         ws.set_resolve_behavior();
+        // The find_root function is used here to traverse the directory tree and locate the root of the workspace.
```
weihanglo, line 286: A bit worried about this change. The in-memory property isn't held for ephemeral workspaces after this patch. It is also a bit fragile because the workspace discovery logic is only guarded by [the other place](
https://github.com/rust-lang/cargo/blob/0a5cb8977a4393e9e814d9c049bb3e441c18f72b/src/cargo/core/workspace.rs#L1836-L1848), which is not immediately clear and hard to track.

I wonder if we really need this line here. `git_install_reads_workspace_manifest` seems to fail if we don't do `find_root`, but does cargo ever use `profile` or anything from there? Things we might want to figure out:

* What kind of workspace support `cargo install --git` current provide?
* Can we skip this and still find all members and inheritable fields?

tests/testsuite/install.rs
```diff
@@ -2064,6 +2064,76 @@ fn git_install_reads_workspace_manifest() {
...
+[COMPILING] bin1 v0.1.0 [..]
+[FINISHED] [..]
+[INSTALLING] [..]home/.cargo/bin/bin1[..]
+[INSTALLED] package `bin1 [..]
+[WARNING] be sure to add [..]
+",
+    )
+    .run();
+
+    cargo_process(&format!(
```
weihanglo, line 2117: Should have a comment or two pointing out we want to observe a rebuild? 
Or we can actually do a shasum comparsion like the original did?

weihanglo (commented):
Thanks for looking back into the history!

src/cargo/core/workspace.rs
hi-rustin: > but does cargo ever use `profile` or anything from there?

After this PR we can only check the manifest for errors. Before that, I belive that we could use any information defined by the workspace. For example lints etc.



> * What kind of workspace support `cargo install --git` current provide?

Before this PR, we just load the whole workspace and create a workspace based on the root manifest. 
After this PR, we create the workspace based on the specific package and we don't use the information from the root manifest. Only searched the root to see if there is an error in the manifest.




> * Can we skip this and still find all members and inheritable fields?

After this PR, we only checked the manifest, we don't respect workspace information anymore. 

For example:
If we try `cargo install --git https://github.com/hi-rustin/test-cargo-git-install bin-tool;`

Before this change:
```console
test-cargo-git-install on  master via 🦀 v1.77.2 
❯ cargo install --git https://github.com/hi-rustin/test-cargo-git-install bin-tool 
    Updating git repository `https://github.com/hi-rustin/test-cargo-git-install`
warning: virtual workspace defaulting to `resolver = "1"` despite one or more workspace members being on edition 2021 which implies `resolver = "2"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
note: for more details see https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions
  Installing bin-tool v0.1.0 (https://github.com/hi-rustin/test-cargo-git-install#d7a594e5)
   Compiling lib1 v0.1.0 (/Users/hi-rustin/.cargo/git/checkouts/test-cargo-git-install-fae8728302b9b2da/d7a594e/lib1)
   Compiling bin-tool v0.1.0 (/Users/hi-rustin/.cargo/git/checkouts/test-cargo-git-install-fae8728302b9b2da/d7a594e/bin-tool)
error: usage of an `unsafe` block
 --> bin-tool/src/main.rs:2:5
  |
2 | /     unsafe {
3 | |         println!("Hello, world!");
4 | |     }
  | |_____^
  |
  = note: requested on the command line with `-F unsafe-code`

warning: unnecessary `unsafe` block
 --> bin-tool/src/main.rs:2:5
  |
2 |     unsafe {
  |     ^^^^^^ unnecessary `unsafe` block
  |
  = note: `#[warn(unused_unsafe)]` on by default

warning: `bin-tool` (bin "bin-tool") generated 1 warning
error: could not compile `bin-tool` (bin "bin-tool") due to 1 previous error; 1 warning emitted
error: failed to compile `bin-tool v0.1.0 (https://github.com/hi-rustin/test-cargo-git-install#d7a594e5)`, intermediate artifacts can be found at `/var/folders/j1/7l6snwpx6svgqxh79bz_d27m0000gn/T/cargo-installbO7SHc`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```

After this change:
```console
test-cargo-git-install on  master via 🦀 v1.77.2 
❯ ../cargo/target/debug/cargo install --git  https://github.com/hi-rustin/test-cargo-git-install bin-tool

    Updating git repository `https://github.com/hi-rustin/test-cargo-git-install`
  Installing bin-tool v0.1.0 (https://github.com/hi-rustin/test-cargo-git-install#d7a594e5)
   Compiling lib1 v0.1.0 (https://github.com/hi-rustin/test-cargo-git-install#d7a594e5)
   Compiling bin-tool v0.1.0 (https://github.com/hi-rustin/test-cargo-git-install#d7a594e5)
    Finished `release` profile [optimized] target(s) in 2.57s
  Installing /Users/hi-rustin/.cargo/bin/bin-tool
   Installed package `bin-tool v0.1.0 (https://github.com/hi-rustin/test-cargo-git-install#d7a594e5)` (executable `bin-tool`)
```

hi-rustin: From a workspace information perspective, I think this PR introduces some regression.

soloturn: > From a workspace information perspective, I think this PR introduces some regression.

workspace is a feature which should not be there as it complicates all optimizations? it is only there because CARGO_TARGET_DIR does not work very well, isn't it?

weihanglo: workspace is already there so to move forward we need to first figure out the feature set `cargo install --git` offers. However, I agree with soloturn, we may need to step back a bit to re-evaludate the bug. 

Both workspace initialization and `CARGO_TARGET_DIR` are symptoms of the bug. From my previous investigation, the bug is in workspace, members are collected as from `PathSource`, hence `cargo install` recognizes the as path dependencies. In contrast, when using in `git` dependency, workspace members will be marked as from `GitSource`. I wonder if we should mark the entire workspace with the same source as the root one for `cargo install`. Thinking of nested packages <https://github.com/rust-lang/rfcs/pull/3452>, if we had that today, we might also need this, otherwise nested package will be marked as from local path and never get updated.

hi-rustin: > workspace is a feature which should not be there as it complicates all optimizations? it is only there because CARGO_TARGET_DIR does not work very well, isn't it?

I think it's hard to say the workspace feature complicates all optimizations. As my example in https://github.com/hi-rustin/test-cargo-git-install, respect workspace information will help us avoid inconsistency.

The problem is not caused by the CARGO_TARGET_DIR. If users set a CARGO_TARGET_DIR, then all compilation would happen in the CARGO_TARGET_DIR. 

The problem here is as @weihanglo said, we cannot rebuild the binary after the git repository is updated.

hi-rustin: > I wonder if we should mark the entire workspace with the same source as the root one for `cargo install`. Thinking of nested packages [rust-lang/rfcs#3452](https://github.com/rust-lang/rfcs/pull/3452), if we had that today, we might also need this, otherwise nested package will be marked as from local path and never get updated.

That sounds like a reasonable solution, I will take a look.

hi-rustin: > I wonder if we should mark the entire workspace with the same source as the root one for `cargo install`

I've looked at how we calculate the fingerprint of the path member. 

```rust
pub fn path_args(ws: &Workspace<'_>, unit: &Unit) -> (PathBuf, PathBuf) {
    let ws_root = ws.root();
    let src = match unit. target.src_path() {
        TargetSourcePath::Path(path) => path.to_path_buf(),
        TargetSourcePath::Metabuild => unit.pkg.manifest().metabuild_path(ws.target_dir()),
    };
    assert!(src.is_absolute());
    if unit.pkg.package_id().source_id().is_path() {
        if let Ok(path) = src.strip_prefix(ws_root) {
            return (path.to_path_buf(), ws_root.to_path_buf());
        }
    }
    (src, unit.pkg.root().to_path_buf())
}
```

It seems we always try to track the workspace as much as possible.

So I am not sure what you mean by `mark the entire workspace with the same source`. Can you explain it a little bit?

weihanglo: When we read a dependency from a Git source, no matter it is a single package or workspace member, it always returns a `PackageId` with `SourceKind::Git`. Under the hood it is actually a `PathSource` assoicated with a Git `SourceId`. This helps Cargo track the actual source. 

https://github.com/rust-lang/cargo/blob/b9d913e532df8bb6e14a507376d9fbbcacdcc0b0/src/cargo/sources/git/source.rs#L360-L365

However, in `cargo install` we use workspace so we lost that  sticky fashion.

PR #13693, opened 2024-04-02
refactor(toml): Split out an explicit step to resolve `Cargo.toml`

epage:

This builds on #13664 and #13666.   Currently, once we have deserialized `Cargo.toml`, we pass it to a large machinery (`to_real_manifest`, `to_virtual_manifest`) so that
- `Cargo.toml` is resolved
- `Summary` is created
- `Manifest` is created

This splits out the resolving of `Cargo.toml` which is mostly workspace inheritance today.

While splitting logic conjoined like this can be a bit messy in the short term, the hope is that overall this makes the logic easier to follow (more condensed, focused sections to view; more explicit inputs/outputs).

In particular, I hope that this will make it clearer and easier to shift more logic into the resolving step, specifically the inferring of build targets for #13456.

This is broken up into very small steps in the hope that it makes it easier to analyze a step.


Muscraft (approved):
Thanks!

This makes stuff much easier to understand

PR #13706, opened 2024-04-04
Don't depend on `?` affecting type inference in weird ways

WaffleLapkin:
This is likely to stop working in the future versions of Rust, see https://github.com/rust-lang/rust/pull/122412#issuecomment-2037813927.

### DIFF
--- a/crates/cargo-util/src/paths.rs
+++ b/crates/cargo-util/src/paths.rs
@@ -510,8 +510,7 @@ fn _remove_file(p: &Path) -> Result<()> {
         }
     }
 
-    Err(err).with_context(|| format!("failed to remove file `{}`", p.display()))?;
-    Ok(())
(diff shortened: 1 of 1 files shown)


epage:
Merging this more as a code cleanup. For this breaking in the future, I'm assuming that is non-urgent as I'm assuming a lot needs to be worked out to uphold stability expectations before that will matter.

@bors r+

weihanglo (commented):
Does this block any crate run?

WaffleLapkin:
@epage yes, this is non urgent (although this needs to happen before the change, so that rustc's CI can pass), so I just made a PR since this is a simple change & is indeed a little cleanup.

@weihanglo this is required for the crater run on #122412, although I think I can workaround that.

weihanglo:
Thanks. I could do a cargo submodule update earlier if needed. Usually it happens on Tuesday and Friday.

Issue #6313, opened 2018-11-13
Duplicate artifact tracking issue.


ehuss:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/issues/6313)*

#6308 added a check if multiple jobs produce the same files. To be safe, it is currently a warning. At some point in the future it should be turned into a hard error if no problems are reported. Collisions are almost certainly bad behavior that will cause problems, so rejecting it is probably the right thing to do.

Related issues:
- #5524: Request to warn on duplicate artifact creation — wherein `.rmeta` files started to collide because bin targets started producing them, and were corrupting one another if built simultaneously. `.rmeta` collisions fixed in #6292.
- #6293: Handling multiple targets in a workspace with the same name.
- #5444: Hardlink collisions when target is built multiple times (like panic/no-panic). Fixed by #5460 by only hardlinking what was requested.

Known situations where collisions may occur:
- Multiple binary/example/lib targets in a workspace with the same name. (Or reusing a shared target directory.)
- Using `--out-dir` with an example and binary with the same name.
- Multiple path dependencies with dylibs with the same name.
- Multiple dependencies with the same name and you select both of them on the command-line, for example: `cargo build -p rand:0.4.3 -p rand:0.5.5`.
- rustdoc in a workspace where multiple crates have the same name.  This can arise from a variety of situations (renamed dependencies, multiple versions of a package, different packages with the same crate names, etc.). (rust-lang/rust#56169, rust-lang/rust#61378)
- `panic="abort"` and cdylibs and tests:  Create a project with a lib (cdylib crate type), binary, and an integration test, with `panic="abort"` in the profile. When `cargo test` runs, the cdylib is built twice (once with panic=abort for the binary, and once without for the test), with the same filename.  Building the lib for the test should probably skip the `cdylib` crate type (assuming rlib is also available), but implementing this is very difficult. See https://github.com/rust-lang/cargo/issues/6313#issuecomment-480014371.
- Multiple targets (particularly executables) that differ only by case on case-insensitive filesystems.
- PDB collisions on Windows. If the package has a binary and a dylib library of the same name, then the `.pdb` file for each target will have the same name. This is easy to hit if you have a proc-macro package with a binary.
- Dylib built multiple times with different features
    - for example with new feature resolver, see #9278.
    - and with workspace members, see #12345
[...]

Issue #7754, opened 2019-12-29
cargo-metadata always resolves features at the workspace level


sfackler:
**Problem**
`cargo-metadata` appears to always resolve features at the "workspace" level, rather than for root crate. This means that `cargo metadata` will report features as enabled that aren't actually when `cargo build` is run in the same directory.

**Steps**

1. Clone https://github.com/sfackler/rust-postgres
2. Run `cargo metadata --format-version 1 --manifest-path postgres-derive/Cargo.toml | jq '.resolve.nodes|.[]|select(.id|test("postgres-types"))'`.
3. Note that the entry for `postgres-types` has the `derive` and `postgres-derive` features enabled even though they are off by default for that crate. The `postgres-derive-test` crate in the same workspace enables that feature, which I assume is where that's coming from:

```json
{
  "id": "postgres-types 0.1.0 (path+file:///home/sfackler/code/rust-postgres/postgres-types)",
  "dependencies": [
    "bytes 0.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
    "fallible-iterator 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
    "postgres-derive 0.4.0 (path+file:///home/sfackler/code/rust-postgres/postgres-derive)",
    "postgres-protocol 0.5.0 (path+file:///home/sfackler/code/rust-postgres/postgres-protocol)"
  ],
  "deps": [
    {
      "name": "bytes",
      "pkg": "bytes 0.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
      "dep_kinds": [
        {
          "kind": null,
          "target": null
        },
        {
          "kind": null,
          "target": null
        }
      ]
    },
    {
      "name": "fallible_iterator",
      "pkg": "fallible-iterator 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
      "dep_kinds": [
        {
          "kind": null,
          "target": null
        },
        {
          "kind": null,
          "target": null
        }
      ]
    },
    {
      "name": "postgres_derive",
      "pkg": "postgres-derive 0.4.0 (path+file:///home/sfackler/code/rust-postgres/postgres-derive)",
      "dep_kinds": [
        {
          "kind": null,
          "target": null
        }
      ]
    },
    {
      "name": "postgres_protocol",
      "pkg": "postgres-protocol 0.5.0 (path+file:///home/sfackler/code/rust-postgres/postgres-protocol)",
      "dep_kinds": [
        {
[...]

PR #13709, opened 2024-04-05
feat: implement RFC 3553 to add SBOM support

justahero:

This PR is an implementation of [RFC 3553] to add support to generate pre-cursor SBOM files for compiled artifacts in Cargo.

The [RFC 3553] adds a new option to Cargo to emit SBOM pre-cursor files. A project can be configured either by the new Cargo config field `sbom`.

```toml
# .cargo/config.toml
[build]
sbom = true
```

or using the environment variable `CARGO_BUILD_SBOM=true`. The `sbom` option is an unstable feature and requires the `-Zsbom` flag to enable it.

Check out this branch & compile Cargo. Pick a Cargo project to test it on, then run:

```
CARGO_BUILD_SBOM=true <path/to/compiled/cargo>/target/debug/cargo build -Zsbom
```

All generated `*.cargo-sbom.json` files are located in the `target` folder alongside their artifacts. To list all generated files use:

```
find ./target -name "*.cargo-sbom.json"
```

then check their content. To see the current output format, see [these examples](https://gist.github.com/justahero/2683f5520bf5e921c6b839a0e91cd01c).

### What does the PR not solve?

The PR leaves a task(s) open that are either out of scope or should be done in a follow-up PRs.

* does not address #6313 (see [comment](https://github.com/rust-lang/cargo/pull/13709#discussion_r1588314920))

### Additional information

There are a few things that I would like to get feedback on, in particular the generated JSON format is not final. Currently it holds the information listed in the [RFC 3553], but it could be further enriched with information only available during builds.

During the implementation a number of questions arose:

- [ ] Should the graph be packages or crates?
  - The unit graph that the SBOM is based on is units. The current SBOM graph is identical to the unit graph, with the run build script nodes merged with building build scripts.
  - Artifact dependencies may impact this
- [ ] Which outputs should get SBOMs files?
  - Currently: executables (including examples and tests), dylib, cdylib, staticlib
- [ ] How do we refer to "normal" dependencies?  https://github.com/rust-lang/cargo/pull/13709#discussion_r1873994078
- [ ] What case should we use?  https://github.com/rust-lang/cargo/pull/13709#discussion_r1873996092
- [ ] Should this be `build.sbom` or `profile.*.sbom`
- [ ] Is sbom the right name for this?

Thanks @arlosi, @RobJellinghaus and @lfrancke for initial guidance & feedback.

* RFC: #rust-lang/rfcs/pull/3553

[RFC 3553]: https://github.com/rust-lang/rfcs/pull/3553


src/cargo/core/compiler/build_runner/mod.rs
```diff
@@ -422,6 +426,21 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
         self.files().metadata(unit)
     }
 
+    /// Returns the list of SBOM output file paths for a given Unit.
+    ///
+    /// Only call this function when `sbom` is active.
+    pub fn sbom_output_files(&self, unit: &Unit) -> CargoResult<Vec<PathBuf>> {
```
justahero, line 432: I'm uncertain if this function should return a `Vec` or if a single `PathBuf` is actually sufficient.

heisen-li:
Much respect for your contribution. 

From my kind reminders, it seems appropriate to modify the documentation of the corresponding sections, e.g. [Configuration](https://doc.rust-lang.org/cargo/reference/config.html#configuration), [Environment Variables](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables).

weihanglo:
Thanks for the reminder, @heisen-li. Would love to see a doc update, though we should probably focus on the design discussion first, as the location of the configuration is not yet decided. (See <https://github.com/rust-lang/rfcs/pull/3553#discussion_r1442335016>).

epage:
One approach for the docs (if this is looking to be merged) is to put the env and config documentation fragments in the Unstable docs.

src/cargo/core/compiler/mod.rs
```diff
@@ -684,12 +687,23 @@ fn prepare_rustc(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> CargoResult
...
     base.args(build_runner.bcx.rustflags_args(unit));
-    if build_runner.bcx.gctx.cli_unstable().binary_dep_depinfo {
+    if gctx.cli_unstable().binary_dep_depinfo {
         base.arg("-Z").arg("binary-dep-depinfo");
     }
 
     if is_primary {
         base.env("CARGO_PRIMARY_PACKAGE", "1");
+
+        if gctx.cli_unstable().sbom && build_runner.bcx.build_config.sbom {
```
weihanglo, line 697: IMO, paths are better to be joined using semi-colons, maybe via [`std::env::join_paths`](https://doc.rust-lang.org/stable/std/env/fn.join_paths.html), to maximize the compatibility.

src/cargo/core/compiler/build_runner/mod.rs
weihanglo: It might output several different crate-types, e.g. `rlib`, `cdylib`.

src/cargo/core/compiler/build_runner/mod.rs
```diff
@@ -422,6 +426,21 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
...
+        assert!(self.bcx.build_config.sbom);
+        let files = self
+            .outputs(unit)?
+            .iter()
+            .filter(|o| matches!(o.flavor, FileFlavor::Normal | FileFlavor::Linkable))
+            .filter_map(|output_file| output_file.hardlink.as_ref())
+            .map(|link_dst| link_dst.with_extension("cargo-sbom.json"))
```
weihanglo, line 439: Do we want to handle name collisions here? For example dylibs might collide with each their as their name have no `-<hash>` suffix.

See <https://github.com/rust-lang/cargo/issues/6313> for some details.

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,241 @@
+//! cargo-sbom precursor files for external tools to create SBOM files from.
+//! See [`output_sbom`] for more.
+
+use std::{
```
weihanglo, line 4: nit: in Cargo we don't do nested multi-line import, though this is not enforced :)

tests/testsuite/sbom.rs
```diff
@@ -0,0 +1,164 @@
...
+        .build();
+
+    p.cargo("build -Zsbom")
+        .masquerade_as_nightly_cargo(&["sbom"])
+        .run();
+
+    assert!(p.bin("foo").with_extension("cargo-sbom.json").is_file());
+    assert_eq!(
+        1,
+        p.glob(p.target_debug_dir().join("libfoo.cargo-sbom.json"))
```
weihanglo, line 87: I guess we might need to deal with different naming convention on different platform. (Windows specifically?)

tests/testsuite/sbom.rs
```diff
@@ -0,0 +1,164 @@
...
+        .file(
+            "bar/Cargo.toml",
+            r#"
+                [package]
+                name = "bar"
+                version = "0.1.0"
+                build = "build.rs"
+
+                [build-dependencies]
+                cc = "1.0.46"
```
weihanglo, line 149: Should avoid introducing external crates.io dependency. Use something like

```
    Package::new("baz", "0.0.1").publish();
```

tests/testsuite/sbom.rs
weihanglo, line 1: I would like to see some UI test for the JSON output. It's easier to check what's inside the SBOM file.

https://doc.crates.io/contrib/tests/writing.html#ui-tests

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,241 @@
...
+    let bcx = build_runner.bcx;
+    let rustc: SbomRustc = bcx.rustc().into();
+
+    let packages = fetch_packages(build_runner, unit);
+
+    for sbom_output_file in build_runner.sbom_output_files(unit)? {
+        let sbom = Sbom::new(unit, packages.clone(), rustc.clone());
+
+        let mut outfile = BufWriter::new(paths::create(sbom_output_file)?);
+        let output = serde_json::to_string_pretty(&sbom)?;
```
weihanglo, line 202: Perhaps pretty not important at this moment. Just wonder how big the size of the output sbom would be, and at some point we might want to use [`serde_json::to_writer_pretty`](https://docs.rs/serde_json/latest/serde_json/fn.to_writer_pretty.html) instead.

weihanglo (commented):
Just note that I reviewed this as-is, didn't really think too much for the design itself. Thank you for working on this!

src/cargo/core/compiler/build_config.rs
```diff
@@ -102,6 +104,16 @@ impl BuildConfig {
             anyhow::bail!("-Zbuild-std requires --target");
         }
 
+        // If sbom flag is set, it requires the unstable feature
+        let sbom = match gctx.get_env_os("CARGO_BUILD_SBOM") {
```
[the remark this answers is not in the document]
justahero: That results in a compile error:

```shell
error[E0624]: method `get_bool` is private
   --> src/cargo/core/compiler/build_config.rs:114:31
    |
114 |         let sbom = match gctx.get_bool("build.sbom")? {
    |                               ^^^^^^^^ private method
    |
   ::: src/cargo/util/context/mod.rs:120:9
    |
120 |         fn $name(&self, key: &ConfigKey) -> Result<OptValue<$ty>, ConfigError> {
    |         ---------------------------------------------------------------------- private method defined here
```

I also couldn't find any reference to this method or one of its siblings (e.g. `get_integer`).

src/cargo/core/compiler/build_runner/mod.rs
justahero: Tbh, I'm not sure how to solve this.

tests/testsuite/sbom.rs
justahero: Maybe the `glob` call can be simplified in a way that it works for all platforms.

src/cargo/core/compiler/build_runner/mod.rs
Shnatsel: This is already a problem for debug info, e.g. .pdb files on Windows. They have to be attributable to the file they correspond to, so they have the same name and cause a name collision. Cargo currently prints a warning about it, but otherwise just lets the collision happen and one of the debug info files is lost. However, losing SBOM info is highly undesirable.

We are not constrained by having the exact same name for the SBOM (unlike for debug info where we don't control the way it is discoverd), so we could do something about this.

I think we should emit all SBOM files in case of a collision, with a predictable naming scheme such as adding a number to the extension - `foo.sbom`, then `foo.1.sbom` and so on. The SBOM file will then contain the name of the binary it corresponds to, so that it can be disambiguated.

src/cargo/core/compiler/output_sbom.rs
Shnatsel: As a rule of thumb, about the same as `cargo metadata`.

Since this is meant to be consumed by machines (and humans can easily pipe it to `jq` or open the file in a browser to inspect it), I think we should stick to minified output because this could be emitted and parsed on every release build, including incremental ones.

weihanglo: > Since this is meant to be consumed by machines

Agree in that regard

src/cargo/core/compiler/build_config.rs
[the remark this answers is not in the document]
weihanglo: That was my misread. I shouldn't have reviewed this while meeting with you people 😆.

Here is my new suggestion:

```rust
        let sbom = match (cfg.sbom, gctx.cli_unstable().sbom) {
            (Some(sbom), true) => sbom,
            (Some(_), false) => {
                gctx.shell()
                    .warn("ignoring 'sbom' config, pass `-Zsbom` to enable it")?;
                false
            }
            (None, _) => false,
        };
```

`cfg` is from `CargoBuildConfig` so it is a merged value from both env and config files. We don't need to query `CARGO_BUILD_SBOM` again.

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,239 @@
...
+    crate_type: Option<CrateType>,
+    name: String,
+    edition: String,
+}
+
+impl From<&Target> for SbomTarget {
+    fn from(target: &Target) -> Self {
+        SbomTarget {
+            kind: target.kind().clone(),
+            crate_type: target.kind().rustc_crate_types().first().cloned(),
```
weihanglo, line 123: Regarding the fact that we might have multiple crate-types in one unit, this seems to lose that information.

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,239 @@
...
+            name: target.name().to_string(),
+            edition: target.edition().to_string(),
+        }
+    }
+}
+
+#[derive(Serialize, Clone)]
+struct SbomRustc {
```
weihanglo, line 131: How do we want to represent a local build rustc?

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,239 @@
...
+        let output = serde_json::to_string(&sbom)?;
+        write!(outfile, "{}", output)?;
+    }
+
+    Ok(())
+}
+
+/// Fetch all dependencies, including transitive ones. A dependency can also appear multiple times
+/// if it's included with different versions.
+fn fetch_packages(build_runner: &mut BuildRunner<'_, '_>, unit: &Unit) -> Vec<SbomPackage> {
```
weihanglo, line 209: I may avoid using the word "fetch". In Cargo "fetch" usually refers to dealing things with the internet.

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,239 @@
...
+    while let Some(package) = queue.pop_first() {
+        if visited.contains(package) {
+            continue;
+        }
+
+        let build_type = if package.unit.mode.is_run_custom_build() {
+            SbomBuildType::Build
+        } else {
+            SbomBuildType::Normal
+        };
```
weihanglo, line 226: Could you share the idea of distinguishing the build script execution?

It is a bit odd that the only "crate" dependency a build script execution has is the build script "compilation".

Let me explain it a bit: In Cargo's unit graph there are two kinds of unit for build scripts. One for build script compilation, the other for execution. See this doc for more: <https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/compiler/custom_build/index.html>.

Yeah we got another similar question: what do we want to track in SBOM?

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,239 @@
...
+
+/// Fetch all dependencies, including transitive ones. A dependency can also appear multiple times
+/// if it's included with different versions.
+fn fetch_packages(build_runner: &mut BuildRunner<'_, '_>, unit: &Unit) -> Vec<SbomPackage> {
+    let unit_graph = &build_runner.bcx.unit_graph;
+    let root_deps = build_runner.unit_deps(unit);
+
+    let mut result: Vec<SbomPackage> = Vec::new();
+    let mut queue: BTreeSet<&UnitDep> = root_deps.iter().collect();
+    let mut visited: BTreeSet<&UnitDep> = BTreeSet::new();
```
weihanglo, line 215: Just personal taste, though I believe type some types can be inferred and don't need to explicit annotate.

tests/testsuite/sbom.rs
```diff
@@ -0,0 +1,333 @@
...
+                "debuginfo": 2,
+                "split_debuginfo": "{...}",
+                "debug_assertions": true,
+                "overflow_checks": true,
+                "rpath": false,
+                "incremental": false,
+                "panic": "unwind",
+                "strip": {
+                    "deferred": "None"
+                }
```
weihanglo, line 263: This needs to be resolved. Deferred is not too meaningful for SBOM if we do want this field.

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,239 @@
...
+    dependencies: Vec<SbomDependency>,
+}
+
+impl SbomPackage {
+    pub fn new(
+        dep: &UnitDep,
+        dependencies: Vec<SbomDependency>,
+        build_type: SbomBuildType,
+    ) -> Self {
+        let package_id = dep.unit.pkg.package_id();
```
weihanglo, line 89: Same as the top level package. Package ID Spec is the only stable representation in Cargo to describe a package version coordinate.

```suggestion
        let package_id = dep.unit.pkg.package_id().to_spec();
```

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,239 @@
...
+            version: rustc.version.to_string(),
+            wrapper: rustc.wrapper.clone(),
+            commit_hash: rustc.commit_hash.clone(),
+            host: rustc.host.to_string(),
+        }
+    }
+}
+
+#[derive(Serialize)]
+struct Sbom {
```
weihanglo, line 150: Do we need a dependencies field for this top-level `Sbom`?

(Just a question. I don't really know if other SBOM formats need it to recover the dependency graph)

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,239 @@
...
+    format_version: SbomFormatVersion<1>,
+    package_id: PackageIdSpec,
+    name: String,
+    version: String,
+    source: String,
+    target: SbomTarget,
+    profile: Profile,
```
weihanglo, line 157: Bad news. People can [override profiles for individual packages](https://doc.rust-lang.org/nightly/cargo/reference/profiles.html#overrides). So, only a top-level profile might not be enough to represent the build.

(The truth is, I am not an SBOM expert, so just provide information for you to consider)

tests/testsuite/sbom.rs
weihanglo, line 1: Could we have some extra tests

* A package with multiple `crate-types`.
* A package that artifact name conflicts.

tests/testsuite/sbom.rs
```diff
@@ -0,0 +1,333 @@
...
+                build = "build.rs"
+
+                [build-dependencies]
+                baz = "0.1.0"
+            "#,
+        )
+        .file("src/lib.rs", "pub fn bar() -> i32 { 2 }")
+        .file(
+            "build.rs",
+            r#"fn main() { println!("cargo::rustc-cfg=foo"); }"#,
```
weihanglo, line 206: You may need to add one more build script instruction `cargo::rustc-check-cfg=cfg(has_foo)` to avoid `unexpected_cfg` lint error.

See <https://blog.rust-lang.org/2024/05/06/check-cfg.html>.

weihanglo (commented):
Now I like the idea of having this PR to explore SBOM format. I'll post back issues we've found so far to the RFC. Thank you :)

src/cargo/core/compiler/output_sbom.rs
weihanglo: Just re-read the RFC and asked there <https://github.com/rust-lang/rfcs/pull/3553#discussion_r1593283479>.

src/cargo/core/compiler/output_sbom.rs
arlosi: Great point. This makes it seem like the profile needs to be captured for each package.

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,239 @@
...
+    let rustc: SbomRustc = bcx.rustc().into();
+
+    let packages = fetch_packages(build_runner, unit);
+
+    for sbom_output_file in build_runner.sbom_output_files(unit)? {
+        let sbom = Sbom::new(unit, packages.clone(), rustc.clone());
+
+        let mut outfile = BufWriter::new(paths::create(sbom_output_file)?);
+        let output = serde_json::to_string(&sbom)?;
+        write!(outfile, "{}", output)?;
```
arlosi, line 201: You can use `to_writer` here.
```suggestion
        serde_json::to_writer(outfile, &sbom)?;
```

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,239 @@
...
+            name: target.name().to_string(),
+            edition: target.edition().to_string(),
+        }
+    }
+}
+
+#[derive(Serialize, Clone)]
+struct SbomRustc {
+    version: String,
+    wrapper: Option<PathBuf>,
```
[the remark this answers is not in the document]
weihanglo: We should either record all or nothing.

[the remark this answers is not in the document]
Shnatsel: The wrapper is usually something like `sccache`. Could help with forensics, I suppose? Seems like a good idea to include all of them it if it's not too difficult, for more transparency in the build process.

src/cargo/core/compiler/output_sbom.rs
Shnatsel: Ideally, yes. Copying my comment from the RFC:

> Note that there are two ways of looking at dependencies: what each package needs, and the final resolved graph.
> 
> For example, if one package depends on `rand` with `features = ["std", "getrandom"]`, and another with `features = ["std", "simd_support"]`, the final resolved features will be `["std", "getrandom", "simd_support"]`. Depending on the use case you may need either or both representations (direct package dependencies and the resolved graph).
> 
> `cargo metadata` [exposes both](https://doc.rust-lang.org/cargo/commands/cargo-metadata.html) (under "packages" and "resolve" fields), but inaccurately: 
>  - https://github.com/rust-lang/cargo/issues/7754
>  - Under resolver v2, it [conflates the normal and build-dependency trees](https://rust-lang.github.io/rfcs/2957-cargo-features2.html#cargo-metadata)
> 
> I think it would be best for the SBOM to also expose both, accurately this time.

So what I would like to see is *two* resolved dependency trees: one for normal dependencies and one for build dependencies, matching the way feature resolver v2 works.

src/cargo/core/compiler/output_sbom.rs
weihanglo: Yes. It will be the same. `rustc.version` is `semver::Version`. However, we'll lose the information that rustc was built from a source tarball, not exactly from official distribution. Is this something SBOM want to track?

```
rustc 1.77.0 (aedd173a2 2024-03-17) (built from a source tarball)
binary: rustc
commit-hash: aedd173a2c086e558c2b66d3743b344f977621a7
commit-date: 2024-03-17
release: 1.77.0
LLVM version: 17.0.6
```

src/cargo/core/compiler/build_config.rs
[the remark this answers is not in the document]
justahero: Updated the code accordingly.

src/doc/src/reference/unstable.md
```diff
@@ -377,6 +378,29 @@ It's possible to update `my-dependency` to a pre-release with `update -Zunstable
...
+The `sbom` build config allows to generate so-called SBOM pre-cursor files
+alongside each compiled artifact. A Software Bill Of Material (SBOM) tool can
+incorporate these generated files to collect important information from the cargo
+build process that are difficult or impossible to obtain in another way.
+
+To enable this feature either set the `sbom` field in the `.cargo/config.toml`
+
+```toml
+[build]
+sbom = true
```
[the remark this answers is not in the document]
justahero: Updated.

src/cargo/core/compiler/output_sbom.rs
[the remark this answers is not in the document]
justahero: Added the workspace wrapper too.

src/cargo/core/compiler/output_sbom.rs
justahero: I think this information is tracked via [`Rustc.verbose_version`](https://github.com/rust-lang/cargo/blob/master/src/cargo/util/rustc.rs#L27).

src/cargo/core/compiler/output_sbom.rs
Shnatsel: Yes. It sounds like a good idea to me. This behavior just needs to be documented.

src/cargo/core/compiler/output_sbom.rs
justahero: The verbatim text is present in the SBOM output now.

src/cargo/core/compiler/output_sbom.rs
justahero: Updated the `package_id`.

justahero:
Rebased the PR.

tests/testsuite/sbom.rs
justahero: @weihanglo could you give insight in how best to set up these tests?

arlosi: I think we already resolved the artifact name conflict issue by appending the extension rather than changing it.

I'd like to see a test where we have a single crate that gets compiled multiple times (once through a build dependency, and once as a normal dependency) with different feature sets enabled. The crate should then be listed twice in the output. Resolver v2 is needed for this (enabled by default in 2021 edition crates).

tests/testsuite/sbom.rs
```diff
@@ -0,0 +1,637 @@
...
+                "panic": "unwind",
+                "strip": {
+                    "deferred": "None"
+                }
+            },
+            "packages": [
+                {
+                    "build_type": "build",
+                    "dependencies": [
+                        {
```
arlosi, line 361: I'm concerned with what happens when a crate is compiled two different ways under resolver v2. For example, once as a build dependency and once as a normal dependency. In this list of `dependencies` how would we tell the two entries apart?

Would it be better for the `dependencies` to be a sequence of indicies into the `packages` list?

arlosi: I'd also really like to see a test of this case.

src/cargo/core/compiler/build_runner/mod.rs
justahero: The SBOM file extension is appended, the existing file extension is not replaced.

src/cargo/core/compiler/output_sbom.rs
justahero: Updated

tests/testsuite/sbom.rs
justahero: It's the output of [`Profile`](https://github.com/rust-lang/cargo/blob/master/src/cargo/core/profiles.rs#L50) struct, but I think it's more sensible to not expose this type directly, in case its properties ever change or new fields are added.

justahero: The `strip` field got removed from the JSON output.

tests/testsuite/sbom.rs
justahero: A test has been added.

src/cargo/core/compiler/output_sbom.rs
justahero: Has this question been resolved?

tests/testsuite/sbom.rs
justahero: Hello @arlosi ,

I refactored the code that collects all packages & their dependencies. First all packages are collected, then for each package the indices of its dependencies are determined. It simplifies the serialization code slightly. Would like to get your feedback, if this is what you had in mind.

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+use super::{unit_graph::UnitDep, BuildRunner, CrateType, Unit};
+
+#[derive(Serialize, Clone, Debug, Copy)]
+#[serde(rename_all = "kebab-case")]
+enum SbomBuildType {
+    /// A package dependency
+    Normal,
+    /// A build script dependency
+    Build,
+}
```
epage, line 28: Should we be consistent with `cargo metadata` wrt th schema for this?

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+        serializer.serialize_u32(V)
+    }
+}
+
+/// A profile can be overriden for individual packages.
+///
+/// This wraps a [`Profile`] object.
+/// See <https://doc.rust-lang.org/nightly/cargo/reference/profiles.html#overrides>
+#[derive(Serialize, Clone, Debug)]
+struct SbomProfile {
```
epage, line 47: No case is put on this.  Is `snake_case` intentional? Looks like thats what we use for `cargo metadata`

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+            split_debuginfo: profile.split_debuginfo.map(|x| x.to_string()),
+            debug_assertions: profile.debug_assertions,
+            overflow_checks: profile.overflow_checks,
+            rpath: profile.rpath,
+            incremental: profile.incremental,
+            panic: profile.panic.to_string(),
+            rustflags,
+        }
+    }
+}
```
epage, line 88: A part of me wants to have us `profile.clone()` and destructure `Profile` so we make sure we update this whenever a new `Profile` field is added (and to make it easier to review to make sure all fields are present)

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+struct SbomPackage {
+    package_id: PackageIdSpec,
+    package: String,
+    /// Package profile is given when it differs from the root level [`Profile`].
+    #[serde(skip_serializing_if = "Option::is_none")]
+    profile: Option<SbomProfile>,
+    version: Option<Version>,
+    features: Vec<String>,
+    build_type: SbomBuildType,
+    extern_crate_name: String,
```
epage, line 101: Why do we need this?

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+            rustflags,
+        }
+    }
+}
+
+/// Describes a package dependency
+#[derive(Serialize, Clone, Debug)]
+struct SbomPackage {
+    package_id: PackageIdSpec,
+    package: String,
```
epage, line 94: whats `package`?

epage: I guess its the name, but why include it (and `version`) when its in `package_id`?

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+        let profile = if &unit_dep.unit.profile != root_profile {
+            Some((&unit_dep.unit.profile).into())
+        } else {
+            None
+        };
+        let build_type = if unit_dep.unit.mode.is_run_custom_build() {
+            SbomBuildType::Build
+        } else {
+            SbomBuildType::Normal
+        };
```
epage, line 119: The documentation for `is_run_custom_build` says
```rust
    /// Returns `true` if this is the *execution* of a `build.rs` script.
```

So this isn't saying whether this is a normal or build instance but whether this will be run as a build script.

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+            incremental: profile.incremental,
+            panic: profile.panic.to_string(),
+            rustflags,
+        }
+    }
+}
+
+/// Describes a package dependency
+#[derive(Serialize, Clone, Debug)]
+struct SbomPackage {
```
epage, line 92: Did we decide whether the sbom will track packages instead of crates?

src/cargo/core/compiler/output_sbom.rs
epage: as in build scripts show up twice, once for the Build mode and once for the Run mode

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+            commit_hash: rustc.commit_hash.clone(),
+            host: rustc.host.to_string(),
+            verbose_version: rustc.verbose_version.clone(),
+        }
+    }
+}
+
+#[derive(Serialize)]
+struct Sbom {
+    format_version: SbomFormatVersion<1>,
```
epage, line 186: Not-blocking: eventually we should move this out into `cargo-util-schemas` and provide `Deserialize` impls so people can use that to read this file.  `SbomFormatVersion` isn't compatible with that.

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+    package_id: PackageIdSpec,
+    name: String,
+    version: String,
+    source: String,
+    target: SbomTarget,
+    profile: SbomProfile,
+    packages: Vec<SbomPackage>,
+    features: Vec<String>,
```
epage, line 194: Isn't most of this redundant with `packages`, is there a reason we don't just have a `root_package` field?

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+    package: String,
+    /// Package profile is given when it differs from the root level [`Profile`].
+    #[serde(skip_serializing_if = "Option::is_none")]
+    profile: Option<SbomProfile>,
+    version: Option<Version>,
+    features: Vec<String>,
+    build_type: SbomBuildType,
+    extern_crate_name: String,
+    /// Indices into the `packages` array
+    dependencies: Vec<usize>,
```
epage, line 103: Should we newtype the `usize`?  On one hand, it will make it easy to find all all `usize`s that fill those role.  On the other hand, it will make indexing more annoying.

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+            .iter()
+            .filter_map(|dep| {
+                visited_dependencies
+                    .iter()
+                    .position(|unit_dep| dep == unit_dep)
+            })
+            .collect::<Vec<_>>();
+
+        if let Some(package) = packages.get_mut(index) {
+            package.dependencies.append(&mut indices);
```
epage, line 287: This assumes that the order we iterate through `visitied_dependencies` (`BTreeSet`) is the same order things appear in `packages (`Vec`).

Should `visited_dependencies` be an `IndexSet`?

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+    // They have the same number of objects in the same order, therefore using an index is fine.
+    for (index, package) in visited_dependencies.iter().enumerate() {
+        let dependencies: BTreeSet<&UnitDep> = unit_graph[&package.unit].iter().collect();
+
+        let mut indices = dependencies
+            .iter()
+            .filter_map(|dep| {
+                visited_dependencies
+                    .iter()
+                    .position(|unit_dep| dep == unit_dep)
```
epage, line 282: Is this `O(n^3)`?  Is the equality check doing a straight pointer comparison or a deep check?

src/doc/src/reference/unstable.md
```diff
@@ -396,6 +397,32 @@ It's possible to update `my-dependency` to a pre-release with `update -Zunstable
 This is because `0.1.2-pre.0` is considered compatible with `0.1.1`.
 It would not be possible to upgrade to `0.2.0-pre.0` from `0.1.1` in the same way.
 
+## sbom
```
epage, line 400: Should we document the schema in here?  yes, the RFC will have it but that will then be a snapshot in time and eventually we'll want to have it documented.

tests/testsuite/sbom/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+        .file("src/main.rs", r#"fn main() {}"#)
+        .build();
+
+    p.cargo("build -Zsbom")
+        .masquerade_as_nightly_cargo(&["sbom"])
+        .run();
+
+    let file = with_sbom_suffix(&p.bin("foo"));
+    let output = std::fs::read_to_string(file).unwrap();
+    assert_e2e().eq(output, snapbox::file!["build_sbom_using_cargo_config.json"]);
```
epage, line 64: Hmm, having external snapshots will make it a lot harder to review these tests because I have to find the right test and right file and jump back and forth between them

tests/testsuite/sbom/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+    p.cargo("build -Zsbom")
+        .masquerade_as_nightly_cargo(&["sbom"])
+        .run();
+
+    assert!(with_sbom_suffix(&p.bin("foo")).is_file());
+    assert_eq!(
+        2,
+        p.glob(p.target_debug_dir().join("*.cargo-sbom.json"))
+            .count()
+    );
```
epage, line 111: Under what conditions is an rlib sbom created?

tests/testsuite/sbom/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+
+fn configured_project() -> ProjectBuilder {
+    project().file(
+        ".cargo/config.toml",
+        r#"
+            [build]
+            sbom = true
+        "#,
+    )
+}
```
epage, line 28: Personally, I find helpers like this not too helpful
- They obfuscate what the test is doing (it should at least say what the config is)
- They don't scale if you then want to do something else with the config

tests/testsuite/sbom/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+
+#[cargo_test]
+fn build_sbom_with_artifact_name_conflict() {
+    Package::new("deps", "0.1.0")
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "deps" # name conflict
+                version = "0.1.0"
```
epage, line 122: What is this test showing?

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,292 @@
...
+
+impl SbomPackage {
+    pub fn new(unit_dep: &UnitDep, root_profile: &Profile) -> Self {
+        let package_id = unit_dep.unit.pkg.package_id().to_spec();
+        let package = package_id.name().to_string();
+        let profile = if &unit_dep.unit.profile != root_profile {
+            Some((&unit_dep.unit.profile).into())
+        } else {
+            None
+        };
```
epage, line 114: Not seeing a test that demonstrates this conditional profile

src/doc/src/reference/unstable.md
arlosi: Yes, I think that would be helpful. Added.

arlosi (commented):
I've the PR to address most of the feedback from @epage. Thanks!

I re-wrote the algorithm for building the sbom graph and made changes to the format. There's also now documentation for the format in `unstable.md` that should help reviewers understand what it's looking like.
 
Before merging, I still want to add more test coverage for additional cases.

src/cargo/core/features.rs
```diff
@@ -1297,10 +1297,10 @@ impl CliUnstable {
             "rustdoc-map" => self.rustdoc_map = parse_empty(k, v)?,
             "rustdoc-scrape-examples" => self.rustdoc_scrape_examples = parse_empty(k, v)?,
             "sbom" => self.sbom = parse_empty(k, v)?,
-            "script" => self.script = parse_empty(k, v)?,
             "separate-nightlies" => self.separate_nightlies = parse_empty(k, v)?,
             "checksum-freshness" => self.checksum_freshness = parse_empty(k, v)?,
             "skip-rustdoc-fingerprint" => self.skip_rustdoc_fingerprint = parse_empty(k, v)?,
+            "script" => self.script = parse_empty(k, v)?,
             "target-applies-to-host" => self.target_applies_to_host = parse_empty(k, v)?,
```
epage, line 1304: Why was this moved?  I believe we generally try to be sorted here

src/cargo/core/compiler/output_sbom.rs
epage: As for the use of the name `normal`, we use it today in
- [cargo tree](https://doc.rust-lang.org/cargo/commands/cargo-tree.html#tree-options)
- [index](https://doc.rust-lang.org/cargo/reference/registry-index.html#json-schema)
- [publish](https://doc.rust-lang.org/cargo/reference/registry-web-api.html#publish)

seems like there is enough precedence for this.

We should probably make sure this information gets reflected in the RFC

src/cargo/core/features.rs
arlosi: It was inadvertently moved in the 1st commit when merging. The 2nd commit fixes it.

If you prefer, I can squash the two commits together. I was hoping it would be easier to review only the changes since last time, so I added a commit rather than rewriting the 1st one.

src/cargo/core/compiler/output_sbom.rs
epage: When I wrote up our schema docs, I said to be `kebab-case`.  I don't remember what all led to that (focusing on `Cargo.toml`?)

However, looking at our code, we never explicitly state `snake_case`, only `kebab-case`.

src/cargo/core/compiler/output_sbom.rs
epage: I suspect artifact dependencies could make things interesting

Note that we do call the field in the format `crates`

src/cargo/core/features.rs
epage: I've done enough of a pass; let's squash

src/cargo/core/compiler/build_runner/mod.rs
```diff
@@ -291,6 +293,14 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
             }
 
             super::output_depinfo(&mut self, unit)?;
+
+            if self.bcx.build_config.sbom {
+                let sbom = super::build_sbom(&mut self, unit)?;
+                for sbom_output_file in self.sbom_output_files(unit)? {
+                    let outfile = BufWriter::new(paths::create(sbom_output_file)?);
+                    serde_json::to_writer(outfile, &sbom)?;
+                }
```
epage, line 302: We are outputting the same sbom for each output.

If we're package focused, that makes sense.  If we're create focused, then that doesn't quite work.

Is there anything specific about each root artifact that we'd want to call out?

src/cargo/core/compiler/mod.rs
```diff
@@ -711,6 +714,11 @@ fn prepare_rustc(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> CargoResult
...
+            let file_list = std::env::join_paths(build_runner.sbom_output_files(unit)?)?;
+            base.env("CARGO_SBOM_PATH", file_list);
```
epage, line 720: Can you document this in the unstable docs like it will be when moved to the stable docs?

I want to make sure we call out that it can be multiple files

src/cargo/core/compiler/build_runner/mod.rs
```diff
@@ -446,6 +456,44 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
         self.files().metadata(unit).unit_id()
     }
 
+    /// Returns the list of SBOM output file paths for a given [`Unit`].
+    ///
+    /// Only call this function when `sbom` is active.
+    pub fn sbom_output_files(&self, unit: &Unit) -> CargoResult<Vec<PathBuf>> {
```
epage, line 462: Do these need to be calculated in `prepare_units` and added to `files`?

src/cargo/core/compiler/build_runner/mod.rs
```diff
@@ -446,6 +456,44 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
...
+            TargetKind::CustomBuild => false,
+        };
+        if !sbom_enabled {
+            return Ok(Vec::new());
+        }
+
+        let files = self
+            .outputs(unit)?
+            .iter()
+            .filter(|o| matches!(o.flavor, FileFlavor::Normal | FileFlavor::Linkable))
```
epage, line 491: Do we have a place to track specific points we want to have called out in the RFC?  For me, when we do this is one of them

src/cargo/core/compiler/output_sbom.rs
arlosi: This is resolved now. The "build script build" and "run build script" units do not appear in the SBOM graph anymore. The graph edges that go to dependencies of build scripts are marked as "build" edges.

src/cargo/core/compiler/output_sbom.rs
arlosi: I did `snake_case` for consistency with `cargo_metadata`, but if there's a reason for switching to `snake-case` for some of these, let me know.

src/cargo/core/compiler/output_sbom.rs
arlosi: I haven't tried artifact deps. I can add a test and we can revisit.

src/cargo/core/compiler/output_sbom.rs
weihanglo: it's called kebab-case btw 😛

arlosi: Oops, typo!

src/cargo/core/compiler/output_sbom.rs
```diff
@@ -0,0 +1,245 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    codegen_backend: Option<InternedString>,
+    debuginfo: DebugInfo,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    split_debuginfo: Option<InternedString>,
+    debug_assertions: bool,
+    overflow_checks: bool,
+    rpath: bool,
+    panic: PanicStrategy,
+}
```
weihanglo, line 43: Not a blocker but I would love to understand why some are not picked.

* codegen-units might have impact on final artifacts just like lto does.
* Why codegen-backend is selected, and rustflags isn't. Both of them are unstable though.

…and etc.

src/cargo/core/compiler/output_sbom.rs
weihanglo, line 1: Should we also track `--target` in the sbom.json? It is fairly important to know what platform the artifact is for.

https://github.com/rust-lang/rfcs/pull/3553#discussion_r1955292307

weihanglo: Not a blocker for the MVP though.

src/cargo/core/compiler/build_runner/mod.rs
weihanglo: Had an offline discussion with Arlo. Here are some thoughts:

If it were an `OutputFile`, then

* It is tracked in `.d` depfinfo.
* It will be incluced in [JSON artifact message](https://doc.rust-lang.org/nightly/cargo/reference/external-tools.html#artifact-messages).
* It may need to output to `target/debug/deps` first then be hardlink'd to top level `target/debug`.

It is a bit odd making sbom an `OutputFile` because:

* Arlo has experimented on it, and found that SBOM will be tracked by depinfo if it were an tracked `OutputFile`.
* `OutputFile`, as [its comment says](https://github.com/rust-lang/cargo/blob/3b784a42e3075ebf34591ebfd5b79457e680e69d/src/cargo/core/compiler/build_runner/compilation_files.rs?plain=1#L134), is for files emitted by compiler. SBOM is not a thing from compiler.
* External tools might be confused and need to exclude SBOM from JSON message `filenames` array, as they might not be interested in it.
* The current SBOM format doesn't really tracked the output artifact in its content. The associated artifact information is embedded in the SBOM's file name. If SBOM is seen as a output artifact, do we need to have an SBOM for SBOM itself. (This is half-joking though).

src/cargo/core/compiler/output_sbom.rs
arlosi: Done.

src/cargo/core/compiler/build_runner/mod.rs
arlosi: I've added a list to the PR description

src/cargo/core/compiler/output_sbom.rs
arlosi: For this initial implementation, I've dropped the `Profile` section. We can re-add it if needed when a rationale is available.

src/cargo/core/compiler/output_sbom.rs
epage: Added this to the PR description so we ensure we track this

src/cargo/core/compiler/output_sbom.rs
epage: Rather than block on this, I noted it in the PR description for raising this in the RFC

tests/testsuite/sbom/mod.rs
epage, line 1: Why is this in its own directory?

src/cargo/core/compiler/build_runner/mod.rs
epage: > It is tracked in .d depfinfo.

What are the trade offs of this?

> It will be incluced in [JSON artifact message](https://doc.rust-lang.org/nightly/cargo/reference/external-tools.html#artifact-messages).

That seems like a positive to be communicated out about a generated artifact file

src/cargo/core/compiler/output_sbom.rs
epage: Any updates on this?

src/cargo/core/compiler/build_runner/mod.rs
arlosi: Focusing on the packages makes more sense for SBOMs to me, since that's what's published to crates.io, has CVEs issued against, etc. That said, it's important that we call out if a package is compiled in different ways (different features, etc).

I can't think of anything specific about each root that we'd need to call out.  I don't see an issue with having duplicate contents. If needed, we can always add output-specific information.

src/cargo/core/compiler/output_sbom.rs
arlosi: I've added a test for artifact deps. They seem to work fine -- similar to other dependencies. If the artifact dep is a build-dependency, then it's noted as such in the sbom graph. Artifact deps that are normal dependencies also show up as expected.

We could add special support in the sbom format for denoting artifact dependencies, but I don't think it's necessary for this MVP.

src/cargo/core/compiler/build_runner/mod.rs
arlosi: > That seems like a positive to be communicated out about a generated artifact file

It's still a bit of a grey area for me, since the generated artifact file is not produced by rustc. I'm leaning back towards including it in the artifact message though.

src/cargo/core/compiler/build_runner/mod.rs
epage: If we're including the same output, then we're including non-relevant dependencies.  I thought providing focused dependencies was one of the things they were wanting compared to `cargo metadata`.  Granted, this is much easier to trim the unreachable branches but I think its still relevant.

It also feels weird to me to output per artifact the exact same output but I guess that also allows more future evolution.

src/cargo/core/compiler/output_sbom.rs
epage: What is necessary for the MVP is knowing whether we want one or two entries if a package is used both as a lib dependency and an artifact dependency.

Thinking about this and metabuild (which will build on artifact deps), I wonder if we should focus the schema on crates *and* represent build scripts in this with a crate having a build dependency on a build script and build scripts having normal dependencies on everything else.  As we add artifact dep support, we mark the build script and each proc macro as a build artifact dep and all normal artifact deps as such.  Then when we introduce metabuild, each build artifact dep will be a represented the same was as the build script itself.

src/cargo/core/compiler/build_runner/mod.rs
arlosi: The duplicate contents only occurs when a build unit produces multiple outputs (like `staticlib` + `dylib`). And I don't think it's possible to configure different dependencies for each output type, so the SBOMs should be correctly pruned.

Each root unit within the primary package (say an `example` vs the `main` binary) *can* have different SBOMs. The example, may have `dev-dependencies` that that `main` would not. 

If we're compiling for multiple target triples, those would each also get their own root unit (and their own individual SBOM).

If there's a specific case you're thinking of, let me know and I can add a test to show how it behaves.

epage: Ok, I must have misunderstood this to be across units, not within a unit (yup, its within a unit for loop)

src/cargo/core/compiler/output_sbom.rs
epage: Since this is tracking crates, can we rename `SbomPackage` to `SbomCrate` and `Sbom::packages` to `Sbom::crates`?

src/cargo/core/compiler/build_runner/mod.rs
arlosi: I've added it as a new `FileFlavor::Sbom`

arlosi:
I've updated the PR and it should be ready for another round of review. Notable changes include:
* The graph is no longer combining dependencies within the same package. This means that things like libs and build scripts within a package get unique nodes in the graph.
* The SBOM is listed in the JSON output as an output file.
* Added a test for RUSTC_WRAPPER

epage (approved):

justahero:
Huge thank you @arlosi & to the cargo team for investing the time & effort to get this feature integrated. 🎉

nazar-pc:
Is this supposed to work with `cargo rustc` (I hope so)? It doesn't complain about the option, but doesn't produce `*.cargo-sbom.json` file either, while `cargo build` does.

weihanglo:
IIRC that hasn't yet been discussed.

PR #6329, opened 2018-11-18
Make autodiscovery disable inferred targets

dwijnand:
Fixes #6205

PR #5335, opened 2018-04-10
Introduce autoXXX keys for target auto-discovery

dwijnand:
In Rust 2015 absence of the configuration makes it default to not
include auto-discovered targets (i.e false), with a warnings message.

In Rust 2018 absence makes it default to include auto-discovered
targets (i.e true).

Fixes #5330

---

_original_

Fixes #5330

submitted for early review. feedback very welcome, I'm happy to iterate (and learn).

in particular I require assistance with the borrowing of `warnings` in `clean_benches`.

Issue #13695, opened 2024-04-02
Include cargo_vcs_info.json even for dirty working directories


kornelski:
### Problem

Currently, Cargo includes a `.cargo_vcs_info.json` file when publishing, but only if the working copy is clean and `--allow-dirty` is not used.

However, this metadata is needed regardless whether the directory was modified or not. It's unlikely the path in repo would change, and it is still needed for resolving relative paths in README, creating links to the crate, and still helpful for quickly finding the crate in its repository.

Even having the commit hash is still useful, because it provides a base commit to diff from.

There are situations where it's still necessary to edit the `Cargo.toml` file, e.g. to publish crates with circular dev dependencies (#4242), or explicitly specified tests/benchmarks that are excluded from the tarball (#13456). In this case the commit hash is basically still accurate.

Note that Cargo omitting the commit hash when the directory is dirty is not a guarantee that packages with a commit hash are matching their commit. Dishonest users can always modify Cargo to make it lie or upload a manipulated tarball themselves, so this file is informational, not an integrity guarantee.

### Proposed Solution

Cargo could always include the commit hash when available. Perhaps add `"dirty":true` if reporting that state is deemed important. Alternatively, provide an option of `--allow-dirty-but-still-include-the-commit-hash`.

epage:
Is there a reason to be mutating `Cargo.toml` for `dev-dependencies` these days with the stripping of non-registry dev-dependencies?

For #13456, I am actively working on a fix.  I posted the last major refactor today (#13693) before I can make the relevant change.

That said, I agree with the idea and it makes me wonder why this wasn't considered before.  #5886 and its predecessor don't really say much on motivation.  Same with #9866.

epage:
The biggest risk with this is if someone assumes the presence of `cargo_vcs_info.json` means the `.crate` is clean.

I talked to @joshtriplett to get some historical perspective and they don't see an issue and don't think that above use case should be large enough for us to work around (e.g. adding a `cargo_vcs_info-2.json` file)

kornelski:
https://docs.rs/crate/async-recursion/1.1.0/source/Cargo.toml.orig

is an example of a crate that removed non-workspace crates-io dev-dep to avoid a cycle (tokio-test)

epage:
I'm not seeing `tokio-test` in the [source Cargo.toml](https://docs.rs/crate/async-recursion/1.1.0/source/Cargo.toml.orig).

[...]

PR #13713, opened 2024-04-05
fix(toml): Warn, rather than fail publish, if a target is excluded

### LINKED ISSUE #5806: Cargo isn't consistent wrt "automatic" files and vendoring
Say you have a crate vendored. That crate doesn't have a build script. Now, there are two ways to add a build script. One involves adding an entry to Cargo.toml, which fails because of the checksum change in Cargo.toml, and another is to just add a `build.rs` file.

The latter silently succeeds (and yes, the build script is built and run).

Or is silently ignored if the crate was already built thanks to #5805.

### LINKED ISSUE #13456: cargo publish has bad error message when explicit `[[bench]]` isn't in `package.include`

epage:

We have a couple of problems with publishing
- Inconsistent errors: if a target that `package` doesn't verify is missing `path`, it will error, while one with `path` won't, see #13456
- Users may want to exclude targets and their choices are
  - Go ahead and include them.  I originally excluded my examples before doc-scraping was a think.  The problem was if I had to set `required-features`, I then could no longer exclude them
  - Muck with `Cargo.toml` during publish and pass `--allow-dirty`

This fixes both by auto-stripping targets on publish.  We will warn the user that we did so.

This is a mostly-one-way door on behavior because we are turning an error case into a warning.
For the most part, I think this is the right thing to do.
My biggest regret is that the warning is only during `package`/`publish` as it will be too late to act on it and people who want to know will want to know when the problem is introduced.
The error is also very late in the process but at least its before a non-reversible action has been taken.
Dry-run and `yank` help.

Fixes #13456
Fixes #5806

Tests are added in the first commit and you can then follow the commits to see how the test output evolved.

The biggest risk factors for this change are
- If the target-stripping logic mis-identifies a path as excluded because of innocuous path differences (e.g. case)
- Setting a minimum MSRV for published packages: `auto*` were added in 1.27 (#5335) but were insta-stable.  `autobins = false` did nothing until 1.32 (#6329).  I have not checked to see how this behaves pre-1.32  or pre-1.27.  Since my memory of that error is vague, I believe it will either do a redundant discovery *or* it will implicitly skip discovery

Resolved risks
- #13729 ensured our generated target paths don't have `\` in them
- #13729 ensures the paths are normalize so the list of packaged paths

For case-insensitive filesystems, I added tests to show the original behavior (works locally but will fail when depended on from a case-sensitive filesystem) and tracked how that changed with this PR (on publish warn that those targets are stripped).  We could try to normalize the case but it will also follow symlinks and is likely indicative of larger casing problems that the user had.  Weighing how broken things are now , it didn't seem changing behavior on this would be too big of a deal.

We should do a Call for Testing when this hits nightly to have people to `cargo package` and look for targets exclusion warnings that don't make sense.

### Additional information

This builds on #13701 and the work before it.

By enumerating all targets in `Cargo.toml`, it makes it so rust-lang/crates.io#5882 and rust-lang/crates.io#814 can be implemented without any other filesystem interactions.

A follow up PR is need to make much of a difference in performance because we unconditionally walk the file system just in case `autodiscover != Some(false)` or a target is missing a `path`.

We cannot turn off auto-discovery of libs, so that will always be done for bin-only packages.


epage:
Blocked on #13701.  Posting now to get Windows testing.

epage:
@ChrisDenton I'd be curious if you have any thoughts on the case-preserving change in this PR.

Existing behavior (shown in the first commit): we can detect `SrC/MaIn.rs` as `src/main.rs` and publish it.  Windows consumers will be able to use the package.  Linux consumers of it won;'t.

With us tracking targets in the published package, `src/main.rs` (what cargo sees) and `SrC/MaIn.rs` (what the filesystem sees) wouldn't match and we'd strip the target from the published manifest, preventing anyone from using it.  The publisher will get a warning.

In the last commit, I use `canonicalize` to try to read the case from the filesystem and preserve that through Cargo, allowing the targets to be included and Linux users can now consume them.

This is shown in the commit history of the PR with the changes to the `package::normalize_case` test.

The one case not handled is if the user sets `path = "src/main.rs"` but the file is `SrC/MaIn.rs` then we will still strip the target.

ChrisDenton:
Hmm... I would just quickly note the case sensitivity is not necessarily just a cross-platform concern. E.g. Windows can have case-sensitive directories. So anywhere there is a mismatch between expected and actual case means that it's non-portable even possibly within the same system. This does make erroring on detected case mismatches more appealing to me as I'd personally want to be made very aware of the situation before I publish, especially if Cargo is unable to handle all such cases automatically. However, I can also see how that might be frustrating for other users so I don't feel that strongly about it. "It just works" is always appealing.

On a technical level I would worry a bit about canonical paths not necessarily matching the packaged directory structure, due to symlinks or something. Admittedly I'm uncertain if this would be a problem in practice or not.

epage:
@ChrisDenton thanks for the input!  I've taken out that commit for now and instead going with the approach of it needing a strong justification to go in.  I want to do a call-for-testing on this when it hits nightly and will particularly call out publishing from Windows.

epage:
Starting the conversation on the behavior changes in parallel to the review of the implementation

This change snapshots implicit build targets into the published package, making them explicit.

Immediate benefits
- Makes implicit targets, explicit-without-path targets, and explicit-with-path targets consistent for error reporting (#13456) and for intentionally excluding targets (not needing `--allow-dirty` (as discussed in #13695)
- Disallows adding `build.rs` to vendored source (#5806)
- As a crate maintainer, makes it easier to inspect what the published result is

This unblocks
- Performance optimizations.  This allows us to shift all of the burden onto TOML parsing which would likely be easier to optimize
- crates.io specializing their rendering on bin vs lib (rust-lang/crates.io#5882, rust-lang/crates.io#814)

Behavior changes that I'm looking for consensus on:
- Switching some publish errors to warnings
  - Excluded implicit target
    - Before: silently ignored
    - After: silently ignored
  - Excluded explicit target without a path
    - Before: error since a path can't be discovered
    - After: silently ignored
  - Excluded explicit target with a path
    - Before: error if its a bin/lib, silently ignored otherwise
    - After: silently ignored
- Case mismatches
  - Situations:
    - Implicit file target with alt case of (e.g. `src/Main.rs`)
    - Implicit target search with alt case of (e.g. `Src/Bin/`)
    - Explicit target without `path` whose `name` has different case than actual file
    - Explicit target whose `path` has different case than the actual file
  - Before:
    - Publish worked
    - Package would work on case-insensitive filesystems
    - Package would fail on case-sensitive filesystems
  - After:
    - Ignored with a warning during publish
  - Justification:
    - This is a portability hazard, even within the same machine
    - We've not seen people organically bring up the problem that their packages don't work on other platforms.  The main organic case situation was `cargo.toml` and that mostly came about because windows users didn't think to add the required case.  This is the opposite.
    - These case mismatches could be indicative of more fundamental case problems and this could have a long tail of holes to patch
  - Alternative: normalize the case
    - One option is to use `canonicalize` but that will also resolve symlinks
    - We could enumerate directories and find a best match.  This avoids "how is case-insensitivity implemented" because the cases we care about are ASCII.  However, we don't really know whether the filesystem is case-sensitive or not

@rfcbot fcp merge

src/cargo/ops/cargo_package.rs
```diff
@@ -689,7 +689,11 @@ fn tar(
 
     let base_name = format!("{}-{}", pkg.name(), pkg.version());
     let base_path = Path::new(&base_name);
-    let publish_pkg = prepare_for_publish(pkg, ws)?;
+    let included = ar_files
```
weihanglo, line 692: mit: Should this be a set instead of an array?

weihanglo (approved):
Implementation-wise looks great. Thanks a lot!

Just wonder if we could document somewhere about what “resolved” means in these contexts. Seems like at least

* workspace inheritance is resolved.
* target discovery is resolved.

That would help contributors pretty much I believe.

weihanglo (approved):

weihanglo:
@bors r+

Thanks!

glandium:
How is this related to #5806?

epage:
If I understand correctly `cargo vendor` will vendor "published" `Cargo.toml` files.  With #5806, you could drop a `build.rs` and it would be automatically picked up.  After this change, it will be ignored and the only way to add a `build.rs` is to hand-edit the `Cargo.toml` which will result in a checksum error.

glandium:
So, if I understand correctly, this only "fixes" #5806 for crates published to crates.io, but e.g. crates vendored from a git repo, will still have the problem.

epage:
From my understanding `cargo vendor` effectively runs `cargo package` on every dependency, regardless of the source.  This change affects `cargo package`.

I tried this out locally by running `cargo vendor` on https://github.com/crate-ci/cargo-release/ which has git dev-dependencies.  The vendored versions of those git dev-dependencies had the standard "this is from cargo package" boilerplate at the top and they had all of the `auto*` fields set to `false` from this change.

So I believe this fixes it for git as well.  Feel free to test it when this hits nightly and let us know if it works or not!

glandium:
I tested 1.80.0-nightly (c987ad527 2024-05-01), and vendoring indeed adds the `auto*` and `build` fields (and more).

ia0:
FYI: This PR might have introduced a "warning" regression: https://github.com/rust-lang/cargo/issues/14290.

PR #13715, opened 2024-04-06
doc(semver): remove mention of deprecated tool rust-semverver

weihanglo:
See also

* https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Creating.20a.20release.20for.20rust-semverver
* https://github.com/rust-lang/team/pull/1390
* https://github.com/rust-lang/rust-semverver/issues/373


Kobzol:
Maybe we could mention cargo-semver-checks instead? It seems to be a spiritual successor of ruat-semverver.

PR #13718, opened 2024-04-06
Fix github fast path redirect.

ehuss:
This fixes the GitHub fast-path check to look up the sha of a git ref. At some point, GitHub changed the API to redirect to a different URL. Currently cargo is failing the fast-path lookup with 301 response code.

This can be tested in a project with a git dependency, and running `CARGO_LOG=cargo::sources::git::utils=debug cargo fetch` to verify it is picking up the fast path. This currently can't be tested in CI due to #13563.


weihanglo (approved):

weihanglo:
Always following directions looks pretty reasonable. Thanks!

@bors r+

PR #13721, opened 2024-04-07
[doc]:Changed `cargo:paseto` to `cargo::paseto`.

heisen-li:

I understand from [here](https://github.com/heisen-li/cargo/commit/bd230e810cbf307de33e974d1b59a3cf7389fe8c) that `cargo:` is currently deprecated, but there doesn't seem to be a change in the documentation.

This is a not-so-important issue, so feel free to close it if I've misunderstood.

### DIFF
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -108,7 +108,7 @@ For the latest nightly, see the [nightly version] of this page.
     * [`cargo config`](#cargo-config) --- Adds a new subcommand for viewing config files.
 * Registries
     * [publish-timeout](#publish-timeout) --- Controls the timeout between uploading the crate and being available in the index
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR! The original is correct. The single colon syntax is used for [built-in providers](https://doc.rust-lang.org/nightly/cargo/reference/registry-authentication.html#built-in-providers) which is something very different from build-script instructions.

heisen-li:
> Thanks for the PR! The original is correct. The single colon syntax is used for [built-in providers](https://doc.rust-lang.org/nightly/cargo/reference/registry-authentication.html#built-in-providers) which is something very different from build-script instructions.

Thanks for the correction, that's knowledge I didn't have. I was too reckless.

PR #13725, opened 2024-04-09
Improve Workspaces docs page

### LINKED ISSUE #13580: Improve Workspaces docs page
### Problem

https://doc.rust-lang.org/cargo/reference/workspaces.html is comprehensive, but could benefit from some clarifications and examples of how to create a workspace

Anyone trying to create a workspace would have to piece together multiple code snippets and go by trial and error.
Most people probably get it right the first time, but making it clearer in the docs will help.

### Proposed Solution

- explain the behavior of a root package (or primary package or root mainfest, it has different names even in the source code 😓).
- explain the resolver part where it is first mentioned
- add examples of creating a workspace step-by-step, manual and/or using cargo commands
- consider https://matklad.github.io/2021/08/22/large-rust-workspaces.html

Copied from and earlier discussion (https://github.com/rust-lang/cargo/issues/13579#issuecomment-1996186120)

### Notes

Related:

- Stems from #13579  (Improve workspace resolver error message when root edition is missing)
- #8365 (Add cargo new --workspace to create a workspace)
- #5151 (Feature request: Cargo templates)
- #11234 (cargo init is not working great when creating a new workspace with multiple members)

heisen-li:

Fixes #13580


src/doc/src/reference/workspaces.md
```diff
@@ -15,6 +15,9 @@ The key points of workspaces are:
   sections in `Cargo.toml` are only recognized in the *root* manifest, and
   ignored in member crates' manifests.
 
+The *workspace root* is the directory where the workspace's `Cargo.toml` 
+is located.
```
epage, line 19: This feels a bit disjointed to define this term here and in this way.

src/doc/src/reference/workspaces.md
```diff
@@ -30,6 +33,10 @@ In the `Cargo.toml`, the `[workspace]` table supports the following sections:
 * [`[replace]`](overriding-dependencies.md#the-replace-section) --- Override dependencies (deprecated).
 * [`[profile]`](profiles.md) --- Compiler settings and optimizations.
 
+## Creating a Workpaces
+
+todo!：Added example of creating a workspace.
+
```
epage, line 39: Due to the challenges in writing this documentation, I agree with @weihanglo about keeping each PR focused.  These changes will likely involve a lot of iteration and coupling them together in one PR will likely make this longer and more exhausting

src/doc/src/reference/workspaces.md
```diff
@@ -30,6 +33,10 @@ In the `Cargo.toml`, the `[workspace]` table supports the following sections:
 * [`[replace]`](overriding-dependencies.md#the-replace-section) --- Override dependencies (deprecated).
 * [`[profile]`](profiles.md) --- Compiler settings and optimizations.
 
+## Creating a Workpaces
+
+todo!：Added example of creating a workspace.
```
epage, line 38: btw this is a reference-level page.  I feel like this would probably belong in the Guide section.  However, I'm not quite too sure what all we'd want to cover in a guide-level section on workspaces.  Might be good to discuss this in the issue first.

src/doc/src/reference/workspaces.md
```diff
@@ -57,6 +65,10 @@ version = "0.1.0"    # the current version, obeying semver
 authors = ["Alice <a@example.com>", "Bob <b@example.com>"]
 ```
 
+If no other members are specified in [workspace.members filed](#the-members-and-exclude-fields) 
+in `Cargo.toml` in the workspace root directory, the root package will 
+become the default workspace member.
```
epage, line 70: For me, this feels a bit out of place with the section it is in.  Nothing else is talking about members or default members.  We also have a section down below for both members and default members which this pertains to.

src/doc/src/reference/workspaces.md
heisen-li: Sorry this is a bit of my own immature consideration.

[cargo book](https://doc.rust-lang.org/stable/cargo/reference/workspaces.html#workspaces):

> All packages share a common [Cargo.lock](https://doc.rust-lang.org/stable/cargo/guide/cargo-toml-vs-cargo-lock.html) file which resides in the workspace root.
> All packages share a common [output directory](https://doc.rust-lang.org/stable/cargo/guide/build-cache.html), which defaults to a directory named target in the workspace root.

The use of `workspace root` has already appeared in previous descriptions, and I personally feel that it may be easier to read the definitions by placing them in the nearest location.

src/doc/src/reference/workspaces.md
heisen-li: Ok, about this part I will first submit a preliminary idea for discussion in issue.

src/doc/src/reference/workspaces.md
heisen-li: Since the main purpose here is to describe the possible default behavior of the root package, it might be more appropriate to put it here. 
You make a reasonable point, so I've added the description of the member in question earlier.

src/doc/src/reference/workspaces.md
weihanglo: I agree with epage. This paragraph is more like an overview. Putting term definition here seems odd. This might be more appropriate if moved to "The `[workspace]` section" before "Root package", such like

> The directory where the workspace's `Cargo.toml` is located then becomes so-called the _workspace root_.

heisen-li: Done. Thank you.

src/doc/src/reference/workspaces.md
```diff
@@ -33,21 +33,24 @@ In the `Cargo.toml`, the `[workspace]` table supports the following sections:
...
+
 ```toml
 [workspace]
 # ...
 ```
 
+The directory where the workspace's Cargo.toml is located then becomes so-called the workspace root.
+
 At minimum, a workspace has to have a member, either with a root package or as
 a virtual manifest.
```
epage, line 45: This is splitting up the text.  One way to read this is we are describing the minimum of what is needed in a workspace
1. A `[workspace]` table
2. A member

and this inserts the workspace root definition inbetween those

heisen-li: If it doesn't harmonize, it's still the previous one that makes more sense and flows better.

heisen-li:
Sorry. Are there any other changes suggested?

src/doc/src/reference/workspaces.md
```diff
@@ -57,6 +59,14 @@ version = "0.1.0"    # the current version, obeying semver
 authors = ["Alice <a@example.com>", "Bob <b@example.com>"]
 ```
 
+In the [worksapce section](#the-workspace-section), you can specify the members of 
+the workspace by [workspace.members filed](#the-members-and-exclude-fields). 
+
+If no other members are specified in [workspace.members filed](#the-members-and-exclude-fields) 
+in `Cargo.toml` in the workspace root directory, the root package will 
+become the default workspace member. In addition, you can also specify the default 
+workspace members through [workspace.default-members](#the-default-members-field).
```
weihanglo, line 68: Most of this addition seems a bit redundant, since some points are already covered in other parts of this doc. Perhaps we could do something like:

```suggestion
Unless explicitly set through [`workspace.default-members`](#the-default-members-field),
the root package is considered as the only default member of the workspace.
```

src/doc/src/reference/workspaces.md
```diff
@@ -80,6 +90,10 @@ edition = "2021"     # the edition, will have no effect on a resolver used in th
 authors = ["Alice <a@example.com>", "Bob <b@example.com>"]
 ```
 
+In a virtual workspace, at least one member needed be included and if
```
weihanglo, line 93:
```suggestion
Unless explicitly set through [`workspace.default-members`](#the-default-members-field),
every member of a virtual workspace is considered as a default member.
```

src/doc/src/reference/workspaces.md
```diff
@@ -48,6 +49,7 @@ If the [`[workspace]` section](#the-workspace-section) is added to a
 the *root package* of the workspace. The *workspace root* is the directory
 where the workspace's `Cargo.toml` is located.
 
+Example:
```
weihanglo, line 52: I don't see this too relevant. Could we delete it?

src/doc/src/reference/workspaces.md
```diff
@@ -33,6 +33,7 @@ In the `Cargo.toml`, the `[workspace]` table supports the following sections:
 ## The `[workspace]` section
 
 To create a workspace, you add the `[workspace]` table to a `Cargo.toml`:
+
```
weihanglo, line 36: This is not necessary. If you'd like to format the content, please do it in a separate commit.

weihanglo (commented):
Thanks for the doc update.
Could you rework on Git commits in a reasonable manner? Some intermediate commits like “merge master” shouldn't be here.

epage:
@heisen-li I talked the challenge of writing this section with @weihanglo and had an idea on how to cover this topic.  What are your thoughts on #13794?

heisen-li:
> @heisen-li I talked the challenge of writing this section with @weihanglo and had an idea on how to cover this topic. What are your thoughts on #13794?

I think it's fine, no other suggestions. I'll update the #13766 parser section a bit.

PR #13727, opened 2024-04-09
refactor: make `resolve_with_previous` clearer

weihanglo:

Added more comments and extract patch registration to a function from `resolve_with_previous`, so we'll have both `register_previous_lock` and `register_patch_entries` looking symmetrical.

There should have no behavior change.

The construction of `pre_patch_keep` closure is moved into `cargo update` module, since it is the only place using the closure, and I can't foresee anything will use it.

### DIFF
--- a/src/cargo/ops/resolve.rs
+++ b/src/cargo/ops/resolve.rs
@@ -307,115 +307,11 @@ pub fn resolve_with_previous<'gctx>(
         version_prefs.max_rust_version(ws.rust_version().cloned().map(RustVersion::into_partial));
     }
 
-    // This is a set of PackageIds of `[patch]` entries, and some related locked PackageIds, for
-    // which locking should be avoided (but which will be preferred when searching dependencies,
-    // via prefer_patch_deps below)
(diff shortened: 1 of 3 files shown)


src/cargo/ops/resolve.rs
```diff
@@ -308,12 +308,13 @@ pub fn resolve_with_previous<'gctx>(
...
     let avoid_patch_ids = if register_patches {
-        register_patch_entries(registry, ws, previous, &mut version_prefs, pre_patch_keep)?
+        register_patch_entries(registry, ws, previous, &mut version_prefs, keep)?
     } else {
         HashSet::new()
     };
 
-    let keep = |p: &PackageId| pre_patch_keep(p) && !avoid_patch_ids.contains(p);
+    // Refine `keep` with patches that should avoid locking.
+    let keep = |p: &PackageId| keep(p) && !avoid_patch_ids.contains(p);
```
epage, line 317: While I'm normally a fan of shadowing, I feel like decorating a closure with another closure is likely going to make this easy to miss which could make reading the code confusing

PR #13728, opened 2024-04-09
fix(cargo-fix): dont apply same suggestion twice

weihanglo:

This assumes that if any of the machine applicable fixes (`solution`)
in a diagnostic is a duplicate, then `cargo fix` should only apply
the entire suggestion once.

The first commit demonstrates the case in rust-lang/rust#123304.

One caveat is that  `cargo fix` would show `Fixed .. (2 fixes)` for duplicate suggestions,
but actually it only applies the fix once and for all.

### Additional information

r? ehuss


src/cargo/ops/fix.rs
```diff
@@ -738,7 +738,18 @@ fn rustfix_and_fix(
...
+        let mut already_applied = HashSet::new();
         for suggestion in suggestions.iter().rev() {
+            // Skip seemly duplicated suggestions.
+            if suggestion
+                .solutions
+                .iter()
+                .any(|sol| !already_applied.insert(sol))
+            {
+                // already applied
+                fixed_file.fixes_applied += 1;
```
epage, line 750: Why are we incrementing this if nothing happened? It seems weird in the test to say two fixes were applied when really only one change was made.

tests/testsuite/fix.rs
```diff
@@ -1897,3 +1897,38 @@ warning: `foo` (lib) generated 1 warning (run `cargo fix --lib -p foo` to apply
...
+        .with_stderr(
+            "\
+[CHECKING] foo v0.0.1 ([CWD])
+[FIXED] src/lib.rs (2 fixes)
+[FINISHED] `dev` profile [..]
+",
+        )
+        .run();
+
+    assert!(!p.read_file("src/lib.rs").contains("unsafe { unsafe"));
```
epage, line 1933: This seems like a brittle assert. If the fix was formatted just slightly differently, it won't fire.

src/cargo/ops/fix.rs
weihanglo: I was wondering if we match the number of warnings Cargo show, but after seeing the actual message Cargo emitted, it seems better not to increment.

```
warning: `foo` (lib) generated 2 warnings (run `cargo fix --lib -p foo` to apply 2 suggestions)
```

weihanglo: Actually, I am not so sure…

tests/testsuite/fix.rs
```diff
@@ -1897,3 +1897,38 @@ warning: `foo` (lib) generated 1 warning (run `cargo fix --lib -p foo` to apply
 ")
         .run();
 }
+
+#[cargo_test]
+fn fix_only_once_for_duplicates() {
```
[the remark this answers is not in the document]
weihanglo: Good catch. Comment added. The `FIXED src/lib.rs (1 fix)` status line should help catch it to some extent as well

tests/testsuite/fix.rs
weihanglo: Updated to find occurrence of `unsafe` keyword.

src/cargo/ops/fix.rs
weihanglo: Anyway, I think it should reflect the truth at this moment. Updated.

tests/testsuite/fix.rs
```diff
@@ -1897,3 +1897,45 @@ warning: `foo` (lib) generated 1 warning (run `cargo fix --lib -p foo` to apply
...
+[FIXED] src/lib.rs (1 fix)
+[FINISHED] `dev` profile [..]
+",
+        )
+        .run();
+
+    assert_eq!(
+        p.read_file("src/lib.rs").matches("unsafe").count(),
+        4,
+        "2 in lint name, 1 from original unsafe fn, 1 from newly-applied unsafe blocks"
```
epage, line 1939: Can the message include the adjusted `src/lib.rs`?  That would removes one step in the debugging process

crates/rustfix/src/lib.rs
```diff
@@ -253,8 +253,19 @@ impl CodeFix {
 
 /// Applies multiple `suggestions` to the given `code`.
 pub fn apply_suggestions(code: &str, suggestions: &[Suggestion]) -> Result<String, Error> {
+    let mut already_applied = HashSet::new();
     let mut fix = CodeFix::new(code);
     for suggestion in suggestions.iter().rev() {
+        // This assumes that if any of the machine applicable fixes in
```
weihanglo, line 259: Decided to duplicate the work in also `rustfix`. I didn't do it in `CodeFix` as it might require a longer lifetime bound than it was.

PR #13729, opened 2024-04-10
fix(package): Normalize paths in `Cargo.toml`

epage:

On the surface, this resolves problems that aren't too big of a deal
- Clean up redundant information in paths (e.g. `.////.//foo.rs` being `foo.rs`) which is just visual
- Switch paths with `\` in them to `/`

However, this is prep for #13713 where these will be a much bigger deal
- If the user does `./foo.rs`, we might fail to compare that with the list of files included in the package
- We'll generate paths with `\` and need to make sure the packages can still be used on Windows


ChrisDenton:
Would it be worth asserting that `std::path::is_separator(`/`)` is true for the platform? Mind you, off the top of my head, I can't think of any host platforms where that would be false.

epage:
> Would it be worth asserting that std::path::is_separator('/') is true for the platform? Mind you, off the top of my head, I can't think of any host platforms where that would be false.

I had wondered about something along those lines but then almost everything else will already be broken in cargo that I shrugged my shoulders.  For example, every dependency you try to use will have `/` in the paths.

tests/testsuite/package.rs
```diff
@@ -3638,30 +3638,30 @@ edition = "2015"
...
 description = "foo"
 documentation = "docs.rs/foo"
-readme = 'docs/README.md'
-license-file = 'docs/LICENSE'
+readme = "docs/README.md"
+license-file = "docs/LICENSE"
 
 [lib]
-path = 'src/lib.rs'
+path = "src/lib.rs"
```
epage, line 3648: cargo was outputting `path = 'src\lib.rs'` but cargo-test-support converts `\` to `/`.

So the only hint that `\` was present is the use of single-quotes because `toml` prefers to avoid escaping by emitting string literals.

Muscraft (approved):
Thanks!

Issue #13546, opened 2024-03-05
i686-unknown-linux-gnu/bin/cargo: error while loading shared libraries: libatomic.so.1


messense:
Maintainer's note: See <https://github.com/rust-lang/cargo/issues/13546#issuecomment-2463102513> for why Cargo is stuck in OpenSSL v1.

---

When running `cargo` in a `quay.io/pypa/manylinux2014_i686:latest` Docker container, it gives the following error

> /root/.rustup/toolchains/nightly-i686-unknown-linux-gnu/bin/cargo: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory

Is `libatomic.so.1` a hard requirement for cargo now?

### Meta

`rustc --version --verbose`:
```
rustc 1.78.0-nightly (d18480b84 2024-03-04)
```

weihanglo:
Probably this <https://github.com/openssl/openssl/issues/23376>?

weihanglo:
https://github.com/sfackler/rust-openssl/issues/2163 and I mentioned it earlier in <https://github.com/rust-lang/cargo/pull/13449#issuecomment-1948572149>
(I forgot everything older than three days ago)

messense:
This seems to happen again in `rust version 1.79.0-nightly (8b2459c1f 2024-04-09)`

```
 info: downloading installer
  Warning: Not enforcing strong cipher suites for TLS, this is potentially less secure
  info: profile set to 'minimal'
  info: default host triple is i686-unknown-linux-gnu
  info: syncing channel updates for 'nightly-i686-unknown-linux-gnu'
  info: latest update on 2024-04-10, rust version 1.79.0-nightly (8b2459c1f 2024-04-09)
  info: downloading component 'cargo'
  info: downloading component 'rust-std'
  info: downloading component 'rustc'
  info: installing component 'cargo'
  info: installing component 'rust-std'
  info: installing component 'rustc'
  
  info: default toolchain set to 'nightly-i686-unknown-linux-gnu'
    nightly-i686-unknown-linux-gnu installed - rustc 1.79.0-nightly (8b2459c1f 2024-04-09)
  
  
  Rust is installed now. Great!
  
  To get started you may need to restart your current shell.
  This would reload your PATH environment variable to include
  Cargo's bin directory ($HOME/.cargo/bin).
  
  To configure your current shell, you need to source
  the corresponding env file under $HOME/.cargo.
  
  This is usually done by running one of the following (note the leading DOT):
[...]

PR #13731, opened 2024-04-10
chore: downgrade to openssl v1.1.1 (again)

weihanglo:
Accidentally updated by <https://github.com/rust-lang/cargo/pull/13674>

See https://github.com/rust-lang/cargo/issues/13546#issuecomment-2047366361 

I am not sure if we should pin this. Somebody might want to build Cargo with openssl v3.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -78,6 +78,8 @@
       matchUpdateTypes: [
         'patch',
       ],
+      // See rust-lang/cargo#13546 and openssl/openssl#23376 for the exclusion
+      excludePackageNames: ['openssl', 'openssl-src', 'openssl-sys'],
       automerge: false,
       groupName: 'compatible',
     },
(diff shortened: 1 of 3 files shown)


ehuss:
Why did #13674 update it if it wasn't listed as updated in the description?

I would lean towards pinning it, since it has been an issue for a while. If some distro wants to use a newer version, they can patch `Cargo.toml` to remove the pin. Just make sure there is a comment explaining why it is pinned. Or if there is some other way to force dependabot to not touch it.

.github/renovate.json5
```diff
@@ -78,6 +78,8 @@
       matchUpdateTypes: [
         'patch',
       ],
+      // See rust-lang/cargo#13546 and openssl/openssl#23376 for the exclusion
```
weihanglo, line 81: End up using [`excludePackageNames`](https://docs.renovatebot.com/configuration-options/#excludepackagenames), though I completely don't know if it can detect lockfile-only updates.

weihanglo: @epage any chance you know this is gonna work?

epage: RenovateBot will no try to upgrade us but if something else causes an upgrade, then it will still happen.

Granted, if they are running `cargo update <dep>` then nothing should update it.

weihanglo: I don't know why they got upgraded by renovatebot. There was nothing preventing me from `cargo update openssl --precise <oldversion>`. Weird.

PR #13738, opened 2024-04-11
feat(reslve): Respect '--ignore-rust-version'

epage:

This is a part of #9930.

I had considered several ways of implementing this.  I first looked at passing this into `ops::resolve*`.
.This would get a bit annoying with the function signature, so I considered moving it to a builder..
Each of the entry points is slightly different with different ownership needs, making it hard to have a common abstraction.
In doing this, I noticed we currently pass some state around to the resolver via `Workspace`, so I mirrored that.

The nice thing about this location is it provides a good place to hook in config and `package.resolve` so they affect this.

### DIFF
--- a/src/cargo/core/workspace.rs
+++ b/src/cargo/core/workspace.rs
@@ -771,7 +771,7 @@ impl<'gctx> Workspace<'gctx> {
             }
         }
 
-        debug!("find_members - {}", manifest_path.display());
+        debug!("find_path_deps - {}", manifest_path.display());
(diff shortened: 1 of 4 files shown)


Muscraft (approved):
Thanks!

Muscraft:
I think the failure is similar to the one I described in [#13740](https://github.com/rust-lang/cargo/pull/13740#issuecomment-2050565325)

weihanglo (approved):
Scott beat me by speed

PR #13739, opened 2024-04-11
test: deal with lockfile checksums dynamically

cuviper:
When generating local registries in the testsuite, the specific package
checksums may vary from the original simply due to different `zlib`
versions. For example, Fedora 40 is now using `zlib-ng-compat`. We can
account for this dynamically by either formatting the checksum in the
expected output string, or using `generate-lockfile` on projects and
their expected output.


tests/testsuite/cargo_remove/update_lock_file/mod.rs
```diff
@@ -33,5 +35,7 @@ fn case() {
         .stdout_matches(str![""])
         .stderr_matches(file!["stderr.term.svg"]);
 
-    assert_ui().subset_matches(current_dir!().join("out"), &project_root);
+    let expected = Project::from_expected(current_dir!().join("out"));
```
weihanglo, line 38: The dynamic check seems fine to me, though I am a bit unsure since most UI tests in Cargo only asserts one command invocation.

@epage, do you have any input for this?

cuviper: Note that it's not *just* about expected output in the tests. In `cargo_add::locked_unchanged`, cargo itself complains that the input lockfile looks corrupt with its mismatched checksum.

cuviper:
Alternatively, I think setting `Compression::none()` in the `cargo-test-support` registry will sufficiently avoid zlib differences. At least, it appears to be consistent at first glance, but I'll do more testing and send a new PR if that works out.

cuviper:
Closing in favor of #13744.

PR #13740, opened 2024-04-11
refactor: Remove `rust_2024_compatibility` lint group

Muscraft:
The `rust_2024_compatibility` lint group was added as a way to be compatible with `Rust`. This group is meant to be used when switching to the 2024 edition (usually enabled by `cargo fix --edition`). Since we are not going to be interacting with `cargo fix` in the standard way to fix edition lints, this group is not needed at this time. Removing this will (slightly) reduce the complexity of working on things for the 2024 edition.


Muscraft:
Not sure how bors failed, as the CI run for the PR did not fail on the same test. Must've been an odd CI bug.

PR #13742, opened 2024-04-11
feat(cli): Add --ignore-rust-version to update/generate-lockfile

epage:
### What does this PR try to resolve?

This is part of #9930 and extends `--ignore-rust-version` to `cargo update` and `cargo generate-lockfile`

### How should we test and review this PR?

First commit sets up tests

PR #13743, opened 2024-04-11
feat(resolve): Fallback to 'rustc -V' for MSRV resolving

epage:

This is part of #9930 and adds a fallback if the rust-version isn't set

Tests are added in a separate commit so people can see how the behavior changed.

### DIFF
--- a/src/cargo/ops/resolve.rs
+++ b/src/cargo/ops/resolve.rs
@@ -73,7 +73,6 @@ use crate::util::cache_lock::CacheLockMode;
 use crate::util::errors::CargoResult;
 use crate::util::CanonicalUrl;
 use anyhow::Context as _;
-use cargo_util_schemas::manifest::RustVersion;
 use std::collections::{HashMap, HashSet};
 use tracing::{debug, trace};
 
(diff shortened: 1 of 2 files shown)


epage:
Blocked on #13742

src/cargo/ops/resolve.rs
```diff
@@ -304,8 +303,14 @@ pub fn resolve_with_previous<'gctx>(
         version_prefs.version_ordering(VersionOrdering::MinimumVersionsFirst)
     }
     if ws.resolve_honors_rust_version() {
-        let rust_version = ws.rust_version().cloned().map(RustVersion::into_partial);
-        version_prefs.max_rust_version(rust_version);
+        let rust_version = if let Some(ver) = ws.rust_version() {
+            ver.clone().into_partial()
+        } else {
+            let rustc = ws.gctx().load_global_rustc(None)?;
```
weihanglo, line 309: Should we respect `.rustc_info.json`? There are fingerprint detection so it's pretty much safe if rustc has switched to other channel.

epage: That is done automatically through `load_global_rustc`

weihanglo: I didn't get it. It seems that rustc info cache is disabled when no workspace provided. 

https://github.com/rust-lang/cargo/blob/7e31f62a80260c70b0e7604c615ed06fcf98928c/src/cargo/util/rustc.rs#L236-L246

weihanglo (approved):
Thanks!

PR #13744, opened 2024-04-11
test: don't compress test registry crates

cuviper:
They are still nominally gzipped, but using `Compression::none()` makes
them consistent even across zlib and zlib-ng, and this fixes checksum
differences in the testsuite. There is a one-time update of all those
checksums to catch up with this change though.

r? @weihanglo


weihanglo (approved):
Pretty clever! Thank you!

weihanglo:
Tried turning `zlib-ng` feature on, it works flawlessly on Linux for me.

@bors r+

Issue #13629, opened 2024-03-22
Next step for deprecations in Cargo


weihanglo:
## Problem

Cargo has been emitting a bunch deprecation message for a (long?) while. Since edition 2024 is approaching (1.82.0 anticipated), and a Cargo linting system is on the way (<https://github.com/rust-lang/cargo/pull/13621>), we might want to take advantage of this timing to evaluate whether this deprecations could be turned into a hard error.

The main risk here is CI automation might fail. Unlike rustc and clippy lints, people hardly noticed Cargo warnings becuase because there was no way to set `-D` on them.

## Deprecations

<sup>last updated: 2024-03-22T20:37:03.809Z</sup>

#### user-defined alias is shadowing an external subcommand found:

* Warn since: 1.58.0 (2022-01-13)
* Risk: Custom workflow might fail. `cargo clippy` and `cargo fmt` are what we concerned.
* <https://github.com/rust-lang/cargo/pull/10082>
* [Link to code](https://github.com/rust-lang/cargo/blob/f0ae76519127d16d87362c6d88416d92cba86c6a/src/bin/cargo/cli.rs#L321)
* Next step: 

#### `license-file` and `readme` pointing to a non-exsting file

* Warn since: 1.43.0 (2020-04-23) for `license-file`; 1.71.0 (2023-08-03) for `readme`
* Risk: Publish might fail. Easy to fix in normal cases.
* <https://github.com/rust-lang/cargo/pull/7905>
* <https://github.com/rust-lang/cargo/pull/12036>
* [Link to code](https://github.com/rust-lang/cargo/blob/f0ae76519127d16d87362c6d88416d92cba86c6a/src/cargo/ops/cargo_package.rs#L418)
* Next step: 

#### good old `[project]` (replaced with `[package]`)

* Warn since: 1.66 (2022-12-15)
* Risk: Old crates fail to compile. Need an edition boundary.
* <https://github.com/rust-lang/cargo/pull/11135>
* [Link to code](https://github.com/rust-lang/cargo/blob/f0ae76519127d16d87362c6d88416d92cba86c6a/src/cargo/util/toml/mod.rs#L498)
* Next step: 

#### `default-features` in inherited dependencies

* Warn since: 1.69.0 (2023-04-29)
* Risk: Build failures. Easy to fix.
* <https://github.com/rust-lang/cargo/pull/11409>
[...]

PR #11135, opened 2022-09-23
Reduce references to `[project]` within cargo

Muscraft:
There was an issue (#11129) with `[project]` being put in the docs but nothing else stating what `[project]` is. This was an oversight by me when writing the docs as `[project]` is still supported in cargo. [Back in 2014](https://github.com/rust-lang/cargo/commit/86b2a2a432a73b3747cca3593850293a5fc3951a) `[project]` was [renamed to `[package]`](https://github.com/rust-lang/cargo/issues/3388#issuecomment-266504735), since then it has kinda sat as something that exists and is supported but not really. I brought this up in a [zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/removal.20of.20.60.5Bproject.5D.60), and it was suggested that we could show a warning when `[project]` is used in a manifest of a _top-level_ crate. 

To go slightly further than this I tried to change as many references from `[project]` to `[package]` as possible. This should hopefully help with confusion between `[project]` and `[package]` in the future.

This PR also includes a cherry-picked commit from #11131 as it was having issues with bors and fits well with the changes that are being made in this PR.

This should ideally be reviewed commit by commit

PR #13747, opened 2024-04-12
fix(toml): Error on `[project]` in Edition 2024

epage:

`[package]` was added in 86b2a2a432a73b3747cca3593850293a5fc3951a in the pre-1.0 days but `[project]` was never removed nor did we warn on its use until recently in #11135.  So likely we can't remove it completely but we can remove it in Edition 2024+.

This includes `cargo fix` support which is hard coded directly into the `cargo fix` command.

This is part of
- #13629
- rust-lang/rust#123754

While we haven't signed off on everything in #13629, I figured this (and a couple other changes) are not very controversial.

Per commit.  Tests are added to show the behavior changes, including in `cargo fix`.

### DIFF
--- a/src/bin/cargo/commands/fix.rs
+++ b/src/bin/cargo/commands/fix.rs
@@ -1,5 +1,6 @@
 use crate::command_prelude::*;
 
+use cargo::core::Workspace;
 use cargo::ops;
 
 pub fn cli() -> Command {
@@ -69,7 +70,8 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
 
     // Unlike other commands default `cargo fix` to all targets to fix as much
     // code as we can.
-    let ws = args.workspace(gctx)?;
+    let root_manifest = args.root_manifest(gctx)?;
+    let ws = Workspace::new(&root_manifest, gctx)?;
     let mut opts = args.compile_options(gctx, mode, Some(&ws), ProfileChecking::LegacyTestOnly)?;
 
     if !opts.filter.is_specific() {
--- a/src/bin/cargo/commands/fix.rs
+++ b/src/bin/cargo/commands/fix.rs
@@ -60,7 +60,6 @@ pub fn cli() -> Command {
 }
 
 pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
-    let ws = args.workspace(gctx)?;
     // This is a legacy behavior that causes `cargo fix` to pass `--test`.
     let test = matches!(
         args.get_one::<String>("profile").map(String::as_str),
@@ -70,6 +69,7 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
 
     // Unlike other commands default `cargo fix` to all targets to fix as much
     // code as we can.
+    let ws = args.workspace(gctx)?;
     let mut opts = args.compile_options(gctx, mode, Some(&ws), ProfileChecking::LegacyTestOnly)?;
 
     if !opts.filter.is_specific() {
--- a/src/bin/cargo/commands/fix.rs
+++ b/src/bin/cargo/commands/fix.rs
@@ -71,7 +71,8 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
     // Unlike other commands default `cargo fix` to all targets to fix as much
     // code as we can.
     let root_manifest = args.root_manifest(gctx)?;
-    let ws = Workspace::new(&root_manifest, gctx)?;
+    let mut ws = Workspace::new(&root_manifest, gctx)?;
+    ws.set_honor_rust_version(args.honor_rust_version());
     let mut opts = args.compile_options(gctx, mode, Some(&ws), ProfileChecking::LegacyTestOnly)?;
 
     if !opts.filter.is_specific() {
@@ -80,7 +81,9 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
     }
 
     ops::fix(
+        gctx,
         &ws,
+        &root_manifest,
         &mut ops::FixOptions {
             edition: args.flag("edition"),
             idioms: args.flag("edition-idioms"),
--- a/src/cargo/ops/fix.rs
+++ b/src/cargo/ops/fix.rs
@@ -87,11 +87,19 @@ pub struct FixOptions {
     pub broken_code: bool,
 }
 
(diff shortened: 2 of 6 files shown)


tests/testsuite/check.rs
```diff
@@ -1005,6 +1005,37 @@ fn warn_manifest_with_project() {
         .run();
 }
 
+#[cargo_test(nightly, reason = "edition2024")]
```
weihanglo, line 1008: Nightly gate is for rustc. I don't think we need the gate here.

```suggestion
```

src/bin/cargo/commands/fix.rs
```diff
@@ -69,7 +70,8 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
 
     // Unlike other commands default `cargo fix` to all targets to fix as much
     // code as we can.
-    let ws = args.workspace(gctx)?;
+    let root_manifest = args.root_manifest(gctx)?;
```
weihanglo, line 73: What about `set_honor_rust_version`?That is not relevant for now though.

tests/testsuite/fix.rs
```diff
@@ -1942,3 +1942,107 @@ fn fix_only_once_for_duplicates() {
...
+#[cargo_test(nightly, reason = "edition2024")]
+fn migrate_project_to_package() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+cargo-features = ["edition2024"]
+
+# Before project
+[ project ] # After project header
```
weihanglo, line 1963: why do we need extra spaces in-between?

tests/testsuite/check.rs
weihanglo: Ignore this sugggestion if you want to keep edition tests for consistency reason.

tests/testsuite/fix.rs
epage: I put in as much non-standard formatting as I could think of to highlight that we are format-preserving

tests/testsuite/check.rs
epage: Honestly, I worry about that.

I had assumed we pass the new edition down but we only pass the lint group.  Its a bit odd that we have the lint group working on stable.

weihanglo (approved):

weihanglo:
This only errors out on edition 2024, though for hard errors it's better to get a consensus from the team.

@rfcbot fcp merge

weihanglo:
The cargo team went to the proposed deprecations in <https://github.com/rust-lang/cargo/issues/13629> today, and we are fine with this erroring on Edition 2024. No need to wait for a full FCP.

@bors r+

src/cargo/ops/fix.rs
```diff
@@ -87,11 +88,26 @@ pub struct FixOptions {
...
+    original_ws: &Workspace<'_>,
+    root_manifest: &Path,
+    opts: &mut FixOptions,
+) -> CargoResult<()> {
+    check_version_control(gctx, opts)?;
+
     if opts.edition {
-        check_resolver_change(ws, opts)?;
+        let specs = opts.compile_opts.spec.to_package_id_specs(&original_ws)?;
+        let members: Vec<&Package> = original_ws
```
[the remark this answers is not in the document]
epage: We do filter by the `--package` flag which is us just doing what the user told us.  I'm not seeing how presence of `[proiject]` causes a problem.  We also have a test showing that it works.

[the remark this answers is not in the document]
heisen-li: Sorry, I don't seem to have been clear.

What I meant was that if the members in the workspace use `[project]` instead of `[package]`, then executing the command `cargo fix --edition` in the workspace does not modify the members.

I'm wondering if the changes should be done on all members at once.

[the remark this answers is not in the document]
epage: I think I'm still missing the problem case.

Could you create a test case to demonstrate it?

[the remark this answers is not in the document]
heisen-li: If there is a workspace `ws` containing the member `hello`.

which:
```
# [ws]/Cargo.toml
[workspace]
members = ["hello"]

[project]     # Pay attention here 
name = "ws"
version = "0.1.0"
edition = "2021"
```

```
# [ws]/hello/Cargo.toml
[project]     # Pay attention here 
name = "hello"
version = "0.1.0"
edition = "2021"
```

After running `cargo fix --edition --allow-no-vcs` in the workspace. The `cargo.toml` of `ws` completes the modification, but the `cargo.toml` of member `hello` does not.

I think both should be modified to `[package]` here.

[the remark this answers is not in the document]
epage: I've written that up as a test case which is passing.

Yes, we aren't migrating `hello`'s `Cargo.toml` but we also aren't migrating `hello`s `src/lib.rs`.  Manifest and rust migration should be done hand-in-hand.  The user has to opt-in to migrating `hello` with the `--workspace` flag.  `cargo fix --edition --workspace` will migrate both packages, manifest and rust.

<details>

```rust
#[cargo_test]
fn migrate_project_to_package_ws() {
    let p = project()
        .file(
            "Cargo.toml",
            r#"

[workspace]
members = ["hello"]

# Before project
[ project ] # After project header
# After project header line
name = "foo"
edition = "2021"
# After project table
"#,
        )
        .file("src/lib.rs", "")
        .file(
            "hello/Cargo.toml",
            r#"

# Before project
[ project ] # After project header
# After project header line
name = "hello"
edition = "2021"
# After project table
"#,
        )
        .file("hello/src/lib.rs", "")
        .build();

    p.cargo("fix --edition --allow-no-vcs")
        .masquerade_as_nightly_cargo(&["edition2024"])
        .with_stderr(
            "\
[MIGRATING] Cargo.toml from 2021 edition to 2024
[FIXED] Cargo.toml (1 fix)
[WARNING] [CWD]/hello/Cargo.toml: `[project]` is deprecated in favor of `[package]`
[CHECKING] foo v0.0.0 ([CWD])
[MIGRATING] src/lib.rs from 2021 edition to 2024
[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [..]s
",
        )
        .run();
    assert_eq!(
        p.read_file("Cargo.toml"),
        r#"

[workspace]
members = ["hello"]

# Before project
[ package ] # After project header
# After project header line
name = "foo"
edition = "2021"
# After project table
"#
    );
    assert_eq!(
        p.read_file("hello/Cargo.toml"),
        r#"

# Before project
[ project ] # After project header
# After project header line
name = "hello"
edition = "2021"
# After project table
"#
    );
}
```

</details>

PR #13748, opened 2024-04-12
refactor(config): Consistently use kebab-case

epage:
This shouldn't change the behavior but makes it safer if
- We add new fields where it will matter
- Copy/paste these for new structs

I did not change things related to the Index because we are already stuck with that case (whether we want it or not)

Came across this when working on #13540 and almost made the mistake of copying what was already there

### DIFF
--- a/src/cargo/ops/cargo_new.rs
+++ b/src/cargo/ops/cargo_new.rs
@@ -131,6 +131,7 @@ impl NewOptions {
 }
 
 #[derive(Deserialize)]
+#[serde(rename_all = "kebab-case")]
 struct CargoNewConfig {
     #[deprecated = "cargo-new no longer supports adding the authors field"]
     #[allow(dead_code)]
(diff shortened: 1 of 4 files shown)


weihanglo:
The third time we hit this failure?

```

---- death::kill_cargo_add_never_corrupts_cargo_toml stdout ----
thread 'death::kill_cargo_add_never_corrupts_cargo_toml' panicked at crates\cargo-test-support\src\paths.rs:150:17:
failed to remove "D:\\a\\cargo\\cargo\\target\\tmp\\cit\\t1104\\foo": Os { code: 32, kind: Uncategorized, message: "The process cannot access the file because it is being used by another process." }
```

weihanglo:
`kill_cargo_add_never_corrupts_cargo_toml` failed also in #13738 and #13740

weihanglo (approved):

PR #13751, opened 2024-04-12
docs(ref): Update unstable docs for msrv-policy

epage:

### DIFF
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -327,8 +327,24 @@ Documentation updates:
 ## msrv-policy
 - [#9930](https://github.com/rust-lang/cargo/issues/9930) (MSRV-aware resolver)
 
-The `msrv-policy` feature enables experiments in MSRV-aware policy for cargo in
-preparation for an upcoming RFC.
+Catch-all unstable feature for MSRV-aware cargo features under
+[RFC 2495](https://github.com/rust-lang/rfcs/pull/2495).
+
+### MSRV-aware cargo add
+
+This was stabilized in 1.79 in [#13608](https://github.com/rust-lang/cargo/pull/13608).
+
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/unstable.md
```diff
@@ -327,8 +327,24 @@ Documentation updates:
...
+### MSRV-aware cargo add
+
+This was stabilized in 1.79 in [#13608](https://github.com/rust-lang/cargo/pull/13608).
+
+### MSRV-aware resolver
+
+By default, `-Zmsrv-policy` enables an MSRV-aware resolver.
+The resolver will prefer dependencies with a `package.rust-version` that is the same or newer as your project's MSRV.
+Your project's MSRV is determined by taking the lowest `package.rust-version` set in you workspace.
+If there is none set, `rustc -V` will be used to pick up the rust version from rustup's `rust-toolchain.toml`.
```
weihanglo, line 342: It might be tricker to explain the rustc invocation when <https://github.com/rust-lang/cargo/pull/13659> is merged. rust-toolchain.toml does not always exist either. I might suggest something like:

```suggestion
Your project's MSRV is determined by taking the lowest `package.rust-version` set among all members of your workspace.
If there is none set, the version of rustc Cargo will invoke would be picked as the project's MSRV.
```

BTW, what does "project" refer to here? 😆

epage: Went in a slightly different direction.

I also did not write this with an eye for what would go in stable documentation.  This is meant to help people in using the unstable feature and I don't think we need to block on this being ideal.

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

weihanglo:
bors is dead?

@bors r-

PR #13753, opened 2024-04-13
Reword sentence describing workspace toml for clarity

jw013:
Judging by the commit history, the original sentence was written before the `[patch]` and `[profile]` sections were added and since those sections do not go underneath the workspace table the original sentence needed to be updated.


weihanglo (approved):
Looks good! Thank you.

PR #13754, opened 2024-04-14
feat(resolve): Tell the user the style of resovle done 

epage:

This is to help with https://github.com/rust-lang/cargo/issues/9930

Example changes:
```diff
-[LOCKING] 4 packages
+[LOCKING] 4 packages to latest compatible version
-[LOCKING] 2 packages
+[LOCKING] 2 packages to latest Rust 1.60.0 compatible versions
-[LOCKING] 2 packages
+[LOCKING] 2 packages to earliest compatible versions
```

Benefits
- The package count is of "added" packages and this makes that more
  logically clear
- This gives users transparency into what is happening, especially with
  - what rust-version is use
  - the transition to this feature in the new edition
  - whether the planned config was applied or not (as I don't want it to
    require an MSRV bump)
- Will make it easier in tests to show what changed
- Provides more motiviation to show this message in `cargo update` and
  `cargo install` (that will be explored in a follow up PR)

This does come at the cost of more verbose output but hopefully not too
verbose.  This is why I left off other factors, like avoid-dev-deps.

### DIFF
--- a/tests/testsuite/cargo_add/default_features/stderr.term.svg
+++ b/tests/testsuite/cargo_add/default_features/stderr.term.svg
@@ -26,7 +26,7 @@
 </tspan>
     <tspan x="10px" y="64px"><tspan class="fg-green bold">      Adding</tspan><tspan> my-package2 v0.4.1 to dependencies</tspan>
 </tspan>
-    <tspan x="10px" y="82px"><tspan class="fg-green bold">     Locking</tspan><tspan> 3 packages</tspan>
+    <tspan x="10px" y="82px"><tspan class="fg-green bold">     Locking</tspan><tspan> 3 packages to latest compatible versions</tspan>
 </tspan>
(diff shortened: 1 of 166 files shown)


tests/testsuite/cargo_add/default_features/stderr.term.svg
weihanglo, line 1: It is a bit confused that `Locking 3 packages to latest version` is followed by a status indicating `(latest: some-version)`. So do we really have my-package2 on the latest version?

We could put something like `latest compatible version` though that is more verbose.  Call out here and see if anyone has a better idea. I am fine that we move forward with the current proposal.

weihanglo (approved):

tests/testsuite/cargo_add/default_features/stderr.term.svg
epage: Adding "compatible" shouldn't be too bad because we'll be showing that word by default in the 2024 edition

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #13756, opened 2024-04-15
Recategorize cargo test's `--doc` flag under "Target Selection"

dtolnay:

In `cargo help test`, the `--doc` flag is listed under a section called "Target Selection" next to `--lib`, `--bin`, `--bins`, `--example`, `--examples`, `--test`, `--tests`, `--bench`, `--benches`, and `--all-targets`.

But in `cargo test --help`, it was instead listed in an "Options" section next to `--no-run`, `--message-format`, `--color`, etc, which seems less appropriate than "Target Selection".

- `cargo build --release`
- `cargo test --release --test testsuite -- cargo_test::help::case`
- `target/release/cargo test --help`
- `target/release/cargo help test` (unchanged)

### DIFF
--- a/src/bin/cargo/commands/test.rs
+++ b/src/bin/cargo/commands/test.rs
@@ -18,7 +18,6 @@ pub fn cli() -> Command {
                 .num_args(0..)
                 .last(true),
         )
-        .arg(flag("doc", "Test only this library's documentation"))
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
Although `doctest` is actually a `CompileMode`, to me that is an implementation detail. Agree that `cargo test --help` should align with `cargo help test`. Thank you!

PR #13759, opened 2024-04-15
feat(update): Include a Locking message

epage:

This extends #13561 to `cargo update`.  I previously left it out because the locking message was redundant.  However the `Locking` message has been extended in #13754 to include the resolving policy which can be a useful point of interest (e.g. "why does `cargo update` do nothing? Oh, `-Zminimal-versions` is enabled").

I still left out the message for `--precise` because the user is overriding all of that.

I'd still like to extend all of this to `cargo install` (and maybe all resolves) but that is taking more investigation.

### DIFF
--- a/src/cargo/ops/cargo_generate_lockfile.rs
+++ b/src/cargo/ops/cargo_generate_lockfile.rs
@@ -358,8 +358,12 @@ fn print_lockfile_updates(
     resolve: &Resolve,
     registry: &mut PackageRegistry<'_>,
 ) -> CargoResult<()> {
+    let diff = PackageDiff::diff(&previous_resolve, &resolve);
+    let num_pkgs: usize = diff.iter().map(|d| d.added.len()).sum();
+    status_locking(ws, num_pkgs)?;
+
     let mut unchanged_behind = 0;
-    for diff in PackageDiff::diff(&previous_resolve, &resolve) {
+    for diff in diff {
         fn format_latest(version: semver::Version) -> String {
             let warn = style::WARN;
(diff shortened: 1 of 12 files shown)


tests/testsuite/git.rs
```diff
@@ -616,7 +616,10 @@ fn recompilation() {
 
     p.cargo("update")
         .with_stderr(&format!(
-            "[UPDATING] git repository `{}`",
+            "\
+[UPDATING] git repository `{}`
+[LOCKING] 0 packages to latest compatible versions
```
weihanglo, line 621: Is zero plural or singular? 🤔

src/cargo/ops/cargo_generate_lockfile.rs
weihanglo, line 1: > which can be a useful point of interest (e.g. "why does cargo update do nothing? Oh, -Zminimal-versions is enabled").

I can't connect the current locking message to this. It basically prints zero without any explanation. How does it help user figure out?

tests/testsuite/git.rs
epage: Its plural

src/cargo/ops/cargo_generate_lockfile.rs
epage: It will print
```
[LOCKING] 0 packages to earliest compatible versions
```

minimal versions, in its current form, is an odd example as its on the command line.  When MSRV becomes the default, it will be "holding things back" and we want to give the users hints as to why it might be.

weihanglo (approved):

weihanglo:
Thanks for the clarification!

@bors r+

PR #13762, opened 2024-04-16
deprecate: Discard the existence of `[project]` and change `WARNING` to `ERROR`

heisen-li:

a part of  #13629

Maybe start this now? It's always good to get rid of something：)


epage:
Couple of notes
- #13747 is also open for this
-  The next steps on that issue are proposals, we still need to check with the team on them (I made a guess with #13747 that our next steps would be approved)
- As a heads up, "deprecating" means to discourage the use of something and that it no longer gets new features, rather than remove support for it.

weihanglo:
Close as #13747 was just merged.

PR #13764, opened 2024-04-16
feat(install): Including Locking message 

epage:

This extends #13561 to include `cargo install`, like #13759 did for `cargo update`.

As we switch to MSRV-aware resolver, this will help users work out why
MSRV-aware resolving isn't helping them.

This will also make it more obvious if we breaking things when
developing the MSRV-aware resolver.

### Additional information

This still leaves `cargo publish` and a couple other misc situations that I'm intentionally avoiding because
- They hit some weird cases that can confuse the user (e.g. causing `cargo install --locked` to show that 1 package is being added) and we can't distinguish these cases too well from where this is happening
- The value is lower

### DIFF
--- a/src/cargo/ops/resolve.rs
+++ b/src/cargo/ops/resolve.rs
@@ -147,6 +147,7 @@ pub fn resolve_ws_with_opts<'gctx>(
             specs,
             add_patches,
         )?;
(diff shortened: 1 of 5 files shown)


ehuss:
Thanks!

@bors r+

PR #12083, opened 2023-05-03
do not try an exponential number of package names

Eh2406:
re #11934, and as discussed in the cargo team meeting, this changes the strategy to "the original, all underscore, and all dashes".

I was excessively proud of the `hyphen_combination_num` based implementation when I came up with it. But it's always been a hack. I'm glad to be the one to remove it.

PR #13765, opened 2024-04-17
Fix cargo add behaving different when translating package name

### LINKED ISSUE #10680: `cargo add` disables previously enabled features when converting between `_` and `-`
### Problem

`cargo add` can be used to add an optional feature to an existing dependency:

```sh
> cargo add once_cell -F unstable
    Updating crates.io index
      Adding once_cell v1.11.0 to dependencies.
             Features:
             + alloc
             + race
             + std
             + unstable
             - atomic-polyfill
             - parking_lot
             - parking_lot_core

> cargo add once_cell -F parking_lot
    Updating crates.io index
      Adding once_cell v1.11.0 to dependencies.
             Features:
             + alloc
             + parking_lot
             + parking_lot_core
             + race
             + std
             + unstable
             - atomic-polyfill
# note that `unstable` is still enabled from before!
```

Furthermore, `cargo add` can convert `_` to `-` and vice versa in crate names. However, **when doing this translation, existing features are dropped**:

```sh
> cargo add once-cell -F parking_lot
    Updating crates.io index
warning: translating `once-cell` to `once_cell`
      Adding once_cell v1.11.0 to dependencies.
             Features:
             + alloc
             + parking_lot
             + parking_lot_core
             + race
             + std
             - atomic-polyfill
             - unstable
# note that `unstable` is no longer enabled!
```

`cargo add` should behave the same whether the crate name was written in its canonical form or not.

### Version

```text
cargo 1.63.0-nightly (a4c1cd0eb 2022-05-18)
release: 1.63.0-nightly
commit-hash: a4c1cd0eb6b18082a7e693f5a665548fe1534be4
commit-date: 2022-05-18
libcurl: 7.80.0-DEV (sys:0.4.51+curl-7.80.0 vendored ssl:OpenSSL/1.1.1n)
os: Manjaro 21.2.6 (Qonos) [64-bit]
```

### LINKED ISSUE #13702: cargo add to package within a virtual workspace behaves differently when translating package name
### Problem

`cargo add` behaves differently when operating on a package in a virtual workspace depending whether the package name is not translated or translated (cargo shows `warning: translating …`). 

In the first case cargo adds a reference to dependency declared in `[workspace.dependencies]` - which I believe is the desired result.

In the later case cargo adds the dependency and the version, but does not reference `[workspace.dependencies]` - I believe this is not desired.

### Steps

Reproduction script when adding `parking-lot`:
```
> mkdir workspace
> cd workspace
> cat > Cargo.toml
[workspace]
resolver = "2"

[workspace.dependencies]
parking_lot = "0.12.1"
> cargo new x
     Created binary (application) `x` package
> cargo add -p x parking-lot
    Updating crates.io index
warning: translating `parking-lot` to `parking_lot`
      Adding parking_lot v0.12.1 to dependencies.
             Features:
             - arc_lock
             - deadlock_detection
             - hardware-lock-elision
             - nightly
             - owning_ref
             - send_guard
             - serde
    Updating crates.io index
> cat x/Cargo.toml
[package]
name = "x"
version = "0.1.0"
edition = "2021"

[dependencies]
parking_lot = "0.12.1"
```

Reproduction script when adding `parking_lot`:
```
> mkdir workspace
> cd workspace
> cat > Cargo.toml
[workspace]
resolver = "2"

[workspace.dependencies]
parking_lot = "0.12.1"
> cargo new x
     Created binary (application) `x` package
> cargo add -p x parking_lot
    Updating crates.io index
      Adding parking_lot (workspace) to dependencies.
             Features as of v0.12.1:
             - arc_lock
             - deadlock_detection
             - hardware-lock-elision
             - nightly
             - owning_ref
             - send_guard
             - serde
    Updating crates.io index
> cat x/Cargo.toml
[package]
name = "x"
version = "0.1.0"
edition = "2021"

[dependencies]
parking_lot.workspace = true
```

### Possible Solution(s)

_No response_

### Notes

I tested this with the latest rust release.

This might be related to this issue: https://github.com/rust-lang/cargo/issues/10680

### Version

```text
> cargo version --verbose
cargo 1.77.1 (e52e36006 2024-03-26)
release: 1.77.1
commit-hash: e52e360061cacbbeac79f7f1215a7a90b6f08442
commit-date: 2024-03-26
os: Ubuntu 20.04 (focal) [64-bit]
```

dohse:
Fixes #13702
Fixes #10680

TODOs

- [x] ~Fuzzy match registry dependencies in `cargo remove` as well~
     `cargo remove` does not need fuzzy matching, because there is no unexpected behavior for the user
- [x] ~Don't duplicate name permutation generation~
     Unsure whether this is worth it
- [ ] Shall we reject cases where multiple different permutations match?
- [x] Add comments that explain the behavior


tests/testsuite/cargo_add/detect_workspace_inherit_fuzzy/mod.rs
```diff
@@ -7,7 +7,6 @@ use cargo_test_support::str;
 use cargo_test_support::Project;
 
 #[cargo_test]
-#[ignore]
```
epage: Please update your PR so the tests are never ignored and the first two commits pass on their own.  THat will make it so this commit will then change their output, showing the change in behavior.

dohse: Ok, I updated the PR.

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -364,24 +364,28 @@ fn resolve_dependency(
     };
     selected_dep = populate_dependency(selected_dep, arg);
 
-    let old_dep = get_existing_dependency(manifest, selected_dep.toml_key(), section)?;
-    let mut dependency = if let Some(mut old_dep) = old_dep.clone() {
-        if old_dep.name != selected_dep.name {
-            // Assuming most existing keys are not relevant when the package changes
-            if selected_dep.optional.is_none() {
-                selected_dep.optional = old_dep.optional;
+    let get_dependency = |mut selected_dep: Dependency| {
```
epage, line 367: > I created a fix for this issue. I did not follow the route of doing fuzzy lookups within cargo add, but just repeat the workspace existing dependency lookup and workspace dependency lookup when the latest dependency lookup returns a different name.

This approach is pretty complicated, requiring the code to duplicate its logic to cycle things back through.  At minimum, it seems like we could look in the registry from the beginning and then pass the literal name given and the registry normalized name around to do lookups.

Though that will also have problems.  This is using crates.io to normalize package names regardless of whether their source is crates.io or not.  We really should be normalizing as we go.  Unlike the registry, we don't need to try a bunch of combinations but we can extract the names from a table, normalize them to either `-` or `_`, normalize our value to `-` or `_` to match and see what the final result would be.

dohse: Sorry, for the late reply.

> This approach is pretty complicated, requiring the code to duplicate its logic to cycle things back through.

I agree that this is not pretty. On the other hand it checks some boxes: It avoids duplicating logic in the code, It is pretty minimal, It avoids querying the registry if it is not needed, It applies the same fuzzy search logic as the registry.

> At minimum, it seems like we could look in the registry from the beginning and then pass the literal name given and the registry normalized name around to do lookups.

This has the problem that we always have to lookup the registry. This breaks a lot of tests by changing their outputs  by adding "Updating crates.io index" lines.

> This is using crates.io to normalize package names regardless of whether their source is crates.io or not.

Mhh, not sure whether I understand this properly. Sources of dependencies can be workspace dependencies, the members of the workspace and crates.io. As far as I understand only crates.io does fuzzy lookups. In which scenario crates.io be used to normalize package names that are not from crates.io?

> We really should be normalizing as we go.

I understand that you want to try several versions of the name - the original one, one that normalizes to underscores and one that normalizes to dashes, right? I can try that.

epage: > Mhh, not sure whether I understand this properly. Sources of dependencies can be workspace dependencies, the members of the workspace and crates.io. As far as I understand only crates.io does fuzzy lookups. In which scenario crates.io be used to normalize package names that are not from crates.io?

git dependencies

> I understand that you want to try several versions of the name - the original one, one that normalizes to underscores and one that normalizes to dashes, right? I can try that.

That should be fine as that is all we are checking for the registry today, see #12083

dohse: I updated this PR to check permutations

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -364,7 +364,24 @@ fn resolve_dependency(
     };
     selected_dep = populate_dependency(selected_dep, arg);
 
-    let old_dep = get_existing_dependency(manifest, selected_dep.toml_key(), section)?;
+    let mut old_dep = get_existing_dependency(manifest, selected_dep.toml_key(), section)?;
+    if old_dep.is_none() && selected_dep.source().is_none() && selected_dep.rename().is_none() {
```
epage, line 368: Why the `selected_dep` checks?

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -364,7 +364,24 @@ fn resolve_dependency(
...
+        for name_permutation in [
+            selected_dep.name.replace('-', "_"),
+            selected_dep.name.replace('_', "-"),
+        ] {
```
epage, line 372: Instead of looking up `old_dep` and then looking it up again, could we include the user-selected name in the list of permutations?

epage, line 372: Should we break this out into a function to be reused between this and `find_workspace_dep`?

src/cargo/ops/cargo_add/mod.rs
dohse: `selected_dep.source().is_none()` to try the mutations only on when looking for registry dependencies and `selected_dep.rename().is_none()` to not try permutations when adding a renamed dependency.

src/cargo/ops/cargo_add/mod.rs
dohse: I will try that, but I guess I have handle the rename case then separately.

dohse: I did this and pushed a new version with that

torhovland:
This PR looks to be basically ready. @dohse, is there anything else you need to do before a final review? I see an unfinished TODO list in the description and one review comment that may not be resolved yet.

dohse:
Yes, this PR should be ready. I need feedback to my responses (@epage). I would take another look at my TODOs when the general direction of the fix is clear.

src/cargo/ops/cargo_add/mod.rs
epage: Looks like you went ahead and did this in 543a138c285cec20f1454af02b4859cd90a03487 without any aspect of rename.  So is that not needed then?

src/cargo/ops/cargo_add/mod.rs
epage: Could you expand on the rename case?

dohse: I changed the implementation - I believe this discussion is not relevant any more.

src/cargo/ops/cargo_add/mod.rs
dohse: I am handling renames in [these lines](https://github.com/rust-lang/cargo/pull/13765/files#diff-7d08d5b8dd5fdeaca6bb19b814d06790d42f63f052b42a41f3eb5cf8ec89456cR463-R465) by not doing fuzzy lookups when the dependency is renamed

dohse:
I responded to your comments - I will also rebase this PR.

> In looking at the responses, its not clear to me what I needed responding to.

Sorry - this PR became confusing, because it contains two possible solutions. I was mostly referring to [this comment](https://github.com/rust-lang/cargo/pull/13765#discussion_r1589338337), that I broke out a function as suggested.

src/cargo/ops/cargo_add/mod.rs
epage: Please squash when you completely change directions in a PR.  It can be confusing to track what changed or didn't change between commits.

epage: So while the code is changed, the concepts from this discussion are still there
```rust
    if let Some(rename) = dependency.rename() {
        return lookup(rename);
    }
```
```rust
            if !matches!(dep.source, Some(Source::Registry(_))) {
                continue;
            }
```

The motivation for these checks is not obvious.  Could you explain these?

Also, if there aren't tests that would break with their removal, could you add them?

dohse: I added comments that explain the conditions.

When I remove the first condition these tests fail:

cargo_add::change_rename_target::case
cargo_add::invalid_key_rename_inherit_dependency::case
cargo_add::overwrite_rename_with_rename_noop::case

so that is covered.

That is not the case for the second condition, so I added a case to the test suite.

dohse:
I hope that I addressed the comment.

[I am also now only check permutations that are unique](https://github.com/rust-lang/cargo/pull/13765/files#diff-7d08d5b8dd5fdeaca6bb19b814d06790d42f63f052b42a41f3eb5cf8ec89456cR471)

tests/testsuite/cargo_add/add_workspace_non_fuzzy/in/target/.rustc_info.json
epage, line 1: Looks like you accidentally pulled in your target directory

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -449,6 +452,45 @@ fn resolve_dependency(
...
+    ]
+    .into_iter()
+    .dedup()
+    {
+        let Some(dep) = lookup(&name_permutation)? else {
+            continue;
+        };
+
+        if dependency.name != name_permutation {
+            // Ignore fuzzy matches with non-registry sources
```
epage, line 478: This comment is a re-phrasing of the code and isn't describing the motivation behind it.

Based on the test, my guess is
```suggestion
            // Mirror the fuzzy matching policy of `PackageRegistry::query_vec`
```

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -449,6 +452,45 @@ fn resolve_dependency(
     Ok(dependency)
 }
 
+fn fuzzy_lookup(
+    dependency: &mut Dependency,
+    lookup: impl Fn(&str) -> CargoResult<Option<Dependency>>,
+    gctx: &GlobalContext,
+) -> CargoResult<Option<Dependency>> {
+    if let Some(rename) = dependency.rename() {
+        // If it has a rename only lookup that name - analogue to Dependency::toml_key()
```
epage, line 461: This comment is a re-phrasing of the code and isn't describing the motivation behind it.

Oh, I think I see the key reason for this. Before we did lookups with `toml_key`.  This manually implements `toml_key` so that fuzzy lookups only happen on package names, not dependency names, to scope this to only happen when other normalization happens (`PackageRegistr::query_vec`).  Is that it?

tests/testsuite/cargo_add/add_workspace_non_fuzzy/in/target/.rustc_info.json
dohse: Sorry - removed it

src/cargo/ops/cargo_add/mod.rs
dohse: Sounds good.

I believe referencing `PackageRegistry::query()` is more clear, because `query_vec()` is only defined on the trait.

src/cargo/ops/cargo_add/mod.rs
dohse: The intention is to mirror the behavior of `Dependency::toml_key()` to lookup _only_ on the rename if that is present. I agree with your description that fuzzy lookups should only happen on packages names.

I will write `Mirror behavior of to Dependency::toml_key()`

dohse:
I changed the comments according to the discussion.

src/cargo/ops/cargo_add/mod.rs
epage: That describes the whole function in terms of checking `rename` and then `name`.

The specific nuance is why you are doing that, rather than using `toml_key`.

Something like
```
Manually implement `toml_key` to restrict fuzzy lookups to only package names to mirror `PackageRegistry::query_vec`
```

dohse: I used the suggested comment.

dohse:
I updated the comment.

I also removed the `.dedup()` of the combinations - it would not prevent all lookups of the unmodified original basename.

epage:
Thanks!

@bors r+

Issue #8365, opened 2020-06-16
Add `cargo new --workspace`


camelid:
**Describe the problem you are trying to solve**

One has to manually create a new directory and setup a `Cargo.toml`, rather than being able to simply run `cargo new foo --workspace` and have a basic template set-up.

**Describe the solution you'd like**

Add the option `--workspace` to `cargo new` that creates a new workspace with the provided name.

**Notes**

The feature is not too important, but it seems like it would streamline the process of creating a new workspace a bit.

pksunkara:
Just wanted to chime in and say that we have been looking to implement this with https://github.com/pksunkara/cargo-workspaces/pull/15. Once that PR is merged, you should be able to do `cargo ws init`.

buzmeg:
Should this also be closed as postponed per #8411 ?

Stargateur:
@buzmeg why is there another issue that ask about this feature ?

H2rmone:
Same confusion here. In Javascript community, tools like lerna, yarn workspace and nx both have command to create workspace project. Cargo workspace includes many feature, I think it will save a lot of beginners' time if we have command to create workspace project like `cargo new --workspace`.

dobromyslov:
Are there any plans to implement this feature? Looks like the `templates` feature was postponed last year.

drebbe-intrepid:
Any update with this? I would love this feature also.

epage:
I've clarified the title and linked out to #6378 since I always get these mixed up.

As for plans to move forward with this, someone would need to step up and work on a design for this which at least acknowledges the needs of #6378 so one doesn't block the other.

My gut feeling is there are enough use cases with various needs that we'd be better served by research into #5151.

rosti-il:
I think it's worth to support the `--workspace` option not only in `cargo new` and `cargo init` but also in `cargo remove`.

epage:
`cargo remove` automatically removes workspace dependencies when they are no longer used.

rosti-il:
@epage you're right. I just got confused between `cargo new` and `cargo add` when thought about the opposite of `cargo remove`.

snubwoody:
Any news on this? I use workspaces all the time and I think this would be a great feature. I would be willing to help if that's possible?

vdudouyt:
I have the same question

solidiquis:
+1

jhpratt:
Any updates will be posted here. Please refrain from asking if there are any updates or similar.

[...]

PR #13766, opened 2024-04-17
[doc]: Add create workspace doc

heisen-li:

The second part of #13580.

There was a discussion in the previous issue and it looks like there are preliminary results?

Reference:
https://github.com/rust-lang/cargo/issues/13580#issuecomment-2026364311
https://github.com/rust-lang/cargo/issues/13580#issuecomment-2052044699

### DIFF
--- a/src/doc/src/SUMMARY.md
+++ b/src/doc/src/SUMMARY.md
@@ -14,6 +14,7 @@
     * [Package Layout](guide/project-layout.md)
     * [Cargo.toml vs Cargo.lock](guide/cargo-toml-vs-cargo-lock.md)
     * [Tests](guide/tests.md)
+    * [Creating a Workspace](guide/creating-a-new-workspace.md)
     * [Continuous Integration](guide/continuous-integration.md)
     * [Cargo Home](guide/cargo-home.md)
     * [Build Cache](guide/build-cache.md)
--- a/src/doc/src/guide/creating-a-new-workspace.md
+++ b/src/doc/src/guide/creating-a-new-workspace.md
@@ -1,0 +1,141 @@
+# Creating a New Workspace
+
+A [workspace][def-workspace] is a collection of one or more packages, 
+called workspace members, that are managed together.
+
+In this chapter, we will create a workspace `new_workspace` containing 
+binary member `foo` and library member `bar`.
+
+As mentioned in [`[workspace]` section][workspace-section], the workspace must 
+have at least one member, either the [root package] or a [virtual manifest].
+
+Next we create a workspace containing [root package].
+For convenience, you can first create a package using the command `cargo new new_workspace`.
+Then add the `[workspace]` section to the `Cargo.toml` file in the root directory 
+to make it a manifest of the workspace:
+
+```toml
+# [new_workspace]/Cargo.toml
+[workspace]
+
+[package]
+name = "new_workspace"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+```
+
+Then, continue adding members `foo` and `bar` to the workspace:
+
+```console
+$ cd new_workspace
+$ cargo new foo
+$ cargo new bar --lib
+```
+
+Cargo will automatically add members to `Cargo.toml`。
+At this point, the workspace will contain three members: `foo` and `bar` and 
+the default member `new_workspace`.
+
+```toml
+# [new_workspace]/Cargo.toml
+[workspace]
+members = [ "bar", "foo" ]
+
+[package]
+name = "new_workspace"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+```
+
+The package at this point contains the following files:
+
+```console
+$ cd new_workspace
+$ tree .
+.
+├── bar
+│   ├── Cargo.toml
+│   └── src
+│       └── lib.rs
+├── Cargo.toml
+├── foo
+│   ├── Cargo.toml
+│   └── src
+│       └── main.rs
+└── src
+    └── main.rs
+
+5 directories, 6 files
+```
+
+Let's move on and create a virtual workspace.
+
+In the another `new_workspace` empty directory, create a new `Cargo.toml` file and 
+add the `[workspace]` section:
+
+```toml
+# [new_workspace]/Cargo.toml
+[workspace]
+```
+
+If using a virtual workspace, then the version of [resolver] needs to be specified 
+in the table (if not, the default version of resolver for a workspace is `1`, 
+even if the default resolver version for workspace members is `2`), for example:
+
+```toml
+# [new_workspace]/Cargo.toml
+[workspace]
+resolver = "2"
+```
+
+Likewise, you can then use the `cargo new <package>` command to create 
+binary member `foo` and library member `bar`.
+
+
+```toml
+# [new_workspace]/Cargo.toml
+[workspace]
+resolver = "2"
+members = [ "bar","foo"]
+
+```
+
+The package at this point contains the following files:
+
+```console
+$ cd new_workspace
+$ tree .
+.
+├── bar
+│   ├── Cargo.toml
+│   └── src
+│       └── lib.rs
+├── Cargo.toml
+└── foo
+    ├── Cargo.toml
+    └── src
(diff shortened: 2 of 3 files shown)


src/doc/src/SUMMARY.md
```diff
@@ -9,6 +9,7 @@
 * [Cargo Guide](guide/index.md)
     * [Why Cargo Exists](guide/why-cargo-exists.md)
     * [Creating a New Package](guide/creating-a-new-project.md)
+    * [Creating a Workspace](guide/creating-a-new-workspace.md)
     * [Working on an Existing Package](guide/working-on-an-existing-project.md)
```
epage, line 13: Within the scope of things, we are introducing workspaces before we introduce dependencies, tests, etc.  This likely should be later

src/doc/src/guide/creating-a-new-workspace.md
```diff
@@ -0,0 +1,75 @@
+# Creating a New Workspace
+
+A [workspace][def-workspace] is a collection of one or more packages, called workspace members, that are managed together.
```
weihanglo, line 3: In Cargo's documentation, we usually either wrap at column like 80/120, or at where makes sense and natural like punctuations or conjunctions.

src/doc/src/guide/creating-a-new-workspace.md
```diff
@@ -0,0 +1,75 @@
...
+
+In this chapter we implement the creation of a workspace `new_workspace` from scratch, in which two members `foo-bin` and `bar-lib` are included.
+
+As mentioned in [[workspace] section](../reference/workspaces.md#the-workspace-section), the workspace must have at least one member, either the root package or a virtual manifest. Here we recommend setting the root of the workspace to be a virtual manifest. The reason is as follows:
```
weihanglo, line 7: To be consistent, use [link reference](https://spec.commonmark.org/0.31.2/#link-reference-definitions) here.

weihanglo, line 7: Need to link to the root package and virtual workspace paragraphs respectively.

src/doc/src/guide/creating-a-new-workspace.md
```diff
@@ -0,0 +1,75 @@
...
+
+- The crate namespace at the Cargo level is flat. The tree layout creates another hierarchy and increases the possibility of inconsistencies. 
+- With a flat structure, adding or splitting crates is very easy. 
+- Even larger flattened lists are easier to see at a glance and easier to maintain than smaller tree structures.
+
+Let's look at how a workspace containing a virtual inventory should be created.
```
weihanglo, line 13: I don't think we used "virtual inventory" elsewhere. Avoid introducing new jargon and keep the guide written in simple English.

```suggestion
Let's move on and create a virtual workspace.
```

src/doc/src/guide/creating-a-new-workspace.md
```diff
@@ -0,0 +1,75 @@
...
+Cargo will automatically add members to `Cargo.toml`:
+
+```toml
+# [new_workspace]/Cargo.toml
+[workspace]
+resolver = "2"
+members = [ "bar-lib","foo-bin"]
+
+[package]
+# ...
```
weihanglo, line 48: A virtual workspace shouldn't contain a `[package]` section.

src/doc/src/guide/creating-a-new-workspace.md
```diff
@@ -0,0 +1,75 @@
...
+resolver = "2"
+```
+
+You can then proceed to create workspace members. As mentioned above, if you have a lot of workspace members, we recommend designing the workspace with a flat structure. So add all the workspace members in the directory `new_workspace/crates`.
+
+There are several ways to add members to the workspace, we recommend using the command `cargo new/init`. For example, add the members `foo-bin` and `bar-lib` to the workspace:
+
+```console
+$ cargo new crates/foo-bin
+$ cargo new crates/bar-lib
```
weihanglo, line 36: it's weird we call this `-lib` but it is actually not a lib.

src/doc/src/guide/creating-a-new-workspace.md
```diff
@@ -0,0 +1,75 @@
...
+# [new_workspace]/Cargo.toml
+[workspace]
+resolver = "2"
+members = [ "bar-lib","foo-bin"]
+
+[package]
+# ...
+```
+
+The project at this point contains the following files:
```
weihanglo, line 51:
```suggestion
The package at this point contains the following files:
```

src/doc/src/guide/creating-a-new-workspace.md
```diff
@@ -0,0 +1,75 @@
...
+    │       └── main.rs
+    └── foo-bin
+        ├── Cargo.toml
+        └── src
+            └── main.rs
+
+5 directories, 5 files
+```
+
+Up to this point, we have achieved creating a workspace that contains multiple `crate` members through a step-by-step process.
```
weihanglo, line 71: I don't understand why quoting "`crate`". Also, could put some basic usages of workspace here maybe?

```suggestion
Up to this point, we have a workspace with two members.
Whenever you run `cargo build` under the workspace root directory, Cargo builds all member at once.
Instead of building the entire workspace, you could use the `--package`/`-p` flag to select certain packages.
For example, `cargo build -p foo-bin` will build only `foo-bin` package.
```

weihanglo (commented):
Just skimmed through. Some minor suggestions added.

src/doc/src/guide/creating-a-new-workspace.md
epage: https://rhodesmill.org/brandon/2012/one-sentence-per-line/

src/doc/src/guide/creating-a-new-workspace.md
```diff
@@ -0,0 +1,75 @@
...
+
+A [workspace][def-workspace] is a collection of one or more packages, called workspace members, that are managed together.
+
+In this chapter we implement the creation of a workspace `new_workspace` from scratch, in which two members `foo-bin` and `bar-lib` are included.
+
+As mentioned in [[workspace] section](../reference/workspaces.md#the-workspace-section), the workspace must have at least one member, either the root package or a virtual manifest. Here we recommend setting the root of the workspace to be a virtual manifest. The reason is as follows:
+
+- The crate namespace at the Cargo level is flat. The tree layout creates another hierarchy and increases the possibility of inconsistencies. 
+- With a flat structure, adding or splitting crates is very easy. 
+- Even larger flattened lists are easier to see at a glance and easier to maintain than smaller tree structures.
```
epage, line 11: This feels like its blurring the line between a guide-level document and the best practices examples that would hang off the reference.

Seeing this proposed document, I feel like its not pulling enough weight and I wonder if we should first focus in that examples / practices document

src/doc/src/guide/creating-a-new-workspace.md
heisen-li: Sorry, I must have been busy.This is stupid.

src/doc/src/guide/creating-a-new-workspace.md
heisen-li: Perhaps you mean that if add this section it might become a best practice type of article rather than an instructional one. Then I have adjusted the article to make it more suitable as an instructional article for newbies.

heisen-li:
Is there anything else that needs to be changed on this, please?

src/doc/src/guide/creating-a-new-workspace.md
epage: > Is there anything else that needs to be changed on this, please?


That went in the opposite direction of what I was suggesting.  I'll have to give this more time to think about this to avoid any bias from my own way of wanting to do things.

src/doc/src/guide/creating-a-new-workspace.md
```diff
@@ -0,0 +1,141 @@
...
+to make it a manifest of the workspace:
+
+```toml
+# [new_workspace]/Cargo.toml
+[workspace]
+
+[package]
+name = "new_workspace"
+version = "0.1.0"
+edition = "2021"
```
weihanglo, line 24: This feels odd because `new_workspace` is actually a package. The step-by-step guide is not an ideal workflow, and even more confusing at this moment. Admittedly this is the tool's failure <https://github.com/rust-lang/cargo/issues/8365>. I wonder if we want to block this on that.

weihanglo:
This PR has been stale for a while, and not in a good state of merging. I am going to close now. Thank you for being interested in improving documentations!

PR #13767, opened 2024-04-17
test(msrv): Re-organize MSRV tests

epage:

- Merge test cases
- Clarify names
- Focus on `cargo tree`, rather than `cargo check` (faster and more to
  what we are testing)


ehuss:
Thanks!

@bors r+

PR #13770, opened 2024-04-17
fix(credential): trim newlines in tokens from stdin

arlosi:

`cargo login` when using a credential provider other than `cargo:token` does not automatically trim whitespace from tokens.

This can lead to extra whitespace being included in the pasted token value (usually a trailing newline) that makes the token invalid.

First commit adds a test showing the problematic behavior. Second commit fixes it.

### DIFF
--- a/tests/testsuite/login.rs
+++ b/tests/testsuite/login.rs
@@ -113,6 +113,7 @@ fn empty_login_token() {
         .with_stderr(
             "\
 [UPDATING] crates.io index
+please paste the token found on [..] below
 [ERROR] credential provider `cargo:token` failed action `login`
 
 Caused by:
--- a/src/cargo/ops/registry/login.rs
+++ b/src/cargo/ops/registry/login.rs
@@ -39,7 +39,7 @@ pub fn registry_login(
     let mut token_from_stdin = None;
     let token = token_from_cmdline.or_else(|| {
         if !std::io::stdin().is_terminal() {
-            let token = std::io::read_to_string(std::io::stdin()).unwrap_or_default();
+            let token = cargo_credential::read_line().unwrap_or_default();
             if !token.is_empty() {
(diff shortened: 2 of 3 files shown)


epage:
I know this is unlikely but is there any case where newlines could be significant to these tokens?

arlosi:
> I know this is unlikely but is there any case where newlines could be significant to these tokens?

The token is directly sent as the `Authorization` HTTP header, which doesn't allow newlines.

If you try to use it, you get:

```
cargo yank asdf --version 0.0.1
    Updating crates.io index
        Yank asdf@0.0.1
error: failed to yank from the registry at https://crates.io

Caused by:
  token contains invalid characters.
  Only printable ISO-8859-1 characters are allowed as it is sent in a HTTPS header.
```

tests/testsuite/login.rs
```diff
@@ -113,6 +113,7 @@ fn empty_login_token() {
         .with_stderr(
             "\
 [UPDATING] crates.io index
+please paste the token found on [..] below
```
epage, line 116: How is this test passing today?

arlosi: In the `read_token` function, if a token is provided by `stdin` or by the command line, then no prompt is made. 

The change causes `login_options.token` to be `None`, instead of a `Some("\n")`, so Cargo now prints the "please paste" message.
 
https://github.com/rust-lang/cargo/blob/852a31615d8ad66cf9768e16ef50119806629027/credential/cargo-credential/src/lib.rs#L281-L290

epage: Could you move that change into the commit that caused it so that each commit passes tests?

arlosi: Fixed. I should have amended that instead of adding a commit.

PR #13775, opened 2024-04-18
fix(toml)!: Disallow source-less dependencies

epage:

This is part of #13629 addressing “dependency without path, version, git, workspace specified”.

This turns deps like
```toml
foo = { optional = true }
```
from `version="*"` deps with a warning into errors. This breaking change was deemed acceptable because this behavior has been considered a bug from the beginning.
We have gotten little to no feedback about people wanting this behavior.

This improves our forwards-compatibility story as we can add new dependency sources and they won't be considered a wildcard registry dependency on older cargo.

I removed the `cargo_add` test because it is redundant.

We no longer get the “unused key” warnings because those warnings get deferred to after this error gets reported.


src/cargo/util/toml/mod.rs
```diff
@@ -1672,14 +1672,11 @@ fn detailed_dep_to_dependency<P: ResolveToPath + Clone>(
     kind: Option<DepKind>,
 ) -> CargoResult<Dependency> {
     if orig.version.is_none() && orig.path.is_none() && orig.git.is_none() {
-        let msg = format!(
-            "dependency ({}) specified without \
+        anyhow::bail!(
```
weihanglo, line 1675: Question: What's the criteria of turning this into

* a normal warning
* * a critical warning that bails out later
* a fatal error bailing out here

weihanglo (approved):

src/cargo/util/toml/mod.rs
epage: > turning this into a normal warning

This was already a warning, so no change there

> a fatal error bailing out here

When we discussed #13629 as a team, we seemed in favor of a hard error as this was deemed a bug that people haven't been clamoring for us to continue to support.

> a critical warning that bails out later

I honestly haven't touched that system much.  I think those "recoverable errors", right?  As in, we continue moving forward, seeing other messages, and but fail after all of that?

While I would want to move more towards error recovery, this has some weird properties
- These only show up when `emit_warnings` is called which looks to only be compile commands and `cargo fetch`
- The overall value of error recovery seems relatively low (not much more is really going to be shown but instead people might see other unrelated warnings)

weihanglo: Thanks for the clarification. I understand this being a fatal error as a team decision. The recovery error is something I am also thinking about, yet it doesn't apply to this situation.

weihanglo (approved):

weihanglo:
@bors r+

Thanks!

PR #13777, opened 2024-04-18
fix 13773 - 'cargo build' fails when list_files() with gix is triggered

### LINKED ISSUE #13773: Error building path dependency inside symlinked git repository
### Problem

`cargo build` fails if a path dependency points to a subdirectory of a symlinked git repository and has a build script.

### Steps

Run the following commands in order.

```sh
cargo new repro
cd repro
mkdir original
cd original
git init
cargo new --lib lib
echo 'fn main() {}' >lib/build.rs
git add .
cd ..
ln -s original symlink
cargo add --path symlink/lib
cargo build
```

Observe that `cargo build` prints an error.

```text
error: failed to determine package fingerprint for build script for lib v0.1.0 (…/repro/symlink/lib)

Caused by:
  failed to determine the most recently modified file in …/repro/symlink/lib

Caused by:
  failed to determine list of files in …/repro/symlink/lib

Caused by:
  Worktree root at '…/repro/symlink' is not a directory
```

### Possible Solution(s)

_No response_

### Notes

This is probably related to gitoxide integration as `__CARGO_GITOXIDE_DISABLE_LIST_FILES=1 cargo build` succeeds.

### Version

```text
cargo 1.79.0-nightly (6f06fe908 2024-04-16)
release: 1.79.0-nightly
commit-hash: 6f06fe908a5ee0f415c187f868ea627e82efe07d
commit-date: 2024-04-16
host: aarch64-apple-darwin
os: Mac OS 14.4.1 [64-bit]
```

Byron:
Fixes #13773.

### Tasks

* [x] reproduce issue with new test-case
* [x] update [fixed `gix-dir`](https://github.com/rust-lang/cargo/pull/13777) in Cargo.lock to turn test green

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1296,9 +1296,9 @@ dependencies = [
 
 [[package]]
 name = "gix-dir"
-version = "0.4.0"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d6943a1f213ad7a060a0548ece229be53f3c2151534b126446ce3533eaf5f14c"
+checksum = "3d6fcd56ffa1133f35525af890226ad0d3b2e607b4490360c94b1869e278eba3"
 dependencies = [
  "bstr",
  "gix-discover",
@@ -1431,9 +1431,9 @@ dependencies = [
 
 [[package]]
 name = "gix-index"
-version = "0.32.0"
+version = "0.32.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3383122cf18655ef4c097c0b935bba5eb56983947959aaf3b0ceb1949d4dd371"
+checksum = "881ab3b1fa57f497601a5add8289e72a7ae09471fc0b9bbe483b628ae8e418a1"
 dependencies = [
  "bitflags 2.5.0",
  "bstr",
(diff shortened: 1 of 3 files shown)


weihanglo:
> Actually I think I did. The first commit https://github.com/rust-lang/cargo/commit/07d2bd751708bd438975f80b3db6aa648baad2a3 adds a test and fails exactly like it should, the second commit includes the fix. Is there anything I am missing?

Yes, you did it pretty good. For myself, I also match the failure output in the test in the first commit, so the next commit we shows both fix and test change.

Should we wait for <https://github.com/Byron/gitoxide/commit/5dea6f10eb8cef0b3ab3650169a557d8c1141359> and update to the new version?

weihanglo:
Ok okay I missed this:

> An immediate change to what's released isn't necessary, I think, so these improvements will come in with the next natural release in a month or so.

Let's merge this. Thanks!

@bors r+

weihanglo (approved):

Byron:
> Yes, you did it pretty good. For myself, I also match the failure output in the test in the first commit, so the next commit we shows both fix and test change.

Oh, right! I don't know how I could miss this opportunity - I just visually inspected the output. With a test-assertion on stderr, it would have been far clearer though, as the test would still have failed. Next time, even though I hope there won't be, who needs bugs ;).

PR #13778, opened 2024-04-18
Unused dependencies cleanup

Muscraft:
The implementation of #12826 was used as a demonstration for #12235, in #13621. The demonstration implementation was far from ideal and was lacking a few features. This PR makes the implementation "feature complete", and fixes known bugs with the initial implementation.


tests/testsuite/lints/implicit_features/warn/mod.rs
```diff
@@ -27,10 +27,10 @@ implicit-features = "warn"
         .build();
 
     snapbox::cmd::Command::cargo_ui()
-        .masquerade_as_nightly_cargo(&["always_nightly"])
+        .masquerade_as_nightly_cargo(&["cargo-lints", "edition2024"])
         .current_dir(p.root())
         .arg("check")
-        .arg("--quiet")
+        .arg("-Zcargo-lints")
```
epage: Unrelated?

tests/testsuite/lints/implicit_features/edition_2021/mod.rs
```diff
@@ -23,12 +23,10 @@ bar = { version = "0.1.0", optional = true }
         .build();
 
     snapbox::cmd::Command::cargo_ui()
-        .masquerade_as_nightly_cargo(&["always_nightly"])
         .current_dir(p.root())
         .arg("check")
-        .arg("--quiet")
```
epage, line 29: Why drop `--quiet`?

src/cargo/util/lints.rs
```diff
@@ -144,10 +144,10 @@ impl From<TomlLintLevel> for LintLevel {
 /// [RFC #3491]: https://rust-lang.github.io/rfcs/3491-remove-implicit-features.html
 const IMPLICIT_FEATURES: Lint = Lint {
     name: "implicit_features",
-    desc: "warn about the use of unstable features",
+    desc: "implicit features for optional dependencies is deprecated and will be unavailable in the 2024 edition",
```
epage, line 160: Unrelated to the current commit

src/cargo/util/lints.rs
```diff
@@ -168,48 +168,41 @@ pub fn check_implicit_features(
...
-    let explicitly_listed = user_defined_features.map_or(HashSet::new(), |f| {
-        f.values()
-            .flatten()
-            .filter_map(|v| match FeatureValue::new(v.into()) {
-                Dep { dep_name } => Some(dep_name),
-                _ => None,
-            })
-            .collect()
-    });
+    let activated_opt_deps = manifest
```
epage, line 184: Unrelated cleanup?

src/cargo/util/lints.rs
```diff
@@ -215,3 +227,98 @@ pub fn check_implicit_features(
...
+    }
+
+    let manifest = pkg.manifest();
+    let user_defined_features = manifest.resolved_toml().features();
+    let features = user_defined_features.map_or(HashSet::new(), |f| {
+        f.keys().map(|k| InternedString::new(&k)).collect()
+    });
+    // Add implicit features for optional dependencies if they weren't
+    // explicitly listed anywhere.
+    let explicitly_listed = user_defined_features.map_or(HashSet::new(), |f| {
```
epage: A previous commit changed the style of this for the other lint.  Is there a reason this one is still this way?

src/cargo/util/lints.rs
```diff
@@ -215,3 +227,98 @@ pub fn check_implicit_features(
...
+                    ),
+                )
+                .fold(true),
+        );
+        if emitted_source.is_none() {
+            emitted_source = Some(format!(
+                "`cargo::{}` is set to `{lint_level}`",
+                UNUSED_OPTIONAL_DEPENDENCY.name
+            ));
+            message = message.footer(Level::Note.title(emitted_source.as_ref().unwrap()));
```
epage: Shouldn't this also be on the other lint as well?

epage: (could be its own commit on that lint)

tests/testsuite/lints/implicit_features/edition_2024/mod.rs
```diff
@@ -23,15 +23,19 @@ baz = { version = "0.1.0", optional = true }
 
 [features]
 baz = ["dep:baz"]
+
+[lints.cargo]
+unused-optional-dependency = "allow"
```
epage, line 28: Why was this added?

src/cargo/util/lints.rs
```diff
@@ -1,3 +1,4 @@
+use crate::core::dependency::DepKind;
 use crate::core::FeatureValue::Dep;
```
epage, line 2: This commit is named after what it does but not the why ("avoid creating implicit features on edition 2024")

tests/testsuite/lints/unused_optional_dependencies/lint_source_once/mod.rs
```diff
@@ -0,0 +1,36 @@
+use cargo_test_support::prelude::*;
+use cargo_test_support::registry::Package;
+use cargo_test_support::str;
+use cargo_test_support::{file, project};
+
+#[cargo_test(nightly, reason = "edition2024 is not stable")]
```
epage, line 6: Isn't this test unrelated to this commit?

tests/testsuite/lints/unused_optional_dependencies/correct_dep_table/mod.rs
```diff
@@ -0,0 +1,42 @@
+use cargo_test_support::prelude::*;
+use cargo_test_support::registry::Package;
+use cargo_test_support::str;
+use cargo_test_support::{file, project};
+
+#[cargo_test(nightly, reason = "edition2024 is not stable")]
+fn case() {
```
epage, line 7: This should probably be a test in `tests/testsuite/package.rs`, rather than a UI test for lints

tests/testsuite/lints/implicit_features/warn/mod.rs
Muscraft: This was needed to not show the warning about using `lints.cargo`

tests/testsuite/lints/implicit_features/edition_2021/mod.rs
Muscraft: It would've been an empty file, which does show that the lint wouldn't have run, but it felt odd. This should probably be switched to a non-UI test in the long run.

tests/testsuite/lints/implicit_features/warn/mod.rs
epage: Is that lint hidden with `--quiet` and so by removing that we now need this?

Shouldn't we have not gotten any lints without this `-Z`?

src/cargo/util/lints.rs
```diff
@@ -144,7 +144,7 @@ impl From<TomlLintLevel> for LintLevel {
 /// [RFC #3491]: https://rust-lang.github.io/rfcs/3491-remove-implicit-features.html
 const IMPLICIT_FEATURES: Lint = Lint {
     name: "implicit_features",
-    desc: "warn about the use of unstable features",
+    desc: "implicit features for optional dependencies is deprecated and will be unavailable in the 2024 edition",
```
epage, line 160: nit: I would have had this in the commit where this message is used

(this is more for future reference)

src/cargo/util/lints.rs
```diff
@@ -192,7 +192,7 @@ pub fn check_implicit_features(
         }
         let level = lint_level.to_diagnostic_level();
         let manifest_path = rel_cwd_manifest_path(path, gctx);
-        let message = level.title("unused optional dependency").snippet(
+        let message = level.title(IMPLICIT_FEATURES.desc).snippet(
```
epage: nit (more for future reference): this is not a `refactor` but a `fix`

src/cargo/util/lints.rs
```diff
@@ -217,3 +235,113 @@ pub fn check_implicit_features(
...
+
+pub fn unused_dependencies(
+    pkg: &Package,
+    path: &Path,
+    lints: &TomlToolLints,
+    error_count: &mut usize,
+    gctx: &GlobalContext,
+) -> CargoResult<()> {
+    let edition = pkg.manifest().edition();
+    // This lint is only relevant for editions 2024 and later.
```
epage, line 255: nit: to get at more of the intent
```suggestion
    // Dependencies can only be used on edition 2024+
```

src/cargo/util/lints.rs
```diff
@@ -157,56 +170,61 @@ pub fn check_implicit_features(
     error_count: &mut usize,
     gctx: &GlobalContext,
 ) -> CargoResult<()> {
-    let lint_level = IMPLICIT_FEATURES.level(lints, pkg.manifest().edition());
+    let edition = pkg.manifest().edition();
+    if edition >= Edition::Edition2024 {
+        return Ok(());
+    }
```
epage, line 176: This should likewise probably have a comment like
```suggestion
    let edition = pkg.manifest().edition();
    // In Edition 2024+, instead of creating optional features, the dependencies are unused.  See [`UNUSED_OPTIONAL_DEPENDENCY`]
    if edition >= Edition::Edition2024 {
        return Ok(());
    }
```

src/cargo/util/lints.rs
```diff
@@ -260,63 +261,87 @@ pub fn unused_dependencies(
...
+            vec!["target", name, DepKind::Development.kind_table()],
+        ));
+        orig_deps.push((
+            platform.build_dependencies.as_ref(),
+            vec!["target", name, DepKind::Normal.kind_table()],
+        ));
+    }
+    for (deps, toml_path) in orig_deps {
+        if let Some(deps) = deps {
+            for (name, _) in deps {
```
epage, line 302: nit: can we do `deps.keys()`?

src/cargo/util/lints.rs
```diff
@@ -260,63 +261,87 @@ pub fn unused_dependencies(
...
-                .flatten()
-                .filter_map(|f| match FeatureValue::new(InternedString::new(f)) {
-                    Dep { dep_name } => Some(dep_name.as_str()),
-                    _ => None,
-                })
-                .collect::<HashSet<_>>()
-        })
-        .unwrap_or_default();
+    let original_toml = manifest.original_toml();
+    let all_dependencies = manifest
```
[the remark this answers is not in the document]
epage: (maybe even add a comment that unused dependencies were stripped)

src/cargo/util/toml/mod.rs
```diff
@@ -610,10 +645,18 @@ fn resolve_dependencies<'a>(
...
 
-        deps.insert(
-            name_in_toml.clone(),
-            manifest::InheritableDependency::Value(resolved.clone()),
-        );
+        // if the dependency is not optional, it is always used
+        // if the dependency is optional and activated, it is used
+        // if the dependency is optional and not activated, it is not used
+        let is_dep_activated =
+            !resolved.is_optional() || activated_opt_deps.contains(name_in_toml.as_str());
```
epage, line 652: > `name_in_toml`

Should we have a test for renamed packages to make sure the semantics of this are correct?

src/cargo/util/toml/mod.rs
```diff
@@ -610,10 +645,18 @@ fn resolve_dependencies<'a>(
...
+        // If the edition is less than 2024, we don't need to check for unused optional dependencies
+        if edition < Edition::Edition2024 || is_dep_activated {
```
epage, line 654: We should have a `cargo publish` test to make sure the published `Cargo.toml` and the `Summary` has these stripped

src/cargo/util/lints.rs
```diff
@@ -260,63 +261,87 @@ pub fn unused_dependencies(
...
+        .dependencies()
+        .into_iter()
+        .map(|d| d.name_in_toml().to_string())
+        .collect::<HashSet<String>>();
```
epage, line 271: We probably shouldn't be converting away from `InternedString` but converting things to it

epage: Though I'm unsure if this will make much of a difference, so its probably fine

src/cargo/util/lints.rs
```diff
@@ -217,3 +235,113 @@ pub fn check_implicit_features(
...
+pub fn unused_dependencies(
+    pkg: &Package,
+    path: &Path,
+    lints: &TomlToolLints,
+    error_count: &mut usize,
+    gctx: &GlobalContext,
+) -> CargoResult<()> {
+    let edition = pkg.manifest().edition();
+    // This lint is only relevant for editions 2024 and later.
+    if edition <= Edition::Edition2021 {
```
arlosi, line 256: Since the comment is referencing Edition 2024, would this be clearer as:

```suggestion
    if edition < Edition::Edition2024 {
```

src/cargo/util/lints.rs
arlosi: This suggested change makes it more confusing to me. We can always have dependencies, right?

Maybe:

> Unused optional dependencies are always allowed prior to Edition 2024.

src/cargo/util/toml/mod.rs
arlosi: Do we also need a test that shows it's not in the lockfile?

src/cargo/util/lints.rs
epage: I think I left off "Unused" at the start of my suggestion

The problem with that suggestion is that unused optional dependencies don't exist prior to 2024

src/cargo/util/lints.rs
```diff
@@ -198,12 +211,19 @@ pub fn check_implicit_features(
...
                 .origin(&manifest_path)
                 .annotation(level.span(get_span(manifest.document(), &toml_path, false).unwrap()))
                 .fold(true),
         );
+        if emitted_source.is_none() {
+            emitted_source = Some(format!(
+                "`cargo::{}` is set to `{lint_level}`",
+                IMPLICIT_FEATURES.name
+            ));
+            message = message.footer(Level::Note.title(emitted_source.as_ref().unwrap()));
```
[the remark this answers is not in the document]
epage: Noted this on #12235

PR #13783, opened 2024-04-19
fix(toml): Report `_` fied variants (e.g. `dev_dependencies`) as deprecated

epage:

This is prep for removing them in the 2024 Edition and is part of rust-lang/rust#123754 and #13629

This doesn't include 2024 Edition work because there is a risk of conflict with other work going on these areas.

This changes us from
- When using `-` and `_` variants: deprecated, will error some time
- Otherwise, nothing

To
- When using `-` and `_` variants: unused field warning
- When using only `_`: deprecation, will be removed in 2024

I decided to model this as an unused field warning because that is what this is and that is how any other use of `_` works.  We might hard error during a transition period but I'd eventually want these to just make the code act like anything else in the end.


ehuss:
Thanks!
The cargo team discussed this in the 2024-04-16 meeting as a part of the 2024 deprecation sweep.
@bors r+

PR #13776, opened 2024-04-18
feat(resolver): Add v3 resolver for MSRV-aware resolving

epage:
### What does this PR try to resolve?
This is a part of #9930 and is important for changing the default with the new edition.

PR #13785, opened 2024-04-20
feat(resolver): Add default Edition2024 to resolver v3

epage:

With #13776 done, we can now make MSRV-aware resolver the default for the new edition as part of #9930

### DIFF
--- a/tests/testsuite/rust_version.rs
+++ b/tests/testsuite/rust_version.rs
@@ -614,6 +614,117 @@ foo v0.0.1 ([CWD])
 }
 
 #[cargo_test(nightly, reason = "edition2024 in rustc is unstable")]
+fn resolve_edition2024() {
+    Package::new("only-newer", "1.6.0")
+        .rust_version("1.65.0")
+        .file("src/lib.rs", "fn other_stuff() {}")
+        .publish();
+    Package::new("newer-and-older", "1.5.0")
+        .rust_version("1.55.0")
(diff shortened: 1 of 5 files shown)


weihanglo (approved):

weihanglo:
I am forgetting. Do we have any previous discussion or FCP for making MSRV resolver the default in edition 2024?

tests/testsuite/rust_version.rs
weihanglo, line 1: Do we have a test for `workspace.resolver = "3"`, and should we have one?

epage: #13776 had tests

epage:
> I am forgetting. Do we have any previous discussion or FCP for making MSRV resolver the default in edition 2024?

The RFC called for making it a default in the next edition.  My plan is to stabilize it before the window closes for 2024.  If not, we can adjust from there.

tests/testsuite/rust_version.rs
weihanglo: Yes, but not for `resolver = "3"` at workspace level specifically. Do we need that?

epage: So long as we have some tests for workspace/package resolver interactions, I don't think we need the tests for every resolver version.

weihanglo: Fair enough!

weihanglo:
@bors r+

Thanks!

PR #13788, opened 2024-04-22
fix(toml)!: Remove support for inheriting badges

### LINKED ISSUE #13643: Undocumented, inconsisten inheritance of `[badges]`
### Problem

Despite the [`[workspace.package]`](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) not mentioning it, you can specify `workspace.package.badges` to inherit `badges` (note its not `package.badges`

Problems
- This feature is undocumented
- We should be inheriting from `workspace.badges`, not `workspace.package.badges`
- `cargo new` will automatically add a `package.badges.workspace = true`

### Steps

Create the following workspace:
```rust
[workspace]
members = ["child"]
resolver = "2"

[workspace.package.badges]
maintenance = { status = "actively-maintained" }
```

Run:
```console
$ cargo new child
$ cat child/Cargo.toml  # see `package.badges.workspace = true`
$ cargo package -p child
$ cat target/package/child-0.1.0/Cargo.toml  # see `badges` table
```

### Possible Solution(s)

Maybe we break this as no one likely knows this exists?

For example, even looking at packages that try to implement `Cargo.toml` support
- `cargo_toml` doesn't acknowledge this
  - See their inheritable `workspace.package` fields: https://docs.rs/cargo_toml/0.19.2/cargo_toml/struct.Manifest.html
  - See their `badges` field: https://docs.rs/cargo_toml/0.19.2/cargo_toml/struct.Manifest.html
- `cargo_manifest` doesn't acknowledge this
  - See their inheritable `workspace.package` fields: https://docs.rs/cargo-manifest/0.13.0/cargo_manifest/struct.WorkspacePackage.html
  - See their `badges` field: https://docs.rs/cargo-manifest/0.13.0/cargo_manifest/struct.Manifest.html

### Notes

Note: I'm in the middle of a big refactor: please don't touch this yet

### Version

_No response_

epage:

We allowed `[badges]` to inherit from `[workspace.package.badges]` which was a bug:
- This was not specified in the RFC
- We did not document this
- Even if someone were to try to guess to use this, it is inconsistent with how inheritance works because this should inherit from `workspace.badges` instead of `workspace.package.badges`

While keeping in mind that `[badges]` is effectively deprecated.

In that context, I think its safe to break support for this without a transition period.

Fixes #13643

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -466,7 +466,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-util-schemas"
-version = "0.3.1"
+version = "0.4.0"
 dependencies = [
  "semver",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ cargo-platform = { path = "crates/cargo-platform", version = "0.1.5" }
 cargo-test-macro = { version = "0.2.0", path = "crates/cargo-test-macro" }
 cargo-test-support = { version = "0.2.0", path = "crates/cargo-test-support" }
 cargo-util = { version = "0.2.9", path = "crates/cargo-util" }
-cargo-util-schemas = { version = "0.3.0", path = "crates/cargo-util-schemas" }
+cargo-util-schemas = { version = "0.4.0", path = "crates/cargo-util-schemas" }
 cargo_metadata = "0.18.1"
 clap = "4.5.4"
 color-print = "0.3.5"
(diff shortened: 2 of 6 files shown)


weihanglo (approved):

weihanglo:
As this is sort of a breaking change to an undocumented behavior that shouldn't have happened for a deprecated field, I'd like to start a poll.

(not going to wait for a full 10-days FCP after we reach a consensus)

@rfcbot fcp merge

weihanglo:
@bors r+

Thanks for the update.

PR #13790, opened 2024-04-23
fix(install): Don't respect MSRV for non-local installs 

epage:

This is part of #9930

### DIFF
--- a/src/cargo/core/workspace.rs
+++ b/src/cargo/core/workspace.rs
@@ -104,7 +104,6 @@ pub struct Workspace<'gctx> {
     /// The resolver behavior specified with the `resolver` field.
     resolve_behavior: ResolveBehavior,
     resolve_honors_rust_version: bool,
-    honor_rust_version: Option<bool>,
 
     /// Workspace-level custom metadata
     custom_metadata: Option<toml::Value>,
@@ -235,7 +234,6 @@ impl<'gctx> Workspace<'gctx> {
             ignore_lock: false,
             resolve_behavior: ResolveBehavior::V1,
             resolve_honors_rust_version: false,
-            honor_rust_version: None,
             custom_metadata: None,
         }
     }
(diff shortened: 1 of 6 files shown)


src/cargo/core/workspace.rs
```diff
@@ -103,7 +103,7 @@ pub struct Workspace<'gctx> {
 
     /// The resolver behavior specified with the `resolver` field.
     resolve_behavior: ResolveBehavior,
-    resolve_honors_rust_version: bool,
+    resolve_honors_rust_version: Option<bool>,
```
weihanglo, line 106: Wonder why we need an `Option` here. It doesn't seem too relevant to this PR.

src/cargo/core/workspace.rs
```diff
@@ -657,10 +657,15 @@ impl<'gctx> Workspace<'gctx> {
         self.honor_rust_version
     }
 
+    pub fn set_resolve_honors_rust_version(&mut self, honor_rust_version: Option<bool>) {
+        self.resolve_honors_rust_version = honor_rust_version;
+    }
+
     pub fn resolve_honors_rust_version(&self) -> bool {
```
weihanglo, line 664: I am now seeing `resolve_honors_rust_version` and `honors_rust_version` quiet confusing. Any change to explain or conslidate them?

src/cargo/core/workspace.rs
epage: Oops, that represents a bug.  Fixing it.

src/cargo/core/workspace.rs
weihanglo: BTW, not a blocker

src/cargo/core/workspace.rs
epage: I misread, there wasn't a bug and I didn't need to take advantage of this.  Simplifying the code.

weihanglo (approved):
Thanks!

Issue #9857, opened 2021-08-30
`cargo fix` may attempt to write to the registry cache


ehuss:
**Problem**
If `rustc` provides a suggestion to modify code in a dependency, `cargo fix` will happily oblige. However, cargo should never apply fixes to non-path dependencies.  I might say it shouldn't apply fixes outside of the current package, but I'm not sure about that (or how to detect that reliably).

For example, it may attempt to write to `~/.cargo/registry/src/github.com-1ecc6299db9ec823` which really should be read-only.

`rustc` normally tries to avoid diagnostics that point into dependencies, but sometimes there are bugs where it points to the wrong span, and ends up giving a suggestion into a dependency.  See https://github.com/rust-lang/rust/issues/88514 for more detail.

**Steps**
It's hard to give a good example, since it either requires using a real crates.io package, or depending on diagnostics in rustc that are arguably not working correctly.

### Method 1 — Use crates.io

With dependency `proconio = "=0.3.8"`

```rust
pub fn foo(
  source: proconio::source::line::LineSource<std::io::BufReader<&[u8]>>,
) {
  proconio::input! {
    from source,
    mut x:usize,
  }
}
```

This will suggest making a change to the remote package:

```
warning: unused variable: `x`
   --> /Users/eric/.cargo/registry/src/github.com-1ecc6299db9ec823/proconio-0.3.8/src/lib.rs:579:22
    |
579 |         let $($mut)* $var = $crate::read_value!(@source [$source] @kind [$($kind)*]);
    |                      ^^^^ help: if this is intentional, prefix it with an underscore: `_x`
    |
    = note: `#[warn(unused_variables)]` on by default
```

### Method 2 — Sample macro

Put this macro in a dependency, and notice rustc will suggest modifying the macro.

```rust
#[macro_export]
macro_rules! m {
    ($i:tt) => {
        let $i = 1;
    };
}
```

The key part is that the identifier uses a `tt` instead of an `ident`.

### Method 3 — cargo_test

This is a cargo testsuite test.  I don't think we can use it since it is dependent on rustc's diagnostic output, but maybe we can simulate it with a rustc wrapper? 

```rust
#[cargo_test]
fn fix_in_dependency() {
    Package::new("bar", "1.0.0")
        .file(
            "src/lib.rs",
            r#"
                #[macro_export]
                macro_rules! m {
                    ($i:tt) => {
                        let $i = 1;
                    };
                }
            "#,
        )
        .publish();

    let p = project()
[...]

PR #13792, opened 2024-04-23
fix(cargo-fix): dont fix into standard library

weihanglo:

Protect standard library source from being fixed by `cargo fix`.

This is a bug in rustc, which it shouldn't have given any fix suggestion to the standard library.
However, I believe Cargo should have a layer of protection regardless it's a bug in rustc or not.

See <https://github.com/rust-lang/cargo/issues/9857#issuecomment-2043477022>

This first commit shows the problematic behavior. The second fixes it.

The alternative idea has been brought up: only paths under workspace root are fixable.
However, it might encounter some corner cases:

* `#[path]` attribute pointing outside a workspace.
* Path dependencies that are not inside the same workspace.

I took this conservative approach that Cargo just stop writing into whatever pointing to sysroot.

r? ehuss

### DIFF
--- a/src/cargo/ops/fix.rs
+++ b/src/cargo/ops/fix.rs
@@ -50,6 +50,7 @@ use rustfix::CodeFix;
 use semver::Version;
 use tracing::{debug, trace, warn};
 
+use crate::core::compiler::CompileKind;
 use crate::core::compiler::RustcTargetData;
 use crate::core::resolver::features::{DiffMap, FeatureOpts, FeatureResolver, FeaturesFor};
 use crate::core::resolver::{HasDevUnits, Resolve, ResolveBehavior};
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks! I still hope we can get this fixed in rustc (https://github.com/rust-lang/rust/issues/88514) so it doesn't provide external suggestions, but for now this seems good.

@bors r+

RalfJung:
> I took this conservative approach that Cargo just stop writing into whatever pointing to sysroot.

So -- it may still be writing into the crates downloaded from the registry? That seems just as problematic as changing std?

weihanglo:
@RalfJung 

Not really. We built a protection for registry caches since 1.57 <https://github.com/rust-lang/cargo/pull/9938>.

PR #13793, opened 2024-04-23
Fix warning suppression for config.toml vs config compat symlinks

### LINKED ISSUE #13667: Make `.cargo/config` deprecation warnings silent
### Problem

I want to be able to run cargo 1.31 and have it respect my `,.cargo/config`.

There seems to be no way to achieve this without causing recent cargo to issue a warning.

### Steps

1. Try: `config` is a file, `config.toml` doesn't exist

```
warning: `/home/rustcargo/.cargo/config` is deprecated in favor of `config.toml`
note: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`
```

(Note that the phrase "symlink X to Y" is ambiguous.  Does it mean make X a symlink to Y, or vice versa?)

1. Try: `config` is a file; `config.toml` is a symlink to it

```
warning: both `/home/rustcargo/.cargo/config` and `/home/rustcargo/.cargo/config.toml` exist. Using `/home/rustcargo/.cargo/config`
```

2. Try `config.toml` is a file, `config` is a symlink to it

```
warning: both `/home/rustcargo/.cargo/config` and `/home/rustcargo/.cargo/config.toml` exist. Using `/home/rustcargo/.cargo/config`
```

### Possible Solution(s)

Check to see if one of the files is a link to the other.  If so, don't issue the warning about both files existing.

The "one file is a link to the other" check should be done with `stat` and comparing `st_dev` and `st_ino`.  (AIUI Windows doesn't have symlinks in the same way so we don't need to do this check there.)

### Notes

This is a blocker for me upgrading the pinned nightly used by CI tests for derive-deftly, because I have UI tests that check for the precise stderr output.  Right now, those tests always produce polluted output in my local environment.

This is happening to me with ` +nightly-2024-03-14`.  I can't easily try a more recent version because my rustup is completely broken due to https://github.com/rust-lang/rustup/issues/3737

### Version

```text
rustcargo@zealot:~/.cargo$ cargo -vV
cargo 1.77.0-beta.5 (837c09f04 2024-02-16)
release: 1.77.0-beta.5
commit-hash: 837c09f04022352c26cf445f37492709922b24ef
commit-date: 2024-02-16
os: Debian 10 (buster) [64-bit]
rustcargo@zealot:~/.cargo$
```

ijackson:

Background: the cargo config file is being renamed from `.cargo/config` to `.cargo/config.toml`.  There's code in new cargo to look for both files (for compatibility), to issue a warning when onliy the old filename is found, and also to issue a warning if both files are found.  The warning suggests making a symlink if compatibility with old cargo is wanted.

An attempt was made to detect when both the old and new files exists, but one is a symlink to the other, but as reported in #13667, this code is not effective.  (It would work only if the symlink had the precise absolute pathname that cargo has decided to use for the lookup, which would be an unnatural way to make the link.)

Logically, the warning should appear when both files exist *but are different*.  That is the anomalous situation that will generate confusing behaviour.   By "different" we ought to mean "aren't the very same file".

That's what this MR implements, where possible.  On Unix, we use the information from stat(2).  That's not available on other platforms; on those, we arrange to also tolerate a symlink referring to precisely `config.toml` as a relative pathname, which is also fine, since by definition the target is then in the same directrory as `config`.

Fixes #13667.

I have interleaved the new tests with the commits that support them.  In each case, a functional commit is followed by a test which fails just beforehand.

(This can be observed by experimentally reordering the branch.)

I have also done ad-hoc testing.

### Additional information

I'm making the assumption that a symlink containing a relative path does something sane on Windows.  This assumption may be unwarranted.  If so, "Handle `config` -> `config.toml` (without full path)" needs to be dropped, and the test case needs to be `#[cfg(unix)]`.

But also, in this case, we should probably put some warnings in the stdlib docs!

### DIFF
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -1542,11 +1542,11 @@ impl GlobalContext {
                 if let Ok(possible_with_extension_handle) =
                     same_file::Handle::from_path(&possible_with_extension)
                 {
-                // We don't want to print a warning if the version
-                // without the extension is just a symlink to the version
-                // WITH an extension, which people may want to do to
-                // support multiple Cargo versions at once and not
-                // get a warning.
+                    // We don't want to print a warning if the version
+                    // without the extension is just a symlink to the version
+                    // WITH an extension, which people may want to do to
+                    // support multiple Cargo versions at once and not
+                    // get a warning.
                     if possible_handle != possible_with_extension_handle {
                         self.shell().warn(format!(
                             "both `{}` and `{}` exist. Using `{}`",
@@ -1556,13 +1556,13 @@ impl GlobalContext {
                         ))?;
                     }
                 } else {
-                        self.shell().warn(format!(
-                            "`{}` is deprecated in favor of `{filename_without_extension}.toml`",
-                            possible.display(),
-                        ))?;
-                        self.shell().note(
-                            format!("if you need to support cargo 1.38 or earlier, you can symlink `{filename_without_extension}` to `{filename_without_extension}.toml`"),
-                        )?;
+                    self.shell().warn(format!(
+                        "`{}` is deprecated in favor of `{filename_without_extension}.toml`",
+                        possible.display(),
+                    ))?;
+                    self.shell().note(
+                        format!("if you need to support cargo 1.38 or earlier, you can symlink `{filename_without_extension}` to `{filename_without_extension}.toml`"),
+                    )?;
                 }
             }
 
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -1537,28 +1537,25 @@ impl GlobalContext {
         let possible = dir.join(filename_without_extension);
         let possible_with_extension = dir.join(format!("{}.toml", filename_without_extension));
 
-        if possible.exists() {
+        if let Ok(possible_handle) = same_file::Handle::from_path(&possible) {
             if warn {
+                if let Ok(possible_with_extension_handle) =
+                    same_file::Handle::from_path(&possible_with_extension)
+                {
                 // We don't want to print a warning if the version
                 // without the extension is just a symlink to the version
                 // WITH an extension, which people may want to do to
                 // support multiple Cargo versions at once and not
                 // get a warning.
-                let skip_warning = if let Ok(target_path) = fs::read_link(&possible) {
-                    target_path == possible_with_extension
-                } else {
-                    false
-                };
-
-                if !skip_warning {
-                    if possible_with_extension.exists() {
+                    if possible_handle != possible_with_extension_handle {
                         self.shell().warn(format!(
                             "both `{}` and `{}` exist. Using `{}`",
                             possible.display(),
                             possible_with_extension.display(),
                             possible.display()
                         ))?;
-                    } else {
+                    }
+                } else {
                         self.shell().warn(format!(
                             "`{}` is deprecated in favor of `{filename_without_extension}.toml`",
                             possible.display(),
@@ -1566,7 +1563,6 @@ impl GlobalContext {
                         self.shell().note(
                             format!("if you need to support cargo 1.38 or earlier, you can symlink `{filename_without_extension}` to `{filename_without_extension}.toml`"),
                         )?;
-                    }
                 }
             }
 
--- a/tests/testsuite/config.rs
+++ b/tests/testsuite/config.rs
@@ -184,7 +184,7 @@ fn symlink_file(target: &Path, link: &Path) -> io::Result<()> {
     os::windows::fs::symlink_file(target, link)
 }
 
-fn symlink_config_to_config_toml() {
+fn make_config_symlink_to_config_toml_absolute() {
     let toml_path = paths::root().join(".cargo/config.toml");
     let symlink_path = paths::root().join(".cargo/config");
     t!(symlink_file(&toml_path, &symlink_path));
@@ -298,7 +298,7 @@ f1 = 1
 ",
     );
 
-    symlink_config_to_config_toml();
+    make_config_symlink_to_config_toml_absolute();
 
     let gctx = new_gctx();
 
--- a/tests/testsuite/config.rs
+++ b/tests/testsuite/config.rs
@@ -195,6 +195,18 @@ fn make_config_symlink_to_config_toml_relative() {
     t!(symlink_file(Path::new("config.toml"), &symlink_path));
 }
 
+fn rename_config_toml_to_config_replacing_with_symlink() {
+    let root = paths::root();
+    t!(fs::rename(
+        root.join(".cargo/config.toml"),
+        root.join(".cargo/config")
+    ));
+    t!(symlink_file(
+        Path::new("config"),
+        &root.join(".cargo/config.toml")
+    ));
+}
+
 #[track_caller]
 pub fn assert_error<E: Borrow<anyhow::Error>>(error: E, msgs: &str) {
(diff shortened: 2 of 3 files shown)


epage:
Please add all test cases in the initial commit, passing.  This should show what the current behavior is.  As the commits are made that change behavior, the test cases will then be updated to pass with the behavior changes.

This has a couple of benefits
- Helps test the tests
- Makes it more clear to anyone reading this what the behavior changes are

tests/testsuite/config.rs
```diff
@@ -309,6 +308,13 @@ f1 = 1
     assert_match("", &output);
 }
 
+#[cargo_test]
+fn config_ambiguous_filename_symlink_doesnt_warn() {
+    config_ambiguous_filename_symlink_doesnt_warn_general(|| {
+        symlink_config_to_config_toml();
+    });
+}
```
epage: nit: personally i would rather duplicate the tests than do this.  This makes it harder to follow what the tests are doing and to evolve them over time.

src/cargo/util/context/mod.rs
```diff
@@ -1546,7 +1546,10 @@ impl GlobalContext {
                 // support multiple Cargo versions at once and not
                 // get a warning.
                 let skip_warning = if let Ok(target_path) = fs::read_link(&possible) {
-                    target_path == possible_with_extension
+                    target_path == possible_with_extension ||
+                        // allow `config` -> `config.toml`, without the parent dir(s),
+                        // which is how you'd naturally do it
+                        target_path == Path::new(&filename_with_extension)
```
epage, line 1575: is there a reason not to just make both paths absolute and compare them?

src/cargo/util/context/mod.rs
```diff
@@ -1544,8 +1545,34 @@ impl GlobalContext {
...
+                // vs absolute pathname, etc.)  Also, this is symmetrical: it doesn't mind
+                // which way round the symlink is.
+                #[cfg(unix)]
+                let skip_warning = if let (Ok(metadata_1), Ok(metadata_2)) = (
+                    fs::metadata(&possible),
+                    fs::metadata(&possible_with_extension),
+                ) {
+                    use std::os::unix::fs::MetadataExt;
+                    let devino = |md: fs::Metadata| (md.dev(), md.ino());
+                    devino(metadata_1) == devino(metadata_2)
```
weihanglo, line 1561: Cargo depends on [`same-file`](https://crates.io/crates/same-file) transitively via `cargo-util` btw. We might want to make use of it instead.

src/cargo/util/context/mod.rs
ijackson: Yes.  That is considerably slower (which may not matter very much, but this happens every time cargo starts up).

Perhaps that could be done in the  non-Unix case, but surely the perf problem is worse on Windows (and symlinks aren't really a thing there anyway).

ijackson:
> Please add all test cases in the initial commit, passing. This should show what the current behavior is. As the commits are made that change behavior, the test cases will then be updated to pass with the behavior changes.

OK.  When should I rebase the branch then?  Normally I wouldn't do that without an OK from the reviewers.

epage:
> OK. When should I rebase the branch then? Normally I wouldn't do that without an OK from the reviewers.

yes, please

src/cargo/util/context/mod.rs
epage: And yes, I had overlooked it, `same-file` should instead be used

src/cargo/util/context/mod.rs
ijackson: Interesting.  I was all ready to say "oh I should use that instead".

But I RTFM and `samefile` is unreliable on Windows.  I don't really have an opinion about that (other than "it is impossible to write reliable software on Windows, anyway").  Perhaps it's OK for this to be unreliable on Windows since the only misbehaviour would be to suppress a warning that ought to have been issued.

I would be happy to change this to use `samefile` everywhere.  The documentation is clear about the algorithm on Unix, and it's a reliable one.

LMK what you prefer.

tests/testsuite/config.rs
ijackson: Gosh.  OK.

ijackson:
OK.  Thanks everyone for the comments.  LMK about the `samefile` thing, and I'll probably provide a revised version tomorrow.

src/cargo/util/context/mod.rs
ijackson: > And yes, I had overlooked it, same-file should instead be used

OK, thanks, willdo.

tests/testsuite/config.rs
ijackson: Now done.

ijackson:
I've revised this and rebased/reordered it, as discussed.

While I was writing the code, I felt dirty writing `if let Ok(...) = `, because that's an ignored error bug.  But I don't think this MR is the right time to address this, so I just added a comment about it.  Ideally we'd change this here, but also  use clippy to forbid use of the buggy `exists` function throughout the codebase.  If folks agree, I could do that as a followup.

ijackson:
Err, I have discovered a mistake.  Give me a moment, will rewind this again.  Sorry.

ijackson:
I think this is ready now.  CI failures are unrelated.

src/cargo/util/context/mod.rs
```diff
@@ -1537,6 +1537,11 @@ impl GlobalContext {
         let possible = dir.join(filename_without_extension);
         let possible_with_extension = dir.join(format!("{}.toml", filename_without_extension));
 
+        // Arguably this is wrong, because it ignores errors.
+        // I.e. it is "error-prone", as discussed in the docs for `std::path::Path::exists`.
+        //
+        // Instead, if we got an error, we should bail, but perhaps
+        // that might be too much risk of being a breaking change.
```
epage, line 1544: imo code comments shouldn't be used for a backlog.  Please pull out this commit

epage: If you have a reproduction case for a bug, feel free to open an Issue.

src/cargo/util/context/mod.rs
```diff
@@ -1537,36 +1537,32 @@ impl GlobalContext {
...
                 // get a warning.
-                let skip_warning = if let Ok(target_path) = fs::read_link(&possible) {
-                    target_path == possible_with_extension
-                } else {
-                    false
-                };
-
-                if !skip_warning {
-                    if possible_with_extension.exists() {
+                    if possible_handle != possible_with_extension_handle {
```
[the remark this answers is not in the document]
ijackson: I did this deliberately.  I think it's fine for either side to be a symlink.  Let me explain:

The reason we are warning is as follows: if the user has both these files, and they are different, then different versions of cargo might behave differently due to reading different files.  That would be quite confusing.

If both files are the same, whichever way round they are symlinked, then this risk doesn't arise.

So insisting on the symlink being made in one particular direction would just be being pernickety for no good reason.  (And also, our current  warning message, inviting the user to make a symlink, is ambiguous, because the phrase "link A to B" might mean "make A -> B" or "make B -> A".)

If we *do* want to insist on the symlink being made in one particular direction (which I think we don't) then we certainly ought to have a bespoke error message for this situation.

We would also then have to decide what we think if *both* files are symlinks referring, ultimately, to the same file, somewhere else.  Currently we're happy with this, as I think we should be.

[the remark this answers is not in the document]
epage: A change in behavior like that should be called out explicitly in an easy-to-discover way in the PR description

PR #13794, opened 2024-04-23
docs(ref): Index differences between virtual / real manifests

epage:

For a user to read the reference and to understand when each type of workspace might be right for them, they have to know to also read the section on Package Selection.

This reframes the section on needing to set `resolver = "2"` to being about differences when there isn't a root package and extends it to summarize a part of Package Selection, linking out to it. The hope is that this will make it all of the differences more discoverable without retreading too much of the same ground within Reference-style documentation.

Part of #13580

### Additional information

r? @weihanglo

### DIFF
--- a/src/doc/src/reference/workspaces.md
+++ b/src/doc/src/reference/workspaces.md
@@ -120,10 +120,8 @@ is not inside a subdirectory of the workspace root.
 In a workspace, package-related Cargo commands like [`cargo build`] can use
 the `-p` / `--package` or `--workspace` command-line flags to determine which
 packages to operate on. If neither of those flags are specified, Cargo will
-use the package in the current working directory. If the current directory is
-a [virtual workspace](#virtual-workspace), it will apply to all members (as if
-`--workspace` were specified on the command-line).  See also
-[`default-members`](#the-default-members-field).
+use the package in the current working directory. However, if the current directory is
+a workspace root, the [`default-members`](#the-default-members-field) will be used.
 
 ## The `default-members` field
 
@@ -136,6 +134,10 @@ used:
 members = ["path/to/member1", "path/to/member2", "path/to/member3/*"]
 default-members = ["path/to/member2", "path/to/member3/foo"]
 ```
+
+When unspecified, the [root package](#root-package) will be used.
+In the case of a [virtual workspace](#virtual-workspace), all members will be used
+(as if `--workspace` were specified on the command-line).
 
 ## The `package` table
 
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/workspaces.md
```diff
@@ -80,10 +80,14 @@ edition = "2021"     # the edition, will have no effect on a resolver used in th
...
-field which is absent in virtual manifests and the edition field of a member
-won't affect the resolver used by the workspace.
+When there isn't a root package:
+
+- [`resolver = "2"`](resolver.md#resolver-versions) must be
+  set explicitly in virtual workspaces as they have no
+  [`package.edition`][package-edition] to infer it from.
+  [resolver version](resolver.md#resolver-versions).
+- Commands run in the workspace's directory will run against all workspace
+  members by default, see [Package selection](#package-selection).
```
weihanglo, line 90: Feel a bit disconnected with the previous paragraph. What about this:

```suggestion
Some major differences bewteen root packages and virtual workspaces:

- In virtual workspaces, [`resolver = "2"`](resolver.md#resolver-versions) must be
  set explicitly as they have no
  [`package.edition`][package-edition] to infer it from.
  [resolver version](resolver.md#resolver-versions).
- Commands run in the a virtual workspace's directory will run against all workspace
  members by default, see [Package selection](#package-selection).
```

src/doc/src/reference/workspaces.md
weihanglo, line 1: There is still something missing. The doc never explains that a root package is effectively a default workspace member.

epage: This was implied in the old "Package selection" wording by saying that the package for the current working directory is used.  However, I think the current wording is up for misinterpretation and I'm tweaking it in hopes that it will be less confusing.

src/doc/src/reference/workspaces.md
epage: I'm not quite seeing the problem.

When I look at the previous paragraph, it ends on the note of the role of virtual workspaces contrasted with root packages

> This is typically useful when there isn't a "primary" package, or
you want to keep all the packages organized in separate directories.

I tweaked the leading line for the new section a little bit which now reads

> By having a workspace without a root package,

Could you help me understand what challenge you are seeing in the transition?

weihanglo: Thanks for that. It feels better now. I guess it is because that there is an extra term “primary package”, so root package is a bit far when mentioned.

weihanglo (approved):

weihanglo:
@bors r+

Nice work!

PR #13797, opened 2024-04-24
Cleanup linting system

Muscraft:
There are a number of problems with the current linting system, most notably that lints could run without `-Zcargo-lints` being set. This PR fixes that issue and a few others that are low-hanging fruit.

### DIFF
--- a/src/cargo/util/lints.rs
+++ b/src/cargo/util/lints.rs
@@ -142,6 +142,61 @@ impl From<TomlLintLevel> for LintLevel {
     }
 }
 
+const IM_A_TEAPOT: Lint = Lint {
+    name: "im_a_teapot",
+    desc: "`im_a_teapot` is specified",
+    groups: &[],
+    default_level: LintLevel::Allow,
+    edition_lint_opts: None,
+};
+
+pub fn check_im_a_teapot(
+    pkg: &Package,
+    path: &Path,
+    lints: &TomlToolLints,
+    error_count: &mut usize,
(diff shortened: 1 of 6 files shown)


src/cargo/util/lints.rs
```diff
@@ -142,6 +142,61 @@ impl From<TomlLintLevel> for LintLevel {
...
+
+pub fn check_im_a_teapot(
+    pkg: &Package,
+    path: &Path,
+    lints: &TomlToolLints,
+    error_count: &mut usize,
+    gctx: &GlobalContext,
+) -> CargoResult<()> {
+    let manifest = pkg.manifest();
+    let lint_level = IM_A_TEAPOT.level(lints, manifest.edition());
```
epage, line 182: `IM_A_TEAPOT` should be an unstable lint that we hide (when it comes to that)

epage:
@bors r+

As this is unstable, IM_A_TEAPOT is non-blocking.  So we dont' forget it, I'll note it in #12235

src/cargo/util/lints.rs
```diff
@@ -79,23 +86,37 @@ pub struct Lint {
...
-                    if edition >= lint_edition {
-                        return lint_level;
-                    }
+                if let Some(level) = edition_level {
+                    level
+                } else {
+                    self.default_level
                 }
-                self.default_level
             }
```
epage, line 120: The implementation for default and edition exists in two places.  We should do this in one place.

PR #13800, opened 2024-04-24
fix(toml): Don't double-warn when underscore is used in workspace dep

epage:

This is prep for removing them in the 2024 Edition and is part of rust-lang/rust#123754 and #13629

Particularly, I wanted to make sure I didn't make things worse and in doing so found there was room for improvement.


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #13802, opened 2024-04-25
refactor(toml): extract dependency-to-source-id to function

weihanglo:

Miscellaneous refactors around extracting dependency-to-source-id to a function.

This is a side effect when trying to rebase #13779.

Should be more readable IMO?

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -1897,15 +1897,15 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
     let new_source_id = match (
         orig.git.as_ref(),
         orig.path.as_ref(),
-        orig.registry.as_ref(),
+        orig.registry.as_deref(),
         orig.registry_index.as_ref(),
     ) {
-        (Some(_), _, Some(_), _) | (Some(_), _, _, Some(_)) => bail!(
+        (Some(_git), _, Some(_registry), _) | (Some(_git), _, _, Some(_registry)) => bail!(
             "dependency ({}) specification is ambiguous. \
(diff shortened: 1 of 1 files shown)


src/cargo/util/toml/mod.rs
```diff
@@ -1810,6 +1810,90 @@ fn detailed_dep_to_dependency<P: ResolveToPath + Clone>(
...
+                "'{}' specifier cannot be used without an 'artifact = …' value ({})",
+                specifier,
+                name_in_toml
+            )
+        }
+    }
+    Ok(dep)
+}
+
+fn resolve_source_id_from_dependency<P: ResolveToPath + Clone>(
```
epage, line 1892: Naming wise, `resolve` is generally being used in this file to "resolve" the original toml to a resolved toml.

For taking toml content and generating a cargo-native type, the `to_` naming scheme is used.  So this could be `to_source_id` or `to_dependency_source_id`.

src/cargo/util/toml/mod.rs
```diff
@@ -1901,20 +1901,17 @@ fn resolve_source_id_from_dependency<P: ResolveToPath + Clone>(
         orig.registry_index.as_ref(),
     ) {
         (Some(_git), _, Some(_registry), _) | (Some(_git), _, _, Some(_registry)) => bail!(
-            "dependency ({}) specification is ambiguous. \
+            "dependency ({name_in_toml}) specification is ambiguous. \
```
epage, line 1904: aside: isn't there a clippy lint we could turn on for this?

weihanglo: Yes, but not sure if we want to turn it on for every place. For here it seems fine.

PR #13805, opened 2024-04-25
Error when unstable lints are specified but not enabled

Muscraft:
In [#13797, it was noted that](https://github.com/rust-lang/cargo/pull/13797#discussion_r1578162057) the `im-a-teapot` lint should always be unstable. This PR makes it so that `im-a-teapot` is unstable, and it is an error to specify it without the `test-dummy-unstable` cargo feature.

It does this by adding a `feature-gate` field to `Lint` and `LintGroup` that optionally
puts the lint/lint group behind a feature. The `feature-gate` is then checked during the new `analyze_cargo_lints_table` step that runs across all lints (and groups) specified in `[lints.cargo]` or `[workspace.lints]` if the package is inheriting its lints from a workspace.

The error looks like the following:
No inherit
No inherit
Inherited
Inhrtited

### DIFF
--- a/src/cargo/core/features.rs
+++ b/src/cargo/core/features.rs
@@ -343,7 +343,7 @@ impl FromStr for Edition {
     }
 }
 
-#[derive(PartialEq)]
+#[derive(Debug, PartialEq)]
 enum Status {
     Stable,
     Unstable,
@@ -387,11 +387,11 @@ macro_rules! features {
             $(
                 $(#[$attr])*
                 #[doc = concat!("\n\n\nSee <https://doc.rust-lang.org/nightly/cargo/", $docs, ">.")]
-                pub fn $feature() -> &'static Feature {
+                pub const fn $feature() -> &'static Feature {
                     fn get(features: &Features) -> bool {
                         stab!($stab) == Status::Stable || features.$feature
                     }
-                    static FEAT: Feature = Feature {
+                    const FEAT: Feature = Feature {
                         name: stringify!($feature),
                         stability: stab!($stab),
                         version: $version,
@@ -512,6 +512,7 @@ features! {
 }
 
 /// Status and metadata for a single unstable feature.
+#[derive(Debug)]
 pub struct Feature {
     /// Feature name. This is valid Rust identifier so no dash only underscore.
     name: &'static str,
--- a/src/cargo/core/features.rs
+++ b/src/cargo/core/features.rs
@@ -406,6 +406,10 @@ macro_rules! features {
             fn is_enabled(&self, features: &Features) -> bool {
                 (self.get)(features)
             }
+
+            pub(crate) fn name(&self) -> &str {
+                self.name
+            }
         }
 
         impl Features {
--- a/src/cargo/util/lints.rs
+++ b/src/cargo/util/lints.rs
@@ -1,6 +1,6 @@
 use crate::core::dependency::DepKind;
 use crate::core::FeatureValue::Dep;
-use crate::core::{Edition, FeatureValue, Package};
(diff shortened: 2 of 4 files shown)


src/cargo/util/lints.rs
```diff
@@ -238,6 +238,15 @@ pub fn check_im_a_teapot(
...
     gctx: &GlobalContext,
 ) -> CargoResult<()> {
+    // This lint is only available when `test-dummy-unstable` is enabled
+    if !pkg
+        .manifest()
+        .unstable_features()
+        .is_enabled(Feature::test_dummy_unstable())
+    {
+        return Ok(());
+    }
```
epage, line 248: Should we error if its in the lints table?

src/cargo/core/features.rs
```diff
@@ -515,7 +515,7 @@ features! {
 #[derive(Debug)]
 pub struct Feature {
     /// Feature name. This is valid Rust identifer so no dash only underscore.
-    name: &'static str,
+    pub name: &'static str,
```
epage, line 518: Should we instead have an accessor?

src/cargo/util/lints.rs
```diff
@@ -247,9 +247,39 @@ pub fn check_im_a_teapot(
     gctx: &GlobalContext,
 ) -> CargoResult<()> {
     let manifest = pkg.manifest();
+    let manifest_path = rel_cwd_manifest_path(path, gctx);
     let (lint_level, reason) = IM_A_TEAPOT.level(pkg_lints, ws_lints, manifest.edition());
 
     if IM_A_TEAPOT.is_feature_gated(manifest) {
+        if reason == LintLevelReason::Package || reason == LintLevelReason::Workspace {
```
epage: Add function to `reason` for this to make it easier to understand and evolve, like `is_user_specified`

src/cargo/util/lints.rs
```diff
@@ -12,6 +12,160 @@ use std::ops::Range;
 use std::path::Path;
 use toml_edit::ImDocument;
 
+const LINTS: &[Lint] = &[IM_A_TEAPOT, IMPLICIT_FEATURES, UNUSED_OPTIONAL_DEPENDENCY];
+
+pub fn verify_lints(
```
epage, line 18: nit: `analyze_lints_table`?

src/cargo/util/lints.rs
```diff
@@ -60,6 +70,33 @@ pub fn verify_lints(
...
 }
 
+fn find_lint_or_group<'a>(
+    name: &str,
+) -> Option<(
+    &'static str,
+    &LintLevel,
+    &Option<(Edition, LintLevel)>,
+    &Option<&'static Feature>,
+)> {
```
epage, line 80: Thats a bit large of a tuple with non-obvious positional meaning.  Normally this would be changed into a struct  or a trait but its unclear if another struct or trait is worth it.

epage: Because its only called in one place, its probably not worth fixing

src/cargo/util/lints.rs
```diff
@@ -249,10 +405,6 @@ pub fn check_im_a_teapot(
     let manifest = pkg.manifest();
     let (lint_level, reason) = IM_A_TEAPOT.level(pkg_lints, ws_lints, manifest.edition());
 
-    if IM_A_TEAPOT.is_feature_gated(manifest) {
-        return Ok(());
-    }
```
epage: If I'm understanding this, unstable lints run, we just also error.  Is that right?

Should we have the "allow" logic also force these as "allow", causing them to be bypassed automatically for every unstable lint?

PR #13806, opened 2024-04-25
Strip cargo lints table

Muscraft:
`[lints.cargo]` should be removed when publishing/packaging to avoid unintentional MSRV bumps, and this PR makes it so the table is removed.


epage:
I was wondering if we should just completely strip the table on publish.

This got me thinking, is there a downside to doing it completely?  If so, is there a downside to doing it here?

Maybe if `forbid` is used but that won't work anyways.

If we wanted to include it, we could make it based on your `package.rust-version`.

Muscraft:
#13833 supersedes this

Issue #6790, opened 2019-03-28
`--artifact-dir` (nee `--out-dir`) Tracking Issue


ehuss:
Note: this was renamed from `--out-dir` to `--artifact-dir` in #13809

Original issue: #4875
Implementation PR: #5203
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#artifact-dir
Issues: https://github.com/rust-lang/cargo/labels/Z-out-dir

**Summary**
Adds `--artifact-dir=PATH` flag to `cargo build` to specify a directory to place final artifacts.

**Tasks:**
- [x] #5203
- [x] #13809
- [ ] Remove `--out-dir`
- [ ] Switch `--artifact-dir` to being the direct we uplift directly into with a fallback to the current location (done across all commands including `cargo package`)
- [ ] Evaluate recording requested uplifts and performing them at the end
- [ ] Evaluate only creating `examples/` if needed (helps if we know a priori what is uplifted)
- [ ] Detect conflicts in uplifts due to platform-tuple not being included
- [ ] *(design needed)* explore templating of artifact-dir
- [ ] Rework to better match https://github.com/rust-lang/cargo/issues/14125#issuecomment-2258708917

**Unresolved questions**
- [ ] Multi-target builds (`--target <one> --target <two>`)
- [ ] #7493
- [x] #6100 — concern about similarity with `--target-dir`, proliferation of flags
- [x] #6313 — concern about name collisions in workspaces or with examples
- [x] What exactly is the use case for this?  In what scenario would it be used where you don't already have an outer build system or script that can copy the files into the correct location?
- [x] Confusion with rustc's `--out-dir` (#13809)
- [x] Discrepancy or confusion with `build.rs`s `OUT_DIR` (depends on if `build.rs` is generating intermediate or final artifacts) (#13809)

**Future possibilities**
- Deprecate `--target-dir`, hiding it from the CLI help

ultrasaurus:
I find this feature very helpful, since I get output in a directory named like this `./target/debug/build/mylib-79d5eadf36eadf53` and with a build.rs file there are two similarly named directories. 

However, it doesn't seem that OUT_DIR environment variable is set to the value of `--out-dir` when build.rs is executed.  Is that intentional?  is there some other way to discover the value of `--out-dir`?

haze:
[...]

Issue #6100, opened 2018-09-26
Is `--out-dir` justified or can we find another way of accomplishing it?


withoutboats:
I noticed that `cargo build` has two subcommands, both added this year. I believe this accurately describes their behavior:

* `--target-dir`: Replaces `$PWD/target` with this directory (e.g. a debug build will be built into `target-dir/debug`). Stable flag, added in April 2018. #5393 #5308. Mimicks the behavior of `CARGO_TARGET_DIR`
* `--out-dir`: Things will be built into `$PWD/target` as normal, but copies some of the artifacts into the directory specified by `out-dir` (not a profile specific subdirectory). Unstable flag, added in March 2018.  #5203  #4875. Mimicks the behavior of `OUT_DIR`.

(I'm not 100% certain I've accurately described their behavior in relationship to the profile specific subdirectories.)

--out-dir is still unstable, but --target-dir was instantly stabilized (by mistake, I believe).

I can't help but notice that these two flags do very similar things! They also seem to have been designed for very similar purposes (build integration, the meson project is mentioned specifically in both of them).

I'm not certain I agree that either of them should have been added as flags (an argument is made that flags are better than env vars, and while I agree in principle for *user oriented* APIs, I don't think that build system integration is a *user oriented* API in the relevant sense), but I'm definitely suspicious of having both of these very similar flags provided. Since target-dir is stable, I'm not sure if there's an available avenue to unify these flags other than just only providing target-dir.

cc @rust-lang/cargo

joshtriplett:
I *do* think it makes sense to keep these separate. I like the idea of having `--out-dir` to put the final build artifacts and not all the intermediate bits.

matklad:
I think it is important that we have both of these features:

* ability to override "scratch space" dir for intermediate build artifacts 
* ability to override the place where the final build results go

UI-wise, using flags for these seems better: the "build system" code, like ci-scripts, tends to be poked by many people, so it's important to keep it easy to understand without prior knowledge, and flags are better than env vars in this respect. 

I do agree that for most of the users, interacting with Cargo's CLI, these flags are irrelevant. However, this could be said about many other flags, like `--manifest-path`, `--message-format`, `--build-plan`, `--frozen`, `--locked`. Perhaps we should hide such "useful in very specific circumstances" flags from `cargo cmd --help` by default?

withoutboats:
[...]

PR #5203, opened 2018-03-19
cargo build --out-dir

matklad:
This is intended to fix https://github.com/rust-lang/cargo/issues/4875. Very much work in progress, just to figure out how exactly `--out-dir` should work :) 

The path of least resistance for implementing `out-dir` would be to just link everything from `target/debug` to the `out-dir`. However, the problem with this approach is that we link *too much* to the `target/debug`. For example, if you run `cargo build --bin foo`, you'll find not only the `foo` itself there, but also rlibs from the same workspace. 

I think it's rather important not to copy extra stuff to out-dir, so it is implemented as a completely new parameter, which is threaded through various config structs and applies *only* to the top-level units (i.e, to the stuff user explicitly requested to compile on the command line).

I also plan to add platform specific tests here, to check that we get .pdb on windows and .dSYM on macs for various crate-types. 

Because, in general, a single target may produce multiple files, `out-dir` has to be a directory, you can't directly specify the output *file*. 

Note that artifacts are still generated to the `target` dir as well. 

Also cc @nirbheek, I hope that this might be useful for Meson integration, because `--out-dir` should be more straightforward to use than `--message-format=json`. 

The end result, for `cargo build --bin rustraytracer --out-dir out` on windows looks like this:

image

Note how we have fewer files in the `out` :) 

r? @alexcrichton

Issue #4875, opened 2017-12-29
Ability to specify output artifact name


matklad:
Hi!

I'd like to suggest that we should add `--output-path` option to certain Cargo commands, to make it easier to use Cargo as a part of the larger build system. 

Currently, the way to get output artifacts location is to run `cargo build --message-format=json` and parse the resulting JSON. This is very convenient if you want to "enhance" the stuff that Cargo does. For example, IntelliJ Rust uses this feature to learn what binaries it should launch via debugger. 

However, this does not work well for the case where you use Cargo as a part of larger build, and are interested only in the artifact. The main drawback here is that parsing the JSON is not that easy. 

Here some implementation considerations:

* Cargo should still put all stuff  into `target` as usual, `--output-path` is only a copy at the end
* The flag works only when the build produces a single artifact. That is, you'll need `--bin name` sort flags for disambiguation. 
* The flag definitely should work for `cdylib`, `clib` and `bin` crate types. Not sure if we'd want to support rlibs.
* It's not clear what to do with extensions. Do we add `.exe` on windows automatically? Or do we use the path as-is? I think the letter is preferable. 
* It's not clear what to do if the artifact consists of several files (i.e, debug info on Macs is in a separate file). Should we just place it alongside? Should we add more config?

Issue inspired by this discussion on Gitter: https://gitter.im/rust-lang/rust?at=5a468c4b0163b028108fe9b6

PR #13809, opened 2024-04-26
Rename --out-dir to --artifact-dir

valadaptive:
Progress towards unblocking #6790. Renames the experimental `--out-dir` argument to `--artifact-dir`, both to reflect that it's where the final build *artifacts* will be copied to, and to avoid confusion with the `OUT_DIR` environment variable which serves an entirely different purpose.

For transition purposes, `--out-dir` argument and `out-dir` config key will still work with a deprecation message encouraging the use of the new arg and config key.

### Rationale

A lot of people seem to be confused by the naming of the `--out-dir` argument, and are misled into thinking it serves the same purpose as the `OUT_DIR` environment variable:

> [However, it doesn't seem that OUT_DIR environment variable is set to the value of --out-dir when build.rs is executed.](https://github.com/rust-lang/cargo/issues/6790#issuecomment-574754790)

> [I understand that the worry is that there could be confusion between --out-dir for cargo and the environment variable OUT_DIR for build.rs, but doesn't it mean exactly the same in both cases?](https://github.com/rust-lang/cargo/issues/6790#issuecomment-2091259652)

> [--out-dir: Things will be built into $PWD/target as normal, but copies some of the artifacts into the directory specified by out-dir (not a profile specific subdirectory). Unstable flag, added in March 2018. cargo build --out-dir #5203 Ability to specify output artifact name #4875. **Mimicks the behavior of OUT_DIR.**](https://github.com/rust-lang/cargo/issues/6100#issue-364115234)

> [I recently had a couple of people express an interest in --out-dir being stabilized and from my initial digging it seems like what they may actually want is to switch to OUT_DIR, which is already stable.](https://github.com/rust-lang/cargo/issues/6100#issuecomment-820783316)


workingjubilee:
Grim news. Perhaps both flags must be briefly supported to deal with the bootstrap problem.

valadaptive:
Never mind; it seems I was a bit zealous in replacing `--out-dir` and accidentally search-and-replaced it in some rustc invocations. Might still need to do some fixups but we're looking a lot better now.

valadaptive:
This should be ready for review now. Not sure if this needs to go through any sort of design process since it's renaming a nightly-only API. There does seem to be some consensus in #6790 that `--artifact-dir` is a better name.

valadaptive:
@epage Just wanted to make sure this hasn't fallen off your radar--it's ready for review now.

epage:
I have about a month backlog of github notifications that I still need to respond to.

epage:
Could you update the commits for how you'd want them presented for review and merging?

src/bin/cargo/commands/build.rs
```diff
@@ -50,14 +50,45 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
...
-    if let Some(out_dir) = args.value_of_path("out-dir", gctx) {
-        compile_opts.build_config.export_dir = Some(out_dir);
-    } else if let Some(out_dir) = gctx.build_config()?.out_dir.as_ref() {
-        let out_dir = out_dir.resolve_path(gctx);
-        compile_opts.build_config.export_dir = Some(out_dir);
+    // If the user specifies `--artifact-dir`, use that
+    let mut artifact_dir = args.value_of_path("artifact-dir", gctx);
+
+    // `--out-dir` is deprecated, but still supported for now
+    if artifact_dir.is_none() {
```
epage, line 57: Shouldn't the `if/else-if` approach still work?  Keeping the structure would make it easier to read and this embeds higher level context into the code structure (makes it clear to readers that they are mutually exclusive)

epage: You could even combine the approaches and do
```rust
compile_opts.build_config.export_dir = if let Some(artifact_dir) = ...
};
```

valadaptive: I've changed it back to the old code style.

The if-expression style may not be equivalent to the existing code--right now `compile_opts.build_config.export_dir` is preserved if none of the command-line flags or config flags are present, whereas the if-expression would unconditionally reassign (in the final `else`, we *could* just assign it to its existing value, but that feels too "clever"). I don't *think* `compile_opts.build_config.export_dir` is set by any other code, but I'd rather not risk it.

valadaptive:
I've updated the flag-checking code and squashed most of the commits.

Should the deprecation messages provide some sort of warning that the old options will eventually be removed, or are they fine currently?

tests/testsuite/artifact_dir.rs
```diff
@@ -304,6 +304,55 @@ For more information, try '--help'.
         .run();
 }
 
+#[cargo_test]
+fn deprecated_out_dir() {
```
epage, line 308: nit: not important enough for this change but in the future, I generally recommend adding the tests as  the first commit (with them passing).  Then as you change behavior, that is reflected in how the tests are changed to ensure they still pass.  This makes it easier for reviewers to see how behavior changed, we can then show those diffs in posts to the community for them to understand changes, and it helps "test the test" (I've made changes where I thought I was testing the right condition and wasn't).

PR #13811, opened 2024-04-26
Update SleepTraker returns_in_order unit test

ehuss:
This updates the `returns_in_order` SleepTracker unit test so that it is not so sensitive to how fast the system is running. Previously it assumed that the function calls would take less than a millisecond to finish, but that is not a valid assumption for a slow-running system.

I have changed it to simplify the test, with the assumption that it takes less than 30 seconds for it to run, which should have a safety margin of a few orders of magnitude.


weihanglo (approved):
Thanks!

So [this failed](https://github.com/rust-lang/rust/pull/124416#issuecomment-2080187478) because arm linux is too fast!?

Issue #9339, opened 2021-04-09
Add cargo watch


Zerotask:
**Describe the problem you are trying to solve**

During development you are very often in the process of making changes and then build or run your project or run your tests. This always costs some time, therefore an automation would be useful, to save some time and increase the developer experience

**Describe the solution you'd like**

I'd like to have a built-in watcher similar to nodemon for NodeJS or like this project https://github.com/passcod/cargo-watch
This would allow you to run your whole project if you change something or you can also specify to run only tests when your files change:
`cargo watch -x run`
`cargo watch -x test`

Since this is completely optional to use, developers are not forced to use the suggested watcher.

**Notes**

sr2ds:
@Zerotask, I usage the cargo-watch package and is like nodemon. Why do you think that is necessary include this on cargo if the package cargo-watch solve this problem?

When I need make any tests without cargo, I run on homemade script like this:

```
while inotifywait -e close_write *; do rustc main.rs; done
```

Sorry if I did not understand correctly your request.

Zerotask:
Yes, I also mentioned this package and the reason why I think it should be added to the default toolchain of cargo.

If a command is often used, one could consider to add it to the default toolchain.

sr2ds:
Yes! I understand your wish now and this make sense. I use every day cargo watch too.

weihanglo:
Just note that Cargos has a function [`list_files`] respecting `package.include` and `package.exclude`. This might be worth a reuse than inventing a brand new file listing mechanism. There are also [change detection instructions] in build scripts we need to consider. O

To be fair, I don't know any details of cargo-watch. [epage's suggestion] seems like a great starting point to me.

[`list_files`]: https://github.com/rust-lang/cargo/blob/777960615afe01ca3a822398e645027f2aeb2ce3/src/cargo/sources/path.rs#L100
[change detection instructions]: https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection
[epage's suggestion]: https://internals.rust-lang.org/t/cargo-run-watch-flag-to-auto-recompile/17971/4?u=weihanglo

epage:
Another thing I'd add is to study multiple forms of prior art for features, pitfalls, etc so and report on them and what you suggest we should apply from them.

We do this when adding commands but I suspect a `--watch` flag has enough little pieces of policy to be worth it as well.

epage:
While this might only be implemented a bit at at time, we should consider what all commands should support this.

[...]

PR #13813, opened 2024-04-27
refactor: Move diagnostic printing to Shell

Muscraft:
As I have been working on cargo's diagnostic system, I have disliked copying around the code for a new `Renderer`, and then emitting the diagnostic:
```rust
let renderer = Renderer::styled().term_width(
    gctx.shell()
        .err_width()
        .diagnostic_terminal_width()
        .unwrap_or(annotate_snippets::renderer::DEFAULT_TERM_WIDTH),
);
writeln!(gctx.shell().err(), "{}", renderer.render(message))?;
```

Moving to a method on `Shell` helps mitigate the risk of updating duplicate code and missing one. It should also make it nearly impossible to get into scenarios where `gctx.shell()` is borrowed twice, leading to panics. This problem was one I ran into early on as I tried to write messages to `stderr` like so:
```rust
writeln!(
    gctx.shell().err(),
    "{}",
    Renderer::styled()
        .term_width(
            gctx.shell()
                .err_width()
                .diagnostic_terminal_width()
                .unwrap_or(annotate_snippets::renderer::DEFAULT_TERM_WIDTH),
        )
        .render(message)
)?;
```

### DIFF
--- a/src/cargo/core/shell.rs
+++ b/src/cargo/core/shell.rs
@@ -2,6 +2,7 @@ use std::fmt;
 use std::io::prelude::*;
 use std::io::IsTerminal;
 
+use annotate_snippets::{Message, Renderer};
 use anstream::AutoStream;
 use anstyle::Style;
 
@@ -390,6 +391,19 @@ impl Shell {
         // ... but don't fail due to a closed pipe.
         drop(writeln!(self.out(), "{}", encoded));
         Ok(())
+    }
+
+    /// Prints the passed in [Message] to stderr
+    pub fn print_message(&mut self, message: Message<'_>) -> std::io::Result<()> {
+        let term_width = self
+            .err_width()
+            .diagnostic_terminal_width()
+            .unwrap_or(annotate_snippets::renderer::DEFAULT_TERM_WIDTH);
+        writeln!(
+            self.err(),
+            "{}",
+            Renderer::styled().term_width(term_width).render(message)
+        )
     }
 }
 
--- a/src/cargo/util/lints.rs
+++ b/src/cargo/util/lints.rs
@@ -3,7 +3,7 @@ use crate::core::FeatureValue::Dep;
 use crate::core::{Edition, FeatureValue, Package};
 use crate::util::interning::InternedString;
 use crate::{CargoResult, GlobalContext};
-use annotate_snippets::{Level, Renderer, Snippet};
+use annotate_snippets::{Level, Snippet};
 use cargo_util_schemas::manifest::{TomlLintLevel, TomlToolLints};
 use pathdiff::diff_paths;
 use std::collections::HashSet;
@@ -270,13 +270,8 @@ pub fn check_im_a_teapot(
                     .fold(true),
             )
             .footer(Level::Note.title(&emitted_reason));
-        let renderer = Renderer::styled().term_width(
-            gctx.shell()
-                .err_width()
-                .diagnostic_terminal_width()
-                .unwrap_or(annotate_snippets::renderer::DEFAULT_TERM_WIDTH),
-        );
-        writeln!(gctx.shell().err(), "{}", renderer.render(message))?;
+
+        gctx.shell().print_message(message)?;
     }
     Ok(())
 }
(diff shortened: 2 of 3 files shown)


weihanglo (commented):
Overall looks good to me, especially for the prevention of double borrowing `RefCell`!

However, I wonder if we need diagnostic with to be computed every time before Cargo prints messages. I guess the answer is no, we don't care people resize their terminal width during a build.

epage:
I'm a bit uncomfortable generating this in `Context::new` because it can be quite easy for us to add config for controlling this (honestly, I'm surprised it doesn't exist).

As for the terminal width changing, that is more likely to come up when it comes to `cargo watch` support (#9339).

So far, a `Renderer` is "cheap".  We could instead make the renderer on every call.

src/cargo/util/context/mod.rs
```diff
@@ -412,6 +413,22 @@ impl GlobalContext {
         self.shell.borrow_mut()
     }
 
+    /// Prints the passed in [Message] to stderr
+    pub fn renderer_diagnostic(&self, message: Message<'_>) -> std::io::Result<()> {
```
weihanglo, line 417: Any reason that this method is on `GlobalContext` instead of `Shell`?

Muscraft: None that I can think of, I went ahead and moved it to `Shell`

epage: One of the problems with #12875 is that it put diagnostic tracking in `Shell`.  For me, I see `Shell` as lower level than that.  I feel like putting diagnotic rendering in `Shell` could encourage that kind of thinking.

Muscraft: I agree diagnostic tracking should not be in `Shell`, but I am not as worried about encouraging that thinking with this change. This change moves to a function where things get rendered and then emitted, similar to: https://github.com/rust-lang/cargo/blob/ba6fb405284df654362275405175f3fac2b79dbf/src/cargo/core/shell.rs#L387-L393

The JSON gets "rendered" by `serde_json::to_string` and then printed to `stdout`.

The fact `Renderer` gets created inside of `emit_diagnostic` is an implementation detail.

src/cargo/core/shell.rs
```diff
@@ -1,3 +1,4 @@
+use annotate_snippets::{Message, Renderer};
```
weihanglo, line 1: nit: we don't really have a style convention for imports, though I believe if this was added by auto-import via Rust Analyzer, it would be grouped together with `anstream` and `anstyle.

weihanglo (approved):

src/cargo/util/context/mod.rs
weihanglo: @epage do you consider it as a blocker? I am fine with either way.

src/cargo/core/shell.rs
```diff
@@ -391,6 +392,19 @@ impl Shell {
         drop(writeln!(self.out(), "{}", encoded));
         Ok(())
     }
+
+    /// Prints the passed in [Message] to stderr
+    pub fn emit_diagnostic(&mut self, message: Message<'_>) -> std::io::Result<()> {
```
epage, line 397: Should this be `emit_message`?  This is lower level than a diagnostic.

src/cargo/core/shell.rs
Muscraft: Nice catch! Normally, things get added correctly; I'm not sure why it didn't in this case.

 I wish [`group_imports`](https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=import#group_imports) and [`imports_granularity`](https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=import#imports_granularity) were stable so we could use them.

src/cargo/util/context/mod.rs
weihanglo: Maybe we could put some design principles in the module-level for shell, so at least this discussion won't just disappear in the air.

epage: The current naming is low level enough that I'm fine with moving this forward

weihanglo:
@bors r+

Thank you all :)

PR #13817, opened 2024-04-28
Fix target entry in .gitignore

rukai:
The directories at:
* https://github.com/rust-lang/cargo/tree/master/tests/testsuite/cargo_remove/target
* https://github.com/rust-lang/cargo/tree/master/tests/testsuite/cargo_add/target

Are getting picked up by the `target` entry in .gitignore, causing git to ignore any extra files added to those directories.
A minor issue I know, but I still think its worth fixing.

## Alternative

An alternative solution would be to rename those directories which I'm more than happy to do.
But I think we are better off just changing .gitignore as that is more future proof.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,11 @@
-target
-Cargo.lock
+/target
+/Cargo.lock
 /config.stamp
 /Makefile
 /config.mk
-src/doc/build
-src/etc/*.pyc
-src/registry/target
+/src/doc/build
+/src/etc/*.pyc
+/src/registry/target
 rustc
 __pycache__
 .idea/


weihanglo (approved):

.gitignore
```diff
@@ -1,4 +1,4 @@
-target
+/target
 Cargo.lock
```
weihanglo, line 2: Could you also add a `/` prefix to `Cargo.lock`? I believe it has the same issue. Thank you :)

```console
$ fd -HI Cargo.lock
Cargo.lock
tests/testsuite/cargo_add/locked_unchanged/in/Cargo.lock
tests/testsuite/cargo_add/lockfile_updated/in/Cargo.lock
tests/testsuite/cargo_add/lockfile_updated/out/Cargo.lock
tests/testsuite/cargo_remove/gc_patch/out/Cargo.lock
tests/testsuite/cargo_remove/update_lock_file/in/Cargo.lock
tests/testsuite/cargo_remove/update_lock_file/out/Cargo.lock
```

rukai: done!
I also added it to the `src/*` entries since they all seem to be referring to the root as well.

weihanglo (approved):
Feel like some ignore entries are no longer necessary, but anyway we can look at them later.

Thanks!

Issue #5931, opened 2018-08-23
Per-user compiled artifact cache


djc:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/issues/5931)*

MVP design: cache intermediate artifacts across projects with minimal cache poisoning
- Only immutable items (registry, git, not path)
- Idempotent builds (no build script executions + dependents, no proc macro dependents)

Roadmap:
- [x] Split out a `build-dir` to communicate out early what will be broken (#14125)
- [x] Re-organize `build-dir` into cacheable units (#15010)
- [ ] Experiments
  - [ ] More granular locking in `build-dir` to see what could apply to a shared cache (#4282)
  - [ ] GC for `build-dir` in prep for GC for shared cache (#5026)
  - [ ] Implement shared cache without gc and with coarse locking
- [ ] Polish shared cache
  - [ ] GC
  - [ ] granular locking

Parallel efforts to improve cache efficiency
- #14948
- #14903
- rust-lang/rust#143549
- rust-lang/rust#143547

Future possibilities
- Remote cache plugin system
- Means for proc-macros and build scripts to be cached
  - Pinky promise flag to be cacheable
  - Means for guarenteeing they are only accessing declared resources

Open questions
- How does this interact with `cargo clean`?
- How does this interact with pipelined builds?

---

I was wondering if anyone has contemplated somehow sharing compiled crates. If I have a number of projects on disk that often have similar dependencies, I'm spending a lot of time recompiling the same packages. (Even correcting for features, compiler flags and compilation profiles.) Would it make sense to store symlinks in `~/.cargo` or equivalent pointing to compiled artefacts?

alexcrichton:
There's been musings about this historically but never any degree of serious consideration. I've always wanted to explore it though! (I think it's definitely plausible)

aidanhs:
[sccache](https://github.com/mozilla/sccache) is one option here - it has a local disk cache in addition to the more exotic options to store compiled artifacts in the cloud.

djc:
sccache would be good for the compilation time part, but it'd be nice to also get a handle on the disk size part of it.

Eh2406:
cc https://github.com/rust-lang/cargo/issues/6229

Vlad-Shcherbina:
I think you can put
```
[build]
target-dir = "/my/shared/target/dir"
```
in `~/.cargo/config`.

But I have no idea if this mode is officially supported. Is it?

[...]

PR #13818, opened 2024-04-28
Clean package perf improvements

osiewicz:

I've noticed that `cargo clean -p` execution time scales poorly with size of target directory; in my case (~250GB target directory on M1 Mac) running `cargo clean -p` takes circa 35 seconds. Notably, it's the file listing that takes that time, not deleting the package itself. That is, when running `cargo clean -p SOME_PACKAGE` twice in a row, both executions take roughly the same time.

I've tracked it down to the fact that we seem quite happy to use `glob::glob` function, which iterates over contents of target dir. It also was a bit sub-optimal when it came to doing that, for which I've already filled a PR in https://github.com/rust-lang/glob/pull/144 - that PR alone takes down cleaning time down to ~14 seconds. While it is a good improvement for a relatively straightforward change, this PR tries to take it even further. With glob PR applied + changes from this PR, my test case goes down to ~6 seconds. I'm pretty sure that we could squeeze this further, but I'd rather do so in a follow-up PR. 

Notably, this PR doesn't help with *just* super-large target directories. `cargo clean -p serde` on cargo repo (with ~7Gb target directory size) went down from ~380ms to ~100ms for me. Not too shabby.

I've mostly tested it manually, running `cargo clean` against multiple different repos.

### Additional information
TODO:
- [x] [c770700](https://github.com/rust-lang/cargo/commit/c770700885d486b5bf7d360afb269f90c50780db) is not quite correct; we need to consider that it changes how progress reporting works; as is, we're gonna report all progress relatively quickly and stall at the end (when we're actually iterating over directories, globbing, removing files, that kind of jazz). I'll address that.


epage:
#5931 will require re-working the target directory to be organized around package names.  Depending on the complexity this entails, I wonder if its better to wait on that.  Granted, my bias is that I never use `cargo clean -p` and instead do `cargo clean`.

osiewicz:
Ah, I see; thanks for pointing that issue out. That's definitely another good argument for not squeezing every last bit out of `cargo clean -p` if it's gonna be reworked in near-ish term.
The only pending item I want to address is that TODO for progress reporting; I expect this PR to be ready for review tomorrow-ish.

weihanglo (commented):
Thanks for the contribution!

weihanglo (approved):
Thanks!

PR #13820, opened 2024-04-28
[doc]:fix project to package

heisen-li:

As `[project]` is being [phased out](https://github.com/rust-lang/cargo/pull/13747), the filename should be consistent with the title of the file content.


heisen-li:
I'm going to check to see if the link jump is affected and set it to draft for now.

weihanglo (commented):
I forgot how things actually work, but we need to ensure that every old link won't be dead and will redirect to the new one. Should check both page and fragment redirections.

heisen-li:
> I forgot how things actually work, but we need to ensure that every old link won't be dead and will redirect to the new one. Should check both page and fragment redirections.

I'm just talking about my own knowledge that there is a check for document linking in CI, but I don't know if it includes a check for fragment redirection.

https://raw.githubusercontent.com/rust-lang/rust/master/src/tools/linkchecker/linkcheck.sh

weihanglo:
We need to make sure the redirection work. For example, <https://doc.rust-lang.org/nightly/cargo/guide/project-layout.html> must redirect to <https://doc.rust-lang.org/nightly/cargo/guide/package-layout.html>.

weihanglo:
Speaking of that. This only affects URLs. It is not really a thing user preceive, so might not be worth having this churn.

heisen-li:
> Speaking of that. This only affects URLs. It is not really a thing user preceive, so might not be worth having this churn.

Of course, this is not relevant to the big picture. I was just browsing the article and noticed this.

The content of the URL doesn't match the content of the document, I just think it should be kept consistent. If this is difficult or otherwise, then feel free to turn it off.

Thank you. @weihanglo

weihanglo:
mdbook supports [`[output.html.redirect]`](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#outputhtmlredirect), though it doesn't seem to support fragment jumps after redirections. We then need to add additional JavaScript like this:

https://github.com/rust-lang/cargo/blob/05364cb2f61a2c2b091e061c1f42b207dfb5f81f/src/doc/src/reference/manifest.md?plain=1#L654-L694

I feel like there are more works with less consistency gain, so lean toward closing. What do you think?

heisen-li:
Ok, no problem there. In the future, it would be nice to be able to check the validity of the link jump in the CI.

PR #13826, opened 2024-04-29
ci: exclude semver check for `cargo` against referenced commit

weihanglo:

We always bump major versions for `cargo` package,
so the risk of shipping breaking change should be zero.

We still verify against crates.io as a sanity check.

Check if CI is all green.

### Additional information

See also https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/What.20is.20the.20reason.20for.20this.20error.3F/near/436053158


weihanglo:
Close in favor of #13827.

PR #13830, opened 2024-04-29
fix(toml): Improve granularity of traces

epage:

Once we move resolving out of `to_targets`, we won't have tracing visibility.  As we don't have a top-level function for resolving, I put it on each. I mirrored this for "to_" for consistency and it seems useful when looking at the output.


weihanglo (approved):

weihanglo:
Those instruments make sense to me. Thanks.

@bors r+

PR #13833, opened 2024-04-30
fix(lint): Warn not Error on unsupported lint tool

Muscraft:
In a recent Cargo Team meeting, it was decided to lessen the error on an unsupported tool in `[lints]` to a warning. This PR implements that change.

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -2231,8 +2231,13 @@ fn verify_lints(
     for (tool, lints) in lints {
         let supported = ["cargo", "clippy", "rust", "rustdoc"];
         if !supported.contains(&tool.as_str()) {
-            let supported = supported.join(", ");
-            anyhow::bail!("unsupported `{tool}` in `[lints]`, must be one of {supported}")
+            let message = format!(
+                "unrecognized lint tool `lints.{tool}`, specifying unrecognized tools may break in the future.
+supported tools: {}",
+                supported.join(", "),
+            );
+            warnings.push(message);
+            continue;
         }
         if tool == "cargo" && !gctx.cli_unstable().cargo_lints {
             warn_for_cargo_lint_feature(gctx, warnings);
--- a/tests/testsuite/lints_table.rs
+++ b/tests/testsuite/lints_table.rs
@@ -105,13 +105,12 @@ fn fail_on_invalid_tool() {
         .build();
 
     foo.cargo("check")
-        .with_status(101)
         .with_stderr(
             "\
-[..]
-
-Caused by:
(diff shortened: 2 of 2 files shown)


src/cargo/util/toml/mod.rs
```diff
@@ -2231,8 +2231,8 @@ fn verify_lints(
     for (tool, lints) in lints {
         let supported = ["cargo", "clippy", "rust", "rustdoc"];
         if !supported.contains(&tool.as_str()) {
-            let supported = supported.join(", ");
-            anyhow::bail!("unsupported `{tool}` in `[lints]`, must be one of {supported}")
+            warnings.push(format!("unused manifest key `lints.{tool}`",));
```
epage, line 2234: The request in the meeting was to specifically call out that this is reserved syntax and may break in the future.  From that, we should probably be more specific than unused key but unused lint tool info

src/cargo/util/toml/mod.rs
```diff
@@ -2231,8 +2231,15 @@ fn verify_lints(
...
         let supported = ["cargo", "clippy", "rust", "rustdoc"];
         if !supported.contains(&tool.as_str()) {
-            let supported = supported.join(", ");
-            anyhow::bail!("unsupported `{tool}` in `[lints]`, must be one of {supported}")
+            let message = format!(
+                "unrecognized lint tool `lints.{tool}`
+
+specifying an unrecognized lint tool could lead to future breakage, as `cargo`
+reserves all of the `lints` namespace for itself and provides no compatibility
+guarantees when encountering unrecognized tools"
```
Muscraft, line 2239: I also considered:
```
`cargo` reserves all of the `lints` namespace for itself and provides no
future compatibility guarantees when encountering unrecognized tools, as they
may be supported in the future.
```

weihanglo: I prefer this one.

Should we also suggest using `package.metadata`?

epage: I feel like this makes it sound like the lints table is special and this doesn't apply elsewhere.

Something to keep in mind is people might also be here from a typo.

Muscraft: I think this should be the top level message:
```
unrecognized lint tool `{}`, must be one of {supported}
```

What about:
```
no compatibility guarantees are provided when specifying unrecognized manifest
key(s), as the entire namespace is reserved for future use. Tools that wish to
store package configuration should use `package.metadata`
```
or 
```
no compatibility guarantees are provided when specifying unrecognized manifest
key(s) as `cargo` reserves the entire namespace for future use. `package.metadata`
should be used for tools which would like to store package configuration
```

epage: I'm good with the now updated wording

PR #13836, opened 2024-05-01
fix(toml): Don't lose 'public' when inheriting a dep 

epage:

When inheriting a simple dep, we preserved the package dep's `public`
field but lost it when inheriting a detailed dep.

This was done by consolidating the code paths.  This also reduces the
risk of us doing this again in the future.

I didn't write tests for this specific case because I refactored the root cause away.


Muscraft (approved):
Thanks!

PR #13837, opened 2024-05-01
fix(lints): Remove ability to specify `-` in lint name

Muscraft:
In a recent Cargo Team meeting, it was discussed whether our lint should use `-` or `_` and if we should rewrite the wrong form to the correct one. It was decided that Cargo would use `_` for lint names and would not convert `-` to `_` automatically; instead, we would warn about an "unknown lint" and mention the similarly named lint with `_`, if found.

The decision to ise `_` was made because it is the canonical representation, as well as [RFC #344](https://rust-lang.github.io/rfcs/0344-conventions-galore.html#lints) specifies:
> Use snake case in the same way you would for function names.

This PR implements these changes.

Note: This adds an `unknown_lints` lint, that tries to mirror [the lint `rustc` has with the same name](https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#unknown-lints).

### DIFF
--- a/src/cargo/util/lints.rs
+++ b/src/cargo/util/lints.rs
@@ -29,45 +29,68 @@ pub fn analyze_cargo_lints_table(
     let manifest = pkg.manifest();
     let manifest_path = rel_cwd_manifest_path(path, gctx);
     let ws_path = rel_cwd_manifest_path(ws_path, gctx);
-
-    for lint_name in pkg_lints
+    let mut unknown_lints = Vec::new();
+    for (lint_name, specified_in) in pkg_lints
         .keys()
-        .chain(ws_lints.map(|l| l.keys()).unwrap_or_default())
+        .map(|name| (name, SpecifiedIn::Package))
+        .chain(
+            ws_lints
+                .map(|l| l.keys())
+                .unwrap_or_default()
+                .map(|name| (name, SpecifiedIn::Workspace)),
+        )
     {
-        if let Some((name, default_level, edition_lint_opts, feature_gate)) =
+        let Some((name, default_level, edition_lint_opts, feature_gate)) =
             find_lint_or_group(lint_name)
-        {
-            let (_, reason, _) = level_priority(
+        else {
+            unknown_lints.push((lint_name, specified_in));
+            continue;
+        };
+
+        let (_, reason, _) = level_priority(
+            name,
+            *default_level,
+            *edition_lint_opts,
+            pkg_lints,
+            ws_lints,
+            manifest.edition(),
+        );
+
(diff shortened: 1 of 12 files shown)


src/cargo/util/lints.rs
```diff
@@ -117,31 +117,21 @@ fn verify_feature_enabled(
     gctx: &GlobalContext,
 ) -> CargoResult<()> {
     if !manifest.unstable_features().is_enabled(feature_gate) {
-        let dash_name = lint_name.replace("_", "-");
         let dash_feature_name = feature_gate.name().replace("_", "-");
-        let title = format!("use of unstable lint `{}`", dash_name);
+        let title = format!("use of unstable lint `{}`", lint_name);
```
weihanglo, line 144: nit: implicit arg capture (should we force that in clippy?)

```suggestion
        let title = format!("use of unstable lint `{lint_name}`");
```

src/cargo/util/lints.rs
```diff
@@ -573,6 +601,120 @@ pub fn check_implicit_features(
     Ok(())
 }
 
+const UNKNOWN_LINTS: Lint = Lint {
```
weihanglo, line 604: You know. I'll request for more doc comments like what `IMPLICIT_FEATURES` has :)

src/cargo/util/lints.rs
```diff
@@ -588,6 +601,120 @@ pub fn check_implicit_features(
...
+        }
+        let title = format!("{}: `{lint_name}`", UNKNOWN_LINTS.desc);
+        let second_title = format!("`cargo::{}` was inherited", lint_name);
+        let underscore_lint_name = lint_name.replace("-", "_");
+        let matching = if let Some(lint) = LINTS.iter().find(|l| l.name == underscore_lint_name) {
+            Some((lint.name, "lint"))
+        } else if let Some(group) = LINT_GROUPS.iter().find(|g| g.name == underscore_lint_name) {
+            Some((group.name, "group"))
+        } else {
+            None
```
weihanglo, line 649: Besides dash-replaced, should we also leverage `util::edit_distance::closest` to find one more possible suggestion?

src/cargo/util/lints.rs
epage: If its something we bring up in code review but could be automated, then yes

with the caveat on how often we don't want it which I believe is pretty rare for this one.

src/cargo/util/lints.rs
epage: We should.  I saw that as a future improvement.  The minimum we needed in this case was to help with dropping support for `-`.

PR #11409, opened 2022-11-22
fix(toml): Add `default-features` to `TomlWorkspaceDependency`

Muscraft:
In #11329 it was noted that `default-features` is ignored when used in a dependency that inherits from a workspace i.e. 
```toml
[workspace]
members = []
[workspace.dependencies]
dep = "0.1"

[package]
name = "bar"
version = "0.2.0"
authors = []
[dependencies]
dep = { workspace = true, default-features = false }
```

This problem is caused by problems with deserializing a `TomlDependency` and not emitting an unused manifest key correctly. When discussed in a recent Cargo team meeting we felt the best course of action was to allow `default-features = false` when inheriting a dependency, but it does not change actually set `default-features`. It will be used to warn when there is a difference between the definition in `[workspace.dependencies]` and `[dependencies]` i.e.
```toml
[package]
name = "bar"
version = "0.2.0"
[dependencies]
dep = { workspace = true, default-features = false }

[workspace]
members = []
[workspace.dependencies]
dep = { version = "0.1", default-features = true }
```

This does not entirely resolve the problem with unused manifest keys. A follow up PR and issue will be created to address those concerns.

close #11329

PR #13839, opened 2024-05-01
fix(toml): On 2024 Edition, disallow ignored `default-features` when inheriting

epage:

This is part of rust-lang/rust#123754

This is a follow up to #11409 which tweaked how we do inheritance of default-features, including warning when `default-features = false` is ignored.  This turns those warnings into an error.


epage:
Migration is now implemented.

Muscraft (approved):
This looks great!

PR #13840, opened 2024-05-01
chore(ci): Ignore openssl deps

epage:
We excluded the packages in #13731 but that just means they fell into the default logic, rather than being ignored (see #13835).  This at least made it easier to reject the change.

This should prevent the PR from being created.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -8,6 +8,8 @@
   ignorePaths: [
     '**/tests/**',
   ],
+  // See rust-lang/cargo#13546 and openssl/openssl#23376 for the exclusion
+  ignoreDeps: ['openssl', 'openssl-src', 'openssl-sys'],
   customManagers: [
     {
       customType: 'regex',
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -8,6 +8,8 @@
   ignorePaths: [
     '**/tests/**',
   ],
+  // See rust-lang/cargo#13546 and openssl/openssl#23376 for the exclusion
+  ignoreDeps: ['openssl', 'openssl-src', 'openssl-sys'],
```
weihanglo, line 12: For reference: <https://docs.renovatebot.com/configuration-options/#ignoredeps>.

I am fine merging this as-is. 

Do you think we should remove the exact requirement from `openssl` crate?

weihanglo (approved):

.github/renovate.json5
epage: > Do you think we should remove the exact requirement from openssl crate?

Probably worth keeping to avoid accidental upgrades (e.g. changing another dep causes it to be upgraded) as the manifest has a bit more visibility

weihanglo:
@bors r+

Thanks!

PR #13842, opened 2024-05-02
Fix: Build only the specified artifact library when multiple types are available

### LINKED ISSUE #12109: artifact-dependencies seems to build both cdylib and staticlib when only one is specified
### Problem

When an artifact dependency specifies `artifact='staticlib'`, and the dependency declares its crate type as `["staticlib", "cdylib"]`, cargo will build both the staticlib and cdylib separately. I would expect it to only build the staticlib.

### Steps

Create a project with an artifact dependency:

```toml
[package]
name = "foo"
version = "0.1.0"
edition = "2021"

[dependencies]
bar = {path = "bar", artifact = "staticlib"}
```

and in the bar dependency, specify it with multiple crate types:

```toml
[package]
name = "bar"
version = "0.1.0"
edition = "2021"

[lib]
crate-type = ["staticlib", "cdylib"]
```

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
cargo 1.71.0-nightly (569b648b5 2023-05-05)
release: 1.71.0-nightly
commit-hash: 569b648b5831ae8a515e90c80843a5287c3304ef
commit-date: 2023-05-05
host: aarch64-apple-darwin
libgit2: 1.6.3 (sys:0.17.0 vendored)
libcurl: 7.87.0 (sys:0.4.61+curl-8.0.1 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 13.3.1 [64-bit]
```

skogseth:
Fixes #12109. 

#### TL;DR:
A crate `bar` exposes it's library as both a `staticlib` and a `cdylib`. A second crate `foo` specifies an artifact dependency of type `staticlib` on `bar`, meaning cargo should build `bar` as a `staticlib` and expose certain environment variables (e.g. `CARGO_STATICLIB_FILE_BAR`). However, due to a bug, cargo ends up building (and exposing the associated environment variables) for both the `staticlib` and `cdylib`. 

The same happens if `foo` specifies an artifact dependency of type `cdylib`; both artifact types are built.

The first commit introduces a test which reproduces the issue, the second commit introduces the fix. This setup was recommended by @ehuss.

### Additional information
Artifact dependencies: https://rust-lang.github.io/rfcs/3028-cargo-binary-dependencies.html

#### TL;DR
If a crate `foo` requests an artifact dependency of kind <artifact_kind> from a crate `bar` then the following happens:

- `bar` is built with crate-type <artifact_kind> and a directory is created at `target/<profile>/deps/artifact/bar-<build_hash_or_something>/<artifact_kind>`. The binary artifact is placed in this directory.
- Cargo exposes certain environment variables, the most important for this PR is `CARGO_<artifact_kind>_FILE_BAR`, which points to the binary artifact that was specified. This environment variable is available at compile time for normal dependencies and at runtime for build-dependencies.

If multiple artifact-kinds are requested cargo will create a unit for each, and so they will all be built separately.

### DIFF
--- a/src/cargo/core/compiler/unit_dependencies.rs
+++ b/src/cargo/core/compiler/unit_dependencies.rs
@@ -24,7 +24,7 @@ use crate::core::compiler::unit_graph::{UnitDep, UnitGraph};
 use crate::core::compiler::{
     CompileKind, CompileMode, CrateType, RustcTargetData, Unit, UnitInterner,
 };
-use crate::core::dependency::{Artifact, ArtifactTarget, DepKind};
+use crate::core::dependency::{Artifact, ArtifactKind, ArtifactTarget, DepKind};
 use crate::core::profiles::{Profile, Profiles, UnitFor};
 use crate::core::resolver::features::{FeaturesFor, ResolvedFeatures};
 use crate::core::resolver::Resolve;
@@ -555,17 +555,20 @@ fn artifact_targets_to_unit_deps(
     let ret =
         match_artifacts_kind_with_targets(dep, artifact_pkg.targets(), parent.pkg.name().as_str())?
             .into_iter()
-            .map(|(_artifact_kind, target)| target)
-            .flat_map(|target| {
+            .flat_map(|(artifact_kind, target)| {
                 // We split target libraries into individual units, even though rustc is able
-                // to produce multiple kinds in an single invocation for the sole reason that
+                // to produce multiple kinds in a single invocation for the sole reason that
                 // each artifact kind has its own output directory, something we can't easily
                 // teach rustc for now.
                 match target.kind() {
                     TargetKind::Lib(kinds) => Box::new(
                         kinds
                             .iter()
-                            .filter(|tk| matches!(tk, CrateType::Cdylib | CrateType::Staticlib))
+                            .filter(move |tk| match (tk, artifact_kind) {
+                                (CrateType::Cdylib, ArtifactKind::Cdylib) => true,
+                                (CrateType::Staticlib, ArtifactKind::Staticlib) => true,
+                                _ => false,
+                            })
                             .map(|target_kind| {
                                 new_unit_dep(
                                     state,
--- a/tests/testsuite/artifact_dep.rs
+++ b/tests/testsuite/artifact_dep.rs
@@ -3189,3 +3189,77 @@ fn check_transitive_artifact_dependency_with_different_target() {
         .with_status(101)
         .run();
 }
+
+#[cargo_test]
+fn build_only_specified_artifact_library() {
+    // Create a project with:
+    // - A crate `bar` with both `staticlib` and `cdylib` as crate-types.
+    // - A crate `foo` which depends on either the `staticlib` or `cdylib` artifact of bar,
+    //   whose build-script simply checks which library artifacts are present.
+    let create_project = |artifact_lib| {
+        project()
+            .file(
+                "bar/Cargo.toml",
+                r#"
+                [package]
+                name = "bar"
+                version = "1.0.0"
+
+                [lib]
+                crate-type = ["staticlib", "cdylib"]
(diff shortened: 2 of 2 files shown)


tests/testsuite/artifact_dep.rs
```diff
@@ -3189,3 +3189,77 @@ fn check_transitive_artifact_dependency_with_different_target() {
...
+            .build()
+    };
+
+    let cdylib = create_project("cdylib");
+    cdylib
+        .cargo("build -Z bindeps")
+        .masquerade_as_nightly_cargo(&["bindeps"])
+        .run();
+    match_exact(
+        "cdylib present: true\nstaticlib present: false",
```
weihanglo, line 3244: > Just make sure it has a test, preferably as the first commit and then include the fix in the second commit to show how the test changes.

So this should be

```suggestion
        "cdylib present: true\nstaticlib present: true",
```

in the first commit. And the next commit would fix both the behavior and this test. This setup is better because:

* Every commit passes CI pipeline.
* Changes of the test shows we really fixes the behavior.

tests/testsuite/artifact_dep.rs
```diff
@@ -3189,3 +3189,77 @@ fn check_transitive_artifact_dependency_with_different_target() {
...
+        None,
+    )
+    .unwrap();
+
+    let staticlib = create_project("staticlib");
+    staticlib
+        .cargo("build -Z bindeps")
+        .masquerade_as_nightly_cargo(&["bindeps"])
+        .run();
+    match_exact(
```
weihanglo, line 3257: Not sure why this test module chose to use match_exact but not  `cargo_test_support::compare::assert_match_exact`. Anyway, I'll find a time clean this up. For now we just follow what has been written :)

tests/testsuite/artifact_dep.rs
```diff
@@ -3189,3 +3189,77 @@ fn check_transitive_artifact_dependency_with_different_target() {
...
+        "cdylib present: false\nstaticlib present: true",
```
weihanglo, line 3258: ditto

```suggestion
        "cdylib present: true\nstaticlib present: true",
```

src/cargo/core/compiler/unit_dependencies.rs
```diff
@@ -555,17 +555,20 @@ fn artifact_targets_to_unit_deps(
...
                 match target.kind() {
                     TargetKind::Lib(kinds) => Box::new(
                         kinds
                             .iter()
-                            .filter(|tk| matches!(tk, CrateType::Cdylib | CrateType::Staticlib))
+                            .filter(|tk| match tk {
+                                CrateType::Cdylib => *artifact_kind == ArtifactKind::Cdylib,
+                                CrateType::Staticlib => *artifact_kind == ArtifactKind::Staticlib,
+                                _ => false,
+                            })
```
weihanglo, line 571: What about this? 

(just personal taste, not too important)

```suggestion
                            .filter(move |tk| match (tk, artifact_kind) {
                                (CrateType::Cdylib, ArtifactKind::Cdylib) => true,
                                (CrateType::Staticlib, ArtifactKind::Staticlib) => true,
                                _ => false,
                            })
```

weihanglo (commented):
Overall looks good! I added some minor suggestions.

Thanks for your contribution!

tests/testsuite/artifact_dep.rs
skogseth: Ahh, that makes a lot of sense!

src/cargo/core/compiler/unit_dependencies.rs
skogseth: Ooohh, I like it

weihanglo (approved):

weihanglo:
Thank you for the patch!

@bors r+

Issue #10662, opened 2022-05-12
Make `cargo --list` print description for custom subcommands


dtolnay:
### Problem

Currently `cargo --list` prints descriptions only for the builtin subcommands. For third-party subcommands like `cargo expand` it does not print a description.

```console
$ cargo --list
Installed Commands:
    …
    doc                  Build a package's documentation
    expand
    fetch                Fetch dependencies of a package from the network
    …
```

### Proposed Solution

First of all, I don't want `cargo --list` to **run** any of the binaries (in order to, for example, run `--help` on them and parse the output, or some other handshake to cause the binary to print out a description of itself). Arbitrary binaries in the PATH are not necessarily safe to run. An arbitrary binary might completely ignore `--help` or any other handshake being passed by Cargo, and immediately do unwanted stuff instead, which would be unexpected and undesirable to the person running `cargo --list`.

Instead, Cargo subcommands should be able to embed a description into their compiled binary, which can then be directly retrieved from the binary by Cargo in a way that does not involve running it.

The way this is exposed to subcommands can be as simple as:

```rust
// subcommand's main.rs

cargo_subcommand_metadata::description!("…");

fn main() {
    /* … */
}
```

The underlying implementation would need to be platform-specific based on the file format used for executables on the platform. On platforms that use ELF executables, I would propose that this uses a Note. On those platforms the macro shown above would expand to:

```rust
#[used]
#[link_section = ".note.cargo.subcommand"]
static ELF_NOTE: ElfNote = {
    …
}
```

This note can then be retrieved efficiently from the ELF executable during `cargo --list` by using the [`object`](https://github.com/gimli-rs/object) crate to parse the executable's program header.

From testing on my machine with the `cargo-expand` crate, `cargo build --release` produces a 13 MB executable, and retrieving the subcommand description ELF note from it using the `object` crate takes 0.03 milliseconds, so this should be plenty fast enough for `cargo --list`.

We do not necessarily need to implement support for a large number of platforms immediately up front. Supporting just ELF would already benefit a large fraction of Cargo users.

### Notes

Yes, exactly.

dtolnay:
As proof of concept, I have released a version of the `cargo-expand` crate containing a suitable ELF note (https://github.com/dtolnay/cargo-expand/pull/142) and created a draft PR #10663 which makes `cargo --list` retrieve the subcommand description from that ELF note.

[...]

PR #13847, opened 2024-05-02
Show description for user-installed binaries when running "cargo --list"

Oakchris1955:

I've noticed that when running `cargo --list`, user-installed subcommands wouldn't have a description. In order to resolve this, cargo will now communicate with the [crates.io] registry (if those packages were installed from [crates.io] ofc) and retrieve the package description from there. This also works if the binary name is different than the name of the package (for example, if a package installs multiple binaries)

 I've noticed that some test use the `--list` flag, so if they pass, this should be fine

### Additional information

[crates.io] allows someone to go to `https://crates.io/api/v1/crates/{crate}` and get info about this crate in a JSON format. However, from what I've found, this isn't standardized for other Rust registries, and so, if a subcommand was installed from another registry, it won't print a description. Please verify that this is the case. If not, I can remove the relevant checks for other registries.

**IMPORTANT**

I haven't fixed the relevant tests *yet*, so **DON'T** merge this immediately

[crates.io]: https://crates.io

### DIFF
--- a/crates/crates-io/lib.rs
+++ b/crates/crates-io/lib.rs
@@ -128,6 +128,11 @@ struct Users {
     users: Vec<User>,
 }
 #[derive(Deserialize)]
+struct CrateInfo {
+    #[serde(alias = "crate")]
+    crate_info: Crate,
+}
+#[derive(Deserialize)]
 struct TotalCrates {
     total: u32,
 }
@@ -262,6 +267,22 @@ impl Registry {
         Ok(serde_json::from_str::<Users>(&body)?.users)
     }
 
+    pub fn description(&mut self, krate: &str) -> Result<Option<String>> {
+        // Looks like this isn't a standalized method for getting packages' descriptions
+        // so it will only be used if the current registry host is crates.io
+        if self.host_is_crates_io() {
+            self.handle.get(true)?;
+            let body = self
+                .req(&format!("/crates/{}", krate), None, Auth::Unauthorized)
+                .unwrap();
+            Ok(serde_json::from_str::<CrateInfo>(&body)?
+                .crate_info
+                .description)
+        } else {
+            Ok(None)
+        }
+    }
+
     pub fn publish(&mut self, krate: &NewCrate, mut tarball: &File) -> Result<Warnings> {
         let json = serde_json::to_string(krate)?;
         // Prepare the body. The format of the upload request is:
@@ -392,6 +413,8 @@ impl Registry {
         if body.is_some() {
             headers.append("Content-Type: application/json")?;
         }
+
+        headers.append(&format!("User-Agent: cargo-cli"))?;
 
         if self.auth_required || authorized == Auth::Authorized {
             headers.append(&format!("Authorization: {}", self.token()?))?;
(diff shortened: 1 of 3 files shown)


crates/crates-io/lib.rs
```diff
@@ -393,6 +414,8 @@ impl Registry {
             headers.append("Content-Type: application/json")?;
         }
 
+        headers.append(&format!("User-Agent: cargo-cli"))?;
```
weihanglo, line 417: This is not ideal. Cargo has its own [user-agent header](https://doc.rust-lang.org/nightly/cargo/reference/registry-web-api.html?highlight=user-agent#web-api) when communicating with crates.io. This kind of setting should be in `cargo` main crate, not `crates-io`. Also there is `curl::easy::Easy::useragent` you could use. `

src/bin/cargo/cli.rs
```diff
@@ -180,14 +210,27 @@ fn print_list(gctx: &GlobalContext, is_verbose: bool) {
...
-                    drop_println!(gctx, "    {literal}{name}{literal:#}");
+                    drop_print!(gctx, "    {literal}{name:<20}{literal:#}");
+
+                    // The .unwrap() in the path below will probably never panic, since we are dealing with files here
+                    let installed_by = user_installed
+                        .get(&path.file_name().unwrap().to_string_lossy().to_string())
+                        .cloned();
+
+                    if let Some(installed_by) = installed_by {
+                        let desc = registry.description(&installed_by).ok().flatten();
```
weihanglo, line 222: Personally, I feel like it's overkill talking over the network for descriptions. Yet, let's back to design discussion first.

weihanglo (commented):
Thanks for the contribution.

There is an issue tracking this <https://github.com/rust-lang/cargo/issues/10662>. Per [the contributing guide](https://doc.crates.io/contrib/process/index.html), we'd like to have a design discussion before diving into implementation details.

crates/crates-io/lib.rs
Oakchris1955: Yeah, I did some searching around and found that in `cargo::util::network::http` exists a function named `http_handle` that autoconfigures things like the user-agent, etc. based on the `gctx` (GloablContext) variable. I assume it is ok to use that?

src/bin/cargo/cli.rs
Oakchris1955: Querying the network for such little things is overkill indeed. However, from what I've found,currently no other better way to do this exists.
Perhaps what could be done here is implement some form of cache to store binaries' descriptions so that we don't have to query them each time.

Oakchris1955:
I might have found why those tests are failing. For some reason, by making those changes to the code, `cargo --list` now generates trailing whitespaces on every end of line of subcommands without a description, even though I couldn't narrow down why this happens. Any help would be appreciated

weihanglo:
I am not going to discuss any specific implementation further until we have a conclusion on the design. Could we bring it back to <https://github.com/rust-lang/cargo/issues/10662>, and gather ideas there?

ehuss:
I'm going to close per the comment about about not having a chosen design, yet. Additionally, I don't think we will want to do network requests in the `--list` command. Thanks for the contribution, though!

PR #13849, opened 2024-05-02
perf(toml): Avoid inferring when targets are known

epage:

We read the file system to infer two different data points
- Implicit targets
- Implicit `path` values for targets

I took a shortcut for this case and recognize the scenario where we can
bypass both and do so.
I went with a bypass, rather than this being integrating into the
inferring code because the inferring code is complex and I didn't want
to add to it further in isolating the inferring to only when its needed.

The validation gets duplicated because having it in the middle of the resolve code provides a better user experience and it would be messy to add the conditionals to get all of that working.  I at least worked to keep the duplicated validation close to each other.

### DIFF
--- a/src/cargo/util/toml/targets.rs
+++ b/src/cargo/util/toml/targets.rs
@@ -261,48 +261,60 @@ pub fn resolve_bins(
     errors: &mut Vec<String>,
     has_lib: bool,
 ) -> CargoResult<Vec<TomlBinTarget>> {
-    let inferred = inferred_bins(package_root, package_name);
+    if is_resolved(toml_bins, autodiscover) {
+        let toml_bins = toml_bins.cloned().unwrap_or_default();
+        for bin in &toml_bins {
+            validate_bin_name(bin, warnings)?;
+            validate_bin_crate_types(bin, edition, warnings, errors)?;
+            validate_bin_proc_macro(bin, edition, warnings, errors)?;
+        }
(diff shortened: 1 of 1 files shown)


weihanglo (approved):

weihanglo:
Looks pretty neat. Thanks!

@bors r+

src/cargo/util/toml/targets.rs
```diff
@@ -536,11 +548,24 @@ fn to_bench_targets(
...
+        return false;
+    }
+
+    let Some(toml_targets) = toml_targets else {
+        return true;
+    };
+    toml_targets
+        .iter()
+        .all(|t| t.name.is_some() && t.path.is_some())
+}
```
Turbo87, line 562: does this mean that if in the `Cargo.toml` (whether normalized or not) all `[[bin]]` entries contain a `name` and `path` field then the auto-discovery of additional bins is disabled? that seems to have potential for false positives and would not match https://doc.rust-lang.org/cargo/reference/cargo-targets.html#target-auto-discovery. last time this behavior was changed it was done at an edition boundary 🤔

epage: We only get here if autodiscover is explicitly disabled. We reuse autodiscover logis is reused to fill in `name` or `path` and this is checking to see if that is needed.

If there is a behavior change you noticed, could you be more explicit about what it is?

PR #13850, opened 2024-05-02
Update continuous-integration.md: Include CircleCI reference

eddiewebb:

Adds simple example to run `cargo test` on CircleCI, alongside the other popular CI engines.

Change is purely documentation. I would ask that the PR be checked for tone and formatting that is in line with Rust/Cargo community standards.


epage:
Unsure what the bar should be for inclusion but if we have sr.ht then it seems fine to include CircleCI.  At a later point, I could see us creating more guidance on what gets included which might shake things up.

@bors r+

Issue #13843, opened 2024-05-02
[regression] `cargo vendor` keeps workspace inheritance for some things


glandium:
### Problem

With 1.77.2, a vendored Cargo.toml might contain something like:
```
[lints.clippy.pedantic]
level = "warn"
priority = -1
```
instead of what's in the original crate (vendored via git):
```
[lints]
workspace = true
```

with 1.79.0-beta.1, you see what's in the original crate.

It happens with nightly-2024-04-28 but not nightly-2024-04-27, which would put the regression in the range c9392675917..b60a1555155. Edit: fixed the versions.

Presumably this could affect `cargo publish` too.

### Steps

1. `cargo new testcase; cd testcase`
2. add the following dependency to `Cargo.toml`: `neqo-common = { tag = "v0.7.5", git = "https://github.com/mozilla/neqo" }`
3. `cargo vendor vendoring`
4. Look at the file `vendoring/neqo-common/Cargo.toml`

glandium:
Looks like it could be a regression from d5bc35d84454fd4653b5ffe546e86363803b0a1b

epage:
@Muscraft unsure how I overlooked this.

Our options
- Inherit this during `prepare_for_publish`
- Don't differentiate between workspace and package source
- Use `original_toml` like you had proposed

As this is in beta, we will need to fix this soonish and do a beta backport.  Unsure how bad this will be for merge conflicts.  I wonder if we should take a very hacky route just for beta (like revert the change) to keep risk down.

weihanglo:
* Beta backport is just merged <https://github.com/rust-lang/rust/pull/124647>. We'll have this fix in the next beta release tomorrow.
* I'll do a nightly update for <https://github.com/rust-lang/cargo/pull/13852> later today.

I believe this is resolved, so close. Thanks people.

PR #13801, opened 2024-04-24
Add where lint was set

Muscraft:
`rustc` and `clippy` both show why the lint was emitted and where the level was set the first time it was emitted for a package. We already showed why the list was being emitted but did not show where the lint level was set. This PR adds where the lint was set at.

PR #13852, opened 2024-05-02
fix(lints): Prevent inheritance from bring exposed for published packages

Muscraft:
#13843 demonstrated a regression caused by #13801, where we started to keep `[lints]` and `[workspace.lints]` separate, and not truly resolve `[lints]`. This was a nice thing to have and made it easier to tell when a lint came from a workspace. The downside of doing so is the lints table would not get resolved when vendoring or publishing. 

To fix this issue, I reverted the change for keeping `[lints]` and `[workspace.lints]` separate and modified how cargo's linting system figures out where a lint is coming from. Due to this change, we no longer specify that a lint was set by `[workspace.lints]`, only `[lints]`. It is true that a lint level is set by `[lints]` always, as it would've had to specify the lint outright or specify that it was inheriting it, seeing that, I do not think this is a regression in diagnostic quality. I still manage to keep the ability to render a lint's location in the workspace's manifest when running ` analyze_cargo_lints_table`, which I am pleased about.

### DIFF
--- a/src/cargo/util/lints.rs
+++ b/src/cargo/util/lints.rs
@@ -19,7 +19,6 @@ pub fn analyze_cargo_lints_table(
     pkg: &Package,
     path: &Path,
     pkg_lints: &TomlToolLints,
-    ws_lints: Option<&TomlToolLints>,
     ws_contents: &str,
     ws_document: &ImDocument<String>,
     ws_path: &Path,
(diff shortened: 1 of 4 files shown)


src/cargo/util/lints.rs
```diff
@@ -151,55 +137,50 @@ fn verify_feature_enabled(
...
+                        .fold(true),
+                )
+                .footer(Level::Help.title(&help))
+        } else {
+            let lint_span = get_span(
+                ws_document,
+                &["workspace", "lints", "cargo", lint_name],
+                false,
+            )
+            .unwrap();
```
epage, line 159: Thoughts on this unwrap?

src/cargo/util/lints.rs
```diff
@@ -651,50 +603,45 @@ fn output_unknown_lints(
...
+                    .annotation(Level::Error.span(span))
+                    .fold(true),
+            )
+        } else {
+            let lint_span = get_span(
+                ws_document,
+                &["workspace", "lints", "cargo", lint_name],
+                false,
+            )
+            .unwrap();
```
epage, line 621: ditto

epage (approved):
Feel free to r= me when done

weihanglo:
Are we gonna backport this to beta?

Muscraft:
Yes, I plan to backport this once it is merged

PR #13861, opened 2024-05-04
fix(toml): Remove unstable rejrected frontmatter syntax for cargo script

epage:

With rust-lang/rfcs#3503 approved, we no longer need to allow easy, high fidelity experiments with alternative cargo script syntax.

### Additional information

We still need to improve the experience for users writing bad syntax but that can come later.


ehuss:
Thanks!

@bors r+

Eh2406:
@epage looks like the documentation did not get updated for this change https://github.com/rust-lang/cargo/blob/7297f0f06f46e500881edc0e72e1925128fcf96e/src/doc/src/reference/unstable.md?plain=1#L1242

PR #8248, opened 2020-05-16
Provide better error messages for a bad `patch`.

ehuss:
This attempts to provide more user-friendly error messages for some situations with a bad `patch`.  This is a follow-up to #8243.

I think this more or less covers all the issues from #4678.  I imagine there are other corner cases, but those will need to wait for another day. The main one I can think of is when the patch location is missing required features.  Today you get a "blah was not used in the crate graph." warning, with some suggestions added in #6470, but it doesn't actually check if there is a feature mismatch.

Closes #4678

PR #13863, opened 2024-05-04
Don't squash path dependencies with same package name

### LINKED ISSUE #13862: Unnecessary index fetching upon patching two dependencies with same package name
### Problem

If you have two patches on two dependencies with same package name, you will see `Updating crates.io index` every time you run cargo check. The index update is unnecessary and slows down development speed. 

### Steps

Simple reproduction repo: 
https://github.com/ldm0/cargo-bug-repo/tree/31fb297e58ffa3be877062940a300883d2cb580a

1. Run cargo check
2. Run cargo check (you will still see `Updating crates.io index`)

### Possible Solution(s)

I'll file a fix PR later.

### Notes

_No response_

### Version

```text
Can't find a cargo version which doesn't have this issue.
```

ldm0:
Fixes #13862

Dependencies with same package name in the `Cargo.lock` was squashed in `EncodableResolve::into_resolve()`.

e.g. In the problem reproduction repo's Cargo.lock, there are two entries:

```toml
[[package]]
name = "once_cell"
version = "0.1.0"

[[package]]
name = "once_cell"
version = "1.19.0"
```

However, they were squashed in `build_path_deps()`, then we endup getting two package ids with same path like this:

```
PackageId { name: "once_cell", version: "0.1.0", source: "pwd/vendor/once_cell_1_19" }
PackageId { name: "once_cell", version: "1.19.0", source: "pwd/vendor/once_cell_1_19" }
```

This package version inconsistency triggers patch update introduced in https://github.com/rust-lang/cargo/pull/8248 everytime you compile this crate.

### DIFF
--- a/src/cargo/core/resolver/encode.rs
+++ b/src/cargo/core/resolver/encode.rs
@@ -154,7 +154,7 @@ impl EncodableResolve {
     /// primary uses is to be used with `resolve_with_previous` to guide the
     /// resolver to create a complete Resolve.
     pub fn into_resolve(self, original: &str, ws: &Workspace<'_>) -> CargoResult<Resolve> {
-        let path_deps = build_path_deps(ws)?;
+        let path_deps = PathDeps::from_workspace(ws)?;
         let mut checksums = HashMap::new();
 
         let mut version = match self.version {
@@ -202,14 +202,18 @@ impl EncodableResolve {
                 if !all_pkgs.insert(enc_id.clone()) {
                     anyhow::bail!("package `{}` is specified twice in the lockfile", pkg.name);
                 }
-                let id = match pkg.source.as_deref().or_else(|| path_deps.get(&pkg.name)) {
+                let id = match pkg.source.as_deref().copied().or_else(|| {
+                    path_deps
+                        .get_path_dep(&pkg.name, &pkg.version)
+                        .map(|package_id| package_id.source_id())
+                }) {
                     // We failed to find a local package in the workspace.
                     // It must have been removed and should be ignored.
                     None => {
                         debug!("path dependency now missing {} v{}", pkg.name, pkg.version);
                         continue;
                     }
-                    Some(&source) => PackageId::try_new(&pkg.name, &pkg.version, source)?,
+                    Some(source) => PackageId::try_new(&pkg.name, &pkg.version, source)?,
                 };
 
                 // If a package has a checksum listed directly on it then record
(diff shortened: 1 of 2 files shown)


src/cargo/core/resolver/encode.rs
```diff
@@ -408,68 +416,98 @@ impl EncodableResolve {
...
-    }
 
-    fn build_dep(
-        dep: &Dependency,
-        ws: &Workspace<'_>,
-        ret: &mut HashMap<String, SourceId>,
-        visited: &mut HashSet<SourceId>,
-    ) {
-        let id = dep.source_id();
-        if visited.contains(&id) || !id.is_path() {
```
ldm0, line 462: `!id.is_path()` this was removed, since non-path dependencies were filtered in the beginning of `from_workspace()`

src/cargo/core/resolver/encode.rs
```diff
@@ -408,68 +416,98 @@ impl EncodableResolve {
...
+    /// If no exact version match, pickup a random version for patch updating trial
+    fn get_path_dep(&self, name: &str, version: &str) -> Option<&PackageId> {
+        let pkg_version = semver::Version::parse(version).ok()?;
+        self.path_deps.get(name).and_then(|versions| {
+            versions
+                .iter()
+                .find(|package_id| {
+                    package_id.version().cmp_precedence(&pkg_version) == std::cmp::Ordering::Equal
+                })
+                .or_else(|| versions.last())
```
ldm0, line 435: Use `last()` for behaviour consistency with old cargo. Old cargo use HashMap and insert for path deps construction, therefore the last inserted package was picked.

src/cargo/core/resolver/encode.rs
ldm0: Oops, seems to make `patch::patch_git` fail. Reverted.

tests/testsuite/patch.rs
```diff
@@ -2914,3 +2914,120 @@ Caused by:
...
+[CHECKING] once_cell [..] ([CWD]/vendor/once_cell_[..])
+[CHECKING] [..] v0.1.0 ([CWD]/vendor/[..])
+[CHECKING] [..] v0.1.0 ([CWD]/vendor/[..])
+[CHECKING] cargo-bug-repo v0.1.0 ([CWD])
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [..]
+",
+        )
+        .run();
+
+    p.cargo("check").with_stderr("[FINISHED] [..]").run();
```
epage, line 3032: The fact that the following commit doesn't change the test, I'm assuming the test is failing within this commit.

Tests should pass in every commit.  Please update this to be passing and then update the test in the following commit to still be passing.  Showing how the test changes helps to test the test and "documents" for people looking at this the exact change in behavior.

epage:
Note: I just did a quick scan of the issue (and commit structure).  I've not looked at the test or code change in detail yet

tests/testsuite/patch.rs
ldm0: Fixed :-)

ldm0:
It seems that CI failure is caused by `check-cfg` https://github.com/rust-lang/cargo/actions/runs/8964218728/job/24615675343?pr=13863
Related: https://github.com/rust-lang/cargo/pull/13571

ldm0:
ping @weihanglo

weihanglo:
Sorry for the late reply. Thought epage was on it. Feel free to open a new one whenever ready.

PR #13874, opened 2024-05-07
Old syntax suggestion

### LINKED ISSUE #13868: `cargo::` syntax should be permitted despite MSRV with the right configuration
Currently, zerocopy's `build.rs` script contains the following:

```rust
    for version_cfg in version_cfgs {
        if rustc_version >= version_cfg.version {
            println!("cargo:rustc-cfg={}", version_cfg.cfg_name);
        }
    }
```

We have a lot of `cfg`s that are now triggering the recently-added `unexpected_cfgs` lint. In order to fix this, I added the following to our build script:

```rust
    if rustc_version >= (Version { major: 1, minor: 77, patch: 0 }) {
        for version_cfg in &version_cfgs {
            println!("cargo::rustc-check-cfg=cfg({})", version_cfg.cfg_name);
        }
    }
```

Note that this is gated to only run on Rust version 1.77 or later, since that's when the `cargo::` syntax was introduced.

Unfortunately, this causes the following error to be emitted:

```text
error: the `cargo::` syntax for build script output instructions was added in Rust 1.77.0, but the minimum supported Rust version of `zerocopy v0.8.0-alpha.9 (.../zerocopy)` is 1.56.0.
See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
```

This is spurious since our version detection logic will ensure that we don't emit these annotations prior to 1.77, but since it's a hard error, there's no way to work around it.

It would be great if there were a way to ask Cargo to disable this error for cases like these. The alternative for us is to just add a blanket `#![allow(unexpected_cfgs)]`, which would be a shame since that lint is genuinely useful.

(I will also note that this error seems especially odd given that pre-1.77 toolchains simply ignore `cargo::`; it doesn't cause a compilation error or even a warning)

torhovland:
Fixes #13868.

The build error in the issue will now include a suggestion:

```
   Compiling zerocopy v0.8.0-alpha.9
error: the `cargo::` syntax for build script output instructions was added in Rust 1.77.0, but the minimum supported Rust version of `zerocopy v0.8.0-alpha.9` is 1.56.0.
Consider using the old `cargo:` syntax in front of `rustc-check-cfg=`.
See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
```

The suggestion is only included for reserved prefixes.

A test has been added.

### DIFF
--- a/src/cargo/core/compiler/custom_build.rs
+++ b/src/cargo/core/compiler/custom_build.rs
@@ -722,24 +722,28 @@ impl BuildOutput {
         const DOCS_LINK_SUGGESTION: &str = "See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script \
                 for more information about build script outputs.";
 
+        fn has_reserved_prefix(flag: &str) -> bool {
+            RESERVED_PREFIXES
+                .iter()
+                .any(|reserved_prefix| flag.starts_with(reserved_prefix))
+        }
+
         fn check_minimum_supported_rust_version_for_new_syntax(
             pkg_descr: &str,
             msrv: &Option<RustVersion>,
-            key: &str,
+            flag: &str,
         ) -> CargoResult<()> {
             if let Some(msrv) = msrv {
                 let new_syntax_added_in = RustVersion::from_str("1.77.0")?;
                 if !new_syntax_added_in.is_compatible_with(msrv.as_partial()) {
-                    let prefix = format!("{key}=");
-
-                    let old_syntax_suggestion = RESERVED_PREFIXES
-                        .contains(&&*prefix)
-                        .then(|| {
-                            format!(
-                                "Consider using the old `cargo:` syntax in front of `{prefix}`.\n"
-                            )
-                        })
-                        .unwrap_or_default();
+                    let old_syntax_suggestion = if has_reserved_prefix(flag) {
+                        format!("Consider using the old `cargo:` syntax in front of `{flag}`.\n")
+                    } else if flag.starts_with("metadata=") {
+                        let old_format_flag = flag.strip_prefix("metadata=").unwrap();
+                        format!("Consider using the old `cargo:{old_format_flag}` syntax instead of `cargo::{flag}` (note the single colon).\n")
+                    } else {
+                        String::new()
+                    };
 
                     bail!(
                         "the `cargo::` syntax for build script output instructions was added in \
@@ -806,17 +810,13 @@ impl BuildOutput {
             };
             let mut old_syntax = false;
             let (key, value) = if let Some(data) = line.strip_prefix("cargo::") {
+                check_minimum_supported_rust_version_for_new_syntax(pkg_descr, msrv, data)?;
                 // For instance, `cargo::rustc-flags=foo` or `cargo::metadata=foo=bar`.
-                let (key, value) = parse_directive(whence.as_str(), line, data, old_syntax)?;
-                check_minimum_supported_rust_version_for_new_syntax(pkg_descr, msrv, key)?;
-                (key, value)
+                parse_directive(whence.as_str(), line, data, old_syntax)?
             } else if let Some(data) = line.strip_prefix("cargo:") {
                 old_syntax = true;
                 // For instance, `cargo:rustc-flags=foo`.
-                if RESERVED_PREFIXES
-                    .iter()
-                    .any(|prefix| data.starts_with(prefix))
-                {
+                if has_reserved_prefix(data) {
                     parse_directive(whence.as_str(), line, data, old_syntax)?
                 } else {
                     // For instance, `cargo:foo=bar`.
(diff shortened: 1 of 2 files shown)


tests/testsuite/build_script.rs
```diff
@@ -5508,6 +5508,47 @@ for more information about build script outputs.
         .run();
 }
 
+#[cargo_test]
+fn test_new_syntax_with_old_msrv_and_reserved_prefix() {
```
epage, line 5512: Generally, we recommend splitting up PRs into at least two commits
- Adding of tests (must pass)
- Changing of behavior with requisite test updates

This makes it easier to review how the behavior changed and helps test the test.  A little less important in this case but would still be nice if you have to update this PR

src/cargo/core/compiler/custom_build.rs
```diff
@@ -725,13 +725,26 @@ impl BuildOutput {
...
             if let Some(msrv) = msrv {
                 let new_syntax_added_in = RustVersion::from_str("1.77.0")?;
                 if !new_syntax_added_in.is_compatible_with(msrv.as_partial()) {
+                    let prefix = format!("{key}=");
+
+                    let old_syntax_suggestion = RESERVED_PREFIXES
+                        .contains(&&*prefix)
+                        .then(|| {
+                            format!(
+                                "Consider using the old `cargo:` syntax in front of `{prefix}`.\n"
```
epage, line 739: imo splitting up the suggestion like this feels a little more awkward to read.

Maybe

```suggestion
                                "Switch to `cargo:{prefix}` (note the single colon).\n"
```

src/cargo/core/compiler/custom_build.rs
```diff
@@ -725,13 +725,26 @@ impl BuildOutput {
         fn check_minimum_supported_rust_version_for_new_syntax(
             pkg_descr: &str,
             msrv: &Option<RustVersion>,
+            key: &str,
         ) -> CargoResult<()> {
             if let Some(msrv) = msrv {
                 let new_syntax_added_in = RustVersion::from_str("1.77.0")?;
                 if !new_syntax_added_in.is_compatible_with(msrv.as_partial()) {
+                    let prefix = format!("{key}=");
```
epage, line 733: Shouldn't this be inside of the `then` or inlined directly into that `format`?

src/cargo/core/compiler/custom_build.rs
```diff
@@ -725,13 +725,26 @@ impl BuildOutput {
...
+
+                    let old_syntax_suggestion = RESERVED_PREFIXES
```
epage, line 735: When it isn't in `RESERVED_PREFIXES`, if the key is `metadata`, we should suggest `cargo:{value}`

Unsure if we should say much more otherwise.

src/cargo/core/compiler/custom_build.rs
torhovland: I wanted to make it clear that the single colon is an old syntax. What do you think?

tests/testsuite/build_script.rs
```diff
@@ -5508,6 +5508,47 @@ for more information about build script outputs.
...
+            "#,
+        )
+        .build();
+
+    p.cargo("build")
+        .with_status(101)
+        .with_stderr_contains(
+            "\
+[COMPILING] foo [..]
+error: the `cargo::` syntax for build script output instructions was added in Rust 1.77.0, \
```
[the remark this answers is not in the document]
torhovland: That seems like an unrelated change, as it is handled by `cargo::core::shell::ShellOut`, and affects old tests.

tests/testsuite/build_script.rs
torhovland: Done.

src/cargo/core/compiler/custom_build.rs
torhovland: It was used one more place, but it's refactored now anyway.

src/cargo/core/compiler/custom_build.rs
torhovland: Done.

tests/testsuite/build_script.rs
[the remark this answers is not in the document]
arlosi: This is just a request to use the substitution for the `error: ` prefix used by Cargo's test framework. Similar to the `[COMPILING]` on the previous line.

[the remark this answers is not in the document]
torhovland: I see! Updated.

src/cargo/core/compiler/custom_build.rs
torhovland: OK, how about now?

weihanglo (approved):
Thanks for the work!

It would be great if you could clean up some intermediate commits, and make the git history easier to track in the future.

torhovland:
OK, I squashed a couple commits. Hope that's better.

PR #13878, opened 2024-05-07
docs(ref): Correct heading level of `[lints]` documentation

kpreid:
It was two levels too deep. (Compare with the following `[badges]` section header.)


weihanglo (approved):
Nice catch. Thank you!

(don't know how we missed that 😅)

PR #13879, opened 2024-05-07
test: clean up unnecessary uses of `match_exact`

weihanglo:
It was found during the review of <https://github.com/rust-lang/cargo/pull/13842#discussion_r1591789376>

We should be happy using `assert_match_exact` directly.
The extra arguments to `match_exact` are not necessary for those test cases.


Muscraft (approved):
Thanks!

PR #13880, opened 2024-05-08
Refactor cargo lint tests

Muscraft:
In #13621, it was brought up that [the lints tests are nested more deeply than other UI tests](https://github.com/rust-lang/cargo/pull/13621#discussion_r1538065181). This got me wondering if there was a better way to structure all the lint tests.
What I came up with was:
- Lints should not have UI tests, only parts of the diagnostic system, i.e., how warnings, errors, notes, etc., look
  - This is because UI tests should focus on parts of the system that make up each lint's output
  - We can always add UI tests for each lint if desired
- All tests related to the linting system should live in `tests/testsuite/lints/`
- Tests related to `[lints.cargo]` should stay in `lints_table.rs` as it is for the whole lints table
- Each lint will get a file in `lints/` for all of its tests
  - This makes `lints/mod.rs` smaller and targeted only at tests for the linting system itself
  - It makes it much easier to know where to place a test


tests/testsuite/cargo_lints.rs
```diff
@@ -0,0 +1,677 @@
...
+  | ----------------
+  |
+  = help: consider adding `cargo-features = [\"test-dummy-unstable\"]` to the top of the manifest
+error: encountered 2 errors(s) while verifying lints
+",
+        )
+        .run();
+}
+
+// LINT_SPECIFIC_TESTS
```
weihanglo, line 445: I am fine with have `cargo_lints.rs` for organizing every lint test, though I don't understand why we can't have nested UI tests. I believe the amount of lint specific tests will grow beyond controls like `testsuite/build.rs`.

weihanglo (commented):
I am fine with this change, though I wonder

tests/testsuite/cargo_lints.rs
Muscraft: Nesting tests more deeply would not be a problem; the problem, from my understanding, is that they would diverge from how other UI tests are structured.

weihanglo (approved):
Looks good to me, and looking forward to documenting the lint development process in the future :)

Something is failing rustfix tests on nightly and I am working on it:

```

failures:

---- everything stdout ----
failed: ./tests/everything/multiple-solutions.rs
thread 'everything' panicked at crates/rustfix/tests/parse_and_replace.rs:233:9:
1 out of 9 fixture asserts failed
error: test failed, to rerun pass `-p rustfix --test parse_and_replace`
(run with `env RUST_LOG=parse_and_replace=info` to get more details)
```

weihanglo:
@bors r+

Thanks for the refactor!

Issue #2904, opened 2016-07-21
Provide better diagnostics for why crates are rebuilt


alexcrichton:
Right now Cargo can sometimes choose to recompile crates for surprising reasons, but it has lots of information internally to inform users as to why crates are recompiling. It'd be great to expose all this information!

Currently the best way to get this information is:

```
CARGO_LOG=cargo::core::compiler::fingerprint=trace cargo build
```

but then you have to wade through a bunch of other logging output. It'd be best if cargo would automatically do this on `-v`. Something like "recompiling crate `foo` because: ...". Maybe even `-vv`.

jwilm:
> but then you have to wade through a bunch of other logging output.

Adding such debug information to `-v` or `-vv` would result in even more log output to parse through in comparison with that `RUST_LOG` setting. A separate flag might be nicer to get only the rebuild diagnostics.

davidbarsky:
I’m opposed to hiding that information behind a flag—

davidbarsky:
I'm opposed to hiding that information behind a flag—in particular, the issue I had in #3067 would’ve been noticed far sooner. An (relatively) rare event like full recompilation of a dependency should—in my amateurish opinion—be communicated to the user _as quickly as possible_, and I think the regular `cargo build` output is a good place to put it.

Caveat: I am ignorant of most of cargo is built. I just wandered in here, all oblivious.

alexcrichton:
@davidbarsky yeah that sounds reasonable to me. We try to keep Cargo's UI pretty tight to ensure that you're not inundated with output, but perhaps we could do something like print `Recompiling` instead of `Compiling` with a small snippet of a reason at one end. Then if you pass `-v` we could print more information about why something is being recompiled.

davidbarsky:
@alexcrichton Thanks Alex! Got a quick question though: wouldn’t passing `-v` only show the status for the currently executing build, not previous one? On other hand, a user might not care _why_ a previous build recompiled, only if the build keeps recompiling.

alexcrichton:
It's true yeah, `-v` may be too late in some situations. For debugging though it could be useful?

jwilm:
Whenever I've encountered a recompiling issue it's usually very easy to reproduce, and opting into extra info via `-v` for debugging would be fine. If it's a one time issue and `-v` would be too late, is it really a problem?

davidbarsky:
[...]

PR #13883, opened 2024-05-08
Log the name of the dependency that changed

### LINKED ISSUE #13859: Show more details on "Dirty ...: dependency info changed"
### Problem

Debugging unexpected rebuilds got a lot nicer with `cargo build -v` showing a reason for "Dirty" files. However, sometimes one gets
```
       Dirty miri v0.1.0 (D:\a\miri\miri): dependency info changed
```
which is not really actionable.

### Proposed Solution

Would be great to get some more details about what changed here. For this specific case, the `cargo::core::compiler::fingerprint` log shows
```
2024-05-04T06:41:27.4318113Z    0.102008300s  INFO prepare_target{force=false package_id=miri v0.1.0 (D:\a\miri\miri) target="miri"}: cargo::core::compiler::fingerprint: fingerprint dirty for miri v0.1.0 (D:\a\miri\miri)/Build/TargetInner { name: "miri", tested: false, doc: true, ..: with_path("D:\\a\\miri\\miri\\src\\bin\\miri.rs", Edition2021) }
2024-05-04T06:41:27.4320415Z    0.102041200s  INFO prepare_target{force=false package_id=miri v0.1.0 (D:\a\miri\miri) target="miri"}: cargo::core::compiler::fingerprint:     dirty: UnitDependencyInfoChanged { old_name: "ctrlc", old_fingerprint: 6202229392602597251, new_name: "ctrlc", new_fingerprint: 4374262352119191711 }
```
So apparently the ctrlc crate changed somehow. I don't know what this means, but maybe it would make sense to mention ctrlc in the `cargo build -v` output. Maybe it could even explain what changed about ctrlc -- I have no idea what flows into this fingerprint that apparently changed.

### Notes

_No response_

torhovland:
Fixes #13859 by adding the name of the dependency to the "dependency info changed" log message.

There is a surprising test output after this. See code review.

### DIFF
--- a/tests/testsuite/lto.rs
+++ b/tests/testsuite/lto.rs
@@ -604,7 +604,7 @@ fn dylib() {
 [COMPILING] registry-shared v0.0.1
 [FRESH] registry v0.0.1
 [RUNNING] `rustc --crate-name registry_shared [..]-C embed-bitcode=no[..]
-[DIRTY] bar v0.0.0 ([..]): dependency info changed
+[DIRTY] bar v0.0.0 ([..]): dependency info changed (registry_shared)
 [COMPILING] bar [..]
 [RUNNING] `rustc --crate-name bar [..]--crate-type dylib [..]-C embed-bitcode=no[..]
 [FINISHED] [..]
@@ -622,7 +622,7 @@ fn dylib() {
 [FRESH] registry-shared v0.0.1
 [COMPILING] registry v0.0.1
 [RUNNING] `rustc --crate-name registry [..]
-[DIRTY] bar v0.0.0 ([..]): dependency info changed
+[DIRTY] bar v0.0.0 ([..]): dependency info changed (registry_shared)
 [COMPILING] bar [..]
 [RUNNING] `rustc --crate-name bar [..]--crate-type dylib [..]-C embed-bitcode=no[..]
 [RUNNING] `rustc --crate-name bar [..]-C lto [..]--test[..]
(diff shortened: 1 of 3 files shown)


tests/testsuite/lto.rs
```diff
@@ -622,7 +622,7 @@ fn dylib() {
 [FRESH] registry-shared v0.0.1
 [COMPILING] registry v0.0.1
 [RUNNING] `rustc --crate-name registry [..]
-[DIRTY] bar v0.0.0 ([..]): dependency info changed
+[DIRTY] bar v0.0.0 ([..]): dependency info changed (registry_shared)
```
torhovland, line 625: This message looks strange to me. Should it not output `registry` rather than `registry_shared` here?

epage: Part of the problem is we only report the first dependency, rather than all.

I suspect what is happening here is we are reusing `registry_shared` with a version from a couple builds back.  This is a different hash than the immediately prior build, so it requires rebuilding `bar` but this isn't quite clear without more context.

I wonder if this change will cause more confusion than good and we should defer to #2904 which would add a `cargo report` subcommand for reporting why the last command rebuilt and we can include a lot more context.

weihanglo: See <https://github.com/rust-lang/cargo/issues/13859#issuecomment-2103227615>, yet it might need some design discussions before proceed.

weihanglo:
Going to close this as it needs more designs before making a PR. Thank you for being interested in this.

torhovland:
Fair enough, but the issue should probably not be `S-accepted`, then.

PR #13885, opened 2024-05-08
Case insensitive file name matching when packaging files

### LINKED ISSUE #13722: Cargo packages duplicate files on case-insensitive file systems
### Problem

It seems that Cargo is excluding an already-packaged files using exact name comparison, which doesn't always match how the file system sees name equality. 

```text
   Archiving Cargo.lock
   Archiving Cargo.toml
   Archiving Cargo.toml.orig
   Archiving README.md
   Archiving readme.Md
   Archiving src/main.rs
```

Example crate:

https://docs.rs/crate/rosu/0.6.0/source/

### Steps

```toml
[package]
…
readme = "README.md"
```

```bash
echo case > readme.Md
cargo package
```

The same applies to `license-file`, `cargo.lock`.

### Possible Solution(s)

Theoretically there could be other gotchas of this kind, e.g. HFS+ file system on macOS forces file names to use NFD Unicode form, while most text has NFC form, which makes codepoint-by-codepoint comparisons not equal. However HFS+ is on its way out, so perhaps a simple case-insensitive comparison will suffice.

### Notes

_No response_

### Version

```text
cargo 1.79.0-nightly (499a61ce7 2024-03-26)
```

torhovland:
Fixes #13722.

Handles the readme and license files. The `Cargo.lock` file is trickier. There are a bunch of places in the code where the existence of `Cargo.lock` (case sensitive) is checked, including when packaging.

I think it should be OK to keep `Cargo.lock` out of this. While the readme and license files are typically created by the user, the `Cargo.lock` file is typically created by Cargo, hence casing should be consistent anyway.

### DIFF
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -366,23 +366,31 @@ fn check_for_file_and_add(
     result: &mut Vec<ArchiveFile>,
     ws: &Workspace<'_>,
 ) -> CargoResult<()> {
+    fn compare_file_names_case_insensitive(result: &Vec<ArchiveFile>, rel_str: &str) -> bool {
+        result
+            .iter()
+            .any(|ar| ar.rel_path.to_string_lossy().to_lowercase() == rel_str.to_lowercase())
+    }
+
     match abs_file_path.strip_prefix(&pkg.root()) {
         Ok(rel_file_path) => {
-            if !result.iter().any(|ar| ar.rel_path == rel_file_path) {
+            let rel_str = rel_file_path
+                .to_str()
+                .expect("everything was utf8")
+                .to_string();
+
+            if !compare_file_names_case_insensitive(result, &rel_str) {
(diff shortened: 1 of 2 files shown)


epage:
Note that #13722 is not marked as [S-Accepted](https://doc.crates.io/contrib/process/index.html#working-on-issues).  I don't think we've come to a conclusion on if it should be fixed and what the right constraints are.

torhovland:
I am now noticing that Cargo uses the `unicase` crate for case insensitive comparisons elsewhere.

```rust
use unicase::Ascii as UncasedAscii;
```

This PR should do that rather than explicitly converting to lowercase and comparing.

weihanglo:
Thanks for the PR. Marked it as draft for now to reflect the fact that it's still in discussion.

epage:
As this is stale and #13722  is the place for discussion a proposal, I'm going to go ahead and close this.

PR #13886, opened 2024-05-08
Silence the warning about forgetting the vendoring

### LINKED ISSUE #12802: re-vendored suggestion is incorrect with sparse crates.io
### Problem

The error message for a resolver activation error includes this comment when used with crates.io:

```
perhaps a crate was updated and forgotten to be re-vendored?
```

This is not correct. This message should only be displayed when vendoring is enabled. Cargo is getting confused somehow, thinking index.crates.io is a replaced source. This message should only be displayed when vendoring is actually enabled.

### Steps

Try to build with:

```toml
[dependencies]
regex = "99"
```

displays:

```
    Updating crates.io index
error: failed to select a version for the requirement `regex = "^99"`
candidate versions found which didn't match: 1.10.0, 1.9.6, 1.9.5, ...
location searched: crates.io index
required by package `foo v0.1.0 (/Users/eric/Temp/foo)`
perhaps a crate was updated and forgotten to be re-vendored?
```

### Possible Solution(s)

The error message is generated [here](https://github.com/rust-lang/cargo/blob/e4fe8f04e76dd6172dce67a840bfc6a1ebacd76a/src/cargo/core/resolver/errors.rs#L297-L299).

I'm uncertain at this moment why the source seems to be replaced. I vaguely recall having cargo treat index.crates.io as equivalent for the git index, but I forget how that all works.

### Notes

_No response_

### Version

```text
cargo 1.75.0-nightly (794d0a825 2023-10-03)
release: 1.75.0-nightly
commit-hash: 794d0a82547f3081044c0aca7b6083733ce51344
commit-date: 2023-10-03
host: aarch64-apple-darwin
libcurl: 8.1.2 (sys:0.4.66+curl-8.3.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 14.0.0 [64-bit]
```

jneem:
When sparse crates.io is used, dependencies from crates.io always come from `ReplacedSource`s, as the non-sparse crates.io gets replaced by the sparse one. As far as diagnostics go, this case shouldn't "count" as a replaced source.

I had trouble adding a test for this change, because the test harness replaces everything with a dummy registry.

Fixes #12802

### DIFF
--- a/src/cargo/sources/replaced.rs
+++ b/src/cargo/sources/replaced.rs
@@ -152,7 +152,7 @@ impl<'gctx> Source for ReplacedSource<'gctx> {
     }
 
     fn is_replaced(&self) -> bool {
-        true
+        !self.is_builtin_replacement()
     }
 
     fn add_to_yanked_whitelist(&mut self, pkgs: &[PackageId]) {
--- a/src/cargo/sources/source.rs
+++ b/src/cargo/sources/source.rs
@@ -146,6 +146,8 @@ pub trait Source {
     fn describe(&self) -> String;
 
     /// Returns whether a source is being replaced by another here.
+    ///
+    /// Builtin replacement of `crates.io` doesn't count as replacement here.
     fn is_replaced(&self) -> bool {
         false
     }
--- a/tests/testsuite/registry.rs
+++ b/tests/testsuite/registry.rs
@@ -3833,3 +3833,43 @@ fn builtin_source_replacement() {
         )
         .run();
 }
+
+#[cargo_test]
+fn builtin_source_replacement_no_vendor_error() {
+    // errors for builtin source replacement of crates.io
+    // should not mention outdated vendor dependencies
+    let server = RegistryBuilder::new().build();
+
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.0.1"
+                edition = "2021"
+
+                [dependencies]
+                dep = "0.2.0"
+            "#,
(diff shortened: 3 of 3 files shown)


ehuss:
Thanks!

I believe you can use the `replace_crates_io()` function to replace crates.io without using the typical source replacement. See the test `builtin_source_replacement` as an example. Can you give that a try and see if it works?

Also, can you update the docstring on `Registry::is_replaced` to mention that it is false for built-in replacements?

Perhaps @arlosi could give this a second set of eyes to determine if this is the best strategy to fix the issue. I *think* it should be fine. This method is only used for this one error message. I just have a slight concern that it could cause confusion in the future (returning false for something that is replaced internally).

jneem:
Thanks for the hint! I managed to get a test working (which fails before this PR, and passes after).

> I just have a slight concern that it could cause confusion in the future (returning false for something that is replaced internally).

I was a little worried about this too. I think the usages of `ReplacedSource::is_builtin_replacement` suggests that it's often necessary to distinguish between builtin and non-builtin replacements. So if it ever becomes necessary to do so on the `Source` trait, we could replace `is_replaced` by a method returning a three-value enum (`NotReplaced`, `BuiltinReplaced`, `OtherReplaced`)

arlosi:
I think this is OK:
* The fact that Cargo is using source replacement to handle sparse registry access to crates.io is an implementation detail.
* Consumers of the `Source` trait should not see it as a source replacement since the *user* did not configure source replacement here.

Thanks for this fix!

@bors r+

PR #13887, opened 2024-05-08
docs: clarify dash replacement rule in target name

weihanglo:

A follow-up of <https://github.com/rust-lang/cargo/issues/13867> and <https://github.com/rust-lang/cargo/pull/13882>.

This clarifies that dashes will be replaced in every JSON messages documentation we have.
Also fixes a documentation bug that Cargo in fact doesn't replace dashes for the default binary (`src/main.rs`).

Read the doc and suggest changes.

### DIFF
--- a/src/doc/man/cargo-metadata.md
+++ b/src/doc/man/cargo-metadata.md
@@ -151,7 +151,9 @@ The JSON output has the following format:
                     "crate_types": [
                         "bin"
                     ],
-                    /* The name of the target. */
+                    /* The name of the target.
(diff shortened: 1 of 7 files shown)


Muscraft (approved):
Thanks!

Turbo87:
> Also fixes a documentation bug that Cargo in fact doesn't replace dashes for the default binary (`src/main.rs`).

lol, I was just about to open a PR for fixing that in the docs and then I discovered it's already fixed on `master` 😆

PR #13890, opened 2024-05-09
test(rustfix): run some tests only on nightly

weihanglo:

This is a temporary fix for <https://github.com/rust-lang/cargo/pull/13880#pullrequestreview-2047809180>, which failed due to rustc improvement <https://github.com/rust-lang/rust/pull/123344>.

I am copying `cargo-test-macro` without thinking too much for checking Rust channel.
Plan to reuse the infra and migrate to snapbox for rustfix tests in the future.


weihanglo:
Note that I didn't even look into whether the `unused_import` lint still gave us multiple solutions.

ehuss:
Thanks!

These tests could probably use a different strategy in the future. For example, perhaps they could not run `rustc` if the JSON snapshot is already there. I think they only really buy us a cross-check that the compiler has not broken the JSON format, but I'm not sure that is worth it.

@bors r+

Nilstrieb:
@pietroalbini wow I can't believe you'd dare breaking cargo :(

PR #13898, opened 2024-05-10
Preserve file permissions on unix during `write_atomic`

### LINKED ISSUE #13896: `cargo add` changes file permissions of `Cargo.toml` to 600
### Problem

When you run `cargo add`, it changes the file permissions of `Cargo.toml` to 600 (user read+write only). This is a little bit painful when you're building the code as a different user than the user writing the code, for example if you're running the code in a container. This applies to `cargo remove` as well. I tested this behaviour on Cargo 1.78.0 and nightly.

### Steps

```bash
$ mkdir cargo-add-test && cd cargo-add-test
$ cargo init
$ ls -l
-rw-rw-r-- 1 user user   85 May  9 23:09 Cargo.toml
drwxrwxr-x 2 user user 4096 May  9 23:09 src
$ cargo add libc
$ ls -l
-rw-rw-r-- 1 user user  376 May  9 23:09 Cargo.lock
-rw------- 1 user user  102 May  9 23:09 Cargo.toml
drwxrwxr-x 2 user user 4096 May  9 23:09 src
$ umask
0002
```

### Possible Solution(s)

I'm guessing that Cargo replaces the existing file with a new file, and this is why the permissions change. In this case, I think it would be better to create the file with the original user/group/world permissions (ignoring things like suid permissions), ideally bypassing the umask as well.

**Edit:** It appears to use the tempfile crate, which [uses 600 by default](https://docs.rs/tempfile/3.10.1/tempfile/struct.Builder.html#method.permissions) on Linux.

https://github.com/rust-lang/cargo/blob/4de0094ac78743d2c8ff682489e35c8a7cafe8e4/crates/cargo-util/src/paths.rs#L186-L194

### Notes

_No response_

### Version

```text
cargo 1.80.0-nightly (0ca60e940 2024-05-08)
release: 1.80.0-nightly
commit-hash: 0ca60e940821c311c9b25a6423b59ccdbcea218f
commit-date: 2024-05-08
os: Ubuntu 20.4.0 (focal) [64-bit]
```

stevenengler:

Fixes #13896.

> When you run `cargo add`, it changes the file permissions of `Cargo.toml` to 600 (user read+write only). This is a little bit painful when you're building the code as a different user than the user writing the code, for example if you're running the code in a container. This applies to `cargo remove` as well. I tested this behaviour on Cargo 1.78.0 and nightly.

I'm not entirely sure how permissions are handled on Windows, but the tempfile lib [doesn't seem to support them](https://docs.rs/tempfile/3.10.1/tempfile/struct.Builder.html#windows-and-others), so I haven't changed the behaviour on Windows.

Only the user/group/other read/write/execute permission bits are copied.

This PR sets the permissions ~twice~ once:
~1. When creating the file. This has the umask applied, but means that we don't create a file that is more permissive than the original.~
2. After the file has been created. This doesn't apply the umask, resulting in the file having the same u/g/o r/w/x permissions as the original file.

Since this PR changes a util function, it has a wider scope than just changing the behaviour of `cargo add` and `cargo remove`. `write_atomic` is called from the following functions:

- [`migrate_manifests`](https://github.com/rust-lang/cargo/blob/4de0094ac78743d2c8ff682489e35c8a7cafe8e4/src/cargo/ops/fix.rs#L340)
- [`update_manifest_with_new_member`](https://github.com/rust-lang/cargo/blob/4de0094ac78743d2c8ff682489e35c8a7cafe8e4/src/cargo/ops/cargo_new.rs#L1008)
- [`LocalManifest::write`](https://github.com/rust-lang/cargo/blob/4de0094ac78743d2c8ff682489e35c8a7cafe8e4/src/cargo/util/toml_mut/manifest.rs#L299)
- [`gc_workspace`](https://github.com/rust-lang/cargo/blob/4de0094ac78743d2c8ff682489e35c8a7cafe8e4/src/bin/cargo/commands/remove.rs#L274)

Unit test was added (`cargo test -p cargo-util write_atomic_permissions`).

### DIFF
--- a/crates/cargo-util/src/paths.rs
+++ b/crates/cargo-util/src/paths.rs
@@ -824,6 +824,32 @@ mod tests {
     }
 
     #[test]
+    #[cfg(unix)]
+    fn write_atomic_permissions() {
+        use std::os::unix::fs::PermissionsExt;
+
+        let original_perms = std::fs::Permissions::from_mode(u32::from(
+            libc::S_IRWXU | libc::S_IRGRP | libc::S_IWGRP | libc::S_IROTH,
+        ));
+
+        let tmp = tempfile::Builder::new().tempfile().unwrap();
+
+        // need to set the permissions after creating the file to avoid umask
+        tmp.as_file()
+            .set_permissions(original_perms.clone())
+            .unwrap();
+
+        // after this call, the file at `tmp.path()` will not be the same as the file held by `tmp`
+        write_atomic(tmp.path(), "new").unwrap();
+        assert_eq!(std::fs::read_to_string(tmp.path()).unwrap(), "new");
+
+        let new_perms = std::fs::metadata(tmp.path()).unwrap().permissions();
+
+        let mask = u32::from(libc::S_IRWXU | libc::S_IRWXG | libc::S_IRWXO);
+        assert_eq!(0o600, new_perms.mode() & mask);
+    }
+
+    #[test]
     fn join_paths_lists_paths_on_error() {
         let valid_paths = vec!["/testing/one", "/testing/two"];
         // does not fail on valid input
--- a/crates/cargo-util/src/paths.rs
+++ b/crates/cargo-util/src/paths.rs
@@ -4,7 +4,7 @@ use anyhow::{Context, Result};
 use filetime::FileTime;
 use std::env;
 use std::ffi::{OsStr, OsString};
-use std::fs::{self, File, Metadata, OpenOptions};
+use std::fs::{self, File, Metadata, OpenOptions, Permissions};
 use std::io;
 use std::io::prelude::*;
 use std::iter;
@@ -185,10 +185,34 @@ pub fn write<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()>
 /// write_atomic uses tempfile::persist to accomplish atomic writes.
 pub fn write_atomic<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()> {
     let path = path.as_ref();
+
+    // On unix platforms, get the permissions of the original file. Copy only the user/group/other
+    // read/write/execute permission bits. The tempfile lib defaults to an initial mode of 0o600,
+    // and we'll set the proper permissions after creating the file.
+    #[cfg(unix)]
+    let perms = path.metadata().ok().map(|meta| {
+        use std::os::unix::fs::PermissionsExt;
+
+        // these constants are u16 on macOS
+        let mask = u32::from(libc::S_IRWXU | libc::S_IRWXG | libc::S_IRWXO);
+        let mode = meta.permissions().mode() & mask;
+
+        Permissions::from_mode(mode)
+    });
+
     let mut tmp = TempFileBuilder::new()
         .prefix(path.file_name().unwrap())
         .tempfile_in(path.parent().unwrap())?;
     tmp.write_all(contents.as_ref())?;
+
+    // On unix platforms, set the permissions on the newly created file. We can use fchmod (called
+    // by the std lib; subject to change) which ignores the umask so that the new file has the same
+    // permissions as the old file.
+    #[cfg(unix)]
+    if let Some(perms) = perms {
+        tmp.as_file().set_permissions(perms)?;
+    }
+
     tmp.persist(path)?;
     Ok(())
 }
@@ -846,7 +870,7 @@ mod tests {
         let new_perms = std::fs::metadata(tmp.path()).unwrap().permissions();
 
         let mask = u32::from(libc::S_IRWXU | libc::S_IRWXG | libc::S_IRWXO);
-        assert_eq!(0o600, new_perms.mode() & mask);
+        assert_eq!(original_perms.mode(), new_perms.mode() & mask);
     }
 
     #[test]


crates/cargo-util/src/paths.rs
```diff
@@ -185,10 +185,45 @@ pub fn write<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()>
...
-        .prefix(path.file_name().unwrap())
-        .tempfile_in(path.parent().unwrap())?;
+    let mut builder = TempFileBuilder::new();
+
+    builder.prefix(path.file_name().unwrap());
+
+    // On unix platforms, use the same permissions as the original file. Copy only the
+    // user/group/other read/write/execute permission bits. While we need to also set the
+    // permissions again later to bypass the umask, we still need to set the permissions here as
+    // well so that we don't create a more-permissive file than the original. The tempfile lib says
```
weihanglo, line 195: The default seems to be [0o600](https://docs.rs/tempfile/3/tempfile/struct.Builder.html#method.permissions), which is pretty safe I believe.  Do we really need to set permission for tempfiles?
Granted, if the `Cargo.toml` was readonly, maybe `cargo add` should fail.

crates/cargo-util/src/paths.rs
```diff
@@ -823,6 +857,30 @@ mod tests {
         assert_eq!(contents, original_contents);
     }
 
+    #[test]
+    #[cfg(unix)]
+    fn write_atomic_permissions() {
```
weihanglo, line 863: Could we have this test in the first commit showing the problematic behavior, and the next commit fixes both the test and the behavior? By doing so it's a bit clearer to reviewers to just read the diff and understand what has been changed.

crates/cargo-util/src/paths.rs
stevenengler: I can remove that part and just use the default 600 permissions to simplify it. tempfile could change their default in the future, but that's probably unlikely.

crates/cargo-util/src/paths.rs
```diff
@@ -185,10 +185,45 @@ pub fn write<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()>
...
+    };
+
+    let mut tmp = builder.tempfile_in(path.parent().unwrap())?;
     tmp.write_all(contents.as_ref())?;
+
+    // On unix platforms, we set the permissions on the newly created file. When we set the
+    // permissions above, they were subject to the umask. Now that the file is created, we can use
+    // fchmod (called by the std lib; subject to change) to set the permissions which ignores the
+    // umask so that the new file has the same permissions as the old file.
+    #[cfg(unix)]
```
weihanglo, line 222: Maybe on Windows we could try [setting read-only](https://doc.rust-lang.org/stable/std/fs/struct.Permissions.html#method.set_readonly) if it was read-only?

I don't know whether it would fail or not when replacing a read-only file though.

crates/cargo-util/src/paths.rs
weihanglo: Yeah that could simplify the logic a bit.

crates/cargo-util/src/paths.rs
ChrisDenton: The standard library does not actually change any permissions on Windows. It sets the read-only attribute but from a security pov this is mostly useless as anyone with write permissions could just unset it again.

ChrisDenton: I mean, I guess it might make sense to preserve some attributes in any case. Seems not as important as perms though.

weihanglo: > I mean, I guess it might make sense to preserve some attributes in any case. Seems not as important as perms though.

True. I was thinking from that angle. Just a nice-to-have, not a blocker.

BTW, how could you always notice there is a Windows related issue happening, even when nobody pinged you?

ChrisDenton: Ha, I do miss things. But often people mention something to me privately or on discord. In this case I just happened to be browsing new PRs and this one looked interesting so I opened it then noticed a Windows thing.

crates/cargo-util/src/paths.rs
stevenengler: Reorganized it so that the first commit adds the tests, and the second commit adds the `write_atomic` changes.

crates/cargo-util/src/paths.rs
stevenengler: It now only sets the permissions once after the file has been created. The file will be created with tempfile's default of 600.

crates/cargo-util/src/paths.rs
stevenengler: I've added code to maintain the readonly property on non-unix platforms, but this doesn't work on Windows since unlike Linux you cannot delete or replace a read-only file on Windows. So when tempfile::persist is called it tries to replace the old read-only file, which fails with an "access denied" error.

I think there are workarounds, but we couldn't do it atomically without modifying the tempfile code or changing the readonly property on the original file. So I think it would be better to ignore Windows (and non-unix platforms). What do you think?

weihanglo: That makes sense. Go ahead :)


---

BTW, when I mentioned the commit organization, I meant something like this: 
* The first commit asserts the bad behavior. CI is all green with this commit.
* The second commit fixes both bug and **test**. CI is still green, and the diff bewteen shows the test change so we're more confident it fixes the previously "bad" behavior.
* See https://github.com/rust-lang/cargo/pull/13808 as a reference.

Not really a hard requirement though.

ChrisDenton: You should be able to delete read only files. Std has an issue open about renaming (https://github.com/rust-lang/rust/issues/123985) so that may be changed in the future. Not that it helps in the here and now.

stevenengler: Ah okay. I had borrowed a Windows computer and tried to delete a read-only file (and tried replacing a read-only file) on the command line with `del` and `move` and wasn't able to. But I have no idea what Windows APIs exist or what was being used by those commands. tempfile seems to use [`MoveFileExW`](https://github.com/Stebalien/tempfile/blob/e5418bd64758d1d0444e9158005f00ca7d2bc6ee/src/file/imp/windows.rs#L106).

If you have any suggestions about how to make this work on Windows let me know. My Windows API knowledge is very old and I don't have a good way to test things on Windows other than using the CI.

stevenengler: I removed the non-unix code that tries to preserve the readonly property.

> BTW, when I mentioned the commit organization, I meant something like this:

Ah sorry I misunderstood. I believe I fixed this now.

crates/cargo-util/src/paths.rs
stevenengler: As mentioned in one of the other threads, I fixed this so that the first commit adds the test for the existing behaviour, and the second commit fixes the behaviour and updates the test for the new behaviour.

weihanglo (approved):
Thanks! Looks pretty good now. We shouldn't block this from merge.

PR #10395, opened 2022-02-17
Enable propagating host rustflags to build scripts

jonhoo:
### What does this PR try to resolve?

This PR primarily fixes #10206, but in doing so it also slightly modifies the interface for the unstable `target-applies-to-host` feature (#9453), and adds the unstable `target-applies-to-host-kind` flag to mirror `target-applies-to-host` for build scripts and other host artifacts.

The commit messages have more in-depth discussion.

### How should we test and review this PR?

The test case from #10206 now works rather than producing an error. It has also been added a regression test case. A few additional test cases have also been added to handle the expected behavior around rustflags for build scripts with and without `target-applies-to-host-kind` enabled. 

### Additional information

1. This changes the interface for `target-applies-to-host` so that it does not need to be specified twice to be used. And it can still be set through configuration files using the `[unstable]` table. However, we may(?) want to pick a stable format for in-file configuration of this setting unless we intend for it to only ever be a command-line flag.
2. It may be that `target-applies-to-host-kind` is never behavior we want users to turn on, and that it should therefore simply be removed and hard-coded as being `false`.
3. It's not entirely clear how `target-applies-to-host-kind` should interact with `-Zmultitarget`. If, for example, `requested_kinds = [HostTarget, SomeOtherTarget]` and `kind.is_host()`, should `RUSTFLAGS` take effect or not? For the time being I've just hard-coded the behavior for single targets, and the answer would be "no".

PR #9753, opened 2021-08-01
Stabilize target-applies-to-host feature.

jameshilliard:
Details: #9453, #9322

PR #9634, opened 2021-06-28
Use cross compile style target/host isolation for all builds.

jameshilliard:
Note: I disabled a few tests that I wasn't sure how to fix but I think I got things working for the most part. This is still rather rough and needs a good bit of cleanup.

Issue #9453, opened 2021-05-03
Tracking Issue for target-applies-to-host


jameshilliard:
**Summary**

Original issue: #3349
Implementation: #9322
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#target-applies-to-host

The `target-applies-to-host` key in a config file can be used set the desired behavior for passing target config flags to build scripts.

rich-g:
Any chance of stabilizing this?

rich-g:
Just bumping this again. Would really like this stabilized so I can get off of nightly builds. :)

jameshilliard:
> Just bumping this again. Would really like this stabilized so I can get off of nightly builds. :)

Not sure what status of stabilizing this is, but if you want to use it on stable just set [this](https://github.com/buildroot/buildroot/blob/2022.08.1/package/pkg-cargo.mk#L43-L45) in your env.

PR #13900, opened 2024-05-11
Pass rustflags to artifacts built with implicit targets when using target-applies-to-host

### LINKED ISSUE #10744: RUSTFLAGS / RUSTDOCFLAGS are not being passed to rustdoc when target-applies-to-host nightly feature is used
### Problem

For two project I maintain, the docs generated by CI no longer includes an index file despite the appropriate nightly flags being used. I initially opened https://github.com/rust-lang/rust/issues/97881 for this, but now found out that it's actually a bug in Cargo that doesn't pass `RUSTDOCFLAGS` on to `rustdoc` in the affected workspaces (though it doesn't seem to affect all workspaces).

### Steps

* Clone https://github.com/ruma/ruma/tree/rustdoc-bug
* With the environment variable `RUSTDOCFLAGS` set to `--cfg compile_error`, run `cargo +nightly doc -p ruma`
* Observe no errors despite this in `crates/ruma/src/lib.rs`
  ```rust
  #[cfg(compile_error)]
  compile_error!("!!!");
  ```

### Notes

* Reproducible with a recent version of cargo (`1.63.0-nightly (85e457e15 2022-06-07)`)
* Reproducible with cargo `1.61.0-nightly (3d6970d 2022-02-28)` (rustup `nightly-2022-03-02`)
* Not reproducible with cargo `1.61.0-nightly (d6cdde5 2022-02-22)` (rustup `nightly-2022-03-01`).

### Version

```text
cargo 1.63.0-nightly (85e457e15 2022-06-07)
release: 1.63.0-nightly
commit-hash: 85e457e158db216a2938d51bc3b617a5a7fe6015
commit-date: 2022-06-07
libcurl: 7.83.1-DEV (sys:0.4.55+curl-7.83.1 vendored ssl:OpenSSL/1.1.1n)
```

gmorenz:
## What this PR does

This fixes #10744, a long-standing bug with `target-applies-to-host=false` where `RUSTFLAGS` are not being passed to artifact-units when built with `cargo build` (as opposed to `cargo build --target <host>`).

It doesn't fix a similar problem with `linker` and `links`. If the architecture in this PR is accepted, I expect I will be able to fix `linker` and `links` in the same way in a subsequent PR.

Below is a hopefully useful table of what flags are passed when, with new behavior bolded (without these changes the flag is not passed). I've only changed values in the top right cell, I've included the whole table for completeness and to hopefully make clear what exactly this feature is doing (which took me awhile to understand).

The table was generated with a host of x86_64-unknown-linux-gnu. "Flag" refers to values in RUSTFLAGS. "Linker" refers to the value of `--config target.<host>.linker` . The table was built with [this repo](https://github.com/gmorenz/target_application_to_host_matrix) and then hand modify to bold changed values.

| | target_applies_to_host=true | target_applies_to_host=false |
|-|-|-|
| no --target flag | Flag passed to bin<br/>Flag passed to shared dep from bin<br/>Linker passed to bin<br/><br/>Flag passed to proc macro<br/>Flag passed to shared dep from proc macro<br/>Linker passed to proc macro<br/><br/>Built with 5 invocations<br/>Without rustflags, built with 5 invocations<br/> | **Flag passed to bin**<br/>**Flag passed to shared dep from bin**<br/>Linker not passed to bin<br/><br/>Flag not passed to proc macro<br/>Flag not passed to shared dep from proc macro<br/>Linker not passed to proc macro<br/><br/>**Built with 6 invocations**<br/>Without rustflags, built with 5 invocations<br/> |
| --target x86_64-unknown-linux-gnu | Flag passed to bin<br/>Flag passed to shared dep from bin<br/>Linker passed to bin<br/><br/>Flag not passed to proc macro<br/>Flag not passed to shared dep from proc macro<br/>Linker passed to proc macro<br/><br/>Built with 6 invocations<br/>Without rustflags, built with 6 invocations<br/> | Flag passed to bin<br/>Flag passed to shared dep from bin<br/>Linker passed to bin<br/><br/>Flag not passed to proc macro<br/>Flag not passed to shared dep from proc macro<br/>Linker not passed to proc macro<br/><br/>Built with 6 invocations<br/>Without rustflags, built with 6 invocations<br/> |
| --target i686-unknown-linux-gnu | Flag passed to bin<br/>Flag passed to shared dep from bin<br/>Linker not passed to bin<br/><br/>Flag not passed to proc macro<br/>Flag not passed to shared dep from proc macro<br/>Linker passed to proc macro<br/><br/>Built with 6 invocations<br/>Without rustflags, built with 6 invocations<br/> | Flag passed to bin<br/>Flag passed to shared dep from bin<br/>Linker not passed to bin<br/><br/>Flag not passed to proc macro<br/>Flag not passed to shared dep from proc macro<br/>Linker not passed to proc macro<br/><br/>Built with 6 invocations<br/>Without rustflags, built with 6 invocations<br/> |

 ## How it is implemented

There are two stages in the `UnitGraph`'s life. It gets built, with correct `CompileKind`: `CompileKind::Host` for host units, `CompileKind::Target(HostTriple)` for artifact units. Then it gets rebuilt with artifact units that are intended for the host  having their kind changed to `CompileKind::Host`.

This rebuilding step serves to de-duplicate host and artifact units. A step that we want to maintain where possible. Because de-duplicating host and artifact dependencies is only possible if their `rustflags` don't differ we must calculate `rustflags` for units before this step, and this step necessarily throws away the information necessary to calculate them.

The possible rustflags have already been determined before creating units. "Calculating rustflags for units" just means determining which set of rustflags go with a particular unit, and storing that somewhere. The obvious place to do that is in the unit itself, so that's what this PR does, which has the convenient side affect of also handing the de-duplication logic for us. If the rustflags are the same, two units can be merged, if they differ, they cannot.

In the above table that's why it takes 5 invocations to build the repo without rustflags, but 6 with them. The shared_dependency merges when it is built without rustflags, and not when it is built with rustflags.

 ## Related PRs, comments and issues

fixes #10744

#9453 - Tracking issue

#9753 - Stabilization PR

#9634 - I believe this was another attempt (going down another implementation route) to fix the same issue

Comments from Alex Crichton noting that this must be fixed [1](https://github.com/rust-lang/cargo/pull/10395#discussion_r814239607) [2](https://github.com/rust-lang/cargo/pull/9753#issuecomment-891135066)

[My own comment describing exactly how I ran into this](https://github.com/rust-lang/cargo/pull/9753#issuecomment-2101475693)

[Documentation for the feature](https://doc.rust-lang.org/cargo/reference/unstable.html#target-applies-to-host)


gmorenz:
@rustbot label: Z-target-applies-to-host

The original work for this feature was done under Alex Crichton, and no team member has seemingly picked up the feature after he stepped down. Since the original bug isn't marked as "S-accepted" I'm rather ignoring the contributor documentation by submitting this PR in the first place (sorry).

If this isn't the direction whatever team member picks this up would like to go in, I'm happy to do something else. If no team member picks this up (which is rather what the contributing docs suggest should happen)... I guess the PR sits here harmlessly ¯\\\_(ツ)\_/¯ (I'll probably show up at office hours in that case and see if I can prompt some movement).

gmorenz:
If I'm reading the test error correctly, it's a formatting change in the benchmark harness *completely* unrelated to this PR

It's expecting

```
test bin_bench ... bench:           0 ns/iter (+/- 0)
```

It's finding

```
test bin_bench ... bench:           0.00 ns/iter (+/- 0.00)
```

With extra `.00`s

ehuss:
If you rebase on latest master, the test issue should be fixed.

gmorenz:
Rebased on master. Also merged the commit fixing the doc links into the commits moving the the docs around, since I noticed comments on other PRs asking that every commit pass tests.

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -900,9 +907,21 @@ impl<'gctx> RustcTargetData<'gctx> {
         // needs access to the target config data, create a copy so that it
         // can be found. See `rebuild_unit_graph_shared` for why this is done.
         if requested_kinds.iter().any(CompileKind::is_host) {
-            let ct = CompileTarget::new(&rustc.host)?;
-            target_info.insert(ct, host_info.clone());
-            target_config.insert(ct, gctx.target_cfg_triple(&rustc.host)?);
+            target_config.insert(host_target, gctx.target_cfg_triple(&rustc.host)?);
```
weihanglo, line 911: This looks correct. 

Would you mind sharing why we need the entire infra change (`Arc` and move out rustflags from targetinfo), instead of just the change here?

src/cargo/ops/cargo_compile/unit_generator.rs
```diff
@@ -162,13 +163,15 @@ impl<'a> UnitGenerator<'a, '_> {
                     unit_for,
                     kind,
                 );
+                let kind = kind.for_target(target);
                 self.interner.intern(
                     pkg,
                     target,
                     profile,
                     kind.for_target(target),
```
weihanglo, line 171:
```suggestion
                    kind
```

weihanglo (commented):
Thanks and sorry for the long wait.

Could you follow the ["atomic commit" pattern](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request)? That is, the first commit show the current buggy behavior and pass all tests, followed by the other commits fix both tests and the bug.

src/cargo/core/compiler/build_context/target_info.rs
gmorenz: This question is basically what "How it is implemented" tries to answer in the original comment on this PR. To take another stab at explaining it here:

The `CompileKind` for artifact `Unit` starts out as `CompileKind::Target(host_target)` (assuming we're not cross compiling). The `CompileKind` for host `Unit` starts out as `CompileKind::Host`
  
This change sets up `target_info` so that lookups for `CompileKind::Target(host_target)` get the correct rustflags (as well as units with `CompileKind::Host`, which need different rustflags). Which is great provided we lookup artifact `Unit`'s with their original `CompileKind`.

`rebuild_unit_graph_shared` changes artifact `Unit`s with `CompileKind::Target(host_target)` to have `CompileKind::Host`. It does this so that we can deduplicate identical otherwise identical units and only compile them once\*.

The current infrastructure attempts to look up `rustflags` via `kind` after `rebuild_unit_graph_shared`. Thus even though we have correctly set up the `target_info` for `CompileKind::Target(host_target)` we're going to look up the flags in `host_info`, which are wrong. The infrastructure change is to push this lookup forwards to before we rebuild the unit with a different `kind`. It is also necessary because we deduplicate `Unit`s based on their hash, so even if we could somehow lookup the correct rustflags later, there wouldn't necessarily be a unique set of them (in the case where host and target artifacts have different rustflags, and a shared dependency).

\* To draw your attention to the table at the top. This is why `no --target flag`, `target_applies_to_host=false` has "Without rustflags, built with 5 invocations" rather than 6. A dependency shared between the host artifacts and the target artifacts only has to be built once because it is built the exact same way for both of them.

---

However this question caused me to double check that my understanding was correct by... trying it. I believe it is, however it turns out my test is broken and will (incorrectly) pass (fail to compile) because passing an invalid rustflag will cause `TargetInfo::new(host_target)` to fail because that command calls rustc with rustflags to query file-names/sysroot/... even though we never build a target with those rustflags.

I'll figure out how to write a version of that test without the false-pass on the partial change and push it later tonight.

src/cargo/ops/cargo_compile/unit_generator.rs
gmorenz: Fixed during rebase, thanks!

gmorenz:
> Could you follow the ["atomic commit" pattern](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request)? That is, the first commit show the current buggy behavior and pass all tests, followed by the other commits fix both tests and the bug.

I've done this as requested, as well as incorporation the suggested 1 line change, rebasing on master, and fixing the use of a now-deprecated testing function.

I owe you fixing the test so that it doesn't pass on the partial-fix you asked about in the other review commit.

gmorenz:
Alright, pushed an updated test that passes `--cfg feature="flag"` in RUSTFLAGS instead of an unexpected cfg. This doesn't change the behavior of anything in this branch at all, but it means that partial fix now passes the current-behavior version of the test and doesn't pass the fixed behavior version of the test, because it doesn't error when querying rustc for information prior to building anything.

[The partial fix branch is here](https://github.com/rust-lang/cargo/compare/master...gmorenz:cargo:rustflags_partial_change), just for illustrative purposes.

gmorenz:
Oops, pushing again with fixed formatting in the test in the first commit, and a fixed comment in the test in the second commit (just sloppy rebasing on my part)

src/cargo/core/compiler/build_context/target_info.rs
weihanglo: Thanks. I like the idea that `Unit` holds the exact rustflags it is going to invoke. It opens a new door that we might be able to use this information to create the acutal “build plans” for external systems.

There's only one nit left and I'll merge this :)

weihanglo (approved):

tests/testsuite/rustflags.rs
```diff
@@ -1587,3 +1587,29 @@ fn host_config_rustflags_with_target() {
         .arg("host.rustflags=[\"--cfg=foo\"]")
         .run();
 }
+
+#[cargo_test]
+fn target_applies_to_host_rustflags_works() {
```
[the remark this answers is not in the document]
gmorenz: Sure, just pushed this

weihanglo (approved):

weihanglo:
Thanks a lot!

@bors r+

PR #13907, opened 2024-05-12
fix: support IPv6-only network for cargo fix

### LINKED ISSUE #13906: Diagnostic Server Fails on macOS 14.4
### Problem

```
$ cargo clippy --fix
error: failed to bind TCP listener to manage locking

Caused by:
  Can't assign requested address (os error 49)
```

### Notes

Panic arises in the diagnostic server: https://github.com/rust-lang/cargo/blob/3a18c89a55cbe249e69042b0f51d329726a16d15/src/cargo/util/diagnostic_server.rs#L231

This may be a bug in macOS or rust itself as this simple program fails on my platform:

```
$ cat > example.rs && rustc example.rs && ./example
use std::net::TcpListener;

fn main() {
  let _ = TcpListener::bind("127.0.0.1:0").unwrap();
}
thread 'main' panicked at example.rs:4:44:
called `Result::unwrap()` on an `Err` value: Os { code: 49, kind: AddrNotAvailable, message: "Can't assign requested address" }
```

https://github.com/rust-lang/rust/issues/123715#issuecomment-2106082837

### Version

```text
$ cargo version --verbose                          
cargo 1.78.0 (54d8815d0 2024-03-26)
release: 1.78.0
commit-hash: 54d8815d04fa3816edc207bbc4dd36bf18014dbc
commit-date: 2024-03-26
host: aarch64-apple-darwin
os: Mac OS 14.4.0 [64-bit]
```

weihanglo:

To coordinate diagnositcs between different cargo-as-rustc-proxies,
`cargo fix` launches a diagnostc server on localhost.
However, the TCP server was hard-coded with an IPv4 address 127.0.0.1,
which didn't work with IPv6-only network.

This commit fixes the issue by attempting to bind both IPv4 and IPv6
localhost addessses.

Fixes #13906

I have no idea how to test it. Maybe in Docker follow the [IPv6 network setup guide](https://docs.docker.com/config/daemon/ipv6/), and disable IPv4 stack? Or create a VM that is completely IPv6-only.

I've manually tested it on a macbook. Use at your on risk:

```console
$ ifconfig lo0
# make sure lo0 have both IPv4 and IPv6, and back up if needed
$ ifconfig lo0 inet delete
$ cargo clippy --fix # the original cargo should fail; this patch should pass
$ ifconfig lo0 inet 127.0.0.1 # restore
```

### DIFF
--- a/src/cargo/util/diagnostic_server.rs
+++ b/src/cargo/util/diagnostic_server.rs
@@ -16,6 +16,7 @@ use tracing::warn;
 
 use crate::core::Edition;
 use crate::util::errors::CargoResult;
+use crate::util::network::LOCALHOST;
 use crate::util::GlobalContext;
 
 const DIAGNOSTICS_SERVER_VAR: &str = "__CARGO_FIX_DIAGNOSTICS_SERVER";
@@ -266,7 +267,7 @@ pub struct StartedServer {
 
 impl RustfixDiagnosticServer {
     pub fn new() -> Result<Self, Error> {
-        let listener = TcpListener::bind("127.0.0.1:0")
+        let listener = TcpListener::bind(&LOCALHOST[..])
             .with_context(|| "failed to bind TCP listener to manage locking")?;
         let addr = listener.local_addr()?;
 
--- a/src/cargo/util/lockserver.rs
+++ b/src/cargo/util/lockserver.rs
@@ -20,6 +20,8 @@ use std::thread::{self, JoinHandle};
 
 use anyhow::{Context, Error};
 
+use crate::util::network::LOCALHOST;
+
 pub struct LockServer {
     listener: TcpListener,
     addr: SocketAddr,
@@ -44,7 +46,7 @@ struct ServerClient {
 
 impl LockServer {
     pub fn new() -> Result<LockServer, Error> {
-        let listener = TcpListener::bind("127.0.0.1:0")
+        let listener = TcpListener::bind(&LOCALHOST[..])
             .with_context(|| "failed to bind TCP listener to manage locking")?;
         let addr = listener.local_addr()?;
         Ok(LockServer {
--- a/src/cargo/util/network/mod.rs
+++ b/src/cargo/util/network/mod.rs
@@ -1,11 +1,22 @@
 //! Utilities for networking.
 
+use std::net::Ipv4Addr;
+use std::net::Ipv6Addr;
+use std::net::SocketAddr;
+use std::net::SocketAddrV4;
+use std::net::SocketAddrV6;
 use std::task::Poll;
 
 pub mod http;
 pub mod proxy;
 pub mod retry;
 pub mod sleep;
+
+/// LOCALHOST constants for both IPv4 and IPv6.
+pub const LOCALHOST: [SocketAddr; 2] = [
+    SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, 0)),
+    SocketAddr::V6(SocketAddrV6::new(Ipv6Addr::LOCALHOST, 0, 0, 0)),
+];
 
 pub trait PollExt<T> {
     fn expect(self, msg: &str) -> T;


abhillman:
Wow, nice!

```bash
$ export PATH="/Users/aryehh/Development/cargo/target/release:$PATH"
$ cargo --version
cargo 1.80.0 (ced58bfd6 2024-05-12)
$ cargo clippy --fix
    Updating crates.io index
  Downloaded rowan v0.15.15
  Downloaded countme v3.0.1
  Downloaded strum v0.26.2
  Downloaded rustc-hash v1.1.0
  Downloaded quote v1.0.36
  Downloaded memoffset v0.9.1
  Downloaded text-size v1.1.1
  Downloaded proc-macro2 v1.0.81
  Downloaded serde_derive v1.0.198
  Downloaded serde v1.0.198
  Downloaded rnix v0.11.0
  Downloaded syn v2.0.60
  Downloaded 12 crates (765.5 KB) in 0.25s
   Compiling proc-macro2 v1.0.81
   Compiling unicode-ident v1.0.12
   Compiling serde v1.0.198
   Compiling autocfg v1.2.0
   Compiling serde_json v1.0.116
    Checking countme v3.0.1
    Checking hashbrown v0.14.3
    Checking rustc-hash v1.1.0
    Checking itoa v1.0.11
    Checking ryu v1.0.17
    Checking strum v0.26.2
   Compiling memoffset v0.9.1
   Compiling quote v1.0.36
   Compiling syn v2.0.60
   Compiling serde_derive v1.0.198
    Checking text-size v1.1.1
    Checking rowan v0.15.15
    Checking rnix v0.11.0
    Checking rnix-json v0.1.0 (/Users/aryehh/Development/rnix-json)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.73s
$ ifconfig lo0                                                      
lo0: flags=8049<UP,LOOPBACK,RUNNING,MULTICAST> mtu 16384
	options=1203<RXCSUM,TXCSUM,TXSTATUS,SW_TIMESTAMP>
	inet6 ::1 prefixlen 128 
	inet6 fe80::1%lo0 prefixlen 64 scopeid 0x1 
	nd6 options=201<PERFORMNUD,DAD>
```

abhillman:
For good measure, here is the result I get with the parent commit:

```
$ cargo --version
cargo 1.80.0 (2f17770a1 2024-05-11)
$ cargo clippy --fix
error: failed to bind TCP listener to manage locking

Caused by:
  Can't assign requested address (os error 49)
```

abhillman:
Looks good. I did also check it on linux, but I would recommend adding a regression test to help ensure that this works on other systems -- perhaps something like this would be fine (fails without your patch) https://github.com/rust-lang/cargo/compare/master...abhillman:cargo:diag-server-regression-test.

weihanglo:
> Looks great. I did also check it on linux, but I might recommend adding a regression test to help ensure that this works on other systems -- perhaps something like? Fails without your patch: [master...abhillman:cargo:diag-server-regression-test](https://github.com/rust-lang/cargo/compare/master...abhillman:cargo:diag-server-regression-test). wdyt?

Thanks, though I am not sure. What is the difference between this test and [the entire end-to-end tests for `cargo fix`](https://github.com/rust-lang/cargo/blob/2f17770a15b5e6abe3348d4fcefc08b223e05968/tests/testsuite/fix.rs)? The latter should always create a diagnostic server when exercising.

abhillman:
> > Looks great. I did also check it on linux, but I might recommend adding a regression test to help ensure that this works on other systems -- perhaps something like? Fails without your patch: [master...abhillman:cargo:diag-server-regression-test](https://github.com/rust-lang/cargo/compare/master...abhillman:cargo:diag-server-regression-test). wdyt?
> 
> Thanks, though I am not sure. What is the difference between this test and [the entire end-to-end tests for `cargo fix`](https://github.com/rust-lang/cargo/blob/2f17770a15b5e6abe3348d4fcefc08b223e05968/tests/testsuite/fix.rs)? The latter should always create a diagnostic server when exercising.

Ah! Thanks! I hadn't noticed these! Indeed, they fail without your patch and look more than sufficient 🚀

ehuss:
Thanks! I have not tested this, but seems like it should work.

@bors r+

PR #11631, opened 2023-01-26
Add experimental [cfgs] section to decribe the expected cfgs

Urgau:
This PR adds new section experimental `[cfgs]` to the manifest to describe the expected cfgs and their values. This change was proposed on the Cargo zulip [thread](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/RFC.203013.3A.20Checking.20conditional.20compilation.20at.20compile.20time/near/323550538) without any objection, so I'm proposing it.

The new section is gated under cargo [`-Zcheck-cfg`](https://github.com/rust-lang/cargo/issues/10554) with `cfgs` (`-Zcheck-cfg=cfgs`).

The syntax is straight forward, a simple key/values<sup>*</sup>. <sub>With a custom `values` sub-key for the rare occasion where the values are unknown (empty list != unknown). This could be removed if not desired.</sub>

```toml
[cfgs]
"use_libc" = []                # means `--check-cfg=values('use_libc')`
"loom" = ["yes", "no", "auto"] # means `--check-cfg=values('feature', "yes", "no", "auto")`

[cfgs."unknow_values"] # means `--check-cfg=names('unknow_values')`
values = false         # aka values are unknown (so everything is accepted)
```

### What does this PR try to resolve?

This PR tries to add the missing piece of the all "Checking conditional compilation at compile time", that is, statically defining for a package it's expected cfgs. With all the other [`-Zcheck-cfg` options](https://doc.rust-lang.org/cargo/reference/unstable.html#check-cfg) it was possible to enable checking of well known values, well known names, feature cfgs and add thoses with build script; but there was no way (except `RUSTFLAGS`) to statically set the expected cfgs so that anyone compiling the project could have the unexpected cfgs warnings.

Also by adding this section we also unblock the feature, especially the well know names, which without this section (and probably and edition) would be a breaking change if enable as there would be no way of setting the expected cfgs.

### How should we test and review this PR?

This PR should be reviewed commit by commit and tested with the automated tests and examples.

### Additional information

The implementation use the already existent `check_cfg_args` function that adds the `--check-cfg` args to the `rustc`/`rustdoc` invocation, so nothing new here.

The implementation also use `#[serde(untagged)]` to be able to dezerialize (and serialize?) the `[cfgs]` and `[cfgs.*]` sections into one struct. First time using it, so if you think it's not the right tool, let me know.

#### Questions

 - Should there be any restrictions on the name (like disallowing `feature`) or on the values (like disallowing any value with `"`) ?
[...]

PR #13913, opened 2024-05-14
Add special `check-cfg` lint config for the `unexpected_cfgs` lint

Urgau:

This PR adds a special `check-cfg` lint config for the `unexpected_cfgs` lint, as it was decided by T-cargo (in today's meeting).

The goal of this lint config is to provide a simple and cost-less alternative to the build-script `cargo::rustc-check-cfg` instruction.

```toml
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ['cfg(foo, values("bar"))'] }
```

I recommand reviewing commit by commit; and looking at all the new tests added in `check_cfg.rs`, I tried making them as exhaustive as I could, many of them are very similar to their non-config counterpart.

### Additional information

I didn't add *(actually removed from the 1st version of this PR)* the possibility to omit the `level` field if `check-cfg` is specified, https://github.com/rust-lang/cargo/pull/13913#discussion_r1600609229.

Regarding the implementation, I tried making it is as straight forward as possible, nothing over-engineered or complex.

r? @epage (or @weihanglo maybe)

### DIFF
--- a/crates/cargo-util-schemas/src/manifest/mod.rs
+++ b/crates/cargo-util-schemas/src/manifest/mod.rs
@@ -1503,6 +1503,13 @@ impl TomlLint {
             Self::Config(config) => config.priority,
         }
     }
+
+    pub fn config(&self) -> Option<&toml::Table> {
+        match self {
+            Self::Level(_) => None,
+            Self::Config(config) => Some(&config.config),
+        }
+    }
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
@@ -1511,6 +1518,8 @@ pub struct TomlLintConfig {
     pub level: TomlLintLevel,
     #[serde(default)]
     pub priority: i8,
+    #[serde(flatten)]
+    pub config: toml::Table,
 }
 
 #[derive(Serialize, Deserialize, Debug, Copy, Clone, Eq, PartialEq)]
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1353,12 +1353,37 @@ fn check_cfg_args(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> Vec<OsStri
         // Cargo and docs.rs than rustc and docs.rs. In particular, all users of docs.rs use
         // Cargo, but not all users of rustc (like Rust-for-Linux) use docs.rs.
 
-        vec![
+        let mut args = vec![
             OsString::from("--check-cfg"),
             OsString::from("cfg(docsrs)"),
             OsString::from("--check-cfg"),
             arg_feature,
-        ]
+        ];
+
+        // Also include the custom arguments specified in `[lints.rust.unexpected_cfgs.check_cfg]`
+        if let Ok(Some(lints)) = unit.pkg.manifest().resolved_toml().resolved_lints() {
+            if let Some(rust_lints) = lints.get("rust") {
+                if let Some(unexpected_cfgs) = rust_lints.get("unexpected_cfgs") {
+                    if let Some(config) = unexpected_cfgs.config() {
+                        if let Some(check_cfg) = config.get("check-cfg") {
+                            if let Ok(check_cfgs) =
+                                toml::Value::try_into::<Vec<String>>(check_cfg.clone())
+                            {
+                                for check_cfg in check_cfgs {
+                                    args.push(OsString::from("--check-cfg"));
+                                    args.push(OsString::from(check_cfg));
+                                }
+                            // warn (if wise) about `check-cfg` not being a list-of-string
+                            } else if unit.show_warnings(&build_runner.bcx.gctx) {
+                                let _ = build_runner.bcx.gctx.shell().warn("`lints.rust.unexpected_cfgs.check-cfg` must be a list of string");
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        args
     } else {
         Vec::new()
     }
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -63,7 +63,7 @@ use std::io::{BufRead, Write};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
-use anyhow::{Context as _, Error};
+use anyhow::{bail, Context as _, Error};
 use lazycell::LazyCell;
 use tracing::{debug, trace};
 
@@ -732,7 +732,7 @@ fn prepare_rustdoc(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> CargoResu
     let doc_dir = build_runner.files().out_dir(unit);
     rustdoc.arg("-o").arg(&doc_dir);
     rustdoc.args(&features_args(unit));
-    rustdoc.args(&check_cfg_args(build_runner, unit));
+    rustdoc.args(&check_cfg_args(build_runner, unit)?);
 
     add_error_format_and_color(build_runner, &mut rustdoc);
     add_allow_features(build_runner, &mut rustdoc);
@@ -1125,7 +1125,7 @@ fn build_base_args(
     }
 
     cmd.args(&features_args(unit));
-    cmd.args(&check_cfg_args(build_runner, unit));
+    cmd.args(&check_cfg_args(build_runner, unit)?);
 
     let meta = build_runner.files().metadata(unit);
     cmd.arg("-C").arg(&format!("metadata={}", meta));
@@ -1310,7 +1310,7 @@ fn trim_paths_args(
 }
 
 /// Generates the `--check-cfg` arguments for the `unit`.
-fn check_cfg_args(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> Vec<OsString> {
+fn check_cfg_args(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> CargoResult<Vec<OsString>> {
     if build_runner
         .bcx
         .target_data
@@ -1373,9 +1373,9 @@ fn check_cfg_args(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> Vec<OsStri
                                     args.push(OsString::from("--check-cfg"));
                                     args.push(OsString::from(check_cfg));
                                 }
-                            // warn (if wise) about `check-cfg` not being a list-of-string
-                            } else if unit.show_warnings(&build_runner.bcx.gctx) {
-                                let _ = build_runner.bcx.gctx.shell().warn("`lints.rust.unexpected_cfgs.check-cfg` must be a list of string");
+                            // error about `check-cfg` not being a list-of-string
+                            } else {
+                                bail!("`lints.rust.unexpected_cfgs.check-cfg` must be a list of string");
                             }
                         }
                     }
@@ -1383,9 +1383,9 @@ fn check_cfg_args(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> Vec<OsStri
             }
         }
 
-        args
+        Ok(args)
     } else {
-        Vec::new()
+        Ok(Vec::new())
     }
 }
 
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -2265,7 +2265,7 @@ supported tools: {}",
         if tool == "cargo" && !gctx.cli_unstable().cargo_lints {
             warn_for_cargo_lint_feature(gctx, warnings);
         }
-        for name in lints.keys() {
+        for (name, config) in lints {
             if let Some((prefix, suffix)) = name.split_once("::") {
                 if tool == prefix {
                     anyhow::bail!(
@@ -2277,6 +2277,14 @@ supported tools: {}",
                     )
                 } else {
                     anyhow::bail!("`lints.{tool}.{name}` is not a valid lint name")
+                }
+            } else if let Some(config) = config.config() {
+                for config_name in config.keys() {
+                    // manually report unused manifest key warning since we collect all the "extra"
+                    // keys and values inside the config table
+                    let message =
+                        format!("unused manifest key: `lints.{tool}.{name}.{config_name}`");
+                    warnings.push(message);
                 }
             }
         }
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -2282,9 +2282,14 @@ supported tools: {}",
                 for config_name in config.keys() {
                     // manually report unused manifest key warning since we collect all the "extra"
                     // keys and values inside the config table
-                    let message =
-                        format!("unused manifest key: `lints.{tool}.{name}.{config_name}`");
-                    warnings.push(message);
+                    //
+                    // except for `rust.unexpected_cfgs.check-cfg` which is used by rustc/rustdoc
+                    if !(tool == "rust" && name == "unexpected_cfgs" && config_name == "check-cfg")
+                    {
+                        let message =
+                            format!("unused manifest key: `lints.{tool}.{name}.{config_name}`");
+                        warnings.push(message);
+                    }
                 }
             }
         }
--- a/src/doc/src/reference/build-scripts.md
+++ b/src/doc/src/reference/build-scripts.md
@@ -292,32 +292,6 @@ if foo_bar_condition {
 }
 ```
 
-##### Example of local-only `build.rs`/build script
-
-Build scripts can impose costs on downstream users, and crate authors who wish to avoid
-this can use "local-only" build scripts, which are active locally but not packaged in the
-distributed package. Completly removing the cost from downstream users. 
-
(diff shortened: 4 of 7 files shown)


crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -1503,14 +1503,29 @@ impl TomlLint {
...
+            Self::Config(config) => config.sub_config.as_ref(),
+        }
+    }
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(rename_all = "kebab-case")]
 pub struct TomlLintConfig {
-    pub level: TomlLintLevel,
+    pub level: Option<TomlLintLevel>,
```
epage, line 1518: imo this should be at least a separate commit, probably a separate PR

crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -1503,14 +1503,29 @@ impl TomlLint {
...
     pub priority: i8,
+    #[serde(flatten)]
+    pub sub_config: Option<TomlLintSubConfig>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "kebab-case")]
+pub enum TomlLintSubConfig {
+    CheckCfg(Vec<String>),
 }
```
epage, line 1529:
```suggestion
    #[serde(flatten)]
    pub config: toml::Table,
}
```
- this is lint config, so calling it `sub_config` is weird
- this design forces the every lint to have knowledge about every other lint's config.  Instead, we should treat these as unused keys

src/doc/src/reference/build-scripts.md
```diff
@@ -282,6 +282,16 @@ It is recommended to group the `cargo::rustc-check-cfg` and
...
 
+It is also possible to add to the list of expected config names and values using the special
+[`check-cfg`][lint-table-check-cfg] sub-config under the `[lints.rust.unexpected_cfgs]` lint:
+
+```toml
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ["cfg(foo, values(\"bar\"))"] }
+```
+
+[lint-table-check-cfg]: manifest.md#special-check-cfg
```
epage: With this change, the way I see it working is
- `build.rs` is used if using `cargo::rustc-cfg`
- `Cargo.toml` is used for everything else

In that case, this doesn't really fit in the build script documentation.

src/doc/src/reference/manifest.md
```diff
@@ -589,6 +589,22 @@ Cargo only applies these to the current package and not to dependencies.
...
+
+This special sub-config is only allowed for `[lints.rust.unexpected_cfgs]`, it can be
+used like this:
+
+```toml
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ["cfg(tokio_unstable)"] }
+```
+
+[rustc-check-cfg]: build-scripts.md#rustc-check-cfg
```
epage: Are we expected to be documenting rustc's lint configs or should we be pointing to rustc's docs (once they are updated)?

crates/cargo-util-schemas/src/manifest/mod.rs
Urgau: It's already a separate commit, albeit dependent on the first one, and the reason it's dependant on the first one is to force the `level` if no `check-cfg` is provided, which can only be done when we have knowledge of `check-cfg`.

tests/testsuite/check_cfg.rs
```diff
@@ -521,6 +521,81 @@ fn lint_table_config() {
         .run();
 }
 
+#[cargo_test(>=1.79, reason = "--check-cfg was stabilized in Rust 1.79")]
+fn lint_table_config_and_features() {
```
epage, line 525: could you be more clear in the test name what you are trying to test here.  Is it that `lint_config doesn't cargo native check-cfg's?

epage: (ditto for the other tests)

tests/testsuite/check_cfg.rs
```diff
@@ -521,6 +521,81 @@ fn lint_table_config() {
...
+
+    p.cargo("check -v")
+        .with_stderr_contains(x!("rustc" => "cfg" of "has_foo"))
+        .with_stderr_contains(x!("rustc" => "cfg" of "has_bar" with "yes" "no"))
+        .with_stderr_contains(x!("rustc" => "cfg" of "feature" with "alloc" "my_feature"))
+        .run();
+}
+
+#[cargo_test(>=1.79, reason = "--check-cfg was stabilized in Rust 1.79")]
+fn lint_table_config_doc() {
```
epage, line 554: Looks like these are to mirror tests with `build.rs`.  Maybe mention that in the commit message?

tests/testsuite/check_cfg.rs
epage: Also,. if thats the case, should we have `build.rs` / lint config tests?

tests/testsuite/lints_table.rs
```diff
@@ -150,6 +150,7 @@ fn invalid_type_in_lint_value() {
 }
 
 #[cargo_test]
+#[ignore = "currently broken by serde(flatten) on TomlLintConfig"]
```
epage, line 153: Please fix

src/doc/src/reference/build-scripts.md
Urgau: Ideally yes, it shouldn't be in the build script documentation but we currently don't have a canonical place for check-cfg (`rustc` link directly to this doc) and more importantly I see this as an alternative to the instruction, so mentioning it doesn't seems that out of place.

crates/cargo-util-schemas/src/manifest/mod.rs
epage: That is a detail too buried in everything which makes me feel more strongly this needs to be its own PR to make sure we know what semantics we are agreeing to.

src/doc/src/reference/build-scripts.md
epage: > Ideally yes, it shouldn't be in the build script documentation but we currently don't have a canonical place for check-cfg (rustc link directly to this doc) 

This is a rust lint config and belongs in their documentation.  

> I see this as an alternative to the instruction, so mentioning it doesn't seems that out of place.

I gave reasons for why I disagree.  Could you address those?

crates/cargo-util-schemas/src/manifest/mod.rs
Urgau: Will do that in a follow-up PR.

crates/cargo-util-schemas/src/manifest/mod.rs
Urgau: It's now called `config` and is now as asked completely agnostic.

src/doc/src/reference/build-scripts.md
Urgau: I've removed this documentation update, the build-script doc should only be about build-script.

tests/testsuite/check_cfg.rs
Urgau: I've renamed most tests to be a more descriptive. I add more description is desired.

tests/testsuite/lints_table.rs
Urgau: Done, but I'm not 100% satisfied with the result.

src/doc/src/reference/manifest.md
Urgau: I've also removed this documentation update.

weihanglo:
> 💔 Test failed - [checks-actions](https://github.com/rust-lang/cargo/actions/runs/9103372526/job/25025359548)

```
Updating git repository `[https://127.0.0.1:32770/repos/bar.git`](https://127.0.0.1:32770/repos/bar.git%60)
warning: spurious network error (3 tries remaining): failed to receive HTTP 200 response: got 500; class=Net (12)
warning: spurious network error (2 tries remaining): failed to receive HTTP 200 response: got 500; class=Net (12)
warning: spurious network error (1 tries remaining): failed to receive HTTP 200 response: got 500; class=Net (12)
error: failed to get `bar` as a dependency of package `foo v0.1.0 (/home/runner/work/cargo/cargo/target/tmp/cit/t1616/foo)`

Caused by:

Caused by:
  Unable to update https://127.0.0.1:32770/repos/bar.git

Caused by:
  failed to clone into: /home/runner/work/cargo/cargo/target/tmp/cit/t1616/home/.cargo/git/db/bar-f70683a84217eb2c

Caused by:
  network failure seems to have happened
  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here

Caused by:
  failed to receive HTTP 200 response: got 500; class=Net (12)
```

A container test failed. Switch from [latest apline (3.19)](https://github.com/rust-lang/cargo/blob/fc13634f78023381b55452fa7f4d7a974449a5e8/crates/cargo-test-support/containers/apache/Dockerfile#L1) to `httpd:2.4-alpine3.18` works. I guessed that might be some package has changed, and found that `git` [^git] was bumped earlier today (2024-05-15) for fixing security vulnerabilities[^vuln].

Still investigating. If I cannot make it on time, I'll file a PR to use apline 3.18 temporarily.


[^git]: https://git.alpinelinux.org/aports/commit/?id=f9dddd8074787cca180745095c9cdd14cfc60b39
[^vuln]: https://github.blog/2024-05-14-securing-git-addressing-5-new-vulnerabilities/

src/cargo/core/compiler/mod.rs
```diff
@@ -1353,12 +1353,37 @@ fn check_cfg_args(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> Vec<OsStri
...
+                            {
+                                for check_cfg in check_cfgs {
+                                    args.push(OsString::from("--check-cfg"));
+                                    args.push(OsString::from(check_cfg));
+                                }
+                            // warn (if wise) about `check-cfg` not being a list-of-string
+                            } else if unit.show_warnings(&build_runner.bcx.gctx) {
+                                let _ = build_runner.bcx.gctx.shell().warn("`lints.rust.unexpected_cfgs.check-cfg` must be a list of string");
+                            }
+                        }
```
epage, line 1380: Should this be an error instead?

src/cargo/util/toml/mod.rs
```diff
@@ -2282,9 +2282,14 @@ supported tools: {}",
...
-                        format!("unused manifest key: `lints.{tool}.{name}.{config_name}`");
-                    warnings.push(message);
+                    //
+                    // except for `rust.unexpected_cfgs.check-cfg` which is used by rustc/rustdoc
+                    if !(tool == "rust" && name == "unexpected_cfgs" && config_name == "check-cfg")
+                    {
+                        let message =
+                            format!("unused manifest key: `lints.{tool}.{name}.{config_name}`");
+                        warnings.push(message);
+                    }
```
epage, line 2292: For where we are at with this effort, I think this is a good enough solution

src/cargo/core/compiler/mod.rs
Urgau: Could be, I'm just being cautious here, writing `check-cfg` is currently a warning, but I doubt anyone has written it before, so it could probably be a error.

epage: We can always relax the error into a warning in the future but we can't do the other, so probably safer to start with an error.

Urgau: Should in be an unconditional error? or also be gated on `unit.show_warnings`?

Hum, 🤔, it would be weird if it isn't unconditional ?

epage: Let's start unconditional.

Urgau: Sure. Done in https://github.com/rust-lang/cargo/pull/13913/commits/980afaabd484af10150089b2d7ed8ba224f2e552.

src/cargo/core/compiler/mod.rs
```diff
@@ -1353,14 +1353,39 @@ fn check_cfg_args(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> Vec<OsStri
...
+        let mut args = vec![
             OsString::from("--check-cfg"),
             OsString::from("cfg(docsrs)"),
             OsString::from("--check-cfg"),
             arg_feature,
-        ]
+        ];
+
+        // Also include the custom arguments specified in `[lints.rust.unexpected_cfgs.check_cfg]`
+        if let Ok(Some(lints)) = unit.pkg.manifest().resolved_toml().resolved_lints() {
```
epage, line 1364: We should remove the ["local only"](https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#example-of-local-only-buildrsbuild-script) build script example as this is meant to target that use case?

Urgau: Yes, I think, I don't see any reason to use a "local only" build script over this new config. Removed.

epage:
@rfcbot fcp merge

This is a straw poll (no 10 day waiting period) to get a more official sign off on this solution to some of the concerns raised in rust-lang/rust#124800

This adds support for `lints.rust.unexpected_cfgs.check-cfg` as a declarative alternative to `build.rs`.  My assumption with this is `cargo::rustc-check-cfg` should be used if `cargo::rustc-cfg` is used, otherwise `lints.rust.unexpected_cfgs.check-cfg` should be used.

While we had rejected a `[cfg]` table in https://github.com/rust-lang/cargo/pull/11631#issuecomment-1487424886, the supposition here is that this strikes a better balance by being a low level mechanism tied to rustc.  Longer term, we are hoping [global features](https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618) can reduce the need for this.

Example:
```toml
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ["cfg(has_foo)", "cfg(has_bar, values(\"yes\", \"no\"))"] }
```
- To keep this PR focused, `level` is still required.

Follow up work
- Update `unexpected_cfgs` docs to mention this
- Generally scour docs and error messages to make sure they made sense
- Consider making `level` optional

wesleywiser:
Does whatever library Cargo uses for TOML parsing support [literal strings](https://toml.io/en/#:~:text=no%20line%20breaks\).-,Literal%20strings,-are%20surrounded%20by)? If so, it seems that we should document and recomend people use those instead of needing to escape every `"` within the check configuration argument. 

eg:

```toml
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ['cfg(has_foo)', 'cfg(has_bar, values("yes", "no"))'] }
```

Urgau:
> Does whatever library Cargo uses for TOML parsing support literal strings?

It does, I just checked locally. That is indeed much better to suggest.

I will make sure we suggest string literal in the diagnostic output of the lint. (I already have the rust-lang/rust PR in the works)

epage:
An idea that came up in Office Hour when talking about this is doing beta "backport" that disables the unused manifest key warning just for `lints.rust.unexpected_cfgs.check-cfg` so to reduce the time period of people facing the dilemma of seeing warnings on nightly or stable.

If we want to buy us more time, we could switch the lint to `allow` for a couple of releases.

epage:
Brought up quoting in our [long-running style thread](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/.60Cargo.2Etoml.60.20style.20guide)

epage:
I don't know why but I just noticed the inconsistent use of `-`: `lints.rust.unexpected_cfgs.check-cfg`.  Wanted to call that out incase it bothers anyone else.

src/cargo/core/compiler/mod.rs
```diff
@@ -1353,14 +1353,39 @@ fn check_cfg_args(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> Vec<OsStri
...
+            if let Some(rust_lints) = lints.get("rust") {
+                if let Some(unexpected_cfgs) = rust_lints.get("unexpected_cfgs") {
```
epage, line 1366: ... speaking of inconsistent `-` / `_`.  We didn't do anything to force `-` or `_` in `lints` so you could have `lints.rust.unexpected-cfgs.check-cfg` and it would not be read but instead give an unused key warning

epage:
> I don't know why but I just noticed the inconsistent use of -: lints.rust.unexpected_cfgs.check-cfg. Wanted to call that out incase it bothers anyone else.

We talked about this as a team.  Overall, there weren't strong thoughts one way or the other on this.

One re-framing for myself was that this is similar to features and dependencies which can have inconsistent word separators from the rest of the format.

src/cargo/core/compiler/mod.rs
epage: See #13943

jhpratt:
> I don't know why but I just noticed the inconsistent use of `-`: `lints.rust.unexpected_cfgs.check-cfg`. Wanted to call that out incase it bothers anyone else.

I will note that when using `unexpected-cfgs`, `check-cfg` is not known. This is unlike every other lint config.

PR #13916, opened 2024-05-15
Verify build-std resolve against original lockfile

adamgemmell:

Ensures that the resolve of the standard library when using build-std is a subset of the lockfile from the library source, as per discussion on [rust-lang/wg-cargo-std-aware#38](https://github.com/rust-lang/wg-cargo-std-aware/issues/38). It is a runtime version of the test added here: https://github.com/rust-lang/cargo/pull/13404

Fixes https://github.com/rust-lang/wg-cargo-std-aware/issues/38

It can be manually tested by running cargo build -Zbuild-std --target=[host] and seeing the command succeed. I'm not sure what behaviour can exercise the negative test case.

### Additional information

Co-authored-by: Lawrence Tang [lawrence.tang@arm.com](mailto:lawrence.tang@arm.com)

### DIFF
--- a/src/cargo/core/compiler/standard_lib.rs
+++ b/src/cargo/core/compiler/standard_lib.rs
@@ -5,7 +5,7 @@ use crate::core::compiler::UnitInterner;
 use crate::core::compiler::{CompileKind, CompileMode, RustcTargetData, Unit};
 use crate::core::profiles::{Profiles, UnitFor};
 use crate::core::resolver::features::{CliFeatures, FeaturesFor, ResolvedFeatures};
-use crate::core::resolver::HasDevUnits;
+use crate::core::resolver::{EncodableResolve, HasDevUnits};
 use crate::core::{Dependency, PackageId, PackageSet, Resolve, SourceId, Workspace};
 use crate::ops::{self, Packages};
 use crate::util::errors::CargoResult;
@@ -125,9 +125,12 @@ pub fn resolve_std<'gctx>(
     // now. Perhaps in the future features will be decoupled from the resolver
     // and it will be easier to control feature selection.
     let current_manifest = src_path.join("library/sysroot/Cargo.toml");
-    // TODO: Consider doing something to enforce --locked? Or to prevent the
-    // lock file from being written, such as setting ephemeral.
-    let mut std_ws = Workspace::new_virtual(src_path, current_manifest, virtual_manifest, gctx)?;
+    let mut std_ws =
+        Workspace::new_virtual(src_path.clone(), current_manifest, virtual_manifest, gctx)?;
+
+    // The source checkout isn't managed by us, so setting ephemeral here
+    // ensures the lockfile isn't updated.
+    std_ws.set_ephemeral(true);
     // Don't require optional dependencies in this workspace, aka std's own
     // `[dev-dependencies]`. No need for us to generate a `Resolve` which has
     // those included because we'll never use them anyway.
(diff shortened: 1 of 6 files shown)


src/cargo/core/compiler/standard_lib.rs
```diff
@@ -125,9 +125,12 @@ pub fn resolve_std<'gctx>(
...
     let current_manifest = src_path.join("library/sysroot/Cargo.toml");
-    // TODO: Consider doing something to enforce --locked? Or to prevent the
-    // lock file from being written, such as setting ephemeral.
-    let mut std_ws = Workspace::new_virtual(src_path, current_manifest, virtual_manifest, gctx)?;
+    let mut std_ws =
+        Workspace::new_virtual(src_path.clone(), current_manifest, virtual_manifest, gctx)?;
+
+    // The source checkout isn't managed by us, so setting ephemeral here
+    // ensures the lockfile isn't updated.
+    std_ws.set_ephemeral(true);
```
epage, line 133: Could this be split out into a separate commit or even separate PR?  Seems like this is fairly important we shouldn't hide it away in these other changes

src/cargo/core/compiler/standard_lib.rs
```diff
@@ -158,6 +161,47 @@ pub fn resolve_std<'gctx>(
         HasDevUnits::No,
         crate::core::resolver::features::ForceAllTargets::No,
     )?;
+
+    // Verify that we have resolved to a subset of the lockfile
+    let lockfile = std::fs::read_to_string(&src_path.join("Cargo.lock"))
+        .expect("Couldn't read the Rust source's lockfile");
+
+    let encoded_lockfile: EncodableResolve = toml::from_str(&lockfile).unwrap();
+    let lockfile_packages = encoded_lockfile.package().expect("libstd has no packages!");
```
epage, line 170: The general practice I've heard and prefer for `expect` is it should be listing the invariant you are relying on

src/cargo/core/compiler/standard_lib.rs
```diff
@@ -158,6 +161,47 @@ pub fn resolve_std<'gctx>(
         HasDevUnits::No,
         crate::core::resolver::features::ForceAllTargets::No,
     )?;
+
+    // Verify that we have resolved to a subset of the lockfile
+    let lockfile = std::fs::read_to_string(&src_path.join("Cargo.lock"))
+        .expect("Couldn't read the Rust source's lockfile");
+
+    let encoded_lockfile: EncodableResolve = toml::from_str(&lockfile).unwrap();
```
epage, line 169: We should probably list our invariant for this

src/cargo/core/compiler/standard_lib.rs
```diff
@@ -158,6 +161,47 @@ pub fn resolve_std<'gctx>(
...
+    let lockfile_packages = encoded_lockfile.package().expect("libstd has no packages!");
+
+    for resolved_pkg in resolve.targeted_resolve.iter() {
+        let pkg_name = resolved_pkg.name().to_string();
+        let pkg_ver = resolved_pkg.version().to_string();
```
epage, line 174: why do we need to `to_string` these?

src/cargo/core/compiler/standard_lib.rs
```diff
@@ -158,6 +161,47 @@ pub fn resolve_std<'gctx>(
...
+                        pkg.name() == dep.name().as_str()
+                        && (pkg.version() == None
+                            || pkg.version() == Some(dep.version().to_string().as_str()))
+                    })
+                })
+                .unwrap_or_else(||
+                    panic!("Standard library's package graph changed during resolution:
+                           Package '{}' ({}) was present as a dependency of '{} ({}) in the resolve but not in the lockfile",
+                           dep.name(), dep.version(), resolved_pkg.name(), resolved_pkg.version())
+                );
```
epage, line 201: Do we need to check the dependencies?  I would assume that so long as the package is deemed the same, the dependencies should be the same.

@Eh2406 in case you know of a reason otherwise.

adamgemmell: Say rustc and std depend on the same package, but rustc enabled an optional dependency (a). I believe they'd resolve to the same package (with an edge to a) in the lockfile, but the std resolve in build-std would be free to eliminate the edge to a.

This is fine in the patch since the edge exists in the lockfile but not in the resolve - the reverse situation is an error.

src/cargo/core/compiler/standard_lib.rs
```diff
@@ -158,6 +161,47 @@ pub fn resolve_std<'gctx>(
...
+    let encoded_lockfile: EncodableResolve = toml::from_str(&lockfile).unwrap();
+    let lockfile_packages = encoded_lockfile.package().expect("libstd has no packages!");
+
+    for resolved_pkg in resolve.targeted_resolve.iter() {
+        let pkg_name = resolved_pkg.name().to_string();
+        let pkg_ver = resolved_pkg.version().to_string();
+        let lockfile_pkg = lockfile_packages.binary_search_by_key(
+                &(pkg_name.as_str(), pkg_ver.as_str()),
+                |p| (p.name(), p.version())
+            )
```
epage, line 178: Shouldn't we also be checking the `source`?

src/cargo/core/compiler/standard_lib.rs
```diff
@@ -158,6 +161,47 @@ pub fn resolve_std<'gctx>(
...
     )?;
+
+    // Verify that we have resolved to a subset of the lockfile
+    let lockfile = std::fs::read_to_string(&src_path.join("Cargo.lock"))
+        .expect("Couldn't read the Rust source's lockfile");
+
+    let encoded_lockfile: EncodableResolve = toml::from_str(&lockfile).unwrap();
+    let lockfile_packages = encoded_lockfile.package().expect("libstd has no packages!");
+
+    for resolved_pkg in resolve.targeted_resolve.iter() {
```
epage, line 172: `resolved_pkg` is a PackageId, right?  Could we clarify that by renaming it to `resolved_pkgid`?

src/cargo/core/compiler/standard_lib.rs
```diff
@@ -158,6 +161,47 @@ pub fn resolve_std<'gctx>(
...
+            .and_then(|idx| Ok(&lockfile_packages[idx]))
+            .unwrap_or_else(|_|
+                panic!("Standard library's package graph changed during resolution:
+                       Package '{}' ({}) was present in the resolve but not in the original lockfile",
+                       resolved_pkg.name(), resolved_pkg.version())
+            );
+        // If the lockfile wasn't sorted then the binary search result can be nonsensical
+        assert!(lockfile_pkg.name() == pkg_name);
+
+        for (dep, _) in resolve.targeted_resolve.deps(resolved_pkg) {
```
epage, line 188: ditto about `dep` -> `dep_id`

epage: Maybe even `resolved_dep_id`

src/cargo/core/compiler/standard_lib.rs
epage: My first thought was that maybe we should have a comment that we are checking for extra features being activated. As an alternative, my second thought is to update the panic to clarify that.

PR #13921, opened 2024-05-16
fix(toml): Convert warnings that `licence` and `readme` files do not exist into errors

heisen-li:

In this PR:
- Changed the warning to a hard error and modified the associated test function;
- Removed what should have been a redundant test function：`publish::publish_with_missing_readme`；
- Since `cargo publish` is preceded by the execution of `cargo package`, the error message in the test `function bad_license_file` needs to be modified.

issue: https://github.com/rust-lang/cargo/issues/13629#license-file-and-readme-pointing-to-a-non-existent-file.

### Additional information

It seems that this is not enough, the current situation is that `cargo package` warns if `package.readme` is an empty string or the wrong file location, but if I cancel `package.readme`, no warning is generated.

I'm wondering if I should judge `package.readme&licence` when executing `cargo package` and return an error if it doesn't exist?

As this has not been done before, your advice is sought.

### DIFF
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -310,6 +310,8 @@ fn build_ar_list(
             });
     }
 
+    let mut invalid_manifest_field: Vec<String> = vec![];
+
     let mut result = result.into_values().flatten().collect();
     if let Some(license_file) = &pkg.manifest().metadata().license_file {
         let license_path = Path::new(license_file);
@@ -324,7 +326,12 @@ fn build_ar_list(
                 ws,
             )?;
         } else {
-            warn_on_nonexistent_file(&pkg, &license_path, "license-file", &ws)?;
+            error_on_nonexistent_file(
+                &pkg,
+                &license_path,
+                "license-file",
+                &mut invalid_manifest_field,
+            );
         }
     }
     if let Some(readme) = &pkg.manifest().metadata().readme {
@@ -333,8 +340,12 @@ fn build_ar_list(
         if abs_file_path.is_file() {
             check_for_file_and_add("readme", readme_path, abs_file_path, pkg, &mut result, ws)?;
         } else {
-            warn_on_nonexistent_file(&pkg, &readme_path, "readme", &ws)?;
+            error_on_nonexistent_file(&pkg, &readme_path, "readme", &mut invalid_manifest_field);
         }
+    }
+
+    if !invalid_manifest_field.is_empty() {
+        return Err(anyhow::anyhow!(invalid_manifest_field.join("\n")));
     }
 
     for t in pkg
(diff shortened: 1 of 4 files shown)


src/cargo/ops/cargo_package.rs
```diff
@@ -500,11 +513,7 @@ fn check_metadata(pkg: &Package, gctx: &GlobalContext) -> CargoResult<()> {
             )*
         }}
     }
-    lacking!(
-        description,
-        license || license_file,
-        documentation || homepage || repository
-    );
+    lacking!(description, documentation || homepage || repository);
```
weihanglo, line 516: Why did `license || license_file` part get removed? It is still required to have license specified.

src/cargo/ops/cargo_package.rs
```diff
@@ -406,25 +417,27 @@ fn check_for_file_and_add(
...
+    invalid: &mut Vec<String>,
+) {
     let rel_msg = if path.is_absolute() {
         "".to_string()
     } else {
         format!(" (relative to `{}`)", pkg.root().display())
     };
-    ws.gctx().shell().warn(&format!(
+
+    let msg = format!(
```
weihanglo, line 432: Cargo checks only two fields, so it is fine just returning an `Err` here . That would make the code easier to reason.

tests/testsuite/package.rs
weihanglo, line 1: >    Hard error, even for empty string

Should we add test checking empty string like `license-file = ""` or `readme = ""` will fail?

src/cargo/ops/cargo_package.rs
heisen-li: I thought not specifying `license || license_file` was supposed to be an error and not a warning?

src/cargo/ops/cargo_package.rs
heisen-li: It would certainly be possible to return `Err` directly in the function `error_on_nonexistent_file`, but I think if neither `licence` nor `readme` exists, then they should all be prompted at the same time, not one by one.

And it's also easier to expand the other options later on as required fields as well. I hope my understanding is reasonable, thank you.

tests/testsuite/package.rs
heisen-li: Sorry, it seems that there is already a corresponding test in the test function.

[fn empty_readme_path()](https://github.com/rust-lang/cargo/pull/13921/files#diff-536dea47bdf1670f771547855b858c1507baad3565eefe663ada0b1954134b7dR1944)
[fn empty_license_file_path](https://github.com/rust-lang/cargo/pull/13921/files#diff-536dea47bdf1670f771547855b858c1507baad3565eefe663ada0b1954134b7dR2039)

src/cargo/ops/cargo_package.rs
heisen-li: Thanks. It has returned to its previous state.

weihanglo (approved):
Thanks. Also schedule this to merge after 1.81 nightly comes out.

weihanglo (approved):

Issue #10948, opened 2022-08-08
cargo package --workspace is not very useful 


conradludgate:
### Tracking

Unstable flag: `-Zpackage-workspace`

Stabilizing this would also close
- #1169

Testing instructions: https://github.com/rust-lang/cargo/issues/10948#issuecomment-2540365225

Implementation
- [x] #13947
- [x] #14433
- [x] #14659
  - [ ] When stabilizing, do a clean up of d30fde980f5 man page duplicates. 

Changes since original plan
- Added`--registry` and `--index` flags to `cargo package` to know what registry will be used for generating the `Cargo.lock` file as if the internal dependencies were already published
- `cargo publish` is not atomic but it does verify all before publish

Open questions
- Are we ok with a slight compatibility breakage in `cargo package`?  See https://github.com/rust-lang/cargo/issues/10948#issuecomment-2253530842
- Are we ok stabilizing this and #1169 at the same time?  Currently, they are behind the same flag
- What is the desired behavior for the publish timeout?  #14433 uploads the crates in batches (depending on the dependency graph), and we only timeout if nothing in the batch is available within the timeout, deferring the rest to the next wait-for-publish. So for example, if you have packages `a`, `b`, `c` then we'll wait up to 60 seconds and if only `a` and `b` were ready in that time, we'll then wait another 60 seconds for `c`.
- What is the desired behavior when publishing some packages in a workspace that have `publish = false`? #14433 raises an error whenever any of the selected packages has `publish = false`, so it will error on `cargo publish --workspace` in a workspace with an unpublishable package. An alternative interface would implicitly exclude unpublishable packages in this case, but still error out if you explicitly select an unpublishable package with `-p package-name` (see #14356). #14433 behavior is the most conservative one as it can change from an error to implicit excludes later.

Known issues
- #14396
- #15412

Non-blocking issues
- #14941

### Problem

Let's say you have a workspace

```
workspace/
  package_a/
    Cargo.toml (name="a", version="0.1.0")
  package_b/
    Cargo.toml (name="a", version="0.1.0", [dependencies] a="0.1.0")
  Cargo.toml (members = ["package_a","package_b"])
```

If you have already published `a` to crates.io, then `cargo package --workspace` will complete successfully.

Now, you update `a` to `0.1.1`, and update `b` to use that new minimum version. If you try `cargo package --workspace` again, it will no longer work. You will get an error that `0.1.1` was not found.

[...]

PR #13926, opened 2024-05-17
Add local registry overlays

jneem:
This PR adds (private to cargo internals) support for local registry overlays, in which you can locally pretend to add packages to remote registries; the local packages will have the same source ids as the remote registry that you're overlaying.

There are two ways to set up these overlays: programmatically using `GlobalContext::local_overlays` and through the `__CARGO_TEST_PACKAGE_CONFUSION_VULNERABILITY_DO_NOT_USE_THIS` environment variable. You can't set up these overlays with `.cargo/config`.

The motivation for this is [packaging workspaces](https://github.com/rust-lang/cargo/issues/10948). When we're packing a workspace, we'd like to be able to pretend (for lockfile generation and verification) that some workspace packages are already published even though they aren't.

### DIFF
--- a/crates/cargo-test-support/src/lib.rs
+++ b/crates/cargo-test-support/src/lib.rs
@@ -877,6 +877,17 @@ impl Execs {
         self
     }
 
+    pub fn overlay_registry(&mut self, url: &Url, path: &str) -> &mut Self {
+        if let Some(ref mut p) = self.process_builder {
+            let env_value = format!("{}={}", url, path);
+            p.env(
+                "__CARGO_TEST_DEPENDENCY_CONFUSION_VULNERABILITY_DO_NOT_USE_THIS",
+                env_value,
+            );
+        }
+        self
+    }
+
     pub fn enable_split_debuginfo_packed(&mut self) -> &mut Self {
         self.env("CARGO_PROFILE_DEV_SPLIT_DEBUGINFO", "packed")
             .env("CARGO_PROFILE_TEST_SPLIT_DEBUGINFO", "packed")
--- a/crates/cargo-test-support/src/registry.rs
+++ b/crates/cargo-test-support/src/registry.rs
@@ -1649,7 +1649,12 @@ impl Package {
     /// Returns the path to the compressed package file.
     pub fn archive_dst(&self) -> PathBuf {
         if self.local {
-            registry_path().join(format!("{}-{}.crate", self.name, self.vers))
+            let path = if self.alternative {
+                alt_registry_path()
+            } else {
+                registry_path()
+            };
+            path.join(format!("{}-{}.crate", self.name, self.vers))
         } else if self.alternative {
             alt_dl_path()
                 .join(&self.name)
--- a/src/cargo/ops/resolve.rs
+++ b/src/cargo/ops/resolve.rs
@@ -116,7 +116,7 @@ version. This may also occur with an optional dependency that is not enabled.";
 /// This is a simple interface used by commands like `clean`, `fetch`, and
 /// `package`, which don't specify any options or features.
 pub fn resolve_ws<'a>(ws: &Workspace<'a>, dry_run: bool) -> CargoResult<(PackageSet<'a>, Resolve)> {
-    let mut registry = PackageRegistry::new(ws.gctx())?;
+    let mut registry = ws.package_registry()?;
     let resolve = resolve_with_registry(ws, &mut registry, dry_run)?;
     let packages = get_resolved_packages(&resolve, registry)?;
     Ok((packages, resolve))
@@ -142,7 +142,7 @@ pub fn resolve_ws_with_opts<'gctx>(
     force_all_targets: ForceAllTargets,
     dry_run: bool,
 ) -> CargoResult<WorkspaceResolve<'gctx>> {
-    let mut registry = PackageRegistry::new(ws.gctx())?;
+    let mut registry = ws.package_registry()?;
     let (resolve, resolved_with_overrides) = if ws.ignore_lock() {
         let add_patches = true;
         let resolve = None;
--- a/src/cargo/sources/config.rs
+++ b/src/cargo/sources/config.rs
@@ -5,6 +5,7 @@
 //! sources to one another via the `replace-with` key in `.cargo/config`.
 
 use crate::core::{GitReference, PackageId, SourceId};
+use crate::sources::overlay::DependencyConfusionThreatOverlaySource;
 use crate::sources::source::Source;
 use crate::sources::{ReplacedSource, CRATES_IO_REGISTRY};
 use crate::util::context::{self, ConfigRelativePath, OptValue};
@@ -24,6 +25,8 @@ pub struct SourceConfigMap<'gctx> {
     cfgs: HashMap<String, SourceConfig>,
     /// Mapping of [`SourceId`] to the source name.
     id2name: HashMap<SourceId, String>,
+    /// Mapping of sources to local registries that will be overlaid on them.
+    overlays: HashMap<SourceId, SourceId>,
     gctx: &'gctx GlobalContext,
 }
 
@@ -81,6 +84,18 @@ impl<'gctx> SourceConfigMap<'gctx> {
                 base.add_config(key, value)?;
             }
         }
+
+        Ok(base)
+    }
+
+    /// Like [`SourceConfigMap::new`] but includes sources from source
+    /// replacement configurations.
+    pub fn new_with_overlays(
+        gctx: &'gctx GlobalContext,
+        overlays: impl IntoIterator<Item = (SourceId, SourceId)>,
+    ) -> CargoResult<SourceConfigMap<'gctx>> {
+        let mut base = SourceConfigMap::new(gctx)?;
+        base.overlays = overlays.into_iter().collect();
         Ok(base)
     }
 
@@ -90,6 +105,7 @@ impl<'gctx> SourceConfigMap<'gctx> {
         let mut base = SourceConfigMap {
             cfgs: HashMap::new(),
             id2name: HashMap::new(),
+            overlays: HashMap::new(),
             gctx,
         };
         base.add(
@@ -136,7 +152,7 @@ impl<'gctx> SourceConfigMap<'gctx> {
         debug!("loading: {}", id);
 
         let Some(mut name) = self.id2name.get(&id) else {
-            return id.load(self.gctx, yanked_whitelist);
+            return self.load_overlaid(id, yanked_whitelist);
         };
         let mut cfg_loc = "";
         let orig_name = name;
@@ -161,7 +177,7 @@ impl<'gctx> SourceConfigMap<'gctx> {
                     name = s;
                     cfg_loc = c;
                 }
-                None if id == cfg.id => return id.load(self.gctx, yanked_whitelist),
+                None if id == cfg.id => return self.load_overlaid(id, yanked_whitelist),
                 None => {
                     break cfg.id.with_precise_from(id);
                 }
@@ -178,8 +194,8 @@ impl<'gctx> SourceConfigMap<'gctx> {
             }
         };
 
-        let new_src = new_id.load(
-            self.gctx,
+        let new_src = self.load_overlaid(
+            new_id,
             &yanked_whitelist
                 .iter()
                 .map(|p| p.map_source(id, new_id))
@@ -213,6 +229,23 @@ restore the source replacement configuration to continue the build
         }
 
         Ok(Box::new(ReplacedSource::new(id, new_id, new_src)))
+    }
+
+    /// Gets the [`Source`] for a given [`SourceId`] without performing any source replacement.
+    fn load_overlaid(
+        &self,
+        id: SourceId,
+        yanked_whitelist: &HashSet<PackageId>,
+    ) -> CargoResult<Box<dyn Source + 'gctx>> {
+        let src = id.load(self.gctx, yanked_whitelist)?;
+        if let Some(overlay_id) = self.overlays.get(&id) {
+            let overlay = overlay_id.load(self.gctx(), yanked_whitelist)?;
+            Ok(Box::new(DependencyConfusionThreatOverlaySource::new(
+                overlay, src,
+            )))
+        } else {
+            Ok(src)
+        }
     }
 
     /// Adds a source config with an associated name.
--- a/src/cargo/sources/overlay.rs
+++ b/src/cargo/sources/overlay.rs
@@ -1,0 +1,148 @@
+use std::task::ready;
+
+use tracing::debug;
+
+use crate::sources::IndexSummary;
+
+use super::source::{MaybePackage, Source};
+
+/// A `Source` that overlays one source over another, pretending that the packages
+/// available in the overlay are actually available in the other one.
+///
+/// This is a massive footgun and a terrible idea, so we do not (and never will)
+/// expose this publicly. However, it is useful for some very specific private
+/// things, like locally verifying a bunch of packages at a time before any of
+/// them have been published.
+pub struct DependencyConfusionThreatOverlaySource<'gctx> {
+    // The overlay source. The naming here comes from the main application of this,
+    // where there is a remote registry that we overlay some local packages on.
+    local: Box<dyn Source + 'gctx>,
+    // The source we're impersonating.
+    remote: Box<dyn Source + 'gctx>,
+}
+
+impl<'gctx> DependencyConfusionThreatOverlaySource<'gctx> {
+    pub fn new(local: Box<dyn Source + 'gctx>, remote: Box<dyn Source + 'gctx>) -> Self {
+        debug!(
+            "overlaying {} on {}",
+            local.source_id().as_url(),
+            remote.source_id().as_url()
+        );
+        Self { local, remote }
+    }
+}
+
+impl<'gctx> Source for DependencyConfusionThreatOverlaySource<'gctx> {
+    fn source_id(&self) -> crate::core::SourceId {
+        self.remote.source_id()
+    }
+
+    fn supports_checksums(&self) -> bool {
+        self.local.supports_checksums() && self.remote.supports_checksums()
+    }
+
+    fn requires_precise(&self) -> bool {
+        self.local.requires_precise() || self.remote.requires_precise()
+    }
+
+    fn query(
+        &mut self,
+        dep: &crate::core::Dependency,
+        kind: super::source::QueryKind,
+        f: &mut dyn FnMut(super::IndexSummary),
+    ) -> std::task::Poll<crate::CargoResult<()>> {
+        let local_source = self.local.source_id();
+        let remote_source = self.remote.source_id();
+
+        let local_dep = dep.clone().map_source(remote_source, local_source);
+        let mut local_packages = std::collections::HashSet::new();
+        let mut local_callback = |index: IndexSummary| {
+            let index = index.map_summary(|s| s.map_source(local_source, remote_source));
+            local_packages.insert(index.as_summary().clone());
+            f(index)
+        };
+        ready!(self.local.query(&local_dep, kind, &mut local_callback))?;
+
+        let mut package_collision = None;
+        let mut remote_callback = |index: IndexSummary| {
+            if local_packages.contains(index.as_summary()) {
+                package_collision = Some(index.as_summary().clone());
+            }
+            f(index)
+        };
+        ready!(self.remote.query(dep, kind, &mut remote_callback))?;
+
+        if let Some(collision) = package_collision {
+            std::task::Poll::Ready(Err(anyhow::anyhow!(
+                "found a package in the remote registry and the local overlay: {}@{}",
+                collision.name(),
+                collision.version()
+            )))
+        } else {
+            std::task::Poll::Ready(Ok(()))
+        }
+    }
+
+    fn invalidate_cache(&mut self) {
+        self.local.invalidate_cache();
+        self.remote.invalidate_cache();
+    }
+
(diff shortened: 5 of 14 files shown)


crates/cargo-test-support/src/lib.rs
```diff
@@ -844,6 +844,17 @@ impl Execs {
         self
     }
 
+    pub fn overlay_registry(&mut self, url: &Url, path: &str) -> &mut Self {
+        if let Some(ref mut p) = self.process_builder {
+            let env_value = format!("{}={}", url, path);
+            p.env(
+                "__CARGO_TEST_PACKAGE_CONFUSION_VULNERABILITY_DO_NOT_USE_THIS",
```
Eh2406, line 851: Nit: The term of art for this kind of vulnerability is "Dependency Confusion". As with so many names in computer science, it's better to have consistency then trying to find a better name. This term was coined in the blog post https://medium.com/@alex.birsan/dependency-confusion-4a5d60fec610 which might be worth linking in comments about how users could easily miss use this.

src/cargo/sources/overlay.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+impl<'gctx> Source for OverlayConfusionAttack<'gctx> {
+    fn source_id(&self) -> crate::core::SourceId {
+        self.remote.source_id()
+    }
+
+    fn supports_checksums(&self) -> bool {
+        true
```
Eh2406, line 32: I would've guessed that this depends on whether both of the sources being overlaid support checksum. Did I miss something?

src/cargo/sources/overlay.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    fn requires_precise(&self) -> bool {
+        false
```
Eh2406, line 36: I would've thought this was true if either of the sources required precise, what am I missing?

src/cargo/sources/overlay.rs
```diff
@@ -0,0 +1,119 @@
...
+        let local_source = self.local.source_id();
+        let remote_source = self.remote.source_id();
+
+        let local_dep = dep.clone().map_source(remote_source, local_source);
+        let mut callback = |index: IndexSummary| {
+            let index = index.map_summary(|s| s.map_source(local_source, remote_source));
+            f(index)
+        };
+        ready!(self.local.query(&local_dep, kind, &mut callback))?;
+        self.remote.query(dep, kind, f)
```
Eh2406, line 54: Does this need to handle the same package/version being available from both registries? Possibly given its limited use case, just by detecting the situation and crashing.

Eh2406:
Catching Eric up on some of the context, Arlo and I discussed the questions from https://github.com/rust-lang/cargo/issues/10948#issuecomment-2112683333 at office hours on Thursday. Adding this API, exclusively for internal use, seemed like the least hacky solution.

src/cargo/sources/overlay.rs
jneem: Good point, I've added that check and a test.

src/cargo/sources/overlay.rs
jneem: Good point, I'd had in my head (and in the documentation) that the two sources were both `RegistrySource`s. I've updated the docs and this method to reflect that they could be arbitrary `Source`s.

src/cargo/sources/overlay.rs
jneem: (As above)

src/cargo/sources/overlay.rs
```diff
@@ -0,0 +1,141 @@
+use std::task::ready;
+
+use crate::sources::IndexSummary;
+
+use super::source::{MaybePackage, Source};
+
+/// A `Source` that overlays a one source over another, pretending that the packages
```
torhovland, line 7:
```suggestion
/// A `Source` that overlays one source over another, pretending that the packages
```

src/cargo/util/context/mod.rs
```diff
@@ -523,6 +526,13 @@ impl GlobalContext {
             .borrow_mut()
     }
 
+    /// Local overlay configuration.
+    ///
+    /// See [`crate::sources::overlay::OverlayConfusionAttack`] for why you shouldn't use this.
+    pub fn local_overlays(&self) -> RefMut<'_, HashMap<SourceId, PathBuf>> {
```
torhovland, line 532: Should probably mention that the paths need to point to local registries, not just to any crate file or directory.

src/cargo/sources/overlay.rs
```diff
@@ -0,0 +1,141 @@
...
+/// available in the overlay are actually available in the other one.
+///
+/// This is a massive footgun and a terrible idea, so we do not (and never will)
+/// expose this publicly. However, it is useful for some very specific private
+/// things, like locally verifying a bunch of packages at a time before any of
+/// them have been published.
+pub struct OverlayConfusionAttack<'gctx> {
```
torhovland, line 14: Frankly, I think `OverlaySource` is a better fitting name 🙂

Cargo.toml
```diff
@@ -60,7 +60,7 @@ indexmap = "2.2.6"
 itertools = "0.12.1"
 jobserver = "0.1.31"
 lazycell = "1.3.0"
-libc = "0.2.154"
+libc = "0.2.153"
```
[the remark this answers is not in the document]
jneem: 0.2.154 got [yanked](https://github.com/rust-lang/libc/issues/3677#issuecomment-2093991553). I see that 0.2.155 just got released, so I'll switch to that

src/cargo/sources/overlay.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl<'gctx> OverlayConfusionAttack<'gctx> {
+    pub fn new(local: Box<dyn Source + 'gctx>, remote: Box<dyn Source + 'gctx>) -> Self {
+        Self { local, remote }
+    }
+}
+
+impl<'gctx> Source for OverlayConfusionAttack<'gctx> {
+    fn source_id(&self) -> crate::core::SourceId {
+        self.remote.source_id()
```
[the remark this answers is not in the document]
jneem: In the current implementation that can't happen because the overlays are configured by a `HashMap<SourceId, _>`, so only one overlay can be registered for each remote.

src/cargo/sources/overlay.rs
jneem: That name definitely makes more sense, yes. But in office hours someone suggested giving it a scary-looking name to make people less likely to misuse it...

tests/testsuite/registry_overlay.rs
```diff
@@ -0,0 +1,161 @@
...
+    let p = simple_project();
+
+    Package::new("bar", "0.0.1").publish();
+    // The one we want is in the main registry.
+    Package::new("baz", "0.1.1").publish();
+    // An older version in the local registry doesn't confuse it.
+    Package::new("baz", "0.0.1")
+        .alternative(true)
+        .local(true)
+        .publish();
```
epage, line 68: I don't think this is really testing anything because `baz` is too low of a version for the version requirement

tests/testsuite/registry_overlay.rs
```diff
@@ -0,0 +1,161 @@
...
+        .overlay_registry(&reg.index_url(), &alt_path)
+        .run();
+}
+
+#[cargo_test]
+fn version_fallback() {
+    let (reg, alt_path) = setup();
+    let p = simple_project();
+
+    Package::new("bar", "0.0.1").publish();
```
epage, line 61: What is the role of `bar` in these tests?   We should either simplify the test cases or name the packages to convey the role they play in the tests

tests/testsuite/registry_overlay.rs
```diff
@@ -0,0 +1,161 @@
...
+fn multiversion() {
+    let (reg, alt_path) = setup();
+    let p = simple_project();
+
+    Package::new("bar", "0.0.1").publish();
+    Package::new("baz", "0.1.1").dep("baz", "=0.0.1").publish();
+    Package::new("baz", "0.0.1")
+        .alternative(true)
+        .local(true)
+        .publish();
```
epage, line 115: Should we have the local / selected version depend on the registry version?  Thats more like what we'll be dealing with for `cargo package`

tests/testsuite/registry_overlay.rs
```diff
@@ -0,0 +1,161 @@
...
+            r#"
+                [package]
+                name = "foo"
+                version = "0.0.1"
+                edition = "2015"
+                authors = []
+
+                [dependencies]
+                bar = ">= 0.0.0"
+                baz = ">= 0.1.0"
```
epage, line 32: Why are non-standard version reqs being used?  Chekhov's gun would say that there is something specific about this to the tests but I'm not seeing it.

(ditto for the other package)

src/cargo/util/context/mod.rs
```diff
@@ -523,6 +526,16 @@ impl GlobalContext {
             .borrow_mut()
     }
 
+    /// Local overlay configuration.
+    ///
+    /// The values in this map are paths pointing to a local registry. Inserted overlays will
+    /// affect any newly created `PackageRegistry`s.
+    ///
+    /// See [`crate::sources::overlay::OverlayConfusionAttack`] for why you shouldn't use this.
+    pub fn local_overlays(&self) -> RefMut<'_, HashMap<SourceId, PathBuf>> {
```
epage, line 535: I'd hope we can avoid putting this on GlobalContext but its hard to say without reviewing the follow up at the same time.  Should we defer this to the follow up PR?

My motivation is that it gets messy adding and removing from the Global Context and if we can directly instantiate things, the easier it will be to maintain.

src/cargo/sources/overlay.rs
epage: I like `OverlaySource` being the suffix.  Maybe `DependencyConfusionThreatOverlaySource`?

src/cargo/sources/config.rs
```diff
@@ -136,7 +170,8 @@ impl<'gctx> SourceConfigMap<'gctx> {
         debug!("loading: {}", id);
 
         let Some(mut name) = self.id2name.get(&id) else {
-            return id.load(self.gctx, yanked_whitelist);
+            let src = id.load(self.gctx, yanked_whitelist)?;
+            return self.overlay_source(id, src, yanked_whitelist);
```
epage, line 174: We only call `overlay_source` on the result of a `load`

Should we name it `fn load` and perform the load and apply the overlay?  I worry that the current structure could make it easy to overlook applying the overlay while this pattern would make it the default to go through `Self::load` and we'd be less likely to introduce bugs to this

src/cargo/sources/config.rs
```diff
@@ -81,6 +84,22 @@ impl<'gctx> SourceConfigMap<'gctx> {
...
+
+        if let Ok(overlay) =
+            gctx.get_env("__CARGO_TEST_DEPENDENCY_CONFUSION_VULNERABILITY_DO_NOT_USE_THIS")
+        {
+            let (url, path) = overlay.split_once('=').ok_or(anyhow::anyhow!(
+                "invalid overlay format. I won't tell you why; you shouldn't be using it anyway"
+            ))?;
+            let overlay_id = SourceId::for_local_registry(path.as_ref())?;
+            base.overlays.insert(SourceId::from_url(url)?, overlay_id);
+        }
```
epage, line 101: Should we move this hack up into the caller?  `PackageRegistry` is the immediate caller but maybe a layer above that?  That would help with the `GlobalContext`.  The main downside is it would make it more annoying to support `__CARGO_TEST_DEPENDENCY_CONFUSION_VULNERABILITY_DO_NOT_USE_THIS`

Thoughts?

src/cargo/util/context/mod.rs
jneem: I can give a little more context before we decide whether or not to defer. `cargo package` needs the overlay to take effect during a call to `ops::compile`. AFAICT there are three possibilities:

- put the overlay configuration in `GlobalContext`
- put the overlay configuration in the ephemeral workspace that `cargo package` creates
- have `cargo package` do something lower-level than `ops::compile`

The workspace option has the advantage that the overlay configuration doesn't need to be removed (because the workspace is discarded). It seems like it would just require a little work in `resolve_ws` and `resolve_ws_with_opts` to get the overlay configuration out of the workspace and insert it into the `PackageRegistry`

tests/testsuite/registry_overlay.rs
jneem: I think I just copied it from some other test. I'll switch to normal version reqs.

tests/testsuite/registry_overlay.rs
jneem: I think it's not needed; I'll remove it.

tests/testsuite/registry_overlay.rs
jneem: The motivation here was to test that the overlay is correctly the union of the versions in the overlay and the overlaid registries. (If it were only returning the overlay versions, it would only return the too-old version and fail to resolve)

tests/testsuite/registry_overlay.rs
jneem: I don't think so, but also happy to discuss it further. My understanding was that the behavior of `cargo package` w.r.t. intra-workspace deps should be exactly the same as if the deps were published before packaging. Resolution should have the option to select the versions that are being packaged simultaneously, but it's also allowed not to.

src/cargo/sources/config.rs
jneem: Unfortunately we're already in a `fn load`. I renamed `overlay_source` to `load_and_overlay`, and modified it to do the load and the overlay. I think this is an improvement but not sure it it really addresses the concern.

src/cargo/sources/config.rs
jneem: I think if we moved the overlay configuration into `Workspace` as discussed below then this could be pushed up into `resolve_ws` and `resolve_ws_with_opts`

src/cargo/util/context/mod.rs
epage: The workspace seems the most beneficial.  Whichever way we go, still seems like it could be split out into a follow up PR

tests/testsuite/registry_overlay.rs
epage: Yes, that should be the behavior but we aren't testing that situation to confirm it.  `cargo package` will involve a local package depending on an overlayed package which may depend on a registry package.

I guess there are also situations where you depend on a registry package which will depend on the overlaid package (if `clap_complete` used `criterion` it would both depend on clap directly and through `criterion`).  So maybe its best to cover both cases.

tests/testsuite/registry_overlay.rs
```diff
@@ -0,0 +1,156 @@
...
+        alt.index_url()
+            .to_file_path()
+            .unwrap()
+            .into_os_string()
+            .into_string()
+            .unwrap(),
+    )
+}
+
+fn simple_project() -> Project {
```
epage, line 19: Let's avoid using this.  This reduces "duplication" but that is false code duplication.  They have the appearance of being the same but they do not share requirements.  Instead it affects the code locality (the ability to read tightly coupled code).  I'm trying to double check what these tests are doing and jumping back and forth is making this more difficult.

tests/testsuite/registry_overlay.rs
epage: Except we aren't checking the resolve error for whether it sees `baz@0.0.1`

We also should probably have a test the other direction, that the overlay source properly shadows the regular source.

tests/testsuite/registry_overlay.rs
jneem: Ah, makes sense, I just misunderstood your original comment. I'm now testing both directions.

> I guess there are also situations where you depend on a registry package which will depend on the overlaid package
I think the registry package will never *require* an overlaid package, right? Because the registry package can satisfy all its dependencies from the registry. I added a test where the registry package will pick up an overlaid dependency because it's newer (but compatible with) the registry version of that dependency.

tests/testsuite/registry_overlay.rs
jneem: Got it. I'm now testing both directions, and checking the appropriate versions in the stderr.

Cargo.toml
```diff
@@ -60,7 +60,7 @@ indexmap = "2.2.6"
 itertools = "0.12.1"
 jobserver = "0.1.31"
 lazycell = "1.3.0"
-libc = "0.2.154"
+libc = "0.2.155"
```
epage, line 63: Why is libc being bumped as part of this PR?

tests/testsuite/registry_overlay.rs
epage: Not require but should resolve to use it.

src/cargo/sources/config.rs
```diff
@@ -125,6 +145,22 @@ impl<'gctx> SourceConfigMap<'gctx> {
         self.gctx
     }
 
+    fn load_and_overlay(
```
epage, line 148: nit: I feel like `and` in a name is a code smell, that you are making a fucntion do unrelated things.  Ideally, we find a name that gets at the heart of what distinguishes this from `load`

epage, line 148: nit: I tend to prefer putting internal functions after their use.  People tend to read top-down and tend to be looking at interfaces and then look into implementations where relevant.

Cargo.toml
jneem: I've removed it. (It had something to do with 0.2.154 being yanked, but I don't remember anymore why that was a problem)

src/cargo/sources/config.rs
jneem: What about `load_unreplaced`? Because the difference between `load` and `load_unreplaced` is that `load` does source replacement.

jneem: Done

epage: Describing something by what it isn't doing is a bit messy.

Your name hints that `pub fn load` is `load_replaced`.  So maybe this could be `load_overlaid`.  Still not thrilled but probably the best we can do without blocking this further.

src/cargo/sources/config.rs
epage: Just wanted to check in if there is anything you are waiting on me for this change as you've done other work on this PR but not this part yet.

epage:
btw feel free to clean up your commit history for how you want to present it for review and merge.

src/cargo/sources/config.rs
jneem: Ah sorry, I thought this would go in a follow-up (along with moving the overlay configuration to `Workspace`). On Monday I'll clean up the commit history and look into doing this too.

src/cargo/ops/resolve.rs
```diff
@@ -116,7 +116,8 @@ version. This may also occur with an optional dependency that is not enabled.";
 /// This is a simple interface used by commands like `clean`, `fetch`, and
 /// `package`, which don't specify any options or features.
 pub fn resolve_ws<'a>(ws: &Workspace<'a>, dry_run: bool) -> CargoResult<(PackageSet<'a>, Resolve)> {
-    let mut registry = PackageRegistry::new(ws.gctx())?;
+    let source_config = SourceConfigMap::new_with_overlays(ws.gctx(), ws.local_overlays()?)?;
+    let mut registry = PackageRegistry::new_with_source_config(ws.gctx(), source_config)?;
```
epage, line 120: How do we decide when to replace `PackageRegistry::new` with this?

We only replaced two uses of it:
```
src/cargo/ops/resolve.rs
119:    let mut registry = PackageRegistry::new(ws.gctx())?;
145:    let mut registry = PackageRegistry::new(ws.gctx())?;

src/cargo/ops/cargo_package.rs
475:    let mut tmp_reg = PackageRegistry::new(ws.gctx())?;

src/cargo/ops/cargo_update.rs
35:    let mut registry = PackageRegistry::new(ws.gctx())?;
76:                    let mut registry = PackageRegistry::new(opts.gctx)?;
91:    let mut registry = PackageRegistry::new(opts.gctx)?;
229:    let mut registry = PackageRegistry::new(gctx)?;

src/cargo/ops/cargo_add/mod.rs
81:    let mut registry = PackageRegistry::new(options.gctx)?;

crates/xtask-bump-check/src/xtask.rs
393:    let mut registry = PackageRegistry::new(gctx)?;
```

epage, line 120: Depending on which type is more correct to know about the other, should we have either
- `Workspace::package_registry(&self) -> PackageRegistry`
- `PackageRegistry::with_workspace(ws: &Workspace) -> Self`

to simplify this pattern and make it more discoverable so that when new features get added, people are more likely to use this and not break support for this in `cargo package` in odd corner cases?

crates/cargo-test-support/src/registry.rs
```diff
@@ -1649,7 +1649,12 @@ impl Package {
     /// Returns the path to the compressed package file.
     pub fn archive_dst(&self) -> PathBuf {
         if self.local {
-            registry_path().join(format!("{}-{}.crate", self.name, self.vers))
+            let path = if self.alternative {
```
epage, line 1652: Please move this to the appropriate commit

src/cargo/ops/resolve.rs
jneem: I don't have a great sense for which is better, but since `Workspace` already mentions `PackageRegistry` (and `PackageRegistry` doesn't mention `Workspace`) I went with `Workspace::package_registry`. All the usages of `PackageRegistry::new` had a workspace available, so I converted them all and then deleted `PackageRegistry::new` so that future changes are likely to default to `Workspace::package_registry`

epage:
Thanks!

@bors r+

PR #13929, opened 2024-05-18
Silence warnings running embedded unittests.

ehuss:
The `util::toml::embedded::test_expand` unittests were printing warnings directly to stderr, circumventing the libtest override. They were displaying ``warning: `package.edition` is unspecified, defaulting to `2021` ``. That warning doesn't look particularly important for those tests, so this changes it to swallow the output.

We may want to have some kind of utility function for generating a GlobalContext for testing purposes that does this, since this is a repeated pattern.

### DIFF
--- a/src/cargo/util/toml/embedded.rs
+++ b/src/cargo/util/toml/embedded.rs
@@ -265,12 +265,12 @@ mod test_expand {
 
     macro_rules! si {
         ($i:expr) => {{
-            expand_manifest(
-                $i,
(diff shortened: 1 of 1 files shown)


epage:
Hopefully we'll be able to improve capturing to include `stdout` / `stderr`, not just `print`/`eprint`, see https://github.com/rust-lang/testing-devex-team/issues/8

weihanglo (approved):

weihanglo:
Make sense. I've seen that as well.

@bors r+

PR #13931, opened 2024-05-18
Temporarily fix standard_lib tests on linux.

ehuss:
This fixes the standard_lib tests which are broken in the latest nightly. The latest nightly now requires rust-lld to be in the sysroot for x86_64-unknown-linux-gnu. This broke these tests which were trying to verify that the standard library is not required. This temporarily removes this validation, but we should have some way of enforcing it (https://github.com/rust-lang/wg-cargo-std-aware/issues/31).

cc https://github.com/rust-lang/rust/issues/125246


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #13934, opened 2024-05-19
Make `level` field optional in the `[lints]` table

Urgau:

This PR makes the `level` field in the individual lint under the `[lints]` optional if some other key (except `priority` since it would make sense to have it alone) is defined. The goal is to avoid a empty lint definition.

This is to allow special configs like `check-cfg` in the `unexpected_cfgs` to omit the `level` field.

Users would now be able to omit the `level` field, resulting in:
```diff
- unexpected_cfgs = { level = "warn", check-cfg = ['cfg(has_foo)'] }
+ unexpected_cfgs = { check-cfg = ['cfg(has_foo)'] }
```

Test putting removing the `level` field with/without this PR.

Reviewing this PR should be straightforward, look at the tests that change to see the effect of this change.

### Additional information

https://github.com/rust-lang/cargo/pull/13913#issuecomment-2115949246

r? @epage

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -2292,7 +2292,21 @@ supported tools: {}",
                     }
                 }
             }
-            if config.level().is_none() && config.config().map(|c| c.is_empty()).unwrap_or(true) {
+            if matches!(
+                config,
+                manifest::TomlLint::Config(manifest::TomlLintConfig {
+                    priority: Some(_),
+                    level: None,
+                    ..
+                })
+            ) || matches!(
+                config,
+                manifest::TomlLint::Config(manifest::TomlLintConfig {
+                    priority: None,
+                    level: None,
+                    config,
+                }) if config.is_empty())
+            {
                 anyhow::bail!("`lints.{tool}.{name}` missing field `level`");
             }
         }
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -2292,6 +2292,9 @@ supported tools: {}",
                     }
                 }
             }
+            if config.level().is_none() && config.config().map(|c| c.is_empty()).unwrap_or(true) {
+                anyhow::bail!("`lints.{tool}.{name}` missing field `level`");
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


tests/testsuite/lints_table.rs
```diff
@@ -856,3 +856,51 @@ warning: `im_a_teapot` is specified
...
+                [lints.rust]
+                unsafe_code = { }
+            "#,
+        )
+        .file("src/main.rs", "fn main() {}")
+        .build();
+
+    p.cargo("check")
+        .with_status(101)
+        .with_stderr_contains("[..]missing field `level`[..]")
```
epage, line 880: At least for myself, I prefer to not use the `contains` methods because they are brittle.  A slight change and you can be testing for the wrong condition.
(this is especially true for `does_not_contain`)

src/cargo/util/toml/mod.rs
```diff
@@ -2292,6 +2292,9 @@ supported tools: {}",
                     }
                 }
             }
+            if config.level().is_none() && config.config().map(|c| c.is_empty()).unwrap_or(true) {
+                anyhow::bail!("`lints.{tool}.{name}` missing field `level`");
+            }
```
epage, line 2297: This is the big question for me: what should be allows

Today
```toml
lint_name = {}  # disallowed
lint_name = { config_name = "config_value" }  # disallowed
lint_name = { priority = 1 }  # disallowed
lint_name = { priority = 1, config_name = "config_value" }  # disallowed
lint_name = { level = "warn" }  # allowed
lint_name = { level = "warn", priority = 1 }  # allowed
```

With this PR
```toml
lint_name = {}  # disallowed
lint_name = { config_name = "config_value" }  # allowed
lint_name = { priority = 1 }  # disallowed
lint_name = { priority = 1, config_name = "config_value" }  # allowed
lint_name = { level = "warn" }  # allowed
lint_name = { level = "warn", priority = 1 }  # allowed
```

This allows `priority` without `level`.  Should we allow that?

This requires `level` or config, should we bother with an empty table error?

I almost lean towards either keeping the requirement on `level` or dropping all requirements (though `priority` alone feels weird).

Thoughts?

Urgau: > This allows `priority` without `level`. Should we allow that?

I don't think so, if `priority` is set then `level` must also be set, that's a oversight on my part. Will fix it.

> This requires `level` or `config`, should we bother with an empty table error?

My stance is that it is useless and weird to have empty table.

tests/testsuite/lints_table.rs
Urgau: Switch to `with_stderr`.

src/cargo/util/toml/mod.rs
Urgau: Fixed the `priority` field alone without the `level` field. It now bail out.

epage:
We talked about this in today's Cargo team meeting and decided to defer this out.
- We can always merge this later but we cannot undo it.  We should probably learn from the current behavior works before iterating on it further.
- There was some preference on the team that we should be explicit on the level when config is present.

PR #13935, opened 2024-05-19
Make `git::use_the_cli` test truly locale independent

Urgau:

The current `git::use_the_cli` test part the output of Git but that output is locale dependant, making it fail on my non-english system.

Specifically this part of the test is local-dependent:
```
From [..]
 * [new ref] [..] -> origin/HEAD[..]
```

Adding the `LC_ALL=C` env to the cargo invocation solve the issue, making the test locale independent.

Trying putting a different `LC_ALL` locale and the test will fail without this PR and will not fail with my PR.

### DIFF
--- a/tests/testsuite/git.rs
+++ b/tests/testsuite/git.rs
@@ -2831,7 +2831,11 @@ From [..]
 [FINISHED] [..]
 ";
 
-    project.cargo("check -v").with_stderr(stderr).run();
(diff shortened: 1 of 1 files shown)


epage:
@bors r+

Main question I had is whether we should apply this to all tests but figured we can iterate more in the future.

PR #13937, opened 2024-05-20
docs(ref): Simplify check-cfg build.rs docs

epage:

After looking over the rust documentation updates, decided to look over Cargo's build.rs documentation.

- Removed link to the blog post as that is not evergreen documentation
- Collapsed down the example to put more emphasis on the more complete example

### DIFF
--- a/src/doc/src/reference/build-scripts.md
+++ b/src/doc/src/reference/build-scripts.md
@@ -258,8 +258,6 @@ identifier, the value should be a string.
 
 ### `cargo::rustc-check-cfg=CHECK_CFG` {#rustc-check-cfg}
 
-*See the announcement [blog post][check-cfg-blog-post] for more a global view of the feature.*
-
 Add to the list of expected config names and values that is used when checking
 the _reachable_ cfg expressions.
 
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
Thanks for the simplification.

src/doc/src/reference/build-scripts.md
weihanglo, line 1: Should we have a link to the new page in rustc book?

https://github.com/rust-lang/rust/pull/125219

epage: Sounds like that page is a short-term hack

> However, I'd rather keep Cargo-specific explainers that are as long as this one out of the rustc book

(which I agree with)

so I'm a bit hesitant linking to a page that will soon be broken.

weihanglo: Yes, but docs are designed for people to read. They are less useful if there is no route to them. We also haven't put a date for removing that, so I think it's better to link to that within the same version. The linkchcker should be able to detect breakages when the page is removed.

epage: Looking more closely at each piece of documentation, I'm not seeing much reason to link to the cargo-specific page.  It just duplicates what we are already trying to cover.  Instead, we are missing a link to the parent page which I added.

epage: (the cargo page is more of a starting point from the lint output)

PR #13939, opened 2024-05-20
Use `i32` rather than `usize` as "default integer" in library template

tbu-:
`usize` was renamed from `uint` in order to convey that it's not the "default integer type". Guide new users to use `i32` instead of `usize`.

### DIFF
--- a/src/cargo/ops/cargo_new.rs
+++ b/src/cargo/ops/cargo_new.rs
@@ -876,7 +876,7 @@ fn main() {
 "
         } else {
             b"\
-pub fn add(left: usize, right: usize) -> usize {
+pub fn add(left: u64, right: u64) -> u64 {
     left + right
 }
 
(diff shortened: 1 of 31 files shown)


weihanglo (commented):
This looks reasonable. I am more curious on how you found this and why you think knowing the default integer type is important for new Rust users. Mind sharing a bit?

tbu-:
> I am more curious on how you found this and why you think knowing the default integer type is important for new Rust users.

I found this by creating a new library with Cargo and seeing that the example code was unnecessarily using `usize`.

I wouldn't exactly say it's important to new Rust users, but I wouldn't set a bad example by using `usize` when a more concrete integer type would be reasonable.

tbu-:
> I think `u64` might be better from a teaching standpoint.

I don't particularly care about the exact integer type used here, `i32`, `i64`, `u32`, `u64` would all work here, in my opinion. Does anyone else have an opinion on this or should I just change it to `u64`?

tbu-:
> I think `u64` might be better from a teaching standpoint.

Changed to `u64`.

weihanglo (approved):

weihanglo:
The template `cargo new` creates doesn't really have stability guarantee so I am happy to just merge this. If there turns out to be an issue, please call it out and we can consider reverting.

@bors r+

PR #10807, opened 2022-06-30
Apply GitHub fast path even for partial hashes

dtolnay:
### What does this PR try to resolve?

As flagged in https://github.com/rust-lang/cargo/pull/10079#issuecomment-1170940132, it's not great to assume that git SHAs would always be 40-character hex strings. In the future they will be longer.

> Git is on a long-term trajectory to move to SHA256 hashes ([current status](https://lwn.net/SubscriberLink/898522/f267d0e9b4fe9983/)). I suppose when that becomes available/the default it's possible for a 40-digit hex-encoded hash not to be the full hash. Will this fail for that case?

The implementation from #10079 fails in that situation because it turns dependencies of the form `{ git = "…", rev = "[…40 hex…]" }` into fetches with a refspec `+[…40 hex…]:refs/commit/[…40 hex…]`. That only works if the 40 hex digits are the *full* long hash of your commit. If it's really a prefix ("short hash") of a 64-hex-digit SHA-256 commit hash, you'd get a failure that resembles:

```console
error: failed to get `dependency` as a dependency of package `repro v0.0.0`

Caused by:
  failed to load source for dependency `dependency`

Caused by:
  Unable to update https://github.com/rust-lang/dependency?rev=b30694b4d9b29141298870b7993e9aee10940524

Caused by:
  revspec 'b30694b4d9b29141298870b7993e9aee10940524' not found; class=Reference (4); code=NotFound (-3)
```

This PR updates the implementation so that Cargo will curl GitHub to get a resolved long commit hash *even if* the `rev` specified for the git dependency in Cargo.toml already looks like a SHA-1 long hash.

### Performance considerations

⛔ This reverses a (questionable, negligible) benefit of #10079 of skipping the curl when `rev` is a long hash and is not already present in the local clone. These curls take 200-250ms on my machine.

🟰 We retain the much larger benefit of #10079 which comes from being able to precisely fetch a single `rev`, instead of fetching all branches and tags in the upstream repo and hoping to find the rev somewhere in there. This accounts for the entire performance difference explained in the summary of that PR.

🟰 We still skip the curl when `rev` is a **long hash** of a commit that is already previously fetched.

[...]

PR #13946, opened 2024-05-22
fetch specific commits even if the github fast path fails

### LINKED ISSUE #13555: `object not found` when fetching git dependency

iliana:

This PR fixes #13555, which describes a regression from 1.64.0 to 1.65.0 where the inability to fetch commit information from api.github.com (the "GitHub fast path") silently changes Cargo's behavior.

Cargo can fetch a specific Git commit from a remote without having to fetch all refs. Prior to #10807, this functionality required a repository hosted on github.com and providing the full commit hash (usually available from the Cargo.lock); after that change, any revision (including abbreviated revisions) that could be resolved by GitHub's API could be fetched directly. However, this logic requires the "GitHub fast path", which was not intended to be robust, to successfully return the resolved commit hash; if a client is currently rate-limited by api.github.com (very common in CI and shared cloud / corporate environments) this fails and Cargo falls back to fetching all refs.

Usually this is not noticeable. However, GitHub allows fetching commits that are related to the repository but not actually part of any of its refs, including commits pushed to a fork. This results in the same command working fine in some environments where api.github.com is accessible, and not working in other environments that are rate-limited, which is very confusing and difficult to debug.

This change adds another branch to cover the regression case: if we are going through the GitHub fast path with a full commit hash, return early indicating that we need to fetch it. (Previously: ~~when the GitHub fast path was unsuccessful, the user is not using the unstable shallow clone options, and we have a full commit hash and expect to be able to fetch it directly because we know it's a github.com repository.~~)

I have been testing this PR by temporarily adding a `0.0.0.0 api.github.com` entry to my `/etc/hosts`, which causes the GitHub fast path to always fail, then running:

```
target/release/cargo install --git https://github.com/haha-business/unstable-test-repo.git --rev c9040898c9183ddbb9402dcbf749ed06d6ea90ad
```

This refers to [a particular commit on a fork of the repo](https://github.com/iliana/unstable-test-repo/commit/c9040898c9183ddbb9402dcbf749ed06d6ea90ad) which won't be found by the fallback path or current Cargo.

**Note** that you will need to delete `~/.cargo/git/checkouts/unstable-test-repo-*` and `~/.cargo/git/db/unstable-test-repo-*` after a successful run with this change in order to reproduce the broken behavior of the current release.

I am having trouble getting the test suite to run at all on my system so I haven't experimented with writing a specific test for this case, but I probably should.

### Additional information

This uses the same logic as the unstable shallow clone support to detect if the revision is a full commit hash. This is not compatible with SHA-256 commit hashes; `git2::Oid` specifically expects a 40-character hexadecimal string. Given that the change introducing this bug was meant to future-proof SHA-256 support (despite only doing so for GitHub repositories), it might be good to make the logic more explicit within Cargo and allow either 40- or 64-character hex strings.

I wanted to keep this change focused on the regression fix, but in testing, pretty much every Git repository I could think of (including non-forges, like git.kernel.org and some repositories I host on my own infrastructure with cgit) supports fetching directly from a commit, so it would be ideal to eventually relax the GitHub requirement for this functionality. However, it would need some sort of fallback logic because I suspect the HTTP [dumb protocol](https://git-scm.com/book/en/v2/Git-Internals-Transfer-Protocols) doesn't support commit references, and I haven't researched when this functionality was added to the smart protocol.

### DIFF
--- a/src/cargo/sources/git/utils.rs
+++ b/src/cargo/sources/git/utils.rs
@@ -1417,6 +1417,14 @@ fn github_fast_path(
                         return Ok(FastPathRev::UpToDate);
                     }
                 }
+                // If `rev` is a full commit hash, the only thing it can resolve
+                // to is itself. Don't bother talking to GitHub in that case
+                // either. (This ensures that we always attempt to fetch the
+                // commit directly even if we can't reach the GitHub API.)
+                if let Ok(oid) = rev.parse() {
+                    debug!("github fast path is already a full commit hash {rev}");
+                    return Ok(FastPathRev::NeedsFetch(oid));
+                }
                 rev
             } else {
                 debug!("can't use github fast path with `rev = \"{}\"`", rev);


src/cargo/sources/git/utils.rs
```diff
@@ -977,6 +977,13 @@ pub fn fetch(
                 // The reason we write to `refs/remotes/origin/HEAD` is that it's of special significance
                 // when during `GitReference::resolve()`, but otherwise it shouldn't matter.
                 refspecs.push(format!("+{0}:refs/remotes/origin/HEAD", rev));
+            } else if Url::parse(remote_url).map_or(false, |url| is_github(&url))
```
weihanglo, line 980: Thank you for the nice finding and write-up!

I understand supporting other Git hosting would be awesome, though this is still a GitHub specific fast path. I wonder if we could move this patch to under [`github_fast_path`](https://github.com/rust-lang/cargo/blob/84dc5dc11a9007a08f27170454da6097265e510e/src/cargo/sources/git/utils.rs#L1420). Maybe something like this?

```diff
                if let Some(local_object) = local_object {
                    if is_short_hash_of(rev, local_object) {
                        debug!("github fast path already has {local_object}");
                        return Ok(FastPathRev::UpToDate);
                    }
                }
+               if let Ok(oid) = rev.parse::<Oid>() {
+                    debug!("github fast path is already a full commit hash {rev}");
+                    return Ok(FastPathRev::NeedsFetch(oid))
+               }
                rev
            } else {
                debug!("can't use github fast path with `rev = \"{}\"`", rev);
                return Ok(FastPathRev::Indeterminate);
            }
```

iliana: Oh that's clever! Let me give that a shot.

iliana:
I added a `Co-authored-by` header to my commit since you wrote it! Let me know if that should be adjusted.

It looks like that approach also has the benefit of making the behavior around what the ref is fetched as more consistent; previously if the GitHub API request failed the ref would instead be fetched to `refs/remotes/origin/HEAD` instead of `refs/commit/{commit_hash}`. Given that was never supposed to happen, and the ref we fetch as is pretty meaningless in the context of these Cargo-managed repos, I think that's fine.

weihanglo (approved):
Looks great! Thank you for the contribution.

Let's move on and leave off the issue of handling 64-hex-digit commit hash, believing this part won't be the only place need to change at that time 😬.

PR #13947, opened 2024-05-22
Package workspaces

### LINKED ISSUE #10948: cargo package --workspace is not very useful

jneem:
Adds support for packaging an entire workspace, even when there are dependencies between the crates. The generated packages should be identical to the ones produced by packaging and publishing the crates one-by-one in dependency order, but the main benefit of this PR is that the packages can be created and verified locally before anything is published.

The main mechanism is the one in #13926, where we create a temporary local registry that "overlays" the true registry. We "publish" the crates in the local registry, which enables lockfile generation and verification of the dependent crates.

This adds `--registry` and `--index` flags to `cargo package`. They act
much like the same arguments to `cargo publish`, except that of course
we are not actually publishing to the specified registry. Instead, these
arguments affect lock-file generation for intra-workspace dependencies:
when simultaneously packaging a crate and one of its dependencies, the
lock-file will be generated under the assumption that the dependency
will be published to the specified registry.

You can also publish a subset of a workspace using `-p` arguments. In this case, there will be an error unless the chosen subset contains all of the dependencies of everything in the subset.

Fixes #10948. Based on #13926.

### Compatibility issue

This PR introduces a case where `cargo package` will fail where it did not before: if you have a workspace containing two packages, `main` and `dep@0.1.0`, where `main` depends on `dep@0.1.0` and `dep@0.1.0` is already published in crates.io then attempting to package the whole workspace will fail. To be specific, it will package `dep@0.1.0` successfully and then fail when trying to package `main` because it will see the two different packages for `dep@0.1.0`. Note that `cargo publish` will already fail in this scenario.

This shouldn't interfere with crates.io running `cargo package` for each package to diff the `.crate` files
- This might interfere if someone tried to verify their "published" MSRV by running `cargo package`.

The failure could be avoided by changing the local overlay source to not error out if there's a duplicate package; see [here](https://github.com/rust-lang/cargo/pull/13926#discussion_r1606822615). However, failing early has the advantage of catching errors early.


jneem:
I think this is ready for a look.

There's a fair amount of churn in the tests: for workspace packaging we need to finish all the packaging before we start verifying, and this changes the order of console messages. I pushed as much of the test churn as I could into the first commit.

src/cargo/ops/cargo_package.rs
```diff
@@ -231,9 +212,33 @@ pub fn package(ws: &Workspace<'_>, opts: &PackageOpts<'_>) -> CargoResult<Option
...
+    let src_files = src.list_files(pkg)?;
+
+    // Check (git) repository state, getting the current commit hash if not
+    // dirty.
+    let vcs_info = if !opts.allow_dirty {
+        // This will error if a dirty repo is found.
+        check_repo_state(pkg, &src_files, gctx)?
+    } else {
+        None
+    };
```
epage, line 378: metadata seems unrelated to this step

I feel like repo state is similar. If this function read the files also, then I could see that.

src/cargo/ops/cargo_package.rs
```diff
@@ -82,48 +83,33 @@ struct GitVcsInfo {
...
+// Builds a tarball and places it in the output directory.
+fn create_package(
+    ws: &Workspace<'_>,
+    pkg: &Package,
+    ar_files: Vec<ArchiveFile>,
+) -> CargoResult<FileLock> {
+    let gctx = ws.gctx();
     let filecount = ar_files.len();
 
-    if opts.list {
```
epage, line 119: imo `package-one` should still do this.  `cargo publish` might not need it but from an API perspective.

Alternatively, we assert if its set

src/cargo/ops/cargo_package.rs
```diff
@@ -145,10 +131,6 @@ pub fn package_one(
     dst.file().set_len(0)?;
     let uncompressed_size = tar(ws, pkg, ar_files, dst.file(), &filename)
         .with_context(|| "failed to prepare local package for uploading")?;
-    if opts.verify {
-        dst.seek(SeekFrom::Start(0))?;
-        run_verify(ws, pkg, &dst, opts).with_context(|| "failed to verify package tarball")?
-    }
 
     dst.seek(SeekFrom::Start(0))?;
```
epage, line 142: I assume the first `seek` is no longer needed because we are relying on the second seek with `run_verify` moved

tests/testsuite/package.rs
```diff
@@ -5038,3 +5038,303 @@ path = "examples/z.rs"
...
+
+For more information, try '--help'.
+",
+        )
+        .run();
+}
+
+/// level1 -> level2 checks that we handle local dependencies.
+/// level2 -> level3 checks that we handle dependencies with local dependencies.
+fn build_workspace_with_local_deps(registry: Option<&str>) -> Project {
```
epage, line 5342: Generally not too big of a fan of fixture setups like this
- The abstraction is leaky: in reading the test, you have to know all of the details of this to make sense of but but this has bad code locality
- This tends towards "god tests" that cover too many cases.  There are times when a test needs to hit a bunch of cases at once but when you have helpers like this, they tend to get over-reused and every test hits every case when only a few need to.  This causes slower tests that are likely to hit storage limits (a regular problem for cargo's tests on CI)

src/cargo/ops/cargo_package.rs
```diff
@@ -28,6 +28,7 @@ use tar::{Archive, Builder, EntryType, Header, HeaderMode};
 use tracing::debug;
 use unicase::Ascii as UncasedAscii;
```
epage, line 33: As for commit titles, strictly speaking this is not a refactor as it has user-visible changes

src/cargo/ops/registry/publish.rs
```diff
@@ -319,14 +319,12 @@ fn verify_dependencies(
...
-fn transmit(
+pub(crate) fn prepare_transmit(
     gctx: &GlobalContext,
     pkg: &Package,
-    tarball: &File,
-    registry: &mut Registry,
     registry_id: SourceId,
-    dry_run: bool,
-) -> CargoResult<()> {
+    require_readme_and_license: bool,
```
epage, line 327: Why is this `bool` in here?

src/bin/cargo/commands/package.rs
```diff
@@ -5,6 +5,8 @@ use cargo::ops::{self, PackageOpts};
 pub fn cli() -> Command {
     subcommand("package")
         .about("Assemble the local package into a distributable tarball")
+        .arg_index("Registry index URL to prepare the package for")
+        .arg_registry("Registry to prepare the package for")
```
epage, line 9: Commit message and PR should talk of why these are needed

src/cargo/ops/cargo_package.rs
```diff
@@ -157,7 +164,74 @@ fn create_package(
...
+                        .iter()
+                        .all(|p| p.publish().as_ref() == Some(allowed))
+                {
+                    Some(allowed[0].clone())
+                } else {
+                    None
+                }
+            })
+        }
+    };
```
epage, line 195: nit: this is too complex of an expression to use combinators.  It particularly is confusing as `pkgs[1..]` is read inside of the combinator.  The general expectation is that the combinator is checking whats inside of itself, with inputs.

src/bin/cargo/commands/package.rs
epage: (this will also help serve as the stabilization report for these flags when proposing to merge these)

src/cargo/ops/cargo_package.rs
```diff
@@ -178,34 +250,100 @@ pub fn package(ws: &Workspace<'_>, opts: &PackageOpts<'_>) -> CargoResult<Option
...
     }
 
-    if opts.list {
-        // We're just listing, so there's no file output
-        Ok(None)
-    } else {
-        Ok(Some(dsts))
+    // Verify all packages in the workspace. This can be done in any order, since the dependencies
+    // are already all in the local registry overlay.
+    if opts.verify {
```
epage, line 287: Are we no verifying when using `--list`?

src/cargo/ops/cargo_package.rs
```diff
@@ -178,34 +250,100 @@ pub fn package(ws: &Workspace<'_>, opts: &PackageOpts<'_>) -> CargoResult<Option
...
+        packages.map(|pkg| (pkg.0.package_id(), pkg)).collect();
+
+    let mut graph = Graph::new();
+    for (pkg, _features) in packages.values() {
+        graph.add(pkg.package_id());
+        for dep in pkg.dependencies() {
+            let dep_id = dep.source_id();
+            if dep.kind() == DepKind::Development || !dep_id.is_path() {
+                continue;
+            };
```
epage, line 330: Should likely have a comment on this

src/cargo/ops/cargo_package.rs
```diff
@@ -178,34 +250,100 @@ pub fn package(ws: &Workspace<'_>, opts: &PackageOpts<'_>) -> CargoResult<Option
...
+
+            let Ok(dep_path) = dep_id.url().to_file_path() else {
+                continue;
+            };
+            let dep_manifest = dep_path.join("Cargo.toml");
+            let Ok(dep_pkg) = ws.load(&dep_manifest) else {
+                continue;
+            };
```
epage, line 338: I feel like there should be a better way of doin this as appending `Cargo.toml` will cause us problems when we have cargo script.

Can we instead walk workspace members to see if we can find one with the same source?  I feel like that will also be less likely to cause problems in other ways

src/cargo/ops/cargo_package.rs
```diff
@@ -900,12 +1055,14 @@ fn run_verify(
...
     pkg: &Package,
     tar: &FileLock,
+    local_reg: Option<&TmpRegistry<'_>>,
     opts: &PackageOpts<'_>,
 ) -> CargoResult<()> {
     let gctx = ws.gctx();
 
     gctx.shell().status("Verifying", pkg)?;
 
+    tar.file().seek(SeekFrom::Start(0))?;
```
epage, line 1065: Why is this needed now when it wasn't before?

src/etc/man/cargo-package.1
```diff
@@ -9,7 +9,8 @@ cargo\-package \[em] Assemble the local package into a distributable tarball
 \fBcargo package\fR [\fIoptions\fR]
 .SH "DESCRIPTION"
 This command will create a distributable, compressed \fB\&.crate\fR file with the
-source code of the package in the current directory. The resulting file will
+source code of the package in the current directory. When run in a workspace,
+this command will create one \fB\&.crate\fR file per crate. The resulting file(s) will
```
epage, line 13: I almost feel like this isn't needed.  Its trying to describe a common pattern within cargo but incompletely

src/etc/man/cargo-package.1
```diff
@@ -9,7 +9,8 @@ cargo\-package \[em] Assemble the local package into a distributable tarball
 \fBcargo package\fR [\fIoptions\fR]
 .SH "DESCRIPTION"
```
epage, line 10: We do need to document the new flags.  They should be in the same commit as the flags

src/cargo/ops/cargo_package.rs
```diff
@@ -82,48 +83,33 @@ struct GitVcsInfo {
...
     ws: &Workspace<'_>,
     pkg: &Package,
     opts: &PackageOpts<'_>,
-) -> CargoResult<Option<FileLock>> {
-    let gctx = ws.gctx();
-    let mut src = PathSource::new(pkg.root(), pkg.package_id().source_id(), gctx);
-    src.update()?;
+) -> CargoResult<FileLock> {
+    let ar_files = build_ar_list(ws, pkg, opts)?;
+    let tarball = create_package(ws, pkg, ar_files)?;
```
epage: If you want to reduce the risk of conflicts, I think this commit stands enough on its own that it can be split out and not wait for the rest of the review

src/cargo/ops/cargo_package.rs
```diff
@@ -1076,3 +1239,101 @@ fn check_filename(file: &Path, shell: &mut Shell) -> CargoResult<()> {
...
+            "name": new_crate.name,
+            "vers": new_crate.vers,
+            "deps": deps,
+            "cksum": cksum,
+            "features": new_crate.features,
+            "yanked": false,
+            "links": new_crate.links,
+            "v": 2,
+        })
+        .to_string();
```
epage, line 1328: How about we use [IndexPackage](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/sources/registry/index/struct.IndexPackage.html) for this?

src/cargo/ops/cargo_package.rs
jneem: No, and this matches the old behavior, where with `--list` we return early before getting to `run_verify`.

src/cargo/ops/cargo_package.rs
jneem: Previously it was the caller's responsibility. Since we now call `run_verify` twice, I thought it was nicer the move the seek inside.

src/cargo/ops/cargo_package.rs
jneem: The first `seek` just got moved inside `run_verify`

src/cargo/ops/cargo_package.rs
jneem: My main motivation here was simplify the return value, because the old behavior was to return `None` when `--list` was set and `Some` otherwise. That was a bit awkward given that the only current caller hard-codes `opts.list` to `false`.

If it's ok with you, I'll go with assert + a comment explaining the situation.

src/cargo/ops/registry/publish.rs
jneem: I don't remember, sorry. Removing it doesn't break any tests...

src/cargo/ops/cargo_package.rs
jneem: Thanks for the pointer! It's a little awkward because `NewCrate` and `IndexPackage` make different choices about `String` vs `InternedString`/`Cow`. But I guess the extra conversions are outweighed by the fact that it'll be easier to keep the index format in sync.

src/cargo/ops/cargo_package.rs
jneem: Yes, that works!

jneem:
Ok, I think this is ready for another look.

While revising it I noticed that the file-locking of the `TmpRegistry` was a bit suspicious, so I reworked that following the example in `core::compiler::layout`

src/cargo/ops/cargo_package.rs
```diff
@@ -186,34 +256,102 @@ pub fn package(ws: &Workspace<'_>, opts: &PackageOpts<'_>) -> CargoResult<Option
...
+    let mut graph = Graph::new();
+    for (pkg, _features) in packages.values() {
+        graph.add(pkg.package_id());
+        for dep in pkg.dependencies() {
+            // Ignore dev-dependencies because they aren't needed for intra-workspace
+            // lockfile generation or verification.
+            if dep.kind() == DepKind::Development || !dep.source_id().is_path() {
+                continue;
+            };
+
```
epage, line 347:
```suggestion
            // Ignore local dev-dependencies because they aren't needed for intra-workspace
            // lockfile generation or verification as they get stripped on publish
            if dep.kind() == DepKind::Development || !dep.source_id().is_path() {
                continue;
            };

```

src/bin/cargo/commands/package.rs
epage, line 9: We need to add these arguments to the shell completions in `src/etc`

src/bin/cargo/commands/package.rs
```diff
@@ -5,6 +5,8 @@ use cargo::ops::{self, PackageOpts};
 pub fn cli() -> Command {
```
epage, line 5: Something I think forgot to ask earlier is whether we should insta- stabilize this or make it unstable with a testing period before stabilizing.

Thoughts?

jneem: Sure, that makes sense, especially since there could be some unforeseen interactions with publishing workspaces. I've made it unstable, and also made the new `--registry` and `--index` flags unstable for `package`.

Quick question about the tests: do we want to duplicate some or all of them so that we can see the difference with/without the unstable flag?

epage: Since this has behavior under stable Cargo, probably

src/bin/cargo/commands/package.rs
```diff
@@ -41,6 +43,15 @@ pub fn cli() -> Command {
 }
 
 pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
+    if args._value_of("registry").is_some() {
+        gctx.cli_unstable()
+            .fail_if_stable_opt("--registry", 13947)?;
+    }
+    if args._value_of("index").is_some() {
+        gctx.cli_unstable()
+            .fail_if_stable_opt("--registry", 13947)?;
```
epage, line 52:
```suggestion
    if args._value_of("index").is_some() {
        gctx.cli_unstable()
            .fail_if_stable_opt("--index", 13947)?;
```

epage, line 52: Shouldn't these check `-Zpackage_workspace` instead of `-Zunstable-options`?

tests/testsuite/package.rs
```diff
@@ -5149,23 +5149,105 @@ fn workspace_with_local_deps() {
...
+        .masquerade_as_nightly_cargo(&["package-workspace"])
         .replace_crates_io(crates_io.index_url())
-        .with_status(101)
-        .with_stdout("")
-        .with_stderr(
+        .with_stdout_data("")
+        .with_stderr_data(
             "\
-[PACKAGING] level1 v0.0.1 ([CWD]/level1)
+[PACKAGING] level3 v0.0.1 ([ROOT]/foo/level3)
```
epage, line 5158: You've handled the `with_stderr` deprecation but a large part of the intent behind that deprecation is to switch to snapshot testing, see #14039 for more details.

The switch in the assert method would ideally also be in the prior commit

tests/testsuite/package.rs
```diff
@@ -5213,22 +5295,23 @@ fn workspace_with_local_deps_packaging_one_fails() {
...
 
     // We can't package just level1, because there's a dependency on level2.
-    p.cargo("package -p level1")
+    p.cargo("package -p level1 -Zpackage-workspace")
+        .masquerade_as_nightly_cargo(&["package-workspace"])
         .replace_crates_io(crates_io.index_url())
         .with_status(101)
-        .with_stdout("")
-        .with_stderr(
+        .with_stdout_data("")
```
epage, line 5302: Note that if the test is very intentionally wanting blank output on a stream, we don't put `str![]` on it

tests/testsuite/package.rs
```diff
@@ -5077,3 +5077,573 @@ path = "examples/z.rs"
...
+            license = "MIT"
+            description = "level2"
+            repository = "bar"
+        "#,
+        )
+        .file("level2/src/lib.rs", "")
+        .build();
+
+    // We can't package just level1, because there's a dependency on level2.
+    p.cargo("package -p level1")
```
epage: Should we have a test case when a version-req compatible `level2` is already published?

src/bin/cargo/commands/package.rs
jneem: Ok, I didn't duplicate all of the tests because some of them differ in ways that are only interesting when `-Zpackage-workspace` is set, but I did a few.

tests/testsuite/package.rs
jneem: Good point, I added one and it raises a possible issue: a case where `-Zpackage-workspace` fails but old cargo didn't.

If you have a local crate `dep@0.1.0` which is published with the same version `0.1.0`, and you have a crate in the same workspace that depends on `dep@0.1.0` then packaging the workspace used to succeed (because the dependency on `dep` will be satisfied by the registry) but now it fails because the overlay source complains about the version conflict between the published `dep@0.1.0` and the overlaid version...

tests/testsuite/package.rs
jneem: Ah, this snapshot testing is much nicer!

tests/testsuite/package.rs
epage: Talked about this in office hour.

Our choices
- Break compatibility on this
- Allow the internal-only source overlap to shadow versions from the original source, rather than error.

`cargo publish` will already fail in this scenario, so the question is strictly about `cargo package` use cases.  crates.io is using `cargo package` but for verifying `.crate` files.  This shouldn't be a problem and we should add a test for this use case.  Multi-package packaging is already fairly busted.  Maybe its ok to break compatibility?

In the end, we should go with one solution, note it in the PR description, and make this an explicit point during the FCP.

epage: Remembered the other use case I was wondering about: packaging for the sake of testing the MSRV for registry dependents: https://github.com/taiki-e/cargo-hack/issues/216

jneem: Makes sense. I think this should also be fine, as again it's per-package instead of per-workspace.

jneem:
@epage I think the latest round of comments have all been addressed now.

tests/testsuite/package.rs
epage: I've tweaked the compatibility note

Issue #13923, opened 2024-05-17
gix integration breaks doc building


nerditation:
to be honest, until I encountered this problem, I didn't know about `gix`, but it seems to cause problem when I build docs. below is the commands to reproduce this error:

```console
❯ echo $env.PROCESSOR_ARCHITECTURE
AMD64

❯ ver

Microsoft Windows [Version 10.0.19045.4412]

❯ cargo --version
cargo 1.80.0-nightly (4de0094ac 2024-05-09)

❯ cargo new --lib dummy
    Creating library `dummy` package
note: see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

❯ cd dummy\

❯ git add .

❯ cargo doc
error: failed to determine package fingerprint for documenting dummy v0.1.0 (Y:\dummy)

Caused by:
  failed to determine the most recently modified file in Y:\dummy

Caused by:
  failed to determine list of files in Y:\dummy

Caused by:
  The value for `core.excludesFile` could not be read from configuration

Caused by:
  path is missing
```

by grepping through the code base for the error messages and some experiments, I managed to pinpoint the failed code is at this line, where the `repo.walkdir_iter()` method returns `Err`:

https://github.com/rust-lang/cargo/blob/8d68ed4fb2554ae7ff2f0a3193c2651e31ee4e1e/src/cargo/sources/path.rs#L519

I did some investigation of the `gix` code, and it turns out my git config file contains a blank value for `core.excludesfile` which gix treats differently than a missing `core.excludesfile` entry.

I opened an issue for gix https://github.com/Byron/gitoxide/issues/1370, but I want to at least mention it here, in case someone else enountered similar problem and want to search for reasons.

also, I'm not sure if this is considered a bug for gix or not, but at least it behaves differently from `libgit2`. when I browse the source code of `cargo` here:

https://github.com/rust-lang/cargo/blob/8d68ed4fb2554ae7ff2f0a3193c2651e31ee4e1e/src/cargo/sources/path.rs#L145-L155

 I discovered this `__CARGO_GITOXIDE_DISABLE_LIST_FILES` environment variable )(undocumented anywhere) so I tried to set it, and then `cargo doc` runs without showing any errors.

Byron:
[...]

PR #13948, opened 2024-05-22
upgrade gix from 0.62 to 0.63

Byron:
This is a bugfix release with various improvements, some of which relevant to safety when checking out (possibly freshly cloned) repositories.

See https://github.com/Byron/gitoxide/security/advisories/GHSA-7w47-3wg8-547c and https://github.com/Byron/gitoxide/security/advisories/GHSA-49jc-r788-3fc9 (might not be published yet as it was waiting for feedback at the time of writing).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -147,12 +147,6 @@ checksum = "4c7f02d4ea65f2c1853089ffd8d2787bdbc63de2f0d29dedbcf8ccdfa0ccd4cf"
 
 [[package]]
 name = "base64"
-version = "0.21.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


weihanglo:
Wonder if this is worth a beta backport. The issue <https://github.com/rust-lang/cargo/issues/13923> shows that it was too strict on malformed gitconfig

epage:
I assumed there would be.  If the risk from this upgrade is too high, we should probably find a different way to fixing beta.

Byron:
> I assumed there would be. If the risk from this upgrade is too high, we should probably find a different way to fixing beta.

I don't know if it helps to know that this is a bugfix release with no features added (unfortunately). If the `gix` version in `beta` is 0.62, then I would expect no issues with this one.

Byron:
I don't know what to do about the CI failure - when I click it, it all looks green. Maybe it's something unrelated, so I am setting this as ready for review.

PR #13955, opened 2024-05-23
pragmatically compare timestamps (as suggested by kornelski)

### LINKED ISSUE #12060: Spurious rebuilds due to filesystem rounding file modification timestamps
### Problem

The issue #2874 was not completely fixed by #2880. I'm still seeing the exact same problem in latest Cargo:

> Dirty libc v0.2.142: the dependency build_script_build was rebuilt (1682215868.647155753s, 647155753ns after last build at 1682215868.000000000s)
> Dirty openssl-sys v0.9.86: the dependency build_script_main was rebuilt (1682214777.891768399s, 891768399ns after last build at 1682214777.000000000s)

### Steps

I'm using Docker and mounting `target/` dir as a Docker volume (in order to cache it). I think the docker volume rounds the timestamps to 1-second precision. However, this doesn't always cause spurious rebuilds, only maybe 25% of the time, and it's always `build_script_build` or `build_script_main`, not other files.

### Possible Solution(s)

I assume Cargo wouldn't want to introduce imprecision in comparison, such as always comparing the integer number of seconds or allowing 1-second difference between timestamps. Therefore, I propose detecting this problem explicitly: there's only one in a billion chance that the nanosecond part will be exactly 0. So if either of the timestamps being compared has exactly 0 nanosecond part, compare only the integer number of seconds. Otherwise compare in full precision. I can make a PR with that.

### Notes

_No response_

### Version

```text
cargo 1.71.0-nightly (d0a4cbcee 2023-04-16)
release: 1.71.0-nightly
commit-hash: d0a4cbcee614fdb7ba66e860e603a00a644d71f8
commit-date: 2023-04-16
host: aarch64-apple-darwin
libgit2: 1.6.3 (sys:0.17.0 vendored)
libcurl: 7.88.1 (sys:0.4.61+curl-8.0.1 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 13.4.0 [64-bit]
```

gilescope:
AKA 'make docker great again'.

Fixes #12060 (as coded here: https://github.com/rust-lang/cargo/compare/master...kornelski:cargo:nanotime )

Let's cut out needless rebuilds. (Thank you @kornelski for suggesting this)

If we are concerned about backward compatibility, we can put this behind a nightly only flag so this slight behaviour change is not the default.

### DIFF
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -1887,9 +1887,9 @@ where
         None
     };
 
-    let mut truncated_precision_count : u64 = 0;
+    let mut truncated_precision_count: u64 = 0;
     let mut total_checked: u64 = 0;
-    let mut first : Option<StaleItem> = None;
+    let mut first: Option<StaleItem> = None;
     for path in paths {
         let path = path.as_ref();
 
@@ -1941,12 +1941,12 @@ where
         let fresh = if truncated_precision {
             truncated_precision_count += 1;
             if first.is_none() {
-               first = Some(StaleItem::ChangedFile {
-                   reference: reference.to_path_buf(),
-                   reference_mtime,
-                   stale: path.to_path_buf(),
-                   stale_mtime: path_mtime,
-               }); 
+                first = Some(StaleItem::ChangedFile {
+                    reference: reference.to_path_buf(),
+                    reference_mtime,
+                    stale: path.to_path_buf(),
+                    stale_mtime: path_mtime,
+                });
             }
             path_mtime.unix_seconds() <= reference_mtime.unix_seconds()
         } else {
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -1952,6 +1952,13 @@ where
         });
     }
 
+    // If a few files in many have a .0ns mtime then still regard them as changed files.
+    if total_checked > 1 && truncated_precision_count == 1 {
+        return first;
+    } else if truncated_precision_count > 0 {
+        debug!("ignoring files that look changed due to mtime precision differences");
+    }
+
     debug!(
         "all paths up-to-date relative to {:?} mtime={}",
         reference, reference_mtime
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -1887,6 +1887,9 @@ where
         None
     };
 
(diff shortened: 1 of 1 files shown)


kornelski (approved):

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1927,7 +1927,20 @@ where
...
-        if path_mtime <= reference_mtime {
+        //
+        // Additionally, the build can span different volumes, some which support high-precision
+        // file timestamps and some that don't (e.g. mounted Docker volumes). This can make
+        // one of the timestamps lose the nanosecond part and appear up to a second in the past.
+        let truncated_precision =
+            path_mtime.nanoseconds() == 0 || reference_mtime.nanoseconds() == 0;
+
+        let fresh = if truncated_precision {
+            path_mtime.unix_seconds() <= reference_mtime.unix_seconds()
```
weihanglo, line 1938: Could we have a `debug!` or a `trace!` log when the precision is truncated?

weihanglo:
@rfcbot fcp merge

This helps mitigate the needless rebuild when a filesystem supports only low-precision mtime, with a caveat:

>  there's only one in a billion chance that the nanosecond part will be exactly 0.

I propose to merge without any nightly flag this, as the probability is relatively low.

joshtriplett:
@rfcbot concern one-in-a-billion-chances-happen-at-scale

I'm all for handling timestamp truncation in a reasonable way. But I don't want us making the assumption that one-in-a-billion chances don't happen. If you're running a large fleet of systems doing many many operations per day, one-in-a-billion chances can be a regular occurrence.

Can we guard against this to ensure we only hit this case due to timestamp granularity, and not by random chance?

ehuss:
I don't think 1 billion is an accurate representation of the precision or probability that the sub-second time will be 0. Different systems and filesystems have different precision. For example, NTFS is 100-nanosecond, some virtualization environments or limited hardware might be much less.

I'm not sure why the updated patch here only checks for one file. I would assume if one side or the other is stored on a truncated filesystem, then all files would be truncated, and if not some number from 0 to less than all could possibly be 0.

gilescope:
Doing more experimentation, I've seen drift from 0ns to 14ns on a large codebase.
I'd be inclined to be lenient on the check only if every file is < 25ns (overridable by `CARGO_MAX_MTIME_DRIFT_NS`)

kornelski:
I've noticed that the timestamps are wrong only on `build_script_build` files, so I'm wondering whether there's a different bug in the way build scripts are written or fingerprinted that causes this problem.

PR #13956, opened 2024-05-23
Improve error description when deserializing partial field struct 

### LINKED ISSUE #13688: Unhelpful error messages with env-vars that don't set all fields of a config struct
### Problem

I have been using the `gitoxide` integration for a while, via setting env-vars so it doesn't impact running the stable compiler:

```
CARGO_UNSTABLE_GITOXIDE_CHECKOUT=true
CARGO_UNSTABLE_GITOXIDE_FETCH=true
CARGO_UNSTABLE_GITOXIDE_INTERNAL_USE_GIT2=false
CARGO_UNSTABLE_GITOXIDE_SHALLOW_DEPS=true
CARGO_UNSTABLE_GITOXIDE_SHALLOW_INDEX=true
```

After updating to a version that includes https://github.com/rust-lang/cargo/pull/13592 any usage of cargo starts failing with

```console
> cargo test -- --ignored zero_byte
error: missing field `list_files`
```

which makes no mention of _where_ this field is missing

### Steps

```
> CARGO_UNSTABLE_GITOXIDE_FETCH=true cargo build
error: missing field `list_files`
```

### Possible Solution(s)

_No response_

### Notes

https://github.com/rust-lang/cargo/pull/13687 will likely fix this specific issue by making the fields of this struct default, but the general issue of the error message not having enough context will remain.

### Version

```text
cargo 1.79.0-nightly (a59aba136 2024-03-28)
release: 1.79.0-nightly
commit-hash: a59aba136aab5510c16b0750a36cbd9916f91796
commit-date: 2024-03-28
os: NixOS 23.11.0 [64-bit]
```

linyihai:

Fixes #13688

one commit add test, one commit fixed and update the test.

### DIFF
--- a/src/cargo/util/context/de.rs
+++ b/src/cargo/util/context/de.rs
@@ -35,7 +35,7 @@ macro_rules! deserialize_method {
                 .ok_or_else(|| ConfigError::missing(&self.key))?;
             let Value { val, definition } = v;
             let res: Result<V::Value, ConfigError> = visitor.$visit(val);
-            res.map_err(|e| e.with_key_context(&self.key, definition))
+            res.map_err(|e| e.with_key_context(&self.key, Some(definition)))
         }
     };
 }
@@ -60,7 +60,7 @@ impl<'de, 'gctx> de::Deserializer<'de> for Deserializer<'gctx> {
                 CV::Boolean(b, def) => (visitor.visit_bool(b), def),
             };
             let (res, def) = res;
-            return res.map_err(|e| e.with_key_context(&self.key, def));
+            return res.map_err(|e| e.with_key_context(&self.key, Some(def)));
         }
         Err(ConfigError::missing(&self.key))
     }
@@ -178,7 +178,7 @@ impl<'de, 'gctx> de::Deserializer<'de> for Deserializer<'gctx> {
         let Value { val, definition } = value;
         visitor
             .visit_enum(val.into_deserializer())
-            .map_err(|e: ConfigError| e.with_key_context(&self.key, definition))
+            .map_err(|e: ConfigError| e.with_key_context(&self.key, Some(definition)))
     }
 
     // These aren't really supported, yet.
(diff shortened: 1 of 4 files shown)


tests/testsuite/config.rs
```diff
@@ -1863,3 +1863,54 @@ fn trim_paths_parsing() {
...
+    let gctx = GlobalContextBuilder::new()
+        .env("CARGO_FOO_BAR_BAZ", "true")
+        .build();
+    assert_error(
+        gctx.get::<Foo>("foo").unwrap_err(),
+        "\
+failed to parse config key `foo`
+
+Caused by:
+  could not load config key `foo.bar`
```
linyihai, line 1889: There is no better way to deal with the problem of lost environment variables. Because we never pass `CARGO_FOO_BAR`. So the `defination` will be None here.

linyihai:
Other tests will fail and need to be manually updated, more than 20 in total. I can update them later

src/cargo/util/context/mod.rs
```diff
@@ -2030,18 +2030,22 @@ impl ConfigError {
         }
     }
 
+    fn is_missing_field(&self) -> bool {
+        self.error.to_string().contains("missing field")
+    }
```
epage, line 2035: Could we have [serde tell us a field is missing](https://docs.rs/serde/latest/serde/de/trait.Error.html#method.missing_field) so we don't have to be doing string matching?

src/cargo/util/context/de.rs
```diff
@@ -345,11 +345,26 @@ impl<'de, 'gctx> de::MapAccess<'de> for ConfigMapAccess<'gctx> {
...
+                    return e;
+                }
+                e.with_key_context(
+                    &self.de.key,
+                    self.de
+                        .gctx
+                        .get_cv_with_env(&self.de.key)
+                        .map_or(None, |cv| {
+                            cv.map_or(None, |cv| Some(cv.get_definition().clone()))
+                        }),
```
weihanglo, line 365: nit:
```suggestion
                        .ok()
                        .and_then(|cv| cv.map(|cv| cv.get_definition().clone())),
```

src/cargo/util/context/mod.rs
```diff
@@ -2030,18 +2030,22 @@ impl ConfigError {
         }
     }
 
+    fn is_missing_field(&self) -> bool {
+        self.error.to_string().contains("missing field")
```
weihanglo, line 2034: Not a blocker. Just wonder if it is possible to compare with error types instead of strings.

src/cargo/util/context/mod.rs
linyihai: Good idea. By Implementing the `missing_field` and return the `MissingField` error which we defined

src/cargo/util/context/mod.rs
linyihai: Have Added a custom `MissingField` errot to match.

weihanglo:
Looks awesome. Thanks!


@bors r+

PR #13957, opened 2024-05-23
Fix wrong type of rustc-flags in documentation

iawia002:
When I tried to add some custom configuration according to the documentation in https://doc.rust-lang.org/cargo/reference/config.html#configuration-format, I encountered the following error:

```
error: expected a string, but found a array for `rustc-flags` in /xxx/.cargo/config.toml
```

I believe the type of `rustc-flags` should be a string:

https://github.com/rust-lang/cargo/blob/95b921ba592ec968d4638e124f5273bb42a25528/src/doc/src/reference/config.md?plain=1#L1243-L1253


weihanglo (approved):

weihanglo:
@bors r+

Thanks for the contribution!

PR #13958, opened 2024-05-23
Include `lints.rust.unexpected_cfgs.check-cfg` in the fingerprint

Urgau:

When changing the `--check-cfg` args in the `lints.rust.unexpected_cfgs.check-cfg` lint config, the build should be restarted as the arguments we pass to the compiler change, and they can change the diagnostics output by generating new or removing some `unexpected_cfgs` warning(s).

Look at the before and after test (separate commit).

### Additional information

Similar to https://github.com/rust-lang/cargo/pull/13012 which did that for the declared features.
Contrary to that PR, I didn't add a new `DirtyReason` variant, since even the `[lints]` table didn't get one.

r? @epage

### DIFF
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -80,6 +80,7 @@
 //! config settings[^5]                        | ✓           |
 //! is_std                                     |             | ✓
 //! `[lints]` table[^6]                        | ✓           |
+//! `[lints.rust.unexpected_cfgs.check-cfg]`   | ✓           |
 //!
 //! [^1]: Build script and bin dependencies are not included.
 //!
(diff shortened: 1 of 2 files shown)


src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1420,12 +1421,31 @@ fn calculate_normal(
...
+                        if let Ok(check_cfgs) =
+                            toml::Value::try_into::<Vec<String>>(check_cfg.clone())
+                        {
+                            lint_check_cfg = check_cfgs;
+                        }
+                    }
+                }
+            }
+        }
+    }
```
epage, line 1440: Rather than re-looking up this entry, we should see if there is a good way to make this reuseable

Example areas to explore
- I assume we fingerprint the build script check-cfg.  Can we mix the two
- We already check `lint_rustflags`.  Should we instead move the lookup to be a part of that?

Urgau: > I assume we fingerprint the build script check-cfg. Can we mix the two

I don't know, I'm not sure we do anything special for build-script. I think they are just seen as deps in fingerprint.

> We already check lint_rustflags. Should we instead move the lookup to be a part of that?

`lint_rustflags` creates the different `--warn`, `--allow`, ... flags, which we could add the `--check-cfg` args at some point, but not right now since we need to gate all `--check-cfg` args behind the compiler supporting check-cfg stably, which can only be done when we have access to the `Unit` (I think) which is way after `lint_rustflags` is called

Urgau: Created #13975 (so the hack can be tracked) and updated the code with your suggested wording (as well as changing it to a `HACK`).

PR #13960, opened 2024-05-24
Include vcs_info even if workspace is dirty

### LINKED ISSUE #13695: Include cargo_vcs_info.json even for dirty working directories

torhovland:

Fixes #13695.

Generates and packages the vcs_info file even if the worktree is dirty, as long as `--allow-dirty` is passed.

Tests are included.


src/cargo/ops/cargo_package.rs
weihanglo, line 1: See <https://github.com/rust-lang/cargo/issues/13695#issuecomment-2130353956>. Still in dicussion though.

weihanglo (commented):
Thanks for the contribution. Could you adjust the commit order a bit?

The first commit adding a new test (`issue_13695_dirty_vcs_info`) to show the old behavior, and should pass the CI pipeline. The next commits fix both the behavior and test assertions.

weihanglo:
r? weihanglo

src/cargo/ops/cargo_package.rs
epage: I had made the assumption that `dirty:true` would be present.  I wasn't aware there was a question on it

weihanglo: Nice. We are aligned now.

@torhovland could you add that and we'll do an FCP then?

torhovland: @weihanglo OK, done.

weihanglo:
@rfcbot fcp merge

With this change, the  `.cargo_vcs_info.json` will always be generated regardless the dirty status of the VCS. 

One risk is that people may already rely on the existence of vcs info file to determine if the repository is dirty. That's actually [documented in the cargo-package manpage][1]. However, a package without `.cargo_vcs_info.json` only implies that it was packaged with `--allow-dirty` passed, but not the repository was actually dirty at that time. Hence I think the risk should be low, and this patch makes it more accurate.

---

@rfcbot concern doc-update

Since we are going to include VCS info file always if in a Git repository. This might give people a false impression that Cargo handles package publish provenance against the linked repository. We may want to explicitly call out that Cargo hasn't yet provided the feature in [the manpage][1], as people can fake the file. We should update the doc of the JSON format as well, and add a line saying that new fields may be added in the future and not considered as breaking changes.


[1]: https://doc.rust-lang.org/nightly/cargo/commands/cargo-package.html

src/cargo/ops/cargo_package.rs
```diff
@@ -80,6 +80,8 @@ struct VcsInfo {
 #[derive(Serialize)]
 struct GitVcsInfo {
     sha1: String,
+    /// Indicate whether or not the Git worktree is dirty.
+    dirty: bool,
```
epage, line 84: Should we skip-on-serialize-if-default?  That way the file stays the same in the regular cases?

torhovland: That makes sense to me too. Just let me know when and if to make this change.

weihanglo: I am fine with either.

weihanglo:
@torhovland I've posted <https://github.com/rust-lang/cargo/pull/13984> to address the concern. If that is merged, would you mind adjusting the doc [here](https://github.com/rust-lang/cargo/blob/0b72605141073a7e8cd50214a53b5ec053e5e820/src/doc/man/cargo-package.md?plain=1#L33-L35) and adding a line about the new `dirty` field below [this line](https://github.com/rust-lang/cargo/blob/0b72605141073a7e8cd50214a53b5ec053e5e820/src/doc/man/cargo-package.md?plain=1#L61-L62)?

weihanglo:
@rfcbot resolve doc-update

torhovland:
@weihanglo I've adjusted the docs.

src/cargo/ops/cargo_package.rs
weihanglo: @torhovland feel free to go ahead and do it :)

weihanglo (approved):

src/cargo/ops/cargo_package.rs
torhovland: Updated.

Turbo87:
> One risk is that people may already rely on the existence of vcs info file to determine if the repository is dirty.

FYI crates.io is parsing this file these days. might be good to at least ping the crates.io team on semantic changes like this 😉

weihanglo:
Thanks for letting us know, @Turbo87!

Does this change make sense to crates.io?

Turbo87:
> Does this change make sense to crates.io?

yeah, should be fine now that we know about it and can treat the new field correctly :)

epage:
> FYI crates.io is parsing this file these days. might be good to at least ping the crates.io team on semantic changes like this 😉

When I saw Adam's blogpost, I informed him of this.  That wasn't yer "crates.io is parsing" though.  Was the use of this previously communicated in some other way so we would know to tell crates.io?

djc:
I wonder if there would be value here in being a little more fine-grained, for example by noting whether any of the `src` files are modified? For example, indicatif is included in https://lawngnome.github.io/divine-provenance/ but because I usually publish from my local dev environment which includes several untracked examples (which I've used for testing) hasn't included `vcs_info` for most recent releases. Rather than just concluding the workspace is "dirty" which could mean any number of things, in terms of "provenance" it would be more interesting to say whether some subset of the published files are modified relative to the `HEAD` commit.

epage:
What value do you want to derive from that? Keep in mind the presence and content of this file is not sufficient for provenance since people can hack of cargo for custom publishes or take a custom `.crate` and push that.

djc:
It's just more precise information on exactly how the "dirtiness" of the Git index state affects the published code. @LawnGnome any thoughts on whether this would be helpful?

LawnGnome:
> any thoughts on whether this would be helpful?

I do see where you're coming from, but speaking for myself, I don't think it would help the checks I'm doing specifically — my code is pretty much doing the equivalent of a full `diff -ur` between the crate file and repo anyway to be able to generate fine-grained diffs, so the nuance of _how_ dirty a workspace is when a crate is packaged is something that is essentially going to be reconstructed anyway.

I can see that it might be helpful for someone who wanted to do a quick analysis without having to get into the weeds of having both trees available to diff them, but my suspicion — although this is obviously t-cargo's call — is that it's probably more complexity than the `cargo` implementation would really want to be responsible for, and it has the issues @epage noted.

It's entirely possible I'm looking at this too much through my own lens, though, so I may well be missing a use case here. But from where I am on a sleepy Friday morning, I guess I'd be roughly -0 on adding more detailed metadata. Maybe +0 if `cargo` could generate it _really_ easily, although then we'd have to figure out what the schema would look like.

(I guess the other case where it would be useful would be if the repo is no longer available or the revision was never pushed, but in those cases, there's really no concept of provenance at any level anyway.)

djc:
Okay, sorry for the noise/distraction!

weihanglo:
Thanks for the conversation, people. Based on the FCP, I'll merge this now.

@bors r+

torhovland:
Rebased on master, which interestingly succeeded without any conflicts.

@bors r+

weihanglo (approved):

weihanglo:
Thank you!

@bors r+

PR #13970, opened 2024-05-28
fix: adjust custom err from cert-check due to libgit2 1.8 change

weihanglo:

libgit2 disallows overriding errors from certificate check since v1.8.0,
so we store the error additionally and unwrap it later.

I believe it is due to this change: https://github.com/libgit2/libgit2/commit/9a9f220119d9647a352867b24b0556195cb26548

Actually, in the doc of `git_error_set` it does state that the custom message might [be overridden by libgit2 internals](https://github.com/libgit2/libgit2/blob/6c5520f334e5652d5f0476c00a3188d1d97754e7/include/git2/sys/errors.h#L24-L28).

I am not sure if this should be fixed in upstream libgit2 or by our own,
but it's better to do it now then waiting for libgit2 1.8.2

### Additional information

r? ehuss

### DIFF
--- a/src/cargo/sources/git/known_hosts.rs
+++ b/src/cargo/sources/git/known_hosts.rs
@@ -23,6 +23,7 @@
 //! and revoked markers. See "FIXME" comments littered in this file.
 
 use crate::util::context::{Definition, GlobalContext, Value};
+use crate::CargoResult;
 use base64::engine::general_purpose::STANDARD;
 use base64::engine::general_purpose::STANDARD_NO_PAD;
 use base64::Engine as _;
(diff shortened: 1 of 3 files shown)


weihanglo:
This commit shows that the code change are compatible with both libgit2 1.7 and 1.8: <https://github.com/rust-lang/cargo/commit/aa2ad97a77fcf126bdd7ceb73a0c5fbf936a2abb>. I believe this is good to merge now without waiting for a new git2-rs release.

ehuss:
Yea, it was a little precarious how the error was getting shuttled around. Thank you very much for looking into this and getting it fixed!

@bors r+

PR #13971, opened 2024-05-28
fix: using `--release/debug` and `--profile` together becomes an error

heisen-li:

part of #13629 

issue https://github.com/rust-lang/cargo/issues/13629#release-is-ignored-when-paired-with-profile

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -558,11 +558,18 @@ Run `{cmd}` to see possible targets."
 
     fn get_profile_name(
         &self,
-        gctx: &GlobalContext,
         default: &str,
         profile_checking: ProfileChecking,
     ) -> CargoResult<InternedString> {
         let specified_profile = self._value_of("profile");
+
+        let err_message = |flag: &str| {
+            format!(
+                "\
+the `--{flag}` flag can not be specified with the `--profile` flag
+Please remove one of the flags."
+            )
+        };
 
         // Check for allowed legacy names.
         // This is an early exit, since it allows combination with `--release`.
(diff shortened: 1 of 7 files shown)


epage:
Is there something this PR is waiting on for why it is a Draft

heisen-li:
> Is there something this PR is waiting on for why it is a Draft

Sorry, I thought there would be some documentation that would need to be changed, but it doesn't seem to be there when I looked today.

src/cargo/util/command_prelude.rs
```diff
@@ -558,12 +558,19 @@ Run `{cmd}` to see possible targets."
...
     ) -> CargoResult<InternedString> {
         let specified_profile = self._value_of("profile");
 
+        let err_message = |flag: &str| {
+            format!(
+                "\
+the `--{flag}` flag can not be specified with the `--profile` flag
+Please remove one of the flags."
+            )
+        };
```
weihanglo, line 572: This should work and below we can just call `check_conflict(flag)?` instead.

```suggestion
        fn check_conflict(flag: &str) -> CargoResult<()> {
            bail!(
                "\
the `--{flag}` flag can not be specified with the `--profile` flag
Please remove one of the flags."
            )
        };
```

weihanglo, line 572: This should work and below we can just call `conflict(flag)?` instead.

```suggestion
        fn conflict(flag: &str) -> CargoResult<()> {
            bail!(
                "\
the `--{flag}` flag can not be specified with the `--profile` flag
Please remove one of the flags."
            )
        };
```

weihanglo (commented):
Looks good to me. Thanks!

epage:
Should we leverage clap's [conflict](https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with) support?  We likely would need validation lower level like this.  Unsure which style of error messages we'd prefer.
```diff
    fn arg_release(self, release: &'static str) -> Self {
        self._arg(
            flag("release", release)
                .short('r')
+              .conflicts_with("profile")
                .help_heading(heading::COMPILATION_OPTIONS),
        )
    }
...
        .arg(flag(
            "debug",
            "Build in debug mode (with the 'dev' profile) instead of release mode",
-        ))
+       ).conflicts_with("profile"))
```
You don't need to put it on `--profile`.

weihanglo:
> Should we leverage clap's [conflict](https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with) support? We likely would need validation lower level like this. Unsure which style of error messages we'd prefer.


This sound better and we can remove most of those tests. (maybe leave one for preventing regression?)

weihanglo (approved):
Thanks for the contribution.

Since this turns a warning into a hard error, I would like to schedule the merge after 1.81 nightly is out.

src/cargo/util/command_prelude.rs
weihanglo, line 1: https://github.com/rust-lang/cargo/pull/13971#issuecomment-2142877924

@heisen-li have you considered this option?

heisen-li: Thanks to you guys, I used this suggestion. Regarding tests, I keep a relevant test for each command (`rustc`, `build`, `check`, `install`).

weihanglo:
@bors r+

Thanks!

PR #13979, opened 2024-05-29
Add `cargo update --breaking`

torhovland:
Related to #12425.

There are two kinds of manifest mutations here. In `upgrade_manifests` we have to mutate `cargo::core::manifest::Manifest` so that `resolve_ws` does what it needs to do, and outputs a correct lock file. Then, in `write_manifest_upgrades` we mutate `cargo::util::toml_mut::manifest::Manifest`, because that is how we can preserve the existing formatting in the manifest files on disk.

Some of the code here is copied from `cargo-edit`.

# Comparison with `cargo upgrade`

Running on the Cargo source itself gives the following:

```
❯ cargo upgrade --dry-run --incompatible allow --compatible ignore 
    Updating 'https://github.com/rust-lang/crates.io-index' index
    Checking benchsuite's dependencies
    Checking capture's dependencies
    Checking cargo's dependencies
name               old req compatible latest  new req note      
====               ======= ========== ======  ======= ====      
anstream           0.6.13  0.6.14     0.6.14  0.6.13  compatible
anstyle            1.0.6   1.0.7      1.0.7   1.0.6   compatible
anyhow             1.0.82  1.0.86     1.0.86  1.0.82  compatible
itertools          0.12.1  0.12.1     0.13.0  0.13.0            
libc               0.2.154 0.2.155    0.2.155 0.2.154 compatible
opener             0.7.0   0.7.1      0.7.1   0.7.0   compatible
openssl            0.10.57 0.10.64    0.10.64 0.10.57 compatible
openssl-sys        =0.9.92 0.9.92     0.9.102 =0.9.92 pinned    
pulldown-cmark     0.10.3  0.10.3     0.11.0  0.11.0            
security-framework 2.10.0  2.11.0     2.11.0  2.10.0  compatible
semver             1.0.22  1.0.23     1.0.23  1.0.22  compatible
serde              1.0.199 1.0.203    1.0.203 1.0.199 compatible
serde-untagged     0.1.5   0.1.6      0.1.6   0.1.5   compatible
serde_json         1.0.116 1.0.117    1.0.117 1.0.116 compatible
tar                0.4.40  0.4.41     0.4.41  0.4.40  compatible
thiserror          1.0.59  1.0.61     1.0.61  1.0.59  compatible
toml               0.8.12  0.8.14     0.8.14  0.8.12  compatible
toml_edit          0.22.12 0.22.14    0.22.14 0.22.12 compatible
unicode-width      0.1.12  0.1.13     0.1.13  0.1.12  compatible
    Checking cargo-credential's dependencies
    Checking cargo-credential-1password's dependencies
    Checking cargo-credential-libsecret's dependencies
    Checking cargo-credential-macos-keychain's dependencies
    Checking cargo-credential-wincred's dependencies
    Checking cargo-platform's dependencies
    Checking cargo-test-macro's dependencies
    Checking cargo-test-support's dependencies
    Checking cargo-util's dependencies
    Checking cargo-util-schemas's dependencies
    Checking crates-io's dependencies
    Checking home's dependencies
    Checking mdman's dependencies
    Checking resolver-tests's dependencies
    Checking rustfix's dependencies
    Checking semver-check's dependencies
    Checking xtask-build-man's dependencies
    Checking xtask-bump-check's dependencies
    Checking xtask-stale-label's dependencies
note: Re-run with `--pinned` to upgrade pinned version requirements
note: Re-run with `--verbose` to show all dependencies
  local: cargo
  unchanged: annotate-snippets, base64, bytesize, cargo-credential, cargo-credential-libsecret, cargo-credential-macos-keychain, cargo-credential-wincred, cargo-platform, cargo-test-macro, cargo-test-support, cargo-util, cargo-util-schemas, cargo_metadata, clap, color-print, core-foundation, crates-io, criterion, curl, curl-sys, filetime, flate2, git2, git2-curl, gix, glob, handlebars, hex, hmac, home, http-auth, humantime, ignore, im-rc, indexmap, jobserver, lazycell, libgit2-sys, libloading, memchr, miow, os_info, pasetors, pathdiff, percent-encoding, pkg-config, proptest, rand, regex, rusqlite, rustfix, same-file, serde-value, serde_ignored, sha1, sha2, shell-escape, similar, snapbox, supports-hyperlinks, supports-unicode, tempfile, time, tracing, tracing-chrome, tracing-subscriber, unicase, unicode-xid, url, varisat, walkdir, windows-sys
warning: aborting upgrade due to dry run

❯ target/debug/cargo update --breaking --dry-run -Zunstable-options
    Updating crates.io index
   Upgrading itertools ^0.12.1 -> ^0.13.0
   Upgrading pulldown-cmark ^0.10.3 -> ^0.11.0
    Updating crates.io index
     Locking 3 packages to latest compatible versions
    Updating itertools v0.12.1 -> v0.13.0
    Updating pulldown-cmark v0.10.3 -> v0.11.0
    Updating pulldown-cmark-escape v0.10.0 -> v0.11.0
warning: not updating any files due to dry run
```

In both cases we see an upgrade of `itertools` to `^0.13.0` and `pulldown-cmark` to `^0.11.0`.

The diff to the manifest (when it isn't a dry run) is as follows:

```
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ humantime = "2.1.0"
 ignore = "0.4.22"
 im-rc = "15.1.0"
 indexmap = "2.2.6"
-itertools = "0.12.1"
+itertools = "0.13.0"
 jobserver = "0.1.31"
 lazycell = "1.3.0"
 libc = "0.2.154"
@@ -74,7 +74,7 @@ pathdiff = "0.2.1"
 percent-encoding = "2.3.1"
 pkg-config = "0.3.30"
 proptest = "1.4.0"
-pulldown-cmark = { version = "0.10.3", default-features = false, features = ["html"] }
+pulldown-cmark = { version = "0.11.0", default-features = false, features = ["html"] }
 rand = "0.8.5"
 regex = "1.10.4"
 rusqlite = { version = "0.31.0", features = ["bundled"] }
```

# TODO

- [x] In the case of `--incompatible`, we also need to let `update_lockfile` use `upgrades` in order to only update the incompatible dependencies.
- [x] Testing all the different cases of package sources, version requirements, pinned versions, renamed dependencies, inherited workspace dependencies, multiple versions of the same dependency, selecting a subset `--package`, etc.
- [x] Passing tests.
- [x] Implement suggestions from reviews.
- [x] The preservation of formatting in manifest files should be improved.
- [x] Compare with `cargo upgrade`.


ehuss:
r? epage
If you don't mind.

tests/testsuite/update.rs
```diff
@@ -1659,3 +1659,56 @@ fn update_with_missing_feature() {
         )
         .run();
 }
+
+#[cargo_test]
+fn update_incompatible() {
+    Package::new("compatible", "1.0.0").publish();
+    Package::new("incompatible", "1.0.0").publish();
+    Package::new("pinned", "1.0.0").publish();
```
epage: Also renamed

tests/testsuite/update.rs
```diff
@@ -1659,3 +1659,56 @@ fn update_with_missing_feature() {
...
+                [dependencies]
+                compatible = "1.0"
+                incompatible = "1.0"
+                pinned = "<99.0"
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("check").run();
```
epage, line 1686: `generate-lockfile`

tests/testsuite/update.rs
```diff
@@ -1659,3 +1659,56 @@ fn update_with_missing_feature() {
...
+                [package]
+                name = "bar"
+                version = "0.0.1"
+                edition = "2015"
+                authors = []
+
+                [dependencies]
+                compatible = "1.0"
+                incompatible = "1.0"
+                pinned = "<99.0"
```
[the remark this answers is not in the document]
epage: Granted, it should still be ignored, just the name of "pinned" could be confusing

tests/testsuite/update.rs
```diff
@@ -1659,3 +1659,56 @@ fn update_with_missing_feature() {
...
+    Package::new("incompatible", "1.0.1").publish();
+    Package::new("pinned", "1.0.1").publish();
+
+    Package::new("compatible", "2.0.0").publish();
+    Package::new("incompatible", "2.0.0").publish();
+    Package::new("pinned", "2.0.0").publish();
+
+    p.cargo("update --incompatible")
+        .with_status(1)
+        .with_stderr_contains("[ERROR] unexpected argument '--incompatible' found")
```
epage: Let's start with the name `--breaking`

Incompatible is a broader term that can be "incomptible with this version req" and not just "semver incompatible".  "Breaking" is semantically more tied to semver

epage: (and it is easier to add a short flag for)

src/cargo/ops/cargo_update.rs
```diff
@@ -41,6 +45,162 @@ pub fn generate_lockfile(ws: &Workspace<'_>) -> CargoResult<()> {
...
+
+        let new_summary = member
+            .manifest()
+            .summary()
+            .clone()
+            .map_dependencies(|dependency| {
+                let req = if let OptVersionReq::Req(current) = dependency.version_req() {
+                    let query = crate::core::dependency::Dependency::parse(
+                        dependency.package_name(),
+                        None,
```
epage, line 78: I believe we should skip on renames

src/cargo/ops/cargo_update.rs
```diff
@@ -41,6 +45,162 @@ pub fn generate_lockfile(ws: &Workspace<'_>) -> CargoResult<()> {
...
+                            .iter()
+                            .map(|s| s.as_summary())
+                            .map(|s| s.version().clone())
+                            .max()
+                    } else {
+                        None
+                    };
+
+                    if let Some(latest) = latest.clone() {
+                        if !current.matches(&latest) {
```
epage, line 105: I'm not seeing this restricted to only semver operator

src/cargo/ops/cargo_update.rs
```diff
@@ -41,6 +45,162 @@ pub fn generate_lockfile(ws: &Workspace<'_>) -> CargoResult<()> {
...
+                                .shell()
+                                .status_with_color(
+                                    "Upgrading",
+                                    format!(
+                                        "{} {} -> v{}",
+                                        dependency.package_name(),
+                                        current,
+                                        latest.to_string()
+                                    ),
+                                    &style::GOOD,
```
epage, line 116: This is going to duplicate messages if two members have the same dependency, right?

tests/testsuite/update.rs
epage: We should also have a `ws_incompatible` to demonstrate workspace inherited dependencies get updated

tests/testsuite/update.rs
```diff
@@ -1659,3 +1659,56 @@ fn update_with_missing_feature() {
         )
         .run();
 }
+
+#[cargo_test]
+fn update_incompatible() {
```
epage: We should have a separate dry-run test case that does the same thing

tests/testsuite/update.rs
```diff
@@ -1659,3 +1659,56 @@ fn update_with_missing_feature() {
...
+    Package::new("compatible", "2.0.0").publish();
+    Package::new("incompatible", "2.0.0").publish();
+    Package::new("pinned", "2.0.0").publish();
+
+    p.cargo("update --incompatible")
+        .with_status(1)
+        .with_stderr_contains("[ERROR] unexpected argument '--incompatible' found")
+        .run();
+
+    p.cargo("update")
```
epage: We need to verify the manifests were updated

src/cargo/util/toml/mod.rs
```diff
@@ -60,6 +61,12 @@ pub fn read_manifest(
     let mut warnings = Default::default();
     let mut errors = Default::default();
 
+    debug!(
+        "read_manifest; path={}; source-id={}",
+        path.display(),
+        source_id
+    );
```
epage, line 68: These already get logged at the `trace` level

epage:
When this is ready for review, what I expect to see for the commits
1. The refactor to create `cargo_update.rs`
2. Tests
3. Feature with test updates

tests/testsuite/update.rs
```diff
@@ -1659,3 +1659,99 @@ fn update_with_missing_feature() {
...
+
+    Package::new("compatible", "1.0.1").publish();
+    Package::new("incompatible", "2.0.0").publish();
+
+    p.cargo("update --incompatible")
+        // .env("CARGO_LOG", "cargo::ops=trace,cargo::util::toml=debug")
+        .with_stderr(
+            "\
+[UPDATING] `[..]` index
+[UPGRADING] incompatible ^1.0 -> v2.0.0
```
epage, line 1734: Should we use https://github.com/killercup/cargo-edit/blob/522160cdfd057bfc993034ce8d5f776782acf609/src/version.rs#L145

tests/testsuite/update.rs
torhovland: Changed to `--breaking`. Although I will have to update this initial test as well when rewriting the Git history.

epage: Before you next want me to look at this, could you clean up the history?

tests/testsuite/update.rs
torhovland: Added these cases and more.

tests/testsuite/update.rs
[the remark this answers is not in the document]
torhovland: Testing both.

tests/testsuite/update.rs

tests/testsuite/update.rs
torhovland: Done.

tests/testsuite/update.rs
torhovland: I added `--dry-run` as a separate step in the main test. Is that OK?

src/cargo/util/toml/mod.rs
torhovland: I plan on cleaning out debug logging in the final PR.

src/bin/cargo/commands/update.rs
```diff
@@ -84,11 +92,18 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
...
         recursive: args.flag("recursive"),
         precise: args.get_one::<String>("precise").map(String::as_str),
+        incompatible: args.flag("incompatible"),
         to_update,
         dry_run: args.dry_run(),
         workspace: args.flag("workspace"),
         gctx,
     };
+
+    ops::update_manifests(&ws, &update_opts)?;
```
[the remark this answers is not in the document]
torhovland: Done.

src/cargo/ops/cargo_update.rs

src/cargo/ops/cargo_update.rs
torhovland: Changed now. At the moment I am _only_ allowing the caret operator. Not sure if that is too restrictive.

src/cargo/ops/cargo_update.rs
torhovland: Correct. Should we keep messages in a `HashSet` and use that to ignore duplicates?

tests/testsuite/update.rs
torhovland: The manifests are _correct_ now, but they don't preserve formatting as well as they should. Any ideas how to address this? I guess I should look more into cargo-edit.

tests/testsuite/update.rs
torhovland: Done.

src/cargo/ops/mod.rs
```diff
@@ -44,7 +44,6 @@ pub(crate) mod cargo_compile;
 pub mod cargo_config;
 mod cargo_doc;
 mod cargo_fetch;
-mod cargo_generate_lockfile;
```
epage, line 47: Thank you for splitting this out as it makes things much clearer.  I think there was something I had overlooked.  I thought you split `cargo_generate_lockfile` rather than straight rename.  It feels odd to favor one name over the other for the command.

What is your motivation for renaming?  `cargo_generate_lockfile` is so small I wonder if its a separate file.  I had considered suggesting it `lockfile.rs` but we appear to have that.  Maybe its worth being in there still?

Thoughts?

src/cargo/ops/cargo_update.rs
epage: Messages or more source-level information.

tests/testsuite/update.rs
epage: I tend to prefer separate tests. `--dry-run` doesn't have to be as exhaustive.  However, my preference on this isn't too strong, so I leave it to you.

tests/testsuite/update.rs
```diff
@@ -1740,34 +1772,113 @@ fn update_breaking() {
...
+                [workspace.dependencies]
+                ws  = "2.0"
+            "#,
+    );
+
+    let crate_manifest = p.read_file("foo/Cargo.toml");
+
+    // FIXME: Not sure how to preserve all formatting here.
+    snapbox::assert_data_eq!(
+        &crate_manifest,
```
epage, line 1848: Can you please add these asserts to the previous commit?

tests/testsuite/update.rs
```diff
@@ -1740,34 +1772,113 @@ fn update_breaking() {
...
+
+                [workspace.dependencies]
+                ws  = "2.0"
+            "#,
+    );
+
+    let crate_manifest = p.read_file("foo/Cargo.toml");
+
+    // FIXME: Not sure how to preserve all formatting here.
+    snapbox::assert_data_eq!(
```
epage, line 1847: We should be using `cargo_test_support::assert_e2e().eq()` to get some cargo-specific configuration

tests/testsuite/update.rs
```diff
@@ -1740,34 +1772,113 @@ fn update_breaking() {
...
+        &crate_manifest,
+        r#"
```
epage, line 1849: This should be created with `cargo_test_support::str![]` which will give us snapshot updating capabilities

tests/testsuite/update.rs
```diff
@@ -1661,36 +1689,40 @@ fn update_breaking() {
...
             r#"
+                # Check if formatting is preserved
+
                 [package]
-                name = "foo"
-                version = "0.0.1"
-                edition = "2015"
-                authors = []
+                name  =  "foo"
+                version  =  "0.0.1"
```
epage, line 1708: Please put all  of these formatting checks in the previous commit

tests/testsuite/update.rs
```diff
@@ -1740,34 +1772,113 @@ fn update_breaking() {
         .alternative(true)
         .publish();
 
-    p.cargo("update --dry-run --breaking")
-        .with_status(1)
-        .with_stderr_contains("[ERROR] unexpected argument '--breaking' found")
+    p.cargo("update -Zunstable-options --dry-run --breaking")
+        .masquerade_as_nightly_cargo(&["update-breaking"])
+        // .env("CARGO_LOG", "cargo::ops=trace")
```
epage, line 1777: Don't forget to remove this

tests/testsuite/update.rs
```diff
@@ -1740,34 +1772,113 @@ fn update_breaking() {
...
+        .with_stderr_unordered(
```
epage, line 1778: Is the ordering non-deterministic?  Our existing `cargo update` tests do not use `unordered`

tests/testsuite/update.rs
```diff
@@ -1637,3 +1638,220 @@ fn update_with_missing_feature() {
...
+
+    Package::new("just-foo", "1.0.1").publish();
+    Package::new("just-bar", "1.0.1").publish();
+    Package::new("shared", "1.0.1").publish();
+    Package::new("ws", "1.0.1").publish();
+
+    Package::new("just-foo", "2.0.0").publish();
+    Package::new("just-bar", "2.0.0").publish();
+    Package::new("shared", "2.0.0").publish();
+    Package::new("ws", "2.0.0").publish();
```
epage, line 1949: A test case for `update --breaking just-foo shared ws` that we need to cover is if a breaking version depends on a newer, compatible version of a package.  The upgrade of that transitive dependency should still go through.

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +221,215 @@ pub fn print_lockfile_changes(
...
+        .acquire_package_cache_lock(CacheLockMode::DownloadExclusive)?;
+
+    let mut registry = PackageRegistry::new(opts.gctx)?;
+    registry.lock_patches();
+
+    for member in ws.members_mut() {
+        debug!("upgrading manifest for `{}`", member.name());
+
+        *member.manifest_mut().summary_mut() = member.manifest().summary().clone().try_map_dependencies(
+            |dependency| -> CargoResult<_> {
```
epage, line 247: Can we split this out?  For myself, I find overly long closures tend to make the code harder to read.

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +221,215 @@ pub fn print_lockfile_changes(
...
+                            &current.comparators
+                        );
+                    }
+                } else {
+                    trace!(
+                        "skipping dependency `{}` without a simple version requirement: {}",
+                        name,
+                        version_req
+                    );
+                }
```
epage, line 350: Would there be a good way to shift to early returns / breaks rather than this deep nesting for trivial branches?

I find code like this tends to be harder to read and I generally recommend nesting logic to be focused on business cases while using early returns for bookkeeping makes things a lot clearer.

src/cargo/ops/cargo_update.rs
epage: Yes, `^` is the "semver" operator and is the only one we should be supporting atm (hence the name change of the flag)

tests/testsuite/update.rs
```diff
@@ -1637,3 +1638,220 @@ fn update_with_missing_feature() {
         )
         .run();
 }
+
+#[cargo_test]
+fn update_breaking() {
+    registry::alt_init();
+    Package::new("compatible", "1.0.0").publish();
```
epage, line 1673: Other test cases
- incompatible-prerelease
- incompatible-yanked

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +221,215 @@ pub fn print_lockfile_changes(
...
+                                    opts.gctx.shell().status_with_color(
+                                        "Upgrading",
+                                        format!(
+                                            "{} {} -> v{}",
+                                            name,
+                                            current,
+                                            latest.to_string()
+                                        ),
+                                        &style::GOOD,
+                                    )?;
```
epage, line 310: We're reporting `Upgrading {name} {current_req} -> v{latest_version}`.  Should we be doing `Upgrading {name} {current_req} -> {new_req}`?

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +221,215 @@ pub fn print_lockfile_changes(
...
+
+/// Update manifests with upgraded versions, and write to disk. Based on cargo-edit.
+pub fn write_manifests(
+    ws: &Workspace<'_>,
+    opts: &UpdateOptions<'_>,
+    upgrades: &HashMap<String, Version>,
+) -> CargoResult<()> {
+    if !opts.breaking {
+        return Ok(());
+    }
```
epage, line 368: nit: Could we instead do this based on `!upgrades.is_empty()`.  Either way, it should still work.  The benefit comes when we do the other half of this feature which is `--precise <breaking-version>`.

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +221,215 @@ pub fn print_lockfile_changes(
...
+
+        if opts.dry_run {
+            opts.gctx.shell().warn(format!(
+                "not upgrading manifest `{}` due to dry run",
+                manifest_path.display()
+            ))?;
+        } else {
+            debug!("writing upgraded manifest to {}", manifest_path.display());
+            local_manifest.write()?;
+        }
```
epage, line 427: - Should we avoid the `write` / dry-run message if this manifest didn't change?
- Do we get duplicate `write`s / dry-run messages when there is a root package in the workspace?

src/cargo/ops/cargo_update.rs
```diff
@@ -554,6 +777,92 @@ fn fill_with_deps<'a>(
     }
 }
 
+/// Upgrade an existing requirement to a new version.
+/// Copied from cargo-edit.
+fn upgrade_requirement(req: &str, version: &semver::Version) -> CargoResult<Option<String>> {
```
epage, line 782: nit: if we want to reduce the size of `cargo_update.rs`, this might also make sense to live in `toml_mut`.

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +221,215 @@ pub fn print_lockfile_changes(
...
+                    dep_key,
+                    dep_item,
+                ) {
+                    Ok(dependency) => dependency,
+                    Err(err) => {
+                        opts.gctx
+                            .shell()
+                            .warn(&format!("ignoring {dep_key}, unsupported entry: {err}"))?;
+                        continue;
+                    }
```
epage, line 394: Unsure whether its best to error or warn

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +221,215 @@ pub fn print_lockfile_changes(
...
+                if let crate::util::toml_mut::dependency::MaybeWorkspace::Other(_) =
+                    dependency.source_id(opts.gctx)?
+                {
+                    if let (Some(current), Some(latest)) =
+                        (dependency.version(), upgrades.get(&dependency.name))
+                    {
+                        if let Some(new_req_string) = upgrade_requirement(current, latest)? {
+                            let mut dep = dependency.clone();
+
+                            if let Some(Source::Registry(source)) = dependency.source() {
```
epage, line 406: I feel like this could benefit from
```rust
let Some(_) = _ else {
    continue
};
```

src/cargo/ops/cargo_update.rs
epage, line 782: We should probably pull in the unit tests as well: https://github.com/killercup/cargo-edit/blob/d395bc51bf1fd7b97edeabadbb1df78d83643f03/src/version.rs#L295

epage:
Thanks for all of your work on this!  This is looking good and making a lot of progress!

tests/testsuite/update.rs
torhovland: It is. These can come in any order:

```
Updating `alternative` index
```

Also, this can come first or last, depending on when `bar/Cargo.toml` is upgraded.

```
Upgrading multiple ^2.0 -> v3.0.0
```

tests/testsuite/update.rs
torhovland: Done.

tests/testsuite/update.rs

tests/testsuite/update.rs

tests/testsuite/update.rs
torhovland: Done.

tests/testsuite/update.rs

tests/testsuite/update.rs
torhovland: Added.

tests/testsuite/update.rs
torhovland: Added.

src/cargo/ops/cargo_update.rs
torhovland: Do you want to have it changed? It would certainly be easier to just return a CargoResult.

torhovland:
> Thanks for all of your work on this! This is looking good and making a lot of progress!

And thank you for constructive review feedback!

I have some other work I need to do, but will get back to this later.

tests/testsuite/update.rs
```diff
@@ -1637,3 +1639,318 @@ fn update_with_missing_feature() {
...
+                members  =  ["foo", "bar"]
+
+                [workspace.dependencies]
+                ws  =  "1.0"  # This line gets partially rewritten
+            "#]],
+    );
+
+    let crate_manifest = p.read_file("foo/Cargo.toml");
+
+    // FIXME: Not sure how to preserve all formatting here.
```
epage, line 1923: Tracking this FIXME

tests/testsuite/update.rs
```diff
@@ -1646,21 +1646,124 @@ fn update_breaking_unstable() {
...
+            "\
+[ERROR] the `--breaking` flag is unstable, pass `-Z unstable-options` to enable it
+See https://github.com/rust-lang/cargo/issues/12425 for more information about the `--breaking` flag.
+",
+        )
+        .run();
+}
+
+#[cargo_test]
+fn update_breaking_dry_run() {
```
epage, line 1672: Could you have this separate out in the prior commit?

src/cargo/util/toml_mut/manifest.rs
```diff
@@ -276,21 +276,24 @@ impl LocalManifest {
...
+            //         "found virtual manifest at {}, but this command requires running against an \
+            //              actual package in this workspace.",
+            //         self.path.display()
+            //     );
+            // } else {
+            anyhow::bail!(
+                "missing expected `package`, `project`, or `workspace` fields in {}",
+                self.path.display()
+            );
+            // }
```
epage, line 296: To be resolved

src/cargo/ops/cargo_update.rs
epage: You went out of your way to not fail on this.  What was the motivation?

tests/testsuite/update.rs
```diff
@@ -1646,21 +1646,124 @@ fn update_breaking_unstable() {
...
-        .with_status(1)
-        .with_stderr_contains("[ERROR] unexpected argument '--breaking' found")
+        .with_stderr(
+            "\
+[UPDATING] `dummy-registry` index
+[UPGRADING] incompatible ^1.0 -> v2.0.0
+[UPGRADING] ws ^1.0 -> v2.0.0
+[LOCKING] 2 packages to latest compatible versions
+[UPDATING] incompatible v1.0.0 -> v2.0.0
+[UPDATING] ws v1.0.0 -> v2.0.0
```
epage, line 1725: Do we need both UPGRADING and UPDATING?  Is UPDATING enough to imply UPGRADING?

They are covering distinct aspects so I'm not entirely sure on this.

tests/testsuite/update.rs
```diff
@@ -1646,21 +1646,124 @@ fn update_breaking_unstable() {
...
+[WARNING] not updating lockfile due to dry run
+[WARNING] not upgrading manifest `[..]/Cargo.toml` due to dry run
+[WARNING] not upgrading manifest `[..]/foo/Cargo.toml` due to dry run
```
epage, line 1728: I feel like we don't need three dry-run warnings.

- It'd be messy to implement but we could say "not writing the following files due to dry run" and list them
- We could say "not writing updates due to dry run"
- We could be even more generic and say "aborting update due to dry-run" like some other commands do

tests/testsuite/update.rs
```diff
@@ -1936,20 +2042,25 @@ fn update_breaking_specific_packages() {
...
     Package::new("transitive-incompatible", "2.0.0").publish();
 
-    p.cargo("update --breaking just-foo shared ws")
-        .with_status(1)
-        .with_stderr_contains("[ERROR] unexpected argument '--breaking' found")
-        .run();
-
-    p.cargo("update just-foo shared ws")
-        .with_stderr(
+    // FIXME: This should not upgrade just-bar
```
epage, line 2045: Tracking FIXME

epage: What if in the `breaking` case, we called `resolve_ws` instead of `ops::update_lockfile(&ws, &update_opts, &upgrades)?;`  That would make the behavior more like `cargo check` without having to muck with `keep`

src/doc/src/reference/unstable.md
```diff
@@ -378,6 +379,21 @@ It's possible to update `my-dependency` to a pre-release with `update -Zunstable
 This is because `0.1.2-pre.0` is considered compatible with `0.1.1`.
 It would not be possible to upgrade to `0.2.0-pre.0` from `0.1.1` in the same way.
 
+## update-breaking
+
+* Tracking Issue: [#12425](https://github.com/rust-lang/cargo/issues/12425)
+
+The `update-breaking` feature allows upgrading dependencies to breaking versions
```
epage, line 386: Not quite correct to refer to this as `update-breaking` (backticks included) as a feature name as that implies its `-Zupdate-breaking`  when this is `-Zunstable-options`.  We have other examples of `-Zunstable-options` here, like `--out-dir`, that you can look to as examples

tests/testsuite/update.rs
```diff
@@ -1637,3 +1639,429 @@ fn update_with_missing_feature() {
...
+
+                [workspace.dependencies]
+                ws  = "2.0"
+            
+"#]],
+    );
+
+    let crate_manifest = p.read_file("foo/Cargo.toml");
+
+    // FIXME: Not sure how to preserve all formatting here.
```
epage, line 1923: Tracking this FIXME

epage: I believe `cargo add` is able to preserve formatting in these cases, so I don't think this is a problem with `toml_edit` but how we are using it.

epage: The distinction is that this code is overwriting the toml table
```rust
*dep_item = dep.to_toml(manifest_path);
```
while cargo-add updates it
```rust
dep.update_toml(&crate_root, &mut dep_key, dep_item);
```

src/cargo/util/toml_mut/manifest.rs
torhovland: Does the change look OK to you?

tests/testsuite/update.rs
torhovland: I have changed the upgrade message to show the updated version requirement rather than the latest version, so the upgrade and update messages are more distinct now. But I am fine with dropping one or the other if you like.

```
[UPGRADING] incompatible ^1.0 -> ^2.0
[UPDATING] incompatible v1.0.0 -> v2.0.0
```

src/cargo/ops/cargo_update.rs

src/cargo/util/toml/mod.rs
torhovland: This one is removed. I'm leaving some debug and trace logging that I think makes sense.

tests/testsuite/update.rs
torhovland: I'm sorting the crates before upgrading, to make this deterministic. Fixed.

src/cargo/ops/cargo_update.rs
torhovland: Agreed. Fixed.

src/cargo/ops/cargo_update.rs

src/cargo/ops/cargo_update.rs

src/cargo/ops/cargo_update.rs

src/cargo/ops/cargo_update.rs
torhovland: Nice! Fixed.

src/cargo/ops/cargo_update.rs

tests/testsuite/update.rs
torhovland: Fixed.

tests/testsuite/update.rs
torhovland: That works indeed. Thanks. Although I did have to pepper the code with `dry_run` arguments. Not sure which approach is best. The actual FIXME above was unrelated to this, as it was about not honouring the `to_update` collection.

src/doc/src/reference/unstable.md
torhovland: Fixed.

tests/testsuite/update.rs
torhovland: Excellent! Fixed.

benches/benchsuite/benches/resolve.rs
```diff
@@ -33,6 +33,7 @@ fn do_resolve<'gctx>(gctx: &'gctx GlobalContext, ws_root: &Path) -> ResolveInfo<
     let force_all_targets = ForceAllTargets::No;
     // Do an initial run to download anything necessary so that it does
     // not confuse criterion's warmup.
+    let dry_run = false;
```
epage, line 36: Good call on doing this to make the intent of a `false` parameter clearer!

src/cargo/ops/cargo_update.rs
epage: Good call on applying concepts from `cargo upgrade`.  However, I think the reason I did this was because it was decoupled from the development of cargo and didn't want new features in Cargo (or bugs in my code) from stopping things.

tests/testsuite/update.rs
epage: Having the messages more distinct helps.  Part of the challenge (and why we didn't want a separate `cargo upgrade` command) is the distinction between an "update" and an "upgrade" can be confusing to users.  This also causes a lot more output.  However, people also want to know when things are happening.

How about we keep this for now and use the post-merge call-for-testing to see what feedback we get.  If we don't have it, the conversation is only about adding it or not.  If we have it, the conversation is about whether we should have it and how we can tweak what we have.

tests/testsuite/update.rs
epage: Something feels off to me about that.  Overly casual?  A blend of detail (files) but vagueness (which ones?)?

Our existing messages:
```console
$ cargo update -n
warn: not updating lockfile due to dry run
$ cargo clean -n
warn: no files deleted due to --dry-run
$ cargo publish -n
warn: aborting upload due to dry run
$ cargo rm foo -n
warn: aborting remove due to dry run
$ cargo add foo -n
warn: aborting add due to dry run
```
(`cargo add` and `cargo rm` were modeled off of `cargo publish` without any deeper thought)

tests/testsuite/update.rs
epage: Ah, the source of non-determinism is that `Workspace::members_mut` iterates over `Packages::packages` which is as `HashMap<PathBuf, Package>`.

In that case, the `sort` is likely fine

tests/testsuite/update.rs
```diff
@@ -1891,9 +1910,28 @@ fn update_breaking() {
...
+        .masquerade_as_nightly_cargo(&["update-breaking"])
+        .with_stderr(
+            "\
+[UPDATING] `dummy-registry` index
+[UPGRADING] multiple ^2.0 -> ^3.0
+[UPGRADING] shared ^1.0 -> ^2.0
+[UPDATING] `alternative` index
+[UPGRADING] alternative-1 ^1.0 -> ^2.0
+[UPGRADING] alternative-2 ^1.0 -> ^2.0
+[UPGRADING] incompatible ^1.0 -> ^2.0
```
epage, line 1923: This is neither the original version requirement nor what we are updating it to (neither has `^`)

tests/testsuite/update.rs
```diff
@@ -1637,3 +1639,429 @@ fn update_with_missing_feature() {
...
+        .with_status(1)
+        .with_stderr_contains("[ERROR] unexpected argument '--breaking' found")
+        .run();
+
+    let root_manifest = p.read_file("Cargo.toml");
+    assert_e2e().eq(
+        &root_manifest,
+        str![[r#"
+
+                # Check if formatting is preserved. Nothing here should change, due to dry-run.
```
epage, line 1736: nit: for something like this, I'd make assign the manifest into a variable and use that with `file` and with `eq!()`.  This makes the intent clearer that it is to be left unchanged.

src/cargo/ops/mod.rs
epage: Thinking more on this, the primary thing people look for is `update` and not `generate_lockfile`.

Also, adding `Cargo.toml` editing makes this so `resolver` and `lockfile` names are less applicable.

Seems like this rename is fine for now.  We can evolve it as the programs evolve.

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +214,251 @@ pub fn print_lockfile_changes(
...
+        return Ok(upgrades);
+    }
+
+    // Updates often require a lot of modifications to the registry, so ensure
+    // that we're synchronized against other Cargos.
+    let _lock = ws
+        .gctx()
+        .acquire_package_cache_lock(CacheLockMode::DownloadExclusive)?;
+
+    let mut registry = PackageRegistry::new(opts.gctx)?;
```
epage, line 234: The caller handles the policy of checking `opts.breaking`, maybe we should not do that here and stop accepting `UpdateOptions`, only accepting `GlobalContext`

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +214,251 @@ pub fn print_lockfile_changes(
...
+    let upgrade_message = format!("{} {} -> {}", name, current, new_req_string);
+    trace!(upgrade_message);
+
+    if upgrade_messages.insert(upgrade_message.clone()) {
+        opts.gctx
+            .shell()
+            .status_with_color("Upgrading", &upgrade_message, &style::GOOD)?;
+    }
+
+    upgrades.insert(name.to_string(), latest.clone());
```
epage, line 359: What if we have the same dependency name with different versions on different registries?

epage, line 359: Since you are using the package name, rather than the dependency name, we should likely always `insert` the same thing here (assuming its the same registry).

There is an optimization that could be done to skip some of the above steps if an entry already exists in `upgrades` but likely not important.

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +214,251 @@ pub fn print_lockfile_changes(
...
+                    continue;
+                };
+
+                let Some(latest) = upgrades.get(&dependency.name) else {
+                    trace!(
+                        "skipping dependency without an upgrade: {}",
+                        dependency.name
+                    );
+                    continue;
+                };
```
epage, line 426: What about the following cases
- A `path` dependency with a `version` specified
- A `path` dependency with a `version` specified **and** a version-only dependency with the same name

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +214,251 @@ pub fn print_lockfile_changes(
...
+
+                let Some(new_req_string) = upgrade_requirement(current, latest)? else {
```
epage, line 428: Why do we call `upgrade_requirement in both `write_manifests` and `upgrade_manifests`?

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +214,251 @@ pub fn print_lockfile_changes(
...
+
+    let req = OptVersionReq::Req(VersionReq::parse(&latest.to_string())?);
+    let mut dep = dependency.clone();
+    dep.set_version_req(req);
+    Ok(dep)
+}
+
+/// Update manifests with upgraded versions, and write to disk. Based on cargo-edit.
+/// Returns true if any file has changed.
+pub fn write_manifests(
```
epage, line 369: We should probably be more specific, like `write_manifest_upgrades`

src/cargo/ops/cargo_update.rs
```diff
@@ -207,6 +214,251 @@ pub fn print_lockfile_changes(
...
+                        "skipping dependency `{}` because the version requirement didn't change",
+                        dependency.name
+                    );
+                    continue;
+                };
+
+                let Some(Source::Registry(source)) = dependency.source() else {
+                    trace!("skipping non-registry dependency: {}", dependency.name);
+                    continue;
+                };
```
epage, line 439: Shouldn't we also be skipping this in `upgrade_dependency`?

tests/testsuite/update.rs
```diff
@@ -1791,6 +1806,7 @@ fn update_breaking() {
     Package::new("pre-release", "1.0.0").publish();
     Package::new("yanked", "1.0.0").publish();
     Package::new("ws", "1.0.0").publish();
+    Package::new("shared", "1.0.0").publish();
```
epage, line 1809: Shouldn't this be in the prior commit?

tests/testsuite/update.rs
epage: I think I'm missing something.  Could you explain the situation?

src/cargo/util/toml_mut/manifest.rs
epage: We should just remove the error check.  I think its an artifact of a previous version of `cargo add` from before it got merged into cargo and it had to do more of the heavy lifting.

tests/testsuite/update.rs
torhovland: OK, I changed this to "aborting update due to dry run".

tests/testsuite/update.rs
torhovland: The fix for the FIXME above was this:

```
    if !opts.to_update.is_empty() && !opts.to_update.contains(&name.to_string()) {
        trace!("skipping dependency `{}` not selected for upgrading", name);
        return Ok(dependency);
    }
```

Your suggestion to use `resolve_ws` works as well as ops::update_lockfile. The only difference seemed to be that `resolve_ws` insisted on writing the lock file, so it needed a `dry_run` argument. This caused a number of changes across the code base. If that looks fine to you, let's keep it. Otherwise, we can revert to using ops::update_lockfile.

tests/testsuite/update.rs
torhovland: The issue here is that the upgrading messages are logged in `upgrade_manifests`, which does not know the exact dependency syntax in the TOML files. I could move the logging to `write_manifests` (which does handle the TOML) and improve the messages, but then they come after the update messages (from `resolve_ws`), which looks odd.

I could also just drop the upgrading messages, as you've hinted at before. Or I can wipe out any carets in the messages (but I guess that's equally wrong if the manifest actually has carets).

What do you think?

src/cargo/ops/cargo_update.rs
torhovland: I can remove that extra check, but not `UpdateOptions`, as we are also using `opts.to_update`.

src/cargo/ops/cargo_update.rs
torhovland: The first case is already covered:

```toml
bar  =  { path  =  "../bar", version  =  "0.0.1" }  # Comment
```

src/cargo/ops/cargo_update.rs
torhovland: OK, I changed it to just return an error.

src/cargo/util/toml_mut/manifest.rs
torhovland: Removed.

tests/testsuite/update.rs

src/cargo/ops/cargo_update.rs
torhovland: Check removed.

src/cargo/ops/cargo_update.rs
torhovland: I added a test dependency: `conflicting-name`. And upgraded the hashmap to include the source id.

src/cargo/ops/cargo_update.rs
torhovland: I added a test case, `conflicting-source-type`. Hope that is what you had in mind.

src/cargo/ops/cargo_update.rs
torhovland: `write_manifests` needs to know what version req to put in the manifest file. `upgrade_manifests` just needs it for the upgrade message. If we could use the latest version rather than the new version req in the upgrade message, we wouldn't need to call it. Also, we cannot reuse the output from the first in the second, because `write_manifests` will call it with the actual syntax in the TOML file and preserve the formatting, while `upgrade_manifests` doesn't know about any formatting and will call it with a stringified version req.

src/cargo/ops/cargo_update.rs

tests/testsuite/update.rs
torhovland: Fixed.

tests/testsuite/update.rs
epage: Let's keep it for now and see what feedback we get.

While its not the exact version requirement, this also helps in distinguishing it from the update messages (makes it clearer that its *a* version requirement being affected).

tests/testsuite/update.rs
epage: Honestly,. I would just pass the parameters directly and not use `opts` at all in this case (which means `UpdateOptions` doesn't need `breaking`).

The alternative is to move all of the logic in the `command/update.rs` into the `ops` and `UpdateOptions` can be passed to that but the lower level functions probably shouldn't accept it.

src/cargo/ops/cargo_update.rs
epage: Honestly,. I would just pass the parameters directly and not use `opts` at all in this case (which means `UpdateOptions` doesn't need `breaking`).

The alternative is to move all of the logic in the `command/update.rs` into the `ops` and `UpdateOptions` can be passed to that but the lower level functions probably shouldn't accept it.

src/cargo/ops/cargo_update.rs
epage: I worry that use of `package` for renaming could make it we aren't testing the case I'm interested in.

Also, its not quite conflicting.

tests/testsuite/update.rs
epage: Thanks!  I thought this was going in a different direction.  Unsure which approach is better overall though we can see over time.  I just hesitate over the modification of `keep` so avoiding that is a win for me.

src/cargo/ops/cargo_update.rs
torhovland: OK, I implemented the first option (not passing `upgrade_opts`).

src/cargo/ops/cargo_update.rs
torhovland: I changed the names to something more accurate than `conflicting-*`. And I updated the multiple source types test to not do renaming.

epage:
Thanks for all of the work you put into this!

@bors r+

PR #13980, opened 2024-05-29
refactor: Transition direct assertions from cargo-test-support to snapbox

epage:

Cargo has a bespoke testing framework for functional tests
- Extra stuff for us to maintain
- Don't leverage benefits from contributions related to other projects
- Less incentive to be thoroughly documented

UI tests are written using snapbox.  The latest release of snapbox (#13963) was geared at supporting cargo's needs in the hope that we can consolidate on testing frameworks.

Besides having a single set of semantics, benefits we'd gain include
- Updating of test snapshots
- Fancier redacting of test output (e.g. #13973)

This is the first incremental step in this direction.  This replaces direct assertions with snapbox assertions.  This still leaves all of the CLI output assertions. These will be done incrementally.


tests/testsuite/artifact_dep.rs
```diff
@@ -777,19 +784,22 @@ fn build_script_with_selected_dashed_bin_artifact_and_lib_true() {
...
     } else {
-        compare::assert_match_exact(
-            "[..]/artifact/bar-baz-[..]/bin\n\
-        [..]/artifact/bar-baz-[..]/bin/baz_suffix-[..]",
+        assert_e2e().eq(
             &build_script_output,
+            str![[r#"
+[ROOT]/foo/target/debug/deps/artifact/bar-baz-[..]/bin
+[ROOT]/foo/target/debug/deps/artifact/bar-baz-[..]/bin/baz_suffix-[..]
+
```
hi-rustin, line 801: Can I ask when we need to add a blank line? I tried this locally, it seemed to pass the test after I removed it.
image

tests/testsuite/alt_registry.rs
```diff
@@ -1492,8 +1494,10 @@ dependencies = [
 [[package]]
 name = "foo"
 version = "0.1.0"
-source = "sparse+http://[..]/"
-checksum = "458c1addb23fde7dfbca0410afdbcc0086f96197281ec304d9e0e10def3cb899""#,
+source = "sparse+http://127.0.0.1:[..]/index/"
```
hi-rustin, line 1497: It seems that `source = "source = "sparse+http://[..]/index/"` also passes the test.

hi-rustin (commented):
Thanks for adding it.

tests/testsuite/alt_registry.rs
epage: Yes.  I had snapshot updating process all of these and re-added the globbing.  I tried to be more limited in where I added globs.  I also tried to find where we could replace globs with something else (e.g. the `[BROKEN_PIPE]`) so snapshot updating can be more automatic.

I could possibly detect localhost with a port and redact the port (`127.0.0.1:[PORT]`).  The main risk with redactions is we'd redact something that another test explicitly checks (e.g. most of the time we don't care about the `.crate` size reported by `cargo publish` but some tests specifically do)

hi-rustin: Got it. Make sense. Thanks!

tests/testsuite/artifact_dep.rs
hi-rustin: Got it.

hi-rustin (approved):
Looks good to me. Thanks! 👍

hi-rustin:
I guess you need to rerun the tests.

crates/cargo-test-support/src/compare.rs
```diff
@@ -96,6 +97,121 @@ pub fn assert_ui() -> snapbox::Assert {
...
+/// Updating of snapshots is controlled with the `SNAPSHOTS` environment variable:
+///
+/// - `skip`: do not run the tests
+/// - `ignore`: run the tests but ignore their failure
+/// - `verify`: run the tests
+/// - `overwrite`: update the snapshots based on the output of the tests
+///
+/// # Patterns
+///
+/// - `[..]` is a character wildcard, stopping at line breaks
```
weihanglo, line 115: Guess we might want to add `[CWD]` in the future. Nevertheless it is not a blocker.

weihanglo (approved):
Thank you all for working on and reviewing this!

PR #13989, opened 2024-05-31
fix(toml): Ensure targets are in a deterministic order

### LINKED ISSUE #13988: Non determinist vendoring of git dependencies
### Problem

The output of `cargo vendor` is not determinist for git dependencies. Depending on the machine where the command is executed, the `Cargo.toml` of the vendored dependencies can be different. All tests were done on various versions of Ubuntu but with the same version of cargo. The difference is only about order of arrays of tables in the `Cargo.toml`.

I made a minimal reproduction example in [this repo](https://github.com/stormshield-guillaumed/vendor-repro) for which I ran the vendor command on my computer. You can see in [this CI run](https://github.com/stormshield-guillaumed/vendor-repro/actions/runs/9318120254) that running `cargo vendor` results in a diff with master. Also, the diff is not the same between the 3 jobs, so it doesn't seems like a mistake on my part.

### Steps

_No response_

### Possible Solution(s)

I'm not very familiar with the cargo codebase, but after digging a bit, it seems to only happen with git dependencies because they are treated differently. Those dependencies are not published on crates.io so their `Cargo.toml` is not normalised. Due to this, the `Cargo.toml` is normalised before the copy to the vendor directory. This normalisation yields different results depending on the machine.

### Notes

_No response_

### Version

```text
cargo 1.80.0-nightly (431db31d0 2024-05-28)
release: 1.80.0-nightly
commit-hash: 431db31d0dbeda320caf8ef8535ea48eb3093407
commit-date: 2024-05-28
```

epage:

With #13713, we enumerate all targets in `Cargo.toml` on `cargo publish` and `cargo vendor`.
However, the order of the targets is non-determistic. This can be annoying for comparing the published `Cargo.toml` across releases but even worse is the churn it causes for `cargo vendor`.

So we sort all the targets during publish.
This keeps costs minimal with the following risks
- If the non-determinism shows up in a way that affects developers during development
- If there is a reason the user wants to control target order for explicit targets

Fixes #13988

As for writing of tests, I'm unsure why none of our existing tests failed which makes it unclear to me what would be needed to write a test for this.

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -2777,6 +2777,11 @@ fn prepare_targets_for_publish(
         };
         prepared.push(target);
     }
+    // Ensure target order is deterministic, particularly for `cargo vendor` where re-vendoring
+    // shuld not cause changes.
+    //
+    // `unstable` should be deterministic because we enforce that `t.name` is unique
+    prepared.sort_unstable_by_key(|t| t.name.clone());
 
     if prepared.is_empty() {
         Ok(None)


weihanglo (approved):
This makes sense. I just checked the other sections. They should all be controlled by BTreeMap so should be fine.

Thanks for the fix!

PR #13992, opened 2024-05-31
fix(config): Ensure `--config net.git-fetch-with-cli=true` is respected

### LINKED ISSUE #13991: `cargo update --config net.git-fetch-with-cli=true` fetches using git2
### Problem

As reported in renovatebot/renovate#29277
- Private git repo dependency fetching worked on 1.77 but not 1.78
- They used `net.git-fetch-with-cli=true` on the command-line

I git bisected it to #13479, specifically f525e1f383fc0d0a7adbdfb1bac6d6228ae79b33

### Steps

```console
$ git clone https://github.com/crate-ci/cargo-release.git
$ cd cargo-release
$ rm -rf ~/.cargo/git  # avoiding the github fast path
$ cargo update --config net.git-fetch-with-cli=true
```
Notice that the fetch is done using libgit2 instead of git.  To make this more obvious, I ran from master with a `dbg!` around the `git_fetch_with_cli` check at
https://github.com/rust-lang/cargo/blob/d04355a8c42f8c173a0c02625477edd668fc0cdd/src/cargo/sources/git/utils.rs#L1008

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

_No response_

epage:

#13479 changed the global context initialization order so that command line stuff is processed after we read some config.
This had a side effect of breaking `--config net.git-fetch-with-cli=true`.
I reverted the change to restore support for `--config`.

Fixes #13991

### Additional information

This reverts commit https://github.com/rust-lang/cargo/commit/f525e1f383fc0d0a7adbdfb1bac6d6228ae79b33.

This removes color control from warnings for unstable features.
For some reason this removed color support from `cargo -Zhelp` in the
tests but I can't reproduce it locally.

The most important thing was getting the config fix in.
There are two follow ups
- Can we have the config working *and* color?
- Why did this fail for this field and not the others we already had
  tests for?

I ran out my immediate time box for looking into these.

### DIFF
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -1020,6 +1020,31 @@ impl GlobalContext {
         unstable_flags: &[String],
         cli_config: &[String],
     ) -> CargoResult<()> {
+        for warning in self
+            .unstable_flags
+            .parse(unstable_flags, self.nightly_features_allowed)?
+        {
+            self.shell().warn(warning)?;
+        }
+        if !unstable_flags.is_empty() {
+            // store a copy of the cli flags separately for `load_unstable_flags_from_config`
+            // (we might also need it again for `reload_rooted_at`)
+            self.unstable_flags_cli = Some(unstable_flags.to_vec());
+        }
+        if !cli_config.is_empty() {
+            self.cli_config = Some(cli_config.iter().map(|s| s.to_string()).collect());
+            self.merge_cli_args()?;
+        }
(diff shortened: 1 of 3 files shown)


epage:
I suspect we should do a beta-backport for this (another reason to keep this PR simple).

Should we also do a 1.78.1?  I suspect that 1.79 is close enough that it won't be worth it.

weihanglo (approved):
Yes we should do a beta backport

PR #13993, opened 2024-05-31
refactor(source): Split `RecursivePathSource` out of `PathSource`

epage:

`PathSource` serves a couple of roles
- When a dependency / patch uses `path` (non-recursive)
- As the implementation details of a `git` source (recursive)
- Dependency overrides (recursive)

Instead of using a `PathSource::new` vs `PathSouce::new_recursive`, this does `RecursivePathSource::new`.  This makes the intent a lot clearer and makes it easier to customize the behavior to each role that is played.

Specifically, there are two ways I expect to leverage this refactor
- Improve the interplay between `RecursivePathSource` and `read_packages` to reduce the duplicate package warnings for git sources (#13992)
- cargo script will change the semantics of path sources slightly and I'm assuming having a distinct source will make this easier

### DIFF
--- a/src/cargo/core/workspace.rs
+++ b/src/cargo/core/workspace.rs
@@ -1142,8 +1142,7 @@ impl<'gctx> Workspace<'gctx> {
                 MaybePackage::Package(ref p) => p.clone(),
                 MaybePackage::Virtual(_) => continue,
             };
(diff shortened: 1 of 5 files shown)


ehuss:
Thanks!

@bors r+

PR #14000, opened 2024-06-01
Fix: Skip deserialization of unrelated fields with overlapping name

linyihai:

Split from https://github.com/rust-lang/cargo/pull/13687#discussion_r1622694446

This fixes the overlap of environment variable names:

> For example, when env_key is UNSTABLE_GITOXIDE_FETCH
and field_key is UNSTABLE_GIT, the field shouldn't be
added because `unstable.gitoxide.fetch` doesn't
belong to `unstable.git` struct.

Updates of test cases `struct_with_overlapping_inner_struct_and_defaults` can be used to compare changes before and after changes

### Additional information
r? weihanglo and very appreciate your more optimized code


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #14008, opened 2024-06-04
Cargo Tree: New format string parameter

timbess:
Add new format parameter to allow querying what platform each dependency is built for.

[rules_rust is using](https://github.com/bazelbuild/rules_rust/blob/e3f6258b1e1660652ae044f01c6dac074f9fb206/crate_universe/src/metadata.rs#L530-L558) `cargo tree` to query the dependency graph and create an equivalent internal graph of dependencies per architecture it's configured to build for. The issue I observed with their current implementation is that they were querying all build dependencies when querying for each target architecture, which caused host architecture only transitive build deps to be added to graph of the target architecture.

I tried querying with just `build` edges, but then we don't get to see the `normal` transitive deps under the `build` deps so we can't actually build a dependency graph of all the build deps accurately using filtering. Next I went to look through the format parameters and saw there wasn't really a way for me to extract what platform it's building for, hence this PR.

### DIFF
--- a/src/cargo/ops/tree/format/mod.rs
+++ b/src/cargo/ops/tree/format/mod.rs
@@ -2,6 +2,8 @@ use std::fmt;
 
 use anyhow::{bail, Error};
 
+use crate::core::dependency::DepKind;
+
 use self::parse::{Parser, RawChunk};
 use super::{Graph, Node};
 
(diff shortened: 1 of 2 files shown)


epage:
While I appreciate this is a draft, we should likely reserve any conversations about "what is the problem", "what the the possible solutions", and "which solution do we want" to an issue.

timbess:
@epage I can copy this to an Issue. I was just filling out the template Github presented when I opened the PR.

PR #14014, opened 2024-06-05
docs(contrib): Suggest atomic commits with separate test commits

epage:

This came up in a discussion with @torhovland.

I'm unsure what the right weight is to put behind these suggestions
- I don't want a lack of git history editing experience to get in the way of someone contributing
- If people do anything, I'd like to see the tests split out

### DIFF
--- a/src/doc/contrib/src/process/working-on-cargo.md
+++ b/src/doc/contrib/src/process/working-on-cargo.md
@@ -124,6 +124,21 @@ typically finish in under 30 minutes.
 The reviewer might point out changes deemed necessary. Large or tricky changes
 may require several passes of review and changes.
 
+> **tip:** Prefer atomic commits where each commit is a single, complete, and coherent unit of work.
+> For example, if your feature work leads to renaming a module, make the rename its own commit.
+> However, adding an internal function that is unused is not complete or coherent.
+>
+> As part of your atomic commits, prefer adding tests as their own commit *before* any functionality changes.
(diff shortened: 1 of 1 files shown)


src/doc/contrib/src/process/working-on-cargo.md
```diff
@@ -124,6 +124,16 @@ typically finish in under 30 minutes.
 The reviewer might point out changes deemed necessary. Large or tricky changes
 may require several passes of review and changes.
 
+> **tip:** Prefer atomic commits where each commit is a single, complete, and coherent unit of work.
+> For example, if your feature work leads to renaming a module, make the rename its own commit.
+> However, adding an internal function that is unused is not complete of coherent.
+>
+> As part of your atomic commits, prefer adding tests as their own commit *before* any functionality changes.
```
weihanglo, line 131: Could we have an example PR here?

Some people write a failing test first, not a test passing the CI.

epage: Added examples and tried to be clearer that we want the tests passing on each commit.

weihanglo (approved):
It looks reasonable as a tip not an enforcement.
Also it is something we can copy over and link to when communicating with contributors.

epage:
Went ahead and remove draft status.  Started it as one to be very intentional about merging this in but based on @weihanglo's comment, seems like this is good to go forward.

weihanglo:
@bors r+

Thanks!

Issue #14016, opened 2024-06-05
Bad error message when a strong dep features's depedency is an unused optional dependency


epage:
### Problem

If nothing can activate an optional dependency, cargo acts as if the dependency doesn't exist which creates poor error messages.

Also, with how things are arranged, the unused optional dependency lint doesn't get reported which could at least reduce the burden on the error message.

### Steps

Baseline `Cargo.toml`:
```toml
cargo-features = ["edition2024"]
[package]
name = "cargo-14010"
version = "0.1.0"
edition = "2024"

[dependencies]
serde = { version = "1.0.203", optional = true }
```
```console
$ cargo +nightly check -Zcargo-lints
warning: unused optional dependency
 --> Cargo.toml:8:1
  |
8 | serde = { version = "1.0.203", optional = true }
  | -----
  |
  = note: `cargo::unused_optional_dependency` is set to `warn` by default
  = help: remove the dependency or activate it in a feature with `dep:serde`
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.00s
```

Add the following to `Cargo.toml`:
```toml

[features]
serde = ["serde/derive"]
```
```console
$ cargo +nightly check -Zcargo-lints
error: failed to parse manifest at `/home/epage/src/personal/dump/cargo-14010-1/Cargo.toml`

Caused by:
  feature `serde` includes `serde/derive`, but `serde` is not a dependency
```

### Possible Solution(s)

- Improve the error message
- Implicitly create `dep:serde` if `serde/derive` is used
- Deprecate `serde/derive` syntax, removing it in a new edition (#10556)

### Notes

This was split out of #14015 assuming we'll go with a different solution than it

epage:
https://github.com/rust-lang/cargo/issues/14010#issuecomment-2148408439

> This makes me worried about `dep_name/feature_name` syntax on Edition 2024. It now requires also adding `dep:dep_name` at which point it isn't much different than `dep_name?/feature_name`.
> 
> I had missing this part in past discussions (e.g. [this zulip topic](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Weak.20features.20syntax)). It seems like we should have either changed it in Edition 2024 to not need `dep:dep_name` _or_ transition out the syntax according to #10556. I worry that changing the meaning of `dep_name/feature_name` would be too subtle. Maybe we should at least go forward with the deprecation of `dep_name/feature_name` as mentioned in the comments of #10556?

[...]

Issue #14015, opened 2024-06-05
Bad error message when a weak dep features's depedency is an unused optional dependency


epage:
### Problem

If nothing can activate an optional dependency, cargo acts as if the dependency doesn't exist which creates poor error messages.

Also, with how things are arranged, the unused optional dependency lint doesn't get reported which could at least reduce the burden on the error message.

### Steps

Baseline `Cargo.toml`:
```toml
cargo-features = ["edition2024"]
[package]
name = "cargo-14010"
version = "0.1.0"
edition = "2024"

[dependencies]
serde = { version = "1.0.203", optional = true }
```
```console
$ cargo +nightly check -Zcargo-lints
warning: unused optional dependency
 --> Cargo.toml:8:1
  |
8 | serde = { version = "1.0.203", optional = true }
  | -----
  |
  = note: `cargo::unused_optional_dependency` is set to `warn` by default
  = help: remove the dependency or activate it in a feature with `dep:serde`
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.00s
```

Add the following to `Cargo.toml`:
```toml

[features]
serde = ["serde?/derive"]
```
```console
$ cargo +nightly check -Zcargo-lints
error: failed to parse manifest at `/home/epage/src/personal/dump/cargo-14010-1/Cargo.toml`

Caused by:
  feature `serde` includes `serde?/derive`, but `serde` is not a dependency
```

### Possible Solution(s)

- Improve the error message
- Show the lint (which might be more difficult)

linyihai:
@rustbot claim 
I have a unpolish improvement need to try.

PR #14018, opened 2024-06-05
fix(fix): Address problems with implicit -> explicit feature migration

### LINKED ISSUE #14010: 2024 implicit feature migration deletes dependency features
### Problem

Migrating to 2024 causes `foo/feature_name` features to get removed for optional dependencies. Those need to be preserved.

### Steps

Example manifest:

```toml
[package]
name = "foo"
version = "0.1.0"
edition = "2021"

[dependencies]

[features]
serde = ["serde/derive"]
```

Running `cargo fix --edition` on this will result in:

```toml
serde = ["dep:serde"]
```

Note that the `/derive` feature got removed.

### Possible Solution(s)

When there is a `["foo/feature_name"]` syntax in the `[features]` table, and `dep:foo` isn't found anywhere in the `[features]` table, it should be rewritten to be `["dep:foo", "foo/feature_name"]`. Some things to be careful of:

* Adjust all places that `foo/xxx` is found (such as `x = ["foo/a"]` and `y = ["foo/b"]`, both x and y need a "`dep:foo`")
* Should not disturb any other features (like `["foo/feature_name", "bar"] would be rewritten to ["dep:foo", "foo/feature_name", "bar"]`)
* Should avoid duplicates (`["foo/a", "foo/b"]` should not have two `dep:foo` entries added).

### Notes

_No response_

### Version

```text
cargo 1.80.0-nightly (7a6fad098 2024-05-31)
release: 1.80.0-nightly
commit-hash: 7a6fad0984d28c8330974636972aa296b67c4513
commit-date: 2024-05-31
host: aarch64-apple-darwin
libcurl: 8.6.0 (sys:0.4.72+curl-8.6.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 14.5.0 [64-bit]
```

epage:

Within the scope of `cargo fix` there  are two problems
- We wipe out existing feature activations if it has the same name as an optional dependency
- The `Cargo.toml` isn't parseable because the unused optional dependency won't "exist" if just `dep_name/feature_name` is used

Fixes #14010

As for the unused optional dependency not "existing" error,
- #14015 is for improving the message for weak dep features
- #14016 is for re-evaluating how we handle this for strong dep features

Depending on what solution we go with for #14016, we might want to revisit the second migration within this PR.  This is one reason I made the commit separate (in addition to just making it clearer whats happening as this gets into some finer details of features).

### DIFF
--- a/src/cargo/ops/fix.rs
+++ b/src/cargo/ops/fix.rs
@@ -456,18 +456,19 @@ fn add_feature_for_unused_deps(pkg: &Package, parent: &mut dyn toml_edit::TableL
     for dep in manifest.dependencies() {
         let dep_name_in_toml = dep.name_in_toml();
         if dep.is_optional() && !activated_opt_deps.contains(dep_name_in_toml.as_str()) {
-            fixes += 1;
             if let Some(features) = parent
                 .entry("features")
                 .or_insert(toml_edit::table())
                 .as_table_like_mut()
             {
-                features.insert(
-                    dep_name_in_toml.as_str(),
-                    toml_edit::Item::Value(toml_edit::Value::Array(toml_edit::Array::from_iter(
-                        &[format!("dep:{}", dep_name_in_toml)],
-                    ))),
-                );
+                features
+                    .entry(dep_name_in_toml.as_str())
+                    .or_insert_with(|| {
+                        fixes += 1;
+                        toml_edit::Item::Value(toml_edit::Value::Array(
+                            toml_edit::Array::from_iter(&[format!("dep:{}", dep_name_in_toml)]),
+                        ))
+                    });
             }
         }
     }
--- a/src/cargo/ops/fix.rs
+++ b/src/cargo/ops/fix.rs
@@ -285,7 +285,7 @@ fn migrate_manifests(ws: &Workspace<'_>, pkgs: &[&Package]) -> CargoResult<()> {
                 fixes += rename_dep_fields_2024(workspace, "dependencies");
             }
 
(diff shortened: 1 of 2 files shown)


src/cargo/ops/fix.rs
```diff
@@ -461,14 +465,43 @@ fn add_feature_for_unused_deps(pkg: &Package, parent: &mut dyn toml_edit::TableL
...
+                // Ensure `dep:dep_name` is present for `dep_name/feature_name` since `dep:` is the
+                // only way to guarantee an optional dependency is available for use.
+                //
+                // The way we avoid implicitly creating features in Edition2024 is we remove the
+                // dependency from `resolved_toml` if there is no `dep:` syntax as that is the only
+                // syntax that suppresses the creation of the implicit feature.
+                for (feature_name, activations) in features.iter_mut() {
+                    let Some(activations) = activations.as_array_mut() else {
+                        let _ = gctx.shell().warn(format_args!("skipping fix of feature `{feature_name}` in package `{}`: unsupported feature schema", pkg.name()));
+                        continue;
```
epage, line 487: This check was written with rust-lang/rfcs#3416 in mind

That syntax is likely to work on any Edition and if we forget to touch this code, I'd prefer to make that known so people can report it to us and we can fix it rather than silently ignoring it.

The question that remains is to warn or error.  I figured we shouldn't block the user from getting other fixes for something they can't workaround in a reasonable way.

ehuss:
Thanks!

@bors r+

PR #14019, opened 2024-06-06
fix:  improve message for inactive weak optional feature with edition2024

### LINKED ISSUE #14015: Bad error message when a weak dep features's depedency is an unused optional dependency

linyihai:

This will improve the message for inactive weak optional feature with edition2024.

This doesn't distinguish whether the dependency had set, which needs to get the dependency from origin that will make more complex.

Fixes #14015 

one commit add the test, one commit fix and update the test

### DIFF
--- a/tests/testsuite/lints/unused_optional_dependencies.rs
+++ b/tests/testsuite/lints/unused_optional_dependencies.rs
@@ -188,7 +188,7 @@ fn inactive_weak_optional_dep() {
 error: failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
-  feature `feat` includes `dep?/feat`, but `dep` is not a dependency
+  feature `feat` includes `dep?/feat`, activate it in a feature with `dep:dep` if `dep` is an enabled dependency
 ",
         )
         .run();
@@ -221,7 +221,7 @@ Caused by:
 error: failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
-  feature `feat` includes `dep?/feat`, but `dep` is not a dependency
+  feature `feat` includes `dep?/feat`, activate it in a feature with `dep:dep` if `dep` is an enabled dependency
 ",
         )
         .run();
--- a/tests/testsuite/lints/unused_optional_dependencies.rs
+++ b/tests/testsuite/lints/unused_optional_dependencies.rs
@@ -157,3 +157,72 @@ warning: unused optional dependency
         )
         .run();
 }
+
+#[cargo_test(nightly, reason = "edition2024 is not stable")]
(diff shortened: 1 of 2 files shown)


tests/testsuite/lints/unused_optional_dependencies.rs
```diff
@@ -221,7 +221,7 @@ Caused by:
 error: failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
-  feature `feat` includes `dep?/feat`, but `dep` is not a dependency
+  feature `feat` includes `dep?/feat`, activate it in a feature with `dep:dep` if `dep` is an enabled dependency
```
epage, line 224: > if `dep` is an enabled dependency

I feel the wording could be improved here.  To help in coming up with ideas, could you expand on what you are trying to convey

linyihai: Yes, the messge need to include the cases that dependency inactive and dependency missing, `if dep is an enabled dependency` means the former.

https://github.com/rust-lang/cargo/pull/14026 is another PR to distinguish these. And this PR may require more detailed wording to address both cases.

linyihai: I suggest keep this PR to polish the message unless https://github.com/rust-lang/cargo/pull/14026https://github.com/rust-lang/cargo/pull/14026 is merged

linyihai:
Close this in favor of https://github.com/rust-lang/cargo/pull/14026

PR #14024, opened 2024-06-07
fix(lints): Add unknown_lints to lints list

Muscraft:
When working on the linting system, I noticed that `UNKNOWN_LINTS` was not in the list of all lints, when it probably should be.


weihanglo (approved):
Thanks!

We could have a way to automate this.

Muscraft:
I plan on adding something to check that all lints have been added at some point, but I didn't want to block this change on getting that working.

PR #14025, opened 2024-06-07
Add tooling to document lints

Muscraft:
One thing the linting system has been missing is documentation. There is no list of all `cargo`'s lints and their associated documentation, like `clippy` or `rustc` has. To resolve this problem, I wrote a very basic tool that gathers the doc comments for a `Lint` and copies them to a markdown file in the Cargo book. This solution is not perfect but it is a way to work towards a better overall experience.

### DIFF
--- a/crates/xtask-lint-docs/src/main.rs
+++ b/crates/xtask-lint-docs/src/main.rs
@@ -1,0 +1,108 @@
+use cargo::util::command_prelude::{flag, ArgMatchesExt};
+use cargo::util::lints::{Lint, LintLevel};
+use itertools::Itertools;
+use std::fmt::Write;
+use std::path::PathBuf;
+
+fn cli() -> clap::Command {
+    clap::Command::new("xtask-lint-docs").arg(flag("check", "Check that the docs are up-to-date"))
+}
+
+fn main() -> anyhow::Result<()> {
+    let args = cli().get_matches();
+    let check = args.flag("check");
+
+    let mut allow = Vec::new();
+    let mut warn = Vec::new();
+    let mut deny = Vec::new();
+    let mut forbid = Vec::new();
+
+    let mut lint_docs = String::new();
+    for lint in cargo::util::lints::LINTS
+        .iter()
+        .sorted_by_key(|lint| lint.name)
+    {
+        if lint.docs.is_some() {
+            let sectipn = match lint.default_level {
+                LintLevel::Allow => &mut allow,
+                LintLevel::Warn => &mut warn,
+                LintLevel::Deny => &mut deny,
+                LintLevel::Forbid => &mut forbid,
+            };
+            sectipn.push(lint.name);
+            add_lint(lint, &mut lint_docs)?;
+        }
+    }
+
+    let mut buf = String::new();
+    writeln!(buf, "# Lints\n")?;
+    writeln!(
+        buf,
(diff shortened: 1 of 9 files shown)


src/cargo/util/lints.rs
```diff
@@ -677,6 +721,41 @@ fn output_unknown_lints(
     Ok(())
 }
 
+/// Note: This only runs on edition 2024+
+///
```
epage, line 726: I feel this is inaccurate.  This form of unused optional depedencies only exist in 2024+ which is different

crates/xtask-lint-docs/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+
+fn cli() -> clap::Command {
+    clap::Command::new("xtask-lint-docs").arg(flag("check", "Check that the docs are up-to-date"))
+}
+
+fn main() -> anyhow::Result<()> {
+    let args = cli().get_matches();
+    let check = args.flag("check");
+
+    let file = std::fs::read_to_string("src/cargo/util/lints.rs")?;
```
epage, line 59: `lint_docs_path` is relative to `CARGO_MANIFEST_DIR` but this isn't

crates/xtask-lint-docs/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+    let mut allow = LintSection::new(LintLevel::Allow);
+    let mut warn = LintSection::new(LintLevel::Warn);
+    let mut deny = LintSection::new(LintLevel::Deny);
+    let mut forbid = LintSection::new(LintLevel::Forbid);
+    for lint in cargo::util::lints::LINTS {
+        let docs = get_lint_docs(lint.name, &file)
```
epage, line 65: I see no justification for why are we parsing the code for lints.  Its especially odd that we are using cargo-as-a-library to enumerate the lints but then not asking the library for the documentation.

src/cargo/util/lints.rs
Muscraft: I don't think this is inaccurate, as [this lint doesn't run if it is less than edition 2024](https://github.com/rust-lang/cargo/blob/a72ce7d278f35e78a168ba5f270974c9f4f91255/src/cargo/util/lints.rs#L698-L701).

crates/xtask-lint-docs/src/main.rs
Muscraft: I originally used doc comments because that is how you normally comment things in Rust. I switched to having it be a field on `Lint`, and I think it will work fine for now.

src/doc/src/reference/lints.md
```diff
@@ -0,0 +1,115 @@
...
+[RFC #3143]: https://rust-lang.github.io/rfcs/3143-cargo-weak-namespaced-features.html
+[RFC #3491]: https://rust-lang.github.io/rfcs/3491-remove-implicit-features.html
+
+
+## Warn-by-default
+
+These lints are all set to the 'warn' level by default.
+- [`unknown_lints`](#unknown_lints)
+- [`unused_optional_dependency`](#unused_optional_dependency)
+
```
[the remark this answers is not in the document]
Muscraft: I went with all the lint levels at the top, and then all the lints sorted by name. I went this route since we do not have many lints currently, and we can always move to separate files in the future

PR #14026, opened 2024-06-07
fix: improve message for inactive weak optional feature with edition2024 through unused dep collection

linyihai:

Collect the unused dependencies to check whether a weak optional dependency had set. Then we can improve the message when weak optional dependency inactive. 

Fixes https://github.com/rust-lang/cargo/issues/14015

One commit test added, one commit fixed and updated

### Additional information
Part of https://github.com/rust-lang/cargo/issues/14039
- migrate `tests/testsuite/lints/unused_optional_dependencies.rs` to snapshot 

And rename `MissingField` to `MissingFieldError`

### DIFF
--- a/src/cargo/core/summary.rs
+++ b/src/cargo/core/summary.rs
@@ -30,6 +30,35 @@ struct Inner {
     rust_version: Option<RustVersion>,
 }
 
+/// Indicates the dependency inferred from the `dep` syntax that should exist,
+/// but missing on the resolved dependencies tables.
+#[derive(Debug)]
+pub struct MissingDependencyError {
+    pub dep_name: InternedString,
+    pub feature: InternedString,
+    pub feature_value: FeatureValue,
+    /// Indicates the dependency inferred from the `dep?` syntax that is weak optional
+    pub weak_optional: bool,
+}
+
+impl std::error::Error for MissingDependencyError {}
+
+impl fmt::Display for MissingDependencyError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let Self {
+            dep_name,
+            feature,
+            feature_value: fv,
+            ..
+        } = self;
+
+        write!(
+            f,
+            "feature `{feature}` includes `{fv}`, but `{dep_name}` is not a dependency",
+        )
+    }
+}
+
 impl Summary {
     #[tracing::instrument(skip_all)]
     pub fn new(
@@ -274,7 +303,12 @@ fn build_feature_map(
 
                     // Validation of the feature name will be performed in the resolver.
                     if !is_any_dep {
-                        bail!("feature `{feature}` includes `{fv}`, but `{dep_name}` is not a dependency");
+                        bail!(MissingDependencyError {
+                            feature: *feature,
+                            feature_value: (*fv).clone(),
+                            dep_name: *dep_name,
+                            weak_optional: *weak,
+                        })
                     }
                     if *weak && !is_optional_dep {
                         bail!(
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -15,6 +15,7 @@ use cargo_util_schemas::manifest::{RustVersion, StringOrBool};
 use itertools::Itertools;
 use lazycell::LazyCell;
 use pathdiff::diff_paths;
+use toml_edit::ImDocument;
 use url::Url;
 
 use crate::core::compiler::{CompileKind, CompileTarget};
@@ -29,6 +30,7 @@ use crate::core::{GitReference, PackageIdSpec, SourceId, WorkspaceConfig, Worksp
 use crate::sources::{CRATES_IO_INDEX, CRATES_IO_REGISTRY};
 use crate::util::errors::{CargoResult, ManifestError};
 use crate::util::interning::InternedString;
+use crate::util::lints::{get_span, rel_cwd_manifest_path};
 use crate::util::{self, context::ConfigRelativePath, GlobalContext, IntoUrl, OptVersionReq};
 
 mod embedded;
@@ -1459,7 +1461,14 @@ fn to_real_manifest(
         // need to check whether `dep_name` is stripped as unused dependency
         if let Err(ref err) = summary {
             if let Some(missing_dep) = err.downcast_ref::<MissingDependencyError>() {
-                check_weak_optional_dep_unused(&original_toml, missing_dep)?;
+                missing_dep_diagnostic(
+                    missing_dep,
+                    &original_toml,
+                    &document,
+                    &contents,
+                    manifest_file,
+                    gctx,
+                )?;
             }
         }
         summary?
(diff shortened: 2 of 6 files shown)


linyihai:
r? @epage

I'm sorry split the fix into two.  If  https://github.com/rust-lang/cargo/pull/14019 more reasonable and finally merge, this PR can be closed.

tests/testsuite/lints/unused_optional_dependencies.rs
```diff
@@ -221,7 +221,7 @@ Caused by:
 error: failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
-  feature `feat` includes `dep?/feat`, but `dep` is not a dependency
+  feature `feat` includes `dep?/feat`, activate it in a feature with `dep:dep`
```
epage, line 224: This never finishing saying whats wrong but jumps straight to a suggestion.

e.g.
```
feature `feat` includes `dep?/feat` but no feature activates `dep`.
To activate `dep`, a feature must specify `dep:dep`.
```
```
`dep?/feat` (in `feat`) activates `feat` when `dep` is activated but no feature activates `dep`.
To activate `dep`, a feature must specify `dep:dep`.
```
```
feature `feat` includes `dep?/feat` but no feature activates `dep` with `dep:dep`.
```

tests/testsuite/lints/unused_optional_dependencies.rs
```diff
@@ -157,3 +157,72 @@ warning: unused optional dependency
...
+            cargo-features = ["edition2024"]
+            [package]
+            name = "foo"
+            version = "0.1.0"
+            edition = "2024"
+
+            [dependencies]
+
+            [features]
+            feat = ["dep?/feat"]
```
epage, line 177: - `dep` doesn't have the `feature` feature
- We should clarify the names so people are less likely to be confused  / make the test output clearer
  - `dep` -> `dep_name`
  - `dep`s `feature` -> `dep_feature`
  - `foo`s `feature -> `foo_feature`

epage:
@Muscraft would be interested in your thoughts on this approach to solving #14015.

#14019 is an attempt but it lacks the information for a clear message, so this tries to track the additional information to make that happen.

At least one potential ramification of this is we could simplified `unused optional dependency` lint to just checking the `unused_dependencies` tables

I do think #14028 should be merged first so we make sure we don't have regressions.

epage:
(also I should note that I've only looked at the high level details and not all of the nitty gritty of the implementation yet)

tests/testsuite/lints/unused_optional_dependencies.rs
linyihai: Oops， my local commit dosn't push to here, but it is similar to this
> feature `feat` includes `dep?/feat` but no feature activates `dep` with `dep:dep`.

linyihai:
Try to minimize code changes and performance impact,  which will return a custom Error `MissingDependency` when `Summary::new?` found a dependency missing.  And then handle the `MissingDependency` in `to_real_manifest`.

linyihai:
The suggestions above have been considered. 

Then in the subsequent commit, add these alternation by the way.
-  Rename `MissingField` to `MissingFieldError`
-  Migrate `unused_optional_dependencies` to snapshot
The changes are small and easy to review :)

src/cargo/util/toml/mod.rs
```diff
@@ -1435,24 +1436,53 @@ fn to_real_manifest(
...
+                    for deps in orig_deps {
+                        if let Some(deps) = deps {
+                            if deps.keys().any(|p| *p.as_str() == missing_dep.dep_name()) {
+                                missing_dep.set_unused_dependency(true);
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
weihanglo, line 1483: To me, it is a bit odd to mutate an error. Could we extract this login to a function, which accepts the `original_toml` and the `err` as arguments and returns a refined error?

By doing that, `MissingDependencyError` doesn't need the field `unused_dependency`. We could do the logic complete outside `core::summary`.`Summary` is used by both index file parsing and manifest parsing, but the missing dep feature error doesn't make sense for index parsing at this moment. Therefore, `MissingDependencyError` will emit only one kind of error message.

Also, we could make all fields `pub` because the error is not intended to be modified.

weihanglo (commented):
Have talked to @Muscraft and we think the current mechanism of this is fine!

In the future, we might want to make a dedicated `Error` type (probably an enum) for `Summary::new()` through `thiserror` crate, so we don't need to `downcast` them anymore. The `Error` type could additionally be handled in `cargo::util::toml` with original manifest TOML to get an enhanced error with richer information. 

The future version would be something like this:

```rust
fn error_with_original_toml(err: &SummaryError, original_toml: &TomlManifest) -> CargoResult<()> {
    match err {
        SmmmaryError::Foo => return Err(...),
        SmmmaryError::BarError => return Err(...),
        SummaryError::MissingDependency(missing_dep) => {
		    let message = "...";
            if let Err(err) = gctx.shell().print_message(message) {
            	return Err(err.into());
            }
            return Err(AlreadyPrintedError::new(anyhow!("").into()).into());
        }
    }
}
```

which might be rendered to something like:


image

or

image


Credit to @Muscraft doing this experiment. See <https://github.com/Muscraft/cargo/commit/c0b44e06a213bd0d680111f2f386d68504a60af5> for the code that generate the snapshots above.

src/cargo/core/summary.rs
```diff
@@ -30,6 +30,55 @@ struct Inner {
...
+#[derive(Debug)]
+pub struct MissingDependencyError {
+    dep_name: InternedString,
+    feature: InternedString,
+    feature_value: FeatureValue,
+    // Indicates the dependency inferred from the `dep?` syntax that is weak optional
+    weak_optional: bool,
+    // Indicates the dependency is unused but not absent in the manifest
+    unused_dependency: bool,
+}
```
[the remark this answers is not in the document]
linyihai: No problem.

src/cargo/util/toml/mod.rs
linyihai: Thanks for your suggestions. 

All have been rebased into [fix: improve message for inactive weak optional feature with edition2024](https://github.com/rust-lang/cargo/pull/14026/commits/1306429464e8182d987cf4c4715a162b04e10df2)

linyihai:
> Credit to @Muscraft doing this experiment. See https://github.com/Muscraft/cargo/commit/c0b44e06a213bd0d680111f2f386d68504a60af5 for the code that generate the snapshots above.

I have added the `missing_dep_diagnostic` into the last commit with tiny twist, thanks @Muscraft.

These tips are more user-friendly.

src/cargo/util/toml/mod.rs
linyihai: And the last commit has added the `missing_dep_diagnostic`,  but I shamless copy the code, may be it can be improved in latter PR.

weihanglo (approved):

weihanglo:
Thanks! That looks awesome. Will go ahead and merge this.

@bors r+

PR #14027, opened 2024-06-07
fix(env): `cargo:rerun-if-env-changed` doesn't work with env configuration

### LINKED ISSUE #10358: `cargo:rerun-if-env-changed` doesn't work with `env` configuration
### Problem

When a `build.rs` script emits `cargo:rerun-if-env-changed`, it is not re-run when the value of the specified variable is changed via the `env` configuration.

### Steps

`build.rs`:
```rust
pub fn main() {
    println!("cargo:rerun-if-env-changed=FOO");
    if let Ok(foo) = std::env::var("FOO") {
        assert!(&foo != "bad");
    }
}
```

`.cargo/config.toml`:
```toml
[env]
FOO = "good"
```

1. Run `cargo build`, it succeeds.
2. Change `.cargo/config.toml` to this:
```toml
[env]
FOO = "bad"
```
3. Run `cargo build` again, and it uses the cached build and succeeds. It should run `build.rs` again, and it should fail.

# Workaround

4. Run `FOO=bar cargo build`, ignore the result.
5. Run `cargo build` again, now it runs `build.rs` with `FOO=bad` from the config and fails as expected.

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
cargo 1.58.0 (7f08ace4f 2021-11-24)
release: 1.58.0
commit-hash: 7f08ace4f1305de7f3b1b0e2f765911957226bd4
commit-date: 2021-11-24
libcurl: 7.80.0-DEV (sys:0.4.51+curl-7.80.0 vendored ssl:OpenSSL/1.1.1l)
os: Debian 11 (bullseye) [64-bit]
```

heisen-li:

Fixes #10358


src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1695,6 +1720,7 @@ fn local_fingerprints_deps(
     deps: &BuildDeps,
     target_root: &Path,
     pkg_root: &Path,
+    envs: BTreeMap<String, Option<OsString>>,
```
epage, line 1723: Why do we pass in an owned `BTreeMap` rather than a borrowed?

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -588,6 +590,10 @@ impl TargetInfo {
             .iter()
             .any(|sup| sup.as_str() == split.as_str())
     }
+
+    pub fn get_target_envs(&self) -> CargoResult<&BTreeMap<String, Option<OsString>>> {
+        return Ok(self.crate_type_process.get_envs());
+    }
 }
```
epage, line 597: This feels very brittle because any change to how we pass in environment variables automatically gets used as  fingerprinting, rather than intentionally choosing what we fingerprint.  Just to check to see if this is correct is requiring checking several other places.

In fact, I think this will cause every build to rebuild everything when using jobserver because I think we are capturing the env after that is configured and that env includes file descriptors.

src/cargo/core/compiler/fingerprint/mod.rs
heisen-li: done

src/cargo/core/compiler/build_context/target_info.rs
epage: This was overlooking that we aren't fingerprinting all of it, only what the user requests.  So maybe its not all that of a problem?

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -772,10 +774,21 @@ impl LocalFingerprint {
...
         let var = key.to_owned();
-        let val = env::var(key).ok();
+
+        let val = envs
+            .get(key)
+            .and_then(|opt| {
+                opt.as_ref()
+                    .and_then(|os_str| os_str.clone().into_string().ok())
+            })
+            .or_else(|| env::var(key).ok());
```
epage, line 790: I find it strange to fallback to `env::var` if the `envs[key]` is non-UTF8.  Maybe we should call `env::var_os` and deal with `into_string` at the end.

epage, line 790: iiuc on Windows, `env::var` calls `GetEnvironmentVariableW` which is case insensitive.  Since we're recreating our own environment look up, should this matter?

heisen-li: To be honest, I didn't consider using `env::var_os` instead of `env::var` because I wanted to minimize the impact of the change and avoid affecting other behaviors. If there is an up-to-date value in the current environment variable, then use envs.get(key); otherwise the previous `env::var(key)` is still used.

It looks like the only difference between `env::var` and `env::var_os` is whether the environment variable is a valid Unicode.

heisen-li: I don't have a more in-depth study of windows, I'm very sorry. 

Do you mean that you shouldn't use `env::var` or `env::var_os` to query environment variables? That doesn't seem to work : see: [rerun_if_env_changes](https://github.com/rust-lang/cargo/pull/14027/files#diff-bec1ea82298c9855dced8a66db57ca380f385e70254ca1cd3b8f99152465854dR55) and [rerun_if_env_or_file_changes](https://github.com/rust-lang/cargo/pull/14027/files#diff-bec1ea82298c9855dced8a66db57ca380f385e70254ca1cd3b8f99152465854dR109)

PR #14028, opened 2024-06-07
test(lints): Ensure unused optional dep fires for shadowed dep

epage:
This is a way to have an unused optional dependency before 2024 edition.
In prior editions, it is currently a hard error.
I'm tempted to switch that hard error to this lint in prior editions but at minimum, we need to make sure we don't make changes that cause this to revert back to a hard error on 2024+


ehuss:
Thanks!

@bors r+

PR #14030, opened 2024-06-07
Keep lints updated

Muscraft:
In #14024, [it was noted](https://github.com/rust-lang/cargo/pull/14024#pullrequestreview-2103598233) that we should probably have some automated way to ensure that `LINTS` stays up to date. I went ahead and did this, but I extended the idea to also include `LINT_GROUPS` and added support for ensuring they both stay sorted.

Error message for `LINTS` not being sorted:
Screenshot from 2024-06-07 10-18-43

Error message for missing `Lint`s:
Screenshot from 2024-06-07 10-51-09

### DIFF
--- a/src/cargo/util/lints.rs
+++ b/src/cargo/util/lints.rs
@@ -14,8 +14,8 @@ use toml_edit::ImDocument;
 
 const LINT_GROUPS: &[LintGroup] = &[TEST_DUMMY_UNSTABLE];
 pub const LINTS: &[Lint] = &[
-    IM_A_TEAPOT,
     IMPLICIT_FEATURES,
+    IM_A_TEAPOT,
     UNKNOWN_LINTS,
     UNUSED_OPTIONAL_DEPENDENCY,
 ];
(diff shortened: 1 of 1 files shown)


src/cargo/util/lints.rs
```diff
@@ -875,3 +875,45 @@ pub fn unused_dependencies(
...
+        let actual = super::LINTS
+            .iter()
+            .map(|l| l.name.to_uppercase())
+            .collect::<Vec<_>>();
+
+        let mut expected = actual.clone();
+        expected.sort();
+        let expected = format!("{:#?}", expected);
+        let actual = format!("{:#?}", actual);
+        snapbox::assert_data_eq!(actual, expected);
```
epage, line 899: nit: `snapbox` has a `ToDebug::to_debug()` so this can be `snapbox::assert_data_eq!(actual.to_debug(), expected.to_debug())`

src/cargo/util/lints.rs
```diff
@@ -916,4 +918,78 @@ mod tests {
...
+
+    #[test]
+    fn ensure_updated_lints() {
+        let path = file!();
+        let expected = std::fs::read_to_string(path).unwrap();
+        let expected = expected
+            .lines()
+            .filter_map(|l| {
+                if l.ends_with(": Lint = Lint {") {
+                    Some(
```
epage, line 930: I worry about this *but* if its a problem, we can always remove it later

src/cargo/util/lints.rs
```diff
@@ -916,4 +918,78 @@ mod tests {
         let actual = format!("{:#?}", actual);
         snapbox::assert_data_eq!(actual, expected);
     }
+
+    #[test]
+    fn ensure_updated_lints() {
+        let path = file!();
```
epage, line 924: nit: unlikely to be a problem for `cargo` but `file!` doesn't always work which is what `snapbox::utils::current_rs!()` is for

PR #14040, opened 2024-06-10
Bump to 0.82.0; update changelog

weihanglo:
[rendered](https://github.com/weihanglo/cargo/blob/version-bump/CHANGELOG.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,22 +1,177 @@
 # Changelog
 
 ## Cargo 1.80 (2024-07-25)
-[b60a1555...HEAD](https://github.com/rust-lang/cargo/compare/b60a1555...HEAD)
+[b60a1555...rust-1.80.0](https://github.com/rust-lang/cargo/compare/b60a1555...rust-1.80.0)
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


ehuss:
Thanks!

@bors r+

CHANGELOG.md
```diff
@@ -1,23 +1,235 @@
...
+- 🎉 cargo-update: Allows `--precise` to specify a yanked version of a package,
+  and will update the lockfile accordingly.
+  [#13974](https://github.com/rust-lang/cargo/pull/13974)
+
 ### Changed
 
+- ❗️ manifest: Disallow `[badges]` to inherit from `[workspace.package.badges]`.
+  This was considered a bug and should inherit from `[workspace.badges]`.
+  Keep in mind that `[badges]` is effectively deprecated.
+  [#13788](https://github.com/rust-lang/cargo/pull/13788)
```
epage, line 99: > Even if someone were to try to guess to use this, it is inconsistent with how inheritance works because this should inherit from workspace.badges instead of workspace.package.badges

I think that line from the PR was misinterpreted.  It meant to say "the pattern for how to inherit would have us use `workspace.badges`.  This does not mean that `workspace.badges` is supported

Issue #14039, opened 2024-06-10
Port from bespoke assertions of snapbox


epage:
Traditionally, [`cargo-test-support`](https://doc.rust-lang.org/nightly/nightly-rustc/cargo_test_support/index.html) has had bespoke assertions, requiring hand implemented diff algorithms.  As there is less of a community around this, the knowledge is more specialized, its less likely to be documented, and we are on our own for feature development.

In #13980 and #14031, we introduced the use of [snapbox](https://crates.io/crates/snapbox) as replacement for our own assertions and need to work to migrate to them.

Aside: doc updates related to this transition
- #14243
- #14266
- #14269
- #14268
- #14270
- #14272

### 1. Select a file to port

The files should have `#![allow(deprecated)]` at the top, e.g.
```console
$ git pull --rebase
$ rg '#!.allow.deprecated' tests
```

Please check the comments on this issue for anyone to have claimed the file and then post that you claim the file

### 2. Remove `#![allow(deprecated)]` to identify what work is needed

If nothing, congrats, that was easy!

### 3. Resolve basic deprecations

Replace `Execs::with_stdout("...")` with
```rust
use cargo_test_support::prelude::*;
use cargo_test_support::str;

.with_stdout_data(str![])
```

Replace `Execs::with_stderr("...")` with
```rust
use cargo_test_support::prelude::*;
use cargo_test_support::str;

.with_stderr_data(str![])
```
(prelude is only needed for some steps)

> *Side note:* If the test is specifically wanting to assert an empty string, use `""` rather than `str![]` so that people are less likely to accidentally do a snapshot update to bless output that gets added.

Commit these changes

Run `SNAPSHOTS=overwrite cargo test && cargo check --test testsuite`.  In rare cases, `SNAPSHOTS=overwrite` may make bad edits.  Feel free to create a reproduction case and [create an issue](https://github.com/assert-rs/snapbox/issues/new/choose).  They are usually obvious how to fix, so don't worry.

Diff the snapshots.  Is there anything machine or run specific in them that previously was elided out with `[..]`?
- Can you auto-redact it like 5ea1c8f?  If so, revert the snapshot updates, add the auto-redaction (as a separate commit before any test changes), and re-run the snapshots
- Otherwise, replace it with `[..]`

Once its working, amend your commit with the snapshots

### 4. Resolve non-literal deprecations

Like Step 3, but not just `with_stdout("...")` but with variables or `format!`.

Evaluate whether a literal could be used
[...]

PR #14041, opened 2024-06-11
test: migrate binary_name to snapbox

weihanglo:

Part of <https://github.com/rust-lang/cargo/issues/14039>.

Migrate `tests/testsuite/binary_name.rs` to snapbox.

Two things need to beware of

* A new redaction is added: `[ELAPSED]` for libtest output.
* Snapbox's [`JsonLines`](https://docs.rs/snapbox/0.6.10/snapbox/data/enum.DataFormat.html#variant.JsonLines) doesn't seems to support pretty printed JSON.

### DIFF
--- a/tests/testsuite/binary_name.rs
+++ b/tests/testsuite/binary_name.rs
@@ -1,9 +1,11 @@
-#![allow(deprecated)]
+//! Tests for `cargo-features = ["different-binary-name"]`.
 
-use cargo_test_support::install::{
-    assert_has_installed_exe, assert_has_not_installed_exe, cargo_home,
-};
+use cargo_test_support::install::assert_has_installed_exe;
+use cargo_test_support::install::assert_has_not_installed_exe;
+use cargo_test_support::install::cargo_home;
+use cargo_test_support::prelude::*;
 use cargo_test_support::project;
+use cargo_test_support::str;
 
 #[cargo_test]
 fn gated() {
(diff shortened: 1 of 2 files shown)


tests/testsuite/binary_name.rs
```diff
@@ -284,25 +312,13 @@ fn check_msg_format_json() {
...
-
-{"reason":"build-finished", "success":true}
-"#;
-
     // Run cargo build.
     p.cargo("build --message-format=json")
         .masquerade_as_nightly_cargo(&["different-binary-name"])
-        .with_json(output)
+        .with_stdout_data(str![[r#"
+{"executable":"[ROOT]/foo/target/debug/007bar[EXE]","features":[],"filenames":"{...}","fresh":false,"manifest_path":"[ROOT]/foo/Cargo.toml","package_id":"path+[ROOTURL]/foo#0.0.1","profile":"{...}","reason":"compiler-artifact","target":"{...}"}
```
weihanglo, line 319: This is something worth fixing in snapbox. We need a pretty-printed JSON lines support.

epage: Technically, this is no longer jsonlines.  It looks like cargo-test-support had a baked in policy of taking `expected` and using double-newlines to delimit pretty-printed json.

One trick I can look into is we could parse a json array and then convert that to jsonlines.

epage:
@bors r+

Checked in with @weihanglo  on jsonlines

> It's a regression when reviewing those outputs. We can track it in the issue and revisit then

tests/testsuite/binary_name.rs
weihanglo: Yeah I am aware of the double-newlines trick in cargo-test-support. I don't consider it as a block for this test file. For test suites like `tests/testsuite/message_format.rs` we should consider postponing the migrations.

PR #14044, opened 2024-06-11
test: migrate build_script_env to snapbox

heisen-li:

part of https://github.com/rust-lang/cargo/issues/14039.

### DIFF
--- a/tests/testsuite/build_script_env.rs
+++ b/tests/testsuite/build_script_env.rs
@@ -1,10 +1,9 @@
 //! Tests for build.rs rerun-if-env-changed and rustc-env
 
-#![allow(deprecated)]
-
 use cargo_test_support::basic_manifest;
 use cargo_test_support::project;
 use cargo_test_support::sleep_ms;
+use cargo_test_support::str;
 
 #[cargo_test]
 fn rerun_if_env_changes() {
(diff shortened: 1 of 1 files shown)


tests/testsuite/build_script_env.rs
```diff
@@ -171,17 +186,28 @@ fn rustc_bootstrap() {
...
+[ERROR] Cannot set `RUSTC_BOOTSTRAP=1` from build script of `foo v0.0.1 ([ROOT]/foo)`.
+[NOTE] Crates cannot set `RUSTC_BOOTSTRAP` themselves, as doing so would subvert the stability guarantees of Rust for your project.
+[HELP] If you're sure you want to do this in your project, set the environment variable `RUSTC_BOOTSTRAP=1` before running cargo instead.
+
+"#]])
         .with_status(101)
         .run();
 }
 
+#[allow(deprecated)]
```
heisen-li, line 210: If the following test cases are all expanded, the content will be very large. I don't think of a better solution, maybe the current treatment is the most appropriate.

epage, line 210: Highlighting this for better tracking

tests/testsuite/build_script_env.rs
```diff
@@ -237,6 +264,7 @@ fn build_script_sees_cfg_target_feature() {
 /// what the value of `RUSTFLAGS` should be in this case. We chose to give a
 /// warning. However, no matter what we do, it's important that build scripts
 /// and rustc see a consistent picture
+#[allow(deprecated)]
```
epage, line 267: Highlighting this for better tracking

tests/testsuite/build_script_env.rs
```diff
@@ -198,6 +224,7 @@ fn build_script_env_verbose() {
         .run();
 }
 
+#[allow(deprecated)]
```
epage, line 227: Highlighting this for better tracking

tests/testsuite/build_script_env.rs
```diff
@@ -298,6 +326,7 @@ fn cfg_paradox() {
 /// ```
 /// CARGO_CFG_TARGET_THREAD_LOCAL=
 /// ```
+#[allow(deprecated)]
```
epage, line 329: Highlighting this for better tracking

tests/testsuite/build_script_env.rs
heisen-li: Sorry, I don't know how to emphasize this to make it more obvious (does that mean add notes?) . Can you go into more detail or examples?

epage: We had a race condition.  I was posting at the same time as you.  What you did was what I would recommend: start discussions on topics that you know reviewers need to look at.

tests/testsuite/build_script_env.rs
epage: Yes, for `-vv` output, we should probably still do "contains" checks as mentioned in the instructions

So
```rust
    p.cargo("check -vv")
        .with_stderr_contains("[WARNING] non-trivial mutual dependency between target-specific configuration and RUSTFLAGS")
        .with_stderr_contains("[foo 0.0.1] cfg!(bertrand)=true")
        .with_stderr_contains("[..]--cfg=bertrand[..]")
        .run();
```
could be
```rust
    p.cargo("check -vv")
        .with_stderr_contains("\
...
[WARNING] non-trivial mutual dependency between target-specific configuration and RUSTFLAGS
...")
        .with_stderr_contains("\
...
[foo 0.0.1] cfg!(bertrand)=true
...")
        .with_stderr_contains("\
...
[..]--cfg=bertrand[..]
...")
        .run();
```
`...` is a multi-line glob.  `str![]` isn't as important in these cases as we can't really update snapshots like these.

epage:
Thanks!

@bors r+

PR #14048, opened 2024-06-11
Migrate a few test files to snapbox

Muscraft:
This migrates the following files to `snapbox`
- `artifact_dep`
  - Has a few `does_not_contain`
- `artifact_dir`
- `bad_config`
- `bad_manifest_path`
  - Does not use `str!` for all tests  
- `bench`

Note: This also adds auto-redactions for:
- `[HOST_TARGET]`
- `[ALT_TARGET]` 
  - Only added if cross-compilation is allowed for the target
- `[AVG_ELAPSED]`
  - For `bench` output 
- `[JITTER]`
  - For `bench` output

Part of #14039

### DIFF
--- a/tests/testsuite/artifact_dep.rs
+++ b/tests/testsuite/artifact_dep.rs
@@ -1,9 +1,8 @@
 //! Tests specific to artifact dependencies, designated using
 //! the new `dep = { artifact = "bin", … }` syntax in manifests.
 
-#![allow(deprecated)]
-
 use cargo_test_support::compare::assert_e2e;
+use cargo_test_support::prelude::*;
 use cargo_test_support::registry::{Package, RegistryBuilder};
 use cargo_test_support::str;
 use cargo_test_support::{
@@ -35,14 +34,13 @@ fn check_with_invalid_artifact_dependency() {
         .build();
     p.cargo("check -Z bindeps")
         .masquerade_as_nightly_cargo(&["bindeps"])
-        .with_stderr(
-            "\
-[ERROR] failed to parse manifest at `[..]/Cargo.toml`
+        .with_stderr_data(str![[r#"
+[ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
   'unknown' is not a valid artifact specifier
-",
-        )
+
+"#]])
         .with_status(101)
         .run();
 
@@ -79,14 +77,13 @@ Caused by:
         .build();
     run_cargo_with_and_without_bindeps_feature(&p, "check", &|cargo| {
         cargo
-            .with_stderr(
-                "\
-[ERROR] failed to parse manifest at `[..]/Cargo.toml`
+            .with_stderr_data(str![[r#"
+[ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
   'lib' specifier cannot be used without an 'artifact = …' value (bar)
-",
-            )
+
+"#]])
             .with_status(101)
             .run();
     });
@@ -112,14 +109,13 @@ Caused by:
         .build();
     run_cargo_with_and_without_bindeps_feature(&p, "check", &|cargo| {
         cargo
-            .with_stderr(
-                "\
-[ERROR] failed to parse manifest at `[..]/Cargo.toml`
+            .with_stderr_data(str![[r#"
+[ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
   'target' specifier cannot be used without an 'artifact = …' value (bar)
-",
-            )
+
+"#]])
             .with_status(101)
             .run();
     })
@@ -149,9 +145,16 @@ fn check_with_invalid_target_triple() {
         .build();
     p.cargo("check -Z bindeps")
         .masquerade_as_nightly_cargo(&["bindeps"])
-        .with_stderr_contains(
-            r#"[..]Could not find specification for target "unknown-target-triple"[..]"#,
-        )
+        .with_stderr_data(str![[r#"
+[ERROR] failed to run `rustc` to learn about target-specific information
+
+Caused by:
+  process didn't exit successfully: `rustc - --crate-name ___ --print=file-names --target unknown-target-triple [..]` ([EXIT_STATUS]: 1)
+  --- stderr
+  [ERROR] Error loading target specification: Could not find specification for target "unknown-target-triple". Run `rustc --print target-list` for a list of built-in targets
+
+
+"#]])
         .with_status(101)
         .run();
 }
@@ -179,14 +182,13 @@ fn build_without_nightly_aborts_with_error() {
         .build();
     p.cargo("check")
         .with_status(101)
-        .with_stderr(
-            "\
-[ERROR] failed to parse manifest at [..]
+        .with_stderr_data(str![[r#"
+[ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
   `artifact = …` requires `-Z bindeps` (bar)
-",
-        )
+
+"#]])
         .run();
 }
 
@@ -215,11 +217,12 @@ fn disallow_artifact_and_no_artifact_dep_to_same_package_within_the_same_dep_cat
     p.cargo("check -Z bindeps")
         .masquerade_as_nightly_cargo(&["bindeps"])
         .with_status(101)
-        .with_stderr("\
+        .with_stderr_data(str![[r#"
 [LOCKING] 2 packages to latest compatible versions
-[WARNING] foo v0.0.0 ([CWD]) ignoring invalid dependency `bar_stable` which is missing a lib target
-[ERROR] the crate `foo v0.0.0 ([CWD])` depends on crate `bar v0.5.0 ([CWD]/bar)` multiple times with different names",
-        )
+[WARNING] foo v0.0.0 ([ROOT]/foo) ignoring invalid dependency `bar_stable` which is missing a lib target
+[ERROR] the crate `foo v0.0.0 ([ROOT]/foo)` depends on crate `bar v0.5.0 ([ROOT]/foo/bar)` multiple times with different names
+
+"#]])
         .run();
 }
 
@@ -324,15 +327,14 @@ fn features_are_unified_among_lib_and_bin_dep_of_same_target() {
 
     p.cargo("build -Z bindeps")
         .masquerade_as_nightly_cargo(&["bindeps"])
-        .with_stderr(
-            "\
+        .with_stderr_data(str![[r#"
 [LOCKING] 3 packages to latest compatible versions
-[COMPILING] d2 v0.0.1 ([CWD]/d2)
-[COMPILING] d1 v0.0.1 ([CWD]/d1)
-[COMPILING] foo v0.0.1 ([CWD])
-[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [..]
-",
-        )
+[COMPILING] d2 v0.0.1 ([ROOT]/foo/d2)
+[COMPILING] d1 v0.0.1 ([ROOT]/foo/d1)
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
         .run();
 }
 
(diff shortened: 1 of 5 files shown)


crates/cargo-test-support/src/cross_compile.rs
```diff
@@ -209,18 +209,27 @@ pub fn native_arch() -> &'static str {
 ///
 /// Only use this function on tests that check `cross_compile::disabled`.
 pub fn alternate() -> &'static str {
+    if let Some(alt) = try_alternate() {
+        alt
+    } else {
+        panic!("This test should be gated on cross_compile::disabled.");
+    }
```
epage, line 216: nit: this could be `try_alternate().expect(...)`

tests/testsuite/artifact_dep.rs
```diff
@@ -1087,26 +1156,32 @@ fn build_script_deps_adopt_specified_target_unconditionally() {
         .file("bar/src/lib.rs", "pub fn doit() {}")
         .build();
 
+    #[allow(deprecated)]
```
epage, line 1159: highlighting for discussion

tests/testsuite/artifact_dep.rs
```diff
@@ -1206,21 +1281,23 @@ fn non_build_script_deps_adopt_specified_target_unconditionally() {
...
```
epage, line 1284: highlighting for discussion

tests/testsuite/artifact_dep.rs
```diff
@@ -1351,26 +1430,26 @@ fn build_script_deps_adopts_target_platform_if_target_equals_target() {
         .build();
 
     let alternate_target = cross_compile::alternate();
+    #[allow(deprecated)]
```
epage, line 1433: highlighting for discussion

tests/testsuite/artifact_dep.rs
```diff
@@ -436,9 +438,32 @@ fn features_are_not_unified_among_lib_and_bin_dep_of_different_target() {
...
+  |
+[NOTE] found an item that was configured out
+ --> [ROOT]/foo/d2/src/lib.rs:3:49
+  |
+3 |                 #[cfg(feature = "d2f2")] pub fn f2() {}
+  |                                                 ^^
+  = [NOTE] the item is gated behind the `d2f2` feature
+
+For more information about this error, try `rustc --explain E0425`.
+[ERROR] could not compile `d1` (bin "d1") due to 1 previous error
```
epage, line 464: Should we use `...` to avoid having rustc output in the snapshot?  That seems to cause frustrations in making rustc changed

tests/testsuite/artifact_dep.rs
```diff
@@ -673,7 +703,22 @@ fn build_script_with_bin_artifact_and_lib_false() {
     p.cargo("build -Z bindeps")
         .masquerade_as_nightly_cargo(&["bindeps"])
         .with_status(101)
-        .with_stderr_does_not_contain("[..]sentinel[..]")
+        // This should not contain "sentinel"
+        .with_stderr_data(str![[r#"
```
epage, line 707: I'll want to look closer at this as comments could be missed in snapshot updates

tests/testsuite/artifact_dep.rs
```diff
@@ -673,7 +703,22 @@ fn build_script_with_bin_artifact_and_lib_false() {
...
+[LOCKING] 2 packages to latest compatible versions
+[COMPILING] bar v0.5.0 ([ROOT]/foo/bar)
+[COMPILING] foo v0.0.0 ([ROOT]/foo)
+error[E0433]: failed to resolve: use of undeclared crate or module `bar`
+ --> build.rs:3:16
+  |
+3 |                bar::doit()
+  |                ^^^ use of undeclared crate or module `bar`
```
epage, line 715: Also, pulling in rustc errors is not idea

tests/testsuite/artifact_dep.rs
```diff
@@ -715,7 +760,21 @@ fn lib_with_bin_artifact_and_lib_false() {
     p.cargo("build -Z bindeps")
         .masquerade_as_nightly_cargo(&["bindeps"])
         .with_status(101)
-        .with_stderr_does_not_contain("[..]sentinel[..]")
+        // This should not contain "sentinel"
+        .with_stderr_data(str![[r#"
```
epage, line 764: ditto

crates/cargo-test-support/src/compare.rs
```diff
@@ -108,6 +109,20 @@ pub fn assert_ui() -> snapbox::Assert {
         regex::Regex::new("home/\\.cargo/registry/src/-(?<redacted>[a-z0-9]+)").unwrap(),
     )
     .unwrap();
+    subs.insert("[TARGET]", rustc_host()).unwrap();
```
weihanglo, line 112: I forgot why we separate redactions into UI and e2e. Can/Should we combine them?

tests/testsuite/artifact_dep.rs
```diff
@@ -1267,17 +1344,15 @@ fn no_cross_doctests_works_with_artifacts() {
...
 [LOCKING] 2 packages to latest compatible versions
-[COMPILING] bar v0.5.0 ([CWD]/bar)
-[COMPILING] foo v0.0.1 ([CWD])
-[FINISHED] `test` profile [unoptimized + debuginfo] target(s) in [..]
-[RUNNING] [..] (target/{triple}/debug/deps/foo-[..][EXE])
+[COMPILING] bar v0.5.0 ([ROOT]/foo/bar)
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+[FINISHED] `test` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+[RUNNING] unittests src/lib.rs (target/[TARGET]/debug/deps/foo-d7d4d3e3948ead58)
```
epage, line 1352: Is that hash going to cause a problem?

tests/testsuite/artifact_dep.rs
```diff
@@ -2085,17 +2192,16 @@ fn env_vars_and_build_products_for_various_build_targets() {
...
-[COMPILING] bar [..]
-[COMPILING] foo [..]
-[FINISHED] `test` profile [unoptimized + debuginfo] target(s) in [..]
-[RUNNING] unittests [..]
-[RUNNING] tests/main.rs [..]
+[COMPILING] bar v0.5.0 ([ROOT]/foo/bar)
+[COMPILING] foo v0.0.0 ([ROOT]/foo)
+[FINISHED] `test` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+[RUNNING] unittests src/lib.rs (target/debug/deps/foo-[..])
+[RUNNING] tests/main.rs (target/debug/deps/main-[..])
```
epage, line 2201: Should we try to have an auto-redaction for these?

tests/testsuite/artifact_dir.rs
```diff
@@ -1,8 +1,8 @@
 //! Tests for --artifact-dir flag.
 
-#![allow(deprecated)]
-
+use cargo_test_support::prelude::*;
```
epage, line 3: Note to self: I stopped on this commit

tests/testsuite/artifact_dep.rs
Muscraft: Thanks for catching this, I thought I caught all of them

tests/testsuite/artifact_dep.rs
Muscraft: I can add back in explicit `.with_stderr_does_not_contain` if you would like

crates/cargo-test-support/src/compare.rs
weihanglo, line 112: what about this?

```suggestion
    subs.insert("[HOST]", rustc_host()).unwrap();
```

tests/testsuite/artifact_dep.rs
Muscraft: Switched to `...`

tests/testsuite/artifact_dep.rs
Muscraft: Switched to `...`

crates/cargo-test-support/src/compare.rs
epage: If i see `HOST`, I think something else. `HOST_TARGET`?

epage: e2e has more. Maybe when this is done we can reevaluate that but rather not do it during the port.

As for duplicated ones, thats something worth reevaluating at this point.  The first had to have the regex specialized between the two (because i like to keep regexes as simple as possible) and that set the example to follow. We have enough now to be worth reevaluating.

crates/cargo-test-support/src/compare.rs
```diff
@@ -185,6 +185,10 @@ static MIN_LITERAL_REDACTIONS: &[(&str, &str)] = &[
     ("[EXE]", std::env::consts::EXE_SUFFIX),
     ("[BROKEN_PIPE]", "Broken pipe (os error 32)"),
     ("[BROKEN_PIPE]", "The pipe is being closed. (os error 232)"),
+    // Unix message for exit status
+    ("[EXIT_STATUS]", "exit status"),
+    // Windows message for exit status
+    ("[EXIT_STATUS]", "exit code"),
```
epage, line 196: There is a risk these are too general but we can always revisit it later

crates/cargo-test-support/src/compare.rs
```diff
@@ -165,6 +165,11 @@ fn add_common_redactions(subs: &mut snapbox::Redactions) {
         regex::Regex::new("home/\\.cargo/registry/src/-(?<redacted>[a-z0-9]+)").unwrap(),
     )
     .unwrap();
+    subs.insert(
+        "[HASH]",
+        regex::Regex::new(r"[a-z0-9\-_]+-(?<redacted>[0-9a-f]{16})").unwrap(),
+    )
+    .unwrap();
```
epage, line 172: Is this too general?

tests/testsuite/artifact_dep.rs
epage: Let's do that

tests/testsuite/artifact_dep.rs
```diff
@@ -1206,21 +1260,23 @@ fn non_build_script_deps_adopt_specified_target_unconditionally() {
...
-        ))
-        .with_stderr_contains(format!(
-            "[RUNNING] `rustc --crate-name bar --edition=2015 bar/src/main.rs [..]--target {} [..]",
-            target
-        ))
         .with_stderr_does_not_contain(format!(
             "[RUNNING] `rustc --crate-name foo [..]--target {} [..]",
             target
         ))
-        .with_stderr_contains("[RUNNING] `rustc --crate-name foo [..]")
```
epage, line 1223: I feel like these are easier to contrast with the previous way

tests/testsuite/bench.rs
```diff
@@ -1663,31 +2111,13 @@ fn json_artifact_includes_executable_for_benchmark() {
...
-                }
+        .with_stdout_data(
+            str![[r#"
+{"executable":"[ROOT]/foo/target/release/deps/benchmark-[HASH][EXE]","features":[],"filenames":"{...}","fresh":false,"manifest_path":"[ROOT]/foo/Cargo.toml","package_id":"path+[ROOTURL]/foo#0.0.1","profile":"{...}","reason":"compiler-artifact","target":{"crate_types":["bin"],"doc":false,"doctest":false,"edition":"2015","kind":["bench"],"name":"benchmark","src_path":"[ROOT]/foo/benches/benchmark.rs","test":false}}
+{"reason":"build-finished","success":true}
 
-                {"reason": "build-finished", "success": true}
-            "#,
+"#]]
+            .json_lines(),
```
epage, line 2120: Like with #14041, we'll need to revisit this later

crates/cargo-test-support/src/compare.rs
Muscraft: Taken care of in #14054

tests/testsuite/artifact_dep.rs
Muscraft: Taken care of in #14054

crates/cargo-test-support/src/compare.rs
Muscraft: The common redactions were merged into one function in #14054 so that `assert_ui` and `assert_e2e` could share them

tests/testsuite/artifact_dep.rs
```diff
@@ -670,10 +685,24 @@ fn build_script_with_bin_artifact_and_lib_false() {
...
+error[E0433]: failed to resolve: use of undeclared crate or module `bar`
+...
+
+For more information about this error, try `rustc --explain E0433`.
+[ERROR] could not compile `foo` (build script) due to 1 previous error
+...
+
+"#]])
         .run();
 }
```
epage, line 707: If I'm reading this right, we added new output coverage we didn't have before.  What was the motivation?

epage: (ditto for other sentinel cases)

Muscraft: I believe I did this so we could see that the actual was while still keeping the `does_not_contain` separate. I went a head and removed it.

PR #14049, opened 2024-06-12
More `update --breaking` tests

torhovland:
Related to https://github.com/rust-lang/cargo/issues/12425#issuecomment-2161374345 in #12425.

### DIFF
--- a/src/cargo/ops/cargo_update.rs
+++ b/src/cargo/ops/cargo_update.rs
@@ -348,7 +348,7 @@ fn upgrade_dependency(
         return Ok(dependency);
     }
 
-    let Some(new_req_string) = upgrade_requirement(&current.to_string(), latest)? else {
+    let Some((new_req_string, _)) = upgrade_requirement(&current.to_string(), latest)? else {
         trace!("skipping dependency `{name}` because the version requirement didn't change");
         return Ok(dependency);
     };
@@ -369,8 +369,17 @@ fn upgrade_dependency(
     Ok(dep)
 }
 
-/// Update manifests with upgraded versions, and write to disk. Based on cargo-edit.
-/// Returns true if any file has changed.
+/// Update manifests with upgraded versions, and write to disk. Based on
+/// cargo-edit. Returns true if any file has changed.
+///
+/// Some of the checks here are duplicating checks already done in
+/// upgrade_manifests/upgrade_dependency. Why? Let's say upgrade_dependency has
+/// found that dependency foo was eligible for an upgrade. But foo can occur in
+/// multiple manifest files, and even multiple times in the same manifest file,
+/// and may be pinned, renamed, etc. in some of the instances. So we still need
+/// to check here which dependencies to actually modify. So why not drop the
+/// upgrade map and redo all checks here? Because then we'd have to query the
+/// registries again to find the latest versions.
 pub fn write_manifest_upgrades(
     ws: &Workspace<'_>,
     upgrades: &UpgradeMap,
@@ -407,6 +416,11 @@ pub fn write_manifest_upgrades(
                 )?;
                 let name = &dependency.name;
 
+                if let Some(renamed_to) = dependency.rename {
+                    trace!("skipping dependency renamed from `{name}` to `{renamed_to}`");
+                    continue;
+                }
+
                 let Some(current) = dependency.version() else {
                     trace!("skipping dependency without a version: {name}");
                     continue;
@@ -424,12 +438,26 @@ pub fn write_manifest_upgrades(
                     continue;
                 };
 
-                let Some(new_req_string) = upgrade_requirement(current, latest)? else {
+                let Some((new_req_string, new_req)) = upgrade_requirement(current, latest)? else {
                     trace!(
                         "skipping dependency `{name}` because the version requirement didn't change"
                     );
                     continue;
                 };
+
+                let [comparator] = &new_req.comparators[..] else {
+                    trace!(
+                        "skipping dependency `{}` with multiple version comparators: {:?}",
+                        name,
+                        new_req.comparators
+                    );
+                    continue;
+                };
+
+                if comparator.op != Op::Caret {
+                    trace!("skipping non-caret dependency `{}`: {}", name, comparator);
+                    continue;
+                }
 
                 let mut dep = dependency.clone();
                 let mut source = source.clone();
--- a/src/cargo/ops/cargo_update.rs
+++ b/src/cargo/ops/cargo_update.rs
@@ -263,46 +263,37 @@ fn upgrade_dependency(
     let renamed_to = dependency.name_in_toml();
 
     if name != renamed_to {
-        trace!(
-            "skipping dependency renamed from `{}` to `{}`",
-            name,
-            renamed_to
-        );
+        trace!("skipping dependency renamed from `{name}` to `{renamed_to}`");
         return Ok(dependency);
     }
 
     if !to_update.is_empty() && !to_update.contains(&name.to_string()) {
-        trace!("skipping dependency `{}` not selected for upgrading", name);
+        trace!("skipping dependency `{name}` not selected for upgrading");
         return Ok(dependency);
     }
 
     if !dependency.source_id().is_registry() {
-        trace!("skipping non-registry dependency: {}", name);
+        trace!("skipping non-registry dependency: {name}");
         return Ok(dependency);
     }
 
     let version_req = dependency.version_req();
 
     let OptVersionReq::Req(current) = version_req else {
-        trace!(
-            "skipping dependency `{}` without a simple version requirement: {}",
-            name,
-            version_req
-        );
+        trace!("skipping dependency `{name}` without a simple version requirement: {version_req}");
         return Ok(dependency);
     };
 
     let [comparator] = &current.comparators[..] else {
         trace!(
-            "skipping dependency `{}` with multiple version comparators: {:?}",
-            name,
+            "skipping dependency `{name}` with multiple version comparators: {:?}",
             &current.comparators
         );
         return Ok(dependency);
     };
 
     if comparator.op != Op::Caret {
-        trace!("skipping non-caret dependency `{}`: {}", name, comparator);
+        trace!("skipping non-caret dependency `{name}`: {comparator}");
         return Ok(dependency);
     }
 
@@ -332,30 +323,21 @@ fn upgrade_dependency(
     };
 
     let Some(latest) = latest else {
-        trace!(
-            "skipping dependency `{}` without any published versions",
-            name
-        );
+        trace!("skipping dependency `{name}` without any published versions");
         return Ok(dependency);
     };
 
     if current.matches(&latest) {
-        trace!(
-            "skipping dependency `{}` without a breaking update available",
-            name
-        );
+        trace!("skipping dependency `{name}` without a breaking update available");
         return Ok(dependency);
     }
 
     let Some(new_req_string) = upgrade_requirement(&current.to_string(), latest)? else {
-        trace!(
-            "skipping dependency `{}` because the version requirement didn't change",
-            name
-        );
+        trace!("skipping dependency `{name}` because the version requirement didn't change");
         return Ok(dependency);
     };
 
-    let upgrade_message = format!("{} {} -> {}", name, current, new_req_string);
+    let upgrade_message = format!("{name} {current} -> {new_req_string}");
     trace!(upgrade_message);
 
     if upgrade_messages.insert(upgrade_message.clone()) {
@@ -389,10 +371,7 @@ pub fn write_manifest_upgrades(
         .collect::<Vec<_>>();
 
     for manifest_path in manifest_paths {
-        trace!(
-            "updating TOML manifest at `{:?}` with upgraded dependencies",
-            manifest_path
-        );
+        trace!("updating TOML manifest at `{manifest_path:?}` with upgraded dependencies");
 
         let crate_root = manifest_path
             .parent()
@@ -410,31 +389,28 @@ pub fn write_manifest_upgrades(
                     dep_key_str,
                     dep_item,
                 )?;
+                let name = &dependency.name;
 
                 let Some(current) = dependency.version() else {
-                    trace!("skipping dependency without a version: {}", dependency.name);
+                    trace!("skipping dependency without a version: {name}");
                     continue;
                 };
 
                 let (MaybeWorkspace::Other(source_id), Some(Source::Registry(source))) =
                     (dependency.source_id(ws.gctx())?, dependency.source())
                 else {
-                    trace!("skipping non-registry dependency: {}", dependency.name);
+                    trace!("skipping non-registry dependency: {name}");
                     continue;
                 };
 
-                let Some(latest) = upgrades.get(&(dependency.name.to_owned(), source_id)) else {
-                    trace!(
-                        "skipping dependency without an upgrade: {}",
-                        dependency.name
-                    );
+                let Some(latest) = upgrades.get(&(name.to_owned(), source_id)) else {
+                    trace!("skipping dependency without an upgrade: {name}");
                     continue;
                 };
 
                 let Some(new_req_string) = upgrade_requirement(current, latest)? else {
                     trace!(
-                        "skipping dependency `{}` because the version requirement didn't change",
-                        dependency.name
+                        "skipping dependency `{name}` because the version requirement didn't change"
                     );
                     continue;
                 };
@@ -444,7 +420,7 @@ pub fn write_manifest_upgrades(
                 source.version = new_req_string;
                 dep.source = Some(Source::Registry(source));
 
-                trace!("upgrading dependency {}", dependency.name);
+                trace!("upgrading dependency {name}");
                 dep.update_toml(&crate_root, &mut dep_key, dep_item);
                 manifest_has_changed = true;
                 any_file_has_changed = true;
--- a/src/cargo/ops/cargo_update.rs
+++ b/src/cargo/ops/cargo_update.rs
@@ -222,6 +222,11 @@ pub fn upgrade_manifests(
     let mut upgrades = HashMap::new();
     let mut upgrade_messages = HashSet::new();
 
+    let to_update = to_update
+        .iter()
+        .map(|s| PackageIdSpec::parse(s))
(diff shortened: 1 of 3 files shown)


torhovland:
r? epage

tests/testsuite/update.rs
```diff
@@ -2096,3 +2095,273 @@ fn update_breaking_specific_packages() {
...
+        str![[r#"
+
+                [package]
+                name = "foo"
+                version = "0.0.1"
+                edition = "2015"
+                authors = []
+
+                [dependencies]
+                renamed-to = { package = "renamed-from", version = "2.0" }
```
torhovland, line 2349: @epage Hmm, this is a bug, then, isn't it?

tests/testsuite/update.rs
```diff
@@ -2096,3 +2095,324 @@ fn update_breaking_specific_packages() {
...
+
+    Package::new("compatible", "1.0.1").publish();
+    Package::new("incompatible", "1.0.1").publish();
+
+    Package::new("incompatible", "2.0.0").publish();
+
+    p.cargo("update -Zunstable-options --breaking incompatible@1.0.0")
+        .masquerade_as_nightly_cargo(&["update-breaking"])
+        .with_stderr(
+            // FIXME: This is wrong.
```
torhovland, line 2250: I found another problem. We're not matching `incompatible` with `incompatible@1.0.0`.

torhovland: But maybe that's not a bug. Strictly speaking, `1.0.0` does not match the version requirement `1.0`.

torhovland: Anyway, I pushed a fix.

tests/testsuite/update.rs
epage: Yes

tests/testsuite/update.rs
epage: Think of this feature as a specially named `--force` flag. They identified the entry they want to update so we should.

What we also need a test for is if they specify a compatible version that isn't in the lockfile.

src/cargo/ops/cargo_update.rs
```diff
@@ -271,7 +276,14 @@ fn upgrade_dependency(
         return Ok(dependency);
     }
 
-    if !to_update.is_empty() && !to_update.contains(&name.to_string()) {
+    if !to_update.is_empty()
+        && !to_update.iter().any(|spec| {
+            spec.name() == name.as_str()
+                && spec
```
epage, line 282: We should check the spec has a registry source and if its for the  same registry.

tests/testsuite/update.rs
```diff
@@ -2096,3 +2095,341 @@ fn update_breaking_specific_packages() {
...
+    Package::new("renamed-from", "1.0.1").publish();
+    Package::new("non-semver", "1.0.1").publish();
+    Package::new("transitive-compatible", "1.0.1").publish();
+    Package::new("transitive-incompatible", "1.0.1").publish();
+
+    Package::new("renamed-from", "2.0.0").publish();
+    Package::new("non-semver", "2.0.0").publish();
+    Package::new("transitive-incompatible", "2.0.0").publish();
+
+    p.cargo("update -Zunstable-options --breaking compatible renamed-from non-semver transitive-compatible transitive-incompatible")
```
epage, line 2145: Should this fail?

I lean towards "yes"
- During testing, it will be easier to get feedback on an unwanted failure
- Compatibility wise, its easier to stop erroring than to start

We can split fixing this into a separate PR if you want.

tests/testsuite/update.rs
torhovland: I'm actually not using the lock file. This should work whether or not there is a lock file. I'm just checking if the version matches the version req.

epage: Thats an implementation detail while I'm talking of user-facing semantics. We should match `spec` behavior with `cargo update` without `--breaking`.

tests/testsuite/update.rs
torhovland: I'm making a FIXME comment. Let's fix this in a separate PR.

tests/testsuite/update.rs
torhovland: Imagine a non-existing `compatible@1.0.10`. When run without a lock file, `cargo update compatible@1.0.10` does not complain. Then, if the same command is run again (now with a lock file), it complains about `compatible@1.0.10` not matching any packages. So it is not idempotent.

Is it OK if `cargo update --breaking compatible@1.0.10` also doesn't complain, whether or not there is a lock file?

If not OK, we'll have to query a `previous_resolve` just for this error message. And it might be a good idea to go back to reusing the `ops::update_lockfile` that the non-breaking update is running (but then having to make changes to `keep`). In any case, can we put that in a separate PR?

tests/testsuite/update.rs
torhovland: Yes, let's do a separate PR.

src/cargo/ops/cargo_update.rs
torhovland: Good point. Fixed.

torhovland:
@epage I found another bug, I think. See the latest commit: https://github.com/rust-lang/cargo/pull/14049/commits/901f78e225219af9c29c5f0c10a73b4a81e32924

epage:
If I'm understanding, thats a question of whether non-workspace path dependencies should be affected by this (and whether we should error if nothing matched).

I think this makes a good case for erroring. Our mental model is this is a specialized force, so it makes sense the lockfile can update when others can't if we error.

As for local deps like this, we should make this a task in the tracking issue to verify the general approach of cargo wrt mutations (cargo add, cargo fix, etc).

torhovland:
Looks like this is actually consistent. We can not run `cargo add` on non-ws crates. When running `cargo fix` only ws members are being fixed. And `cargo-edit` also does not make upgrades in non-ws crates.

torhovland:
@epage I consider this PR ready on my part now. Awaiting further review comments.

tests/testsuite/update.rs
torhovland: Never mind, I added the fix here.

weihanglo:
While #14115 is not about bumping SemVer-breaking versions, this behavior described in <https://github.com/rust-lang/cargo/issues/14115#issuecomment-2181617556> could be a thing to consider supporting or adding a test: `cargo update <pkg> --breaking` shouldn't touch other unrelated packages from the same source.

tests/testsuite/update.rs
```diff
@@ -2096,3 +2095,466 @@ fn update_breaking_specific_packages() {
...
+[UPGRADING] mixed-compatibility ^1.0 -> ^2.0
+[LOCKING] 1 package to latest compatible version
+[ADDING] mixed-compatibility v2.0.1
+",
+        )
+        .run();
+}
+
+#[cargo_test]
+/// In order to handle cases like this, it's not enough to keep an upgrade map
```
epage, line 2457: Stylistically, I feel like documentation should go before attributes

tests/testsuite/update.rs
```diff
@@ -2096,3 +2095,466 @@ fn update_breaking_specific_packages() {
...
+
+    p.cargo("generate-lockfile").run();
+
+    Package::new("dep", "1.1.1").publish();
+    Package::new("dep", "2.0.0").publish();
+
+    // Will upgrade the direct dependency
+    p.cargo("update -Zunstable-options --breaking dep@1.0")
+        .masquerade_as_nightly_cargo(&["update-breaking"])
+        // FIXME: Should upgrade a dependency here.
```
epage: Let's have a single place to track remaining work, your task list, rather than having multiple backlogs, with one being harder to track (the source) than the other (issues)

src/cargo/ops/cargo_update.rs
```diff
@@ -410,41 +426,62 @@ pub fn write_manifest_upgrades(
...
+                let name = &dependency.name;
+
+                if let Some(renamed_to) = dependency.rename {
+                    trace!(
+                        "skipping dependency renamed from `{}` to `{}`",
+                        name,
+                        renamed_to
+                    );
+                    continue;
+                }
```
epage, line 438: nit: where it works, please prefer inline interpolation

```suggestion
                if let Some(renamed_to) = dependency.rename {
                    trace!(
                        "skipping dependency renamed from `{name}` to `{renamed_to}`"
                    );
                    continue;
                }
```
(ditto with `dependency.name` -> `name` changes)

src/cargo/ops/cargo_update.rs
```diff
@@ -426,41 +426,62 @@ pub fn write_manifest_upgrades(
                     dep_key_str,
                     dep_item,
                 )?;
+                let name = &dependency.name;
```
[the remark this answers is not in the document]
torhovland: Done.

tests/testsuite/update.rs
```diff
@@ -1950,6 +1989,7 @@ fn update_breaking() {
                 yanked  =  "1.0"  # Comment
                 ws.workspace  =  true  # Comment
                 shared  =  "2.0"  # Comment
+                multiple-locations  =  { path  =  "../multiple-locations", version  =  "1.0" }  # Comment
```
torhovland, line 1992: @epage I'm wondering how you feel about this version not getting upgraded.

tests/testsuite/update.rs
torhovland: Moved docs elsewhere (to implementation).

tests/testsuite/update.rs
torhovland: These FIXMEs were just intermediate ones, and are gone in the final PR.

But task list added here: https://github.com/rust-lang/cargo/issues/12425#issuecomment-2186198258

tests/testsuite/update.rs
```diff
@@ -2096,3 +2095,466 @@ fn update_breaking_specific_packages() {
...
+[ADDING] mixed-compatibility v2.0.1
+",
+        )
+        .run();
+}
+
+#[cargo_test]
+/// In order to handle cases like this, it's not enough to keep an upgrade map
+/// keyed on dependency name and source, we also need to know which manifest
+/// files to update and which to keep unchanged, because they may be pinned or
```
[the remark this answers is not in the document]
torhovland: I removed that part of the comment, moved the rest (updated) to the implementation, and added a test case for mixed pinning within the same manifest.

src/cargo/ops/cargo_update.rs

src/cargo/ops/cargo_update.rs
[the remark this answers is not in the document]
torhovland: Fixed.

torhovland:
Does anybody understand the Clippy error? The file in question has had `#![allow(deprecated)]` the whole time, and the Clippy command has worked fine until now and still works on my machine.

tests/testsuite/update.rs
torhovland: @epage This may or may not be a task to put on the task list, but it depends on your answer to the question about `cargo update --breaking compatible@1.0.10` not complaining.

tests/testsuite/update.rs
torhovland: Done. See https://github.com/rust-lang/cargo/issues/12425#issuecomment-2186198258

torhovland:
> While #14115 is not about bumping SemVer-breaking versions, this behavior described in [#14115 (comment)](https://github.com/rust-lang/cargo/issues/14115#issuecomment-2181617556) could be a thing to consider supporting or adding a test: `cargo update <pkg> --breaking` shouldn't touch other unrelated packages from the same source.

I agree about adding a test. But I feel this particular case is so specific to #14115 that I suggest any test cases for breaking as well as non-breaking updates go there.

tests/testsuite/update.rs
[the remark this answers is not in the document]
epage: The comment removal was done in a later commit.  Should it be done in the initial commit?

tests/testsuite/update.rs
epage: Leaving this alone is the correct behavior.  Cargo will error out if the `version` req does not match the version at `path`.

tests/testsuite/update.rs
```diff
@@ -2064,3 +2112,508 @@ fn update_breaking_specific_packages() {
...
+[ERROR] expected a version like "1.32"
+
+"#]])
+        .run();
+
+    // Spec version not matching our current dependencies
+    p.cargo("update -Zunstable-options --breaking incompatible@2.0.0")
+        .masquerade_as_nightly_cargo(&["update-breaking"])
+        .with_stderr_data(str![[r#""#]])
+        .run();
```
epage, line 2275: If I run `cargo update clap@5.0.0` in the cargo repo, it fails.  This should do similar.  This can be added to the task list.

tests/testsuite/update.rs
```diff
@@ -2064,3 +2112,508 @@ fn update_breaking_specific_packages() {
...
+
+    // Spec source not matching our current dependencies
+    p.cargo("update -Zunstable-options --breaking https://alternative.com#incompatible@1.0.0")
+        .masquerade_as_nightly_cargo(&["update-breaking"])
+        .with_stderr_data(str![[r#""#]])
+        .run();
```
epage, line 2281: ditto

tests/testsuite/update.rs
```diff
@@ -2064,3 +2112,508 @@ fn update_breaking_specific_packages() {
...
+    // Non-existing versions
+    p.cargo("update -Zunstable-options --breaking incompatible@9.0.0")
+        .masquerade_as_nightly_cargo(&["update-breaking"])
+        .with_stderr_data(str![[r#""#]])
+        .run();
+
+    p.cargo("update -Zunstable-options --breaking compatible@9.0.0")
+        .masquerade_as_nightly_cargo(&["update-breaking"])
+        .with_stderr_data(str![[r#""#]])
+        .run();
```
epage, line 2326: ditto

tests/testsuite/update.rs
```diff
@@ -2064,3 +2112,508 @@ fn update_breaking_specific_packages() {
...
+
+"#]])
+        .run();
+
+    let root_manifest = p.read_file("Cargo.toml");
+    let foo_manifest = p.read_file("foo/Cargo.toml");
+    let bar_manifest = p.read_file("bar/Cargo.toml");
+    let baz_manifest = p.read_file("baz/Cargo.toml");
+
+    assert_e2e().eq(
```
epage, line 2556: nit: Personally, I'd read next to use to keep knowledge as local as possible

tests/testsuite/update.rs
```diff
@@ -2064,3 +2112,508 @@ fn update_breaking_specific_packages() {
...
+[ADDING] mixed-ws-pinned v2.0.0
+[ADDING] renamed-from v2.0.0
+
+"#]])
+        .run();
+
+    let root_manifest = p.read_file("Cargo.toml");
+    let foo_manifest = p.read_file("foo/Cargo.toml");
+    let bar_manifest = p.read_file("bar/Cargo.toml");
+    let baz_manifest = p.read_file("baz/Cargo.toml");
```
epage, line 2554: Would it work to give these workspace members names tied to their role in the test?  For example, they can be "pinned", "unpinned", "mixed-pinned".  That'd make it easier to follow what the test is doing

tests/testsuite/update.rs
epage: Let's put it in the task list either way and not block this conversation further.

tests/testsuite/update.rs
epage: Tracked elsewhere

tests/testsuite/update.rs
epage: Tracked elsewhere

tests/testsuite/update.rs
epage: Tracked elsewhere

tests/testsuite/update.rs
torhovland: By the way, this gets improved in #14140, and becomes:

```
        .with_stderr_data(str![[r#"
Did you mean one of these?

  incompatible@1.0.0

"#]])
```

tests/testsuite/update.rs

tests/testsuite/update.rs

tests/testsuite/update.rs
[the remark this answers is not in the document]

tests/testsuite/update.rs
torhovland: FYI, this is changed in #14140. If there isn't a lock file, `update -Zunstable-options --breaking incompatible@9.0.0` will do this:

```
[ADDING] incompatible v1.0.1 (latest: v2.0.0)
```

If there is a lock file, it will do this:

```
[ERROR] package ID specification `incompatible@9.0.0` did not match any packages
Did you mean one of these?

  incompatible@1.0.0
```

torhovland:
@epage Ready for review again.

torhovland:
@epage Sorry, my rebasing on master probably made the force push diff unreadable for you.

But all I changed was deleting that long comment in the test commit, and applying your suggestions [here](https://github.com/rust-lang/cargo/pull/14049/commits/8882d0504eef97ccc919e1788ad54af644289ac1#diff-7a958bda8ff2a6c02818273dd5fb4c03d79d3c699affd9da3d9a67d8f4eaed30R2436).

epage:
Thanks!

@bors r+

tests/testsuite/update.rs
```diff
@@ -2064,3 +2112,507 @@ fn update_breaking_specific_packages() {
...
+    Package::new("incompatible", "1.0.1").publish();
+
+    Package::new("incompatible", "2.0.0").publish();
+
+    // Invalid spec
+    p.cargo("update -Zunstable-options --breaking incompatible@foo")
+        .masquerade_as_nightly_cargo(&["update-breaking"])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] expected a version like "1.32"
```
epage, line 2266: Hadn't notice this in your prior PR.  This error message is bad.  We should at least be calling out that we are parsing a package id spec

Existing behavior:
```console
$ cargo update clap@foo
error: invalid package ID specification: `clap@foo`

Caused by:
  expected a version like "1.32"
```
(again, task is list is fine)

torhovland: Added.

PR #14051, opened 2024-06-12
test: migrate features_are_quoted to snapbox

henry40408:

Migrate `tests/testsuite/shell_quoting.rs` to snapbox.

N/A

### Additional information

N/A

### DIFF
--- a/tests/testsuite/shell_quoting.rs
+++ b/tests/testsuite/shell_quoting.rs
@@ -2,9 +2,8 @@
 //! in the output, their arguments are quoted properly
 //! so that the command can be run in a terminal.
 
-#![allow(deprecated)]
-
 use cargo_test_support::project;
+use cargo_test_support::str;
 
 #[cargo_test]
 fn features_are_quoted() {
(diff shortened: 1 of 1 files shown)


tests/testsuite/shell_quoting.rs
```diff
@@ -26,14 +25,24 @@ fn features_are_quoted() {
...
+        .with_stderr_data(str![[r#"
+[CHECKING] foo v0.1.0 ([ROOT]/foo)
+[RUNNING] `rustc --crate-name foo --edition=2015 src/main.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type bin --emit=dep-info,metadata -C embed-bitcode=no -C debuginfo=2 --cfg 'feature="default"' --cfg 'feature="some_feature"' -C metadata=c2925e00c1458fcb -C extra-filename=-c2925e00c1458fcb --out-dir [ROOT]/foo/target/debug/deps -L dependency=[ROOT]/foo/target/debug/deps`
+error[E0425]: cannot find value `error` in this scope
+ --> src/main.rs:1:12
+  |
+1 | fn main() {error}
+  |            ^^^^^ not found in this scope
+
+For more information about this error, try `rustc --explain E0425`.
```
weihanglo, line 40: Snapshotting rustc output is a bit risky. If applicable we should omit it by `...`.

```suggestion
...
```

tests/testsuite/shell_quoting.rs
```diff
@@ -26,14 +25,24 @@ fn features_are_quoted() {
...
-            .with_stderr_contains(
-                r#"[RUNNING] `rustc [..] --cfg 'feature="default"' --cfg 'feature="some_feature"' [..]`"#
-            ).with_stderr_contains(
-                r#"
+        .env("MSYSTEM", "1")
+        .with_status(101)
+        .with_stderr_data(str![])
+        .with_stderr_data(str![[r#"
+[CHECKING] foo v0.1.0 ([ROOT]/foo)
+[RUNNING] `rustc --crate-name foo --edition=2015 src/main.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type bin --emit=dep-info,metadata -C embed-bitcode=no -C debuginfo=2 --cfg 'feature="default"' --cfg 'feature="some_feature"' -C metadata=c2925e00c1458fcb -C extra-filename=-c2925e00c1458fcb --out-dir [ROOT]/foo/target/debug/deps -L dependency=[ROOT]/foo/target/debug/deps`
```
weihanglo, line 33: We don't have redactions for hashes (metadata and extra-filename) for now. Let's redact most of those args as before?

```suggestion
[RUNNING] `rustc [..] --cfg 'feature="default"' --cfg 'feature="some_feature"' [..]`
```

tests/testsuite/shell_quoting.rs
```diff
@@ -26,14 +25,24 @@ fn features_are_quoted() {
...
     p.cargo("check -v")
-            .env("MSYSTEM", "1")
-            .with_status(101)
-            .with_stderr_contains(
-                r#"[RUNNING] `rustc [..] --cfg 'feature="default"' --cfg 'feature="some_feature"' [..]`"#
-            ).with_stderr_contains(
-                r#"
+        .env("MSYSTEM", "1")
+        .with_status(101)
+        .with_stderr_data(str![])
```
weihanglo, line 30:
```suggestion
```

tests/testsuite/shell_quoting.rs
weihanglo: By risky I mean rustc output is not guaranteed stable and might break cargo test suites between different toolchain versions.

epage: I've added this to the porting steps

> - If replacing a `contains` with an equality check, use multi-line globs (`...`) for rustc errors that weren't previously matched to minimize tying Cargo's tests to the exact output of rustc

tests/testsuite/shell_quoting.rs
epage: Should we work to address to auto-redact these first?  The downside is delaying @henry40408's contribution.  The upside is we don't have to remember to update these later.

Hmm, maybe this isn't a problem.  If we add auto-redactions later, this will still work.  If/when the output changes, the snapshot updating will take care of it for us.  The only benefit remaining is helping with future test ports and so auto-redacting now is at a "nit" level and I don't see it as a blocker.

weihanglo: I am fine with moving this forward now and come back later

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #14052, opened 2024-06-12
docs(contrib): Improve triage instructions

epage:

@torhovland brought up some confusions they had when looking at our triage instructions and this attempts to fix them.

### DIFF
--- a/src/doc/contrib/src/issues.md
+++ b/src/doc/contrib/src/issues.md
@@ -225,9 +225,9 @@ There are several things to consider when triaging an issue:
     needs to discuss whether or not to proceed and what needs to be done to
     address the issue.
   * [S-needs-mentor] --- This is something the Cargo team wants to address,
-    but does not currently have the capacity to help with reviewing.
+    but does not currently have the capacity to help with reviewing. **(reserved for Cargo team)**
   * [S-accepted] --- This is something that clearly needs to be addressed, and
-    a Cargo team member has volunteered to help review.
+    a Cargo team member has volunteered to help review. **(reserved for Cargo team)**
 
 Anyone is welcome to help with the triaging process. You can help with
 reproducing issues, checking for duplicates, gathering more information from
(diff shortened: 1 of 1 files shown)


src/doc/contrib/src/issues.md
```diff
@@ -225,9 +228,9 @@ There are several things to consider when triaging an issue:
     needs to discuss whether or not to proceed and what needs to be done to
     address the issue.
   * [S-needs-mentor] --- This is something the Cargo team wants to address,
-    but does not currently have the capacity to help with reviewing.
+    but does not currently have the capacity to help with reviewing. **(reserved for Cargo team)**
```
weihanglo, line 231: What about `S-needs-rfc`?

> Anyone is welcome to help at this stage, but it should be clear that it is not yet accepted. However, this should only be tagged for changes that are somewhat likely to be accepted.

Should we include that or adjust its text a bit?

epage: I took out the "likely to be accepted".  I didn't realize that and have been tagging things that are RFC scope.  I did link out to the instructions on socializing an RFC which kind of fills a similar role of pre-vetting.

src/doc/contrib/src/issues.md
```diff
@@ -217,17 +221,17 @@ There are several things to consider when triaging an issue:
 * Assuming the issue looks valid, remove the [S-triage] label and move it onto
```
torhovland, line 221: > remove the [S-triage] label

This is part of what doesn't sit well with me. What if the next natural status is one of the **(reserved for Cargo team)** ones? Isn't removing [S-triage] reserved as well then?

I suggest something like this:

> Assuming the issue looks valid, consider assigning it a new status label:
>
> ... (bullet points)
>
> Remove the [S-triage] label if you moved the issue to a new status.

epage: I tried a slightly different wording

> Assuming the issue looks valid, switch the [S-triage] label for one of the following:

My thinking is that this makes it clearer that this is an "atomic" operation; that you should only remove `S-triage` if you can add the new label.  In practice, if there is no more design work, then its "needs-team-input".

weihanglo: This change looks reasonable. Thanks!

weihanglo (approved):

PR #14054, opened 2024-06-12
Add assert redactions

Muscraft:
This was split out from #14048 so that the test changes in that PR do not block the redactions.

This adds auto-redactions for:
- A new `[HASH]` for `/<file>-<16 char hash>`
- `[HOST_TARGET]`
- `[ALT_TARGET]` 
  - Only added if cross-compilation is allowed for the target
- `[AVG_ELAPSED]`
  - For `bench` output 
- `[JITTER]`
  - For `bench` output

This also moves all common redactions to a function that `assert_e2e` and `assert_ui` call to reduce the amount of duplicate code and makes it so we only compile regex redactions once.

### DIFF
--- a/crates/cargo-test-support/src/compare.rs
+++ b/crates/cargo-test-support/src/compare.rs
@@ -36,8 +36,9 @@
 //!   a problem.
 //! - Carriage returns are removed, which can help when running on Windows.
 
-use crate::diff;
+use crate::cross_compile::try_alternate;
 use crate::paths;
+use crate::{diff, rustc_host};
 use anyhow::{bail, Context, Result};
 use serde_json::Value;
 use std::fmt;
@@ -173,6 +174,10 @@ fn add_common_redactions(subs: &mut snapbox::Redactions) {
         regex!("home/\\.cargo/registry/src/-(?<redacted>[a-z0-9]+)"),
     )
     .unwrap();
+    subs.insert("[HOST_TARGET]", rustc_host()).unwrap();
+    if let Some(alt_target) = try_alternate() {
+        subs.insert("[ALT_TARGET]", alt_target).unwrap();
+    }
 }
 
 static MIN_LITERAL_REDACTIONS: &[(&str, &str)] = &[
--- a/crates/cargo-test-support/src/compare.rs
+++ b/crates/cargo-test-support/src/compare.rs
@@ -45,6 +45,15 @@ use std::path::Path;
 use std::str;
 use url::Url;
 
+/// This makes it easier to write regex replacements that are guaranteed to only
+/// get compiled once
+macro_rules! regex {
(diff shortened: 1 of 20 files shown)


crates/cargo-test-support/src/compare.rs
```diff
@@ -165,6 +165,11 @@ fn add_common_redactions(subs: &mut snapbox::Redactions) {
         regex::Regex::new("home/\\.cargo/registry/src/-(?<redacted>[a-z0-9]+)").unwrap(),
     )
     .unwrap();
+    subs.insert(
+        "[HASH]",
+        regex::Regex::new(r"[a-z0-9\-_]+-(?<redacted>[0-9a-f]{16})").unwrap(),
```
weihanglo, line 170: This doesn't work with `-C metadata=<hash>`.

crates/cargo-test-support/src/compare.rs
weihanglo, line 1: We're compiling regex object every time when `assert_e2e`/`assert_ui`  is called. I didn't measure it but it won't surprise me that there is a performance loss when we get more and more regexes

crates/cargo-test-support/src/compare.rs
```diff
@@ -178,15 +165,35 @@ pub fn assert_e2e() -> snapbox::Assert {
         regex::Regex::new("home/\\.cargo/registry/src/-(?<redacted>[a-z0-9]+)").unwrap(),
     )
     .unwrap();
-    snapbox::Assert::new()
-        .action_env(snapbox::assert::DEFAULT_ACTION_ENV)
-        .redact_with(subs)
+    subs.insert(
+        "[HASH]",
+        regex::Regex::new(r"[a-z0-9\-_]+-(?<redacted>[0-9a-f]{16})").unwrap(),
```
epage, line 170: This is still fairly broad. Is there something else about these situations we can pick up on, like use of `/`?

crates/cargo-test-support/src/compare.rs
epage: Technically, this is trying to cover other cases so that would be another entry to cover that which wouldn't be a blocker for this PR but a nice to have

weihanglo: Is there a way to forced regenerate existing snapshots, so we can add and remove redactions more freely, and won't bother forgetting to update some overly omitted snapshots?

epage: Not at this time.  If we remove a redaction, it'll require an update anyways.

crates/cargo-test-support/src/compare.rs
Muscraft: I switched to check for a `/` in front

PR #14055, opened 2024-06-13
[experiment] patch with patch files (re-resolve)

weihanglo:

This is an alternative of <https://github.com/rust-lang/cargo/pull/13779>.

Most parts of the design are the same. The differences are:

* `patch` entry no longer requires an exact version requirement, e.g. `=1.2.3`.
  Instead, only after a few rounds of dependency resolutions can a patch entry know which package version it is going to patch.
* One `patch` is able to match multiple versions, if applicable.
* Patching `package.version` field to a SemVer incompat version is allowed but kinda useless. The metadata of patched package, for example v2.0.0-to-v1.0.0, is different from the metadata of the real v1.0.0. Thus rustc things they are different versions.

This is built upon #13779 with two new commits.

While this works, I don't think the code change is in a good state to merge. The design (hacking on a new `SourceKind`) could be inherently wrong.

The ugliest part is code that handles the reuse of existing patched in lockfile edd3172c2da0f274367a6fdeecf9916619e7bb75. There are rooms for cleanup but I don't believe it'll be significantly better.

### Additional information

Doc hasn't got updated.


weihanglo:
See <https://github.com/rust-lang/cargo/pull/13779#issuecomment-2278138457>.

Issue #10273, opened 2022-01-08
config.toml support [env] with [env.<cfg>]


eric5f3759df:
### Problem

some build need arch assotiated targe environment var. like rusty_v8 RUSTY_V8_ARCHIVE, so it's usefull for [env] supporting multi arch.

weihanglo:
I guess cargo maintainers here possibly have little knowledge about V8 and Deno stuff. It would be great if you can provide more contexts, such like

- What exactly you're trying to resolve. A concrete example and links are great.
- Why the feature should be built-in Cargo.
- What the current solution/workaround looks like without this feature.

This would help not only maintainers but other random contributors to come up with better ideas. 😁

Gaubee:
eg:
```toml
[env]
RUSTY_V8_ARCHIVE = { value = "assets/rusty_v8_mirror/v0.42.0/librusty_v8_release_aarch64-apple-darwin.a", relative = true }

[env.'cfg(target_os="android")']
RUSTY_V8_ARCHIVE = { value = "assets/rusty_v8_mirror/v0.42.0/librusty_v8_release_aarch64-linux-android.a", relative = true }

```

nmittler:
Same issue here, but was using boringssl. I captured the details here: https://github.com/rust-lang/cargo/issues/11385

MatrixDev:
IMHO it would be much much better if we could just setup all these configs (including most of the cargo.toml) from the `rs` script (similar to what gradle does). This would simplify non-trivial configuration so much....

Because currently instead of "everything is rust" it is "everything is rust hidden behind toml and shell scripts" :(

weihanglo:
> if we could just setup all these configs…

For configuring environment variables, it's available today in build script with [`cargo:rustc-env=VAR=VALUE`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-envvarvalue).

MatrixDev:
@weihanglo, correct me if I'm wrong, but does it propagate to the libraries before their `build.rs` is executed?

oblique:
@weihanglo Not really. This does not propagate to dependencies, so you can not just define `OPENSSL_DIR` from a `build.rs`.

weihanglo:
THank you all. That is absolutely the limitation of build script today. Configuration is the current way to deal with that.

MatrixDev:
[...]

PR #14058, opened 2024-06-13
fix(env): cargo:rerun-if-env-changed doesn't work with env configuration

### LINKED ISSUE #8693: rerun-if-env-changed=CARGO_MANIFEST_DIR doesn't trigger a rebuild on change

heisen-li:

Fixes https://github.com/rust-lang/cargo/issues/10358
Fixes #8693 

### Additional information

Due to my stupidity, I screwed up the [previous PR](https://github.com/rust-lang/cargo/pull/14027). I didn't know what to do with it, so I created a new  PR.

r?@epage

### DIFF
--- a/src/cargo/core/compiler/build_context/target_info.rs
+++ b/src/cargo/core/compiler/build_context/target_info.rs
@@ -21,6 +21,8 @@ use cargo_util::{paths, ProcessBuilder};
 use serde::{Deserialize, Serialize};
 use std::cell::RefCell;
 use std::collections::hash_map::{Entry, HashMap};
+use std::collections::BTreeMap;
+use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::str::{self, FromStr};
 
@@ -587,6 +589,10 @@ impl TargetInfo {
         self.support_split_debuginfo
             .iter()
             .any(|sup| sup.as_str() == split.as_str())
+    }
+
+    pub fn get_target_envs(&self) -> &BTreeMap<String, Option<OsString>> {
+        return self.crate_type_process.get_envs();
     }
 }
 
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -357,7 +357,9 @@ mod dirty_reason;
 
 use std::collections::hash_map::{Entry, HashMap};
 
+use std::collections::BTreeMap;
 use std::env;
+use std::ffi::OsString;
 use std::hash::{self, Hash, Hasher};
 use std::io;
 use std::path::{Path, PathBuf};
@@ -768,14 +770,35 @@ pub enum StaleItem {
 impl LocalFingerprint {
     /// Read the environment variable of the given env `key`, and creates a new
     /// [`LocalFingerprint::RerunIfEnvChanged`] for it.
-    ///
-    // TODO: This is allowed at this moment. Should figure out if it makes
-    // sense if permitting to read env from the config system.
     #[allow(clippy::disallowed_methods)]
-    fn from_env<K: AsRef<str>>(key: K) -> LocalFingerprint {
-        let key = key.as_ref();
+    fn from_env<K: AsRef<str>>(
+        key: K,
+        envs: &BTreeMap<String, Option<OsString>>,
+    ) -> LocalFingerprint {
+        fn get_envs_case_insensitive(
+            key: &str,
+            envs: &BTreeMap<String, Option<OsString>>,
+        ) -> Option<OsString> {
+            let upper_case_key: String = key.to_uppercase();
+            for (k, v) in envs {
+                if k.to_uppercase().eq(&upper_case_key) {
+                    return v.to_owned();
+                }
+            }
+            None
+        }
+
+        let key: &str = key.as_ref();
         let var = key.to_owned();
-        let val = env::var(key).ok();
+
+        let val: Option<String> = if cfg!(windows) {
+            get_envs_case_insensitive(key, envs)
+        } else {
+            envs.get(key).and_then(|v| v.to_owned())
+        }
+        .or(env::var_os(key))
+        .and_then(|os_str| os_str.into_string().ok());
+
         LocalFingerprint::RerunIfEnvChanged { var, val }
     }
 
@@ -1608,6 +1631,12 @@ fn build_script_local_fingerprints(
     bool,
 ) {
     assert!(unit.mode.is_run_custom_build());
+    let envs = build_runner
+        .bcx
+        .target_data
+        .info(unit.kind)
+        .get_target_envs()
+        .clone();
     // First up, if this build script is entirely overridden, then we just
     // return the hash of what we overrode it with. This is the easy case!
     if let Some(fingerprint) = build_script_override_fingerprint(build_runner, unit) {
@@ -1660,7 +1689,12 @@ fn build_script_local_fingerprints(
             // Ok so now we're in "new mode" where we can have files listed as
             // dependencies as well as env vars listed as dependencies. Process
             // them all here.
-            Ok(Some(local_fingerprints_deps(deps, &target_dir, &pkg_root)))
+            Ok(Some(local_fingerprints_deps(
+                deps,
+                &target_dir,
+                &pkg_root,
+                &envs,
+            )))
         };
 
     // Note that `false` == "not overridden"
@@ -1695,6 +1729,7 @@ fn local_fingerprints_deps(
     deps: &BuildDeps,
     target_root: &Path,
     pkg_root: &Path,
+    envs: &BTreeMap<String, Option<OsString>>,
 ) -> Vec<LocalFingerprint> {
     debug!("new local fingerprints deps {:?}", pkg_root);
     let mut local = Vec::new();
@@ -1719,7 +1754,7 @@ fn local_fingerprints_deps(
     local.extend(
         deps.rerun_if_env_changed
             .iter()
-            .map(LocalFingerprint::from_env),
+            .map(|v| LocalFingerprint::from_env(v, &envs)),
     );
 
     local
--- a/tests/testsuite/build_script_env.rs
+++ b/tests/testsuite/build_script_env.rs
@@ -6,6 +6,165 @@ use cargo_test_support::sleep_ms;
 use cargo_test_support::str;
 
 #[cargo_test]
+fn rerun_if_env_changes_config() {
+    let p = project()
+        .file("Cargo.toml", &basic_manifest("foo", "0.1.0"))
+        .file("src/main.rs", "fn main() {}")
+        .file(
+            ".cargo/config.toml",
+            r#"
+                [env]
+                FOO = "good"
+            "#,
+        )
+        .file(
+            "build.rs",
+            r#"
+                fn main() {
+                    println!("cargo:rerun-if-env-changed=FOO");
+                    if let Ok(foo) = std::env::var("FOO") {
+                        assert!(&foo != "bad");
+                    }
+                }
+            "#,
+        )
+        .build();
+
+    p.cargo("check")
+        .with_stderr_data(str![[r#"
+[COMPILING] foo v0.1.0 ([ROOT]/foo)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
+        .run();
+
+    p.change_file(
+        ".cargo/config.toml",
+        r#"
+            [env]
+            FOO = "bad"
+        "#,
+    );
+
+    p.cargo("check")
+        .with_stderr_data(str![[r#"
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
+        .run();
+
+    p.cargo("clean").run();
+    p.cargo("check")
+        .with_status(101)
+        .with_stderr_data(
+            "\
+            [COMPILING] foo v0.1.0 ([ROOT]/foo)
+[ERROR] failed to run custom build command for `foo v0.1.0 ([..])`
+...",
+        )
+        .run();
+}
+
+#[cfg(windows)]
+#[cargo_test]
+fn rerun_if_env_changes_config_windows() {
+    let p = project()
+        .file("Cargo.toml", &basic_manifest("foo", "0.1.0"))
+        .file("src/main.rs", "fn main() {}")
+        .file(
+            ".cargo/config.toml",
+            r#"
+                [env]
+                FOO = "good"
+            "#,
+        )
+        .file(
+            "build.rs",
+            r#"
+                fn main() {
+                    println!("cargo:rerun-if-env-changed=Foo");
+                    if let Ok(foo) = std::env::var("Foo") {
+                        assert!(&foo != "bad");
+                    }
+                }
+            "#,
+        )
+        .build();
+
+    p.cargo("check")
+        .with_stderr_data(str![[r#"
+[COMPILING] foo v0.1.0 ([ROOT]/foo)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
+        .run();
+
+    p.change_file(
+        ".cargo/config.toml",
+        r#"
+            [env]
+            FOO = "bad"
+        "#,
+    );
+
+    p.cargo("check")
+        .with_stderr_data(str![[r#"
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
+        .run();
+
+    p.cargo("clean").run();
+    p.cargo("check")
+        .with_status(101)
+        .with_stderr_data(
+            "\
+[COMPILING] foo v0.1.0 ([ROOT]/foo)
+[ERROR] failed to run custom build command for `foo v0.1.0 ([..])`
+...",
+        )
+        .run();
+}
+
+#[cargo_test]
+fn rerun_if_env_changes_cargo_set_env_variable() {
+    let p = project()
+        .file("src/main.rs", "fn main() {}")
+        .file(
+            "build.rs",
+            r#"
+                fn main() {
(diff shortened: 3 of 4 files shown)


src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -772,10 +774,20 @@ impl LocalFingerprint {
...
         let key = key.as_ref();
         let var = key.to_owned();
-        let val = env::var(key).ok();
+        let val = envs
+            .get(key)
+            .and_then(|opt| {
+                opt.as_ref()
+                    .and_then(|os_str| os_str.clone().into_string().ok())
+            })
+            .or_else(|| env::var_os(key).and_then(|os_str| os_str.into_string().ok()));
```
epage, line 789: With using `env::var_os`, we can consolidate `.and_then(|os_str| os_str.into_string().ok())` to happen after both branches

```suggestion
        let val = envs
            .get(key)
            .map(|v| v.to_owned())
            .or_else(|| env::var_os(key))
            .and_then(|os_str| {
                os_str.into_string().ok()
            });
```

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -772,10 +774,20 @@ impl LocalFingerprint {
...
+    ) -> LocalFingerprint {
         let key = key.as_ref();
         let var = key.to_owned();
-        let val = env::var(key).ok();
+        let val = envs
+            .get(key)
+            .and_then(|opt| {
+                opt.as_ref()
+                    .and_then(|os_str| os_str.clone().into_string().ok())
+            })
```
epage, line 788: https://github.com/rust-lang/cargo/pull/14027#discussion_r1636789147

> iiuc on Windows, `env::var` calls `GetEnvironmentVariableW` which is case insensitive. Since we're recreating our own environment look up, should this matter?

epage: https://github.com/rust-lang/cargo/pull/14027#discussion_r1637708208

> I don't have a more in-depth study of windows, I'm very sorry.
> 
> Do you mean that you shouldn't use `env::var` or `env::var_os` to query environment variables? That doesn't seem to work : see: [rerun_if_env_changes](https://github.com/rust-lang/cargo/pull/14027/files#diff-bec1ea82298c9855dced8a66db57ca380f385e70254ca1cd3b8f99152465854dR55) and [rerun_if_env_or_file_changes](https://github.com/rust-lang/cargo/pull/14027/files#diff-bec1ea82298c9855dced8a66db57ca380f385e70254ca1cd3b8f99152465854dR109)

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -588,6 +590,10 @@ impl TargetInfo {
             .iter()
             .any(|sup| sup.as_str() == split.as_str())
     }
+
+    pub fn get_target_envs(&self) -> CargoResult<&BTreeMap<String, Option<OsString>>> {
+        return Ok(self.crate_type_process.get_envs());
+    }
```
epage, line 596: Why does this return a `CargoResult` if its always `Ok`.  The `unwrap` in the caller seemed dubious.

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1608,6 +1620,13 @@ fn build_script_local_fingerprints(
     bool,
 ) {
     assert!(unit.mode.is_run_custom_build());
+    let envs = build_runner
+        .bcx
+        .target_data
+        .info(unit.kind)
+        .get_target_envs()
+        .unwrap()
+        .clone();
```
epage, line 1629: I've not looked at the full control flow.  Do we have to `clone`?

src/cargo/core/compiler/fingerprint/mod.rs
heisen-li: thanks

src/cargo/core/compiler/build_context/target_info.rs
heisen-li: I was wrong.

src/cargo/core/compiler/fingerprint/mod.rs
epage: Ah, `envs` gets captured by a closure that this function returns so everything has to be `'static`.

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -772,10 +774,18 @@ impl LocalFingerprint {
     // TODO: This is allowed at this moment. Should figure out if it makes
     // sense if permitting to read env from the config system.
```
epage, line 775: With your change, this `TODO` is no longer relevant, right?

src/cargo/core/compiler/fingerprint/mod.rs
epage: @ChrisDenton looking for thoughts on case-insensitive env var lookup on Windows.

When generating a `rerun-if-env-changed` fingerprint, this PR emulates env variable lookup that happens within build scripts by reading the environment we set on build scripts and prioritizing that over `std::env::var` calls (which the build script would implicitly inherit).  The build script and `std::env::var`, iiuc, would be doing case insensitive lookups on Windows.  For the variables we explicitly set on build scripts, we'd be doing a case sensitive lookup.

weihanglo: BTW for env snapshot in Cargo we have done this: <https://github.com/rust-lang/cargo/blob/00ea165442eba4bb25d50b754c3a64667110e593/src/cargo/util/context/environment.rs#L132>.

ChrisDenton: > When generating a rerun-if-env-changed fingerprint, this PR emulates env variable lookup that happens within build scripts by reading the environment we set on build scripts and prioritizing that over std::env::var calls (which the build script would implicitly inherit). The build script and std::env::var, iiuc, would be doing case insensitive lookups on Windows. For the variables we explicitly set on build scripts, we'd be doing a case sensitive lookup.

I've not looked at this PR yet but that sounds not ideal. E.g. It's not uncommon for `PATH` to be spelled  `Path` or `path`.

epage: If we are already doing case insensitive lookups in other parts of cargo, seems like we should here as well

heisen-li: Sorry, I need to confirm this. If it's in a windows environment, getting the environment variables in the build script is not case sensitive?

epage: Yes, in the windows environment, look ups are case insensitive.

heisen-li: Added `windows` environment case insensitive scenarios and corresponding tests.

epage:
btw I believe this would also fix #8693.

This would likely need to revert #12482.

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -775,13 +775,26 @@ impl LocalFingerprint {
...
-        let val = envs
-            .get(key)
-            .map(|v| v.to_owned())
-            .or_else(|| Some(env::var_os(key)))
-            .and_then(|os_str| os_str?.into_string().ok());
+
+        let val: Option<String> = match envs.get(key) {
+            Some(v) => v.to_owned(),
+            None => get_env_value(key, envs),
+        }
```
epage, line 796: This feels like a strange split in logic
- Direct look up happens
- Call a function
  - Case insensitive lookup happens
  - actual env look up happens

heisen-li: My previous thought was to try to keep the previous habit of all prioritizing finding case sensitive variables. Then look for case insensitive variables.

Maybe what I'm considering is redundant.

tests/testsuite/build_script_env.rs
```diff
@@ -7,6 +7,26 @@ use cargo_test_support::str;
...
+    } else {
+        r#"
+fn main() {
+    println!("cargo:rerun-if-env-changed=FOO");
+    if let Ok(foo) = std::env::var("FOO") {
+        assert!(&foo != "bad");
+    }
+}
+"#
+    };
```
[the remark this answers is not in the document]
heisen-li: done

src/cargo/core/compiler/fingerprint/mod.rs
epage: Oh, I think that makes sense.  Its more the division of top-level code / function call felt odd.  We split the lookup into the map between the two calls and hide away the `std::env` look up in the case insensitive call (which has a generic, unclear name)

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -775,13 +775,29 @@ impl LocalFingerprint {
...
-            .map(|v| v.to_owned())
-            .or_else(|| Some(env::var_os(key)))
-            .and_then(|os_str| os_str?.into_string().ok());
+
+        let val: Option<String> = if cfg!(windows) {
+            get_envs_case_insensitive(key, envs)
+        } else {
+            envs.get(key).and_then(|v| v.to_owned())
+        }
+        .xor(env::var_os(key))
```
epage, line 799: I missed this before, why are we using `xor` rather than `or`?

tests/testsuite/build_script_env.rs
```diff
@@ -48,12 +48,63 @@ fn rerun_if_env_changes_config() {
...
 [ERROR] failed to run custom build command for `foo v0.1.0 ([ROOT]/foo)`
-...",
+...
+"#]])
+        .run();
+}
+
+#[cfg(windows)]
+#[cargo_test]
+fn rerun_if_env_changes_config_in_windows() {
```
epage, line 61: Can you please add all tests in commits before any functionality changes?

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -768,14 +770,35 @@ pub enum StaleItem {
 impl LocalFingerprint {
     /// Read the environment variable of the given env `key`, and creates a new
     /// [`LocalFingerprint::RerunIfEnvChanged`] for it.
-    ///
-    // TODO: This is allowed at this moment. Should figure out if it makes
-    // sense if permitting to read env from the config system.
     #[allow(clippy::disallowed_methods)]
-    fn from_env<K: AsRef<str>>(key: K) -> LocalFingerprint {
-        let key = key.as_ref();
+    fn from_env<K: AsRef<str>>(
```
epage, line 774: > This would likely need to revert https://github.com/rust-lang/cargo/pull/12482.

Don't we need to update the documentation with this?

tests/testsuite/build_script_env.rs
```diff
@@ -5,6 +5,109 @@ use cargo_test_support::project;
 use cargo_test_support::sleep_ms;
 use cargo_test_support::str;
 
+#[cargo_test]
+fn rerun_if_env_changes_config() {
```
epage, line 9: > btw I believe this would also fix https://github.com/rust-lang/cargo/issues/8693.

Since we should be getting that fix "for free", could you add a test for a cargo-set env variable (don't think it has to literally be that one) and mark this PR as fixing that issue as well?

tests/testsuite/build_script_env.rs
heisen-li: If I understand correctly, they should all be added.

src/cargo/core/compiler/fingerprint/mod.rs
heisen-li: Sorry for the long delay with work and life stuff.

Does `revert` mean to revert to a previous state?I used google translate.

tests/testsuite/build_script_env.rs
heisen-li: It has been added and you can continue to review it.

tests/testsuite/build_script_env.rs
```diff
@@ -5,6 +5,165 @@ use cargo_test_support::project;
...
+"#]])
+        .run();
+
+    p.change_file(
+        ".cargo/config.toml",
+        r#"
+            [env]
+            OUT_DIR = "somedir"
+        "#,
+    );
```
epage, line 139: The intention of this test is for detecting changes made by cargo, and not the user.  Us manually setting one runs counter to that

src/cargo/core/compiler/fingerprint/mod.rs
epage: > Sorry for the long delay with work and life stuff.

Fully understand.  I'm behind on a lot of stuff.

> Does revert mean to revert to a previous state?I used google translate.

Yes.  Its a term commonly used in VCS, like git.  Basically "undo".

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -768,14 +770,35 @@ pub enum StaleItem {
...
+        let key: &str = key.as_ref();
         let var = key.to_owned();
-        let val = env::var(key).ok();
+
+        let val: Option<String> = if cfg!(windows) {
+            get_envs_case_insensitive(key, envs)
+        } else {
+            envs.get(key).and_then(|v| v.to_owned())
+        }
+        .or(env::var_os(key))
```
weihanglo, line 799: Lazy evaluation
```suggestion
        .or_else(|| env::var_os(key))
```

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -768,14 +770,35 @@ pub enum StaleItem {
...
+    fn from_env<K: AsRef<str>>(
+        key: K,
+        envs: &BTreeMap<String, Option<OsString>>,
+    ) -> LocalFingerprint {
+        fn get_envs_case_insensitive(
+            key: &str,
+            envs: &BTreeMap<String, Option<OsString>>,
+        ) -> Option<OsString> {
+            let upper_case_key: String = key.to_uppercase();
+            for (k, v) in envs {
```
weihanglo, line 783: This conversion seems to be in a deeply nested loop. M `rerun_if_env_changed` * N `[env]` keys on Windows. We should probably find a way to do it once and for all.

This doesn't really seem to be a performance issue for general usage though.

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -588,6 +590,10 @@ impl TargetInfo {
             .iter()
             .any(|sup| sup.as_str() == split.as_str())
     }
+
+    pub fn get_target_envs(&self) -> &BTreeMap<String, Option<OsString>> {
+        return self.crate_type_process.get_envs();
```
weihanglo, line 595: Accessing `crate_type_process` seems like a hacky shortcut to me. The process is for learning target info from rustc, but fingerprint is for actual rustc compilation. One extension point for putting it here is for supporting <https://github.com/rust-lang/cargo/issues/10273>, though I believe that will need other rounds of env var system overhual in Cargo.

tests/testsuite/build_script_env.rs
epage: I'm not too sure what the concern is.

weihanglo: Setting `env.OUR_DIR` has no effect on build script execution.

epage: Oh, that.  The intent of my comment was to say that we should not test this with `[env]` but making a change that causes the cargo-provided env to change.

src/cargo/core/compiler/build_context/target_info.rs
ehuss: I would agree, the chance of this having unintended consequences seems kinda high since it isn't obvious when looking at `TargetInfo` that this would also influence build-script environment fingerprinting.

Would it be possible to rewrite `apply_env_config` so that there is a separate function that yields a map of resolved config values? Then the code in the fingerprint module could call that and collect the results into a map that it could then pass into the closure.

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -768,14 +770,35 @@ pub enum StaleItem {
 impl LocalFingerprint {
     /// Read the environment variable of the given env `key`, and creates a new
     /// [`LocalFingerprint::RerunIfEnvChanged`] for it.
-    ///
-    // TODO: This is allowed at this moment. Should figure out if it makes
-    // sense if permitting to read env from the config system.
```
ehuss, line 773: Instead of removing this, can we have a comment explaining why the `allow` can't be removed? It's not obvious to me that there is some fundamental problem that is preventing this.

I presume the problem is that the closure cannot access the GlobalContext, and thus is unable to access the env snapshot, and thus must access the raw environment directly? It seems like there are still options we can consider, such as wrapping the environment in Arc or something like that?

src/cargo/core/compiler/fingerprint/mod.rs
ehuss: A small fix would be to check the map with the key as-is, and then only check for the case-insensitive if that fails. 99% of the time the case will match. [`EnvConfig::get_env_os`](https://github.com/rust-lang/cargo/blob/a7917fdc3ae499b2324e45b94ed7fd08cc242b1c/src/cargo/util/context/environment.rs#L115-L126) does something similar.

ehuss:
I'm not clear, how does this fix #8693?

src/cargo/core/compiler/fingerprint/mod.rs
weihanglo: > such as wrapping the environment in Arc

Agree `Arc` is the way to resolve this.

MabezDev:
@heisen-li Do you plan on finishing this up? It would be great to see this land.

weihanglo:
No worries @heisen-li. Your help here is appreciated. I'll close this in favor of #14756. Thanks you!

PR #14060, opened 2024-06-13
test: migrate help to snapbox

choznerol:

Part of #14039.

Migrate `tests/testsuite/help.rs` to snapbox.

I followed #14039. The `#![allow(deprecated)]` was removed and `tests/testsuite/help.rs` is still passing.

### Additional information

~I accidentally noticed that [L155](https://github.com/rust-lang/cargo/pull/14060/files#diff-5a3eeb8da75123b44c76e8ecd96f6b78b9a921fb1f1ae5dcf6d1d229bc41bd7fL150-R155) would always pass despite the `.with_stderr_data("whatever I write here")`, either before or after this PR. It's probably related to `run_expect_error()`. Will investigate and open an issue.~ 👉 Filed #14076

### DIFF
--- a/tests/testsuite/help.rs
+++ b/tests/testsuite/help.rs
@@ -140,7 +140,11 @@ fn help_alias() {
     // The `empty-alias` returns an error.
     cargo_process("help empty-alias")
         .env("PATH", Path::new(""))
-        .with_stderr_data("[..] The subcommand 'empty-alias' wasn't recognized [..]")
+        .with_stderr_data("The subcommand 'empty-alias' wasn't recognized
+
+FIXME: #14076 This assertion isn't working, as this line should have caused a test failure but didn't.
+        ",
+        )
         .run_expect_error();
 
     // Because `simple-alias` aliases a subcommand with no arguments, help shows the manpage.
--- a/tests/testsuite/help.rs
+++ b/tests/testsuite/help.rs
@@ -14,11 +14,6 @@ fn help() {
     cargo_process("help build").run();
     cargo_process("build -h").run();
     cargo_process("help help").run();
-    // Ensure that help output goes to stdout, not stderr.
-    cargo_process("search --help").with_stderr("").run();
-    cargo_process("search --help")
-        .with_stdout_contains("[..] --frozen [..]")
-        .run();
 }
 
 #[cargo_test]
--- a/tests/testsuite/help.rs
+++ b/tests/testsuite/help.rs
@@ -155,18 +155,3 @@ fn help_alias() {
     let out = help_with_stdout_and_path("complex-alias", Path::new(""));
     assert_eq!(out, "`complex-alias` is aliased to `build --release`\n");
 }
-
-#[cargo_test]
-fn alias_z_flag_help() {
-    for cmd in ["build", "run", "check", "test", "b", "r", "c", "t"] {
-        cargo_process(&format!("{cmd} -Z help"))
-            .with_stdout_data(
-                "\
-...
-    -Z allow-features[..]  Allow *only* the listed unstable features
-...
-",
-            )
-            .run();
-    }
-}
--- a/tests/testsuite/help.rs
+++ b/tests/testsuite/help.rs
@@ -1,6 +1,7 @@
 //! Tests for cargo's help output.
 
 use cargo_test_support::registry::Package;
+use cargo_test_support::str;
 use cargo_test_support::{basic_manifest, cargo_exe, cargo_process, paths, process, project};
 use std::fs;
 use std::path::Path;
@@ -34,7 +35,10 @@ fn help_external_subcommand() {
         .publish();
     cargo_process("install cargo-fake-help").run();
     cargo_process("help fake-help")
-        .with_stdout_data("fancy help output\n")
+        .with_stdout_data(str![[r#"
+fancy help output
+
+"#]])
         .run();
 }
 
@@ -140,12 +144,17 @@ fn help_alias() {
     // The `empty-alias` returns an error.
     cargo_process("help empty-alias")
         .env("PATH", Path::new(""))
-        .with_stderr_data("The subcommand 'empty-alias' wasn't recognized
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] no such command: `empty-alias`
 
-FIXME: #14076 This assertion isn't working, as this line should have caused a test failure but didn't.
-        ",
-        )
-        .run_expect_error();
+	Did you mean `empty-alias`?
+
+	View all installed commands with `cargo --list`
+	Find a package to install `empty-alias` with `cargo search cargo-empty-alias`
+
+"#]])
+        .run();
 
     // Because `simple-alias` aliases a subcommand with no arguments, help shows the manpage.
     help_with_man_and_path("", "simple-alias", "build", Path::new(""));
--- a/tests/testsuite/help.rs
+++ b/tests/testsuite/help.rs
@@ -1,7 +1,5 @@
 //! Tests for cargo's help output.
 
-#![allow(deprecated)]
-
 use cargo_test_support::registry::Package;
 use cargo_test_support::{basic_manifest, cargo_exe, cargo_process, paths, process, project};
 use std::fs;
@@ -17,9 +15,15 @@ fn help() {
     cargo_process("build -h").run();
     cargo_process("help help").run();
     // Ensure that help output goes to stdout, not stderr.
-    cargo_process("search --help").with_stderr("").run();
+    cargo_process("search --help").with_stderr_data("").run();
     cargo_process("search --help")
-        .with_stdout_contains("[..] --frozen [..]")
+        .with_stdout_data(
+            "\
+...
+[..] --frozen [..]
+...
+",
+        )
         .run();
 }
 
@@ -41,7 +45,7 @@ fn help_external_subcommand() {
         .publish();
     cargo_process("install cargo-fake-help").run();
     cargo_process("help fake-help")
-        .with_stdout("fancy help output\n")
+        .with_stdout_data("fancy help output\n")
         .run();
 }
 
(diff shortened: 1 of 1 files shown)


tests/testsuite/help.rs
```diff
@@ -17,9 +15,15 @@ fn help() {
     cargo_process("build -h").run();
     cargo_process("help help").run();
     // Ensure that help output goes to stdout, not stderr.
-    cargo_process("search --help").with_stderr("").run();
+    cargo_process("search --help").with_stderr_data("").run();
     cargo_process("search --help")
-        .with_stdout_contains("[..] --frozen [..]")
+        .with_stdout_data(
```
weihanglo, line 20: You could use `str![]` and regenerate snapshots by running this command `SNAPSHOTS=overwrite cargo t --test testsuite help::`

```suggestion
        .with_stdout_data(str![])
```

epage, line 20: We can probably delete this test because we have help UI tests for every command, like

https://github.com/rust-lang/cargo/blob/master/tests/testsuite/cargo_search/help/mod.rs

tests/testsuite/help.rs
```diff
@@ -147,7 +151,7 @@ fn help_alias() {
     // The `empty-alias` returns an error.
     cargo_process("help empty-alias")
         .env("PATH", Path::new(""))
-        .with_stderr_contains("[..]The subcommand 'empty-alias' wasn't recognized[..]")
+        .with_stderr_data("[..] The subcommand 'empty-alias' wasn't recognized [..]")
```
epage, line 154: Should we drop the `[..]`s?  We are trying to move away from them where it makes sense because they get in the way of snapshot updating

tests/testsuite/help.rs
```diff
@@ -162,8 +166,12 @@ fn help_alias() {
...
-            .with_stdout_contains(
-                "    -Z allow-features[..]  Allow *only* the listed unstable features",
+            .with_stdout_data(
+                "\
+...
+    -Z allow-features[..]  Allow *only* the listed unstable features
+...
+",
             )
             .run();
```
epage, line 176: We also might be able to remove this as well

https://github.com/rust-lang/cargo/blob/master/tests/testsuite/cargo/z_help/stdout.term.svg

tests/testsuite/help.rs
choznerol: Thanks, but somehow I can get `SNAPSHOTS=overwrite` taking effect for other tests, but not here.

<details><summary>Can auto fix other test with <code>SNAPSHOTS=overwrite cargo test --test testsuite binary_name::gated</code></summary>

```
$ SNAPSHOTS=overwrite cargo test --test testsuite binary_name::gated
    Blocking waiting for file lock on build directory
   Compiling cargo v0.82.0 (/Users/ychou/choznerol/contrib/cargo)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 7.24s
     Running tests/testsuite/main.rs (target/debug/deps/testsuite-00971f03e698357e)

running 1 test
Fixing:
---- expected: tests/testsuite/binary_name.rs:34:27
++++ actual:   stderr
   1    1 | [ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
   2    2 |
   3      - fix me
   4    3 | Caused by:
   5    4 |   feature `different-binary-name` is required
   6    5 |
   7      -   The package requires the Cargo feature called `different-binary-name`, but that feature is not stabilized in this version of Cargo ([..]).
        6 +   The package requires the Cargo feature called `different-binary-name`, but that feature is not stabilized in this version of Cargo (1.81.0 (5f0294b15 2024-06-13)).
   8    7 |   Consider adding `cargo-features = ["different-binary-name"]` to the top of Cargo.toml (above the [package] table) to tell Cargo you are opting in to use this unstable feature.
   9    8 |   See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#different-binary-name for more information about the status of this feature.


test binary_name::gated ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 3350 filtered out; finished in 1.21s
```

</details>


<details><summary>Can't auto fix this with <code>SNAPSHOTS=overwrite cargo test --test testsuite help::help</code></summary>

```diff
commit cbaa4a89e60507366401ac0076d652f5eafc4998
Author: Lawrence Chou <choznerol@protonmail.com>
Date:   Sat Jun 15 14:36:42 2024 +0800

    Tmp commit prior to `SNAPSHOTS=overwrite cargo t --test testsuite help::`

diff --git a/tests/testsuite/help.rs b/tests/testsuite/help.rs
index 80d236034..c1d8383a2 100644
--- a/tests/testsuite/help.rs
+++ b/tests/testsuite/help.rs
@@ -3,6 +3,7 @@
 #![allow(deprecated)]
 
 use cargo_test_support::registry::Package;
+use cargo_test_support::str;
 use cargo_test_support::{basic_manifest, cargo_exe, cargo_process, paths, process, project};
 use std::fs;
 use std::path::Path;
@@ -17,9 +18,11 @@ fn help() {
     cargo_process("build -h").run();
     cargo_process("help help").run();
     // Ensure that help output goes to stdout, not stderr.
-    cargo_process("search --help").with_stderr("").run();
+    cargo_process("search --help").with_stderr(str![]).run();
     cargo_process("search --help")
-        .with_stdout_contains("[..] --frozen [..]")
+        .with_stdout_contains(str![[r#"
+HEY SNAPBOX, FIX THIS AUTOMATICALLY PLEASE
+"#]])
         .run();
 }
```

```
$ SNAPSHOTS=overwrite cargo test --test testsuite help::help
   Compiling cargo v0.82.0 (/Users/ychou/choznerol/contrib/cargo)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 6.23s
     Running tests/testsuite/main.rs (target/debug/deps/testsuite-00971f03e698357e)

running 5 tests
test cargo_help::help::case ... ok
test help::help_alias ... ok
test help::help ... FAILED
test help::help_external_subcommand ... ok
test help::help_man ... ok

failures:

---- help::help stdout ----
running `/Users/ychou/choznerol/contrib/cargo/target/debug/cargo`
running `/Users/ychou/choznerol/contrib/cargo/target/debug/cargo help`
running `/Users/ychou/choznerol/contrib/cargo/target/debug/cargo -h`
running `/Users/ychou/choznerol/contrib/cargo/target/debug/cargo help build`
running `/Users/ychou/choznerol/contrib/cargo/target/debug/cargo build -h`
running `/Users/ychou/choznerol/contrib/cargo/target/debug/cargo help help`
running `/Users/ychou/choznerol/contrib/cargo/target/debug/cargo search --help`
thread 'help::help' panicked at tests/testsuite/help.rs:21:56:

test failed running `/Users/ychou/choznerol/contrib/cargo/target/debug/cargo search --help`
error: stderr did not match:
1        -tests/testsuite/help.rs:21:48


other output:
Search packages in the registry. Default registry is crates.io

Usage: cargo search [OPTIONS] [QUERY]...

Arguments:
  [QUERY]...

Options:
      --limit <LIMIT>        Limit the number of results (default: 10, max: 100)
      --index <INDEX>        Registry index URL to search packages in
      --registry <REGISTRY>  Registry to search packages in
  -v, --verbose...           Use verbose output (-vv very verbose/build.rs output)
  -q, --quiet                Do not print cargo log messages
      --color <WHEN>         Coloring: auto, always, never
      --config <KEY=VALUE>   Override a configuration value
  -Z <FLAG>                  Unstable (nightly-only) flags to Cargo, see 'cargo -Z help' for details
  -h, --help                 Print help

Manifest Options:
      --locked   Assert that `Cargo.lock` will remain unchanged
      --offline  Run without accessing the network
      --frozen   Equivalent to specifying both --locked and --offline

Run `cargo help search` for more detailed information.


failures:
    help::help

test result: FAILED. 4 passed; 1 failed; 0 ignored; 0 measured; 3346 filtered out; finished in 2.80s

error: test failed, to rerun pass `--test testsuite`
```

</details>


I initially suspected that `help.rs` mostly uses `cargo_process("...")` instead of `p.cargo("...")`, but couldn't really trace the root cause.

tests/testsuite/help.rs
choznerol: Removed in e5e9f2f, but in fact there is another issue #14076 with this test case. I mentioned that as well in the same commit.

tests/testsuite/help.rs
choznerol: Thanks. Removed in 3fd3879.

tests/testsuite/help.rs
choznerol: Thanks. Removed in eef057e.

choznerol: Note that this specific test was later removed in https://github.com/rust-lang/cargo/pull/14060#discussion_r1640859247

tests/testsuite/help.rs
```diff
@@ -147,7 +140,11 @@ fn help_alias() {
     // The `empty-alias` returns an error.
     cargo_process("help empty-alias")
         .env("PATH", Path::new(""))
-        .with_stderr_contains("[..]The subcommand 'empty-alias' wasn't recognized[..]")
+        .with_stderr_data("The subcommand 'empty-alias' wasn't recognized
+
+FIXME: #14076 This assertion isn't working, as this line should have caused a test failure but didn't.
+        ",
+        )
         .run_expect_error();
```
weihanglo, line 148:
```suggestion
        .with_status(101)
        .with_stderr_data(str![[r#"
[ERROR] no such command: `empty-alias`

	Did you mean `empty-alias`?

	View all installed commands with `cargo --list`
	Find a package to install `empty-alias` with `cargo search cargo-empty-alias`

"#]])
        .run();
```

Something like this should work. See <https://github.com/rust-lang/cargo/issues/14076#issuecomment-2169593172>.

(just realized that the error message regressed…)

weihanglo: Would you mind posting another PR fixing #14076 and then update this afterward?

choznerol: Thanks! Just get `SNAPSHOTS=overwrite` working and pushed 8db30bc.

> Would you mind posting another PR fixing #14076 and then update this afterward?

Sure! Will address #14076 first with #14078, and then resolve the conflict here.

tests/testsuite/help.rs
choznerol: Thanks! Just get `SNAPSHOTS=overwrite` working and pushed https://github.com/rust-lang/cargo/commit/8db30bc4bdd67de9947f1f9a5dd653edcc2c8d59.

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

Issue #6645, opened 2019-02-07
Proposal: Better shell completions


ehuss:
This is a proposal to add better shell completion support to Cargo. The outline of the strategy is:

- Add a `cargo completions` subcommand to assist with completion support.
- `cargo completions <shell>` will output a completion script for the given shell (bash, zsh, etc.).
- The completion script will execute `cargo completions` to implement the work of actually emitting completions. This can significantly simplify the work of supporting completions for different shells. The completion script is very small (see `npm completion` for an example). This also means all the logic can be written in one language (Rust) and can be more easily tested.

Benefits:
- Support richer completions (such as `--bin` or `--example` completing the target name).
- Maintain consistent completions across shells.
- Easier to test.
- (Maybe) easier to keep in sync with clap.
- More easily discoverable (who knows the completions even exist?).
- Completions automatically stay in sync with the release via rustup.

Drawbacks:
- More stuff to maintain.
- Minor performance drawback compared to native-shell code. Many of the existing completions already execute commands, so in many cases it is a wash. Cargo's launch speed isn't terrible (~50ms for me).
- Won't be backwards compatible with older versions of Cargo.

Please let me know if you have any thoughts or objections.

nrc:
> `cargo completions <shell>` will output a completion script for the given shell (bash, zsh, etc.)

I don't know how this works at all, but why do some tool's shell completions just work (e.g., Git) whereas the Rust tools seem to need some kind of opt-in step? Of course, I think it would be nice if we didn't need this opt-in step.

ehuss:
> didn't need this opt-in step

It depends on how rust/cargo is installed.  Some installers (like ubuntu) will automatically add the completion file to the global location (such as `/usr/share/bash-completion/completions/`) and it will "just work" assuming your shell is configured to load them.  rustup does not do that, since it avoids touching files outside of the user's home directory.  rustup could install them in the user's home directory (`$XDG_DATA_HOME/bash-completion/completions`, and probably something similar for other shells).  That would need to be a decision for rustup to make, if it is acceptable to dump files in various locations.

jpalomaki:
FWIW, I just started learning Rust (cargo), using rustup, and this was pretty much the first thing I hit. :smile: 

If we want to keep stuff in the user's home directory (as opposed to e.g. `/etc/bash_completion.d`), one way to do this (for bash) is:

```
echo 'source <(cargo completion bash)' >> ~/.bashrc
```

scooter-dangle:
[...]

PR #14084, opened 2024-06-16
zsh completion: allow setting custom registries for 'cargo add --registry'

pseyfert:

So far when I type in zsh `cargo add --regi<TAB>`, I get `cargo add --registry=` as suggestion. So far so good. But even though I have a private not-default registry configured in `~/.cargo/config.toml`, there is no further suggestion for what should come after the equal sign. My crates come basically from two registries in equal parts, so setting a default wouldn't help, as I still have to specify the registry on the command line half of the times. In the PR here, I read a zstyle where a user can specify custom registries. In my case, I can add
```
zstyle ':completion:*:complete:cargo:*:registries' registries "sevensense" "my employer"
```

In my .zshrc and get `sevensense` suggested. The format for the style is that of an associative array. I.e. for multiple registries:
```
zstyle ':completion:*:complete:cargo:*:registries' registries "wubbel" "I like wubbel" "wopper" "you can eat that at: Burger King" "yuhu" "the best"
```

And the tab completion would suggest
```
 > cargo add --registry=<TAB>
registry:
wubbel  -- I like wubbel
wopper  -- you can eat that at: Burger King
yuhu    -- the best
```

I also looked into reading the configured registries from `~/.cargo/config.toml` but afaiu `cargo` can't print them to the command line. I looked at https://github.com/z-shell/zsh-string-lib/tree/main but adding this as dependency seemed a bit too much complexity added.

*todo*: Having this code block where it is, is probably not optimal. Also, the intended usage should be documented somewhere, where users can find it, I'm open for suggestions.

`cargo add --registry=<TAB>` in zsh, with the _cargo completion in use, should not do anything with the style unset. With the style set as explained above, you should see that zsh suggests the registries, and completes other options as before.

I only tested `cargo add`, but also other commands take the --registry argument. I guess generally completion for more complex command lines with registry should be tested.

### DIFF
--- a/src/etc/_cargo
+++ b/src/etc/_cargo
@@ -60,11 +60,18 @@ _cargo() {
         "--no-default-features[don't build the default features]"
     )
 
+    typeset -A registries_tmp
+    typeset -a registries
+    zstyle -a ":completion:${curcontext}:registries" registries registries_tmp
+    for key in ${(k)registries_tmp}; do
+        registries+=("$key:${registries_tmp[$key]}")
+    done
+
     msgfmt='--message-format=[specify error format]:error format [human]:(human json short)'
     triple='--target=[specify target triple]:target triple:_cargo_target_triple'
     target='--target-dir=[specify directory for all generated artifacts]:directory:_directories'
     manifest='--manifest-path=[specify path to manifest]:path:_directories'
-    registry='--registry=[specify registry to use]:registry'
+    registry='--registry=[specify registry to use]:registry:((${registries[@]}))'
 
     case $state in
         args)


src/etc/_cargo
weihanglo, line 1: Thanks for being interested in improving the shell completion experience!

It is always painful for the team (at least for me) to maintain each shell completion. So, the current proposal is to move the completion to clap. The project is currently a GSoC (Google Summer of Code) project, and is actively developed I believe. Some useful links:

* GSoC project link: <https://summerofcode.withgoogle.com/programs/2024/projects/jjnidpgn>
* Zulip topic for the discussion <https://rust-lang.zulipchat.com/#narrow/stream/421156-gsoc/topic/Project.3A.20Move.20cargo.20shell.20completions.20to.20Rust>
* Issue in Cargo repo tracking it: https://github.com/rust-lang/cargo/issues/6645
* With a Rust native shell completion we may be able to reduce the risk of running code in malicious projects: https://github.com/rust-lang/rustup/issues/3740

---

Back to your implementation. I don't think it is wrong, yet it is a bit too specific (you need to define zstyle value to enable the auto-completion for a certain flag). Also, like you've mentioned, there is no a good way to document this feature, so users might not find its existence. If there are going to have more shell compleitions implemented like this one, it may be a burden to maintain in the long run.

weihanglo: (In an ideal world, a completion should collect registry names from Cargo configurations, from both environment variables and the hierarchy of `.cargo/config.toml`. That's too far a future I know 🥲 )

pseyfert: Thanks for the prompt feedback and providing the pointers,
I agree that maintaining the completions is not trivial (thanks for doing it so far!) and agree moving it to clap has huge maintainability benefits.

weihanglo: I've tracked this pull request in #6645, so that we can look into a better integration with Cargo configurations when implementing. Going to close this and thanks for providing the use case and idea!

PR #14089, opened 2024-06-17
Simplify checking for dependency cycles

Eh2406:

In my PubGrub work I had to try to understand the `check_cycles` post resolution pass. I found the code tricky to understand and doing some unnecessary reallocations. I cleaned up the code to remove the unnecessary data structures so that I could better understand.

It's an internal re-factor, and the tests still pass. But this code is not extensively tested by our test suite. In addition I ran my "resolve every crate on crates.io" script against a local check out of this branch. With the old code as `check_cycles_old` and the PR code as `check_cycles_new` the following assertion did not hit: 
```rust
fn check_cycles(resolve: &Resolve) -> CargoResult<()> {
    let old = check_cycles_old(resolve);
    let new = check_cycles_new(resolve);
    assert_eq!(old.is_err(), new.is_err());

    old
}
```

That experiment also demonstrated that this was not a significant performance change in either direction. 

### Additional information

If you're having questions about this code while you're reviewing, this would be a perfect opportunity for better naming and comments. Please ask questions.

### DIFF
--- a/src/cargo/core/resolver/mod.rs
+++ b/src/cargo/core/resolver/mod.rs
@@ -1004,69 +1004,57 @@ fn find_candidate(
 }
 
 fn check_cycles(resolve: &Resolve) -> CargoResult<()> {
-    // Create a simple graph representation alternative of `resolve` which has
-    // only the edges we care about. Note that `BTree*` is used to produce
-    // deterministic error messages here. Also note that the main reason for
-    // this copy of the resolve graph is to avoid edges between a crate and its
-    // dev-dependency since that doesn't count for cycles.
-    let mut graph = BTreeMap::new();
-    for id in resolve.iter() {
-        let map = graph.entry(id).or_insert_with(BTreeMap::new);
-        for (dep_id, listings) in resolve.deps_not_replaced(id) {
-            let transitive_dep = listings.iter().find(|d| d.is_transitive());
-
-            if let Some(transitive_dep) = transitive_dep.cloned() {
-                map.insert(dep_id, transitive_dep.clone());
-                resolve
-                    .replacement(dep_id)
-                    .map(|p| map.insert(p, transitive_dep));
-            }
-        }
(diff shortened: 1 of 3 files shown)


src/cargo/core/resolver/mod.rs
```diff
@@ -1004,69 +1004,57 @@ fn find_candidate(
...
     // same node twice that's a cycle.
-    let mut checked = HashSet::new();
-    let mut path = Vec::new();
-    let mut visited = HashSet::new();
-    for pkg in graph.keys() {
-        if !checked.contains(pkg) {
-            visit(&graph, *pkg, &mut visited, &mut path, &mut checked)?
+    let mut checked = HashSet::with_capacity(resolve.len());
+    let mut path = Vec::with_capacity(4);
+    let mut visited = HashSet::with_capacity(4);
```
weihanglo, line 1013: Why are these initialized with a capacity of 4?

weihanglo (approved):
Apart from the question, things look fine. Thank you!

src/cargo/core/resolver/mod.rs
Eh2406: Random guess. Initializing them with a capacity of 0 (`::new()`) seems silly when we know were going to push an item to it. Initializing them to the largest possible length (`::with_capacity(resolve.len());`) prevents a reallocation ever being needed, but is excessive as dependency trees tend to be wide not deep. So this was a compromise value I came up with off the top of my head. In effect it's just moving the allocation from "the first time we push" up to here. Given that none of this PR seems to have affected perf, I'd be happy to change it to whatever seems clear. What would you prefer?

weihanglo: I am fine with merge this as-is, as the rationale has been called out here.

weihanglo:
@bors r+

Thanks!

PR #14090, opened 2024-06-17
remove some legacy public dependency code from the resolver

Eh2406:
These conflict reasons are no longer constructed, since we removed backtracking caused by public dependencies. Let's go ahead and remove them.

r? @epage


weihanglo (approved):
Looks reasonable. Thanks!

PR #14091, opened 2024-06-17
test: migrate testsuites to snapbox

henry40408:

Migrate the following testsuites to snapbox:

- tests/testsuite/death.rs
- ~~tests/testsuite/locate_project.rs~~ see https://github.com/rust-lang/cargo/pull/14091#discussion_r1646183025
- tests/testsuite/message_format.rs
- tests/testsuite/precise_pre_release.rs
- tests/testsuite/verify_project.rs
- tests/testsuite/version.rs

N/A

### Additional information

N/A

### DIFF
--- a/tests/testsuite/message_format.rs
+++ b/tests/testsuite/message_format.rs
@@ -1,8 +1,6 @@
 //! Tests for --message-format flag.
 
-#![allow(deprecated)]
-
-use cargo_test_support::{basic_lib_manifest, basic_manifest, project};
+use cargo_test_support::{basic_lib_manifest, basic_manifest, project, str};
 
 #[cargo_test]
 fn cannot_specify_two() {
@@ -13,12 +11,14 @@ fn cannot_specify_two() {
 
     let formats = ["human", "json", "short"];
 
-    let two_kinds = "error: cannot specify two kinds of `message-format` arguments\n";
     for a in formats.iter() {
         for b in formats.iter() {
             p.cargo(&format!("build --message-format {},{}", a, b))
                 .with_status(101)
-                .with_stderr(two_kinds)
+                .with_stderr_data(str![[r#"
+[ERROR] cannot specify two kinds of `message-format` arguments
+
+"#]])
                 .run();
         }
     }
@@ -45,6 +45,7 @@ fn double_json_works() {
         .run();
 }
 
+#[allow(deprecated)]
 #[cargo_test]
 fn cargo_renders() {
     let p = project()
@@ -89,17 +90,17 @@ fn cargo_renders_short() {
 
     p.cargo("check --message-format json-render-diagnostics,json-diagnostic-short")
         .with_status(101)
-        .with_stderr_contains(
-            "\
-[CHECKING] foo [..]
-error[..]`main`[..]
-",
-        )
-        .with_stderr_does_not_contain("note:")
+        .with_stderr_data(str![[r#"
+[CHECKING] foo v0.1.0 ([ROOT]/foo)
+error[E0601]: `main` function not found in crate `foo`
+[ERROR] could not compile `foo` (bin "foo") due to 1 previous error
+
+"#]])
         .run();
 }
 
 #[cargo_test]
+#[allow(deprecated)]
 fn cargo_renders_ansi() {
     let p = project()
         .file("Cargo.toml", &basic_manifest("foo", "0.1.0"))
@@ -129,7 +130,24 @@ fn cargo_renders_doctests() {
 
     p.cargo("test --doc --message-format short")
         .with_status(101)
-        .with_stdout_contains("src/lib.rs:2:1: error[E0425]:[..]")
-        .with_stdout_contains("[..]src/lib.rs - bar (line 1)[..]")
+        .with_stdout_data(str![[r#"
+
+running 1 test
+test src/lib.rs - bar (line 1) ... FAILED
+
+failures:
+
+---- src/lib.rs - bar (line 1) stdout ----
+src/lib.rs:2:1: error[E0425]: cannot find function `bar` in this scope
+[ERROR] aborting due to 1 previous error
+Couldn't compile the test.
+
+failures:
+    src/lib.rs - bar (line 1)
+
+test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in [ELAPSED]s
+
+
+"#]])
         .run();
 }
--- a/tests/testsuite/version.rs
+++ b/tests/testsuite/version.rs
@@ -1,7 +1,5 @@
 //! Tests for displaying the cargo version.
 
-#![allow(deprecated)]
-
 use cargo_test_support::{cargo_process, project};
 
 #[cargo_test]
@@ -9,15 +7,15 @@ fn simple() {
     let p = project().build();
 
     p.cargo("version")
-        .with_stdout(&format!("cargo {}\n", cargo::version()))
+        .with_stdout_data(&format!("cargo {}\n", cargo::version()))
         .run();
 
     p.cargo("--version")
-        .with_stdout(&format!("cargo {}\n", cargo::version()))
+        .with_stdout_data(&format!("cargo {}\n", cargo::version()))
         .run();
 
     p.cargo("-V")
-        .with_stdout(&format!("cargo {}\n", cargo::version()))
+        .with_stdout_data(&format!("cargo {}\n", cargo::version()))
         .run();
 }
 
@@ -53,10 +51,19 @@ fn version_works_with_bad_target_dir() {
 fn verbose() {
     // This is mainly to check that it doesn't explode.
     cargo_process("-vV")
-        .with_stdout_contains(&format!("cargo {}", cargo::version()))
-        .with_stdout_contains("host: [..]")
-        .with_stdout_contains("libgit2: [..]")
-        .with_stdout_contains("libcurl: [..]")
-        .with_stdout_contains("os: [..]")
+        .with_stdout_data(format!(
+            "\
+cargo {}
+release: [..]
+commit-hash: [..]
+commit-date: [..]
+host: [HOST_TARGET]
+libgit2: [..] (sys:[..] [..])
+libcurl: [..] (sys:[..] [..])
+...
+os: [..]
+",
+            cargo::version()
+        ))
         .run();
 }
--- a/tests/testsuite/death.rs
+++ b/tests/testsuite/death.rs
@@ -1,7 +1,5 @@
 //! Tests for ctrl-C handling.
 
-#![allow(deprecated)]
-
 use cargo_test_support::{project, slow_cpu_multiplier};
 use std::fs;
 use std::io::{self, Read};
--- a/tests/testsuite/precise_pre_release.rs
+++ b/tests/testsuite/precise_pre_release.rs
@@ -1,8 +1,6 @@
 //! Tests for selecting pre-release versions with `update --precise`.
 
-#![allow(deprecated)]
-
-use cargo_test_support::project;
+use cargo_test_support::{project, str};
 
 #[cargo_test]
 fn requires_nightly_cargo() {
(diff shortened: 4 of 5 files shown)


tests/testsuite/locate_project.rs
```diff
@@ -1,15 +1,16 @@
...
+use cargo_test_support::{project, str};
 
 #[cargo_test]
 fn simple() {
     let p = project().build();
 
     p.cargo("locate-project")
-        .with_json(r#"{"root": "[ROOT]/foo/Cargo.toml"}"#)
+        .with_stdout_data(str![[r#"
+{"root":"[..]"}
```
hi-rustin, line 11: Do we need to use `{"root":"[ROOT]/foo/Cargo.toml"}` here?

tests/testsuite/message_format.rs
```diff
@@ -108,7 +113,13 @@ fn cargo_renders_ansi() {
...
...` function not found in crate `foo`","code":{"code":"E0601","explanation":"No `main` function was found in a binary crate./n/nTo fix this error, add a `main` function:/n/n```/nfn main() {/n    // Your program will start here./n    println!(/"Hello world!/");/n}/n```/n/nIf you don't know the basics of Rust, you can look at the/n[Rust Book][rust-book] to get started./n/n[rust-book]: https://doc.rust-lang.org/book//n"},"level":"error","spans":[{"file_name":"src/main.rs","byte_start":0,"byte_end":0,"line_start":0,"line_end":0,"column_start":1,"column_end":1,"is_primary":true,"text":[],"label":null,"suggested_replacement":null,"suggestion_applicability":null,"expansion":null}],"children":[{"message":"consider adding a `main` function to `src/main.rs`","code":null,"level":"note","spans":[],"children":[],"rendered":null}],"rendered":"/u001b[0m/u001b[1m/u001b[38;5;9merror[E0601]/u001b[0m/u001b[0m/u001b[1m: `main` function not found in crate `foo`/u001b[0m/n/u001b[0m  /u001b[0m/u001b[0m/u001b[1m/u001b[38;5;12m|/u001b[0m/n/u001b[0m  /u001b[0m/u001b[0m/u001b[1m/u001b[38;5;12m= /u001b[0m/u001b[0m/u001b[1mnote/u001b[0m/u001b[0m: consider adding a `main` function to `src/main.rs`/u001b[0m/n/n"}}
```
hi-rustin, line 117: I guess we only need to check `\\u001b[38;5;9merror` here?

tests/testsuite/locate_project.rs
weihanglo: I guess yes. That redaction should be preferred.

weihanglo: I also wonder if we want to use `str![].json()` for these JSON outputs. It can help future change that adding new fields, and we want to make the JSON pretty printed. Granted, it is just a one-liner so probably not really needed.

tests/testsuite/locate_project.rs
```diff
@@ -57,22 +67,33 @@ fn workspace() {
         .file("inner/src/lib.rs", "")
         .build();
 
-    let outer_manifest = r#"{"root": "[ROOT]/foo/Cargo.toml"}"#;
-    let inner_manifest = r#"{"root": "[ROOT]/foo/inner/Cargo.toml"}"#;
```
weihanglo, line 61: Guess it's also possible to do something alike? By extracting them as variables, we can make sure invocations that are ought to be the same still get the same JSON.

```suggestion
    let outer_manifest = str![[r#"{"root": "[ROOT]/foo/Cargo.toml"}"#]];
    let inner_manifest = str!r#"{"root": "[ROOT]/foo/inner/Cargo.toml"}"#]];
```

tests/testsuite/message_format.rs
```diff
@@ -66,17 +66,23 @@ fn cargo_renders() {
...
-             {\"reason\":\"build-finished\",\"success\":false}",
-        )
-        .with_stderr_contains(
-            "\
-[CHECKING] bar [..]
-[CHECKING] foo [..]
-error[..]`main`[..]
-",
-        )
+        .with_stdout_data(str![[r#"
```
weihanglo, line 69: This is something we want to avoid at this moment <https://github.com/rust-lang/cargo/issues/14039#issuecomment-2160827084>. Could you revert and annotate the function with `#[allow(deprecated)]`?

tests/testsuite/message_format.rs
```diff
@@ -108,7 +113,13 @@ fn cargo_renders_ansi() {
 
     p.cargo("check --message-format json-diagnostic-rendered-ansi")
         .with_status(101)
-        .with_stdout_contains("[..]\\u001b[38;5;9merror[..]")
+        .with_stdout_data(str![[r#"
```
weihanglo, line 116: ditto, for JSON messages we keep it as-is.

tests/testsuite/version.rs
```diff
@@ -1,23 +1,30 @@
...
-use cargo_test_support::{cargo_process, project};
+use cargo_test_support::{cargo_process, project, str};
 
 #[cargo_test]
 fn simple() {
     let p = project().build();
 
     p.cargo("version")
-        .with_stdout(&format!("cargo {}\n", cargo::version()))
+        .with_stdout_data(str![[r#"
```
weihanglo, line 10: It feel like the variable is important enough so should be kept with `.with_stdout(&format!("cargo {}\n", cargo::version()))` or similar. It also won't be a burden for manual updates, as this should be the only place we assert the version string.

See #14039 "4. Resolve non-literal deprecations"

tests/testsuite/version.rs
```diff
@@ -53,10 +60,16 @@ fn version_works_with_bad_target_dir() {
...
-        .with_stdout_contains("os: [..]")
+        .with_stdout_data(str![[r#"
+cargo [..] ([..] [..])
+release: [..]
+commit-hash: [..]
+commit-date: [..]
+host: [HOST_TARGET]
+libgit2: [..] (sys:[..] [..])
+libcurl: [..] (sys:[..] [..])
+os: [..]
```
weihanglo, line 71: This one is a bit tricky. SSL information might be there if people build cargo with `openssl` Cargo feature. The test will fail under that circumstance. 

https://github.com/rust-lang/cargo/blob/d899ce72ecff0a7486ca5efa5347f719e5a44214/src/bin/cargo/cli.rs#L260-L269

The official Rust toolchain distribution uses the [`all-static`](https://github.com/rust-lang/cargo/blob/d899ce72ecff0a7486ca5efa5347f719e5a44214/Cargo.toml#L258) Cargo feature so it won't affect any CI jobs ran in rust-lang/rust repo. However, some distro packagers manage their own dependencies and don't like vendoring openssl or libgit2. They might be impacted by this when running the test suite without `all-static` feature.

One way to solve this is using make the assertion unordered by calling `.unordered()`, and put `...` at the end of the literal. And the same we may want to remove `str![]` for this.

tests/testsuite/locate_project.rs
weihanglo: https://github.com/rust-lang/cargo/actions/runs/9564623389/job/26365780379?pr=14091#step:11:4113

Saw the failure, however confusing. It did work in other test case like [this](https://github.com/rust-lang/cargo/blob/83a3964417d01ec4abd4af919be46f851b3dc349/tests/testsuite/bench.rs#L2116). Not sure what happened.

weihanglo: Perhaps this is missing in snapbox?

https://github.com/rust-lang/cargo/blob/83a3964417d01ec4abd4af919be46f851b3dc349/crates/cargo-test-support/src/compare.rs#L320-L339

tests/testsuite/locate_project.rs
eth3lbert: It appears the suggested change might have a typo
```rust
let inner_manifest = str![[r#"{"root": "[ROOT]/foo/inner/Cargo.toml"}"#]];
```

tests/testsuite/locate_project.rs
henry40408: I merged the `master` branch and tried again at commit 2fbba53, but the CI still failed.

weihanglo: Thanks for the attempt. Let's leave off `locate_project` for now and focus on others. We can investigate it later.

henry40408: Should I revert the `locate_project` testsuite to the original version?

Specifically, to the base commit d899ce72e of this pull request on the `master` branch?

tests/testsuite/version.rs
henry40408: > One way to solve this is using make the assertion unordered by calling `.unordered()`, and put `...` at the end of the literal. And the same we may want to remove `str![]` for this.

Take `tests/testsuite/bench.rs` for example: isn't `str![]` needed by `.unordered()`?

https://github.com/rust-lang/cargo/blob/103e6758f11af08b7bd3bc6ebe7e7b270bfc7233/tests/testsuite/bench.rs#L1608-L1619

Is there any testsuite that already uses this technique that I can refer to?

tests/testsuite/locate_project.rs
weihanglo: Could also just annotate those failing test with `#[allow(deprecated)]` and move on

tests/testsuite/version.rs
weihanglo: True and not ture. [`unordered()`](https://docs.rs/snapbox/latest/snapbox/data/struct.Data.html#method.unordered) is a method of `Data`. Haven't tested it but you could probably do something like:

```rust
with_stdout_data(format!("<omit>").into().unordered()))
```

weihanglo: BTW, see https://github.com/rust-lang/cargo/issues/14039 "6. Contains / Does not Contains deprecations"

> A `contains` can be modeled by surrounding a entry with `...` (text) or `"...",` (json lines).  The question is whether we should still do this or use a different method.
> ....
> - Multiple `contains` can be merged into a single `eq` with `..` interspersed (or one `...` and `.unordered()`).  For example, see 3054936cab3d4627d5d1d878426c4fd88fab41f8
>   - When using `unordered`, likely string literals, rather than `str![]` should be used, as these can't correctly be updated

tests/testsuite/message_format.rs
henry40408: addressed at commit a1f4d03

tests/testsuite/message_format.rs
henry40408: addressed at commit a1f4d03

tests/testsuite/locate_project.rs
henry40408: addressed at commit 09fe7b8

tests/testsuite/version.rs
henry40408: addressed at commit 5d44b52

tests/testsuite/version.rs
henry40408: addressed at commit 5d44b52

tests/testsuite/version.rs
henry40408: Is `unordered` necessary? From the source code, the order of lines of version information should be idempotent:

https://github.com/rust-lang/cargo/blob/d899ce72ecff0a7486ca5efa5347f719e5a44214/src/bin/cargo/cli.rs#L208-L217

tests/testsuite/version.rs
```diff
@@ -53,10 +52,18 @@ fn version_works_with_bad_target_dir() {
...
+release: [..]
+commit-hash: [..]
+commit-date: [..]
+host: [HOST_TARGET]
+libgit2: [..] (sys:[..] [..])
+libcurl: [..] (sys:[..] [..])
+...
+"#]]
+            .unordered(),
+        )
```
weihanglo, line 67: Sorry but we're missing the assertion of `os` part. We could preserve the exact match of cargo version string, and omit ssl part with `...`, and drop `.unordered` call. What do you think?

```suggestion
        .with_stdout_data(format!(
            "\
cargo {}
release: [..]
commit-hash: [..]
commit-date: [..]
host: [HOST_TARGET]
libgit2: [..] (sys:[..] [..])
libcurl: [..] (sys:[..] [..])
...
os: [..]
",
            cargo::version()
        ))
```

weihanglo (commented):
Thanks. Just one thing left and we're ready to go.

tests/testsuite/version.rs
henry40408: Agreed.

henry40408: apply suggestion at commit 842579e

henry40408:
I'll tidy up commits after CI passes.

henry40408:
@weihanglo I've finished arranging the commits.

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #14092, opened 2024-06-18
Add `CodeFix::apply_solution` and impl `Clone`

Alexendoo:

In Clippy we have a good few lints that produce mutually exclusive suggestions e.g.

```
error: octal-looking escape in a literal
  --> tests/ui/octal_escapes.rs:6:19
   |
LL |     let _bad2 = b"\033[0m";
   |                   ^^^^
   |
help: if an octal escape is intended, use a hex escape instead
   |
LL |     let _bad2 = b"\x1b[0m";
   |                   ~~~~
help: if a null escape is intended, disambiguate using
   |
LL |     let _bad2 = b"\x0033[0m";
   |                   ~~~~~~
```

For these we have to disable rustfix tests since the suggestions are overlapping, this PR adds a method to `rustfix` that `ui_test` could use in order to produce multiple `.fixed` files, one for each alternative suggestion

### Additional information

It does not work for for multiple suggestions coming from a single subdiagnostic ([`Diag::span_suggestions`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/diagnostic/struct.Diag.html#method.span_suggestions)) e.g.

```
help: consider importing one of these items
  |
1 + use std::collections::HashMap;
  |
1 + use ahash::HashMap;
```

Solving this would be blocked on https://github.com/rust-lang/rust/issues/53934, on the Clippy side we only have one use of `span_suggestions` however so it's still very useful without this

The test cases use Clippy lints that I generated by setting the `parse_and_replace` test to use `clippy-driver` because of familiarity, if there's a rustc case that does multiple suggestions it would be good to go with that

### DIFF
--- a/crates/rustfix/src/lib.rs
+++ b/crates/rustfix/src/lib.rs
@@ -213,6 +213,7 @@ pub fn collect_suggestions<S: ::std::hash::BuildHasher>(
 /// 1. Feeds the source of a file to [`CodeFix::new`].
 /// 2. Calls [`CodeFix::apply`] to apply suggestions to the source code.
 /// 3. Calls [`CodeFix::finish`] to get the "fixed" code.
+#[derive(Clone)]
 pub struct CodeFix {
     data: replace::Data,
     /// Whether or not the data has been modified.
@@ -230,12 +231,18 @@ impl CodeFix {
 
     /// Applies a suggestion to the code.
     pub fn apply(&mut self, suggestion: &Suggestion) -> Result<(), Error> {
-        for sol in &suggestion.solutions {
-            for r in &sol.replacements {
-                self.data
-                    .replace_range(r.snippet.range.clone(), r.replacement.as_bytes())?;
-                self.modified = true;
-            }
+        for solution in &suggestion.solutions {
+            self.apply_solution(solution)?;
+        }
+        Ok(())
+    }
+
+    /// Applies an individual solution from a [`Suggestion`].
+    pub fn apply_solution(&mut self, solution: &Solution) -> Result<(), Error> {
+        for r in &solution.replacements {
+            self.data
+                .replace_range(r.snippet.range.clone(), r.replacement.as_bytes())?;
+            self.modified = true;
         }
         Ok(())
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2943,7 +2943,7 @@ checksum = "08d43f7aa6b08d49f382cde6a7982047c3426db949b1424bc4b7ec9ae12c6ce2"
 
 [[package]]
 name = "rustfix"
-version = "0.8.4"
+version = "0.8.5"
 dependencies = [
  "anyhow",
  "proptest",
--- a/crates/rustfix/Cargo.toml
+++ b/crates/rustfix/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rustfix"
-version = "0.8.4"
+version = "0.8.5"
 authors = [
     "Pascal Hertleif <killercup@gmail.com>",
     "Oliver Schneider <oli-obk@users.noreply.github.com>",


Alexendoo:
Ah I misread how the test suite worked, thought it only ran `rustc` on bless so I'll need to find some rustc examples that produce multiple suggestions

crates/rustfix/src/lib.rs
```diff
@@ -270,3 +280,34 @@ pub fn apply_suggestions(code: &str, suggestions: &[Suggestion]) -> Result<Strin
...
+///
+/// When a suggestion has multiple solutions that they are applied separately
+/// each output being the result of only applying the Nth solution. If a
+/// suggestion has fewer solutions than other suggestions the first solution
+/// is repeated.
+///
+/// Currently individual subdiagnostics with multiple solutions are not
+/// detected, they are treated as a single combined solution. See
+/// [rust-lang/rust#53934](https://github.com/rust-lang/rust/issues/53934).
+pub fn apply_suggestions_separately(
```
weihanglo, line 294: Do we consider duplicate solutions inside a suggestion? We should only apply once in that situation. See <https://github.com/rust-lang/cargo/pull/13728>.
That said, it can be considered a rustc bug and we do nothing here if the only consumer is clippy.

weihanglo (commented):
Thanks for the pull request. I saw you've changed the test snapshot to rustc diagnostics. Would you mind

* Providing issue links and code in Clippy of where Clippy disables rustfix tests.
* Sharing the Clippy diagnositc JSON which rustfix is not happy with.
* Also how Clippy is going to use that new public interface so we won't mess up stuff in the future.

That would make the intent clear and connected. Thank you!

Alexendoo:
I don't think we have an issue open for it but here's a few test examples with disabled rustfix due to having multiple suggestions

- https://github.com/rust-lang/rust-clippy/blob/29cc5c691c574b54b2fcfed51c22f8bb51110919/tests/ui/assign_ops2.rs#L1
- https://github.com/rust-lang/rust-clippy/blob/29cc5c691c574b54b2fcfed51c22f8bb51110919/tests/ui/borrow_deref_ref_unfixable.rs#L1
- https://github.com/rust-lang/rust-clippy/blob/29cc5c691c574b54b2fcfed51c22f8bb51110919/tests/ui/join_absolute_paths.rs#L1

https://github.com/rust-lang/cargo/blob/50fa303d5c3e61c95e1a87b5b12ef3de19d415db/crates/rustfix/tests/separate/separate.json is a good Clippy diagnostic example

The plan would be to add support to [`ui_test`](https://docs.rs/ui_test/latest/ui_test/) that Clippy uses for its tests, similar to the added tests here there would be multiple `fixed` files when suggestions contain multiple solutions

I just noticed though that since the fields of [`Suggestion`](https://docs.rs/rustfix/latest/rustfix/struct.Suggestion.html) are `pub` this could be done without an API change to `rustfix`. Do you think this should live in `rustfix` or be implemented outside? The `Clone`/`apply_solution` additions for `CodeFix` would still be useful if it's implemented outside

weihanglo:
Thanks for the information! Now I can clearly see where it came from :)

> I just noticed though that since the fields of [`Suggestion`](https://docs.rs/rustfix/latest/rustfix/struct.Suggestion.html) are `pub` this could be done without an API change to `rustfix`. Do you think this should live in `rustfix` or be implemented outside? 

Sounds that it's better being done on Clippy/ui_test side if it is the only consumer. It also reduces cross repo works if something needs to change in the futre.

> The `Clone`/`apply_solution` additions for `CodeFix` would still be useful if it's implemented outside

Yeah I am good with these additions.

Alexendoo:
A much smaller PR now 😄

weihanglo (approved):

weihanglo:
Thanks!

@bors+

BTW, do you know what [rust-cloud-vms](https://github.com/apps/rust-cloud-vms) is? I saw it force-pushed code you commited. Or, are you a bot?

Alexendoo:
🤖

It's because I pushed from one of the [dev desktops](https://forge.rust-lang.org/infra/docs/dev-desktop.html#dev-desktops), they have a [funky credential helper](https://forge.rust-lang.org/infra/docs/dev-desktop.html#how-to-interact-with-github) setup going on that I guess appears as though it's coming from the bot

PR #14106, opened 2024-06-19
Simplify checking feature syntax

Eh2406:

Similar to #14089, in my PubGrub work I had to try to understand the `build_feature_map` code that determines if the name refers to a feature or an optional dependency. There were a couple of little things I wanted to clean up while I was staring at the code. Specifically:
- there was a lot of vertical space taken up with arguments to `bail` that could be in line at the format string.
- some match statements were used where a named helper method would be clearer.
- `split_once` could replace a `find ... split_at` dance.
- in `dep_map` we were constructing a full `Vec<Dependency>`, when we only cared about whether there was any dependency and whether any of the dependencies were optional.

It's an internal re-factor, and the tests still pass.

### Additional information

If you're having questions about this code while you're reviewing, this would be a perfect opportunity for better naming and comments. Please ask questions.

@epage: After this PR I am likely to copy this code into my pubgrub tester. Are there other users who might be interested in looking at a cargo.toml or an index entry and determining what feature names are available and what dependencies they enable? If so maybe this function should be moved to one of the new stable-ish reusable libraries.

### DIFF
--- a/src/cargo/core/summary.rs
+++ b/src/cargo/core/summary.rs
@@ -156,13 +156,12 @@ fn build_feature_map(
     dependencies: &[Dependency],
 ) -> CargoResult<FeatureMap> {
     use self::FeatureValue::*;
-    let mut dep_map = HashMap::new();
+    // A map of dependency names to whether there are any that are optional.
+    let mut dep_map: HashMap<InternedString, bool> = HashMap::new();
     for dep in dependencies.iter() {
-        dep_map
-            .entry(dep.name_in_toml())
-            .or_insert_with(Vec::new)
-            .push(dep);
+        *dep_map.entry(dep.name_in_toml()).or_insert(false) |= dep.is_optional();
     }
+    let dep_map = dep_map; // We are done mutating this variable
 
     let mut map: FeatureMap = features
         .iter()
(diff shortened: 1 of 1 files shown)


src/cargo/core/summary.rs
```diff
@@ -403,25 +356,35 @@ impl FeatureValue {
...
-    pub fn has_dep_prefix(&self) -> bool {
-        matches!(self, FeatureValue::Dep { .. })
+    fn explicitly_name(&self) -> Option<InternedString> {
+        match self {
+            FeatureValue::Dep { dep_name, .. } => Some(*dep_name),
+            _ => None,
+        }
+    }
+
+    fn dep_name(&self) -> InternedString {
```
ehuss, line 366: I think this might be a confusing name, since `FeatureValue::Feature` is not necessarily a dependency name. For example:

```toml
[features]
foo = ["bar"]
bar = []
```

Here the `FeatureValue::Feature("bar")` is not a dependency name.

I realize this is refactoring code that was already using `dep_name`, but the original was somewhat abusing the nomenclature.

I would go with `feature_or_dep_name` to make it clearer it can be either. What do you think?

src/cargo/core/summary.rs
```diff
@@ -403,25 +356,35 @@ impl FeatureValue {
         }
     }
 
-    /// Returns `true` if this feature explicitly used `dep:` syntax.
-    pub fn has_dep_prefix(&self) -> bool {
-        matches!(self, FeatureValue::Dep { .. })
+    fn explicitly_name(&self) -> Option<InternedString> {
```
ehuss, line 359: Here `name` seems ambiguous to me. It could be a "feature name" or a "dependency name". I might go with something like `explicit_dep_name` or something like that.

Also, can you add a docstring to say exactly what it does?

ehuss (commented):
Looks good, just some bikeshedding on the name.

epage:
`cargo add` and `cargo-information do their # own feature activation work.

Eh2406:
@ehuss tweak names as requested.

@epage are you saying that other projects would or would not find making this method public useful?

epage:
I've not checked the function logic to see if it would work (mostly out for the next couple weeks) so I was just noting potential applications, one being in Cargo.

ehuss:
Thanks!

@bors r+

Issue #13922, opened 2024-05-16
Build host architecture is written into metadata hash for compilation units, breaking reproducibility of output between host platforms


kanavin:
### Problem

hash_rustc_version() writes rustc().verbose_version into the hash. That data has one problematic field:

host: x86_64-unknown-linux-gnu
or
host: aarch64-unknown-linux-gnu

Due to this, when one is using a mix of x86 and aarch build hosts with exactly same rust compiler and building for the same cross-target, the output becomes non-reproducible, and differs between the two; even the file names become different.

### Steps

This requries two build hosts that have a different architecture, e.g. x86 and aarch, and running an identical cross-compile on them, using exact same host rust compiler. The output is going to be different (even in the filenames if those include hashes), even though it should be the same.

### Possible Solution(s)

The situation occurred in the context of Yocto project builds, where we have a cluster of build machines (a mix of x86 and arm64), the rust version in use is tightly controlled, and we expect the output to be the same (and check for it). The quick and hacky solution was to patch out the problematic lines in cargo, as shown in the attached patch, but perhaps a better option would be to not write the host architecture into the hash when the build is a cross one.

```
From 065d7c263091118437465d714d8a29dbb6296921 Mon Sep 17 00:00:00 2001
From: Alexander Kanavin <alex@linutronix.de>
Date: Mon, 13 May 2024 14:57:54 +0200
Subject: [PATCH] cargo: do not write host information into compilation unit
 hashes

This breaks reproducibility in cross-builds where the cross-target
can be the same, but build hosts are different, as seen with
"rustc --version -v":
...
host: x86_64-unknown-linux-gnu

vs.

host: aarch64-unknown-linux-gnu

This can possibly be improved by only hashing host info if the build
is a native one (e.g. there's no --target option passed to cargo
invocation) but I'm not sure how.

Signed-off-by: Alexander Kanavin <alex@linutronix.de>
---
 .../src/cargo/core/compiler/context/compilation_files.rs      | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/src/tools/cargo/src/cargo/core/compiler/context/compilation_files.rs b/src/tools/cargo/src/cargo/core/compiler/context/compilation_files.rs
index d83dbf10c..b2ad8d9f3 100644
--- a/src/tools/cargo/src/cargo/core/compiler/context/compilation_files.rs
+++ b/src/tools/cargo/src/cargo/core/compiler/context/compilation_files.rs
[...]

Issue #8140, opened 2020-04-21
Non-reproducible -C metadata=hash passed to rustc depending on the compiling OS


gendx:
**Problem**

[Tock](https://github.com/tock/tock) is an operating system written in Rust for embedded platforms (e.g. `thumbv7em-none-eabi` target). We've been trying to make builds reproducible, with good progress (https://github.com/tock/tock/issues/1666). In particular:
- pinning a version of the Rust compiler,
- using `--remap-path-prefix`,
- using a Cargo workspace so that local dependencies are identified by a relative path (Tock is split into multiple crates inside the workspace),
- using `cargo rustc` to avoid passing custom linker arguments (that include paths on the filesystem) to the dependencies.

With these we've managed to make the builds reproducible across various Linux machines.

However, the builds are different on Linux (`nightly-2020-02-03-x86_64-unknown-linux-gnu`) and MacOS (`nightly-2020-02-03-x86_64-apple-darwin`), as evidenced by CI on a GitHub workflow (https://github.com/google/OpenSK/pull/94#issuecomment-616671549).

In particular, the issue seems to stem from a **different `-C metadata=` (and `-C extra-filename=`) passed to `rustc`** (while the steps that I mentioned above make sure that the same `metadata` hash is passed across Linux machines for each crate in the project).

These are built with the `--verbose` parameter passed to `cargo`, and an example `rustc` invocation is the following (`tock_cells` crate of the Tock project).

[...]

PR #14107, opened 2024-06-19
Make `-Cmetadata` consistent across platforms

alexcrichton:
This commit updates how `-Cmetadata` is calculated for each unit to optionally exclude the `host: ...` line that rustc prints for cross-compiled builds. Previously the full `verbose_version` was hashed for stable builds and the `host` was explicitly hashed for non-stable builds. For a build using `--target`, however, that means that the `-Cmetadata` will be different when producing the same target on different hosts (e.g. producing the binary once on Linux and once on macOS).

This can hinder reproduction of a binary across different platforms even when the `--target` flag is used. For example in rust-lang/rust#117597 it was seen that a WebAssembly binary produced on different platforms was slightly different and this appears due to the differing `-Cmetadata` flags. After this commit the `-Cmetadata` flag is the same for two different platforms meaning that different platforms produce the same binary.

I've tested locally and a simple project produces a different binary before this change but produces the same binary on two platforms after this change. Unfortunately automated testing of this change will be difficult since it requires two different host compilers, though.

### DIFF
--- a/src/cargo/core/compiler/build_runner/compilation_files.rs
+++ b/src/cargo/core/compiler/build_runner/compilation_files.rs
@@ -620,7 +620,7 @@ fn compute_metadata(
     unit.target.name().hash(&mut hasher);
     unit.target.kind().hash(&mut hasher);
 
-    hash_rustc_version(bcx, &mut hasher);
+    hash_rustc_version(bcx, &mut hasher, unit);
 
     if build_runner.bcx.ws.is_member(&unit.pkg) {
         // This is primarily here for clippy. This ensures that the clippy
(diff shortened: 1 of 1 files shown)


weihanglo:
Does this fix the same issues in <https://github.com/rust-lang/cargo/issues/8140> and <https://github.com/rust-lang/cargo/issues/13922>?

alexcrichton:
Ah good catch! I suspect this fixes #13922 but I think there's more concerns brought up in #8140 so while this may fix some of the issues there I don't think it would fix all of them

ehuss:
Thanks! I'm not 100% certain that different host compilers will always produce the exact same output, but I imagine that should be considered a bug in rustc if it does not.

@bors r+

BlackAsLight:
A bit off topic, but how does one know when this merge will show up in the stable or nightly release?

bjorn3:
I believe cargo gets updated on nightly once a week or so. And from there it will ride the release train to stable, which can take anywhere from 6-12 weeks depending on where in the release cycle a change gets merged.

alexcrichton:
> Thanks! I'm not 100% certain that different host compilers will always produce the exact same output, but I imagine that should be considered a bug in rustc if it does not.

I definitely agree yeah. Getting deterministic output across different host platforms for the same target to me is basically a big game of whack-a-mole. This PR is probably one piece of the puzzle but I also have the same hunch that there are many other pieces for completely full fidelity.

joshtriplett:
Adding the `relnotes` label here. Sketch of release notes:

The Rust compiler now uses the same metadata for symbol ABI compatibility regardless of the host the Rust compiler was compiled for. This makes it possible for cross-compiles to the same target from different hosts to produce identical binaries. (There may still be other issues with doing so, which we hope to resolve over time.)

PR #14111, opened 2024-06-19
test: migrate weak_dep_features, workspaces and yank to snapbox

eth3lbert:

Migrate following to snapbox:

- `tests/testsuite/weak_dep_features.rs`
- `tests/testsuite/workspaces.rs`
- `tests/testsuite/yank.rs`

### DIFF
--- a/tests/testsuite/workspaces.rs
+++ b/tests/testsuite/workspaces.rs
@@ -1,8 +1,7 @@
 //! Tests for workspaces.
 
-#![allow(deprecated)]
-
 use cargo_test_support::registry::Package;
+use cargo_test_support::str;
 use cargo_test_support::{basic_lib_manifest, basic_manifest, git, project, sleep_ms};
 use std::env;
 use std::fs;
@@ -112,22 +111,20 @@ fn non_virtual_default_members_build_other_member() {
         .build();
 
     p.cargo("check")
-        .with_stderr(
-            "\
+        .with_stderr_data(str![[r#"
 [LOCKING] 3 packages to latest compatible versions
-[CHECKING] baz v0.1.0 ([..])
-[..] Finished `dev` profile [unoptimized + debuginfo] target(s) in [..]
-",
-        )
+[CHECKING] baz v0.1.0 ([ROOT]/foo/baz)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
         .run();
 
     p.cargo("check --manifest-path bar/Cargo.toml")
-        .with_stderr(
-            "\
-[CHECKING] bar v0.1.0 ([..])
-[..] Finished `dev` profile [unoptimized + debuginfo] target(s) in [..]
-",
-        )
+        .with_stderr_data(str![[r#"
+[CHECKING] bar v0.1.0 ([ROOT]/foo/bar)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
         .run();
 }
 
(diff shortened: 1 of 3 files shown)


tests/testsuite/workspaces.rs
```diff
@@ -430,14 +424,13 @@ fn invalid_parent_pointer() {
...
-            "\
-error: failed to read `[..]Cargo.toml`
+        .with_stderr_data(str![[r#"
+[ERROR] failed to read `[ROOT]/foo/foo/Cargo.toml`
 
 Caused by:
-  [..]
-",
-        )
+  [..] (os error [..])
```
epage, line 431: Why was this glob made more specific?

tests/testsuite/workspaces.rs
```diff
@@ -1251,7 +1230,14 @@ fn rebuild_please() {
...
         .with_status(101)
-        .with_stderr_contains("[..]assertion[..]")
+        .with_stderr_data(str![[r#"
+...
+assertion `left == right` failed
+  left: 1
+ right: 0
+[NOTE] run with `RUST_BACKTRACE=1` environment variable to display a backtrace
+...
+"#]])
```
epage, line 1240: We try to minimize rustc/std output that we match because we support multiple versions of rustc and changing rustc's behavior can be more difficult when cargo tests are dependent on it.  We should probably match on as little as possible.

For some cases, we don't even think about this.  I'm only thinking about this one because the output recently changed.

tests/testsuite/workspaces.rs
eth3lbert: Originally, I attempted a more specific approach without using any globs. However, this resulted in different error messages on different platforms https://github.com/rust-lang/cargo/actions/runs/9587495931/job/26437670924#step:11:4084.

tests/testsuite/workspaces.rs
eth3lbert: Ok, I'll update to use the original matching pattern.

tests/testsuite/workspaces.rs
epage: We have an auto-redaction for `[BROKEN_PIPE]` that matches the text for each platform's broken pipe message.  We could do something similar here.  As this is a common error message, auto-redacting it would likely help a lot!

eth3lbert: Sounds good!

One question: How should we handle situations where the same error occurs on Darwin/Linux but potentially results in different errors on Windows?

e.g.
https://github.com/rust-lang/cargo/actions/runs/9587495931/job/26437670924#step:11:4065
```
   7    7 | Caused by:
   8      -   No such file or directory (os error 2)
        8 +   The system cannot find the file specified. (os error 2)
```
https://github.com/rust-lang/cargo/actions/runs/9587495931/job/26437670924#step:11:4084
```
   7    7 | Caused by:
   8      -   No such file or directory (os error 2)
        8 +   The system cannot find the path specified. (os error 3)
 ```

Comment (weihanglo) [reply tests/testsuite/workspaces.rs:431]: There is another one https://github.com/rust-lang/cargo/pull/14068/files#r1639170882

Comment (epage) [reply tests/testsuite/workspaces.rs:431]: See https://github.com/rust-lang/cargo/blob/ce90393a11758106040628f77568337b1a5b3a0e/crates/cargo-test-support/src/compare.rs#L202-L203

Define the redaction multiple times, once for each unique text string.  This should be in `MIN_LITERAL_REDACTIONS`

**Comment by eth3lbert** [comment]
Rebased and feedback addressed!

Additionally, I've replaced an empty `str![]` with `""`.
```diff
diff --git a/tests/testsuite/weak_dep_features.rs b/tests/testsuite/weak_dep_features.rs
index f5d2e18e5..da281bfe3 100644
--- a/tests/testsuite/weak_dep_features.rs
+++ b/tests/testsuite/weak_dep_features.rs
@@ -517,7 +517,7 @@ foo v0.1.0 ([ROOT]/foo)
     // This is a little strange in that it produces no output.
     // Maybe `cargo tree` should print a note about why?
     p.cargo("tree -e features -i bar --features bar?/feat")
-        .with_stdout_data(str![])
+        .with_stdout_data("")
         .run();
 
     p.cargo("tree -e features -i bar --features bar?/feat,bar")
```

Thanks for the review! :)

PR #14096, opened 2024-06-18
test: migrate clean to snapbox

dieterplex:
Part of #14039, migrating `tests/testsuite/clean.rs` to snapbox.

PR #14113, opened 2024-06-20
test: migrate some files to snapbox 

heisen-li:

Changes to the following documents were completed:

`tests/testsuite/cargo_command.rs`
`tests/testsuite/cargo_env_config.rs`
`tests/testsuite/cargo_features.rs`
`tests/testsuite/cargo_targets.rs`

part of #14039

### DIFF
--- a/tests/testsuite/cargo_command.rs
+++ b/tests/testsuite/cargo_command.rs
@@ -1,7 +1,5 @@
 //! Tests for custom cargo commands and other global command features.
 
-#![allow(deprecated)]
-
 use std::env;
 use std::fs;
 use std::io::Read;
@@ -12,6 +10,7 @@ use std::str;
 use cargo_test_support::basic_manifest;
 use cargo_test_support::paths::CargoPathExt;
 use cargo_test_support::registry::Package;
+use cargo_test_support::str;
 use cargo_test_support::tools::echo_subcommand;
 use cargo_test_support::{
     basic_bin_manifest, cargo_exe, cargo_process, paths, project, project_in_home,
@@ -26,25 +25,23 @@ fn path() -> Vec<PathBuf> {
 fn list_commands_with_descriptions() {
     let p = project().build();
     p.cargo("--list")
-        .with_stdout_contains(
-            "    build                Compile a local package and all of its dependencies",
+        .with_stdout_data(
+            "\
+...
+    b                    alias: build
+...
+    build                Compile a local package and all of its dependencies
+...
+    c                    alias: check
+...
+    r                    alias: run
+...
+    read-manifest        Print a JSON representation of a Cargo.toml manifest.
+...
+    t                    alias: test
+...
+",
         )
-        // Assert that `read-manifest` prints the right one-line description followed by another
-        // command, indented.
-        .with_stdout_contains(
-            "    read-manifest        Print a JSON representation of a Cargo.toml manifest.",
-        )
-        .run();
-}
-
-#[cargo_test]
-fn list_builtin_aliases_with_descriptions() {
-    let p = project().build();
-    p.cargo("--list")
-        .with_stdout_contains("    b                    alias: build")
-        .with_stdout_contains("    c                    alias: check")
-        .with_stdout_contains("    r                    alias: run")
-        .with_stdout_contains("    t                    alias: test")
         .run();
 }
 
@@ -62,8 +59,12 @@ fn list_custom_aliases_with_descriptions() {
         .build();
 
     p.cargo("--list")
-        .with_stdout_contains("    myaliasstr           alias: foo --bar")
-        .with_stdout_contains("    myaliasvec           alias: foo --bar")
+        .with_stdout_data(str![[r#"
+...
+    myaliasstr           alias: foo --bar
+    myaliasvec           alias: foo --bar
+...
+"#]])
         .run();
 }
 
@@ -81,7 +82,11 @@ fn list_dedupe() {
 
     p.cargo("--list")
         .env("PATH", &path)
-        .with_stdout_contains_n("    dupe", 1)
+        .with_stdout_data(str![[r#"
+...
+    dupe
+...
+"#]])
         .run();
 }
 
@@ -132,6 +137,7 @@ fn list_command_looks_at_path_case_mismatch() {
     );
 }
 
+#[allow(deprecated)]
 #[cargo_test]
 fn list_command_handles_known_external_commands() {
     let p = project()
@@ -153,7 +159,10 @@ fn list_command_handles_known_external_commands() {
 
     p.cargo("--list")
         .env("PATH", &path)
-        .with_stdout_contains(fmt_desc)
+        .with_stdout_data(str![[r#"
+...
+    fmt                  Formats all bin and lib files of the current crate using rustfmt.
+..."#]])
         .run();
 }
 
@@ -182,13 +191,15 @@ fn list_command_resolves_symlinks() {
 fn find_closest_capital_c_to_c() {
     cargo_process("C")
         .with_status(101)
-        .with_stderr_contains(
-            "\
-error: no such command: `C`
+        .with_stderr_data(str![[r#"
+[ERROR] no such command: `C`
 
-<tab>Did you mean `c`?
-",
-        )
+	Did you mean `c`?
+
+	View all installed commands with `cargo --list`
+	Find a package to install `C` with `cargo search cargo-C`
+
+"#]])
         .run();
 }
 
@@ -196,13 +207,15 @@ error: no such command: `C`
 fn find_closest_capital_b_to_b() {
     cargo_process("B")
         .with_status(101)
-        .with_stderr_contains(
-            "\
-error: no such command: `B`
+        .with_stderr_data(str![[r#"
+[ERROR] no such command: `B`
 
-<tab>Did you mean `b`?
-",
-        )
+	Did you mean `b`?
+
+	View all installed commands with `cargo --list`
+	Find a package to install `B` with `cargo search cargo-B`
+
+"#]])
         .run();
 }
 
@@ -210,13 +223,15 @@ error: no such command: `B`
 fn find_closest_biuld_to_build() {
     cargo_process("biuld")
         .with_status(101)
-        .with_stderr_contains(
-            "\
-error: no such command: `biuld`
+        .with_stderr_data(str![[r#"
+[ERROR] no such command: `biuld`
 
-<tab>Did you mean `build`?
-",
-        )
+	Did you mean `build`?
+
+	View all installed commands with `cargo --list`
+	Find a package to install `biuld` with `cargo search cargo-biuld`
+
+"#]])
         .run();
 
     // But, if we actually have `biuld`, it must work!
@@ -234,13 +249,17 @@ error: no such command: `biuld`
 
     cargo_process("install cargo-biuld").run();
     cargo_process("biuld")
-        .with_stdout("Similar, but not identical to, build\n")
+        .with_stdout_data(str![[r#"
+Similar, but not identical to, build
+
+"#]])
         .run();
     cargo_process("--list")
-        .with_stdout_contains(
-            "    build                Compile a local package and all of its dependencies\n",
-        )
-        .with_stdout_contains("    biuld\n")
+        .with_stdout_data(str![[r#"
+...
+    biuld
+    build                Compile a local package and all of its dependencies
+..."#]])
         .run();
 }
 
@@ -250,7 +269,7 @@ fn find_closest_alias() {
     let my_home = root.join("my_home");
     fs::create_dir(&my_home).unwrap();
     fs::write(
-        &my_home.join("config"),
+        &my_home.join("config.toml"),
         r#"
             [alias]
             myalias = "build"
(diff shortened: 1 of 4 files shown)


tests/testsuite/cargo_command.rs
```diff
@@ -26,26 +25,65 @@ fn path() -> Vec<PathBuf> {
...
-        // Assert that `read-manifest` prints the right one-line description followed by another
-        // command, indented.
-        .with_stdout_contains(
-            "    read-manifest        Print a JSON representation of a Cargo.toml manifest.",
-        )
-        .run();
-}
-
-#[cargo_test]
-fn list_builtin_aliases_with_descriptions() {
```
heisen-li, line 41: Since the output on the right is complete, the test case here is deleted.

tests/testsuite/cargo_command.rs
```diff
@@ -132,6 +178,7 @@ fn list_command_looks_at_path_case_mismatch() {
     );
 }
 
+#[allow(deprecated)]
```
heisen-li, line 181: Since `with_stdout_does_not_contain`，add `#[allow(deprecated)]`.

tests/testsuite/cargo_env_config.rs
```diff
@@ -231,16 +250,29 @@ fn env_applied_to_target_info_discovery_rustc() {
...
-        .with_stderr_contains("MAIN ENV_TEST:from-env")
+        .with_stderr_data(str![[r#"
+[COMPILING] foo v0.5.0 ([ROOT]/foo)
+WRAPPER ENV_TEST:from-env
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+[RUNNING] `target/debug/foo`
+MAIN ENV_TEST:from-env
+
+"#]])
+        // .with_stderr_contains("MAIN ENV_TEST:from-env")
```
heisen-li, line 276: Forgot to delete the comment here, will do so later, sorry.

heisen-li: deleted

crates/cargo-test-support/src/compare.rs
```diff
@@ -174,6 +176,11 @@ fn add_common_redactions(subs: &mut snapbox::Redactions) {
         regex!(r"home/\.cargo/registry/src/-(?<redacted>[a-z0-9]+)"),
     )
     .unwrap();
+    subs.insert(
+        "[VERSION]",
+        regex!(r"(?<redacted>\d+\.\d+\.\d+ \([a-f0-9]{9}+ \d{4}-\d{2}-\d{2}\))"),
+    )
```
epage, line 182: This feels too general.  Should we do `[CARGO_VERSION]` redacts `cargo::version()`?

heisen-li: Need to convert the type.

crates/cargo-test-support/src/compare.rs
```diff
@@ -146,6 +147,7 @@ fn add_common_redactions(subs: &mut snapbox::Redactions) {
         .unwrap();
     subs.insert("[ROOT]", root).unwrap();
     subs.insert("[ROOTURL]", root_url).unwrap();
+    subs.insert("[CARGO]", cargo).unwrap();
```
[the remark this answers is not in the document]
heisen-li: It's here:https://github.com/rust-lang/cargo/pull/14113/files#diff-9e3e31cc30eb728fcb75dcddf441dc127b0eb8cf764b4ce6cc260c3ec9487c26R393

tests/testsuite/cargo_command.rs
```diff
@@ -381,11 +401,13 @@ fn cargo_subcommand_env() {
         .with_extension(std::env::consts::EXE_EXTENSION)
         .canonicalize()
         .unwrap();
-    let envtest_bin = envtest_bin.to_str().unwrap();
     cargo_process("envtest")
         .env("PATH", &path)
         .env(cargo::CARGO_ENV, &envtest_bin)
-        .with_stdout(envtest_bin)
+        .with_stdout_data(str![[r#"
+[ROOT]/cargo-envtest/target/debug/cargo-envtest[EXE]
```
heisen-li, line 409: There may have to be an error here I haven't figured out what it is yet. Any suggestions?
```
expected: [ROOT]/cargo-envtest/target/debug/cargo-envtest[EXE]
actually  :\\?\[ROOT]/cargo-envtest/target/debug/cargo-envtest[EXE]
```

weihanglo: See <https://github.com/rust-lang/cargo/pull/14096#discussion_r1653536860>.

epage:
Could you clean up the commit history?

tests/testsuite/cargo_command.rs
```diff
@@ -365,14 +383,16 @@ fn cargo_subcommand_env() {
...
     path.push(target_dir.clone());
     let path = env::join_paths(path.iter()).unwrap();
 
     cargo_process("envtest")
         .env("PATH", &path)
-        .with_stdout(cargo.to_str().unwrap())
+        .with_stdout_data(str![[r#"
+[CARGO_PATH]
+
+"#]])
```
epage, line 395: So this was already a "self updating" test output and I believe the design of it means that there is little to evolve that we'd need to allow snapshotting more from this.

Should we keep the expected-str generation that was in here before, just switching the assert method used?

tests/testsuite/version.rs
```diff
@@ -1,21 +1,30 @@
...
 #[cargo_test]
 fn simple() {
     let p = project().build();
 
     p.cargo("version")
-        .with_stdout_data(&format!("cargo {}\n", cargo::version()))
+        .with_stdout_data(str![[r#"
+cargo [CARGO_VERSION]
+
+"#]])
```
epage, line 13: ditto about maintaining the old expected-value and just updating the assert method

tests/testsuite/cargo_command.rs
heisen-li: It is not possible to just convert `with_stdout` to `with_stdout_data`, such as the error reported below:
```
--- Expected
++++ actual:   stdout
   1      - /home/heisen/temp/temp/temp/cargo/target/debug/cargo∅
        1 + /home/heisen/temp/temp/temp/cargo/target/debug/cargo
```

tests/testsuite/cargo_command.rs
heisen-li: > See [#14096 (comment)](https://github.com/rust-lang/cargo/pull/14096#discussion_r1653536860).

Thanks for the suggestion. If nothing better, use `#[allow(deprecated)]` for now?

tests/testsuite/cargo_command.rs
epage: Adding in a newline is trivial and doesn't invalidate the idea I mentioned that this is unlikely to have surrounding text that gets regularly changed.

heisen-li: It has been modified and you can see if it meets the requirements.

tests/testsuite/cargo_command.rs
```diff
@@ -252,69 +271,68 @@ fn find_closest_alias() {
     fs::write(
         &my_home.join("config"),
```
weihanglo, line 272: Would you mind updating to `config.toml` so that we can avoid the "[ROOT]/my_home/config is deprecated in favor of config.toml" warning.

```suggestion
        &my_home.join("config.toml"),
```

tests/testsuite/cargo_command.rs
```diff
@@ -326,9 +344,9 @@ fn override_cargo_home() {
...
         r#"
-            [cargo-new]
-            vcs = "none"
-        "#,
+			[cargo-new]
```
weihanglo, line 347: Why whitespaces changes?

tests/testsuite/cargo_command.rs
```diff
@@ -365,14 +383,16 @@ fn cargo_subcommand_env() {
...
 
-    let cargo = cargo_exe().canonicalize().unwrap();
     let mut path = path();
     path.push(target_dir.clone());
     let path = env::join_paths(path.iter()).unwrap();
 
     cargo_process("envtest")
         .env("PATH", &path)
-        .with_stdout(cargo.to_str().unwrap())
+        .with_stdout_data(str![[r#"
```
weihanglo, line 392: This asserts an exact value. I don't think we need a snapshot here.

```suggestion
        .with_stdout_data(cargo.to_str())
```

tests/testsuite/cargo_command.rs
```diff
@@ -385,7 +405,10 @@ fn cargo_subcommand_env() {
     cargo_process("envtest")
         .env("PATH", &path)
         .env(cargo::CARGO_ENV, &envtest_bin)
-        .with_stdout(envtest_bin)
+        .with_stdout_data(str![[r#"
```
weihanglo, line 408: ditto. Just check the value against `envtest_bin`.

tests/testsuite/cargo_command.rs
```diff
@@ -477,16 +517,43 @@ fn cargo_subcommand_args() {
...
+    _ => {/* ... */} // All possible cases have already been handled
+}
+```
+
+`match` blocks have their patterns matched in order, so, for example, putting
+a wildcard arm above a more specific arm will make the latter arm irrelevant.
+
+Ensure the ordering of the match arm is correct and remove any superfluous
+arms.
+
```
weihanglo, line 555: The error is from rustc and subject to change. 

```suggestion
...
[..]This error suggests that the expression arm corresponding to the noted pattern[..]
...
```

tests/testsuite/cargo_command.rs
heisen-li: Sorry, this modification doesn't work, I tried several different modification options and got different results. I would like to describe it in detail:
1.Try modifying it to：`.with_stdout_data(cargo.to_str().unwrap())`;
The results obtained are：
```
--- Expected
++++ actual:   stdout
   1      - /home/heisen/temp/temp/temp/cargo/target/debug/cargo∅
        1 + /home/heisen/temp/temp/temp/cargo/target/debug/cargo
```
Line breaks need to be added. If line breaks are added after: `.with_stdout_data(format!("{}\n",cargo.to_str().unwrap())`, then the suffix `[EXE]` needs to be handled. If `[EXE]` is added, then an error occurs:
```
--- Expected
++++ actual:   stdout
   1      - \\?\D:\Rust\temp\temp\temp\cargo\target\debug\cargo.exe[EXE]
        1 + //?/D:/Rust/temp/temp/temp/cargo/target/debug/cargo[EXE]
```
2.Try changing it to `.with_stdout_data(cargo.to_str())`; it gives an error:
```
error[E0277]: the trait bound `std::option::Option<&str>: snapbox::IntoData` is not satisfied
```
3. Trying to change to `.with_stdout_data(envtest_bin)` in the following will result in an error:
```
--- Expected
++++ actual:   stdout
   1      - \\?\D:\Rust\temp\temp\temp\cargo\target\tmp\cit\t0\cargo-envtest\target\debug\cargo-envtest.exe∅
        1 + //?/[ROOT]/cargo-envtest/target/debug/cargo-envtest[EXE]
```

All I can think of at the moment is to use the current modification or [add `#[allow(deprecated)]`](https://github.com/rust-lang/cargo/pull/14113#discussion_r1653987076).

tests/testsuite/cargo_command.rs
```diff
@@ -26,44 +25,46 @@ fn path() -> Vec<PathBuf> {
...
             &paths::home().join(".cargo").join("config"),
             r#"
-            [alias]
-            myaliasstr = "foo --bar"
-            myaliasvec = ["foo", "--bar"]
-        "#,
+			[alias]
+			myaliasstr = "foo --bar"
+			myaliasvec = ["foo", "--bar"]
+		"#,
```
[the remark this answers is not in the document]
heisen-li: This one is actually caused by the use of `tap` instead of spaces.

tests/testsuite/cargo_command.rs
heisen-li: done

tests/testsuite/cargo_command.rs
heisen-li: Ditto. It's normal in my local area.

tests/testsuite/cargo_command.rs
heisen-li: modified

tests/testsuite/cargo_command.rs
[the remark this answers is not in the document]
weihanglo: Could you revert these kinds of change? They are not too relevant to this PR.

If there is a reason to convert spaces to tabs, open a new issue.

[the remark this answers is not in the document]
heisen-li: Thankfully, I believe it has been restored.

tests/testsuite/cargo_command.rs
```diff
@@ -341,6 +357,7 @@ fn override_cargo_home() {
     assert!(paths::root().join("foo2/.git").is_dir());
 }
 
+#[allow(deprecated)]
```
weihanglo, line 360: Note: https://github.com/rust-lang/cargo/pull/14113#discussion_r1661273136

weihanglo (approved):

weihanglo:
Looks good. Thanks!


@bors r+

PR #14121, opened 2024-06-21
test: Auto-redact file number

dieterplex:
This is from <https://github.com/rust-lang/cargo/pull/14096#discussion_r1648889843>.

Although the number of files in `cargo package` is important,
we have `validate_crate_contents` and `validate_upload_with_contents`
that verify the exact contents.
Redacting `Packaged` status should be fine.

### DIFF
--- a/crates/cargo-test-support/src/compare.rs
+++ b/crates/cargo-test-support/src/compare.rs
@@ -165,6 +165,11 @@ fn add_common_redactions(subs: &mut snapbox::Redactions) {
     )
     .unwrap();
     subs.insert(
+        "[FILE_NUM]",
+        regex!(r"\[(PACKAGED|REMOVED|SUMMARY)\] (?<redacted>[0-9]+) files"),
+    )
+    .unwrap();
+    subs.insert(
         "[FILE_SIZE]",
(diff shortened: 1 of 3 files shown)


weihanglo (approved):

epage:
Why is this being redacted?  I am not seeing any motivation in this PR or the linked comment?

weihanglo:
See <https://github.com/rust-lang/cargo/pull/14096#discussion_r1644832897>. The PR description is updated as well <https://github.com/rust-lang/cargo/pull/14121>.

epage:
I haven't really thought through the `cargo clean` case to really  say on that, I disagree about `cargo package`.  At least for now, I would like to lean towards redacting less, rather than redacting more.  Currently, the porting instructions are very limited in only suggesting auto-redactions for machine-specific information.  We have snapshot updating which removes the burden of updating the tests.

weihanglo:
We have an imperfect CI job that needs to accommodate to stable, beta, and nightly toolchains. `cargo package` is controlled by Cargo so probably safe not to redact. Artifacts emitted by rustc are not.

epage:
Posted #14123 for the `cargo package` side of this.

PR #14126, opened 2024-06-22
test: migrate rustflags, rustup and script to snapbox

eth3lbert:

Migrate following to snapbox:

- `tests/testsuite/rustflags.rs`
- `tests/testsuite/rustup.rs`
- `tests/testsuite/script.rs`

### DIFF
--- a/tests/testsuite/rustflags.rs
+++ b/tests/testsuite/rustflags.rs
@@ -1,9 +1,7 @@
 //! Tests for setting custom rustc flags.
 
-#![allow(deprecated)]
-
 use cargo_test_support::registry::Package;
-use cargo_test_support::{basic_manifest, paths, project, project_in_home, rustc_host};
+use cargo_test_support::{basic_manifest, paths, project, project_in_home, rustc_host, str};
 use std::fs;
 
 #[cargo_test]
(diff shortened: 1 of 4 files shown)


tests/testsuite/rustflags.rs
```diff
@@ -1237,7 +1456,12 @@ fn env_rustflags_misspelled() {
     for cmd in &["check", "build", "run", "test", "bench"] {
         p.cargo(cmd)
             .env("RUST_FLAGS", "foo")
-            .with_stderr_contains("[WARNING] Cargo does not read `RUST_FLAGS` environment variable. Did you mean `RUSTFLAGS`?")
+            .with_stderr_data(
+                "\
+[WARNING] Cargo does not read `RUST_FLAGS` environment variable. Did you mean `RUSTFLAGS`?
+...
+",
+            )
```
eth3lbert, line 1464: Ideally, it's better to use the specific output without globbing. However, the output varies depending on the cmd used. Alternatively, we could simply check it separately if necessary.

tests/testsuite/script.rs
```diff
@@ -30,19 +30,18 @@ fn basic_rs() {
...
+"#]])
+        .with_stderr_data(str![[r#"
 [WARNING] `package.edition` is unspecified, defaulting to `2021`
 [COMPILING] echo v0.0.0 ([ROOT]/foo)
-[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [..]s
-[RUNNING] `[..]/debug/echo[EXE]`
-",
-        )
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+[RUNNING] `[ROOT]/[..]/debug/echo[EXE]`
```
epage, line 42: How come `echo` needs globbing?

eth3lbert: We're globbing it because the path looks like
```
[RUNNING] `[ROOT]/home/.cargo/target/a4/7b05b555ee1ab6/debug/echo`
```

epage: We have a `[HASH]` redaction.  Could you extend it to include this?  Something like `r".cargo/target/(?<redacted>[0-9a-f]{2}/[0-9a-f]{14})/"`?

tests/testsuite/script.rs
```diff
@@ -1107,79 +1046,81 @@ fn cmd_metadata_with_embedded() {
...
-        "workspace_root": "[..]/foo",
-        "metadata": null
-    }"#,
+      "authors": [],
+      "categories": [],
+      "default_run": null,
+      "dependencies": [],
+      "description": null,
+      "documentation": null,
+      "edition": "[..]",
```
epage, line 1061: With snapshotting, we could just show the current stable edition here.

tests/testsuite/script.rs
```diff
@@ -1107,79 +1046,81 @@ fn cmd_metadata_with_embedded() {
...
+      {
+        "dependencies": [],
+        "deps": [],
+        "features": [],
+        "id": "path+[ROOTURL]/foo#script@0.0.0"
+      }
+    ],
+    "root": "path+[ROOTURL]/foo#script@0.0.0"
+  },
+  "target_directory": "[ROOT]/home/.cargo/target/[..]",
```
epage, line 1107: Does this no longer need `[..]`?

eth3lbert: Great catch! I've just updated the regex to allow directories that don't end with a slash.

PR #14135, opened 2024-06-25
Change tests to support `rustc` wording changes

estebank:
Between https://github.com/rust-lang/rust/pull/126810 and https://github.com/rust-lang/rust/pull/126810 the output of `rustc` for resolution errors is going to change in such a way that some existing cargo tests will fail. Change them to support both the current and future output, so that those PRs can land in `rustc`.

### DIFF
--- a/tests/testsuite/check.rs
+++ b/tests/testsuite/check.rs
@@ -347,7 +347,8 @@ fn rustc_check_err() {
         .with_status(101)
         .with_stderr_contains("[CHECKING] bar [..]")
         .with_stderr_contains("[CHECKING] foo [..]")
-        .with_stderr_contains("[..]cannot find function `qux` in [..] `bar`")
+        .with_stderr_contains("[..]cannot find function `qux`[..]")
+        .with_stderr_contains("[..]in crate `bar`")
         .run();
 }
 
--- a/tests/testsuite/message_format.rs
+++ b/tests/testsuite/message_format.rs
@@ -138,7 +138,7 @@ test src/lib.rs - bar (line 1) ... FAILED
 failures:
 
 ---- src/lib.rs - bar (line 1) stdout ----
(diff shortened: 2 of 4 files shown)


tests/testsuite/check.rs
```diff
@@ -347,7 +347,8 @@ fn rustc_check_err() {
         .with_status(101)
         .with_stderr_contains("[CHECKING] bar [..]")
         .with_stderr_contains("[CHECKING] foo [..]")
-        .with_stderr_contains("[..]cannot find function `qux` in [..] `bar`")
+        .with_stderr_contains("[..]cannot find function `qux`[..]")
```
weihanglo, line 350: Should we stop asserting the string, and instead check only the appearance of `E0425`? I feel like the error code will always be there, no?

weihanglo (approved):

weihanglo:
Let's not block rustc change. We can clean up then.

Thank you!

@bors r+

tests/testsuite/check.rs
estebank: The error code will indeed always be there. I'm ok with either approach.

weihanglo: @epage
Any opinion on adding a redaction for the first line rustc diagnostic message? Something like

```
error[E0425]: [RUSTC_ERROR_MESSAGE]
```

We still need to use `...` to omit the multiline sub-diagnostics though.

estebank: Interesting, not a bad idea, but you can already accomplish that with `error[E025]: [..]` right?

epage: So we only match on the error number and not the error message?  I guess that would be specific enough.

epage: > Interesting, not a bad idea, but you can already accomplish that with error[E025]: [..] right?

The more we can auto-redact, the easier to is to write new tests and do the right thing (or update existing tests)

Issue #6529, opened 2019-01-09
(Option to) Fingerprint by file contents instead of mtime


illicitonion:
**Describe the problem you are trying to solve**
The primary problem I have is that when building my code on travis, the actual code in my workspace builds every time, even though much of it hasn't changed and I have target directory caching on. The reason is that travis makes a new clone of my git repo, which doesn't preserve mtimes. This can add about 5 minutes to every travis run. My project is mixed rust and non-rust code, so this adds 5 minutes to those runs even if no rust code has been affected. I started futzing with mtimes, but that seems fragile and not solving the root of the problem.

Additionally, edit-undo loops cause re-compilation locally, which is a little annoying.

**Describe the solution you'd like**
Add a new `LocalFingerprint::ContentBased(Digest, PathBuf)` variant to https://github.com/rust-lang/cargo/blob/b84e625c4401be3893a4d6d44d5dbac82b33f1c4/src/cargo/core/compiler/fingerprint.rs#L204-L209 which reads the content of the `PathBuf`, passes it through a  `SipHasher`, and mixes that into any aggregate fingerprints. Use this instead of `LocalFingerprint::MtimeBased`.

**Notes**
This will probably slow down no-op builds slightly (in some circumstances, such as with large build script inputs over NFS, significantly), so may want to be behind a flag (perhaps `--fingerprint-strategy={mtime,content}`).

This would probably also make more shared caching (that people talk about a lot, most recently at https://internals.rust-lang.org/t/idea-cargo-global-binary-cache/9002) easier.

I'd be happy to implement this if the PR is likely to be accepted :)

This would probably also fix
- #4425
- #3076
- #7775
- #5918
- #8868 ?
- #10175 ?
- #12060
- #13119

alexcrichton:
We're not really too interested in cryptographic hashing here for its security properties, so I think picking any reasonable fast algorithm should be fine (and I think `SipHasher` is reasonably fast as well). I would personally prefer we not wholesale switch away from mtimes just yet, but having this as at least a runtime option to choose from is a great start!

dwijnand:
Would be interesting to compare it to using FxHash.

Eh2406:
[...]

Issue #14136, opened 2024-06-25
Tracking Issue for checksum freshness


Xaeroxe:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/issues/14136)*

### Summary

Original issue: #6529 
Implementation: https://github.com/rust-lang/cargo/pull/14137 
Rustc implementation: https://github.com/rust-lang/rust/pull/126930
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#checksum-freshness

Add the ability to use content checksums rather than filesystem mtimes to determine if a crate should be recompiled.

### Unresolved Issues

* [ ] How should we handle checksums for files processed by build scripts?
  * One option is to [provide an officially sanctioned API for build scripts.](https://github.com/rust-lang/cargo/issues/12432) Then we can use that API to compute and output the correct checksum value when the relevant function is invoked in the build script.
* [ ] Figure out how this will be stabilized (always on, config)
  * [ ] Get benchmarked https://github.com/rust-lang/cargo/issues/14722
* [ ] If this is on by default, migrate Cargo's tests to it (hack on and make pass)

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

```[tasklist]
### Tasks
- [ ] https://github.com/rust-lang/cargo/pull/14137
- [ ] https://github.com/rust-lang/rust/pull/126930
```

weihanglo:
Here is a result of a performance benchmark that mostly tests the rustc part.

* It adds `aws-sdk-ec2@1.80.0` as a dependency and runs `cargo clean && cargo check`.
* `rg 'checksum:blake3' target/debug/deps | wc -l` shows 8313 matches. It means rustc roughly computes blake3 checksum 8313 times.
* Checksum computations on Cargo side are mostly ignored because Cargo assumes registry are immutable and won't do either mtime or checksum check .

**Result:** On par if you have zero local code.

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `cargo +nightly check --frozen` | 52.986 ± 0.366 | 52.549 | 53.681 | 1.00 |
[...]

PR #14137, opened 2024-06-25
initial version of checksum based freshness

Xaeroxe:
Implementation for https://github.com/rust-lang/cargo/issues/14136 and resolves https://github.com/rust-lang/cargo/issues/6529

This PR implements the use of checksums in cargo fingerprints as an alternative to using mtimes. This is most useful on systems with poor mtime implementations.

This has a dependency on https://github.com/rust-lang/rust/pull/126930. It's expected this will increase the time it takes to declare a build to be fresh. Still this loss in performance may be preferable to the issues the ecosystem has had with the use of mtimes for determining freshness.


Cargo.toml
```diff
@@ -204,6 +204,7 @@ unicode-width.workspace = true
 url.workspace = true
 walkdir.workspace = true
 supports-unicode = "3.0.0"
+md-5 = "0.10.6"
```
tgross35, line 207: Does this get used anywhere? Only sha256 is mentioned in these commits (there probably isn't much reason to support md5 anyway)

Xaeroxe: I'm not happy about supporting md5 myself, but it is something that my rustc PR could feasibly emit. I'm happy to remove it from both PRs if we can build a consensus on that.

Xaeroxe: This deserves more explanation. Md5 is used by rustc for debug info files, so the support for it was already present in rustc.

tgross35: Makes sense, just doesn’t look like it is used anywhere in this diff?

bjorn3: It's imported in src/cargo/core/compiler/fingerprint/mod.rs, which github hides by default due to the size of the diff.

Xaeroxe: It gets used here [Line 380 of src/cargo/core/compiler/fingerprint/mod.rs](https://github.com/rust-lang/cargo/pull/14137/files#diff-33fc5e62ca2fa6b54fd70249825e4f28dcc490317cd3a89e3c07aa62754406ceR380)

arlosi: The only reason that md5 and sha-1 were included as options for the debuginfo checksum was compatibility with PDB (pre-sha256 support) & DWARF5 debug formats.

Since we're making something new here, I don't think there's any reason to support deprecated hashing algorithms.

tgross35: > It gets used here [Line 380 of src/cargo/core/compiler/fingerprint/mod.rs](https://github.com/rust-lang/cargo/pull/14137/files#diff-33fc5e62ca2fa6b54fd70249825e4f28dcc490317cd3a89e3c07aa62754406ceR380)

🤦‍♂️ didn't see the collapsed file when I ctrl+F'd for it

Xaeroxe: md5 and sha1 are no longer supported for this. I've also added an option to use xxhash. xxhash is not cryptographically secure, but that's not necessary for this use case. So I figured it was worth comparing the performance of it to sha256.

Xaeroxe: xxhash has been swapped out for blake3. It was decided during the rustc MCP that cryptographically secure hashes are necessary in order for this freshness algorithm to be correct.

Xaeroxe:
Merge conflicts resolved.

src/cargo/core/features.rs
```diff
@@ -757,6 +757,7 @@ unstable_cli_options!(
     build_std: Option<Vec<String>>  = ("Enable Cargo to compile the standard library itself as part of a crate graph compilation"),
     build_std_features: Option<Vec<String>>  = ("Configure features enabled for the standard library itself when building the standard library"),
     cargo_lints: bool = ("Enable the `[lints.cargo]` table"),
+    checksum_freshness: bool = ("Use a checksum to determine if output is fresh rather than filesystem mtime"),
```
epage, line 760: Unstable features should also be documented at https://doc.rust-lang.org/nightly/cargo/reference/unstable.html

See `src/doc/src/reference/unstable.md`

src/cargo/util/command_prelude.rs
```diff
@@ -708,6 +708,7 @@ Run `{cmd}` to see possible targets."
         build_config.build_plan = self.flag("build-plan");
         build_config.unit_graph = self.flag("unit-graph");
         build_config.future_incompat_report = self.flag("future-incompat-report");
+        build_config.checksum_freshness = self.flag("checksum-freshness");
```
epage, line 711: I think there might be a misunderstanding which makes me wonder how the tests are working. `self.flag` is for reading CLI flags from clap but no `checksum-freshness` CLI flag was created.  Instead there is `checksum-freshness` unstable feature flag that should be accessible as `gctx.cli_unstable().checksum_freshness`

src/cargo/util/command_prelude.rs
```diff
@@ -742,6 +743,10 @@ Run `{cmd}` to see possible targets."
             gctx.cli_unstable()
                 .fail_if_stable_opt("--unit-graph", 8002)?;
         }
+        if build_config.checksum_freshness {
+            gctx.cli_unstable()
+                .fail_if_stable_opt("--checksum-freshness", 14136)?;
+        }
```
epage, line 749: If we're defining a feature flag for this, we don't need to require `-Zunstable-options` (which means we should also stop having the tests set it)

src/cargo/util/context/mod.rs
```diff
@@ -91,6 +91,7 @@ use serde::Deserialize;
 use serde_untagged::UntaggedEnumVisitor;
 use time::OffsetDateTime;
 use toml_edit::Item;
+use tracing::warn;
```
epage, line 94: This looks like dead code?

tests/testsuite/freshness_checksum.rs
```diff
@@ -0,0 +1,2880 @@
...
+use std::thread;
+
+use super::death;
+use cargo_test_support::paths::CargoPathExt;
+use cargo_test_support::prelude::*;
+use cargo_test_support::registry::Package;
+use cargo_test_support::{basic_lib_manifest, basic_manifest, project, rustc_host, rustc_host_env};
+
+#[cargo_test]
+fn checksum_actually_uses_checksum() {
```
epage, line 18: At minimum, could you structure your PR so its
- A commit with these tests without `-Zchecksum-freshness`
- A commit with the checksum work that also updates the tests to pass `-Zchecksum-freshness`

A big benefit to this is it shows to reviewers / the community how this feature is comparing to what was being done before

(sometimes, I also break out "adding an unstable feature" into its own commit which is the flag + docs)

tests/testsuite/freshness_checksum.rs
```diff
@@ -0,0 +1,2880 @@
...
+    let p = project()
+        .file("src/main.rs", "mod a; fn main() {}")
+        .file("src/a.rs", "")
+        .build();
+
+    p.cargo("build")
+        .masquerade_as_nightly_cargo(&["checksum-freshness"])
+        .args(&["-Z", "unstable-options", "-Z", "checksum-freshness"])
```
epage, line 26: nit: could you do
```rust
p.cargo("build -Zchecksum-freshness")
```

Identifying quickly in the test case what command is being run makes it a lot easier to read the test.  This buries it compared to most tests.

tests/testsuite/freshness_checksum.rs
```diff
@@ -0,0 +1,2880 @@
...
+use cargo_test_support::{basic_lib_manifest, basic_manifest, project, rustc_host, rustc_host_env};
+
+#[cargo_test]
+fn checksum_actually_uses_checksum() {
+    let p = project()
+        .file("src/main.rs", "mod a; fn main() {}")
+        .file("src/a.rs", "")
+        .build();
+
+    p.cargo("build")
```
epage, line 24: Is there a reason we need `cargo build` instead of `cargo check`?  The latter helps keep test times down

tests/testsuite/freshness_checksum.rs
epage: Note: I've not dug deep into the tests, waiting on this change

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1399,7 +1518,11 @@ fn calculate_normal(
     } else {
         let dep_info = dep_info_loc(build_runner, unit);
         let dep_info = dep_info.strip_prefix(&target_root).unwrap().to_path_buf();
-        vec![LocalFingerprint::CheckDepInfo { dep_info }]
+        if build_runner.bcx.gctx.cli_unstable().checksum_freshness {
+            vec![LocalFingerprint::CheckDepInfoChecksums { dep_info }]
+        } else {
+            vec![LocalFingerprint::CheckDepInfo { dep_info }]
+        }
```
epage, line 1523: Why did you go with separate enums?  Are we not able to read the checksum feature flag?  Would it be better to have a `checksum: bool` in it?

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1865,16 +1988,32 @@ pub fn parse_dep_info(
...
+            .insert(path, (file_len, Checksum::from_str(&checksum)?));
+    }
     Ok(Some(ret))
 }
 
+fn make_absolute_path(
+    ty: DepInfoPathType,
+    pkg_root: &Path,
+    path: PathBuf,
+    target_root: &Path,
```
epage, line 2006: nit: I'd make `path` the last parameter.  The first three are all related in determining what the `root` should be

Alternatively, you could just make this a `path_root` function that only takes the first three and the join is done in the caller

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -858,12 +899,48 @@ impl LocalFingerprint {
...
     }
 }
 
+fn dep_info_shared(
+    pkg_root: &Path,
+    target_root: &Path,
+    dep_info: &PathBuf,
+    cargo_exe: &Path,
+    gctx: &GlobalContext,
+) -> Result<Either<StaleItem, RustcDepInfo>, anyhow::Error> {
```
epage, line 913: This name doesn't tell the reader what this function is doing

epage, line 913: Not thrilled with using `Either`, I feel like its obscuring what this is doing.

epage: For both of these, the problems go away I think if we merge the `LocalFingerprint` variants

src/cargo/core/compiler/fingerprint/mod.rs
epage: Ah, I assume this is tied to the serialization.  I wonder if we should have a slight decoupling as I feel like this makes things more complicated.  I hope having two types of fingerprints is short term and so we shouldn't try to over-generalize

Cargo.toml
ReillyBrogan: Blake3 appears to more mature over-all anyway as it has runtime CPU feature detection for SSE2/SSE4.1/AVX2/AVX-512. The most commonly used XXH library only supports compile-time configuration of these which makes it less suited to a tool like cargo that's compiled once and ran on a wide variety of hardware.

When I was benchmarking algorithms for a similar purpose (source fingerprinting in go-task) both Blake3 and XXH were fast enough that the bottleneck was almost always disk IO and the runtime differences were minimal.

src/cargo/core/compiler/fingerprint/mod.rs
Xaeroxe: Can you help be better understand what such a decoupling might look like?

Xaeroxe: Nevermind, I think I've got it.

tests/testsuite/freshness_checksum.rs
Xaeroxe: So it's worth noting that `freshness_checksum.rs` derives very heavily from `freshness.rs`. So a version of `freshness_checksum.rs` without the freshness flag would just be a subset of `freshness.rs`. I'm not sure how this provides new information. There are two tests which are truly unique to `freshness_checksum.rs`, which are `same_size_different_content()` and `checksum_actually_uses_checksum()`.

One might debate the merit of duplicating the tests like that. If you really wanted to deduplicate the tests then this would likely require a special case be added to the test runner code.

src/cargo/core/features.rs
Xaeroxe: Resolved in 6cf92e1f1677ebe030ddfabea8782e2f721378ce

src/cargo/util/command_prelude.rs
Xaeroxe: Resolved in 6cf92e1f1677ebe030ddfabea8782e2f721378ce, this was just vestigial.

src/cargo/util/command_prelude.rs
Xaeroxe: Resolved in 0240bf05d640eae07d6d74523e14eea303db983a

src/cargo/util/context/mod.rs
Xaeroxe: Resolved in 5afdc115d295f74c475a2252df40ff7d2351c3eb

src/cargo/core/compiler/fingerprint/mod.rs
Xaeroxe: Resolved in 5afdc115d295f74c475a2252df40ff7d2351c3eb

src/cargo/core/compiler/fingerprint/mod.rs
Xaeroxe: Function was removed as part of d422f64ca56a7b6dfab5c8e5d4bbb06cccbbe5b3

Xaeroxe: Function was removed as part of d422f64ca56a7b6dfab5c8e5d4bbb06cccbbe5b3

tests/testsuite/freshness_checksum.rs
Xaeroxe: No, where it was possible to do so without breaking tests I downgraded these to `cargo check` in 3cc6814ee8ede4ab4f3defb581f37c0ae1a1623a

tests/testsuite/freshness_checksum.rs
Xaeroxe: Resolved in 3cc6814ee8ede4ab4f3defb581f37c0ae1a1623a

src/cargo/core/compiler/fingerprint/mod.rs
Xaeroxe: Resolved in d422f64ca56a7b6dfab5c8e5d4bbb06cccbbe5b3

Xaeroxe: The original thinking was that I should maintain backwards compatibility with prior build caches, but then I remembered that cargo releases alongside rust, and rust doesn't provide backwards compatibility of build caches.

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -2206,6 +2419,22 @@ pub fn parse_rustc_dep_info(rustc_dep_info: &Path) -> CargoResult<RustcDepInfo>
                 None => None,
             };
             ret.env.push((unescape_env(env_var)?, env_val));
+        } else if let Some(rest) = line.strip_prefix("# checksum:") {
```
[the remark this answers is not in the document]
weihanglo: As a side note: is it possible to combine a comment and the corresponding files into one line in rustc? It might make dep-info harder to read, though it is not meant to read by human, and combining them could potentially write less bytes. I dont know if it is really criticial in terms of performance.

```
src/lib.rs: src/someotherfile.rs # checksum:blake3=660e91cbd3 file_len:9
```

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -2102,13 +2271,14 @@ pub struct RustcDepInfo {
 struct EncodedDepInfo {
     files: Vec<(DepInfoPathType, PathBuf)>,
     env: Vec<(String, Option<String>)>,
+    checksum: Vec<(DepInfoPathType, PathBuf, u64, String)>,
```
weihanglo, line 2274: (not a blocker) 
I feel like `files` and `checksum` should eventually merge into one field.

src/cargo/core/compiler/fingerprint/dirty_reason.rs
```diff
@@ -235,6 +256,43 @@ impl DirtyReason {
...
+                        new_checksum,
+                    } => {
+                        let file = source.strip_prefix(root).unwrap_or(&source);
+                        s.dirty_because(
+                            unit,
+                            format_args!(
+                                "the file `{}` has changed (checksum didn't match, {} != {})",
+                                file.display(),
+                                stored_checksum,
+                                new_checksum,
```
weihanglo, line 271: nit: arg capture whenver possible and making sense

```suggestion
                                "the file `{}` has changed (checksum didn't match, {stored_checksum} != {new_checksum})",
                                file.display(),
```

src/cargo/core/compiler/fingerprint/dirty_reason.rs
```diff
@@ -183,6 +187,16 @@ impl DirtyReason {
             DirtyReason::PrecalculatedComponentsChanged { .. } => {
                 s.dirty_because(unit, "the precalculated components changed")
             }
+            DirtyReason::ChecksumUseChanged { old, new: _ } => {
```
weihanglo, line 190: If we are not using `new` anywhere, should we just remove this field?

tests/testsuite/freshness_checksum.rs
weihanglo, line 1: Oops. The test module might need to be rewritten. See #14039

We can sort this out later as follow-ups. Don't worry

weihanglo: This file looks like a copy of `freshness.rs` with `-Zchecksum-freshness` and `masquerade_as_nightly_cargo` everywhere.

Could you point out which tests are new and worth a review? Guess they are

* `checksum_actually_uses_checksum`
* `same_size_different_content`
* `modifying_and_moving`

Also could you add a test that verifying `-Zchecksum-freshness` is gated behind nightly? (A test without `masquerade_as_nightly_cargo` is sufficient.

weihanglo (commented):
Looks like we are almost there! Thank for yours efforts, epage and Xaeroxe. Thought it would be a tough review, but honestly it was a happy time :)

src/doc/src/reference/unstable.md
```diff
@@ -504,6 +505,13 @@ the crate will be rebuilt). The primary use case is for building the compiler
 itself, which has implicit dependencies on the standard library that would
 otherwise be untracked for change-detection.
 
+## checksum-freshness
+* Tracking issue: [#14136](https://github.com/rust-lang/cargo/issues/14136)
+
+The `-Z checksum-freshness` flag will replace the use of file mtimes in cargo's
```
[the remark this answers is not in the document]
Xaeroxe: Addressed as part of ba5899ae0f2211184731b2a8a98313881cb85675

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1922,44 +2030,82 @@ where
...
+                    };
+                    if current_file_len != file_len {
+                        return Some(StaleItem::FileSizeChanged {
+                            path: path.to_path_buf(),
+                            new_size: current_file_len,
+                            old_size: file_len,
+                        });
+                    }
+                    let Ok(checksum) = Checksum::compute(prior_checksum.algo, file) else {
+                        return Some(StaleItem::MissingFile(path.to_path_buf()));
```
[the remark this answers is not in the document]
Xaeroxe: Addressed as part of ba5899ae0f2211184731b2a8a98313881cb85675

This wasn't a great use of this error type, so I added a new one.

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1922,44 +2030,82 @@ where
...
+
+            let path_checksum = match checksum_cache.entry(path_buf) {
+                Entry::Occupied(o) => *o.get(),
+                Entry::Vacant(v) => {
+                    let Ok(file) = File::open(path) else {
+                        return Some(StaleItem::MissingFile(path.to_path_buf()));
+                    };
+                    let Ok(current_file_len) = file.metadata().map(|m| m.len()) else {
+                        return Some(StaleItem::FailedToReadMetadata(path.to_path_buf()));
+                    };
```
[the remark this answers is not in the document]
Xaeroxe: Addressed as part of ba5899ae0f2211184731b2a8a98313881cb85675

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -805,10 +829,11 @@ impl LocalFingerprint {
             // matches, and for each file we see if any of them are newer than
             // the `dep_info` file itself whose mtime represents the start of
             // rustc.
-            LocalFingerprint::CheckDepInfo { dep_info } => {
+            LocalFingerprint::CheckDepInfo { dep_info, checksum } => {
                 let dep_info = target_root.join(dep_info);
+                let cargo_exe = cargo_exe;
                 let Some(info) = parse_dep_info(pkg_root, target_root, &dep_info)? else {
-                    return Ok(Some(StaleItem::MissingFile(dep_info)));
+                    return Ok(Some(StaleItem::MissingFile(dep_info.clone())));
```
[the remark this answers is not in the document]
Xaeroxe: Addressed in ba5899ae0f2211184731b2a8a98313881cb85675

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -835,15 +860,35 @@ impl LocalFingerprint {
                         current,
                     }));
                 }
-                Ok(find_stale_file(mtime_cache, &dep_info, info.files.iter()))
+                macro_rules! find_stale_file {
```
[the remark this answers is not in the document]
Xaeroxe: Removed the macro as part of ba5899ae0f2211184731b2a8a98313881cb85675

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -835,15 +860,35 @@ impl LocalFingerprint {
...
+                            $iter,
+                            *checksum,
+                        ))
+                    };
+                }
+                if *checksum {
+                    find_stale_file!(info
+                        .files
+                        .iter()
+                        .map(|file| (file.clone(), info.checksum.get(file).cloned())))
```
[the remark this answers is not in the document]
Xaeroxe: Added as part of ba5899ae0f2211184731b2a8a98313881cb85675, thanks for the code suggestion!

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -2250,3 +2479,201 @@ pub fn parse_rustc_dep_info(rustc_dep_info: &Path) -> CargoResult<RustcDepInfo>
...
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum InvalidChecksum {
+    InvalidChecksumAlgo(InvalidChecksumAlgo),
+    InvalidChecksum(ChecksumAlgo),
+    InvalidFormat,
+}
+
+impl Display for InvalidChecksum {
```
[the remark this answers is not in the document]
Xaeroxe: Addressed in ba5899ae0f2211184731b2a8a98313881cb85675

src/cargo/core/compiler/fingerprint/dirty_reason.rs
Xaeroxe: Addressed in ba5899ae0f2211184731b2a8a98313881cb85675

src/cargo/core/compiler/fingerprint/dirty_reason.rs
Xaeroxe: Yeah, new is defined as precisely the opposite of `old`, so let's get rid of it. Done in ba5899ae0f2211184731b2a8a98313881cb85675

src/cargo/core/compiler/fingerprint/mod.rs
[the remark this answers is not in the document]
weihanglo: See <https://github.com/rust-lang/rust/pull/126930#discussion_r1783822618>, which shows that my suggestion is not even better.

src/cargo/core/compiler/fingerprint/mod.rs
Xaeroxe: They are now the same field

tests/testsuite/freshness_checksum.rs
weihanglo: I've reworked on tests so they should be better to review commit-by-commit

* Updated to new test syntax.
* Split copying `freshness.rs` into its own commit.
* Split non-functional changes from the commit that enables `-Zfreshness-checksum`

Other than these, everything should be the same as before. I haven't added any nightly-gate tests.

weihanglo: BTW this rework now passes all tests with <https://github.com/rust-lang/rust/pull/126930/commit/58c5ac43aea5c94d61e3a7a4e8646e714230574d> in <https://github.com/rust-lang/rust/pull/126930>. Feel free to continue.

Xaeroxe: The nightly gate tests have been added. Thanks for your help with the other notes here!

weihanglo:
@bors try

Let's see how it really goes.

weihanglo:
All tests are passed. I will do another round of review today.

tests/testsuite/freshness_checksum.rs
```diff
@@ -0,0 +1,2883 @@
...
+
+"#]])
+        .run();
+}
+
+#[cargo_test(nightly, reason = "requires -Zchecksum-hash-algorithm")]
+#[cfg_attr(
+    not(all(target_arch = "x86_64", target_os = "windows", target_env = "msvc")),
+    ignore
+)]
```
weihanglo, line 2515: This should be sufficient

```suggestion
#[cfg(all(target_arch = "x86_64", target_os = "windows", target_env = "msvc"))]
```

tests/testsuite/dep_info.rs
weihanglo, line 1: Could we have a test verifying that the Cargo-flavored dep-info under fingerprint folder is correctly encoded? Something like [`relative_depinfo_paths_ws`](https://github.com/rust-lang/cargo/blob/4ddd017694a0415a7c91598153e642e3c2b26321/tests/testsuite/dep_info.rs#L251) but simpler

weihanglo (commented):
@Xanewok 

Would you be willing to rearrange commits and make them easier to review/track for future readers?

Some intermediate commits like a45d66345acad70e930f883a207cd1f9fb9b70b6 and  5b36d643b23172c9cbac05ccb47e217bf5b6f63b could be squashed, and some like what I've done for tests are helpful to understand they were copied from the other file.

If that sounds too much a burden, no worries. I can do a simple cleanup before this PR merges. Thank you :)

Xaeroxe:
@weihanglo All comments addressed in 9ede54d000c09e1824080d02bfbee647009def96

Also the PR has been rebased to remove some of the rejected designs for this feature.

weihanglo (approved):
I just rearrange commits a bit, so we can remove unnecessary commit like 2dafe11009d3228612397c6fe87b02a5bef25b13.

Thank you for the hard work! I am going to merge this when the CI is green.

jonhoo:
Is there a way we can tag this for mention in the release notes? May be worth giving a bit of celebration to :)

weihanglo:
It has been listed in changelog of 1.83 already, under the nightly section! https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-183-2024-11-28

Though we haven't figured out the build script part (see <https://github.com/rust-lang/cargo/issues/14136>), maybe we could celebrate a bit later when we resolve that? :)

weihanglo:
If anyone is interested in helping this go stabilized, please share your benchmark results for `-Zchecksum-freshness` on [this Zulip topic](https://rust-lang.zulipchat.com/#narrow/channel/334885-t-cargo.2Fbuild-integration/topic/Call.20for.20testing.3A.20use.20of.20checksums.20over.20mtime), or in #14722.

PR #14140, opened 2024-06-25
Add capability of making breaking changes in `update --precise`

torhovland:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/pull/14140)*

Implements the second half of #12425.

With the `unstable-options` feature enabled, `cargo update --precise` will allow making upgrades/downgrades that require changes to be made to the manifest files, similar to `cargo update --breaking`.

Blocked on #14259. See https://github.com/rust-lang/cargo/pull/14140/files#r1682381732.

### DIFF
--- a/src/bin/cargo/commands/update.rs
+++ b/src/bin/cargo/commands/update.rs
@@ -1,8 +1,11 @@
+use std::collections::HashMap;
+
 use crate::command_prelude::*;
 
 use anyhow::anyhow;
 use cargo::ops::{self, UpdateOptions};
 use cargo::util::print_available_packages;
+use tracing::trace;
 
 pub fn cli() -> Command {
     subcommand("update")
@@ -92,28 +95,38 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
     let update_opts = UpdateOptions {
         recursive: args.flag("recursive"),
         precise: args.get_one::<String>("precise").map(String::as_str),
+        breaking: args.flag("breaking"),
         to_update,
         dry_run: args.dry_run(),
         workspace: args.flag("workspace"),
         gctx,
     };
 
-    if args.flag("breaking") {
-        gctx.cli_unstable()
-            .fail_if_stable_opt("--breaking", 12425)?;
+    let breaking_update = update_opts.breaking; // or if doing a breaking precise update, coming in #14140.
 
-        let upgrades = ops::upgrade_manifests(&mut ws, &update_opts.to_update)?;
-        ops::resolve_ws(&ws, update_opts.dry_run)?;
-        ops::write_manifest_upgrades(&ws, &upgrades, update_opts.dry_run)?;
-
-        if update_opts.dry_run {
-            update_opts
-                .gctx
-                .shell()
-                .warn("aborting update due to dry run")?;
+    // We are using the term "upgrade" here, which is the typical case, but it
+    // can also be a downgrade (in the case of a precise update). In general, it
+    // is a change to a version req matching a precise version.
+    let upgrades = if breaking_update {
+        if update_opts.breaking {
+            gctx.cli_unstable()
+                .fail_if_stable_opt("--breaking", 12425)?;
         }
+
+        trace!("allowing breaking updates");
+        ops::upgrade_manifests(&mut ws, &update_opts.to_update)?
     } else {
-        ops::update_lockfile(&ws, &update_opts)?;
+        HashMap::new()
+    };
+
+    ops::update_lockfile(&ws, &update_opts, &upgrades)?;
+    ops::write_manifest_upgrades(&ws, &upgrades, update_opts.dry_run)?;
+
+    if update_opts.dry_run {
+        update_opts
+            .gctx
+            .shell()
+            .warn("aborting update due to dry run")?;
     }
 
     Ok(())
--- a/src/bin/cargo/commands/update.rs
+++ b/src/bin/cargo/commands/update.rs
@@ -102,7 +102,8 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
         gctx,
     };
 
-    let breaking_update = update_opts.breaking; // or if doing a breaking precise update, coming in #14140.
+    let breaking_update = update_opts.breaking
+        || (update_opts.precise.is_some() && gctx.cli_unstable().unstable_options);
 
     // We are using the term "upgrade" here, which is the typical case, but it
     // can also be a downgrade (in the case of a precise update). In general, it
@@ -114,7 +115,7 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
         }
 
         trace!("allowing breaking updates");
-        ops::upgrade_manifests(&mut ws, &update_opts.to_update)?
+        ops::upgrade_manifests(&mut ws, &update_opts.to_update, &update_opts.precise)?
     } else {
         HashMap::new()
     };
--- a/src/cargo/ops/cargo_update.rs
+++ b/src/cargo/ops/cargo_update.rs
@@ -15,6 +15,7 @@ use crate::util::toml_mut::manifest::LocalManifest;
 use crate::util::toml_mut::upgrade::upgrade_requirement;
 use crate::util::{style, OptVersionReq};
 use crate::util::{CargoResult, VersionExt};
+use anyhow::Context as _;
 use itertools::Itertools;
 use semver::{Op, Version, VersionReq};
 use std::cmp::Ordering;
@@ -262,6 +263,7 @@ pub fn print_lockfile_changes(
 pub fn upgrade_manifests(
     ws: &mut Workspace<'_>,
     to_update: &Vec<String>,
+    precise: &Option<&str>,
 ) -> CargoResult<UpgradeMap> {
     let gctx = ws.gctx();
     let mut upgrades = HashMap::new();
@@ -290,6 +292,7 @@ pub fn upgrade_manifests(
                 upgrade_dependency(
                     &gctx,
                     &to_update,
+                    precise,
                     &mut registry,
                     &mut upgrades,
                     &mut upgrade_messages,
@@ -304,6 +307,7 @@ pub fn upgrade_manifests(
 fn upgrade_dependency(
     gctx: &GlobalContext,
     to_update: &Vec<PackageIdSpec>,
+    precise: &Option<&str>,
     registry: &mut PackageRegistry<'_>,
     upgrades: &mut UpgradeMap,
     upgrade_messages: &mut HashSet<String>,
@@ -361,7 +365,7 @@ fn upgrade_dependency(
     let query =
         crate::core::dependency::Dependency::parse(name, None, dependency.source_id().clone())?;
 
-    let possibilities = {
+    let possibilities = if precise.is_none() {
         loop {
             match registry.query_vec(&query, QueryKind::Exact) {
                 std::task::Poll::Ready(res) => {
@@ -370,6 +374,8 @@ fn upgrade_dependency(
                 std::task::Poll::Pending => registry.block_until_ready()?,
             }
         }
+    } else {
+        Vec::new()
     };
 
     let latest = if !possibilities.is_empty() {
@@ -383,17 +389,23 @@ fn upgrade_dependency(
         None
     };
 
-    let Some(latest) = latest else {
+    let new_version = if let Some(precise) = precise {
+        Version::parse(precise)
+            .with_context(|| format!("invalid version format for precise version `{precise}`"))?
+    } else if let Some(latest) = latest {
+        latest.clone()
+    } else {
+        // Breaking (not precise) upgrade did not find a latest version
         trace!("skipping dependency `{name}` without any published versions");
         return Ok(dependency);
     };
 
-    if current.matches(&latest) {
+    if current.matches(&new_version) {
         trace!("skipping dependency `{name}` without a breaking update available");
         return Ok(dependency);
     }
 
-    let Some((new_req_string, _)) = upgrade_requirement(&current.to_string(), latest)? else {
+    let Some((new_req_string, _)) = upgrade_requirement(&current.to_string(), &new_version)? else {
         trace!("skipping dependency `{name}` because the version requirement didn't change");
         return Ok(dependency);
     };
@@ -401,17 +413,36 @@ fn upgrade_dependency(
     let upgrade_message = format!("{name} {current} -> {new_req_string}");
     trace!(upgrade_message);
 
+    let old_version = semver::Version::new(
+        comparator.major,
+        comparator.minor.unwrap_or_default(),
+        comparator.patch.unwrap_or_default(),
+    );
+    let is_downgrade = new_version < old_version;
+    let status = if is_downgrade {
+        "Downgrading"
+    } else {
+        "Upgrading"
+    };
+
     if upgrade_messages.insert(upgrade_message.clone()) {
         gctx.shell()
-            .status_with_color("Upgrading", &upgrade_message, &style::GOOD)?;
+            .status_with_color(status, &upgrade_message, &style::WARN)?;
     }
 
     upgrades.insert(
         (name, dependency.source_id()),
-        (current.clone(), latest.clone()),
+        (current.clone(), new_version.clone()),
     );
 
-    let req = OptVersionReq::Req(VersionReq::parse(&latest.to_string())?);
+    let new_version_req = VersionReq::parse(&new_version.to_string())?;
+
+    let req = if precise.is_some() {
+        OptVersionReq::Precise(new_version, new_version_req)
+    } else {
+        OptVersionReq::Req(new_version_req)
+    };
+
     let mut dep = dependency.clone();
     dep.set_version_req(req);
     Ok(dep)
--- a/src/cargo/ops/cargo_update.rs
+++ b/src/cargo/ops/cargo_update.rs
@@ -9,6 +9,7 @@ use crate::ops;
 use crate::sources::source::QueryKind;
 use crate::util::cache_lock::CacheLockMode;
 use crate::util::context::GlobalContext;
+use crate::util::interning::InternedString;
 use crate::util::toml_mut::dependency::{MaybeWorkspace, Source};
 use crate::util::toml_mut::manifest::LocalManifest;
 use crate::util::toml_mut::upgrade::upgrade_requirement;
@@ -20,12 +21,25 @@ use std::cmp::Ordering;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use tracing::{debug, trace};
 
-pub type UpgradeMap = HashMap<(String, SourceId), Version>;
+/// A map of all breaking upgrades which is filled in by
+/// upgrade_manifests/upgrade_dependency when going through workspace member
+/// manifests, and later used by write_manifest_upgrades in order to know which
+/// upgrades to write to manifest files on disk. Also used by update_lockfile to
+/// know which dependencies to keep unchanged if any have been upgraded (we will
+/// do either breaking or non-breaking updates, but not both).
+pub type UpgradeMap = HashMap<
+    // The key is a package identifier consisting of the name and the source id.
+    (InternedString, SourceId),
+    // The value is the original version requirement before upgrade, and the
+    // upgraded version.
+    (VersionReq, Version),
+>;
 
 pub struct UpdateOptions<'a> {
     pub gctx: &'a GlobalContext,
     pub to_update: Vec<String>,
     pub precise: Option<&'a str>,
+    pub breaking: bool,
     pub recursive: bool,
     pub dry_run: bool,
     pub workspace: bool,
@@ -49,7 +63,11 @@ pub fn generate_lockfile(ws: &Workspace<'_>) -> CargoResult<()> {
     Ok(())
 }
 
-pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoResult<()> {
+pub fn update_lockfile(
+    ws: &Workspace<'_>,
+    opts: &UpdateOptions<'_>,
+    upgrades: &UpgradeMap,
+) -> CargoResult<()> {
     if opts.recursive && opts.precise.is_some() {
         anyhow::bail!("cannot specify both recursive and precise simultaneously")
     }
@@ -91,7 +109,38 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
     let mut registry = ws.package_registry()?;
     let mut to_avoid = HashSet::new();
 
-    if opts.to_update.is_empty() {
+    if opts.breaking {
+        // We don't necessarily want to update all specified packages. If we are
+        // doing a breaking update (or precise upgrades, coming in #14140), we
+        // don't want to touch any packages that have no breaking updates. So we
+        // want to only avoid all packages that got upgraded.
+
+        for name in opts.to_update.iter() {
+            // We still want to query any specified package, for the sake of
+            // outputting errors if they don't exist.
+            previous_resolve.query(name)?;
+        }
+
+        for ((name, source_id), (version_req, _)) in upgrades.iter() {
+            if let Some(matching_dep) = previous_resolve.iter().find(|dep| {
+                dep.name() == *name
+                    && dep.source_id() == *source_id
+                    && version_req.matches(dep.version())
+            }) {
+                let spec = PackageIdSpec::new(name.to_string())
+                    .with_url(source_id.url().clone())
+                    .with_version(matching_dep.version().clone().into())
+                    .to_string();
+                let pid = previous_resolve.query(&spec)?;
+                to_avoid.insert(pid);
+            } else {
+                // Should never happen
+                anyhow::bail!(
+                    "no package named `{name}` with source `{source_id}` and version matching `{version_req}` in the previous lockfile",
+                )
+            }
+        }
+    } else if opts.to_update.is_empty() {
         if !opts.workspace {
             to_avoid.extend(previous_resolve.iter());
             to_avoid.extend(previous_resolve.unused_patches());
@@ -185,11 +234,7 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
         opts.precise.is_some(),
         &mut registry,
     )?;
-    if opts.dry_run {
-        opts.gctx
-            .shell()
-            .warn("not updating lockfile due to dry run")?;
-    } else {
+    if !opts.dry_run {
         ops::write_pkg_lockfile(ws, &mut resolve)?;
     }
     Ok(())
@@ -361,7 +406,10 @@ fn upgrade_dependency(
             .status_with_color("Upgrading", &upgrade_message, &style::GOOD)?;
     }
 
-    upgrades.insert((name.to_string(), dependency.source_id()), latest.clone());
+    upgrades.insert(
+        (name, dependency.source_id()),
+        (current.clone(), latest.clone()),
+    );
 
     let req = OptVersionReq::Req(VersionReq::parse(&latest.to_string())?);
     let mut dep = dependency.clone();
@@ -433,7 +481,7 @@ pub fn write_manifest_upgrades(
                     continue;
                 };
 
-                let Some(latest) = upgrades.get(&(name.to_owned(), source_id)) else {
+                let Some((_, latest)) = upgrades.get(&(name.into(), source_id)) else {
                     trace!("skipping dependency without an upgrade: {name}");
                     continue;
                 };
--- a/src/cargo/ops/cargo_update.rs
+++ b/src/cargo/ops/cargo_update.rs
@@ -109,12 +109,13 @@ pub fn update_lockfile(
     };
     let mut registry = ws.package_registry()?;
     let mut to_avoid = HashSet::new();
+    let breaking_update = !upgrades.is_empty();
 
-    if opts.breaking {
+    if breaking_update {
         // We don't necessarily want to update all specified packages. If we are
-        // doing a breaking update (or precise upgrades, coming in #14140), we
-        // don't want to touch any packages that have no breaking updates. So we
-        // want to only avoid all packages that got upgraded.
+        // doing a breaking update or precise upgrade, we don't want to touch
+        // any packages that have no breaking updates. So we want to only avoid
+        // all packages that got upgraded.
 
         for name in opts.to_update.iter() {
             // We still want to query any specified package, for the sake of
--- a/src/cargo/util/semver_ext.rs
+++ b/src/cargo/util/semver_ext.rs
@@ -178,6 +178,8 @@ impl From<VersionReq> for OptVersionReq {
 
 #[cfg(test)]
 mod matches_prerelease {
+    use semver::{Version, VersionReq};
+
     use super::OptVersionReq;
 
     #[test]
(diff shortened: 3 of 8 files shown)


weihanglo:
Should we wait for #14049, or this can be reviewed independently?

torhovland:
The 4 new commits here can be reviewed. I think #14049 is close to being merged anyway.

epage:
> The 4 new commits here can be reviewed. I think https://github.com/rust-lang/cargo/pull/14049 is close to being merged anyway.

In the future, please make a blocking thing like that explicit either by making this a draft or opening an issue on an arbitrary part of the code.

tests/testsuite/update.rs
```diff
@@ -1687,7 +1687,7 @@ fn update_breaking_dry_run() {
     Package::new("ws", "2.0.0").publish();
 
     p.cargo("update -Zunstable-options --dry-run --breaking")
-        .masquerade_as_nightly_cargo(&["update-breaking"])
+        .masquerade_as_nightly_cargo(&[])
```
epage: Why were these changed?

src/cargo/core/features.rs
```diff
@@ -786,6 +786,7 @@ unstable_cli_options!(
     target_applies_to_host: bool = ("Enable the `target-applies-to-host` key in the .cargo/config.toml file"),
     trim_paths: bool = ("Enable the `trim-paths` option in profiles"),
     unstable_options: bool = ("Allow the usage of unstable options"),
+    update_precise_breaking: bool = ("Allow `update --precise` to do breaking upgrades"),
```
epage, line 789: Why a new unstable option rather than `unstable_options`?

tests/testsuite/update.rs
torhovland: Because there is no `update-breaking` feature. There's only the `-Zunstable-options`.

tests/testsuite/update_duplicated_with_precise_breaking_feature.rs
```diff
@@ -0,0 +1,712 @@
+//! Duplicating tests for `cargo update --precise` with the
+//! update-precise-breaking feature enabled. When the feature is stabilized,
+//! this file can be deleted.
+
+#![allow(deprecated)]
+
```
epage, line 6: Still trying to decide what the best approach is here but in the mean time, could you split this into two commits
- One that duplicates the tests
- One that enables the unstable feature

That way it shows what behavior change happened (or not) with the feature.

tests/testsuite/update.rs
```diff
@@ -2037,11 +2030,19 @@ fn update_precise_breaking_shared_non_ws() {
     p.cargo("update -Zupdate-precise-breaking -p shared@0.1.0 --precise 0.2.0")
         .masquerade_as_nightly_cargo(&[])
         .with_status(101)
+        // The lockfile update is failing for the same reason a non-breaking
+        // update would. There is still a shared@0.1.0 here that didn't get
+        // upgraded (the transitive dependency within bar), and we are now
+        // asking the lockfile update to update it to 0.2.0, which it cannot do.
         .with_stderr_data(str![[r#"
```
epage, line 2037: Like with `--breaking`, should we limit this to just the ones that can be upgraded?

Like other cases, this can be broken out into your task list

tests/testsuite/update.rs
```diff
@@ -2318,10 +2314,12 @@ fn update_precise_breaking_pinned() {
         .masquerade_as_nightly_cargo(&[])
         .with_status(101)
         .with_stderr_data(str![[r#"
-[ERROR] unknown `-Z` flag specified: update-precise-breaking
-
-For available unstable features, see https://doc.rust-lang.org/nightly/cargo/reference/unstable.html
-If you intended to use an unstable rustc feature, try setting `RUSTFLAGS="-Zupdate-precise-breaking"`
+[UPDATING] `dummy-registry` index
+[ERROR] failed to select a version for the requirement `pinned = "=0.1"`
+candidate versions found which didn't match: 0.2.0
```
epage: So `--precise` will error if the mentioned item can't upgrade but `--breaking` won't...

src/cargo/core/features.rs
torhovland: Because since there isn't any difference between the breaking command `cargo update foo --precise 2.0.1` and its non-breaking counterpart, I thought we needed a feature.

Should the breaking instead be `cargo update foo --breaking --precise 2.0.1`?

Or is it fine to just use `unstable-options` to allow the breaking update?

tests/testsuite/update.rs
```diff
@@ -1489,6 +1489,1195 @@ fn update_precise_yanked_multiple_presence() {
...
+
+    let crate_manifest_after = p.read_file("foo/Cargo.toml");
+    assert_e2e().eq(&crate_manifest_after, crate_manifest);
+
+    let lock_file_after = p.read_file("Cargo.lock");
+    assert_e2e().eq(&lock_file_after, lock_file);
+}
+
+#[cargo_test]
+fn update_precise_breaking_incompatible() {
```
epage: Is there a rest that makes sure `cargo update incompatible --precise 2.3.4` sets the version req correctly when 2.5.7 is present?

src/bin/cargo/commands/update.rs
```diff
@@ -1,8 +1,11 @@
+use std::collections::HashMap;
```
epage, line 1: Only did a quick glance at the tests; haven't gotten to the implementation yet.

tests/testsuite/update.rs
```diff
@@ -2016,51 +1927,1940 @@ fn update_breaking_specific_packages() {
...
+
+"#]])
+        .run();
+
+    // Invalid spec
+    p.cargo("update -Zupdate-precise-breaking -p incompatible@foo --precise 0.3.0")
+        .masquerade_as_nightly_cargo(&[])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] expected a version like "1.32"
```
epage, line 2101: In case my comment gets lost track of on a merged pr

Hadn't notice this before.  This error message is bad.  We should at least be calling out that we are parsing a package id spec

Existing behavior:
```console
$ cargo update clap@foo
error: invalid package ID specification: `clap@foo`

Caused by:
  expected a version like "1.32"
```
(again, task is list is fine)

torhovland:
> Except this PR description does not include an explanation as to why this PR addresses that and why an error is not the way forward

Description updated.

> (btw probably best not to mark things as done when they aren't merged as things can change especially when the "solution" is not decided yet)

Got it.

tests/testsuite/update.rs
epage: We still want an identifying label on all uses of `masquerade_as_nightly_cargo` so we can easily find all tests related to an unstable feature even if there isn't an "unstable feature" for it.

src/cargo/core/features.rs
epage: I view this as a new supported value within the option which could be used with `-Zunstable-options`.  If there is doing more than turning error cases into success cases, then we might want to consider a more explicit opt-in.

tests/testsuite/update.rs
```diff
@@ -1489,6 +1489,1195 @@ fn update_precise_yanked_multiple_presence() {
...
+
+                [dependencies]
+                incompatible = "1.0"
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("update -p incompatible --precise 2.0.0")
+        .masquerade_as_nightly_cargo(&[])
```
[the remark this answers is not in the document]
torhovland: I thought it was a list of features, but I see now it is a list of reasons. I'll update all the empty ones.

epage:
> Description updated.

I'm still not seeing the description talk to

> why an error is not the way forward

btw it would be good to focus on user impact and not implementation.  The paragraph that alludes to this starts off by discussing the implementation which someone is likely to gloss over when looking for user impact.

tests/testsuite/update.rs

src/bin/cargo/commands/update.rs
```diff
@@ -92,28 +95,39 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
...
         workspace: args.flag("workspace"),
         gctx,
     };
 
-    if args.flag("breaking") {
-        gctx.cli_unstable()
-            .fail_if_stable_opt("--breaking", 12425)?;
-
-        let upgrades = ops::upgrade_manifests(&mut ws, &update_opts.to_update)?;
-        ops::resolve_ws(&ws, update_opts.dry_run)?;
```
torhovland, line 106: With this change, it is now possible to revert the introduction of the `dry_run` argument in `resolve_ws`. Let me know what you think.

tests/testsuite/update.rs
```diff
@@ -1823,14 +1812,7 @@ fn update_precise_breaking_pre_release_cannot_upgrade_nonexplicit_version_req()
...
         .with_stderr_data(str![[r#"
-[UPDATING] `dummy-registry` index
-[ERROR] failed to select a version for the requirement `pre = "^0.1"`
-candidate versions found which didn't match: 0.2.0-beta
-location searched: `dummy-registry` index (which is replacing registry `crates-io`)
-required by package `foo v0.0.1 ([ROOT]/foo)`
-if you are looking for the prerelease package it needs to be specified explicitly
-    pre = { version = "0.2.0-beta" }
-perhaps a crate was updated and forgotten to be re-vendored?
+[ERROR] New requirement ^0.2 is invalid, because it doesn't match 0.2.0-beta
```
torhovland, line 1815: @epage I'm curious what you think of this.

src/cargo/ops/cargo_update.rs
```diff
@@ -14,6 +14,7 @@ use crate::util::toml_mut::manifest::LocalManifest;
 use crate::util::toml_mut::upgrade::upgrade_requirement;
 use crate::util::{style, OptVersionReq};
 use crate::util::{CargoResult, VersionExt};
+use anyhow::Context;
```
weihanglo, line 17: nit: `Context` is a name so general that may conflict with other structs.

```suggestion
use anyhow::Context as _;
```

tests/testsuite/update.rs
```diff
@@ -2357,31 +2357,31 @@ fn update_breaking_spec_version_transitive() {
...
+    Package::new("dep", "2.0.1").publish();
+    Package::new("dep", "3.0.0").publish();
 
     // Will upgrade the direct dependency
     p.cargo("update -Zunstable-options --breaking dep@1.0")
         .masquerade_as_nightly_cargo(&["update-breaking"])
         .with_stderr_data(str![[r#"
 [UPDATING] `[..]` index
-[UPGRADING] dep ^1.0 -> ^2.0
+[UPGRADING] dep ^1.0 -> ^3.0
```
weihanglo, line 3748: There is discrepancy between the normal update and breaking update. 

* With and without `v` prefix
* Show SemVer requirement operator or not.

Not a blocker for this PR, but we might want  track this somewhere.

tests/testsuite/update.rs
```diff
@@ -2357,31 +2357,31 @@ fn update_breaking_spec_version_transitive() {
...
             "#,
         )
         .file("bar/src/lib.rs", "")
         .build();
 
     p.cargo("generate-lockfile").run();
 
-    Package::new("dep", "1.1.1").publish();
-    Package::new("dep", "2.0.0").publish();
+    Package::new("dep", "2.0.1").publish();
```
weihanglo, line 3740: What's the motivation behind 45bc28c9363d90557a4379ca56f4ca59c0d28cd7?

src/cargo/util/semver_ext.rs
```diff
@@ -117,9 +117,9 @@ impl OptVersionReq {
     /// and we're not sure if this part of the functionality should be implemented in semver or cargo.
     pub fn matches_prerelease(&self, version: &Version) -> bool {
         if version.is_prerelease() {
-            let mut version = version.clone();
-            version.pre = semver::Prerelease::EMPTY;
-            return self.matches(&version);
+            let mut version_cleaned = version.clone();
+            version_cleaned.pre = semver::Prerelease::EMPTY;
+            return self.matches(&version) || self.matches(&version_cleaned);
```
torhovland, line 122: This needs more looking into. A test is failing. I'll also move this and related changes to a separate commit.

tests/testsuite/update.rs
```diff
@@ -1489,6 +1489,1381 @@ fn update_precise_yanked_multiple_presence() {
...
+                authors = []
+
+                [dependencies]
+                incompatible = "1.0"
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("update -p incompatible --precise 2.0.0")
```
weihanglo, line 1514: `-p` is not needed anymore :)

tests/testsuite/update.rs
```diff
@@ -1489,6 +1489,1381 @@ fn update_precise_yanked_multiple_presence() {
...
+
+    p.cargo("update -Zunstable-options renamed-from non-semver transitive-compatible transitive-incompatible --precise 2.0.0")
+        .masquerade_as_nightly_cargo(&["update-precise-breaking"])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[UPDATING] `[..]` index
+[ERROR] failed to select a version for the requirement `renamed-from = "^1.0"`
+candidate versions found which didn't match: 2.0.0
+location searched: `dummy-registry` index (which is replacing registry `crates-io`)
+required by package `foo v0.0.1 ([..]/foo)`
```
weihanglo, line 2813: `[..]` glob is not necessary.

tests/testsuite/update.rs
```diff
@@ -1489,6 +1489,1381 @@ fn update_precise_yanked_multiple_presence() {
...
+    p.cargo("update -Zunstable-options -p pre --precise 2.0.0")
+        .masquerade_as_nightly_cargo(&["update-precise-breaking"])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[UPDATING] `dummy-registry` index
+[ERROR] failed to select a version for the requirement `pre = "^1.0.0-alpha"`
+candidate versions found which didn't match: 2.0.0
+location searched: `dummy-registry` index (which is replacing registry `crates-io`)
+required by package `foo v0.0.1 ([ROOT]/foo)`
+perhaps a crate was updated and forgotten to be re-vendored?
```
weihanglo: Oh no this error message is bad 😢

tests/testsuite/update_duplicated_with_precise_breaking_feature.rs
weihanglo: Instead of duplicating tests, one approach we could take is using a test only env var to activate the unstable behavior, for example [`__CARGO_USE_GITOXIDE_INSTEAD_OF_GIT2`](https://doc.crates.io/contrib/tests/running.html#running-with-gitoxide-as-default-git-backend-in-tests).

(Yes I admit that this opens a door for people without `-Zunstable-options`, though we already have `__CARGO_TEST_CHANNEL_OVERRIDE_DO_NOT_USE_THIS` hence 😬)

src/cargo/core/features.rs
torhovland: Fixed.

tests/testsuite/update_duplicated_with_precise_breaking_feature.rs
torhovland: `__CARGO_USE_GITOXIDE_INSTEAD_OF_GIT2` is used from tests to modify Cargo behaviour. You would still need duplicate tests if you wanted to run Cargo with both options.

I was thinking the duplicate test file is OK, as it is meant to be temporary and can be deleted later.

I have put it in a separate commit, and it didn't need any modification after implementing the feature.

tests/testsuite/update.rs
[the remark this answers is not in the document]

tests/testsuite/update.rs
torhovland: I'm curious what you think the correct behaviour should be. We are upgrading `shared`, so we are not trying to keep it unchanged. So the update tries to do its thing, but fails. Anyway, I'll add this to the task list.

tests/testsuite/update.rs
torhovland: Unlike with `--breaking`, there is no command line difference between a breaking and a non-breaking precise update. So we just try to do any upgrades, and whether or not we did any, the same lockfile update is run. So a precise update may fall back to being non-breaking, and may fail in the same way as a precise update could break before.

If we changed the API to `--breaking --precise` we could stop if there aren't any upgrades.

tests/testsuite/update.rs
torhovland: Does adding `0.4.5` here cover it?

```
    Package::new("incompatible", "0.3.0").publish();
    Package::new("incompatible", "0.3.1").publish();
    Package::new("incompatible", "0.4.5").publish();

    p.cargo("update -Zunstable-options -p incompatible@0.1.0 --precise 0.3.0")
        .with_stderr_data(str![[r#"
[UPGRADING] incompatible ^0.1 -> ^0.3
[UPDATING] incompatible v0.1.0 -> v0.3.0
```

tests/testsuite/update.rs
torhovland: Added to list.

torhovland:
> > Description updated.
> 
> I'm still not seeing the description talk to
> 
> > why an error is not the way forward

I have updated the [task list](https://github.com/rust-lang/cargo/issues/12425#issuecomment-2186198258). Hope that's OK.

tests/testsuite/update.rs
torhovland: This was discussed here: https://github.com/rust-lang/cargo/pull/13979#discussion_r1628470396

I'll add it to the [task list](https://github.com/rust-lang/cargo/issues/12425#issuecomment-2186198258).

tests/testsuite/update.rs
torhovland: In this test (`update_breaking_spec_version_transitive`) I wanted to specifically target `foo.dep` but not `bar.dep`. I can do that if `foo.dep=1.0` and `bar.dep=2.0`:

```
    p.cargo("update -Zunstable-options --breaking dep@1.0")
        .with_stderr_data(str![[r#"
[UPGRADING] dep ^1.0 -> ^3.0
[UPDATING] dep v1.0.0 -> v3.0.0
```

But not if `bar.dep=1.1`, because that makes both `foo.dep` and `bar.dep` resolve to `1.1`.

tests/testsuite/update.rs
torhovland: Lower casing fixed. Awaiting feedback on the error case.

src/cargo/util/semver_ext.rs
torhovland: Two commits inserted at the beginning of the PR.

tests/testsuite/update.rs

tests/testsuite/update.rs

tests/testsuite/update.rs
torhovland: Keep in mind this is without the feature implemented. In the next commit this becomes:

```
[UPGRADING] pre ^1.0.0-alpha -> ^2.0.0
[UPDATING] pre v1.0.0-alpha -> v2.0.0
```

torhovland:
This is ready for another look now.

src/cargo/util/semver_ext.rs
```diff
@@ -117,9 +117,15 @@ impl OptVersionReq {
...
-            return self.matches(&version);
+            // Only in the case of "ordinary" version requirements with pre-release
+            // versions do we need to help the version matching. In the case of Any,
+            // Locked, or Precise, the `matches()` function is already doing the
+            // correct handling.
+            if let OptVersionReq::Req(_) = self {
+                let mut version = version.clone();
+                version.pre = semver::Prerelease::EMPTY;
+                return self.matches(&version);
+            }
```
epage, line 128: I don't think this is an appropriate place to be putting this fix.

Note the comment in the description

> we're not sure if this part of the functionality should be implemented in semver or cargo.

At this time, this is an abstraction that we shouldn't be breaking with cargo-specific logic

btw these two commtis seem unrelated with the rest and seem like they should be their own PR

tests/testsuite/update_duplicated_with_precise_breaking_feature.rs
epage: This was split out into its own commit but I'd like to see it split into two commits.

Currently, the commit is
```console
$ cp tests/testsuite/update.rs
$ edit tests/testsuite/update_duplicated_with_precise_breaking_feature.rs
$ git commit
```
I was asking to see
```console
$ cp tests/testsuite/update.rs
$ git commit
$ edit tests/testsuite/update_duplicated_with_precise_breaking_feature.rs
$ git commit
```
Depending on how well it diffs, maybe it should be more

This makes it easier to audit in this review, and in the future, what the differences are between the two tests files.

tests/testsuite/update.rs
epage: This is 

> Can we limit updates more in case of breaking upgrades? See https://github.com/rust-lang/cargo/pull/14140#discussion_r1656927375 and the test update_precise_breaking_shared_non_ws.

See https://github.com/rust-lang/cargo/issues/12425#issuecomment-2186198258

epage: I've renamed the task in the hope that it has enough context to stand on its own

> Should cargo update --precise <breaking> only attempt to update direct dependencies (which can succeed) or also transitive dependencies (which will typically fail) See https://github.com/rust-lang/cargo/pull/14140#discussion_r1656927375 and the test update_precise_breaking_shared_non_ws.

tests/testsuite/update.rs
epage: I've renamed the existing task to

>  Should cargo update --breaking error if the command completed without doing any upgrades? 

and added this as a parallel task

> Should cargo update --precise <breaking> error if the command completed without doing any upgrades?

tests/testsuite/update.rs
epage: That covers
- Compatible upgrade available
- Incompatible upgrade available

Seems to cover every case.

epage:
>  Except this PR description does not include an explanation as to why this PR addresses that and why an error is not the way forward
>
> Description updated.

The PR says:

> This PR is making a change that also affects cargo update --breaking. We'll reuse ops::update_lockfile() for both breaking and non-breaking updates. The benefit is more consistent output and behaviour between the two. In particular, it addresses a task about an error output if there is nothing to upgrade. See https://github.com/rust-lang/cargo/issues/12425#issuecomment-2186198258.

This is focused on the implementation, glosses over the reason, and doesn't say what is being changed.

I think its best to split this out from this PR, either making ti independent or making one depend on the other.  Trying to slip it in like this is making it harder to review and discuss both.

tests/testsuite/precise_pre_release.rs
```diff
@@ -65,11 +65,12 @@ fn update_pre_release() {
     p.cargo("update my-dependency --precise 0.1.2-pre.0 -Zunstable-options")
         .masquerade_as_nightly_cargo(&["precise-pre-release"])
         .with_stderr_data(str![[r#"
+[UPGRADING] my-dependency ^0.1.1 -> ^0.1.2-pre.0
 [UPDATING] `dummy-registry` index
-[UPDATING] my-dependency v0.1.1 -> v0.1.2-pre.0
 
 "#]])
```
epage, line 71: This seems to incorrectly be interacting with the precise-prerelease feature.  The intention of that feature is that  we should occasionally allow matching pre-releases with a regular version requirement.

tests/testsuite/update.rs
```diff
@@ -1713,14 +1709,11 @@ fn update_precise_breaking_consistent_output() {
...
         .with_stderr_data(str![[r#"
+[UPGRADING] incompatible ^0.1 -> ^0.2
 [UPDATING] `dummy-registry` index
-[ERROR] failed to select a version for the requirement `incompatible = "^0.1"`
-candidate versions found which didn't match: 0.2.0
-location searched: `dummy-registry` index (which is replacing registry `crates-io`)
-required by package `foo v0.0.1 ([ROOT]/foo)`
-perhaps a crate was updated and forgotten to be re-vendored?
+[UPDATING] incompatible v0.1.0 -> v0.2.0
+[NOTE] pass `--verbose` to see 1 unchanged dependencies behind latest
```
epage, line 1716: > `update_precise_breaking_consistent_output`

Consistent with what?  How will this ensure we keep remaining consistent?

tests/testsuite/update.rs
epage: Why is this an error?
- manifest points to `0.1`
- `0.2.0-beta` is published
- user does `cargo update --precise 0.2.0-beta`

We gave it a valid version that it a breaking change from the current one, so it seems like we should just use that as our version requirement.

The fact that the error doesn't make sense from the users perspective ("I never said the version req was `^0.2` so why is it trying that?") I think helps highlight that.

epage: Looks like its only an error because the user's req was `0.1` instead of `0.1.0`?  In that case, I think its reasonable for us to act as if it was `0.1.0`.

tests/testsuite/update.rs
```diff
@@ -2255,14 +2218,10 @@ fn update_precise_breaking_spec_version() {
     // No spec
     p.cargo("update -Zunstable-options incompatible --precise 0.3.1")
```
epage, line 2219: > `update_precise_breaking_spec_version`

I'm confused as to what this test is trying to capture compared to other tests

epage: Oh, its not that there is no spec but that we aren't specifying a version in the spec, only the name

tests/testsuite/update.rs
```diff
@@ -2693,14 +2648,10 @@ fn update_precise_breaking_incompatible_build_metadata() {
     p.cargo("generate-lockfile").run();
     p.cargo("update -Zunstable-options incompatible --precise 2.0.0+b")
```
epage, line 2649: Does `cargo update --precise <no-meta>` work on packages with metadata?  Whichever way, this should do the same if its a breaking change

src/cargo/ops/cargo_update.rs
```diff
@@ -165,7 +182,15 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
...
-    let keep = |p: &PackageId| !to_avoid_sources.contains(&p.source_id()) && !to_avoid.contains(p);
+    let breaking_precise_upgrades = opts.precise.is_some() && !upgrades.is_empty();
+    let breaking_mode = opts.breaking || breaking_precise_upgrades;
+
+    let keep = |p: &PackageId| {
+        (!to_avoid_sources.contains(&p.source_id()) && !to_avoid.contains(p))
+            // In case of `--breaking` or precise upgrades, we want to keep all
+            // packages unchanged that didn't get upgraded.
+                || (breaking_mode && !upgrades.contains_key(&(p.name().to_string(), p.source_id())))
+    };
```
epage, line 193: @torhovland I was very intentionally trying to avoid changing this in past PRs.  To go back and change it requires a good explanation.  Even better if you can also split this into its own commit or PR

@Eh2406 I'm very cautious of messing with `keep` and would appreciate your input on tweaking it for this case

src/cargo/ops/cargo_update.rs
```diff
@@ -338,32 +364,60 @@ fn upgrade_dependency(
         None
     };
 
-    let Some(latest) = latest else {
+    let new_version = if let Some(precise) = precise {
+        Version::parse(precise)
+            .with_context(|| format!("invalid version format for precise version `{precise}`"))?
```
epage, line 369: I'm not seeing this error covered in the tests

src/cargo/ops/cargo_update.rs
```diff
@@ -338,32 +364,60 @@ fn upgrade_dependency(
...
+        "Downgrading"
+    } else {
+        "Upgrading"
+    };
+
     if upgrade_messages.insert(upgrade_message.clone()) {
         gctx.shell()
-            .status_with_color("Upgrading", &upgrade_message, &style::GOOD)?;
+            .status_with_color(status, &upgrade_message, &style::WARN)?;
     }
```
epage, line 406: Why did the style for `Upgrading` change?

src/cargo/util/toml_mut/upgrade.rs
```diff
@@ -27,15 +29,8 @@ pub(crate) fn upgrade_requirement(
             new_req_text.remove(0);
         }
         // Validate contract
-        #[cfg(debug_assertions)]
-        {
-            assert!(
```
epage, line 32: With my other comment in mind, is there a reason this should be a `bail`, rather than an `assert`?

src/cargo/ops/cargo_update.rs
Eh2406: `keep` also makes my head spin. I feel like it's something about the triple negative structure, we update the ones that are not `keep`, we `keep` the ones that are not `avoid`, we `avoid` the things that asked to be updated. But I've never found it clearer naming structure, so maybe that's not it. It's also trying to do a linear prediction of resolution which feels like an approximation.

In specific, why does this need to be a separate lookup as opposed to adding the matching packages to `to_avoid`?

Also, this callback is run a lot on no-op builds, so very small performance problems end up being a high percentage of runtime. I suspect the `to_string` is going to end up being expensive. Can the `upgrades` map be `InternedStrings` or `&str`?

torhovland: I'm pulling this out into a separate PR: https://github.com/rust-lang/cargo/pull/14259

epage: I assume with that pulled out this PR will be updated so we can get it merged without #14259 so this doesn't get blocked on that decision?

torhovland: I don't like the idea of not building on #14259. I suggest we rather let this PR depend on #14259 and leave it as draft for now. Here's why:

- Some of the duplicated existing precise tests with the unstable feature enabled are no longer passing. In other words, without #14259 we will be changing the behaviour of the non-breaking precise update.

- Some invocations are now silently finishing rather than reporting an error. For example:

```
Did you mean one of these?

  incompatible@0.3.1
```

- The output is generally no longer consistent with the non-breaking update. For example, without #14259 the output changes like this:

```
-[UPDATING] incompatible v0.1.0 -> v0.2.0
-[NOTE] pass `--verbose` to see 1 unchanged dependencies behind latest
+[UPDATING] incompatible v0.1.0 -> v0.2.0 (latest: v0.2.1)
```

src/cargo/util/semver_ext.rs
torhovland: Hmm, but all I'm doing here is relaxing the cargo-specific logic. Without this fix, I get this bug:

```
   Upgrading pre ^0.1.0 -> ^0.2.0-beta
error: failed to select a version for the requirement `pre = "^0.2.0-beta"`
candidate versions found which didn't match: 0.2.0-beta
```

What do you think I should do?

epage: If we take `0.2.0-beta`, remove `-beta`, we get `0.2.0`.  That seems like `0.2.0-beta` should be able to match that.

Any insight into why this isn't working?

src/cargo/ops/cargo_update.rs
epage: > I don't like the idea of not building on https://github.com/rust-lang/cargo/pull/14259. I suggest we rather let this PR depend on https://github.com/rust-lang/cargo/pull/14259 and leave it as draft for now.

That PR has behavior changes that have not been agreed to, internal and user-facing.  We should not block progress forward on something on changes like that.  If there is some intermediate change that side steps the controversial parts that unblocks `--precise`, then let's go for that.

torhovland: Is #14259 still controversial? It no longer changes `keep`. I hesitate moving forward with this PR without merging #14259 first, particularly because of the first bullet point above (breaking compatibility with the existing precise update).

src/cargo/util/semver_ext.rs
linyihai: This is a bug for this PR.

`matches_prerelease` will alaways a `OptVersionReq::Req` in my previous [PR](https://github.com/rust-lang/cargo/pull/13626),  but this PR will add `OptVersionReq::Precise` to match , see 

```

    let req = if precise.is_some() {
        OptVersionReq::Precise(new_version, new_version_req)
    } else {
        OptVersionReq::Req(new_version_req)
    };
```

tests/testsuite/update.rs
```diff
@@ -1390,103 +1390,1501 @@ fn update_precise_git_revisions() {
...
+required by package `bar v0.0.1 ([ROOT]/foo/bar)`
+    ... which satisfies path dependency `bar` (locked to 0.0.1) of package `foo v0.0.1 ([ROOT]/foo)`
+perhaps a crate was updated and forgotten to be re-vendored?
+
+"#]])
+        .run();
+}
+
+#[cargo_test]
+fn update_precise_breaking_direct_plus_transitive() {
```
torhovland, line 2568: @epage Can I ask you to take a quick look at this test, please? I believe it answers the concern you raised in office hours. The test demonstrates that we do support upgrading a direct dependency while leaving the transitive dependency at its original version.

If this is reassuring to you, it means #14259 is a valid step forward and we should get it merged. Then I'll clean up this PR.

Issue #14081, opened 2024-06-16
Tracking Issue for merging `cargo-information`/`cargo info` to Cargo


Rustin170506:
### Summary

Rust Internals: https://internals.rust-lang.org/t/feedback-on-cargo-info-to-prepare-it-for-merging/20369
Original issue: https://github.com/rust-lang/cargo/issues/948
Implementation: TBD
Documentation: TBD

We developed 'cargo-information' as a solution for #948

**Features:**

- Works with all registries that are compatible with Cargo
- Fetches and displays basic package information (name, version, owners, etc.)
- Shows package dependencies and their versions
- Provides information about package features

And in the last six months, we have released 8 versions and had 3,428 downloads. At the same time, we have covered almost all the features it needs. So I think it is a good time to merge it back into Cargo and prove it to all Cargo users.

### Unresolved Issues

* [ ] We want to report some crates.io metrics in this command as well, we will do that after we merge it back into Cargo See: https://github.com/hi-rustin/cargo-information/issues/20 
* [ ] After the public dependency feature is stable, we should consider what dependencies we want to show in this command. See: https://github.com/hi-rustin/cargo-information/issues/138

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

epage:
Curious, what is the role of this tracking issue.  I'm assuming we can't have this feature be unstable and instead must insta-stable it like `cargo add` and `cargo remove`.  At which point, we are just tracking the command being merged which the original issue can handle.  Likely, the tasks listed are not blockers.  This is a UX heavy command and we have a lot of flexibility of adding / removing functionality.

weihanglo:
As a hub connecting to other issues? That said, we can create a label `Command-info` for this purpose.

Issue #948, opened 2014-11-21
Something like `cargo query` or `cargo info` for information about a registry package


tomjakubowski:
It would be nice to be able to get the description, latest version(s), author, license, and the like, for a given registry package. Prior art includes `npm view` and `gem query`.

Somewhat related/complementary to #925

felixc:
Now that cargo supports external installable sub-commands, does this make more sense to add via that mechanism, rather than being built-in?

That said, given that `cargo search` presumably already does most of this, maybe it just needs to get some new flags to print more info?

alexcrichton:
@felixc yeah I'd be in favor of seeing this bake externally first

g-k:
I gave this a shot: https://github.com/g-k/cargo-show.  I'm curious about how people might use it and how to improve it for them.

For example, I'd like to cache package GET requests like crates.io, but I'm not sure if it'd be better to use a sub command specific cache like `~/.cargo-show/` or a shared cache that other sub commands could use like `~/.cargo/cache/package-details/`.

PaulDance:
@tomjakubowski Is it still desirable today considering the possibility of simply using the linked tool?

epage:
There is also https://crates.io/crates/cargo-info which has stalled out.

epage:
A thought I had on this command is that it `cargo info <dep>` would choose the latest version of `<dep>` in the associated `Cargo.lock` file and show that.  If the user specifies `cargo info <dep>@<ver>`, `ver` would act like `cargo install` and would be a `=` version requirement.

We could also show either `cargo tree -i <dep>` or tell the user about the command.  Similarly, `cargo search` could suggest `cargo info`

We could also show feature activation, either baesd on if `<dep>` is a direct dependency of the relevant manifest (showing a view like `cargo add`) or based on resolving the dependency graph (behaving more like `cargo tree`).

Some prior art
- https://python-poetry.org/docs/cli/#show
- https://yarnpkg.com/cli/info
- https://bundler.io/v2.2/man/bundle-info.1.html
- https://github.com/luarocks/luarocks/wiki/show
- https://lib.haxe.org/documentation/using-haxelib/#info
- https://docs.racket-lang.org/pkg/cmdline.html#(part._raco-pkg-show)

epage:
Looks like #6666 is covered by existing commands but could still be useful to either highlight those commands or to include their output

yugaego:
[...]

PR #11204, opened 2022-10-09
Add completions for `cargo remove`

cassaundra:
### What does this PR try to resolve?

This PR continues the deferred work of #11099 by adding bash and zsh completions for the cargo remove subcommand.

### How should we test and review this PR?

There doesn't seem to be much in the way of testing for these completions automatically, so I would suggest verifying that they work in practice and sufficiently cover the subcommand's surface area.

### Additional Information

I will also soon post a PR for cargo remove's documentation.

PR #14141, opened 2024-06-25
feat: Add `info` cargo subcommand

Rustin170506:
close https://github.com/rust-lang/cargo/issues/14081
close https://github.com/rust-lang/cargo/issues/948

fcp https://github.com/rust-lang/cargo/pull/14141#issuecomment-2243683924

This PR added a new `info` cargo subcommand.

# Background

This adds a new subcommand to Cargo, `cargo info`. This subcommand would allow users to get information about a crate from the command line, without having to go to the web.

The main motivation for this is to make it easier to get information about a crate from the command line. Currently, the way to get information about a crate is to go to the web and look it up on [crates.io] or find the crate's source code and look at the `Cargo.toml` file. This is not very convenient, especially not all information is displayed on the [crates.io] page.
This command also has been requested by the community for a long time. You can find more discussion about this in [cargo#948].

Another motivation is to make the workflow of finding and evaluating crates more efficient. In the current workflow, users can search for crates using `cargo search`, but then they have to go to the web to get more information about the crate. This is not very efficient, especially if the user is just trying to get a quick overview of the crate. This would allow users to quickly get information about a crate without having to leave the terminal.

[crates.io]: https://crates.io
[cargo#948]: https://github.com/rust-lang/cargo/issues/948

Example usage:

```console
./target/debug/cargo info clap --verbose
  Credential cargo:token get crates-io
clap #argument #cli #arg #parser #parse
A simple to use, efficient, and full-featured Command Line Argument Parser
version: 4.5.8 (latest 4.5.9)
license: MIT OR Apache-2.0
rust-version: 1.74
documentation: https://docs.rs/clap/4.5.8
repository: https://github.com/clap-rs/clap
crates.io: https://crates.io/crates/clap/4.5.8
features:
 +default         = [std, color, help, usage, error-context, suggestions]
  color           = [clap_builder/color]
  error-context   = [clap_builder/error-context]
  help            = [clap_builder/help]
  std             = [clap_builder/std]
  suggestions     = [clap_builder/suggestions]
  usage           = [clap_builder/usage]
  cargo           = [clap_builder/cargo]
  debug           = [clap_builder/debug, clap_derive?/debug]
  deprecated      = [clap_builder/deprecated, clap_derive?/deprecated]
  derive          = [dep:clap_derive]
  env             = [clap_builder/env]
  string          = [clap_builder/string]
  unicode         = [clap_builder/unicode]
  unstable-doc    = [clap_builder/unstable-doc, derive]
  unstable-styles = [clap_builder/unstable-styles]
  unstable-v5     = [clap_builder/unstable-v5, clap_derive?/unstable-v5, deprecated]
  wrap_help       = [clap_builder/wrap_help]
dependencies:
 +clap_builder@=4.5.8
  clap_derive@=4.5.8
owners:
  kbknapp (Kevin K.)
  github:rust-cli:maintainers (Maintainers)
  github:clap-rs:admins (Admins)
note: to see how you depend on clap, run `cargo tree --invert --package clap@4.5.8`
```

image

*note:* this is showing the `--verbose` output to show every thing the user can possibly see.  Normal operation does not include
- dependencies

## Detailed design

| Content                                                                    | Explanation                         | Why                                                                               |
|----------------------------------------------------------------------------|-------------------------------------|-----------------------------------------------------------------------------------|
| clap                                                                       | Name                                | The basic information.                                                            |
| #argument #cli #arg #parser #parse                                         | Keywords (clickable)                           | It's more like a category, which you can use to search for relevant alternatives. |
| A simple to use, efficient, and full-featured Command Line Argument Parser | Description                         | The basic information.                                                            |
| version: 4.5.8 (latest 4.5.9)                                                           | Version                             | The basic information.                                                            |
| license: MIT OR Apache-2.0                                                 | License                             | When choosing a crate, it is crucial to consider the license.                     |
| rust-version: 1.74                                                       | MSRV                                | When choosing a crate, it is crucial to make sure it can work with your MSRV.     |
| documentation: <https://docs.rs/clap/4.5.8>                               | Documentation Link                  | Use these links can find more docs and information.                               |
| repository: <https://github.com/clap-rs/clap>                              | Repo Link                           | Use these links can find more docs and information.                               |
| crates.io: https://crates.io/crates/clap/4.5.8                             | crates.io Link                           | Use these links can find more docs and information.                               |
| features:                                                                  | Default Features And Other Features | It helps for enabling features.                                                   |
| dependencies:                                                              | All dependencies                    | It indicates what it depends on.                                                  |
| owners:                                                                    | Owners                              | It indicates who maintains the crate.                                             |
| note: to see how you depend on clap, run `cargo tree --invert --package clap@4.5.8`                                                                   | A note for cargo tree command                              | It will prompt the user that the package is depended on under the workspace, and the dependencies can be viewed using the cargo tree command.           |

## Rendering features

- For features enabled by users, a + prefix and colored output are now used for better visibility.
- For features enabled automatically, colored output is used to distinguish them.
- For disabled features, non-colored output is used to clearly indicate their status.

## Rendering deps

Only show dependencies in verbose mode.

- For dependencies required by the package, a + prefix and colored output are now used for better visibility.
- For dependencies that are optional and activated, colored output is used to distinguish them.
- For dependencies that are optional and not activated, non-colored output is used to clearly indicate their status.

# Some important notes

## Downloading the crate from any Cargo compatible registry

The `cargo info` command will download the crate from any Cargo compatible registry. It will then extract the information from the `Cargo.toml` file and display it in the terminal.

If the crate is already in the local cache, it will not download the crate again. It will get the information from the local cache.

## Pick the correct version from the workspace

When executed in a workspace directory, the cargo info command chooses the version that the workspace is currently using.

If there's a lock file available, the version from this file will be used. In the absence of a lock file, the command attempts to select a version that is compatible with the Minimum Supported Rust Version (MSRV). And the lock file will be generated automatically.

The following hierarchy is used to determine the MSRV:

- First, the MSRV of the parent directory package is checked, if it exists.
- If the parent directory package does not specify an MSRV, the minimal MSRV of the workspace is checked.
- If neither the workspace nor the parent directory package specify an MSRV, the version of the current Rust compiler (rustc --version) is used.

# Prior art

## NPM

[npm] has a similar command called `npm info`.
For example:

```console
$ npm info lodash

lodash@4.17.21 | MIT | deps: none | versions: 114
Lodash modular utilities.
https://lodash.com/

keywords: modules, stdlib, util

dist
.tarball: https://registry.npmjs.org/lodash/-/lodash-4.17.21.tgz
.shasum: 679591c564c3bffaae8454cf0b3df370c3d6911c
.integrity: sha512-v2kDEe57lecTulaDIuNTPy3Ry4gLGJ6Z1O3vE1krgXZNrsQ+LFTGHVxVjcXPs17LhbZVGedAJv8XZ1tvj5FvSg==
.unpackedSize: 1.4 MB

maintainers:
- mathias <mathias@qiwi.be>
- jdalton <john.david.dalton@gmail.com>
- bnjmnt4n <benjamin@dev.ofcr.se>

dist-tags:
latest: 4.17.21

published over a year ago by bnjmnt4n <benjamin@dev.ofcr.se>
```

[npm]: https://www.npmjs.com/

## Poetry

[Poetry] has a similar command called `poetry show`.

For example:

```console
$ poetry show pendulum

name        : pendulum
version     : 1.4.2
description : Python datetimes made easy

dependencies
 - python-dateutil >=2.6.1
 - tzlocal >=1.4
 - pytzdata >=2017.2.2

required by
 - calendar >=1.4.0
```

[Poetry]: https://python-poetry.org/

# insta-stable

As @weihanglo mentioned in https://github.com/rust-lang/cargo/pull/14141#issuecomment-2189223552, commands that shadow third-party commands tend to be insta-stabilized to avoid an intermediate period where users can't access the third-party command (built-ins get priority) nor the built-in command (requires nightly)

For the cargo-info command, there are two commands that this would shadow
- [cargo-information](https://github.com/hi-rustin/cargo-information): hasn't been around too long and only has 4k downloads
- [cargo-info](https://gitlab.com/imp/cargo-info) : been around longer and has 63k downloads

We might be able to get away with having this unstable but starting from the assumption of insta-stabilization.


weihanglo:
I remembered last time when we imported new commands we insta-stablized them (`cargo-add` and `cargo-remove`) because built-in commands shadow external ones, making them unusable.

If the Cargo team are happy with making this a built-in command, I suppose we could do the same thing.

src/bin/cargo/commands/info.rs
```diff
@@ -0,0 +1,39 @@
+use cargo::ops::cargo_info::info;
+use cargo::util::command_prelude::*;
+use cargo_util_schemas::core::PackageIdSpec;
+
+pub fn cli() -> Command {
+    Command::new("info")
+        .about("Display information about a package in the registry")
+        .arg(
+            Arg::new("package")
```
hi-rustin, line 9: I am not sure whether we need to use `crate` here.

src/cargo/ops/cargo_info/mod.rs
```diff
@@ -0,0 +1,398 @@
...
+    /// The source for crates.io may be replaced by a built-in source for accessing crates.io with
+    /// the sparse protocol, or a source for the testing framework (when the replace_crates_io
+    /// function is used)
+    ///
+    /// User-defined source replacement is not applied.
+    /// Note: This will be utilized when interfacing with the registry API.
+    replacement: SourceId,
+}
+
+fn get_source_id(
```
hi-rustin, line 256: This function is pretty similar with https://github.com/hi-rustin/cargo/blob/476969f3cfa4ec2fd8ca202132ce0405093aa863/src/cargo/ops/registry/mod.rs#L187

I am not sure whether we should extend the original code or try to move out the `package_id` check here.

hi-rustin: Even if we decided to move the `package_id` check out, we still cannot use [registry](https://github.com/hi-rustin/cargo/blob/476969f3cfa4ec2fd8ca202132ce0405093aa863/src/cargo/ops/registry/mod.rs#L117) directly. Because it calls the original `get_source_id`.

src/bin/cargo/commands/info.rs
hi-rustin: We do inspect the crate, but we support package ID as a specification. So I don't know which term we should use here.

src/cargo/ops/cargo_info/mod.rs
hi-rustin: So I prefer to extend [registry](https://github.com/hi-rustin/cargo/blob/476969f3cfa4ec2fd8ca202132ce0405093aa863/src/cargo/ops/registry/mod.rs#L117) and [get_source_id](https://github.com/hi-rustin/cargo/blob/476969f3cfa4ec2fd8ca202132ce0405093aa863/src/cargo/ops/registry/mod.rs#L187) to support specify the package ID.

hi-rustin: @epage Do you have any thoughts about it? Thanks!

src/bin/cargo/commands/info.rs
epage: Are you referring to `Arg::new(name)`?  The `name` is only for internal processing purposes though I'd be consistent with other SPECs.

src/cargo/ops/cargo_info/mod.rs
epage: So the idea is to extract the `get_source_id` out of `registry`?  That seems reasonable.

hi-rustin: I think it makes sense to move the info ops into the registry rather than extracting it out of the registry. What I really want to do is try to extend the registry API to support the use of source IDs from a specific package. I did it in:[`55db322` (#14141)](https://github.com/rust-lang/cargo/pull/14141/commits/55db322a9187aa5d099fc7b4c922948342609cdb)

I am not sure if this is a really good API design. Maybe we should consider moving the `use_package_source_id` into the RegistrySourceIds instead of exposing it directly as a tuple field?

I know I should update the comments of these APIs. I will do it after you take a quick glance to make sure I am on the right track.

src/cargo/ops/registry/info/view.rs
```diff
@@ -169,7 +169,7 @@ fn pretty_source(source: SourceId, ctx: &GlobalContext) -> String {
         .local_path()
         .and_then(|path| pathdiff::diff_paths(path, ctx.cwd()))
     {
-        let path = std::path::Path::new(".").join(relpath);
+        let path = std::path::Path::new("../../../../..").join(relpath);
```
epage, line 172: This doesn't look right

src/cargo/ops/registry/mod.rs
```diff
@@ -116,12 +119,13 @@ impl RegistryCredentialConfig {
...
     gctx: &GlobalContext,
+    package_id: Option<PackageId>,
     token_from_cmdline: Option<Secret<&str>>,
     reg_or_index: Option<&RegistryOrIndex>,
     force_update: bool,
     token_required: Option<Operation<'_>>,
-) -> CargoResult<(Registry, RegistrySourceIds)> {
-    let source_ids = get_source_id(gctx, reg_or_index)?;
+) -> CargoResult<(Registry, bool, RegistrySourceIds)> {
+    let (use_package_source_id, source_ids) = get_source_id(gctx, package_id, reg_or_index)?;
```
epage, line 128: Not thrilled with this change.

As I mentioned in the other comment, it seems like we use some parameters only to call `get_source_id`.  What if instead we moved `get_source_id` out of this function, passing whats needed into it?

src/cargo/ops/registry/mod.rs
```diff
@@ -186,12 +191,35 @@ fn registry(
...
     gctx: &GlobalContext,
+    package_id: Option<PackageId>,
     reg_or_index: Option<&RegistryOrIndex>,
-) -> CargoResult<RegistrySourceIds> {
-    let sid = match reg_or_index {
-        None => SourceId::crates_io(gctx)?,
-        Some(RegistryOrIndex::Index(url)) => SourceId::for_registry(url)?,
-        Some(RegistryOrIndex::Registry(r)) => SourceId::alt_registry(gctx, r)?,
+) -> CargoResult<(bool, RegistrySourceIds)> {
+    let (use_package_source_id, sid) = match (&reg_or_index, package_id) {
```
epage, line 197: Also not too thrilled with this.  I haven't dug into this but is there a way we can split this into layers?  If not, should we just keep these as two functions?

src/cargo/ops/registry/info/view.rs
hi-rustin: Oops. I believe it changed when I tried to drag it into the subdirectory.

src/cargo/ops/registry/mod.rs
hi-rustin: Sounds good!

src/cargo/ops/registry/mod.rs
```diff
@@ -116,13 +118,12 @@ impl RegistryCredentialConfig {
 /// * `token_required`: If `true`, the token will be set.
 fn registry(
     gctx: &GlobalContext,
+    source_ids: &RegistrySourceIds,
```
epage, line 121: Should we split this into a separate PR?  At least when I did command merges, I tried to keep the merge-PR minimal.

src/cargo/ops/registry/info/view.rs
```diff
@@ -1,7 +1,9 @@
 use std::collections::HashMap;
 use std::io::Write;
 
+use crate::core::FeatureValue;
 use crate::util::style::{ERROR, HEADER, LITERAL, NOP, NOTE, WARN};
+use crate::util::OptVersionReq;
```
epage: If we're doing a refactor like this, could you instead do it against `cargo-information`s repo to keep this PR minimal?

src/cargo/ops/registry/mod.rs
```diff
@@ -191,25 +191,11 @@ fn get_source_id(
...
-        Some(RegistryOrIndex::Registry(r)) => SourceId::alt_registry(gctx, r)?,
-    };
-    // Load source replacements that are built-in to Cargo.
-    let builtin_replacement_sid = SourceConfigMap::empty(gctx)?
-        .load(sid, &HashSet::new())?
-        .replaced_source_id();
-    let replacement_sid = SourceConfigMap::new(gctx)?
-        .load(sid, &HashSet::new())?
-        .replaced_source_id();
+    let sid = get_initial_source_id(gctx, reg_or_index)?;
```
epage, line 194: ditto for this refactor.

src/bin/cargo/commands/info.rs
```diff
@@ -0,0 +1,39 @@
+use cargo::ops::info;
+use cargo::util::command_prelude::*;
+use cargo_util_schemas::core::PackageIdSpec;
+
+pub fn cli() -> Command {
+    Command::new("info")
```
epage, line 6: Where are the integration tests for this command?

src/bin/cargo/commands/info.rs
```diff
@@ -0,0 +1,39 @@
+use cargo::ops::cargo_info::info;
+use cargo::util::command_prelude::*;
+use cargo_util_schemas::core::PackageIdSpec;
+
+pub fn cli() -> Command {
```
epage, line 5: Do we have a "new command proposal" somewhere for us to have an FCP on?  For cargo add, it was the PR description itself, see https://github.com/rust-lang/cargo/pull/10472

Looks like we had our insta-stable conversation at https://github.com/rust-lang/cargo/pull/10472#issuecomment-1076778255

This is a bit different of a case because there is a command unrelated to this effort currently that we'd be shadowing.

src/cargo/ops/registry/mod.rs
hi-rustin: Ok, I will split this into a separate PR.

src/cargo/ops/registry/info/view.rs
hi-rustin: Let me revert it. I think the old one also is OK.

src/bin/cargo/commands/info.rs
hi-rustin: I will add it later. I just to make sure the main change is on the right track first.

src/bin/cargo/commands/info.rs
hi-rustin: > This is a bit different of a case because there is a command unrelated to this effort currently that we'd be shadowing.

Do you mean https://crates.io/crates/cargo-info?

epage: Correct.  It might not be a blocker but we need to make sure we call it out

src/bin/cargo/commands/info.rs
epage: We should also call out in the PR if the docs and completions will be merged in this PR or immediately after (with enough time before a beta branch that they don't get split).

I split them for `cargo add` mostly because the CLI was under such heavy churn during the review.  That is less likely to happen here so it seems reasonable to have docs and completions included.

epage:
Overall, seems ready to go for adding in tests, completions, and docs.

src/bin/cargo/commands/info.rs
hi-rustin: I have added it to the PR description.

hi-rustin (commented):

tests/testsuite/cargo_info/git_dependency/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+    let project_root = foo.root();
+    let cwd = &project_root;
+
+    snapbox::cmd::Command::cargo_ui()
+        .arg("info")
+        .arg_line("--verbose foo")
+        .current_dir(cwd)
+        .assert()
+        .success()
+        .stdout_eq(file!["stdout.term.svg"])
```
epage, line 42: For empty asserts, we switched to `""`

tests/testsuite/cargo_info/git_dependency/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+        .stderr_eq(file!["stderr.term.svg"]);
```
epage, line 43: For empty asserts, we switched to `""`

src/doc/man/cargo-info.md
```diff
@@ -0,0 +1,81 @@
...
+* For features enabled automatically, colored output is used to distinguish them.
+* For disabled features, non-colored output is used to clearly indicate their status.
+
+Dependency information is displayed only in verbose mode and follows the following principles:
+
+* For dependencies required by the package, a _+_ prefix and colored output are now used for better visibility.
+* For dependencies that are optional and activated, colored output is used to distinguish them.
+* For dependencies that are optional and not activated, non-colored output is used to clearly indicate their status.
+
+If the specified package is a package in the current workspace, it displays the information in the local Cargo.toml.
```
epage, line 38: This should be moved up with the MSRV part as its all related to the package selection

src/doc/man/cargo-info.md
```diff
@@ -0,0 +1,81 @@
...
+
+* For features enabled by users, a _+_ prefix and colored output are now used for better visibility.
+* For features enabled automatically, colored output is used to distinguish them.
+* For disabled features, non-colored output is used to clearly indicate their status.
+
+Dependency information is displayed only in verbose mode and follows the following principles:
+
+* For dependencies required by the package, a _+_ prefix and colored output are now used for better visibility.
+* For dependencies that are optional and activated, colored output is used to distinguish them.
+* For dependencies that are optional and not activated, non-colored output is used to clearly indicate their status.
```
epage, line 36: - This is unrelated to the section its in
- Do we need to explain the symbols? 
- Unsure how we should document dependencies.  Under `--verbose`?  Not at all?  In the high level description?

epage:
@hi-rustin I modified the PR description
- Added a note on the example usage about `--verbose`
- Made some clarifications to the stabilization plan

epage:
@rfcbot fcp merge

This insta-stabilizes a `cargo info <pkgname>` command, as suggested in #948 and with parallels in other ecosystems.  The command is a UI-focused command (few compatibility constraints) with a focus on what the caller-facing side of a library.

tests/testsuite/cargo_info/git_dependency/mod.rs
hi-rustin: Got it. Thank you! Updated.

src/doc/man/cargo-info.md
hi-rustin: Moved. Thank you!

src/doc/man/cargo-info.md
hi-rustin: > * This is unrelated to the section its in

I've added a new sub-section in the description called `Feature and Dependency Display Principles`. Does this make sense? What do you think?

> * Do we need to explain the symbols?

I don't think it's necessary to explain that separately. I believe we already explained its meaning in the statement.


> * Unsure how we should document dependencies.  Under `--verbose`?  Not at all?  In the high level description?

I think putting it in the high-level description would make it easier to find.

epage: > I don't think it's necessary to explain that separately. I believe we already explained its meaning in the statement.

More my point is whether we even should have a "Feature and Dependency Display Principles" section

> I think putting it in the high-level description would make it easier to find.

For me, when i said "high level description", I meant under the "Description" section

src/doc/man/cargo-info.md
```diff
@@ -0,0 +1,83 @@
...
+If the specified package is a package in the current workspace, it displays the information in the local Cargo.toml.
+
+If no version is specified, it will select the appropriate version based on the Minimum Supported Rust Version (MSRV)
+The following hierarchy is used to determine the MSRV:
+
+* First, the MSRV of the parent directory package is checked, if it exists.
+* If the parent directory package does not specify an MSRV, the minimal MSRV of the workspace is checked.
+* If neither the workspace nor the parent directory package specify an MSRV, the version of the current Rust compiler (
+  rustc --version) is used.
+
```
epage, line 27: imo its unnecessary to go into this much detail and is likely to go stale as we evolve things.

hi-rustin: I worry about it too. I just put too many implementations here. Deleted.

src/doc/src/commands/manifest-commands.md
```diff
@@ -1,5 +1,7 @@
 # Manifest Commands
+
 * [cargo add](cargo-add.md)
+* [cargo_info](cargo-info.md)
 * [cargo generate-lockfile](cargo-generate-lockfile.md)
```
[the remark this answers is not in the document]
hi-rustin: Good catch!

src/doc/man/cargo-info.md
```diff
@@ -0,0 +1,83 @@
...
+This command displays information about a package in the registry. It fetches data from the package's Cargo.toml file
+and presents it in a user-friendly format.
+
+_crate_`@`_version_: Fetch from a registry with a version constraint of "_version_"
+
+If the specified package is a package in the current workspace, it displays the information in the local Cargo.toml.
+
+If no version is specified, it will select the appropriate version based on the Minimum Supported Rust Version (MSRV)
+The following hierarchy is used to determine the MSRV:
```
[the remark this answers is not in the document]
hi-rustin: I agree that I put too many details here. I just deleted it.

src/etc/cargo.bashcomp.sh
```diff
@@ -63,6 +63,7 @@ _cargo()
 	local opt__fix="$opt_common $opt_pkg_spec $opt_feat $opt_mani $opt_parallel $opt_targets $opt_lock --release --target --message-format --broken-code --edition --edition-idioms --allow-no-vcs --allow-dirty --allow-staged --profile --target-dir --ignore-rust-version"
 	local opt__generate_lockfile="$opt_common $opt_mani $opt_lock"
 	local opt__help="$opt_help"
+	local opt_info="$opt_common $opt_lock --registry --index"
```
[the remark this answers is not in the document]
weihanglo: And step to test shell completions https://github.com/rust-lang/cargo/pull/11204#issuecomment-1272908297

src/doc/man/cargo-info.md
```diff
@@ -0,0 +1,63 @@
...
+cargo-info --- Display information about a package in the registry. Default registry is crates.io
+
+## SYNOPSIS
+
+`cargo info` [_options_] _crate_@_version_
+
+## DESCRIPTION
+
+This command displays information about a package in the registry. It fetches data from the package's Cargo.toml file
+and presents it in a user-friendly format.
```
weihanglo, line 14: Should we highlight that this is not machine-readable nor programmable format?

src/etc/cargo.bashcomp.sh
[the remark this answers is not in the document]
hi-rustin: Good catch! Thanks for the testing script!

src/doc/man/cargo-info.md
hi-rustin: Sounds good to me. Let me rephrase it.

hi-rustin: Updated.

epage (approved):

weihanglo (commented):
You might also want to add cargo-info to `src/doc/man/cargo.md`. See <https://github.com/rust-lang/cargo/pull/14339>.

hi-rustin:
> You might also want to add cargo-info to `src/doc/man/cargo.md`. See #14339.

Added. Thank you!

hi-rustin:
@epage @weihanglo I guess this PR is ready to merge. Could you please help merge it when you have time? If there is anything I need to update, please let me know.

@epage Thank you for guiding me in this work, I really appreciate it. Thank you for all your time.

src/cargo/ops/registry/info/mod.rs
```diff
@@ -0,0 +1,285 @@
+use anyhow::bail;
```
weihanglo, line 1: Since it isn't obvious. Could we put a module level doc comment here to describe that this is a module for `cargo info` command?

```suggestion
//! Implementation of `cargo info`.

use anyhow::bail;
```

weihanglo (commented):
I didn't look too closer to the implementation. That seems pretty solid. Thank you hi-rustin!

ehuss:
I tried giving this a spin, and was very surprised to find that it was requesting my token for crates.io. Is there a reason it must do that? That seems like it will be a bad user experience if someone needs to enter a password or use a biometric or hardware key in order to use this command.

Rustin170506:
> I tried giving this a spin, and was very surprised to find that it was requesting my token for crates.io. Is there a reason it must do that? That seems like it will be a bad user experience if someone needs to enter a password or use a biometric or hardware key in order to use this command.

I will take a look. It shouldn't require it. I believe I fixed it in my previous local test.

Rustin170506:
I tested it locally:
```console
❯ ./target/debug/cargo logout
      Logout token for `crates-io` has been removed from local storage
note: This does not revoke the token on the registry server.
    If you need to revoke the token, visit <https://crates.io/me> and follow the instructions there.

cargo on  rustin-patch-info is 📦 v0.83.0 via 🐍 v3.12.4 via 🦀 v1.82.0-nightly 
❯ ./target/debug/cargo info serde  
serde #serde #serialization #no_std
A generic serialization/deserialization framework
version: 1.0.204
license: MIT OR Apache-2.0
rust-version: 1.31
documentation: https://docs.rs/serde
homepage: https://serde.rs
repository: https://github.com/serde-rs/serde
crates.io: https://crates.io/crates/serde/1.0.204
features:
 +default      = [std]
  std          = []
  alloc        = []
  derive       = [serde_derive]
  rc           = []
  serde_derive = [dep:serde_derive]
  unstable     = []
note: to see how you depend on serde, run `cargo tree --invert --package serde@1.0.204`
```

It seems it doesn't require it.

src/cargo/ops/registry/info/mod.rs
Rustin170506: Updated.

src/bin/cargo/commands/info.rs
```diff
@@ -0,0 +1,39 @@
...
+        ))
+}
+
+pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
+    let package = args
+        .get_one::<String>("package")
+        .map(String::as_str)
+        .unwrap();
+    let spec = PackageIdSpec::parse(package).map_err(|e| {
+        anyhow::format_err!(
```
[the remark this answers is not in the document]
Rustin170506: Updated. Thank you!

src/cargo/ops/registry/mod.rs
```diff
@@ -206,6 +208,47 @@ fn get_source_id(
     }
 }
 
+// Very similar to get_source_id, but this function is used when the package_id is known.
```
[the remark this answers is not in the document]
Rustin170506: Updated.

src/doc/man/cargo-info.md
```diff
@@ -0,0 +1,63 @@
+# cargo-info(1)
+
+## NAME
+
+cargo-info --- Display information about a package in the registry. Default registry is crates.io
+
+## SYNOPSIS
+
+`cargo info` [_options_] _crate_@_version_
```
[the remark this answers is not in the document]
Rustin170506: I personally think the current way is easier to understand.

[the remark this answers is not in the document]
Rustin170506: But I think it makes sense to mention it is a package id.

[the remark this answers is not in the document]
Rustin170506: Added.

src/doc/man/cargo-info.md
```diff
@@ -0,0 +1,63 @@
...
+
+## DESCRIPTION
+
+This command displays information about a package in the registry. It fetches data from the package's Cargo.toml file
+and presents it in a human-readable format.
+
+_crate_`@`_version_: Fetch from a registry with a version constraint of "_version_"
+
+If the specified package is part of the current workspace, the information from the local Cargo.toml file will be
```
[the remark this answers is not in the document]
Rustin170506: Added.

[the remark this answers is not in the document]
Rustin170506: > Also, what does it mean by being part of the current workspace, being a workspace member or tracked in `Cargo.lock`?

being a workspace member.

For dependencies, we don't have their information in local Cargo.toml.

src/cargo/ops/registry/info/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+        .into_iter()
+        .flat_map(|p| resolve.deps(p))
+        .map(|(p, _)| p)
+        .filter(|&p| spec.matches(p))
+        .max_by_key(|&p| p.version())
+    {
+        return (Some(package_id), false);
+    }
+
+    if let Some(package_id) = ws
```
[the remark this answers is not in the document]
Rustin170506: I'm not sure if we should mention this detail here. From the start, I provided all the details about how we select a package, but Ed thought it was too detailed. He believes it might create more work if we want to change these implementation details.

src/doc/man/cargo-info.md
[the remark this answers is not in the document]
Rustin170506: You can specify a specific version(a different version) or you need to check it out of the workspace. I guess it's okay to not mention it. I believe checking it out of the workspace is quite straightforward.

Rustin170506 (commented):

weihanglo (approved):

epage:
I also logged out and couldn't reproduce.

As we develop reproduction steps, we can iterate from there.

@bors r+

ehuss:
Opened https://github.com/rust-lang/cargo/issues/14409 regarding the authentication problem.

PR #14144, opened 2024-06-25
Docs: Fix curly quotes in config docs.

ehuss:
mdBook automatically converts the "quotation mark" character to use smart quotes such as “example”. This isn't valid TOML syntax, and means you can't copy and paste these values. This wraps these TOML strings as code spans so that this conversion doesn't happen (and helps emphasize these are literal code spans).


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #14149, opened 2024-06-25
test: migrate registry and registry_auth to snapbox

eth3lbert:

Migrate following to snapbox:

- `tests/testsuite/registry.rs`
- `tests/testsuite/registry_auth.rs`

### Additional information

Parameterize shared logic over inline snapshots to prevent the content from being appended multiple times in `str![]` when it's called multiple times within a reusable function.

Big thanks to @weihanglo for helping me out!

### DIFF
--- a/tests/testsuite/registry.rs
+++ b/tests/testsuite/registry.rs
@@ -1,16 +1,16 @@
 //! Tests for normal registry dependencies.
 
-#![allow(deprecated)]
-
 use cargo::core::SourceId;
 use cargo_test_support::cargo_process;
 use cargo_test_support::paths::{self, CargoPathExt};
+use cargo_test_support::prelude::*;
 use cargo_test_support::registry::{
     self, registry_path, Dependency, Package, RegistryBuilder, Response, TestRegistry,
 };
-use cargo_test_support::{basic_manifest, project};
+use cargo_test_support::{basic_manifest, project, str};
 use cargo_test_support::{git, install::cargo_home, t};
 use cargo_util::paths::remove_dir_all;
+use snapbox::data::Inline;
 use std::fmt::Write;
 use std::fs::{self, File};
 use std::path::Path;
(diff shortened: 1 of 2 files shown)


tests/testsuite/registry_auth.rs
```diff
@@ -85,7 +84,7 @@ fn simple() {
         .build();
 
     let p = make_project();
-    cargo(&p, "build").with_stderr(SUCCESS_OUTPUT).run();
+    cargo(&p, "build").with_stderr_data(SUCCESS_OUTPUT).run();
```
weihanglo, line 87: Is it possible to make `SUCCESS_OUTPUT` snapshot aware?

tests/testsuite/registry.rs
```diff
@@ -30,15 +29,15 @@ fn test_server_stops() {
...
 }
 
 #[cargo_test]
 fn simple_git() {
-    simple();
+    simple(false);
 }
 
-fn simple() {
+fn simple(is_http: bool) {
```
weihanglo, line 40: I feel it a bit confusing. By just looking at code, I don't know why this `is_http` is needed. Apparently there is not different between each branch. Is there a way to document it, or this should be fixed in snapbox or go the other way round?

weihanglo (commented):
Apart from those two comments, LGTM

tests/testsuite/registry_auth.rs
weihanglo: I feel that this comment contradict to <https://github.com/rust-lang/cargo/pull/14149#discussion_r1653730915>.

tests/testsuite/registry.rs
eth3lbert: Currently, if a function contains `str![]` and is called twice, the content will become something like `str![[...] [...]]`. 

This PR introduces an argument `is_http` to handle `str![]` separately, allowing it to be filled with the correct content (even though the current testsuite doesn't show the difference).

If appending multiple times is not the expected behavior, I think this should be fixed in snapbox.

Alternatively, it's possible to eliminate the need for the `is_http` argument by avoiding `str![]` altogether.

tests/testsuite/registry_auth.rs
eth3lbert: > Is it possible to make `SUCCESS_OUTPUT` snapshot aware?

Is this referring to using its own `str![]` instead of the same `SUCCESS_OUTPUT`?

tests/testsuite/registry.rs
eth3lbert: I think `tests/testsuite/list_availables.rs` is also worth mentioning here. It reuses a function for assertions with distinct commands multiple times.

weihanglo: Thanks for the reply!

I still found it cumbersome to understand. Maybe we should parameterize on inline snapshot itself. That may look like

```rust
fn simple_git() {
    simple(str![[r#"
[DOWNLOADING] crates ...
[DOWNLOADED] bar v0.0.1 (registry `dummy-registry`)
[CHECKING] bar v0.0.1

"#]], str[[r#"
[CHECKING] bar v0.0.1

"#]]);
}
```


(Note that we have two shared assertions hence two arguments)

What do you think?

eth3lbert: Yeah, this also make sense to me! Thanks!

tests/testsuite/registry_auth.rs
eth3lbert: Got it, thanks!

eth3lbert:
All feedback addressed!

weihanglo (approved):

weihanglo:
Filed <https://github.com/assert-rs/snapbox/issues/345> though I don't think this is blocked on it.

I appreciate the effort you put on this back-and-forth snapshot changes. Thanks a lot!

@bors r+

PR #14150, opened 2024-06-25
docs(contrib): carefully considered when introducing errors

### LINKED ISSUE #14147: `build.rs` `cargo:rustc-check-cfg` spurious error with MSRV 1.56
### Problem

The check for the `cargo::` (double colon) syntax should apply only to directives that the MSRV would understand.

### Steps

`Cargo.toml`:

```
[package]
name = "derive-deftly-macros"
version = "0.13.0"
edition = "2021"
readme = "README.md"
authors=["Ian Jackson <ijackson@chiark.greenend.org.uk>",
         "and the contributors to Rust derive-deftly"]
license = "MIT"
description="Macros that implement the derive_deftly crate"
homepage = "https://gitlab.torproject.org/Diziet/rust-derive-deftly"
repository = "https://gitlab.torproject.org/Diziet/rust-derive-deftly"
rust-version = "1.56"

# After editing this file, you will probably need to run
#   maint/update-bizarre
# to update the "bizarre" testing versions in tests/pub-export/bizarre-*

[features]
case = ["heck"]
expect = ["sha3", "syn/full"]
meta-as-expr = ["syn/full"]
meta-as-items = ["syn/full"]

[dependencies]
indexmap = ">=1.8, <3"
itertools = ">=0.10.1, <0.14"
proc-macro-crate = ">=1.1.3, <4"
proc-macro2 = "1.0.53"
quote = "1"
sha3 = { version = "0.10", optional = true }
strum = { version = ">=0.24, <0.27", features = ["derive"] }
syn = { version = "2.0.53", features = ["extra-traits"] }
void = "1"

heck = { version = ">=0.4, <0.6", optional = true }

[lib]
path = "macros.rs"
proc-macro = true
```

`build.rs`:

```
fn main() {
    println!(
        r#"cargo::rustc-check-cfg=cfg(derive_deftly_dprint)
cargo::rustc-check-cfg=cfg(feature, values("bizarre"))"#
    );
}
```

Expected output: it compiles.

Actual output:

```
error: the `cargo::` syntax for build script output instructions was added in Rust 1.77.0, but the minimum supported Rust version of `derive-deftly-macros v0.13.0 (/volatile/rustcargo/Rustup/Derive-Deftly/derive-deftly/macros)` is 1.56.
Switch to the old `cargo:rustc-check-cfg=cfg(derive_deftly_dprint)` syntax (note the single colon).
See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
```

### Possible Solution(s)

Accept `cargo::` syntax unless the directive that follows `::` is one that a compiler which only supports `:` would understand.

### Notes

Unknown diretives are ignored, so `cargo::rustc-check-cfg` is completely fine.  I htink no version of rust that would understand `rustc-check-cfg` at all, doesn't understand `::`.

### Version

```text
cargo 1.80.0-nightly (34a6a87d8 2024-06-04)
release: 1.80.0-nightly
commit-hash: 34a6a87d8a2330d8c9d578f927489689328a652d
commit-date: 2024-06-04
os: Debian 10.0.0 (buster) [64-bit]
```

weihanglo:

Resolves #14147

Document that we should always consider warnings unless fixing critical bugs.

I am unsure if it is a principle we'd like to hold.
Just want to document a thing we need to pay attention to.

### DIFF
--- a/src/doc/contrib/src/design.md
+++ b/src/doc/contrib/src/design.md
@@ -77,6 +77,13 @@ data in Cargo's home directory) should not *prevent* older versions from
 running, but they may cause older versions to recreate the cache, which may
 result in a performance impact.
 
+Introducing new hard errors should be carefully considered, given their
+potential impact on compatibility. New errors might cause newer versions of
+Cargo to fail builds that older versions would have succeeded on, contradicting
+the expectation that packages compiling with older versions should also compile
+with newer ones unless critical bugs are addressed. Therefore, such changes
+should be thoroughly evaluated to minimize disruption.
+
 Changes to the registry index should not prevent older versions from working.
 Generally, older versions ignore new fields, so the format should be easily
 extensible. Changes to the format or interpretation of existing fields should


src/doc/contrib/src/design.md
```diff
@@ -77,6 +77,13 @@ data in Cargo's home directory) should not *prevent* older versions from
 running, but they may cause older versions to recreate the cache, which may
 result in a performance impact.
 
+Changes to emitting hard errors, unless for fixing critical bugs, should
```
epage, line 80: Following the existing wording practice in this section makes this hard to read.  This makes it sound like we are editing existing hard errors, like there message.

src/doc/contrib/src/design.md
```diff
@@ -77,6 +77,13 @@ data in Cargo's home directory) should not *prevent* older versions from
...
+consider using warnings instead. Introducing new hard errors may cause newer
+versions of Cargo to fail builds that older versions would have succeeded on,
+as the older versions lack the new behavior that emits those hard errors.
+This breaks the expectation that if a package compiles with an older version of
+Cargo, it should also compile with a newer version.
```
epage, line 85: There are perfectly good reasons to add new hard errors.  I feel like this isn't doing a good job explaining when to do so or not but making it sound like there is a blanket reason we should never do them.

weihanglo: It's hard for me to think of a set of criteria. Just revised the wording a bit, making it less like we don't accept errors.

epage: I can't quite place the issue I have with the new wording.  I wonder if this is worth covering.  We have a good treatment on this problem when it comes to [unstable features](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/features/index.html).  The case we ran into here was an odd mixture of things and a case where we likely wouldn't have done something different anyways.

ijackson: > The case we ran into here was an odd mixture of things and a case where we likely wouldn't have done something different anyways.

Maybe I have misunderstood.

From my point of view what happened here was a breach of Rust's stability guarantees.  It was a foreseeable consequence of adding a new warning that would need people to add suppressions.  Those suppressions would then, inevitably, need to be tolerated without complaint by older Rust versions.  Note that the message that needs suppressing even advises using the syntax which certain range of versions of Rust would reject., rather than the one which is tolerated by all Rust versions. 

It doesn't appear that there was anything that I could sensibly have done differently (other than perhaps to take a completely different design approach in my package) to detect and avoid this.

I *do* have very thorough tests, but it didn't occur to me that I might need to test intermediate versions of Rust.  The Rust stability guarantee ought to mean that testing with my MSRV (1.56 in this case) would be enough.

Now, I understand that mistakes happen.  Thanks to @zydou for trying to help avoid future similar mistakes.  I really appreciate it.  But:

It sounds like @epage is saying that what happened here is fine, was not a mistake, and in a similar situation you'd do the same again.  I do hope I have misunderstood.

epage: > From my point of view what happened here was a breach of Rust's stability guarantees. 

The "standard" stabilization process for Cargo has been to take a `Cargo.toml` field that would be considered "unused", have a period of time where using it is a hard error, and then after that it is processed according to a specific set of rules or it hard errors.  We followed that model here though we didn't have the "unused" warnings.

Like I alluded to earlier, we have recently formalized alternative models to avoid the hard errors that force an MSRV bump but they only apply in specific circumstances.  We currently have no process to help people to catch this broken phase except for where we have "unused" warnings and people notice them (e.g. I never look at the non-fatal output from my MSRV runs).

The only violations of stability guarantees were that we took `:` prefix link metadata and turned them into hard errors and you can no longer set the `check-cfg` link metadata key.  The former we did very consciously as a team.  The latter was from before we had fully considered and decided to avoid conflicting with link metadata keys and was grandfathered in and to avoid this is why we did the `::` change.

>  It was a foreseeable consequence 

That is a view from hindsight and is unfair to the developers involved.  Please be empathetic to developers who are working with the limited perspective they have during development.

Check-cfg had been in development for a long time but had stagnated.  During that lull, we had problem with evolving build script syntax and found a unlikely corner case that we could break compatibility on (link metadata that starts with `:;`) that we used as an out to make it so we could add new build script directives (technically, adding check-cfg, broke compatiblity as it took a potential link metadata name).  We did blanket documentation updates.  Check-cfg was then polished up and stabilized by someone seeing and acting on this new documented syntax.  We considered MSRV through this process but had overlooked `:` vs `::` in the documentation / messages.  Some of that get polished up as this rode the train to being on stable.

> It sounds like @epage is saying that what happened here is fine, was not a mistake, and in a similar situation you'd do the same again. I do hope I have misunderstood.

I don't think this proposed documentation change would have made a difference if it had been in place.  We already have been taking MSRV into account in our designs recently and had discussions during the stabilization of both halves of this related to MSRV.  Its a hard trade off to decide whether unsupported build script directives should be an error or a warning but starting with an error is the safest route.

This specific situation was partially from two features being developed in parallel without a good opportunity to reflect on how they interacted which is hard to do outside of hindsight and this documentation change does not address.

weihanglo:
Since we have documented the design process for [unstable features](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/features/index.html), and for new build script instructions Cargo reserves the `cargo::metadata` namespace, so the same situation is unlikely to happen again in build scripts. I myself cannot think of any other situation at this moment. Based on these and [the thread above](https://github.com/rust-lang/cargo/pull/14150#discussion_r1658020233), I am going to close this.

PR #14158, opened 2024-06-27
test: migrate some files to snapbox

eth3lbert:

Migrate following to snapbox:

- `tests/testsuite/local_registry.rs`
- `tests/testsuite/locate_project.rs`
- `tests/testsuite/lockfile_compat.rs`
- `tests/testsuite/login.rs`
- `tests/testsuite/logout.rs`

### DIFF
--- a/tests/testsuite/local_registry.rs
+++ b/tests/testsuite/local_registry.rs
@@ -1,10 +1,9 @@
 //! Tests for local-registry sources.
 
-#![allow(deprecated)]
-
 use cargo_test_support::paths::{self, CargoPathExt};
+use cargo_test_support::prelude::*;
 use cargo_test_support::registry::{registry_path, Package};
-use cargo_test_support::{basic_manifest, project, t};
+use cargo_test_support::{basic_manifest, project, str, t};
 use std::fs;
 
 fn setup() {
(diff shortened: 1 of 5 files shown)


tests/testsuite/local_registry.rs
```diff
@@ -241,17 +240,16 @@ fn multiple_names() {
...
-[CHECKING] [..]
-[CHECKING] foo v0.0.1 ([CWD])
-[FINISHED] [..]
-",
-        )
+[UNPACKING] bar v0.0.1 (registry `[ROOT]/registry`)
+[UNPACKING] baz v0.1.0 (registry `[ROOT]/registry`)
+[CHECKING] ba[..] v0.[..]
+[CHECKING] ba[..] v0.[..]
+[CHECKING] foo v0.0.1 ([ROOT]/foo)
```
epage, line 249: How come we are doing these globs?  If its unordered should we instead set that property?

eth3lbert: Yes, to address the ordering issue, I'll use `.unordered()` instead!

eth3lbert: Updated!

PR #14160, opened 2024-06-27
gix: remove `revision` feature from cargo

klensy:
Feature was added in #12731, but cargo builds without it and looks like used in tests only, so unset it.


ehuss:
Thanks!

cc @Byron FYI in case this has meaning.

@bors r+

Byron:
A great catch, thanks!

PR #14161, opened 2024-06-27
test: Auto-redact `... after last build at ...`; Migrate `freshness` to Snapbox

choznerol:

Part of #14039.

Migrate `tests/testsuite/freshness.rs` to snapbox.

I followed #14039. The `#![allow(deprecated)]` was removed and `tests/testsuite/freshness.rs` is still passing.

### Additional information

The redaction for "dirty reason" was a bit tricky

https://github.com/rust-lang/cargo/blob/7dcf764cbcff23b54fd061473f5ce66223cc0f86/src/cargo/core/compiler/fingerprint/dirty_reason.rs#L131

Current implementation would redact
- `1719325877.527949100s, 61549498ns after last build at 1719325877.466399602s`
- `1719503592.218193216s, 1h 1s after last build at 1719499991.982681034s`

into `[DIRTY_REASON_NEW_TIME], [DIRTY_REASON_DIFF] after last build at [DIRTY_REASON_OLD_TIME]`

Please let me know if that seems right.

### DIFF
--- a/crates/cargo-test-support/src/compare.rs
+++ b/crates/cargo-test-support/src/compare.rs
@@ -200,6 +200,16 @@ fn add_common_redactions(subs: &mut snapbox::Redactions) {
         regex!(r"ns/iter \(\+/- (?<redacted>[0-9]+(\.[0-9]+)?)\)"),
     )
     .unwrap();
+
+    // Following 3 subs redact:
+    //   "1719325877.527949100s, 61549498ns after last build at 1719325877.466399602s"
+    //   "1719503592.218193216s, 1h 1s after last build at 1719499991.982681034s"
+    // into "[DIRTY_REASON_NEW_TIME], [DIRTY_REASON_DIFF] after last build at [DIRTY_REASON_OLD_TIME]"
+    subs.insert(
+        "[TIME_DIFF_AFTER_LAST_BUILD]",
+        regex!(r"(?<redacted>[0-9]+(\.[0-9]+)?s, (\s?[0-9]+(\.[0-9]+)?(s|ns|h))+ after last build at [0-9]+(\.[0-9]+)?s)"),
+       )
+       .unwrap();
 }
 
 static MIN_LITERAL_REDACTIONS: &[(&str, &str)] = &[
--- a/tests/testsuite/freshness.rs
+++ b/tests/testsuite/freshness.rs
@@ -1,8 +1,7 @@
 //! Tests for fingerprinting (rebuild detection).
 
-#![allow(deprecated)]
-
 use filetime::FileTime;
+use snapbox::IntoData;
 use std::fs::{self, OpenOptions};
 use std::io;
 use std::io::prelude::*;
@@ -17,7 +16,7 @@ use cargo_test_support::paths::{self, CargoPathExt};
 use cargo_test_support::registry::Package;
 use cargo_test_support::{
     basic_lib_manifest, basic_manifest, is_coarse_mtime, project, rustc_host, rustc_host_env,
-    sleep_ms,
+    sleep_ms, str,
 };
 
 #[cargo_test]
@@ -28,34 +27,43 @@ fn modifying_and_moving() {
         .build();
 
     p.cargo("build")
-        .with_stderr(
-            "\
-[COMPILING] foo v0.0.1 ([CWD])
-[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [..]
-",
-        )
+        .with_stderr_data(str![[r#"
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
         .run();
 
-    p.cargo("build").with_stderr("[FINISHED] [..]").run();
+    p.cargo("build")
+        .with_stderr_data(str![[r#"
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
+        .run();
     p.root().move_into_the_past();
     p.root().join("target").move_into_the_past();
 
     p.change_file("src/a.rs", "#[allow(unused)]fn main() {}");
     p.cargo("build -v")
-        .with_stderr(
-            "\
-[DIRTY] foo v0.0.1 ([CWD]): the file `src/a.rs` has changed ([..])
-[COMPILING] foo v0.0.1 ([CWD])
+        .with_stderr_data(str![[r#"
+[DIRTY] foo v0.0.1 ([ROOT]/foo): the file `src/a.rs` has changed ([TIME_DIFF_AFTER_LAST_BUILD])
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
 [RUNNING] `rustc --crate-name foo [..]
-[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [..]
-",
-        )
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
         .run();
 
     fs::rename(&p.root().join("src/a.rs"), &p.root().join("src/b.rs")).unwrap();
     p.cargo("build")
         .with_status(101)
-        .with_stderr_contains("[..]file not found[..]")
+        .with_stderr_data(str![[r#"
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+error[E0583]: file not found for module `a`
+...
+[ERROR] could not compile `foo` (bin "foo") due to 1 previous error
+
+"#]])
         .run();
 }
 
@@ -70,12 +78,11 @@ fn modify_only_some_files() {
         .build();
 
     p.cargo("build")
-        .with_stderr(
-            "\
-[COMPILING] foo v0.0.1 ([CWD])
-[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [..]
-",
-        )
+        .with_stderr_data(str![[r#"
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
         .run();
     p.cargo("test").run();
     sleep_ms(1000);
@@ -89,14 +96,13 @@ fn modify_only_some_files() {
 
     // Make sure the binary is rebuilt, not the lib
     p.cargo("build -v")
-        .with_stderr(
-            "\
-[DIRTY] foo v0.0.1 ([CWD]): the file `src/b.rs` has changed ([..])
-[COMPILING] foo v0.0.1 ([CWD])
+        .with_stderr_data(str![[r#"
+[DIRTY] foo v0.0.1 ([ROOT]/foo): the file `src/b.rs` has changed ([TIME_DIFF_AFTER_LAST_BUILD])
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
 [RUNNING] `rustc --crate-name foo [..]
-[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [..]
-",
-        )
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
         .run();
     assert!(p.bin("foo").is_file());
 }
@@ -138,15 +144,14 @@ fn rebuild_sub_package_then_while_package() {
         .build();
 
     p.cargo("build")
-        .with_stderr(
-            "\
+        .with_stderr_data(str![[r#"
 [LOCKING] 3 packages to latest compatible versions
-[COMPILING] b [..]
-[COMPILING] a [..]
-[COMPILING] foo [..]
-[FINISHED] `dev` profile [..]
-",
-        )
+[COMPILING] b v0.0.1 ([ROOT]/foo/b)
+[COMPILING] a v0.0.1 ([ROOT]/foo/a)
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
         .run();
 
     if is_coarse_mtime() {
@@ -155,14 +160,13 @@ fn rebuild_sub_package_then_while_package() {
     p.change_file("b/src/lib.rs", "pub fn b() {}");
 
     p.cargo("build -pb -v")
-        .with_stderr(
-            "\
-[DIRTY] b v0.0.1 ([..]): the file `b/src/lib.rs` has changed ([..])
-[COMPILING] b [..]
+        .with_stderr_data(str![[r#"
+[DIRTY] b v0.0.1 ([ROOT]/foo/b): the file `b/src/lib.rs` has changed ([TIME_DIFF_AFTER_LAST_BUILD])
+[COMPILING] b v0.0.1 ([ROOT]/foo/b)
 [RUNNING] `rustc --crate-name b [..]
-[FINISHED] `dev` profile [..]
-",
-        )
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
         .run();
 
     p.change_file(
@@ -171,18 +175,17 @@ fn rebuild_sub_package_then_while_package() {
     );
 
     p.cargo("build -v")
-        .with_stderr(
-            "\
-[FRESH] b [..]
-[DIRTY] a [..]: the dependency b was rebuilt ([..])
-[COMPILING] a [..]
+        .with_stderr_data(str![[r#"
+[FRESH] b v0.0.1 ([ROOT]/foo/b)
+[DIRTY] a v0.0.1 ([ROOT]/foo/a): the dependency b was rebuilt ([TIME_DIFF_AFTER_LAST_BUILD])
+[COMPILING] a v0.0.1 ([ROOT]/foo/a)
 [RUNNING] `rustc --crate-name a [..]
-[DIRTY] foo [..]: the dependency b was rebuilt ([..])
-[COMPILING] foo [..]
-[RUNNING] `rustc --crate-name foo [..]
-[FINISHED] `dev` profile [..]
-",
-        )
+[DIRTY] foo v0.0.1 ([ROOT]/foo): the dependency b was rebuilt ([TIME_DIFF_AFTER_LAST_BUILD])
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+[RUNNING] `rustc --crate-name foo [..] src/lib.rs [..]
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
         .run();
 }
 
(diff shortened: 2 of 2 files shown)


crates/cargo-test-support/src/compare.rs
```diff
@@ -200,6 +200,27 @@ fn add_common_redactions(subs: &mut snapbox::Redactions) {
...
     .unwrap();
+
+    // Following 3 subs redact:
+    //   "1719325877.527949100s, 61549498ns after last build at 1719325877.466399602s"
+    //   "1719503592.218193216s, 1h 1s after last build at 1719499991.982681034s"
+    // into "[DIRTY_REASON_NEW_TIME], [DIRTY_REASON_DIFF] after last build at [DIRTY_REASON_OLD_TIME]"
+    // The ordering of these 3 subs matters.
+    subs.insert(
+        "[DIRTY_REASON_NEW_TIME]",
+        regex!(r"(?<redacted>[0-9]+(\.[0-9]+)?s), (\s?[0-9]+(\.[0-9]+)?(s|ns|h))+ after last build at ([0-9]+(\.[0-9]+)?s)"),
```
choznerol, line 211: Some example output I tested: https://regex101.com/r/fh0SEl/1

crates/cargo-test-support/src/compare.rs
```diff
@@ -200,6 +200,27 @@ fn add_common_redactions(subs: &mut snapbox::Redactions) {
         regex!(r"ns/iter \(\+/- (?<redacted>[0-9]+(\.[0-9]+)?)\)"),
     )
     .unwrap();
+
+    // Following 3 subs redact:
+    //   "1719325877.527949100s, 61549498ns after last build at 1719325877.466399602s"
+    //   "1719503592.218193216s, 1h 1s after last build at 1719499991.982681034s"
+    // into "[DIRTY_REASON_NEW_TIME], [DIRTY_REASON_DIFF] after last build at [DIRTY_REASON_OLD_TIME]"
+    // The ordering of these 3 subs matters.
```
epage, line 208: The order they are inserted?  They shouldn't.  We process redactions in a priority order, rather than insertion order.

tests/testsuite/freshness.rs
```diff
@@ -28,34 +27,39 @@ fn modifying_and_moving() {
...
 
     fs::rename(&p.root().join("src/a.rs"), &p.root().join("src/b.rs")).unwrap();
     p.cargo("build")
         .with_status(101)
-        .with_stderr_contains("[..]file not found[..]")
+        .with_stderr_data(str![[r#"
+...
+[..]file not found[..]
+...
+"#]])
```
epage, line 62: Does this still need to be a "contains" test with so much globbing?

tests/testsuite/freshness.rs
```diff
@@ -1841,12 +1940,24 @@ fn script_fails_stay_dirty() {
...
     // There was a bug where this second call would be "fresh".
     p.cargo("build")
-        .with_stderr_contains("[..]Crash![..]")
+        .with_stderr_data(str![[r#"
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+[ERROR] failed to run custom build command for `foo v0.0.1 ([ROOT]/foo)`
+...
+[..]Crash!
+...
+"#]])
```
epage, line 1960: Do these still need the globbing?

crates/cargo-test-support/src/compare.rs
```diff
@@ -200,6 +200,27 @@ fn add_common_redactions(subs: &mut snapbox::Redactions) {
         regex!(r"ns/iter \(\+/- (?<redacted>[0-9]+(\.[0-9]+)?)\)"),
     )
     .unwrap();
+
+    // Following 3 subs redact:
+    //   "1719325877.527949100s, 61549498ns after last build at 1719325877.466399602s"
+    //   "1719503592.218193216s, 1h 1s after last build at 1719499991.982681034s"
+    // into "[DIRTY_REASON_NEW_TIME], [DIRTY_REASON_DIFF] after last build at [DIRTY_REASON_OLD_TIME]"
```
epage, line 207: @weihanglo what are your thoughts on these redactions?

crates/cargo-test-support/src/compare.rs
choznerol: Thanks for pointing out. Indeed, `SNAPBOX=overwrites` still works after changing order. I think I just got this misunderstanding when tweaking regex.
Fixed in 281d123

tests/testsuite/freshness.rs
choznerol: Fixed in 34af338

tests/testsuite/freshness.rs
choznerol: Fixed in 34af338

tests/testsuite/freshness.rs
```diff
@@ -56,9 +56,19 @@ fn modifying_and_moving() {
...
+ --> src/main.rs:1:1
+  |
+1 | mod a; fn main() {}
+  | ^^^^^^
+  |
+  = [HELP] to create the module `a`, create file "src/a.rs" or "src/a/mod.rs"
+  = [NOTE] if there is a `mod a` elsewhere in the crate already, import it with `use crate::...` instead
+
+For more information about this error, try `rustc --explain E0583`.
+[ERROR] could not compile `foo` (bin "foo") due to 1 previous error
```
epage, line 70: Ok, so this was for a compiler error.

For these we replace the bulk of the compiler message (the lines I have highlighted) with `...`

tests/testsuite/freshness.rs
```diff
@@ -492,62 +508,84 @@ fn changing_bin_features_caches_targets() {
...
-        e.with_stderr(
-            "\
-[DIRTY] foo v0.0.1 ([..]): the list of features changed
-[COMPILING] foo[..]
-[RUNNING] `rustc [..]
-[FINISHED] `dev`[..]",
-        );
+        e.with_stderr_data(str![[r#"
+[DIRTY] foo v0.0.1 ([ROOT]/foo): the dependency foo was rebuilt ([DIRTY_REASON_NEW_TIME], [DIRTY_REASON_DIFF] after last build at [DIRTY_REASON_OLD_TIME])
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
```
epage: Any insight from how this happened?  Overall, we should be using `SNAPSHOTS=overwrite` so it seems like this shouldn't have happened so I'm wondering if you were doing some things manually (and if so, if there was a reason) or if there was a bug somewhere.

epage:
Could you squash your "fixup" commits into their appropriate commit?

tests/testsuite/freshness.rs
choznerol: I mostly use (or at least start from) `SNAPSHOTS=overwrite`, but for these `if cfg!(target_env = "msvc") {...}` blocks I had to do manual edit because I didn't have a Windows machine to run it locally 😅.

tests/testsuite/freshness.rs
choznerol: Sure, but I think the L70 is [from Cargo](https://github.com/choznerol/cargo/blob/issue-14039-snapshot-port-freshness/src/cargo/core/compiler/mod.rs#L421-L425). I replaced L61-L69 with `...` in latest commit.

tests/testsuite/freshness.rs
epage: Oh, yeah, that is a pain :)

tests/testsuite/freshness.rs
```diff
@@ -1524,18 +1612,20 @@ fn reuse_panic_pm() {
...
+[COMPILING] bar v0.0.1 ([ROOT]/foo/bar)
+[RUNNING] `rustc --crate-name bar [..] bar/src/lib.rs [..]--crate-type lib --emit=[..]link[..]
+[RUNNING] `rustc --crate-name bar [..] bar/src/lib.rs [..]--crate-type lib --emit=[..]link -C panic=abort[..]-C debuginfo=2 [..]
+[COMPILING] somepm v0.0.1 ([ROOT]/foo/somepm)
+[RUNNING] `rustc --crate-name somepm [..]`
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+[RUNNING] `rustc --crate-name foo [..]-C panic=abort[..]`
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]]
```
choznerol, line 1627: I'm trying to fix this `reuse_panic_pm` [CI failure](https://github.com/rust-lang/cargo/actions/runs/9724974056/job/26841764339#step:11:4052), which passed on `Tests on macOS aarch64 stable` but failed on `Test on macOS x86_64 stable` with below:

```diff

---- expected: tests/testsuite/freshness.rs:1616:13
++++ actual:   stderr
   1    1 | [LOCKING] 3 packages to latest compatible versions
   2    2 | [COMPILING] bar v0.0.1 ([ROOT]/foo/bar)
   3    3 | [RUNNING] `rustc --crate-name bar [..] bar/src/lib.rs [..]--crate-type lib --emit=[..]link[..]
-  4      - [RUNNING] `rustc --crate-name bar [..] bar/src/lib.rs [..]--crate-type lib --emit=[..]link -C panic=abort[..]-C debuginfo=2 [..]
   5    4 | [COMPILING] somepm v0.0.1 ([ROOT]/foo/somepm)
   6    5 | [RUNNING] `rustc --crate-name somepm [..]`
   7    6 | [COMPILING] foo v0.0.1 ([ROOT]/foo)
   8    7 | [RUNNING] `rustc --crate-name foo [..]-C panic=abort[..]`
   9    8 | [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+       9 + [RUNNING] `rustc --crate-name bar --edition=2015 bar/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no --check-cfg 'cfg(docsrs)' --check-cfg 'cfg(feature, values())' -C metadata=9229e8e86d572a0d -C extra-filename=-9229e8e86d572a0d --out-dir [ROOT]/foo/target/debug/deps -L dependency=[ROOT]/foo/target/debug/deps`
```

Some thought & findings so far:

Since the test logic was unchanged, I think the arch difference can be ruled out first.

Me next suspect is flakiness, which leads me to this `with_stderr_unordered` caveat from its nice comment:

>     /// Be careful when using patterns such as `[..]`, because you may end up
>     /// with multiple lines that might match, and this is not smart enough to
>     /// do anything like longest-match. For example, avoid something like:
>     ///
>     /// ```text
>     ///  [RUNNING] `rustc [..]
>     ///  [RUNNING] `rustc --crate-name foo [..]
>     /// ```
>     ///
>     /// This will randomly fail if the other crate name is `bar`, and the
>     /// order changes.

The original L1531-L1532 seems to already match this scenario (A line matching L1532 could also match L1531). However, I assume `reuse_panic_pm` has been stable until this something introduced in my PR.

According to the comment form test:
> bar is built once without panic (for proc-macro) and once with (for the normal dependency).

If the order of two `bar` build is non-deterministic this could be a problem; In contrast, if the order is deterministic, that makes me wonder why using `unordered` [in the first place](https://github.com/rust-lang/cargo/commit/86489946d0828af1617ba921d3cc16146a9e0e53#diff-1d79c710a0473fc4ffb021e2bd34a3cdd5a2431a0a9dc905baba7458090064b1R1166-R1176).

In fact, I removed `.unordered`, reran and passed `reuse_panic_pm` a few times locally, so let my try this first.

I also tried to reproduce any randomness introduce by [Snapbox `unordered`], but no finding:
<details>

```rust
use snapbox::prelude::*;
use snapbox::str;
use snapbox::assert_data_eq;

#[cargo_test]
fn test_snapbox_unorder_randomness() {
    let actual = str![[r#"
rustc --crate-name foo --a=b --ccc=ddd
rustc --crate-name foo --a=b --panic=abort --ccc=ddd
eee fff
"#]]
    .unordered();
    let expected = str![[r#"
rustc --crate-name foo [..]
rustc --crate-name foo [..] --panic=abort [..]
eee fff
"#]]
    .unordered();
    assert_data_eq!(actual, expected); // Always pass. Not flaky.
}
```

</details>

P.s. For reference, the full output from my local runs are like:
```
[COMPILING] bar v0.0.1 ([ROOT]/foo/bar)
[RUNNING] `rustc --crate-name bar --edition=2015 bar/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no --check-cfg 'cfg(docsrs)' --check-cfg 'cfg(feature, values())' -C metadata=3eb56625d2059feb -C extra-filename=-3eb56625d2059feb --out-dir [ROOT]/foo/target/debug/deps -L dependency=[ROOT]/foo/target/debug/deps`
[RUNNING] `rustc --crate-name bar --edition=2015 bar/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type lib --emit=dep-info,metadata,link -C panic=abort -C embed-bitcode=no -C debuginfo=2 -C split-debuginfo=unpacked --check-cfg 'cfg(docsrs)' --check-cfg 'cfg(feature, values())' -C metadata=e34f62847e3b5fb0 -C extra-filename=-e34f62847e3b5fb0 --out-dir [ROOT]/foo/target/debug/deps -L dependency=[ROOT]/foo/target/debug/deps`
[COMPILING] somepm v0.0.1 ([ROOT]/foo/somepm)
[RUNNING] `rustc --crate-name somepm --edition=2015 somepm/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type proc-macro --emit=dep-info,link -C prefer-dynamic -C embed-bitcode=no --check-cfg 'cfg(docsrs)' --check-cfg 'cfg(feature, values())' -C metadata=a7ef9925b27e01b6 -C extra-filename=-a7ef9925b27e01b6 --out-dir [ROOT]/foo/target/debug/deps -L dependency=[ROOT]/foo/target/debug/deps --extern bar=[ROOT]/foo/target/debug/deps/libbar-[HASH].rlib --extern proc_macro`
[COMPILING] foo v0.0.1 ([ROOT]/foo)
[RUNNING] `rustc --crate-name foo --edition=2015 src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type lib --emit=dep-info,metadata,link -C panic=abort -C embed-bitcode=no -C debuginfo=2 -C split-debuginfo=unpacked --check-cfg 'cfg(docsrs)' --check-cfg 'cfg(feature, values())' -C metadata=ebe9fb0a09bbfc5f -C extra-filename=-ebe9fb0a09bbfc5f --out-dir [ROOT]/foo/target/debug/deps -L dependency=[ROOT]/foo/target/debug/deps --extern bar=[ROOT]/foo/target/debug/deps/libbar-[HASH].rmeta --extern somepm=[ROOT]/foo/target/debug/deps/libsomepm-[HASH].dylib`
```

I guess some of my questions are:
- Just in case, was there any similar `reuse_panic_pm` flakiness you recall seeing prior to this PR?
- Does `reuse_panic_pm` really need `with_stderr_unordered` or `.unordered()`?
- Anything else I should look into?

[Snapbox `unordered`]: https://doc.rust-lang.org/nightly/nightly-rustc/cargo_test_support/prelude/trait.IntoData.html#method.unordered

tests/testsuite/freshness.rs
```diff
@@ -492,62 +510,86 @@ fn changing_bin_features_caches_targets() {
...
-[RUNNING] `rustc [..]
-[FINISHED] `dev`[..]",
-        );
+        e.with_stderr_data(str![[r#"
+[DIRTY] foo v0.0.1 ([ROOT]/foo): the list of features changed
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+[RUNNING] `rustc --crate-name [..]
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
```
choznerol, line 552: ok this one is even more surprising 🤯 

https://github.com/rust-lang/cargo/actions/runs/9729712870/job/26851897369#step:11:4057

```diff
 thread 'freshness::changing_bin_features_caches_targets' panicked at tests\testsuite\freshness.rs:561:7:

---- expected: tests\testsuite\freshness.rs:546:28
++++ actual:   stderr
-  1      - [DIRTY] foo v0.0.1 ([ROOT]/foo): the list of features changed
-  2      - [COMPILING] foo v0.0.1 ([ROOT]/foo)
-  3      - [RUNNING] `rustc --crate-name [..]
+       1 + [FRESH] foo v0.0.1 ([ROOT]/foo)
   4    2 | [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
```
How could MSVC suddenly get this behavior change, I don't think there is any logical change on `changing_bin_features_caches_targets`. 

It was still `[DIRTY]` on 6/28: https://github.com/rust-lang/cargo/actions/runs/9700016140/job/26770429920#step:11:4070

But since yesterday it became `[FRESH]`: https://github.com/rust-lang/cargo/actions/runs/9724808711/job/26841405762#step:11:4057

I did [rebase](https://github.com/rust-lang/cargo/pull/14161#event-13339691283) onto `main` when squashing yesterday, but there was no new change other than tests from `main` according to [diff](https://github.com/rust-lang/cargo/compare/cbf4b1bf5cd5fc45719be5ab4231ae8b77813ec9..56ca35f7a65bbed6f75b5cf973f729a2ab0be328).

tests/testsuite/freshness.rs
choznerol: Update: a453efd CI passes `reuse_panic_pm` on all stable pipeline: [linux](https://github.com/rust-lang/cargo/actions/runs/9729603770/job/26851692382#step:11:2249), [mac x86](https://github.com/rust-lang/cargo/actions/runs/9729603770/job/26851692839#step:11:2246), [mac aarch64](https://github.com/rust-lang/cargo/actions/runs/9729603770/job/26851693000#step:11:2245), [Windows](https://github.com/rust-lang/cargo/actions/runs/9729603770/job/26851693071#step:11:2270)

crates/cargo-test-support/src/compare.rs
weihanglo: Should we instead reduct everything in `()` that contains "after last build"?

choznerol: Thanks. That turns out to make it much simpler 😅 
Will squash 75f6f5f and 14cbe9e later.

tests/testsuite/freshness.rs
choznerol: allow me to push a few temporary commits to test this out ⌛

weihanglo: Did you know on which toolchain version that the behavior changes?

choznerol: > It was still `[DIRTY]` on 6/28: [Run 9700016140](https://github.com/rust-lang/cargo/actions/runs/9700016140/job/26770429920#step:11:4070)

Sorry I was wrong about this. `changing_bin_features_caches_targets` has been failing has been fail since beginning. It just failed in earlier assertion 🤦.

Reverting my change to `changing_bin_features_caches_targets` actually fix the test (c826531 and its [ci](https://github.com/rust-lang/cargo/actions/runs/9747539766/job/26900484318#step:11:2240)), so there must be something incorrect from my change. Let me check this later.

choznerol: So this was just from a silly typo. I accidentally added a duplicated `.run()` at L552. Removed in latest amended commit.

tests/testsuite/freshness.rs
choznerol: Update: The same error happened again after a453efd on `Tests Windows x86_64 gnu nightly`, so a453efd didn't fix the problem 😞 

https://github.com/rust-lang/cargo/actions/runs/9747539766/job/26900484318#step:11:4053

This seems like a flaky error. Dropping a453efd for now.

crates/cargo-test-support/src/compare.rs
weihanglo: I am fine with the current redaction. Would could remove/change it when needed.

tests/testsuite/freshness.rs
weihanglo: Might need a deeper investigation of it. Fine with holding off for now.

weihanglo (approved):
Let's not block this. Thank you!

PR #14172, opened 2024-06-30
test: Migrate git_auth to snapbox

dieterplex:
Part of #14039.

### DIFF
--- a/tests/testsuite/git_auth.rs
+++ b/tests/testsuite/git_auth.rs
@@ -1,7 +1,5 @@
 //! Tests for git authentication.
 
-#![allow(deprecated)]
-
 use std::collections::HashSet;
 use std::io::prelude::*;
 use std::io::BufReader;
@@ -12,7 +10,7 @@ use std::thread::{self, JoinHandle};
 
 use cargo_test_support::git::cargo_uses_gitoxide;
 use cargo_test_support::paths;
-use cargo_test_support::{basic_manifest, project};
+use cargo_test_support::{basic_manifest, project, str};
 
 fn setup_failed_auth_test() -> (SocketAddr, JoinHandle<()>, Arc<AtomicUsize>) {
     let server = TcpListener::bind("127.0.0.1:0").unwrap();
@@ -138,10 +136,10 @@ fn http_auth_offered() {
     // may span multiple lines, and isn't relevant to this test.
     p.cargo("check")
         .with_status(101)
-        .with_stderr_contains(&format!(
+        .with_stderr_data(&format!(
             "\
 [UPDATING] git repository `http://{addr}/foo/bar`
-[ERROR] failed to get `bar` as a dependency of package `foo v0.0.1 [..]`
+[ERROR] failed to get `bar` as a dependency of package `foo v0.0.1 ([ROOT]/foo)`
 
 Caused by:
   failed to load source for dependency `bar`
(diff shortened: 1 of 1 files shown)


tests/testsuite/git_auth.rs
```diff
@@ -366,7 +435,29 @@ Caused by:
...
         .with_status(101)
-        .with_stderr_contains("[..]Unable to update[..]")
+        .with_stderr_data(str![[r#"
+[UPDATING] git repository `ssh://needs-proxy.invalid/git`
+[RUNNING] `git fetch --verbose --force --update-head-ok [..]ssh://needs-proxy.invalid/git[..] [..]+HEAD:refs/remotes/origin/HEAD[..]`
+ssh: Could not resolve hostname needs-proxy.invalid: [..] not known
+fatal: Could not read from remote repository.
+
+Please make sure you have the correct access rights
+and the repository exists.
```
weihanglo, line 445: I was thinking we shouldn't over matching against Git and SSH output, but meh we've already done that so leave it.

tests/testsuite/git_auth.rs
```diff
@@ -209,25 +215,42 @@ fn https_something_happens() {
...
+  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli
+
+Caused by:
+  An IO error occurred when talking to the server
+
+Caused by:
+  [35] SSL connect error ([..])"
             } else if cfg!(windows) {
-                "[..]failed to send request: [..]"
+                "  failed to send request: A connection with the server could not be established\n  ; class=Os (2)"
```
weihanglo, line 246: as this is not a snapshot, could we keep the `[..]failed to send request:[..]` in case the platform-dependent error message changes in the future?

tests/testsuite/git_auth.rs
```diff
@@ -209,25 +215,42 @@ fn https_something_happens() {
...
             } else if cfg!(target_os = "macos") {
                 // macOS is difficult to tests as some builds may use Security.framework,
                 // while others may use OpenSSL. In that case, let's just not verify the error
                 // message here.
-                "[..]"
+                "..."
             } else {
-                "[..]SSL error: [..]"
+                r"  SSL [ERROR] syscall failure: ; class=Os (2)"
```
weihanglo, line 253: ditto for platform-dependent error message 
```suggestion
                r"[..]SSL [ERROR][..]"
```

tests/testsuite/git_auth.rs
```diff
@@ -277,31 +300,74 @@ fn ssh_something_happens() {
...
             // But since there is no common meaningful sequence or word, we can only match a small telling sequence of characters.
-            "[..]onnect[..]"
+            format!(
+                "  banner exchange: Connection to {} port {}: Software caused connection abort",
+                addr.ip(),
+                addr.port()
+            )
         } else {
-            "[..]Connection [..] by [..]"
+            "[..]Connection closed by [..]".to_string()
```
weihanglo, line 309: Could just revert the assertion string of `message` back? By reading the comment I am afraid we'll bump into some error for different targets (musl vs gnu vs MSVC I guess?).

dieterplex: Seems not just `message`, it is flaky sometimes with following pattern.

```
thread 'git_auth::ssh_something_happens' panicked at tests/testsuite/git_auth.rs:347:10:

--- Expected
++++ actual:   stderr
   1    1 | [UPDATING] git repository `ssh://127.0.0.1:45997/foo/bar`
        2 + kex_exchange_identification: read: Connection reset by peer
        3 + Connection reset by 127.0.0.1 port 45997
        4 + [WARNING] spurious network error (3 tries remaining): failed to fill whole buffer
   2    5 | [ERROR] failed to get `bar` as a dependency of package `foo v0.0.1 ([ROOT]/foo)`
   3    6 | 
   4    7 | Caused by:
   5    8 |   failed to load source for dependency `bar`
   6    9 | 
          ⋮
  17   20 | 
  18   21 | Caused by:
  19   22 |   An IO error occurred when talking to the server
  20   23 | 
  21   24 | Caused by:
  22      - [..]Connection [..] by [..]
       25 +   ssh: connect to host 127.0.0.1 port 45997: Connection refused
```
instead of current snapshot with last line `  Connection closed by {ip} port {port}`

Could this expected string be below one?
```
[UPDATING] git repository `ssh://127.0.0.1:45997/foo/bar`
...
{message}
...
```

tests/testsuite/git_auth.rs
dieterplex: https://github.com/rust-lang/cargo/actions/runs/9767225029/job/26962030918?pr=14172#step:11:4063
```
  13   13 | Caused by:
  14   14 | [..]failed to send request: [..]
       15 +   ; class=Os (2)
```

should it be `[..]failed to send request:[..]\n...`?

weihanglo: That hurts 😞.

Yes please.

tests/testsuite/git_auth.rs
weihanglo: > Seems not just message, it is flaky sometimes in local testing with following pattern.

That's the reason I fear for. We shouldn't assert outputs from external tools (ssh, git). 
I am fine with your suggestion. Go ahead.

weihanglo (approved):
git_auth assertion are quite hairy 😞. Anyway, thanks!

PR #14177, opened 2024-07-01
test: migrate rust_version and rustc* to snapbox

eth3lbert:

Migrate following to snapbox:

- `tests/testsuite/rust_version.rs`
- `tests/testsuite/rustc.rs`
- `tests/testsuite/rustc_info_cache.rs`

### DIFF
--- a/tests/testsuite/rustc.rs
+++ b/tests/testsuite/rustc.rs
@@ -1,12 +1,7 @@
 //! Tests for the `cargo rustc` command.
 
-#![allow(deprecated)]
-
-use cargo_test_support::{basic_bin_manifest, basic_lib_manifest, basic_manifest, project};
-
-const CARGO_RUSTC_ERROR: &str =
-    "[ERROR] extra arguments to `rustc` can only be passed to one target, consider filtering
-the package by passing, e.g., `--lib` or `--bin NAME` to specify a single target";
+use cargo_test_support::prelude::*;
+use cargo_test_support::{basic_bin_manifest, basic_lib_manifest, basic_manifest, project, str};
 
 #[cargo_test]
 fn build_lib_for_foo() {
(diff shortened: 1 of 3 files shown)


tests/testsuite/rustc.rs
```diff
@@ -462,17 +427,17 @@ fn build_only_bar_dependency() {
...
-[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [..]
-",
-        )
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
         .run();
 }
 
+#[allow(deprecated)]
```
eth3lbert, line 440: `with_stderr_does_not_contain` is being used in this function.

tests/testsuite/rustc.rs
```diff
@@ -622,6 +589,7 @@ fn rustc_with_other_profile() {
     p.cargo("rustc --profile test").run();
 }
 
+#[allow(deprecated)]
```
eth3lbert, line 592: `with_stderr_does_not_contain` is being used in this function.

tests/testsuite/rustc_info_cache.rs
```diff
@@ -10,6 +8,7 @@ const MISS: &str = "[..] rustc info cache miss[..]";
 const HIT: &str = "[..]rustc info cache hit[..]";
 const UPDATE: &str = "[..]updated rustc info cache[..]";
 
+#[allow(deprecated)]
```
eth3lbert, line 11: `with_stderr_does_not_contain` is being used in this function.

tests/testsuite/rustc_info_cache.rs
```diff
@@ -105,6 +104,7 @@ fn rustc_info_cache() {
         .run();
 }
 
+#[allow(deprecated)]
```
eth3lbert, line 107: `with_stderr_does_not_contain` is being used in this function.

tests/testsuite/rustc_info_cache.rs
eth3lbert, line 1: Based on the assertion that shows a pair of `contain` with `does_not_contain`, I'm inclined to keep it as is.

tests/testsuite/rustc.rs
```diff
@@ -736,21 +708,36 @@ fn rustc_with_print_cfg_multiple_targets() {
...
-        .with_stdout_contains("target_pointer_width=\"64\"")
-        .with_stdout_contains("target_vendor=\"pc\"")
-        .with_stdout_contains("windows")
-        .with_stdout_contains("target_env=\"gnu\"")
-        .with_stdout_contains("target_family=\"unix\"")
-        .with_stdout_contains("target_pointer_width=\"32\"")
-        .with_stdout_contains("target_vendor=\"unknown\"")
-        .with_stdout_contains("target_os=\"linux\"")
-        .with_stdout_contains("unix")
+        .with_stdout_data(str![[r#"
```
weihanglo, line 711: The order is not so importnat, should we use `.unordered()` and just one `...` instead?

tests/testsuite/rustc_info_cache.rs
weihanglo: How bad is it if we compare the full output? I think we may need to add a comment for each cargo invocation for the expected hit/miss/update output.

I do want to kill `does_not_contain` as it is even more unreliable.

tests/testsuite/rustc.rs
eth3lbert: Ok, this also works and make sense to me!

tests/testsuite/rustc_info_cache.rs
eth3lbert: Sounds good!  I'll give it a shot and see what results we can get.

eth3lbert: Oh, it seems like there are extra log messages related to `rustup`.

weihanglo: That is terribly bad. Let's keep the old tests.

eth3lbert: done reverting!

weihanglo (approved):

weihanglo:
Looks great. Thank you for the migration.

@bors r+

PR #14183, opened 2024-07-03
Check build target supports std when building with -Zbuild-std=std

harmou01:
**What does this PR try to resolve?**

Ensures that Cargo first verifies whether a given target supports building the standard library when the `-Zbuild-std=std` option is passed to Cargo ([see issue here](https://github.com/rust-lang/wg-cargo-std-aware/issues/87)). This information can be obtained by querying `rustc --print=target-spec-json`. The target spec "metadata" contains an `Option<bool>` determining whether the target supports building std.

In the case this value is `false`, cargo will stop the build. This avoids the numerous "use of unstable library" errors, giving a cleaner, and simpler, "building std is not supported on this target".

**How should we test and review this PR?**

It can be manually tested by running `cargo build --target <target> -Zbuild-std=std`. If a target who's target-spec marks std as false is passed, cargo will exit. This works with multiple `--target`'s, and if any of them don't support std, cargo will exit.

**Additional Information**

This change relies on two things:
* The target-spec metadata in rustc needs to be filled out. Currently, most fields are None, which is treated as OK with this change. Meaning this can be merged before the rustc changes.
* The new test case added with this change will fail without at least `aarch64-unknown-none` having it's target-spec metadata completed.
* Some targets have std support marked as "?". If this state is properly represented in the target-spec in the future, it needs to be determined whether this is allowed, so the build can continue, or whether it fails.

### DIFF
--- a/src/cargo/core/compiler/build_context/target_info.rs
+++ b/src/cargo/core/compiler/build_context/target_info.rs
@@ -43,6 +43,8 @@ pub struct TargetInfo {
     crate_types: RefCell<HashMap<CrateType, Option<(String, String)>>>,
     /// `cfg` information extracted from `rustc --print=cfg`.
     cfg: Vec<Cfg>,
+    /// `supports_std` information extracted from `rustc --print=target-spec-json`
+    pub supports_std: Option<bool>,
     /// Supported values for `-Csplit-debuginfo=` flag, queried from rustc
     support_split_debuginfo: Vec<String>,
     /// Path to the sysroot.
@@ -294,6 +296,44 @@ impl TargetInfo {
                 gctx.shell().warn("non-trivial mutual dependency between target-specific configuration and RUSTFLAGS")?;
             }
 
+            let mut supports_std: Option<bool> = None;
+
+            // The '--print=target-spec-json' is an unstable option of rustc, therefore only
+            // try to fetch this information if rustc allows nightly features. Additionally,
+            // to avoid making two rustc queries when not required, only try to fetch the
+            // target-spec when the '-Zbuild-std' option is passed.
+            if gctx.cli_unstable().build_std.is_some() {
+                let mut target_spec_process = rustc.workspace_process();
+                apply_env_config(gctx, &mut target_spec_process)?;
+                target_spec_process
+                    .arg("--print=target-spec-json")
+                    .arg("-Zunstable-options")
+                    .args(&rustflags)
+                    .env_remove("RUSTC_LOG");
+
+                if let CompileKind::Target(target) = kind {
+                    target_spec_process
+                        .arg("--target")
+                        .arg(target.rustc_target());
+                }
+
+                #[derive(Deserialize)]
+                struct Metadata {
+                    pub std: Option<bool>,
+                }
+
+                #[derive(Deserialize)]
+                struct TargetSpec {
+                    pub metadata: Metadata,
+                }
+
+                if let Ok(output) = target_spec_process.output() {
+                    if let Ok(spec) = serde_json::from_slice::<TargetSpec>(&output.stdout) {
+                        supports_std = spec.metadata.std;
+                    }
+                }
+            }
+
             return Ok(TargetInfo {
                 crate_type_process,
                 crate_types: RefCell::new(map),
@@ -310,6 +350,7 @@ impl TargetInfo {
                 )?
                 .into(),
                 cfg,
+                supports_std,
                 support_split_debuginfo,
             });
         }
@@ -1025,6 +1066,16 @@ impl<'gctx> RustcTargetData<'gctx> {
             CompileKind::Host => &self.host_config,
             CompileKind::Target(s) => &self.target_config[&s],
         }
+    }
+
+    pub fn get_unsupported_std_targets(&self) -> Vec<&str> {
+        let mut unsupported = Vec::new();
+        for (target, target_info) in &self.target_info {
+            if target_info.supports_std == Some(false) {
+                unsupported.push(target.short_name());
+            }
+        }
+        unsupported
     }
 }
 
--- a/src/cargo/core/compiler/standard_lib.rs
+++ b/src/cargo/core/compiler/standard_lib.rs
@@ -72,6 +72,17 @@ pub fn resolve_std<'gctx>(
             .warn("-Zbuild-std does not currently fully support --build-plan")?;
     }
 
+    // check that targets support building std
+    if crates.contains(&"std".to_string()) {
+        let unsupported_targets = target_data.get_unsupported_std_targets();
+        if !unsupported_targets.is_empty() {
+            anyhow::bail!(
+                "building std is not supported on the following targets: {}",
+                unsupported_targets.join(", ")
+            )
+        }
+    }
+
     let src_path = detect_sysroot_src_path(target_data)?;
     let std_ws_manifest_path = src_path.join("Cargo.toml");
     let gctx = ws.gctx();
--- a/tests/testsuite/standard_lib.rs
+++ b/tests/testsuite/standard_lib.rs
@@ -324,6 +324,32 @@ fn check_core() {
 }
 
 #[cargo_test(build_std_mock)]
+fn test_std_on_unsupported_target() {
+    let setup = setup();
+
+    let p = project()
+        .file(
+            "src/main.rs",
+            r#"
+        fn main() {
+            println!("hello");
+        }
+        "#,
+        )
+        .build();
+
+    p.cargo("build")
+        .arg("--target=aarch64-unknown-none")
+        .arg("--target=x86_64-unknown-none")
+        .build_std(&setup)
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] building std is not supported on the following targets: [..]
+"#]])
+        .run();
+}
+
+#[cargo_test(build_std_mock)]
 fn depend_same_as_std() {
     let setup = setup();
 


ehuss:
Thanks for the PR! Unfortunately I'm not sure this is something we can move forward with in the short term. The `--print target-spec-json` option would need to be stabilized in order for this to be supported in build-std. I'm not sure how feasible that is. My impression is that it seems like that will be difficult, since some of the fields are somewhat inherently unstable.

There is some more information at https://github.com/rust-lang/rust/issues/38338 and https://github.com/rust-lang/wg-cargo-std-aware/issues/6.

I would suggest getting in contact with the compiler team to suss out the viability of that. There might be other alternatives, such as providing other rustc options for querying/extracting information like this.

adamgemmell:
I didn't think that target-spec-json would need to be stable because Cargo and Rustc are versioned together - this creates a contract between the two, but one that can be changed. Would a test in Rust that tests this integration alleviate your concerns? We only need to parse part of the spec.

Otherwise I doubt the spec can be stabilised, so I think we'd need some other mechanism of getting target info from Rustc. I think this might be useful for things like getting the default panic or unwind strategy too.

adamgemmell:
I appreciate the detailed answer, thanks. Agreed on the RUSTC_BOOTSTRAP=1 hack. On checking the target-spec-json tracking issue it does seem like there's some support for stabilising the print option which surprised me - I think I had it mixed up in my head with the idea of providing the JSON as an input target.

Considering the case where it was stabilised with all fields, current and future, optional. In the case of this PR, it would mean that Cargo handles the "not found" case for the `Metadata::std` field, probably by printing out a warning, but not error out or break Cargo's CI. Then we'd basically have to wait for a bug report from someone who got the warning and parse the new format while continuing to support the old one.

This seems not ideal, but might it be good enough? The alternative would be a stricter form of stabilisation that makes all current fields mandatory, any future fields optional and all enums non-exhaustive and strings have an undefined format.

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -266,6 +281,36 @@ impl TargetInfo {
...
+                res
+            };
+
+            let target_spec: TargetSpec = serde_json::from_str(&target_spec_json.as_str())
+                .with_context(|| {
+                    format!(
+                    "failed to parse target spec from `rustc --print=target-spec-json`, got:\n{}",
+                    output
+                )
+                })?;
```
Nilstrieb, line 312: can you please make this non-fatal, such that you just keep a `None` if the format doesn't match what you expect (and probably also when not using the nightly channel, the `RUSTC_BOOTSTRAP=1` is quite problematic above and should be avoided), and then avoid emitting any error in that case.

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -294,6 +306,38 @@ impl TargetInfo {
                 gctx.shell().warn("non-trivial mutual dependency between target-specific configuration and RUSTFLAGS")?;
             }
 
+            let mut supports_std: Option<bool> = None;
+
+            // The '--print=target-spec-json' is an unstable option of rustc, therefore if
+            // rustc is not a 'Prerelease' build, it's safe to assume that we cannot fetch
+            // the target-spec-json. So don't bother trying.
+            if rustc.version.is_prerelease() {
```
weihanglo, line 314: We normally use [`GlobalContext::nightly_features_allowed`](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/util/context/struct.GlobalContext.html#structfield.nightly_features_allowed) to detect it. Any reason preferring `is_prerelease` over that?

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -294,6 +306,38 @@ impl TargetInfo {
...
+                let mut target_spec_process = rustc.workspace_process();
+                apply_env_config(gctx, &mut target_spec_process)?;
+                target_spec_process
```
weihanglo, line 317: >  I would suggest collecting the JSON information in a separate call to cached_output. That would also avoid needing to guess where the JSON boundaries are.

Couldn't we again use `--print=crate-name`, which print `___` as a delimiter again? I know that is quite hacky, though a second call to `rustc` isn't a good thing for cargo startup time, especially for the stabilization of `cargo -Zscript`.

src/cargo/core/compiler/standard_lib.rs
```diff
@@ -72,6 +72,18 @@ pub fn resolve_std<'gctx>(
...
     }
 
+    // check that targets support building std
+    if crates.contains(&"std".to_string()) {
+        for (target, target_info) in target_data.target_info() {
+            if target_info.supports_std == Some(false) {
+                return Err(anyhow::Error::msg(format!(
+                    "building std is not supported on this target: {:?}",
+                    target.short_name()
+                )));
```
weihanglo, line 82: `anyhow::bail!` is a better way to handle this.

tests/testsuite/standard_lib.rs
```diff
@@ -323,6 +329,33 @@ fn check_core() {
...
+        .build();
+
+    p.cargo("build")
+        .build_std(&setup)
+        .target("aarch64-unknown-none")
+        .with_status(101)
+        .with_stderr_data(
+            "\
+[ERROR] building std is not supported on this target: [..]
+",
```
weihanglo, line 354: Unless intentionally, would recommend wrapping with `str![[...]]` so that the snapshot could get auto-updated.

tests/testsuite/freshness.rs
```diff
@@ -2493,6 +2493,9 @@ LLVM version: 9.0
             "src/main.rs",
             r#"
             fn main() {
+                if std::env::args_os().any(|s| s.into_string().unwrap().contains("unstable-options")) {
```
weihanglo, line 2496: Why do we need this guard? Seems the nightly channel detection is off?

weihanglo (commented):
Regardless of my review, Eric's concern in <https://github.com/rust-lang/cargo/pull/14183#issuecomment-2206815406> should be deeply considered.

src/cargo/core/compiler/build_context/target_info.rs
harmou01: No reason, I wasn't aware of the preferred method for detecting this. I'll make this change.

src/cargo/core/compiler/standard_lib.rs
harmou01: Noted.

src/cargo/core/compiler/build_context/target_info.rs
harmou01: I'm not sure what the purpose of adding the delimiter back in would be? Do you mean reverting back to doing this all in a single rustc call?

harmou01:
@weihanglo Regarding Eric's concerns about the use of the `--print=target-spec-json` and the possibility of it changing, @adamgemmell has an RFC open to add a `--print=std-support` flag, which gives us a simpler and more reliable method for determining std support. We wouldn't have to grab the entire target-spec-json anymore, and therefore wouldn't risk things changing between versions, because rustc would always give a true/false answer. I think this should address Eric's concerns.

The second rustc query wouldn't be needed either if this RFC is implemented, we could revert back to collecting everything with a single query.

https://github.com/rust-lang/rfcs/pull/3693

src/cargo/core/compiler/build_context/target_info.rs
adamgemmell: We put this in a second rustc call to avoid RUSTC_BOOTSTRAP being used on the first one. If startup time is a concern we could ensure it only runs if build-std is requested.

adamgemmell:
> [rust-lang/rfcs#3693](https://github.com/rust-lang/rfcs/pull/3693)

To elaborate on this we tried to make this patch improve things without waiting for the RFC's proposal to become stable by removing RUSTC_BOOTSTRAP and making Cargo tolerate all failure cases. If you'd feel more comfortable waiting for that solution then we're fine with that of course.

tests/testsuite/standard_lib.rs
harmou01: Noted, will make this change.

src/cargo/core/compiler/build_context/target_info.rs
weihanglo: > If startup time is a concern we could ensure it only runs if build-std is requested.


That would be ideal! build_std can be accessed via `gctx.cli_unstable(). build_std.is_some()` 

https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/features/struct.CliUnstable.html#structfield.build_std

tests/testsuite/freshness.rs
harmou01: This test case was failing for me locally, and this fixed the issue. However i'm not seeing this fail now, so it's possible it can be reverted.

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -294,6 +306,42 @@ impl TargetInfo {
...
+                if target_spec_cached_output.is_ok() {
+                    let (stdout, _stderr) = target_spec_cached_output.unwrap();
+                    match serde_json::from_str(&stdout) {
+                        Result::Ok(val) => {
+                            let target_spec: TargetSpec = val;
+                            supports_std = target_spec.metadata.std;
+                        }
+                        Result::Err(_) => (),
+                    }
+                }
```
weihanglo, line 342: `rustc.cached_output` caches the rustc output to `target/.rustc_info.json` file. I guess it is better not to cache it. A direct call to `ProcessBuilder::output` is probably good enough. Would be something like:

```suggestion
                #[derive(Deserialize)]
                pub struct Metadata {
                    pub std: Option<bool>,
                }

                #[derive(Deserialize)]
                pub struct TargetSpec {
                    pub metadata: Metadata,
                }

                if let Ok(output) = target_spec_process.output() {
                    if let Ok(spec) = serde_json::from_slice::<TargetSpec>(&output.stdout) {
                        supports_std = spec.metadata.std;
                    }
                }
```

tests/testsuite/cfg.rs
weihanglo, line 1: As Eric has noted, target-spec-json is not going to be stabilized in a near feature, I would suggest not adding a new mode to this test. It is already complicated. The `test_std_on_unsupported_target` test should have covered the parsing end-to-end pretty much.

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -56,6 +58,16 @@ pub struct TargetInfo {
     pub rustdocflags: Rc<[String]>,
 }
 
+#[derive(Deserialize)]
+pub struct Metadata {
+    pub std: Option<bool>,
+}
+
+#[derive(Deserialize)]
+pub struct TargetSpec {
```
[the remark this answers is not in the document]
weihanglo: See also the suggestion in <https://github.com/rust-lang/cargo/pull/14183#discussion_r1839034621>

tests/testsuite/cfg.rs
adamgemmell: The test checks something different - that Cargo is tolerant of parsing failures should the output of target-spec-json change, which could cause confusing failures in the future if that behaviour were to change.

weihanglo: Hmm… what kind of change will fail this cfg test but not `test_std_on_unsupported_target`?

To myself I hope that we don't do this. We don't enable build-std related in rust-lang/rust CI because we don't want any change made in the compiler is blocked on Cargo side. This test is run unconditionally which may increase the chance of it.

We're tracking the overall build-std testing strategy in <https://github.com/rust-lang/wg-cargo-std-aware/issues/85>.

Maybe an alternative is providing trace logs when Cargo failed on probing/parsing.

adamgemmell: My mistake, I can't actually think of a scenario that wouldn't fail the other test too.

tests/testsuite/standard_lib.rs
```diff
@@ -183,6 +184,11 @@ impl BuildStd for Execs {
         self.arg("--target").arg(rustc_host());
         self
     }
+
+    fn target(&mut self, target: &str) -> &mut Self {
```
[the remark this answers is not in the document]
weihanglo: Thinking about it more. I think this PR shouldn't block on #14843, given #14843 may or may not land anytime sooner.

weihanglo (approved):
Thanks for working through the feedback—looks great now!

PR #14193, opened 2024-07-05
test: migrate build_plan and build_script to snapbox

eth3lbert:

Migrate following to snapbox:

- `tests/testsuite/build_plan.rs`
- `tests/testsuite/build_script.rs`

### DIFF
--- a/tests/testsuite/build_plan.rs
+++ b/tests/testsuite/build_plan.rs
@@ -1,9 +1,8 @@
 //! Tests for --build-plan feature.
 
-#![allow(deprecated)]
-
+use cargo_test_support::prelude::*;
 use cargo_test_support::registry::Package;
-use cargo_test_support::{basic_bin_manifest, basic_manifest, main_file, project};
+use cargo_test_support::{basic_bin_manifest, basic_manifest, main_file, project, str};
 
 #[cargo_test]
 fn cargo_build_plan_simple() {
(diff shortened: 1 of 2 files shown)


tests/testsuite/build_plan.rs
```diff
@@ -142,58 +154,75 @@ fn cargo_build_plan_build_script() {
...
+      ],
+      "package_name": "foo",
+      "package_version": "0.5.0",
+      "program": "rustc",
+      "target_kind": [
+        "custom-build"
+      ]
+    },
+    {
+      "args": [],
```
epage, line 184: Should we glob this like it did before?

eth3lbert: Great catch! I'll fix it.

tests/testsuite/build_script.rs
```diff
@@ -3741,6 +3780,7 @@ fn custom_target_dir() {
     p.cargo("build -v").run();
 }
 
+#[allow(deprecated)]
```
eth3lbert, line 3783: `with_stderr_does_not_contain` is being used in this function.

tests/testsuite/build_script.rs
```diff
@@ -4921,55 +5021,54 @@ fn rerun_if_directory() {
...
 
     let fresh = || {
-        p.cargo("check").with_stderr("[FINISHED] [..]").run();
+        p.cargo("check")
+            .with_stderr_data(
+                "\
+[FINISHED] [..]
+",
+            )
+            .run();
```
eth3lbert, line 5035: I think this is less important compared to the `dirty`, which I'm currently just leaving as-is.  Let me know if we want to rewrite it with expected argument.

Issue #7723, opened 2019-12-19
Tracking issue for config-include


ehuss:
Implementation: #7649
Original proposal: #6699
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#config-include
Issues: https://github.com/rust-lang/cargo/labels/Z-config-include

**Summary**

Adds the `include` config key to include other config files.

**Unresolved issues**

* [x] Should there be any restriction on the filename? Currently there isn't. I was initially considering restricting to `.cargo/config-*` in case we want to safely use other filenames in `.cargo/` for other purposes. However, it seems like an unnatural restriction. Another restriction is that it should maybe end in `.toml`?
  * The team has decided on putting a restriction on allowing `.toml` extension only. See <https://github.com/rust-lang/cargo/issues/7723#issuecomment-1602641026>
* [ ] Inconsistent relative path behavior in configurations: https://github.com/rust-lang/cargo/issues/15769
* [x] Optional include supports and syntax: <https://github.com/rust-lang/cargo/issues/7723#issuecomment-2386278512>.

[FCP](https://github.com/rust-lang/cargo/issues/7723#issuecomment-1602641051)

ensc:
Some additional wishes (from #9306):

- allow to include multiple files

  ```toml
  include = "foo.toml"
  include = "bar.toml"
  include = "conf.d/*.toml"
  ```

 - allow to ignore missing files; e.g. either by a special syntax like leading "-":

   ```toml
   include = "-can-be-missing.toml"
   include = "required.toml"
   ```

   or by using special keywords (e.g. `required` vs `include`)

   This feature is required for local configuration (e.g. the setup of `build.target-dir`) within a more compliex environment

- substitute placeholders within the filename; e.g.

  - `%h` ... local hostname
  - `%u` ... local username
  - `%H` ... hash of absolute `Cargo.toml` path
  - ...

jonhoo:
Some quick thoughts on this:

[...]

PR #14196, opened 2024-07-05
Allow enabling `config-include` feature in config

lschuermann:

Prior to this change, it is not possible to enable the unstable `config-include` flag from within a config file itself. This is due to the fact that, while cargo does reload configuration files if this flag is set, it does not parse the top-level configuration file's unstable flags before checking whether this flag is present.

This commit forces cargo to load unstable features before this check, and if the flag is present, re-loads configuration files with `config-include` enabled.

Addresses https://github.com/rust-lang/cargo/issues/7723#issuecomment-2206667822.

This PR adds a testcase for this behavior to cargo's testsuite. In general, it can be replicated with a config similar to that in the testcase, or as illustrated in https://github.com/rust-lang/cargo/issues/7723#issuecomment-2206667822.

### Additional information

~I'm not sure whether this is the right(TM) fix, or whether the second `self.load_unstable_flags_from_config()?;` at the bottom of `configure()` would still be required. I have a feeling it might not be, as `reload_rooted_at` also calls out to `load_unstable_flags_from_config`, to collect flags from any newly included configs. If that is not called, no other file was loaded, and thus there should not be a need to re-load unstable flags.~

Resolved: https://github.com/rust-lang/cargo/pull/14196#discussion_r1667205897

### DIFF
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -1035,6 +1035,11 @@ impl GlobalContext {
             self.cli_config = Some(cli_config.iter().map(|s| s.to_string()).collect());
             self.merge_cli_args()?;
         }
+
+        // Load the unstable flags from config file here first, as the config
+        // file itself may enable inclusion of other configs. In that case, we
+        // want to re-load configs with includes enabled:
+        self.load_unstable_flags_from_config()?;
         if self.unstable_flags.config_include {
             // If the config was already loaded (like when fetching the
             // `[alias]` table), it was loaded with includes disabled because
@@ -1090,8 +1095,6 @@ impl GlobalContext {
                 .unwrap_or(false);
         let cli_target_dir = target_dir.as_ref().map(|dir| Filesystem::new(dir.clone()));
         self.target_dir = cli_target_dir;
-
-        self.load_unstable_flags_from_config()?;
 
         Ok(())
     }
--- a/tests/testsuite/config_include.rs
+++ b/tests/testsuite/config_include.rs
@@ -49,6 +49,37 @@ fn simple() {
 }
 
 #[cargo_test]
+fn enable_in_unstable_config() {
+    // config-include enabled in the unstable config table:
+    write_config_at(
+        ".cargo/config.toml",
+        "
+        include = 'other.toml'
+        key1 = 1
+        key2 = 2
+
+        [unstable]
+        config-include = true
+        ",
+    );
+    write_config_at(
+        ".cargo/other.toml",
+        "
+        key2 = 3
+        key3 = 4
+        ",
+    );
+    let gctx = GlobalContextBuilder::new()
+        .nightly_features_allowed(true)
+        .build();
+    // On this commit, enabling `config-include` in the top-level
+    // configuration does not yet work.
+    assert_eq!(gctx.get::<i32>("key1").unwrap(), 1);
+    assert_eq!(gctx.get::<i32>("key2").unwrap(), 2);
+    assert_eq!(gctx.get::<i32>("key3").ok(), None);
+}
+
+#[cargo_test]
 fn works_with_cli() {
     write_config_at(
         ".cargo/config.toml",
--- a/tests/testsuite/config_include.rs
+++ b/tests/testsuite/config_include.rs
@@ -72,11 +72,9 @@ fn enable_in_unstable_config() {
     let gctx = GlobalContextBuilder::new()
         .nightly_features_allowed(true)
         .build();
-    // On this commit, enabling `config-include` in the top-level
-    // configuration does not yet work.
     assert_eq!(gctx.get::<i32>("key1").unwrap(), 1);
     assert_eq!(gctx.get::<i32>("key2").unwrap(), 2);
-    assert_eq!(gctx.get::<i32>("key3").ok(), None);
+    assert_eq!(gctx.get::<i32>("key3").unwrap(), 4);
 }
 
 #[cargo_test]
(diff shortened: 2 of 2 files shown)


epage:
r? weihanglo

If thats ok with you as you have a lot more context on this feature.

tests/testsuite/config_include.rs
```diff
@@ -48,6 +48,35 @@ fn simple() {
...
+        ",
+    );
+    let gctx = GlobalContextBuilder::new()
+        .nightly_features_allowed(true)
+        .build();
+    assert_eq!(gctx.get::<i32>("key1").unwrap(), 1);
+    assert_eq!(gctx.get::<i32>("key2").unwrap(), 2);
+    assert_eq!(gctx.get::<i32>("key3").unwrap(), 4);
+}
+
```
weihanglo, line 79: Could you add a test to ensure that

* A mix of hierarchy and include works correctly.
* Fields under the `[unstable]` table are merged correctly.



```
#[cargo_test]
fn mix_of_hierarchy_and_include() {
    write_config_at(
        "foo/.cargo/config.toml",
        "
        include = 'other.toml'
        key1 = 1

        # also make sure unstable flags merge in the correct order
        [unstable]
        features = ['1']
        ",
    );
    write_config_at(
        "foo/.cargo/other.toml",
        "
        key1 = 2
        key2 = 2

        [unstable]
        features = ['2']
        ",
    );
    write_config_at(
        ".cargo/config.toml",
        "
        include = 'other.toml'
        key1 = 3
        key2 = 3
        key3 = 3

        [unstable]
        features = ['3']
        ",
    );
    write_config_at(
        ".cargo/other.toml",
        "
        key1 = 4
        key2 = 4
        key3 = 4
        key4 = 4

        [unstable]
        features = ['4']
        ",
    );
    let gctx = GlobalContextBuilder::new()
        .unstable_flag("config-include")
        .cwd("foo")
        .nightly_features_allowed(true)
        .build();
    assert_eq!(gctx.get::<i32>("key1").unwrap(), 1);
    assert_eq!(gctx.get::<i32>("key2").unwrap(), 2);
    assert_eq!(gctx.get::<i32>("key3").unwrap(), 3);
    assert_eq!(gctx.get::<i32>("key4").unwrap(), 4);
    assert_eq!(
        gctx.get::<Vec<String>>("unstable.features").unwrap(),
        vec![
            "4".to_string(),
            "3".to_string(),
            "2".to_string(),
            "1".to_string()
        ]
    );
}
```

weihanglo (changes requested):
Thanks for the contribution!

Could you split this into two commits that the first one includes the new test, which demonstrates the buggy behavior and passes the CI pipeline. The second commit fixes both the behavior and the test. This pattern is [documented here as a tip](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request), which helps use make sure we're not fixing a nonexistent bug. The first commit is also served as an [MRE](https://en.wikipedia.org/wiki/Minimal_reproducible_example).

src/cargo/util/context/mod.rs
```diff
@@ -1035,6 +1035,11 @@ impl GlobalContext {
             self.cli_config = Some(cli_config.iter().map(|s| s.to_string()).collect());
             self.merge_cli_args()?;
         }
+
+        // Load the unstable flags from config file here first, as the config
+        // file itself may enable inclusion of other configs. In that case, we
+        // want to re-load configs with includes enabled:
+        self.load_unstable_flags_from_config()?;
```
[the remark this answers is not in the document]
lschuermann: Removed, thanks for the confirmation!

tests/testsuite/config_include.rs
weihanglo: Yeah. That sounds good to me :)

lschuermann:
> Could you split this into two commits that the first one includes the new test, which demonstrates the buggy behavior and passes the CI pipeline. The second commit fixes both the behavior and the test. This pattern is [documented here as a tip](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request), which helps use make sure we're not fixing a nonexistent bug. The first commit is also served as an [MRE](https://en.wikipedia.org/wiki/Minimal_reproducible_example).

I don't know whether there's any infrastructure in place to automate this check (other than pushing every commit individually), but all commits pass the respective tests for me locally:

```
> git rebase -i --exec 'git log --format="%C(auto) %h %s" -1 && cargo test -q --test testsuite -- config_include' HEAD~3
hint: Waiting for your editor to close the file... Waiting for Emacs...
Executing: git log --format="%C(auto) %h %s" -1 && cargo test -q --test testsuite -- config_include
 8bfc31f97 test(add): show config-include cannot be enabled in top-level config

running 13 tests
.............
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 3353 filtered out; finished in 0.31s

Executing: git log --format="%C(auto) %h %s" -1 && cargo test -q --test testsuite -- config_include
 45e5cb496 test(add): show config-include correctly merges unstable configs

running 15 tests
...............
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 3353 filtered out; finished in 0.35s

Executing: git log --format="%C(auto) %h %s" -1 && cargo test -q --test testsuite -- config_include
 0e35bea6c Allow enabling `config-include` feature in config

running 15 tests
...............
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 3353 filtered out; finished in 0.33s

Successfully rebased and updated refs/heads/dev/config-include-enable-in-config.
```

weihanglo (approved):
Awesome. Let's merge it!

weihanglo:
> I don't know whether there's any infrastructure in place to automate this check (other than pushing every commit individually), but all commits pass the respective tests for me locally:

Not really. CI doesn't enforce that. It's a good idea if we make it an optional CI job, though PR with lots of commits might not like it 😆.

PR #14197, opened 2024-07-05
Support target-spec json file extension in case-insensitive form

boozook:
Following https://github.com/rust-lang/rust/pull/127389 to complete it on both sides. Small minor improvement without breaking changes.

Currently user's target json file can have only `.json` ext, but for file that ends with `.JSON' (in caps).

It could be useful in some old OS and old file-systems such as FAT, for example.

## Performance cost:

- In case of file ext is normal `json` - no changes.
- Otherwise added comparison using `eq_ignore_ascii_case` for __only latest 5 chars__ of the string.

_\*"no changes" means there in negative case may be two additional derefs, alloc one little `iter`-over-`&str` and ≈1 char-comparison iteration. It could be optimised if firstly search `.` to get to know "is it normal name or maybe file"._

## Testing

I did not add any new tests for specs, such as with file-extensions like `JsOn` because minor change and doesn't changes existing behaviour, so existing tests are often. Should I add tests?

How to test:
1. use rustc from https://github.com/rust-lang/rust/pull/127389
1. `rustc --print target-spec-json -Zunstable-options > ./some-target.JSON`
1. remove `is-builtin` field from the `some-target.JSON` or set to `false`
1. `cargo build --target=some-target.JSON` should work

## Implementation note

I've separated the comparison into a methods for use instead of just `str.ends_with(".json")` and to maintain future compatibility. 
The method is public for the same reasons and for potential future usage, including other dependent crates.

- - -

_This text looks like a copypasta of that PR. That's not exactly so. The changes are similar, and there are noticeable differences in the two texts. I apologize if this has caused inconvenience._


epage:
Moved to draft as this is dependent on an unreviewer PR in rust-lang/rust for an Issue that hasn't had an MCP.

weihanglo:
Close as the counterpart in rust-lang/rust got closed due to inactivity: https://github.com/rust-lang/rust/pull/127389#issuecomment-3015141164

PR #14185, opened 2024-07-03
test: migrate cfg and check to snapbox

eth3lbert:
### What does this PR try to resolve?

Part of https://github.com/rust-lang/cargo/issues/14039.

Migrate following to snapbox:

- `tests/testsuite/cfg.rs`
- `tests/testsuite/check.rs`

PR #14181, opened 2024-07-03
test: migrate patch, pkgid, proc_macro and progress to snapbox

eth3lbert:
### What does this PR try to resolve?

Part of https://github.com/rust-lang/cargo/issues/14039.

Migrate following to snapbox:

- `tests/testsuite/patch.rs`
- `tests/testsuite/pkgid.rs`
- `tests/testsuite/proc_macro.rs`
- `tests/testsuite/progress.rs`

PR #14198, opened 2024-07-05
fix(test): Restore `does_not_contain` for check

eth3lbert:
This is partial revert of #14185
(27062476a86e864479ef6bde1828c624b643de4b)

Although the changes made in that commit are valid, for the same reason mentioned in the comment https://github.com/rust-lang/cargo/pull/14181#discussion_r1667089718, "People are unlikely to see these comments when updating snapshots". Therefore, it would be better to revert these parts.

--- 
Let me know if anything's missing!

### DIFF
--- a/tests/testsuite/check.rs
+++ b/tests/testsuite/check.rs
@@ -407,6 +407,7 @@ fn check_all() {
         .run();
 }
 
+#[allow(deprecated)]
 #[cargo_test]
 fn check_all_exclude() {
     let p = project()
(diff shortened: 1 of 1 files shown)


weihanglo (approved):

weihanglo:
That was my fault. Thanks for the prompt fix.

eth3lbert:
> That was my fault. Thanks for the prompt fix.

No worries :)

PR #14202, opened 2024-07-06
Add workflow to publish Cargo automatically

pietroalbini:
One of the last manual pieces of the Rust release process is publishing Cargo, and as it happens after the release itself it is often forgotten (I am often guilty of that). This PR aims to fully automate that.

Nowadays tagging Cargo happens automatically, as the release process creates and pushes a signed tag for the release. So the only piece remaining is automatically executing the `publish.py` script. This PR adds a workflow triggered by the tag push to run `publish.py` with a [rust-lang-owner](https://crates.io/users/rust-lang-owner) scoped token.

The secret is stored in the [`release` environment](https://github.com/rust-lang/cargo/settings/environments/3357627654/edit), which can only be used during tag pushes.

cc @Mark-Simulacrum

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,33 @@
+# Publish Cargo to crates.io whenever a new tag is pushed. Tags are pushed by
+# the Rust release process (https://github.com/rust-lang/promote-release),
+# which will cause this workflow to run.
+
+name: Release
+on:
+  push:
+    tags:
+      - "**"
+
+# Prevent multiple releases from starting at the same time.
+concurrency:
+  group: release
+
+jobs:
+  crates-io:
+    name: Publish on crates.io
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    # Gain access to the crates.io publishing token.
+    environment:
+      name: release
+
+    steps:
+      - name: Checkout the source code
+        uses: actions/checkout@v4
+
+      - name: Publish Cargo to crates.io
+        run: ./publish.py
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
--- a/publish.py
+++ b/publish.py
@@ -15,6 +15,9 @@ import urllib.request
 from urllib.error import HTTPError
 
 
+# Whenever you add a new crate to this list that does NOT start with "cargo-"
+# you must reach out to the infra team to add the crate to the list of crates
+# allowed to be published from the "cargo CI" crates.io token.
 TO_PUBLISH = [
     'credential/cargo-credential',
     'credential/cargo-credential-libsecret',


.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
+# Publish Cargo to crates.io whenever a new tag is pushed. Tags are pushed by
+# the Rust release process (https://github.com/rust-lang/promote-release),
+# which will cause this workflow to run.
```
weihanglo, line 3: I believe I am allowed to manually create tags. Would that also trigger this workflow and how should we prevent it?

image

weihanglo:
> The secret is stored in the [`release` environment](https://github.com/rust-lang/cargo/settings/environments/3357627654/edit), which can only be used during tag pushes.

BTW I have no permission to read this page (and shouldn't have). So yeah need t-infra people to verity it.

.github/workflows/release.yml
pietroalbini: Yes you creating tags would trigger the workflow. I guess we could setup a ruleset only allowing promote-release to create tags in the repository, but I'm not sure if it's worth it.

In the other repositories we have auto-publish configured in (https://github.com/rust-lang/infra-team/issues/117) we let everyone with write access to the repo publish releases. Also, having the Cargo team be able to issue new patch releases of Cargo if problems with the crates.io-published versions are discovered (like a wrong dependency) could reduce friction.

@Mark-Simulacrum do you have any concerns?

pietroalbini:
> BTW I have no permission to read this page (and shouldn't have). So yeah need t-infra people to verity it.

Current contents of the page:

image

ehuss:
Thanks @pietroalbini!

The cargo team has discussed trying to do this in the past. We were stuck on two issues:
* How to selectively publish individual crates independent of the Rust release process? We had a few different ideas, but didn't really settle on anything.
* How to guard or otherwise monitor when releases are made? GitHub doesn't really provide a nice way to do this like there is with PRs (which require an independent reviewer, and send out notifications to everyone). We also had some ideas on this, but didn't settle on one. It would be nice if *something* (crates.io or GitHub) at least sent out a notification.

I don't want to block anything on these, just wanted to let you know what we've been thinking.

weihanglo:
> * How to selectively publish individual crates independent of the Rust release process? We had a few different ideas, but didn't really settle on anything.

Perhaps this helps <https://github.com/rust-lang/cargo/issues/13397>?

> * How to guard or otherwise monitor when releases are made? GitHub doesn't really provide a nice way to do this like there is with PRs (which require an independent reviewer, and send out notifications to everyone). We also had some ideas on this, but didn't settle on one. It would be nice if *something* (crates.io or GitHub) at least sent out a notification.

if we create a release for each new tag, there is notification for new releases

.github/workflows/release.yml
Mark-Simulacrum: Ideally, I think across all of our repositories we would require 2 people to approve release issuance. For most of the repositories we have that's hard though. (And 2 people is still somewhat weak in terms of mechanism).

Maybe we can at least restrict publication to just cargo team members? e.g., triagebot needs write access, but ideally would not be able to publish a release (at least trivially). If we gated it behind an environment with manual approval (similar to what we're planning to do for bors production deployments), that could be just the Cargo team. And I'd feel a little better about the approval being an intentional UI click rather than "just" `git push ...` with a typo that makes it a tag.

ehuss:
> Perhaps this helps #13397?

Not really. We need to be able to do something like "publish `home` on the `master` branch, but don't publish `cargo`". Just looking at the version number doesn't help since we proactively bump cargo's version.

Which makes me think we won't be able to tag releases like `home` with this PR... 🤔
We might need to consider filtering based on the tag name?

In git2-rs, I use workflow dispatches which have a little GUI that lets you select which crates to publish.

> if we create a release for each new tag, there is notification for new releases

Yea, I use that in some other repos and it works pretty well. Something we could consider, though I wouldn't want to block on it.

pietroalbini:
I would keep publishing other crates as out of scope right now, my main concern here is automating the last bit of the Rust release process :D

> Which makes me think we won't be able to tag releases like home with this PR... 🤔
We might need to consider filtering based on the tag name?
>
> In git2-rs, I use workflow dispatches which have a little GUI that lets you select which crates to publish.

Do you plan in the future to tag the release of other crates (like `home-1.0.0`), while keeping the Cargo tags as just the version number (like `0.80.0`)? If so, I can change the trigger rule and the workflow to only apply on Cargo releases, leaving the door open for publishing other crates outside of this process in the future.

.github/workflows/release.yml
pietroalbini: So you'd like to require the Cargo team's approval even during normal releases (i.e. when pushed by promote-release)?

Mark-Simulacrum: No, not necessarily. I'm comfortable with automation making releases, that seems fine to do without approvals (though with if required is likely also ok). But manual publishes I think do need more than just write access.

pietroalbini: Created a repository rule to only allow promote-release to push tags to this repository.

pietroalbini:
OK, I think with the restriction of only allowing promote-release to publish tags I think this is ready to be merged? Assuming we defer the discussion of other crates to a future PR.

weihanglo (approved):
Thanks. This is not a one-way-door decision so  I am gonna merge it

Issue #14195, opened 2024-07-05
Linker and links are ignored with target-applies-to-host and no --target flag


gmorenz:
### Problem

This is the exact same bug as #10744 except with regards to `linker` and `links` instead of `rustflags` and `rustdocflags`. When building with `target-applies-to-host=false` and *not* passing a `--target` flag `linker` and `links` are ignored because the host-values of those are set to the default and then the artifact units end up picking up the host values (see "How it is implemented" in #13900 for an explanation of how that happens).

### Steps

Running on x86_64-unknown-linux-gnu (if you have a different host substitute in the correct architecture)

```
cargo new demo
cd demo
cargo build -Z target-applies-to-host --config target-applies-to-host=false  --config target.x86_64-unknown-linux-gnu.linker="\"nope\""
```

Expected behavior: ``error: linker `nope` not found``

Actual behavior: It builds without error

```
cargo new demo
cd demo
echo "int myfunc() {}" | gcc -xc -shared -o libmylib.so -
sed -i '/\[package\]/a links="mylib"' Cargo.toml
echo 'fn main() { panic!("build.rs run") }' > build.rs
cargo build \
    --config 'target.x86_64-unknown-linux-gnu.mylib.rustc-link-lib=["mylib"]' \
    --config 'target.x86_64-unknown-linux-gnu.mylib.rustc-link-search=["native=."]' \
    -Z target-applies-to-host \
    --config target-applies-to-host=false
```

Expected behavior: It builds without error

Expected *bug*: It fails to find mylib

Actual behavior:

gmorenz:
Oops, posted before I intended to, wait a second if you're reading this already

weihanglo:
I am not @rustbot :P

gmorenz:
I should say I intend to fix this since it's a natural continuation of what I already fixed, but thought that I should at least create an issue before doing so instead of completely ignoring process.

gmorenz:
Alright, went ahead and tracked down the root cause of the panic for the links-override half of this issue, tl;dr it is a side effect of failing to properly apply the link overrides flags.

`compute_deps_custom_build` gets called through `compile_ws -> create_bcx (before rebuild_unit_depdendencies) -> build_unit_dependencies --<main path>--> deps_of -> compute_deps -> compute_deps_custom_build`, sees that `script_override(links, unit.kind)` returns `Some`, and returns that there are no dependencies because it is overridden.

[...]

PR #14205, opened 2024-07-06
Fix passing of links-overrides with target-applies-to-host and an implicit target

gmorenz:

This fixes the link-overrides half of #14195, both the panic, and the fact that the field is being discarded, the latter of which caused the former as discussed in [the issue](https://github.com/rust-lang/cargo/issues/14195#issuecomment-2211481773).

It does so following the blueprint laid out in #13900 - which is also in my opinion the current best summary of the broader context.

For reviewing, comparing to the changes in #13900 might be useful.

### Additional information

I'm pushing a PR for the other half of #14195 simultaneously. I thought it better to keep the PRs small since they're independent, though if merged simultaneously there will be a conflict over the ordering of fields in `Unit`.


gmorenz:
Well, the test doesn't work on windows, converting to draft until I fix that.

gmorenz:
Marking this as ready for review, but leaving the fix for the test on windows as a separate comment for now because I'm hoping that there's a better way to link to a cdylib built with cargo on windows that I'm missing. See [the last commit](https://github.com/rust-lang/cargo/pull/14205/commits/0943b5c538dcb27c5e5c1b8c41ec2f93d707729e) for my current understanding of what you have to do to do that.

If there isn't a better way, I should probably squash the last commit into the first commit that introduces the test before this is merged.

src/cargo/core/compiler/unit.rs
```diff
@@ -82,6 +84,12 @@ pub struct UnitInner {
     /// [`BuildContext::extra_args_for`]: crate::core::compiler::build_context::BuildContext::extra_args_for
     /// [`TargetInfo.rustdocflags`]: crate::core::compiler::build_context::TargetInfo::rustdocflags
     pub rustdocflags: Arc<[String]>,
+    /// Build script override for the given library name.
+    ///
+    /// Any package with a `links` value for the given library name will skip
+    /// running its build script and instead use the given output from the
+    /// config file.
+    pub links_overrides: Arc<BTreeMap<String, BuildOutput>>,
```
weihanglo, line 92: Just realized that it should be possible to use `Rc` instead of `Arc`, for all these usages.

src/cargo/core/compiler/custom_build.rs
```diff
@@ -61,7 +61,7 @@ const OLD_CARGO_WARNING_SYNTAX: &str = "cargo:warning=";
 /// [the doc]: https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#cargo-warning
 const NEW_CARGO_WARNING_SYNTAX: &str = "cargo::warning=";
 /// Contains the parsed output of a custom build script.
-#[derive(Clone, Debug, Hash, Default)]
+#[derive(Clone, Debug, Hash, Default, PartialEq, Eq, PartialOrd, Ord)]
```
weihanglo, line 64: Also realized that there is an implication that these newly added fields would affect the order of any sorted units list. I don't think that would cause real issues though.

tests/testsuite/build_script.rs
```diff
@@ -5828,6 +5828,25 @@ fn links_overrides_with_target_applies_to_host() {
...
+    // The path is different on windows because we need the "name.dll.lib" file
+    // to link to it, and as per #3934 that only lives in `target/{profile}/deps`
+    // and isn't copied to `target/{profile}` like the .dll is. #3951 removed the
+    // hash mentioned in #3934 but did not change the directory of the .lib file.
+
+    // Meanwhile on other operating systems everything we need to link (which is
+    // just the linux .so/the macos .dylib) is copied to `target/{profile}` from
+    //  `deps` and as the blessed output artifact that seems to be the appropriate
+    // version to use.
+    let mylib_name = if cfg!(windows) { "mylib.dll" } else { "mylib" };
```
weihanglo, line 5843: Instead of doing this, perhaps we can go an easier route like other tests: Asserting rustflags for rustc invocation?

https://github.com/rust-lang/cargo/blob/6d685637f1853b705053500ff4abacdc51df9c91/tests/testsuite/build_script.rs#L1653-L1670

src/cargo/core/compiler/unit.rs
gmorenz: Huh, cargo check passes with that change at least. I think I must have just assumed they would cross a thread boundary somewhere. Will fix.

src/cargo/core/compiler/custom_build.rs
gmorenz: True... if this is a significant concern we could move the new fields to the end.

Ultimately I don't think I have the context to decide if it is. I think it is almost certainly publicly exposed given comments like [this one](https://github.com/rust-lang/cargo//blob/master/src/cargo/core/compiler/unit_dependencies.rs#L140)), and where these fields currently are in the struct is probably high enough to change the order in the cases [described in this comment](https://github.com/rust-lang/cargo//blob/master/src/cargo/core/compiler/unit.rs#L98). I wouldn't have thought that the ordering of units was something that was guaranteed to be stable between cargo versions.

I'll hold off on updating the rustflags `Arc` to `Rc` until we've decided if we need to move it to the bottom since that might as well be one PR.

tests/testsuite/build_script.rs
gmorenz: This definitely looks better, and is exactly the pre-existing test for link overrides that I failed to find. Will do this.

gmorenz:
Made promised changes, rebased on master.

weihanglo (approved):
Thanks. Looks pretty great!

weihanglo:
@bors r+

Haven't got time to review <https://github.com/rust-lang/cargo/issues/14195>, though I guess you need to rebase?

gmorenz:
> Haven't got time to review https://github.com/rust-lang/cargo/pull/14206, though I guess you need to rebase?

Yeah, also I need to incorporate the Arc -> Rc feedback. Will get to this later today, I've just been a bit busy and was putting it off (though I don't anticipate it taking more than 10 minutes).

When you say "haven't got time" are you looking forwards and saying I should ask someone else to review it, or just "not yet/right now"?

weihanglo:
It is in my queue, and I am happy to review when ready that :)

PR #14207, opened 2024-07-06
use "bootstrap" instead of "rustbuild"

onur-ozkan:
Let's stick with the single name "bootstrap" to refer to the bootstrap project to avoid confusion.

cc https://github.com/rust-lang/rust/pull/127434


weihanglo (approved):

weihanglo:
LGTM.

@bors r+

PR #14211, opened 2024-07-08
dont make new constant InternedString in hot path

Eh2406:

`InternedString::new` always takes a `Mutex` lock. This is not a big deal because `new` is usually only called as external io is done (i.e not in a hot loop) and cargo is single threaded so the lock is not contested. However, dealing with the "default" feature happens deep within the resolver (i.e in a hot loop). So we may as well remove this overhead. I can get runtime benchmarks if that is desired.

For context: My pubgrub work would very much like to run cargoes resolver simultaneously in multiple threads. This `Mutex` is a major obstacle for that effort.

Internal re-factor and test still pass.

Also confirmed the semantics of `static` in https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Const.20evaluation.20and.20equality

### Additional information

There are other constant InternedString's in `src/cargo/core/profiles.rs`. I don't think they are in a hot enough path to justify more `static`s. But I will change if requested.

### DIFF
--- a/src/cargo/util/interning.rs
+++ b/src/cargo/util/interning.rs
@@ -13,9 +13,18 @@ use std::str;
 use std::sync::Mutex;
 use std::sync::OnceLock;
 
+pub static INTERNED_DEFAULT: InternedString = InternedString { inner: "default" };
+
 fn interned_storage() -> std::sync::MutexGuard<'static, HashSet<&'static str>> {
     static STRING_CACHE: OnceLock<Mutex<HashSet<&'static str>>> = OnceLock::new();
-    STRING_CACHE.get_or_init(Default::default).lock().unwrap()
+    STRING_CACHE
+        .get_or_init(|| {
+            let mut out: HashSet<&'static str> = Default::default();
+            out.insert(INTERNED_DEFAULT.as_str());
+            Mutex::new(out)
+        })
+        .lock()
+        .unwrap()
 }
 
 #[derive(Clone, Copy)]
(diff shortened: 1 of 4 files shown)


src/cargo/util/interning.rs
```diff
@@ -13,7 +13,19 @@ use std::str;
 use std::sync::Mutex;
 use std::sync::OnceLock;
 
-static STRING_CACHE: OnceLock<Mutex<HashSet<&'static str>>> = OnceLock::new();
+pub static DEFAULT_INTERNED_STRING: InternedString = InternedString { inner: "default" };
```
Eh2406, line 16: I'm not sure what the clearest naming pattern for this is. The current name suggests `Default::default()`. I am open to suggestions!

epage: `INTERNED_DEFAULT`?

Naming it for the content (and not bothering with the type/storage) or naming it with `DEFAULT` last helps to break that confusion.

src/cargo/util/interning.rs
```diff
@@ -13,7 +13,19 @@ use std::str;
...
+
+fn cache_init() -> std::sync::MutexGuard<'static, HashSet<&'static str>> {
```
epage, line 18: nit: it would be easier to audit the behavior change if the pulling out of `cache_init` was in its own commit.

epage, line 18: imo this should just be called `interned_storage` or something.  The initialization is an implementation detail.

PR #14212, opened 2024-07-08
Fix compatible_with_older_cargo test.

ehuss:
This fixes the `global_cache_tracker::compatible_with_older_cargo` test which is failing with the latest nightly, with the following error:

```
thread 'main' panicked at src/cargo/core/compiler/fingerprint/mod.rs:1834:9:
assertion `left == right` failed
  left: "0b455c154b949b0d"
 right: "ce6d4698ea4438f2"
```

The problem is that it was running stable `cargo` from CARGO_HOME, but was always using nightly `rustc`. The assertion was triggering because it was reusing the same fingerprint file between stable and nightly, but the Hash impl changed in nightly via https://github.com/rust-lang/rust/pull/127297. Under normal circumstances, when switching between toolchains, the hash should change ensuring the toolchains each use a separate directory.

The fix here is to ensure that when running stable cargo that it also uses stable rustc. This helps ensure that the hashes between cargo invocations changes, and they use separate directories.

### DIFF
--- a/tests/testsuite/global_cache_tracker.rs
+++ b/tests/testsuite/global_cache_tracker.rs
@@ -20,6 +20,7 @@ use cargo_test_support::{
     thread_wait_timeout, Execs, Project,
 };
 use itertools::Itertools;
+use std::env;
 use std::fmt::Write;
 use std::path::Path;
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


epage:
Thanks!

@bors r+

PR #14213, opened 2024-07-08
test: Ensure --list test works with cargo-bloat

epage:
This is a follow up to #14113


ehuss:
Thanks!

I wonder if our test sandbox should sanitize PATH to prevent things like this?

@bors r+

PR #14214, opened 2024-07-09
test: migrate fetch and list_availables to snapbox

eth3lbert:

Migrate following to snapbox:

- `tests/testsuite/fetch.rs`
- `tests/testsuite/list_availables.rs`

### DIFF
--- a/tests/testsuite/list_availables.rs
+++ b/tests/testsuite/list_availables.rs
@@ -1,8 +1,6 @@
 //! Tests for packages/target filter flags giving suggestions on which
 //! packages/targets are available.
 
-#![allow(deprecated)]
-
 use cargo_test_support::prelude::*;
 use cargo_test_support::project;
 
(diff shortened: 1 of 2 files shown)


tests/testsuite/list_availables.rs
eth3lbert, line 1: Should we make all this snapshot-aware?

weihanglo: Yes. That helps if we are going to tweak the error messages.

eth3lbert: In order to achieve this goal, I've added a `SnapshotBuilder` to the testsuite. This, along with `ExpectedSnapshots`, also eliminates the need for bit masking in favor of pattern matching.

weihanglo: I am not sure whether `SnapshotBuilder` is better, though it isn't really a blocker. If somebody feels unease of it, we can always revert or improve later

weihanglo (approved):

weihanglo:
@bors r+

Thank you!

PR #14218, opened 2024-07-09
refactor: move get_source_id out of registry

hi-rustin:

This PR tries to do some refactorings to make sure we can use the registry API in https://github.com/rust-lang/cargo/pull/14141.

The code splits from [`7261982` (#14141)](https://github.com/rust-lang/cargo/pull/14141/commits/7261982a94531b978ee5be95415f3e2122d9285b) and [`3c17779` (#14141)](https://github.com/rust-lang/cargo/pull/14141/commits/3c177798b79782fc979d51145437eec23073bd31).

Because `cargo-information` has its own logic to obtain the RegistrySourceIds, we need to pass the source_ids into the registry function instead of calling it within the registry function.

Also, I refactored the function get_source_id to ensure we can reuse some common functions from it. This will help us avoid code duplication in https://github.com/rust-lang/cargo/pull/14141.

Original proposed in https://github.com/rust-lang/cargo/pull/14141#discussion_r1669278421 and https://github.com/rust-lang/cargo/pull/14141#discussion_r1669279924.

Just refactoring, so the old tests are enough.

### Additional information

r? @epage

### DIFF
--- a/src/cargo/ops/registry/login.rs
+++ b/src/cargo/ops/registry/login.rs
@@ -26,8 +26,15 @@ pub fn registry_login(
 ) -> CargoResult<()> {
     let source_ids = get_source_id(gctx, reg_or_index)?;
 
-    let login_url = match registry(gctx, token_from_cmdline.clone(), reg_or_index, false, None) {
-        Ok((registry, _)) => Some(format!("{}/me", registry.host())),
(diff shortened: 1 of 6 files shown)


hi-rustin (commented):

PR #14221, opened 2024-07-09
fix: Ensure dep/feature activates the dependency on 2024 

### LINKED ISSUE #14016: Bad error message when a strong dep features's depedency is an unused optional dependency

epage:

Fixes #14016

This doesn't revert the last commit of #14018 so that it works properly
on Editions 2021 and 2024.

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -322,7 +322,7 @@ fn resolve_toml(
             });
         resolved_toml.package = Some(resolved_package);
 
-        resolved_toml.features = resolve_features(original_toml.features.as_ref())?;
+        resolved_toml.features = resolve_features(original_toml.features.as_ref(), edition)?;
 
         resolved_toml.lib = targets::resolve_lib(
             original_toml.lib.as_ref(),
@@ -691,10 +691,33 @@ fn default_readme_from_package_root(package_root: &Path) -> Option<String> {
 #[tracing::instrument(skip_all)]
 fn resolve_features(
     original_features: Option<&BTreeMap<manifest::FeatureName, Vec<String>>>,
+    edition: Edition,
 ) -> CargoResult<Option<BTreeMap<manifest::FeatureName, Vec<String>>>> {
-    let Some(resolved_features) = original_features.cloned() else {
+    let Some(mut resolved_features) = original_features.cloned() else {
         return Ok(None);
     };
+
+    if Edition::Edition2024 <= edition {
+        for activations in resolved_features.values_mut() {
+            let mut deps = Vec::new();
+            for feature_value in activations.iter() {
+                let feature_value = FeatureValue::new(InternedString::new(feature_value));
+                let FeatureValue::DepFeature {
+                    dep_name,
+                    dep_feature: _,
+                    weak: false,
+                } = feature_value
+                else {
+                    continue;
+                };
+                let dep = FeatureValue::Dep { dep_name }.to_string();
+                if !activations.contains(&dep) {
+                    deps.push(dep);
+                }
+            }
+            activations.extend(deps);
+        }
+    }
 
     Ok(Some(resolved_features))
 }
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -291,7 +291,7 @@ fn resolve_toml(
         dev_dependencies2: None,
         build_dependencies: None,
         build_dependencies2: None,
-        features: original_toml.features.clone(),
+        features: None,
         target: None,
         replace: original_toml.replace.clone(),
         patch: original_toml.patch.clone(),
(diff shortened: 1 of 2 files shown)


epage:
For CI failure, see #14223

src/cargo/util/toml/mod.rs
```diff
@@ -691,11 +691,34 @@ fn default_readme_from_package_root(package_root: &Path) -> Option<String> {
...
 fn resolve_features(
     original_features: Option<&BTreeMap<manifest::FeatureName, Vec<String>>>,
+    edition: Edition,
 ) -> CargoResult<Option<BTreeMap<manifest::FeatureName, Vec<String>>>> {
-    let Some(resolved_features) = original_features.cloned() else {
+    let Some(mut resolved_features) = original_features.cloned() else {
         return Ok(None);
     };
 
+    if Edition::Edition2024 <= edition {
```
weihanglo, line 700: I understand you like Yoda style more, but forgot the reason.

Anyway, this is not a blocker.

src/cargo/util/toml/mod.rs
```diff
@@ -686,6 +688,40 @@ fn default_readme_from_package_root(package_root: &Path) -> Option<String> {
...
+) -> CargoResult<Option<BTreeMap<manifest::FeatureName, Vec<String>>>> {
+    let Some(mut resolved_features) = original_features.cloned() else {
+        return Ok(None);
+    };
+
+    if Edition::Edition2024 <= edition {
+        for activations in resolved_features.values_mut() {
+            let mut deps = Vec::new();
+            for feature_value in activations.iter() {
+                let feature_value = FeatureValue::new(InternedString::new(feature_value));
```
weihanglo, line 704: So, this PR chose the approach:

> Implicitly create `dep:serde` if `serde/derive` is used

I forgot if we had a conclusion somewhere.

src/cargo/util/toml/mod.rs
epage: I actually really dislike Yoda style for equality checks.  For these checks, I prefer number line order (the lower value is always on the left, so `<` and `<=` are only ever used and not `>` or `>=`)

src/cargo/util/toml/mod.rs
epage: See https://github.com/rust-lang/cargo/issues/14016#issuecomment-2161083450

> We discussed this in today's team meeting and lean towards `"dep_name/feature_name"` implying `"dep:dep_name"` in Edition 2024 with it being normalized in the published `Cargo.toml` file.

weihanglo: Why did we choose to do it here instead of in `Summary`?

https://github.com/rust-lang/cargo/blob/471287ffed354a468114caa5c72db97da6b6ddfb/src/cargo/util/toml/mod.rs#L1442-L1459

weihanglo: We want everything to be normalized on `resolved_toml`, right?

epage: By doing it here, we get `cargo publish` normalization "for free" and all data-structures are consistent for less chance of bugs (which also makes it easier to introspect for debugging)

weihanglo:
@bors r+

Thanks!

src/cargo/util/toml/mod.rs
TimDiekmann: > Implicitly create `dep:serde` if `serde/derive` is used

Doesn't this fail if `serde` is not an optional dependency?

PR #14224, opened 2024-07-09
docs(ref): Note MSRV for features in the docs

epage:
I mostly focused on manifests and build scripts as those seemed the highest priority.
I also went back only so far, assuming these would be the highest value items.

### DIFF
--- a/src/doc/src/reference/build-scripts.md
+++ b/src/doc/src/reference/build-scripts.md
@@ -77,13 +77,6 @@ Build scripts communicate with Cargo by printing to stdout. Cargo will
 interpret each line that starts with `cargo::` as an instruction that will
 influence compilation of the package. All other lines are ignored.
 
-> **Note:** The old invocation prefix `cargo:` (one colon only) is deprecated
-> and won't get any new features. To migrate, use two-colons prefix `cargo::`,
-> which was added in Rust 1.77. If you were using `cargo:KEY=VALUE` for
-> arbitrary links manifest key-value pairs, it is encouraged to switch to
-> `cargo::metadata=KEY=VALUE`. Stick to `cargo:` only if the support of Rust
-> version older than 1.77 is required.
-
 > The order of `cargo::` instructions printed by the build script *may*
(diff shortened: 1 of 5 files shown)


weihanglo (commented):
To me, it's nicer to scan if the MSRV requirement is at the top of the section.

Otherwise LGTM.

weihanglo:
To make it more outstanding, maybe we can have an icon noting MSRV requirement.

BTW, this reminds me anchors for environment variables: <https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Anchors.20for.20each.20environment.20variable.20in.20The.20Cargo.20Book>

epage:
> To me, it's nicer to scan if the MSRV requirement is at the top of the section.

I had considered that at one point.  I kept it at the bottom because
- The happy path would be that people not worry about this (especially 5 years out for these specific ones) and so putting it at the top seems to be optimizing for the non-happy path
- If this grows to need multiple notes on MSRV, then that could really clog things up, hiding away the details that say what the section they are viewing is even about.

weihanglo:
> The happy path would be that people not worry about this (especially 5 years out for these specific ones) and so putting it at the top seems to be optimizing for the non-happy path

Not entire agree on this. If there is only one line MSRV description, it doesn't really hurt people on the happy path. However, it's easier to slip from users on non-happy path. They are often library authors and may affect more downstream packages. [CMake doc](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#transitive-compile-properties) is the baseline of what we'll have if moved to the top.

One alternative is adding some styles, like Python [documetation](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop) and [pyproject.toml](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#metadata-generation) (which the former is at the bottom and the latter at the top).

For manifest fields for long term, I would like to see a metadata like what we have for configuration about optionality, default value, etc.

---

Given it is better than none, I am not going to block this. Thanks for the enhancement!

@bors r+

weihanglo (approved):

PR #14169, opened 2024-06-30
refactor(source): Clean up after PathSource/RecursivePathSource split

epage:
### What does this PR try to resolve?

This is a follow up to #13993 and prep for future improvements (e.g. cargo script, #10752)

Issue #10752, opened 2022-06-13
warning: skipping duplicate package `case` found at `~\.cargo\git\checkouts\cargo-..`


aminya:
### Problem

When using `cargo install --git`  to install a package like cargo itself, a lot of warnings are emitted about duplicated packages. It seems that these packages are only used for the tests, and not sure if the warnings affect the final build.

```
warning: skipping duplicate package `cargo-list-test-fixture` found at `~\.cargo\git\checkouts\cargo-..`
...
```

It seems that this is caused by this commit:
https://github.com/rust-lang/cargo/commit/e903f7dfc0e50fe3e11dcbc5456094c902fce6ce

![image](https://user-images.githubusercontent.com/16418197/173324225-e53f6ade-8498-4e20-964f-b2bf7bc42343.png)

### Steps

Try to build cargo with itself:
```
cargo install --git https://github.com/rust-lang/cargo
```

### Version

```text
cargo 1.63.0-nightly (38472bc19 2022-05-31)
```

epage:
Interesting.  cargo is searching recursively through cargo's git repo and finding all of the test cases we have in the repo and reporting a warning for those as we duplicate the manifest names.   These should not even be given as options for install for the warning to be activated.

Besides telling the user to run `cargo install --git https://github.com/rust-lang/cargo cargo` I wonder if there is some way we can allow repos to exclude directories from being searched for manifests.

epage:
We discussed this in the cargo team meeting and came up with a short term and long-term idea

Short term is to see if we can make the warning only appear when the package is `publish = true`

There is still the wider problem if packages being looked up that are internal.  Ideas
- a breadcrumb file that says "stop recursively walking for path/git sources"
- a new manifest key that says "stop recursing and use workspace lookup rules"

epage:
@danilhendrasr since you worked on the warnings in #10701, I thought I'd first check if you would be interested in seeing if the warning could be restricted to packages with `publish = false` to reduce noise from the warning.

danilhendrasr:
Sure, I'll take a stab at it.

epage:
Hmm, thinking about this, i suspect #10767 only gave people the tool to reduce the warnings but cargo still needs to opt-in to that tool, so going ahead and re-opening for now.

ehuss:
[...]

PR #14231, opened 2024-07-10
refactor(source): More RecursivePathSource clean up

epage:

This is a follow up to #13993 and #14169 and is part of my work towards #10752.


Muscraft (approved):
Thanks!

PR #14234, opened 2024-07-10
fix(overrides): Don't warn on duplicate packages from using '..'

epage:

As part of #10752, I was changing the "duplicate package" warning to be like:
```
[WARNING] skipping duplicate package `a2 v0.5.0 ([ROOT]/foo/b/../a)`:
  [ROOT]/foo/b/../a/a2/Cargo.toml
in favor of [ROOT]/foo/a/a2/Cargo.toml
```
and it showed that we were considering two paths to the same package to be duplicates.

This suppresses that warning.


weihanglo (approved):

weihanglo:
@bors r+


Looks fair!

PR #14233, opened 2024-07-10
fix(test): Redact elapsed time in the minutes time frame 

epage:
### What does this PR try to resolve?

This came up in #14231
```
---- expected: tests/testsuite/lto.rs:611:27
++++ actual:   stderr
   1    1 | [FRESH] registry-shared v0.0.1
   2    2 | [FRESH] registry v0.0.1
   3    3 | [COMPILING] bar v0.0.0 ([ROOT]/foo/bar)
   4    4 | [RUNNING] `rustc --crate-name bar [..]-C lto [..]--test [..]`
   5    5 | [RUNNING] `rustc --crate-name b [..]-C lto [..]--test [..]`
   6      - [FINISHED] `release` profile [optimized] target(s) in [ELAPSED]s
   7    6 | [RUNNING] `[ROOT]/foo/target/release/deps/bar-[HASH][EXE]`
   8    7 | [RUNNING] `[ROOT]/foo/target/release/deps/b-[HASH][EXE]`
   9    8 | [DOCTEST] bar
  10    9 | [RUNNING] `rustdoc --edition=2015 --crate-type cdylib --crate-type rlib --crate-name bar --test [..]-C lto [..]
       10 + [FINISHED] `release` profile [optimized] target(s) in 1m 00s

Update with SNAPSHOTS=overwrite
```

PR #14235, opened 2024-07-10
test: migrate check_cfg to snapbox

eth3lbert:

Migrate following to snapbox:

- `tests/testsuite/check_cfg.rs`

---

This commit allow deprecation for most of the tests currently in this file. This is because the tests either use `does_not_contain` or perform multiple `contains` assertion for single tool with macro `x!`.

Here are some potential options to further eliminate the deprecation:

  1. Remove the `x!` macro and replace the it with `str![]` for both Windows and non-Windows platforms.
  2. Modify the `x!` macro to only generate the `--check-cfg` part for different platforms and manually compose it with `format!`.

### DIFF
--- a/tests/testsuite/check_cfg.rs
+++ b/tests/testsuite/check_cfg.rs
@@ -1,8 +1,7 @@
 //! Tests for Cargo usage of rustc `--check-cfg`.
 
-#![allow(deprecated)]
-
-use cargo_test_support::{basic_manifest, project};
+use cargo_test_support::prelude::*;
+use cargo_test_support::{basic_manifest, project, str};
 
 macro_rules! x {
     ($tool:tt => $what:tt $(of $who:tt)?) => {{
@@ -31,6 +30,7 @@ macro_rules! x {
     }};
 }
 
+#[allow(deprecated)]
 #[cargo_test]
 fn features() {
     let p = project()
@@ -53,10 +53,10 @@ fn features() {
     p.cargo("check -v")
         .with_stderr_contains(x!("rustc" => "cfg" of "feature" with "f_a" "f_b"))
         .with_stderr_contains(x!("rustc" => "cfg" of "docsrs"))
-        .with_stderr_does_not_contain("[..]-Zunstable-options[..]")
         .run();
 }
 
+#[allow(deprecated)]
 #[cargo_test]
 fn features_with_deps() {
     let p = project()
@@ -87,6 +87,7 @@ fn features_with_deps() {
         .run();
 }
 
+#[allow(deprecated)]
 #[cargo_test]
 fn features_with_opt_deps() {
     let p = project()
@@ -118,6 +119,7 @@ fn features_with_opt_deps() {
         .run();
 }
 
+#[allow(deprecated)]
 #[cargo_test]
 fn features_with_namespaced_features() {
     let p = project()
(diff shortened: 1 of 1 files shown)


eth3lbert:
Failed in CI with the following:
```
---- expected: tests/testsuite/lto.rs:372:13
++++ actual:   stderr
   1    1 | [COMPILING] foo v0.0.0 ([ROOT]/foo)
   2    2 | [RUNNING] `rustc --crate-name a [..]-C lto [..]`
   3    3 | [RUNNING] `rustc --crate-name b [..]-C lto [..]`
   4    4 | [RUNNING] `rustc --crate-name foo [..]-C lto [..]--test [..]`
   5    5 | [RUNNING] `rustc --crate-name foo [..]--crate-type bin [..]-C lto [..]`
   6      - [FINISHED] `release` profile [optimized] target(s) in [ELAPSED]s
   7    6 | [RUNNING] `[ROOT]/foo/target/release/deps/foo-[HASH][EXE]`
   8    7 | [RUNNING] `[ROOT]/foo/target/release/deps/a-[HASH][EXE]`
   9    8 | [RUNNING] `[ROOT]/foo/target/release/deps/b-[HASH][EXE]`
        9 + [FINISHED] `release` profile [optimized] target(s) in 1m 00s
```

This should be resolved once #14233 is merged.

epage:
#14233 is now merged.

tests/testsuite/check_cfg.rs
```diff
@@ -256,6 +269,7 @@ fn features_test() {
         .run();
 }
 
+#[allow(deprecated)]
 #[cargo_test]
 fn features_doctest() {
```
epage, line 274: I think the `does_not_contains` could be removed.  If we were passing `-Z`, it would fail as this test runs on stable

epage:
> Remove the x! macro and replace the it with str![] for both Windows and non-Windows platforms.

Overall, not a fan of `x!`.

epage:
To add though, we can merge this now-ish (maybe remove some of the weirder `does_not_contains`) and further improve it in follow up PRs, rather than do it all at once.

eth3lbert:
> To add though, we can merge this now-ish (maybe remove some of the weirder `does_not_contains`) and further improve it in follow up PRs, rather than do it all at once.

I remove all the `-Z`. However, if we only want to remove the one within the `features_doctest` function, please let me know!

Urgau:
The `x!` macro was added very early the development of the feature in Cargo and has always been awkward, it was meant as a simple way to create the cmd argument without relying to much on the actual argument; in part because the syntax was still evolving and changing. It also was a convenient way to handle the Windows vs non-Windows escaping.

It has serve us well, but since then many things have evolved, `--check-cfg` is now stable, Cargo has a new testing engine...

In other words, feel free to remove the macro.

eth3lbert:
> The `x!` macro was added very early the development of the feature in Cargo and has always been awkward, it was meant as a simple way to create the cmd argument without relying to much on the actual argument; in part because the syntax was still evolving and changing. It also was a convenient way to handle the Windows vs non-Windows escaping.
> 
> It has serve us well, but since then many things have evolved, `--check-cfg` is now stable, Cargo has a new testing engine...
> 
> In other words, feel free to remove the macro.

Thanks for the explanation and the addition from the author's perspective.

epage:
I'll merge as-is and we can handle `x!` in a follow up

@bors r+

PR #14238, opened 2024-07-11
Add more profiling traces

epage:

Someone suggested I look at
[zed](https://github.com/zed-industries/zed).
It has some interesting performance characteristics compared to projects I looked at before:
- A lot of workspace members
- git dependencies
- patching with git dependencies

This adds traces that help provide more context for those scenarios when looking at them.

image

### Additional information

Some of these areas might be of interest for
optimizing:
- On every invocation, we do a git2 `copy_to` call for each git checkout
- On every invocation, we do a `du` on each git checkout
- It'd be great to avoid parsing every manifest in a git checkout but we need to do a full parse to make sure we find all packages (see https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Redundant.20code.20in.20.60GitSouce.60.3F)
- Its suspicious how much time we spend in the "poison" step of resolving when its a no-op

### DIFF
--- a/src/cargo/ops/resolve.rs
+++ b/src/cargo/ops/resolve.rs
@@ -387,6 +387,7 @@ pub fn resolve_with_previous<'gctx>(
         register_previous_locks(ws, registry, r, &keep, dev_deps);
 
         // Prefer to use anything in the previous lock file, aka we want to have conservative updates.
+        let _span = tracing::span!(tracing::Level::TRACE, "prefer_package_id").entered();
         for id in r.iter().filter(keep) {
             debug!("attempting to prefer {}", id);
             version_prefs.prefer_package_id(id);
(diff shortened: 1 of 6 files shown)


src/cargo/ops/resolve.rs
```diff
@@ -387,6 +387,8 @@ pub fn resolve_with_previous<'gctx>(
         register_previous_locks(ws, registry, r, &keep, dev_deps);
 
         // Prefer to use anything in the previous lock file, aka we want to have conservative updates.
+        let span = tracing::span!(tracing::Level::TRACE, "prefer_package_id");
```
arlosi, line 390: You can use [`entered()`](https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.entered) here and avoid the extra `_enter` variable below.

```suggestion
        let _span = tracing::span!(tracing::Level::TRACE, "prefer_package_id").entered();
```

epage: I like how none of the examples reference that

arlosi:
Thanks

@bors r+

Issue #13724, opened 2024-04-08
build error in example code since 0d62ae2


ckoehler:
### Problem

We're seeing the following error during a `cargo build`: 

```
error: invalid character `{` in package name: `{{package_name}}`, the first character must be a Unicode XID start character (most letters or `_`)
 --> ../../.cargo/git/checkouts/my-crate-e3ad1ac7415c84bc/3314343/tools/generator/thing/thing/{{thing_name}}/Cargo.toml:2:8
  |
2 | name = "{{package_name}}"
  |        ^^^^^^^^^^^^^^
```

This started with Rust 1.77, and I bisected it to commit 0d62ae2. That one looks to really just show errors differently, so I don't know if it caused our error or just surfaced it. The build is not affected, because it's just the `cargo-generator` code for some of our stuff. That path not in any of our workspace manifests, I think cargo just finds it by `walk`ing the source files of our crate and finds the templated Cargo.toml.

### Possible Solution(s)

We can move the templated stuff out of our repo and somewhere else, but since it's a change in behavior, I wanted to flag it.

### Version

```text
cargo 1.77.1 (e52e36006 2024-03-26)
release: 1.77.1
commit-hash: e52e360061cacbbeac79f7f1215a7a90b6f08442
commit-date: 2024-03-26
host: aarch64-apple-darwin
libgit2: 1.7.2 (sys:0.18.2 vendored)
libcurl: 8.4.0 (sys:0.4.70+curl-8.5.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Mac OS 14.2.1 [64-bit]
```

Muscraft:
Can you share the repo where this happened or create an example? It would help in debugging what is going on.

ckoehler:
Sorry, unfortunately it's a private corp repo :/ I may be able to do an example, tho I am not quite sure what all causes this...

ehuss:
@Muscraft Here is a test demonstrating it:

```rust
#[cargo_test]
fn invalid_name_in_git_repo() {
    // Checks to ignore invalid package name in git repository.
    let git_project = git::new("bar", |p| {
        p.file("Cargo.toml", &basic_manifest("bar", "1.0.0"))
            .file("src/lib.rs", "")
            .file(
                "dont_look_at_me/Cargo.toml",
                r#"
                    [package]
                    name = "{{project_name}}"
                    version = "1.0.0"
[...]

PR #14239, opened 2024-07-12
fix(source): Don't warn about unreferenced duplicate packages

### LINKED ISSUE #10752: warning: skipping duplicate package `case` found at `~\.cargo\git\checkouts\cargo-..`

epage:

This also improves the message, consolidating multiple duplicates and saying which was loaded instead, as it naturally fell out of the design

Fixes #10752

### Additional information

We're still subject to #13724 and fully load every manifest, even if we don't use it.  I'm exploring that topic at https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Redundant.20code.20in.20.60GitSouce.60.3F/near/450783427

This change builds on
- #13993
- #14169
- #14231
- #14234

### DIFF
--- a/src/cargo/sources/path.rs
+++ b/src/cargo/sources/path.rs
@@ -226,7 +226,9 @@ pub struct RecursivePathSource<'gctx> {
     /// Whether this source has loaded all package information it may contain.
     loaded: bool,
     /// Packages that this sources has discovered.
-    packages: HashMap<PackageId, Package>,
+    ///
+    /// Tracking all packages for a given ID to warn on-demand for unused packages
+    packages: HashMap<PackageId, Vec<Package>>,
     gctx: &'gctx GlobalContext,
 }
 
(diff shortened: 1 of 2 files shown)


src/cargo/sources/path.rs
```diff
@@ -226,7 +226,8 @@ pub struct RecursivePathSource<'gctx> {
     /// Whether this source has loaded all package information it may contain.
     loaded: bool,
     /// Packages that this sources has discovered.
-    packages: HashMap<PackageId, Package>,
+    packages: HashMap<PackageId, Vec<Package>>,
```
weihanglo, line 229: Could we document what the key-value pair for? I can see it a bit unintuitive because `RecurisivePathSource` ignores all but the first.

weihanglo (approved):
Thanks for the fix.

Review comments added are all minor. `r=weihanglo` when you feel ready.

Issue #353, opened 2014-08-09
Allow in-crate tests and doc tests to be disabled separately


sfackler:
Several of my projects use out-of-crate tests, which means I end up building and running an extra, large test binary with no tests in it.

PR #14242, opened 2024-07-12
test: migrate messages to snapbox

eth3lbert:

Migrate following to snapbox:

- `tests/testsuite/messages.rs`

### DIFF
--- a/tests/testsuite/messages.rs
+++ b/tests/testsuite/messages.rs
@@ -2,8 +2,7 @@
 //!
 //! Tests for message caching can be found in `cache_messages`.
 
-#![allow(deprecated)]
-
+use cargo_test_support::compare::assert_e2e;
 use cargo_test_support::prelude::*;
 use cargo_test_support::{process, project, Project};
 use cargo_util::ProcessError;
@@ -48,6 +47,14 @@ pub fn raw_rustc_output(project: &Project, path: &str, extra: &[&str]) -> String
     result
 }
 
+fn redact_rustc_message(msg: &str) -> impl IntoData {
+    use snapbox::filter::{Filter, FilterPaths};
+    let assert = assert_e2e();
+    let redactions = assert.redactions();
+    let msg = redactions.redact(msg);
+    FilterPaths.filter(msg.into())
+}
+
 #[cargo_test]
 fn deduplicate_messages_basic() {
     let p = project()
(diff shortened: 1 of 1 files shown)


tests/testsuite/messages.rs
```diff
@@ -47,6 +45,13 @@ pub fn raw_rustc_output(project: &Project, path: &str, extra: &[&str]) -> String
     result
 }
 
+fn redact_rustc_message(msg: &str) -> String {
+    msg.replace("help:", "[HELP]")
+        .replace("note:", "[NOTE]")
+        .replace("warning:", "[WARNING]")
+        .replace("error:", "[ERROR]")
+}
```
eth3lbert, line 53: I'm not sure if removing the `raw_rustc_output` function and relying solely on snapshots is a viable option. Therefore, I've currently redacted the messages with it. However, even if removing `raw_rustc_output` is acceptable, there's another file, `cache_messages`, that depends on `raw_rustc_output`. Therefore, I believe it would be better to refactor it in a follow-up pull request.

tests/testsuite/messages.rs
```diff
@@ -48,6 +46,20 @@ pub fn raw_rustc_output(project: &Project, path: &str, extra: &[&str]) -> String
...
+fn redact_rustc_message(msg: &str) -> String {
+    let mut msg = msg
+        .replace("help:", "[HELP]")
+        .replace("note:", "[NOTE]")
+        .replace("warning:", "[WARNING]")
+        .replace("error:", "[ERROR]");
+
+    if cfg!(windows) {
+        msg = msg.replace("\\", "/");
+    }
```
epage, line 58: We do this for all platforms, not just windows

`snapbox` provides a `Filter` for this you can reuse

tests/testsuite/messages.rs
```diff
@@ -48,6 +46,20 @@ pub fn raw_rustc_output(project: &Project, path: &str, extra: &[&str]) -> String
     result
 }
 
+fn redact_rustc_message(msg: &str) -> String {
+    let mut msg = msg
+        .replace("help:", "[HELP]")
+        .replace("note:", "[NOTE]")
+        .replace("warning:", "[WARNING]")
+        .replace("error:", "[ERROR]");
```
epage, line 54: I'd rather we provide a way to expose these redactions and run them on `msg`.  For example, we could add a getter to `snapbox::Assert`

tests/testsuite/messages.rs
```diff
@@ -48,6 +46,20 @@ pub fn raw_rustc_output(project: &Project, path: &str, extra: &[&str]) -> String
     result
 }
 
+fn redact_rustc_message(msg: &str) -> String {
```
epage, line 49: I think a core part of the problem is that we don't redact expected values.  Maybe we should revisit that?

tests/testsuite/messages.rs
dieterplex: I found that the slash replacement need  `[Data::is_path_set`](https://github.com/assert-rs/snapbox/blob/main/crates%2Fsnapbox%2Fsrc%2Fassert%2Fmod.rs#L124-L124) in condition to filter but no way for Data::filters to set `paths()` using test support crate. Or I misunderstood what you suggested?

tests/testsuite/messages.rs
eth3lbert: Indeed! I think providing a function to normalize/redact the expected value manually or a toggle to automatically normalize expected value would be more flexible and less troublesome than always normalizing, as there might be cases where we don't want this behavior.

epage: #353

eth3lbert: Ah, I think you're referring to https://github.com/assert-rs/snapbox/pull/353?

tests/testsuite/messages.rs
```diff
@@ -48,6 +47,14 @@ pub fn raw_rustc_output(project: &Project, path: &str, extra: &[&str]) -> String
     result
 }
 
+fn redact_rustc_message(msg: &str) -> impl IntoData {
+    use snapbox::filter::{Filter, FilterPaths};
+    let assert = assert_e2e();
+    let redactions = assert.redactions();
+    let msg = redactions.redact(msg);
+    FilterPaths.filter(msg.into())
+}
```
eth3lbert, line 56: And of course, we can just rewrite the assertions in this file using `assert_e2e` if needed.

PR #14243, opened 2024-07-12
feat(test): Add cargo_test to test-support prelude

epage:

With where we are at with #14039, I was revisiting our test writing documentation.  I was wanting to put more emphasis on rustdoc and found it would be helpful to talk about the concepts if a prelude was used instead of `extern crate`.

Yes, this included some `use` clean ups in tangentially-related commit but its already painful enough
walking through every test file, I didn't want to do it twice.


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #14244, opened 2024-07-12
Migrate global_cache_tracker snapbox

dieterplex:
Part of #14039.

Some of these tests may require to assert specific removed file count or size that would be redact by current design. Need feedback to come out solution.

Temporary commit to pass the tests and illustrate what need to fix.

### DIFF
--- a/crates/cargo-test-support/src/compare.rs
+++ b/crates/cargo-test-support/src/compare.rs
@@ -188,7 +188,7 @@ fn add_regex_redactions(subs: &mut snapbox::Redactions) {
     .unwrap();
     subs.insert(
         "[HASH]",
-        regex!(r"home/\.cargo/registry/src/-(?<redacted>[a-z0-9]+)"),
+        regex!(r"home/\.cargo/registry/(cache|index|src)/-(?<redacted>[a-z0-9]+)"),
     )
     .unwrap();
     subs.insert(
--- a/tests/testsuite/global_cache_tracker.rs
+++ b/tests/testsuite/global_cache_tracker.rs
@@ -6,8 +6,6 @@
 //! what happens when time passes. The [`days_ago_unix`] and
 //! [`months_ago_unix`] functions help with setting this value.
 
-#![allow(deprecated)]
-
 use std::env;
 use std::fmt::Write;
 use std::path::Path;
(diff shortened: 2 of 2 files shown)


epage:
> Temporary commit to pass the tests and illustrate what need to fix.

How about we skip those specific tests for now and focus on getting the rest of this in?

dieterplex:
> > Temporary commit to pass the tests and illustrate what need to fix.
> 
> How about we skip those specific tests for now and focus on getting the rest of this in?

Added `#[allow(deprecated)]` to these tests.

tests/testsuite/global_cache_tracker.rs
```diff
@@ -778,10 +778,24 @@ fn clean_gc_dry_run() {
...
+        let p = expected_files.replace(&root, "[ROOT]");
+        let p = if cfg!(windows) {
+            p.replace("\\", "/")
+        } else {
+            p
+        };
+        let p = re_crate.replace_all(&p, "-[HASH]");
+        let p = re_pack.replace_all(&p, "pack-[HASH]");
+        format!("{p}\n")
+    };
```
epage, line 794: Let's skip this one

dieterplex: done

dieterplex: Note for similar case https://github.com/rust-lang/cargo/pull/14096#discussion_r1646855403

crates/cargo-test-support/src/compare.rs
```diff
@@ -196,7 +196,12 @@ fn add_regex_redactions(subs: &mut snapbox::Redactions) {
         regex!(r"\.cargo/target/(?<redacted>[0-9a-f]{2}/[0-9a-f]{14})"),
     )
     .unwrap();
-    subs.insert("[HASH]", regex!(r"/[a-z0-9\-_]+-(?<redacted>[0-9a-f]{16})"))
+    subs.insert(
+        "[HASH]",
+        regex!(r"/[a-z0-9\-_]+-(?<redacted>[0-9a-f]{16})\b"),
+    )
+    .unwrap();
+    subs.insert("[HASH]", regex!(r"/pack-(?<redacted>[0-9a-f]{40})\."))
```
epage, line 204: This commit is last. Are tests passing in each commit and this commit is unused?

dieterplex: Yes, it come with the skipped change to redact crate name and git pack filename correctly. Will dispose it.

Issue #14178, opened 2024-07-02
`cargo update --breaking`  doesn't respect pre release


thewh1teagle:
### Problem

Hello!
I use `cargo update` from `nightly` release to upgrade dependencies
However looks like `cargo` doesn't respect beta semver:

```console
cargo +nightly -Zunstable-options update --breaking
    Updating crates.io index
   Upgrading env_logger ^0.10.1 -> ^0.11.3
   Upgrading tauri ^2.0.0-beta.21 -> ^1.7.0
   Upgrading tauri-plugin-deep-link ^2.0.0-beta.5 -> ^0.1.2
   Upgrading windows ^0.56.0 -> ^0.57.0
   Upgrading reqwest ^0.11.23 -> ^0.12.5
    Updating crates.io index
error: failed to select a version for `webkit2gtk-sys`.
```

Notice the line

```console
 Upgrading tauri ^2.0.0-beta.21 -> ^1.7.0
```

### Steps

1. Install tauri beta release
2. upgrade it 

```console
cd /tmp
cargo init --bin sample
cd sample
cargo add tauri@=2.0.0-beta.22
cargo +nightly -Zunstable-options update --breaking
```

Then compare with https://crates.io/crates/tauri/versions

### Possible Solution(s)

Respect beta versions

### Notes

Also, I wanted to say that I'm really looking forward to seeing this feature, as I'm using cargo upgrade from `cargo-edit`, which takes tremendous time to upgrade dependencies, while cargo update from nightly does the same job much quicker.

### Version

```console
$ cargo -Vv
cargo 1.75.0 (1d8b05cdd 2023-11-20)
release: 1.75.0
commit-hash: 1d8b05cdd1287c64467306cf3ca2c8ac60c11eb0
commit-date: 2023-11-20
host: aarch64-apple-darwin
libgit2: 1.7.1 (sys:0.18.1 vendored)
libcurl: 8.6.0 (sys:0.4.68+curl-8.4.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1u  30 May 2023
os: Mac OS 14.5.0 [64-bit]

$ cargo +nightly -Vv                                 
cargo 1.81.0-nightly (4ed7bee47 2024-06-25)
release: 1.81.0-nightly
commit-hash: 4ed7bee47f7dd4416b36fada1909e9a62c546246
commit-date: 2024-06-25
host: aarch64-apple-darwin
[...]

PR #14250, opened 2024-07-15
Don't downgrade on prerelease `VersionReq` when update with --breaking.

linyihai:
Do nothing with prerelease when update with `--breaking`. 

Fixes https://github.com/rust-lang/cargo/issues/14178

Previous commit add a test, next commit to fix and update.

### DIFF
--- a/src/cargo/util/toml_mut/upgrade.rs
+++ b/src/cargo/util/toml_mut/upgrade.rs
@@ -217,5 +217,10 @@ mod test {
             assert_req_bump("1.1.1", "=1.0.0", "=1.1.1");
             assert_req_bump("2.0.0", "=1.0.0", "=2.0.0");
         }
+
+        #[test]
+        fn caret_prerelease() {
+            assert_req_bump("1.7.0", "2.0.0-beta.21", "1.7.0");
+        }
     }
 }
--- a/src/cargo/util/toml_mut/upgrade.rs
+++ b/src/cargo/util/toml_mut/upgrade.rs
@@ -15,12 +15,16 @@ pub(crate) fn upgrade_requirement(
         // Empty matches everything, no-change.
         Ok(None)
     } else {
-        let comparators: CargoResult<Vec<_>> = raw_req
+        let comparators: Vec<_> = raw_req
             .comparators
             .into_iter()
+            // Don't downgrade if pre-release was used, see https://github.com/rust-lang/cargo/issues/14178 and https://github.com/rust-lang/cargo/issues/13290.
+            .filter(|p| p.pre.is_empty() || matches_greater(p, version))
             .map(|p| set_comparator(p, version))
-            .collect();
-        let comparators = comparators?;
+            .collect::<CargoResult<_>>()?;
+        if comparators.is_empty() {
+            return Ok(None);
+        }
         let new_req = semver::VersionReq { comparators };
         let mut new_req_text = new_req.to_string();
         if new_req_text.starts_with('^') && !req.starts_with('^') {
@@ -72,6 +76,33 @@ fn set_comparator(
             Err(unsupported_version_req(user_pred))
         }
     }
+}
+
+// See https://github.com/dtolnay/semver/blob/69efd3cc770ead273a06ad1788477b3092996d29/src/eval.rs#L64-L88
+fn matches_greater(cmp: &semver::Comparator, ver: &semver::Version) -> bool {
+    if ver.major != cmp.major {
+        return ver.major > cmp.major;
+    }
+
+    match cmp.minor {
+        None => return false,
+        Some(minor) => {
+            if ver.minor != minor {
+                return ver.minor > minor;
+            }
+        }
+    }
+
+    match cmp.patch {
+        None => return false,
+        Some(patch) => {
+            if ver.patch != patch {
+                return ver.patch > patch;
+            }
+        }
+    }
+
+    ver.pre > cmp.pre
 }
 
 fn assign_partial_req(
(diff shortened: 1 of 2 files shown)


src/cargo/util/toml_mut/upgrade.rs
```diff
@@ -217,5 +223,10 @@ mod test {
             assert_req_bump("1.1.1", "=1.0.0", "=1.1.1");
             assert_req_bump("2.0.0", "=1.0.0", "=2.0.0");
         }
+
+        #[test]
+        fn caret_prerelease() {
+            assert_req_bump("1.7.0", "2.0.0-beta.21", None);
```
weihanglo, line 229: Could we test against other SemVer operators? Like `assert_req_bump("1.7.0", ">=2.0.0-beta.21", None)`. We might want to ignore this one instead of panicking. Thought?

src/cargo/util/toml_mut/upgrade.rs
```diff
@@ -77,7 +80,10 @@ fn set_comparator(
 fn assign_partial_req(
     version: &semver::Version,
     mut pred: semver::Comparator,
-) -> semver::Comparator {
+) -> Option<semver::Comparator> {
+    if !pred.pre.is_empty() {
```
weihanglo, line 84: Better to have common for the meaning of `Option` here or on `set_comparator`.

linyihai:
wait a second, I squashed the two commit into one accidentally, let me split them.

src/cargo/util/toml_mut/upgrade.rs
linyihai: Ok, I had added other Semver operators test but in the renamed test `ignore_prerelease`

src/cargo/util/toml_mut/upgrade.rs
linyihai: That's a good idea. I moved the changes up to the `upgrade_requirement` to avoid fixing them here

linyihai:
```
---- expected: tests/testsuite/https.rs:159:27
++++ actual:   stderr
   1    1 | [UPDATING] git repository `[https://github.com/rust-lang/bitflags.git`](https://github.com/rust-lang/bitflags.git%60)
        2 + [WARNING] spurious network error (3 tries remaining): failed to connect to github.com: Connection timed out; class=Os (2)
   2    3 | [LOCKING] 2 packages to latest compatible versions

```

May be supcious network connection error?

src/cargo/util/toml_mut/upgrade.rs
weihanglo: My fault. I think I have missed this <https://github.com/rust-lang/cargo/issues/14178#issuecomment-2203083535>. Let's try clarifying it :)

@epage Did you suggest allowing prerelease upgrade under the current stable semver semantic? Should we instead ignore prerelease and document the behavior, until we settle the semantic in `cargo update --precise <prerelease>` feature?

linyihai: https://github.com/rust-lang/cargo/blob/a2b58c3dad4d554ba01ed6c45c41ff85390560f2/src/cargo/ops/cargo_update.rs#L330-L336

The `update --breaking` had filtered the `prerelease` version to update to. So here equals to no upgrade to prerelease also. :/

epage:
CC @torhovland

I feel like one of your PRs had touched on this already but we wanted to split it out for focusing the reviews,  or am I thinking of something else?

torhovland:
> CC @torhovland
> 
> I feel like one of your PRs had touched on this already but we wanted to split it out for focusing the reviews, or am I thinking of something else?

I think these are separate issues. This one here is about `update --breaking` not respecting that the user already has a higher pre-release installed, thus trying to "upgrade" to a lower version.

The one I have encountered is about a precise version req not matching its exact version if it is a pre-release. See https://github.com/rust-lang/cargo/pull/14140/commits/de1a4a2512cac7d6e8e698fc0f3637893543d708

linyihai:
I copied the  `matches_greater` from [`semver`](https://github.com/dtolnay/semver/blob/master/src/eval.rs#L64) to make sure prerelease VersionReq won't downgrade.

tests/testsuite/update.rs
```diff
@@ -2616,3 +2616,37 @@ fn update_breaking_mixed_pinning_renaming() {
...
+        [dependencies]
+        bar = "2.0.0-beta.21"
+    "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("generate-lockfile").run();
+
+    Package::new("bar", "1.7.0").publish();
```
torhovland, line 2643: A comment would be useful here.

```
// The purpose of this test is to demonstrate that `update --breaking` will not try to downgrade to the latest stable version (1.7.0), but will rather keep the latest pre-release (2.0.0-beta.21).
```

linyihai: Good idea. Have already added.

src/cargo/util/toml_mut/upgrade.rs
```diff
@@ -18,9 +18,18 @@ pub(crate) fn upgrade_requirement(
...
             .comparators
             .into_iter()
-            .map(|p| set_comparator(p, version))
+            .filter_map(|p| {
+                if p.pre.is_empty() {
+                    Some(set_comparator(p, version))
+                } else {
+                    None
+                }
+            })
```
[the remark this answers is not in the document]
linyihai: Done, thanks

tests/testsuite/update.rs
```diff
@@ -2616,3 +2616,39 @@ fn update_breaking_mixed_pinning_renaming() {
             "#]],
     );
 }
+
+#[cargo_test]
+fn update_breaking_pre_release() {
```
weihanglo, line 2621: From <https://github.com/rust-lang/cargo/pull/14250/files#r1680348122>:

> The `update --breaking` had filtered the `prerelease` version to update to. So here equals to no upgrade to prerelease also. :/

Since that is the case, we might want to have UI tests for

1. `2.0.0-beta.22` is available, but `2.0.0-beta.21` won't upgrade to it.
2. `2.0.0` is available, but `2.0.0-beta.21` won't upgrade to it.
2. `3.0.0` is available, and `2.0.0-beta.21` will upgrade to it.

We should document the first two as an unresolved issue. That is something we want to support, especially the second one.

linyihai: Added `update_breaking_pre_release_upgrade` to reflect the current behavior.

> 2.0.0 is available, but 2.0.0-beta.21 won't upgrade to it.

This case out of my expection, it seems need to support.

weihanglo (approved):

weihanglo:
While there are still some edge cases having to be supported, we shouldn't block this from fixing wrong downgrades. Thank you for the contribution!

@bors r+

tests/testsuite/update.rs
weihanglo: Task added: <https://github.com/rust-lang/cargo/issues/12425#issuecomment-2186198258>.

PR #14259, opened 2024-07-16
Improved consistency between breaking and non-breaking updates

torhovland:
Related to #12425.

This is ensuring more consistent output messages between breaking and non-breaking updates, as well as better error messages, as we are now running the same `ops::update_lockfile` in either case. 

See how the the tests changed for examples. This becomes even more important after adding support for breaking precise updates in #14140, and this PR is necessary to unblock that PR. Here's why:

- Some of the duplicated existing precise tests with the unstable feature enabled are not passing without this PR. In other words, we would be changing the behaviour of the non-breaking precise update.

- Some invocations are silently finishing rather than reporting an error such as this:

```
Did you mean one of these?

  incompatible@0.3.1
```

- The output is generally not consistent with the non-breaking update. For example, without this PR the output changes like this:

```
-[UPDATING] incompatible v0.1.0 -> v0.2.0
-[NOTE] pass `--verbose` to see 1 unchanged dependencies behind latest
+[UPDATING] incompatible v0.1.0 -> v0.2.0 (latest: v0.2.1)
```

Requested as a separate PR [here](https://github.com/rust-lang/cargo/pull/14140#issuecomment-2226299518) and [here](https://github.com/rust-lang/cargo/pull/14140#discussion_r1676447754). 

This PR does not introduce an error message if there is nothing to update. Instead, it adds more output, as demonstrated by the test `update_breaking_specific_packages_that_wont_update`:

```
[LOCKING] 0 packages to latest compatible versions
[NOTE] pass `--verbose` to see 5 unchanged dependencies behind latest
```

Whether or not to throw an error is still an item on the TODO list [here](https://github.com/rust-lang/cargo/issues/12425#issuecomment-2186198258).

### DIFF
--- a/src/cargo/ops/cargo_update.rs
+++ b/src/cargo/ops/cargo_update.rs
@@ -9,6 +9,7 @@ use crate::ops;
 use crate::sources::source::QueryKind;
 use crate::util::cache_lock::CacheLockMode;
 use crate::util::context::GlobalContext;
+use crate::util::interning::InternedString;
 use crate::util::toml_mut::dependency::{MaybeWorkspace, Source};
 use crate::util::toml_mut::manifest::LocalManifest;
 use crate::util::toml_mut::upgrade::upgrade_requirement;
@@ -20,12 +21,25 @@ use std::cmp::Ordering;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use tracing::{debug, trace};
 
-pub type UpgradeMap = HashMap<(String, SourceId), Version>;
+/// A map of all breaking upgrades which is filled in by
+/// upgrade_manifests/upgrade_dependency when going through workspace member
+/// manifests, and later used by write_manifest_upgrades in order to know which
+/// upgrades to write to manifest files on disk. Also used by update_lockfile to
+/// know which dependencies to keep unchanged if any have been upgraded (we will
+/// do either breaking or non-breaking updates, but not both).
+pub type UpgradeMap = HashMap<
+    // The key is a package identifier consisting of the name and the source id.
+    (InternedString, SourceId),
+    // The value is the original version requirement before upgrade, and the
+    // upgraded version.
+    (VersionReq, Version),
+>;
 
 pub struct UpdateOptions<'a> {
     pub gctx: &'a GlobalContext,
     pub to_update: Vec<String>,
     pub precise: Option<&'a str>,
+    pub breaking: bool,
     pub recursive: bool,
     pub dry_run: bool,
     pub workspace: bool,
@@ -49,7 +63,11 @@ pub fn generate_lockfile(ws: &Workspace<'_>) -> CargoResult<()> {
     Ok(())
 }
 
-pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoResult<()> {
+pub fn update_lockfile(
+    ws: &Workspace<'_>,
+    opts: &UpdateOptions<'_>,
+    upgrades: &UpgradeMap,
+) -> CargoResult<()> {
     if opts.recursive && opts.precise.is_some() {
         anyhow::bail!("cannot specify both recursive and precise simultaneously")
     }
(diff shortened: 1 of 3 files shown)


src/cargo/ops/cargo_update.rs
```diff
@@ -165,7 +182,13 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
         .filter(|s| !s.is_registry())
         .collect();
 
-    let keep = |p: &PackageId| !to_avoid_sources.contains(&p.source_id()) && !to_avoid.contains(p);
+    let keep = |p: &PackageId| {
+        (!to_avoid_sources.contains(&p.source_id()) && !to_avoid.contains(p))
+                // In case of `--breaking` (or precise upgrades, coming in
+                // #14140), we want to keep all packages unchanged that didn't
+                // get upgraded.
+                    || (opts.breaking && !upgrades.contains_key(&(p.name(), p.source_id())))
```
torhovland, line 190: This is an important change that has been discussed here: https://github.com/rust-lang/cargo/pull/14140#discussion_r1676447754

epage: To put another way, there is concern with using `keep` like this and resolving this is a blocker for moving forward.

src/cargo/ops/cargo_update.rs
```diff
@@ -49,7 +62,11 @@ pub fn generate_lockfile(ws: &Workspace<'_>) -> CargoResult<()> {
     Ok(())
 }
 
-pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoResult<()> {
+pub fn update_lockfile(
```
epage, line 65: > This is ensuring more consistent output messages between breaking and non-breaking updates, as well as better error messages, as we are now running the same ops::update_lockfile in either case.

Reusing `ops::update_lockfile` is one way of improving these.  Are there alternative designs we could use?

epage:
> Requested as a separate PR https://github.com/rust-lang/cargo/pull/14140#issuecomment-2226299518 and https://github.com/rust-lang/cargo/pull/14140#discussion_r1676447754.

Quoting from the first link:

> > Except this PR description does not include an explanation as to why this PR addresses that and why an error is not the way forward
> > Description updated.
> 
> The PR says:
> 
> > This PR is making a change that also affects cargo update --breaking. We'll reuse ops::update_lockfile() for both breaking and non-breaking updates. The benefit is more consistent output and behaviour between the two. In particular, it addresses a task about an error output if there is nothing to upgrade. See [#12425 (comment)](https://github.com/rust-lang/cargo/issues/12425#issuecomment-2186198258).
> 
> This is focused on the implementation, glosses over the reason, and doesn't say what is being changed.
> 
> I think its best to split this out from this PR, either making ti independent or making one depend on the other. Trying to slip it in like this is making it harder to review and discuss both.

If this PR is related to this comment, its not clear.  It doesn't talk about about the error case mentioned in that comment or give a justification for why this is the approach to go.

src/cargo/ops/cargo_update.rs
torhovland: I am curious what the actual concern is, though. We are only modifying `keep` in the case of breaking or precise updates, and we have quite extensive test coverage of those.

Eh2406: > The reason this is not put in to_avoid is that those require PackageIds, i.e. they require a version. When doing upgrades, we don't have any versions, only version requirements.

In general converting from a requirement to a version is not possible and converting to a set of versions requires scanning  all the versions on the index. But in this case `keep`  is only called on versions that are in the lock file. I think we already scan all the versions in the lock file to construct `to_avoid`, so I'd rather see this check  done there.

torhovland: I have changed this PR so it no longer changes `keep`, but instead builds up what it needs in `to_avoid`. Please review!

torhovland:
> If this PR is related to this comment, its not clear. It doesn't talk about about the error case mentioned in that comment or give a justification for why this is the approach to go.

I have updated the PR description with more context. Not sure what else to add. If you think it needs more, let me know.

epage:
> See how the the tests changed for examples.

That is what I am wanting you to talk about.  This is supposed to change behavior related to one of the tasks in a direction that I have serious doubs about.  I'd like the intended behavior changes enumerated and why those behaviors were selected, not an explanation for what this unblocks.

src/cargo/ops/cargo_update.rs
```diff
@@ -91,7 +109,38 @@ pub fn update_lockfile(ws: &Workspace<'_>, opts: &UpdateOptions<'_>) -> CargoRes
...
+        for ((name, source_id), (version_req, _)) in upgrades.iter() {
+            if let Some(matching_dep) = previous_resolve.iter().find(|dep| {
+                dep.name() == *name
+                    && dep.source_id() == *source_id
+                    && version_req.matches(dep.version())
+            }) {
+                let spec = PackageIdSpec::new(name.to_string())
+                    .with_url(source_id.url().clone())
+                    .with_version(matching_dep.version().clone().into());
+                let spec = format!("{spec}");
```
epage, line 133: isn't that just `spec.to_string()`?

torhovland:
> If this PR is related to this comment, its not clear. It doesn't talk about about the error case mentioned in that comment or give a justification for why this is the approach to go.

I have added to the description something about whether or not to error if there is nothing to do.

torhovland:
> > See how the the tests changed for examples.
> 
> That is what I am wanting you to talk about. This is supposed to change behavior related to one of the tasks in a direction that I have serious doubs about. I'd like the intended behavior changes enumerated and why those behaviors were selected, not an explanation for what this unblocks.

This PR isn't really related to the task list at all. The only purpose of this PR is to factor out any change to `to_avoid` or `keep` in `fn update_lockfile`.

src/cargo/ops/cargo_update.rs

torhovland:
@epage Could you let me know if you have any further concerns about this PR? #14140 is blocked on this.

torhovland:
@epage Could you take a look at https://github.com/rust-lang/cargo/pull/14140#discussion_r1719691926, please? I believe it will unblock this PR as well as #14140.

PR #14260, opened 2024-07-16
make summary sync by using Arc not Rc

Eh2406:

For my PubGrub testing work I want to be able to read the entire crates.io index into memory and then run lots of resolution questions against that data in parallel. Currently cargoes `Summary` and `Dependency` types use `Rc` internally which prevents this pattern. Using `Arc` in cargo makes my life a lot easier! It does not noticeably slow down single threaded performance. (Measured by running my PubGrub testing in single threaded mode before and after.)

The team largely agreed to this at a meeting and in discussions https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Make.20summary.20sync

Tests still pass

### Additional information

Should be add a test a this is sync?

### DIFF
--- a/src/cargo/core/dependency.rs
+++ b/src/cargo/core/dependency.rs
@@ -5,7 +5,7 @@ use serde::Serialize;
 use std::borrow::Cow;
 use std::fmt;
 use std::path::PathBuf;
-use std::rc::Rc;
+use std::sync::Arc;
 use tracing::trace;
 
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
> Should be add a test a this is sync?

Yes please.

weihanglo (approved):

weihanglo:
Nice!

@bors r+

Issue #13953, opened 2024-05-23
The output from `cargo package --list` is not good for workspaces


torhovland:
### Problem

This issue was discovered when working on #13947, but the problem is also on `master`.

When run in a workspace root directory, `cargo package --list` outputs something like this:

```
Cargo.lock
Cargo.toml
Cargo.toml.orig
src/main.rs
Cargo.toml
Cargo.toml.orig
src/lib.rs
```

It doesn't indicate that half the files are from one crate, and the other half from another crate.

The question is how to improve this in a way that is backward compatible. It is likely that the current output is used by CI/CD scripts, and should ideally stay as it is when packaging single crates.

### Version

```text
cargo 1.78.0 (54d8815d0 2024-03-26)
release: 1.78.0
commit-hash: 54d8815d04fa3816edc207bbc4dd36bf18014dbc
commit-date: 2024-03-26
host: aarch64-apple-darwin
libgit2: 1.7.2 (sys:0.18.2 vendored)
libcurl: 8.4.0 (sys:0.4.72+curl-8.6.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Mac OS 14.2.1 [64-bit]
```

epage:
This is likely broken enough that we can change behavior.  I wonder if we should just error in this case.  If we had json output, then I could see specializing the list per package.

epage:
See #11666 for json output

torhovland:
One approach that is probably better than an error message would be this:

- In the case of one crate, keep the output as it is.
- In the case of multiple crates, output the name or full path of each crate followed by its file contents. Let each crate be separated by a blank line.

epage:
I think a lot depends on what we consider the output intended to be.  The current single-crate output is used programmatically.  Extending the format with a per-crate header extends what we are offering.  This is why I was looking to crate output.

PR #14262, opened 2024-07-17
Add a JSON output mode to `cargo package`.

### LINKED ISSUE #11666: Provide a json output format of `cargo package --list`

LawnGnome:
In multi-crate workspaces, it can be ambiguous where files are sourced from when building packages. This extends the existing `--list` command line flag to optionally emit JSON when called as `-l=json` or `--list=json`. Existing behaviour is unaffected.

The JSON format is slightly different to that proposed in #11666: it exposes the distinction within `cargo::ops::cargo_package` itself around whether an archive file is generated or sourced from the disk. A new crate created with `cargo new --lib foo` generates this output:

```json
{
  ".cargo_vcs_info.json": {
    "Generated": {
      "VcsInfo": {
        "git": {
          "sha1": "af0bf5917934ec4684cd23763f90c16242d7c5ce"
        },
        "path_in_vcs": ""
      }
    }
  },
  ".gitignore": {
    "OnDisk": "/tmp/foo/.gitignore"
  },
  "Cargo.toml": {
    "Generated": "Manifest"
  },
  "Cargo.toml.orig": {
    "OnDisk": "/tmp/foo/Cargo.toml"
  },
  "src/lib.rs": {
    "OnDisk": "/tmp/foo/src/lib.rs"
  }
}
```

I'm open minded as to whether the `VcsInfo` should actually be included, but it feels like there's little harm in doing so.

Fixes #11666, and is at least extremely relevant to #13953.

### DIFF
--- a/src/bin/cargo/commands/package.rs
+++ b/src/bin/cargo/commands/package.rs
@@ -1,16 +1,22 @@
 use crate::command_prelude::*;
 
-use cargo::ops::{self, PackageOpts};
+use cargo::ops::{self, ListMode, PackageOpts};
+use clap::builder::PossibleValuesParser;
 
 pub fn cli() -> Command {
     subcommand("package")
         .about("Assemble the local package into a distributable tarball")
         .arg(
-            flag(
+            opt(
                 "list",
                 "Print files included in a package without making one",
             )
-            .short('l'),
+            .short('l')
+            .default_missing_value("basic")
+            .num_args(0..=1)
+            .require_equals(true)
+            .value_name("MODE")
+            .value_parser(PossibleValuesParser::new(["basic", "json"])),
         )
         .arg(flag(
             "no-verify",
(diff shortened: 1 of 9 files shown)


epage:
FYI I'm moving this to Draft as this has [not been `accepted` or `needs-mentor`](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo) yet. 

We generally reserve PRs for discussing the implementation and leaving problem, requirements, and solution discussions to the issues.  I'd recommend furthering the discussion on the Issue.

weihanglo:
Close in favor of <https://github.com/rust-lang/cargo/pull/15311>. Thank you!

PR #14266, opened 2024-07-18
fix(test): Move path2url to CargoPathExt::to_url 

epage:

This is a small step, like #14243, to improve the clarity of `cargo-test-support`s API.

Overall, I'm trying to make it more obvious on https://docs.rs/cargo-test-support/latest/cargo_test_support/ which items to reach for when.  I figured this is one that could be demoted to `paths` When doing so, I noticed `CargoPathExt`. I figured if we had any extension traits for `Path`, then this is a
reasonable one to add.

### DIFF
--- a/crates/cargo-test-support/src/git.rs
+++ b/crates/cargo-test-support/src/git.rs
@@ -38,7 +38,7 @@ use some of the helper functions in this file to interact with the repository.
 
 */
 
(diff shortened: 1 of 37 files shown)


weihanglo (approved):

weihanglo:
Looks reasonable.

@bors r+

PR #14269, opened 2024-07-19
fix(test)!: Clarify extension trait role with rename

epage:
When browsing the docs, I feel like this will make it easier to tell the role of these traits within the API.
Or in other words, I can easily tell that these fill a support role and for what, so I know when I care to look into them.

We use this naming pattern for
`cargo_test_support::paths::CargoPathExt`.


ehuss:
Thanks!

@bors r+

PR #14270, opened 2024-07-19
fix(test): Move 'cargo_home' from 'install' to 'paths' 

epage:

This is used outside of `cargo install` contexts and this makes it more
discoverable for those use cases.  That I found places that weren't using `cargo_home` but could helps illustrate the point.


weihanglo (approved):

weihanglo:
Seems reasonable.

@bors r+

PR #14271, opened 2024-07-19
Add `TomlPackage::new`, `Default` for `TomlWorkspace`

Alexendoo:
Ran into this when using it to create a `Cargo.toml` rather than consume an existing one

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -461,7 +461,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-util-schemas"
-version = "0.5.0"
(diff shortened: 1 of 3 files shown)


epage:
Note that the release cycle for this package follows the rest of Cargo which means this won't get released until Cargo 1.81 is released (iirc)

Alexendoo:
> Note that the release cycle for this package follows the rest of Cargo which means this won't get released until Cargo 1.81 is released (iirc)

`// TODO` it is then 😄

PR #14273, opened 2024-07-20
Use `Rc` instead of `Arc` for storing rustflags

gmorenz:
# What does this PR try to resolve?

There's no reason that these reference counted pointers (which I introduced in #13900) need to be atomic, so let's use non-atomic pointers.

# Additional information

First noticed by @weihanglo [here](https://github.com/rust-lang/cargo/pull/14205#discussion_r1678347073).

r? @weihanglo


weihanglo (approved):

weihanglo:
Thanks!

If we need it Sync in the future, we can always change later.

@bors r+

PR #14279, opened 2024-07-21
Improved error message when `update --breaking` invalid spec.

torhovland:
Improves an error message when trying to do `cargo update --breaking clap@foo`:

```
- [ERROR] expected a version like "1.32"
+ [ERROR] invalid package ID specification: `clap@foo`
+ Caused by:
+   expected a version like "1.32"
```

Related to #12425. Fixes an item [here](https://github.com/rust-lang/cargo/issues/12425#issuecomment-2186198258), as noted [here](https://github.com/rust-lang/cargo/pull/14049#discussion_r1656929921).

### DIFF
--- a/tests/testsuite/update.rs
+++ b/tests/testsuite/update.rs
@@ -2263,7 +2263,10 @@ fn update_breaking_spec_version() {
         .masquerade_as_nightly_cargo(&["update-breaking"])
         .with_status(101)
         .with_stderr_data(str![[r#"
-[ERROR] expected a version like "1.32"
+[ERROR] invalid package ID specification: `incompatible@foo`
+
+Caused by:
+  expected a version like "1.32"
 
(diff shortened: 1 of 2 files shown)


tests/testsuite/update.rs
```diff
@@ -2263,7 +2263,10 @@ fn update_breaking_spec_version() {
         .masquerade_as_nightly_cargo(&["update-breaking"])
         .with_status(101)
         .with_stderr_data(str![[r#"
-[ERROR] expected a version like "1.32"
+[ERROR] invalid package ID specification: `incompatible@foo`
+
+Caused by:
+  expected a version like "1.32"
```
weihanglo, line 2269: Not optimal but better. Thanks!

weihanglo (approved):

PR #14280, opened 2024-07-21
Add a `--dry-run` flag to the `install` command

Flowrey:

This PR add the `--dry-run` flag to the `cargo install` command (see #11123).
I've tried to do the bare minimal for this flag to work without changing anything in the output.

In my opinion, the `--dry-run` flag should mimic as close as possible the behavior of the normal command to avoid missing potential issue in the normal execution. ~~Currently we're missing information about where the binary will be installed.~~

Unlike #13598 this PR:
- Include as much of the compilation process as possible without actually compiling
- use the information provided by `BuildContext` instead of `InstallablePackage::new`
- in the same way as `unit_graph`, it add a `dry_run` to the `CompileOptions` and return a `Compilation::new` from the function `compile_ws` without actually compiling.
- keeps the output the same rather than adding  status messages indicating which very broad actions would be performed
- ~~remove some warning not relevant in the case of  a `--dry-run`~~

Like #13598, the version check and crate downloads still occur.

The first commit include a unit tests to ensure that no binary is actually installed after the dry run.
There is also a snapshot test that show the diff output of the `--help` flag.

### Additional information

Tests and documentation done in #13598, may be cherry picked into this PR if needed.

### DIFF
--- a/src/bin/cargo/commands/install.rs
+++ b/src/bin/cargo/commands/install.rs
@@ -69,6 +69,7 @@ pub fn cli() -> Command {
         )
         .arg(opt("root", "Directory to install packages into").value_name("DIR"))
         .arg(flag("force", "Force overwriting existing crates or binaries").short('f'))
+        .arg_dry_run("Perform all checks without installing (unstable)")
         .arg(flag("no-track", "Do not save tracking information"))
         .arg(flag(
             "list",
@@ -200,7 +201,9 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
 
     compile_opts.build_config.requested_profile =
         args.get_profile_name("release", ProfileChecking::Custom)?;
-
+    if args.dry_run() {
+        gctx.cli_unstable().fail_if_stable_opt("--dry-run", 11123)?;
+    }
     if args.flag("list") {
         ops::install_list(root, gctx)?;
     } else {
@@ -213,6 +216,7 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
             &compile_opts,
             args.flag("force"),
             args.flag("no-track"),
+            args.dry_run(),
         )?;
     }
     Ok(())
--- a/src/cargo/core/compiler/build_runner/mod.rs
+++ b/src/cargo/core/compiler/build_runner/mod.rs
@@ -126,6 +126,27 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
         })
     }
 
+    /// Dry-run the compilation without actually running it.
+    ///
+    /// This is expected to collect information like the location of output artifacts.
+    /// Please keep in sync with non-compilation part in [`BuildRunner::compile`].
+    pub fn dry_run(mut self) -> CargoResult<Compilation<'gctx>> {
+        let _lock = self
+            .bcx
+            .gctx
+            .acquire_package_cache_lock(CacheLockMode::Shared)?;
+        self.lto = super::lto::generate(self.bcx)?;
+        self.prepare_units()?;
+        self.prepare()?;
+        self.check_collisions()?;
+
+        for unit in &self.bcx.roots {
+            self.collect_tests_and_executables(unit)?;
+        }
+
+        Ok(self.compilation)
+    }
+
     /// Starts compilation, waits for it to finish, and returns information
     /// about the result of compilation.
     ///
--- a/src/cargo/core/compiler/build_runner/mod.rs
+++ b/src/cargo/core/compiler/build_runner/mod.rs
@@ -214,31 +214,7 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
 
         // Collect the result of the build into `self.compilation`.
         for unit in &self.bcx.roots {
-            // Collect tests and executables.
-            for output in self.outputs(unit)?.iter() {
-                if output.flavor == FileFlavor::DebugInfo || output.flavor == FileFlavor::Auxiliary
-                {
-                    continue;
-                }
-
-                let bindst = output.bin_dst();
-
-                if unit.mode == CompileMode::Test {
-                    self.compilation
-                        .tests
-                        .push(self.unit_output(unit, &output.path));
-                } else if unit.target.is_executable() {
-                    self.compilation
-                        .binaries
-                        .push(self.unit_output(unit, bindst));
-                } else if unit.target.is_cdylib()
-                    && !self.compilation.cdylibs.iter().any(|uo| uo.unit == *unit)
-                {
-                    self.compilation
-                        .cdylibs
-                        .push(self.unit_output(unit, bindst));
-                }
-            }
+            self.collect_tests_and_executables(unit)?;
 
             // Collect information for `rustdoc --test`.
             if unit.mode.is_doc_test() {
@@ -305,6 +281,33 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
             }
         }
         Ok(self.compilation)
+    }
+
+    fn collect_tests_and_executables(&mut self, unit: &Unit) -> CargoResult<()> {
+        for output in self.outputs(unit)?.iter() {
+            if output.flavor == FileFlavor::DebugInfo || output.flavor == FileFlavor::Auxiliary {
+                continue;
+            }
+
+            let bindst = output.bin_dst();
+
+            if unit.mode == CompileMode::Test {
+                self.compilation
+                    .tests
+                    .push(self.unit_output(unit, &output.path));
+            } else if unit.target.is_executable() {
+                self.compilation
+                    .binaries
+                    .push(self.unit_output(unit, bindst));
+            } else if unit.target.is_cdylib()
+                && !self.compilation.cdylibs.iter().any(|uo| uo.unit == *unit)
+            {
+                self.compilation
+                    .cdylibs
+                    .push(self.unit_output(unit, bindst));
+            }
+        }
+        Ok(())
     }
 
     /// Returns the executable for the specified unit (if any).
--- a/src/cargo/ops/cargo_compile/mod.rs
+++ b/src/cargo/ops/cargo_compile/mod.rs
@@ -156,7 +156,11 @@ pub fn compile_ws<'a>(
     }
     crate::core::gc::auto_gc(bcx.gctx);
     let build_runner = BuildRunner::new(&bcx)?;
-    build_runner.compile(exec)
+    if options.build_config.dry_run {
+        build_runner.dry_run()
+    } else {
+        build_runner.compile(exec)
+    }
 }
 
 /// Executes `rustc --print <VALUE>`.
--- a/src/cargo/ops/cargo_install.rs
+++ b/src/cargo/ops/cargo_install.rs
@@ -297,7 +297,7 @@ impl<'gctx> InstallablePackage<'gctx> {
         Ok(duplicates)
     }
 
-    fn install_one(mut self) -> CargoResult<bool> {
+    fn install_one(mut self, dry_run: bool) -> CargoResult<bool> {
         self.gctx.shell().status("Installing", &self.pkg)?;
 
         let dst = self.root.join("bin").into_path_unlocked();
@@ -321,6 +321,7 @@ impl<'gctx> InstallablePackage<'gctx> {
         self.check_yanked_install()?;
 
         let exec: Arc<dyn Executor> = Arc::new(DefaultExecutor);
+        self.opts.build_config.dry_run = dry_run;
         let compile = ops::compile_ws(&self.ws, &self.opts, &exec).with_context(|| {
             if let Some(td) = td_opt.take() {
                 // preserve the temporary directory, so the user can inspect it
@@ -419,13 +420,15 @@ impl<'gctx> InstallablePackage<'gctx> {
         let staging_dir = TempFileBuilder::new()
             .prefix("cargo-install")
             .tempdir_in(&dst)?;
-        for &(bin, src) in binaries.iter() {
-            let dst = staging_dir.path().join(bin);
-            // Try to move if `target_dir` is transient.
-            if !self.source_id.is_path() && fs::rename(src, &dst).is_ok() {
-                continue;
+        if !dry_run {
+            for &(bin, src) in binaries.iter() {
+                let dst = staging_dir.path().join(bin);
+                // Try to move if `target_dir` is transient.
+                if !self.source_id.is_path() && fs::rename(src, &dst).is_ok() {
+                    continue;
+                }
+                paths::copy(src, &dst)?;
             }
-            paths::copy(src, &dst)?;
         }
 
         let (to_replace, to_install): (Vec<&str>, Vec<&str>) = binaries
@@ -441,11 +444,13 @@ impl<'gctx> InstallablePackage<'gctx> {
             let src = staging_dir.path().join(bin);
             let dst = dst.join(bin);
             self.gctx.shell().status("Installing", dst.display())?;
-            fs::rename(&src, &dst).with_context(|| {
-                format!("failed to move `{}` to `{}`", src.display(), dst.display())
-            })?;
-            installed.bins.push(dst);
-            successful_bins.insert(bin.to_string());
+            if !dry_run {
+                fs::rename(&src, &dst).with_context(|| {
+                    format!("failed to move `{}` to `{}`", src.display(), dst.display())
+                })?;
+                installed.bins.push(dst);
+                successful_bins.insert(bin.to_string());
+            }
         }
 
         // Repeat for binaries which replace existing ones but don't pop the error
@@ -456,10 +461,12 @@ impl<'gctx> InstallablePackage<'gctx> {
                     let src = staging_dir.path().join(bin);
                     let dst = dst.join(bin);
                     self.gctx.shell().status("Replacing", dst.display())?;
-                    fs::rename(&src, &dst).with_context(|| {
-                        format!("failed to move `{}` to `{}`", src.display(), dst.display())
-                    })?;
-                    successful_bins.insert(bin.to_string());
+                    if !dry_run {
+                        fs::rename(&src, &dst).with_context(|| {
+                            format!("failed to move `{}` to `{}`", src.display(), dst.display())
+                        })?;
+                        successful_bins.insert(bin.to_string());
+                    }
                 }
                 Ok(())
             };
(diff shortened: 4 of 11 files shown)


tests/testsuite/install.rs
```diff
@@ -2730,3 +2730,20 @@ fn uninstall_running_binary() {
...
+        .masquerade_as_nightly_cargo(&["install::dry-run"])
+        .with_stderr_data(str![[r#"
+[UPDATING] `dummy-registry` index
+[DOWNLOADING] crates ...
+[DOWNLOADED] foo v0.0.1 (registry `dummy-registry`)
+[INSTALLING] foo v0.0.1
+
+"#]])
+        .run();
+    assert_has_not_installed_exe(paths::cargo_home(), "foo");
```
epage, line 2748: `cargo test` should pass on every commit.  So instead, this should either be a test without `--dry-run` and then ext commit adds it, or this errors about the flag not existing.

tests/testsuite/cargo_install/help/stdout.term.svg
```diff
@@ -35,103 +35,105 @@
 </tspan>
     <tspan x="10px" y="154px"><tspan class="fg-green bold">Options:</tspan>
 </tspan>
-    <tspan x="10px" y="172px"><tspan>      </tspan><tspan class="fg-cyan bold">--version</tspan><tspan class="fg-cyan"> </tspan><tspan class="fg-cyan">&lt;VERSION&gt;</tspan><tspan>     Specify a version to install</tspan>
+    <tspan x="10px" y="172px"><tspan>  </tspan><tspan class="fg-cyan bold">-n</tspan><tspan>, </tspan><tspan class="fg-cyan bold">--dry-run</tspan><tspan>               Perform all checks without installing (unstable)</tspan>
```
epage, line 38: `cargo test` should pass on every commit.

src/bin/cargo/commands/install.rs
```diff
@@ -16,6 +16,7 @@ use cargo_util::paths;
 pub fn cli() -> Command {
     subcommand("install")
         .about("Install a Rust binary")
+        .arg_dry_run("Perform all checks without installing (unstable)")
```
epage, line 19: I'm thinking this would go better next to `--force`.  Thoughts?

epage, line 19: > Tests and documentation done in https://github.com/rust-lang/cargo/pull/13598, may be cherry picked into this PR if needed.

This would be expected for merging

src/bin/cargo/commands/install.rs
```diff
@@ -16,6 +16,7 @@ use cargo_util::paths;
 pub fn cli() -> Command {
```
epage, line 16: In your PR description, could you compare this with #13598?

Also, could you explicitly call out in the description your stance on the discussion in #11123 about how much is being done during dry-run?

tests/testsuite/install.rs
Flowrey: I've rebased my fisrt commit so it test the output without the `--dry-run` flag, the second commit change this test and add the `--dry-run` flag to the options so we can see the diff in git

tests/testsuite/cargo_install/help/stdout.term.svg
Flowrey: I've moved this in the same commit as the one who add the flag.

tests/testsuite/install.rs
epage: Thanks!

src/bin/cargo/commands/install.rs
Flowrey: I agree, i've moved it next to `--force`.

src/bin/cargo/commands/install.rs
Flowrey: I've updated the description as requested, does it suits you?

epage: Thanks!

src/cargo/ops/cargo_install.rs
```diff
@@ -385,7 +386,7 @@ impl<'gctx> InstallablePackage<'gctx> {
                 .iter()
                 .filter(|t| t.is_executable())
                 .collect();
-            if !binaries.is_empty() {
+            if !binaries.is_empty() && !dry_run {
```
epage, line 389: We should not bypass this warning on dry-run as this could affect people's decisions for what command-line arguments they want to pass

src/cargo/ops/cargo_install.rs
```diff
@@ -745,7 +747,7 @@ pub fn install(
         let path = gctx.get_env_os("PATH").unwrap_or_default();
         let dst_in_path = env::split_paths(&path).any(|path| path == dst);
 
-        if !dst_in_path {
+        if !dst_in_path && !dry_run {
```
epage, line 750: I'm mixed about whether we should show this in dry-run or not.

tests/testsuite/install.rs
```diff
@@ -2730,3 +2730,20 @@ fn uninstall_running_binary() {
...
+fn dry_run() {
+    pkg("foo", "0.0.1");
+
+    cargo_process("-Z unstable-options install --dry-run foo")
+        .masquerade_as_nightly_cargo(&["install::dry-run"])
+        .with_stderr_data(str![[r#"
+[UPDATING] `dummy-registry` index
+[DOWNLOADING] crates ...
+[DOWNLOADED] foo v0.0.1 (registry `dummy-registry`)
+[INSTALLING] foo v0.0.1
```
epage, line 2744: We need an "aborting install" warning message at the end

src/bin/cargo/commands/install.rs
epage: Sorry, I forgot something.  Once you've reslved y comments, if you could more explicitly call out what warnings are skipped and why, that'd help!

(probably should also say why in comments in the code)

tests/testsuite/install.rs
```diff
@@ -2730,3 +2730,20 @@ fn uninstall_running_binary() {
 
 "#]]).run();
 }
+
+#[cargo_test]
+fn dry_run() {
+    pkg("foo", "0.0.1");
```
epage, line 2736: A case I think would be important for us to test for is `cargo install --dry-run some-package-from-the-distant-future` (we tend to use a rust_version of `"1.2345"`).

Ideally, we'd get the error about it being incompatible.

Its produced in `create_bcx` which looks like its called in dry-run mode, so it should be a simple change of adding the test to the test commit and updating it in the feature commit.  If its anything more, I'm fine with i being deferred out and tracked in #11123 as part of stabilization

src/cargo/ops/cargo_install.rs
Flowrey: Fixed in 86415010617b63c981fef8c18328331e11d47434

src/cargo/ops/cargo_install.rs
Flowrey: Fixed in 86415010617b63c981fef8c18328331e11d47434.
I added it in the first place because i was missing information about the executable, notably the path, so he was not relevant. Since now we know where the binary will be located this waning seems relevant to me.

src/bin/cargo/commands/install.rs
Flowrey: Finally with 86415010617b63c981fef8c18328331e11d47434, no warning are skipped.

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -155,7 +155,8 @@ pub fn compile_ws<'a>(
         return Compilation::new(&bcx);
     }
     if options.build_config.dry_run {
-        return Compilation::new(&bcx);
+        let build_runner = BuildRunner::new(&bcx)?;
+        return build_runner.dry_run();
```
epage, line 159: I'm seeing other `build_runner.compile` checks that might be worth having, like collisions.  Should we be like `unit_graph` and be in the `build_context` and have `build_runner.compile` decide how much dry-run behavior to do?

src/cargo/ops/cargo_install.rs
```diff
@@ -442,9 +442,11 @@ impl<'gctx> InstallablePackage<'gctx> {
...
-            fs::rename(&src, &dst).with_context(|| {
-                format!("failed to move `{}` to `{}`", src.display(), dst.display())
-            })?;
+            if !dry_run {
+                fs::rename(&src, &dst).with_context(|| {
+                    format!("failed to move `{}` to `{}`", src.display(), dst.display())
+                })?;
+            }
             installed.bins.push(dst);
             successful_bins.insert(bin.to_string());
```
epage, line 451: Should we include `installed.bins` and `successful_bins` pushes to only be on real runs?

tests/testsuite/install.rs
Flowrey: According to you, is it still needed to have this message ? Since the last message is now `[INSTALLED] package ...` i think that it would be misleading to have an ''aborting install" message.

src/cargo/ops/cargo_install.rs
```diff
@@ -442,9 +442,11 @@ impl<'gctx> InstallablePackage<'gctx> {
...
         }
```
epage, line 452: Can't highlight it but we also need to have dry-run affect
- `replace_to`s rename (should also get a test)
- orphan removal (would be nice to have a test)

weihanglo: Thanks for calling this out. Coming from <https://github.com/rust-lang/cargo/issues/11123#issuecomment-1910469992> we need to answer these user questions. I think the first item is the only one left.


- [ ] How: Is it an update (replace), a fresh install, or no-op?
- [x] What: which binary from which source is going to be installed
- [x] Where: to where the binary will be installed

tests/testsuite/install.rs
epage: Without it, we've told the user we've installed.  If there is a concern with other messages, we can consider tweaking them

tests/testsuite/install.rs
Flowrey: I rebased my two first commit to add this test in it

tests/testsuite/install.rs
Flowrey: Changed in 63a4955c7ed614df8f6f454b66c5fe9488c215e7.
I've decided to display an `[ABORTING] install` message instead of `[INSTALLED] package` to avoid confusion , are you okay with that ?

src/cargo/ops/cargo_install.rs
Flowrey: If i add `installed.bins` and `successful_bins` inside the if statement it change nothing in the output, according to you should i do this to avoid any potential side effect ?

tests/testsuite/install.rs
epage: 1. I'd like the style of dry-run between commands to be consistent
2. I'd a command's output to be consistent whether dry-run or not

What are your thoughts on doing something like #13367 and switching the messages to `Installing` / `Replacing` and printing them before we act on them?

src/bin/cargo/commands/install.rs
Flowrey: I added the doc made in #13598 and removed the part about the specific dry run messages

tests/testsuite/install.rs
Flowrey: So if i understood correctly, you want to replace the `Installed` message by an `Installing` message, before we act on them, in both dry-run an normal mode, is it this ?
Should we still keep the `[ABORTING]` message then ?

src/cargo/ops/cargo_install.rs
Flowrey: Oh yes indeed, the `InstallTracker` persist data on disk, so yes i think we should include them in the `if` statement.

src/cargo/ops/cargo_compile/mod.rs
Flowrey: I added the `collision` check.
Maybe I'm not understanding this correctly but according to me we can't do like `unit_graph` because we need to call method that are private to the `BuildRunner` class and are needed notably for getting the destination path of the executable we are installing.

src/cargo/ops/cargo_install.rs
Flowrey: I made `installed.bins` and `successful_bins` pushes exclusive to real runs

tests/testsuite/install.rs
epage: Its one way of resolving this problem that I'm hopeful will work out.  If we go this route, we should probably break this out into its own PR so we can evaluate it on its own.

> Should we still keep the [ABORTING] message then ?

We should just do the `[WARNING]` like other dry-runs

src/doc/man/cargo-install.md
```diff
@@ -121,6 +121,13 @@ Filesystem path to local crate to install from.
 List all installed packages and their versions.
 {{/option}}
 
+{{#option "`-n`" "`--dry-run`" }}
+(unstable) Cut the install operation short and display what it *would* have
+usually done. The normal version check and crate downloads still occur exactly
+like without the option flag, but the compilation and the actual installation
```
epage, line 127: imo this over-specificies (I could see us changing compilation)

Our other dry-run help descriptions are pretty brief

publish

> Perform all checks without uploading.

add

> Don’t actually write the manifest

update

> Displays what would be updated, but doesn’t actually write the lockfile.

tests/testsuite/install.rs
```diff
@@ -2780,3 +2780,85 @@ fn dry_run_incompatible_package() {
         .run();
     assert_has_not_installed_exe(paths::cargo_home(), "some-package-from-the-distant-future");
 }
+
+#[cargo_test]
+fn dry_run_upgrade() {
```
epage, line 2785: Can these tests be moved to the original test commit?

tests/testsuite/install.rs
Flowrey: Ok so for now i just removed the `[ABORTING]` message and displayed a `[WARNING]` message like for `cargo add --dry-run`

src/doc/man/cargo-install.md
Flowrey: Changed to something less specific

tests/testsuite/install.rs
Flowrey: Done

src/cargo/ops/cargo_install.rs
```diff
@@ -422,7 +423,7 @@ impl<'gctx> InstallablePackage<'gctx> {
         for &(bin, src) in binaries.iter() {
             let dst = staging_dir.path().join(bin);
             // Try to move if `target_dir` is transient.
-            if !self.source_id.is_path() && fs::rename(src, &dst).is_ok() {
+            if (!self.source_id.is_path() && fs::rename(src, &dst).is_ok()) || dry_run {
```
epage, line 426: Shouldn't this be `if dry_run ||` so we don't do the `fs::rename`?

tests/testsuite/install.rs
```diff
@@ -2734,3 +2734,132 @@ fn uninstall_running_binary() {
...
+                "some-package-from-the-distant-future"
+            ),
+        )
+        .publish();
+
+    cargo_process("install some-package-from-the-distant-future")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[UPDATING] `dummy-registry` index
+[ERROR] cannot install package `some-package-from-the-distant-future 0.0.1`, it requires rustc 1.2345.0 or newer, while the currently active rustc version is [..]
```
epage, line 2777: This tests for the error when installing `some-package-from-the-distant-future`.  We also need a test for when `some-package-from-the-distant-future` is a dependency.  That is what will hit `create_bcx` and that section of MSRV code, ensuring we are doing enough of the compile process to hit it.

src/cargo/ops/cargo_install.rs
Flowrey: Absolutely, I fixed this

tests/testsuite/install.rs
Flowrey: I added a test for when `some-package-from-the-distant-future` is a dependency.

Flowrey:
> Note: I tweaked the commits to make this easier to review.

Thanks, sorry if i made a mess with the commits :sweat_smile:

src/cargo/ops/cargo_install.rs
```diff
@@ -422,7 +423,7 @@ impl<'gctx> InstallablePackage<'gctx> {
         for &(bin, src) in binaries.iter() {
             let dst = staging_dir.path().join(bin);
             // Try to move if `target_dir` is transient.
-            if !self.source_id.is_path() && fs::rename(src, &dst).is_ok() {
+            if dry_run || (!self.source_id.is_path() && fs::rename(src, &dst).is_ok()) {
```
weihanglo, line 426: Instead of this, should we just not execute the entire loop when `dry_run` is true?

src/cargo/core/compiler/build_runner/mod.rs
```diff
@@ -126,6 +126,23 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
         })
     }
 
+    pub fn dry_run(mut self) -> CargoResult<Compilation<'gctx>> {
```
weihanglo, line 129: Should we put a comment here about the expectation of how dryrun works? Supposedly we want to get the information of where dylibs and executables reside. So probably something like:

```suggestion
    /// Dry-run the compilation without actually running it.
    ///
    /// This is expected to collect information like the location of output artifacts.
    /// Please keep in sync with non-compilation part in [`BuildRunner::compile`].
    pub fn dry_run(mut self) -> CargoResult<Compilation<'gctx>> {
```

tests/testsuite/install.rs
```diff
@@ -2734,3 +2734,186 @@ fn uninstall_running_binary() {
...
+    Package::new("bar", "1.0.0")
+        .file("src/lib.rs", "")
+        .file(
+            "src/bin/client.rs",
+            &format!("extern crate {}; fn main() {{}}", "bar"),
+        )
+        .file(
+            "src/bin/server.rs",
+            &format!("extern crate {}; fn main() {{}}", "bar"),
+        )
```
weihanglo, line 2873: This should be sufficient. We don't need the lib.
Likewise in other places. We just need bin targets with `fn main() {}`

```suggestion
        .file("src/bin/client.rs", "fn main() {}")
        .file("src/bin/server.rs", "fn main() {}")
```

src/cargo/ops/cargo_install.rs
Flowrey: You're right, i fixed this!

src/cargo/core/compiler/build_runner/mod.rs
Flowrey: I applied your suggestion

tests/testsuite/install.rs
Flowrey: Fixed

weihanglo (approved):
Great job! This should be ready to merge.

PR #14296, opened 2024-07-24
fix(config): Adjust MSRV resolve config field name / values

### LINKED ISSUE #13540: Config support for controlling MSRV-aware resolver

epage:

Fixes #13540

### DIFF
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -344,7 +344,7 @@ This was stabilized in 1.79 in [#13608](https://github.com/rust-lang/cargo/pull/
 
 The resolver will prefer dependencies with a `package.rust-version` that is the same or older than your project's MSRV.
 Your project's MSRV is determined by taking the lowest `package.rust-version` set among your workspace members.
-If there is none set, your toolchain version will be used with the intent to pick up the version from rustup's `rust-toolchain.toml`, if present.
+If there is no MSRV set then your toolchain version will be used, allowing it to pick up the toolchain version from pinned in rustup (e.g. `rust-toolchain.toml`).
 
 #### `resolver.incompatible-rust-versions`
 * Type: string
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -338,7 +338,7 @@ This was stabilized in 1.79 in [#13608](https://github.com/rust-lang/cargo/pull/
 ### MSRV-aware resolver
 
 `-Zmsrv-policy` allows access to an MSRV-aware resolver which can be enabled with:
-- `resolver.something-like-precedence` config field
+- `resolver.incompatible-rust-versions` config field
 - `workspace.resolver = "3"` / `package.resolver = "3"`
 - `package.edition = "2024"` (only in workspace root)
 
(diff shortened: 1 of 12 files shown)


src/doc/src/reference/unstable.md
```diff
@@ -338,26 +338,26 @@ This was stabilized in 1.79 in [#13608](https://github.com/rust-lang/cargo/pull/
...
 
 `-Zmsrv-policy` allows access to an MSRV-aware resolver which can be enabled with:
-- `resolver.something-like-precedence` config field
+- `resolver.incompatible-rust-versions` config field
 - `workspace.resolver = "3"` / `package.resolver = "3"`
 - `package.edition = "2024"` (only in workspace root)
 
 The resolver will prefer dependencies with a `package.rust-version` that is the same or older than your project's MSRV.
 Your project's MSRV is determined by taking the lowest `package.rust-version` set among your workspace members.
 If there is none set, your toolchain version will be used with the intent to pick up the version from rustup's `rust-toolchain.toml`, if present.
```
weihanglo, line 347: Slightly off-topic, but we may not want to expose the implementation detail of rustup.

```suggestion
If there is none set, your toolchain version will be used. The toolchain version is usually determined by `rustup` or the system package manager. 
```

src/doc/src/reference/unstable.md
```diff
@@ -338,26 +338,26 @@ This was stabilized in 1.79 in [#13608](https://github.com/rust-lang/cargo/pull/
...
 * Type: string
-* Default: "something-like-maximum"
-* Environment: `CARGO_RESOLVER_SOMETHING_LIKE_PRECEDENCE`
+* Default: "allow"
+* Environment: `CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS`
 
-Select which policy should be used when resolving dependencies.  Values include
-- `something-like-maximum`: prefer highest compatible versions of a package
-- `something-like-rust-version`: prefer versions of packages compatible with your project's Rust version
+Select how packages with incompatible rust-versions should be resolved.  Values include:
```
weihanglo, line 354: It is unclear what an incompatible rust-version is. Should we link "incompatible rust-versions" to `package.rust-version` in the manifest?

src/doc/src/reference/unstable.md
```diff
@@ -338,26 +338,26 @@ This was stabilized in 1.79 in [#13608](https://github.com/rust-lang/cargo/pull/
...
+- `allow`: treat them like any other dependency
```
weihanglo, line 355: This is also unclear as user doesn't know how other dependency is treated. Are we making it vague for rooms for future behavior changes?

src/doc/src/reference/unstable.md
```diff
@@ -338,26 +338,26 @@ This was stabilized in 1.79 in [#13608](https://github.com/rust-lang/cargo/pull/
...
+- `fallback`: only consider them if no other dependency version matched
 
 Can be overridden with
 - `--ignore-rust-version` CLI option
-- Setting the dependency's version requirement too high
+- Setting the dependency's version requirement above any version that is rust-version compatible
```
weihanglo, line 360: Should we highlight the field name?

```suggestion
- Setting the dependency's version requirement above any version with a compatible `package.rust-version`
```

src/doc/src/reference/unstable.md
epage: This section isn't as important long-term as this is just describing the unstable feature.

However, in describing it this way, I'm trying to accomplish two things
- Emphasize the motivating use case for this functionality
- Help rustup users discover this by using terms they would expect

I am making a slight tweak to it so hopefully that will help.

src/doc/src/reference/unstable.md
epage: I can't think of a reason to be intentionally vague about `allow`, I'm just unsure of what more needs to be described.

src/doc/src/reference/unstable.md
epage: I've called out the field name.  We should eventually turn all of those into links.

epage:
Removing the draft status as I've given time for some community input with only a passing reference to not liking something but the person didn't engage to get more details.

weihanglo:
> Removing the draft status as I've given time for some community input with only a passing reference to not liking something but the person didn't engage to get more details.

Where can I find this?

src/doc/src/reference/unstable.md
weihanglo: I think the problem is pronouns and “other”. What is included in “other”? Maybe something like


```suggestion
- `allow`: treat `package.rust-version` incompatible dependency as compatible
- `fallback`: only consider `package.rust-version` incompatible dependency if no other compatible dependency version matched
```

Pronouns are often confusing, and I prefer to use them less.

(Maybe it is just me 😬)

epage:
> > Removing the draft status as I've given time for some community input with only a passing reference to not liking something but the person didn't engage to get more details.
> 
> Where can I find this?

From https://hachyderm.io/@djc/112842279033657989

> [@epage](https://hachyderm.io/@epage) so the current behavior is allow and the new default would be fallback, right? I don’t love fallback because it describes a sort of meta-behavior that the resolver implements rather than a decision on the particular package version. From that perspective, I feel like “ignore” might be a better fit? “When Cargo encounters a dependency version with an incompatible rust-version, it allows/ignores that version.”

src/doc/src/reference/unstable.md
epage: To be fully correct, its an "incompatible dependency version" (not just dependency).  This starts to become a mouthful.  I gave it a try by skipping on `package.` and just saying "version".

weihanglo (approved):
This seems fine to merge now. We will do an FCP when stabilizing it.

PR #14298, opened 2024-07-25
fix grammar

zzz6519003:


ehuss:
Thanks for the PR! However, I think the original is correct. It is setting up a hypothetical, and it is common in English to use past tense to refer to that hypothetical situation. Thanks, though!

PR #14299, opened 2024-07-25
chore: Update dependencies

epage:
I just published all of my packages using Cargo 1.80 and wanted to see the effect of the enumerating targets during publish (#13849).

### DIFF
--- a/crates/rustfix/Cargo.toml
+++ b/crates/rustfix/Cargo.toml
@@ -26,7 +26,7 @@ tracing.workspace = true
 [dev-dependencies]
 anyhow.workspace = true
 proptest.workspace = true
-similar = "2.6.0"
+similar.workspace = true
 tempfile.workspace = true
 tracing-subscriber.workspace = true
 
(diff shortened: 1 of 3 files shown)


epage:
bstr 1.9.2 broke our build.  Following up at BurntSushi/bstr#187

weihanglo:
I must be forgetting something. Could you make the purpose a bit clear?

crates/rustfix/Cargo.toml
```diff
@@ -26,7 +26,7 @@ tracing.workspace = true
 [dev-dependencies]
 anyhow.workspace = true
 proptest.workspace = true
-similar = "2.5.0"
+similar = "2.6.0"
```
weihanglo, line 29: This should move to `workspace.dependencies`.

epage:
> I must be forgetting something. Could you make the purpose a bit clear?

I've updated it to link out to #13849.  I re-published clap, toml, snapbox, etc with 1.80 so all build-targets are explicitly enumerated in the `.crate`, allowing dependents to not enumerate them themselves, reducing the no-op runtime of cargo commands that parse the dependency tree (well, shifting the time over to TOML parsing which will hopefully be faster either now or in the future).

weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #14305, opened 2024-07-26
feat: Add matches_prerelease semantic

linyihai:
One implementaion for https://github.com/rust-lang/cargo/issues/13290

Thanks to @Eh2406 feedback, a working version came out, and I think it should be able to test under the nightly feature.

This PR proposes a `matches_prerelease semantic`.

| req              | matches             | matches_prerelease     | matches_prerelease_mirror_node [<sub>2<sub>](#mirror-node) |
|------------------|---------------------|------------------------| ----------------------------------|
| `Op::Exact`      |                     |                        |                                   |
| =I.J.K           | =I.J.K              | >=I.J.K, <I.J.(K+1)-0  | >=I.J.K, <I.J.(K+1)-0             |
| =I.J             | >=I.J.0, <I.(J+1).0 | >=I.J.0, <I.(J+1).0-0  | >=I.J.0-0, <I.(J+1).0-0           |
| =I               | >=I.0.0, <(I+1).0.0 | >=I.0.0, <(I+1).0.0-0  | >=I.0.0-0, <(I+1).0.0-0           |
| `Op::Greater`    |                     |                        |                                   |
| >I.J.K           | >I.J.K              | >I.J.K                 | >I.J.K                            |
| >I.J             | >=I.(J+1).0         | >=I.(J+1).0-0          | >=I.(J+1).0-0                     |
| >I               | >=(I+1).0.0         | >=(I+1).0.0-0          | >=(I+1).0.0-0                     |
| `Op::GreaterEq`  |                     |                        |                                   |
| >=I.J.K          | >=I.J.K             | >=I.J.K                | >=I.J.K                           |
| >=I.J            | >=I.J.0             | >=I.J.0                | >=I.J.0-0                         |
| >=I              | >=I.0.0             | >=I.0.0                | >=I.0.0-0                         |
| `Op::Less`       |                     |                        |                                   |
| <I.J.K           | <I.J.K              | <I.J.K or I.J.K-0 depends [<sub>1<sub>](#op-less) | <I.J.K |
| <I.J             | <I.J.0              | <I.J.0-0               | <I.J.0-0                          |
| <I               | <I.0.0              | <I.0.0-0               | <I.0.0-0                          |
| `Op::LessEq`     |                     |                        |                                   |
| <=I.J.K          | <=I.J.K             | <=I.J.K                | <=I.J.K                           |
| <=I.J            | <I.(J+1).0          | <I.(J+1).0-0           | <I.(J+1).0-0                      |
| <=I              | <(I+1).0.0          | <(I+1).0.0-0           | <(I+1).0.0-0                      |
| `Op::Tilde`      |                     |                        |                                   |
| ~I.J.K           | >=I.J.K, <I.(J+1).0 | >=I.J.K, <I.(J+1).0-0  | >=I.J.K, <I.(J+1).0-0             |
| ~I.J             | =I.J                | >=I.J.0, <I.(J+1).0-0  | >=I.J.0, <I.(J+1).0-0             |
| ~I               | =I                  | >=I.0.0, <(I+1).0.0-0  | >=I.0.0, <(I+1).0.0-0             |
| `Op::Caret`      |                     |                        |                                   |
| ^I.J.K (for I>0) | >=I.J.K, <(I+1).0.0 | >=I.J.K, <(I+1).0.0-0  | >=I.J.K, <(I+1).0.0-0             |
| ^0.J.K (for J>0) | >=0.J.K, <0.(J+1).0 | >=0.J.K,  <0.(J+1).0-0 | >=0.J.K-0, <0.(J+1).0-0           |
| ^0.0.K           | =0.0.K              | >=0.0.K,  <0.0.(K+1)-0 | >=0.J.K-0, <0.(J+1).0-0           |
| ^I.J             | ^I.J.0              | >=I.J.0,  <(I+1).0.0-0 | >=I.J.0-0, <(I+1).0.0-0           |
| ^0.0             | =0.0                | >=0.0.0, <0.1.0-0      | >=0.0.0-0, <0.1.0-0               |
| ^I               | =I                  | >=I.0.0, <(I+1).0.0-0  | >=I.0.0-0, <(I+1).0.0-0           |
| `Op::Wildcard`   |                     |                        |                                   |
| `I.J.*`          | =I.J                | >=I.J.0, <I.(J+1).0-0  | >=I.J.0-0, <I.(J+1).0-0           |
| `I.*` or `I.*.*` | =I                  | >=I.0.0, <(I+1).0.0-0  | >=I.0.0-0, <(I+1).0.0-0           |

Notes:

<div id="op-less"></div>

- `<I.J.K`: This is equivalent to `<I.J.K-0` if no lower bound or the lower bound isn't pre-release, otherwise this is equivalent to `<I.J.K`.

<div id="mirror-node"></div>

- [matches_prerelease_mirror_node](https://github.com/dtolnay/semver/pull/321/commits/3464fd136a180fd64c6b0a394ccb0479ccb8d967) is yet another  implementation of [node semver compatibility](https://github.com/npm/node-semver?tab=readme-ov-file#semver1----the-semantic-versioner-for-npm) (with [includePrerelease=true](https://github.com/npm/node-semver?tab=readme-ov-file#functions)) test. This is extrapolated from the test cases and is not necessarily the same as the node implementation

Besides, the proposed semtantic left a [unsolved issuse ](https://github.com/dtolnay/semver/pull/321#issuecomment-2251058953), I don't have clear idea to handle it.

The tests in `src/cargo/util/semver_eval_ext.rs` are designed to reflect current semantic.
 

### Additional information
Migrated from https://github.com/dtolnay/semver/pull/321

TBO, this feature was not conceived in advance plus the semantic is unclear at first.  I need to experiment step by step and find out what I think makes sense.  My experiment can be seen this [comment](https://github.com/rust-lang/cargo/issues/13290#issuecomment-2250051428).


tests/testsuite/precise_pre_release.rs
```diff
@@ -90,30 +90,71 @@ fn update_pre_release_differ() {
...
+
+"#]])
+        .run();
+
+    // Test upgrade
+    // 0.1.3 is in the range, so it match
+    p.cargo("update -p my-dependency --precise 0.1.3 -Zunstable-options")
         .masquerade_as_nightly_cargo(&["precise-pre-release"])
         .with_stderr_data(str![[r#"
 [UPDATING] `dummy-registry` index
```
linyihai, line 139: Here should be some `upgrade` print, I remember which PR or issue mentioned this, didn't I

src/cargo/util/semver_ext.rs
```diff
@@ -111,15 +121,13 @@ impl OptVersionReq {
...
     pub fn matches_prerelease(&self, version: &Version) -> bool {
         if version.is_prerelease() {
-            let mut version = version.clone();
-            version.pre = semver::Prerelease::EMPTY;
-            return self.matches(&version);
+            if let OptVersionReq::Req(req) = self {
+                return req.matches_prerelease(version);
+            }
         }
         self.matches(version)
```
epage, line 132: Should we remove the `is_prerelease` check? I worry there might be some weird corner case we might mask and it seems like it shouldn't be relevant anymore

src/cargo/util/semver_ext.rs
```diff
@@ -217,18 +225,18 @@ mod matches_prerelease {
             //
             (">1.2.3, <1.2.4", "1.2.3-0", false),
             (">1.2.3, <1.2.4", "1.2.3-1", false),
-            (">1.2.3, <1.2.4", "1.2.4-0", false), // upper bound semantic
+            (">1.2.3, <1.2.4", "1.2.4-0", true), // upper bound semantic
```
epage, line 228: I've not followed all of the conversation but this change seems odd.

Technically, its also an odd test case as I would expect this to be an empty range

src/cargo/util/semver_ext.rs
```diff
@@ -217,18 +225,18 @@ mod matches_prerelease {
...
             //
-            (">=1.2.3, <1.2.4", "1.2.3-0", true), // bug, must be false
-            (">=1.2.3, <1.2.4", "1.2.3-1", true), // bug, must be false
-            (">=1.2.3, <1.2.4", "1.2.4-0", false), // upper bound semantic
+            (">=1.2.3, <1.2.4", "1.2.3-0", false),
+            (">=1.2.3, <1.2.4", "1.2.3-1", false),
+            (">=1.2.3, <1.2.4", "1.2.4-0", true), // upper bound semantic
```
epage, line 232: ditto

epage:
I've not gone into a lot of the details, more focusing on the high level

src/cargo/util/semver_ext.rs
weihanglo: If we are going to follow the upper bound semantic in the RFC, this change seems wrong. The equivalent requirements will be `>1.2.3, <1.2.4-0`, which doesn't include 1.2.4. The comment highlights them not because they need to be flipped, but something we must discuss and decide.

src/cargo/util/semver_eval_ext.rs
```diff
@@ -0,0 +1,520 @@
+use semver::{Comparator, Op, Prerelease, Version, VersionReq};
```
weihanglo, line 1: Should we have some mod-level comments explaining what have been copied over from `semver` and why?

weihanglo (commented):
Really appreciate your efforts on prerelease!

src/cargo/util/semver_ext.rs
linyihai: Oh no, I missed this piece in the [Unresolved questions](https://rust-lang.github.io/rfcs/3493-precise-pre-release-cargo-update.html#unresolved-questions). 

The RFC  suggests to pad '0' for `Less op`,  considering all partial version included, the semantic will change the `Less op`
to
```
<I.J.K equivalent to < I.J.K-0
<I.J equivalent to <I.J.0-0
<I equivalent to <I.0.0-0
```

And in another way, if we always treat the "<I.J.K" as "<I.J.K-0", then peple will imposible to speficy that any prerelease version less than `I.J.K`, that seems a problem also?

src/cargo/util/semver_ext.rs
linyihai: OK

src/cargo/util/semver_eval_ext.rs
linyihai: No problems.

linyihai:
For reference, `matches_prerelease_mirror_node` is yet another implementaion of `Node`, this is a little different from what we proposed, see below,


| req              | matches             | matches_prerelease     | matches_prerelease_mirror_node    |
|------------------|---------------------|------------------------| ----------------------------------|
| `Op::Exact`      |                     |                        |                                   |
| =I.J.K           | =I.J.K              | ***>=I.J.K, <I.J.(K+1)-0***  | ***>=I.J.K, <I.J.(K+1)-0*** |
| =I.J             | >=I.J.0, <I.(J+1).0 | >=I.J.0, <I.(J+1).0-0  | >=I.J.0-0, <I.(J+1).0-0           |
| =I               | >=I.0.0, <(I+1).0.0 | >=I.0.0, <(I+1).0.0-0  | >=I.0.0-0, <(I+1).0.0-0           |
| `Op::Greater`    |                     |                        |                                   |
| >I.J.K           | >I.J.K              | ***>I.J.K***           | ***>I.J.K***                      |
| >I.J             | >=I.(J+1).0         | ***>=I.(J+1).0-0***    | ***>=I.(J+1).0-0***               |
| >I               | >=(I+1).0.0         | ***>=(I+1).0.0-0***    | ***>=(I+1).0.0-0***               |
| `Op::GreaterEq`  |                     |                        |                                   |
| >=I.J.K          | >=I.J.K             | ***>=I.J.K***          | ***>=I.J.K***                     |
| >=I.J            | >=I.J.0             | >=I.J.0                | >=I.J.0-0                         |
| >=I              | >=I.0.0             | >=I.0.0                | >=I.0.0-0                         |
| `Op::Less`       |                     |                        |                                   |
| <I.J.K           | <I.J.K              | ***<I.J.K***           | ***<I.J.K***                      |
| <I.J             | <I.J.0              | <I.J.0                 | <I.J.0-0                          |
| <I               | <I.0.0              | <I.0.0                 | <I.0.0-0                          |
| `Op::LessEq`     |                     |                        |                                   |
| <=I.J.K          | <=I.J.K             | ***<=I.J.K***          | ***<=I.J.K***                     |
| <=I.J            | <I.(J+1).0          | ***<I.(J+1).0-0***     | ***<I.(J+1).0-0***                |
| <=I              | <(I+1).0.0          | ***<(I+1).0.0-0***     | ***<(I+1).0.0-0***                |
| `Op::Tilde`      |                     |                        |                                   |
| ~I.J.K           | >=I.J.K, <I.(J+1).0 | ***>=I.J.K, <I.(J+1).0-0***  | ***>=I.J.K, <I.(J+1).0-0*** |
| ~I.J             | =I.J                | ***>=I.J.0, <I.(J+1).0-0***  | ***>=I.J.0, <I.(J+1).0-0*** |
| ~I               | =I                  | ***>=I.0.0, <(I+1).0.0-0***  | ***>=I.0.0, <(I+1).0.0-0*** |
| `Op::Caret`      |                     |                        |                                   |
| ^I.J.K (for I>0) | >=I.J.K, <(I+1).0.0 | ***>=I.J.K, <(I+1).0.0-0***  | ***>=I.J.K, <(I+1).0.0-0*** |
| ^0.J.K (for J>0) | >=0.J.K, <0.(J+1).0 | >=0.J.K,  <0.(J+1).0-0 | >=0.J.K-0, <0.(J+1).0-0           |
| ^0.0.K           | =0.0.K              | >=0.0.K,  <0.0.(K+1)-0 | >=0.J.K-0, <0.(J+1).0-0           |
| ^I.J             | ^I.J.0              | >=I.J.0,  <(I+1).0.0-0 | >=I.J.0-0, <(I+1).0.0-0           |
| ^0.0             | =0.0                | >=0.0.0, <0.1.0-0      | >=0.0.0-0, <0.1.0-0               |
| ^I               | =I                  | >=I.0.0, <(I+1).0.0-0  | >=I.0.0-0, <(I+1).0.0-0           |
| `Op::Wildcard`   |                     |                        |                                   |
| `I.J.*`          | =I.J                | >=I.J.0, <I.(J+1).0-0  | >=I.J.0-0, <I.(J+1).0-0           |
| `I.*` or `I.*.*` | =I                  | >=I.0.0, <(I+1).0.0-0  | >=I.0.0-0, <(I+1).0.0-0           |

**notes**:
- `matches` flows the [limit](https://github.com/npm/node-semver?tab=readme-ov-file#prerelease-tags), that says if the version has a prerelease tag then its MAJOR.MINOR.PATCH must be same with the VersionReq at least once and this VersionReq prerelase tag must speficy, ie. ^1.2.3-0 meets 1.2.3-1 but not ^1.2.4-1, ^1.2.3 will not meets any prerelase version. 
- `matches_prerelease_mirror_node` is a implementation of [`node semver compatibility`](https://github.com/npm/node-semver?tab=readme-ov-file#semver1----the-semantic-versioner-for-npm) (with [includePrerelease=true](https://github.com/npm/node-semver?tab=readme-ov-file#functions)) test. This is extrapolated from the test cases and is not necessarily the same as the node implementation
- the semantic decorated with ***Bold Italic***  between `matches_prerelease` and `matches_prerelease_mirror_node` is totally same

<details>
 <summary>table source</summary>

```
| req              | matches             | matches_prerelease     | matches_prerelease_mirror_node    |
|------------------|---------------------|------------------------| ----------------------------------|
| `Op::Exact`      |                     |                        |                                   |
| =I.J.K           | =I.J.K              | ***>=I.J.K, <I.J.(K+1)-0***  | ***>=I.J.K, <I.J.(K+1)-0*** |
| =I.J             | >=I.J.0, <I.(J+1).0 | >=I.J.0, <I.(J+1).0-0  | >=I.J.0-0, <I.(J+1).0-0           |
| =I               | >=I.0.0, <(I+1).0.0 | >=I.0.0, <(I+1).0.0-0  | >=I.0.0-0, <(I+1).0.0-0           |
| `Op::Greater`    |                     |                        |                                   |
| >I.J.K           | >I.J.K              | ***>I.J.K***           | ***>I.J.K***                      |
| >I.J             | >=I.(J+1).0         | ***>=I.(J+1).0-0***    | ***>=I.(J+1).0-0***               |
| >I               | >=(I+1).0.0         | ***>=(I+1).0.0-0***    | ***>=(I+1).0.0-0***               |
| `Op::GreaterEq`  |                     |                        |                                   |
| >=I.J.K          | >=I.J.K             | ***>=I.J.K***          | ***>=I.J.K***                     |
| >=I.J            | >=I.J.0             | >=I.J.0                | >=I.J.0-0                         |
| >=I              | >=I.0.0             | >=I.0.0                | >=I.0.0-0                         |
| `Op::Less`       |                     |                        |                                   |
| <I.J.K           | <I.J.K              | ***<I.J.K***           | ***<I.J.K***                      |
| <I.J             | <I.J.0              | <I.J.0                 | <I.J.0-0                          |
| <I               | <I.0.0              | <I.0.0                 | <I.0.0-0                          |
| `Op::LessEq`     |                     |                        |                                   |
| <=I.J.K          | <=I.J.K             | ***<=I.J.K***          | ***<=I.J.K***                     |
| <=I.J            | <I.(J+1).0          | ***<I.(J+1).0-0***     | ***<I.(J+1).0-0***                |
| <=I              | <(I+1).0.0          | ***<(I+1).0.0-0***     | ***<(I+1).0.0-0***                |
| `Op::Tilde`      |                     |                        |                                   |
| ~I.J.K           | >=I.J.K, <I.(J+1).0 | ***>=I.J.K, <I.(J+1).0-0***  | ***>=I.J.K, <I.(J+1).0-0*** |
| ~I.J             | =I.J                | ***>=I.J.0, <I.(J+1).0-0***  | ***>=I.J.0, <I.(J+1).0-0*** |
| ~I               | =I                  | ***>=I.0.0, <(I+1).0.0-0***  | ***>=I.0.0, <(I+1).0.0-0*** |
| `Op::Caret`      |                     |                        |                                   |
| ^I.J.K (for I>0) | >=I.J.K, <(I+1).0.0 | ***>=I.J.K, <(I+1).0.0-0***  | ***>=I.J.K, <(I+1).0.0-0*** |
| ^0.J.K (for J>0) | >=0.J.K, <0.(J+1).0 | >=0.J.K,  <0.(J+1).0-0 | >=0.J.K-0, <0.(J+1).0-0           |
| ^0.0.K           | =0.0.K              | >=0.0.K,  <0.0.(K+1)-0 | >=0.J.K-0, <0.(J+1).0-0           |
| ^I.J             | ^I.J.0              | >=I.J.0,  <(I+1).0.0-0 | >=I.J.0-0, <(I+1).0.0-0           |
| ^0.0             | =0.0                | >=0.0.0, <0.1.0-0      | >=0.0.0-0, <0.1.0-0               |
| ^I               | =I                  | >=I.0.0, <(I+1).0.0-0  | >=I.0.0-0, <(I+1).0.0-0           |
| `Op::Wildcard`   |                     |                        |                                   |
| `I.J.*`          | =I.J                | >=I.J.0, <I.(J+1).0-0  | >=I.J.0-0, <I.(J+1).0-0           |
| `I.*` or `I.*.*` | =I                  | >=I.0.0, <(I+1).0.0-0  | >=I.0.0-0, <(I+1).0.0-0           |
```

 </details>

src/cargo/util/semver_ext.rs
epage: Hence the conundrum.  One way to solve is messy: only determine how to evaluate `<` bounds by looking at any lower bounds that also exist.

linyihai: > only determine how to evaluate < bounds by looking at any lower bounds that also exist.

I had updated the semantic for `Less Op` now it will meet the `upper bound semantic`

linyihai: > only determine how to evaluate < bounds by looking at any lower bounds that also exist.

I have updated the `Less Op` semantic. Now it will pad prerelase with "0"

Eh2406:
Thank you all for doing work on this. I am going on vacation and will not be able to look at or think about this for 2 weeks.

One open design question is whether "`matches` implies `matches_prerelease`" is an axiom of our design. If it is then there are all kinds of (obscure) corner cases we need to track down and decide how to deal with. In general we should modify the new thing to meet our axioms. But given how obscure the cases are we should also consider modifying the existing behavior. Breakage will probably be only theoretical.

Please do not let my absence slow down this effort. When people feel it's ready, let's merge this PR. Unstable features are cheap to change if we change our mind.

tests/testsuite/precise_pre_release.rs
linyihai: My Bad. I shouldn't publish `0.1.3` in advance, the reason without `UPDATING` due to it alrealy is.

linyihai:
The forced update commits just updated the comments of the code .  

And I had updated the semantic in the PR description, exspecially for `<I.J.K`.

src/cargo/util/semver_eval_ext.rs
```diff
@@ -0,0 +1,595 @@
+//! This module provides the [`matches_prerelease`] to extends the semantic for how to match a pre-release version.
+//! In particular, this allows non pre-release version to update to Semver-compatible pre-release version.
```
epage, line 2: nit: when writing a doc comment, consider what the one-line explaination that `rustdoc` will show when browsing [`util`](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/util/index.html).

So maybe something like

```suggestion
//! Extend `semver::VersionReq` with  [`matches_prerelease`] which doesn't preclude pre-releases by default
```

src/cargo/util/semver_ext.rs
```diff
@@ -111,16 +121,13 @@ impl OptVersionReq {
...
     pub fn matches_prerelease(&self, version: &Version) -> bool {
-        if version.is_prerelease() {
-            let mut version = version.clone();
-            version.pre = semver::Prerelease::EMPTY;
-            return self.matches(&version);
+        if let OptVersionReq::Req(req) = self {
+            return req.matches_prerelease(version);
         }
+
         self.matches(version)
```
epage, line 131: nit: these two are mutually exclusive in their business logic (and its not just a guard statement).  For myself, I find using an `else` clarifies this relationship over an early return

tests/testsuite/precise_pre_release.rs
```diff
@@ -76,44 +76,92 @@ fn update_pre_release() {
 }
 
 #[cargo_test]
-fn update_pre_release_differ() {
+fn update_pre_release_range() {
```
epage, line 79: How much of these test changes can happen in a commit before the fix?  Separating out expanded coverage from behavior changes would make this a lot easier to review

src/cargo/util/semver_ext.rs
```diff
@@ -203,12 +210,12 @@ mod matches_prerelease {
         // https://rust-lang.github.io/rfcs/3493-precise-pre-release-cargo-update.html#version-ranges-with-pre-release-upper-bounds
         let cases = [
             //
-            ("1.2.3", "1.2.3-0", true), // bug, must be false
-            ("1.2.3", "1.2.3-1", true), // bug, must be false
+            ("1.2.3", "1.2.3-0", false),
+            ("1.2.3", "1.2.3-1", false),
             ("1.2.3", "1.2.4-0", true),
```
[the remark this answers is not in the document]
epage: Should we also add a proptest that verifies that every time `matches() == true` that `matches_prerelease() == true`?

tests/testsuite/precise_pre_release.rs
linyihai: I split the change into the first commit, which include the two cases, 

- `pre_release_should_unmatched` explains the drawback that needs to be fixed in propose semantic
- `pre_release_should_matched` can be treat as a normal test that must be meet

src/cargo/util/semver_ext.rs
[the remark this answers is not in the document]
linyihai:
```
("1.2.3", "2.0.0-0", _), // upper bound semantics
("=1.2.3-0", "1.2.3", false)
("=1.2.3-0", "1.2.3-0", true)
("=1.2.3-0", "1.2.4", false)  // upper bound semantics
(">=1.2.3-2, <1.2.3-4", "1.2.3-0", false)
(">=1.2.3-2, <1.2.3-4", "1.2.3-3", true)
(">=1.2.3-2, <1.2.3-4", "1.2.3-5", false)  // upper bound semantics
```
These tests had been added in the first commit before the fiexed applied.

src/cargo/util/semver_ext.rs
linyihai: Done. Thank you.

src/cargo/util/semver_eval_ext.rs
linyihai: I've changed it this way

src/cargo/util/semver_ext.rs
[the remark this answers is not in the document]
linyihai: > Should we also add a proptest that verifies that every time matches() == true that matches_prerelease() == true?

I don't have a clear idea of how to write a sufficiently complete test case.  The challenge is to find all the test cases for 'matches()=true' that you can, and to use different operators.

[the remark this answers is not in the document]
epage: a prop test wouldn't be a blocker

src/cargo/util/semver_eval_ext.rs
```diff
@@ -75,7 +75,7 @@ fn matches_exact(cmp: &Comparator, ver: &Version) -> bool {
 }
 
 // See https://github.com/dtolnay/semver/blob/69efd3cc770ead273a06ad1788477b3092996d29/src/eval.rs#L64-L88
-fn matches_greater(cmp: &Comparator, ver: &Version) -> bool {
+pub(crate) fn matches_greater(cmp: &Comparator, ver: &Version) -> bool {
```
epage, line 78: Unsure if this commit is worth it.  In theory, `semver_eval_ext` will be going away and we'll have to undo this.

linyihai: Well, this commit is not very important for this PR.  We can decide what to do with it later.

epage:
Thanks!

@bors r+

PR #14315, opened 2024-07-28
Fix warnings building tests on Windows

ehuss:
This fixes a few warnings when building cargo's testsuite on Windows due to some imports only being used inside some `cfg` elided tests.


weihanglo (approved):

weihanglo:
Nice catch. Thanks!

@bors r+

PR #14318, opened 2024-07-29
dont call wrap in a no-op source_id::with*

Eh2406:

When running resolution in parallel (which my pubgrub tests do but cargo does not) there can be a lot of contention on the lock for constructing new `source_id`. When investigating much of this is due to `without_precise` in `encodable_package_id` in `check_duplicate_pkgs_in_lockfile`. There are many ways to solve this, the simplest seems to be to return `self` if the requested modification made no difference.

All tests still pass and it's an internal re-factor.

In addition running all crates on crates.io through cargoes resolver in parallel on 190 cores went from >20k sec cpu time to ~10k.

### DIFF
--- a/src/cargo/core/source_id.rs
+++ b/src/cargo/core/source_id.rs
@@ -468,34 +468,51 @@ impl SourceId {
 
     /// Creates a new `SourceId` from this source with the given `precise`.
     pub fn with_git_precise(self, fragment: Option<String>) -> SourceId {
-        SourceId::wrap(SourceIdInner {
-            precise: fragment.map(|f| Precise::GitUrlFragment(f)),
-            ..(*self.inner).clone()
-        })
+        let precise = fragment.map(|f| Precise::GitUrlFragment(f));
+        if self.inner.precise == precise {
+            self
+        } else {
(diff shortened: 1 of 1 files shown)


epage:
I posted #14319 for the test failure.

src/cargo/core/source_id.rs
```diff
@@ -468,34 +468,51 @@ impl SourceId {
...
-        })
+        let precise = fragment.map(|f| Precise::GitUrlFragment(f));
+        if self.inner.precise == precise {
+            self
+        } else {
+            SourceId::wrap(SourceIdInner {
+                precise,
+                ..(*self.inner).clone()
+            })
+        }
```
epage, line 479: Should we factor this out with a `with_precise` that takes an `Option` so it'd work in every case?

epage (approved):

epage:
#14319 has been merged

src/cargo/core/source_id.rs
Eh2406: One of `with_precise_from` or `with_git_precise` will need to do one extra clone. But it seams worth it.

Issue #13889, opened 2024-05-09
assertion failed: edges.remove(&key)


fenollp:
### Problem

I'm working on my own `RUSTC_WRAPPER` and it may slow down some calls to `rustc` made by `cargo`. Toying with the amount of `--jobs=N`, I notice I can fail a `cargo` assertion: 
```shell
2024-05-09T07:46:34.9182036Z thread 'main' panicked at src/cargo/util/dependency_queue.rs:191:13:
2024-05-09T07:46:34.9183450Z assertion failed: edges.remove(&key)
2024-05-09T07:46:34.9184589Z note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2024-05-09T07:46:36.0109959Z ##[error]Process completed with exit code 101.
```

Now I'm not sure if I'm doing something wrong or if it's on `cargo`'s side?

### Notes

* the mentioned code: https://doc.rust-lang.org/beta/nightly-rustc/src/cargo/util/dependency_queue.rs.html#191
* the build logs https://productionresultssa6.blob.core.windows.net/actions-results/fe4adb8f-2924-4ee1-9d51-2ddc2c68553e/workflow-job-run-5d30e883-8d76-5db4-605e-068efef1550f/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-05-09T08%3A10%3A05Z&sig=eU4enoWXQZTBDKBXjwWMnbSXRC4rLMx1g3pJP0nBhC4%3D&sp=r&spr=https&sr=b&st=2024-05-09T08%3A00%3A00Z&sv=2021-12-02

### Version

```text
cargo 1.78.0 (54d8815d0 2024-03-26)
release: 1.78.0
commit-hash: 54d8815d04fa3816edc207bbc4dd36bf18014dbc
commit-date: 2024-03-26
host: x86_64-unknown-linux-gnu
libgit2: 1.7.2 (sys:0.18.2 vendored)
libcurl: 8.6.0-DEV (sys:0.4.72+curl-8.6.0 vendored ssl:OpenSSL/1.1.1w)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Ubuntu 23.10 (mantic) [64-bit]
```

weihanglo:
Could you share a minimal reproduction for this, and also a log without authentication to view? Without them it's a bit hard to come out with anything valuable.

fenollp:
Ah well sorry about that link
[...]

PR #14322, opened 2024-07-30
Go around 2 crash locations that trigger when using a RUSTC_WRAPPER

fenollp:

I am working on a `$RUSTC_WRAPPER` that you can find at https://github.com/fenollp/supergreen and, running it, I often trigger an assertion error or a panic from the unwrap. This usually triggers when calling cargo with `>=8` jobs and never with only `1` job.

My understanding is that this code I'm patching is concurrently updating `dep_map`, and the various slownesses my cargo plugin and/or rustc wrapper introduce confuse that code.
I believe a `RwLock` could also be used around `dep_map`. I tried to avoid using a lock but you may prefer it.

Here is a couple CI jobs that show the failure:
* https://github.com/fenollp/supergreen/actions/runs/10158581256/job/28090986965?pr=54
```
2024-07-30T08:09:59.2200020Z thread 'main' panicked at src/cargo/util/dependency_queue.rs:190:56:
2024-07-30T08:09:59.2236785Z called `Option::unwrap()` on a `None` value
2024-07-30T08:09:59.2237867Z note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
* https://github.com/fenollp/supergreen/actions/runs/10125251212/job/28040951263?pr=54
```
2024-07-27T17:28:04.9303457Z thread 'main' panicked at src/cargo/util/dependency_queue.rs:191:13:
2024-07-27T17:28:04.9304284Z assertion failed: edges.remove(&key)
2024-07-27T17:28:04.9305150Z note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This uses this CI code:
* https://github.com/fenollp/supergreen/blob/480fc800b8107ff44aa50cd9700fea91160dfd58/.github/workflows/self.yml#L199
* https://github.com/fenollp/supergreen/blob/480fc800b8107ff44aa50cd9700fea91160dfd58/.github/workflows/clis.yml#L508

Basically, it compiles my plugin and my wrapper and uses them to call cargo 1.80 to build some code.

### Additional information

I believe this should be easily reproducible simply by adding a `sleep` before executing `rustc` when using `cargo` to build things.

I'd love some directions on this issue. It's causing me some troubles developing my things :)

### DIFF
--- a/src/cargo/util/dependency_queue.rs
+++ b/src/cargo/util/dependency_queue.rs
@@ -187,10 +187,11 @@ impl<N: Hash + Eq + Clone, E: Eq + Hash + Clone, V> DependencyQueue<N, E, V> {
         let key = (node.clone(), edge.clone());
         let mut result = Vec::new();
         for dep in reverse_deps.iter() {
-            let edges = &mut self.dep_map.get_mut(dep).unwrap().0;
-            assert!(edges.remove(&key));
-            if edges.is_empty() {
-                result.push(dep);
+            if let Some((edges, _)) = &mut self.dep_map.get_mut(dep) {
+                edges.remove(&key);
+                if edges.is_empty() {
+                    result.push(dep);
+                }
             }
         }
         result


weihanglo:
> My understanding is that this code I'm patching is concurrently updating `dep_map`, and the various slownesses my cargo plugin and/or rustc wrapper introduce confuse that code.

No. It should be run only on the main thread in Cargo. From [`drain_the_queue`](https://github.com/rust-lang/cargo/blob/11dccd1df09b0f50706d72f6e886b67f1f00b732/src/cargo/core/compiler/job_queue/mod.rs#L532-L537) down to [`handle_event`](https://github.com/rust-lang/cargo/blob/11dccd1df09b0f50706d72f6e886b67f1f00b732/src/cargo/core/compiler/job_queue/mod.rs#L784), to each message (from a spawn worker thread) processing.

The assertion, while not clear, help guard that Cargo handles a node only once. If you have bumped into any assertion failure, there might be a more severe bug. Instead of allowing this, maybe we should find which kind of `RUSTC_WRAPPER` was failing.

Could you create a new issue and provide some minimal reproducible examples with your `RUSTC_WRAPPER`?

fenollp:
Oh ok, I'm very surprised this code isn't multi-threaded. I can definitely say I reproduce my issue with `cargo --jobs=8` and not with `cargo --jobs=1`.

weihanglo:
If I remember correctly you've filed a similar issue <https://github.com/rust-lang/cargo/issues/13889> and said that the tool delays some rustc invocation. This is a suspect behavior to me. I don't have time walking through the entire repository. If you can point to a portion of code that contributes the delay logic, or create a minimal one that proves its a bad behavior in Cargo, we can then look into it together.

In the meanwhile, I will close this. Thank you.

Issue #5707, opened 2018-07-11
RFE: add option to specify path to Cargo.lock


ignatenkobrain:
In Fedora we package all crate sources (including tests) in `/usr/share/cargo/registry` which is not user-writable. However, we would like to run tests for them after installation. There is already option for `target-dir` and it would be very nice to get option to put Cargo.lock in some other directory so that we could run `cargo test` from there.

ignatenkobrain:
@alexcrichton any idea how the option could be named? I'm thinking about making a patch.

alexcrichton:
Perhaps `--lockfile-path` to mirror `--manifest-path`?

epage:
We talked about this in the Cargo team meeting and are open to people experimenting with this
- CLI flag only, no config, no manifest support
- Name: `--lockfile-path <path>`
- Behind `-Zunstable-options` and docs
  - The flag's docs / `--help` should have `(unstable)` in them
- Looks like it belongs on all commands that take `--manifest-path` except `cargo locate-project`
- Must require the `file_name` be `Cargo.lock` (like we do for `--manifest-path`).  We can always re-evaluate this in the future
- Main question is how this should work if the lockfile or a parent directory is not present
  - I would lean towards "create them" rather than "error"

Ifropc:
@epage sorry it took me some time, but the PR is ready for review.
I followed the guideline you posted in [your comment](https://github.com/rust-lang/cargo/issues/5707#issuecomment-2231276362) (thank you very much!)
1 slight difference is I haven't added `--lockfile-path` on `cargo install` yet, as it's not one of the command that takes the `--manifest-path`. 
I think the main usage of the `lockfile-path` is to enable folks to run commands on readonly folders (and adding it to almost all --`manifest-path` commands for consistency is a good idea). However, `cargo install` doesn't really fit into this category, so I doubt there's going to be any demand for that. 
That being sad, adding it is very easy (and to be completely honest I just completely missed your suggestion up until I opened the PR), so if you think it'd be a nice addition it's not a problem to add it to the set of supported commands.

weihanglo:
@Ifropc No worry. As `--lockfile-path` is not going to be insta-stable, that seems fine to me. We can always discuss use cases later.

weihanglo:
Most major tasks in https://github.com/rust-lang/cargo/issues/5707#issuecomment-2231276362 were addressed via #14326. Below are remaining tasks:

[...]

PR #14326, opened 2024-07-31
Add `--lockfile-path` flag

Ifropc:
This change implements a new `--lockfile-path` proposed in #5707 .

Functionality added:
- Add `--lockfile-path <PATH>` to all commands that support `manifest-path` with exception of:
   - `locate-project` (doesn't use lock file)
   - `verify-project` (is deprecated)
   - `read-manifest` (doesn't use lock file)
- Behind -Zunstable-options and docs
   - The flag's docs / `--help` has (unstable) in them
- `<PATH>` must end with `Cargo.lock`. If specified path doesn't exist (or parent director(ies), create all the parent directories and the lockfile itself

Implementation TLDR: add `requested_lockfile_path` into `Workspace` and set it on `workspace(gctx)` call (setting from the context)
Update `lockfile.lock_root()` to respect `requested_lockfile_path` (if set)
Add test cases covering all affected commands. Tested creating lockfile, reading lockfile, overriding default (`./Cargo.lock`) lockfile, symlink tests. Extra tests for package to make sure pinned versions from path's lockfile are respected (i.e. double check correct lockfile is used)
I doubt this flag will be used for any command that's not read-only, but I tried to cover all the commands.


src/doc/man/includes/options-lockfile-path.md
```diff
@@ -0,0 +1,12 @@
+{{#option "`--lockfile-path` _PATH_"}}
+Changes the path of the lockfile from the default (`./Cargo.lock`) to _PATH_. _PATH_ must end with
```
weihanglo, line 2: The default is `Cargo.lock` at the workspace root.

src/cargo/util/command_prelude.rs
```diff
@@ -1,6 +1,25 @@
+use std::ffi::{OsStr, OsString};
```
weihanglo, line 1: Although I love this "standard" way to arrange import orders, could we follow atomic commit pattern and leave off this to a separate commit or pull request?

src/cargo/util/command_prelude.rs
```diff
@@ -295,6 +297,15 @@ pub trait CommandExt: Sized {
         )
     }
 
+    fn arg_lockfile_path(self) -> Self {
+        self._arg(
+            opt("lockfile-path", "Path to the Cargo.lock file (unstable)")
```
weihanglo, line 302: What about this for consistency?

```suggestion
            opt("lockfile-path", "Path to Cargo.lock (unstable)")
```

src/cargo/util/command_prelude.rs
```diff
@@ -295,6 +297,15 @@ pub trait CommandExt: Sized {
...
+                .value_name("FILE")
+                // TODO: seemed to be the best option, but maybe should be something else?
```
weihanglo, line 304: I feel it makes sense to put `--manifest-path` and `--lockfile-path` side by side.

```suggestion
```

src/cargo/util/command_prelude.rs
```diff
@@ -986,6 +1003,64 @@ pub fn root_manifest(manifest_path: Option<&Path>, gctx: &GlobalContext) -> Carg
...
+pub fn lockfile_path(
+    lockfile_path: Option<&Path>,
+    gctx: &GlobalContext,
+) -> CargoResult<Option<PathBuf>> {
+    let path;
+
+    if let Some(lockfile_path) = lockfile_path {
+        if !gctx.cli_unstable().unstable_options {
+            bail!("`--lockfile-path` option requires `-Zunstable-options`")
+        }
```
weihanglo, line 1015:
```suggestion
        gctx.cli_unstable()
            .fail_if_stable_opt("--lockfile-path", 5707)?;

```

src/cargo/util/command_prelude.rs
```diff
@@ -986,6 +1003,64 @@ pub fn root_manifest(manifest_path: Option<&Path>, gctx: &GlobalContext) -> Carg
...
+
+        if lockfile_path.is_absolute() {
+            path = lockfile_path.to_path_buf();
+        } else {
+            path = gctx.cwd().join(lockfile_path);
+        }
```
weihanglo, line 1021: This should be sufficient.

* [`Path::join`](https://doc.rust-lang.org/1.80.0/std/path/struct.Path.html#method.join)  replaces the original path if the given is absolute.
* Place the variable declaration together when binding.


```suggestion
        let path = gctx.cwd().join(lockfile_path);
```

src/cargo/util/command_prelude.rs
```diff
@@ -986,6 +1003,64 @@ pub fn root_manifest(manifest_path: Option<&Path>, gctx: &GlobalContext) -> Carg
...
+        if path.is_dir() {
+            bail!(
+                "lockfile path `{}` is a directory but expected a file",
+                lockfile_path.display()
+            )
+        }
+        if !path.exists() {
+            // Root case should already be covered above
+            let parent_path = lockfile_path
+                .parent()
```
weihanglo, line 1038: If it is already covered, should we use `expect("...")` instead?

src/cargo/util/command_prelude.rs
```diff
@@ -986,6 +1003,64 @@ pub fn root_manifest(manifest_path: Option<&Path>, gctx: &GlobalContext) -> Carg
     }
 }
 
+pub fn lockfile_path(
+    lockfile_path: Option<&Path>,
+    gctx: &GlobalContext,
+) -> CargoResult<Option<PathBuf>> {
+    let path;
+
+    if let Some(lockfile_path) = lockfile_path {
```
weihanglo, line 1012: Use `let-else` so we can have less indentation levels.

```suggestion
    let Some(lockfile_path) = lockfile_path else {
        return Ok(None);
    };
```

weihanglo:
Thanks for the contribution btw!

epage:
Thanks for your work on this!

The commit history looks like its from how this was developed over time.  Do you think you could clean it up?  Maybe an example set of commits
1. Refactor commits for making it easier to implement
2. Commit that adds any new tests, showing the current behavior (ie running `cargo test` on this commit should pass)
3. Feature commit w/ test updates so they continue passing

weihanglo:
> I'm guessing I should just squash + commits together? From my understanding on each separate commit cargo test must always pass, and each separate commit should represent a block of added functionality (or a new test scenario)

Squashing all commtis into one isn't needed. We follow atomic commits as possible as we could. It makes reviews smoother and easier, increasing the chance of PR being merged.

Usually for bugfix I would expect one commit shows the problemtic behavior (and passes CI by assertion the incorrect behavior), and the other commit fixes both the bad behavior and test. You can read [some examples](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request) here in the contributing guide.

If you feel like some refactors can be break out to a separate pull request. Feel free to do so!

epage:
To expand on that, generally atomic commits are fix/feature, test updates, documentation updates, etc.   We only break from this in more extreme cases (#14141).

That said, something we encourage, as shown in those examples, is adding tests in a commit before feature / fix work that shows the existing behavior (ie tests must pass).  Its a little odd when its something like a new CLI flag as the tests will show just "unknown command line flag".  What you can instead do is write the tests without the flag and then in the commit with the flag added, add the flag to the tests.  This will show to reviewers how the behavior changed with the introduction of this flag.

If you are doing TDD, you might be intentionally having failing tests.  That is understandable but separate from how to organize commits.  Have the "new test" commit then locally make the tests fail and work on the new behavior in that situation and then commit when they start passing again.  Of course, actual development can be a lot messier than that and commits might need to be re-organized after the fact and thats ok too.  Sometimes, the re-organization is too messy and not worth it.  That is the case where a mixture of squashing and leaving things as-is is best.

Ifropc:
Thanks Ed, I think I'll try a mixture of squashing and re-organizing, just to get the general idea and not have messy history again

src/cargo/util/command_prelude.rs
Ifropc: This wasn't an intended change, and simply auto-refactoring of the imports. I also (usually) prefer not to mess with imports to avoid merge conflicts. Rolled back this change :+1:

src/doc/man/includes/options-lockfile-path.md
Ifropc: Good catch. It was under assumption that we already running a command from the workspace root 😄

src/cargo/util/command_prelude.rs
Ifropc: Good catch, thanks!

src/cargo/util/command_prelude.rs
```diff
@@ -986,6 +1003,64 @@ pub fn root_manifest(manifest_path: Option<&Path>, gctx: &GlobalContext) -> Carg
...
+            })?;
+
+            if !exists {
+                fs::create_dir_all(parent_path).with_context(|| {
+                    format!(
+                        "Failed to create lockfile-path parent directory {}",
+                        parent_path.display()
+                    )
+                })?
+            }
```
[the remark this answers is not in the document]
Ifropc: Missed that, thanks! Def would want to be consistent across the codebase

src/cargo/core/workspace.rs
```diff
@@ -647,6 +651,14 @@ impl<'gctx> Workspace<'gctx> {
         self
     }
 
+    pub fn requested_lockfile_path(&self) -> Option<&PathBuf> {
+        self.requested_lockfile_path.as_ref()
+    }
+
+    pub fn set_requested_lockfile_path(&mut self, path: Option<PathBuf>) {
+        self.requested_lockfile_path = path;
+    }
```
weihanglo, line 660: Could we make `lockfile_path` work in the similar pattern like `target_dir`? That is,

* `pub fn lockfile_path()` returns a  `FileSystem`. By default check `Workspace.requested_lockfile_path`, which is `Option<FileSystem>`.
* If none, calls `fn default_lockfile_path()` to get `ws.root.join("Cargo.lock")`.

src/cargo/ops/lockfile.rs
```diff
@@ -247,6 +248,14 @@ fn emit_package(dep: &toml::Table, out: &mut String) {
 }
 
 fn lock_root(ws: &Workspace<'_>) -> Filesystem {
+    if let Some(requested) = ws.requested_lockfile_path() {
```
weihanglo, line 251: If we had `Workspace::lockfile_path()`, this function can be rewritten to

```rust
fn lock_root(...) {
    if ws.root_maybe().is_embedded() {
        ws.target_dir()
    } else {
        ws.lockfile_path().parent().expect("lockfile path exists")
    }
}
```
What do you think?

src/bin/cargo/commands/fix.rs
```diff
@@ -71,8 +71,7 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
     // Unlike other commands default `cargo fix` to all targets to fix as much
     // code as we can.
     let root_manifest = args.root_manifest(gctx)?;
-    let mut ws = Workspace::new(&root_manifest, gctx)?;
-    ws.set_resolve_honors_rust_version(args.honor_rust_version());
+    let ws = args.workspace(gctx)?;
```
weihanglo, line 74: Using `args.workspace()` here have `cargo fix` to respect [`-Zavoid-dev-deps`](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html?highlight=avoid-dev-d#avoid-dev-deps), which previous it didnt.

I think it is intentionally, as `cargo fix` by default also builds test targets, which requires dev-dependencies.

src/cargo/core/workspace.rs
```diff
@@ -103,6 +103,9 @@ pub struct Workspace<'gctx> {
     // file. This is set for `cargo install` without `--locked`.
     ignore_lock: bool,
 
+    // Requested path of the lockfile (i.e. passed as the cli flag)
+    requested_lockfile_path: Option<PathBuf>,
```
weihanglo, line 107: Use doc comment is preferrable.

(Just notice some fields here should switch to doc comemnt as well)

```suggestion
    /// Requested path of the lockfile (i.e. passed as the cli flag)
    requested_lockfile_path: Option<PathBuf>,
```

src/cargo/ops/cargo_package.rs
```diff
@@ -248,7 +248,12 @@ pub fn package(ws: &Workspace<'_>, opts: &PackageOpts<'_>) -> CargoResult<Vec<Fi
     }
     let pkgs = ws.members_with_features(specs, &opts.cli_features)?;
 
-    if ws.root().join("Cargo.lock").exists() {
+    let default_lockfile = ws.root().join("Cargo.lock");
```
weihanglo, line 251: Call `ws.lockfile_path()` and we don't need to duplicate ourselves.
See https://github.com/rust-lang/cargo/pull/14326#discussion_r1702450833

src/cargo/util/command_prelude.rs
```diff
@@ -986,6 +1001,41 @@ pub fn root_manifest(manifest_path: Option<&Path>, gctx: &GlobalContext) -> Carg
...
+    let Some(lockfile_path) = lockfile_path else {
+        return Ok(None);
+    };
+
+    gctx.cli_unstable()
+        .fail_if_stable_opt("--lockfile-path", 5707)?;
+
+    let path = gctx.cwd().join(lockfile_path);
+
+    if !path.ends_with(LOCKFILE_NAME) && !crate::util::toml::is_embedded(&path) {
```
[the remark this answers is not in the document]
Ifropc: I think it _should_ work. I initially added a basic test for it https://github.com/Ifropc/cargo/commit/164aae02d27339c6e5f33e1a4c619e0c4ed79bef but removed it since (not sure why actually). 
Do you think we would need a better test coverage for `-Zscript`? If yes, I'd prefer to cover it with tests in a separate PR. If no, I can just bring the test back.
(FYI: there's a missing `.no_manifest()` on 
```rust
  let p = project()
        .no_manifest()
```  
But it still works with adding `.no_manifest()`)

src/cargo/util/command_prelude.rs
```diff
@@ -986,6 +1001,41 @@ pub fn root_manifest(manifest_path: Option<&Path>, gctx: &GlobalContext) -> Carg
...
+        bail!(
+            "lockfile path `{}` is a directory but expected a file",
+            lockfile_path.display()
+        )
+    }
+    if !path.exists() {
+        // Root case should already be covered above
+        let parent_path = lockfile_path.parent().expect("lockfile path can't be root");
+
+        paths::create_dir_all(parent_path)?;
```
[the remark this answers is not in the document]
Ifropc: Hmm that's fair. I can move the logic of creating directory into `write_pkg_lockfile`. 
In that case, I think we should bring back the logic of validating the path, while creating the dirs on writing the lockfile.
WDYT?

src/cargo/ops/lockfile.rs
```diff
@@ -1,19 +1,20 @@
-use std::io::prelude::*;
-
 use crate::core::{resolver, Resolve, ResolveVersion, Workspace};
 use crate::util::errors::CargoResult;
 use crate::util::Filesystem;
+use std::io::prelude::*;
```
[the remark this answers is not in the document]
Ifropc: Sorry it's my IDE auto-imports stuff again. Let me disable it entirely

src/cargo/core/workspace.rs
Ifropc: So to clarify, we basically want to move logic on of `cargo::ops::lockfile::lock_root` into worspace instead, right? I was originally thinking about it, but wasn't sure if I should move that piece of code around

src/cargo/ops/lockfile.rs
Ifropc: Yep that's what I was originally thinking of doing too actually! Let me fix that

tests/testsuite/lockfile_path.rs
weihanglo, line 1: Just found I missed reviewing this file 😞.

It was folded by GitHub. I'll take a look later!

src/cargo/ops/cargo_package.rs
```diff
@@ -248,7 +248,12 @@ pub fn package(ws: &Workspace<'_>, opts: &PackageOpts<'_>) -> CargoResult<Vec<Fi
...
+    if ws
+        .requested_lockfile_path()
+        .unwrap_or_else(|| &default_lockfile)
```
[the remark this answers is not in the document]
Ifropc: Agree on that, will do some refactoring, good catch!

src/cargo/ops/lockfile.rs
Ifropc: Moved `lock_root()` all together into `worspace`, to keep current behavior. It also looks like the idea of `Filesystem` is to hold directory path to work with files inside it.
LMK if that's what you were thinking 
132ba16779eb2c28d43796a82160ce02348158d1

src/cargo/util/command_prelude.rs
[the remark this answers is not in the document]
weihanglo: I mean. `path` not ending with `Cargo.lock` is already emitting error. Why we need to check if it is not embedded (not ends with `.rs`)?

For `-Zscript` support, I think have a test for `cargo run` with both `--lockfile-path` and `--manifest-path` provided should cover it.

weihanglo (commented):
Sorry for the delay and we are making progress!

src/cargo/util/command_prelude.rs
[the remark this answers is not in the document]
Ifropc: Yeah that makes sense, I think I got accidentally mixed logic for lockfile directory in there, not sure why this check is still there, good catch! I will also cover it with a test case ^

tests/testsuite/lockfile_path.rs
```diff
@@ -0,0 +1,420 @@
...
+        let lockfile_path_argument = "mylockfile/Cargo.lock";
+        let p = project()
+            .file("Cargo.toml", TOML)
+            .file("src/main.rs", "fn main() {}")
+            .file("mylockfile/Cargo.lock", PINNED_VERSION)
+            .build();
+
+        make_basic_command(&mut p.cargo("package"), lockfile_path_argument.to_string()).run();
+
+        assert!(!p.root().join("Cargo.lock").is_file());
```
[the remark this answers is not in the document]
Ifropc: Haha fair point 😆

tests/testsuite/lockfile_path.rs
```diff
@@ -0,0 +1,420 @@
...
+    if lockfile_must_exist(command) {
+        return;
+    }
+
+    let lockfile_path_argument = "mylockfile/is/burried/Cargo.lock";
+    let p = make_project().build();
+    let registry = RegistryBuilder::new().http_api().http_index().build();
+
+    make_execs(&mut p.cargo(command), lockfile_path_argument.to_string())
+        .replace_crates_io(registry.index_url())
```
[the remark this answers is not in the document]
Ifropc: Yeah I went overboard with dryness that hurt readability and made tests to be overloaded. Let me simplify it!

tests/testsuite/lockfile_path.rs
```diff
@@ -0,0 +1,420 @@
...
+fn make_basic_command(execs: &mut Execs, lockfile_path_argument: String) -> &mut Execs {
+    return execs
+        .masquerade_as_nightly_cargo(&["unstable-options"])
+        .arg("-Zunstable-options")
+        .arg("--lockfile-path")
+        .arg(lockfile_path_argument);
+}
+
+fn lockfile_must_exist(command: &str) -> bool {
+    return command == "pkgid" || command == "publish" || command == "package";
```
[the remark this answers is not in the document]
Ifropc: This will go away when I rewrite the test structure!

tests/testsuite/lockfile_path.rs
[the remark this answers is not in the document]
Ifropc: I went with dry approach mainly because I'm very new to codebase and wanted to cover as much as possible to make sure I'm not breaking everything (and copy-pasting same test case for each command felt like something I don't want to do). Now I'm 100% re-thinking this approach,
> You don't need to test every command in every aspect. It is sufficient to just pick one command and test it exhaustively (unless some command diverges).

I agree that we should remove testing every command feels absolutely unnecessary to do that and introduces bunch of very similar tests, I will leave only "unique" once (that require special setup, like publish); `manifest` and `run`

> or have a loop over them in one test

I think it's better to have separate `#[test]` because the way we setup project or command invocation is different from command to command.

> Just need to make sure we're using with_{stderr,stdout}_data with the snapshot str! macro (we are migrating to snapthos as possible as we could

Got it! I will check a bit more into the reasoning  

> Personally, I would drop declarative macros

Yep agree on that too. Really didn't want to copy-paste the same cases during development 😆 

 > For code sharing, it's better to write test helpers for dumb things like project creations, and try to keep them as a whole for readability, don't break them into small pieces just for sharing.

I think I personally prefer test helpers for stuff like initializing the test environment that is same between the tests, but I do know a lot of people who just prefer copy-pasting it everywhere for readability. If we are making tests wet I think I'll just get rid away of helpers too, so there's a clear test scenario

Ifropc:
Thank you for the patience Weihang! I usually try to get back on the comments and resolve them ASAP but it's getting late here and I'm quite busy over the weekend. Will try to fix the remaining pieces early next week (Monday)

tests/testsuite/lockfile_path.rs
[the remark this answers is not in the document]
weihanglo: BTW I didn't mean copy-pasting everything. We need to weigh on different aspects. 

Another example is that assertions are easier to extract into shared code, like
https://github.com/rust-lang/cargo/blob/b66cad8038bd0a94164a4ecd1b2c12075a42610d/tests/testsuite/vendor.rs#L154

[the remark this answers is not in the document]
Ifropc: Thanks for this example, I'll do another iteration on the tests and please feel free to point out to anything that you think should be improved!

src/cargo/core/workspace.rs
Ifropc: > (Just notice some fields here should switch to doc comemnt as well)

Addressed it in #14397 before it is forgotten

tests/testsuite/lockfile_path.rs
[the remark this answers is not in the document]
Ifropc: Refactored tests, should be in a better shape now. LMK if you still think we should trim something down

weihanglo:
I am not sure if this is ready for another round of review, so

@rustbot author

When ready, please click the GitHub “Re-request review” button, or add a comment `@rustbot review`.

src/cargo/util/command_prelude.rs
```diff
@@ -986,6 +1001,44 @@ pub fn root_manifest(manifest_path: Option<&Path>, gctx: &GlobalContext) -> Carg
...
+        let parent_path = lockfile_path.parent().expect("lockfile path can't be root");
+
+        // Fail early if parent path existence can't be verified (e.g. due to permissions)
+        parent_path.try_exists().with_context(|| {
+            format!(
+                "failed to fetch lock file's parent path metadata {}",
+                parent_path.display()
+            )
+        })?;
+    }
```
epage, line 1037: I've not been following the review too closely lately but it looks like this errors if the parent directory is not present

From #5707

> Main question is how this should work if the lockfile or a parent directory is not present
>
> > I would lean towards "create them" rather than "error"

Could you call out in the PR description why you made that behavior decision?

src/cargo/util/command_prelude.rs
```diff
@@ -986,6 +1001,44 @@ pub fn root_manifest(manifest_path: Option<&Path>, gctx: &GlobalContext) -> Carg
...
+        )
+    }
+    if !path.exists() {
+        // Root case should already be covered above
+        let parent_path = lockfile_path.parent().expect("lockfile path can't be root");
+
+        // Fail early if parent path existence can't be verified (e.g. due to permissions)
+        parent_path.try_exists().with_context(|| {
+            format!(
+                "failed to fetch lock file's parent path metadata {}",
```
epage, line 1033: If we keep this error description, we should frame it in terms of the user (which the comment focuses on) rather than the implementation

src/cargo/util/command_prelude.rs
```diff
@@ -986,6 +1001,44 @@ pub fn root_manifest(manifest_path: Option<&Path>, gctx: &GlobalContext) -> Carg
...
+    };
+
+    gctx.cli_unstable()
+        .fail_if_stable_opt("--lockfile-path", 5707)?;
+
+    let path = gctx.cwd().join(lockfile_path);
+
+    if !path.ends_with(LOCKFILE_NAME) {
+        bail!("the lockfile-path must be a path to a {LOCKFILE_NAME} file")
+    }
```
epage, line 1019: nit: a part of me would prefer to be more explicit that this is about `file_name`

src/cargo/util/command_prelude.rs
Ifropc: Hi Ed, we are creating parent directory in [write_pkg_lockfile](https://github.com/rust-lang/cargo/blob/76d75059091923fc00a4749090d93151d1d418f7/src/cargo/ops/lockfile.rs#L87-L89). I added this check so we can fail early if we can't check for parent path and know it can't be created (e.g. loop symlinks).
But now that I think of it, if we have write only permission on a directory (`-w-`), then this check will fail when we still can create directory. Let me actually remove it

src/cargo/util/command_prelude.rs
Ifropc: [Removed the check](https://github.com/rust-lang/cargo/pull/14326#discussion_r1717333464)

Ifropc:
Thanks @weihanglo , yes it should be ready. I'm more available now and will be able to address comments the same day!

@rustbot review

tests/testsuite/lockfile_path.rs
```diff
@@ -0,0 +1,510 @@
...
+    let args = args.collect::<Vec<_>>();
+    println!("bin: {bin}");
+    println!("args: {args:?}");
+}
+
+#[test]
+fn test () {}
+"#;
+    let lockfile_path_argument = "mylockfile/Cargo.lock";
+    let p = project().file("src/main.rs", ECHO_SCRIPT).build();
```
weihanglo, line 496: I guess we don't need any content inside the main function, since asserting them is the job of `script.rs` test suite.

```suggestion
    let p = project().file("src/main.rs", "fn main()").build();
```

tests/testsuite/lockfile_path.rs
```diff
@@ -0,0 +1,510 @@
...
+version = "0.5.0"
+authors = ["wycats@example.com"]
+edition = "2015"
+
+[[bin]]
+
+name = "test_foo"
+
+[dependencies]
+hello = "1.0.0"
```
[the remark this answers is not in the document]
Ifropc: Yep that's true, I was using actual dependency just to see everything is working and forgot to replace it with mock dependency. Let me fix that. Good catch!

tests/testsuite/lockfile_path.rs
```diff
@@ -0,0 +1,510 @@
...
+"#
+        ]]
+    } else {
+        str![[
+            r#"[WARNING] please specify `--format-version` flag explicitly to avoid compatibility problems
+[ERROR] failed to create directory `[ROOT]/foo/somedir/link`
+
+Caused by:
+  Cannot create a file when that file already exists. (os error 183)
+
```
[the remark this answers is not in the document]
Ifropc: Good point, I didn't know we can do that 👍🏼

tests/testsuite/lockfile_path.rs
```diff
@@ -0,0 +1,510 @@
...
+    let dst = "dst";
+    let src = "somedir/link";
+    let lockfile_path_argument = format!("{src}/Cargo.lock");
+
+    let p = make_project().symlink_dir(dst, src).build();
+
+    fs::create_dir(p.root().join("dst"))
+        .unwrap_or_else(|e| panic!("could not create directory {}", e));
+    assert!(p.root().join(src).is_dir());
```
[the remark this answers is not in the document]
Ifropc: I think I just copied from the main code I had, fixed!

Ifropc:
Thanks for your patience @weihanglo !
@rustbot review

weihanglo (approved):
Thanks @Ifropc for your patience and contribution! 

This is in a good state to merge. It is still unstable so if any bug emerges we can always fix later!

weihanglo:
@Ifropc 

Here are some tasks we can do in follow-ups, if you're still interested :)

https://github.com/rust-lang/cargo/issues/5707#issuecomment-2293655336

For feature gated tests, we should have

* One without `-Zunstable-options` and fails.
* One with `-Zunstable-options but not with `masquerade_as_nightly_cargo`.

For reference: https://github.com/rust-lang/cargo/blob/d3e84f01ff48d19892035b77db7e9c4eda87bf1f/tests/testsuite/build.rs#L298-L319

epage:
If we're going back and making changes, please rename `make_execs` to something that describes what it does, rather than based on its return type.  Personally, I would recommend removing it.  It adds an extra layer of indirection (opaqueness) for little benefit.

PR #14329, opened 2024-07-31
fix(toml): Resolve regression from toml_edit 0.22.18

epage:
A fuzzer found a stack overflow with very large amounts of input. Been a while since I touched the parser and introduced recursion without thinking about it, whoops.


weihanglo (approved):
Thanks.

For references, here are related pull requests:

* https://github.com/toml-rs/toml/pull/775
* https://github.com/toml-rs/toml/pull/774

PR #14338, opened 2024-08-01
Bail publish job before packaging and upload (passing tests)

### LINKED ISSUE #3662: Fail fast when package version is already published

torhovland:
Builds on #13501 with a fix that makes the tests pass. Fixes #3662.

### DIFF
--- a/src/cargo/ops/registry/publish.rs
+++ b/src/cargo/ops/registry/publish.rs
@@ -161,9 +161,10 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
 
     if !duplicate_query.is_empty() {
         bail!(
-            "crate {} already has version {}. Aborting publish.",
+            "crate {}@{} already exists on {}",
             pkg.name(),
-            pkg.version()
+            pkg.version(),
+            source.describe()
         );
     }
 
--- a/src/cargo/ops/registry/publish.rs
+++ b/src/cargo/ops/registry/publish.rs
@@ -28,6 +28,7 @@ use crate::ops;
 use crate::ops::PackageOpts;
 use crate::ops::Packages;
 use crate::sources::source::QueryKind;
+use crate::sources::source::Source;
 use crate::sources::SourceConfigMap;
 use crate::sources::CRATES_IO_REGISTRY;
 use crate::util::auth;
@@ -129,7 +130,7 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
         val => val,
     };
     let source_ids = super::get_source_id(opts.gctx, reg_or_index.as_ref())?;
-    let mut registry = super::registry(
+    let (mut registry, mut source) = super::registry(
         opts.gctx,
         &source_ids,
         opts.token.as_ref().map(Secret::as_deref),
@@ -141,9 +142,7 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
 
     // Bail before packaging and uploading if same version already exists in the registry
 
-    let mut source = SourceConfigMap::empty(opts.gctx)?.load(reg_ids.original, &HashSet::new())?;
-
-    let query = Dependency::parse(pkg.name(), Some(&ver), reg_ids.original)?;
+    let query = Dependency::parse(pkg.name(), Some(&ver), source_ids.replacement)?;
 
     let _lock = opts
         .gctx
--- a/src/cargo/ops/registry/publish.rs
+++ b/src/cargo/ops/registry/publish.rs
@@ -139,6 +139,35 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
     )?;
     verify_dependencies(pkg, &registry, source_ids.original)?;
 
+    // Bail before packaging and uploading if same version already exists in the registry
+
+    let mut source = SourceConfigMap::empty(opts.gctx)?.load(reg_ids.original, &HashSet::new())?;
+
+    let query = Dependency::parse(pkg.name(), Some(&ver), reg_ids.original)?;
+
+    let _lock = opts
+        .gctx
+        .acquire_package_cache_lock(CacheLockMode::DownloadExclusive)?;
+
+    let duplicate_query = loop {
+        match source.query_vec(&query, QueryKind::Exact) {
+            std::task::Poll::Ready(res) => {
+                break res?;
+            }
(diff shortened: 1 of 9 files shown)


torhovland:
@hi-rustin, you recently [refactored](https://github.com/rust-lang/cargo/commit/416a2b96fcb19689091b079680d41e4291843923#diff-277403efaf7f6382e405c12bf52e33337060618739c801f4e8690f3cb10d73f6) away the return tuple in `fn cargo::ops::registry::registry()`. Any thoughts about reintroducing a return tuple here?

epage:
> @hi-rustin, you recently [refactored](https://github.com/rust-lang/cargo/commit/416a2b96fcb19689091b079680d41e4291843923#diff-277403efaf7f6382e405c12bf52e33337060618739c801f4e8690f3cb10d73f6) away the return tuple in `fn cargo::ops::registry::registry()`. Any thoughts about [reintroducing](https://github.com/rust-lang/cargo/pull/14338/files#diff-bc1a08bfe2fe76626f5fee55bfadc3df43043860d436c26a25a01a7fd613aff9) a return tuple here?

The goal with that refactor is less about removing the tuple and more about pulling work out of that function for reuse, like @jneem's recent `transmit` -> `prepare_transmit` + `transmit` refactor.  Adding back in a tuple to access some of the other state is reasonable.

src/cargo/ops/registry/publish.rs
```diff
@@ -139,6 +139,35 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
...
+            }
+            std::task::Poll::Pending => source.block_until_ready()?,
+        }
+    };
+
+    drop(_lock);
+
+    if !duplicate_query.is_empty() {
+        bail!(
+            "crate {} already has version {}. Aborting publish.",
```
epage, line 164: nit: what do you think of
```rust
"cannot publish {}@{} as that version already exists on {registry}",
```
or
```rust
"a {}@{} already exists on {registry}",
```

tests/testsuite/publish.rs
```diff
@@ -130,6 +130,54 @@ You may press ctrl-c to skip waiting; the crate should be available shortly.
...
+        .add_responder("/index/3/f/foo", move |req, server| {
+            let mut lock = arc.lock().unwrap();
+            *lock += 1;
+            if *lock <= 1 {
+                server.not_found(req)
+            } else {
+                server.index(req)
+            }
+        })
+        .build();
```
epage, line 148: I can't remember, was there a reason we couldn't publish the package like we normally do in tests?

stupendoussuperpowers:
As for the 2 failing tests - 

**credential_process::token_caching** - This is failing because the test case tries to upload a package twice to test out two different scenarios. However it does it with the same version number and therefore our new code changes mean that the second test publish job bails. I have verified a fix for this locally where we just need to change the version and expected stdout.

**registry_auth::token_not_logged** - Here, we are logging the token 8 times instead of the previous 7. I wonder if this is because of the extra check we make to confirm whether the crate@version exists already? If so, we can either change the test to expect 8 logs (if that is the new accepted behaviour), or check why there's an extra log. 

Let me know how I should proceed if we are okay with continuing with this solution.

CC: @torhovland

epage:
I think this approach, rather than `set_quiet(true)` makes more sense.

tests/testsuite/publish.rs
stupendoussuperpowers: This was initially added because publish during tests did not have any memory of previously uploaded packages by default. Not sure if that changed, but test-publish works fine without this responder now.

stupendoussuperpowers:
[stupendoussuperpowers:dupl](https://github.com/stupendoussuperpowers/cargo/tree/dupl) 

@torhovland - all tests are passing with this. you can apply this patch and we should be good to go.

src/cargo/ops/registry/publish.rs
torhovland: Updated by @stupendoussuperpowers.

tests/testsuite/publish.rs
torhovland: And it's updated.

torhovland:
Good job!

@epage This is ready for another look.

@stupendoussuperpowers Will you follow up any feedback, please?

stupendoussuperpowers:
I'm assuming I'll have to rebase to be 2 commits - first with the test and second with the updates. Will get to it.

torhovland:
For simplicity, feel free to reopen #13501 and finish the work there.

hi-rustin:
> @hi-rustin, you recently [refactored](https://github.com/rust-lang/cargo/commit/416a2b96fcb19689091b079680d41e4291843923#diff-277403efaf7f6382e405c12bf52e33337060618739c801f4e8690f3cb10d73f6) away the return tuple in `fn cargo::ops::registry::registry()`. Any thoughts about [reintroducing](https://github.com/rust-lang/cargo/pull/14338/files#diff-bc1a08bfe2fe76626f5fee55bfadc3df43043860d436c26a25a01a7fd613aff9) a return tuple here?

I think you are attempting to return a different type. What I removed from it is `RegistrySourceIds`, but what you actually want to return here is `RegistrySource`. So, I believe it is okay to add it.

tests/testsuite/publish.rs
```diff
@@ -130,6 +130,57 @@ You may press ctrl-c to skip waiting; the crate should be available shortly.
...
+                license = "MIT"
+                description = "foo"
+            "#,
+        )
+        .file("src/main.rs", "fn main() {}")
+        .build();
+
+    p.cargo("publish")
+        .replace_crates_io(registry_dupl.index_url())
+        .with_stderr(
```
epage: Can this commit be updated to be using the current testing style, rather than deferring that to later commits?

tests/testsuite/publish.rs
```diff
@@ -132,20 +132,7 @@ You may press ctrl-c to skip waiting; the crate should be available shortly.
...
-            let mut lock = arc.lock().unwrap();
-            *lock += 1;
-            if *lock <= 1 {
-                server.not_found(req)
-            } else {
-                server.index(req)
-            }
-        })
-        .build();
+    let registry_dupl = RegistryBuilder::new().http_api().http_index().build();
```
epage, line 135: Can we update the original test with these changes?

epage:
It'd be a big help for the commits to be cleaned up

Either

1. Add tests with them passing
2. Bail early
3. Fix double-message

or

1. Add tests with them passing
2. Change return type of `registry`
3. Bail early

stupendoussuperpowers:
#14448 was merged with these fixes. I believe we can close this.

PR #14339, opened 2024-08-01
fix: also build manpage for cargo.md

weihanglo:

Generate manpage for the main `cargo` command.

Run `mdbook serve src/doc` and check the modified page.

### Additional information

This was found during reviewing #14336


src/doc/man/generated_txt/cargo.txt
```diff
@@ -50,6 +50,9 @@ COMMANDS
            Execute unit and integration tests of a package.
 
    Manifest Commands
+       cargo-add(1)
```
weihanglo, line 53: We should automate this.

PR #14340, opened 2024-08-01
Infer registry

jneem:
While working on `cargo publish` for workspaces, I noticed a few ways in which the registry inference for `cargo package` seemed wrong. (This only affects the unstable `-Zpackage-workspace` feature.) This is all about the interaction between `--index`/`--registry`, and the `package.publish` fields.

- Previously, if multiple crates were packaged and they had different `package.publish` settings, we would ignore those setting and fall back to "crates-io". After this PR, we will bail out with an error instead (unless they've specified `--registry` or `--index`).
- Previously, we would validate the inferred registry against the `package.publish` settings, even if the `--index` argument would have taken precedence. After this PR, we will skip registry validation in this case.
- Previously, we were just ignoring the registry value inferred from `package.publish`: we'd validate that it was ok, but then just not use it.


tests/testsuite/package.rs
```diff
@@ -5943,3 +5943,191 @@ fn workspace_with_local_and_remote_deps() {
...
+                ]
+            "#,
+        )
+        .file("src/main.rs", "fn main() {}")
+        .build();
+
+    p.cargo("package -Zpackage-workspace")
+        .masquerade_as_nightly_cargo(&["package-workspace"])
+        .arg("--registry")
+        .arg("alternative")
```
epage, line 5971: For myself, I didn't notice the `arg` calls which made me suspicious.

At least put the `arg` calls immediately after `cargo`.  Preferably just put it in the `cargo` call.  Most end-to-end tests do that (unless the arg is dynamic) and so people are more likely to notice it that way.

src/cargo/ops/cargo_package.rs
```diff
@@ -185,44 +185,52 @@ fn infer_registry(
...
+                    None | Some([]) => RegistryOrIndex::Registry(CRATES_IO_REGISTRY.to_owned()),
+                    _ => bail!(
+                "cannot infer a registry\n\
+                Multiple `package.publish` settings found. Use --registry or --index to choose one."
+            ),
+                }
+            } else {
+                bail!(
+            "cannot infer a registry\n\
+            Not all packages have the same `package.publish` setting. Use --registry or --index to choose one."
```
epage, line 204: I feel like these messages aren't entirely clear from the end-users perspective, especially for running `cargo package`.  I'm just not sure how to entirely solve it if `cargo publish` will end up going through this code path.

At a high level, for the first message, I'd focus on it being ambiguous and I'd name the choices.  Maybe "`--registry is required to disambiguate between A, B, or C registries". Yes, this doesn't mention `--index` maybe that could be an "or" in there.

For the second, the issue is that it is either ambiguous (there is a common registry but other registries as well) or there is a conflict (for which `--registry` won't help).

jneem: >  I'm just not sure how to entirely solve it if cargo publish will end up going through this code path.

If we end up needing different error messages, my plan was to factor out a function that returns an enum of the possible inference errors. But it will indeed be easier if the errors can make sense for both `package` and `publish`.

I added a distinction between "non-overlapping `publish` fields" and "not-all-equal `publish` fields", but I haven't managed to come up with a great error message. I remember there was some discussion about diagnostics that point to locations in manifest files. Does that exist yet? Because I think that if we could point to the multiple distinct `publish` fields then it will be clear even if the message isn't great.

tests/testsuite/package.rs
```diff
@@ -6121,11 +6123,104 @@ The registry `alternative` is not listed in the `package.publish` value in Cargo
...
+[COMPILING] dep v0.1.0 (registry `alternative`)
+[COMPILING] main v0.0.1 ([ROOT]/foo/target/package/main-0.0.1)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 
 "#]])
     .run();
 }
+
+#[cargo_test]
+fn registry_not_inferred_because_of_multiple_options() {
```
epage, line 6147: Could you please add all tests in the same commit?

tests/testsuite/package.rs
```diff
@@ -6121,11 +6123,104 @@ The registry `alternative` is not listed in the `package.publish` value in Cargo
...
+        "#,
+        )
+        .file("dep/src/lib.rs", "")
+        .build();
+
+    p.cargo("package -Zpackage-workspace")
+        .masquerade_as_nightly_cargo(&["package-workspace"])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] --registry needed to disambiguate between alternative or alternative2 registries
```
epage, line 6201: - I think `is required` is clearer than `needed`
- Could be quote the registries?  In the middle of a sentence, they are hard to scan for.

tests/testsuite/package.rs
```diff
@@ -6100,8 +6103,7 @@ fn registry_not_inferred_because_of_missing_option() {
         .masquerade_as_nightly_cargo(&["package-workspace"])
         .with_status(101)
         .with_stderr_data(str![[r#"
-[ERROR] `dep` cannot be packaged.
-The registry `crates-io` is not listed in the `package.publish` value in Cargo.toml.
+[ERROR] `package.publish` settings have no registry in common
```
epage, line 6106: - We tend to refer to these as "fields" rather than "settings"
- This feels passive

What about

```
[ERROR] conflicts between `package,publish` fields between the selected packages
```

src/cargo/ops/cargo_package.rs
```diff
@@ -185,44 +185,77 @@ fn infer_registry(
...
+                            .unwrap_or_default()
+                            .iter()
+                            .collect::<HashSet<_>>()
+                    })
+                    .reduce(|xs, ys| xs.intersection(&ys).cloned().collect())
+                    .unwrap_or_default();
+                if common_regs.is_empty() {
+                    bail!("`package.publish` settings have no registry in common");
+                } else {
+                    bail!("--registry needed because not all `package.publish` settings agree",);
```
epage, line 229: We don't have a test for this case

epage, line 229: Are these errors taking into account that `None` / `vec![]` means "all" registries?

jneem: I thought `vec![]` meant "no registries"? But no, it isn't handing `None` correctly...

jneem:
I'm mystified by the failure in the build-std tests. When I run it locally I don't get the `[UPDATING]` line, but it seems to be reliably present in CI. (Also, I don't see why this PR would affect that test)

jneem:
Ok, I rebased and the mysterious build-std error seems to have gone away.

tests/testsuite/package.rs
```diff
@@ -6143,7 +6156,7 @@ fn registry_not_inferred_because_of_multiple_options() {
             "Cargo.toml",
             r#"
             [workspace]
-            members = ["dep", "main", "other"]
+            members = ["dep", "main"]
```
epage, line 6159: Should these test case changes be in a prior commit?

jneem: Sorry, I think I messed up the rebase. It should be fixed now

PR #14341, opened 2024-08-01
Change tests to support `rustc` wording changes

estebank:
With rust-lang/rust#126818 the output of rustc for resolution errors is going to change in such a way that some existing cargo tests will fail. Change them to support both the current and future output, so that those PRs can land in `rustc`. Follow up to #14135. Noticed in https://github.com/rust-lang-ci/rust/actions/runs/10192263962/job/28194880547#step:25:7529 when trying to land rust-lang/rust#126818.

### DIFF
--- a/tests/testsuite/check.rs
+++ b/tests/testsuite/check.rs
@@ -819,7 +819,7 @@ fn short_message_format() {
         .with_status(101)
         .with_stderr_data(str![[r#"
 [CHECKING] foo v0.0.1 ([ROOT]/foo)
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
Thanks and sorry that you always bump into this.😞.

estebank:
@weihanglo no problem! It's a shame that we have to deal with a complex multi-repo relationship, but at the same time I prefer the mild annoyance over uncaught unintended mistakes. :)

PR #14357, opened 2024-08-05
Don't specify the depedency name in the `cargo add` inferred name test

dpaoliello:
The `cargo_add/path_inferred_name` was incorrectly adding the name of the dependency to add in its command line.

PR #8834, opened 2020-11-05
Check if rust-src contains a vendor dir, and patch it in

Gankra:
This is the cargo side of https://github.com/rust-lang/wg-cargo-std-aware/issues/23

Note that this design naively assumes there is only one version of each package. It does not robustly verify this, and will presumably just cryptically fail to resolve dependencies.

See https://github.com/rust-lang/rust/pull/78790 for the other half of this change.

PR #14358, opened 2024-08-06
fix(build-std): remove hack on creating virtual std workspace 

weihanglo:

Starting from https://github.com/rust-lang/rust/pull/128534 (nightly-2024-08-05),
standard library has its own Cargo workspace.
Hence `-Zbuild-std` no longer need to fake a virtual workspace.

This also adjusts Cargo.toml in `mock-std` to align with std's Cargo.toml

I haven't done any e2e test for `-Zbuild-std` manually. Would appreciate if someone would like to.

### Additional information

There might be more clean-up we could do.

### DIFF
--- a/src/cargo/core/compiler/standard_lib.rs
+++ b/src/cargo/core/compiler/standard_lib.rs
@@ -6,13 +6,12 @@ use crate::core::compiler::{CompileKind, CompileMode, RustcTargetData, Unit};
 use crate::core::profiles::{Profiles, UnitFor};
 use crate::core::resolver::features::{CliFeatures, FeaturesFor, ResolvedFeatures};
 use crate::core::resolver::HasDevUnits;
-use crate::core::{Dependency, PackageId, PackageSet, Resolve, SourceId, Workspace};
+use crate::core::{PackageId, PackageSet, Resolve, Workspace};
 use crate::ops::{self, Packages};
 use crate::util::errors::CargoResult;
 use crate::GlobalContext;
 use std::collections::{HashMap, HashSet};
 use std::path::PathBuf;
-use std::rc::Rc;
 
 use super::BuildConfig;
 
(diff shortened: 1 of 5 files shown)


weihanglo:
See build failures:

* https://github.com/rust-lang/cargo/actions/runs/10246133822/job/28342625398?pr=14352
* https://github.com/rust-lang/cargo/actions/runs/10250133690/job/28355129599?pr=13709
* https://github.com/rust-lang/cargo/pull/14357#issuecomment-2270081936

glandium:
```
---- expected: tests/build-std/main.rs:116:27
++++ actual:   stderr
        1 + [UPDATING] crates.io index
```

That's an interesting error. -Zbuild-std should probably not touch crates.io

src/cargo/core/compiler/standard_lib.rs
```diff
@@ -74,60 +73,11 @@ pub fn resolve_std<'gctx>(
...
-    // to be the "current" member. `sysroot` is the root, and all other
-    // standard library crates are dependencies from there. Since none of the
-    // other crates need to alter their features, this should be fine, for
-    // now. Perhaps in the future features will be decoupled from the resolver
-    // and it will be easier to control feature selection.
-    let current_manifest = src_path.join("library/sysroot/Cargo.toml");
     // TODO: Consider doing something to enforce --locked? Or to prevent the
     // lock file from being written, such as setting ephemeral.
-    let mut std_ws = Workspace::new_virtual(src_path, current_manifest, virtual_manifest, gctx)?;
+    let mut std_ws = Workspace::new(&std_ws_manifest_path, gctx)?;
```
bjorn3, line 80: Can you build this workspace with `--locked` to avoid masking any future case where the lockfile is not up to date for whatever reason?

weihanglo: I understand the needs, though that requires more tweaks. I slightly lean toward get this fix first to unblock Cargo CI pipeline and then we make it `--locked`.

Regardless, which one is preferable?

* Build succeeded. Lockfile got update but was written to file system.
* Build failed because of any lockfile update.

Note that `tests/build-std/main.rs` should have guarded basic cases that won't update registry index, though they are not run in rust-lang/rust main CI pipeline.

bjorn3: I think the build should fail. An outdated lockfile means that something went wrong when building the rust-src component (it should be a verbatim copy of the lockfile used while building the standard library in the rust build system, which uses `--locked` on CI afaik) or that the rust-src component was (accidentally) tampered with on the user's system.

weihanglo: Tracking issue of it: https://github.com/rust-lang/wg-cargo-std-aware/issues/38

glandium:
FWIW, this didn't change anything wrt -Zbuild-std not working with the use of a vendoring tree. Parts of https://github.com/rust-lang/rust/pull/78790 and/or https://github.com/rust-lang/cargo/pull/8834 are probably still necessary.

weihanglo:
@glandium 
Could you open a new issue with reproduction if it is not supported?

glandium:
The issue already exists as https://github.com/rust-lang/wg-cargo-std-aware/issues/23

PR #14359, opened 2024-08-06
Fix: `cargo package` failed on bare commit git repo.

### LINKED ISSUE #14354: Failed to `cargo package --allow-dirty` for repo with no commit
### Problem

<https://github.com/rust-lang/cargo/pull/13960> unconditionally checks VCS status when doing `cargo package`.
However, a VCS (Git specifically) repository could have no commit history, and then it fails `cargo package` with such an error:

```
cargo +beta package --allow-dirty
warning: manifest has no description, license, license-file, documentation, homepage or repository.
error: revspec 'HEAD' not found; class=Reference (4); code=NotFound (-3)
```

### Steps

1. `cargo new foo`
2. cd foo
3. `cargo package --allow-dirty`

### Possible Solution(s)

The current `.cargo_vcs_info.json` format looks like:

```json
{
 "git": {
   "sha1": "aac20b6e7e543e6dd4118b246c77225e3a3a1302",
   "dirty": true
 },
 "path_in_vcs": ""
}
```

If the `HEAD` hash cannot be determined, should we instead emit a JSON like

```json
{
 "git": {
   "dirty": true
 },
 "path_in_vcs": ""
}
```

or just don't generate `.cargo_vcs_info.json` at all?

### Notes

_No response_

### Version

```text
cargo 1.81.0-beta.2 (a2b58c3da 2024-07-16)
release: 1.81.0-beta.2
commit-hash: a2b58c3dad4d554ba01ed6c45c41ff85390560f2
commit-date: 2024-07-16
```

linyihai:
Fixes #14354 

This approach chose to not generate a `.cargo_vcs_info.json` for bare commit git repo.

Compare the test changes before and after the two commits

### DIFF
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -87,8 +87,7 @@ struct VcsInfo {
 
 #[derive(Serialize)]
 struct GitVcsInfo {
-    #[serde(skip_serializing_if = "Option::is_none")]
-    sha1: Option<String>,
+    sha1: String,
     /// Indicate whether or not the Git worktree is dirty.
     #[serde(skip_serializing_if = "std::ops::Not::not")]
     dirty: bool,
@@ -744,10 +743,12 @@ fn check_repo_state(
                         .and_then(|p| p.to_str())
                         .unwrap_or("")
                         .replace("\\", "/");
-                    return Ok(Some(VcsInfo {
-                        git: git(p, src_files, &repo, &opts)?,
-                        path_in_vcs,
-                    }));
+                    let Some(git) = git(p, src_files, &repo, &opts)? else {
+                        // If the git repo lacks essensial field like `sha1`, and since this field exists from the beginning,
+                        // then don't generate the corresponding file in order to maintain consistency with past behavior.
+                        return Ok(None);
+                    };
+                    return Ok(Some(VcsInfo { git, path_in_vcs }));
                 }
             }
             gctx.shell().verbose(|shell| {
@@ -773,7 +774,7 @@ fn check_repo_state(
         src_files: &[PathBuf],
         repo: &git2::Repository,
         opts: &PackageOpts<'_>,
-    ) -> CargoResult<GitVcsInfo> {
+    ) -> CargoResult<Option<GitVcsInfo>> {
         // This is a collection of any dirty or untracked files. This covers:
         // - new/modified/deleted/renamed/type change (index or worktree)
         // - untracked files (which are "new" worktree files)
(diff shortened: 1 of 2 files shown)


linyihai:
It hits the API limit 

> #2 [internal] load metadata for docker.io/library/alpine:3.20
#2 ERROR: failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/library/alpine/manifests/sha256:eddacbc7e24bf8799a4ed3cdcfa50d4b88a323695ad80f317b6629883b2c2a78: 429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

weihanglo (approved):
Thank you!

weihanglo:
@rfcbot fcp merge

This insta-stabilizes an edge case of `.cargo_vcs_info.json`:

When a Git repo has no commits, the file content would be

```json5
{
  "git": {
    # there is no "sha1" field because `HEAD` is missing
    "dirty": true
  },
  "path_in_vcs": ""
}
```

epage:
@weihanglo there are several solutions for this
- Don't include the file
- Do whatever non-git locations do
- not have the git table

and likely more.

Is there a motivation for the proposed solution?  I didn't see it here or in the issue.

weihanglo:
I am sorry that I just did a really quick FCP without thinking hard on it.

I don't have a good answer. Thinking backward from #13695, the requirement is to have commit hash as well as the `path_in_vcs` field for other tools if they want to post-process further. The current approach this PR takes, generating vcs info file even with no `sha` field, is still useful for preserving `path_in_vcs`.

However, I personally doubt it is useful in real life, since that happens on in an edge case that no commit exists. I am fine with whatever approach we choose. We just need to fix the regression.

linyihai:
@weihanglo Thanks you rapidly response.

Sorry for that, I had change my mind. I'm not sure if the absence of this field will cause problems in other ways,  so it's better to not include/generate the `.cargo_vcs_info.json` in this case.

Of course, this is only a reasonable change that I currently feel, which should be very rare for this scenario, and it may be better to maintain the consistency with the previous

src/cargo/ops/cargo_package.rs
```diff
@@ -743,10 +744,12 @@ fn check_repo_state(
                         .and_then(|p| p.to_str())
                         .unwrap_or("")
                         .replace("\\", "/");
-                    return Ok(Some(VcsInfo {
-                        git: git(p, src_files, &repo, &opts)?,
-                        path_in_vcs,
-                    }));
+                    let git = git(p, src_files, &repo, &opts)?;
```
weihanglo, line 747: Could `fn git()` return a `None` when nothing needed to be recorded? So that we don't need `skip_serializing_if`  for `sha1` field.

weihanglo:
Agree that not generating vcs info file is also a reasonable approach to take. It stays with the old behavior (not generating when `--allow-dirty` presents) without blocking `cargo package`.

weihanglo:
Let's discuss different approaches in the issue then.

src/cargo/ops/cargo_package.rs
linyihai: Agree with you and updated it

weihanglo (approved):

weihanglo:
Thanks for the contribution. This should be the least controversial way to fix this. See the  discussion in #14354.

@bors r+

Issue #14355, opened 2024-08-05
Tracking Issue for path-bases


dpaoliello:
### Summary

RFC: [#3529](https://github.com/rust-lang/rfcs/pull/3529)
Original issue: #9855
Implementation: #12974
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#path-bases

Introduce shared base directories in Cargo configuration files that in turn enable base-relative path dependencies.

### Unresolved Issues

- [ ] What exact names we should use for the table (`path-bases`) and field names
(`base`)?
- [ ] What other built-in base paths could be useful?
  * `package` or `current-dir` for the directory of the current project?
  * `home` or `user_home` for the user's home directory?
  * `sysroot` for the current rustc sysroot?

### Future Extensions

- [ ] Add support for declaring path bases in the manifest
- [ ] Path bases relative to other path bases
- [ ] Path dependency with just a base
- [ ] Git dependencies

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

Sajjon:
@dpaoliello this makes it possible to shorten the chars needed when specifying paths in workspaces, which is a great feature.

But even better a feature would be to be able to **declare paths like constants, once** and reuse them everywhere. Has this been pitched / discussed? I understand this is a tracking issue for path-bases specifically, so sorry to hijack this issue, but what I propose is related to the problem which this tries to solve: *better ergonomics for `my-crate = { path = "../my-crate" }` config*.

Imagine many crates (30+) in a single workspace, where I would like to group some crates together on filesystem by use of folders. We note that this is ofc opaque to cargo, but filesystem grouping is still good for IDE exploration and git.

Project root:
```sh
crates/
    core/
        utils/
            Cargo.toml
            src/
                lib.rs
        error/
            Cargo.toml
            src/
                lib.rs
    clients/
        http/
            Cargo.toml
            src/
                lib.rs
        database
            Cargo.toml
            src/
                lib.rs
    really_huge_structure/
        logic/
            foo/
                Cargo.toml
                src/
                    lib.rs
            bar/
                Cargo.toml
                src/
                    lib.rs
        models/
            foo/
                Cargo.toml
                src/
                    lib.rs
            bar/
[...]

PR #14360, opened 2024-08-06
Implement base paths (RFC 3529) 1/n: path dep and patch support

dpaoliello:

This PR add support in path dependencies and `patch` table and the `workspace` built-in path base.

### DIFF
--- a/crates/cargo-util-schemas/src/manifest/mod.rs
+++ b/crates/cargo-util-schemas/src/manifest/mod.rs
@@ -776,6 +776,7 @@ pub struct TomlDetailedDependency<P: Clone = String> {
     // `path` is relative to the file it appears in. If that's a `Cargo.toml`, it'll be relative to
     // that TOML file, and if it's a `.cargo/config` file, it'll be relative to that file.
     pub path: Option<P>,
+    pub base: Option<PathBaseName>,
     pub git: Option<String>,
     pub branch: Option<String>,
     pub tag: Option<String>,
@@ -815,6 +816,7 @@ impl<P: Clone> Default for TomlDetailedDependency<P> {
             registry: Default::default(),
             registry_index: Default::default(),
             path: Default::default(),
+            base: Default::default(),
             git: Default::default(),
             branch: Default::default(),
             tag: Default::default(),
@@ -1409,6 +1411,16 @@ impl<T: AsRef<str>> FeatureName<T> {
     /// Validated feature name
     pub fn new(name: T) -> Result<Self, NameValidationError> {
         restricted_names::validate_feature_name(name.as_ref())?;
+        Ok(Self(name))
+    }
+}
+
+str_newtype!(PathBaseName);
+
+impl<T: AsRef<str>> PathBaseName<T> {
+    /// Validated path base name
+    pub fn new(name: T) -> Result<Self, NameValidationError> {
+        restricted_names::validate_path_base_name(name.as_ref())?;
         Ok(Self(name))
     }
 }
--- a/crates/cargo-util-schemas/src/restricted_names.rs
+++ b/crates/cargo-util-schemas/src/restricted_names.rs
@@ -238,6 +238,10 @@ pub(crate) fn validate_feature_name(name: &str) -> Result<()> {
     Ok(())
 }
 
+pub(crate) fn validate_path_base_name(name: &str) -> Result<()> {
+    validate_name(name, "path base name")
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/src/cargo/core/features.rs
+++ b/src/cargo/core/features.rs
@@ -513,6 +513,9 @@ features! {
 
     /// Allow multiple packages to participate in the same API namespace
     (unstable, open_namespaces, "", "reference/unstable.html#open-namespaces"),
+
+    /// Allow paths that resolve relatively to a base specified in the config.
+    (unstable, path_bases, "", "reference/unstable.html#path-bases"),
 }
 
 /// Status and metadata for a single unstable feature.
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -10,7 +10,9 @@ use crate::AlreadyPrintedError;
 use anyhow::{anyhow, bail, Context as _};
 use cargo_platform::Platform;
 use cargo_util::paths::{self, normalize_path};
-use cargo_util_schemas::manifest::{self, TomlManifest};
+use cargo_util_schemas::manifest::{
+    self, PackageName, PathBaseName, TomlDependency, TomlDetailedDependency, TomlManifest,
+};
 use cargo_util_schemas::manifest::{RustVersion, StringOrBool};
 use itertools::Itertools;
 use lazycell::LazyCell;
@@ -296,7 +298,7 @@ fn normalize_toml(
         features: None,
         target: None,
         replace: original_toml.replace.clone(),
-        patch: original_toml.patch.clone(),
+        patch: None,
         workspace: original_toml.workspace.clone(),
         badges: None,
         lints: None,
@@ -310,6 +312,7 @@ fn normalize_toml(
         inherit_cell
             .try_borrow_with(|| load_inheritable_fields(gctx, manifest_file, &workspace_config))
     };
+    let workspace_root = || inherit().map(|fields| fields.ws_root());
 
     if let Some(original_package) = original_toml.package() {
         let package_name = &original_package.name;
@@ -390,6 +393,7 @@ fn normalize_toml(
             &activated_opt_deps,
             None,
             &inherit,
+            &workspace_root,
             package_root,
             warnings,
         )?;
@@ -410,6 +414,7 @@ fn normalize_toml(
             &activated_opt_deps,
             Some(DepKind::Development),
             &inherit,
+            &workspace_root,
             package_root,
             warnings,
         )?;
@@ -430,6 +435,7 @@ fn normalize_toml(
             &activated_opt_deps,
             Some(DepKind::Build),
             &inherit,
+            &workspace_root,
             package_root,
             warnings,
         )?;
@@ -443,6 +449,7 @@ fn normalize_toml(
                 &activated_opt_deps,
                 None,
                 &inherit,
+                &workspace_root,
                 package_root,
                 warnings,
             )?;
@@ -463,6 +470,7 @@ fn normalize_toml(
                 &activated_opt_deps,
                 Some(DepKind::Development),
                 &inherit,
+                &workspace_root,
                 package_root,
                 warnings,
             )?;
@@ -483,6 +491,7 @@ fn normalize_toml(
                 &activated_opt_deps,
                 Some(DepKind::Build),
                 &inherit,
+                &workspace_root,
                 package_root,
                 warnings,
             )?;
@@ -498,6 +507,13 @@ fn normalize_toml(
             );
         }
         normalized_toml.target = (!normalized_target.is_empty()).then_some(normalized_target);
+
+        normalized_toml.patch = normalize_patch(
+            gctx,
+            original_toml.patch.as_ref(),
+            &workspace_root,
+            features,
+        )?;
 
         let normalized_lints = original_toml
             .lints
@@ -517,6 +533,37 @@ fn normalize_toml(
     }
 
     Ok(normalized_toml)
+}
+
+fn normalize_patch<'a>(
+    gctx: &GlobalContext,
+    original_patch: Option<&BTreeMap<String, BTreeMap<PackageName, TomlDependency>>>,
+    workspace_root: &dyn Fn() -> CargoResult<&'a PathBuf>,
+    features: &Features,
+) -> CargoResult<Option<BTreeMap<String, BTreeMap<PackageName, TomlDependency>>>> {
+    if let Some(patch) = original_patch {
+        let mut normalized_patch = BTreeMap::new();
+        for (name, packages) in patch {
+            let mut normalized_packages = BTreeMap::new();
+            for (pkg, dep) in packages {
+                let dep = if let TomlDependency::Detailed(dep) = dep {
+                    let mut dep = dep.clone();
+                    normalize_path_dependency(gctx, &mut dep, workspace_root, features)
+                        .with_context(|| {
+                            format!("resolving path for patch of ({pkg}) for source ({name})")
+                        })?;
+                    TomlDependency::Detailed(dep)
+                } else {
+                    dep.clone()
+                };
+                normalized_packages.insert(pkg.clone(), dep);
+            }
+            normalized_patch.insert(name.clone(), normalized_packages);
+        }
+        Ok(Some(normalized_patch))
+    } else {
+        Ok(None)
+    }
 }
 
 #[tracing::instrument(skip_all)]
@@ -710,6 +757,7 @@ fn normalize_dependencies<'a>(
     activated_opt_deps: &HashSet<&str>,
     kind: Option<DepKind>,
     inherit: &dyn Fn() -> CargoResult<&'a InheritableFields>,
+    workspace_root: &dyn Fn() -> CargoResult<&'a PathBuf>,
     package_root: &Path,
     warnings: &mut Vec<String>,
 ) -> CargoResult<Option<BTreeMap<manifest::PackageName, manifest::InheritableDependency>>> {
@@ -768,6 +816,8 @@ fn normalize_dependencies<'a>(
                     }
                 }
             }
+            normalize_path_dependency(gctx, d, workspace_root, features)
+                .with_context(|| format!("resolving path dependency {name_in_toml}"))?;
         }
 
         // if the dependency is not optional, it is always used
@@ -784,6 +834,23 @@ fn normalize_dependencies<'a>(
         }
     }
     Ok(Some(deps))
+}
+
+fn normalize_path_dependency<'a>(
+    gctx: &GlobalContext,
+    detailed_dep: &mut TomlDetailedDependency,
+    workspace_root: &dyn Fn() -> CargoResult<&'a PathBuf>,
+    features: &Features,
+) -> CargoResult<()> {
+    if let Some(base) = detailed_dep.base.take() {
+        if let Some(path) = detailed_dep.path.as_mut() {
+            let new_path = lookup_path_base(&base, gctx, workspace_root, features)?.join(&path);
+            *path = new_path.to_str().unwrap().to_string();
+        } else {
+            bail!("`base` can only be used with path dependencies");
+        }
+    }
+    Ok(())
 }
 
 fn load_inheritable_fields(
@@ -901,13 +968,17 @@ impl InheritableFields {
         };
         let mut dep = dep.clone();
         if let manifest::TomlDependency::Detailed(detailed) = &mut dep {
-            if let Some(rel_path) = &detailed.path {
-                detailed.path = Some(resolve_relative_path(
-                    name,
-                    self.ws_root(),
-                    package_root,
-                    rel_path,
-                )?);
+            if detailed.base.is_none() {
+                // If this is a path dependency without a base, then update the path to be relative
+                // to the workspace root instead.
+                if let Some(rel_path) = &detailed.path {
+                    detailed.path = Some(resolve_relative_path(
+                        name,
+                        self.ws_root(),
+                        package_root,
+                        rel_path,
+                    )?);
+                }
             }
         }
         Ok(dep)
@@ -2151,6 +2222,33 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
     }
 }
 
+pub(crate) fn lookup_path_base<'a>(
+    base: &PathBaseName,
+    gctx: &GlobalContext,
+    workspace_root: &dyn Fn() -> CargoResult<&'a PathBuf>,
+    features: &Features,
+) -> CargoResult<PathBuf> {
+    features.require(Feature::path_bases())?;
+
+    // HACK: The `base` string is user controlled, but building the path is safe from injection
+    // attacks since the `PathBaseName` type restricts the characters that can be used to exclude `.`
+    let base_key = format!("path-bases.{base}");
+
+    // Look up the relevant base in the Config and use that as the root.
+    if let Some(path_bases) = gctx.get::<Option<ConfigRelativePath>>(&base_key)? {
+        Ok(path_bases.resolve_path(gctx))
+    } else {
+        // Otherwise, check the built-in bases.
+        match base.as_str() {
+            "workspace" => Ok(workspace_root()?.clone()),
+            _ => bail!(
+                "path base `{base}` is undefined. \
+            You must add an entry for `{base}` in the Cargo configuration [path-bases] table."
+            ),
+        }
+    }
+}
+
 pub trait ResolveToPath {
     fn resolve(&self, gctx: &GlobalContext) -> PathBuf;
 }
@@ -2865,6 +2963,7 @@ fn prepare_toml_for_publish(
                 let mut d = d.clone();
                 // Path dependencies become crates.io deps.
                 d.path.take();
+                d.base.take();
                 // Same with git dependencies.
                 d.git.take();
                 d.branch.take();
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -101,6 +101,7 @@ Each new feature described below should explain how to use it.
     * [Edition 2024](#edition-2024) — Adds support for the 2024 Edition.
     * [Profile `trim-paths` option](#profile-trim-paths-option) --- Control the sanitization of file paths in build outputs.
     * [`[lints.cargo]`](#lintscargo) --- Allows configuring lints for Cargo.
+    * [path bases](#path-bases) --- Named base directories for path dependencies.
 * Information and metadata
     * [Build-plan](#build-plan) --- Emits JSON information on which commands will be run.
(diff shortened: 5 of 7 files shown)


src/cargo/util/toml/mod.rs
```diff
@@ -2135,7 +2139,16 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
...
+                let path = if let Some(base) = orig.base.as_ref() {
+                    lookup_path_base(&base, manifest_ctx.gctx, manifest_ctx.root)
+                        .with_context(|| {
+                            format!("resolving path base for dependency ({name_in_toml})")
+                        })?
+                        .join(path)
+                } else {
+                    // This is a standard path with no prefix.
+                    manifest_ctx.root.join(path)
+                };
```
epage, line 2151: I was hoping this could be taken care of during `normalize_dependencies`, dropping `base`.  I've been moving this code more generally to the model of data transformation happens during `normalize` and `to_real_manifest` is about adapting data types.

We could then do a follow up refactor to unify that with the inheriting dependencies without a base, as the base then is just "workspace"

src/cargo/util/toml/mod.rs
epage, line 1: We should have an error check that `base` is only present if `path` is

epage, line 1: We need to strip `dep.base` when publishing inside of `map_dependency`

src/cargo/util/toml/mod.rs
```diff
@@ -2151,6 +2164,40 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
...
 }
 
+pub(crate) fn lookup_path_base(
+    base: &str,
+    gctx: &GlobalContext,
+    manifest_root: &Path,
+) -> CargoResult<PathBuf> {
+    if !gctx.cli_unstable().path_bases {
+        bail!("usage of path bases requires `-Z path-bases`");
+    }
```
epage, line 2174: Can we use this instead
```rust
    gctx.cli_unstable().fail_if_stable_command(
        gctx,
        "clean gc",
        12633,
        "gc",
        gctx.cli_unstable().gc,
    )?;
```

src/cargo/util/toml/mod.rs
```diff
@@ -2151,6 +2164,40 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
...
+        // Otherwise, check the built-in bases.
+        match base {
+            "workspace" => {
+                if let Some(workspace_root) = find_workspace_root(manifest_root, gctx)? {
+                    Ok(workspace_root.parent().unwrap().to_path_buf())
+                } else {
+                    bail!(
+                        "the `workspace` built-in path base cannot be used outside of a workspace."
+                    )
+                }
```
epage, line 2191: finding a workspace root is expensive.  We should reuse the workspace root found for inheritance.

crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -776,6 +776,7 @@ pub struct TomlDetailedDependency<P: Clone = String> {
     // `path` is relative to the file it appears in. If that's a `Cargo.toml`, it'll be relative to
     // that TOML file, and if it's a `.cargo/config` file, it'll be relative to that file.
     pub path: Option<P>,
+    pub base: Option<String>,
```
epage, line 779: We should use a newtype, like `PackageName`, to enforce the naming rules
```rust
str_newtype!(PackageName);

impl<T: AsRef<str>> PackageName<T> {
    /// Validated package name
    pub fn new(name: T) -> Result<Self, NameValidationError> {
        restricted_names::validate_package_name(name.as_ref())?;
        Ok(Self(name))
    }
}
```

src/cargo/util/toml/mod.rs
```diff
@@ -2151,6 +2164,40 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
...
+    gctx: &GlobalContext,
+    manifest_root: &Path,
+) -> CargoResult<PathBuf> {
+    if !gctx.cli_unstable().path_bases {
+        bail!("usage of path bases requires `-Z path-bases`");
+    }
+
+    // Look up the relevant base in the Config and use that as the root.
+    if let Some(path_bases) =
+        gctx.get::<Option<ConfigRelativePath>>(&format!("path-bases.{base}"))?
```
epage, line 2178: We should probably have a comment explaining why this path-injection is safe from abuse

src/cargo/util/toml/mod.rs
epage: Sorry, I missed that you do this

src/cargo/core/features.rs
```diff
@@ -776,6 +776,7 @@ unstable_cli_options!(
     no_index_update: bool = ("Do not update the registry index even if the cache is outdated"),
     package_workspace: bool = ("Handle intra-workspace dependencies when packaging"),
     panic_abort_tests: bool = ("Enable support to run tests with -Cpanic=abort"),
+    path_bases: bool = ("Allow paths that resolve relatively to a base specified in the config"),
```
epage, line 779: Is there a reason you did this as a `-Z` instead of as `cargo-features`?

`-Z` with `Cargo.toml` unstable features is usually reserved for "we gracefully degrade when we ignore these parts of the manifest".

tests/testsuite/patch.rs
```diff
@@ -3028,3 +3028,45 @@ foo v0.0.0 ([ROOT]/foo)
...
+        .file("Cargo.toml", &basic_manifest("bar", "0.5.0"))
+        .file("src/lib.rs", "pub fn hello() {}")
+        .build();
+    Package::new("bar", "0.5.0").publish();
+
+    let p = project()
+        .file(
+            ".cargo/config.toml",
+            &format!(
+                "[path-bases]\ntest = '{}'",
```
epage, line 3045: Please prefer actual newlines to escapes ones.  It makes the files easier to read

tests/testsuite/patch.rs
```diff
@@ -3028,3 +3028,45 @@ foo v0.0.0 ([ROOT]/foo)
...
+                version = "0.5.0"
+                authors = ["wycats@example.com"]
+                edition = "2018"
+
+                [dependencies]
+                bar = "0.5.0"
+
+                [patch.crates-io.bar]
+                path = 'bar'
+                base = 'test'
```
epage, line 3063: nt: for myself at least, most dependency tables are inline (`{}`) and I keep having a hard time reading these tests that use standard TOML tables (`[]`)

tests/testsuite/path.rs
```diff
@@ -589,6 +589,354 @@ Caused by:
...
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("build -v -Zpath-bases")
+        .masquerade_as_nightly_cargo(&["path-bases"])
+        .run();
+}
+
+#[cargo_test]
+fn workspace_with_base() {
```
epage, line 680: I misread this at first to be about using `base = "workspace"`.  Might be clearer to comment about "inheriting` than `workspace`.

tests/testsuite/path.rs
```diff
@@ -589,6 +589,354 @@ Caused by:
...
+        )
+        .file("src/main.rs", "fn main() {}")
+        .file(
+            "child/Cargo.toml",
+            r#"
+                [package]
+                name = "child"
+                version = "0.1.0"
+                authors = []
+                workspace = ".."
```
epage, line 719: Is there a reason this is explicitly set?

tests/testsuite/patch.rs
epage, line 3063: nit: I would also find it easier to read for `base` to come before `path` as we are doing `base.join(path)`

tests/testsuite/path.rs
```diff
@@ -589,6 +589,354 @@ Caused by:
...
+                name = "parent"
+                version = "0.1.0"
+                authors = []
+
+                [workspace]
+                members = ["child"]
+
+                [workspace.dependencies.dep_with_base]
+                path = '../dep_with_base'
+                base = 'workspace'
```
epage, line 841: Let's simplify this test and only do one thing in it (shadow base) rather than also covering inheritance.

It would also help if this had a clear parallel with `workspace_builtin_base`

tests/testsuite/path.rs
```diff
@@ -589,6 +589,354 @@ Caused by:
...
+            "\
+[ERROR] failed to parse manifest at `[..]/foo/Cargo.toml`
+
+Caused by:
+  resolving path base for dependency (bar)
+
+Caused by:
+  the `workspace` built-in path base cannot be used outside of a workspace.
+",
+        )
```
epage, line 936: Cargo does see this as a workspace.  I'm unsure if its worth preventing this or not

tests/testsuite/path.rs
epage, line 1: What odd cases might we need to test?
- `path = """`
- `path = "."`

Anything else?

src/cargo/util/toml/mod.rs
epage, line 2178: Aside: I wonder how this is on performance and if we should load all of `path-bases` and then index into it (that also removes tracking of invariants to prevent path-injection)

src/cargo/util/toml/mod.rs
dpaoliello: Ok, moved it into `normalize`

src/cargo/util/toml/mod.rs
dpaoliello: Added

src/cargo/util/toml/mod.rs
dpaoliello: Done

dpaoliello: I added a new `fail_if_stable_feature` function that prints ``the `{feature}` feature is unstable`` otherwise it was a bit weird to see ``the `cargo path bases` command`` in the error message.

src/cargo/util/toml/mod.rs
dpaoliello: I wasn't 100% sure of the best way to do this - so please let me know if what I've done is ok.

src/cargo/core/features.rs
dpaoliello: Ah, didn't realize that was a thing (sorry, I inherited this from Jonhoo a while ago...)

src/cargo/util/toml/mod.rs
dpaoliello: Switched to using a feature instead of a `-Z`

crates/cargo-util-schemas/src/manifest/mod.rs
dpaoliello: Done, and added validation per the RFC

src/cargo/util/toml/mod.rs
dpaoliello: Done

tests/testsuite/patch.rs
dpaoliello: Changed, here and elsewhere

tests/testsuite/patch.rs
dpaoliello: Switched to inline, here and elsewhere

tests/testsuite/path.rs
dpaoliello: Nope, removed.

tests/testsuite/patch.rs
dpaoliello: Switched.

tests/testsuite/path.rs
dpaoliello: Removed the inheritance.

Not sure what you mean by "clear parallel" - it now looks the same as `workspace_builtin_base` except for the config.toml and a different `path`.

tests/testsuite/path.rs
dpaoliello: Maybe I'm misunderstanding what it means to be "in a workspace" - it would certainly simplify the logic of finding a workspace root.

When I look at the docs, it mentions that a workspace is created using the `[workspace]` table, so I find it surprising that this is considered a workspace.

src/cargo/util/toml/mod.rs
dpaoliello: I guess it depends on whether we think there will be more items in the config.toml (i.e., making the index is expensive) or more uses in the Cargo.toml (i.e., having to look up the same items many times). For now, I'd rather wait and see how the feature is used, but I definitely add this as a note in the tracking issue.

tests/testsuite/path.rs
dpaoliello: Does it really make sense to have dedicated tests for these? I'm using `PathBuf::join`, so I'd expect any unusual value for the path base or the path to be handled by the Standard Library/OS.

epage:
> Done

Wanting to double check that you intended to not push the mentioned changes.

src/cargo/core/features.rs
epage: In case you didn't find it, the docs for this is at https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/features/index.html

tests/testsuite/path.rs
epage: Every package has a workspace, some are just implicitly created.

dpaoliello:
> Wanting to double check that you intended to not push the mentioned changes.

Yes, sorry, not sure what the correct etiquette is with this. Usually I reply to comments and resolve conversations -> push updated code -> require reviews.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 [a2b58c3d...HEAD](https://github.com/rust-lang/cargo/compare/a2b58c3d...HEAD)
 
 ### Added
+- Added the `path-bases` feature to support paths that resolve relatively to a
+  base specified in the config.
+  [docs](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#path-bases)
+  [#14360](https://github.com/rust-lang/cargo/pull/14360)
```
epage, line 10: Normally changelogs are handled semi-automatically.  Unsure if a manually written entry will interfere.  @weihanglo ?

If we do keep this, it needs to be in a "Nightly only" section

crates/cargo-util-schemas/src/restricted_names.rs
```diff
@@ -238,6 +238,37 @@ pub(crate) fn validate_feature_name(name: &str) -> Result<()> {
...
 
+pub(crate) fn validate_path_base_name(name: &str) -> Result<()> {
+    let what = "path base name";
+    let mut chars = name.chars();
+
+    let Some(first_char) = chars.next() else {
+        return Err(ErrorKind::Empty(what).into());
+    };
+
+    if !first_char.is_alphabetic() {
```
epage, line 249: Is there a reason we went with ASCII only names in the RFC?  For myself, I'd prefer we try to aim for consistency between our different validation rules which would likely mean delegating to `validate_name`

tests/testsuite/patch.rs
```diff
@@ -3028,3 +3028,49 @@ foo v0.0.0 ([ROOT]/foo)
 
     assert_eq!(p.read_file("Cargo.lock"), p.read_file("Cargo.lock.orig"));
 }
+
+#[cargo_test]
+fn patch_with_base() {
```
epage, line 3033: Do we also need a test for `[patch]` in `.cargo/config.toml`?

tests/testsuite/path.rs
```diff
@@ -589,6 +589,444 @@ Caused by:
...
+
+Caused by:
+  path base `test` is undefined. You must add an entry for `test` in the Cargo configuration [path-bases] table.
+",
+        )
+        .run();
+}
+
+#[cargo_test]
+fn base_pathout_path() {
```
epage, line 919:
```suggestion
fn base_without_path() {
```

src/cargo/util/toml/mod.rs
```diff
@@ -768,6 +822,8 @@ fn normalize_dependencies<'a>(
                     }
                 }
             }
+            normalize_path_dependency(gctx, d, workspace_root, features)
+                .with_context(|| format!("resolving path dependency ({name_in_toml})"))?;
```
epage, line 826:
```suggestion
                .with_context(|| format!("resolving path dependency `{name_in_toml}`"))?;
```
seems weird to use a parenthetical when its not needed

src/cargo/util/toml/mod.rs
epage: ok

crates/cargo-util-schemas/src/manifest/mod.rs
epage: ok

src/cargo/util/toml/mod.rs
epage: ok

src/cargo/util/toml/mod.rs
```diff
@@ -2151,6 +2225,41 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
...
+    workspace_root: &dyn Fn() -> CargoResult<Option<&'a PathBuf>>,
+    features: &Features,
+) -> CargoResult<PathBuf> {
+    features.require(Feature::path_bases())?;
+
+    // Look up the relevant base in the Config and use that as the root.
+    // NOTE: The `base` string is user controlled, but building the path is safe from injection
+    // attacks since the `PathBaseName` type restricts the characters that can be used.
+    if let Some(path_bases) =
+        gctx.get::<Option<ConfigRelativePath>>(&format!("path-bases.{base}"))?
```
epage, line 2240: imo the structure makes this harder to follow
```suggestion
    // HACK: The `base` string is user controlled, but building the path is safe from injection
    // attacks since the `PathBaseName` type restricts the characters that can be used to exclude `.`
    let base_key = format!("path-bases.{base}");
    
    // Look up the relevant base in the Config and use that as the root.
    if let Some(path_bases) =
        gctx.get::<Option<ConfigRelativePath>>(&base_key)?
```

src/cargo/core/features.rs
epage: ok

tests/testsuite/patch.rs
epage: ok

tests/testsuite/patch.rs
epage: ok

tests/testsuite/path.rs
epage: ok

tests/testsuite/path.rs
epage: ok

tests/testsuite/patch.rs
epage: patch and the docs are still the other way.

Seeing the change, I'm mixed about what is "right" but being consistent would be good.

tests/testsuite/path.rs
epage: ok

tests/testsuite/path.rs
epage: That depends on how the paths are used after this  I've had tests that don't do well in odd cases like this.  It depends on if you are dependent on the paths being "clean (prefix checks) and making sure the cleaning comes after.

src/cargo/util/toml/mod.rs
epage: If its in the tracking issue, we can cover it later

epage:
> Yes, sorry, not sure what the correct etiquette is with this. Usually I reply to comments and resolve conversations -> push updated code -> require reviews.

Not many people remember to "request review" and you never know whether someone will or not (in general, I just assume people won't do it).

In general, its strange to look at a PR with things marked "done" without code pushes, particularly if I'm coming back to the PR several days after the message.

tests/testsuite/path.rs
dpaoliello: Fair enough - I've removed this error

CHANGELOG.md
dpaoliello: Moved into the "Nightly only" section, but I'm happy to remove this if you'd prefer.

crates/cargo-util-schemas/src/restricted_names.rs
dpaoliello: I was following the wording in the RFC. I switched to `validate_name` to be consistent with the rest of Cargo.

tests/testsuite/patch.rs
dpaoliello: Good catch - I'd tried adding this and implementing it, but it raised a few big issues:
1. Configuration doesn't have a `cargo-features` key (as I understand it), so how do we handle unstable features? (Just allow using base paths? Require that every manifest using this configuration allow base paths?)
2. How to handle the `workspace` built-in? (block it?)
3. `patch` and `replace` have a unified code path for resolving dependencies, so do I need to differentiate to only permit `base` in `patch` tables?
4. Do I need to add back the code to handle base paths in `to_dependency_source_id`?

dpaoliello: I might need to move support for `[patch]` in config files into a new PR. Or maybe drop support for `[patch]` given the complications it causes...

src/cargo/util/toml/mod.rs
epage: ok

> We could then do a follow up refactor to unify that with the inheriting dependencies without a base, as the base then is just "workspace"

I've noted this as unresolved issue on #14355

tests/testsuite/path.rs
epage: We should still have a test for this, either way.

I've noted the direction *I think* you went in #14355 for evaluation during stabilization

src/cargo/util/toml/mod.rs
epage: ok

tests/testsuite/patch.rs
epage: > I might need to move support for [patch] in config files into a new PR. Or maybe drop support for [patch] given the complications it causes...

I think its doable but I also agree we should probably split it out at this point so it doesn't block the rest of your work

> Configuration doesn't have a cargo-features key (as I understand it), so how do we handle unstable features? (Just allow using base paths? Require that every manifest using this configuration allow base paths?)

This would use config's unstable mechanism,

> How to handle the workspace built-in? (block it?)

config is relative to itself and is independent of workspaces, so it should fail

> patch and replace have a unified code path for resolving dependencies, so do I need to differentiate to only permit base in patch tables?

Unsure.

> Do I need to add back the code to handle base paths in to_dependency_source_id?

We need to handle patch somewhere.  Where I don't care but preferably we don't change the normalization from how it is now in this PR.

tests/testsuite/path.rs
epage: ok

src/cargo/util/toml/mod.rs
epage: ok

src/cargo/util/toml/mod.rs
epage: ok

tests/testsuite/patch.rs
dpaoliello: I've added "`[patch]` table in config support" as one of the items to implement in the tracking issue

tests/testsuite/path.rs
dpaoliello: Even though I removed the error, it looks like the code that finds a workspace root already raises an error for this case, so I'm going to stick with using that error. I've removed the note from the tracking issue.

tests/testsuite/path.rs
dpaoliello: Ok, added a few different tests: `.`, `..`, `"`
Also added a test for self-dependency via a path base.

tests/testsuite/patch.rs
epage: ok

tests/testsuite/patch.rs
epage: ok

tests/testsuite/path.rs
epage: ok

I've added it as an unresolved question

tests/testsuite/path.rs
epage: I see `self_dependency_using_base` which is a good case but not the others

CHANGELOG.md
weihanglo: The changelog process skips pull requests that are already listed in the file, so should be fine.

We usually combine relevant PRs into one entry, but that should be fine here also.

A risk here is we are not 100% sure which version this PR is going to get into, though that's also fine as we can delete the entry then.

src/doc/src/reference/unstable.md
```diff
@@ -1570,6 +1571,58 @@ implicit-features = "warn"
 workspace = true
 ```
 
+## Path Bases
+
+* Tracking Issue: [#14355](https://github.com/rust-lang/cargo/issues/14355)
```
epage, line 1576: This never makes mention of the unstable feature to activate this.

tests/testsuite/path.rs
dpaoliello: `invalid_path_with_base`, `parent_dir_with_base` and `current_dir_with_base`

CHANGELOG.md
dpaoliello: I've removed the note: I don't want to block this PR on getting the release note exactly right. I'm happy to discuss the text here or on zulip.

src/doc/src/reference/unstable.md
dpaoliello: Added `cargo-features` to the example Cargo.toml

tests/testsuite/path.rs
epage: I'm sorry, that was a typo.  I didn't mean `path = """` but `path = ""`.  I was really confused when you mentioned an "invalid" test.

src/doc/src/reference/unstable.md
epage: Thanks!

PR #14361, opened 2024-08-06
Implement base paths (RFC 3529) 2/n: unstable docs

dpaoliello:

This PR documents the feature in the unstable book.


src/doc/src/reference/unstable.md
```diff
@@ -1570,6 +1571,58 @@ implicit-features = "warn"
 workspace = true
 ```
 
+## Path Bases
```
epage, line 1574: imo we shouldn't merge functionality without the docs (except in some very specific circumstances).  This fully belongs with the prior PR.  I think what got confusing was my comment about the unstable docs.   I was more suggesting something like b8a684674a6be0c4235834530527d5b085503357 or 4d09a6a7453c9bcc9d3226a26de7a4cf52433942 (though in its own PR rather than a commit in a PR.  Then as each PR adds functionality, the docs get updated to reflect that.

PR #14364, opened 2024-08-07
`cargo package --workspace` must support `publish = false`

### LINKED ISSUE #14356: `-Zpackage-workspace` is not smart about `publish = false`
### Problem

I was hoping to include `-Zpackage-workspace` as part of a workspace's CI so that it doesn't repeat certain embarrassing mistakes, but it didn't work as expected with respect to certain crates that have `publish = false` on them.

### Steps

After bumping the versions of all packages I wished to publish a new release for, I used this on the workspace with the following command:
```console
$ cargo +nightly package --workspace -Zpackage-workspace
     Locking 7 packages to latest compatible versions
    Updating cargo-pgrx v0.12.0-beta.3 (/home/jubilee/tcdi/pgrx/cargo-pgrx) -> v0.12.0-beta.4
    Updating pgrx v0.12.0-beta.3 (/home/jubilee/tcdi/pgrx/pgrx) -> v0.12.0-beta.4
    Updating pgrx-macros v0.12.0-beta.3 (/home/jubilee/tcdi/pgrx/pgrx-macros) -> v0.12.0-beta.4
    Updating pgrx-pg-config v0.12.0-beta.3 (/home/jubilee/tcdi/pgrx/pgrx-pg-config) -> v0.12.0-beta.4
    Updating pgrx-pg-sys v0.12.0-beta.3 (/home/jubilee/tcdi/pgrx/pgrx-pg-sys) -> v0.12.0-beta.4
    Updating pgrx-sql-entity-graph v0.12.0-beta.3 (/home/jubilee/tcdi/pgrx/pgrx-sql-entity-graph) -> v0.12.0-beta.4
    Updating pgrx-tests v0.12.0-beta.3 (/home/jubilee/tcdi/pgrx/pgrx-tests) -> v0.12.0-beta.4
error: `aggregate` cannot be packaged.
The registry `crates-io` is not listed in the `package.publish` value in Cargo.toml.
```

### Possible Solution(s)

This seems a curious error, given that this is intentional? I wish to not publish that crate and thus do not wish to package it for publishing?

Cargo has the option of doing the "smart" thing ("What I mean") here and simply ignoring such packages. The other obvious alternative is that it can simply continue to fail-fast, but if so it should probably have a more helpful error message. Note these are "leaf" packages on which nothing depends, so automatically excluding them _would_ work.

### Notes

I do have the option of simply removing it from the workspace, but it all gets a bit awkward. To complicate matters further, these packages being in the workspace does affect feature resolution which also affects whether or not basic `cargo check` or `cargo test` work (as the library is one of those "you must enable one and only one of these features" sorts).

Related:
- https://github.com/rust-lang/cargo/issues/10948

### Version

```text
cargo 1.82.0-nightly (b5d44db1d 2024-07-26)
release: 1.82.0-nightly
commit-hash: b5d44db1daf0469b227a6211b987162a39a54730
commit-date: 2024-07-26
libcurl: 8.8.0-DEV (sys:0.4.73+curl-8.8.0 vendored ssl:OpenSSL/3.3.1)
ssl: OpenSSL 3.3.1 4 Jun 2024
```

torhovland:
Fixes #14356.

OK, so the issue _can_ be fixed quite easily. But I'm not sure if it is the correct fix. So in this PR we are simply skipping validating the published-to registry if `publish = false` when inferring what registry to use. But that means we allow any workspace where one of the packages, possibly a crucial dependency, is configured to not get published. So the final, published packages may not work for anybody.

Do we need a smarter validation, that only allows this for packages that are not dependencies of other packages that are scheduled to get published?

### DIFF
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -207,7 +207,7 @@ fn infer_registry(
         let reg_name = publish_registry.as_deref().unwrap_or(CRATES_IO_REGISTRY);
         for pkg in pkgs {
             if let Some(allowed) = pkg.publish().as_ref() {
-                if !allowed.iter().any(|a| a == reg_name) {
+                if !allowed.is_empty() && !allowed.iter().any(|a| a == reg_name) {
                     bail!(
                         "`{}` cannot be packaged.\n\
                          The registry `{}` is not listed in the `package.publish` value in Cargo.toml.",
--- a/tests/testsuite/package.rs
+++ b/tests/testsuite/package.rs
@@ -2868,21 +2868,30 @@ fn in_workspace_with_publish_false() {
         .build();
 
     p.cargo("package --workspace")
-        // FIXME: This reproduces https://github.com/rust-lang/cargo/issues/14356
-        .with_status(101)
         .with_stderr_data(str![[r#"
-[ERROR] `no-publish` cannot be packaged.
-The registry `crates-io` is not listed in the `package.publish` value in Cargo.toml.
+[WARNING] manifest has no documentation, homepage or repository.
+See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for more info.
+[PACKAGING] foo v0.0.1 ([ROOT]/foo)
+[PACKAGED] 4 files, [FILE_SIZE]B ([FILE_SIZE]B compressed)
+[WARNING] manifest has no documentation, homepage or repository.
+See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for more info.
+[PACKAGING] no-publish v0.0.1 ([ROOT]/foo/no-publish)
+[PACKAGED] 4 files, [FILE_SIZE]B ([FILE_SIZE]B compressed)
+[VERIFYING] foo v0.0.1 ([ROOT]/foo)
+[COMPILING] foo v0.0.1 ([ROOT]/foo/target/package/foo-0.0.1)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+[VERIFYING] no-publish v0.0.1 ([ROOT]/foo/no-publish)
+[COMPILING] no-publish v0.0.1 ([ROOT]/foo/target/package/no-publish-0.0.1)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 
 "#]])
         .run();
 
-    // FIXME: Enable these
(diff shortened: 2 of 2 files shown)


jneem:
`infer_registry` has a bunch of problems that are already addressed in #14340, so I think any further fixes should build on that.

Other than that, I think if you also skip putting the `publish = false` packages in the local overlay then this fix is reasonable. Because that way, verification will fail if and only it would have failed without publishing those packages. (e.g. maybe it's ok not to publish them because the dependency on them can be satisfied by some other already-published version).

torhovland:
OK, let's regard this one blocked on #14340. Either the issue gets fixed there, or we fix it here when that one is merged.

epage:
>  So in this PR we are simply skipping a validation if publish = false. But that means we allow any workspace where one of the packages, 

`cargo package --workspace` should have the same behavior as `cargo package -p foo && cargo package -p no-publish`.  It sounds like this PR is trying to solve this problem by having the behavior diverge (skip the verify step).

btw how are we skipping the verify step?

jneem:
The "verify" step @torhovland is talking about isn't "the" verify step where you test-build the packages. It's the step where we're verifying that the inferred registry is valid for all the packages.

epage:
Thanks for the clarification.  I've updated the PR description to try to clarify that.

So if I'm understanding the concern, it is that if you depend on a package with `publish = false`, we won't catch that mistake during the verify step?

imo this seems less of a problem for `cargo package` and more of a problem for `cargo publish` which won't allow these to be published anyways.

jneem:
> if you depend on a package with publish = false, we won't catch that mistake during the verify step?

The goal of my suggestion is to replicate the old behavior. There are a few cases, so imagine we have a workspace with `main` depending on `no-publish`, and the current local version of `no-publish` is `0.1.1`

- if `main` depends on `no-publish 0.1.0` and `no-publish 0.1.0` is already published on crates-io, the verify step will succeed because it will pull in `no-publish` from crates-io
- if `main` depends on `no-publish 0.1.0` and it isn't published, the verify step will fail: since `no-publish` is marked as not publishable, it won't be put in the local registry overlay, and so when verifying `main` we won't find it
- if `main` depends on `no-publish 0.1.1` but only `0.1.0` is published on crates-io, the verify step will fail for the same reason

epage:
I'm not too worried about the exact behavior when a published package depends on an unpublished package that changes state between unpublishable and publishable.  That feels like a corner case not worth designing towards.  If anything, I think I'd prefer to err on the side of allowing it to be packaged (since the publish state is independent of packaging).  We could note it in the issue for consderation when stabilizing.

torhovland:
Superseded by #14408.

Issue #14348, opened 2024-08-03
Cargo vendor regression for git sources containing `exclude = ["build.rs"]`


dtolnay:
### Problem

In Rust 1.79 and older releases, `cargo vendor` used to work for crates that have build scripts applying only to their source repo, and not published to crates.io. An example of such a crate is https://github.com/dtolnay/cxx/blob/1.0.124/macro/Cargo.toml#L8. The build script was correctly omitted from being vendored.

Since Rust 1.80, Cargo still does not vendor the build script, but now it inserts `build = "build.rs"` into the vendored Cargo.toml, which causes compilation of the vendored sources to fail.

### Steps

Create an empty project with this Cargo manifest:

```toml
# Cargo.toml

[package]
name = "repro"
version = "0.0.0"
edition = "2021"
publish = false

[dependencies]
cxxbridge-macro = { git = "https://github.com/dtolnay/cxx", tag = "1.0.124" }
```

```console
$ mkdir .cargo
$ cargo vendor > .cargo/config.toml
$ cargo check
...
   Compiling cxxbridge-macro v1.0.124 (https://github.com/dtolnay/cxx?tag=1.0.124#afd4aa3f)
error: couldn't read /path/to/vendor/vendor/cxxbridge-macro/build.rs: No such file or directory (os error 2)

error: could not compile `cxxbridge-macro` (build script) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

### Possible Solution(s)

Do not inject `build = "build.rs"` into the vendored manifest if build.rs is not being vendored.

### Notes

Bisects to https://github.com/rust-lang/cargo/pull/13713/commits/1e6047763d72edf5dfaee52e1c2305b94d0b7ccc.

epage:
cxx-bridge-macro has an inferred `build.rs`

https://github.com/dtolnay/cxx/blob/master/macro%2FCargo.toml

We cover a lot of these cases for `build.rs` for package
https://github.com/rust-lang/cargo/blob/fa646583675d7c140482bd906145c71b7fb4fc2b/tests/testsuite/package.rs#L3909

But only had minimal tests for `cargo vendor`

epage:
We are using a `PathSource` to list the files to copy for vendor

https://github.com/rust-lang/cargo/blob/fa646583675d7c140482bd906145c71b7fb4fc2b/src/cargo/ops/vendor.rs#L222

but only vendoring the normlalized git `Cargo.toml`

[...]

PR #14366, opened 2024-08-07
[beta-1.81] Revert "fix: Ensure dep/feature activates the dependency on 2024"

weihanglo:
Beta backports

* https://github.com/rust-lang/cargo/pull/14295

In order to make CI pass, the following PRs are also cherry-picked:

* #14352
* Ignore `build_std::{cross_custom,custom_test_framework,remap_path_scope}` tests (see also #14368)

PR #14367, opened 2024-08-07
fix(vendor): Strip excluded build targets 

epage:

Fixes https://github.com/rust-lang/cargo/issues/14348

The commits are setup to be able to show, through the tests, how the `cargo package` and `cargo vendor` behavior diverge and then how `cargo vendor`s behavior changes over time.

This is a **very** hacky solution, duplicating the minimum of what
`prepare_for_publish` does to fix this one issue and in the least
intrusive way to the vendor code.

The intention is to keep this low risk for backporting to beta and
stable.
We need to revisit this, refactoring the `cargo package` code so that we
can call into that for each vendored dependency.


src/cargo/ops/vendor.rs
```diff
@@ -392,6 +398,114 @@ fn cp_sources(
     Ok(())
 }
 
+fn prepare_for_vendor(
```
weihanglo, line 401: Even if we plan to revisit and refactor, a notice here could be useful for people about to diverge the logic to think twice.

```suggestion
/// This is a hack for resolving cargo/#14348.
/// It basically copies the normalization logic from `prepare_for_publish`.
/// TODO: need to revisit and consolidate both for `cargo package` and `cargo vendor`. 
fn prepare_for_vendor(
```

tests/testsuite/vendor.rs
```diff
@@ -956,10 +1521,6 @@ fn git_deterministic() {
 # will likely look very different (and much more reasonable).
 # See Cargo.toml.orig for the original contents.
 
-bin = []
-test = []
-bench = []
```
weihanglo, line 961: I am sorry that I didn't even notice this when reviewing that PR.

weihanglo (approved):

src/cargo/ops/vendor.rs
weihanglo: `r=weihanglo` whenever you feel ready. The comment is not necessary.

tests/testsuite/vendor.rs
epage: Notice which?

weihanglo:
I don't think it is controversial for beta backport. #14366 contains two commits to make the beta channel CI happy. @epage, are you working on the backport? I have some free time now and can handle it if needed.

epage:
I can do the beta backport.

PR #14368, opened 2024-08-07
[beta-1.81] fix(vendor): Strip excluded build targets 

epage:
Beta backports

* #14367

In order to make CI pass, the following PRs are also cherry-picked:

* Ignore `build_std::basic` test (see also #14366)
* 830db6f06105389160436e88c852afd86edd0179 (#14341, see also #14366)

PR #11414, opened 2022-11-23
fix(vendor): Make `vendor` work similar to `package` since it resolve…

Muscraft:
In #11192 it was noted that workspace inheritance causes vendored git dependencies to fail since `workspace = true` does not get resolved during vendoring. 

To fix this, it was suggested to make `cargo vendor` work similar to `cargo package` since it resolves manifests before packaging them. I ended up implementing this proposed solution but sadly I could not find a way to share a ton of the code other than making a few things `pub`.

Issue #11192, opened 2022-10-07
Cannot build vendored git dependency if it inherits info from a workspace


hunts:
### Problem

If we `cargo vendor` our dependencies, and one of the crates in the vendor folder is inheriting package info from its Workspace, `cargo build --offline` would fail with errors like:

```
Caused by:
  error inheriting `version` from workspace root manifest's `workspace.package.version`
Caused by:
  failed to find a workspace root
```

### Steps

1. Add a crate that is from a project using Workspace inheritance
```
$ tail Cargo.toml
...
[dependencies]
lib-in-workspace = { git = "ssh://git@internal-git-service/rust-project-using-workspace.git", tag = "v0.176.0" }
...
```

2. Vendor dependencies:

```
$ cargo vendor >> .cargo/config.toml

$ tail .cargo/config.toml
...
[source."ssh://git@internal-git-service/rust-project-using-workspace.git"]
git = "ssh://git@internal-git-service/rust-project-using-workspace.git"
tag = "v0.176.0"
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "vendor"

$ cat Cargo.lock
...
[[package]]
name = "lib-in-workspace"
version = "0.176.0"
source = "git+ssh://git@internal-git-service/rust-project-using-workspace.git?tag=v0.176.0#7efcf95d618e9af79431e036414903fe3b4954b8"
dependencies = [
 "aho-corasick",
 "bstr",
 "cc",
 ...
]
...
```

3. Build app offline
```
$ cargo build --offline --release

error: failed to get `aho-corasick` as a dependency of package `app v0.1.0 (/home/hunts/code/app)`

Caused by:
  failed to load source for dependency `aho-corasick`

Caused by:
  Unable to update registry `crates-io`

Caused by:
  failed to update replaced source registry `crates-io`

Caused by:
  failed to parse manifest at `/home/hunts/code/app/vendor/lib-in-workspace/Cargo.toml`

Caused by:
  error inheriting `version` from workspace root manifest's `workspace.package.version`

Caused by:
  failed to find a workspace root
```

4. Check what's in the crate's Cargo.toml

```
$ head vendor/lib-in-workspace/Cargo.toml

[package]
name = "lib-in-workspace"
version.workspace = true
authors.workspace = true
description.workspace = true
edition.workspace = true
publish.workspace = true
```

### Possible Solution(s)

[...]

PR #14369, opened 2024-08-08
[stable-1.80]  fix(vendor): Strip excluded build targets 

epage:
Stable backports

* #14368


src/cargo/ops/vendor.rs
```diff
@@ -220,7 +220,7 @@ fn sync(
         let pathsource = PathSource::new(src, id.source_id(), gctx);
```
epage, line 220: this is targeting `rust-1.80.0`.  Did that as a placeholder until we have `rust-1.80.2`

weihanglo: That branch is for `1.80` and would never change as far as I can tell.

weihanglo (approved):

epage:
> [weihanglo](https://github.com/weihanglo) approved these changes

FYI, this intentionallty is broken.  I wanted to get this up for the conversation but I still need to work through the build breakages from other parts of cargo being different and then figure out what unblocking backports are needed.

weihanglo:
@rfcbot fcp merge

I propose to backport #14368 

There is a bug in 1.80.0/1.80.1 that `cargo vendor` generates broken unbuildable vendored Git sources when

* The git dependency contains a build script
* The build script is explicitly excluded via `package.exclude`.

epage:
FYI I've given a heads up to the release team at https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/1.2E80.2E2

weihanglo:
Should we bump the version of cargo crate to 0.81.1?

epage:
According to @pietroalbini, this doesn't meet the bar for a new point release, only piggy-backing off of an existing point release, as this is limited to `cargo vendor` users and can be worked around by running `cargo vendor` on a different version.

This aligns with our last `cargo vendor` breakage (#11192).  That was broken in 1.64 and we didn't fix it until 1.68 (#11414).  It doesn't look like we even did a beta backport for that.

weihanglo:
> This aligns with our last `cargo vendor` breakage (#11192). That was broken in 1.64 and we didn't fix it until 1.68 (#11414). It doesn't look like we even did a beta backport for that.


I have no strong inclination for stable backport after you pointed this out.

PR #14352, opened 2024-08-05
Fix renamed disallowed cfg lint name

Urgau:
### What does this PR try to resolve?

In https://github.com/rust-lang/cargo/pull/14153 we allow the to be merged https://github.com/rust-lang/rust/pull/126158 lint. Since then a FCP was conducted and the lint was renamed to `explicit_builtin_cfgs_in_flags`. This PR renames the lint in the test so that https://github.com/rust-lang/rust/pull/126158 can be merged.

### Additional information

https://github.com/rust-lang/rust/pull/126158#issuecomment-2267984311 shows the failing test.

PR #14374, opened 2024-08-09
[beta-1.81] Fix: `cargo package` failed on bare commit git repo.

weihanglo:
Beta backports

* #14359

* #14352
* Ignore `build_std::{cross_custom,custom_test_framework,remap_path_scope}` tests (see also #14368)

### DIFF
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -583,10 +583,12 @@ fn check_repo_state(
                         .and_then(|p| p.to_str())
                         .unwrap_or("")
                         .replace("\\", "/");
-                    return Ok(Some(VcsInfo {
-                        git: git(p, src_files, &repo, &opts)?,
-                        path_in_vcs,
-                    }));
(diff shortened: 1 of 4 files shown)


epage:
Whats the reason for backporting this?  Empty git repos seems like a pretty rare use case.

weihanglo:
> Whats the reason for backporting this? Empty git repos seems like a pretty rare use case.

True but 

* Still a regression. We don't know if there is an automation relying on `cargo new`–adding stuff–`cargo package` combos.
* This is a relatively simple backport.

Fine with not backporting it. Whoever does this should find any workaround easily (like making a git commit). Closing.

Turbo87:
FWIW not backporting this has broken the crates.io smoke test setup (https://github.com/rust-lang/crates.io/blob/main/crates/crates_io_smoke_test/). we're basically doing the same "steps" as in https://github.com/rust-lang/cargo/issues/14354#issue-2449185198 leading to the "revspec 'HEAD' not found" error :-/

PR #14376, opened 2024-08-09
Correctly parse arguments from environment variables

GuillaumeGomez:
Fixes https://github.com/rust-lang/cargo/issues/14381.

The current parsing doesn't correctly handle arguments like `--x="y z"`. To be noted: if you don't want the dependency, I can write a parser by hand too, shouldn't take many lines of code to do.

r? @epage

### DIFF
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -2971,8 +2971,12 @@ fn disables_multiplexing_for_bad_curl(
 mod tests {
     use super::disables_multiplexing_for_bad_curl;
     use super::CargoHttpConfig;
+    use super::ConfigKey;
+    use super::Definition;
     use super::GlobalContext;
     use super::Shell;
+
+    use std::collections::HashMap;
 
     #[test]
     fn disables_multiplexing() {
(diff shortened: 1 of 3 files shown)


src/cargo/util/context/mod.rs
```diff
@@ -973,8 +974,8 @@ impl GlobalContext {
             }
         } else {
             output.extend(
-                env_val
-                    .split_whitespace()
+                Shlex::new(env_val)
+                    .into_iter()
                     .map(|s| (s.to_string(), def.clone())),
```
epage, line 979: We'll need a test for this.

Ideally, the PR gets split up into two commits
- The first adds the test, showing the bad behavior
- This commit updates the test showing the good behavior

As discussed at https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request

epage:
I'm unsure what the compatibility guarantees are around this.  @ehuss if you have any insight, it'd be appreciated.

In general, we [recommend starting with an issue](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo).  That way we make sure we are all on the same page (e.g. the definition of the problem in this PR is unclear) and, if we go through multiple possible PRs for the problem (which is common enough), the conversation on the problem and solution space is centralized rather than split among all of the PRs.

src/cargo/util/context/mod.rs
GuillaumeGomez: Noted! Rewriting git history and using this occasion to add a test.

GuillaumeGomez:
I opened https://github.com/rust-lang/cargo/issues/14381 too.

src/cargo/util/context/mod.rs
```diff
@@ -3010,4 +3015,37 @@ mod tests {
...
+
+        for (key, value) in values {
+            let mut args = Vec::new();
+            gctx.get_env_list(&ConfigKey::from_str(key), &mut args)
+                .unwrap();
+
+            let mut expected = Vec::new();
+            for sub in *value {
+                expected.push((
+                    sub.replace("\"", ""),
```
GuillaumeGomez, line 3044: `shlex` removes (double) quotes. Not sure whether or not it's a good thing. If it's not the expected behaviour, I can implement a simple shell parser which keeps them.

src/cargo/util/context/mod.rs
epage, line 1: Blocked on discussion in #14381

ehuss:
I'm going to close as this is unlikely the direction that we want to go. Using shlex as-is would be a breaking change (for example `a\b` would be interpreted differently). If we do decide on some direction, that discussion will likely happen on https://github.com/rust-lang/cargo/issues/14381.

PR #14380, opened 2024-08-09
Use longhand gitoxide path-spec patterns

### LINKED ISSUE #14379: Unimplemented short keyword: '_' when Cargo.toml under path starting with _
### Problem

Since `rust-1.79` we have been unable to build packages that include `build.rs` within our rust workspace with cargo due to the following error.
```
error: failed to determine package fingerprint for build script for foo v0.1.0 (/Users/jemtucker/src/mvp/_code/rust/foo)

Caused by:
  failed to determine the most recently modified file in /Users/jemtucker/src/mvp/_code/rust/foo

Caused by:
  failed to determine list of files in /Users/jemtucker/src/mvp/_code/rust/foo

Caused by:
  Unimplemented short keyword: '_'
```

Our directory structure is as follows.
```
.git/...
_code/rust/Cargo.toml
_code/rust/foo/Cargo.toml
_code/rust/foo/build.rs
_code/rust/foo/src/lib.rs
```

This error occurs when gitoxide is attempting to parse pathspecs of the form `:/_code/rust` and `:!/_code/rust` which are invalid. 

The fix is to use the non-shorthand pathspec keywords e.g. `:(top)_code/rust`.

### Steps

_No response_

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

_No response_

jemtucker:
Closes #14379

### DIFF
--- a/src/cargo/sources/path.rs
+++ b/src/cargo/sources/path.rs
@@ -610,11 +610,11 @@ fn list_files_gix(
 
     let pathspec = {
         // Include the package root.
-        let mut include = BString::from(":/");
+        let mut include = BString::from(":(top)");
         include.push_str(package_prefix.as_ref());
 
         // Exclude the target directory.
-        let mut exclude = BString::from(":!/");
+        let mut exclude = BString::from(":!(exclude,top)");
         exclude.push_str(target_prefix.as_ref());
 
         vec![include, exclude]
--- a/tests/testsuite/build_script.rs
+++ b/tests/testsuite/build_script.rs
@@ -14,7 +14,7 @@ use cargo_test_support::tools;
 use cargo_test_support::{
     basic_manifest, cargo_exe, cross_compile, is_coarse_mtime, project, project_in,
 };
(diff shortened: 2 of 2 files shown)


epage:
CC @Byron

src/cargo/sources/path.rs
```diff
@@ -610,11 +610,11 @@ fn list_files_gix(
...
     let pathspec = {
         // Include the package root.
-        let mut include = BString::from(":/");
+        let mut include = BString::from(":(top)");
         include.push_str(package_prefix.as_ref());
 
         // Exclude the target directory.
-        let mut exclude = BString::from(":!/");
+        let mut exclude = BString::from(":!(exclude,top)");
         exclude.push_str(target_prefix.as_ref());
```
epage, line 618: Could you add a test for this?

Ideally this PR would be split up such that
- The first commit adds the test, showing the bad behavior
- This commit updates the test, showing the good behavior

See also https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request

PR #14397, opened 2024-08-14
doc: convert comments to rustdoc in workspace

Ifropc:
Converting workspace comments into docs and adding docs on members that were missing it.
[Original discussion](https://github.com/rust-lang/cargo/pull/14326#discussion_r1702452550)

### DIFF
--- a/src/cargo/core/workspace.rs
+++ b/src/cargo/core/workspace.rs
@@ -46,65 +46,70 @@ use pathdiff::diff_paths;
 /// package is loaded and/or learned about.
 #[derive(Debug)]
 pub struct Workspace<'gctx> {
+    /// Cargo configuration information. See [`GlobalContext`].
     gctx: &'gctx GlobalContext,
 
-    // This path is a path to where the current cargo subcommand was invoked
-    // from. That is the `--manifest-path` argument to Cargo, and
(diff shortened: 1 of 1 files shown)


src/cargo/core/workspace.rs
```diff
@@ -46,65 +46,69 @@ use pathdiff::diff_paths;
...
-    // If `true`, then the resolver will ignore any existing `Cargo.lock`
-    // file. This is set for `cargo install` without `--locked`.
+    /// If `true`, then the resolver will ignore any existing `Cargo.lock`
+    /// file. This is set for `cargo install` without `--locked`.
     ignore_lock: bool,
 
     /// The resolver behavior specified with the `resolver` field.
     resolve_behavior: ResolveBehavior,
+    /// If `true`, then workspace `rust_version` would be used in `resolve`
+    /// This is set for v3 resolver.
```
ehuss, line 111: I think this might be a bit misleading since how this gets set is a little more complicated than just the resolver version. I might say something like "This is set based on the resolver version, config settings, and CLI flags.".

Ifropc: Yeah, I was honestly trying to understand the logic and simplify the comment to not get too much into details. Like the updated description, let me fix it!

Issue #13908, opened 2024-05-13
cargo update should not highlight non-semver-compatible "(latest: v0.23.5)" for indirect dependencies


djc:
### Problem

I just ran my weekly `cargo update` against the workspace at work, and it highlighted this change:

```
    Updating rustls v0.21.11 -> v0.21.12 (latest: v0.23.5)
```

This workspace does not have a direct dependency on rustls. As such, I would suggest this information isn't actionable enough to merit a mention in the non-verbose `cargo update` output (it would definitely fit in with the verbose output).

### Version

```text
cargo 1.78.0 (54d8815d0 2024-03-26)
release: 1.78.0
commit-hash: 54d8815d04fa3816edc207bbc4dd36bf18014dbc
commit-date: 2024-03-26
host: aarch64-apple-darwin
libgit2: 1.7.2 (sys:0.18.2 vendored)
libcurl: 8.4.0 (sys:0.4.72+curl-8.6.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Mac OS 14.4.1 [64-bit]
```

weihanglo:
Thanks for the suggestion!

Agree it is not actionable locally and quite verbose, especially when you have only few direct dependencies and that's very noisy. However, I see it as an opportunity to encourage user to collaborate with upstreams. Putting in behind `--verbose` would reduce the change of it. It is also a more consistent console output that an update always comes with latest version information, but not some are behind `--verbose` some are not. 

One similar work we've done is future-compatible warnings. Cargo indeed puts them behind `cargo report`, though I am not sure if they should have the same behavior. Should also weigh on the complexity added and the verbosity it reduces.

djc:
> Agree it is not actionable locally and quite verbose, especially when you have only few direct dependencies and that's very noisy. (..) Putting in behind `--verbose` would reduce the change of it. It is also a more consistent console output that an update always comes with latest version information, but not some are behind `--verbose` some are not.

This seems like bad reasoning, since `cargo update --verbose` reveals a bunch more crates that similarly have semver-incompatible versions (but somehow these are not being shown without `--verbose`):

```
Unchanged hyper-rustls v0.24.2 (latest: v0.27.1)
```

I suppose that the `(latest ..)` notice for rustls is only shown because there **was** a semver-compatible bump, but IMO that's not a good reason to also show the `(latest ..)` notice for non-direct semver-incompatible versions.

[...]

PR #14401, opened 2024-08-14
feat(update): Report when incompatible-rust-version packages are selected

epage:

In discussin this in #13873, it highlighted that we need to make sure we
tell people when incompatible-rust-version packages are selected.

I decided to keep "latest" and "required rust version" messages mutually
exclusive to avoid too much noise.  I gave "required rust version"
higher precedence as its the more critical to operation and, if you are
using an MSRV-incompatible package, it likely is "latest" already.

### Additional information

I was tempted to change colors to make "required rust version" stand out
from "latest" but was unsure what direction to go, so I held off.
Options included
- red for "required rust version", yellow for "latest"
- yellow for "required rust version", nothing for "latest"

There is also more discussion on how to format "latest" at #13908.

### DIFF
--- a/src/cargo/ops/cargo_update.rs
+++ b/src/cargo/ops/cargo_update.rs
@@ -498,10 +498,6 @@ fn print_lockfile_generation(
     status_locking(ws, num_pkgs)?;
 
     for diff in diff {
-        fn format_latest(version: semver::Version) -> String {
-            let warn = style::WARN;
-            format!(" {warn}(latest: v{version}){warn:#}")
-        }
         let possibilities = if let Some(query) = diff.alternatives_query() {
             loop {
                 match registry.query_vec(&query, QueryKind::Exact) {
@@ -555,10 +551,6 @@ fn print_lockfile_sync(
     status_locking(ws, num_pkgs)?;
 
     for diff in diff {
-        fn format_latest(version: semver::Version) -> String {
-            let warn = style::WARN;
-            format!(" {warn}(latest: v{version}){warn:#}")
-        }
         let possibilities = if let Some(query) = diff.alternatives_query() {
             loop {
                 match registry.query_vec(&query, QueryKind::Exact) {
@@ -650,10 +642,6 @@ fn print_lockfile_updates(
 
     let mut unchanged_behind = 0;
     for diff in diff {
-        fn format_latest(version: semver::Version) -> String {
-            let warn = style::WARN;
-            format!(" {warn}(latest: v{version}){warn:#}")
-        }
         let possibilities = if let Some(query) = diff.alternatives_query() {
             loop {
                 match registry.query_vec(&query, QueryKind::Exact) {
@@ -805,6 +793,11 @@ fn status_locking(ws: &Workspace<'_>, num_pkgs: usize) -> CargoResult<()> {
         .shell()
         .status("Locking", format!("{num_pkgs} package{plural}{cfg}"))?;
     Ok(())
+}
+
+fn format_latest(version: semver::Version) -> String {
+    let warn = style::WARN;
+    format!(" {warn}(latest: v{version}){warn:#}")
 }
 
 fn is_latest(candidate: &semver::Version, current: &semver::Version) -> bool {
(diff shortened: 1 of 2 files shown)


src/cargo/ops/cargo_update.rs
```diff
@@ -795,6 +743,20 @@ fn status_locking(ws: &Workspace<'_>, num_pkgs: usize) -> CargoResult<()> {
     Ok(())
 }
 
+fn report_latest(possibilities: &[IndexSummary], package: PackageId) -> Option<String> {
+    if possibilities.is_empty() {
+        return None;
+    }
```
weihanglo, line 790: nit: [`Iterator::max`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.max) returns `None` when empty. So this is kinda redundant?

src/cargo/ops/cargo_update.rs
```diff
@@ -514,12 +514,14 @@ fn print_lockfile_generation(
         };
 
         for package in diff.added.iter() {
+            let required_rust_version = report_required_rust_version(ws, resolve, *package);
             let latest = report_latest(&possibilities, *package);
+            let note = required_rust_version.or(latest);
```
weihanglo, line 519: nit: should we make `report_latest` lazy via `or_else`?
(I know we keep this for consistency for the unchanged branch, though)

src/cargo/ops/cargo_update.rs
```diff
@@ -637,15 +641,17 @@ fn print_lockfile_updates(
...
 
             let msg = if removed.source_id().is_git() {
                 format!(
-                    "{removed} -> #{}",
+                    "{removed} -> #{}{note}",
                     &added.source_id().precise_git_fragment().unwrap()[..8],
                 )
             } else {
-                format!("{removed} -> v{}{latest}", added.version())
+                format!("{removed} -> v{}{note}", added.version())
```
weihanglo, line 654: How does it looks like when removing a package required incompatible rust version? It seems that we don't have such a test.

src/cargo/ops/cargo_update.rs
epage: The consistency with `unchanged` was my motivation with the assumption that this wouldn't have a meaningful performance impact.  Consistency has the benefit of making it easier to spot branches you need to update (or that a branch was forgotten)

src/cargo/ops/cargo_update.rs
epage: This case isn't actually removing but "Updating" or "Downgrading" and we are reporting the status of the package we are upgrading or downgrading to.

Pulled from our tests
```
-[UPDATING] bar v1.5.0 -> v1.6.0
+[UPDATING] bar v1.5.0 -> v1.6.0 (requires Rust 1.65.0)
```

A proper remove has no status being reported
```rust
            for package in diff.removed.iter() {
                ws.gctx().shell().status_with_color(
                    "Removing",
                    format!("{package}"),
                    &style::ERROR,
                )?;
            }
```

weihanglo: I was confused. Checked this branch again and yes it is not "removed". Thanks for the explanation.

weihanglo (approved):

src/cargo/ops/cargo_update.rs
```diff
@@ -807,6 +749,60 @@ fn status_locking(ws: &Workspace<'_>, num_pkgs: usize) -> CargoResult<()> {
...
 
+fn required_rust_version(ws: &Workspace<'_>) -> Option<PartialVersion> {
+    if !ws.resolve_honors_rust_version() {
+        return None;
+    }
+
+    if let Some(ver) = ws.rust_version() {
+        Some(ver.clone().into_partial())
+    } else {
+        let rustc = ws.gctx().load_global_rustc(Some(ws)).ok()?;
```
weihanglo, line 760: May not be worth a fix. I didn't benchmark it.

This will load and parse `.rustc_info.json` in a loop. May have some potential performance impact.

epage: Yeah. I had thought of that and had considered passing it around but wasn't sure if it was worth it.  We already are writing to the screen which is pretty slow.

PR #14402, opened 2024-08-14
test: Migrate some json tests to snapbox

epage:

This is part of #14039

### Additional information

This was unblocked because of assert-rs/snapbox#350

### DIFF
--- a/tests/testsuite/doc.rs
+++ b/tests/testsuite/doc.rs
@@ -1879,24 +1879,25 @@ fn doc_message_format() {
 
     p.cargo("doc --message-format=json")
         .with_status(101)
-        .with_json_contains_unordered(
-            r#"
-            {
-                "message": {
-                    "$message_type": "diagnostic",
-                    "children": "{...}",
-                    "code": "{...}",
-                    "level": "error",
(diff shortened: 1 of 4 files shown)


tests/testsuite/doc.rs
```diff
@@ -1879,24 +1879,25 @@ fn doc_message_format() {
 
     p.cargo("doc --message-format=json")
         .with_status(101)
-        .with_json_contains_unordered(
```
weihanglo, line 1882: Does the snapbox patch now correctly handle unordered JSON lines, or this test doesn't need unordered assertion?

epage: If by "now correctly" you mean "implicitly", then no, as that would not be correct.  I am not seeing anything in this test case that needs `.unordered()` applied.

weihanglo: I meant to comment under `doc_json_artifacts` but that one doesn't seem to need unordered either.

weihanglo (approved):

weihanglo:
@bors r+

Thanks!

PR #14404, opened 2024-08-15
feat: add CARGO_MANIFEST_PATH env variable

Ifropc:
Adds `CARGO_MANIFEST_PATH` variable as part of https://github.com/rust-lang/cargo/issues/12207
Context: `CARGO_MANIFEST_DIR` is not very useful, because there is no `Cargo.toml` file when running a cargo script. In cases when multiple scripts are stored in the same folder, we can't tell which script exactly is being run using `CARGO_MANIFEST_DIR`

### DIFF
--- a/src/cargo/core/compiler/compilation.rs
+++ b/src/cargo/core/compiler/compilation.rs
@@ -356,6 +356,7 @@ impl<'gctx> Compilation<'gctx> {
         // in BuildContext::target_metadata()
         let rust_version = pkg.rust_version().as_ref().map(ToString::to_string);
         cmd.env("CARGO_MANIFEST_DIR", pkg.root())
+            .env("CARGO_MANIFEST_PATH", pkg.manifest_path())
             .env("CARGO_PKG_VERSION_MAJOR", &pkg.version().major.to_string())
             .env("CARGO_PKG_VERSION_MINOR", &pkg.version().minor.to_string())
             .env("CARGO_PKG_VERSION_PATCH", &pkg.version().patch.to_string())
--- a/src/doc/src/reference/environment-variables.md
+++ b/src/doc/src/reference/environment-variables.md
@@ -226,6 +226,7 @@ corresponding environment variable is set to the empty string, `""`.
 
 * `CARGO` --- Path to the `cargo` binary performing the build.
 * `CARGO_MANIFEST_DIR` --- The directory containing the manifest of your package.
+* `CARGO_MANIFEST_PATH` --- The path to the manifest of your package.
 * `CARGO_PKG_VERSION` --- The full version of your package.
 * `CARGO_PKG_VERSION_MAJOR` --- The major version of your package.
 * `CARGO_PKG_VERSION_MINOR` --- The minor version of your package.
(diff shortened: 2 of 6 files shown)


src/doc/src/reference/environment-variables.md
```diff
@@ -225,7 +225,8 @@ Note that if one of these values is not provided in the manifest, the
 corresponding environment variable is set to the empty string, `""`.
 
 * `CARGO` --- Path to the `cargo` binary performing the build.
-* `CARGO_MANIFEST_DIR` --- The directory containing the manifest of your package.
+* `CARGO_MANIFEST_DIR` --- The directory containing the manifest of your package. When running a [Cargo script], this variable will be set to the script location.
```
epage, line 228: That wording change is ambiguous.  To me, it reads as we change the meaning to point to `CARGO_MANIFEST_PATH`.

I don't think this change is needed and it is unrelated to the intent of this PR, so let's remove it.

epage:
@rfcbot merge

This PR insta-stabilizes `CARGO_MANIFEST_PATH`, a companion to `CARGO_MANIFEST_DIR` for when you actually want the manifest because  `CARGO_MANIFEST_DIR.join("Cargo.toml")` will no longer be sufficient due to cargo-script.

src/cargo/core/compiler/compilation.rs
```diff
@@ -356,6 +356,7 @@ impl<'gctx> Compilation<'gctx> {
         // in BuildContext::target_metadata()
         let rust_version = pkg.rust_version().as_ref().map(ToString::to_string);
         cmd.env("CARGO_MANIFEST_DIR", pkg.root())
+            .env("CARGO_MANIFEST_PATH", pkg.manifest_path())
```
weihanglo, line 359: This change needs an FCP, and maybe an unstable gate before `-Zscript` stabilization.

weihanglo: Resolved by <https://github.com/rust-lang/cargo/pull/14404#issuecomment-2292019184> 😁.

epage: I figure this is small enough in scope and likely useful enough outside of cargo-script that we could stabilize now, with an FCP, independent of the cargo-script progress.

src/doc/src/reference/environment-variables.md
Ifropc: Oops the second sentence should've been in `CARGO_MANIFEST_PATH` let me fix that. Accidentally put it to the DIR instead. ([fix diff](https://github.com/rust-lang/cargo/compare/994083b14c1b9905a242419c15246be0ee5e0d23..d9e1028c1f77b6f11c0227fcd1de967008e87389))
LMK if you still think it's better to just remove it

epage: Let's leave it off for now.  We'll need to decide how we document cargo script.  Besides, most unstable documentation gets left in https://doc.rust-lang.org/cargo/reference/unstable.html

epage (approved):

tests/testsuite/script.rs
weihanglo, line 1: I understand that this patch is pretty self-explained, but should we verify `CARGO_MANIFEST_PATH` in the `crate_env_vars` test? It is a pretty a standard env var Cargo set, not only for `cargo -Zscript`.

https://github.com/rust-lang/cargo/blob/d7bffc302ef83dd8a4bf06bd4fb7579e7988263e/tests/testsuite/build.rs#L1640-L1642

weihanglo (approved):

weihanglo (approved):

weihanglo:
Thanks for working with us! Glad to have you along!

I am going to merge this. This feature is not controversial at all, so no need to wait for the full 10-days FCP. Please speak up if you have any concern.

@bors r+

PR #14408, opened 2024-08-15
Be more permissive while packaging unpublishable crates.

### LINKED ISSUE #14356: `-Zpackage-workspace` is not smart about `publish = false`

jneem:
This PR allows for packaging workspaces that include `publish = false` crates, in some circumstances:

- unpublishable crates are ignored when inferring the publish registry
- when checking whether the publish registry is valid, we ignore unpublishable crates
- we don't put unpublishable crates in the local registry overlay, so if any workspace crates depend on an unpublishable crate then they will fail to verify.

This PR also contains a refactor, moving the registry inference logic to `registry/mod.rs`, where it will be reused by the upcoming publish-workspace feature. I put the refactor and the logic changes in different commits.

Fixes #14356

### DIFF
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -243,7 +243,9 @@ pub fn package(ws: &Workspace<'_>, opts: &PackageOpts<'_>) -> CargoResult<Vec<Fi
         } else {
             let tarball = create_package(ws, &pkg, ar_files, local_reg.as_ref())?;
             if let Some(local_reg) = local_reg.as_mut() {
-                local_reg.add_package(ws, &pkg, &tarball)?;
+                if pkg.publish() != &Some(Vec::new()) {
+                    local_reg.add_package(ws, &pkg, &tarball)?;
+                }
             }
             outputs.push((pkg, opts, tarball));
         }
@@ -284,7 +286,10 @@ fn get_registry(
     if let RegistryOrIndex::Registry(reg_name) = reg {
         for pkg in pkgs {
             if let Some(allowed) = pkg.publish().as_ref() {
-                if !allowed.iter().any(|a| a == &reg_name) {
+                // If allowed is empty (i.e. package.publish is false), we let it slide.
+                // This allows packaging unpublishable packages (although packaging might
+                // fail later if the unpublishable package is a dependency of something else).
+                if !allowed.is_empty() && !allowed.iter().any(|a| a == &reg_name) {
                     bail!(
                         "`{}` cannot be packaged.\n\
                          The registry `{}` is not listed in the `package.publish` value in Cargo.toml.",
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -6,7 +6,6 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::task::Poll;
 
-use super::RegistryOrIndex;
 use crate::core::compiler::{BuildConfig, CompileMode, DefaultExecutor, Executor};
 use crate::core::dependency::DepKind;
 use crate::core::manifest::Target;
@@ -15,6 +14,7 @@ use crate::core::resolver::HasDevUnits;
 use crate::core::{Feature, PackageIdSpecQuery, Shell, Verbosity, Workspace};
 use crate::core::{Package, PackageId, PackageSet, Resolve, SourceId};
 use crate::ops::lockfile::LOCKFILE_NAME;
+use crate::ops::registry::RegistryOrIndex;
 use crate::sources::registry::index::{IndexPackage, RegistryDependency};
 use crate::sources::{PathSource, SourceConfigMap, CRATES_IO_REGISTRY};
 use crate::util::cache_lock::CacheLockMode;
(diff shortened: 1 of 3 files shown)


src/cargo/ops/cargo_package.rs
```diff
@@ -354,6 +316,12 @@ impl LocalDependencies {
             .map(|name| self.packages[&name].clone())
             .collect()
     }
+
+    pub fn has_no_dependencies(&self) -> bool {
```
epage, line 325: This commit seems to have multiple refactors in it, not just moving of a function.  I'm having a hard time following what each change is intended to do.  Could you split this further up?

src/cargo/ops/registry/mod.rs
```diff
@@ -325,39 +325,52 @@ pub(crate) struct RegistrySourceIds {
...
-                regs.dedup();
-                // unwrap: the match block ensures that there's more than one reg.
-                let (last_reg, regs) = regs.split_last().unwrap();
-                bail!(
-                    "--registry is required to disambiguate between {} or {} registries",
-                    regs.join(", "),
-                    last_reg
-                )
+    // Ignore "publish = false" packages while inferring the registry.
+    let pkgs: Vec<_> = pkgs
```
epage, line 329: While I'm a big fan of shadowing, its generally best when the meanings are the same.  These have two different meanings.

src/cargo/ops/registry/mod.rs
```diff
@@ -325,39 +325,52 @@ pub(crate) struct RegistrySourceIds {
...
+        .iter()
+        .filter(|p| p.publish() != &Some(Vec::new()))
+        .collect();
+
+    match &pkgs[..] {
+        [] => Ok(None),
+        [first, rest @ ..] => {
```
epage, line 336: If these are our only cases, this seems better served as a guard statement.  This causes a lot of unnecessary rightward drift and it gives the impression there will be other interesting cases when there aren't, breaking users expectations when reading the code, making it harder to read.

src/cargo/ops/cargo_package.rs
jneem: I've split the second commit into a pure move, followed by some changes. "Refactor" was probably the wrong word; there are a few logic changes because I wanted to reuse `publish`'s `get_source_id` logic for `package`.

src/cargo/ops/registry/mod.rs
epage: By guard statement, I was referring to an early return, something like changing
```rust
if let Some((first, rest)) = publishable_pkgs.split_first() {
```
to
```rust
let Some((first, rest)) = publishable_pkgs.split_first()  else {
    return Ok(None);
};
```

jneem:
Ok, I've split that commit further. We can drop the "Use the shared source building..." commit if you want; it changes some behavior mostly just to get some more code re-use.

src/cargo/ops/cargo_package.rs
```diff
@@ -202,19 +202,34 @@ pub fn package(ws: &Workspace<'_>, opts: &PackageOpts<'_>) -> CargoResult<Vec<Fi
...
+        Ok(sid) => {
+            debug!("packaging for registry {}", sid);
+            Some(sid)
+        }
+        Err(e) => {
+            if deps.has_no_dependencies() && opts.reg_or_index.is_none() {
+                // The publish registry doesn't matter unless there are local dependencies,
+                // so ignore any errors if we don't need it. If they explicitly passed a registry
+                // on the CLI, we check it no matter what.
+                None
```
epage, line 218: I feel uncomfortable taking any error a function may return and ignoring it based on certain criteria, assuming its a very specific case.  Someone could refactor that function, changing error cases and not realize that it doesn't propagate through.

epage:
Thank you for splitting things out further.  That was a big help in following what is happening!

src/cargo/ops/cargo_package.rs
jneem: Good point, what about this new version? It just avoids calling the function altogether instead of ignoring the error.

PR #14412, opened 2024-08-16
fix: Limiting pre-release match semantics to use only on `OptVersionReq::Req`

linyihai:

The prerelease matches semantics should be only used on `OptVersionReq::Req`, but it dosn't.  The other `OptVersionReq` types have specify matches logic already,  for example a `OptVersionReq::Precise` will failed on `matches_prerelease`, see https://github.com/rust-lang/cargo/pull/14140#discussion_r1682777074

the first commit added the test, the second commit updated the test and fixed the issue.

### DIFF
--- a/src/cargo/util/semver_ext.rs
+++ b/src/cargo/util/semver_ext.rs
@@ -178,7 +178,10 @@ impl From<VersionReq> for OptVersionReq {
 
 #[cfg(test)]
 mod matches_prerelease {
+    use semver::VersionReq;
+
     use super::OptVersionReq;
+    use super::Version;
 
     #[test]
     fn prerelease() {
@@ -237,5 +240,20 @@ mod matches_prerelease {
             let matched = OptVersionReq::Req(version_req).matches_prerelease(&version);
             assert_eq!(expected, matched, "req: {req}; ver: {ver}");
         }
+    }
+
+    #[test]
+    fn opt_version_req_matches_prerelease() {
+        let req_ver: VersionReq = "^1.2.3-rc.0".parse().unwrap();
+        let to_ver: Version = "1.2.3-rc.0".parse().unwrap();
+
+        let req = OptVersionReq::Req(req_ver.clone());
+        assert!(req.matches_prerelease(&to_ver));
+
+        let req = OptVersionReq::Locked(to_ver.clone(), req_ver.clone());
+        assert!(!req.matches_prerelease(&to_ver));
+
+        let req = OptVersionReq::Precise(to_ver.clone(), req_ver.clone());
+        assert!(!req.matches_prerelease(&to_ver));
     }
 }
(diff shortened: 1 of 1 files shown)


src/cargo/util/semver_ext.rs
```diff
@@ -238,4 +241,19 @@ mod matches_prerelease {
             assert_eq!(expected, matched, "req: {req}; ver: {ver}");
         }
     }
+
+    #[test]
+    fn opt_version_req_mathes_prerelease() {
```
epage, line 246:
```suggestion
    fn opt_version_req_matches_prerelease() {
```

src/cargo/util/semver_ext.rs
```diff
@@ -238,4 +241,19 @@ mod matches_prerelease {
...
+
+        let req = OptVersionReq::Req(req_ver.clone());
+        assert!(req.matches_prerelease(&to_ver));
+
+        let req = OptVersionReq::Locked(to_ver.clone(), req_ver.clone());
+        assert!(!req.matches_prerelease(&to_ver));
+
+        let req = OptVersionReq::Locked(to_ver.clone(), req_ver.clone());
+        assert!(!req.matches_prerelease(&to_ver));
+    }
```
epage: Am I missing something or are these two cases the same?

epage:
> matches_prerelease should be only used on OptVersionReq::Req, but it dosn't. 

Can you explain why?

linyihai:
Thanks your swift review and forgive my typos.

> matches_prerelease should be only used on OptVersionReq::Req, but it dosn't.

The currrent approach of `matches_prerelease` ignores the `Prerelease` tag,   which will allows updating to any  semver-compatible pre-rerelease version. (This approach is temporray).  

But the `OptVersionReq::Locked` and  `OptVersionReq::Precise` involve  the `Prerelease` tag to match, they will fail in match if the VersiongReq is pre-release.  

The current use of this function is  `OptVersionReq::Req`, 
https://github.com/rust-lang/cargo/blob/d585829270c758468b9fd20b2b7fd3d4f4557763/src/cargo/sources/registry/mod.rs#L750-L765

https://github.com/rust-lang/cargo/blob/d585829270c758468b9fd20b2b7fd3d4f4557763/src/cargo/sources/registry/mod.rs#L795-L809


so it's safe to limit its usage

src/cargo/util/semver_ext.rs
```diff
@@ -116,7 +116,7 @@ impl OptVersionReq {
     /// The logic here is temporary, we'll have to consider more boundary conditions later,
     /// and we're not sure if this part of the functionality should be implemented in semver or cargo.
     pub fn matches_prerelease(&self, version: &Version) -> bool {
-        if version.is_prerelease() {
+        if version.is_prerelease() && matches!(self, OptVersionReq::Req(_)) {
```
epage, line 119: nit: to make the semantics more clear, we should have `matches!(self, OptVersionReq::Req(_))` on the outside (left side) as that is guarding our overriding of the   `matches` logic to get the pre-release behavior.

PR #14413, opened 2024-08-16
Correct diagnostic for `TomlDebugInfo`

ctz:
This is missing the friendly aliases introduced in 3dbb474c.

I wrote: 

```
[profile.bench]
debug = "all"
```

I got this diagnostic:

```
error: invalid value: string "all", expected a boolean, 0, 1, 2, "line-tables-only", or "line-directives-only"
  --> Cargo.toml:31:9
   |
31 | debug = "all"
   |         ^^^^^
   |
```

I meant to write "full", as [documented here](https://doc.rust-lang.org/cargo/reference/profiles.html#debug), but that wasn't mentioned in the diagnostic.

Updated the existing tests for this diagnostic.


epage:
Thanks!

Ideally, we provide a way for this to be auto-updated *but* that is a concern out of the scope of just fixing it.

@bors r+

Issue #14001, opened 2024-06-02
Building in-tree allows brotli feature simd in dependency; building with install --git errors "error[E0554]: `#![feature]` may not be used on the stable release channel"


nabijaczleweli:
### Problem

As given in title.

### Steps

1. `git clone --branch v2.0.0 https://github.com/thecoshman/http`
2. `cargo build` it from the directory (works)
3. `cargo install --git --branch v2.0.0 https://github.com/thecoshman/http` (or whatever the spelling is, this is currently HEAD) – 
```
   Compiling rand v0.7.3
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> /home/nabijaczleweli/.cargo/registry/src/index.crates.io-6f17d22bba15001f/brotli-6.0.0/src/lib.rs:8:39
  |
8 | #![cfg_attr(feature = "simd", feature(portable_simd))]
  |                                       ^^^^^^^^^^^^^
```
4. Removing features = simd from brotli fixes it.

### Notes

May have something to do with `.cargo/config.toml` setting `env.RUSTC_BOOTSTRAP = 1`? But the thing it's used for in the crate itself *does* work.

### Version

```text
$ cargo version --verbose
cargo 1.78.0 (54d8815d0 2024-03-26)
release: 1.78.0
commit-hash: 54d8815d04fa3816edc207bbc4dd36bf18014dbc
commit-date: 2024-03-26
host: x86_64-unknown-linux-gnu
libgit2: 1.7.2 (sys:0.18.2 vendored)
libcurl: 8.6.0-DEV (sys:0.4.72+curl-8.6.0 vendored ssl:OpenSSL/1.1.1w)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Debian 12 (bookworm) [64-bit]
```

Eh2406:
The config `env.RUSTC_BOOTSTRAP = 1` is telling stable rust to ignore that error. And config files are ignored for install. That is why you are seeing the error. `RUSTC_BOOTSTRAP` is a very heavy hammer, one only intended for building the compiler itself.

nabijaczleweli:
And yet it is the only way to achieve feature parity of win32 targets with unix ones, since I use it for st_dev and st_ino (https://github.com/rust-lang/rust/issues/63010). But, whatever.

Indeed, I haven't tested cargo install --git on windows, but this does also break that:
```
error[E0554]: `#![feature]` may not be used on the stable release channel
[...]

Issue #2930, opened 2016-07-29
Cargo does not always search for .cargo/config file in project root


bennofs:
Currently, the search path for `.cargo/config` files is dependent on the current working directory at the time when cargo is executed. I find this behavior very surprising: I would expect that cargo at least always reads the `.cargo/config` file at the root of the project it builds.

It is important to note though that for some applications, this is probably the right thing to do (for example, cargo new does not have any associated project root to work with).

##### Examples

```
$ cargo build --manifest-path foo/Cargo.toml # will not read foo/.cargo/config
$ cargo build -p foo # in a workspace with foo, will not read foo's .cargo/config
```

##### Test case demonstrating the problem

```rust
#[test]
fn search_config_relative_to_project_root() {
    let foo = project("foo")
        .file("Cargo.toml", r#"
            [package]
            name = "foo"
            authors = []
            version = "0.0.0"
            build = "build.rs"
        "#)
        .file("src/lib.rs", "")
        .file("build.rs", r#"
            use std::env;
            fn main() {
                if env::var("NUM_JOBS").unwrap() == "5673" {
                    panic!("Config value `jobs` taken from config file located in working directory \
                            and not from the config file at foo's project root.");
                }

                if env::var("NUM_JOBS").unwrap() != "2423" {
                    panic!("Config value `jobs` not taken from config file located at foo's project root \
                            but config file in working directory was not read either.");
                }
            }
        "#)
        .file(".cargo/config", r#"
          [build]
          jobs = 2423
        "#);

    let bar = project("bar")
        .file(".cargo/config", r#"
          [build]
          jobs = 5673
        "#);

    foo.build();
    assert_that(bar.cargo_process("build")
                   .arg("--manifest-path").arg(foo.root().join("Cargo.toml")),
                execs().with_status(0));
}
```

eberkund:
Possibly related: https://forum.snapcraft.io/t/bug-rust-plugin-not-using-correct-pkg-config-path/7302/2

benma:
I agree, the config resolution should not be based on the current working directory, but starting at the project (where Cargo.toml is, i.e. `manifest-path`).

duskmoon314:
Is there any progress? I wonder how can I organize more than one crate in a directory having same dependencies but different  config.toml ?

It seems that now I can only manage their cargo.toml and config.toml by hand.

dvdhrm:
[...]

PR #14416, opened 2024-08-16
Implement path-bases (RFC 3529) 2/n: support for nested packages

dpaoliello:
Cargo has the ability to discover "nested packages" when loading the dependencies of a package. For example, if a git dependency has a path dependency pointing to somewhere else in the repo, or a submodule, then Cargo needs to discover those dependencies as well.

This change makes the "nested packages" loading code aware of the "path bases" feature so that it can build the correct manifest path for any path dependencies that may be using a `base`.

### DIFF
--- a/src/cargo/sources/path.rs
+++ b/src/cargo/sources/path.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::fmt::{self, Debug, Formatter};
 use std::fs;
@@ -5,7 +6,9 @@ use std::io;
 use std::path::{Path, PathBuf};
 use std::task::Poll;
 
-use crate::core::{Dependency, EitherManifest, Manifest, Package, PackageId, SourceId};
+use crate::core::{
+    find_workspace_root, Dependency, EitherManifest, Manifest, Package, PackageId, SourceId,
+};
 use crate::ops;
 use crate::sources::source::MaybePackage;
 use crate::sources::source::QueryKind;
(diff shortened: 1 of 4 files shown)


epage:
> Cargo has the ability to discover "nested packages" when loading the dependencies of a package. For example, if a git dependency has a path dependency pointing to somewhere else in the repo, or a submodule, then Cargo needs to discover those dependencies as well.
>
> This change makes the "nested packages" loading code aware of the "path bases" feature so that it can build the correct manifest path for any path dependencies that may be using a base.
>
> Note that this change also fixes a bug in the "nested package" loading code: specifically, that all nested packages were read using the GlobalContext of the "root" package, effectively ignoring any config.toml files of those nested packages.

I believe the existing behavior is what is "expected" and this PR is a regression in Cargo.

dpaoliello:
> I believe the existing behavior is what is "expected" and this PR is a regression in Cargo.
> 
> EDIT: To expand on this, configuration is environmental configuration and not project configuration. The latter is what `Cargo.toml` is for but we aren't supporting `Cargo.toml` yet. See also #14001, #2930

Ah, ok, #2930 is what I was tripping over when writing my test: I'll adjust the test and revert the regression.

I'll also add a note to the docs to mention this: if you depend on a package that uses a path base, then that path base must be defined in the configuration loaded for the root package.

tests/testsuite/git.rs
```diff
@@ -380,6 +382,109 @@ hello world
...
+    git::add_submodule(&repo, &url, Path::new("submods/dep2"));
+    git::commit(&repo);
+
+    write_config_at(
+        paths::home().join(".cargo/config.toml"),
+        r#"
+            [path-bases]
+            submodules = '../dep1/submods'
+        "#,
+    );
```
epage, line 446: imo writing path-bases on behalf of git dependencies seems very questionable
- Constructing the right absolute path is a mess
- Constructing the right relative path is a mess

If you want to test `RecursivePathSource`, https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#paths-overrides is another user of that tyoe

src/cargo/sources/path.rs
```diff
@@ -1000,8 +1005,31 @@ fn read_nested_packages(
...
+                    Ok(base) => Cow::Owned(base.join(p)),
+                    Err(err) => {
+                        errors.push(err);
+                        continue;
+                    }
+                }
+            } else {
+                Cow::Borrowed(p)
+            };
+            let path = paths::normalize_path(&path.join(p.as_path()));
```
epage, line 1032: Why is this needed?  Aren't we operating on a normalized package and don't we resolve path-bases during normaization?

For either of us to verify that is one of the reasons I recommend splitting commits into
- Test showing the existing behavior
- Fix/feature with test updates to show how the behavior changed

dpaoliello:
Looks like this change isn't needed: #2930 really confused me.
Sorry for the noise!

PR #14417, opened 2024-08-17
Lockfile path tests (follow-up)

Ifropc:
Addressing comments by [Weihang](https://github.com/rust-lang/cargo/pull/14326#issuecomment-2293664459) and [Ed](https://github.com/rust-lang/cargo/pull/14326#issuecomment-2293671992) from #14326

### DIFF
--- a/tests/testsuite/lockfile_path.rs
+++ b/tests/testsuite/lockfile_path.rs
@@ -7,7 +7,7 @@ use snapbox::str;
 use cargo_test_support::compare::assert_e2e;
 use cargo_test_support::registry::{Package, RegistryBuilder};
 use cargo_test_support::{
-    basic_bin_manifest, cargo_test, project, symlink_supported, Execs, ProjectBuilder,
+    basic_bin_manifest, cargo_test, project, symlink_supported, ProjectBuilder,
 };
 
 #[cargo_test]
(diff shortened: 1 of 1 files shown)


tests/testsuite/lockfile_path.rs
weihanglo, line 1: While this patch looks like, we'd better follow [the atomic commit patterns](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request). That is, nightly gate tests are in one commit. Removing `make_exec` helper in the other.

weihanglo: Could also be two separate PR, as they are not really related. I am fine with one PR though.

Ifropc: Still learning the best way of doing commits for rust ecosystem 👍🏼 
I will just split it into 2 separate commits, I think logically one PR should be just fine

weihanglo (approved):
Looks better. Thank you!

@bors r+

weihanglo:
I forgot. We also need an unstable doc. Something like <https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#precise-pre-release>. See the new tracking issue <https://github.com/rust-lang/cargo/issues/14421>

Issue #14409, opened 2024-08-15
`cargo info` requires login


ehuss:
### Problem

Running `cargo info foo` seems to require a login to crates.io. I would not expect a simple information command to require authentication (or for those with a hardware token, a hardware key) to access it.

### Steps

1. Set up config.toml with:
```toml
[registry]
global-credential-providers = ["false"]
```
2. Run `cargo info syn`
3. See that it fails:
```
    Updating crates.io index
error: credential provider `false` failed action `get`

Caused by:
  failed to deserialize hello

Caused by:
  EOF while parsing a value at line 1 column 0
```

### Possible Solution(s)

I'm guessing this is because of the call to `try_list_owners`. I do not think `cargo info` should be trying to access API endpoints that require authentication.

### Version

```text
commit-hash: ff6df29857a5fc5f6bb0b9da8436cb501ee5894c
```

Rustin170506:
@rustbot claim

I will take a look.

Rustin170506:
I can reproduce it.

Rustin170506:
> I do not think `cargo info` should be trying to access API endpoints that require authentication.

From the current design, if we can access the owner list, then we will display it; otherwise, it will skip the missing token error.

I think the bug from here:
https://github.com/rust-lang/cargo/blob/c76c8fd9aadff705336d348676d87b15a86b50ce/src/cargo/ops/registry/info/mod.rs#L239

We only checked the missing token error, maybe we should ignore all errors here.

The reason we introduced this information is that sometimes knowing who maintained the crate should be helpful.

ehuss:
> We only checked the missing token error, maybe we should ignore all errors here.

It's not an issue of checking for errors. Credentials can pop up requests for passwords, biometrics, or hardware keys. I do not think it would be good if `cargo info` pops up a password request every time you try to use it.

AFAIK, the [`owners list`](https://doc.rust-lang.org/nightly/cargo/reference/registry-web-api.html#owners-list) API endpoint does not require authentication on crates.io. Does the call to `registry()` require the `Operation::Read` entitlement?

Rustin170506:
> It's not an issue of checking for errors. Credentials can pop up requests for passwords, biometrics, or hardware keys. I do not think it would be good if `cargo info` pops up a password request every time you try to use it.

I see. I didn't consider that. You are right. It doesn't make sense at all.



[...]

PR #13972, opened 2024-05-28
WIP: feat(crates-io): add the downloads API

Rustin170506:
### What does this PR try to resolve?

Ref https://github.com/hi-rustin/cargo-information/issues/20

To report the crates.io metrics on cargo-information, a new API is required to obtain the download count from crates.io.

So in this PR, I added the downloads API to the `crates-io` crate.

### How should we test and review this PR?

[ ] I will test it on the cargo information side as an integration test case.

### Additional information

r? @ghost

PR #14418, opened 2024-08-17
fix: remove list owners feature of info subcommand

Rustin170506:

close https://github.com/rust-lang/cargo/issues/14409

Because this API requires authentication, the cargo info command should prompt the user to enter a password every time it is used(for those with a hardware token, a hardware key).

Since this commit https://github.com/rust-lang/cargo/commit/0c25226b35447a51918d5a17ba62a36b109757fe, authentication is required for this API.

So we cannot simply 'fix' it. So we decided to delete this feature first.

Check the updated test snapshots.

### Additional information

None

### DIFF
--- a/src/cargo/ops/registry/info/mod.rs
+++ b/src/cargo/ops/registry/info/mod.rs
@@ -1,9 +1,7 @@
 //! Implementation of `cargo info`.
 
 use anyhow::bail;
-use cargo_credential::Operation;
 use cargo_util_schemas::core::{PackageIdSpec, PartialVersion};
-use crates_io::User;
 
 use crate::core::registry::PackageRegistry;
 use crate::core::{Dependency, Package, PackageId, PackageIdSpecQuery, Registry, Workspace};
@@ -12,7 +10,6 @@ use crate::ops::registry::{get_source_id_with_package_id, RegistryOrIndex, Regis
 use crate::ops::resolve_ws;
 use crate::sources::source::QueryKind;
 use crate::sources::{IndexSummary, SourceConfigMap};
-use crate::util::auth::AuthorizationErrorReason;
 use crate::util::cache_lock::CacheLockMode;
 use crate::util::command_prelude::root_manifest;
 use crate::{CargoResult, GlobalContext};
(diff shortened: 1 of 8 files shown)


src/cargo/ops/registry/info/mod.rs
weihanglo, line 1: > So we cannot simply 'fix' it

Does this mean owner information may not be added back when `cargo info` hits stable? If it is the case, should we need a regression test ensuring this not happen again?

Rustin170506: I will open a new issue to discuss this, as we need to consider including the download number [in this PR](https://github.com/rust-lang/cargo/pull/13972). It may also be worth considering using a special API to fetch the owners list. However, we need to discuss this in the new issue.

When I say "fix it," I refer to not requiring authentication for this API. We cannot simply change that because, in our document, we state that it does require authentication. See [owners list](https://doc.rust-lang.org/nightly/cargo/reference/registry-web-api.html#owners-list)

weihanglo: I understood that. I was talking about the necessity of adding regression tests in this PR. If the feature requires more discussions and is not expected to be handled in a short term, maybe regression tests are worthy. What do you think?

weihanglo (approved):
Confirm that with this `.cargo/config.toml` 

```
[registry]
global-credential-providers = ["false"]
```

`cargo info syn` succeeds to query crate information without requiring registry authentication.

src/cargo/ops/registry/info/mod.rs
Rustin170506: Not sure what kind of regression tests you are referring to. Do you mean to check there is no owners output or check for no credential operation output?

Rustin170506: If it is the latter one, I guess we can check it in the 'verbose' case.

weihanglo: https://github.com/rust-lang/cargo/pull/14418#pullrequestreview-2244114099

The command won't failed under the reproduction step described in #14409

weihanglo:
@bors r+

Going to merge this as-is. Fine with or without regression test. Thank you!

src/cargo/ops/registry/info/mod.rs
Rustin170506: Oh, I understand now. I was mistaken. I forgot that we can use that case. 😆 I will add a regression test for it.

Issue #14421, opened 2024-08-18
Tracking Issue for `-Zlockfile-path`


weihanglo:
### Summary

Original issue: #5707
Implementation: #14326
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#lockfile-path

The `-Zlockfile-path` flag enables the `resolver.lockfile-path` configuration option,
which allows you to specify the path of the lockfile `Cargo.lock`.

### Unresolved Issues

- [ ] Inconsistency in `resolver` table being applied to `cargo install`
- [ ] How this should work if the lockfile or a parent directory is not present
  - In [#14326], missing parent directories will be created when Cargo is about to write the lockfile. This minimizes the chance of creating dangling directories. 
- [ ] `cargo install --lockfile-path=<path>` implies `--locked`, and requires the alternative lockfile to exist.
  - This behavior diverges from other commands, which allow the absence of lockfile at the alternative path.
- [x] #15510

[#14326]: https://github.com/rust-lang/cargo/pull/14326

### Tasks

- [x] Add an unstable doc under <https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#lockfile-path> #14423
- [x] `cargo install` should make `--lockfile-path` imply `--locked` #14556
- [x] Add support for `resolver.lockfile-path` in #16510
- [x] Expand `resolver.lockfile-path` coverage in #16617
- [ ] Remove `--lockfile-path` in 1.95.0-nightly

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

Ifropc:
As far as shell completions go, I don't think we need to do anything in the scope of this task. In fact, it's alreadyready working on the latest nightly (as part of https://github.com/rust-lang/cargo/issues/14520)
![Image](https://github.com/user-attachments/assets/2846f465-4c09-4169-8e33-b146c757e2da)

epage:
@Ifropc thats the new, unstable completion system.

The task is referring to our handwritten, stable completions
- https://github.com/rust-lang/cargo/blob/master/src/etc/_cargo
- https://github.com/rust-lang/cargo/blob/master/src/etc/cargo.bashcomp.sh

Ifropc:
[...]

PR #14423, opened 2024-08-18
doc: add lockfile-path unstable doc section

Ifropc:
Adds lockfile-path into the unstable doc section as part of tracking issue https://github.com/rust-lang/cargo/issues/14421 tasks

### DIFF
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -115,6 +115,7 @@ Each new feature described below should explain how to use it.
 * Other
     * [gitoxide](#gitoxide) --- Use `gitoxide` instead of `git2` for a set of operations.
     * [script](#script) --- Enable support for single-file `.rs` packages.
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/unstable.md
```diff
@@ -112,6 +112,8 @@ Each new feature described below should explain how to use it.
 * Registries
     * [publish-timeout](#publish-timeout) --- Controls the timeout between uploading the crate and being available in the index
     * [asymmetric-token](#asymmetric-token) --- Adds support for authentication tokens using asymmetric cryptography (`cargo:paseto` provider).
+* Mafinest and lockfile
```
Ifropc, line 115: Not sure if we should add another section, but existing ones don't seem to be relative enough for this feature, WDYT?

weihanglo: Maybe under Other suffices

weihanglo (approved):

weihanglo:
Thank you for updating the doc!

@bors r+

PR #14425, opened 2024-08-18
fix: doctest respects Cargo's color options

### LINKED ISSUE #14403: `cargo t --doc` does not respect `--color` when a compilation error occurs
### Problem

When using `cargo` to run doctests, `--color` is not respected during a compilation failure; however `cargo t --tests` _does_ respect it and `rustdoc --test` _almost_ respects it. Similarly `cargo t --doc` does respect `--color` when a compilation failure does not occur.

### Steps

```rust
/// ```
/// bar
/// ```
pub fn foo() {}
#[cfg(test)]
mod tests {
    #[test]
    fn foo() {
        bar
    }
}
```

```
[zack@laptop src]$ cargo t --color never --doc -- --color never
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.00s
   Doc-tests bar

running 1 test
test src/lib.rs - foo (line 1) ... FAILED

failures:

---- src/lib.rs - foo (line 1) stdout ----
error[E0423]: expected value, found crate `bar`
 --> src/lib.rs:3:1
  |
5 | bar
  | ^^^ not a value

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0423`.
Couldn't compile the test.

failures:
    src/lib.rs - foo (line 1)

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s

error: doctest failed, to rerun pass `--doc`
```

This portion above is always color and bold formatted:

```bash
error[E0423]: expected value, found crate `bar`
 --> src/lib.rs:3:1
  |
5 | bar
  | ^^^ not a value

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0423`.
```

In contrast,

```
[zack@laptop src]$ cargo t --color never --tests -- --color never
   Compiling bar v0.1.0 (/home/zack/projects/bar)
error[E0425]: cannot find value `bar` in this scopear(test)
 --> src/lib.rs:9:9
  |
9 |         bar
  |         ^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
error: could not compile `bar` (lib test) due to 1 previous error
```

has no color or formatting. Also notice the different error code, E0425, compared to E0423.

With the exception of the two instance of `FAILED` as red, the following has no color or formatting:

```
[zack@laptop src]$ rustdoc --test --color never lib.rs

running 1 test
test lib.rs - foo (line 1) ... FAILED

failures:

---- lib.rs - foo (line 1) stdout ----
error[E0425]: cannot find value `bar` in this scope
 --> lib.rs:2:1
  |
3 | bar
  | ^^^ not found in this scope

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0425`.
Couldn't compile the test.

failures:
    lib.rs - foo (line 1)

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
```

Also notice that it has the same error code, E0425, as  `cargo t --tests` even though I would have guessed for it to be the same as `cargo t --doc`.

Unsurprisingly when trying to retain color via `--color always` when piping `stdout` into a different process, the color is not retained presumably for the same reason that `--color never` is not being respected when writing to a TTY.

I _think_ this might be related to the fact that `rustdoc` writes the compilation error to `stdout`; thus `cargo t --doc` also writes the formatted output to `stdout`; however `cargo t --tests` writes the compilation error to `stderr`. Normally both `stderr` and `stdout` are not formatted though, so not sure why that would matter.

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
cargo 1.80.1 (376290515 2024-07-16)
release: 1.80.1
commit-date: 2024-07-16
os: Arch Linux [64-bit]
```

ShashiSugrim:

Explain the motivation behind this change.

This commit fixes the issue where cargo's test command doesn't respect the color parameter when it gets passed in like this command: `cargo t --color never --doc -- --color never`

Fixes #14403
Test on a basic rust project, with a file called lib.rs that looks like this 
```
/// ```
/// bar
/// ```
pub fn foo() {}
#[cfg(test)]
mod tests {
    #[test]
    fn foo() {
        bar
    }
}
```
You can try to replicate the same commands @zacknewman used in the description of this issue. 
`cargo t --color never --doc -- --color never`

You will see that compared to the official build of cargo, this build will respect the --color argument

### DIFF
--- a/src/cargo/ops/cargo_test.rs
+++ b/src/cargo/ops/cargo_test.rs
@@ -1,5 +1,6 @@
 use crate::core::compiler::{Compilation, CompileKind, Doctest, Metadata, Unit, UnitOutput};
 use crate::core::profiles::PanicStrategy;
+use crate::core::shell::ColorChoice;
 use crate::core::shell::Verbosity;
 use crate::core::{TargetKind, Workspace};
 use crate::ops;
@@ -176,6 +177,7 @@ fn run_doc_tests(
     let gctx = ws.gctx();
     let mut errors = Vec::new();
     let doctest_xcompile = gctx.cli_unstable().doctest_xcompile;
+    let color = gctx.shell().color_choice();
 
     for doctest_info in &compilation.to_doc_test {
         let Doctest {
@@ -215,6 +217,14 @@ fn run_doc_tests(
         for (var, value) in env {
             p.env(var, value);
         }
+
+        let color_arg = match color {
+            ColorChoice::Always => "always",
+            ColorChoice::Never => "never",
+            ColorChoice::CargoAuto => "auto",
+        };
+        p.arg("--color").arg(color_arg);
+
         p.arg("--crate-name").arg(&unit.target.crate_name());
         p.arg("--test");
 
--- a/tests/testsuite/build_script.rs
+++ b/tests/testsuite/build_script.rs
@@ -2871,7 +2871,7 @@ fn env_test() {
 [RUNNING] `[ROOT]/foo/target/debug/deps/foo-[HASH][EXE]`
 [RUNNING] `[ROOT]/foo/target/debug/deps/test-[HASH][EXE]`
 [DOCTEST] foo
-[RUNNING] `rustdoc --edition=2015 --crate-type lib --crate-name foo[..]`
+[RUNNING] `rustdoc --edition=2015 --crate-type lib --color auto --crate-name foo[..]`
 
 "#]])
         .with_stdout_data(str![[r#"
--- a/tests/testsuite/lto.rs
+++ b/tests/testsuite/lto.rs
@@ -619,7 +619,7 @@ fn cdylib_and_rlib() {
 [RUNNING] `[ROOT]/foo/target/release/deps/bar-[HASH][EXE]`
 [RUNNING] `[ROOT]/foo/target/release/deps/b-[HASH][EXE]`
 [DOCTEST] bar
-[RUNNING] `rustdoc --edition=2015 --crate-type cdylib --crate-type rlib --crate-name bar --test [..]-C lto [..]
+[RUNNING] `rustdoc --edition=2015 --crate-type cdylib --crate-type rlib --color auto --crate-name bar --test [..]-C lto [..]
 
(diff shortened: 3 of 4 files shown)


src/cargo/ops/cargo_test.rs
```diff
@@ -212,6 +214,13 @@ fn run_doc_tests(
         gctx.shell().status("Doc-tests", unit.target.name())?;
         let mut p = compilation.rustdoc_process(unit, *script_meta)?;
 
+        let color_arg = match color {
+            ColorChoice::Always => "always",
+            ColorChoice::Never => "never",
+            ColorChoice::CargoAuto => "auto",
+        };
+        p.arg("--color").arg(color_arg);
+
```
epage, line 223: Mind moving this later? Within this function, we had "add env, add args" and now we have "add arg, add env, add args"

Also, as indicated by the test suite, some tests will need to be updated that are checking for some but not all arguments

ShashiSugrim:
Sorry for the problems. I am trying to fix my branch commits

ShashiSugrim:
Hello, please review the latest changes, thank you

src/cargo/ops/cargo_test.rs
```diff
@@ -215,6 +217,14 @@ fn run_doc_tests(
         for (var, value) in env {
             p.env(var, value);
         }
+
+        let color_arg = match color {
+            ColorChoice::Always => "always",
+            ColorChoice::Never => "never",
+            ColorChoice::CargoAuto => "auto",
+        };
+        p.arg("--color").arg(color_arg);
```
weihanglo, line 226: Just want to call out that this may break others' builds if they already have `--color` in their `RUSTDOCFLAGS`. `rustdoc` doesn't allow more then one occurrence of the `--color` flag. While we [don't consider this a breaking change](https://doc.rust-lang.org/nightly/cargo/reference/config.html#buildrustdocflags), this is worth mentioning in changelog.

weihanglo (approved):

PR #14427, opened 2024-08-19
Implement path-bases (RFC 3529) 2/n: `cargo [add|remove|update]` support

dpaoliello:

This PR adds the `--base` option to `cargo add` to allow adding a path dependency with a path base.


src/bin/cargo/commands/add.rs
```diff
@@ -101,6 +101,12 @@ Example uses:
                 .help("Filesystem path to local crate to add")
                 .group("selected")
                 .conflicts_with("git"),
+            clap::Arg::new("base")
+                .long("base")
+                .action(ArgAction::Set)
+                .value_name("BASE")
+                .help("The path base to use when adding from a local crate.")
```
epage, line 108: This should have a `(unstable)` suffix

src/bin/cargo/commands/add.rs
```diff
@@ -224,6 +230,7 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
 
 fn parse_dependencies(gctx: &GlobalContext, matches: &ArgMatches) -> CargoResult<Vec<DepOp>> {
     let path = matches.get_one::<String>("path");
+    let base = matches.get_one::<String>("base");
```
epage, line 233: I'm not seeing an unstable gate on this flag

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -843,7 +863,11 @@ impl PathSource {
 
 impl std::fmt::Display for PathSource {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        self.path.display().fmt(f)
+        if let Some((base_name, _)) = &self.base_name_and_value {
+            write!(f, "[{base_name}]/{}", self.path.display())
```
epage, line 867: `[base_name]/foo/bar` reads in an odd way to me

What do you think of `base_name:/foo/bar`?

src/cargo/util/toml/mod.rs
```diff
@@ -2225,10 +2226,12 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
 pub(crate) fn lookup_path_base<'a>(
     base: &PathBaseName,
     gctx: &GlobalContext,
-    workspace_root: &dyn Fn() -> CargoResult<&'a PathBuf>,
-    features: &Features,
+    workspace_root: &dyn Fn() -> CargoResult<&'a Path>,
```
epage, line 2229: Could you split this change out into its own commit?

src/cargo/util/toml/mod.rs
```diff
@@ -2225,10 +2226,12 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
...
+    features: Option<&Features>,
```
epage, line 2230: This feels sketchy to make optional.  We still need people to do the check somewhere

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -349,10 +369,15 @@ fn resolve_dependency(
...
-            let mut source = crate::sources::PathSource::new(&path, src.source_id()?, gctx);
+            let mut source = crate::sources::PathSource::new(&src.path, src.source_id()?, gctx);
             let package = source.root_package()?;
             Dependency::from(package.summary())
         };
+        if let Some(selected_source) = selected.source.as_mut() {
+            if let Source::Path(selected_source) = selected_source {
+                selected_source.base_name_and_value = base_name_and_value;
+            }
+        }
```
epage, line 380: I feel like this should be part of the creation of `src`.  Thst leaves `Dependency::from(package.summary())` but that should then be handled within that branch

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -331,10 +336,25 @@ fn resolve_dependency(
...
+        let (path, base_name_and_value) = if let Some(base_name) = &arg.base {
+            let workspace_root = || Ok(ws.root_manifest().parent().unwrap());
+            let base_value = lookup_path_base(
+                &PathBaseName::new(base_name.clone())?,
+                gctx,
+                &workspace_root,
+                None,
+            )?;
+            (
+                base_value.join(raw_path),
```
epage, line 348: This assumes users will be doing
```console
$ cargo add --base foo --path bar
```
to get
```toml
bar = { base = "foo", path = "bar" }
```

Would users instead be expecting
```console
$ cargo add --base foo --path ../../../foo/bar
```
to get
```toml
bar = { base = "foo", path = "bar" }
```
or maybe even auto-base detection?

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -412,10 +412,18 @@ impl Dependency {
...
                         }
+                        let relative_to = if let Some((base_name, base_value)) =
+                            src.base_name_and_value.as_ref()
+                        {
+                            table.insert("base", base_name.into());
+                            base_value
+                        } else {
+                            crate_root
+                        };
+                        let relpath = path_field(relative_to, &src.path);
```
epage, line 426: imo this code is misleading, both returning a value and having a side effect.  I'd rather just have the two inserts inside of it I think

src/cargo/util/toml_mut/dependency.rs
```diff
@@ -493,12 +501,20 @@ impl Dependency {
...
+                    let relative_to =
+                        if let Some((base_name, base_value)) = src.base_name_and_value.as_ref() {
+                            overwrite_value(table, "base", base_name);
+                            base_value
+                        } else {
+                            table.remove("base");
+                            crate_root
+                        };
+                    let relpath = path_field(relative_to, &src.path);
                     overwrite_value(table, "path", relpath);
```
epage, line 518: ditto

src/doc/man/generated_txt/cargo-add.txt
```diff
@@ -5,7 +5,7 @@ NAME
 
 SYNOPSIS
        cargo add [options] crate…
-       cargo add [options] --path path
+       cargo add [options] --path path [--base base]
```
epage, line 8: `--base` seems uncommon enough that I'm unsure if we should have it in the usage

src/doc/man/generated_txt/cargo-add.txt
```diff
@@ -56,6 +56,11 @@ OPTIONS
            <https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies>
            to local crate to add.
 
+       --base base
+           The path base
+           <https://doc.rust-lang.org/cargo/reference/unstable.html#path-bases>
+           to use when adding a local crate.
```
epage, line 62: We should call out that this is unstable

tests/testsuite/cargo_add/overwrite_git_with_path_base/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+            .publish();
+    }
+
+    let project = Project::from_template(current_dir!().join("in"));
+    let project_root = project.root();
+    let cwd = project_root.join("primary");
+
+    snapbox::cmd::Command::cargo_ui()
+        .arg("add")
+        .arg_line("cargo-list-test-fixture-dependency --path dependency --base my_base")
```
epage, line 30: > tests/testsuite/cargo_add/overwrite_git_with_path_base/mod.rs

I don't think this test case is needed.  The important part is that any `--path` causes `git` fields to be wiped out.

Testing this would be like testing that `git` without any other fields and with `["rev", "tag", "branch"]` all successfully replace a `path`

tests/testsuite/cargo_add/path_base/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+        "0.2.0+my-package",
+        "0.2.3+my-package",
+        "0.4.1+my-package",
+        "20.0.0+my-package",
+        "99999.0.0+my-package",
+        "99999.0.0-alpha.1+my-package",
+    ] {
+        cargo_test_support::registry::Package::new("cargo-list-test-fixture-dependency", ver)
+            .publish();
+    }
```
epage, line 22: We don't need these packages published, right?

src/cargo/util/toml_mut/dependency.rs
dpaoliello: Might be confusing for folks on Windows.

We could use a pipe, but it might be hard to differentiate between that and a slash: `base_name|foo/bar`

I was using brackets to mimic how the argument help text uses brackets to represent the name of a thing to be inserted/set rather than a literal argument. Could use braces to feel more Rust-format-string like instead?

src/cargo/util/toml/mod.rs
dpaoliello: Is there a good way to get `Features` from within `cargo add`? I didn't see any other use of it, and trying to construct one manually turned into a bit of a mess.

src/bin/cargo/commands/add.rs
dpaoliello: How do I do that check?

We changed `path-bases` to a Cargo feature, so there is no `-Z` flag. I also see that the `--public` flag is unstable, but has no checks for `-Z` flags or Cargo features.

epage: Well, thats a bug that `public` doesn't check...

In your case, the options are:
- Gate it on `-Zunstable-options`
- In `resolve_dependency`, you have a `Package` so you can get a `Manifest` and check `unstable_features` for this

src/cargo/util/toml_mut/dependency.rs
epage: > I was using brackets to mimic how the argument help text uses brackets to represent the name of a thing to be inserted/set rather than a literal argument. 

In CLI usage `[]` means "optional".  What you are more looking for is `<>` for "required".

That would make it `<base_name>/foo/bar`

> Could use braces to feel more Rust-format-string like instead?

So that would be `{base_name}/foo/bar`.  That aligns well with a prior version of this RFC, existing features in Cargo, and additional features we are looking at adding to cargo.  I lean towards this.

src/bin/cargo/commands/add.rs
dpaoliello: Ah, I didn't realize that `spec` was the `Package` for the current package. Thanks!

src/cargo/ops/cargo_add/mod.rs
dpaoliello: Unfortunately, this doesn't work: `select_package` creates a new `Dependency` object without re-using the original `Source` information, so the `base` information is lost.

src/cargo/ops/cargo_add/mod.rs
dpaoliello: I like the idea of `--path` always being relative the working directory, it also means that absolute paths will work as well.

I'm not a fan of auto-detection: I'd rather that the user be explicit.

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -331,10 +337,41 @@ fn resolve_dependency(
...
+                    .as_ref()
+                    .unwrap_or(&empty);
+                let features = Features::new(cargo_features, gctx, &mut warnings, true)?;
+                if !warnings.is_empty() {
+                    for warning in warnings {
+                        gctx.shell().warn(warning)?;
+                    }
+                }
+                features
+            };
```
epage, line 360: FYI `spec.manifest().unstable_features()`

src/bin/cargo/commands/add.rs
epage: Looks like this was resolved by delegating to `lookup_path_base`

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -331,10 +337,41 @@ fn resolve_dependency(
...
+            let base_value = lookup_path_base(
+                &PathBaseName::new(base_name.clone())?,
+                gctx,
+                &workspace_root,
+                &features,
+            )?;
+            Some((base_name.clone(), base_value))
+        } else {
+            None
+        };
```
epage, line 370: If this has switched to `--path` always being local paths (rather than relative to base), could we only store `base` in `PathSource` instead of also a `base_value` and do all of that processing there?

We already have the processing for package-relative paths and this is all very similar.  In fact, each of these reviews is increasing my desire for us to add ``base = "package"`.  `--base package` could then serve as a way to remove a `base`.

src/cargo/ops/cargo_add/mod.rs
epage: two-phase initialization is not great and can lead to its own bugs, in addition to being messy.

I think I'd prefer us to handle it in `select_package`

dpaoliello: Ok, I pushed this down into the `Summary` to `Dependency` conversion.

src/cargo/ops/cargo_add/mod.rs
dpaoliello: > do all of that processing there?

I'm not sure what you mean by "there" - I tried moving it into `to_toml` and `update_toml` but we're missing all of the workspace and config information. I could move it into `PathSource::new_with_base`, but I don't think that's really much neater.

src/cargo/util/toml_mut/dependency.rs
epage, line 1: Two cases this doesn't support atm:
- `cargo add --path foo --base bar && cargo add foo --dev`
- ... add a workspace dep with a base ... `cargo add workspace_dep`

In these cases, we call `Dependency::from_toml` but it doesn't look like thats been updated

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -733,7 +749,12 @@ fn select_package(
...
                 }
                 1 => {
-                    let mut dep = Dependency::from(&possibilities[0]);
+                    let base_name_and_value = dependency
+                        .source()
+                        .map(|s| s.as_path().map(|p| &p.base_name_and_value))
+                        .flatten()
+                        .unwrap_or(&None);
+                    let mut dep = Dependency::from_summary(&possibilities[0], base_name_and_value);
                     if let Some(reg_name) = dependency.registry.as_deref() {
```
epage, line 758: Like we d with `registry`, I think I'd rather handle it here than in `from_summary`.  `from_summary` is constructing it according to the `Summary`.  We then manually manipulate it for our needs.  Trying to mix the two is mixing of concerns and inconsistently doing so since we have the registry code here.

src/cargo/ops/cargo_add/mod.rs
epage: You are right that passing more parameters along with `crate_root` would be janky.

We have two different means of dealing with relatiev paths.  Bases are handled here and package-relative paths are handled in `to_toml`.  We also have the awkwardness of storing two paths, each used in a different scenario.  I'm trying to find a way to fix both problems.


Let's step back and look at what we are trying to all have work.

We keep paths as absolute because
- Its easier when converting from a `Summary`
- It helps when moving a `Dependency` from a workspace to a package manifest

We could
- Make the path relative only when `base` is present.  This solves part of the problem but not the whole.
- Pass all of those extra parameters to `to_toml`
- Always track a relative path in `PathSource`, requiring us to specially manage `Dependency::from_summary`

Thoughts?

src/cargo/util/toml_mut/dependency.rs
dpaoliello: Fixed up `from_toml`, and added tests for the cases you called out.

src/cargo/ops/cargo_add/mod.rs
dpaoliello: I think I've found a reasonable approach.

The key was that the callers of `to_toml` and `from_toml` already have all of the information we need to lookup path bases, they just weren't passing it down to those functions.

The final design is that `PathSource` is always absolute, but now has a `base` field which contains the name of the path base. I'm doing the fixups to set the `base` in `resolve_dependency` and `select_package`, so `Dependency::from(Summary)` stays exactly the same as it currently is.

PR #14428, opened 2024-08-19
[mdman] Normalize newlines when rendering options

dpaoliello:
When using `cargo build-man` on Windows, I noticed that some files I didn't modify were getting modified, specifically they had an extra space added in the middle of a code block in an option description.

The root cause appears to be that if a code block in a handlebars template was split across multiple lines, then both the `\r` and the `\n` were being converted to spaces.

The fix for this is to normalize newlines to `\n` after templates are expanded by handlebars but before the individual formatters render the block as an option.

### DIFF
--- a/crates/mdman/src/hbs.rs
+++ b/crates/mdman/src/hbs.rs
@@ -139,6 +139,9 @@ impl HelperDef for OptionHelper<'_> {
         // Render the block.
         let block = t.renders(r, gctx, rc)?;
 
+        // Windows newlines can break some rendering, so normalize.
+        let block = block.replace("\r\n", "\n");
+
         // Get the name of this page.
         let man_name = gctx
             .data()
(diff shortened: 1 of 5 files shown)


crates/mdman/src/hbs.rs
```diff
@@ -139,6 +139,9 @@ impl HelperDef for OptionHelper<'_> {
         // Render the block.
         let block = t.renders(r, gctx, rc)?;
 
+        // Windows newlines can break some rendering, so normalize.
+        let block = block.replace("\r\n", "\n");
```
epage, line 143: Could you split this into two commits, test addition and fix, so the fix commit shows the behavior change through how the test changes?

dpaoliello: Done

PR #14432, opened 2024-08-21
fix: -Cmetadata includes whether extra rustflags is same as host

### LINKED ISSUE #14253: warning: output filename collision.
### Problem

Got lots of warnings like below
```
warning: output filename collision.
The lib target `cfg_if` in package `cfg-if v1.0.0` has the same output filename as the lib target `cfg_if` in package `cfg-if v1.0.0`.
Colliding filename is: target/debug/deps/libcfg_if-b07c5e93d2d15f7e.rlib
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
```

### Steps

no warnings:
- stable
- cargo 1.81.0-nightly (a515d4634 2024-07-02)

warnings:
- cargo 1.81.0-nightly (154fdac39 2024-07-07) start to pop up warnings.

### Possible Solution(s)

_No response_

### Version

```text
nightly-x86_64-apple-darwin
```

weihanglo:

Fixes #14253

The root cause is described in <https://github.com/rust-lang/cargo/issues/14253#issuecomment-2299595480>:

> `-Ztarget-applies-to-host --config target-applies-to-host=false` make it possible to build two units for the same output directory with different rustflags (one with artifact/target rustflags, one with none/host rust flags). Those flags aren't included in the `-Cmetadata` hash which is used to disambiguate different versions of crates, resulting in a conflict. The actual error being produced appears to be the result of two invocations of `rustc` racing and clobbering each other…

While we don't hash RUSTFLAGS because it may contain absolute paths that
hurts reproducibility, we track whether a unit's RUSTFLAGS is from host
config, so that we can generate a different metadata hash for runtime
and compile-time units.

This hack is only enabled when `target-applies-to-host=false` and when compile kind is the host platform, so the shouldn't affect any stable behavior.

### Additional information

cc @gmorenz

I found time working on this PR before my travel. Feel free to cherry pick and create a new PR if this is not correct.

### DIFF
--- a/src/cargo/core/compiler/build_runner/compilation_files.rs
+++ b/src/cargo/core/compiler/build_runner/compilation_files.rs
@@ -649,6 +649,26 @@ fn compute_metadata(
     // with user dependencies.
     unit.is_std.hash(&mut hasher);
 
+    // While we don't hash RUSTFLAGS because it may contain absolute paths that
+    // hurts reproducibility, we track whether a unit's RUSTFLAGS is from host
+    // config, so that we can generate a different metadata hash for runtime
+    // and compile-time units.
+    //
+    // HACK: This is a temporary hack for fixing rust-lang/cargo#14253
+    // Need to find a long-term solution to replace this fragile workaround.
+    // See https://github.com/rust-lang/cargo/pull/14432#discussion_r1725065350
+    if unit.kind.is_host() && !bcx.gctx.target_applies_to_host().unwrap_or_default() {
+        let host_info = bcx.target_data.info(CompileKind::Host);
+        let target_configs_are_different = unit.rustflags != host_info.rustflags
+            || unit.rustdocflags != host_info.rustdocflags
+            || bcx
+                .target_data
+                .target_config(CompileKind::Host)
+                .links_overrides
+                != unit.links_overrides;
+        target_configs_are_different.hash(&mut hasher);
+    }
+
     MetaInfo {
         meta_hash: Metadata(hasher.finish()),
         use_extra_filename: should_use_metadata(bcx, unit),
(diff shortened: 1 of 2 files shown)


tests/testsuite/rustflags.rs
```diff
@@ -1642,3 +1642,67 @@ fn target_applies_to_host_rustdocflags_works() {
...
+            rustflags = ['--cfg', 'from_target']
+            ",
+        )
+        .build();
+
+    p.cargo("build -v")
+        .masquerade_as_nightly_cargo(&["target-applies-to-host"])
+        .arg("-Ztarget-applies-to-host")
+        .arg("-Zhost-config")
+        // Sometimes it compiles. Not deterministic...
```
gmorenz, line 1688:
```suggestion
```

Nit: This comment should be removed (at least in the fixed version).

gmorenz (approved):
lgtm (up to nit) :+1:

src/cargo/core/compiler/build_runner/compilation_files.rs
```diff
@@ -649,6 +649,23 @@ fn compute_metadata(
...
+    // This is essentially a hack for fixing rust-lang/cargo#14253.
+    if unit.kind.is_host() && !bcx.gctx.target_applies_to_host().unwrap_or_default() {
+        let host_info = bcx.target_data.info(CompileKind::Host);
+        let target_configs_are_different = unit.rustflags != host_info.rustflags
+            || unit.rustdocflags != host_info.rustdocflags
+            || bcx
+                .target_data
+                .target_config(CompileKind::Host)
+                .links_overrides
+                != unit.links_overrides;
```
epage, line 665: This feels a bit brittle. If we don't remember to update this with anything else that is different but not in metadata already, we'll see this problem again.

The only thoughts I had are
- Putting this question in the unit or host info to answer
- Hashing `unit.kind.is_host() && !bcx.gctx.target_applies_to_host().unwrap_or_default()`

Note: answering this is more important for long term reliability and stabilization.  This doesn't need to block fixing people who are using the unstable feature

weihanglo:
Removed the comment and added a note above the hack.

Thanks people. Going to merge this.

@bors r=epage

PR #10677, opened 2022-05-18
fix(publish): add more check when use `publish -p <SPEC>` 

likzn:
### Main issue
As issue say #10536 , we need add more check when user use `cargo publish -p <SPEC>`

>@ehuss point outs: 
>From a behavior standpoint, here are some things to check:
> - In the root of a virtual workspace, it should be an error to run without -p.
>- It should be an error to pass -p for a non-workspace member.
>- It should be an error for -p to match multiple packages.
>- When using -p, it should publish that package, not the one in the current directory (which can be different).

PR #14433, opened 2024-08-21
Publish workspace

jneem:
Adds support for simultaneously publishing multiple (possibly inter-dependent) packages in a workspace, gated by the `-Zpackage-workspace` flag.

We had a couple of questions:
- Is it worth adding a new `-Zpublish-workspace` flag, or is it ok to reuse the old one?
- This breaks the `publish_namespaced` test, because we now try to parse all packages as package specs, and "foo::bar" fails to parse. This could probably be avoided by a refactor (adding a different package function that takes a list of packages instead of trying to get them from cli arguments), but I also think it's a bug in the package spec parser. Thoughts?


ehuss:
> Is it worth adding a new `-Zpublish-workspace` flag, or is it ok to reuse the old one?

I don't have enough context to know if `-Zpackage-workspace` was intended as a stepping-stone for publishing, or if there really is an independent use-case for packaging versus publishing.

> This breaks the `publish_namespaced` test, because we now try to parse all packages as package specs, and "foo::bar" fails to parse. This could probably be avoided by a refactor (adding a different package function that takes a list of packages instead of trying to get them from cli arguments), but I also think it's a bug in the package spec parser. Thoughts?

Can the [check here](https://github.com/rust-lang/cargo/blob/ef854d2f66df7bfcd803a6ca1cfa3b619603142c/src/cargo/util/toml/mod.rs#L2707-L2714) be moved to an earlier stage, like in `ops::registry::publish::publish` by iterating over the `just_pkgs` list and checking if any of them have the `open-namespaces` feature?

ehuss:
I realistically won't have time to give this a good review for at least a few days if not a week or more. I was wondering if @weihanglo or @epage would have more bandwidth in the short term? If not, that's ok, I'll try to review this when I can.

epage:
> This breaks the publish_namespaced test, because we now try to parse all packages as package specs, and "foo::bar" fails to parse. This could probably be avoided by a refactor (adding a different package function that takes a list of packages instead of trying to get them from cli arguments), but I also think it's a bug in the package spec parser. Thoughts?

@jneem I posted #14467

epage:
Could you rebase on top of #14467?

jneem:
Rebased, thanks!

src/cargo/ops/cargo_package.rs
```diff
@@ -187,7 +187,11 @@ pub fn package(ws: &Workspace<'_>, opts: &PackageOpts<'_>) -> CargoResult<Vec<Fi
             spec.query(member_ids)?;
         }
     }
-    let pkgs = ws.members_with_features(specs, &opts.cli_features)?;
+    let mut pkgs = ws.members_with_features(specs, &opts.cli_features)?;
+
+    // In `members_with_features_old`, it will add "current" package (determined by the cwd)
+    // So we need filter
+    pkgs.retain(|(pkg, _feats)| specs.iter().any(|spec| spec.matches(pkg.package_id())));
```
epage, line 193: This change mirrors similar code in `cargo publish` that was added in #10677.

epage: Could you add that as a note to the commit?

btw I think this and the prior commit could be their own PR?

epage:
> > Is it worth adding a new -Zpublish-workspace flag, or is it ok to reuse the old one?
>
> I don't have enough context to know if -Zpackage-workspace was intended as a stepping-stone for publishing, or if there really is an independent use-case for packaging versus publishing.

`-Zpackage-workspace` is a stepping stone.  In that spirit, it seems like the feature would be named `-Zpublish-workspace` but probably not worth changing at this point.

The questions relevant for this is would we want to stabilize these together?  Are there design risks for one that by separating the stabilization of them would make it easier to stabilize the other?

tests/testsuite/publish.rs
epage, line 1: We should have a test case for the publish equivalent of #14356.

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
+            validate_registry(&just_pkgs, reg.as_ref())?;
+            if let Some(RegistryOrIndex::Registry(ref registry)) = &reg {
+                if registry != CRATES_IO_REGISTRY {
+                    // Don't warn for crates.io.
+                    opts.gctx.shell().note(&format!(
+                        "found `{}` as only allowed registry. Publishing to it automatically.",
+                        registry
+                    ))?;
+                }
             }
```
epage, line 109: nit: looks like switching to `infer_registry` could be broken out into a refactor commit?  Within the existing single-package publish, it should have no behavior change but its a stepping stone towards multi-package publish.

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
-    assert_eq!(pkgs.len(), 1);
 
-    let (pkg, cli_features) = pkgs.pop().unwrap();
-
-    let mut publish_registry = match opts.reg_or_index.as_ref() {
-        Some(RegistryOrIndex::Registry(registry)) => Some(registry.clone()),
-        _ => None,
-    };
-    if let Some(ref allowed_registries) = *pkg.publish() {
-        if publish_registry.is_none() && allowed_registries.len() == 1 {
```
epage, line 87: nit: a potential refactor that could be broken out is to switch this code into the style of `match` in the new code.

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
-        val => val,
-    };
     let source_ids = super::get_source_id(opts.gctx, reg_or_index.as_ref())?;
     let mut registry = super::registry(
         opts.gctx,
         &source_ids,
         opts.token.as_ref().map(Secret::as_deref),
         reg_or_index.as_ref(),
         true,
-        Some(operation).filter(|_| !opts.dry_run),
```
epage, line 138: nit: removing `operation` variable could be its own refactor commit

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
     )?;
-    verify_dependencies(pkg, &registry, source_ids.original)?;
 
-    // Prepare a tarball, with a non-suppressible warning if metadata
-    // is missing since this is being put online.
-    let tarball = ops::package_one(
+    // Validate all the packages before publishing any of them.
+    for (pkg, _) in &pkgs {
+        verify_dependencies(pkg, &registry, source_ids.original)?;
+    }
```
epage, line 129: note for people reviewing this: at this time, `verify_dependencies` does not contact the registry, so the local, unpublished dependencies shouldn't be a problem

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
+
+    let specs = Packages::from_flags(
+        false,
+        vec![],
+        just_pkgs.iter().map(|p| p.name().to_string()).collect(),
```
epage, line 134: package names are not always sufficient specs.  I don't know what paths this code goes down but if it hits the wrong one now, or in the future, this will break in less commonly seen but important cases.

For example, `clap` has transitive dependencies on its self so I can't run `cargo check -p clap`, I have to do `cargo check -p clap@...`

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
+    )?;
+    let pkg_dep_graph = ops::cargo_package::package_with_dep_graph(
```
epage, line 136: In general, it feels odd for us to have process the specs and then convert them back into specs and process them again.

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
+            )?;
+        }
+
+        for pkg_id in &ready {
+            let (pkg, (_features, tarball)) = &pkg_dep_graph.packages[pkg_id];
+
+            if !opts.dry_run {
+                let ver = pkg.version().to_string();
+
+                tarball.file().seek(SeekFrom::Start(0))?;
```
epage, line 171: While I usually try to minimize noise, we should print as we are uploading because its important context for when an upload fails.

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
+            )?;
+
+            if !opts.dry_run {
+                // Short does not include the registry name.
+                let short_pkg_description = format!("{} v{}", pkg.name(), pkg.version());
+                let source_description = source_ids.original.to_string();
+                ws.gctx().shell().status(
+                    "Uploaded",
+                    format!("{short_pkg_description} to {source_description}"),
+                )?;
```
epage, line 207: At the time, the uploaded message was important context for what phase we are at with waiting.  I wonder if its just extra noise in this scenario.

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
-            wait_for_publish(opts.gctx, source_ids.original, pkg, timeout)?;
+
+        for id in &finished {
+            to_confirm.remove(id);
+            remaining.remove(id);
+        }
+        if finished.is_empty() {
+            // If nothing finished, it means we timed out while waiting for confirmation.
+            // We're going to exit, but first we need to check: if we pretend that the unconfirmed
+            // uploads are confirmed, did we finish everything?
```
epage, line 236: What does this mean?

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
-        source_ids.original,
-        opts.dry_run,
-    )?;
-    if !opts.dry_run {
-        const DEFAULT_TIMEOUT: u64 = 60;
-        let timeout = if opts.gctx.cli_unstable().publish_timeout {
-            let timeout: Option<u64> = opts.gctx.get("publish.timeout")?;
-            timeout.unwrap_or(DEFAULT_TIMEOUT)
+    while !to_confirm.is_empty() {
+        if !ready.is_empty() {
```
epage, line 158: Why do we continue the loop when `ready` is empty?

src/cargo/ops/registry/publish.rs
```diff
@@ -210,34 +258,28 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
 fn wait_for_publish(
```
epage, line 258: Oh, we are waiting until the first set of packages become available.

We should probably rename this function to clarify that.

src/cargo/ops/registry/publish.rs
```diff
@@ -247,43 +289,67 @@ fn wait_for_publish(
...
             "Published",
             format!("{short_pkg_description} at {source_description}"),
         )?;
     }
 
-    Ok(())
+    Ok(available)
+}
+
+fn poll_one_package(
```
epage, line 336: nit: could pulling this out be its own refactor commit?

src/cargo/ops/registry/publish.rs
```diff
@@ -498,3 +564,152 @@ fn transmit(
...
+    // The reverse-dependency graph. It has an edge p -> q if package q depends on package p.
+    rev_graph: Graph<PackageId, ()>,
+    // The weight of a package is the number of unpublished dependencies it has.
+    weights: HashMap<PackageId, usize>,
+}
+
+impl PublishOrder {
+    // Given a package dependency graph, creates a `PublishOrder` for tracking state and also
+    // a list of packages that have no dependencies (i.e. can be published immediately).
+    fn new(graph: &Graph<PackageId, ()>) -> (Self, Vec<PackageId>) {
```
epage, line 579: API wise this feels weird that we return `ready` as well as `Self`.

Options
- We keep it.  Its all internal
- We add a `take_ready` which removes items with `weight == 0`
- likely other approaches as well

src/cargo/ops/registry/publish.rs
```diff
@@ -498,3 +564,152 @@ fn transmit(
...
+        [] => String::new(),
+        [a] => a.clone(),
+        [a, b] => format!("{a} {final_sep} {b}"),
+        [names @ .., last] => {
+            format!("{}, {final_sep} {last}", names.join(", "))
+        }
+    }
+}
+
+fn validate_registry(pkgs: &[&Package], reg_or_index: Option<&RegistryOrIndex>) -> CargoResult<()> {
```
epage, line 630: nit: could extracting this be put in its own refactor commit?

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
+                let timeout: Option<u64> = opts.gctx.get("publish.timeout")?;
+                timeout.unwrap_or(DEFAULT_TIMEOUT)
+            } else {
+                DEFAULT_TIMEOUT
+            };
+            if 0 < timeout {
+                let timeout = Duration::from_secs(timeout);
+                wait_for_publish(opts.gctx, source_ids.original, &to_confirm, timeout)?
+            } else {
+                Vec::new()
```
epage, line 225: The timeout behavior is not intuitive and should at least be called out in the PR description and, once merged, in the tracking issue so we make sure we evaluate this when determining whether its ready for stabilization

tests/testsuite/publish.rs
```diff
@@ -961,7 +961,6 @@ fn publish_failed_with_index_and_only_allowed_registry() {
         .arg(registry.index_url().as_str())
         .with_status(101)
         .with_stderr_data(str![[r#"
-[NOTE] found `alternative` as only allowed registry. Publishing to it automatically.
```
epage, line 964: Why was the note removed?

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
-            &opts.gctx,
-            &source_ids.original,
-            None,
-            operation,
-            vec![],
-            false,
-        )?));
-    }
+    let (mut order, mut ready) = PublishOrder::new(&pkg_dep_graph.graph);
+    let mut to_confirm: HashSet<_> = ready.iter().copied().collect();
```
epage, line 154: For readability, it feels weird that `to_confirm` has things that are `ready` but not published.

src/cargo/ops/registry/publish.rs
```diff
@@ -57,151 +64,192 @@ pub struct PublishOpts<'gctx> {
...
+            const DEFAULT_TIMEOUT: u64 = 60;
+            let timeout = if opts.gctx.cli_unstable().publish_timeout {
+                let timeout: Option<u64> = opts.gctx.get("publish.timeout")?;
+                timeout.unwrap_or(DEFAULT_TIMEOUT)
+            } else {
+                DEFAULT_TIMEOUT
+            };
+            if 0 < timeout {
+                let timeout = Duration::from_secs(timeout);
+                wait_for_publish(opts.gctx, source_ids.original, &to_confirm, timeout)?
```
epage, line 223: Publish and waiting order is non-deterministic.  `ready` is created from a `HashMap` and `to_confirm` is a `HashSet`.

Internally, this makes testing a mess (likely the tests will break if `.unordered()` isn't used).

For users, its generally nice to put things in a deterministic order.

For `wait_for_publish`, we should probably wait in the order of upload to reduce the risk of hitting the timeout unnecessarily.

tests/testsuite/publish.rs
```diff
@@ -3758,7 +3846,23 @@ You may press ctrl-c to skip waiting; the crate should be available shortly.
...
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+[VERIFYING] b v0.0.1 ([ROOT]/foo/b)
+[UPDATING] crates.io index
+[ERROR] failed to verify package tarball
+
+Caused by:
+  failed to get `a` as a dependency of package `b v0.0.1 ([ROOT]/foo/target/package/b-0.0.1)`
+
+Caused by:
+  found a package in the remote registry and the local overlay: a@0.0.1
```
epage, line 3865: I'm assuming #14448 will improve this.  its a race for who is merged first and causes the other a conflict :)

src/cargo/ops/registry/publish.rs
epage: Looking over how this is done, maybe it does make sense.  Unsure

tests/testsuite/publish.rs
jneem: `infer_registry` doesn't print this warning if `--index` or `--registry` was provided. The warning is misleading: we didn't automatically choose where to publish.

src/cargo/ops/registry/publish.rs
jneem: Prior to multi-package publishing, it wasn't an error if the verification timed out: we just assume that it will appear in the registry eventually. We try to imitate the same behavior when publishing multiple packages, but we *do* raise an error if the package(s) that timed out had dependencies that we haven't uploaded yet.

To be precise, we exit with success if and only if "confirmed packages" union "uploaded but not confirmed packages" is everything.

src/cargo/ops/registry/publish.rs
jneem: Say we have packages `main`, `dep1`, and `dep2`, where `main` depends on both of the `dep`s. The first time through the loop, both of the deps are ready, so we upload them both. But now suppose that when we poll, we only see `dep1` as ready. Then we'll run through the loop again even though nothing is ready for upload.

Having said that, I'm going to remove this check in favor of printing the "Uploading..." messages one-by-one for more precise context.

src/cargo/ops/registry/publish.rs
jneem: I'll sort both `ready` and `to_confirm` so they'll be in the same order. In case it wasn't clear, btw, `wait_for_publish` polls all the packages on each run through the loop.

src/cargo/ops/registry/publish.rs
jneem: The API is definitely a little weird, but it has the advantage that it requires the least state-tracking: when you mark a package as finished, it's immediately obvious (in constant time, with no extra state) which packages are ready. To have a two-step API that separates out the "mark as done" and "get next ready" steps, we'd have to either add API for mutating `Graph` or (better, IMO) add an extra field to store packages that are ready to be taken in a `take_ready` function.

Given the extra complication, I prefer the "keep it" option...

tests/testsuite/publish.rs
jneem: I brought one of those tests over, but this PR's behavior is intentionally different from `cargo package`: we abort with an error as soon as any package in our list has `publish = false`. Maybe we could be more permissive, by filtering out unpublishable packages if `to_publish` is `Packages::Default`?

tests/testsuite/publish.rs
epage: Thanks for that fix!

src/cargo/ops/registry/publish.rs
epage: Thanks.  I think the way the code is structured can be a bit confusing.  I'll try to figure out what led me to be confused and how to improve it.

src/cargo/ops/registry/publish.rs
```diff
@@ -57,187 +65,224 @@ pub struct PublishOpts<'gctx> {
...
+            to_confirm.remove(id);
+            remaining.remove(id);
+        }
+        if finished.is_empty() {
+            // If nothing finished, it means we timed out while waiting for confirmation.
+            // We're going to exit, but first we need to check: if we pretend that the unconfirmed
+            // uploads are confirmed, did we finish everything?
+            for id in &ready {
+                remaining.remove(id);
+            }
```
epage, line 238: Would it work to do this?  I feel like it would make the intent more clear.
```suggestion
            for id in &to_confirm {
                remaining.remove(id);
            }
```

src/cargo/ops/registry/publish.rs
```diff
@@ -57,187 +65,224 @@ pub struct PublishOpts<'gctx> {
...
+                let short_pkg_description = format!("{} v{}", pkg.name(), pkg.version());
+                let source_description = source_ids.original.to_string();
+                ws.gctx().shell().status(
+                    "Uploaded",
+                    format!("{short_pkg_description} to {source_description}"),
+                )?;
+            }
+        }
+
+        let finished = if opts.dry_run {
```
epage, line 205: Since we use the term `to_confirm`, it seems like it would make it easier to follow if we made this match
```suggestion
        let confirmed = if opts.dry_run {
```

src/cargo/ops/registry/publish.rs
```diff
@@ -57,187 +65,224 @@ pub struct PublishOpts<'gctx> {
...
+            ready.clone()
```
epage, line 206: If we wait on `to_confirm`, why don't we return that here?
```suggestion
            to_confirm.clone()
```

src/cargo/ops/registry/publish.rs
```diff
@@ -57,187 +65,224 @@ pub struct PublishOpts<'gctx> {
...
+                remaining.remove(id);
+            }
+
+            if remaining.is_empty() {
+                // It's ok that we timed out, because nothing was waiting on dependencies to
+                // be confirmed.
+                break;
+            } else {
+                let failed_list = package_list(remaining, "and");
+                bail!("failed to publish {failed_list} because we timed out waiting for dependencies.");
```
epage, line 246: Feels weird to say something failed that we haven't attempted
```suggestion
                bail!("unable to publish {failed_list} because we timed out waiting for dependencies.");
```

tests/testsuite/publish.rs
epage: Whichever way we go, let's call it out in the PR description and in the tracking issue / stabilization report so we can make sure we make a conscious decision about this.

For myself, I think I lean towards `--workspace` implicitly excluding packages while `--package foo` should fail.  Unsure how cleanly we can implement those semantics.  However, erroring is the safer choice as we could probably change to implicit exclude later.

src/cargo/ops/registry/publish.rs
epage: Thanks!  This makes it really clear that that commit is actually diagnostic fix!

(technically, it could be pulled out into its own PR then... but not going to block on that)

src/cargo/ops/registry/publish.rs
epage: Do we need to mutate the graph or only `weights`?  If we did so, then wouldn't it still be trivial to track what packages are ready (its `self.weights.keys()`).

src/cargo/ops/registry/publish.rs
epage: > Say we have packages main, dep1, and dep2, where main depends on both of the deps. The first time through the loop, both of the deps are ready, so we upload them both. But now suppose that when we poll, we only see dep1 as ready. Then we'll run through the loop again even though nothing is ready for upload.

That we are doubling the timeout for some packages is non-obvious and this should probably have a comment in the code.

src/cargo/ops/registry/publish.rs
jneem: Sure, they should be the same in this case (except that one is a Vec and one is a BTreeSet; I've made them both BTreeSets)

src/cargo/ops/registry/publish.rs
jneem: Not only is the intent more clear, the `ready` one was incorrect in some cases. I tried adding a test to exercise this, but it relies on specific timing that the test harness doesn't offer (we'd have to upload two crates at once, but only see one as ready when we poll).

src/cargo/ops/registry/publish.rs
jneem: Oh sorry, for some reason I misunderstood that you were talking about the `new` function. Yes, this can have a nicer API.

jneem:
Ok, I think I've addressed all of the comments. I tweaked the main upload loop to be hopefully a bit clearer (we're checking `remaining.is_empty()` instead of `to_confirm.is_empty()` for termination, and as a result we get to populate `to_confirm` after the upload instead of before).

I've called out the behavioral questions (the behavior of unpublishable packages, and the behavior of timeouts) in the PR description. I don't *know* of any other design questions that would need an unstable flag. I was thinking that the main application of packaging workspaces is for publishing workspaces, so there isn't much value in stabilizing one before the other.

src/cargo/ops/registry/publish.rs
```diff
@@ -485,6 +586,74 @@ fn transmit(
...
+            .collect();
+        Self { rev_graph, weights }
+    }
+
+    // The set of packages that are ready for publishing (i.e. have no outstanding dependencies).
+    fn ready(&self) -> Vec<PackageId> {
+        let mut ready: Vec<_> = self
+            .weights
+            .iter()
+            .filter_map(|(id, weight)| (*weight == 0).then_some(*id))
```
epage, line 614: There a reason we do this rather than remove them in `mark_published`?

This is sort of a continuation of https://github.com/rust-lang/cargo/pull/14433#pullrequestreview-2273538538

epage:
> I've called out the behavioral questions (the behavior of unpublishable packages, and the behavior of timeouts) in the PR description. 

You had:

> What is the desired behavior for the publish timeout? This PR uploads the crates in batches (depending on the dependency graph), and it applies the timeout for each batch. So for example, if you have a dependency chain `a -> b -> c` then we'll wait up to 60 seconds for `a`, and then up to 60 seconds for `b`, and then up to 60 seconds for `c`.

But I don't think a straight dependency tree was the concern.  Is my update accurate?

> What is the desired behavior for the publish timeout? This PR uploads the crates in batches (depending on the dependency graph), and we only timeout if nothing in the batch is available within the timeout, deferring the rest to the next wait-for-publish. So for example, if you have packages `a`, `b`, `c` then we'll wait up to 60 seconds and if only `a` and `b` were ready in that time, we'll then wait another 60 seconds for `c`.

src/cargo/ops/registry/publish.rs
```diff
@@ -115,82 +123,165 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
...
+    //    available (the set    we poll for is `to_confirm`).
+    //
+    // Note that `to_confirm` might contains packages from previous rounds: we
+    // might upload a bunch of packages in step 1, observe in step 2 that one
+    // has become available, and then go back to step 1 and upload a bunch more.
+    // Then `to_confirm` will contain the new packages we just uploaded, plus
+    // most of the packages from the previous round.
+    let mut order = PublishOrder::new(&pkg_dep_graph.graph);
+    let mut ready = order.ready();
+    ready.sort();
```
epage, line 167: `ready` returns sorted data

epage: Could we just make that an invariant with a `BTreeSet`?

src/cargo/ops/registry/publish.rs
epage: The comment that was added to resolve this, while good and covering literally what I asked for, didn't fully address the concern I had. The comment was put deeper, inside of `if 0 < timeout {` but this doesn't help someone to understand why you can have `remaining` without `ready` where they are reading this non-obvious code.

src/cargo/ops/registry/publish.rs
epage: `wait_for_any_publish_confirmation`

src/cargo/ops/registry/publish.rs
```diff
@@ -115,82 +123,165 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
...
+                let short_pkg_description = format!("{} v{}", pkg.name(), pkg.version());
+                let source_description = source_ids.original.to_string();
+                ws.gctx().shell().status(
+                    "Uploaded",
+                    format!("{short_pkg_description} to {source_description}"),
+                )?;
+            }
+        }
+
+        to_confirm.extend(ready.iter().copied());
```
epage, line 219: When reading this, it implies to me that `ready` will continue to be used.

Some ideas
- `to_confirm.extend(ready.drain())` (depending on `ready`s type)
- `to_confirm.extend(std::mem::take(&mut ready)` 
- Inserting into `to_confirm` on the line right after `transmit`

... wait, does `ready` need to cross iterations of the loop?  We could loop over `remaining` and then immediately create `ready` before looping over it.  I feel like that would be a big help for understanding the code.

very minor nit: If we don't go with the last option, we *might* want to move `to_confirm` up a line at least.  Blank lines in code convey logical grouping.  Seeing a line like this suggests to me that we are initializing `to_confirm` for the following line but its for that and for future iterations of the loop.  Attaching it to the transmit operation (directly after it or at least after the block without a newline) conveys that what we transmit must be confirmed at some point.

src/cargo/ops/registry/publish.rs
```diff
@@ -115,82 +123,165 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
...
+                BTreeSet::new()
+            }
         };
-        if 0 < timeout {
-            let timeout = Duration::from_secs(timeout);
-            wait_for_publish(opts.gctx, source_ids.original, pkg, timeout)?;
+
+        for id in &confirmed {
+            to_confirm.remove(id);
+            remaining.remove(id);
```
epage, line 249: What if we removed from `remaining` when we transmit?  The loop will need to be updated to be `while !remaining.is_empty() || !to_confirm.is_empty()` which might negate the need for the comment I suggested.

Wait, do we need `remaining`?  If `PublishOrder` removes 0 weights, we can have a `any_ready` function that checks if weights is 0.  By removing `remaining`, we reduce the number of data structure we need to keep in sync and remove the question of "remaining to do what" (I keep assuming "to upload" while currently its" to do anything").

src/cargo/ops/registry/publish.rs
jneem: Graph nodes aren't removable (and with the current graph representation, making them removable has possible footguns: either we allow dangling edges to nodes that no longer exist, or we need to scan the whole graph on each node removal, making it easy to accidentally introduce quadratic behavior).

Rather than mess with the graph type, I've added some additional state to `PublishOrder` that hopefully makes things more clear.

src/cargo/ops/registry/publish.rs
jneem: I went with the last option.

I also pushed some of the state into `PublishOrder` so that the loop is hopefully clearer. It would be also be possible to remove some of that state altogether, but (AFAICS) only at the cost of making things quadratic in the number of packages. (For example, we could remove ready, and just recalculate it every time as the set of remaining packages with weight zero.) Maybe being quadratic in the number of packages isn't important? But on the other hand I think the state is reasonably well-encapsulated now...

jneem:
> But I don't think a straight dependency tree was the concern. Is my update accurate?

Yes, your update is accurate. I thought that a long straight dependency tree might also be a concern, because it effectively extends the timeout.

src/cargo/ops/registry/publish.rs
jneem: Ah, since I'd removed the `!ready.is_empty()` check it wasn't clear to me that the explanation was still needed. I've added one, at the point where we retrieve the ready packages.

epage:
> Yes, your update is accurate. I thought that a long straight dependency tree might also be a concern, because it effectively extends the timeout.

Not concerned about the straight line because the timeout is for publishing a package, not publishing everything, so its natural they get concatenated when doing them back-to-back.

src/cargo/ops/registry/publish.rs
jneem: Ah, thanks for the clarification. Yes, I think this is nicer and the names look good to me. My only concern is that I think this makes the whole process quadratic in the number of packages, since `take_ready` needs to scan over all the remaining packages. There's an intermediate solution that avoids the quadratic behavior by keeping the `ready` field (but still removing the `remaining` field). But also maybe workspaces are small enough to not worry about it...

epage: Large workspaces are likely a hundred members or so.  Publish performance isn't in a critical path as well.

jneem: Sounds good. I've pulled in your changes and squashed them in.

epage:
Thanks!

@bors r+

epage:
btw I've noted in #10948 the open questions

Issue #10159, opened 2021-12-04
Presentation of build script errors is noisy, difficult to understand


kornelski:
### Problem

When a custom build script fails, Cargo prints a very verbose output. It's often difficult to find the actual error message from the build script among all other debugging information printed just in case.

For example, it's common for pkg-config crate to fail, but the name of the library that user needs to install is buried under 30 other lines (and even worse with RUST_BACKTRACE=1)

```text
error: failed to run custom build command for `tst v0.1.0 (/private/tmp/tst)`

Caused by:
  process didn't exit successfully: `/private/tmp/tst/target/debug/build/tst-2429227eb6585f3e/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=IDONTEXIST_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_aarch64-apple-darwin
  cargo:rerun-if-env-changed=PKG_CONFIG_aarch64_apple_darwin
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=IDONTEXIST_STATIC
  cargo:rerun-if-env-changed=IDONTEXIST_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_aarch64-apple-darwin
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_aarch64_apple_darwin
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_aarch64-apple-darwin
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_aarch64_apple_darwin
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_LIBDIR
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_aarch64-apple-darwin
[...]

PR #14435, opened 2024-08-21
Allow build scripts to report error messages through `cargo::error`

torhovland:
Related to #10159.

Adds the `cargo::error` build script directive. It is similar to `cargo::warning`, but it emits an error message and it also fails the build.


tests/testsuite/build_script.rs
```diff
@@ -3925,14 +3929,16 @@ fn warnings_emitted_when_build_script_panics() {
 }
 
 #[cargo_test]
-fn warnings_hidden_for_upstream() {
+fn errors_and_warnings_hidden_for_upstream() {
```
epage, line 3936: The name of this test implies errors are hidden but they aren't.

In general, I'm feeling we should probably test errors and warnings separately.

tests/testsuite/warn_on_failure.rs
```diff
@@ -90,17 +88,14 @@ fn no_error_or_warning_on_bin_failure() {
...
-
-"#]])
+        .with_stderr_does_not_contain("hidden stderr")
+        .with_stderr_does_not_contain(&format!("[ERROR] {}", ERROR))
+        .with_stderr_does_not_contain(&format!("[WARNING] {}", WARNING1))
+        .with_stderr_does_not_contain(&format!("[WARNING] {}", WARNING2))
+        .with_stderr_contains("[UPDATING] `[..]` index")
+        .with_stderr_contains("[DOWNLOADED] bar v0.0.1 ([..])")
+        .with_stderr_contains("[COMPILING] bar v0.0.1")
+        .with_stderr_contains("[COMPILING] foo v0.0.1 ([..])")
```
epage, line 98: Why was the test style changed?  We are trying to move away from `does_not_contain` and `contains`

tests/testsuite/warn_on_failure.rs
```diff
@@ -113,16 +108,13 @@ fn errors_and_warnings_on_lib_failure() {
...
-[LOCKING] 2 packages to latest compatible versions
-[DOWNLOADING] crates ...
-[DOWNLOADED] bar v0.0.1 (registry `dummy-registry`)
-[COMPILING] bar v0.0.1
-[ERROR] invalid output in build script of `bar v0.0.1`: `cargo::error=Hello! I'm an error. :)`
-Unknown key: `error`.
-See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
-
-"#]])
+        .with_stderr_does_not_contain("hidden stderr")
```
epage, line 111: ditto

src/doc/src/reference/build-scripts.md
```diff
@@ -313,13 +314,18 @@ link-arg=FLAG` option][link-arg] to the compiler, but only when building a
...
+script has finished running. Errors are only shown for `path` dependencies
+(that is, those you're working on locally), so for example errors printed
 out in [crates.io] crates are not emitted by default. The `-vv` "very verbose"
-flag may be used to have Cargo display warnings for all crates.
+flag may be used to have Cargo display errors for all crates.
+
+### `cargo::warning=MESSAGE` {#cargo-warning}
+
+The `warning` instruction is just like [`error`](#cargo-error), except that it
+will output warnings.
```
epage, line 328: This does not match what the cargo team accepted in https://github.com/rust-lang/cargo/issues/10159#issuecomment-988104375

`cargo::error` should force the build to fail just like a bad exit code, independent of where it is in the dependency graph.  This means we should always print a `cargo::error`, even for transitive dependencies.

tests/testsuite/build_script.rs
epage: Ah, this looks to be part of a larger misunderstanding, see https://github.com/rust-lang/cargo/pull/14435#discussion_r1725100785

tests/testsuite/warn_on_failure.rs
torhovland: It's actually not changed overall. I upgraded the style in the first commit, but left it as it was in this one, so that the PR as a whole only changes one line. But I can update the styles overall.

src/doc/src/reference/build-scripts.md
torhovland: My thinking was to leave this for now, as there was a discussion about this later in the issue. But I can change it.

tests/testsuite/warn_on_failure.rs
epage: Ah, it was switched back and forth

src/doc/src/reference/build-scripts.md
torhovland: I don't think a feature flag makes a lot of sense here, as a library that emits `cargo::error` would have to require that any client app runs with the feature flag enabled. I'll go with alternative 1.

torhovland:
@epage This is ready for another look.

tests/testsuite/build_script.rs
```diff
@@ -3871,14 +3871,104 @@ fn warnings_emitted() {
...
+                }
+            "#,
+        )
+        .build();
+
+    p.cargo("build")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[COMPILING] foo v0.5.0 ([ROOT]/foo)
+[ERROR] invalid output in build script of `foo v0.5.0 ([ROOT]/foo)`: `cargo::error=foo err`
```
epage: nit: to really make the behavior obvious in the test change, this commit could use the `cargo::warning`s instead of `cargo::error`s and then the following commit switches it with the introduction of `cargo::error`

Went ahead and tried this out and it works quite well.  Looks like I don't have permission to edit your branch.  Feel free to pull in https://github.com/epage/cargo/tree/build-script-errors

tests/testsuite/build_script.rs
```diff
@@ -4030,20 +4270,10 @@ fn warnings_printed_on_vv() {
...
+...
 [WARNING] bar@0.1.0: foo
 [WARNING] bar@0.1.0: bar
-[RUNNING] `[..] rustc --crate-name bar [..]`
-[COMPILING] foo v0.5.0 ([ROOT]/foo)
-[RUNNING] `[..] rustc --crate-name foo [..]`
-[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
-
+...
 "#]])
```
epage, line 4275: There a reason this test was changed?  Using `...` makes things not update as cleanly

tests/testsuite/warn_on_failure.rs
```diff
@@ -86,13 +86,23 @@ fn no_warning_on_bin_failure() {
...
+[DOWNLOADED] bar v0.0.1 (registry `dummy-registry`)
+[COMPILING] bar v0.0.1
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+error[E0425]: cannot find function `hi` in this scope
+ --> src/main.rs:1:13
+  |
+1 | fn main() { hi() }
+  |             ^^ not found in this scope
+
+For more information about this error, try `rustc --explain E0425`.
```
epage, line 102: We try to use wildcards in place of rustc error messages to make it easier for them to change them

tests/testsuite/warn_on_failure.rs
```diff
@@ -86,13 +86,23 @@ fn no_warning_on_bin_failure() {
...
         .with_status(101)
         .with_stdout_does_not_contain("hidden stdout")
-        .with_stderr_does_not_contain("hidden stderr")
-        .with_stderr_does_not_contain(&format!("[WARNING] {}", WARNING1))
-        .with_stderr_does_not_contain(&format!("[WARNING] {}", WARNING2))
-        .with_stderr_contains("[UPDATING] `[..]` index")
-        .with_stderr_contains("[DOWNLOADED] bar v0.0.1 ([..])")
-        .with_stderr_contains("[COMPILING] bar v0.0.1")
-        .with_stderr_contains("[COMPILING] foo v0.0.1 ([..])")
+        .with_stderr_data(str![[r#"
```
epage, line 89: Ideally this cleanup would be in its own commit

src/doc/src/reference/build-scripts.md
```diff
@@ -313,13 +314,19 @@ link-arg=FLAG` option][link-arg] to the compiler, but only when building a
...
 ### `cargo::warning=MESSAGE` {#cargo-warning}
 
 The `warning` instruction tells Cargo to display a warning after the build
 script has finished running. Warnings are only shown for `path` dependencies
 (that is, those you're working on locally), so for example warnings printed
-out in [crates.io] crates are not emitted by default. The `-vv` "very verbose"
-flag may be used to have Cargo display warnings for all crates.
+out in [crates.io] crates are not emitted by default, unless the build fails.
+The `-vv` "very verbose" flag may be used to have Cargo display warnings for
+all crates.
```
epage, line 329: This should be its own commit

tests/testsuite/build_script.rs
torhovland: Done.

tests/testsuite/build_script.rs
torhovland: Reverted to the old snapshot.

tests/testsuite/warn_on_failure.rs
torhovland: Fixed.

tests/testsuite/warn_on_failure.rs
torhovland: Done.

src/doc/src/reference/build-scripts.md
torhovland: Fixed.

torhovland:
@epage This is ready for another look.

src/doc/src/reference/build-scripts.md
```diff
@@ -313,13 +314,19 @@ link-arg=FLAG` option][link-arg] to the compiler, but only when building a
 `cdylib` library target. Its usage is highly platform specific. It is useful
 to set the shared library version or the runtime-path.
 
+### `cargo::error=MESSAGE` {#cargo-error}
+
+The `error` instruction tells Cargo to display an error after the build script
+has finished running, and then fail the build.
```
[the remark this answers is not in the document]
torhovland: So... should I commit this?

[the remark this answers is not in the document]
weihanglo: I guess sk and we can always update it later

tests/testsuite/warn_on_failure.rs
```diff
@@ -95,7 +95,6 @@ fn no_warning_on_bin_failure() {
 [COMPILING] foo v0.0.1 ([ROOT]/foo)
 error[E0425]: cannot find function `hi` in this scope
 ...
-For more information about this error, try `rustc --explain E0425`.
```
epage: Could you squash this into the relevant commit?

src/doc/src/reference/build-scripts.md
```diff
@@ -319,6 +319,8 @@ to set the shared library version or the runtime-path.
 The `error` instruction tells Cargo to display an error after the build script
 has finished running, and then fail the build.
 
+> **MSRV:** Respected as of 1.82
```
epage, line 322: Keeping a comment open on this so we remember to update it if this slips

epage:
@rfcbot fcp merge

@torhovland is proposing we insta-stabilize `cargo::error=` build directive which was requested in #10159.

This is following the semantics proposed in a cargo team meeting (https://github.com/rust-lang/cargo/issues/10159#issuecomment-988104375):  `cargo::error=` forces the build script to fail, independent of whether its local or not

@kornelski counters that this means we have two ways to cause a build script to error when we should only have one.  They suggest warning if an error appears without a failure (https://github.com/rust-lang/cargo/issues/10159#issuecomment-2305139091).  However, that does mean that we have error messages that might not be tied to a failure which can cause user confusion and I think we should keep the Cargo team's previous proposal on this.

This is only supported with `cargo::` syntax and not `cargo:` syntax.  Use of `cargo:error` would create an `error` metadata entry.  As `cargo::` is strictly checked, this will require an MSRV bump to use which is good in this case if people are relying on `cargo::error` causing a bad  exit code.  Because `cargo::error` will fail on previous versions of Cargo, no transition period should be needed for this.

As for insta-stabiliaze, I can understand that it seems like a lot to create an unstable feature just for this build directive.  We don't have anything like `-Zunstable-options` for build directives.  This is relatively small in scope and low level / mechanism focused, that besides @kornelski's counter proposal, I'm not seeing much we can learn from an extended testing period.  I also don't expect much nightly testing on this.

tests/testsuite/warn_on_failure.rs
torhovland: Done.

kornelski:
The primary motivation for this feature was #10159, where the main problem was noisy output from `pkg_config`. However, I don't see how to integrate `cargo::error` with the `pkg_config` crate ;(

[By default](https://docs.rs/pkg-config/latest/pkg_config/struct.Config.html#method.cargo_metadata) `pkg_config` prints cargo metadata as a side effect. But this metadata must not contain `cargo::error`, because the API returns a `Result` and many crates try a fallback after that. `pkg_config` can't stuff `cargo::error` prefix in `Display` of `pkg_config::Error` either, because some crates have `printf("cargo:warning={pkg_config_error}");` and that would nest syntax, or `cargo::error` after a newline would fail the build instead of warning.

ehuss:
I would assume that the caller of pkg-config would be responsible for printing `cargo::error` with the appropriate error if they want to. Can you explain why that might not work?

kornelski:
@ehuss The downside is that instead of improving one crate to benefit the whole ecosystem, it'll be necessary to change 750 downstream crates.

 `pkg_config` instructs users to `.unwrap()` it: https://docs.rs/pkg-config/latest/pkg_config/#example

system-deps [too](https://docs.rs/system-deps/latest/system_deps/#usage).

which means these crates can't adopt `cargo:error` on their own, in their current API, to automatically provide a good baseline to all current users who `unwrap()` their errors.

epage:
@torhovland looks like some other changes cause the output in the new tests to change and the commits need to be updated.

ehuss:
Closing in favor of #14743.

PR #14436, opened 2024-08-21
More helpful missing feature error message

kornelski:
The "none of the selected packages contains these features" error does not say which packages are selected, and does not say which packages do have the requested features.

If any package in the workspace has the requested feature (matching exactly), the error suggests selecting that package, instead of trying alternative spellings.

### DIFF
--- a/src/cargo/core/workspace.rs
+++ b/src/cargo/core/workspace.rs
@@ -1363,12 +1363,12 @@ impl<'gctx> Workspace<'gctx> {
         }
     }
 
-    fn report_unknown_features_error(
+    fn missing_feature_spelling_suggestions(
         &self,
-        specs: &[PackageIdSpec],
+        selected_members: &[&Package],
         cli_features: &CliFeatures,
         found_features: &BTreeSet<FeatureValue>,
-    ) -> CargoResult<()> {
+    ) -> Vec<String> {
         // Keeps track of which features were contained in summary of `member` to suggest similar features in errors
         let mut summary_features: Vec<InternedString> = Default::default();
 
@@ -1387,10 +1387,7 @@ impl<'gctx> Workspace<'gctx> {
         let mut optional_dependency_names_per_member: BTreeMap<&Package, BTreeSet<InternedString>> =
             Default::default();
 
-        for member in self
-            .members()
-            .filter(|m| specs.iter().any(|spec| spec.matches(m.package_id())))
-        {
+        for &member in selected_members {
             // Only include features this member defines.
             let summary = member.summary();
 
@@ -1428,7 +1425,7 @@ impl<'gctx> Workspace<'gctx> {
             edit_distance(a.as_str(), b.as_str(), 3).is_some()
         };
 
-        let suggestions: Vec<_> = cli_features
+        cli_features
             .features
             .difference(found_features)
             .map(|feature| match feature {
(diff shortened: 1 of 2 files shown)


src/cargo/core/workspace.rs
```diff
@@ -1522,27 +1519,83 @@ impl<'gctx> Workspace<'gctx> {
...
-                suggestions.join(", ")
+            let names = selected_members.iter().map(|m| m.name()).join(", ");
+            format!("none of the selected packages contains {these_features}: {}\nselected packages: {names}", unknown.join(", "))
+        };
+
+        use std::fmt::Write;
+        if !missing_packages_with_the_features.is_empty() {
+            write!(
+                &mut msg,
+                "\nhelp: include package{} with the missing feature{}:",
```
epage, line 1572: I feel like this better matches the rustc style of "saying options exist"
```suggestion
                "\nhelp: package{} with the missing feature{}:",
```

src/cargo/core/workspace.rs
```diff
@@ -1522,27 +1519,83 @@ impl<'gctx> Workspace<'gctx> {
...
+                    "s"
+                } else {
+                    ""
+                },
+                if unknown.len() != 1 { "s" } else { "" }
+            )?;
+            for member in &missing_packages_with_the_features {
+                write!(&mut msg, " --package {member}")?;
+            }
+        } else {
```
epage, line 1583: Should we show both hints?

src/cargo/core/workspace.rs
```diff
@@ -1522,27 +1519,83 @@ impl<'gctx> Workspace<'gctx> {
...
+            let suggestions = self.missing_feature_spelling_suggestions(
+                &selected_members,
+                cli_features,
+                found_features,
             );
+            if !suggestions.is_empty() {
+                write!(
+                    &mut msg,
+                    "\nhelp: did you mean: {}?",
```
epage, line 1592: non-blocker (since existing issue): `did you mean` is against rustc's [suggestion style guide](https://rustc-dev-guide.rust-lang.org/diagnostics.html#suggestion-style-guide) and we should be moving away from it. 

However, if we show both hints, we should probably reword this so the two hints work in tandem with each other.

epage:
Thanks for taking the time to improve this!

src/cargo/core/workspace.rs
kornelski: No, I don't think so. If the workspace has a package with this exact feature name, then it's very unlikely to be a misspelling.

I've made it like this after adding the tests, and seeing how silly it looks when I use the correct real feature name, and it suggests an unrelated one.

src/cargo/core/workspace.rs
kornelski: The suggestion includes the `-p` flag which I think is helpful, but without the verb it's weird to have flags there.

epage: You're right, I had missed the `-p`.

For myself, I'd prefer dropping `include` and dropping `--package`.

I feel like `include packages with the following features" is an imperative mood (we are telling the user what to do) which doesn't feel appropriate in this circumstance (or most).  

rustc's style guide does offer

> The message may contain further instruction such as "to do xyz, use" or "to do xyz, use abc".

In this case that would be "to include a package with the missing features, add one of these..."

I feel like that is best used when we are teaching someone which I don't think is needed here and it should build on information we've already mentioned which would be that packages exist with this feature but that comes later.

src/cargo/core/workspace.rs
epage: Ok.  I had thought there were real world cases where we might want both but I can't reconstruct them at this time.  We can always change it later.

src/cargo/core/workspace.rs
epage: As an example where telling the user what to do, I commonly have features through most of my workspace and I might leave it enabled when building on a package without it.  My action item would be to remove the feature, not to specify a different package.

epage:
Thanks!

@bors r+

PR #14439, opened 2024-08-21
move modules from kebab-case to snake_case

DSchroer:

It is currently unclear reading the docs how files should be named. [RFC 430](https://rust-lang.github.io/rfcs/0430-finalizing-naming-conventions.html) is clear on crates and module names, `kebab-case` is never mentioned and while there are historical considerations, we should guide new development towards `snake_case`. We should align the documented defaults to that RFC. 

Review the docs and determine if we view value in the change.

### DIFF
--- a/src/doc/src/guide/project-layout.md
+++ b/src/doc/src/guide/project-layout.md
@@ -48,6 +48,8 @@ files, place a `main.rs` file along with the extra [*modules*][def-module]
 within a subdirectory of the `src/bin`, `examples`, `benches`, or `tests`
 directory. The name of the executable will be the directory name.
 
+> **Note:** By convention, binaries, examples, benches and integration tests follow `kebab-case` naming style, unless there are compatibility reasons to do otherwise (e.g. compatibility with a pre-existing binary name). Modules within those targets are `snake_case` following the [Rust standard](https://rust-lang.github.io/rfcs/0430-finalizing-naming-conventions.html).
+
 You can learn more about Rust's module system in [the book][book-modules].
 
 See [Configuring a target] for more details on manually configuring targets.


cristian-sandru (approved):

src/doc/src/guide/project-layout.md
```diff
@@ -11,24 +11,24 @@ Cargo [package][def-package]:
 │   ├── lib.rs
 │   ├── main.rs
 │   └── bin/
-│       ├── named-executable.rs
-│       ├── another-executable.rs
-│       └── multi-file-executable/
+│       ├── named_executable.rs
+│       ├── another_executable.rs
+│       └── multi_file_executable/
```
epage, line 16: I believe this would have end-user affects, making the default binary names use `_` when the status quo is for binaries to use `-`.

As these are not just `mod` names but build target names, i would be inclined to leave all of these as `-`

ehuss:
Thanks for the PR! I think there may be some confusion, as these filenames do not represent crate names, they are target names. Considering that there isn't a standard for how those should be named, I'm disinclined to change these here. Indeed, cargo's own repository makes use of hyphen in most places.

DSchroer:
> Considering that there isn't a standard for how those should be named, I'm disinclined to change these here. Indeed, cargo's own repository makes use of hyphen in most places.

Should we call that out in this page then? I am happy with `kebab-case` for build targets as they do seem out of scope from that RFC. However we should spell it out to avoid future confusion.

DSchroer:
I have tried to add some clarity by explicitly stating my current understanding here. What do you think? 

One thing I debated was if using a single word for targets should be recommended but that does not feel right.

src/doc/src/guide/project-layout.md
```diff
@@ -48,6 +48,8 @@ files, place a `main.rs` file along with the extra [*modules*][def-module]
 within a subdirectory of the `src/bin`, `examples`, `benches`, or `tests`
 directory. The name of the executable will be the directory name.
 
+Binaries, examples, benches and integration tests follow `kebab-case` naming style. Modules within those targets are `snake_case` following the [Rust standard](https://rust-lang.github.io/rfcs/0430-finalizing-naming-conventions.html).
```
epage, line 51: This is likely best left to a style guide, with us finding an appropriate place to link out to it.

DSchroer: Is there a style guide that makes sense to use?

DSchroer: The way I see it, right now no one in the rust ecosystem directly says that build targets are `kebab-case`. This document is the closest but it is not clear.

DSchroer:
Are [cargo targets](https://doc.rust-lang.org/cargo/appendix/glossary.html#target) a cargo concept? If so can we define the naming convention somewhere that says those targets should be `kebab-case`?

src/doc/src/guide/project-layout.md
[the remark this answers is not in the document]
DSchroer: The thing is without a codified standard its hard to enforce. So people will start drifting and you will end up with multiple groups doing different things. If it becomes part of the docs then its easier to PR lints into Clippy in order to enable automatic enforcement for projects and makes the ecosystem easier and more coherent.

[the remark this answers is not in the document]
weihanglo: Thank you for explaining that :)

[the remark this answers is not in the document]
epage: > If it becomes part of the docs then its easier to PR lints into Clippy in order to enable automatic enforcement for projects and makes the ecosystem easier and more coherent.

I suspect this lint should live in cargo, not clippy.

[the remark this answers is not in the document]
epage: I disagree that this is the place for it.

Also, even better than documenting this is to lint for it.  If we have specific naming recommendations, we can use #12235 (unstable implementation is in place) to lint for naming like rustc does for functions, structs, etc.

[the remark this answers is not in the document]
DSchroer: > Also, even better than documenting this is to lint for it.

100% agree. However you end up in a chicken and egg problem. If I propose a semantic lint like this with no docs to support it, then I will end up with a really hard time getting it merged. So its just a first step here to remove an ambiguous situation.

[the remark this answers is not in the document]
DSchroer: So where specifically should I try to put this?

src/doc/src/guide/project-layout.md
joshtriplett, line 51:
```suggestion
By convention, binaries, examples, benches and integration tests follow `kebab-case` naming style, unless there are compatibility reasons to do otherwise (e.g. compatibility with a pre-existing binary name). Modules within those targets are `snake_case` following the [Rust standard](https://rust-lang.github.io/rfcs/0430-finalizing-naming-conventions.html).
```

epage: Could you also include the changes from @joshtriplett ?  The current text is very prescriptive, leaving out room for people to make decisions to follow other conventions when there is a motivating case.

DSchroer: Sure. Made the modification.

DSchroer:
Updated to a single squashed commit.

epage (approved):

PR #14440, opened 2024-08-21
refactor(update): Prepare for smarter update messages

epage:

This is to make it easier to
- Make #14401 path aware
- Work on #13908, depending on what is decided
- Improve the heuristics for when we show `Locking` messages

There are fixes along the way that were found by making the code more consistent in prep for consolidating it, mostly for #14401.

Along the way, some odd code structures are used for the sake of making the diff easier to follow.  These get cleaned up towards the end

I didn't add tests for the fixes because of the code consolidation that happens that causes us to cover more real-world scenarios with fewer tests.


Muscraft (approved):
Thanks!

PR #14441, opened 2024-08-21
Tests rely on absence of RUST_BACKTRACE

kornelski:
I have it set in my default environment, and it causes tests to fail


weihanglo (approved):
Sounds correct. Will merge this when CI is green. Thank you!

PR #14445, opened 2024-08-22
fix(resolve): Dont show locking workspace members 

epage:
This is for `cargo generate-lockfile` and when syncing the lockfile with
the manifest.
We still show it for `cargo update` because of `cargo update
--workspace`.

We hacked around this previously by filtering out the `num_pkgs==1` case
for single packages but this didn't help with workspaces.

### Additional information

This builds on #14440

### DIFF
--- a/src/cargo/ops/cargo_update.rs
+++ b/src/cargo/ops/cargo_update.rs
@@ -492,15 +492,21 @@ fn print_lockfile_generation(
     resolve: &Resolve,
     registry: &mut PackageRegistry<'_>,
 ) -> CargoResult<()> {
-    let changes = PackageChange::new(resolve);
-    let num_pkgs: usize = changes.iter().filter(|change| change.kind.is_new()).count();
-    if num_pkgs <= 1 {
-        // just ourself, nothing worth reporting
+    let changes = PackageChange::new(ws, resolve);
+    let num_pkgs: usize = changes
+        .iter()
+        .filter(|change| change.kind.is_new() && !change.is_member.unwrap_or(false))
+        .count();
+    if num_pkgs == 0 {
+        // nothing worth reporting
         return Ok(());
     }
     status_locking(ws, num_pkgs)?;
 
     for change in changes {
+        if change.is_member.unwrap_or(false) {
+            continue;
+        };
         match change.kind {
             PackageChangeKind::Added => {
                 let possibilities = if let Some(query) = change.alternatives_query() {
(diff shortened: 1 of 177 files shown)


src/cargo/ops/cargo_update.rs
```diff
@@ -787,20 +800,23 @@ struct PackageChange {
...
-    pub fn diff(previous_resolve: &Resolve, resolve: &Resolve) -> Vec<Self> {
+    pub fn diff(ws: &Workspace<'_>, previous_resolve: &Resolve, resolve: &Resolve) -> Vec<Self> {
         let diff = PackageDiff::diff(previous_resolve, resolve);
-        Self::with_diff(diff)
+        Self::with_diff(diff, ws)
     }
 
-    fn with_diff(diff: impl Iterator<Item = PackageDiff>) -> Vec<Self> {
+    fn with_diff(diff: impl Iterator<Item = PackageDiff>, ws: &Workspace<'_>) -> Vec<Self> {
+        let member_ids: HashSet<_> = ws.members().map(|p| p.package_id()).collect();
```
weihanglo, line 758: Use `Workspace::is_member` instead?

src/cargo/ops/cargo_update.rs
```diff
@@ -787,20 +800,23 @@ struct PackageChange {
     package_id: PackageId,
     previous_id: Option<PackageId>,
     kind: PackageChangeKind,
+    is_member: Option<bool>,
```
weihanglo, line 743: I can't see the necessity of wrapping with an `Option` here except for representing “not relevant because it is a removed package”. Is that the reason of using it?

src/cargo/ops/cargo_update.rs
weihanglo, line 1: The truth is that Cargo actually also locks versions of workspaces members, though I believe users often just don't care.

src/cargo/ops/cargo_update.rs
epage: Thats annoying, it requires a `Package` just to do a lookup with `PackageId`

src/cargo/ops/cargo_update.rs
epage: Yes, I wanted to make sure that when I accessed it, this state was clear.  At first, I thought I was also going to have to act on it but that didn't end up happening.

src/cargo/ops/cargo_update.rs
epage: Yes.  When I first did this, the intent was to inform users of relevant information.  It can also be confusing to see these mesaages.  I just made a change to some of my workspaces, removing `version` and `cargo check` reported that it downgraded those packages.

epage:
This PR had a refactor on it, merging the `cargo generate-lockfile` and the lockfile sync (e.g. `cargo check`) reporting but I realized there was a difference that we might want to preserve: `cargo generate-lockfile` only shows `Adds` with relevant information to avoid overwhelming the user.

weihanglo (approved):

weihanglo:
@bors r+

Thanks!

PR #14448, opened 2024-08-24
bail before packaging on same version 

### LINKED ISSUE #3662: Fail fast when package version is already published

stupendoussuperpowers:
Fixes #3662. Cleaned up commits from #14338.

### DIFF
--- a/tests/testsuite/publish.rs
+++ b/tests/testsuite/publish.rs
@@ -130,6 +130,58 @@ You may press ctrl-c to skip waiting; the crate should be available shortly.
     validate_upload_foo();
 }
 
+#[cargo_test]
+fn duplicate_version() {
+    let registry_dupl = RegistryBuilder::new()
+        .http_api()
+        .http_index()
+        // test registry doesn't error on duplicate versions, we need to
+        .add_responder("/api/v1/crates/new", move |_req, _server| Response {
+            code: 200,
+            headers: vec![],
+            body: br#"{"errors": [{"detail": "crate version `0.0.1` is already uploaded"}]}"#
+                .to_vec(),
+        })
+        .build();
+
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.0.1"
+                authors = []
+                license = "MIT"
+                description = "foo"
+            "#,
+        )
+        .file("src/main.rs", "fn main() {}")
+        .build();
+
+    p.cargo("publish")
+        .replace_crates_io(registry_dupl.index_url())
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[UPDATING] crates.io index
+[WARNING] manifest has no documentation, homepage or repository.
+See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for more info.
+[PACKAGING] foo v0.0.1 ([ROOT]/foo)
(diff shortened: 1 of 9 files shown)


stupendoussuperpowers:
r? @epage 

Comments are rebased into 3 like we discussed on the last PR.

tests/testsuite/publish.rs
```diff
@@ -158,23 +162,17 @@ fn duplicate_version() {
         .file("src/main.rs", "fn main() {}")
         .build();
 
+    p.cargo("publish")
+        .replace_crates_io(registry_dupl.index_url())
+        .without_status()
+        .run();
```
epage, line 155: This is publishing the package to ensure we get the error.

Does
```rust
    Package::new("too_new_child", "0.0.1")
        .rust_version("1.2345.0")
        .file("src/lib.rs", "fn other_stuff() {}")
        .publish();
```
work with the `RegistryBuilder` created earlier?

tests/testsuite/publish.rs
```diff
@@ -132,14 +132,18 @@ You may press ctrl-c to skip waiting; the crate should be available shortly.
...
-                .to_vec(),
+        .add_responder("/index/3/f/foo", move |req, server| {
+            let mut lock = arc.lock().unwrap();
+            *lock += 1;
+            if *lock <= 1 {
+                server.not_found(req)
+            } else {
+                server.index(req)
+            }
         })
```
epage: If we have a published package, would `RegistryBuilder` do this for us or does it not correctly handle these cases?

epage:
> Cleaned up commits

FYI what would be very helpful in reviewing would be
- Have the full structure of the tests in place in the test commit.  This would mean having the extra publish in that commit and the `RegistryBuilder` changes
- Squashing the text change into the primary commit
- Ensuring every commit passes tests

ehuss:
I do not think this fixes https://github.com/rust-lang/cargo/issues/3662. We have been using that as a general-purpose issue to track validation of all sorts. There's a more detailed description of the pre-flight validation that we were considering in https://github.com/rust-lang/crates-io-cargo-teams/issues/82 that I think would need to be addressed to close #3662.

stupendoussuperpowers:
This fix is one of many pre-flight checks mentioned in https://github.com/rust-lang/crates-io-cargo-teams/issues/82. Should all those checks be worked on separately or should they be all be added in the same PR? The discussion on #3662 mostly mentioned just the same version pre-flight check.

ehuss:
I think it is good to land this PR; I don't think we need to block incremental improvements. The registry-based validation would be nice, and in theory could address all of the issues in one go, but I think is also significantly more work and will likely take a very long time even if someone worked hard on it.

epage:
@ehuss what do you see the difference is between #3662 and #4377?

ehuss:
Those look like duplicates of one another, though #4377 has more details.

tests/testsuite/publish.rs
epage: This works with the fix, replacing the extra `cargo publish` call.

tests/testsuite/publish.rs
epage: We do need a custom responder before the fix because our dev registry does not report duplicate package errors, so we can't just double publish to hit the test case.

epage:
I went ahead and resolved the rebase and applied my review feedback

@bors r+

PR #8141, opened 2020-04-21
Uplift windows gnu DLL import libraries.

ehuss:
This is a repost of #6875, rebased with tests fixed.

PR #14451, opened 2024-08-25
uplift windows gnullvm import libraries

jschwe:
Same changes as #8141, but for gnullvm.
gnullvm does not seem to be tested in CI, so tests were not adjusted.

The `implib` for [gnullvm](https://doc.rust-lang.org/rustc/platform-support/pc-windows-gnullvm.html) targets should end up at the same location as for `gnu` targets.

### Background

I'm not a windows developer, so I may be wrong, but I believe the implib artifacts for gnullvm should be needed in the same way as for the normal windows-gnu target. In my downstream code, I need to know the location of the implib artifact, and would prefer for the location to be the same for both gnullvm and normal gnu.

### DIFF
--- a/src/cargo/core/compiler/build_context/target_info.rs
+++ b/src/cargo/core/compiler/build_context/target_info.rs
@@ -389,7 +389,10 @@ impl TargetInfo {
                     crate_type: Some(crate_type.clone()),
                     should_replace_hyphens: true,
                 });
-            } else if target_triple.ends_with("windows-gnu") && suffix == ".dll" {
+            } else if suffix == ".dll"
+                && (target_triple.ends_with("windows-gnu")
+                    || target_triple.ends_with("windows-gnullvm"))
+            {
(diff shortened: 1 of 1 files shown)


jschwe:
CC the target maintainers @mati865 and @thomcc

mati865 (approved):
Although LLD is able to link DLL directly in most of the cases, it's preferred to use an import library.
Thank you for the fix.

jschwe:
I guess there is no chance of this getting merged before the next beta tomorrow, is there?

weihanglo:
> I guess there is no chance of this getting merged before the next beta tomorrow, is there?

1.82 beta has already branched a few days ago.

weihanglo (approved):
I don't have the test infra either, though this seems pretty reasonable.

Given target maintainers @mati865 and @thomcc has given 👍🏾 to this change, I am going to merge this. Thank you for the patch!

Issue #2608, opened 2016-04-22
Start requiring a reason for all yanks


alexcrichton:
@wycats, @nikomatsakis and I have discussed recently the ability to yank a crate with a reason, and then whenever that yanked crate is compiled locally it will present the reason as a warning.

The intended use cases for this would be:

* When a crate is fixed because it will be broken in the next version of the compiler (e.g. a soundness fix or bug fix) then the previous versions can be yanked and nudge users forward.
* If a crate is fixed for a security reason, the old versions can be yanked and the new version can be suggested
* If a crate is renamed (or perhaps deprecated) to another then the yank message can indicate what to do in that situation.

Technically, I believe this should be implemented through:

1. Update crates.io to accept an optional message for yanks. Store it in the database.
2. Update `cargo yank` to **require a message** indicating why the crate is being yanked.
3. Change the logic for **when a yanked crate is compiled** to something like:
  * Use whatever state the registry is currently in, do not update it.
  * Only issue a warning if the crate is compiled, not if it is "fresh" (e.g. has already been compiled).
  * When compilation starts, fire off an asynchronous HTTP request to crates.io to fetch the yank message.
  * Print out the warning whenever the HTTP request finishes or block the build finishing on these HTTP requests to return.
  * Ignore any errors that happen fetching the yank message, they should not fail the build.

I think this should suffice as an "opportunistic nudge" for these situations where possible, although it's certainly no form of ironclad guarantee.

Thoughts @wycats, @nikomatsakis?

Mark-Simulacrum:
If we can standardize the yank text to a small list we could store it in the index and avoid the HTTP overhead for the common case.

kornelski:
For the yank of futures 0.2 something like "please use futures-preview instead" would be a useful yank message. Small list of predefined messages wouldn't cover it, unless you allowed parametrized messages, but that might be too much complexity for little gain.
If yank messages have a length limit (e.g. 80 chars?) putting them in the index shouldn't be too bad.

tarcieri:
Note: there's some discussion of this issue in the context of this thread: https://internals.rust-lang.org/t/pre-rfc-reviving-security-advisories-in-crates-io-rfc-pr-1752/9017/11?u=bascule

Namely that if the "yank message" that gets persisted to the crates.io index can be something like a structured TOML file, a mechanism like this could potentially be used for storing RustSec security advisories.

However, for that purpose, it would be helpful if the "yank message" could span yanking multiple releases of the same crate.

[...]

PR #14452, opened 2024-08-25
Workaround for type inference breakage in the time crate

kornelski:
The `time` crate is widely used, and the compatibility issue https://github.com/rust-lang/rust/issues/127343 affects many crates. This may have long-lasting consequences, as the broken versions of the `time` crate are locked in lockfiles.

If @jhpratt doesn't mind this, I propose here breaking a taboo, and slightly modify the already published crates to fix the incompatibility.

With this patching-patch the affected `time` crates build again with Rust 1.80. It's an ad-hoc bit of code in tar unpacker, because hopefully Cargo won't need a proper infrastructure for this.

### DIFF
--- a/src/cargo/sources/registry/mod.rs
+++ b/src/cargo/sources/registry/mod.rs
@@ -195,6 +195,7 @@ use std::task::{ready, Poll};
 use anyhow::Context as _;
 use cargo_util::paths::{self, exclude_from_backups_and_indexing};
 use flate2::read::GzDecoder;
+use semver::VersionReq;
 use serde::Deserialize;
 use serde::Serialize;
 use tar::Archive;
@@ -694,6 +695,10 @@ impl<'gctx> RegistrySource<'gctx> {
             .open(&path)
             .with_context(|| format!("failed to open `{}`", path.display()))?;
 
+        if let Err(e) = patch_extracted_package_if_needed(pkg, unpack_dir) {
+            tracing::warn!("error patching {pkg}: {e}");
+        }
+
         let lock_meta = LockMetadata { v: 1 };
         write!(ok, "{}", serde_json::to_string(&lock_meta).unwrap())?;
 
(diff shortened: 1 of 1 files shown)


jhpratt (approved):
While certainly unconventional, this does work around the problem about as much as possible.

Without checking the version bounds (I presume you did), the patch itself is acceptable to me. Whether this _should_ be done is presumably up to the relevant team, so I'll defer to them on that.

tgross35:
Repeating a suggestion from the possible mitigations in the compiler: it may be good to add a `// FIXME(2025-Q2): ...` as a reminder that this should be revisited at some point in the future. It might get dropped at that point or it might be extended based on how much usage of the old version still exists, but just something to note this isn't intended to be a permanent hack in Cargo.

alyssais:
For Nixpkgs, I think the best thing would be to have an environment variable to turn this off, given we've already mostly handled the breakage anyway.  I imagine an opt out for Cargo modifying crate sources is desirable for other reasons as well.

kornelski:
This is applied only when unpacking. For users with the crate already in unpacked, I'd need to add some extra check.

Perhaps patching could copy the whole crate to another directory in case tools check the integrity of the original dir, but I'll need to check if that works with the rest of the build pipeline.

emilazy:
It’s pretty big, in my opinion. It silently alters the source tree of a huge number of historical crate dependency graphs from what their locked hashes say they should be. It seems like a dangerous precedent to set and the compiler would be the more appropriate way to hack around this, if a hack is desired. I’m biased because it’s bad for Nixpkgs, but I don’t think we’d be the only ecosystem with serious concerns about this.

(Ideally the compiler would have deployed a workaround *before* it hit a stable release…)

epage:
> While discussion on how these hotfixes should be done in general is no doubt useful, trying to properly come up with a general solution before merging this would delay this hotfix in particular indefinitely. 

To be clear, I was not saying this had to be blocked on those longer term items.  This is blocked on the shorter term discussion of "what are our options and what is the best of those" which needs to be discussed across teams.

This is an invasive change and a major policy change and should not be done lightly, as others have said.

epage:
As this is initially a cross-team discussion, I started a focused thread at https://internals.rust-lang.org/t/brainstorming-how-to-help-old-code-locked-to-old-time-build-on-new-toolchains/21438/2

ehuss:
We understand that the `time` issue has caused a lot of unexpected churn and like you are interested in finding ways of improving things now and in the future.  To that end, we devoted most of @rust-lang/cargo's meeting to discuss this topic. We considered the possibility of patching the specific ranges of versions of the time crate, by various means. We're concerned about trying to ship a feature like this in a hurry, and potentially unexpected fallout of it, now and in the future:

- Cargo crossing the deontological line of patching users' sources and not building exactly the original sources.
- Having the user be surprised by this patching (e.g. because they were expecting the failure and were testing the failure).
- Having to take this patching into account for future features like caching and verification.

Given that, we would prefer to not move forward with a short-term fix for this issue with the `time` crate. We are however considering solutions that could help in the future. We have opened https://github.com/rust-lang/cargo/issues/14458 to explore those ideas. In particular, we would like to continue with more incremental work that has already begun, such as:

- [Yank reasons](https://rust-lang.github.io/rust-project-goals/2024h2/yank-crates-with-a-reason.html) which is currently a Project Goal, and has the potential to eventually ease towards a [mutable database](https://github.com/rust-lang/cargo/issues/2608#issuecomment-1932818615) that could possibly support registry patches.
- [Experiments with patches](https://github.com/rust-lang/cargo/pull/13779). That work halted due to concerns over resolver interactions. Resolver interaction was a major component behind that change, but a situation like this with `time` shows that perhaps a solution that does not support `Cargo.toml` changes would be useful.
- Providing an interface between rustc and cargo so that either side can better support reporting diagnostics related to situations like this. (https://github.com/rust-lang/compiler-team/issues/635 is tangentially related to this.)

A mitigation is being added to rustc via https://github.com/rust-lang/rust/pull/129343. We acknowledge that likely does not go far enough for many since it only provides a diagnostic and not a fix.

Other teams, including libs-api, lang, compiler, and release, are also working on improvements to try to better handle or prevent situations like this.

PR #14453, opened 2024-08-25
Log details of failure if no errors were seen

Mark-Simulacrum:

Crater has started (some time ago, though exact point is not known) to see failures in crates without logging an error. We're suspecting that there's something wrong in Cargo or rustc, but it's also possible that Crater itself is at fault. In any case, this should provide more details.

cc https://github.com/rust-lang/crater/issues/733

My hope is that this is reasonable to land without additional tests -- as-is, we expect it to never happen (there's even a debug assert later in the same code path) but it seems it *is* happening sometimes in the wild.

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -409,7 +409,17 @@ fn rustc(
                         )
                     },
                 )
-                .map_err(verbose_if_simple_exit_code)
+                .map_err(|e| {
+                    if output_options.errors_seen == 0 {
+                        // If we didn't expect an error, do not require --verbose to fail.
+                        // This is intended to debug
+                        // https://github.com/rust-lang/crater/issues/733, where we are seeing
+                        // Cargo exit unsuccessfully while seeming to not show any errors.
+                        e
+                    } else {
+                        verbose_if_simple_exit_code(e)
+                    }
+                })
                 .with_context(|| {
                     // adapted from rustc_errors/src/lib.rs
                     let warnings = match output_options.warnings_seen {


src/cargo/core/compiler/mod.rs
```diff
@@ -409,7 +409,17 @@ fn rustc(
                         )
                     },
                 )
-                .map_err(verbose_if_simple_exit_code)
+                .map_err(|e| {
+                    if output_options.errors_seen == 0 {
+                        // If we didn't expect an error, do not require --verbose to fail.
+                        // This is intended to debug
```
RalfJung, line 415: `verbose_if_simple_exit_code` doesn't do anything with `--verbose` so I am confused by the comment.

(I am also confused by `verbose_if_simple_exit_code`... the comment says that on signals and "abnormal termination", "don't hide the error details" -- but the code and function name seem to make it *more* verbose when it is a "simple" error, i.e. *not* a signal / abnormal termination.

RalfJung: Oh, I see... `VerboseError` is an error that is *less* verbose, except if `--verbose` is passed then it is always verbose. That's quite confusing.

weihanglo: Agree. Should have to rename it to something. Do you have any suggestion?

weihanglo (approved):

src/cargo/core/compiler/mod.rs
RalfJung: Maybe something like `SilenceErrorUnlessVerbose`?

PR #14457, opened 2024-08-27
fix(resolve): Report incompatible packages with precise Rust version

epage:

In #14401, we reported about MSRV issues as if we were the resolver.  We can be smarter than that though and report precise MSRV information.  This allows us to elevate the message from color from yellow to red.

This is also prep work for telling users when a newer, MSRV-compatible version of a package is available.

The report function I added is a little odd because a follow up commit will update it to report when a package is incompatible with rustc when the MSRV resolver is disabled and do this on stable.

### Additional information

This builds on #14445 to improve #14401


ehuss:
Thanks!

@bors r+

PR #14459, opened 2024-08-27
feat(resolve): Report incompatible-with-rustc when MSRV-resolver is disabled

epage:

This builds on #14457 to report when deps are incompatible with rustc when an MSRV-aware resolver is not used.
This affects stable code.


Muscraft (approved):
Thanks!

PR #14464, opened 2024-08-28
fix: avoid inserting duplicate `dylib_path_envvar` when calling `cargo run` recursively

### LINKED ISSUE #14194: Cargo appends paths to `PATH` while building without checking if those paths already exist
### Problem

`cargo` appears to be appending paths to the `PATH` environment variable without checking if those paths are already present in it, causing unneeded rebuilds in some situations. I would expect cargo to not add to the path if it doesn't have to.

### Steps

This issue can be reduced to a single file.
1. Create a new cargo project.
2. Enter the following in `main.rs`:
```rust
use std::process::Command;

fn main() {
    for path in std::env::split_paths(&std::env::var("PATH").unwrap()) {
        println!("{}", path.to_string_lossy());
    }

    if std::env::var("EXIT").is_ok() {
        return;
    }

    Command::new("cargo")
        .arg("run")
        .env("EXIT", "1")
        .status()
        .unwrap();
}
```
3. Inspect the printed `PATH` variables. The beginning of the path of the first cargo invocation should be:
```
C:\\Users\\natha\\OneDrive\\Desktop\\cargo-path-duplicate\\target\\debug\\deps
C:\\Users\\natha\\OneDrive\\Desktop\\cargo-path-duplicate\\target\\debug
C:\\Users\\natha\\.rustup\\toolchains\\stable-x86_64-pc-windows-msvc\\lib\\rustlib\\x86_64-pc-windows-msvc\\lib
...rest of PATH
```
While the nested cargo invocation should be:
```
C:\\Users\\natha\\OneDrive\\Desktop\\cargo-path-duplicate\\target\\debug\\deps
C:\\Users\\natha\\OneDrive\\Desktop\\cargo-path-duplicate\\target\\debug
C:\\Users\\natha\\.rustup\\toolchains\\stable-x86_64-pc-windows-msvc\\lib\\rustlib\\x86_64-pc-windows-msvc\\lib
C:\\Users\\natha\\OneDrive\\Desktop\\cargo-path-duplicate\\target\\debug\\deps
C:\\Users\\natha\\OneDrive\\Desktop\\cargo-path-duplicate\\target\\debug
C:\\Users\\natha\\.rustup\\toolchains\\stable-x86_64-pc-windows-msvc\\lib\\rustlib\\x86_64-pc-windows-msvc\\lib
...rest of PATH
```
As can be seen, cargo has added the same 3 paths again to the path, making crates that watch the PATH variable for changes be rebuilt.

### Possible Solution(s)

I'm not sure where cargo inserts these paths, but perhaps just avoiding adding a path if the path occurs within the first 3 paths is sufficient? Another solution could be a hashset over all paths in the `PATH`, but that seems excessive.

### Notes

I'm using an [xtask](https://github.com/matklad/cargo-xtask) setup to run `cargo build`, while running clippy through cargo directly. Switching between xtask and cargo causes a decent number of crates to be rebuilt. This is because `cc`, which a few of my crates use, emits a `cargo:rerun-if-env-changed` directive for `PATH`, which is dirty between cargo and xtask. I would expect a `cargo build` run from within xtask and a `cargo build` run outside of xtask to not need to rebuild each other's artifacts.

As an alternative, I could just add a new command to my xtask setup for clippy. However, I would prefer if I didn't have to wrap each cargo build-like command and `cargo install`-ed subcommand that I want to use.

### Version

```text
cargo 1.79.0 (ffa9cf99a 2024-06-03)
release: 1.79.0
commit-hash: ffa9cf99a594e59032757403d4c780b46dc2c43a
commit-date: 2024-06-03
host: x86_64-pc-windows-msvc
libcurl: 8.6.0-DEV (sys:0.4.72+curl-8.6.0 vendored ssl:Schannel)
os: Windows 10.0.22631 (Windows 11 Professional) [64-bit]
```
```

### PR DESCRIPTION
**Pull request by linyihai**

If the current program started by `cargo run` recursively call into `cargo run`, the second `cargo run` will insert  `search_path`  into `dylib_path_envvar` again.

Fixes #14194

The first commit adds the test to reflect the issue. The first call to `cargo run` stores the dylib search path env var to a file. Subsequent calls verify that env var remains the same.

The second commit fixes the behavior by checking if env vars in `search_path` are a prefix of the slice of env vars in `dylib_path_envvar`.

### DIFF
--- a/src/cargo/core/compiler/compilation.rs
+++ b/src/cargo/core/compiler/compilation.rs
@@ -323,7 +323,11 @@ impl<'gctx> Compilation<'gctx> {
 
         let dylib_path = paths::dylib_path();
         let dylib_path_is_empty = dylib_path.is_empty();
-        search_path.extend(dylib_path.into_iter());
+        if dylib_path.starts_with(&search_path) {
+            search_path = dylib_path;
+        } else {
+            search_path.extend(dylib_path.into_iter());
+        }
         if cfg!(target_os = "macos") && dylib_path_is_empty {
             // These are the defaults when DYLD_FALLBACK_LIBRARY_PATH isn't
             // set or set to an empty string. Since Cargo is explicitly setting
--- a/tests/testsuite/build.rs
+++ b/tests/testsuite/build.rs
@@ -2154,6 +2154,66 @@ fn crate_library_path_env_var() {
     setenv_for_removing_empty_component(p.cargo("run")).run();
 }
 
+// See https://github.com/rust-lang/cargo/issues/14194
+#[cargo_test]
+fn issue_14194_deduplicate_library_path_env_var() {
+    let p = project()
+        .file(
+            "src/main.rs",
+            &format!(
+                r#"
+                    use std::process::Command;
+                    fn main() {{
+                        let level: i32 = std::env::args().nth(1).unwrap().parse().unwrap();
+                        let txt = "var.txt";
+                        let lib_path = std::env::var("{}").unwrap();
+
+                        // Make sure we really have something in dylib search path.
+                        let count = std::env::split_paths(&lib_path).count();
+                        assert!(count > 0);
+
+                        if level >= 3 {{
+                            std::fs::write(txt, &lib_path).unwrap();
+                        }} else {{
+                            let prev_lib_path = std::fs::read_to_string(txt).unwrap();
+                            // Ensure no duplicate insertion to dylib search paths
+                            // when calling `cargo run` recursively.
+                            assert_ne!(lib_path, prev_lib_path);
+                        }}
+
+                        if level == 0 {{
+                            return;
+                        }}
+                        
+                        let _  = Command::new(std::env!("CARGO"))
+                        .arg("run")
+                        .arg("--")
+                        .arg((level - 1).to_string())
+                        .status()
+                        .unwrap();
+                    }}
+                "#,
+                dylib_path_envvar(),
+            ),
+        )
+        .build();
+
+    setenv_for_removing_empty_component(p.cargo("run -- 3"))
+        .with_stderr_data(str![[r#"
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+[RUNNING] `target/debug/foo[EXE] 3`
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+[RUNNING] `target/debug/foo[EXE] 2`
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+[RUNNING] `target/debug/foo[EXE] 1`
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+[RUNNING] `target/debug/foo[EXE] 0`
+
+"#]])
+        .run();
+}
+
 // Regression test for #4277
 #[cargo_test]
 fn build_with_fake_libc_not_loading() {
--- a/tests/testsuite/build.rs
+++ b/tests/testsuite/build.rs
@@ -2178,7 +2178,7 @@ fn issue_14194_deduplicate_library_path_env_var() {
                             let prev_lib_path = std::fs::read_to_string(txt).unwrap();
                             // Ensure no duplicate insertion to dylib search paths
                             // when calling `cargo run` recursively.
-                            assert_ne!(lib_path, prev_lib_path);
+                            assert_eq!(lib_path, prev_lib_path);
                         }}
 
                         if level == 0 {{

### CODE REVIEW

File: tests/testsuite/build.rs [hunk 1]
```diff
@@ -2147,6 +2147,40 @@ fn crate_library_path_env_var() {
...
+                r#"
+                    use std::process::Command;
+                    fn main() {{
+                        if std::env::var("SUB").is_ok() {{
+                            let paths = match std::env::var_os("{}") {{
+                                Some(var) => std::env::split_paths(&var).collect(),
+                                None => Vec::new(),
+                            }};
+                            let first = paths.first().unwrap();
+                            assert!(paths.iter().skip(1).any(|p| p == first ));
```
epage: Should we test for the sympton rather than the specific detail?  In this case, its an extra rebuild

src/cargo/core/compiler/compilation.rs
```diff
@@ -323,7 +323,12 @@ impl<'gctx> Compilation<'gctx> {
 
         let dylib_path = paths::dylib_path();
         let dylib_path_is_empty = dylib_path.is_empty();
-        search_path.extend(dylib_path.into_iter());
+        // if current program call into `cargo`, we need to avoid inserting `search_path` before `dylib_path` since
+        // `dylib_path` was already prefixed with `search_path`.
+        if &dylib_path.first() != &search_path.first() {
```
epage, line 328: Why is only comparing `first()` sufficient?

linyihai:
This issue seems need more exploration, thanks for your in advance review.

src/cargo/core/compiler/compilation.rs
weihanglo, line 1: The reason this approach didn't work is pretty much due to <https://github.com/rust-lang/cargo/issues/10358>. When Cargo computes fingerprint for a `Unit` of work, [it asks `env::var`](https://github.com/rust-lang/cargo/blob/40b6638df6ee1360680a7577168eb1be4a167a83/src/cargo/core/compiler/fingerprint/mod.rs#L774-L779) which doesn't contain `LIBRARY_PATH` or `PATH` because that is set by Cargo. `rerun-if-env-changed` only captures environment variables set by users via shell like `MY_VAR=1 cargo build`. Variables added by Cargo only available when consulting the `Command` struct of the build script execution via `Command::get_envs` (like what has done in f91dc782245cc5fcf75f9acc191ee9b018605328).



As a consequence, if you open the fingerprint JSON file of the build script execution (normally at `target/debug/.fingerprint/foo-559e30894cda9925/run-build-script-build-script-build.json`), you'll see

```json5
{
  "rustc": 11851488446658826000,
  // ...
  "local": [
    {
      "RerunIfEnvChanged": {
        "var": "DYLD_FALLBACK_LIBRARY_PATH",
        "val": null
      }
    }
  ],
  "rustflags": [],
  // ...
}
```

which shows that Cargo failed to capture the initial library search path set by itself.

---

There is another way to observe this bug: set the `LEVEL` env var in the test to a higher number like 5. Any subsequent Cargo invocation after the first one should be fresh. Before doing this verification you need to change `std::env!("LEVEL")` to `std::env::var!("LEVEL")` because the former will be captured by rustc dep-info, which trigger a rebuild.

tests/testsuite/build.rs
```diff
@@ -2147,6 +2147,90 @@ fn crate_library_path_env_var() {
     setenv_for_removing_empty_component(p.cargo("run")).run();
 }
 
+// See https://github.com/rust-lang/cargo/issues/14194
+#[cargo_test]
+fn issue_14194_deduplicate_library_path_env_var() {
```
weihanglo, line 2152: Should we re-target this PR to fix the unnecessary-env-var-prepend issue?

So that the test might be rewritten like this:

* A `src/main.rs` that calling itself.
* The first, the second call contains the same set of environment variables.
* No `build.rs` needed.

src/cargo/core/compiler/compilation.rs
linyihai: > Before doing this verification you need to change std::env!("LEVEL") to std::env::var!("LEVEL") because the former will be captured by rustc dep-info, which trigger a rebuild.

I didn't expect such a difference. 

I decide to to get the `"LEVEL"` from the command line argument.

tests/testsuite/build.rs
```diff
@@ -2154,6 +2154,57 @@ fn crate_library_path_env_var() {
...
+            "src/main.rs",
+            &format!(
+                r#"
+                    use std::process::Command;
+                    fn main() {{
+                        let args: Vec<String> = std::env::args().collect();
+                        if args.len() == 1 {{
+                            return;
+                        }}
+                        let level: i32  = args[1].parse().unwrap();
```
weihanglo, line 2171: This is probably better?

```suggestion
                        let level: i32 = std::env::args().nth(1).unwrap().parse().unwrap();
```

tests/testsuite/build.rs
```diff
@@ -2154,6 +2154,56 @@ fn crate_library_path_env_var() {
...
+                        .status()
+                        .unwrap();
+                    }}
+                "#,
+                dylib_path_envvar(),
+            ),
+        )
+        .build();
+
+    setenv_for_removing_empty_component(p.cargo("run -- 1"))
```
weihanglo, line 2195: I would suggest 3, so it can assure env var stay the same for more than one recursive call.

```suggestion
    setenv_for_removing_empty_component(p.cargo("run -- 3"))
```

tests/testsuite/build.rs
weihanglo, line 1: What about something like

```rust
                        let level: i32 = std::env::args().nth(1).unwrap().parse().unwrap();
                        let txt = "var.txt";
                        let lib_path = std::env::var("{}").unwrap();

                        // Make sure we really have something in dylib search path.
                        let count = std::env::split_paths(&lib_path).count();
                        assert!(count > 0);

                        if level >= 3 {{
                            std::fs::write(txt, &lib_path).unwrap();
                        }} else {{
                            let prev_lib_path = std::fs::read_to_string(txt).unwrap();
                            // Ensure no duplicate insertion to dylib search paths
                            // when calling `cargo run` recursively.
                            assert_eq!(lib_path, prev_lib_path);
                        }}

                        if level == 0 {{
                            return;
                        }}
						/// Command::new(...)
```

So that we really check every run dylib search paths stay the same?

weihanglo (approved):

PR #14467, opened 2024-08-28
fix(pkgid): Allow open namespaces in PackageIdSpec's

epage:

This is a part of #13576

This unblocks #14433.  We have a test to ensure you can't publish a namespaced package and the error for that is getting masked in #14433 because the package name is getting  parsed as a `PackageIdSpec` which wasn't supported until this PR.

### DIFF
--- a/crates/cargo-util-schemas/src/core/package_id_spec.rs
+++ b/crates/cargo-util-schemas/src/core/package_id_spec.rs
@@ -210,7 +210,10 @@ impl PackageIdSpec {
 }
 
 fn parse_spec(spec: &str) -> Result<Option<(String, Option<PartialVersion>)>> {
-    let Some((name, ver)) = spec.split_once([':', '@']) else {
+    let Some((name, ver)) = spec
+        .rsplit_once('@')
+        .or_else(|| spec.rsplit_once(':').filter(|(n, _)| !n.ends_with(':')))
+    else {
         return Ok(None);
     };
     let name = name.to_owned();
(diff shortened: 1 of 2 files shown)


crates/cargo-util-schemas/src/core/package_id_spec.rs
```diff
@@ -217,6 +209,18 @@ impl PackageIdSpec {
     }
 }
 
+fn parse_spec(spec: &str) -> Result<Option<(String, Option<PartialVersion>)>> {
```
jneem, line 212: I think the second `Option` is redundant, because if the version is missing then you return `Ok(None)`

epage: Yeah, I should have called out this non-obvious design choice.

Yes, we always return `Some(version)`.   However, that makes things nicer in the caller.  One caller doesn't care, they can just `Some(version)` or pass it through, depending on what we do.  The other caller would require a `.map(|(name, version)| (name, Some(version))` which when added to `let (name, version) = parse_spec(spec)?.unwrap_or_else(|| (spec.to_owned(), None));` was making things more complicated than it seemed worth it (or I'd break this out into a `match`).  Since the scope of `parse_spec` is hyper-local, it felt okay-ish to consider how the callers used it for what the API looked like.  I already had to do that once by the fact that each caller handles the lack of version differently.

Muscraft (approved):
Thanks!

PR #14471, opened 2024-08-30
feat(resolve): Report MSRV compatible version instead of incomptible 

epage:

This expands on #14461 to where only MSRV-compatible versions are
"actionable".  MSRV-incompatible versions are therefore unstyled.

We report the MSRV needed so people can choose to unblock by updating
their MSRV.  I had wondered if we should report the the absolute latest
MSRV-incompatible version or the one with the next higher MSRV from
where the user is at.  Both are reasonable use cases, so I erred with
absolute latest version.

```console
$ cargo update --workspace -v
```
image

I changed the label from `latest` to `available` to not have to keep coming up with relevant terms for each case.

### Additional information

This is the final step before asking for new feedback on #13908

### DIFF
--- a/src/cargo/ops/cargo_update.rs
+++ b/src/cargo/ops/cargo_update.rs
@@ -757,9 +757,20 @@ fn report_latest(possibilities: &[IndexSummary], change: &PackageChange) -> Opti
     }
 
     let version_req = package_id.version().to_caret_req();
+    let required_rust_version = change.required_rust_version.as_ref();
+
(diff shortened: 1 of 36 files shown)


ehuss:
Thanks!

I was wondering if we should maybe have some documentation somewhere that directly discusses the situations where cargo is unable to update to the latest version. Possibly with some illustrations that show these different kinds of `(available...)` messages.

@bors r+

PR #14473, opened 2024-08-31
Add location information in patch error messages 

### LINKED ISSUE #13952: Location of patch information
### Problem

If there are errors in a `[patch]` entry, it can be hard to locate _where the patch was specified_.

I work on a large code base that spans multiple workspace crates and cross repo crates.
I have experienced a few times that we migrate a set of dependencies, and get caught with resolution errors. If the error is because of a `[patch]` entry which is no longer applicable, it only gives the crate name that is failing.
Most of the time we define patches in the workspace `Cargo.toml`, however there have been occasion to put them into the `.cargo/config.toml` for various reasons.

The `.cargo/config.toml` is not a common location to look for patches, it would be helpful if the error message could point to where the patch is being specified.

### Proposed Solution

Cargo gives more detail for error messages involving `[patch]`es. Specifically the location of the `[patch]` entry would quickly help pinpoint where to fix the issue.

### Notes

_No response_

antoniospg:

Fixes #13952
* Similar to what was explained in the issue page, I've tried to preserve location information in `Workspace::root_patch` and use it in `registry::patch` to improve the error messages.
* I've also updated the tests to take this change in consideration.


tests/testsuite/patch.rs
```diff
@@ -1437,7 +1437,7 @@ fn replace_with_crates_io() {
         .with_status(101)
         .with_stderr_data(str![[r#"
 [UPDATING] `dummy-registry` index
-[ERROR] failed to resolve patches for `https://github.com/rust-lang/crates.io-index`
+[ERROR] failed to resolve patches for `https://github.com/rust-lang/crates.io-index` specified in `[ROOT]/foo/Cargo.toml`
```
epage, line 1440: Please make all commits atomic which includes all tests must pass for each commit

src/cargo/core/resolver/encode.rs
```diff
@@ -465,7 +465,7 @@ fn build_path_deps(
         build_pkg(member, ws, &mut ret, &mut visited);
     }
     for deps in ws.root_patch()?.values() {
-        for dep in deps {
+        for dep in deps.iter().map(|(d, _)| d) {
             build_dep(dep, ws, &mut ret, &mut visited);
         }
```
epage, line 470: I feel like `for (dep, _) in deps` would be cleaner.

src/cargo/core/workspace.rs
```diff
@@ -506,13 +528,43 @@ impl<'gctx> Workspace<'gctx> {
...
+        let mut from_manifest: HashMap<_, Vec<_>> = HashMap::new();
+        for (url, dep) in patch_from_manifest {
+            from_manifest.insert(
+                url.clone(),
+                dep.iter()
+                    .map(|dep| (dep.clone(), manifest_path.clone()))
+                    .collect(),
+            );
+        }
+
```
epage, line 567: This was
1. manifest
2. config
3. config
4. manifest

But became
1. config
2. manifest
5. manifest
6. config
7. manifest
8. config
9. manifest

Already the original code wasn't idea, it should either be completely parallel or consistently interleaved.  As the code grows, I would suggest interleaving makes the code harder to follow and we should be completely parallel.

src/cargo/ops/resolve.rs
```diff
@@ -813,7 +813,7 @@ fn register_patch_entries(
 ) -> CargoResult<HashSet<PackageId>> {
     let mut avoid_patch_ids = HashSet::new();
     for (url, patches) in ws.root_patch()?.iter() {
-        for patch in patches {
+        for patch in patches.iter().map(|(d, _)| d) {
```
epage, line 816: I feel like `for (patch, _) in patches` would be cleaner.

src/cargo/ops/resolve.rs
```diff
@@ -832,7 +832,8 @@ fn register_patch_entries(
         // previous resolve graph, which is primarily what's done here to
         // build the `registrations` list.
         let mut registrations = Vec::new();
-        for dep in patches {
+        for dep_with_location in patches {
+            let (dep, _) = dep_with_location;
```
epage, line 836: I think I'd prefer destructuring the tuple in the `for` and then re-creating it in the `push`

src/cargo/core/workspace.rs
```diff
@@ -506,13 +528,43 @@ impl<'gctx> Workspace<'gctx> {
...
+        let mut from_manifest: HashMap<_, Vec<_>> = HashMap::new();
+        for (url, dep) in patch_from_manifest {
+            from_manifest.insert(
+                url.clone(),
+                dep.iter()
+                    .map(|dep| (dep.clone(), manifest_path.clone()))
+                    .collect(),
+            );
+        }
```
epage, line 566: Should this be done in the functions we call to get a patch?  Please understand, this is a serious question and not an indirect way of saying what to do.

For manifests, its a little weird because it will always be the same path.  However, it feels weird to be calling back into config to get the location.  It feels like we are re-implementing things.  Calling back into config can also have some negative performance implications.

src/cargo/core/workspace.rs
```diff
@@ -506,13 +528,43 @@ impl<'gctx> Workspace<'gctx> {
     /// Returns the root `[patch]` section of this workspace.
     ///
     /// This may be from a virtual crate or an actual crate.
-    pub fn root_patch(&self) -> CargoResult<HashMap<Url, Vec<Dependency>>> {
-        let from_manifest = match self.root_maybe() {
+    pub fn root_patch(&self) -> CargoResult<HashMap<Url, Vec<(Dependency, PathOrDefinition)>>> {
```
epage, line 531: I'm a bit mixed on tuple vs struct.

antoniospg: We could have a struct like:

```
#[derive(Clone, Debug)]
pub struct Patch {
    dep: Dependency,
    location: PathOrDefinition,
}
```
and have the `PathOrDefinition::get_location_string` implemented for it instead. But I'm not sure what would be the other benefits, as we would still need to destructure it in a lot of places or keep using `iter().map(|patch| patch.dep)` instead. But I'm open for better ideas.

src/cargo/core/workspace.rs
antoniospg: Not sure if this is what you said, but maybe it's better to move the` get_cv_with_env `call to  `Workspace::config_patch` as it's already calling the config there with the following: 
```
        let config_patch: Option<
            BTreeMap<String, BTreeMap<String, TomlDependency<ConfigRelativePath>>>,
        > = self.gctx.get("patch")?;
```
But it would still have the negative performance implications. I'm not sure if there's a better way to get the config location.

Issue #12883, opened 2023-10-27
Publish `.cargo/config.toml` jsonschema to `schemastore.org` for editor 


loynoir:
### Problem

`.cargo/config.toml` does not have jsonshema.

So editor can not use schemastore to support tab-completion and validation.

### Proposed Solution

`cargo` team implement jsonschema, publish and maintain related jsonschema in https://github.com/SchemaStore/schemastore.

So editor can use schemastore to support tab-completion and validation.

### Notes

- doc

https://doc.rust-lang.org/cargo/reference/config.html

- `vcpkg`

https://github.com/microsoft/vcpkg/issues/34706

linyihai:
In my use-cases, i use vscode with the plugins `Even Better TOML` and `crates`, it is support completion and validation. In my opinion, the most of  Rust configuration written in toml format, may be it is not a easy-cake to support another format.

loynoir:
Although its name is `jsonschema`, but actually, it can also provide for `yaml` and `toml`

https://github.com/SchemaStore/schemastore/blob/24a1f6e5e134d1f2dddf9d781fb5717a308c2822/src/api/json/catalog.json#L19

epage:
We could have a `feature` to enable https://crates.io/crates/schemars/0.8.15 on our types with a xtask to update the file checked in.  There would be bugs in it but it can serve as a basis for community members to then update the schemastore based off of it.

weihanglo:
We will need to pull each fields in `.cargo/config.toml` out as struct definitions to a module. Not all of those fields have a struct defined, and their access scatters over [cargo/util/context/mod.rs](https://github.com/rust-lang/cargo/blob/1ea02ffc5ee2b4e3584871f8bd95717e1456ec3d/src/cargo/util/context/mod.rs#L225-L230) and anywhere in the codebase. This might need some more explorations and efforts involved.

For `Cargo.toml`, although there is already a JSON schema as [aforementioned](https://github.com/rust-lang/cargo/issues/12883#issuecomment-1782854956), I think we can still experiment on it first to understand what kind of edge cases `schemars` has. It is also simpler because `cargo-util-schemas` should already have all struct definitions of `Cargo.toml`.

[...]

PR #14482, opened 2024-09-01
add cargo config jsonschema support for editor auto completion and docs jumping

loynoir:

There is not yet editor hint and completion support for `.cargo/config.toml`.

So I impl it.

Related

- https://github.com/rust-lang/cargo/issues/12883

- https://github.com/SchemaStore/schemastore/issues/3349

### Goals

[x] vscode tab completion on key

[x] vscode tab completion on value candidates

[x] vscode err hint when invalid key

[x] vscode hint on document url

Review `schema/cargo-config.schema.d.mts`.

The jsonschema is generate from typescript using `npm:typescript-json-schema@0.65.1`.

Try genereated jsonschema using any supported editor toml extension.

```jsonc
        "evenBetterToml.schema.associations": {
            "^.*/\\.cargo/config\\.toml$": "https://raw.githubusercontent.com/loynoir/cargo/feat-json-schema-for-editor/schema/cargo-config.schema.json",
```

### Additional information

As result, you can see the output jsonschema is very large.

Not able to let human write it.

So I generate it from typescript dts.

Maybe someone can impl macro rules to generate jsonschema, website docs and cli parser at the same time in the future.


epage:
Generally we use cargo xtasks for automating things in the repo with a CI check to ensure that the generator and the output haven't drifted (make the generator was run, avoid hand-edits).  This also duplicates information in an out-of-the-way place (currently in source and in our own documentation).

Also, I'm unsure what the implications are of us having incomplete or wrong information for targets since a straight combination of `arch`, `sub`, `vendor`, `sys`, and `abi` are being used.  We might want to just leave that as a string.

I wonder if we could use [schemars](https://crates.io/crates/schemars) or something like would be a good way to single-source this information.  Maybe then we could also code-generate some of our documentation from it as well.

If you go down this route, I'd recommend playing with this for `Cargo.toml` first in `crates/cargo-util-schemas`.  We could have an `unstable-schemagen` feature and the xtask could enable it and get the schema and dump it.  If/when we get that settled, we could work to move the config schemas to `cargo-util-schemas` and under `unstable-schemagen` have a top-level `TomlConfig` struct.

PR #14487, opened 2024-09-02
Fix elided lifetime

ehuss:
This fixes an issue with the recent nightly that has added a lint (https://github.com/rust-lang/rust/pull/129207) that warns about the lack of a lifetime, which looks like:

```
warning: elided lifetime has a name
   --> src/cargo/core/workspace.rs:580:66
    |
580 |     pub fn default_members<'a>(&'a self) -> impl Iterator<Item = &Package> {
    |                            -- lifetime `'a` declared here        ^ this elided lifetime gets resolved as `'a`
    |
    = note: `#[warn(elided_named_lifetimes)]` on by default
```

### DIFF
--- a/src/cargo/core/workspace.rs
+++ b/src/cargo/core/workspace.rs
@@ -577,7 +577,7 @@ impl<'gctx> Workspace<'gctx> {
     }
 
     /// Returns an iterator over default packages in this workspace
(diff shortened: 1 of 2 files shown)


ehuss:
Pushed a fix for the test matching the nightly version.

PR #14493, opened 2024-09-04
feat: Switch hand-written completions to native completions

shannmu:

Related issue https://github.com/rust-lang/cargo/issues/6645
This PR is the first step to move cargo shell completions to native completions by using `clap_complete` crate. It makes users could complete cargo subcommand and flags.
See https://github.com/clap-rs/clap/issues/3166 to get more context about what features `clap_complete` has supported.

1. Build a test environment, including the necessary short completion scripts, and the `complete` function to start an interactive shell with the help of a pty device and obtain completion results.
2. Simply test the completion results of subcommands and flags in bash and zsh.


Cargo.toml
```diff
@@ -205,6 +205,8 @@ unicode-width.workspace = true
 url.workspace = true
 walkdir.workspace = true
 supports-unicode = "3.0.0"
+completest = "0.4.2"
+completest-pty = "0.5.3"
```
epage, line 209: `completest-pty` re-exports `completest` so we only need one dep which will avoid needing to keep these in sync

epage: These also should be defined in `workspace.dependencies` and sorted

shannmu: Thanks for your reminder!

epage: And pulled into `dev-dependencies`, rather than `dependencies`

src/bin/cargo/main.rs
```diff
@@ -28,6 +29,13 @@ fn main() {
         }
     };
 
+    let nightly_features_allowed = matches!(&*features::channel(), "nightly" | "dev");
+    if nightly_features_allowed {
+        clap_complete::CompleteEnv::with_factory(|| cli::cli(&mut gctx))
+            .var("CARGO_COMPLETE")
```
epage, line 35: We need this documented in https://doc.rust-lang.org/cargo/reference/unstable.html

src/bin/cargo/main.rs
```diff
@@ -28,6 +29,13 @@ fn main() {
         }
     };
 
+    let nightly_features_allowed = matches!(&*features::channel(), "nightly" | "dev");
+    if nightly_features_allowed {
+        clap_complete::CompleteEnv::with_factory(|| cli::cli(&mut gctx))
```
epage, line 34: This is using the built-in shell support, expanding our shell coverage from bash and zsh to also include elvish, fish, and powershell.

Do we want to do that?  Even if we experiment with this, we would need to track that in the PR description and the tracking issue.

tests/testsuite/auto_complete/basic/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+use snapbox::assert_data_eq;
+
+#[cargo_test]
+fn bash() {
+    let project = Project::from_template(current_dir!().join("in"));
+    let project_root = project.root();
+    let cwd = &project_root;
+
+    // Change current directory to the project root
+    std::env::set_current_dir(cwd).unwrap();
```
epage, line 13: This affects all tests and shouldn't be done

tests/testsuite/auto_complete/basic/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+--color            -Z                 fix                new                rustdoc            
+--locked           -h                 generate-lockfile  owner              search             
+--offline          add                help               package            test               
+--frozen           bench              info               pkgid              tree               
+--config           build              init               publish            uninstall          "
+    ];
+    let actual = super::common::complete(input, "bash");
+    assert_data_eq!(actual, expected);
+
+    let input = "cargo build --\t\t";
```
epage, line 32: We probably only need one per shell.  Our focus for these tests should be on the completions being registered.  For everything else, we should either directly test custom completers or rely on `clap_complete`s tests

epage: In that situation, do we need to have this setup for many `.rs` files?

tests/testsuite/auto_complete/common.rs
```diff
@@ -0,0 +1,68 @@
...
+    let shell_name = R::name();
+
+    let home = std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+        .join("tests/testsuite/auto_complete/snapshots")
+        .join(shell_name);
+
+    let scratch = snapbox::dir::DirRoot::mutable_temp()
+        .unwrap()
+        .with_template(&home)
+        .unwrap();
```
epage, line 32: Could we instead use the callers home directory?

tests/testsuite/auto_complete/common.rs
```diff
@@ -0,0 +1,68 @@
+pub(crate) fn complete(input: &str, shell: impl Into<Shell>) -> String {
+    let shell = shell.into();
+
+    // Load the runtime
+    let (mut runtime, _scratch) = match shell {
+        Shell::Bash => load_runtime::<completest_pty::BashRuntimeBuilder>(),
+        Shell::Zsh => load_runtime::<completest_pty::ZshRuntimeBuilder>(),
+    };
+
```
epage, line 9: This feels roundabout to go from `&str` -> `Shell` -> `completest_pty::*RuntimeBuilder`

tests/testsuite/auto_complete/common.rs
```diff
@@ -0,0 +1,68 @@
...
+    let home = scratch.path().unwrap().to_owned();
+
+    let bin_path = cargo_test_support::cargo_exe();
+    let bin_root = bin_path.parent().unwrap().to_owned();
+
+    let runtime = Box::new(R::with_home(bin_root, home).unwrap());
+
+    (runtime, scratch)
+}
```
epage, line 42: This is reading in a checked-in `HOME` but nothing makes sure that the checked-in `HOME` is up-to-date.

Also, do we need to check-in `HOME`?

tests/testsuite/auto_complete/common.rs
epage, line 32: Could we copy it into the caller's `HOME`?

tests/testsuite/auto_complete/common.rs
epage: See https://doc.rust-lang.org/nightly/nightly-rustc/cargo_test_support/paths/fn.home.html

tests/testsuite/auto_complete/common.rs
epage: See https://doc.rust-lang.org/nightly/nightly-rustc/cargo_test_support/paths/fn.home.html

tests/testsuite/main.rs
```diff
@@ -6,6 +6,7 @@ mod advanced_env;
 mod alt_registry;
 mod artifact_dep;
 mod artifact_dir;
+mod auto_complete;
```
epage, line 9: I think `shell_completions` might be a better name

src/bin/cargo/main.rs
epage: > By using clap_complete crate, we could extend the supported shells to Bash, Zsh, Elvish, Fish, and PowerShell. However, at the current stage, the support for PowerShell in clap_complete is not fully developed, so this PR does not include any content related to PowerShell.

Thanks for adding this to the PR. I've added a similar Open Questioln to #14520.

Note: currently, you are supporting `powershell`, you just might not be documenting it.

tests/testsuite/shell_completions/mod.rs
```diff
@@ -1,2 +1,6 @@
 pub mod common;
 
+mod bash;
+mod elvish;
+mod fish;
+mod zsh;
```
epage, line 6: Do we need these to be separate files?  We do not need to exhaustively test completions.  I'm not even sure we need to test every shell.  Our primary aim is to test that the `CompleteEnv` call is in place and works as expected.

For shell-particulars or native completion particulars, `clap_complete`s tests are covering us.  For our custom completers, we should probably skip end-to-end tests for those and do integration tests.

src/bin/cargo/main.rs
epage: Also, I'm fine with us doing powershell for now, we just need to evaluate whether to stabilize it later.

Exposing it now gets more visibility for more testing and feedback rather than waiting for some point for it to be ready enough.

tests/testsuite/shell_completions/mod.rs
```diff
@@ -1,2 +1,3 @@
 pub mod common;
 
+mod register;
```
epage, line 3: Please move all of this into `tests/testsuite/shell_completions.rs`

epage:
Also, it looks like you need to `git rebase` against the latest master; there is a merge conflict.

src/bin/cargo/main.rs
shannmu: If we support `powershell` for now. I need to add test for `powershell` and first add PowerShell support to the completest crate (continuing my previous work).

epage: Having it validated *somehow* is a blocker for stabilization.  Note that none of our existing completions are validated in an automated way, so automating it is not necessarily a requirement.

tests/testsuite/shell_completions.rs
```diff
@@ -0,0 +1,183 @@
+use cargo_test_support::cargo_test;
+use completest_pty::Runtime;
+use snapbox::assert_data_eq;
+
+#[cargo_test]
+fn bash() {
```
epage, line 6: These tests need `#[cfg(unix)]`

epage:
Looks like there is a doc and a test failure (besides windows)

tests/testsuite/shell_completions.rs
```diff
@@ -0,0 +1,56 @@
...
+    let shell = shell.into();
+
+    // Load the runtime
+    let mut runtime = match shell {
+        "bash" => load_runtime::<completest_pty::BashRuntimeBuilder>("bash"),
+        "elvish" => load_runtime::<completest_pty::ElvishRuntimeBuilder>("elvish"),
+        "fish" => load_runtime::<completest_pty::FishRuntimeBuilder>("fish"),
+        "zsh" => load_runtime::<completest_pty::ZshRuntimeBuilder>("zsh"),
+        _ => panic!("Unsupported shell: {}", shell),
+    };
```
epage, line 145: Any code using `completest_pty` needs a `#[cfg(unix)]`

shannmu: I saw it in the CI, thanks for your reminder!

src/bin/cargo/main.rs
shannmu: I added instructions on how to use the feature in PowerShell to `unstable.md`. Call for everybody to try this new feature.

As for testing, I haven't added tests for PowerShell yet. I will add tests once PowerShell support in the `clap_complete` crate stabilizes.

shannmu:
On macOS, the test code for bash, fish, and elvish cannot pass. The default shell in macOS CI seems to be zsh, and only zsh works correctly. So I added `cfg!(target_os = "macos")` branch for these tests.

ehuss:
Hi @shannmu! There seems to be a problem with this PR that I can't quite figure out. The `shell_completions::fish` test seems to be failing. See [this CI run](https://github.com/rust-lang/cargo/actions/runs/10841889343/job/30086629807) for example. The error looks like it is only emitting `cargo ∅`.

I am able to reproduce on my system (I am unable to get it to pass at all). I can't tell how this PR was able to pass CI, so maybe there is some flakiness with it?  Do you think you can investigate?

And FYI, cargo's CI is currently broken for other reasons (see https://github.com/rust-lang/rust/issues/130291), but that only affects a subset of CI jobs.

PR #14503, opened 2024-09-06
Bump ci's version of cargo-semver-version

jneem:
I just started seeing CI failures like this:
```
error: rustdoc format v33 for file /home/jneeman/tweag/cargo/target/semver-checks/local-cargo-0_83_0/target/semver-checks/target/doc/cargo.json is not supported
```
I think the issue is that the new rust version needs a new cargo-semver-checks, so this PR bumps the version of cargo-semver-checks in CI.


weihanglo:
Thanks!

@bors r+

PR #14504, opened 2024-09-06
Include public/private dependency status in `cargo metadata`

### LINKED ISSUE #14502: Include public/private dependency status in `cargo metadata`
### Problem

related to https://github.com/rust-lang/rust/issues/44663

Information about whether or not a dependency is public is not included in the output of `cargo metadata`.

This could be useful for custom subcommands. `cargo override`, would find this useful for https://github.com/eopb/cargo-override/issues/121

### Proposed Solution

When `-Zpublic-dependency` is enabled, include a boolean flag in `cargo metadata`s output at `packages[].dependencies[].public`.
If  `-Zpublic-dependency` is not enabled,  this flag should not appear in `cargo metadata` output.

### Notes

If this feature request is accepted, I'm happy to claim and take a crack at working on it

eopb:
fixes #14502

> [!TIP]
> This change can be reviewed commit-by-commit.
> Descriptions on individual commits are available to justify some decisions

@rustbot label Command-metadata Z-public-dependency


eopb:
This is a breaking change to the `cargo` crate, but I'm not sure which `Cargo.toml` version I should bump, if any.

epage:
> This is a breaking change to the cargo crate, but I'm not sure which Cargo.toml version I should bump, if any.

Our general policy is that every new `cargo` release is a breaking change, so you don't have to worry about that.  For any other crate in the workspace, yes, you'd have to bump it.

tests/testsuite/metadata.rs
```diff
@@ -627,6 +627,326 @@ fn cargo_metadata_with_deps_and_version() {
         .run();
 }
 
+#[cargo_test]
+fn cargo_metadata_public_private_dependencies() {
```
epage, line 631: Would you be up for splitting your commit into
- Add test (passes on this commit)
- Make behavior change (will require updating the test so it still passes)

This both makes it easier to review and easier to demonstrate to others what the new feature does.  You just look at the feature commit and see how the tests have changed.

tests/testsuite/metadata.rs
```diff
@@ -627,6 +627,326 @@ fn cargo_metadata_with_deps_and_version() {
...
+          "source": "registry+https://github.com/rust-lang/crates.io-index",
+          "target": null,
+          "uses_default_features": true
+        },
+        {
+          "features": [],
+          "kind": null,
+          "name": "foobar",
+          "optional": false,
+          "public": true,
```
epage, line 783: https://doc.rust-lang.org/cargo/commands/cargo-metadata.html needs to be updated with this, with an `(unstable)` in the description

src/cargo/core/dependency.rs
```diff
@@ -182,6 +164,26 @@ impl Dependency {
         }
     }
 
+    pub fn serialized(&self, unstable_flags: &CliUnstable) -> SerializedDependency {
```
epage, line 167: Please split this change off into its own refactor commit

src/cargo/core/dependency.rs
```diff
@@ -52,50 +52,32 @@ struct Inner {
 }
 
 #[derive(Serialize)]
-struct SerializedDependency<'a> {
-    name: &'a str,
+pub struct SerializedDependency {
+    name: InternedString,
```
epage, line 56: Please split this change off into its own refactor commit

src/cargo/core/dependency.rs
```diff
@@ -182,6 +164,26 @@ impl Dependency {
...
+            uses_default_features: self.uses_default_features(),
+            features: self.features().to_vec(),
+            target: self.inner.platform.clone(),
+            rename: self.explicit_name_in_toml(),
+            registry: self.registry_id().as_ref().map(|sid| sid.url().to_string()),
+            path: self.source_id().local_path(),
+            artifact: self.inner.artifact.clone(),
+            public: unstable_flags
+                .public_dependency
+                .then_some(self.inner.public),
```
epage, line 183: When parsing a manifest, we consider `-Z` or `cargo-features` (`Feature::public_dependency()`).  Should we do that here as well?

epage:

src/cargo/core/dependency.rs
eopb: I'm not sure what considering `Feature::public_dependency()` would look like. Do you have a link to where this happens in the code for parsing the manifest?

tests/testsuite/metadata.rs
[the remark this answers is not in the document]
eopb: I added a second test in https://github.com/rust-lang/cargo/pull/14504/commits/9b5981e1ab8b73378609c217c2ced2b6b7a483e5 but I'm not sure it's very helpful. I can't change the public/private status of dependencies on stable so I'm only testing that `"public": false` is not present, which should already be covered by already existing tests.

tests/testsuite/metadata.rs
eopb: Resolved in https://github.com/rust-lang/cargo/pull/14504/commits/5d9f554aada95e94076291fc6a4513d1c19df360

tests/testsuite/metadata.rs
eopb: I like this :+1:.
I've split everything into multiple commits

src/cargo/core/dependency.rs
epage: https://github.com/rust-lang/cargo/blob/64c4b6d320b09de9fc5085dafd99c020f7128821/src/cargo/util/toml/mod.rs#L789-L794 is us considering it during manifest parsing.  The information is in one of the related types (manifest? package?).  Unsure if you have access to it all to pass it where its needed.

tests/testsuite/metadata.rs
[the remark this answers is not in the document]
epage: > I can't change the public/private status of dependencies on stable so I'm only testing that "public": false is not present, which should already be covered by already existing tests.

Hmm, I wonder if thats a bug.

If you masquerade as nightly but do not enable the public dependency feature, then `public` is allowed but ignored.  We can use that case for testing.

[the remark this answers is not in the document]
epage: Posted #14507

tests/testsuite/metadata.rs
```diff
@@ -627,6 +627,497 @@ fn cargo_metadata_with_deps_and_version() {
...
+          "edition": "2015",
+          "kind": [
+            "lib"
+          ],
+          "name": "bar",
+          "src_path": "[ROOT]/home/.cargo/registry/src/-[HASH]/bar-0.0.1/src/lib.rs",
+          "test": true
+        }
+      ],
+      "version": "0.0.1"
```
epage, line 878: Could you redact the fields that are not relevant for these tests?  I'd recommend keeping all keys in a `dependencies` table but most package metadata keys and the `resolve` graph could probably go

#### Examples

Redact a value
```json
  "workspace_members": "{...}",
```

Redact all other keys in a table
```json
"...": "{...}"
```

src/cargo/core/dependency.rs
```diff
@@ -52,52 +52,28 @@ struct Inner {
 }
 
 #[derive(Serialize)]
-struct SerializedDependency<'a> {
-    name: &'a str,
+pub struct SerializedDependency {
+    name: InternedString,
```
epage, line 56: > I replaced borrowed types in `SerializedDependency` with owned variants to satisfy the borrow checker.
This matches `SerializedPackage` and shouldn't be an issue since `Dependency` is cheap to copy.

Could you help me understand why this didn't borrow check?

One of the goals we have is to eventually migrate all serde types over to `cargo-util-schemas` so people can load the data themselves.  The use of `InternedString` gets in the way of that (granted, we might need to switch to `String` for that...)

eopb: I had issues with temporaries being dropped in `Dependency::serialized` after copying the code from `Dependency::serialize`. Specificity with [`name`](https://github.com/rust-lang/cargo/blob/dcb4ef9a10630df4fb8b1b2e67072b1267d7d0e3/src/cargo/core/dependency.rs#L84) and [`registry_id`](https://github.com/rust-lang/cargo/blob/dcb4ef9a10630df4fb8b1b2e67072b1267d7d0e3/src/cargo/core/dependency.rs#L82). Looking again, it should be possible to work around this issue.

That being said, I don't think it would be worth changing back. Since the refactor makes `SerializedPackage` contain a `SerializedDependency` rather than a `Dependency` and `SerializedPackage` is an owned type, it would probably be awkward to make `SerializedDependency` require a lifetime, since it would require changes in `SerializedPackage`.

epage: Hmm, this all matches `SerializedPackage` so I guess its no worse than where we are at today.

tests/testsuite/metadata.rs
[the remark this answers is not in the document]
eopb: > If you masquerade as nightly but do not enable the public dependency feature, then public is allowed but ignored. We can use that case for testing.

Addressed with this suggestion

src/cargo/core/dependency.rs
eopb: Addressed :+1:

tests/testsuite/metadata.rs
eopb: Oh nice!

Addressed with a force push

tests/testsuite/metadata.rs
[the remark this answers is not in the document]
eopb: Since it's now in `master` I've rebased my PR onto your change and removed `masquerade_as_nightly_cargo` from that test

epage:
Thanks!

@bors r+

PR #14507, opened 2024-09-06
fix(toml): Don't require MSRV bump for pub/priv

epage:

In #13308, we decided to make the feature gate for public/private dependencies non-blocking.
Generally, people opt-in to a feature that is non-blocking through `-Z` but some nightly users want an "always on" mode for this, so we offered both in #13340.
In #13340, we made both feature gates work but we did not make them non-blocking for stable, only nightly.

This change makes the feature gate non-blocking on stable.

Unfortunately, this means that 1.83 will be the MSRV for people using public dependencies.  Good thing the feature is indefinitely blocked in rustc as that gives us more time.

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -786,12 +786,8 @@ fn normalize_dependencies<'a>(
                 warnings,
             )?;
             if d.public.is_some() {
-                let public_feature = features.require(Feature::public_dependency());
-                let with_public_feature = public_feature.is_ok();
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
> Unfortunately, this means that 1.83 will be the MSRV for people using public dependencies.

Why is this unfortunate? I mean without this patch users can't even use it with stable toolchains.

epage:
Its not a matter of the feature being active on stable now but that in the future, people can use the feature without bumping their MSRV, like when we did `[lints]`.

PR #14515, opened 2024-09-07
docs: bin source can be `src/main.rs` and/or in `src/bin/`

### LINKED ISSUE #14463: Document having only `src/bin/*.rs` binaries is valid
### Problem

I found the following section of the cargo book could make readers think they must provide `src/main.rs`.

https://github.com/rust-lang/cargo/blob/59ecb11a298f0c3740216220361024cf05ed5b5a/src/doc/src/reference/cargo-targets.md?plain=1#L30-L34

However, it is also valid to have only `src/bin/*.rs` binaries and not to have `src/main.rs`.  
I believe clarifying this situation on the book will help readers when arranging codes for binaries.

### Proposed Solution

Changing the section like below should work:
```
Binary targets are executable programs that can be run after being compiled. 
The default binary filename is `src/main.rs`, which defaults to the name of 
the package. Additionally one or more binaries are stored in the [`src/bin/`
 directory][package layout]. In this case the default binary `src/main.rs` 
is not required. The settings for each binary can be [customized] in the
`[[bin]]` tables in `Cargo.toml`.
```

### Notes

_No response_

nekomachi-touge:
This commit clarifies the doc by stating we can place a binary source as `src/main.rs` and/or in `src/bin/`.
Wording is suggested by Ed Page (@epage).

(It seems the only way is to check the chenge visually...)

### Additional information
Closes #14463.

### DIFF
--- a/src/doc/src/reference/cargo-targets.md
+++ b/src/doc/src/reference/cargo-targets.md
@@ -28,10 +28,10 @@ bench = false
 ## Binaries
 
 Binary targets are executable programs that can be run after being compiled.
-The default binary filename is `src/main.rs`, which defaults to the name of
-the package. Additional binaries are stored in the [`src/bin/`
-directory][package layout]. The settings for each binary can be [customized]
-in the `[[bin]]` tables in `Cargo.toml`.
+A binary's source can be `src/main.rs` and/or stored in the [`src/bin/`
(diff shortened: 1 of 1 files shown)


weihanglo (approved):

weihanglo:
Thank you!

@bors r+

PR #14516, opened 2024-09-08
replace !Vec::is_empty() then Some(Vec) with bool::then()

Sk44rt:

None of existing issues.

Just replace if !vec.is_empty() {Some(vec)} with (!vec.is_empty()).then(|| vec) in src/bin/cargo/commands/rustc.rs
looks a little better i think

### DIFF
--- a/src/bin/cargo/commands/rustc.rs
+++ b/src/bin/cargo/commands/rustc.rs
@@ -80,11 +80,7 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
         compile_opts.build_config.requested_profile = InternedString::new("dev");
     }
     let target_args = values(args, "args");
-    compile_opts.target_rustc_args = if target_args.is_empty() {
-        None
-    } else {
-        Some(target_args)
-    };
+    compile_opts.target_rustc_args = (!target_args.is_empty()).then(|| target_args);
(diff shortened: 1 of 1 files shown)


src/bin/cargo/commands/rustc.rs
```diff
@@ -80,23 +80,15 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
         compile_opts.build_config.requested_profile = InternedString::new("dev");
     }
     let target_args = values(args, "args");
-    compile_opts.target_rustc_args = if target_args.is_empty() {
```
weihanglo, line 83: It has to be wrapped with additional parentheses so doesn't seem more readable to me. The clippy lint [`if_then_some_else_none`](https://rust-lang.github.io/rust-clippy/master/index.html#/if_then_some_else_none) also only hints for the pattern of `if-some-else-none` not for the reversed.

Is there any other purpose of this change? Otherwise I tend to close this. Thank you.

epage: Agreed, I've found that if `!` is used, I would rather have an `if` / `else if`

weihanglo:
Thank you. I am going to close this. See <https://github.com/rust-lang/cargo/pull/14516#discussion_r1749220740>.

Issue #14525, opened 2024-09-10
Vendoring of git dependencies is sometimes incorrectly skipped


stormshield-guillaumed:
### Problem

The vendoring of git dependencies is incorrectly skipped under certain conditions. From my tests and by looking at the source code of cargo, the conditions for a crate `A` to be incorrectly skipped during vendoring seems to be :
- `A` is a git dependency
- a version of `A` is already vendored with a version suffix, i.e. `log-0.4.22`
- you change the git revision of `A` to a different commit, but the crate version defined in the `Cargo.toml` of `A` is the same between the two commits

I made a minimal reproduction example in [this repo](https://github.com/stormshield-guillaumed/vendor-repro) with a CI job checking the update of the log dependency should change the vendor directory.

### Steps

1. `git clone https://github.com/stormshield-guillaumed/vendor-repro.git`
2. `sed -i s/d1a8306aadb88d56b74c73cdce4ef0153fb549cb/96dbf58ec05e844ec55aa2282f32f612372fbced/ Cargo.toml`
3. `cargo vendor --versioned-dirs > .cargo/config.toml`

The log's sources should have changed after this but they have not.

### Possible Solution(s)

The part where cargo decides whether it skips the vendoring of a crate or not seems to assume there is nothing to do if the crate version hasn't changed, which is true for dependencies on `crates.io`, but not for git dependencies where the same crate version can span multiples commits. Maybe this particular skip could be done only for dependencies on `crates.io` and not for git dependencies. That way, the optimisation is kept for `crates.io` dependencies which are the most common.

### Notes

The skip doesn't happen when the dependency is vendored without suffix, i.e. `log`.

### Version

```text
cargo 1.83.0-nightly (c1fa840a8 2024-08-29)
release: 1.83.0-nightly
commit-hash: c1fa840a85eca53818895901a53fae34247448b2
commit-date: 2024-08-29
host: x86_64-unknown-linux-gnu
libgit2: 1.8.1 (sys:0.19.0 vendored)
libcurl: 8.9.0-DEV (sys:0.4.74+curl-8.9.0 vendored ssl:OpenSSL/1.1.1w)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Ubuntu 24.4.0 (noble) [64-bit]
```

ehuss:
Thanks for the report! I believe this is a duplicate of #8181 (and probably the same underlying cause as #11897), so closing in favor of that.

Issue #11897, opened 2023-03-27
`cargo vendor` doesn't notice that local vendored content has diverged from crate registry


jhfrontz:
### Problem

I had been hoping that I could use `cargo vendor` to quickly identify any unofficial/local changes to a crate.

To test this, I made an innocuous change to a local copy of a crate -- I was expecting when I did `cargo vendor` that it would either point out the difference (or even just silently overwrite it).  However, it doesn't seem to do either.

The workaround seems to be to delete the entire `vendor` directory and then compare the result to what had been checked in (a la `git diff --exit-code`) -- after ignoring any semi-ephemeral content (e.g., `git add $(find . -type f -name Cargo.lock -o -name .cargo-checksum.json)`)

### Possible Solution(s)

It's not clear to me whether this is the intended behavior for `cargo vendor` or not -- but if it is, there should be a major note on the man page about `cargo vendor` retaining local changes.

### Version

```text
$ cargo version --verbose
cargo 1.65.0
release: 1.65.0
host: x86_64-unknown-linux-gnu
libgit2: 1.5.0 (sys:0.15.0 vendored)
libcurl: 7.29.0 (sys:0.4.55+curl-7.83.1 system ssl:NSS/3.53.1)
os: CentOS 7.9.2009 (Core) [64-bit]
```
```

**Comment by weihanglo** [comment]
Hmm… `cargo vendor` does check the integrity of vendored packages. It also overwrites files if you modify them manually. Can you provide steps to help the reproduction?

BTW, to make vendoring take effect, you need to put some configs under `.cargo/config.toml` first, like

```toml
[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "vendor"
```

weihanglo:
This is a reproducer that verifies Cargo is aware of vendored source being modified:

```
cargo new cargo-issue-11897
pushd cargo-issue-11897
cargo add empty-library
mkdir -p .cargo
cargo vendor > .cargo/config.toml
echo "pub fn f() {}" > vendor/empty-library/src/lib.rs
cargo c
```

You'll get an error like

```
error: the listed checksum of `/projects/cargo-issue-11897/vendor/empty-library/src/lib.rs` has changed:
expected: b81956f17c56fffd4b5e81f93eb647dd8ce78a89bccec826b5525368f172baba
actual:   16c2fca9e371936458d01576b4ca311c22d166a45539ccbc9104823d0b10db47
```

[...]

PR #14530, opened 2024-09-11
fix(vendor): trust crate version only when coming from registries

### LINKED ISSUE #8181: cargo vendor doesn't re-vendor if the version number hasn't changed
**Problem**

`cargo vendor` fails to re-vendor a package if:
- the package vendor dir has a version number (eg because it's not the most recent, or `--versioned-dirs` is in use)
- the code has changed

This can happen if the code is coming from a git repo and the `rev` has been updated, if a patch has been added in the `[patch]` section of the Cargo.toml, or when switching between crates.io and git as a source.

This doesn't happen for non-versioned directories because `cargo vendor` unconditionally re-vendors their contents.

**Possible Solution(s)**
Rather than just comparing the version, it should also check for changes to the patch and/or origin specification.

**Notes**

Output of `cargo version`: 1.42, but this hasn't changed in recent versions

stormshield-guillaumed:

Fixes #8181
Relates to #11897 and #14525

As mentioned in the contribution guide, I made a first commit adding a test that passes with the actual behaviour. Then, I made a second commit with a fix and modified the test with the new expected  behaviour.

### Additional information

The fix doesn't take into account switching from a git dependency to crates.io, which is not handled correctly on master either, and would probably require the vendoring to serialize the source ID to detect source changes.

I specifically limited the trust of immutable version to crates.io, but it could be extended to other registries.

### DIFF
--- a/src/cargo/ops/vendor.rs
+++ b/src/cargo/ops/vendor.rs
@@ -207,8 +207,10 @@ fn sync(
         let dst = canonical_destination.join(&dst_name);
         to_remove.remove(&dst);
         let cksum = dst.join(".cargo-checksum.json");
-        if dir_has_version_suffix && cksum.exists() {
-            // Always re-copy directory without version suffix in case the version changed
+        // Registries are the only immutable sources,
+        // path and git dependencies' versions cannot be trusted to mean "no change"
+        if dir_has_version_suffix && id.source_id().is_registry() && cksum.exists() {
+            // Don't re-copy directory with version suffix in case it comes from a registry
             continue;
         }
 
--- a/tests/testsuite/vendor.rs
+++ b/tests/testsuite/vendor.rs
@@ -1569,6 +1569,67 @@ path = [..]
 }
 
 #[cargo_test]
+fn git_update_rev() {
+    let (git_project, git_repo) = git::new_repo("git", |p| {
+        p.file("Cargo.toml", &basic_manifest("a", "0.1.0"))
+            .file("src/lib.rs", "")
+    });
+    let url = git_project.url();
+    let ref_1 = "initial";
+    let ref_2 = "update";
+
+    git::tag(&git_repo, ref_1);
+
+    git_project.change_file("src/lib.rs", "pub fn f() {}");
+    git::add(&git_repo);
+    git::commit(&git_repo);
+    git::tag(&git_repo, ref_2);
+
+    let p = project()
+        .file(
+            "Cargo.toml",
+            &format!(
+                r#"
+                    [package]
+                    name = "foo"
+                    version = "0.1.0"
+
+                    [dependencies]
+                    a = {{ git = '{url}', rev = '{ref_1}' }}
+                "#
+            ),
+        )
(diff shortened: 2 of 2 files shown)


src/cargo/ops/vendor.rs
```diff
@@ -207,8 +207,10 @@ fn sync(
         let dst = canonical_destination.join(&dst_name);
         to_remove.remove(&dst);
         let cksum = dst.join(".cargo-checksum.json");
-        if dir_has_version_suffix && cksum.exists() {
-            // Always re-copy directory without version suffix in case the version changed
+        // The standard registry is the only immutable source,
+        // path and git dependencies' versions cannot be trusted to mean "no change"
+        if dir_has_version_suffix && id.source_id().is_crates_io() && cksum.exists() {
```
jsgf, line 212: Will this have the effect of *always* revendoring git-sourced packages even if they haven't changed?

weihanglo: Yes. Unfortunately in `.cargo-checksum.json` there is no checksum recorded for git dependencies.

src/cargo/ops/vendor.rs
```diff
@@ -207,8 +207,10 @@ fn sync(
         let dst = canonical_destination.join(&dst_name);
         to_remove.remove(&dst);
         let cksum = dst.join(".cargo-checksum.json");
-        if dir_has_version_suffix && cksum.exists() {
-            // Always re-copy directory without version suffix in case the version changed
+        // The standard registry is the only immutable source,
+        // path and git dependencies' versions cannot be trusted to mean "no change"
```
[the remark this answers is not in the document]
stormshield-guillaumed: Could we keep this comment for when they will be or would you prefer I remove the mention to path dependencies ?
I wasn't sure whether to include all registries or not, but you are right, if a registry is not immutable, it will probably cause other problems before that. I'll change this.

src/cargo/ops/vendor.rs
stormshield-guillaumed: I think there is no way currently to retrieve the git rev of the vendored dependency.

src/cargo/ops/vendor.rs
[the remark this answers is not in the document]
weihanglo: We can leave the comment there. It just a note that Cargo doesn't vendor local path deps.

>  I wasn't sure whether to include all registries or not, but you are right, if a registry is not immutable, it will probably cause other problems before that. I'll change this.

Rebuild detection also holds that assumption, so yeah thanks for the update.

weihanglo (approved):
While this may affect performance loss as it becomes always-re-vendor for git dependencies, the bug itself is more severe because it vendors the wrong stuff.

To get further for vendor correctness, `cargo vendor` could always compare file checksums, though it implies more non-trivial refactors than this bugfix. I think we can proceed with this simple patch first.

ehuss:
CI issue is tracked in https://github.com/rust-lang/rust/issues/130291, will be blocked until that is fixed.

Issue #14520, opened 2024-09-09
Tracking Issue for native-completion


shannmu:
### Summary

Original issue: #6645  
Implementation: #14493  
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#native-completions

Please enter a short, one-sentence description here.

### Unresolved Issues

* [ ] Add custom completer for `cargo uninstall <TAB>`
* [ ] Add custom completer for `cargo build --bin=<TAB>`
* [ ] Add custom completer for `cargo test <TAB>` and `cargo `
* [ ] Add custom completer for `cargo build --bench=<TAB>`
* [ ] Add custom completer for `cargo build --example=<TAB>`
* [ ] Add custom completer for `cargo help <TAB>`
* [ ] Add custom completer for `cargo -Z <TAB>`
* [ ] Add custom completer for `cargo build --target=<TAB>`
* [ ] Add custom completer for `cargo +<TAB>` to complete toolchain name
* [ ] Add custom completer for `cargo <TAB>` to complete aliases defined in `config.toml`

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

epage:
@shannmu in testing this, some of the output isn't ideal.  The new completer is respecting the documented behavior in [`ValueHint::Unknown`](https://docs.rs/clap/latest/clap/enum.ValueHint.html#variant.Unknown) which means there are cases like `cargo add [TAB]` that show file paths when we shouldn't.  On the other hand, this is a good default for `cargo install --path [TAB]`.

We'll need to identify places to sprinkle `ValueHint::Other`

epage:
Fish does a lot of parsing of Cargo output and people doing that is a reason we're careful about changing our output.  It would be good to see if we could be feature parity with them and get them to switch to our completion generation.

Looking over what they do in addition to things specified
- They complete `--features`
- `cargo add` and `cargo install` complete using the crates.io search API
  - This is questionable because that can disclose confidential information

See https://github.com/fish-shell/fish-shell/blob/master/share/completions/cargo.fish

shannmu:
> We'll need to identify places to sprinkle `ValueHint::Other`

[...]

PR #14531, opened 2024-09-11
feat: Add custom completer for `cargo build --example=<TAB>`

shannmu:

Add custom completer for `cargo build --example=<TAB>`


src/cargo/util/command_prelude.rs
```diff
@@ -1027,6 +1036,37 @@ pub fn lockfile_path(
...
+    match get_examples_from_cwd() {
+        Ok(examples) => examples
+            .into_iter()
+            .map(|example| clap_complete::CompletionCandidate::new(example))
+            .collect(),
+        Err(_) => vec![],
+    }
+}
+
+fn get_examples_from_cwd() -> CargoResult<Vec<OsString>> {
```
epage, line 1049: We need this to be reworked like how I suggested we re-work bins, which will make this PR build on top of that.

PR #14532, opened 2024-09-11
feat: Add custom completer for completing benchmark names

shannmu:

Add custom completer for `cargo build --bench=<TAB>`

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -163,7 +163,9 @@ pub trait CommandExt: Sized {
             )
             ._arg(flag("benches", benches).help_heading(heading::TARGET_SELECTION))
             ._arg(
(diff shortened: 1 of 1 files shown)


src/cargo/util/command_prelude.rs
```diff
@@ -1027,6 +1031,43 @@ pub fn lockfile_path(
...
+        .map(|benches| {
+            benches
+                .into_iter()
+                .map(|bench| clap_complete::CompletionCandidate::new(bench))
+                .collect()
+        })
+        .unwrap_or(vec![])
+}
+
+fn get_benches_from_manifest() -> Option<Vec<String>> {
```
epage, line 1045: We need this to be reworked like how I suggested we re-work bins, which will make this PR build on top of that.

PR #14533, opened 2024-09-11
feat: Add custom completer for completing bin names

shannmu:

Add custom completer for `cargo build --bin=<TAB>`

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -1,6 +1,6 @@
 use crate::core::compiler::{BuildConfig, MessageFormat, TimingOutput};
 use crate::core::resolver::CliFeatures;
-use crate::core::{Edition, Workspace};
+use crate::core::{shell, Edition, Target, TargetKind, Workspace};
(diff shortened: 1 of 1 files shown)


src/cargo/util/command_prelude.rs
```diff
@@ -1027,6 +1040,42 @@ pub fn lockfile_path(
...
+    let menifest = env!("CARGO_MANIFEST_DIR");
+    let menifest = Path::new(menifest).join("Cargo.toml");
+
+    let value = toml::from_str::<toml::Value>(&std::fs::read_to_string(menifest)?)?;
+
+    match value {
+        toml::Value::Table(table) => Ok(table),
+        _ => Err(anyhow!("Error parsing Cargo.toml")),
+    }
+}
```
epage, line 1077: Most bins won't be found in `Cargo.toml`.

We should run logic like `cargo metadata` to load the workspace, enumerating targets and then filter the build targets for the ones that are bins.  This will then be easy to extend to examples, tests, benches, etc.

Issue #14518, opened 2024-09-08
Recommend using `cargo info` after the `cargo search` results


Rustin170506:
### Problem

Now that we have introduced `cargo info', I think it would be useful if we could advise users to use the `cargo info' command to inspect the specific crate from the search result.

### Proposed Solution

```console
$ cargo +nightly search clap
clap = "4.5.17"                     # A simple to use, efficient, and full-featured Command Line Argument Parser
clap-v3 = "3.0.0-beta.1"            # A simple to use, efficient, and full-featured Command Line Argument Parser
nameless-clap = "3.0.0-beta.2.2"    # A simple to use, efficient, and full-featured Command Line Argument Parser
conf = "0.1.0"                      # A derive-based config parser for CLI args and env parameters
clap-serde = "0.5.1"                # Provides a wrapper to deserialize clap app using serde.
clap-digest = "0.3.0"               # clap and digest integration
clap-nested = "0.4.0"               # Convenient `clap` for CLI apps with multi-level subcommands.
clap-markdown-dfir = "0.2.0"        # Autogenerate Markdown documentation for clap command-line tools
clap-stdin = "0.5.1"                # Provides a type for easily accepting Clap arguments from stdin
clap_derive-v3 = "3.0.0-beta.1"     # Parse command line argument by defining a struct, derive crate.
... and 785 crates more (use --limit N to see more)

note: to see more details about a specific crate, try `cargo info crate_name`
```

epage:
I agree, having a way to bridge users from command-to-command would be useful.  We do this today with `cargo update` -> `cargo tree` and `cargo info` -> `cargo tree`.  Knowing how to see more about a search result would be useful.

I would make a couple tweaks
```diff
-note: to see more details about a specific crate, try `cargo info crate_name`
+note: to learn more a package, run `cargo info <name>`
```

Rustin170506:
@rustbot claim

I will add it.

PR #14537, opened 2024-09-11
Suggest `cargo info` command in the `cargo search` result

Rustin170506:

close https://github.com/rust-lang/cargo/issues/14518

Suggest `cargo info` command in the `cargo search` result.

The old tests were updated.

### Additional information

~~I am not sure if the `note` function should be put into util. Does it make sense? Do you have a better place for it?~~ Use the `shell.note` instead.

### DIFF
--- a/src/cargo/ops/registry/search.rs
+++ b/src/cargo/ops/registry/search.rs
@@ -8,6 +8,7 @@ use anyhow::Context as _;
 use url::Url;
 
 use crate::util::style;
+use crate::util::style::LITERAL;
 use crate::util::truncate_with_ellipsis;
 use crate::CargoResult;
 use crate::GlobalContext;
@@ -85,6 +86,13 @@ pub fn search(
             total_crates - limit,
             extra
         );
+    }
+
+    if total_crates > 0 {
+        let literal = LITERAL;
+        shell.note(format_args!(
+            "to learn more about a package, run `{literal}cargo info <name>{literal:#}`",
+        ))?;
     }
 
     Ok(())
--- a/tests/testsuite/search.rs
+++ b/tests/testsuite/search.rs
@@ -115,7 +115,11 @@ fn not_update() {
     cargo_process("search postgres")
         .replace_crates_io(registry.index_url())
         .with_stdout_data(SEARCH_RESULTS)
-        .with_stderr_data("") // without "Updating ... index"
+        // without "Updating ... index"
+        .with_stderr_data(str![[r#"
+[NOTE] to learn more about a package, run `cargo info <name>`
+
+"#]])
         .run();
 }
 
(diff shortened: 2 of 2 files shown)


src/cargo/util/style.rs
```diff
@@ -11,3 +15,14 @@ pub const NOTE: Style = AnsiColor::Cyan.on_default().effects(Effects::BOLD);
 pub const GOOD: Style = AnsiColor::Green.on_default().effects(Effects::BOLD);
 pub const VALID: Style = AnsiColor::Cyan.on_default().effects(Effects::BOLD);
 pub const INVALID: Style = AnsiColor::Yellow.on_default().effects(Effects::BOLD);
+
+
+/// Write a note to the given writer.
```
Rustin170506, line 20: I am not sure if the note function should be put into here. Does this make sense? Do you have a better place for it?

Rustin170506 (commented):

src/cargo/ops/registry/search.rs
```diff
@@ -87,5 +90,28 @@ pub fn search(
...
+    let literal = LITERAL;
+
+    let _ = writeln!(stdout);
+    note(
+        format_args!("to learn more a package, run `{literal}cargo info <name>{literal:#}`",),
+        stdout,
+    )
+}
+
+pub(super) fn note(msg: impl std::fmt::Display, stdout: &mut dyn Write) -> CargoResult<()> {
```
epage: Why is this being used instead of `shell.note`?

Rustin170506: I didn't notice that. I have updated it.

tests/testsuite/search.rs
```diff
@@ -115,7 +115,10 @@ fn not_update() {
     cargo_process("search postgres")
         .replace_crates_io(registry.index_url())
         .with_stdout_data(SEARCH_RESULTS)
-        .with_stderr_data("") // without "Updating ... index"
+        .with_stderr_data(str![[r#"
```
epage, line 118: Please update the commits so each one passes tests

src/cargo/ops/registry/search.rs
```diff
@@ -87,5 +88,12 @@ pub fn search(
         );
     }
 
+    if total_crates > 0 {
+        let literal = LITERAL;
+        shell.note(format_args!(
+            "to learn more a package, run `{literal}cargo info <name>{literal:#}`",
```
epage, line 94:
```suggestion
            "to learn more about a package, run `{literal}cargo info <name>{literal:#}`",
```

Rustin170506 (commented):

src/cargo/ops/registry/search.rs
Rustin170506: Updated.

tests/testsuite/search.rs
Rustin170506: I have rebased them into one commit.

epage (approved):

PR #14540, opened 2024-09-12
docs: Feature resolver version 2: clarify use of 'target'

RalfJung:
IIUC, these two uses of the word `target` right next to each other in the docs actually have different meaning. That is kind of confusing, so add disambiguators and links to the glossary that should help clarify this.


weihanglo (approved):
Wish we could rename all this mess  someday 😂

weihanglo:
@bors r+

Thanks!

PR #14541, opened 2024-09-13
Gate shell_completions tests on availability

ehuss:
This changes the shell_completions tests to only run if the corresponding external program is installed. Generally, cargo's tests shouldn't require any external commands to be installed.

This also uses the `ignore` attribute for disabling on macos. It is quite confusing when running on macos to see a successful test when it isn't running. This ensures that the test output indicates that it is ignored and the reason.

cc @shannmu

### DIFF
--- a/tests/testsuite/shell_completions.rs
+++ b/tests/testsuite/shell_completions.rs
@@ -4,13 +4,9 @@ use cargo_test_support::cargo_test;
 use completest_pty::Runtime;
 use snapbox::assert_data_eq;
 
-#[cargo_test]
+#[cargo_test(requires_bash)]
+#[cfg_attr(target_os = "macos", ignore = "bash is not working on macOS")]
 fn bash() {
-    // HACK: At least on CI, bash is not working on macOS
-    if cfg!(target_os = "macos") {
-        return;
-    }
-
     let input = "cargo \t\t";
     let expected = snapbox::str![
         "% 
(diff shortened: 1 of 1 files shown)


weihanglo:
> error: failed to install component: 'llvm-tools-preview-x86_64-unknown-linux-gnu', detected conflict: 'lib/rustlib/x86_64-unknown-linux-gnu/bin/llc'

@bors try

epage:
I recommend we be free with reverts when a code change is blocking.  We can always fix this in another pass.

weihanglo (approved):

weihanglo:
https://github.com/rust-lang/rust/pull/130292 is merged. This shouldn't be blocked anymore.

@bors r+

ehuss:
@bors r-

This isn't going to work, since not all jobs have the shells installed.

I'm looking at a different approach where the jobs will be identified as being "extended" to test these extra things that might not be availbale.

ehuss:
And BTW, this will block updating in rust-lang/rust, since those won't have the shells installed either. I'm poking around now, but I'm a little confused why the `fish` test seems to be randomly failing.

shannmu:
The Fish shell test failures might be due to the test depending on the completest crate, which tests the completion results based on an interactive shell. We cannot be certain when the completion results are fully generated, so there is a built-in timeout parameter. If the completion results are not generated within the timeout period, it may cause the test to fail.

ehuss:
Hmm. That's unfortunate.

Let's just disable the tests for now. I have opened https://github.com/rust-lang/cargo/pull/14546 to do that, and https://github.com/rust-lang/cargo/issues/14545 lists the problems that I noticed while looking at this.

I'm probably not going to look at this further for now.

sway5959g-lgtm:
14546

PR #14543, opened 2024-09-14
feat(resolve): Direct people to working around less optimal MSRV-resolver results

epage:

In discussing #14414, the general problem of the resolver picking a
version older than a package needs for its MSRV (or lack of one) because of the MSRV of
other packages came up.
This tries to patch over that problem by telling users that a dependency
might be able to be newer than the resolver selected.

The message is fairly generic and might be misread to be about any MSRV
update which an MSRV `fallback` strategy allows, which would make the
count off.
The reason it is so generic is we don't know with precision why it was
held back
- Direct dependents may have a non-semver upper bound on the version as
  we aren't trying to unify the version requirements across direct
  dependents at this time
- A dependency could have removed a feature without making a breaking
  change
  - This seems like it should instead be an error but thats a
    conversation for another day
- ~~The user enabled `-Zminimal-versions`~~
  - This is now detected and the message skipped

Note: separate from this, we may also print the status suffix for this
case if the package was not selected for update (e.g. passing
`--workspace`).


tests/testsuite/workspaces.rs
```diff
@@ -698,6 +698,7 @@ fn share_dependencies() {
 [UPDATING] `dummy-registry` index
 [LOCKING] 1 package to latest compatible version
 [ADDING] dep1 v0.1.3 (available: v0.1.8)
+[NOTE] 1 package may have a higher, compatible versions. To update it, run `cargo update <name> --precise <version>
```
epage, line 701: This is an example of where it might show up for a non-MSRV case (and the suggested action will fail)

epage:
@joshtriplett this is a first pass at supporting a summary message and what wording we might use.

src/cargo/ops/cargo_update.rs
```diff
@@ -559,6 +561,22 @@ fn print_lockfile_generation(
...
     }
 
+    let compat_ver_compat_msrv = changed_stats.compat_ver_compat_msrv;
+    if 0 < compat_ver_compat_msrv
+        && !ws.gctx().cli_unstable().direct_minimal_versions
+        && !ws.gctx().cli_unstable().minimal_versions
+    {
+        if compat_ver_compat_msrv == 1 {
+            ws.gctx().shell().note(format!(
+                "{compat_ver_compat_msrv} package may have a higher, compatible versions. To update it, run `cargo update <name> --precise <version>"
```
joshtriplett, line 571:
```suggestion
                "{compat_ver_compat_msrv} package may have a higher, compatible version. To update it, run `cargo update <name> --precise <version>"
```

tests/testsuite/workspaces.rs
joshtriplett, line 701:
```suggestion
[NOTE] 1 package may have a higher, compatible version. To update it, run `cargo update <name> --precise <version>
```

tests/testsuite/rust_version.rs
```diff
@@ -1210,10 +1234,13 @@ fn report_rust_versions() {
...
 [LOCKING] 9 packages to latest Rust 1.60.0 compatible versions
-[ADDING] dep-only-high-incompatible v1.75.0 (requires Rust 1.75.0)
-[ADDING] dep-only-low-incompatible v1.75.0 (requires Rust 1.75.0)
+[ADDING] dep-only-high-compatible v1.55.0 (available: v1.65.0)
+[ADDING] dep-only-high-incompatible v1.55.0 (available: v1.75.0, requires Rust 1.75.0)
+[ADDING] dep-only-low-incompatible v1.55.0 (available: v1.75.0, requires Rust 1.75.0)
+[ADDING] dep-only-unset-compatible v1.55.0 (available: v1.75.0)
 [ADDING] dep-only-unset-incompatible v1.2345.0 (requires Rust 1.2345.0)
 [ADDING] dep-shared-incompatible v1.75.0 (requires Rust 1.75.0)
+[NOTE] 2 packages may have a higher, compatible version. To update them, run `cargo update <name> --precise <version>
```
epage, line 1243: For me, this message is vague to the point of being unhelpful.

joshtriplett: In the case where we're *already* showing verbose messages about the individual crates, I agree. Could we show it only if there are messages about possible upgrades that we're currently hiding because of the verbosity level?

epage: Thats sounding like the message we already have
```
"pass `--verbose` to see {unchanged_behind} unchanged dependencies behind latest"
```

PR #14552, opened 2024-09-17
feat: Add support for completing `cargo update <TAB>`

shannmu:

Add custom completer for `cargo update <TAB>`

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -1,9 +1,11 @@
-use crate::core::compiler::{BuildConfig, MessageFormat, TimingOutput};
-use crate::core::resolver::CliFeatures;
-use crate::core::{shell, Edition, Target, TargetKind, Workspace};
+use crate::core::compiler::{
+    BuildConfig, CompileKind, MessageFormat, RustcTargetData, TimingOutput,
+};
+use crate::core::resolver::{CliFeatures, ForceAllTargets, HasDevUnits};
+use crate::core::{shell, Edition, Package, Target, TargetKind, Workspace};
 use crate::ops::lockfile::LOCKFILE_NAME;
 use crate::ops::registry::RegistryOrIndex;
-use crate::ops::{CompileFilter, CompileOptions, NewOptions, Packages, VersionControl};
+use crate::ops::{self, CompileFilter, CompileOptions, NewOptions, Packages, VersionControl};
 use crate::util::important_paths::find_root_manifest_for_wd;
 use crate::util::interning::InternedString;
 use crate::util::is_rustup;
@@ -20,6 +22,8 @@ use cargo_util_schemas::manifest::RegistryName;
 use cargo_util_schemas::manifest::StringOrVec;
 use clap::builder::UnknownArgumentValueParser;
 use home::cargo_home_with_cwd;
+use semver::Version;
+use std::collections::HashMap;
 use std::ffi::{OsStr, OsString};
 use std::path::Path;
 use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


src/cargo/util/command_prelude.rs
```diff
@@ -1067,6 +1068,29 @@ fn get_targets_from_metadata() -> CargoResult<Vec<Target>> {
     Ok(targets)
 }
 
+pub fn get_dependency_name_candidates() -> Vec<clap_complete::CompletionCandidate> {
+    get_dependencies_from_metadata()
+        .unwrap_or_default()
+        .into_iter()
+        .map(|dep| clap_complete::CompletionCandidate::new(dep.name_in_toml().as_str()))
```
epage, line 1075: Let's include the descriptions

epage, line 1075: I think this is supposed to be package name, not dependency name (`name_in_toml`)

epage, line 1075: package names may not always be sufficient.

In clap, run
```console
$ cargo update clap
error: There are multiple `clap` packages in your project, and the specification `clap` is ambiguous.
Please re-run this command with one of the following specifications:
  clap@4.5.9
  clap@4.5.17
```

`PackageIdSpec::query::minimize` has a crude algorithm for uniquifing things.  We should have a more robust form.
- Put all packages into a map of `name -> Vec<Package>`
- For any entries with duplicates but unique versions, report `name@version`
- If there are duplicate versions, report the full `PackageIdSpec`

src/cargo/util/command_prelude.rs
```diff
@@ -1067,6 +1068,29 @@ fn get_targets_from_metadata() -> CargoResult<Vec<Target>> {
     Ok(targets)
 }
 
+pub fn get_dependency_name_candidates() -> Vec<clap_complete::CompletionCandidate> {
```
epage, line 1071:
```suggestion
pub fn get_package_candidates() -> Vec<clap_complete::CompletionCandidate> {
```

src/cargo/util/command_prelude.rs
```diff
@@ -1067,6 +1068,29 @@ fn get_targets_from_metadata() -> CargoResult<Vec<Target>> {
...
+    get_dependencies_from_metadata()
+        .unwrap_or_default()
+        .into_iter()
+        .map(|dep| clap_complete::CompletionCandidate::new(dep.name_in_toml().as_str()))
+        .collect::<Vec<_>>()
+}
+
+fn get_dependencies_from_metadata() -> CargoResult<Vec<Dependency>> {
```
epage, line 1079:
```suggestion
fn get_packages() -> CargoResult<Vec<Dependency>> {
```

src/cargo/util/command_prelude.rs
```diff
@@ -1067,6 +1068,29 @@ fn get_targets_from_metadata() -> CargoResult<Vec<Target>> {
...
+
+    let dependencies = ws
+        .members()
+        .flat_map(|pkg| pkg.dependencies().into_iter().cloned())
+        .collect::<HashSet<_>>()
+        .into_iter()
+        .collect::<Vec<_>>();
+
+    Ok(dependencies)
+}
```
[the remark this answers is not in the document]
shannmu: > We can either updated our `GlobalContext` to be quiet (should probably go back and do that to all of these completions anyways)

Are you referring to updating the `GlobalContext`(in the main function) within the custom completer function?

If this is what you mean, I think we will face a problem: the custom completer function cannot carry additional parameters. Capturing `&mut gctx` with a closure would likely be restricted by lifetime limitations. If there are other methods I haven’t considered, please let me know. Thank you!

src/cargo/util/command_prelude.rs
shannmu: Thanks for your guidance!

src/cargo/util/command_prelude.rs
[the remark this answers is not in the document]
epage: > `    let gctx = GlobalContext::new(shell::Shell::new(), cwd.clone(), cargo_home_with_cwd(&cwd)?);`

For the `gctx` we have, we should force it to be quiet.  The primary-path in the CLI handles that through calling [`configure`](https://docs.rs/cargo/latest/cargo/util/context/struct.GlobalContext.html#method.configure).  Unsure if there are more direct ways of configuring it.  We might also want to call `configure` anyways as our context is only partially initialized and I've not audited it for any problems we can run into without fully initializing it.

src/cargo/util/command_prelude.rs
```diff
@@ -1174,6 +1178,122 @@ fn get_target_triples_from_rustc() -> CargoResult<Vec<clap_complete::CompletionC
         .collect())
 }
 
+pub fn get_package_candidates() -> Vec<clap_complete::CompletionCandidate> {
```
epage, line 1181: we should clarify this is `get_pkg_id_spec_candidates` with this change.

I only realized this as I wondered if we could reuse this for `cargo info` (maybe) and `cargo add` (no)

src/cargo/util/command_prelude.rs
```diff
@@ -1174,6 +1178,122 @@ fn get_target_triples_from_rustc() -> CargoResult<Vec<clap_complete::CompletionC
...
+                })
+                .collect::<Vec<_>>()
+        })
+        .collect::<Vec<_>>()
+}
+
+fn get_packages() -> CargoResult<Vec<Package>> {
+    let cwd = std::env::current_dir()?;
+    let mut gctx = GlobalContext::new(shell::Shell::new(), cwd.clone(), cargo_home_with_cwd(&cwd)?);
+    gctx.configure(0, true, None, false, true, false, &None, &[], &[])?;
```
epage, line 1266: Could you split these out into variables to clarify their names?

We'll likely need this in all of the completions that create a context, so maybe we should provide a helper for this?  Feels too specialized and accident prone to live on `GlobalContext` but unsure.

src/cargo/util/command_prelude.rs
```diff
@@ -1174,6 +1178,122 @@ fn get_target_triples_from_rustc() -> CargoResult<Vec<clap_complete::CompletionC
...
+                                    .description
+                                    .to_owned()
+                                    .map(From::from),
+                            )
+                        })
+                        .collect::<Vec<_>>()
+                })
+                .collect::<Vec<_>>()
+        })
+        .collect::<Vec<_>>()
```
epage, line 1260: 1. are all of these `collect`s needed?  I thought `flat_map` accepted an `IntoIterator`
2. This feels like too complex for combinators for me

src/cargo/util/command_prelude.rs
shannmu: Change done! Thanks!

src/cargo/util/command_prelude.rs
shannmu: > We'll likely need this in all of the completions that create a context, so maybe we should provide a helper for this?

I will refactor other completers used `gctx` in another pr.

> Feels too specialized and accident prone to live on GlobalContext but unsure.

For now, I make this function a non-pub function for now. I think putting it in the impl of `GlobalContext` is too specialized.

src/cargo/util/command_prelude.rs
shannmu: Modify it a clearer form! Thanks!

shannmu:
The completion speed of this feature is not very ideal. On my local machine, I can feel a slight lag.

epage:
> The completion speed of this feature is not very ideal. On my local machine, I can feel a slight lag.

Looked at it using https://doc.crates.io/contrib/tests/profiling.html

image

Note: this was done on a debug build, doing `cargo update [TAB]` within Cargo.  There was a pause but not too terrible.

- We should probably switch approaches so tracing output is flushed
- We have some unidentified gaps in time
- I've been looking to improve package load times which is about half of the total execution time
- If we're okay with dropping the package description, we could remove the `download_accessible` call by switching from `resolve_ws_with_opts` to `resolve_ws`, removing all of the manifest parsing time
  - If we want to keep descriptions, I suspect there is a faster way than running the feature resolver (which `resolve_ws_with_opts` does)
- The resolve times I've already been having conversations on how to speed up

shannmu:
> The resolve times I've already been having conversations on how to speed up

May I ask where is the discussion about how to speed up resolve times?

Issue #9289, opened 2021-03-19
`cargo install --locked` is not really locked


ehuss:
**Problem**
`cargo install --locked` causes cargo to use a `Cargo.lock` file if it exists (instead of ignoring it).  However, unlike all other commands it does not *require* that the `Cargo.lock` file be up-to-date.  This may be confusing since it is inconsistent.

For installing registry dependencies, this is probably not a big deal because the way cargo publishes, it generates a fresh `Cargo.lock`, so we can usually assume it is not outdated.

For `cargo install --path . --locked`, this may be a little more likely since there is a greater chance that `Cargo.lock` is out-of-date.  For example, you add or change a dependency in `Cargo.toml`, and don't run any other commands.

**Steps**

```rust
#[cargo_test]
fn install_outdated_locked() {
    // This Cargo.lock is intentionally out-of-date, missing the `bar` dependency.
    Package::new("bar", "0.1.0").publish();
    Package::new("foo", "0.1.0")
        .dep("bar", "0.1")
        .file("src/main.rs", "extern crate bar; fn main() {}")
        .file(
            "Cargo.lock",
            r#"
[[package]]
name = "foo"
version = "0.1.0"
"#,
        )
        .publish();

    // This succeeds when it should fail, since the published `Cargo.lock` is
    // out of date.
    cargo_process("install --locked foo").run();
}
```

**Possible Solution(s)**

This may be difficult to solve for a few reasons:
* `cargo install` ignores dev-dependencies during resolution, which results in a different graph if there are dev-dependencies.
* `cargo install` for a registry filters out the root package from the initial resolve graph [here](https://github.com/rust-lang/cargo/blob/c9ce0ec40dd7f12ee3ee81c334a587798cf278ec/src/cargo/core/resolver/encode.rs#L406-L413). I don't fully understand the motivation for that, but it makes it more difficult to compare the original lock.

**Notes**

`cargo 1.52.0-nightly (90691f2bf 2021-03-16)`

foxpy:
I have personally ran into a problem with bad documentation. `cargo install --help` says that:

```
--locked                    Require Cargo.lock is up to date
```

This is very confusing. The behaviour I am experiencing is that `cargo install` builds executable using fresh libraries and `cargo install --locked` build executable using libraries locked in `Cargo.lock`.

I believe this fragment of help documentation should really be updated to prevent further confusion.

mulkieran:
"up to date" wrt. what specifically, would be a good first step.

[...]

PR #14556, opened 2024-09-17
feat: lockfile path implies --locked on cargo install

Ifropc:
Follow-up of https://github.com/rust-lang/cargo/issues/14421
Resolving one of the items 
> cargo install should make --lockfile-path imply --locked

Simply mirrored behavior as if `--locked` was provided (on creating the workspace)

### DIFF
--- a/src/cargo/ops/cargo_install.rs
+++ b/src/cargo/ops/cargo_install.rs
@@ -46,7 +46,6 @@ struct InstallablePackage<'gctx> {
     vers: Option<VersionReq>,
     force: bool,
     no_track: bool,
-
     pkg: Package,
     ws: Workspace<'gctx>,
     rustc: Rustc,
@@ -68,6 +67,7 @@ impl<'gctx> InstallablePackage<'gctx> {
         no_track: bool,
         needs_update_if_source_is_index: bool,
         current_rust_version: Option<&PartialVersion>,
+        lockfile_path: Option<&Path>,
     ) -> CargoResult<Option<Self>> {
         if let Some(name) = krate {
             if name == "." {
@@ -155,6 +155,7 @@ impl<'gctx> InstallablePackage<'gctx> {
                     &root,
                     &dst,
                     force,
+                    lockfile_path,
                 ) {
                     let msg = format!(
                         "package `{}` is already installed, use --force to override",
@@ -179,15 +180,32 @@ impl<'gctx> InstallablePackage<'gctx> {
             }
         };
 
-        let (ws, rustc, target) =
-            make_ws_rustc_target(gctx, &original_opts, &source_id, pkg.clone())?;
-        // If we're installing in --locked mode and there's no `Cargo.lock` published
-        // ie. the bin was published before https://github.com/rust-lang/cargo/pull/7026
-        if gctx.locked() && !ws.root().join("Cargo.lock").exists() {
-            gctx.shell().warn(format!(
-                "no Cargo.lock file published in {}",
-                pkg.to_string()
-            ))?;
+        let (ws, rustc, target) = make_ws_rustc_target(
+            gctx,
+            &original_opts,
+            &source_id,
+            pkg.clone(),
+            lockfile_path.clone(),
+        )?;
+
+        if gctx.locked() {
+            // When --lockfile-path is set, check that passed lock file exists
+            // (unlike the usual flag behavior, lockfile won't be created as we imply --locked)
+            if let Some(requested_lockfile_path) = ws.requested_lockfile_path() {
+                if !requested_lockfile_path.is_file() {
+                    bail!(
+                        "no Cargo.lock file found in the requested path {}",
+                        requested_lockfile_path.display()
+                    );
+                }
+            // If we're installing in --locked mode and there's no `Cargo.lock` published
+            // ie. the bin was published before https://github.com/rust-lang/cargo/pull/7026
+            } else if !ws.root().join("Cargo.lock").exists() {
+                gctx.shell().warn(format!(
+                    "no Cargo.lock file published in {}",
+                    pkg.to_string()
+                ))?;
+            }
         }
         let pkg = if source_id.is_git() {
             // Don't use ws.current() in order to keep the package source as a git source so that
@@ -246,7 +264,6 @@ impl<'gctx> InstallablePackage<'gctx> {
             vers: vers.cloned(),
             force,
             no_track,
-
             pkg,
             ws,
             rustc,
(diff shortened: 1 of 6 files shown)


src/cargo/ops/cargo_install.rs
```diff
@@ -825,6 +825,10 @@ fn make_ws_rustc_target<'gctx>(
         ws
     };
     ws.set_ignore_lock(gctx.lock_update_allowed());
+    // if --lockfile-path is set, imply --locked
+    if ws.get_requested_lockfile_path().is_some() {
+        ws.set_ignore_lock(false);
+    }
```
epage, line 831: Could you add a test for this?

tests/testsuite/lockfile_path.rs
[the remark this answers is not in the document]
Ifropc: Will do!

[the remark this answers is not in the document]
Ifropc: @weihanglo re-organized commits

tests/testsuite/lockfile_path.rs
```diff
@@ -400,6 +400,56 @@ bar = "0.1.0"
...
+        .run();
+    cargo_process("install foo -Zunstable-options --lockfile-path lockfile_dir/Cargo.lock")
+        .masquerade_as_nightly_cargo(&["lockfile-path"])
+        .with_stderr_data(str![[r#"
+...
+[ERROR] unexpected argument '--lockfile-path' found
+...
+"#]])
+        .with_status(1)
+        .run();
```
epage: The intent behind splitting the commits is to show the change in behavior.

Showing the flag is not supported is the less interesting change in behavior.  I believe the intent would be to have the test without the flag in this commit, showing what behavior you get today.  The next commit would add the flag and show how the behavior changed.

Ifropc: Hmm okay I see what you are saying so we want to pass just --locked without --lockfile-path on the first commit and then change the behavior. Let me do that!

Ifropc: ✅

src/cargo/ops/cargo_install.rs
weihanglo, line 1: @epage I wonder the idea of `cargo install --locked-path` implying `--locked`. Is it because `cargo install` doesn't respect `Cargo.lock` at all? The semantic of using a lockfile is different from `--locked`. There might be an extra lockfile that is a bit outdated and `cargo install --lockfile-path` could help update it.

(I don't know if it is a valid use case. Just try to think it through).

epage: Thats true, `--locked` can have two meanings
- use lockfile (whats used in `cargo install`)
- lockfile is up-to-date (whats used elsewhere)

Note that `cargo install` does not currently have both behaviors, see #9289.

I think here, the care about is only "use lockfile".  We should not imply "lockfile is up-to-date" if that ever gets added.

weihanglo: Understood!

I think the current implementation (turn the `--locked` flag on) meets what we expect now. We may just need some tests ensuring `--lockfile-path` not to imply "lockfile is up-to-date".

src/cargo/ops/cargo_install.rs
```diff
@@ -189,6 +195,13 @@ impl<'gctx> InstallablePackage<'gctx> {
                 pkg.to_string()
             ))?;
         }
+        // When --lockfile-path is set, move lock file to the new location
+        // (the new location is expected downstream and will be used during compilation)
+        if gctx.locked() && ws.get_requested_lockfile_path().is_some() {
+            let requested_lockfile_path = ws.get_requested_lockfile_path().unwrap();
+            paths::create_dir_all(ws.lock_root().as_path_unlocked())?;
+            fs::rename(ws.root().join("Cargo.lock"), requested_lockfile_path)?;
```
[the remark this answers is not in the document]
Ifropc: Oh okay makes sense. I was confused about the exact flow we are going for here. I'll update the implementation

src/cargo/ops/cargo_install.rs
Ifropc: > We may just need some tests ensuring --lockfile-path not to imply "lockfile is up-to-date".

Test that was added (`install_respects_lock_file_path`) should [cover this case](https://github.com/rust-lang/cargo/blob/02c7b5e805769ce1258df9ffa45b9962714e93b1/tests/testsuite/lockfile_path.rs#L407-L410) I think

src/cargo/ops/cargo_install.rs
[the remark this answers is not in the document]
Ifropc: Forgot to mention, the implementation has been updated

src/cargo/ops/cargo_install.rs
```diff
@@ -189,6 +195,18 @@ impl<'gctx> InstallablePackage<'gctx> {
...
+        // When --lockfile-path is set, check that passed lock file exists
+        // (unlike the usual flag behavior, lockfile won't be created as we imply --locked)
+        if gctx.locked() {
+            if let Some(requested_lockfile_path) = ws.requested_lockfile_path() {
+                if !ws.root().join("Cargo.lock").exists() {
+                    bail!(
+                        "no Cargo.lock file found in {}",
+                        requested_lockfile_path.display()
+                    );
+                }
```
weihanglo, line 207: We are going to use the alternative lockfile, so don't need to check lockfile in the workspace root I believe.

```suggestion
                bail!(
                    "no Cargo.lock file found in {}",
                    requested_lockfile_path.display()
                );
```

Also, I don't think we want to emit the `no Cargo.lock file published in ...` on line 194, so make then conditioned on the existence of `requested_lockfile_path`, like

```rust
if gctx.locked() {
    if let Some(requested_lockfile_path) = ws.requested_lockfile_path() {
        // error
    } else if !ws.root().join("Cargo.lock").exists() {
        // warn
    }
} 
```

Ifropc: Apologies, I believe it was left from when the logic was to copy it. Fixed! Thank you 🙏🏼

src/cargo/ops/cargo_install.rs
```diff
@@ -179,15 +180,32 @@ impl<'gctx> InstallablePackage<'gctx> {
...
+            &source_id,
+            pkg.clone(),
+            lockfile_path.clone(),
+        )?;
+
+        if gctx.locked() {
+            // When --lockfile-path is set, check that passed lock file exists
+            // (unlike the usual flag behavior, lockfile won't be created as we imply --locked)
+            if let Some(requested_lockfile_path) = ws.requested_lockfile_path() {
+                if (!requested_lockfile_path.exists()) {
```
weihanglo, line 195: Perhaps better.



```suggestion
                if !requested_lockfile_path.is_file() {
```

After this change I believe we're ready to go :)

weihanglo (approved):

weihanglo:
Ready to merge. Thanks!

I'll track this behavior discrepancy in the tracking issue since people may have different thoughts when stabilization.

@bors r+

PR #14557, opened 2024-09-17
feat: Add custom completer for `cargo help <TAB>`

shannmu:

Add custom completer for `cargo help <TAB>`

### Additional information
The current completer function is quite slow because it executes too many functions. One idea I have is to use the file!() macro to list the filenames under the commands directory, excluding mod.rs, and return them as the completion results. Would this approach be too hacky?

### DIFF
--- a/src/bin/cargo/commands/help.rs
+++ b/src/bin/cargo/commands/help.rs
@@ -15,7 +15,19 @@ const COMPRESSED_MAN: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/man.tgz"
 pub fn cli() -> Command {
     subcommand("help")
         .about("Displays help for a cargo subcommand")
-        .arg(Arg::new("COMMAND").action(ArgAction::Set))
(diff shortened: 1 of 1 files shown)


src/bin/cargo/commands/help.rs
```diff
@@ -15,7 +15,20 @@ const COMPRESSED_MAN: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/man.tgz"
 pub fn cli() -> Command {
     subcommand("help")
         .about("Displays help for a cargo subcommand")
-        .arg(Arg::new("COMMAND").action(ArgAction::Set))
+        .arg(Arg::new("COMMAND").action(ArgAction::Set).add(
+            clap_complete::ArgValueCandidates::new(|| {
+                // TODO: Need a faster way to get all subcommand names.
```
epage, line 20: Unless we have something telling us this is slow, we shouldn't bother with this TODO, especially tracking it in source.

PR #14562, opened 2024-09-18
Revert "remove reference to incomplete crates.io feature from docs"

epage:
This reverts commit 1735917adc11aaa14011a187baf47a320f1c5bc5.

As discussed in #13660, my stance on removing the docs was if there wasn't a path forward.
There is work progressing on the docs.rs side which will unblock the crates.io side.

If the concern for removing the docs is for new crates, then no harm in removing it for now but also little benefit.
If its to get existing crates to change, I don't think that ecosystem churn is worth it to try to get everyone to add the link and then remove it again.

### DIFF
--- a/src/doc/src/reference/manifest.md
+++ b/src/doc/src/reference/manifest.md
@@ -209,13 +209,17 @@ description = "A short description of my package"
 ### The `documentation` field
 
 The `documentation` field specifies a URL to a website hosting the crate's
-documentation.
+documentation. If no URL is specified in the manifest file, [crates.io] will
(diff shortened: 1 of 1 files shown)


weihanglo:
Since there are at least two maintainers agree with reverting, and myself have no strong opinion on it, I am going to click the merge button.

@bors r+

RalfJung:
I continue to think this is doing our users a disservice, but I am being overruled here so I guess I won't try to improve this corner of Rust any further. :shrug:

PR #14563, opened 2024-09-18
docs(unstable): Expand on completion documentation

epage:

Wanting to improve the workflow for people testing completions

### DIFF
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -1695,6 +1695,23 @@ This feature moves the handwritten completion scripts to Rust native, making it
 easier for us to add, extend and test new completions. This feature is enabled with the
 nightly channel, without requiring additional `-Z` options.
 
+Areas of particular interest for feedback
+- Arguments that need escaping or quoting that aren't handled correctly
+- Inaccuracies in the information
+- Bugs in parsing of the command-line
+- Arguments that don't report their completions
+- If a known issue is being problematic
+
+Feedback can be broken down into
+- What completion candidates are reported
+  - Known issues: [#14520](https://github.com/rust-lang/cargo/issues/14520), [`A-completions`](https://github.com/rust-lang/cargo/labels/A-completions)
+  - [Report an issue](https://github.com/rust-lang/cargo/issues/new) or [discuss the behavior](https://github.com/rust-lang/cargo/issues/14520)
+- Shell integration, command-line parsing, and completion filtering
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/unstable.md
```diff
@@ -1695,21 +1695,38 @@ This feature moves the handwritten completion scripts to Rust native, making it
...
+- Shell integration, command-line parsing, and completion filtering
+  - Known issues: [clap#3166](https://github.com/clap-rs/clap/issues/3166), [clap's `A-completions`](https://github.com/clap-rs/clap/labels/A-completion)
+  - [Report an issue](https://github.com/clap-rs/clap/issues/new/choose) or [discuss the behavior](https://github.com/clap-rs/clap/discussions/new/choose)
+
+When in doubt, you can discuss this in [#14520](https://github.com/rust-lang/cargo/issues/14520) or on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo)
+
 ### How to use native-completions feature:
 - bash:
-  Add `source <(CARGO_COMPLETE=bash cargo)` to your .bashrc.
+  Add `source <(CARGO_COMPLETE=bash cargo +nightly)` to your .bashrc.
```
Rustin170506, line 1717: Do we need to specify the minimal nightly version here? Because if users are using the old nightly version, it doesn't work.

epage: We don't generally specify that in the unstable documentation.

src/doc/src/reference/unstable.md
```diff
@@ -1695,21 +1695,38 @@ This feature moves the handwritten completion scripts to Rust native, making it
 easier for us to add, extend and test new completions. This feature is enabled with the
 nightly channel, without requiring additional `-Z` options.
 
+Areas of particular interest for feedback
```
weihanglo, line 1698: We don't really document the feedback process for other unstable features. Why do we decide to do it here?

epage: - Most of this feature lives outside of the cargo's repo so usual processes will likely add more friction
- There are enough corner cases with this that I want to be very cautious in saying this is good enough to replace the existing completions so I need to depend more on broad feedback than we typically do

weihanglo: Fair enough. We might also want a copy of this process in the tracking issue.

weihanglo (approved):

weihanglo:
@bors r+

Thanks!

PR #14564, opened 2024-09-18
fix(complete): Harden `--target` completions

epage:

I found `--target` wasn't working with rustup as I expected, so this fixes it.  We generally only reference rustup if we are running under it but I think in UX code like this, with a fallback scheme, this should be reasonable enough.

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -1123,11 +1123,13 @@ fn get_target_triples() -> Vec<clap_complete::CompletionCandidate> {
 
     if is_rustup() {
         if let Ok(targets) = get_target_triples_from_rustup() {
-            candidates.extend(targets);
+            candidates = targets;
         }
-    } else {
+    }
+
+    if candidates.is_empty() {
         if let Ok(targets) = get_target_triples_from_rustc() {
-            candidates.extend(targets);
+            candidates = targets;
         }
     }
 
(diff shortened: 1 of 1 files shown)


epage:
Its further described in b9fc8d4

> clap-rs/clap#5733 removed the rustup proxy so that
> `CARGO_COMPLETE=bash cargo +nightly` works
> (with a side benefit of removing the proxy overhead).
> 
> As a downside, cargo no longer knows it is running within rustup, so we
> aren't reading `--target` candidates from rustup.
> 
> This changes the code to always try rustup.  It is likely a good enough
> source, even if the user isn't currently using it.
> The candidates should be about the same, just rustup hides some by
> default.
> Hiding just means it isn't shown by default but if only hidden
> candidates match, then we show them.

src/cargo/util/command_prelude.rs
weihanglo, line 1: This looks reasonable, though I had a hard time making it work on my machine (macOS).

* [x] Set the default toolchain to latest nightly
* [x] Ran the `source <(CARGO_COMPLETE=$SHELL cargo)`


To test this manually, what am I missing here?

epage: That should be it.  We'll likely need to gather more information to debug whats going wrong, like what shell you are using.

weihanglo (approved):
Was my bash too old or something, anyway it is fixed and working. Thanks!

Issue #13975, opened 2024-05-28
Deduplicate fingerprint logic and `check-cfg` lint config


Urgau:
https://github.com/rust-lang/cargo/pull/13958 introduce a duplication in the code between the fingerprint logic and the `check-cfg` lint config, because of Cargo MSRV, the `--check-cfg` code generation is dependant on a minimum `rustc` version (1.79).

It would be good to consolidate https://github.com/rust-lang/cargo/blob/431db31d0dbeda320caf8ef8535ea48eb3093407/src/cargo/core/compiler/fingerprint/mod.rs#L1424-L1443 and https://github.com/rust-lang/cargo/blob/2415192bcd369a375311eec14422e3b4970ad6e3/src/cargo/core/compiler/mod.rs#L1363-L1384 in https://github.com/rust-lang/cargo/blob/2415192bcd369a375311eec14422e3b4970ad6e3/src/cargo/util/toml/mod.rs#L2333

Originate from https://github.com/rust-lang/cargo/pull/13958#discussion_r1611728409

PR #14567, opened 2024-09-19
Cleanup duplicated check-cfg lints logic

Urgau:

This PR clean-ups some duplication left by https://github.com/rust-lang/cargo/pull/13958, because of Cargo MSRV.

Fixes https://github.com/rust-lang/cargo/issues/13975

The tests in `tests/testsuite/check_cfg.rs` show no change in behaviours (except for the better error messages). I suggest maybe reviewing commit by commit.


epage:
Thanks for coming back to fix this!

PR #14568, opened 2024-09-19
fix(resolve): Don't list transitive, incompatible dependencies as available

epage:

We have limited capability to clearly communicate to the user the different update states without overwhelming them. 
Before we showed all versions that were behind, with color distinguishing how actionable they are.
Color doesn't communicate enough though and we don't want to add footnotes to clarify everything.
For now, the least useful messages will be removed, available, transitive, incompatible versions.  These only give the user a hint of tech debt within their dependencies and can't be acted upon.

This is part of #13908

### DIFF
--- a/src/cargo/ops/cargo_update.rs
+++ b/src/cargo/ops/cargo_update.rs
@@ -774,7 +774,7 @@ fn report_latest(possibilities: &[IndexSummary], change: &PackageChange) -> Opti
     let version_req = package_id.version().to_caret_req();
     let required_rust_version = change.required_rust_version.as_ref();
 
-    if let Some(summary) = possibilities
+    let compat_ver_compat_msrv_summary = possibilities
         .iter()
         .map(|s| s.as_summary())
         .filter(|s| {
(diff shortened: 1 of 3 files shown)


src/cargo/ops/cargo_update.rs
```diff
@@ -787,45 +787,43 @@ fn report_latest(possibilities: &[IndexSummary], change: &PackageChange) -> Opti
...
-    {
-        let warn = if change.is_transitive.unwrap_or(true) {
-            Default::default()
-        } else {
-            style::WARN
-        };
-        let version = summary.version();
-        let report = format!(" {warn}(available: v{version}){warn:#}");
-        return Some(report);
+    if !change.is_transitive.unwrap_or(true) {
```
weihanglo, line 798: I might start forgetting things again. What does `is_transitive` mean here? It doesn't seem to be "transtive dependency" comparing to direct dependency. Or, is everything in `PackageDiff` is transitive?

epage: It is transitive vs direct dependency.

weihanglo: Understood now. I missed this part that sets it back to false.

https://github.com/rust-lang/cargo/blob/9d5c1497f61d0e9a62e81c6f275e4750f404c420/src/cargo/ops/cargo_update.rs#L989-L994

weihanglo (approved):

PR #14578, opened 2024-09-22
WIP: Allow credential helper to provide SSH password

w4:

I'm currently working on extending docs.rs to support our internal crate registry that usually uses SSH keys for authentication. However, since our crate registry is permissioned, we don't want all users to see all crates, nor should the docs.rs instance have unfettered access to the registry. Users authenticate to the docs.rs instance using an ephemeral OAuth token, and in turn we're using that token to authenticate with the registry.

This works great for the pulling of .crate files since crates.io uses its own libgit2 instance we can hook into to provide credentials, however in the case of actually building the documentation it shells out to cargo to fetch dependencies. In this case, since Cargo uses the credential-helper as part of its authentication hooks for libgit2, we'd like to use that instead.

The credential-helper in this case is a very simple script that reads from the environment variables passed through on the `Command` invoked by docs.rs, `git config --global credential.helper '!f() { echo "username=oauth-token"; echo "password=${GIT_PASSWORD}"; }; f'`. With this, we can use our ephemeral OAuth tokens as SSH passwords.

The easiest way to test is temporarily removing your ssh keys from your agent and using username/password auth to clone a Github repository. For future readers, don't use this for that :smile:

### Additional information

This is in draft whilst I keep hacking to see if there's a way around this by other means (maybe smuggling the token in the username via ssh_config and using git-fetch-with-cli or something).

### DIFF
--- a/src/cargo/sources/git/utils.rs
+++ b/src/cargo/sources/git/utils.rs
@@ -681,6 +681,7 @@ where
             // fails we'll get called again, but we don't have another option so
             // we bail out.
             let mut attempts = 0;
+            let mut user_pass_attempted = false;
             res = f(&mut |_url, username, allowed| {
                 if allowed.contains(git2::CredentialType::USERNAME) {
                     return git2::Cred::username(&s);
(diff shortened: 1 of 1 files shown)


epage:
As this has been a draft for over a year, I'm going to go ahead and close. This by no means is a statement on the idea here.

Note that we prefer that problems and proposed solutions are agreed to in Issues before we focus on implementation on PRs.

PR #14579, opened 2024-09-23
fix(config): Don't double-warn about `$CARGO_HOME/config`

### LINKED ISSUE #14560: Duplicate warnings of `config` on Windows OS When using git-bash
### Problem

This only happend when I use the git bash and in the /tmp folder

Recently, I found the `cargo` on Windows will output the warnings of `config` twice 

1、with only `config` exists

```
cargo -vV
warning: `C:\Users\ticktockit\.cargo\config` is deprecated in favor of `config.toml`
note: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`
warning: `C:\Users\ticktockit\.cargo\config` is deprecated in favor of `config.toml`
note: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`
cargo 1.83.0-nightly (c1fa840a8 2024-08-29)
release: 1.83.0-nightly
commit-hash: c1fa840a85eca53818895901a53fae34247448b2
commit-date: 2024-08-29
host: x86_64-pc-windows-msvc
libcurl: 8.9.0-DEV (sys:0.4.74+curl-8.9.0 vendored ssl:Schannel)
os: Windows 10.0.22631 (Windows 11 CoreCountrySpecific) [64-bit]
``` 

2. both `config` and `config.toml` exist

```
cargo -vV
warning: both `C:\Users\ticktockit\.cargo\config` and `C:\Users\ticktockit\.cargo\config.toml` exist. Using `C:\Users\ticktockit\.cargo\config`
warning: both `C:\Users\ticktockit\.cargo\config` and `C:\Users\ticktockit\.cargo\config.toml` exist. Using `C:\Users\ticktockit\.cargo\config`
cargo 1.83.0-nightly (c1fa840a8 2024-08-29)
release: 1.83.0-nightly
commit-hash: c1fa840a85eca53818895901a53fae34247448b2
commit-date: 2024-08-29
host: x86_64-pc-windows-msvc
libcurl: 8.9.0-DEV (sys:0.4.74+curl-8.9.0 vendored ssl:Schannel)
os: Windows 10.0.22631 (Windows 11 CoreCountrySpecific) [64-bit]
```

### Steps

1. make sure the  `C:\Users\[Your PC user account]\.cargo\config` exists
2. using the `git-bash`
3. cd `/tmp`
4. run `cargo -vV` 

### Possible Solution(s)

It is not clear if this problem still exists in the latest version, but duplicate information needs to be removed

### Notes

_No response_

### Version

cargo 1.83.0-nightly (c1fa840a8 2024-08-29)
release: 1.83.0-nightly
commit-hash: c1fa840a85eca53818895901a53fae34247448b2
commit-date: 2024-08-29
host: x86_64-pc-windows-msvc
libcurl: 8.9.0-DEV (sys:0.4.74+curl-8.9.0 vendored ssl:Schannel)
os: Windows 10.0.22631 (Windows 11 CoreCountrySpecific) [64-bit]

linyihai:

The core requirements for this bug are
- You have a `$CARGO_HOME/.config`
- Your project is inside of `$HOME`

We have a check to make sure we don't double-walk `$CARGO/config` but
that check is *after* we warn about there being no `.toml` extension.

To fix this, we just need to move that check outside of the file lookup.
This required changing the `seen` check from checking whether walked the
config file to checking if we've walked the config dir.  As we only have
one file per directory, this should work.

Fixes #14560

test commit added the test, fix commit fixed the issue.

### DIFF
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -1581,12 +1581,12 @@ impl GlobalContext {
     where
         F: FnMut(&Path) -> CargoResult<()>,
     {
-        let mut stash: HashSet<PathBuf> = HashSet::new();
+        let mut seen = HashSet::new();
 
         for current in paths::ancestors(pwd, self.search_stop_path.as_deref()) {
             if let Some(path) = self.get_file_path(&current.join(".cargo"), "config", true)? {
                 walk(&path)?;
-                stash.insert(path);
+                seen.insert(path);
             }
         }
 
@@ -1594,7 +1594,7 @@ impl GlobalContext {
         // in our history to be sure we pick up that standard location for
         // information.
         if let Some(path) = self.get_file_path(home, "config", true)? {
-            if !stash.contains(&path) {
+            if !seen.contains(&path) {
                 walk(&path)?;
             }
         }
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -1581,20 +1581,21 @@ impl GlobalContext {
     where
         F: FnMut(&Path) -> CargoResult<()>,
     {
-        let mut seen = HashSet::new();
+        let mut seen_dir = HashSet::new();
 
         for current in paths::ancestors(pwd, self.search_stop_path.as_deref()) {
-            if let Some(path) = self.get_file_path(&current.join(".cargo"), "config", true)? {
+            let config_root = current.join(".cargo");
+            if let Some(path) = self.get_file_path(&config_root, "config", true)? {
                 walk(&path)?;
-                seen.insert(path);
             }
+            seen_dir.insert(config_root);
         }
 
         // Once we're done, also be sure to walk the home directory even if it's not
         // in our history to be sure we pick up that standard location for
         // information.
-        if let Some(path) = self.get_file_path(home, "config", true)? {
-            if !seen.contains(&path) {
+        if !seen_dir.contains(home) {
+            if let Some(path) = self.get_file_path(home, "config", true)? {
                 walk(&path)?;
             }
         }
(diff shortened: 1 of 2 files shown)


src/cargo/util/context/mod.rs
```diff
@@ -1590,10 +1590,14 @@ impl GlobalContext {
...
 
+        // Need to check the `$CARGO_HOME/config` have been warned already.
+        let home_config_without_extention = home.join("config");
+        let already_warned = stash.contains(&home_config_without_extention);
+
         // Once we're done, also be sure to walk the home directory even if it's not
         // in our history to be sure we pick up that standard location for
         // information.
-        if let Some(path) = self.get_file_path(home, "config", true)? {
+        if let Some(path) = self.get_file_path(home, "config", !already_warned)? {
```
epage, line 1600: Isn't this issue really about us somtimes processing `CARGO_HOME`s config twice?  Seems like we should organize the fix (including comments) around that concept.

Would that be as simple as `if pwd.starts_with(&home)`?

linyihai: Should avoid processing `CARGO_HOME`s config twice, I mean.

Totally not processing the `CARGO_HOME`s if it had been walked already.

linyihai: > Would that be as simple as if pwd.starts_with(&home)?

It depends on the relation betweent `pwd` and `home`, the `starts_with` works only pwd is `home`'s subdirectory.

And the other problem is `$CARGO_HOME` can be change by env, so we must `join` the actual `home` to retrieve the  config we had walked.

epage: I had an idea for another iteration and decided to try it out myself rather than trying to describe it and found I misunderstood this!  I thought it was a problem with what file gets put in `seen`.  It turns out its because `seen` isn't covering enough.  Reusing that check by extending the scope it covers fixes this.  I've pushed that to the branch and would be interested in your thoughts on it.

linyihai: > Your project is inside of `$HOME`

This is exacly the root reason.  

I compiled the branch on local and run the fixed `cargo`, that seems the problem was addressed.

PR #14581, opened 2024-09-23
Lint against unexpected cfgs in `[target.'cfg(...)']`

Urgau:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/pull/14581)*

Since https://github.com/rust-lang/cargo/pull/13571, we now lint by default on unexpected cfgs in Rust code, but one area that is missing are cfg expressions in `Cargo.toml` and `.cargo/config.toml`. This PR address this missing piece by (unstably) introducing `-Zcheck-target-cfgs` based on `rustc --print=check-cfg`.

The feature ~~voluntarily does not always follow `RUSTFLAGS` (to avoid mistakes with one-off and for consistency)~~ but instead follows Rust `unexpected_cfgs` lint configuration:

```toml
[target.'cfg(foo)'.dependencies] # will not get linted about
cfg-if = "1.0"

[lints.rust.unexpected_cfgs]
level = "warn"
check-cfg = ['cfg(foo)'] # because it's mark as expected here
```

In terms of implementation, one global `--print=check-cfg` is retrieved and fetched, packages that have their own `check-cfg` inputs have each their own `--print=check-cfg`, otherwise the expected cfgs will get mixed up. Each `--print=check-cfg` is cached and only done if necessary.

As asked I tried putting everything into small commits, they may not reflect the de-coupling possible as they follow the implementation but they are still IMO very useful. However I strongly recommand doing a first-pass of the complete diff first.

In terms of tests, I added some combinations that I though would be useful, I can add more if desired.

### Additional information

Documentation for [`--print=check-cfg`](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/print-check-cfg.html). It was one of motivation of the compiler MCP https://github.com/rust-lang/compiler-team/issues/743.

r? @epage


crates/cargo-platform/src/cfg.rs
```diff
@@ -89,6 +89,19 @@ impl CfgExpr {
             CfgExpr::Value(ref e) => cfg.contains(e),
         }
     }
+
+    pub fn fold<E>(&self, f: &impl Fn(&Cfg) -> Result<(), E>) -> Result<(), E> {
```
epage: This doesn't align with how `fold` is used in [`Iterator::fold`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold) and found that confusing when trying to understand this.

It might also be good to have a comment explaining the intended semantics.

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -56,6 +57,80 @@ pub struct TargetInfo {
     pub rustdocflags: Rc<[String]>,
 }
 
+/// Check Config (aka `--print=check-cfg` representation)
+#[derive(Debug, Default, Clone)]
+pub struct CheckCfg {
```
epage, line 62: Should this live in `cargo-platform`?

A valid answer can "maybe later" so we can keep it close for now as we figure out what this should look like

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -56,6 +57,80 @@ pub struct TargetInfo {
...
+        } else {
+            if let Some((name, val)) = line.split_once('=') {
+                let name = name.to_string();
+
+                // strip "" around the value, e.g. "linux" -> linux
+                let val = val
+                    .strip_prefix('"')
+                    .map(|v| v.strip_suffix('"'))
+                    .flatten()
+                    .unwrap_or(val);
```
epage, line 130: Can a value have quotes in it?

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -56,6 +57,80 @@ pub struct TargetInfo {
...
+
+                self.expecteds
+                    .entry(name)
+                    .and_modify(|v| match v {
+                        ExpectedValues::Some(_) if val.is_empty() => {}
+                        ExpectedValues::Some(_) if val == "any()" => {
```
epage, line 136: Should we have a constant for `any()`?

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -56,6 +57,80 @@ pub struct TargetInfo {
...
+                    .or_insert_with(|| {
+                        if val.is_empty() {
+                            // no value to add
+                            ExpectedValues::Some(Default::default())
+                        } else if val == "any()" {
+                            ExpectedValues::Any
+                        } else {
+                            let values = HashSet::from([Some(val.to_string())]);
+                            ExpectedValues::Some(values)
+                        }
```
epage, line 153: I almost feel like we should centralize the meaning of this and then have an option to merge it into an existing entry, single-sourcing how we interpret  these fields

epage: That could possibly help consolidate with the "no `=`" case as well

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -56,6 +57,80 @@ pub struct TargetInfo {
...
+                        ExpectedValues::Some(v) => {
+                            v.insert(Some(val.to_string()));
+                        }
+                        ExpectedValues::Any => {}
+                    })
+                    .or_insert_with(|| {
+                        if val.is_empty() {
+                            // no value to add
+                            ExpectedValues::Some(Default::default())
```
epage, line 147: I feel like I'd prefer us to be consistent on whether we use `None` or `Default::default()` so we can more easily correlate the code

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -56,6 +57,80 @@ pub struct TargetInfo {
...
+#[derive(Debug, Clone)]
+pub enum ExpectedValues<T> {
+    /// List of expected values
+    Some(HashSet<Option<T>>),
+    /// All values expected
+    Any,
+}
+
+impl CheckCfg {
+    fn process_line(&mut self, line: &str) {
```
epage, line 118: As this re-implements some rustc logic, are there any unit tests they have that would be worth pulling over to make sure we conform on what any of this means/

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -990,6 +993,35 @@ impl<'gctx> RustcTargetData<'gctx> {
...
+                .cached_output(&process, 0)
+                .with_context(|| "failed to run `rustc` to learn about check-cfg information")?;
+
+            let lines = output.lines();
+
+            lines.fold(CheckCfg::default(), |mut check_cfg, line| {
+                check_cfg.process_line(line);
+                check_cfg
+            })
+        };
```
epage: > The feature voluntarily does not follow RUSTFLAGS (to avoid mistakes with one-off and for consistency) but instead follows Rust unexpected_cfgs lint configuration:

Could you clarify what you mean by this?

Calling rustc is "expensive" for no-op builds which is especially important for cargo-script.  Yes, this is cached, but I'd like to better understand why we need this over adding it to another call.

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -990,6 +993,35 @@ impl<'gctx> RustcTargetData<'gctx> {
...
+
+            // Removes `FD_CLOEXEC` set by `jobserver::Client` to pass jobserver
+            // as environment variables specify.
+            if let Some(client) = gctx.jobserver_from_env() {
+                process.inherit_jobserver(client);
+            }
+
+            let (output, _error) = rustc
+                .cached_output(&process, 0)
+                .with_context(|| "failed to run `rustc` to learn about check-cfg information")?;
```
epage: Note: I've not yet inspected this for alignment with our other rustc calls

tests/testsuite/cfg.rs
```diff
@@ -515,9 +515,73 @@ fn exclusive_dep_kinds() {
...
 [COMPILING] foo v0.1.0 ([ROOT]/foo)
 error[E0463]: can't find crate for `bar`
 ...
 "#]])
         .run();
 }
+
+#[cargo_test]
+fn unexpected_cfgs_target() {
+    let p = project()
```
epage: Any new tests like this should be added in commits before the behavior change and that commit should pass.

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -476,6 +480,49 @@ pub fn create_bcx<'a, 'gctx>(
...
+            let target_cfgs = gctx.target_cfgs()?;
+            for (key, _fields) in target_cfgs {
+                if !key.starts_with("cfg(") || !key.ends_with(')') {
+                    continue;
+                }
+
+                let cfg_expr = &key[4..key.len() - 1];
+                let Ok(cfg_expr) = CfgExpr::from_str(cfg_expr) else {
+                    continue;
+                };
```
epage, line 498: Thats a
```rust
let Some(key) = key.strip_prefix("cfg(") else { continue ; };
let Some(key) = key.strip_suffix(")") else { continue ; };
```

Even better is to not re-implement this parser and instead use `Platform::from_str`

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -476,6 +480,49 @@ pub fn create_bcx<'a, 'gctx>(
             .extend(args);
     }
 
+    {
+        let check_cfg = target_data.check_cfg();
+
+        if check_cfg.exhaustive {
+            let target_cfgs = gctx.target_cfgs()?;
```
epage: Is there a reason we are doing this here instead of in our lint phase?

https://github.com/rust-lang/cargo/blob/01e1ab55dd7f391cef809c05a50030ae3bb018d8/src/cargo/core/workspace.rs#L1172-L1249

We could even tie it into our unstable linting system so people can turn it off.

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -542,6 +589,52 @@ where `<compatible-ver>` is the latest version supporting rustc {rustc_version}"
...
+            None => {
+                gctx.shell().warn(format!(
+                    "{prefix}unexpected `cfg` condition name: `{name}`{cfg} in `[target.'cfg({cfg_expr})'{suffix}]`",
+                    cfg = if value.is_some() {
+                        format!(" for `{cfg}`")
+                    } else {
+                        format!("")
+                    }
+                ))?;
+            }
```
epage: As this aligns closely with our new linting system, I'm assuming @Muscraft would prefer these to be using annotate-snippets

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -994,32 +1033,7 @@ impl<'gctx> RustcTargetData<'gctx> {
         };
 
         let check_cfg = if gctx.cli_unstable().check_target_cfgs {
-            let mut process = rustc.workspace_process();
```
epage: > Use specific check-cfg args if provided for each unit 

This appears to be a refactor combined with a fix?  Could you please split these out so its easier to tell what you are referring to?

src/doc/src/reference/unstable.md
```diff
@@ -1728,6 +1729,29 @@ When in doubt, you can discuss this in [#14520](https://github.com/rust-lang/car
 - powershell:
   Add `CARGO_COMPLETE=powershell cargo +nightly | Invoke-Expression` to `$PROFILE`.
 
+## check-target-cfgs
```
epage, line 1732: This should be in the commit it was added in

crates/cargo-platform/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-platform"
-version = "0.1.8"
+version = "0.1.9"
```
epage, line 3: Please move this to the commit before changing `cargo-platform`

src/cargo/core/compiler/build_context/target_info.rs
epage: Also, why are we moving code within a PR rather than making it that way in the first place?

Note: I question moving this as I assume `CheckCfg` would eventually live in `cargo-platform`

epage:
This probably could have used a conversation with the Cargo team first to ensure alignment with where we are going with lints.  If there are questions about the direction I'm pointing this towards, feel free to reach out!

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -493,14 +497,36 @@ pub fn create_bcx<'a, 'gctx>(
...
-                warn_on_unexpected_cfgs(gctx, check_cfg, &cfg_expr, None, "")?;
+                warn_on_unexpected_cfgs(
+                    gctx,
+                    check_cfg,
+                    &cfg_expr,
+                    LintLevel::Warn,
+                    &mut error_count,
+                    None,
+                    "",
+                )?;
```
epage, line 508: The fact that this isn't subject to lint config calls into question whether we should warn.  We are very cautious of adding warnings that cannot be suppressed in some way.

Does this correspond with config?

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -542,6 +629,77 @@ where `<compatible-ver>` is the latest version supporting rustc {rustc_version}"
...
 
+fn get_specific_check_cfg_for_unit<'a>(
+    gctx: &GlobalContext,
+    rustc: &Rustc,
+    unit: &Unit,
+) -> CargoResult<Option<CheckCfg>> {
+    let lint_rustflags = unit.pkg.manifest().lint_rustflags();
+
+    if lint_rustflags.iter().any(|a| a == "--check-cfg") {
+        Ok(Some(CheckCfg::new(gctx, rustc, Some(lint_rustflags))?))
```
epage, line 640: I don't feel comfortable having us parse a command line

epage: I guess we don't parse it but forward it to rustc but then we are conditionally including unrelated flags in the call to rustc which feels odd and potentially brittle

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -493,14 +497,36 @@ pub fn create_bcx<'a, 'gctx>(
...
+                            lint_level = unexpected_cfgs.level().into();
+                            max_lint_level = std::cmp::max(max_lint_level, lint_level);
+                        }
+                    }
+                }
+
+                if lint_level == LintLevel::Allow {
+                    continue;
+                }
+
```
epage, line 529: This incorrectly implements lint control in that its not subject to any groups which cargo currently has no knowledge of.

This leads into a more fundamental question we've been asking about lint control: when should we act on lints for `rust` / others vs use our own?

Examples
- unknown_lint
- snake_case, etc

This also ties into the question of how to handle clippy-like cargo lints, whether to be in cargo's namespace or clippy's.

src/cargo/core/compiler/build_context/target_info.rs
Urgau: Idk, but one thing to keep in mind is that `--print=check-cfg` is still unstable and I don't if we should expose unstable rustc parts in `cargo-platform`.

src/cargo/core/compiler/build_context/target_info.rs
Urgau: Yes,

```
$ rustc a.rs --cfg "aa=\"aa\\\"\"" --print cfg
aa="aa""
```

src/cargo/core/compiler/build_context/target_info.rs
Urgau: `rustc` currently only has unit tests for `--check-cfg` to `--print=check-cfg`, but not the inverse.
[tests/run-make/print-check-cfg/rmake.rs](https://github.com/rust-lang/rust/blob/76ed7a1fa40c3f54d3fd3f834e12bf9c932d0146/tests/run-make/print-check-cfg/rmake.rs)

src/cargo/core/compiler/build_context/target_info.rs
Urgau: I want to avoid confusing users who might add a `--check-cfg` args in `RUSTFLAGS`, thinking that it works while at the second they will remove them from `RUSTFLAGS` it will not anymore; but now that I think about it more it's already the case for `rustc`, so this would be worse as it would work with `rustc` but not Cargo.

I will revert this behaviour and merge it into the others `--print`s.

src/cargo/core/compiler/build_context/target_info.rs
Urgau: > Also, why are we moving code within a PR rather than making it that way in the first place?

I wanted to have small commits with clear purpose instead a giant commit implementing all the logic in one go.

crates/cargo-platform/Cargo.toml
Urgau: I'm sorry but I don't understand this comment, do you want me to update the `cargo-platform` version in the same commit as the changes to it?

src/cargo/ops/cargo_compile/mod.rs
Urgau: Yeah, I don't like this either, but I don't know to which config it should depend.

Maybe we could let this a future possibility until we figure out the anwser.

src/cargo/ops/cargo_compile/mod.rs
Urgau: Yeah, it's a hack and I don't really like it either, but we just deduplicated all the logic in https://github.com/rust-lang/cargo/pull/14567, that I didn't wanted to reintroduce one that will potentially stay indefinitely and the unrelated flags are only allow/warn/deny which is fine for `--print=check-cfg`.

src/cargo/ops/cargo_compile/mod.rs
Urgau: > This incorrectly implements lint control in that its not subject to any groups which cargo currently has no knowledge of.

Yeah, the special `warnings` group, didn't think of that.

> This leads into a more fundamental question we've been asking about lint control: when should we act on lints for rust / others vs use our own?

I also asked my-self the same question and I didn't wanted to introduce a cargo specific lint for it as it would mean duplicating the `[lints.rust.unexpected_cfgs.check-cfg]` entry for users.

-----

I don't know what the right answer is, but I feel like this could wait until there is a more general policy.

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -476,6 +481,88 @@ pub fn create_bcx<'a, 'gctx>(
...
+                    check_cfg,
+                    &cfg_expr,
+                    LintLevel::Warn,
+                    &mut error_count,
+                    None,
+                    "",
+                )?;
+            }
+
+            for unit in &units {
```
epage, line 511: Another problem with doing this in compilation is that a package can have multiple build units and they share dependencies, so you can get duplicate dependency reports

src/cargo/core/compiler/build_context/target_info.rs
epage: While that flag is unstable, the formatting of check-cfg's is already stable and this is using the same format, right?

src/cargo/core/compiler/build_context/target_info.rs
epage: .... ok

src/cargo/core/compiler/build_context/target_info.rs
epage: Having this respect RUSTFLAGS would likely require this to be processed during compilation.  Properly processing RUSTFLAGS is a complex, expensive operation as it is fixed point as it evaluates how new `--cfg`s opt-in more RUSTFLAGS.

src/cargo/core/compiler/build_context/target_info.rs
epage: I believe thats separate from what I'm asking.  The function was added in one commit and then moved in another commit as part of a refactor to unblock feature work (so not quite atomic).  If it was instead in that new location in the original commit, I don't think anything would be negatively affected (if we keep it in that location, which is dependent on other factors).

crates/cargo-platform/Cargo.toml
epage: Or before the commit with the changes.

src/cargo/ops/cargo_compile/mod.rs
epage: Another option is we keep it in the unstable flag and not stabilize it with the rest until we figure out our story for linting of config.

src/cargo/core/compiler/build_context/target_info.rs
Urgau: Negative, the input of `--check-cfg` and the output of `--print=check-cfg` are different.
I choose to make the output of `--print=check-cfg` very close to `--print=cfg`, so it would be easier to parse for consumer.

Urgau: As suggested I put those in `cargo-platform` for now. [`f8179c5` (#14581)](https://github.com/rust-lang/cargo/pull/14581/commits/f8179c53c782dc10f6a6f77ec1a81d6acfb4bec8)

src/cargo/core/compiler/build_context/target_info.rs
Urgau: I completely simplified the parsing in [`f8179c5` (#14581)](https://github.com/rust-lang/cargo/pull/14581/commits/f8179c53c782dc10f6a6f77ec1a81d6acfb4bec8)

src/cargo/ops/cargo_compile/mod.rs
Urgau: I've moved it into to the lint phase/`emit_lints` as suggested. (It's now double gated ;-) `-Zcargo-lints`&`-Zcheck-target-cfgs`) [`cc5b088` (#14581)](https://github.com/rust-lang/cargo/pull/14581/commits/cc5b088b17d0f9c5b564e026bca17f9c579d5850)

src/cargo/core/compiler/build_context/target_info.rs
Urgau: `TargetInfo` as some handling of `RUSTFLAGS` so now use and it's handling to have `RUSTFLAGS`. [`18f2caf` (#14581)](https://github.com/rust-lang/cargo/pull/14581/commits/18f2caf13dd37e95720d2d9b151d7af2d2f2e520)

src/cargo/core/compiler/build_context/target_info.rs
epage: > I choose to make the output of --print=check-cfg very close to --print=cfg, so it would be easier to parse for consumer.

Is there a place that documents differences?

I worry that having check-cfg and print-cfg be different will make things more difficult, not less, as we now have two distinct ways of talking about the same kind of thing.

Urgau: Yes, the documentation of [`--print=check-cfg`](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/print-check-cfg.html), on the left is the `--check-cfg` input and on the right is the `--print=check-cfg` output.

My concern about re-using the `--check-cfg` syntax is that it's not easy to parse and would basically require every user to have a [MetaItem](https://doc.rust-lang.org/reference/attributes.html#meta-item-attribute-syntax) parser, while with the simplified output you only need to split the first `=` and strip the `"` for the value, much easier.

crates/cargo-platform/src/check_cfg.rs
```diff
@@ -0,0 +1,71 @@
+//! check-cfg
+
+use std::collections::{HashMap, HashSet};
+
+/// Check Config (aka `--print=check-cfg` representation)
+#[derive(Debug, Default, Clone)]
+pub struct CheckCfg {
```
epage, line 7: So this isn't actually a `check-cfg` parser but a `print-check-cfg` parser, right?  Seems like the name should convey that.

crates/cargo-platform/src/check_cfg.rs
```diff
@@ -0,0 +1,71 @@
...
+        let name: String;
+
+        if let Some((n, val)) = line.split_once('=') {
+            name = n.to_string();
+
+            if val == "any()" {
+                value_any_specified = true;
+            } else if val.is_empty() {
+                // no value to add
+                // strip "" around the value, e.g. "linux" -> linux
```
epage, line 42: This comment isn't quite in the right place.

Maybe just split the stripping out to a function would help

crates/cargo-platform/src/check_cfg.rs
```diff
@@ -0,0 +1,71 @@
...
+            } else if let Some(val) = val.strip_prefix('"').map(|v| v.strip_suffix('"')).flatten() {
+                value.insert(Some(val.to_string()));
+            } else {
+                // missing quotes and non-empty, ignore
+                return;
+            }
```
epage, line 48: Should this API report back  a parse error in some form?

crates/cargo-platform/src/check_cfg.rs
```diff
@@ -0,0 +1,71 @@
...
+
+        if let Some((n, val)) = line.split_once('=') {
+            name = n.to_string();
+
+            if val == "any()" {
+                value_any_specified = true;
+            } else if val.is_empty() {
+                // no value to add
+                // strip "" around the value, e.g. "linux" -> linux
+            } else if let Some(val) = val.strip_prefix('"').map(|v| v.strip_suffix('"')).flatten() {
```
epage, line 43: Isn't `.map(_).flatten()` just a `.and_then(_)`?

crates/cargo-platform/src/check_cfg.rs
```diff
@@ -0,0 +1,71 @@
...
+            value.insert(None);
+        }
+
+        self.expecteds
+            .entry(name)
+            .and_modify(|v| match v {
+                ExpectedValues::Some(v) if !value_any_specified => {
+                    v.extend(value.clone());
+                }
+                ExpectedValues::Some(_) => *v = ExpectedValues::Any,
```
epage, line 60: Generally people find it easier to read positive logic over negative logic.  If you just swapped these two lines, I think you can remove `!` and have the same result.

crates/cargo-platform/tests/test_check_cfg.rs
```diff
@@ -0,0 +1,79 @@
...
+    assert_eq!(
+        *check_cfg.expecteds.get("cfg_c").unwrap(),
+        ExpectedValues::Any
+    );
+
+    check_cfg.process_line("cfg_c");
+    assert_eq!(
+        *check_cfg.expecteds.get("cfg_c").unwrap(),
+        ExpectedValues::Any
+    );
```
epage, line 40: Does `check-cfg` not allow you to specify that a `--cfg` can be either `name` or `name=value`?

crates/cargo-platform/tests/test_check_cfg.rs
```diff
@@ -0,0 +1,79 @@
...
+        ExpectedValues::Some(HashSet::from([Some("double_quote\"".to_string())]))
+    );
+
+    check_cfg.process_line("cfg_e=any()"); // this shouldn't be possible but better handle this
+                                           // correctly anyway
+    assert_eq!(
+        *check_cfg.expecteds.get("cfg_e").unwrap(),
+        ExpectedValues::Any
+    );
+}
```
epage, line 79: Ideally, the weird cases would be named in the test name, calling out what behavior you are trying to show.

In this case, two unrelated behaviors are being tested, building on each other, which is harder to follow and brittle

crates/cargo-platform/src/check_cfg.rs
```diff
@@ -0,0 +1,71 @@
...
+            if val == "any()" {
+                value_any_specified = true;
+            } else if val.is_empty() {
+                // no value to add
+                // strip "" around the value, e.g. "linux" -> linux
+            } else if let Some(val) = val.strip_prefix('"').map(|v| v.strip_suffix('"')).flatten() {
+                value.insert(Some(val.to_string()));
+            } else {
+                // missing quotes and non-empty, ignore
+                return;
```
epage, line 47: this is another "weird" case that isn't tested

src/cargo/core/compiler/build_context/target_info.rs
epage: I'm concerned we may add support for something that isn't expressible in this syntax, e.g. https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618#additional-globalsvalues-validation-rules-24

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -253,16 +253,29 @@ impl TargetInfo {
...
-                .with_context(|| {
-                    format!(
-                        "failed to parse the cfg from `rustc --print=cfg`, got:\n{}",
-                        output
-                    )
-                })?;
+            let cfg = {
+                let mut res = Vec::new();
+                loop {
+                    match lines.next() {
```
epage, line 269: I think
```rust
while let Some(line) = lines.next() {
}
if res.is_empty() {
}
```
would be clearer in intent and reduce rightward shift

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -253,16 +253,29 @@ impl TargetInfo {
...
+                                    "failed to parse the cfg from `rustc --print=cfg`, got:\n{}",
+                                    output
+                                )
+                            })?;
+                            if TargetInfo::not_user_specific_cfg(&cfg) {
+                                res.push(cfg);
+                            }
+                        }
+                        None if res.is_empty() => {
+                            return error_missing_print_output("cfgs", &process, &output, &error)
```
epage, line 283: I'm not seeing this in the old code.  What is this replacing?

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -43,6 +43,8 @@ pub struct TargetInfo {
     crate_types: RefCell<HashMap<CrateType, Option<(String, String)>>>,
     /// `cfg` information extracted from `rustc --print=cfg`.
     cfg: Vec<Cfg>,
+    /// `CheckCfg` informations extracted from `rustc --print=check-cfg`.
+    check_cfg: Option<CheckCfg>,
```
epage, line 47: Please call this something else if its not `check-cfg`

src/cargo/core/compiler/build_context/target_info.rs
epage: imo the name `--print=check-cfg` needs to change because that is not what its doing (but printing another format with the same intent) and giving it this name is confusing as I look through the PR.  I know in the context of this PR what it means **now** but in viewing the code as someone without that context, that isn't clear at all.

Similarly, this is terrible UI design for rustc's command-line.  The user has to not just read the docs but carefully parse a bullet list to try to understand that this is not check-cfg thats being printed.

tests/testsuite/cfg.rs
```diff
@@ -566,6 +566,7 @@ fn unexpected_cfgs_target() {
 
     p.cargo("check -Zcheck-target-cfgs")
         .masquerade_as_nightly_cargo(&["requires -Zcheck-target-cfgs"])
+        // FIXME: We should warn on multiple cfgs
```
epage: How is this being tracked?

src/cargo/util/lints.rs
```diff
@@ -880,6 +882,136 @@ pub fn unused_dependencies(
...
 
+pub fn unexpected_target_cfgs(
+    ws: &Workspace<'_>,
+    pkg: &Package,
+    path: &Path,
+    error_count: &mut usize,
+    gctx: &GlobalContext,
+) -> CargoResult<()> {
+    // FIXME: This method is broken is several ways, it doesn't take into account
+    // `rustc` flags 'via `RUSTFLAGS`), nor the possible lints groups, ...
```
epage, line 893: How are we tracking this?

src/cargo/util/lints.rs
```diff
@@ -880,6 +882,136 @@ pub fn unused_dependencies(
...
+    let lint_level = get_unexpected_cfgs_lint_level(pkg).unwrap_or(LintLevel::Warn);
+
+    if lint_level == LintLevel::Allow {
+        return Ok(());
+    }
+
+    let rustc = gctx.load_global_rustc(Some(ws))?;
+    // FIXME: While it doesn't doesn't really matter for `--print=check-cfg`, wee should
+    // still pass the actual requested targets instead of an empty array so that the
+    // target info can be de-duplicated.
```
epage, line 971: How are we tracking this?

tests/testsuite/cfg.rs
epage: Looks like this gets resolved before the end of the PR

src/cargo/util/lints.rs
```diff
@@ -880,6 +882,111 @@ pub fn unused_dependencies(
...
+                    ))?;
+                }
+                _ => { /* not unexpected */ }
+            }
+            Ok(())
+        })
+    }
+
+    let rustc = gctx.load_global_rustc(Some(ws))?;
+    // FIXME: While it doesn't doesn't really matter for `--print=check-cfg`, wee should
```
epage, line 969: `wee` -> `we`

src/cargo/util/lints.rs
```diff
@@ -889,6 +889,19 @@ pub fn unexpected_target_cfgs(
...
+    fn get_unexpected_cfgs_lint_level(pkg: &Package) -> Option<LintLevel> {
+        if let Ok(Some(lints)) = pkg.manifest().normalized_toml().normalized_lints() {
+            if let Some(rust_lints) = lints.get("rust") {
+                if let Some(unexpected_cfgs) = rust_lints.get("unexpected_cfgs") {
+                    return Some(unexpected_cfgs.level().into());
+                }
+            }
+        }
+        None
+    }
```
epage, line 903: Is there a reason this isn't implemented like our other lints?

src/cargo/util/lints.rs
```diff
@@ -976,6 +976,20 @@ pub fn unexpected_target_cfgs(
         return Ok(());
     };
 
+    // If we have extra `--check-cfg` args comming from the lints config, we need to
+    // refetch the `--print=check-cfg` with those extra args.
+    let lint_rustflags = pkg.manifest().lint_rustflags();
+    let check_cfg = if lint_rustflags.iter().any(|a| a == "--check-cfg") {
+        Some(target_info.check_cfg_with_extra_args(gctx, &rustc, lint_rustflags)?)
```
epage, line 983: Is there a reason we are doing both a "global" and " extra_args" calls rather than just the latter?

src/cargo/util/lints.rs
```diff
@@ -880,6 +882,111 @@ pub fn unused_dependencies(
...
 }
 
+pub fn unexpected_target_cfgs(
+    ws: &Workspace<'_>,
+    pkg: &Package,
+    path: &Path,
+    error_count: &mut usize,
+    gctx: &GlobalContext,
+) -> CargoResult<()> {
+    fn warn_on_unexpected_cfgs(
```
epage, line 905: Stylistically, we are not using inner function for any of our other lints that do similar processing and ... that argument is a big red flag to me.

src/cargo/util/lints.rs
```diff
@@ -976,6 +976,20 @@ pub fn unexpected_target_cfgs(
...
+    let check_cfg = if lint_rustflags.iter().any(|a| a == "--check-cfg") {
+        Some(target_info.check_cfg_with_extra_args(gctx, &rustc, lint_rustflags)?)
+    } else {
+        None
+    };
+    let check_cfg = check_cfg.as_ref().unwrap_or(&global_check_cfg);
+
+    if !check_cfg.exhaustive {
+        return Ok(());
+    }
```
epage, line 991: Why was this moved in this commit, rather than always being in this location?

src/cargo/util/lints.rs
```diff
@@ -880,6 +882,111 @@ pub fn unused_dependencies(
...
+        lint_level: LintLevel,
+        error_count: &mut usize,
+        path: Option<&Path>,
+        suffix: &str,
+    ) -> CargoResult<()> {
+        let prefix = if let Some(path) = path {
+            format!("{path}: ", path = path.display())
+        } else {
+            "".to_string()
+        };
```
epage, line 918: Why does this deal with `Option<Path>` and why are we creating this before we even know if we have a lint?

src/cargo/util/lints.rs
```diff
@@ -880,6 +882,111 @@ pub fn unused_dependencies(
...
+        let mut emit_lint = |msg| match lint_level {
+            LintLevel::Warn => gctx.shell().warn(msg),
+            LintLevel::Deny | LintLevel::Forbid => {
+                *error_count += 1;
+                gctx.shell().error(msg)
+            }
+            LintLevel::Allow => Ok(()),
+        };
+
+        cfg_expr.try_for_each(|cfg| -> CargoResult<()> {
```
epage, line 929: Why a `for_each` rather than just a `iter` and using a `for` loop?

src/cargo/util/lints.rs
```diff
@@ -880,6 +882,111 @@ pub fn unused_dependencies(
...
+
+            match check_cfg.expecteds.get(name) {
+                Some(ExpectedValues::Some(values)) if !values.contains(&value) => {
+                    let value = if let Some(ref value) = value {
+                        value
+                    } else {
+                        "(none)"
+                    };
+                    emit_lint(format!(
+                        "{prefix}unexpected `cfg` condition value: `{value}` for `{cfg}` in `[target.'cfg({cfg_expr})'{suffix}]`"
```
epage, line 943: Not seeing a `(none)` in a test case but in my looking at this string, it feels weird to have a
```
some_path/Cargo.toml: unexpected `cfg` condition value: `(none)` for `unix` in `[target.'cfg(unix="foo"`.dependencies]`
```
specifically having a description in quotes that implies a literal

src/cargo/util/lints.rs
```diff
@@ -880,6 +882,111 @@ pub fn unused_dependencies(
...
+
+        warn_on_unexpected_cfgs(
+            gctx,
+            &global_check_cfg,
+            cfg_expr,
+            // FIXME: We should get the lint level from `[lints.rust.unexpected_cfgs]`
+            LintLevel::Warn,
+            error_count,
+            Some(path),
+            ".dependencies",
```
epage, line 1008: 1. Why is this passed in rather than hard coded?
2. Why is it passed in in a dotted form?

src/cargo/util/lints.rs
epage: (or whatever `(none)` maps to)

crates/cargo-platform/src/check_cfg.rs
Urgau: The `CheckCfg` struct is representing the internal of check-cfg, it doesn't really have anything to do with how it's parsed. (The compiler is using the same representation [internaly](https://doc.rust-lang.org/stable/nightly-rustc/rustc_session/config/struct.CheckCfg.html))

It *could* be used to parse the `--check-cfg` without an issue.

Instead of renaming the struct what about renaming the parsing method instead?
`CheckCfg::process_line` -> `CheckCfg::parse_print_check_cfg_line`?

src/cargo/util/lints.rs
Urgau: I think that would require having generators (nightly-only) or a very heavy manual state machine in a `CfgExpr`-iterator.

Which I don't is something we want for maintainability.

src/cargo/core/compiler/build_context/target_info.rs
Urgau: I following what was done above for the `--print=split-debuginfo`, I think it's a safe guard just in case there is no cfgs. I can remove it if desired.

src/cargo/util/lints.rs
Urgau: The "global" call is always done, cached and already contains the `--print=check-cfg`, so getting it is mostly "free", and even if we don't use the print-check-cfg it parsed it is still useful as it gives us all the `RUSTFLAGS` args that we can just re-use for free to the "extra_args" case.

crates/cargo-platform/src/check_cfg.rs
epage: > Instead of renaming the struct what about renaming the parsing method instead?

That might work

src/cargo/util/lints.rs
epage: I assume the answer then is "its walking a recursive data structure and its easier to write it that way".  btw I think then the name `walk` would better clarify intent.

src/cargo/core/compiler/build_context/target_info.rs
epage: In the wider context of how this is parsing, I guess the current structure makes sense.

src/cargo/core/compiler/build_context/target_info.rs
epage: A behavior change like that should probably be called out.

Is there ever a reason rustc will print nothing?

src/cargo/util/lints.rs
epage: If we don't need to ask for it from rustc twice, we probably shouldn't. Removing the global check_cfg seems like  it would simplify this overall PR as we no longer need to interject ourselves into the regular rustc query, even if we still need it for processing rustflags.

src/cargo/util/lints.rs
Urgau: Yeah, I also realized that none of the lints are using inner functions, but putting this function at the same level felt even weirder. What about putting all of this logic in a submodule?

src/cargo/util/lints.rs
Urgau: Well we don't really ask for rustc twice, we are just re-using a already available information for when there is no `[lints.rust.unexpected_cfgs.check-cfg]` specified, which I think is the vast majority of projects.

The same "global" call will be used for cfgs, split-debuginfo support, ...

-----

Schematically, if we have 3 packages, we would currently do 1 `rustc` call for the "global".

Now imaging that one the package as extra check-cfg args from the lint config, with the current strategy we would only do 2 `rustc` calls, the "global" and the one specialized for the package.

If we were to not "interject ourselves into the regular rustc query" (aka "global") there would 4 `rustc` calls (the global and 1 per package) instead of 2.

-----

Btw I'm getting mixed signal with this comment and the one this [conversation](https://github.com/rust-lang/cargo/pull/14581#discussion_r1775906095):
> Calling rustc is "expensive" for no-op builds which is especially important for cargo-script. Yes, this is cached, but I'd like to better understand why we need this **over adding it to another call**.

Which is why I changed strategy to reduce the number of calls.

crates/cargo-platform/src/check_cfg.rs
Urgau: Renamed to `CheckCfg::parse_print_check_cfg_line`.

crates/cargo-platform/src/check_cfg.rs
Urgau: I made the API report error.

crates/cargo-platform/src/check_cfg.rs
Urgau: Split the stripping in it's own function for clarity.

crates/cargo-platform/src/check_cfg.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub exhaustive: bool,
+    /// List of expected cfgs
+    pub expecteds: HashMap<String, ExpectedValues>,
+}
+
+/// List of expected check-cfg values
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ExpectedValues {
+    /// List of expected values
+    Some(HashSet<Option<String>>),
```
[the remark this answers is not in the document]
Urgau: Added documentation about the representation `HashSet<Option<String>>` representation.

src/cargo/core/compiler/build_context/target_info.rs
Urgau: Revert to not error if no cfgs printed.

src/cargo/util/lints.rs
Urgau: Mistake on my side, fixed.

src/cargo/util/lints.rs
Urgau: It was coming from a previous iteration. Fixed.

src/cargo/ops/cargo_compile/mod.rs
Urgau: The last revision of this PR is now using annotate-snippets.

src/cargo/util/lints.rs
Urgau: These were coming from a previous iteration of this PR. Fixed.

src/cargo/util/lints.rs
```diff
@@ -605,6 +608,157 @@ fn output_unknown_lints(
...
+                        let help_where = format!("occurred in `{path}`", path = path.display());
+                        let diag = level.title(&*title).footer(Level::Help.title(&*help_where));
+                        gctx.shell().print_message(diag)?;
+                    }
+                }
+                _ => { /* not unexpected */ }
+            }
+            Ok(())
+        })?;
+    }
```
Muscraft, line 757: If you use the original manifest's `target` table, any `cfg` will be in its original format, so we can always get a `span`. One thing to note is that it will be the span for the whole `cfg`, not a specific part, which would be ideal. To get the particular part of a `cfg`, we would either need to "find" that part in the `cfg`, or we would need to track spans while parsing `CfgExpr` (I tested a basic implementation of this, and it does work).
```suggestion
    let manifest_path = rel_cwd_manifest_path(path, gctx);

    let Some(target) = manifest.original_toml().target.as_ref() else {
        return Ok(());
    };
    for name in target.keys() {
        if let Some(cfg_expr_orig) = name.strip_prefix("cfg(").and_then(|s| s.strip_suffix(')')) {
            let Ok(cfg_expr) = cfg_expr_orig.parse::<CfgExpr>() else {
                continue;
            };
            cfg_expr.walk(|cfg| -> CargoResult<()> {
                let (name, value) = match cfg {
                    Cfg::Name(name) => (name, None),
                    Cfg::KeyPair(name, value) => (name, Some(value.to_string())),
                };
                let cfg_str = format!("cfg({cfg_expr_orig})");
                if let Some(title) = match check_cfg.expecteds.get(name) {
                    Some(ExpectedValues::Some(values)) if !values.contains(&value) => {
                        let value = match value {
                            Some(value) => Cow::from(format!("`{value}`")),
                            None => Cow::from("(none)"),
                        };
                        Some(format!("unexpected `cfg` condition value: {}", value))
                    }
                    None => Some(format!("unexpected `cfg` condition name: `{}`", name)),
                    _ => None,
                } {
                    if let Some(span) = get_span(manifest.document(), &["target", &cfg_str], false)
                    {
                        let level = lint_level.to_diagnostic_level();
                        if lint_level == LintLevel::Forbid || lint_level == LintLevel::Deny {
                            *error_count += 1;
                        }

                        let diag = level.title(&*title).snippet(
                            Snippet::source(manifest.contents())
                                .origin(&manifest_path)
                                .annotation(level.span(span))
                                .fold(true),
                        );
                        gctx.shell().print_message(diag)?;
                    }
                }
                Ok(())
            })?;
        }
    }
```

Urgau: This looks cleaner than re-using the dependencies, as it avoids duplicate and is more easily forward-compatible. Applied your suggestion. Thanks.

As for getting the span of the particular cfg we want to lint on I think the best way would be to track the spans when parsing the `CfgExpr`/`Cfg` as you already alluded to. I think this would be good follow-up work as this PR is already quite large.

sosthene-nitrokey:
Hello,

Not sure if this is the best place to discuss this, but [the zulip stream](https://rust-lang.zulipchat.com/#narrow/stream/233931-xxx/topic/Add.20--print.3Dcheck-cfg.20compiler-team.23743) linked in the MCP is pretty empty, the MCP has already been accepted, and I don't see prior discussion on it, so I'm putting my concern here.

[We recently encountered](https://github.com/trussed-dev/interchange/pull/20) a case that would be linted by this PR if I understand correctly. We have in the `Cargo.toml`:

```toml
[target.'cfg(loom)'.dependencies]
loom = "0.5"
```

and some `cfg(loom)` in the rust code.

Rustc linted the rust code and generated the warning for that. We can silence them through the `[lints.rust.unexpected_cfgs]` section in `Cargo.toml`. If I understand correctly, with this PR, the lint would also have warned for the `[target.'cfg(loom)'.dependencies]` section, and `[lints.rust.unexpected_cfgs]` would also silence that.


My intuition was different, I would have expected `Cargo` to automatically add `cfg(loom)` to the `check-cfg` config of the lint, based on the fact that there is a dependency for this specific `cfg` that is defined.

This goes with my intuition that `Cargo.toml` would define all that can be used for the `cfg()` and minimize the amount of manual bookkeeping of what is and isn't linted, and also minimize the amount of "false positives" for this lint with the default configuration.
To me the presence of the `[target.cfg(loom)'] section implies that the `cfg(loom)` is intended to be used, so shouldn't be linted against, rather than being a possible mistake.

epage:
> My intuition was different, I would have expected Cargo to automatically add cfg(loom) to the check-cfg config of the lint, based on the fact that there is a dependency for this specific cfg that is defined.

`--cfg` / `--check-cfg` is operating at a layer of abstraction below Cargo which is why `check-cfg` is configured through either build scripts or the rust lints table.  While we support conditional dependencies, I don't expect it to be likely to infer cfg's from them, especially since the case you gave is fairly special.  IN a lot of other cases, users can benefit from being told what cfg's are supported vs accidentally creating a new one.

Currently, the only feature at Cargo's level of abstraction is `features`.  [global, mutually exclusive features](https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618) is an idea for another one.

src/doc/src/reference/unstable.md
```diff
@@ -1929,6 +1930,31 @@ For example:
...
+```toml
+[target.'cfg(foo)'.dependencies]
+cfg-if = "1.0"
+
+[lints.cargo.unexpected_cfgs]
+level = "deny"
+
+[lints.rust.unexpected_cfgs]
+level = "warn"
+check-cfg = ['cfg(foo)']
```
epage, line 1955: > I don't remember if we had settle on the lint config position, ie if we should share check-cfg = [...] from rust.unexpected_cfgs or if cargo.unexpected_cfgs should have a distinct one. I've left the sharing for now.

I would say if we have our own lint, we have our own lint config.

If anything, in light of lints being fractured like this, I would see us re-evaluating whether we should have per-lint config or a general config table.

tests/testsuite/cfg.rs
```diff
@@ -871,3 +872,255 @@ fn unexpected_cfgs_target() {
 "#]])
         .run();
 }
+
+#[cargo_test(nightly, reason = "--print=check-cfg is unstable in rustc")]
```
epage, line 905: Could we consolidate all of the tests into one commit?

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -208,6 +210,14 @@ impl TargetInfo {
             process.arg("--print=crate-name"); // `___` as a delimiter.
             process.arg("--print=cfg");
 
+            if gctx.cli_unstable().check_target_cfgs {
+                process.arg("--print=crate-name"); // `___` as a delimiter.
+                process.arg("--print=check-cfg");
+
+                process.arg("--check-cfg=cfg()"); // otherwise `--print=check-cfg` won't output
```
epage, line 217: Why is that?

src/cargo/core/compiler/build_context/target_info.rs
```diff
@@ -277,6 +292,22 @@ impl TargetInfo {
                 res
             };
 
+            let check_cfg = if gctx.cli_unstable().check_target_cfgs {
+                let mut check_cfg = CheckCfg::default();
+                check_cfg.exhaustive = true;
```
epage, line 297: API-wise, this feels dirty that we are having to tell it that the default state before parsing is `exhaustive = true`

src/cargo/util/lints.rs
```diff
@@ -605,6 +608,164 @@ fn output_unknown_lints(
...
+    let Some(global_check_cfg) = target_info.check_cfg() else {
+        return Ok(());
+    };
+
+    if !global_check_cfg.exhaustive {
+        return Ok(());
+    }
+
+    // FIXME: If the `[lints.rust.unexpected_cfgs.check-cfg]` config is set we should
+    // re-fetch the check-cfg informations with those extra args
```
epage: How are we expected to track these FIXMEs?

src/cargo/util/lints.rs
```diff
@@ -605,6 +608,164 @@ fn output_unknown_lints(
     Ok(())
 }
 
+const UNEXPECTED_CFGS: Lint = Lint {
+    name: "unexpected_cfgs",
+    desc: "lint on unexpected target cfgs",
+    groups: &[],
+    default_level: LintLevel::Warn,
+    edition_lint_opts: None,
+    feature_gate: None,
```
epage, line 617: Right now only the acquisition of `check-cfg`s from the compiler is feature gated.  The lint should be feature gated as well.

src/cargo/util/lints.rs
```diff
@@ -665,13 +665,20 @@ pub fn unexpected_target_cfgs(
         return Ok(());
     };
 
-    if !global_check_cfg.exhaustive {
+    // If we have extra `--check-cfg` args comming from the lints config, we need to
+    // refetch the `--print=check-cfg` with those extra args.
+    let lint_rustflags = pkg.manifest().lint_rustflags();
+    let check_cfg = if lint_rustflags.iter().any(|a| a == "--check-cfg") {
```
epage, line 671: I don't feel too comfortable having this stringly-typed API from our `Cargo.toml` parser to here

src/cargo/util/lints.rs
```diff
@@ -665,13 +665,20 @@ pub fn unexpected_target_cfgs(
...
+        Some(target_info.check_cfg_with_extra_args(gctx, &rustc, lint_rustflags)?)
```
epage, line 672: Why are we calling rustc for this rather than processing them ourselves?

src/cargo/core/compiler/build_context/target_info.rs
Urgau: The comment is a bit misleading, `--print=check-cfg` alone would print something, just not what we want, without `--check-cfg=cfg()` it would be like we didn't enable check-cfg checking and so `--print=check-cfg` would only return the default state (of no checking).

We also have `--check-cfg=cfg()` when generating the [`check_cfg_args`](https://github.com/rust-lang/cargo/blob/a9e8aa2c65091b520519fd1065f41d156d45f9dd/src/cargo/core/compiler/mod.rs#L1500).

src/cargo/util/lints.rs
Urgau: Me either. Looking at the flags like this is ugly, I would love to not have to do this. A better way would probably be to pass always pass `--print=check-cfg` (with a file path) in the check/build invocation of the crate, that way we only have one compilation unit. It would be a much larger change though.

src/cargo/util/lints.rs
epage: Except instead we have a parser for something not quite like `check-cfg` and have to process that.

src/cargo/core/compiler/build_context/target_info.rs
Urgau: We only know if it's exhaustive by the absence of `any()=any()` and `any()`, which we can't know inside `parse_print_check_cfg_line` since it doesn't know if we have finished processing all the lines.

What about having associated function `CheckCfg::parse_print_check_cfg_lines` (note the `s` at the end), which would return a `CheckCfg` (no Default impl anymore), this function would do the business of parsing all the lines and setting the `exhaustive` flag appropriately.

src/cargo/util/lints.rs
Urgau: `--check-cfg` and `--print=check-cfg` are different, `--check-cfg` is an input that accepts multiple ways for doing the same thing, it can evolve, and new way of doing things, for example taking one day the [shorthand `target(...)`](https://rust-lang.github.io/rfcs/3239-cfg-target.html), on the other-hand the output of `--print=check-cfg` is meant to be stable, no matter the inputs given in `--check-cfg`.

With `--print=check-cfg` each line a expected key-value (except for `any()=any()` or `any()` but those are exceptions), it's much simpler to parse and it doesn't vary if the input changes, like if introduced the [shorthand `target(...)`](https://rust-lang.github.io/rfcs/3239-cfg-target.html) to `--check-cfg`, the output would stay with the full form, no matter what.

Issue #9029, opened 2020-12-31
Open semver range does not properly unify with closed semver ranges


CAD97:
**Problem**

If a dependency's version is constrained to cover multiple semver-incompatible releases, the version selected does not unify with more specific restrictions from elsewhere in the build graph. E.g. a requirement for `>=0.0.0, <0.9.0` and a requirement for `^0.4.0` results in both version `0.4.x` and `0.8.y` being pulled in (or `0.4.x` and `0.1.0` with `-Zminimal-versions`), rather than unifying both versions to just `0.4.x`.

Manually adjusting the lockfile to use `0.4.x` in both cases works as expected; this is just an issue with the "find the best solution" codepath, not the "check the lockfile satisfies" codepath.

Originally reported by @MaximilianKoestler on URLO: https://users.rust-lang.org/t/crate-interoperability-and-3rd-party-types-in-interfaces/53431

**Steps**

Repro from @MaximilianKoestler : https://github.com/MaximilianKoestler/crate-version-testing

Using the `rgb` crate as our target, this is exemplified using just the three packages:

```toml
[package]
name = "lib_a"

[dependencies]
rgb = ">=0.4.0, <0.5.0"
```

```toml
[package]
name = "lib_b"

[dependencies]
rgb = ">=0.0.0, <0.9.0"
```

```toml
[package]
name = "bin_c"

[dependencies]
lib_a = { path = "../lib_a" }
lib_b = { path = "../lib_b" }
```

The generated lockfile:

```toml
# irrelevant parts removed and reordered
[[package]]
name = "lib_a"
dependencies = [
 "rgb 0.4.0",
]

[[package]]
name = "lib_b"
dependencies = [
 "rgb 0.8.25",
]

[[package]]
name = "bin_c"
dependencies = [
 "lib_a",
 "lib_b",
]

[[package]]
name = "rgb"
version = "0.4.0"
source = "registry+https://github.com/rust-lang/crates.io-index"

[[package]]
name = "rgb"
version = "0.8.25"
source = "registry+https://github.com/rust-lang/crates.io-index"
```

**Possible Solution(s)**

Ideally, if all constraints overlap, only a single version should be selected. Otherwise, maintaining support for old semver-incompatible versions of public dependencies (when the subset you're using didn't change) is almost meaningless, as cargo won't unify the dependency off of the most recent semver-compatible range (or first with `-Zminimal-versions`).

**Notes**

[...]

PR #5718, opened 2018-07-12
Remove a redundant heuristic

Eh2406:
This adds a test for #5529, then removes the heuristic that caused that odd behavior. 
I believe this heuristic is redundant with what the resolver already does. 

If you believe this heuristic is needed please suggest a test case and I will add it to this PR. 

This tests and closes #5529 and should close #5702.

PR #14582, opened 2024-09-23
fix(update): don't reuse a previously locked dependency if multiple major versions are compatible

### LINKED ISSUE #5529: `cargo update -p SPEC` can silently downgrade dependencies of other packages
This is kinda convoluted, but it bit me when I wasn't paying attention when updating rust's lock file.  Here's a relatively small repro.  Create a workspace with the following members and dependencies:

```
a → bitflags="0.9"
b → bitflags=">= 0.9, < 2.0"
c → bitflags="1.0"
```

`cargo update` will generate a lockfile that looks good (`a`→0.9, `b`→1.0, `c`→1.0).

`cargo update -p c` will *downgrade* `b`'s copy of bitflags to 0.9!  I'd expect it to leave it as-is.

The real-world example of this is in rust where these chains exist:
```
rustbook → ... → html5ever → .. → rand 0.3
rls → rayon → rayon-core → rand >=0.3, < 0.5
cargo → ... → rand 0.4
```
Updating cargo caused rayon's version of rand to downgrade.  I didn't use `rand` in my example above because version 0.3 is a little odd in that it also depends on rand 0.4, but doesn't seem to be important and just complicates the example.

### LINKED ISSUE #14115: unrelated dependency changed in Cargo.lock when updating another dependency
### Problem

This dependabot PR updates `lz4`, but Cargo.lock for (one of the) `opendal` is also updated. https://github.com/risingwavelabs/risingwave/pull/17207

From `cargo tree -i lz4`, they seem to be not related.

I thought it's dependabot's issue, but I can also reproduce this locally.
I also found that only modifying the `Cargo.toml` will trigger this change, but not `cargo update -p lz4`

### Steps

1. `git clone https://github.com/risingwavelabs/risingwave` (HEAD as of the post is `af8f9a5815388c6d3147bebda77a1830a05f8c7e`)
2. modify `src/storage/Cargo.toml`  `lz4` 1.24.0 -> 1.25.0 

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
❯ cargo version --verbose
cargo 1.78.0-nightly (a4c63fe53 2024-03-06)
release: 1.78.0-nightly
commit-hash: a4c63fe5388beaa09e5f91196c86addab0a03580
commit-date: 2024-03-06
host: aarch64-apple-darwin
os: Mac OS 14.4.1 [64-bit]
```

### LINKED ISSUE #14446: Cargo downgrades transitive dependency that should not change
### Problem

Our project, uv, depends on a git version of pubgrub. pubgrub depends on `rustc-hash = ">=1.0.0, <3.0.0"`. uv has dependencies with both rustc-hash 1 and rustc-hash 2. At , `Cargo.lock` in uv has:
```toml

[[package]]
name = "pubgrub"
version = "0.2.1"
source = "git+https://github.com/astral-sh/pubgrub?rev=aaef464c1b0d8eea4ff9ffaee4f3458c236d10da#aaef464c1b0d8eea4ff9ffaee4f3458c236d10da"
dependencies = [
 "indexmap",
 "log",
 "priority-queue",
 "rustc-hash 2.0.0",
 "thiserror",
]
```

Each time i bump the pubgrub revision in `Cargo.toml`, cargo downgrades the rustc version, even though pubgrub's dependencies didn't change:

```toml
[[package]]
name = "pubgrub"
version = "0.2.1"
source = "git+https://github.com/astral-sh/pubgrub?rev=388685a8711092971930986644cfed152d1a1f6c#388685a8711092971930986644cfed152d1a1f6c"
dependencies = [
 "indexmap",
 "log",
 "priority-queue",
 "rustc-hash 1.1.0",
 "thiserror",
]
```

The only way to fix this is running `cargo update -p rustc-hash@1.1.0` manually after each pubgrub update.

### Steps

* `git clone https://github.com/astral-sh/uv && cd uv && git checkout 6bd677d60d15ab950ae5466197cab4f3f2405bf4`
* Update the git revision in `Cargo.toml`, e.g. to `388685a8711092971930986644cfed152d1a1f6c`
* Update the lockfile, e.g. `cargo check`
* `git diff`:
```diff
diff --git a/Cargo.lock b/Cargo.lock
index d90529ebcf5e..15f283c3788c 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2777,12 +2777,12 @@ dependencies = [
 [[package]]
 name = "pubgrub"
 version = "0.2.1"
-source = "git+https://github.com/astral-sh/pubgrub?rev=aaef464c1b0d8eea4ff9ffaee4f3458c236d10da#aaef464c1b0d8eea4ff9ffaee4f3458c236d10da"
+source = "git+https://github.com/astral-sh/pubgrub?rev=388685a8711092971930986644cfed152d1a1f6c#388685a8711092971930986644cfed152d1a1f6c"
 dependencies = [
  "indexmap",
  "log",
  "priority-queue",
- "rustc-hash 2.0.0",
+ "rustc-hash 1.1.0",
  "thiserror",
 ]
 
diff --git a/Cargo.toml b/Cargo.toml
index 94a65c8d0168..933484fe9740 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,7 +112,7 @@ pathdiff = { version = "0.2.1" }
 petgraph = { version = "0.6.4" }
 platform-info = { version = "2.0.2" }
 proc-macro2 = { version = "1.0.86" }
-pubgrub = { git = "https://github.com/astral-sh/pubgrub", rev = "aaef464c1b0d8eea4ff9ffaee4f3458c236d10da" }
+pubgrub = { git = "https://github.com/astral-sh/pubgrub", rev = "388685a8711092971930986644cfed152d1a1f6c" }
 pyo3 = { version = "0.21.0" }
 pyo3-log = { version = "0.10.0" }
 quote = { version = "1.0.36" }
```

### Possible Solution(s)

During version resolution, cargo should use the version previously used for the package as preference, not another version in the lockfile, or if it has to use a version from the lockfile, it should use the higher version.

### Notes

_No response_

### Version

```text
cargo 1.80.1 (376290515 2024-07-16)
release: 1.80.1
commit-date: 2024-07-16
```

stormshield-kg:

Fixes https://github.com/rust-lang/cargo/issues/5529.
Fixes https://github.com/rust-lang/cargo/issues/14115.
Fixes https://github.com/rust-lang/cargo/issues/14446.

This first commit adds a test documenting the current behavior, and the second commit implements the fix and updates the test.

### DIFF
--- a/src/cargo/core/registry.rs
+++ b/src/cargo/core/registry.rs
@@ -12,6 +12,7 @@
 use std::collections::{HashMap, HashSet};
 use std::task::{ready, Poll};
 
+use crate::core::resolver::VersionPreferences;
 use crate::core::PackageSet;
 use crate::core::{Dependency, PackageId, SourceId, Summary};
 use crate::sources::config::SourceConfigMap;
@@ -63,6 +64,12 @@ pub trait Registry {
 
     /// Block until all outstanding [`Poll::Pending`] requests are [`Poll::Ready`].
     fn block_until_ready(&mut self) -> CargoResult<()>;
+
+    /// Get version preferences.
+    fn version_prefs(&self) -> &VersionPreferences;
+
+    /// Set version preferences.
+    fn set_version_prefs(&mut self, version_prefs: VersionPreferences);
 }
 
 /// This structure represents a registry of known packages. It internally
@@ -131,6 +138,9 @@ pub struct PackageRegistry<'gctx> {
     /// This is constructed during calls to [`PackageRegistry::patch`],
     /// along with the `patches` field, thoough these entries never get locked.
     patches_available: HashMap<CanonicalUrl, Vec<PackageId>>,
+
+    /// Version preferences.
+    version_prefs: VersionPreferences,
 }
 
 /// A map of all "locked packages" which is filled in when parsing a lock file
@@ -209,6 +219,7 @@ impl<'gctx> PackageRegistry<'gctx> {
             patches: HashMap::new(),
             patches_locked: false,
             patches_available: HashMap::new(),
+            version_prefs: VersionPreferences::default(),
         })
     }
 
@@ -503,7 +514,12 @@ impl<'gctx> PackageRegistry<'gctx> {
         for summaries in self.patches.values_mut() {
             for summary in summaries {
                 debug!("locking patch {:?}", summary);
-                *summary = lock(&self.locked, &self.patches_available, summary.clone());
+                *summary = lock(
+                    &self.locked,
+                    &self.patches_available,
+                    summary.clone(),
+                    &self.version_prefs,
+                );
             }
         }
         self.patches_locked = true;
@@ -581,7 +597,12 @@ impl<'gctx> PackageRegistry<'gctx> {
     /// through.
     pub fn lock(&self, summary: Summary) -> Summary {
         assert!(self.patches_locked);
-        lock(&self.locked, &self.patches_available, summary)
+        lock(
+            &self.locked,
+            &self.patches_available,
+            summary,
+            &self.version_prefs,
+        )
     }
 
     fn warn_bad_override(
@@ -734,7 +755,12 @@ impl<'gctx> Registry for PackageRegistry<'gctx> {
                         }
                     }
                     let summary = summary.into_summary();
-                    f(IndexSummary::Candidate(lock(locked, all_patches, summary)))
+                    f(IndexSummary::Candidate(lock(
+                        locked,
+                        all_patches,
+                        summary,
+                        &self.version_prefs,
+                    )))
                 };
                 return source.query(dep, kind, callback);
             }
@@ -799,6 +825,14 @@ impl<'gctx> Registry for PackageRegistry<'gctx> {
         }
         Ok(())
     }
+
+    fn version_prefs(&self) -> &VersionPreferences {
+        &self.version_prefs
+    }
+
+    fn set_version_prefs(&mut self, version_prefs: VersionPreferences) {
+        self.version_prefs = version_prefs;
+    }
 }
 
 /// See [`PackageRegistry::lock`].
@@ -806,6 +840,7 @@ fn lock(
     locked: &LockedMap,
     patches: &HashMap<CanonicalUrl, Vec<PackageId>>,
     summary: Summary,
+    version_prefs: &VersionPreferences,
 ) -> Summary {
     let pair = locked
         .get(&(summary.source_id(), summary.name()))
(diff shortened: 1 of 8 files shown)


epage:
As a heads up, none of these issues are marked as `A-accepted` but instead `S-needs-design` (see [Issues in contrib docs](https://doc.crates.io/contrib/issues.html)).  It would generally be best to explore the high level ideas in the Issue before moving on to a PR (even if playing with an implementation locally) so we keep the solution space exploration in one place and leave the PR to focus on the implementation.  Sometimes PRs come and go for the same issue and if we explore designs in PRs, it fractures the conversation.

@Eh2406 can better speak to whether we can shortcircuit that and move forward with this.

r? @Eh2406

btw another issue with multi-major version reqs is #9029 with #13535 as a proposed change for it.

src/cargo/core/registry.rs
```diff
@@ -63,6 +64,12 @@ pub trait Registry {
 
     /// Block until all outstanding [`Poll::Pending`] requests are [`Poll::Ready`].
     fn block_until_ready(&mut self) -> CargoResult<()>;
+
+    /// Get version preferences.
+    fn version_prefs(&self) -> &VersionPreferences;
+
+    /// Set version preferences.
+    fn set_version_prefs(&mut self, version_prefs: VersionPreferences);
```
epage, line 72: For myself, I feel uncomfortable how much resolver logic lives in `PackageRegistry`, I feel even more uncomfortable adding `VersionPreferences` to it and to `trait Registry` as well.

src/cargo/core/registry.rs
```diff
@@ -892,12 +927,30 @@ fn lock(
...
-        // If anything does then we lock it to that and move on.
-        let v = locked
-            .get(&(dep.source_id(), dep.package_name()))
-            .and_then(|vec| vec.iter().find(|&&(id, _)| dep.matches_id(id)));
+        // If this dependency did not have a locked version, we check if multiple
+        // versions of the dependency already exists in the lockfile. In this case we want
+        // to do the full resolving instead of locking a specific version.
+        // Otherwise, we query all known locked packages to see if they match this dependency,
+        // and if anything does then we lock it to that and move on.
+
```
Eh2406, line 935: My instinct has long been that this heuristic should just be removed. Apparently I tried that when I was a little baby contributor new to open source in #5718. A lot has changed since then, so it might work now.

Another approach discussed in that PR is to remove this heuristic and double our calls to `resolve`. We try resolving with `--offline` and only try normal resolution after the `--offline` did not work. But now that I say that (I think, without having try it), that would lead to `cargo update` randomly updating unrelated GIT dependencies. In addition to the performance impacts.

Eh2406: After some more investigation, this heuristic seems to be loadbearing for cleaning up different ways to refer to the same git package. For example a dependency might require the default branch (or the `master` branch) but what's available in the lock file is a particular checkout. This heuristic identifies that the particular hash could match and uses it. This seems accidental to me. I think there must be a better way of handling this. If we had a `matches_with_compatible_git_source` when iterating over `locked_deps` things could be solved much more elegantly. In the time I had available I did not find one laying around, but this still seems a promising direction.

stormshield-kg: This test checks this heuristic and is not related to git dependencies:
https://github.com/rust-lang/cargo/blob/43e3fa90257d899c7281d3808d37cbf569a21587/tests/testsuite/registry.rs#L2487

Eh2406: You are correct. I forgot to mention that case. That test was added at the same time as the heuristic. Based on the PR that added it was to allow use of cargo with fewer network requests. This was a big issue at the time because we did not yet have `--offline`. So if cargo decided that you needed a network request and you did not have network access you were stuck. I have not formally discussed it with the team, but my vote would be that we just remove that test.

0xPoe:
As this PR has been stale for about 2 years and has a lot of conflicts, I would like to close it. If you want to rework this PR, please let me know, and I will help reopen it.

PR #11938, opened 2023-04-06
Remove disabled optional weak dependencies from Cargo.lock

est31:
In #10801, it was noted that Cargo treats weak dependencies just like strong ones in Cargo.lock, as in, includes them unconditionally. Cargo does not build these dependencies though, only uses them during resolution.

Cargo [has two resolvers](https://github.com/rust-lang/cargo/blob/7bf43f028ba5eb1f4d70d271c2546c38512c9875/src/cargo/core/resolver/mod.rs#L40):

* one that performs aggressive unification and is used to generate Cargo.lock files
* one that takes the enabled features on the command line into account, as well as the target platform

The implementation in #8818 only changed the latter resolver, which implemented weak dependency support for most important workflows, like determining which crates are actually built. The former resolver still keeps a list of enabled and disabled features for each crate, and the way more aggressive unification it performs isn't an issue for weak dependency support, so fundamentally one can add weak dependency support to it, too. This PR does that precisely:

* it changes the parts of the resolver that generate Cargo.lock to not include weak dependencies.
* it adds a new Cargo.lock `ResolveVersion::V4` variant, which isn't emitted by default. The idea is though that at some point in the future cargo switches to the new version by default. I'm open to feedback for different suggestions how to phase in the change, maybe via `resolver = 3`, or maybe additionally by having an unstable flag so that we can iterate before version 4 is stabilized.

Fixes #10801

PR #14583, opened 2024-09-23
test: add support for features in the sat resolver

x-hgg-x:

This PR implements the first step of https://github.com/rust-lang/cargo/pull/11938#issuecomment-1868426431.

The first commit does some refactorings, and the second commit updates the SAT resolver.

List of boolean variables in the SAT resolver:
* One variable representing the activation of each registry package.
* One variable representing the activation of each feature of a given registry package.
* In the `sat_resolve()` method, we create an additional representing the activation of the root package.

List of clauses in the SAT resolver:
* If a package feature is activated, then the package should be activated.
* No two packages with the same links set.
* No two semver compatible versions of the same package.
* For each package:
    - For each feature:
        - If the package feature is activated and it depends of another feature of the same package, then it is also activated.
        - If the package feature is activated and it depends of a dependency, then at least one of the compatible dependency package should be activated.
        - If the package feature is activated and it depends of a feature of a dependency, then the feature of a compatible dependency package should be activated only if the compatible dependency package is activated. If this is not a weak dependency feature, then at least one of the compatible dependency package should be activated.
    - For each dependency, if the package is activated and if the dependency is non-optional or has been activated, then at least one of the compatible dependency package and its required features should be activated.
* The root package has the same dependency clauses but has no features.

List of assumptions in the SAT resolver:
* The root package is activated.
* Old root packages from previous calls to `sat_resolve()` are deactivated. This is necessary since the proptest call `sat_resolve()` several times with a different root package using the same SAT resolver, and clauses relative to the root package are not removable.

### DIFF
--- a/crates/resolver-tests/src/lib.rs
+++ b/crates/resolver-tests/src/lib.rs
@@ -1,7 +1,7 @@
 #![allow(clippy::print_stderr)]
 
 use std::cmp::{max, min};
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::fmt;
 use std::fmt::Write;
 use std::sync::OnceLock;
@@ -10,12 +10,14 @@ use std::time::Instant;
 
 use cargo::core::dependency::DepKind;
 use cargo::core::resolver::{self, ResolveOpts, VersionOrdering, VersionPreferences};
-use cargo::core::Resolve;
+use cargo::core::FeatureMap;
 use cargo::core::ResolveVersion;
 use cargo::core::{Dependency, PackageId, Registry, Summary};
+use cargo::core::{FeatureValue, Resolve};
 use cargo::core::{GitReference, SourceId};
 use cargo::sources::source::QueryKind;
 use cargo::sources::IndexSummary;
+use cargo::util::interning::{InternedString, INTERNED_DEFAULT};
 use cargo::util::{CargoResult, GlobalContext, IntoUrl};
 
 use proptest::collection::{btree_map, vec};
@@ -25,7 +27,12 @@ use proptest::string::string_regex;
 use varisat::ExtendFormula;
 
 pub fn resolve(deps: Vec<Dependency>, registry: &[Summary]) -> CargoResult<Vec<PackageId>> {
-    resolve_with_global_context(deps, registry, &GlobalContext::default().unwrap())
+    Ok(
+        resolve_with_global_context(deps, registry, &GlobalContext::default().unwrap())?
+            .into_iter()
+            .map(|(pkg, _)| pkg)
+            .collect(),
+    )
 }
 
 // Verify that the resolution of cargo resolver can pass the verification of SAT
@@ -33,7 +40,7 @@ pub fn resolve_and_validated(
     deps: Vec<Dependency>,
     registry: &[Summary],
     sat_resolver: &mut SatResolver,
-) -> CargoResult<Vec<PackageId>> {
+) -> CargoResult<Vec<(PackageId, Vec<InternedString>)>> {
     let resolve =
         resolve_with_global_context_raw(deps.clone(), registry, &GlobalContext::default().unwrap());
 
@@ -66,7 +73,7 @@ pub fn resolve_and_validated(
                     }));
                 }
             }
-            let out = resolve.sort();
+            let out = collect_features(&resolve);
             assert_eq!(out.len(), used.len());
 
             if !sat_resolver.sat_is_valid_solution(&out) {
@@ -80,13 +87,21 @@ pub fn resolve_and_validated(
     }
 }
 
+fn collect_features(resolve: &Resolve) -> Vec<(PackageId, Vec<InternedString>)> {
+    resolve
+        .sort()
+        .iter()
+        .map(|&pkg| (pkg, resolve.features(pkg).to_vec()))
+        .collect()
+}
+
 pub fn resolve_with_global_context(
     deps: Vec<Dependency>,
     registry: &[Summary],
     gctx: &GlobalContext,
-) -> CargoResult<Vec<PackageId>> {
+) -> CargoResult<Vec<(PackageId, Vec<InternedString>)>> {
     let resolve = resolve_with_global_context_raw(deps, registry, gctx)?;
-    Ok(resolve.sort())
+    Ok(collect_features(&resolve))
 }
 
 pub fn resolve_with_global_context_raw(
@@ -201,7 +216,7 @@ fn log_bits(x: usize) -> usize {
 // At this point is possible to select every version of every package.
 // So we need to mark certain versions as incompatible with each other.
 // We could add a clause not A, not B for all A and B that are incompatible,
-fn sat_at_most_one(solver: &mut impl varisat::ExtendFormula, vars: &[varisat::Var]) {
+fn sat_at_most_one(solver: &mut varisat::Solver<'_>, vars: &[varisat::Var]) {
     if vars.len() <= 1 {
         return;
     } else if vars.len() == 2 {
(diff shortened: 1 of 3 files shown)


Eh2406:
Thank you so much for working on this! I look forward to digging into what you've done.

That being said, there's a lot here. Any advice for where to start? Is it possible to split up the change into smaller logically separable commits? For example changing `resolve` to take a `Summary` instead of a list of deps, could probably be its own commit that happens before the code that actually looks for the feature table. That might make it easier to review what is "just" mechanically moving the arguments around, and what is fundamental changes. And would be a good place for me to ask "would it be clearer if we leave `resolve` with the same simple API for the tests that exist and add a `resolve_with_features` for things that use the new functionality?".

x-hgg-x:
I splitted the PR in two commits and updated the PR description.

crates/resolver-tests/src/lib.rs
```diff
@@ -274,22 +417,47 @@ pub struct SatResolver(Rc<RefCell<SatResolverInner>>);
...
     var_for_is_packages_used: HashMap<PackageId, varisat::Var>,
-    by_name: HashMap<&'static str, Vec<PackageId>>,
+    var_for_is_packages_features_used: HashMap<PackageId, HashMap<InternedString, varisat::Var>>,
+    by_name: HashMap<InternedString, Vec<Summary>>,
 }
 
 impl SatResolver {
     pub fn new(registry: &[Summary]) -> Self {
-        let mut cnf = varisat::CnfFormula::new();
+        let mut solver = varisat::Solver::new();
```
x-hgg-x, line 428: The `Solver` is used directly without creating a `CnfFormula`, since otherwise the variable counter is reset when creating the `Solver` and the root package variable conflicts with the first registry package.

crates/resolver-tests/src/lib.rs
```diff
@@ -411,22 +620,62 @@ impl SatResolver {
...
+                    writeln!(&mut out, "    + {feature_name}").unwrap();
+                }
+            }
+
             out
         })
     }
+
+    #[cfg(test)]
+    fn reset_solver_lits(&self) {
```
x-hgg-x, line 658: This is used for the test below, since a solution activating more packages than necessary is still valid.

crates/resolver-tests/src/lib.rs
```diff
@@ -298,110 +466,151 @@ impl SatResolver {
...
+            .map(|&f| (f, solver.new_var()))
+            .collect();
+
+        for (_, root_feature_var) in &var_for_root_features_used {
+            solver.add_clause(&[root_feature_var.negative(), root_var.positive()]);
+        }
+
+        // replace previous root package
+        var_for_is_packages_used.insert(root_pkg, root_var);
+        var_for_is_packages_features_used.insert(root_pkg, var_for_root_features_used);
```
x-hgg-x, line 537: the root package variables are inserted in these maps to allow reusing `add_pkg_clauses()` and `process_feature_value()` functions.

crates/resolver-tests/src/lib.rs
Eh2406: As I recall the CNF was used because of a performance problem when adding clauses directly to `Solver`. Although that may have been fixed in the intervening years. It seems like there's a fair bit of complexity caused by adding clauses in side `sat_resolve`, on the other hand there's some good code sharing thanks to your new `add_pkg_clauses`.

I wonder if we could avoid the complexity in `sat_resolve` by deciding that the root package cannot have features. Its dependencies can activate transitive features, which may activate transitive optional dependencies. But, the exact process of going from user requested features to a list of dependencies for the root crate has to be done before calling `sat_resolve`. That's obviously less flexible. But are there real bugs that this simplification could not test? What are your thoughts?

I will continue to review when I have time tomorrow. Possibly during my office hours. Depending on who attends. If you attend we could discuss this work face to face!

x-hgg-x: > As I recall the CNF was used because of a performance problem when adding clauses directly to `Solver`. Although that may have been fixed in the intervening years. It seems like there's a fair bit of complexity caused by adding clauses in side `sat_resolve`, on the other hand there's some good code sharing thanks to your new `add_pkg_clauses`.

I didn't see any performance regression in the existing prop tests, since `SatResolver::new()` is  only called one time and then reused.

> I wonder if we could avoid the complexity in `sat_resolve` by deciding that the root package cannot have features. Its dependencies can activate transitive features, which may activate transitive optional dependencies. But, the exact process of going from user requested features to a list of dependencies for the root crate has to be done before calling `sat_resolve`. That's obviously less flexible. But are there real bugs that this simplification could not test? What are your thoughts?

We cannot replace clauses in `sat_resolve` by simple assumptions, because a required dependency for the root package is represented by a `OR` clause with all compatible packages.

What about the following example:
```toml
# Cargo.toml
[package]
name = "root"

[dependencies]
indexmap = { version = "2.5.0", optional = true }
# ...
# other deps that may activate `indexmap`
# ...
```
```sh
# command-line
cargo c -F 'indexmap?/serde'
```
Assuming `--all-features` from the cli is not enough, since the argument introduces a new feature not specified anywhere. Keeping the feature clauses for the root package can help reusing code in this case.

Additionally this example doesn't work with vendoring, since the generated `Cargo.lock` doesn't contains `serde` (the version resolution for `serde` is done on the fly when executing the command, but the result is not saved anywhere).

I don't think there are any test for this, since all existing tests assume `--all-features` for the root package.

Eh2406: > I didn't see any performance regression in the existing prop tests, since SatResolver::new() is only called one time and then reused.

Each prop test is called hundreds of times each with a distinct registry, `SatResolver::new()` is called once per generated registry. I was also not able to reproduce the performance problem, so I guess it's been solved at some point in the past few years. So all good.

> We cannot replace clauses in `sat_resolve` by simple assumptions, because a required dependency for the root package is represented by a `OR` clause with all compatible packages.

Right, but that's not new.  The old code also had to deal with  a dependency from the root being an `OR`  clause. That is what the old code was using `this_call`/`matches` to do. It seems that the old code successfully added a clause to the solver without clobbering the internal numbering system  or hitting the conflict you described.

> I don't think there are any test for this, since all existing tests assume `--all-features` for the root package.

Yes, I don't think there any tests for this behavior. If I remember correctly, the ability for the command line to  turn on any arbitrary dependencies arbitrary feature was accidentally introduced.  Whether we are locked in by backwards compatibility, or whether we can consider it a bug, is being discussed in other places.

The question at hand is... Imagine we  had a prop test that completely covered all possible `root` crates that only have required features, so it will test the registry with `indexmap = { version = "2.5.0" }` and `indexmap = { version = "2.5.0", features =["serde"]}`. (Prop tests never give  guarantees, but if we've been running them long enough we have a probabilistic statement of coverage.) Do we think it's possible that a user  will find a bug that uses command line `-F`  that cannot be reproduced by the prop test?

crates/resolver-tests/src/lib.rs
Eh2406: In practice, this seems to work, and be required. But it makes me a little nervous, the documentation for `.assume` says "This replaces the current set of assumed literals". So the API does not guarantee that  what's done in this method has any effect on subsequent calls to `sat_resolve`. I'm also not sure why it ends up being needed when `assumption.extend(old_root_vars.iter().map(|v| v.negative()))` is in `sat_resolve`.

What am I missing?

crates/resolver-tests/src/lib.rs
```diff
@@ -884,3 +1179,121 @@ pub fn assert_same<A: PartialEq + std::fmt::Debug>(a: &[A], b: &[A]) {
...
+            &[
+                ("default", &["log", "image/default"]),
+                ("man", &["rgb?/man"]),
+            ],
+        );
+
+        let checks = [
+            (
+                &CliFeatures::new_all(true),
+                r#"used:
```
Eh2406, line 1219: Do we want to be testing that each of these matches a hardcoded value? 
- If no: If we move them to `crates/resolver-tests/tests/resolve.rs`, we could use `resolve_and_validated` to make sure we get the same output as cargo, without needing to maintain the hardcoded value.
- If yes: The rest of cargo is moving toward using `snapbox` for maintaining these hardcoded strings.

crates/resolver-tests/src/lib.rs
```diff
@@ -227,51 +247,217 @@ fn sat_at_most_one(solver: &mut impl varisat::ExtendFormula, vars: &[varisat::Va
...
+                    var_for_is_packages_features_used[&dep.package_id()][&dep_feature_name];
+
+                solver.add_clause(&[
+                    bound_var.negative(),
+                    dep_var.negative(),
+                    dep_feature_var.positive(),
+                ]);
+            }
+
+            if !weak {
```
Eh2406, line 331: This looks basically correct. I wish I could make a stronger statement. I wouldn't be surprised if there are corner cases about situations where there are features and dependencies by the same name or around distinguishing between the default feature and a feature called "default". But I don't know how to find these corner cases except comparing cargo and that SAT resolver across a large number of examples. Which is not part of this PR because it's follow-up work.

Some of that follow-up work will look like:
- making the Prop Tests output features that might hit this corner case.
- I found a lot of corner cases the hard way when comparing cargo and the PubGrub based resolver. The minimize examples are at [Eh2406/pubgrub-crates-benchmark/index_ron](https://github.com/Eh2406/pubgrub-crates-benchmark/tree/main/out/index_ron). We could turn those test examples into hand coded tests in `crates/resolver-tests/tests/resolve.rs`.
- We could bring some of that approach to this project and have a command to run `resolve_and_validated` for each version of every crate on crates.io.

Anyway just some context and potential follow-up work.

Eh2406:
In reviewing this I'm noticing how very long `lib.rs` and `resolve.rs` are getting. Let's not do it in this PR, but as a follow-up it would be really nice to split these monoliths into separate files.

ehuss:
r? Eh2406

crates/resolver-tests/src/lib.rs
x-hgg-x: > The question at hand is... Imagine we had a prop test that completely covered all possible root crates that only have required features, so it will test the registry with indexmap = { version = "2.5.0" } and indexmap = { version = "2.5.0", features = ["serde"] }.

Yes this should work: testing my above example is equivalent to testing `root1`, `root2`, `root3`, `root4` like this:
```toml
[package]
name = "dep"
version = "1.0.0"

[dependencies]
indexmap = { version = "2.5.0", optional = true }
# ...

[features]
cli1 = []
cli2 = ["dep:indexmap"]
cli3 = ["indexmap/serde"]
cli4 = ["indexmap?/serde"]
```
```toml
[package]
name = "root1"

[dependencies]
dep = { version = "1.0.0", features = ["cli1"]
```
```toml
[package]
name = "root2"

[dependencies]
dep = { version = "1.0.0", features = ["cli2"]
```
```toml
[package]
name = "root3"

[dependencies]
dep = { version = "1.0.0", features = ["cli3"]
```
```toml
[package]
name = "root4"

[dependencies]
dep = { version = "1.0.0", features = ["cli4"]
```

So we can limit the root package to have only non-optional dependencies and no features, so that `sat_resolve` don't need to handle cli features.

> It seems that the old code successfully added a clause to the solver without clobbering the internal numbering system or hitting the conflict you described.

The solver allows reusing a variable not bound by any clause after adding a cnf formula. With the following registry and dependencies, the old code gives the same variable value `3` for the dep `b` and the `root` package:
```
let reg = registry(vec![
    pkg!(("a", "1.5.0")),
    pkg!(("a", "1.6.0")),
    pkg!(("b", "1.5.0")),
    pkg!(("c", "1.4.0")),
]);
let deps = vec![dep_req("a", "1.5.0"), dep("b")];
```
This is not a problem for the old code since it doesn't keep a reference of the root variable inside the `SatResolve`.

crates/resolver-tests/src/lib.rs
x-hgg-x: I think when calling `solve` several times using [incremental solving](https://jix.github.io/varisat/manual/0.2.1/lib/incremental.html), the solver starts from the previous solution. Here I start with the `--all-features` case, which is also a solution of all the following cases (`default-features`, `--no-default-features`, `man` and `man,rgb`), so no step is needed for the solver and the variables are not updated. A call to `reset_solver_lits()` move away the starting point so it has more chances to give the preferred solution.

In any case this test is not very robust, since there is no guarantee that the solver will stop at the given set of dependencies and features. And the `resolve_and_validated` function only test the satisfiability property and not the solution found.

crates/resolver-tests/src/lib.rs
x-hgg-x: > where there are features and dependencies by the same name or around distinguishing between the default feature and a feature called "default"

This is managed when constructing the `Summary`:

https://github.com/rust-lang/cargo/blob/01e1ab55dd7f391cef809c05a50030ae3bb018d8/src/cargo/core/summary.rs#L189

An implicit feature `dep1 = ["dep:dep1"]` from an optional dependency is only created if the feature `dep1` is not already defined, and if the feature `default` is not already defined, an implicit feature `default = []` is added.
Since the SAT resolver receives `Summary` objects, theses corner cases don't need to be handled.

crates/resolver-tests/src/lib.rs
Eh2406: > And the `resolve_and_validated` function only test the satisfiability property and not the solution found.

It's a little better than that. If cargo says there is no solution, then it checks that SAT thinks there's no solution. This is the best we can do in the no solution case (given the current resolver). If cargo says there is a solution, then we check that that solution is acceptable to SAT with `sat_is_valid_solution`.

crates/resolver-tests/src/lib.rs
```diff
@@ -227,51 +242,201 @@ fn sat_at_most_one(solver: &mut impl varisat::ExtendFormula, vars: &[varisat::Va
...
-/// find a valid resolution if one exists. The big thing that the real resolver does,
-/// that this one does not do is work with features and optional dependencies.
+/// For the subset of functionality that are currently made by `registry_strategy`,
+/// this will find a valid resolution if one exists.
 ///
-/// The SAT library dose not optimize for the newer version,
+/// The SAT library does not optimize for the newer version,
 /// so the selected packages may not match the real resolver.
-#[derive(Clone)]
-pub struct SatResolve(Rc<RefCell<SatResolveInner>>);
```
Eh2406, line 255: Why the removal of this wrapper type? It was nice to allow callers of `resolve and validate` to reuse a `SatResolve` if they had one, or pass `None` if they just wanted one to be created. Happy to switch to the more direct code if there's a good reason to. I was just wondering what the reason was?

x-hgg-x: I modified the signature of `sat_resolve` and `sat_is_valid_solution` to take a `&mut self`, so that is more obvious for the reader that the internal solver state is updated. Since the struct doesn't need to be clonable, I removed the wrapper.

crates/resolver-tests/src/lib.rs
```diff
@@ -227,51 +242,201 @@ fn sat_at_most_one(solver: &mut impl varisat::ExtendFormula, vars: &[varisat::Va
...
-        sat_at_most_one(cnf, key);
+        sat_at_most_one(solver, key);
     }
     by_keys
 }
 
+fn find_compatible_dep_summaries_by_name_in_toml(
+    pkg_dependencies: &[Dependency],
+    by_name: &HashMap<InternedString, Vec<Summary>>,
+) -> IndexMap<InternedString, Vec<Summary>> {
```
Eh2406, line 262: Why does this need to be in `IndexMap`? Not objecting to it, it's a good library, I'm just not sure why we need to keep the order.

crates/resolver-tests/src/lib.rs
```diff
@@ -227,51 +242,201 @@ fn sat_at_most_one(solver: &mut impl varisat::ExtendFormula, vars: &[varisat::Va
...
-/// The SAT library dose not optimize for the newer version,
+/// The SAT library does not optimize for the newer version,
 /// so the selected packages may not match the real resolver.
-#[derive(Clone)]
-pub struct SatResolve(Rc<RefCell<SatResolveInner>>);
-
-struct SatResolveInner {
+pub struct SatResolver {
     solver: varisat::Solver<'static>,
+    old_root_vars: Vec<varisat::Var>,
```
Eh2406, line 402: I don't think we need to keep `old_root_var` anymore. If cargo found a solution, then we make a assumption of everything we care about. If cargo did not find a solution, then SAT will opt out of as many constraints as it can including all the old `root_var` before deciding there is no solution.

So removing all code that refers to `old_root_vars` the code still passes all tests.

crates/resolver-tests/src/lib.rs
```diff
@@ -280,147 +445,206 @@ impl SatResolve {
...
+            for (&p, v) in self.var_for_is_packages_used.iter() {
                 if lits.contains(v) {
-                    writeln!(&mut out, "    {}", p).unwrap();
+                    used_packages.entry(p).or_default();
                 }
             }
+            for (&p, map) in &self.var_for_is_packages_features_used {
+                for (&f, v) in map {
+                    if lits.contains(v) {
+                        used_packages.entry(p).or_default().insert(f);
```
Eh2406, line 594: Super nit: If we ever use the `or_default` something's gone very wrong. Because we've activated a feature for package that was not activated. And this will be hard to debug because the output will claim that it was activated. I think I'd rather see a panic here. `.entry(p).or_default()` -> `.get_mut(&p).expect("The feature is activated without the package being activated")`

Eh2406 (approved):
I'm down to some nits at this point. Thank you so much for all the work.

crates/resolver-tests/src/lib.rs
x-hgg-x: I have currently this code which depends of the iteration order of the map:
```rust
for (dep, compatible_dep_summaries) in pkg_dependencies
        .iter()
        .zip(compatible_dep_summaries_by_name_in_toml.values())
```
But this is not necessary since I can directly use the key, so I will use an `HashMap` here.

crates/resolver-tests/src/lib.rs
x-hgg-x: The `old_root_vars` field is used to optimize the solving time. If we construct a `SatResolver` and then solve for a large number of root packages, adding assumptions that old root variables are negative allows the solver to completely ignore all previously added clauses, so it shouldn't take too much time to solve even after with a large number of total clauses.

Eh2406:
Thanks for the work!
@bors r+

PR #14585, opened 2024-09-24
chore: Bump MSRV to 1.81

epage:
This is prep for using `#[expect]`.

Its not clear why RenovateBot didn't do this.


ehuss:
Thanks!

@bors r+

ehuss:
> Its not clear why RenovateBot didn't do this.

It's not clear to me from the docs, but I get the impression that packageRules.schedule can't really override the global schedule (https://github.com/renovatebot/renovate/discussions/25829 is related). Maybe it can only make it less frequent? So it is only running once a month on the first day.

epage:
Except it worked as of #14303 and we haven't changed our config since then unless you are saying something at the org level and there was a change there.

PR #14587, opened 2024-09-24
fix(cargo-rustc): give trailing flags higher precedence on nightly

### LINKED ISSUE #14346: Precedence of `cargo rustc -- <flags>`  is too low to override flags set by Cargo
<sup>**Original issue title:** `-C strip=symbols` doesn't work like `strip = true`</sup>

### Problem

Reproducer: https://github.com/konstin/strip-reproducer/blob/main/compile.sh

With the default linker on linux, passing `-C link-arg=-s`, `-C strip=symbols` and `-C strip=debuginfo` to `cargo rustc` all don't strip the binary properly as setting `strip = true` does, even though the [docs](https://doc.rust-lang.org/cargo/reference/profiles.html#strip) say "The strip option controls the [-C strip flag](https://doc.rust-lang.org/rustc/codegen-options/index.html#strip)".

In [maturin](https://github.com/PyO3/maturin) we expose a `--strip` option, since small size is important for publishing. We pass this on to `cargo rustc`, historically with `-C link-arg=-s`, but i also tried with `-C strip=symbols`. This does strip some debuginfo, 40MB -> 37MB in the case of uv, but not down to 30MB for the fully stripped binary (https://github.com/astral-sh/uv/issues/5683). In the minimal reproducer, only `strip = true` shows any effect.

Note that this only happens with the default linker, mold strips properly in both cases.

### Steps

https://github.com/konstin/strip-reproducer/blob/main/compile.sh

Create a new project with `cargo new strip-reproducer`, then run:

```bash
cargo build -q --release
echo "Release: $(stat --printf="%s" target/release/strip-reproducer)"

cargo rustc -q --release -- -C strip=symbols
echo "strip=symbols: $(stat --printf="%s" target/release/strip-reproducer)"

cargo rustc -q --release -- -C strip=debuginfo
echo "strip=debuginfo: $(stat --printf="%s" target/release/strip-reproducer)"

printf "[profile.release]\nstrip = true\n" >> Cargo.toml
cargo build -q --release
echo "strip = true: $(stat --printf="%s" target/release/strip-reproducer)"
```

```
Release: 398360
strip=symbols: 398360
strip=debuginfo: 398360
strip = true: 334304
```

### Possible Solution(s)

There should ideally be a cli option to strip like `strip = true` does, or the docs should be adjusted on the differences between `strip = true` and the `-C` options.

### Notes

_No response_

### Version

```text
cargo 1.80.0 (376290515 2024-07-16)
release: 1.80.0
commit-date: 2024-07-16
```

weihanglo:

Previously `cargo rustc -- <flags>` got a lower precedence than
some of the flags set by cargo internal.
This is a bit unintuitive as Cargo generally treats user-provided
CLI flags with the highest priority.

This commit changes `cargo rustc -- <flags>` to a higher precedence:
higher than most of flags set by Cargo, and only lower than
`build.rustflags` family.

Unsure if this affects people's workflow, so this behavior is only
enabled on nightly for collectin feedback. A environment variable
`__CARGO_RUSTC_ORIG_ARGS_PRIO=1` is provided for users to opt-out.
If everything goes well, the nightly gate will be removed after a
few of releases.

### Additional information

See discussion on https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/rustflags.20precendence.20of.20.60cargo.20rustc.60

Fixes #14346

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -694,6 +694,11 @@ fn prepare_rustc(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> CargoResult
     base.inherit_jobserver(&build_runner.jobserver);
     build_deps_args(&mut base, build_runner, unit)?;
     add_cap_lints(build_runner.bcx, unit, &mut base);
+    if cargo_rustc_higher_args_precedence(build_runner) {
+        if let Some(args) = build_runner.bcx.extra_args_for(unit) {
+            base.args(args);
+        }
+    }
     base.args(&unit.rustflags);
     if build_runner.bcx.gctx.cli_unstable().binary_dep_depinfo {
         base.arg("-Z").arg("binary-dep-depinfo");
@@ -753,8 +758,11 @@ fn prepare_rustdoc(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> CargoResu
     }
 
     rustdoc.args(unit.pkg.manifest().lint_rustflags());
-    if let Some(args) = build_runner.bcx.extra_args_for(unit) {
-        rustdoc.args(args);
+
+    if !cargo_rustc_higher_args_precedence(build_runner) {
+        if let Some(args) = build_runner.bcx.extra_args_for(unit) {
+            rustdoc.args(args);
+        }
     }
 
     let metadata = build_runner.metadata_for_doc_units[unit];
@@ -795,6 +803,11 @@ fn prepare_rustdoc(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> CargoResu
 
     rustdoc::add_output_format(build_runner, unit, &mut rustdoc)?;
 
+    if cargo_rustc_higher_args_precedence(build_runner) {
+        if let Some(args) = build_runner.bcx.extra_args_for(unit) {
+            rustdoc.args(args);
+        }
+    }
     rustdoc.args(&unit.rustdocflags);
 
     if !crate_version_flag_already_present(&rustdoc) {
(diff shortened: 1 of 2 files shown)


src/cargo/core/compiler/mod.rs
```diff
@@ -694,6 +694,11 @@ fn prepare_rustc(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> CargoResult
     base.inherit_jobserver(&build_runner.jobserver);
     build_deps_args(&mut base, build_runner, unit)?;
     add_cap_lints(build_runner.bcx, unit, &mut base);
+    if build_runner.bcx.gctx.nightly_features_allowed {
```
epage, line 697: > Unsure if this affects people's workflow, so this behavior is only
enabled on nightly for collectin feedback. If everything goes well,
the nightly gate will be removed after a few of releases.

When we discussed this in the Cargo team meeting, one idea was a short-term env variable to opt-in to this and then we transition to opt-out from this.

What were your thoughts on this direction?

weihanglo: Should we go with what we've done to git2-to-gitx transition for cargo-publish, like having an opt-out env var for a while? I slightly doubt that people will use a short-live opt-in environment variable, not to say it's pretty hard to get people onboard and solicit feedback.

epage: > Should we go with what we've done to git2-to-gitx transition for cargo-publish, like having an opt-out env var for a while? I slightly doubt that people will use a short-live opt-in environment variable, not to say it's pretty hard to get people onboard and solicit feedback.

I can understand skipping opt-in.

If you are referring to doing that on stable, the one difference is that git2/gix were meant to be behavior parity while in this case we are changing behavior.  Having an opt-out on stable buys people time to update if it breaks them but it doesn't let us take feedback and change it as it will already be stable.

If you are only referring to nightly, then I'm fine.

weihanglo: Nightly only for now. Let's see how it goes.

> Having an opt-out on stable buys people time to update if it breaks them but it doesn't let us take feedback and change it as it will already be stable.

To me this is the way to go. We have done similar process for config array merging, and expected it to have a larger impact. Arlo ever wrote an [inside rust post](https://blog.rust-lang.org/inside-rust/2023/08/24/cargo-config-merging.html) for it. However, we have not heard back from anyone, not even a linked issue to [the original PR](https://github.com/rust-lang/cargo/pull/12515). Yet, no need to decide how to stabilize now.

weihanglo: An environment variable is added `__CARGO_RUSTC_ORIG_ARGS_PRIO`.

epage:
Should we keep #14346 open for tracking stabilization?

PR #14588, opened 2024-09-24
feat(timings): support dark color scheme in HTML output

### LINKED ISSUE #14586: cargo-timing.html supports dark mode
### Problem

For users using dark mode, opening cargo-timing.html is very glaring

I can use some browser ext to force the web page to switch to dark mode, but the canvas will still be white background.

Image

### Proposed Solution

For web pages, there is already prefers-color-scheme

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

For canvas, maybe just invert the basic white and black, and keep the other colors?

### Notes

_No response_

ahaoboy:

Use CSS variables to control colors in different modes

Fixes #14586

Switch to dark mode to preview cargo-timing.html file

### Additional information

I am not an expert in CSS. The color table is generated by gpt. Maybe there are other better color schemes.

Snipaste_2024-09-24_19-36-14
Snipaste_2024-09-24_19-36-50

### DIFF
--- a/src/cargo/core/compiler/timings.js
+++ b/src/cargo/core/compiler/timings.js
@@ -22,6 +22,21 @@ let UNIT_COORDS = {};
 // Map of unit index to the index it was unlocked by.
 let REVERSE_UNIT_DEPS = {};
 let REVERSE_UNIT_RMETA_DEPS = {};
+
+// Colors from css
+const getCssColor = name => getComputedStyle(document.body).getPropertyValue(name);
+const TEXT_COLOR = getCssColor('--text');
+const BG_COLOR = getCssColor('--background');
+const CANVAS_BG = getCssColor('--canvas-background');
+const AXES_COLOR = getCssColor('--canvas-axes');
+const GRID_COLOR = getCssColor('--canvas-grid');
+const BLOCK_COLOR = getCssColor('--canvas-block');
+const CUSTOM_BUILD_COLOR = getCssColor('--canvas-custom-build');
+const NOT_CUSTOM_BUILD_COLOR = getCssColor('--canvas-not-custom-build');
+const DEP_LINE_COLOR = getCssColor('--canvas-dep-line');
+const DEP_LINE_HIGHLIGHTED_COLOR = getCssColor('--canvas-dep-line-highlighted');
+const CPU_COLOR = getCssColor('--canvas-cpu');
+
 for (let n=0; n<UNIT_DATA.length; n++) {
   let unit = UNIT_DATA[n];
   for (let unlocked of unit.unlocked_units) {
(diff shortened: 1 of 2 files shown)


src/cargo/core/compiler/timings.rs
```diff
@@ -609,8 +609,59 @@ static HTML_TMPL: &str = r#"
   <title>Cargo Build Timings — {ROOTS}</title>
   <meta charset="utf-8">
 <style type="text/css">
+:root {
+  --error-text: #e80000;
+  --text: #000;
+  --background: #fff;
+  --h1-border-bottom: #c0c0c0;;
```
weihanglo, line 616:
```suggestion
  --h1-border-bottom: #c0c0c0;
```

src/cargo/core/compiler/timings.rs
```diff
@@ -609,8 +609,59 @@ static HTML_TMPL: &str = r#"
...
+    --table-tr-border-top: #333;
+    --table-tr-border-bottom: #333;
+    --table-tr-odd-background: #1e1e1e;
+    --table-td-background: #262626;
+    --table-td-border-right: #333;
+    --canvas-background: #1a1a1a;
+    --axes: #b0b0b0;
+    --grid: #333;
+    --dep-line: #444;
+    --dep-line-highlighted: #fff;
```
weihanglo, line 657: These are all used inside canvas, right? If that is the case I would prefer naming them with a `canvas-` prefix.

src/cargo/core/compiler/timings.rs
```diff
@@ -609,8 +609,59 @@ static HTML_TMPL: &str = r#"
...
+  --grid: #e6e6e6;
+  --block: #aa95e8;
+  --custom-build: #f0b165;
+  --not-custom-build: #95cce8;
+  --dep-line: #ddd ;
+  --dep-line-highlighted: #000;
+  --cpu: rgba(250, 119, 0, 0.2)
+}
+
+@media (prefers-color-scheme: dark) {
```
weihanglo, line 638: Could we split the dark theme support to the other commit?

That is, the first commit would be theming the current CSS color in CSS variables, followed by adding a dark color scheme media qeury.

src/cargo/core/compiler/timings.js
weihanglo, line 1: I am pretty sure that somebody will complain about dark theme. Not sure if this is a blocker, but should we have a simple theme toggle?

weihanglo (commented):
This is cool! Thank you for the contribution.

src/cargo/core/compiler/timings.js
ahaoboy: Adding a button to switch modes is more complicated, the button style, position, mode priority handling, etc. Here is an example, if it is acceptable then I will add it to this pr

https://github.com/rust-lang/cargo/commit/b3f4f57464a1da0b249d9db0a3be90159288bf7c

[cargo-timing-toggle.zip](https://github.com/user-attachments/files/17143523/cargo-timing-toggle.zip)

weihanglo: Some colors in the canvas seem broken, but thank you for the attempt :)

I am happy to merge this PR as-is and look forward to your follow-up PR!

weihanglo (approved):

src/cargo/core/compiler/timings.js
ahaoboy: Haha, this example only implements the toggle, missing some css colors, maybe we can add it when someone needs this feature :D

PR #14589, opened 2024-09-24
Implement `--locked` for build-std

adamgemmell:

This fixes https://github.com/rust-lang/wg-cargo-std-aware/issues/38

Drop the version of a dependency in the rust-src Cargo.lock (for example, libc down to 0.2.149). Then build something with build-std.
Before this patch, libc would be silently upgraded to the latest semver-compatible version. Now Cargo will return an error.

### Additional information

In order for Cargo to actually attempt to write the std lockfile we need to resolve with optional_deps again. Optional packages aren't actually built.

This works for now but the behaviour will need a rethink for rust-lang/wg-cargo-std-aware#64 - perhaps adding some concept of --locked but only for the std parts of a dependency graph..


adamgemmell:
r? author

src/cargo/core/compiler/standard_lib.rs
```diff
@@ -78,6 +78,7 @@ pub fn resolve_std<'gctx>(
     // TODO: Consider doing something to enforce --locked? Or to prevent the
     // lock file from being written, such as setting ephemeral.
     let mut std_ws = Workspace::new(&std_ws_manifest_path, gctx)?;
+    std_ws.set_is_std(true);
```
epage, line 81: Should we name this for the specific policy involved (forcing the workspace to be locked, independent of any flags)?

src/cargo/core/workspace.rs
```diff
@@ -103,6 +103,12 @@ pub struct Workspace<'gctx> {
     /// file. This is set for `cargo install` without `--locked`.
     ignore_lock: bool,
 
+    /// If `true`, then the resolver will not update the `Cargo.lock` file and
+    /// return an error if the lockfile is missing or out of date, similar
+    /// to the `GlobalContext::locked()` behaviour. Note that
```
weihanglo, line 108: There are already too many fields controlling the `locked` behavior. Per Ed's suggestion, I'd like to see if we can add one line explaining this field is independent from `--locked` CLI flag. Perhaps something like:

```suggestion
    /// If `true`, then the resolver will not update the `Cargo.lock` file and
    /// return an error if the lockfile is missing or out of date.
    ///
    /// This field is independent from any flags like `--locked` or `--frozen`.
    /// Use [`GlobalContext::locked()`] whenever you have an access to mutable
    /// to [`GlobalContext`], such as at Command arg parsing level.
```

src/cargo/ops/lockfile.rs
```diff
@@ -66,6 +66,14 @@ pub fn write_pkg_lockfile(ws: &Workspace<'_>, resolve: &mut Resolve) -> CargoRes
         );
     }
 
+    if ws.is_locked() {
+        anyhow::bail!(
+            "Attempted to write to the standard library's lockfile.\n\
+            This most likely means the lockfile has been previously modified by mistake.\
+            Try removing and readding the `rust-src` component."
```
weihanglo, line 73: * Lower case for first letter by convention in this repo.
* For rust-src component, it should be conditioned on the availability of rustup. You may use [`cargo::util::is_rustup`](https://github.com/rust-lang/cargo/blob/15fbd2f607d4defc87053b8b76bf5038f2483cf4/src/cargo/util/mod.rs?plain=1#L71) to check.

```suggestion
            "Attempted to write to the standard library's lockfile at: {}.\n\
            This most likely means the lockfile has been previously modified by mistake.\ 
            Try removing and readding the `rust-src` rustup component.
            lock_root.as_path_unlocked().join(LOCKFILE_NAME).display(),
```

src/cargo/core/compiler/standard_lib.rs
```diff
@@ -75,13 +75,8 @@ pub fn resolve_std<'gctx>(
     let src_path = detect_sysroot_src_path(target_data)?;
     let std_ws_manifest_path = src_path.join("Cargo.toml");
     let gctx = ws.gctx();
-    // TODO: Consider doing something to enforce --locked? Or to prevent the
-    // lock file from being written, such as setting ephemeral.
     let mut std_ws = Workspace::new(&std_ws_manifest_path, gctx)?;
-    // Don't require optional dependencies in this workspace, aka std's own
-    // `[dev-dependencies]`. No need for us to generate a `Resolve` which has
-    // those included because we'll never use them anyway.
-    std_ws.set_require_optional_deps(false);
```
weihanglo, line 84: Why do we remove this? Did it cause any issue?

weihanglo: > In order for Cargo to actually attempt to write the std lockfile we need to resolve with optional_deps again. Optional packages aren't actually built.

I guess it is for this?

weihanglo: https://github.com/rust-lang/cargo/blob/15fbd2f607d4defc87053b8b76bf5038f2483cf4/src/cargo/ops/resolve.rs?plain=1#L289-L294

Okay it is this one. 

I am unsure if removing this is the correct approach. `require_optional_deps` affects how we resolve dependencies. And with that std starts respecting [path override](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#paths-overrides).

https://github.com/rust-lang/cargo/blob/15fbd2f607d4defc87053b8b76bf5038f2483cf4/src/cargo/ops/resolve.rs?plain=1#L163-L173

weihanglo (commented):
I am thinking over the UI design of this again. See and discuss in <https://github.com/rust-lang/wg-cargo-std-aware/issues/38#issuecomment-2412384869>.

adamgemmell:
Going to close this for now pending a more comprehensive plan for the build-std as part of the project goal.

PR #14590, opened 2024-09-24
test: Remove completion tests

### LINKED ISSUE #14545: shell_completions test issues
### Problem

The new shell_completions tests have a few issues that need to be resolved before they can be enabled.

1. The tests randomly fail. I have seen `fish` and `zsh` randomly fail.  https://github.com/rust-lang/cargo/pull/14541#issuecomment-2351096767 indicates that it might be an interactive timeout.
2. They require external tools to be installed to run. Tests generally should not require that.
    * Looking for the binaries is not sufficient, since for example the `bash` test also requires the bash-completions package to be installed.
3. The elvish test has a weird failure for me:
```
++++ actual:   In-memory
        1 + Deprecation: the legacy temporary assignment syntax is deprecated; use "tmp" instead
        2 +   /tmp/cargo/target/tmp/cit/t0/home/elvish/rc.elv:3:7: eval (E:CARGO_COMPLETE=elvish cargo | slurp)
```
4. Tests should be using the `ignore` attribute so that it reports in the output when it isn't running. Currently they exit with `return;` on macos, but that should use the `ignore` attribute instead.
5. Should figure out why the tests are ignored on macos, since I would expect that to work with homebrew. I'm unaware why these are currently ignored.

### Steps

1. `cargo test --test testsuite -- shell_completions`

### Possible Solution(s)

One idea I had for limiting where these tests run is to add a CI_EXTENDED environment variable. The tests would only be required to run if that environment variable is set, and then we can set that in the appropriate jobs (like `test`). Then the cargo_test macro could have something like `requires_extended_fish` or something like that. That would also be useful for the currently hard-coded `hg` and `lldb`. 

However, that doesn't help with some of the more complex issues like the bash-completions problem.

We could extend the cargo_test attribute to have `ignore_macos` to fix the macos ignore problem.

### Notes

_No response_

### Version

Currently on 643a025b3c3ad6f7d3acea558d223784ea8ab932 of master.

epage:
The tests are intended to spot check that shell completions are registered correctly.  That is a low change, low risk area.  For shell integration, we're relying on `clap_complete`s tests. For our own candidates, we should test the candidate generation directly, rather than end-to-end.

This reverts parts of commit e7ca9bec80ab5c010c1a84690816da7b64008257, reversing changes made to bd5f32bb1c7ca273b5d86815bf0ae4adba59ddd8.

Fixes #14545


weihanglo (approved):
Make sense.

Issue #5330, opened 2018-04-09
Make explicit opt-out of target discovery neccessary for Rust 2018


matklad:
Currently, Cargo automatically discovers tests, examples and binaries form the files on the file system. One can also add an explicit `[[example]]` section to Cargo.toml, to add non-discoverable examples, or to tweak an existing example (for example, by specifying required features).

However, adding explicit `[[target]]` disables automatic discovery for other targets of similar type, and this behavior is highly surprising. 

## Current Proposal

New keys will be available under the `[package]` section of `Cargo.toml`:

```toml
[package]
# ... 

autobins = true
autoexamples = true
autotests = true
autobenches = true
```

The value of each of these keys indicates whether or not Cargo's automatic inference logic is in effect for discovering targets. Turning these keys to `false` will completely disable inference and Cargo will not attempt to find more targets. For example setting `autobins = false` will disable Cargo probing `src/bin` looking for binaries.

If these keys are turned to `true` then Cargo will probe the standard set of directories for targets. Any explicitly named targets in the manifest will be merged with the automatically found targets. For example if `autotests = true` is enabled and you've got `tests/foo.rs` and `tests/bar.rs`, you could disable the test harness of `tests/foo.rs` with:

```toml
[[test]]
name = "foo"
harness = false
```

and you'll still be able to use `cargo test --test bar` as Cargo will automatically find the `tests/bar.rs` test.

In the 2015 edition (today in Cargo) these keys are set with the following rules:

* If an explicit target is specified, the key is set to `false`. For example writing `[[bin]]` will set `autobins = false` automatically.
* If no explicit targets are specified, the key is set to `true`.

In the 2018 edition these keys will be unconditionally set to `true` by default.

### Impact on Users Today

Unconditionally turning these keys to `true` is a **breaking change** as it can cause files that weren't previously compiled as a test, for example, to get compiled as a test. As a result projects will need to reorganize their files or otherwise set `autotests = true`, for example. 

If you receive a command-line warning though and don't know what to do with that, please leave a comment here!

## Original Proposal

We would like to change this behavior eventually and for that in Rust 2018  we should:

* add a flag `autodiscover` for each target type, with values true and false (precise flag name and syntax are subject to bike shedding).
* in Rust 2018, warn if `autodiscover` is not specified, at least a single `[[target]]` is listed, and some other target would be discovered if `autodiscover` was set to `true`. 
* in the edition *after* 2018, flip `autodiscover` default to `true`. 

[...]

PR #14591, opened 2024-09-24
feat(toml): Add `autolib`

### LINKED ISSUE #14476: Cargo scripts try to build a nearby `src/lib.rs` if present
### Problem

If I place a script in the same directory as a crate, it uses crate's source files, but does not use crate's dependencies, causing unexpected compilation errors.

### Steps

`./Cargo.toml`:

```toml
[package]
name = "mainpackage"
version = "0.1.0"
edition = "2021"

[dependencies]
hex = "0.4.3"
```

`./src/lib.rs`:

```rust
use hex;
```

`./script.rs`:

```rust
#!/usr/bin/env cargo +nightly -Zscript

//! ```cargo
//! [package]
//! edition = "2021"
//! ```

fn main() {
}
```

----

Running `./script.rs` causes:

```text
warning: `package.edition` is unspecified, defaulting to `2021`
   Compiling script v0.0.0 (…/mainpackage)
error[E0432]: unresolved import `hex`
 --> src/lib.rs:1:5
  |
1 | use hex;
  |     ^^^ no external crate `hex`

```

This looks broken, because `src/lib.rs` is compiled with scripts dependencies, not its usual crate dependencies.

### Possible Solution(s)

I would expect `Cargo.toml` and `src/` to be completely ignored by the scripts. Or if they're not ignored, to be compiled correctly as a library and included as an external crate in the script (like `tests/` or `main.rs` get). But compiling the crate, but with a wrong `Cargo.toml` context seems like the worst of both.

### Notes

BTW, "warning: `package.edition` is unspecified, defaulting to `2021`" is printed, even though I've tried to specify it in all places.

### Version

```text
cargo 1.82.0-nightly (8f40fc59f 2024-08-21)
release: 1.82.0-nightly
commit-hash: 8f40fc59fb0c8df91c97405785197f3c630304ea
commit-date: 2024-08-21
host: aarch64-apple-darwin
os: Mac OS 15.0.0 [64-bit]
```

epage:

PR #5335 added `autobins`, etc for #5330.  Nowhere in there is
discussion of `autolib`.

Cargo script disables support for additional build-targets by disabling
discovery.
Except we don't have a way to disable discovery of `autolib`, leading to #14476.
By adding `autolib`, we can continue in that direction.

This also allows us to bypass inferring of libs on published packages,
like all other build-targets which were handled in #13849.

Fixes #14476

### Additional information

As this seems fairly low controversy, this insta-stabilizes the field.
In prior versions of Cargo, users will get an "unused manifest key"
warning.
For packages where this is set by `cargo publish`, the warning will be suppressed and things will work as normal.
For `cargo vendor`, the same except there will be some churn in the
vendored source as this field will now be set.
For local development, it should be rare to set `autolib` so the lack of
error by discovering a file when this is set shouldn't be a problem.


epage:
Starting FCP in parallel to the PR review

@rfcbot fcp merge

src/cargo/util/toml/targets.rs
```diff
@@ -526,7 +541,18 @@ fn to_bench_targets(
...
-fn is_normalized(toml_targets: Option<&Vec<TomlTarget>>, autodiscover: Option<bool>) -> bool {
+fn is_normalized(toml_target: Option<&TomlTarget>, autodiscover: Option<bool>) -> bool {
+    if autodiscover != Some(false) {
+        return false;
+    }
+
+    let Some(toml_target) = toml_target else {
+        return true;
+    };
+    toml_target.name.is_some() && toml_target.path.is_some()
```
weihanglo, line 552: What about this?

(Guess we'll need to turn `Option<&Vec<TomlTarget>` to `Option<&[TomlTarget]>` in are_normalized` if going to do that)

```suggestion
    are_normalized(toml_target.map(std::slice::from_ref), autodiscover)                                                                     
```

crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -173,6 +173,7 @@ pub struct TomlPackage {
     pub publish: Option<InheritableVecStringOrBool>,
     pub workspace: Option<String>,
     pub im_a_teapot: Option<bool>,
+    pub autolib: Option<bool>,
```
weihanglo, line 176: Everything looks, except this could potentially break other third-party build tools like [`cargo-manifest`](https://crates.io/crates/cargo-manifest) in the future. I'd like to give them a heads-up in <https://rust-lang.zulipchat.com/#narrow/stream/334885-t-cargo.2Fbuild-integration> and merge on Friday. Any other opinion?

weihanglo (approved):

crates/cargo-util-schemas/src/manifest/mod.rs
weihanglo: Posted: https://rust-lang.zulipchat.com/#narrow/stream/334885-t-cargo.2Fbuild-integration/topic/New.20.60package.2Eautolib.60.20for.20disabling.20auto-discovery.20for.20.5Blib.5D

weihanglo: cc @kornelski and @LukeMathWalker since you're maintainers of relevant crates.

LukeMathWalker: cc @Turbo87 who is covering for me while I'm on pat leave

Turbo87: yep yep, already aware of it :)

weihanglo:
It's Friday! Let us not wait for the entire FCP. Feel free to comment below if you have any further concerns. 

@bors r+

PR #14593, opened 2024-09-24
Fix panic when running cargo tree on a package with a cross compiled bindep

elchukc:

This is an attempt to close out @rukai's [PR](https://github.com/rust-lang/cargo/pull/13207#issue-2056019109) for #12358 and #10593 by adjusting the new integration test and resolving merge conflicts.

I have also separated the changes into atomic commits as per [previous review](https://github.com/rust-lang/cargo/pull/13207#discussion_r1543447659).

The integration test that has been edited here is sufficient, plus the new integration test that confirms a more specific case where `cargo tree` throws an error.

### Additional information

I have confirmed the test `artifact_dep_target_specified` fails on master branch and succeeds on this branch.

The first commit fixes the panic and the integration test. Commits 2 and 3 add other tests that confirm behaviour mentioned in related issues.

Commits:
1. [Fix panic when running cargo tree on a package with a cross compiled bindep](https://github.com/rust-lang/cargo/pull/14593/commits/5c5ea78ecd4d92b04cd4d4f8f867a3dffacf9a79) - fixes some panics and changes the integration test to succeed
2. [test: cargo tree panic on artifact dep target deactivated](https://github.com/rust-lang/cargo/pull/14593/commits/ed294ab4e7743454472b13119e110973ed4c56fd) - adds test to confirm the behaviour for the specific panic from [#10539 (comment)](https://github.com/rust-lang/cargo/issues/10593#issuecomment-1317759526)

### DIFF
--- a/src/cargo/ops/tree/graph.rs
+++ b/src/cargo/ops/tree/graph.rs
@@ -391,10 +391,32 @@ fn add_pkg(
         let dep_pkg = graph.package_map[&dep_id];
 
         for dep in deps {
-            let dep_features_for = if dep.is_build() || dep_pkg.proc_macro() {
-                FeaturesFor::HostDep
-            } else {
-                features_for
+            let dep_features_for = match dep
+                .artifact()
+                .and_then(|artifact| artifact.target())
+                .and_then(|target| target.to_resolved_compile_target(requested_kind))
+            {
+                // Dependency has a `{ …, target = <triple> }`
+                Some(target) => FeaturesFor::ArtifactDep(target),
+                // Get the information of the dependent crate from `features_for`.
+                // If a dependent crate is
+                //
+                // * specified as an artifact dep with a `target`, or
+                // * a host dep,
+                //
+                // its transitive deps, including build-deps, need to be built on that target.
+                None if features_for != FeaturesFor::default() => features_for,
+                // Dependent crate is a normal dep, then back to old rules:
+                //
+                // * normal deps, dev-deps -> inherited target
+                // * build-deps -> host
+                None => {
+                    if dep.is_build() || dep_pkg.proc_macro() {
+                        FeaturesFor::HostDep
+                    } else {
+                        features_for
+                    }
+                }
             };
             let dep_index = add_pkg(
                 graph,
--- a/tests/testsuite/artifact_dep.rs
+++ b/tests/testsuite/artifact_dep.rs
@@ -1579,6 +1579,86 @@ foo v0.0.0 ([ROOT]/foo)
         .run();
 }
 
+/// From issue #10593
+/// The case where:
+/// *   artifact dep is { target = <specified> }
+/// *   dependency of that artifact dependency specifies the same target
+/// *   the target is not activated.
+#[cargo_test]
+fn dep_of_artifact_dep_same_target_specified() {
+    if cross_compile::disabled() {
+        return;
+    }
+    let target = cross_compile::alternate();
+    let p = project()
+        .file(
+            "Cargo.toml",
+            &format!(
+                r#"
+                    [package]
+                    name = "foo"
+                    version = "0.1.0"
+                    edition = "2015"
+                    resolver = "2"
+
+                    [dependencies]
+                    bar = {{ path = "bar", artifact = "bin", target = "{target}" }}
+                "#,
+            ),
+        )
+        .file("src/lib.rs", "")
+        .file(
+            "bar/Cargo.toml",
+            &format!(
+                r#"
+                    [package]
+                    name = "bar"
+                    version = "0.1.0"
+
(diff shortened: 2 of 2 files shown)


elchukc:
For continuity, probably
r? weihanglo

Also note I have not re-added `None if features_for != FeaturesFor::default() => features_for` but would be open to creating another test to catch that case.

src/cargo/ops/tree/graph.rs
```diff
@@ -391,10 +391,20 @@ fn add_pkg(
         let dep_pkg = graph.package_map[&dep_id];
 
         for dep in deps {
-            let dep_features_for = if dep.is_build() || dep_pkg.proc_macro() {
-                FeaturesFor::HostDep
-            } else {
-                features_for
+            let dep_features_for = match dep
```
weihanglo, line 394: Thanks for reviving this PR! 

However, this is missing an important part of [my original patch](https://github.com/rust-lang/cargo/issues/10593#issuecomment-1317759526). Please see my [previous comment](https://github.com/rust-lang/cargo/pull/13207#discussion_r1543502588) :)

I am tempted to merge this PR as it fixes some panics. I will be more existed if we can implement the design proposed here: <https://github.com/rust-lang/cargo/issues/10593#issuecomment-1341214604>.

src/cargo/core/resolver/features.rs
```diff
@@ -319,8 +319,32 @@ impl ResolvedFeatures {
         pkg_id: PackageId,
         features_for: FeaturesFor,
     ) -> Vec<InternedString> {
-        self.activated_features_int(pkg_id, features_for)
-            .expect("activated_features for invalid package")
+        let fk = features_for.apply_opts(&self.opts);
```
weihanglo, line 322: I am fine if we split this refactor to another PR.

src/cargo/ops/tree/graph.rs
elchukc: I'm very happy to keep implementing the design! (I'm now emotionally invested), but I think we should split it into separate PRs.'ve updated the PR message to explain each commit as they currently exist.

elchukc: Update: Moved the proposed design change work to #14658 to keep this PR clear.

tests/testsuite/artifact_dep.rs
```diff
@@ -1568,12 +1568,174 @@ fn artifact_dep_target_specified() {
...
+}
+
+#[cargo_test]
+fn check_bindep_depending_on_build_only_dep_with_optional_dep() {
+    if cross_compile::disabled() {
+        return;
+    }
+    // i686-unknown-linux-gnu did not trigger a panic so we need to manually specify wasm32-unknown-unknown instead of relying on cross_compile::alternate()
+    // This is possibly because the `cfg(unix)` later on needs to evaluate to false on the cross compile but true on the host.
+    let target = "wasm32-unknown-unknown";
```
elchukc, line 1667: Isn't it fine that this test fails, since `wasm32-unknown-unknown` isn't installed on the ?

After adding the target on my own computer (`rustup target add wasm32-unknown-unknown`), the test passes. Therefore I will delete this unnecessary test.

tests/testsuite/artifact_dep.rs
```diff
@@ -1568,12 +1568,174 @@ fn artifact_dep_target_specified() {
...
+        .with_stdout_data(str![[r#"
+foo v0.0.0 ([ROOT]/foo)
+└── bindep v0.0.0 ([ROOT]/foo/bindep)
+
+"#]])
+        .with_status(0)
+        .run();
+}
+
+/// From issue #10061
```
elchukc, line 1582: This should say [#10593 (comment)](https://github.com/rust-lang/cargo/issues/10593#issuecomment-1341214604)

tests/testsuite/artifact_dep.rs
weihanglo: I feel that we don't really need this test. My instinct tells me that they share the same bug.

tests/testsuite/artifact_dep.rs
```diff
@@ -1568,12 +1568,174 @@ fn artifact_dep_target_specified() {
...
+                    [package]
+                    name = "bar"
+                    version = "0.1.0"
+
+                    [target.{target}.dependencies]
+                    baz = {{ path = "../baz" }}
+                "#,
+            ),
+        )
+        .file("bar/src/lib.rs", "")
```
weihanglo, line 1623: Should we put a bin here otherwise `cargo check` always fails?

```suggestion
        .file("bar/src/main.rs", "")
```

weihanglo (approved):
Thanks!

While it doesn't really resolve all panics, at least we are one step forward :)

PR #14595, opened 2024-09-25
feat: make lockfile v4 the default 

weihanglo:

This commit makes lockfile version 4 the default version when Cargo
tries to write to a lockfile.

The lockfile version 4 has been stabilized since 1.78.0,
and will become default in 1.83.0.
the length of transition period is pretty similar as before.

One caveat is that in other output from Cargo,
e.g., `cargo metatada`, status messages,
`SourceID` will display in the v4 URL encoded format.
This shouldn't affect the majority of Rust users,
as `SourceId` representation should be opaque to them,
unless comparing `SourceId` across different version of toolchains.

Some of those tests don't really need to be `version = 4`.
They were updated for consistency.

### Additional information

This was discussed in Cargo meeting on 2024-09-24.


tests/testsuite/git.rs
```diff
@@ -283,7 +283,7 @@ fn cargo_compile_git_dep_pull_request() {
         .with_stderr_data(str![[r#"
 [UPDATING] git repository `[ROOTURL]/dep1`
 [LOCKING] 1 package to latest compatible version
-[COMPILING] dep1 v0.5.0 ([ROOTURL]/dep1?rev=refs/pull/330/head#[..])
+[COMPILING] dep1 v0.5.0 ([ROOTURL]/dep1?rev=refs%2Fpull%2F330%2Fhead#[..])
```
epage, line 286: Oof, slashes in query strings need percent encoding? Thats too bad.

epage (approved):

epage:
Also important to call out that we respect MSRV for this so we don't generate a lockfile you can't use.

@bors r+

tests/testsuite/git.rs
weihanglo: This makes me wonder if we did the wrong encoding…

weihanglo:
```
curl: (28) Failed to connect to github.com port 443 after 135371 ms: Connection timed out

gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

Network error.

@bors retry

tests/testsuite/git.rs
epage: Huh, the wikipedia article i read didn't mention slashes being allowed.

PR #14598, opened 2024-09-26
fix: correct error count for `cargo check --message-format json`

### LINKED ISSUE #14472: Inconsistent summary on error count with --message-format=json
### Problem

`cargo check` emits a good summary on error count, but with `--message-format json`, the count number is increased by 1 because it seems to include the error summary itself as a new error. This also influences summary count of clippy in the same way.

As for warning count, the former reports the count, but with json format, no count is reported.

### Steps

Repoducible repo: https://github.com/os-checker/MRE-cargo-error-count

```rust
// src/main.rs
#![deny(warnings)]
fn main() {
    let a = 1;
}
```

```rust
$ cargo check
error: could not compile `cargo-error-count` (bin "cargo-error-count") due to 1 previous error

$ cargo check --message-format json
error: could not compile `cargo-error-count` (bin "cargo-error-count") due to 2 previous errors
```

Note: this inconsistency might be a bit different on warning report

```rust
# with no #![deny(warnings)]
$ cargo check
warning: `cargo-error-count` (bin "cargo-error-count") generated 1 warning

# with no #![deny(warnings)]
$ cargo check --message-format json
# no count reported
```

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
cargo 1.82.0-nightly (0d8d22f83 2024-08-08)
release: 1.82.0-nightly
commit-hash: 0d8d22f83b066503f6b2b755925197e959e58b4f
commit-date: 2024-08-08
libcurl: 8.9.0-DEV (sys:0.4.74+curl-8.9.0 vendored ssl:OpenSSL/3.3.1)
ssl: OpenSSL 3.3.1 4 Jun 2024
```

yichi170:
Hi! This is my first time contributing to Cargo. If there is anything that I need to do, please let me know!
(I'm not sure whether the commit message is aligned with the Cargo's convention. If it doesn't, I'm willing to modify it!)

This PR resolves the issue with incorrect error count and ensures warnings are correctly displayed when using `cargo check --message-format json`.

Fixes #14472

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1867,10 +1867,19 @@ fn on_stderr_line_inner(
 
     #[derive(serde::Deserialize)]
     struct CompilerMessage {
+        message: String,
         level: String,
     }
-    if let Ok(message) = serde_json::from_str::<CompilerMessage>(compiler_message.get()) {
-        count_diagnostic(&message.level, options);
+
+    if let Ok(msg) = serde_json::from_str::<CompilerMessage>(compiler_message.get()) {
+        if msg.message.starts_with("aborting due to")
+            || msg.message.ends_with("warning emitted")
+            || msg.message.ends_with("warnings emitted")
+        {
+            // Skip this line; we'll print our own summary at the end.
+            return Ok(true);
+        }
+        count_diagnostic(&msg.level, options);
     }
 
     let msg = machine_message::FromCompiler {
--- a/tests/testsuite/build.rs
+++ b/tests/testsuite/build.rs
@@ -95,7 +95,7 @@ fn cargo_fail_with_no_stderr() {
         .with_status(101)
         .with_stderr_data(str![[r#"
 [COMPILING] foo v0.5.0 ([ROOT]/foo)
-[ERROR] could not compile `foo` (bin "foo") due to 2 previous errors
+[ERROR] could not compile `foo` (bin "foo") due to 1 previous error
 
 "#]])
         .run();
--- a/tests/testsuite/pkgid.rs
+++ b/tests/testsuite/pkgid.rs
@@ -330,12 +330,6 @@ fn pkgid_json_message_metadata_consistency() {
   {
     "manifest_path": "[ROOT]/foo/Cargo.toml",
     "package_id": "path+[ROOTURL]/foo#0.5.0",
-    "reason": "compiler-message",
-    "...": "{...}"
-  },
-  {
-    "manifest_path": "[ROOT]/foo/Cargo.toml",
-    "package_id": "path+[ROOTURL]/foo#0.5.0",
     "reason": "compiler-artifact",
     "...": "{...}"
   },


src/cargo/core/compiler/mod.rs
weihanglo, line 1: Thanks for the contribution. 

We encourage adding tests when fixing bugs. And when adding tests, we recommend following [a variant of atomic commit pattern](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request) — the first commit adds a test showing the current behavior and passes CI. The next commit implements the fix and updates the test showing how the behavior has changed. This pattern could help us identify there is really a bug as the first commit can be considered as a minimal reproduction of that bug.

src/cargo/core/compiler/mod.rs
```diff
@@ -1865,12 +1865,30 @@ fn on_stderr_line_inner(
...
+                .map(|child| {
+                    child
+                        .spans
+                        .iter()
+                        .filter_map(|span| span.suggestion_applicability)
+                        .any(|app| app == Applicability::MachineApplicable)
+                })
+                .any(|b| b);
+            count_diagnostic(&msg.level, options);
+            state.emit_diag(msg.level, rendered, machine_applicable)?;
```
weihanglo, line 1890: I am not sure if we need to emit diagnostics here. Could you give an example for the necessity of this?

(The example can be done as a test case, if applicable)

epage:
> Hi! This is my first time contributing to Cargo. If there is anything that I need to do, please let me know!

Welcome!

If you ever want some face-time with a Cargo team member, we hold office hours on jitsi, see https://github.com/rust-lang/cargo/wiki/Office-Hours

It would be great if this change could have a test!  Looking around, my guess is one of the following files will have relevant tests to put this next to: `tests/testsuite/check.rs`, `tests/testsuite/message_format.rs`, or `tests/testsuite/messages.rs`.

Something we've found very helpful when writing tests is to add them in a commit before a fix, with every commit passing the test.  See https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request for more details.

> (I'm not sure whether the commit message is aligned with the Cargo's convention. If it doesn't, I'm willing to modify it!)

We don't have an official style though some of us use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)

yichi170:
Thanks for providing the information! I would add the test for this modification.

src/cargo/core/compiler/mod.rs
yichi170: I'm not sure whether this diagnostic is expected or not, but in issue #14472, running `cargo check` will emit a warning, but `cargo check --message-format json` won't emit a warning. Do you think this should be aligned?

weihanglo: The message format is explicitly chosen as JSON, so I don't expect any message to be emitted in a human-readable format.

yichi170:
I think the original test already cover this PR, instead of creating one, I modify the test to make it produce the correct result.

src/cargo/core/compiler/mod.rs
```diff
@@ -1865,12 +1865,15 @@ fn on_stderr_line_inner(
         return Ok(true);
     }
 
-    #[derive(serde::Deserialize)]
-    struct CompilerMessage {
-        level: String,
-    }
-    if let Ok(message) = serde_json::from_str::<CompilerMessage>(compiler_message.get()) {
-        count_diagnostic(&message.level, options);
+    if let Ok(msg) = serde_json::from_str::<CompilerMessage>(compiler_message.get()) {
```
weihanglo, line 1868: Instead of reusing a larger struct definition, should we just add an extra field `message`, and change `count_diagnostic` to accept three args `(level, message, options)` ?

Although having no data, my intuition tells me that skipping unused fields is generally more performant.


```json5
{
  "rendered": "error: expected one of `!` or `::`, found `<eof>`\n --> src/main.rs:3:1\n  |\n3 | xxx\n  | ^^^ expected one of `!` or `::`\n\n",
  "children": [], // may contain nested diagnostics
  "level": "error",
  "message": "expected one of `!` or `::`, found `<eof>`",
}
```

compares to

```json5
{
  "level": "error",
  "message": "expected one of `!` or `::`, found `<eof>`",
}
```

weihanglo (commented):
Yeah existing tests seem to cover the case.

src/cargo/core/compiler/mod.rs
yichi170: Do you mean I should check the message in `count_diagnostic` instead of checking before passing the `(level, options` to `count_diagnostic`?

weihanglo: Ignore the `count_diagnostic` part. I have no idea what I was talking about.

I meant that we may want to keep a separate `CompilerMessage` struct, since the current patch deserializes more fields than needed.

yichi170: fixed!

weihanglo (approved):

weihanglo:
Awesome! Thanks for helping.

@bors r+

PR #14599, opened 2024-09-26
docs(charter): Declare new Intentional Artifacts as 'small' changes

epage:
### What does this PR try to resolve?

The default stance for new "Intentional Artifact" crates is that an RFC is needed, see [Crate ownership policy](https://forge.rust-lang.org/policies/crate-ownership.html).
However, it gives room for team's to have their charter define the process.  As we don't generally need wide input for these decisions, I'm proposing we treat these as "small" changes and only require an FCP.

PR #14600, opened 2024-09-26
docs: Declare support level for each crate in our Charter / docs

epage:

This is to bring us into conformance with the [Rust crate ownership policy](https://forge.rust-lang.org/policies/crate-ownership.html).

Items of note
- `cargo` does not have a status specified on it.  `cargo install cargo` shouldn't be done and `cargo add cargo` is "internal" but putting in our README that Cargo is "internal" feels weird from a messaging perspective.
- `cargo-credential-1password` is declared as Experimental as it is intended for the community but I was unsure if we wanted to commit to full support for it.  In my mind, the ideal thing to do would be to expatriate this to 1password.
- `home` is declared as Internal despite its wide use within the ecosystem.
- `cargo-credential` is declared as Intentional as its an API intended for the wider ecosystem and I didn't see a reason to declare it experimental.
- `cargo-platform`, `cargo-util-schemas`, and `crates-io` are declared as Intentional as they are both used internally and intended for others to use for logic that integrates with cargo/registries. I wondered about these being Experimental or Internal instead.

I was mixed on what to do with unpublished crates (and didn't fully check what all is unpublished). 
I ended up only skipping xtasks.


epage:
While the policy asks for an RFC for making things Intentional and #14599 isn't approved yet, I'm going to at least start this attempting to codify where things are at, rather than set new policy.

@rfcbot fcp merge

weihanglo (approved):

crates/cargo-platform/README.md
```diff
@@ -0,0 +1,2 @@
+> This crate is maintained by the Cargo team for use by the wider
+> ecosystem. This crate follows semver compatibility for its APIs.
```
ehuss, line 2: I'm on the fence here. My intention when I created it was to be more "internal". I was thinking that if others found it useful, we could expand that to a more "experimental" status. I didn't know if anyone would actually have a use for it, and it wasn't really something I wanted to put effort into. 

Also, as evident by the lack of a README, this isn't really up to my standards for an intentional artifact that we intend to support.

It looks like it has a few dependents now, though: https://crates.io/crates/cargo-platform/reverse_dependencies.

I'm fine with elevating this to be "intentional" if ya'll agree. It has required almost no maintenance effort, and I don't expect us to delete it, and we already have semver checks. I just wanted to make clear what I was originally thinking.

(We probably should eventually add a real README, though 🤣)

ehuss (commented):
Thanks for putting this together, it looks great!

ehuss:
> `cargo` does not have a status specified on it. 

This is probably should say *something*. I'm personally fine with putting it in the root README. Can you say more about your concern about the messaging? Is it related to the README having different audiences?

crates/cargo-platform/README.md
epage: Thanks for the clarification!

I think its reasonable for the Rust Project to provide an API for this micro-DSL.  My main concern is what shape any of that should take and thats where I feel like the commitment of "Intentional" is strong.  Granted, the definition of "Intentional" is strong enough I'm hesitant to mark *anything* intentional but anything else would also be too weak / scare people away.

epage:
> This is probably should say something. I'm personally fine with putting it in the root README. Can you say more about your concern about the messaging? Is it related to the README having different audiences?

I think thats a way of summing it up.  If someone goes to the Cargo repo to learn about Cargo and sees a big banner saying that Cargo is experimental or internal, that will send the wrong message.  However, we do want to send a message like that to anyone who builds the binary directly or tries to depend on it as a library.

ehuss:
I think it should be fine to include the notice in the root README with the right wording to make it clear what it means. The README already has some low-level documentation for building cargo, which is for a very limited audience.

epage:
I've done a pass at wording for `cargo` in both `README.md` and `src/cargo/lib.rs`

README.md
```diff
@@ -9,6 +9,13 @@ Cargo downloads your Rust project’s dependencies and compiles your project.
 [The Cargo Book]: https://doc.rust-lang.org/cargo/
 [Cargo Contributor Guide]: https://rust-lang.github.io/cargo/contrib/
 
+> The Cargo binary distributed through with Rust is maintained by the Cargo
```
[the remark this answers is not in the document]
epage: That section heavily implies "building the binary from source" and doesn't really cover the "depending on it as a library" case.

I feel like a lot of this should be re-worked, mostly by moving in to the contributor guide.  I recommend we remove the install instructions and defer re-working the compilation instructions.

weihanglo (approved):
Since Eric's concern has been addressed. I am going to merge this.

If there are other concerns, we can address later.

PR #14602, opened 2024-09-26
test: relax panic output assertion

weihanglo:

rust-lang/rust#122565 adds a new line to thread panic output. To make the current test suites works on stable, beta, and nightly, this relaxes the assertion around that by globbing everything.

Switch to rust-lang/rust#122565 and build a stage0 std, then use it to run these Cargo tests.

Linking to stage0 toolchain should work: https://rustc-dev-guide.rust-lang.org/building/how-to-build-and-run.html#creating-a-rustup-toolchain

### Additional information

An extra newline from the beginning of the panic message seems a bit odd to me, though I understand until we find a better approach, it is probably the only way to fix the message interleaved issue.

### DIFF
--- a/tests/testsuite/freshness.rs
+++ b/tests/testsuite/freshness.rs
@@ -1891,6 +1891,7 @@ test t1 ... FAILED
 failures:
 
 ---- t1 stdout ----
+...
 thread 't1' panicked at src/lib.rs:8:21:
 assertion `left == right` failed: doit assert failure
   left: 2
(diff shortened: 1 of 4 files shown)


ehuss:
I'm concerned about these tests matching on the output of libtest. Is there a reason the full output is being checked here? The originals were intentionally done to avoid that, because otherwise it causes a conflict whenever the libs/compiler people want to change the output of any of these strings.

weihanglo:
No. I'll try moving to minimal assertions for them.

PR #14605, opened 2024-09-27
git: do not validate submodules of fresh checkouts

### LINKED ISSUE #14603: Loading of a git source is slow due to always checking if the submodule is updated
When you have some git dependencies or patches, loading them takes a long time

For the traces from #14238, notice how slow the first `resolve` is compared to the second.  A lot of the time is taken up checking each git dependencies submodules
image

Note: this slow down is most relevant for tooling like `rust-analyzer` or rust-defined completions when it may be called in interactive contexts without any compilation happening or "cargo script" where `cargo build` is performed every time you run your script.

Related: #14395

osiewicz:
Fixes #14603

As is, we unconditionally validate freshness of the submodules of a checkout, even though we could assume that a fresh checkout has to have up-to-date submodules as well.

N/A

### Additional information

N/A

### DIFF
--- a/src/cargo/sources/git/utils.rs
+++ b/src/cargo/sources/git/utils.rs
@@ -181,9 +181,14 @@ impl GitDatabase {
             .filter(|co| co.is_fresh())
         {
             Some(co) => co,
-            None => GitCheckout::clone_into(dest, self, rev, gctx)?,
+            None => {
+                let (checkout, guard) = GitCheckout::clone_into(dest, self, rev, gctx)?;
+                checkout.update_submodules(gctx)?;
+                guard.mark_ok()?;
+                checkout
+            }
         };
-        checkout.update_submodules(gctx)?;
+
         Ok(checkout)
     }
 
@@ -280,7 +285,7 @@ impl<'a> GitCheckout<'a> {
         database: &'a GitDatabase,
         revision: git2::Oid,
         gctx: &GlobalContext,
-    ) -> CargoResult<GitCheckout<'a>> {
+    ) -> CargoResult<(GitCheckout<'a>, CheckoutGuard)> {
         let dirname = into.parent().unwrap();
         paths::create_dir_all(&dirname)?;
         if into.exists() {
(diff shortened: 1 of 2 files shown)


src/cargo/sources/git/utils.rs
```diff
@@ -181,9 +181,13 @@ impl GitDatabase {
...
             Some(co) => co,
-            None => GitCheckout::clone_into(dest, self, rev, gctx)?,
+            None => {
+                let checkout = GitCheckout::clone_into(dest, self, rev, gctx)?;
+                checkout.update_submodules(gctx)?;
+                checkout
+            }
         };
-        checkout.update_submodules(gctx)?;
+
```
epage, line 190: `is_fresh` is determined by the presence of the `CHECKOUT_READY_LOCK` file.  That file is written in `reset` which is called at the end of `clone_into`.  As-written, if the submodule update fails or gets interrupted, we won't be able to detect that and recover on the next invocation.  We need to make sure that `CHECKOUT_READY_LOCK` is not written to until after the submodule update to make sure we can recover gracefully.

epage:
We likely should update `_corrupted_checkout` test to have a submodule to show that we recovery gracefully on interruption,

src/cargo/sources/git/utils.rs
```diff
@@ -80,6 +80,25 @@ pub struct GitCheckout<'a> {
     repo: git2::Repository,
 }
 
+/// See [`GitCheckout::reset`] for rationale on this type.
+struct CheckoutGuard {
+    ok_file: PathBuf,
+}
+
+impl CheckoutGuard {
+    #[must_use]
```
epage, line 89: Should this be put on the `struct` itself?

epage:
Could you update the commits to be how you want them reviewed and merged?

For example, I could see this being
- Test commit
- Refactor for checkout guard
- Improve performance by adjusting the scope of the checkout guard

epage (approved):

PR #14607, opened 2024-09-27
Update cc to 1.1.22

ChrisDenton:
This version of `cc` contains [a fix](https://github.com/rust-lang/cc-rs/releases/tag/cc-v1.1.22) to prevent spurious rebuilds. Hopefully this should help avoid the CI issues rustc has been having.


weihanglo:
It sounds like rust-lang/rust build is affected. I guess we should include it in <https://github.com/rust-lang/rust/pull/130929>?

ChrisDenton:
That would be ideal, imho. Though the CI failure rate isn't as bad at the moment so it'll still be tolerable if it slips to the next update.

PR #14609, opened 2024-09-28
Migrate Cargo's &Option<T> into Option<&T>

nyurik:
As part of this, also modify `Option<&String>` to `Option<&str>`, same for `PathBuf`

Trying out my new lint https://github.com/rust-lang/rust-clippy/pull/13336 - according to the [video](https://www.youtube.com/watch?v=6c7pZYP_iIE), this could lead to some performance and memory optimizations.

Basic thoughts expressed in the video that seem to make sense:
* `&Option<T>` in an API breaks encapsulation:
  * caller must own T and move it into an Option to call with it
  * if returned, the owner must store it as Option<T> internally in order to return it
* Performance is subject to compiler optimization, but at the basics, `&Option<T>` points to memory that has `presence` flag + value, whereas `Option<&T>` by specification is always optimized to a single pointer.

### DIFF
--- a/src/cargo/core/source_id.rs
+++ b/src/cargo/core/source_id.rs
@@ -468,30 +468,30 @@ impl SourceId {
 
     /// Creates a new `SourceId` from this source with the given `precise`.
     pub fn with_git_precise(self, fragment: Option<String>) -> SourceId {
-        self.with_precise(&fragment.map(|f| Precise::GitUrlFragment(f)))
+        self.with_precise(fragment.map(|f| Precise::GitUrlFragment(f)))
     }
 
     /// Creates a new `SourceId` from this source without a `precise`.
     pub fn without_precise(self) -> SourceId {
-        self.with_precise(&None)
+        self.with_precise(None)
     }
 
     /// Creates a new `SourceId` from this source without a `precise`.
     pub fn with_locked_precise(self) -> SourceId {
-        self.with_precise(&Some(Precise::Locked))
+        self.with_precise(Some(Precise::Locked))
     }
 
     /// Creates a new `SourceId` from this source with the `precise` from some other `SourceId`.
     pub fn with_precise_from(self, v: Self) -> SourceId {
-        self.with_precise(&v.inner.precise)
+        self.with_precise(v.inner.precise.clone())
     }
 
-    fn with_precise(self, precise: &Option<Precise>) -> SourceId {
-        if &self.inner.precise == precise {
+    fn with_precise(self, precise: Option<Precise>) -> SourceId {
+        if self.inner.precise == precise {
             self
         } else {
             SourceId::wrap(SourceIdInner {
-                precise: precise.clone(),
+                precise,
                 ..(*self.inner).clone()
             })
         }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -359,7 +359,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-credential-1password"
-version = "0.4.4"
+version = "0.5.0"
 dependencies = [
  "cargo-credential",
  "serde",
--- a/benches/benchsuite/benches/global_cache_tracker.rs
+++ b/benches/benchsuite/benches/global_cache_tracker.rs
@@ -43,7 +43,7 @@ fn initialize_context() -> GlobalContext {
         false,
         false,
         false,
-        &None,
+        None,
         &["gc".to_string()],
         &[],
     )
--- a/benches/benchsuite/src/bin/capture-last-use.rs
+++ b/benches/benchsuite/src/bin/capture-last-use.rs
@@ -36,7 +36,7 @@ fn main() {
         false,
         false,
         false,
-        &None,
+        None,
         &["gc".to_string()],
         &[],
     )
--- a/benches/benchsuite/src/lib.rs
+++ b/benches/benchsuite/src/lib.rs
@@ -188,7 +188,7 @@ impl Fixtures {
             false,
             false,
             false,
-            &Some(self.target_dir()),
+            Some(&self.target_dir()),
             &[],
             &[],
         )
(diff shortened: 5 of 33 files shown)


src/cargo/core/source_id.rs
```diff
@@ -468,30 +468,30 @@ impl SourceId {
...
 
     /// Creates a new `SourceId` from this source with the `precise` from some other `SourceId`.
     pub fn with_precise_from(self, v: Self) -> SourceId {
-        self.with_precise(&v.inner.precise)
+        self.with_precise(v.inner.precise.clone())
     }
 
-    fn with_precise(self, precise: &Option<Precise>) -> SourceId {
-        if &self.inner.precise == precise {
+    fn with_precise(self, precise: Option<Precise>) -> SourceId {
```
nyurik, line 489: note that in this change the fn signature is changed to take the ownership - seems to be ok, but maybe less efficient?

src/cargo/core/source_id.rs
```diff
@@ -468,30 +468,30 @@ impl SourceId {
...
     /// Creates a new `SourceId` from this source without a `precise`.
     pub fn with_locked_precise(self) -> SourceId {
-        self.with_precise(&Some(Precise::Locked))
+        self.with_precise(Some(Precise::Locked))
     }
 
     /// Creates a new `SourceId` from this source with the `precise` from some other `SourceId`.
     pub fn with_precise_from(self, v: Self) -> SourceId {
-        self.with_precise(&v.inner.precise)
+        self.with_precise(v.inner.precise.clone())
```
nyurik, line 486: this seems wrong -- if both `v` and `self` are owned, why can't I take ownership of `v.inner.precise` ?

epage:
Note that we intentionally run with very few clippy lints enabled and there are times we intentionally do what this PR changes away from because it actually makes the code better and a change like this is unlikely to have a significant impact on performance.

This might make some parts nicer but having to make the connection between one API change and how it changes callers is hard when everything is bundled in one PR.

nyurik:
@epage do you disagree with the arguments presented in the video?  Or do you propose to simply break this PR into multiple ones to make reviewing easier?  If so, could you give some guidance on the partitioning. Thx!

epage:
The video is about API design and micro-optimizations without data to back it.

Those are good baseline principles to operate on but they hardly map to every case in reality.  In particular, `cargo` is not developed as a library, but as a binary that we don't block people from accessing the internals.  So API design principles are thrown out *for that package* and what matters is what helps with the implementation of Cargo.  Take https://github.com/rust-lang/cargo/blob/master/src/cargo/util/toml/targets.rs, that commits various "sins" around references, both `&Option<T>` and `Option<&Vec<T>>` rather than `Option<&[T]>`.  Thats because that is whats easiest for the caller.  And of all of the performance hits we take in manifest parsing, no one will notice the difference.

Particularly when you get into

> modify Option<&String> to Option<&str>, same for PathBuf

you hit limits.  Yes, you can change `&String` to `&str` and `&Vec<T>` to `&[T]` to reduce ownership restrictions but then what happens when you have `&Vec<String>`?  The only way to remove restrictions is to use `impl Iterator<Item=&str>` and thats not always convenient.

I'm not against people generally improving this as they make changes (putting it into its own commit of course) or having tools call out it for the dev to reconsider (which is why [I think we need a new lint level](https://internals.rust-lang.org/t/forbid-deny-warn-allow-and-notice/19986)), but I don't want to dig into a 33 file PR to see if there are places this made it worse.

weihanglo (commented):
> Note that we intentionally run with very few clippy lints enabled and there are times we intentionally do what this PR changes away from because it actually makes the code better and a change like this is unlikely to have a significant impact on performance.

To expand this a bit, without enabling the lint in Cargo's `[lints]` table, we may need to do it again in the future. I would suggest at least not to do it at this moment, and revisit when the lint is in clippy officially.

weihanglo:
> Trying out my new lint https://github.com/rust-lang/rust-clippy/pull/13336 - according to the [video](https://www.youtube.com/watch?v=6c7pZYP_iIE), this could lead to some performance and memory optimizations.

Does that mean you're looking for crate maintainers' feekback for the lint, or you just want to understand how the new lint impacts a medium-size codecase?

nyurik:
thx for feedback @weihanglo and @epage!  The lint is part of `pedantic`, and I do want to see how it impacts the codebases as well as get maintainers' perspectives.  Ed does bring valid point that this is more important to a crate's external API rather than internal function signatures, and performance benefits are yet to be determined.  The video's author made a lot of sense in their logic, but in this PR i did end up adding a lot of `.as_ref()` and `.as_deref()` instead of `&`, but at the same time a few of them were removed on the caller's side, plus using `None` instead of `&None` did look cleaner.  I think this will be important for the crate authors in building the right APIs, but internal usefulness and perf is TBD.

nyurik:
Micro-benchmarking is notoriously hard to do right.  I tried running this benchmark (assuming I got it done right), it shows about 11% difference.  If I change `Option<Vec<usize>>` to `Option<(usize, usize)>`, the difference grows to about 20%. 

```rust
use criterion::{criterion_group, criterion_main, Criterion};

#[inline(never)]
pub fn do_ref(val: &Option<Vec<usize>>) -> usize {
    match val {
        None => 42,
        Some(val) => val.iter().sum(),
    }
}

#[inline(never)]
pub fn do_as_ref(val: Option<&Vec<usize>>) -> usize {
    match val {
        None => 42,
        Some(val) => val.iter().sum(),
    }
}

fn criterion_benchmark(c: &mut Criterion) {
    let data: Option<Vec<usize>> = Some(vec![0; 2]);
    c.bench_function("ref", |b| b.iter(|| do_ref(&data)));
    c.bench_function("as_ref", |b| b.iter(|| do_as_ref(data.as_ref())));
}

criterion_group!(benches, criterion_benchmark);
criterion_main!(benches);
```

epage:
To be clear, I'm less interested in micro-benchmarks because this kind of performance chance is unlikely to beat out the algorithmic complexity challenges in Cargo.

PR #14614, opened 2024-09-29
Add more SAT resolver tests

x-hgg-x:

This is a follow-up of #14583.

* Commit 1 splits tests into smaller modules.
* Commit 2 adds more helper trait methods.
* Commit 3 refactors computation of the SAT clauses, by introducing intermediate boolean variables for each dependency and each dependency feature declared in a package.
The old behavior was incorrect: package features specified in an optional dependency were activated if the package was activated, even if the dependency wasn't activated.
* Commit 4 add tests from https://github.com/Eh2406/pubgrub-crates-benchmark/tree/main/out/index_ron.

r? Eh2406


crates/resolver-tests/tests/pubgrub.rs
```diff
@@ -0,0 +1,453 @@
...
+            "a",
+            vec!["b".opt(), "x".opt()],
+            &[("b", &["x"]), ("default", &["b/native"])],
+        ),
+        pkg_dep_with("b", vec![], &[("native", &[])]),
+    ]);
+
+    let deps = vec!["a".with(&["default"])];
+
+    // Bug in the dependency resolver: feature `b` of package `a` should not be activated by `b/native`
```
x-hgg-x, line 365: I think this is a bug in the dependency resolver. Has an issue already been created for this ?

Eh2406:
First off, thank you so much for splitting the PR into independently understandable commits. It makes reviewing this PR much more feasible. For example the first 2 commits look 100% and ready for +1. If the other commits require extensive discussion, maybe we should get those first two merged asap.

Secondly, thank you for giving descriptive names to the pubgrub tests! Would you mind if I took your good names back to that project? (Or even better do you want to do a PR?)

crates/resolver-tests/tests/pubgrub.rs
Eh2406: I thought it was a bug when I first came across it as well. But it is intentional documented behavior: https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Odd.20cargo.20resolver.20behavior

crates/resolver-tests/src/helpers.rs
```diff
@@ -0,0 +1,247 @@
...
+
+pub fn pkg_dep_link<T: ToPkgId>(name: T, link: &str, dep: Vec<Dependency>) -> Summary {
+    Summary::new(name.to_pkgid(), dep, &BTreeMap::new(), Some(link), None).unwrap()
+}
+
+pub fn local_pkg_id(name: &str) -> PackageId {
+    PackageId::try_new(
+        name,
+        "1.0.0",
+        SourceId::for_path(Path::new("/some-local-path")).unwrap(),
```
Eh2406, line 152: In pubgrub-crates-benchmark I needed to do something a little more complicated to make it work on Windows. https://github.com/Eh2406/pubgrub-crates-benchmark/blob/main/src/cargo_resolver.rs#L131 

In retrospect I wonder if it would be easier to use `SourceId::for_registry(&"https://other.example.com".into_url().unwrap()).unwrap()` in both projects.

x-hgg-x:
> Would you mind if I took your good names back to that project? (Or even better do you want to do a PR?)

I posted a PR: https://github.com/Eh2406/pubgrub-crates-benchmark/pull/10.

Eh2406:
I meant to do a new round of review today, looking at your new changes and being sure your old changes still made sense with fresh eyes and making sure the tests had translated correctly. (If I had to do the translations I would've made more than a few copy/paste errors.) Unfortunately, had a family emergency come up overnight. Luckily, everyone's okay. The only casualty was my sleep/concentration schedule. With the holy days the end of this week, I cannot guarantee availability of review time. I will make time for this Monday, if I have not gotten to it sooner. Given how enthusiastic I am to see your progress on this work, I am very sorry at how I'm slowing you down.

crates/resolver-tests/src/sat.rs
```diff
@@ -59,40 +61,130 @@ fn sat_at_most_one_by_key<K: std::hash::Hash + Eq>(
...
+>;
+
+fn create_dependencies_vars<'a>(
+    solver: &mut varisat::Solver<'_>,
+    pkg_var: varisat::Var,
+    pkg_dependencies: &'a [Dependency],
+    pkg_features: &FeatureMap,
+) -> (DependencyVarMap<'a>, DependencyFeatureVarMap<'a>) {
+    let mut var_for_is_dependencies_used = HashMap::<_, HashMap<_, _>>::new();
+    let mut var_for_is_dependencies_features_used = HashMap::<_, HashMap<_, HashMap<_, _>>>::new();
```
Eh2406, line 79: Nit: these can use the type aliases you've declared above. `DependencyVarMap::new()`

Eh2406:
SAT is slippery, I wouldn't be surprised if we find more weird cases as we add more test cases/fuzzing. But I don't know what they are. So for now this looks good. Thank you!

@bors r+

x-hgg-x:
The `lockfile` and `msrv` test in the CI are failing, the lockfile has changed.

x-hgg-x:
It should be good now after a rebase on `master`.

PR #11436, opened 2022-11-28
Emit progress state escape codes

albertlarsan68:
Emit progress state escape codes when drawing progress bar.

Fixes #11432

PR #8165, opened 2020-04-27
Add a term option to configure the progress bar

mchernyavsky:
Closes https://github.com/rust-lang/cargo/issues/7587.

PR #14615, opened 2024-09-29
Add terminal integration via ANSI OSC 9;4 sequences

### LINKED ISSUE #11432: Use ANSI codes to report progress
### Problem

When the terminal is not in focus, there is no way to know the progress state of the compilation.

### Proposed Solution

There exist escape codes that allow control of a "progress meter", that for example is shown on the taskbar in Windows.
Winget uses them, and the implementation can be seen at https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerCLICore/VTSupport.cpp. It mostly consists of adding the correct progress number, and clean up after the end.

### Notes

I volunteer to make this change to this repo.

Gordon01:

A few terminal emulators support progress output to Windows taskbar. `winget` uses this to show install progress. 

Notably, Windows Terminal [recently (2020) added support](https://github.com/microsoft/terminal/pull/8055) for ANSI codes [specified](https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC) in ConEmu (another terminal emulator for Windows) documentation. Also, in "[Learn Windows](https://learn.microsoft.com/en-us/windows/terminal/tutorials/progress-bar-sequences)".

I've found the previous attempt to add this feature: https://github.com/rust-lang/cargo/pull/11436 

As per @weihanglo's request, I've added the config option to enable/disable this feature. **It's enabled on supported terminal emulators.**

Fixes #11432 

FCP: https://github.com/rust-lang/cargo/pull/14615#issuecomment-2546346489

Run `cargo build` in Windows Terminal with configuration option `term.progress.taskbar` set to `true`.

### Not sure

- [x] Should all the code be `#[cfg(windows)]`? Probably no, because the feature is also usable in WSL.
> Solved by introducing heuristic based on environment variable set by terminal

- [ ] If Ctrl+C is pressed, a progressbar will stay in a last state forever (shown in the ConEmu video). `winget` is also behaves like alike. I've experimented with `ctrl_c handler` and it's totally fixable.
- [x] `Enabled` is a sensible default for WSL because it works on linux builds in Windows Terminal too
> Solved by introducing heuristic based on environment variable set by terminal

- [x] Downloading stage may produce unpleasant blinking due to a rapid 0-100 changes 
> Solved by not displaying bar when downloading and using indeterminate state in other cases so amination don't reset

### TLDR

* An `term.progress.taskbar` option with bool type is added
* On Windows Terminal and ConEmu is enabled by default
* If enabled reports build progress to taskbar icon and/or tab header

### Videos

https://github.com/user-attachments/assets/48bb648a-e819-490e-b3ac-3502bc5f2f3a

https://github.com/user-attachments/assets/1d7ddf7a-34dd-4db1-b654-e64d7170798e

### DIFF
--- a/src/cargo/core/shell.rs
+++ b/src/cargo/core/shell.rs
@@ -56,6 +56,7 @@ impl Shell {
                 stderr_tty: std::io::stderr().is_terminal(),
                 stdout_unicode: supports_unicode(&std::io::stdout()),
                 stderr_unicode: supports_unicode(&std::io::stderr()),
+                stderr_term_integration: supports_term_integration(&std::io::stderr()),
             },
             verbosity: Verbosity::Verbose,
             needs_clear: false,
@@ -119,6 +120,18 @@ impl Shell {
         match self.output {
             ShellOut::Stream { stderr_tty, .. } => stderr_tty,
             _ => false,
+        }
+    }
+
+    pub fn is_err_term_integration_available(&self) -> bool {
+        if let ShellOut::Stream {
+            stderr_term_integration,
+            ..
+        } = self.output
+        {
+            stderr_term_integration
+        } else {
+            false
         }
     }
 
@@ -426,6 +439,7 @@ enum ShellOut {
         hyperlinks: bool,
         stdout_unicode: bool,
         stderr_unicode: bool,
+        stderr_term_integration: bool,
     },
 }
 
@@ -573,6 +587,16 @@ fn supports_hyperlinks() -> bool {
     }
 
     supports_hyperlinks::supports_hyperlinks()
+}
+
+/// Determines whether the terminal supports ANSI OSC 9;4.
+#[allow(clippy::disallowed_methods)] // Read environment variables to detect terminal
+fn supports_term_integration(stream: &dyn IsTerminal) -> bool {
+    let windows_terminal = std::env::var("WT_SESSION").is_ok();
+    let conemu = std::env::var("ConEmuANSI").ok() == Some("ON".into());
+    let wezterm = std::env::var("TERM_PROGRAM").ok() == Some("WezTerm".into());
+
+    (windows_terminal || conemu || wezterm) && stream.is_terminal()
 }
 
 pub struct Hyperlink<D: fmt::Display> {
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -2832,6 +2832,8 @@ pub struct TermConfig {
 pub struct ProgressConfig {
     pub when: ProgressWhen,
     pub width: Option<usize>,
+    /// Communicate progress status with a terminal
+    pub term_integration: Option<bool>,
 }
 
 #[derive(Debug, Default, Deserialize)]
@@ -2864,10 +2866,12 @@ where
                 "auto" => Ok(Some(ProgressConfig {
                     when: ProgressWhen::Auto,
                     width: None,
+                    term_integration: None,
                 })),
                 "never" => Ok(Some(ProgressConfig {
                     when: ProgressWhen::Never,
                     width: None,
+                    term_integration: None,
                 })),
                 "always" => Err(E::custom("\"always\" progress requires a `width` key")),
                 _ => Err(E::unknown_variant(s, &["auto", "never"])),
@@ -2889,6 +2893,7 @@ where
             if let ProgressConfig {
                 when: ProgressWhen::Always,
                 width: None,
+                ..
             } = pc
             {
                 return Err(serde::de::Error::custom(
--- a/src/cargo/util/progress.rs
+++ b/src/cargo/util/progress.rs
@@ -74,6 +74,108 @@ struct Format {
     style: ProgressStyle,
     max_width: usize,
     max_print: usize,
+    term_integration: TerminalIntegration,
+}
+
+/// Controls terminal progress integration via OSC sequences.
+struct TerminalIntegration {
+    enabled: bool,
+    error: bool,
+}
+
+/// A progress status value printable as an ANSI OSC 9;4 escape code.
+#[cfg_attr(test, derive(PartialEq, Debug))]
+enum StatusValue {
+    /// No output.
+    None,
+    /// Remove progress.
+    Remove,
+    /// Progress value (0-100).
+    Value(f64),
+    /// Indeterminate state (no bar, just animation)
+    Indeterminate,
+    /// Progress value in an error state (0-100).
+    Error(f64),
+}
+
+enum ProgressOutput {
+    /// Print progress without a message
+    PrintNow,
+    /// Progress, message and progress report
+    TextAndReport(String, StatusValue),
+    /// Only progress report, no message and no text progress
+    Report(StatusValue),
+}
+
+impl TerminalIntegration {
+    #[cfg(test)]
+    fn new(enabled: bool) -> Self {
+        Self {
+            enabled,
+            error: false,
+        }
+    }
+
+    /// Creates a `TerminalIntegration` from Cargo's configuration.
+    /// Autodetect support if not explicitly enabled or disabled.
+    fn from_config(gctx: &GlobalContext) -> Self {
+        let enabled = gctx
+            .progress_config()
+            .term_integration
+            .unwrap_or_else(|| gctx.shell().is_err_term_integration_available());
+
+        Self {
+            enabled,
+            error: false,
+        }
+    }
+
+    fn progress_state(&self, value: StatusValue) -> StatusValue {
+        match (self.enabled, self.error) {
+            (true, false) => value,
+            (true, true) => match value {
+                StatusValue::Value(v) => StatusValue::Error(v),
+                _ => StatusValue::Error(100.0),
+            },
+            (false, _) => StatusValue::None,
+        }
+    }
+
+    pub fn remove(&self) -> StatusValue {
+        self.progress_state(StatusValue::Remove)
+    }
+
+    pub fn value(&self, percent: f64) -> StatusValue {
+        self.progress_state(StatusValue::Value(percent))
+    }
+
+    pub fn indeterminate(&self) -> StatusValue {
+        self.progress_state(StatusValue::Indeterminate)
+    }
+
+    pub fn error(&mut self) {
+        self.error = true;
+    }
+}
+
+impl std::fmt::Display for StatusValue {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // From https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC
+        // ESC ] 9 ; 4 ; st ; pr ST
+        // When st is 0: remove progress.
+        // When st is 1: set progress value to pr (number, 0-100).
+        // When st is 2: set error state in taskbar, pr is optional.
+        // When st is 3: set indeterminate state, pr is ignored.
+        // When st is 4: set paused state, pr is optional.
+        let (state, progress) = match self {
+            Self::None => return Ok(()), // No output
+            Self::Remove => (0, 0.0),
+            Self::Value(v) => (1, *v),
+            Self::Indeterminate => (3, 0.0),
+            Self::Error(v) => (2, *v),
+        };
+        write!(f, "\x1b]9;4;{state};{progress:.0}\x1b\\")
+    }
 }
 
 impl<'gctx> Progress<'gctx> {
@@ -126,6 +228,7 @@ impl<'gctx> Progress<'gctx> {
                     // 50 gives some space for text after the progress bar,
                     // even on narrow (e.g. 80 char) terminals.
                     max_print: 50,
+                    term_integration: TerminalIntegration::from_config(gctx),
                 },
                 name: name.to_string(),
                 done: false,
@@ -223,7 +326,7 @@ impl<'gctx> Progress<'gctx> {
     /// calling it too often.
     pub fn print_now(&mut self, msg: &str) -> CargoResult<()> {
         match &mut self.state {
-            Some(s) => s.print("", msg),
+            Some(s) => s.print(ProgressOutput::PrintNow, msg),
             None => Ok(()),
         }
     }
@@ -232,6 +335,13 @@ impl<'gctx> Progress<'gctx> {
     pub fn clear(&mut self) {
         if let Some(ref mut s) = self.state {
             s.clear();
+        }
+    }
+
+    /// Sets the progress reporter to the error state.
+    pub fn indicate_error(&mut self) {
+        if let Some(s) = &mut self.state {
+            s.format.term_integration.error()
         }
     }
 }
@@ -269,6 +379,11 @@ impl Throttle {
 impl<'gctx> State<'gctx> {
     fn tick(&mut self, cur: usize, max: usize, msg: &str) -> CargoResult<()> {
         if self.done {
+            write!(
+                self.gctx.shell().err(),
+                "{}",
+                self.format.term_integration.remove()
+            )?;
             return Ok(());
         }
 
@@ -280,21 +395,30 @@ impl<'gctx> State<'gctx> {
         // return back to the beginning of the line for the next print.
         self.try_update_max_width();
         if let Some(pbar) = self.format.progress(cur, max) {
-            self.print(&pbar, msg)?;
+            self.print(pbar, msg)?;
         }
         Ok(())
     }
 
-    fn print(&mut self, prefix: &str, msg: &str) -> CargoResult<()> {
+    fn print(&mut self, progress: ProgressOutput, msg: &str) -> CargoResult<()> {
         self.throttle.update();
         self.try_update_max_width();
 
+        let (mut line, report) = match progress {
+            ProgressOutput::PrintNow => (String::new(), None),
+            ProgressOutput::TextAndReport(prefix, report) => (prefix, Some(report)),
+            ProgressOutput::Report(report) => (String::new(), Some(report)),
+        };
+
         // make sure we have enough room for the header
         if self.format.max_width < 15 {
+            // even if we don't have space we can still output progress report
+            if let Some(tb) = report {
+                write!(self.gctx.shell().err(), "{tb}\r")?;
+            }
             return Ok(());
         }
 
-        let mut line = prefix.to_string();
         self.format.render(&mut line, msg);
         while line.len() < self.format.max_width - 15 {
             line.push(' ');
@@ -305,7 +429,11 @@ impl<'gctx> State<'gctx> {
             let mut shell = self.gctx.shell();
             shell.set_needs_clear(false);
             shell.status_header(&self.name)?;
-            write!(shell.err(), "{}\r", line)?;
+            if let Some(tb) = report {
+                write!(shell.err(), "{line}{tb}\r")?;
+            } else {
+                write!(shell.err(), "{line}\r")?;
+            }
             self.last_line = Some(line);
             shell.set_needs_clear(true);
         }
@@ -314,6 +442,12 @@ impl<'gctx> State<'gctx> {
     }
 
     fn clear(&mut self) {
+        // Always clear the progress report
+        let _ = write!(
+            self.gctx.shell().err(),
+            "{}",
+            self.format.term_integration.remove()
+        );
         // No need to clear if the progress is not currently being displayed.
         if self.last_line.is_some() && !self.gctx.shell().is_cleared() {
             self.gctx.shell().err_erase_line();
@@ -331,7 +465,7 @@ impl<'gctx> State<'gctx> {
 }
 
 impl Format {
-    fn progress(&self, cur: usize, max: usize) -> Option<String> {
+    fn progress(&self, cur: usize, max: usize) -> Option<ProgressOutput> {
         assert!(cur <= max);
         // Render the percentage at the far right and then figure how long the
         // progress bar is
@@ -339,11 +473,21 @@ impl Format {
         let pct = if !pct.is_finite() { 0.0 } else { pct };
         let stats = match self.style {
             ProgressStyle::Percentage => format!(" {:6.02}%", pct * 100.0),
-            ProgressStyle::Ratio => format!(" {}/{}", cur, max),
+            ProgressStyle::Ratio => format!(" {cur}/{max}"),
             ProgressStyle::Indeterminate => String::new(),
         };
+        let report = match self.style {
+            ProgressStyle::Percentage | ProgressStyle::Ratio => {
+                self.term_integration.value(pct * 100.0)
+            }
+            ProgressStyle::Indeterminate => self.term_integration.indeterminate(),
+        };
+
         let extra_len = stats.len() + 2 /* [ and ] */ + 15 /* status header */;
         let Some(display_width) = self.width().checked_sub(extra_len) else {
+            if self.term_integration.enabled {
+                return Some(ProgressOutput::Report(report));
+            }
             return None;
         };
 
@@ -371,7 +515,7 @@ impl Format {
         string.push(']');
         string.push_str(&stats);
 
-        Some(string)
+        Some(ProgressOutput::TextAndReport(string, report))
     }
 
     fn render(&self, string: &mut String, msg: &str) {
@@ -398,7 +542,11 @@ impl Format {
 
     #[cfg(test)]
     fn progress_status(&self, cur: usize, max: usize, msg: &str) -> Option<String> {
-        let mut ret = self.progress(cur, max)?;
+        let mut ret = match self.progress(cur, max)? {
+            // Check only the variant that contains text.
+            ProgressOutput::TextAndReport(text, _) => text,
+            _ => return None,
+        };
         self.render(&mut ret, msg);
         Some(ret)
     }
@@ -420,6 +568,7 @@ fn test_progress_status() {
         style: ProgressStyle::Ratio,
         max_print: 40,
         max_width: 60,
+        term_integration: TerminalIntegration::new(false),
     };
     assert_eq!(
         format.progress_status(0, 4, ""),
@@ -493,6 +642,7 @@ fn test_progress_status_percentage() {
         style: ProgressStyle::Percentage,
         max_print: 40,
         max_width: 60,
+        term_integration: TerminalIntegration::new(false),
     };
     assert_eq!(
         format.progress_status(0, 77, ""),
@@ -518,6 +668,7 @@ fn test_progress_status_too_short() {
         style: ProgressStyle::Percentage,
         max_print: 25,
         max_width: 25,
+        term_integration: TerminalIntegration::new(false),
     };
     assert_eq!(
         format.progress_status(1, 1, ""),
@@ -528,6 +679,25 @@ fn test_progress_status_too_short() {
         style: ProgressStyle::Percentage,
         max_print: 24,
         max_width: 24,
+        term_integration: TerminalIntegration::new(false),
     };
     assert_eq!(format.progress_status(1, 1, ""), None);
+}
+
+#[test]
+fn test_term_integration_disabled() {
+    let report = TerminalIntegration::new(false);
+    let mut out = String::new();
+    out.push_str(&report.remove().to_string());
+    out.push_str(&report.value(10.0).to_string());
+    out.push_str(&report.indeterminate().to_string());
+    assert!(out.is_empty());
+}
+
+#[test]
+fn test_term_integration_error_state() {
+    let mut report = TerminalIntegration::new(true);
+    assert_eq!(report.value(10.0), StatusValue::Value(10.0));
+    report.error();
+    assert_eq!(report.value(50.0), StatusValue::Error(50.0));
 }
--- a/src/doc/src/reference/config.md
+++ b/src/doc/src/reference/config.md
@@ -191,13 +191,14 @@ metadata_key1 = "value"
 metadata_key2 = "value"
 
 [term]
-quiet = false          # whether cargo output is quiet
-verbose = false        # whether cargo provides verbose output
-color = 'auto'         # whether cargo colorizes output
-hyperlinks = true      # whether cargo inserts links into output
-unicode = true         # whether cargo can render output using non-ASCII unicode characters
-progress.when = 'auto' # whether cargo shows progress bar
-progress.width = 80    # width of progress bar
+quiet = false                    # whether cargo output is quiet
+verbose = false                  # whether cargo provides verbose output
+color = 'auto'                   # whether cargo colorizes output
+hyperlinks = true                # whether cargo inserts links into output
+unicode = true                   # whether cargo can render output using non-ASCII unicode characters
+progress.when = 'auto'           # whether cargo shows progress bar
+progress.width = 80              # width of progress bar
+progress.term-integration = true # whether cargo reports progress to terminal emulator
 ```
 
 ## Environment variables
@@ -1360,6 +1361,13 @@ Controls whether or not progress bar is shown in the terminal. Possible values:
 * Environment: `CARGO_TERM_PROGRESS_WIDTH`
 
 Sets the width for progress bar.
+
+#### `term.progress.term-integration`
+* Type: bool
+* Default: auto-detect
+* Environment: `CARGO_TERM_PROGRESS_TERM_INTEGRATION`
+
+Report progress to the teminal emulator for display in places like the task bar.
 
 [`cargo bench`]: ../commands/cargo-bench.md
 [`cargo login`]: ../commands/cargo-login.md
--- a/src/cargo/core/compiler/job_queue/mod.rs
+++ b/src/cargo/core/compiler/job_queue/mod.rs
@@ -854,7 +854,7 @@ impl<'gctx> DrainState<'gctx> {
     }
 
     fn handle_error(
-        &self,
+        &mut self,
         shell: &mut Shell,
         err_state: &mut ErrorsDuringDrain,
         new_err: impl Into<ErrorToHandle>,
@@ -863,6 +863,7 @@ impl<'gctx> DrainState<'gctx> {
         if new_err.print_always || err_state.count == 0 {
             crate::display_error(&new_err.error, shell);
             if err_state.count == 0 && !self.active.is_empty() {
+                self.progress.indicate_error();
                 let _ = shell.warn("build failed, waiting for other jobs to finish...");
             }
             err_state.count += 1;


src/doc/src/reference/config.md
```diff
@@ -1331,6 +1331,13 @@ Controls whether or not progress bar is shown in the terminal. Possible values:
 
 Sets the width for progress bar.
 
+#### `term.progress.taskbar`
```
epage, line 1334: The [example config](https://doc.rust-lang.org/cargo/reference/config.html#configuration-format) would also need to be updated

src/doc/src/reference/config.md
```diff
@@ -1331,6 +1331,13 @@ Controls whether or not progress bar is shown in the terminal. Possible values:
...
+* Type: bool
+* Default: true on Windows, false on everything else
```
epage, line 1336: Like with hyperlinks and unicode, we could say "auto-detect"

src/doc/src/reference/config.md
```diff
@@ -1331,6 +1331,13 @@ Controls whether or not progress bar is shown in the terminal. Possible values:
...
+* Environment: `CARGO_TERM_PROGRESS_TASKBAR`
+
+Enables output of progress to a taskbar via ANSI escape codes.
```
epage, line 1339: What do you think of
```suggestion
Report progess to the teminal emulator for display in places like the task bar
```

src/cargo/util/context/mod.rs
epage, line 1: Could you clean up the commits to represent how you want this reviewed and merged?

src/cargo/util/progress.rs
```diff
@@ -76,6 +77,76 @@ struct Format {
...
+
+impl TaskbarProgress {
+    /// Creates a new `TaskbarProgress` from a cargo's config system.
+    ///
+    /// * `config == None` enables taskbar progress reporting on Windows
+    /// by default and disables on other platforms
+    fn from_config(config: Option<bool>) -> Self {
+        let show = match config {
+            Some(v) => v,
+            None => cfg!(target_os = "windows"),
```
epage, line 97: I don't think `target_os` is sufficient.  We should be checking the terminal as well.

src/cargo/util/progress.rs
```diff
@@ -76,6 +77,76 @@ struct Format {
...
+}
+
+impl std::fmt::Display for TaskbarValue {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // From https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC
+        // ESC ] 9 ; 4 ; st ; pr ST
+        // When st is 0: remove progress.
+        // When st is 1: set progress value to pr (number, 0-100).
+        // When st is 2: set error state in taskbar, pr is optional.
+        // When st is 3: set indeterminate state.
```
epage, line 139: Should we respect `ProgressStyle::Indeterminate` and have a way of outputting this?

src/cargo/util/progress.rs
```diff
@@ -280,6 +353,7 @@ impl<'gctx> State<'gctx> {
         // return back to the beginning of the line for the next print.
         self.try_update_max_width();
         if let Some(pbar) = self.format.progress(cur, max) {
+            self.taskbar.percent = ((cur as f32) / (max as f32) * 100.0) as u8;
```
epage, line 356: `Format` has a `progress` function to abstract things like this.  I feel like `self.taskbar` should as well.

src/cargo/util/progress.rs
```diff
@@ -76,6 +77,76 @@ struct Format {
...
+    Value(u8),
+}
+
+impl std::fmt::Display for TaskbarValue {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // From https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC
+        // ESC ] 9 ; 4 ; st ; pr ST
+        // When st is 0: remove progress.
+        // When st is 1: set progress value to pr (number, 0-100).
+        // When st is 2: set error state in taskbar, pr is optional.
```
epage, line 138: Only slightly related to this PR: Any details on the error state?  Wondering how we could use that.

src/cargo/util/progress.rs
weihanglo: This is a blocker if we want to turn it on by default for certain platforms. Otherwise, we may figure it out later if it is an opt-in feature.

src/cargo/util/progress.rs
Gordon01: Good catch! We should, and it works beatifully:

https://github.com/user-attachments/assets/70ec0a05-3565-412c-be2e-fcf525522f99

src/cargo/util/progress.rs
Gordon01: I've reworked code a little: moved `TaskbarProgress` into `Format`. That also solved the problem with this line. No code duplication now, I've reused a value calculated and normalized in `Format::progress()`.

Now `Format::progress()` also returns not just a string, but a `ProgressOutput` enum with 3 variants: `PrintNow` (used in `Progress::print_now()`), `TextAndTaskbar` and `Taskbar` (for cases when we don't have space or terminal width is low).

I believe this makes intents clear, otherwise code in `Context::print()` becames too complicated because of all those corner cases when we want to output taskbar progress but not text.

src/cargo/util/progress.rs
Gordon01: We can but since cargo exits on error it would be visible for a fraction of second. Probably more in a case when sources contain error and it takes time for other codegen units to complete. Do you think it is worth it?

This also raises another small issue: if user stops cargo with ctrl+c, taskbar progress is not cleared and it stays until the next invocation of a program with taskbar progress report capability. Microsoft's own winget behaves like this.
I've tried to set ctrl+c handler in cargo and clear taskbar from there. It works, but it 50-100 LOC which would be used rarely. What do you think?

Gordon01:
Hello, Ed, thank you for the review and sorry for the late feedback. It's been a lot of work lately.

src/cargo/util/context/mod.rs
epage: Huh, looks like squash isn't enabled for this repo.

Besides, the question of how you want this reviewed and merged is best determined by you than us.

src/cargo/util/progress.rs
```diff
@@ -74,6 +74,94 @@ struct Format {
...
+    }
+}
+
+impl std::fmt::Display for TaskbarValue {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // From https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC
+        // ESC ] 9 ; 4 ; st ; pr ST
+        // When st is 0: remove progress.
+        // When st is 1: set progress value to pr (number, 0-100).
+        // When st is 2: set error state in taskbar, pr is optional.
```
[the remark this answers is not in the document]
epage: I am hesitant to have a ctrl+c handler but I also have had enough bad experience on Windows with colors being leaked from a program and could see myself being annoyed about progress state leaking as well.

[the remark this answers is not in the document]
Gordon01: Here's my prvious work on ctrl+c handler: https://github.com/rust-lang/cargo/commit/de22a7f4f559f129fb909d67ef073687fc73fb94

The approach works, but then we need to set the handler depending on Progress.state. Can you please advice on this? Do you want to have a handler in `src/bin/cargo/main.rs` ?

src/cargo/util/progress.rs
Gordon01: Thank you for this link! For some reason I was thinking that this is much more complicated than reading env.
I've used a proper `ConEmuANSI` variable.
I haven't added OS check because we may run cargo in WSL and it "just works" now. 

Who-ho! This now works better then I expected.

src/cargo/util/progress.rs
Gordon01: I will experiment on this in the next few days and will get back with feedback.

Gordon01: https://github.com/user-attachments/assets/b1f0264b-263f-425f-b07d-254f0ac76e6f

src/cargo/util/progress.rs
```diff
@@ -74,6 +74,113 @@ struct Format {
...
+    /// Do not output anything
+    None,
+    /// Remove progress
+    Remove,
+    /// Progress value 0-100
+    Value(f64),
+    /// Indeterminate state (no bar, just animation)
+    Indeterminate,
+    Error,
+}
```
epage, line 100: This is treating error and value is mutually exclusive.  Is that a limitation of the escape codes?

src/cargo/util/progress.rs
epage: Neat!

src/cargo/util/progress.rs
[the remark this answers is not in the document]
epage: @ehuss I'd be interested in your opinion on this.

epage:
This is looking really great and making me jealous that my terminal doesn't support it!

src/cargo/util/progress.rs
Gordon01: No there is no limitation, again your remark was hot to the point!

https://github.com/user-attachments/assets/d07325a5-bc21-4259-b90b-4dff2c80d535

src/cargo/util/progress.rs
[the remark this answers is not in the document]
Gordon01: Since we're now detecting a terminal regardless of the OS, a ctrl+c handler needs to be portable too, not like in my first implementation. If we still want to go this way, a few questions:
1. Can we use external crate like https://crates.io/crates/ctrlc? I assume your policy is to use as few external crates as possible
2. Will it iterfere with https://github.com/rust-lang/cargo/blob/master/crates/cargo-util/src/process_builder.rs#L259-L276? I don't fully understand when this code is used.

src/cargo/util/progress.rs
```diff
@@ -74,6 +74,113 @@ struct Format {
...
+        // When st is 3: set indeterminate state, pr is ignored.
+        // When st is 4: set paused state, pr is optional.
+        let (state, progress) = match self {
+            Self::None => return Ok(()),
+            Self::Remove => (0, 0.0),
+            Self::Value(v) => (1, *v),
+            Self::Indeterminate => (3, 0.0),
+            Self::Error(v) => (2, *v),
+        };
+        write!(f, "\x1b]9;4;{state};{progress:.0}\x1b\\")
```
Gordon01, line 182: According to [1](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html), [wiki](https://en.wikipedia.org/wiki/ANSI_escape_code) and [MS Learn](https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences) `<ST>` is `ESC \ (0x1B 0x5C)` or `BEL (0x7)`. Both are supported but `ESC \` is preferred

Gordon01:
Hello, @epage and everyone interested. Hope, you're well!

I have a good news! The progress in standartisation of the sequence is going faster than I initially thought. Soon it's gonna be supported by systemd and GNOME vte. Here is the [message](https://gitlab.gnome.org/GNOME/vte/-/issues/2845#note_2289317) from [Christian Persch](https://gitlab.gnome.org/chpe):
> So my conclusion after reading through this issue, the windows terminal issue/pr, and conemu implementation, is that while this API has flaws and might be improved, it's reasonably widely implemented, and small enough (and the flaws small enough) so as to make it not worthwhile to try to invent yet another protocol for this.
Therefore I'm going to commit a patch to support `OSC 9 ; 4` to vte (plus hook it up in the test app). I've changed it a bit since the last wip/progress branch; please check and test if this still meets the requirements. The parser at least should fully copy conemu behaviour.

And, cosequently [support to Ptyxis was added](https://blogs.gnome.org/chergert/2024/12/03/ptyxis-progress-support/).

From [Lennart Poettering](https://mastodon.social/@pid_eins/113406672373007116):
>Since a while systemd has been showing nice terminal progress bars when doing certain slow operations (for example, when systemd-repart initializes a disk). With v257 we go one step further with this. Whenever we show the progress bar we'll now also issue certain terminal ANSI sequences that tell your terminal emulator that a slow operation is going on and what the progress currently is.

And the bug in kitty about conflict with their OSC 9 implementation was fixed [here](https://github.com/kovidgoyal/kitty/issues/8011).

So, to sum it up, Ed, it looks like you will get the support in your terminal earlier than you anticipated.

weihanglo:
Thank you for driving this!

I use Kitty so may gain nothing, but still a great feature!

src/cargo/util/progress.rs
Gordon01: Seems like it was a right choise, see [a recent patch comments](https://github.com/systemd/systemd/pull/34964/files#diff-59ed3cf1ea99efb36edcb76142b4da2ffe9588f4a9a3483408c03fc704672b00R30-R35) for systemd.

src/cargo/util/progress.rs
```diff
@@ -74,6 +74,113 @@ struct Format {
...
+        }
+    }
+
+    /// Creates a new `TaskbarProgress` from a cargo's config.
+    /// If not explicitly enabled or disabled, detect a supported terminal.
+    fn from_config(gctx: &GlobalContext) -> Self {
+        let enabled = match gctx.progress_config().taskbar {
+            Some(v) => v,
+            None => {
+                gctx.get_env("WT_SESSION").is_ok()
```
weihanglo, line 126: Could we have a comment for the source of how each env var works for auto-detection?

src/cargo/util/progress.rs
weihanglo, line 1: If we feel too risky (like excessive notification on Kitty), should we roll this out only on nightly for a while?

Gordon01: I agree. It will take time for linux users to upgrade their terminals. And who knows what other bugs systemd will catch. I definitely want people to associate Rust ecosystem with quality.

src/cargo/util/progress.rs
Gordon01: Done. Also replaced the match block with `unwrap_or_else()`

src/cargo/util/progress.rs
```diff
@@ -74,6 +74,112 @@ struct Format {
...
+
+    /// Create a `TaskbarProgress` from Cargo's configuration.
+    /// Autodetects support if not explicitly enabled or disabled.
+    fn from_config(gctx: &GlobalContext) -> Self {
+        let enabled = gctx.progress_config().taskbar.unwrap_or_else(|| {
+            // Windows Terminal session.
+            gctx.get_env("WT_SESSION").is_ok()
+            // Compatibility with ConEmu's ANSI support.
+                || gctx.get_env("ConEmuANSI").ok() == Some("ON".into())
+        });
```
epage, line 129: Can we break the detection out into a function like we do for hyperlinks?

(and this would be allowed to use `std::env::var_os`, like hyperlinks)

src/cargo/util/progress.rs
epage: Unsure if this is needed.  This feature is "auto" by default and the only places where it auto-enables are inside of terminals that are known to support this feature.

So the only way this would be a problem is if the user does `CARGO_TERM_PROGRESS_TASKBAR = true cargo ...` and that is by design so people can experiment with this feature and report back on support so we can expand it.

When it comes to a terminal having bad behavior in an old version and good behavior in a new, we can sometimes do version detection, e.g. https://github.com/zkat/supports-hyperlinks/blob/a2cc083668eb47c046f32ae0bc89d5bbdb8398ef/src/lib.rs#L23-L28

src/cargo/util/progress.rs
Gordon01: I'll see what I can do.

> Ideally I see us breaking this out into a crate as some point

After Lennart's push I think that it may be beneficial for the community to have a crate exactly for this one purpose: detecting supported terminal and serializing progress value and status into escape sequences.

Gordon01: I've added `cargo::core::shell::supports_progress_report() -> bool`, `cargo::core::shell::Stream::progress_report` and `cargo::core::shell::Shell::progress_report() -> bool`. Hope this wording is clear:

```rust
let enabled = gctx
    .progress_config()
    .taskbar
    .unwrap_or_else(|| gctx.shell().progress_report());
```

epage:
@rfcbot fcp merge

In terms of stability, this adds a new config field
```toml
[term]
progress.taskbar = true # whether cargo reports progress to terminal emulator
```

Like with `hyperlinks` and `unicode`, it is auto-detect by default and can be explicitly turned on/off.

For rendering, this is supported by at least
- ConEmu
- Windows Terminal
- GNOME vte / Ptyxis

For emitting, this is supported by at least
- systemd 

In the current implementation, f the user hits `ctrl-c`, then nothing will remove the taskbar status.  We could add a ctrl-c handler but I'm always hesitant about those and would prefer to see if we can get away without it.  It also sounds like other existing programs don't add a ctrl-c handler for this.

src/cargo/core/shell.rs
```diff
@@ -565,6 +576,14 @@ fn supports_unicode(stream: &dyn IsTerminal) -> bool {
     !stream.is_terminal() || supports_unicode::supports_unicode()
 }
 
+#[allow(clippy::disallowed_methods)] // ALLOWED: to read terminal app signature
+fn supports_progress_report() -> bool {
```
epage, line 580: This should be dependent on the state of the stream.

If stderr is being piped to a file, we shouldn't emit these ANSI escape codes (unless the user overrides t)

nagisa: Out of curiosity, who is going to maintain this detection logic as the terminal support for this feature increases?

epage: For now, us. As I said earlier in this PR, I expect this will get split out into a crate.

src/doc/src/reference/config.md
```diff
@@ -1361,6 +1362,13 @@ Controls whether or not progress bar is shown in the terminal. Possible values:
 
 Sets the width for progress bar.
 
+#### `term.progress.taskbar`
+* Type: bool
+* Default: auto-detect
+* Environment: `CARGO_TERM_PROGRESS_TASKBAR`
```
ehuss, line 1368: Just a mild note, this environment variable doesn't work by itself (because it requires the other `progress` settings).

src/doc/src/reference/config.md
```diff
@@ -191,13 +191,14 @@ metadata_key1 = "value"
...
-progress.when = 'auto' # whether cargo shows progress bar
-progress.width = 80    # width of progress bar
+quiet = false           # whether cargo output is quiet
+verbose = false         # whether cargo provides verbose output
+color = 'auto'          # whether cargo colorizes output
+hyperlinks = true       # whether cargo inserts links into output
+unicode = true          # whether cargo can render output using non-ASCII unicode characters
+progress.when = 'auto'  # whether cargo shows progress bar
+progress.width = 80     # width of progress bar
+progress.taskbar = true # whether cargo reports progress to terminal emulator
```
joshtriplett, line 201: Different terminal emulators handle this differently; for instance, some of the Linux terminal emulators show this in the terminal's tab bar, instead. Rather than having a name that's specific to the rendering of a subset of terminals, could we use a name that's a little more generic?

Perhaps `progress.ansi` or `progress.osc`? Or, if we want something less technical, then perhaps `progress.escapes`, indicating that it sends escape codes?

src/doc/src/reference/config.md
ehuss: BTW, you might be able to fix this by adding `#[serde(default)]` to the `ProgressConfig::when` field.

src/cargo/core/shell.rs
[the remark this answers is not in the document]
Gordon01: From @joshtriplett:
> Different terminal emulators handle this differently; for instance, some of the Linux terminal emulators show this in the terminal's tab bar, instead. Rather than having a name that's specific to the rendering of a subset of terminals, could we use a name that's a little more generic?

Windows Terminal also displays progress in the tab. And here I also started to think about more generic term, because it's really more like a progress and status messaging to terminal. 

I'm fine to change it back to `taskbar_progress` because naming is hard anyway.

src/doc/src/reference/config.md
Gordon01: Here are my points:
- `progress.taskbar` - I like this one for easy interchangeability with `taskbar_progress`-named fields and functions
- `progress.terminal` - clearly indicates a connection with terminal emulators
- `progress.osc` - people name PRs and issues mostly with something like "OSC 9;4" so it's somewhat familiar
- `progress.state` - because report is a percentage and state

src/doc/src/reference/config.md
Gordon01: Great point! Definitely more ergonomic now.

image
image

src/cargo/core/shell.rs
Gordon01: > If stderr is being piped to a file, we shouldn't emit these ANSI escape codes (unless the user overrides t)

Makes sense. I've modified the condition to `*progress_report && *stderr_tty`

src/doc/src/reference/config.md
Gordon01: This breaks test `bad_progress_config_missing_when` 

I don't understand whe `ProgressWhen` has a `Default` implementation then...

src/doc/src/reference/config.md
Gordon01: https://github.com/rust-lang/cargo/pull/8165

src/cargo/core/shell.rs
```diff
@@ -286,6 +287,17 @@ impl Shell {
...
     }
 
+    pub fn supports_osc9_4(&self) -> bool {
+        match &self.output {
+            ShellOut::Write(_) => false,
+            ShellOut::Stream {
+                progress_report,
+                stderr_tty,
+                ..
+            } => *progress_report && *stderr_tty,
```
epage, line 297: Generally this check is done in the `supports` functions, should we make that consistent?

src/doc/src/reference/config.md
epage: For me, my care abouts are
- Easy discover for people wanting to change this behavior
- Clear meaning when looking at a config or config docs
- Consistency with wider community

I feel like `ansi`, `osc`, `escapes`,  and `osc9_4 do neither unless you are someone already in the nitty gritty implementation details.  Also, our other OSC configs do not speak this way (`color`, `unicode`, `hyperlinks, though `color can also support wincon APIs).

src/doc/src/reference/config.md
ehuss: I don't recall any particular decisions made regarding `bad_progress_config_missing_when`. I personally would not object to changing that to allow it to pass. I can't think of a reason it would be required to specify `when` with `width`.

src/doc/src/reference/config.md
joshtriplett: > progress.taskbar - I like this one for easy interchangeability with taskbar_progress-named fields and functions

That's circular. :)

src/doc/src/reference/config.md
```diff
@@ -1361,6 +1362,13 @@ Controls whether or not progress bar is shown in the terminal. Possible values:
 
 Sets the width for progress bar.
 
+#### `term.progress.taskbar`
+* Type: bool
+* Default: auto-detect
+* Environment: `CARGO_TERM_PROGRESS_TASKBAR`
+
+Report progess to the teminal emulator for display in places like the task bar.
```
weihanglo, line 1370:
```suggestion
Report progress to the teminal emulator for display in places like the task bar.
```

src/doc/src/reference/config.md
joshtriplett: @epage :+1: to those premises. In addition to those, I'd like to avoid ending up with a name that assumes specifics of how this gets displayed. (This is somewhat similar to https://www.w3.org/QA/Tips/noClickHere .)

Some more possibilities:

- `term.progress.notification` (suggested by @Gordon01)
- `term.progress.terminal` or `term.progress.term` (slightly odd to repeat but they do have distinct meanings)
- `term.progress.terminal-ui`
- `term.progress.ui`
- `term.progress.gui`
- `term.progress.terminal-reporting`

Happy to see any other possibility that doesn't specifically assume a UI display mechanism (e.g. not "title" or "taskbar" or "tab-title" or similar).

weihanglo: I don't think there is a clear name to explain itself. `notification` sounds too far away. `osc9_4` is  too tied to implementation details. Maybe we'll have other reporting mechanism for some specific terminal emulator in the future (I hope not though).

Since what the feature does is reporting the progress. So, perhaps a verb `progress.report` is a bit clearer?

(Hey report is also a noun 😅)

joshtriplett:
@rfcbot reviewed
@rfcbot concern config-name

*Temporary* concern until https://github.com/rust-lang/cargo/pull/14615#discussion_r1888730634 gets resolved, because folks may be about to go on holiday soon and I don't want this to get lost. I am *not* at this time attempting to actually block the FCP if this isn't changed, and I'm happy to defer to consensus; I just want to make sure that this doesn't accidentally FCP without resolving it one way or another.

src/doc/src/reference/config.md
epage: `notification` can also sound like the kitty feature that slightly overlaps om the codes.

Another name I thought of was `progress.system` to convey that its reporting through the system that calls into cargo.

Eventually I want to add support for multi-line progress which might need its own config and then if we ever get a watch, I would like it to look like `bacon`.  `ui` and `gui` might get confusing in those respects (well, more `tui`).

Gordon01:
Happy New Year and Merry Christmas!

I had a lot of work in the last weeks of December and now I'm on New Year vacation. I'll be back in the second half of January, I'll fix the tests, decide on naming. 

Wishing everyone a happy new 2025!

src/doc/src/reference/config.md
weihanglo: Just looked at this again in 2025, notification sounds fine as well :)

wez:
Just wanted to chime in here to say that WezTerm just now added support for these escapes in `main`.
It should be fine to emit the escapes in earlier versions, but in case you prefer to probe to be sure, you can look for the following environment variables:

```
TERM_PROGRAM=WezTerm
TERM_PROGRAM_VERSION=20250209-182623-44866cc1
```

`$TERM_PROGRAM_VERSION` will be `20250209-182623-44866cc1` or greater in wezterms that support this sequence.

https://github.com/wezterm/wezterm/issues/6581 is the wezterm issue tracking this feature.

src/cargo/core/shell.rs
Gordon01: Looking on the other code it seems that it's a separate thing:

- `supports_progress_report()` detects the supported terminal from environment variables
- `progress_report_available()` signals that terminal supports the feature and it's a tty

epage: I'm not sure I'm following
- `supports_unicode` checks `!stream.is_terminal() || supports_unicode::supports_unicode()`
- `AutoStream::new(std::io::stdout(), stdout_choice)` will check `is_terminal` as well as what the terminal supports

hyperlinks is a bit odd but that is later combined with the `AutoStream` check

Gordon01:
Hi everyone!

I've changed the feature name to progress report. This name keeps the focus on what the feature does (reporting progress) rather than how it’s rendered.

@wez also called is like this.

I can't make test to work. It passes locally on Windows and Linux but here on pipeline it behaves differently and I can't figure out why.

src/cargo/core/shell.rs
```diff
@@ -565,6 +579,21 @@ fn supports_unicode(stream: &dyn IsTerminal) -> bool {
...
+    // Windows Terminal session
+    std::env::var("WT_SESSION").is_ok()
+    // Compatibility with ConEmu's ANSI support
+        || std::env::var("ConEmuANSI").ok() == Some("ON".into())
+    // WezTerm: check if TERM_PROGRAM is "WezTerm" and the version is recent enough
+    // https://github.com/rust-lang/cargo/pull/14615#issuecomment-2646738819
+    || (std::env::var("TERM_PROGRAM").ok() == Some("WezTerm".into())
+        && std::env::var("TERM_PROGRAM_VERSION")
+            .ok()
+            .map_or(false, |v| v.as_str() >= "20250209-182623-44866cc1"))
```
epage, line 594: I think the main reason for version checks is to deal with buggy terminal implementations.  If older wezterms gracefully handle these codes, should we skip the version check?

epage, line 594: @wez are versions string comparable or do they need to be split apart?

wez: FWIW, older versions of wezterm will swallow an OSC they doesn't understand, so you could remove the version number check and just check for wezterm without it.  That feels like less magic to keep track of over here.

wez: yeah, they are comparable lexicographically like this. It's YYYYMMDD-HHMMSS of the commit (UTC) then the commit hash which is informational; the hash isn't really comparable this way, but the date+time check will be sufficient except in pathological cases that I try to keep out of the commit history.

src/doc/src/reference/config.md
epage: Within a terminal emulator, I think that name can make sense (though that is one specific terminals name and not a common name)

Within the context of the term table
```toml
[term]
quiet = false          # whether cargo output is quiet
verbose = false        # whether cargo provides verbose output
color = 'auto'         # whether cargo colorizes output
hyperlinks = true      # whether cargo inserts links into output
unicode = true         # whether cargo can render output using non-ASCII unicode characters
progress.when = 'auto' # whether cargo shows progress bar
progress.width = 80    # width of progress bar
progress.report = true # whether cargo reports progress to terminal emulator
```
especially if you ignore the comments, `report` sounds like `when`

Gordon01: Should I roll it back to `term.progress.taskbar` or `term.progress.notification`, `term.progress.integration`

Integration sounds great for me

Gordon01: From https://github.com/rust-lang/cargo/pull/14615#discussion_r1889484062

> I don't think there is a clear name to explain itself. notification sounds too far away. osc9_4 is too tied to implementation details. Maybe we'll have other reporting mechanism for some specific terminal emulator in the future (I hope not though).
> 
> Since what the feature does is reporting the progress. So, perhaps a verb progress.report is a bit clearer?
> 
> (Hey report is also a noun 😅)

src/cargo/core/shell.rs
Gordon01: Makes sense. Initially, I thought it's a good idea to add field to ShellOut for caching the heuristic result. But in the end it's more code with unclear goal so I followed KISS principle.

Reworked everything into a single `Shell::is_progress_report_available()`

src/doc/src/reference/config.md
weihanglo: Thank you for trying to find a good name!

The team plans to discuss this tomorrow in the meeting. Hope we'll have a final decision after that.

src/cargo/core/shell.rs
Gordon01: I've removed the version check for WezTerm in the latest edit.

src/cargo/core/shell.rs
epage: This is still not quite following the existing patterns.  I was expecting to see a separate function like `supports_unicode`

Gordon01: Should we have a field `progress_report` in `ShellOut`? Is it mainly for caching?

If yes, what to do in `ProgressReport::from_config()` then? It needs to read config and fall back to `Shell`'s getter to determine terminal support.

```
fn from_config(gctx: &GlobalContext) -> Self {
        let enabled = gctx
            .progress_config()
            .report
            .unwrap_or_else(|| gctx.shell().is_progress_report_available());
```

previously I've made something like:

```
pub fn is_progress_report_available(&self) -> bool {
    if let ShellOut::Stream { stderr_tty, progress_report,  .. } = self.output {
        progress_report && stderr_tty
    } else {
        false
    }
}
```

epage: I'm fine with us caching a `stdout_progress_report` and `stderr_progress_report`, like we do for `std*_unicode` but that then calls a non-member `supports_unicode`.

src/doc/src/reference/config.md
Gordon01: I've fixed the test in the latest edit. Completely forgot that in the test environment `auto` is `none`. Therefore, just changed the test to check for successful completion because now a single `term.progress.width = 100` is correct but the progress itself is not shown because `'auto' == 'none'`.

src/cargo/util/progress.rs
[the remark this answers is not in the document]
Gordon01: Should I do anything else here?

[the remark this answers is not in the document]
epage: I would lean away from a ctrl-c handler.

Should we have `cargo run` and `cargo <plugin>` clear the progress state before `exec`ing the new process?

src/cargo/core/shell.rs
Gordon01: Ah, looks like I finally got it. This is non-member because in the future it would be moved to a separate crate, right?

```rust
fn supports_progress_report(stream: &dyn IsTerminal) -> bool {
    let windows_terminal = std::env::var("WT_SESSION").is_ok();
    let conemu = std::env::var("ConEmuANSI").ok() == Some("ON".into());
    let wezterm = std::env::var("TERM_PROGRAM").ok() == Some("WezTerm".into());

    (windows_terminal || conemu || wezterm ) && stream.is_terminal()
}
```

Why do we need `stdout_progress_report`? `util::progress::Progress` only outputs to stderr.

```rust
impl Shell {
    pub fn new() -> Shell {
        Shell {
            output: ShellOut::Stream {
                stderr_progress_report: supports_progress_report(&std::io::stderr()),
```

And finally, `Shell::is_progress_report_available()`

```rust
pub fn is_progress_report_available(&self) -> bool {
    if let ShellOut::Stream { stderr_progress_report,  .. } = self.output {
        stderr_progress_report
    } else {
        false
    }
}
```

Is this correct?)

joshtriplett:
@rfcbot resolved config-name

epage:
@rfcbot concern config-name

I don't feel the name is resolved yet :)

(and the PR should make sure to update all references to the final name)

epage:
Hmm, either a delay or it doesn't like me reusing the name

@rfcbot concern config-name-part-2

src/doc/src/reference/config.md
epage: We talked about this in today's cargo team meeting and landed on `term.progress.term-integration`.  When renaming, be sure to update relevant functions to use this name.

When discussing this, we were working off of this summary of the discussion so far:
- Care abouts
  - Easy discover for people wanting to change this behavior
  - Clear meaning when looking at a config or config docs
  - Consistency with wider community
    - [ConEmu](https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC): "Set progress state on Windows 7 taskbar and ConEmu title"
    - [Windows Terminal](https://github.com/microsoft/terminal/pull/8055): "Taskbar", "Taskbar Progress"
    - [systemd](https://github.com/systemd/systemd/pull/34929): "Windows Terminal progress indication"
    - [ptyxis blog post](https://blogs.gnome.org/chergert/2024/12/03/ptyxis-progress-support/): "progress state"
    - [VTE](https://gitlab.gnome.org/GNOME/vte/-/issues/2845): "progress indication" ("progress hint" describes how to interpret "progress value")
- Names
  - `term.progress.taskbar`: original intent but terminal/OS specific
  - `term.progress.<term|terminal>` may seem redundant, may be confused with content inside terminal
  - `term.progress.notification`: another escape code exists for toaster pop ups which this sounds more like (intrusive), rather than a passive indicator
  - `term.progress.<ansi|osc|escapes|osc9_4>` are fairly low level; hard to discover or understand from reading the config unless you know nitty gritty terminal details
  - `term.progress.<terminal-ui|ui|gui|system>`
  - `term.progress.report` may be confused with `when`
  - `term.progress.integration`

From our discussion:

The use cases for caring about the config are:
- Your terminal added support but cargo hasn't whitelisted it yet and so you want to turn it on
  - You will be operating from the context of how your terminal frames this
- You see your terminal doing something when you run cargo that you don't like and want to turn it off.
  - You will be operating from only what you can see and trying to find the relevant terminal or cargo feature to turn it off which can make it hard to intuit the words

Without a standardized way of framing this, users are likely going to be lost anyways.  Any incremental benefit of one name or the other was likely going to be lost.  However, a bad name could cause problems.

tests/testsuite/progress.rs
```diff
@@ -53,28 +53,19 @@ fn bad_progress_config_missing_width() {
 }
 
 #[cargo_test]
-fn bad_progress_config_missing_when() {
+fn default_progress_is_auto() {
```
epage, line 56: Can you split the behavior change behind this test change into its own commit?  This change in behavior should be more explicitly acknowledged rather than conjoined in the same commit with the new feature that wants to leverage it

src/cargo/util/progress.rs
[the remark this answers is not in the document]
Gordon01: `cargo {run, clippy, hack}` works correct out of the box 

https://github.com/user-attachments/assets/4d8df720-303b-4154-a67a-8f94ab5a7396

https://github.com/user-attachments/assets/70e6965f-ce7f-4b66-b539-796219027efd

src/doc/src/reference/config.md
Gordon01: Thank you and a team for this resolution!
I've renamed everything to `term_integration` and a control struct to `TerminalIntegration`.

src/cargo/core/shell.rs
Gordon01: - Re-added a `stderr_term_integration` field to `ShellOut`
- `Shell::is_err_term_integration_available()`
- freestanding `supports_term_integration()`

Should be fine now

tests/testsuite/progress.rs
Gordon01: Done: https://github.com/rust-lang/cargo/pull/15287

kovidgoyal:
> Thank you for driving this!
> 
> I use Kitty so may gain nothing, but still a great feature!

kitty displays OSC 9 progress percentage in tab titles as of version 0.39, see https://sw.kovidgoyal.net/kitty/conf/#opt-kitty.tab_title_template

kovidgoyal:
> > > Thank you for driving this!
> > > I use Kitty so may gain nothing, but still a great feature!
> > 
> > 
> > kitty displays OSC 9 progress percentage in tab titles as of version 0.39, see https://sw.kovidgoyal.net/kitty/conf/#opt-kitty.tab_title_template
> 
> Can you please advice on how is the best to detect kitty?

Depends on how robust you want to be. The simplest is checking the TERM env var, the most robust is using the XTVERSION escape code combined with a DA1 escape code to avoid timeouts.

epage:
So `TERM=xterm-kitty` detects kitty.  However, we should use version detection with kitty because of kovidgoyal/kitty#8011

I'm not seeing any way to do version detection for kitty besides `XTVERSION` and, due to my lack of experience with those call/response escape codes, I'm a bit cautious about what all gotchas there might be.

epage:
@rfcbot resolved config-name-part-2

epage (approved):

Gordon01:
Hello, cargo team! Looks like we're not getting issues with this feature!
Thanks a lot for your support and help. I've learned a lot while doing this PR and I'm impressed with your passion with quality software. No wonder that you're making Rust.


@[epage](https://github.com/epage) [on Dec 9, 2024](https://github.com/rust-lang/cargo/pull/14615#discussion_r1876472499)
> Ideally I see us breaking this out into a crate as some point

Now I also would like to have a separate crate for this feature for reusability in other projects. Can I make it as a crate in

Gordon01:
Hello, Сargo team! 
It looks like we're not encountering any issues with this feature.
Thank you so much for your support and help - I learned a lot while working on this PR, and I'm really impressed by your passion for high-quality software. No wonder you're the ones making Rust!


@[epage](https://github.com/epage) [on Dec 9, 2024](https://github.com/rust-lang/cargo/pull/14615#discussion_r1876472499)
> Ideally I see us breaking this out into a crate as some point

I would also love to break this feature out into a separate crate for reusability in other projects.
Can I create a new crate within Cargo’s workspace, following [these contribution guidelines](https://doc.crates.io/contrib/implementation/packages.html)?

epage:
> I would also love to break this feature out into a separate crate for reusability in other projects.
Can I create a new crate within Cargo’s workspace, following [these contribution guidelines](https://doc.crates.io/contrib/implementation/packages.html)?

There is a high bar for crates to be opened by Cargo.  I'd be fine having such a crate in the [rust-cli org](https://github.com/orgs/rust-cli/repositories?type=all).  I've been trying to decide whether it'd be appropriate to have in [anstyle repo](https://github.com/rust-cli/anstyle) or not (along with hyperlink helpers)

Gordon01:
> I've been trying to decide whether it'd be appropriate to have in [anstyle repo](https://github.com/rust-cli/anstyle) or not (along with hyperlink helpers)

anstyle looks great! The existing [`anstyle-query`](https://docs.rs/anstyle-query/latest/anstyle_query/) crate would be perfect for `supports_term_integration()` and `supports_hyperlinks()`.

For `Progress` and `Hyperlink` and their `Display` impls:
   - The main [`anstyle` crate](https://github.com/rust-cli/anstyle/blob/main/crates/anstyle/src/style.rs) (which already handles styles)
   - A new dedicated crate like `anstyle-osc`

Would you like me to open an issue in the anstyle repo?

epage:
> Would you like me to open an issue in the anstyle repo?

An issue would be a good place to start.

`anstyle` has a fairly specific role, so I'm guessing we won't add features to that crate but we could add a new crate for some of this.

PR #14620, opened 2024-09-30
docs: Minor re-grouping of pages

epage:

In figuring out where MSRV documentation should live, I found the location of some items confusing
- Publishing is written in more of a guide-form, rather than reference
- Caching is written in more of a reference form, rather than a guide, and is more advanced
- Dependency references were scattered, making it harder to find


src/doc/src/SUMMARY.md
```diff
@@ -15,33 +15,33 @@
     * [Cargo.toml vs Cargo.lock](guide/cargo-toml-vs-cargo-lock.md)
     * [Tests](guide/tests.md)
     * [Continuous Integration](guide/continuous-integration.md)
+    * [Publishing on crates.io](guide/publishing.md)
```
weihanglo, line 18: Redirections are needed. Maybe use this? https://rust-lang.github.io/mdBook/format/configuration/renderers.html#outputhtmlredirect

weihanglo: > This will generate an HTML page which will automatically redirect to the given location. Note that the source location does not support # anchor redirects.

This is the interesting part of this config. We might need to do this as well?

https://github.com/rust-lang/cargo/blob/2fd7321ee49307c328e0414c85fb4c36a5653759/src/doc/src/reference/specifying-dependencies.md#L631-L648

epage: So instead of using the redirect feature, you are proposing that we place a hand-written file in the old location that includes the script to redirect fragments?  I'm not fully confident in making sure all of these features interact well with each other, search engines, etc.

weihanglo: The mdbook config one is for generating those old pages with a `<meta http-equiv="refresh" content="0; URL=/.path/to/new/location.html">`, which is a reasonable way to getting a better search engine optimization. However it doesn't support hash. The hand-written `<script>` hack doesn't seem to help this situation, either.

Off the top of my head, the only way to support hash redirections is using JavaScript in the old location of the markdown file, and make the markdown file hidden. It is not good, though.

weihanglo: I have another "bad idea": don't rename file names. Just swap the order in SUMMARY.md.

weihanglo: cc @ehuss as you may know how well we need to maintain link validity. Like would tools in rust-lang/rust check this?

weihanglo (approved):
I am fine with the current state of this PR.

src/doc/src/SUMMARY.md
ehuss: Hm, yea, this might actually be a little bit of a problem. I think there are quite a few links to https://doc.rust-lang.org/cargo/reference/publishing.html#github-permissions in particular. 

I actually don't think there is a way to include a redirect file without leaving an entry in SUMMARY which I think is undesirable here.

I might recommend just moving the entries in SUMMARY.md, but don't rename the file. I realize that is not optimal, since the URL is a little confusing that way.

Perhaps in the future we can extend mdbook to support javascript redirects with HTML fragments on removed pages? We can rename the files at that time.

ehuss: Oh, I see what you mean by leaving an html file in its place. I *think* that should work. That seems like a reasonable option, too.

epage: For now, I'm leaving the path in the old location.

src/doc/src/guide/build-cache.md
```diff
@@ -1,108 +1 @@
...
-[def-workspace]:  ../appendix/glossary.md#workspace  '"workspace" (glossary entry)'
-[target]: ../appendix/glossary.md#target '"target" (glossary entry)'
-[environment variable]: ../reference/environment-variables.md
-[incremental output]: ../reference/profiles.md#incremental
-[sccache]: https://github.com/mozilla/sccache
-[profile]: ../reference/profiles.md
-[binary executables]: ../reference/cargo-targets.md#binaries
-[library targets]: ../reference/cargo-targets.md#library
-[example targets]: ../reference/cargo-targets.md#examples
+# Build Cache
```
weihanglo, line 1: Looks like there are some more outdated links.

```
src/doc/src/reference/cargo-targets.md
371:[Build cache]: ../guide/build-cache.md

src/doc/src/reference/config.md
557:info](../guide/build-cache.md#dep-info-files) file paths. This config setting

src/doc/src/reference/workspaces.md
293:[output directory]: ../guide/build-cache.md

src/doc/src/reference/profiles.md
374:[`target` directory]: ../guide/build-cache.md

src/doc/book.toml
12:"/guide/build-cache.html" = "../reference/build-cache.html"

src/doc/man/includes/options-target-triple.md
15:[build cache](../guide/build-cache.html) documentation for more details.
```

And the one in `src/doc/src/reference/config.md` will be an invalid link. 

I think there are a few options

* Just fix the wrong links above, and merge this PR.
* Move the chapter without renaming to `reference/guide`.
* Put a `<script>` redirection for hash anchors in this empty chapter, instead of using `[output.html.redirect]`.

src/doc/src/guide/build-cache.md
```diff
@@ -1,108 +1 @@
...
-process. Their layout is considered internal to Cargo, and is subject to
-change. Some of these directories are:
-
-Directory | Description
-----------|------------
-<code style="white-space: nowrap">target/debug/deps/</code> | Dependencies and other artifacts.
-<code style="white-space: nowrap">target/debug/incremental/</code> | `rustc` [incremental output], a cache used to speed up subsequent builds.
-<code style="white-space: nowrap">target/debug/build/</code> | Output from [build scripts].
-
-## Dep-info files
```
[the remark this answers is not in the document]
weihanglo: Just made a dead link in unstable book and run `./x.py test linkchecker`. It seems not affected.

src/doc/src/guide/build-cache.md
epage: Yeah, the link checker isn't able to cache these. I've fixed them.

As for the anchor, I thought we weren't too concerned about that for this case?

weihanglo: Looked it again. It is an intra-link inside The Cargo Book so should always be valid I believe. And according to [my test](https://github.com/rust-lang/cargo/pull/14620#discussion_r1786690819) against rust-lang/rust it doesn't seem to cause any link check problem.

weihanglo (approved):
`r=me` when you rebuild those man pages.

PR #14630, opened 2024-10-01
fix: Remove implicit feature removal

epage:

Due to problems we ran into with #14016, we're removing implicit features from the 2024 edition to give ourselves more time to design it as we should.

### Additional information

I could have added a new flag for this or made an EditionNext but I decided to remove it in the hopes to avoid any path dependency in solving this the next time.


epage:
> Feel like code changes I have reviewed for this feature are more than this removal.

My hope is that feeling is coming from seeing PRs that tweaked existing behavior and that this also came with the introduction of a linting system.  I looked through the "implemention" PRs on the tracking issue and this looked to be it.

PR #14631, opened 2024-10-01
docs(config): make `--config <PATH>` more prominent

### LINKED ISSUE #14629: Improve `--config=<PATH>` documentation
The current documentation for `--config=<PATH>` is a bit hard to find and non-obvious. It is in the man page under [`--config`](https://doc.rust-lang.org/nightly/cargo/commands/cargo.html#common-options), but I suggest a few other places that should clearly document it:

- https://doc.rust-lang.org/nightly/cargo/reference/config.html
- The `cargo -h` output ([here](https://github.com/rust-lang/cargo/blob/bd31b4a805ca62e5661437d5ce9dda58d5a161e2/src/bin/cargo/cli.rs#L671))

weihanglo:

Make `--config <PATH>` a bit more prominent in both help text and The Cargo Book.

Fixes #14629 

Probably just me but I hate wasting space between flags and help texts in some commands

Screenshot 2024-10-01 at 14 27 08

I wonder if we could adopt what `cargo add --help` has done, but that also uses more vertical space.

### DIFF
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -668,7 +668,7 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
             .action(ArgAction::SetTrue)
             .global(true)
             .hide(true))
-        .arg(multi_opt("config", "KEY=VALUE", "Override a configuration value").global(true))
+        .arg(multi_opt("config", "KEY=VALUE|PATH", "Override a configuration value").global(true))
         // Better suggestion for the unsupported lowercase unstable feature flag.
         .arg( Arg::new("unsupported-lowercase-unstable-feature-flag")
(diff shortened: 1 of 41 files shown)


src/bin/cargo/cli.rs
```diff
@@ -668,7 +668,7 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
             .action(ArgAction::SetTrue)
             .global(true)
             .hide(true))
-        .arg(multi_opt("config", "KEY=VALUE", "Override a configuration value").global(true))
+        .arg(multi_opt("config", "KEY=VALUE or PATH", "Override a configuration value").global(true))
```
epage, line 671: The convention for this would be `KEY=VALUE|PATH`

PR #14632, opened 2024-10-01
chore(deps): update tar to 0.4.42

### LINKED ISSUE #14594: Next `tar-rs` update creates sparse entries by default
Recently, the `tar` crate got support for creating archives with sparse entries (https://github.com/alexcrichton/tar-rs/pull/375). It is going to be enabled by default in the next release.

For Cargo (which depends on this crate), keeping the default would mean that some very old versions of Cargo would not be able to read some archives created by the latest Cargo. Specifically, the ability to read sparse entries was added in `tar` `0.4.6` (released 2016-06-09). The corresponding `tar-rs` bump for Cargo was #2973, which is included in Cargo `0.13` which is included in rust `1.12`.

The sparseness check does not affect Rust source files (as text files generally do not contain large chunks of null bytes), but it may affect crates containing binary files. For example, [`protobuf-build`](https://docs.rs/crate/protobuf-build/0.14.0/source/bin/protoc-linux-ppcle_64) and [`metal`](https://docs.rs/crate/metal/0.29.0/source/examples/mps/shaders.metallib) crates contain binaries that could possibly be sparse. The command `rg -ca '\x00{4096}' ~/.cargo/registry/` might reveal other similar binaries.

I'm creating this issue at the [request](https://github.com/alexcrichton/tar-rs/pull/375#issuecomment-2364276823) of @alexcrichton to let the Cargo team decide whether to
- keep the new default as-is, or
- add a few `ar.sparse(false)` calls explicitly everywhere `tar::Builder` is used.

xzfc:
The new version of tar enables the creation of sparse tar archives by default.  The ability to read such sparse entries was added in tar 0.4.6, which has been in use starting from Cargo 0.13 and Rust 1.12. Additionally, `docker cp` doesn't support sparse tar entries in the GNU format.  For compatibility with older versions of Rust and Cargo, as well as with `docker cp`, this commit disables sparse archive creation everywhere the `tar::Builder` is used.

Closes #14594. CC: @weihanglo, @alexcrichton.

### DIFF
--- a/crates/cargo-test-support/src/containers.rs
+++ b/crates/cargo-test-support/src/containers.rs
@@ -122,6 +122,7 @@ impl Container {
             return;
         }
         let mut ar = tar::Builder::new(Vec::new());
+        ar.sparse(false);
         let files = std::mem::replace(&mut self.files, Vec::new());
         for mut file in files {
             ar.append_data(&mut file.header, &file.path, file.contents.as_slice())
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3371,9 +3371,9 @@ dependencies = [
 
 [[package]]
 name = "tar"
-version = "0.4.41"
+version = "0.4.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cb797dad5fb5b76fcf519e702f4a589483b5ef06567f160c392832c1f5e44909"
+checksum = "4ff6c40d3aedb5e06b57c6f669ad17ab063dd1e63d977c6a88e7f4dfa4f04020"
 dependencies = [
  "filetime",
  "libc",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,7 +94,7 @@ shell-escape = "0.1.5"
 similar = "2.6.0"
 supports-hyperlinks = "3.0.0"
 snapbox = { version = "0.6.17", features = ["diff", "dir", "term-svg", "regex", "json"] }
-tar = { version = "0.4.41", default-features = false }
+tar = { version = "0.4.42", default-features = false }
 tempfile = "3.10.1"
 thiserror = "1.0.63"
 time = { version = "0.3.36", features = ["parsing", "formatting", "serde"] }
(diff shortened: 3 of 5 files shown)


build.rs
```diff
@@ -24,6 +24,7 @@ fn compress_man() {
         .write(dst, Compression::best());
     let mut ar = tar::Builder::new(encoder);
     ar.mode(tar::HeaderMode::Deterministic);
+    ar.sparse(false);
```
weihanglo, line 27: This is for internal use only, so may be find with sparse.

benches/capture/src/main.rs
```diff
@@ -48,6 +48,7 @@ fn capture(source_root: &Path, dest: &Path, force: bool) {
         .write(dst, Compression::best());
     let mut ar = tar::Builder::new(encoder);
     ar.mode(tar::HeaderMode::Deterministic);
+    ar.sparse(false);
```
weihanglo, line 51: This is for internal benchmark only, so should be fine with sparse as well.

crates/cargo-test-support/src/containers.rs
```diff
@@ -122,6 +122,7 @@ impl Container {
             return;
         }
         let mut ar = tar::Builder::new(Vec::new());
+        ar.sparse(false);
```
weihanglo, line 125: Container tests are pretty internal stuff. Fine with sparse.

weihanglo:
Thanks for posting this PR!

crates/cargo-test-support/src/containers.rs
weihanglo: > Additionally, docker cp doesn't support sparse tar entries in the GNU format.

Ah. I didn't see this. Thanks for letting us know. Resolved!

build.rs
weihanglo: To be precise, this tarball will be read into the binary at build time, so there shouldn't be any compatibility issue between versions.

benches/capture/src/main.rs
xzfc: Not sure what this bench exactly does, but my thought was that it should be closer to the real-world scenario as checking each file to sparseness might (or might not) affect performance.

Reverted.

weihanglo: These tarballs are snapshots of some real world Cargo workspaces, we don't benchmark on tar/untar of them.

https://github.com/rust-lang/cargo/blob/b5d44db1daf0469b227a6211b987162a39a54730/benches/capture/src/main.rs#L1-L5

Anyhow, thanks for the revert!

weihanglo (approved):

weihanglo:
Thank you!

@bors r+

PR #14636, opened 2024-10-02
docs(ref): Expand on MSRV

epage:

The need for this has been growing and becomes especially important with the upcoming MSRV aware resolver
We will be making having an MSRV easier so we need to meet those maintainers where they are and help them.

In particular, this covers
- What setting an MSRV does
- What "support" in MSRV means, according to the definition we developed during the review of [RFC 3537](https://rust-lang.github.io/rfcs/3537-msrv-resolver.html)
- Background and guidance on setting an MSRV policy

### Additional information

This intentionally leaves out anything related to the MSRV-aware resolver as that will be handled as the feature is stabilized.

### DIFF
--- a/src/doc/src/reference/rust-version.md
+++ b/src/doc/src/reference/rust-version.md
@@ -15,8 +15,6 @@ identifiers such as -nightly will be ignored while checking the Rust version.
 
 To find the minimum `rust-version` compatible with your project, you can use third-party tools like [`cargo-msrv`](https://crates.io/crates/cargo-msrv).
 
-When used on packages that get published, we recommend [verifying the `rust-version`](../guide/continuous-integration.md#verifying-rust-version).
-
 > **MSRV:** Respected as of 1.56
 
 ## Uses
@@ -37,3 +35,31 @@ Other tools may also take advantage of it, like `cargo clippy`'s
 [`incompatible_msrv` lint](https://rust-lang.github.io/rust-clippy/stable/index.html#/incompatible_msrv).
 
 > **Note:** The `rust-version` may be ignored using the `--ignore-rust-version` option.
+
+## Support Expectations
+
+These are general expectations; some packages may document when they do not follow these.
+
+**Complete:**
+
+All functionality, including binaries and API, are available on the supported Rust versions under every [feature](features.md).
+
+**Verified:**
+
+A package's functionality is verified on its supported Rust versions, including automated testing.
+See also our
+[Rust version CI guide](../guide/continuous-integration.md#verifying-rust-version).
+
+**Patchable:**
+
+When licenses allow it,
+users can [override their local dependency](overriding-dependencies.md) with a fork of your package.
+In this situation, Cargo may load the entire workspace for the patched dependency which should work on the supported Rust versions, even if other packages in the workspace have different supported Rust versions.
+
+**Dependency Support:**
+
+In support of the above,
+it is expected that each dependency's version-requirement supports at least one version compatible with your `rust-version`.
+However,
+it is **not** expected that the dependency specification excludes versions incompatible with your `rust-version`.
+In fact, supporting both allows you to balance the needs of users that support older Rust versions with those that don't.
--- a/src/doc/src/reference/rust-version.md
+++ b/src/doc/src/reference/rust-version.md
@@ -1,7 +1,7 @@
 # Rust Version
 
 The `rust-version` field is an optional key that tells cargo what version of the
-Rust language and compiler your package can be compiled with.
+Rust language and compiler you support for your package.
 If the currently selected version of the Rust compiler is older than the stated
 version, cargo will exit with an error, telling the user what version is
 required.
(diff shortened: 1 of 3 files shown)


src/doc/src/reference/rust-version.md
```diff
@@ -0,0 +1,154 @@
...
+identifiers such as -nightly will be ignored while checking the Rust version.
+
+> **MSRV:** Respected as of 1.56
+
+## Uses
+
+**Diagnostics:**
+
+When your package is compiled on an unsupported toolchain,
+Cargo will provide clearer diagnostics than reporting invalid syntax or missing functionality in `std`.
```
joshtriplett, line 23:
```suggestion
Cargo will provide clearer diagnostics about the insufficient toolchain version rather than reporting invalid syntax or missing functionality in the standard library.
```

src/doc/src/reference/rust-version.md
```diff
@@ -0,0 +1,154 @@
...
+This affects all build-targets in the package, including binaries, examples, test suites,
+benchmarks, etc.
+
+**Development aid:**
+
+`cargo add` will auto-select the dependency's version requirement to be low enough to be compatible with your `rust-version`.
```
joshtriplett, line 29:
```suggestion
`cargo add` will auto-select the dependency's version requirement to be the latest version compatible with your `rust-version`, which may not be the latest version of the dependency. `cargo add` will warn if it selects an older version, and tell you what MSRV would be required to use the latest version.
```

epage: For the second half of that, I feel like that is clear from the first half and would rather not get into the specifics of how we behave.

joshtriplett: I think "which may not be the latest version of the dependency" naturally follows but is still helpful reinforcement to help people who aren't familiar with the problem domain wrap their heads around it.

For the latter half, the exact details aren't load-bearing. The main things I'm trying to convey are:
- `cargo add` doesn't do this *silently*
- If you'd rather use the latest version, here's how you can know what MSRV you'd need to use to do that: `cargo add` will tell you

src/doc/src/reference/rust-version.md
```diff
@@ -0,0 +1,154 @@
...
+
+For any dependency shared across policies,
+the lowest common versions must be used as Cargo
+[unifies SemVer-compatible versions](resolver.md#semver-compatibility),
+potentially limiting access to features of the shared dependency for the workspace member with the higher `rust-version`.
+
+### One or More Policies
+
+One way to mitigate the downsides of supporting older Rust versions is to apply your policy to older major or minor versions of your package that you continue to support.
+You likely need still need a policy for what Rust versions the development branch support compared to the release branches for those major or minor versions.
```
weihanglo, line 137:
```suggestion
You likely still need a policy for what Rust versions the development branch support compared to the release branches for those major or minor versions.
```

src/doc/src/reference/rust-version.md
```diff
@@ -0,0 +1,154 @@
...
+By backporting new functionality between minor versions,
+the next available version would be missing it which could be considered a breaking change, violating SemVer.
+Backporting changes also comes with the risk of introducing bugs.
+
+Supporting older versions comes at a cost.
+This cost is dependent on the risk and impact of bugs within the package and what is acceptable for backporting.
+Creating the release branches on-demand and putting the backport burden on the community are ways to balance this cost.
+
+There is not yet a way for dependency management tools to report that the version is still supported,
+shifting the responsibility for users to notice this in documentation.
```
weihanglo, line 151:
```suggestion
shifting the responsibility to users to notice this in documentation.
```

src/doc/src/reference/rust-version.md
epage: Further tweaked it.

src/doc/src/reference/rust-version.md
```diff
@@ -0,0 +1,155 @@
+# Rust Version
+
+The `rust-version` field is an optional key that tells cargo what version of the
```
weihanglo, line 3: Found this :)

```suggestion
The `rust-version` field is an optional key that tells Cargo what version of the
```

src/doc/src/reference/rust-version.md
```diff
@@ -0,0 +1,158 @@
...
+However, this is too unpredictable to be relied on for aligning with the Rust version users track.
+
+The further `rust-version` drifts from your specified policy,
+the more likely users are to infer a policy you did not intend,
+leading to frustration at the unmet expectations.
+
+When drift is allowed,
+there is the question of what is "justifiable enough" to drop supported Versions.
+Each person can come to a reasonably different justification and
+working through that discussion can be frustrating for the involved parties.
```
weihanglo, line 113: Better to read.

```suggestion
Each person can come to a reasonably different justification,
and working through that discussion can be frustrating for the involved parties.
```

src/doc/src/reference/rust-version.md
joshtriplett: LGTM, thanks!

src/doc/src/reference/rust-version.md
epage: That was mirroed from the manifest page which uses that a lot.  I'm unsure if I want to go down that rabbit hole in this PR.

weihanglo (approved):

weihanglo:
Read through this again this morning, this seems good enough to merge. If we'd like to clarify more things or add some examples, we can always deal with that in follow-ups.

@bors r+

PR #14638, opened 2024-10-03
docs(ref): Highleft whats left for msrv-policy

epage:

While there is a tracking issue for these,
I didn't want to have everything under `msrv-policy` to be stabilized,
making it look like it should be moved to the Stable section of the
page, when there are independently stabilizable pieces missing.

### Additional information

This is prep for stabilizing MSRV-aware resover


ehuss:
Thanks!

@bors r+

PR #14639, opened 2024-10-03
feat: Stabilize MSRV-aware resolver config

epage:

This includes
- `cargo generate-lockfile --ignore-rust-version`
- `cargo update --ignore-rust-version`

This does not include
- `edition = "2024"`
- `resolver = "3"`

This is part of #9930

### Additional information

This is stacked on top of #14636.  The commits for this PR start with the commit with a title that matches the PR title.

[FCP](https://github.com/rust-lang/cargo/pull/14639#issuecomment-2392172716)

### DIFF
--- a/src/doc/src/reference/config.md
+++ b/src/doc/src/reference/config.md
@@ -143,6 +143,9 @@ rpath = false            # Sets the rpath linking option.
 [profile.<name>.package.<name>]  # Override profile for a package.
 # Same keys for a normal profile (minus `panic`, `lto`, and `rpath`).
 
+[resolver]
+incompatible-rust-versions = "allow"  # Specifies how resolver reacts to these
+
 [registries.<name>]  # registries other than crates.io
 index = "…"          # URL of the registry index
 token = "…"          # authentication token for the registry
@@ -972,6 +975,30 @@ See [rpath](profiles.md#rpath).
 
 See [strip](profiles.md#strip).
 
+### `[resolver]`
+
+The `[resolver]` table overrides [dependency resolution behavior](resolver.md) for local development (e.g. excludes `cargo install`).
+
+#### `resolver.incompatible-rust-versions`
+* Type: string
+* Default: `"allow"`
+* Environment: `CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS`
+
+When resolving which version of a dependency to use, select how versions with incompatible `package.rust-version`s are treated.
+Values include:
+- `allow`: treat `rust-version`-incompatible versions like any other version
+- `fallback`: only consider `rust-version`-incompatible versions if no other version matched
+
+Can be overridden with
+- `--ignore-rust-version` CLI option
+- Setting the dependency's version requirement higher than any version with a compatible `rust-version`
+- Specifying the version to `cargo update` with `--precise`
+
+See the [resolver](resolver.md#rust-version) chapter for more details.
+
+> **MSRV:**
+> - `allow` is supported on any version
+> - `fallback` is respected as of 1.84
 
 ### `[registries]`
 
--- a/src/doc/src/reference/resolver.md
+++ b/src/doc/src/reference/resolver.md
@@ -73,7 +73,7 @@ Key steps:
 
 ### Version numbers
 
-Cargo prefers the highest version currently available.
+Generally, Cargo prefers the highest version currently available.
 
 For example, if you had a package in the resolve graph with:
 ```toml
@@ -82,6 +82,8 @@ bitflags = "*"
 ```
 If at the time the `Cargo.lock` file is generated, the greatest version of
 `bitflags` is `1.2.1`, then the package will use `1.2.1`.
+
+For an example of a possible exception, see [Rust version](#rust-version).
 
 ### Version requirements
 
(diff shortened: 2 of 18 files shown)


epage:
I propose that we stabilize the MSRV-aware resolver.
- Offer an opt-in config for preferring MSRV-compatible packages over those that are incompatible, regardless of version
- Packages without an MSRV are always considered compatible
- If no MSRV is defined, fallback to the current toolchain's version
- Small tweaks in the algorithm are allowed as changes in the algorithm won't cause changes in lockfiles
- This only applies to local lockfile generation and not to `cargo install`

This includes stabilizing
- `cargo generate-lockfile --ignore-rust-version`
- `cargo update --ignore-rust-version`

This does not include stabilizing
- `edition = "2024"`
- `resolver = "3"`
- Features around the resolver that are meant to support MSRV workflows (e.g. `--update-rust-version` flag)

Deviations from the RFC
- Instead of the MSRV being the lowest among the workspace's MSRVs, we prioritze packages based on how many MSRVs from the workspace they are compatible with
  - We only fallback to the current toolchain's version if no MSRV is set, rather than treating that as the MSRV for those packages

I had considered proposing we have this cover `cargo install`
- When viewed only as a config and keeping in mind that `cargo install` only reads from the user config (and so wouldn't be affected by the current repo's config), it sounds reasonable on the surface
- However, people setting something in their config for `cargo install` would affect any repo that doesn't have its own config.  Really, `cargo install` does need its own `[install.resolver]` table.
- This also gets weirder when it comes to `resolver = "3"` as we'd need to describe that as conditional

@rfcbot fcp merge

src/doc/src/reference/rust-version.md
epage, line 1: #14636 needs to be merged first

src/doc/src/reference/rust-version.md
```diff
@@ -0,0 +1,154 @@
...
+benchmarks, etc.
+
+**Development aid:**
+
+`cargo add` will auto-select the dependency's version requirement to be the latest version compatible with your `rust-version`.
+
+Other tools may also take advantage of it, like `cargo clippy`'s
+[`incompatible_msrv` lint](https://rust-lang.github.io/rust-clippy/master/index.html#/incompatible_msrv).
+
+> **Note:** The `rust-version` may be ignored using the `--ignore-rust-version` option.
```
epage, line 34: This needs to be updated for the MSRV-aware resolver

src/doc/src/reference/rust-version.md
```diff
@@ -0,0 +1,154 @@
...
+
+Cargo allows supporting multiple policies within one workspace.
+
+Verifying specific packages under specific versions can get complicated.
+Tools like [`cargo-hack`](https://crates.io/crates/cargo-hack) can help.
+
+For any dependency shared across policies,
+the lowest common versions must be used as Cargo
+[unifies SemVer-compatible versions](resolver.md#semver-compatibility),
+potentially limiting access to features of the shared dependency for the workspace member with the higher `rust-version`.
```
epage, line 128: This needs to be updated for the MSRV-aware resolver

src/doc/src/reference/config.md
epage, line 1: I also need to talk about this in `resolver.md`

Theghost256 (approved):

src/doc/src/reference/config.md
```diff
@@ -143,6 +143,9 @@ rpath = false            # Sets the rpath linking option.
 [profile.<name>.package.<name>]  # Override profile for a package.
 # Same keys for a normal profile (minus `panic`, `lto`, and `rpath`).
 
+[resolver]
+incompatible-rust-versions = "allow"  # Specifies how resolver reacts to thes
```
ehuss, line 147: The comment here looks truncated, or maybe the last word is misspelled?

src/doc/src/reference/unstable.md
```diff
@@ -343,31 +343,7 @@ This was stabilized in 1.79 in [#13608](https://github.com/rust-lang/cargo/pull/
...
-When resolving a version for a dependency, select how versions with incompatible `package.rust-version`s are treated.
-Values include:
-- `allow`: treat `rust-version`-incompatible versions like any other version
-- `fallback`: only consider `rust-version`-incompatible versions if no other version matched
-
-Can be overridden with
-- `--ignore-rust-version` CLI option
-- Setting the dependency's version requirement higher than any version with a compatible `rust-version`
-- Specifying the version to `cargo update` with `--precise`
+This was stabilized in 1.83 in #.
```
weihanglo, line 346:
```suggestion
This was stabilized in 1.83 in [#14639](https://github.com/rust-lang/cargo/pull/14639).
```

BTW, do you plan not to move anything to "Stabilized" section until every feature is implemented? Should we move just these two stabilized feature under independent h2 headings, which won't break link validity?

src/doc/src/reference/config.md
```diff
@@ -972,6 +975,28 @@ See [rpath](profiles.md#rpath).
 
 See [strip](profiles.md#strip).
 
+## `[resolver]`
```
weihanglo, line 978:
```suggestion
### `[resolver]`
```

src/doc/src/reference/unstable.md
epage: Hadn't really thought that through.  Its a question of whether the stable section is documenting features or flags.  Overall, I don't have a strong preference.

weihanglo: I am slightly towards moving them, but not going to blocker this PR for that reason.

weihanglo (approved):

src/doc/src/reference/config.md
```diff
@@ -972,6 +975,30 @@ See [rpath](profiles.md#rpath).
...
+Can be overridden with
+- `--ignore-rust-version` CLI option
+- Setting the dependency's version requirement higher than any version with a compatible `rust-version`
+- Specifying the version to `cargo update` with `--precise`
+
+See the [resolver](resolver.md#rust-version) chapter for more details.
+
+> **MSRV:**
+> - `allow` is supported on any version
+> - `fallback` is respected as of 1.83
```
epage, line 1001: this needs to be updated if I missed the window before beta branch

src/doc/src/reference/unstable.md
epage: As its already that way for some parts, let's handle the decision separately

src/doc/src/reference/resolver.md
```diff
@@ -201,6 +203,87 @@ ecosystem if you publish a SemVer-incompatible version of a popular library.
...
+
+[dependencies]
+clap = "4.5"  # resolves to 4.5.20
+```
+Though each package has a version requirement for `clap` that would meet its own Rust version,
+because of [version unification](#version-numbers),
+the resolver will need to pick one version that works in both cases and that would be a version like 4.5.20.
+
+[Rust version]: rust-version.md
+[`incompatible-rust-version`]: config.md#resolverincompatible-rust-versions
```
weihanglo, line 285:
```suggestion
[`incompatible-rust-versions`]: config.md#resolverincompatible-rust-versions
```

src/doc/src/reference/resolver.md
```diff
@@ -201,6 +203,87 @@ ecosystem if you publish a SemVer-incompatible version of a popular library.
...
+[package]
+name = "my-cli"
+rust-version = "1.62"
+
+[dependencies]
+clap = "4.0"  # resolves to 4.0.32
+```
+The resolver would pick 4.0.32 as its Rust version is 1.60.0.
+- 4.0.0 also has a Rust version of 1.60.0 but that is a [lower version number](#version-numbers)
+- 4.5.20 is a much [higher version](#version-numbers) and it has a Rust version of 1.74.0 which is compatible with your 1.85 toolchain but it is incompatible with `my-cli`'s Rust version of 1.62
```
weihanglo, line 225: This feels more cohesive to me.

```suggestion
The resolver would pick version 4.0.32 for dependency `clap` because
- 4.0.32 has a Rust version of 1.60.0, which is compatible with `my-cli`'s Rust version of 1.62.
- 4.0.0 also has a Rust version of 1.60.0 but that is a [lower version number](#version-numbers).
- Despite 4.1.0 and later versions all have a much [higher version](#version-numbers), their Rust versions are incompatible with `my-cli`'s Rust vesion. For example, 4.5.20 has a Rust version of 1.74.0 which is compatible with your 1.85 toolchain, but not with `my-cli`'s Rust version of 1.62.
```

src/doc/src/reference/resolver.md
```diff
@@ -201,6 +203,87 @@ ecosystem if you publish a SemVer-incompatible version of a popular library.
...
+that is compatible with Rust version 1.62.
+The resolver will then pick an incompatible version, like 4.5.20 despite it having a Rust version of 1.74.
+
+When the resolver selects a dependency version of a package,
+it does not know all the workspace members that will eventually have a transitive dependency on that version
+and so it cannot take into account only the Rust versions relevant for that dependency.
+The resolver has heuristics to find a "good enough" solution when workspace members have different Rust versions.
+This applies even for packages in a workspace without a Rust version.
+
+With multiple Rust versions, the resolver may pick a lower dependency version than necessary.
```
weihanglo, line 248: Make it explicit?

```suggestion
When a workspace has members with multiple different Rust versions,
the resolver may pick a lower dependency version than necessary.
```

or be clear we're talking about the above case?

```suggestion
In the above situation, the resolver may pick a lower dependency version than necessary.
```

weihanglo (commented):
Some nitpicks as a reader that often forgot what I was reading in previous paragraphs.

src/doc/src/reference/resolver.md
epage: Note that that is mixing sentence structures.  The header was changed to end with "because" and "because despite" doesn't make sense to say.  I'm working on my own teak.

src/doc/src/reference/resolver.md
```diff
@@ -201,6 +203,88 @@ ecosystem if you publish a SemVer-incompatible version of a popular library.
...
+equal to or greater than your own Rust version.
+For example, you are using Rust 1.85 to develop the following package:
+```toml
+[package]
+name = "my-cli"
+rust-version = "1.62"
+
+[dependencies]
+clap = "4.0"  # resolves to 4.0.32
+`
```
weihanglo, line 222: Should be three backticks

src/doc/src/reference/resolver.md
weihanglo: Thanks! Looks nicer :)

weihanglo (approved):

weihanglo:
@bors r+

The FCP has ended. To maximise the testing window I am going to merge this soon. Thank you all for your inputs!

PR #14642, opened 2024-10-04
test: Migrate publish snapshotting to snapbox

epage:

This is part of #14039 and allows snapshotting in more places.

### Additional information

`InMemoryDir` is taken from some experiments I've been doing in snapshot for filesystem assertions.  That got held up because of design complexity but publish validation's needs are simpler than I was designing for.


crates/cargo-test-support/src/publish.rs
```diff
@@ -216,15 +231,7 @@ pub fn validate_crate_contents(
...
         );
     }
-    if !expected_contents.is_empty() {
-        for (e_file_name, e_file_contents) in expected_contents {
-            let e_file_name = Path::new(e_file_name);
-            let actual_contents = files
-                .get(e_file_name)
-                .unwrap_or_else(|| panic!("file `{}` missing in archive", e_file_name.display()));
-            assert_match_exact(e_file_contents, actual_contents);
-        }
```
epage, line 215: Something still missing is the reporting of which path failed

Theghost256 (approved):

crates/cargo-test-support/src/compare.rs
```diff
@@ -706,6 +669,136 @@ impl fmt::Debug for WildStr<'_> {
...
+}
+
+impl From<()> for InMemoryDir {
+    fn from(_files: ()) -> Self {
+        let files = Vec::new();
+        Self { files }
+    }
+}
+
+macro_rules! impl_from_tuple_for_dirsnapshot {
```
weihanglo, line 767: I know it is working, and I understand it is for supporting up to 9 tuples (right?). However the macro is a bit hard to read for me. Is this a de-facto decl macro pattern? Is it possible to document what it is doing?

weihanglo (approved):

weihanglo (approved):

crates/cargo-test-support/src/compare.rs
weihanglo: Thanks!

`r=me` when the version bump is fixed

PR #14647, opened 2024-10-05
improve error reporting when feature not found in `activated_features`

elchukc:
Pulls the error message refactor out of #14593 (originally #13207) to improve error reporting when we fail to get the list of activated features enabled for the given package. It now fully lists the activated_features hashmap keys too.

From the [original author](https://github.com/rust-lang/cargo/pull/13207#issue-2056019109):

> I moved `activated_features_int` into `activated_features` and `activated_features_unverified` as I was concerned of the performance cost of generating the full error when its not a fatal error and may occur many times.

Old vs new error message:
```diff
- activated_features for invalid package: features did not find PackageId { name: "bindep", version: "0.0.0", source: "[ROOT]/foo/bindep" } NormalOrDev
+ did not find features for (PackageId { name: "bindep", version: "0.0.0", source: "[ROOT]/foo/bindep" }, NormalOrDev) within activated_features:
+ [
+     (
+         PackageId {
+             name: "bindep",
+             version: "0.0.0",
+             source: "[ROOT]/foo/bindep",
+         },
+         ArtifactDep(
+             CompileTarget {
+                 name: "[ALT_TARGET]",
+             },
+         ),
+     ),
+     (
+         PackageId {
+             name: "foo",
+             version: "0.0.0",
+             source: "[ROOT]/foo",
+         },
+         NormalOrDev,
+     ),
+ ]
```
r? weihanglo

### DIFF
--- a/src/cargo/core/resolver/features.rs
+++ b/src/cargo/core/resolver/features.rs
@@ -319,8 +319,30 @@ impl ResolvedFeatures {
         pkg_id: PackageId,
         features_for: FeaturesFor,
     ) -> Vec<InternedString> {
-        self.activated_features_int(pkg_id, features_for)
-            .expect("activated_features for invalid package")
+        if let Some(res) = self.activated_features_unverified(pkg_id, features_for) {
+            res
+        } else {
+            panic!(
+                "did not find features for ({pkg_id:?}, {features_for:?}) within activated_features:\n{:#?}",
+                self.activated_features.keys()
+            )
+        }
+    }
+
+    /// Variant of `activated_features` that returns `None` if this is
+    /// not a valid pkg_id/is_build combination. Used in places which do
+    /// not know which packages are activated (like `cargo clean`).
+    pub fn activated_features_unverified(
+        &self,
+        pkg_id: PackageId,
+        features_for: FeaturesFor,
+    ) -> Option<Vec<InternedString>> {
+        let fk = features_for.apply_opts(&self.opts);
+        if let Some(fs) = self.activated_features.get(&(pkg_id, fk)) {
+            Some(fs.iter().cloned().collect())
+        } else {
+            None
+        }
     }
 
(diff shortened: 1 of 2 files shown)


src/cargo/core/resolver/features.rs
```diff
@@ -319,8 +319,32 @@ impl ResolvedFeatures {
...
-        self.activated_features_int(pkg_id, features_for)
-            .expect("activated_features for invalid package")
+        let fk = features_for.apply_opts(&self.opts);
+        let key = (pkg_id, fk);
+        if let Some(fs) = self.activated_features.get(&key) {
+            fs.iter().cloned().collect()
+        } else {
+            panic!(
+                "did not find features for {key:?} within activated_features:\n{:#?}",
+                self.activated_features
```
elchukc, line 329: I could also print `self.activated_features.keys()` to clean up the output a tiny bit
```diff
...
    (
        PackageId {
            name: "foo",
            version: "0.0.0",
            source: "[ROOT]/foo",
        },
        NormalOrDev,
+    ),
-    ): {},
...
```

weihanglo: That sounds great!!

src/cargo/core/resolver/features.rs
```diff
@@ -319,8 +319,32 @@ impl ResolvedFeatures {
         pkg_id: PackageId,
         features_for: FeaturesFor,
     ) -> Vec<InternedString> {
-        self.activated_features_int(pkg_id, features_for)
```
weihanglo, line 322: What about reusing and calling the unverified variant in this function. If none then panic?

elchukc: Good idea!

weihanglo (approved):

weihanglo:
Thank you!

@bors r+

PR #14649, opened 2024-10-06
Implement RFC3695: Allow boolean literals as cfg predicates

Urgau:

This PR implements https://github.com/rust-lang/rfcs/pull/3695: allow boolean literals as cfg predicates, i.e. `cfg(true)` and `cfg(false)`.

The PR should be reviewed commit by commit and tested by looking at the tests and using `[target.'cfg(<true/false>)']` in `Cargo.toml`/`.cargo/config.toml`.

### Additional information

I had to bump `cargo-platform` to `0.3.0` has we are changing `CfgExpr` enum in a semver incompatible change.

We currently have (thks to https://github.com/rust-lang/cargo/pull/14671) a forward compatibility warning against `cfg(true/false)` as identifiers instead of keywords.

~~I choose a use a `Cargo.toml` feature (for the manifest) as well as a unstable CLI flag for the `.cargo/config.toml` part.~~

~~Given the very small (two occurrences on Github Search) for [`cfg(true)`](https://github.com/search?q=lang%3Atoml+%2F%28%3F-i%29%5C%5Btarget%5C.%5B%27%22%5Dcfg.*true%2F&type=code) and [`cfg(false)`](https://github.com/search?q=lang%3Atoml+%2F%28%3F-i%29%5C%5Btarget%5C.%5B%27%22%5Dcfg.*false%2F&type=code), I choose to gate the feature under a error and not a warning.~~

### DIFF
--- a/crates/cargo-platform/src/lib.rs
+++ b/crates/cargo-platform/src/lib.rs
@@ -98,6 +98,7 @@ impl Platform {
                         ))
                     },
                 }
+                CfgExpr::True | CfgExpr::False => {},
             }
         }
 
@@ -115,30 +116,18 @@ impl Platform {
                         check_cfg_expr(e, warnings, path);
                     }
                 }
+                CfgExpr::True | CfgExpr::False => {}
                 CfgExpr::Value(ref e) => match e {
                     Cfg::Name(name) | Cfg::KeyPair(name, _) => {
                         if !name.raw && KEYWORDS.contains(&name.as_str()) {
-                            if name.as_str() == "true" || name.as_str() == "false" {
-                                warnings.push(format!(
-                                    "[{}] future-incompatibility: the meaning of `cfg({e})` will change in the future\n \
-                                     | Cargo is erroneously allowing `cfg(true)` and `cfg(false)`, but both forms are interpreted as false unless manually overridden with `--cfg`.\n \
-                                     | In the future these will be built-in defines that will have the corresponding true/false value.\n \
-                                     | It is recommended to avoid using these configs until they are properly supported.\n \
-                                     | See <https://github.com/rust-lang/rust/issues/131204> for more information.\n \
-                                     |\n \
-                                     | help: use raw-idents instead: `cfg(r#{name})`",
-                                    path.display()
-                                ));
-                            } else {
-                                warnings.push(format!(
-                                    "[{}] future-incompatibility: `cfg({e})` is deprecated as `{name}` is a keyword \
-                                     and not an identifier and should not have have been accepted in this position.\n \
-                                     | this was previously accepted by Cargo but is being phased out; it will become a hard error in a future release!\n \
-                                     |\n \
-                                     | help: use raw-idents instead: `cfg(r#{name})`",
-                                     path.display()
-                                ));
-                            }
+                            warnings.push(format!(
+                                "[{}] future-incompatibility: `cfg({e})` is deprecated as `{name}` is a keyword \
+                                 and not an identifier and should not have have been accepted in this position.\n \
+                                 | this was previously accepted by Cargo but is being phased out; it will become a hard error in a future release!\n \
+                                 |\n \
+                                 | help: use raw-idents instead: `cfg(r#{name})`",
+                                 path.display()
+                            ));
                         }
                     }
                 },
--- a/tests/testsuite/cfg.rs
+++ b/tests/testsuite/cfg.rs
@@ -659,3 +659,238 @@ fn cfg_keywords() {
 "#]])
         .run();
 }
+
+#[cargo_test]
+fn cfg_booleans() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "a"
+                version = "0.0.1"
+                edition = "2015"
+                authors = []
+
+                [target.'cfg(true)'.dependencies]
+                b = { path = 'b' }
+                
+                [target.'cfg(false)'.dependencies]
+                c = { path = 'c' }
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .file("b/Cargo.toml", &basic_manifest("b", "0.0.1"))
+        .file("b/src/lib.rs", "")
+        .file("c/Cargo.toml", &basic_manifest("c", "0.0.1"))
+        .file("c/src/lib.rs", "")
+        .build();
+
+    p.cargo("check")
+        // FIXME: `b` should be compiled
+        .with_stderr_data(str![[r#"
+[WARNING] [[ROOT]/foo/Cargo.toml] future-incompatibility: the meaning of `cfg(false)` will change in the future
+ | Cargo is erroneously allowing `cfg(true)` and `cfg(false)`, but both forms are interpreted as false unless manually overridden with `--cfg`.
+ | In the future these will be built-in defines that will have the corresponding true/false value.
+ | It is recommended to avoid using these configs until they are properly supported.
+ | See <https://github.com/rust-lang/rust/issues/131204> for more information.
+ |
+ | [HELP] use raw-idents instead: `cfg(r#false)`
+[WARNING] [[ROOT]/foo/Cargo.toml] future-incompatibility: the meaning of `cfg(true)` will change in the future
+ | Cargo is erroneously allowing `cfg(true)` and `cfg(false)`, but both forms are interpreted as false unless manually overridden with `--cfg`.
+ | In the future these will be built-in defines that will have the corresponding true/false value.
+ | It is recommended to avoid using these configs until they are properly supported.
+ | See <https://github.com/rust-lang/rust/issues/131204> for more information.
+ |
+ | [HELP] use raw-idents instead: `cfg(r#true)`
+[LOCKING] 2 packages to latest compatible versions
+[CHECKING] a v0.0.1 ([ROOT]/foo)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
+        .run();
+}
+
+#[cargo_test]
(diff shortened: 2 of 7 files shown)


src/cargo/util/context/target.rs
```diff
@@ -53,6 +54,17 @@ pub(super) fn load_target_cfgs(
     let target: BTreeMap<String, TargetCfgConfig> = gctx.get("target")?;
     tracing::debug!("Got all targets {:#?}", target);
     for (key, cfg) in target {
+        if let Ok(Platform::Cfg(cfg_expr)) = key.parse() {
+            cfg_expr.walk_expr(|e| match e {
+                CfgExpr::True | CfgExpr::False => {
+                    if !gctx.cli_unstable().cfg_boolean_literals {
```
epage, line 60: nit: Should this condition be moved out of the new if-let?  If we know its not going to fail, why bother doing all of this?

epage, line 60: Should we have a deprecation warning for `Name("true")` / `Name("false")` in existing versions, here and in `toml`?

If we do this, then no point in moving the if like I mentioned

epage, line 60: How does stable and unstable rustc respond to `--cfg true` and `--cfg false`?  I wonder if we (cargo+rustc) should also report future-incompat messages in those cases

epage: Strictly speaking, so long as a user doesn't set `--cfg false`, the use of `#[cfg(false)]` won't fundamentally change, so maybe we don't need a future-incompat for that?

epage, line 60: > Given the very small (two occurrences on Github Search) for cfg(true) and cfg(false), I choose to gate the feature under a error and not a warning.

As `cfg(true)` would have dramatically different behavior on stable before and after this is the correct choice.

We could have leeway on `cfg(false)` as I just mentioned.

Urgau: `rustc` happily accepts `--cfg true`/`--cfg false`, as it's well defined in the Rust language.

`cfg(true)`/`cfg(false)` was never accepted on stable (or nightly even), to use it you would have needed to use raw idents `cfg(r#true)`, which works in `rustc` but not in Cargo

Urgau: > Should we have a deprecation/future-incompat warning for `Name("true")` / `Name("false")` in existing versions, here and in toml?

We could, but I don't think it's worth it, we only have two users and I expect the RFC to become stable relatively quickly.

> Strictly speaking, so long as a user doesn't set `--cfg false`, the use of `#[cfg(false)]` won't fundamentally change

Technically that's not true for `rustc` as `#[cfg(false)]` was never legal before the RFC.

epage:
A more generalized search: https://github.com/search?q=%2F%28%3F-i%29cfg%5C%28.*%5Cb%28true%7Cfalse%29%5Cb%2F+language%3ATOML&type=code&ref=advsearch

src/cargo/util/context/target.rs
epage: Two users on github.  That isn't a complete search of all code.  I'm just hesitant about turning this into a hard error without any prior notice.

@weihanglo, thoughts?

weihanglo: > Technically that's not true for rustc as #[cfg(false)] was never legal before the RFC.

If we treat it as a Cargo bug, we can literally ship this breaking change 😬.

Being conservative, I vote for warnings. We could give a “clear timeline” when it will become a hard error, and note that in our CHANGELOG.

weihanglo: I wonder what other cfg value rustc disallows but Cargo is happy with. We may also fix them altogether.

Urgau: > If we treat it as a Cargo bug, we can literally ship this breaking change 😬.

That's my position, I would consider the Cargo parser to be in-sufficient and buggy, and just ship the breaking change ...

> Being conservative, I vote for warnings. We could give a “clear timeline” when it will become a hard error, and note that in our CHANGELOG.

... but that doesn't seems to be either of your position. 😄 

Which leads me to ask, what kind of warning you want? A future-incompatibility warning with no change in behaviour? A warning but with change in behaviour? Or a future-incompatibility warning but with unstable flag and feature to change the behaviour?

That last one would be my preferred one, as it would allow nightly testing, but it would be the most involved as currently the parsing is done via [`<CfgExpr as FromStr>::from_str(&str)`](https://doc.rust-lang.org/beta/nightly-rustc/cargo_platform/enum.CfgExpr.html#impl-FromStr-for-CfgExpr) which doesn't take any options, which will would need to configure the parser behaviour.

epage: I think an unstable-opt-in is fine to change from the future-incompat warning to the new behavior.

weihanglo: We may also want to update the doc to note that some cfg values are not intended to use. Even if they are valid in `Cargo.toml` or `.cargo/config.toml`, they are unlikely to be picked by rustc.

There are two categories I can think of people might misuse:

* `[target.'cfg(true)'.dependencies]` being a way to activate additional features
  * The `[features]` table is the right approach of it.
* `target.'cfg(ture).rustflags` being a way to set/overwrite rustflags
  * `cargo --config` should be used instead, or `RUSTFLAGS` or `CARGO_BUILD_RUSTFLAGS`.

Urgau: Opened #14671 to address this thread *(as this PR is already quite large)*.

Urgau:
The future incompatibility warning was merged in #14671.

I tried adding an unstable Cargo feature but that would I mean either making not insignificant refactor to the parsing or evaluation, or adding a hack to transform back, neither felt justified enough for such a temporary mechanism.

Therefore this PR is blocked on enough time having passed to being able to unconditional apply to new logic.

@rustbot label -S-waiting-on-review +S-blocked-external

weihanglo:
> Therefore this PR is blocked on enough time having passed to being able to unconditional apply to new logic.

* Do we have any discussion of “enough time” elsewhere like on Zulip?
* Should we wait for stabilization of <https://github.com/rust-lang/rust/issues/131204>?

Urgau:
> * Do we have any discussion of “enough time” elsewhere like on Zulip?

Not that I'm aware.

> * Should we wait for stabilization of [Tracking issue for RFC 3695: Allow boolean literals as cfg predicates rust#131204](https://github.com/rust-lang/rust/issues/131204)?

I'm planning on issuing after the holidays a call-for-testing and then directly propose a T-lang stabilization. This could mean only 1 or 2 releases with the future compact warning, not sure if you (T-cargo) would be comfortable with that.

We could also wait and stabilize both at the same time, but I don't think that's required.

Urgau:
We are currently having a T-lang FCP of the feature in https://github.com/rust-lang/rust/pull/138632 (with 5/5 check-boxes).

I'm expecting it to land in 1.87 (it branches at the end of this week), this should let a full release cycle for T-cargo to decide to land this or not. The future incompatibility warning landed in 1.85 so it would be 2 release cycles with the warning.

@rustbot labels -S-blocked-external -S-waiting-on-author -A-unstable +S-waiting-on-review

crates/cargo-platform/src/lib.rs
```diff
@@ -115,30 +116,18 @@ impl Platform {
...
-                                warnings.push(format!(
-                                    "[{}] future-incompatibility: `cfg({e})` is deprecated as `{name}` is a keyword \
-                                     and not an identifier and should not have have been accepted in this position.\n \
-                                     | this was previously accepted by Cargo but is being phased out; it will become a hard error in a future release!\n \
-                                     |\n \
-                                     | help: use raw-idents instead: `cfg(r#{name})`",
-                                     path.display()
-                                ));
-                            }
+                            warnings.push(format!(
```
epage, line 123: Do we not have a test for this remaining future-incompat warning?

epage:
Thanks @Urgau for continuing to drive this!

crates/cargo-platform/src/lib.rs
Urgau: I don't think so, the keywords list currently only contains `true` and `false`.

https://github.com/rust-lang/cargo/blob/a6c604d1b8a2f2a8ff1f3ba6092f9fda42f4b7e9/crates/cargo-platform/src/cfg.rs#L47-L53

Urgau:
From my side yes, I don't know if you also want to do an FCP or if you consider the T-lang FCP to be enough, but either way the changes in this PR are at minimum ready for review.

tests/testsuite/cfg.rs
```diff
@@ -659,3 +659,238 @@ fn cfg_keywords() {
...
+ |
+ | [HELP] use raw-idents instead: `cfg(r#true)`
+[LOCKING] 2 packages to latest compatible versions
+[CHECKING] b v0.0.1 ([ROOT]/foo/b)
+[CHECKING] c v0.0.1 ([ROOT]/foo/c)
+[CHECKING] a v0.0.1 ([ROOT]/foo)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
+        .env("RUSTFLAGS", "--cfg true --cfg false")
```
epage: Could you put all of the inputs for the test case before the outputs?  I was confused for a while at how this was a rustflags test before I finally noticed this.

Sorry I didn't notice this earlier.

epage:
Thanks for driving this!

We aren't holding our own FCP, relying on the T-compiler FCP

Once the test is cleaned up, this should be good to merge on our side.

tests/testsuite/cfg.rs
Urgau: Sure, fixed.

epage (approved):

PR #7902, opened 2020-02-18
Remove the `--build-plan` option

alexcrichton:
This commit removes the "build plan" feature from Cargo. More
information can be found in #7614.

PR #14657, opened 2024-10-08
Document build-plan as being deprecated

ehuss:
This adds a note to the build-plan documentation that it is deprecated. I do not think it will make any progress as it is. In the future, we should remove the feature (#7902), possibly with a code-warning ahead of time.


weihanglo (approved):
image


https://rust-lang.github.io/mdBook/format/mdbook.html#classwarning

I didn't know there is a predefined CSS class for warning in mdbook. Awesome.

PR #14658, opened 2024-10-09
WIP Add bin/cdylib/staticlib suffix for artifact deps

### LINKED ISSUE #14804: Display artifact dependency information in `cargo tree`
Some questions I asked in <https://github.com/rust-lang/cargo/issues/10593#issuecomment-1341214604>:

* Should we annotate artifact dependency nodes with a special `(artifact deps)` suffix?
* Should we show the information about which `target` platform an artifact dependency will be built on?
* How do we want to display the node of an artifact dependency with `{ …, lib = true }`?

And some replies from @bstrie 

> Should we annotate artifact dependency nodes with a special (artifact deps) suffix?

There are multiple possible values for the `artifact` field, so I wouldn't mind having a `(bin)` or `(staticlib)` or `(cdylib)` suffix.

> Should we show the information about which target platform an artifact dependency will be built on?

No, I think that's a detail that can be easily answered by checking Cargo.toml. But if you did decide to display this information, I would include it as part of the above annotation, e.g. `(wasm32-wasi bin)` or `(x86_64-unknown-linux-musl staticlib)`.

> How do we want to display the node of an artifact dependency with { …, lib = true }?

I think it should show up multiple times in the tree graph. The lib node should show up as an entirely normal dependency, whereas the artifact node should be annotated as mentioned above.

## Propose solution

Since this is an unstable, we can ship an initial version and iterate on it then.

## Version

d8cb5fb7f8f66fc9b4ba3d80b089474164a8a83d

elchukc:

This is an implementation of cargo tree as defined in [#10593 (comment)](https://github.com/rust-lang/cargo/issues/10593#issuecomment-1341214604)

Fixes #14804

### Additional information

r? @weihanglo

### DIFF
--- a/src/cargo/core/package.rs
+++ b/src/cargo/core/package.rs
@@ -19,7 +19,7 @@ use serde::Serialize;
 use tracing::debug;
 
 use crate::core::compiler::{CompileKind, RustcTargetData};
-use crate::core::dependency::DepKind;
+use crate::core::dependency::{ArtifactKind, DepKind};
 use crate::core::resolver::features::ForceAllTargets;
 use crate::core::resolver::{HasDevUnits, Resolve};
 use crate::core::{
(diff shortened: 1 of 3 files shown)


src/cargo/core/package.rs
```diff
@@ -167,6 +167,14 @@ impl Package {
     pub fn proc_macro(&self) -> bool {
         self.targets().iter().any(|target| target.proc_macro())
     }
+    // TODO fix this. For now, just wanted it to return a plausible value. Must figure out why .kinds() returns a Vec.
+    /// Gets crate-type in { .., artifact = <crate-type> } of this package
```
elchukc, line 171: This returns ArtifactKind of a package's _dependencies_, not the ArtifactKind of the package itself. This is why (bin/cdylib/staticlib) suffixes currently print on the wrong line.
```diff
- expected
+ actual
- foo v0.0.0 ([ROOT]/foo)
-└── bindep v0.0.0 (bin) ([ROOT]/foo/bindep)
+ foo v0.0.0 (bin) ([ROOT]/foo)
+ └── bindep v0.0.0 ([ROOT]/foo/bindep)
```

weihanglo:
@rustbot author

@elchukc are you still pursuing this?

epage:
As there has been no contact with the author for over a year, I'm going to go ahead and close this.

PR #14662, opened 2024-10-09
docs(resolver): Lay groundwork for documenting MSRV-aware logic

epage:

This is prep for document the MSRV-aware resolver (see #14639), in particular
- This give more breathing room for adding this new heuristic to the resolver documentation
- This provides the context for understanding the limitations

In moving documentation, I asked the question "where would I look to find this if I had a question on it".  I tried to balance this by not putting too much formal / technical documentation in more guide-level descriptions.  In particular, while "Specifying Dependencies" is in the reference, its also written in somewhat of a guide-style.

There is likely more work that can be done, including
- Maybe making the "SemVer Compatibility" chapter the de facto reference for Cargo's version of semver that other sections reference for a more exhaustive description.
- Splitting discussion of the Feature resolver out of the resolver and features documentation.  In the current implementation, we have 3 resolve phases (1) lockfile, (2) adapt to the current compilation, (3) resolve features.  The last two really serve the same role and I'd consider merging discussion of them.

I tried to break up changes in smaller to digest chunks.  This means some times a new section doesn't fully jive with another section until a follow up commit.  I'd recommend reviewing by commit while having the full diff up on the side to see if a concern is still relevant.

### DIFF
--- a/src/doc/src/reference/manifest.md
+++ b/src/doc/src/reference/manifest.md
@@ -92,6 +92,20 @@ a keyword. [crates.io] imposes even more restrictions, such as:
 
 ### The `version` field
 
+The `version` field is formatted according to the [SemVer] specification:
+
+Versions must have three numeric parts,
+the major version, the minor version, and the patch version.
+
+A pre-release part can be added after a dash such as `1.0.0-alpha`.
+The pre-release part may be separated with periods to distinguish separate
+components. Numeric components will use numeric comparison while
+everything else will be compared lexicographically.
+For example, `1.0.0-alpha.11` is higher than `1.0.0-alpha.4`.
+
+A metadata part can be added after a plus, such as `1.0.0+21AF26D3`.
+This is for informational purposes only and is generally ignored by Cargo.
+
 Cargo bakes in the concept of [Semantic
 Versioning](https://semver.org/), so make sure you follow some basic rules:
 
@@ -103,7 +117,6 @@ Versioning](https://semver.org/), so make sure you follow some basic rules:
 * After 1.0.0, don’t add any new public API (no new `pub` anything) in patch-level
   versions. Always increment the minor version if you add any new `pub` structs,
   traits, fields, types, functions, methods or anything else.
-* Use version numbers with three numeric parts such as 1.0.0 rather than 1.0.
 
 See the [Resolver] chapter for more information on how Cargo uses versions to
 resolve dependencies, and for guidelines on setting your own version. See the
@@ -114,6 +127,7 @@ This field is optional and defaults to `0.0.0`.  The field is required for publi
 
 > **MSRV:** Before 1.75, this field was required
 
+[SemVer]: https://semver.org
 [Resolver]: resolver.md
 [SemVer compatibility]: semver.md
 
--- a/src/doc/src/reference/manifest.md
+++ b/src/doc/src/reference/manifest.md
@@ -106,22 +106,10 @@ For example, `1.0.0-alpha.11` is higher than `1.0.0-alpha.4`.
 A metadata part can be added after a plus, such as `1.0.0+21AF26D3`.
 This is for informational purposes only and is generally ignored by Cargo.
 
-Cargo bakes in the concept of [Semantic
-Versioning](https://semver.org/), so make sure you follow some basic rules:
-
-* Before you reach 1.0.0, anything goes, but if you make breaking changes,
-  increment the minor version. In Rust, breaking changes include adding fields to
-  structs or variants to enums.
-* After 1.0.0, only make breaking changes when you increment the major version.
-  Don’t break the build.
-* After 1.0.0, don’t add any new public API (no new `pub` anything) in patch-level
-  versions. Always increment the minor version if you add any new `pub` structs,
-  traits, fields, types, functions, methods or anything else.
-
+Cargo bakes in the concept of [Semantic Versioning](https://semver.org/),
+so versions are considered considered [compatible](semver.md) if their left-most non-zero major/minor/patch component is the same.
 See the [Resolver] chapter for more information on how Cargo uses versions to
-resolve dependencies, and for guidelines on setting your own version. See the
-[SemVer compatibility] chapter for more details on exactly what constitutes a
-breaking change.
+resolve dependencies.
 
 This field is optional and defaults to `0.0.0`.  The field is required for publishing packages.
 
--- a/src/doc/src/reference/resolver.md
+++ b/src/doc/src/reference/resolver.md
@@ -137,41 +137,6 @@ ecosystem if you publish a SemVer-incompatible version of a popular library.
 [semver trick]: https://github.com/dtolnay/semver-trick
 [`downcast_ref`]: ../../std/any/trait.Any.html#method.downcast_ref
 
-### Pre-releases
-
-SemVer has the concept of "pre-releases" with a dash in the version, such as
-`1.0.0-alpha`, or `1.0.0-beta`. Cargo will avoid automatically using
-pre-releases unless explicitly asked. For example, if `1.0.0-alpha` of package
-`foo` is published, then a requirement of `foo = "1.0"` will *not* match, and
-will return an error. The pre-release must be specified, such as `foo =
-"1.0.0-alpha"`. Similarly [`cargo install`] will avoid pre-releases unless
-explicitly asked to install one.
-
-Cargo allows "newer" pre-releases to be used automatically. For example, if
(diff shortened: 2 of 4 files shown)


src/doc/src/reference/resolver.md
```diff
@@ -5,99 +5,127 @@ use based on the version requirements specified in each package. This process
...
+```
 
+Key steps:
+- Walking dependencies (`pick_next_dep`):
+  The order dependencies are walked can affect
+  how related version requirements for the same dependency get resolved, see unifying versions,
+  and how much the resolver backtracks, affecting resolver performance,
+- Unifying versions (`try_unify_version`, `needs_version_unification`):
+  Ensuring Cargo reuses versions affects build time and allows types from common dependencies to be passed between APIs.
+  Cargo would prefer to backtrack or error rather than include two versions that would have unified except their [dependency specifications] conflicted.
```
Eh2406, line 67: I don't understand what the sentence is trying to say.

src/doc/src/reference/resolver.md
```diff
@@ -5,99 +5,127 @@ use based on the version requirements specified in each package. This process
...
+- Ignoring versions (`filter_versions`):
+  A [dependency specification] or Cargo may decide that a version is undesirable,
+  preferring to backtrack or error rather than use it.
+- Preferring versions (`pick_next_version`):
+  Cargo may decide that we should prefer a specific version,
+  falling back to the next when backtracking.
+  The resolver cannot backtrack because a more-preferred version becomes available.
```
Eh2406, line 74: Also having trouble parsing this one.

epage: Which part?

The first half is trying to describe our sorting of versions.

The second half is trying to deal with limitations in relying on version order, e.g. a package without an MSRV depending on one that does and we can't backtrack to pick a better version of the non-MSRV package.

src/doc/src/reference/resolver.md
Eh2406: I was confused by "Cargo would prefer to backtrack or error rather than include two versions that would have unified except their [dependency specifications] conflicted."

src/doc/src/reference/resolver.md
Eh2406: I was having trouble with "The resolver cannot backtrack because a more-preferred version becomes available."

src/doc/src/reference/resolver.md
epage: That is talking about the inability to unify `"=1.2.4"` with `"=1.5.6"`.

src/doc/src/reference/resolver.md
epage: With the explanation I gave, any thoughts on how to cover this?  Or should we skip it and leave it to the MSRV section that I'll add later?

Eh2406: Can you elaborate on that situation? What's the alternative design that we would love to have if it was technically possible?

The hope of these questions is that having written it out it becomes clear what point should be summarized here, or it becomes clear that it should be its own paragraph in the MSRV section. In either case the writing/thinking that is done in this comment can be used in that follow-up.

epage: This is less about design decisions but more about implications.  I'm trying to help the user take the principles and see the effect.

src/doc/src/reference/resolver.md
epage: > If two [dependency specifications] conflict then they can not be unified on a single version. Cargo backtracks or errors in this case.

A challenge in writing this is to be so clear we cannot be misunderstood.  saying two dependency specifications that can conflict is open ended for the user to make assumptions about what may conflict, like major versions.  This is why I was emphasizing that we attempt to unify but can't.  Working on a new sentence based on your feedback.

epage: Next proposal

> If multiple versions would have been unified if it wasn't for conflicts in their [dependency specifications], Cargo will backtrack, erroring if no solution is found, rather than selecting multiple versions.

src/doc/src/reference/resolver.md
epage: I went ahead and removed it as its hard to talk about it in a theoretical sense as its hard to describe how the resolver could hit this today.  With my MSRV documentation, it will be eaiser.

src/doc/src/reference/resolver.md
Eh2406: That's definitely better. We still have some ambiguity about what can and can't be unified.

Eh2406 (approved):
Thank you for adding this information and clarification. I think it makes the situation a lot clearer for our users!

I'm not going to single-handedly r+, because documentation requires review from many perspectives. But I'm excited to see this land once  others have had a chance to review it.

src/doc/src/reference/resolver.md
```diff
@@ -71,31 +71,48 @@ Key steps:
...
-compatible.
-
-When multiple packages specify a dependency for a common package, the resolver
-attempts to ensure that they use the same version of that common package, as
-long as they are within a SemVer compatibility range. It also attempts to use
-the greatest version currently available within that compatibility range. For
-example, if there are two packages in the resolve graph with the following
-requirements:
+Cargo assumes packages follow [SemVer] and will unify dependency versions if they are
+[SemVer] compatible according to the [Caret version requirements].
```
weihanglo, line 106: Should we call it default version requirement instead of caret?

src/doc/src/reference/manifest.md
```diff
@@ -92,28 +92,30 @@ a keyword. [crates.io] imposes even more restrictions, such as:
...
+the major version, the minor version, and the patch version.
+
+A pre-release part can be added after a dash such as `1.0.0-alpha`.
+The pre-release part may be separated with periods to distinguish separate
+components. Numeric components will use numeric comparison while
+everything else will be compared lexicographically.
+For example, `1.0.0-alpha.11` is higher than `1.0.0-alpha.4`.
+
+A metadata part can be added after a plus, such as `1.0.0+21AF26D3`.
+This is for informational purposes only and is generally ignored by Cargo.
```
weihanglo, line 107: nit: Should these link to their sections respectively?

weihanglo (approved):
Great work! Thanks a lot!

Left some nits but feel free to `@bors r=Eh2406,weihanglo` (not sure if it is the correct syntax for multiple reviewers)

I am unsure about how resolver actually works. Since Jacob has approved, I am comfortable :)

src/doc/src/reference/resolver.md
epage: To me, it seems weird to title something as "Default" rather than "X, which is the default".  I think this also matches for how I see people talk about version requirements.

I didn't make the version requirement syntax docs do this because

>  I tried to balance this by not putting too much formal / technical documentation in more guide-level descriptions. In particular, while "Specifying Dependencies" is in the reference, its also written in somewhat of a guide-style.

src/doc/src/reference/resolver.md
```diff
@@ -39,19 +39,6 @@ with leading zeros. For example, `0.1.0` and `0.1.2` are compatible, but
 `0.1.0` and `0.2.0` are not. Similarly, `0.0.1` and `0.0.2` are not
 compatible.
 
-As a quick refresher, the
```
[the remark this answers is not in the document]
epage: The question is where to put it.  I would normally put it at the beginning of a reference but:

> I tried to balance this by not putting too much formal / technical documentation in more guide-level descriptions. In particular, while "Specifying Dependencies" is in the reference, its also written in somewhat of a guide-style.

We could have a summary section at the end that also includes the recommendations.

src/doc/src/reference/manifest.md
epage: Do you mean to their respective sections in the semver spec?

weihanglo: Inside this chapter just for convenience.

Linking to the official https://semver.org/ could/should happen in those sections.

epage: Which parts of this chapter are you referring to?  I thought this was that.

weihanglo: Never mind I found this is in `manifest.md`, not in `specifying-dependencies.md`. Sorry for the noise.

src/doc/src/reference/resolver.md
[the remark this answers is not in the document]
weihanglo: Either works for me.

Let's not block this PR on this, as the original one was in `resolver.md`, and I don't expect average users to read that chapter. Users above average don't need that either.

If someone requests, we can add it back to somewhere.

PR #14663, opened 2024-10-10
Improve resolver speed

x-hgg-x:

This PR improves the resolver speed after investigations from https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6.

Commit 1 adds a test showing a slow case in the resolver, where resolving time is cubic over the number of versions of a crate. It can be used for benchmarking the resolver.

Comparison of the resolving time with various values of `N=LAST_CRATE_VERSION_COUNT` and `C=TRANSITIVE_CRATES_COUNT`:

|      | N=100 | N=200 | N=400 |
|------|-------|-------|-------|
| C=1  | 0.25s | 0.5s  | 1.4s  |
| C=2  | 7s    | 44s   | 314s  |
| C=3  | 12s   | 77s   | 537s  |
| C=6  | 30s   | 149s  | 1107s |
| C=12 | 99s   | 447s  | 2393s |

Commit 2 replaces the default hasher with the hasher from `rustc-hash`, decreasing resolving time by more than 50%.

Performance comparison with the test from commit 1 by setting `LAST_CRATE_VERSION_COUNT = 100`:
| commit          | duration |
|-----------------|----------|
| master          | 16s      |
| with rustc-hash | 6.7s     |

Firefox profiles, can be read with https://profiler.firefox.com:
[perf.tar.gz](https://github.com/user-attachments/files/17318243/perf.tar.gz)

r? Eh2406

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -342,6 +342,7 @@ dependencies = [
  "rand",
  "regex",
  "rusqlite",
+ "rustc-hash 2.0.0",
  "rustfix",
  "same-file",
  "semver",
@@ -3007,6 +3008,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "08d43f7aa6b08d49f382cde6a7982047c3426db949b1424bc4b7ec9ae12c6ce2"
 
 [[package]]
+name = "rustc-hash"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "583034fd73374156e66797ed8e5b0d5690409c9226b22d87cb7f19821c05d152"
+
+[[package]]
 name = "rustfix"
 version = "0.8.7"
 dependencies = [
@@ -3748,7 +3755,7 @@ dependencies = [
  "log",
  "ordered-float",
  "partial_ref",
- "rustc-hash",
+ "rustc-hash 1.1.0",
  "serde",
  "thiserror",
  "varisat-checker",
@@ -3768,7 +3775,7 @@ dependencies = [
  "anyhow",
  "log",
  "partial_ref",
- "rustc-hash",
+ "rustc-hash 1.1.0",
  "smallvec",
  "thiserror",
  "varisat-dimacs",
--- a/crates/resolver-tests/tests/resolve.rs
+++ b/crates/resolver-tests/tests/resolve.rs
@@ -4,8 +4,8 @@ use cargo::util::GlobalContext;
 
 use resolver_tests::{
     helpers::{
-        assert_contains, assert_same, dep, dep_kind, dep_loc, dep_req, loc_names, names, pkg_id,
-        pkg_loc, registry, ToPkgId,
+        assert_contains, assert_same, dep, dep_kind, dep_loc, dep_req, loc_names, names, pkg,
+        pkg_dep, pkg_dep_with, pkg_id, pkg_loc, registry, ToDep, ToPkgId,
     },
     pkg, resolve, resolve_with_global_context,
 };
@@ -935,6 +935,50 @@ fn large_conflict_cache() {
     }
     let reg = registry(input);
     let _ = resolve(root_deps, &reg);
+}
+
+#[test]
+fn resolving_slow_case_missing_feature() {
+    let mut reg = Vec::new();
+
+    const LAST_CRATE_VERSION_COUNT: usize = 50;
+
+    // increase in resolve time is at least cubic over `INTERMEDIATE_CRATES_VERSION_COUNT`.
+    // it should be `>= LAST_CRATE_VERSION_COUNT` to reproduce slowdown.
+    const INTERMEDIATE_CRATES_VERSION_COUNT: usize = LAST_CRATE_VERSION_COUNT + 5;
+
+    // should be `>= 2` to reproduce slowdown
+    const TRANSITIVE_CRATES_COUNT: usize = 3;
+
+    reg.push(pkg_dep_with(("last", "1.0.0"), vec![], &[("f", &[])]));
+    for v in 1..LAST_CRATE_VERSION_COUNT {
+        reg.push(pkg(("last", format!("1.0.{v}"))));
+    }
+
+    reg.push(pkg_dep(
+        ("dep", "1.0.0"),
+        vec![
+            dep("last"), // <-- needed to reproduce slowdown
+            dep_req("intermediate-1", "1.0.0"),
+        ],
+    ));
+
+    for n in 0..INTERMEDIATE_CRATES_VERSION_COUNT {
+        let version = format!("1.0.{n}");
+        for c in 1..TRANSITIVE_CRATES_COUNT {
+            reg.push(pkg_dep(
+                (format!("intermediate-{c}"), &version),
+                vec![dep_req(&format!("intermediate-{}", c + 1), &version)],
+            ));
+        }
+        reg.push(pkg_dep(
+            (format!("intermediate-{TRANSITIVE_CRATES_COUNT}"), &version),
+            vec![dep_req("last", "1.0.0").with(&["f"])],
+        ));
+    }
+
+    let deps = vec![dep("dep")];
+    let _ = resolve(deps, &reg);
 }
 
 #[test]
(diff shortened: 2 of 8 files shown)


src/cargo/core/resolver/conflict_cache.rs
```diff
@@ -209,12 +210,12 @@ impl ConflictCache {
         for c in con.keys() {
             self.dep_from_pid
                 .entry(*c)
-                .or_insert_with(HashSet::new)
+                .or_insert_with(FxHashSet::default)
```
Eh2406, line 213: Nit: `.or_default()`

Eh2406 (approved):
Absolutely love the switch to `rustc-hash`, [from my testing](https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6#issuecomment-2403545404) this is a huge speed up with no downside.

Still reviewing the other commits.

crates/resolver-tests/tests/resolve.rs
```diff
@@ -937,6 +937,46 @@ fn large_conflict_cache() {
     let _ = resolve(root_deps, &reg);
 }
 
+#[test]
+fn resolving_worst_case() {
+    let mut reg = Vec::new();
+
+    // increase in resolve time is at least cubic over `N`
```
Eh2406, line 944: Do we have a regression fit demonstrating that it is technically `cubic`? I'd love to see it, or help make it. If not perhaps we should be less specific year "the resolver gets rapidly slower as you increase N".

src/cargo/core/package_id.rs
```diff
@@ -10,6 +11,7 @@ use std::sync::OnceLock;
 use serde::de;
 use serde::ser;
 
+use crate::core::resolver::activation_key::ActivationKey;
```
Eh2406, line 14: If the `ActivationKey` is going to be part of a fundamental data structure like `PackageId` then it should live in the same area, not in the resolver. @epage will probably have opinions about where, but I would guess `crate::core::ActivationKey;`.

src/cargo/core/resolver/activation_key.rs
```diff
@@ -0,0 +1,73 @@
...
+
+/// The activated version of a crate is based on the name, source, and semver compatibility.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct ActivationKey {
+    inner: &'static ActivationKeyInner,
+}
+
+impl From<ActivationKeyInner> for ActivationKey {
+    fn from(inner: ActivationKeyInner) -> Self {
+        let mut cache = ACTIVATION_KEY_CACHE
```
Eh2406, line 22: Note for self: This lock/cash is not a problem for parallelism because at the moment it's only called in `PackageId::new` which is in its own lock/cash.

src/cargo/core/package_id.rs
epage: Cargo's subsystems are a bit of a mess right now and I'd love to go through and make them more strictly hierarchical.  We have other "violations" right now so I'm not too sure if changing this will make much of a differene.

I do think I'd want us to have a clearer name within `package_id.rs`

src/cargo/core/resolver/context.rs
```diff
@@ -22,56 +22,26 @@ pub struct ResolverContext {
...
 /// of decisions made to get to this state.
 /// Several structures store the `ContextAge` when it was added,
 /// to be used in `find_candidate` for backtracking.
 pub type ContextAge = usize;
 
-/// Find the activated version of a crate based on the name, source, and semver compatibility.
-/// By storing this in a hash map we ensure that there is only one
-/// semver compatible version of each crate.
-/// This all so stores the `ContextAge`.
-pub type ActivationsKey = (InternedString, SourceId, SemverCompatibility);
```
epage, line 42: For myself, I'd prefer the moving of these types to be in its own refactor commit

epage:
> Absolutely love the switch to rustc-hash, https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6#issuecomment-2403545404 this is a huge speed up with no downside.

Should we split the test and rustc-hash commits out into its own PR and get that merged while we work through nohash?

src/cargo/core/resolver/activation_key.rs
```diff
@@ -0,0 +1,73 @@
...
+impl ActivationKey {
+    /// Returns a value that implements a "no hash" hashable value.
+    /// This is possible since all `ActivationKey` are already interned in a `HashSet`.
+    pub fn no_hash(self) -> ActivationKeyNoHash {
+        ActivationKeyNoHash(std::ptr::from_ref(self.inner) as u64)
+    }
+}
+
+#[derive(Clone, Copy, Eq, PartialEq)]
+pub struct ActivationKeyNoHash(u64);
```
Eh2406, line 65: Should this be a separate type? Or can it just be the hash implementation for `ActivationKey`?

crates/resolver-tests/tests/resolve.rs
x-hgg-x: Yes, I edited the PR description.

src/cargo/core/resolver/context.rs
```diff
@@ -135,8 +107,13 @@ impl ResolverContext {
                 // versions came from a `[patch]` source.
                 if let Some((_, dep)) = parent {
                     if dep.source_id() != id.source_id() {
-                        let key = (id.name(), dep.source_id(), id.version().into());
-                        let prev = self.activations.insert(key, (summary.clone(), age));
+                        let new_id =
+                            PackageId::new(id.name(), id.version().clone(), dep.source_id());
```
Eh2406, line 111: Note for self: This call to `PackageId::new` is a hazard for parallelism. It would still be a hazard if we directly called `ActivationKey::into`. This is in the hottest part of the resolver constructing a new lock/cached object. But, this case only occurs when we are dealing with a patched source. This is not a problem for my benchmarking tool because I don't deal with patched sources. This is not a problem for cargo because it does not run in parallel.

x-hgg-x:
I splitted to `nohash` commit in #14665.

weihanglo:
While it doesn't seem like this part of code is under the risk of hash collision, there is a concern around the other part in Cargo, see <https://github.com/rust-lang/cargo/issues/13171#issuecomment-2181266802>.

I also did some exploration of blake3 as a stable hash algorithm candidate, and blake3 has become a dependency since #14137. Have we benchmarked that as well?

Eh2406:
All of these hashes are being used exclusively in memory. So the concerns about consistency between builds let alone consistency between architectures are not relevant here. Happy to see experimentation with other algorithms though.

x-hgg-x:
I rebased on master.

Cargo.lock
weihanglo, line 1: @Eh2406 I moved conversion here for us to follow easier.

> All of these hashes are being used exclusively in memory. So the concerns about consistency between builds let alone consistency between architectures are not relevant here.

Despite that, the lang team has recently decided to migrate away from weak hash algorithms. I assume that type_id is also mostly an in-memory stuff. Do we need to follow suit?

* https://github.com/rust-lang/rust/issues/10389#issuecomment-2284562874
* https://github.com/rust-lang/rust/issues/129014

weihanglo: How easy is it for an attack to fake a PackageId, and how bad is it when it really happens?

Eh2406: The problem with `type_id` is that it assumes that if the hash is equal than the object is equal. In its case, it even goes so far as to do an unsafe transmute based on that assumption.

In this PRs case we only use it for a HashMap. If two things spuriously have the same hash than they will be checked by `eq`. If an attacker could manage to get a lot of things to hash to the same thing then cargo would spend `O(n)` time finding the right one instead of `O(1) amortized`.

> How easy is it for an attacker to fake a PackageId, and how bad is it when it really happens?

PackageId's are only generated for dependencies. If a dependency is controlled by the attacker, then serious hash collisions are the least of our problems.

weihanglo: Even SipHash is not an choice from t-lang's perspective in that linked issue, which is what Cargo is using now from std. As you said, we have `eq` so should be even harder to exploit Cargo.

> If an attacker could manage to get a lot of things to hash to the same thing then cargo would spend `O(n)` time finding the right one instead of `O(1) amortized`.

This might be a problem for companies using either cargo binary or library as a service. I am not sure if we need to take this into account. Without considering this I am 👍🏾 on this PR.

PR #14669, opened 2024-10-11
docs: document official external commands

### LINKED ISSUE #14640: `cargo-fmt` and `cargo-clippy` are not documented in The Cargo Book
### Problem

I am looking for documentation on `cargo fmt`

So I came to [here](https://doc.rust-lang.org/cargo/commands/index.html) 

It is an undocumented Cargo command.

It seems like a good place to document it.

thank you

### Proposed Solution

_No response_

### Notes

_No response_

weihanglo:

This documents external commands that are distributed along with the Rust toolchain, including

* `cargo-clippy`
* `cargo-fmt`
* `cargo-miri`

The purpose of this is increasing the visibility of these tools, and redirecting people to the correct place to discuss and learn them.

Fixes #14640

* Is “External commands” the right name to put in The Cargo Book?
* Do we really want to document the nightly-only `cargo miri`?
  * I feel like it is officially distributed, so we should.


src/doc/src/SUMMARY.md
```diff
@@ -92,6 +92,10 @@
         * [cargo package](commands/cargo-package.md)
         * [cargo publish](commands/cargo-publish.md)
         * [cargo yank](commands/cargo-yank.md)
+    * [External Commands](commands/external-commands.md)
+        * [cargo clippy](commands/cargo-clippy.md)
+        * [cargo fmt](commands/cargo-fmt.md)
+        * [cargo miri](commands/cargo-miri.md)
```
epage, line 98: > Is “External commands” the right name to put in The Cargo Book?

The main question I have is whether its worth distinguishing these as external or to list them with the others.

src/doc/src/commands/cargo-miri.md
```diff
@@ -0,0 +1,20 @@
...
+
+cargo-miri --- Runs binary crates and tests in Miri
+
+## DESCRIPTION
+
+This is an external command distributed with the Rust toolchain as an optional component.
+It is not built into Cargo, and may require additional installation.
+
+For information about usage and installation,
+see <https://github.com/rust-lang/miri>.
```
epage, line 13: > Do we really want to document the nightly-only cargo miri? 

When we have unstable options, we mark them. Unsure about unstable commands.  As this is a nightly command, we should probably mark it as unstable

src/doc/src/SUMMARY.md
weihanglo: That also works, and might be better, though `cargo-fmt` is pretty hard to categorize. It doesn't involve either build or manidest

epage: Likely it should follow clippy

weihanglo: They are all under build commands. If we find a better to categorize them we can do it later.

PR #14670, opened 2024-10-11
Remove the support for `Cargo.toml` of the cargo-script

Rustin170506:

ref https://github.com/rust-lang/cargo/issues/12207#issuecomment-2358818671

I had a discussion with @epage two weeks ago, and we decided to defer this feature and remove it now. The reason is that we couldn't always resolve the symbolic link that breaks some use cases, and the use cases of `cargo Cargo.toml` may be rare.

The unit test has been removed.

### Additional information

r? @epage

### DIFF
--- a/src/bin/cargo/commands/run.rs
+++ b/src/bin/cargo/commands/run.rs
@@ -90,9 +90,7 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
 /// See also `util/toml/mod.rs`s `is_embedded`
 pub fn is_manifest_command(arg: &str) -> bool {
     let path = Path::new(arg);
-    1 < path.components().count()
-        || path.extension() == Some(OsStr::new("rs"))
-        || path.file_name() == Some(OsStr::new("Cargo.toml"))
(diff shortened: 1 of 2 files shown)


epage:
> The reason is that we couldn't always resolve the symbolic link that breaks some use cases

We can't resolve symlinks for all manifests immediately on load as some people do weird trickery with templated `Cargo.toml` files where symlinks exist to point to the generated version of them.  Resolving the symlinks could change `target-dir` and other areas.  I'm not making any statement on whether this is meant to be officially supported workflow but likely not worth investigating.

That means we'd have to resolve the symlink only for determining how to parse the manifest.  The downside to that is a symlink to a `Cargo.toml` with a `#!` in it won't reuse your package's `target-dir`.

epage:
@bors r+

Thanks!

Qix-:
It's been kind of hard to track the changes to `cargo -Zscript` - does this mean that manifests currently do not work and thus dependencies cannot be specified in cargo scripts at the moment? Can't find any updated examples that work for pulling in dependencies.

epage:
Documentation is at https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#script

This only removed support for `cargo Cargo.toml`

PR #14671, opened 2024-10-11
Add future-incompat warning against keywords in cfgs and add raw-idents

Urgau:

This PR tries to address this thread https://github.com/rust-lang/cargo/pull/14649#discussion_r1790468829 in #14649 regarding `cfg(true)`/`cfg(false)` (and keywords more generally) which are wrongly accepted[^1] as ident.

To address this, this PR does two things:
 1. it introduce a future-incompatibility warning against those (wrongly) accepted keywords as ident
 2. it add parsing for raw-idents (`r#true`) add suggest-it in the warning

This PR should be reviewed commit-by-commit.
Tests are included in preliminary commits. 

### Additional information

I added a new struct for representing `Ident`s which is rawness aware.
Which implied updating `cargo-platform` to `0.2.0` due to the API changes.

r? @epage

[^1]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ccfb9c894dbf14e791c8ae7e4798efd0

### DIFF
--- a/crates/cargo-platform/src/cfg.rs
+++ b/crates/cargo-platform/src/cfg.rs
@@ -31,6 +31,14 @@ enum Token<'a> {
     String(&'a str),
 }
 
+/// The list of keywords.
+///
+/// We should consider all the keywords, but some are conditional on
+/// the edition so for now we just consider true/false.
+///
+/// <https://doc.rust-lang.org/reference/keywords.html>
+pub(crate) const KEYWORDS: &[&str; 2] = &["true", "false"];
+
 #[derive(Clone)]
 struct Tokenizer<'a> {
     s: iter::Peekable<str::CharIndices<'a>>,
--- a/crates/cargo-platform/src/cfg.rs
+++ b/crates/cargo-platform/src/cfg.rs
@@ -1,5 +1,6 @@
 use crate::error::{ParseError, ParseErrorKind::*};
 use std::fmt;
+use std::hash::{Hash, Hasher};
 use std::iter;
 use std::str::{self, FromStr};
 
@@ -16,16 +17,28 @@ pub enum CfgExpr {
 #[derive(Eq, PartialEq, Hash, Ord, PartialOrd, Clone, Debug)]
 pub enum Cfg {
     /// A named cfg value, like `unix`.
-    Name(String),
+    Name(Ident),
     /// A key/value cfg pair, like `target_os = "linux"`.
-    KeyPair(String, String),
+    KeyPair(Ident, String),
+}
+
+/// A identifier
+#[derive(Eq, Ord, PartialOrd, Clone, Debug)]
+pub struct Ident {
+    /// The identifier
+    pub name: String,
+    /// Is this a raw ident: `r#async`
+    ///
+    /// It's mainly used for display and doesn't take
+    /// part in the hash or equality (`foo` == `r#foo`).
+    pub raw: bool,
 }
 
 #[derive(PartialEq)]
 enum Token<'a> {
     LeftParen,
     RightParen,
-    Ident(&'a str),
+    Ident(bool, &'a str),
     Comma,
     Equals,
     String(&'a str),
@@ -47,6 +60,45 @@ struct Tokenizer<'a> {
 
 struct Parser<'a> {
     t: Tokenizer<'a>,
+}
+
+impl Ident {
+    pub fn as_str(&self) -> &str {
+        &self.name
+    }
+}
+
+impl Hash for Ident {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.name.hash(state);
+    }
+}
+
+impl PartialEq<str> for Ident {
+    fn eq(&self, other: &str) -> bool {
+        self.name == other
+    }
+}
+
+impl PartialEq<&str> for Ident {
+    fn eq(&self, other: &&str) -> bool {
+        self.name == *other
+    }
+}
+
+impl PartialEq<Ident> for Ident {
+    fn eq(&self, other: &Ident) -> bool {
+        self.name == other.name
+    }
+}
+
+impl fmt::Display for Ident {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.raw {
+            f.write_str("r#")?;
+        }
+        f.write_str(&*self.name)
+    }
 }
 
 impl FromStr for Cfg {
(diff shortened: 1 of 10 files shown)


tests/testsuite/cfg.rs
```diff
@@ -550,9 +550,26 @@ fn cfg_keywords() {
 
     p.cargo("check")
         .with_stderr_data(str![[r#"
+[WARNING] [[ROOT]/foo/Cargo.toml] future-incompatibility: `cfg(async)` is deprecated as `async` is a keyword and not an identifier and should not have have been accepted in this position.
```
epage, line 641: Rustc makes this conditioned on the edition

Compare
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3e7b39ec2db740f36b54d9fd0e93eaaa
with
https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=8e0ec47ce0d8c34e642dbcdc5abf622c

We can make this dependent on the edition for `Cargo.toml` but not `.cargo/config.toml`.  We'll need to figure out how we want to handle this

epage: One solution is to make this only relevant for `true` / `false` which is what I had scoped my comments to in https://github.com/rust-lang/cargo/pull/14649#discussion_r1790468829

Urgau: I noticed that to *(and forgot to mentioned it, my bad)*. Idk how Cargo should handle edition keywords.

However I don't think we should only do it for `true`/`false`, the other ones are wrong (even if some are not only wrong in some peculiar way), it would inconsistent to allow some and warn on the others.

My stance on the PR is to consider then independently of their "edition"-ness, which could be reversed by excluding any edition keyword until we figure out how to handle them (which would only remove a few of them and keep the vast majority).

tests/testsuite/cfg.rs
```diff
@@ -543,12 +543,12 @@ fn cfg_raw_idents() {
         .build();
 
     p.cargo("check")
-        .with_status(101)
         .with_stderr_data(str![[r#"
-[ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
-
-Caused by:
-  failed to parse `any(r#fn, r#all, r#target_os = "<>")` as a cfg expression: unexpected character `#` in cfg, expected parens, a comma, an identifier, or a string
+[WARNING] [[ROOT]/foo/Cargo.toml] future-incompatibility: `cfg(r#fn)` is deprecated as `r#fn` is a keyword and not an identifier and should not have have been accepted in this position.
```
epage: Are raw identifiers like this meant to warn?

epage: nm future commit handles it

tests/testsuite/cfg.rs
epage: The conversation on this was specifically scoped to "we shouldn't break people now for `true` / `false`"

If nothing else, expanding the scope beyond that will mire this in decisions that need to be made vs a simple sign off.

tests/testsuite/cfg.rs
Urgau: No, certainly not, it's the point of raw-ident to well be ident and not keywords, it's a bug, I will fix it.

tests/testsuite/cfg.rs
Urgau: I'm not in a rush.

I think it would be better to handle all of them in one go, but if you specifically only want to handle `true`/`false` I can separate the others keywords and do a follow up PR with them.

epage: I'd prefer it if we split these conversations up.

Urgau: I've restricted the set of keywords we warn on to only true/false.

crates/cargo-platform/src/cfg.rs
```diff
@@ -31,6 +31,14 @@ enum Token<'a> {
     String(&'a str),
 }
 
+/// The list of keywords.
+///
+/// We should consider all the keywords, but some are consitional on
```
epage, line 48: `consitional` -> `conditional`

epage:
@rfcbot fcp merge

This PR adds a future-incompat warning for the use of `true` and `false` values in a `cfg` within Cargo.  This is to make room supporting RFC3695 (see #14649) which we verbally confirmed approval for (see https://github.com/rust-lang/rfcs/pull/3695#issuecomment-2371444527)

The main focus of this FCP is insta-stabilizing support for raw identifiers (e.g. `r#true`) in `cfg`s in Cargo.  The future-incompat warning (this PR) and support for `true` / `false` (#14649) is implemented in a way that mirrors rustc where `true` and `false` are considered keywords and to reference the identifers with those names you need to use raw identifiers.

ehuss:
I'm not quite sure I understand the wording of the error message. Saying `cfg(true) is deprecated` and `will become a hard error in a future` doesn't seem correct, since when https://github.com/rust-lang/rust/issues/131204 is stabilized, it will become a supported feature (and thus not deprecated, and will intentionally be supported), right?

Would it be possible to change the wording to be something more like:

> warning: the meaning of `cfg(true)` may change in the future
> Cargo used to erroneously allow `cfg(true)` and `cfg(false)`, but both forms were interpreted as false unless manually overridden with `--cfg`. In the future these will be built-in defines that will have the corresponding true/false value. It is recommended to avoid using these values until they are properly supported. See https://github.com/rust-lang/rust/issues/131204 for more information.

Urgau:
> [..] it will become a supported feature (and thus not deprecated, and will intentionally be supported), right?

Yes, `cfg(true)` and `cfg(false)` will have the right semantic after #14649 is stabilized.

Changed the wording as suggest.

epage (approved):
Thanks @Urgau !

weihanglo:
```
---- test_proc_macro stdout ----
running `/home/runner/work/cargo/cargo/target/debug/cargo test --lib -Zbuild-std -Zpublic-dependency`
thread 'test_proc_macro' panicked at tests/build-std/main.rs:391:10:

---- expected: tests/build-std/main.rs:385:27
++++ actual:   stderr
        1 + [BLOCKING] waiting for file lock on package cache
        2 + [BLOCKING] waiting for file lock on package cache
   1    3 | [COMPILING] foo v0.0.0 ([ROOT]/foo)
   2    4 | [FINISHED] `test` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
   3    5 | [RUNNING] unittests src/lib.rs (target/debug/deps/foo-[HASH])
```

Hmm… build-std seems to clobber each others after <https://github.com/rust-lang/cargo/pull/14850>. Let me check how to fix it.

weihanglo:
Got a PR up: https://github.com/rust-lang/cargo/pull/14861

weihanglo:
Re-queuing as #14861 has been merged.

PR #14674, opened 2024-10-11
Don't copy in `From<String> for InternedString`.

jacobbramley:
If we have an owned `String` already, then don't clone it just to internalise it.

Before, `InternedString::from(String::new("blahblah"))` would take ownership of the `String` but then make and leak a copy (`str.to_string().leak()`) for use in the `interned_storage()`. This change just avoids that copy, which I think was the intention behind the interface and the interning mechanism.

Several existing tests hit this implementation, so I'd expect `cargo test` to suffice for functional correctness. The implementation of `InternedString::new()` (a few lines down) is useful as a reference.

### Additional information

Whilst it looks like an optimisation, I've no particular performance objective here. This was just something I noticed whilst looking around the code.

### DIFF
--- a/src/cargo/util/interning.rs
+++ b/src/cargo/util/interning.rs
@@ -46,7 +46,14 @@ impl<'a> From<&'a String> for InternedString {
 
 impl From<String> for InternedString {
     fn from(item: String) -> Self {
-        InternedString::new(&item)
+        let mut cache = interned_storage();
+        let s = cache.get(item.as_str()).copied().unwrap_or_else(|| {
+            let s = item.leak();
+            cache.insert(s);
+            s
+        });
+
+        InternedString { inner: s }
     }
 }
 


src/cargo/util/interning.rs
```diff
@@ -46,7 +46,14 @@ impl<'a> From<&'a String> for InternedString {
...
     fn from(item: String) -> Self {
-        InternedString::new(&item)
+        let mut cache = interned_storage();
+        let s = cache.get(item.as_str()).copied().unwrap_or_else(|| {
+            let s = item.leak();
+            cache.insert(s);
+            s
+        });
+
+        InternedString { inner: s }
```
epage, line 56: Without performance data, this falls to being a refactor.  My main concern is that we're duplicating the logic for creating an InternedString.

Unsure whether thats sufficient to block this or not.  Will think more on this

weihanglo: Another way to handle this is having a `Cow` variant of `new()`, and then both `new()` and `from()` call it.

Eh2406: Uncertainty about the additional complexity is why this was not done originally.

If we want to make it part of the type system without duplicating the core logic we could do something like:

```rust
fn new_inner<S: AsRef<str>>(str: S, make_stat: impl Fn(S) -> &'static str) -> InternedString {
    let mut cache = interned_storage();
    let s = cache.get(str.as_ref()).copied().unwrap_or_else(|| {
        let s = make_stat(str);
        cache.insert(s);
        s
    });
    InternedString { inner: s }
}
```

I don't know if it's worth the additional complexity.

epage: imo cow offers us a fairly clean design and I don't expect the check will impact performance all that much

PR #14675, opened 2024-10-11
cargo test --help: clarify --tests and --benches

RalfJung:
This tries to reduce the confusion expressed in https://github.com/rust-lang/cargo/issues/10936.

I am not sure what it means to "test a target in benchmark mode", this is copied verbatim from `cargo help test`.

### DIFF
--- a/src/bin/cargo/commands/test.rs
+++ b/src/bin/cargo/commands/test.rs
@@ -41,9 +41,9 @@ pub fn cli() -> Command {
             "Test only the specified example",
             "Test all examples",
             "Test only the specified test target",
-            "Test all test targets",
+            "Test all targets that have `test = true` set",
             "Test only the specified bench target",
-            "Test all bench targets",
+            "Test all targets that have `bench = true` set",
             "Test all targets (does not include doctests)",
         )
         .arg(
--- a/src/doc/man/generated_txt/cargo-bench.txt
+++ b/src/doc/man/generated_txt/cargo-bench.txt
@@ -172,11 +172,11 @@ OPTIONS
            multiple times and supports common Unix glob patterns.
 
        --tests
-           Benchmark all targets in test mode that have the test = true
-           manifest flag set. By default this includes the library and binaries
-           built as unittests, and integration tests. Be aware that this will
-           also build any required dependencies, so the lib target may be built
-           twice (once as a unittest, and once as a dependency for binaries,
+           Benchmark all targets that have the test = true manifest flag set.
+           By default this includes the library and binaries built as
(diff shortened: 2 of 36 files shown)


src/bin/cargo/commands/test.rs
```diff
@@ -41,9 +41,9 @@ pub fn cli() -> Command {
             "Test only the specified example",
             "Test all examples",
             "Test only the specified test target",
-            "Test all test targets",
+            "Test all targets that have the `test = true` manifest flag set",
```
weihanglo, line 44: Could we make them shorter, like

```suggestion
            "Test all targets that have `test = true` set",
```

and ditto `--benches`?

(I still find it confusing though, because it doesn't really test "all" targets. Anyway this PR is already an improvement)

src/bin/cargo/commands/test.rs
```diff
@@ -41,9 +41,9 @@ pub fn cli() -> Command {
...
             "Test only the specified bench target",
-            "Test all bench targets",
+            "Test all targets in benchmark mode that have the `bench = true` manifest flag set",
```
weihanglo, line 46: > I am not sure what it means to "test a target in benchmark mode",

I don't really know, either. I think that should be removed in man pages. Basically how I understand it is collecting Cargo targets that have `bench = true` and nothing else.

RalfJung: Would be good to confirm that before adjusting the docs.
Seems like this was added in https://github.com/rust-lang/cargo/commit/f78ab2ca4b64f8160c21738629e5b362d3ab8310, so Cc @ehuss

ehuss: I think "in benchmark mode" is left over from when cargo used to change modes/profiles based on which targets were being built. When we stabilized named profiles, we also changed cargo to stop doing that and use only a single profile.

I would delete it here and in the docs.

RalfJung: Okay, I have done that. This was duplicated in a bunch of places so I hope this will work.

The term `benchmark mode` is now gone entire; `test mode` still appears in a few places -- probably better for one of you to take a look since I am lacking the necessary context here.

RalfJung:
I realized the `cargo test --help`  texts are duplicated for `cargo check --help` and a bunch of others as well, I hope I now got them all.

src/doc/man/generated_txt/cargo-bench.txt
RalfJung, line 1: Here it was also talking about "benchmarking in test mode" vs "benchmarking in bench mode" -- I removed it for consistency, or is that still relevant somehow?

PR #14676, opened 2024-10-11
docs(env): minor improvements in environment variables doc

### LINKED ISSUE #10956: CARGO_PRIMARY_PACKAGE is set for dependencies inside workspace
### Problem

The Cargo reference states:
>- `CARGO_PRIMARY_PACKAGE` — This environment variable will be set if the package being built is primary. Primary packages are the ones the user selected on the command-line, either with -p flags or the defaults based on the current directory and the default workspace members. **This environment variable will not be set when building dependencies.** This is only set when compiling the package (not when running binaries or tests).

Ref: https://doc.rust-lang.org/cargo/reference/environment-variables.html

Specifically this statement is not always true:

> This environment variable will not be set when building dependencies.

However, it appears the env var will be set when a crate in a workspace is being built, even when used as a dependency of another crate.

Without CARGO_PRIMARY_PACKAGE always being not set for dependencies, it is not possible to accurately detect if a crate is being built for use as a dependency or not.

### Steps

See example repo: https://github.com/leighmcculloch/rustlang--cargo--issue-10956

Running the Makefile in the repo produces:
```
---- build workspace, and run bbb
cargo clean
cargo build
   Compiling aaa v0.1.0 (/Users/leighmcculloch/Code/rusttest/aaa)
   Compiling bbb v0.1.0 (/Users/leighmcculloch/Code/rusttest/bbb)
    Finished dev [unoptimized + debuginfo] target(s) in 0.34s
cargo run bbb
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
     Running `target/debug/bbb bbb`
bbb: true
aaa: true
---- build bbb, and run bbb
cargo clean
cargo build -p bbb
   Compiling aaa v0.1.0 (/Users/leighmcculloch/Code/rusttest/aaa)
   Compiling bbb v0.1.0 (/Users/leighmcculloch/Code/rusttest/bbb)
    Finished dev [unoptimized + debuginfo] target(s) in 0.33s
cargo run bbb
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
     Running `target/debug/bbb bbb`
bbb: true
aaa: false
---- build workspace (again), and run bbb
cargo clean
cargo build
   Compiling aaa v0.1.0 (/Users/leighmcculloch/Code/rusttest/aaa)
   Compiling bbb v0.1.0 (/Users/leighmcculloch/Code/rusttest/bbb)
    Finished dev [unoptimized + debuginfo] target(s) in 0.34s
cargo run bbb
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
     Running `target/debug/bbb bbb`
bbb: true
aaa: false
```

Note that in the second build and run `aaa` thinks it is primary, even when it is used as a dependency.

Note that in the third build and run `aaa` knowledge of whether it is primary is unstable, because building the workspace in the final run produces a different result as the first run.

### Possible Solution(s)

_No response_

### Notes

It appears that CARGO_PRIMARY_PACKAGE is also unstable. If I cd into the `B` crate directory and build, `A` will be built as a dependency with the env var not set. If I then cd into the workspace directory and built all crates in the workspace, the `A` crate is not rebuilt, meaning the cached built artifact for `A` is inconsistent with a clean build.

### Version

```text
cargo 1.62.1 (a748cf5a3 2022-06-08)
release: 1.62.1
commit-hash: a748cf5a3e666bc2dcdf54f37adef8ef22196452
commit-date: 2022-06-08
host: aarch64-apple-darwin
libcurl: 7.79.1 (sys:0.4.51+curl-7.80.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 12.4.0 [64-bit]
```

### LINKED ISSUE #13294: There are no documentation that cargo currently does not tell build script if `cfg!(test)` or not
In #4789, an member of the repository says "This is currently intentional behavior because build scripts don't have a "I'm running tests" mode." but it looks there are no documentation about that in cargo book. 

https://github.com/rust-lang/cargo/blob/92395d90106b3b61bcb68bcf2069052c93771764/src/doc/src/reference/environment-variables.md#environment-variables-cargo-sets-for-build-scripts
https://github.com/rust-lang/cargo/blob/92395d90106b3b61bcb68bcf2069052c93771764/src/doc/src/reference/build-scripts.md

I think it's good to mention that in the cargo book.

weihanglo:

Minor improvements in environment variables doc.

Fixes #10956
Fixes #13294

### DIFF
--- a/src/doc/src/reference/environment-variables.md
+++ b/src/doc/src/reference/environment-variables.md
@@ -260,9 +260,10 @@ corresponding environment variable is set to the empty string, `""`.
 * `CARGO_PRIMARY_PACKAGE` --- This environment variable will be set if the
   package being built is primary. Primary packages are the ones the user
   selected on the command-line, either with `-p` flags or the defaults based
-  on the current directory and the default workspace members. This environment
-  variable will not be set when building dependencies. This is only set when
-  compiling the package (not when running binaries or tests).
+  on the current directory and the default workspace members.
+  This variable will not be set when building dependencies,
+  unless a dependency is also a workspace member that was also selected on the command-line.
+  This is only set when compiling the package (not when running binaries or tests).
 * `CARGO_TARGET_TMPDIR` --- Only set when building [integration test] or benchmark code.
   This is a path to a directory inside the target directory
   where integration tests or benchmarks are free to put any data needed by
--- a/src/doc/src/reference/environment-variables.md
+++ b/src/doc/src/reference/environment-variables.md
@@ -355,6 +355,9 @@ let out_dir = env::var("OUT_DIR").unwrap();
     * `CARGO_CFG_TARGET_FEATURE=mmx,sse` --- List of CPU [target features] enabled.
   > Note that different [target triples][Target Triple] have different sets of `cfg` values,
   > hence variables present in one target triple might not be available in the other.
+  >
+  > Some cfg values like `debug_assertions`, `test`, and Cargo features like
+  > `feature="foo"` are not available.
 * `OUT_DIR` --- the folder in which all output and intermediate artifacts should
   be placed. This folder is inside the build directory for the package being built,
   and it is unique for the package in question.


src/doc/src/reference/environment-variables.md
```diff
@@ -260,9 +260,10 @@ corresponding environment variable is set to the empty string, `""`.
 * `CARGO_PRIMARY_PACKAGE` --- This environment variable will be set if the
   package being built is primary. Primary packages are the ones the user
   selected on the command-line, either with `-p` flags or the defaults based
-  on the current directory and the default workspace members. This environment
-  variable will not be set when building dependencies. This is only set when
-  compiling the package (not when running binaries or tests).
+  on the current directory and the default workspace members.
+  This variable will not be set when building dependencies,
+  unless a dependency is also a workspace member.
```
ehuss, line 265: I'm concerned that this statement isn't entirely true. In many (most?) cases, it won't be set for workspace member dependencies. For example in cargo's own workspace, it won't be set for any of its dependencies when doing `cargo run`.

I think to make this more accurate, perhaps change it to something like this?

```suggestion
  unless a dependency is also a workspace member that was also selected on the command-line.
```

However, as noted in the issue, what is "selected on the command-line" is sometimes not intuitive.

ehuss: I just want to highlight that there is a bug in how cargo cache's this behavior, which is described in #14678. I don't think that needs to change anything here, but indicates that the answer to when it is set is "buggy".

weihanglo: Thank you. That is more accurate. Updated!

I think one of the problems here is that the paragraph wasn't well arranged. There is no paragraph or newline there. We may really need to give each variable a heading.

ehuss:
Thanks!

@bors r+

Issue #8179, opened 2020-04-29
Proc macros can cause non-json output in stdout despite message-format=json


roblabla:
**Problem**

When proc macros print to stdout, the output is reflected in the cargo output. For instance, the following proc macro will cause `Hello` to be printed in the compiler output, even when `message-format=json` is passed, even though it's not a valid JSON object. This will cause various tools (such as `cargo-metadata`, `rust-analyzer`, `jq`...) to fail to parse the message, and give up.

```rust
extern crate proc_macro;
use proc_macro::TokenStream;

#[proc_macro_attribute]
pub fn hello(_attr: TokenStream, item: TokenStream) -> TokenStream {
    println!("Hello");
    item
}
```

I'm not sure if this is considered a cargo or a rustc bug ^^'.

**Steps**

1. Install `jq`
2. `git clone https://github.com/roblabla/cargo-proc-macro-print-bug`
3. `cargo build --message-format=json | jq .`

This will cause `jq` to error with "parse error: Invalid numeric literal at line 3, column 0", which is caused by the string "Hello", printed by the proc macro, to appear in `stdout`.

**Possible Solution(s)**
I imagine the proc macro output could be:

- Ignored/silenced
- Wrapped in a `CompilerMessage` JSON object
- Wrapped in a new kind of JSON object, maybe a `GenericMessage`?

**Notes**

cargo 1.42.0 (86334295e 2020-01-31)
rustc 1.42.0 (b8cedc004 2020-03-09)

ehuss:
Tools should only interpret a line as JSON if it starts with `{`.  I thought that was documented, but it is not (as you know, we are working on the docs on the cargo_metadata side in https://github.com/oli-obk/cargo_metadata/pull/106).

Otherwise, there is nothing that can be done for now. Proc-macros run unrestricted within the rustc process, and can do whatever they want.  There are experiments like [watt](https://github.com/dtolnay/watt) for running proc-macros in an isolated environment, but I think that is very very far off, and may not be flexible enough in some cases.

roblabla:
> Tools should only interpret a line as JSON if it starts with {. I thought that was documented, but it is not.

In fact, documentation [explicitly says otherwise](https://doc.rust-lang.org/cargo/reference/external-tools.html#json-messages):

> When passing --message-format=json, Cargo will output the following information during the build:
> - compiler errors and warnings,
> - produced artifacts,
> - results of the build scripts (for example, native dependencies).
> 
[...]

PR #14677, opened 2024-10-11
docs: tools should only interpret a line starting with `{` as JSON

weihanglo:

Minor doc improvement that helps #8179

### DIFF
--- a/src/doc/src/reference/external-tools.md
+++ b/src/doc/src/reference/external-tools.md
@@ -46,6 +46,12 @@ The `package_id` field is a unique identifier for referring to the package, and
 as the `--package` argument to many commands. The syntax grammar can be found in
 chapter [Package ID Specifications].
 
+> **Note:** `--message-format=json` only controls Cargo and Rustc's output.
+> This cannot control the output of other tools,
+> e.g. `cargo run --message-format=json`,
+> or arbitrary output from procedural macros.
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/external-tools.md
```diff
@@ -46,6 +46,10 @@ The `package_id` field is a unique identifier for referring to the package, and
 as the `--package` argument to many commands. The syntax grammar can be found in
 chapter [Package ID Specifications].
 
+> **Note:** Not all output is guaranteed to be in JSON format with the `--message-format=json` option.
+> For example, output from procedural macros and logs from external tools may appear in plain text.
+> Tools should only interpret a line as JSON if it starts with `{`.
+
```
epage, line 52: I think I'd prefer we don't bless mixed output as much as this makes it sound like it.

Maybe instead
```suggestion
> **Note:** `--message-format=json` only controls Cargo and Rustc's output.  This cannot control the output of other tools, e.g. `cargo run --message-format=json`.  A possible workaround in these situations is to only interpret a line as JSON if it starts with `{`.

```

weihanglo: Applied your suggestion with a mention of proc macros. I think it is still worth since proc macros are not seen as “other tools” and people may not realize immediately.

epage: Yeah, they are a nother of those "outside of the control of cargo/rustc" things that are worth calling out

PR #14680, opened 2024-10-12
fix(registry): make sure to process a download after it has completed

x-hgg-x:

This PR solves the bug found in https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6#issuecomment-2406197867.

With this PR we can avoid refetching the same dependency multiple times because a finished download was not processed before returning from the `block_until_ready()` method. It also improves the resolving speed since `activate_deps_loop()` is called less times overall.

### DIFF
--- a/src/cargo/sources/registry/http_remote.rs
+++ b/src/cargo/sources/registry/http_remote.rs
@@ -796,15 +796,15 @@ impl<'gctx> RegistryData for HttpRegistry<'gctx> {
         self.downloads.blocking_calls += 1;
 
         loop {
-            self.handle_completed_downloads()?;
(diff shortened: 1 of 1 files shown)


Eh2406:
I've never had a deep grasp of the semantics of the curl crate. So I'm not going to take lead on reviewing this PR.

arlosi:
@x-hgg-x thanks for reporting and creating this fix!

I did some additional investigation and added additional testing in this area in a new PR #14694.

x-hgg-x:
Closed in favor of https://github.com/rust-lang/cargo/pull/14694.

Issue #11203, opened 2022-10-09
Allow having broken symlinks in the packaged cargo archive


Ekleog:
### Problem

In https://github.com/Ekleog/kannader/tree/main/smtp-queue-fs, I have a `res/` folder that intentionally contains a broken symlink. This folder is properly tracked by git, and used by tests to validate that `smtp-queue-fs` behaves correctly when faced with a queue that has a broken symlink

### Proposed Solution

It would be nice if `cargo` were able to store the same types of files as `git`, so including broken symlinks. This would make it possible to upload this kind of crates with its tests on crates.io

### Notes

I guess an alternative would be to not have the broken symlink be checked-in, and instead handle the test creation and validation differently. However, it makes it harder to just add tests, given that it means manual care needs to be given to each test that has a broken symlink.

I would understand if you think that'd be a non-feature and you didn't want to allow including broken symlinks in the crate on crates.io, but I thought it was worth opening this issue to gather feedback over whether it'd be of interest. Sorry if it has already been discussed elsewhere!

weihanglo:
Thank you for the proposal! 

I totally understand what you're trying to solve. Also have no idea why Cargo implemented like this, as the logic dates back to [Cargo v0.0.1-pre]😆. However, I do have concerns on this being implemented. 

Cargo tries its best to make every `.crate` file always compile. Based on that fact, Cargo resolves all symlinks, and copies target contents over into tarball. A common use case is sharing root README and LICENSE file within a workspace, so that each member gets a real copy of that file and published to crates.io with a sense of integrity. 

If Cargo now switches the behaviour to align with git, some projects will start failing as they cannot find the actual contents during compiling. If we keep the current behaviour and only treat broken symlinks as you proposed, the inconsistency sounds even more confusing, at least to me.

We may loose the restriction a bit, such like when Cargo.toml `includes` explicitly includes a broken symlink, just includes it and emits a warning. I am not entirely sure what it might cause if we go this route. People might find their files missing after publish and that is really unfortunate.

[Cargo v0.0.1-pre]: https://github.com/rust-lang/cargo/blob/v0.0.1-pre/src/cargo/ops/cargo_package.rs#L130

Ekleog:
Hmmm… as you're saying, I guess silently changing the behavior or reusing `include` wouldn't make much sense.

[...]

PR #14684, opened 2024-10-14
docs: More information on what is and isn't included by cargo package

sourcefrog:

Add more detail to `cargo package` explaining what is and isn't included. 

This doesn't add the feature requested by #11203 (shipping symlinks) but it does at least make it more clear that it's not supported.

Docs only change.

I wrote this by reading [src/cargo/sources/path.rs](https://github.com/rust-lang/cargo/blob/82c489f1c612096c2dffc48a4091d21af4b8e046/src/cargo/sources/path.rs), in addition to my observations of cargo behavior, so it might be helpful for reviewers to check the new docs against the code.

### DIFF
--- a/src/doc/man/cargo-package.md
+++ b/src/doc/man/cargo-package.md
@@ -32,6 +32,16 @@ stored in the `target/package` directory. This performs the following steps:
     - A `.cargo_vcs_info.json` file is included that contains information
       about the current VCS checkout hash if available, as well as a flag if the
       worktree is dirty.
+    - Symlinks are flattened to their target files.
+    - Subdirectories containing `Cargo.toml` are excluded, as they are assumed
+      to be separate packages.
+    - A `target` directory in the root of the package is excluded.
+    - Files and directories matching the `[include]` and `[exclude]` fields in
+      the manifest are included or excluded. Include patterns can override the
+      following rules.
+    - Files that are ignored by VCS are excluded.
+    - Unix hidden files and directories (starting with `.`) are excluded.
+
 3. Extract the `.crate` file and build it to verify it can build.
     - This will rebuild your package from scratch to ensure that it can be
       built from a pristine state. The `--no-verify` flag can be used to skip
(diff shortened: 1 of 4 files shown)


sourcefrog:
cc? @weihanglo

sourcefrog:
The failures seem to be unrelated flakes?

sourcefrog:
https://github.com/rust-lang/cargo/actions/runs/11330170805/job/31507413375?pr=14684

Perhaps this is getting rate-limited based on the GitHub workers IP pool, if the cargo build does not currently pass any credentials? If so, just switching to an authenticated pull, even without paying, might help. 

It also might be possible to cache the image in GitHub Actions, per https://docs.docker.com/build/ci/github-actions/cache/. That would help stay below the limit of authenticated pulls. Even if we're using just the GitHub IPs it would probably reduce the error rate.

We could split out a separate issue? I can have a look if you like, although as I'm not a project member I might need help with registering an account or uploading secrets.

src/doc/man/cargo-package.md
```diff
@@ -32,6 +32,16 @@ stored in the `target/package` directory. This performs the following steps:
...
       worktree is dirty.
+    - Symlinks are flattened to their target files.
+    - Subdirectories containing `Cargo.toml` are excluded, as they are assumed
+      to be separate packages.
+    - A `target` directory in the root of the package is excluded.
+    - Files and directories matching the `[include]` and `[exclude]` fields in
+      the manifest are included or excluded. Include patterns can override the
+      following rules.
+    - Files that are ignored by VCS are excluded.
+    - Unix hidden files and directories (starting with `.`) are excluded.
```
weihanglo, line 43: Thank you. I love your writing!

Yet, I just recalled that we already have the rule written down in <https://doc.rust-lang.org/nightly/cargo/reference/manifest.html#the-exclude-and-include-fields>.

Instead of duplicating the writing, what about providing a relative link to `../reference/manifest.html#the-exclude-and-include-fields`? Something like:

```suggestion
    - Files and directories are included or excluded based on rules mentioned in
      [the `[include]` and `[exclude]` fields](../reference/manifest.html#the-exclude-and-include-fields).
```

weihanglo (approved):

weihanglo:
Thank you!

@bors r+

PR #14682, opened 2024-10-14
docs: install options -> uninstall options

DriesDeMaeseneer:
### What does this PR try to resolve?

Fixed a typo in the cargo-uninstall.md documentation file.

PR #14686, opened 2024-10-14
chore: switch to ECR Public from Docker Hub

weihanglo:

We're hitting this hard.

```
[internal] load metadata for docker.io/library/httpd:2.4-alpine
ERROR: failed to copy: httpReadSeeker:
failed open: unexpected status code https://registry-1.docker.io/v2/library/httpd/manifests/sha256:66c49302c02430619abb84240a438bcfc083015661009fcaaeaac931450f62cd:
  429 Too Many Requests - Server message: toomanyrequests:
You have reached your pull rate limit.
You may increase the limit by authenticating and upgrading:
https://www.docker.com/increase-rate-limit
```

And the rate limit seems quite easy to hit.

> The rate limits of 100 container image requests per six hours for
> anonymous usage, > and 200 container image requests per six hours
> for free Docker accounts are now in effect. Image requests exceeding
> these limits will be denied until the six hour window elapses.

This PR switches to Amazon ECR public, and alternative image registry,
to unblock us from the rate limit of the official Docker registry.

Both of the two images are verified and uploaded by Docker

* https://gallery.ecr.aws/docker/library/alpine
* https://gallery.ecr.aws/docker/library/httpd

See also https://gallery.ecr.aws/

### DIFF
--- a/crates/cargo-test-support/containers/apache/Dockerfile
+++ b/crates/cargo-test-support/containers/apache/Dockerfile
@@ -1,4 +1,4 @@
-FROM httpd:2.4-alpine
+FROM public.ecr.aws/docker/library/httpd:2.4-alpine
 
 RUN apk add --no-cache git git-daemon openssl
 
--- a/crates/cargo-test-support/containers/sshd/Dockerfile
+++ b/crates/cargo-test-support/containers/sshd/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.20
+FROM public.ecr.aws/docker/library/alpine:3.20
 
 RUN apk add --no-cache openssh git
 RUN ssh-keygen -A


sourcefrog:
Just a thought: arguably we should switch to Azure's registry instead, since GitHub workers probably run on Azure? (I'm not totally sure they have public mirrors of this but it seems like they would?)

sourcefrog:
.. In a bit of searching I can't find public mirrors like this on Azure. 

Within GH itself, there is https://github.com/linuxcontainers/alpine/pkgs/container/alpine but I don't think this includes Apache and I'm not sure it should be equally trusted.

weihanglo:
Seems like we have recovered from the rate limit: <https://github.com/rust-lang/cargo/pull/14682#issuecomment-2412539891>. I am fine with either closing or merging this.

weihanglo:
> .. In a bit of searching I can't find public mirrors like this on Azure.

Neither can I find Azure equivalent. And nothing in <https://github.com/orgs/apache/packages> nor <https://github.com/orgs/alpinelinux/packages> 😞.

ehuss:
I opened https://github.com/rust-lang/infra-team/issues/176 to figure out what we should do here.

weihanglo:
Going to close in favor of the discussion <https://github.com/rust-lang/infra-team/issues/176>.

PR #14690, opened 2024-10-15
fix(resolver): avoid cloning when iterating using RcVecIter

x-hgg-x:

Follow up of https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6#issuecomment-2408905423.

This PR modifies the internal `RcVecIter` struct so that we can iterate on its items without cloning them. This improves performance of the resolver because eagerly cloning `Arc<T>` is not free.

Comparison of performance using `solana-core = "=1.0.5"` in `Cargo.toml`:

| branch     | duration |
|------------|----------|
| master     | 213s     |
| PR         | 202s     |

r? Eh2406

### DIFF
--- a/src/cargo/core/resolver/types.rs
+++ b/src/cargo/core/resolver/types.rs
@@ -5,7 +5,6 @@ use crate::util::interning::InternedString;
 use crate::util::GlobalContext;
 use std::cmp::Ordering;
 use std::collections::{BTreeMap, BTreeSet};
-use std::ops::Range;
 use std::rc::Rc;
 use std::time::{Duration, Instant};
 
@@ -181,13 +180,13 @@ impl DepsFrame {
     fn min_candidates(&self) -> usize {
         self.remaining_siblings
             .peek()
-            .map(|(_, (_, candidates, _))| candidates.len())
+            .map(|(_, candidates, _)| candidates.len())
             .unwrap_or(0)
     }
 
-    pub fn flatten(&self) -> impl Iterator<Item = (PackageId, Dependency)> + '_ {
+    pub fn flatten(&self) -> impl Iterator<Item = (PackageId, &Dependency)> + '_ {
         self.remaining_siblings
-            .clone()
+            .remaining()
             .map(move |(d, _, _)| (self.parent.package_id(), d))
     }
 }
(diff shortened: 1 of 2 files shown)


src/cargo/core/resolver/types.rs
```diff
@@ -251,15 +250,16 @@ impl RemainingDeps {
             // Figure out what our next dependency to activate is, and if nothing is
             // listed then we're entirely done with this frame (yay!) and we can
             // move on to the next frame.
-            if let Some(sibling) = deps_frame.remaining_siblings.next() {
+            let sibling = deps_frame.remaining_siblings.iter_by_ref().next().cloned();
+            if let Some(sibling) = sibling {
                 let parent = Summary::clone(&deps_frame.parent);
-                self.data.insert((deps_frame, insertion_time));
+                self.data.insert((deps_frame.clone(), insertion_time));
```
Eh2406, line 256: If were hyper counting clones, I don't think this one's needed.

src/cargo/core/resolver/types.rs
```diff
@@ -324,22 +324,24 @@ pub type ConflictMap = BTreeMap<PackageId, ConflictReason>;
...
-        self.rest
-            .clone()
-            .next()
-            .and_then(|i| self.vec.get(i).map(|val| (i, &*val)))
+    pub fn remaining(&self) -> impl Iterator<Item = &T> + '_ {
+        self.vec.get(self.offset..).into_iter().flatten()
+    }
+
+    pub fn iter_by_ref(&mut self) -> impl Iterator<Item = &T> + '_ {
+        (self.vec.get(self.offset..).into_iter().flatten()).inspect(|_| self.offset += 1)
```
Eh2406, line 344: We are relying on some spooky action keeping the iteration counter inside the range `self.offset..` to the same value as the `self.offset`. Put a different way, using `inspect` for its side effects makes my skin crawl.

The best I've got for an alternative way to spell it is:
```rust
        std::iter::from_fn(|| {
            let out = self.vec.get(self.offset);
            if out.is_some() {
                self.offset += 1;
            }
            out
        })
```

Although that may end up being slower because of the loss of `ExactSizeIterator`.

If we can't find a better way of spelling it, we should at least put a comment within the `inspect` explain why it's critical for correctness.

src/cargo/core/resolver/types.rs
```diff
@@ -324,22 +324,24 @@ pub type ConflictMap = BTreeMap<PackageId, ConflictReason>;
...
-    fn peek(&self) -> Option<(usize, &T)> {
-        self.rest
-            .clone()
-            .next()
-            .and_then(|i| self.vec.get(i).map(|val| (i, &*val)))
+    pub fn remaining(&self) -> impl Iterator<Item = &T> + '_ {
+        self.vec.get(self.offset..).into_iter().flatten()
+    }
+
+    pub fn iter_by_ref(&mut self) -> impl Iterator<Item = &T> + '_ {
```
Eh2406, line 343: I think we can just call this `iter`, as most plane `iter` yield `&`.

Eh2406 (approved):
Overall, a good improvement. I'm surprised it makes such a difference. Thank you for finding it.

PR #14691, opened 2024-10-15
fix(perf): switch global allocator to mimalloc

x-hgg-x:

Follow-up of https://github.com/Eh2406/pubgrub-crates-benchmark/issues/6#issuecomment-2408905423.

This PR uses `mimalloc` as the global allocator for `cargo`. This improves performance of the resolver by reducing allocation and deallocation cost. The performance gains are the same as with `jemalloc`, but `mimalloc` is compatible with Windows. 

Comparison of performance using `solana-core = "=1.0.5"` in `Cargo.toml`:

| allocator  | duration |
|------------|----------|
| system     | 66.3s    |
| jemalloc   | 57.7s    |
| mimalloc   | 58s      |

r? Eh2406

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,7 @@ libc = "0.2.155"
 libgit2-sys = "0.17.0"
 libloading = "0.8.5"
 memchr = "2.7.4"
+mimalloc = "0.1.43"
 miow = "0.6.0"
 opener = "0.7.1"
 openssl = "=0.10.57" # See rust-lang/cargo#13546 and openssl/openssl#23376 for pinning
@@ -181,6 +182,7 @@ jobserver.workspace = true
 lazycell.workspace = true
 libgit2-sys.workspace = true
 memchr.workspace = true
+mimalloc = { workspace = true, optional = true }
 opener.workspace = true
 os_info.workspace = true
 pasetors.workspace = true
@@ -260,6 +262,7 @@ test = false
 doc = false
 
 [features]
+mimalloc = ["dep:mimalloc"]
 vendored-openssl = ["openssl/vendored"]
 vendored-libgit2 = ["libgit2-sys/vendored"]
 # This is primarily used by rust-lang/rust distributing cargo the executable.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -334,6 +334,7 @@ dependencies = [
  "libc",
  "libgit2-sys",
  "memchr",
+ "mimalloc",
  "opener",
  "openssl",
  "os_info",
@@ -2246,6 +2247,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4ec2a862134d2a7d32d7983ddcdd1c4923530833c9f2ea1a44fc5fa473989058"
 
 [[package]]
+name = "libmimalloc-sys"
+version = "0.1.39"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "23aa6811d3bd4deb8a84dde645f943476d13b248d818edcf8ce0b2f37f036b44"
+dependencies = [
+ "cc",
+ "libc",
+]
+
+[[package]]
 name = "libnghttp2-sys"
 version = "0.1.10+1.61.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2360,6 +2371,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fe751422e4a8caa417e13c3ea66452215d7d63e19e604f4980461212f3ae1322"
 dependencies = [
  "libc",
+]
+
+[[package]]
+name = "mimalloc"
+version = "0.1.43"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "68914350ae34959d83f732418d51e2427a794055d0b9529f48259ac07af65633"
+dependencies = [
+ "libmimalloc-sys",
 ]
 
 [[package]]
--- a/src/bin/cargo/main.rs
+++ b/src/bin/cargo/main.rs
@@ -1,5 +1,9 @@
 #![allow(clippy::self_named_module_files)] // false positive in `commands/build.rs`
 
+#[cfg(feature = "mimalloc")]
+#[global_allocator]
+static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
+
 use cargo::core::features;
 use cargo::core::shell::Shell;
 use cargo::util::network::http::http_handle;


Cargo.toml
```diff
@@ -243,6 +244,9 @@ features = [
   "Win32_System_Threading",
 ]
 
+[target.'cfg(not(target_env = "msvc"))'.dependencies]
```
x-hgg-x, line 247: Should we restrict this to only linux and macos ?

Cargo.toml
weihanglo, line 1: This reminds me that since Cargo is no longer a workspace member in rust-lang/rust, we may have missed some optimization they've done in the past. We may also want to check what we should pick up again in Cargo's `[profile]`.

weihanglo, line 1: BTW, rust-lang/rust uses jemalloc for sure. See <https://github.com/rust-lang/rust/blob/00367d523e3dbc7bc7c7a5f5772c2928453fdbc6/compiler/rustc/src/main.rs#L4-L36>.

Eh2406:
I don't know have context about how different parts of rust releases are supposed to link to `jemallocator`. @ehuss how should we link and who do we need to talk to?

@bors r? ehuss

Cargo.toml
weihanglo: Maybe? Most other sibling projects test only on macOS and Linux for it.

* https://github.com/rust-lang/rust/blob/9322d183f45e0fd5a509820874cc5ff27744a479/config.example.toml?plain=1#L723-L725
* https://github.com/rust-lang/rust/blob/9322d183f45e0fd5a509820874cc5ff27744a479/src/tools/miri/Cargo.toml?plain=1#L31-L36

But rust-analyzer doesn't.

* https://github.com/rust-lang/rust/blob/9322d183f45e0fd5a509820874cc5ff27744a479/src/tools/rust-analyzer/crates/rust-analyzer/Cargo.toml?plain=1#L77-L78
* https://github.com/rust-lang/rust-analyzer/blob/418c1365eccf20c9261b6948a6e637f789224af9/crates/rust-analyzer/src/bin/main.rs#L31-L33

x-hgg-x: CI failed for `Windows x86_64 gnu`, so we may want to disable it for Windows.

Cargo.toml
weihanglo: Some random concerns I've found:

* https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Why.20is.20rustc.20on.20Alpine.20so.20much.20slower.20and.20can.20it.20be.20fixed.3F/near/475549598
* https://rust-lang.zulipchat.com/#narrow/stream/197181-t-libs.2Fwg-allocators/topic/mimalloc.20as.20default.20allocator/near/474810148

weihanglo:
Opened a Zulip topic for this: https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Replace.20system.20allocator.20with.20jemalloc.2Fmimalloc

x-hgg-x:
Comparison on Linux between `system`, `jemalloc` and `mimalloc` allocators with the last improvements on master, using `solana-core = "=1.0.5"` in `Cargo.toml`:
[profiles.tar.gz](https://github.com/user-attachments/files/17395197/profiles.tar.gz)

| allocator  | duration |
|------------|----------|
| system     | 66.3s    |
| jemalloc   | 57.7s    |
| mimalloc   | 58s      |

Output of `perf diff` between `system` and `jemalloc`:
```
# Baseline  Delta Abs  Shared Object  Symbol                                                                                                                                                                                                                                                                                                                               
    10.33%     -3.56%  cargo          [.] cargo::core::resolver::RemainingCandidates::next
     3.28%     -3.28%  libc.so.6      [.] malloc
     3.10%     -3.10%  libc.so.6      [.] _int_malloc
     2.37%     -2.37%  libc.so.6      [.] _int_free
     5.60%     +1.84%  libc.so.6      [.] __memmove_avx_unaligned_erms
     2.67%     +1.80%  cargo          [.] core::hash::Hash::hash_slice
     3.10%     +1.20%  cargo          [.] alloc::rc::Rc<T,A>::make_mut
     7.71%     +1.06%  cargo          [.] <cargo::core::dependency::Dependency as core::hash::Hash>::hash
               +0.99%  cargo          [.] _rjem_sdallocx
     0.80%     -0.79%  libc.so.6      [.] _int_free_merge_chunk
     4.45%     +0.63%  cargo          [.] <alloc::rc::Rc<T,A> as core::ops::drop::Drop>::drop
               +0.57%  cargo          [.] _rjem_malloc
     3.59%     +0.47%  cargo          [.] <core::hash::sip::Hasher<S> as core::hash::Hasher>::write
```

Output of `perf diff` between `system` and `mimalloc`:
```
# Baseline  Delta Abs  Shared Object  Symbol                                                                                                                                                                                                                                                                                                                               
    10.33%     -3.76%  cargo          [.] cargo::core::resolver::RemainingCandidates::next
     3.28%     -3.27%  libc.so.6      [.] malloc
     3.10%     -3.10%  libc.so.6      [.] _int_malloc
     2.37%     -2.37%  libc.so.6      [.] _int_free
     2.67%     +1.84%  cargo          [.] core::hash::Hash::hash_slice
     5.60%     +1.64%  libc.so.6      [.] __memmove_avx_unaligned_erms
               +1.05%  cargo          [.] mi_free
     4.45%     +1.03%  cargo          [.] <alloc::rc::Rc<T,A> as core::ops::drop::Drop>::drop
     0.98%     -0.98%  libc.so.6      [.] unlink_chunk.isra.0
     3.10%     +0.98%  cargo          [.] alloc::rc::Rc<T,A>::make_mut
     7.71%     +0.72%  cargo          [.] <cargo::core::dependency::Dependency as core::hash::Hash>::hash
     6.21%     +0.66%  cargo          [.] <bitmaps::bitmap::Iter<Size> as core::iter::traits::iterator::Iterator>::next
               +0.62%  cargo          [.] _mi_malloc_generic
     0.59%     -0.59%  libc.so.6      [.] cfree@GLIBC_2.2.5
```

The main gain seems to be the memory access to the interned `PackageId` in the resolver hot loop.
Since `jemalloc` and `mimalloc` have about the same performance profiles, we can enable `mimalloc` for Windows and use either one for the other platforms.

epage:
> Since jemalloc and mimalloc have about the same performance profiles, we can enable mimalloc for Windows and use either one for the other platforms.

If there is little performance gain to specialize allocators for a given platform, I would lean towards keeping things simple and using a single allocator across all of them.

cuviper:
What version of glibc was used for the system tests? 

Also, could this be set as a feature so third parties like distros can more easily opt out? I think for integration, I'd rather keep using the system allocator.

weihanglo:
> Also, could this be set as a feature so third parties like distros can more easily opt out? I think for integration, I'd rather keep using the system allocator.

If we're going to do it, we will, just like openssl and libgit2.

Cargo.toml
x-hgg-x: Solved by switching to `mimalloc` instead.

Cargo.toml
```diff
@@ -260,6 +262,8 @@ test = false
 doc = false
 
 [features]
+default = ["mimalloc"]
```
weihanglo, line 265: Custom allocator should not be default. Opt-in is a better option to respect people's choices.

PR #14694, opened 2024-10-15
fix(registry): `HttpRegistry` `block_until_ready` returns early when work is still pending

arlosi:

`block_until_ready` is returning early when there are still pending transfers. This leads to extra restarts of the resolver, impacting performance.

In the existing implementation:
- `handle_completed_downloads` runs and there are no completed downloads
- `multi.perform()` is called and completes new downloads, with no pending transfers remaining
- Since there are no items `remaining_in_multi`, the function returns early

This fixes the issue by reordering the calls to `multi.perform()`, `handle_completed_downloads`, then the completion check.

A test is added that uses cargo's tracing to show the number of blocking calls. First commit shows existing behavior, second commit shows fix. 

Originally based on the PR #14680 by @x-hgg-x. The ordering fix in this PR also avoids an additional `block_until_ready` call for retried failed downloads.

### DIFF
--- a/src/cargo/sources/registry/http_remote.rs
+++ b/src/cargo/sources/registry/http_remote.rs
@@ -789,8 +789,8 @@ impl<'gctx> RegistryData for HttpRegistry<'gctx> {
     }
 
     fn block_until_ready(&mut self) -> CargoResult<()> {
-        trace!(target: "network",
-            "block_until_ready: {} transfers pending",
+        trace!(target: "network::HttpRegistry::block_until_ready",
+            "{} transfers pending",
             self.downloads.pending.len()
         );
         self.downloads.blocking_calls += 1;
--- a/src/cargo/sources/registry/http_remote.rs
+++ b/src/cargo/sources/registry/http_remote.rs
@@ -796,19 +796,22 @@ impl<'gctx> RegistryData for HttpRegistry<'gctx> {
         self.downloads.blocking_calls += 1;
 
         loop {
-            self.handle_completed_downloads()?;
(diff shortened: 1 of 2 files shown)


Eh2406:
does the code https://github.com/rust-lang/cargo/blob/8040c0051dadf95bee8c78eca89c339a8a10df3a/src/cargo/core/package.rs#L960  also need adjustment?

weihanglo:
> does the code
> 
> https://github.com/rust-lang/cargo/blob/8040c0051dadf95bee8c78eca89c339a8a10df3a/src/cargo/core/package.rs#L960
> 
> also need adjustment?

Seem pretty likely, though this is downloading the package so the perf gain may not be as significant as registry index one.

weihanglo (approved):
👍🏾 
Good either with or without <https://github.com/rust-lang/cargo/pull/14694#issuecomment-2414882510>.

arlosi:
> does the code
> 
> https://github.com/rust-lang/cargo/blob/8040c0051dadf95bee8c78eca89c339a8a10df3a/src/cargo/core/package.rs#L960
> also need adjustment?

The current implementation looks correct to me. It doesn't have the same bug with early returning when there is still work to be done.

My notes on how it's working:
```
loop {
- wait()
  - wait_for_curl() 
     loop {
     - add_sleepers() [self.sleeping -> self.pending (may not move all, depending on timing)]
     - multi.perform() [self.pending -> self.results]
     }
  - self.pending.remove()
  - self.sleeping.push() [for failed downloads that need retry]
}
```

Issue #14154, opened 2024-06-26
Changing workspace Cargo.toml doesn't invalidate build cache for workspace members


illicitonion:
### Problem

[The environment variables Cargo sets when building crates](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) don't appear to get updated if the crate's Cargo.toml inherits values from the workspace, and the only change between builds is the value in the workspace's Cargo.toml.

### Steps

```console
% cat >Cargo.toml <<EOF
[workspace]
members = ["lib"]
package.version = "0.2.0"
package.readme = "beep.md"
EOF
% mkdir -p lib/src
% cat >lib/Cargo.toml <<EOF
[package]
name = "lib"
version = "0.1.0"
edition = "2021"
readme = { workspace = true }
EOF
% cat >lib/src/main.rs <<EOF
fn main() {
    println!("Readme: {}", env!("CARGO_PKG_README"));
}
EOF
% cargo run -p lib 2>/dev/null
Readme: ../beep.md
% cat >Cargo.toml <<EOF
[workspace]
members = ["lib"]
package.version = "0.2.0"
package.readme = true
EOF
% # Next command should print ../README.md because the workspace value changed:
% cargo run -p lib 2>/dev/null
Readme: ../beep.md
% # Clearing the cache forces a rebuild, which gives the correct value:
% rm -rf target
% cargo run -p lib 2>/dev/null
Readme: ../README.md
```

### Version

```text
cargo 1.79.0 (ffa9cf99a 2024-06-03)
release: 1.79.0
commit-hash: ffa9cf99a594e59032757403d4c780b46dc2c43a
commit-date: 2024-06-03
host: aarch64-apple-darwin
libgit2: 1.7.2 (sys:0.18.3 vendored)
libcurl: 8.1.2 (sys:0.4.72+curl-8.6.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Mac OS 13.5.2 [64-bit]
```

epage:
Just reproduced this without workspace inheritance involved.

epage:
Not seeing an existing issue for cargo-generated env variables but this seems similar in internals to #13280 and #10358.

weihanglo:
[...]

PR #14701, opened 2024-10-18
Fix: trace `config` `[env]` table in dep-info.

### LINKED ISSUE #13280: Cargo fails to detect environment variable is newly set and rebuild if the environment variable is set in `config.toml`

linyihai:

The `env` defined in `config.toml` or `--config env`  are stripped before recording in the dep-info file.  The absence of the env table doesn't rebuild when env changed 

By tracing the `env` table into the `dep-info` file, If the envs changed then the rebuild can be triggered. 

Fixes #13280

One commit add the test, the latter commit update the test through the fixes.

### Additional information
The PR only fixed the `env` table changes can't trigger a rebuild, other `CARGO-like` envs 
 doesn't take into account.

### DIFF
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -362,8 +362,8 @@
 mod dirty_reason;
 
 use std::collections::hash_map::{Entry, HashMap};
-
 use std::env;
+use std::ffi::OsString;
 use std::fmt::{self, Display};
 use std::fs::{self, File};
 use std::hash::{self, Hash, Hasher};
@@ -849,7 +849,11 @@ impl LocalFingerprint {
                                 .to_string(),
                         )
                     } else {
-                        gctx.get_env(key).ok()
+                        if let Some(value) = gctx.env_config()?.get(key) {
+                            value.to_str().and_then(|s| Some(s.to_string()))
+                        } else {
+                            gctx.get_env(key).ok()
+                        }
                     };
                     if current == *previous {
                         continue;
@@ -2124,6 +2128,9 @@ enum DepInfoPathType {
 ///
 /// The serialized Cargo format will contain a list of files, all of which are
 /// relative if they're under `root`. or absolute if they're elsewhere.
+///
+/// The `env_config` argument is a set of environment variables that are
+/// defined in `[env]` table of the `config.toml`.
 pub fn translate_dep_info(
     rustc_dep_info: &Path,
     cargo_dep_info: &Path,
@@ -2132,6 +2139,7 @@ pub fn translate_dep_info(
     target_root: &Path,
     rustc_cmd: &ProcessBuilder,
     allow_package: bool,
+    env_config: &Arc<HashMap<String, OsString>>,
 ) -> CargoResult<()> {
     let depinfo = parse_rustc_dep_info(rustc_dep_info)?;
 
@@ -2168,9 +2176,11 @@ pub fn translate_dep_info(
     // This also includes `CARGO` since if the code is explicitly wanting to
     // know that path, it should be rebuilt if it changes. The CARGO path is
     // not tracked elsewhere in the fingerprint.
-    on_disk_info
-        .env
-        .retain(|(key, _)| !rustc_cmd.get_envs().contains_key(key) || key == CARGO_ENV);
+    //
+    // For cargo#13280, We trace env vars that are defined in the `[env]` config table.
+    on_disk_info.env.retain(|(key, _)| {
+        env_config.contains_key(key) || !rustc_cmd.get_envs().contains_key(key) || key == CARGO_ENV
+    });
 
     let serialize_path = |file| {
         // The path may be absolute or relative, canonical or not. Make sure
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1987,10 +1987,7 @@ pub(crate) fn apply_env_config(
         if cmd.get_envs().contains_key(key) {
             continue;
         }
-
-        if value.is_force() || gctx.get_env_os(key).is_none() {
-            cmd.env(key, value.resolve(gctx));
-        }
+        cmd.env(key, value);
     }
     Ok(())
 }
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -330,7 +330,7 @@ fn rustc(
     if hide_diagnostics_for_scrape_unit {
         output_options.show_diagnostics = false;
     }
-
+    let env_config = Arc::clone(build_runner.bcx.gctx.env_config()?);
     return Ok(Work::new(move |state| {
         // Artifacts are in a different location than typical units,
         // hence we must assure the crate- and target-dependent
@@ -459,6 +459,7 @@ fn rustc(
                 &rustc,
                 // Do not track source files in the fingerprint for registry dependencies.
                 is_local,
+                &env_config,
             )
             .with_context(|| {
                 internal(format!(
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -63,7 +63,7 @@ use std::io::SeekFrom;
 use std::mem;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
-use std::sync::Once;
+use std::sync::{Arc, Once};
 use std::time::Instant;
 
 use self::ConfigValue as CV;
@@ -227,7 +227,7 @@ pub struct GlobalContext {
     target_cfgs: LazyCell<Vec<(String, TargetCfgConfig)>>,
     doc_extern_map: LazyCell<RustdocExternMap>,
     progress_config: ProgressConfig,
-    env_config: LazyCell<EnvConfig>,
+    env_config: LazyCell<Arc<HashMap<String, OsString>>>,
     /// This should be false if:
     /// - this is an artifact of the rustc distribution process for "stable" or for "beta"
     /// - this is an `#[test]` that does not opt in with `enable_nightly_features`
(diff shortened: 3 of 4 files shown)


src/cargo/core/compiler/mod.rs
```diff
@@ -330,7 +330,13 @@ fn rustc(
...
         output_options.show_diagnostics = false;
     }
-
+    let config_envs: HashSet<String> = build_runner
+        .bcx
+        .gctx
+        .env_config()?
+        .keys()
+        .cloned()
+        .collect();
```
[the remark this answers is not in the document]
weihanglo: Another way is we could have a set of internal env + a set of env from `cargo::rustc-env` build script invocation. dep_info should first remove those and anything left is what we're interested in (including `[env]`).

<details><summary>See the PoC</summary>
<p>

* Can avoid cloning `cargo::rustc-env` by acquiring mutex again, though this is just a PoC so didn't do that.
* Should also need to filter out `CARGO_BIN_<exe>`
* Need an `debug_assertion` to check exhaustiveness of `internal_envs` in rustc invocation.

```diff
diff --git a/src/cargo/core/compiler/fingerprint/mod.rs b/src/cargo/core/compiler/fingerprint/mod.rs
index e6fff90932a..03489d2c99a 100644
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -361,8 +361,10 @@
 
 mod dirty_reason;
 
+use std::cell::LazyCell;
 use std::collections::hash_map::{Entry, HashMap};
 
+use std::collections::HashSet;
 use std::env;
 use std::fmt::{self, Display};
 use std::fs::{self, File};
@@ -2130,7 +2132,7 @@ pub fn translate_dep_info(
     rustc_cwd: &Path,
     pkg_root: &Path,
     target_root: &Path,
-    rustc_cmd: &ProcessBuilder,
+    envs_from_build_script_output: &HashSet<String>,
     allow_package: bool,
 ) -> CargoResult<()> {
     let depinfo = parse_rustc_dep_info(rustc_dep_info)?;
@@ -2140,6 +2142,33 @@ pub fn translate_dep_info(
     let mut on_disk_info = EncodedDepInfo::default();
     on_disk_info.env = depinfo.env;
 
+    let internal_envs = LazyCell::new(|| HashSet::from([
+        "CARGO",
+        "CARGO_MANIFEST_DIR",
+        "CARGO_MANIFEST_PATH",
+        "CARGO_PKG_VERSION",
+        "CARGO_PKG_VERSION_MAJOR",
+        "CARGO_PKG_VERSION_MINOR",
+        "CARGO_PKG_VERSION_PATCH",
+        "CARGO_PKG_VERSION_PRE",
+        "CARGO_PKG_AUTHORS",
+        "CARGO_PKG_NAME",
+        "CARGO_PKG_DESCRIPTION",
+        "CARGO_PKG_HOMEPAGE",
+        "CARGO_PKG_REPOSITORY",
+        "CARGO_PKG_LICENSE",
+        "CARGO_PKG_LICENSE_FILE",
+        "CARGO_PKG_RUST_VERSION",
+        "CARGO_PKG_README",
+        "CARGO_CRATE_NAME",
+        "CARGO_BIN_NAME",
+        "OUT_DIR",
+        "CARGO_PRIMARY_PACKAGE",
+        "CARGO_TARGET_TMPDIR",
+        "CARGO_RUSTC_CURRENT_DIR",
+        paths::dylib_path_envvar(),
+    ]));
+
     // This is a bit of a tricky statement, but here we're *removing* the
     // dependency on environment variables that were defined specifically for
     // the command itself. Environment variables returned by `get_envs` includes
@@ -2170,7 +2199,10 @@ pub fn translate_dep_info(
     // not tracked elsewhere in the fingerprint.
     on_disk_info
         .env
-        .retain(|(key, _)| !rustc_cmd.get_envs().contains_key(key) || key == CARGO_ENV);
+        .retain(|(key, _)| 
+            !internal_envs.contains(key.as_str()) && !envs_from_build_script_output.contains(key)
+            || key == CARGO_ENV
+        );
 
     let serialize_path = |file| {
         // The path may be absolute or relative, canonical or not. Make sure
diff --git a/src/cargo/core/compiler/mod.rs b/src/cargo/core/compiler/mod.rs
index ee5fdbbecd6..2f3e547c762 100644
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -337,8 +337,11 @@ fn rustc(
         // directory is present.
         if artifact.is_true() {
             paths::create_dir_all(&root)?;
+
         }
 
+        let mut envs_from_build_script_output = HashSet::new();
+
         // Only at runtime have we discovered what the extra -L and -l
         // arguments are for native libraries, so we process those here. We
         // also need to be sure to add any -L paths for our plugins to the
@@ -359,7 +362,7 @@ fn rustc(
                 )?;
                 add_plugin_deps(&mut rustc, &script_outputs, &build_scripts, &root_output)?;
             }
-            add_custom_flags(&mut rustc, &script_outputs, script_metadata)?;
+            add_custom_flags(&mut rustc, &script_outputs, script_metadata, &mut envs_from_build_script_output)?;
         }
 
         for output in outputs.iter() {
@@ -456,7 +459,7 @@ fn rustc(
                 &cwd,
                 &pkg_root,
                 &target_dir,
-                &rustc,
+                &envs_from_build_script_output,
                 // Do not track source files in the fingerprint for registry dependencies.
                 is_local,
             )
@@ -880,6 +883,7 @@ fn rustdoc(build_runner: &mut BuildRunner<'_, '_>, unit: &Unit) -> CargoResult<W
             &mut rustdoc,
             &build_script_outputs.lock().unwrap(),
             script_metadata,
+            &mut HashSet::new(),
         )?;
 
         // Add the output of scraped examples to the rustdoc command.
@@ -1525,6 +1529,7 @@ fn add_custom_flags(
     cmd: &mut ProcessBuilder,
     build_script_outputs: &BuildScriptOutputs,
     metadata: Option<Metadata>,
+    envs: &mut HashSet<String>,
 ) -> CargoResult<()> {
     if let Some(metadata) = metadata {
         if let Some(output) = build_script_outputs.get(metadata) {
@@ -1536,6 +1541,7 @@ fn add_custom_flags(
             }
             for (name, value) in output.env.iter() {
                 cmd.env(name, value);
+                envs.insert(name.to_owned());
             }
         }
     }
```

</p>
</details>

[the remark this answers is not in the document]
weihanglo: I believe the original purpose of excluding Cargo internal environment variables from Cargo's dep-info is just they are already tracked by Cargo, so not necessary. (Though we know the statement is not entirely true according to #14154.)

If we choose to use a patch similar to the one I shared in <https://github.com/rust-lang/cargo/pull/14701#discussion_r1807208627>, it might become a headache for syncing between the hash set and env vars actually applied to rustc invocations. That's why I propose to check the exhaustiveness in `debug_assertion`. That will help at least when we run cargo's test suite.

> Does it means that check all environment variables in internal_var are present in rust innovations?

We'll need to do the other way round — check every env var in `rustc_cmd.envs()` is an element in internal env var hash set. However, as you've noticed, dynamic ones are hard to track, and may need some more refactors. Yet I think here we focus on environment variables set for crates, not set for running build scripts. Environment variables for artifact-dependency do not seem to belong to this category?

---

At this stage, I am fine with `Arc` on `gctx.env_config` to reduce the clone cost, if the second approach looks too brittle and need more explorations.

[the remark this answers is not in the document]
linyihai: Thanks for you explanations. The second approach seems require more preparation and prior knowledge for me. 

Besides, I had apply the Arc on `gctx.env_config`.

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -2168,9 +2172,11 @@ pub fn translate_dep_info(
     // This also includes `CARGO` since if the code is explicitly wanting to
     // know that path, it should be rebuilt if it changes. The CARGO path is
     // not tracked elsewhere in the fingerprint.
-    on_disk_info
-        .env
-        .retain(|(key, _)| !rustc_cmd.get_envs().contains_key(key) || key == CARGO_ENV);
+    //
+    // For issue#13280, We trace env vars that are defined in the `config.toml`.
+    on_disk_info.env.retain(|(key, _)| {
+        !rustc_cmd.get_envs().contains_key(key) || key == CARGO_ENV || config_envs.contains(key)
```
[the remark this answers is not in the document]
linyihai: One clumsy solution is to filter these `internal_envs` out of `[env]` table 
```
 let internal_envs = LazyCell::new(|| HashSet::from([
        "CARGO",
        "CARGO_MANIFEST_DIR",
        "CARGO_MANIFEST_PATH",
        "CARGO_PKG_VERSION",
        "CARGO_PKG_VERSION_MAJOR",
        "CARGO_PKG_VERSION_MINOR",
        "CARGO_PKG_VERSION_PATCH",
        "CARGO_PKG_VERSION_PRE",
        "CARGO_PKG_AUTHORS",
        "CARGO_PKG_NAME",
        "CARGO_PKG_DESCRIPTION",
        "CARGO_PKG_HOMEPAGE",
        "CARGO_PKG_REPOSITORY",
        "CARGO_PKG_LICENSE",
        "CARGO_PKG_LICENSE_FILE",
        "CARGO_PKG_RUST_VERSION",
        "CARGO_PKG_README",
        "CARGO_CRATE_NAME",
        "CARGO_BIN_NAME",
        "OUT_DIR",
        "CARGO_PRIMARY_PACKAGE",
        "CARGO_TARGET_TMPDIR",
        "CARGO_RUSTC_CURRENT_DIR",
        paths::dylib_path_envvar(),
    ]));

```

src/cargo/core/compiler/mod.rs
```diff
@@ -330,7 +330,7 @@ fn rustc(
     if hide_diagnostics_for_scrape_unit {
         output_options.show_diagnostics = false;
     }
-
+    let config_envs = build_runner.bcx.gctx.env_config()?.clone();
```
weihanglo, line 333: Explicit `Arc::clone` is preferred.

```suggestion
    let config_envs = Arc::clone(&build_runner.bcx.gctx.env_config()?);
```

weihanglo (commented):
Thanks. Only one tiny suggestion.

tests/testsuite/cargo_env_config.rs
```diff
@@ -276,3 +276,94 @@ MAIN ENV_TEST:from-env
...
+        .run();
+
+    p.cargo("run")
+        .env("ENV_TEST", "from-env")
+        .with_stdout_data(str![[r#"
+from-env
+
+"#]])
+        .with_stderr_data(str![[r#"
+[COMPILING] foo v0.5.0 ([ROOT]/foo)
```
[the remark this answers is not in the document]
linyihai: https://github.com/rust-lang/cargo/blob/6f92aaa31f5404edb2ce401b4461b3b3de06804f/src/cargo/core/compiler/fingerprint/mod.rs?plain=1#L851-L853

Thank your for pointing out these.  Here shoudld also apply [ the contions](https://github.com/rust-lang/cargo/blob/6f92aaa31f5404edb2ce401b4461b3b3de06804f/src/cargo/core/compiler/mod.rs?plain=1#L1991) to figure out which env change should trigger a rebuild.

src/cargo/util/context/mod.rs
```diff
@@ -1833,34 +1833,41 @@ impl GlobalContext {
...
+                    if env_config.contains_key(*disallowed) {
+                        bail!(
+                            "setting the `{disallowed}` environment variable is not supported \
+                            in the `[env]` configuration table"
+                        );
+                    }
+                }
+                env_config
+                    .into_iter()
+                    .filter(|(k, v)| v.is_force() || self.get_env_os(k).is_none())
```
weihanglo, line 1867: Is it possible that it calls `value.resolve(gctx)` so other caller don't need to resolve on their own?

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -2168,9 +2179,11 @@ pub fn translate_dep_info(
     // This also includes `CARGO` since if the code is explicitly wanting to
     // know that path, it should be rebuilt if it changes. The CARGO path is
     // not tracked elsewhere in the fingerprint.
-    on_disk_info
-        .env
-        .retain(|(key, _)| !rustc_cmd.get_envs().contains_key(key) || key == CARGO_ENV);
+    //
+    // For issue#13280, We trace env vars that are defined in the `config.toml`.
```
weihanglo, line 2183: nit:
```suggestion
    // For cargo#13280, We trace env vars that are defined in the `[env]` config table.
```

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -2124,6 +2131,9 @@ enum DepInfoPathType {
 ///
 /// The serialized Cargo format will contain a list of files, all of which are
 /// relative if they're under `root`. or absolute if they're elsewhere.
+///
+/// The `config_envs` argument is a set of environment variables that are
```
weihanglo, line 2135: I found it hard to text search this variable because the naming is not consistent. Do you have any specific reason calling it `config_envs` not `env_config`?

tests/testsuite/cargo_env_config.rs
```diff
@@ -276,3 +276,166 @@ MAIN ENV_TEST:from-env
...
+
+"#]])
+        .with_stderr_data(str![[r#"
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+[RUNNING] `target/debug/foo[EXE]`
+
+"#]])
+        .run();
+
+    p.cargo("run")
```
weihanglo, line 329: This invocation is identical to the previous one. I believe this is for checking that there is no rebuild. It is not obvious without any comment. Should we add one for this and `env_changed_defined_in_config_args`?

```suggestion
    // This identical cargo invocation is to ensure no rebuild happen.
    p.cargo("run")
```

tests/testsuite/cargo_env_config.rs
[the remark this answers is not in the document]
weihanglo: Thanks for updating the PR so swiftly!

src/cargo/util/context/mod.rs
linyihai: I split a single commit from the previous commit for the purpose of fixing to apply all the changes to this function because this part looks like refactoring the code

src/cargo/core/compiler/fingerprint/mod.rs
linyihai: I like `env_config` also.

src/cargo/util/context/mod.rs
weihanglo: Makes sense. Thank you for doing that! The refactor may have a downside: the resolved paths may be long and consume lots of memory. Splitting it into a separate helps us discover it when needed.

(I may move `Arc` change to be along with the next fix commit, though).

weihanglo (approved):

weihanglo:
@bors r+

Thanks for working on this along with us!

Note that this fix has a caveat <https://github.com/rust-lang/cargo/pull/14701#discussion_r1807184527>. If a user sets cargo internal env like `CARGO_MAINFEST_DIR` in `[env]` it will track that env var in Cargo's dep info with the "correct" value, not the value specified in `[env]`. This is not a feature we advertise and considered as a bug, though it could potentially be used as a workaround for #14154.

PR #14707, opened 2024-10-19
test: add fixes in the sat resolver

x-hgg-x:

This is a follow-up of https://github.com/rust-lang/cargo/pull/14614.

Commit 1 removes duplicate variables in the sat resolver.
Commit 2 removes useless clones in the sat resolver.

r? Eh2406

### DIFF
--- a/crates/resolver-tests/src/sat.rs
+++ b/crates/resolver-tests/src/sat.rs
@@ -1,3 +1,4 @@
+use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::fmt::Write;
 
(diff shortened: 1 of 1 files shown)


crates/resolver-tests/src/sat.rs
```diff
@@ -86,16 +87,17 @@ fn create_dependencies_vars<'a>(
...
+        let dep_feature_vars = var_for_is_dependencies_features_used
             .entry(name)
             .or_default()
-            .insert((kind, platform), dep_feature_var_map);
+            .entry((kind, platform))
+            .or_default();
+
+        for &feature_name in dep.features() {
+            if let Entry::Vacant(entry) = dep_feature_vars.entry(feature_name) {
+                entry.insert(solver.new_var());
```
Eh2406, line 98: Why the change?  Is this just to avoid  allocating  several `new_var`'s? Do we really end up with dependencies that  activate the same feature more than once?

Also would this be clearer as `.entry(feature_name).or_insert_with(|| solver.new_var())`?

x-hgg-x: > Why the change? Is this just to avoid allocating several `new_var`'s?

Yes
 
> Also would this be clearer as `.entry(feature_name).or_insert_with(|| solver.new_var())`?

I didn't use `or_insert_with` because I don't need the returned `Entry`.

Eh2406: Okay, that makes sense.

> Do we really end up with dependencies that activate the same feature more than once?

For testing purposes I added a `else {todo!()}` to each of these, and all tests pass. So I don't think our test suite hits this case. What am I missing? Do you have plans that will add test code that hits this case? Are there significant downsides to a few calls to `solver.new_var()`? I'm not objecting, I'm just trying to understand.

x-hgg-x: The `else {todo!()}` will trigger if we have a dependency with a list of features containing duplicates, or if the feature `"dep/feature"` or `"dep?/feature"` is listed multiple times among all features declared in a summary. I will add a test for this.

x-hgg-x: > Are there significant downsides to a few calls to solver.new_var()?

Maybe not, but removing duplicates will improve the debugging if we want to manually inspect the clauses.

Eh2406:
No need for a test.

@bors r+

Issue #11406, opened 2022-11-22
Add the ability to replace any source in .cargo/config.toml  file


jfrog-qa:
### Problem

When deploying a crate with a registry dependency, the client tries to resolve the dependency from the specified URL.
In JFrog we want to support cargo sparse index for a virtual repository as well - which means that we are resolving files without really knowing their source.
The problem is that the client will fetch the dependencies from the real source, instead of the registry in the config.toml file.
In order to resolve this issue we need the option to tell the cargo client to replace any source (maybe [source.any]), and not only specific sources like [source.crates-io]

### Proposed Solution

Provide another known source that will be relevant for any source. or an option to direct all cargo requests to a specific registry (including URLs of dependencies)

weihanglo:
Just try to scratch the surface of syntax. Probably `[source."*"]` is a choice? Cargo already has the config key `[profile.dev.package."*"]` in [profile overrides](https://doc.rust-lang.org/nightly/cargo/reference/profiles.html#overrides).

weihanglo:
`[source."*"]` may not be the best choice, since it's already a valid source name today. Even `.cargo/config.toml` of every package in the world were migrated to the new semantic of `[source."*"]`, old Cargo still doesn't understand the new meaning and understands it in the old way, which may lead to a complete different source replacement. It looks like a compatibility hazard.

For instance, this config setting also valid and works perfectly fine with `cargo vendor`.

```
[source.'🦀']
directory = "vendor"

[source.crates-io]
replace-with = "🦀"
```

 I don't have an alternative at this moment 😞.

Dr-Emann:
I believe `virtual repository` is an [Artifactory concept](https://stackoverflow.com/a/76734613), which provides one repository which combines multiple repositories: e.g. one repository which delegates to either a cache of crates.io, or a private internal repository (or multiple repositories).

Eh2406:
Registries that mix crates from multiple sources are going to hit this problem. The current workaround is to use source replacement (specifically at published time) so that all crates claim that their dependencies come from crates.io. This recommendation is despite the documentation on source replacement. (Sorry for that.)

The fact that cargo keeps track of what source crates come from and does not silently interpose the same name from different sources is an important protection against "dependency confusion" attacks. The fact that it's going haywire here is because a virtual repository is "dependency confusion" as a service.

Dr-Emann:
[...]

PR #14713, opened 2024-10-22
 fix source replacement for deps

zhangweikop:
**one line** change to set the package source to be the actual registry when source replacement is used.

Example of this issue is demonstrated as:
 https://github.com/zhangweikop/cargo-dep-source-replace-issue
#### Details:

1.  The current dep resolve fail with *"conflict error"* when the resolve loop encounters the following items:
> a. packageX (that contains native link) whose registry has been explicitly specified as the source Y 
> b. the same package X whose source has been replaced by sourceY.

Such case are not uncommon in company with multiple teams where various teams may use:
For crate1:  its Cargo.toml has `packageX = { version = "1.2.3", registry = "SourceY" }`
For crate2: its Cargo.toml has `packageX = { version = "1.2.33" }` (with source replacement  "SourceY")
Also, packageX have some (transitive) dependency which has native links.
Now if a developer creates crate3 which depends on crate1&crate2. It would encounter failure.

Example of error is shown in the background. 
2. This new change will let the resolve process treat "same" dep package same and get rid of this "native link" conflict error.

3. This change should only affect the resolve process.
     The existing source download function has handled the source replacement properly.
     This change doesn't affect behavior of how the crate is packaged and published. 

4. ([wip] A bunch of tests output still need to be updated) 

#### Background:
Our company uses a private "registry" called "external" for the "crates-io". 
And when develop crate, we add the source replacement like the following to ".cargo/config.toml "
```
[source.crates-io]
replace-with = 'external'
```
It normally works well. However, it sometimes complains as:
```
failed to select a version for `wasm-bindgen-shared`.
    ... required by package `wasm-bindgen-macro-support v0.2.40`
    ... which satisfies dependency `wasm-bindgen-macro-support = "=0.2.40"` of package `wasm-bindgen-macro v0.2.40`
    ... which satisfies dependency `wasm-bindgen-macro = "=0.2.40"` of package `wasm-bindgen v0.2.40`
    ... which satisfies dependency `wasm-bindgen = "^0.2.23"` of package `js-sys v0.3.0`
    ... which satisfies dependency `js-sys = "^0.3"` of package `chrono v0.4.20`
    ... which satisfies dependency `chrono_0_4 = "^0.4.20"` of package `serde_with v3.3.0`
..
versions that meet the requirements `=0.2.40` are: 0.2.40

the package `wasm-bindgen-shared` links to the native library `wasm_bindgen`, but it conflicts with a previous package which links to `wasm_bindgen` as well:
package `wasm-bindgen-shared v0.2.95 (registry `external`)`
    ... which satisfies dependency `wasm-bindgen-shared = "=0.2.95"` (locked to 0.2.95) of package `wasm-bindgen-macro-support v0.2.95 (registry `external`)`
    ... which satisfies dependency `wasm-bindgen-macro-support = "=0.2.95"` (locked to 0.2.95) of package `wasm-bindgen-macro v0.2.95 (registry `external`)`
    ... which satisfies dependency `wasm-bindgen-macro = "=0.2.95"` (locked to 0.2.95) of package `wasm-bindgen v0.2.95 (registry `external`)`
    ... which satisfies dependency `wasm-bindgen = "^0.2.89"` (locked to 0.2.95) of package `reqwest v0.12.8 (registry `external`)`
    ... ..
```
It's sort of annoying because we don't even use any Wasm in our crates at all. The dependency system perform some check the , if two **different** [package ids link to](https://github.com/rust-lang/cargo/blob/master/src/cargo/core/resolver/mod.rs#L768) same native library, then resolve fails. But actually, all the packages are downloaded from the same source which is "external".
When I was trying to figure out what lead to: `Package(wasm-bindgen-shared, source="default crate-io") ! = Package(wasm-bindgen-shared, source="external")` . 
I notice "possible mistaken usage" of  `.map_source` in `sources/replaced.rs`.  
From the meaning, the dependency info from source which may be empty (treated as crates-io) should be explicitly set as the active source? 

Test building my crate using this new cargo fix. And it works.
 
I picked a few test and update it. It seems the new test output makes more sense?
 For example :`testsuite/alt_registry.rs` 

Considering this replace block in [test setup](https://github.com/rust-lang/cargo/blob/master/crates/cargo-test-support/src/registry.rs#L443C25-L443C56): the output should not show `github.com/rust-lang/crates.io-index`.

### Additional information
Would like to hear more opinions about this PR. 
If the fix makes sense, I would go ahead to fix other unit tests.

### DIFF
--- a/src/cargo/sources/replaced.rs
+++ b/src/cargo/sources/replaced.rs
@@ -73,7 +73,7 @@ impl<'gctx> Source for ReplacedSource<'gctx> {
 
         self.inner
             .query(&dep, kind, &mut |summary| {
-                f(summary.map_summary(|s| s.map_source(replace_with, to_replace)))
+                f(summary.map_summary(|s| s.map_source(to_replace, replace_with)))
             })
             .map_err(|e| {
                 if self.is_builtin_replacement() {
--- a/tests/testsuite/alt_registry.rs
+++ b/tests/testsuite/alt_registry.rs
@@ -439,7 +439,7 @@ fn alt_registry_and_crates_io_deps() {
 [DOWNLOADING] crates ...
 [DOWNLOADED] crates_io_dep v0.0.1 (registry `dummy-registry`)
 [DOWNLOADED] alt_reg_dep v0.1.0 (registry `alternative`)
-[CHECKING] crates_io_dep v0.0.1
+[CHECKING] crates_io_dep v0.0.1 (registry `dummy-registry`)
 [CHECKING] alt_reg_dep v0.1.0 (registry `alternative`)
 [CHECKING] foo v0.0.1 ([ROOT]/foo)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
@@ -1105,7 +1105,7 @@ fn alt_reg_metadata() {
       "edition": "2015",
       "features": {},
       "homepage": null,
-      "id": "registry+https://github.com/rust-lang/crates.io-index#bar@0.0.1",
+      "id": "registry+[ROOTURL]/registry#bar@0.0.1",
       "keywords": [],
       "license": null,
       "license_file": null,
@@ -1117,7 +1117,7 @@ fn alt_reg_metadata() {
       "readme": null,
       "repository": null,
       "rust_version": null,
-      "source": "registry+https://github.com/rust-lang/crates.io-index",
+      "source": "registry+[ROOTURL]/registry",
       "targets": [
         {
           "crate_types": [
@@ -1225,7 +1225,7 @@ fn alt_reg_metadata() {
       "edition": "2015",
       "features": {},
       "homepage": null,
-      "id": "registry+https://github.com/rust-lang/crates.io-index#iodep@0.0.1",
+      "id": "registry+[ROOTURL]/registry#iodep@0.0.1",
       "keywords": [],
       "license": null,
       "license_file": null,
@@ -1237,7 +1237,7 @@ fn alt_reg_metadata() {
       "readme": null,
       "repository": null,
       "rust_version": null,
-      "source": "registry+https://github.com/rust-lang/crates.io-index",
+      "source": "registry+[ROOTURL]/registry",
       "targets": [
         {
           "crate_types": [
@@ -1261,7 +1261,7 @@ fn alt_reg_metadata() {
     "nodes": [
       {
         "dependencies": [
-          "registry+https://github.com/rust-lang/crates.io-index#bar@0.0.1"
+          "registry+[ROOTURL]/registry#bar@0.0.1"
         ],
         "deps": [
           {
@@ -1272,7 +1272,7 @@ fn alt_reg_metadata() {
               }
             ],
             "name": "bar",
-            "pkg": "registry+https://github.com/rust-lang/crates.io-index#bar@0.0.1"
+            "pkg": "registry+[ROOTURL]/registry#bar@0.0.1"
           }
         ],
         "features": [],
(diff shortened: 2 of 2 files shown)


src/cargo/sources/replaced.rs
weihanglo, line 1: You can run `SNAPSHOTS=overwrite cargo t --test testsuite` to get a sense of how it changes tests. Don't bother manually editing it or read through the CI logs.

weihanglo: I really appreciate your writing in the PR description! To make us understand the issue more, is it possible for you to create a minimal reproducible example so that we can look into and brainstorm solutions?

zhangweikop: > I really appreciate your writing in the PR description! To make us understand the issue more, is it possible for you to create a minimal reproducible example so that we can look into and brainstorm solutions?

Yeah, I would do that. (I have been using a minimal example when testing with our company's private registry)
Do you whether there is some open source `crates-io` alternative registry to help demonstrate the behavior?

weihanglo: Some third-party registries can be found at <https://github.com/rust-lang/cargo/wiki/Third-party-registries>. They may have bugs or haven't catch up to some new features in Cargo / registry specs.

(I wish we could have an (official) reference implementation of it, or a testsuite asserting the registry behavior. None of them has yet happened 😞)

zhangweikop: I created this example for demonstration.

https://github.com/zhangweikop/cargo-dep-source-replace-issue

weihanglo: This is what I've got from the reproduction with `Kellnr`. Not sure what has happened with Kellnr mirror but apparently it was not what your original issue.

```console
    Updating `test-mirror` index
error: failed to select a version for the requirement `syn = "^1.0.67"`
candidate versions found which didn't match: 0.12.10, 0.12.9, 0.12.7, ...
location searched: `test-mirror` index
required by package `wasm-bindgen-macro-support v0.2.81 (registry `test-mirror`)`
    ... which satisfies dependency `wasm-bindgen-macro-support = "=0.2.81"` of package `wasm-bindgen-macro v0.2.81 (registry `test-mirror`)`
    ... which satisfies dependency `wasm-bindgen-macro = "=0.2.81"` of package `wasm-bindgen v0.2.81 (registry `test-mirror`)`
    ... which satisfies dependency `wasm-bindgen = "^0.2.81"` of package `iana-time-zone v0.1.41 (registry `test-mirror`)`
    ... which satisfies dependency `iana-time-zone = "^0.1.41"` of package `chrono v0.4.21 (registry `test-mirror`)`
    ... which satisfies dependency `chrono = "^0.4.21"` of package `deps-conflict v0.1.0 (home/user/projects/cargo-dep-source-replace-issue/deps-conflict)
```

weihanglo: Regardless, by just reading the error message you encountered, I feel like that is the expected behavior. 

One of the design decisions made into Cargo is that it takes `source` into account for package uniqueness, to avoid dependency confusion attack [^1]. A `registry` key is in fact a custom registry you host not a mirror, whereas a source replaced one is expected to be as if it were the original one, since [source replacement mostly for mirroring purpose](https://doc.rust-lang.org/nightly/cargo/reference/source-replacement.html). From Cargo's perspective, they are deemed different, even they eventually pointed to the same registry. Hence I think there is a confusion between these two different concepts.

Really appreciate your reproduction and detailed explanation. I have some questions about the environment setup in your work place. What is the use case of specifying `registry = "external"` when you've already mirrored crates.io? That is not immediately clear to me, and according to [the doc](https://doc.rust-lang.org/nightly/cargo/reference/source-replacement.html#:~:text=Cargo%20has%20a%20core%20assumption%20about%20source%20replacement%20that%20the%20source%20code%20is%20exactly%20the%20same%20from%20both%20sources.%20Note%20that%20this%20also%20means%20that%20a%20replacement%20source%20is%20not%20allowed%20to%20have%20crates%20which%20are%20not%20present%20in%20the%20original%20source.), Cargo made an assumption that source code is exactly the same from both replacement and original source. A crates.io mirror and an internal custom registry should be separate, if you also want to publish crates to a internal custom registry.


[^1]: See <https://github.com/rust-lang/cargo/issues/11406#issuecomment-2210972081> and https://medium.com/@alex.birsan/dependency-confusion-4a5d60fec610

weihanglo: Per the other review comment <https://github.com/rust-lang/cargo/pull/14713#discussion_r1815888268>, the fix seems incorrect. Also, we should really discuss the use case in an issue not PR. There going to close this. Feel free to open an issue, if you want to continue dicussing.

PR #14715, opened 2024-10-22
fix: add source replacement info when no matching package found

### LINKED ISSUE #14697: Misleading error message when crates.io is `replace-with` a local source
### Problem

When using a local source via a `replace-with` in `config.toml` and there is a missing package, `cargo build` gives confusing errors saying that it searched `crates-io` for the package when it only looked in the local file system. The error it gives is identical to the error from an actually missing package:
```
error: no matching package named `<package name>` found
location searched: registry `crates-io`
required by package `example v0.1.0`
```

### Possible Solution(s)

If you run `cargo seach` it gives much more specific information:
```
$ cargo search ahash
error: crates-io is replaced with non-remote-registry source dir /.../cargo-bug/vendor;
include `--registry crates-io` to use crates.io
```
It would be great if the original error included this "crates-io is replaced with non-remote-registry source" remark.

### Steps

1. Unpack this zip archive: [cargo-bug.zip](https://github.com/user-attachments/files/17397502/cargo-bug.zip)
2. Run `cargo build` in root
3. Get error:
    ```
    error: no matching package named `ahasha` found
    location searched: registry `crates-io`
    required by package `example v0.1.0 (/home/rchatham/Documents/programming/tmp/cargo-bug)`
    ```

As an alternative, instead of unpacking, make an empty folder, cd into it, make the following files and then do the following shell instructions:

#### `Cargo.toml`
```toml
[package]
name = "example"
version = "0.1.0"
edition = "2021"

[dependencies]
ahash = { version = "^0.8.9" }
```

#### `.cargo/config.toml`

```toml
[source.crates-io]

[source.vendored-sources]
directory = "vendor"
```

#### Shell instructions

Then:
```sh
mkdir src
touch src/lib.rs
mkdir vendor
cargo build
```

### Version

cargo 1.84.0-nightly (15fbd2f60 2024-10-08)
release: 1.84.0-nightly
commit-hash: 15fbd2f607d4defc87053b8b76bf5038f2483cf4
commit-date: 2024-10-08

chentodd:

Fixes #14697 

I updated error message in `resolver/errors` with [Registry::describe_source](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/registry/trait.Registry.html#tymethod.describe_source)

- Use message from `Registry::describe_source` as default message
- If the message is empty, use `SourceId` message

and I also updated related test cases.

1. Run following test cases:

- resolver-tests
- source_replacement
- build
- directory
- local_registry
- offline
- package
- publish
- registry
- update

2. Run [cargo-bug.zip](https://github.com/user-attachments/files/17493272/cargo-bug.zip)

- Unpack the zip archive
- Run `cargo build` in the root

following message should be seen:
```
error: no matching package named `ahasha` found
location searched: directory source `/home/foo/Desktop/cargo-bug/vendor` (which is replacing registry `crates-io`)
required by package `example v0.1.0 (/home/foo/Desktop/cargo-bug)`
```

### DIFF
--- a/tests/testsuite/build.rs
+++ b/tests/testsuite/build.rs
@@ -6043,7 +6043,7 @@ fn avoid_dev_deps() {
         .with_stderr_data(str![[r#"
 [UPDATING] `dummy-registry` index
 [ERROR] no matching package named `baz` found
-location searched: registry `crates-io`
+location searched: `dummy-registry` index (which is replacing registry `crates-io`)
 required by package `bar v0.1.0 ([ROOT]/foo)`
 
 "#]])
--- a/tests/testsuite/offline.rs
+++ b/tests/testsuite/offline.rs
@@ -184,7 +184,7 @@ fn cargo_compile_offline_not_try_update() {
         .with_status(101)
         .with_stderr_data(str![[r#"
 [ERROR] no matching package named `not_cached_dep` found
-location searched: registry `crates-io`
+location searched: crates.io index
 required by package `bar v0.1.0 ([ROOT]/bar)`
 As a reminder, you're using offline mode (--offline) which can sometimes cause surprising resolution failures, if this error is too confusing you may wish to retry without the offline flag.
 
@@ -195,7 +195,7 @@ As a reminder, you're using offline mode (--offline) which can sometimes cause s
     p.change_file(".cargo/config.toml", "net.offline = true");
     p.cargo("check").with_status(101).with_stderr_data(str![[r#"
 [ERROR] no matching package named `not_cached_dep` found
-location searched: registry `crates-io`
+location searched: crates.io index
 required by package `bar v0.1.0 ([ROOT]/bar)`
 As a reminder, you're using offline mode (--offline) which can sometimes cause surprising resolution failures, if this error is too confusing you may wish to retry without the offline flag.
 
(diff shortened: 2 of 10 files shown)


tests/testsuite/build.rs
```diff
@@ -6043,7 +6043,7 @@ fn avoid_dev_deps() {
         .with_stderr_data(str![[r#"
 [UPDATING] `dummy-registry` index
 [ERROR] no matching package named `baz` found
-location searched: registry `crates-io`
+location searched: `dummy-registry` index (which is replacing registry `crates-io`)
```
epage, line 6046: As there is no production change, I assume this belongs in the previous commit?

Each commit should be atomic which includes passing all tests

tests/testsuite/source_replacement.rs
```diff
@@ -294,3 +294,47 @@ fn source_replacement_with_registry_url() {
 "#]])
         .run();
 }
+
+#[cargo_test]
+fn source_replacement_with_no_package_in_directoy() {
```
epage, line 299: I'm not seeing this test updated in a later commit, so I assume this commit will fail on its own.

We ask for
- Commits to be atomic, including tests passing
- New tests to be added before a fix or feature that shows the existing behavior so that we can then see how the behavior changes with the fix (where possible)

chentodd: Sorry for the inconvenience I caused. I'll try to redo the work and follow the practice.

tests/testsuite/offline.rs
```diff
@@ -383,30 +383,39 @@ fn update_offline_not_cached() {
...
+    }
+
     p.cargo("update --offline")
         .with_status(101)
-        .with_stderr_data(
+        .with_stderr_data(&format!(
             "\
 [ERROR] no matching package named `bar` found
-location searched: registry `crates-io`
+location searched: {location_searched_message}
```
epage, line 400: We should try to keep snapshotting where possible.

Looking at this, I don't think the location is critical to the test and we could replace it with the `[..]`  wildcard

chentodd: Thanks for the feedback. I'm new to cargo test,  I want to make sure I fully understand your suggestion. Do I need to replace all instances of the hard-coded `location searched: XXX` with `[..]` in the `offline.rs` test cases, or only in this specific test case (`update_offline_not_cached`)?

epage: Only in this specific case to make it work in both situations

epage:
Thanks!

@bors r+

hoodmane:
Thanks @chentodd!

PR #14716, opened 2024-10-22
Cargo vendor: add source rewriting for private registries included in deps

P-E-Meunier:
In the case where the exact same dependency (hash-wise) is supplied by two different registries, Cargo.lock keeps only one sources. However, in that case, the output of `cargo vendor` contains only the registry from Cargo.lock, which isn't enough to compile with `--offline`.

This PR fixes this by adding the dependencies' registries of all vendored packages to the output of `cargo vendor`.


src/cargo/ops/vendor.rs
weihanglo, line 1: The contributing guide also encourages writing tests. Tests added before the commit of the actual fix can be seen as an [MRE](https://stackoverflow.com/help/minimal-reproducible-example), letting people understand the problematic behavior indeed exists. See <https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request>

weihanglo (commented):
Thanks for the pull request. Could you provide a bug reproduction and open an issue first? Without that it is not immediately clear what this patch is for. That workflow is also encouraged in [Cargo Contributor Guide](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo).

src/cargo/ops/vendor.rs
P-E-Meunier: This only happens when you have two copies of the same private registry, which is hard to test. But I'd of course welcome guidance on that if this is implemented somewhere.

P-E-Meunier:
Thanks for the review, I added a detailed explanation in a new issue.

PR #14723, opened 2024-10-23
download targeted transitive deps of with artifact deps'  target platform

### LINKED ISSUE #12554: bindeps: Enhance download_accessible to download all potential dependencies
#12421 has made some enhancements to support bindeps in a registry. However, this trips a problem with the implementation of [`download_accessible`](https://github.com/rust-lang/cargo/blob/8c08800172a4ecbb5971d90323a46457312dc23a/src/cargo/core/package.rs#L492). That function is responsible for downloading all packages before feature resolution. However, the current implementation does not handle the `target` dependency field causing some dependencies to not get downloaded. This in turn can cause a panic within the feature resolver.

I think it would be good to try to enhance `download_accessible` to be aware of the `target` field so that it can ensure that everything gets downloaded. I think that would involve having `collect_used_deps` check if a dependency is for a specific target to call itself with that different target (using a different value for `requested_kinds`). I think the `used` set might also need to be target-aware, since a dependency used for one target might have different dependencies than a dependency used for another target. There might also need to be some logic to handle `target="target"`, such as keeping track of what "target" means at the start.

#12421 has added the test `proc_macro_in_artifact_dep` for this that needs to be enabled. It might be good to try to think of other tests that might reveal issues in `download_accessible` (like the target="target" example).

elchukc:

Fixes #12554. `download_accessible` will now download platform-specified deps of artifact deps with `target = ...`.

It will also resolve the panic in `cargo tree -Z bindeps` in [#10593 (comment)](https://github.com/rust-lang/cargo/issues/10593#issuecomment-1317759526), where:

- a dependency of an artifact dependency is platform specified
- artifact dep itself is { target = } with the same platform as its own dependency
- the platform is not activated.

Essentially, `no entry found for key` was happening because for artifact deps with `{.., target = <target>}`, transitive deps that specified their platform as `<target>` were not downloaded. This is why adding `--target all` to `cargo tree -Z bindeps` made the bug dissapear.

Tests included in this PR should be enough.

~~Test `artifact_dep::proc_macro_in_artifact_dep` still throws; this PR will be ready for review once the test does not panic.~~

### Additional Information

`used` set needs to be target-aware
r? @weihanglo

### DIFF
--- a/src/cargo/core/package.rs
+++ b/src/cargo/core/package.rs
@@ -519,7 +519,7 @@ impl<'gctx> PackageSet<'gctx> {
                 target_data,
                 force_all_targets,
             );
-            for (pkg_id, _dep) in filtered_deps {
+            for (pkg_id, deps) in filtered_deps {
                 collect_used_deps(
                     used,
                     resolve,
@@ -529,6 +529,24 @@ impl<'gctx> PackageSet<'gctx> {
                     target_data,
                     force_all_targets,
                 )?;
+                let artifact_kinds = deps.iter().filter_map(|dep| {
+                    Some(
+                        dep.artifact()?
+                            .target()?
+                            .to_resolved_compile_kind(*requested_kinds.iter().next().unwrap()),
+                    )
+                });
+                for artifact_kind in artifact_kinds {
+                    collect_used_deps(
+                        used,
+                        resolve,
+                        pkg_id,
+                        has_dev_units,
+                        artifact_kind,
+                        target_data,
+                        force_all_targets,
+                    )?;
+                }
             }
             Ok(())
         }
--- a/src/cargo/core/package.rs
+++ b/src/cargo/core/package.rs
@@ -499,17 +499,18 @@ impl<'gctx> PackageSet<'gctx> {
         force_all_targets: ForceAllTargets,
     ) -> CargoResult<()> {
         fn collect_used_deps(
-            used: &mut BTreeSet<PackageId>,
+            used: &mut BTreeSet<(PackageId, CompileKind)>,
             resolve: &Resolve,
             pkg_id: PackageId,
             has_dev_units: HasDevUnits,
-            requested_kinds: &[CompileKind],
+            requested_kind: CompileKind,
             target_data: &RustcTargetData<'_>,
             force_all_targets: ForceAllTargets,
         ) -> CargoResult<()> {
-            if !used.insert(pkg_id) {
+            if !used.insert((pkg_id, requested_kind)) {
                 return Ok(());
             }
+            let requested_kinds = &[requested_kind];
             let filtered_deps = PackageSet::filter_deps(
                 pkg_id,
                 resolve,
@@ -524,7 +525,7 @@ impl<'gctx> PackageSet<'gctx> {
                     resolve,
                     pkg_id,
                     has_dev_units,
-                    requested_kinds,
+                    requested_kind,
                     target_data,
                     force_all_targets,
                 )?;
@@ -538,16 +539,22 @@ impl<'gctx> PackageSet<'gctx> {
         let mut to_download = BTreeSet::new();
 
         for id in root_ids {
-            collect_used_deps(
-                &mut to_download,
-                resolve,
-                *id,
-                has_dev_units,
-                requested_kinds,
-                target_data,
-                force_all_targets,
-            )?;
+            for requested_kind in requested_kinds {
+                collect_used_deps(
+                    &mut to_download,
+                    resolve,
+                    *id,
+                    has_dev_units,
+                    *requested_kind,
+                    target_data,
+                    force_all_targets,
+                )?;
+            }
         }
+        let to_download = to_download
+            .into_iter()
+            .map(|(p, _)| p)
+            .collect::<BTreeSet<_>>();
         self.get_many(to_download.into_iter())?;
         Ok(())
     }
--- a/tests/testsuite/artifact_dep.rs
+++ b/tests/testsuite/artifact_dep.rs
@@ -1777,9 +1777,7 @@ perhaps a crate was updated and forgotten to be re-vendored?
         .run();
 }
 
-// FIXME: `download_accessible` should work properly for artifact dependencies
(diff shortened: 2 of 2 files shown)


tests/testsuite/artifact_dep.rs
```diff
@@ -1636,8 +1651,9 @@ fn dep_of_artifact_dep_same_target_specified() {
     p.cargo("check -Z bindeps")
         .masquerade_as_nightly_cargo(&["bindeps"])
         .with_stderr_data(str![[r#"
-[LOCKING] 2 packages to latest compatible versions
+[LOCKING] 3 packages to latest compatible versions
 [COMPILING] baz v0.1.0 ([ROOT]/foo/baz)
+[COMPILING] one v0.1.0 ([ROOT]/foo/one)
```
elchukc, line 1656: The problem with this is it's non-deterministic whether `bar` or `one` get compiled first, which means this output causes the test to fail half the time. Looking for a way around this.

Rustin170506: Maybe you can try `with_stderr_contains`.

epage: We're trying to de-emphasize the `conains` cases. 

You can call `.unordered()` on the `str![]`

src/cargo/core/package.rs
```diff
@@ -518,16 +519,34 @@ impl<'gctx> PackageSet<'gctx> {
...
                     used,
                     resolve,
                     pkg_id,
                     has_dev_units,
-                    requested_kinds,
+                    requested_kind,
                     target_data,
                     force_all_targets,
                 )?;
+                let artifact_kinds = deps.iter().filter_map(|dep| {
```
epage, line 532: Can the rest of the changes in this commit be split out into a refactor commit?

src/cargo/core/package.rs
```diff
@@ -499,17 +499,18 @@ impl<'gctx> PackageSet<'gctx> {
         force_all_targets: ForceAllTargets,
     ) -> CargoResult<()> {
         fn collect_used_deps(
-            used: &mut BTreeSet<PackageId>,
+            used: &mut BTreeSet<(PackageId, CompileKind)>,
             resolve: &Resolve,
             pkg_id: PackageId,
             has_dev_units: HasDevUnits,
-            requested_kinds: &[CompileKind],
+            requested_kind: CompileKind,
```
weihanglo, line 506: This seems mostly correct to me. Cargo allows only one target platform per artifact deps. Thanks folks. 👍🏾


With this RFC the approach here might still have some issues but it is not even landed yet so no need to worry
https://rust-lang.github.io/rfcs/3176-cargo-multi-dep-artifacts.html

tests/testsuite/artifact_dep.rs
```diff
@@ -1603,6 +1603,7 @@ fn dep_of_artifact_dep_same_target_specified() {
 
                     [dependencies]
                     bar = {{ path = "bar", artifact = "bin", target = "{target}" }}
+                    one = {{ path = "one", artifact = "bin", target = "x86_64-fortanix-unknown-sgx" }}
```
elchukc, line 1606: Is there some way to test for _another_ target besides the one already represented by `target` on line 1605? I know this will fail in Github Actions because the OS images don't have the fortanix platform installed.

epage: Is https://docs.rs/cargo-test-support/latest/cargo_test_support/cross_compile/fn.alternate.html what you are looking for?

src/cargo/core/package.rs
epage: What we would do in that situation is just call `collect_used_deps` for each `target`.  This errs on the side of walking the tree multiple times, once for each target, rather than trying to keep state so that we only walk once.

src/cargo/core/package.rs
elchukc: Not totally sure what you mean here. Are you talking about everything changed about collect_deps besides filtering for artifact kinds? The separate commit wouldn't compile.

tests/testsuite/artifact_dep.rs
elchukc: Actually, I don't think this extra package in the project is necessary. I added it during development to see more helpful print output.

src/cargo/core/package.rs
```diff
@@ -538,16 +539,22 @@ impl<'gctx> PackageSet<'gctx> {
...
+                    resolve,
+                    *id,
+                    has_dev_units,
+                    *requested_kind,
+                    target_data,
+                    force_all_targets,
+                )?;
+            }
         }
+        let mut to_download = to_download
```
epage: I think this `mut` is unused

epage: Ah, I accidentally removed it in the follow up commit.  That should be in this one

tests/testsuite/artifact_dep.rs
```diff
@@ -1829,7 +1828,18 @@ fn proc_macro_in_artifact_dep() {
...
-        .with_stderr_data(str![[r#""#]])
+        .with_stderr_data(
+            r#"...
+[UPDATING] `dummy-registry` index
+[LOCKING] 2 packages to latest compatible versions
+[DOWNLOADING] crates ...
+[ERROR] failed to download from `[ROOTURL]/dl/pm/1.0.0/download`
+
+Caused by:
+  [37] Could not read a file:// file (Couldn't open file [ROOT]/dl/pm/1.0.0/download)
```
elchukc, line 1839: Ok so the deployment is failing because the OS tests expect a slightly different error message. When running locally, this test passes. Do the OS images need to be update?
```diff
- expected
+ actual
Caused by:
- [37] Could not read a file:// file (Couldn't open file [ROOT]/dl/pm/1.0.0/download)
+ [37] Couldn't read a file:// file (Couldn't open file [ROOT]/dl/pm/1.0.0/download)
```

epage: You can use the wildcard `[..]` to ignore a section that differs

PR #14728, opened 2024-10-24
refactor(fingerprint): avoid unnecessary fopen calls

weihanglo:

This was found by @kobzol.

While not in a hot path, it is no harm to do this tiny thing now.

Usually if `File::open` could failed,
`fs::metadata` should have already failed,
so should have no real user-facing change.


ehuss:
Thanks!

@bors r+

PR #14730, opened 2024-10-25
refactor(env): remove unnecessary clones

weihanglo:

This was found during the review of #14701.

Not every environment variable access requires an owned type.
To be precise, most of our usages don't.

Refactor only. Should not have any change in behavior.

One drawback is that the fn signatures deviate from `std::env::var{_os}`.


ehuss:
Thanks! Yea, I think the original intent was to mirror `std`'s API, but I don't think we necessarily need to do that since we aren't under the same OS-level constraints.

@bors r+

PR #14736, opened 2024-10-27
Deprecate `cargo verify-project`

### LINKED ISSUE #14679: Deprecate `cargo verify-project`
### Problem

I never found `cargo verify-project` useful. It never checks semantics of `Cargo.toml`. What it does is merely

* Locating a workspace.
* Parsing `Cargo.toml` and ensuring that be valid TOML, during the course of locating a workspace.

I don't know if in the wild it is heavily used, but apparently on GitHub it's not (see [this search result](https://github.com/search?q=%22cargo+verify-project%22&type=code)). It also causes confusions like people expecting it to verify more stuff (#11202).

I wonder if it's time to decide how this command should be.

### Proposed Solution

We have at least two choices.

* Make it clear that what `cargo verify-project` verifies. Perhaps at this moment only TOML syntax. We could expand to simple manifest semantic, though the line will never be clear and can't be a complete verification.

* Mark `cargo verify-project` as _deprecated_. Whomever want to use it could continue, though we don't add any support in the future.
  
  `cargo locate-project --workspace` might be a good alternative. (Well, not completely. `locate-project` additionally checks UTF8 for some unknown reason. And IMO it should be allowed.) 

### Notes

Commit: e8adba9d24257e14fef8793cecabb55b2fbcd0e3

MoskalykA:

Information is available in the following issue #14679

### Additional information

Closes #14679


src/bin/cargo/commands/verify_project.rs
```diff
@@ -5,12 +5,16 @@ use std::process;
 
 pub fn cli() -> Command {
     subcommand("verify-project")
-        .about("Check correctness of crate manifest")
+        .about(color_print::cstr!(
+            "\
+Check correctness of crate manifest.
+
+Deprecated, use `<cyan,bold>cargo locate-project --workspace</>` instead (also checks UTF8).\
```
weihanglo, line 12: I would avoid mentioning `cargo locate-project`. Same as `cargo verify-project`, it basically checks only TOML validity. The purpose of `cargo locate-project` is not checking manifest either. This is not like the relationship of cargo-read-manifest to cargo-metadata.

I would probably put this instead:

```
Deprecated, see https://github.com/rust-lang/cargo/issues/14679
```

src/doc/src/commands/cargo-verify-project.md
weihanglo, line 1: `build-man` is bad at tracking removed commands (we seldom do it).

I believe these files also need to remove.

* src/etc/man/cargo-verify-project.1
* src/doc/man/cargo-verify-project.md
* src/doc/man/generated_txt/cargo-verify-project.txt

MoskalykA: thanks u <3 !

weihanglo (approved):

weihanglo:
Thanks for the contribution!

@bors r+

MoskalykA:
@weihanglo `failed to connect to github.com: Connection timed out` we just need to re-launch the bors tests?

Issue #11298, opened 2022-10-27
Add more optimizations to the release build profile.


horacehoff:
Team updates:
- https://github.com/rust-lang/cargo/issues/11298#issuecomment-3258950086

---

### Problem

The release profile is used when wanting to build an optimized binary. Thus, I think it would be logical to optimize it to the fullest, sometimes at the cost of a greater build time, but I think it is a good compromise for many people, including me.

### Proposed Solution

More optimizations flags should be added to the release build profile, notably lto=true and codegen-units=1, as well as optimizing all the packages with opt-level=3.
This greatly enhances runtime performance, although yes, at the cost of a bigger binary size and longer build time.

### Notes

I have tested what I am proposing, with the following added in my cargo.toml:
```
[profile.release]
lto = true
codegen-units = 1

[profile.release.package."*"]
opt-level = 3
```
This improved the output binary's runtime performance by orders of magnitude, thus this issue.

epage:
I suspect `lto = "thin"` is a sufficient compromise that it'd be worth being the default compared to `"fat"` / `true`.

Muscraft:
Compile times are a sensitive area and increasing them in any way is always a worry.

I ran a test on my machine (M1 Max chip 32GB ram) against a private project that has 395 dependencies (many are very heavy) and the jump in compile time is large for "fat" LTO. The results also appear to agree that `lto = "thin"` could work. 

One thing to consider is if changing the `release` profile could be breaking in some unintended way. [Crater](https://doc.crates.io/contrib/tests/crater.html?highlight=crater#crater) might need to be run to see if it changes anything. Either way, it might be a good idea to add a new profile that contains these changes, instead of changing `release`.

<details><summary>Test Results</summary>

| `profile` | time |
| -- | -- |
| "fat" LTO codegen=1 | 2m 30s |
| "fat" | 2m 27s |
| "thin" LTO codegen=1 |  1m 44s |
| "thin" |  1m 08s |
| No LTO codegen=1 | 1m 39s |
| `release` | 1m 09s |


<details><summary>"fat" codegen=1</summary>

```toml
[profile.production]
inherits = "release"
lto = true
codegen-units = 1
opt-level = 3
```

</details>

<details><summary>"fat"</summary>

```toml
[profile.production]
inherits = "release"
lto = true
opt-level = 3
```

</details>


<details><summary>"thin" codegen=1</summary> 

```toml
[profile.production]
inherits = "release"
lto = "thin"
[...]

PR #14738, opened 2024-10-27
Enable LTO for profile.release and profile.bench in new packages

foxtran:
Hi!

This PR enables fat Link-Time Optimization (LTO) by default for new packages which are started via `cargo init` command. To speed-up development cycle, LTO is enabled only for performance-related build types (`release`, `bench`). For library targets, only `bench` profile gets enabled LTO.

The main advantage is that new projects will no longer need to enable LTO manually. This means users will benefit from binaries that are a bit more optimized for both size and speed just from the start of their projects. Since projects are generally small in the early stages, this change should not significantly impact compilation time or memory usage — these issues usually arise only in the compilation with LTO of really large projects.

This PR does not affect existing packages. I've test these changes via `cargo test` command or more precisely:
```bash
cargo test --test testsuite -- cargo_init::
cargo test --test testsuite -- cargo_new::
```
because only these tests were affected by my changes.

Hope these changes will be merged and all Rust community gets faster packages from scratch :)

### DIFF
--- a/src/cargo/ops/cargo_new.rs
+++ b/src/cargo/ops/cargo_new.rs
@@ -784,6 +784,9 @@ fn mk(gctx: &GlobalContext, opts: &MkOptions<'_>) -> CargoResult<()> {
         manifest["profile.release"]["lto"] = toml_edit::value(true);
     }
 
+    manifest["profile.bench"] = toml_edit::Item::Table(toml_edit::Table::new());
+    manifest["profile.bench"]["lto"] = toml_edit::value(true);
+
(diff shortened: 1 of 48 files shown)


epage:
Agreed with @Rustin170506.  Thank you for your interest in improving the ecosystem.   To add to why we prefer Issues is that there are many angles to look at (what settings, set in `init` or default profile, etc) and PRs aren't ideal for this type of setup.

#11298 might be a good place for discussing this.  Closing this until there is resolution.

Issue #12114, opened 2023-05-09
Add `cargo  help --commands`


ToBinio:
### Problem

currently, the only real way to know which cargo subcommands are available is through `cargo --list`. However the output of `cargo --list` is not very clear.

```
b                    alias: build
bench                Execute all benchmarks of a local package
build                Compile a local package and all of its dependencies
build-man            alias: run --package xtask-build-man --
c                    alias: check
```

the main problems for me are:

- all types of commands (built-in commands, custom aliases, and installed commands) are in one big list
- built in commands and there aliases are visually not connected
- there is no computer-friendly way to get this information

a bit more information about these problems can be found in #12093 

### Proposed Solution

Adding an alternative/replacement for `cargo --list`. `cargo help --commands` this change would also make it clear what the comment does. Printing information about all possible commands.
this could then fix the before-noted problems.

example output:

```
Built-in commands:
    add                     Add dependencies to a Cargo.toml manifest file
    bench                   Execute all benchmarks of a local package
    build, b                Compile a local package and all of its dependencies
    ...

Custom aliases:
    build-man               run --package xtask-build-man --
    ...

Installed commands:
    clippy                  Checks a package to catch common mistakes and improve your Rust code.
    ...
```

## Spliting types
instead of having all commands in one long list, it would split them into 3 categories.

- Built-in commands: all commands which are found in the cargo code. this would also include their aliases 
- custom aliases: this would include all aliases defined by users in their settings files. and aliases which are more complex than a one-to-one mapping to another command. e.g `build-man`
- installed commands: all commands not found in the cargo code. mostly installed through `cargo install` but also pre-installed things like `clippy` and `fmt`

an example of a similar splitting behavior can be seen with `git --help`

## aliases
instead of having aliases and their base explanation in separate lines.

like now :

```
b                    alias: build
bench                Execute all benchmarks of a local package
build                Compile a local package and all of its dependencies
build-man            alias: run --package xtask-build-man --
c                    alias: check
```

it could look more like `cargo --help` | default `clap` help

```
build, b     Compile the current package
check, c     Analyze the current package and report errors, but don't build object files
clean        Remove the target directory
doc, d       Build this package's and its dependencies' documentation
```

## Other formats
instead of just the normal human readable code, it would also have support for the 
`--message-format` flag. This would enable printing the information in a more computer-friendly way like `json`.

### Notes

In the future, this could be expanded to extra commands for every single command category.

- cargo help --built-in-commands
[...]

PR #14739, opened 2024-10-28
fix: clean up for deprecated and removed commands

weihanglo:

This PR is a follow-up of <https://github.com/rust-lang/cargo/issues/14679>.

* A new doc page for deprecated and removed commands
* Remove shell completions for deprecated commands
* Remove `verify-project` and `read-manifest` from `cargo --list` output.

`mdbook serve src/doc` and read the doc.

I've set up redirections for all removed/deprecated commands to maintain consistency,
though only `verify-project` really need it I believe.


src/bin/cargo/commands/read_manifest.rs
```diff
@@ -1,7 +1,10 @@
+//! Deprecated.
+
 use crate::command_prelude::*;
 
 pub fn cli() -> Command {
     subcommand("read-manifest")
+        .hide(true)
```
epage, line 7: This is the only commit I'm unsure about.

On one hand, this will reduce clutter and make completions nicer.

On the other hand, `--list` is no longer exhaustive which can be important for knowing what command names are available.  Unsure what other use cases might be affected.

weihanglo: That reminds me https://github.com/rust-lang/cargo/issues/12114.

`cargo --list` may be a little brittle to change. I've restored the behavior of `--list` and only hide commands for a nicer shell completion.
Also add some DEPRECATED/REMOVED prefix for making the status more standing out in `cargo --list`.

epage (approved):
Feel free to r= me when this is ready

PR #14742, opened 2024-10-29
fix(publish): Downgrade version-exists error to warning on dry-run 

### LINKED ISSUE #14721: Cannot dry-run `cargo publish` without bumping versions
### Problem

When a tool like `cargo release` runs in a dry-run mode, it shouldn't be bumping version numbers.  However, as of #14448, `cargo publish --dry-run` will fail unless the version numbers are bumped

### Steps

```console
$ git clone git@github.com:clap-rs/clap.git
$ cd clap
$ cargo +beta release major -vvv
```

### Possible Solution(s)

Revert #14448 which makes us lose out on early error reporting.  However, if we view publish as plumbing with tools like `cargo release` handling it, make its not too bad

Skip #14448's check in dry-run mode or downgrade it to a warning.  Deviating behavior between dry-run or not can be confusing and break other workflows.

Tell people to use `cargo package`.  There are slight behavior differences though.

### Notes

cargo 1.83.0-beta.2 (15fbd2f60 2024-10-08)

### Version

_No response_

epage:

Especially for beta, this was the most conservative, minimal change.

Fixes #14721

### Additional information

This will get cherry-picked to beta

### DIFF
--- a/src/cargo/ops/registry/publish.rs
+++ b/src/cargo/ops/registry/publish.rs
@@ -133,7 +133,7 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
             .acquire_package_cache_lock(CacheLockMode::DownloadExclusive)?;
 
         for (pkg, _) in &pkgs {
-            verify_unpublished(pkg, &mut source, &source_ids)?;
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks!

@bors r+

PR #14747, opened 2024-10-30
Add transactional semantics to `rustfix`

### LINKED ISSUE #13027: `cargo fix`: duplicate diagnostic detection does not handle insert-only replacements
The following example:

```rust
macro_rules! foo {
    () => {
        &1;
    };
}

fn main() {
    foo!();
    foo!();
}
```

generates a suggestion (on beta 1.55) to insert a fix like this:

```rust
let _ = &1;
```

However, since this diagnostic is triggered from macros, rustc emits two separate machine-applicable suggestions at the exact same spot. That causes rustfix to end up changing the code to:

```rust
let _ = let _ = &1;
```

which fails to compile.

rust-lang/rustfix#131 added detection to avoid applying duplicate suggestions, but I believe it does not handle the case when it is an `insert_only` suggestion.

### LINKED ISSUE #14699: [rustfix] Two suggestions applying to the same span cause syntax errors
### Problem

https://github.com/rust-lang/rust-clippy/issues/13549

Two suggestions all making changes to the same span have the potential to cause a syntax error.

### Steps

1. `cargo new --lib test-crate`
2. Write the following code into `src/lib.rs`:

```rust
struct Foo(i64); 
impl Into<i64> for Foo {
    fn into(self) -> i64 {
        self.0.clone()
    }
}`
```

3. `cargo clippy --fix`

### Possible Solution(s)

Making sure that a spans don't overlap each other when having made a suggestion. See more context in the linked issue

### Notes

_No response_

### Version

cargo version --verbose

saites:

This PR adds transactional semantics to `rustfix::Data`, enabling `rustfix::CodeFix` to apply `Suggestion`s as atomic units and rollback partially-applied changes when they conflict with existing ones. The basic approach and goals are discussed [in a comment on issue 14699](https://github.com/rust-lang/cargo/issues/14699#issuecomment-2427501232). In that comment, I proposed a solution which extended the existing `State` enumeration, but described an alternative that simplifies the overall tracking of incoming changes; this PR implements the latter.

I've added an additional test and updated the existing ones. The tests in `parse_and_replace` already cover this case, particularly thanks to https://github.com/rust-lang/cargo/pull/14765 added by @weihanglo. It's still a good idea to experiment with `cargo clippy --fix` on repos that match the cases described in these open issues. 

### Additional information

Fixes #14699
Fixes rust-lang/rust-clippy/issues/13549

### DIFF
--- a/crates/rustfix/src/lib.rs
+++ b/crates/rustfix/src/lib.rs
@@ -243,9 +243,11 @@ impl CodeFix {
     pub fn apply_solution(&mut self, solution: &Solution) -> Result<(), Error> {
         for r in &solution.replacements {
             self.data
-                .replace_range(r.snippet.range.clone(), r.replacement.as_bytes())?;
-            self.modified = true;
+                .replace_range(r.snippet.range.clone(), r.replacement.as_bytes())
+                .inspect_err(|_| self.data.restore())?;
         }
+        self.data.commit();
+        self.modified = true;
         Ok(())
     }
 
--- a/crates/rustfix/src/replace.rs
+++ b/crates/rustfix/src/replace.rs
@@ -1,56 +1,100 @@
-//! A small module giving you a simple container that allows easy and cheap
-//! replacement of parts of its content, with the ability to prevent changing
-//! the same parts multiple times.
+//! A small module giving you a simple container that allows
+//! easy and cheap replacement of parts of its content,
+//! with the ability to commit or rollback pending changes.
+//!
+//! Create a new [`Data`] struct with some initial set of code,
+//! then record changes with [`Data::replace_range`],
+//! which will validate that the changes do not conflict with one another.
+//! At any time, you can "checkpoint" the current changes with [`Data::commit`]
+//! or roll them back (perhaps due to a conflict) with [`Data::restore`].
+//! When you're done, use [`Data::to_vec`]
+//! to merge the original data with the changes.
+//!
+//! # Notes
+//!
+//! The [`Data::to_vec`] method includes uncommitted changes, if present.
+//! The reason for including uncommitted changes is that typically, once you're calling those,
+//! you're done with edits and will be dropping the [`Data`] struct in a moment.
+//! In this case, requiring an extra call to `commit` would be unnecessary work.
+//! Of course, there's no harm in calling `commit`---it's just not strictly necessary.
+//!
+//! Put another way, the main point of `commit` is to checkpoint a set of known-good changes
+//! before applying additional sets of as-of-yet unvalidated changes.
+//! If no future changes are expected, you aren't _required_ to pay the cost of `commit`.
+//! If you want to discard uncommitted changes, simply call [`Data::restore`] first.
 
+use std::ops::Range;
 use std::rc::Rc;
 
 use crate::error::Error;
 
-/// Indicates the change state of a [`Span`].
-#[derive(Debug, Clone, PartialEq, Eq)]
-enum State {
-    /// The initial state. No change applied.
-    Initial,
-    /// Has been replaced.
-    Replaced(Rc<[u8]>),
-    /// Has been inserted.
-    Inserted(Rc<[u8]>),
-}
-
-impl State {
-    fn is_inserted(&self) -> bool {
-        matches!(*self, State::Inserted(..))
-    }
-}
-
-/// Span with a change [`State`].
-#[derive(Clone, PartialEq, Eq)]
+/// Data that should replace a particular range of the original.
+#[derive(Clone)]
 struct Span {
-    /// Start of this span in parent data
-    start: usize,
-    /// up to end excluding
-    end: usize,
-    /// Whether the span is inserted, replaced or still fresh.
-    data: State,
+    /// Span of the parent data to be replaced, inclusive of the start, exclusive of the end.
+    range: Range<usize>,
+    /// New data to insert at the `start` position of the `original` data.
+    data: Rc<[u8]>,
+    /// Whether this data is committed or provisional.
+    committed: bool,
 }
 
 impl std::fmt::Debug for Span {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let state = match self.data {
-            State::Initial => "initial",
-            State::Replaced(_) => "replaced",
-            State::Inserted(_) => "inserted",
+        let state = if self.is_insert() {
+            "inserted"
+        } else {
+            "replaced"
         };
-        write!(f, "({}, {}: {state})", self.start, self.end)
+
+        let committed = if self.committed {
+            "committed"
+        } else {
+            "uncommitted"
+        };
+
+        write!(
+            f,
+            "({}, {}: {state}, {committed})",
+            self.range.start, self.range.end
+        )
     }
 }
 
-/// A container that allows easily replacing chunks of its data
+impl Span {
+    fn new(range: Range<usize>, data: &[u8]) -> Self {
+        Self {
+            range,
+            data: data.into(),
+            committed: false,
+        }
+    }
+
+    /// Returns `true` if and only if this is a "pure" insertion,
+    /// i.e. does not remove any existing data.
+    ///
+    /// The insertion point is the `start` position of the range.
+    fn is_insert(&self) -> bool {
+        self.range.start == self.range.end
+    }
+}
+
+impl PartialEq for Span {
+    /// Returns `true` if and only if this `Span` and `other` have the same range and data,
+    /// regardless of `committed` status.
+    fn eq(&self, other: &Self) -> bool {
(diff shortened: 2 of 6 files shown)


src/cargo/ops/fix.rs
```diff
@@ -968,20 +968,12 @@ fn rustfix_and_fix(
...
-            if suggestion
-                .solutions
-                .iter()
-                .any(|sol| !already_applied.insert(sol))
-            {
-                continue;
-            }
             match fixed.apply(suggestion) {
                 Ok(()) => fixed_file.fixes_applied += 1,
+                Err(rustfix::Error::AlreadyReplaced {
```
weihanglo, line 974: While I don't think the compiler will get this fixed in any near future, we may want to restore this comment in case there is such a future.

(probably ditto in `apply_suggestions()`?)

```
            // If this replacement matches exactly the part that we would
            // otherwise split then we ignore this for now. This means that you
            // can replace the exact same range with the exact same content
            // multiple times and we'll process and allow it.
            //
            // This is currently done to alleviate issues like
            // rust-lang/rust#51211 although this clause likely wants to be
            // removed if that's fixed deeper in the compiler.
```

crates/rustfix/src/replace.rs
```diff
@@ -297,28 +303,35 @@ mod tests {
...
+        d.replace_range(1..1, "b").unwrap();
+        assert_eq!("fboo", str(&d.to_vec()));
+        assert!(matches!(
+            d.replace_range(1..1, "b").unwrap_err(),
+            Error::AlreadyReplaced {
+                is_identical: true,
+                ..
+            },
+        ));
+        assert_eq!("fboo", str(&d.to_vec()));
```
weihanglo: This test change should move to the second commit maybe.

weihanglo: While this has already covered the case of <https://github.com/rust-lang/cargo/issues/13027>, I feel we should generate a snapshot test from the exact macro diagnostics, like <https://github.com/rust-lang/cargo/blob/de5832fe8fe1a05e278eee82bbc801d544ed4bd1/crates/rustfix/tests/everything/E0178.rs>.

I can help with that if it is too much a burden.

crates/rustfix/src/replace.rs
```diff
@@ -97,96 +152,41 @@ impl Data {
...
-
-        if part_to_split.data != State::Initial {
-            return Err(Error::AlreadyReplaced);
-        }
-
-        let mut new_parts = Vec::with_capacity(self.parts.len() + 2);
+        // Keep sorted by start position, with pure inserts before replacements.
+        let ins_point = self.parts.partition_point(|span| {
+            span.range.start < range.start
+                || (span.range.start == range.start && span.range.end < range.end)
```
weihanglo, line 171: Could you explain a bit why we need `span.range.end < range.end`?

crates/rustfix/src/lib.rs
weihanglo, line 1: Looks like the last three commits are mostly refactoring. To make things clearer and more traceable, could we move them to a separate or a follow-up PR?

weihanglo (commented):
This is awesome! Thank you for this patch!

crates/rustfix/src/replace.rs
```diff
@@ -97,96 +152,41 @@ impl Data {
...
-            .iter()
-            .position(|p| !p.data.is_inserted() && p.start <= range.start && p.end >= range.end)
-        else {
-            tracing::debug!(
-                "no single slice covering {}..{}, current slices: {:?}",
-                range.start,
-                range.end,
-                self.parts,
-            );
-            return Err(Error::MaybeAlreadyReplaced(range));
```
weihanglo, line 122: If this is no longer needed, we should remove it (and maybe add `#[mon_exhaustive]` to the error enum as a follow-up).

crates/rustfix/src/replace.rs
weihanglo: > with pure inserts before replacements.

I think got it :)

saites:
@weihanglo Thank you so much for your review! 

I'll reply soon to your comments and submit some updates, but I won't be able to complete the changes until tomorrow at the earliest.

crates/rustfix/src/lib.rs
```diff
@@ -262,22 +262,21 @@ impl CodeFix {
...
+
+/// Discard errors for identical changes while passing through other error types.
+pub fn ignore_identical<T: Default>(err: Error) -> Result<T, Error> {
+    match err {
+        Error::AlreadyReplaced {
+            is_identical: true, ..
+        } => Ok(Default::default()),
+        _ => Err(err),
+    }
+}
```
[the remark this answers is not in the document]
saites: Fair enough -- and I'm happy to inline it. I originally had it split out and `pub` so I could make use of it in `fix.rs`, but then decided against it because it would change the "number of fixes" applied.

crates/rustfix/src/replace.rs
saites: I agree that it should be removed, but _should_ it be marked `#[non_exhaustive]`? It seems like this variant was a bit of a hedge against an implementation error, but not really a mistake from the caller.

crates/rustfix/src/lib.rs
saites: Yes, certainly. I got a bit over-zealous during the original work, and that's actually why I pulled them out into these separate commits. I'll move those commits out of this PR. 

> the last three commits

Do you mean last three, or last four? That is, should I remove all except the first commit, or do you want to include the changes to `Error`? If the former, I think this PR should still remove the `MaybeAlreadyReplaced` variant, as discussed in a different comment.

As for submitting those commits as potential refactors, I was thinking I'd open an issue for them first, since in retrospect, they could use some discussion. But also, I don't want to increase your workload unnecessarily, so no worries if the additional commits aren't really something the project is looking to review/adopt at this time.

crates/rustfix/src/replace.rs
saites: Yeah, it's to keep inserts ahead, but maybe I should elaborate in the comment why that's important: it ensures the checks that come next are valid, particularly if there might be multiple inserts/replacements at the same point. A single insert and replacement can have the same start, but any others at that point need to return an error.

crates/rustfix/src/replace.rs
saites: I think I can handle it. I wanted to ask, though, do you know a good way to capture the original issue I raised, since it's `clippy` producing those suggestions?

crates/rustfix/src/replace.rs
```diff
@@ -1,56 +1,100 @@
...
 }
 
-/// A container that allows easily replacing chunks of its data
+/// A container that allows easily replacing chunks of its data.
 #[derive(Debug, Clone, Default)]
 pub struct Data {
     /// Original data.
     original: Vec<u8>,
     /// [`Span`]s covering the full range of the original data.
+    /// Important: it's expected that the underlying implementation maintains this in order,
```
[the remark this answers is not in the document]
saites: Good idea.

crates/rustfix/src/replace.rs
weihanglo: Sounds good to me!

crates/rustfix/src/lib.rs
weihanglo: Yeah. IMO ideally this PR could be splitted into three maybe?

* transactional semantics (fixes #14699)
* `AlreadyReplaced` handling (fixes #13027)
* Other refactors/optimizations 

I don't think we need extra issues for any of them.

crates/rustfix/src/replace.rs
weihanglo: See <https://github.com/rust-lang/cargo/blob/7421ccf9159d5ac43883c53abe41ef1da657acfd/crates/rustfix/tests/parse_and_replace.rs?plain=1#L1-L21>.

Though I believe it is already fixed via <https://github.com/rust-lang/cargo/pull/13728>>. Granted your solution is better 👍🏾 

Just opened https://github.com/rust-lang/cargo/pull/14765 to switch and close it, so you don't need to worry setting up a test.

weihanglo:
@saites 
Since some public API has changed, to make CI pass, we'll need to bump the major version of rustfix. Would you mind helping with that?

saites:
> @saites Since some public API has changed, to make CI pass, we'll need to bump the major version of rustfix. Would you mind helping with that?

Yes, I think I can handle that.

weihanglo (approved):

weihanglo:
Thank you for working on this!

I just did a real world test: `cargo clippy --fix --allow-dirty` with both this commit and Cargo from 1.82 toolchain. They've both addressed 551 warnings (527 duplicates), and outputs the same fixed code. Looks cool!

@bors r+

weihanglo:
Feel free to work on refactoring pull requests!

PR #14749, opened 2024-10-30
Change config paths to only check CARGO_HOME for cargo-script

Rustin170506:

ref https://github.com/rust-lang/cargo/issues/12207#issuecomment-2358818671
After this change, we will only check the configs from the `CARGO_HOME`.

- [ ] TBD, Need to add a test for it.

### Additional information

r? @epage


epage:
> What do you mean by downloading a script? And what kind of spurious files might end up being created? I thought it only read the config files.

If I download a script off the internet, my browser will put it in `~/Downloads`.  That is a grab bag directory with who knows what in it.  A user might have extracted a project into that directory to look at that could have dropped a `.cargo/config.toml` into it.  The concern is with that being used when it wasn't intended.  That unfortunately means we'll not load any config for cargo scripts.

Rustin170506 (commented):

tests/testsuite/script.rs
```diff
@@ -306,20 +309,25 @@ fn use_script_config() {
...
-[build]
-rustc = "non-existent-rustc"
-"#,
-        )
         .file("script.rs", script)
         .build();
 
     // Verify the config is bad
     p.cargo("-Zscript script.rs -NotAnArg")
+        .env("CARGO_HOME", &my_home)
```
epage, line 330: Why are we overriding CARGO_HOME rather than using the cargo home created for you?

epage: See https://doc.rust-lang.org/nightly/nightly-rustc/cargo_test_support/paths/fn.cargo_home.html

Rustin170506: I don't want the second case to be affected by the default cargo home either.

Rustin170506: I believe I should change the test case to focus solely on checking the cargo home configuration. I will update it.

Rustin170506 (commented):

tests/testsuite/script.rs
Rustin170506: Updated. 
1. I checked that the config from the current directory is used.
2. I checked that the config from the parent directory is not used.
3. I checked that config from the home directory is used.
4. Renamed it to use_cargo_home_config.

Rustin170506 (commented):

Rustin170506:

PR #14750, opened 2024-10-30
fix(util): Respect all `..`s in `normalize_path`

epage:

The fact that `normalize_path` was only designed for absolute paths bit us when working out #14497 and so I decided to make sure it worked.  The other alternative I considered was having it assert that the path was absolute.

Since I did try out the assert and Cargo tests hit it, this likely fixes something but I haven't dug through to be able to say what.


weihanglo (approved):

weihanglo:
It seems correct, though I don't know if there is anyone relying on this problematic behavior 😆.


@bors r+

Issue #14712, opened 2024-10-21
"memory allocation of .. bytes failed" appears when using `cargo run test` to run specific test name


chentodd:
### Problem

`cargo run test` shows "memory allocation of .. bytes failed" when I run specific test case:

```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/cargo test source_replacement`
memory allocation of 63063237072 bytes failed
Aborted (core dumped)
```
I expect there are no issues when I use `cargo run test` to run the tests.

### Steps

1. Run `rustup update` to update needed tools to build `cargo`
2. Clone `cargo` repository
3. `cd` into `cargo` repository
4. Change to `master` branch
5. Run `cargo check --workspace`
6. Run `cargo build`
7. Run `cargo run test source_replacement`

here is the test video that contains step.5 to step.7 for the reference: [20241021_cargo_test_video.zip](https://github.com/user-attachments/files/17464077/20241021_cargo_test_video.zip)

### Notes

The error message in this issue is similar to #7543

### Version

cargo 1.84.0 (edd36eba5 2024-10-21)
release: 1.84.0
commit-hash: edd36eba5e0d6e0cfcb84bd0cc651ba8bf5e7f83
commit-date: 2024-10-21
host: x86_64-unknown-linux-gnu
libgit2: 1.8.1 (sys:0.19.0 vendored)
libcurl: 8.5.0 (sys:0.4.74+curl-8.9.0 system ssl:OpenSSL/3.0.13)
os: Ubuntu 24.4.0 (noble) [64-bit]

additional hardware information of test machine:
Memory: 16 GB

epage:
I wasn't able to reproduce.  Unsure what the difference is.

weihanglo:
The RAM size is critical.

> additional hardware information of test machine:
> Memory: 16 GB

I reproduced it on a machine with 4 GiB RAM.

```
(24-10-24 17:12:24) <0> [~/cargo]  
%  cargo -vV
cargo 1.82.0 (8f40fc59f 2024-08-21)
release: 1.82.0
commit-hash: 8f40fc59fb0c8df91c97405785197f3c630304ea
commit-date: 2024-08-21
host: x86_64-unknown-linux-gnu
libgit2: 1.8.1 (sys:0.19.0 vendored)
libcurl: 8.9.0-DEV (sys:0.4.74+curl-8.9.0 vendored ssl:OpenSSL/1.1.1w)
[...]

PR #14751, opened 2024-10-30
fix: track version in fingerprint dep-info files

weihanglo:

Encodes the version information into Cargo's fingerprint dep-info files,
so that when the format encoding changes in the future,
Cargo understands a dep-info file was outdated and doesn't bother parsing it.

Since there was no version info encoded in the old format (call it v0),
to be compatible with older cargoes,
this PR works around it with a horrible hack.
It is explained in the doc comment of `EncodedDepInfo`.

This PR also adds a static file which is v0 depinfo,
to ensure newer cargoes give up parsing it successfully without allocation errors.
See the allocation failures in this CI job result: https://github.com/weihanglo/cargo/actions/runs/11601065840/job/32302943597

It is no harm to keep the v0 test file also fine with removing it from git.

### Additional information

Fixes rust-lang/cargo#14712


src/cargo/core/compiler/fingerprint/dep_info.rs
```diff
@@ -84,7 +84,7 @@ pub enum DepInfoPathType {
 ///   | len of key | key bytes | value exists? | len of value | value bytes |
```
epage, line 100: Should we also have a sanity-check value on "len of key"  We likely should never have a dep info key that is larger than 10k for instance

epage: (and ture for any length we deal with)

src/cargo/core/compiler/fingerprint/dep_info.rs
```diff
@@ -61,13 +64,26 @@ pub enum DepInfoPathType {
...
+///   Since [`EncodedDepInfo`] is just an optimization, to avoid adding any complexity,
+///   Cargo recognizes only one version of [`CURRENT_ENCODED_DEP_INFO_VERSION`].
+///   To support v0 (prior to checksum support in [`f4ca7390`]),
+///   which doesn't have this field, we do a hacky workaround. The current layout looks like
+///   ```text
+///   +----------------------------+---------+
+///   | [0x01 0x00 0x00 0x00 0xff] | version |
+///   +----------------------------+---------+
+///   ```
+///   The first 5 bytes will be interpreted as "one file trackced and invalid
```
epage, line 84: `trackced` -> `tracked`

src/cargo/core/compiler/fingerprint/dep_info.rs
```diff
@@ -129,6 +152,16 @@ impl EncodedDepInfo {
         }
         return Some(EncodedDepInfo { files, env });
 
+        fn read_version(bytes: &mut &[u8]) -> Option<u8> {
+            let _size = read_usize(bytes)?;
+            let path_type = read_u8(bytes)?;
+            if path_type != u8::MAX {
+                // Old depinfo. Give up parsing it.
+                return None;
+            }
```
epage, line 161: imo this function deserves comments, even if its pointing back to `EncodedDepInfo`s comment

src/cargo/core/compiler/fingerprint/dep_info.rs
weihanglo: Not sure. What if they do track that many environment variables?

An allocation error seems reasonable in that case. Cargo may crash in other places even it doesn't here.

src/cargo/core/compiler/fingerprint/dep_info.rs
epage, line 161: I wonder if this should be split out into a `read_header` or `read_magic_marker` or something.  I found it confusing when I saw the doc comment and then immediately saw `read_version`.  I was then going to comment in `read_version` about how it was behaving

src/cargo/core/compiler/fingerprint/dep_info.rs
```diff
@@ -667,4 +667,30 @@ mod encoded_dep_info {
...
+    // │000000d0│ 41 44 44 5f 42 49 4e 00 ┊ 08 00 00 00 54 46 5f 42 │ADD_BIN⋄┊•⋄⋄⋄TF_B│
+    // │000000e0│ 55 49 4c 44 00          ┊                         │UILD⋄   ┊        │
+    // └────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
+    #[test]
+    #[cfg_attr(not(target_os = "macos"), ignore = "OOM")]
+    fn parse_v0_fingerprint_dep_info() {
+        let data = include_bytes!("dep-lib-cargo_test_macro");
+        assert!(EncodedDepInfo::parse(&data[..]).is_none());
+    }
 }
```
epage: If this test isn't updated in the following commit, what is it we are expecting out of this test?

src/cargo/core/compiler/fingerprint/dep_info.rs
```diff
@@ -129,6 +152,16 @@ impl EncodedDepInfo {
         }
         return Some(EncodedDepInfo { files, env });
 
+        fn read_version(bytes: &mut &[u8]) -> Option<u8> {
+            let _size = read_usize(bytes)?;
+            let path_type = read_u8(bytes)?;
+            if path_type != u8::MAX {
+                // Old depinfo. Give up parsing it.
+                return None;
```
epage, line 160: Should we have a test for this case?

src/cargo/core/compiler/fingerprint/dep_info.rs
```diff
@@ -129,6 +155,18 @@ impl EncodedDepInfo {
         }
         return Some(EncodedDepInfo { files, env });
 
+        /// See [`EncodedDepInfo`] for why a magic marker exists.
+        fn read_magic_marker(bytes: &mut &[u8]) -> Option<()> {
+            let _size = read_usize(bytes)?;
+            let path_type = read_u8(bytes)?;
+            if path_type != u8::MAX {
+                // Old depinfo. Give up parsing it.
+                return None;
```
epage, line 164: nit: `return None;` could now be `None`

epage:
Feel free to `r=` me now or after resolving the nit

src/cargo/core/compiler/fingerprint/dep_info.rs
weihanglo: An alternative is <https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.try_with_capacity>.
(not yet stable though)

weihanglo:
@bors r=epage

While none of the test covers the OOM cases described in <https://github.com/rust-lang/cargo/issues/14712>, <https://github.com/weihanglo/cargo/pull/69> shows both failed and successful builds before and after this patch.

PR #14752, opened 2024-10-30
add unstable -Zroot-dir flag to configure the path from which rustc should be invoked

RalfJung:
This implements the proposal described [here](https://github.com/rust-lang/cargo/issues/9887#issuecomment-2273742881): we add a new flag, for now called `-Zroot-dir`, that configures the directory relative to which rustc is given the crate root filenames to build. (Files outside this directory are passed absolutely.)

This is necessary to be able to fix (no github don't close that issue yet) https://github.com/rust-lang/rust/issues/128726: in multi-workspace repositories that use scripts to manage a whole bunch of cargo invocations, currently the output cargo+rustc produce is often hard or even impossible to interpret for both human and machine consumption. This is because directories in the output are always relative to the workspace root, but when cargo is invoked many times for different workspaces, it is quite unclear what the workspace root is for the invocation that failed.

So I suggest we should have a new flag that the build script in such a repo can set to the consistent "root dir" that the user would recognize as such (e.g., the root of the rustc source tree), and all paths emitted by cargo and rustc should be relative to that directory.

I don't know all the places that cargo itself emits paths (if any), but this PR changes the way we invoke rustc to honor the new flag, so all paths emitted by rustc will be relative to the `-Zroot-dir`.

See https://github.com/rust-lang/rust/pull/132390 for the changes needed in rustc bootstrap to wire this up; together, that suffices to finally properly show errors in RA for all parts of the rustc src tree. :)

### DIFF
--- a/src/cargo/core/features.rs
+++ b/src/cargo/core/features.rs
@@ -121,6 +121,7 @@
 use std::collections::BTreeSet;
 use std::env;
 use std::fmt::{self, Write};
+use std::path::PathBuf;
 use std::str::FromStr;
 
 use anyhow::{bail, Error};
@@ -783,6 +784,7 @@ unstable_cli_options!(
     profile_rustflags: bool = ("Enable the `rustflags` option in profiles in .cargo/config.toml file"),
     public_dependency: bool = ("Respect a dependency's `public` field in Cargo.toml to control public/private dependencies"),
     publish_timeout: bool = ("Enable the `publish.timeout` key in .cargo/config.toml file"),
+    root_dir: Option<PathBuf> = ("Set the root directory relative to which paths are printed (defaults to workspace root)"),
     rustdoc_map: bool = ("Allow passing external documentation mappings to rustdoc"),
     rustdoc_scrape_examples: bool = ("Allows Rustdoc to scrape code examples from reverse-dependencies"),
     script: bool = ("Enable support for single-file, `.rs` packages"),
@@ -1287,6 +1289,7 @@ impl CliUnstable {
             "profile-rustflags" => self.profile_rustflags = parse_empty(k, v)?,
             "trim-paths" => self.trim_paths = parse_empty(k, v)?,
             "publish-timeout" => self.publish_timeout = parse_empty(k, v)?,
+            "root-dir" => self.root_dir = v.map(|v| v.into()),
             "rustdoc-map" => self.rustdoc_map = parse_empty(k, v)?,
             "rustdoc-scrape-examples" => self.rustdoc_scrape_examples = parse_empty(k, v)?,
             "separate-nightlies" => self.separate_nightlies = parse_empty(k, v)?,
--- a/tests/testsuite/directory.rs
+++ b/tests/testsuite/directory.rs
@@ -802,13 +802,17 @@ fn root_dir_diagnostics() {
         .file("ws_root/src/lib.rs", "invalid;")
         .build();
 
+    // Crucially, the rustc error message below says `ws_root/...`, i.e.
+    // it is relative to our fake home, not to the workspace root.
     p.cargo("check")
+        .arg("-Zroot-dir=.")
         .arg("--manifest-path=ws_root/Cargo.toml")
+        .masquerade_as_nightly_cargo(&["-Zroot-dir"])
         .with_status(101)
         .with_stderr_data(str![[r#"
 [CHECKING] foo v0.1.0 ([ROOT]/ws_root)
 [ERROR] [..]
- --> src/lib.rs:1:8
+ --> ws_root/src/lib.rs:1:8
   |
 1 | invalid;
   | [..]
@@ -839,10 +843,23 @@ fn root_dir_file_macro() {
         )
         .build();
 
+    // Crucially, the path is relative to our fake home, not to the workspace root.
     p.cargo("run")
+        .arg("-Zroot-dir=.")
         .arg("--manifest-path=ws_root/Cargo.toml")
+        .masquerade_as_nightly_cargo(&["-Zroot-dir"])
         .with_stdout_data(str![[r#"
-src/main.rs
+ws_root/src/main.rs
+
+"#]])
+        .run();
+    // Try again with an absolute path for `root-dir`.
+    p.cargo("run")
+        .arg(format!("-Zroot-dir={}", p.root().display()))
+        .arg("--manifest-path=ws_root/Cargo.toml")
+        .masquerade_as_nightly_cargo(&["-Zroot-dir"])
+        .with_stdout_data(str![[r#"
+ws_root/src/main.rs
 
 "#]])
         .run();
--- a/tests/testsuite/directory.rs
+++ b/tests/testsuite/directory.rs
@@ -784,3 +784,66 @@ Caused by:
         .with_status(101)
         .run();
 }
+
+#[cargo_test]
+fn root_dir_diagnostics() {
+    let p = ProjectBuilder::new(paths::root())
+        .no_manifest() // we are placing it in a different dir
+        .file(
+            "ws_root/Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.1.0"
+                edition = "2015"
+                authors = []
+            "#,
+        )
+        .file("ws_root/src/lib.rs", "invalid;")
+        .build();
+
+    p.cargo("check")
+        .arg("--manifest-path=ws_root/Cargo.toml")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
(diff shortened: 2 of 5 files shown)


RalfJung:
@epage indicated [here](https://github.com/rust-lang/cargo/issues/9887#issuecomment-2274139604) that they'd be okay with unstable experimentation, so I hope this is sufficient from a procedural standpoint.

epage:
> @epage indicated https://github.com/rust-lang/cargo/issues/9887#issuecomment-2274139604 that they'd be okay with unstable experimentation, so I hope this is sufficient from a procedural standpoint.

Something that isn't clear from the PR that was in my post is that I specifically called out a feature that is intended to be unstable only.  Granted, I worry this will take the pressure off of fixing this for everyone, rather than a rustc-only fix.

src/cargo/core/features.rs
```diff
@@ -783,6 +784,7 @@ unstable_cli_options!(
     profile_rustflags: bool = ("Enable the `rustflags` option in profiles in .cargo/config.toml file"),
     public_dependency: bool = ("Respect a dependency's `public` field in Cargo.toml to control public/private dependencies"),
     publish_timeout: bool = ("Enable the `publish.timeout` key in .cargo/config.toml file"),
+    root_path: Option<PathBuf> = ("Set the root directory relative to which paths in diagnostics are printed (defaults to workspace root)"),
```
RalfJung, line 787: FWIW I am not sure if it's *just* diagnostics, or also affects other paths rustc emits, like `file!()`.

src/cargo/core/features.rs
```diff
@@ -1287,6 +1289,15 @@ impl CliUnstable {
...
             "trim-paths" => self.trim_paths = parse_empty(k, v)?,
             "publish-timeout" => self.publish_timeout = parse_empty(k, v)?,
+            "root-path" => self.root_path = v.map(|v| {
+                // Make the path absolute, if we can.
+                let v = PathBuf::from(v);
+                if v.is_absolute() {
+                    v
+                } else {
+                    v.canonicalize().unwrap_or(v)
+                }
```
epage, line 1299: Cargo generally does not use `canonicalize` (suprised we haven't banned it with clippy).  We should use `gctx.cwd()`. 

We also don't generally do `is_absolute` checks but always do the join, maybe with a `cargo_util::paths::normalize_path` call.

tests/testsuite/directory.rs
```diff
@@ -784,3 +784,41 @@ Caused by:
...
+        .arg("--manifest-path=ws_root/Cargo.toml")
+        .masquerade_as_nightly_cargo(&["-Zroot-path"])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[CHECKING] foo v0.1.0 ([ROOT]/ws_root)
+[ERROR] expected one of `!` or `::`, found `;`
+ --> ws_root/src/lib.rs:1:8
+  |
+1 | invalid;
+  |        ^ expected one of `!` or `::`
```
epage, line 818: can you elide this with a `...` wildcard to help reduce our dependence on rustc output which makes updating rustc more difficult?

src/cargo/core/features.rs
RalfJung: Yeah it does affect `file!()`, so I'll update the description

tests/testsuite/directory.rs
```diff
@@ -784,3 +784,41 @@ Caused by:
...
+    // it is relative to our fake home, not to the workspace root.
+    p.cargo("check")
+        .arg("-Zroot-path=.")
+        .arg("--manifest-path=ws_root/Cargo.toml")
+        .masquerade_as_nightly_cargo(&["-Zroot-path"])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[CHECKING] foo v0.1.0 ([ROOT]/ws_root)
+[ERROR] expected one of `!` or `::`, found `;`
+ --> ws_root/src/lib.rs:1:8
```
epage, line 815: Could you add this test in a previous commit, showing the output without `-Zroot-path=.` and passing?  Then the feature commit would only have the feature work, addition of `-Zroot-path=.`, and the test output update.  This makes it very clear how the output changes with this.

RalfJung:
> Something that isn't clear from the PR that was in my post is that I specifically called out a feature that is intended to be unstable only.

I think this will eventually be needed on stable: there's clearly a missing feature in cargo here, and I haven't seen a better solution that could replace this unstable flag. However, the most pressing need is fixing https://github.com/rust-lang/rust/issues/128726 so that one can finally get proper RA feedback for rustc library work again (I am holding off writing certain PRs as it's just annoyingly painful to work on the library right now) -- and an unstable-only flag is enough for that.

tests/testsuite/directory.rs
RalfJung: I'll add that to my todo list, but by dev time is extremely limited so no promises when I can get around to fine-tune the git history here.

tests/testsuite/directory.rs
RalfJung: Ah, that's how you do this. :)

src/cargo/core/features.rs
RalfJung: I was wondering about that since I recalled `canonicalize` behaving oddly on Windows... I thought I had seen a WIP new function for std that would behave better, but couldn't find it.

`normalize_path` is probably a good idea since we need the prefix to match exactly for `strip_prefix` to kick in.

RalfJung:
> bikeshed the name of the flag

I have now painted this in the `-Zroot-dir` color. That works best with the current flag description, I think:

"Set the root directory relative to which paths are printed (defaults to workspace root)"

RalfJung:
Seems like the path prefix stripping is not working on Windows... not quite sure how to debug this, without access to a Windows machine. :/

RalfJung:
*oops* I had forgotten to remove the `canonicalize`... let's hope that is enough for Windows.

epage:
> I think this will eventually be needed on stable: there's clearly a missing feature in cargo here 

*Something* will eventually be needed in stable. 
- We need to find a way to improve the default of what gets rendered
- If we make it further configurable from that default, it likely should live in config somehow (templated variable like we want to do with build/target dirs?)

RalfJung:
You seem to have a larger design space in mind here than I realized -- is that spelled out anywhere? https://github.com/rust-lang/cargo/issues/9887 doesn't have a lot of concrete ideas, the only ones I could find are
- what is being implemented here
- have cargo translate the output so rustc is always invoked from the workspace root but diagnostics are printed relative to the working dir cargo is invoked from

Anyway, as I said I am fine with this being unstable without a clear path to stabilization. But I think we have to do *something* to make rustc development more pleasant, and this is the most actionable path forward I can see. (Having cargo translate paths is quite complicated to get right and has some open design questions itself, like how cargo should know about paths embedded in the human-readable part of a diagnostic.)

I have done some git history surgery and I think it now has the commit structure you asked for.

src/cargo/core/features.rs
```diff
@@ -783,6 +784,7 @@ unstable_cli_options!(
     profile_rustflags: bool = ("Enable the `rustflags` option in profiles in .cargo/config.toml file"),
     public_dependency: bool = ("Respect a dependency's `public` field in Cargo.toml to control public/private dependencies"),
     publish_timeout: bool = ("Enable the `publish.timeout` key in .cargo/config.toml file"),
+    root_dir: Option<PathBuf> = ("Set the root directory relative to which paths are printed (defaults to workspace root)"),
```
epage, line 787: I'm sorry, I forgot to call out that we need an entry in https://doc.rust-lang.org/cargo/reference/unstable.html

You can find it in https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/unstable.md

epage:
> You seem to have a larger design space in mind here than I realized -- is that spelled out anywhere? https://github.com/rust-lang/cargo/issues/9887 doesn't have a lot of concrete ideas, the only ones I could find are

Not really.  I think it would be great for us to always show diagnostics relative to the current working directory.  How we do that is less important (re-render diagnostics, add a flag to rustc to re-parent paths, etc).

If there remain use cases after that, then we can explore something else.  People generally bring up a CLI for this but I think config is more relevant as I'd like us to reserve stuff for the CLI that is commonly used.  The templating part is an idea I came up with on the fly as config can be as static or dynamic as the user wants and having a template helps for if the user wants something specific relative to something else.

src/cargo/core/features.rs
RalfJung: I did my best at guessing the format and organization here, please have a look.

epage:
Thanks!

@bors r+

PR #14755, opened 2024-10-31
Fix --package <non-existent> --workspace

### LINKED ISSUE #12978: With package selection, --package <non-existent> --workspace doesn't warn or error
### Problem

Originally reported as https://github.com/nextest-rs/nextest/issues/1108.

With Cargo's package selection, this errors out:

```
$ cargo build --package non-existent
error: package ID specification `non-existent` did not match any packages
```

However, this does not error out:

```
$ cargo build --package non-existent --workspace
<build occurs...>
```

I think this should at least produce a warning and maybe eventually error out.

(I tried searching for issues a bit but I couldn't figure out the right keywords.)

### Steps

As above.

### Possible Solution(s)

Produce a warning on `--workspace` and `-p <non-existent>`. Maybe eventually turn it into an error

### Notes

_No response_

### Version

```text
% cargo +nightly version --verbose           
cargo 1.76.0-nightly (6790a5127 2023-11-10)
release: 1.76.0-nightly
commit-hash: 6790a5127895debec95c24aefaeb18e059270df3
commit-date: 2023-11-10
os: Pop!_OS 22.04 (jammy) [64-bit]
```

766974616c79:

Information is available in the following issue #12978

### Additional information

Closes #12978

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -688,7 +688,14 @@ Run `{cmd}` to see possible targets."
         workspace: Option<&Workspace<'_>>,
         profile_checking: ProfileChecking,
     ) -> CargoResult<CompileOptions> {
-        let spec = self.packages_from_flags()?;
+        let w = self.flag("workspace");
+        let mut all = w || self.flag("all");
+        let package = self._values_of("package");
+        if w && !package.is_empty() {
+            all = false;
+        }
+
+        let spec = Packages::from_flags(all, self._values_of("exclude"), package)?;
         let mut message_format = None;
         let default_json = MessageFormat::Json {
             short: false,


src/cargo/util/command_prelude.rs
weihanglo, line 1: We suggest a PR coming with some tests to show how the behavior changed. It is better if we can have a commit capturing the current behavior, followed by the fix and the test change with CI passing.

src/cargo/util/command_prelude.rs
```diff
@@ -688,7 +688,14 @@ Run `{cmd}` to see possible targets."
         workspace: Option<&Workspace<'_>>,
         profile_checking: ProfileChecking,
     ) -> CargoResult<CompileOptions> {
-        let spec = self.packages_from_flags()?;
+        let w = self.flag("workspace");
+        let mut all = w || self.flag("all");
+        let package = self._values_of("package");
+        if w && !package.is_empty() {
+            all = false;
+        }
```
weihanglo, line 696: This changes the behavior and is not even proposed in the original issue.

I would recommend starting with a warning first.

ehuss:
@rustbot author

Also, we recently updated our CI configuration, and this will need to be rebased on the latest master.

src/cargo/util/command_prelude.rs
MoskalykA: here I am again, sorry for my lack of seriousness, i have a small question. issuing an error when the user uses `--package <package> & --workspace / --all` would be good for this issue? I have the impression that the fact that the package doesn't exist doesn't change anything.

weihanglo: You can create a workspace and test it out. Before this patch, `cargo b -p foo --workspace` works without error. After this, it reports `foo` matching nothing. This should be a warning first, not a hard error.

PR #14756, opened 2024-10-31
fix: env table config can't trigger rebuild with `rerun-if-env-changed`.

### LINKED ISSUE #10358: `cargo:rerun-if-env-changed` doesn't work with `env` configuration

linyihai:

As #10358 said, `When a build.rs script emits cargo:rerun-if-env-changed, it is not re-run when the value of the specified variable is changed via the env configuration.`  

Fixes #10358 

Add test bofore fixing to reflect the issue, the next commit fixed it.

### Additional information

The PR is a sucessor of https://github.com/rust-lang/cargo/pull/14058, so the previous dicussion can be refer to it.

### DIFF
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -374,6 +374,7 @@ mod dirty_reason;
 
 use std::collections::hash_map::{Entry, HashMap};
 use std::env;
+use std::ffi::OsString;
 use std::fs;
 use std::fs::File;
 use std::hash::{self, Hash, Hasher};
@@ -509,7 +510,7 @@ pub fn prepare_target(
         // thunk we can invoke on a foreign thread to calculate this.
         let build_script_outputs = Arc::clone(&build_runner.build_script_outputs);
         let metadata = build_runner.get_run_build_script_metadata(unit);
-        let (gen_local, _overridden) = build_script_local_fingerprints(build_runner, unit);
+        let (gen_local, _overridden) = build_script_local_fingerprints(build_runner, unit)?;
         let output_path = build_runner.build_explicit_deps[unit]
             .build_script_output
             .clone();
@@ -801,15 +802,24 @@ pub enum StaleItem {
 
 impl LocalFingerprint {
     /// Read the environment variable of the given env `key`, and creates a new
-    /// [`LocalFingerprint::RerunIfEnvChanged`] for it.
+    /// [`LocalFingerprint::RerunIfEnvChanged`] for it. The `env_config` is used firstly
+    /// to check if the env var is set in the config system as some envs need to be overridden.
+    /// If not, it will fallback to `std::env::var`.
     ///
-    // TODO: This is allowed at this moment. Should figure out if it makes
-    // sense if permitting to read env from the config system.
+    // TODO: `std::env::var` is allowed at this moment. Should figure out
+    // if it makes sense if permitting to read env from the env snapshot.
     #[allow(clippy::disallowed_methods)]
-    fn from_env<K: AsRef<str>>(key: K) -> LocalFingerprint {
+    fn from_env<K: AsRef<str>>(
+        key: K,
+        env_config: &Arc<HashMap<String, OsString>>,
+    ) -> LocalFingerprint {
         let key = key.as_ref();
         let var = key.to_owned();
-        let val = env::var(key).ok();
+        let val = if let Some(val) = env_config.get(key) {
+            val.to_str().map(ToOwned::to_owned)
+        } else {
+            env::var(key).ok()
+        };
         LocalFingerprint::RerunIfEnvChanged { var, val }
     }
 
@@ -1577,7 +1587,7 @@ fn calculate_run_custom_build(
     // the build script this means we'll be watching files and env vars.
     // Otherwise if we haven't previously executed it we'll just start watching
     // the whole crate.
-    let (gen_local, overridden) = build_script_local_fingerprints(build_runner, unit);
+    let (gen_local, overridden) = build_script_local_fingerprints(build_runner, unit)?;
     let deps = &build_runner.build_explicit_deps[unit];
     let local = (gen_local)(
         deps,
@@ -1671,7 +1681,7 @@ See https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-change
 fn build_script_local_fingerprints(
     build_runner: &mut BuildRunner<'_, '_>,
     unit: &Unit,
-) -> (
+) -> CargoResult<(
     Box<
         dyn FnOnce(
                 &BuildDeps,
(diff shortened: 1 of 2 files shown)


src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -376,6 +377,7 @@ use anyhow::format_err;
 use anyhow::Context as _;
 use cargo_util::paths;
 use filetime::FileTime;
+use lazycell::LazyCell;
```
epage, line 380: imo we should not be adding any new uses of lazycell but should instead be trying to remove it in favor of OnceCell

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -796,14 +798,36 @@ pub enum StaleItem {
...
+        env_config: &Arc<HashMap<String, OsString>>,
+        env_config_insensitive: &Arc<LazyCell<HashMap<String, OsString>>>,
+    ) -> LocalFingerprint {
         let key = key.as_ref();
         let var = key.to_owned();
-        let val = env::var(key).ok();
+        let val = if let Some(val) = env::var(key).ok() {
+            Some(val)
+        } else {
+            match env_config.get(key) {
```
epage, line 812: which is supposed to have higher precedence?

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1701,6 +1725,7 @@ fn build_script_local_fingerprints(
     // obvious.
     let pkg_root = unit.pkg.root().to_path_buf();
     let target_dir = target_root(build_runner);
+    let env_config = Arc::clone(build_runner.bcx.gctx.env_config().unwrap());
```
epage, line 1728: Whats the assumption we are operating under that its safe to `unwrap`?

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -796,14 +798,36 @@ pub enum StaleItem {
...
+                    if cfg!(windows) {
+                        env_config_insensitive
+                            .borrow_with(|| {
+                                env_config
+                                    .iter()
+                                    .map(|(k, v)| (k.to_uppercase().clone(), v.clone()))
+                                    .collect::<HashMap<String, OsString>>()
+                            })
+                            .get(&key.to_uppercase())
+                            .and_then(|s| s.to_str().map(|s| s.to_string()))
```
epage, line 824: Please use a more specific conversion than `to_string` as it can mask incidental semantic conversions

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -796,14 +798,36 @@ pub enum StaleItem {
...
         let var = key.to_owned();
-        let val = env::var(key).ok();
+        let val = if let Some(val) = env::var(key).ok() {
+            Some(val)
+        } else {
+            match env_config.get(key) {
+                Some(value) => value.to_str().map(|s| s.to_string()),
+                None => {
+                    if cfg!(windows) {
+                        env_config_insensitive
```
epage, line 816: Rather than having to do all of this extra work, I wonder if we should make the env map itself case insensitive.  Looking at how `env_config()` is used, I think we have several places where we don't handle case sensitivity and maybe we should?

What we could do is define a type def for `EnvMap` where the key is `unicase::Ascii<String>` and have the global context return that.

@weihanglo thoughts?

weihanglo: That should be conditioned on platform, maybe only on Windows?

Cargo does that upfront for environment variables from configuration enviornment: https://github.com/rust-lang/cargo/blob/master/src/cargo/util/context/environment.rs

epage: Sorry, Yes, I meant conditioned on the platform (which is why the typedef would be there)

epage: Hmm, if we also have a normalization process, I wonder if we should apply that more generally

src/cargo/core/compiler/fingerprint/mod.rs
linyihai: I think `env_config` should have higer precedence as `env_config`  may contains the env that needs  force overwrite or that doesn't exist in the env snapshot

src/cargo/core/compiler/fingerprint/mod.rs
linyihai: Updated.

src/cargo/core/compiler/fingerprint/mod.rs
linyihai: I guessed you intention is explictly to use `str::to_string` instead ?

epage: For myself, I tend to prefer `to_owned` as that signals the intent

src/cargo/core/compiler/fingerprint/dep_info.rs
epage, line 1: >  refactor: env_config supports resolve case insensitive env. 

What was the intention with this last commit?

linyihai: Its intention is https://github.com/rust-lang/cargo/pull/14756#discussion_r1824555604.  Should I squash it into the previous fixing commit ?

epage: The intention behind that is to abstract away the platform-specific logic and make it easier to do the right thing across cargo.  I don't feel the current implementation meets that need. Granted, this might be a bigger change, so depending on  the direction we should take with it, maybe its better to defer out of this

linyihai: Ok,  I leave this as a future PR.

linyihai:
I removed the code that makes `env_config` case insensitivity in Windows.  Resolving env_config's case sensitivity due to operating system platform differences is far more complicated than fixing this issue, so I left it to a possible future PR and made this PR easier to merge.

weihanglo (approved):
This all makes sense. Thank you.

weihanglo (approved):

PR #14767, opened 2024-11-01
Fix building count off-by-one error

griccardos:
When building a project, the count of the current `Building` item is always one less than it should be.
'Building' implies the current item being built.
One would expect the **last** item being built to be for example 330/330 and NOT 329/330

This is because we are currently calculating the `finished / total`
Instead we should be calculating `current / total`
The current will always be 1 more than the finished up until the last item.

### Before
note `20/21`

cargo_old1

### After
note `21/21`

cargo_new1

### DIFF
--- a/src/cargo/core/compiler/job_queue/mod.rs
+++ b/src/cargo/core/compiler/job_queue/mod.rs
@@ -887,7 +887,7 @@ impl<'gctx> DrainState<'gctx> {
             .map(|u| self.name_for_progress(u))
             .collect::<Vec<_>>();
         let _ = self.progress.tick_now(
-            self.finished,
+            self.finished + 1,
             self.total_units,
             &format!(": {}", active_names.join(", ")),
         );


epage:
There are multiple ways to read it.  The way I read it and I think its intended to be read is "We are building, here is how much is done".  It sounds like you are reading it as "We building number N".  However, thats not quite accurate because we build things in parallel.  Its not "We are building 1-N" because that includes finished items.

As I believe this is working as intended, I'm going to go ahead and close.  If you have a reason to further discuss this, we encourage having these discussions in Issues before working on PRs, see https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo

griccardos:
Thanks @epage for reviewing, I appreciate the work you do. 
I see how you read it, although for me, and I assume many others, it says 'Building', so I take it literally, that the item it says it's building should include the current item. This generally doesn't matter except for the last item which I want to know is being built, but it never is because it never gets to N/N, although it's obvious by the name of the package being there at the end. 
Since this is merely cosmetic, I don't think you need waste more time on it, and I'll try read it differently.

PR #14777, opened 2024-11-03
Enable transfer feature in triagebot

Urgau:
This PR enables the `transfer` feature from triagebot, it permits issue transfers within the org.

Documentation at: https://forge.rust-lang.org/triagebot/transfer.html

It would have been useful in https://github.com/rust-lang/rust/issues/132572#issuecomment-2453566084

r? @ehuss

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -26,6 +26,8 @@ label = "O-windows"
 
 [shortcut]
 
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks!

@bors r+

PR #14781, opened 2024-11-04
test: Update some emaining unordered tests to snapbox

epage:

This is part of #14039

This leaves `global_cache_tracker.rs` as it requires some more thinking.

As for the flakiness in `freshness.rs`  that was seen in #14161, `compare.rs` would prioritize expected lines according to their length (assuming its more specific).  Currently, snapbox prioritizes according to the line order.  So we just need to put the proc-macro line before the other one to ensure it gets precedence.


weihanglo (approved):

weihanglo:
Thanks!

@bors r+

PR #14785, opened 2024-11-05
test(gc): Update remaining unordered tests to snapbox

epage:

This gets rid of the last unordered tests and removes the functions from `cargo-test-support` as part of #14039

Some tests are being less specific than they were before but in talking to ehuss, it sounded like that was ok.


ehuss:
Thanks!

@bors r+

PR #11359, opened 2022-11-10
Add `home` crate to cargo crates

LucioFranco:
### What does this PR try to resolve?

The `home` crate is the crate used by both `rustup` and `cargo` to define their home directories. The crate is quite small but plays an important role in gluing rust tooling together. This PR adds this crate the `crates` folder within cargo to allow the cargo team and the rust project to maintain this small crate.

### Additional information

I've also added both `rust-lang-owners` and `ehuss` as owners of the crate. Once, merged I will archive the old repo. 

Reference zulip conversation: https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Moving.20the.20home.20crate.20into.20rust-lang.20and.20under.20cargo/near/299975263

cc @weihanglo @ehuss

PR #14786, opened 2024-11-05
Migrate build-rs to the Cargo repo

### LINKED ISSUE #12432: Official API for build scripts

epage:

Fixes #12432

This pulls in https://github.com/cad97/build-rs at eb389d1104dae2dc2eb4eafa9d6e821a7dcb45a7 with the following changes:
- `Cargo.toml` metadata
- Removal of `.github`, `.gitignore`, `Cargo.lock`

We'll need to integrate `test-lib` into our processes but that seemed more invasive, so I wanted to leave that for a future PR.

### Additional information

Infra changes are being coordinated in https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Transfering.20.60build-rs.60.20crate.20to.20rust-lang/near/480779960

Context: per [Cargo's charter](https://doc.crates.io/contrib/team.html#decision-process), we approved this transfer in an [FCP](https://github.com/rust-lang/cargo/issues/12432#issuecomment-2389528102).

### DIFF
--- a/crates/build-rs/Cargo.toml
+++ b/crates/build-rs/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "build-rs"
+version = "0.2.0"
+edition = "2021"
+rust-version = "1.76"
+publish = false
+
+[dependencies]
+unicode-ident = "1.0.13"
+
+[features]
+
+## Experimental API. This feature flag is **NOT** semver stable.
+unstable = []
--- a/crates/build-rs/Cargo.toml
+++ b/crates/build-rs/Cargo.toml
@@ -1,14 +1,16 @@
 [package]
 name = "build-rs"
 version = "0.2.0"
-edition = "2021"
-rust-version = "1.76"
-publish = false
-
-[dependencies]
-unicode-ident = "1.0.13"
+rust-version.workspace = true
+edition.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+description = "API for writing Cargo `build.rs` files"
 
 [features]
-
 ## Experimental API. This feature flag is **NOT** semver stable.
 unstable = []
+
+[dependencies]
+unicode-ident.workspace = true
--- a/crates/build-rs/README.md
+++ b/crates/build-rs/README.md
@@ -1,0 +1,2 @@
+> This crate is maintained by the Cargo team for use by the wider
+> ecosystem. This crate follows semver compatibility for its APIs.
--- a/crates/build-rs/src/output.rs
+++ b/crates/build-rs/src/output.rs
@@ -1,0 +1,408 @@
+//! Outputs from the build script to the build system.
+//!
+//! This crate assumes that stdout is at a new line whenever an output directive
+//! is called. Printing to stdout without a terminating newline (i.e. not using
+//! [`println!`]) may lead to surprising behavior.
+//!
+//! Reference: <https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script>
+
+use crate::{
+    allow_use,
+    ident::{is_ascii_ident, is_ident},
+};
+use std::{ffi::OsStr, fmt::Display, fmt::Write, path::Path, str};
+
+fn emit(directive: &str, value: impl Display) {
+    if allow_use::double_colon_directives() {
+        println!("cargo::{}={}", directive, value);
+    } else {
+        println!("cargo:{}={}", directive, value);
+    }
+}
+
+/// The `rerun-if-changed` instruction tells Cargo to re-run the build script if the
+/// file at the given path has changed. Currently, Cargo only uses the filesystem
+/// last-modified “mtime” timestamp to determine if the file has changed. It
+/// compares against an internal cached timestamp of when the build script last ran.
+///
+/// If the path points to a directory, it will scan the entire directory for any
+/// modifications.
+///
+/// If the build script inherently does not need to re-run under any circumstance,
+/// then calling `rerun_if_changed("build.rs")` is a simple way to prevent it from
+/// being re-run (otherwise, the default if no `rerun-if` instructions are emitted
+/// is to scan the entire package directory for changes). Cargo automatically
+/// handles whether or not the script itself needs to be recompiled, and of course
+/// the script will be re-run after it has been recompiled. Otherwise, specifying
+/// `build.rs` is redundant and unnecessary.
+#[track_caller]
+pub fn rerun_if_changed(path: impl AsRef<Path>) {
+    let Some(path) = path.as_ref().to_str() else {
+        panic!("cannot emit rerun-if-changed: path is not UTF-8");
+    };
+    if path.contains('\n') {
+        panic!("cannot emit rerun-if-changed: path contains newline");
+    }
+    emit("rerun-if-changed", path);
+}
+
+/// The `rerun-if-env-changed` instruction tells Cargo to re-run the build script
+/// if the value of an environment variable of the given name has changed.
+///
+/// Note that the environment variables here are intended for global environment
+/// variables like `CC` and such, it is not possible to use this for environment
+/// variables like `TARGET` that [Cargo sets for build scripts][build-env]. The
+/// environment variables in use are those received by cargo invocations, not
+/// those received by the executable of the build script.
+///
+/// [build-env]: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts
+#[track_caller]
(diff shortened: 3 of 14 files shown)


crates/build-rs/Cargo.toml
weihanglo, line 1: https://github.com/rust-lang/cargo/pull/11359#issuecomment-1310921393

should we import the full history of that repo, like what we did for `home` crate?

crates/build-rs/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "build-rs"
+version = "0.2.0"
+rust-version.workspace = true
```
weihanglo, line 4: Just want to call out that this MSRV seems way too high for this crate.
We may want to lower this in the future.

crates/build-rs/Cargo.toml
CAD97: I rewrote basically everything on the main branch from v0.1 anyway, so it doesn't particularly matter to me either way.

CAD97:
FYI I didn't see this had been created already and ~~finally~~ [added some cleanup and items I had missed in the v0.2 refactor](https://github.com/CAD97/build-rs/commit/7010eb5e6ec9923877dae6c521472b0cb43ea739) that will want to be pulled into this as well.

build-rs-test-lib is just a package that hosts a buildscript that can smoke test invoking the API in a build script context; I don't *think* it should need anything special.

The changes from v0.1 to  v0.2 are a simplification of implementation, removal of the glob reexports, and basic integration of how the crate can do whatever version smoothing is considered desirable.

crates/build-rs/Cargo.toml
epage, line 1: Need to re-sync with the build-rs repo before we merge

crates/build-rs/src/output.rs
```diff
@@ -0,0 +1,408 @@
...
+/// This form is for keys without an expected value, such as `cfg(name)`.
+///
+/// It is recommended to group the `rustc_check_cfg` and `rustc_cfg` calls as
+/// closely as possible in order to avoid typos, missing check_cfg, stale cfgs,
+/// and other mistakes.
+///
+/// [`unexpected_cfgs`]: https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#unexpected-cfgs
+#[doc = msrv!("1.80")]
+#[track_caller]
+pub fn rustc_check_cfgs(keys: &[&str]) {
```
CAD97, line 291: `rustc-check-cfg` is the most "interesting" API to translate, and I'm the least happy with the way it's exposed here. 

The split between `rustc_cfg` and `rustc_cfg_value` isn't *great*, but it's at least straightforward to translate optional arguments to multiple function names. 

I'm like 20% tempted to expose this as a macro so that the directive syntax for 1+ keys and 0+ values can be directly replicated here. But it's been an explicit goal for this bridge to expose *simple*, straightforward API, and macro tricks are anything but. 

Sketch, without the necessary macro trickery:

```rust
macro_rules! rustc_check_cfg {
    ($($key:expr),+) => { rustc_check_cfg(&[$($key),+]) };
    ($($key:expr),+, values()) => { rustc_check_cfg_value(&[$($key),+], &[]) };
    ($($key:expr),+, values(none())) => { rustc_check_cfg(&[$($key),+]) };
    ($($key:expr),+, values(any())) => { rustc_check_cfg_any(&[$($key),+]) };
    ($($key:expr),+, values($($value:expr),*)) => { rustc_check_cfg_value(&[$($key),+], &[$($value),*) };
}
```

Other options: 
- Trait for accepting `&str` or `&[&str]`. Still a bespoke API generalization that makes for worse type errors/inference.
- Builder style API. Requires either `.emit()` or printing on drop.

crates/build-rs/src/output.rs
```diff
@@ -0,0 +1,408 @@
...
+/// [cargo-env]: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates
+#[track_caller]
+pub fn rustc_env(key: &str, value: &str) {
+    if key.contains(['=', '\n']) {
+        panic!("cannot emit rustc-env: invalid key {key:?}");
+    }
+    if value.contains('\n') {
+        panic!("cannot emit rustc-env: invalid value {value:?}");
+    }
+    emit("rustc-env", format_args!("{key}={value}"));
```
CAD97, line 361: Need to check/test that the various directives apply quoting/escaping correctly

CAD97 (approved):
One relevant comment on API stability of the API for unstable cfg directives, and the rest is mostly just me boggling at code I wrote weirdly while transcribing my understanding of the buildscript protocol into code, and don't need to be addressed here.

Plus my explicit public :+1: to making build-rs an officially produced cargo artifact.

crates/build-rs/src/output.rs
```diff
@@ -0,0 +1,408 @@
...
+#[track_caller]
+pub fn rustc_cfg(key: &str) {
+    if !is_ident(key) {
+        panic!("cannot emit rustc-cfg: invalid key {key:?}");
+    }
+    emit("rustc-cfg", key);
+}
+
+/// Like [`rustc_cfg`], but with the value specified separately. To replace the
+/// less convenient `rustc_cfg(r#"my_component="foo""#)`, you can instead use
```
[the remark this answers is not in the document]
epage: While code feedback is good, we should probably decouple it from this PR and handle it in follow ups.

crates/build-rs/src/output.rs
```diff
@@ -0,0 +1,408 @@
...
+/// like `rustc_cfg(r#"my_component="foo""#)`. The key should be a Rust identifier,
+/// the value should be a string.
+///
+/// [cfg]: https://doc.rust-lang.org/rustc/command-line-arguments.html#option-cfg
+/// [Cargo features]: https://doc.rust-lang.org/cargo/reference/features.html
+#[track_caller]
+pub fn rustc_cfg(key: &str) {
+    if !is_ident(key) {
+        panic!("cannot emit rustc-cfg: invalid key {key:?}");
+    }
```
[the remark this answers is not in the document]
epage: While code feedback is good, we should probably decouple it from this PR and handle it in follow ups.

crates/build-rs/src/output.rs
epage: While code feedback is good, we should probably decouple it from this PR and handle it in follow ups.

crates/build-rs/src/output.rs
epage: While code feedback is good, we should probably decouple it from this PR and handle it in follow ups.

crates/build-rs/README.md
```diff
@@ -0,0 +1,2 @@
+> This crate is maintained by the Cargo team for use by the wider
+> ecosystem. This crate follows semver compatibility for its APIs.
```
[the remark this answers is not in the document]
epage: My assumption is that its fine as that feature is documented as unstable.   I'm not even sure why that line is encouraged in the policy.  Pretty much every lib is assumed to follow semver.

weihanglo (approved):
Looks good to me. Some areas may need to polish, but as Ed said we should defer them to separate PRs.

weihanglo:
Since the FCP was completed. Going to merge this now.

@bors r+

PR #14792, opened 2024-11-07
fix(toml): Update frontmatter parser for RFC 3503

epage:

We removed rejected syntax in #13861 haven't update frontmatter parsing to what [RFC 3503](https://rust-lang.github.io/rfcs/3503-frontmatter.html#reference-level-explanation) says.

This adds tests and fixes various cases to ensure we correctly detect the frontmatter and the infostring.

This is part of #12207

### DIFF
--- a/src/cargo/util/toml/embedded.rs
+++ b/src/cargo/util/toml/embedded.rs
@@ -267,6 +267,7 @@ fn split_source(input: &str) -> CargoResult<Source<'_>> {
 
 #[cfg(test)]
 mod test_expand {
+    use snapbox::assert_data_eq;
     use snapbox::str;
 
     use super::*;
@@ -284,7 +285,7 @@ mod test_expand {
 
     #[test]
     fn test_default() {
-        snapbox::assert_data_eq!(
+        assert_data_eq!(
             si!(r#"fn main() {}"#),
             str![[r#"
 [[bin]]
@@ -312,7 +313,7 @@ strip = true
 
     #[test]
     fn test_dependencies() {
-        snapbox::assert_data_eq!(
+        assert_data_eq!(
             si!(r#"---cargo
 [dependencies]
 time="0.1.25"
@@ -348,7 +349,7 @@ strip = true
 
     #[test]
     fn test_no_infostring() {
-        snapbox::assert_data_eq!(
+        assert_data_eq!(
             si!(r#"---
 [dependencies]
 time="0.1.25"
--- a/src/cargo/util/toml/embedded.rs
+++ b/src/cargo/util/toml/embedded.rs
@@ -228,8 +228,17 @@ fn split_source(input: &str) -> CargoResult<Source<'_>> {
 
     const FENCE_CHAR: char = '-';
 
-    let fence_end = source
-        .content
+    let mut trimmed_content = source.content;
+    while !trimmed_content.is_empty() {
+        let c = trimmed_content;
+        let c = c.trim_start_matches([' ', '\t']);
+        let c = c.trim_start_matches(['\r', '\n']);
+        if c == trimmed_content {
+            break;
+        }
+        trimmed_content = c;
+    }
+    let fence_end = trimmed_content
         .char_indices()
         .find_map(|(i, c)| (c != FENCE_CHAR).then_some(i))
         .unwrap_or(source.content.len());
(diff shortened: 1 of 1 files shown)


src/cargo/util/toml/embedded.rs
```diff
@@ -267,65 +281,363 @@ fn split_source(input: &str) -> CargoResult<Source<'_>> {
...
-            let cwd = std::env::current_dir().unwrap();
-            let home = home::cargo_home_with_cwd(&cwd).unwrap();
-            let gctx = GlobalContext::new(shell, cwd, home);
-            expand_manifest($i, std::path::Path::new("/home/me/test.rs"), &gctx)
-                .unwrap_or_else(|err| panic!("{}", err))
-        }};
+    #[test]
+    fn split_default() {
+        snapbox::assert_data_eq!(
+            split_source(r#"fn main() {}"#).to_debug(),
```
ehuss, line 292: I'm a little uncomfortable checking the output of the debug implementation. It isn't stable, and would prevent updating the formatting in rust-lang/rust since it could cause these tests to fail. Would it be possible to check this in a way that doesn't depend on that?

epage: One route I went was asserting the individual fields which gives good failures and allows some snapshotting but not very ergonomic
```rust
    #[track_caller]
    fn assert_source(
        source: &str,
        shebang: Option<Inline>,
        info: Option<Inline>,
        frontmatter: Option<Inline>,
        content: Inline,
    ) {
        let actual = match split_source(source) {
            Ok(actual) => actual,
            Err(err) => panic!("unexpected err: {err}"),
        };
        assert_some_data_eq(actual.shebang, shebang);
        assert_some_data_eq(actual.info, info);
        assert_some_data_eq(actual.frontmatter, frontmatter);
        assert_data_eq!(actual.content, content);
    }

    #[track_caller]
    fn assert_some_data_eq(actual: Option<impl IntoData>, expected: Option<impl IntoData>) {
        match (actual, expected) {
            (Some(actual), Some(expected)) => {
                assert_data_eq!(actual, expected);
            }
            (Some(actual), None) => {
                let actual = actual.into_data();
                panic!(
                    "expected `None`, got:
```
{actual}
```
"
                );
            }
            (None, Some(actual)) => {
                let actual = actual.into_data();
                panic!(
                    "got `None`, expected:
```
{actual}
```
"
                );
            }
            (None, None) => {}
        }
    }

    #[track_caller]
    fn assert_err(
        result: Result<impl std::fmt::Debug, impl std::fmt::Display>,
        err: impl IntoData,
    ) {
        match result {
            Ok(d) => panic!("unexpected Ok({d:#?})"),
            Err(actual) => snapbox::assert_data_eq!(actual.to_string(), err),
        }
    }

    #[test]
    fn split_dependencies() {
        assert_source(
            r#"---
[dependencies]
time="0.1.25"
---
fn main() {}
"#,
            None,
            None,
            Some(str![[r#"
[dependencies]
time="0.1.25"

"#]]),
            str![[r#"
fn main() {}

"#]],
        );
    }

```

epage: A custom render for tests is likely the best hybrid solution.

epage: Is this better?  It does rely on the debug for strings but that is relative stable
```rust
    #[track_caller]
    fn assert_source(source: &str, expected: impl IntoData) {
        use std::fmt::Write as _;

        let actual = match split_source(source) {
            Ok(actual) => actual,
            Err(err) => panic!("unexpected err: {err}"),
        };

        let mut rendered = String::new();
        write_optional_field(&mut rendered, "shebang", actual.shebang);
        write_optional_field(&mut rendered, "info", actual.info);
        write_optional_field(&mut rendered, "frontmatter", actual.frontmatter);
        writeln!(&mut rendered, "content: {}", actual.content).unwrap();
        assert_data_eq!(rendered, expected.raw());
    }

    fn write_optional_field(writer: &mut dyn std::fmt::Write, field: &str, value: Option<&str>) {
        if let Some(value) = value {
            writeln!(writer, "{field}: {value:?}").unwrap();
        } else {
            writeln!(writer, "{field}: None").unwrap();
        }
    }

    #[track_caller]
    fn assert_err(
        result: Result<impl std::fmt::Debug, impl std::fmt::Display>,
        err: impl IntoData,
    ) {
        match result {
            Ok(d) => panic!("unexpected Ok({d:#?})"),
            Err(actual) => snapbox::assert_data_eq!(actual.to_string(), err),
        }
    }

    #[test]
    fn split_dependencies() {
        assert_source(
            r#"---
[dependencies]
time="0.1.25"
---
fn main() {}
"#,
            str![[r#"
shebang: None
info: None
frontmatter: "[dependencies]\ntime=\"0.1.25\"\n"
content: fn main() {}


"#]],
        );
    }

```

PR #14793, opened 2024-11-07
docs(test):  Document Execs assertions based on port effort 

### LINKED ISSUE #14039: Port from bespoke assertions of snapbox

epage:

Closes #14039

A lot of this was pulled from #14039, common assertions people need to write, and from seeing people not noticing features that exist.

I'm leaving behind the `contains` assertions, rather than finding a way to make them work with snapbox.  Thankfully, snapbox is designed to let people build their own thing (see #14790).  I considered reusing snapbox's `[..]` matches but the code here is pretty minimal, the logic is similar enough, and I don't have a great abstraction for snapbox for it.  If there was more of a need, I'd make something work in snapbox.  As such, the `contains` assertions are no longer deprecated.

While doing this last pass through, I did some polish on the code as well.

### DIFF
--- a/crates/cargo-test-support/src/lib.rs
+++ b/crates/cargo-test-support/src/lib.rs
@@ -816,7 +816,6 @@ impl Execs {
     /// - `expected` can end up being ambiguous, causing the assertion to succeed when it should fail
     ///
     /// </div>
-    #[deprecated(note = "replaced with `Execs::with_stdout_data(expected)`")]
     pub fn with_stdout_contains<S: ToString>(&mut self, expected: S) -> &mut Self {
         self.expect_stdout_contains.push(expected.to_string());
         self
@@ -834,7 +833,6 @@ impl Execs {
     /// - `expected` can end up being ambiguous, causing the assertion to succeed when it should fail
     ///
     /// </div>
-    #[deprecated(note = "replaced with `Execs::with_stderr_data(expected)`")]
     pub fn with_stderr_contains<S: ToString>(&mut self, expected: S) -> &mut Self {
         self.expect_stderr_contains.push(expected.to_string());
         self
@@ -857,7 +855,6 @@ impl Execs {
     /// [`Execs::with_stdout_contains`].
     ///
     /// </div>
-    #[deprecated]
     pub fn with_stdout_does_not_contain<S: ToString>(&mut self, expected: S) -> &mut Self {
         self.expect_stdout_not_contains.push(expected.to_string());
         self
@@ -879,7 +876,6 @@ impl Execs {
     /// with [`Execs::with_stderr_line_without`].
     ///
     /// </div>
-    #[deprecated]
     pub fn with_stderr_does_not_contain<S: ToString>(&mut self, expected: S) -> &mut Self {
         self.expect_stderr_not_contains.push(expected.to_string());
         self
@@ -916,7 +912,6 @@ impl Execs {
     /// This will check that a build line includes `-C opt-level=3` but does
     /// not contain `-C debuginfo` or `-C incremental`.
     ///
-    #[deprecated]
     pub fn with_stderr_line_without<S: ToString>(
         &mut self,
         with: &[S],
--- a/crates/cargo-test-support/src/lib.rs
+++ b/crates/cargo-test-support/src/lib.rs
@@ -667,6 +667,57 @@ impl Execs {
     /// Verifies that stdout is equal to the given lines.
     ///
     /// See [`compare::assert_e2e`] for assertion details.
+    ///
+    /// <div class="warning">
+    ///
+    /// Prefer passing in [`str!`] for `expected` to get snapshot updating.
+    ///
+    /// If `format!` is needed for content that changes from run to run that you don't care about,
+    /// consider whether you could have [`compare::assert_e2e`] redact the content.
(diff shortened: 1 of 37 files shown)


crates/cargo-test-support/src/lib.rs
```diff
@@ -706,6 +808,14 @@ impl Execs {
     /// its output.
     ///
     /// See [`compare`] for supported patterns.
+    ///
+    /// <div class="warning">
```
weihanglo, line 812: Cool! I didn't know rustdoc has special CSS classes until today.

weihanglo (approved):

crates/cargo-test-support/src/lib.rs
```diff
@@ -737,12 +864,18 @@ impl Execs {
     ///
     /// See [`compare`] for supported patterns.
     ///
-    /// Care should be taken when using this method because there is a
-    /// limitless number of possible things that *won't* appear. A typo means
-    /// your test will pass without verifying the correct behavior. If
-    /// possible, write the test first so that it fails, and then implement
-    /// your fix/feature to make it pass.
-    #[deprecated]
```
weihanglo, line 745: Not going to block this, but still prefer them being deprecated, as people have less change to misuse.

epage: If you have other ideas, I'm open

crates/cargo-test-support/src/lib.rs
epage: I think they actually have something newer but its not documented.  I'll ping them about it.

PR #14796, opened 2024-11-08
chore(ci): Check for clippy `correctness`

epage:

The fact that we don't check for `derive_ord_xor_partial_ord ` almost bit us in #14663.  If we used clippy's default lint levels, this would have been caught automatically.

Instead of just enabling this one lint, I figured it'd be good to audit
all of the `deny` by default lints.
That is long enough that I was concerned about maintaining it (or
bikeshedding which to enable or disable).
All `deny` by default lints are `correctness` lints and I figure we
could just enable the group.

We normally opt-in to individual clippy lints.
From what I remember of that conversation, it mostly stems from how
liberal clippy is with making a lint `warn` by default.
It also makes an unpinned CI more brittle.
I figured clippy is more conservative about `deny` by default lints
and slower to add them that this is unlikely to be a problem.

As for what existing problems this found,
- Some missing serde functions that would be useful for formats besides toml (since `toml` never uses borrowed strings)
- Code that behaves differently than the syntax says (a 0-1 iteration loop)
- A redundant assignment (wasn't even removing `mut`ness

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -121,7 +121,8 @@ rust_2018_idioms = "warn" # TODO: could this be removed?
 private_intra_doc_links = "allow"
 
 [workspace.lints.clippy]
-all = { level = "allow", priority = -1 }
+all = { level = "allow", priority = -2 }
+correctness = { level = "warn", priority = -1 }
 dbg_macro = "warn"
 disallowed_methods = "warn"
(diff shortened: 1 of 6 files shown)


ehuss (approved):
Thanks! The team discussed and agreed to move forward with this.

Issue #5478, opened 2018-05-04
Patching dependencies does not work if it's for the same location but a different branch


sdroege:
E.g.
```toml
[patch.'https://github.com/sdroege/gobject-subclass']
gobject-subclass = { git = 'https://github.com/sdroege/gobject-subclass', branch = 'refactoring' }
```

This gives

```
error: failed to resolve patches for `https://github.com/sdroege/gobject-subclass`

Caused by:
  patch for `gobject-subclass` in `https://github.com/sdroege/gobject-subclass` points to the same source, but patches must point to different sources
```

alexcrichton:
Ah yeah this is sort of an unfortunate side-effect of the implementation of `patch` in that it only looks at URLs, not at the full source. That is, you can't even say you only want to replace the source where the branch is `master`, instead a `[patch]` applies to *all* git sources with the same url (regardless of rev/branch/etc). That means that this is currently replacing itself as well, which is disallowed.

jplatte:
For GitHub repositories, this can be worked around by `[patch]`ing https://github.com/user/repo to https://github.com/user/repo.git (or the other way around).

est31:
> For GitHub repositories, one can work around this by [patch]ing https://github.com/user/repo to https://github.com/user/repo.git (or the other way around).

Nice workaround hack! Further ways to change the URL string include adding `/` to the url like https://github.com////user///repo . This method scales to unlimited amounts of different urls for the same repo.

brenzi:
The above workaround doesn't work for us, because what we patch has dependencies as well which need to be patched by the workspace patch. However, as the url now changed, it won't be patched. 
I thought I could just add another level of patching:

```
[patch.'https://github.com/scs/substrate']
primitives = { git = "https://github.com/scs//substrate", package = 'substrate-primitives', tag = "substraTEE-M5.1" }

[patch.'https://github.com/scs//substrate']
primitives = { git = "https://github.com/scs///substrate", package = 'substrate-primitives', tag = "substraTEE-M5.1" }
```
but that won't work:
```
cargo update
    Updating git repository `https://github.com/scs/ed25519-dalek.git`
    Updating git repository `https://github.com/scs///substrate`
warning: spurious network error (2 tries remaining): unexpected HTTP status code: 404; class=Net (12)
warning: spurious network error (1 tries remaining): unexpected HTTP status code: 404; class=Net (12)
error: failed to resolve patches for `https://github.com/scs//substrate`

Caused by:
[...]

PR #14800, opened 2024-11-08
simplify SourceID Hash

Eh2406:

Despite being interned `SourceId::Eq` is not a `ptr::eq`. Which in turn is because `SourceId`s concept of identity is a complete mess. The code relies on having to IDs that are `Eq` but do not have the same values for their fields.

As one measure of this `SourceId` has an `impl Hash` which does something different from `fn full_hash` and `fn stable_hash`. Separately `SourceIdInner` has a different implementation. Similar levels of complexity exist for `Eq`. Every one of these `impl`s  was added due to a real bug/issue we've had that needs to stay fixed. Not all of witch are reproducible enough to have made it into our test suite.

I [have some ideas](https://github.com/rust-lang/cargo/pull/14665#issuecomment-2412077472) for how to reorganize the types so that this is easier to reason about and faster. But given the history and the complexity I want to move extremely carefully.

The test pass, and it's a one line change,  but this still needs careful review.

### Additional information

r? @ehuss I remember you and Alex working very hard to track down most of these bugs.

### DIFF
--- a/src/cargo/core/source_id.rs
+++ b/src/cargo/core/source_id.rs
@@ -532,7 +532,10 @@ impl SourceId {
         url == CRATES_IO_INDEX || url == CRATES_IO_HTTP_INDEX || is_overridden_crates_io_url(url)
     }
 
-    /// Hashes `self`.
+    /// Hashes `self` to be used in the name of some Cargo folders, so shouldn't vary.
+    ///
+    /// For git and url, `as_str` gives the serialisation of a url (which has a spec) and so
+    /// insulates against possible changes in how the url crate does hashing.
     ///
     /// For paths, remove the workspace prefix so the same source will give the
     /// same hash in different locations, helping reproducible builds.
@@ -550,7 +553,11 @@ impl SourceId {
                 return;
             }
         }
-        self.hash(into)
+        self.inner.kind.hash(into);
+        match self.inner.kind {
+            SourceKind::Git(_) => (&self).inner.canonical_url.hash(into),
+            _ => (&self).inner.url.as_str().hash(into),
+        }
     }
 
     pub fn full_eq(self, other: SourceId) -> bool {
@@ -665,9 +672,6 @@ impl fmt::Display for SourceId {
     }
 }
 
-/// The hash of `SourceId` is used in the name of some Cargo folders, so shouldn't
-/// vary. `as_str` gives the serialisation of a url (which has a spec) and so
-/// insulates against possible changes in how the url crate does hashing.
 impl Hash for SourceId {
     fn hash<S: hash::Hasher>(&self, into: &mut S) {
         self.inner.kind.hash(into);
--- a/src/cargo/core/source_id.rs
+++ b/src/cargo/core/source_id.rs
@@ -675,10 +675,7 @@ impl fmt::Display for SourceId {
 impl Hash for SourceId {
     fn hash<S: hash::Hasher>(&self, into: &mut S) {
         self.inner.kind.hash(into);
-        match self.inner.kind {
-            SourceKind::Git(_) => self.inner.canonical_url.hash(into),
-            _ => self.inner.url.as_str().hash(into),
-        }
+        self.inner.canonical_url.hash(into);
     }
 }
 


Eh2406:
If this turns out not to be a valid transformation I am happy to update the PR to be tests to prove why it's not okay.

src/cargo/core/source_id.rs
```diff
@@ -666,15 +666,12 @@ impl fmt::Display for SourceId {
...
 /// insulates against possible changes in how the url crate does hashing.
 impl Hash for SourceId {
     fn hash<S: hash::Hasher>(&self, into: &mut S) {
         self.inner.kind.hash(into);
-        match self.inner.kind {
-            SourceKind::Git(_) => self.inner.canonical_url.hash(into),
-            _ => self.inner.url.as_str().hash(into),
-        }
+        self.inner.canonical_url.hash(into);
     }
```
epage, line 675: Most CanonicalUrl differences are related to git somehow.  I'm assuming that means we can't hit them in any other way.  There is the stripping of a trailing `/`.  However, we probably should have been respecting that in the hash in the first place I assume.  The question is if this will change anything significant enough

Eh2406: Perhaps if you have a index hosted on git, then `GITHUB.com/me/my-index.git/` could change a lot. I think the biggest concern here is having to recheck out dependencies. So even if this does affect people we might just eat that cost. On the other hand there's a lot of complexity here and maybe there are bigger problems I forgot about.

weihanglo: Haven't yet looked closer. I wonder if this would affect hacks like this one: <https://github.com/rust-lang/cargo/issues/5478#issuecomment-1055365283>.

Eh2406: I don't think that would be affected by this change. But it is something we have to watch for as I "clean up" this code.

weihanglo: That is true, since `[patch.<url>]` uses URLs directly, not SourceId.

This change may affect anywhere we let people define URL and parse it as `SourceId`, like the `[registries]` table? We could also deem this as a bug and fix it.

Eh2406: I would hope that we rely on `Eq` for those cases and not `Hash`. So that the problem you identified is not a problem with this PR, rather one to investigate on the next PR that changes `Eq`.

weihanglo (approved):
Since the paths to crates.io Git and spare index on users disks are not affected by this change, you have my vote.

src/cargo/core/source_id.rs
weihanglo: Fair enough. Have checked other usages of it, none of them seems user-facing like the path to index files. It should be fine to move over.

ehuss:
To be clear, the goal here is to simplify the code a little bit? Are there any other intended improvements?

And the downside is that if someone has a custom registry that canonicalizes differently (like a trailing slash, or upper-case characters in a GitHub URL), it will hash differently and they will end up with new directories in `~/.cargo/registry/*/...`, and duplicated cache contents?

If that's the tradeoff, I'm a little on the fence and could go either way.

If I'm understanding it correctly, would it be possible to update the `test_stable_hash` test with an example that illustrates the hash changing with this PR?

Eh2406:
>>    And the downside is that if someone has a custom registry that canonicalizes differently (like a trailing slash, or upper-case characters in a GitHub URL), it will hash differently and they will end up with new directories in ~/.cargo/registry/*/..., and duplicated cache contents?
> 
> That is the only concern I'm aware of. But it is possible that I missed something.

While trying to update the tests as requested I thought of another concern.  If the user has two custom registries that canonical eyes to the same thing, before this PR they would end up in separate directories in `~/.cargo/registry/*/...` and now they would end up in the same folder. With this collision cause problems?

weihanglo:
> , before this PR they would end up in separate directories in `~/.cargo/registry/*/...` and now they would end up in the same folder. With this collision cause problems?

Theoretically yes. However I doubt somebody relies on this behavior to distinguish custom registries. It either needs to have trailing `/` or `.git`. or on gitub.com with protocol scheme other than `https`. We could notify interest groups in advance. Not sure what is the correct channel, [t-cargo/build-integration](https://rust-lang.zulipchat.com/#narrow/channel/334885-t-cargo.2Fbuild-integration) doesn't seem to be the best fit.

Eh2406:
I was going to abandon this PR, but before I do... @weihanglo do those filenames  still go through this code path at all? Where do they go through `stable_hash`?

weihanglo:
> I was going to abandon this PR, but before I do... @weihanglo do those filenames still go through this code path at all? Where do they go through `stable_hash`?

The code path doesn't change at all. `stable_hash` still relies on Hash impl for SourceId

https://github.com/rust-lang/cargo/blob/652623b779c88fe44afede28bf7f1c9c07812511/src/cargo/core/source_id.rs?plain=1#L553

Eh2406:
Good point! But we can just in line that.  I just rebased and added a commit.  The first commit changes `stable_hash` to not depend on `hash`  but to have its own copy of the code. Then the second commit simplifies the main `hash`.

I feel like this alleviates the relevant concerns.

weihanglo:
> Good point! But we can just in line that. […] I feel like this alleviates the relevant concerns.

That is true. However I am forgetting the motivation of this PR. It doesn't seem to simplify the current code. Does this change open any door for future improvement?

Eh2406:
> It doesn't seem to simplify the current code. 

Separating the "must return the same hash value across versions of the code" into 1  function, even though there's more code, I do think it is a very small when for clarity. The Hash trait does not normally have that requirement, our reliance on it always feels a little awkward.

> I am forgetting the motivation of this PR. [..] Does this change open any door for future improvement?

Yes, and I'm happy to post a draft PR of where this is going if that would help. The idea is that  I like to separately intern the parts of SourceID that are part of `Eq` from the parts that are mutable. In order for that to be effective, I need to minimize the parts that are relevant to `Eq`. But, `Hash == Hash  must imply Eq == Eq`  so in order to simplify `Eq`  I needed to start by simplifying `Hash`. I think the three changes deserve their own PRs is because:
- this PR, the change to Hash, is mostly worried about the stable hash interactions.
- the next PR, the change to Eq,  It is mostly concerned with the interactions between the interning strategy and its implications on  the rest of the code.
- the last PR, changes to interning, are mostly concerned about performance.
If I did it in one PR, I'd be very concerned that  one or more of these concerns might get lost in the other conversations.

Opinions can and do vary. I am open to however you wish to proceed.

weihanglo (approved):
Thanks for the explanation, Jacob!

PR #14803, opened 2024-11-09
test: adjust `cargo_test_env` to unblock rust submodule update

weihanglo:

`[HOST_TARGET]` from `$CARGO` was not replaced.
See each commit message for the detail.

* Run `cargo t --test testsuite -- cargo_test_env`
* Apply the change to rust-lang/rust and run `./x.py test src/tools/cargo -- cargo_test_env`.

### DIFF
--- a/tests/testsuite/test.rs
+++ b/tests/testsuite/test.rs
@@ -3888,12 +3888,15 @@ fn cargo_test_env() {
         .file("src/lib.rs", &src)
         .build();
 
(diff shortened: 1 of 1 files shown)


epage (approved):

weihanglo:
This is ready for another round of review. I've removed the switch to `with_stderr_data` as it caused churn of path normalization and may worth exploring separately.

Issue #14621, opened 2024-09-30
Error message regression in Cargo 1.80 when fetching particular git revision


RobJellinghaus:
### Problem

I am working on Dependabot support for Rust. Currently Dependabot includes Rust 1.79. I was testing with Rust 1.80 and discovered that a Dependabot unit test started failing.

The specific failing Dependabot unit test exercises fetching a nonexistent revision from a git dependency: https://github.com/dependabot/dependabot-core/blob/main/cargo/spec/dependabot/cargo/update_checker/version_resolver_spec.rb#L355

I created a minimal repro with a .zip file below.

With Cargo 1.79, the final reported error from git is:
`revspec '11111b376b93484341c68fbca3ca110ae5cd2708' not found; class=Reference (4); code=NotFound (-3)`

With Cargo 1.80, the final reported error becomes:
`process didn't exit successfully: ``git fetch --verbose --force --update-head-ok 'https://github.com/BurntSushi/utf8-ranges' '+11111b376b93484341c68fbca3ca110ae5cd2708:refs/commit/11111b376b93484341c68fbca3ca110ae5cd2708'`` (exit status: 128)`

This updated error message doesn't describe the nature of the error as usefully as the preceding error, and can't be matched by Dependabot's error message parsing.

Evidently error message fidelity is not necessarily a Cargo goal in all cases, so if this is considered not to be a bug, I will work on a change to make Dependabot able to tolerate the new behavior. But it definitely is a regression from a user perspective to lose clarity on the actual git-level problem.

This was discovered on Linux (where Dependabot runs) but my most recent testing was on Windows; same behavior.

### Steps

1. Unzip [dbot_git_locked_ref_unit_test_break.zip](https://github.com/user-attachments/files/17196854/dbot_git_locked_ref_unit_test_break.zip)
2. Run `cargo update -p time:0.1.39 -vv`

Expected outcome (with 1.79):
```
Updating crates.io index
    Updating git repository `https://github.com/BurntSushi/utf8-ranges`
     Running `git fetch --tags --verbose --force --update-head-ok 'https://github.com/BurntSushi/utf8-ranges' '+refs/heads/*:refs/remotes/origin/*' '+HEAD:refs/remotes/origin/HEAD'`
POST git-upload-pack (332 bytes)
From https://github.com/BurntSushi/utf8-ranges
 = [up to date]      master     -> origin/master
 = [up to date]                 -> origin/HEAD
 = [up to date]      0.1.1      -> 0.1.1
[...]

PR #14806, opened 2024-11-10
fix(): error context for git_fetch refspec not found

dacianpascu06:

Fixes Issue : #14621 

Adds more error context for fetching a commit that doesn't exists.

I've created two tests, one for fast_path and one for libgit2.

r? @weihanglo

### DIFF
--- a/src/cargo/sources/git/utils.rs
+++ b/src/cargo/sources/git/utils.rs
@@ -1007,156 +1007,11 @@ pub fn fetch(
     }
 
     if let Some(true) = gctx.net_config()?.git_fetch_with_cli {
-        return fetch_with_cli(repo, remote_url, &refspecs, tags, gctx);
-    }
-
-    if gctx.cli_unstable().gitoxide.map_or(false, |git| git.fetch) {
-        let git2_repo = repo;
-        let config_overrides = cargo_config_to_gitoxide_overrides(gctx)?;
-        let repo_reinitialized = AtomicBool::default();
-        let res = oxide::with_retry_and_progress(
-            &git2_repo.path().to_owned(),
-            gctx,
-            &|repo_path,
-              should_interrupt,
-              mut progress,
-              url_for_authentication: &mut dyn FnMut(&gix::bstr::BStr)| {
-                // The `fetch` operation here may fail spuriously due to a corrupt
-                // repository. It could also fail, however, for a whole slew of other
-                // reasons (aka network related reasons). We want Cargo to automatically
-                // recover from corrupt repositories, but we don't want Cargo to stomp
-                // over other legitimate errors.
-                //
-                // Consequently we save off the error of the `fetch` operation and if it
-                // looks like a "corrupt repo" error then we blow away the repo and try
-                // again. If it looks like any other kind of error, or if we've already
-                // blown away the repository, then we want to return the error as-is.
-                loop {
-                    let res = oxide::open_repo(
-                        repo_path,
-                        config_overrides.clone(),
-                        oxide::OpenMode::ForFetch,
-                    )
-                    .map_err(crate::sources::git::fetch::Error::from)
-                    .and_then(|repo| {
-                        debug!("initiating fetch of {refspecs:?} from {remote_url}");
-                        let url_for_authentication = &mut *url_for_authentication;
-                        let remote = repo
-                            .remote_at(remote_url)?
-                            .with_fetch_tags(if tags {
-                                gix::remote::fetch::Tags::All
-                            } else {
-                                gix::remote::fetch::Tags::Included
-                            })
-                            .with_refspecs(
-                                refspecs.iter().map(|s| s.as_str()),
-                                gix::remote::Direction::Fetch,
-                            )
-                            .map_err(crate::sources::git::fetch::Error::Other)?;
-                        let url = remote
-                            .url(gix::remote::Direction::Fetch)
-                            .expect("set at init")
-                            .to_owned();
-                        let connection = remote.connect(gix::remote::Direction::Fetch)?;
-                        let mut authenticate = connection.configured_credentials(url)?;
-                        let connection = connection.with_credentials(
-                            move |action: gix::protocol::credentials::helper::Action| {
-                                if let Some(url) = action.context().and_then(|gctx| {
-                                    gctx.url.as_ref().filter(|url| *url != remote_url)
-                                }) {
-                                    url_for_authentication(url.as_ref());
-                                }
-                                authenticate(action)
-                            },
-                        );
-                        let outcome = connection
-                            .prepare_fetch(&mut progress, gix::remote::ref_map::Options::default())?
-                            .with_shallow(shallow.clone().into())
-                            .receive(&mut progress, should_interrupt)?;
-                        Ok(outcome)
-                    });
-                    let err = match res {
-                        Ok(_) => break,
-                        Err(e) => e,
-                    };
-                    debug!("fetch failed: {}", err);
-
-                    if !repo_reinitialized.load(Ordering::Relaxed)
-                        // We check for errors that could occur if the configuration, refs or odb files are corrupted.
-                        // We don't check for errors related to writing as `gitoxide` is expected to create missing leading
-                        // folder before writing files into it, or else not even open a directory as git repository (which is
-                        // also handled here).
-                        && err.is_corrupted()
-                        || has_shallow_lock_file(&err)
-                    {
-                        repo_reinitialized.store(true, Ordering::Relaxed);
-                        debug!(
-                            "looks like this is a corrupt repository, reinitializing \
-                     and trying again"
-                        );
-                        if oxide::reinitialize(repo_path).is_ok() {
-                            continue;
-                        }
-                    }
-
-                    return Err(err.into());
-                }
-                Ok(())
-            },
-        );
-        if repo_reinitialized.load(Ordering::Relaxed) {
-            *git2_repo = git2::Repository::open(git2_repo.path())?;
-        }
-        res
+        fetch_with_cli(repo, remote_url, &refspecs, tags, gctx)
+    } else if gctx.cli_unstable().gitoxide.map_or(false, |git| git.fetch) {
+        fetch_with_gitoxide(repo, remote_url, refspecs, tags, shallow, gctx)
     } else {
-        debug!("doing a fetch for {remote_url}");
-        let git_config = git2::Config::open_default()?;
-        with_fetch_options(&git_config, remote_url, gctx, &mut |mut opts| {
-            if tags {
-                opts.download_tags(git2::AutotagOption::All);
-            }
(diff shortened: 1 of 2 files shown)


tests/testsuite/git.rs
```diff
@@ -4052,6 +4052,90 @@ src/lib.rs
         .run();
 }
 
+#[cargo_test(public_network_test)]
```
weihanglo, line 4055: You'll need `requires_git` as well

```suggestion
#[cargo_test(public_network_test, requires_git)]
```

src/cargo/sources/git/utils.rs
```diff
@@ -1152,6 +1152,13 @@ pub fn fetch(
                         continue;
                     }
                 }
+                if matches!(err.class(), ErrorClass::Net) && err.code() == git2::ErrorCode::Eof {
+                    let revspec_id: String = refspecs
```
weihanglo, line 1156: I would avoid doing this re-parsing, especially when we need to do that for every kind of fetch approaches (like, we've covered git CLI and `libgit2` here but missed gitoxide).

What do you think of doing it one level up as [mentioned here](https://github.com/rust-lang/cargo/issues/14621#issuecomment-2397279846), so we don't need to repeat ourselves for git/libgit2/gitoxide?

tests/testsuite/git.rs
weihanglo, line 1: If possible, please follow [this guide](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request), which one commit contains a test showing the current problematic behavior with assertions, followed by the other commits fixing the behavior and test. Each commit is expected to pass CI.

src/cargo/sources/git/utils.rs
```diff
@@ -1221,9 +1228,21 @@ fn fetch_with_cli(
...
-    cmd.exec()?;
+
+    let revspec_id: String = refspecs
+        .iter()
+        .map(|s| s.split(':').next().unwrap_or("").to_string())
+        .collect();
+
+    cmd.exec().with_context(|| {
+        format!(
+            "Failed to fetch via Github fast path \nrevspec {:?} not found",
```
ehuss, line 1242: BTW, this does not look related to the GitHub fast path. I'm not sure I understand why it is mentioning it here.

src/cargo/sources/git/utils.rs
```diff
@@ -944,9 +944,15 @@ pub fn fetch(
...
     let shallow = remote_kind.to_shallow_setting(repo.is_shallow(), gctx);
 
+    // Flag to keep track if the rev is a full commit hash
+    let mut fast_path_rev: bool = false;
+
     let oid_to_fetch = match github_fast_path(repo, remote_url, reference, gctx) {
         Ok(FastPathRev::UpToDate) => return Ok(()),
-        Ok(FastPathRev::NeedsFetch(rev)) => Some(rev),
+        Ok(FastPathRev::NeedsFetch(rev)) => {
+            fast_path_rev = true;
```
weihanglo, line 953: Instead of adding a flag here, we should probably add one in `} else if let Some(oid_to_fetch) = oid_to_fetch {` after line 995, which the `ref/commit/{0}` really happened and caused the problem

tests/testsuite/git.rs
```diff
@@ -4052,6 +4052,94 @@ src/lib.rs
...
+  failed to clone into: [ROOT]/home/.cargo/git/db/bitflags-[HASH]
+
+Caused by:
+  process didn't exit successfully: `git fetch --no-tags --force --update-head-ok 'https://github.com/rust-lang/bitflags.git' '+11111b376b93484341c68fbca3ca110ae5cd2790:refs/commit/11111b376b93484341c68fbca3ca110ae5cd2790'` ([EXIT_STATUS]: 128)
+
+"#]])
+        .run();
+}
+
+#[cargo_test(public_network_test, requires_git)]
```
weihanglo, line 4099: libgit2 doesn't require git CLI.

```suggestion
#[cargo_test(public_network_test)]
```

src/cargo/sources/git/utils.rs
```diff
@@ -1011,152 +1011,9 @@ pub fn fetch(
...
-                    if reinitialize(repo).is_ok() {
-                        continue;
-                    }
-                }
-
-                return Err(err.into());
-            }
-            Ok(())
-        })
+        return fetch_with_libgit2(repo, remote_url, refspecs, tags, shallow, gctx);
```
weihanglo: `return` is redundant.

src/cargo/sources/git/utils.rs
```diff
@@ -958,6 +964,8 @@ pub fn fetch(
 
     clean_repo_temp_files(repo);
 
+    let mut revision: String = String::new();
```
weihanglo, line 967: Probably don't need this extra variable. `oid_to_fetch` already contains the revision information we want.

src/cargo/sources/git/utils.rs
```diff
@@ -1007,13 +1016,36 @@ pub fn fetch(
     }
 
     if let Some(true) = gctx.net_config()?.git_fetch_with_cli {
-        return fetch_with_cli(repo, remote_url, &refspecs, tags, gctx);
+        match fetch_with_cli(repo, remote_url, &refspecs, tags, gctx) {
```
weihanglo, line 1019: A better way to handling this is

```rust
let result = if let Some(true) = gctx.net_config()?.git_fetch_with_cli {
  // ..
} else if gctx.cli_unstable().gitoxide.map_or(false, |git| git.fetch) {
  // ..
} else {
  // ..
};

if fast_path_rev
  result.with_context(|| format("revision {oid_to_fetch} not found")
} else {
  result
}
```

src/cargo/sources/git/utils.rs
```diff
@@ -1007,13 +1016,36 @@ pub fn fetch(
...
+            }
+        };
     }
 
     if gctx.cli_unstable().gitoxide.map_or(false, |git| git.fetch) {
-        return fetch_with_gitoxide(repo, remote_url, refspecs, tags, shallow, gctx);
+        match fetch_with_gitoxide(repo, remote_url, refspecs, tags, shallow, gctx) {
+            Ok(result) => return Ok(result),
+            Err(e) => {
+                if fast_path_rev == true {
```
weihanglo, line 1031: if it is a boolean, just write `if bool` directly . Don't compare `if bool == true`.

src/cargo/sources/git/utils.rs
```diff
@@ -1007,13 +1016,36 @@ pub fn fetch(
...
+                    return Err(e).context(format!("Revision {:?} not found", revision));
```
weihanglo, line 1032: `.context` is eagerly evaluated — a string of `Revision ...` would be allocated even when it is not on the error path. use `with_context` to avoid that.

weihanglo, line 1032: In Cargo by convention we use lowercase for the first character for the first sentence of a message.

src/cargo/sources/git/utils.rs
```diff
@@ -1227,6 +1084,171 @@ fn fetch_with_cli(
...
+}
+
+fn fetch_with_libgit2(
+    repo: &mut git2::Repository,
+    remote_url: &str,
+    refspecs: Vec<String>,
+    tags: bool,
+    shallow: gix::remote::fetch::Shallow,
+    gctx: &GlobalContext,
+) -> Result<(), anyhow::Error> {
```
weihanglo, line 1234: In Cargo  `CargoResult<T>` is used as a synonym of `Result<T, anyhow::Error>`. You're not wrong here just we have a subtle convention.

```suggestion
) -> CargoResult<()> {
```

tests/testsuite/git.rs
```diff
@@ -4052,6 +4052,100 @@ src/lib.rs
...
+Caused by:
+  revision 11111b376b93484341c68fbca3ca110ae5cd2790 not found
+
+Caused by:
+  network failure seems to have happened
+  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
+  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli
+
+Caused by:
+  could not read from remote repository; class=Net (12); code=Eof (-20)
```
weihanglo, line 4143: It is interesting that the error message doesn't seem to be emitted on GitHub Actions machine, but only on my local machine.

If it still happens, instead of asserting the entire message, try omitting some parts we're not interested in with `...` (For the syntax, see https://doc.rust-lang.org/nightly/nightly-rustc/cargo_test_support/compare/index.html).


```suggestion
[UPDATING] git repository `https://github.com/rust-lang/bitflags.git`
...
[ERROR] failed to get `bitflags` as a dependency of package `foo v0.1.0 ([ROOT]/foo)`
Caused by:
  failed to load source for dependency `bitflags`
Caused by:
  Unable to update https://github.com/rust-lang/bitflags.git?rev=11111b376b93484341c68fbca3ca110ae5cd2790
Caused by:
  failed to clone into: [ROOT]/home/.cargo/git/db/bitflags-[HASH]
Caused by:
  revision 11111b376b93484341c68fbca3ca110ae5cd2790 not found
...
```

weihanglo: For the context:

```
---- git::git_fetch_libgit2_error_message stdout ----
thread 'git::git_fetch_libgit2_error_message' panicked at tests/testsuite/git.rs:4163:10:

---- expected: tests/testsuite/git.rs:4135:27
++++ actual:   stderr
   1    1 | [UPDATING] git repository `[https://github.com/rust-lang/bitflags.git`](https://github.com/rust-lang/bitflags.git%60)
   2      - [WARNING] spurious network error (3 tries remaining): could not read from remote repository; class=Net (12); code=Eof (-20)
   3      - [WARNING] spurious network error (2 tries remaining): could not read from remote repository; class=Net (12); code=Eof (-20)
   4      - [WARNING] spurious network error (1 tries remaining): could not read from remote repository; class=Net (12); code=Eof (-20)
   5    2 | [ERROR] failed to get `bitflags` as a dependency of package `foo v0.1.0 ([ROOT]/foo)`
   6    3 | 
   7    4 | Caused by:
   8    5 |   failed to load source for dependency `bitflags`
   9    6 | 
          ⋮
  15   12 | 
  16   13 | Caused by:
  17   14 |   revision 11111b376b93484341c68fbca3ca110ae5cd2790 not found
  18   15 | 
  19   16 | Caused by:
  20      -   network failure seems to have happened
  21      -   if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
  22      -   https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli
       17 +   Could not decode server reply
  23   18 | 
  24   19 | Caused by:
  25      -   could not read from remote repository; class=Net (12); code=Eof (-20)
       20 +   Unknown or unsupported header: "ERR upload-pack: not our ref 11111b376b93484341c68fbca3ca110ae5cd2790"
       ```

PR #3857, opened 2017-03-23
Include package props with corresponding env vars into target metadata

antonlarin:
Previously, when changing package properties with corresponding environment variables (such as authors, which has CARGO_PKG_AUTHORS), it didn't invalidate the build, even though there could have been a dependency on such variables in the source code.

This commit includes 3 such properties: authors list, description and homepage in the target metadata.

I've added a test only for description change, can add more if necessary.
Fixes #3696.

PR #14811, opened 2024-11-12
fix(env): dont track envs set by cargo in dep-info file 

### LINKED ISSUE #14798: CARGO_RUSTC_CURRENT_DIR is causing forced rebuilds
### Problem

In the cargo repo, there is an environment override setting the `CARGO_RUSTC_CURRENT_DIR` environment variable: https://github.com/rust-lang/cargo/blob/79866e12c663e866a904d713c29745563a64aafa/.cargo/config.toml#L7-L10

This is causing `testsuite` to always be considered dirty on my system. I am getting:

```
FsStatusOutdated(StaleItem(ChangedEnv 
{ var: "CARGO_RUSTC_CURRENT_DIR", 
previous: Some("/Users/eric/Proj/rust/cargo"), 
current: Some("/Users/eric/Proj/rust/cargo/") 
}
```

### Steps

1. Check out rust-lang/cargo
2. `cargo test --test testsuite --no-run`
3. `cargo test --test testsuite --no-run` -- This should build testsuite again.

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
cargo 1.84.0-nightly (031049782 2024-11-01)
release: 1.84.0-nightly
commit-hash: 0310497822a7a673a330a5dd068b7aaa579a265e
commit-date: 2024-11-01
host: aarch64-apple-darwin
os: Mac OS 14.6.1 [64-bit]
```

weihanglo:

Don't track environment variables set by Cargo[1] in Cargo's dep-info file.

This patch is quite hacky since the "set-by-Cargo" env vars are
hardcoded. However, not all environment variables set by Cargo are
statically known. To prevent the actual environment variables applied to
rustc invocation get out of sync from what we hard-code, a debug time
assertion is put to help discover missing environment variables earlier.

Fixes #14798

A new test is added to show no rebuild triggered.

Try adding a random new env [here][2] and see if a debug build fails.

[1]: https://doc.rust-lang.org/nightly/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates
[2]: https://github.com/rust-lang/cargo/blob/2e7fc43158d9621dccb4012756a5ad3efba849e9/src/cargo/core/compiler/compilation.rs?plain=1#L399

### DIFF
--- a/src/cargo/core/compiler/compilation.rs
+++ b/src/cargo/core/compiler/compilation.rs
@@ -527,3 +527,98 @@ fn target_linker(bcx: &BuildContext<'_, '_>, kind: CompileKind) -> CargoResult<O
     }
     Ok(matching_linker.map(|(_k, linker)| linker.val.clone().resolve_program(bcx.gctx)))
 }
+
+/// This tracks environment variables Cargo sets to rustc when building a crate.
+///
+/// This only inclues variables with statically known keys.
+/// For environment variable keys that vary like `CARG_BIN_EXE_<name>` or
+/// `-Zbindeps` related env vars, we compare only their prefixes to determine
+/// if they are internal.
+/// See [`is_env_set_by_cargo`] and
+/// <https://doc.rust-lang.org/nightly/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates>.
+const ENV_VARS_SET_FOR_CRATES: [&str; 23] = [
+    crate::CARGO_ENV,
+    "CARGO_MANIFEST_DIR",
+    "CARGO_MANIFEST_PATH",
+    "CARGO_PKG_VERSION",
+    "CARGO_PKG_VERSION_MAJOR",
+    "CARGO_PKG_VERSION_MINOR",
+    "CARGO_PKG_VERSION_PATCH",
+    "CARGO_PKG_VERSION_PRE",
+    "CARGO_PKG_AUTHORS",
+    "CARGO_PKG_NAME",
+    "CARGO_PKG_DESCRIPTION",
+    "CARGO_PKG_HOMEPAGE",
+    "CARGO_PKG_REPOSITORY",
+    "CARGO_PKG_LICENSE",
+    "CARGO_PKG_LICENSE_FILE",
+    "CARGO_PKG_RUST_VERSION",
+    "CARGO_PKG_README",
+    "CARGO_CRATE_NAME",
+    "CARGO_BIN_NAME",
+    "OUT_DIR",
+    "CARGO_PRIMARY_PACKAGE",
+    "CARGO_TARGET_TMPDIR",
+    paths::dylib_path_envvar(),
+];
+/// Asserts if the given env vars are controlled by Cargo.
+///
+/// This is only for reminding Cargo developer to keep newly added environment
+/// variables in sync with [`ENV_VARS_SET_FOR_CRATES`].
+#[cfg(debug_assertions)]
+pub(crate) fn assert_only_envs_set_by_cargo<'a>(
+    keys: impl Iterator<Item = impl AsRef<str>>,
+    env_config: &HashMap<String, OsString>,
+) {
+    for key in keys {
+        let key = key.as_ref();
+        // When running Cargo's test suite,
+        // we're fine if it is from the `[env]` table
+        if env_config.contains_key(key) {
+            continue;
+        }
+        assert!(
+            is_env_set_by_cargo(key),
+            "`{key}` is not tracked as an environment variable set by Cargo\n\
+            Add it to `ENV_VARS_SET_FOR_CRATES` if you intend to introduce a new one"
+        );
+    }
+}
+
+/// True if the given env var is controlled or set by Cargo.
+/// See [`ENV_VARS_SET_FOR_CRATES`].
+pub(crate) fn is_env_set_by_cargo(key: &str) -> bool {
+    ENV_VARS_SET_FOR_CRATES.contains(&key)
+        || key.starts_with("CARGO_BIN_EXE_")
+        || key.starts_with("__CARGO") // internal/test-only envs
+        || key == "RUSTC_BOOTSTRAP" // for -Zbuild-std
+        || is_artifact_dep_env_vars(key)
+}
+
+/// Whether an env var is set because of `-Zbindeps`.
+fn is_artifact_dep_env_vars(key: &str) -> bool {
+    let Some(key) = key.strip_prefix("CARGO_") else {
+        return false;
+    };
+    let Some(key) = key
+        .strip_prefix("BIN_")
+        .or_else(|| key.strip_prefix("CDYLIB_"))
+        .or_else(|| key.strip_prefix("STATICLIB_"))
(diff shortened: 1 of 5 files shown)


src/cargo/core/compiler/compilation.rs
```diff
@@ -529,3 +532,88 @@ fn target_linker(bcx: &BuildContext<'_, '_>, kind: CompileKind) -> CargoResult<O
...
+        "CARGO_PKG_RUST_VERSION",
+        "CARGO_PKG_README",
+        "CARGO_CRATE_NAME",
+        "CARGO_BIN_NAME",
+        "OUT_DIR",
+        "CARGO_PRIMARY_PACKAGE",
+        "CARGO_TARGET_TMPDIR",
+        paths::dylib_path_envvar(),
+    ])
+});
```
epage, line 570: Should we make this an array with a test to enforce uniqueness?

We could make `dylib_path_envvar` const

tests/testsuite/cargo_env_config.rs
```diff
@@ -460,8 +460,10 @@ fn override_env_set_by_cargo() {
         .build();
 
     let args = [
-        "--config", "env.CARGO_MANIFEST_DIR='Sauron'",
-        "--config", "env.CARGO_PKG_NAME='Saruman'",
+        "--config",
+        "env.CARGO_MANIFEST_DIR='Sauron'",
+        "--config",
+        "env.CARGO_PKG_NAME='Saruman'",
```
epage, line 466: Re-formatting?  Can this be moved into the other commit?

src/cargo/core/compiler/compilation.rs
```diff
@@ -529,3 +532,88 @@ fn target_linker(bcx: &BuildContext<'_, '_>, kind: CompileKind) -> CargoResult<O
...
+        if env_config.contains_key(key) {
+            continue;
+        }
+        assert!(
+            is_env_set_by_cargo(key),
+            "`{key}` is not tracked as an environment variable set by Cargo\n\
+            Add it to `ENV_VARS_SET_FOR_CRATES` if you intend to introduce a new one"
+        );
+    }
+}
```
epage, line 594: > This patch is quite hacky since the "set-by-Cargo" env vars are
hardcoded. However, not all environment variables set by Cargo are
statically known. To prevent the actual environment variables applied to
rustc invocation get out of sync from what we hard-code, a debug time
assertion is put to help discover missing environment variables earlier.

@ehuss I'm curious about your thoughts on this

src/cargo/core/compiler/compilation.rs
weihanglo: Do you suggest not using lazy hashset at all, or keeping an array here and move lazy hashset into `is_env_set_by_cargo`?

epage: Not using a lazy hashset at all.  With the code complication, data set size, and lazy lock overhead, I'm assuming a slice would be better

src/cargo/core/compiler/compilation.rs
```diff
@@ -529,3 +529,98 @@ fn target_linker(bcx: &BuildContext<'_, '_>, kind: CompileKind) -> CargoResult<O
     }
     Ok(matching_linker.map(|(_k, linker)| linker.val.clone().resolve_program(bcx.gctx)))
 }
+
```
linyihai, line 532: In order to better discover these variables, whether to put this part of the content in a separate file management is better?

weihanglo: A simple search for `ENV_VARS_SET_FOR_CRATES` could lead the way, though no objection to this proposal, just lacking of a better module name (naming is hard 😆).

src/cargo/core/compiler/compilation.rs
```diff
@@ -527,3 +527,98 @@ fn target_linker(bcx: &BuildContext<'_, '_>, kind: CompileKind) -> CargoResult<O
...
+    "CARGO_PKG_VERSION_MAJOR",
+    "CARGO_PKG_VERSION_MINOR",
+    "CARGO_PKG_VERSION_PATCH",
+    "CARGO_PKG_VERSION_PRE",
+    "CARGO_PKG_AUTHORS",
+    "CARGO_PKG_NAME",
+    "CARGO_PKG_DESCRIPTION",
+    "CARGO_PKG_HOMEPAGE",
+    "CARGO_PKG_REPOSITORY",
+    "CARGO_PKG_LICENSE",
```
epage, line 553: To remove these requires that we are tracking these somewhere else. For things like the package name, thats more obvious. For ones like this, where is that happening? Should we have a test that changing the license causes a rebuild?

weihanglo: That's a good catch. We actually only rebuild for `authors`, `homepage`, `repository`, and `repository`.

https://github.com/rust-lang/cargo/blob/298e403317d231892e11bf93d3073f011841be48/src/cargo/core/compiler/fingerprint/mod.rs?plain=1#L1534

If we're going to hand-pick some of them, I would rather drop this out.

Since we were trying to fix #14811, the alternative is a “wontfix” and documenting it. 

> If you're setting `CARGO_SOMETHING` in `[env]`, you are screwed because it won't work and may hurt your rebuild detection.

weihanglo: Apparently, `package.description` was tracked by fingerprint because of the usage of `CARGO_PKG_DESCRIPTION` <https://github.com/rust-lang/cargo/pull/3857>.

Okay, maybe the other way round—We do hand-pick whatever we want to "remove" from cargo's dep-info file. The other stay in dep-info so they are happy to rebuild when requires by `env!`. And we don't even need to track them in fingerprint.

epage: I think it makes sense for us to track metadata in dep info rather than fingerprint so we only rebuild if the user cares.

As for the other part, I'm unusre

weihanglo: https://github.com/rust-lang/cargo/blob/3cd245ce48ee2646509f589d06b341334db8d9cd/src/cargo/core/manifest.rs?plain=1#L145-L146

It is so fun that `package.links` and `package.rust-version` are considered metadata and and won't trigger any rebuild when changed.

epage: imo there isn't a reason to rebiild on `rust-version` change unless the user is using it.

weihanglo: That is fair. It is a kind of metadata for the actual compilation.

weihanglo:
Going to close this.

* It is not good that a patch relying on debug assertions
* We may want to track metadata by Cargo's dep-info file, insteado of manual track in fingerprint. https://github.com/rust-lang/cargo/pull/14811#discussion_r1880431940
* `CARGO_RUSTC_CURRENT_DIR` has been removed. If other folks has bumped into the same situation of setting internal envs in `[env]` table, we can at least tell them it is not supported.

epage:
`CARGO_RUSTC_CURRENT_DIR` may come back because `file!` is being specified, see rust-lang/rust#134442

PR #14812, opened 2024-11-13
feat(rustdoc): diplay env vars in extra verbose mode 

weihanglo:

This was found when doing #14811.

Other counterparts display environment variables,
but rustdoc does not.
This patch makes rustdoc follow suit.

Run `cargo doc -vv`


ehuss (approved):
Thanks!

PR #14817, opened 2024-11-13
fix(build-rs)!: Updates from an audit

epage:

Did a quick audit for parts of the API and implementation that stood out to me.  So far, I've mostly focused on "input".

### Additional information

CC @CAD97

### DIFF
--- a/crates/build-rs/src/input.rs
+++ b/crates/build-rs/src/input.rs
@@ -27,7 +27,7 @@ macro_rules! invalid {
 }
 
 #[track_caller]
-fn get_bool(key: &str) -> bool {
+fn is_present(key: &str) -> bool {
     env::var_os(key).is_some()
 }
 
@@ -135,7 +135,7 @@ pub fn cargo_feature(name: &str) -> bool {
     }
     let name = name.to_uppercase().replace('-', "_");
     let key = format!("CARGO_FEATURE_{name}");
-    get_bool(&key)
+    is_present(&key)
 }
 
 /// For each [configuration option] of the package being built, this will contain
@@ -162,31 +162,31 @@ mod cfg {
     #[cfg(any())]
     #[track_caller]
     pub fn cargo_cfg_clippy() -> bool {
-        get_bool("CARGO_CFG_CLIPPY")
+        is_present("CARGO_CFG_CLIPPY")
     }
 
     /// If we are compiling with debug assertions enabled.
     #[track_caller]
     pub fn cargo_cfg_debug_assertions() -> bool {
-        get_bool("CARGO_CFG_DEBUG_ASSERTIONS")
+        is_present("CARGO_CFG_DEBUG_ASSERTIONS")
     }
 
     #[cfg(any())]
     #[track_caller]
     pub fn cargo_cfg_doc() -> bool {
-        get_bool("CARGO_CFG_DOC")
+        is_present("CARGO_CFG_DOC")
     }
 
     #[cfg(any())]
     #[track_caller]
     pub fn cargo_cfg_docsrs() -> bool {
-        get_bool("CARGO_CFG_DOCSRS")
+        is_present("CARGO_CFG_DOCSRS")
     }
 
     #[cfg(any())]
     #[track_caller]
     pub fn cargo_cfg_doctest() -> bool {
-        get_bool("CARGO_CFG_DOCTEST")
+        is_present("CARGO_CFG_DOCTEST")
     }
 
     /// The level of detail provided by derived [`Debug`] implementations.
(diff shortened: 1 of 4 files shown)


crates/build-rs/src/input.rs
```diff
@@ -480,112 +445,159 @@ pub fn rustc_wrapper() -> Option<PathBuf> {
...
 
 /// The pre-release version of your package.
 #[track_caller]
-pub fn cargo_pkg_version_pre() -> String {
-    get_str("CARGO_PKG_VERSION_PRE")
+pub fn cargo_pkg_version_pre() -> Option<String> {
+    to_opt(var_or_panic("CARGO_PKG_VERSION_PRE")).map(to_string)
 }
 
 /// Colon separated list of authors from the manifest of your package.
```
CAD97, line 497: Since we parse the colon separation, this description isn't accurate anymore. 

```suggestion
/// The list of authors from the manifest of your package.
```

crates/build-rs/src/input.rs
```diff
@@ -480,112 +445,159 @@ pub fn rustc_wrapper() -> Option<PathBuf> {
...
-    get_path("CARGO_PKG_README")
+pub fn cargo_pkg_readme() -> Option<PathBuf> {
+    to_opt(var_or_panic("CARGO_PKG_README")).map(to_path)
+}
+
+fn is_present(key: &str) -> bool {
+    var_os(key).is_some()
+}
+
+#[track_caller]
```
CAD97, line 556: Note to reviewers: buildscripts are compiled without debug info by default, so backtraces aren't helpful. That's why `#[track_caller]` is applied so liberally here, so the caller actually knows where the bad call was made.

crates/build-rs/src/input.rs
```diff
@@ -480,112 +445,159 @@ pub fn rustc_wrapper() -> Option<PathBuf> {
...
+fn to_strings(value: std::ffi::OsString, sep: char) -> Vec<String> {
+    if value.is_empty() {
+        return Vec::new();
+    }
+    let value = to_string(value);
+    let values = value.split(sep).map(str::to_owned).collect();
+    values
```
CAD97, line 587: Why doesn't [clippy::let_and_return](https://rust-lang.github.io/rust-clippy/master/index.html#let_and_return) fire here? Something to do with being conservative about temporary lifetimes, maybe? 

```suggestion
    value.split(sep).map(str::to_owned).collect()
```

CAD97 (approved):
Improvements all look good to me 👍

crates/build-rs/src/input.rs
epage: I wonder where we should record design principles like that.  For example, I had considered moving to `env!` but realized that re-running is cheaper than recompilation

CAD97: The default answer would be a DESIGN.md alongside README.md. That seems as reasonable a choice as any other[^1], although any such document always is difficult to organize and keep up to date.

[^1]: The main alternative being a wiki page, which is even less likely to be updated or even looked at ime since it isn't colocated.

As for `env!` — it's lots of "fun" to keep track of whether we're querying the environment for building build-rs (set at our compile time) or for the client buildscript (only set at our run time).

weihanglo: For build-rs I think we should in general query only runtime envs, though I don't fully get the context of using `env!` for this function.

weihanglo (approved):
Thank you both. Looks great!

crates/build-rs/src/input.rs
epage: The only time we use `env!` is to get `build-rs`s MSRV.  If we were to do things at compile time, we'd have to turn all of these functions into macros so that `env!` runs in the callers context.

crates/build-rs/src/input.rs
```diff
@@ -480,112 +445,158 @@ pub fn rustc_wrapper() -> Option<PathBuf> {
...
+    T: std::str::FromStr,
+    T::Err: std::fmt::Display,
+{
+    let value = to_string(value);
+    match value.parse() {
+        Ok(s) => s,
+        Err(err) => {
+            panic!("{err}")
+        }
+    }
```
weihanglo, line 601: nit: any reason we can't do this?

```suggestion
    value.parse().unwrap()
```

weihanglo:
Anyway, the only nit is just a nit. Going to merge this now.

crates/build-rs/src/input.rs
epage: The question is whether to report Debug or Display

Issue #14814, opened 2024-11-13
Support underline and shortline compatible for cargo-remove


A4-Tacks:
### Problem

**Reproduction**:
```sh
lrne@Acer-Lrne[/path/to/]
    $ cargo new foo && cd foo
    Creating binary (application) `foo` package
note: see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
lrne@Acer-Lrne[/path/to/foo/]
    $ cargo add cfg_if
    Updating `rsproxy-sparse` index
warning: translating `cfg_if` to `cfg-if`
      Adding cfg-if v1.0.0 to dependencies
             Features:
             - compiler_builtins
             - core
             - rustc-dep-of-std
    Updating `rsproxy-sparse` index
     Locking 2 packages to latest compatible versions
lrne@Acer-Lrne[/path/to/foo/]
    $ cargo remove cfg_if
    Removing cfg_if from dependencies
error: the dependency `cfg_if` could not be found in `dependencies`
lrne@Acer-Lrne[/path/to/foo/]
    $ cargo remove cfg-if
    Removing cfg-if from dependencies
lrne@Acer-Lrne[/path/to/foo/]
    $ 
```

epage:
At least providing a "did you mean" more generally and not just for word separators) in the error messages would be good.

However, I'm unsure how far we should take normalization in Cargo. Yes, we do it in `cargo-add`. I don't remember the reason it did this in cargo-edit and I think we overlooked discussing this when merging. While `remove` is parallel to `add`, its also like other commands in that you've now been warned that the name isn't correct and you are operating on your existing dependencies.

A4-Tacks:
If cargo-remove cannot translating underline and shortline, at least try giving a similar package name hint in the error message

epage:
#14818 improved the error message.

A4-Tacks:
Thanks!

PR #14818, opened 2024-11-13
fix(remove): On error, suggest other dependencies 

epage:

`cargo remove` already will suggest other tables if appropriate.
`cargo add` auto-corrects `_` and `-`.

This is an extension of the two by suggesting existing dependencies
within the same table that are "close".

Related to #14814

I chose to make alt tables and alt deps mutually exclusive in the
message because I didn't wantto deal with how to separate things if both
show up.
Most likely, people will only expect one or the other and if its in an
alt table, then most likely they mean that.


ehuss (approved):
Thanks!

Issue #14821, opened 2024-11-14
Source replacements are ignored when producing compilation units


P-E-Meunier:
### Problem

When replacing sources in `.cargo/config.toml` so that two different sources point to the same vendored directory, cargo identifies a single package from two different sources as different, giving rise to a confusing error message where a type from a file is claimed to be distinct from itself:

```
error[E0308]: mismatched types
 --> src/main.rs:2:10
  |
2 |     b::b(c::C {});
  |     ---- ^^^^^^^ expected `c::C`, found `C`
  |     |
  |     arguments to this function are incorrect
  |
  = note: `C` and `c::C` have similar names, but are actually distinct types
note: `C` is defined in crate `c`
 --> /tmp/cargo-vendor-test/a/vendor/c/src/lib.rs:1:1
  |
1 | pub struct C {}
  | ^^^^^^^^^^^^
note: `c::C` is defined in crate `c`
 --> /tmp/cargo-vendor-test/a/vendor/c/src/lib.rs:1:1
  |
1 | pub struct C {}
  | ^^^^^^^^^^^^
  = note: perhaps two different versions of crate `c` are being used?
note: function defined here
 --> /tmp/cargo-vendor-test/a/vendor/b/src/lib.rs:1:8
  |
1 | pub fn b(_c: c::C) {}
  |        ^
```

### Steps

1. Clone https://github.com/P-E-Meunier/cargo-vendor-test
2. Run `cargo vendor`, which yields a wrong output (as per #14729).
3. The correct output should be:
```
[source."sparse+http://localhost:8080/registry-y/"]
registry = "sparse+http://localhost:8080/registry-y/"
replace-with = "vendored-sources"

[source."sparse+http://localhost:8080/registry-x/"]
registry = "sparse+http://localhost:8080/registry-x/"
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "vendor"
```
4. Run `cargo build`

### Version

```text

```

weihanglo:
Ran `cargo vendor` didn't give me that output. With which version of Cargo under which directory should I run `cargo vendor`?

P-E-Meunier:
So, in my example repo, you need to run those steps from subdirectory `a`.
In #14729, you additionally need to start a fake private registry before using the following command from the root of the repo:
```
python -m http.server 8080 --bind 127.0.0.1 --directory .
```

weihanglo:
[...]

Issue #14729, opened 2024-10-25
Cargo vendor doesn't replace all sources used by dependencies in its output


P-E-Meunier:
### Problem

Let's say we have three crates, A, B and C, where A depends on B and C and B depends on C.

We serve them using two identical registries x and y. Crate A has the following line in its Cargo.toml:

```
B = { version = ..., registry = "x" }
C = { version = ..., registry = "x" }
```

Now, let's also assume that the authors of B preferred registry y, B would have the following Cargo.toml:

```
C = { version = ..., registry = "y" }
```

Since the hashes of C, as served by x and y, are identical, cargo chooses only one source in Cargo.lock. In our example let's say it chooses x.

When running `cargo vendor`, the stdout shows what replacements are needed, and that stdout is meant to be consumed by Cargo itself when doing `cargo build --offline`.

Because the sources are taken from Cargo.lock, in our case the output will replace x only. While this seems intuitively correct (all crates are indeed vendored, and all their source registries are replaced), `cargo build --offline` still fails in our example, since the Cargo.toml of `B` mentions registry y.

### Possible Solution(s)

One fix could have been to tell cargo build to ignore source registries and only look for crate names and versions, but the same crate+version may appear multiple times in Cargo.lock if the different registries disagree on the hash.

A simpler fix, implemented by #14716, is to read the Cargo.toml of all dependencies when vendoring them to make sure we replace all the sources.

### Notes

Now, I agree some parts of that explanation sound like shooting yourself in the foot:

- Why would anyone mirror registries in this way rather than use the same domain name? Well, in our case this was caused by a migration to a different provider, with a transition period where both addresses were still in use.
- Why would you want two different registries to serve the same crate name+version with different hashes? Now, *that* sounds like setting you up for really hard debugging sessions, but I suppose our mirror of B on x could have used `C = { registry = "x" }` to avoid this issue in the first place.

### Version

Last commit on master.

P-E-Meunier:
Minimal reproduction: https://github.com/P-E-Meunier/cargo-vendor-test

Eh2406:
[...]

PR #14822, opened 2024-11-14
Use the replacement source instead of the original source when generating `-C metadata` and `-C extra-filename`

P-E-Meunier:

This commit fixes an issue when replacing two different sources with the same local one (for example using `cargo vendor`), where `rustc` errs saying that two types "defined at the same place in the same file" are "different".

See #14821.

Use the steps described in #14821.

### Additional information

This is really useful in the context of private registries and vendored dependencies.

### DIFF
--- a/src/cargo/core/compiler/build_runner/compilation_files.rs
+++ b/src/cargo/core/compiler/build_runner/compilation_files.rs
@@ -585,6 +585,7 @@ fn compute_metadata(
     // to pull crates from anywhere without worrying about conflicts.
     unit.pkg
         .package_id()
+        .with_source_id(unit.replaced_source)
         .stable_hash(bcx.ws.root())
         .hash(&mut hasher);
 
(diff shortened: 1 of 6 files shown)


src/cargo/core/compiler/build_runner/compilation_files.rs
epage, line 1: To help us track this, #14729 hasn't been accepted yet

(we generally ask for issues to be accepted before posting PRs, see https://doc.crates.io/contrib/process/index.html)

P-E-Meunier: Oops, I meant #14821, which is an issue. I updated the initial report.

ranger-ross:
Given this PR is stale and discussion has moved into https://github.com/rust-lang/cargo/issues/14821 I am gonna close this until the design is decided.

PR #14823, opened 2024-11-15
docs: Clean up doc comments

epage:

I'm experimenting with API changes in Cargo and wanted to make it easier to find doc comments that need updating. I didn't see a lint to help with this but I did go ahead and apply the result of other documentation lints.

I've not turned these lints on more generally
- `clippy::doc_markdown` has some false positives we'd need to record in our config (`SemVer` and `SQLite`)
- `clippy::too_long_first_doc_paragraph` pointed out a doc comment that, at a quick glance, I didn't see what summary to provide

I also didn't wanted to get this in without making a policy decision of what lints are enabled.


ehuss:
Thanks! Can you bump the version of `cargo-credential` to fix the version-bump job (and any other packages that need bumping)?

ehuss:
Looks like Cargo.lock also needs updating.

ehuss:
Thanks!

@bors r+

ehuss (approved):
Oops, wrong bot.

Issue #8716, opened 2020-09-19
Reconsider RUSTFLAGS artifact caching.


ehuss:
When RUSTFLAGS is changed between invocations, this can trigger a complete rebuild. This can be annoying, and the only workaround is to use a separate target directory for each RUSTFLAGS setting. I think we should revisit how this works, and see if there is a way to address the problems.

**Motivation**:
* Changing RUSTFLAGS can be very expensive time-wise, and frustrating if your project relies on being able to swap between different settings.
* Supporting a shared build directory cache would be thwarted any time someone changes RUSTFLAGS.

**History**:
* #6503 added RUSTFLAGS to the metadata hash, allowing independent caching. 1.33 (2019-02-28)
* #6966 Removed `--remap-path-prefix` from the hash, because it broke reproducible builds. 1.37 (2019-08-15)
* #7417 Reverted #6503 because it broke profile guided optimization. 1.39 (2019-11-07)

**Solutions?**
I am uncertain what a good approach would be. Some rough ideas:

* Go back to hashing RUSTFLAGS, but only in the `-C extra-filename`, and *not* `-C metadata`. This should (probably) avoid the PGO problems, since the symbols won't change. However, this could still cause problems with reproducible builds if the rlib filename somehow leaks into the resulting binary (I don't have a reason to think it does, but it seems risky).
* Add a mechanism for specifying RUSTFLAGS so that the user can decide whether or not it gets hashed. (Maybe a separate env var, or some string in the current one, or a side setting?)  I feel like there is a problem that specifying RUSTFLAGS is already too complicated (considering the multiple overrides) and at the same time not flexible enough.
* Provide first-class support for some of the things that people use RUSTFLAGS for today.  For example, Cargo could have a "reproducible" mode where it tries to create a reproducible build, making whatever sacrifices are necessary. First class support for remap-path-prefix and PGO could also alleviate some of the problems.

jyn514:
> Provide first-class support for some of the things that people use RUSTFLAGS for today.

My main pain point with RUSTFLAGS is https://github.com/xd009642/tarpaulin/issues/416#issuecomment-623152929, which has some niche flags and wouldn't be helped much by this. AFAIK all of those flags are codegen/linking related though, and don't affect `.rlib`s.

> Add a mechanism for specifying RUSTFLAGS so that the user can decide whether or not it gets hashed.

[...]

PR #14826, opened 2024-11-15
refactor(fingerprint): Track the intent for each use of `UnitHash`

epage:

We have
- `-C metadata`
- `-C extra-filename`
- Uniquifying build scripts
- Uniquifying scraped documentation

Currently, these are all the same value.  For #8716, we might want to have `-C metadata` and `-C extra-filename` hash different parts of the `Unit`.  I figured that this change helps to clarify intent so that even if we don't change the definitions, this could still be worth it.

The last two I'm tracking as a `unit_id`.  As we evolve the first two
hashes, we can decide which would be a better fit for backing the `unit_id`.
For scraping,  I could have treated this as
`-C extra-filename` but then we'd have a lot of `.expect()`s.

I moved the accessors from `CompilationFiles` to `MetaInfo` so we could closely associate the documentation, API, and implementation.  Making `CompilationFiles::c_metadata` the main entry point that gets documented would be weird because the hashing is associated with `MetaInfo` (umbrella type, was private, now `Metadata`) or `Metadata` (agnostic of each use, now `UnitHash`)

### DIFF
--- a/src/cargo/core/compiler/build_runner/compilation_files.rs
+++ b/src/cargo/core/compiler/build_runner/compilation_files.rs
@@ -671,7 +671,7 @@ fn compute_metadata(
 
     MetaInfo {
         meta_hash: Metadata(hasher.finish()),
-        use_extra_filename: should_use_metadata(bcx, unit),
+        use_extra_filename: use_extra_filename(bcx, unit),
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/cargo/core/compiler/build_runner/compilation_files.rs
```diff
@@ -25,6 +25,21 @@ use crate::util::{self, CargoResult, StableHasher};
 /// use the same rustc version.
 const METADATA_VERSION: u8 = 2;
 
+#[derive(Copy, Clone, Hash, Eq, PartialEq, Ord, PartialOrd)]
```
weihanglo, line 28: Since this is exposed to other functions as returning type, this may worth a doc comment. Something like this:

```suggestion
/// A hash value uniquely identifying a [`Unit`] in a build. See [`Metadata`] for more.
#[derive(Copy, Clone, Hash, Eq, PartialEq, Ord, PartialOrd)]
```

weihanglo (approved):
Thanks!

PR #14829, opened 2024-11-15
you we distinction

KGrewal1:
Cleaning up the usage of we vs you in the guide section of the book for greater consistency and clarity.

You should now refer to the reader and we to the authors of the book

### DIFF
--- a/src/doc/src/guide/cargo-toml-vs-cargo-lock.md
+++ b/src/doc/src/guide/cargo-toml-vs-cargo-lock.md
@@ -16,8 +16,8 @@ We recommend pairing this with
 
 Let’s dig in a little bit more.
 
-`Cargo.toml` is a [**manifest**][def-manifest] file in which we can specify a
-bunch of different metadata about our package. For example, we can say that we
+`Cargo.toml` is a [**manifest**][def-manifest] file in which you can specify a
+bunch of different metadata about your package. For example, you can say that you
 depend on another package:
 
 ```toml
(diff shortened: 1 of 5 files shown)


Rustin170506:
I'm not sure which one is better, but using "we" is also quite common in tutorial or guide-type documents. Anyway, thanks for your contribution. Also, you need to squash your second commit. Since Cargo merges code using the `merge` method, it seems like the second commit could be merged into the first one!

KGrewal1:
> I'm not sure which one is better, but using "we" is also quite common in tutorial or guide-type documents. Anyway, thanks for your contribution. Also, you need to squash your second commit. Since Cargo merges code using the `merge` method, it seems like the second commit could be merged into the first one!

Squashed into one commit now - aim was to make this more consistent between pages / paragraphs - advantage of 'you / we' is greater clarity in who is being referred to, but agree that we for everything is also common - just ideally whichever choice should be consistent

ehuss:
Microsoft's style guide suggests [Avoid plural first person](https://learn.microsoft.com/en-us/style-guide/grammar/person#avoid-plural-first-person-we-us), which I agree with.

weihanglo (commented):
https://github.com/rust-lang/cargo/blob/e858736b561805af163ad9dbe0dbd30aa889701f/src/doc/src/guide/working-on-an-existing-project.md?plain=1#L6

This also need a tweak?

KGrewal1:
Couldn't think of a non clunky way to remove the pronoun here, "you'll use" sounds imperative and worse, "let's use" is still 1st person plural, and removing the pronoun (In this example, using `regex`) I thought was subjectively worse

weihanglo:
That's true. Let's merge this now, as people seem to like this :)

Issue #14506, opened 2024-09-06
Make empty/default fields in the index metadata optional


kornelski:
### Problem

It is a bit surprising that the index metadata needs explicit `"optional": false` and `"default_features":true` for every dependency, even though these fields have defaults and are optional in manifests.

It's also odd that `features: {}` must still be present in the serialized JSON even when using only `features2` field. 

Making these fields optional would make it a bit easier to serialize index metadata, and can make the index files a bit smaller (for packages with high enough MSRV).

### Proposed Solution

https://github.com/rust-lang/cargo/pull/14491

```diff
--- a/src/cargo/sources/registry/index/mod.rs
+++ b/src/cargo/sources/registry/index/mod.rs
@@ -206,6 +206,7 @@ pub struct IndexPackage<'a> {
     #[serde(borrow)]
     pub deps: Vec<RegistryDependency<'a>>,
     /// Set of features defined for the package, i.e., `[features]` table.
+    #[serde(default)]
     pub features: BTreeMap<InternedString, Vec<InternedString>>,
     /// This field contains features with new, extended syntax. Specifically,
     /// namespaced features (`dep:`) and weak dependencies (`pkg?/feat`).
@@ -268,10 +269,13 @@ pub struct RegistryDependency<'a> {
     #[serde(borrow)]
     pub req: Cow<'a, str>,
     /// Set of features enabled for this dependency.
+    #[serde(default)]
     pub features: Vec<InternedString>,
     /// Whether or not this is an optional dependency.
+    #[serde(default)]
     pub optional: bool,
     /// Whether or not default features are enabled.
+    #[serde(default = "default_true")]
     pub default_features: bool,
     /// The target platform for this dependency.
     pub target: Option<Cow<'a, str>>,
@@ -292,6 +296,10 @@ pub struct RegistryDependency<'a> {
     pub lib: bool,
 }

+fn default_true() -> bool {
+    true
+}
+
```

epage:
> Making these fields optional would make it a bit easier to serialize index metadata, 

What makes it easier?

> can make the index files a bit smaller (for packages with high enough MSRV).

This is where things get tricky.

We could only do this when we know the MSRV or a registry looks at `Cargo.toml` to infer an MSRV (e.g. from `package.edition`).  That means this is opt-in which can limit the benefit registries would see from it.

[...]

PR #14838, opened 2024-11-19
Allow registries to omit empty/default fields in JSON

kornelski:
Applies the patch from #14506 + documentation update.

The `kind` field has already been optional in the implementation, but documented as possibly missing due to bugs. I've changed the documentation to simply state it's optional.

### DIFF
--- a/src/doc/src/reference/registry-index.md
+++ b/src/doc/src/reference/registry-index.md
@@ -140,26 +140,24 @@ explaining the format of the entry.
             "req": "^0.6",
             // Array of features (as strings) enabled for this dependency.
             "features": ["i128_support"],
-            // Boolean of whether or not this is an optional dependency.
+            // Boolean of whether or not this is an optional dependency. Defaults to `false` if not specified.
             "optional": false,
-            // Boolean of whether or not default features are enabled.
+            // Boolean of whether or not default features are enabled. Defaults to `true` if not specified.
             "default_features": true,
             // The target platform for the dependency.
-            // null if not a target dependency.
+            // If not specified or `null`, it is not a target dependency.
             // Otherwise, a string such as "cfg(windows)".
             "target": null,
             // The dependency kind.
             // "dev", "build", or "normal".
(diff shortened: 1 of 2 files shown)


epage (approved):
Thanks!

src/doc/src/reference/registry-index.md
```diff
@@ -140,26 +140,24 @@ explaining the format of the entry.
             "req": "^0.6",
             // Array of features (as strings) enabled for this dependency.
             "features": ["i128_support"],
-            // Boolean of whether or not this is an optional dependency.
+            // Boolean of whether or not this is an optional dependency. Defaults to `false` if not specified.
```
Eh2406, line 143: Can this be more precise? Something like:
```
If not specified:
- cargo >= 1.8x: defaults this as `false`.
- cargo >= 1.19: ignore the one version.
- cargo <= 1.18: ignore the entire package.
```
Or
```
Current versions of cargo defaults to `false` if not specified.
Cargos before 1.8x the version will be ignored.
Very old cargos before 1.19 ignore the entire package.
```
When someone looks to skip this data on crates.io, this is where I would look to see if it is ok. So I want to leave a note that it is not ok for the git index.

src/doc/src/reference/registry-index.md
```diff
@@ -140,26 +140,24 @@ explaining the format of the entry.
...
             "optional": false,
-            // Boolean of whether or not default features are enabled.
+            // Boolean of whether or not default features are enabled. Defaults to `true` if not specified.
```
Eh2406, line 145: Same as above.

src/doc/src/reference/registry-index.md
```diff
@@ -140,26 +140,24 @@ explaining the format of the entry.
             "req": "^0.6",
             // Array of features (as strings) enabled for this dependency.
             "features": ["i128_support"],
```
Eh2406, line 142: This one also needs a comment.

Eh2406 (commented):
I guess I am to late with this comment.

PR #14843, opened 2024-11-20
tests: refactor testsuite `cargo build --target`

harmou01:

Mentioned [here](https://github.com/rust-lang/cargo/pull/14183#discussion_r1838979612), this PR is a small refactor to add three additional functions to `Execs` in order to pass a target triple, host target or target directory more cleanly when running a test case.

### DIFF
--- a/tests/testsuite/build_script.rs
+++ b/tests/testsuite/build_script.rs
@@ -457,7 +457,7 @@ fn custom_build_env_var_rustc_linker() {
 
     // no crate type set => linker never called => build succeeds if and
     // only if build.rs succeeds, despite linker binary not existing.
-    p.cargo("build --target").arg(&target).run();
+    p.cargo("build").target(&target).run();
 }
 
 // Only run this test on linux, since it's difficult to construct
@@ -494,7 +494,7 @@ fn custom_build_env_var_rustc_linker_with_target_cfg() {
 
     // no crate type set => linker never called => build succeeds if and
     // only if build.rs succeeds, despite linker binary not existing.
-    p.cargo("build --target").arg(&target).run();
+    p.cargo("build").target(&target).run();
 }
 
 #[cargo_test]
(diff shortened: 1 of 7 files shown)


tests/testsuite/custom_target.rs
```diff
@@ -140,7 +145,10 @@ fn custom_bin_target() {
         .file("custom-bin-target.json", SIMPLE_SPEC)
         .build();
 
-    p.cargo("build --target custom-bin-target.json -v").run();
+    p.cargo("build")
```
weihanglo, line 148: One drawback of it is that the argument list is spliited into multiple lines of indirection, making it harder to read at first glance.

tests/testsuite/custom_target.rs
```diff
@@ -53,8 +53,13 @@ fn custom_target_minimal() {
         .file("custom-target.json", SIMPLE_SPEC)
         .build();
 
-    p.cargo("build --lib --target custom-target.json -v").run();
-    p.cargo("build --lib --target src/../custom-target.json -v")
+    p.cargo("build --lib")
```
weihanglo, line 56: So does this one. The original one is more readable IMO.

tests/testsuite/build_script.rs
```diff
@@ -494,7 +494,7 @@ fn custom_build_env_var_rustc_linker_with_target_cfg() {
 
     // no crate type set => linker never called => build succeeds if and
     // only if build.rs succeeds, despite linker binary not existing.
-    p.cargo("build --target").arg(&target).run();
+    p.cargo("build").target(&target).run();
```
weihanglo, line 497: This kind of dynamic one is something in my mind could benefit from this change, though I haven't seen other 100+ occurrences changed in this pull request.

Upon this point, maybe this refactor might not be worthy with this scale of churn on your side?
(updating 100+ occurrences of `--target` sounds tedious)

weihanglo:
Just chatted with the PR author on Zulip. This is likely a bit too tedious to change 100+ occurrences all at once. They are not going to chase this down at this cmoment and instead focusing on the original issue #14183. Thanks for their contribution anyway :)

PR #14846, opened 2024-11-20
test(rustflags): Verify -Cmetadata directly, not through -Cextra-filename

epage:

This unblocks experimenting with having these diverge for #8716

### DIFF
--- a/tests/testsuite/rustflags.rs
+++ b/tests/testsuite/rustflags.rs
@@ -4,7 +4,10 @@ use std::fs;
 
 use cargo_test_support::prelude::*;
 use cargo_test_support::registry::Package;
-use cargo_test_support::{basic_manifest, paths, project, project_in_home, rustc_host, str};
+use cargo_test_support::{
+    basic_manifest, paths, project, project_in_home, rustc_host, str, RawOutput,
+};
+use snapbox::assert_data_eq;
 
 #[cargo_test]
 fn env_rustflags_normal_source() {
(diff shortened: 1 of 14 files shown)


tests/testsuite/rustflags.rs
```diff
@@ -1478,36 +1481,44 @@ fn env_rustflags_misspelled_build_script() {
...
+
+    let build_output = p
+        .cargo("build -v")
         .env(
             "RUSTFLAGS",
             "--remap-path-prefix=/abc=/zoo --remap-path-prefix /spaced=/zoo",
         )
         .run();
-    check_metadata_same();
+    let build_c_metadata = dbg!(get_c_metadata(build_output));
```
weihanglo, line 1512:
```suggestion
    let build_c_metadata = get_c_metadata(build_output);
```

tests/testsuite/rustflags.rs
```diff
@@ -1478,36 +1481,44 @@ fn env_rustflags_misspelled_build_script() {
...
-    assert_eq!(rlibs.len(), 1);
-    p.cargo("clean").run();
-
-    let check_metadata_same = || {
-        let rlibs2 = p
-            .glob("target/debug/deps/*.rlib")
-            .collect::<Result<Vec<_>, _>>()
-            .unwrap();
-        assert_eq!(rlibs, rlibs2);
+    let get_c_metadata_re =
```
weihanglo, line 1484: Generally I think a regex worth a commment to explain itself.

tests/testsuite/rustflags.rs
```diff
@@ -1478,36 +1481,44 @@ fn env_rustflags_misspelled_build_script() {
...
     p.cargo("clean").run();
-    p.cargo("rustc -- --remap-path-prefix=/abc=/zoo --remap-path-prefix /spaced=/zoo")
+
+    let rustc_output = p
+        .cargo("rustc -v -- --remap-path-prefix=/abc=/zoo --remap-path-prefix /spaced=/zoo")
         .run();
-    check_metadata_same();
+    let rustc_c_metadata = dbg!(get_c_metadata(rustc_output));
+
+    Assert::new().eq(rustc_c_metadata, build_c_metadata);
```
weihanglo, line 1521: This is the first direct usage of `snapbox::Assert` in our test suite. When should we use it instead `assert_e2e` or `assert_ui`.

tests/testsuite/rustflags.rs
epage: I intentionally left `dbg` statements in so people can vouch for the output thats being asserted since the assertion's expected result is another command-under-test which is questionable.  The very first time I ran this test, it passed.  The output was blank.  I added the assert for that reason but figured it would still be good to be able to verify it.

(lots of other tests print results as they go)

tests/testsuite/rustflags.rs
epage: For very complex regexes, I can see that.  Maybe its a curse of knowledge thing but this seems relatively straightforward and I feel like any comment I write would just be repeating what the code does (extract two parameters from the CLI).

tests/testsuite/rustflags.rs
epage: I didn't want to risk the test becoming invalidated because it was asserting a `[HASH]` equaled a `[HASH]` that both evaluated to different values.

There is no need for any redactions on this test because we extract only what we need.

weihanglo: [`Assert::new().eq`](https://docs.rs/snapbox/0.6.20/snapbox/assert/struct.Assert.html#method.eq) still applies some normalizations. If we want a plain old assertion, should we use `assert_eq!` instead? I think it is the choice of assertion function confusing me.

tests/testsuite/rustflags.rs
weihanglo: Fair enough!

Will it be better that we adds extra leading space so flags are less likely to be captured from weird place like `my--crate-name`?
Also `+` to ensure we actually capture something?

(Surprisingly `rustc -C metadata= src/main.rs` works`)

```rust
        regex::Regex::new(r".* (--crate-name [^ ]+).* (-C ?metadata=[^ ]+).*").unwrap();
```

tests/testsuite/rustflags.rs
weihanglo: Nah, that's fine.

weihanglo (approved):

PR #14848, opened 2024-11-21
test: Switch from 'exec_with_output' to 'run'

epage:

This is a follow up to #14846 which changed `run` to return the `RawOutput`.

Reasons I didn't "update" some code to the new `run` return value
- We were actually using `ProcessBuilder::exec_with_output` and I didn't want to disentangle what it would take to switch to `Execs`
- We did processing on the `Result` and I didn't want to check how that could be updated


weihanglo (approved):
Thanks!

PR #14857, opened 2024-11-25
feat(toml): Allow adding/removing from cargo scripts

epage:

This adds support for cargo script for more cargo commands and is a part of #12207

Unfortunately, there is a double-warning for unspecified editions for `cargo remove`.  Rather than addressing that here, I'll be noting it in the tracking issue.

### DIFF
--- a/src/cargo/util/toml_mut/manifest.rs
+++ b/src/cargo/util/toml_mut/manifest.rs
@@ -11,6 +11,7 @@ use crate::core::dependency::DepKind;
 use crate::core::{FeatureValue, Features, Workspace};
 use crate::util::closest;
 use crate::util::interning::InternedString;
+use crate::util::toml::{is_embedded, ScriptSource};
 use crate::{CargoResult, GlobalContext};
 
 /// Dependency table to add deps to.
@@ -245,6 +246,10 @@ pub struct LocalManifest {
     pub path: PathBuf,
     /// Manifest contents.
     pub manifest: Manifest,
+    /// The raw, unparsed package file
+    pub raw: String,
+    /// Edit location for an embedded manifest, if relevant
+    pub embedded: Option<Embedded>,
 }
 
 impl Deref for LocalManifest {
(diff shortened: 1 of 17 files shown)


src/cargo/util/toml/embedded.rs
```diff
@@ -189,94 +189,112 @@ fn sanitize_name(name: &str) -> String {
...
+        // See rust-lang/rust's compiler/rustc_lexer/src/lib.rs's `strip_shebang`
+        // Shebang must start with `#!` literally, without any preceding whitespace.
+        // For simplicity we consider any line starting with `#!` a shebang,
+        // regardless of restrictions put on shebangs by specific platforms.
+        if let Some(rest) = source.content.strip_prefix("#!") {
+            // Ok, this is a shebang but if the next non-whitespace token is `[`,
+            // then it may be valid Rust code, so consider it Rust code.
+            if rest.trim_start().starts_with('[') {
+                return Ok(source);
+            }
```
ehuss, line 217: Just a comment: This isn't quite the same behavior, as the compiler allows comments before the opening `[`. I realize parsing that would be difficult, and that particular scenario is a bit silly.

src/cargo/util/toml_mut/manifest.rs
```diff
@@ -267,18 +270,56 @@ impl LocalManifest {
...
     pub fn write(&self) -> CargoResult<()> {
-        let s = self.manifest.data.to_string();
-        let new_contents_bytes = s.as_bytes();
+        let mut manifest = self.manifest.data.to_string();
+        let raw = match self.embedded.as_ref() {
+            Some(Embedded::Implicit(start)) => {
+                if !manifest.ends_with("\n") {
+                    manifest.push_str("\n");
+                }
+                let fence = "---\n";
```
ehuss, line 306: Just a comment: These look like it will cause mixed line endings on a CRLF file. However, that looks like it is an existing problem with `cargo add` in general.

src/cargo/util/toml_mut/manifest.rs
```diff
@@ -245,6 +246,8 @@ pub struct LocalManifest {
     pub path: PathBuf,
     /// Manifest contents.
     pub manifest: Manifest,
+    pub raw: String,
+    pub embedded: Option<Embedded>,
```
ehuss, line 250: Can you add docstrings to these explaining these new fields?

src/cargo/util/toml_mut/manifest.rs
```diff
@@ -531,6 +572,44 @@ impl std::fmt::Display for LocalManifest {
     }
 }
 
+#[derive(Clone, Debug)]
+pub enum Embedded {
+    Implicit(usize),
+    Explicit(std::ops::Range<usize>),
+}
```
ehuss, line 579: Can you add docstrings on this enum and its variants explaining what they are?

src/cargo/util/toml_mut/manifest.rs
```diff
@@ -267,18 +270,56 @@ impl LocalManifest {
...
-        let new_contents_bytes = s.as_bytes();
+        let mut manifest = self.manifest.data.to_string();
+        let raw = match self.embedded.as_ref() {
+            Some(Embedded::Implicit(start)) => {
+                if !manifest.ends_with("\n") {
+                    manifest.push_str("\n");
+                }
+                let fence = "---\n";
+                let prefix = &self.raw[0..*start];
+                let suffix = &self.raw[*start..];
```
ehuss, line 308: Only if you want, either form seems fine:

```suggestion
                let (prefix, suffix) = self.raw.split_at(*start);
```

src/cargo/util/toml/embedded.rs
epage: This is existing code, just moved.  Ok if we put this in the tracking issue or an issue and track separately?

src/cargo/util/toml_mut/manifest.rs
epage: As this is another manifestation of an existing issue, are we ok with deferring this out of this PR?

ehuss (approved):
Thanks!

src/cargo/util/toml_mut/manifest.rs
epage: Created #14863 and noted this in it

src/cargo/util/toml/embedded.rs
epage: Added this to the checklist: #12207

PR #14830, opened 2024-11-16
fix(fingerprint): Don't throwaway the cache on RUSTFLAGS changes 

epage:
### What does this PR try to resolve?

Fixes #8716

This splits `-C metdata` and `-C extra-filename` and adds `RUSTFLAGS` to `-C extra-filename` in the hope that we can still make PGO and reproducible builds work while caching artifacts per RUSTFLAGS used.

Issue #7416, opened 2019-09-23
PGO cannot be used via cargo because of how `RUSTFLAGS` are handled.


michaelwoerister:
**Problem**

PGO works in two phases: 

1. Compile an instrumented binary of your program that enables collecting profile data. This is done via `RUSTFLAGS=-Cprofile-generate`.
2. Compile an optimized binary that makes use of the profile data collected in step (1). This is done via `RUSTFLAGS=-Cprofile-use`.

In order for this to work, the symbol names within the two program versions must match up because LLVM generates profiling data in terms of symbol names. 

However, since https://github.com/rust-lang/cargo/pull/6503, `RUSTFLAGS` are fed into the `-Cmetadata` argument to `rustc`. This causes symbol names to differ because one version has `-Cprofile-generate` and the other has `-Cprofile-use` in their `RUSTFLAGS`. 

I think I would prefer not feeding `RUSTFLAGS` into `-C metadata` at all. It does not seem right to me that symbol names are affected by random command line arguments to the compiler.

alexcrichton:
I've proposed a revert of #6503 at https://github.com/rust-lang/cargo/pull/7417 to fix this.

@michaelwoerister out of curiosity was this the cause of why y'all weren't seeing great wins for PGO in Firefox?

michaelwoerister:
@alexcrichton I have confirmed that this issue prevents Firefox from getting any PGO wins, yes. However, I don't know yet if it is the only issue (e.g. GNU ld seems to have a bug that sometimes causes instrumented builds to not record anything 😱). 

I have some microbenchmarks that show a 10-15% win with PGO but have not been able to replicate that with the `regex` benchmark suite, which is my next bigger test case.

alexcrichton:
Oh dear that must have been absolutely gnarly to figure this out, sorry about that!

michaelwoerister:
It's exactly the kind of issue I expected to encounter `:)`

PR #14859, opened 2024-11-26
test(pgo): ensure PGO works

weihanglo:

This is a regression test to prevent issues like #7416.

The test only run on Linux,
as other platforms have different requirements for PGO,
or emit different PGO function missing warnings.

Not sure how brittle it is. We can optionally run it only on Cargo's CI?

cc #14830

### DIFF
--- a/tests/testsuite/pgo.rs
+++ b/tests/testsuite/pgo.rs
@@ -1,0 +1,111 @@
+//! Test if PGO works.
+
+use std::path::PathBuf;
+use std::process::Command;
+
+use cargo_test_support::prelude::*;
+use cargo_test_support::project;
+use cargo_test_support::str;
+
+fn llvm_profdata() -> Option<PathBuf> {
+    let output = Command::new("rustc")
+        .arg("--print=target-libdir")
+        .output()
+        .expect("rustc to run");
+    assert!(output.status.success());
+    let mut libdir = PathBuf::from(String::from_utf8(output.stdout).unwrap());
+    assert!(libdir.pop());
+    let mut bin = libdir.join("bin").join("llvm-profdata");
+    bin.exists().then(|| bin.clone()).or_else(|| {
+        bin.set_extension("exe");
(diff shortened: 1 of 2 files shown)


epage (approved):

tests/testsuite/pgo.rs
```diff
@@ -0,0 +1,111 @@
...
+#[cargo_test]
+fn pgo_works() {
+    if cfg!(not(target_os = "linux")) {
+        // macOS may emit different LLVM PGO warnings.
+        // Windows LLVM has different requirements.
+        return;
+    }
+
+    let Some(llvm_profdata) = llvm_profdata() else {
+        return;
```
ehuss, line 34: Would it be possible to enforce that this exists in CI? Otherwise we don't know if this test is ever actually running.

tests/testsuite/pgo.rs
```diff
@@ -0,0 +1,111 @@
...
+    })
+}
+
+#[cargo_test]
+fn pgo_works() {
+    if cfg!(not(target_os = "linux")) {
+        // macOS may emit different LLVM PGO warnings.
+        // Windows LLVM has different requirements.
+        return;
+    }
```
ehuss, line 31: Can you make sure to use the `ignore` attribute when disabling a test? Otherwise it can be confusing when tests don't seem to change when you expect them to, are say that they are passing when you expect them to fail.

I realize there's still a big hole when llvm-profdata is missing. We could maybe add that to our requires mechanism?

```suggestion
#[cargo_test]
// macOS may emit different LLVM PGO warnings.
// Windows LLVM has different requirements.
#[cfg_attr(not(target_os = "linux"), ignore = "linux only")]
fn pgo_works() {
```

tests/testsuite/pgo.rs
weihanglo: Yeah I've considered that and found it needs more work than the current one.

I'll try to improve it!

PR #14860, opened 2024-11-26
git-fetch-with-cli: Set `GIT_DIR` for bare repository compatibility

### LINKED ISSUE #14758: net.git-fetch-with-cli fails with "cannot use bare repository" with `safe.bareRepository = explicit`
### Problem

When using `net.git-fetch-with-cli = true` with a sufficiently new git, it fails to clone the git repository into the cache because `safe.bareRepository` defaults to `explicit`. 

With `Cargo.toml`:
```
[package]
name = "git-cli-test"
version = "0.1.0"
edition = "2021"

[dependencies]
cargo = { git = "https://github.com/rust-lang/cargo.git" }
```

I get 
```
    Updating git repository `https://github.com/rust-lang/cargo.git`
fatal: cannot use bare repository '/usr/local/google/home/jamuraa/.cargo/git/db/cargo-e7ff1db891893a9e' (safe.bareRepository is 'explicit')
error: failed to get `cargo` as a dependency of package `git-cli-test v0.1.0 (/usr/local/google/home/jamuraa/tmp/git-cli-test)`

Caused by:
  failed to load source for dependency `cargo`

Caused by:
  Unable to update https://github.com/rust-lang/cargo.git

Caused by:
  failed to clone into: /usr/local/google/home/jamuraa/.cargo/git/db/cargo-e7ff1db891893a9e

Caused by:
  process didn't exit successfully: `git fetch --force --update-head-ok 'https://github.com/rust-lang/cargo.git' '+HEAD:refs/remotes/origin/HEAD'` (exit status: 128)
```

This is likely due to the mitigation for https://github.com/justinsteven/advisories/blob/main/2022_git_buried_bare_repos_and_fsmonitor_various_abuses.md

### Steps

1. set `net.git-fetch-with-cli = true`
2. Add a git source to Cargo.toml
3. `cargo update` (or any other command that will fetch the git repo

### Possible Solution(s)

It's likely that adding `--git-dir` to the git command line will get around this, since cargo is using bare repositories on purpose.

### Notes

_No response_

### Version

cargo 1.81.0-nightly (154fdac39 2024-07-07)
release: 1.81.0-nightly
commit-hash: 154fdac39ae9629954e19e9986fd2cf2cdd8d964
commit-date: 2024-07-07
libcurl: 8.8.0-DEV (sys:0.4.73+curl-8.8.0 vendored ssl:OpenSSL/1.1.1w)
os: Debian n/a (rodete) [64-bit]

charmitro:
When using `net.git-fetch-with-cli = true` , Cargo fails to clone the git repository into the cache because `safe.bareRepository` defaults to `explicit`. This results in an error stating that a bare repository cannot be used.

This patch sets the `GIT_DIR` environment variable to the correct repository path instead of removing it. This ensures that Git uses the correct repository directory and avoids the "cannot use bare repository" error.

Fixes #14758

### DIFF
--- a/src/cargo/sources/git/utils.rs
+++ b/src/cargo/sources/git/utils.rs
@@ -1101,9 +1101,8 @@ fn fetch_with_cli(
         .args(refspecs)
         // If cargo is run by git (for example, the `exec` command in `git
         // rebase`), the GIT_DIR is set by git and will point to the wrong
-        // location (this takes precedence over the cwd). Make sure this is
-        // unset so git will look at cwd for the repo.
-        .env_remove("GIT_DIR")
+        // location. This makes sure GIT_DIR is always the repository path.
+        .env("GIT_DIR", repo.path())
         // The reset of these may not be necessary, but I'm including them
         // just to be extra paranoid and avoid any issues.
         .env_remove("GIT_WORK_TREE")


src/cargo/sources/git/utils.rs
```diff
@@ -1099,11 +1099,8 @@ fn fetch_with_cli(
         .arg("--update-head-ok") // see discussion in #2078
         .arg(url)
         .args(refspecs)
-        // If cargo is run by git (for example, the `exec` command in `git
-        // rebase`), the GIT_DIR is set by git and will point to the wrong
-        // location (this takes precedence over the cwd). Make sure this is
-        // unset so git will look at cwd for the repo.
-        .env_remove("GIT_DIR")
+        // Set the GIT_DIR environment variable to the repository path
```
weihanglo, line 1102: This comment is not really meaningful, as it doesn't describes why but what it does. Perhaps something like this?

```suggestion
        // If cargo is run by git (for example, the `exec` command in `git
        // rebase`), the GIT_DIR is set by git and will point to the wrong
        // location. This makes sure GIT_DIR is always the repository path.
```

weihanglo (commented):
Thanks for the patch!

src/cargo/sources/git/utils.rs
charmitro: You're right. Replaced with your suggestion!

weihanglo (approved):

PR #14861, opened 2024-11-26
test(build-std): download deps first

weihanglo:

Download dependencies first,
So that we can assert the full output of `cargo test` without wildcard.

This regressed since #14850

CI passes.

### DIFF
--- a/tests/build-std/main.rs
+++ b/tests/build-std/main.rs
@@ -379,6 +379,9 @@ fn test_proc_macro() {
         .file("src/lib.rs", "")
         .build();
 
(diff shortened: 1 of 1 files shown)


weihanglo:
We can alternatively remove these two lines, though I am not ready to get deeper into build-std test strategy now.

https://github.com/rust-lang/cargo/blob/8ba3ec289534bb25332f7882d1faae4fe312eb8c/tests/build-std/main.rs?plain=1#L29-L30

epage (approved):

PR #14864, opened 2024-11-27
fix(fix): Migrate cargo script manifests across editions

epage:

At this point it is unlikely for a script to be written for pre-2024
edition and migrated to 2024+ but this code is independent of Editions
so this means we have one less bug and are better prepared for the next
edition.

When we add `cargo fix` support for regular manifest warnings, we'll
need to take into account cargo scripts.

This is a part of #12207.

### Additional information

While looking for where the tests go, I found a couple places with a hard coded latest-edition in test output and fixed it.


ehuss (approved):
Thanks!

PR #14865, opened 2024-11-28
enhance(build): finish to log compilied pkg.

damonxue:
This pull request includes a little change to the `DrainState` implementation in the `src/cargo/core/compiler/job_queue/mod.rs` file. The change enhances the logging and tracking of the job queue's progress by adding detailed logging for different types of artifacts.

Enhancements to logging and tracking:

* [`src/cargo/core/compiler/job_queue/mod.rs`](diffhunk://#diff-760fbef71d557f3e4344d337e6b6e81c6f860bb32498b751d015d26f9a5951cfL1121-R1136): Added detailed logging for the completion of a unit of work, including specific statuses for documentation, scraping, and compilation. This helps in tracking the progress of the job queue more effectively.

### DIFF
--- a/src/cargo/core/compiler/job_queue/mod.rs
+++ b/src/cargo/core/compiler/job_queue/mod.rs
@@ -1125,13 +1125,21 @@ impl<'gctx> DrainState<'gctx> {
                 debug!("Finished unit: {:?}", unit);
                 if unit.mode.is_doc() {
                     self.documented.insert(unit.pkg.package_id());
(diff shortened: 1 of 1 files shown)


weihanglo:
Saw you just closed it, though thanks for the contribution anyway!

The major issue of console output overhaul is <https://github.com/rust-lang/cargo/issues/8889>, which you might be interested in. One of complaint is the current output too noisy, which is on the opposite side of this PR.

PR #14799, opened 2024-11-08
Revert "feat: Add `CARGO_RUSTC_CURRENT_DIR`"

epage:
### What does this PR try to resolve?

This reverts commit e81d84cc17d7df6a249e60cbc620f25edbe782db.

We decided to not stabilize this in #13644.

### Additional information

This does not fix the underlying issue of #14798

PR #14869, opened 2024-11-29
test: remove the use of `CARGO_RUSTC_CURRENT_DIR`

weihanglo:

The environment variable has been removed in #14799.
Switch to plain old `CARGO_MANIFEST_DIR`
because snapbox's `current_rs` favors furthest Cargo.toml,
blocking tests running in rust-lang/rust as a submodule.

See also build failures in rust-lang/rust#133533

Pull this change into rust-lang/rust and run

```
./x.py test src/tools/cargo
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -455,7 +455,7 @@ version = "0.3.4"
 
 [[package]]
 name = "cargo-test-support"
-version = "0.6.1"
+version = "0.7.0"
 dependencies = [
  "anstream",
  "anstyle",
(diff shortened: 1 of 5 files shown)


epage:
Snapbox also uses `CARGO_RUSTC_CURRENT_DIR`.  However, that is only on test failure when `SNAPSHOTS=overwrite` is specified which is why it isn't a problem.

epage:
Do we know why this doesn't work for bootstrap but it works for everything else?  Are they running our tests outside of our repo, making it so the config isn't loaded?  Could that change?

weihanglo:
> Do we know why this doesn't work for bootstrap but it works for everything else? Are they running our tests outside of our repo, making it so the config isn't loaded? Could that change?

I believe that is because bootstrap invokes cargo with `--manifest-path` without `cd` to the package root.

weihanglo:
> […] Are they running our tests outside of our repo, making it so the config isn't loaded? Could that change?

It might be possible to change, but that requires bootstrap team's involvement. I would suggest we unblock this first, and then discuss with other folks for any further refactors.

weihanglo:
Suggest a patch to that PR: <https://github.com/rust-lang/rust/pull/133533#issuecomment-2508666180>.

weihanglo:
Close as [this patch](https://github.com/rust-lang/cargo/pull/14869#issuecomment-2508666546) works.

PR #14873, opened 2024-11-29
Introduce `CARGO_PKG_EDITION` env var

### LINKED ISSUE #14872: Export `CARGO_PKG_EDITION` variable to build environment
### Problem

Some tools like `bindgen` may need to know the Rust language edition of the root crate, especially when executing as part of `build.rs`.  For example, `bindgen` needs to know if the -sys package uses edition 2021 (so it can use C-string literals), or older (in which case it must generate some unsafe code with the same literals).  Requiring the user to always pass in bindgen args that match their environment is less than ideal if the default could be computed on the fly using environment variables.

See https://github.com/rust-lang/rust-bindgen/pull/3000

### Proposed Solution

Publish `CARGO_PKG_EDITION` or `CARGO_PKG_RUST_EDITION` or `CARGO_PKG_LANG_EDITION` or ... (naming tbd)

### Notes

_No response_

nyurik:
Add `CARGO_PKG_EDITION` environment variable, allowing build scripts to determine the language edition of the crate being compiled.

Fixes #14872


epage:
As a heads up, our contribution process asks that issues be accepted before moving on to PRs. There are two directions to go and this direction has non-obvious problems to work through. We should work through this first before looking at the implementation.

crates/build-rs/src/input.rs
```diff
@@ -536,6 +536,12 @@ pub fn cargo_pkg_license_file() -> Option<PathBuf> {
     to_opt(var_or_panic("CARGO_PKG_LICENSE_FILE")).map(to_path)
 }
 
+/// The Rust language edition from the manifest of your package.
+#[track_caller]
+pub fn cargo_pkg_edition() -> Option<String> {
```
epage, line 541: This should probably list MSRV information

src/doc/src/reference/environment-variables.md
```diff
@@ -239,6 +239,7 @@ corresponding environment variable is set to the empty string, `""`.
 * `CARGO_PKG_REPOSITORY` --- The repository from the manifest of your package.
 * `CARGO_PKG_LICENSE` --- The license from the manifest of your package.
 * `CARGO_PKG_LICENSE_FILE` --- The license file from the manifest of your package.
+* `CARGO_PKG_EDITION` --- The Rust language edition from the manifest of your package.
```
epage, line 242: This should list the MSRV.

epage, line 242: Should this also warn people about the deprecated behavior?

nyurik:
@epage i rebased this PR. WRT your MSRV comments --- what kind of message do you think would be needed, and which version? Thx!

crates/build-rs/src/input.rs
epage: Looking at #14902, this should be
```rust
/// The Rust language edition from the manifest of your package.
#[doc = requires_msrv!("1.88")]
#[track_caller]
pub fn cargo_pkg_edition() -> String {
```

src/doc/src/reference/environment-variables.md
epage: Hmm, looks like #14902 and #14404 didn't add MSRV messages.

We can hold off on this for now because this may require formatting changes, etc.

nyurik:
It seems the smoke test does not work until after this change is released as part of cargo, so commenting it out.  Otherwise seems to work ok?  I updated the MSRV message, but not certain about @epage's "deprecated behavior" [comment](https://github.com/rust-lang/cargo/pull/14873#discussion_r2040045778)

crates/build-rs/src/input.rs
```diff
@@ -580,6 +580,13 @@ pub fn cargo_pkg_description() -> Option<String> {
     to_opt(var_or_panic("CARGO_PKG_DESCRIPTION")).map(to_string)
 }
 
+/// The Rust language edition from the manifest of your package.
+#[doc = requires_msrv!("1.91")]
+#[track_caller]
+pub fn cargo_pkg_edition() -> Option<String> {
+    to_opt(var_or_panic("CARGO_PKG_EDITION")).map(to_string)
```
epage, line 587: Why is this using `to_opt`?  Won;t the edition always have a non-empty value, like `cargo_pkg_name`?

src/doc/src/reference/environment-variables.md
epage: > but not certain about @epage's "deprecated behavior" 

That this would be insufficient for projects using the deprecated [cargo test editions](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-edition-field).

Unsure how important that would be considering it is deprecated, its use was very low, etc.

crates/build-rs-test-lib/build.rs
```diff
@@ -51,6 +51,8 @@ fn smoke_test_inputs() {
     dbg!(cargo_manifest_links());
     dbg!(cargo_pkg_authors());
     dbg!(cargo_pkg_description());
+    // FIXME: enable once released?
+    // dbg!(cargo_pkg_edition());
```
epage, line 55: If only we had `cfg_version`...

How do we plan to track addressing this?

PR #14875, opened 2024-11-30
test: `requires` attribute accepts string literals for cmds

weihanglo:

The new syntax is key-value pair like `requires = "rustfmt"`,
comparing to the previous `requires_<cmd>`.

This enables `cargo_test` to require that the CLI contains non-alphabetic names, such as `llvm-profdata`.

CI passes

### Additional information

This is a request from <https://github.com/rust-lang/cargo/pull/14859#discussion_r1862522117>, but I found it a bit tricky so split off this change first.

### DIFF
--- a/crates/cargo-test-macro/src/lib.rs
+++ b/crates/cargo-test-macro/src/lib.rs
@@ -34,8 +34,8 @@ use std::sync::Once;
 ///   This is useful for tests that use unstable options in `rustc` or `rustdoc`.
 ///   These tests are run in Cargo's CI, but are disabled in rust-lang/rust's CI due to the difficulty of updating both repos simultaneously.
 ///   A `reason` field is required to explain why it is nightly-only.
-/// * `requires_<cmd>` --- This indicates a command that is required to be installed to be run.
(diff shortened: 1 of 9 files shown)


Rustin170506 (approved):
It looks reasonable to me. Thanks! :shipit:

crates/cargo-test-macro/src/lib.rs
```diff
@@ -133,8 +133,18 @@ pub fn cargo_test(attr: TokenStream, item: TokenStream) -> TokenStream {
...
                 );
             }
-            s if s.starts_with("requires_") => {
+            s if s.starts_with("requires=") => {
                 let command = &s[9..];
+                let Ok(literal) = command.parse::<Literal>() else {
+                    panic!("expect a string literal, found: {command}");
+                };
+                let literal = literal.to_string();
+                let Some(command) = literal
```
epage, line 142: should be able to call `.value()`, see https://github.com/clap-rs/clap/blob/b4ea2d4be9f01b6c7945084cbd158d6075cf2408/clap_derive/src/item.rs#L1390

weihanglo: No? Because we are not using `syn` :(

PR #14887, opened 2024-12-03
test(pgo): only run on nightly

weihanglo:

Only run PGO test on nightly,
so it won't run on rust-lang/rust's CI
because `CARGO_TEST_DISABLE_NIGHTLY` is set on that.

This is a temporary fix to unblock Cargo submoule update (assume that rust-lang/rust#133675 won't be fixed this week).

In the future We should seek a solution that

* Disable test on rust-lang/rust's CI
* Maximize test coverage on different channels and platforms on Cargo's CI.

### DIFF
--- a/tests/testsuite/pgo.rs
+++ b/tests/testsuite/pgo.rs
@@ -22,10 +22,15 @@ fn llvm_profdata() -> Option<PathBuf> {
     })
 }
 
-#[cargo_test]
+// Rustc build may be without profiling support.
+// Mark it as nightly so it won't run on rust-lang/rust CI.
+#[cfg_attr(
+    target_os = "linux",
+    cargo_test(nightly, reason = "rust-lang/rust#133675")
+)]
 // macOS may emit different LLVM PGO warnings.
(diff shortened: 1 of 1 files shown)


epage (approved):

ehuss:
Looks like this needs a rustfmt.

weihanglo:
image

So a force push won't dismiss already enabled auto-merge? That is surprising to me.

ehuss:
I think we can change that if you want. Just send a PR to the team repo to set `dismiss-stale-review=true`.

I think the idea with having it disabled is that it allows someone to approve something with the assumption that the author will resolve any remaining issues before it is merged. For example, I could approve something with CI failing and say "this will merge when CI is green", with trust that the author will take reasonable steps to resolve the failures (like running rustfmt). But I'm personally fine with requiring a re-review, it just means a little more hassle in those situations.

weihanglo:
I am fine with the current policy. We mostly trust team members so that might not be needed. Even we accidentally merge something we can always revert it.

PR #14890, opened 2024-12-03
fix(fix): Migrate workspace dependencies

### LINKED ISSUE #14886: 2024 edition migration of `default_features` doesn't work for virtual workspace inherited dependencies
### Problem

A virtual workspace with a workspace dependency that uses an underscore for `default_features` doesn't get migrated to 2024. After updating to 2024, this results in an error.

It does print a warning, though. However, I believe it is intended that the auto-fix should apply to this case to (it supports [workspaces](https://github.com/rust-lang/cargo/blob/ef5edffb89b8c3212ae1061619e5c2309cd8da98/src/cargo/ops/fix.rs#L282-L289), just not a virtual one).

### Steps

Root workspace:

```toml
[workspace]
members = ["a"]

[workspace.dependencies]
empty-library = {version="*", default_features=false}
```

and member "a":

```toml
[package]
name = "a"
version = "0.1.0"
edition = "2021"

[dependencies]
empty-library.workspace = true
```

and run `cargo fix --edition`.

### Possible Solution(s)

I think `cargo fix --edition` should also update these dependencies.

### Notes

_No response_

### Version

```text
cargo 1.85.0-nightly (4c39aaff6 2024-11-25)
release: 1.85.0-nightly
commit-hash: 4c39aaff66862cc0da52fe529aa1990bb8bb9a22
commit-date: 2024-11-25
host: aarch64-unknown-linux-gnu
os: Debian 12.0.0 [64-bit]
```

epage:

Technically, the edition doesn't affect the workspace. We could adjust this as we're inheriting to not be a problem.

But it feels weird to keep this around in newer editions. We could frame this as around the what the package is, including inheritance. The nice thing is that default-features works for all versions that inheritance works so we're not forcing the users hand with multiple editions in a workspace.

Fixes #14886

### DIFF
--- a/src/cargo/ops/fix.rs
+++ b/src/cargo/ops/fix.rs
@@ -252,6 +252,59 @@ fn check_version_control(gctx: &GlobalContext, opts: &FixOptions) -> CargoResult
 }
 
 fn migrate_manifests(ws: &Workspace<'_>, pkgs: &[&Package]) -> CargoResult<()> {
+    // HACK: Duplicate workspace migration logic between virtual manifests and real manifests to
+    // reduce multiple Migrating messages being reported for the same file to the user
+    if matches!(ws.root_maybe(), MaybePackage::Virtual(_)) {
+        // Warning: workspaces do not have an edition so this should only include changes needed by
+        // packages that preserve the behavior of the workspace on all editions
+        let highest_edition = pkgs
+            .iter()
+            .map(|p| p.manifest().edition())
+            .max()
+            .unwrap_or_default();
+        let prepare_for_edition = highest_edition.saturating_next();
+        if highest_edition == prepare_for_edition
+            || (!prepare_for_edition.is_stable() && !ws.gctx().nightly_features_allowed)
+        {
+            //
+        } else {
+            let mut manifest_mut = LocalManifest::try_new(ws.root_manifest())?;
+            let document = &mut manifest_mut.data;
+            let mut fixes = 0;
+
+            if Edition::Edition2024 <= prepare_for_edition {
+                let root = document.as_table_mut();
+
+                if let Some(workspace) = root
+                    .get_mut("workspace")
+                    .and_then(|t| t.as_table_like_mut())
+                {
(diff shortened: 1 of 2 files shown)


src/cargo/ops/fix.rs
```diff
@@ -252,6 +252,59 @@ fn check_version_control(gctx: &GlobalContext, opts: &FixOptions) -> CargoResult
 }
 
 fn migrate_manifests(ws: &Workspace<'_>, pkgs: &[&Package]) -> CargoResult<()> {
+    // HACK: Duplicate workspace migration logic between virtual manifests and real manifests to
+    // reduce duplicate messages being reported to the user
+    if matches!(ws.root_maybe(), MaybePackage::Virtual(_)) {
+        // Warning: workspace's do not have an edition so this should only include changes needed by
```
Rustin170506, line 258:
```suggestion
        // Warning: workspaces do not have an edition so this should only include changes needed by
```

I guess you mean this?

Rustin170506 (commented):
Thanks!

src/cargo/ops/fix.rs
```diff
@@ -252,6 +252,59 @@ fn check_version_control(gctx: &GlobalContext, opts: &FixOptions) -> CargoResult
 }
 
 fn migrate_manifests(ws: &Workspace<'_>, pkgs: &[&Package]) -> CargoResult<()> {
+    // HACK: Duplicate workspace migration logic between virtual manifests and real manifests to
+    // reduce duplicate messages being reported to the user
```
Rustin170506, line 256: I'm not sure what kind of duplication message you're referring to here.

Rustin170506 (approved):
rest LGTM.

Thank you!

src/cargo/ops/fix.rs
epage: I've made the comment more explicit

weihanglo (approved):
The duplicate doesn't look too bad. Thanks!

And thanks for the review from @Rustin170506 as well!

PR #14896, opened 2024-12-05
test(build-std): make mock-std closer to real world

weihanglo:

`test_std_on_unsupported_target` never really succeed to build those targets, due to

* local rustup may not have `{aarch64,x86_64}-unknown-none` installed.
* `core` and `compiler-builtins` mock crate are not `no_std` nor `no_core`
* the dummy `main.rs` uses `println!` and is not `no_std`.

This commit make it compile, if you have those targets installed.

Change this line to `.build_std_arg(&setup, "core")`.

https://github.com/rust-lang/cargo/blob/05f54fdc34310f458033af8a63ce1d699fae8bf6/tests/testsuite/standard_lib.rs?plain=1#L410

And delete these liens: https://github.com/rust-lang/cargo/blob/05f54fdc34310f458033af8a63ce1d699fae8bf6/src/cargo/core/compiler/standard_lib.rs?plain=1#L75-L84

Then the test project should compile.

### DIFF
--- a/tests/testsuite/standard_lib.rs
+++ b/tests/testsuite/standard_lib.rs
@@ -395,12 +395,13 @@ fn test_std_on_unsupported_target() {
 
     let p = project()
         .file(
-            "src/main.rs",
+            "src/lib.rs",
             r#"
-        fn main() {
-            println!("hello");
-        }
-        "#,
+                #![no_std]
+                pub fn foo() {
+                    assert_eq!(u8::MIN, 0);
+                }
+            "#,
         )
         .build();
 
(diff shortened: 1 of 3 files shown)


tests/testsuite/standard_lib.rs
weihanglo, line 1: We still need `{aarch64,x86_64}-unknown-none` targets installed to make `test_std_on_unsupported_target` compile. We need to tweak cargo CI and `cargo-test-macro` to support this use case. That is out of scope of this PR.

epage (approved):

Issue #10623, opened 2022-05-02
cargo <1.60 failed to select a version of a dependency due to either `?` (weak) or `dep:` (namespaced) features


epage:
### Problem

I did a `cargo update` and my CI broke when verifying MSRV. 
```console
error: failed to select a version for the requirement `libgit2-sys = "^0.13.3"`
candidate versions found which didn't match: 0.13.2+1.4.2, 0.13.1+1.4.2, 0.13.0+1.4.1, ...
location searched: crates.io index
required by package `git2 v0.14.3`
    ... which satisfies dependency `git2 = "^0.14"` (locked to 0.14.3) of package `repor v0.1.0 (/home/epage/src/personal/repro)`
```

- cargo 1.60 works
- cargo 1.59 doesn't work

The error being from the resolver and being cargo-version specific is why I am posting here rather than on the git2 repo.

### Steps

```console
$ git clone git@github.com:crate-ci/committed.git
cd committed
cargo +1.59.0 check
cargo +1.60.0 update
cargo +1.59.0 check
```

### Possible Solution(s)

Track parse bad candidates when resolving and report them to the user if they are relevant

### Notes

Old cargo cannot see crates in the registry that use [weak or namespaced features](https://rust-lang.github.io/rfcs/3143-cargo-weak-namespaced-features.html) (or similar new index data).

### Version

```text
cargo 1.60.0 (d1fd9fe 2022-03-01)
release: 1.60.0
commit-hash: d1fd9fe2c40a1a56af9132b5c92ab963ac7ae422
commit-date: 2022-03-01
host: x86_64-unknown-linux-gnu
libgit2: 1.3.0 (sys:0.13.23 vendored)
libcurl: 7.80.0-DEV (sys:0.4.51+curl-7.80.0 vendored ssl:OpenSSL/1.1.1m)
os: Linux Mint 20.3 (una) [64-bit]
```

epage:
Looks like I missed that libgit2-sys uses a weak dependency (really hard to see that `?`), see https://github.com/rust-lang/git2-rs/commit/d8ee105a8f3ce4d5c57cd091b67943aab86b176a

My guess is that this is expected behavior when loading this into old versions of cargo.

epage:
Considering the usability issue of this, re-opening for us providing better errors.

epage:
[...]

PR #14897, opened 2024-12-05
fix(resolver): Report unmatched versions, rather than saying no package

epage:

Instead of saying no package found when there are hidden `Summary`s, we'll instead say why the summary was hidden in the cases of
- Yanked packages
- Schema mismatch
- Offline packages?

The schema mismatch covers part of #10623.  Whats remaining is when we can't parse the `Summary` but can parse a subset (name, version, schema version, optionally rust-version).  That will be handled in a follow up.

This has a couple of risky areas
- Moving the filtering of `IndexSummary` variants from the Index to the Registry Source.  On inspection, there were other code paths but they seemed innocuous to not filter, like asking for a hash of a summary
- Switching `PackageRegistry` to preserve the `IndexSummary` variant for regular sources and overrides
  - I did not switch patches to preserve `IndexSummary` as that was more invasive and the benefits seemed more minor (normally people patch over registry dependencies and not to them)

### DIFF
--- a/src/cargo/core/resolver/errors.rs
+++ b/src/cargo/core/resolver/errors.rs
@@ -304,7 +304,7 @@ pub(super) fn activation_error(
         // Maybe the user mistyped the name? Like `dep-thing` when `Dep_Thing`
         // was meant. So we try asking the registry for a `fuzzy` search for suggestions.
         let candidates = loop {
-            match registry.query_vec(&new_dep, QueryKind::Alternatives) {
+            match registry.query_vec(&new_dep, QueryKind::AlternativeNames) {
                 Poll::Ready(Ok(candidates)) => break candidates,
                 Poll::Ready(Err(e)) => return to_resolve_err(e),
                 Poll::Pending => match registry.block_until_ready() {
--- a/src/cargo/core/resolver/errors.rs
+++ b/src/cargo/core/resolver/errors.rs
@@ -303,7 +303,7 @@ pub(super) fn activation_error(
     } else {
         // Maybe the user mistyped the name? Like `dep-thing` when `Dep_Thing`
         // was meant. So we try asking the registry for a `fuzzy` search for suggestions.
-        let candidates = loop {
+        let name_candidates = loop {
             match registry.query_vec(&new_dep, QueryKind::AlternativeNames) {
                 Poll::Ready(Ok(candidates)) => break candidates,
                 Poll::Ready(Err(e)) => return to_resolve_err(e),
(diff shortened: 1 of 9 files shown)


Eh2406:
I dropped this in a half done state at some point. Thank you for pushing this forward.

src/cargo/core/resolver/errors.rs
```diff
@@ -362,12 +362,23 @@ pub(super) fn activation_error(
                         let _ = writeln!(&mut msg, "  version {} is not cached", summary.version());
                     }
                     IndexSummary::Unsupported(summary, schema_version) => {
-                        let _ = writeln!(
+                        if let Some(rust_version) = summary.rust_version() {
+                            // HACK: technically its unsupported and we shouldn't make assumptions
```
Rustin170506, line 366: What do you mean by `technically its unsupported`?

src/cargo/core/resolver/errors.rs
```diff
@@ -327,7 +341,37 @@ pub(super) fn activation_error(
...
+        if !version_candidates.is_empty() {
+            let _ = writeln!(
+                &mut msg,
+                "no matching versions for `{}` found",
+                dep.package_name()
+            );
+            for candidate in version_candidates {
+                match candidate {
+                    IndexSummary::Candidate(summary) => {
+                        // HACK: If this was a real candidate, we wouldn't hit this case.
```
Rustin170506, line 353: Not sure how this happens. It it possible to add a unit test to cover this?

Rustin170506 (commented):
Just asked some questions that I didn't understand. 

Thank you!

src/cargo/core/resolver/errors.rs
epage: For now, the `Unsupported` branch means that we can parse the entry but the schema version was bumped. How we interpret the fields could have changed.

It'll be even worse in a follow up where I'll add "failed to parse but could get basic data".

src/cargo/core/resolver/errors.rs
epage: I don't know how to hit this. This was originally an assert but I hit it because we switched everything to `Candidate`. That should be fixed but I figured a vague error is better for users than a crash in case I missed something or things change in the future.

src/cargo/core/resolver/errors.rs
Rustin170506: Got it. Thank you!

Rustin170506:
@epage There is a typo in your last commit message. `;` -> `:`

Muscraft (approved):
Thanks for this! The error messages are so much better!

weihanglo (commented):
Did a late review. Brilliant! I love the new `AlternativeVersions` query kind.

PR #14899, opened 2024-12-05
fix(build-std): determine root crates by target spec `std:bool`

weihanglo:

In rust-lang/cargo#14183 Cargo starts bailing out if the `metadata.std`
field in a target spec JSON is set to `false`.
This is problematic because std for some targets are actually buildable
even they've declared as std-unsupported.

This patch removes the hard error, and instead determines the required
root crates by the `metadata.std` field. That is to say, if a target
is explicitly declared as `std: false`, `-Zbuild-std` will build `core`
and `compiler-builtins` only, no `std` will be built.

This patch doesn't change the behavior of `-Zbuild-std` with explicit
crates set. For example `-Zbuild-std=std` will force building `std`.

See Zulip discussion:
https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/help.20debugging.20a.20docs.2Ers.20issue.20with.20a.20new.20cargo.20error

e18b64ff80c7f43758c6330beebfbb0e78da5584 and 125e873dffc4b68b263c5decd88750ec10fd441e might need a closer look.
Cargo relies on how std workspace is organized with these commits. 

Here is an e2e test, if you are willing to test manually.

* Use the current nightly toolchain `rustc 1.85.0-nightly (acabb5248 2024-12-04)`
* `rustup component add rust-src --toolchain nightly`
* `rustup target add aarch64-unknown-none --toolchain nightly`
* Create a `#![no_std]` lib package.
* Run `target/debug/cargo build -Zbuild-std --target aarch64-unknown-none` in that package. Previously it failed with a message `building std is not supported on the following targets`

### DIFF
--- a/tests/testsuite/standard_lib.rs
+++ b/tests/testsuite/standard_lib.rs
@@ -389,8 +389,29 @@ fn check_core() {
         .run();
 }
 
-#[cargo_test(build_std_mock)]
-fn test_std_on_unsupported_target() {
+#[cargo_test(build_std_mock, requires = "rustup")]
+fn build_std_with_no_arg_for_core_only_target() {
+    let has_rustup_aarch64_unknown_none = std::process::Command::new("rustup")
+        .args(["target", "list", "--installed"])
+        .output()
+        .ok()
+        .map(|output| {
+            String::from_utf8(output.stdout)
+                .map(|stdout| stdout.contains("aarch64-unknown-none"))
+                .unwrap_or_default()
+        })
+        .unwrap_or_default();
+    if !has_rustup_aarch64_unknown_none {
+        let msg =
+            "to run this test, run `rustup target add aarch64-unknown-none --toolchain nightly`";
+        if cargo_util::is_ci() {
+            panic!("{msg}");
+        } else {
+            eprintln!("{msg}");
+        }
+        return;
+    }
+
     let setup = setup();
 
     let p = project()
@@ -405,13 +426,14 @@ fn test_std_on_unsupported_target() {
         )
         .build();
 
-    p.cargo("build")
+    p.cargo("build -v")
         .arg("--target=aarch64-unknown-none")
-        .arg("--target=x86_64-unknown-none")
         .build_std(&setup)
         .with_status(101)
         .with_stderr_data(str![[r#"
-[ERROR] building std is not supported on the following targets: [..]
+...
+error[E0463]: can't find crate for `std`
+...
 "#]])
         .run();
 }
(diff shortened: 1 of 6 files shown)


tests/testsuite/standard_lib.rs
```diff
@@ -389,8 +389,29 @@ fn check_core() {
         .run();
 }
 
-#[cargo_test(build_std_mock)]
-fn test_std_on_unsupported_target() {
+#[cargo_test(build_std_mock, requires = "rustup")]
+fn build_std_with_no_arg_for_core_only_target() {
+    let has_rustup_aarch64_unknown_none = std::process::Command::new("rustup")
```
weihanglo, line 394: I hate this. This is added to my list of cargo-test-macro overhaul.

epage: Overhaul `cargo-test-macro` or work on T-testing-devex to make it not needed anymore ;)

epage: We have a standard pattern for alternative targets in [`cargo_test_support::cross_compile`](https://docs.rs/cargo-test-support/latest/cargo_test_support/cross_compile/index.html), should we tie into that?

epage:
> Cargo relies on how std workspace is organized with these commits.

Could you expand on this.  I didn't quite get how some of those parts fit together and tie back to this.  Is this the sysroot stuff?  How much of a problem is that?

weihanglo:
> > Cargo relies on how std workspace is organized with these commits.
> 
> Could you expand on this. I didn't quite get how some of those parts fit together and tie back to this. Is this the sysroot stuff? How much of a problem is that?

It will be a problem if `sysroot` is removed from std workspace, or stop depending on `std`, or a Cargo features changes and one dependency is not activated anymore. The result of std dependency resolution will fail or miss important dependencies due to these reasons. This change makes it a bit more fragile because Cargo got broken whenever `sysroot` is broken, which before this we might have a little chance to build a half-broken std.

tests/testsuite/standard_lib.rs
weihanglo: That is actually a good idea, though I am not sure. The probe is based on building a simple cross-compile project. I don't want to disable all cross-compile test locally just because `aarch64-unknow-none` is missing.

epage: We could have a `cargo_test_support::build_std` with its own `disabled`.  More so I was wanting to highlight that pattern for consideration.

weihanglo: Yeah. I'll keep this in mind. If we get any new tests with the same pattern we can extract the common part then.

epage (approved):

PR #14900, opened 2024-12-06
fix(cargo-rustc): stabilize higher precedence trailing flags

### LINKED ISSUE #14346: Precedence of `cargo rustc -- <flags>`  is too low to override flags set by Cargo

weihanglo:

This was always enabled on nightly since 1.83-nightly (2024-09).
We have no feedback since then, so assume it is a low-impact change.

This stabilization is targeted at 1.85 (2025-02-20)

Fixes #14346

Let's do an FCP.


weihanglo:
@rfcbot fcp merge

epage:
Treating `cargo rustc` extra args the same as rustflags is the right move imo

epage (approved):

PR #14901, opened 2024-12-06
fix(build-rs)!: Remove meaningless 'cargo_cfg_debug_assertions'

epage:

The documentation that was added was pulled straight from a comment in `custom_build.rs`:

> This cfg is always true and misleading, so avoid setting it.
> That is because Cargo queries rustc without any profile settings.

It is therefore misleading to expose this in the API

### DIFF
--- a/crates/build-rs/src/input.rs
+++ b/crates/build-rs/src/input.rs
@@ -109,6 +109,11 @@ mod cfg {
     }
 
     /// If we are compiling with debug assertions enabled.
+    ///
+    /// Build scripts are not passed this cfg because
+    /// this cfg is always true and misleading.
+    /// That is because Cargo queries rustc without any profile settings.
+    #[cfg(any())]
     #[track_caller]
     pub fn cargo_cfg_debug_assertions() -> bool {
         is_present("CARGO_CFG_DEBUG_ASSERTIONS")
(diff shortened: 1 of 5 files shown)


crates/build-rs/src/input.rs
```diff
@@ -109,6 +109,10 @@ mod cfg {
     }
 
     /// If we are compiling with debug assertions enabled.
+    ///
+    /// This cfg is always true and misleading, so avoid setting it.
+    /// That is because Cargo queries rustc without any profile settings.
```
ehuss, line 114:
```suggestion
    /// This cfg is always false and misleading, so avoid setting it.
    /// That is because Cargo explicitly filters out this setting since it queries rustc without any profile settings, and historically led to confusion.
```

Correct? The cfg is [filtered](https://github.com/rust-lang/cargo/blob/a4a3651d662c44e23827c8b42107844873f108df/src/cargo/core/compiler/custom_build.rs#L351-L355) so it is always unset.

epage: Yeah, copying it directly can be misleading.  I'm wanting to keep the comment in line with where its filtered, so I tweaked it in a different way.

ehuss (approved):
Oh, I see. I guess there are two ways to look at this, from the angle of the build script (the env is never set and thus false) and the angle from cargo (rustc always tells us it is always true).

Thanks!

PR #14902, opened 2024-12-06
feat(build-script): Pass CARGO_CFG_FEATURE 

### LINKED ISSUE #3702: feature to env var mapping is not unique
Build scripts receive information about activated features as environment variables `CARGO_FEATURE_$name`, where `$name` is the feature name uppercased and with dashes converted to underscores.

However, feature names are allowed to contain underscores, so this transformation can create collisions. Example:

- Cargo.toml
  ```toml
  [package]
  name = "tester"
  version = "0.1.0"
  
  build = "build.rs"
  
  [features]
  foo-bar = []
  foo_bar = []
  ```
- build.rs
  ```rust
  use std::env;
  
  fn main() {
      for (k, v) in env::vars() {
          if k.starts_with("CARGO_FEATURE") {
              println!("{} => {}", k, v);
          }
      }
      panic!();
  }
  ```
- output
  ```
  --- stdout
  CARGO_FEATURE_FOO_BAR => 1
  
  --- stderr
  thread 'main' panicked at 'explicit panic', build.rs:9
  note: Run with `RUST_BACKTRACE=1` for a backtrace.
  ```

epage:

This may look redundant with `CARGO_FEATURE_<CASE_CONVERTED_NAME>=1`
except that doesn't provide a lossless way of getting the names, e.g. for
forwarding to child builds like tests that need to build examples. which clap does to [test examples](https://github.com/clap-rs/clap/blob/master/tests/examples.rs). Maintaining that manually is easy to mess up.

This also makes things more consistent as users
conditionalize on features through `cfg` and this even fits with what
the `CARGO_CFG_` docs say:

> For each configuration option of the package being built, this
> environment variable will contain the value of the configuration, where
> <cfg> is the name of the configuration uppercased and having -
> translated to _. Boolean configurations are present if they are set, and
> not present otherwise. Configurations with multiple values are joined to
> a single variable with the values delimited by , (comma). This includes values
> built-in to the compiler (which can be seen with rustc --print=cfg) and
> values set by build scripts and extra flags passed to rustc (such as
> those defined in RUSTFLAGS). Some examples of what these variables are: 

Fixes #3702

### DIFF
--- a/crates/build-rs-test-lib/build.rs
+++ b/crates/build-rs-test-lib/build.rs
@@ -9,6 +9,7 @@ fn main() {
 fn smoke_test_inputs() {
     use build_rs::input::*;
     dbg!(cargo());
+    dbg!(cargo_cfg_feature());
     dbg!(cargo_cfg("careful"));
     #[cfg(feature = "unstable")]
     dbg!(cargo_cfg_fmt_debug());
--- a/crates/build-rs/src/input.rs
+++ b/crates/build-rs/src/input.rs
@@ -102,6 +102,13 @@ mod cfg {
     // those disabled with #[cfg(any())] don't seem meaningfully useful
     // but we list all cfg that are default known to check-cfg
 
+    /// Each activated feature of the package being built
+    #[doc = requires_msrv!("1.85")]
+    #[track_caller]
+    pub fn cargo_cfg_feature() -> Vec<String> {
+        to_strings(var_or_panic(&cargo_cfg_var("target_feature")), ',')
+    }
+
     #[cfg(any())]
     #[track_caller]
     pub fn cargo_cfg_clippy() -> bool {
--- a/src/doc/src/reference/environment-variables.md
+++ b/src/doc/src/reference/environment-variables.md
@@ -342,6 +342,7 @@ let out_dir = env::var("OUT_DIR").unwrap();
   values built-in to the compiler (which can be seen with `rustc --print=cfg`)
   and values set by build scripts and extra flags passed to `rustc` (such as
   those defined in `RUSTFLAGS`). Some examples of what these variables are:
+    * `CARGO_CFG_FEATURE` --- Each activated feature of the package being built.
     * `CARGO_CFG_UNIX` --- Set on [unix-like platforms].
     * `CARGO_CFG_WINDOWS` --- Set on [windows-like platforms].
     * `CARGO_CFG_TARGET_FAMILY=unix,wasm` --- The [target family].
@@ -356,8 +357,7 @@ let out_dir = env::var("OUT_DIR").unwrap();
   > Note that different [target triples][Target Triple] have different sets of `cfg` values,
   > hence variables present in one target triple might not be available in the other.
   >
-  > Some cfg values like `debug_assertions`, `test`, and Cargo features like
-  > `feature="foo"` are not available.
+  > Some cfg values like `debug_assertions` and `test` are not available.
 * `OUT_DIR` --- the folder in which all output and intermediate artifacts should
   be placed. This folder is inside the build directory for the package being built,
   and it is unique for the package in question.
--- a/crates/build-rs/src/lib.rs
+++ b/crates/build-rs/src/lib.rs
@@ -40,6 +40,20 @@ MSRV: Respected as of "#,
     };
 }
 
+macro_rules! requires_msrv {
+    ($ver:literal) => {
+        concat!(
+            r#"<div class="warning">
+
+MSRV: Requires "#,
+            $ver,
+            r#".
+
+</div>"#
+        )
+    };
+}
+
 mod ident;
 
 pub mod input;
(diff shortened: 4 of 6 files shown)


crates/build-rs/src/input.rs
```diff
@@ -102,6 +102,13 @@ mod cfg {
     // those disabled with #[cfg(any())] don't seem meaningfully useful
     // but we list all cfg that are default known to check-cfg
 
+    /// Each activated feature of the package being built
+    #[doc = requires_msrv!("1.85")]
```
epage, line 106: This will need to be updated based on when this gets merged

epage:
#14901 will perform the version bump for `build-rs`

src/doc/src/reference/environment-variables.md
```diff
@@ -342,6 +342,7 @@ let out_dir = env::var("OUT_DIR").unwrap();
   values built-in to the compiler (which can be seen with `rustc --print=cfg`)
   and values set by build scripts and extra flags passed to `rustc` (such as
   those defined in `RUSTFLAGS`). Some examples of what these variables are:
+    * `CARGO_CFG_FEATURE` --- Each activated feature of the package being built.
```
weihanglo, line 345: > Some cfg values like debug_assertions, test, and Cargo features like feature="foo" are not available.

We did this a while back in <https://github.com/rust-lang/cargo/pull/14676>. And Cfg `test` and `debug_assertions` are not there anyway, so `CARGO_CFG_*` still doesn't cover everything. 

May need to remove the mention of features from that part of the doc as well.

crates/build-rs-test-lib/build.rs
weihanglo, line 1: I am not sure about it. To me, it seems like a misfeature supporting both `foO_bar` and `foo-bar` as feature keys. This patch does resolve the issue though.

epage: My use case where this would help isn't with ambiguous feature names but with being able to enumerate all features and turn them back into feature names.

e.g. I have something like this in several projects
```rust
#[test]
#[cfg(feature = "help")]
#[cfg(feature = "error-context")]
#[cfg(feature = "usage")]
fn example_tests() {
    let t = trycmd::TestCases::new();
    let features = [
        // Default
        #[cfg(feature = "std")]
        "std",
        #[cfg(feature = "color")]
        "color",
        #[cfg(feature = "help")]
        "help",
        #[cfg(feature = "usage")]
        "usage",
        #[cfg(feature = "error-context")]
        "error-context",
        #[cfg(feature = "suggestions")]
        "suggestions",
        // Optional
        #[cfg(feature = "derive")]
        "derive",
        #[cfg(feature = "cargo")]
        "cargo",
        #[cfg(feature = "wrap_help")]
        "wrap_help",
        #[cfg(feature = "env")]
        "env",
        #[cfg(feature = "unicode")]
        "unicode",
        #[cfg(feature = "string")]
        "string",
        // In-work
        //#[cfg(feature = "unstable-v5")]  // Currently has failures
        //"unstable-v5",
    ]
    .join(" ");
    t.register_bins(trycmd::cargo::compile_examples(["--features", &features]).unwrap());
    t.case("examples/**/*.md");
}
```
https://github.com/clap-rs/clap/blob/master/tests/examples.rs

If I get that wrong, then I get longer build times.

I also want to put forward an ACP for reading `cfg` values so I wouldn't need a build script.

src/doc/src/reference/environment-variables.md
epage: The fun of reading stable docs.

Ah, it was also at the end so I didn't notice it when reading the source.

crates/build-rs-test-lib/build.rs
weihanglo: How is it related to build script, do you plan to use `cargo::rustc-env` to pass them to example?

Regardless, for posterity I think use cases worth listing in the PR description.

epage: Thats the easiest way.

It was covered in the issue.  Someone else reported wanting to do the same forwarding.  I can include i if i helps.

epage:
CC @CAD97

crates/build-rs/src/input.rs
CAD97: It'd probably be beneficial to set up a `CURRENT_RUST_VERSION` placeholder like is used for stabilization in rustc

CAD97 (approved):

crates/build-rs/src/input.rs
epage: Yes, we only recently started tracking these situations more regularly in 1.81 and haven't looked at updating our processes yet.  So far, this has been working out ok.

epage:
As for the syntax for multiple values, the [docs](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts) do say

> Configurations with multiple values are  joined to a single variable with the values delimited by `,`.

weihanglo (approved):
Looks good. Thank you!

I cannot think of any reason this may break existing code, so will merge it now. If there is anything bad happening, we can revert then.

weihanglo:
```
---- ssh::known_host_works stdout ----
SSH_AGENT_PID=79472
thread 'ssh::known_host_works' panicked at tests/testsuite/ssh.rs:187:10:
called `Option::unwrap()` on a `None` value
```

Spurious network error?

Issue #13919, opened 2024-05-16
Windows `\\?\` verbatim paths break idiomatic use of OUT_DIR and `include!`


dtolnay:
### Problem

When Cargo is invoked on Windows in either of the following 2 scenarios, and possibly others:

- a current directory that begins with `\\?\`
- `--manifest-path` that begins with `\\?\`

then the following widespread idiom for including build.rs-generated code does not work.

```rust
include!(concat!(env!("OUT_DIR"), "/repro.rs"));
```

Because `/` is not a directory separator in verbatim paths, this fails to compile with an error like this:

```console
error: couldn't read \\?\D:\a\repro\repro\target\debug\build\repro-59592f29f689633e\out/repro.rs: The filename, directory name, or volume label syntax is incorrect. (os error 123)
 --> src\lib.rs:3:1
  |
3 | include!(concat!(env!("OUT_DIR"), "/repro.rs"));
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the macro `include` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Normally in Windows paths `/` works just as well as `\`, but from [Maximum Path Length Limitation](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation): _File I/O functions in the Windows API convert "/" to "\\" as part of converting the name to an NT-style name, **except when using the "\\\\?\\" prefix** as detailed in the following sections_.

### Steps

The problem can be reproduced in GitHub Actions with the following source code.

```toml
# Cargo.toml

[package]
name = "repro"
version = "0.0.0"
```

```rust
// build.rs

fn main() {
    let out_dir = std::env::var_os("OUT_DIR").unwrap();
    let repro = std::path::Path::new(&out_dir).join("repro.rs");
    std::fs::write(repro, "").unwrap();
}
```

```rust
// src/lib.rs

include!(concat!(env!("OUT_DIR"), "/repro.rs"));
```

```yaml
# .github/workflows/ci.yml

name: CI
on: [push]
jobs:
  repro:
    runs-on: windows-latest
    steps:
      - uses: actions/checkout@v4
      - uses: dtolnay/rust-toolchain@nightly

      - run: pwd
      - run: echo ${{github.workspace}}
      - run: echo \\?\${{github.workspace}}

      - run: cargo check --verbose && cargo clean

[...]

PR #14908, opened 2024-12-09
refactor: use Path::push to construct remap-path-prefix

weihanglo:

It creates paths with correct separators for different systems.

Try out `-Ztrim-paths` on Windows. Run `cargo build --verbose` and see if slashes in `--remap-path-prefix` use the system's path separators.

```toml
cargo-features = ["trim-paths"]

[package]
name = "foo"

[profile.release]trim-paths = true
```

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1310,10 +1310,16 @@ fn trim_paths_args(
 /// This remap logic aligns with rustc:
 /// <https://github.com/rust-lang/rust/blob/c2ef3516/src/bootstrap/src/lib.rs#L1113-L1116>
 fn sysroot_remap(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> OsString {
-    let sysroot = &build_runner.bcx.target_data.info(unit.kind).sysroot;
     let mut remap = OsString::from("--remap-path-prefix=");
-    remap.push(sysroot);
-    remap.push("/lib/rustlib/src/rust"); // See also `detect_sysroot_src_path()`.
+    remap.push({
+        // See also `detect_sysroot_src_path()`.
+        let mut sysroot = build_runner.bcx.target_data.info(unit.kind).sysroot.clone();
+        sysroot.push("lib");
+        sysroot.push("rustlib");
+        sysroot.push("src");
+        sysroot.push("rust");
+        sysroot
+    });
     remap.push("=");
     remap.push("/rustc/");
     if let Some(commit_hash) = build_runner.bcx.rustc().commit_hash.as_ref() {


src/cargo/core/compiler/mod.rs
```diff
@@ -1313,7 +1313,11 @@ fn sysroot_remap(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> OsString {
     let sysroot = &build_runner.bcx.target_data.info(unit.kind).sysroot;
     let mut remap = OsString::from("--remap-path-prefix=");
     remap.push(sysroot);
-    remap.push("/lib/rustlib/src/rust"); // See also `detect_sysroot_src_path()`.
+    // See also `detect_sysroot_src_path()`
+    remap.push("lib");
```
weihanglo, line 1317: Oops I forgot it is a OsString

src/cargo/core/compiler/mod.rs
```diff
@@ -1313,7 +1313,11 @@ fn sysroot_remap(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> OsString {
     let sysroot = &build_runner.bcx.target_data.info(unit.kind).sysroot;
     let mut remap = OsString::from("--remap-path-prefix=");
     remap.push(sysroot);
-    remap.push("/lib/rustlib/src/rust"); // See also `detect_sysroot_src_path()`.
```
Muscraft, line 1316: [`detect_sysroot_src_path`](https://github.com/rust-lang/cargo/blob/master/src/cargo/core/compiler/standard_lib.rs#L200-L206) achieves the same thing with `.join` on `sysroot` so I think this should be:
```rust
    remap.push(
        // See also `detect_sysroot_src_path()`.
        sysroot
            .join("lib")
            .join("rustlib")
            .join("src")
            .join("rust")
            .as_os_str(),
    );
```

src/cargo/core/compiler/mod.rs
```diff
@@ -1310,9 +1310,16 @@ fn trim_paths_args(
...
+    remap.push({
+        // See also `detect_sysroot_src_path()`.
+        let mut sysroot = build_runner.bcx.target_data.info(unit.kind).sysroot.clone();
+        sysroot.push("lib");
+        sysroot.push("rustlib");
+        sysroot.push("src");
+        sysroot.push("rust");
+        sysroot
+    });
     remap.push("/lib/rustlib/src/rust"); // See also `detect_sysroot_src_path()`.
```
Muscraft, line 1323:
```suggestion
```

weihanglo: Oops. Thanks!

ehuss:
Can you say more about what this is fixing? Is path remapping just not working on Windows for sysroot paths? Do we not have a test that would catch that?

weihanglo:
> Can you say more about what this is fixing? Is path remapping just not working on Windows for sysroot paths? Do we not have a test that would catch that?

We don't have any test for windows now, though path remapping works on Windows. 
I just got paranoid about it is not using the correct `\` separator and am afraid of <https://github.com/rust-lang/cargo/issues/13919> happening if `rustc --print sysroot` starts printing one.

ehuss (approved):
Ok, I'll go ahead and merge. I just wanted to comment on a few things:

* It would be good to eventually have some Windows tests.
* Without trim-paths, diagnostic paths look like `C:\Users\eric\.rustup\toolchains\nightly-aarch64-pc-windows-msvc\lib/rustlib/src/rust\library\core\src\hash\mod.rs`. With trim-paths (with or without this PR), they get remapped to ` /rustc/4d669fb34e7db6f3825d01e4c59b7996f0531431\library\core\src\hash\mod.rs`. Is rustc maybe doing some backslash/forwardslash translation?
* I don't see `rustlib` absolute paths in the `.exe` with trim-paths turned off. I'm not sure if there are scenarios where that might happen?
* I do see absolute paths in the `.pdb` file. With or without this PR, they are still there, so I assume Windows debug stripping is either not implemented, or broken, or fundamentally not feasible. They look to always use backslashes. For example, `C:\Users\eric\.rustup\toolchains\nightly-aarch64-pc-windows-msvc\lib\rustlib\aarch64-pc-windows-msvc\lib\libstd-42ec610576778e0a.rlib`.

weihanglo:
> I do see absolute paths in the .pdb file. With or without this PR, they are still there

Yes I noticed it yesterday as well and posted a comment: <https://github.com/rust-lang/rust/issues/111540#issuecomment-2526523455>.

The entire stuff on Windows is still a new thing to me. Sorry off the top of my head I don't have any useful answers for you now :(

I'll add them to unresolved part respectively

PR #14910, opened 2024-12-09
feat(build-rs): Add the 'error' directive

epage:

My previous audit mostly focused on env variables and missed that `cargo::error` hadn't been added yet.

This also includes other clean up along the way to #14902 and fixing of some env variable handling that I'm working towards.


crates/build-rs/src/output.rs
```diff
@@ -6,8 +6,11 @@
 //!
 //! Reference: <https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script>
 
+use std::ffi::OsStr;
+use std::path::Path;
```
weihanglo, line 10: Do we want to automate this?

crates/build-rs/src/output.rs
```diff
@@ -309,13 +306,11 @@ pub fn rustc_check_cfgs(keys: &[&str]) {
         }
     }
 
-    if allow_use::check_cfg() {
-        let mut directive = keys[0].to_string();
-        for key in &keys[1..] {
-            write!(directive, ", {key}").expect("writing to string should be infallible");
-        }
-        emit("rustc-check-cfg", format_args!("cfg({directive})"));
+    let mut directive = keys[0].to_string();
```
weihanglo, line 312: This is pretty much unrelated to the error directive and may have other impacts (which is extremely rare I also believe). While I am fine with each change in this PR, should they be broken down to multiple PRs?

crates/build-rs/src/output.rs
epage: The rustfmt feature is unstable :(

crates/build-rs/src/output.rs
epage: The MSRV for this crate is 1.81.  `check-cfg` was added in 1.80.  This would only be a concern if/when we drop the MSRV further.

weihanglo (approved):

epage:
CC @CAD97

crates/build-rs/src/output.rs
```diff
@@ -411,6 +403,26 @@ pub fn warning(message: &str) {
...
+/// returning a `Result`. It may be better to return a `Result`, and allow the caller to decide if the
+/// error is fatal or not. The caller can then decide whether or not to display the `Err` variant
+/// using [`error`].
+///
+/// </div>
+#[doc = respected_msrv!("1.84")]
+#[track_caller]
+pub fn error(message: &str) {
+    if message.contains('\n') {
+        panic!("cannot emit warning: message contains newline");
```
CAD97, line 421:
```suggestion
        panic!("cannot emit error: message contains newline");
```

CAD97 (approved):

crates/build-rs/src/output.rs
weihanglo: 😅
Nice catch!

epage: Handled in #14913

PR #14911, opened 2024-12-09
fix(build-rs): Implicitly report rerun-if-env-changed for input

epage:

As we abstract away the env variables, users can't do a good job of
reporting these, so we'll do it ourselves.

This isn't the shortest path to this solution because I had originally designed this like more like [`build_env`](https://docs.rs/build-env/latest/build_env/struct.BuildEnv.html) but in a way to allow users to choose what policies they want (rather than baking in their HOST/TARGET scheme).
I decided to not support a custom `env` yet to make this more incremental and to see what feedback we get but I left the implementation in a way that would make it easy to support it.

I made `is_present` part of `Env` because I was unsure if callers of the API would want to handle that differently for `rerun-if-env-changed` or other parts.

### Additional information

CC @CAD97

### DIFF
--- a/crates/build-rs/src/input.rs
+++ b/crates/build-rs/src/input.rs
@@ -9,6 +9,10 @@
 use std::path::PathBuf;
 
 use crate::ident::{is_ascii_ident, is_crate_name, is_feature_name};
+use crate::output::rerun_if_env_changed;
+
+/// [`ProcessEnv`] wrapper that implicit calls [`rerun_if_env_changed`]
+const ENV: RerunIfEnvChanged<ProcessEnv> = RerunIfEnvChanged::new();
 
 /// Abstraction over environment variables
 trait Env {
@@ -41,6 +45,28 @@ impl Env for ProcessEnv {
     }
 }
 
+/// [`Env`] wrapper that implicitly calls [`rerun_if_env_changed`]
+struct RerunIfEnvChanged<E: Env>(E);
+
+impl RerunIfEnvChanged<ProcessEnv> {
+    const fn new() -> Self {
+        Self(ProcessEnv)
+    }
+}
+
+impl<E: Env> Env for RerunIfEnvChanged<E> {
+    #[track_caller]
+    fn get(&self, key: &str) -> Option<std::ffi::OsString> {
+        rerun_if_env_changed(key);
+        self.0.get(key)
+    }
+
+    #[track_caller]
+    fn is_present(&self, key: &str) -> bool {
+        self.get(key).is_some()
+    }
+}
+
 /// Path to the `cargo` binary performing the build.
 #[track_caller]
 pub fn cargo() -> PathBuf {
@@ -60,8 +86,7 @@ pub fn cargo_manifest_dir() -> PathBuf {
 /// The path to the manifest of your package.
 #[track_caller]
 pub fn cargo_manifest_path() -> PathBuf {
-    ProcessEnv
-        .get("CARGO_MANIFEST_PATH")
+    ENV.get("CARGO_MANIFEST_PATH")
         .map(to_path)
         .unwrap_or_else(|| {
             let mut path = cargo_manifest_dir();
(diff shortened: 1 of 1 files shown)


crates/build-rs/src/input.rs
```diff
@@ -6,10 +6,65 @@
...
+    fn get(&self, key: &str) -> Option<std::ffi::OsString> {
+        std::env::var_os(key)
+    }
+
+    fn is_present(&self, key: &str) -> bool {
+        self.get(key).is_some()
+    }
+}
+
+/// [`Env`] wrapper that implicit calls [`rerun_if_env_changed`]
```
CAD97, line 47:
```suggestion
/// [`Env`] wrapper that implicitly calls [`rerun_if_env_changed`]
```

crates/build-rs/src/input.rs
```diff
@@ -554,13 +609,10 @@ pub fn cargo_pkg_readme() -> Option<PathBuf> {
     to_opt(var_or_panic("CARGO_PKG_README")).map(to_path)
 }
 
-fn is_present(key: &str) -> bool {
-    var_os(key).is_some()
-}
-
 #[track_caller]
 fn var_or_panic(key: &str) -> std::ffi::OsString {
```
CAD97, line 613: Changing this to e.g. `ENV.require(key)` may be a good idea to keep env access more uniform, and paves the way towards maybe supporting a different protocol in the future.

CAD97 (commented):
:+1: on direction from me. 

`cargo::rerun-on-env-changed` is purely additive and doesn't disable a default watch set like `cargo::rerun-on-changed` does, so emitting it automatically seems like a good idea. 

*Although*, I'm not sure it's actually needed for the `CARGO_` environment? It depends on how aggressive Cargo is at caching the buildscript and if any of this environment can change without necessarily requiring a recompile.

epage:
> Although, I'm not sure it's actually needed for the CARGO_ environment? It depends on how aggressive Cargo is at caching the buildscript and if any of this environment can change without necessarily requiring a recompile.

The only cases I saw us specially handle are
https://github.com/rust-lang/cargo/blob/2f74b541301e676c884cdae1a59d530858b6d657/src/cargo/core/compiler/fingerprint/mod.rs#L852-L865
and
https://github.com/rust-lang/cargo/blob/2f74b541301e676c884cdae1a59d530858b6d657/src/cargo/core/compiler/fingerprint/dep_info.rs#L306-L338

Which makes sense.  We can set a lot and we likely don't want to re-run if all of it has changed.  Some we'll rebuild for anyways because some of the fingerprint has changed but I figure its better to be over aggressive than under aggressive or having to maintain a perfectly curated set of ENVs

crates/build-rs/src/input.rs
epage: I intentionally kept it off.

I see two ways of going
- Minimum interface for what is needed
- Every input being an inherent function on the trait

I went with the former.  I only included `is_present` on the trait because I was concerned about what the right policy is for re-run if changed in that case.  The direction I went with doesn't specialize it but I figured I'd still leave that door open.  I'm not seeing a reason to specialize this atm and would rather wait for use cases rather than speculatively adding functionality to the trait.

Since we can add this as an inherent function, we can likely do so without a breaking change (yes, there are caveats).  Even then, its not a big deal to make a breaking change.

CAD97: Especially since the trait/type aren't `pub` (yet?) and are just internal helpers.

Muscraft (approved):
Thanks! This seems like a great incremental step overall.

weihanglo (approved):
> Which makes sense. We can set a lot and we likely don't want to re-run if all of it has changed. Some we'll rebuild for anyways because some of the fingerprint has changed but I figure its better to be over aggressive than under aggressive or having to maintain a perfectly curated set of ENVs

Agree with this. While we don't recommend user to hand-write `rerun-if-env-changed` with cargo internal envs, as a build script helper crate it should be fine as an implementation detail.

PR #14912, opened 2024-12-09
chore: update auto-label to include build-rs crate

weihanglo:

We have imported [build-rs](https://crates.io/crates/build-rs)
so update auto-label trigger files.

Should we create a dedicated label for build-rs?
I personally don't think so,
and I think we should have a test verifying directives in build-rs are always aligned with what cargo has.


epage (approved):

epage:
I think it makes sense to keep it under one label, at least for now.

PR #14915, opened 2024-12-10
a faster hash for ActivationsKey

Eh2406:
This is a perf improvement I suggested https://github.com/rust-lang/cargo/pull/14665#issuecomment-2412077472

I mostly want this landed to make it easier to compare the cost and benefits of more complicated changes. As this is the thing to compare against.

### DIFF
--- a/src/cargo/core/resolver/types.rs
+++ b/src/cargo/core/resolver/types.rs
@@ -168,11 +168,11 @@ impl ResolveOpts {
 /// semver compatible version of each crate.
 /// Find the activated version of a crate based on the name, source, and semver compatibility.
 #[derive(Clone, PartialEq, Eq, Debug, Ord, PartialOrd, Hash)]
-pub struct ActivationsKey(InternedString, SourceId, SemverCompatibility);
+pub struct ActivationsKey(InternedString, SemverCompatibility, SourceId);
 
 impl ActivationsKey {
-    pub fn new(name: InternedString, source_id: SourceId, ver: SemverCompatibility) -> ActivationsKey {
-        ActivationsKey(name, source_id, ver)
+    pub fn new(name: InternedString, ver: SemverCompatibility, source_id: SourceId) -> ActivationsKey {
+        ActivationsKey(name, ver, source_id)
     }
 }
 
@@ -200,7 +200,7 @@ impl From<&semver::Version> for SemverCompatibility {
 
 impl PackageId {
     pub fn as_activations_key(self) -> ActivationsKey {
-        ActivationsKey(self.name(), self.source_id(), self.version().into())
+        ActivationsKey(self.name(), self.version().into(), self.source_id())
     }
 }
 
--- a/src/cargo/core/resolver/types.rs
+++ b/src/cargo/core/resolver/types.rs
@@ -167,12 +167,20 @@ impl ResolveOpts {
 /// A key that when stord in a hash map ensures that there is only one
 /// semver compatible version of each crate.
 /// Find the activated version of a crate based on the name, source, and semver compatibility.
-#[derive(Clone, PartialEq, Eq, Debug, Ord, PartialOrd, Hash)]
+#[derive(Clone, PartialEq, Eq, Debug, Ord, PartialOrd)]
 pub struct ActivationsKey(InternedString, SemverCompatibility, SourceId);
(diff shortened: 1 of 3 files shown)


Eh2406:
Force pushed to split up into separate commits for easier review.

src/cargo/util/interning.rs
```diff
@@ -90,6 +90,12 @@ impl InternedString {
     pub fn as_str(&self) -> &'static str {
         self.inner
     }
+
+    /// A faster implementation of hash that is completely compatible with HashMap,
+    /// but does not have a stable value between runs of the program.
+    pub fn fast_hash<H: std::hash::Hasher>(&self, state: &mut H) {
```
epage, line 96: Not quite compatible with `HashMap` because a lookup with a `""` will always fail because this diverges in the hash from `&str`

src/cargo/core/resolver/types.rs
```diff
@@ -176,6 +176,14 @@ impl ActivationsKey {
     }
 }
 
+impl std::hash::Hash for ActivationsKey {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.0.fast_hash(state);
+        self.1.hash(state);
+        // self.2.hash(state); // Packages that only differ by SourceId are rare enough to not be worth hashing
```
epage, line 187: So we have 3 improvments
- Re-order
- Pointer hash on the name
- Skipping the source hash

Do we know how much of an improvement each one of these is?  I'm mostly wondering about skipping the source hash as that feels icky for some reason.

Not going to give this feeling too much weight though.  I'll mostly defer to you on this part.

Eh2406:
Running `cargo r -r -- -t 100 -m cargo --with-solana --filter solana` on https://github.com/Eh2406/pubgrub-crates-benchmark on the `index commit hash: 15350ef77c6486d35d20bb1ca81f2bc051500fb6`

c2b0d50451b7f93013d90851d4daa1c219b08689
```
           Cargo CPU time: 22966.32s == 382.77min
                Wall time:   549.50s ==   9.16min
```
7a67e88baf0f73196973cadf887bf9a31c13e95f
```
           Cargo CPU time: 21391.77s == 356.53min
                Wall time:   520.00s ==   8.67min
```
That is a 7% improvement, which is nothing to sneeze at given the size of the change.

src/cargo/core/resolver/types.rs
Eh2406: > Do we know how much of an improvement each one of these is?

I just collected overall performance numbers, I will try and get you some intermediate numbers.

>  I'm mostly wondering about skipping the source hash as that feels icky for some reason.

Can you articulate why? HashMaps can  devolve to `O(n)`  if  skip fields are the only thing distinguishing `n`  items, but we specifically commented to record why we didn't think that would happen. Skipping a field when hashing is a pretty standard performance trick.

src/cargo/util/interning.rs
Eh2406: I will try and update comment. It is not safe to use this hash for intern string in general while it has the deref to `str`. But, I think it's safe in this case. I need to articulate when it is safe to use it.

Eh2406:
> * Re-order

f2b499847b3cbb702628bf74f48deac7d5c8186c
```
           Cargo CPU time: 21846.75s == 364.11min
                Wall time:   515.96s ==   8.60min
```
So 5% of the improvement! (It would be reasonable to land just this part.)

> *   Re-order
> *   Pointer hash on the name

Doing thees 2 gets us:
```
           Cargo CPU time: 22780.97s == 379.68min
                Wall time:   540.51s ==   9.01min
```
That is slower... let me re-run that.
```
           Cargo CPU time: 21908.46s == 365.14min
                Wall time:   526.55s ==   8.78min
```
So lets not do that.


> *    Re-order
> *    Skipping the source hash

For completeness lets try this combination:
```
           Cargo CPU time: 20367.65s == 339.46min
                Wall time:   480.66s ==   8.01min
```
Oddly that looks like a winner. Let me try it again, just in case.
```
           Cargo CPU time: 21254.67s == 354.24min
                Wall time:   499.23s ==   8.32min
```
So push incoming.

epage (approved):
Thanks!

Issue #14795, opened 2024-11-08
Path to registry index is not cross-platform stable (i.e. hash part of `registry/index/index.crates.io-<hash>`)


uncomfyhalomacro:
Hello. I have an issue with `cargo-fetch` behaving differently than `cargo-vendor`. In a part of the help section of `cargo fetch`, it says for the flag `--target`
> Fetch for the given architecture. The default is all architectures. The
general format of the triple is `<arch><sub>-<vendor>-<sys>-<abi>`. 
Run `rustc --print target-list` for a list of supported targets. This flag may
be specified multiple times.

However, I believe this is untrue since in some architectures, they cannot find their dependencies. I haven't tried the `--target` flag yet as a workaround but given that some target architectures cannot find their dependencies is weird to me, since it says that the default is to use "all" target architectures.

Here is a comparison between `cargo vendor` vs `cargo fetch`.
- **Vendor** https://build.opensuse.org/package/show/home:uncomfyhalomacro:branches:editors:vendor/kak-lsp
- **Fetch** https://build.opensuse.org/package/show/home:uncomfyhalomacro:branches:editors:registry/kak-lsp

You can see here that in https://build.opensuse.org/package/show/home:uncomfyhalomacro:branches:editors:registry/kak-lsp, only **aarch64**, **x86_64**, and **ppc64le** are able to find their dependencies and compile (whether it fails or not). **i586**, **armv7l**, and **s390x** are the architectures that were not able to find their dependencies.

Everything in https://build.opensuse.org/package/show/home:uncomfyhalomacro:branches:editors:vendor/kak-lsp are able to find their dependencies and are able to compile (whether it fails or not).

I would like to know if there is an inaccuracy with the documentation or not.

For context, I've been helping out to build a tool to ease the process of vendoring dependencies in openSUSE/SUSE. https://github.com/openSUSE-Rust/obs-service-cargo.

weihanglo:
Thanks for the bug report. Could you attach some additional information, like
* The version of Cargo (the output of `cargo -vV`) was observed with this behavior.
* If this is a regression, the last successful version of Cargo.
* The exact Cargo commands or scripts ran in that service.
* Any other `[source]` replacement or proxies configured in the server or not.

`cargo vendor` and `cargo fetch` has no difference in the sense of dependency resolution (except `cargo vendor` accepts multiple workspaces). However, there are slightly different when filtering platforms:

[...]

PR #14116, opened 2024-06-20
feat!: use stable hash from rustc-stable-hash

weihanglo:
### What does this PR try to resolve?

This helps `-Ztrim-paths` build a stable cross-platform path for the
registry and git sources. Sources files then can be found from the same
path when debugging.

See https://github.com/rust-lang/cargo/issues/13171#issuecomment-1864899037

### How should we test and review this PR?

There are a few caveats, and we should do an FCP before merge:

* This will invalidate the current downloaded caches.
  Need to put this in the Cargo CHANGELOG.
* As a consequence of changing how `SourceId` is hashed, the global cache
  tracker is also affected because [Cargo writes source identifiers (e.g.
  `index.crates.io-6f17d22bba15001f`) to SQLite](https://github.com/rust-lang/cargo/blob/6e236509b2331eef64df844b7bbc8ed352294107/src/cargo/core/global_cache_tracker.rs#L388-L391).
* The performance of rustc-stable-hash is slightly worse than the old
  SipHasher in std on short things like `SourceId`, but for long stuff
  like fingerprint. See Additional information.

StableHasher is used in several places. We should consider if there is a need for cryptographyic hash (see <https://github.com/rust-lang/cargo/issues/13171#issuecomment-2179705494>).

* Rebuild detection (fingerprints)
  * [Rustc version]( https://github.com/rust-lang/cargo/blob/6e236509b2331eef64df844b7bbc8ed352294107/src/cargo/util/rustc.rs#L326), including [all the CLI args](https://github.com/rust-lang/cargo/blob/6e236509b2331eef64df844b7bbc8ed352294107/src/cargo/util/rustc.rs#L381) running `rustc -vV`.
  * [Build caches](https://github.com/rust-lang/cargo/blob/6e236509b2331eef64df844b7bbc8ed352294107/src/cargo/core/compiler/fingerprint/mod.rs#L1456)
* Compute rustc `-C metadata`
  * [stable hash for SourceId](https://github.com/rust-lang/cargo/blob/6e236509b2331eef64df844b7bbc8ed352294107/src/cargo/core/package_id.rs#L207)
  * Also read and hash contents from [custom target JSON file](https://github.com/rust-lang/cargo/blob/6e236509b2331eef64df844b7bbc8ed352294107/src/cargo/core/compiler/compile_kind.rs#L81-L91).
[...]

PR #14917, opened 2024-12-10
feature(SourceId): use stable hash from rustc-stable-hash

weihanglo:

This helps `-Ztrim-paths` build a stable cross-platform path for the
registry and git sources. Sources files then can be found from the same
path when debugging.

It also helps cache registry index all at once for all platforms,
for example the use case in https://github.com/rust-lang/cargo/issues/14795
(despite they should use `cargo vendor` instead IMO).

Some caveats:

* Newer cargo will need to re-download files for global caches
  (index files, git/registry sources).
  The old cache is still kept and used when running with older cargoes.
* Absolute paths on windows iarenot really covered by the "cross-platform" hash,
  because path prefix components like `C:` are always there.
  That means hashes of some sources kind,
  like local registry and local path,
  are not going to be real cross-platform stable.

#### Security concern

There might be hash collisions if you have two registries under the same
domain. This won't happen to crates.io, as the infra would have to
intentionally put another registry on index.crates.io to collide.
We don't consider this is an actual threat model, so we are not going to
use any cryptographically secure hash algorithm like BLAKE3.

At least, the current unstable SipHash isn't in a better situation.
We might switch to a cryptographic secure one when needed.

See also <https://github.com/rust-lang/cargo/issues/13171#issuecomment-2181465128>

We have an FCP in <https://github.com/rust-lang/cargo/issues/14795#issuecomment-2498446176>.

This PR implements the proposal,
The path-length concern in <https://github.com/rust-lang/cargo/issues/14795#issuecomment-2532076078> is automatically addressed
because we don't need cryptographically secure hash for now.

### Additional information

See more information and benchmark results in <https://github.com/rust-lang/cargo/pull/14116>.

### DIFF
--- a/src/cargo/core/compiler/build_runner/compilation_files.rs
+++ b/src/cargo/core/compiler/build_runner/compilation_files.rs
@@ -718,8 +718,8 @@ fn compute_metadata(
         }
     }
 
-    let c_metadata = UnitHash(c_metadata_hasher.finish());
-    let c_extra_filename = UnitHash(c_extra_filename_hasher.finish());
+    let c_metadata = UnitHash(Hasher::finish(&c_metadata_hasher));
+    let c_extra_filename = UnitHash(Hasher::finish(&c_extra_filename_hasher));
     let unit_id = c_extra_filename;
 
     let c_extra_filename = use_extra_filename.then_some(c_extra_filename);
--- a/src/cargo/core/source_id.rs
+++ b/src/cargo/core/source_id.rs
@@ -782,70 +782,89 @@ mod tests {
     // Otherwise please just leave a comment in your PR as to why the hash value is
     // changing and why the old value can't be easily preserved.
     //
-    // The hash value depends on endianness and bit-width, so we only run this test on
-    // little-endian 64-bit CPUs (such as x86-64 and ARM64) where it matches the
-    // well-known value.
+    // The hash value should be stable across platforms, and doesn't depend on
+    // endianness and bit-width. One caveat is that absolute paths on Windows
+    // are inherently different than on Unix-like platforms. Unless we omit or
+    // strip the prefix components (e.g. `C:`), there is not way to have a true
+    // cross-platform stable hash for absolute paths.
     #[test]
-    #[cfg(all(target_endian = "little", target_pointer_width = "64"))]
-    fn test_cratesio_hash() {
-        let gctx = GlobalContext::default().unwrap();
-        let crates_io = SourceId::crates_io(&gctx).unwrap();
-        assert_eq!(crate::util::hex::short_hash(&crates_io), "1ecc6299db9ec823");
-    }
-
-    // See the comment in `test_cratesio_hash`.
-    //
-    // Only test on non-Windows as paths on Windows will get different hashes.
-    #[test]
-    #[cfg(all(target_endian = "little", target_pointer_width = "64", not(windows)))]
     fn test_stable_hash() {
         use std::hash::Hasher;
         use std::path::Path;
 
+        use crate::util::StableHasher;
+
+        #[cfg(not(windows))]
+        let ws_root = Path::new("/tmp/ws");
+        #[cfg(windows)]
+        let ws_root = Path::new(r"C:\\tmp\ws");
+
         let gen_hash = |source_id: SourceId| {
-            let mut hasher = std::collections::hash_map::DefaultHasher::new();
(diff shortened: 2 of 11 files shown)


src/cargo/core/compiler/build_runner/compilation_files.rs
```diff
@@ -718,8 +718,8 @@ fn compute_metadata(
         }
     }
 
-    let c_metadata = UnitHash(c_metadata_hasher.finish());
-    let c_extra_filename = UnitHash(c_extra_filename_hasher.finish());
+    let c_metadata = UnitHash(Hasher::finish(&c_metadata_hasher));
+    let c_extra_filename = UnitHash(Hasher::finish(&c_extra_filename_hasher));
```
epage, line 722: Why do we need to make this change in changing the hasher?

Both implement the trait and its in scope.

src/cargo/util/hasher.rs
```diff
@@ -1,25 +1,3 @@
-//! Implementation of a hasher that produces the same values across releases.
-//!
-//! The hasher should be fast and have a low chance of collisions (but is not
-//! sufficient for cryptographic purposes).
```
epage, line 4: Should we keep this part of the comment?

src/cargo/core/source_id.rs
```diff
@@ -782,70 +782,89 @@ mod tests {
...
     // changing and why the old value can't be easily preserved.
     //
-    // The hash value depends on endianness and bit-width, so we only run this test on
-    // little-endian 64-bit CPUs (such as x86-64 and ARM64) where it matches the
-    // well-known value.
+    // The hash value should be stable across platforms, and doesn't depend on
+    // endianness and bit-width. One caveat is that absolute paths on Windows
+    // are inherently different than on Unix-like platforms. Unless we omit or
+    // strip the prefix components (e.g. `C:`), there is not way to have a true
+    // cross-platform stable hash for absolute paths.
```
epage, line 789: I've updated the PR description to clarify that the Windows issue is specifically when hashing paths.

Granted, requiring people to have the same paths on different systems, regardless of platform, is already not ideal and is part of the point of trim paths.

tests/testsuite/global_cache_tracker.rs
```diff
@@ -2004,7 +2004,16 @@ fn compatible_with_older_cargo() {
     assert_eq!(get_registry_names("src"), ["middle-1.0.0", "new-1.0.0"]);
     assert_eq!(
         get_registry_names("cache"),
-        ["middle-1.0.0.crate", "new-1.0.0.crate", "old-1.0.0.crate"]
+        // Duplicate crates from two different cache location
```
weihanglo, line 2007: I am not sure if garbage collection should support cases like this.
Ideal it should?

weihanglo: cc @ehuss

src/cargo/core/compiler/build_runner/compilation_files.rs
weihanglo: Because it resolves to `StableHasher::finish` first.

https://github.com/rust-lang/rustc-stable-hash/blob/a7bb3b77597092f1743362b6327d4324544d6d95/src/stable_hasher.rs#L167-L169

Do you have other workarounds to make it resolves to `Hasher::finish`?

tests/testsuite/global_cache_tracker.rs
weihanglo: Actually it is doing its job correctly, so no need to worry about changing these tests.

src/cargo/util/hasher.rs
weihanglo: Added it back!

src/cargo/core/compiler/build_runner/compilation_files.rs
epage: ... if their `finish` isn't compatible with regular usage of `Hasher::finish` then thats an odd choice.

How come we can't rely on `FromStableHash`?  ... they have no impls.  *sigh*

epage (approved):
FCP in #14795.  We likely don't need to wait the 10 day waiting period.  If something comes up, we have time to revert before the next release.

ayanwaar:
5eb7480565adfaf79bfdf90de6a7ff5ea5dcd0eb

ayanwaar:
5eb7480565adfaf79bfdf90de6a7ff5ea5dcd0eb

PR #14919, opened 2024-12-11
feat(toml): Add support for setting the default package version (#6583)

sunhaitao:
Allow changing the default package version by setting the `CARGO_SUGGESTED_PKG_VERSION` environment variable.

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -17,6 +17,7 @@ use cargo_util_schemas::manifest::{RustVersion, StringOrBool};
 use itertools::Itertools;
 use lazycell::LazyCell;
 use pathdiff::diff_paths;
+use semver::Version;
 use toml_edit::ImDocument;
 use url::Url;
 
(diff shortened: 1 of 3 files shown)


src/cargo/util/toml/mod.rs
```diff
@@ -566,7 +567,16 @@ fn normalize_package_toml<'a>(
             .clone()
             .map(|value| field_inherit_with(value, "version", || inherit()?.version()))
             .transpose()?
-            .map(manifest::InheritableField::Value),
+            .map(manifest::InheritableField::Value)
```
weihanglo, line 570: Thanks for the contribution, though this PR seems incomplete. How does this one play well with SemVer? How does the env affect the entire dependency graph?

Note that there are a couple of different use cases in <https://github.com/rust-lang/cargo/issues/6583>. The design of it is not yet settled, so I am going to mark this as draft. In Cargo we encourage discussions before implementations. See https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo

epage: Note that its not just `S-needs-design` or `S-needs-team-input` but its `S-propose-close`.

weihanglo: Yes. Going to close this as it is incomplete and unlikely to merge. Feel free to put your use cases in #6583 and we may consider changing its label status.

sunhaitao: It plays as well as current practices with SemVer. When building multiple packages in a call of `cargo`, all unversioned packages will be affected -- as many similar tools for other languages. I'll post more details on #6583.

Issue #12118, opened 2023-05-10
cross-compilation doctests warning should not be behind verbose flag


timfish:
### Problem

I just spent a lot longer than I'm willing to admit trying to work out why my doctests were not running. After working out I'd done everything correctly I finally tried `--verbose` and found the warning about cross-compilation not being supported.

`cargo test` completes successfully and I've not even known that the doctests have been skipped for months! Why is this warning hidden behind the verbose flag?

weihanglo:
It was an intentional choice in <https://github.com/rust-lang/cargo/pull/10132>. Warning for every doctest target may be noisy. Could be improved by warning just once in normal mode if there is a cross-compile doctest, hinting people to use `--verbose` to see the full list of skipped doctest targets.

Relevant code is here:
https://github.com/rust-lang/cargo/blob/0e93732f26420ecca7803b2eabe1e5ec98015d32/src/cargo/ops/cargo_test.rs#L190-L205

weihanglo:
Just a friendly reminder. It would be great for collaboration if people always strive to provide reproducible steps and version information for their issues. That may increase the chance for anyone wanting to fix gets started sooner.

timfish:
> provide reproducible steps and version information for their issues

Apologies. Just added these.

weihanglo:
Labeled as <https://github.com/rust-lang/cargo/labels/P-low> as we are considering to stabilize `-Zdoctest-xcompile`, so this warning may be unnecessary. There is no ETA though.

1garo:
Would be okay if I work on this one? @weihanglo

1garo:
I'm just having trouble building, I been having the message below in some crates:

`error[E0658]: use of unstable library feature 'once_cell'`

Any ideia?

hi-rustin:
Try to use the latest stable version or nightly version rustc.

1garo:
> Try to use the latest stable version or nightly version rustc.

Thanks! Now I'm having some problem with linking `error: linking with `cc` failed: exit status: 1` when `cargo build`, but `cargo check` is fine!

Mac M1 ventura 13.2

weihanglo:
> Labeled as <https://github.com/rust-lang/cargo/labels/P-low> as we are considering to stabilize `-Zdoctest-xcompile`, so this warning may be unnecessary. There is no ETA though.

As it is labeled as `P-low` and `S-needs-mentor`, I'd suggest working on other issues if you're interested in contributing Cargo.

[...]

PR #14920, opened 2024-12-11
fix(doctest): warn only once if doctest xcompile is skipped

progressive-galib:
as per issue/comment https://github.com/rust-lang/cargo/issues/12118#issuecomment-1541860834 added warning about skipping doctest on xcompiling even when not under verbose flag

#this pr is an attempt at solving https://github.com/rust-lang/cargo/issues/12118. i tested them myself still might need a little work as concise waring shows every time we skip a test instead of once.

### DIFF
--- a/src/cargo/ops/cargo_test.rs
+++ b/src/cargo/ops/cargo_test.rs
@@ -178,6 +178,7 @@ fn run_doc_tests(
     let mut errors = Vec::new();
     let doctest_xcompile = gctx.cli_unstable().doctest_xcompile;
     let color = gctx.shell().color_choice();
+    let mut warned_once = false;
 
     for doctest_info in &compilation.to_doc_test {
         let Doctest {
@@ -194,10 +195,13 @@ fn run_doc_tests(
                 CompileKind::Host => {}
                 CompileKind::Target(target) => {
                     if target.short_name() != compilation.host {
+                        if !warned_once {
+                            gctx.shell().concise(|shell| {
+                                shell.warn("skipping cross-compilation doctest(s), use --verbose to see the full list in detail".to_string())
+                            })?;
+                            warned_once = true;
+                        }
 
-                        gctx.shell().concise(|shell| {
-                            shell.warn("skipping cross-compilation doctest(s), use --verbose to see the full list in detail".to_string())
-                        })?;
                         // Skip doctests, -Zdoctest-xcompile not enabled.
                         gctx.shell().verbose(|shell| {
                             shell.note(format!(
--- a/src/cargo/ops/cargo_test.rs
+++ b/src/cargo/ops/cargo_test.rs
@@ -197,7 +197,7 @@ fn run_doc_tests(
                     if target.short_name() != compilation.host {
                         if !warned_once {
                             gctx.shell().concise(|shell| {
-                                shell.warn("skipping cross-compilation doctest(s), use --verbose to see the full list in detail".to_string())
+                                shell.warn("skipping cross-compilation doctest(s), use --verbose to see the full list in detail.".to_string())
                             })?;
                             warned_once = true;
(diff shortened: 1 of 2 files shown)


src/cargo/ops/cargo_test.rs
```diff
@@ -194,6 +194,10 @@ fn run_doc_tests(
                 CompileKind::Host => {}
                 CompileKind::Target(target) => {
                     if target.short_name() != compilation.host {
+
```
weihanglo, line 197: Thanks for the contribution.

We need a test for this warning to make sure it doesn't emit excessive warnings. 

https://github.com/rust-lang/cargo/issues/12118#issuecomment-1541860834

progressive-galib: currently it will show every time a doctest is passed over (i.e skipped)

else we are gonna need to use something like, 

`pub static ref WARNONCE: Mutex<bool> = Mutex::new(false);`

(btw i really like lazy_static crate but i kinda admire  package count of a self hosted compiler)

weihanglo: This part of cargo is single-threaded, so perhaps a simple variable here serves the purpose.

https://github.com/rust-lang/cargo/blob/94d274d6920aaabf44ee56a786dd0ac6f99759a0/src/cargo/ops/cargo_test.rs#L181

progressive-galib: done, pushed it with without locally building.... man as someone working with very slow computer gotta say i just 💖 your CI/CD set up.

progressive-galib: done, all tests have passed

ehuss:
I just want to mildly push back on this, since the original warning was hidden for a reason. It is difficult or impossible for the user to remove the warning. I suspect this will show up for everyone who runs `cargo test --target` which I suspect will be a large number. For example, I think it will start appearing in every CI run for rust-lang/rust. I understand the desire to bring awareness that something is being skipped, but if there is nothing the user can do about it, I'm not sure that is helpful. I'd like to better understand the motivation here.

progressive-galib:
> I just want to mildly push back on this, since the original warning was hidden for a reason. It is difficult or impossible for the user to remove the warning. I suspect this will show up for everyone who runs `cargo test --target` which I suspect will be a large number. For example, I think it will start appearing in every CI run for rust-lang/rust. I understand the desire to bring awareness that something is being skipped, but if there is nothing the user can do about it, I'm not sure that is helpful. I'd like to better understand the motivation here.

•)it will only warn only once on the first doctest, not for each doctest,
•)its just one line in a 10000 line CI log, I agree with you better to know it was skipped for those that need them
•)if users need the warning turned off there  can be a flag for once users use it and ask for the flag

progressive-galib:
> @epage any update ?

sorry i meant @ehuss

weihanglo:
> > I'm not sure that is helpful. I'd like to better understand the motivation here.
>
> •)it will only warn only once on the first doctest, not for each doctest,
> •)its just one line in a 10000 line CI log, I agree with you better to know it was skipped for those that need them
> •)if users need the warning turned off there can be a flag for once users use it and ask for the flag

I don't consider these motivations ehuss was asking for. These are implementation decisions.

progressive-galib:
@ehuss please approve the review, i  think its worth it, read the motivations above


> > 
> > 
> > •)it will only warn only once on the first doctest, not for each doctest,
> > •)its just one line in a 10000 line CI log, I agree with you better to know it was skipped for those that need them
> > •)if users need the warning turned off there can be a flag for once users use it and ask for the flag
>

ehuss:
I'm going to close as resolved by https://github.com/rust-lang/cargo/pull/15462.

Issue #4260, opened 2017-07-08
Mismatch against yanked version doesn't inform user about yank


crumblingstatue:
Depending on `bincode = "1.0.0-alpha7"` produces the following error:

```
no matching version `^1.0.0-alpha7` found for package `bincode` (required by `whatever`)
location searched: registry https://github.com/rust-lang/crates.io-index
versions found: 0.8.0, 0.7.0, 0.6.1, ...
```

It doesn't mention the fact that there _is_ a `1.0.0-alpha7`, but it has been yanked.

This can leave the user confused about whether they mistyped something, or they are just trying to depend on a yanked version.

Semi-related: https://github.com/onur/docs.rs/issues/112#issuecomment-313868337

epage:
Looks like the error hasn't improved in the intervening time
```rust
#!/usr/bin/env nargo
```cargo
[dependencies]
bincode = "=1.0.0-alpha7"
```

fn main () {}
```
```console
$ ./cargo-4260.rs
warning: `package.edition` is unspecified, defaulting to `2021`
    Updating crates.io index
error: failed to select a version for the requirement `bincode = "=1.0.0-alpha7"`
candidate versions found which didn't match: 2.0.0-rc.3, 2.0.0-rc.2, 2.0.0-rc.1, ...
location searched: crates.io index
required by package `cargo-4260 v0.0.0 (/home/epage/src/personal/dump)`
if you are looking for the prerelease package it needs to be specified explicitly
    bincode = { version = "2.0.0-rc.3" }
perhaps a crate was updated and forgotten to be re-vendored?
```

@Eh2406 how bad would this be to improve the error in this case?

Eh2406:
The resolver is reporting that the package doesn't exist, because it got filtered out and a much lower level of abstraction.
This is part of the work in https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/better.20error.20messages.20for.20filtered.20versions.2E

glandium:
Note sure I should open a separate issue, but relatedly, the error from `cargo update -p crate --precise yanked-version` says `error: no matching package named "crate" found`, which is confusing.

epage:
Allowing `--precise` with yanked packages was implemented in #13333.  It is currently unstable but we expect to stabilize it soon.

PR #14921, opened 2024-12-11
fix(resolver): Don't report all versions as rejected 

epage:

When I copy/pasted the alternative-names code in #14897, I didn't notice that the
wildcard dependency was used.  This includes a rename of the variable to make it stand out more.

I also renamed `AlternativeVersions` because there is another check we do that better matches that name, so I renamed it to `RejectedVersions`.

This does not have a test yet because overly-constrictive dep specs have
higher precedence atm (which is what I was working on when I found this).

### DIFF
--- a/crates/resolver-tests/src/lib.rs
+++ b/crates/resolver-tests/src/lib.rs
@@ -143,7 +143,7 @@ pub fn resolve_with_global_context_raw(
             for summary in self.list.iter() {
                 let matched = match kind {
                     QueryKind::Exact => dep.matches(summary),
-                    QueryKind::AlternativeVersions => dep.matches(summary),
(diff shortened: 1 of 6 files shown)


weihanglo (approved):
This does not have a test yet because overly-constrictive dep specs have
higher precedence atm (which is what I was working on when I found this).

What is your plan?

Anyway this seems good to merge.

epage:
> What is your plan?

Our current error precedence is
1. bad version req
2. rejected version
3. bad name

Bad version req going first makes it hard to see this.  I was already working on changing the order to
1. rejected version
2. bad version req
3. bad name

which is needed to fix issues like #4260.

Once that is done, I can better handle tests for this case.

PR #14922, opened 2024-12-11
Faster conflict cache

Eh2406:
Builds on #14915 

The current code loops over all package versions we have had past problems with to see which ones are currently active. This branch checks which version of the package is active to see if we have a conflict for it. If resolution is dominated by a package with many versions, then this find operation should go from `O(n)` to `O(1)`.


Eh2406:
After fixing the bug, the performance numbers seem worse than master.  The added layer of indirection may be too expensive. One more thing to try before I close.

Eh2406:
No it is not faster. Sorry.

Issue #14894, opened 2024-12-05
Index parsing errors are swallowed


jonhoo:
### Problem

When using a private registry, I recently ran into a case where a crate version that _is_ in the index (confirmed by cURLing the registry directly) could not be resolved by Cargo, giving me an error like
```
error: no matching package named `something` found
location searched: `private-registry` index
```

I quickly suspected that the index entry was invalid somehow, but Cargo gave no errors or warnings, even with `CARGO_LOG=trace`. Furthermore, by tweaking `http_remote.rs`, I found that its query to the registry (with an `if-none-match` header) returned `NotModified`, indicating Cargo _did_ have the etag of the latest index file, which did include the crate/version in question. By removing the etag header, I also confirmed that the contents Cargo got from its request contained the version in question.

Long story short, I discovered that Cargo ends up swallowing errors related to invalid index entries. The first time Cargo downloads an index file with an invalid entry, it hits

https://github.com/rust-lang/cargo/blob/05f54fdc34310f458033af8a63ce1d699fae8bf6/src/cargo/core/summary.rs#L345-L348

However, that gets turned into a `tracing::info` in

https://github.com/rust-lang/cargo/blob/05f54fdc34310f458033af8a63ce1d699fae8bf6/src/cargo/sources/registry/index/mod.rs#L604-L619

The user is likely to miss this since Cargo tracing logs are off by default, so they will just observe that Cargo claims the version doesn't exist.

Unfortunately, the issue doesn't stop there. If the user _then_ tries to run with `CARGO_LOG=trace`, they will then see _no_ errors or warnings from Cargo. In fact, the logs will make it seem like the invalid entry doesn't exist at all. There will be no trace of it anywhere. This is because when Cargo parses an index entry, it caches the parsed lines so that it doesn't have to parse them again

https://github.com/rust-lang/cargo/blob/05f54fdc34310f458033af8a63ce1d699fae8bf6/src/cargo/sources/registry/index/mod.rs#L622

+

https://github.com/rust-lang/cargo/blob/05f54fdc34310f458033af8a63ce1d699fae8bf6/src/cargo/sources/registry/index/mod.rs#L627-L631

[...]

PR #14923, opened 2024-12-11
fix(resolver): In errors, show rejected versions over alt versions

### LINKED ISSUE #4260: Mismatch against yanked version doesn't inform user about yank

epage:

The user likely intended what they said and we should tell them why it
didn't work.
Rejected versions also imply alt versions below them.

Fixes #4260

In changing the errors from alt-versions to rejected-versions, I found part of the old error message was better and changed the message introduced in #14897 from matching alt-names to alt-versions.

This includes the test for #14921.  The "fix" was done before this because as it was part of the refactors in prep for this.

### Additional information

Next steps after this are to still produce an `IndexSummary::Unsupported`  even if there are deserialization errors, so long as we know the name and version which should take care of #10623 and #14894.

### DIFF
--- a/src/cargo/core/resolver/errors.rs
+++ b/src/cargo/core/resolver/errors.rs
@@ -221,7 +221,51 @@ pub(super) fn activation_error(
     // give an error message that nothing was found.
     let mut msg = String::new();
     let mut hints = String::new();
-    if let Some(candidates) = alt_versions(registry, dep) {
+    if let Some(version_candidates) = rejected_versions(registry, dep) {
+        let version_candidates = match version_candidates {
+            Ok(c) => c,
+            Err(e) => return to_resolve_err(e),
+        };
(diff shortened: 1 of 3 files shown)


ehuss (approved):
Thanks, looks great!

PR #12224, opened 2023-06-02
test(z-flags): Verify `-Z` flags list is sorted

Muscraft:
#12182 made it so that `unstable_cli_options!` was sorted, but had no way to make sure it would stay sorted in the future. 

In [this thread](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Sorting.20features!.20and.20ensuring.20it.20stays.20sorted), I proposed that a test should be added that would fail if the list is not sorted, this PR adds that test.

If the list is not sorted the test output looks like this:
Screenshot from 2023-06-02 10-17-08

PR #14925, opened 2024-12-11
fix(script): Don't override the release profile

epage:
This has been around since #12224 and isn't in the RFC, so its being removed.

This is part of #12207.


weihanglo (approved):
I forgot why it was there. Anyway thanks for catching this!

PR #14926, opened 2024-12-11
fix: emit_serialized_unit_graph uses the configured shell

aatifsyed:
I'm writing a compatibility layer between the cargo and meson build systems.
This change allows me to plumb that through better :)

### DIFF
--- a/src/cargo/core/compiler/unit_graph.rs
+++ b/src/cargo/core/compiler/unit_graph.rs
@@ -10,7 +10,6 @@ use crate::util::interning::InternedString;
 use crate::util::CargoResult;
 use crate::GlobalContext;
 use std::collections::HashMap;
-use std::io::Write;
 
 /// The dependency graph of Units.
 pub type UnitGraph = HashMap<Unit, Vec<UnitDep>>;
@@ -121,15 +120,10 @@ pub fn emit_serialized_unit_graph(
             }
         })
         .collect();
-    let s = SerializedUnitGraph {
+
+    gctx.shell().print_json(&SerializedUnitGraph {
         version: VERSION,
         units: ser_units,
         roots,
-    };
-
-    let stdout = std::io::stdout();
-    let mut lock = stdout.lock();
-    serde_json::to_writer(&mut lock, &s)?;
-    drop(writeln!(lock));
-    Ok(())
+    })
 }


epage:
What difference is this making for you?

aatifsyed:
I'm using cargo-as-a-library to get the unit graph.

Before this change:
```
fn main() -> anyhow::Result<()> {
    match env::var_os("CARGO2MESON_CHILD").is_some() {
        // we're the child process, so run cargo so that it emits the unit graph
        true => {
            // ...
            let mut options = CompileOptions::new(&gctx, CompileMode::Build)?;
            options.build_config.unit_graph = true;
            cargo::ops::compile(&Workspace::new(&manifest_path, &gctx)?, &options)?;
        },
        // we're the parent process, so spawn ourselves and capture stdout
        false => {
            // ...
            let output = Command::new(env::current_exe()).env("CARGO2MESON_CHILD", "1") /* .arg(..) */ .output()?;
            serde_json::from_slice(&output.stdout)?;
            // ...
        }
    }
    Ok(())
}
```

After this change:
```
fn main() -> anyhow::Result<()> {
    // ...
    let shared_writer = SharedWriter::new(vec![]);
    let gtcx = GlobalContext::new(Shell::from_writer(Box::new(shared_writer.clone())), path, path);
    let mut options = CompileOptions::new(&gctx, CompileMode::Build)?;
    options.build_config.unit_graph = true;
    cargo::ops::compile(&Workspace::new(&manifest_path, &gctx)?, &options)?;
    serde_json::from_slice(&shared_writer.take())?;
    // ...
}
```

I won't have to workaround by spawning myself, and can use existing cargo mechanisms to graph the output :)

Happy to provide less abridged examples, but I hope the above gets the point across :)

src/cargo/core/compiler/unit_graph.rs
```diff
@@ -127,9 +126,9 @@ pub fn emit_serialized_unit_graph(
...
     };
 
-    let stdout = std::io::stdout();
-    let mut lock = stdout.lock();
-    serde_json::to_writer(&mut lock, &s)?;
-    drop(writeln!(lock));
+    let mut shell = gctx.shell();
+    let stdout = shell.out();
+    serde_json::to_writer(&mut *stdout, &s)?;
+    writeln!(stdout)?;
```
weihanglo, line 132: Maybe better using `gctx.shell().print_json(&result)?` as we have that function

weihanglo (dismissed):
The change itself looks reasonable.

Just remind that libcargo is provided as-is, so you may find it painful when a `cargo` version bump is needed 😆.

src/cargo/core/compiler/unit_graph.rs
aatifsyed: Thanks! applied :)

aatifsyed:
I've applied @weihanglo great suggestion :)

weihanglo (approved):

PR #6880, opened 2019-04-26
Parse less JSON on null builds 

alexcrichton:
This commit fixes a performance pathology in Cargo today. Whenever Cargo
generates a lock file (which happens on all invocations of `cargo build`
for example) Cargo will parse the crates.io index to learn about
dependencies. Currently, however, when it parses a crate it parses the
JSON blob for every single version of the crate. With a lock file,
however, or with incremental builds only one of these lines of JSON is
relevant. Measured today Cargo building Cargo parses 3700 JSON
dependencies in the registry.

This commit implements an optimization that brings down the number of
parsed JSON lines in the registry to precisely the right number
necessary to build a project. For example Cargo has 150 crates in its
lock file, so now it only parses 150 JSON lines (a 20x reduction from
3700). This in turn can greatly improve Cargo's null build time. Cargo
building Cargo dropped from 120ms to 60ms on a Linux machine and 400ms
to 200ms on a Mac.

The commit internally has a lot more details about how this is done but
the general idea is to have a cache which is optimized for Cargo to read
which is maintained automatically by Cargo.

Closes #6866

PR #14927, opened 2024-12-12
fix(resolver): Report invalid index entries 

### LINKED ISSUE #10623: cargo <1.60 failed to select a version of a dependency due to either `?` (weak) or `dep:` (namespaced) features

### LINKED ISSUE #14894: Index parsing errors are swallowed

epage:

While #14897 reported packages with an unsupported index schema version,
that only worked if the changes in the schema version did not cause
errors in deserializing `IndexPackage` or in generating a `Summary`.

This extends that change by recoverying on error with a more lax,
incomplete parse of `IndexPackage` which should always generate a valid
`Summary`.

To help with a buggy Index, we also will report as many as we can.
This does not provide a way to report to users or log on cache reads if
the index entry is not at least `{"name": "<string>", "vers": "<semver>"}`.

Fixes #10623
Fixes #14894

My biggest paranoia is some bad interaction with the index cache including more "invalid" index entries.
That should be ok as we will ignore the invalid entry in the cache when
loading it.
Ignoring of invalid entries dates back to #6880 when the index cache was
introduced.

### DIFF
--- a/tests/testsuite/registry.rs
+++ b/tests/testsuite/registry.rs
@@ -2916,7 +2916,9 @@ fn ignore_invalid_json_lines_git() {
 
 fn ignore_invalid_json_lines() {
     Package::new("foo", "0.1.0").publish();
-    Package::new("foo", "0.1.1").invalid_json(true).publish();
+    Package::new("foo", "0.1.1")
+        .invalid_index_line(true)
+        .publish();
     Package::new("foo", "0.2.0").publish();
 
     let p = project()
--- a/tests/testsuite/registry.rs
+++ b/tests/testsuite/registry.rs
@@ -3012,10 +3012,13 @@ fn invalid_json_lines_error() {
         .with_stderr_data(str![[r#"
 [UPDATING] `dummy-registry` index
 [ERROR] failed to select a version for the requirement `foo = "^0.1.1"`
-candidate versions found which didn't match: 0.2.0, 0.1.0
+  version 0.1.3 requires cargo 1.2345
+  version 0.1.4 requires a Cargo version that supports index version 1000000000
+  version 0.1.5's index entry is invalid
+  version 0.1.6 requires a Cargo version that supports index version 1000000000
+  version 0.1.7's index entry is invalid
 location searched: `dummy-registry` index (which is replacing registry `crates-io`)
 required by package `a v0.5.0 ([ROOT]/foo)`
-perhaps a crate was updated and forgotten to be re-vendored?
 
 "#]])
         .run();
@@ -3024,10 +3027,13 @@ perhaps a crate was updated and forgotten to be re-vendored?
         .with_stderr_data(str![[r#"
 [UPDATING] `dummy-registry` index
 [ERROR] failed to select a version for the requirement `foo = "^0.1.1"`
-candidate versions found which didn't match: 0.2.0, 0.1.0
+  version 0.1.3 requires cargo 1.2345
+  version 0.1.4 requires a Cargo version that supports index version 1000000000
+  version 0.1.5's index entry is invalid
+  version 0.1.6 requires a Cargo version that supports index version 1000000000
+  version 0.1.7's index entry is invalid
 location searched: `dummy-registry` index (which is replacing registry `crates-io`)
 required by package `a v0.5.0 ([ROOT]/foo)`
-perhaps a crate was updated and forgotten to be re-vendored?
 
 "#]])
         .run();
--- a/tests/testsuite/registry.rs
+++ b/tests/testsuite/registry.rs
@@ -2943,6 +2943,97 @@ fn ignore_invalid_json_lines() {
 }
 
 #[cargo_test]
+fn invalid_json_lines_error() {
+    Package::new("foo", "0.1.0")
+        .rust_version("1.0")
+        .schema_version(2)
+        .publish();
+    Package::new("foo", "0.1.1")
+        // Bad name field, too corrupt to use
+        .invalid_index_line(true)
+        .publish();
+    Package::new("foo", "0.1.2")
+        // Bad version field, too corrupt to use
+        .index_line(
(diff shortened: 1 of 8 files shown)


weihanglo (approved):
This is nice! Thank you.

tests/testsuite/registry.rs
```diff
@@ -2940,6 +2942,103 @@ fn ignore_invalid_json_lines() {
...
+required by package `a v0.5.0 ([ROOT]/foo)`
+
+"#]])
+        .run();
+    p.cargo("generate-lockfile")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[UPDATING] `dummy-registry` index
+[ERROR] failed to select a version for the requirement `foo = "^0.1.1"`
+  version 0.1.3 requires cargo 1.2345
```
weihanglo, line 3030: I wish we could perhaps have a log on disk recording all this internal error

weihanglo (approved):

PR #14928, opened 2024-12-12
feat: Implement `--depth workspace` for `cargo tree` command

### LINKED ISSUE #14420: Option for `cargo tree` to filter out (non-)workspace members
### Problem

I'm opening this issue because https://github.com/rust-lang/cargo/issues/8105#issuecomment-1284860085 said to open one if one wanted that feature.

`cargo tree` allows with `--prune` to remove dependencies from the list. However, as mentioned in https://github.com/rust-lang/cargo/issues/8105#issuecomment-846336129 it only takes in `SPEC`, and not the proposed "non-members" or "members".

This would be useful to have, since it would allow easily viewing the dependency tree with only workspace members.

### Proposed Solution

The solution would be adding some argument, that would then take the values `both`, `member`, and `non-member` (as mentioned in https://github.com/rust-lang/cargo/issues/8105#issuecomment-846336129).

I have no proposed name yet for the argument.

### Notes

At the time of writing, there does not exist any such command, see https://github.com/rust-lang/cargo/blob/ba8b39413c74d08494f94a7542fe79aa636e1661/src/doc/src/commands/cargo-tree.md for the `cargo tree` documentation at the time of writing.

ShoyuVanilla:
Resolves #14420

### DIFF
--- a/src/bin/cargo/commands/tree.rs
+++ b/src/bin/cargo/commands/tree.rs
@@ -2,7 +2,7 @@ use crate::cli;
 use crate::command_prelude::*;
 use anyhow::{bail, format_err};
 use cargo::core::dependency::DepKind;
-use cargo::ops::tree::{self, EdgeKind};
+use cargo::ops::tree::{self, DisplayDepth, EdgeKind};
 use cargo::ops::Packages;
 use cargo::util::print_available_packages;
 use cargo::util::CargoResult;
@@ -162,6 +162,12 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
 
     let pkgs_to_prune = args._values_of("prune");
 
+    let display_depth = args
+        ._value_of("depth")
+        .map(|s| s.parse::<DisplayDepth>())
+        .transpose()?
+        .unwrap_or(DisplayDepth::MaxDisplayDepth(u32::MAX));
+
     let packages = args.packages_from_flags()?;
     let mut invert = args
         .get_many::<String>("invert")
@@ -222,7 +228,7 @@ subtree of the package given to -p.\n\
         duplicates: args.flag("duplicates"),
         format: args.get_one::<String>("format").cloned().unwrap(),
         graph_features,
-        max_display_depth: args.value_of_u32("depth")?.unwrap_or(u32::MAX),
+        display_depth,
         no_proc_macro,
     };
 
--- a/src/cargo/ops/tree/mod.rs
+++ b/src/cargo/ops/tree/mod.rs
@@ -43,8 +43,9 @@ pub struct TreeOptions {
     pub format: String,
     /// Includes features in the tree as separate nodes.
     pub graph_features: bool,
-    /// Maximum display depth of the dependency tree.
-    pub max_display_depth: u32,
+    /// Display depth of the dependency tree.
+    /// If non-negative integer, display dependencies with that amount of max depth.
+    pub display_depth: DisplayDepth,
     /// Excludes proc-macro dependencies.
     pub no_proc_macro: bool,
 }
@@ -82,6 +83,30 @@ impl FromStr for Prefix {
             "indent" => Ok(Prefix::Indent),
             "depth" => Ok(Prefix::Depth),
             _ => Err("invalid prefix"),
+        }
+    }
+}
+
+#[derive(Clone, Copy)]
+pub enum DisplayDepth {
+    MaxDisplayDepth(u32),
+}
+
+impl FromStr for DisplayDepth {
+    type Err = clap::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            s => s.parse().map(Self::MaxDisplayDepth).map_err(|_| {
+                clap::Error::raw(
+                    clap::error::ErrorKind::ValueValidation,
+                    format!(
+                        "supported values for --depth are non-negative integers, \
+                                but `{}` is unknown",
+                        s
+                    ),
+                )
+            }),
         }
     }
 }
@@ -250,7 +275,7 @@ fn print(
             pkgs_to_prune,
             opts.prefix,
             opts.no_dedupe,
-            opts.max_display_depth,
+            opts.display_depth,
             &mut visited_deps,
             &mut levels_continue,
             &mut print_stack,
@@ -270,7 +295,7 @@ fn print_node<'a>(
     pkgs_to_prune: &[PackageIdSpec],
     prefix: Prefix,
     no_dedupe: bool,
-    max_display_depth: u32,
+    display_depth: DisplayDepth,
     visited_deps: &mut HashSet<usize>,
     levels_continue: &mut Vec<bool>,
     print_stack: &mut Vec<usize>,
@@ -329,7 +354,7 @@ fn print_node<'a>(
             pkgs_to_prune,
             prefix,
             no_dedupe,
-            max_display_depth,
+            display_depth,
             visited_deps,
             levels_continue,
             print_stack,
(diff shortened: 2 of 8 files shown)


tests/testsuite/tree.rs
```diff
@@ -1846,6 +1846,61 @@ c v1.0.0
...
+
+            [dependencies]
+            somedep = "1"
+            otherdep = "1"
+            "#,
+        )
+        .file("c/src/lib.rs", "")
+        .build();
+
+    p.cargo("tree --depth workspace")
```
epage, line 1891: Could you split this PR into two commits
1. Add the test without the `--depth workspace` flag
2. Add support for `--depth workspace` and update the test to pass it in

With this, it'll make the intent of this change a lot clearer and help test the test

ShoyuVanilla: Sure!

src/cargo/ops/tree/mod.rs
```diff
@@ -341,15 +369,15 @@ fn print_node<'a>(
 
 /// Prints all the dependencies of a package for the given dependency kind.
 fn print_dependencies<'a>(
-    gctx: &GlobalContext,
+    ws: &Workspace<'_>,
```
epage, line 372: nit: splitting the change in parameter would shrink the focus of the main commit to make it easier to see whats happening

src/cargo/ops/tree/mod.rs
```diff
@@ -341,15 +369,15 @@ fn print_node<'a>(
...
     graph: &'a Graph<'_>,
     node_index: usize,
     format: &Pattern,
     symbols: &Symbols,
     pkgs_to_prune: &[PackageIdSpec],
     prefix: Prefix,
     no_dedupe: bool,
-    max_display_depth: u32,
+    display_depth: DisplayDepth,
```
epage, line 380: nit: switching to a new type in its own commit would also make it clearer

epage:
FYI I gave some suggestions that would make the review easier to follow.  The main one i'm interested in is the test one but if you are up for doing the others, that'd be great!

src/cargo/ops/tree/mod.rs
```diff
@@ -389,6 +422,11 @@ fn print_dependencies<'a>(
             // Filter out packages to prune.
             match graph.node(**dep) {
                 Node::Package { package_id, .. } => {
+                    if filter_non_workspace_member
+                        && !ws.is_member(graph.package_for_id(*package_id))
```
epage, line 426: nit: mind adding (in its own commit before this) a `is_member_id`?  We have to look up the package just for `Workspace` to use it to look up the id.  Seems like it'd be better if we called `is_member_id` here (and in other places)

src/bin/cargo/commands/tree.rs
```diff
@@ -42,7 +42,14 @@ pub fn cli() -> Command {
             "SPEC",
             "Prune the given package from the display of the dependency tree",
         ))
-        .arg(opt("depth", "Maximum display depth of the dependency tree").value_name("DEPTH"))
+        .arg(
+            opt(
+                "depth",
+                "Maximum display depth of the dependency tree, \
+                or --depth=worktree to display dependencies from workspace members only",
```
epage, line 49: We should also update `cargo help tree`, see https://github.com/rust-lang/cargo/blob/master/src/doc/man/cargo-tree.md

src/cargo/ops/tree/mod.rs
ShoyuVanilla: Sounds more convenient

src/bin/cargo/commands/tree.rs
epage, line 49: Like with `--jobs default`, I wonder if we should leave this for `cargo help tree`

`--jobs default` is less common but not thrilled with how long this has made the help.  There are things we could do to trim it some.

src/cargo/ops/tree/mod.rs
```diff
@@ -86,6 +86,32 @@ impl FromStr for Prefix {
     }
 }
 
+#[derive(Clone, Copy)]
+pub enum DisplayDepth {
+    MaxDisplayDepth(u32),
+    Workspace,
+}
```
epage, line 95: As this is dead code, please quash it into the commit that follows

It should also just be a newtype at this point as there won't be any workspace support yet

epage:
Thanks!  Splitting it up makes reviews much easier!

src/cargo/ops/tree/mod.rs
```diff
@@ -422,6 +422,11 @@ fn print_dependencies<'a>(
             // Filter out packages to prune.
             match graph.node(**dep) {
                 Node::Package { package_id, .. } => {
+                    if filter_non_workspace_member
+                        && !ws.is_member(graph.package_for_id(*package_id))
```
ShoyuVanilla, line 426: Oh, I forgot to use new-added function here 😅

src/cargo/ops/tree/mod.rs
epage: Thanks for squashing.  This should also be switched to a new type, only parsing numbers.

Every commit should be atomic and this commit accepts new user input and then ignores it.

epage: The adding of workspace support should be in the commit that reads the variable

tests/testsuite/tree.rs
```diff
@@ -1888,15 +1888,12 @@ fn depth_workspace() {
         .file("c/src/lib.rs", "")
         .build();
 
-    p.cargo("tree")
+    p.cargo("tree --depth workspace")
```
epage, line 1891: This should be squashed into the previous commit.

Every commit should be atomic which includes making sure the tests pass.

src/doc/man/cargo-tree.md
```diff
@@ -96,6 +96,9 @@ Prune the given package from the display of the dependency tree.
 {{#option "`--depth` _depth_" }}
 Maximum display depth of the dependency tree. A depth of 1 displays the direct
 dependencies, for example.
+
+If the given value is `workspace`, only shows the dependencies that are member
```
epage, line 100: nit: In most cases, I would also consider doc updates to be a part of the commit that changed the behavior being documented

src/cargo/ops/tree/mod.rs
ShoyuVanilla: Oh, I understood

tests/testsuite/tree.rs
ShoyuVanilla: Squashing would be better and I'm gonna do so, but I think that the test may pass as it is, because the output is not updated yet

src/cargo/ops/tree/mod.rs
ShoyuVanilla: It is being read - though not used - here in the same commit 😅 
https://github.com/rust-lang/cargo/pull/14928/commits/a66e96a2b6bcd81d16c461108fc47d2f41127e6d#diff-52fd799ef4f821832efb7407f8d6d8501d8067e08dfb6f189203e19942c30d1bR409-R413

epage:
Thank you for all of the clean up!

With this having been discussed previously and the scope being relatively small, I'm wondering if we need an FCP.  @weihanglo you second that?

epage (approved):

weihanglo:
Yes that seems reasonable. If something arises then we reconsider.

weihanglo (approved):

epage:
Thanks @ShoyuVanilla!

ShoyuVanilla:
> Thanks @ShoyuVanilla!

Thank you for the nice and detailed reviews!

Issue #14685, opened 2024-10-14
cargo package could run tests in the packaged tree


sourcefrog:
### Problem

Today, cargo package has an option (by default on) to run a build in the packaged tree, to increase confidence that the package will build when downloaded.

Some users would like to check that the tests also pass in the package crate. 

In particular, Debian and probably other distributions that repackage Rust software would like to:
1. start with a crate tarball, not a git checkout
2. run the tests to increase confidence that their build is correct

However in cargo-mutants we found that the tests didn't pass in a tree unpacked from a cargo crate, because some testdata files were excluded. (See https://sources.debian.org/patches/rust-cargo-mutants/23.10.0-1/disable-tests-missing-testdata.patch/, and cc @jelmer.)

Arguably there's also value in crates.io archiving a copy of the tree that's sufficiently complete that someone could resume development from that crate if the VCS repository was lost, and this will be much more feasible if the tests are known to pass. For widely cloned git repos this might seem unlikely, but there might be some crates where the released source is open but the VCS is not.

### Proposed Solution

If this is acceptable in principle I could send a PR that adds `cargo package --test` (and `cargo publish --test`) that will run the tests in a copy of the packaged tree.  The option would be off by default.

### Notes

One conceivable argument against this is that including all the tests and testdata to make the tests pass will make the package larger, which will increase bandwidth consumption for both crates.io and users. Probably a large majority of clients that download packages do not run or even build their tests, as they only want to build the library or binary.

Perhaps some crates have very large testdata directories that today are excluded, but adding this option might motivate their authors to include them in the package.

I think that might happen at the margin but I don't think it's a conclusive reason not to do it:
* There's already a cap on package size, and this would not change it.
* cargo and crates.io as far as I know doesn't make any attempt to strip packages down to only what is needed to build the binary/library: many crates will include docs and other files that aren't strictly needed for the build.
* Probably many packages do already include most or all of their tests and testdata; this would just help packages that use the option make sure that the shipped files are actually usable to achieve passing tests.
* I would guess that for many crates the integration tests and testdata are not going to greatly increase the size of the package.

epage:
> Arguably there's also value in crates.io archiving a copy of the tree that's sufficiently complete that someone could resume development from that crate if the VCS repository was lost, and this will be much more feasible if the tests are known to pass. For widely cloned git repos this might seem unlikely, but there might be some crates where the released source is open but the VCS is not.

[...]

PR #14930, opened 2024-12-12
perf(package): Speed up verify with 'check'

### LINKED ISSUE #14941: `cargo package` (and hence `cargo publish`) verify step is slow from doing a full build
In testing workspace publishing, I was really wishing for the build to be faster when I noticed the sea of status messages was "Compiling", instead of "Checking".
I've been glossing over those messages for years and never noticed!

If we switch from `cargo build` to `cargo check it would make builds faster by skipping the compiler backend / codegen at the cost of not getting post-monomorphization errors. That seems like a small price to pay.

epage:

In testing workspace publishing, I was really wishing for the build to be faster when I noticed the sea of status messages was "Compiling", instead of "Checking".
I've been glossing over those messages for years and never noticed!

This makes builds faster by skipping the compiler backend / codegen at the cost of not getting post-monomorphization errors. That seems like a small price to pay.

Fixes #14941

I searched through the issues, open and closed, and saw no previous discussions of this.  In particular, I was looking to see if this was previously rejected.
My only assumption is this was missed when `cargo check` was added.

### DIFF
--- a/src/cargo/ops/cargo_package/verify.rs
+++ b/src/cargo/ops/cargo_package/verify.rs
@@ -94,7 +94,7 @@ pub fn run_verify(
                 opts.jobs.clone(),
                 opts.keep_going,
                 &opts.targets,
-                CompileMode::Build,
+                CompileMode::Check { test: false },
             )?,
             cli_features: opts.cli_features.clone(),
             spec: ops::Packages::Packages(Vec::new()),
--- a/tests/testsuite/alt_registry.rs
+++ b/tests/testsuite/alt_registry.rs
@@ -351,8 +351,8 @@ See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for
 [VERIFYING] foo v0.0.1 ([ROOT]/foo)
 [DOWNLOADING] crates ...
 [DOWNLOADED] bar v0.0.1 (registry `alternative`)
-[COMPILING] bar v0.0.1 (registry `alternative`)
-[COMPILING] foo v0.0.1 ([ROOT]/foo/target/package/foo-0.0.1)
+[CHECKING] bar v0.0.1 (registry `alternative`)
+[CHECKING] foo v0.0.1 ([ROOT]/foo/target/package/foo-0.0.1)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 [UPLOADING] foo v0.0.1 ([ROOT]/foo)
 [UPLOADED] foo v0.0.1 to registry `alternative`
@@ -515,7 +515,7 @@ See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for
 [PACKAGING] foo v0.0.1 ([ROOT]/foo)
 [PACKAGED] 4 files, [FILE_SIZE]B ([FILE_SIZE]B compressed)
 [VERIFYING] foo v0.0.1 ([ROOT]/foo)
-[COMPILING] foo v0.0.1 ([ROOT]/foo/target/package/foo-0.0.1)
+[CHECKING] foo v0.0.1 ([ROOT]/foo/target/package/foo-0.0.1)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 [UPLOADING] foo v0.0.1 ([ROOT]/foo)
 [UPLOADED] foo v0.0.1 to registry `alternative`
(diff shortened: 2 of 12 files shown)


src/cargo/ops/cargo_package.rs
```diff
@@ -1157,7 +1157,7 @@ fn run_verify(
                 opts.jobs.clone(),
                 opts.keep_going,
                 &opts.targets,
-                CompileMode::Build,
+                CompileMode::Check { test: false },
```
weihanglo, line 1160: For myself I would keep it a full build.
One reason is that `cargo publish` is not an operation easy to revert.

The other is, per [RFC 3477](https://github.com/rust-lang/rfcs/blob/master/text/3477-cargo-check-lang-policy.md),

> A Rust program must compile with cargo build to be covered by Rust's standard stability guarantee.

If we want the same level of stability guarantee for published crates, we'd better do a full build.

epage: I guess the question is what is it we intend to be verified and at what cost.

Things that can be verified
- Basic check
- Post-monomorphization errors
- feature combinations
- platform combinations
- logic (ie `cargo test`, #14685)
- Miri for verifying `unsafe`

In commonly used crates, there is most likely a CI checking for what the author feels is important enough.  I don't think `cargo publish`, especially be default, should or can replicate that.

What verify can help with is helping specifically with packaging specific issues
- Are all the right files present
- Can this build in isolation (to a degree)

For those, `check` is sufficient.

For people who don't have a CI, verify can tell them some but not much.  Personally, I don't think we need to be verifying the stability guarantees in these cases, especially at the cost of expensive packaging, which has the most impact when doing it in dry-run mode which is where I most notice this (and tend to avoid dry run because of it).

weihanglo: On one hand, commonly used crates usually tend to have better CI checking support, so less likely to hit errors like this. Yet it depends on where they run `cargo build` or `cargo test` under release profile.

On the other hand, we are getting more API in const contexts (81 in 1.83). While `const fn` doesn't always contribute to error during [monomorphization](https://github.com/rust-lang/rust/issues/99682), it still increases the chances of hitting it.
(Check the minimal example in <https://github.com/rust-lang/rust/issues/112301> to understand how to get bitten).

There are also [diagnostics during MIR passes](https://github.com/rust-lang/rust/issues/49292) cannot be caught by `cargo check`. By just looking at how easy [the example](https://github.com/rust-lang/rust/issues/49292#issuecomment-1691455732) is. I am afraid of this may be more troublesome than monomorphization errors.

> I searched through the issues, open and closed, and saw no previous discussions of this.

Supposedly this is a starting point of examining all verification Cargo provides and re-position cargo commands, to embrace the potential plumbing commands reorganization (https://github.com/rust-lang/rust-project-goals/pull/178/ perhaps). However, instead of merging into this change now, perhaps we could create an issue and solicit feedback first?

weihanglo: By all means, I love how you’re thinking ahead with all these verification categories.

epage: > Yet it depends on where they run cargo build or cargo test under release profile.

Good point that there is also `--release`

> While const fn doesn't always contribute to error during https://github.com/rust-lang/rust/issues/99682, it still increases the chances of hitting it.

> There are also https://github.com/rust-lang/rust/issues/49292 cannot be caught by cargo check

If post-`check` errors are getting that bad, then I have strong words for T-lang...  The RFC test focuses on linker errors and doesn't justify its existence with it being prevalent.

epage: Created #14941

joshtriplett:
I don't personally think this should be the default, because `cargo build` does catch things `cargo check` does not, per [RFC 3477](https://github.com/rust-lang/rfcs/blob/master/text/3477-cargo-check-lang-policy.md) as mentioned.

If people want this, for performance, it might make sense as an option people could choose to use. I personally don't think that option should be the default, but zero objections to an option if someone wants it.

ranger-ross:
Just pitching in my 2 cents. 
I also tend to lean towards keeping `cargo build` as part `cargo publish` for the additional verification. 

I find that most crates publish new version once or twice a week at most. 
So to me the tradeoff is saving of one or two builds a week vs keeping the additional validation.
With that framing, I think the extra time building are worth the additional protection against publishing a broken library.

Although I do think having the ability skip the build during publishing would be useful for projects that are large and publish frequently. (ie. internal business logic libraries that change frequently)
So having something like `cargo publish --skip-build` would make sense to me :+1:

weihanglo:
Going to close this. The discussion has been captured in #14941. Can always revisit when we make the call. Thank you!

PR #14931, opened 2024-12-13
fix(base): Support bases in patches in virtual manifests 

epage:

This bug has been there since #14360

Found this when looking to change the normalization code for cargo script.  As a result, I also changed `cargo-util-schemas` in the hope that grouping the fields would make the intent clearer.  Since I was already changing field order, I also re-ordered for my personal taste (user facing features first)

### DIFF
--- a/crates/cargo-util-schemas/src/manifest/mod.rs
+++ b/crates/cargo-util-schemas/src/manifest/mod.rs
@@ -35,11 +35,13 @@ use crate::schema::TomlValueWrapper;
 #[serde(rename_all = "kebab-case")]
 #[cfg_attr(feature = "unstable-schema", derive(schemars::JsonSchema))]
 pub struct TomlManifest {
-    // when adding new fields, be sure to check whether `requires_package` should disallow them
     pub cargo_features: Option<Vec<String>>,
+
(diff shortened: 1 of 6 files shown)


crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -35,11 +35,13 @@ use crate::schema::TomlValueWrapper;
...
 #[cfg_attr(feature = "unstable-schema", derive(schemars::JsonSchema))]
 pub struct TomlManifest {
-    // when adding new fields, be sure to check whether `requires_package` should disallow them
     pub cargo_features: Option<Vec<String>>,
+
+    // Update `requires_package` when adding new package-specific fields
     pub package: Option<Box<TomlPackage>>,
     pub project: Option<Box<TomlPackage>>,
-    pub profile: Option<TomlProfiles>,
+    pub badges: Option<BTreeMap<String, BTreeMap<String, String>>>,
```
weihanglo, line 43: I wish we could have a more automated way to group these. Having a separate strict is too overkill. Anyway reorder looks fine!

weihanglo (approved):

PR #14943, opened 2024-12-16
test(build-std): Isolate output test to avoid spurious `[BLOCKING]` messages from concurrent runs

weihanglo:

47c2095b1dd580a91e42cb6197b58a318526b8c4 didn't really fix the flakiness.

Spun off from <https://github.com/rust-lang/cargo/pull/14938>  2a54190c9c789756d2281eefc856704e98c9941e

build-std tests use the users `CARGO_HOME` for downloading registry dependencies of the standard library.
This reduces disk needs of the tests, speeds up the tests, and reduces the number of network requests that could fail.

However, this means all of the tests access the same locks for the package cache.  In one test, we assert on the output and a `[BLOCKING]` message can show up, depending on test execution time from concurrent test runs.

We are going to hack around this by having the one test that asserts on test output to use the standard `cargo-test-support` `CARGO_HOME`, rather than the users `CARGO_HOME`.  There will then only be one process accessing the lock and no `[BLOCKING]` messages.

No more assertion errors like this:

```
---- test_proc_macro stdout ----
running `/home/runner/work/cargo/cargo/target/debug/cargo fetch -Zbuild-std -Zpublic-dependency`
running `/home/runner/work/cargo/cargo/target/debug/cargo test --lib -Zbuild-std -Zpublic-dependency`
thread 'test_proc_macro' panicked at tests/build-std/main.rs:394:10:

---- expected: tests/build-std/main.rs:388:27
++++ actual:   stderr
        1 + [BLOCKING] waiting for file lock on package cache
error: test failed, to rerun pass `-p cargo --test build-std`
        2 + [BLOCKING] waiting for file lock on package cache
   1    3 | [COMPILING] foo v0.0.0 ([ROOT]/foo)
   2    4 | [FINISHED] `test` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
   3    5 | [RUNNING] unittests src/lib.rs (target/debug/deps/foo-[HASH])
```

### DIFF
--- a/tests/build-std/main.rs
+++ b/tests/build-std/main.rs
@@ -25,9 +25,11 @@ use cargo_test_support::{basic_manifest, paths, project, rustc_host, str, Execs}
 use std::env;
 use std::path::Path;
 
-fn enable_build_std(e: &mut Execs, arg: Option<&str>) {
-    e.env_remove("CARGO_HOME");
-    e.env_remove("HOME");
+fn enable_build_std(e: &mut Execs, arg: Option<&str>, isolated: bool) {
+    if !isolated {
+        e.env_remove("CARGO_HOME");
+        e.env_remove("HOME");
+    }
 
     // And finally actually enable `build-std` for now
     let arg = match arg {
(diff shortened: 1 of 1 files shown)


Muscraft (approved):
Thanks!

tests/build-std/main.rs
```diff
@@ -42,17 +44,23 @@ fn enable_build_std(e: &mut Execs, arg: Option<&str>) {
 trait BuildStd: Sized {
     fn build_std(&mut self) -> &mut Self;
     fn build_std_arg(&mut self, arg: &str) -> &mut Self;
+    fn build_std_isolated(&mut self) -> &mut Self;
```
epage, line 47: Can we document when to use each?

tests/build-std/main.rs
```diff
@@ -107,9 +115,12 @@ fn basic() {
         )
         .build();
 
-    p.cargo("check").build_std().target_host().run();
+    // An isolated CARGO_HOME environment is used elesewhere in this test
+    // to ensure no extra index updates is performed.
+    // It is achieve by asserting the complete stderr without any wildcard.
```
epage, line 120:
```suggestion
    // HACK: use an isolated the isolated CARGO_HOME environment (`build_std_isolated`) to avoid `[BLOCKING]` messages (from lock contention with other tests) from getting in this test's asserts
```

epage (approved):

Issue #13586, opened 2024-03-14
Not reproducible when a project is copied to another directory


cathy-sjh:
The Compiled Binary Files Are Inconsistent After the Directory of the Rust Package Is Changed.
It is important to have [reproducible builds](https://en.wikipedia.org/wiki/Reproducible_builds).
Let's say I have two crate in my project, crate mangle2 depends on crate mangle1 and they are located in directory1.
```
root -> /usr1/directory1 $  tree 
.
├── mangle1
│   ├── Cargo.toml
│   └── src
│       └── lib.rs
└── mangle2
    ├── Cargo.toml
    └── src
        └── lib.rs
```
mangle1/Cargo.toml file:
```
[package]
name = "mangle1"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
```
mangle1/src/lib.rs file:
```
pub fn sjhadd(left: usize, right: usize) -> usize {
    left + right
}
```

in mangle2/Cargo.toml file, we see mangle2 depends on mangle1.
```
[package]
name = "mangle2"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
mangle1 = { path = "../mangle1" }

```
in mangle2/src/lib.rs file
```
use mangle1::sjhadd;

#[inline(never)]
pub fn super_add(left: usize, right: usize) -> usize {
    sjhadd(left, right)
}
```
Then I went into the mangle2 directory and compiled the whole project.
```
root -> /usr1/directory1/mangle2 $ cargo build --release -v
   Compiling mangle1 v0.1.0 (/usr1/s00659936/rust_program/mangle_disambiguator_test/test2/mangle1)
[...]

PR #14951, opened 2024-12-17
Do not hash absolute sysroot path into stdlib crates metadata.

Dirbaio:
stdlib crates get a SourceId which has an absolute path pointing into the sysroot. This makes the metadata hash change depending on where you've installed Rust. This is causing problems because the different hashes snowball into the optimizer making different decisions which ends up changing the binary size. 

(Some context: at work we're working with embedded devices with little flash storage so it often happens that a binary builds locally and then fails to fit in flash in CI, just because CI has installed rustc to a different path. Improving binary size is *not* a goal of this PR, after the fix the size will be whatever, but at least it won't change based on the rustc path anymore)

Overview of the fix:
- For libstd crates, the metadata hash now contains the path relative to the sysroot, instead of the absolute path.
- The absolute path is still hashed into the fingerprint (not the metadata) so moving the rustc installation triggers a rebuild. This ensures stdlib crates are rebuilt when upgrading nightly versions.
- The rustc version is still hashed into the metadata as usual, so upgrading Rust releases (not nightly versions) does cause a metadata change.

Repro of the bug:

```
$ git clone https://github.com/embassy-rs/embassy --branch cargo-nondet-repro
$ cd embassy/
$ cd examples/nrf52840
$ RUSTUP_HOME=~/.rustup1 cargo build --release --bin wifi_esp_hosted
....
    Finished `release` profile [optimized + debuginfo] target(s) in 13.33s
$ llvm-size target/thumbv7em-none-eabi/release/wifi_esp_hosted
   text	   data	    bss	    dec	    hex	filename
 114500	     80	  48116	 162696	  27b88	target/thumbv7em-none-eabi/release/wifi_esp_hosted
$ RUSTUP_HOME=~/.rustup2 cargo build --release --bin wifi_esp_hosted
....
    Finished `release` profile [optimized + debuginfo] target(s) in 9.64s
$ llvm-size target/humbv7em-none-eabi/release/wifi_esp_hosted
   text	   data	    bss	    dec	    hex	filename
 114272	     80	  48116	 162468	  27aa4	target/thumbv7em-none-eabi/release/wifi_esp_hosted
```

### DIFF
--- a/src/cargo/core/compiler/build_runner/compilation_files.rs
+++ b/src/cargo/core/compiler/build_runner/compilation_files.rs
@@ -604,11 +604,20 @@ fn compute_metadata(
 
     METADATA_VERSION.hash(&mut shared_hasher);
 
+    let ws_root = if unit.is_std {
+        // SourceId for stdlib crates is an absolute path inside the sysroot.
+        // Pass the sysroot as workspace root so that we hash a relative path.
+        // This avoids the metadata hash changing depending on where the user installed rustc.
+        &bcx.target_data.get_info(unit.kind).unwrap().sysroot
+    } else {
+        bcx.ws.root()
+    };
+
     // Unique metadata per (name, source, version) triple. This'll allow us
     // to pull crates from anywhere without worrying about conflicts.
     unit.pkg
         .package_id()
-        .stable_hash(bcx.ws.root())
+        .stable_hash(ws_root)
         .hash(&mut shared_hasher);
 
     // Also mix in enabled features to our metadata. This'll ensure that


Dirbaio:
An alternative approach I tried is making `stable_hash()` strip the sysroot prefix, just like it does now with the workspace root. Diff is [here](https://github.com/Dirbaio/cargo/commit/b641d1ae022575a1db7efa8401c3670132d9d89a). It felt a bit less clean to me because it requires threading the sysroot through a few places, but OTOH it doesn't require special-casing `is_std`. If that approach seems preferable let me know.

src/cargo/core/compiler/build_runner/compilation_files.rs
```diff
@@ -604,12 +604,18 @@ fn compute_metadata(
...
+    if unit.is_std {
+        // SourceId for build-std crates is a full absolute path inside the sysroot.
+        // Don't hash it to avoid the metadata hash changing depending on where the user installed rustc.
+        unit.pkg.name().hash(&mut shared_hasher);
+    } else {
+        // Unique metadata per (name, source, version) triple. This'll allow us
+        // to pull crates from anywhere without worrying about conflicts.
+        unit.pkg
+            .package_id()
+            .stable_hash(bcx.ws.root())
```
weihanglo, line 616: Another approach to this: if a unit is from `-Zbuild-std`, pass its workspace root path to `stable_hash` instead.

weihanglo (commented):
I don't have a definitive answer here. Overall the more reproducibility the better. OTOH, it is also a niche use case like [this one](https://github.com/rust-lang/cargo/issues/13586#issuecomment-2550024395), and relies on how the unstable `-Zbuild-std` work underneath.

src/cargo/core/compiler/build_runner/compilation_files.rs
```diff
@@ -604,12 +604,18 @@ fn compute_metadata(
...
-    // Unique metadata per (name, source, version) triple. This'll allow us
-    // to pull crates from anywhere without worrying about conflicts.
-    unit.pkg
-        .package_id()
-        .stable_hash(bcx.ws.root())
-        .hash(&mut shared_hasher);
+    if unit.is_std {
+        // SourceId for build-std crates is a full absolute path inside the sysroot.
+        // Don't hash it to avoid the metadata hash changing depending on where the user installed rustc.
+        unit.pkg.name().hash(&mut shared_hasher);
```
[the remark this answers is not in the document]
Dirbaio: good point, I didn't realize `is_std` also applied to crates.io deps of stdlib crates.

I've changed it to pass the sysroot as workspace root for libstd crates instead, as per your other comment.

src/cargo/core/compiler/build_runner/compilation_files.rs
Dirbaio: that's cleaner yes! I've changed the code to do this.

weihanglo: Sorry but I got paranoid that this could be applied to the same logic, for consistency.

https://github.com/rust-lang/cargo/blob/99dff6d77db779716dda9ca3b29c26addd02c1be/src/cargo/core/compiler/build_runner/compilation_files.rs#L203-L206

(It doesn't really matter though. The hash is for target directory layout for fingerprint and build-script output…)

weihanglo: Never mind. It needs more tweaks and seems meaningless.

weihanglo (approved):
@Dirbaio 
I am about to merge this. Could you update the PR description a bit to reflect the latest change? Thank you!

src/cargo/core/compiler/build_runner/compilation_files.rs
```diff
@@ -604,11 +604,20 @@ fn compute_metadata(
 
     METADATA_VERSION.hash(&mut shared_hasher);
 
+    let ws_root = if unit.is_std {
+        // SourceId for stdlib crates is an absolute path inside the sysroot.
+        // Pass the sysroot as workspace root so that we hash a relative path.
+        // This avoids the metadata hash changing depending on where the user installed rustc.
+        &bcx.target_data.get_info(unit.kind).unwrap().sysroot
```
weihanglo, line 611: Just a note for future self: sysroot is something like
`/home/user/.rustup/toolchains/nightly-aarch64-apple-darwin`

It is not the real workspace root path of the standard library.
While this seems a bit out of sync, the relative path inside sysroot should be fairly deterministic. It's fine now and I don't think we need any hardcoded path logic like [this](https://github.com/rust-lang/cargo/blob/addcc8ca715bc7fe20df66afd6efbf3c77ef43f8/src/cargo/core/compiler/standard_lib.rs?plain=1#L217).

Dirbaio:
> @weihanglo  Could you update the PR description a bit to reflect the latest change? Thank you!

 done!

PR #14957, opened 2024-12-18
fixed the error message for a user to open the crate

### LINKED ISSUE #14776: `cargo doc --open` error message is confusing when all targets are skipped
### Problem

I use an alias for this command:
```console
$ cargo doc -v --open --workspace --document-private-items --all-features --no-deps --examples
```
When first developing a new project, there are no examples yet.  The output is:
```
warning: target filter `examples` specified, but no targets matched; this is a no-op
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
error: no crates with documentation
```

So it is NOT a "no-op".  I believe it should be.  The warning alone is fine.

### Proposed Solution

Instead of failing to provide crate documentation, it should create crate documentation even if there's no examples yet.

### Notes

_No response_

ibilalkayy:

This PR is resolving an [issue](https://github.com/rust-lang/cargo/issues/14776#issuecomment-2515017963) where `cargo doc --open --examples` does not give a clear message when the examples file is given to it. 

I changed the message from `no crates with documentation` in the [cargo_doc.rs](https://github.com/rust-lang/cargo/blob/58b2d609ece6bd0333e4d0f63a024fe3dd350b4b/src/cargo/ops/cargo_doc.rs#L62) and [doc.rs](https://github.com/rust-lang/cargo/blob/58b2d609ece6bd0333e4d0f63a024fe3dd350b4b/tests/testsuite/doc.rs#L1504) to `requested crate documentation is not available to open`. Now it becomes easy for the user to understand what they're missing.

 Fixes #14776

Here is the command through which you can test this change: 

`cargo test SNAPSHOTS=overwrite -- doc::open_no_doc_crate`

### DIFF
--- a/src/cargo/ops/cargo_doc.rs
+++ b/src/cargo/ops/cargo_doc.rs
@@ -59,7 +59,7 @@ pub fn doc(ws: &Workspace<'_>, options: &DocOptions) -> CargoResult<()> {
         let name = &compilation
             .root_crate_names
             .get(0)
-            .ok_or_else(|| anyhow::anyhow!("no crates with documentation"))?;
+            .ok_or_else(|| anyhow::anyhow!("requested crate documentation is not available to open"))?;
         let kind = options.compile_opts.build_config.single_requested_kind()?;
 
         let path = path_by_output_format(&compilation, &kind, &name, &options.output_format);
--- a/src/cargo/ops/cargo_doc.rs
+++ b/src/cargo/ops/cargo_doc.rs
@@ -57,7 +57,9 @@ pub fn doc(ws: &Workspace<'_>, options: &DocOptions) -> CargoResult<()> {
 
     if options.open_result {
         let name = &compilation.root_crate_names.get(0).ok_or_else(|| {
-            anyhow::anyhow!("cannot open specified crate's documentation: no documentation generated")
+            anyhow::anyhow!(
+                "cannot open specified crate's documentation: no documentation generated"
+            )
         })?;
         let kind = options.compile_opts.build_config.single_requested_kind()?;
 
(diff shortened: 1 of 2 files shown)


src/cargo/ops/cargo_doc.rs
```diff
@@ -59,7 +59,7 @@ pub fn doc(ws: &Workspace<'_>, options: &DocOptions) -> CargoResult<()> {
         let name = &compilation
             .root_crate_names
             .get(0)
-            .ok_or_else(|| anyhow::anyhow!("no crates with documentation"))?;
+            .ok_or_else(|| anyhow::anyhow!("requested crate documentation is not available to open"))?;
```
epage, line 62: The hard part of this is fix is in the nuance.  By saying its "not available" implies it is there but something prevents it.

What about
```suggestion
            .ok_or_else(|| anyhow::anyhow!("cannot open specified crate's documentation: no documentation generated"))?;
```

ibilalkayy: What if the examples file is not even present? How it can open? Will this message still be suitable?

Rustin170506:
@ibilalkayy The fmt check failed. You need to run `cargo fmt --all`.

ibilalkayy:
Thank you, @Rustin170506! I was not aware of this. Once I get the clarification message from @epage, I will push again.

src/cargo/ops/cargo_doc.rs
epage: To fix the problem, it needs to discuss opening as that is the operation that is failing.

ibilalkayy: Alright, Now I understand the point.

epage:
Would you be ok cleaning up the commit history?

Issue #14831, opened 2024-11-16
Add `cargo pkgid` for cargo-script


Rustin170506:
### Problem

ref https://github.com/rust-lang/cargo/issues/12207#issuecomment-2358818671

For now, if you try to use `cargo pkgid` against to a cargo script file. It will generate an error:
```shell
cargo pkgid sql2.rs
error: could not find `Cargo.toml` in `/Users/rustin/code/rs-scripts` or any parent directory
```

### Proposed Solution

We need to support `cargo pkgid` for cargo-script. 

For instance, for itself, it would look like:
```shell
$pwd
/Users/rustin/code/rs-scripts

$ls
sql.rs  sql1.rs sql2.rs sql3.rs sql4.rs sql5.rs sql6.rs sql7.rs sql8.rs sql9.rs

$cargo pkgid sql2.rs
path+file:///Users/rustin/code/rs-scripts/sql2.rs#embedded
```

For its dependencies, it would look like:
```shell
cargo pkgid sql2.rs -p clap

https://github.com/rust-lang/crates.io-index#clap:4.5.21
```

### Notes

Not sure specifying the rust file as an argument is a good idea. Because `cargo pkgid` itself already takes the `SPEC` as the argument. So it might be a very confusing syntax.

weihanglo:
I've tried thinking it hard but not sure about the use case of getting `pkgid` from a cargo script. Mind sharing more on your thought?

weihanglo:
Regarding that, I feel like this should not block the stabilization. Unless there is a clear use case we want to support, I'd suggest we make sure pkgid not leaking when stabilization.
Adding an explicit unsupported statement is also helpful, since we can mark it as a bug if we any unexpected stuff is exposed.

Rustin170506:
> Regarding that, I feel like this should not block the stabilization. Unless there is a clear use case we want to support, I'd suggest we make sure pkgid not leaking when stabilization.

I think a reasonable use case is that we don't include the `cargo.lock` file under the script file dir. So there is no good way to know which version your dependency is using. So can we just get it with this command and use that specification?

epage:
```console
$ cargo pkgid sql2.rs
path+file:///Users/rustin/code/rs-scripts/sql2.rs#embedded
```
My hope is we can get away without a `#` or `?` for this.  It puts it completely on the `Source` to then disambiguate whether `sql2.rs` is a file or a directory.

```console
$ cargo pkgid sql2.rs -p clap

[...]

PR #14961, opened 2024-12-19
NOT READY FOR REVIEW : feat: add `cargo pkgid` support for cargo-script

Rustin170506:

In this PR, we added the `cargo pkgid` support for the cargo-script.

For the package itself:

```console
cargo pkgid --manifest-path foo.rs
path+file:///my-project/foo.rs#0.86.0
```

For its dependence:

```console
cargo pkgid --manifest-path foo.rs -p sqlx
registry+https://github.com/rust-lang/crates.io-index#sqlx@0.7.4
```

I have updated the unit tests and also added more test cases for it.

### Additional information

None


epage:
Another test we'll need is that we error when a cargo script is specified as a dependency which might become possible with this change.

Rustin170506:
> Another test we'll need is that we error when a cargo script is specified as a dependency which might become possible with this change.

Do you mean by `cargo pkgid --manifest-path foo.rs -p foo`?

epage:
I mean
```toml
[package]
name = "foo"

[dependencies]
bar.path = "bar.rs"
```

Rustin170506:
@epage [`dedc674` (#14961)](https://github.com/rust-lang/cargo/pull/14961/commits/dedc674f48caa515e10050b1ffb15495a1dbc78c) 

I just did a really rough commit to prove the idea here. It seems to work. But the code is totally messed up. 

I want to check two things with you before I move forward and improve the code. 
1. Can you take a grlance at the implementation, I think we have to change the SourceId and PackageIdSpec. I just want to check if I am on the right track.
2. I am a little hesitant that should we include the filename or just mark it is embedded. Since from the RFC, we don't allow users to specify the name at all, it seems the information is enough to construct the filename. But I also believe that including the full name would be easier to use and expand in the future. So I am not sure what is your thought.

epage:
> I am a little hesitant that should we include the filename or just mark it is embedded. Since from the RFC, we don't allow users to specify the name at all, it seems the information is enough to construct the filename. B

The RFC says `package.name` is defaulted but users can set it.

`package.name` cannot be mapped unambiguously back to a file name because we sanitize it
https://github.com/rust-lang/cargo/blob/0499e318a6b804c9b361605c04c3b744596de26c/src/cargo/util/toml/embedded.rs#L115
https://github.com/rust-lang/cargo/blob/0499e318a6b804c9b361605c04c3b744596de26c/src/cargo/util/toml/embedded.rs#L147-L176

epage:
> Can you take a grlance at the implementation, I think we have to change the SourceId and PackageIdSpec. I just want to check if I am on the right track.

```rust
    // FIXME: It should be `path+[ROOTURL]/foo#script.rs@0.0.0`.
    p.cargo("-Zscript pkgid --manifest-path script.rs")
        .masquerade_as_nightly_cargo(&["script"])
        .with_stdout_data(str![[r#"
path+[ROOTURL]/foo#script@0.0.0
"#]])
```
this should instead be
```rust
    // FIXME: It should be `path+[ROOTURL]/foo/script.rs#0.0.0`.
    p.cargo("-Zscript pkgid --manifest-path script.rs")
        .masquerade_as_nightly_cargo(&["script"])
        .with_stdout_data(str![[r#"
path+[ROOTURL]/foo/script.rs#0.0.0
"#]])
```
according to https://github.com/rust-lang/cargo/issues/14831#issuecomment-2539177293

Switching to that scheme should simplify the code, not requiring updates to `PackageIdSpec`, `PackageId`, or `SourceId` struct fields, only functions

Rustin170506:
> The RFC says `package.name` is defaulted but users can set it.

Oh, I misunderstood `package.name = <slugified file stem>`. I thought it always used the file name.



> `package.name` cannot be mapped unambiguously back to a file name because we sanitize it

Yes. Got it.

Rustin170506:
> this should instead be
> 
> ```rust
>     // FIXME: It should be `path+[ROOTURL]/foo/script.rs#0.0.0`.
>     p.cargo("-Zscript pkgid --manifest-path script.rs")
>         .masquerade_as_nightly_cargo(&["script"])
>         .with_stdout_data(str![[r#"
> path+[ROOTURL]/foo/script.rs#0.0.0
> "#]])
> ```
> 
> according to [#14831 (comment)](https://github.com/rust-lang/cargo/issues/14831#issuecomment-2539177293)
> 
> Switching to that scheme should simplify the code, not requiring updates to `PackageIdSpec`, `PackageId`, or `SourceId` struct fields, only functions


So you mean we don't include the package name in the spec at all? Always just show the filename?

epage:
> So I guess if the package name is foo, then we have:
> 
> ```rust
> path+[ROOTURL]/foo/script.rs#foo@0.0.0
> ```

If the user explicitly overrode it, yes

src/cargo/core/workspace.rs
```diff
@@ -1800,7 +1800,10 @@ impl<'gctx> Packages<'gctx> {
         match self.packages.entry(key.to_path_buf()) {
```
epage, line 1800: Looks like this will be ambiguous when we add workspace support and will be hard to find.  By working around this now, we are laying a trap for ourselves in the future.

Maybe we should have a refactor commit before this one that changes the `key` from being the manifest-dir to being the manifest-path.

Rustin170506: Yes. Totally argee this is a bit tricky. I will try to refactor it and change the key.

tests/testsuite/update.rs
```diff
@@ -1886,37 +1880,7 @@ fn update_breaking() {
...
-[UPDATING] multiple-registries v2.0.0 (registry `alternative`) -> v3.0.0
-[UPDATING] multiple-registries v1.0.0 -> v2.0.0
-[UPDATING] multiple-source-types v1.0.0 -> v2.0.0
-[ADDING] multiple-versions v3.0.0
-[UPDATING] platform-specific v1.0.0 -> v2.0.0
-[UPDATING] shared v1.0.0 -> v2.0.0
-[UPDATING] ws v1.0.0 -> v2.0.0
-
-"#]])
+        .with_stderr_data(str![""])
```
Rustin170506, line 1883: I'm not sure why this has been updated. I will revert it.

tests/testsuite/script.rs
```diff
@@ -1370,6 +1370,47 @@ registry+https://github.com/rust-lang/crates.io-index#dep@1.0.0
...
+        .build();
+
+    p.cargo("build")
+        .masquerade_as_nightly_cargo(&["script"])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[WARNING] no edition set: defaulting to the 2015 edition while the latest is 2024
+[ERROR] failed to get `script` as a dependency of package `foo v0.1.0 ([ROOT]/foo)`
+
+Caused by:
```
Rustin170506, line 1398: Working on a better error message for it.

src/cargo/sources/path.rs
```diff
@@ -124,6 +124,9 @@ impl<'gctx> PathSource<'gctx> {
     }
 
     fn read_package(&self) -> CargoResult<Package> {
+        if crate::util::toml::is_embedded(&self.path) {
```
Rustin170506, line 127: I am not sure if this is the best place to check this. Maybe we should move it to where we create the source ID?

Rustin170506 (commented):

src/cargo/core/workspace.rs
```diff
@@ -271,7 +271,9 @@ impl<'gctx> Workspace<'gctx> {
         ws.require_optional_deps = require_optional_deps;
         let id = package.package_id();
         let package = MaybePackage::Package(package);
-        ws.packages.packages.insert(ws.current_manifest.clone(), package);
+        ws.packages
+            .packages
+            .insert(ws.current_manifest.clone(), package);
         ws.target_dir = if let Some(dir) = target_dir {
```
epage, line 277: Does this reformatting belong on the previous commit?

tests/testsuite/script.rs
```diff
@@ -1370,6 +1370,47 @@ registry+https://github.com/rust-lang/crates.io-index#dep@1.0.0
         .run();
 }
 
+#[cargo_test(nightly, reason = "edition2024 hasn't hit stable yet")]
+fn script_as_dep() {
```
epage, line 1374: Is there a reason this test wasn't added before the feature commit?

src/cargo/sources/path.rs
epage: imo this feels too spooky as this function is not directly tied to dependencies.

tests/testsuite/script.rs
Rustin170506: Just want to make sure it already has an error for it. The feature is just to improve the error message.

src/cargo/core/workspace.rs
Rustin170506: Oops. I squashed it into the wrong commit.

src/cargo/sources/path.rs
Rustin170506: Then I guess we should do the check when we read the manifest file.

epage: Maybe it isn't so bad? If we view the lack of dependency / workspace support as not far away, this just needs to hold us over until then. If its far away or unlikely, then more care might be needed.

epage: Dependency support is dependent on either
- `[lib]` support which is hazy
- artifact dependencies which I'd really like to move forward for build script delegation

Workspace support could be really useful for cargo-xtasks

These were deferred out of the first RFC just to focus on the minimal, core feature set.  I am not lined up to immediately follow with these features but I could see them being high demand when this gets stabilized.

Rustin170506: I have moved it before we create the source ID for the path. I think this location is better than the previous one (read package). It also helped us bail out an error earlier.

Rustin170506 (commented):

src/cargo/core/source_id.rs
```diff
@@ -393,6 +393,9 @@ impl SourceId {
                     .url
                     .to_file_path()
                     .expect("path sources cannot be remote");
+                if crate::util::toml::is_embedded(&path) {
+                    anyhow::bail!("The embedded script cannot be used as a dependency");
```
epage, line 397: "embedded script" isn't quite the right term to use for this.

We refer to the manifest as "embedded" because it is encapsulated within the script.  We have unofficially been referring to the files as scripts but I'm not fully satisfied with that term.  In the RFC, I also called these "single file packages".

src/cargo/core/source_id.rs
```diff
@@ -393,6 +393,9 @@ impl SourceId {
                     .url
                     .to_file_path()
                     .expect("path sources cannot be remote");
+                if crate::util::toml::is_embedded(&path) {
```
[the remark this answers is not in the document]
Rustin170506: No, I think we only support the directory path now. So it wouldn't reach here.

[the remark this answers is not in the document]
Rustin170506: I will add a test case for it.

src/cargo/core/source_id.rs
Rustin170506: Updated.

src/cargo/core/source_id.rs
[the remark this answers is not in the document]
Rustin170506: Added in [`9fe45b2` (#14961)](https://github.com/rust-lang/cargo/pull/14961/commits/9fe45b20100536ba2efb41c57e9866acda4d7682)

epage (approved):

epage:
Thanks!

Rustin170506:

Issue #14955, opened 2024-12-18
Performance Regression in `cargo package` on `1.81.0`


landonxjames:
### Problem

The aws-sdk-rust recently updated our MSRV to `1.81.0`. This caused a 3-4x slowdown in our `cargo package` invocation. We traced the likely culprit of this slowdown to https://github.com/rust-lang/cargo/pull/13960 which removes an `if !opts.allow_dirty` check around the `check_repo_state` function causing it to run on ever packaging step.

This causes a problem with the [aws-sdk-rust](https://github.com/awslabs/aws-sdk-rust) repo. Since it is very large and has a huge history all invocations to `git` in that repository are slow. Since `cargo package` is now invoking `git` on every packaging step in the workspace it slows the whole process down to a crawl. 

### Steps

1. Run `git clone https://github.com/awslabs/aws-sdk-rust.git` to checkout the `aws-sdk-rust` repo (this will be a bit slow, it is huge)
2. Ensure that you are using a cargo version `<1.81`
3. In the `aws-sdk-rust` repo run `cargo package --no-verify --allow-dirty --workspace` observe the approximate pace at which crates are packaged (you can also use the `time` command, but this fails locally for me with `Too many open files (os error 24)` before it completes)
4. Upgrade your cargo version to `=1.81`
5. Again run `time cargo package --no-verify --allow-dirty --workspace` (this goes too slowly for me to wait for it to fail, but it is very easy to observe the difference in speed)

### Possible Solution(s)

Potentially when packaging multiple crates it might be possible to only invoke `git` once at the beginning of the run? Or potentially add a new option to disable the behavior introduced in  https://github.com/rust-lang/cargo/pull/13960 that always generates a `.cargo_vcs_info.json`

### Version

```text
$ cargo version --verbose
cargo 1.81.0 (2dbb1af80 2024-08-20)
release: 1.81.0
commit-hash: 2dbb1af80a2914475ba76827a312e29cedfa6b2f
commit-date: 2024-08-20
host: aarch64-apple-darwin
libgit2: 1.8.1 (sys:0.19.0 vendored)
libcurl: 8.7.1 (sys:0.4.73+curl-8.8.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Mac OS 14.7.1 [64-bit]
```

epage:
[...]

PR #14962, opened 2024-12-19
perf(cargo-package): match certain path prefix with pathspec

weihanglo:

Improve #14955

`check_repo_state` checks the entire git repo status.
This is usually fine if you have only a few packages in a workspace.

For huge monorepos, it may hit performance issues.

For example, on awslabs/aws-sdk-rust@2cbd34db53389aed1cafe4c56d23186ef5ff304a the workspace has roughly 434 members to publish. `git ls-files` reported us 204379 files in this Git repository. That means git may need to check status of all files 434 times.
That would be `204379 * 434 = 88,700,486` checks!

Moreover, the current algorithm is finding the intersection of `PathSource::list_files` and `git status`. It is an `O(n^2)` check. Let's assume files are evenly distributed into each package, so roughly 470 files per package. If we're unlucky to have some dirty files, say 100 files. We will have to do `470 * 100 = 47,000` times of path comparisons.

Even worse, because we `git status` everything in the repo, we'll have to it for all members,
even when those dirty files are not part of the current package in question.
So it becomes `470 * 100 * 434 = 20,398,000`!

#### Solution

Instead of comparing with the status of the entire repository,
this patch use the magic pathspec[1] to tell git only reports
paths that match a certain path prefix.

This wouldn't help the `O(n^2)` algorithm,
but at least it won't check dirty files outside the current package.

[1]: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec

Run this command against awslabs/aws-sdk-rust@2cbd34db53389aed1cafe4c56d23186ef5ff304a,
and see if it is getting better.

```
CARGO_LOG_PROFILE=1 cargor package --no-verify --offline --allow-dirty -p aws-sdk-accessanalyzer -p aws-sdk-apigateway
```

### Additional information

There are some other alternatives, like making `PathSource::list_files` additionally reports dirty files. I don't want to touch that part right now because it is already overly complicated.

Some other approaches like

* switching to gitoxide
* a flag `--no-vcs` to skip vcs at all
* improve the `O(n^2)` algorithm.

This approach should be the most straighforward one.

### DIFF
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -824,7 +824,7 @@ fn check_repo_state(
         // Find the intersection of dirty in git, and the src_files that would
         // be packaged. This is a lazy n^2 check, but seems fine with
         // thousands of files.
-        let dirty_src_files: Vec<String> = src_files
+        let mut dirty_src_files: Vec<_> = src_files
             .iter()
             .filter(|src_file| dirty_files.iter().any(|path| src_file.starts_with(path)))
             .map(|path| {
@@ -847,6 +847,7 @@ fn check_repo_state(
                 dirty,
             }))
         } else {
+            dirty_src_files.sort_unstable();
             anyhow::bail!(
                 "{} files in the working directory contain changes that were \
                  not yet committed into git:\n\n{}\n\n\
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -805,7 +805,7 @@ fn check_repo_state(
     return Ok(Some(VcsInfo { git, path_in_vcs }));
 
     fn git(
-        _pkg: &Package,
+        pkg: &Package,
         src_files: &[PathBuf],
         repo: &git2::Repository,
         opts: &PackageOpts<'_>,
@@ -816,7 +816,8 @@ fn check_repo_state(
         // - ignored (in case the user has an `include` directive that
         //   conflicts with .gitignore).
         let mut dirty_files = Vec::new();
-        collect_statuses(repo, &mut dirty_files)?;
+        let pathspec = relative_pathspec(repo, pkg.root());
+        collect_statuses(repo, &[pathspec.as_str()], &mut dirty_files)?;
         // Include each submodule so that the error message can provide
         // specifically *which* files in a submodule are modified.
         status_submodules(repo, &mut dirty_files)?;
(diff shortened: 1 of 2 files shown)


weihanglo:
From https://github.com/rust-lang/cargo/issues/14955#issuecomment-2555139730

> Here are profiling data: [trace.tar.gz](https://github.com/user-attachments/files/18201025/trace.tar.gz)
> 
> #### On 59b2ddd4484 (trace-offline.json)
>  
> Before
> 
> #### #14962 (trace-offline-pathspec.json)
> 
> After

src/cargo/ops/cargo_package.rs
```diff
@@ -816,7 +816,8 @@ fn check_repo_state(
         // - ignored (in case the user has an `include` directive that
         //   conflicts with .gitignore).
         let mut dirty_files = Vec::new();
-        collect_statuses(repo, &mut dirty_files)?;
+        let pathspec = relative_pathspec(repo, pkg.root());
+        collect_statuses(repo, &[pathspec.as_str()], &mut dirty_files)?;
```
epage, line 820: What if the readme is `../README.md`?

weihanglo: I don't think this PR changed anything for relative paths. It is a bug itself.

They will be filtered out later in

https://github.com/rust-lang/cargo/blob/59b2ddd4484a04a0c202052202cc9f30ac48165c/src/cargo/ops/cargo_package.rs#L829

That said, I don't think either dirty symlinks or relative paths were correctly handled on master.

weihanglo: Let me create a test an verify it.

epage: That is saying if a `src_file` should start with a `dirty_file` path, else filter it out.  I'm not seeing how thats relevant to a `../README.md`.  It would if `Cargo.toml` is in the repo root but it is most likely to be in a sub-directory within the repo, so that `README.md` could be among `dirty_files`.

Or is `src_files` not including these `../README.md` files?

epage: Either way, isn't this cementing the bug in further, making it harder for us to fix?

weihanglo: `src_files` includes symlink files as ordinary files. By the time we get `src_files`, file metadata has been lost.

I don't think this change worsen the situation, unless we consider rewriting `list_files` to do more as a fix to #14955.

epage: What about the non-symlink case?

weihanglo: See #14967.

Going to close this and discuss the performance issue in #14955

PR #14968, opened 2024-12-19
fix(package): show dirty filepaths relative to git workdir

weihanglo:

Dirty file paths in the original message were stripped relative to package root.
User is not able to know the full path to find dirty files.

This PR makes it relative to git workdir.

This was found during #14962


epage (approved):

epage:
This doesn't matter for stable `cargo package` because its not ambiguous but it will for `cargo package --workspace`.

I think the ideal solution is to be relative to the current directory but that takes more work and this incremental improvement is great!

weihanglo:
> This doesn't matter for stable `cargo package` because its not ambiguous but it will for `cargo package --workspace`.

We have  the `--workspace` flag for `cargo package` on stable for along while, so it does affect stable Cargo.

PR #14969, opened 2024-12-19
fixed the error message for a user to open the crate

### LINKED ISSUE #14776: `cargo doc --open` error message is confusing when all targets are skipped

ibilalkayy:

This PR is resolving an [issue](https://github.com/rust-lang/cargo/issues/14776#issuecomment-2515017963) where `cargo doc --open --examples` does not give a clear message when the examples file is given to it. 

I changed the message from `no crates with documentation` in the [cargo_doc.rs](https://github.com/rust-lang/cargo/blob/58b2d609ece6bd0333e4d0f63a024fe3dd350b4b/src/cargo/ops/cargo_doc.rs#L62) and [doc.rs](https://github.com/rust-lang/cargo/blob/58b2d609ece6bd0333e4d0f63a024fe3dd350b4b/tests/testsuite/doc.rs#L1504) to `requested crate documentation is not available to open`. Now it becomes easy for the user to understand what they're missing.

 Fixes #14776

Here is the command through which you can test this change: 

`cargo test SNAPSHOTS=overwrite -- doc::open_no_doc_crate`

### DIFF
--- a/src/cargo/ops/cargo_doc.rs
+++ b/src/cargo/ops/cargo_doc.rs
@@ -56,10 +56,11 @@ pub fn doc(ws: &Workspace<'_>, options: &DocOptions) -> CargoResult<()> {
     let compilation = ops::compile(ws, &options.compile_opts)?;
 
     if options.open_result {
-        let name = &compilation
-            .root_crate_names
-            .get(0)
-            .ok_or_else(|| anyhow::anyhow!("no crates with documentation"))?;
+        let name = &compilation.root_crate_names.get(0).ok_or_else(|| {
+            anyhow::anyhow!(
+                "cannot open specified crate's documentation: no documentation generated"
+            )
+        })?;
         let kind = options.compile_opts.build_config.single_requested_kind()?;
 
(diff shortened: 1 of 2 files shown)


epage:
Previously reviewed at #14957

epage (approved):

epage:
Thanks!

ibilalkayy:
You're welcome, Page!

armujahid:
Thanks. for future PRs, remember that you can clean commit history and force push in your local branch instead of creating another PR.

PR #14970, opened 2024-12-20
fix(package): use relpath to cwd for vcs dirtiness report

weihanglo:

Address https://github.com/rust-lang/cargo/pull/14968#issuecomment-2555901072

> I think the ideal solution is to be relative to the current directory but that takes more work and this incremental improvement is great!

Sorry I should have noticed your comment earlier.

### DIFF
--- a/src/cargo/ops/cargo_package.rs
+++ b/src/cargo/ops/cargo_package.rs
@@ -796,7 +796,7 @@ fn check_repo_state(
         .and_then(|p| p.to_str())
         .unwrap_or("")
         .replace("\\", "/");
-    let Some(git) = git(src_files, &repo, &opts)? else {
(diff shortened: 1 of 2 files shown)


epage:
> Sorry I should have noticed your comment earlier.

I intentionally didn't block on it or raise it as a concern.  This is my "ideal" but there are multiple points of view to be had and the risk of details that weren't blocking the PR.

I leave it to you if you'd want to move this forward.

epage (approved):

weihanglo:
The working directory can be seen as either git's workdir, or user's current working directory. Since the dirtiness check is more a Git thing, this PR makes more sense to me.

```
[ERROR] 1 files in the working directory contain changes that were not yet committed into git:

../Cargo.toml

to proceed despite this and include the uncommitted changes, pass the `--allow-dirty` flag
```

PR #14973, opened 2024-12-21
Moved manifest metadata tracking from fingerprint to dep info

### LINKED ISSUE #14154: Changing workspace Cargo.toml doesn't invalidate build cache for workspace members

ranger-ross:

Fixes #14154

~~Added some of the missing package metadata to the fingerprint so that rebuilds are triggered correctly.~~
~~Also updated the test to prevent a future regression.~~

Moved the manifest metadata tracking to use dep info in favor of fingerprint so that rebuilds are only triggered if the metadata is actually used.

### DIFF
--- a/src/cargo/core/compiler/compilation.rs
+++ b/src/cargo/core/compiler/compilation.rs
@@ -351,8 +351,6 @@ impl<'gctx> Compilation<'gctx> {
             }
         }
 
-        let metadata = pkg.manifest().metadata();
-
         let cargo_exe = self.gctx.cargo_exe()?;
         cmd.env(crate::CARGO_ENV, cargo_exe);
 
@@ -360,7 +358,6 @@ impl<'gctx> Compilation<'gctx> {
         // crate properties which might require rebuild upon change
         // consider adding the corresponding properties to the hash
         // in BuildContext::target_metadata()
-        let rust_version = pkg.rust_version().as_ref().map(ToString::to_string);
         cmd.env("CARGO_MANIFEST_DIR", pkg.root())
             .env("CARGO_MANIFEST_PATH", pkg.manifest_path())
             .env("CARGO_PKG_VERSION_MAJOR", &pkg.version().major.to_string())
@@ -368,37 +365,13 @@ impl<'gctx> Compilation<'gctx> {
             .env("CARGO_PKG_VERSION_PATCH", &pkg.version().patch.to_string())
             .env("CARGO_PKG_VERSION_PRE", pkg.version().pre.as_str())
             .env("CARGO_PKG_VERSION", &pkg.version().to_string())
-            .env("CARGO_PKG_NAME", &*pkg.name())
-            .env(
-                "CARGO_PKG_DESCRIPTION",
-                metadata.description.as_ref().unwrap_or(&String::new()),
-            )
-            .env(
-                "CARGO_PKG_HOMEPAGE",
-                metadata.homepage.as_ref().unwrap_or(&String::new()),
-            )
-            .env(
-                "CARGO_PKG_REPOSITORY",
-                metadata.repository.as_ref().unwrap_or(&String::new()),
-            )
-            .env(
-                "CARGO_PKG_LICENSE",
-                metadata.license.as_ref().unwrap_or(&String::new()),
-            )
-            .env(
-                "CARGO_PKG_LICENSE_FILE",
-                metadata.license_file.as_ref().unwrap_or(&String::new()),
-            )
-            .env("CARGO_PKG_AUTHORS", &pkg.authors().join(":"))
-            .env(
-                "CARGO_PKG_RUST_VERSION",
-                &rust_version.as_deref().unwrap_or_default(),
-            )
-            .env(
-                "CARGO_PKG_README",
-                metadata.readme.as_ref().unwrap_or(&String::new()),
-            )
-            .cwd(pkg.root());
+            .env("CARGO_PKG_NAME", &*pkg.name());
+
+        for (key, value) in pkg.manifest().metadata().env_vars() {
+            cmd.env(key, value.as_ref());
+        }
+
+        cmd.cwd(pkg.root());
 
         apply_env_config(self.gctx, &mut cmd)?;
 
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -78,7 +78,6 @@
 //! [`CompileKind`] (host/target)              | ✓           | ✓                   | ✓                      | ✓
 //! `__CARGO_DEFAULT_LIB_METADATA`[^4]         |             | ✓                   | ✓                      | ✓
 //! `package_id`                               |             | ✓                   | ✓                      | ✓
-//! authors, description, homepage, repo       | ✓           |                     |                        |
 //! Target src path relative to ws             | ✓           |                     |                        |
 //! Target flags (test/bench/for_host/edition) | ✓           |                     |                        |
 //! -C incremental=… flag                      | ✓           |                     |                        |
@@ -189,6 +188,8 @@
 //! files to learn about environment variables that the rustc compile depends on.
 //! Cargo then later uses this to trigger a recompile if a referenced env var
 //! changes (even if the source didn't change).
+//! This also includes env vars generated from Cargo metadata like `CARGO_PKG_DESCRIPTION`.
+//! (See [`crate::core::manifest::ManifestMetadata`]
 //!
 //! #### dep-info files for build system integration.
 //!
@@ -612,10 +613,6 @@ pub struct Fingerprint {
     memoized_hash: Mutex<Option<u64>>,
     /// RUSTFLAGS/RUSTDOCFLAGS environment variable value (or config value).
     rustflags: Vec<String>,
-    /// Hash of some metadata from the manifest, such as "authors", or
-    /// "description", which are exposed as environment variables during
-    /// compilation.
-    metadata: u64,
     /// Hash of various config settings that change how things are compiled.
     config: u64,
     /// The rustc target. This is only relevant for `.json` files, otherwise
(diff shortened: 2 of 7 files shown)


src/cargo/core/compiler/fingerprint/mod.rs
weihanglo, line 1: Thanks for the PR, Ross!

In https://github.com/rust-lang/cargo/pull/14811#discussion_r1880431940 we had discussed th other direction: remove all manifest metadata fields tracking from fingerprint, and only rely on dep-info tracking when needed.

Here is my rough idea:

* For [`ManifestMetadata`](https://github.com/rust-lang/cargo/blob/652623b779c88fe44afede28bf7f1c9c07812511/src/cargo/core/manifest.rs?plain=1#L133) struct, add some helper functions:
   * One for returning all metadata environment variable names Cargo has to emit. This can be used later for keeping those variables in Cargo's dep-info [here](https://github.com/rust-lang/cargo/blob/652623b779c88fe44afede28bf7f1c9c07812511/src/cargo/core/compiler/fingerprint/dep_info.rs?plain=1#L337)
   * Extracting [this code](https://github.com/rust-lang/cargo/blob/652623b779c88fe44afede28bf7f1c9c07812511/src/cargo/core/compiler/compilation.rs?plain=1#L372-L400), so we can consolidate the set of variables Cargo emits in one place.
 * Remove manifest metadata tracking in fingerprint from this file. We rely purely on dep-info tracking on-demand, avoiding eager rebuilds.
 
Some fields like `categories`, `keywords`, and `badges` we can put off, since they aren't emitted elsewhere. And `links` is perhaps worth its own fix.

weihanglo: Whatever approach we take, don't forget to update [this gigantic table](https://github.com/rust-lang/cargo/blob/652623b779c88fe44afede28bf7f1c9c07812511/src/cargo/core/compiler/fingerprint/mod.rs?plain=1#L81).
(We have a good online rendered version [here](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/compiler/fingerprint/index.html#fingerprints-and-unithashs))

ranger-ross: Thanks for the detailed explanation. 
I think the approach makes sense to me. :+1: 

I'll take a shot at implementing that in this PR.

ranger-ross: I reworked this PR to move the metadata fields from the fingerprint to dep-info like you proposed above.

> Extracting [this code](https://github.com/rust-lang/cargo/blob/652623b779c88fe44afede28bf7f1c9c07812511/src/cargo/core/compiler/compilation.rs?plain=1#L372-L400), so we can consolidate the set of variables Cargo emits in one place.

I tried this approach by creating a `EmittablePackageMetadata` enum. I am not in love with the `EMITTED_METADATA_MAPPINGS` array but it works and is in the same file/area as the enum definition.

Let me know what you think

weihanglo:
Huh! Didn't realize you pushed more updates.

src/cargo/core/compiler/compilation.rs
```diff
@@ -351,54 +352,21 @@ impl<'gctx> Compilation<'gctx> {
...
         // in BuildContext::target_metadata()
-        let rust_version = pkg.rust_version().as_ref().map(ToString::to_string);
-        cmd.env("CARGO_MANIFEST_DIR", pkg.root())
-            .env("CARGO_MANIFEST_PATH", pkg.manifest_path())
-            .env("CARGO_PKG_VERSION_MAJOR", &pkg.version().major.to_string())
-            .env("CARGO_PKG_VERSION_MINOR", &pkg.version().minor.to_string())
-            .env("CARGO_PKG_VERSION_PATCH", &pkg.version().patch.to_string())
-            .env("CARGO_PKG_VERSION_PRE", pkg.version().pre.as_str())
-            .env("CARGO_PKG_VERSION", &pkg.version().to_string())
-            .env("CARGO_PKG_NAME", &*pkg.name())
```
weihanglo, line 371: I would probably keep these untouched, and work on metatata for now.

They are either already tracked by other means (PackageId), or still under discussion (https://github.com/rust-lang/cargo/issues/8693).

src/cargo/core/compiler/fingerprint/mod.rs
weihanglo: Yeah the mapping, while working, doesn't look too pleasant 😆.

I've spent a bit time hacking on a macro solution. I feel like it is a bit clearer (?). 


```rust
macro_rules! get_metadata_env {
    ($meta:ident, $field:ident) => ($meta.$field.as_deref().unwrap_or_default().into());
    ($meta:ident, $field:ident, $to_var:expr) => ($to_var($meta).into());
}

macro_rules! metadata_envs {
    (
        $(
            ($field:ident, $key:literal$(, $to_var:expr)?),
        )*
    ) => {
        struct MetadataEnvs;
        impl MetadataEnvs {
            $(
                fn $field(meta: &ManifestMetadata) -> Cow<'_, str> {
                    get_metadata_env!(meta, $field$(, $to_var)?)
                }
            )*

            pub fn should_track(key: &str) -> bool {
                let keys = [$($key),*];
                key.strip_prefix("CARGO_PKG_")
                    .map(|key| keys.iter().any(|k| *k == key))
                    .unwrap_or_default()
            }

            pub fn var<'a>(meta: &'a ManifestMetadata, key: &str) -> Option<Cow<'a, str>> {
                key.strip_prefix("CARGO_PKG_").and_then(|key| match key {
                    $($key => Some(Self::$field(meta)),)*
                    _ => None,
                })
            }

            pub fn vars(meta: &ManifestMetadata) -> impl Iterator<Item = (&'static str, Cow<'_, str>)> {
                [
                    $(
                        (
                            concat!("CARGO_PKG_", $key),
                            Self::$field(meta),
                        ),
                    )*
                ].into_iter()
            }
        }
    }
}

metadata_envs! {
    (description, "DESCRIPTION"),
    (homepage, "HOMEPAGE"),
    (repository, "REPOSITORY"),
    (license, "LICENSE"),
    (license_file, "LICENSE_FILE"),
    (authors, "AUTHORS", |m: &ManifestMetadata| m.authors.join(":")),
    (rust_version, "RUST_VERSION", |m: &ManifestMetadata| m.rust_version.as_ref().map(ToString::to_string).unwrap_or_default()),
    (readme, "README"),
}

impl ManifestMetadata {
    /// Whether the given env var should be tracked by Cargo's dep-info.
    pub fn should_track(env_key: &str) -> bool {
        MetadataEnvs::should_track(env_key)
    }

    pub fn env_var<'a>(&'a self, env_key: &str) -> Option<Cow<'a, str>> {
        MetadataEnvs::var(self, env_key)
    }

    pub fn env_vars(&self) -> impl Iterator<Item = (&'static str, Cow<'_, str>)> {
        MetadataEnvs::vars(&self)
    }
}
```

src/cargo/core/compiler/compilation.rs
weihanglo: I mean. We can revisit them later, after we figure out manifest metdata part. Doesn't need to do everything in one PR :)

src/cargo/core/compiler/compilation.rs
```diff
@@ -351,54 +352,21 @@ impl<'gctx> Compilation<'gctx> {
...
-            .env(
-                "CARGO_PKG_README",
-                metadata.readme.as_ref().unwrap_or(&String::new()),
-            )
-            .cwd(pkg.root());
+        for (key, value) in pkg.emitted_env_vars() {
+            match value {
+                Cow::Borrowed(value) => cmd.env(key, value),
+                Cow::Owned(ref value) => cmd.env(key, value.as_str()),
+            };
```
weihanglo, line 366: nit: 

```suggestion
            cmd.env(key, value.as_ref());
```

src/cargo/core/compiler/fingerprint/mod.rs
ranger-ross: ah nice, this looks better to me. 
I need to work on my "macro-fu" for scenarios like this :stuck_out_tongue_closed_eyes: 

I updated the PR to use this and address a few more of your other comments in the latest commit. :+1:

weihanglo: Ah. Nice!

Thanks for adopting the hacky hacky solution 😆

Would you mind rearranging commits? I mean like the first commit made some intermediate changes later got deleted, which may be confusing when reading through `git blame`.

To myself I may squash into one commit for this one.

ranger-ross: sure, I squashed everything into one commit and gave it a better commit message :+1:  https://github.com/rust-lang/cargo/pull/14973/commits/3d7b154d158eea5fbf1564320f9043371ff92c62

weihanglo (approved):
Thank you for working on this with us!

PR #14975, opened 2024-12-22
upgrade `gix` to the latest release 0.69.

Byron:
The main benefit is that it won't facilitate hangs due to attempts to read from untrackable directory entries, like names pipes or sockets.

Related to https://github.com/GitoxideLabs/gitoxide/pull/1629

### Tasks

* [x] upgrade
* [x] incorporate updated `gix` [once everything is fixed](https://github.com/GitoxideLabs/gitoxide/pull/1740).
* [x] assure tests work

### Postponed

It turns out that the new `gix` version doesn't magically fix the FIFO issue, so the following test still fails.

It's not super-trivial to fix apparently (I tried), so let's do it in a separate PR.

Here is the patch I have so far in case anyone is interested to fix it earlier or wants to share insights :).

```patch
commit dfef545eae215f0b9da9f3d4424b52cba7edaec3
Author: Sebastian Thiel <sebastian.thiel@icloud.com>
Date:   Sun Dec 22 19:05:40 2024 +0100

    fix: assure possibly blocking non-files (like FIFOs) won't be picked up for publishing.
    
    This would otherwise cause the publish to hang.

diff --git a/src/cargo/sources/path.rs b/src/cargo/sources/path.rs
index 776590697..c78463a32 100644
--- a/src/cargo/sources/path.rs
+++ b/src/cargo/sources/path.rs
@@ -626,8 +626,11 @@ fn list_files_gix(
         .filter(|res| {
             // Don't include Cargo.lock if it is untracked. Packaging will
             // generate a new one as needed.
+            // Also don't include untrackable directory entries, like FIFOs.
             res.as_ref().map_or(true, |item| {
-                !(item.entry.status == Status::Untracked && item.entry.rela_path == "Cargo.lock")
+                item.entry.disk_kind != Some(gix::dir::entry::Kind::Untrackable)
+                    && !(item.entry.status == Status::Untracked
+                        && item.entry.rela_path == "Cargo.lock")
             })
         })
         .map(|res| res.map(|item| (item.entry.rela_path, item.entry.disk_kind)))
diff --git a/tests/testsuite/package.rs b/tests/testsuite/package.rs
index 1740de4ac..1c6b3db89 100644
--- a/tests/testsuite/package.rs
+++ b/tests/testsuite/package.rs
@@ -6873,3 +6873,29 @@ See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for
 "#]])
         .run();
 }
+
+#[cargo_test]
+#[cfg(unix)]
+fn simple_with_fifo() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+            "#,
+        )
+        .build();
+
+    std::process::Command::new("mkfifo")
+        .current_dir(p.root())
+        .arg(p.root().join("blocks-when-read"))
+        .status()
+        .expect("a FIFO can be created");
+
+    // If this hangs, Cargo tried to package a FIFO and is reading it forever.
+    p.cargo("package").run();
+}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1161,9 +1161,9 @@ dependencies = [
 
 [[package]]
 name = "gix"
-version = "0.68.0"
+version = "0.69.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b04c66359b5e17f92395abc433861df0edf48f39f3f590818d1d7217327dd6a1"
+checksum = "8d0eebdaecdcf405d5433a36f85e4f058cf4de48ee2604388be0dbccbaad353e"
 dependencies = [
  "gix-actor",
  "gix-attributes",
@@ -1197,6 +1197,7 @@ dependencies = [
  "gix-revision",
  "gix-revwalk",
  "gix-sec",
+ "gix-shallow",
  "gix-submodule",
  "gix-tempfile",
  "gix-trace",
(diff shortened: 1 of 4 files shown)


weihanglo (approved):
Interesting FIFO issue 😆.

Agree we can deal with it separately. Thank you for this update!

PR #14980, opened 2024-12-24
simplify SourceID Ord/Eq

Eh2406:

This is a followup to #14800. Like that PR, this is a small incremental change that does not pull its own weight. If this PR is accepted, the next PR will unlock large performance wins. I am not posting them together because the logic of why this PR is correct is subtle and deserves to be discussed and reviewed without unrelated code changes.

All tests pass on all commits. This **should** be reviewed one commit at a time. 

### Additional information

I pushed one commit at a time, so that CI can confirm that the assert (in the first commit) is never hit.

### DIFF
--- a/src/cargo/core/source_id.rs
+++ b/src/cargo/core/source_id.rs
@@ -598,14 +598,38 @@ impl Ord for SourceId {
             other => return other,
         }
 
-        // If the `kind` and the `url` are equal, then for git sources we also
-        // ensure that the canonical urls are equal.
+        let ord = self.inner.canonical_url.cmp(&other.inner.canonical_url);
+
         match (&self.inner.kind, &other.inner.kind) {
             (SourceKind::Git(_), SourceKind::Git(_)) => {
-                self.inner.canonical_url.cmp(&other.inner.canonical_url)
+                // In the pre-PR code we returned Ord here,
+                // so there is no chance that this commit has broken anything about this match arm.
             }
-            _ => self.inner.url.cmp(&other.inner.url),
+            _ => {
+                // In the pre-PR code we returned cmp of url here, so let's make sure that's the same.
+                assert_eq!(self.inner.url.cmp(&other.inner.url), ord);
+                // I am quite sure that this assert will never fire.
+                // In order for it to fire either `url`s are equal but `canonical_url`s are not,
+                // or the other way around. The algorithm for constructing a canonical URL is deterministic,
+                // so if it's given the same URL it will return the same canonical URL.
+
+                // But what if we have two different URLs that canonical eyes the same?
+                // I assert that the second one would get thrown out when the second `SourceId` was interned.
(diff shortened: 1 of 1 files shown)


weihanglo:
Potential risk:

> If someone has two alternate registry hosted on github whose name has one of the weird languages where `a < b` does not imply `a.to_lower() < b.to_lower()` (we did this normalization in `CanonicialUrl` for only `github.com` domain), then the package order in the lockfile file could switch between different versions of Cargo.

Though this is pretty niche and I doubt it impacts any real world use case.

weihanglo:
@rfcbot fcp merge

I propose to the team to merge this. See the comment in 2a9527be4926c75f99030d76cc6715641acf48b9 for detailed explanation.

tl;dr

* Cargo can never have two `SourceId`s with everything the same except for the `url` field, because when interning, we hash only `kind`, `precise`, and `canonical_url`, not `url`.
* One of the risks we can think of is described in https://github.com/rust-lang/cargo/pull/14980?new_mergebox=true#issuecomment-2561384117
* The other one might be a `path` dependency having a path ending with `.git` [got chopped off](https://github.com/rust-lang/cargo/blob/c86f4b3a1b153218e6e50861214b0b4b4e695f23/src/cargo/util/canonical_url.rs?plain=1#L49-L57), making the package order different in lockfile. I also think this is a weird use case we can ignore.

If there is any implication we're missing here, please call it out!

Eh2406:
Note that  all  of the risks identified so far involve the implications to `ord`  and the performance benefits mostly relate to `eq`. If more risks are found, or the risks related to `ord`  are to painful this PR can be updated to leave `ord` as is and  instead do the comparable thing directly on `eq`. I prefer the code simplicity of `eq`  delegating to `ord`, but opinions can legitimately vary.

Rustin170506:
> If someone has two alternate registry hosted on github whose name has one of the weird languages where `a < b` does not imply `a.to_lower() < b.to_lower()`

May I ask, do you have a real example of this?

weihanglo:
I don't, unfortunately. I don't even know if there is a language has such odd upper/lowercase comparison. CJK might potentially maybe?

cc @Eh2406, is it possible to create one real example on your side?

Eh2406:
I only speak English. So I've no idea how realistic these examples are. But some googling found this list https://stackoverflow.com/questions/67106784/why-is-upper-casing-not-enough-for-case-insensitive-comparison and with the help of some intelligent auto complete here are sum reproducers https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f4bc9bc63fa18a63f2e503442521e82c

weihanglo (approved):
Thanks Jacob.

Since we have reached a consensus, going to merge this now. I don't think waiting for an entire 10-day FCP will make any difference. If there is a real use case, only when the change hits nightly and people start using it will we know.

Issue #14967, opened 2024-12-19
`cargo package` VCS dirty check corner cases


weihanglo:
### Problem

During the investigation of #14955 and #14962, we found more corner cases that `cargo package` doesn't consider dirty. You don't even need to pass `--allow-dirty` flag to explicitly allow it. `.cargo_vcs_info.json` shows no dirty at all. However, the actual packaged `.crate` file contains dirty files. 

Let's consider this virtual workspace:

```
./
├── foo/
│   ├── src/
│   ├── Cargo.toml
│   └── README.md -> ../README.md
├── Cargo.toml
├── LICENSE
└── README.md
```

```toml
# foo/Cargo.toml
[package]
name = "foo"
edition.workspace = true
license-file = "../LICENSE" # link to parent LICENSE
```

```toml
# Cargo.toml
[workspace]
members = ["foo"]
[workspace.package]
edition = "2015"
```

There are at least three corner cases:

* [x] Symlinks — if you change README.md at workspace root level, when packaging it would be included. However, since [`PathSource::list_files`](https://github.com/rust-lang/cargo/blob/59b2ddd4484a04a0c202052202cc9f30ac48165c/src/cargo/sources/path.rs?plain=1#L97) list symlink as ordinary file. When doing dirty check, the [naive path prefix check](https://github.com/rust-lang/cargo/blob/59b2ddd4484a04a0c202052202cc9f30ac48165c/src/cargo/sources/path.rs?plain=1#L97) is not able to find the symlink source file dirty.
  * Fixed by https://github.com/rust-lang/cargo/pull/14981
* [x] `package.license-file` and `package.readme` — they are copied during `cargo package`. `PathSource::list_files` doesn't even report them because they are not under the directory of the package. 
  * Fixed by https://github.com/rust-lang/cargo/pull/14966
* [x] Workspace inheritance — like changing `workspace.package.edition` to `2021` would actually make foo's Cargo.toml dirty, but the current dirty status check doesn't capture that.
  * Fixed by <https://github.com/rust-lang/cargo/pull/15276> and <https://github.com/rust-lang/cargo/pull/15341>. We do an aggressively check that if the workspace Cargo.toml is dirty, consider it dirty.
* [ ] Symlinks linking to Git submodules. See #15384.
  * It also has a performance issue and needs refactor. When considering performance, also keep in mind large monorepo-level workspace performance like #14955

### Steps

See https://github.com/rust-lang/cargo/pull/14966

### Version

```text
[...]

Issue #14965, opened 2024-12-19
cargo vendor is not consistent about paths and symbolic links on Windows vs. other platforms


glandium:
### Problem

About paths:
when a crate contains paths for e.g. dependencies in a workspace, `cargo vendor` uses / on Unices and \ on Windows, changing whatever the original path was.

About symbolic links:
when a crate contains symbolic links to another file in a workspace, `cargo vendor` expands the content on Unices and replaces it with the path of the link target on Windows.

### Steps

1. `cargo new foo`
2. `cd foo`
3. `cargo add libz-rs-sys`
4. `cargo vendor`

On Linux or Mac, you end up with a proper `vendor/libz-rs-sys/LICENSE` file, and `vendor/libz-rs-sys/Cargo.toml` has `dependencies.zlib-rs.path = "../zlib-rs"`.
On Windows, `vendor/libz-rs-sys/LICENSE` literally contains `../LICENSE`, and `vendor/libz-rs-sys/Cargo.toml` has `dependencies.zlib-rs.path = "..\zlib-rs"`.

### Possible Solution(s)

cargo vendor should be consistent, wherever it's run.

### Version

```text

```

weihanglo:
@glandium I had a hard time reproducing this. The repro seems to depend on a non-exact package version on crates.io. And even after I tried  some recent versions, I didn't see `dependencies.zlib-rs-sys.path` in the manifest file. Were you using git dependencies then? If yes which commit and repo?

glandium:
Did crates.io repackage crates? I'm pretty sure I tested with crates.io back then, and I can't reproduce myself now.

But this is reproducible with `libz-rs-sys = {git="https://github.com/memorysafety/zlib-rs"}`

weihanglo:
> Did crates.io repackage crates? I'm pretty sure I tested with crates.io back then, and I can't reproduce myself now.

I am not aware of that. Even it really did, `cargo publish` should copy LICENSE and normalize the path (there were some bugs though).

I'll go ahead and close this as a duplicate of #7051, as they are all git-dependency related. Thank you!

Issue #13331, opened 2024-01-21
Add `include` and `exclude` information to metadata


GuillaumeGomez:
### Problem

I'm currently working on https://github.com/rust-lang/rust-clippy/issues/11677. However, to finish this lint, I need to have the `include` and `exclude` information in the metadata (to be used with `cargo_metadata`).

### Proposed Solution

I think that to prevent having to recompute `include` and `exclude` when using it from `metadata` and duplicate what cargo does, it should contain the equivalent of `cargo package --list`. An initial implementation was done in

epage:
FYI the title and problem are still focused on one specific solution, rather than the problem at hand.

Going to copy oer from #13330 @weihanglo's comment which has some good context when considering solutions:

> While it might be useful to have `include` and `exclude` in `cargo metadata` output, it doesn't seem to be a proper way to fix [rust-lang/rust-clippy#11677](https://github.com/rust-lang/rust-clippy/issues/11677). The solution might end up reimplement the logic [Cargo find and package files](https://github.com/rust-lang/cargo/blob/7bb7b539558dc88bea44cee4168b6269bf8177b0/src/cargo/sources/path.rs#L133-L200), which includes `.gitignore`, shell glob syntax, symlinks handling, git dirty status, and other details (see #11405). `cargo package --list` might be more reliable, but not 100% accurate (see #8407 and #11666).
> 
> I would suggest starting the discussion from an issue, and jump back to implementation when the design is settled.

For the most part, `cargo package --list` should be fine for your use case since this involves the source code looking up paths the same way.  For other use cases, 
we'd likely need to output more than `cargo package --list` but the mapping cargo makes from source to destination (file moves, resolving of symlinks, whether the `Cargo.lock` is generated or pulled from the root, etc).  We'd need to be confident enough in this new data structure to meet our compatibility guarantees.

Besides those limitations in `cargo package --list`, it can be slow.  I currently use it in `cargo release` to identify which packages are affected by which commits (making me a candidate for that extended use case).  I worry that if we include it within `cargo metadata`, we'll negatively impact all other users.

We could have `clippy` run `cargo package --list` only if that lint is enabled but then that will forever keep this as a niche lint.

GuillaumeGomez:
I was pretty sure that I was missing on the overall complexity on cargo side hence why I implemented it with just the fields values and why I was planning to retrieve the information from `clippy` by emulating how cargo was computing it. ^^'

weihanglo:
[...]

PR #14981, opened 2024-12-24
fix(package): check dirtiness of symlinks source files

weihanglo:

This adds a special case for checking source files are symlinks
and have been modified when under a VCS control

This is required because those paths may link to a file outside the
current package root, but still under the git workdir, affecting the
final packaged `.crate` file.

Pretty similar to #14966, as a part of #14967.

This may have potential performance issue. If a package contains thousands of symlinks, Cargo will fire `git status` for each of them. Not sure if we want to do anything proactively now.

The introduction of the `PathEntry` struct gives us more room for storing file metadata to satisfiy use cases in the future. For instances,

* Knowing a source file is a symlink and resolving it when packaging on Windows
  * #5664
  * #14965
* Knowing more about a file's metadata (e.g. permission bits from Git)
  * #4413
  * #8006
* Provide richer information for `cargo package --list`, for example JSON output mode
  * #11666
  * #13331
  * #13953

### DIFF
--- a/src/cargo/ops/cargo_package/vcs.rs
+++ b/src/cargo/ops/cargo_package/vcs.rs
@@ -242,12 +242,7 @@ fn dirty_metadata_paths(pkg: &Package, repo: &git2::Repository) -> CargoResult<V
         if let Ok(rel_path) = paths::strip_prefix_canonical(&abs_path, workdir) {
             // Outside package root but under git workdir,
             if repo.status_file(&rel_path)? != git2::Status::CURRENT {
-                dirty_files.push(if abs_path.is_symlink() {
-                    // For symlinks, shows paths to symlink sources
-                    workdir.join(rel_path)
-                } else {
-                    abs_path
-                });
+                dirty_files.push(workdir.join(rel_path))
             }
         }
     }
--- a/src/cargo/ops/cargo_package/vcs.rs
+++ b/src/cargo/ops/cargo_package/vcs.rs
@@ -1,5 +1,6 @@
 //! Helpers to gather the VCS information for `cargo package`.
 
+use std::collections::HashSet;
 use std::path::Path;
 use std::path::PathBuf;
 
@@ -187,6 +188,7 @@ fn git(
         .filter(|src_file| dirty_files.iter().any(|path| src_file.starts_with(path)))
         .map(|p| p.as_ref())
         .chain(dirty_metadata_paths(pkg, repo)?.iter())
+        .chain(dirty_symlinks(pkg, repo, src_files)?.iter())
         .map(|path| {
             pathdiff::diff_paths(path, cwd)
                 .as_ref()
@@ -247,6 +249,34 @@ fn dirty_metadata_paths(pkg: &Package, repo: &git2::Repository) -> CargoResult<V
         }
     }
     Ok(dirty_files)
+}
+
+/// Checks whether source files are symlinks and have been modified.
+///
+/// This is required because those paths may link to a file outside the
+/// current package root, but still under the git workdir, affecting the
+/// final packaged `.crate` file.
+fn dirty_symlinks(
+    pkg: &Package,
+    repo: &git2::Repository,
+    src_files: &[PathEntry],
+) -> CargoResult<HashSet<PathBuf>> {
+    let workdir = repo.workdir().unwrap();
+    let mut dirty_symlinks = HashSet::new();
+    for rel_path in src_files
+        .iter()
+        .filter(|p| p.is_symlink_or_under_symlink())
+        .map(|p| p.as_ref().as_path())
+        // If inside package root. Don't bother checking git status.
+        .filter(|p| paths::strip_prefix_canonical(p, pkg.root()).is_err())
+        // Handle files outside package root but under git workdir,
+        .filter_map(|p| paths::strip_prefix_canonical(p, workdir).ok())
+    {
+        if repo.status_file(&rel_path)? != git2::Status::CURRENT {
+            dirty_symlinks.insert(workdir.join(rel_path));
+        }
+    }
+    Ok(dirty_symlinks)
 }
 
 /// Helper to collect dirty statuses for a single repo.
--- a/src/cargo/ops/cargo_package/vcs.rs
+++ b/src/cargo/ops/cargo_package/vcs.rs
@@ -235,11 +235,11 @@ fn dirty_metadata_paths(pkg: &Package, repo: &git2::Repository) -> CargoResult<V
             continue;
         };
         let abs_path = paths::normalize_path(&root.join(path));
-        if paths::strip_prefix_canonical(abs_path.as_path(), root).is_ok() {
+        if paths::strip_prefix_canonical(&abs_path, root).is_ok() {
             // Inside package root. Don't bother checking git status.
             continue;
         }
-        if let Ok(rel_path) = paths::strip_prefix_canonical(abs_path.as_path(), workdir) {
+        if let Ok(rel_path) = paths::strip_prefix_canonical(&abs_path, workdir) {
             // Outside package root but under git workdir,
             if repo.status_file(&rel_path)? != git2::Status::CURRENT {
                 dirty_files.push(if abs_path.is_symlink() {
--- a/src/cargo/ops/cargo_package/vcs.rs
+++ b/src/cargo/ops/cargo_package/vcs.rs
@@ -1,7 +1,6 @@
 //! Helpers to gather the VCS information for `cargo package`.
 
 use std::collections::HashSet;
-use std::path::Path;
 use std::path::PathBuf;
 
 use anyhow::Context as _;
@@ -187,8 +186,7 @@ fn git(
         .iter()
         .filter(|src_file| dirty_files.iter().any(|path| src_file.starts_with(path)))
         .map(|p| p.as_ref())
-        .chain(dirty_metadata_paths(pkg, repo)?.iter())
-        .chain(dirty_symlinks(pkg, repo, src_files)?.iter())
+        .chain(dirty_files_outside_pkg_root(pkg, repo, src_files)?.iter())
         .map(|path| {
             pathdiff::diff_paths(path, cwd)
                 .as_ref()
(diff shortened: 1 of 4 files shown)


src/cargo/ops/cargo_package.rs
```diff
@@ -897,6 +903,36 @@ fn check_repo_state(
...
+            .map(|p| p.as_path())
+            // If inside package root. Don't bother checking git status.
+            .filter(|p| paths::strip_prefix_canonical(*p, pkg.root()).is_err())
+            // Handle files outside package root but under git workdir,
+            .filter_map(|p| paths::strip_prefix_canonical(p, workdir).ok())
+        {
+            // TODO: Should we warn users if there are like thousands of symlinks,
+            // which may hurt the performance?
+            if repo.status_file(&rel_path)? != git2::Status::CURRENT {
+                dirty_symlinks.insert(workdir.join(rel_path));
```
weihanglo, line 930: We could potentially cache git status for each repo then. That leaves to future improvement if this one has got merged.

weihanglo:
This doesn't handle symlink directories correctly with git. Fixing…

weihanglo:
> This doesn't handle symlink directories correctly with git. Fixing…

This has been fixed.

src/cargo/ops/cargo_package.rs
weihanglo: See <https://github.com/rust-lang/cargo/pull/14985>.

src/cargo/sources/path.rs
```diff
@@ -404,6 +404,32 @@ impl<'gctx> Source for RecursivePathSource<'gctx> {
     }
 }
 
+/// [`PathBuf`] with extra metadata.
+#[derive(Clone, Debug)]
```
epage, line 443: > feat(source): wrap PathBuf with PathEntry

nit: as the lib is not intended as an API, this is more of a `refactor` than a `feat`

src/cargo/sources/path.rs
```diff
@@ -404,16 +404,70 @@ impl<'gctx> Source for RecursivePathSource<'gctx> {
     }
 }
 
+/// Type that abstracts over [`gix::dir::entry::Kind`] and [`fs::FileType`].
+#[derive(Debug, Clone, Copy)]
+enum FileType {
```
epage, line 409: > feat(package): preserve file type information in PathEntry 

nit: ditto about `refactor`

src/cargo/sources/path.rs
```diff
@@ -834,10 +854,27 @@ fn list_files_walk(
...
+                    Some(_) | None => {
+                        // Not under any parent symlink dir, update the current one.
+                        current_symlink_dir = if file_type.is_dir() {
+                            entry.path_is_symlink().then(|| entry.path().to_path_buf())
+                        } else {
+                            None
+                        };
+                    }
+                }
+
```
epage, line 877: This assumes child files are walked before child directories, is that valid?

- `a/`
  - `b/`
  - `c/`
  - `d.txt`

Walking that:
1. `a/`, `current_symlink_dir='a'`
2. `a/b/`, `current_symlink_dir='a/b'`
3. `a/c/`, `current_symlink_dir='a/c'`
4. `a/d.txt`, `current_symlink_dir=None`

src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -133,12 +134,14 @@ fn git(
...
     // thousands of files.
+    let mut dirty_files_outside_pkg_root = dirty_symlinks(pkg, repo, src_files)?;
+    dirty_files_outside_pkg_root.extend(dirty_metadata_paths(pkg, repo)?);
     let cwd = gctx.cwd();
     let mut dirty_src_files: Vec<_> = src_files
         .iter()
         .filter(|src_file| dirty_files.iter().any(|path| src_file.starts_with(path)))
         .map(|p| p.as_ref())
-        .chain(dirty_metadata_paths(pkg, repo)?.iter())
+        .chain(dirty_files_outside_pkg_root.iter())
```
epage, line 144: There a reason you aren't chaining them?

src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -206,6 +209,36 @@ fn dirty_metadata_paths(pkg: &Package, repo: &git2::Repository) -> CargoResult<V
...
+        .iter()
+        .filter(|p| p.is_symlink_or_under_symlink())
+        .map(|p| p.as_ref().as_path())
+        // If inside package root. Don't bother checking git status.
+        .filter(|p| paths::strip_prefix_canonical(*p, pkg.root()).is_err())
+        // Handle files outside package root but under git workdir,
+        .filter_map(|p| paths::strip_prefix_canonical(p, workdir).ok())
+    {
+        // TODO: Should we warn users if there are like thousands of symlinks,
+        // which may hurt the performance?
```
epage, line 234: If we need to track this, we should probably do this in a way that has visibility

src/cargo/sources/path.rs
weihanglo: Huh, we may need a stack.

src/cargo/ops/cargo_package/vcs.rs
weihanglo: Did you mean opening an issue?

I can do a benchmark and find a reasonable number to emit a warning.

src/cargo/sources/path.rs
weihanglo: Fixed in #14994

src/cargo/sources/path.rs
weihanglo: Fixed in #14994

src/cargo/ops/cargo_package/vcs.rs
weihanglo: Deduplicate same symlinks both reported in `dirty_metadata_paths` and `dirty_symlinks`.

I would go either

* make `dirty_metadata_paths` stop handling symlinks.
* make this a separate commit to show the dedup change in Git history
* combine `dirty_metadata_paths` and `dirty_symlinks` into a single function.

Any preference?

src/cargo/ops/cargo_package/vcs.rs
epage: Alternatively, if this is speculation, should we wait on a user to have a problem?

src/cargo/ops/cargo_package/vcs.rs
epage: In my C++ days, I would use an out-param instead of every function generating a data structure that allocates.  That approach isn't in too much favor in the Rust community though.

What we have is probably fine.

src/cargo/ops/cargo_package/vcs.rs
weihanglo: Sounds like we can just remove the TODO comment.

src/cargo/sources/path.rs
weihanglo: Probably not?

- `a/`
  - `b/`
  - `c/`
  - `d.txt`

Walking that:
1. `a/`, `current_symlink_dir='a'`
2. `a/b/`, `current_symlink_dir='a'` because still `.starts_with(dir)`
3. `a/c/`, `current_symlink_dir='a'` because still `.starts_with(dir)`
4. `a/d.txt`, `current_symlink_dir='a'` because still `.starts_with(dir)`

We keep the farthest symlink directory as long as possible, until we walk to the other completely outside the previous symlink directory.

weihanglo: Given this project layout, I got these from `list_files`:

```
./
├── dir/
│   └── b
├── src/
│   ├── dir -> ../dir/
│   ├── regular-dir/
│   │   ├── symlink-dir -> ../../dir/
│   │   └── c
│   ├── symlink-dir -> ../dir/
│   ├── a
│   ├── lib.rs
│   └── z
├── symlink-src -> src/
├── symlink-src-but-actually-not/
│   └── x
└── Cargo.toml
```

```rust
[src/cargo/ops/cargo_package/vcs.rs:49:5] src_files = [
    PathEntry {
        path: "/foo/Cargo.toml",
        ty: File,
        under_symlink_dir: false,
    },
    PathEntry {
        path: "/foo/symlink-src/regular-dir/symlink-dir/b",
        ty: File,
        under_symlink_dir: true,
    },
    PathEntry {
        path: "/foo/symlink-src/regular-dir/c",
        ty: File,
        under_symlink_dir: true,
    },
    PathEntry {
        path: "/foo/symlink-src/symlink-dir/b",
        ty: File,
        under_symlink_dir: true,
    },
    PathEntry {
        path: "/foo/symlink-src/lib.rs",
        ty: File,
        under_symlink_dir: true,
    },
    PathEntry {
        path: "/foo/symlink-src/z",
        ty: File,
        under_symlink_dir: true,
    },
    PathEntry {
        path: "/foo/symlink-src/a",
        ty: File,
        under_symlink_dir: true,
    },
    PathEntry {
        path: "/foo/symlink-src/dir/b",
        ty: File,
        under_symlink_dir: true,
    },
    PathEntry {
        path: "/foo/.gitignore",
        ty: File,
        under_symlink_dir: false,
    },
    PathEntry {
        path: "/foo/dir/b",
        ty: File,
        under_symlink_dir: false,
    },
    PathEntry {
        path: "/foo/symlink-src-but-actually-not/x",
        ty: File,
        under_symlink_dir: false,
    },
    PathEntry {
        path: "/foo/src/regular-dir/symlink-dir/b",
        ty: File,
        under_symlink_dir: true,
    },
    PathEntry {
        path: "/foo/src/regular-dir/c",
        ty: File,
        under_symlink_dir: false,
    },
    PathEntry {
        path: "/foo/src/symlink-dir/b",
        ty: File,
        under_symlink_dir: true,
    },
    PathEntry {
        path: "/foo/src/lib.rs",
        ty: File,
        under_symlink_dir: false,
    },
    PathEntry {
        path: "/foo/src/z",
        ty: File,
        under_symlink_dir: false,
    },
    PathEntry {
        path: "/foo/src/a",
        ty: File,
        under_symlink_dir: false,
    },
    PathEntry {
        path: "/foo/src/dir/b",
        ty: File,
        under_symlink_dir: true,
    },
]
```

src/cargo/ops/cargo_package/vcs.rs
weihanglo: I've done one additional commit to make it clear that we dedup.

src/cargo/sources/path.rs
epage: Yeah, I forgot you only track the root-most one.  So this is likely fine.  Still feels confusing or brittle but likely it will be fine.

epage (approved):

Issue #14863, opened 2024-11-27
`cargo add`, etc do not preserve `\r\n` line endings


epage:
### Problem

Generally people want line endings to be consistent.  When Cargo modifies a `Cargo.toml`, it may insert `\n` independent of what line ending type is used

### Steps

1. Have a `Cargo.toml` file with `\r\n` line endings
2. Run `cargo add serde`
3. See that `\n` line endings are mixed in

### Possible Solution(s)

Check if `\r\n` is present and then convert any stray `\n`s to `\r\n`

### Notes

See also #14857 as cargo-script has another area where this can be hit

### Version

```text

```

epage:
Somewhat related: #12897

ranger-ross:
I took a look at how `file` decides if a file has CRLF line terminators and it also appears to be using the simple approach of just [checking the number](https://github.com/file/file/blob/68cc5f974a36477dd507cd9eddd9457cbbefc1f7/src/ascmagic.c#L175) `/r/n` in the file is > 0.

I think the proposed solution if checking if the file contains at least one `\r\n` should work.
A small downside of this is that if a file has mixed line terminates (both `\n` and `\r\n`) Cargo would not be able to preserve both.
But this is probably a rare scenario and I don't see a usecase for having both. 

Fix no concerns with this approach, I could probably take a stab at implementing this

weihanglo:
FWIW, https://github.com/rust-lang/cargo/issues/12897 is a similar issue

PR #14983, opened 2024-12-26
Added logic to preserve carriage returns when updating the manifest file

ranger-ross:

This is an attempt to preserve `/r/n` line endings in `Cargo.toml` during `cargo add`/`remove` by checking if it already had at least one `/r/n`.

See #14863

### Additional information

For testing, it appears that snapbox will [normalize line ending](https://github.com/assert-rs/snapbox/blob/0de225bb3383cc8b9217fe664d16605e180b8062/crates/snapbox/src/filter/mod.rs#L36) so the tests will directly compare the file contents.


ranger-ross:
oops, looks like I broke a bunch of tests on windows. :sweat_smile: 
Let me look into why they started failing

src/cargo/util/toml_mut/manifest.rs
weihanglo, line 1: Overall this does what we want. However, I am not sure if this should be in toml_edit or in Cargo. Found one relevant comment from an issue: <https://github.com/toml-rs/toml/issues/752#issuecomment-2237732455>.

tests/testsuite/cargo_add/preserve_carriage_returns/mod.rs
weihanglo, line 1: This may be auto-converted, if `core.autocrlf = true` is set in git configuration on Linux. So the test always passes even without this patch. I haven't tested it locally though. Maybe we should have [a `.gitattributes` file](https://git-scm.com/docs/gitattributes) and set the file `eol=crlf`?

Slightly related, in [contributing doc](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request) we recommend splitting new tests into its own commit, and it still passes CI. Then followed by a fix and a test update. The test change in the second commit tell that that the bug is actually there, and the patch really fixes it.

src/cargo/util/toml_mut/manifest.rs
weihanglo: #14863 isn't marked as `S-accepted`. I would generally recommend discussing there, and leave this for implementation only dicussion.

ranger-ross: > isn't marked as S-accepted. I would generally recommend discussing there, and leave this for implementation only dicussion.

ah my bad :sweat_smile: I was looking through `C-bug` issues for small bugs to work on to learn/get some experience with the Cargo codebase. I'll be sure to stick to S-accepted or at least discuss further on the issue before starting/opening a PR.

ranger-ross:
I am going to close this PR as its stale and as mentioned above its not accepted and its not clear if this should be fixed in cargo or toml-rs

PR #14985, opened 2024-12-26
feat(package): implement vcs status cache

weihanglo:

Implement a VCS status cache for `cargo package`.

It may not be worthy for the majority. Single file `git status` is generally fast.
Also, it might only benefit whoever does `cargo package` for the entire workspace in one cargo invocation, such as `cargo package --workspace` or using `-Zpackage-workspace`.

On some slow file systems or lower-end machines,
skipping file system access might be helpful.

However, slow file system access usually happens on Window.
And we'll only have large amount of `git status` when
#14981 merges and the repo contains lots of symlinks.
Given symlink handling story is crazy in real world Windows

I didn't really benchmark it.
If any reviewer doesn't think it is worth the complexity,
feel free to close hits.

### Additional information

### DIFF
--- a/src/cargo/ops/cargo_package/mod.rs
+++ b/src/cargo/ops/cargo_package/mod.rs
@@ -226,6 +226,7 @@ fn do_package<'a>(
 
     // Packages need to be created in dependency order, because dependencies must
     // be added to our local overlay before we can create lockfiles that depend on them.
+    let mut vcs_info_builder = vcs::VcsInfoBuilder::new(ws, opts);
     let sorted_pkgs = deps.sort();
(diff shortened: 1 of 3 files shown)


epage:
So this only really helps us with source files outside of the package root?  Is this going to buy us enough gains to be worth the performance hit?

weihanglo:
> So this only really helps us with source files outside of the package root?  Is this going to buy us enough gains to be worth the performance hit?

Yes. It only helps in that situation and when having lots of workspace members sharing the same set of files-outside-package-root.

I lean toward closing this and then revive #14962, once #14981 has merged. The workspace inheritance issue won't be harder to fix with or without #14962

weihanglo:
Close as it is just a speculation. Could revisit when needed. At this moment for performance I favor something like #14997

Issue #5664, opened 2018-06-28
cargo package / publish don't handle symlink correctly on Windows


upsuper:
If you package a crate with symlink, the file would not be followed but be packed as a file with relative path as content. For example, if you checkout [serde-rs/serde](https://github.com/serde-rs/serde), and run `cargo package` in `serde` directory, the license files in the package directory would only have content like `../LICENSE-APACHE`.

It's probably actually a problem of Git for Windows which may not checkout symlink correctly. But maybe Cargo can try harder to query the VCS for whether it's a symlink or not (e.g. use `git ls-files -s`).

Related issues: alexcrichton/cargo-vendor#77, #3896, #2748.

alexcrichton:
According to https://stackoverflow.com/questions/5917249/git-symlinks-in-windows this may be an msys/MinGW thing. In that case this may be quite difficult to fix...

mati865:
Due to permission error (yes, one cannot simply create symlink without admin rights on Windows) `ln -s` was configured to do copy instead.
See [here](https://sourceforge.net/p/msys2/tickets/41/) how to disable it.

Tastaturtaste:
Since I lost some hours today to this problem I want to document what solved this problem for me here.
Basically just follow the instruction in [this stackoverflow answer](https://stackoverflow.com/questions/5917249/git-symbolic-links-in-windows/59761201#59761201). This will not fix symlinks that are already broken, though maybe another answer or comment there will be helpful.

I will outline the general steps here, too:

1. Enable developer mode or give your account symlink permissions
2. Enable symlinks in git
    - `git config core.symlinks` should return true in your repo
3. Make sure your global settings are not overwritten by your local settings

vallentin:
Using `git ls-tree HEAD [file-name]` for a specific file or just `git ls-files -s` for all files, then the mode (i.e. the first number) will be `120000` if a file is symlink

weihanglo:
Cargo  since 1.85 will warn if you uses git and git `core.symlinks` is `false`,
and some symlinks were checked out as plain files during packaging.

https://github.com/rust-lang/cargo/pull/14994

PR #14994, opened 2024-12-31
fix(package): warn if symlinks checked out as plain text files

weihanglo:

`cargo package` will warn users when git `core.symlinks` is `false`
and some symlinks were checked out as plain files during packaging.

Git config [`core.symlinks`] defaults to true when unset.
In git-for-windows (and git as well),
the config should be set to false explicitly when the repo was created,
if symlink support wasn't detected [^1].

We assume the config was always set at creation time and never changed.
So, if it is true, we don't bother users with any warning.

[^1]: https://github.com/git-for-windows/git/blob/f1241afcc7956918d5da33ef74abd9cbba369247/setup.c#L2394-L2403

[`core.symlinks`]: https://git-scm.com/docs/git-config#Documentation/git-config.txt-coresymlinks

CI passes.

This shares two commits 42dc4eff43ce480350979e07c8605e0f12983b79 and c8c8223c265ba8e7ea4cd29f829583ce786834f6 with #14981.

I didn't commit to fix all symlink issues all at once.
This PR demonstrates how we could leverage metadata in `PathEntry`.
Maybe later we can really follow plain-text symlinks and resolve the issue.

### Additional information

cc #5664

### DIFF
--- a/src/cargo/ops/cargo_package/vcs.rs
+++ b/src/cargo/ops/cargo_package/vcs.rs
@@ -9,6 +9,7 @@ use serde::Serialize;
 use tracing::debug;
 
 use crate::core::Package;
+use crate::sources::PathEntry;
 use crate::CargoResult;
 use crate::GlobalContext;
 
@@ -41,7 +42,7 @@ pub struct GitVcsInfo {
 #[tracing::instrument(skip_all)]
 pub fn check_repo_state(
     p: &Package,
-    src_files: &[PathBuf],
+    src_files: &[PathEntry],
     gctx: &GlobalContext,
     opts: &PackageOpts<'_>,
 ) -> CargoResult<Option<VcsInfo>> {
@@ -114,7 +115,7 @@ pub fn check_repo_state(
 fn git(
     pkg: &Package,
     gctx: &GlobalContext,
-    src_files: &[PathBuf],
+    src_files: &[PathEntry],
     repo: &git2::Repository,
     opts: &PackageOpts<'_>,
 ) -> CargoResult<Option<GitVcsInfo>> {
@@ -136,6 +137,7 @@ fn git(
     let mut dirty_src_files: Vec<_> = src_files
         .iter()
         .filter(|src_file| dirty_files.iter().any(|path| src_file.starts_with(path)))
+        .map(|p| p.as_ref())
         .chain(dirty_metadata_paths(pkg, repo)?.iter())
         .map(|path| {
             pathdiff::diff_paths(path, cwd)
--- a/src/cargo/ops/cargo_package/vcs.rs
+++ b/src/cargo/ops/cargo_package/vcs.rs
@@ -92,6 +92,8 @@ pub fn check_repo_state(
         return Ok(None);
     }
 
+    warn_symlink_checked_out_as_plain_text_file(gctx, src_files, &repo)?;
+
     debug!(
         "found (git) Cargo.toml at `{}` in workdir `{}`",
         path.display(),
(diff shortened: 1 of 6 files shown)


tests/testsuite/package.rs
```diff
@@ -7025,3 +7025,89 @@ src/main.rs
 "#]])
         .run();
 }
+
+#[cargo_test]
+fn git_core_symlinks_false() {
```
epage, line 7030: Can you do before/after commits for this test?

weihanglo: Thanks for the catch! Done.

src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -111,6 +113,50 @@ pub fn check_repo_state(
...
+    if core_symlinks_disabled && src_files.iter().any(|f| f.maybe_plain_text_symlink()) {
+        let mut shell = gctx.shell();
+        shell.warn(format_args!(
+            "symbolic links may be checked out as regular files for git repo at `{}`",
+            repo.workdir().unwrap().display(),
+        ))?;
+        shell.warn("this might cause `cargo package` to include incorrect or incomplete files")?;
+        shell.note("to avoid this, set the Git config `core.symlinks` to `true`")?;
+        #[cfg(windows)]
+        shell.note("and on Windows, enable the Developer Mode to support symlinks")?;
```
epage, line 154: These seem like they are one note, should we use a newline?

src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -111,6 +113,50 @@ pub fn check_repo_state(
...
+        }
+    };
+
+    if core_symlinks_disabled && src_files.iter().any(|f| f.maybe_plain_text_symlink()) {
+        let mut shell = gctx.shell();
+        shell.warn(format_args!(
+            "symbolic links may be checked out as regular files for git repo at `{}`",
+            repo.workdir().unwrap().display(),
+        ))?;
+        shell.warn("this might cause `cargo package` to include incorrect or incomplete files")?;
```
epage, line 151: These seem like they are one warning, should we use a newline?

src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -111,6 +113,50 @@ pub fn check_repo_state(
...
+    let core_symlinks_disabled = match repo.config().and_then(|c| c.get_bool("core.symlinks")) {
+        Ok(enabled) => !enabled,
+        Err(e) => {
+            debug!(
+                "no git config `core.symlinks` set for repo at `{}`: {e}",
+                repo.path().display(),
+            );
+            return Ok(());
+        }
+    };
```
epage, line 143: I'm not sure if this is an improvement but thought I'd share it in case you had thoughts:
```rust
if !repo.config().and_then(|c| c.get_bool("core.symlinks")).unwrap_or(true) {
    return Ok(());
}

File: src/cargo/ops/cargo_package/vcs.rs [hunk 5]
```diff
@@ -111,6 +113,50 @@ pub fn check_repo_state(
...
+
+    if core_symlinks_disabled && src_files.iter().any(|f| f.maybe_plain_text_symlink()) {
+        let mut shell = gctx.shell();
+        shell.warn(format_args!(
+            "symbolic links may be checked out as regular files for git repo at `{}`",
```
epage, line 148: What do you think of
```
"found symbolic links that may be checked out as regular files for git repo at `{}`",
```

weihanglo:
All suggestions are resolved, and also slightly tweaked the message, as the code path is not specific to `cargo package`. Let me know if it is desired.

epage (approved):

PR #14997, opened 2025-01-01
perf(cargo-package): match certain path prefix with pathspec

weihanglo:

An improvement https://github.com/rust-lang/cargo/issues/14955.
This revives #14962. See benchmark chart in <https://github.com/rust-lang/cargo/pull/14962#issuecomment-2555318312>.

`check_repo_state` checks the entire git repo status. This is usually fine if you have only a few packages in a workspace.

For huge monorepos, it may hit performance issues.

For example,
on awslabs/aws-sdk-rust@2cbd34d
the workspace has roughly 434 members to publish.
`git ls-files` reported us 204379 files in this Git repository. That means git may need to check status of all files 434 times. That would be `204379 * 434 = 88,700,486` checks!

Moreover, the current algorithm is finding the intersection of `PathSource::list_files` and `git status`.
It is an `O(n^2)` check.
Let's assume files are evenly distributed into each package, so roughly 470 files per package.
If we're unlucky to have some dirty files, say 100 files. We will have to do `470 * 100 = 47,000` times of path comparisons.

Even worse, because we `git status` everything in the repo, we'll have to it for all members,
even when those dirty files are not part of the current package in question. So it becomes `470 * 100 * 434 = 20,398,000`!

#### Solution

Instead of comparing with the status of the entire repository, this patch use the magic pathspec[1] to tell git only reports paths that match a certain path prefix.

This wouldn't help the `O(n^2)` algorithm,
but at least it won't check dirty files outside the current package. Also, we don't `git status` against entire git worktree/index anymore.

[1]: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec

Run this command against awslabs/aws-sdk-rust@2cbd34db53389aed1cafe4c56d23186ef5ff304a,
and see if it is getting better.

```
CARGO_LOG_PROFILE=1 cargor package --no-verify --offline --allow-dirty -p aws-sdk-accessanalyzer -p aws-sdk-apigateway
```

I've verified checksums of `.crate` files generated from master (d85d76199f89d7a312f308f6e37fc40f70508923) and this commit (3dabdcdd20d6382e7efa2c01833efa20944f3a79). They are the same.

### Additional information

There are some other alternatives, like making `PathSource::list_files` additionally reports dirty files. While we already have rooms to do it, this approach should be the most straightforward one at this moment.

Some other approaches like

* Switch to gitoxide (I tried and it didn't as good as expected. Maybe I did something wrong).
* A flag `--no-vcs` to skip vcs at all
* Improve the `O(n^2)` algorithm

### DIFF
--- a/src/cargo/ops/cargo_package/vcs.rs
+++ b/src/cargo/ops/cargo_package/vcs.rs
@@ -1,6 +1,7 @@
 //! Helpers to gather the VCS information for `cargo package`.
 
 use std::collections::HashSet;
+use std::path::Path;
 use std::path::PathBuf;
 
 use anyhow::Context as _;
@@ -173,7 +174,9 @@ fn git(
     // - ignored (in case the user has an `include` directive that
     //   conflicts with .gitignore).
     let mut dirty_files = Vec::new();
-    collect_statuses(repo, &mut dirty_files)?;
+    let pathspec = relative_pathspec(repo, pkg.root());
+    collect_statuses(repo, &[pathspec.as_str()], &mut dirty_files)?;
+
     // Include each submodule so that the error message can provide
     // specifically *which* files in a submodule are modified.
     status_submodules(repo, &mut dirty_files)?;
@@ -263,12 +266,18 @@ fn dirty_files_outside_pkg_root(
 }
 
 /// Helper to collect dirty statuses for a single repo.
-fn collect_statuses(repo: &git2::Repository, dirty_files: &mut Vec<PathBuf>) -> CargoResult<()> {
+fn collect_statuses(
+    repo: &git2::Repository,
+    pathspecs: &[&str],
+    dirty_files: &mut Vec<PathBuf>,
+) -> CargoResult<()> {
     let mut status_opts = git2::StatusOptions::new();
     // Exclude submodules, as they are being handled manually by recursing
     // into each one so that details about specific files can be
     // retrieved.
-    status_opts
+    pathspecs
+        .iter()
+        .fold(&mut status_opts, git2::StatusOptions::pathspec)
         .exclude_submodules(true)
         .include_ignored(true)
         .include_untracked(true);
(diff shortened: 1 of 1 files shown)


src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -173,7 +174,9 @@ fn git(
     // - ignored (in case the user has an `include` directive that
     //   conflicts with .gitignore).
     let mut dirty_files = Vec::new();
-    collect_statuses(repo, &mut dirty_files)?;
+    let pathspec = relative_pathspec(repo, pkg.root());
+    collect_statuses(repo, &[pathspec.as_str()], &mut dirty_files)?;
```
epage, line 178: > https://github.com/rust-lang/cargo/pull/14962 was closed because we found more bugs in cargo package, and https://github.com/rust-lang/cargo/pull/14962 could potentially make them even harder to fix. Two of them have been fixed so this is good to ship IMO with its own good.

My concern was about doing this in one collection but instead we have separate passes for files inside and files outside.  That works.

epage: That does mean we could have duplicate checks for the files outside but those should be on a different order of magnitude so it shouldn't be as much of an issue

src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -263,12 +266,18 @@ fn dirty_files_outside_pkg_root(
 }
 
 /// Helper to collect dirty statuses for a single repo.
-fn collect_statuses(repo: &git2::Repository, dirty_files: &mut Vec<PathBuf>) -> CargoResult<()> {
+fn collect_statuses(
+    repo: &git2::Repository,
+    pathspecs: &[&str],
+    dirty_files: &mut Vec<PathBuf>,
+) -> CargoResult<()> {
```
epage, line 273: The downside is if your workspace has a real manifest, rather than virtual.  It will end up querying everything.

Collecting statuses once across all packages would help with this (and the outside root files).  This was proposed in #14985 but the real workspace manifest case wasn't considered and it had some drawbacks.

epage (approved):

PR #15000, opened 2025-01-01
fix(schema): Correct and update the JSON Schema

### LINKED ISSUE #14999: Invalid JSON escaping in cargo-util-schemas
### Problem

https://github.com/rust-lang/cargo/pull/14683 introduces generation of schema for Cargo.toml. However, it's not a valid JSON.

### Steps

```shell
$ cat manifest.schema.json | jq
jq: parse error: Invalid numeric literal at line 989, column 105
```

### Possible Solution(s)

```patch
diff --git a/manifest.schema.json b/manifest.schema.json.1
index a7ea5bd..020be02 100644
--- a/manifest.schema.json
+++ b/manifest.schema.json.1
@@ -986,11 +986,11 @@
     "TomlDependency_for_String": {
       "anyOf": [
         {
-          "description": "In the simple format, only a version is specified, eg. `package = /"<version>/"`",
+          "description": "In the simple format, only a version is specified, eg. `package = \"<version>\"`",
           "type": "string"
         },
         {
-          "description": "The simple format is equivalent to a detailed dependency specifying only a version, eg. `package = { version = /"<version>/" }`",
+          "description": "The simple format is equivalent to a detailed dependency specifying only a version, eg. `package = { version = \"<version>\" }`",
           "allOf": [
             {
               "$ref": "#/definitions/TomlDetailedDependency_for_String"
```

### Notes

_No response_

### Version

```text

```

epage:

Fixes #14999

### DIFF
--- a/crates/cargo-util-schemas/src/manifest/mod.rs
+++ b/crates/cargo-util-schemas/src/manifest/mod.rs
@@ -1787,5 +1787,5 @@ pub struct UnresolvedError;
 fn dump_manifest_schema() {
     let schema = schemars::schema_for!(crate::manifest::TomlManifest);
     let dump = serde_json::to_string_pretty(&schema).unwrap();
-    snapbox::assert_data_eq!(dump, snapbox::file!("../../manifest.schema.json"));
+    snapbox::assert_data_eq!(dump, snapbox::file!("../../manifest.schema.json").raw());
 }
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -205,6 +205,13 @@ jobs:
     - name: Fetch smoke test
       run: ci/fetch-smoke-test.sh
 
+  schema:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - run: rustup update stable && rustup default stable
+    - run: cargo test -p cargo-util-schemas -F unstable-schema
+
   resolver:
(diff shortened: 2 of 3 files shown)


crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -1787,5 +1787,5 @@ pub struct UnresolvedError;
 fn dump_manifest_schema() {
     let schema = schemars::schema_for!(crate::manifest::TomlManifest);
     let dump = serde_json::to_string_pretty(&schema).unwrap();
-    snapbox::assert_data_eq!(dump, snapbox::file!("../../manifest.schema.json"));
+    snapbox::assert_data_eq!(dump, snapbox::file!("../../manifest.schema.json").raw());
```
weihanglo, line 1790: Should we add a round-trip check for serialization here?

Something like this?

```rust
serde_json::from_str::<crate::manifest::TomlManifest>(&std::fs::read_to_string("mainfest.schema.json").unwrap()).unwrap();
```

weihanglo (approved):
Feel free to either add a check or proceed.

crates/cargo-util-schemas/src/manifest/mod.rs
epage: I feel like that won't generally buy us much.  If we did anything, it should probably be checking the schema against valid manifests.

PR #15002, opened 2025-01-02
test(git): Clean up shallow fetch tests

epage:

When looking to add shallow fetch for the git CLI, I found it hard to tell what we test.  This tries to clean up the tests and make it easier to add new ones.  My hope is to later generalize the test bodies so we can more easily test any git backend in any combination of shallow or deep fetches.

This is part of #13285.

### DIFF
--- a/tests/testsuite/git_shallow.rs
+++ b/tests/testsuite/git_shallow.rs
@@ -4,22 +4,27 @@ use cargo_test_support::{basic_manifest, git, paths, project};
 
 use crate::git_gc::find_index;
 
+enum Backend {
+    Git2,
+    Gitoxide,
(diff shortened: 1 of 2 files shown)


tests/testsuite/git_shallow.rs
```diff
@@ -9,11 +9,29 @@ enum Backend {
     Gitoxide,
 }
 
+impl Backend {
+    fn to_arg(&self) -> &'static str {
```
weihanglo, line 13: nit: this method name seems too general? We may have features like checkout or status some day.

weihanglo (approved):
Thanks!

Issue #15001, opened 2025-01-02
Allow specifying pull request deps using the github PR link


joshka:
### Problem

From time to time I need to submit a PR to a lib and then want to rely on that PR in my local code.
Every time I do this I have to search the [docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#choice-of-commit) to work out the specific syntax that I need.
It would be nice if I could just use GitHub's PR syntax for this (or at the very least the error message should point me in the right direction)

### Proposed Solution

# Option 1. Support this as-is:

```toml
[dependencies]
foo = { version = "1", git = "https://github.com/bar/foo/pull/123" }
```

I understand why this might not be ideal even if it's the most developer friendly fix for this. So alternatively:

# Option 2. Provide a better error message for this scenario that provides the right syntax:

Current error message:
```
    Updating git repository `https://github.com/bar/foo/pull/123`
warning: spurious network error (3 tries remaining): unexpected http status code: 404; class=Http (34)
warning: spurious network error (2 tries remaining): unexpected http status code: 404; class=Http (34)
warning: spurious network error (1 tries remaining): unexpected http status code: 404; class=Http (34)
error: failed to get `foo` as a dependency of package `some-package v0.1.0 (/Users/joshka/local/some-package)`

Caused by:
  failed to load source for dependency `some-package`

Caused by:
  Unable to update https://github.com/bar/foo/pull/123

Caused by:
  failed to fetch into: /Users/joshka/.cargo/git/db/161-9e1cf949600a4ac3

Caused by:
  network failure seems to have happened
  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli

Caused by:
  unexpected http status code: 404; class=Http (34)
```

Expected error message (something like):
```
The git dependency for `foo` points at the pull request url. Fix this by changing to `foo = { version = "1" git = "https://github.com/bar/foo" rev = "refs/pull/123/head"
```

### Notes

I understand that there's other code forges than GitHub, but it's reasonable to fix the 99% case even when other cases are not fixed.

epage:
For Option 1, we'd be pulling github into our compatibility story which is not great.

Option 2 is more viable.  One question would be what hosting services (and enterprise deployments) should we support. 

As patches are likely what should be encouraged in these scenarios, we should make sure we support those as well with whatever we do. 

[...]

PR #15003, opened 2025-01-02
fix: add error message for github PR url in dep

joshka:
Prior to this, using a github PR URL would cause cargo to attempt to
fetch from an incorrect URL several times before failing.
Providing a github pull request url now fails with an error message
that shows how to fix the problem.

E.g.:
```toml
bar = { git = "https://github.com/foo/bar/pull/123" }
```
Now gives the following error message:
```
dependency (bar) specifies a GitHub pull request link. If you were trying to specify a specific github PR, replace the URL with the git URL (e.g. `git = "https://github.com/foo/bar.git"`) and add `rev = "refs/pull/123/head"` in the dependency declaration.
```

Fixes: #15001

<!--
Thanks for submitting a pull request 🎉! Here are some tips for you:

* If this is your first contribution, read "Cargo Contribution Guide" first:
  https://doc.crates.io/contrib/
* Run `cargo fmt --all` to format your code changes.
* Small commits and pull requests are always preferable and easy to review.
* If your idea is large and needs feedback from the community, read how:
  https://doc.crates.io/contrib/process/#working-on-large-features
* Cargo takes care of compatibility. Read our design principles:
  https://doc.crates.io/contrib/design.html
* When changing help text of cargo commands, follow the steps to generate docs:
  https://github.com/rust-lang/cargo/tree/master/src/doc#building-the-man-pages
* If your PR is not finished, set it as "draft" PR or add "WIP" in its title.
* It's ok to use the CI resources to test your PR, but please don't abuse them.

Passing a github PR issue url as the URL for a dependency causes cargo to attempt to incorrectly fetch from the wrong URL. Instead this PR makes that an error with meaningful next steps.

Unit test for this is added to the test suite which checks that the error message aligns with the expected value.

### Additional information

This could be expanded to other forges (Gitlab, Codeberg, etc.) fairly simply if someone was so inclined, but for now this solves the 99% problem.

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -2144,6 +2144,8 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
                 .unwrap_or(GitReference::DefaultBranch);
             let loc = git.into_url()?;
 
+            bail_if_github_pull_request(&name_in_toml, &loc)?;
+
             if let Some(fragment) = loc.fragment() {
                 let msg = format!(
                     "URL fragment `#{fragment}` in git URL is ignored for dependency ({name_in_toml}). \
@@ -2180,6 +2182,26 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
         }
         (None, None, None, None) => SourceId::crates_io(manifest_ctx.gctx),
     }
+}
+
+/// Checks if the URL is a GitHub pull request URL.
+///
+/// If the URL is a GitHub pull request URL, an error is returned with a message that explains
+/// how to specify a specific git revision.
+fn bail_if_github_pull_request(name_in_toml: &str, url: &Url) -> CargoResult<()> {
+    if url.host_str() != Some("github.com") {
+        return Ok(());
+    }
+    let path_components = url.path().split('/').collect::<Vec<_>>();
+    if let ["", owner, repo, "pull", pr_number, ..] = path_components[..] {
+        bail!(
+            "dependency ({name_in_toml}) specifies a GitHub pull request link. \
+                If you were trying to specify a specific github PR, replace the URL with the git \
+                URL (e.g. `git = \"https://github.com/{owner}/{repo}.git\"`) \
+                and add `rev = \"refs/pull/{pr_number}/head\"` in the dependency declaration.",
+        );
+    }
+    Ok(())
 }
 
 pub(crate) fn lookup_path_base<'a>(
--- a/tests/testsuite/bad_config.rs
+++ b/tests/testsuite/bad_config.rs
@@ -2152,6 +2152,37 @@ Caused by:
 }
 
 #[cargo_test]
+fn github_pull_request_url() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.0.0"
+                edition = "2015"
+                authors = []
+
+                [dependencies.bar]
+                git = "https://github.com/foo/bar/pull/123"
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("check -v")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
+
+Caused by:
+  dependency (bar) specifies a GitHub pull request link. If you were trying to specify a specific github PR, replace the URL with the git URL (e.g. `git = "https://github.com/foo/bar.git"`) and add `rev = "refs/pull/123/head"` in the dependency declaration.
+
+"#]])
+        .run();
+}
+
+#[cargo_test]
 fn fragment_in_git_url() {
     let p = project()
         .file(


tests/testsuite/bad_config.rs
```diff
@@ -2151,6 +2151,37 @@ Caused by:
         .run();
 }
 
+#[cargo_test]
+fn github_pull_request_url() {
```
epage, line 2155: As noted in #14941, this should cover patches as well and we should have a test for it.

src/cargo/util/toml/mod.rs
```diff
@@ -2144,6 +2144,8 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
                 .unwrap_or(GitReference::DefaultBranch);
             let loc = git.into_url()?;
 
+            bail_if_github_pull_request(&name_in_toml, &loc)?;
```
epage, line 2147: I'm a bit uncomfortable proactively erroring if the URL pattern matches to github, rather than providing context if what the user specified fails.  I guess the worst that can happen is github allows you to clone PR URLs and we people report it and we have to remove it to allow its use.  Not the worst outcome for something that is unclear if it would ever happen.

tests/testsuite/bad_config.rs
```diff
@@ -2151,6 +2151,37 @@ Caused by:
...
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("check -v")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
+
+Caused by:
+  dependency (bar) specifies a GitHub pull request link. If you were trying to specify a specific github PR, replace the URL with the git URL (e.g. `git = "https://github.com/foo/bar.git"`) and add `rev = "refs/pull/123/head"` in the dependency declaration.
```
epage, line 2179: As an error message, it should first saw what the problem is.  I also feel like this requires people to parse a lot of the message to understand the fix and apply it (e.g. rather than copy/pasting)

What do you think of
```
  `dependencies.{name}.git = "{url}"` is not a repository.
  hint: The repository path looks like a pull request, try replacing the repository path with:
    `git = "{url}", rev = "{rev}"`
```

tests/testsuite/bad_config.rs
weihanglo: #15001 you mean?

epage: Yes :)

tests/testsuite/bad_config.rs
joshka: The first part is for consistency with the other error messages. How about:
```
dependency (bar) git url https://github.com/foo/bar/pull/123 is not a repository.
The path looks like a pull request.
Try replacing the dependency with: `git = "https://github.com/foo/bar.git" rev = "refs/pull/123/head"` in the dependency declaration.
```

tests/testsuite/bad_config.rs
joshka: Done in d743987ca

src/cargo/util/toml/mod.rs
joshka: The behavior here is the same as for how fragments are handled directly below this (though they are handled as a warning. The difference is that this causes an error later in the processing while the fragment does not. I think it's reasonable to avoid attempting to do cause an error by downloading. If this was wrong, then removing this code or changing it to a warning would be fairly low cost. How about we keep this as-is?

weihanglo: A warning is safer if we are unsure about the effect of a change. I would vote for turning into a warning for now :)

tests/testsuite/bad_config.rs
weihanglo: I am fine with keeping it consistent with existing messages. We could tweak them later altogether when needed.

src/cargo/util/toml/mod.rs
joshka: Will fix

joshka: Done

weihanglo (approved):
Thanks! If needed, we can iterate it later :)

joshka:
I'm not quite sure what the right approach to getting the tests to succeed both in CI and locally for this change. Any guidance?

tests/testsuite/bad_config.rs
```diff
@@ -2151,6 +2151,56 @@ Caused by:
...
+Caused by:
+  failed to clone into: [ROOT]/home/.cargo/git/db/123-[HASH]
+
+Caused by:
+  network failure seems to have happened
+  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
+  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli
+
+Caused by:
+  unexpected http status code: 404; class=Http (34)
```
weihanglo, line 2198: Interesting. I thought it was network flakiness. Anyway, you can use `...` to omit the unimportant part.

```suggestion
...
[ERROR] failed to get `bar` as a dependency of package `foo v0.0.0 ([ROOT]/foo)`

Caused by:
  failed to load source for dependency `bar`

Caused by:
  Unable to update https://github.com/foo/bar/pull/123
...
```

epage:
Sorry I've not been able to get back to this and thanks @weihanglo for picking it up.

A non-blocking thought (since some improvement is better than no improvement) is to offer something similar for `cargo add --git URL`

tests/testsuite/bad_config.rs
weihanglo: `[patch]` test also needs the same trick :)

motorailgun:
@joshka Sorry for heads-up, wondering if you're still working on this? Would you mind if I take over this PR?

joshka:
> @joshka Sorry for heads-up, wondering if you're still working on this? Would you mind if I take over this PR?

Please do. I've been busy with life and not as available to get things like this over the line.

weihanglo:
completed in https://github.com/rust-lang/cargo/pull/16207, closing

Issue #13828, opened 2024-04-29
cargo doesn't include test = false `[[test]]`s in --tests or --all-targets


vojtechkral:
### Problem

A test marked as `test = false` is supposed to not be run when invoking `cargo test`.

This works, however, not only is the test not run, it's not built or checked (when using `cargo check`) either, not even with `--tests` or `--all-targets` passed.

As a workaround, I'm currently using `--tests \*` instead of `--tests` to actually include all tests including `test = false` ones.

### Steps

1. Mark a test with `test = false` in `Cargo.toml`
2. Run `cargo check --tests`
3. The test is not checked.
4. Run `cargo check --test \*`
5. The test is checked.

### Version

```text
cargo 1.77.2 (e52e36006 2024-03-26)
release: 1.77.2
commit-hash: e52e360061cacbbeac79f7f1215a7a90b6f08442
commit-date: 2024-03-26
host: aarch64-apple-darwin
libgit2: 1.7.2 (sys:0.18.2 vendored)
libcurl: 8.4.0 (sys:0.4.70+curl-8.5.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Mac OS 14.4.1 [64-bit]
```
```

vojtechkral:
Yeah, `test = false` tests are a bit of an oddball. They sort of are and aren't tests.

Wdyt about having `--tests` build these "non-tests" but not run them?

The full context is here [using cargo-fixture](https://github.com/vojtechkral/cargo-fixture).

weihanglo:
>  I'm assuming it would be safe to update it to include all targets.

Hmm… I can't see it. Changing that will include Cargo targets not supposed to build, and is a breaking change for some existing scripts relying on the current behavior.

epage:
Yes, there is a risk that people would have `test = false` programs that can't compile in all situations they run `cargo check --all-targets` and may not even run when they do `cargo test --test bad`.  The question is whether `--all-targets` behavior is a bug (which I feel like it is even if the docs don't quite say it) and whether that is enough of a corner case to be concerned with.

epage:
Interesting, the behavior for `test = false` is different between bins and integration tests, see also #8338

Issue #13668, opened 2024-03-29
Cargo runs unit tests when they are disabled and `--lib` is passed


RalfJung:
### Problem

When a crate has no unit tests (via `test = false` in its manifest), I would expect cargo to never ever try to run unit tests for that crate.

And yet it turns out that `cargo test --lib` will run the unit tests.

### Steps

1. Run `cargo test --lib --bins --examples --tests --benches` (in an attempt to run "everything except for doc tests"), or just `cargo test --lib`
2. Note how even if the library has unit tests are disabled, cargo will run them

```
     Running unittests src/lib.rs (build/x86_64-unknown-linux-gnu/stage0-std/x86_64-unknown-linux-gnu/release/deps/core-04ceb2ced8c36de5)

running 0 tests


test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 7.54µs
```

### Possible Solution(s)

`--lib` should just be a filter, removing from the usual test set everything that's not a "library unit test". It should not add things that were not present to begin with.

### Notes

This is related to but different from https://github.com/rust-lang/cargo/issues/10936

RalfJung:
Something similar seems to happen with benchmarks. With a manifest like this
```
[[bench]]
name = "allocbenches"
path = "benches/lib.rs"
test = true

[[bench]]
name = "vec_deque_append_bench"
path = "benches/vec_deque_append.rs"
harness = false
```
Running `cargo test` will only run the first benchmark, but running `cargo test --lib --bins --examples --tests --benches` will run both.

weihanglo:
This seems to be covered by <https://github.com/rust-lang/testing-devex-team/issues/1>> Once we have a new JSON output format, Cargo could get more controls over tests than relying on libtest.

RalfJung:
I don't think this issue has anything to do with handling the output from the tests. It's all about which tests cargo decides to run.

weihanglo:
I see. Sorry I understood it wrong.

`--lib` is a flag for user to explicitly requests tests in library target to be run, so that is expected. So does other flags like `--test`, `--bench`, etc.

`test = false` controls only the default set of tests being run. See [the target selection section](https://doc.rust-lang.org/nightly/cargo/commands/cargo-test.html#target-selection).

> Setting targets to `test = false` will stop them from being tested by default. Target selection options that take a target by name (such as `--example` foo) ignore the test flag and will always test the given target.

weihanglo:
[...]

Issue #10958, opened 2022-08-09
Provide an option to `cargo check` everything except benchmarks


RalfJung:
IDE plugins like RA like to invoke `cargo check` with `--all-targets` to ensure that, for example, examples are checked as well, not just the main crate sources. However, this often fails on stable since it will also check benchmarks, which essentially always need nightly features (Cc https://github.com/crossbeam-rs/crossbeam/issues/890).

So I thought I could work around this by telling RA to run `"cargo" "check" "--lib" "--bins" "--tests" "--examples" "--workspace" "--message-format=json"` instead -- however, this command sadly will fail when there is no lib target. Is there a way to say "please check everything except for the benchmarks"?

The documentation claims that `--all-targets` is equivalent to specifying `--lib --bins --tests --benches --examples`, but that is wrong: `--lib --bins --tests --benches --examples` will fail when there is no lib target, but `--all-targets` will still work.

RalfJung:
E.g. there could be a `--libs` flag that selects all the libs targets that exist -- it's going to be either 0 or 1, but that still leaves two options, and cargo currently doesn't have a uniform way to deal with this situation.

I'd rather not have to manually configure RA for each and every package and then remember to change the configuration when a libs crate is added to a package.

weihanglo:
Sorry for the subtlety between these flags. `--tests` implies lib target is included as well, so I believe `cargo check --tests` is effectively the same as `cargo check --lib --tests` except [checking the existence of lib target](https://github.com/rust-lang/cargo/blob/e6827eee9b6720a1bd69c6115cac22f8125b24aa/src/cargo/ops/cargo_compile.rs#L1132-L1139). Does `cargo check --bins --tests --examples` work for you?

RalfJung:
What if my Cargo.toml contains
```
[lib]
test = false
```
will `cargo check --tests` then still also check the lib? And is that a stable guarantee I can rely on? The documentation doesn't say anything like this (as far as I can see).

RalfJung:
FWIW for now I am simply sticking to nightly rustc for crates where `cargo check --all-targets` does not work on stable. But at least the documentation issue in `--all-targets`  should still be tracked.

weihanglo:
> What if my Cargo.toml contains
> 
> ```
> [lib]
> test = false
> ```
> 
> will `cargo check --tests` then still also check the lib? 

[...]

PR #15007, opened 2025-01-02
Conditionally mark the `test` cfg as a well known cfg

Urgau:

This PR conditionally mark the `test` cfg as a well known cfg depending on the target unit "test" field (ie `lib.test = false`, `[[bin]] test = false` and others).

This is related to https://github.com/rust-lang/rust/issues/117778 and https://users.rust-lang.org/t/cargo-what-is-the-purpose-of-lib-test-false/102361.

When defining `lib.test = false` (and others), any use of `cfg(test)` will trigger the `unexpected_cfgs` lint. 
```toml
[lib]
test = false  # will now warn on cfg(test) 
```

Best reviewed commit by commit. Second commit removes the `test` cfg from the `--check-cfg` args.

### Additional information

T-compiler [MCP#785](https://github.com/rust-lang/compiler-team/issues/785) and https://github.com/rust-lang/cargo/pull/14963 were of preparatory work.

r? @epage

### DIFF
--- a/tests/testsuite/check_cfg.rs
+++ b/tests/testsuite/check_cfg.rs
@@ -318,6 +318,127 @@ fn well_known_names_values_doctest() {
 }
 
 #[cargo_test]
+fn test_false_lib() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.1.0"
+                edition = "2018"
+
+                [lib]
+                test = false
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("check -v")
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test"))
+        .run();
+
+    p.cargo("clean").run();
+    p.cargo("test -v")
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test"))
+        .run();
+
+    p.cargo("clean").run();
+    p.cargo("test --lib -v")
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test"))
+        .run();
+}
+
+#[cargo_test]
+fn test_false_bins() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.1.0"
+                edition = "2018"
+
+                [[bin]]
+                name = "daemon"
+                test = false
+                path = "src/deamon.rs"
+            "#,
+        )
+        .file("src/main.rs", "fn main() {}")
+        .file("src/deamon.rs", "fn main() {}")
+        .build();
+
+    p.cargo("check -v")
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test")) // for foo & deamon
+        .run();
+}
+
+#[cargo_test]
+fn test_false_examples() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.1.0"
+                edition = "2018"
+
+                [lib]
+                test = false
+
+                [[example]]
+                name = "daemon"
+                test = false
+                path = "src/deamon.rs"
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .file("src/deamon.rs", "fn main() {}")
+        .build();
+
+    p.cargo("check --examples -v")
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test"))
+        .run();
+}
+
+#[cargo_test(nightly, reason = "bench is nightly")]
+fn test_false_benches() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.0.0"
+                edition = "2018"
+
+                [[bench]]
+                name = "ben1"
+                test = false
+                path = "benches/ben1.rs"
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .file(
+            "benches/ben1.rs",
+            r#"
+            #![feature(test)]
+            extern crate test;
+            #[bench] fn run1(_ben: &mut test::Bencher) { }
+            "#,
+        )
+        .build();
+
+    p.cargo("bench --bench ben1 -v")
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test"))
+        .run();
+}
+
+#[cargo_test]
 fn features_doc() {
     let p = project()
         .file(
--- a/tests/testsuite/check_cfg.rs
+++ b/tests/testsuite/check_cfg.rs
@@ -336,17 +336,18 @@ fn test_false_lib() {
         .build();
 
     p.cargo("check -v")
-        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test"))
+        .with_stderr_does_not_contain(x!("rustc" => "cfg" of "docsrs,test"))
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs"))
         .run();
 
     p.cargo("clean").run();
     p.cargo("test -v")
-        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test"))
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs"))
         .run();
 
     p.cargo("clean").run();
     p.cargo("test --lib -v")
-        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test"))
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs"))
         .run();
 }
 
@@ -372,7 +373,8 @@ fn test_false_bins() {
         .build();
 
     p.cargo("check -v")
-        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test")) // for foo & deamon
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test")) // for foo
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs")) // for deamon
         .run();
 }
 
@@ -401,7 +403,8 @@ fn test_false_examples() {
         .build();
 
     p.cargo("check --examples -v")
-        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test"))
+        .with_stderr_does_not_contain(x!("rustc" => "cfg" of "docsrs,test"))
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs"))
         .run();
 }
 
--- a/tests/testsuite/check_cfg.rs
+++ b/tests/testsuite/check_cfg.rs
@@ -317,7 +317,7 @@ fn well_known_names_values_doctest() {
         .run();
 }
 
-#[cargo_test]
+#[cargo_test(nightly, reason = "warning currently only on nightly")]
 fn test_false_lib() {
     let p = project()
         .file(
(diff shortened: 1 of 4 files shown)


tests/testsuite/check_cfg.rs
```diff
@@ -336,17 +336,18 @@ fn test_false_lib() {
         .build();
 
     p.cargo("check -v")
-        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs,test"))
+        .with_stderr_does_not_contain(x!("rustc" => "cfg" of "docsrs,test"))
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs"))
```
epage, line 340: One way this test would silently keep passing is if we split the calls to `--check-cfg` so its `--check-cfg cfg(docsrs) --check-cfg cfg(test)`

Would really prefer testing the desired end-user result but that would mean checking for more rustc warnings.

epage: More generally, I'm not thrilled with these tests assertions.  They do not communicate what the expected behavior is.  This reads as if `--cfg docsrs` is being passed in.  They also rely on brittle `contains` checking which we try to avoid.

Urgau: I agree checking for those arguments is a bit brittle and the macro could use some rework (or better removal).

I will include some partial warning checking (as to not depend on the exact warning output which changes often for check-cfg/`unexpected_cfgs`). Just waiting for https://github.com/rust-lang/rust/pull/131729 to hit nightly.

Urgau: I've added the partial stderr checking for checking that the warning(s) are indeed emitted.

I have added those as the separate commit. Let me know if it isn't sufficient.

tests/testsuite/check_cfg.rs
```diff
@@ -317,6 +317,174 @@ fn well_known_names_values_doctest() {
...
+        .file("src/lib.rs", "#[cfg(test)] mod tests {}")
+        .build();
+
+    p.cargo("check -v")
+        .with_stderr_does_not_contain(x!("rustc" => "cfg" of "docsrs,test"))
+        .with_stderr_contains(x!("rustc" => "cfg" of "docsrs"))
+        .with_stderr_data(str![[r#"
+...
+[WARNING] unexpected `cfg` condition name: `test`
+ --> src/lib.rs:1:7
```
epage, line 344: We should elide the body of the message to reduce the risk of rustc changing things and being blocked on cargo

tests/testsuite/check_cfg.rs
```diff
@@ -332,26 +332,57 @@ fn test_false_lib() {
...
+
+"#]])
         .run();
 
     p.cargo("clean").run();
     p.cargo("test --lib -v")
         .with_stderr_contains(x!("rustc" => "cfg" of "docsrs"))
+        .with_stderr_data(str![[r#"
+...
+[WARNING] unexpected `cfg` condition name: `test`
```
weihanglo, line 373: If user has specified `--lib`, doesn't it mean that they requested to test the lib crate so `cfg(test)` is expected to be there?

tests/testsuite/test.rs
```diff
@@ -4191,7 +4191,19 @@ fn test_hint_workspace_virtual() {
         .file("a/src/lib.rs", "#[test] fn t1() {}")
         .file("b/Cargo.toml", &basic_manifest("b", "0.1.0"))
         .file("b/src/lib.rs", "#[test] fn t1() {assert!(false)}")
-        .file("c/Cargo.toml", &basic_manifest("c", "0.1.0"))
+        .file(
```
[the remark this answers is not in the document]
Urgau: It's a bit tricky because of what the test is trying to test but in nutshell the example "ex1" contains a `#[test]` and to not get the warning it must have `test = true` so that's what I did, since by default for examples `test = false`.

tests/testsuite/check_cfg.rs
Urgau: What the command line arguments are shouldn't matter, otherwise users would get a different behavior between a "normal" `cargo test` and one with arguments like `--all-targets`.

weihanglo: However, `lib.test = false` doesn't mean there is no test. It, as of today, is defined as "test" won't run by default. `cfg(test)` is still relevant regardless if a test is in the default test set.

Maybe I should put the question this way: Why is `cfg(test)` in a `test=false` crate unexpected?

Urgau: IMO having `test = false` meaning not tested by default compared to not tests at all is unexpected, at least before reading the documentation I would have assumed that it disabled all the testing.

@jplatte made a further argument a in his [comment](https://github.com/rust-lang/rust/issues/117778#issuecomment-1805587670):
> Hm, maybe. I guess it depends on what the intention of `lib.test = false` is. I think the vast majority of projects that set it don't have unit tests at all, and if any were to be introduced, getting a warning would be valuable (whether it results in the test author removing the flag or moving the test code into unit tests). I wonder if there are _any_ projects that use `lib.test = false` while having unit tests, intentionally.

weihanglo: Thanks for linking the thread here. Yeah I've read that, and agree that `lib.test = false` largely implies no test in lib crate at all. That is a valid use case and deserves a lint or warning.

However, fields under Cargo targets are mostly about including in or excluding from the default crate set, and is largely affected by [the command-line argument](https://doc.rust-lang.org/nightly/cargo/commands/cargo-test.html#target-selection). If Cargo has started emitting unexpected warning for Cargo targets, it may also bite users that disabling tests for other reasons. For example, I've seen people setting `test=false` because some tests are not expected to run by default for daily development, but only on CI or other special environments.

While that kind of scenario I guess is far less than who don't want tests at all, it is still a valid use case aligning with what the current doc describes. Treating `unexpected_cfgs` as a indirect way of banning tests in a crate doesn't seem to be the best venue. It is more like a hack because there is no lint for really banning tests.

That being said, I am not surprised if Cargo targets settings doesn't meet people's expectation. It is always a source of confusions[^1] 😞.


[^1]: I can actually name a lot of issues here. Let's just have a little peek of them:
		  * https://github.com/rust-lang/cargo/issues/8338
		  * https://github.com/rust-lang/cargo/issues/10936
		  * https://github.com/rust-lang/cargo/issues/10958
		  * https://github.com/rust-lang/cargo/issues/13668
		  * https://github.com/rust-lang/cargo/issues/13437
		  * https://github.com/rust-lang/cargo/issues/13828

hanna-kruppe: The current behavior is also unexpected and undesirable to me, I agree with the quote and linked comment. I've been keeping an eye on this series of PRs (thanks, @Urgau!) because I'd like to start using `lib.test = false` for crates that don't have unit tests without risking the silent failure mode mentioned by @jplatte.

For tests that shouldn't run *by default*, `#[ignore]` is a more discoverable and more robust choice. Ignored tests are at least called out as such in libtest output, as long as the test binary containing them is built and run. In contrast, `cfg(test)` code that is never compiled or run at all is mostly invisible. Even if you get in the habit of `cargo test --all-targets`, you'll have a similar problem with doctests not being run any more! Examples with `#[test]`s have this problem by default, which [the documentation highlights as a pitfall to avoid](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-test-field). The unexpected-cfg warning added by this PR also calls attention to cases where this has been missed.

If any project intentionally uses `test = false` for crates with unit tests in them, they can still turn off the unexpected-cfg warning by adding `cfg(test)` to the allowlist in `[lints.rust]` or build script output. So it's not like this change would make it fundamentally harder to use `test = false` for tests not built by default -- it just adds the ability to distinguish that usage from "not intended to contain any tests at all".

weihanglo: To clarify a bit, the proposed behavior may be worth for `lib.test = false`, as it often implies no unit tests. For `test.test = false` it’s a bit harder to say. We could diverge the logic for different target kinds, but I think nobody wants another confusions added to Cargo targets 😓.

weihanglo: >  it just adds the ability to distinguish that usage from "not intended to contain any tests at all".

My argument is that this may deserve an `unexpected_tests` lint. Relying on `unexpected_cfgs ` only partially resolves the issue because people can still have `#[test]` without any `cfg(test)`.

hanna-kruppe: Yes, it's a somewhat indirect way of achieving the goal, and it would be nice to also warn on bare `#[test]` not nested under an explicit `cfg(test)`. Clippy [has a lint for that](https://rust-lang.github.io/rust-clippy/rust-1.83.0/index.html#/tests_outside_test_module), but ironically it only fires when you run `cargo clippy -- --test` because otherwise these items are stripped out before Clippy can see them. For the same reason, a dedicated `unexpected_tests` lint would also have to warn on any occurrence of `cfg(... test ...)`, including nested inside complex `cfg` expressions, just like `unexpected_cfgs` would after this PR. I'd also expect a warning for `#[cfg_attr(... test ..., blah)]` in a crate where tests are "unexpected". So I guess `unexpected_tests` would be identical to `unexpected_cfgs` plus warning on `#[test]`?

But it also wouldn't be a bad solution to just treat bare `#[test]` as if it came with an implied `cfg(test)` and emit the same diagnostic if `cfg(test)` is unexpected. I don't know if that's implementable without unduly complicating `unexpected_cfgs` and/or `#[test]` expansion, but it seems semantically appropriate: it matches how `#[test]` is expanded, except for the subtle difference between `rustc --test` and `rustc --cfg test` (that nobody should be relying on). Tailored diagnostics that talk about unit tests in particular can be achieved without an entire new lint. The main thing that's lost is the ability to toggle the lint independently of other `unexpected_cfgs` warnings, but that seems strange when `cfg(... test ...)` is part of what the lint would warn about.

Urgau: The `unexpected_cfgs` lint do trigger on `#[test]` but only when passed with `--test`.

<details>

```text
$ rustc +nightly --crate-type=lib --test --check-cfg='cfg()' src/lib.rs
warning: unexpected `cfg` condition name: `test`
 --> src/lib.rs:9:5
  |
9 |     #[test]
  |     ^^^^^^^
  |
  = help: expected names are: `clippy`, `debug_assertions`, `doc`, `doctest`, `fmt_debug`, `miri`, `over
flow_checks`, `panic`, `proc_macro`, `relocation_model`, `rustfmt`, `sanitize`, `sanitizer_cfi_generaliz
e_pointers`, `sanitizer_cfi_normalize_integers`, `target_abi`, `target_arch`, `target_endian`, `target_e
nv`, `target_family`, `target_feature`, `target_has_atomic`, `target_has_atomic_equal_alignment`, `target_has_atomic_load_store`, `target_os`, `target_pointer_width`, `target_thread_local`, `target_vendor`, `ub_checks`, `unix`, and `windows`
  = note: using a cfg inside a attribute macro will use the cfgs from the destination crate and not the ones from the defining crate
  = help: try referring to `test` crate for guidance on how handle this unexpected cfg
  = help: to expect this configuration use `--check-cfg=cfg(test)`
  = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg.html> for more information about checking conditional configuration
  = note: `#[warn(unexpected_cfgs)]` on by default
  = note: this warning originates in the attribute macro `test` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: 1 warning emitted
```

</details>

epage: Unless I missed something in catching up on this, I feel like we'd get more out of `test = false` being consistent, independent of the build-target type.  I'm not too concerned about false positives from people who have `#[test]` but set `test = false`.

weihanglo: My main point is the meaning of `test = false` will change after this PR. I don't mind shipping the particular change. Just want to make sure people aware of and feel good with it. And probably need doc update as well.

epage:
Talked it over with @weihanglo.  While this is an observable change in behavior it seems relatively innocuous and could likely be revisited if needed.

epage (approved):

tests/testsuite/check_cfg.rs
kpreid: There are multiple ways which a target may be tested even when it has `test = false`, notably, `cargo test --all-targets` and `cargo test some_test_name_filter`. Therefore, I believe this change is problematic (and have filed #15131 for tracking of that).

PR #15013, opened 2025-01-04
Make --allow-dirty imply --allow-staged

### LINKED ISSUE #14176: cargo-fix: --allow-dirty to imply --allow-staged?
### Problem

`cargo fix`/`clippy --fix` don't modify git's staging area, so the requirement to specify `--allow-staged` if there are any staged files is very cautious. Technically the staged files are not stored directly in the working directory, and can't be overwritten by the fix commands.

OTOH "dirty" state is much more volatile, and potentially has file changes in the working directory that haven't been saved anywhere, and which could be irreversibly lost when files are modified. This makes `--allow-dirty` a much stronger option that implies the local state is unimportant and can be lost.

This makes use of `--allow-dirty` *without* `--allow-staged` a weird edge case. It says the local changes are unimportant and can be overwritten even if they're unrecoverable, except being extra cautious about existence of local changes that are safe, and won't be overwritten. It's like having a `--force-overwrite` flag, but requiring `--allow-not-forced-non-overwrite` flag too.

### Proposed Solution

Having to add both `--allow-dirty --allow-staged` flags to the `fix` commands is verbose and a bit tedious to type. The extra `--allow-staged` flag could be avoided if the more destructive `--allow-dirty` also implied the lesser `--allow-staged`. Alternatively, maybe there could be a single option like `--force`/`-f` that implies both.

### Notes

_No response_

kornelski:
Staged changes don't really need protecting, and `--allow-dirty` is stronger than `--allow-staged`, so it can imply `--allow-staged` to make usage of `cargo fix` less verbose.

Closes #14176

### DIFF
--- a/src/bin/cargo/commands/fix.rs
+++ b/src/bin/cargo/commands/fix.rs
@@ -21,7 +21,7 @@ pub fn cli() -> Command {
         ))
         .arg(flag(
             "allow-dirty",
-            "Fix code even if the working directory is dirty",
+            "Fix code even if the working directory is dirty or has staged changes",
         ))
         .arg(flag(
             "allow-staged",
@@ -86,6 +86,8 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
         opts.filter = ops::CompileFilter::new_all_targets();
     }
 
+    let allow_dirty = args.flag("allow-dirty");
+
     ops::fix(
         gctx,
         &ws,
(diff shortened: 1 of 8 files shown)


epage:
As a reminder, the Cargo team's policy is that [Issues need to be S-accepted before moving onto PRs](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo)

kornelski:
I don't know what else I can do to move the issue forward. My feature request hasn't been accepted nor rejected for months.

weihanglo (approved):
https://github.com/rust-lang/cargo/issues/14176#issuecomment-2571346417


We've got a quorum for the  FCP, and I don't think this is significant enough we need to wait for 10 days. So, I'm going to merge this now. Thanks for the contribution!

PR #15014, opened 2025-01-04
feat: Added warning when failing to update index cache

### LINKED ISSUE #13712: Cargo updates registry index on every command with bad permissions
For a while now I've seen "Updating crates.io index" before every build in the rust-lang/rust repository on one of my computers. I got a chance to investigate today and it looks like that's because of this error:

```
$ CARGO_LOG=info ./x.py build --stage 0 library/std
Building bootstrap
    Finished `dev` profile [unoptimized] target(s) in 0.03s
Building stage0 library artifacts {std} (x86_64-unknown-linux-gnu)
    Updating crates.io index
   0.424170519s  INFO main:exec:compile_ws:create_bcx:resolve_with_registry:resolve_with_previous:resolve: cargo::sources::registry::index::cache: failed to write cache: Permission denied (os error 13) cache_path="/home/mark/.cargo/registry/index/index.crates.io-6f17d22bba15001f/.cache/r-/ef/r-efi"
    Finished `release` profile [optimized + debuginfo] target(s) in 0.67s
```

That error is because a bunch of the files in the cache directory ended up owned by root rather than my user, probably as a result of a docker build of test-various (see [lockfile](https://github.com/rust-lang/rust/blob/master/src/ci/docker/host-x86_64/test-various/uefi_qemu_test/Cargo.lock) which contains this crate).

I'm filing this because it seems like we ought to:

* Not silently ignore this error if it's going to cause us to "Updating..." every build
* Potentially delete/prune the cache (not sure if possible) rather than failing in this way

In my case `r-/ef/r-efi` where all owned by root.

ranger-ross:

Fixes #13712 

Adds a warning if there is an error updating the index cache. 
It also attempts to avoid warning spam as requested in [this comment](https://github.com/rust-lang/cargo/issues/13712#issuecomment-2514991597)

Below is an example output

```
    Updating crates.io index
warning: failed to write cache, path: /home/ross/.cargo/registry/index/index.crates.io-1949cf8c6b5b557f/.cache/ba/se/base64, error: Permission denied (os error 13)
   Compiling serde v1.0.217
   Compiling r-efi v5.2.0
   Compiling base64 v0.22.1
   Compiling cargo-13712 v0.1.0 (/home/ross/projects/cargo-13712)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.20s

```

I tested this on my machine by manually changing the owner/permission of the index files 
```sh
sudo chown root ~/.cargo/registry/index/.../.cache/r-
sudo chmod 700 ~/.cargo/registry/index/.../.cache/r-

# in a project that uses the `r-efi` crate
cargo build 
```

I wasn't quiet sure about the best way to add a test to the testsuite for this. :sweat_smile: 
If there is good way to create a test for this lmk

### DIFF
--- a/src/cargo/sources/registry/index/cache.rs
+++ b/src/cargo/sources/registry/index/cache.rs
@@ -65,6 +65,7 @@
 //! [`IndexSummary::parse`]: super::IndexSummary::parse
 //! [`RemoteRegistry`]: crate::sources::registry::remote::RemoteRegistry
 
+use std::cell::RefCell;
 use std::fs;
 use std::io;
 use std::path::PathBuf;
@@ -226,6 +227,9 @@ pub struct CacheManager<'gctx> {
     cache_root: Filesystem,
     /// [`GlobalContext`] reference for convenience.
     gctx: &'gctx GlobalContext,
+    /// Keeps track of if we have sent a warning message if there was an error updating the cache.
+    /// The motivation is to avoid warning spam if the cache is not writable.
+    has_warned: RefCell<bool>,
 }
 
 impl<'gctx> CacheManager<'gctx> {
@@ -233,7 +237,11 @@ impl<'gctx> CacheManager<'gctx> {
     ///
     /// `root` --- The root path where caches are located.
     pub fn new(cache_root: Filesystem, gctx: &'gctx GlobalContext) -> CacheManager<'gctx> {
-        CacheManager { cache_root, gctx }
+        CacheManager {
+            cache_root,
+            gctx,
+            has_warned: Default::default(),
+        }
     }
 
     /// Gets the cache associated with the key.
@@ -251,14 +259,26 @@ impl<'gctx> CacheManager<'gctx> {
     /// Associates the value with the key.
     pub fn put(&self, key: &str, value: &[u8]) {
         let cache_path = &self.cache_path(key);
-        if fs::create_dir_all(cache_path.parent().unwrap()).is_ok() {
-            let path = Filesystem::new(cache_path.clone());
-            self.gctx
-                .assert_package_cache_locked(CacheLockMode::DownloadExclusive, &path);
-            if let Err(e) = fs::write(cache_path, value) {
-                tracing::info!(?cache_path, "failed to write cache: {e}");
+        if let Err(e) = self.put_inner(cache_path, value) {
+            tracing::info!(?cache_path, "failed to write cache: {e}");
+
+            if !*self.has_warned.borrow() {
+                let _ = self.gctx.shell().warn(format!(
+                    "failed to write cache, path: {}, error: {e}",
+                    cache_path.to_str().unwrap_or_default()
+                ));
+                *self.has_warned.borrow_mut() = true;
             }
         }
+    }
+
+    fn put_inner(&self, cache_path: &PathBuf, value: &[u8]) -> std::io::Result<()> {
+        fs::create_dir_all(cache_path.parent().unwrap())?;
+        let path = Filesystem::new(cache_path.clone());
+        self.gctx
+            .assert_package_cache_locked(CacheLockMode::DownloadExclusive, &path);
+        fs::write(cache_path, value)?;
+        Ok(())
     }
 
     /// Invalidates the cache associated with the key.
--- a/tests/testsuite/registry.rs
+++ b/tests/testsuite/registry.rs
@@ -2,7 +2,7 @@
 
 use std::fmt::Write;
 use std::fs::{self, File};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::sync::Mutex;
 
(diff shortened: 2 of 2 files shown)


src/cargo/sources/registry/index/cache.rs
weihanglo, line 1: We have a similar test, maybe we can take reuse that test to assert their stderr?

https://github.com/rust-lang/cargo/blob/2939e969f5b923c9fbdf79c1448e63a875bd6b51/tests/testsuite/registry.rs?plain=1#L3053-L3098

ranger-ross: thanks, I was able to use this to create a test based off of this in https://github.com/rust-lang/cargo/pull/15014/commits/9f2d0a3bf1301bef966f798710417b3a2dc8e6ee

src/cargo/sources/registry/index/cache.rs
```diff
@@ -226,14 +227,21 @@ pub struct CacheManager<'gctx> {
     cache_root: Filesystem,
     /// [`GlobalContext`] reference for convenience.
     gctx: &'gctx GlobalContext,
+    /// Keeps track of if we have sent a warning message if there was an error updating the cache.
+    /// The motivation is to avoid warning spam if the cache is not writable.
+    has_warned: Mutex<bool>,
```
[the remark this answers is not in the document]
ranger-ross: That is fair enough :+1: 

I updated `has_warned` to `RefCell<bool>` since we need to modify it behind a shared reference and I wanted to avoid making `.put()` taking a `&mut self` (as that would create more refactoring that I don't think is beneficial)

src/cargo/sources/registry/index/cache.rs
weihanglo: Shameless copied from <https://github.com/rust-lang/cargo/pull/15026#discussion_r1905769013>.

> We generally ask that PRs are split into
>
> * One commit that adds the test, showing the currently broken behavior (ie it passes)
> * The fix commit that updates the test to pass with the change
>
> This makes the diff between the two commits show the change in behavior

Mind doing that?

tests/testsuite/registry.rs
```diff
@@ -3097,6 +3097,70 @@ fn readonly_registry_still_works() {
     }
 }
 
+#[cargo_test]
+#[cfg(not(windows))]
```
weihanglo, line 3101: Wonder why it cannot be done on Windows. Is it possible to use [`Permissions::set_readonly`](https://doc.rust-lang.org/stable/std/fs/struct.Permissions.html#method.set_readonly) here?

src/cargo/sources/registry/index/cache.rs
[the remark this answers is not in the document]
weihanglo: Nice! Thank you.

tests/testsuite/registry.rs
```diff
@@ -3097,6 +3097,70 @@ fn readonly_registry_still_works() {
...
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "a"
+                version = "0.5.0"
+                edition = "2015"
+                authors = []
+
+                [dependencies]
+                foo = '0.1.0'
```
weihanglo, line 3116: Should we add more than one dependency, so we know it really emits only once?

tests/testsuite/registry.rs
ranger-ross: Well the failure case is actually a write failure, not read. So `set_readonly` is not enough. 

That's why I used `std::os::unix::fs::PermissionsExt`.

I suppose it should also be possible to modify the inner `set_permissions` function to use a `#[cfg]` to run on both targets.
I didn't have a windows machine handy so I skipped it for now :sweat_smile:

tests/testsuite/registry.rs
ranger-ross: I was testing with this and notice the dependencies would get compiled in different orders causing `.with_stderr_data()` to fail sometimes.

But I think this can be worked around by using the `[..]` to ignore the names since they are not the point of this test.
```rs
        .with_stderr_data(str![[r#"
[WARNING] failed to write cache, path: [ROOT]/home/.cargo/registry/index/-[HASH]/.cache/3/f/fo[..], [ERROR] Permission denied (os error 13)
[COMPILING] fo[..] v0.1.0
[COMPILING] fo[..] v0.1.0
[COMPILING] a v0.5.0 ([ROOT]/foo)

"#]])
```

I think something like this should. 
But let me know if you are aware of a smarter way to do this. :+1:

tests/testsuite/registry.rs
weihanglo: Hmm. Looks like with ignore_windows the test still compiles on Windows.
My bad 🙇🏾‍♂️.  Let's go back to `cfg(unix)` then.

weihanglo: Your new approach should also work :)

ranger-ross: hmmm well CI did not compile on windows with this change ([job](https://github.com/rust-lang/cargo/actions/runs/12711114224/job/35433732061?pr=15014))

It appears the `cargo_test` only adds `#[ignore]` to the test so `#[cfg(not(windows))]` is still needed so the test will compile on windows.

I put the macro on the inner `set_permissions` function so that the test still shows up as skipped if its run on windows.

```rs
fn set_permissions(path: &Path, permissions: u32) {
    #[cfg(not(windows))]
    {
        use std::os::unix::fs::PermissionsExt;
        let mut perms = t!(path.metadata()).permissions();
        perms.set_mode(permissions);
        t!(fs::set_permissions(path, perms));
    }

    #[cfg(windows)]
    panic!("This test is not supported on windows. See the reason in the #[cargo_test] macro");
}
```

weihanglo (approved):

tests/testsuite/registry.rs
weihanglo: Thanks!

PR #15025, opened 2025-01-07
Avoid naming variables `str`

joshtriplett:
This renames variables named `str` to other names, to make sure `str`
always refers to a type.


Rustin170506 (approved):
Thanks! :shipit:

epage:
@joshtriplett what motivated this? if its not a clippy lint, should an issue for one be created?

Issue #9357, opened 2021-04-13
Tracking Issue for `cargo rustc --print=VAL`


ehuss:
**Summary**

Original issue: #8923
Implementation: #9002

`cargo rustc --print=VAL` forwards the `--print` flag to `rustc` in order to extract information from `rustc`. This runs `rustc` with the corresponding [`--print`](https://doc.rust-lang.org/rustc/command-line-arguments.html#--print-print-compiler-information) flag, and then immediately exits without compiling. Exposing this as a cargo flag allows cargo to inject the correct target and RUSTFLAGS based on the current configuration.

The primary use case is to run `cargo rustc --print=cfg` to get config values for the appropriate target and influenced by any other RUSTFLAGS.

**Unresolved issues**

(These aren't necessarily blockers, just random thoughts about the current design.)

* [ ] I think `cfg` is the only value that is worth printing. It is not clear from a UI perspective if it makes sense to expose the other values, since several will just fail or don't add anything different than running `rustc` directly. Forwarding the print value to rustc is more flexible and probably fine, though it seems like it could cause confusion.
* [ ] The intended use case of `--print=cfg` means anyone using it needs to implement a parser for the ad-hoc format. The original implementation parsed it and returned JSON, which might be easier to consume? There was some back-and-forth discussion on the PR about this, and possibly using a more general purpose subcommand (like `cargo config`) for "expose internal information". It might be worth considering that more.
* [ ] The behavior with multiple `--target` flags is a little strange.
* [ ] Profiles are ignored. This has historically caused confusion for people (for things like `debug_assertions`). However, all of cargo ignores profiles for the intended use case (`--print=cfg`). It's not clear what this should do.

**Future extensions**

**About tracking issues**

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

volks73:
Thank you for creating this issue to track stabilization. I just wanted to briefly comment about a couple of the unresolved issues (not sure if I should be creating separate issues for each of these):

[...]

PR #15026, opened 2025-01-07
Setup cargo environment for `cargo rustc --print`

Veykril:
It turns out, running `cargo rustc --print cfg -Zunstable-options` (and the like, https://github.com/rust-lang/cargo/issues/9357) fail with `.cargo/config.toml` setups like
```toml
[build]
# custom target json that lives in `./targets/my-super-cool-target.json`
target = "my-super-cool-target"

[env]
RUST_TARGET_PATH = { value = "./targets", relative = true }
```
resulting in
```

❯ cargo rustc --print cfg -Zunstable-options
error: Error loading target specification: Could not find specification for target "my-super-cool-target". Run `rustc --print target-list` for a list of built-in targets

error: process didn't exit successfully: `C:\Users\lukas\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\bin\rustc.exe --target my-super-cool-target --print cfg` (exit code: 1)
```

The reason for that is that cargo recognizes the target from the `.cargo/config` and then implicitly passes that along to the spawned rustc process, but it does so without passing along the important environment that is required for the target tuple to make sense.

(can add a test if desired, just tell me where)

### DIFF
--- a/src/cargo/ops/cargo_compile/mod.rs
+++ b/src/cargo/ops/cargo_compile/mod.rs
@@ -41,7 +41,7 @@ use std::sync::Arc;
 
 use crate::core::compiler::unit_dependencies::build_unit_dependencies;
 use crate::core::compiler::unit_graph::{self, UnitDep, UnitGraph};
-use crate::core::compiler::{standard_lib, CrateType, TargetInfo};
+use crate::core::compiler::{apply_env_config, standard_lib, CrateType, TargetInfo};
 use crate::core::compiler::{BuildConfig, BuildContext, BuildRunner, Compilation};
 use crate::core::compiler::{CompileKind, CompileMode, CompileTarget, RustcTargetData, Unit};
 use crate::core::compiler::{DefaultExecutor, Executor, UnitInterner};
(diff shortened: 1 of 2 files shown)


epage (approved):
Dug around and not seeing other places where we miss `apply_env_config`

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -188,6 +188,7 @@ pub fn print<'a>(
         }
         let target_info = TargetInfo::new(gctx, &build_config.requested_kinds, &rustc, *kind)?;
         let mut process = rustc.process();
+        apply_env_config(gctx, &mut process)?;
```
epage, line 191: > (can add a test if desired, just tell me where)

See https://github.com/rust-lang/cargo/blob/master/tests/testsuite/rustc.rs

Looks like we have `--print` tests in there.  Unsure about custom targets.

We generally ask that PRs are split into
- One commit that adds the test, showing the currently broken behavior (ie it passes)
- The fix commit that updates the test to pass with the change

This makes the diff between the two commits show the change in behavior

epage (changes requested):
Forgot about a test

epage (approved):

epage:
Thanks!

PR #15029, opened 2025-01-07
shorten comment on Ord for SourceKind

Eh2406:
As discussed in today's cargo meeting, and in https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/while.20we're.20breaking.20stable.20hash, we are keeping the manual impl  but shortening the  comment about why it is needed.

### DIFF
--- a/crates/cargo-util-schemas/src/core/source_kind.rs
+++ b/crates/cargo-util-schemas/src/core/source_kind.rs
@@ -31,59 +31,9 @@ impl SourceKind {
     }
 }
 
-/// Note that this is specifically not derived on `SourceKind` although the
-/// implementation here is very similar to what it might look like if it were
-/// otherwise derived.
-///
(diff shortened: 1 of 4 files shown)


crates/cargo-util-schemas/src/core/source_kind.rs
weihanglo, line 1: Could we remove some of those intermediate back-and-forth commits? That would make tracing down git history in the future easier. To me, maybe one for snapshotization, one for shortening comment, and one for impl hash?

(Sorry for being a picky a**hole)

Eh2406: done

weihanglo (approved):

weihanglo:
feel free to merge this when the version bump issue gets fixed

Issue #15030, opened 2025-01-08
Broken Cargo Manifest


JP-Ellis:
### Problem

A recent update has resulted in the Cargo manifest incorrectly flagging the `[lints.*]` sections as invalid.

### Steps

Run `taplo lint` against the following Cargo manifest

```toml
#:schema https://github.com/rust-lang/cargo/blob/b696870dd8cd5009a16339817cec11a882f68083/crates/cargo-util-schemas/manifest.schema.json
[package]
name = "example"

[lints.rust]
missing_docs = "warn"

[lints.clippy]
pedantic = { level = "warn", priority = -1 }

alloc_instead_of_core = "warn"

[lints.rustdoc]
missing-crate-level-docs = "warn"
```

generates the following:

<details>
<summary><code>taplo lint Cargo.toml</code></summary>

```log
 INFO taplo:lint_files:collect_files: found files total=1 excluded=0 cwd="/private/tmp/cargo"
error: {"rust":{"missing_docs":"warn"},"clippy":{"pedantic":{"level":"warn","priority":-1},"alloc_instead_of_core":"warn"},"rustdoc":{"missing-crate-level-docs":"warn"}} is not valid under any of the schemas listed in the 'anyOf' keyword
   ┌─ /private/tmp/cargo/Cargo.toml:5:1
   │  
 5 │ ╭ [lints.rust]
 6 │ │ missing_docs = "warn"
 7 │ │ 
 8 │ │ [lints.clippy]
   · │
13 │ │ [lints.rustdoc]
14 │ │ missing-crate-level-docs = "warn"
   │ ╰─────────────────────────────────^ {"rust":{"missing_docs":"warn"},"clippy":{"pedantic":{"level":"warn","priority":-1},"alloc_instead_of_core":"warn"},"rustdoc":{"missing-crate-level-docs":"warn"}} is not valid under any of the schemas listed in the 'anyOf' keyword

error: {"rust":{"missing_docs":"warn"},"clippy":{"pedantic":{"level":"warn","priority":-1},"alloc_instead_of_core":"warn"},"rustdoc":{"missing-crate-level-docs":"warn"}} is not valid under any of the schemas listed in the 'anyOf' keyword
  ┌─ /private/tmp/cargo/Cargo.toml:5:8
  │
5 │ [lints.rust]
[...]

PR #15033, opened 2025-01-08
fix(schemas): Fix 'metadata' JSON Schema

epage:
Instead of allowing any type in metadata, we were specifying fields like `"string": "<any string>"`.

Found this when looking into #15030

### DIFF
--- a/crates/cargo-util-schemas/manifest.schema.json
+++ b/crates/cargo-util-schemas/manifest.schema.json
@@ -621,40 +621,7 @@
         }
       ]
     },
-    "TomlValue": {
-      "type": "object",
-      "properties": {
-        "string": {
-          "type": "string"
-        },
-        "integer": {
-          "type": "integer",
-          "format": "int64"
-        },
-        "float": {
-          "type": "number",
-          "format": "double"
-        },
-        "boolean": {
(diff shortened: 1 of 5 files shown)


crates/cargo-util-schemas/manifest.schema.json
```diff
@@ -621,40 +621,7 @@
...
-        },
-        "table": {
-          "type": "object",
-          "additionalProperties": {
-            "$ref": "#/definitions/TomlValue"
-          }
-        }
-      }
-    },
+    "TomlValue": true,
```
weihanglo, line 624: I can't find the definition of this boolean in JSON schema spec, but it seems to accept anything so indeed fixes the issue. Tested it on <https://www.jsonschemavalidator.net/>

weihanglo (approved):

crates/cargo-util-schemas/manifest.schema.json
epage: https://json-schema.org/draft/2020-12/json-schema-core#name-boolean-json-schemas

> The boolean schema values "true" and "false" are trivial schemas that always produce themselves as assertion results, regardless of the instance value. They never produce annotation results.

PR #15037, opened 2025-01-08
docs(contrib): Start guidelines for schema design

epage:

This was inspired by a recent Cargo team discussion on whether we should generally elide default values.
This will also help with https://rust-lang.github.io/rust-project-goals/2025h1/cargo-plumbing.html

Case studies in schema design:
- https://github.com/rust-lang/cargo/issues/14506
- https://github.com/rust-lang/cargo/issues/10543

### DIFF
--- a/src/doc/contrib/src/implementation/schemas.md
+++ b/src/doc/contrib/src/implementation/schemas.md
@@ -1,0 +1,47 @@
+# Data Schemas
+
+Cargo reads and writes user and machine facing data formats, like
+- `Cargo.toml`, read and written on `cargo package`
+- `Cargo.lock`, read and written
+- `.cargo/config.toml`, read-only
+- `cargo metadata` output
+- `cargo build --message-format` output
+
+## Schema Design
+
+Generally,
+- Fields should be kebab case
+  - `#[serde(rename_all = "kebab-case")]` should be applied defensively
+- Fields should only be present when needed, saving space and parse time
+  - Also, we can always switch to always outputting the fields but its harder to stop outputting them
+  - `#[serde(skip_serializing_if = "Default::default")]` should be applied liberally
+- For output, prefer [jsonlines](https://jsonlines.org/) as it allows streaming output and flexibility to mix content (e.g. adding diagnostics to output that didn't prevously have it
+- `#[serde(deny_unknown_fields)]` should not be used to allow evolution of formats, including feature gating
+
+## Schema Evolution Strategies
+
+When changing a schema for data that is read, some options include:
+- Adding new fields is relatively safe
+  - If the field must not be ignored when present,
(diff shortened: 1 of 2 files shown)


src/doc/contrib/src/implementation/schemas.md
```diff
@@ -0,0 +1,29 @@
+# Data Schemas
```
weihanglo, line 1: Should we have an overview of what this "data schemas" refer to?

epage: I've added some context

src/doc/contrib/src/implementation/schemas.md
```diff
@@ -0,0 +1,36 @@
...
+## Schema Design
+
+Generally,
+- Fields should be kebab case
+  - `#[serde(rename_all = "kebab-case")]` should be applied defensively
+- Fields should only be present when needed, saving space and parse time
+  - Also, we can always switch to always outputting the fields but its harder to stop outputting them
+  - `#[serde(skip_serializing_if = "Default::default")]` should be applied liberally
+- For output, prefer [jsonlines](https://jsonlines.org/) as it allows streaming output and flexibility to mix content (e.g. adding diagnostics to output that didn't prevously have it
+- `deny_unknown_fields` should not be used to allow evolution of formats, including feature gating
```
weihanglo, line 19:
```suggestion
- `#[serde(deny_unknown_fields)]` should not be used to allow evolution of formats, including feature gating
```

Make it a bit explicit?

src/doc/contrib/src/implementation/schemas.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Schema Evolution Strategies
```
weihanglo, line 21: When reading this, each bullet point is a bit disconnected to me.
I can't tell what it means exactly by evolution strategies. Maybe it is a pretty common term but I didn't know?

Also like "When reading data, add new fields", I had a hard time interpreting it as part of evolution strategies. For "Versioned with the `edition`" I am able to guess it means "When a schema needs to change some existing field, versioning it with `edition`", though the whole paragraph still looks pretty sparse and needs some guesses for me to read.

epage: Tried to expand on it a bit more.  Hope it helps!

weihanglo: Thanks. That is good enough for me as a maintainer. Not sure about casual contributors, but we can always iterate later :)

weihanglo (approved):
I don't think this need any FCP as it is just some design principles we've used throughout our meeting. It just has never been written down.

Now we have a place for. Perhaps in the future we expand it to an automation to check our code.

PR #15040, opened 2025-01-09
Small resolver cleanups

Eh2406:
This is just three small resolver cleanups I found while doing some other work.

### DIFF
--- a/src/cargo/core/resolver/context.rs
+++ b/src/cargo/core/resolver/context.rs
@@ -3,10 +3,10 @@ use super::errors::ActivateResult;
 use super::types::{ActivationsKey, ConflictMap, ConflictReason, FeaturesSet, ResolveOpts};
 use super::RequestedFeatures;
 use crate::core::{Dependency, PackageId, Summary};
-use crate::util::interning::InternedString;
+use crate::util::interning::{InternedString, INTERNED_DEFAULT};
 use crate::util::Graph;
 use anyhow::format_err;
-use std::collections::HashMap;
+use std::collections::{BTreeSet, HashMap};
 use tracing::debug;
 
 // A `Context` is basically a bunch of local resolution information which is
@@ -121,6 +121,7 @@ impl ResolverContext {
             }
         }
         debug!("checking if {} is already activated", summary.package_id());
+        let empty_features = BTreeSet::new();
         match &opts.features {
             // This returns `false` for CliFeatures just for simplicity. It
             // would take a bit of work to compare since they are not in the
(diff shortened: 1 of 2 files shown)


src/cargo/core/resolver/context.rs
```diff
@@ -135,16 +136,16 @@ impl ResolverContext {
...
-                                || !has_default_feature)
-                    }
-                    None => features.is_empty() && (!uses_default_features || !has_default_feature),
-                })
+                let has_default_feature = summary.features().contains_key(&INTERNED_DEFAULT);
+                let prev = self
+                    .resolve_features
+                    .get(&id)
+                    .map(|f| &**f)
+                    .unwrap_or(&empty_features);
```
weihanglo, line 144: Does this work?

```suggestion
                    .unwrap_or_default();
```

src/cargo/core/resolver/mod.rs
```diff
@@ -789,12 +775,27 @@ impl RemainingCandidates {
...
+                }
+            }
+
             // Well if we made it this far then we've got a valid dependency. We
             // want this iterator to be inherently "peekable" so we don't
             // necessarily return the item just yet. Instead we stash it away to
             // get returned later, and if we replaced something then that was
             // actually the candidate to try first so we return that.
-            if let Some(r) = mem::replace(&mut self.has_another, Some(b.clone())) {
+            if let Some(r) = self.has_another.replace(b.clone()) {
```
weihanglo, line 798:
```
warning: unused import: `std::mem`
  --> src/cargo/core/resolver/mod.rs:62:5
   |
62 | use std::mem;
   |     ^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: `cargo` (lib) generated 1 warning (run `cargo fix --lib -p cargo` to apply 1 suggestion)
```

src/cargo/core/resolver/mod.rs
```diff
@@ -789,12 +775,27 @@ impl RemainingCandidates {
                 }
             }
 
+            // Otherwise the `links` key in the manifest dictates that there's only one
```
weihanglo, line 778: Noticed the code moved. Any specific reason?

Eh2406: Conceptually I think  maintaining "one per  compatibility range"  is more fundamental to how cargo is resolver works then the links attributes, "Otherwise" really belongs on the links. Also, I suspect that we hit the semver  check far more often than the links check, but it was not measurable in benchmarking. Branch predictors are really good.

src/cargo/core/resolver/mod.rs
```diff
@@ -142,9 +142,7 @@ pub fn resolve(
     let mut past_conflicting_activations = conflict_cache::ConflictCache::new();
 
     let resolver_ctx = loop {
-        let resolver_ctx = ResolverContext::new();
```
Eh2406, line 145: Thinking about this more, I wonder if this was a (failed) attempt to reuse memory. Perhaps `let mut resolver_ctx = ResolverContext::new();` outside the loop and `resolver_ctx.clear()`  inside the loop was the original intent.

src/cargo/core/resolver/mod.rs
weihanglo: Thanks for the explanation. I think I have guessed your reason behind correctly :)

Eh2406:
I forgot what this is waiting on. @weihanglo, what would you like to see done?

weihanglo (approved):
I forgot as well, perhaps [this warning](https://github.com/rust-lang/cargo/pull/15040/commits/92f096362c6beea8104ade579a1989ebfa8dc8f5#r1909308316)?

Anyway it seems to be fixed. Thanks!

PR #15052, opened 2025-01-11
docs: Added authentication schemes note in registry docs

### LINKED ISSUE #15021: Improve documentation about private registry credentials
### Problem

In the [Cargo Book](https://doc.rust-lang.org/cargo/reference/config.html#credentials) it states that you can set a token like so.
```toml
[registries.<name>]
token = "…"   # Access token for the named registry
```
However, the Cargo uses this token directly in the `Authorization` header without modifying it.
Many private registries will expect the token in the bearer format (`Bearer {token}`) so you will need to add `Bearer ` to the token field in `credentials.toml`.
```toml
[registries.<name>]
token = "Bearer my-token-123"
```

Generally when I see a `token` field in a config file, I expect to just add my token without caring about the format the registry expected the token to be delivered in. This lead to a good amount of confusion today 😅 

### Possible Solution(s)

1. At the bare minimum we should at least improve the documentation in the Cargo book about how the token is used and that the user might need to add the format like `Basic ` or `Bearer ` to the token field.
2. An alternative would be to retry the request to the registry prefixing the token with `Bearer ` if there is a 401 response.

### Notes

There is already precedence for retry failed request for getting the index `config.json` so extending it to downloading crates seems reasonable. 
See the [sparse authentication](https://doc.rust-lang.org/cargo/reference/registry-index.html#sparse-authentication) docs

### Version
```
release: 1.83.0
commit-hash: 5ffbef3211a8c378857905775a15c5b32a174d3b
commit-date: 2024-10-29
ssl: OpenSSL 1.1.1w  11 Sep 202
```

ranger-ross:

Reduce confusion when authenticating to private registries by improving the documentation

closes #15021

r? @weihanglo

### DIFF
--- a/src/doc/src/reference/config.md
+++ b/src/doc/src/reference/config.md
@@ -346,6 +346,10 @@ all capital letters.
 > extension by default. However, for backward compatibility reason, when both
 > files exist, Cargo will read and write the file without the extension.
 
+> **Note:** Cargo will use the token as is in the `Authorization` header without modification. 
+> For registries that require an authentication scheme like (like Bearer) the
+> `registry.<name>.token` should be prefixed with the scheme name (ie. `Bearer `) by the user.
+
 ## Configuration keys
 
 This section documents all configuration keys. The description for keys with


epage (changes requested):
Still being discussed in #15021

Issue #14157, opened 2024-06-27
Tracking Issue for feature-metadata


epage:
### Summary

RFC: [#3416](https://github.com/rust-lang/rfcs/pull/3416) 
Implementation: https://github.com/rust-lang/cargo/pull/15056
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#feature-metadata

Extend the schema for feature activations to be a table, not just an array, so more fields can be added in the future.

### Future Extensions

- https://github.com/rust-lang/rfcs/pull/3485
- https://github.com/rust-lang/rfcs/pull/3486
- https://github.com/rust-lang/rfcs/pull/3487

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

amab8901:
The docs link u posted (https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#feature-metadata) seems obsolete since there is no section called `feature-metadata` in that page

weihanglo:
@amab8901 it is just a placeholder because it hasn't yet implemented

amab8901:
I think I found the place in the `cargo` source code that needs to be changed to implement this feature:
![image](https://github.com/user-attachments/assets/6cfc249b-f6f8-47d4-8251-5cdd210bdd8b)

The `Vec<String>` should probably be replaced with `BTreeMap<String,String>`. At least that's my first thought. But then I realize it will prob break the old syntax. So now I'm thinking maybe we need some kind of generic? Or perhaps some kind of match statement to deal with Vec vs BTreeMap separately?

epage:
That section of code is an internal data structure to track `cargo add`s command-line.  `Cargo.toml`s definition is in `crates/cargo-util-schemas`.  An untagged enum would need to be used, much like we have "simple" and "detailed" dependencies.  All of this would need to be behind an unstable feature as mentioned behind our contributor guide.  We'd also need to make sure that people not using this syntax aren't switched to it on `cargo publish` so we don't force everyone to upgrade cargo.

With all that said, I'm unsure how much benefit there is on moving this forward until we have fields to take advantage of it.

amab8901:
do u have an RFC for the fields that we are waiting for?

[...]

PR #15056, opened 2025-01-12
feat(manifest)!: implement feature-metadata RFC3416

AudaciousAxiom:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/pull/15056)*

This PR implements [RFC3416](https://rust-lang.github.io/rfcs/3416-feature-metadata.html): `feature-metadata`.
It introduces an alternate syntax to define crate features in manifests: instead of simply being an array of features the feature enables, it can now be a table with, for now, one *required* array key, `enables`, which is equivalent to the existing array. This lays the groundwork for later supporting additional keys in that table, to provide feature metadata.

Related to https://github.com/rust-lang/cargo/issues/14157

Integrations tests are included in this PR, including for the normalized manifest used for publishing on a crate registry.

### Additional information

- The JSON schema of the manifest was updated using:

```sh
cargo test -p cargo-util-schemas --features unstable-schema -- 'dump_manifest_schema'
```

#### Questions

- Should there be a Cargo unstable feature for this change?
- This PR of course introduces a breaking change in the `cargo-util-schemas` *crate*, is there anything to do regarding this in this PR?
- Should this PR also attempt to update `core::Summary` or should this be left to future implementations of RFCs providing other keys (e.g., `doc`)?

### DIFF
--- a/doc/book/src/reference/unstable.md
+++ b/doc/book/src/reference/unstable.md
@@ -115,6 +115,7 @@ Each new feature described below should explain how to use it.
     * [artifact dependencies](#artifact-dependencies) --- Allow build artifacts to be included into other build artifacts and build them for different targets.
     * [Profile `trim-paths` option](#profile-trim-paths-option) --- Control the sanitization of file paths in build outputs.
     * [path bases](#path-bases) --- Named base directories for path dependencies.
+    * [feature-metadata](#feature-metadata) --- Table syntax for feature definitions.
     * [`unstable-editions`](#unstable-editions) --- Allows use of editions that are not yet stable.
 * Information and metadata
     * [unit-graph](#unit-graph) --- Emits JSON for Cargo's internal graph structure.
@@ -2443,6 +2444,21 @@ See the [`include` config documentation](config.md#include) for more.
 ## pubtime
 
 The `pubtime` index field  has been stabilized in Rust 1.94.0.
+
+## feature-metadata
+
+* Tracking Issue: [#14157](https://github.com/rust-lang/cargo/issues/14157)
+
+This allows to use a table when defining features, with a required `enables` key:
+
+```toml
+[features]
+# same as `foo = []`
+foo = { enables = [] }
+```
+
+This is equivalent to the array-of-strings syntax.
+Support for other keys should be added later.
 
 ## lockfile-path
 
--- a/tests/testsuite/features.rs
+++ b/tests/testsuite/features.rs
@@ -2458,3 +2458,211 @@ required-features = ["feat"]
     p.cargo(&format!("check --target={host} --examples --frozen"))
         .run();
 }
+
+#[cargo_test]
+fn feature_metadata() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                cargo-features = ["feature-metadata"]
+
+                [package]
+                name = "foo"
+                edition = "2015"
+
+                [features]
+                a = []
+                b = []
+                c = { enables = ["a", "b"] }
+            "#,
+        )
+        .file(
+            "src/main.rs",
+            r#"
+                fn main() {
+                    #[cfg(not(all(feature = "a", feature = "b")))]
+                    compile_error!("Cargo features `a` and `b` must be enabled");
+                }
+            "#,
+        )
+        .build();
+
+    p.cargo("check --features c")
+        .masquerade_as_nightly_cargo(&["feature-metadata"])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] invalid type: map, expected a sequence
+  --> Cargo.toml:11:21
+   |
+11 |                 c = { enables = ["a", "b"] }
+   |                     ^^^^^^^^^^^^^^^^^^^^^^^^
+
+"#]])
+        .run();
+}
+
+#[cargo_test]
+fn feature_metadata_is_unstable() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                edition = "2015"
+
+                [features]
+                a = { enables = [] }
+            "#,
+        )
+        .file("src/main.rs", "")
+        .build();
+
+    p.cargo("check --features a")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] invalid type: map, expected a sequence
+ --> Cargo.toml:7:21
+  |
+7 |                 a = { enables = [] }
+  |                     ^^^^^^^^^^^^^^^^
+
+"#]])
+        .run();
+}
+
+#[cargo_test]
+fn feature_metadata_missing_enables() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                cargo-features = ["feature-metadata"]
+
+                [package]
+                name = "foo"
+                edition = "2015"
+
+                [features]
+                foo = {}
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("check")
+        .masquerade_as_nightly_cargo(&["feature-metadata"])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] invalid type: map, expected a sequence
+ --> Cargo.toml:9:23
+  |
+9 |                 foo = {}
+  |                       ^^
+
+"#]])
+        .run();
+}
+
+#[cargo_test]
+fn feature_metadata_empty_enables() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                cargo-features = ["feature-metadata"]
+
+                [package]
+                name = "foo"
+                edition = "2015"
+
+                [features]
+                foo = { enables = [] }
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("check")
+        .masquerade_as_nightly_cargo(&["feature-metadata"])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] invalid type: map, expected a sequence
+ --> Cargo.toml:9:23
+  |
+9 |                 foo = { enables = [] }
+  |                       ^^^^^^^^^^^^^^^^
+
+"#]])
+        .run();
+}
+
+#[cargo_test]
+fn unused_keys_in_feature_metadata() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                cargo-features = ["feature-metadata"]
+
+                [package]
+                name = "foo"
+                edition = "2015"
+
+                [features]
+                foo = { enables = ["bar"], a = false, b = true }
+                bar = []
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("check")
+        .masquerade_as_nightly_cargo(&["feature-metadata"])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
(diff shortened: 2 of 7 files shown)


src/cargo/util/toml/mod.rs
```diff
@@ -1764,6 +1767,30 @@ fn to_virtual_manifest(
...
+    warnings: &mut Vec<String>,
+) -> CargoResult<()> {
+    let Some(features) = features else {
+        return Ok(());
+    };
+
+    for (feature, feature_definition) in features {
+        match feature_definition {
+            FeatureDefinition::Array(..) => {}
+            FeatureDefinition::Metadata(FeatureMetadata { _unused_keys, .. }) => {
```
weihanglo, line 1781: haven't got time into full review, though I think the meta field should be behind a nightly feature flag.

src/cargo/util/toml/mod.rs
```diff
@@ -1764,6 +1767,30 @@ fn to_virtual_manifest(
...
+                warnings.extend(
+                    _unused_keys
+                        .keys()
+                        .map(|k| format!("unused manifest key: `features.{feature}.{k}`")),
```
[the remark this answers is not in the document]
weihanglo: While not really enforced, backticks are generally the convention in Cargo

tests/testsuite/features.rs
```diff
@@ -2282,3 +2285,162 @@ fn invalid_feature_name_slash_error() {
...
+[package]
+edition = "2015"
+name = "foo"
+build = false
+autolib = false
+autobins = false
+autoexamples = false
+autotests = false
+autobenches = false
+readme = false
```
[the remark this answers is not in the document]
weihanglo: Three dots `...`
See https://doc.rust-lang.org/nightly/nightly-rustc/cargo_test_support/compare/index.html

src/cargo/util/toml/mod.rs
weihanglo: > * Should there be a Cargo unstable feature for this change?

Yes. Probably behind a cargo-feature "feature-metadata".

> * This PR of course introduces a breaking change in the `cargo-util-schemas` _crate_, is there anything to do regarding this in this PR?

Could add a doc comment on relevant field/variant indicating it is unstable/nightly only.

https://github.com/rust-lang/cargo/blob/f15df8fd4990f1b87734343c4e44f3adec3c8b00/crates/cargo-util-schemas/src/manifest/mod.rs?plain=1#L936-L937

There is a [CI job](https://github.com/rust-lang/cargo/tree/master/crates/xtask-bump-check) checking if a member crate needs a version bump. It didn't warn you so I assume it has already been bumped in this release cycle. You do not need to do anything.

> * Should this PR also attempt to update `core::Summary` or should this be left to future implementations of RFCs providing other keys (e.g., `doc`)?

Summary is more like a thing for dependency resolution. I think we revisit it in the future. Regardless, see epage's comment <https://github.com/rust-lang/cargo/issues/14157#issuecomment-2310378604> that the feature itself is not particularly useful until other RFC gets merged. Anyway, thanks for the contribution!

epage:
What is your motivation for moving this forward?

In https://github.com/rust-lang/cargo/issues/14157#issuecomment-2308615506 I was wondering if this was worth it without a feature depending on it.

epage:
> I wanted to see how much of this could be introduced into Cargo (I am aware of other [existing](https://internals.rust-lang.org/t/pre-rfc-cargo-mutually-exclusive-features/13182) [effort](https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618) in that direction), and the first step for that was to tackle feature-metadata.

Interesting effort.  Some cautions though
- Trying to add mutually exclusive support for the existing feature system is likely a dead end because features must be additive
- This would created "unused manifest fields" which are not subject to our compatibility guarantees.  I'd recommend finding a different way to declare the relevant information

> I have introduced a feature-metadata unstable feature which can be dynamically enabled through cargo-features. However, I realize now that the same cannot be done for the cargo-util-schemas crate (we can only document the FeatureDefinition::Metadata variant as unstable). So, unless there is a way to dynamically feature-gate the change in the schema, and to not cause unnecessary churn to the dependents of cargo-util-schemas, I agree it may be best not to move forward on this PR for now, until other keys are introduced. Future implementations of these could hopefully use this PR as a base.

`cargo-util-schemas` does not get feature gated.  We do that in  `cargo/utils/toml/mod.rs` or later.

syphar:
I want to invest some time to drive progress on the whole "feature documentation" topic, 

And I see we already have some code :) 

@AudaciousAxiom it looks like this PR stalled, is that something I can help with or even take over?

AudaciousAxiom:
I've rebased and squashed the existing fixup commits; the new fixup commit is for the differences in tests, which are just the slightly changed Cargo output.

@syphar I haven't been through the previous PR comments yet; feel free to point out any comment that still needs addressing or new changes needed due to the modifications of the codebase in the meantime.

tests/testsuite/features.rs
```diff
@@ -2458,3 +2461,190 @@ required-features = ["feat"]
...
+                fn main() {
+                    a();
+                    b();
+                }
+            "#,
+        )
+        .build();
+
+    p.cargo("check --features c")
+        .masquerade_as_nightly_cargo(&[])
```
weihanglo: We probably want to verify those features enabled, maybe through a `compile_error!` if a or b is not enabled?

tests/testsuite/features.rs
```diff
@@ -2458,3 +2461,190 @@ required-features = ["feat"]
...
+            r#"
+                cargo-features = ["feature-metadata"]
+
+                [package]
+                name = "foo"
+                edition = "2015"
+
+                [features]
+                foo = {}
+            "#,
```
weihanglo, line 2551: Should we have a test checking an empty `enables` field?

tests/testsuite/features.rs
```diff
@@ -2458,3 +2461,190 @@ required-features = ["feat"]
...
+    p.cargo("package --no-verify")
+        .masquerade_as_nightly_cargo(&[])
+        .run();
+    let f = File::open(&p.root().join("target/package/foo-0.0.0.crate")).unwrap();
+    let normalized_manifest = str![[r#"
+...
+
+...
+
+...
```
weihanglo: why multiple multiline wildcards?

weihanglo (commented):
Thanks for picking it up! Looks generally good. Feel free to reorganize and rebase your commit history. 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/cargo/15056/e58ed5577389b5cf597dca04bc53f4134b6444dc..9b4843c01fdd7168922a28bcaceca237e02fef05)*

tests/testsuite/features.rs
AudaciousAxiom: The construct with the feature-gated functions was already checking that the Cargo features were properly enabled, but I've replaced it with a feature-gated `compiler_error!` as you suggested, which makes it a bit shorter and more explicit.

tests/testsuite/features.rs
AudaciousAxiom: I've added a simple test case with an empty `enables` field.

tests/testsuite/features.rs
AudaciousAxiom: I think I was under the impression that might otherwise result in some false negatives; I've replaced them with a single wildcard.

AudaciousAxiom:
@weihanglo Thanks for the review! I've fixed the commit history, moved the test commit after the implementation commit as [I see you want each commit to pass the tests](https://doc.crates.io/contrib/process/working-on-cargo.html#:~:text=The%20tests%20should%20pass%20in%20each%20commit), and added the changes from the review as fixup commits. I'll simply autosquash before merging.

weihanglo:
> @weihanglo Thanks for the review! I've fixed the commit history, moved the test commit after the implementation commit as [I see you want each commit to pass the tests](https://doc.crates.io/contrib/process/working-on-cargo.html#:~:text=The%20tests%20should%20pass%20in%20each%20commit), and added the changes from the review as fixup commits. I'll simply autosquash before merging.

Thanks, though rebasing is easier for us to review. Under most circumstances, we don't need contributors to keep the edit history.

Also, the paragraph you quote actually prefers test commits placed before fix commits, so that the git diff can show the behavior change. See those examples in that note, or some newers examples like rust-lang/cargo#17356 and rust-lang/cargo#17337

weihanglo (approved):
Looks good. 

Wonder whether you are open to do git cleanup, or I can edit history and do that for you before merging.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/cargo/15056/e7167a4bac50fd878ce18530e901624d83be218e..7016b1cd32e73c18b40d14910f9ceb2cd255c5d3)*

doc/book/src/reference/unstable.md
```diff
@@ -2346,6 +2445,21 @@ See the [`include` config documentation](config.md#include) for more.
...
+This allows to use a table when defining features, with a required `enables` key:
+
+```toml
+[features]
+# same as `foo = []`
+foo = { enables = [] }
+```
+
+This is equivalent to the array-of-strings syntax.
+Support for other keys should be added later.
```
weihanglo, line 2461: Look at other unstable for example [`min-publish-age`](https://doc.rust-lang.org/1.98.0/cargo/reference/unstable.html#added-to-configuration), we add some doc examples that is meant to be copied verbatim when stabilization. 

Above just FYI, docs updates don't block this PR merge.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/cargo/15056/e7167a4bac50fd878ce18530e901624d83be218e..7016b1cd32e73c18b40d14910f9ceb2cd255c5d3)*

AudaciousAxiom:
> Thanks, though rebasing is easier for us to review. Under most circumstances, we don't need contributors to keep the edit history.

Thanks, I'll keep that in mind, this really depends on the project.

> Also, the paragraph you quote actually prefers test commits placed before fix commits, so that the git diff can show the behavior change. See those examples in that note, or some newers examples like #17356 and #17337

I did look, but all of these are bug fixes; it makes sense to me to first acknowledge the bug with tests, and then fix it and update the tests atomically, but I wasn't sure what to do for this PR, as this is a purely additive feature; adding the tests first wouldn't allow them to pass.

> Wonder whether you are open to do git cleanup, or I can edit history and do that for you before merging.

I've autosquashed the fixup commits; feel free to further edit the commits as you want.

doc/book/src/reference/unstable.md
AudaciousAxiom: Thanks for pointing that out; I didn't include this in this round to not further delay the merge if you want it now.

weihanglo:
> I did look, but all of these are bug fixes; it makes sense to me to first acknowledge the bug with tests, and then fix it and update the tests atomically, but I wasn't sure what to do for this PR, as this is a purely additive feature; adding the tests first wouldn't allow them to pass.

I tweaked a bit the tests. Thanks for the contribution anyway!

doc/book/src/reference/unstable.md
AudaciousAxiom: I've opened #17446 to update the documentation when this eventually stabilizes; I don't think this can be really be added as a doc snippet to the unstable feature docs as this is not purely additive.

weihanglo: We can't really merge #17446 now but when stabilizing.
Thanks it for doing it promptly :)

Issue #13623, opened 2024-03-22
`cargo login` encourages disclosure of secret token


auspicacious:
### Problem

I am new to Rust and working through [the Rust Book](https://doc.rust-lang.org/book/).

`cargo login`'s default syntax encourages users to pass their secret registry authentication token on the command line, e.g. `cargo login [token]`. The Rust Book also encourages this syntax.

The `cargo login` manpage makes it clear that this value requires protection:

> Take care to keep the token secret, it should not be shared with anyone else.

However, in general, passing any secret as a command-line argument creates significant and unnecessary security risk. It is significant for at least these reasons:

* While the `cargo login` command is running, its arguments are visible to all users on the system. A malicious user monitoring processes could easily steal the token.
* The token is saved in the user's shell history file. Some shells may limit permissions on this file to only that user, but this is not guaranteed, and, regardless, exposes the token in a place the user may not anticipate.
* Users may be working on systems that monitor and log the commands they execute, such that the token could be persisted into a centralized logging system with even more relaxed access permissions and many potential vulnerabilities.

And it is unnecessary because there are other options available that do not create this risk:

* `cargo login` could be modified to prompt the user for input, using normal shell conventions for entering a password, and not echoing text back to the console. #7813 touches on this.
    - This, of course, could lead to the token being persisted in a clipboard manager if it is copy/pasted, but mitigating that risk is out of scope for this issue and present regardless.
* The user could write the token into a file or a shell-oriented password manager like `pass` and pipe it into `cargo login` to avoid exposure.
* Even without modifications to `cargo login`, if using `bash`, and possibly some other shells, the user could use this combination of commands to avoid exposure:

```
IFS='' read -r -s token
cargo login <<< "$token"
unset token
```

I am honestly surprised that I could not find any duplicate issues for this, and I apologize if I have missed one that explains the rationale behind the current design. Likewise, I am not using the security vulnerability disclosure process for this bug report, as it is based entirely on public and obvious information.

However, I do believe that this is insecure by design, and encourages Rust developers to adopt patterns that are widely known to be insecure.

epage:
Note that the Cargo book shows the use of stdin, see https://doc.rust-lang.org/cargo/reference/publishing.html#before-your-first-publish

We can't remove the argument due to our compatibility guarantees.

Possible improvements from where we are at:
[...]

PR #15057, opened 2025-01-13
fix(login): Deprecate CLI token

epage:

This came up in #13623 to avoid putting tokens into shell history.

The exact approach to deprecation can vary
- Include `<token>` in at least some docs for discovery (most likely the man page)
- Don't warn yet

etc

I also suspect we could reorganize `cargo help login` but wanted to decouple that from this change.

### DIFF
--- a/src/bin/cargo/commands/login.rs
+++ b/src/bin/cargo/commands/login.rs
@@ -6,7 +6,12 @@ use crate::command_prelude::*;
 pub fn cli() -> Command {
     subcommand("login")
         .about("Log in to a registry.")
-        .arg(Arg::new("token").value_name("TOKEN").action(ArgAction::Set))
+        .arg(
+            Arg::new("token")
+                .value_name("TOKEN")
+                .action(ArgAction::Set)
+                .hide(true),
+        )
         .arg_registry("Registry to use")
         .arg(
             Arg::new("args")
@@ -27,16 +32,18 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
         "must not be index URL"
     );
 
+    let token = args.get_one::<String>("token").map(|s| s.as_str().into());
+    if token.is_some() {
+        let _ = gctx
+            .shell()
+            .warn("`cargo login <token>` is deprecated in favor of reading `<token>` from stdin");
+    }
+
     let extra_args = args
         .get_many::<String>("args")
         .unwrap_or_default()
         .map(String::as_str)
         .collect::<Vec<_>>();
-    ops::registry_login(
-        gctx,
-        args.get_one::<String>("token").map(|s| s.as_str().into()),
-        reg.as_ref(),
-        &extra_args,
-    )?;
+    ops::registry_login(gctx, token, reg.as_ref(), &extra_args)?;
     Ok(())
 }
--- a/src/doc/man/cargo-login.md
+++ b/src/doc/man/cargo-login.md
@@ -6,7 +6,7 @@ cargo-login --- Log in to a registry
 
 ## SYNOPSIS
 
-`cargo login` [_options_] [_token_] [`--` _args_]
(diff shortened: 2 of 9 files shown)


epage:
>  Have we had any consensus or FCP for this change?

I do not believe we've discussed this as a team yet.  Open to FCP or team meeting discussion.

arlosi (approved):

Turbo87:
FWIW I've seen @woodruffw use https://developer.1password.com/docs/cli/reference/commands/read/ to use command line options like this in a seemingly safe way. I wonder if we really need to deprecate this option, or if a strong warning in the documentation/help text would be sufficient.

though to be fair, this is probably irrelevant for the `cargo login` case, since CI systems should probably use env vars to pass in the token instead of saving it via `cargo login` and regular users can use stdin and are unlikely to regularly call `cargo login` for this to matter much 😅

weihanglo:
> I wonder if we really need to deprecate this option, or if a strong warning in the documentation/help text would be sufficient.

Do you suggest keeping doc/helptext for discoverable reasons? I am fine with that as well :)

epage:
I feel like `cargo login -h` help output would need to be carefully tweaked to avoid people seeing `[TOKEN]` and immediately thinking thats what they need.  I'm not even sure thats possible as we've already anchored the users conversation with the help by the usage / argument placeholder column.

woodruffw:
> FWIW I've seen @woodruffw use [developer.1password.com/docs/cli/reference/commands/read](https://developer.1password.com/docs/cli/reference/commands/read/) to use command line options like this in a seemingly safe way. I wonder if we really need to deprecate this option, or if a strong warning in the documentation/help text would be sufficient.

Yep, I frequently use that or similar (e.g. `$(gh auth token)` for GitHub)!

Ultimately, passing secrets between processes without an actual authenticated IPC channel/system secret manager is pretty difficult to do in a completely secure way 😅 -- even `cargo login $(op read ...)` can be sniffed by any other user-level process via `/proc/<pid>/cmdline`, as can any environment variable via `/proc/<pid>/environ`. Whether or not those are considered attacker-readable depends on the threat model.

IMO it'd be reasonable to nudge users towards providing tokens via `stdin`, while still allowing people to pass it via `argv` (perhaps with a loud warning about the risk of token exposure when passing a literal token rather than via a subshell?).

weihanglo:
To clarify, because of the stability guarantee Cargo cannot remove the flag (see <https://github.com/rust-lang/cargo/issues/13623#issuecomment-2014868113>).

woodruffw:
> To clarify, because of the stability guarantee Cargo cannot remove the flag (see [#13623 (comment)](https://github.com/rust-lang/cargo/issues/13623#issuecomment-2014868113)). Deprecation here means, well, a big warning 😆.

Ah, sorry for the confusion! I badly misread the diff as actually removing the positional, not merely hiding it :sweat_smile:

Turbo87:
> To clarify, because of the stability guarantee Cargo cannot remove the flag (see [#13623 (comment)](https://github.com/rust-lang/cargo/issues/13623#issuecomment-2014868113)). Deprecation here means, well, a big warning 😆.

I wasn't aware either that the plan is not to remove the flag. The word "deprecated" to me means that it would go away in a future version.

I think I would personally prefer the second option mentioned in https://github.com/rust-lang/cargo/issues/13623#issuecomment-2014868113:

> help message mentioning stdin as an alternative for more secure operation

weihanglo:
@rfcbot fcp merge

This PR deprecates the optional `<token>` argument in `cargo login`, in the sense of:

* Remove the mention of the `<token>` argument from `cargo login` help text and doc.
* Give a warning when the `<token>` argument is used
  ```
  warn: cargo login <token>` is deprecated in favor of reading `<token>` from stdin`.
  ```

Folks, take your time to check in boxes. This is not an urgent FCP.

provrb (approved):
Useful change. LGTM

PR #15062, opened 2025-01-14
Document that cargo automatically registers variables used in env! macro to trigger rebuilds

Imberflur:
Initially discussed in https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/env!.2Foption_env!.20are.20tracked.2C.20can.20documentation.20be.20improved.3F/near/491793318

There is no documentation that I can find which indicates that cargo will rebuild for changes in the variables used in `env!` macros.

Because this was not always the case, when searching for information on this, the main result indicates otherwise https://users.rust-lang.org/t/should-env-cause-rebuild-is-env-var-changes/18013.

Users misled by this may turn to `rerun-if-changed-env` in a build script to trigger rebuilds, so this documentation is a useful place to catch users and indicate that it isn't needed for `env!` macros.

I'm not sure how to test the new relative link but I matched it to the existing `[env-macro]: ../../std/macro.env.html`.

My main concern when writing this was whether it was clear that this applies to `env!` macro usage within the main code of the crate and not just in `build.rs`.

### DIFF
--- a/src/doc/src/reference/build-scripts.md
+++ b/src/doc/src/reference/build-scripts.md
@@ -403,6 +403,13 @@ variables like `TARGET` that [Cargo sets for build scripts][build-env]. The
 environment variables in use are those received by `cargo` invocations, not
 those received by the executable of the build script.
 
+As of 1.46, using [`env!`][env-macro] and [`option_env!`][option-env-macro] in
+source code will automatically detect changes and trigger rebuilds.
+`rerun-if-env-changed` is no longer needed for variables already referenced by
+these macros.
+
+[option-env-macro]: ../../std/macro.option_env.html
+
 ## The `links` Manifest Key
 
 The `package.links` key may be set in the `Cargo.toml` manifest to declare


src/doc/src/reference/build-scripts.md
```diff
@@ -403,6 +403,12 @@ variables like `TARGET` that [Cargo sets for build scripts][build-env]. The
 environment variables in use are those received by `cargo` invocations, not
 those received by the executable of the build script.
 
+As of 1.46, variables used by [`env!`][env-macro] and
```
weihanglo, line 406: Thanks!

Unsure if we want to put these under `> **MSRV:**` quote like other parts of the book. Maybe unnecessary?
WDYT?

src/doc/src/reference/build-scripts.md
```diff
@@ -403,6 +403,12 @@ variables like `TARGET` that [Cargo sets for build scripts][build-env]. The
 environment variables in use are those received by `cargo` invocations, not
 those received by the executable of the build script.
 
+As of 1.46, variables used by [`env!`][env-macro] and 
+[`option_env!`][option-env-macro] are automatically detected. So using a build
+script to register these with `rerun-if-env-changed` is not necessary.
```
weihanglo, line 408: > My main concern when writing this was whether it was clear that this applies to env! macro usage within the main code of the crate and not just in build.rs.

What about this?

> As of 1.46, using `env!` and `option_env!` in source code will automatically detect changes and trigger rebuilds changes. `rerun-if-env-changed` is no longer needed for variables already referenced by these macros.

Imberflur: Removing `changes` at the end of the first sentence since it looks like a typo:

> As of 1.46, using `env!` and `option_env!` in source code will automatically detect changes and trigger rebuilds. `rerun-if-env-changed` is no longer needed for variables already referenced by these macros.

The wording here looks like an improvement to me. Although I don't know if it alleviates my concern much, I would still be happy with this! I actually considered:
> So using a build script to register these with `rerun-if-env-changed`

to somewhat help imply that the macros used weren't specifically in a build script by focusing the fact that build scripts are involved on the need to use `rerun-if-env-changed`, but that may be a bit of a stretch.

src/doc/src/reference/build-scripts.md
weihanglo: I am fine with this as-is to be honest. No need to put everything in.

src/doc/src/reference/build-scripts.md
weihanglo: Understandable.

To me register is a more technical word that may need more explanation. Anyway, let me know if you want to do further revise. I could merge this as is as well.

Imberflur: I'm going to change it to your suggested wording.

weihanglo:
Probably flaky. Return tests.

```
---- test::panic_abort_tests stdout ----
running `/home/runner/work/cargo/cargo/target/debug/cargo test -Z panic-abort-tests -v`

thread 'test::panic_abort_tests' panicked at tests/testsuite/test.rs:5023:10:

---- expected: tests/testsuite/test.rs:5014:13
++++ actual:   stderr
   1    1 | [RUNNING] `[..]--crate-name a [..]-C panic=abort[..]`
   2    2 | [RUNNING] `[..]--crate-name foo [..]-C panic=abort[..]`
   3      - [RUNNING] `[..]--crate-name foo [..]-C panic=abort[..]--test[..]`
   4    3 | ...∅
```

weihanglo (approved):

Imberflur:
I forgot to check here first and pushed to try to rerun the tests.

weihanglo:
> I forgot to check here first and pushed to try to rerun the tests.

Nah. I added the comment just for a record, since I clicked the rerun button in GitHub CI dashboard.

PR #15066, opened 2025-01-14
fix(config): When merging, replace rather than combining specific configuration keys

### LINKED ISSUE #14906: Configuration merge doesn't make sense for credential-provider
### Problem

When merging multiple `config.toml` files, Cargo concatenates arrays. For the `registries.….credential-provider` setting this merging is counter-productive, because it constructs an invalid command.

### Steps

When there's more than one applicable `cargo.toml` file that contains:

```toml
[registries.custom]
credential-provider = ["cargo:token-from-stdout", "command", "arguments"]
```

after merging, it becomes:

```toml
[registries.custom]
credential-provider = ["cargo:token-from-stdout", "command", "arguments", "cargo:token-from-stdout", "command", "arguments"]
```

and results in Cargo running `command arguments cargo:token-from-stdout command arguments`, which contains wrong arguments, and can fail in very confusing ways.

It's particularly easy to end up with exact duplicates of a custom registry configuration when configs are injected by build tools or projects have their own copy in addition to per-user or per-host configs.

### Possible Solution(s)

The `credential-provider` field should be replaced, not concatenated, when merging config files.

### Notes

_No response_

### Version

```text
```

arlosi:
In general, Cargo merges lists in configuration by concatenating them. However, sometimes the lists don't make sense for merging, such as a program and its arguments. We had the `UnmergedStringList` type that handled this case for merging environment variables, but it did not work for multiple configs.

* Removes the `UnmergedStringList` type, which only worked for preventing merging of environment variables with configuration.
* Adds a new function `is_nonmergable_list` which hard-codes which configuration keys contain lists that should not be merged.

Fixes #14906

### DIFF
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -736,7 +736,7 @@ impl GlobalContext {
                     Ok(Some(CV::List(cv_list, cv_def)))
                 }
                 Some(cv) => {
-                    // This can't assume StringList or UnmergedStringList.
+                    // This can't assume StringList.
                     // Return an error, which is the behavior of merging
                     // multiple config.toml files with the same scenario.
                     bail!(
@@ -910,20 +910,8 @@ impl GlobalContext {
     }
 
     /// Helper for `StringList` type to get something that is a string or list.
-    fn get_list_or_string(
-        &self,
-        key: &ConfigKey,
-        merge: bool,
-    ) -> CargoResult<Vec<(String, Definition)>> {
+    fn get_list_or_string(&self, key: &ConfigKey) -> CargoResult<Vec<(String, Definition)>> {
         let mut res = Vec::new();
-
-        if !merge {
-            self.get_env_list(key, &mut res)?;
-
-            if !res.is_empty() {
-                return Ok(res);
-            }
-        }
 
         match self.get_cv(key)? {
             Some(CV::List(val, _def)) => res.extend(val),
@@ -943,6 +931,7 @@ impl GlobalContext {
     }
 
     /// Internal method for getting an environment variable as a list.
+    /// If the key is a non-mergable list and a value is found in the environment, existing values are cleared.
     fn get_env_list(
         &self,
         key: &ConfigKey,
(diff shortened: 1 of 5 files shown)


epage:
When we talked about this in the Cargo team meeting (2025-01-07), we were ok with this so long as we felt there was a proper way of fixing this later.  One idea to do that was to delay merging until access.  We're unsure how difficult that would be to implement.

src/cargo/util/context/mod.rs
```diff
@@ -2348,6 +2360,16 @@ impl ConfigValue {
     }
 }
 
+/// List of which configuration lists cannot be merged.
+/// Instead of merging, these the higher priority list replaces the lower priority list.
+fn is_nonmergable_list(key: &ConfigKey) -> bool {
+    key.matches("registries.*.credential-provider")
```
ehuss, line 2366: Does this also need to include `"registry.credential-provider"`?

src/cargo/util/context/mod.rs
```diff
@@ -2348,6 +2360,16 @@ impl ConfigValue {
     }
 }
 
+/// List of which configuration lists cannot be merged.
+/// Instead of merging, these the higher priority list replaces the lower priority list.
+fn is_nonmergable_list(key: &ConfigKey) -> bool {
```
ehuss, line 2365: I'm slightly concerned with this getting out of sync. As a small mitigation, can you add a bold sentence to the `PathAndArgs` type that says if you use it in a new key that this needs to be udpated?

ehuss (commented):
Can you rebase on latest master and resolve the conflict?

ehuss (approved):
Thanks!

We discussed the concerns about potentially breaking behavior, since it is feasible someone is relying on something like `runner` concatenating arguments. Our perspective is that this fixes behavior that we weren't expecting, and brings it in line with how env merging works. We decided that we will monitor for any issues that get raised in nightly or beta, and consider backing it out. We'll make sure that the release notes for stable have a notice about this change.

PR #14815, opened 2024-11-13
Always include Cargo.lock in published crates

sdroege:
### What does this PR try to resolve?

Originally it was only included for packages that have executables or examples for `cargo install`, however this causes inconsistencies and is kind of unexpected nowadays, e.g. with cdylib crates.

Including it always only slightly increases the crate size and allows for all crates to know a set of dependency versions that were working, which can make regression tracking easier.

Fixes https://github.com/rust-lang/cargo/issues/13447

### How should we test and review this PR?

The existing tests are covering this change in all kinds of various already, and one test that previously asserted that there is *no* Cargo.lock for library crates was changed to explicitly check for the new behaviour.

PR #15067, opened 2025-01-15
docs(cargo-package): alwasy include the lockfile

weihanglo:

This was overlooked in #14815.

```
cargo build
target/debug/cargo help package
# and read the manpage
```

### DIFF
--- a/src/doc/man/cargo-package.md
+++ b/src/doc/man/cargo-package.md
@@ -26,9 +26,9 @@ stored in the `target/package` directory. This performs the following steps:
     - The original `Cargo.toml` file is rewritten and normalized.
     - `[patch]`, `[replace]`, and `[workspace]` sections are removed from the
       manifest.
-    - `Cargo.lock` is automatically included if the package contains an
-      executable binary or example target. {{man "cargo-install" 1}} will use the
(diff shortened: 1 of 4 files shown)


src/doc/man/cargo-package.md
```diff
@@ -26,9 +26,9 @@ stored in the `target/package` directory. This performs the following steps:
     - The original `Cargo.toml` file is rewritten and normalized.
     - `[patch]`, `[replace]`, and `[workspace]` sections are removed from the
       manifest.
-    - `Cargo.lock` is automatically included if the package contains an
-      executable binary or example target. {{man "cargo-install" 1}} will use the
-      packaged lock file if the `--locked` flag is used.
+    - `Cargo.lock` is included if available, or generated when missing.
```
epage, line 29: I feel like leading with `if available` will set the tone for readers so that they misunderstand this. I know my mind went one direction when starting to read this and I had to bring it back to where things are with the second half. We need to assume people will stop reading documentation at any point, because they do.

epage: The last as it makes a clear statement and then clarifies

epage (approved):

Issue #15027, opened 2025-01-07
Option to set default authors for `cargo new|init`


Kladki:
### Problem

When making a new cargo package, I rarely set the author to anything other than myself. I don't want to have to write out my full name and email whenever I make a new package.

### Proposed Solution

Inside the [cargo configuration file](https://doc.rust-lang.org/cargo/reference/config.html), there should be an option to set default authors which should be set when running `cargo new` or `cargo init`.

epage:
Some combination of
- Document it as deprecated (and reject features)
- Provide a deprecation warning on use
- Remove on next edition

weihanglo:
> * Document it as deprecated (and reject features)

This looks fine, if we focus on rejectin features.

> * Provide a deprecation warning on use
> * Remove on next edition

This would push people away from using `package.authors`. Popular crates like `clap` provide [`crate_authors!` macro](https://docs.rs/clap/4.5.24/clap/macro.crate_authors.html) to populate the value at compile time. And there are [thousands of uses of it on GitHub](https://github.com/search?q=%22crate_authors%21%22&type=code).
Do we prepare to have that churn for the ecosystem?

As an alternative, we could decide to freeze any feature regarding `package.authors` in this issue. Whenever a new feature request pops up, we link to this issue.

epage:
`clap` no longer includes the authors in any output by default.  Looking at a sampling of those search results:
- `package.authors` is set and `authors` is set on clap, but **clap isn't showing it**
- `package.authors` is **not** set and `authors` is set on clap, but **clap isn't showing it**
- they are using clap v2 which means it is likely legacy code

epage:
With `authors` being deprecated, I'm closing this.  If there is a reason to revisit this, let us know!

Issue #14949, opened 2024-12-17
Feature request: `authors-file` package field


timmypidashev:
### Problem

Hello 👋

I am building some rust software, and couldn't find a package field that allows for mentioning/referencing an AUTHORS file in the authors field. This is unfortunate, as I already make use of a non-standard license, and the license-file field came in very handy. I assumed something like authors-file should also exist, but alas.

When a project sees multiple contributions, It is not feasible to keep track of them all in a authors list, nor is it expandable. It would be nice to have a field where we can pass in a markdown or text file with a list of maintainers/contributors, and have cargo fetch that into its package metadata.

### Proposed Solution

I propose adding an authors-file field to Cargo.toml that works similarly to the existing license-file field. Here's how it would work:

```toml
[package/workspace.package]
authors-file = "AUTHORS"
```
The AUTHORS file would support categorization through comments, with lines starting with # being used as category headers:
```md
# Core Team
Your Name <your.email@example.com>

# Contributors
Another Author <another.email@example.com>
```
Additionally, we should provide support for GitHub's CONTRIBUTORS file format, allowing markdown formatting while extracting just the relevant contributor information.

### Notes

This is my first potential contribution to the Rust ecosystem, and I'm really excited about it! It follows the precedent set by license-file and provides a more scalable alternative compared to listing authors directly in Cargo.toml.

I'd love to hear what others think and if its a good idea, I'd love to submit a pull request. Looking forward to getting involved with the Rust community through this feature! 😊

epage:
`package.authors` used to be required and then it was made optional and de-emphasized / hidden in official tooling, see rust-lang/rfcs#3052

In that context, I'm not really seeing us making use of a `package.authors-file` in a meaningful way that having built-in support, especially designing out something semi-structured, would be worth it.

timmypidashev:
Thanks for the reply. That makes a lot more sense, and seeing as it's not even a requirement means anyone who is in need can simply default to using a CONTRIBUTORS file or the like. I will drop the issue 👍

Issue #3052, opened 2016-08-28
Using build.rustflags is awkward right now


CamJN:
Based on https://github.com/rust-lang/cargo/issues/2535 and https://github.com/rust-lang/rust/pull/23238 and https://github.com/rust-lang/rust/issues/13841. I thought I could add 

```
[build]
rustflags = ["-C target-cpu=native"]
```

to my ~/.cargo/config file to pass that flag to my builds. But that breaks everything. Turns out I have to use `["-C", "target-cpu=native"]` which is weird but it works, so w/e. However when I run `cargo build --verbose` I noticed that the `-C target-cpu=native` flag gets passed to rustc twice in immediate succession. ie: `rustc src/main.rs --crate-name min --crate-type bin -g --out-dir /Users/camdennarzt/Developer/Rust/min/target/debug --emit=dep-info,link -L dependency=/Users/camdennarzt/Developer/Rust/min/target/debug -L dependency=/Users/camdennarzt/Developer/Rust/min/target/debug/deps -C target-cpu=native -C target-cpu=native`

rustc: 1.11.0
cargo: 0.11.0 (259324c 2016-05-20)
uname -a: Darwin SECUR-T.local 15.6.0 Darwin Kernel Version 15.6.0: Thu Jun 23 18:25:34 PDT 2016; root:xnu-3248.60.10~1/RELEASE_X86_64 x86_64

nagisa:
"-C target-cpu=native" is a single argument (i.e. shell does not split them for you), so what the target binary really receives is not `"rustc" "-C" "target-cpu=native"` but `"rustc" "-C target-cpu=native"` which is completely different both behaviour and semantics wise.

Alas, you cannot do such splitting either, because that would preclude people from passing over stuff (i.e. filenames) with spaces in it.

CamJN:
I dunno, I would expect that the command is simply concatenated together and passed to the shell, as that's less work (a huge priority of cargo's) and more flexible. You can pass files with spaces in them the same way as people have done for my entire life, by escaping the space with `\`. 

Either way I think we can agree that passing the flag twice is wrong.

alexcrichton:
@CamJN did you happen to also have `RUSTFLAGS` or `CARGO_BUILD_RUSTFLAGS` environment variables set at the time? Passing twice definitely seems wrong, and I couldn't reproduce locally unfortunately.

Also we can and should accept configuration like:

```toml
[build]
rustflags = "-C target-cpu=native"
```

[...]

PR #15068, opened 2025-01-15
docs(ref): Deprecate 'package.authors' 

epage:

This was left as a future-possibility in
[RFC #3052](https://rust-lang.github.io/rfcs/3052-optional-authors-field.html).
Without a clear role for this field within Cargo/crates.io,
there seems little reason to actively support it (e.g. #14949, #15027),
including working through feature requests.
I see the potential for having some fields with more specific roles
(where to go for support, where to go for disclosing vulnerabilities, etc)
but even these run into the mutable data problem of `authors`.
Users can always manage metadata for their own purposes through `package.metadata`.

The RFC called out clap as an example of something that uses this.
Clap no longer does by default but users can opt-in through a
`help_template`.

For now, this notes the status in the documentation.
When we have[user control over Cargo lint (#12235)
(and figure out namespacing with lints that overlap with Rust),
we can give a warning on this.
Based on feedback from users, we can evaluate removing this field in a
future edition.

This is as discussed in a recent Cargo team meeting

### Additional information

We may want to consider doing similar for `badges`


epage:
CI failure seems unrelated
```
     Running benches/global_cache_tracker.rs (target/debug/deps/global_cache_tracker-c5606e28ceb8cc09)
Gnuplot not found, using plotters backend
     Running benches/resolve.rs (target/debug/deps/resolve-954fd14fc20489a6)
Gnuplot not found, using plotters backend
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint:
hint: 	git config --global init.defaultBranch <name>
hint:
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint:
hint: 	git branch -m <name>
Initialized empty Git repository in /home/runner/work/cargo/cargo/target/tmp/bench/index/
From https://github.com/rust-lang/crates.io-index-archive
 * branch                85f7bfd61ea4fee08ec68c468762e886b2aebec6 -> FETCH_HEAD
Testing resolve_ws/cargo
    Updating `local-snapshot` index

thread 'main' panicked at benches/benchsuite/benches/resolve.rs:47:6:
called `Result::unwrap()` on an `Err` value: no matching package named `clap` found
location searched: `local-snapshot` index (which is replacing registry `crates-io`)
required by package `cargo v0.58.0 (/home/runner/work/cargo/cargo/target/tmp/bench/workspaces/cargo)`
```

weihanglo:
See <https://github.com/rust-lang/cargo/pull/15065#issuecomment-2593421395>. Haven't figured out the root cause.

weihanglo (approved):

PR #15070, opened 2025-01-16
fix(publish): Report all unpublishable packages 

epage:
I didn't extend this to multiple packages restricted to specific
registries.
It seems less likely to be a problem and more complex to gather and
report.

This was inspired by feedback left at #10948


joshka (approved):
Looks reasonable to me.

joshka (approved):

weihanglo (approved):
Looks good 👍🏾 

Thanks for the improvement, and also @joshka for the review :)

PR #15071, opened 2025-01-16
feat: emit error if package not found within workspace

linyihai:

Fixes https://github.com/rust-lang/cargo/issues/12978

currently, if `--package `and `--workspace` passe in the same time, the `--package` will be ignore even the package doesnt exist in the workspace, this PR regards this behavior as an error 

one commit add test, one commit fixes the issue.

### DIFF
--- a/tests/testsuite/workspaces.rs
+++ b/tests/testsuite/workspaces.rs
@@ -2648,11 +2648,10 @@ fn nonexistence_package_togother_with_workspace() {
     let p = p.build();
 
     p.cargo("check --package nonexistence --workspace")
+        .with_status(101)
         .with_stderr_data(
             str![[r#"
-[CHECKING] foo v0.1.0 ([ROOT]/foo)
-[CHECKING] baz v0.1.0 ([ROOT]/foo/baz)
-[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+[ERROR] package(s) `nonexistence` not found in workspace `[ROOT]/foo`
 
 "#]]
             .unordered(),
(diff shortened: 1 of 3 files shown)


src/bin/cargo/commands/add.rs
```diff
@@ -185,7 +185,7 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
         print_available_packages(&ws)?;
     }
 
-    let packages = args.packages_from_flags()?;
+    let packages = args.packages_from_flags(gctx, false)?;
```
epage, line 188: > set ignore_package as true to indicate that dont pop-up warnning, these commands needs this

why does `add` not need this?

src/bin/cargo/commands/remove.rs
```diff
@@ -66,7 +66,7 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
         print_available_packages(&workspace)?;
     }
 
-    let packages = args.packages_from_flags()?;
+    let packages = args.packages_from_flags(gctx, false)?;
```
epage, line 69: > set ignore_package as true to indicate that dont pop-up warnning, these commands needs this

why does `remove` not need this?

tests/testsuite/workspaces.rs
```diff
@@ -2658,22 +2658,25 @@ fn warn_package_togother_with_workspace() {
     // baz is present in the workspace and it is a real existing package.
     p.cargo("check --package baz --workspace")
         .with_stderr_data(str![[r#"
-[CHECKING] foo v0.1.0 ([ROOT]/foo)
+[WARNING] ignoring `--package` as the `--workspace` or `--all` flag is set
```
epage, line 2661: The issue wasn't about redundant `--package` flags but about non-existent packages.  *If* we decide to warn on redundant flags, then we can simplify the design of this but that is something that needs to be figured out first.

linyihai: > The issue wasn't about redundant --package flags but about non-existent package.

My PR was deviated from the original question. I tried to add warning for non-existent package.

src/bin/cargo/commands/remove.rs
linyihai: `add` and `remove`  command don't support the --workspace option.

epage (approved):
Thanks!

While this turns a success case into an error case, I'm going to go ahead and merge this without an FCP because this was discussed in a Cargo team meeting where we had consensus.  This was a bug that we are fixing which limits how much focus we put on compatibility.  We are going straight to an error as we feel there is little value gained from a more gradual transition in this case.

If someone does have some concerns, we have nightly and beta to address them.

PR #15075, opened 2025-01-16
took the functionality of the third party subcommand from the list_commands function

ibilalkayy:
Hey there,

I just moved the functionality of the third-party subcommand from the [list_commands](https://github.com/rust-lang/cargo/blob/master/src/bin/cargo/main.rs#L184) function to another new function by the name of third_party_subcommand name and called that function in the list_commands function.

From my understanding regarding the third-party subcommand from this [tracking issue](https://github.com/rust-lang/cargo/issues/14520), following points should be performed.

- The code that gathers third-party subcommands in list_commands should be moved into a separate function that I did.

- This new function will be called both by list_commands and the code for adding subcommand completions. Although I called the function in the list_commands but didn't understand the second point.

- Test the change that I made in the first place but it gave me this error: 

carbon (68)


epage (approved):
Thanks!

PR #15076, opened 2025-01-17
created a function for user defined aliases

ibilalkayy:
Hey there,

As mentioned in this [issue](https://github.com/rust-lang/cargo/issues/14520), I worked on the functionality of autocompleting the user-defined aliases and I moved the code of user-defined aliases into a separate function by defining two parameters. The first one is global context or gctx and the second one is the commands that can be called in the list_commands function.

Thank you!

### DIFF
--- a/src/bin/cargo/main.rs
+++ b/src/bin/cargo/main.rs
@@ -205,16 +205,8 @@ fn list_commands(gctx: &GlobalContext) -> BTreeMap<String, CommandInfo> {
     }
 
     // Add the user-defined aliases
-    if let Ok(aliases) = gctx.get::<BTreeMap<String, StringOrVec>>("alias") {
-        for (name, target) in aliases.iter() {
-            commands.insert(
-                name.to_string(),
-                CommandInfo::Alias {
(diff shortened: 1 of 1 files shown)


src/bin/cargo/main.rs
```diff
@@ -258,6 +249,20 @@ fn third_party_subcommands(gctx: &GlobalContext) -> BTreeMap<String, CommandInfo
     commands
 }
 
+fn user_defined_aliases(gctx: &GlobalContext, commands: &mut BTreeMap<String, CommandInfo>) {
```
epage, line 252: let's be consistent in how we define these functions rather than one using a return value and another an out param

provrb: Agreed

ibilalkayy:
@epage I removed the command parameter this time and returned it to call it in the `list_commands` function. After calling it, I used `.extend` to extend the alias.

epage (approved):

ibilalkayy:
Thank you, for your guidance and approval!

PR #15093, opened 2025-01-23
Print globs when workspace members can't be found

kornelski:
Cargo expands globs when loading workspace members. If the glob happens to match some non-crate directory, it causes an error that may be tricky to understand, because it makes Cargo complain that it failed to read a `Cargo.toml` from a specific directory, but that directory name won't be listed explicitly anywhere in `Cargo.toml`, so it may seem that Cargo is trying to read some made-up phantom manifest.

I've made the error messages mention which glob has been used to select the missing workspace member.


src/cargo/core/workspace.rs
```diff
@@ -1883,16 +1892,19 @@ impl WorkspaceRootConfig {
             // If glob does not find any valid paths, then put the original
             // path in the expanded list to maintain backwards compatibility.
             if expanded_paths.is_empty() {
-                expanded_list.push(pathbuf);
+                expanded_list.push((pathbuf, None));
             } else {
+                let used_glob_pattern = expanded_paths.len() > 1 || expanded_paths[0] != pathbuf;
```
epage, line 1897: Why the `>1` check?

kornelski: When one path expands to multiple paths, that's certain it's due to a glob.

epage (approved):

epage:
Thanks for improving the error

PR #15103, opened 2025-01-25
Fix comment on Ord for SourceId

teor2345:

In PR #14980, the Ord impl for SourceId was changed, but the comment wasn’t updated. So it is now incorrect.

This PR updates the comment to match the implementation.

Read the Ord for SourceId code and make sure the updated comment describes what it does.

### Additional information

None


teor2345:
cc @Eh2406 as the original PR author

weihanglo (approved):
Yeah this looks more correct. Thanks!

Issue #14125, opened 2024-06-21
Redefine `CARGO_TARGET_DIR` to be only an artifacts directory


kornelski:
### Problem

There are a couple of issues with the `CARGO_TARGET_DIR` that are seemingly in conflict with each other:

1. Users don't like that the `target` dir is huge, and that there are many copies of `target` directories scattered around disk for individual projects. Multiple locations complicate excluding them from backups and full-disk search, cleanup of the temp files, moving temp files to dedicated partitions, out of slow network drives or container mounts, etc. Users would prefer a central location to ease management of the temp files, and also to dedupe/reuse dependencies across many projects.

2. People (and tools) are relying on a relative `./target` directory being present to copy or run built files out of there. Additionally, users may not want to configure a shared `CARGO_TARGET_DIR` due to risk of file name conflicts between projects.

However, the dilemma between 1 and 2 exists only because Cargo uses `CARGO_TARGET_DIR` for two different roles:

1. A cache for all intermediate build products (a place where crates.io crates are built, where compiler-private temp files are) which aren't project-specific, and/or files that users don't need to access directly.
2. A location for user-facing final build products (artifacts) that users expect to be there and need to access.

### Proposed Solution

So to satisfy both uses, I suggest to change the thinking about what the role of `CARGO_TARGET_DIR` should be. Instead of thinking *where* to put the same huge all-purpose mixed `CARGO_TARGET_DIR`, think how to deduplicate and slim `CARGO_TARGET_DIR`, and move everything non-user-facing out of it. 

Instead of merging or sharding the `CARGO_TARGET_DIR` as-is with all of its current content, and adding `--artifact-dir` as a separate place where final products are being copied to — make `CARGO_TARGET_DIR` to be the artifact dir (without copying).

As long as the `CARGO_TARGET_DIR` dir is the place for *all* of the build files, of all crates including all the crates.io and local builds, with all the caches, all the temp junk, then this is going to be a problematic large directory that needs to be managed. But if the purpose of the `./target` dir was changed to be only for user-facing files (files that users can name, and would access via `./target` path themselves), then this directory would be relatively small, with a good reason to stay workspace-relative.

### What isn't an intermediate build product? (and should stay in `./target`)

* linked (and stripped) binaries of the current workspace, including binaries for the examples,
* libraries of the current workspace as `.a`/`.so`, where `lib.crate-type` calls for them. Possibly `.rlib`/`.rmeta` in the future if there's a stable ABI.
[...]

Issue #15061, opened 2025-01-14
Make target dir self-ignoring


senekor:
### Problem

Every Rust project using git requires a `.gitignore` file with the content `/target`. This contributes to a cluttered project directory.

### Proposed Solution

Whenever cargo creates the `target` directory, it should also create a file `target/.gitignore` with the content `*` to ensure the directory ignores itself. That way, a top-level `.gitignore` file that's checked into version control can be avoided completely in many cases.

### Notes

This is part of #11548, which also proposes that cargo doesn't generate the top-level `.gitignore` file anymore. As mentioned in [this comment](https://github.com/rust-lang/cargo/issues/11548#issuecomment-2588392241), that should only be done much later (if at all) for backwards-compatibility reasons.

epage:
If this is something we always write out, how do we decide which tools that we make it self-ignoring for?

A lot of this was talked about in #11548 but now we're splitting that conversation.  If we split it, its important to collect context like that.

>  That way, a top-level .gitignore file that's checked into version control can be avoided completely in many cases.

How often do projects only have `target` in their gitignore?

senekor:
> how do we decide which tools that we make it self-ignoring for?

Do you mean which VCS? Judging from their documentation, mercurial, fossil and pijul don't support this kind of approach. (I haven't verified it though, I don't know / use any of them. Their documentation might just be incomplete.) So, I'd say only git (and because of its git-compatibility, jj).

If the other VCS gain support for self-ignoring directories, I don't see a reason not to support it from cargo's side.

> How often do projects only have `target` in their gitignore?

I don't have the numbers. Do you think this is not worth it because most projects need a gitignore anyway?

I realize this proposal isn't a ground-breaking value-add, but I don't see much of a cost to it either.

epage:
> I realize this proposal isn't a ground-breaking value-add, but I don't see much of a cost to it either.

- its adding an inconsistency that we have to keep in mind and deal with
- There is nuance to how/when to write the file to `target/` to make sure we get behavior we want for people to rely on

Neither of these is necessarily a deal breaker but when it comes to allocating my focus, I want to understand "is this worth it".

weihanglo:
Probably this scenario: if you set `build.target-dir` to some path else that is not git-ignored. With this feature it is still ignored without any further user intervention.

senekor:
Code complexity is a valid cost to consider. I don't think there is much benefit beyond avoiding a (relatively harmless) file for small repos.

[...]

Issue #11548, opened 2023-01-06
instead of writing a `.gitignore` with `/target` -- output a `.gitignore` with `*` *inside* `target/`


asottile:
### Problem

currently `cargo init` helpfully writes out a `.gitignore` with the following contents:

```gitignore
/target
```

this ~generally means that any rust project now needs to have this `/target` entry in `.gitignore`

instead of needing this in `.gitignore`, one can create a self-ignoring directory by writing a `.gitignore` with the contents `*` inside of it.  this is (for example) what `virtualenv` does:

```console
$ virtualenv venv >& /dev/null
$ cat venv/.gitignore
# created by virtualenv automatically
*
```

this would remove the need for everyone having `/target` in their `gitignore` and instead have it be automatic!

### Proposed Solution

the proposal is to create a `.gitignore` with `*` in the `target` directory upon creation and phase out the top-level `.gitignore`'s `/target` entry

weihanglo:
This looks pretty interesting! Thanks for your proposal.

There are several things needs to consider. One is backward compatibility. I believe it is rare that a user commits stuff under `target/` directory. It may be fine to completely ignore `target` dir, although I am not 120% sure there is no one what to commit `target/cargo-timings` for further build pipeline investigation. (Cargo should provide a better way to store such information 🤪)

Cargo supports various kinds of VCS, e.g. Git, Mecurial, Pijul, and Fossil. We don't need to do it all at once, as 99% of Rust projects choose Git. Still, we can perform more researches on how other VCS works. 

Going further, “which VCS ignore file should be generated” becomes a question. Is it a good practice for Cargo to check VCS information for each run of `cargo build`? Could we store VCS info somewhere to avoid that cost? Or just generate 4 kinds of ignore files when every time Cargo recreates `target`?

The idea is cool anyway. Thank you again.

LunarLambda:
Writing 4 tiny (single line) files should not be a significant cost compared to any kind of average cargo compile, even hello world.

Meson for example generates ones for git and mercurial only, but others wouldn't be difficult, assuming they provide glob support. However, it only generates them once. Cargo could do the same if `cargo clean` didn't delete target/ entirely, but left the VCS ignores

mathstuf:
Please keep `/target` in the ignore as putting in the directory does not support `target`-as-symlink.

LunarLambda:
In what case would you need target to be a symlink?

mathstuf:
[...]

PR #15104, opened 2025-01-26
Implemented `build.build-dir` config option

ranger-ross:

This PR adds a new `build.build-dir` configuration option that was proposed in https://github.com/rust-lang/cargo/issues/14125#issuecomment-2258708917

This new config option allows the user to specify a directory where intermediate build artifacts should be stored. 
I have shortened it to just `build-dir` from `target-build-dir`, although naming is still subject to change.

#### Final artifacts
These are the files that end users will typically want to access directly or indirectly via a third party tool.

* binary executable for bin crates
* binary executable for examples
* `.crate` files output from `cargo package`
* [depinfo files](https://doc.rust-lang.org/cargo/reference/build-cache.html#dep-info-files) (`.d` files) for third party build-system integrations (see https://github.com/rust-lang/cargo/blob/master/src/cargo/core/compiler/fingerprint/mod.rs#L194)
* `cargo doc` output (html/css/js/etc)
* Cargo's [`--timings`](https://doc.rust-lang.org/cargo/reference/timings.html) HTML report

#### Intermediate build artifact, caches, and state
These are files that are used internally by Cargo/Rustc during the build process

* other depinfo files (generated by rustc, fingerprint, etc. See https://github.com/rust-lang/cargo/blob/master/src/cargo/core/compiler/fingerprint/mod.rs#L164)
* rlibs and debug info from dependencies
* build script `OUT_DIR`
* output from proc macros (previously stored in `target/build`)
* [incremental build](https://doc.rust-lang.org/rustc/codegen-options/index.html#incremental) output from rustc
* fingerprint files used by Cargo for rebuild detection
* scratchpad used for `cargo package` verify step
* Cache of rustc invocations (`.rustc_info.json`)
* "pre and non uplifted" binary executables. (ie. bins for `examples` that contain the hash in the name, bins for `benches`, proc macros, build scripts)
* `CARGO_TARGET_TMPDIR` files (see rational for this [here](https://github.com/rust-lang/cargo/issues/14125#issuecomment-2258708917))
* [future-incompat-report](https://doc.rust-lang.org/cargo/reference/future-incompat-report.html)'s `.future-incompat-report.json` file

## Feature Gating Strategy

We are following the "Ignore the feature that is used without a gate" approach as described [here](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/features/index.html).

The rational for this is: 
The `build.build-dir` is likely going to be set by by users "globally" (ie. `$CARGO_HOME/config.toml`) to set a shared build directory to reduce rebuilding dependencies. For users that multiple cargo versions having having an error would be disrupted.
The fallback behavior is to revert to the behavior of the current stable release (building in `$CARGO_TARGET_DIR`)

## Testing Strategy 

* We have the existing Cargo testsuite to be sure we do not introduce regressions.
  * I have also run the testsuite locally with the cli flag remove to verify all tests pass with the default build dir (which falls back to the target dir)
* For testing thus far, I have been using small hello world project with a few dependencies like `rand` to verify files are being output to the correct directory.
* When this PR is closer to merging, I plan to test with some larger projects with more dependencies, build scripts, ect.
* Other testing recommendations are welcome :bow: 

This is probably best reviewed commit by commit. I documented each commit. 
I tied to follow the atomic commits recommendation in the Cargo contributors guide, but I split out some commits for ease of review. (Otherwise I think this would have ended up being 1 or 2 large commits :sweat_smile:)

## Questions

- [x] What is the expected behavior of `cargo clean`? 
  * At the time of writing it only cleans `target` and does not impact the build-dir but this is easily changable.
  * Answer: See https://github.com/rust-lang/cargo/pull/15104#issuecomment-2616369862
- [x] When using `cargo package` are was expecting just the `.crate` file to be in `target` while all other output be stored in `build.build-dir`?  Not sure if we consider things like `Cargo.toml`, ` Cargo.toml.orig`, `.cargo_vcs_info.json` part of the user facing interface.
  * Current consensus is that only `.crate` is considered a final artifact 
- [x] Where should `cargo doc` output go? HTML/JS for many crates can be pretty large. Moving to the build-dir would help reduce duplication if we find the that acceptable. For `cargo doc --open` this is not a problem but may be problematic for other use cases?
  * Answer: https://github.com/rust-lang/cargo/pull/15104#issuecomment-2619760470
- [x] Are bins generated from `benches` considered final artifacts?
  * Since bins from `examples` are considered final artifacts, it seems natural that `benches` should also be considered final artifacts. However, unlike `examples` the `benches` bins are stored in `target/{profile}/deps` instead of a dedicated directory (like `target/{profile}/examples`). We could move them into a dedicated directory (`target/{profile}/benches`) but that mean would also be changing the structure of the `target` directory which feels out of scope for this change. If we decide that `benches` are final artifacts, it would probably be better to consider that changes as part of https://github.com/rust-lang/cargo/issues/6790
  * Answer: https://github.com/rust-lang/cargo/pull/15104#issuecomment-2636519011
- [ ] [Do we want to include a `CARGO_BUILD_DIR` shortcut env var?](https://github.com/rust-lang/cargo/pull/15104#discussion_r1943404348) 
  * The current commit (2af0c918d415b4de03e627459c3594826ae03cfb) has included the `CARGO_BUILD_DIR` shortcut. This can be removed before merging if there a good reason to.
 
## TODO

- Implementation 
  - [x] Add support in `cargo clean`
  - [ ] ~~Implement templating for `build.build-dir`~~
     * Will implement in a follow up PR [as suggested](https://github.com/rust-lang/cargo/pull/15104#issuecomment-2636514628)
  - [x] Fix issue with `target/examples` still containing "pre-uplifted" binaries
  - [x] Verify `build-dir` with non-bin crate types
- Prepare for review
  - [x] Clean up/improve docs
  - [x] Review tests and add more as needed
  - [x] Fix tests in CI (Windows is currently failing)
  - [x] Clean up commits
  - [ ] Resolve remaining questions
- [x] Request review

### DIFF
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -839,7 +839,7 @@ impl LocalFingerprint {
         mtime_cache: &mut HashMap<PathBuf, FileTime>,
         checksum_cache: &mut HashMap<PathBuf, Checksum>,
         pkg: &Package,
-        target_root: &Path,
+        build_root: &Path,
         cargo_exe: &Path,
         gctx: &GlobalContext,
     ) -> CargoResult<Option<StaleItem>> {
@@ -852,8 +852,8 @@ impl LocalFingerprint {
             // the `dep_info` file itself whose mtime represents the start of
             // rustc.
             LocalFingerprint::CheckDepInfo { dep_info, checksum } => {
-                let dep_info = target_root.join(dep_info);
-                let Some(info) = parse_dep_info(pkg_root, target_root, &dep_info)? else {
+                let dep_info = build_root.join(dep_info);
+                let Some(info) = parse_dep_info(pkg_root, build_root, &dep_info)? else {
                     return Ok(Some(StaleItem::MissingFile(dep_info)));
                 };
                 for (key, previous) in info.env.iter() {
@@ -910,7 +910,7 @@ impl LocalFingerprint {
             LocalFingerprint::RerunIfChanged { output, paths } => Ok(find_stale_file(
                 mtime_cache,
                 checksum_cache,
-                &target_root.join(output),
+                &build_root.join(output),
                 paths.iter().map(|p| (pkg_root.join(p), None)),
                 false,
             )),
@@ -1153,7 +1153,7 @@ impl Fingerprint {
         mtime_cache: &mut HashMap<PathBuf, FileTime>,
         checksum_cache: &mut HashMap<PathBuf, Checksum>,
         pkg: &Package,
-        target_root: &Path,
+        build_root: &Path,
         cargo_exe: &Path,
         gctx: &GlobalContext,
     ) -> CargoResult<()> {
@@ -1261,7 +1261,7 @@ impl Fingerprint {
                 mtime_cache,
                 checksum_cache,
                 pkg,
-                target_root,
+                build_root,
                 cargo_exe,
                 gctx,
             )? {
@@ -1449,13 +1449,13 @@ fn calculate(build_runner: &mut BuildRunner<'_, '_>, unit: &Unit) -> CargoResult
 
     // After we built the initial `Fingerprint` be sure to update the
     // `fs_status` field of it.
-    let target_root = target_root(build_runner);
+    let build_root = build_root(build_runner);
     let cargo_exe = build_runner.bcx.gctx.cargo_exe()?;
     fingerprint.check_filesystem(
         &mut build_runner.mtime_cache,
         &mut build_runner.checksum_cache,
         &unit.pkg,
-        &target_root,
+        &build_root,
         cargo_exe,
         build_runner.bcx.gctx,
     )?;
@@ -1493,7 +1493,7 @@ fn calculate_normal(
     };
 
     // Afterwards calculate our own fingerprint information.
-    let target_root = target_root(build_runner);
+    let build_root = build_root(build_runner);
     let local = if unit.mode.is_doc() || unit.mode.is_doc_scrape() {
         // rustdoc does not have dep-info files.
         let fingerprint = pkg_fingerprint(build_runner.bcx, &unit.pkg).with_context(|| {
@@ -1505,7 +1505,7 @@ fn calculate_normal(
         vec![LocalFingerprint::Precalculated(fingerprint)]
     } else {
         let dep_info = dep_info_loc(build_runner, unit);
-        let dep_info = dep_info.strip_prefix(&target_root).unwrap().to_path_buf();
+        let dep_info = dep_info.strip_prefix(&build_root).unwrap().to_path_buf();
         vec![LocalFingerprint::CheckDepInfo {
             dep_info,
             checksum: build_runner.bcx.gctx.cli_unstable().checksum_freshness,
@@ -1714,7 +1714,7 @@ fn build_script_local_fingerprints(
     // longstanding bug, in Cargo. Recent refactorings just made it painfully
     // obvious.
     let pkg_root = unit.pkg.root().to_path_buf();
-    let target_dir = target_root(build_runner);
+    let build_dir = build_root(build_runner);
     let env_config = Arc::clone(build_runner.bcx.gctx.env_config()?);
     let calculate =
         move |deps: &BuildDeps, pkg_fingerprint: Option<&dyn Fn() -> CargoResult<String>>| {
@@ -1747,7 +1747,7 @@ fn build_script_local_fingerprints(
             // them all here.
             Ok(Some(local_fingerprints_deps(
                 deps,
-                &target_dir,
+                &build_dir,
                 &pkg_root,
                 &env_config,
             )))
@@ -1783,7 +1783,7 @@ fn build_script_override_fingerprint(
 /// [`RunCustomBuild`]: crate::core::compiler::CompileMode::RunCustomBuild
 fn local_fingerprints_deps(
     deps: &BuildDeps,
-    target_root: &Path,
+    build_root: &Path,
     pkg_root: &Path,
     env_config: &Arc<HashMap<String, OsString>>,
 ) -> Vec<LocalFingerprint> {
@@ -1796,7 +1796,7 @@ fn local_fingerprints_deps(
         // absolute prefixes from them.
         let output = deps
             .build_script_output
-            .strip_prefix(target_root)
+            .strip_prefix(build_root)
             .unwrap()
             .to_path_buf();
         let paths = deps
@@ -1854,10 +1854,10 @@ pub fn dep_info_loc(build_runner: &mut BuildRunner<'_, '_>, unit: &Unit) -> Path
     build_runner.files().fingerprint_file_path(unit, "dep-")
 }
 
-/// Returns an absolute path that target directory.
+/// Returns an absolute path that build directory.
 /// All paths are rewritten to be relative to this.
-fn target_root(build_runner: &BuildRunner<'_, '_>) -> PathBuf {
-    build_runner.bcx.ws.target_dir().into_path_unlocked()
+fn build_root(build_runner: &BuildRunner<'_, '_>) -> PathBuf {
+    build_runner.bcx.ws.build_dir().into_path_unlocked()
 }
 
 /// Reads the value from the old fingerprint hash file and compare.
--- a/src/cargo/core/compiler/future_incompat.rs
+++ b/src/cargo/core/compiler/future_incompat.rs
@@ -91,7 +91,7 @@ pub struct Diagnostic {
     pub level: String,
 }
 
-/// The filename in the top-level `target` directory where we store
+/// The filename in the top-level `build-dir` directory where we store
 /// the report
 const FUTURE_INCOMPAT_FILE: &str = ".future-incompat-report.json";
 /// Max number of reports to save on disk.
@@ -166,7 +166,7 @@ impl OnDiskReports {
         }
         let on_disk = serde_json::to_vec(&self).unwrap();
         if let Err(e) = ws
-            .target_dir()
+            .build_dir()
             .open_rw_exclusive_create(
                 FUTURE_INCOMPAT_FILE,
                 ws.gctx(),
@@ -191,7 +191,7 @@ impl OnDiskReports {
 
     /// Loads the on-disk reports.
     pub fn load(ws: &Workspace<'_>) -> CargoResult<OnDiskReports> {
-        let report_file = match ws.target_dir().open_ro_shared(
+        let report_file = match ws.build_dir().open_ro_shared(
             FUTURE_INCOMPAT_FILE,
             ws.gctx(),
             "Future incompatible report",
--- a/src/cargo/core/compiler/layout.rs
+++ b/src/cargo/core/compiler/layout.rs
@@ -128,6 +128,8 @@ pub struct Layout {
     fingerprint: PathBuf,
     /// The directory for examples: `$dest/examples`
     examples: PathBuf,
+    /// The directory for pre-uplifted examples: `$build-dir/debug/examples`
+    build_examples: PathBuf,
     /// The directory for rustdoc output: `$root/doc`
     doc: PathBuf,
     /// The directory for temporary data of integration tests and benches: `$dest/tmp`
@@ -135,6 +137,11 @@ pub struct Layout {
     /// The lockfile for a build (`.cargo-lock`). Will be unlocked when this
     /// struct is `drop`ped.
     _lock: FileLock,
+    /// Same as `_lock` but for the build directory.
+    ///
+    /// Will be `None` when the build-dir and target-dir are the same path as we cannot
+    /// lock the same path twice.
+    _build_lock: Option<FileLock>,
 }
 
 impl Layout {
@@ -150,15 +157,22 @@ impl Layout {
         dest: &str,
     ) -> CargoResult<Layout> {
         let mut root = ws.target_dir();
+        let mut build_root = ws.build_dir();
         if let Some(target) = target {
             root.push(target.short_name());
+            build_root.push(target.short_name());
         }
+        let build_dest = build_root.join(dest);
         let dest = root.join(dest);
         // If the root directory doesn't already exist go ahead and create it
         // here. Use this opportunity to exclude it from backups as well if the
         // system supports it since this is a freshly created folder.
         //
         paths::create_dir_all_excluded_from_backups_atomic(root.as_path_unlocked())?;
+        if root != build_root {
+            paths::create_dir_all_excluded_from_backups_atomic(build_root.as_path_unlocked())?;
+        }
+
         // Now that the excluded from backups target root is created we can create the
         // actual destination (sub)subdirectory.
         paths::create_dir_all(dest.as_path_unlocked())?;
@@ -167,23 +181,37 @@ impl Layout {
         // directory, so just lock the entire thing for the duration of this
         // compile.
         let lock = dest.open_rw_exclusive_create(".cargo-lock", ws.gctx(), "build directory")?;
+
+        let build_lock = if root != build_root {
+            Some(build_dest.open_rw_exclusive_create(
+                ".cargo-lock",
+                ws.gctx(),
+                "build directory",
+            )?)
+        } else {
+            None
+        };
         let root = root.into_path_unlocked();
+        let build_root = build_root.into_path_unlocked();
         let dest = dest.into_path_unlocked();
-        let deps = dest.join("deps");
+        let build_dest = build_dest.as_path_unlocked();
+        let deps = build_dest.join("deps");
         let artifact = deps.join("artifact");
 
         Ok(Layout {
             deps,
-            build: dest.join("build"),
+            build: build_dest.join("build"),
             artifact,
-            incremental: dest.join("incremental"),
-            fingerprint: dest.join(".fingerprint"),
+            incremental: build_dest.join("incremental"),
+            fingerprint: build_dest.join(".fingerprint"),
             examples: dest.join("examples"),
+            build_examples: build_dest.join("examples"),
             doc: root.join("doc"),
-            tmp: root.join("tmp"),
+            tmp: build_root.join("tmp"),
             root,
             dest,
             _lock: lock,
+            _build_lock: build_lock,
         })
     }
 
@@ -193,6 +221,7 @@ impl Layout {
         paths::create_dir_all(&self.incremental)?;
         paths::create_dir_all(&self.fingerprint)?;
         paths::create_dir_all(&self.examples)?;
+        paths::create_dir_all(&self.build_examples)?;
         paths::create_dir_all(&self.build)?;
 
         Ok(())
@@ -209,6 +238,10 @@ impl Layout {
     /// Fetch the examples path.
     pub fn examples(&self) -> &Path {
         &self.examples
+    }
+    /// Fetch the build examples path.
+    pub fn build_examples(&self) -> &Path {
+        &self.build_examples
     }
     /// Fetch the doc path.
     pub fn doc(&self) -> &Path {
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -297,7 +297,7 @@ fn rustc(
     let exec = exec.clone();
 
     let root_output = build_runner.files().host_dest().to_path_buf();
-    let target_dir = build_runner.bcx.ws.target_dir().into_path_unlocked();
+    let build_dir = build_runner.bcx.ws.build_dir().into_path_unlocked();
     let pkg_root = unit.pkg.root().to_path_buf();
     let cwd = rustc
         .get_cwd()
@@ -455,7 +455,7 @@ fn rustc(
                 &dep_info_loc,
                 &cwd,
                 &pkg_root,
-                &target_dir,
+                &build_dir,
                 &rustc,
                 // Do not track source files in the fingerprint for registry dependencies.
                 is_local,
@@ -555,7 +555,7 @@ fn link_targets(
         let path = unit
             .pkg
             .manifest()
-            .metabuild_path(build_runner.bcx.ws.target_dir());
+            .metabuild_path(build_runner.bcx.ws.build_dir());
         target.set_src_path(TargetSourcePath::Path(path));
     }
 
--- a/src/cargo/ops/cargo_package/verify.rs
+++ b/src/cargo/ops/cargo_package/verify.rs
@@ -45,9 +45,11 @@ pub fn run_verify(
 
     tar.file().seek(SeekFrom::Start(0))?;
     let f = GzDecoder::new(tar.file());
-    let dst = tar
-        .parent()
-        .join(&format!("{}-{}", pkg.name(), pkg.version()));
+    let dst = ws.build_dir().as_path_unlocked().join(&format!(
+        "package/{}-{}",
+        pkg.name(),
+        pkg.version()
+    ));
     if dst.exists() {
         paths::remove_dir_all(&dst)?;
     }
@@ -63,7 +65,14 @@ pub fn run_verify(
     let mut src = PathSource::new(&dst, id, ws.gctx());
     let new_pkg = src.root_package()?;
     let pkg_fingerprint = hash_all(&dst)?;
-    let mut ws = Workspace::ephemeral(new_pkg, gctx, None, true)?;
+
+    let target_dir = if gctx.cli_unstable().build_dir {
+        Some(ws.build_dir())
+    } else {
+        None
+    };
+
+    let mut ws = Workspace::ephemeral(new_pkg, gctx, target_dir, true)?;
     if let Some(local_reg) = local_reg {
         ws.add_local_overlay(
             local_reg.upstream,
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -421,11 +421,8 @@ impl GlobalContext {
 
     /// Gets the path to the `rustc` executable.
     pub fn load_global_rustc(&self, ws: Option<&Workspace<'_>>) -> CargoResult<Rustc> {
-        let cache_location = ws.map(|ws| {
-            ws.target_dir()
-                .join(".rustc_info.json")
-                .into_path_unlocked()
-        });
+        let cache_location =
+            ws.map(|ws| ws.build_dir().join(".rustc_info.json").into_path_unlocked());
         let wrapper = self.maybe_get_tool("rustc_wrapper", &self.build_config()?.rustc_wrapper);
         let rustc_workspace_wrapper = self.maybe_get_tool(
             "rustc_workspace_wrapper",
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -603,7 +603,7 @@ impl GlobalContext {
     ///
     /// Returns `None` if the user has not chosen an explicit directory.
     ///
-    /// Callers should prefer `Workspace::target_dir` instead.
+    /// Callers should prefer [`Workspace::target_dir`] instead.
     pub fn target_dir(&self) -> CargoResult<Option<Filesystem>> {
         if let Some(dir) = &self.target_dir {
             Ok(Some(dir.clone()))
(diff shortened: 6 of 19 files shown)


tests/testsuite/build_dir.rs
weihanglo, line 1: We usually add some tests verifying a nightly flag is really gated.

src/cargo/core/compiler/build_runner/mod.rs
```diff
@@ -290,7 +290,9 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
                 });
             }
 
-            super::output_depinfo(&mut self, unit)?;
+            if !self.bcx.gctx.cli_unstable().build_dir {
+                super::output_depinfo(&mut self, unit)?;
```
weihanglo, line 294: Why do we stop generating this when `build-dir` is enabled?

weihanglo (commented):
Just want to make sure I didn't miss something. From what I can tell these directories/files have been removed right?

* `target/<profile>/.metabuild`
* `target/<profile>/.fingerprint`
* `target/<profile>/deps`
* `target/<profile>/incremental`
* `target/<profile>/build`
* `target/.cargo-lock`
* `target/tmp`
* `target/.rustc_info.json`

src/cargo/core/compiler/build_runner/mod.rs
ranger-ross: Sorry, actually I believe this is a mistake on my part :sweat_smile: 

My original assumption was that the `/target/debug/{project-name}.d` file was the same file as `/target/debug/deps/{project-name}-{hash}.d`, so I was trying to avoid copying it into `target`.
But now that I compared the files they appear to be different so it looks like that was a bad assumption.

I believe this should probably be considered an intermediate build artifact so it should probably be moved to the `build-dir` location. 

Modifying this file in particular was a bit tricky as it was not just replacing `.target_dir()` with `.build_dir()`.
I think I will probably need to go to where the [`OutputFile`](https://github.com/rust-lang/cargo/blob/e63457a4cecfc3562d87197c451a3a3146f5e1a6/src/cargo/core/compiler/build_runner/compilation_files.rs#L136) is created and modify the `hardlink` to be the new build-dir path.

ranger-ross:
> Just want to make sure I didn't miss something. From what I can tell these directories/files have been removed right?
> 
> * `target/<profile>/.metabuild`
> * `target/<profile>/.fingerprint`
> * `target/<profile>/deps`
> * `target/<profile>/incremental`
> * `target/<profile>/build`
> * `target/.cargo-lock`
> * `target/tmp`
> * `target/.rustc_info.json`

Yes, that with the exception of `target/.cargo-lock`. 
I think we will still want this cargo lock for backwards compatibility with previous versions of cargo.

Ideally in the longer term it can be removed in favor of fine grain locking like https://github.com/rust-lang/cargo/issues/4282

So a typical `target` directory will be something like 
```
target
├── CACHEDIR.TAG
└── debug
    ├── .cargo-lock
    ├── examples
    └── hello_world // (the binary)
```

src/doc/src/reference/unstable.md
```diff
@@ -237,6 +238,22 @@ This can also be specified in `.cargo/config.toml` files.
...
+The final artifacts will also be contained in this directory but will be hardlinked into the artifact directory (usually `target`)
+
+This can be specified in `.cargo/config.toml` files using the `build.build-dir`.
+
+Example:
+
+```toml
+[build]
+build-dir = "out"
+```
```
epage, line 255: Could you include documentation in this that mirrors what would be in https://doc.rust-lang.org/cargo/reference/config.html

src/cargo/core/compiler/layout.rs
```diff
@@ -135,6 +135,12 @@ pub struct Layout {
     /// The lockfile for a build (`.cargo-lock`). Will be unlocked when this
     /// struct is `drop`ped.
     _lock: FileLock,
+    /// Same as `_lock` but for the build directory.
+    ///
+    /// NOTE: `_lock` and `_build_lock` can eventually be merged once #14125 is stablized.
+    /// Will be `None` when the build-directory and target-directory are the same path as we cannot
+    /// lock the same path twice.
```
epage, line 142: How can we merge these locks?

In the short term, we'll have
- target-dir
- build-dir

which both need locking

In the long term, we'll have
- artifact-dir
- build-dir

which will both need locking

tests/testsuite/build_dir.rs
epage, line 1: Also, we generally add (passing) tests before the change so the affect of the change is obvious.

This is more straightforward when fixing a bug.  In a case like this, a common option is to write the tests without the new feature or using something similar ( like `target-dir`) and then in the commit where the feature is added, you switch over the tests.

epage:
> What is the expected behavior of cargo clean? 

It should clean the build dir

epage:
> When using cargo package are was expecting just the .crate file to be in target while all other output be stored in build.build-dir? Not sure if we consider things like Cargo.toml,  Cargo.toml.orig, .cargo_vcs_info.json part of the user facing interface.

imo The artifact for `cargo package` is the `.crate`.  Everything else is part of the "build" process.

epage:
Can we call out explicitly what our testing strategy is?

We likely should also explicitly document in the PR what is considered an artifact and what is a build output and make sure we have tests for these.

src/cargo/core/compiler/layout.rs
ranger-ross: Sorry, you are right. Not quiet sure what I was thinking when I wrote this comment.

ranger-ross:
One other question that came to my mind was the output of `cargo doc`. HTML/JS for many crates can be pretty large. Moving to the build-dir would help reduce duplication if we find the that acceptable. For `cargo doc --open` this is not a problem but may be problematic for other use cases?

Perhaps symlinking the `index.html` from the build dir into `target` could be an option if we care about keeping an entry point.

ranger-ross:
> Can we call out explicitly what our testing strategy is?
> 
> 
> 
> We likely should also explicitly document in the PR what is considered an artifact and what is a build output and make sure we have tests for these.


@epage sure, I updated the PR description but let me know if I missed anything.

epage:
> One other question that came to my mind was the output of cargo doc. HTML/JS for many crates can be pretty large. Moving to the build-dir would help reduce duplication if we find the that acceptable. For cargo doc --open this is not a problem but may be problematic for other use cases?

imo `cargo doc`s output is an artifact that people will want access to.  I suspect it'd be a breaking change to move it out of `target-dir`.

epage:
> depinfo files (.d files)

There are multiple types of depinfo files.  I suspect the ones next to final artifacts are also considered final artifacts, see https://doc.rust-lang.org/cargo/reference/build-cache.html#dep-info-files

epage:
FYI I added to the PR description a couple more intermediate artifacts

> * rlibs and debug info from dependencies
> * build script `OUT_DIR`

When are workspace member rlibs considered final artifacts?  We're putting them in `target/<profile>` at times, so I take it that has already been answered.

epage:
> > Can we call out explicitly what our testing strategy is?
> > We likely should also explicitly document in the PR what is considered an artifact and what is a build output and make sure we have tests for these.
> 
> @epage sure, I updated the PR description but let me know if I missed anything.

Sorry if I wasn't clear but my discussion of testing strategy was in the context of tracking the categorization of artifacts.  I'm not saying we have to exhaustively test it but we should think about it and document the choice made and why.

src/cargo/core/compiler/build_runner/mod.rs
ranger-ross: Given [this](https://github.com/rust-lang/cargo/pull/15104#issuecomment-2619773382) comment I am going to revert this change

ranger-ross:
I spent some more time this weekend making some progress on this PR.

Here are my updates

* Fixed some issues to make `cargo clean` & `cargo package` work with `build.build-dir`
* Added templating support for `build.build-dir`
* Cleanup the commits a bit and tried to follow the request in [this comment](https://github.com/rust-lang/cargo/pull/15104#discussion_r1930859849) regarding tests

I updated the PR description with more details as well. notably:

* Added another question I would like input form the Cargo team on.
  *  Are bins generated from benches considered final artifacts? (see PR desc for more details)
* Updated the TODO list with a more detailed list of tasks before marking this PR as ready for review.
* Cleaned up the "What is a final artifact vs an intermediate build artifact" section
  * Also added the following file types to the list of intermediate build artifacts
    *  "pre-uplifted" binary executables. (ie. bins for `examples` that contain the hash in the name)
    * `CARGO_TARGET_TMPDIR` files (see rational for this [here](https://github.com/rust-lang/cargo/issues/14125#issuecomment-2258708917))

epage:
> Added templating support for build.build-dir

I haven't looked but would it work to split that out into a follow up PR?  Keeping PRs focused can help with ensuring a smooth review and that we are less likely to overlook something.

> Are bins generated from benches considered final artifacts? (see PR desc for more details)

My take is that if it isn't staged/uplifted, then its not a final artifact.

I assume this also applies to test binaries, build scripts, and proc-macros.

ranger-ross:
> I haven't looked but would it work to split that out into a follow up PR? Keeping PRs focused can help with ensuring a smooth review and that we are less likely to overlook something.

Sure. Its not a ton of code, but I have no problems breaking that out into anther PR to reduce the scope of this one :+1:

ranger-ross:
> > Are bins generated from benches considered final artifacts? (see PR desc for more details)
> 
> My take is that if it isn't staged/uplifted, then its not a final artifact.
> 
> I assume this also applies to test binaries, build scripts, and proc-macros.

Thanks, I think that sounds reasonable.
I will add that to the classification list in PR description with this as well.

tests/testsuite/build_dir.rs
```diff
@@ -0,0 +1,193 @@
...
+    );
+}
+
+#[track_caller]
+fn assert_executeable_exists(path: PathBuf, name: &str, exists: bool) {
+    let executable = if cfg!(target_os = "windows") {
+        format!("{name}.exe")
+    } else {
+        name.to_string()
+    };
```
epage, line 226: You can always insert https://doc.rust-lang.org/std/env/consts/constant.EXE_SUFFIX.html, making this not have to deal with conditional compilation

tests/testsuite/build_dir.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    assert_eq!(path.join(executable).exists(), exists);
```
epage, line 228: Can you emit the path in the assert message?

tests/testsuite/build_dir.rs
```diff
@@ -96,51 +104,79 @@ fn cargo_package_should_output_to_target_dir() {
...
+        .masquerade_as_nightly_cargo(&["build-dir"])
+        .enable_mac_dsym()
+        .run();
+
+    assert!(!p.root().join("build").exists());
+    assert!(!p.root().join("target").exists());
+}
+
+#[cargo_test]
+fn verify_build_dir_is_disabled_by_feature_flag() {
```
epage, line 161: Can you add this test to the first commit?

tests/testsuite/build_dir.rs
```diff
@@ -96,51 +104,79 @@ fn cargo_package_should_output_to_target_dir() {
...
+
+    p.cargo("build")
         .masquerade_as_nightly_cargo(&[])
         .enable_mac_dsym()
         .run();
 
+    assert_build_dir(p.root().join("target"), "debug", true);
+    assert_executeable_exists(p.root().join("target/debug"), "foo", true);
     assert!(!p.root().join("build").exists());
 }
```
epage, line 181: Looks like you are taking the approach "Ignore the feature that is used without a gate " (see https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/features/index.html)

Could you speak to that (and update the PR description with that)

src/cargo/util/context/mod.rs
```diff
@@ -634,6 +634,35 @@ impl GlobalContext {
...
+    /// Fallsback to the target directory if not specified.
+    ///
+    /// Callers should prefer [`Workspace::build_dir`] instead.
+    pub fn build_dir(&self) -> CargoResult<Option<Filesystem>> {
+        if !self.cli_unstable().build_dir {
+            return self.target_dir();
+        }
+
+        if let Some(val) = &self.build_config()?.build_dir {
+            let path = val.resolve_path(self);
```
epage, line 645: Compared to `pub fn target_dir()`,
- We don't have `self.build_dir` which is fine as we don't have a CLI flag
- We don't check `CARGO_BUILD_DIR` env variable as a shortcut for `CARGO_BUIILD_BUILD_DIR`.  **Should we have this shortcut?**

src/cargo/util/context/mod.rs
```diff
@@ -634,6 +634,35 @@ impl GlobalContext {
...
+                    val.value().definition
+                )
+            }
+
+            Ok(Some(Filesystem::new(path)))
+        } else {
+            // For now, fallback to the previous implementation.
+            // This will change in the future.
+            return self.target_dir();
+        }
```
epage, line 660: Should we handle this fallback here or in `Workspace::build_dir`?

(I can see both ways)

src/cargo/core/compiler/fingerprint/mod.rs
```diff
@@ -1857,7 +1857,7 @@ pub fn dep_info_loc(build_runner: &mut BuildRunner<'_, '_>, unit: &Unit) -> Path
 /// Returns an absolute path that target directory.
 /// All paths are rewritten to be relative to this.
 fn target_root(build_runner: &BuildRunner<'_, '_>) -> PathBuf {
-    build_runner.bcx.ws.target_dir().into_path_unlocked()
+    build_runner.bcx.ws.build_dir().into_path_unlocked()
```
epage, line 1860: Should `target_root` be renamed?

src/cargo/core/compiler/future_incompat.rs
```diff
@@ -166,7 +166,7 @@ impl OnDiskReports {
         }
         let on_disk = serde_json::to_vec(&self).unwrap();
         if let Err(e) = ws
-            .target_dir()
+            .build_dir()
             .open_rw_exclusive_create(
                 FUTURE_INCOMPAT_FILE,
```
epage, line 171: I'm not seeing this discussed in the PR as intermediate vs final.

src/cargo/core/compiler/mod.rs
```diff
@@ -297,7 +297,7 @@ fn rustc(
     let exec = exec.clone();
 
     let root_output = build_runner.files().host_dest().to_path_buf();
-    let target_dir = build_runner.bcx.ws.target_dir().into_path_unlocked();
+    let target_dir = build_runner.bcx.ws.build_dir().into_path_unlocked();
```
epage, line 300: Should we rename this variable?

src/cargo/util/context/mod.rs
epage: Looks like it is in the follow up commit, duplicating the fallback

tests/testsuite/build_dir.rs
```diff
@@ -12,19 +12,22 @@ fn binary_with_debug() {
...
-        .masquerade_as_nightly_cargo(&[])
+    p.cargo("build -Z unstable-options -Z build-dir")
+        .masquerade_as_nightly_cargo(&["build-dir"])
         .enable_mac_dsym()
         .run();
 
     assert_build_dir(p.root().join("build"), "debug", true);
-    assert_executeable_exists(p.root().join("build/debug"), "foo", true);
-    assert!(!p.root().join("target").exists());
+    assert_build_dir(p.root().join("target"), "debug", false);
```
epage, line 27: In reading these tests, the intent of the assertion isn't too clear (asserting that the target dir is a build dir? what does `false` mean?).  I think it'd be better to have two separate functions.  If they truly must share an implementation, then they can just be wrappers on top of that.

tests/testsuite/build_dir.rs
ranger-ross: Sure, I updated the PR description with a feature gating section

src/cargo/core/compiler/future_incompat.rs
ranger-ross: Thanks for pointing this out. I reverted the change and added the following to the "final artifacts" category 
> * reports generated from `cargo report` like [future-incompat-report](https://doc.rust-lang.org/cargo/reference/future-incompat-report.html)

src/cargo/util/context/mod.rs
ranger-ross: yes, I added the fallback in both places "just in case"

src/cargo/core/compiler/future_incompat.rs
epage: From those docs, it sounds like the user doesn't interact directly with the files but views them only through a command.   That sounds more like an intermediate artifact.

src/cargo/ops/cargo_package/verify.rs
```diff
@@ -45,9 +45,14 @@ pub fn run_verify(
...
-        .parent()
-        .join(&format!("{}-{}", pkg.name(), pkg.version()));
+    let dst = if gctx.cli_unstable().build_dir {
+        ws.build_dir()
+            .as_path_unlocked()
+            .join(&format!("package/{}-{}", pkg.name(), pkg.version()))
+    } else {
+        tar.parent()
+            .join(&format!("{}-{}", pkg.name(), pkg.version()))
+    };
```
epage, line 55: Why make this conditional?

src/doc/src/reference/unstable.md
```diff
@@ -237,6 +238,27 @@ This can also be specified in `.cargo/config.toml` files.
...
+
+
+This can be specified in `.cargo/config.toml` files using the `build.build-dir`.
+
+Example:
+
+```toml
+[build]
+build-dir = "out"
+```
```
epage, line 260:
````suggestion
* Tracking Issue: [#14125](https://github.com/rust-lang/cargo/issues/14125)

The directory where intermediate build artifacts will be stored.
Intermediate artifacts are produced by Rustc/Cargo during the build process.

```toml
[build]
build-dir = "out"
```

### `build.build-dir`

* Type: string (path)
* Default: Defaults to the value of `build.target-dir`
* Environment: `CARGO_BUILD_BUILD_DIR` or `CARGO_BUILD_DIR`

The path to where internal files used as part of the build are placed.

````

File: src/cargo/ops/cargo_package/verify.rs (hunk 14 above)
Comment (ranger-ross) [reply src/cargo/ops/cargo_package/verify.rs:55]: Good point, looking a bit closer, it appears these should always be equal since the `tar` should [always be placed in `target/package`](https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/cargo_package/mod.rs#L104).

I'll simplify this to just 
```rs
let dst = ws.build_dir()
            .as_path_unlocked()
            .join(&format!("package/{}-{}", pkg.name(), pkg.version()))
```

tests/testsuite/build_dir.rs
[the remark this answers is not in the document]
ranger-ross: Sorry, I think I misunderstood your request. (perhaps multiple times)

I added a doc comment at the [top of `build_dir.rs`](https://github.com/rust-lang/cargo/pull/15104/commits/b43e226b8039f1140d381e19790a9d651548b6d6#diff-d37a8e4e98440be7f035daa36682456743a69ede1307f7e189f219179725bf80R3-R10) with a summary of the test strategy for that file and the rational why.

Let me know if that is what you were asking for.

src/cargo/core/compiler/future_incompat.rs
ranger-ross: Hmm good point. Since it is JSON I assumed it was a user facing file. But looking at the docs, they do not reference that file so i guess its meant to be internal. 

I think in that cause it should belong in the build-dir.
I updated that list in the PR description and will revert the code for this.

---

Also, one thing that just came to my mind, is the HTML report of the `--timings` flag in Cargo.
I added this to the final artifacts list in the PR description.

tests/testsuite/build_dir.rs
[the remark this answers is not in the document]
epage: In the ideal scenario, we verify
- each file listed under artifact-dir vs build-dir is in the right place
- ensure new files get categorized correctly

I'm assuming we can't meet that ideal.  I would like to understand why we can't meet that ideal and understand how close we should get to it along with a call out of what gaps are left.

src/cargo/util/context/mod.rs
```diff
@@ -634,6 +634,44 @@ impl GlobalContext {
...
+    /// The directory to use for intermediate build artifacts.
+    ///
+    /// Falls back to the target directory if not specified.
+    ///
+    /// Callers should prefer [`Workspace::build_dir`] instead.
+    pub fn build_dir(&self) -> CargoResult<Option<Filesystem>> {
+        if !self.cli_unstable().build_dir {
+            return self.target_dir();
+        }
+        if let Some(dir) = self.get_env_os("CARGO_BUILD_DIR") {
```
weihanglo, line 643: I understand `CARGO_BUILD_DIR` is proposed here because we already have `CARGO_TARGET_DIR`. However I feel like `CARGO_TARGET_DIR` was a mistake.

The `build.build-dir` config already comes with a `CARGO_BUILD_BUILD_DIR` environment variable for free. We may want to stick with it and stop stacking up extra environment variables.

src/cargo/core/compiler/layout.rs
```diff
@@ -150,15 +157,22 @@ impl Layout {
...
         }
+        let build_dest = build_root.join(dest);
         let dest = root.join(dest);
         // If the root directory doesn't already exist go ahead and create it
         // here. Use this opportunity to exclude it from backups as well if the
         // system supports it since this is a freshly created folder.
         //
         paths::create_dir_all_excluded_from_backups_atomic(root.as_path_unlocked())?;
+        if root != build_root {
+            paths::create_dir_all_excluded_from_backups_atomic(build_root.as_path_unlocked())?;
```
weihanglo, line 173: Not really on topic, but this reminds me <https://github.com/rust-lang/cargo/issues/11548> and <https://github.com/rust-lang/cargo/issues/15061>.

For the new build-dir, we don't have backward compatibility issue, and files inside are really intermediate caches. We might want to reconsider the self-ignoring directory approach.

src/cargo/util/context/mod.rs
ranger-ross: > However I feel like `CARGO_TARGET_DIR` was a mistake.

Can I ask why you feel it was a mistake? 

As a user of cargo I think it might be a bit confusing to see `BUILD` show up twice in the same variable name.
It make sense when strictly following the [Cargo configuration](https://doc.rust-lang.org/cargo/reference/config.html#environment-variables), but newcomers are often not aware of this.

I think there is value in providing a shortcut that is easily understandable.

epage: > And the --target-dir flag can just be a alias equivalent to --config 'build.target-dir=foo', and could eliminate the special handling.

The plan was to not offer a `--build-dir` env variable and build-dir would eventually be decoupled from `--target-dir`

ranger-ross: > Could you elaborate a bit more on this? In the book there is a list of variables for each configuration field, which I assume no discovery issue here. (Could expand the doc of CARGO_BUILD_TARGET_DIR, if the other one is removed, yet we won't remove it for sure.)

I think that anyone remotely familiar with Rust can fairly quickly understand `CARGO_TARGET_DIR` does without going to the documentation page. Seeing `CARGO_BUILD_BUILD_DIR` is much less obvious and will probably require some google searching to figure out what it does. 

But perhaps to your later point, if people just set it once in their shared cargo config, perhaps it does not matter that much.

> While I lean towards `CARGO_BUILD_DIR`, starting with fewer features makes it easier to identify when a feature is needed compared to when you offer everything, its hard to tell what is used and could be removed.

I think this is fair. I can remove it from this PR and we can revisit if it's really needed later

tests/testsuite/build_dir.rs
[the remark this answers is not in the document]
ranger-ross: > I'm assuming we can't meet that ideal. I would like to understand why we can't meet that ideal and understand how close we should get to it along with a call out of what gaps are left.

I spent some time adding some additional tests to get a bit closer to this ideal.
I tried to have a test for every bullet point in the `build-dir`/`artifact-dir` list in the PR description. 

Gaps
* Not all files are explicitly checked for various reason.
  * `cargo doc` output contains a lot of files and I suspect the files are subject to change as rustdoc changes which will make this a maintenance burden.
  * `incremental build` file names are just the hash so it is a bit difficult to check explicitly. I settled for simply checking for `incremental` directory in the build-dir.
* Hashes in the file/dir names is a bit tedious to test with the current approach (primarily using `glob` crate).
  * I am currently using a `*` wildcard, but this leaves 
  * It should be possible to improve this in the future (I am thinking by using `regex` and `walkdir` crates together)
* I am not a cargo expert so there may be other files I could have missed that I am just not aware of.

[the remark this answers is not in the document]
epage: Agreed that we shouldn't need to check every individual file generated but each class of artifact file (where does `cargo doc` get written to generally, where does incremental builds get written to, etc)

tests/testsuite/build_dir.rs
```diff
@@ -0,0 +1,589 @@
...
+    assert_build_dir_layout(p.root().join("build"), "debug");
+    assert_exists_patterns_with_base_dir(
+        &p.root(),
+        &[
+            &format!("build/debug/deps/foo*{EXE_SUFFIX}"),
+            "build/debug/deps/foo*.d",
+            "build/debug/.fingerprint/foo-*/bin-foo",
+            "build/debug/.fingerprint/foo-*/bin-foo.json",
+            "build/debug/.fingerprint/foo-*/dep-bin-foo",
+            "build/debug/.fingerprint/foo-*/invoked.timestamp",
```
epage, line 71: We probably only need to check where `.fingerprint/` is written to, not every class of file inside of it

tests/testsuite/build_dir.rs
```diff
@@ -0,0 +1,589 @@
...
+            .run();
+
+        assert_build_dir_layout(p.root().join("build"), "debug");
+        assert_exists_patterns_with_base_dir(&p.root().join("build/debug"), &expected_files);
+        assert_not_exists(&p.root().join("target"));
+    }
+}
+
+#[cargo_test]
+fn binary_with_release() {
```
epage, line 156: I feel like this should be next to `binary_with_debug` as they seem to be closely related

tests/testsuite/build_dir.rs
```diff
@@ -0,0 +1,589 @@
...
+        &p.root(),
+        &[
+            &format!("build/debug/build/foo-*/build-script-build{EXE_SUFFIX}"),
+            "build/debug/build/foo-*/build_script_build*.d",
+            "build/debug/build/foo-*/invoked.timestamp",
+            "build/debug/build/foo-*/out",
+            "build/debug/build/foo-*/out/foo.txt", // Verify OUT_DIR
+            "build/debug/build/foo-*/output",
+            "build/debug/build/foo-*/root-output",
+            "build/debug/build/foo-*/stderr",
```
epage, line 264: Similarly, do we need to check every build script file or just `build/` exists in only one location?

Hmm,. the `build/*/build` repetition could make this confusing.  Should we have the tests call the first one `build_dir`?

tests/testsuite/build_dir.rs
```diff
@@ -47,17 +47,19 @@ fn binary_with_debug() {
...
-            target-dir = "build"
+            target-dir = "target"
+            build-dir = "build"
             "#,
         )
         .build();
 
-    p.cargo("build")
-        .masquerade_as_nightly_cargo(&[])
+    p.cargo("build -Z unstable-options -Z build-dir")
```
epage, line 56: why is `-Zunstable-options` used?

tests/testsuite/build_dir.rs
```diff
@@ -0,0 +1,589 @@
...
+        .enable_mac_dsym()
+        .run();
+
+    assert_build_dir_layout(p.root().join("build"), "debug");
+    assert_exists(&p.root().join(format!("build/debug/foo{EXE_SUFFIX}")));
+    assert_not_exists(&p.root().join("target"));
+}
+
+#[cargo_test]
+fn build_script_should_output_to_build_dir() {
```
epage, line 223: So this is named in the test commit for what the behavior should be while `cargo_tmpdir_should_output_to_target_dir` is named for the current behavior but not renamed while `examples_should_output_to_target_dir` is named for the current behavior and then renamed when the behavior changed (as examples)

We should probably be consistent and ensure the names are correct

tests/testsuite/build_dir.rs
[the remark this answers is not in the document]
epage: Opened more specific comments.  Thanks for exploring this!

tests/testsuite/build_dir.rs
ranger-ross: oops, this is from a copy/paste from the `--artfiact-dir` tests. Since build-dir does not a CLI option its not needed. 
Removed it :+1:

tests/testsuite/build_dir.rs
ranger-ross: > Hmm,. the build/*/build repetition could make this confusing. Should we have the tests call the first one build_dir?

Yeah that makes sense. Also I think I might have made the review commits a bit confusing by naming `target-dir = "build"` in the first test commit and then renamed it in the impl commit. :sweat_smile: 

I changed it to `target-dir = "target-dir"` and only added `build-dir = "build-dir"` in the impl commit.
I think this makes the git diffs much easier to understand.

ranger-ross:
Rebased to pick up the CI fixes in #15222 to hopefully make CI green

tests/testsuite/build_dir.rs
epage: I think "making the names either consistently renamed or consistently always the final behavior" is the only thing left?

ranger-ross: Sorry, I totally forgot to up the test names in the second commit :sweat_smile: 

Updated them in https://github.com/rust-lang/cargo/compare/dca7eb511c..5874b6f3474a

epage (approved):

epage:
Thanks for putting in all this work on this!

PR #15110, opened 2025-01-26
Update tests to fix nightly errors

ehuss:
There were some changes in the latest nightly which is breaking some tests:

* https://github.com/rust-lang/rust/pull/133154 updated the wording of a message. I adjusted the test to be less sensitive to the exact wording.
* https://github.com/rust-lang/rust/pull/119286 added the `linker-messages` lint which shows the output from the linker. Some of our tests were passing dummy flags to the linker, which it was complaining about. This updates it so that it uses real directories.
* This also fixes an incidental issue, where `build_script_needed_for_host_and_target` was not testing for the correct command-line flags. These got lost in https://github.com/rust-lang/cargo/pull/14132.

### DIFF
--- a/tests/testsuite/build_script.rs
+++ b/tests/testsuite/build_script.rs
@@ -878,20 +878,26 @@ fn custom_build_script_rustc_flags() {
             "foo/build.rs",
             r#"
                 fn main() {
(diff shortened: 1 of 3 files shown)


epage (approved):
Feel free to merge when CI is green

PR #15111, opened 2025-01-26
Fix shared_std_dependency_rebuild running on Windows

ehuss:
This fixes the `standard_lib::shared_std_dependency_rebuild` test while running on Windows. On my system, `CARGO_MANIFEST_DIR` is a normal windows-style path (`D:\rust\cargo`) with backslashes. That is not valid TOML syntax. I don't know why this doesn't fail on CI (maybe CI sets a unix-style current dir?).


weihanglo (approved):
This looks correct, though I don't think we are able to merge this before fixing other CI failures.

PR #15119, opened 2025-01-29
Move the changelog to the cargo book

ehuss:
This moves cargo's changelog to the cargo book. We're bumping into rendering limits on GitHub's side, so we're moving the changelog to the cargo book where it can be statically served under our control.

This PR is structured in such a way that the original CHANGELOG.md file is kept in place (and history is preserved in the new copy). The intent is to wait until this PR makes it to the nightly channel, and then open a new PR that will delete the original. That will look something like https://github.com/ehuss/cargo/commit/5ce79476939b5adff820abef6fb488765ede635b.

This makes use of a new mdbook feature in order to exclude the changelog chapter from the search index. It completely swamps the search results, so I think it is best to just exclude it.

We'll need to coordinate with the release team so that future release notes link directly to the cargo book instead of GitHub.


weihanglo (approved):
```
     Running `linkchecker/target/release/linkchecker linkcheck`
cargo/CHANGELOG.html:2848: broken intra-doc link - [<code>patch</code>]
cargo/print.html:22667: broken intra-doc link - [<code>patch</code>]
checked links in: 18.5s
number of HTML files scanned: 44199
number of HTML redirects found: 13800
number of links checked: 2423387
number of links ignored due to external: 92227
number of links ignored due to exceptions: 24
number of intra doc links ignored: 8
errors found: 2
found some broken links
```

Guess the CI failure was this:

```patch
diff --git a/src/doc/src/CHANGELOG.md b/src/doc/src/CHANGELOG.md
index 8ec33623364..c1ee4f243e7 100644
--- a/src/doc/src/CHANGELOG.md
+++ b/src/doc/src/CHANGELOG.md
@@ -2694,7 +2694,7 @@
   [#11790](https://github.com/rust-lang/cargo/pull/11790)
 - Broke endless loop on cyclic features in added dependency in `cargo add`.
   [#11805](https://github.com/rust-lang/cargo/pull/11805)
-- Don't panic when [`patch`] involved in dependency resolution results in a conflict.
+- Don't panic when `[patch]` involved in dependency resolution results in a conflict.
   [#11770](https://github.com/rust-lang/cargo/pull/11770)
 - Fixed credential token format validation.
   [#11951](https://github.com/rust-lang/cargo/pull/11951)
```

Feel free to merge when resolved.

epage:
> We'll need to coordinate with the release team so that future release notes link directly to the cargo book instead of GitHub.

Won't this break links from previous releases?

weihanglo:
> > We'll need to coordinate with the release team so that future release notes link directly to the cargo book instead of GitHub.
> 
> Won't this break links from previous releases?

Linking to the Cargo book won't.

> > The intent is to wait until this PR makes it to the nightly channel, and then open a new PR that will delete the original. That will look something like https://github.com/ehuss/cargo/commit/5ce79476939b5adff820abef6fb488765ede635b.

This will.

---

Maybe instead, we put a big notice on the top of the file, and we keep only the heading of each release, and put a link under it to the new location of the cargo book.

epage:
Other options
- Could the book symlink to the root's changelog?  Limits book building to Windows
- We fork the changelog and freeze the current location

No strong preference on what we do.

PR #15123, opened 2025-01-29
Remove the original changelog

ehuss:
This removes the original changelog now that it is located in the cargo book.

This places a notice in the original location that it has moved to a new place. There are a few other alternatives we could consider:
* Keep the headings but replace the content with a link. This would make it a little easier for someone to jump to the exact position in the new page. However, it looks a little messy. Each section could be something like:

   > This content has moved to [CHANGELOG.md#186-2025-04-03](https://doc.rust-lang.org/cargo/nightly/CHANGELOG.md#186-2025-04-03)

* Keep the content, with a notice at the top that the page is no longer being updated. I was a little uncomfortable with this, since I wasn't sure people would see the notice, it duplicates some content, and still leaves the potential problems of GitHub not rendering the page. If we cut some content, it ensures the page still renders reliably.
  * As a variant, we could delete the content but keep the last 1 or 2 releases since those are probably what people will be looking for.

Opening as draft until https://github.com/rust-lang/cargo/pull/15119 is published on nightly.


weihanglo:
The other alternative is we patching every link in blog.rust-lang.org. Admittedly this won't fix other links in the wild.

ehuss:
This should be ready to go now that the docs are up at https://doc.rust-lang.org/nightly/cargo/CHANGELOG.html.

> The other alternative is we patching every link in blog.rust-lang.org. Admittedly this won't fix other links in the wild.

I'm happy to post a PR to fix those links, too.

weihanglo:
> This should be ready to go now that the docs are up at [doc.rust-lang.org/nightly/cargo/CHANGELOG.html](https://doc.rust-lang.org/nightly/cargo/CHANGELOG.html).
> 
> > The other alternative is we patching every link in blog.rust-lang.org. Admittedly this won't fix other links in the wild.
> 
> I'm happy to post a PR to fix those links, too.

Should that be a blocker of merging this? I am fine with either way.


cc @epage, wonder what your thought is, as you were in the previous discussion.

epage:
That works for those links but not for any others (bookmarks, browser history, links from issues, etc).  If they land on a generic 404 page, that isn't great.   It is probably fine for them to have to search within a page.  Releases are easy to find.  We didn't have anchors to specific items, so work for that was already needed.

I also find that the first place I look for a changelog is by looking for a file in a repo root.  I rarely think to check Releases.  I'll likely miss it in the body of the text here.

Options
- Symlink the book to the root or symlink the root to the book
- Make the changelog only contain a link to the web page
- Fork and freeze the changelog with a top-level item pointing to the web page

ehuss:
> If they land on a generic 404 page

The page isn't being removed. There is a placeholder that directs them to the new location.

> Make the changelog only contain a link to the web page

This is the option I'm proposing here.

The reason for not symlinking is that I would prefer to avoid having users assume they can look at that file on GitHub, since there is risk it won't render at all. I'd also like to make it easier to do relative links within the book (which can help with offline viewing, link validation, etc.).

The reason I refrained from the fork-and-freeze approach is because I wanted to avoid having two copies of a large amount of content. It may be hard for the user to see a notice at the top of the page that it is no longer being updated if there is still a large amount of content displayed to the user. It also doesn't help with the "possibly doesn't render" problem.

epage:
> The page isn't being removed. There is a placeholder that directs them to the new location.

Sorry, I looked too closely and thought it was deleted.

That should work.

weihanglo (approved):
Looks like we are okay with the caveat that some links in the wild may be invalid, so I am going to merge it. Thanks for fixing the issue!

PR #15124, opened 2025-01-29
Don't suggest `cargo login` when using incompatible credental providers

kornelski:
Authentication errors shouldn't unconditionally recommend running `cargo login`, because alternative registries configured to use custom credential providers won't use the token obtained by `cargo login`.

### DIFF
--- a/tests/testsuite/credential_process.rs
+++ b/tests/testsuite/credential_process.rs
@@ -101,7 +101,8 @@ fn credential_provider_auth_failure() {
         .with_status(101)
         .with_stderr_data(str![[r#"
 [UPDATING] `alternative` index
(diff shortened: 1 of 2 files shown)


tests/testsuite/credential_process.rs
```diff
@@ -87,6 +87,31 @@ You may press ctrl-c [..]
...
+[ERROR] token rejected for `alternative`
+You may need to log in using this registry's credential provider
+
+Caused by:
+  failed to get successful HTTP response from [..]
+  body:
+  [..]
+"#]])
+        .run();
+}
```
epage, line 113: Could you add the test in a prior commit, with the test passing?  Then this commit will show how your change affected behavior.

weihanglo (approved):

PR #11600, opened 2023-01-19
Error on invalid alphanumeric token for crates.io

Akida31:
ref #11571 

When using `cargo login` and calling an api which requires authentification there will be an error if the given token is not a valid alphanumerical string.
This check is only enabled for crates.io because
only for that registry we can be certain, that the generated token should have been alphanumeric, see [the code here](https://github.com/rust-lang/crates.io/blob/7ea41e9d345f05566ee776b7cbb62e46ccf6b393/src/util/token.rs#L15). So if I'm not mistaken, this should not be a breaking change, since crates.io only generates fitting tokens. (Should I add a comment to the crates.io code that modifying this logic can break cargo?)

I'm not sure if the fix works and is enough to close the issue, please say if you have any corrections or improvements!

I don't know if the check should also be enabled for other registries and it would be really bad if the check is too strict.
In the linked issue it was recommended to encode invalid characters, but I don't know in which encoding. I saw in [this http rfc](https://www.rfc-editor.org/rfc/rfc7230#section-3.2.4) that only the ISO-8859-1 charset is allowed and everything else must be [encoded](https://www.rfc-editor.org/rfc/rfc7230#section-3.2.4) but this seems somewhat complex and hard to implement. There is a crate `rust-encoding` which should be capable doing this (from a first look), but I don't know if a new dependency only for this is justified. There seems to be `percent encoding` already in the dependency tree but I have no idea if it would be correct and work.
If you have any idea about this encoding, please say so.

r? @Eh2406 (since you suggested the encoding part)

PR #15127, opened 2025-01-31
Check token validity when loading registry config

kornelski:
This change checks validity of registry tokens earlier, when they're read from the config. Previously Cargo would check the token only when sending a request.

This makes the error message more helpful, because it can tell how the token has been defined. This is especially helpful when users do something like:

```
export CARGO_REGISTRIES_FOO_TOKEN=$(misconfigured tool printing error messages)
```

### DIFF
--- a/tests/testsuite/credential_process.rs
+++ b/tests/testsuite/credential_process.rs
@@ -95,7 +95,7 @@ fn credential_provider_auth_failure() {
         .auth_required()
         .alternative()
         .no_configure_token()
-        .credential_provider(&["cargo:token-from-stdout", "true"])
+        .credential_provider(&["cargo:token-from-stdout", "echo", "incorrect token"])
         .build();
 
     cargo_process("install libc --registry=alternative")
--- a/tests/testsuite/registry_auth.rs
+++ b/tests/testsuite/registry_auth.rs
@@ -391,6 +391,34 @@ Caused by:
 }
 
 #[cargo_test]
+fn syntactically_invalid_token() {
+    let _registry = RegistryBuilder::new()
+        .alternative()
+        .auth_required()
+        .no_configure_token()
+        .http_index()
+        .build();
+
+    let p = make_project();
+    cargo(&p, "build")
+        .env("CARGO_REGISTRIES_ALTERNATIVE_TOKEN", "\t\n悪")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[UPDATING] `alternative` index
+[ERROR] failed to get `bar` as a dependency of package `foo v0.0.1 ([ROOT]/foo)`
+
+Caused by:
+  Token for registry `alternative` is invalid (defined in environment variable `CARGO_REGISTRIES_ALTERNATIVE_TOKEN`)
+
+Caused by:
+  token contains invalid characters.
+  Only printable ASCII characters are allowed as it is sent in a HTTPS header.
+
+"#]])
+        .run();
+}
+
+#[cargo_test]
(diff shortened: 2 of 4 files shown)


crates/crates-io/lib.rs
```diff
@@ -545,7 +545,7 @@ pub fn check_token(token: &str) -> Result<()> {
         Ok(())
     } else {
         Err(Error::InvalidToken(
-            "token contains invalid characters.\nOnly printable ISO-8859-1 characters \
+            "token contains invalid characters.\nOnly printable ASCII characters \
```
epage, line 548: Looking at #11600, it seems that what was checked evolved but the error message was never updated for the subset encoding that more people will also recognize.

So this seems reasonable

src/cargo/util/auth/mod.rs
```diff
@@ -236,6 +237,17 @@ pub fn registry_credential_config_raw(
...
+        token: Some(token), ..
+    }) = &cfg
+    {
+        check_token(&token.val.as_deref().expose()).with_context(|| {
+            format!(
+                "Token for {sid} is invalid (defined in {})",
+                token.definition
+            )
+        })?;
+    }
```
epage, line 250: This is moving the check from
- token credential provider

To
- token credential provider
- paseto credential provider
- the enumeration of all credential providers

I'd want to better understand if its a good idea to be running this check in all of those other situations

CC @arlosi

kornelski: Construction of `RegistryConfig` is closer to TOML deserialization of `credentials.toml` than to implementation of credential providers. The token here is the *input* for downstream users of the registry config, and does not touch *output* token of credential providers (that's `CredentialResponse`).

Paseto uses `secret_key` and `secret_key_subject` fields, not `token`.

This field is only read by `TokenCredential`, apart from code that warns when `cargo::token` is not enabled.

kornelski:
I've found another case where extra `\n` in the token causes an internal error: 

```text
Caused by:
  download of config.json failed

Caused by:
  failed to download from `[…]/index/config.json`

Caused by:
  [43] A libcurl function was given a bad argument
```

Fixed in 9ef2dd10a66f96442090cc6c52e9716fd79c1dfd

src/cargo/util/auth/mod.rs
epage: @arlo I'd be interested in your thoughts on this

tests/testsuite/credential_process.rs
```diff
@@ -95,7 +95,7 @@ fn credential_provider_auth_failure() {
         .auth_required()
         .alternative()
         .no_configure_token()
-        .credential_provider(&["cargo:token-from-stdout", "true"])
+        .credential_provider(&["cargo:token-from-stdout", "echo", "incorrect token"])
```
epage, line 98: Why is this test being changed?

tests/testsuite/registry_auth.rs
```diff
@@ -390,6 +390,34 @@ Caused by:
         .run();
 }
 
+#[cargo_test]
+fn syntactically_invalid_token() {
```
epage, line 394: As a reminder, we generally prefer tests to be added in their own commit, showing the current behavior.

tests/testsuite/credential_process.rs
kornelski: This makes the test more precise, because it requires the token to be accepted syntactically, and rejected only by the server. `true` creates an empty string, which effectively skips per-byte checks of the token's content.

PR #15133, opened 2025-02-03
Suggest similar feature names on CLI

DJMcNab:

When you typo a feature name on the CLI, the error message isn't very helpful. Concretely, I was testing a PR which adds a feature called `cosmic_text` to enable a `cosmic-text` dependency, and got a correct but unhelpful error message:
```rust
error: Package `scenes v0.0.0 ([ELIDED]/linebender/vello/examples/scenes)` does not have feature `cosmic-text`. It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.
```

I had to dig into the Cargo.lock file to find out how to fix this.

Observe the new test cases

### DIFF
--- a/src/cargo/core/resolver/dep_cache.rs
+++ b/src/cargo/core/resolver/dep_cache.rs
@@ -20,11 +20,13 @@ use crate::core::{
     Dependency, FeatureValue, PackageId, PackageIdSpec, PackageIdSpecQuery, Registry, Summary,
 };
 use crate::sources::source::QueryKind;
+use crate::util::closest_msg;
 use crate::util::errors::CargoResult;
 use crate::util::interning::{InternedString, INTERNED_DEFAULT};
 
 use anyhow::Context as _;
 use std::collections::{BTreeSet, HashMap, HashSet};
+use std::fmt::Write;
 use std::rc::Rc;
 use std::task::Poll;
 use tracing::debug;
@@ -514,11 +516,20 @@ impl RequirementError {
                     .collect();
                 if deps.is_empty() {
                     return match parent {
-                        None => ActivateError::Fatal(anyhow::format_err!(
-                            "Package `{}` does not have the feature `{}`",
-                            summary.package_id(),
-                            feat
-                        )),
+                        None => {
+                            let closest = closest_msg(
+                                &feat.as_str(),
+                                summary.features().keys(),
+                                |key| &key,
+                                "feature",
+                            );
+                            ActivateError::Fatal(anyhow::format_err!(
+                                "Package `{}` does not have the feature `{}`{}",
+                                summary.package_id(),
+                                feat,
+                                closest
+                            ))
+                        }
                         Some(p) => {
                             ActivateError::Conflict(p, ConflictReason::MissingFeatures(feat))
                         }
@@ -526,13 +537,32 @@ impl RequirementError {
                 }
                 if deps.iter().any(|dep| dep.is_optional()) {
                     match parent {
-                        None => ActivateError::Fatal(anyhow::format_err!(
-                            "Package `{}` does not have feature `{}`. It has an optional dependency \
-                             with that name, but that dependency uses the \"dep:\" \
-                             syntax in the features table, so it does not have an implicit feature with that name.",
-                            summary.package_id(),
-                            feat
-                        )),
+                        None => {
+                            let mut features =
+                                features_enabling_dependency_sorted(summary, feat).peekable();
+                            let mut suggestion = String::new();
+                            if features.peek().is_some() {
+                                suggestion = format!(
+                                    "\nDependency `{}` would be enabled by these features:",
+                                    feat
+                                );
+                                for feature in (&mut features).take(3) {
+                                    let _ = write!(&mut suggestion, "\n\t- `{}`", feature);
+                                }
+                                if features.peek().is_some() {
+                                    suggestion.push_str("\n\t  ...");
+                                }
+                            }
+                            ActivateError::Fatal(anyhow::format_err!(
+                                "\
+Package `{}` does not have feature `{}`. It has an optional dependency \
+with that name, but that dependency uses the \"dep:\" \
+syntax in the features table, so it does not have an implicit feature with that name.{}",
+                                summary.package_id(),
+                                feat,
+                                suggestion
+                            ))
+                        }
                         Some(p) => ActivateError::Conflict(
                             p,
                             ConflictReason::NonImplicitDependencyAsFeature(feat),
@@ -544,7 +574,7 @@ impl RequirementError {
                             "Package `{}` does not have feature `{}`. It has a required dependency \
                              with that name, but only optional dependencies can be used as features.",
                             summary.package_id(),
-                            feat
+                            feat,
                         )),
                         Some(p) => ActivateError::Conflict(
                             p,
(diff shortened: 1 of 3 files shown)


src/cargo/core/resolver/dep_cache.rs
```diff
@@ -529,9 +532,10 @@ impl RequirementError {
                         None => ActivateError::Fatal(anyhow::format_err!(
                             "Package `{}` does not have feature `{}`. It has an optional dependency \
                              with that name, but that dependency uses the \"dep:\" \
-                             syntax in the features table, so it does not have an implicit feature with that name.",
+                             syntax in the features table, so it does not have an implicit feature with that name.{}",
                             summary.package_id(),
-                            feat
+                            feat,
+                            closest
                         )),
```
epage, line 539: This error message is already telling the user that something exists that could be mistaken for a feature but wasn't.  It is unlikely that a typo would happen here and it feels like the two messages don't compose well.

src/cargo/core/resolver/dep_cache.rs
```diff
@@ -542,9 +546,10 @@ impl RequirementError {
                     match parent {
                         None => ActivateError::Fatal(anyhow::format_err!(
                             "Package `{}` does not have feature `{}`. It has a required dependency \
-                             with that name, but only optional dependencies can be used as features.",
+                             with that name, but only optional dependencies can be used as features.{}",
                             summary.package_id(),
-                            feat
+                            feat,
+                            closest
                         )),
```
epage, line 553: This error message is already telling the user that something exists that could be mistaken for a feature but wasn't.  It is unlikely that a typo would happen here and it feels like the two messages don't compose well.

epage:
Thanks for adding this!

src/cargo/core/resolver/dep_cache.rs
DJMcNab: We could search through all the features which exist, and find those which enable this dependency?

Fwiw, this error message would have helped with the reason I opened this PR, but that is a very niche case.

DJMcNab: Yes, I did. This is entirely right, and I will remove the suggestion from this branch.

(I had pattern-matched this to the thing which sometimes happens on GitHub of comments double-sending)

src/cargo/core/resolver/dep_cache.rs
DJMcNab: Hopefully 875e1cfb7a151e38a41c113441cd1be41f303b6f makes this better

DJMcNab:
Does rustbot know something I don't? If it insists on that label, I'm not going to fight it.

arlosi:
> Does rustbot know something I don't? If it insists on that label, I'm not going to fight it.

It's because you're editing files within the resolver. Don't worry about it.

src/cargo/core/resolver/dep_cache.rs
```diff
@@ -513,25 +515,60 @@ impl RequirementError {
...
+                                }
+                                _ => (),
+                            }
+                        }
+                    }
+                    let mut suggestion = String::new();
+                    if !features.is_empty() {
+                        // Features is already sorted because it was constructed from a BTreeMap.
+                        suggestion =
+                            String::from("\nThis feature would be enabled by these features:");
```
epage, line 555:
```suggestion
                            String::from("\nDependency `{dep_name}` is enabled by these features:");
```

src/cargo/core/resolver/dep_cache.rs
```diff
@@ -513,25 +515,60 @@ impl RequirementError {
...
+                        let mut features = features.iter();
+                        for feature in (&mut features).take(3) {
+                            let _ = write!(&mut suggestion, "\n\t- `{}`", feature);
+                        }
+                        if features.next().is_some() {
+                            suggestion.push_str("\n\t  ...");
+                        }
+                    }
```
epage, line 563: 1. Is there a reason this isn't being done in the only relevant branch?
2. Should we move some or all of this out to a helper function to not detract from the flow of this code?a

epage:
Can you restructure your PR
- A commit that extends the tests to cover more cases (should pass)
- A commit for behavior changes (tests should be updated to pass)

src/cargo/core/resolver/dep_cache.rs
```diff
@@ -513,25 +515,60 @@ impl RequirementError {
                     .filter(|dep| dep.name_in_toml() == feat)
                     .collect();
                 if deps.is_empty() {
+                    let closest =
+                        closest_msg(&feat.as_str(), summary.features().keys(), |key| &key);
```
Eh2406, line 519: How expensive is the `closest_msg` call? and same for the coding added below?

This part of the code can be hit very often in a build. So operations like an allocation should be done only where absolutely necessary, if possible only in the branch where `ActivateError::Fatal` is returned. Happy to eat the performance hit if it's necessary for better error messages. But, if we can avoid it I would like to.

epage: Is there any happy-path performance hit if this is limited to `Fatal` branches?

src/cargo/core/resolver/dep_cache.rs
DJMcNab: When I've applied this (not yet pushed), I've used the "old style" `format_args` argument, as that is the prevailing style.

src/cargo/core/resolver/dep_cache.rs
DJMcNab: The changes are now only limited to the `Fatal` branches, I believe

src/cargo/core/resolver/dep_cache.rs
```diff
@@ -514,25 +516,51 @@ impl RequirementError {
...
+                                for feature in (&mut features).take(3) {
+                                    let _ = write!(&mut suggestion, "\n\t- `{}`", feature);
+                                }
+                                if features.peek().is_some() {
+                                    suggestion.push_str("\n\t  ...");
+                                }
+                            }
+                            ActivateError::Fatal(anyhow::format_err!(
+                                // We left-align this error message here so that rustfmt won't refuse to
+                                // format the remaining parts of this file.
```
epage, line 554: I can see making this a comment on the PR but leaving this as an artifact in the code adds noise imo

DJMcNab: Hmm, I would think one would want to know why this formatting is different than elsewhere as a reader of this code.

I can remove if this is blocking, but I don't intend to otherwise, barring some other solution.

src/cargo/core/resolver/dep_cache.rs
Eh2406: Thank you so much for the change.

src/cargo/core/resolver/dep_cache.rs
epage: Speaking for myself in reviewing this, please remove this.

We are not too consistent on our multi-line string formatting generally, a deviation like this is small, this seems like a bug in rustfmt and so would be transient in nature without this having anything tracking it, and has a big comment, raising its visibility more than its relative importance.

epage:
Could you clean up the commits for how you'd want them merged?

epage:
FYI you'll need to rebase on top of #15138 (this PR inspired it but I didn't expect it to get in first)

epage (approved):
Thanks!  This is a great improvement to our error reporting!  Thank you for your patience through the code review process (and the semantic merge conflicts)

Dorian-Rozan (approved):

PR #10322, opened 2022-01-24
Test if reserved filenames are allowed in Windows

eholk:
Recent versions of Windows have removed the limitation on filenames like `aux` or `con`. This change allows the `package::reserved_windows_name` to still pass by first trying to create a file with a reserved name to see if Windows supports it. If so, it skips the rest of the test. Otherwise, we keep the same behavior as before.

PR #15135, opened 2025-02-03
allow windows reserved names in CI

marcoieni:
We are progressively moving the windows CI from windows 2022 to windows 2025 because we found windows 2025 more stable.

In https://github.com/rust-lang/rust/pull/136478 a cargo test failed and this might solve the issue. See [this](https://github.com/rust-lang/rust/pull/136478#issuecomment-2631307932) comment. What do you think?

If you have a better way of solving this, let me know 👍

### DIFF
--- a/crates/cargo-test-support/src/paths.rs
+++ b/crates/cargo-test-support/src/paths.rs
@@ -373,11 +373,6 @@ pub fn sysroot() -> String {
 pub fn windows_reserved_names_are_allowed() -> bool {
     use cargo_util::is_ci;
 
-    // Ensure tests still run in CI until we need to migrate.
-    if is_ci() {
-        return false;
-    }
-
     use std::ffi::OsStr;
     use std::os::windows::ffi::OsStrExt;
     use std::ptr;


epage:
That check was added at the request of @ehuss at https://github.com/rust-lang/cargo/pull/10322#discussion_r932825120

I feel like the comment in the code is ambiguous but in looking at that context, it seems like upgrading CI to a new Windows will cause us to lose coverage and this test was to make sure we got as much coverage as possible until that day.  I don't see much we can do about ensuring old coverage except running two versions of Windows in Ci which doesn't seem ideal, so I'm assuming this should be fine but will wait a bit in case anyone else with more context can pipe in.

arlosi:
This change makes sense to me. 

Eventually, once more users are on a Windows release that doesn't have this restriction, we can remove the test altogether.

ehuss (approved):
This seems fine to me. My original intent was that once we drop Windows 10 as a tier-1 platform, we would migrate our CI to Windows 11 (and remove this test). However, it looks like CI migration is happening faster than our tier support migration. Probably not too much risk, though we could maybe consider running both windows-2022 and windows-2025 (at least on the cargo side), but probably not worth the bother.

ChrisDenton:
I would just note that even the Windows 10 runner is a fair way above Rust's minimum tier 1 support (the runner is Windows Server 2019 whereas our minimum tier 1 supported version is Windows Server 2016 or Windows Desktop 1507). And that's not just a technicality as major "feature" releases were made on a yearly (or twice yearly) release schedule during Windows 10's lifetime. Ideally we would also test RTM Windows 10 but alas we don't have the infra.

PR #15138, opened 2025-02-03
fix: Don't use "did you mean" in errors

epage:

This is an attempt at moving Cargo closer to [rustc's diagnostic guide](https://rustc-dev-guide.rust-lang.org/diagnostics.html#suggestion-style-guide) (clap already did this)

> Suggestions should not be a question. In particular, language like "did you mean" should be avoided. Sometimes, it's unclear why a particular suggestion is being made. In these cases, it's better to be upfront about what the suggestion is.
>
> Compare "did you mean: Foo" vs. "there is a struct with a similar name: Foo".

### Additional information

I did leave behind three "did you mean"s as I was a little less sure in how to handle them


weihanglo (approved):
Thanks!

Issue #9528, opened 2021-05-31
Cargo gets lost in the root of a Linux filesystem


ehuss:
**Problem**
`cargo doc` will mysteriously hang when run from the root of a Linux filesystem. The problem is that the fingerprinting code tries to find the file with the newest mtime by walking the entire filesystem. Unfortunately there are some paths that are infinite black holes. 

An example is `/sys/block`. This tree has a complex set of cyclical symbolic links. Cargo follows a cycle until it reaches ELOOP (too many symbolic links), and then starts following another cycle. There is a permutation of cycles, and so cargo essentially gets lost forever.

I believe this also affects "old" style build scripts which also use `LocalFingerprint::Precalculated`.

**Steps**
1. Start a bash prompt with the `rust:latest` docker image.
2. In the root directory, run `cargo init --name foo`
3. Run `cargo doc`. This step will never finish.

**Possible Solution(s)**
Uncertain.  A few ideas:
* Use `walkdir` for walking, which has symbolic link cycle detection.
* Check for special devices and ignore them.
* Detect running in root, and return an error.
* Set an upper bound on the number of files the walker will visit before it gives up.

As a side note, another problem with the implementation is that it collects the path of every file it visits in memory. This is done with the assumption that `PathSource::list_files` will return a reasonable number of files. However, the mtime computation code doesn't need a full list, but instead just needs to know which file has the newest mtime. Perhaps that code could be changed to take a callback instead of accumulating all the files in memory.

**Notes**

Output of `cargo version`:
cargo 1.54.0-nightly (e931e4796 2021-05-24)

Linux 4.19.121

jyn514:
@ehuss it seems strange to me that cargo looks at all files in the first place - I can look into adding --emit=dep-info to rustdoc if that would help.

weihanglo:
> Perhaps that code could be changed to take a callback instead of accumulating all the files in memory.

I prototyped [an iterative version], but I don't feel comfortable about its complexity. 😞
Is there anything can do to make it simpler?

[an iterative version]: https://github.com/rust-lang/cargo/compare/master...weihanglo:refactor-walk

LizardWizzard:
Faced the same issue during cargo build when I copy sources into root in docker. I see endless wall of warnings with the original repro. 

Workaround for me was to create directory and copy sources there instead of root.

Will it be better to terminate with an error instead of being stuck with warnings?

PR #15141, opened 2025-02-04
fix: detect infinite file system loop and bail out

### LINKED ISSUE #12600: `warning: File system loop found: xxx` printed when running `cargo run` on file not ever needed by `cargo`
### Problem

`cargo run` prints a `warning: File system loop found: xxx` warning message on every build, for a symlink that is not ever needed by Cargo.

This file is in a `fixture/` directory. The fixture exists specifically to test the code with this file system loop case.

### Steps

1. Clone https://github.com/denoland/deno.git
2. Install `protoc` and `cmake`
3. `cargo run`

Observe:

```
warning: File system loop found: /path/to/deno/test_util/std/wasi/testdata/fixtures/symlink_to_parent_directory/fixtures/symlink_to_parent_directory points to an ancestor /path/to/deno/test_util/std/wasi/testdata/fixtures/symlink_to_parent_directory
```

### Possible Solution(s)

_No response_

### Notes

This did not occur in 1.71.0.

### Version

```text
release: 1.72.0
commit-date: 2023-08-15
host: aarch64-apple-darwin
libcurl: 7.88.1 (sys:0.4.63+curl-8.1.2 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 13.4.1 [64-bit]
```

weihanglo:

For #9528 if the loop in inside `/sys/block` it would be an infinite loop.
In that case Cargo is better bailing out instead of emitting endless warnings.
We set a relatively large limit 1024
to ensure there is any legitimate reason to have file system loop.

This also turn the file system loop warning into a `debug!` log, per #12600 request.

I don't really know if 1024 is really a big enough number.
My naive instinct told me that there is no a valid use case for so many loops in a Cargo package.

### Additional information

Fixes #12600

### DIFF
--- a/src/cargo/sources/path.rs
+++ b/src/cargo/sources/path.rs
@@ -95,7 +95,7 @@ impl<'gctx> PathSource<'gctx> {
     /// `package.exclude` to filter the list of files.
     #[tracing::instrument(skip_all)]
     pub fn list_files(&self, pkg: &Package) -> CargoResult<Vec<PathEntry>> {
-        list_files(pkg, self.gctx)
+        list_files(pkg)
     }
 
     /// Gets the last modified file in a package.
@@ -106,7 +106,7 @@ impl<'gctx> PathSource<'gctx> {
                 self.path
             )));
         }
-        last_modified_file(&self.path, pkg, self.gctx)
+        last_modified_file(&self.path, pkg)
     }
 
     /// Returns the root path of this source.
@@ -279,7 +279,7 @@ impl<'gctx> RecursivePathSource<'gctx> {
     /// use other methods like `.gitignore`, `package.include`, or
     /// `package.exclude` to filter the list of files.
     pub fn list_files(&self, pkg: &Package) -> CargoResult<Vec<PathEntry>> {
-        list_files(pkg, self.gctx)
+        list_files(pkg)
     }
 
     /// Gets the last modified file in a package.
@@ -290,7 +290,7 @@ impl<'gctx> RecursivePathSource<'gctx> {
                 self.path
             )));
         }
-        last_modified_file(&self.path, pkg, self.gctx)
+        last_modified_file(&self.path, pkg)
     }
 
     /// Returns the root path of this source.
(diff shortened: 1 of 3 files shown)


epage (changes requested):
For myself, I'd like to better understand more about the use case before moving forward.

weihanglo:
> For myself, I'd like to better understand more about the use case before moving forward.

While this is a fix of #12600, it is also a solution to <https://github.com/rust-lang/cargo/issues/9528#issuecomment-2090154025>. Fixing #12600 is more like a side effect.

The warning of file system loop was a solution to let people know they are stuck in infinite loop (#9528). However endless warnings are not particularly useful. Especially if it happens in CI and then users need to wait for CI service timing out, which could be hours. Breaking the loop is more pragmatic than that I believe.

ijackson:
By "bail out" I hope we mean "skip the recursion and continue scanning other files" rather than "fai".

Currently I'm getting these warnings but I just want them suppressed, not to become fatal.  See #12600

weihanglo:
@ijackson 

> We set a relatively large limit 1024 
> to ensure there is any legitimate reason to have file system loop.

Unsure whether 1024 is large enough to say that you have an “infinite file system loop”. Can increase it if you have a use case that loops back to the same directory more than 1024 times.

ijackson:
> @ijackson
> 
> > We set a relatively large limit 1024
> > to ensure there is any legitimate reason to have file system loop.
> 
> Unsure whether 1024 is large enough to say that you have an “infinite file system loop”. Can increase it if you have a use case that loops back to the same directory more than 1024 times.

I'm not sure I follow.  In one of my use cases I have, for example, a symlink `maint -> `.  Every time you try to follow it, it will come back to the same directory.

Currently I'm getting a warning about this, but cargo seems to function.  If this warning becomes an error my project will become unbuildable.

The value of 1024 doesn't seem relevant.

epage:
From my reading of the code and walkdir, the 1024 is more of a "if you find 1024 unique symlinks pointing to an ancestor, bail".  Putting a cap on that seems like an odd solution.

That was the solution proposed in https://github.com/rust-lang/cargo/issues/9528#issuecomment-2090154025 but that comment is overall light on details and solution seems more focused on "error rather than make my CI slow" which seems like a false dichotomy and that we should explore other solutions.  As that person didn't create an issue, should we continue that conversation over at #12600 (as our PRs are generally more focused on the implementation than problem/solution discussions)?

weihanglo:
> From my reading of the code and walkdir, the 1024 is more of a "if you find 1024 unique symlinks pointing to an ancestor, bail". Putting a cap on that seems like an odd solution.

Agree that it is just a random cap. <https://github.com/rust-lang/cargo/issues/12235> is probably one solution to it. Let's close this PR and discuss in #12600 then.

PR #15143, opened 2025-02-04
Don't use on Solaris libc::LOCK_* which were removed from libc in ver…

psumbera:
Relevant libc change was: https://github.com/rust-lang/libc/commit/251e8e839a22e3d3140091f780e5123d9e8aac15

### DIFF
--- a/src/cargo/util/flock.rs
+++ b/src/cargo/util/flock.rs
@@ -436,24 +436,41 @@ mod sys {
     use std::io::{Error, Result};
     use std::os::unix::io::AsRawFd;
 
+    #[cfg(not(target_os = "solaris"))]
+    const LOCK_SH: i32 = libc::LOCK_SH;
+    #[cfg(target_os = "solaris")]
+    const LOCK_SH: i32 = 1;
+    #[cfg(not(target_os = "solaris"))]
+    const LOCK_EX: i32 = libc::LOCK_EX;
+    #[cfg(target_os = "solaris")]
+    const LOCK_EX: i32 = 2;
+    #[cfg(not(target_os = "solaris"))]
+    const LOCK_NB: i32 = libc::LOCK_NB;
+    #[cfg(target_os = "solaris")]
+    const LOCK_NB: i32 = 4;
+    #[cfg(not(target_os = "solaris"))]
+    const LOCK_UN: i32 = libc::LOCK_UN;
+    #[cfg(target_os = "solaris")]
+    const LOCK_UN: i32 = 8;
+
     pub(super) fn lock_shared(file: &File) -> Result<()> {
-        flock(file, libc::LOCK_SH)
+        flock(file, LOCK_SH)
     }
 
     pub(super) fn lock_exclusive(file: &File) -> Result<()> {
-        flock(file, libc::LOCK_EX)
+        flock(file, LOCK_EX)
     }
 
     pub(super) fn try_lock_shared(file: &File) -> Result<()> {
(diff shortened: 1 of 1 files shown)


epage:
Wouldn't that be a breaking change?  I'm not seeing that discussed in rust-lang/libc#4017

src/cargo/util/flock.rs
```diff
@@ -437,23 +437,58 @@ mod sys {
     use std::os::unix::io::AsRawFd;
 
     pub(super) fn lock_shared(file: &File) -> Result<()> {
-        flock(file, libc::LOCK_SH)
+        #[cfg(not(target_os = "solaris"))]
+        {
+            flock(file, libc::LOCK_SH)
+        }
```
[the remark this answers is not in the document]
psumbera: Without `{ }`,  Rust complains like this:
```
error: expected `;`, found `#`
   --> /export/home/test/cargo-solaris-libc/src/cargo/util/flock.rs:441:35
    |
440 |         #[cfg(not(target_os = "solaris"))]
    |         ---------------------------------- only `;` terminated statements or tail expressions are allowed after this attribute
441 |         flock(file, libc::LOCK_SH)
    |                                   ^ expected `;` here
442 |         #[cfg(target_os = "solaris")]
    |         - unexpected token
    |
```

psumbera:
> Wouldn't that be a breaking change? I'm not seeing that discussed in [rust-lang/libc#4017](https://github.com/rust-lang/libc/pull/4017)

Solaris never had `flock()` and corresponding `LOCK_`* definition. Only Illumos (Solaris fork) added this. And since Rust libc definitions are partially shared between these two it was there defined for both. While I was fixing Rust libc tests for Solaris I had to remove it from there.

src/cargo/util/flock.rs
```diff
@@ -437,23 +437,58 @@ mod sys {
...
 
     pub(super) fn try_lock_exclusive(file: &File) -> Result<()> {
-        flock(file, libc::LOCK_EX | libc::LOCK_NB)
+        #[cfg(not(target_os = "solaris"))]
+        {
+            flock(file, libc::LOCK_EX | libc::LOCK_NB)
+        }
+        #[cfg(target_os = "solaris")]
+        {
+            flock(file, 2 | 4)
```
epage, line 479: Should we make our own constants so we limit all solaris-specific code to just the solaris implementation of `flock`

psumbera: I'm open for any suggestion. But not sure what do you exactly mean here...

weihanglo: Guess something like?

```rust
#[cfg(all(unix, not(target_os = "solaris")))]
const LOCK_UN = libc::LOCK_UN;
#[cfg(all(unix, target_os = "solaris"))]
const LOCK_UN = 8;
```

weihanglo (approved):
I am good with this. Thanks!

src/cargo/util/flock.rs
```diff
@@ -430,30 +430,48 @@ fn is_on_nfs_mount(_path: &Path) -> bool {
     false
 }
 
+#[cfg(all(unix, not(target_os = "solaris")))]
```
weihanglo, line 433: Sorry for being picky. Feel like this can move one level down into the `#[cfg(unix)] mod sys` so we can remove `all(unix,` part from cfg?

weihanglo (approved):
Thanks!

Issue #15145, opened 2025-02-05
macOS: `cargo clean gc` doesn't understand `.DS_Store`


madsmtm:
### Problem

Trying to remove a directory that has been opened in macOS' Finder fails with the following error:

```console
$ cargo +nightly clean gc -Zgc
error: failed to clean entries from the global cache

Caused by:
  failed to sync tracking database

Caused by:
  failed to read path `"~/.cargo/registry/cache/.DS_Store"`

Caused by:
  Not a directory (os error 20)
```

Same goes for `~/.cargo/registry/src`.

The problem is resolved by manually removing the `.DS_Store`.

### Steps

1. Run `open ~/.cargo/registry/cache` to open the cache directory in Finder.
2. Fiddle around with the subfolders in the Finder (open/close if in list view, move around if in symbol view).
3. Run `cargo clean gc -Zgc`.

### Possible Solution(s)

Perhaps Cargo should ignore the file in it's consistency check?

### Notes

Possibly related: https://github.com/rust-lang/cargo/issues/11441.

### Version

```text
cargo 1.86.0-nightly (0e3d73849 2025-02-01)
release: 1.86.0-nightly
commit-hash: 0e3d73849ab8cbbab3ec5c65cbd555586cb21339
commit-date: 2025-02-01
host: aarch64-apple-darwin
libgit2: 1.9.0 (sys:0.20.0 vendored)
libcurl: 8.7.1 (sys:0.4.74+curl-8.9.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Mac OS 14.7.1 [64-bit]
```

arlosi:
I think Cargo is expecting only directories in that location, not files, however I wasn't able to reproduce this issue on Windows by creating an empty `.DS_Store` file manually.

Could you please try running again with the environment variable `CARGO_LOG=gc`? That should print additional logging information.

ehuss:
I'm able to reproduce. I'll look at making it more resilient.

PR #15147, opened 2025-02-05
Make cache tracking resilient to unexpected files

ehuss:
This adds a check to fix an issue where the cache tracking synchronization was breaking on unexpected files like macOS's `.DS_Store` (see commit for details).

Fixes https://github.com/rust-lang/cargo/issues/15145


epage (approved):

epage:
Thanks!

Issue #14415, opened 2024-08-16
`resolver="2"` does make it impossible to generate feature gated code from a proc macro in a relaiable way


weiznich:
### Problem

Consider the following example: 

-   There is a crate "a" and "a_derives". These are separate as proc macros are required to live in a separate crate.
-   Crate "a" uses "a_derives" internally and offers a set of additive features (lets call them "x", "y", "z").
-   Crate "a_derives" has the same feature set and crate "a" forwards the features to "a_derives".
-   Crate "a_derives" generates code based on this features assuming that crate "a" contains the relevant helper code
-   Now a user of crate "a" adds it as dependency and as build-dependency to their own crate "c". The for the normal dependency they enable feature "x" and for the build dependency feature "y".

Assuming that host and target match:

-   Cargo with resolver="2" now builds crate "a_derive" once with the features "x" and "y". It will build crate "a" twice once with the features "a" and once with "b", which will both fail as both variants of crate "a" do not contain all the relevant code
-   Cargo with resolver="1" will build "a_derive" and "a" once with both features "x" and "y" enabled.

Given this situation: How do I resolve that while supporting both resolver versions. If the answer is: You cannot, then resolver = "2" is a breaking change and should have been marked as such (== shipping rust 2.0). I fully understand that it is far to late to roll back that change but I would like to see the cargo team to at least acknowledge that they broke something and that they should prioritizing working on a fix (and maybe more than working on implementing new features!).

### Steps

For a real word example try to build the following crate:

```toml
[package]
name = "test_resolver"
version = "0.1.0"
edition = "2021"

[dependencies]
diesel = { version = "=2.2.2", features = ["postgres"] }

[build-dependencies]
diesel = { version = "=2.2.1", features = ["sqlite"] }

```

This breaks diesel as we cannot encapsulate the proc-macros and the main crate as "one" unit. 

### Possible Solution(s)

* Do not unify features for proc macros between build-dependencies and normal dependencies
* Do unify not only the proc macro dependencies but also the normal crates between build dependencies and normal dependencies (i.e `resolver="1"` behavior.)

### Notes

See the offtopic part of https://github.com/rust-lang/cargo/issues/14406 for more discussion on this topic

### Version

```text
cargo 1.80.0 (376290515 2024-07-16)
release: 1.80.0
commit-hash: 37629051518c3df9ac2c1744589362a02ecafa99
commit-date: 2024-07-16
host: x86_64-unknown-linux-gnu
[...]

Issue #2980, opened 2016-08-09
Mutually exclusive features


retep998:
In Windows API there is the notion of a family and partitions. Your choice of family is basically which platform you are targeting and there is a choice between `WINAPI_FAMILY_DESKTOP_APP` `WINAPI_FAMILY_PC_APP` `WINAPI_FAMILY_PHONE_APP` `WINAPI_FAMILY_SYSTEM` and `WINAPI_FAMILY_SERVER`. Your choice of family in turn enables certain partitions which enable access to certain functionality. At the moment I cannot use cargo features for this because they are not mutually exclusive, so two downstream dependencies could specify different families which is completely wrong. Even if I use a build script to ensure only one such feature is selected, it would be a nightmare to ensure that _every_ single downstream dependency chooses the same family, threading the choice of family feature through long chains of transitive dependencies that shouldn't have to care about that choice.

What I think would work well here is some sort of feature selection that has to choose exactly one option from a closed set, and that choice is controlled by the root crate. All dependencies that are affected by that choice will then need some way to be informed of the choice so they can adjust which functionality to use depending on that choice.

At the moment the only workaround I see is for the user to pass an environment variable while building the project so winapi's build script can read that to choose the family followed by emitting key=value stuff which downstream dependencies can read via build scripts through the `DEP_WINAPI_key=value` environment variables to emit cfgs to control their own code.

Boddlnagg:
I just had this idea, but haven't thought it through yet: To get mutually exclusive features couldn't you just add some piece of code to the crate that is gated on the features that should not be used together and make `rustc` output some error message that actually points to the problem (do we have something like `#error` in C, probably a macro?).

This could be considered an ugly workaround, but maybe it allows the greatest flexibility.

retep998:
[...]

PR #15148, opened 2025-02-05
docs(cargo-fix): troubleshooting failing fixes

weihanglo:

There are a couple of reasons making `cargo fix` failed.
This doc update lists some approaches to narrow the scope
to help users deal with those failures one by one.

```
cargo build
target/debug/cargo help fix
```

I am not satisfied with this writing. Suggestions are welcome :)

### Additional information

See <https://github.com/rust-lang/rust/issues/136345>.

### DIFF
--- a/src/doc/man/cargo-fix.md
+++ b/src/doc/man/cargo-fix.md
@@ -52,11 +52,22 @@ The `cargo fix` subcommand can also be used to migrate a package from one
    (without the `--edition` flag) to apply any suggestions given by the
    compiler.
 
-And hopefully that's it! Just keep in mind of the caveats mentioned above that
-`cargo fix` cannot update code for inactive features or `cfg` expressions.
-Also, in some rare cases the compiler is unable to automatically migrate all
-code to the new edition, and this may require manual changes after building
-with the new edition.
+If you encounter failing fixes during migration, instead of fixing everything
+in a single `cargo fix` invocation, try fixing only a subset of packages for
+specific features and Cargo targets. Fixing each combination separately can
+help identify the root cause of failures more easily and avoid unintended
+feature unification between packages. For example:
+
+* Use the `--package` flag to fix only some packages, e.g., `cargo fix --package foo`.
+* Instead of `--all-features`, use the `--features` flag to activate a specific
+  set of features, e.g., `cargo fix --package foo --features myfeat`.
+* `cargo fix` implies `--all-targets` by default. Use target selection flags to
+  fix only a subset of Cargo targets, e.g., `cargo fix --bin mybin`.
+
+Also keep in mind of the caveats mentioned above that `cargo fix` cannot update
+code for inactive features or `cfg` expressions. Also, in some rare cases the
+compiler is unable to automatically migrate all code to the new edition,
+and this may require manual changes after building with the new edition.
 
 [edition]: https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html
 [edition field]: ../reference/manifest.html#the-edition-field
--- a/src/doc/man/cargo-fix.md
+++ b/src/doc/man/cargo-fix.md
@@ -37,10 +37,6 @@ pass `--target` to fix code for the given target.
 
     cargo fix --target x86_64-pc-windows-gnu
 
-If you encounter any problems with `cargo fix` or otherwise have any questions
-or feature requests please don't hesitate to file an issue at
-<https://github.com/rust-lang/cargo>.
-
 ### Edition migration
 
 The `cargo fix` subcommand can also be used to migrate a package from one
--- a/src/doc/man/generated_txt/cargo-fix.txt
+++ b/src/doc/man/generated_txt/cargo-fix.txt
@@ -53,11 +53,29 @@ DESCRIPTION
           (without the --edition flag) to apply any suggestions given by the
           compiler.
 
-       And hopefully that’s it! Just keep in mind of the caveats mentioned
-       above that cargo fix cannot update code for inactive features or cfg
-       expressions. Also, in some rare cases the compiler is unable to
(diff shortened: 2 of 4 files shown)


src/doc/man/cargo-fix.md
```diff
@@ -52,15 +52,37 @@ The `cargo fix` subcommand can also be used to migrate a package from one
...
+  cargo fix --bin mybin
+  ```
+
+* If you have a Cargo workspace containing multiple packages,
+  fix them one by one to avoid unwanted feature unification between packages.
+  For instance, fix only the `bar` package :
+
+  ```sh
+  cargo fix --package bar
+  ```
```
epage, line 79: imo these give suggestions without context around them or why other failures may happen which

This also feels like its heavily written for editions but is at the same heading level as editions

src/doc/man/cargo-fix.md
```diff
@@ -52,15 +52,37 @@ The `cargo fix` subcommand can also be used to migrate a package from one
...
+While `cargo fix` strives to automatically apply fixes, it can still fail for
+various reasons. Here are some suggestions for troubleshooting fix failures:
+
+* Unlike other commands, `cargo fix` implies `--all-targets` by default.
+  Try fixing one Cargo target at a time instead of everything in one run.
+  For instance, narrow the scope to a single binary target:
+
+  ```sh
+  cargo fix --bin mybin
+  ```
```
epage, line 71: Looks like we don't have precedence for code fences like this in the man pages.

In other parts of the book, we use `console` (for interactive) instead of `sh` (for scripts)

src/doc/man/cargo-fix.md
weihanglo: Yes it is missing contexts. I am not sure how to put one. The biggest one in my mind is feature unifications. There might be some other I am not aware of.

>  feels like its heavily written for edition

Not really? One can run plain `cargo fix` and hitting it hard due to having mutually exclusive features.

weihanglo: Hmm, maybe I can reframe as edition  troubleshooting, as we do give users specific steps for edition migration

weihanglo: Moved it to under migration and polished it a bit.

src/doc/man/cargo-fix.md
```diff
@@ -52,11 +52,22 @@ The `cargo fix` subcommand can also be used to migrate a package from one
...
-And hopefully that's it! Just keep in mind of the caveats mentioned above that
-`cargo fix` cannot update code for inactive features or `cfg` expressions.
-Also, in some rare cases the compiler is unable to automatically migrate all
-code to the new edition, and this may require manual changes after building
-with the new edition.
+If you encounter failing fixes during migration, instead of fixing everything
+in a single `cargo fix` invocation, try fixing only a subset of packages for
+specific features and Cargo targets. Fixing each combination separately can
+help identify the root cause of failures more easily and avoid unintended
+feature unification between packages. For example:
```
epage, line 59: This is less about "failing fixes" and more about "unrelated compilation errors", right?

For the conntext, if I'm understanding correctly, this is coming down to either helping the user are working around #2980 (a workflow that runs counter to Cargo's design) or dealing with #14415 (a bug).  This is not something unique to `cargo fix` except that `--all-targets` is the default which can get unexpected unification from dev-dependencies when combined with migrating multiple packages at once.  I would assume someone knows that they can't run `cargo check --workspace --all-targets` so I wonder why `cargo fix` failing would be so surprising and without knowing how to workaround it already.  Note that we give a `Migrating <build-target-path>` message for each build target, which is an indirect way of telling users that `--all-targets` is being used.

A small part of me also wonders about the balance in meeting a user where they are at vs documenting (and somewhat blessing) unsupported workflows.

weihanglo: Yes I agree the troubleshooting can be generalized to `cargo check`. To me the distinction is that unlike failures in `cargo check`, `cargo fix --all-features` is a suggested workflow here. People are exposed to potential failure immediately.

It would be great as well if we can find a place for general troubleshooting

weihanglo: > I would assume someone knows that they can't run `cargo check --workspace --all-targets` so I wonder why `cargo fix` failing would be so surprising and without knowing how to workaround it already.

I would also say from my observation at work, average Rust developers don't really know how each command actually works. What they do is fiddling around different flags, not to mention the indirection from `Migrating <path>` message. We are able to tell it's feature unification issue because we have read the cargo doc a hundred times.

> A small part of me also wonders about the balance in meeting a user where they are at vs documenting (and somewhat blessing) unsupported workflows.

[Mutually exclusive features](https://doc.rust-lang.org/nightly/cargo/reference/features.html#mutually-exclusive-features) isn't really officially unsupported. It is indeed unrecommended and Cargo features can't handle it well now though. The officially suggested alternative `cfg-if` may still cause issues like one wants is to test only a specific feature but it got overridden and test failed. So we also want to take care of those alternative workflows.

weihanglo:
Had some discussion during today's office hours.  

Some recaps:  

* The current troubleshooting section is actually more general than `cargo fix`. It may be worth merging it with [resolver troubleshooting](https://doc.rust-lang.org/nightly/cargo/reference/resolver.html#troubleshooting) and moving it to a new chapter in the Cargo Book.  
* This PR focuses on the feature unification failure case. However, `cargo fix` has many different failure modes. We are unsure how common feature unification issues (and mutually exclusive features) are. A crater run with `--all-features` might help.  
* Good troubleshooting should help users understand where they are so they can choose the most likely solution. Telling them to break down Cargo invocations without context might lead them in the wrong direction. We need to think more about the complete troubleshooting process for this scenario.

weihanglo:
Close as it needs a more larger-scale of doc writing

PR #15150, opened 2025-02-06
Simplify backtrack

Eh2406:

Some very small simplifications made while poking around the resolver. I was trying to understand when we would end up hitting a `continue` in the code. The easiest way was to run the tests with a panic instead. Turns out I was right, we cannot/(can no longer) hit that code path. Which allowed us to simplify some other surrounding code.

All tests pass on each commit.

### DIFF
--- a/src/cargo/core/resolver/mod.rs
+++ b/src/cargo/core/resolver/mod.rs
@@ -960,13 +960,6 @@ fn find_candidate(
     };
 
     while let Some(mut frame) = backtrack_stack.pop() {
-        let next = frame
-            .remaining_candidates
-            .next(&mut frame.conflicting_activations, &frame.context);
-        let Some((candidate, has_another)) = next else {
-            panic!("why did we save a frame that has no next?");
-        };
-
         // If all members of `conflicting_activations` are still
         // active in this back up we know that we're guaranteed to not actually
         // make any progress. As a result if we hit this condition we can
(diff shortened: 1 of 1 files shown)


src/cargo/core/resolver/mod.rs
```diff
@@ -999,6 +992,11 @@ fn find_candidate(
             }
         }
 
+        let (candidate, has_another) = frame
```
epage, line 990: Double checking: commit says this is moved out of the loop but its moved to later in the loop, right?

src/cargo/core/resolver/mod.rs
```diff
@@ -960,45 +960,38 @@ fn find_candidate(
...
+            // But lets just double check if the `pop`ed frame agrees.
+            let frame_to_new = frame.context.age >= age;
+            debug_assert!(
+                frame
+                    .context
+                    .is_conflicting(Some(parent.package_id()), conflicting_activations)
+                    == frame_to_new.then_some(age)
+            );
+
+            if frame_to_new {
```
epage, line 978:
```suggestion
            // Above we use `cx` to determine if this is going to be conflicting.
            // But lets just double check if the `pop`ed frame agrees.
            let frame_too_new = frame.context.age >= age;
            debug_assert!(
                frame
                    .context
                    .is_conflicting(Some(parent.package_id()), conflicting_activations)
                    == frame_too_new.then_some(age)
            );

            if frame_to_new {
```

src/cargo/core/resolver/mod.rs
Eh2406: Yes. The structure here is a little weird. The "hot path" is `while ... continue` which eventually exits by `while ... return`. Actually getting to the end of the loop only happens once in a resolution.

Eh2406: I pushed another commit reorganizing the loop to be more normal.

epage:
Thanks!

Issue #14774, opened 2024-11-02
Tracking Issue for workspace `feature-unification`


ehuss:
### Summary

RFC: [#3692](https://github.com/rust-lang/rfcs/pull/3692)
Implementation: TODO
Documentation: TODO

Adds the `resolver.feature-unification` configuration option to control how features are unified across a workspace.

### Unresolved Issues

- [ ] Bikeshed the config option naming.

### Future Extensions

- Support in manifests
- Dependency version unification
- Unify features in other settings

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

weiznich:
Given that my concerns on the RFC were ignored I will repeat them here as well. 

This feature can result in build failures in upstream dependency crates outside of the workspace if the workspace feature unification is used. 

The charge team should make clear in the document and communication around this feature that:

* This can happen and might not be fixable due to other deficiencies in cargo and the wider rust ecosystem. 
* This is entirely up to the user that sets this setting and not to the upstream crate
* Ideally these information should be presented by cargo as part of the compiler error message 


Also as I personal note I want to emphasise again that I'm highly disappointed how the cargo team seems to handle concerns and suggestions to resolve these concerns. I'm also disappointed how the team decided to (not) communicate their decision. For me that doesn't look like a team that's interested in outside contributions. For me personally that means that I will likely not contribute to cargo anymore, including taking care of concerns from the cargo team for potential concerns for features I design/work on. I also cannot recommend other persons to contribute to this team anymore. That effort is likely better spend somewhere else.

weiznich:
Whoever marked my comment as duplicate: Could you point to the "duplicated" comment in this thread?

Also could you point to an official answer to these suggestions from the cargo team and not from individual team members? I'm really not sure what you are expecting to happen if you just keep pretending to ignore meaningful suggestions.

epage:
[...]

PR #15157, opened 2025-02-08
feat: implement workspace feature unification

aliu:

Adds workspace feature unification for #14774 

In a workspace that has dependencies with different activated features depending on the packages being built, add the `resolver.feature-unification = "workspace"` option to `.cargo/config.toml`. Build the entire workspace with `--workspace` and then build individual packages, ensuring no dependencies are being recompiled.

### Additional information

Originally, the RFC and tracking issue mention some more complex changes required in cargo's dependency/feature resolution phases in order to support workspace feature unification. However, it seems like it also works by just modifying the list of `PackageIdSpec`s passed to the workspace resolver to include all workspace members, and then using the original list of specs when generating the build units. I'm wondering if I missed something because this change feels a bit *too* simple...

I tested it on a fairly large workspace containing about 100 packages and didn't see any recompilations when building with different sets of packages. I also added an integration test that verifies the correct features are enabled. If there are any other test cases I should include, please let me know and I'll try to add it.

### DIFF
--- a/src/cargo/core/workspace.rs
+++ b/src/cargo/core/workspace.rs
@@ -21,6 +21,7 @@ use crate::core::{
 use crate::core::{EitherManifest, Package, SourceId, VirtualManifest};
 use crate::ops;
 use crate::sources::{PathSource, SourceConfigMap, CRATES_IO_INDEX, CRATES_IO_REGISTRY};
+use crate::util::context::FeatureUnification;
 use crate::util::edit_distance;
 use crate::util::errors::{CargoResult, ManifestError};
 use crate::util::interning::InternedString;
@@ -112,7 +113,8 @@ pub struct Workspace<'gctx> {
     /// and other places that use rust version.
     /// This is set based on the resolver version, config settings, and CLI flags.
     resolve_honors_rust_version: bool,
-
+    /// The feature unification mode used when building packages.
+    resolve_feature_unification: FeatureUnification,
     /// Workspace-level custom metadata
     custom_metadata: Option<toml::Value>,
 
@@ -246,6 +248,7 @@ impl<'gctx> Workspace<'gctx> {
             requested_lockfile_path: None,
             resolve_behavior: ResolveBehavior::V1,
             resolve_honors_rust_version: false,
+            resolve_feature_unification: FeatureUnification::Selected,
             custom_metadata: None,
             local_overlays: HashMap::new(),
         }
@@ -307,13 +310,20 @@ impl<'gctx> Workspace<'gctx> {
                 }
             }
         }
-        if let CargoResolverConfig {
-            incompatible_rust_versions: Some(incompatible_rust_versions),
-        } = self.gctx().get::<CargoResolverConfig>("resolver")?
-        {
+        let config = self.gctx().get::<CargoResolverConfig>("resolver")?;
+        if let Some(incompatible_rust_versions) = config.incompatible_rust_versions {
             self.resolve_honors_rust_version =
                 incompatible_rust_versions == IncompatibleRustVersions::Fallback;
         }
+        if self.gctx().cli_unstable().feature_unification {
+            self.resolve_feature_unification = config
+                .feature_unification
+                .unwrap_or(FeatureUnification::Selected);
+        } else if config.feature_unification.is_some() {
+            self.gctx()
+                .shell()
+                .warn("ignoring `resolver.feature-unification` without `-Zfeature-unification`")?;
+        };
 
         Ok(())
     }
@@ -661,6 +671,14 @@ impl<'gctx> Workspace<'gctx> {
 
     pub fn resolve_honors_rust_version(&self) -> bool {
         self.resolve_honors_rust_version
+    }
+
+    pub fn set_resolve_feature_unification(&mut self, feature_unification: FeatureUnification) {
+        self.resolve_feature_unification = feature_unification;
+    }
+
+    pub fn resolve_feature_unification(&self) -> FeatureUnification {
+        self.resolve_feature_unification
     }
 
     pub fn custom_metadata(&self) -> Option<&toml::Value> {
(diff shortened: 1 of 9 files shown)


epage:
A couple of high level comments
- This will need to start as unstable. We generally have tests to make sure the feature is correctly gated
- We prefer for PRs to add tests in a commit before the feature, with the tests passing. Then the commit that adds the feature updates the tests to still pass
- We should have tests for any relevant cases from the RFC, e.g. `cargo install` is called out

weihanglo:
To add, for `cargo install`, according to [the RFC](https://github.com/rust-lang/rfcs/blob/1c590ce05d676e72e2217845ee054758d3a6df34/text/3692-feature-unification.md?plain=1#L125), the field should not apply the behavior to `cargo install`.

> This field will not apply to `cargo install` to match the behavior of `resolver.incompatible-rust-versions`.

src/cargo/core/workspace.rs
```diff
@@ -307,9 +307,10 @@ impl<'gctx> Workspace<'gctx> {
                 }
             }
         }
-        if let CargoResolverConfig {
-            incompatible_rust_versions: Some(incompatible_rust_versions),
-        } = self.gctx().get::<CargoResolverConfig>("resolver")?
+        if let Some(incompatible_rust_versions) = self
```
weihanglo, line 310: Seems unrelated to the feature unification behavior. Is this just a cosmetic change? If so, I'd prefer not doing this, or having it in a separate PR with a reason for the change.

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -272,13 +274,23 @@ pub fn create_bcx<'a, 'gctx>(
             HasDevUnits::No
         }
     };
+    let feature_unification = ws
+        .gctx()
+        .get::<CargoResolverConfig>("resolver")?
+        .feature_unification
+        .unwrap_or_default();
+    let resolve_specs = if let FeatureUnification::Workspace = feature_unification {
+        &Packages::All(Vec::new()).to_package_id_specs(ws)?
```
weihanglo, line 283: > However, it seems like it also works by just modifying the list of PackageIdSpecs passed to the workspace resolver to include all workspace members, and then using the original list of specs when generating the build units. I'm wondering if I missed something because this change feels a bit too simple...

For the `workspace` option I think this is pretty close to what we want. `cargo metadata` [does the same thing](https://github.com/rust-lang/cargo/blob/2928e32734b04925ee51e1ae88bea9a83d2fd451/src/cargo/ops/cargo_output_metadata.rs?plain=1#L139).

However, we will need to tweak bit for `cargo install` as <https://github.com/rust-lang/cargo/pull/15157#issuecomment-2646007116> said. It will be something like `Workspace::resolve_honors_rust_version`, or if you have any clever solution, welcome :)

tests/testsuite/feature_unification.rs
```diff
@@ -0,0 +1,112 @@
...
+        .run();
+
+    p.cargo("build --workspace")
+        .with_stderr_data(str![[r#"
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
+        .run();
+
+    p.process(&p.bin("ab")).with_stdout_data(str!["abc"]).run();
```
weihanglo, line 110: This is quite a brilliant way to check the unification result!

We usually use `--verbose` cargo execution and verify flags passed down to rustc. Maybe with that we

* Can replace `main.rs` with `lib.rs`. Building binaries requires linking and may slow down tests. 
* Can remove functions in the `common` lib, as we directly assert rustc invocations.

What do you think?

tests/testsuite/feature_unification.rs
weihanglo, line 1: For adding new unstable configurations, see

* https://doc.crates.io/contrib/process/unstable.html
* https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/features/index.html

We usually also add an [unstable doc](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html) whenever we add a new unstable feature. The [doc draft](https://github.com/rust-lang/rfcs/blob/1c590ce05d676e72e2217845ee054758d3a6df34/text/3692-feature-unification.md?plain=1#L88-L100) has been written in the RFC. Feel free to copy and adjust. If you feel writing doc is too much a burden, let us know we can do it in a follow-up PR.

tests/testsuite/feature_unification.rs
epage: We also try to prefer `cargo check` over `cargo build`

However, checking rustc flags has down sides and I feel like its good to avoid when possible 
- A lot of unrelated data to avoid
- We're not testing the end-user effect

aliu:
Thank you @epage and @weihanglo for the quick feedback! This is my first PR to cargo, so apologies if I didn't get some things right regarding the process, etc. I will try to address the comments and push an update today.

tests/testsuite/feature_unification.rs
aliu: I'm okay with writing it either way, but what is the consensus? If it's enough to just verify there are no recompilations then `cargo check` alone would be sufficient, however I can't think of any ways to actually verify the activated features without either using `--verbose` or `cargo build`.

weihanglo: > emitting a conditional compile_error! if not all the features are enabled

Sounds awesome!

src/cargo/core/workspace.rs
aliu: It was originally a cosmetic change and I reverted it, but I ended up having to change this area again when implementing the unstable feature gate.

weihanglo (approved):
Looks awesome! Thank you for the contribution.

weihanglo:
@aliu if you want to continue working on #14774, feel free to `@rustbot claim` it

PR #15164, opened 2025-02-09
Update deny.toml

ehuss:
`cargo deny` is complaining about some deprecated settings in our `deny.toml`. This updates it to the latest structure, and also synchronizes the file with the latest template.


weihanglo (approved):
Thanks!

PR #15172, opened 2025-02-11
refactor: Consolidate creation of SourceId from manifest path

epage:

This preps more features for cargo-script support and makes it clearer where we don't yet support it.

### DIFF
--- a/src/cargo/core/source_id.rs
+++ b/src/cargo/core/source_id.rs
@@ -204,6 +204,17 @@ impl SourceId {
         SourceId::new(SourceKind::Path, url, None)
     }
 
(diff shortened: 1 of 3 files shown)


weihanglo (approved):
This looks reasonable. Going to merge it. Just wonder what is the plan.

epage:
> Just wonder what is the plan.

This was focused on "consolidate logic" / "reduce the chance for mistakes".

I was doing this as I worked towards `cargo install --path` but I realized that is more like `dependencies.*.path` and held off for now.  When we support that, we'll likely want a similar constructor like this one.

PR #15187, opened 2025-02-14
fix(embedded): Handle more parsing corner cases

epage:

This is part of #12207.  I found these while implementing frontmatter support within rustc.

I'll likely do another pass when I finish rustc support to
- Unify tests between cargo and rustc
- Improve error messages


weihanglo (approved):
Looks good. Thanks!

PR #15190, opened 2025-02-15
Improve package-table-in-virtual-workspace errors

### LINKED ISSUE #13965: Improve the diagnostics in virual workspaces
### Problem

I tried to add the following to my virtual workspace:

```toml
[lints.rust]
rust-2018-idioms = "warn"
```

This results in the following diagnostic:

```console
❯ cargo check
error: failed to parse manifest at `/home/user/src/project/Cargo.toml`

Caused by:
  this virtual manifest specifies a `lints` section, which is not allowed
```

That diagnostic doesn't give a hint about that the solution is to rename the table to `workspace.lints.rust`. And googling the error did not turn up very useful results either.

### Proposed Solution

```console
❯ cargo check
error: failed to parse manifest at `/home/user/src/project/Cargo.toml`

Caused by:
  this virtual manifest specifies a `lints` section, which is not allowed
Suggestion:
  For workspaces, use `workspace.lints` instead
```

Note the added suggestion. Rustc is usually very good about giving useful hints. I assume Cargo should be able to do so too.

There is a question about what to do when you have a *non-virtual* workspace. It is not a feature I use myself, so I have no input on the diagnostic in that case.

### Notes

_No response_

spanishpear:

Fix #13965 - and implements clearer warning for incorrect keys/fields when using virtual workspaces.

Change is fairly obvious, you can view test file for changed output. 

### Additional information

The wording/language in the new message aligns to existing errors. 
For example using `If you were trying...` and `use ... instead.` as with other toml errors.

----
I ran into this today, and spent a decent amount of time before understanding my mistake 🤦 
I came across https://users.rust-lang.org/t/configuring-lints-for-an-entire-workspace/111995 and then #13965 - so decided to put up a quick PR :)

Thanks!

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -525,7 +525,8 @@ fn normalize_toml(
         normalized_toml.badges = original_toml.badges.clone();
     } else {
         if let Some(field) = original_toml.requires_package().next() {
-            bail!("this virtual manifest specifies a `{field}` section, which is not allowed");
+            bail!("this virtual manifest specifies a `{field}` section, which is not allowed.\n\
+                    If you were trying to provide fields for workspaces, use `workspace.{field}` instead.")
         }
     }
 
--- a/tests/testsuite/workspaces.rs
+++ b/tests/testsuite/workspaces.rs
@@ -2321,7 +2321,8 @@ fn ws_err_unused() {
 [ERROR] failed to parse manifest at `[..]/foo/Cargo.toml`
 
 Caused by:
-  this virtual manifest specifies a `{key}` section, which is not allowed
+  this virtual manifest specifies a `{key}` section, which is not allowed.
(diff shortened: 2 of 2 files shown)


src/cargo/util/toml/mod.rs
```diff
@@ -525,7 +525,8 @@ fn normalize_toml(
         normalized_toml.badges = original_toml.badges.clone();
     } else {
         if let Some(field) = original_toml.requires_package().next() {
```
epage, line 527: looking at https://github.com/rust-lang/cargo/blob/ce948f4616e3d4277e30c75c8bb01e094910df39/crates/cargo-util-schemas/src/manifest/mod.rs#L72, the suggestion doesn't apply to most fields.

spanishpear: Ah 🤦‍♀️ I clearly have forgotten how little i remember around crate fields. Will close this for now as I'm not going to be able to look at it for a few weeks.

PR #15191, opened 2025-02-15
Fix man page with malformed `{{#options}}` block

max-heller:

The man page template for `cargo publish` includes a malformed `{{#options}}` block containing a freestanding paragraph outside the scope of the `{{#option}}` it relates to. This results in bad formatting in the generated man pages.

Review the diffs of the generated man pages.


ehuss (approved):
Thanks!

Issue #3772, opened 2017-02-28
Build configure fails due to missing Cargo


itarato:
I've noticed the Cargo documentation seems to require Cargo already installed on the system. Even so that the build instructions have a no-cargo scenario, if I run `./configure` it fails with the following error:

```bash
configure: processing ./configure args
configure: 
configure: CFG_PREFIX           := /usr/local 
configure: CFG_LOCAL_RUST_ROOT  :=  
configure: CFG_CARGO            := cargo 
configure: CFG_RUSTC            := rustc 
configure: CFG_RUSTDOC          := rustdoc 
configure: CFG_CARGO            :=  
configure: error: needed, but unable to find any of: CFG_CARGO cargo
```

In my case I compiled Rust from source, which doesn't have Cargo included anymore. Also as far as I know it is a bad practice to have a pre-existing Rust install (such as via rustup) when compiling Rust and Cargo. In any case, I couldn't figure out from the makefile or the docs if I really need to install Cargo before I build the project.
For clarity - my intention is to build Cargo (and Rust) from source.
What do you recommend?

alexcrichton:
Indeed yeah Cargo is required to compile Cargo. In future Rust releases we'll have `--enable-extended` which will compile Cargo as well as Rust.

In general though having an existing Rust install when compiling from source should work just fine, no need to avoid it.

itarato:
Thanks @alexcrichton

0joshuaolson1:
> In future Rust releases we'll have --enable-extended

@alexcrichton is there an [issue](https://github.com/rust-lang/cargo/issues?utf8=%E2%9C%93&q=enable-extended) for this?

alexcrichton:
@0joshuaolson1 it's currently present on the master branch of rust-lang/rust

0joshuaolson1:
That started with [this commit](https://github.com/rust-lang/rust/commit/44a01b8a54b078d15620d1133b94ee21ee7a6915), right? I see `--enable-extented` in [beta](https://github.com/rust-lang/rust/blob/beta/src/ci/docker/dist-x86_64-linux/Dockerfile#L91) and [stable](https://github.com/rust-lang/rust/blob/stable/src/ci/docker/cross/Dockerfile#L60) too.

Can I use the cargo build outside of Docker if my machine is the right arch?

alexcrichton:
Sorry I'm not really sure what you're asking? How does this relate to this issue?

0joshuaolson1:
[...]

PR #15192, opened 2025-02-16
fix: default to all targets when using `--edition` and ` --edition-idioms` in cargo fix

Rustin170506:

Close https://github.com/rust-lang/cargo/issues/13527

As we discussed, `cargo fix` should use the default target selection with `cargo check`.

In this PR, I modified `cargo fix` to no longer use all targets by default. For `cargo fix --edition` and `cargo fix --edition-idioms`, it will retain the old behavior and select all targets.

Unit tests

### DIFF
--- a/tests/testsuite/fix.rs
+++ b/tests/testsuite/fix.rs
@@ -198,6 +198,103 @@ fn prepare_for_2018() {
 }
 
 #[cargo_test]
+fn fix_tests_with_edition() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.1.0"
+                edition = "2018"
+            "#,
+        )
+        .file(
+            "src/lib.rs",
+            r#"
+                #![allow(ellipsis_inclusive_range_patterns)]
+                pub fn foo() {}
+
+                #[cfg(test)]
+                mod tests {
+                    #[test]
+                    fn it_works() {
+                        f();
+                    }
+                    fn f() -> bool {
+                        let x = 123;
+                        match x {
+                            0...100 => true,
+                            _ => false,
+                        }
+                    }
+                }
+            "#,
+        )
+        .build();
+
+    p.cargo("fix --edition --allow-no-vcs")
+        .with_stderr_data(str![[r#"
+[MIGRATING] Cargo.toml from 2018 edition to 2021
+[CHECKING] foo v0.1.0 ([ROOT]/foo)
+[MIGRATING] src/lib.rs from 2018 edition to 2021
+[FIXED] src/lib.rs (1 fix)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
+        .with_stdout_data("")
+        .run();
+    // Check that the test is fixed.
+    assert!(p.read_file("src/lib.rs").contains(r#"0..=100 => true,"#));
+}
+
+#[cargo_test]
+fn fix_tests_with_edition_idioms() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = 'foo'
+                version = '0.1.0'
+                edition = '2018'
+            "#,
+        )
+        .file(
+            "src/lib.rs",
+            r#"
+                pub fn foo() {}
+
+                #[cfg(test)]
+                mod tests {
+                    #[test]
+                    fn it_works() {
+                        f();
+                    }
+
+                    use std::any::Any;
+                    pub fn f() {
+                        let _x: Box<Any> = Box::new(3);
+                    }
+                }
+            "#,
+        )
+        .build();
+
+    p.cargo("fix --edition-idioms --allow-no-vcs")
+        .with_stderr_data(str![[r#"
+[CHECKING] foo v0.1.0 ([ROOT]/foo)
+[FIXED] src/lib.rs (1 fix)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
+        .with_stdout_data("")
+        .run();
+    // Check that the test is fixed.
+    assert!(p.read_file("src/lib.rs").contains("Box<dyn Any>"));
+}
+
+#[cargo_test]
 fn local_paths() {
     let p = project()
         .file(
--- a/tests/testsuite/fix.rs
+++ b/tests/testsuite/fix.rs
@@ -805,7 +805,7 @@ fn does_not_warn_about_dirty_ignored_files() {
 }
 
 #[cargo_test]
-fn fix_all_targets_by_default() {
+fn do_not_fix_tests_by_default() {
     let p = project()
         .file("src/lib.rs", "pub fn foo() { let mut x = 3; let _ = x; }")
         .file("tests/foo.rs", "pub fn foo() { let mut x = 3; let _ = x; }")
@@ -814,7 +814,7 @@ fn fix_all_targets_by_default() {
         .env("__CARGO_FIX_YOLO", "1")
         .run();
     assert!(!p.read_file("src/lib.rs").contains("let mut x"));
-    assert!(!p.read_file("tests/foo.rs").contains("let mut x"));
+    assert!(p.read_file("tests/foo.rs").contains("let mut x"));
 }
 
 #[cargo_test]
@@ -1424,7 +1424,7 @@ fn fix_to_broken_code() {
     p.cargo("build").cwd("foo").run();
 
     // Attempt to fix code, but our shim will always fail the second compile
-    p.cargo("fix --allow-no-vcs --broken-code")
+    p.cargo("fix --all-targets --allow-no-vcs --broken-code")
         .cwd("bar")
         .env("RUSTC", p.root().join("foo/target/debug/foo"))
         .with_status(101)
--- a/tests/testsuite/fix.rs
+++ b/tests/testsuite/fix.rs
@@ -1424,10 +1424,9 @@ fn fix_to_broken_code() {
     p.cargo("build").cwd("foo").run();
 
     // Attempt to fix code, but our shim will always fail the second compile
-    p.cargo("fix --all-targets --allow-no-vcs --broken-code")
+    p.cargo("fix --allow-no-vcs --broken-code")
         .cwd("bar")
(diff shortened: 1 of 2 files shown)


tests/testsuite/fix.rs
```diff
@@ -747,19 +834,6 @@ fn does_not_warn_about_dirty_ignored_files() {
     p.cargo("fix").run();
 }
 
-#[cargo_test]
-fn fix_all_targets_by_default() {
```
epage, line 711: Should we instead update this test for he new default behAVIOR?

tests/testsuite/fix.rs
```diff
@@ -1370,7 +1444,7 @@ fn fix_to_broken_code() {
     p.cargo("fix --allow-no-vcs --broken-code")
         .cwd("bar")
         .env("RUSTC", p.root().join("foo/target/debug/foo"))
-        .with_status(101)
+        .with_status(0)
```
epage, line 1447: Looking at the intent of this test, it seems like it should still be failing

Rustin170506: Yes. I am still looking at it. I'll fix it.

tests/testsuite/fix.rs
```diff
@@ -220,21 +220,108 @@ fn fix_tests() {
...
+                    #[test]
+                    fn it_works() {
+                        f();
+                    }
+                    fn f() -> bool {
+                        let x = 123;
+                        match x {
+                            0...100 => true,
+                            _ => false,
+                        }
```
epage, line 259: Can we make the edition-fix code reuse a pattern from an existing test, so we constrain rustc the least amount?  Looks like we have 2015 -> 2018 with `crate::`

tests/testsuite/fix.rs
Rustin170506: I used `--all-targets` to fix the status change, but I still don’t understand why this case would be affected by the test target selection. It seems there are no test targets in this case at all. Or did I miss something?

tests/testsuite/fix.rs
```diff
@@ -220,21 +220,108 @@ fn fix_tests() {
...
 [CHECKING] foo v0.0.1 ([ROOT]/foo)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
+"#]])
+        .with_stdout_data("")
+        .run();
+}
+
+#[cargo_test]
+fn fix_tests_with_edition() {
```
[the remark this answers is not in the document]
Rustin170506: Moved.

tests/testsuite/fix.rs
Rustin170506: Yes, that sounds reasonable.

tests/testsuite/fix.rs
epage: Missed that when I was looking.  Thats fine.

tests/testsuite/fix.rs
```diff
@@ -1424,7 +1424,7 @@ fn fix_to_broken_code() {
     p.cargo("build").cwd("foo").run();
 
     // Attempt to fix code, but our shim will always fail the second compile
-    p.cargo("fix --allow-no-vcs --broken-code")
+    p.cargo("fix --all-targets --allow-no-vcs --broken-code")
```
epage, line 1427: As you said, this is suspicious and we should understand why in case there is a bad interaction happening

Rustin170506: Yes, I’m still looking into it. It seems that if we don’t select the test target, the shim foo is called only twice. However, if we select the test target, it will be called three times, I believe.

Rustin170506: And if you have two targets, it also works: `p.cargo("fix --lib --tests --allow-no-vcs --broken-code")`
```console
[COMPILING] bar v0.1.0 ([ROOT]/foo/bar)
[WARNING] failed to automatically apply fixes suggested by rustc to crate `bar`
This likely indicates a bug in either rustc or cargo itself,
and we would appreciate a bug report! You're likely to see
a number of compiler warnings after this message which cargo
attempted to fix but failed. If you could open an issue at
https://github.com/rust-lang/rust/issues
quoting the full output of this command we'd be very appreciative!
Note that you may be able to make some more progress in the near-term
fixing code with the `--broken-code` flag

The following errors were reported:

thread 'main' panicked at src/main.rs:23:29:
explicit panic
[NOTE] run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Original diagnostics will follow.

[WARNING] variable does not need to be mutable
 --> src/lib.rs:1:20
  |
1 | pub fn foo() { let mut x = 3; let _ = x; }
  |                    ----^
  |                    |
  |                    [HELP] remove this `mut`
  |
  = [NOTE] `#[warn(unused_mut)]` on by default

[WARNING] `bar` (lib test) generated 1 warning (run `cargo fix --lib -p bar --tests` to apply 1 suggestion)

thread 'main' panicked at src/main.rs:23:29:
explicit panic
[NOTE] run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[ERROR] could not compile `bar` (lib)

Caused by:
  process didn't exit successfully: `/Volumes/t7/code/cargo/target/debug/cargo [ROOT]/foo/foo/target/debug/foo --crate-name bar --edition=2015 src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type lib --emit=dep-info,metadata -C embed-bitcode=no -C debuginfo=2 -C split-debuginfo=unpacked --check-cfg 'cfg(docsrs,test)' --check-cfg 'cfg(feature, values())' -C metadata=4b0938be085d245b -C extra-filename=-d245623a02104f1e --out-dir [ROOT]/foo/bar/target/debug/deps -L dependency=[ROOT]/foo/bar/target/debug/deps` ([EXIT_STATUS]: 101)
```

Rustin170506: I think it’s related to how many times foo rustc is called in this command. If it’s called only twice for one target, the panic error is caught as a warning. However, if it’s called again, it turns into a compile error and causes a bailout. Then the status becomes 101.
I’ll dig into it further. I’m not sure how we handle the compile error here, but I’ll take a look tomorrow.

Rustin170506: I think this change is reasonable. The test returns a 101 status because rustc was called three times. The first time, it attempts to fix build.rs with no errors. The second time, it tries to fix src/lib.rs, where there is one error, but we have a mechanism to output the fixes as warnings when `--broekn-code` was specified. The third time, it attempts to fix src/lib.rs again, but since no fixes are needed, the panic error is displayed directly, resulting in the 101 status.

See: https://github.com/rust-lang/cargo/blob/99745f078eff140c6e7c28be50d48a33e315674d/src/cargo/ops/fix.rs#L695

Rustin170506: I think we should remove -all-target. The status 0 is expected here

epage: @ehuss any idea what this test *should* be doing?

Rustin170506: @ehuss could you please take a look when you have time? Thanks!

ehuss: Seems fine to remove the `with_status`. From what I can tell, the intent of this test is to check for the `warning` when it fails to run `rustc` to verify.

Rustin170506: How about we keep with_status? I think it’s still useful to verify whether we can catch the error and use the correct status code here when `--broken-code` is specified.

tests/testsuite/fix.rs
```diff
@@ -805,7 +805,7 @@ fn does_not_warn_about_dirty_ignored_files() {
 }
 
 #[cargo_test]
-fn fix_all_targets_by_default() {
+fn do_not_fix_tests() {
```
epage, line 808: `_by_default`?  The name sounds like we should never fix tests

Rustin170506: Fixed!

tests/testsuite/fix.rs
Rustin170506: This is what I am suggesting: [`44a91af` (#15192)](https://github.com/rust-lang/cargo/pull/15192/commits/44a91afb10891bb86748989e73f98a3a2558b12b)

ehuss: 👍

epage (approved):

epage:
@rcbot fcp merge

We previous discussed this as a team (https://github.com/rust-lang/cargo/issues/13527#issuecomment-2634457438) and agreed that changing the non-`--edition` behavior seemed like the right call.  This FCP is to official record that as we are making a behavior change.

Background:

`cargo fix` was developed with a focus on Editions which including running `--all-targets` by default so that you migrated your whole project.  With [RFC #3772](https://rust-lang.github.io/rfcs/3772-build-target-edition.html) deprecating per-build-target editions, this makes even more sense.

However, it has grown to be a general way of applying fixes suggested by the compiler.  In this situation, having `cargo check` and `cargo fix` or `cargo clippy` and `cargo clippy --fix` diverge in behavior does not make sense.

In considering this, we felt that the non-edition use of `cargo fix` / `cargo --fix` is unlikely to be used programmatically in a way negatively impacted by this change, considering it can't fully resolve all lints, it can only work in some situations (due to dirty vcs), and that it can generate broken code.

Rustin170506:
> @rcbot fcp merge

@rfcbot fcp merge

See https://github.com/rust-lang/cargo/pull/15192#issuecomment-2813005009

tests/testsuite/fix.rs
arlosi: I think `with_status(0)` is the default, so it should not be necessary.

tests/testsuite/fix.rs
```diff
@@ -1330,7 +1427,7 @@ fn fix_to_broken_code() {
     p.cargo("fix --allow-no-vcs --broken-code")
         .cwd("bar")
         .env("RUSTC", p.root().join("foo/target/debug/foo"))
-        .with_status(101)
+        .with_status(0)
```
arlosi, line 1430: `with_status(0)` is the default. I think you can just delete the line.

tests/testsuite/fix.rs
Rustin170506: Updated. Thank you!

epage:
Thanks again @Rustin170506 !

Rustin170506:

PR #11428, opened 2022-11-27
core::compiler: remove Executor trait

nrc:
### What does this PR try to resolve?

This PR remove the Executor trait and thus simplifies some code. I added this trait for the RLS and it was never advertised as a publicly available API. The RLS is dead now and I think this API is not the best way to achieve its goals, therefore I think it should be removed.

### How should we test and review this PR?

Change is trivial refactoring, so existing tests should be fine. Eyeball review should be fine.

### Additional information

r? @arlosi

PR #15196, opened 2025-02-17
refactor: Remove executor trait

dev-ardi:
- **docs: Improve comments**
- **refactor: Remove Executor trait**

It looks like there was an `Executor` trait that existed but doesn't make much sense now,
since it's only implemented by a unit struct, so I simplify this by calling a method directly.

Hopefully this will also bring a minuscule perf gain but more importantly simplify cargo a bit.


ehuss:
I believe in https://github.com/rust-lang/cargo/pull/11428 we decided to keep the trait since some API users are using it.

Issue #14544, opened 2024-09-14
workspace.default-members causes cargo run --bin from-other-crate to fail with an unhelpful error


kornelski:
### Problem

In a workspace with `default-members` set, `cargo run --bin` considers only these members as expected. However, when specifying another workspace binary without `-p`, the error message is unhelpful, because it doesn't hint which package needs to be selected.

### Steps

`cargo run --bin other` where the binary is in the workspace, but not in the default workspace member.

### Possible Solution(s)

The error message could search other crates in the workspace and list which have a binary with the given name.

### Version

```text
cargo 1.83.0-nightly (c1fa840a8 2024-08-29)
release: 1.83.0-nightly
commit-hash: c1fa840a85eca53818895901a53fae34247448b2
commit-date: 2024-08-29
host: aarch64-apple-darwin
libgit2: 1.8.1 (sys:0.19.0 vendored)
libcurl: 8.7.1 (sys:0.4.74+curl-8.9.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Mac OS 15.0.0 [64-bit]
```

epage:
Could you provide full reproduction steps and example failure output?

kornelski:
```bash
cargo new crate1
cargo new crate2
```

+ `Cargo.toml` at the root

```toml
[workspace]
members = ["crate1", "crate2"]
default-members = ["crate1"]
```

and

```bash
cargo r --bin crate2
```

gives

```text
error: no bin target named `crate2`

	Did you mean `crate1`?
```

Alternatively, without `default-members`, `cargo r --bin` gives:

```text
error: "--bin" takes one argument.
Available binaries:
    crate1
    crate2
```

with `default-members` it considers just the members, which is technically correct:

```text
error: "--bin" takes one argument.
Available binaries:
    crate1
```

but it could be more helpful if it also looked at other workspace members and suggested `-p` to select them with their binaries.

epage:
From a users perspective, it seems reasonable to improve.  My only question is how much it would take to get the right information into the right place.

linyihai:
The `Available` packages were retrieved by these functions
https://github.com/rust-lang/cargo/blob/3b784a42e3075ebf34591ebfd5b79457e680e69d/src/cargo/util/workspace.rs#L12-L18

`options.spec` is a `Packages::Default` enum type.

[...]

PR #15199, opened 2025-02-18
Get all members as `available targets` even though default-members was specified.

linyihai:

The old behavior gets members from `default-members` if specified in a virtual workspace, this PR gets all workspace members for `Available` target hints. 

Fixes https://github.com/rust-lang/cargo/issues/14544

The first commit added the test, the second commit addressed the issue.

### DIFF
--- a/src/cargo/ops/cargo_compile/unit_generator.rs
+++ b/src/cargo/ops/cargo_compile/unit_generator.rs
@@ -247,15 +247,15 @@ impl<'a> UnitGenerator<'a, '_> {
         mode: CompileMode,
     ) -> CargoResult<Vec<Proposal<'a>>> {
         let is_glob = is_glob_pattern(target_name);
-        let proposals = if is_glob {
-            let pattern = build_glob(target_name)?;
-            let filter = |t: &Target| is_expected_kind(t) && pattern.matches(t.name());
-            self.filter_targets(filter, true, mode)
-        } else {
-            let filter = |t: &Target| t.name() == target_name && is_expected_kind(t);
-            self.filter_targets(filter, true, mode)
+        let pattern = build_glob(target_name)?;
+        let filter = |t: &Target| {
+            if is_glob {
+                is_expected_kind(t) && pattern.matches(t.name())
+            } else {
+                is_expected_kind(t) && t.name() == target_name
+            }
         };
-
+        let proposals = self.filter_targets(filter, true, mode);
         if proposals.is_empty() {
             let targets = self
                 .packages
--- a/src/cargo/ops/cargo_compile/unit_generator.rs
+++ b/src/cargo/ops/cargo_compile/unit_generator.rs
@@ -48,6 +48,7 @@ struct Proposal<'a> {
 pub(super) struct UnitGenerator<'a, 'gctx> {
     pub ws: &'a Workspace<'gctx>,
     pub packages: &'a [&'a Package],
+    pub spec: &'a Packages,
     pub target_data: &'a RustcTargetData<'gctx>,
     pub filter: &'a CompileFilter,
     pub requested_kinds: &'a [CompileKind],
@@ -288,24 +289,42 @@ impl<'a> UnitGenerator<'a, '_> {
                 CargoResult::Ok(())
             };
 
+            let unmatched_packages = || match self.spec {
+                Packages::Default | Packages::OptOut(_) | Packages::All(_) => {
+                    "default-run packages".to_owned()
+                }
+                Packages::Packages(packages) => {
+                    let first = packages
+                        .first()
+                        .expect("The number of packages must be at least 1");
+                    if packages.len() == 1 {
+                        format!("`{}` package", first)
+                    } else {
+                        format!("`{}`, ... packages", first)
+                    }
+                }
+            };
+
             let mut msg = String::new();
             if !suggestion.is_empty() {
                 write!(
                     msg,
-                    "no {} target {} `{}`{}",
+                    "no {} target {} `{}` in {}{}",
                     target_desc,
                     if is_glob { "matches pattern" } else { "named" },
                     target_name,
+                    unmatched_packages(),
                     suggestion,
                 )?;
                 append_targets_elsewhere(&mut msg, "\n")?;
             } else {
                 writeln!(
                     msg,
-                    "no {} target {} `{}`.",
+                    "no {} target {} `{}` in {}.",
                     target_desc,
                     if is_glob { "matches pattern" } else { "named" },
                     target_name,
+                    unmatched_packages()
                 )?;
 
                 append_targets_elsewhere(&mut msg, "")?;
--- a/src/cargo/ops/cargo_compile/unit_generator.rs
+++ b/src/cargo/ops/cargo_compile/unit_generator.rs
@@ -17,6 +17,7 @@ use crate::util::{closest_msg, CargoResult};
 
 use super::compile_filter::{CompileFilter, FilterRule, LibRule};
 use super::packages::build_glob;
+use super::Packages;
 
 /// A proposed target.
 ///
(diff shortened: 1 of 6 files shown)


tests/testsuite/workspaces.rs
```diff
@@ -2716,3 +2716,51 @@ fn nonexistence_package_togother_with_workspace() {
...
+        );
+
+    let p = p.build();
+    p.cargo("run --bin")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] "--bin" takes one argument.
+Available binaries:
+    crate1
+    crate2
```
epage, line 2761: > Fixes #14544

If the fix only adds bins from all targets to this error theo it will lead to user frustration as they then add `--bin crate2` and it fails without saying why.

The core problem in the Issue is the `--bin crate2` error message. It only suggested touching this error message as an addendum and suggested it show what the needed package is. I agree with that with the caveat that we need to double check if our completion or fish's completion script uses the output from this. While this is meant for humans and can change, we still want to play nice with these.

src/cargo/util/workspace.rs
```diff
@@ -47,9 +59,14 @@ fn print_available_targets(
...
-        for target in targets {
-            writeln!(output, "    {}", target)?;
+        for (package, targets) in targets {
+            if targets.is_empty() {
+                continue;
+            }
+            writeln!(output, "Available {} in {}:", plural_name, package)?;
+            for target in targets {
+                writeln!(output, "    {}", target)?;
+            }
```
epage, line 69: This is unrelated to the other changes and should be in its own commit

src/cargo/ops/cargo_compile/unit_generator.rs
```diff
@@ -267,35 +268,61 @@ impl<'a> UnitGenerator<'a, '_> {
...
-                    if is_glob { "matches pattern" } else { "named" },
-                    target_name,
-                    suggestion
-                );
+            let target_elsewhere = if is_glob {
+                let pattern = build_glob(target_name)?;
+                let filter = |t: &Target| is_expected_kind(t) && pattern.matches(t.name());
+                self.get_targets_from_other_packages(filter)?
             } else {
-                let mut msg = String::new();
```
epage, line 279: Refactoring to merge these code paths should probably be in its own commit

src/cargo/ops/cargo_compile/unit_generator.rs
```diff
@@ -267,35 +268,61 @@ impl<'a> UnitGenerator<'a, '_> {
...
-                writeln!(
-                    msg,
-                    "no {} target {} `{}`.",
```
epage, line 282: unifying whether we have a `.` or not should probably be in its own commit.  I would lean instead to keep it instead of get rid of it

src/cargo/ops/cargo_compile/unit_generator.rs
```diff
@@ -267,35 +268,61 @@ impl<'a> UnitGenerator<'a, '_> {
...
+            let mut msg = String::new();
+            write!(
+                msg,
+                "no {} target {} `{}`",
+                target_desc,
+                if is_glob { "matches pattern" } else { "named" },
+                target_name
+            )?;
+            if let Some((package, target)) = target_elsewhere {
+                writeln!(msg, "\nAvailable {target_desc} targets in {package}:")?;
```
epage, line 289: Could we make `in {package}` its own commit as well?  That is an independent improvement

src/cargo/ops/cargo_compile/unit_generator.rs
linyihai: I chose keeping the `.` , if it feels good to get rid of it then it will sperate into another refactor PR

tests/testsuite/workspaces.rs
```diff
@@ -2778,9 +2778,9 @@ Available binaries:
     p.cargo("run --bin crate2")
         .with_status(101)
         .with_stderr_data(str![[r#"
-[ERROR] no bin target named `crate2`
+[ERROR] Available bin targets in crate2:
+ crate2
 
-[HELP] a target with a similar name exists: `crate1`
```
epage: I don't think these are mutually exclusive

tests/testsuite/workspaces.rs
```diff
@@ -2778,9 +2778,9 @@ Available binaries:
     p.cargo("run --bin crate2")
         .with_status(101)
         .with_stderr_data(str![[r#"
-[ERROR] no bin target named `crate2`
+[ERROR] Available bin targets in crate2:
```
epage, line 2782: We are no longer telling the user about the error.  We should preserve the old error message and put this behind a `[HELP]`

src/cargo/ops/cargo_compile/unit_generator.rs
```diff
@@ -267,7 +268,21 @@ impl<'a> UnitGenerator<'a, '_> {
...
             let suggestion = closest_msg(target_name, targets.iter(), |t| t.name(), "target");
-            if !suggestion.is_empty() {
+            let target_elsewhere = if is_glob {
+                let pattern = build_glob(target_name)?;
+                let filter = |t: &Target| is_expected_kind(t) && pattern.matches(t.name());
+                self.get_targets_from_other_packages(filter)?
+            } else {
+                let filter = |t: &Target| t.name() == target_name && is_expected_kind(t);
+                self.get_targets_from_other_packages(filter)?
+            };
```
epage, line 278: Would it be worth extracting the general pattern of filter creation rather than duplicate it?

tests/testsuite/workspaces.rs
epage: While the focus of #14544 is now included, two of the 3 improvements still require investigating their impact on cargo's and fish's completions and one still has the issue of showing values that can't be used.

tests/testsuite/build.rs
```diff
@@ -1352,6 +1352,7 @@ Available bin targets:
 
 [HELP] a target with a similar name exists: `a`
 
+
 "#]])
```
epage, line 1356: This extra newline doesn't seem right

tests/testsuite/workspaces.rs
```diff
@@ -2835,8 +2840,8 @@ Available bin targets:
         .with_status(101)
         .with_stderr_data(str![[r#"
 [ERROR] no bin target named `crate2`.
-Available bin targets:
-    another
+Available bin targets in crate2:
```
epage: Should we say

> no bin target named `crate2` in crate1 package.

or

> no bin target named `crate2` in default-run packages

Without that extra context, I worry users might get confused because they know might know these targets come from other packages but not realize it

(this should probably be its own commit)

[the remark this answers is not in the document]
linyihai: I removed the `targets` that it seems not nessary.

src/cargo/ops/cargo_compile/unit_generator.rs
```diff
@@ -247,15 +248,15 @@ impl<'a> UnitGenerator<'a, '_> {
...
-            let filter = |t: &Target| t.name() == target_name && is_expected_kind(t);
-            self.filter_targets(filter, true, mode)
+        let pattern = build_glob(target_name)?;
+        let filter = |t: &Target| {
+            if is_glob {
+                is_expected_kind(t) && pattern.matches(t.name())
+            } else {
+                is_expected_kind(t) && t.name() == target_name
+            }
         };
```
epage, line 258: A refactor for pulling out the `filter` creation should probably be in its own refactor commit

tests/testsuite/workspaces.rs
```diff
@@ -2716,3 +2716,137 @@ fn nonexistence_package_together_with_workspace() {
...
+        .run();
+
+    p.cargo("run --bin crate2")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] no bin target named `crate2` in default-run packages
+
+[HELP] a target with a similar name exists: `crate1`
+Available bin in `crate2` package:
+    crate2
```
epage, line 2810: Shouldn't these have separate `[HELP]` prefixes?

tests/testsuite/workspaces.rs
epage: @linyihai maybe we should just split this PR so we can move forward the `--bin foo` case and deal with `--bin` separately

linyihai: Let us focus on the `--bin foo case` case

src/cargo/ops/cargo_compile/unit_generator.rs
```diff
@@ -267,14 +268,36 @@ impl<'a> UnitGenerator<'a, '_> {
                 })
                 .collect::<Vec<_>>();
             let suggestion = closest_msg(target_name, targets.iter(), |t| t.name(), "target");
+            let targets_elsewhere = self.get_targets_from_other_packages(filter)?;
+            let need_append_targets_elsewhere = !targets_elsewhere.is_empty();
+            let append_targets_elsewhere = |msg: &mut String| {
+                for (package, targets) in targets_elsewhere {
+                    if !targets.is_empty() {
+                        writeln!(msg, "help: Available {target_desc} in `{package}` package:")?;
```
epage, line 277: Sorry I missed this but `Available` should be lower case, see https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-output-style-guide

epage: Except we already use `Available`, so we can skip this for now.

tests/testsuite/workspaces.rs
```diff
@@ -2800,6 +2800,9 @@ Available binaries:
 [ERROR] no bin target named `crate2`
 
 [HELP] a target with a similar name exists: `crate1`
+[HELP] Available bin in `crate2` package:
+    crate2
+
```
epage, line 2805: extra blank lines are back?

epage: Should be good to merge with this resolved.

linyihai: Also fixed. Thanks.

epage (approved):
Thanks!

Issue #11100, opened 2022-09-16
Improve `cargo add -F crate/invalid_feat` error discovery


Emilgardis:
### Problem

When using `cargo add -F crate/feat`, if the `feat` feature does not exist, the error is not presented until the end.

<img width="285" alt="image" src="https://user-images.githubusercontent.com/1502855/190811822-8fd36e7c-363b-443d-8add-61e1e2bf7f1c.png">

This can be easy to miss, as most cargo warnings are presented with more context/text around them. `cargo add` however spits out a features table as if everything is ok and then presents an error.

### Proposed Solution

Attach more context in the output to clearly signal that "hey, this is not correct, check the error message"

<img width="280" alt="image" src="https://user-images.githubusercontent.com/1502855/190816028-01b9c6c8-d8e9-4422-97ef-f61161a13ebe.png">

### Notes

this has been implemented in #11098, creating this per request.

epage:
> This can be easy to miss, as most cargo warnings are presented with more context/text around them.

I feel like this is the key to the problem that we should focus on.

> cargo add however spits out a features table as if everything is ok and then presents an error.

Good call, we should probably **not** show the table like this.  We should probably split up `print_msg` to separate out the "Adding" from the table and put the table in after everything has been processed for that dependency.

The error should probably follow our usual pattern of providing a suggestion if one exists and showing everything if one doesn't exist.

Emilgardis:
One nice thing with providing the table, even on error, is that you can quickly see if you accidentally typoed.

> The error should probably follow our usual pattern of providing a suggestion if one exists

Perhaps for the attached example:

```
add tokio --features sync,net,rt,rr     
```

the error could be 

```
error: unrecognized feature `rr`
help: a feature with a similar name exists: `rt`
```

>  and showing everything if one doesn't exist.

what does this mean?

epage:
> One nice thing with providing the table, even on error, is that you can quickly see if you accidentally typoed or misremembered the feature flags name.

I think that is also what the alternative I brought up does.

> the error could be
> ...
> what does this mean?

This is an example of an existing "not found" error in cargo: https://github.com/rust-lang/cargo/pull/10999/files

Emilgardis:
so something like this?

`crate` has one feature `existing_feat`

```
$ cargo add -F crate/invalid_feat crate/existing_feat1
Adding crate v0.0.1 to dependencies
error: no feature `invalid_feat` available for `crate`
Available features:
    existing_feat
```

[...]

PR #15200, opened 2025-02-18
fix(add): Focus on error, rather than large feature lists 

epage:

Inspired by #11100 and previous work to collapse feature lists down.


weihanglo (approved):
Reasonable. Thank you!

Issue #12266, opened 2023-06-13
`cargo doc` doesn't detect dirty sources from parent directories


P1n3appl3:
### Problem

When sources aren't in a subdirectory of the cargo manifest, `cargo doc` fails to detect changes to those sources. The issue only occurs with `cargo doc` (not check or build). It isn't affected by the working directory (using `--manifest-path`), and using an absolute vs. relative path for the source makes no difference. 

### Steps

Use the following directory structure and `Cargo.toml` with an empty `lib.rs`:

```
manifest
└── Cargo.toml
lib.rs
```

```toml
[package]
name = "foo"
version = "0.1.0"
edition = "2021"

[lib]
path = "../lib.rs"
```

Running `cargo doc --manifest-path=manifest/Cargo.toml -v` should succeed. Now add a syntax error to `lib.rs`. Subsequent `cargo doc` runs will say that the crate is `Fresh` and exit without detecting the syntax error.

Removing `manifest/target` manually or by `cargo clean`ing will force a rebuild, but changes to `lib.rs` seem to never cause rebuilds. 

### Notes

This behavior was encountered on Fuchsia, where we primarily use a different build system (GN+ninja) but have experimental support for generating cargo manifests for rust targets. The `Cargo.toml`s are generated in an out directory and contain absolute paths pointing to the source root for each crate. We initially thought the issue had to do with the absolute paths or our use of symlinks, but it appears that change detection for `cargo doc` is broken for ANY sources that aren't in the same dir or subdirs of the manifest.

### Version

```text
Verified the behavior with both nightly `cargo 1.72.0-nightly (49b6d9e17 2023-06-09)` and stable `cargo 1.70.0 (ec8a8a0ca 2023-04-25)`
```

weihanglo:
Thanks for the report. I believe this is a known issue around the current fingerprint system (recompile detection). Cargo depends on file mtime to track whether it need to re-compile or not. 

For normal `cargo build` there are [dep-info] files which can be tracked under `target` directory. Unfortunately, for `cargo doc` there is no such file emitted from `rustdoc`[^1]. One of the naive solution here is walking through the filesystem from the root of the package. Finding the last modified file[^2] and comparing it with the output of `cargo doc` (index.html or something). Anything outside the root of the current package will be ignored. That's pretty much the root cause.

[...]

PR #15205, opened 2025-02-19
Fingerprint out of root targets

### LINKED ISSUE #15203: Incremental cargo doc doesn't support [lib] path to parent
### Problem

When running `cargo doc` incrementally on a project with a parent `[lib]` path:
```
[lib]
name = "crate1"
path = "../src/lib.rs"
```

edits to files are not detected, ending up in a previous doc version.

### Steps

See https://github.com/Vrixyz/cargo_parent_path.

1. Run `cargo doc --open`.

Output is:

```
 Documenting crate1 v0.1.0 (/path/to/here/crate1)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.17s
     Opening /path/to/here/target/doc/crate1/index.html
```

2. Change some documentation in src/lib.rs (and save).

Run again `cargo doc --open`.

Output is:

```
     Opening /path/to/here/target/doc/crate1/index.html
```

But should be the same as before.

3. Notice the opened documentation still shows the previous content, where we would have expected it to show the updated content.

### Possible Solution(s)

_No response_

### Notes

My current workaround is to run `cargo clean` (with `-p package` in bigger projects to avoid too expensive recompilation.

This bug had been present for a long time (I don´t believe that's a regression).

I detected this while working on https://github.com/dimforge/parry and similar crates.

### Version

```text
release: 1.83.0
commit-hash: 5ffbef3211a8c378857905775a15c5b32a174d3b
commit-date: 2024-10-29
os: Pop!_OS 22.4.0 (jammy) [64-bit]
```

krtab:
This PR partially fixes #15203 and #12266 but this is admitedly a bit hackish, let me know how you want to proceed with it.

### DIFF
--- a/src/cargo/sources/path.rs
+++ b/src/cargo/sources/path.rs
@@ -622,15 +622,29 @@ fn _list_files(pkg: &Package, gctx: &GlobalContext) -> CargoResult<Vec<PathEntry
 
         ignore_should_package(relative_path, is_dir)
     };
-
-    // Attempt Git-prepopulate only if no `include` (see rust-lang/cargo#4135).
-    if no_include_option {
-        if let Some(repo) = git_repo {
-            return list_files_gix(pkg, &repo, &filter, gctx);
+    let mut ret = match git_repo {
+        Some(repo) if no_include_option => {
+            // Attempt Git-prepopulate only if no `include` (see rust-lang/cargo#4135).
+            list_files_gix(pkg, &repo, &filter, gctx)?
+        }
+        _ => {
+            let mut ret = Vec::new();
+            list_files_walk(pkg.root(), &mut ret, true, &filter, gctx)?;
+            ret
+        }
+    };
+    for target in pkg.manifest().targets() {
+        let Some(path) = target.src_path().path() else {
+            continue;
+        };
+        let path = path.canonicalize()?;
+        if !path.starts_with(pkg.root()) {
+            // The target "root" is not included in the package root
+            // Unwrapping is ok because we know path is to a file
+            let target_root = path.parent().unwrap();
+            list_files_walk(target_root, &mut ret, false, &(|_, _| true), gctx)?;
         }
     }
-    let mut ret = Vec::new();
-    list_files_walk(pkg.root(), &mut ret, true, &filter, gctx)?;
     Ok(ret)
 }
 


krtab:
This is the hack described in https://github.com/rust-lang/cargo/issues/12266#issuecomment-1681297289 and hence has the same caveats that `#[path]` attributes won't be supported.

src/cargo/sources/path.rs
weihanglo, line 1: Thanks for helping with this issue!

As you've said, this is hacky. Some high-level feedback:

* `PathSource::list_file` (which then calls `_list_files`) is used in computing fingerprints for build scripts and `cargo doc`, as well as determining what is going to be packaged into `cargo package`. During packaging, any file outside the package root will not be packed into the resulting `.crate` file. Cargo even [emits a warning](https://github.com/rust-lang/cargo/blob/511e3ab9458ea6a6fa6a3eadc0b46c05829b0243/src/cargo/util/toml/mod.rs?plain=1#L3193-L3196) when the situation is detected. With this change, `cargo-package` starts including those files again. This is undesired, not even mention that where should their path be in the `.crate` tarball.
* How do this hack know what really needs to track? It simply looks at the directory the target source file is in and assume it always at the root directory with all other relevant files. Yeah it cannot handle `#[path]` we already knew, but it also doesn't respect git directory and `include`/`exclude` options. It may accidentally include more files they necessary.

To me I still think depinfo from rustdoc is a more sustainable approach. We might want to put more effort on that route than sorting out a hack.

Some other alternatives are on user side, like breaking them into sub packages for reasonable reuse.

weihanglo:
Also note that the issue is not labeled as `S-accepted`. It is recommended to [discuss designs in issues](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo) rather than in pull requests.

src/cargo/sources/path.rs
Vrixyz: - allow me to link https://github.com/rust-lang/rust/issues/91982 which is the tracking issue for the "sustainable" approach mentioned

weihanglo: Thanks! It is linked in <https://github.com/rust-lang/cargo/issues/12266#issuecomment-1591253839>, which also calls out a relevant Cargo issue <https://github.com/rust-lang/cargo/issues/9931>.

weihanglo:
@krtab have you tried [the `-Zrustdoc-depinfo` unstable flag](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#rustdoc-depinfo)? Did it work well on your sides? It would be appreciated to have your feedback :)

krtab:
Hi @weihanglo  I didn't really have a use case for it, and my bandwidth is pretty limited at the moment so I won't have time to test it unfortunately.
Thanks a lot for your work on this matter.

PR #15206, opened 2025-02-19
fix: build warning in windows_reserved_names_are_allowed

arlosi:
A recent change removed use of the `is_ci` function, but didn't remove the `use` statement.

### DIFF
--- a/crates/cargo-test-support/Cargo.toml
+++ b/crates/cargo-test-support/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-test-support"
-version = "0.7.1"
+version = "0.7.2"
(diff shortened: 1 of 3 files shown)


epage (approved):

crates/cargo-test-support/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-test-support"
-version = "0.7.1"
+version = "0.7.2"
```
epage, line 3: Should you also update https://github.com/rust-lang/cargo/blob/684bca20121f66a4fc7b02638222f50cf9cd8250/Cargo.toml#L36

Issue #10113, opened 2021-11-23
CARGO env var set incorrectly when Cargo invoked through ld


jonhoo:
### Problem

Earlier today I got a report from an internal developer about a build failing with an error like this:

```
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: CargoMetadata(".../Cargo.toml", Metadata(Output { status: ExitStatus(ExitStatus(32512)), stdout: "", stderr: "metadata: error while loading shared libraries: metadata: cannot open shared object file\n" }))', build.rs:30:14
```

The line in question is:
```rust
cbindgen::generate(&crate_dir).unwrap()
```
where
```rust
let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
```

Digging some more using `strace`, the developer found that the culprit was this `execve`:
```
execve("/usr/lib64/ld-2.26.so", ["/usr/lib64/ld-2.26.so", "metadata", "--all-features", "--format-version", "1", "--manifest-path", ".../Cargo.toml"], ...
```

This led me down a bit of a rabbit hole. The error propagates an error from when `cbindgen` tries to [invoke `cargo metadata`](https://github.com/eqrion/cbindgen/blob/93c06c5c9d319f481788c9670700097b4e46d270/src/bindgen/cargo/cargo.rs#L47-L49), which constructs the actual Command [here](https://github.com/eqrion/cbindgen/blob/93c06c5c9d319f481788c9670700097b4e46d270/src/bindgen/cargo/cargo_metadata.rs#L233). The name of the command is determined by the `CARGO` environment variable, which [Cargo sets](https://github.com/rust-lang/cargo/blob/3d3ddd267f31d99d6c72d54218ff935836d6a732/src/cargo/core/compiler/compilation.rs#L312) based on [`Config::cargo_exe`](https://github.com/rust-lang/cargo/blob/f066c50d653a70fb2b7613f2f7ad603f04331fb7/src/cargo/util/config/mod.rs#L383). Now _that_, in turn, is set based on [`std::env::current_exe`](https://doc.rust-lang.org/std/env/fn.current_exe.html), which gets the "full path to the current executable" (Linux impl is [reading `/proc/self/exe`](https://github.com/rust-lang/rust/blob/074f63648bd2368d5ca19aed02b5763a144e5d05/library/std/src/sys/unix/os.rs#L354)). 

[...]

PR #15208, opened 2025-02-20
fix: reset $CARGO if the running program is real `cargo[.exe]` 

### LINKED ISSUE #15099: `$CARGO` passed to external subcommands can point to the wrong `cargo` if already set
### Problem

Observed in https://github.com/rust-lang/rust-clippy/issues/14045, if `cargo +nightly clippy` is ran by a process with `CARGO` pointing to the stable `cargo` then nightly `clippy-driver` will be ran by stable `cargo`

In this case that resulted in a confusing warning

### Steps

1. Put the following into your `PATH` as `cargo-foo`

   ```bash
   #!/usr/bin/env bash

   echo "$CARGO"
   ```
2. With stable cargo use `cargo run` on

   ```rust
   use std::process::Command;
   
   fn main() {
       Command::new("cargo")
           .args(["+nightly", "foo"])
           .status()
           .unwrap();
   }
   ```

3. The output will be `.../.rustup/toolchains/stable-.../bin/cargo`

### Possible Solution(s)

Looks to be due to https://github.com/rust-lang/cargo/pull/11285, perhaps there could be an exception for when the current exe really is `cargo`

### Notes

_No response_

### Version

```text
cargo 1.84.0 (66221abde 2024-11-19)
```

smoelius:
Overwrite `$CARGO` when the current exe is detected to be a cargo binary.

From https://github.com/rust-lang/cargo/issues/15099#issuecomment-2666737150:
> opt-in behavior (set by cargo-the-bin):
> 
> 1. `from_current_exe` if its `cargo[EXE_SUFFIX]`
> 2. `from_argv` if its `cargo[EXE_SUFFIX]`
> 3. `from_env`
> 4. `from_current_exe`
> 5. `from_argv`

r? @epage

### DIFF
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -493,9 +493,25 @@ impl GlobalContext {
                     paths::resolve_executable(&argv0)
                 }
 
+                // Determines whether `path` is a cargo binary.
+                // See: https://github.com/rust-lang/cargo/issues/15099#issuecomment-2666737150
+                fn is_cargo(path: &Path) -> bool {
+                    path.file_stem() == Some(OsStr::new("cargo"))
+                }
+
+                let from_current_exe = from_current_exe();
+                if from_current_exe.as_deref().is_ok_and(is_cargo) {
+                    return from_current_exe;
+                }
+
+                let from_argv = from_argv();
+                if from_argv.as_deref().is_ok_and(is_cargo) {
+                    return from_argv;
+                }
+
                 let exe = from_env()
-                    .or_else(|_| from_current_exe())
-                    .or_else(|_| from_argv())
+                    .or(from_current_exe)
+                    .or(from_argv)
                     .context("couldn't get the path to cargo executable")?;
                 Ok(exe)
             })
--- a/tests/testsuite/cargo_command.rs
+++ b/tests/testsuite/cargo_command.rs
@@ -391,10 +391,13 @@ fn cargo_subcommand_env() {
         .canonicalize()
         .unwrap();
     let envtest_bin = envtest_bin.to_str().unwrap();
+    // Previously, `$CARGO` would be left at `envtest_bin`. However, with the
+    // fix for #15099, `$CARGO` is now overwritten with the path to the current
+    // exe when it is detected to be a cargo binary.
     cargo_process("envtest")
         .env("PATH", &path)
         .env(cargo::CARGO_ENV, &envtest_bin)
-        .with_stdout_data(format!("{}\n", envtest_bin).raw().raw())
+        .with_stdout_data(format!("{}\n", cargo.display()).raw())
         .run();
 }
 
--- a/tests/testsuite/cargo_command.rs
+++ b/tests/testsuite/cargo_command.rs
@@ -573,3 +573,87 @@ fn full_did_you_mean() {
 "#]])
         .run();
 }
+
+#[cargo_test]
+fn overwrite_cargo_environment_variable() {
+    // If passed arguments `arg1 arg2 ...`, this program runs them as a command.
+    // If passed no arguments, this program simply prints `$CARGO`.
+    let p = project()
+        .file("Cargo.toml", &basic_manifest("foo", "1.0.0"))
+        .file(
+            "src/main.rs",
+            r#"
+                fn main() {
+                    let mut args = std::env::args().skip(1);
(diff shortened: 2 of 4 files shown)


smoelius:
I'll fix the errors shortly.

src/cargo/util/context/mod.rs
```diff
@@ -493,9 +493,23 @@ impl GlobalContext {
                     paths::resolve_executable(&argv0)
                 }
 
+                fn is_cargo(path: &Path) -> bool {
```
weihanglo, line 496: We might want to write down a comment for why checking `cargo[.exe]` is needed.

weihanglo (commented):
Thanks for the patch!

Please allow me time to test against the dynamic linker use case.

src/cargo/util/context/mod.rs
```diff
@@ -493,9 +493,23 @@ impl GlobalContext {
...
+                    path.file_stem() == Some(OsStr::new("cargo"))
+                }
+
+                let from_current_exe_result = from_current_exe();
```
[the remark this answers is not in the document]
smoelius: I don't object, but just to be clear, that means that `from_current_exe` and `from_argv` shadow declarations from above.

smoelius:
Just because I am curious, is the reason for the recent CI failures known?

epage:
They were fixed in #15211
- One was because of a short term hack until 1.85 was released which then needed to be removed
- We're experimenting with automating the fix for the other one in #15212.

smoelius:
> * We're experimenting with automating the fix for the other one in [chore(ci): Auto-update cargo-semver-checks #15212](https://github.com/rust-lang/cargo/pull/15212).

Oh, interesting! Thank you for explaining!

smoelius:
If I'm interpreting the results right, path normalization is applied to sterr but not stdout?

Anyway, the PR should be ready for review.

weihanglo (commented):
Did manual tests for ld wrapper from what Joo's has described in #10113. This seems working!

One thing I am not sure is this characteristic:

> * We don't fix this issue for when users have an alternative name for the Cargo binary (or on some platforms, an alternative symlink name)

Should we document our heuristic rule in the CARGO env doc? I feel like it might be excessive and don't want to wait for a perfect doc to merge this fix.

tests/testsuite/freshness.rs
weihanglo, line 1: Feel like this should be moved to either `test/testsuite/cargo_command.rs` or `test/testsuite/run.rs` (perhaps the former is better?).

tests/testsuite/cargo_command.rs
```diff
@@ -391,10 +391,13 @@ fn cargo_subcommand_env() {
         .canonicalize()
         .unwrap();
     let envtest_bin = envtest_bin.to_str().unwrap();
+    // Previously, `$CARGO` would be left at `envtest_bin`. However, with the
```
weihanglo, line 394: Seeing this test change and I wonder if using `$CARGO` as a Cargo wrapper is a supported use case, and we don't want to reset it. 

Maybe we can detect iff the current `$CARGO` is `cargo[.exe]` and then we reset?

weihanglo: cc @epage as you proposed the current solution

smoelius:
> One thing I am not sure is this characteristic:
> 
> > * We don't fix this issue for when users have an alternative name for the Cargo binary (or on some platforms, an alternative symlink name)
> 
> Should we document our heuristic rule in the CARGO env doc? I feel like it might be excessive and don't want to wait for a perfect doc to merge this fix.

I would just point out that the current description of each variable is light on details: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

I think there some benefit to this, as it makes the list easy to scan.

tests/testsuite/cargo_command.rs
smoelius: I have no problem with implementing this is @epage agrees.

epage: `CARGO` was originally not intended for purposes of being a wrapper.  That people could do it was a byproduct of the implementation that was done but it wasn't intentional.  I wouldn't worry too much about it.

weihanglo: That is fair. Thanks Ed.

weihanglo (approved):
Thanks for working with us for this fix!

PR #15213, opened 2025-02-20
docs(ref): Shift focus to resolver v3

epage:

We have the `resolver` field in their as subtle nod to encourage people to explicitly set it in workspaces.  With `resolver = "3"` being out for a couple releases (at the time this hits stable), it seems appropriate for us to bump this value.


weihanglo (approved):

weihanglo (commented):
I don't see this controversial

PR #15218, opened 2025-02-21
chore(ci): Visually group output in Github

epage:

In a recent bump-check failure, it was hard to see how the change was relevant.  The hope is that [grouping the lines](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#grouping-log-lines) will filter the data down and provide enough context to make this easier


ehuss:
Thanks! Looks like clippy is having an issue.

PR #15221, opened 2025-02-22
Make sure search paths inside OUT_DIR precede external paths

### LINKED ISSUE #15220: link-search-path doesn't have a way to specify priority / doesn't auto deprioritize OS link paths
### Problem

I modified the llama.cpp repo that llama-cpp-2 has as a submodule to expose a new symbol & then added a method to call that new symbol in llama-cpp-2 via a Rust pub API. I then called the new Rust pub API in my code and added cpal as a dependency and the path to my local llama-cpp-2 as a dependency with the cuda feature on (cuda doesn't matter but right now that forces llama.cpp to be built as a shared library). I then tried to build my crate. I also installed libllama.so in my OS.

I expected to see this happen: Crate builds

Instead, this happened: Crate fails to link saying the new symbol I added to llama.cpp is missing.

The root cause is that cpal depends indirectly on alsa-sys which uses pkg-config which adds the location of the found library in /usr/lib to the search path. This happens before llama-cpp-2 runs and thus /usr/lib ends up ahead of the path that llama-cpp-2 prints causing it to try to link against the OS library that doesn't have the symbol. 

While there's no way to specify priority and while valuable, that feels like it would be a much larger discussion. I was wondering if a more targeted fix would be to ensure that all paths under OUT_DIR are moved to the front (in an order-preserving way) while the remaining directories are moved later (in an order-preserving way). That way the linker ALWAYS looks for paths within OUT_DIR before looking in other locations.

I'm not sure if this bug is in rust itself or in Cargo, so I also filed to rust: https://github.com/rust-lang/rust/issues/137404

### Steps

1. Install alsa and another dynamic library that conflicts with a crate that tries to build the same dynamic library (tested on Linux with libllama.so and llama-cpp-sys-2) 
2. Add alsa-sys as a dependency & use some symbols from it so that /usr/lib ends up getting added to the search path by pkg-config
3. Modify the crate to build a new version of the conflicting dynamic library with new symbols & try to use those symbols

Expected: build should still work because the dynamic library should be used out of OUT_DIR not /usr/lib.

### Possible Solution(s)

Bubble up link search paths under OUT_DIR to the front and bubble down search paths outside of it in a relative order preserving way.

### Notes

_No response_

### Version

```text
cargo 1.84.0 (66221abde 2024-11-19)
release: 1.84.0
commit-hash: 66221abdeca2002d318fde6efff516aab091df0e
commit-date: 2024-11-19
```

vlovich:
If a library exists both in an added folder inside OUT_DIR and in the OS, prefer to use the one within OUT_DIR. Folders within OUT_DIR and folders outside OUT_DIR do not change their relative order between themselves.

This is accomplished by sorting by whether we think the path is inside the search path or outside.

Fixes #15220. If a Rust crates builds a dynamic library & that same dynamic library is installed in the host OS, the result of the build's success & consistent behavior of executed tools depends on whether or not the user has the conflicting dynamic library in the external search path. If they do, then the host OS library will always be used which is unexpected - updates to your Rust dependency will still have you linking & running against an old host OS library (i.e. someone who doesn't have that library has a different silent behavior).

This is what I did to verify my issue got resolved but I'm sure there's a simpler example one could construct.

* Make sure Alsa and libllama.so are installed (on Arch I installed alsa-lib and llama.cpp-cuda).
* Clone llama-cpp-2 & init llama.cpp submodule & update the submodule to point to https://github.com/ggml-org/llama.cpp/pull/11997 instead.
* Add plumbing to expose the new method within llama-cpp-2 as a public facing function on the LlamaModel struct (it's basically the same code as for n_head, just calling n_head_kv from llama.cpp).
* Add cpal as a dependency in crate "foo"
* Add llama-cpp-2 via path as a dependency in crate "foo" and enable the `dynamic-link` feature.
* Add code using the newly expose n_head_kv method in crate "foo" in main.rs. NOTE: Code just needs to compile & be exported, doesn't have to be correct (fn main is probably easiest.
* Add some basic code that tries to initialize cpal in crate "foo" in fn main.
* Try to build / run crate "foo"

Before my change, it fails with a linker error saying it can't find `llama_model_n_head_kv` because /usr/lib appears in the search path before the directory that contains the libllama.so that was built internally by the crate. This is because cpal depends on alsa-sys which uses pkg-config which adds /usr/lib to the search path before the llama-cpp-sys-2 build.rs is run.

### Additional information

I'm not sure how to add tests so open to some help on that. I wanted to make sure that this approach is even correct. I coded this to change Cargo minimally and defensively since I don't know the internals of Cargo very well (e.g. I don't know if I have to compare against both `script_out_dir` / `script_out_dir_when_generated` since I don't know the difference & there's not really any explanation on what they are).

It's possible this over-complicates the implementation so open to any feedback. Additionally, the sort that happens prior to each build up of the rustc environment is not where I'd ideally place it. I think it would be more efficient to have the list of search paths be free-floating and not tied to a BuildOutput so that they could be kept updated live & resorted only on insertion (since it's changed less frequently than rustc is invoked). Additionally, the generalized sort is correct but pessimistic - maintaining the list sorted could be done efficiently with some minor book keeping (i.e. you'd only need to sort the new paths & then could quickly inject into the middle of a VecDeque).

And of course in terms of correctness, I didn't do a thorough job testing across all possible platforms. From first principles this seems directionally correct but it's always possible this breaks someone else's workflow. I'm also uneasy that the relative position of `-L` / `-l` arguments changes in this PR & I'm not sure if that's observable behavior or not (i.e. it used to be -L for a crate followed by `-l` for a crate), but now it's `-L` for all crates, still grouped by crated internally, followed by `-l` by crate).

### DIFF
--- a/src/cargo/core/compiler/custom_build.rs
+++ b/src/cargo/core/compiler/custom_build.rs
@@ -31,7 +31,7 @@
 //! [`CompileMode::RunCustomBuild`]: super::CompileMode
 //! [instructions]: https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script
 
-use super::{fingerprint, BuildRunner, Job, Unit, Work};
+use super::{fingerprint, get_dynamic_search_path, BuildRunner, Job, Unit, Work};
 use crate::core::compiler::artifact;
 use crate::core::compiler::build_runner::UnitHash;
 use crate::core::compiler::fingerprint::DirtyReason;
@@ -74,11 +74,76 @@ pub enum Severity {
 
 pub type LogMessage = (Severity, String);
 
+/// Represents a path added to the library search path.
+///
+/// We need to keep track of requests to add search paths within the cargo build directory
+/// separately from paths outside of Cargo. The reason is that we want to give precedence to linking
+/// against libraries within the Cargo build directory even if a similar library exists in the
+/// system (e.g. crate A adds `/usr/lib` to the search path and then a later build of crate B adds
+/// `target/debug/...` to satisfy its request to link against the library B that it built, but B is
+/// also found in `/usr/lib`).
+///
+/// There's some nuance here because we want to preserve relative order of paths of the same type.
+/// For example, if the build process would in declaration order emit the following linker line:
+/// ```bash
+/// -L/usr/lib -Ltarget/debug/build/crate1/libs -L/lib -Ltarget/debug/build/crate2/libs)
+/// ```
+///
+/// we want the linker to actually receive:
+/// ```bash
+/// -Ltarget/debug/build/crate1/libs -Ltarget/debug/build/crate2/libs) -L/usr/lib -L/lib
+/// ```
+///
+/// so that the library search paths within the crate artifacts directory come first but retain
+/// relative ordering while the system library paths come after while still retaining relative
+/// ordering among them; ordering is the order they are emitted within the build process,
+/// not lexicographic order.
+///
+/// WARNING: Even though this type implements PartialOrd + Ord, this is a lexicographic ordering.
+/// The linker line will require an explicit sorting algorithm. PartialOrd + Ord is derived because
+/// BuildOutput requires it but that ordering is different from the one for the linker search path,
+/// at least today. It may be worth reconsidering & perhaps it's ok if BuildOutput doesn't have
+/// a lexicographic ordering for the library_paths? I'm not sure the consequence of that.
+#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub enum LibraryPath {
+    /// The path is pointing within the output folder of the crate and takes priority over
+    /// external paths when passed to the linker.
+    CargoArtifact(PathBuf),
+    /// The path is pointing outside of the crate's build location. The linker will always
+    /// receive such paths after `CargoArtifact`.
+    External(PathBuf),
+}
+
+impl LibraryPath {
+    fn new(p: PathBuf, script_out_dir: &Path) -> Self {
+        let search_path = get_dynamic_search_path(&p);
+        if search_path.starts_with(script_out_dir) {
+            Self::CargoArtifact(p)
+        } else {
+            Self::External(p)
+        }
+    }
+
+    pub fn into_path_buf(self) -> PathBuf {
+        match self {
+            LibraryPath::CargoArtifact(p) | LibraryPath::External(p) => p,
+        }
+    }
+}
+
+impl AsRef<PathBuf> for LibraryPath {
+    fn as_ref(&self) -> &PathBuf {
+        match self {
+            LibraryPath::CargoArtifact(p) | LibraryPath::External(p) => p,
+        }
+    }
+}
+
 /// Contains the parsed output of a custom build script.
 #[derive(Clone, Debug, Hash, Default, PartialEq, Eq, PartialOrd, Ord)]
 pub struct BuildOutput {
     /// Paths to pass to rustc with the `-L` flag.
-    pub library_paths: Vec<PathBuf>,
+    pub library_paths: Vec<LibraryPath>,
     /// Names and link kinds of libraries, suitable for the `-l` flag.
     pub library_links: Vec<String>,
     /// Linker arguments suitable to be passed to `-C link-arg=<args>`
@@ -237,7 +302,7 @@ fn emit_build_output(
     let library_paths = output
         .library_paths
         .iter()
-        .map(|l| l.display().to_string())
+        .map(|l| l.as_ref().display().to_string())
         .collect::<Vec<_>>();
 
     let msg = machine_message::BuildScript {
@@ -884,10 +949,16 @@ impl BuildOutput {
                 "rustc-flags" => {
                     let (paths, links) = BuildOutput::parse_rustc_flags(&value, &whence)?;
                     library_links.extend(links.into_iter());
-                    library_paths.extend(paths.into_iter());
+                    library_paths.extend(
+                        paths
+                            .into_iter()
+                            .map(|p| LibraryPath::new(p, script_out_dir)),
+                    );
                 }
                 "rustc-link-lib" => library_links.push(value.to_string()),
-                "rustc-link-search" => library_paths.push(PathBuf::from(value)),
+                "rustc-link-search" => {
+                    library_paths.push(LibraryPath::new(PathBuf::from(value), script_out_dir))
+                }
                 "rustc-link-arg-cdylib" | "rustc-cdylib-link-arg" => {
                     if !targets.iter().any(|target| target.is_cdylib()) {
                         log_messages.push((
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -659,6 +659,13 @@ fn add_plugin_deps(
     Ok(())
 }
 
+fn get_dynamic_search_path(path: &Path) -> &Path {
+    match path.to_str().and_then(|s| s.split_once("=")) {
+        Some(("native" | "crate" | "dependency" | "framework" | "all", path)) => Path::new(path),
+        _ => path,
+    }
+}
+
 // Determine paths to add to the dynamic search path from -L entries
 //
 // Strip off prefixes like "native=" or "framework=" and filter out directories
@@ -670,12 +677,9 @@ where
 {
     let mut search_path = vec![];
     for dir in paths {
-        let dir = match dir.to_str().and_then(|s| s.split_once("=")) {
-            Some(("native" | "crate" | "dependency" | "framework" | "all", path)) => path.into(),
-            _ => dir.clone(),
-        };
+        let dir = get_dynamic_search_path(dir);
         if dir.starts_with(&root_output) {
-            search_path.push(dir);
+            search_path.push(dir.to_path_buf());
         } else {
             debug!(
                 "Not including path {} in runtime library search path because it is \
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -79,7 +79,7 @@ pub use self::compilation::{Compilation, Doctest, UnitOutput};
 pub use self::compile_kind::{CompileKind, CompileTarget};
 pub use self::crate_type::CrateType;
 pub use self::custom_build::LinkArgTarget;
-pub use self::custom_build::{BuildOutput, BuildScriptOutputs, BuildScripts};
+pub use self::custom_build::{BuildOutput, BuildScriptOutputs, BuildScripts, LibraryPath};
 pub(crate) use self::fingerprint::DirtyReason;
 pub use self::job_queue::Freshness;
 use self::job_queue::{Job, JobQueue, JobState, Work};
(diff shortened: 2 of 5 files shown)


vlovich:
Not sure what the CI failure for "check-version-bump" is but  I'm assuming it's some minor book keeping that's validating a version number somewhere needs to be bumped because I made a consequential change? But the output when running locally doesn't make sense to me:

```
--- failure function_requires_different_generic_type_params: function now requires a different number of generic type parameters ---

Description:
A function now requires a different number of generic type parameters than it used to. Uses of this function that supplied the previous number of generic types will be broken.
        ref: https://doc.rust-lang.org/reference/items/generics.html
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.39.0/src/lints/function_requires_different_generic_type_params.ron

Failed in:
  function strip_prefix_canonical (1 -> 0 generic types) in /home/vlovich/projects/cargo/crates/cargo-util/src/paths.rs:713
```

I haven't touched `strip_prefix_canonical` - running `cargo semver-checks check-release --workspace` on master seems to indicate this is broken there too so I think this step might just be broken for everyone right now.

weihanglo:
That CI job failure is completely unrelated to this PR. See <https://github.com/rust-lang/cargo/pull/15222>.

weihanglo:
Just note that it is generally recommended an issue to be discussed and labeled as `S-accepted` and than open a PR, though you are welcome to open this as a draft PR and hack with it :)

https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo

vlovich:
Good note but I also wanted to see how difficult such a change would be to help guide any questions that might come up in the issue.

src/cargo/core/compiler/custom_build.rs
```diff
@@ -884,10 +957,16 @@ impl BuildOutput {
...
+                    library_paths.extend(paths.into_iter().map(|p| {
+                        LibraryPath::new(p, script_out_dir, script_out_dir_when_generated)
+                    }));
                 }
                 "rustc-link-lib" => library_links.push(value.to_string()),
-                "rustc-link-search" => library_paths.push(PathBuf::from(value)),
+                "rustc-link-search" => library_paths.push(LibraryPath::new(
+                    PathBuf::from(value),
+                    script_out_dir,
+                    script_out_dir_when_generated,
```
weihanglo, line 968: Why do we still need `script_out_dir_when_generated` here, isn't `value` already rewritten with `script_out_dir` at line 932?

src/cargo/core/compiler/custom_build.rs
```diff
@@ -74,11 +75,83 @@ pub enum Severity {
 
 pub type LogMessage = (Severity, String);
 
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub enum LibraryPath {
```
weihanglo, line 79: This enum and variants may be worth some documentations about what they are and why we're doing this. And I think we can derive Ord here.

https://doc.rust-lang.org/stable/std/cmp/trait.Ord.html#derivable

> When derived on enums, variants are ordered primarily by their discriminants. Secondarily, they are ordered by their fields. By default, the discriminant is smallest for variants at the top, and largest for variants at the bottom. Here’s an example:

src/cargo/core/compiler/mod.rs
```diff
@@ -649,6 +667,13 @@ fn add_plugin_deps(
     Ok(())
 }
 
+fn get_dynamic_search_path(path: &PathBuf) -> &Path {
```
weihanglo, line 670: Would you mind moving this function extraction to its own commit? That would help review and trace code in the future.

(Git is pretty bad at displaying this kind of diff 😞).

weihanglo, line 670:
```suggestion
fn get_dynamic_search_path(path: &Path) -> &Path {
```

If nothing terribly needs it, we usually favor `&RefType` over `&OwnedType` for function arguments.

src/cargo/core/compiler/mod.rs
```diff
@@ -485,16 +485,34 @@ fn rustc(
...
+            let output = build_script_outputs.get(key.1).ok_or_else(|| {
+                internal(format!(
+                    "couldn't find build script output for {}/{}",
+                    key.0, key.1
+                ))
+            })?;
+            library_paths.extend(output.library_paths.iter());
+        }
+
+        library_paths.sort_by_key(|p| match p {
```
weihanglo, line 500: We've implemented (or derived) Ord, so `sort` should be good enough. No need to

```suggestion
        library_paths.sort()
```

vlovich: Unfortunately not. This does a grouped sort, keeping the relative ordering within a group the same (i.e. it moves all Cargo artifacts to the front of the search path and external paths to the end, but within cargo artifacts & within external paths the relative order of specified paths doesn't change. If you just did .sort, then it would sort by path within the group & screw up the ordering that build.rs requested.

src/cargo/core/compiler/custom_build.rs
vlovich: I was being overly defensive. It should just pass in script_out_dir.

src/cargo/core/compiler/mod.rs
weihanglo: You're correct. I somehow thought it was already handled as I saw the manual `Ord` impl.

vlovich:
I think I've resolved all the outstanding review requests. Please let me know if there's any other concerns. I did add a comment to the LibraryPath documentation that may be worth discussing (but also OK just leaving this for future consideration):

```
/// WARNING: Even though this type implements PartialOrd + Ord, this is a lexicographic ordering.
/// The linker line will require an explicit sorting algorithm. PartialOrd + Ord is derived because
/// BuildOutput requires it but that ordering is different from the one for the linker search path,
/// at least today. It may be worth reconsidering & perhaps it's ok if BuildOutput doesn't have
/// a lexicographic ordering for the library_paths? I'm not sure the consequence of that.
```

@rustbot review

src/cargo/core/compiler/custom_build.rs
```diff
@@ -74,11 +75,96 @@ pub enum Severity {
...
+    }
+
+    pub fn display(&self) -> std::path::Display<'_> {
+        match self {
+            LibraryPath::CargoArtifact(p) | LibraryPath::External(p) => p.display(),
+        }
+    }
+}
+
+impl AsRef<Path> for LibraryPath {
```
weihanglo, line 139: Feel like we only need `AsRef<PathBuf>`?

```diff
diff --git a/src/cargo/core/compiler/custom_build.rs b/src/cargo/core/compiler/custom_build.rs
index e1e7cc49977..75d553c1db3 100644
--- a/src/cargo/core/compiler/custom_build.rs
+++ b/src/cargo/core/compiler/custom_build.rs
@@ -128,20 +128,6 @@ impl LibraryPath {
             LibraryPath::CargoArtifact(p) | LibraryPath::External(p) => p,
         }
     }
-
-    pub fn display(&self) -> std::path::Display<'_> {
-        match self {
-            LibraryPath::CargoArtifact(p) | LibraryPath::External(p) => p.display(),
-        }
-    }
-}
-
-impl AsRef<Path> for LibraryPath {
-    fn as_ref(&self) -> &Path {
-        match self {
-            LibraryPath::CargoArtifact(p) | LibraryPath::External(p) => p,
-        }
-    }
 }
 
 impl AsRef<PathBuf> for LibraryPath {
@@ -152,14 +138,6 @@ impl AsRef<PathBuf> for LibraryPath {
     }
 }
 
-impl AsRef<OsStr> for LibraryPath {
-    fn as_ref(&self) -> &OsStr {
-        match self {
-            LibraryPath::CargoArtifact(p) | LibraryPath::External(p) => p.as_os_str(),
-        }
-    }
-}
-
 /// Contains the parsed output of a custom build script.
 #[derive(Clone, Debug, Hash, Default, PartialEq, Eq, PartialOrd, Ord)]
 pub struct BuildOutput {
@@ -323,7 +301,7 @@ fn emit_build_output(
     let library_paths = output
         .library_paths
         .iter()
-        .map(|l| l.display().to_string())
+        .map(|l| l.as_ref().display().to_string())
         .collect::<Vec<_>>();
 
     let msg = machine_message::BuildScript {
diff --git a/src/cargo/core/compiler/mod.rs b/src/cargo/core/compiler/mod.rs
index 73597978880..74f964c723d 100644
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -518,7 +518,7 @@ fn rustc(
         });
 
         for path in library_paths.iter() {
-            rustc.arg("-L").arg(path);
+            rustc.arg("-L").arg(path.as_ref());
         }
 
         for key in build_scripts.to_link.iter() {

```

tests/testsuite/build_script.rs
```diff
@@ -6087,3 +6087,40 @@ fn directory_with_leading_underscore() {
         .with_status(0)
         .run();
 }
+
+#[cargo_test]
+fn linker_search_path_preference() {
```
weihanglo, line 6092: This looks nice. Thank you!

Could we have one more test that: Has some dependencies, and those dependencies also emit `cargo::rustc-link-search`, and we ensure that the order from the dependencies is maintained?

src/cargo/core/compiler/custom_build.rs
vlovich: Resolved

src/cargo/core/compiler/custom_build.rs
```diff
@@ -74,11 +75,96 @@ pub enum Severity {
 
 pub type LogMessage = (Severity, String);
 
+/// This represents a path added to the library search path. We need to keep track of requests
+/// to add search paths within the cargo build directory separately from paths outside of Cargo.
```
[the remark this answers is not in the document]
vlovich: Resolved

src/cargo/core/compiler/mod.rs
```diff
@@ -670,10 +700,7 @@ where
 {
     let mut search_path = vec![];
     for dir in paths {
-        let dir = match dir.to_str().and_then(|s| s.split_once("=")) {
-            Some(("native" | "crate" | "dependency" | "framework" | "all", path)) => path.into(),
-            _ => dir.clone(),
-        };
+        let dir = get_dynamic_search_path(dir).to_path_buf();
```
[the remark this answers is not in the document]
vlovich: Resolved

src/cargo/core/compiler/custom_build.rs
```diff
@@ -74,11 +75,96 @@ pub enum Severity {
...
+    fn new(p: PathBuf, script_out_dir: &Path) -> Self {
+        let search_path = get_dynamic_search_path(&p);
+        if search_path.starts_with(script_out_dir) {
+            Self::CargoArtifact(p)
+        } else {
+            Self::External(p)
+        }
+    }
+
+    pub fn into_path_buf(self) -> std::path::PathBuf {
```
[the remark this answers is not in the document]
vlovich: Resolved

tests/testsuite/build_script.rs
vlovich: I just added dependencies into the test rather than creating a separate test.

weihanglo (approved):
Thank you for working with us! I really love how you documented the rationale and how the code is organized.

This looks great. Going to merge it :)

PR #15223, opened 2025-02-22
chore: semver-check build-rs against beta channel

weihanglo:

This should have been restored when 1.84 was out,
but nobody remembered.


ehuss (approved):
Thanks!

PR #15166, opened 2025-02-09
Update all dependencies

ehuss:
It's been a long while since dependencies were updated, so let's go ahead and update those. Updates are:

    Updating allocator-api2 v0.2.18 -> v0.2.21
    Updating anstream v0.6.15 -> v0.6.18
    Updating anstyle v1.0.8 -> v1.0.10
    Updating anstyle-lossy v1.1.2 -> v1.1.3
    Updating anstyle-parse v0.2.5 -> v0.2.6
    Updating anstyle-query v1.1.1 -> v1.1.2
    Updating anstyle-svg v0.1.4 -> v0.1.7
    Updating anstyle-wincon v3.0.4 -> v3.0.7
    Updating anyhow v1.0.86 -> v1.0.95
    Updating arrayref v0.3.7 -> v0.3.9
    Updating arrayvec v0.7.4 -> v0.7.6
    Updating autocfg v1.3.0 -> v1.4.0
    Updating bit-set v0.5.3 -> v0.8.0
    Updating bit-vec v0.6.3 -> v0.8.0
    Removing bitflags v1.3.2
    Removing bitflags v2.6.0
      Adding bitflags v2.8.0
    Updating blake3 v1.5.2 -> v1.5.5
    Updating bstr v1.9.1 -> v1.11.3
    Updating bumpalo v3.16.0 -> v3.17.0
      Adding byteorder v1.5.0
    Updating bytes v1.6.0 -> v1.10.0
    Updating camino v1.1.6 -> v1.1.9
    Updating cargo-platform v0.1.8 -> v0.1.9
    Updating cargo_metadata v0.19.0 -> v0.19.1
    Updating cc v1.1.22 -> v1.2.13
    Updating clap v4.5.20 -> v4.5.28
    Updating clap_builder v4.5.20 -> v4.5.27
    Updating clap_lex v0.7.2 -> v0.7.4
    Updating clru v0.6.1 -> v0.6.2
    Updating color-print v0.3.6 -> v0.3.7
    Updating color-print-proc-macro v0.3.6 -> v0.3.7
    Updating colorchoice v1.0.2 -> v1.0.3
    Updating constant_time_eq v0.3.0 -> v0.3.1
    Updating cpufeatures v0.2.12 -> v0.2.17
    Updating crc32fast v1.4.0 -> v1.4.2
    Updating crossbeam-channel v0.5.12 -> v0.5.14
    Updating crossbeam-deque v0.8.5 -> v0.8.6
[...]

PR #15224, opened 2025-02-23
chore: depend on openssl-sys to correctly pin its version

weihanglo:

From https://github.com/rust-lang/cargo/issues/13546#issuecomment-2676629906

> This was an overlook in <https://github.com/rust-lang/cargo/pull/15166> and apparently we forgot to actually add openssl-sys to `cargo` package. The pinned version is [under `workspace.dependencies`](https://github.com/rust-lang/cargo/pull/15166) so it never really affect the dependency resolution.
>
> The bad news is that <https://github.com/rust-lang/cargo/pull/15166> is now in beta toolchain already. If we're going to fix this again we might need to revert it in beta as well.

This PR fixes it. However, I think we should find a way to unpin the version.

If we're going to merge this, we might want to backport to beta

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -351,6 +351,7 @@ dependencies = [
  "memchr",
  "opener",
  "openssl",
+ "openssl-sys",
  "os_info",
  "pasetors",
  "pathdiff",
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!

Approved for beta, too. Can you send a PR for that?

weihanglo:
Thanks for the review!

Beta backport PR is here: <https://github.com/rust-lang/cargo/pull/15226>.

PR #15226, opened 2025-02-23
[beta-1.86] depend on openssl-sys to correctly pin its version

weihanglo:
Beta backports

* #15224

In order to make CI pass, the following PRs are also cherry-picked:

* https://github.com/rust-lang/cargo/commit/7e0da417c071568f51bc5463f46981ce5b87112d

PR #15227, opened 2025-02-23
[beta 1.86] test(gc): Update tests for latest stable

ehuss:
This fixes the `global_cache_tracker::compatible_with_older_cargo` test to work now that stable has been updated. Cherry-picks the fix from https://github.com/rust-lang/cargo/pull/15211


ehuss:
Oh, whoops, didn't see this was picked up in https://github.com/rust-lang/cargo/pull/15226.

PR #15233, opened 2025-02-25
refactor(tree): Abstract the concept of an edge

epage:

I'm playing around with `--depth public` to help view and communicate public/private dependency test states and ended up doing these refactors along the way and felt they were generally useful independent of my `--depth public` work.

### DIFF
--- a/src/cargo/ops/tree/graph.rs
+++ b/src/cargo/ops/tree/graph.rs
@@ -456,6 +456,10 @@ fn add_pkg(
                 requested_kind,
                 opts,
             );
+            let new_edge = Edge {
+                kind: EdgeKind::Dep(dep.kind()),
+                node: dep_index,
+            };
             if opts.graph_features {
                 // Add the dependency node with feature nodes in-between.
                 dep_name_map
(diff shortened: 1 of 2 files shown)


src/cargo/ops/tree/graph.rs
```diff
@@ -579,7 +562,11 @@ fn add_cli_features(
     for fv in to_add {
         match fv {
             FeatureValue::Feature(feature) => {
-                let index = add_feature(graph, feature, None, package_index, EdgeKind::Feature).1;
+                let feature_edge = Edge {
+                    kind: EdgeKind::Feature,
+                    node: package_index,
+                };
```
weihanglo, line 568: Should this be moved out from the for loop, along with line 599?

weihanglo (commented):
Thanks. This looks pretty nice!

src/cargo/ops/tree/graph.rs
epage: I considered it but the weird thing is that on 599 we also have a `dep_edge` and I worried about pulling the definition of `feature_edge` away from that context.

weihanglo: That is true. Too far away. Thanks!

weihanglo (approved):

Issue #15151, opened 2025-02-07
publishing a crate that depends on itself as a dev dependency


mockersf:
### Problem

It used to be possible to publish a crate that depends upon itself as a dev dependency:

```toml
[package]
name = "my_crate"
version = "0.1.1"
edition = "2021"

[dependencies]

[dev-dependencies]
my_crate = { path = ".", version = "0.1.1", features = ["feature-for-test"] }

[features]
default = []
feature-for-test = []
```

This pattern is useful when you have a feature that you want to enable for tests

This worked on January 3rd, see the dev dependencies of https://crates.io/crates/bevy_math/0.15.1/dependencies published with cargo 1.83

This failed today with cargo 1.84.1

```
    Updating crates.io index
   Packaging bevy_math v0.15.2 (crates/bevy_math)
    Updating crates.io index
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for the requirement `bevy_math = "^0.15.2"`
  candidate versions found which didn't match: 0.15.1, 0.15.0, 0.15.0-rc.3, ...
  location searched: crates.io index
  required by package `bevy_math v0.15.2 (crates/bevy_math)`
```

### Steps

1. have a crate that depends on itself as a dev dependency
2. publish it

### Version

```text
cargo 1.84.1 (66221abde 2024-11-19)
release: 1.84.1
commit-hash: 66221abdeca2002d318fde6efff516aab091df0e
commit-date: 2024-11-19
host: aarch64-apple-darwin
libgit2: 1.8.1 (sys:0.19.0 vendored)
libcurl: 8.7.1 (sys:0.4.74+curl-8.9.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Mac OS 15.1.0 [64-bit]
```

epage:
In 1.84, we always include a lockfile, rather than trying to detect when it would be relevant like we did before. When we create a `.crate`, we strip `path`s from dependencies. So when we create the lockfile, we then look to the registry for this package but its not there, giving this error.

Before, were you using `--no-verify`? I assume the verify step would create a lockfile which would have failed.

epage:
Another case affected by lockfiles is #15059

mockersf:
> Before, were you using `--no-verify`? I assume the verify step would create a lockfile which would have failed.

Yup, forgot about that but our publish script use `cargo publish --no-verify --allow-dirty`

[...]

Issue #4242, opened 2017-07-01
Allow crates to be published with cyclic dev-dependencies


Marwes:
In gluon I have the project split up into multiple crates with the intent that users can include only the crates they need if they desire a smaller binary footprint. For convenience I provide a main `gluon` crate which provides an easy to use interface when one uses all the crates.

While this split works fine for mostly everything I want to use the easy interface when writing documentation tests so I added the `gluon` crate as a [dev-dependency in gluon_vm](https://github.com/gluon-lang/gluon/blob/69175032121269b83f60bb586b7a6a38a5c51936/vm/Cargo.toml#L38)  and this compiles without issue since a little bit back. It does however break when trying to publish `gluon_vm` as the `gluon` dev-dependency has not been published yet!

Currently I managed to work around it by specifying a version range but it would be nice if there was a way to publish a group of crates as a whole or to ignore the version check if a dev-dependency does not exist on crates.io (yet).

```
gluon = { path = "..", version = "<0.6.0, >=0.4.2" } # GLUON
```

alexcrichton:
Yeah in general I think we should basically ignore dev-dependencies when we publish. We shouldn't attempt to resolve them, verify they're already published, etc.

BusyJay:
Maybe dev-dependencies don't have to be shown in the crate's page too.

matthewkmayer:
I bumped into this issue as well. From what I can tell the workaround above gets us by this for now. 👍

cramertj:
+1 to this-- it now affects the futures crate.

gnzlbg:
This affects a lot of crates, what do we need to do to fix this. Isn't there a check somewhere that checks for the dependencies before publishing, where we can just filter dev-dependencies away and call it a day? cc @matklad 

I mean, this would even affect core and std, where std could depend on core to build, but core would depend on std for tests. The check is spurious anyways, because this sort of dependencies are not circular: build core -> build std -> build core tests is ok. Tests and "dev-only" artifacts are their own crates (core+tests is a different crate than core, and so are every test/foo.rs in core, and core/examples, etc.).

matklad:
>The only solution I came up with is the same that gluon uses which is insane: publish coresimd prunning dev-dependencies, publish stdsimd, re-publish coresimd with a minor version update to use the newly publish stdsimd as dev-dependencies.

[...]

PR #15234, opened 2025-02-25
feat(package): add --exclude-lockfile flag

### LINKED ISSUE #15059: `cargo package --no-verify` fails if a package's version req is too high for the registry (but works locally)
### Problem

#13447 dd698ff0483bda065d1f78968e70ee86fb116a72 has some unintentional consequences:

`cargo package` after that generates a lockfile also when packaging libraries.
However, if the crate being packaged depends on some unpublished dependencies. When generating the lockfile Cargo then can't find unpublished deps on registry index, so it bailed out.

### Steps

This could be simplly reproduced in rust-lang/cargo as of today (2025-01-13)

1. Checkout to 54df3c7dd914a529b4efda61b1efcc10dbde52cc
2. Bump cargo-test-support to an unpublished version such as `99.99.99`
3. `cargo +1.83.0 package -p cargo-test-support --no-verify` — succeeded
4. `cargo +1.84.0 package -p cargo-test-support --no-verify` — failed

### Possible Solution(s)

Assume if we have passed the local registry unconditionally, we shouldn't have the issue today.

https://github.com/rust-lang/cargo/blob/f15df8fd4990f1b87734343c4e44f3adec3c8b00/src/cargo/ops/cargo_package/mod.rs?plain=1#L219-L225

### Version

* `cargo 1.84.0 (66221abde 2024-11-19)`
* `cargo 1.86.0-nightly (088d49608 2025-01-10)`

### LINKED ISSUE #15159: Provide a flag for `package` to avoid regenerating/generating Cargo.lock, thus decoupling packaging from the registry
### Problem

I would like to work on adding a new flag to `cargo package` to avoid regenerating Cargo.lock when packaging, or to avoid generating it at all in case there is none.

The main use-case I have in mind for this flag is the packaging work we do in Debian. In Debian we adopt two main approaches when it comes to rust packaging. The first is packaging single crates, with source code obtained from crates.io. The second is packaging from upstream sources (usually git repos), which is particularly useful for multi-crate projects, cross-language projects, or projects that simply don't publish to crates.io. Now, the latter may contain crates with rust libraries, and these rust libraries must be packaged as `librust-$CRATENAME-dev` Debian packages in order for them to be used by other packages as build dependencies. When it comes to this, `cargo package` (used either as a command line tool or via the cargo library) is extremely useful in that it creates a canonicalized version of the crate, with the desired included/excluded files, which can then be simply copied to the desired directory and packaged as a .deb. Before this is done (regardless of whether one uses `cargo package` or not), Cargo.lock is removed from the directory because there is no use for it in Debian (verification is done at the level of the Debian archive -- that is, by verifying .debs themselves -- and versioning is locked to whatever we have in the archive at a certain point in time).

Unfortunately, the way `cargo package` currently works -- especially after 1.84 which started regenerating/generating lock files for *all* crates -- couples the packaging process to the registry in a way that makes it quite hard for us to use it in practice the way we would like. Specifically, if Cargo.lock needs to be regenerated/generated, then *all* the crates' dependencies must be in the registry in order to use `cargo package`. This causes some notable issues:

1. it forces us to deal with ordering. Namely, we need to install packages in the local registry in the correct order, so that they can be found by packages that depend on them. At this time I've only seen evidence of ordering issues between crates, but my guess is this will extend to ordering issues between crate *features*, which will be much harder to solve

2. due to the above, it forces us to deal with circular dependencies. An example of this was already reported for cargo publish in #15151, pointing to the same underlying issue I'm reporting here (but possibly needing a different solution). I have a strong feeling that the issue of circular dependencies will not be solved easily

3. in general, it prevents us from packaging crates unless all dependencies are installed in the local registry (this was also reported in #15059). This is not a logic necessity since, were it not for Cargo.lock, canonicalization and file inclusion/exclusion could proceed with no input from the registry. Packaging crates without dependencies being present locally saves us from downloading and installing dependencies unless they are really needed

4. since not only dependencies, but also dev-dependencies must be installed in the local registry, test/example/benchmark dependencies turn into actual build-dependencies. This in itself is a separate issue than circular dependencies, but can ultimately cause (cross-workspace) circular build-dependencies (see e.g. the scenario in [my comment below](https://github.com/rust-lang/cargo/issues/15159#issuecomment-2667243259)).

### Proposed Solution

Since as I said we don't care about Cargo.lock in Debian (again, we actually remove it from the final package), all of this would be solved for us if one could just skip lock file regeneration/generation. I've already tested that, when used together with `--no-verify`, a new `--no-gen-lockfile` flag that implements such skipping would decouple `cargo package` from the registry: in practice one can purge the registry from all crates and use `cargo package` with no dependencies present locally at all. This way the issue of dependency ordering/cyles obviously never arises.

A proof of concept is at https://github.com/NoisyCoil/cargo/tree/no-lockfile, but I would like to hear your thoughts before submitting an actual PR.

### Notes

I am not sure this proposal is suitable for solving #15151, since `publish` should probably have different requirements than `package`. Indeed, in my proof of concept, the `no-gen-lockfile` flag is turned off for the (internal) use that `publish` makes of `package`. On the other hand, it may help (or even completely solve) #15059.

weihanglo:

Fixes #15059
Fixes #15159

This provides an escape hatch `--exclude-lockfile`for uncommon workflows
that don't verify (`--no-verify` is passed) the build with their unpublished packages
In effect, this takes the heuristic removed in #14815 and replaces it with a flag

When `--exclude-lockfile` is enabled,
`cargo package` will not verify the lock file if present,
nor will it generate a new one if absent.
Cargo.lock will not be included in the resulting tarball.

Together with `--no-verify`,
this flag decouples packaging from checking the registry index.
While this is useful for some non-normal workflows that requires
to assemble packages having unpublished dependencies.
It is recommended to use `-Zpackage-workspace` to package the entire
workspace, instead of opting out lockfile.

The first commit was stolen from <https://github.com/NoisyCoil/cargo/commit/1a104b544445b11044f095705ed6dd45f8a60b15> (credit to @NoisyCoil!)

The second added two failing cases we observed in #15059.

### DIFF
--- a/src/doc/man/cargo-package.md
+++ b/src/doc/man/cargo-package.md
@@ -27,8 +27,8 @@ stored in the `target/package` directory. This performs the following steps:
     - `[patch]`, `[replace]`, and `[workspace]` sections are removed from the
       manifest.
     - `Cargo.lock` is always included. When missing, a new lock file will be
-      generated. {{man "cargo-install" 1}} will use the packaged lock file if
-      the `--locked` flag is used.
+      generated unless the `--exclude-lockfile` flag is used. {{man "cargo-install" 1}}
+      will use the packaged lock file if the `--locked` flag is used.
     - A `.cargo_vcs_info.json` file is included that contains information
       about the current VCS checkout hash if available, as well as a flag if the
       worktree is dirty.
@@ -96,6 +96,14 @@ or the license).
 
 {{#option "`--allow-dirty`" }}
 Allow working directories with uncommitted VCS changes to be packaged.
+{{/option}}
+
+{{#option "`--exclude-lockfile`" }}
+Don't include the lock file when packaging.
+
+This flag is not for general use.
+Some tools may expect a lock file to be present (e.g. `cargo install --locked`).
+Consider other options before using this.
 {{/option}}
 
 {{> options-index }}
--- a/tests/testsuite/package.rs
+++ b/tests/testsuite/package.rs
@@ -7304,9 +7304,8 @@ fn exclude_lockfile() {
         .file("src/lib.rs", "")
         .build();
 
-    p.cargo("package --list")
+    p.cargo("package --list --exclude-lockfile")
         .with_stdout_data(str![[r#"
-Cargo.lock
 Cargo.toml
 Cargo.toml.orig
 src/lib.rs
@@ -7315,10 +7314,10 @@ src/lib.rs
         .with_stderr_data("")
         .run();
 
-    p.cargo("package")
+    p.cargo("package --exclude-lockfile")
         .with_stderr_data(str![[r#"
 [PACKAGING] foo v0.0.1 ([ROOT]/foo)
-[PACKAGED] 4 files, [FILE_SIZE]B ([FILE_SIZE]B compressed)
+[PACKAGED] 3 files, [FILE_SIZE]B ([FILE_SIZE]B compressed)
 [VERIFYING] foo v0.0.1 ([ROOT]/foo)
 [COMPILING] foo v0.0.1 ([ROOT]/foo/target/package/foo-0.0.1)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
@@ -7330,7 +7329,7 @@ src/lib.rs
     validate_crate_contents(
         f,
         "foo-0.0.1.crate",
-        &["Cargo.lock", "Cargo.toml", "Cargo.toml.orig", "src/lib.rs"],
+        &["Cargo.toml", "Cargo.toml.orig", "src/lib.rs"],
         (),
     );
 }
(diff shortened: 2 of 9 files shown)


epage:
Mind cleaning up the commit history (reorder test first, squash test update), including updating commint message for new flag name?

weihanglo:
> Mind cleaning up the commit history (reorder test first, squash test update), including updating commint message for new flag name?

Was in a haste leaving the office. Sorry.

src/doc/man/cargo-package.md
```diff
@@ -98,6 +98,12 @@ or the license).
 Allow working directories with uncommitted VCS changes to be packaged.
 {{/option}}
 
+{{#option "`--exclude-lockfile`" }}
+Don't include the lock file when packaging.
+
+This flag is not for general use. Always consider packaging with a lockfile.
```
weihanglo, line 104: Not sure if we really want to mention this. People using `cargo package` usually are already in an unusual workflow.

NoisyCoil:
Thanks for stealing, very much appreciated! (Especially the part where I don't have to come up with a better name for the flag ;-))

src/doc/man/cargo-package.md
epage: I think it might be worth giving a caution on this though we should probably give some semblance of a reason

> This flag is not for general use as some tools may expect a lockfile to be present (e.g. `cargo install --locked`).
> Consider other options before using this.

`cargo-semver-checks` is also something else that would benefit from the lockfile but that feels a bit odd to call out.

tests/testsuite/package.rs
epage, line 1: Should we also add a test that doesn't start in a failing case?

weihanglo: Good idea. Just force push a new revision.

src/doc/man/cargo-package.md
weihanglo: Adopted your suggestion with just one slight tweak. Thanks!

epage:
@fcp merge

This is intended for corner cases where existing or upcoming functionality does not meet people's needs.

epage (approved):

weihanglo:
@rfcbot fcp merge


See the PR description for what is proposed to merge.

People may want to bikeshed on the flag name, though.

Nemo157:
Given this is only on `cargo package`, how are you meant to publish the result?

weihanglo:
> Given this is only on `cargo package`, how are you meant to publish the result?

@Nemo157 
It is not meant to "publish". Quoting from the PR description:

> This provides an escape hatch `--exclude-lockfile` for uncommon workflows
that don't verify (`--no-verify` is passed) the build with their unpublished packages
In effect, this takes the heuristic removed in https://github.com/rust-lang/cargo/pull/14815 and replaces it with a flag

It is for you just want a tarball but some unpublished crates failed `cargo package` due to lockfile generation. The two linked issues have more use cases in details.

Nemo157:
Hmm, so my usecase of publishing circular dependencies (not actually circular in the unit-graph, just the simplified package-graph for resolution) that will actually resolve once the final crate is published is just no longer supported? (This used to work fine with `--no-verify`).

EDIT: ok, I found the mostly relevant issue about this regressing #15151, I'll just have to abandon having usable tests in published crates until #4242 is ever actually implemented.

PR #15236, opened 2025-02-26
Added `build.build_dir` templating support

ranger-ross:

This PR is a follow up on #15104 and and adds support for the path templating in `build.build-dir` as defined in #14125.

Supported templates:
* `{workspace-root}`
* `{cargo-cache}` (pointing to `CARGO_HOME` for now)
* `{workspace-manifest-path-hash}`

#### Unresolved questions

What should we name `{workspace-manifest-path-hash}` and what should it include?  Should we shorten to `{workspace-hash}` or even just `{hash}`?  Should we include the Cargo version so we get unique whole-target directories for easier cleanup (#13136)

How should this handle unknown variables (error) or unclosed `{` / `}` (ignored), see https://github.com/rust-lang/cargo/pull/15236#discussion_r1977898973

When using `{workspace-manifest-path-hash}` this hash will change based on the project path. In the event of a cargo being executed in a symlinked project, the hash will change. 

For example, given the following directory 
```
/Users/
└─ user1/
    └─ projects/
        ├─ actual-crate/
        │  └─ Cargo.toml
        └─ symlink-to-crate/ -> actual-crate/
```

the hash will be unique when running cargo from each of the following directories.
* `/Users/user1/actual-crate`
* `/Users/user1/symlink-to-crate`

Figuring out whether to handle this is deferred out, see
- https://github.com/rust-lang/cargo/pull/15236#discussion_r1971938021
-  https://github.com/poliorcetics/rfcs/blob/cargo-target-directories/text/3371-cargo-target-dir-templates.md#symbolic-links
- https://github.com/rust-lang/cargo/issues/12207#issuecomment-2711402159

This PR is fairly small. I included tests for each template variable.

You can also clone my branch and test it locally with
```console
CARGO_BUILD_BUILD_DIR='{workspace-root}/foo' cargo -Z build-dir build
```

### Additional information

While searching Cargo for any prior art for path templating, I found [`sources/registry/download.rs`](https://github.com/rust-lang/cargo/blob/master/src/cargo/sources/registry/download.rs#L84) doing a simple string replace. Thus I followed the same behavior. 

r? @epage

### DIFF
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -652,13 +652,36 @@ impl GlobalContext {
     /// Falls back to the target directory if not specified.
     ///
     /// Callers should prefer [`Workspace::build_dir`] instead.
-    pub fn build_dir(&self) -> CargoResult<Option<Filesystem>> {
+    pub fn build_dir(&self, workspace_manifest_path: &PathBuf) -> CargoResult<Option<Filesystem>> {
         if !self.cli_unstable().build_dir {
             return self.target_dir();
         }
         if let Some(val) = &self.build_config()?.build_dir {
-            let path = val.resolve_path(self);
+            let replacements = vec![
+                (
+                    "{workspace-root}",
+                    workspace_manifest_path
+                        .parent()
+                        .unwrap()
+                        .to_str()
+                        .context("workspace root was not valid utf-8")?
+                        .to_string(),
+                ),
+                (
+                    "{cargo-cache-home}",
+                    self.home()
+                        .as_path_unlocked()
+                        .to_str()
+                        .context("cargo home was not valid utf-8")?
+                        .to_string(),
+                ),
+                ("{workspace-manifest-path-hash}", {
+                    let hash = crate::util::hex::short_hash(&workspace_manifest_path);
+                    format!("{}{}{}", &hash[0..2], std::path::MAIN_SEPARATOR, &hash[2..])
+                }),
+            ];
 
+            let path = val.resolve_templated_path(self, replacements);
             // Check if the target directory is set to an empty string in the config.toml file.
             if val.raw_value().is_empty() {
                 bail!(
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -681,7 +681,17 @@ impl GlobalContext {
                 }),
             ];
 
-            let path = val.resolve_templated_path(self, replacements);
+            let path = val
+                .resolve_templated_path(self, replacements)
+                .map_err(|e| match e {
+                    path::ResolveTemplateError::UnexpectedVariable {
+                        variable,
+                        raw_template,
+                    } => anyhow!(
+                        "unexpected variable `{variable}` in build.build-dir path `{raw_template}`"
+                    ),
+                })?;
+
             // Check if the target directory is set to an empty string in the config.toml file.
             if val.raw_value().is_empty() {
                 bail!(
--- a/src/cargo/util/context/path.rs
+++ b/src/cargo/util/context/path.rs
@@ -32,6 +32,25 @@ impl ConfigRelativePath {
         self.0.definition.root(gctx).join(&self.0.val)
     }
 
+    /// Same as [`Self::resolve_path`] but will make string replacements
+    /// before resolving the path.
+    ///
+    /// `replacements` should be an an [`IntoIterator`] of tuples with the "from" and "to" for the
+    /// string replacement
+    pub fn resolve_templated_path(
+        &self,
+        gctx: &GlobalContext,
+        replacements: impl IntoIterator<Item = (impl AsRef<str>, impl AsRef<str>)>,
+    ) -> PathBuf {
+        let mut value = self.0.val.clone();
+
+        for (from, to) in replacements {
+            value = value.replace(from.as_ref(), to.as_ref());
+        }
+
+        self.0.definition.root(gctx).join(&value)
+    }
+
     /// Resolves this configuration-relative path to either an absolute path or
     /// something appropriate to execute from `PATH`.
     ///
--- a/src/cargo/util/context/path.rs
+++ b/src/cargo/util/context/path.rs
@@ -1,4 +1,5 @@
 use super::{GlobalContext, StringList, Value};
+use regex::Regex;
 use serde::{de::Error, Deserialize};
 use std::path::PathBuf;
 
@@ -41,14 +42,23 @@ impl ConfigRelativePath {
         &self,
         gctx: &GlobalContext,
         replacements: impl IntoIterator<Item = (impl AsRef<str>, impl AsRef<str>)>,
-    ) -> PathBuf {
+    ) -> Result<PathBuf, ResolveTemplateError> {
         let mut value = self.0.val.clone();
 
         for (from, to) in replacements {
             value = value.replace(from.as_ref(), to.as_ref());
         }
 
-        self.0.definition.root(gctx).join(&value)
+        // Check for expected variables
+        let re = Regex::new(r"\{(.*)\}").unwrap();
+        if let Some(caps) = re.captures(&value) {
+            return Err(ResolveTemplateError::UnexpectedVariable {
+                variable: caps[1].to_string(),
+                raw_template: self.0.val.clone(),
+            });
+        };
+
+        Ok(self.0.definition.root(gctx).join(&value))
     }
 
     /// Resolves this configuration-relative path to either an absolute path or
@@ -121,4 +131,12 @@ impl PathAndArgs {
             args,
         }
     }
+}
+
+#[derive(Debug)]
+pub enum ResolveTemplateError {
+    UnexpectedVariable {
+        variable: String,
+        raw_template: String,
+    },
 }
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -260,6 +260,13 @@ build-dir = "out"
 
 The path to where internal files used as part of the build are placed.
 
+This option supports path templating.
+
+Avaiable template variables:
+* `{workspace-root}` resolves to root of the current workspace.
+* `{cargo-cache-home}` resolves to `CARGO_HOME`
+* `{workspace-manifest-path-hash}` resolves to a hash of the manifest path
+
 
 ## root-dir
 * Original Issue: [#9887](https://github.com/rust-lang/cargo/issues/9887)
--- a/tests/testsuite/build_dir.rs
+++ b/tests/testsuite/build_dir.rs
@@ -12,7 +12,7 @@
 use std::path::PathBuf;
 
 use cargo_test_support::prelude::*;
-use cargo_test_support::project;
+use cargo_test_support::{paths, project};
 use std::env::consts::{DLL_PREFIX, DLL_SUFFIX, EXE_SUFFIX};
 
 #[cargo_test]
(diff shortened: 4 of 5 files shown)


tests/testsuite/build_dir.rs
```diff
@@ -491,6 +491,119 @@ fn future_incompat_should_output_to_build_dir() {
     assert_exists(&p.root().join("build-dir/.future-incompat-report.json"));
 }
 
+mod templating {
```
epage, line 494: nit: We've generally not been subdividing test mods up like this

src/cargo/util/context/mod.rs
```diff
@@ -636,13 +636,36 @@ impl GlobalContext {
...
+                    self.home()
+                        .as_path_unlocked()
+                        .to_str()
+                        .context("cargo home was not valid utf-8")?
+                        .to_string(),
+                ),
+                (
+                    "{workspace-manifest-path-hash}",
+                    crate::util::hex::short_hash(manifest_path),
+                ),
```
epage, line 665: See https://github.com/poliorcetics/rfcs/blob/cargo-target-directories/text/3371-cargo-target-dir-templates.md#handle-possibly-thousands-of-directories-in-a-single-templated-cargo_target_dir-path

src/cargo/util/context/mod.rs
```diff
@@ -636,13 +636,36 @@ impl GlobalContext {
...
+                        .to_string(),
+                ),
+                (
+                    "{cargo-cache}",
+                    self.home()
+                        .as_path_unlocked()
+                        .to_str()
+                        .context("cargo home was not valid utf-8")?
+                        .to_string(),
+                ),
```
epage, line 661: The documentation needs to be updated with these variables

src/cargo/util/context/mod.rs
epage: (the logic for this lives in `default_target_dir`)

src/cargo/util/context/mod.rs
```diff
@@ -636,13 +636,36 @@ impl GlobalContext {
...
+                (
+                    "{workspace-manifest-path-hash}",
+                    crate::util::hex::short_hash(manifest_path),
+                ),
+            ];
+
+            let path = val.resolve_templated_path(self, replacements);
```
epage, line 668: A part of me feels like this should be handled in the caller.

e.g. `default_target_dir` depends on more context, like whether the manifest is embedded or not.

src/cargo/util/context/mod.rs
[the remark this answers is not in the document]
epage: Hmm, or maybe we only expose `{cargo-cache}=CARGO_HOME` and have people add `./build/`, like they have to do with `workspace-root`

src/cargo/util/context/mod.rs
ranger-ross: oh thanks I was not aware of this. I had just assumed a short sha. 
I'll update to follow a similar impl as `default_target_dir`

src/cargo/util/context/mod.rs
ranger-ross: I think this is fair. 
My rational for putting it here was to resolve the templates as early as possible to avoid the potential for bugs being introduced by using the unresolved path.

However, after pointing out `default_target_dir` I think it makes sense for template resolution for both to happen in the same "layer".

I'll move this logic to `workspace.rs`

src/cargo/util/context/mod.rs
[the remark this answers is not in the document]
ranger-ross: Naming it a bit difficult here :sweat: 

`{cargo-build-dir}` feels a bit odd since its being to define `build-dir`. (it seems recursive)
```toml
build-dir = "{cargo-build-dir}/foo"
```
---

Perhaps `{cargo-build-cache}` pointing to `CARGO_HOME/build` might be good.
I am imagining that eventually `build.build-dir` defaults to `{cargo-build-cache}` and we just consider `CARGO_HOME/build` the "build cache". 

What do you think?

ranger-ross:
Another question: Should Cargo error if the user adds something like 
```toml
build-dir =  "/foo/{workspace-root}/bar"
```

This is invalid as `{workspace-root}` points to a specific directory so prefixes don't make logical sense.
The same is true for `cargo-cache` but not `{workspace-manifest-path-hash}` since the hash is not pointing to a specific directory. It's merely part of some path.

In my opinion it would be best to error on an invalid path.
Although another approach would be to simply ignore the prefix.

src/cargo/util/context/mod.rs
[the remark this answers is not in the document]
epage: Let's just start this off generic with `cargo-cache-home` pointing to `CARGO_HOME`

epage:
> Another question: Should Cargo error if the user adds something like

On Linux systems, that is valid and I could see someone doing that, rather than using a hash.

On Windows systems, that will be a mess.

I would lean towards not doing anything special wrt and instead we can note this in the tracking issue for people to be aware of this when we stabilize.

src/cargo/util/context/mod.rs
epage: btw I would love for all defaults to eventually be defined in terms of template variables.

src/cargo/util/context/mod.rs
epage: btw I would love for all defaults to eventually be defined in terms of template variables.

ranger-ross: I spent some time trying to move this logic to `workspace.rs` and I ended up changing my mind and keeping it in global context.

The issue I was running into was doing templating _after_ [`ConfigRelativePath::resolve_path()`](https://github.com/rust-lang/cargo/blob/master/src/cargo/util/context/path.rs#L31) is a bit awkward since it joins the workspace root to the path.

I found that doing templating on the `ConfigRelativePath` (a `String` internally) ended up being much easier.

src/doc/src/reference/unstable.md
```diff
@@ -260,6 +260,14 @@ build-dir = "out"
 
 The path to where internal files used as part of the build are placed.
 
+This option supports path templating.
+
+Avaiable template variables:
+* `{workspace-root}` resolves to root of the current workspace.
+* `{cargo-cache-home}` resolves to `CARGO_HOME`
+* `{workspace-manifest-path-hash}` resolves to a hash of the manifest path
+   * This is split into 2 nested directories where the first dir is the first 2 hex chars and the inner dir is the remaining chars. (ie. `34/f9d02eb8411c05`)
```
epage, line 269: FYI these are implementation details that we do not guarantee.

The second I wouldn't even specify.

src/cargo/util/context/path.rs
```diff
@@ -32,6 +32,25 @@ impl ConfigRelativePath {
...
+        gctx: &GlobalContext,
+        replacements: impl IntoIterator<Item = (impl AsRef<str>, impl AsRef<str>)>,
+    ) -> PathBuf {
+        let mut value = self.0.val.clone();
+
+        for (from, to) in replacements {
+            value = value.replace(from.as_ref(), to.as_ref());
+        }
+
+        self.0.definition.root(gctx).join(&value)
```
epage, line 51: What should this do on undefined variables?

ranger-ross: I gave this some thought and I am of the opinion that we should ignore them. (at least for now)

On unix system's its possible to create paths with curly braces.
```sh
> mkdir 'foo/{bar}/baz'
> ls foo
{bar}
```

Soo the following would be legal

```toml
[build]
build-dir = "{workspace-root}/foo/{bar}/baz"
```

Although this is probably a very niche usecase I think it would be better to allow.

tests/testsuite/build_dir.rs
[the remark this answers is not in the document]
ranger-ross: I added a test for symlink links and added some additional logic for symlink handling (primarily for Windows)

#### Note on the implementation 
In Windows, `std::fs::canonicalize` will prefix the path with `\\?\`. This by itself is not a problem but it means that Windows  no longers considers `/` in path valid. 
This would mean Windows users would be required to use `\` exclusively for `build-dir` which is not ideal since its does not match the target dir.

To overcome this, I created `paths::resolve_symlinks` that only resolve symlinks without fully canonicalizing the path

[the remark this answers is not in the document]
epage: The problem is if we add a new variable, what will be the experience on older Cargo?  Is it better to lossily succeed or fail?

src/cargo/util/context/path.rs
epage: The problem is if we add a new variable, what will be the experience on older Cargo?  Is it better to lossily succeed or fail?

tests/testsuite/build_dir.rs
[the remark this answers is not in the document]
epage: Is there a reason you went in this direction?  As I said, cargo script doesn't seem to.  Either way we should be consistent.  I would also lean towards this being its own commit or PR on top of templating so we can see how the behavior changes.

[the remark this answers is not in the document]
ranger-ross: My rational was that keeping the hash the same across symlinks would result in less duplicates.

Sorry, I have gotten a chance to play with cargo script yet. Could you clarify what you mean by cargo script went in another direction?

src/cargo/util/context/path.rs
ranger-ross: Hmmm, good point. Perhaps it's better to start out strict to avoid potential for breaking changes. 

If there is a demand for curly braces in paths, we could probably add the ability escape them in the future.

tests/testsuite/build_dir.rs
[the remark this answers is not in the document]
epage: I went ahead and wrote up some notes at https://github.com/rust-lang/cargo/issues/12207#issuecomment-2711402159

As this is a two way door, I lean towards deferring this out of this PR at least to keep this better focused / simpler.  We should note it in the tracking issue and come to a starting point decision before stabilization.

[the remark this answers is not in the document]
traviscross: Not sure where the best place to consolidate discussion on that question is, but as one perspective on the question you raised in the tracking issue:

> If someone symlinks a manifest or a parent directory, are they expecting isolation or consistency?

With symlinks, I'd always expect consistency.  That is, I'm expecting the path to be canonicalized.

(As you say, of course, it seems this can be deferred.)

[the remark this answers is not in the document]
ranger-ross: Sure that works for me :+1: 

I removed the symlink handling logic from this PR in the lastest push.

src/cargo/util/context/path.rs
epage: Double checking: this hasn't been put in yet, right?

ranger-ross: Correct, I should have time to add it this weekend.

ranger-ross: Added unexpected variable validation in https://github.com/rust-lang/cargo/pull/15236/commits/707d637b510952ac1c80f2394c5c04cef136b644

tests/testsuite/build_dir.rs
```diff
@@ -491,6 +491,31 @@ fn future_incompat_should_output_to_build_dir() {
     assert_exists(&p.root().join("build-dir/.future-incompat-report.json"));
 }
 
+#[cargo_test]
+fn template_should_error_for_invalid_variables() {
```
epage, line 495: Please add this test in an earlier commit

src/cargo/util/context/mod.rs
```diff
@@ -681,7 +681,17 @@ impl GlobalContext {
...
 
-            let path = val.resolve_templated_path(self, replacements);
+            let path = val
+                .resolve_templated_path(self, replacements)
+                .map_err(|e| match e {
+                    path::ResolveTemplateError::UnexpectedVariable {
+                        variable,
+                        raw_template,
+                    } => anyhow!(
+                        "Unexpected variable `{variable}` in build.build-dir path `{raw_template}`"
```
epage, line 691: We try to follow the [rustc style guide](https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-output-style-guide) which says this error should be lower case

src/cargo/util/context/path.rs
epage: Since you are already making changes, when I added this as a stabilization question in the PR description, I also mentioned unmatched `{` / `}` (and forgot to mention it here).

I'll leave the decision up to you of whether you want to check for that and we can always re-evaluate on the way to stabilization.

ranger-ross: I was also thinking about that while implementing the last commit. 
I am inclined to ignore unterminated brackets for now since they can technically be valid file paths on unix.

But I agree we should re-evaluate before stabilizing.

I think that it will come down to whether we consider `{` and `}` in paths a valid usecase.
I'd be perfectly happy to say "No, you cannot use brackets in file paths in templated path configs." (It would simplify validation to a simple `path.contains("{") || path.contains("}")`  after resolving all known template variables)
But the Cargo team members might have different opinions this

epage (approved):

epage:
Thanks!

PR #15237, opened 2025-02-26
refactor(tree): Abstract the concept of a NodeId

epage:

This is a follow up to #15233.

The primary goal is to make things more type safe / easier to follow through adding semantic information to the types.

A secondary benefit is this allows us to track some light weight debug information in NodeIds making debugging Edges a lot easier.
We could put this behind a `#[cfg(debug_asserts)]` but I'm assuming this is cheap enough to not be worth doing that to the code.


weihanglo (approved):
All seem reasonable. Thanks for the refactor again!

PR #15238, opened 2025-02-27
feat: add completions for `--lockfile-path`

Gmin2:

Add auto completion for `cargo build --lockfile-path <TAB>`

Related to #14520 

Screenshot from 2025-02-27 18-38-21

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -353,10 +353,6 @@ pub trait CommandExt: Sized {
                             None => return false,
                         };
 
-                        // allow directories
-                        if path.is_dir() {
-                            return true;
-                        }
                         // allow `Cargo.lock` file
                         file_name == OsStr::new("Cargo.lock")
                     }),


src/cargo/util/command_prelude.rs
```diff
@@ -345,7 +345,27 @@ pub trait CommandExt: Sized {
...
+
+                        // includes "." and ".." entries as they are required for directory navigation
+                        if file_name == OsStr::new(".") || file_name == OsStr::new("..") {
+                            return true;
+                        }
+
+                        // allow directories
+                        if path.is_dir() {
+                            return true;
+                        }
```
weihanglo, line 364: Directories are already potential implicit candidates. Wonder why this is needed.

ref: https://github.com/clap-rs/clap/blob/f2cb42a4538cf5fb2a64536464040a7e10987692/clap_complete/src/engine/custom.rs#L325-L334

epage:
Feel free to squash your commits for how your PR should be merged

src/cargo/util/command_prelude.rs
```diff
@@ -345,7 +345,27 @@ pub trait CommandExt: Sized {
...
-                .help_heading(heading::MANIFEST_OPTIONS),
+                .help_heading(heading::MANIFEST_OPTIONS)
+                .add(clap_complete::engine::ArgValueCompleter::new(
+                    clap_complete::engine::PathCompleter::any().filter(|path: &Path| {
+                        let file_name = match path.file_name() {
+                            Some(name) => name,
+                            None => return false,
+                        };
+
+                        // includes "." and ".." entries as they are required for directory navigation
```
epage, line 356: If these dir part are needed for files, then thats a bug in `clap_complete`. They shouldn't be needed though since all dirs should always be included, just filtered ones get but last

src/cargo/util/command_prelude.rs
```diff
@@ -345,7 +345,27 @@ pub trait CommandExt: Sized {
...
+                        if file_name == OsStr::new(".") || file_name == OsStr::new("..") {
```
weihanglo, line 357: Do we intentionally make `.` and `..` as candidates to display? I don't think without we can do directory navigation. And `--manifest-path` doesn't have them either.

src/cargo/util/command_prelude.rs
Gmin2: Thanks for pointing that out, yu are right about the directories part `.` and `..` are already part of `is_dir()`

src/cargo/util/command_prelude.rs
Gmin2: see https://github.com/rust-lang/cargo/pull/15238#discussion_r1973616678, the directories are already part of the `is_dir`, removing this check

and also do we need to include hidden folders , i think we need to put a check for it

src/cargo/util/command_prelude.rs
Gmin2: in my opinion we should include directory navigation(should modify the manifest-path flag also )

Gmin2: removed from this pr to keep the consistency with `--manifest-path` and should i open a issue in `clap-rs/clap` regarding this?

epage: > should i open a issue in clap-rs/clap regarding this?

If you can name a concrete problem with a small reproduction case, feel free.

src/cargo/util/command_prelude.rs
```diff
@@ -345,7 +345,22 @@ pub trait CommandExt: Sized {
...
+                    clap_complete::engine::PathCompleter::any().filter(|path: &Path| {
+                        let file_name = match path.file_name() {
+                            Some(name) => name,
+                            None => return false,
+                        };
+
+                        // allow directories
+                        if path.is_dir() {
+                            return true;
+                        }
```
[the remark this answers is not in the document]
Gmin2: done now, sorry for the overlook

epage (approved):

epage:
Thanks!

PR #15240, opened 2025-02-27
fix(package): Ensure we can package directories ending with '.rs'

epage:

This likely only affects `-Zpackage-workspace` but it might have also broken dependencies whose path ends with `.rs` as
well

This broke in https://github.com/rust-lang/cargo/pull/14961

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -45,9 +45,10 @@ pub use embedded::ScriptSource;
 /// See also `bin/cargo/commands/run.rs`s `is_manifest_command`
 pub fn is_embedded(path: &Path) -> bool {
     let ext = path.extension();
-    ext == Some(OsStr::new("rs")) ||
+    (ext == Some(OsStr::new("rs")) ||
         // Provide better errors by not considering directories to be embedded manifests
-        (ext.is_none() && path.is_file())
+        ext.is_none())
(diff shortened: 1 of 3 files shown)


epage:
Should we do a beta backport?

weihanglo (approved):
Thanks!

weihanglo:
> Should we do a beta backport?

If we want to be extra cautious, yes, though I cannot find a way to break any stable features.
I'll do the remaining work if you open a beta backport PR.

epage:
Confirmed: this is visible with stable behavior. 
```console
$ cargo new foo --lib
$ mv foo foo.rs
$ cargo add --path foo.rs
$ cargo check
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.03s
$ cargo +nightly check
error: failed to get `foo` as a dependency of package `cargo-15240 v0.1.0 (/home/epage/src/personal/dump/cargo-15240)`

Caused by:
  failed to load source for dependency `foo`

Caused by:
  Unable to update /home/epage/src/personal/dump/cargo-15240/foo.rs

Caused by:
  Single file packages cannot be used as dependencies
$ nargo check
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.02s
```

weihanglo:
Beta backport submitted: https://github.com/rust-lang/cargo/pull/15248

PR #15243, opened 2025-02-27
feat(tree): Add `--depth public` behind `-Zunstable-options`

epage:

I was investigating some issues around public dependency lints and wanted to see the structure of the public dependencies and had the idea to add this with us having added `--depth workspace`.

See https://github.com/rust-lang/rust/issues/119428#issuecomment-2686384070 for some example output (comparing `cargo tree` with `cargo tree --depth public`)

### DIFF
--- a/src/cargo/ops/tree/graph.rs
+++ b/src/cargo/ops/tree/graph.rs
@@ -78,6 +78,7 @@ impl Node {
 pub struct Edge {
     kind: EdgeKind,
     node: NodeId,
+    public: bool,
 }
 
 impl Edge {
@@ -87,6 +88,10 @@ impl Edge {
 
     pub fn node(&self) -> NodeId {
         self.node
+    }
+
+    pub fn public(&self) -> bool {
+        self.public
     }
 }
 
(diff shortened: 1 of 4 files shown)


weihanglo (approved):
Looks good. Thanks!

weihanglo:
This <https://github.com/rust-lang/cargo/issues/16043> reminds me that maybe this makes more sense to be an edge kind instead of depth, as "public dependency" is less a depth things more a dependency stuff.

epage:
That is worth considering.

Looking over the [docs](https://doc.rust-lang.org/cargo/commands/cargo-tree.html), the weird part is that the edges specified are distinct edge types in `Cargo.toml` (except `no-proc-macros` which is a weird case).  This would effectively be the first time we'd be adding an edge type based on a cross-cutting attribute of an edge.  I assume `-e public` would filter other edge types based on public, but is that clear enough?  In a way, a `no-non-public` would make that type of filtering clearer but that is a bit annoying with the double negative.  `no-private` resolves the double negative but now we've moved away from the naming of the manifest field.

PR #15253, opened 2025-03-02
Remove `Cargo.toml` from `package.include` in example

ZaneErebos:
https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields

>The following files are always included:
>
>* The `Cargo.toml` file of the package itself is always included, it does not need to be listed in `include`.

### DIFF
--- a/src/doc/src/reference/publishing.md
+++ b/src/doc/src/reference/publishing.md
@@ -114,7 +114,6 @@ If you’d rather explicitly list the files to include, Cargo also supports an
 # ...
 include = [
     "**/*.rs",
-    "Cargo.toml",
 ]
 ```
 
--- a/src/doc/src/reference/publishing.md
+++ b/src/doc/src/reference/publishing.md
@@ -107,7 +107,7 @@ exclude = [
 ```
 
 If you’d rather explicitly list the files to include, Cargo also supports an
-`include` key, which if set, overrides the `exclude` key:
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/publishing.md
```diff
@@ -114,7 +114,6 @@ If you’d rather explicitly list the files to include, Cargo also supports an
 # ...
 include = [
     "**/*.rs",
-    "Cargo.toml",
```
weihanglo, line 117: nit: maybe also this to point out not only Rust files can be included.

```suggestion
    "resources/to/include/*",
```

weihanglo (approved):

src/doc/src/reference/publishing.md
ZaneErebos: To avoid confusion would it be better to also add a link to the relevant docs since there is a difference between `resources/to/include/*`, `resources/to/include/`, and `resources/to/include`? Or maybe just mention that it should be `gitignore`-like patterns?

weihanglo: Good point. Yeah we can link to the same paragraph like this for the `include` key as well:
https://github.com/rust-lang/cargo/blob/592fd2ea2f5aeb3f013bae0ca040cd5c896c2ed2/src/doc/src/reference/publishing.md?plain=1#L97

ZaneErebos: Sounds good

weihanglo (approved):
Thanks!

PR #15255, opened 2025-03-02
 Use a dynamic bound for the queue

oriongonza:
If the queue is at capacity (100) and we `push` something to it it will grow, but the `push_bounded` will still block when there's plenty of space left.

### DIFF
--- a/src/cargo/util/queue.rs
+++ b/src/cargo/util/queue.rs
@@ -17,7 +17,6 @@ pub struct Queue<T> {
     state: Mutex<State<T>>,
     popper_cv: Condvar,
     bounded_cv: Condvar,
-    bound: usize,
 }
 
 struct State<T> {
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR! I don't think the capacity of VecDeque is guaranteed to be exact. Can you say more about why this is needed?

oriongonza:
This is not exactly needed, but this solves the issue mentioned above, the unfortunate situation when the sender blocks when it doesn't need to.

> I don't think the capacity of VecDeque is guaranteed to be exact

https://github.com/rust-lang/cargo/blob/master/src/cargo/core/compiler/job_queue/mod.rs#L485 I don't think you care ;)

weihanglo (commented):
> but the push_bounded will still block when there's plenty of space left.

The bound is expected to be static if I understand the code correctly.

Why does this need to support a dynamic bound? Any Cargo bad behavior you saw?

weihanglo:
Close due to inactivity and no response for a few months.

PR #15259, opened 2025-03-03
Improve error when versions start with `v`

### LINKED ISSUE #12331: Improve the error message when running `cargo add cubing @v0.5.0`
This has now been rescoped to providing a good error message if a `v` prefix is present, see https://github.com/rust-lang/cargo/issues/12331#issuecomment-1642551113

---

### Problem

I keep doing something the following:

```shell
cargo add cubing@v0.5.0
```

This causes the following error:

```
error: invalid version requirement `v0.5.0`

Caused by:
  unexpected character 'v' while parsing major version number
```

I do this because I have a strong habit of prefixing my version numbers with `v` where possible, as `vX.Y.Z` tends to be a more unmistakable way to specify a version. And although it is not universal, it is also very common to prefix project release tags with `v` — to the point that the canonical `cargo` test for for a `tag` dependency even uses a value like this: https://github.com/rust-lang/cargo/blob/5b377cece0e0dd0af686cf53ce4637d5d85c2a10/tests/testsuite/cargo_add/git_tag/mod.rs#L22C1-L22C1
(I know that doesn't have a *semantic* meaning in this case, but it illustrates how widespread it is to indicate a version using a prefix of `v`.)

Further, other tools like `npm` allow a prefix of `v` (e.g. `npm install cubing@v0.38.1`), which is an inconsistency that causes a small stumbling block when working with multiple web technologies.

### Proposed Solution

In order of preference:

- Accept an extra `v` before a semantic version number.
- Recognize the extra `v` and provide a helpful error message that includes something like: `If you want to install a specific version, please run: cargo add cubing@0.5.0`

### Notes

_No response_

DeveloperC286:
Fixes #12331

### DIFF
--- a/src/bin/cargo/commands/yank.rs
+++ b/src/bin/cargo/commands/yank.rs
@@ -54,14 +54,26 @@ fn resolve_crate<'k>(
             }
 
             match version {
-                None => Ok((Some(name), embedded_version)),
+                None => {
+                    if embedded_version.starts_with('v') {
+                        Ok((Some(name), &embedded_version[1..]))
+                    } else {
+                        Ok((Some(name), embedded_version))
+                    }
+                }
                 Some(_) => {
                     anyhow::bail!("cannot specify both `@{embedded_version}` and `--version`");
                 }
             }
         }
         None => match version {
-            Some(version) => Ok((krate, version)),
+            Some(version) => {
+                if version.starts_with('v') {
+                    Ok((krate, &version[1..]))
+                } else {
+                    Ok((krate, version))
+                }
+            }
             None => {
                 anyhow::bail!("`--version` is required");
             }
--- a/src/bin/cargo/commands/yank.rs
+++ b/src/bin/cargo/commands/yank.rs
@@ -30,14 +30,11 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
         args.get_one::<String>("crate").map(String::as_str),
         args.get_one::<String>("version").map(String::as_str),
     )?;
-    if version.is_none() {
-        return Err(anyhow::format_err!("`--version` is required").into());
-    }
 
     ops::yank(
         gctx,
         krate.map(|s| s.to_string()),
-        version.map(|s| s.to_string()),
+        version.to_string(),
         args.get_one::<String>("token").cloned().map(Secret::from),
         args.registry_or_index(gctx)?,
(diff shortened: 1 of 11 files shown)


src/bin/cargo/commands/yank.rs
```diff
@@ -30,14 +30,11 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
         args.get_one::<String>("crate").map(String::as_str),
         args.get_one::<String>("version").map(String::as_str),
     )?;
-    if version.is_none() {
```
epage, line 33: Moving this into the function would ideally be a refactor commit on its own

DeveloperC286: Moved the refactoring into its own commit.

DeveloperC286:
While working on this I have the questions

1. Do we want to standardise the error messages for the same error across the commands?
2. Do we want to support a v in a version range?

epage:
> Do we want to standardise the error messages for the same error across the commands?

We should be consistent where it makes sense

> Do we want to support a v in a version range?

As in not error?  This was discussed in the issue and the issue is currently scoped to just improving the error messages.  If there is a reason to reconsider that, the case should be made in the issue.  We should probably still move forward with this PR.

weihanglo:
@DeveloperC286 
I believe what Ed said is backing out the support of v prefix in this PR. That part is still in discussion. We should focus on improving the error messages in this PR.

DeveloperC286:
Oh okay, I got my wires crossed somewhere. I thought there was support for handing versions prefixed with 'v'.

While a version prefixed with a `v` is not strictly a semantic version it is commonly used, e.g. output by automated release tooling etc, so not supporting it just makes it harder to integrate with other tools.

I'm not sure I see the point if I run `cargo install test@v0.5.0` of an error message saying `did you mean 'cargo... @0.5.0'?`. I think if your able to produce an error message that clear then the intent is clear and if the intent is clear I don't see the point in being pedantic and not just doing it.

epage:
Please see my earlier comment at https://github.com/rust-lang/cargo/pull/15259#issuecomment-2707092237

DeveloperC286:
I only have limited free time ATM and was doing this piecemeal, I think reworking this PR to just improved error messages takes me back basically to the start. Someone feel free to take this over, I will close this as I don't have the time ATM.

PR #15260, opened 2025-03-04
fix(vendor): dont remove non-cached source 

### LINKED ISSUE #15244: `cargo vendor --respect-source-config` deletes previously vendored sources
### Problem

Running `cargo vendor > .cargo/config.toml` followed by running `cargo vendor --respect-source-config` results in a failure.  Moreover `cargo vendor` deletes the previous contents of the `vendor` folder.

```
% cargo vendor --versioned-dirs --respect-source-config --no-delete
error: failed to sync

Caused by:
  failed to load pkg lockfile

Caused by:
  no matching package named `bytemuck` found
  location searched: directory source `/Users/alklig/work/bad_vendor/vendor` (which is replacing registry `crates-io`)
  required by package `bad_vendor v0.1.0 (/Users/alklig/work/bad_vendor)`
```

### Steps

1. clone https://github.com/lambdageek/bad-vendor
2. `cargo vendor --versioned-dirs --respect-source-config --no-delete > .cargo/config.toml`
3. `cargo vendor --versioned-dirs --respect-source-config --no-delete`

### Possible Solution(s)

I suspect something in the loop trying to deal with #5956 is not expecting to see replaced sources.  In particular, I wonder if this line is deleting files in the "vendor" directory

https://github.com/rust-lang/cargo/blob/3fe24ab6b4d2474b36ef614b5dea51bd4e0f8cf2/src/cargo/ops/vendor.rs#L137

### Notes

_No response_

### Version

```text
cargo 1.84.1
release: 1.84.1
host: aarch64-apple-darwin
os: Mac OS 15.3.1 [64-bit]
```

weihanglo:

Fixes #15244

With this fix,
`cargo vendor` will not delete original sources,
if you want to vendor things from one directory sources to the other

#### Background

cargo-vendor has a workaround that to mitigate #5956:
it removes all cached sources in order to trigger a re-unpack.
It was meant for dealing with registry sources only,
but accidentally applied to directory source kind.

While directory source kind was invented for vendoring,
and vendoring from one vendored directory to the other seems unusual,
Cargo IMO should not delete any real sources.

It does not mean that registry sources are okay to delete,
In long term, we should explore a way that unpacks `.crate` files
directly, without any removal. See
https://github.com/rust-lang/cargo/pull/12509#issuecomment-1732415990

The added test should suffice.

Also, although this is for fixing #15244,
`cargo vendor` still doesn't support vendor from and to the same location.
Unless we figure out an `rsync`-like solutin to update vendor sources,
it is not going to support in short term.
(And I also doubt the real world use case of it)

### DIFF
--- a/src/cargo/ops/vendor.rs
+++ b/src/cargo/ops/vendor.rs
@@ -1,8 +1,10 @@
 use crate::core::shell::Verbosity;
+use crate::core::SourceId;
 use crate::core::{GitReference, Package, Workspace};
 use crate::ops;
 use crate::sources::path::PathSource;
 use crate::sources::PathEntry;
+use crate::sources::SourceConfigMap;
 use crate::sources::CRATES_IO_REGISTRY;
 use crate::util::cache_lock::CacheLockMode;
 use crate::util::{try_canonicalize, CargoResult, GlobalContext};
@@ -21,6 +23,7 @@ pub struct VendorOptions<'a> {
     pub versioned_dirs: bool,
     pub destination: &'a Path,
     pub extra: Vec<PathBuf>,
+    pub respect_source_config: bool,
 }
 
 pub fn vendor(ws: &Workspace<'_>, opts: &VendorOptions<'_>) -> CargoResult<()> {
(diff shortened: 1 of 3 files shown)


src/cargo/ops/vendor.rs
```diff
@@ -121,20 +152,31 @@ fn sync(
...
                     if let Ok(path) = try_canonicalize(path) {
                         to_remove.remove(&path);
                     }
                 }
                 continue;
             }
-            if pkg.source_id().is_git() {
+            if sid.is_git() {
                 continue;
             }
```
epage, line 172: nit: rather than `if`s with `continue`s that keep falling through cases, it seems like `if`/`else if` would better express the mutually exclusive nature of this loop.

epage (approved):
nit is unlikely worth waiting on a merge

PR #15266, opened 2025-03-05
feat: add completions for install --path

NOOMA-42:

This PR introduces auto-completion for the cargo install --path option. When a user types cargo install --path  and presses the TAB key, the system will automatically detect directories in the current path and suggestion prioritizes directory containing a Cargo.toml file. 

1. In the terminal, type `cargo install --path`
2. Press the TAB key.
3. You should see suggestions only for directories in the current path that contain a `Cargo.toml` file being prioritized, such as `./src/` (if `./src/Cargo.toml` exists). 

https://github.com/user-attachments/assets/281706c0-7956-4551-a22b-125dadccd12e

### Additional information

There's a [discussion in Zulip](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/PR15266.3A.20completions.20for.20install.20--path)  and here're some relevant information from it
1. directories are always shown so that a completer doesn't have to recurse through a directory tree to determine what directories may be relevant. The difference marking a directory as accepted does it sorts it first.
2. A user / config setting might override a sort order over native completion. Try disable it manually if native completion is not there.

### DIFF
--- a/src/bin/cargo/commands/install.rs
+++ b/src/bin/cargo/commands/install.rs
@@ -65,7 +65,11 @@ pub fn cli() -> Command {
         .arg(
             opt("path", "Filesystem path to local crate to install from")
                 .value_name("PATH")
-                .conflicts_with_all(&["git", "index", "registry"]),
+                .conflicts_with_all(&["git", "index", "registry"])
+                .add(clap_complete::engine::ArgValueCompleter::new(
+                    clap_complete::engine::PathCompleter::any()
+                        .filter(|path| path.join("Cargo.toml").exists()),
+                )),
         )
         .arg(opt("root", "Directory to install packages into").value_name("DIR"))
         .arg(flag("force", "Force overwriting existing crates or binaries").short('f'))


src/bin/cargo/commands/install.rs
```diff
@@ -65,7 +68,21 @@ pub fn cli() -> Command {
...
             opt("path", "Filesystem path to local crate to install from")
                 .value_name("PATH")
-                .conflicts_with_all(&["git", "index", "registry"]),
+                .conflicts_with_all(&["git", "index", "registry"])
+                .add(clap_complete::engine::ArgValueCompleter::new(
+                    clap_complete::engine::PathCompleter::any().filter(|path: &Path| {
+                        if path.is_dir() {
+                            return path.join("Cargo.toml").exists();
+                        }
+                        if path.file_name() == Some(OsStr::new("Cargo.toml")) {
```
epage, line 77: I don't think you can `cargo install --path Cargo.toml` only `cargo install --path .`

src/bin/cargo/commands/install.rs
weihanglo, line 1: Thanks for the contribution!

Likewise, could you in the PR description include a simple explanation of what you want to complete from `--path`? 

I think it useful because currently native completion lack for tests. If the behavior changes in the future, we can look at what we wrote before and know it is a regression. See this [PR description](https://github.com/rust-lang/cargo/pull/15225) as an example.

NOOMA-42:
This is my first time contributing to Cargo, so I'm not very familiar with it. Please bear with me if this is a dumb question.
I wanted to ask if cargo native-completion can print debug messages. I followed the instructions here and added some environment variables, but nothing got logged to a file or printed in the terminal.

https://doc.crates.io/contrib/implementation/debugging.html

weihanglo:
> I wanted to ask if cargo native-completion can print debug messages. I followed the instructions here and added some environment variables, but nothing got logged to a file or printed in the terminal.

If you're adding tracing log like `tracing::info!`, try exporting `CARGO_LOG` environment variable, so your subsequent `cargo` call can see the variable. For example,

```
# zsh
export CARGO_LOG=info
cargo install --path <tab>
```

If you feel logging too hassle to use, use `dbg!` or just `panic!` so you'll see it without setting any env var.

epage:
Anything written to `stdout` will be read by the completion shim and interpreted as completions to report to the user
https://github.com/clap-rs/clap/blob/fdbbf66e77c83688f52b7a206d64102582af40d3/clap_complete/src/env/shells.rs#L40-L46

Unsure what else a shell's completion system may do on top of what we're doing.

You can also directly invoke the completer so nothing gets captured.  You can run `CARGO_COMPLETE=zsh cargo +nightly`, see how cargo is invoked and do similar.

src/bin/cargo/commands/install.rs
```diff
@@ -65,7 +66,15 @@ pub fn cli() -> Command {
...
             opt("path", "Filesystem path to local crate to install from")
                 .value_name("PATH")
-                .conflicts_with_all(&["git", "index", "registry"]),
+                .conflicts_with_all(&["git", "index", "registry"])
+                .add(clap_complete::engine::ArgValueCompleter::new(
+                    clap_complete::engine::PathCompleter::any().filter(|path: &Path| {
+                        if path.is_dir() {
+                            return path.join("Cargo.toml").exists();
+                        }
+                        false
```
weihanglo, line 75: * Type annotation doesn't seem necessary here.
* nit: Could simplify to `path.is_dir() && path.join("Cargo.toml").exists()`.

weihanglo (commented):
Mind fixing up the commits for how you prefer them to get merged?

src/bin/cargo/commands/install.rs
```diff
@@ -65,7 +66,15 @@ pub fn cli() -> Command {
...
+                    }),
```
epage, line 76:
```suggestion
                        path.join("Cargo.toml").exists();
                    }),
```
The `is_dir` is redundant

epage:
Feel free to squash your commits for how you would like them to be merged

epage (approved):
Thanks!

PR #8167, opened 2020-04-27
Support multiple `--target` flags on the CLI

alexcrichton:
This commit refactors the internals of Cargo to no longer have a
singular `--target` flag (and singular `requested_target` kind throught)
but to instead have a list. The semantics of multiple `--target` flags
is to build the selected targets for each of the input `--target` flag
inputs.

For now this is gated behind `-Zmultitarget` as an unstable features,
since I'm not entirely sure this is the interface we want. In general
it'd be great if we had a way to simply specify `Unit` structures of
what to build on the CLI, but we're in general very far away from that,
so I figured that this is probably sufficient at least for testing for
now.

cc #8156

PR #15271, opened 2025-03-06
Don't use `$CARGO_BUILD_TARGET` in `cargo metadata`

### LINKED ISSUE #15270: The `cargo metadata --filter-platform` flag defaults to `$CARGO_BUILD_TARGET` and can't be overridden
Today the `--filter-platform` flag to the `cargo metadata` subcommand defaults to `$CARGO_BUILD_TARGET` or the configuration of `build.target` within `.cargo/config.toml`. This behavior does not appear to be overridable (or I'm also perhaps missing something in which case this issue is moot), meaning that if a tool wants to execute `cargo metadata` specifically without a platform filter then there's no way for the tool to opt-in to that behavior.

This issue is downstream of  https://github.com/mozilla/cargo-vet/issues/652 where the bug doesn't actually lie in that repository, but rather in the interaction of the various features in play here. In short `cargo vet` runs `cargo metadata` to learn about the crate graph and generally wants all information about all platforms. Locally I have `build.target` configured in `.cargo/config.toml`, however, which means that the results of `cargo vet` are subtly different on my machine. This discrepancy in behavior was the source of the original bug report.

To reproduce this behavior while I'm sure it can be minimized further I can see this with the [wasmtime repository](https://github.com/bytecodealliance/wasmtime):

```
$ cargo metadata --format-version 1  | shasum
8272634713a68939780da2c8387b5ddcab57dbc7  -
$ cargo metadata --format-version 1 --filter-platform x86_64-unknown-linux-gnu | shasum
4525e3f757ea995e4691188f26b4fc44f7ef6379  -
$ CARGO_BUILD_TARGET=x86_64-unknown-linux-gnu cargo metadata --format-version 1 | shasum
4525e3f757ea995e4691188f26b4fc44f7ef6379  -
```

Effectively `--filter-platform`, as expected, changes the output of `cargo metadata`. The `CARGO_BUILD_TARGET` then affects this by default so even when `--filter-platform` isn't passed it produces the same output as-if it were passed.

I'll note that I'm assuming that the behavior of reading `CARGO_BUILD_TARGET` is intentional, but I also realize that may also not be intentional. If it's not intentional, then that's also a possible fix to this! If it is intentional, then this would boil down to a feature request to have a CLI flag for something like `--filter-platform all` to explicitly request, regardless of other configuration, that no filter be applied.

alexcrichton:
This commit fixes a (five year old!) regression in `cargo metadata` where if `--filter-platform` isn't explicitly specified it will accidentally read `$CARGO_BUILD_TARGET` (or `build.target` configuration) and use that as the default `--filter-platform`. The reason for this is that the calculation for targets changed in #8167 and while the shared function makes sense for other commands such as `cargo build` the targets have a different meaning in `cargo metadata` so a slightly different set of functionality is desired.

This commit fixes the issue by introducing a new constructor for the list of `CompileKind` variants where the fallback of "if nothing is specified" is explicitly chosen.

Closes #15270

### DIFF
--- a/tests/testsuite/metadata.rs
+++ b/tests/testsuite/metadata.rs
@@ -4954,3 +4954,34 @@ local-time = 1979-05-27
         )
         .run();
 }
+
+#[cargo_test]
+fn metadata_ignores_build_target_configuration() -> anyhow::Result<()> {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+
+                [target.'cfg(something)'.dependencies]
+                foobar = "0.0.1"
+           "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+    Package::new("foobar", "0.0.1").publish();
+
+    let output1 = p
+        .cargo("metadata -q --format-version 1")
+        .exec_with_output()?;
+    let output2 = p
+        .cargo("metadata -q --format-version 1")
+        .env("CARGO_BUILD_TARGET", rustc_host())
+        .exec_with_output()?;
+    assert!(
+        output1.stdout != output2.stdout,
+        "metadata should change when `CARGO_BUILD_TARGET` is set",
+    );
+    Ok(())
+}
--- a/tests/testsuite/metadata.rs
+++ b/tests/testsuite/metadata.rs
@@ -1224,7 +1224,7 @@ fn workspace_metadata_with_dependencies_no_deps() {
                 name = "bar"
                 version = "0.5.0"
                 authors = ["wycats@example.com"]
-                
+
                 [dependencies]
                 baz = { path = "../baz/" }
                 artifact = { path = "../artifact/", artifact = "bin" }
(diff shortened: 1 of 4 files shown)


tests/testsuite/metadata.rs
```diff
@@ -4954,3 +4954,34 @@ local-time = 1979-05-27
         )
         .run();
 }
+
+#[cargo_test]
+fn metadata_ignores_build_target_configuration() -> anyhow::Result<()> {
```
epage, line 4959: nit: we prefer contributions split tests into the first commit, with them passing.  That way the fix commit will show the change in behavior by how the test changes.  This helps test the test and provides an illustration of the PR description for the reviewer and the community at large.

We linked to examples in the [contrib docs explaining this](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request)

epage:
Could you fixup the commits for how you would prefer them to be merged?

alexcrichton:
Sure thing, can do! Updated with an extra change to go ahead and strip the trailing whitespace too, but I'm happy to remove that if desired as well.

epage (approved):

PR #15276, opened 2025-03-07
fix(package): report lockfile / workspace manifest is dirty 

### LINKED ISSUE #14967: `cargo package` VCS dirty check corner cases

weihanglo:

Workspace manifest and lockfile might not be under the package root,
but workspace inheritance and outdated lockfile entries
may still affect what is packaged into the `.crate.` tarball.
We take a conservative action that if any of them is dirty,
then all workspace members are considered dirty.

Fixes #14967

Workspace manifest case is simple.

Lockfile requires a bit more works.
It is allowed to be missing, and can be generated during packaging.
We skip checking when it is missing in both workdir and git index,
otherwise cargo will fail with git2 not found error.


tests/testsuite/package.rs
```diff
@@ -1432,6 +1426,133 @@ edition = "2021"
...
+
+to proceed despite this and include the uncommitted changes, pass the `--allow-dirty` flag
+
+"#]])
+        .run();
+
+    // Now check if it is considered dirty when removed.
+    p.cargo("clean").run();
+    fs::remove_file(p.root().join("Cargo.lock")).unwrap();
+    p.cargo("package --workspace --no-verify")
```
weihanglo, line 1543: Might be some path normalization issues on Windows.

```
thread 'package::dirty_ws_lockfile_dirty' panicked at tests\testsuite\package.rs:1553:10:
test failed running `D:\a\cargo\cargo\target\debug\cargo.exe package --workspace --no-verify`
error: process exited with code 0 (expected 101)
--- stdout
--- stderr
   Packaging isengard v0.0.0 (D:\a\cargo\cargo\target\tmp\cit\t2239\foo\isengard)
    Packaged 5 files, 1.5KiB (986B compressed)
```

weihanglo: Fixed by <https://github.com/rust-lang/cargo/compare/4c0e42dac2b05d52b52ef2435a9a25d73dd92c5e..f0907fca16a84101d69fc53272a955e8bb53d9fe>.

It was failing because removed files can't be canonicalized, so when comparing path prefix with a canonicalized `workdir` it will never pass.

src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -248,20 +254,50 @@ fn dirty_files_outside_pkg_root(
...
+        None
+    };
+
     for rel_path in src_files
         .iter()
         .filter(|p| p.is_symlink_or_under_symlink())
         .map(|p| p.as_ref().as_path())
         .chain(metadata_paths.iter().map(AsRef::as_ref))
         .chain([ws.root_manifest()])
+        .chain(lockfile_path.as_deref().into_iter())
```
epage, line 300: I'm trying to think through how this PR might be disruptive to people's workflows and the most likely case I can come ip with is people being sloppy with version bumps and publishing which can run into #5979.

epage (approved):

PR #15281, opened 2025-03-08
Fix cargo add overwriting symlinked Cargo.toml files

### LINKED ISSUE #15241: cargo add overwrites a symlink Cargo.toml
### Problem

I'll start with a codeblock as imo it's the most straightforward explanation:

```
$ ls -l
drwxr-xr-x src
lrwxrwxrwx target -> /home/axlefublr/.cache/wks/hello-world.rs/target
lrwxrwxrwx Cargo.lock -> /home/axlefublr/.cache/wks/hello-world.rs/Cargo.lock
lrwxrwxrwx Cargo.toml -> /home/axlefublr/.cache/wks/hello-world.rs/Cargo.toml

$ cargo --version
cargo 1.87.0-nightly (1d1d646c0 2025-02-21)

$ rustup run stable cargo --version
cargo 1.85.0 (d73d2caf9 2024-12-31)

$ rustup run stable cargo add clap
    Updating crates.io index
      Adding clap v4.5.31 to dependencies
			// --SNIP--

$ ls -l
drwxr-xr-x src
lrwxrwxrwx target -> /home/axlefublr/.cache/wks/hello-world.rs/target
lrwxrwxrwx Cargo.lock -> /home/axlefublr/.cache/wks/hello-world.rs/Cargo.lock
.rw-r--r-- Cargo.toml

$ rm -fr Cargo.toml

$ ln -sf ~/.cache/wks/hello-world.rs/Cargo.toml ./Cargo.toml

$ cargo add clap
    Updating crates.io index
      Adding clap v4.5.31 to dependencies
			// --SNIP--

$ ls -l
drwxr-xr-x src
lrwxrwxrwx target -> /home/axlefublr/.cache/wks/hello-world.rs/target
lrwxrwxrwx Cargo.lock -> /home/axlefublr/.cache/wks/hello-world.rs/Cargo.lock
.rw-r--r-- Cargo.toml
```

I have Cargo.toml symlinked to the real Cargo.toml somewhere else. When using `cargo add`, I expect the symlink to not be broken. However, `cargo add` overrides the symlink into being a normal file, breaking the symlink tie to that other Cargo.toml. \
I don't think that's intentional.

### Steps

```fish
mkdir -p test/src
cd test
touch src/main.rs
ln -sf ~/.cache/wks/hello-world.rs/Cargo.toml ./Cargo.toml # path to some other Cargo.toml goes here, instead of the first path
cargo add clap # doesn't matter which dependency you add
```

viola. `stat` the Cargo.toml to see that it's now a regular file, rather than a symlink. \
the target Cargo.toml stays untouched, btw

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
cargo 1.85.0 (d73d2caf9 2024-12-31)
release: 1.85.0
commit-hash: d73d2caf9e41a39daf2a8d6ce60ec80bf354d2a7
commit-date: 2024-12-31
os: EndeavourOS Rolling Release (rolling) [64-bit]
```

RaghavenderSingh:

This PR fixes a bug where `cargo add` breaks symlinks to Cargo.toml files. Currently, when Cargo.toml is a symlink and `cargo add` is used to add a dependency, the symlink is replaced with a regular file, breaking the link to the original target file.

This issue was reported in #15241 where a user who relies on symlinked Cargo.toml files found that `cargo add` breaks their workflow.

Fixes #15241

I've modified `LocalManifest::write()` to check if the path is a symlink, and if so, follow it to get the actual target path. This ensures we write to the actual file rather than replacing the symlink.

I've also added a test in `tests/testsuite/cargo_add/symlink.rs` that:
1. Creates a symlinked Cargo.toml file
2. Runs `cargo add` to add a dependency
3. Verifies the symlink is preserved and the dependency is added to the target file

I've manually tested this fix and confirmed it works correctly.

### DIFF
--- a/crates/cargo-util/src/paths.rs
+++ b/crates/cargo-util/src/paths.rs
@@ -190,8 +190,19 @@ pub fn write<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()>
 /// Writes a file to disk atomically.
 ///
 /// This uses `tempfile::persist` to accomplish atomic writes.
+/// If the path is a symlink, it will follow the symlink and write to the actual target.
 pub fn write_atomic<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()> {
     let path = path.as_ref();
+
+    // Check if the path is a symlink and follow it if it is
+    let resolved_path;
+    let path = if path.is_symlink() {
+        resolved_path = fs::read_link(path)
+            .with_context(|| format!("failed to read symlink at `{}`", path.display()))?;
+        &resolved_path
+    } else {
+        path
+    };
 
     // On unix platforms, get the permissions of the original file. Copy only the user/group/other
     // read/write/execute permission bits. The tempfile lib defaults to an initial mode of 0o600,
@@ -981,6 +992,33 @@ mod tests {
              "
             );
         }
+    }
+
+    #[test]
+    fn write_atomic_symlink() {
+        let tmpdir = tempfile::tempdir().unwrap();
+        let target_path = tmpdir.path().join("target.txt");
+        let symlink_path = tmpdir.path().join("symlink.txt");
+
+        // Create initial file
+        write(&target_path, "initial").unwrap();
+
+        // Create symlink
+        #[cfg(unix)]
+        std::os::unix::fs::symlink(&target_path, &symlink_path).unwrap();
+        #[cfg(windows)]
+        std::os::windows::fs::symlink_file(&target_path, &symlink_path).unwrap();
+
+        // Write through symlink
+        write_atomic(&symlink_path, "updated").unwrap();
+
+        // Verify both paths show the updated content
+        assert_eq!(std::fs::read_to_string(&target_path).unwrap(), "updated");
+        assert_eq!(std::fs::read_to_string(&symlink_path).unwrap(), "updated");
+
+        // Verify symlink still exists and points to the same target
+        assert!(symlink_path.is_symlink());
+        assert_eq!(std::fs::read_link(&symlink_path).unwrap(), target_path);
     }
 
     #[test]
--- a/tests/testsuite/cargo_add/symlink.rs
+++ b/tests/testsuite/cargo_add/symlink.rs
@@ -47,6 +47,8 @@ fn symlink_case() {
 
     project.cargo("add test-dep").run();
 
-    // Current behavior: symlink is NOT preserved (gets replaced)
-    assert!(!project.root().join("Cargo.toml").is_symlink());
+    assert!(project.root().join("Cargo.toml").is_symlink());
+
+    let target_content = fs::read_to_string(target_dir.join("Cargo.toml")).unwrap();
+    assert!(target_content.contains("test-dep"));
 }
--- a/tests/testsuite/cargo_add/symlink.rs
+++ b/tests/testsuite/cargo_add/symlink.rs
@@ -1,0 +1,52 @@
+use cargo_test_support::prelude::*;
+use cargo_test_support::project;
+use cargo_test_support::registry;
+use std::fs;
+
+#[cargo_test]
+fn symlink_case() {
+    if !cargo_test_support::symlink_supported() {
+        return;
+    }
+
+    registry::init();
+    registry::Package::new("test-dep", "1.0.0").publish();
+
+    let project = project().file("src/lib.rs", "").build();
+
+    let target_dir = project.root().join("target_dir");
+    fs::create_dir_all(&target_dir).unwrap();
+
+    fs::copy(
+        project.root().join("Cargo.toml"),
+        target_dir.join("Cargo.toml"),
+    )
+    .unwrap();
+
+    fs::remove_file(project.root().join("Cargo.toml")).unwrap();
+
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::symlink;
+        symlink(
+            target_dir.join("Cargo.toml"),
+            project.root().join("Cargo.toml"),
+        )
+        .unwrap();
+    }
+
+    #[cfg(windows)]
+    {
+        use std::os::windows::fs::symlink_file;
+        symlink_file(
+            target_dir.join("Cargo.toml"),
+            project.root().join("Cargo.toml"),
+        )
+        .unwrap();
+    }
+
+    project.cargo("add test-dep").run();
+
+    // Current behavior: symlink is NOT preserved (gets replaced)
+    assert!(!project.root().join("Cargo.toml").is_symlink());
+}
--- a/tests/testsuite/cargo_add/mod.rs
+++ b/tests/testsuite/cargo_add/mod.rs
@@ -145,6 +145,7 @@ mod script_bare;
 mod script_frontmatter;
 mod script_shebang;
 mod sorted_table_with_dotted_item;
+mod symlink;
 mod target;
 mod target_cfg;
 mod unknown_inherited_feature;


tests/testsuite/cargo_add/symlink.rs
weihanglo, line 1: nit: we prefer contributions split tests into the first commit, with them passing.  That way the fix commit will show the change in behavior by how the test changes.  This helps test the test and provides an illustration of the PR description for the reviewer and the community at large.

We linked to examples in the [contrib docs explaining this](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request)

RaghavenderSingh: Thanks for the guidance. I've restructured the PR.

weihanglo: Note that the first commit is best to pass CI. That means you'll assert the problematic behavior. The "fix" commit then contains both the fix, and the change in test. See also [this PR](https://github.com/rust-lang/cargo/pull/15260) as an example.

In addition, your test is a [functional test](https://doc.crates.io/contrib/tests/writing.html#functional-tests), but is placed along with other [UI tests](https://doc.crates.io/contrib/tests/writing.html#ui-tests). Is it possible to make it a pure file-based UI test, and change the test name to `case` in order to follow the convention?

(It might not work because cross-platform symlinks are tricky, and git is not good at handling it.)

tests/testsuite/cargo_add/symlink.rs
```diff
@@ -0,0 +1,51 @@
...
+        use std::os::windows::fs::symlink_file;
+        symlink_file(
+            target_dir.join("Cargo.toml"),
+            project.root().join("Cargo.toml"),
+        )
+        .unwrap();
+    }
+
+    // Add a dependency
+    project.cargo("add rand").run();
```
weihanglo, line 47: This is pulling an exteranl dependency from crates.io. We avoid that in cargo's testsuite and instead host a [local registry fixure](https://doc.rust-lang.org/1.85.0/nightly-rustc/cargo_test_support/registry/index.html), like

* https://github.com/rust-lang/cargo/blob/ab1463d632528e39daf35f263e10c14cbe590ce8/tests/testsuite/rename_deps.rs?plain=1#L12
* https://github.com/rust-lang/cargo/blob/ab1463d632528e39daf35f263e10c14cbe590ce8/tests/testsuite/cargo_add/features/mod.rs?plain=1#L10-L16

tests/testsuite/cargo_add/symlink.rs
weihanglo: cc @epage, any thoughts on how to organize these?

src/cargo/util/toml_mut/manifest.rs
```diff
@@ -297,6 +297,7 @@ impl LocalManifest {
         })
     }
 
+    /// Write changes back to the file.
     /// Write changes back to the file.
```
epage, line 301: Why the double comment?

src/cargo/util/toml_mut/manifest.rs
```diff
@@ -323,9 +324,16 @@ impl LocalManifest {
...
-    }
+        // Check if the path is a symlink and follow it if it is
+        let actual_path = if self.path.is_symlink() {
+            std::fs::read_link(&self.path)?
+        } else {
+            self.path.clone()
+        };
 
+        // Write to the actual target path instead of the symlink
+        cargo_util::paths::write_atomic(&actual_path, new_contents_bytes)
```
epage, line 335: Should this be put in `write_atomic`?

tests/testsuite/cargo_add/symlink.rs
epage: This mixing of functional with UI tests seems inline with how we do it elesewhere

src/cargo/util/toml_mut/manifest.rs
epage, line 1: This commit does not change the test.  Does the previous commit fail `cargo test` or is it testing the wrong thing?  Each commit should be atomic which includes passing tests.  When we ask for a test commit to be split out first, it is to show the current behavior.  The commit with the fix will also change the test and that diff shows how behavior changes.

src/cargo/util/toml_mut/manifest.rs
```diff
@@ -323,9 +323,9 @@ impl LocalManifest {
         };
         let new_contents_bytes = raw.as_bytes();
 
+        // The symlink handling is now done in write_atomic
```
epage, line 326: "now", it wasn't done before.  Is this comment needed?

src/cargo/util/toml_mut/manifest.rs
```diff
@@ -323,9 +323,9 @@ impl LocalManifest {
...
         cargo_util::paths::write_atomic(&self.path, new_contents_bytes)
     }
-
```
epage, line 328: Why is this line removed

src/cargo/util/toml_mut/manifest.rs
epage: Note that this is still not addressed.

crates/cargo-util/src/paths.rs
```diff
@@ -220,10 +228,9 @@ pub fn write_atomic<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Res
         tmp.as_file().set_permissions(perms)?;
     }
 
-    tmp.persist(path)?;
+    tmp.persist(&actual_path)?;
     Ok(())
 }
-
```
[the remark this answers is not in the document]
RaghavenderSingh: Hi @epage,
I implemented symlink preservation in write_atomic().
preview

The test now shows how cargo add handles symlinks correctly. Let me know what you think.

src/cargo/util/toml_mut/manifest.rs
epage: From https://github.com/rust-lang/cargo/pull/15281#discussion_r2014385665

> Hi @epage, I implemented symlink preservation in write_atomic(). preview
> 
> The test now shows how cargo add handles symlinks correctly. Let me know what you think.

epage: So the test fails for the first commit and then passes for the second commit?  In that case, these commits are not atomic, as in they do not stand on their own with every commit passing tests.

We [ask for the splitting of test/fix commits](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request) so that the second commit shows how the behavior changes by how the test body changes to ensure it still passes.  This is a clear way of communicating the way the behavior changed.

crates/cargo-util/src/paths.rs
```diff
@@ -190,14 +190,22 @@ pub fn write<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()>
...
+/// If the path is a symlink, it will follow the symlink and write to the actual target.
 pub fn write_atomic<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()> {
     let path = path.as_ref();
 
+    // Check if the path is a symlink and follow it if it is
+    let actual_path = if path.is_symlink() {
+        std::fs::read_link(path)?
+    } else {
+        path.to_path_buf()
+    };
```
weihanglo, line 202: To avoid unnecessary allocations. And we can continue call it `path` so we don't need to change other lines.
(It is good that shadowing is allowed in Rust)

```suggestion
    let path = if path.is_symlink() {
        &std::fs::read_link(path)?
    } else {
        path
    };
```

src/cargo/util/toml_mut/manifest.rs
weihanglo: @RaghavenderSingh would you mind doing what epage suggested?

weihanglo (changes requested):
This may need a rebase onto master to pass CI.

weihanglo (changes requested):
Could you squah the last two commits into the second commit? They are unnecessary.

Issue #13313, opened 2024-01-17
Delegate `Cargo.toml` style guide to style team?


joshtriplett:
@rust-lang/cargo previously discussed the possibility of formatting `Cargo.toml`. There are open questions around that, such as whether `rustfmt` or `cargo fmt` is responsible for doing the formatting.

This issue is to determine who has the responsibility for writing and maintaining the style guide for `Cargo.toml`. The old style team (years ago) did write a style guide for `Cargo.toml` manifests, but there was never any automated tooling to format manifests according to that style, so we don't want to presume that that style is well-established.

When this came up previously, `cargo` had at least one concern with the current style guidance, which I'm addressing with this PR: https://github.com/rust-lang/rust/pull/120072 . Currently reviewing other potential concerns, and will post fixes for those as well, *before* the style guide gets used for any form of automatic formatting.

With my Cargo team hat on, I would propose that the Cargo team delegate responsibility for ~~bikeshedding~~ writing and maintaining the style guide for `Cargo.toml` manifest to the style team, with the understanding that:
- The Cargo team can always raise objections if there's style guidance they have concerns with
- The Cargo team can withdraw the delegation if they have unresolved concerns with how that work is being done

joshtriplett:
Starting an FCP for this, but also nominating it so that we can discuss it in our next meeting.

@rfcbot merge

epage:
@rfcbot concern verify

This seems to almost complete delegate ownership.  I think I would feel more comfortable delegating the task of definition, setting parameters and having a final T-cargo FCP, and as we get things settled, moving on to the more complete delegation mentioned here.

joshtriplett:
@epage So, the initial delegation would be "update the style guide to address all of T-cargo's concerns and we'll review it, FCP it, and consider delegating ongoing maintenance at that time"?

epage:
Thats roughly my thought.  Not required we go that way but just a little concerned with the fact that the previous style guide had no teeth (no rustfmt) and had weird suggestions that ran counter to community norms so it seems like it'd be good to be more in the loop so we make sure we start in the right direction with a style guide that is getting teeth.

PR #15285, opened 2025-03-09
docs(reference): sort manifest package section keys per style guide

scop:
https://doc.rust-lang.org/nightly/style-guide/cargo.html

> Put the `[package]` section at the top of the file; put the `name` and `version` keys in that order at the top of that section, followed by the remaining keys other than `description` in order, followed by the description at the end of that section.

Sorting only, no content changes.

### DIFF
--- a/src/doc/src/reference/manifest.md
+++ b/src/doc/src/reference/manifest.md
@@ -11,31 +11,31 @@ Every manifest file consists of the following sections:
   * [`name`](#the-name-field) --- The name of the package.
   * [`version`](#the-version-field) --- The version of the package.
   * [`authors`](#the-authors-field) --- The authors of the package.
-  * [`edition`](#the-edition-field) --- The Rust edition.
-  * [`rust-version`](rust-version.md) --- The minimal supported Rust version.
-  * [`description`](#the-description-field) --- A description of the package.
-  * [`documentation`](#the-documentation-field) --- URL of the package documentation.
-  * [`readme`](#the-readme-field) --- Path to the package's README file.
-  * [`homepage`](#the-homepage-field) --- URL of the package homepage.
-  * [`repository`](#the-repository-field) --- URL of the package source repository.
-  * [`license`](#the-license-and-license-file-fields) --- The package license.
-  * [`license-file`](#the-license-and-license-file-fields) --- Path to the text of the license.
-  * [`keywords`](#the-keywords-field) --- Keywords for the package.
-  * [`categories`](#the-categories-field) --- Categories of the package.
-  * [`workspace`](#the-workspace-field) --- Path to the workspace for the package.
-  * [`build`](#the-build-field) --- Path to the package build script.
-  * [`links`](#the-links-field) --- Name of the native library the package links with.
-  * [`exclude`](#the-exclude-and-include-fields) --- Files to exclude when publishing.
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/manifest.md
```diff
@@ -11,31 +11,31 @@ Every manifest file consists of the following sections:
   * [`name`](#the-name-field) --- The name of the package.
   * [`version`](#the-version-field) --- The version of the package.
   * [`authors`](#the-authors-field) --- The authors of the package.
-  * [`edition`](#the-edition-field) --- The Rust edition.
-  * [`rust-version`](rust-version.md) --- The minimal supported Rust version.
-  * [`description`](#the-description-field) --- A description of the package.
+  * [`autobenches`](cargo-targets.md#target-auto-discovery) --- Disables bench auto discovery.
```
weihanglo, line 14: For example, personally, `autobenches` is definitely not as important as other fields like `edition`, `rust-version`, and `description`. This might make the reading experience slightly worse for readers who go through it from top to bottom.

src/doc/src/reference/manifest.md
weihanglo, line 1: Thanks for the pull request!

This is a good attempt, though the ownership of the style of Carg.toml doesn't seem clear to me right now.

* https://github.com/rust-lang/cargo/issues/13313

As a consequence, we see some issues like this one:

* https://github.com/rust-lang/rust/pull/120072

I am not sure if the style guide is settled, so maybe we should postpone this until things get clearer?

src/doc/src/reference/manifest.md
```diff
@@ -135,47 +135,84 @@ authors = ["Graydon Hoare", "Fnu Lnu <no-reply@rust-lang.org>"]
 This field is surfaced in package metadata and in the `CARGO_PKG_AUTHORS`
 environment variable within `build.rs` for backwards compatibility.
 
-### The `edition` field
+## The `[badges]` section
```
weihanglo, line 138: Lexicographic order makes sense for a reference-like documentation. I believe we did it for `.cargo/config.toml` docs. However, I feel like fields in configuration are mostly equally-important but in Cargo.toml they are not. Moving an unused field upfront doesn't seem too good. 

Anyway, this is not a suggestion for you to apply. I am just sharing an example that I don't fully agree with the style guide.

weihanglo:
cc @joshtriplett, what is your thought on this?
(as you're a member of both cargo and style teams)

joshtriplett:
@weihanglo This portion of styling has not been finalized and does not have consensus with the Cargo team, so it's not ready to be acted on. The fact that this prompted a PR means we should remove this portion of the style guide until we have a new version that has consensus with Cargo.

src/doc/src/reference/manifest.md
scop: FWIW, I agree with this, the ordering in the cargo docs seems better to me.

scop:
JFYI there's tooling in existence that is already making use of the style guide sort order, which kind of led to me opening this PR: https://github.com/dprint/dprint-plugin-toml/issues/29

This is not to say I'd personally lean towards the style guide on this (I don't -- the one in cargo docs seem "better" to me), but as said, JFYI.

weihanglo:
Since this is kinda waiting for t-style to update the guide, I'll mark it as draft and add a label. Thanks for your contribution!

@rustbot label +S-blocked-external

PR #15287, opened 2025-03-09
Allow `term.progress.when` to default

Gordon01:

The config section `term.progress` contains a collection of options (currently only `when` and `width`). If only `width` is set, Cargo fails with the error: "missing field `when`." This behavior may be surprising and inconvenient for users who want to set the `CARGO_TERM_PROGRESS_WIDTH` environment variable.

The relevant test, `bad_progress_config_missing_when`, has been replaced with `default_progress_is_auto`.

https://github.com/rust-lang/cargo/pull/14615

From https://github.com/rust-lang/cargo/pull/14615#discussion_r1889426752

> I don't recall any particular decisions made regarding bad_progress_config_missing_when. I personally would not object to changing that to allow it to pass. I can't think of a reason it would be required to specify when with width.

### DIFF
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -2830,6 +2830,7 @@ pub struct TermConfig {
 #[derive(Debug, Default, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct ProgressConfig {
+    #[serde(default)]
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
Looks pretty reasonable. Thanks!

Gordon01:
Amazing!

PR #15288, opened 2025-03-10
feat: add completions for add --path

NOOMA-42:

This PR introduces auto-completion for the `cargo add --path` option. When a user types `cargo add --path` and presses the TAB key, the system will automatically detect directories in the current path and suggestion prioritizes directory containing a Cargo.toml file.

In the terminal, type `cargo add --path`
Press the TAB key.
You should see suggestions only for directories in the current path that contain a Cargo.toml file being prioritized, such as ./src/ (if ./src/Cargo.toml exists).

https://github.com/user-attachments/assets/0620ae5e-abca-425c-ad9a-7004929bb98d

### DIFF
--- a/src/bin/cargo/commands/add.rs
+++ b/src/bin/cargo/commands/add.rs
@@ -100,7 +100,11 @@ Example uses:
                 .value_name("PATH")
                 .help("Filesystem path to local crate to add")
                 .group("selected")
-                .conflicts_with("git"),
+                .conflicts_with("git")
+                .add(clap_complete::engine::ArgValueCompleter::new(
+                    clap_complete::engine::PathCompleter::any()
+                        .filter(|path| path.join("Cargo.toml").exists()),
+                )),
             clap::Arg::new("base")
                 .long("base")
                 .action(ArgAction::Set)


weihanglo (approved):
Look nice! Thanks!

I wonder if we want to de-prioritize hidden directories, but that can leave for future discussion.

src/bin/cargo/commands/add.rs
```diff
@@ -100,7 +100,11 @@ Example uses:
                 .value_name("PATH")
                 .help("Filesystem path to local crate to add")
                 .group("selected")
-                .conflicts_with("git"),
+                .conflicts_with("git")
+                .add(clap_complete::engine::ArgValueCompleter::new(
+                    clap_complete::engine::PathCompleter::any()
+                        .filter(|path| path.join("Cargo.toml").exists()),
+                )),
```
epage, line 107: For future changes, should we refactor out these `--path` completers into https://github.com/rust-lang/cargo/blob/master/src/cargo/util/command_prelude.rs ?  We have two already and I believe there is a third *and* we'll be wanting to add more context to these

PR #15290, opened 2025-03-10
Replace unmaintained humantime crate with jiff

oherrala:
The crate [`humantime`](https://crates.io/crates/humantime) appears to be unmaintained. There's open PR in RustSec's advisory-db about this: https://github.com/rustsec/advisory-db/pull/2249

The crates [`clap`](https://crates.io/crates/clap) and [`env_logger`](https://crates.io/crates/env_logger) have already made the switch from `humantime` to [`jiff`](https://crates.io/crates/jiff):

 * https://github.com/clap-rs/clap/pull/5944
 * https://github.com/rust-cli/env_logger/pull/352

The `jiff` crate is already dependency on `cargo` via `gix` (albeit old 0.1 version, but that's probably fixed in [next gix release](https://github.com/GitoxideLabs/gitoxide/commit/3ae99a42f51cd2d6c55c6abbd1ead86bf8bf2e1f)):

```
jiff v0.1.29
└── gix-date v0.9.3
    ├── gix v0.70.0
    │   └── cargo v0.88.0 (/Users/oherrala/rust/cargo)
```

This PR shouldn't have any functional change to cargo itself.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,11 +55,11 @@ hex = "0.4.3"
 hmac = "0.12.1"
 home = "0.5.11"
 http-auth = { version = "0.1.10", default-features = false }
-humantime = "2.1.0"
 ignore = "0.4.23"
 im-rc = "15.1.0"
 indexmap = "2.7.1"
 itertools = "0.14.0"
+jiff = { version = "0.2.3", default-features = false, features = [ "std" ] }
 jobserver = "0.1.32"
 lazycell = "1.3.0"
 libc = "0.2.169"
@@ -176,11 +176,11 @@ hex.workspace = true
 hmac.workspace = true
 home.workspace = true
 http-auth.workspace = true
-humantime.workspace = true
 ignore.workspace = true
 im-rc.workspace = true
 indexmap.workspace = true
 itertools.workspace = true
+jiff.workspace = true
 jobserver.workspace = true
 lazycell.workspace = true
 libgit2-sys.workspace = true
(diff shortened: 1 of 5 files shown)


epage:
```diff
- [DIRTY] foo v0.0.1 ([ROOT]/foo): the dependency foo was rebuilt ([TIME_DIFF_AFTER_LAST_BUILD])
+ [DIRTY] foo v0.0.1 ([ROOT]/foo): the dependency foo was rebuilt (1741638048.881853669s, PT1S after last build at 1741638047.743836231s)
```

Our regex at https://github.com/rust-lang/cargo/blob/340d1239a89d0030fd547beae74d87a1dccb8bd6/crates/cargo-test-support/src/compare.rs#L243-L251 is not matching anymore to redact this test output

Cargo.toml
```diff
@@ -55,11 +55,11 @@ hex = "0.4.3"
 hmac = "0.12.1"
 home = "0.5.11"
 http-auth = { version = "0.1.10", default-features = false }
-humantime = "2.1.0"
 ignore = "0.4.23"
 im-rc = "15.1.0"
 indexmap = "2.7.1"
 itertools = "0.14.0"
+jiff = { version = "0.2.3", default-features = false, features = [ "std" ] }
```
epage, line 62: If we're adopting `jiff`, we may want to consider consolidating on it more generally (ie move off of `time`)

Haven't considered whether that is blocking for this

oherrala: Since `jiff` is already in dependency tree via `gix` this PR shouldn't have big change in build nor binary? Dropping `time` in favor of `jiff` could be done in another PR. Quick grep shows `time` is used in six files and mostly `OffsetDateTime` and RFC3339 time formatting.

epage (approved):

yhx-12243:
`humantime` seems resurrected now (1 hour ago).

Issue #15289, opened 2025-03-10
Cargo reference build script example fails to compile


ostkamp:
### Problem

https://github.com/rust-lang/cargo/blob/42b6ae87b1faf2656b95040bf5337e6d41d5c08d/src/doc/src/reference/build-script-examples.md?plain=1#L255 contains the following lines:

```
extern { fn hello(); }

fn main() {
    unsafe { hello(); }
}
```

This does not or no longer compile with rust 1.85.0:
```
   Compiling hello-world-from-c v0.1.0 (/home/ostkamp/rust/projects/14_More_about_Cargo_and_Crates.io/hello-world-from-c)
error: extern blocks must be unsafe
 --> src/main.rs:3:1
  |
3 | extern { fn hello(); }
  | ^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `hello-world-from-c` (bin "hello-world-from-c") due to 1 previous error

```

Suggested solution: Change the extern line as outlined below:

```
unsafe extern "C" { fn hello(); }

// Note the lack of the `#[link]` attribute. We’re delegating the responsibility
// of selecting what to link over to the build script rather than hard-coding
// it in the source file.

fn main() {
    unsafe { hello(); }
}
```

### Version

```text

```

weihanglo:
Thanks for the report. Feel free to update the docs Rust 2024 edition flavor.

PR #15294, opened 2025-03-10
docs(refs): Add `unsafe` to `extern` while using build scripts in Cargo Book

namanlp:

Since `rust 1.85.0`, It is no longer sufficient to use the `unsafe` keyword only at the call site. It must also be applied to the `extern` statement, hence, Cargo Book must be updated

- closes https://github.com/rust-lang/cargo/issues/15289

and also update some more related content

Run `mdbook build`. The `unsafe extern` should be seen at `cargo/reference/build-script-examples.html#linking-to-system-libraries` , `cargo/reference/build-script-examples.html#building-a-native-library` , `cargo/reference/semver.html#repr-c-shuffle` , `cargo/reference/semver.html#repr-c-remove` and `cargo/reference/semver.html#repr-transparent-remove`

### DIFF
--- a/src/doc/src/reference/build-script-examples.md
+++ b/src/doc/src/reference/build-script-examples.md
@@ -252,7 +252,7 @@ void hello() {
 // Note the lack of the `#[link]` attribute. We’re delegating the responsibility
 // of selecting what to link over to the build script rather than hard-coding
 // it in the source file.
-extern { fn hello(); }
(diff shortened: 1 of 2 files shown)


epage (approved):
Thanks!

epage:
I wonder if we should pull the example code out into a crate that we make a workspace member, like over at https://github.com/rust-cli/book

PR #15295, opened 2025-03-11
cargo-credential-libsecret: load libsecret only once

### LINKED ISSUE #15603: `cargo login` + libsecret provider ⇒ hangs
### Problem

I'm trying to `cargo login` to a custom registry and store the token in the GNOME keyring via the libsecret provider, but it just gets stuck forever.

### Steps

1. configure cargo to use the libsecret provider:

    ~/.cargo/config.toml:
    ```
    [registry]
    global-credential-providers = ["cargo:token", "cargo:libsecret"]
    ```

2. `CARGO_REGISTRIES_TEST1_INDEX=sparse+https://charts.jfrog.io/artifactory/api/cargo/xxx/index/ cargo login --registry test1`
    (this is some fairly random URL that happens to trigger the issue but it's not our private jfrog instance obviously)
3. reply whatever (e.g. "x") to the "please paste the token for test1 below" prompt

Now cargo gets stuck after printing the following errors:
```
(process:19648): GLib-GObject-CRITICAL **: 15:31:30.977: cannot register existing type 'SecretService'
(process:19648): GLib-GObject-CRITICAL **: 15:31:30.977: cannot add private field to invalid (non-instantiatable) type '<invalid>'
(process:19648): GLib-GObject-CRITICAL **: 15:31:30.977: g_type_add_interface_static: assertion 'G_TYPE_IS_INSTANTIATABLE (instance_type)' failed
(process:19648): GLib-GObject-CRITICAL **: 15:31:30.977: g_type_add_interface_static: assertion 'G_TYPE_IS_INSTANTIATABLE (instance_type)' failed
(process:19648): GLib-GObject-CRITICAL **: 15:31:30.977: cannot register existing type 'SecretBackend'
(process:19648): GLib-GObject-CRITICAL **: 15:31:30.977: g_type_interface_add_prerequisite: assertion 'G_TYPE_IS_INTERFACE (interface_type)' failed
(process:19648): GLib-GObject-CRITICAL **: 15:31:30.977: g_type_interface_add_prerequisite: assertion 'G_TYPE_IS_INTERFACE (interface_type)' failed
(process:19648): GLib-CRITICAL **: 15:31:30.977: g_once_init_leave_pointer: assertion 'result != 0' failed
(process:19648): GLib-GObject-CRITICAL **: 15:31:30.977: g_type_add_interface_static: assertion 'G_TYPE_IS_INSTANTIATABLE (instance_type)' failed
```

### Possible Solution(s)

I believe this is because the libsecret-1.so.0 library gets loaded and unloaded multiple times (once for the lookup, another time for the store):
https://github.com/rust-lang/cargo/blob/c8d15a37e36c2dfb5fdff8942b71858b4a5756c2/credential/cargo-credential-libsecret/src/lib.rs#L122
but this isn't supported: https://mail.gnome.org/archives/gtk-list/2017-February/msg00006.html

Using something like [LazyLock](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) to load the library should resolve the issue.

There's a workaround — LD_PRELOAD can be used to prevent the dynamic library from getting unloaded:
`LD_PRELOAD=/lib/aarch64-linux-gnu/libsecret-1.so.0 cargo login --registry`

### Notes

I think this only gets stuck on arm64. I've seen the glib errors on x86_64 as well but it didn't get stuck. `gdb` tells me it's stuck in some futex/condition variable, but I don't think it's especially relevant as the root cause is (imo) well understood.

### Version

```text
cargo 1.87.0 (99624be96 2025-05-06)
release: 1.87.0
commit-hash: 99624be96e9d213b0e9b1e36451271f24e4a41d8
commit-date: 2025-05-06
host: aarch64-unknown-linux-gnu
libcurl: 8.12.1-DEV (sys:0.4.80+curl-8.12.1 vendored ssl:OpenSSL/3.4.1)
ssl: OpenSSL 3.4.1 11 Feb 2025
os: Debian 13.0.0 (trixie) [64-bit]
```

biglben:
Previously, libsecret was repeatedly loaded and unloaded—at least twice during cargo login (once for action get and once for action login). This caused issues where calls could hang indefinitely due to glib failing to clean up properly (some threads still running after unloading) and generating numerous error messages:
```
cargo login --registry ownreg
    Updating `ownreg` index
please paste the token for ownreg below


(process:100727): GLib-GObject-CRITICAL **: 08:59:54.568: cannot register existing type 'SecretService'

(process:100727): GLib-GObject-CRITICAL **: 08:59:54.568: cannot add private field to invalid (non-instantiatable) type '<invalid>'

(process:100727): GLib-GObject-CRITICAL **: 08:59:54.568: g_type_add_interface_static: assertion 'G_TYPE_IS_INSTANTIATABLE (instance_type)' failed

(process:100727): GLib-GObject-CRITICAL **: 08:59:54.568: g_type_add_interface_static: assertion 'G_TYPE_IS_INSTANTIATABLE (instance_type)' failed

(process:100727): GLib-GObject-CRITICAL **: 08:59:54.568: cannot register existing type 'SecretBackend'

(process:100727): GLib-GObject-CRITICAL **: 08:59:54.568: g_type_interface_add_prerequisite: assertion 'G_TYPE_IS_INTERFACE (interface_type)' failed

(process:100727): GLib-GObject-CRITICAL **: 08:59:54.568: g_type_interface_add_prerequisite: assertion 'G_TYPE_IS_INTERFACE (interface_type)' failed

(process:100727): GLib-CRITICAL **: 08:59:54.568: g_once_init_leave_pointer: assertion 'result != 0' failed

(process:100727): GLib-GObject-CRITICAL **: 08:59:54.568: g_type_add_interface_static: assertion 'G_TYPE_IS_INSTANTIATABLE (instance_type)' failed
```

Now, libsecret is stored in a OnceLock, ensuring it is only loaded once, preventing unnecessary unload/reload cycles.

Fixes #15603

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ blake3 = "1.5.5"
 build-rs = { version = "0.3.1", path = "crates/build-rs" }
 cargo = { path = "" }
 cargo-credential = { version = "0.4.2", path = "credential/cargo-credential" }
-cargo-credential-libsecret = { version = "0.4.15", path = "credential/cargo-credential-libsecret" }
+cargo-credential-libsecret = { version = "0.5.0", path = "credential/cargo-credential-libsecret" }
 cargo-credential-macos-keychain = { version = "0.4.15", path = "credential/cargo-credential-macos-keychain" }
 cargo-credential-wincred = { version = "0.4.15", path = "credential/cargo-credential-wincred" }
 cargo-platform = { path = "crates/cargo-platform", version = "0.3.0" }
--- a/credential/cargo-credential-libsecret/src/lib.rs
+++ b/credential/cargo-credential-libsecret/src/lib.rs
@@ -119,7 +119,7 @@ mod linux {
         }
     }
 
-    impl Credential for LibSecretCredential {
+    impl Credential for &LibSecretCredential {
         fn perform(
             &self,
             registry: &RegistryInfo<'_>,
--- a/credential/cargo-credential-libsecret/src/lib.rs
+++ b/credential/cargo-credential-libsecret/src/lib.rs
@@ -83,7 +83,9 @@ mod linux {
         ...
     ) -> *mut gchar;
 
-    pub struct LibSecretCredential;
+    pub struct LibSecretCredential {
+        libsecret: Library,
+    }
 
     fn label(index_url: &str) -> CString {
         CString::new(format!("cargo-registry:{}", index_url)).unwrap()
@@ -105,6 +107,18 @@ mod linux {
         }
     }
 
+    impl LibSecretCredential {
+        pub fn new() -> Result<LibSecretCredential, Error> {
+            // Dynamically load libsecret to avoid users needing to install
+            // additional -dev packages when building this provider.
+            let libsecret = unsafe { Library::new("libsecret-1.so.0") }.context(
+                "failed to load libsecret: try installing the `libsecret` \
+                or `libsecret-1-0` package with the system package manager",
+            )?;
+            Ok(Self { libsecret })
+        }
+    }
+
     impl Credential for LibSecretCredential {
         fn perform(
             &self,
@@ -112,24 +126,22 @@ mod linux {
             action: &Action<'_>,
             _args: &[&str],
         ) -> Result<CredentialResponse, Error> {
-            // Dynamically load libsecret to avoid users needing to install
-            // additional -dev packages when building this provider.
-            let lib;
             let secret_password_lookup_sync: Symbol<'_, SecretPasswordLookupSync>;
             let secret_password_store_sync: Symbol<'_, SecretPasswordStoreSync>;
             let secret_password_clear_sync: Symbol<'_, SecretPasswordClearSync>;
             unsafe {
-                lib = Library::new("libsecret-1.so.0").context(
-                    "failed to load libsecret: try installing the `libsecret` \
-                    or `libsecret-1-0` package with the system package manager",
-                )?;
-                secret_password_lookup_sync = lib
+                secret_password_lookup_sync = self
+                    .libsecret
                     .get(b"secret_password_lookup_sync\0")
                     .map_err(Box::new)?;
-                secret_password_store_sync =
-                    lib.get(b"secret_password_store_sync\0").map_err(Box::new)?;
-                secret_password_clear_sync =
-                    lib.get(b"secret_password_clear_sync\0").map_err(Box::new)?;
+                secret_password_store_sync = self
+                    .libsecret
+                    .get(b"secret_password_store_sync\0")
+                    .map_err(Box::new)?;
+                secret_password_clear_sync = self
+                    .libsecret
+                    .get(b"secret_password_clear_sync\0")
+                    .map_err(Box::new)?;
             }
 
             let index_url_c = CString::new(registry.index_url).unwrap();
--- a/src/cargo/util/auth/mod.rs
+++ b/src/cargo/util/auth/mod.rs
@@ -545,7 +545,7 @@ fn credential_action(
             #[cfg(target_os = "macos")]
             "cargo:macos-keychain" => Box::new(cargo_credential_macos_keychain::MacKeychain {}),
             #[cfg(target_os = "linux")]
-            "cargo:libsecret" => Box::new(cargo_credential_libsecret::LibSecretCredential {}),
+            "cargo:libsecret" => Box::new(cargo_credential_libsecret::LibSecretCredential::new()?),
             name if BUILT_IN_PROVIDERS.contains(&name) => {
                 Box::new(cargo_credential::UnsupportedCredential {})
             }
--- a/src/cargo/util/auth/mod.rs
+++ b/src/cargo/util/auth/mod.rs
@@ -508,6 +508,27 @@ static BUILT_IN_PROVIDERS: &[&'static str] = &[
     "cargo:libsecret",
 ];
 
+/// Retrieves a cached instance of `LibSecretCredential`.
+/// Must be cached to avoid repeated load/unload cycles, which are not supported by `glib`.
+#[cfg(target_os = "linux")]
+fn get_credential_libsecret(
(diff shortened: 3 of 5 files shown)


epage:
The main thing I'm wondering is if this is the right layer of caching.

Alternatively, we could lazy load these fields in `LibSecretCredential` and use a `OnceLock` on that in the cargo side of the code.  Or we could skip `OnceLock` and have a cache on `GlobalContext` which would give users control over lifetimes.

A potential way to side step any of that is for us to not load the library multiple times.  It looks like during `login()` we call `registry()` to get `Registry::new_handle()` so we can call `registry.host()` .  We might be able to split `registry()` into `registry()` and `registry_cfg()` and use the latter in `login()`, avoiding looking up the token just to throw it away.

biglben:
Thanks for your feedback.

> The main thing I'm wondering is if this is the right layer of caching.
> 
> Alternatively, we could lazy load these fields in `LibSecretCredential` and use a `OnceLock` on that in the cargo side of the code. Or we could skip `OnceLock` and have a cache on `GlobalContext` which would give users control over lifetimes.

I will check `GlobalContext`.
 Could you clarify your approach to lazy loading these fields in `LibSecretCredential`? Are you suggesting storing `Library` in `LibSecretCredential` and using a `OnceLock<LibSecretCredential>` in `credential_action`, in the `auth` module itself or somewhere else?

> 
> A potential way to side step any of that is for us to not load the library multiple times. It looks like during `login()` we call `registry()` to get `Registry::new_handle()` so we can call `registry.host()` . We might be able to split `registry()` into `registry()` and `registry_cfg()` and use the latter in `login()`, avoiding looking up the token just to throw it away.

This would be a good optimization, but it wouldn't fully resolve the issue. I discovered that libsecret is also loaded multiple times when multiple registries are present.

By the way, should i open a issue too?

epage:
Let's keep things simple for the first pass and put the `OnceLock` inside of `cargo`-the-lib.  The `GlobalContext` change can come later.

weihanglo:
@rustbot author

Let us know if you have any question!

biglben:
@rustbot ready

Sorry for the delay, I was busy with a lot of other stuff and forgot about this.

src/cargo/util/auth/mod.rs
```diff
@@ -508,6 +508,26 @@ static BUILT_IN_PROVIDERS: &[&'static str] = &[
...
+            let _ = LIBSECRET.set(
+                unsafe { libloading::Library::new("libsecret-1.so.0") }.context(
+                    "failed to load libsecret: try installing the `libsecret` \
+                or `libsecret-1-0` package with the system package manager",
+                )?,
+            );
+            Ok(LIBSECRET.get().unwrap())
+        }
+    }
+}
```
epage: I had envisioned putting `LibSecretCredential` inside of a `OnceLock`.  How come you instead put `libloading::Library`?

biglben: Ah, sorry, I misunderstood. I thought the idea is that libsecret could also be used in other parts of cargo but as i look more into it, that does not really makes sense.
I will change it

src/cargo/util/auth/mod.rs
```diff
@@ -508,6 +508,25 @@ static BUILT_IN_PROVIDERS: &[&'static str] = &[
...
+fn get_credential_libsecret(
+) -> CargoResult<&'static cargo_credential_libsecret::LibSecretCredential> {
+    static CARGO_CREDENTIAL_LIBSECRET: std::sync::OnceLock<
+        cargo_credential_libsecret::LibSecretCredential,
+    > = std::sync::OnceLock::new();
+    // unfortunately `get_or_try_init` is not yet stable
+    match CARGO_CREDENTIAL_LIBSECRET.get() {
+        Some(lib) => Ok(lib),
+        None => {
+            let _ = CARGO_CREDENTIAL_LIBSECRET
```
epage, line 525: Can we at least log something if this happens?

src/cargo/util/auth/mod.rs
```diff
@@ -508,6 +508,25 @@ static BUILT_IN_PROVIDERS: &[&'static str] = &[
...
+                .set(cargo_credential_libsecret::LibSecretCredential::new()?);
```
epage, line 526: This can cause us to double load on a race condition but Cargo is generally single threaded so this should be fine for now.  Famous last words...

credential/cargo-credential-libsecret/src/lib.rs
```diff
@@ -83,7 +83,9 @@ mod linux {
         ...
     ) -> *mut gchar;
 
-    pub struct LibSecretCredential;
+    pub struct LibSecretCredential {
```
epage, line 86: nit: its a big help for reviewers to split up commits.  In this case, I'd do
- Version bump
- Switch to `LibSecretCredential` owning `libsecret`
- Add the `OnceLock`

I mention this more for future reference to the point that I wasn't going to mention it if there wasn't other feedback to avoid you doing extra churn on this without a reason.

src/cargo/util/auth/mod.rs
biglben: If you want, I can log this. As I understand it, loading the library twice just increments the reference counter and decrements it when `drop` is called. So I don't see a problem, even if Cargo becomes multithreaded.

Cargo.toml
```diff
@@ -185,6 +185,7 @@ jiff.workspace = true
 jobserver.workspace = true
 lazycell.workspace = true
 libgit2-sys.workspace = true
+libloading.workspace = true
```
epage: This isn't needed anymore, right?

biglben: Nice catch, thank you

credential/cargo-credential-libsecret/src/lib.rs
```diff
@@ -119,7 +119,7 @@ mod linux {
         }
     }
 
-    impl Credential for LibSecretCredential {
+    impl Credential for &LibSecretCredential {
```
epage, line 122: Had missed this before the commit split, thanks!

Should still be good though.

epage (approved):

PR #15298, opened 2025-03-11
fix(run): Disambiguate bins from different packages that share a name

### LINKED ISSUE #13312: Error for `cargo run --bin {name}` should notice multiple packages have the same binary name
### Problem

```console
$ cargo run --bin strings/pgrx_embed
error: no bin target named `strings/pgrx_embed`.
Available bin targets:
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    cargo-pgrx
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx-version-updater
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
    pgrx_embed
```

This is obviously bad UX, as users can even now address these individual binaries in individual packages, despite having the same name, by using e.g. `cargo run -p strings --bin pgrx_embed`.

### Proposed Solution

While I am reconsidering the UX in this proposed changed to pgrx's tooling which would involve this diff, I would like it if cargo was able to distinguish binaries in different packages with the same name and at least emit a more useful error message.

### Notes

The reason for having a binary of the same name for all these packages is that several other "tooling is bad for this" problems are being worked around by various means, and this is one of the proposed workarounds that is better than other workarounds. The current workarounds involve dlopening a Rust programmer's cdylib in order to generate SQL based on the programmer's Rust code, this one would generate a binary that knows how to emit the SQL. This could all be omitted if Rust tooling simply had a more advanced facility for emitting user-defined metadata so that we could organize it later.

epage:

This builds on the work done in #15199 to improve target selection errors to also disambiguate when the same binary name is used in multiple packages.

This also makes the errors from #15199 more consistent with the rustc style guide and reduces code duplication.

Fixes #13312

This is a first pass and does not do the full `--package foo --bin bar` syntax.  I wanted to focus on the basic functionality before we iterated on it, e.g. investigating how well we can predict the CLI flags, how much noise they might add, etc.

### DIFF
--- a/tests/testsuite/run.rs
+++ b/tests/testsuite/run.rs
@@ -836,6 +836,11 @@ fn ambiguous_bin_name() {
         .with_status(101)
         .with_stderr_data(str![[r#"
 [ERROR] `cargo run` can run at most one executable, but multiple were specified
+[HELP] available targets:
+    bin `ambiguous` in package `crate1`
+    bin `ambiguous` in package `crate2`
+    bin `ambiguous` in package `crate3`
+    bin `ambiguous` in package `crate4`
 
 "#]])
         .run();
@@ -1014,6 +1019,9 @@ fn either_name_or_example() {
         .with_status(101)
         .with_stderr_data(str![[r#"
 [ERROR] `cargo run` can run at most one executable, but multiple were specified
+[HELP] available targets:
+    bin `a` in package `foo`
+    example `b` in package `foo`
 
 "#]])
         .run();
--- a/src/cargo/ops/cargo_run.rs
+++ b/src/cargo/ops/cargo_run.rs
@@ -1,4 +1,5 @@
 use std::ffi::OsString;
+use std::fmt::Write as _;
 use std::iter;
 use std::path::Path;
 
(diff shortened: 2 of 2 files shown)


tests/testsuite/run.rs
```diff
@@ -1014,6 +1019,9 @@ fn either_name_or_example() {
         .with_status(101)
         .with_stderr_data(str![[r#"
 [ERROR] `cargo run` can run at most one executable, but multiple were specified
+[HELP] available targets:
+    bin a in package foo
+    example b in package foo
```
weihanglo, line 1024: Do we want to quote the names? Unsure if it is more readable

```suggestion
    bin `a` in package foo
    example `b` in package foo
```

weihanglo (approved):

tests/testsuite/run.rs
epage: I expect we're very inconsistent about our quoting and would be good to decide when we should and try to clean things up.

PR #15301, opened 2025-03-12
feat: Add custom completer for cargo +<TAB> to complete toolchain name

nooma-42:

This PR introduces auto-completion for the `cargo +<TAB>` option. When a user types `cargo +<TAB>` and presses the TAB key, the system will automatically suggest toolchain in current device

1. In the terminal, type `cargo +<TAB>`
2. Press the TAB key.
3. You should see toolchain suggestions

https://github.com/user-attachments/assets/c8bd8a5e-e97d-4d0a-b284-1ed52734c4a2

### DIFF
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -691,7 +691,30 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
                 }))
             }).collect()
         })))
+        .add(clap_complete::engine::SubcommandCandidates::new(|| {
+            get_toolchains_from_rustup()
+                .into_iter()
+                .map(|t| clap_complete::CompletionCandidate::new(t))
+                .collect()
+        }))
         .subcommands(commands::builtin())
+}
+
+fn get_toolchains_from_rustup() -> Vec<String> {
+    let output = std::process::Command::new("rustup")
+        .arg("toolchain")
+        .arg("list")
+        .arg("-q")
+        .output()
+        .unwrap();
+
+    if !output.status.success() {
+        return vec![];
+    }
+
+    let stdout = String::from_utf8(output.stdout).unwrap();
+
+    stdout.lines().map(|line| format!("+{}", line)).collect()
 }
 
 #[test]


src/bin/cargo/cli.rs
```diff
@@ -690,10 +690,39 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
...
+                .filter(|t| OsStr::new(t).as_encoded_bytes().starts_with(partial.as_encoded_bytes()))
+                .map(|t| clap_complete::CompletionCandidate::new(t))
+                .collect()
         })))
         .subcommands(commands::builtin())
 }
 
+fn get_toolchains_from_rustup() -> Vec<String> {
+    let output = std::process::Command::new("rustup")
+        .arg("toolchain")
```
weihanglo, line 712: `stable-aarch64-apple-darwin (default)` is not a valid toolchain name. Need to handle this.

weihanglo: Also I'd like to check with rustup team whether the output format is stable to parse.

cc @rami3l 

Cargo parse the output from `rustup toolchain list` and `rustup target list` for the unstable native-completion features. Do we have any guarantee of the output format?

src/bin/cargo/cli.rs
```diff
@@ -690,10 +690,39 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
...
+                .arg( Arg::new("toolchain")
+            .help("Specifies the toolchain to use for this command")
+            .value_name("TOOLCHAIN")
+            .action(ArgAction::Set)
+        .add(clap_complete::engine::ArgValueCompleter::new(|partial: &OsStr| {
+            let toolchains = get_toolchains_from_rustup();
+
+            toolchains
+                .into_iter()
+                .filter(|t| OsStr::new(t).as_encoded_bytes().starts_with(partial.as_encoded_bytes()))
```
weihanglo, line 703: Is this OsStr conversion necessary? `CompletionCandidate::new` seems to accept any `Into<OsString>` type.

src/bin/cargo/cli.rs
```diff
@@ -690,10 +690,39 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
                     help.into()
                 }))
             }).collect()
+        })))
+                .arg( Arg::new("toolchain")
```
epage, line 694: Adding a new `Arg` for this completer changes the behavior of Cargo's CLI.

See https://github.com/rust-lang/cargo/issues/14520#issuecomment-2577972361

nooma-42: Is this also relevant with the `did_you_mean_file()` test in `script.rs` ([Tests Linux x86_64 nightly](https://github.com/rust-lang/cargo/actions/runs/13806634324/job/38618678473?pr=15301#logs)) ?  I failed that test but I don't know why

src/bin/cargo/cli.rs
nooma-42: Do I simply need to remove ` (default)`? 
> `stable-aarch64-apple-darwin (default)` is not a valid toolchain name. Need to handle this.

rami3l: > Also I'd like to check with rustup team whether the output format is stable to parse.
> 
> cc @rami3l
> 
> Cargo parse the output from `rustup toolchain list` and `rustup target list` for the unstable native-completion features. Do we have any guarantee of the output format?

Do you want to get a full list of toolchains? Rustup v1.28+ supports the following specifically for subprocess usages:

```console
> rustup toolchain list -q
stable-aarch64-apple-darwin
nightly-aarch64-apple-darwin
system
```

weihanglo: Thanks! Asked because I don't want to force rustup team to be stuck in a format just because Cargo accidentally relies on an output format that isn't supposed to be stable. I don't feel any frustration :)

So sounds like with `-q` the format is guarantee unchanged across versions?

src/bin/cargo/cli.rs
weihanglo: I guess so? This patch adds a new flag, whereas `cargo ./script` expects a file path.

epage:
Would you be able to clean up the commits for how this should be merged?

epage (approved):

PR #15303, opened 2025-03-12
docs: make it clearer that `rust_version` is enforced during compile

ericswpark:

The current wording makes it sound like Cargo will only refuse to compile the package if it encounters any missing functionality in stdlib while compiling the package on an unsupported toolchain. However, cargo will actually refuse to build the package unless overridden or the toolchain is updated.

No tests are required or applicable for this PR unless I am mistaken

### DIFF
--- a/src/doc/src/reference/rust-version.md
+++ b/src/doc/src/reference/rust-version.md
@@ -19,11 +19,10 @@ identifiers such as -nightly will be ignored while checking the Rust version.
 
 **Diagnostics:**
 
-When your package is compiled on an unsupported toolchain, Cargo will provide
-clearer diagnostics about the insufficient toolchain version by always erroring
-during compilation, rather than reporting invalid syntax or missing functionality
+When your package is compiled on an unsupported toolchain, Cargo will report that as an error to the user. This makes the support expectations clear and avoids reporting a less direct diagnostic like invalid syntax or missing functionality
 in the standard library. This affects all [Cargo targets](cargo-targets.md) in the
 package, including binaries, examples, test suites, benchmarks, etc.
+A user can opt-in to an unsupported build of a package with the `--ignore-rust-version` flag.
 
 
 **Development aid:**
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/rust-version.md
```diff
@@ -24,6 +24,9 @@ Cargo will provide clearer diagnostics about the insufficient toolchain version
 This affects all [Cargo targets](cargo-targets.md) in the package, including binaries, examples, test suites,
 benchmarks, etc.
 
+Cargo will not build your package on an unsupported toolchain, unless the `--ignore-rust-version` option
+below is specified.
```
epage, line 28: The previous paragraph is attempting to say the same thing.  By adding this, we are making it come across as these are saying two different things which can make this more confusing.

I can see being more upfront that we will always error with the intent of providing a clearer diagnostic than what you could get otherwise and having a comment at the end that in case the dependency happens to work, you can opt-in to an unsupported build of the package with `--ignore-rust-version`.

ericswpark: That's fair! I'll edit it to be part of the previous paragraph like you suggested.

ericswpark: Pushed new commit with suggestions applied. Thank you!

src/doc/src/reference/rust-version.md
```diff
@@ -19,10 +19,12 @@ identifiers such as -nightly will be ignored while checking the Rust version.
...
 
-When your package is compiled on an unsupported toolchain,
-Cargo will provide clearer diagnostics about the insufficient toolchain version rather than reporting invalid syntax or missing functionality in the standard library.
-This affects all [Cargo targets](cargo-targets.md) in the package, including binaries, examples, test suites,
-benchmarks, etc.
+When your package is compiled on an unsupported toolchain, Cargo will provide
+clearer diagnostics about the insufficient toolchain version by always erroring
+during compilation, rather than reporting invalid syntax or missing functionality
+in the standard library. This affects all [Cargo targets](cargo-targets.md) in the
+package, including binaries, examples, test suites, benchmarks, etc.
```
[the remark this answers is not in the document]
ericswpark: That looks better! Applied

epage (approved):

PR #15308, opened 2025-03-13
feat(complete): Added completion for `--profile`

Gmin2:

This attempts to complete the autocompleter for `cargo build --profile <TAB>`, it loads the built in profiles and the custom profile present in `Cargo.toml` and `.cargo/config.toml` if present when pressing TAB key, in the completion value it also adds an description inspired by [this](https://github.com/rust-lang/cargo/blob/master/src/cargo/core/compiler/job_queue/mod.rs#L805-L813)

Screenshot from 2025-03-14 02-17-40

Related to #14520

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -2,7 +2,7 @@ use crate::core::compiler::{
     BuildConfig, CompileKind, MessageFormat, RustcTargetData, TimingOutput,
 };
 use crate::core::resolver::{CliFeatures, ForceAllTargets, HasDevUnits};
-use crate::core::{shell, Edition, Package, Target, TargetKind, Workspace};
+use crate::core::{profiles::Profiles, shell, Edition, Package, Target, TargetKind, Workspace};
 use crate::ops::lockfile::LOCKFILE_NAME;
 use crate::ops::registry::RegistryOrIndex;
 use crate::ops::{self, CompileFilter, CompileOptions, NewOptions, Packages, VersionControl};
@@ -274,7 +274,11 @@ pub trait CommandExt: Sized {
         self._arg(
             opt("profile", profile)
                 .value_name("PROFILE-NAME")
-                .help_heading(heading::COMPILATION_OPTIONS),
+                .help_heading(heading::COMPILATION_OPTIONS)
+                .add(clap_complete::ArgValueCandidates::new(|| {
+                    let candidates = get_profile_candidates();
+                    candidates
+                })),
         )
     }
 
@@ -1104,6 +1108,53 @@ pub fn get_registry_candidates() -> CargoResult<Vec<clap_complete::CompletionCan
     } else {
         Ok(vec![])
     }
+}
+
+fn get_profile_candidates() -> Vec<clap_complete::CompletionCandidate> {
+    match get_workspace_profile_candidates() {
+        Ok(candidates) if !candidates.is_empty() => candidates,
+        // fallback to default profile candidates
+        _ => default_profile_candidates(),
+    }
+}
+
+fn get_workspace_profile_candidates() -> CargoResult<Vec<clap_complete::CompletionCandidate>> {
+    let gctx = new_gctx_for_completions()?;
+    let ws = Workspace::new(&find_root_manifest_for_wd(gctx.cwd())?, &gctx)?;
+    let profiles = Profiles::new(&ws, InternedString::new("dev"))?;
+
+    let mut candidates = Vec::new();
+    for name in profiles.profile_names() {
+        let Ok(profile_instance) = Profiles::new(&ws, name) else {
+            continue;
+        };
+        let base_profile = profile_instance.base_profile();
+
+        let mut description = String::from(if base_profile.opt_level.as_str() == "0" {
+            "unoptimized"
+        } else {
+            "optimized"
+        });
+
+        if base_profile.debuginfo.is_turned_on() {
+            description.push_str(" + debuginfo");
+        }
+
+        candidates
+            .push(clap_complete::CompletionCandidate::new(&name).help(Some(description.into())));
+    }
+
+    Ok(candidates)
+}
+
+fn default_profile_candidates() -> Vec<clap_complete::CompletionCandidate> {
+    vec![
+        clap_complete::CompletionCandidate::new("dev").help(Some("unoptimized + debuginfo".into())),
+        clap_complete::CompletionCandidate::new("release").help(Some("optimized".into())),
+        clap_complete::CompletionCandidate::new("test")
+            .help(Some("unoptimized + debuginfo".into())),
+        clap_complete::CompletionCandidate::new("bench").help(Some("optimized".into())),
+    ]
 }
 
 fn get_example_candidates() -> Vec<clap_complete::CompletionCandidate> {
--- a/src/cargo/core/profiles.rs
+++ b/src/cargo/core/profiles.rs
@@ -391,6 +391,11 @@ impl Profiles {
             .get(name)
             .ok_or_else(|| anyhow::format_err!("profile `{}` is not defined", name))
     }
+
+    /// Returns an iterator over all profile names known to Cargo.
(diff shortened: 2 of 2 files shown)


src/cargo/util/command_prelude.rs
```diff
@@ -1106,6 +1110,53 @@ pub fn get_registry_candidates() -> CargoResult<Vec<clap_complete::CompletionCan
...
+    Ok(vec![
+        clap_complete::CompletionCandidate::new("dev".to_string())
+            .help(Some("unoptimized + debuginfo".into())),
+        clap_complete::CompletionCandidate::new("release".to_string())
+            .help(Some("optimized".into())),
+        clap_complete::CompletionCandidate::new("test".to_string())
+            .help(Some("unoptimized + debuginfo with tests enabled".into())),
+        clap_complete::CompletionCandidate::new("bench".to_string())
+            .help(Some("optimized with tests enabled".into())),
+    ])
```
epage, line 1157: My general recommendation is
- Keep code blocks organized around business logic
- The one exception is guard statements

To fit within that, I would recommend either
- Using an `else` and not doing an early return
- Use `let Ok() = foo else { hardcoded };`

epage, line 1157: Should we do this as an array of string tuples and make the candidates from that?  I wonder if that would be better

epage: Granted, you arenested two levels deep and this is a catch-all for those.  We don't have `&&` for `if let` yet.  Another option is `.and_then`.  Or not doing this.  Or we do like `get_target_triples` and we add these if `candidates.is_empty()`

epage: But then its more annoying to add anything else

Gmin2: dont yu think following the `get_target_triplets` we will get too many nested if satement(ig 4)

Gmin2: ig `vec` impl is fine only as doing the tuples impl we would need to change both the tuple def and conversion as well

src/cargo/util/command_prelude.rs
```diff
@@ -1106,6 +1110,53 @@ pub fn get_registry_candidates() -> CargoResult<Vec<clap_complete::CompletionCan
...
+                    } else {
+                        "optimized"
+                    });
+
+                    if base_profile.debuginfo.is_turned_on() {
+                        description.push_str(" + debuginfo");
+                    }
+
+                    if matches!(name.as_str(), "test" | "bench") {
+                        description.push_str(" with tests enabled");
```
weihanglo, line 1134: What is the rationale behind this? There is no difference between these profiles and the one they inherit from.

src/cargo/util/command_prelude.rs
```diff
@@ -1106,6 +1110,53 @@ pub fn get_registry_candidates() -> CargoResult<Vec<clap_complete::CompletionCan
...
+                            .help(Some(description.into())),
+                    );
+                }
+            }
+
+            return Ok(candidates);
+        }
+    }
+
+    Ok(vec![
```
weihanglo, line 1148: This is also `profile` in configuration file, though we might not want to go that far in completion code I guess?

https://github.com/rust-lang/cargo/blob/611b7c4ee40b63d0818040f7290a8e14424a9519/src/cargo/core/profiles.rs?plain=1#L1299-L1302

epage: In the main flow, I believe we do load `[profile]` from both config and workspace because we do `Profile::new().get_profile_names` which looked like the main code path for doing this.

This code path is for when that fails.  We could try to make it smarter and give people as much information as can successfully run  but that ends up duplicating a lot more of `Profiles`. My assumption is that it will be fine with the hard coded list.

src/cargo/util/command_prelude.rs
epage: This has two recovery baths
- this
- unwrap_or_default

Use the two method approach can flatten this and drop us to one

src/cargo/util/command_prelude.rs
epage: This still needs resolving

Gmin2: > There is no difference between these profiles and the one they inherit from.
got it, removed in https://github.com/rust-lang/cargo/pull/15308/commits/ed48f5f08917559d9b80079118f8ed0696ac9920

epage:
Could you clean up the commits for how they should be merged?

Gmin2:
> Could you clean up the commits for how they should be merged?

Done

src/cargo/util/command_prelude.rs
```diff
@@ -1106,6 +1110,56 @@ pub fn get_registry_candidates() -> CargoResult<Vec<clap_complete::CompletionCan
...
+}
+
+fn get_workspace_profile_candidates() -> CargoResult<Vec<clap_complete::CompletionCandidate>> {
+    let gctx = new_gctx_for_completions()?;
+    let ws = Workspace::new(&find_root_manifest_for_wd(gctx.cwd())?, &gctx)?;
+    let profiles = Profiles::new(&ws, InternedString::new("dev"))?;
+
+    let mut candidates = Vec::new();
+    for name in profiles.profile_names() {
+        if let Ok(profile_instance) = Profiles::new(&ws, name) {
```
weihanglo, line 1128: We are recreating `Profiles` struct here, which doesn't really make sense, as `Profiles` already has every profile. But for completion, this seems acceptable?

BTW we could use let-else so one less identation level.

```suggestion
        let Ok(profile_instance) = Profiles::new(&ws, name) else {
            continue;
        };
```

src/cargo/util/command_prelude.rs
```diff
@@ -1106,6 +1110,56 @@ pub fn get_registry_candidates() -> CargoResult<Vec<clap_complete::CompletionCan
...
+            } else {
+                "optimized"
+            });
+
+            if base_profile.debuginfo.is_turned_on() {
+                description.push_str(" + debuginfo");
+            }
+
+            candidates.push(
+                clap_complete::CompletionCandidate::new(name.to_string())
```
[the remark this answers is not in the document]
Gmin2: thanks, for the catch

src/cargo/util/command_prelude.rs
```diff
@@ -1106,6 +1110,56 @@ pub fn get_registry_candidates() -> CargoResult<Vec<clap_complete::CompletionCan
     }
 }
 
+fn get_profile_candidates() -> CargoResult<Vec<clap_complete::CompletionCandidate>> {
+    match get_workspace_profile_candidates() {
+        Ok(candidates) if !candidates.is_empty() => Ok(candidates),
+        // fallback to default profile candidates
+        _ => Ok(default_profile_candidates()),
+    }
+}
```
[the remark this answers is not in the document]
Gmin2: thanks

src/cargo/util/command_prelude.rs
Gmin2: > BTW we could use let-else so one less identation level.

done 

> We are recreating Profiles struct here, which doesn't really make sense, as Profiles already has every profile. But for completion, this seems acceptable?

you are right that recreating the `Profiles` struct for each profile name is inefficient, the original profiles object should already contain all the information, I kept this approach for consistency with the existing implementation, but if you think it would be better to refactor this to avoid the redundant Profiles creation i would be happy to refactor it

src/cargo/util/command_prelude.rs
```diff
@@ -1106,6 +1110,56 @@ pub fn get_registry_candidates() -> CargoResult<Vec<clap_complete::CompletionCan
...
+            );
+        }
+    }
+
+    Ok(candidates)
+}
+
+fn default_profile_candidates() -> Vec<clap_complete::CompletionCandidate> {
+    vec![
+        clap_complete::CompletionCandidate::new("dev".to_string())
```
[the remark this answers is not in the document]
epage: FYI it will allocate either way; its just not needed to be explicit (and I prefer limiting `to_string()` to only when doing conversions)

epage (approved):
Thanks for your work on this!

PR #15310, opened 2025-03-14
docs(ref): Mention `x.y.*` as a kind of version requirement to avoid.

kpreid:
This expands the text introduced by #12323.

`x.y.*` is equivalent to `>=x.y.0, <x.(y+1).0`, so it can create an unresolvable conflict in the same way as already discussed.

I also added a little punctuation to improve the structure of some nearby sentences.

@rustbot label A-documenting-cargo-itself

### DIFF
--- a/src/doc/src/reference/specifying-dependencies.md
+++ b/src/doc/src/reference/specifying-dependencies.md
@@ -162,19 +162,19 @@ is ignored and should not be used in version requirements.
 > [#10599]).
 >
 > Avoid constraining the upper bound of a version to be anything less than the
-> next semver incompatible version
-> (e.g. avoid `">=2.0, <2.4"`) as other packages in the dependency tree may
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/specifying-dependencies.md
```diff
@@ -162,19 +162,19 @@ is ignored and should not be used in version requirements.
 > [#10599]).
 >
 > Avoid constraining the upper bound of a version to be anything less than the
-> next semver incompatible version
-> (e.g. avoid `">=2.0, <2.4"`) as other packages in the dependency tree may
+> next semver incompatible version (e.g. avoid `">=2.0, <2.4"` or `"2.0.*"`),
```
weihanglo, line 165: What about `~2.1.0`, isn't it effectively the same?

kpreid: Added that too. Thanks!

weihanglo (approved):
Thanks!

PR #15311, opened 2025-03-14
feat(package): add `--list json` option for JSON output format

weihanglo:

#11666

This adds an option `json` to `--list` for `cargo package`.

* The `--list` option without any argumenet value remains the old path-per-line behavior.
* The `--list json` emits the unstable machine-readable JSON information, which include the absolute paths to the resolved/normalized file paths.

* This adds `dunce` as a direct dependency, which is already a transitive via gix.
  Because I don't want to deal with std canonicalization mess.
  If we don't want to show absolute paths in the JSON output,
  we should also switch to relative path to either package root or cwd
  (I prefer the latter though)
  If we choose relative paths, then `dunce` can be removed.
* The actual schema and field names is open to discuss and change. See also <https://github.com/rust-lang/cargo/issues/12377> for reference.
* Do we want to provide an option for the old broken-with-workspace format?

### Additional information


src/cargo/ops/cargo_package/mod.rs
```diff
@@ -69,16 +88,55 @@ struct ArchiveFile {
...
+    /// A file copied from another location,
+    ///
+    /// The associated file path points to a resolved symlink target
+    /// or a normalized path outside the package root.
+    Copied(PathBuf),
     /// Generates a file.
     Generated(GeneratedFile),
 }
 
+impl serde::ser::Serialize for FileContents {
```
epage, line 102: Can we have the schema defined in `cargo-util-schemas` with a schema.json file?

src/bin/cargo/commands/package.rs
```diff
@@ -1,18 +1,21 @@
...
         .arg(
-            flag(
+            optional_opt(
                 "list",
-                "Print files included in a package without making one",
+                "Print files included in a package without making one: (unstable) json",
             )
-            .short('l'),
+            .short('l')
+            .value_name("FMT"),
```
epage, line 18: Should we instead use `--message-format`?

One reason is I wonder if we should focus on ndjson messages rather than a single json blob.  This makes it easier to extend with more feature.  For example, we can't extend `cargo metadata` with warnings or errors without adding them to that single schema which feels weird.

src/doc/src/reference/unstable.md
```diff
@@ -1882,6 +1883,16 @@ Specify which packages participate in [feature unification](../reference/feature
...
 
+## package list JSON format
+
+* Tracking Issue: [#11666](https://github.com/rust-lang/cargo/issues/11666)
+
+The `--list json` flag value in `cargo package` emits machine-readable JSON 
+information about each file included in a package.
+Requires `-Zunstable-options`.
+See [`cargo package --list`](../commands/cargo-package.md#option-cargo-package---list[fmt])
+for more information.
```
epage, line 1894: Can we document the schema?

src/doc/man/cargo-package.md
```diff
@@ -81,8 +81,45 @@ There is no guarantee that the source code in the tarball matches the VCS inform
...
+
+Specifying an output format (rather than the default) is unstable and requires
+`-Zunstable-options`. Valid output formats:
+
+- `json` (unstable, requires `-Zunstable-options`): Emit machine-readable JSON
+  information about each file.
+  ```javascript
+  {
+    /* The Package ID Spec of the package. */
+    "path+file:///home/foo#0.0.0": {
```
epage, line 95: Like I said at https://github.com/rust-lang/cargo/pull/15311/files#r1995637878, should we do a message per package or a message per file?

src/doc/man/cargo-package.md
```diff
@@ -81,8 +81,45 @@ There is no guarantee that the source code in the tarball matches the VCS inform
...
+      /* Relative path in the archive file. */
+      "Cargo.toml.orig": {
+        /* An absolute path to the actual file content. */
+        "original_path": "/home/foo/Cargo.toml",
```
epage, line 99: snake case or kebab case?

src/doc/man/cargo-package.md
```diff
@@ -81,8 +81,45 @@ There is no guarantee that the source code in the tarball matches the VCS inform
...
+           - "generate" for file being generated during packaging
+           - "copied" for file copied from another location.
+             The associated path points to a resolved symlink target
+             or a normalized path outside the package root.
+        */
+        "source": "copied"
+      },
+      "Cargo.toml": {
+        "source": "generated"
+      },
```
epage, line 111: This isn't generated from nothing, should we say what its being generated from?

src/doc/src/reference/unstable.md
weihanglo: That is in `cargo package --list`. Do we want to duplicate it?

src/doc/man/cargo-package.md
```diff
@@ -81,8 +81,45 @@ There is no guarantee that the source code in the tarball matches the VCS inform
...
+        "original_path": "/home/foo/Cargo.toml",
+        /* The source of the file.
+           - "existing" for file that exists on disk as-is.
+           - "generate" for file being generated during packaging
+           - "copied" for file copied from another location.
+             The associated path points to a resolved symlink target
+             or a normalized path outside the package root.
+        */
+        "source": "copied"
+      },
```
epage, line 108: I keep thinking "source path" when I see "source".   Wonder if there is a better name.  kind?  operation?  Not found one that I'm happy with.

src/doc/src/reference/unstable.md
epage: Oh, forgot that we only document the schemas in the relevant commands, nm

src/doc/man/cargo-package.md
weihanglo: We do that i `-Zsbom`, `cargo metadata`, and [`--message-format=json`](https://doc.rust-lang.org/nightly/cargo/reference/external-tools.html#compiler-messages). This follows.

epage: sbom is just because of `cargo metadata`.  We deferred what it should actually do to the RFC. (I brought this up there).

Uck, didn't know `--message-format=json` does that.

 My hope for the new plumbing commands is that we reuse public schemas as much as possible, like `TomlManifest`.  Unless we copy/paste it, that will be kebab case.

So for right now, our [schema document](https://doc.crates.io/contrib/implementation/schemas.html#schema-design) gives an unqualified statement to use kebab case.  We should probably discuss this and figure out if there needs to be more qualifiers to it.

src/doc/man/cargo-package.md
epage: `cargo-release` runs `cargo package --list` to know what files will affect the final `.crate` so we can check their git history..  We have heuristics on how to use this information.  It would be good to not require people to have to maintain those heuristics.

src/bin/cargo/commands/package.rs
epage: I've taken `message` to mean "output message" which is generic, not specific to the type of message.  The shape would probably need to be consistent across the different messages though.

src/doc/man/cargo-package.md
weihanglo: I am good with per package.

How would message per file look like? Tools may need more work to associate all of them, but tools usually want information for the entire package I assume?
Also, with message-per-file, Cargo then needs to guarantee that there is no interleaving in between. Otherwise tools either need to consume all messages to determine if everything in a single package is collected.

src/bin/cargo/commands/package.rs
weihanglo: In the future `--message-format=json` would also affect status message Cargo emits, right? So here comes another counterpoint: This adds some more work on users side, if they want to keep the human-readable status messages.

src/doc/man/cargo-package.md
weihanglo: What about this?

```javascript
{
  "Cargo.toml": {
    "path": "Cargo.toml",
    "kind": "generated", 
  },
  "Cargo.toml.orig": {
    "path": "Cargo.toml",
    "kind": "on_disk",
  },
  "README.md": {
    "path": "../README.md",
    "kind": "on_disk",
  }
}
```

* Only "generated" and "on_disk".
  Assume that people don't need to know whether a file was copied from the other location.
  If they need,  the `path` field shows the it is different from the object key.
* Use relative path to the current working directory

src/bin/cargo/commands/package.rs
epage: > In the future `--message-format=json` would also affect status message Cargo emits, right? So here comes another counterpoint: This adds some more work on users side, if they want to keep the human-readable status messages.

Thats true when interacting with most of our json APIs and why some have asked for a way to tee.

src/doc/man/cargo-package.md
epage: I think I'd just go with `coped` for all (they are all being copied from somewhere).  This also makes the verbage more consistent.

Speaking of verbage, should these be present tense (`generate`, `copy`)?

src/cargo/ops/cargo_package/mod.rs
weihanglo: Doesn't seem like a blocker. I would like to defer until stabilization, or any request from users.

src/cargo/ops/cargo_package/mod.rs
```diff
@@ -40,10 +41,38 @@ use unicase::Ascii as UncasedAscii;
...
+#[derive(Debug, Clone)]
+pub enum PackageMessageFormat {
+    Plain,
+    Json,
+}
+
+impl PackageMessageFormat {
+    pub const POSSIBLE_VALUES: [&str; 2] = ["plain", "json"];
+
+    pub const DEFAULT: &str = "plain";
```
epage, line 57: Should `plain` be `human`?  Thats what we use for `cargo check`

src/cargo/ops/cargo_package/mod.rs
```diff
@@ -107,6 +108,28 @@ enum FileContents {
...
+                map.end()
+            }
+            FileContents::Generate(_) => {
+                let mut map = serializer.serialize_map(Some(1))?;
+                map.serialize_entry("kind", "generate")?;
+                map.end()
+            }
+        }
+    }
+}
```
epage: Why is this hand written?  If its for `kind`, isn't that just `#[serde(tag = "kind")]`?

src/cargo/ops/cargo_package/mod.rs
epage: This is a schema we are expecting people to use programmatically but
- We are keeping internal the helpers for doing so, leaving them to the documentation.
- The serialization schema is tied directly to the implementation in a location that doesn't call out this is for serialization.  This makes it harder for us to track schema changes (e.g. a refactor to rename a field).  At least the `impl Serialize` is hand written which helps call it out but it might not always be that way.  We do have tests but those can't exhaustively cover every case.

I also would like to see every existing serialization schema moved there.

I'd like to flip the script and instead of waiting on a need, for us to proactively do it.  It will reduce the work we need to do later to take care of this if we do it in the first place.  I held off on this for sboms mostly because I brought it up late in the process.

Yes, this could be done at stabilization.  We'd need to track that as work to do for that.  I also like to see stabilization make as small of a change as possible so its easier to review and has less risk.

Maybe waiting until later in the pre-stabilization process can help if there is expected churn in the overall design that it would require major rework (e.g. `--list json` vs `--message-format json` can end up with very different needs)

src/cargo/ops/cargo_package/mod.rs
weihanglo: Fixed, along with the schema extraction!

src/cargo/ops/cargo_package/mod.rs
weihanglo: Changed it to `human`. If we're going to `--message-format` for every message, the default will be `human`. The current file-per-line format should be "human" since it is the default.

src/cargo/ops/cargo_package/mod.rs
weihanglo: Extracted. Separating internal implementation makes a lot of sense!

According to our JSON compatibility policy, the schema should have room to add more fields, the current `path` and `kind` won't move away I feel like, so it should be fine doing it now than later.

epage: There isn't a `schema.json` file.

However,
- Its unclear what that file should look like for messages (should we put all messages in it?)
- Those are most for people code-genning off of it (less needed with `cargo-util-schemas`) or editors (less needed for programmatic files)

So not going to block on that.

epage (approved):
Thanks!  Looking forward to using this in `cargo-release`!

epage:
btw I added some of the decisions made during the review discussion to the PR description so we can better track them as we move towards stabilization (so the rest of the team can re-affirm or disagree)

Issue #15312, opened 2025-03-14
Multiple --crate-type=X (--crate-type=staticlib --crate-type=staticlib) leads to a warning and then to assertion


azat:
Repro

```
$ cargo --version
cargo 1.87.0-nightly (ab1463d63 2025-03-08)
$ rustc --version
rustc 1.87.0-nightly (cbfdf0b01 2025-03-13)

$ cargo new my_staticlib --lib

$ cargo rustc --lib --crate-type=staticlib --crate-type=staticlib
warning: output filename collision.
The lib target `my_staticlib` in package `my_staticlib v0.1.0 (/tmp/rust-test/my_staticlib)` has the same output filename as the lib target `my_staticlib` in package `my_staticlib v0.1.0 (/tmp/rust-test/my_staticlib)`.
Colliding filename is: /tmp/rust-test/my_staticlib/target/debug/deps/libmy_staticlib-6e3dfca3ee62c788.a
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
warning: output filename collision.
The lib target `my_staticlib` in package `my_staticlib v0.1.0 (/tmp/rust-test/my_staticlib)` has the same output filename as the lib target `my_staticlib` in package `my_staticlib v0.1.0 (/tmp/rust-test/my_staticlib)`.
Colliding filename is: /tmp/rust-test/my_staticlib/target/debug/libmy_staticlib.a
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
   Compiling my_staticlib v0.1.0 (/tmp/rust-test/my_staticlib)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s

$ cargo rustc --lib --crate-type=staticlib --crate-type=staticlib
warning: output filename collision.
The lib target `my_staticlib` in package `my_staticlib v0.1.0 (/tmp/rust-test/my_staticlib)` has the same output filename as the lib target `my_staticlib` in package `my_staticlib v0.1.0 (/tmp/rust-test/my_staticlib)`.
Colliding filename is: /tmp/rust-test/my_staticlib/target/debug/deps/libmy_staticlib-6e3dfca3ee62c788.a
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
[...]

PR #15314, opened 2025-03-15
fix: deduplicate crate types in cargo rustc command

Frank-III:

close: #15312

deduplicate `crate-type` in cargo `rustc` command

### DIFF
--- a/tests/testsuite/rustc.rs
+++ b/tests/testsuite/rustc.rs
@@ -855,3 +855,53 @@ fn precedence() {
 "#]])
         .run();
 }
+
+#[cargo_test]
+fn build_with_duplicate_crate_types() {
+    let p = project().file("src/lib.rs", "").build();
+
+    p
+        .cargo("rustc -v --crate-type staticlib --crate-type staticlib")
+        .with_stderr_data(str![[r#"
+[WARNING] output filename collision.
+The lib target `foo` in package `foo v0.0.1 ([ROOT]/foo)` has the same output filename as the lib target `foo` in package `foo v0.0.1 ([ROOT]/foo)`.
+Colliding filename is: [ROOT]/foo/target/debug/deps/libfoo-[HASH].a
+The targets should have unique names.
+Consider changing their names to be unique or compiling them separately.
+This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
+[WARNING] output filename collision.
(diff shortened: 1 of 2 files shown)


tests/testsuite/rustc.rs
```diff
@@ -855,3 +855,18 @@ fn precedence() {
...
+
+#[cargo_test]
+fn build_with_duplicate_crate_types() {
+    let p = project().file("src/lib.rs", "").build();
+
+    // Test with duplicate crate types specified directly
+    p.cargo("rustc -v --crate-type staticlib --crate-type staticlib")
+        .with_stderr_data(str![[r#"
+[COMPILING] foo v0.0.1 ([ROOT]/foo)
+[RUNNING] `rustc --crate-name foo --edition=2015 src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type staticlib --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 -C split-debuginfo=unpacked --check-cfg 'cfg(docsrs,test)' --check-cfg 'cfg(feature, values())' -C metadata=991ef8c99cd510e0 -C extra-filename=-f0d97c77b50b1fb3 --out-dir [ROOT]/foo/target/debug/deps -L dependency=[ROOT]/foo/target/debug/deps`
```
weihanglo, line 867: Assering the entire rustc argument list is a big fragile. Maybe this?
(`[..]` for wildcard matching.)

```suggestion
[RUNNING] `rustc [..]future-incompat --crate-type staticlib --emit[..]
```

tests/testsuite/rustc.rs
weihanglo, line 1: Thanks for the patch! This looks good.

Would you mind splitting this into two commits?

1. The first commit adds a test with the snapshot of the current problematic behavior (and the test passes). This can be seen as a reproduction verifying the bug does exist.
2. The second commit contains both the actual fix, and the test change, so the test snapshot change could beautifully reflect the behavior change.

weihanglo: It is not a hard blocker btw!

weihanglo (approved):
Thanks! Any reason this is still a draft?

Frank-III:
> Thanks! Any reason this is still a draft?

sorry, the test I added is not correctly capture the behavior before!

weihanglo (approved):
Good catch! Thank you for the contribution @Frank-III :)

PR #15317, opened 2025-03-16
do not pass cdylib link args to test

### LINKED ISSUE #12663: cdylib extra link args apply to unit tests
### Problem

 According to [The Cargo Book](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script), the `build.rs` file can emit the `cargo:rustc-link-arg-cdylib=FLAG` setting to have it passed to rustc as a `-C link-arg=FLAG` but only when building a "cdylib library target". From the documentation, its not clear that building and running the test binaries for unittests in a cdylib crate will also apply `-C link-arg=FLAG`. 
 
 In my specific case, I am passing (among other linker arguments) `rustc-link-arg-cdylib=/INTEGRITYCHECK` to a cdylib crate via `build.rs`.  Doing so also passes it to the unit tests, so it makes it impossible to execute my unit tests.
 
This is sort of the opposite problem of https://github.com/rust-lang/cargo/issues/10937. In that issue, they want `cargo:rustc-link-arg-tests=FLAG` to apply to unittests. In this issue, I want all the other `rustc-link-arg` variants to not apply to unit tests (or at least have some option that lets me do this). 

### Steps

1. Create project with files:
`Cargo.toml`:
```rust
[package]
name = "rustc-link-arg-bug"
version = "0.1.0"
edition = "2021"

[lib]
crate-type = ["cdylib"]
```

`build.rs`
```rust
fn main() {
    println!("cargo:rustc-link-arg-cdylib=/INTEGRITYCHECK");
}
```

`src/lib.rs`:
empty file

2. Run Cargo Test:
```
cargo test
    Finished test [unoptimized + debuginfo] target(s) in 0.02s
     Running unittests src\lib.rs (target\debug\deps\rustc_link_arg_bug-d9c1a47e9d03aa01.exe)
error: test failed, to rerun pass `--lib`

Caused by:
  could not execute process `D:\rustc-link-arg-bug\target\debug\deps\rustc_link_arg_bug-d9c1a47e9d03aa01.exe` (never executed)

Caused by:
  Windows cannot verify the digital signature for this file. A recent hardware or software change might have installed a file that is signed incorrectly or damaged, or that might be malicious software from an unknown source. (os error 577)
```
`cargo test` fails to run because `/INTEGRITYCHECK` was passed to the unittest "targets" (Aside: are these unit test binaries considered a separate target in the world of cargo?)

__note: `/INTEGRITYCHECK` is an msvc-specific linker argument__

### Possible Solution(s)

I think a way to control whether extra link args are passed or not passed to unit tests would be a good thing. The [cargo documentation](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script) could also be improved as it pretty loose with the word "tests", and the way it is worded could confuse people (like me) as to what directives apply to tests(including unittests) and what applies only to integration tests(ie. test targets?).

Note: I would be very happy if anyone knows any workaround to this issue that allows me to pass linker arguments only to the cdylib and not to the cdylib's tests.

### Notes

_No response_

### Version

```text
cargo 1.74.0-nightly (2fc85d15a 2023-09-09)
release: 1.74.0-nightly
commit-hash: 2fc85d15a542bfb610aff7682073412cf635352f
commit-date: 2023-09-09
host: x86_64-pc-windows-msvc
libgit2: 1.7.1 (sys:0.18.0 vendored)
libcurl: 8.2.1-DEV (sys:0.4.65+curl-8.2.1 vendored ssl:Schannel)
os: Windows 10.0.23541 (Windows 11 Enterprise) [64-bit]
```

usamoi:
fixes #12663

### DIFF
--- a/tests/testsuite/build_script_extra_link_arg.rs
+++ b/tests/testsuite/build_script_extra_link_arg.rs
@@ -443,3 +443,40 @@ fn cdylib_both_forms() {
 "#]])
         .run();
 }
+
+// https://github.com/rust-lang/cargo/issues/12663
+#[cargo_test]
+fn cdylib_extra_link_args_should_not_apply_to_unit_tests() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.5.0"
+                edition = "2015"
+
+                [lib]
+                crate-type = ["lib", "cdylib"]
+            "#,
+        )
+        .file(
+            "src/lib.rs",
+            r#"
+                #[test]
+                fn noop() {}
+            "#,
+        )
+        .file(
+            "build.rs",
+            r#"
+                fn main() {
+                    // This would fail if cargo passed `-lhack` to building the test because `hack` doesn't exist.
+                    println!("cargo::rustc-link-arg-cdylib=-lhack");
+                }
+            "#,
+        )
+        .build();
+
+    p.cargo("test --lib").run();
+}
--- a/src/cargo/core/compiler/build_runner/mod.rs
+++ b/src/cargo/core/compiler/build_runner/mod.rs
@@ -262,7 +262,7 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
                         }
 
                         for (lt, arg) in &output.linker_args {
-                            if lt.applies_to(&unit.target) {
+                            if lt.applies_to(&unit.target, unit.mode) {
                                 args.push("-C".into());
                                 args.push(format!("link-arg={}", arg).into());
                             }
--- a/src/cargo/core/compiler/custom_build.rs
+++ b/src/cargo/core/compiler/custom_build.rs
@@ -37,6 +37,7 @@ use crate::core::compiler::build_runner::UnitHash;
 use crate::core::compiler::fingerprint::DirtyReason;
 use crate::core::compiler::job_queue::JobState;
 use crate::core::{profiles::ProfileRoot, PackageId, Target};
+use crate::util::command_prelude::CompileMode;
 use crate::util::errors::CargoResult;
 use crate::util::internal;
 use crate::util::machine_message::{self, Message};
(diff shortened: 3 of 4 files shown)


epage:
r? ehuss

Since you have more of the background on this

ehuss:
Thanks!

I'm a little uncertain about changing `Bin`, `SingleBin`, and `Example`. My thinking is that if you have a flag that is required to link a `bin`, then the flag would also likely be required to link a unittest of the bin (since they are mostly the same thing). Can you say more about why those are being changed?

tests/testsuite/build_script_extra_link_arg.rs
```diff
@@ -403,3 +403,39 @@ fn build_script_extra_link_arg_examples() {
         )
         .run();
 }
+
+#[cargo_test]
+fn issue_12663() {
```
ehuss, line 408: Generally it is best to give the test a descriptive name instead of issue numbers, and a comment explaining what the test does (the comment can include a full URL to the issue).

tests/testsuite/build_script_extra_link_arg.rs
```diff
@@ -403,3 +403,39 @@ fn build_script_extra_link_arg_examples() {
...
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.5.0"
+                authors = ["wycats@example.com"]
```
ehuss, line 416: This isn't needed anymore.

```suggestion
```

tests/testsuite/build_script_extra_link_arg.rs
```diff
@@ -403,3 +403,39 @@ fn build_script_extra_link_arg_examples() {
...
+            "build.rs",
+            r#"
+                fn main() {
+                    println!("cargo::rustc-cdylib-link-arg=-lhack");
+                }
+            "#,
+        )
+        .build();
+
+    p.cargo("test --lib").run();
```
ehuss, line 440: A comment here could be useful to help quickly understand the test.

```suggestion
    // This would fail if cargo passed `-lhack` to building the test because `hack` doesn't exist.
    p.cargo("test --lib").run();
```

tests/testsuite/build_script_extra_link_arg.rs
```diff
@@ -403,3 +403,39 @@ fn build_script_extra_link_arg_examples() {
...
+            r#"
+                #[test]
+                fn noop() {}
+            "#,
+        )
+        .file(
+            "build.rs",
+            r#"
+                fn main() {
+                    println!("cargo::rustc-cdylib-link-arg=-lhack");
```
weihanglo, line 434: <https://github.com/rust-lang/cargo/pull/15326> favors `rustc-link-arg-cdylib` over this one.

usamoi:
> I'm a little uncertain about changing `Bin`, `SingleBin`, and `Example`. My thinking is that if you have a flag that is required to link a `bin`, then the flag would also likely be required to link a unittest of the bin (since they are mostly the same thing). Can you say more about why those are being changed?

I initially assumed they were different based on intuition.

I recalled a case where a binary was compiled for a bare-metal target, and I used custom test frameworks to run tests in the same environment. If the current behavior changes, similar use cases could break.

So, I reverted to the previous behavior for `Bin`, `SingleBin` and `Example`.

ehuss (approved):
Thanks!

PR #15319, opened 2025-03-17
feat: Add custom completer for cargo <TAB> to complete aliases defined in config.toml

nooma-42:

This PR introduces auto-completion for the `cargo <TAB>` option. When a user types `cargo <TAB>` and presses the TAB key, the system will automatically suggest aliases defined in `config.toml`

1. In the terminal, type `cargo <TAB>`
2. Press the TAB key.
3. You should see aliases suggestions

https://github.com/user-attachments/assets/d8a265e8-bbd9-4f58-8121-80caf142d8a6

### DIFF
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -11,6 +11,7 @@ use std::fmt::Write;
 
 use super::commands;
 use super::list_commands;
+use super::user_defined_aliases;
 use crate::command_prelude::*;
 use crate::util::is_rustup;
 use cargo::core::shell::ColorChoice;
@@ -691,11 +692,13 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
                 }))
             }).collect()
         })))
-        .add(clap_complete::engine::SubcommandCandidates::new(|| {
-            get_toolchains_from_rustup()
+        .add(clap_complete::engine::SubcommandCandidates::new(move || {
+            let mut candidates = get_toolchains_from_rustup()
                 .into_iter()
                 .map(|t| clap_complete::CompletionCandidate::new(t))
-                .collect()
+                .collect::<Vec<_>>();
+            candidates.extend(get_alias_candidates());
+            candidates
         }))
         .subcommands(commands::builtin())
 }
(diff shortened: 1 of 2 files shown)


src/bin/cargo/cli.rs
```diff
@@ -691,6 +693,20 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
                 }))
             }).collect()
         })))
+        .add(clap_complete::engine::SubcommandCandidates::new(move || {
```
epage, line 696: As a warning, this will have conflicts with #15301

src/bin/cargo/cli.rs
```diff
@@ -571,6 +571,8 @@ pub fn cli(gctx: &GlobalContext) -> Command {
             .invalid(style::INVALID)
     };
 
+    let alias_map = super::user_defined_aliases(gctx);
```
epage, line 574: This is pessimizing performance.  We'll load user defined aliases an extra time for no other reason than for when completions are performed.  This should be called inside of the closure.  If you look at the ArgValueCompleters in `command_prelude.rs`, you can see how we deal with getting a global context inside of a completer

src/bin/cargo/cli.rs
```diff
@@ -691,6 +693,20 @@ See '<cyan,bold>cargo help</> <cyan><<command>></>' for more information on a sp
...
         })))
+        .add(clap_complete::engine::SubcommandCandidates::new(move || {
+            alias_map.iter()
+                .map(|(alias, cmd_info)| {
+                    let help_text = if let super::CommandInfo::Alias { target } = cmd_info {
+                        let cmd_str = target.iter().map(String::as_str).collect::<Vec<_>>().join(" ");
+                        format!("alias for {}", cmd_str)
+                    } else {
+                        "alias (from config)".to_string()
+                    };
```
epage, line 704: I take it this branch is a "just in case" a new variant is added relevant for aliases and we don't want to break the user?  Otherwise, it looks like this could be an `unreachable!`.

One option is to use a `match` and make each of the existing variants `unreachable`.  If a new variant is added, it will then be a compiler error to let us know to update this by either reporting a candidate or by adding `unreachable!`

src/cargo/util/command_prelude.rs
```diff
@@ -1370,6 +1370,50 @@ fn new_gctx_for_completions() -> CargoResult<GlobalContext> {
     Ok(gctx)
 }
 
+pub fn get_alias_candidates() -> Vec<clap_complete::CompletionCandidate> {
```
epage, line 1373: Why was this moved?

weihanglo:
@nooma-42 I'll look into build-std failure today.

src/cargo/util/command_prelude.rs
nooma-42: Are you asking about `user_defined_aliases`? I move `user_defined_aliases` from `src/bin/cargo/main.rs` to `src/cargo/util/command_prelude.rs` because I cannot access `user_defined_aliases` in `command_prelude.rs` or `cli.rs`

weihanglo:
A rebase onto master should fix the CI failure.

src/cargo/util/command_prelude.rs
epage: `cli.rs` has access to content in `main.rs`

Why does `command_prelude.rs` need access to it?

epage (approved):
Thanks!

Issue #15283, opened 2025-03-08
Tracking Issue for deprecating per-build-target `edition` field


ehuss:
### Summary

RFC: [#3772](https://github.com/rust-lang/rfcs/pull/3772)
Implementation: TODO
Documentation: TODO

This feature is to deprecate the `edition` field in the [cargo target configuration](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-edition-field).

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

PR #15321, opened 2025-03-17
fix(toml): Report '<target>.edition' deprecation to users

epage:

This is a part of rust-lang/cargo#15283

In the RFC, I mentioned this might be blocked on #12235.  
In hindsight, the use of this is [rare enough](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Deprecate.20build-target.20.60edition.60.20field.3F/near/499047806) that I suspect we can go ahead and warn without user controllable lints.

### DIFF
--- a/src/cargo/util/toml/targets.rs
+++ b/src/cargo/util/toml/targets.rs
@@ -76,24 +76,28 @@ pub(super) fn to_targets(
         normalized_toml.bin.as_deref().unwrap_or_default(),
         package_root,
         edition,
+        warnings,
     )?);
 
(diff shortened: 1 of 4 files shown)


ehuss (approved):
Thanks!

Can we close https://github.com/rust-lang/cargo/issues/15283, as I don't think there is anything else to do here?

epage:
> Can we close https://github.com/rust-lang/cargo/issues/15283, as I don't think there is anything else to do here?

Once we have the next edition, we need to remove support for the field.

PR #15325, opened 2025-03-18
test(build-std): address overly-matched snapshot

weihanglo:
Still trying to bisect what went wrong,
but this should be fairly safe to merge.

Failed on nightly-2025-03-18 but not nightly-2025-03-17 btw.


ehuss:
> Still trying to bisect what went wrong,

This was changed by https://github.com/rust-lang/rust/pull/136355 which is being reverted in https://github.com/rust-lang/rust/pull/138661.

PR #9557, opened 2021-06-08
Stabilize the rustc-link-arg option

danielframpton:
This change removes the unstable option (tracked by #9426) and unconditionally accepts additional linker arguments (as implemented in #7811 and #8441). Documentation is moved from unstable to what appeared to be the correct location.

I am not aware of any significant concerns with the option and it appears consistent with some other existing stable linker options.

Please let me know if this is not the appropriate process or if there is anything that I am missing from the PR.

PR #8441, opened 2020-07-02
Finish implementation of `-Zextra-link-arg`.

reitermarkus:
This is a continuation of https://github.com/rust-lang/cargo/pull/7811, which adds tests and a warning if the `-Zextra-link-arg` flag is not specified. Also moved the documentation into `unstable.md`.

PR #15326, opened 2025-03-18
Cleanup for rustc-link-arg-cdylib

ehuss:
This does a little bit of cleanup around rustc-link-arg-cdylib build script instruction:

* Reorders the documentation so that it is consistent.
* Updates the documentation to mention both the new (`rustc-link-arg-cdylib`) and old (`rustc-cdylib-link-arg`) forms are documented, with the new form as the primary form.
* Adds a test for the old form, since we didn't have any.

Somehow I missed all this in https://github.com/rust-lang/cargo/pull/8441 and https://github.com/rust-lang/cargo/pull/9557.

### DIFF
--- a/src/doc/src/reference/build-scripts.md
+++ b/src/doc/src/reference/build-scripts.md
@@ -105,7 +105,7 @@ one detailed below.
   to re-run the script.
 * [`cargo::rustc-link-arg=FLAG`](#rustc-link-arg) --- Passes custom flags to a
   linker for benchmarks, binaries, `cdylib` crates, examples, and tests.
-* [`cargo::rustc-cdylib-link-arg=FLAG`](#rustc-cdylib-link-arg) --- Passes custom
+* [`cargo::rustc-link-arg-cdylib=FLAG`](#rustc-cdylib-link-arg) --- Passes custom
   flags to a linker for cdylib crates.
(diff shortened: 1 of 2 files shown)


src/doc/src/reference/build-scripts.md
```diff
@@ -147,13 +147,16 @@ linker script.
 
 [link-arg]: ../../rustc/codegen-options/index.md#link-arg
 
-### `cargo::rustc-cdylib-link-arg=FLAG` {#rustc-cdylib-link-arg}
+### `cargo::rustc-link-arg-cdylib=FLAG` {#rustc-cdylib-link-arg}
```
weihanglo, line 150: Do we want to maintain the old anchor link?
I am fine wit either way.

weihanglo (approved):
Thanks!

epage:
Do we need to make any changes to `build-rs` api?

src/doc/src/reference/build-scripts.md
ehuss: Yea, I prefer to keep the old anchor, just to avoid breaking any potential links.

ehuss:
> Do we need to make any changes to `build-rs` api?

If you mean maybe rename [rustc_cdylib_link_arg](https://docs.rs/build-rs/latest/build/output/fn.rustc_cdylib_link_arg.html) to `rustc_link_arg_cdylib`, I don't have much of an opinion there. Maybe?

PR #15330, opened 2025-03-19
Use ellipses when truncating progress

WhyNotHugo:

Visual quirkiness in progress that doesn't fit on screen. Instead of using three periods at the end of the screen, use an ellipses.

Any regular operation on a narrow screen should make the difference evident.

### Additional information

n/a

### DIFF
--- a/src/cargo/util/progress.rs
+++ b/src/cargo/util/progress.rs
@@ -261,8 +261,7 @@ impl<'gctx> Progress<'gctx> {
     /// * `cur` should be how far along the progress is.
     /// * `max` is the maximum value for the progress bar.
     /// * `msg` is a small piece of text to display at the end of the progress
-    ///   bar. It will be truncated with `...` if it does not fit on the
-    ///   terminal.
+    ///   bar. It will be truncated with `…` if it does not fit on the terminal.
     ///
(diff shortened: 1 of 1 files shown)


src/cargo/util/progress.rs
```diff
@@ -521,20 +520,20 @@ impl Format {
...
                 avail_msg_len -= display_width;
                 string.push(c);
-                if avail_msg_len >= 3 {
+                if avail_msg_len >= 1 {
                     ellipsis_pos = string.len();
                 }
             } else {
                 string.truncate(ellipsis_pos);
-                string.push_str("...");
+                string.push_str("…");
```
epage, line 536: We do not unconditionally output unicode characters but need to check either `shell().out_unicode()` or `shell().err_unicode()`.

epage:
Thanks for contributing this!  It would be great to find ways to better leverage unicode output now that we can with `term.unicode` being added in #13337.

src/cargo/util/progress.rs
epage: > Unicode is a superset of ascii. I assume you mean non-ascii Unicode?

Yes

PR #15332, opened 2025-03-19
Fix `[env]` `relative` description in reference

MingweiSamuel:
Current short description does not match reality nor the longer description:
> Setting the relative flag evaluates the value as a config-relative path that is relative to the parent directory of the .cargo directory that contains the config.toml file. The value of the environment variable will be the full absolute path.


weihanglo (approved):
Thanks!

PR #15334, opened 2025-03-20
fix(build-dir): Renamed workspace-manifest-path-hash to workspace-path-hash

ranger-ross:

Small PR to rename `workspace-manifest-path-hash` to `workspace-path-hash` in the build-dir template as mentioned [here](https://github.com/rust-lang/cargo/issues/14125#issuecomment-2733611870) (cc: #14125)

r? @epage


epage (approved):
Thanks!

PR #14553, opened 2024-09-17
feat: Add custom completer for completing `cargo build --packge <TAB>`

shannmu:
### What does this PR try to resolve?
Tracking issue https://[github.com/rust-lang/cargo/issues/14520](https://github.com/rust-lang/cargo/issues/14520)

Add custom completer for `cargo build --package <TAB>`

PR #15338, opened 2025-03-21
feat: Add custom completer for completing cargo build --packge <TAB>

nooma-42:


epage:
In https://github.com/rust-lang/cargo/pull/14553#discussion_r1791937807, I expressed concern over keying off of the name of the command for deciding what kind of completions should be done.

epage:
This ended up being resolved by #16210.

PR #15340, opened 2025-03-21
feat(build-dir): Added the Cargo version to the inputs of `workspace-path-hash`

ranger-ross:

This PR adds the Cargo version to the hash inputs of `workspace-path-hash` build-dir template variable.
We only include the major, minor, and patch parts of the version in the hash inputs to avoid including the release channel causing new nightly's from creating many new build-dirs.

See this [comment](https://github.com/rust-lang/cargo/issues/14125#issuecomment-2733611870) in  #14125

I added a test for this change. 
The testing approach is to run `cargo build` twice changing the versions between runs and then validating the build-dir changed between runs.

### Additional information

For testing I had to add a `__CARGO_TEST_CARGO_VERSION` variable as that seemed to be the easiest way to test this. 
Adding `__CARGO_TEST_` env vars for testing purposes appears to be an established pattern (ie. `__CARGO_TESTS_ONLY_SRC_ROOT`, `__CARGO_TEST_MAX_UNPACK_SIZE`, etc)

r? @epage

### DIFF
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -676,7 +676,16 @@ impl GlobalContext {
                         .to_string(),
                 ),
                 ("{workspace-path-hash}", {
-                    let hash = crate::util::hex::short_hash(&workspace_manifest_path);
+                    // We include the version in the hash to prevent external tools from relying on
+                    // our hash implmentation. Note that we explictly do not include the prerelease
+                    // and build parts of the semver to avoid a new hash for every nightly version.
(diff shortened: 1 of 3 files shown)


epage (approved):

epage:
As noted at https://github.com/rust-lang/cargo/issues/14125#issuecomment-2751658701, the aspect of this PR for including the cargo version in the hash is being deferred.  I've not closed this in case you want to keep this PR for just the test change.

PR #15341, opened 2025-03-21
fix: revert the behavior checking lockfile's VCS status

### LINKED ISSUE #15339: Dirty Cargo.lock file reported on package/publish even if in .gitignore
### Problem

Cargo throws an error when trying to publish/package [alloy-rs/core](https://github.com/alloy-rs/core) on nightly (`cargo 1.87.0-nightly (6cf826701 2025-03-14)`), while it works fine on stable (1.85).

This breaks our cargo-release workflow, and I don't really understand this error since Cargo.lock is not dirty, is in .gitignore, and also adding it to Cargo.toml package.exclude doesn't help.

Bisected to nightly-2025-03-11 -> https://github.com/rust-lang/rust/pull/138200 -> probably https://github.com/rust-lang/cargo/pull/15276.

### Steps

```bash
cd /tmp
git clone https://github.com/alloy-rs/core.git alloy-core
# git checkout 5d9b4f22a9091749e87d89058185dd8cf5a03fa2
cd ./alloy-core

args=(package --list)
# No Cargo.lock
cargo +stable "${args[@]}"  # OK
cargo +nightly "${args[@]}" # OK

# Generate Cargo.lock, but still ignored by git
cargo generate-lockfile
cargo +stable "${args[@]}"  # OK
cargo +nightly "${args[@]}" # Fails with:
```

```
error: 1 files in the working directory contain changes that were not yet committed into git:

Cargo.lock

to proceed despite this and include the uncommitted changes, pass the `--allow-dirty` flag
```

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
cargo 1.87.0-nightly (6cf826701 2025-03-14)
```

weihanglo:

Fixes #15339

Lockfile might be gitignore'd,
So it never shows up in `git status`.

However, `cargo packag` vcs checks actually performs `git status --untracked --ignored`.
It is a bit confusing that lockfile is ignored but still counts as dirty from the report of `cargo package`.

There are some more nuances:
We check lockfile's VCS status if the lockfile is ouside the current package root. That means a non-workspace Cargo package will not have this VCS check. I don't think it is good that we have diverged behaviors Hence this revert.

We can always re-evaluate later,
as we've reserved rooms for doing more dirty checks.

https://github.com/rust-lang/cargo/issues/14967#issuecomment-2651329783

See if the revert is what we want now.
Or we should do a more thorough check for both workspace and non-workspace cases.

### Additional information

This is a revert of f0907fca16a84101d69fc53272a955e8bb53d9fe from #15276


src/cargo/ops/cargo_package/vcs.rs
epage, line 1: Note that f0907fca16a84101d69fc53272a955e8bb53d9fe has a comment change that was not reverted in this commit

epage (approved):

epage:
Feel free to merge when CI is green

PR #15344, opened 2025-03-21
Add a test for windows verbatim cwd

ehuss:
This adds a test for running cargo with a verbatim path on Windows. In particular, this checks for workspace globbing support.

This was incidentally fixed by https://github.com/rust-lang/cargo/pull/15166 which updated the glob crate. This was observed in https://github.com/oli-obk/cargo_metadata/issues/288.

You can downgrade the glob crate, or test against 1.85 to see that the test will fail with something like:

```
error: failed to load manifest for workspace member `\\?\C:\Proj\cargo\target\tmp\cit\t0\foo\crates\*`
referenced by workspace at `\\?\C:\Proj\cargo\target\tmp\cit\t0\foo\Cargo.toml`

Caused by:
  failed to read `\\?\C:\Proj\cargo\target\tmp\cit\t0\foo\crates\*\Cargo.toml`

Caused by:
  The filename, directory name, or volume label syntax is incorrect. (os error 123)

```

### DIFF
--- a/tests/testsuite/workspaces.rs
+++ b/tests/testsuite/workspaces.rs
@@ -2716,3 +2716,124 @@ fn nonexistence_package_together_with_workspace() {
 "#]])
         .run();
 }
+
+#[cargo_test]
+#[cfg_attr(not(windows), ignore = "testing windows canonicalization only")]
+fn workspace_glob_with_canonical_wd() {
+    // Tests the behavior of having a glob include with a canonical current
+    // working directory. On Windows this had issues with a verbatum path.
+    //
+    // Note that verbatim path handling is still unreliable in some
+    // situations, see https://github.com/rust-lang/cargo/issues/9770.
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [workspace]
+                members = ["crates/*"]
+            "#,
+        )
+        .file("crates/foo/Cargo.toml", &basic_manifest("foo", "1.0.0"))
+        .file(
+            "crates/foo/src/main.rs",
+            "fn main() { println!(\"hello world\"); }",
+        )
+        .build();
+
+    let cwd = p.root().canonicalize().unwrap();
+    // Ensure that canonicalize is giving us a verbatim path.
+    let std::path::Component::Prefix(prefix) = cwd.components().next().unwrap() else {
+        panic!("expected component")
+    };
+    assert!(matches!(prefix.kind(), std::path::Prefix::VerbatimDisk(_)));
+
+    p.cargo("metadata")
+        .cwd(&cwd)
+        .with_stdout_data(
+            str![[r#"
+{
+  "metadata": null,
+  "packages": [
+    {
+      "authors": [],
+      "categories": [],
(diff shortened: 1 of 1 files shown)


tests/testsuite/workspaces.rs
```diff
@@ -2716,3 +2716,124 @@ fn nonexistence_package_together_with_workspace() {
...
 }
+
+#[cargo_test]
+#[cfg_attr(not(windows), ignore = "testing windows canonicalization only")]
+fn workspace_glob_with_canonical_wd() {
+    // Tests the behavior of having a glob include with a canonical current
+    // working directory. On Windows this had issues with a verbatum path.
+    //
+    // Note that verbatim path handling is still unreliable in some
+    // situations, see https://github.com/rust-lang/cargo/issues/9770.
```
ChrisDenton, line 2727: I would just note that an actual current working directory *should not* be a verbatim path. As the docs for [`SetCurrentDirectory`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-setcurrentdirectory#remarks) say:

> Each process has a single current directory made up of two parts:
> - A disk designator that is either a drive letter followed by a colon, or a server name and share name (`\\servername\sharename`)
> - A directory on the disk designator

That is the path must be either in the form of `C:\path\to\dir` or `\\servername\sharename\path\to\dir`. A verbatim path (i.e. one that starts with `\\?\`) can break assumptions made by other OS APIs. It may work for a particular use but it will be fragile.

ehuss: Hm, yea... The problem is that people often naively do something like:

```rust
Command::new("cargo")
    .current_dir(path.canonicalize()?)
    .status()?;
```

If this seems too fragile, then I'm fine with just closing this PR.

ChrisDenton: That is something I'm considering fixing in the standard library. At the very least I think std should attempt to simplify the path if less than `MAX_PATH`. But doing that is semi-blocked on cleaning up our path handling, which already has a number of adhoc fixes for things.

epage:
For reference, another verbatim path issue we have is #13919.

tests/testsuite/workspaces.rs
ChrisDenton: I went ahead with making changes to `Command::current_dir` so that verbatim paths are usually turned into their non-verbatim equivalent  https://github.com/rust-lang/rust/pull/138869. I think that will solve the original issue.

I don't think cargo should error if the current directory is verbatim (though a warning may be appropriate). In pure rust code we should make a best effort to support this across all tools.

Where it gets tricky is with FFI and especially foreign processes. Attempting to make it work, as in this PR, does appeal to me. I'm just uncertain how well that will work out in practice. Even if Cargo makes it all work internally it can't control what other processes do.

Leaving it undefined is the safest option but somewhat unsatisfying. Having more test coverage for verbatim paths is appealing to me.

epage: Thanks!

PR #15345, opened 2025-03-21
Fix some issues with future-incompat report generation

### LINKED ISSUE #15337: [Diagnostics] --future-incompat-report gave me incorrect information and incorrect instructions
### Problem

I got a message at the end of my cargo output telling me to run `--future-incompat-report`. When I did this, I got a messsage which (1) misdescribed my package and (2) told me to run a *different* cargo invocation; this second cargo invocation simply gave me an error.

There turns out to be a different invocation which *did* work, which either the initial suggestion to run `--future-incompat-report` should have given me exclusively, or which --future-incompat-report should have aliased to.

### Steps

I have a small WIP program here:

https://github.com/mcclure/lisp0-experiment/tree/unstable

What it does isn't important because all we'll be doing is building it. Check out branch unstable, commit `35d955f995`. Build it.

**Observed behavior**

I did the following tests in Rust 1.83.0 and Rust 1.85.1 (the newest version rustup will give me) and saw identical behavior in each.

When I run `cargo build`, it prints at the end

```
warning: the following packages contain code that will be rejected by a future version of Rust: lisp0 v0.1.0 (/home/mcc/work/r/emj)
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 4
```

Gosh, that's great to know! I immediately run `cargo build --future-incompat-report`. Now it prints at the end:

```
warning: the following packages contain code that will be rejected by a future version of Rust: lisp0 v0.1.0 (/home/mcc/work/r/emj)
note: 
To solve this problem, you can try the following approaches:


- If the issue is not solved by updating the dependencies, a fix has to be
implemented by those dependencies. You can help with that by notifying the
maintainers of this problem (e.g. by creating a bug report) or by proposing a
fix to the maintainers (e.g. by creating a pull request):

  - lisp0@0.1.0
  - Repository: <not found>
  - Detailed warning command: `cargo report future-incompatibilities --id 5 --package lisp0@0.1.0`

- If waiting for an upstream fix is not an option, you can use the `[patch]`
section in `Cargo.toml` to use your own version of the dependency. For more
information, see:
https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section
        
note: this report can be shown with `cargo report future-incompatibilities --id 4`
```

Wat

As you can see from the Cargo.toml, "lisp0" is *the package I am building*. I cannot update the dependencies because it's not a dependency. I cannot email the maintainer because I am the maintainer.

Well, fine, I guess I just need to run the detailed warning. I run `cargo report future-incompatibilities --id 5 --package lisp0@0.1.0`, just as it suggests.

```
error: could not find report with ID 5
Available IDs are: 1, 2, 3, 4
```

*Wat*

I notice that the original message also listed `cargo report future-incompatibilities --id 4` as an option. I run that. It gives me a clear, concise explanation of the problem (to summarize: I used `macro_rules!` incorrectly). This "--id 4" message is exactly what I would have expected --future-incompat-report to print to begin with.

**Expected behavior:**

- The text "to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 4`" to me suggests that these two options are identical. If they are not identical, the text suggesting them should have clarified the difference.
- If I am building package `lisp0`, Rust diagnostics should not discuss the `lisp0` package as if it were a dependency.
- `--future-incompat-report` should not under any circumstances tell me to access a report with an id that does not exist.
- The text from `cargo report future-incompatibilities --id 4` is what `--future-incompat-report` should have printed.

### Possible Solution(s)

The text from `cargo report future-incompatibilities --id 4` is what `--future-incompat-report` should have printed.

### Notes

I don't know for a fact if this is being caused by cargo or rustc.

I have not tested with smaller test cases, so I don't know if this is a general problem with `--future-incompat-report` or somehow specific to my project.

### Version

```text
cargo 1.85.1 (d73d2caf9 2024-12-31)
release: 1.85.1
commit-hash: d73d2caf9e41a39daf2a8d6ce60ec80bf354d2a7
commit-date: 2024-12-31
os: Debian n/a (trixie) [64-bit]
```

ehuss:
This fixes various issues with the future-incompat report. In particular, it addresses the off-by-one for the `--id` flag when the report is already cached, and it doesn't recommend updating dependencies when the error comes from a local crate. See each commit for more details.

Fixes #15337

### DIFF
--- a/src/cargo/core/compiler/future_incompat.rs
+++ b/src/cargo/core/compiler/future_incompat.rs
@@ -483,7 +483,7 @@ You may want to consider updating them to a newer version to see if the issue ha
             "
 To solve this problem, you can try the following approaches:
 
-{update_message}
+{update_message}\
 - If the issue is not solved by updating the dependencies, a fix has to be
 implemented by those dependencies. You can help with that by notifying the
 maintainers of this problem (e.g. by creating a bug report) or by proposing a
--- a/src/cargo/core/compiler/future_incompat.rs
+++ b/src/cargo/core/compiler/future_incompat.rs
@@ -485,7 +485,7 @@ fix to the maintainers (e.g. by creating a pull request):
 section in `Cargo.toml` to use your own version of the dependency. For more
 information, see:
 https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section
-        ",
+",
         upstream_info = upstream_info,
         update_message = update_message,
     );
--- a/src/cargo/core/compiler/future_incompat.rs
+++ b/src/cargo/core/compiler/future_incompat.rs
@@ -140,16 +140,10 @@ impl OnDiskReports {
         mut self,
         ws: &Workspace<'_>,
         suggestion_message: String,
-        per_package_reports: &[FutureIncompatReportPackage],
+        per_package: BTreeMap<String, String>,
     ) -> u32 {
-        let per_package = render_report(per_package_reports);
-
-        if let Some(existing_report) = self
-            .reports
-            .iter()
-            .find(|existing| existing.per_package == per_package)
-        {
-            return existing_report.id;
+        if let Some(existing_id) = self.has_report(&per_package) {
+            return existing_id;
         }
 
         let report = OnDiskReport {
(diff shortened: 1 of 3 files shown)


weihanglo (approved):
Thanks for the fix and cleanup!

mcclure:
Cool

PR #15348, opened 2025-03-25
Temporarily ignore cargo_test_doctest_xcompile_ignores

ehuss:
This was broken due to a change in
https://github.com/rust-lang/rust/pull/138535. There is an approved PR in https://github.com/rust-lang/rust/pull/138877 to fix it, but it may take a day or two for it to make its way to nightly. This should be reverted after it hits nightly.

### DIFF
--- a/tests/testsuite/test.rs
+++ b/tests/testsuite/test.rs
@@ -4740,7 +4740,9 @@ fn test_dep_with_dev() {
         .run();
 }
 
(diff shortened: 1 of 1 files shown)


weihanglo (approved):

weihanglo:
The `cfg` stuff is annoying 😆. Should we just comment out the entire test temporarily?

weihanglo (approved):

PR #9991, opened 2021-10-21
Don't canonicalize executable path

Byron:
Otherwise symbolic links may also accidentally be resolved which may lead to unexpected results in the case of 'coreutils', a binary that depends on the executable name being a symbolic link.

This means a path like /bin/echo being canonicalized to /bin/coreutils will loose all information about the desired functionality.

For example, test failures will occur if 'echo' is resolved that way and it's not trivial to find the cause of it in the provided error messages.  For example`doc_workspace_open_different_library_and_package_names` did fail for me on MacOS, Nix packages in PATH, but works with this patch.

With this patch, there is still the possibility that a path gets canonicalized for its relative path components, but still results in changing the name of the binary. I could imagine to check for binary name changes and panic if `coreutils` or `busybox` is encountered, which are known to fail without a symlink telling them which program to emulate.

PR #15355, opened 2025-03-26
Don't canonicalize executable path in `cargo_exe`

ChrisDenton:

Follow up to #9991. The issue explains why but in short canonicalising executable paths causes issues for symlinks. It seems this was missed in #9991.

Note that the comment I deleted is wrong since #9991.

Running existing test should ensure this doesn't break anything.


weihanglo (approved):
This seems reasonable. How did you discover this though?

weihanglo:
Sorry I shouldn't click that Copilot review button… Was just curious 😆.

ChrisDenton:
I was investigating issues caused by the rustup shims now being symlinks by default in 1.28+ and this looked suspicious.

I'm not yet sure if this caused a real world breakage but it definitely looks like it could if `$CARGO` is set and cargo recursively calls into itself.

PR #15359, opened 2025-03-28
feat: rustdoc depinfo rebuild detection via -Zrustdoc-depinfo

### LINKED ISSUE #12266: `cargo doc` doesn't detect dirty sources from parent directories

weihanglo:

This leverages the unstable `--emit=depinfo` option from rustdoc,
so that rustdoc invocation rebuild can be better tracked
without traversing the entire directory.

Some design decisions made in the current implementation:

* Rustdoc's depinfo doesn't and shouldn't emit to `target/doc`,
  as the directory is considered part of the final artifact directory.
  In regard to that, we specify the dep-info output path to
  the fingerprint directory of rustdoc invocation.
  It looks like this
  `target/debug/.fingerprint/serde-12d29d32b3b8b38f/doc-lib-serde.d`.
* We also start supporting `-Zchecksum-freshness` as a side effect.
  Could make it a separate PR if desired.
* `-Zbinary-dep-depinfo` is not enabled along with this,
  since doc generations don't really require any binary dependencies.

The tests added has covered these cases:

* target src outside package root, e.g., `lib.path = "../lib.rs"`
* `#[doc = include_str!("../outside/pkgroot")]`
* `#[path = "../outside/pkgroot"]`
* `env!`

### Additional information

Fixes #12266
Closes #15205


src/cargo/core/compiler/mod.rs
```diff
@@ -763,6 +763,20 @@ fn prepare_rustdoc(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> CargoResu
     add_error_format_and_color(build_runner, &mut rustdoc);
     add_allow_features(build_runner, &mut rustdoc);
 
+    // TODO: create a dedicated unstable flag
```
epage, line 766: What do you mean by this TODO?

weihanglo: Should be removed

epage (approved):

epage:
Thanks!

Issue #15360, opened 2025-03-28
Cannot compile crate with cargo <1.71 if any crate in workspace depends on a crate that uses `dep:` syntax


nicoburns:
### Problem

There is a backwards compatibility hazard with `dep:` syntax in `Cargo.toml` files.

- Given a workspace with N crates
- If any of those crates depend (even transitively) on a crate that uses `dep:` syntax
- Then no crate in the workspace can be compiled with cargo/rustc older than 1.71

This means that a crate that wants to maintain a low MSRV cannot be combined in a workspace with:
- Other crates in the workspace with higher MSRV
- Examples that depend on crates with higher MSRV
- Tooling or scripts that depend on crates with higher MSRV

### Possible Solution(s)

Perhaps there could be a way to include crates in a workspace for the purpose of being accessible with `--package` and `--workspace` being possible to set it as a default crate to compile, but without dependencies being unified with the rest of the workspace?

Something like this would also be good for examples which I find generally work better as separate crates except for the fact that you then can't call them with `--example`.

### Notes

This can be worked around by not using workspaces and splitting the repo into separate Cargo projects. But this is not ideal.

### Version

```text
Any version of cargo older than 1.71 (1.71 is not affected, 1.70 is)
```

epage:
Yes, when working in a multi-MSRV workspace, you have to use the lowest common denominator for
- workspace member manifest files
- config files
- registry dependency index entries
- all other dependency manifest files and maybe their workspace manifest file

There has been some curiosity around allowing separate `Cargo.lock`s within a workspace. This was thought of more in the direction of the feature unification controls (all-or-nothing config for testing purposes). Supporting this for MSRV as well would require more drastic internal changes and maybe a switch to package-by-package controls which might then be a manifest field to control this.

In blessing workspace members to deviate from each other like this, we should also consider how it affects the whole workflow and how far in that direction do we want to take cargo (ie future possibility discussion, not blocking for full design). For example, groups of inheritable configuration has come up in the past.

epage:
Related issues:
- #6015
- #12385
- #13600
- #15360
- #15894

PR #15361, opened 2025-03-28
Set MSRV of home crate to 1.71

nicoburns:

Despite not being intended for external use, 667 crates depend directly on `home`. It would therefore be nice if it could be more conservative with MSRV than the other crates in this repo. This crate will compile with Rust 1.71 without any code changes, so I've set it to that.

Lower than 1.71 is blocked by https://github.com/rust-lang/cargo/issues/15360 (*this* crate could compile with lower MSRV than that, but dependencies of crates in this workspace use `dep:` syntax so cargo doesn't even get that far)


epage:
Please see the discussion in #13270. Since then, the MSRV-aware resolver has been released as well which won't help if your development version is your MSRV but those cases are likely rare.

nicoburns:
@epage I don't see how the MSRV-aware resolver helps here (until the MSRV of one's crate is greater than 1.84). `Cargo.lock` files from libraries don't apply to crates that depend on that library, so even if I apply an MSRV-aware resolution to my dependencies, then the end users of my crate will still get incompatible version resolved unless they are also using the MSRV-aware solver which they obviously won't be able to do if they need a lower MSRV.

I could pin a lower version of each dependency using `=` in `Cargo.toml` but I don't think there's an automated way to do that? (and I'd have to do it proactively for every dependency if I wanted to avoid my own MSRV being silently bumped when a dependency updates their MSRV). And even then, that wouldn't necessarily work for transitive dependencies.

Am I missing something?

I would also ask: is there really a meaningful cost to cargo of doing this? No code changes are required. It's literally just a case of not setting the `rust-version` key in the metadata to an artificially high version.

epage:
> @epage I don't see how the MSRV-aware resolver helps here (until the MSRV of one's crate is greater than 1.84). Cargo.lock files from libraries don't apply to crates that depend on that library, so even if I apply an MSRV-aware resolution to my dependencies, then the end users of my crate will still get incompatible version resolved unless they are also using the MSRV-aware solver which they obviously won't be able to do if they need a lower MSRV.

The interface for enabling the MSRV-aware resolver was intentionally designed to not require an MSRV bump.  It only requires that you update your *development* version of Cargo and use that when generating a `Cargo.lock`.  This works for people who can use a regular toolchain to develop locally and then deploy using a specific toolchain (e.g. from yocto linux).  If someone is doing development on a Linux distribution with an old version, then that won't work.  This is most likely to happen with desktop Linux distributions.  The general vibe I get from the team is that desktop Linux rust is really meant for building the distribution and people should be using rustup otherwise.  iirc distributions like Debian are now or soon will be offering rustup.

> I could pin a lower version of each dependency using = in Cargo.toml but I don't think there's an automated way to do that?

Doing that, especially in libraries, is also harmful to the ecosystem as it can cause incompatibilities between packages.  `Cargo.lock` is the right way to pin dependency versions and that can be managed via `cargo update --precise`

>  (and I'd have to do it proactively for every dependency if I wanted to avoid my own MSRV being silently bumped when a dependency updates their MSRV). And even then, that wouldn't necessarily work for transitive dependencies.

Yes, the MSRV experience for pre-MSRV-aware resolver is not great.  This is why I put so much emphasis on getting that work done.

> I would also ask: is there really a meaningful cost to cargo of doing this? No code changes are required. It's literally just a case of not setting the rust-version key in the metadata to an artificially high version.

This is not a new argument and it is not productive to re-litigate decisions with each new person who is concerned over something.  I'd recommend looking over the discussion in #13270 and our [MSRV documentation](https://doc.rust-lang.org/cargo/reference/rust-version.html#support-expectations) which we wrote after that PR

Issue #14292, opened 2024-07-23
Consider renaming `[gc]` config table.


ehuss:
In https://github.com/rust-lang/cargo/pull/14287#issuecomment-2243770766, it was suggested to have a different name for the `[gc]` table. The intent is to make it something clearer to the user.

My intent is to have the `frequency` option affect all cache cleaning. That is, I think there should be one option for configuring the "once per day" cleaning of caches. I would be concerned that trying to define properties for various different kinds of caches could add complexity. To that end, I might suggest something like:

```toml
[cache]
auto-clean-frequency = "1 day"
```

I don't have a particular opinion about the other config options, since I don't know what those will ultimately look like (tracked in #13061). In the meantime, I would toss them in some sub-table until a more permanent solution is picked. Perhaps something like `cache.global-clean` or something like that.

Feel free to suggest other ideas, preferably with your reasons for what you suggest.

weihanglo:
Maybe some sub-table like

```toml
[cache.global]
auto-clean-frequency = "1 day"
```

And someday later we got #5931 and #13061, we can expand to fine-grained control when needed.

```toml
[cache.global]
# The default for all cache. Overwritable.

[cache.index]
# for registry index cache

[cache.source]
# git or registry source

[cache.shared-artifacts]
# per-user cache

[cache.local-artifacts]
# workspace-local target dirs
```

epage:
I was talking with someone about [logging (~~tele~~ metry)](https://internals.rust-lang.org/t/no-telemetry-in-the-rust-compiler-metrics-without-betraying-user-privacy/19275).  If we do that, we'd need log rotation / reclamation support.

I could see our options being
- See this as another form of GC and put it in the GC table (with a top-level name that is more general than `cache`)
  - `auto` would also cover log rotation
- Want to consolidate all logging config in `.cargo/config.toml` under one table, including log rotation
  - Should `auto` then not cover log rotation?

I don't want to block work on this on future possibilities but wanted to add to the conversation "there might be resources to reclaim besides caches".  Do we want to limit `auto` to only caches?  I can see both sides and wanted others thoughts.

AustinHellerRepo:
I just wanted to add that I read that there was a proposal to "stabilize a minimal amount of garbage collection" and my heart sank. I said aloud "and this is the end of Rust". I forced myself to read further and realized that this is about file cleanup. Please reconsider using the phrase "garbage collection" as this strongly conflicts (on a verbiage-level) with one of the best features of the language.

[...]

PR #15367, opened 2025-03-30
Rename the gc config table

ehuss:
This renames the gc config table to `[cache]` to help avoid some confusion, and to set up a namespace for possible expansion in the future for different kind of cache controls.

Low-level settings are stuffed into the `[cache.global-clean]` table, but we do not expect to stabilize these at this time. Only the top-level `cache.auto-clean-frequency` setting is expected to be stabilized.

Closes https://github.com/rust-lang/cargo/issues/14292


weihanglo (approved):
Thanks. Looks good!

PR #15368, opened 2025-03-31
Prevent undeclared public network access

sebastianpoeplau:
Stop a few tests which aren't marked with `public_network_test` from trying to access crates.io.

### DIFF
--- a/tests/testsuite/cargo_add/normalize_name_path_existing/mod.rs
+++ b/tests/testsuite/cargo_add/normalize_name_path_existing/mod.rs
@@ -7,6 +7,7 @@ use cargo_test_support::Project;
 
 #[cargo_test]
 fn case() {
+    cargo_test_support::registry::init();
     let project = Project::from_template(current_dir!().join("in"));
     let project_root = project.root();
     let cwd = &project_root;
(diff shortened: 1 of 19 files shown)


tests/testsuite/cargo_add/normalize_name_path_existing/mod.rs
```diff
@@ -7,6 +7,9 @@ use cargo_test_support::Project;
 
 #[cargo_test]
 fn case() {
+    // Prevent network access by initializing the dummy registry.
```
epage, line 10: Don't think we need these comments; it should be used regardless

tests/testsuite/cargo_add/normalize_name_path_existing/mod.rs
epage, line 1: I'm seeing several other `cargo_add` tests (and maybe `cargo_info`?) that don't have an `init`.  Could you check them and add it?  If not, incremental improvement is better than no improvement.

tests/testsuite/cargo_add/normalize_name_path_existing/mod.rs
sebastianpoeplau: Removed.

tests/testsuite/cargo_add/normalize_name_path_existing/mod.rs
sebastianpoeplau: Done for `cargo_add` and `cargo_info`.

epage (approved):
Thanks!

PR #15369, opened 2025-03-31
fix(tree): Make output more deterministic

### LINKED ISSUE #15366: `cargo tree --duplicates --target all` non-deterministic output
### Problem

I have been working on reducing duplicate dependencies from a crate.
However I noticed that the output of `cargo -tree --duplicates --target all`
sometimes changes between printing a bunch of duplicates of the same version,
and printing `warning: nothing to print`

```
$ cargo tree --duplicates --target all

warning: nothing to print.

To find dependencies that require specific target platforms, try to use option `--target all` first, and then narrow your search scope accordingly.
```

I'm going to snip the output, but other times I see (what I believe different feature usage of a few crates between build-time and compile time) 

```
$ cargo tree --duplicates --target all

syn v2.0.100
...

syn v2.0.100
...

windows-sys v0.59.0
...

windows-sys v0.59.0
```

I would expect `cargo tree` to print deterministic output when invoked.
This isn't really a released version of the crate, I suppose I could make  a branch of the crate + patches required if this isn't a known issue and can't be reproduced easily.

The output seems to go through "phases", e.g. If you invoke it fast enough
you'll get a few instances of duplicates, followed by a few info instances of "warning: nothing to print".

But doesn't seem to be affected by whether rust-analyzer is running. 

cargo 1.85.0 (d73d2caf9 2024-12-31)

### Steps

_No response_

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text

```

epage:
When you have a dependency show up in multiple dependency tables, the order they are processed was not deterministic. Its hard to predict the exact effects this was having (e.g. for writing tests for this) but changing this made #15366 not reproduce anymore for me.

I'm going to consier this as fixing #15366 mostly because how difficult it is to debug with the non-determinism.
If someone can find a case where this fails, at least it should now always fail and it should be easier to determine why.

Fixes #15366

### DIFF
--- a/src/cargo/ops/tree/graph.rs
+++ b/src/cargo/ops/tree/graph.rs
@@ -478,7 +478,7 @@ fn add_pkg(
             continue;
         }
 
-        deps.sort_unstable_by_key(|dep| dep.name_in_toml());
+        deps.sort_unstable_by_key(|dep| (dep.kind(), dep.name_in_toml()));
         let dep_pkg = graph.package_map[&dep_id];
 
         for dep in deps {


weihanglo (approved):
Reasonable. Thanks!

Issue #13460, opened 2024-02-19
Be able to disable the certificate verification in libcurl


T00mm:
### Problem

While spinning up development images that don't have our enterprise CA included we get the following error on any cargo command that needs to contact the crates-io registry

```
Caused by:
    download of config.json failed
Caused by:
    [60] SSL peer certificate or SSH remote key was not OK (SSL certificate problem: unable to get local issuer certificate)
```

Normally you can bypass this by adding insecure to the cURL config file or by setting an environmental variable but the Easy library of cURL limits the environmental variables to a very limited set which `SSL_VERIFY_PEER` is not one of.

Would it be possible to be able to set the `ssl_verify_peer` option as well through the `Cargo.toml` config?

### Proposed Solution

1. An option in the config with verify_peer set to true as default
```rust
pub struct CargoHttpConfig {
...
    pub no_verify_peer: Option<bool>,
...
}
```

2. Add code the the handle

```rust
pub fn configure_http_handle(config: &Config, handle: &mut Easy) -> CargoResult<HttpTimeout> {
...
    if let Some(no_verify_peer) = http.no_verify_peer {
        handle.ssl_verify_peer(!no_verify_peer)?;
    }
...
```

weihanglo:
Understand the situation. Instead of adding options one by one, is it possible to add an `http.insecure` option to mimic what [`curl --insecure`](https://curl.se/docs/manpage.html#-k)  is doing?

There is also a [`--proxy-insecure`](https://curl.se/docs/manpage.html#--proxy-insecure) we might want to copy.

weihanglo:
Hi. The cargo team has discussed this today. The team is curious to know the situation of needing such an escape hatch. An image that didn't include enterprise CA doesn't sound normal, and many other software would also stop working with the same situation. Could you say more about why you can't configure your own certificates, given that having a self-signed certificate is generally a safer way to fix the issue.

T00mm:
I feel the current set-up limits you too much with these constraints. If the user is aware of the risks, my believe is that (s)he should be allowed to decide how the security should be enforced. 

With GIT you can change it by editting the (git) config so I was expecting that for cURL it would be the same.

The issue came from when I wanted to show some co-workers a quick demo but could not even get past this error. That we then spend a good hour or two with figuring out which certificate should be imported, which format we needed to convert it to, which specific folder and CA-update command that flavour of Linux needed... 

We were spending more time with getting the certificate on an image and importing it then the whole demo would have taken.

[...]

Issue #1180, opened 2015-01-16
Cargo doesn't respect git's http.sslverify


alexcrichton:
This used to be supported in libgit2, but [it was removed](https://github.com/libgit2/libgit2/commit/41698f22). We should probably add back in explicit support somewhere.

cc #636, this has come up once or twice there.

alexcrichton:
This will likely be done by binding the [`certificate_check`](https://libgit2.github.com/libgit2/#HEAD/type/git_remote_callbacks) callback in [`remote_callbacks.rs`](https://github.com/alexcrichton/git2-rs/blob/master/src/remote_callbacks.rs) of [git2-rs](https://github.com/alexcrichton/git2-rs) and then using that in Cargo itself

yug:
It seems a little bit complicated. 

We are behind the R&D proxy, the company classifies the connection zones, we are unfortunately in a zone we can git clone the outside code, no write access to outside.

Cargo should get work by hacked locally, however, the process is that it firstly download a pre-built binary, that init binary couldn't work, so, all things blocked. I am not sure Cargo support cross-compile, if it does not we had to find a linux 64bit box without proxy access problem, build a hacked cargo, substitute the pre-built with that hacked, then get whole things work.

BTW,  there should be configuration item to tell cargo to ignore the SSL_certificate_check, rather than people to hack an ad-hoc version

The problem is not solved, I will try it and update feedback here

the-shank:
@yug any updates on this one? I am behind a http proxy and am unable to use cargo

thebitllc:
I hardcoded the hostnames on my windows box and it worked!

c:\windows\system32\drivers\etc\hosts

174.129.244.175	www.crates.io crates.io s3-us-west-1-r-w.amazonaws.com

alexcrichton:
@thebitllc hm that's interesting, if you don't have that hardcoding, does `crates.io` resolve to a different IP than it normally does?

chrisxue815:
I managed to bypass this problem by setting up a crates.io proxy and a local index repo. See detailed description in [chrisxue815/crates.io-proxy](https://github.com/chrisxue815/crates.io-proxy).

guillon:
Related issue on my side, actually my company has a HTTP/HTTPS proxy service which fakes SSL certificates for doing HTTPS session monitoring (i.e. decrypts https on the fly). To achieve this the proxy provider  fakes https certificates by returning its own certificate to the client.
[...]

PR #15374, opened 2025-04-01
Add `http.proxy-cainfo` config for proxy certs

koxu1996:
This adds a `http.proxy-cainfo` option to Cargo which reads CA information from a bundle to pass through to the underlying `libcurl` call. This should allow configuration of Cargo in situations where SSL proxy is used.

Similar to https://github.com/rust-lang/cargo/pull/2917.

### DIFF
--- a/src/cargo/util/context/mod.rs
+++ b/src/cargo/util/context/mod.rs
@@ -2608,6 +2608,7 @@ pub struct CargoHttpConfig {
     pub low_speed_limit: Option<u32>,
     pub timeout: Option<u64>,
     pub cainfo: Option<ConfigRelativePath>,
+    pub proxy_cainfo: Option<ConfigRelativePath>,
     pub check_revoke: Option<bool>,
     pub user_agent: Option<String>,
     pub debug: Option<bool>,
--- a/src/cargo/util/network/http.rs
+++ b/src/cargo/util/network/http.rs
@@ -61,6 +61,10 @@ pub fn configure_http_handle(gctx: &GlobalContext, handle: &mut Easy) -> CargoRe
         let cainfo = cainfo.resolve_path(gctx);
         handle.cainfo(&cainfo)?;
     }
+    if let Some(proxy_cainfo) = &http.proxy_cainfo {
+        let proxy_cainfo = proxy_cainfo.resolve_path(gctx);
+        handle.proxy_cainfo(&format!("{}", proxy_cainfo.display()))?;
+    }
     if let Some(check) = http.check_revoke {
         handle.ssl_options(SslOpt::new().no_revoke(!check))?;
     }
--- a/src/doc/src/reference/config.md
+++ b/src/doc/src/reference/config.md
@@ -107,6 +107,7 @@ ssl-version.min = "tlsv1.1" # minimum TLS version
 timeout = 30                # timeout for each HTTP request, in seconds
 low-speed-limit = 10        # network timeout threshold (bytes/sec)
 cainfo = "cert.pem"         # path to Certificate Authority (CA) bundle
+proxy-cainfo = "cert.pem"   # path to proxy Certificate Authority (CA) bundle
 check-revoke = true         # check for SSL certificate revocation
 multiplexing = true         # HTTP/2 multiplexing
 user-agent = "…"            # the user-agent header
@@ -707,6 +708,14 @@ Sets the timeout for each HTTP request, in seconds.
 
 Path to a Certificate Authority (CA) bundle file, used to verify TLS
 certificates. If not specified, Cargo attempts to use the system certificates.
+
+#### `http.proxy-cainfo`
+* Type: string (path)
+* Default: none
+* Environment: `CARGO_HTTP_PROXY_CAINFO`
+
+Path to a Certificate Authority (CA) bundle file, used to verify proxy TLS
+certificates.
 
 #### `http.check-revoke`
 * Type: boolean
--- a/src/doc/src/reference/environment-variables.md
+++ b/src/doc/src/reference/environment-variables.md
@@ -105,6 +105,7 @@ In summary, the supported environment variables are:
 * `CARGO_HTTP_PROXY` --- Enables HTTP proxy, see [`http.proxy`].
 * `CARGO_HTTP_TIMEOUT` --- The HTTP timeout, see [`http.timeout`].
 * `CARGO_HTTP_CAINFO` --- The TLS certificate Certificate Authority file, see [`http.cainfo`].
+* `CARGO_HTTP_PROXY_CAINFO` --- The proxy TLS certificate Certificate Authority file, see [`http.proxy-cainfo`].
 * `CARGO_HTTP_CHECK_REVOKE` --- Disables TLS certificate revocation checks, see [`http.check-revoke`].
 * `CARGO_HTTP_SSL_VERSION` --- The TLS version to use, see [`http.ssl-version`].
 * `CARGO_HTTP_LOW_SPEED_LIMIT` --- The HTTP low-speed limit, see [`http.low-speed-limit`].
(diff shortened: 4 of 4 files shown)


src/cargo/util/context/mod.rs
weihanglo, line 1: Thanks. This looks reasonable, though I saw your comment in <https://github.com/rust-lang/cargo/issues/13460#issuecomment-2748973441>, and wonder what people want.
Do people really need the ability to configure custom proxy CA info, or just want to disable all the CA checks? 

There is also git's CA check discussion in <https://github.com/rust-lang/cargo/issues/1180> as well, somehow relevant.

weihanglo: See <https://github.com/rust-lang/cargo/issues/13460#issuecomment-1991902581>.

Let's discuss over there :)

koxu1996: I don't want to disable CA checks - although I think it is really useful option, definitely worth introducing.

koxu1996: I faced the problem where `cargo` doesn't work with an SSL proxy, even though the custom CA is trusted (`curl` doesn't complain about the certificates). This stems from the fact that the automatically vendored `libcurl` has no built-in information about the system trust store. I wrote a fairly detailed description of this issue in #15376; the ability to set a proxy cainfo is a real need to solve it.

koxu1996: @weihanglo Do you need more info?

weihanglo: I would say that the change standalone looks reasonable. Have you tested it satisfies and fixes issues on your side?

I would kick off a straw poll for the Cargo team, as this would add a new insta-stable config field to Cargo.

weihanglo:
@rfcbot fcp merge

I propose to the team that we merge and insta-stabilize the `http.proxy-cainfo` config.

### What is being stabilized

The `http.proxy-cainfo` config. It provides a way to set Certificate Authority (CA) bundle independently to `http.cainfo`  which also affects non-proxy requests.

### How is this useful

This `http.proxy-cainfo` and `http.cainfo` separation is useful in environments where the proxy uses an internal or self-signed certificate that should not be trusted for all HTTPS connections. For example, in a corporate network, you can use `http.proxy-cainfo` to trust the proxy’s certificate without compromising how Cargo validates crates.io or other HTTPS hosts.

* Without `http.proxy-cainfo`, adding internal CA to `http.cainfo` will give it too much trust. For example, if the proxy is comproised, with cainfo separations, traffic to crates.io would not be affected or will reject the connection.
* Changing System CA is usually not an option for non-root user
* Disable certificate checks completely is not in scope of this discussion.

joshtriplett:
@rfcbot concern need-rationale-other-than-curl-bug

According to https://github.com/rust-lang/cargo/issues/15376 , it sounds like the `curl` command-line tool automatically uses the non-proxy CA info as the proxy CA info, if no proxy CA info is specifically provided. This behavior doesn't occur in libcurl, or in other clients like cargo.

I think it'd be appropriate for cargo to implement the same behavior as the cargo command-line tool. That would make this use case work.

There may still be a reason to have this separate configuration, but we'd need a separate motivation for it *other* than the case where the proxy needs a system CA certificate and this config is being used to work around us not automatically using that.

koxu1996:
`http.cainfo` was [introduced](https://github.com/rust-lang/cargo/commit/8e8a2924aaef0086e534dc8c88d7ccecdf1c5eb3) with the following motivation:
> This should allow configuration of Cargo in situations where the default certificate store doesn't contain the relevant certificates, such as behind corporate proxies.

Given that proxies - especially in corporate environments - are expected to use TLS for secure communication, introducing `http.proxy_cainfo` is a reasonable extension that supports this more secure setup.

arlosi:
Regardless of whether we decide to add `proxy_cainfo` as a separate configuration option, I think cargo should match curl's behavior of using `cainfo` for `proxy_cainfo` (if an explicit `proxy_cainfo` is not set).

arlosi:
@koxu1996 are you available to update this PR implementation and docs such that `proxy_cainfo` defaults to the same value as `ca_info` as suggested above?

koxu1996:
@arlosi I can do that, but next week :ok_hand:.

koxu1996:
@arlosi Fallback added in https://github.com/rust-lang/cargo/pull/15374/commits/9a77461bdc8e822a3a1935861eb551a94052ca0f.

src/cargo/util/network/http.rs
```diff
@@ -61,6 +61,16 @@ pub fn configure_http_handle(gctx: &GlobalContext, handle: &mut Easy) -> CargoRe
         let cainfo = cainfo.resolve_path(gctx);
         handle.cainfo(&cainfo)?;
     }
+    // Use `proxy_cainfo` if explicitly set; otherwise, fall back to `cainfo` as curl does #15376.
+    let effective_proxy_cainfo = match (&http.proxy_cainfo, &http.cainfo) {
+        (Some(p), _) => Some(p),
+        (None, Some(ca)) => Some(ca),
+        _ => None,
+    };
+    if let Some(proxy_cainfo) = effective_proxy_cainfo {
```
weihanglo, line 70: This should be sufficient.

```suggestion
    if let Some(proxy_cainfo) = http.proxy_cainfo.as_ref().or(http.cainfo.as_ref()) {
```

src/cargo/util/network/http.rs
```diff
@@ -61,6 +61,16 @@ pub fn configure_http_handle(gctx: &GlobalContext, handle: &mut Easy) -> CargoRe
...
+        let proxy_cainfo = proxy_cainfo.resolve_path(gctx);
+        handle.proxy_cainfo(&format!("{}", proxy_cainfo.display()))?;
```
weihanglo, line 72: I feel like it is a bug in curl-rust, though the implementation here is fine.

src/cargo/util/network/http.rs
koxu1996: That is what I got on the first attempt, but wanted to make the fallback more visible. Accepted anyway.

weihanglo (approved):
Would you mind fixing up these commits and removing some intermediate one? Personally prefer every commit passes tests for better git bisect experience in the future.

weihanglo:
cc @joshtriplett if the fallback implementation and the reply address your concern

koxu1996:
> Would you mind fixing up these commits and removing some intermediate one? Personally prefer every commit passes tests for better git bisect experience in the future.

Cleanup done, now there are 2 commits:

- https://github.com/rust-lang/cargo/pull/15374/commits/ca4a3b1f8b6e39861ebaa08191cbd59723dcc28d - Add `http.proxy-cainfo` config for proxy certs
- https://github.com/rust-lang/cargo/pull/15374/commits/5af783129a688c730a018507f59cb2f06c3fc150 - Make `http.proxy-cainfo` fallback to `http.cainfo`, like curl.

joshtriplett:
OK, if I understand correctly, the rationale here is that you want to use a given CA for the proxy, but *not* permit that CA for the server accessed through the proxy?

Assuming this is an accurate summary of the use case:

@rfcbot resolved need-rationale-other-than-curl-bug

weihanglo:
I updated [the FCP text](<https://github.com/rust-lang/cargo/pull/15374#issuecomment-2791032749>) to capture what we've discussed today, and some scenarios are added. Please take a look.

koxu1996:
> OK, if I understand correctly, the rationale here is that you want to use a given CA for the proxy, but not permit that CA for the server accessed through the proxy?

@joshtriplett No. I just want to use TLS connection to the proxy, which uses custom CA. This is impossible when libcurl gets vendored in `cargo` binary - this is quite common case, all `rustup` users are affected. Adding cert to the OS trust store will NOT help.

koxu1996:
@weihanglo Ready for merge?

weihanglo (approved):
Yeah I believe this is ready

koxu1996:
Clippy check failing?

PR #15377, opened 2025-04-02
Added `build_directory` field to cargo metadata output

ranger-ross:

This PR continues on the build-dir work laid in #14125. (see [this](https://github.com/rust-lang/cargo/issues/14125#issuecomment-2751658701) comment)

We add a new `build_directory` field to the `cargo metadata` output when the `-Z build-dir` feature flag is enabled.

I added a test for metadata output when build-dir is set. There are already many existing tests that verify that build-dir is not included if `-Z build-dir` is not passed.

### Additional information

NOTE: I use `build_directory` instead of `build-dir` to match the existing `target_directory` field in the metadata output

r? @epage


weihanglo (approved):
Looks good. Thank you!

PR #15400, opened 2025-04-06
Added symlink resolution for `workspace-path-hash`

ranger-ross:

This PR adds logic to resolve symlinks before hashing the `workspace-path-hash`  template variable for `build.build-dir`.
See https://github.com/rust-lang/cargo/issues/14125#issuecomment-2751658701

cc: #14125

Note: The behavior on unix systems is unchanged, as the manifest_path was already canonicalized (at least on my system and in CI). However, the Windows behavior did not do this previous.

I added a test which runs `cargo build` twice, once from the real directory and once from inside of a symlinked directory, then verifies that hashes match.

The change is only a few lines. Most of the diffs are testing code

r? @epage

### DIFF
--- a/tests/testsuite/build_dir.rs
+++ b/tests/testsuite/build_dir.rs
@@ -11,8 +11,8 @@
 
 use std::path::PathBuf;
 
-use cargo_test_support::prelude::*;
 use cargo_test_support::{paths, project, str};
+use cargo_test_support::{prelude::*, Project};
 use std::env::consts::{DLL_PREFIX, DLL_SUFFIX, EXE_SUFFIX};
 
 #[cargo_test]
@@ -569,7 +569,7 @@ fn template_cargo_cache_home() {
 }
 
 #[cargo_test]
-fn template_workspace_manfiest_path_hash() {
+fn template_workspace_path_hash() {
     let p = project()
         .file("src/main.rs", r#"fn main() { println!("Hello, World!") }"#)
         .file(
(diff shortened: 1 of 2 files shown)


tests/testsuite/build_dir.rs
```diff
@@ -609,6 +609,87 @@ fn template_workspace_manfiest_path_hash() {
...
+/// directory and then compare the build-dir paths.
+#[cargo_test]
+fn template_workspace_path_hash_should_handle_symlink() {
+    #[cfg(unix)]
+    use std::os::unix::fs::symlink;
+    #[cfg(windows)]
+    use std::os::windows::fs::symlink_dir as symlink;
+
+    let p = project()
+        .file("src/main.rs", r#"fn main() { println!("Hello, World!") }"#)
```
weihanglo, line 624: For faster test execution time.

```suggestion
        .file("src/lib.rs", "")
```

tests/testsuite/build_dir.rs
```diff
@@ -609,6 +609,87 @@ fn template_workspace_manfiest_path_hash() {
...
+            ".cargo/config.toml",
+            r#"
+             [build]
+             build-dir = "foo/{workspace-path-hash}/build-dir"
+             "#,
+        )
+        .build();
+
+    // Build from the non-symlinked directory
+    p.cargo("build -Z build-dir")
```
weihanglo, line 645: Check for faster test execution time.

```suggestion
    p.cargo("check -Z build-dir")
```

weihanglo (commented):
Thanks. Just some nits not really directly related to the change itself.

crates/cargo-test-support/src/lib.rs
```diff
@@ -504,8 +504,20 @@ impl Project {
...
+        self.cargo_with_working_dir(cmd, &self.root())
+    }
+
+    /// Creates a `ProcessBuilder` to run cargo in a given directory.
+    /// [`Project::cargo`] is preferred when executing from the project root.
+    ///
+    /// Arguments can be separated by spaces.
+    ///
+    /// For `cargo run`, see [`Project::rename_run`].
+    pub fn cargo_with_working_dir(&self, cmd: &str, dir: &PathBuf) -> Execs {
```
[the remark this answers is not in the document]
ranger-ross: Thanks, I was looking for this method but I was failing to find it. 
(My instinct was to call `.cwd()` before `.cargo()` but that is not possible :))

I removed the commit with this helper and moved to `.cwd` instead

weihanglo (approved):
Thanks for working on this!

PR #15404, opened 2025-04-06
docs(ref): Use better example value in `CARGO_CFG_TARGET_ABI`

madsmtm:
`target_abi = "sim"` may be deprecated in the future. See https://github.com/rust-lang/rust/pull/139451.


ehuss:
Is there any particular reason to wait for the above PR?

madsmtm:
Actually no, we can do this now

weihanglo (approved):
Thanks!

PR #15405, opened 2025-04-06
feat: print target and package names formatted as file hyperlinks

### LINKED ISSUE #15401: Use file hyperlinks when outputting target names when running `cargo run --bin` and the like.
### Problem

In a project with multiple binaries, examples, tests, benches or packages commands like `cargo run`, `cargo run --bin` `cargo run --example`, `cargo test --bench` will generate an error with the list of available targets and remind the user to pick one of them. This is a good user experience as it helps the user easily discover what targets are available. I think this experience could be improved if the targets where outputted using ANSI hyperlinks, where package targets would be formatted to have a hyperlink to the file path of the crates `Cargo.toml` and binary targets like `--bin` and `--example` would be the file path to the file containing its `main` function. With this the user can easily click on the target if it's unclear where the target orginates from or how its defined and have the file open in their editor.

### Proposed Solution

Change `print_available_targets()` and `print_available_packages()` in `src/cargo/util/workspace.rs` to use `Shell` to format each target as a hyperlink.

### Notes

_No response_

kpbaks:
Resolves #15401

Here is an example of the feature in [kitty](https://sw.kovidgoyal.net/kitty/) `0.40.1` with the following config set in `~/.config/kitty/kitty.conf`

```conf
underline_hyperlinks always
show_hyperlink_targets yes
allow_hyperlinks yes
# ...
```
cargo-target-file-hyperlik-kitty-showcase

Tested on `uname -a`:
```
Linux nixos 6.14.0 #1-NixOS SMP PREEMPT_DYNAMIC Mon Mar 24 14:02:41 UTC 2025 x86_64 GNU/Linux
```
Terminals tested with:
- [x] [kitty](https://sw.kovidgoyal.net/kitty/) `0.40.1`
- [x] [ghostty](https://ghostty.org/) `1.1.4-6f1b22a-nix`
image
- [x] [alacritty](https://alacritty.org/index.html) `0.15.1`
image

- [x] VScode's version `1.98` integrated terminal aka. [xterm.js](https://xtermjs.org/) 

image

The following `cargo` invocations will have their output be modified by this change:
```shell
cargo run # If multiple binaries are defined in the manifest and [package.default-bin] is not defined
cargo run --bin
cargo run --example
cargo run --package
cargo build --bin
cargo build --example
cargo build --package
cargo test --test
cargo test --bench
```

In addition I have done a slight refactor to have the printed indent of targets and packages be the same by using a shared constant named `const ITEM_INDENT: &str = "    ";`

This is my first PR to the cargo codebase, so I am not familiar with what is expected in terms of test for a feature such as this.

### DIFF
--- a/src/cargo/util/workspace.rs
+++ b/src/cargo/util/workspace.rs
@@ -7,13 +7,16 @@ use anyhow::bail;
 use cargo_util::paths::normalize_path;
 use cargo_util::ProcessBuilder;
 use std::fmt::Write;
+use std::path::Path;
 use std::path::PathBuf;
+
+const ITEM_INDENT: &str = "    ";
 
 fn get_available_targets<'a>(
     filter_fn: fn(&Target) -> bool,
     ws: &'a Workspace<'_>,
     options: &'a CompileOptions,
-) -> CargoResult<Vec<&'a str>> {
+) -> CargoResult<Vec<(&'a str, &'a Path)>> {
     let packages = options.spec.get_packages(ws)?;
 
     let mut targets: Vec<_> = packages
@@ -24,7 +27,12 @@ fn get_available_targets<'a>(
                 .iter()
                 .filter(|target| filter_fn(target))
         })
-        .map(Target::name)
+        .map(|target| {
+            (
+                target.name(),
+                target.src_path().path().expect("Target is not a `Metabuild` but one of `Bin` | `Test` | `Bench` | `ExampleBin`")
+            )
+        })
         .collect();
 
     targets.sort();
@@ -48,8 +56,10 @@ fn print_available_targets(
         writeln!(output, "No {} available.", plural_name)?;
     } else {
         writeln!(output, "Available {}:", plural_name)?;
-        for target in targets {
-            writeln!(output, "    {}", target)?;
+        let mut shell = ws.gctx().shell();
+        for (name, src_path) in targets {
+            let link = shell.err_file_hyperlink(src_path);
+            writeln!(output, "{ITEM_INDENT}{link}{}{link:#}", name)?;
         }
     }
     bail!("{}", output)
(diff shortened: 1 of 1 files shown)


tcanabrava:
this would also work on vte and konsole.

src/cargo/util/workspace.rs
weihanglo, line 1: On macOS:

* ghostty 1.0.1 opened the default editor.
* kitty 0.38.1 opened the default editor.
* iterm2 3.5.11 didn't render any link, even the old existing ones. Anyway, I tried
  ```
  printf '\e]8;;file:///Users/user/foo/Cargo.toml\e\\This is a link\e]8;;\e\\'
  ```
  It tried to scp at the first attempt as we found in <https://github.com/rust-lang/cargo/pull/12889#issuecomment-1785563071>. However in subsequent attempts it worked correctly and opened the default editor.
* alacritty 0.15.1 didn't render any link just like iterm2. With the same test link above it worked and opened the default editor.


@ehuss would you happen to know why iterm2 doesn't really render those links? `Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s` should also have a rendered link, but on iterm2 I didn't see it.

weihanglo (approved):
Looks good to me! Thanks you.

src/cargo/util/workspace.rs
weihanglo: I feel like this is a two-way door decision, and we already provide a way to disable the behavior, so the bar accepting this should be fairly low. However, as @ehuss has raised a concern of iterm2, I'd like to hear opinions from them :)

ehuss: IIRC, it is disabled via https://github.com/rust-lang/cargo/blob/c7be0601a74a01f3f62e6259292c15d2f66cc531/src/cargo/core/shell.rs#L584.

weihanglo: Yeah and by turning hyperlink on always, both alacritty and iTerm2 work for me. I am going to merge this since iTerm2 is already always disabled.

PR #14839, opened 2024-11-19
Docs for optional registry JSON fields

kornelski:
Follow up to #14838

I haven't added info about 1.19, because [versions below 1.34.0 have zero users](https://p.datadoghq.com/sb/3a172e20-e9e1-11ed-80e3-da7ad0900002-973f4c1011257befa8598303217bfe3a?fromUser=false&refresh_mode=sliding&from_ts=1732028044807&to_ts=1732042444807&live=true).

I assume it's going to land in 1.84.

PR #15408, opened 2025-04-08
docs: Remove statement that dep features are optional

floers:
I created my own registry by using a git repo and found that leaving the `features` in the dependencies in an index entry does not work. When a crate A has dependencies with features and you try to use it via the registry in crate B the build fails when A uses symbols only available by said features.

### DIFF
--- a/src/doc/src/reference/registry-index.md
+++ b/src/doc/src/reference/registry-index.md
@@ -139,7 +139,6 @@ explaining the format of the entry.
             // https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html.
             "req": "^0.6",
             // Array of features (as strings) enabled for this dependency.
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/registry-index.md
```diff
@@ -139,7 +139,6 @@ explaining the format of the entry.
             // https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html.
             "req": "^0.6",
             // Array of features (as strings) enabled for this dependency.
-            // May be omitted since Cargo 1.84.
```
weihanglo, line 142: The doc might be a bit ambiguous. By looking at #14838 and #14839, this omitting means, if empty then you can omit this.

If a package has a features, then the field cannot be omitted.

PR #15410, opened 2025-04-09
docs(metadata): Added build_directory to cargo metadata documentation

ranger-ross:

Follow up to #15377 that documents `build_directory` in the cargo metadata docs.

I did not mention that this feature is not stable yet, let me know if it would be better to a note that this is nightly only.

cc: #14125


src/doc/man/cargo-metadata.md
```diff
@@ -318,8 +318,10 @@ The JSON output has the following format:
         */
         "root": "file:///path/to/my-package#0.1.0",
     },
-    /* The absolute path to the build directory where Cargo places its output. */
+    /* The absolute path to the target directory where Cargo places its output. */
     "target_directory": "/path/to/my-package/target",
+    /* The absolute path to the build directory where Cargo places intermediate build artifacts. */
```
weihanglo, line 323: Could we put an `(unstable)` at the end of this?

weihanglo: Just like other CLI help texts. If the doc is going to show up in the stable channel, we do note them :)

PR #15416, opened 2025-04-10
fix(package): detect dirtiness for symlinks to submodule 

### LINKED ISSUE #15384: 1.86 `cargo package` fail when workspace crate contains symlink into git submodule
### Problem

Previously it was possible to symlink a shared directory (e.g. testdata) into multiple crates from the root workspace directory where it exists as a git submodule.

Now in the cargo version that comes with rust-1.86, trying to use `cargo publish` or `cargo package` returns nonexistent file errors such as the following:

`error: attempt to get status of nonexistent file 'testdata/toml/dep.toml'; class=Invalid (3); code=NotFound (-3)`

Reverting to an earlier version of rust alleviates the issue.

### Steps

For reference, I'm running into the issue with the project at https://github.com/pkgcraft/pkgcraft that symlinks its testdata directory into multiple crates. See the following steps to reproduce the issue:

1. git clone --recursive https://github.com/pkgcraft/pkgcraft.git
2. cd pkgcraft
3. git checkout pkgcraft-tools-0.0.25
4. cargo package -p pkgcraft-tools

### Version

```text
cargo 1.86.0 (adf9b6ad1 2025-02-28)
release: 1.86.0
commit-hash: adf9b6ad14cfa10ff680d5806741a144f7163698
commit-date: 2025-02-28
libcurl: 8.12.0-DEV (sys:0.4.79+curl-8.12.0 vendored ssl:OpenSSL/1.1.1w)
os: Gentoo Linux 2.17.0 [64-bit]
```

### LINKED ISSUE #15413: cargo 1.86 package (and publish) fail when workspace crate contains symlink into git submodule
I've attached a minimal reproduction [cargo-bug-test.zip](https://github.com/user-attachments/files/19683981/cargo-bug-test.zip). Download it and run the following commands.

1. unzip cargo-bug-test.zip
2. cd cargo-bug-test
3. git status

Verify that git status reports a clean directory. If not, then extracting the symlink out of the zip didnt' work. This might happen on Windows. I've only tested on Linux.

The folder structure looks like this:

```
.git/...
Cargo.lock
Cargo.toml
crates/
crates/workspace-crate/
crates/workspace-crate/Cargo.lock
crates/workspace-crate/Cargo.toml
crates/workspace-crate/link-into-submodule -> ../../the-submodule/submodule-file
crates/workspace-crate/src/
crates/workspace-crate/src/lib.rs
the-submodule/
the-submodule/.git/...
the-submodule/submodule-file
```

Note that there is a git repository at the root and in `the-submodule`. This makes `the-submodule` a submodule, even though it isn't specified by a git URL. I did it this way to get a self contained example. Usually you would have a submodule with URL. The bug is the same.

`cargo +1.85 package -p workspace-crate` succeeds.

`cargo +1.86 package -p workspace-crate` fails with this error:

> error: attempt to get status of nonexistent file 'the-submodule/submodule-file'; class=Invalid (3); code=NotFound (-3)

The change behavior is caused by https://github.com/rust-lang/cargo/pull/14981 . The call `repo.status_file(&rel_path)` returns the error we see in the command line output. This could be a bug in Git2 or it could be that Cargo shouldn't pass submodule paths into Git2.

weihanglo:

If a there is a symlink into a git repository/submodule,
when checking its git status with the wrong outer repo,
we'll get an NotFound error,
as the object doesn't belong to the outer repository.
This kind of error blocked the entire `cargo package` operation.

This fix additionally discovers the nearest Git repository,
and then checks status with that,
assuming the repo is the parent of the source file of the symlink.
This is a best effort solution, so if the check fails we ignore.

If we don't want the complication,
we could drop the last commit, ignore the error, and forget about handling submodules

fixes #15384
fixes #15413

### DIFF
--- a/src/cargo/ops/cargo_package/vcs.rs
+++ b/src/cargo/ops/cargo_package/vcs.rs
@@ -274,9 +274,28 @@ fn dirty_files_outside_pkg_root(
                 dirty_files.insert(workdir.join(rel_path));
             }
             Err(e) => {
+                if e.code() == git2::ErrorCode::NotFound {
+                    // Object not found means this file might be inside a subrepo/submodule.
+                    // Let's check its status from that repo.
+                    let abs_path = workdir.join(&rel_path);
+                    if let Ok(repo) = git2::Repository::discover(&abs_path) {
+                        let is_dirty = if repo.workdir() == Some(workdir) {
+                            false
+                        } else if let Ok(path) =
+                            paths::strip_prefix_canonical(&abs_path, repo.workdir().unwrap())
+                        {
+                            repo.status_file(&path) != Ok(git2::Status::CURRENT)
+                        } else {
+                            false
+                        };
+                        if is_dirty {
+                            dirty_files.insert(abs_path);
+                        }
+                    }
+                }
+
                 // Dirtiness check for symlinks is mostly informational.
-                // And changes in submodule would fail git-status as well (see #15384).
-                // To avoid adding complicated logic to handle that,
+                // To avoid adding more complicated logic,
                 // for now we ignore the status check failure.
                 debug!(
                     "failed to get status from file `{}` in git repo at `{}`: {e}",
--- a/src/cargo/ops/cargo_package/vcs.rs
+++ b/src/cargo/ops/cargo_package/vcs.rs
@@ -268,8 +268,22 @@ fn dirty_files_outside_pkg_root(
         // Handle files outside package root but under git workdir,
         .filter_map(|p| paths::strip_prefix_canonical(p, workdir).ok())
     {
-        if repo.status_file(&rel_path)? != git2::Status::CURRENT {
-            dirty_files.insert(workdir.join(rel_path));
+        match repo.status_file(&rel_path) {
+            Ok(git2::Status::CURRENT) => {}
+            Ok(_) => {
+                dirty_files.insert(workdir.join(rel_path));
+            }
+            Err(e) => {
+                // Dirtiness check for symlinks is mostly informational.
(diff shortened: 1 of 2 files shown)


tests/testsuite/package.rs
```diff
@@ -1425,6 +1425,59 @@ edition = "2021"
     );
 }
 
+#[cargo_test]
+fn dirty_file_outside_pkg_root_inside_submodule() {
+    if !symlink_supported() {
```
epage, line 1430: Is there a reason this test was added in this commit?

epage:
Thanks!  Looks like its mostly nits, so let me know how much you see worth changing.

src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -268,8 +268,43 @@ fn dirty_files_outside_pkg_root(
...
+                                paths::strip_prefix_canonical(&abs_path, repo.workdir().unwrap())
+                            else {
+                                return false;
+                            };
+                            repo.status_file(&path) != Ok(git2::Status::CURRENT)
+                        })
+                        .unwrap_or_default();
+                    if is_dirty {
+                        dirty_files.insert(abs_path);
+                    }
```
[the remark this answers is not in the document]
weihanglo: Suggestion accepted!

tests/testsuite/package.rs
epage: Oops,. to make the suggestions, I was looking at all changes, not a commit, and forgot.

epage (approved):

weihanglo:
Got reverted by #15419 fwiw.

Issue #2526, opened 2016-03-29
Migrate to xz compression for crates.io crates


alexcrichton:
## Why?

**tl;dr;** unless you have a 52MB/s internet connection, xz is probably faster, but my math should be checked!

The statistics here are operating over the entirety of crates.io at this time, including all crates and all versions ever published. Currently we use `flate2` which is backed by miniz using the [best compression](https://github.com/rust-lang/cargo/blob/25e1301a4b72a31c0798aee4a8e8335d12d2a8ee/src/cargo/ops/cargo_package.rs#L119) for tarballs corresponding to [level 9](https://github.com/alexcrichton/flate2-rs/blob/522256b653087b615b5244fd601ab599026d2430/src/lib.rs#L110). The "zlib" numbers here are generated from compiling `flate2` against zlib instead of miniz. The xz numbers are generated with the [xz2 crate](https://github.com/alexcrichton/xz2-rs) also using compression level 9.

First up, let's take a look at what we're just storing on S3. This is just the size of all published crates.

| stat | val | % smaller |
|-----|-------|-------|
| miniz | 3776697673 | 0.0 |
| zlib | 3776147960 | 0.01 |
| xz | 2411082764 | 36.16 |

Next, let's multiply each version's size by how many times it's been downloaded. This in theory the number of bytes that have been transferred out of S3

| stat | val | % smaller |
|-----|-------|-------|
| miniz | 3502228200434 | 0.0 |
| zlib | 3501544526571 | 0.02 |
| xz | 2373770137784 | 32.22 |

Next up is how long it took (in nanoseconds) in total to decompress all crates on my local computer.

| stat | val | ns per byte | % slower |
|-----|-------|-------|------------|
| miniz  | 118891793644 | 31.484 | 0.0 |
| zlib | 118952441288 | 31.501 | 0.05 |
| xz | 144860343353 | 60.081 | 90.83 |

Ok, so the [claims of xz](http://tukaani.org/xz/) are correct in that it's about 30% smaller than gzip, but the decompression time is much larger! If we assume that these numbers are true in the average, however, let's do some math to figure out how fast your bandwidth needs to be to break even.

First up we've got:

```
time = bytes / BW + bytes * time_per_byte
```

So if we assume that xz crates are on average 36.16% smaller and use the timings we found above for decompressing, we have:

```
bytes / BW + bytes * 31.484 = bytes * (1 - .3616) / BW + bytes * (1 - .3616) * 60.081
[...]

PR #15417, opened 2025-04-10
use `zlib-rs` for gzip compression in rust code

folkertdev:

This PR uses `zlib-rs` via the `flate2` crate. It is used for (de)compressing gzip files (e.g. in `cargo package`). 

Using zlib-rs is significantly faster, and produces output of roughly similar size. For the `windows-bindgen` crate 
(a very large crate), the speedup is over 60%, or about 2 seconds.

```
> time cargo package -p windows-bindgen --no-verify
   Packaging windows-bindgen v0.61.0 (/home/folkertdev/rust/windows-rs/crates/libs/bindgen)
    Updating crates.io index
    Packaged 76 files, 31.2MiB (8.2MiB compressed)

________________________________________________________
Executed in    3.30 secs    fish           external
   usr time    3.19 secs  424.00 micros    3.19 secs
   sys time    0.05 secs   61.00 micros    0.05 secs

> time ~/rust/cargo/target/release/cargo package -p windows-bindgen --no-verify
   Packaging windows-bindgen v0.61.0 (/home/folkertdev/rust/windows-rs/crates/libs/bindgen)
    Updating crates.io index
    Packaged 76 files, 31.2MiB (8.3MiB compressed)

________________________________________________________
Executed in    1.25 secs    fish           external
   usr time    1.15 secs    0.00 micros    1.15 secs
   sys time    0.04 secs  589.00 micros    0.04 secs
```

Generally CI/the test suite should handle correctness.

Something to look out for is zlib-rs producing larger binaries than before. So far we're seeing output sizes that are roughly the same (sometimes a bit better, sometimes a bit worse) as the status quo.

We've not specifically looked at decompression yet, mostly because we could not come up with a good command to benchmark. In general zlib-rs is much faster than stock zlib there too.

### Additional information

For the time being, `cargo` still depends on stock zlib via e.g. `curl` and `git`. As far as I know it is the intention to eventually move away from these C dependencies, at which point the dependency on stock zlib also disappears.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1140,12 +1140,12 @@ dependencies = [
 
 [[package]]
 name = "flate2"
-version = "1.1.0"
+version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "11faaf5a5236997af9848be0bef4db95824b1d534ebc64d0f0c6cf3e67bd38dc"
+checksum = "7ced92e76e966ca2fd84c8f7aa01a4aea65b0eb6648d72f7c8f3e2764a67fece"
 dependencies = [
  "crc32fast",
- "libz-sys",
+ "libz-rs-sys",
  "miniz_oxide",
 ]
 
@@ -2598,6 +2598,15 @@ dependencies = [
  "openssl-sys",
  "pkg-config",
  "vcpkg",
+]
+
+[[package]]
+name = "libz-rs-sys"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6489ca9bd760fe9642d7644e827b0c9add07df89857b0416ee15c1cc1a3b8c5a"
+dependencies = [
+ "zlib-rs",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


joshtriplett:
Thanks, @folkertdev!

Nominating for next week's Cargo meeting.

epage:
We talked about how we want to approach these kind of changes in today's Cargo team meeting.

For gitoxide, its still under active development and each feature needs a lot of vetting before adoption, so we're taking an incremental, feature-at-at-ime, approach with an opt-in and opt-out windows to ensure as smooth of a path as possible.

For something like zlib-rs, it feels smaller and more mature.  We likely don't need a a multi-release transition window.  We do want it merged at the beginning of a release window to give as much runtime within the release as possible.  We also appreciate how easy it would be to back this out during beta if need be.  The main risk we see with zlib-rs is platform support.

For our `-sys` crates, we allow people building cargo to decide whether to use the vendored or system version of the package.  The question came up if we should still support those cases in addition to the pure-Rust implementation.  I don't feel that that model needs to be applied when adopting a pure-Rust variant of C code.  We could use C implementations of json, blake3, toml, base64, etc and the question didn't come up of whether that should be made configurable. 

As this is still early in the release Window, we can move forward with this.

epage (approved):

reneleonhardt:
Does this mean a final decision has been made or are alternatives like zstd still on the table? #2526

weihanglo:
This was just a performance improvement. Had no discussion around #2526 when it got merged.

PR #15418, opened 2025-04-11
Improved error message when build-dir template var is invalid

ranger-ross:

This PR improves the error message when an invalid template variable is used in the `build.build-dir` config.
I am using `closest_msg` to find a close match. If there are no close matches, we simply print the available template variables list. 

See #14125

r? @epage

### DIFF
--- a/tests/testsuite/build_dir.rs
+++ b/tests/testsuite/build_dir.rs
@@ -512,6 +512,8 @@ fn template_should_error_for_invalid_variables() {
         .with_stderr_data(str![[r#"
 [ERROR] unexpected variable `fake` in build.build-dir path `{fake}/build-dir`
 
+[HELP] available template variables are `{workspace-root}`, `{cargo-cache-home}`, `{workspace-path-hash}`
+
 "#]])
         .run();
 }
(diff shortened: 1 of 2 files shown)


tests/testsuite/build_dir.rs
```diff
@@ -512,6 +512,8 @@ fn template_should_error_for_invalid_variables() {
         .with_stderr_data(str![[r#"
 [ERROR] unexpected variable `fake` in build.build-dir path `{fake}/build-dir`
 
+[HELP] available template variables are [`{workspace-root}`, `{cargo-cache-home}`, `{workspace-path-hash}`]
```
epage, line 515: I don't think we commonly add square brackets like this.  We usually either do a comma separated list or a line-per-item list

ranger-ross: I removed them in https://github.com/rust-lang/cargo/compare/877e69d6232f8fd55953fbc6bcbaf528d4cb005c..fac2aafa1171bfb664cfa9cd4ea429623199f4aa

epage (approved):
Thanks!

PR #15429, opened 2025-04-13
test(rustfix): Use `snapbox` for snapshot testing

### LINKED ISSUE #13891: Migrate away from ad-hoc snapshot testing for rustfix test suite
### Problem

When [dealing with rustfix snapshot test fixes](https://github.com/rust-lang/cargo/pull/13880#pullrequestreview-2047809180), I found it a bit cumbersome to figure out the actual failure. You need to pass `RUST_LOG=parse_and_replace=info` to see the log, and `RUSTFIX_TEST_BLESS=test-name.rs` to update snapshots.

### Proposed Solution

In the main Cargo crate, we've already integrated [`snapbox`](https://crates.io/crates/snapbox) for [UI tests](https://doc.crates.io/contrib/tests/writing.html#ui-tests). We could migrate rustfix test suite to that, and maybe share some common infra, like nightly channel detection from `cargo-test-macro`.

### Notes

ehuss has a proposal that we don't run rustc if JSON snapshot exists: <https://github.com/rust-lang/cargo/pull/13890#issuecomment-2102747898>.
This could be a good enhancement, and I think we can leave it to follow-ups.

Pyr0de:

- separates each test into different test cases
- `snapbox` is used to test the snapshots
  - difference in `.json` file alone should never cause a test to fail
  - `.json` files updated only if expected fix != actual fix && `SNAPSHOTS=overwrite`
- when `.json` files are updated, the json is pretty printed
- replaced environment variables `RUSTFIX_TEST_*`  for overwriting test snapshots with `SNAPSHOTS=overwrite`
  - ❗ The `RUSTFIX_TEST_RECORD_FIXED_RUST` feature is removed (generate a `*.fixed.rs` on demand`). We can add it back whenever needed.

Fixes #13891 

Run tests with:
```sh
cargo test -p rustfix
```
All the test should run as different test cases
nightly tests run only when using nightly version of rustc is used


crates/rustfix/Cargo.toml
```diff
@@ -29,6 +29,9 @@ proptest.workspace = true
 similar.workspace = true
 tempfile.workspace = true
 tracing-subscriber.workspace = true
+cargo-test-support.workspace = true
+cargo-test-macro.workspace = true
```
weihanglo, line 33: * I would suggest not depending on `cargo-test-*`. It requires too new rustc toolchain version. Not using them might imply some more duplicate code like `static mut VERSION` and `fn version()` but that should be fine I guess.
* You might need to bump `rustfix` to 0.9.1 to get version bump check passed

crates/rustfix/tests/parse_and_replace.rs
```diff
@@ -6,65 +6,28 @@
 //! rustfix and applies the recommended suggestions to the `.rs` file. It then
 //! compares the result with the corresponding `.fixed.rs` file. If they don't
 //! match, then the test fails.
-//!
```
weihanglo, line 9: Can we add a doc comment here for how to use the `SNAPSHOTS` env var? Something like 

```suggestion
//! To override snapshots, run `SNAPSHOTS=overwrite cargo test`.
//! See [`snapbox::assert::Action`] for more different actions.
```

weihanglo (commented):
Thanks!

crates/rustfix/tests/parse_and_replace.rs
```diff
@@ -6,65 +6,28 @@
...
-//! There are several debugging environment variables for this test that you can set:
-//!
-//! - `RUST_LOG=parse_and_replace=debug`: Print debug information.
-//! - `RUSTFIX_TEST_BLESS=test-name.rs`: When given the name of a test, this
-//!   will overwrite the `.json` and `.fixed.rs` files with the expected
-//!   values. This can be used when adding a new test.
-//! - `RUSTFIX_TEST_RECORD_JSON=1`:  Records the JSON output to
-//!   `*.recorded.json` files. You can then move that to `.json` or whatever
-//!   you need.
-//! - `RUSTFIX_TEST_RECORD_FIXED_RUST=1`: Records the fixed result to
```
weihanglo, line 19: We seem to remove this feature in this patch. I am fine living without it, but maybe we should call it out in the PR description?

crates/rustfix/Cargo.toml
Pyr0de: Then a test that is supposed to run only in nightly would show up as passed when it is just being skipped, right? instead of it showing up as ignored

weihanglo: Nice catch! That's a bit unfortunate, but should acceptable ATM.

Pyr0de:
The test `multiple-solutions.nightly.rs` can be changed back to stable since rustc 1.80 is stable
https://github.com/rust-lang/rust/pull/123344

weihanglo (approved):
Thanks!

PR #15434, opened 2025-04-17
Fix formatting of CliUnstable parsing

ehuss:
Because the last string in this match statement was over 100 characters, it prevented rustfmt from formatting the entire thing. By wrapping it, that allows rustfmt to do its job.


weihanglo (approved):
I didn't know that's a problem 😅

PR #15438, opened 2025-04-18
feat(add): suggest similarly named features

### LINKED ISSUE #15436: Feature Request: add suggestion for similarly named features with `cargo add`
### Problem

When adding a crate with certain features and misspelling one of them, `cargo` doesn't suggest features with similar names. This may be confusing if a crate has lots of features. For example:
```sh
$ cargo add axum_extra --features typed-headers  # The actual feature is `typed-header`, see the last row of disabled features
    Updating crates.io index
warning: translating `axum_extra` to `axum-extra`
      Adding axum-extra v0.10.1 to dependencies
error: unrecognized feature for crate axum-extra: typed-headers
disabled features:
    __private_docs, async-read-body, async-stream, attachment, cookie
    cookie-key-expansion, cookie-private, cookie-signed, erased-json
    error-response, file-stream, form, json-deserializer, json-lines, multipart
    protobuf, query, scheme, typed-header, typed-routing
enabled features:
    tracing
```

### Proposed Solution

Instead of just listing all feature flags, `cargo` may additionally search for similarly named features and suggest them, so that it looks like the following:
```sh
$ cargo add axum_extra --features typed-headers
    Updating crates.io index
warning: translating `axum_extra` to `axum-extra`
      Adding axum-extra v0.10.1 to dependencies
error: unrecognized feature for crate axum-extra: typed-headers
help: a feature with a similar name exists: `typed-header`
disabled features:
    __private_docs, async-read-body, async-stream, attachment, cookie
    cookie-key-expansion, cookie-private, cookie-signed, erased-json
    error-response, file-stream, form, json-deserializer, json-lines, multipart
    protobuf, query, scheme, typed-header, typed-routing
enabled features:
    tracing
```

The help output here is worded the same way the compiler does

### Notes

_No response_

Jujumba:
Fixes #15436  

There are 3 tests for each test case:
- there are no feature suggestions
- there's only one feature suggestion (most common)
- there are several feature suggestions


epage:
> There are 3 tests for each test case:

Would you be willing to restructure this to match the recommendation at https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request to add tests in a commit *before* your change with them *passing* and then your commit with your change updates the tests so they still pass so that the diff between the commits highlights the change in behavior?

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -168,6 +169,33 @@ pub fn add(workspace: &Workspace<'_>, options: &AddOptions<'_>) -> CargoResult<(
                 write!(message, "no features available for crate {}", dep.name)?;
             } else {
                 if !deactivated.is_empty() {
+                    let mut suggested_features = Vec::new();
+                    for unknown_feature in &unknown_features {
+                        if let Some(suggested_feature) =
+                            edit_distance::closest(unknown_feature, deactivated.iter(), |dep| *dep)
```
epage, line 175: Why isn't this using `closest_msg`?

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -168,6 +169,33 @@ pub fn add(workspace: &Workspace<'_>, options: &AddOptions<'_>) -> CargoResult<(
...
+                                .coalesce(|x, y| if x.len() + y.len() < 78 {
+                                    Ok(format!("{x}, {y}"))
+                                } else {
+                                    Err((x, y))
+                                })
+                                .into_iter()
+                                .format("\n    ")
+                        )?;
+                    }
                     if deactivated.len() <= MAX_FEATURE_PRINTS {
```
epage, line 199: Should we still display all of the features if there is a closest match for every unknown feature?

src/cargo/ops/cargo_add/mod.rs
```diff
@@ -168,6 +169,33 @@ pub fn add(workspace: &Workspace<'_>, options: &AddOptions<'_>) -> CargoResult<(
                 write!(message, "no features available for crate {}", dep.name)?;
             } else {
                 if !deactivated.is_empty() {
+                    let mut suggested_features = Vec::new();
```
epage, line 172: Should we check even among activated features?

Jujumba:
> > There are 3 tests for each test case:
> 
> 
> 
> Would you be willing to restructure this to match the recommendation at https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request to add tests in a commit *before* your change with them *passing* and then your commit with your change updates the tests so they still pass so that the diff between the commits highlights the change in behavior?

Sure! To make tests pass before I introduce my changes, I have to leave them blank, right?

epage:
> Sure! To make tests pass before I introduce my changes, I have to leave them blank, right?

In this case, they should still be doing adds; just there will be no suggestions given.  If there are questions, the link I gave has links to example PRs.

src/cargo/ops/cargo_add/mod.rs
Jujumba: Perhaps, but I don't see value in it: these features are already activated 🤷‍♂️

src/cargo/ops/cargo_add/mod.rs
Jujumba: I would say yes, although this might be pretty lengthy

Jujumba:
> > Sure! To make tests pass before I introduce my changes, I have to leave them blank, right?
> 
> In this case, they should still be doing adds; just there will be no suggestions given. If there are questions, the link I gave has links to example PRs.

Alright, done

src/cargo/ops/cargo_add/mod.rs
epage: Looking around at how we do this elsewhere
- Most places do not list all options, even without a suggestion
- One place always lists out how to see all options, independent of whether there was a suggestion
- 2 places will conditionally show all options if a suggestion is not present

So there is only a small precedence but it tends to be towards not listing everything if there is a suggestion.

Thinking particularly for this case, we most likely want to focus the users attention on the suggestion.  Also, since there was a concern over the number of suggestions, what there can be a lot more of is number of features.  While we use columns, it can be quite large.

I lean towards only showing the suggestion.

src/cargo/ops/cargo_add/mod.rs
epage: The user might not realize its activated and they may be taking a feature that is indirectly activated and explicitly activating it.

Jujumba:
Ohh, just noticed that `edit_distance::closest_msg` appends two `'\n'` at the begging of the string, messing up the output

Jujumba:
> Ohh, just noticed that `edit_distance::closest_msg` appends two `'\n'` at the begging of the string, messing up the output

I think we have to rollback to `edit_distance::closes` therefore (?)

@epage

src/cargo/ops/cargo_add/mod.rs
Jujumba: > The user might not realize its activated and they may be taking a feature that is indirectly activated and explicitly activating it.

Sounds reasonable. So, do I have to check in activated features as well?

epage: Seems like we should.

Jujumba:
Also, how should I approach fixing other tests? Some of them fail because output of `cargo add` changed

weihanglo:
https://github.com/rust-lang/cargo/actions/runs/14540352012/job/40796970702#step:13:4593

Set the environment variable `SNAPSHOTS=overwrite`. See https://doc.crates.io/contrib/tests/writing.html#updating-snapshots

src/cargo/ops/cargo_add/mod.rs
Jujumba: > Looking around at how we do this elsewhere
> 
>     * Most places do not list all options, even without a suggestion
> 
>     * One place always lists out how to see all options, independent of whether there was a suggestion
> 
>     * 2 places will conditionally show all options if a suggestion is not present
> 
> 
> So there is only a small precedence but it tends to be towards not listing everything if there is a suggestion.
> 
> Thinking particularly for this case, we most likely want to focus the users attention on the suggestion. Also, since there was a concern over the number of suggestions, what there can be a lot more of is number of features. While we use columns, it can be quite large.
> 
> I lean towards only showing the suggestion.

So, how should the output look like?

tests/testsuite/cargo_add/feature_suggestion_multiple/stderr.term.svg
```diff
@@ -25,11 +25,23 @@
...
-    <tspan x="10px" y="100px"><tspan>    bar, foo</tspan>
+    <tspan x="10px" y="100px">
 </tspan>
-    <tspan x="10px" y="118px">
+    <tspan x="10px" y="118px"><tspan>help: a feature with a similar name exists: `bar`</tspan>
+</tspan>
+    <tspan x="10px" y="136px">
+</tspan>
+    <tspan x="10px" y="154px">
+</tspan>
```
epage, line 37: We're showing two blank lines.

This may be an artifact of how `closest_msg` expects error message newlines to be handled.  Could you do a refactor before this commit so that this commit only shows one visible line?

src/cargo/ops/cargo_add/mod.rs
Jujumba: > I'm not too sure what you are looking for with this question.  As I said, I lean towards only showing the suggestion. 

Sorry for the confusion, I mean: how should the suggestion look like? Just say that there `N` similarly spelled features, or perhaps something else?

Also, `closest_msg` appends two line breaks into the middle of the message, so maybe I should use different function?

epage: > Sorry for the confusion, I mean: how should the suggestion look like? Just say that there N similarly spelled features, or perhaps something else?

Unless we have a good reason, we should be use `closest_msg`.  If we do have a good reason, we should mirror it.  So that means we don't say there are "N similarly spelled features". 

> Also, closest_msg appends two line breaks into the middle of the message, so maybe I should use different function?

I commented on that at 
https://github.com/rust-lang/cargo/pull/15438#discussion_r2063736858 .  Other messages don't show two blank lines.

tests/testsuite/cargo_add/feature_suggestion_multiple/stderr.term.svg
Jujumba: Sorry, I'm not quite sure: do you mean removing blank lines or just leaving only one blank line per suggestion?

Sorry if I asking too much questions 😅

epage: There shouldn't be two blank lines between items, only one.

This is likely an outcome of other error messages not being expected to have a trailing newline, so `closest_msg` adds it and then adds a blank line.  But since we have a trailing newline, it makes two blank lines.

We likely need to just these messages to not have a trailing newline.

epage:
I know it can sometimes be a pain to clean up commits but could you
- Pull out the change in newlines to be before any future work was done
- Squash the feature work

That'll be a big help for reviewing

Jujumba:
> I know it can sometimes be a pain to clean up commits but could you
> 
> - Pull out the change in newlines to be before any future work was done
> 
> - Squash the feature work
> 
> 
> 
> That'll be a big help for reviewing

Sure, I'll ping you once done

src/cargo/ops/cargo_add/mod.rs
Jujumba: > To double check, earlier in this thread, I asked about not showing deactivated or activated features. Unless I'm misunderstanding something, it looks like that isn't done yet.

Looks like I missed it. (De)activated features shouldn't be printed only if there are feature suggestions, right?

Jujumba:
Hopefully the commit history is acceptable now

@rustbot ready

epage:
> Hopefully the commit history is acceptable now

Looks like you got rid of the leading test commits which make it easier to see how the behavior changed with each commit.

tests/testsuite/cargo_add/features_unknown_no_features/stderr.term.svg
```diff
@@ -25,11 +25,9 @@
 </tspan>
     <tspan x="10px" y="64px"><tspan class="fg-red bold">error</tspan><tspan class="bold">:</tspan><tspan> unrecognized feature for crate my-package: noze</tspan>
 </tspan>
-    <tspan x="10px" y="82px">
+    <tspan x="10px" y="82px"><tspan>no features available for crate my-package</tspan>
 </tspan>
-    <tspan x="10px" y="100px"><tspan>no features available for crate my-package</tspan>
-</tspan>
-    <tspan x="10px" y="118px">
```
epage: Why did the previous commit add a blank line for this one to remove it?

tests/testsuite/cargo_add/feature_suggestion_none/stderr.term.svg
```diff
@@ -0,0 +1,36 @@
...
+
+  <text xml:space="preserve" class="container fg">
+    <tspan x="10px" y="28px"><tspan class="fg-green bold">    Updating</tspan><tspan> `dummy-registry` index</tspan>
+</tspan>
+    <tspan x="10px" y="46px"><tspan class="fg-green bold">      Adding</tspan><tspan> my-package v0.1.0 to dependencies</tspan>
+</tspan>
+    <tspan x="10px" y="64px"><tspan class="fg-red bold">error</tspan><tspan class="bold">:</tspan><tspan> unrecognized feature for crate my-package: none_existent</tspan>
+</tspan>
+    <tspan x="10px" y="82px"><tspan>disabled features:</tspan>
+</tspan>
```
epage, line 29: Didn't the previous commit add a blank line between these?

Jujumba:
> > Hopefully the commit history is acceptable now
> 
> Looks like you got rid of the leading test commits which make it easier to see how the behavior changed with each commit.

Sorry, I thought that I needed to squash commits regarding the actual feature into. Do you want me to undo all of this?..

epage:
Since it looks like there are other changes to be made, it would be nice.

PR #15451, opened 2025-04-23
[WIP] feat(manifest): add a `force-frame-pointers` profile option

### LINKED ISSUE #15333: Allow profiles to set -Cforce-frame-pointers
### Problem

When profiling Rust code on Linux using `perf`, I can either use `perf record` with `--call-graph=dwarf` or I can enable frame pointers during the build by passing `-Cforce-frame-pointers` to rustc. With frame pointers, a perf run is substantially faster to process. My understanding is that this is because stack traces can be captured in the kernel with eBPF, which can follow the frame pointers, whereas with debug info it's all done by calling `addr2line` after the fact. 

For a profiling run I just did, it took more than 8 minutes to process the perf data when using debug info and less than a second when using frame pointers.

I'd like to be able to turn on frame pointers for a particular profile in my `Cargo.toml`. Currently, I instead turn it on globally in my `~/.cargo/config.toml`.

### Proposed Solution

Add `force-frame-pointers` as an option in cargo profiles. e.g.

```toml
[profile.profiling]
inherits = "release"
force-frame-pointers = true
```

### Notes

If it's agreed that this is a reasonable feature to add, I'm happy to implement it.

brvtalcake:

This PR resolves rust-lang/cargo#15333, and a `force-frame-pointers` option for profiles. It also fixes tests related to profiles, and adds some basic tests for the new option itself.

Run the tests in `tests/testsuite/profiles.rs` :
- `force_frame_pointers_on_works`
- `force_frame_pointers_off_works`
- `force_frame_pointers_unspecified_works`

### Additional information

See, however, the discussion in the linked issue, as there might be some changes needed


crates/cargo-util-schemas/src/manifest/mod.rs
epage, line 1: Some of the solution details are still being discussed in #15451

tests/testsuite/profiles.rs
```diff
@@ -704,6 +704,79 @@ fn do_not_strip_debuginfo_with_requested_debug() {
         .run();
 }
 
+#[cargo_test]
+fn force_frame_pointers_on_works() {
```
epage, line 708: Would you be up for putting the tests in a previous commit without `force-frame-pointers`, having them pass, showing the existing behavior?  Then in this commit, the diff highlights the new feature and what it changes

brvtalcake: No problem, I can do that

brvtalcake: EDIT : I suppose I should first wait for a conclusion to the conversation in the related issue ?

ranger-ross:
I am going to close this PR as its stale and there have been other developments in the issue.  including merging https://github.com/rust-lang/cargo/pull/16742 which appears to do the same thing as this PR

I believe the latest update is: https://github.com/rust-lang/cargo/issues/15333#issuecomment-4550839960

PR #15452, opened 2025-04-24
Format Init/New Directory for Cargo.toml and Package Names

mewfinity06:
# Format Init/New Directory for Cargo.toml and Package Names

## Motivation

This started when I was porting (nob.h)[https://github.com/tsoding/nob.h] to Rust as a fun side project and I named my directory `nob.rs` before initialization (I like to make my directory first then go in and `cargo init` it). I got an error saying that `.` is an illegal char. Wasn't a fan, so I decided to add a function that auto fixes it and still provides the error as a warning (the error on the legality of the parent directory's name)

(This my first PR, so I am sorry if my explanation is not exactly up to par. Thank you in advance for the review!)

## Testing

This passes all of the tests for `cargo_new` and all but one for `cargo_init` (the one for malformed package name)

## TODO

- Make this an optional command. Perhaps `--format-name` which would supersede `--name <NAME>`
- Add more cases, perhaps substituting certain chars for anything other than `_`

## Issues

The only *potential* problem with this merge is it would establish how crates must be named. It would not force this format, but push future crates to use the underscore syntax

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3919,10 +3919,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "test_1__"
-version = "0.1.0"
-
-[[package]]
 name = "thiserror"
 version = "1.0.69"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3919,6 +3919,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "test_1__"
+version = "0.1.0"
+
+[[package]]
 name = "thiserror"
 version = "1.0.69"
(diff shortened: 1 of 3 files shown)


epage:
I'm sympathetic to the idea of "just make things work" especially for a command like this.  We do generally [ask](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo) that people create issues first.  This way we get agreement on the problem and the general solution in a place where people will tend to look for it and the PR focuses on just the implementation.

On the surface, this seems small enough for a go/no-go on just the PR.  However, I do think some due diligence should be done to look into why we chose to error especially since we have test cases specifically for this so it was consciously thought about.

PR #15454, opened 2025-04-25
fix: Suggest similar looking feature names when feature is missing

epage:

I recently depended on a package with `preserve-order` rather than `preserve_order` and the error message didn't help me with the problem so I figure I'd fix that.   I also found other improvements along the way

- Suggest an alternative feature when a feature includes a missing feature
- Suggest an alternative feature when a dependency includes a missing feature
- Lower case error messages
- Re-frame prescriptive information as help
- Change plural "features" error messages to singular "feature" as they can only ever have one (knowing an the `MissingFeature` string only has one feature in it was important for doing a `closest` match on the feature).

### DIFF
--- a/tests/testsuite/features.rs
+++ b/tests/testsuite/features.rs
@@ -37,6 +37,38 @@ Caused by:
 }
 
 #[cargo_test]
+fn feature_activates_typoed_feature() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.0.1"
+                edition = "2015"
+                authors = []
+
+                [features]
+                bar = ["baz"]
+                jaz = []
+            "#,
+        )
+        .file("src/main.rs", "")
+        .build();
+
+    p.cargo("check")
+        .with_status(101)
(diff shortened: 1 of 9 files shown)


tests/testsuite/features.rs
```diff
@@ -280,6 +280,7 @@ fn dependency_activates_typoed_feature() {
 versions that meet the requirements `*` are: 0.0.1
 
 package `foo` depends on `bar` with feature `bar` but `bar` does not have that feature.
+ package `bar` does have feature `baz`
```
weihanglo, line 283: Is the leading space intentional?

epage: Other clauses have a context line like this with a single leading space.

Its weird.  It seems like it should be one of 0, 2, or 4 spaces but I didn't feel like following up what the intent was.

epage: Merging now to get the incremental improvement rather than blocking on figuring out what the intended indentation is.

epage: @Eh2406 when back from vacation, if you'd be willing to look at these errors and let us know what you think would work well, I'd appreciate it.

PR #15455, opened 2025-04-25
Update doctest xcompile flags

ehuss:
This updates the flags used for doctest xcompile to match the upstream changes in https://github.com/rust-lang/rust/pull/137096 which renamed and stabilized the flags.

This cannot be merged until after nightly is published tonight.

### DIFF
--- a/src/cargo/ops/cargo_test.rs
+++ b/src/cargo/ops/cargo_test.rs
@@ -238,12 +238,10 @@ fn run_doc_tests(
         }
 
         if doctest_xcompile {
-            p.arg("-Zunstable-options");
-            p.arg("--enable-per-target-ignores");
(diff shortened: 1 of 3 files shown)


epage (approved):
Feel free to merge when it is no longer blocked

weihanglo (approved):

weihanglo (commented):
> Stabilizes the `--enable-per-target-ignores` flag by removing it and making it unconditionally enabled. This makes it possible to disable a doctest on a per-target basis, which I think will be helpful for rolling out this feature.

Since per-target-ignore is enabled unconditionally, this test needs an adjustment.

https://github.com/rust-lang/cargo/blob/872b92e6719bed523b5563f2858c923763a4e638/tests/testsuite/test.rs?plain=1#L4740

PR #15457, opened 2025-04-26
fix grammar, and remove confusing example

tshepang:
Also, lots of commands need to contact the registry, so seeing an example gives me doubt.

### DIFF
--- a/src/doc/src/reference/source-replacement.md
+++ b/src/doc/src/reference/source-replacement.md
@@ -27,10 +27,13 @@ dependencies], and private registry support is described in [the Registries
 chapter][registries].
 
 When using source replacement, running commands that need to
-contact the registry requires passing the `--registry` option. This helps avoid
+contact the registry directly[^1] requires passing the `--registry` option. This helps avoid
 any ambiguity about which registry to contact, and will use the authentication
 token for the specified registry.
 
+[^1]: Examples of such commands are in [Publishing Commands].
+
+[Publishing Commands]: ../commands/publishing-commands.md
 [overriding dependencies]: overriding-dependencies.md
 [registries]: registries.md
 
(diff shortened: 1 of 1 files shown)


epage:
> remove confusing example
>
> Also, lots of commands need to contact the registry, seeing an example gives me doubt

For commands that contact the registry, we have
- `cargo login`
- `cargo logout`
- `cargo install`
- `cargo yank`
- `cargo publish`
- `cargo package`
- `cargo search`

Could you help me understand how giving the example of `cargo publish` gave you doubt / is confusing?

tshepang:
Because there is a lot of them, listing just one made me doubt that fact. It made me think it's `publish` and just 1 or 2 more.

epage:
Maybe my perspective is biased but that is a small subset of our overall commands and having a concrete example of "this is what touching the registry looks like that might need this flag" I feel like helps take something abstract and make it concrete.

For example, when I saw this PR description of "lots of commands", I thought you were also referring to commands like `cargo check` and I had to go double check because I didn't think that was the case.

tshepang:
> For example, when I saw this PR description of "lots of commands", I thought you were also referring to commands like `cargo check` and I had to go double check because I didn't think that was the case.

is it not `cargo check` also touches registry when lockfile is stale enough, as well as others like `cargo build`, or even `cargo test`

epage:
> is it not cargo check also touches registry when lockfile is stale enough, as well as others like cargo build, or even cargo test

They do but they use the [dependency `registry` field](https://doc.rust-lang.org/nightly/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-other-registries) rather than a `--registry` command line parameter.

Check `cargo check -h`.  It does not list a `--registry` option.

Some commands beyond what I listed have a `--regsitry` option but its for purposes other than listed here (e.g. `cargo add --registry` is for setting the dependency `registry` field, not for directly communicating with registries).

tshepang:
use https://github.com/rust-lang/cargo/pull/15457#issuecomment-2835574716 to help clarify things

PR #15461, opened 2025-04-28
Trivial tweaks to 'target_short_hash'

brson:

The `target_short_hash` function is a fallback used by the `pkg_dir` function for units that should not generate `-C extra-filename`.

This makes the function private since it would be easy to misuse and fixes an outdated method name in the doc comment.

Trivial.


weihanglo (approved):
Makes sense. Thanks!

Issue #13530, opened 2024-03-04
When crates.io gives 429, cargo should back off and retry later


ijackson:
### Problem

Our workspace contains 46 cargo packages.  (Because cargo insists that each crate must be a separate package, and we want to split up crates for code sanity and compilation time reasons.)

This means that in our recent release, [our on-duty release technician hit the rate limit](https://gitlab.torproject.org/tpo/core/arti/-/issues/1310).  This aborted publication of the workspace, requiring manual retries and wrangling.

### Steps

Have a workspace with more than 30 (the current burst rate limit) crates.  Try to publish it by publising each crate, in topo order, with `cargo publish` (using some automated tool).

### Possible Solution(s)

cargo should handle a 429 response by backing off and retrying, using an exponential backoff algorithm.

In #1643 the crates.io team report already having raised the rate limit.  In the error message from crates.io they suggest emailing `help@` to ask for a rate limit increase.  Such a workflow is IMO undesirable, especially as Rust gets more adoption.

### Notes

I don't think increasing the rate limit (globally, or on request) is the right fix.  If 429 is a hard error there is a tension between preventing misuse, and not breaking large projects' releases.  But this tension can be abolished by handling 429 gracefully.

#13397 would probably have assisted the recovery from this situation (and also the local disk space problem our releasae technician also ran into).

See also: #3229 (requesting docs) #6714 (requesting better error message display).

### Version

```text
> cargo version --verbose                                                                                                                                                                                                                                                                             16:57:11
cargo 1.76.0 (c84b36747 2024-01-18)
release: 1.76.0
commit-hash: c84b367471a2db61d2c2c6aab605b14130b8a31b
commit-date: 2024-01-18
host: x86_64-unknown-linux-gnu
libgit2: 1.7.1 (sys:0.18.1 vendored)
libcurl: 8.5.0-DEV (sys:0.4.70+curl-8.5.0 vendored ssl:OpenSSL/1.1.1w)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Arch Linux Rolling Release [64-bit]
```

epage:
People will be more likely to hit this with #1169 (since we'd likely move forward on that without the batch publish on crates.io's side)

`cargo release` tried to detect rate limitation situations and warn users about them so they can break down the publish into smaller steps.

As for strategies to deal with this, I'd want input from crates.io to know what fits with their intent of the rate limit.

Ideas brought up
- Back off and retry
[...]

PR #15463, opened 2025-04-29
feat(network): use Retry-After header for HTTP 429 responses

arlosi:

Cargo registries that return HTTP 429 when the service is overloaded expect the client to retry the request automatically after a delay. Cargo currently does not retry for HTTP 429.

### What changed?
* Adds HTTP 429 (too many requests) as a spurious HTTP error to enable retries.
* Parse the [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After) HTTP header to determine how long to wait before a retry.

In this implementation, the maximum delay is limited to Cargo's existing limit of 10 seconds. We could consider increasing that limit for this case, since the server is explicitly requesting the delay.

### DIFF
--- a/src/cargo/util/network/retry.rs
+++ b/src/cargo/util/network/retry.rs
@@ -104,8 +104,8 @@ impl<'a> Retry<'a> {
     pub fn r#try<T>(&mut self, f: impl FnOnce() -> CargoResult<T>) -> RetryResult<T> {
         match f() {
             Err(ref e) if maybe_spurious(e) && self.retries < self.max_retries => {
-                let err_msg = e
-                    .downcast_ref::<HttpNotSuccessful>()
+                let err = e.downcast_ref::<HttpNotSuccessful>();
+                let err_msg = err
                     .map(|http_err| http_err.display_short())
                     .unwrap_or_else(|| e.root_cause().to_string());
                 let left_retries = self.max_retries - self.retries;
@@ -118,7 +118,12 @@ impl<'a> Retry<'a> {
                     return RetryResult::Err(e);
                 }
                 self.retries += 1;
-                RetryResult::Retry(self.next_sleep_ms())
+                let sleep = err
+                    .and_then(|v| Self::parse_retry_after(v, &jiff::Timestamp::now()))
+                    // Limit the Retry-After to a maximum value to avoid waiting too long.
+                    .map(|retry_after| retry_after.min(MAX_RETRY_SLEEP_MS))
+                    .unwrap_or_else(|| self.next_sleep_ms());
+                RetryResult::Retry(sleep)
             }
             Err(e) => RetryResult::Err(e),
             Ok(r) => RetryResult::Success(r),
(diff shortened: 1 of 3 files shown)


src/cargo/util/network/retry.rs
```diff
@@ -47,6 +47,7 @@ use anyhow::Error;
 use rand::Rng;
 use std::cmp::min;
 use std::time::Duration;
+use time::OffsetDateTime;
```
weihanglo, line 50: I feel like we are on the way of dropping `time`? https://github.com/rust-lang/cargo/pull/15293

weihanglo (commented):
cc <https://github.com/rust-lang/cargo/issues/13530>, which I believe this partially resolves it.

arlosi:
> cc #13530, which I believe this partially resolves it.

This only applies for things that use `Retry`. IIRC the publishing logic doesn't have built-in retries at all.

ehuss:
Would it be possible to add a test using our `HttpServer` to ensure that this works going through curl? I imagine it would be something like adding a `custom_responders`?

src/cargo/util/network/retry.rs
ehuss: Maybe using jiff isn't the right call?

In my testing, it can take 2 to 10 seconds for jiff to initialize.

I suspect part of the problem is that jiff is configured to use the system timezone data, and it's probably loading and initializing a bunch of junk. Unfortunately because `gix-date` enables the default features, we can't easily disable that AFAIK.

Can we maybe get gix-date to not use the default? I don't fully understand what all the jiff features do, or when or if they are necessary.

src/cargo/util/network/retry.rs
```diff
@@ -317,3 +358,57 @@ fn curle_http2_stream_is_spurious() {
...
+                .at(0, 0, 0, 0)
+                .to_zoned(jiff::tz::TimeZone::UTC)
+                .unwrap(),
+        )
+        .unwrap()
+        .total(jiff::Unit::Millisecond)
+        .unwrap() as u64;
+    let actual = Retry::parse_retry_after(&headers).unwrap();
+    let diff = expected.abs_diff(actual.into());
+    assert!((diff < 1000), "{} != {} ({})", expected, actual, diff);
```
ehuss, line 391: Independent of the problem with jiff initialization, I would recommend moving the `expected` initialization to *after* the `actual` line. That should reduce the time variance (since computing `expected` is very efficient, whereas `parse_retry_after` is not).

arlosi: I'll change `parse_retry_after` to take in `now` as a parameter to simplify the testing and eliminate the variation.

ehuss: Correction, I think I had that backwards. It looks like `jiff::Zoned::now()` is what is taking all the time.

src/cargo/util/network/retry.rs
arlosi: It seems like `jiff` init is taking around that time in CI, which I agree is unacceptable. I'm not seeing the delay on Windows, but that's likely because jiff isn't using system timezone data there.

arlosi: I wonder if there's a way to do the parsing without loading the timezone data. The HTTP Date header is always GMT.

src/cargo/util/network/retry.rs
arlosi: Can you try again with latest push and see if it's fixed?

ehuss (approved):
Seems reasonable to me, thanks!

src/cargo/util/network/retry.rs
ehuss: I reported this upstream at https://github.com/BurntSushi/jiff/issues/366.

Issue #15010, opened 2025-01-03
Re-organize build-dir by package + hash, rather than artifact type


epage:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/issues/15010)*

### Call of Testing
Also include anyone is following this issue and missed it, we put out a call of testing of the new v2 `build-dir` layout in a blog post [here](https://blog.rust-lang.org/2026/03/13/call-for-testing-build-dir-layout-v2/)
See the blog post for details and testing instructions

---

Implementation:
- [x] #15848
- [x] #15947
- [x] #16351
- [x] #16807

Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#build-dir-new-layout

Resolves
- Shadowing binaries on Windows (#7919)
- Collisions between intermediate build artifacts (#8794, #16673)
- Inconsistent use of `-Cextra-filename` (#8332)
- Slow performance when intermediate artifacts accumulate (#16665)

Past efforts
- #6668

## Overview

- `.fingerprint`, `deps/`, `build-script`, and `build-script-execution/` move under `build/<pkgname>/<unit-id>`
  - `-L` params to `rustc` were updated
  - `cargo clean -p` was updated
  - `incremental/` is untouched
- `-Cextra-filename` is removed from all bins (#16351)

## Known issues

- As a side effect, we pass a lot more parameters to rustc, likely making `cargo -vv` more annoying, similar to #13941
- It also makes impossible to just copy paste commands into your shell if the limit is exceeded. See [bjorn3's comment](https://github.com/rust-lang/cargo/pull/15947#issuecomment-3393205232).
- This will break people relying on `build-dir` internals
  - See [#t-cargo > build-dir tool compatibility @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/build-dir.20tool.20compatibility/near/546267015)
- Increases the need for solutions for when people rely on `build-dir` internals
  - #13672
  - #5026
  - #13663
  - rust-lang/rfcs#3553 (more precise determination of what is built than `cargo metadata`)
- `cargo clean -p` only cleans up files for the current layout rather than supporting both

## Unresolved issues

Open questions
- [ ] Transition plan: while `build-dir` isn't stable, enough tools rely on the layout that we'd want to setup a transition plan so they can have time to test against the new layout and work to support both
- [x] What do we call the directory?  I said `build/` as its all encompassing
[...]

Issue #4282, opened 2017-07-13
More granular locking in cargo_rustc


alexcrichton:
Right now whenever a build happens we lock the entirety of the `target` directory for the whole build, but it may be possible for us to have a more granular locking strategy which allows multiple instances of Cargo to proceed in parallel instead of serializing them.

alexcrichton:
cc https://bugzilla.mozilla.org/show_bug.cgi?id=1380321

epage:
As part of #5931, we've talked about changing the target directory layout so that there is a directory per package.  This would give us more definitive units to lock.

epage:
In #5931, we discuss having a directory per intermediate artifact inside of a user-wide cache directory.  Each intermediate artifact would be lockable so there is exclusive access on the initial write and then multiple readers after that that prevent during builds.  Depending on packages out of this shared location is important so we don't waste time copying things out.

To keep things simple, ideally we model the target directory the same way.  So in a way, fixing this issue is a perfect subset of #5931.

We'd need a separate lock for the artifact directory though `cargo check` and `cargo clippy` shouldn't need that.

Any compilation that changes the [fingerprint but not the `-Cextra-filename`](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/compiler/fingerprint/index.html) would need to have exclusive access to the directory.

Therefore, the benefit of this feature would be limited to  `check` or`clippy` together or with a single full build command (build, run, test, doc) and at least one of
- Different enough `--features`
- Different enough `--package`s
- Different compilation modes (`check`, `rustdoc`, `build`)
  - `clippy` is not a separate mode but I think we do track it as separate somehow ("rustc"?)
- Different RUSTFLAGS if #14830 is merged
- Different profile

ehuss:
I think it would be good to also keep on eye on performance, since this could introduce more OS overhead. I'm also uncertain if we will run into lock limits, particularly on networked filesystems or filesystems like Docker, etc. It looks like the number of locks is configurable in Linux, but I don't immediately see any strict limits anywhere else.

redactedontop:
Any updates?

ConradIrwin:
One important aspect of this (for me at least) is prioritizing user requests over background requests (for example from LSPs). Building [Zed](https://github.com/zed-industries/zed) consumes all my CPU cores, and I'd like the rust-analyzer diagnostics to get out of my way when I run `cargo run`.

epage:
[...]

PR #15464, opened 2025-04-29
Added tracing spans for rustc invocations

ranger-ross:

While doing some investigation on the theoretical performance implications of #4282 (and #15010 by extension) I was profiling cargo with some experimental changes. (Still a work in progress)

But in the mean time, noticed that we do not have spans for rustc invocations. I think these would be useful when profiling `cargo build`. (`cargo build --timing` exists but is more geared towards debugging a slow building project, not cargo itself)

For reference below is an example before/after of a profile run of a dummy crate with a few random dependencies.

#### Before

image

#### After

image

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -67,7 +67,7 @@ use std::sync::Arc;
 
 use anyhow::{Context as _, Error};
 use lazycell::LazyCell;
-use tracing::{debug, trace};
+use tracing::{debug, instrument, trace};
 
 pub use self::build_config::{BuildConfig, CompileMode, MessageFormat, TimingOutput};
 pub use self::build_context::{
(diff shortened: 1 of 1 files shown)


src/cargo/core/compiler/mod.rs
```diff
@@ -402,6 +402,13 @@ fn rustc(
...
             }
+            let span = tracing::span!(
+                tracing::Level::TRACE,
+                "rustc",
+                package = name.as_str(),
+                process = rustc.to_string()
+            )
+            .entered();
 
             let result = exec
```
weihanglo, line 413: Alternatively, trace all execs?

ranger-ross: Sure, I think that also works.
I put it outside since `Executor` seems a bit generic, but looking at the function signatures it is very much designed to run rustc.

I went ahead and moved it inside of the `Executor` in addc570fabd0ca9139170338c10b51badf048f50 since I think its a bit nicer to use the `#[instrument]` macro :)

arlosi (approved):

src/cargo/core/compiler/mod.rs
```diff
@@ -140,10 +140,11 @@ pub trait Executor: Send + Sync + 'static {
 pub struct DefaultExecutor;
 
 impl Executor for DefaultExecutor {
+    #[instrument(name = "rustc", skip_all, fields(package = id.name().as_str(), process = cmd.to_string()))]
```
weihanglo, line 143: Just note that this is at INFO log level.

Our instruments are all like this, but I feel like we might want to downgrade to debug or trace. Not a blocker though.

weihanglo (approved):

src/cargo/core/compiler/mod.rs
weihanglo: Thanks!

PR #15477, opened 2025-05-01
Only hash host-independent host-dependency metadata when cross-compiling

### LINKED ISSUE #8140: Non-reproducible -C metadata=hash passed to rustc depending on the compiling OS

brson:

This is an attempt to improve reproducibility of wasm artifacts (and other cross-targets) when built on different host platforms.

re https://github.com/rust-lang/cargo/issues/8140#issuecomment-2285036273

The problem: when compiling wasm, the `-C metadata` hash contains metadata from host dependencies, particularly build scripts and proc macros. This metadata is different on Linux and Mac OS, etc, causing symbols in the wasm to differ.

This patch solves this by maintaining a separate `host_independent_hash` that contains only the most basic info about a unit: name, version, etc. When hashing a "target" unit, if a dependency is a host dependency, it only hashes this simple information, not the complete metadata.

In this way, any host-specific metadata is excluded, while still maintaining _some_ information about host dependencies. So if a proc_macro crate _version_ changes it will result in a different wasm, but if the proc_macro's _build profile_, etc. changes it will not.

This may resolve #8140, though I am not familiar with the entirety of that issue.

A way to think about this change: the exact configuration of build scripts and proc macros shouldn't change how they influence the symbols or filenames of wasm binaries; so it's ok to exclude that metadata from the wasm.

Whether this is true or not IDK: think about build scripts that may or may not build native libraries based on feature flags. That has a big influence on the build but after this patch won't be reflected in the wasm's symbol hashes. Ok or not?

This patch is very hard to unit test as it requires building on two different hosts and comparing.

I have done a manual test of building all the wasms in the https://github.com/stellar/soroban-examples repo, on both linux and mac, before and after this patch:

https://gist.github.com/brson/3fbfe7ae70c6ffa88f94b546d18bcb93

There are four files there: before-linux.txt, after-linux.txt, before-mac.txt, after-mac.txt. The relevant thing to notice is that both the "after" wasms have the same hashes across platforms, while that is not true for the "before" wasms.

The test was done on soroban-examples 75f1cc676bd80153a75725f80a4181dedf79a092

cc @leighmcculloch @graydon

### DIFF
--- a/src/cargo/core/compiler/build_runner/compilation_files.rs
+++ b/src/cargo/core/compiler/build_runner/compilation_files.rs
@@ -6,6 +6,7 @@ use std::hash::{Hash, Hasher};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
+use itertools::{Either, Itertools};
 use lazycell::LazyCell;
 use tracing::debug;
 
@@ -88,9 +89,15 @@ impl fmt::Debug for UnitHash {
 /// mangled symbols need to be stable between different builds with different
 /// settings. For example, profile-guided optimizations need to swap
 /// `RUSTFLAGS` between runs, but needs to keep the same symbol names.
+///
+/// [`Metadata::host_independent_hash`] is a simple hash that excludes any metadata
+/// that may differ across hosts. It is used when cross-compiling as the dependency
+/// hash for host crates like build scripts and proc macros, in order to improve
+/// reproducability across build platforms.
 #[derive(Copy, Clone, Debug)]
 pub struct Metadata {
     unit_id: UnitHash,
+    host_independent_hash: UnitHash,
     c_metadata: UnitHash,
     c_extra_filename: Option<UnitHash>,
 }
@@ -617,11 +624,19 @@ fn compute_metadata(
     metas: &mut HashMap<Unit, Metadata>,
 ) -> Metadata {
     let bcx = &build_runner.bcx;
-    let deps_metadata = build_runner
+    let (deps_metadata, deps_metadata_cross_host) = build_runner
         .unit_deps(unit)
         .iter()
-        .map(|dep| *metadata_of(&dep.unit, build_runner, metas))
-        .collect::<Vec<_>>();
+        .partition_map::<Vec<_>, Vec<_>, _, _, _>(|dep| {
+            let meta = *metadata_of(&dep.unit, build_runner, metas);
+            let is_cross_compile_host_dep =
+                matches!(unit.kind, CompileKind::Target(_)) && dep.unit.kind.is_host();
+            if !is_cross_compile_host_dep {
+                Either::Left(meta)
+            } else {
+                Either::Right(meta)
+            }
+        });
     let use_extra_filename = use_extra_filename(bcx, unit);
 
     let mut shared_hasher = StableHasher::new();
@@ -643,6 +658,19 @@ fn compute_metadata(
         .package_id()
         .stable_hash(ws_root)
         .hash(&mut shared_hasher);
+
+    // Mix in only the host-independent target-metadata of any host dependencies
+    // of this target-kind dependency. This ensures that cross-compile targets
+    // like wasm do not mix in host-specific metadata, and are more likely to be
+    // reproducible regardless of build platform.
+    let mut dep_host_independent_hashes = deps_metadata_cross_host
+        .iter()
+        .map(|m| m.host_independent_hash)
+        .collect::<Vec<_>>();
+    dep_host_independent_hashes.sort();
+    dep_host_independent_hashes.hash(&mut shared_hasher);
+
+    let host_independent_hasher = shared_hasher.clone();
 
     // Also mix in enabled features to our metadata. This'll ensure that
     // when changing feature sets each lib is separately cached.
@@ -751,6 +779,7 @@ fn compute_metadata(
         }
     }
 
+    let host_independent_hash = UnitHash(Hasher::finish(&host_independent_hasher));
     let c_metadata = UnitHash(Hasher::finish(&c_metadata_hasher));
     let c_extra_filename = UnitHash(Hasher::finish(&c_extra_filename_hasher));
     let unit_id = c_extra_filename;
@@ -759,6 +788,7 @@ fn compute_metadata(
 
     Metadata {
         unit_id,
+        host_independent_hash,
         c_metadata,
         c_extra_filename,
     }


epage:
Thanks for working on solving this!

As a heads up, I'm going to mark this as a Draft to remind participants that [we prefer to keep discussions on Problems and Solutions in one place](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo), like #8140, while this can serve as an example of a proposed Solution.  Once thats done, we can shift to discussing the implementation here.

src/cargo/core/compiler/build_runner/compilation_files.rs
```diff
@@ -88,9 +89,15 @@ impl fmt::Debug for UnitHash {
 /// mangled symbols need to be stable between different builds with different
 /// settings. For example, profile-guided optimizations need to swap
 /// `RUSTFLAGS` between runs, but needs to keep the same symbol names.
+///
+/// [`Metadata::host_independent_hash`] is a simple hash that excludes any metadata
```
epage, line 93: We'll need to evaluate how the information at https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/compiler/fingerprint/index.html#fingerprints-and-unithashs will need to be updated based on this

epage (commented):
Wanted to record this while I had it on my mind

leighmcculloch:
👏🏻 Independent of #8140 this improves reproducibility so it seems like a net benefit to me, but I'm also not familiar with the entirety of the problem to know if the solution will cut across anything more holistic that might be done.

> before-linux.txt, after-linux.txt, before-mac.txt, after-mac.txt

Is the intent that other hosts such as windows will also be consistent and are just not tested out of inconvenience, or are the fixes unique to linux and mac hosts?

brson:
> Is the intent that other hosts such as windows will also be consistent and are just not tested out of inconvenience, or are the fixes unique to linux and mac hosts?

I expect windows and other hosts to produce the same results as linux and mac. It was just more convenient to test linux/mac here. I can also run the tests on windows if desired.

brson:
Based on feedback in https://github.com/rust-lang/cargo/issues/8140#issuecomment-2845741150 I pushed another commit that restores the host metadata mixing for filenames, leaving only the symbols reproducible across hosts.

I have run the same manual test against linux and macos and verified the results are still identical for soroban-examples.

I am trying to capture this behavior in a unit test by mocking out a rustc-wrapper that provides two different host triples; but not working yet.

PR #15483, opened 2025-05-02
add more details about cargo fetch and when it will regen cargo.lock

glitch003:

I found the docs on `cargo fetch` unclear with respect to whether or not it will regenerate the Cargo.lock file.  This PR just adds some clarity on when it will actually rebuild it.

Just read the docs change and see if it makes sense or needs to be rewritten.

### DIFF
--- a/src/doc/src/commands/cargo-fetch.md
+++ b/src/doc/src/commands/cargo-fetch.md
@@ -17,6 +17,10 @@ fetch` unless the lock file changes.
 If the lock file is not available, then this command will generate the lock
 file before fetching the dependencies.
 
+If the lock file is not compatible with the current `Cargo.toml`, then this
+command will initiate dependency resolution from the current `Cargo.toml` file,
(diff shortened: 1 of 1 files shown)


src/doc/src/commands/cargo-fetch.md
```diff
@@ -17,6 +17,10 @@ fetch` unless the lock file changes.
 If the lock file is not available, then this command will generate the lock
 file before fetching the dependencies.
 
+If the lock file is not compatible with the current `Cargo.toml`, then this
+command will initiate dependency resolution from the current `Cargo.toml` file,
+and update the lock file.
+
```
weihanglo, line 23: https://doc.rust-lang.org/nightly/cargo/commands/cargo-fetch.html#option-cargo-fetch---locked

This is mentioned in `--locked` flag. Unsure if we need to emphasize it here.

PR #15484, opened 2025-05-03
Improved error message for versions prefixed with `v`

### LINKED ISSUE #12331: Improve the error message when running `cargo add cubing @v0.5.0`

Pyr0de:

- Added an error message when version in `CRATE[@<VER>]` or `--version <VER>` starts with 'v' for `install`, `add`, `yank` and `update --precise <VER>`
- Check if version is valid in `cargo yank` 

Fixes #12331

Added tests for each subcommand

### DIFF
--- a/src/bin/cargo/commands/install.rs
+++ b/src/bin/cargo/commands/install.rs
@@ -286,6 +286,12 @@ fn parse_semver_flag(v: &str) -> CargoResult<VersionReq> {
         .next()
         .ok_or_else(|| format_err!("no version provided for the `--version` flag"))?;
 
+    if let Some(stripped) = v.strip_prefix("v") {
+        bail!(
+            "the version provided, `{v}` is not a valid SemVer requirement\n\n\
+            help: try changing the version to `{stripped}`",
+        )
+    }
     let is_req = "<>=^~".contains(first) || v.contains('*');
     if is_req {
         match v.parse::<VersionReq>() {
--- a/src/bin/cargo/commands/yank.rs
+++ b/src/bin/cargo/commands/yank.rs
@@ -1,5 +1,6 @@
 use crate::command_prelude::*;
 
+use anyhow::Context;
 use cargo::ops;
 use cargo_credential::Secret;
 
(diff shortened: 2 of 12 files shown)


src/bin/cargo/commands/install.rs
weihanglo, line 1: The change looks nice.

Would you mind following [the atomic commit pattern](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request) the contributor guide suggests? Which we add a new test first than captures the current behavior, followed by the next commit containing both the fix and test change. That makes the behavior change more prominent.

weihanglo (approved):
I wonder if there is a way to reuse these logic, but given they are simple it is fine. Thank you!

PR #15486, opened 2025-05-03
Include lock file path in verbose mode output

### LINKED ISSUE #15094: Cargo should say which lock it waits on when run with `--verbose`
### Problem

Due to an unfortunate series of events (rust-analyzer OOM'ing after using >44GiB of RAM) I had `cargo`'s package cache permanently locked. When I'd try to build something all I would see is this:
```text
    Blocking waiting for file lock on package cache
```
(I had no cargo/r-a running at that point)

This is only semi-helpful -- it shows you what's the problem, but not how to fix it.

### Proposed Solution

It would be nice if `cargo` would print the path to the lockfile. In this case that would be something like `~/.cargo/.package-cache` and `~/.cargo/.package-cache-mutate` (deleting those fixed the problem for me). It's probably better to only print those with `--verbose`, since usually you _are_ actually waiting for another `cargo` process.

### Notes

_No response_

Gordon01:

Fixes #15094

When Cargo is run with `--verbose`, the message now includes the path to the lockfile it's waiting on. This makes it easier to debug situations where a stale or stuck lock is preventing progress, especially when no other Cargo process appears to be running.

I've personally encountered this a few times.

Run cargo with `--verbose` and some kind of lock engaged.

Previously, the message simply stated:
> Blocking waiting for file lock on package cache

Now it includes the full path:
> Blocking waiting for file lock on package cache (/home/user/.cargo/.package-cache)


src/cargo/util/flock.rs
```diff
@@ -392,18 +393,20 @@ fn acquire(
...
-        // Force borrow to catch invalid borrows outside of contention situations
-        gctx.shell().verbosity();
-    }
+    let verbose = gctx.shell().verbosity() == Verbosity::Verbose;
     if try_acquire(path, lock_try)? {
         return Ok(());
     }
-    let msg = format!("waiting for file lock on {}", msg);
+    let path = path.display();
+    let msg = if verbose {
```
weihanglo, line 401: Unsure if the discoverability of this behind `--verobse`. Personally I am fine with just printing the lock path unconditionally, though I think per [our contribution policy](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo) maybe we want to discuss this in the original issue <https://github.com/rust-lang/cargo/issues/15094>?

weihanglo:
Completed in https://github.com/rust-lang/cargo/pull/16491, closing

PR #15487, opened 2025-05-04
Add version boundary tests

peiqing6888:
This PR adds several boundary condition tests for the `cargo version` command to improve test coverage and ensure robust behavior in edge cases:

- Test handling of extremely long paths (200 characters)
- Test handling of corrupted config files with invalid TOML and Unicode
- Test response to invalid command line flags

These tests help ensure the version command works reliably in various real-world scenarios that might occur during normal usage. The tests cover:

1. File system edge cases (long paths)
2. Configuration file corruption
3. Command line argument validation

The changes are focused on improving test coverage while maintaining compatibility with existing behavior.

Fixes: None (This is a test coverage improvement)

---
### Testing Done
- All version-related tests pass successfully
- Added new test cases for boundary conditions
- Verified error messages and status codes
- Ensured compatibility with existing behavior

### DIFF
--- a/tests/testsuite/version.rs
+++ b/tests/testsuite/version.rs
@@ -71,13 +71,12 @@ os: [..]
 
 #[cargo_test]
 fn version_with_corrupted_config_dir() {
-    let p = project().file(".cargo/config.toml", "").build();
-
-    // Make the config directory unreadable
-    p.change_file(
-        ".cargo/config.toml",
-        &[0xFF, 0xFE, 0xFF, 0xFF], // Invalid UTF-8
-    );
+    let p = project()
+        .file(
+            ".cargo/config.toml",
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR! However, I'm going to close as I don't feel like these tests are needed. There's already a test for an invalid config (`version_works_with_bad_config`), I don't think we need a test for invalid CLI args (that's pretty well handled by clap and not something I think we need to worry about regressing), and long path tests have risks on Windows due to how cargo gets tested in different environments, and I think we have a little coverage on that and not something that seems to be imminently an issue.

Also, generally we are only accepting PR's for accepted issues, please see our [contribution policy](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo).

PR #2026, opened 2015-10-06
Implement `cargo install`

alexcrichton:
This commit is an implementation of [RFC 1200][rfc] which brings two new
subcommands: `cargo install` and `cargo uninstall`. Most of this is a straight
implementation of the RFC, but a few tweaks were made:

* The `-p` or `--package` arguments are no longer needed as you just pass
  `crate` as a bare argument to the command, this means `cargo install foo`
  works and downloads from crates.io by default.
* Some logic around selecting which crate in a multi-crate repo is installed has
  been tweaked slightly, but mostly in the realm of "let's do the thing that
  makes sense" rather than the literal "let's do what's in the RFC".
  Specifically, we don't pick a crate with examples if there are multiple crates
  with binaries (instead an error is generated saying there are multiple binary
  crates).

[rfc]: https://github.com/rust-lang/rfcs/blob/master/text/1200-cargo-install.md

PR #15489, opened 2025-05-04
fix(install): check if git url contains a rev

### LINKED ISSUE #15336: cargo install --git accepts malformed URL with #, discards it for cloning, then re-adds it into crates.toml
### Problem

Downstream report: https://github.com/nabijaczleweli/cargo-update/issues/293#issuecomment-2741121411

### Steps

1. cargo install --git https://github.com/simonrupf/convert2json.git#65d1408a46668787ba34bf10134ac767c5035710
2. grep convert2json ~/.cargo/crates.toml
3. `"convert2json 1.1.6 (git+https://github.com/simonrupf/convert2json.git#65d1408a46668787ba34bf10134ac767c5035710#74a150d03308dee5a084469ca57623e62560b60b)" = [`

### Possible Solution(s)

Either parse the "anchor" as the rev or don't discard it, since cloning a URL like this is actually illegal
```
$ git clone https://github.com/simonrupf/convert2json.git#65d1408a46668787ba34bf10134ac767c5035710
Cloning into 'convert2json.git#65d1408a46668787ba34bf10134ac767c5035710'...
fatal: unable to update url base from redirection:
  asked for: https://github.com/simonrupf/convert2json.git#65d1408a46668787ba34bf10134ac767c5035710/info/refs?service=git-upload-pack
   redirect: https://github.com/simonrupf/convert2json
```

### Notes

_No response_

### Version

```text
cargo 1.85.1 (d73d2caf9 2024-12-31) I think, the original reporter didn't say.
```

dawedawe:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/pull/15489)*

Check if the git url contains a '#' which is likely a sign for the user's desire to specify a rev. Give hint to use `--rev`

Fixes #15336

### DIFF
--- a/tests/testsuite/install.rs
+++ b/tests/testsuite/install.rs
@@ -2334,6 +2334,32 @@ invalid potential workspace manifest: `[ROOT]/home/.cargo/git/checkouts/foo-[HAS
 }
 
 #[cargo_test]
+fn install_git_with_rev_in_url() {
+    let p = git::repo(&paths::root().join("foo"))
+        .file("Cargo.toml", &basic_manifest("foo", "0.1.0"))
+        .file("src/main.rs", "fn main() {}")
+        .build();
+
+    let rev = p.revparse_head();
+    // The malformed URL is passed to libgit2 as a path containing `#`.
+    // Create that path so the clone succeeds and the install can be tracked.
+    let fragment_path = paths::root().join(format!("foo#{rev}"));
+    git2::Repository::clone(p.url().as_str(), &fragment_path).unwrap();
+
+    let mut url = p.url().to_string();
+    url.push('#');
+    url.push_str(&rev);
+
+    cargo_process("install --locked --git").arg(url).run();
+    assert_has_installed_exe(paths::cargo_home(), "foo");
+    let crates_toml = fs::read_to_string(paths::home().join(".cargo/.crates.toml")).unwrap();
+    assert!(
+        crates_toml.contains(&format!("git+{}#{rev}#{rev}", p.url())),
+        "{crates_toml}"
+    );
+}
+
+#[cargo_test]
 fn install_git_with_symlink_home() {
     // Ensure that `cargo install` with a git repo is OK when CARGO_HOME is a
     // symlink, and uses an build script.
--- a/tests/testsuite/install.rs
+++ b/tests/testsuite/install.rs
@@ -2350,13 +2350,16 @@ fn install_git_with_rev_in_url() {
     url.push('#');
     url.push_str(&rev);
 
-    cargo_process("install --locked --git").arg(url).run();
-    assert_has_installed_exe(paths::cargo_home(), "foo");
-    let crates_toml = fs::read_to_string(paths::home().join(".cargo/.crates.toml")).unwrap();
-    assert!(
-        crates_toml.contains(&format!("git+{}#{rev}#{rev}", p.url())),
-        "{crates_toml}"
-    );
+    cargo_process("install --locked --git")
+        .arg(url)
+        .with_stderr_data(str![[r#"
+[ERROR] invalid url `[ROOTURL]/foo#[..]` to create a git reference: URL fragments are not supported
+
+[HELP] remove the fragment, or use `--rev`, `--branch`, or `--tag` to select a git reference
+
+"#]])
+        .with_status(101)
+        .run();
 }
 
 #[cargo_test]
--- a/src/workspace/parser/mod.rs
+++ b/src/workspace/parser/mod.rs
@@ -2407,16 +2407,6 @@ fn to_dependency_source_id<P: ResolveToPath + Clone>(
                 .or_else(|| orig.rev.clone().map(GitReference::Rev))
                 .unwrap_or(GitReference::DefaultBranch);
             let loc = git.into_url()?;
-
-            if let Some(fragment) = loc.fragment() {
-                let msg = format!(
-                    "URL fragment `#{fragment}` in git URL is ignored for dependency ({name_in_toml}). \
-                        If you were trying to specify a specific git revision, \
-                        use `rev = \"{fragment}\"` in the dependency declaration.",
-                );
-                manifest_ctx.warnings.push(msg);
-            }
-
             SourceId::for_git(&loc, reference)
         }
         (None, Some(path), _, _) => {
--- a/src/workspace/source_id.rs
+++ b/src/workspace/source_id.rs
@@ -217,7 +217,14 @@ impl SourceId {
 
     /// Creates a `SourceId` from a Git reference.
     pub fn for_git(url: &Url, reference: GitReference) -> CargoResult<SourceId> {
-        SourceId::new(SourceKind::Git(reference), url.clone(), None)
+        if url.fragment().is_some() {
+            Err(anyhow::format_err!(
+                "invalid url `{url}` to create a git reference: URL fragments are not supported\n\n\
+                    help: remove the fragment, or use `--rev`, `--branch`, or `--tag` to select a git reference"
+            ))
+        } else {
+            SourceId::new(SourceKind::Git(reference), url.clone(), None)
+        }
     }
 
     /// Creates a `SourceId` from a remote registry URL when the registry name
--- a/tests/testsuite/bad_config.rs
+++ b/tests/testsuite/bad_config.rs
@@ -2397,21 +2397,12 @@ fn fragment_in_git_url() {
         // ...
         // [..]127.0.0.1[..]
         .with_stderr_data(str![[r#"
-[WARNING] Cargo.toml: URL fragment `#foo` in git URL is ignored for dependency (bar). If you were trying to specify a specific git revision, use `rev = "foo"` in the dependency declaration.
-[WARNING] `foo` (manifest) generated 1 warning
-[UPDATING] git repository `http://127.0.0.1/#foo`
-...
-[ERROR] failed to get `bar` as a dependency of package `foo v0.0.0 ([ROOT]/foo)`
+[ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
-  failed to load source for dependency `bar`
+  invalid url `http://127.0.0.1/#foo` to create a git reference: URL fragments are not supported
 
-Caused by:
-  unable to update http://127.0.0.1/#foo
-
-Caused by:
-  failed to clone into: [ROOT]/home/.cargo/git/db/_empty-[HASH]
-...
+  [HELP] remove the fragment, or use `--rev`, `--branch`, or `--tag` to select a git reference
 
 "#]])
         .run();


src/bin/cargo/commands/install.rs
```diff
@@ -172,6 +172,14 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
...
     let source = if let Some(url) = args.get_one::<String>("git") {
+        if url.contains('#') {
+            return Err(anyhow::format_err!(
+                "invalid git url to install from\n\n\
+                    help: use `--rev <SHA>` to specify a commit"
+            )
+            .into());
+        }
+
         let url = url.into_url()?;
```
epage, line 183: Should this check be done on the parsed URL, rather than on the string?

src/bin/cargo/commands/install.rs
```diff
@@ -172,6 +172,14 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
...
 
     let source = if let Some(url) = args.get_one::<String>("git") {
+        if url.contains('#') {
+            return Err(anyhow::format_err!(
+                "invalid git url to install from\n\n\
+                    help: use `--rev <SHA>` to specify a commit"
+            )
+            .into());
+        }
+
```
epage, line 182: Is this definitively a bad URL or should this be a hint on the existing error?

dawedawe: I think it's definitely a bad URL, yes.
From the github page:
2025-05-05-223239_hyprshot

I can't think of a valid git url with a `#` besides the rev.

src/bin/cargo/commands/install.rs
dawedawe: Thanks, yeah. That makes a lot of sense.

src/bin/cargo/commands/install.rs
epage: Cargo supports more than Github.  The question would be more about the git protocol and what is expected around it as we could be turning a success case into an error case.

dawedawe: From what I'm seeing [here](https://git-scm.com/docs/http-protocol), other git servers should be the same.

src/bin/cargo/commands/install.rs
```diff
@@ -173,6 +173,14 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
 
     let source = if let Some(url) = args.get_one::<String>("git") {
         let url = url.into_url()?;
+        if url.fragment().is_some() {
+            return Err(anyhow::format_err!(
+                "invalid git url to install from\n\n\
+                    help: use `--rev <SHA>` to specify a commit"
+            )
```
0xPoe: This error message could maybe be further improved, as it could be anything after the fragment in an invalid Git URL.

Maybe we could say:
```log
invalid git url to install from: URL fragments are not supported

help: remove the fragment, or use `--rev`, `--branch`, or `--tag` to select a git reference
```

0xPoe (commented):
I did some digging through the history to understand why we have this issue.

Cargo has long used #<commit> internally to serialize the resolved git revision in SourceId (e465ace4, 2014), which is reasonable for lockfile/source tracking.

cargo install --git was added later in PR #2026 / commit bc60f64b (2015). It parsed the user-provided git URL but did not reject URL fragments. Install tracking then serialized the package source as: original URL + Cargo's internal #<resolved_commit>.

So https://host/repo.git#foo became git+https://host/repo.git#foo#actual_commit.

Thanks for working on this. Not sure if you still want to revisit this one, but I think the check fragment seems reasonable here.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/cargo/15489/6cba807e2ce0b7b4bcdf3bdf0a07dcf83988c05a..dcc8cef7c359515b9906e32fd3ffde4e87c6cae3)*

src/bin/cargo/commands/install.rs
dawedawe: sure, why not

tests/testsuite/install.rs
```diff
@@ -2333,6 +2333,33 @@ invalid potential workspace manifest: `[ROOT]/home/.cargo/git/checkouts/foo-[HAS
...
+[WARNING] spurious network error (1 try remaining): failed to resolve path '[ROOT]/foo#0e8d88b5cfc173c5f5a6a0fe0ce1d4e6018600d4': No such file or directory; class=Os (2)
+[ERROR] failed to clone into: [ROOT]/home/.cargo/git/db/foo-[HASH]
+
+Caused by:
+  failed to resolve path '[ROOT]/foo#0e8d88b5cfc173c5f5a6a0fe0ce1d4e6018600d4': No such file or directory; class=Os (2)
+
+"#]])
+        .with_status(101)
+        .run();
+}
```
0xPoe, line 2363:
```rust
#[cargo_test]
fn install_git_with_rev_in_url() {
    let p = git::repo(&paths::root().join("foo"))
        .file("Cargo.toml", &basic_manifest("foo", "0.1.0"))
        .file("src/main.rs", "fn main() {}")
        .build();

    let rev = p.revparse_head();
    // The malformed URL is passed to libgit2 as a path containing `#`.
    // Create that path so the clone succeeds and the install can be tracked.
    let fragment_path = paths::root().join(format!("foo#{rev}"));
    git2::Repository::clone(p.url().as_str(), &fragment_path).unwrap();

    let mut url = p.url().to_string();
    url.push('#');
    url.push_str(&rev);

    cargo_process("install --locked --git").arg(url).run();
    assert_has_installed_exe(paths::cargo_home(), "foo");
    let crates_toml = fs::read_to_string(paths::home().join(".cargo/.crates.toml")).unwrap();
    assert!(
        crates_toml.contains(&format!("git+{}#{rev}#{rev}", p.url())),
        "{crates_toml}"
    );
}
```

This would make this case really catch the buggy behavior from the issue.

0xPoe (commented):
@epage I'm not sure whether we need to consider this a breaking change or just a bug. It does look like a bug, but I'm not sure if someone would really rely on this. What are your thoughts on this?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/cargo/15489/e009095bbb13e72664be7a0b06a318798f6f5009..f9b37b92e6fc059ad53b87226bc1b1a34279a05b)*

tests/testsuite/install.rs
dawedawe: @0xPoe , thanks, that's a nice repro indeed.

dawedawe:
Hey @0xPoe, @epage, is there anything more I can do to drive this forward?

weihanglo:
> I'm not sure whether we need to consider this a breaking change or just a bug. It does look like a bug, but I'm not sure if someone would really rely on this. What are your thoughts on this?


It is more like a bug indeed, and I sympathize who ever relying on this. Though to be a bit cautious we can give a transition period and turn it a hard warning I guess?

0xPoe:
> > I'm not sure whether we need to consider this a breaking change or just a bug. It does look like a bug, but I'm not sure if someone would really rely on this. What are your thoughts on this?
> 
> It is more like a bug indeed, and I sympathize who ever relying on this. Though to be a bit cautious we can give a transition period and turn it a hard warning I guess?

I tried to check whether that change would have any edge cases. But it seems there are no bad cases, as Git’s standard behavior never allows fragments in the URL: https://git-scm.com/docs/git-clone#_git_urls

https://git-scm.com/docs/http-protocol And from this documentation, Git seems to use only the path and query, and no fragment is used at all. And for most Git platforms, the fragment is used to create the permalink. 

But I did notice that npm would accept git+https://github.com/org/foo.git#v1.0.0 as a valid URL to pick version 1.0.0, and so would other JavaScript tools like Yarn and Bun. On the other hand, would their code prove that it is safe to assume there is no # in the repository name at all? Otherwise, they would hit the bug caused by having too many # characters in the URL. https://github.com/pnpm/pnpm/blob/fbf78bd7d26a3da41b384f1ee35539992b4f799b/pnpm11/resolving/git-resolver/src/parseBareSpecifier.ts#L51

So I guess going with the direct error would be OK?

0xPoe:
> Hey @0xPoe, @epage, is there anything more I can do to drive this forward?

The changes look good to me. I would like to wait for @weihanglo to check my response and decide whether we should add a transition period before turning it into a hard error.

weihanglo:
Sounds reasonable. Can go either way.

0xPoe (approved):
Thanks! :shipit: 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/cargo/15489/400deedfc638db71d7f0539cbaa388faabf41c6f..b6d0cee1539e3c6b67c2b1840248ee66431ddf81)*

0xPoe:
@epage Just want to double-check with you before I merge it. Do you think a hard error would be OK here? Or do you prefer we go with a transition period here?

src/bin/cargo/commands/install.rs
epage: I might be missing it but I'm not seeing anything at that link saying servers will reject fragments.

src/bin/cargo/commands/install.rs
```diff
@@ -178,6 +178,14 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
 
     let source = if let Some(url) = args.get_one::<String>("git") {
         let url = url.into_url()?;
+        if url.fragment().is_some() {
```
epage: This change is directly in `cargo install`.

Does this mean use of fragments in dependencies gets the old behavior?

Does `cargo add --git` error before we write to the manifest?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/cargo/15489/400deedfc638db71d7f0539cbaa388faabf41c6f..b6d0cee1539e3c6b67c2b1840248ee66431ddf81)*

dawedawe: > This change is directly in `cargo install`.
> 
> Does this mean use of fragments in dependencies gets the old behavior?

Yes, that behavior isn't touched.
 
> Does `cargo add --git` error before we write to the manifest?

No, it doesn't error and doesn't warn from what I see.

src/bin/cargo/commands/install.rs
dawedawe: Sorry, I don't remember what I thought back then.

src/bin/cargo/commands/install.rs
epage: > Yes, that behavior isn't touched.

is there a reason we are putting it in this layer so that it only affects one code path, rather than lower down so it affects all? For example, if we keep the error as is, what if we moved this into `SourceId`? That doesn't cover `git add` but it does cover quite a good number of cases.

dawedawe: Mmh, maybe I misunderstand you but wouldn't disallowing a rev in `SourceId` make all the `Cargo.toml` files out there that already have such a git url fail to parse?

epage: Not disallowing it in the `SourceId` URL but in the URL we pass to `SourceId::for_git`.

dawedawe: Having the error inside of `SourceId::for_git` escalates a warning to an error if there's a malformed git dependency in a `Cargo.toml`. I'm afraid this could affect lots of people.
Checking the argument at all callsides or changing the return of `SourceId::for_git` to keep it as a potential warning is, of course, against the idea to catch all code paths with a central check.

PR #15496, opened 2025-05-06
Make cargo script ignore workspaces

arctic-alpaca:

Cargo script considers itself part of the workspace if the script is located in a sub-directory of a workspace (presumably since https://github.com/rust-lang/cargo/pull/15168). This becomes an issue when using a custom registry that is defined in the `.cargo/config.toml` within the workspace. Cargo script does not take that file into account and fails with ``registry index was not found in any configuration: `test-reg` ``.

This PR adds a regression test and makes cargo script ignore the surrounding workspace.
The test ~~will fail without the fix in the second commit and~~ can be used to reproduce the issue.

### Additional information

The issue started occurring with `nightly-2025-02-16`.
Related to https://github.com/rust-lang/cargo/issues/12207.

### DIFF
--- a/tests/testsuite/script.rs
+++ b/tests/testsuite/script.rs
@@ -1923,19 +1923,17 @@ members = [
 
     p.cargo("-Zscript -v script/echo.rs")
         .masquerade_as_nightly_cargo(&["script"])
-        .with_stdout_data(str![[r#""#]])
-        .with_stderr_data(str![[r#"
-[WARNING] `package.edition` is unspecified, defaulting to `2024`
-[ERROR] failed to load manifest for workspace member `[ROOT]/foo/inner`
-referenced by workspace at `[ROOT]/foo/Cargo.toml`
-
-Caused by:
-  failed to parse manifest at `[ROOT]/foo/inner/Cargo.toml`
-
-Caused by:
-  registry index was not found in any configuration: `test-reg`
+        .with_stdout_data(str![[r#"
+bin: [ROOT]/home/.cargo/target/[HASH]/debug/echo[EXE]
+args: []
 
 "#]])
-        .with_status(101)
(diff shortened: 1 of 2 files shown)


tests/testsuite/script.rs
```diff
@@ -1886,3 +1886,54 @@ CARGO_MANIFEST_PATH: [ROOT]/foo/script.rs
 "#]])
         .run();
 }
+
+#[cargo_test(nightly, reason = "edition2024 hasn't hit stable yet")]
+fn ignore_surrounding_workspace() {
```
epage, line 1891: We ask that all commits be atomic which includes tests passing.

If you could update this commit so that the test passes, that'd be appreciated!  We do like having the test commit split out like this.  In doing so, the diff for the fix commit shows how the behavior was fixed (e.g. #15497)

tests/testsuite/script.rs
```diff
@@ -1886,3 +1886,54 @@ CARGO_MANIFEST_PATH: [ROOT]/foo/script.rs
 "#]])
         .run();
 }
+
+#[cargo_test(nightly, reason = "edition2024 hasn't hit stable yet")]
```
epage, line 1890: This test doesn't need nightly.  I know this is copied but I'm prepping a PR to remove it from the rest

epage:
Thanks!  I remember considering this case and am unsure how I lost track of it

tests/testsuite/script.rs
arctic-alpaca: Sure, test now passes in both commits.

tests/testsuite/script.rs
arctic-alpaca: I was wondering about that but decided for consistency, removed now.

epage (approved):
Thanks!

Issue #15505, opened 2025-05-08
Suggest workspace shorthands for boolean dependencie


0xsot:
Error complains about a boolean but only mentions version strings or tables with version = "x.y.z" when you accidentally write:

```toml
[dependencies]
crc32fast = true
```
Cargo reports:
```log
invalid type: boolean `true`, expected a version string like "0.9.8"
or a detailed dependency like { version = "0.9.8" }
in `dependencies.crc32fast`
```
It would be more helpful if the error message also suggested the workspace shorthand. Both of these are valid ways to pull from your workspace
```toml
crc32fast = { workspace = true }
crc32fast.workspace = true
```

### Proposed behavior:

When a boolean is found, append a note pointing out the two valid workspace forms
```log
invalid type: boolean `true`, expected a version string like "0.9.8"
or a detailed dependency like { version = "0.9.8" }
in `dependencies.crc32fast`.

if you meant to use a workspace member, you can write
  crc32fast = { workspace = true }
or
  crc32fast.workspace = true
```

epage:
I've gone ahead and marked this as accepted.

For anyone who pick this up, as a reminder, we would prefer that PRs be split up
- A commit for adding tests in their own commit before other changes, showing the existing error message(s)
- A commit for the bool change that updates the test to show the new error output
- If you do ints/floats, please fix that in a separate commit

For where to put the tests, https://github.com/rust-lang/cargo/blob/a154422cec0fe9ef2ca8aa6501debc2db61186e7/tests/testsuite/bad_config.rs#L2618 is already a test for an integer, so that could be used for that case and a bool test could be added near it.

borngraced:
@epage i want to give this a shot

epage:
Thanks!

motorailgun:
Is this issue already solved by #15507 and should be closed? Just wondering...

PR #15507, opened 2025-05-09
suggest workspace hints for boolean dependencies

0xsot:
via issue #15505, Cargo currently errors on  
```toml
[dependencies]
crc32fast = true
```
with a message about `expected a version string or a detailed dependency` It doesn’t hint that you can depend on a workspace member. This PR updates `cargo` so that when you write dep = true, the error also suggests:
```toml
dep = { workspace = true }
dep.workspace = true
```

In a workspace crate’s Cargo.toml, add
```toml
[dependencies]
crc32fast = true
```
Run `cargo build` and you should see the enhanced error with the two workspace hints.

### Additional information

Regarding support for int/float, we only need a special case for boolean because they’re the only values that get the `workspace hint`. Everything else just uses the normal detailed dependency error

### DIFF
--- a/crates/cargo-util-schemas/src/manifest/mod.rs
+++ b/crates/cargo-util-schemas/src/manifest/mod.rs
@@ -787,12 +787,26 @@ impl<'de, P: Deserialize<'de> + Clone> de::Deserialize<'de> for TomlDependency<P
     where
         D: de::Deserializer<'de>,
     {
+        use serde::de::Error as _;
+        let expected = "a version string like \"0.9.8\" or a \
+                     detailed dependency like { version = \"0.9.8\" }";
         UntaggedEnumVisitor::new()
-            .expecting(
-                "a version string like \"0.9.8\" or a \
-                     detailed dependency like { version = \"0.9.8\" }",
-            )
+            .expecting(expected)
             .string(|value| Ok(TomlDependency::Simple(value.to_owned())))
+            .bool(|value| {
+                let expected = format!("invalid type: boolean `{value}`, expected {expected}");
+                let err = if value {
+                    format!(
+                        "{expected}\n\
+                    note: if you meant to use a workspace member, you can write\n \
+                      dep.workspace = {value}"
+                    )
+                } else {
+                    expected
+                };
+
+                Err(serde_untagged::de::Error::custom(err))
+            })
             .map(|value| value.deserialize().map(TomlDependency::Detailed))
             .deserialize(deserializer)
     }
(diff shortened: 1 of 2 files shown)


crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -776,12 +776,22 @@ impl<'de, P: Deserialize<'de> + Clone> de::Deserialize<'de> for TomlDependency<P
...
-                "a version string like \"0.9.8\" or a \
-                     detailed dependency like { version = \"0.9.8\" }",
-            )
+            .expecting(expected)
             .string(|value| Ok(TomlDependency::Simple(value.to_owned())))
+            .bool(|value| {
+                let err = format!(
+                    "invalid type: boolean `{value}`, expected {expected}\n\
+                    note: if you meant to use a workspace member, you can write\n \
+                      dep = {{ workspace = {value} }}\n\
```
epage, line 789: let's drop this option.

in our docs, we decided to standardize on the other form if they have nothing else to mention which is the case for `dep = true`

crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -776,12 +776,22 @@ impl<'de, P: Deserialize<'de> + Clone> de::Deserialize<'de> for TomlDependency<P
...
+                    or\n \
+                      dep.workspace = {value}",
```
epage, line 791: This will provide a wrong message for `dep = false` (unsupported),

Should we skip the hint in that case?

borngraced: make sense we can show the default(`expected`) hint for `false` bool...this is what you mean rigiht?

epage: Yup

borngraced: @epage I updated my previous comment lmk your thought please else I will continue with showing `expected` message for `false` value.

borngraced: nvm actually, I thought about it for a sec again and I think your suggestion is the obvious choice

epage:
Mind squashing that last commit into the previous one?

weihanglo:
@rustbot author

Could you rebase onto master? Thanks!

epage:
Looks like a new nightly broke `build-std`.  I've asked for input on fixing it at [#t-cargo > &#96;build-std&#96; test failure @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/.60build-std.60.20test.20failure/near/535180418)

epage (approved):

Issue #11577, opened 2023-01-13
Cargo built-in Git/SSH client doesn't support `@cert-authority` or `@revoked` markers


hds:
### Problem

Cargo parses SSH known hosts file. From the Cargo book (https://doc.rust-lang.org/cargo/appendix/git-authentication.html#ssh-known-hosts):

> When connecting to an SSH host, Cargo must verify the identity of the host using "known hosts", which are a list of host keys. Cargo can look for these known hosts in OpenSSH-style known_hosts files located in their standard locations ...

However, there are some additional markers supported by at least some SSH clients (e.g. OpenSSH) to handle more complex cases than verifying a host via a single algorithm/key. The known ones are:
* `@cert-authority`
* `@revoked`

The Cargo SSH client doesn't support these directives. It is quite explicit about this in the code:
https://github.com/rust-lang/cargo/blob/1cd6d3803dfb0b342272862a8590f5dfc9f72573/src/cargo/sources/git/known_hosts.rs#L490-L493

With the release of Rust 1.66.1 and the fix for [CVE-2022-46176](https://www.cve.org/CVERecord?id=CVE-2022-46176) ([security advisory](https://blog.rust-lang.org/2023/01/10/cve-2022-46176.html)), Cargo is now performing host key checking, which will lead to more users needing this functionality because single host key verification may not be practical.

### Proposed Solution

The solution to this issue would be to implement the missing support for the `@cert-authority` or `@revoked` markers.

There is useful documentation on these markers from the OpenSSH project:
* `sshd` SSH_KNOWN_HOSTS FILE FORMAT: https://man.openbsd.org/sshd#SSH_KNOWN_HOSTS_FILE_FORMAT (from @weihanglo's comment)
* Certificate format: https://cvsweb.openbsd.org/src/usr.bin/ssh/PROTOCOL.certkeys?annotate=HEAD (from @est31's comment)

This issue can be mitigated by telling cargo to use the command line Git client (`net.git-fetch-with-cli = true`) as mentioned by @weihanglo on this Zulip thread: https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Cargo.20SSH.20host.20key.20verification.20with.20.40cert-authority.20lines

[...]

PR #15508, opened 2025-05-09
add glob pattern support for known_hosts

motorailgun:

This PR implements glob pattern match for known_hosts file. Hosts written with `*` and `?` now matches correctly.

### Tests

Tests are added accordingly.

### Miscs

This is my first time submitting PR, sorry if there's anything that's off although I've read the contributor guide.

### DIFF
--- a/src/cargo/sources/git/known_hosts.rs
+++ b/src/cargo/sources/git/known_hosts.rs
@@ -23,6 +23,7 @@
 //! and revoked markers. See "FIXME" comments littered in this file.
 
 use crate::util::context::{Definition, GlobalContext, Value};
+use crate::util::restricted_names::is_glob_pattern;
 use crate::CargoResult;
 use base64::engine::general_purpose::STANDARD;
 use base64::engine::general_purpose::STANDARD_NO_PAD;
@@ -588,7 +589,19 @@ impl KnownHost {
         }
         for pattern in self.patterns.split(',') {
             let pattern = pattern.to_lowercase();
-            // FIXME: support * and ? wildcards
+            let is_glob = is_glob_pattern(&pattern);
+
+            if is_glob {
+                match glob::Pattern::new(&pattern) {
+                    Ok(glob) => match_found |= glob.matches(&host),
+                    Err(e) => {
+                        tracing::warn!(
+                            "failed to interpret hostname `{pattern}` as glob pattern: {e}"
+                        )
+                    }
+                }
+            }
+
             if let Some(pattern) = pattern.strip_prefix('!') {
                 if pattern == host {
                     return false;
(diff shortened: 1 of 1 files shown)


motorailgun:
Oops, fixing the failing tests...

src/cargo/sources/git/known_hosts.rs
```diff
@@ -588,7 +590,18 @@ impl KnownHost {
...
         for pattern in self.patterns.split(',') {
             let pattern = pattern.to_lowercase();
-            // FIXME: support * and ? wildcards
+            let is_glob = is_glob_pattern(&pattern);
+
+            if is_glob {
+                let regex_pattern =
+                    regex::escape(&pattern.replace("*", ",,,,").replace("?", ";;;;"))
+                        .replace(",,,,", ".*")
+                        .replace(";;;;", ".");
```
epage, line 599: Is there a reason you are manually ensuring `*` and `?` get passed through `regex::escape` vs using `glob`?

src/cargo/sources/git/known_hosts.rs
```diff
@@ -588,7 +590,18 @@ impl KnownHost {
...
+                if let Ok(regex) = Regex::new(&regex_pattern) {
```
epage, line 600: Should we do anything on errors?

src/cargo/sources/git/known_hosts.rs
motorailgun: Oh, I somehow thought `glob` is only applicable for paths. Thank you for pointing this out!

src/cargo/sources/git/known_hosts.rs
motorailgun: Hmm, my thought is that if the pattern string could not be interpreted as glob pattern it's better to just pass glob matching.

src/cargo/sources/git/known_hosts.rs
```diff
@@ -588,7 +589,14 @@ impl KnownHost {
         }
         for pattern in self.patterns.split(',') {
             let pattern = pattern.to_lowercase();
-            // FIXME: support * and ? wildcards
+            let is_glob = is_glob_pattern(&pattern);
+
+            if is_glob {
+                if let Ok(regex) = glob::Pattern::new(&pattern) {
+                    match_found |= regex.matches(&host);
```
weihanglo, line 596: It is not regex anymore, so perhaps rename to `glob`?

src/cargo/sources/git/known_hosts.rs
```diff
@@ -588,7 +589,14 @@ impl KnownHost {
         }
         for pattern in self.patterns.split(',') {
             let pattern = pattern.to_lowercase();
-            // FIXME: support * and ? wildcards
```
weihanglo, line 591: Was under the impression that this feature is tracked in <https://github.com/rust-lang/cargo/issues/11577>, but apparently not.

It seems like OpenSSH [supports only `*` and `?` wildcards](https://man.openbsd.org/sshd.8#SSH_KNOWN_HOSTS_FILE_FORMAT). One caveat of using `glob` here is we provide more than we should. Anyway, it is good to catch up with the common format.

weihanglo (commented):
Thanks. Just one nit and we are good to go!

src/cargo/sources/git/known_hosts.rs
weihanglo: We could probably emit a `tracing::warn!` on failing cases.

src/cargo/sources/git/known_hosts.rs
motorailgun: Oh! How could I miss this one! Thank you for pointing out :)

src/cargo/sources/git/known_hosts.rs
motorailgun: Sounds nice!

weihanglo (approved):
Great! Thanks for the contribution :)

PR #15512, opened 2025-05-09
Revert "doc: Mention `XDG_DATA_HOME`"

epage:
This reverts commit 15a7672dde9853a8b15b545a1e5da5bbee7244c9.

There was a miscommunication about some outstanding concerns with #15480 and it got merged before we were ready.

Specifically
- How important is this for unstable docs?
- How universal is the use of `bash-completions` and what should we do for people who don't have it?
- How likely is someone to catch that they don't have `XDG_DATA_HOME` set and get tripped up over this documentation
- Should we be aligned with upstream clap?

### DIFF
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -1817,7 +1817,7 @@ When in doubt, you can discuss this in [#14520](https://github.com/rust-lang/car
 
 ### How to use native-completions feature:
 - bash:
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/unstable.md
weihanglo, line 1: > How universal is the use of bash-completions and what should we do for people who don't have it?


I was under the impression that clap-complete relies on bash-completion because the current completion code uses a function from it:

https://github.com/rust-lang/cargo/blob/c566be4742469d8b9ed7a59ecc3f753fd085606d/src/etc/cargo.bashcomp.sh#L9

But that doesn't seem to be the case after I re-read the source code of clap-complete, so yes I think we should revert.

weihanglo (approved):

Issue #13191, opened 2023-12-21
Normalization of `Cargo.toml` can break crates with deps. that read `Cargo.toml`


ErichDonGubler:
### Problem

Commentary surrounding items in a crate's `Cargo.toml` is stripped when `cargo package` or `cargo vendor` are used to vendor in said crate. For most crates, this does not cause a problem. However, this _can_ cause build regressions or changes for crates with dependencies that use `Cargo.toml`'s contents. As a concrete example of this happening, [`document-features`](https://github.com/slint-ui/document-features/) uses `Cargo.toml` to emit documentation for features. However, a few users of crates that consume `document-features` indirectly have found that `cargo vendor` copies of crates run into build regressions (see also [`slint-ui/document-features`#20](https://github.com/slint-ui/document-features/issues/20)), because feature commentary is missing from the "normalized" `Cargo.toml` manifests.

### Steps

1. First, create a "dependency" crate that depends on [`document_features` 0.2.7](https://docs.rs/document-features/0.2.7/document_features/?search=document-features), uses its `document_features!(…)` macro, and which has at least one "documented" dependency according to `document-features`' convention. For the purposes of these steps, it is important that this crate be readily vendor-able via `cargo vendor`, hence a Git repo will be recommended from this point forward.
	* For convenience, one can just use the provided minimum reproducible example "dependency" crate at <https://gitlab.com/erichdongubler-mre/2023-12-cargo-toml-normalization-breakage>.
	
	    * `Cargo.toml`:
	
	    ```toml
		[package]
		name = "lib"
		version = "0.1.0"
		edition = "2021"
		
		[features]
		## ermehgersh so beautiful
		my_awesome_feature = []
		
		[dependencies]
		document-features = "0.2.7"
	    ```
	    
	    * `src/lib.rs`:
	    
	    ```rust
		//! # Features
		//!
		#![doc = document_features::document_features!(feature_label = r#"<span class="stab portability"><code>{feature}</code></span>"#)]
	    ```
  	
2. Next, create the "consumer" crate, which depends on the "dependency" crate 

	* For convenience, one can use this `Cargo.toml` file as a replacement for a crate otherwise generated with `cargo new --lib …`; continuing with the `lib` crate as defined above:
	
		```toml
		[package]
		name = "consumer"
		version = "0.1.0"
		edition = "2021"
		
		[dependencies]
[...]

PR #15514, opened 2025-05-10
fix(vendor)!: direct extraction for registry sources

### LINKED ISSUE #9054: `cargo vendor` vs `cargo package` inconsistency
**Problem**
When using `cargo vendor`, I end up with fewer files than expected compared to if I used `cargo package` 
instead.

More specifically, I expected the `exclude` of `Cargo.toml` to be handled consistently in the `package` and `vendor` commands. For the deno project, the exclusions are listed here (note: there are inverted exclusions too. aka: `!somefile` = include somefile):

https://github.com/denoland/rusty_v8/blob/54a72d7628028063f82802f11e48c7eac89f96bf/Cargo.toml#L11-L50

I expected the vendor folder to contain a file named `.gn`, which is required for building the project.

**Steps**
```sh
$ cd path/to/denoland/deno
$ cargo vendor
$ ls  -la ./vendor/rusty_v8/

# The problem here is that we are not seeing .gn,
# and a few other files that are required for the
# deno build. This means the ./vendor cannot be used
# to build.
# 
# But if we "cargo package"
# instead that seems to work.
```

See the original report to https://github.com/denoland/rusty_v8/issues/566 for more logs and other info.

**Possible Solution(s)**
Maybe the vendoring and publishing code paths are divergent somewhere?

**Notes**

Platform info

```
❯ uname -o
❯ uname -r
❯ cargo version
❯ rustc --version
GNU/Linux
5.10.3-1-default
cargo 1.47.0 (f3c7e066a 2020-08-28)
rustc 1.47.0 (18bf6b4f0 2020-10-07)
```

### LINKED ISSUE #9555: Inconsistency with publishing/vendoring listing of files
Currently I'm seeing some weird behavior in Cargo with respect to vendoring and publishing. Notably this project will succeed to compile but then fail to compile with a vendored build:

```
$ cargo new foo
$ cd foo
$ echo 'zstd-sys = "=1.6.0"' >> Cargo.toml
$ cargo build
...
    Finished dev [unoptimized + debuginfo] target(s) in 30.45s
$ mkdir .cargo
$ cargo vendor >> .cargo/config.toml
$ cargo build
error: failed to run custom build command for `zstd-sys v1.6.0+zstd.1.5.0`

Caused by:
  process didn't exit successfully: `/Users/acrichton/code/foo/target/debug/build/zstd-sys-2992c87aa8f96b70/build-script-build` (exit code: 101)
  --- stderr
  thread 'main' panicked at 'Folder 'zstd/lib' does not exists. Maybe you forgot to clone the 'zstd' submodule?', /Users/acrichton/code/foo/vendor/zstd-sys/build.rs:180:13
```

Sure enough the `zstd` folder is indeed missing from the vendored checkout. Some other tidbits:

* The [tarball itself](https://crates.io/api/v1/crates/zstd-sys/1.6.0+zstd.1.5.0/download) does indeed contain the `zstd` folder
* The manifest indicates an [`exclude` directive](https://github.com/gyscos/zstd-rs/blob/2dead48d0075ccaab20fc27f19c578a74a2b2e8f/zstd-safe/zstd-sys/Cargo.toml#L21-L27) which presumably tries to filter the files included

I believe the problem here is that `cargo vendor` uses `list_files` to list the files that are going to be used for publication/building, and something funny is happening here with the presence of a `.git` folder. For example [checking out `zstd-sys`](https://github.com/gyscos/zstd-rs/blob/2dead48d0075ccaab20fc27f19c578a74a2b2e8f/zstd-safe/zstd-sys) and running `cargo package -v` does indeed vendor the `zstd` folder. If you delete the `zstd/.git` folder, however, and then run `cargo package -v` the `zstd` folder is not included. This I think best simulates what's happening with the registry because it's presumably published from a source checkout (therefore including `zstd`), but it's then vendored from a checkout without `.git` (what was downloaded from crates.io). This means that vendors don't have the `zstd` folder but downloads from crates.io do.

I haven't dug too much into this but the bug is likely around `list_files` and how the `exclude` directory is handled relative to discovered submodules. We should probably warn about this or handle this differently at publication time, ideally publication and vendoring would work with similar sets of files.

### LINKED ISSUE #9575: `cargo vendor` does not include license-file if not in package list

Mono-repos usually have a single `LICENSE` file in the top directory and they reference it from all crates via some `license-file = "../../LICENSE"`.

At the moment, cargo will not package file from such path.

Since the license file needs to be packaged into the crate for it to be able to be vendored in into other projects.

We encountered that problem when trying to vendor in ICU4X (rust mono repo) into Firefox. 

In order to comply with the vendor policy of Mozilla, we need to duplicate the `LICENSE` file to each component of the mono repo and introduce a maintenance script to ensure the licenses stay in sync.

See the [issue](https://github.com/unicode-org/icu4x/issues/771), and the [PR](https://github.com/unicode-org/icu4x/pull/790).

It would help such projects if files referenced in `Cargo.toml` were packaged even if they live in the parent directory to the crate root.

The file could be copied to the root directory of the package.

**Notes**

If that's impossible, I'd at least suggest displaying a warning that cargo skips packaging of the file.

### LINKED ISSUE #11000: `cargo vendor` doesn't generate `.cargo_vcs_info.json`
### Problem

I am not sure if this really is a bug. But rust-skia/skia-bindings' build script expects that there is either a `.cargo_vcs_info.json` in the crate directory, or the crate directory is a git repo. Because `cargo vendor` doesn't create a `.cargo_vcs_info.json`, `skia-bindings` fails to build.

https://github.com/rust-skia/rust-skia/issues/682

### Steps

_No response_

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

_No response_

### LINKED ISSUE #14034: `cargo vendor` misses some files
### Problem

I'm trying to build [this Cargo project](https://github.com/andrewtbiehl/kramdown-syntax_tree_sitter/tree/main/ext/tree_sitter_adapter) with vendored deps. I'm able to build it normally (`cargo build`), and running `cargo vendor` exits without an error, but building using the vendored deps fails (`cargo build` after modifying `.cargo/config.toml`).

It's pretty clear why it fails to build — `cargo vendor` hasn't copied [the `src/generate/templates` directory of the `tree-sitter-cli` dependency](https://github.com/tree-sitter/tree-sitter/tree/master/cli/src/generate/templates), so I see a bunch of errors due to `include_str!` trying to include nonexistent files.

I'm not a Cargo expert, so I might be missing something, but I didn't see any reason why vendoring the dependencies would 
miss this subfolder. Hopefully you can point me in the right direction if there's something obvious I missed.

### Steps

1. `git clone https://github.com/andrewtbiehl/kramdown-syntax_tree_sitter`
2. `cd kramdown-syntax_tree_sitter/ext/tree_sitter_adapter`
3. `RUBY=/path/to/bin/ruby cargo build` (observe that this succeeds)
4. `cargo vendor` (observe that this appears to succeed)
5. Modify `.cargo/config.toml` as directed by the output of `cargo vendor`
6. `cargo build` (observe that this fails)
7. There will also be no folder `vendor/tree-sitter-cli/src/generate/templates`, although it does exist in the upstream repo

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
cargo 1.76.0 (c84b36747 2024-01-18)
release: 1.76.0
commit-hash: c84b367471a2db61d2c2c6aab605b14130b8a31b
commit-date: 2024-01-18
host: aarch64-apple-darwin
libcurl: 8.6.0 (sys:0.4.70+curl-8.5.0 system ssl:(SecureTransport) LibreSSL/3.3.6)
os: Mac OS 14.5.0 [64-bit]
```

### LINKED ISSUE #15080: `cargo vendor` loses some nonempty directories
### Problem

For example, this one: https://docs.rs/crate/bat/0.25.0/source/src/syntax_mapping/builtins/macos/

### Steps

Confirm this directory exists in the published crate, as shown in the docs.rs link above.

```console
$ curl -L https://crates.io/api/v1/crates/bat/0.25.0/download -o bat-0.25.0.crate

$ tar tf bat-0.25.0.crate bat-0.25.0/src/syntax_mapping/builtins/macos
bat-0.25.0/src/syntax_mapping/builtins/macos/.gitkeep
```

Now let's try vendoring.

```toml
# Cargo.toml
[package]
name = "repro-cargo"
edition = "2021"

[dependencies]
bat = "=0.25.0"
```

```console
$ cargo vendor

$ ls vendor/bat/src/syntax_mapping/builtins/
bsd-family  common  linux  README.md  unix-family
```

"macos" is not there.

In bat-0.25.0's case, this actually breaks the crate (when built on macos from vendored sources).

```console
error: failed to run custom build command for `bat v0.25.0`

Caused by:
  process didn't exit successfully: `target/debug/build/bat-126fec914d2804b6/build-script-main` (exit status: 1)
  --- stdout
  cargo:rerun-if-changed=build/
  cargo:rerun-if-changed=src/syntax_mapping/builtins/

  --- stderr
  Error: IO error for operation on src/syntax_mapping/builtins/macos: No such file or directory (os error 2)

  Caused by:
      No such file or directory (os error 2)
```

### Possible Solution(s)

_No response_

### Notes

_No response_

### Version

```text
cargo 1.86.0-nightly (088d49608 2025-01-10)
release: 1.86.0-nightly
commit-hash: 088d496082726091024f1689c124a0c3dccbd775
commit-date: 2025-01-10
```

### LINKED ISSUE #15090: `cargo vendor` should vendor crates in a pristine state
In a large project, where different people might run cargo vendor with different versions of cargo, it would be better to avoid the unwanted noise from each version of cargo introducing changes to the `cargo publish` output.

Things were mostly fine until cargo 1.80, but since then almost each and every release of cargo has introduced changes that introduces churn on `Cargo.toml`, `Cargo.lock` and consequently `.cargo-checksum.json`.

Things like issue #14965 also wouldn't happen if the vendored crates were unmodified.

weihanglo:

`PathSource::list_files` has some heurstic rules for listing files. Those rules are mainly designed for `cargo package`.

Previously, cargo-vendor relies on those rules to understand what files to vendor. However, it shouldn't use those rules because:

* Package extracted from a `.crate` tarball isn't Git-controlled, some rules may apply differently.
* The extracted package already went through `PathSource::list_files` during packaging. It should be clean enough.
* Should keep crate sources from registry sources in a pristine state, which is exactly what vendoring is meant for.

Instead, we switch to direct extraction into the vendor directory
to ensure source code is the same as in the `.crate` tarball.

There is already a test `vendor::package_exclude` for the fix of #9054,
though now I think it is not a good fix. The test change shows the correct behavior change.

I am not sure if we want more tests.

Also, there are some caveats with this fix:

* The overwrite protection in `unpack_package` assumes the unpack
  directory is always `<pkg>-<version`>.
  We don't want to remove this,
  but for cargo-vendor supports vendoring without version suffix.
  For that case, we need a temporary staging area,
  and move the unpacked source then.
* The heurstic in `PathSource::list_files` did something "good" in
  general cases, like excluding hidden directories. That means
  common directories like `.github` or `.config` won't be vendored.
  After this, those get included. This is another round of churns.
  We might want to get other `cargo-vendor` changes along with this
  in one single release.

### Additional information

* Fixes #9054
* Fixes #9555
* Fixes #9575
* Fixes #11000
* Fixes #14034 
* Fixes #15080 
* Fixes #15090

This also opens a door for

* #10310
* #13191

Since we are changing vendored source (again), we might want to remove the `.rej`/`.orig` special rules in cargo-vendor, as well as look into the new source-dedup vendor dir layout.

### Summary Notes

- [benchmark-result](https://github.com/rust-lang/cargo/pull/15514#issuecomment-2870275766) by [weihanglo](https://github.com/weihanglo)

Generated by triagebot, see [help](https://forge.rust-lang.org/triagebot/note.html) for how to add more

### DIFF
--- a/src/cargo/ops/vendor.rs
+++ b/src/cargo/ops/vendor.rs
@@ -40,7 +40,7 @@ pub fn vendor(ws: &Workspace<'_>, opts: &VendorOptions<'_>) -> CargoResult<()> {
         extra_workspaces.push(ws);
     }
     let workspaces = extra_workspaces.iter().chain(Some(ws)).collect::<Vec<_>>();
-    let _lock = gctx.acquire_package_cache_lock(CacheLockMode::MutateExclusive)?;
+    let _lock = gctx.acquire_package_cache_lock(CacheLockMode::DownloadExclusive)?;
     let vendor_config = sync(gctx, &workspaces, opts).context("failed to sync")?;
 
     if gctx.shell().verbosity() != Verbosity::Quiet {
@@ -145,17 +145,10 @@ fn sync(
     let mut source_replacement_cache =
         SourceReplacementCache::new(gctx, opts.respect_source_config)?;
 
-    // First up attempt to work around rust-lang/cargo#5956. Apparently build
-    // artifacts sprout up in Cargo's global cache for whatever reason, although
-    // it's unsure what tool is causing these issues at this time. For now we
-    // apply a heavy-hammer approach which is to delete Cargo's unpacked version
-    // of each crate to start off with. After we do this we'll re-resolve and
-    // redownload again, which should trigger Cargo to re-extract all the
-    // crates.
-    //
-    // Note that errors are largely ignored here as this is a best-effort
-    // attempt. If anything fails here we basically just move on to the next
-    // crate to work with.
+    let mut checksums = HashMap::new();
+    let mut ids = BTreeMap::new();
+
+    // Let's download all crates and start storing internal tables about them.
     for ws in workspaces {
         let (packages, resolve) = ops::resolve_ws(ws, dry_run)
             .with_context(|| format!("failed to load lockfile for {}", ws.root().display()))?;
@@ -165,14 +158,11 @@ fn sync(
             .with_context(|| format!("failed to download packages for {}", ws.root().display()))?;
 
         for pkg in resolve.iter() {
-            let sid = if opts.respect_source_config {
-                source_replacement_cache.get(pkg.source_id())?
-            } else {
-                pkg.source_id()
-            };
+            let sid = source_replacement_cache.get(pkg.source_id())?;
 
-            // Don't delete actual source code!
+            // Don't vendor path crates since they're already in the repository
             if sid.is_path() {
+                // And don't delete actual source code!
                 if let Ok(path) = sid.url().to_file_path() {
                     if let Ok(path) = try_canonicalize(path) {
                         to_remove.remove(&path);
@@ -180,39 +170,7 @@ fn sync(
                 }
                 continue;
             }
-            if sid.is_git() {
-                continue;
-            }
 
-            // Only delete sources that are safe to delete, i.e. they are caches.
-            if sid.is_registry() {
-                if let Ok(pkg) = packages.get_one(pkg) {
-                    drop(fs::remove_dir_all(pkg.root()));
-                }
-                continue;
-            }
-        }
-    }
-
-    let mut checksums = HashMap::new();
-    let mut ids = BTreeMap::new();
-
-    // Next up let's actually download all crates and start storing internal
-    // tables about them.
-    for ws in workspaces {
-        let (packages, resolve) = ops::resolve_ws(ws, dry_run)
-            .with_context(|| format!("failed to load lockfile for {}", ws.root().display()))?;
-
-        packages
-            .get_many(resolve.iter())
-            .with_context(|| format!("failed to download packages for {}", ws.root().display()))?;
-
-        for pkg in resolve.iter() {
-            // No need to vendor path crates since they're already in the
-            // repository
-            if pkg.source_id().is_path() {
-                continue;
-            }
             ids.insert(
                 pkg,
                 packages
--- a/src/cargo/ops/vendor.rs
+++ b/src/cargo/ops/vendor.rs
@@ -647,9 +647,6 @@ fn vendor_this(relative: &Path) -> bool {
         // Temporary Cargo files
         Some(".cargo-ok") => false,
 
-        // Explicitly exclude .cargo_vcs_info.json to keep the old behavior.
-        Some(".cargo_vcs_info.json") => false,
-
         // Skip patch-style orig/rej files. Published crates on crates.io
         // have `Cargo.toml.orig` which we don't want to use here and
         // otherwise these are rarely used as part of the build process.
--- a/src/cargo/ops/vendor.rs
+++ b/src/cargo/ops/vendor.rs
@@ -368,26 +368,9 @@ fn cp_sources(
         let p = p.as_ref();
         let relative = p.strip_prefix(&src).unwrap();
 
-        match relative.to_str() {
-            // Skip git config files as they're not relevant to builds most of
-            // the time and if we respect them (e.g.  in git) then it'll
-            // probably mess with the checksums when a vendor dir is checked
-            // into someone else's source control
-            Some(".gitattributes" | ".gitignore" | ".git") => continue,
-
-            // Temporary Cargo files
-            Some(".cargo-ok") => continue,
-
-            // Skip patch-style orig/rej files. Published crates on crates.io
-            // have `Cargo.toml.orig` which we don't want to use here and
-            // otherwise these are rarely used as part of the build process.
-            Some(filename) => {
-                if filename.ends_with(".orig") || filename.ends_with(".rej") {
-                    continue;
-                }
-            }
-            _ => {}
-        };
+        if !vendor_this(relative) {
+            continue;
+        }
 
         // Join pathname components individually to make sure that the joined
         // path uses the correct directory separators everywhere, since
@@ -576,5 +559,27 @@ fn copy_and_checksum<T: Read>(
         cksum.update(data);
         dst.write_all(data)
             .with_context(|| format!("failed to write to {:?}", dst_path))?;
+    }
+}
+
+/// Filters files we want to vendor.
+///
+/// `relative` is a path relative to the package root.
+fn vendor_this(relative: &Path) -> bool {
+    match relative.to_str() {
+        // Skip git config files as they're not relevant to builds most of
+        // the time and if we respect them (e.g.  in git) then it'll
+        // probably mess with the checksums when a vendor dir is checked
+        // into someone else's source control
+        Some(".gitattributes" | ".gitignore" | ".git") => false,
+
+        // Temporary Cargo files
+        Some(".cargo-ok") => false,
+
+        // Skip patch-style orig/rej files. Published crates on crates.io
+        // have `Cargo.toml.orig` which we don't want to use here and
+        // otherwise these are rarely used as part of the build process.
+        Some(p) if p.ends_with(".orig") || p.ends_with(".rej") => false,
+        _ => true,
     }
 }
--- a/src/cargo/ops/vendor.rs
+++ b/src/cargo/ops/vendor.rs
@@ -4,13 +4,18 @@ use crate::core::{GitReference, Package, Workspace};
 use crate::ops;
 use crate::sources::path::PathSource;
 use crate::sources::PathEntry;
+use crate::sources::RegistrySource;
 use crate::sources::SourceConfigMap;
 use crate::sources::CRATES_IO_REGISTRY;
 use crate::util::cache_lock::CacheLockMode;
 use crate::util::{try_canonicalize, CargoResult, GlobalContext};
+
 use anyhow::{bail, Context as _};
 use cargo_util::{paths, Sha256};
+use cargo_util_schemas::core::SourceKind;
 use serde::Serialize;
+use walkdir::WalkDir;
+
 use std::collections::HashSet;
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::ffi::OsStr;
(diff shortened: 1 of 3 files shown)


epage:
Except this doesn't guarantee things are pristine until we vendor the contents of the `.crate`.  I worry this might cause us to vendor things like `target/` directories.

weihanglo:
> Except this doesn't guarantee things are pristine until we vendor the contents of the `.crate`. I worry this might cause us to vendor things like `target/` directories.

The chance of this happening is low because `cargo-vendor` holds a `MutateExclusive` cache lock during the entire call. Also basically everything is the same except using `walkdir` instead of `PathSource::list_files`. We can add a filter to walkdir to exclude `target/` directory, if that helps.

This patch aims to be as simple as possible. The “direct extraction” is the correct fix but require more architectural changes.

epage:
>     The chance of this happening is low because cargo-vendor holds a MutateExclusive cache lock during the entire call. Also basically everything is the same except using walkdir instead of PathSource::list_files. We can add a filter to walkdir to exclude target/ directory, if that helps.

And that we delete the extracted `.crate` so we get a fresh extraction.

So looking over this again, this does not fix #15090.  This PR only changes which files we list, it doesn't change how we copy them which is what is causing the problem in #15090.

>  The “direct extraction” is the correct fix but require more architectural changes.

Is it all that big of an architectural change?  Can we get an instance of the registry source and and call an extract method directly on it?

weihanglo:
> So looking over this again, this does not fix https://github.com/rust-lang/cargo/issues/15090. This PR only changes which files we list, it doesn't change how we copy them which is what is causing the problem in https://github.com/rust-lang/cargo/issues/15090.

How does it not fix #15090? cargo-vendor doesn't change any content of a registry package, at least `cp_sources()` doesn't. It does skip some files and create a `.cargo-checksum.json` though.

> Is it all that big of an architectural change? Can we get an instance of the registry source and and call an extract method directly on it?

Maybe not really an architectural change, but we'll need to have an abstraction over local/remote/http registry (which is RegistryData/RegistrySource). Ensure `.crates` be prepared. And decouple `unpack_package` and `get_pkg` from the actual download logic.

weihanglo:
Anyway, I got direct extraction working. Doesn't look too terrible than I thought!

weihanglo:
@rustbot note benchmark-result

Also, the benchmark result shows that it is ~60% faster when vendoring cargo and its dependencies. On Apple M1 Pro with 500GiB APFS SSD

```
Benchmark 1: cargo-15514 vendor
  Time (mean ± σ):     12.331 s ±  0.221 s    [User: 2.705 s, System: 7.211 s]
  Range (min … max):   12.083 s … 12.753 s    10 runs

Benchmark 2: cargo-15514 vendor --versioned-dirs
  Time (mean ± σ):     12.737 s ±  0.242 s    [User: 2.704 s, System: 7.681 s]
  Range (min … max):   12.481 s … 13.240 s    10 runs

Benchmark 3: cargor-master vendor
  Time (mean ± σ):     19.639 s ±  0.475 s    [User: 2.947 s, System: 12.391 s]
  Range (min … max):   19.006 s … 20.481 s    10 runs

Benchmark 4: cargor-master vendor --versioned-dirs
  Time (mean ± σ):     19.574 s ±  0.558 s    [User: 2.948 s, System: 12.422 s]
  Range (min … max):   18.739 s … 20.489 s    10 runs

Summary
  cargofast vendor ran
    1.03 ± 0.03 times faster than cargo-15514 vendor --versioned-dirs
    1.59 ± 0.05 times faster than cargo-master vendor --versioned-dirs
    1.59 ± 0.05 times faster than cargo-master vendor
```

ehuss (approved):
Nice!

I have not gone over this with a fine-toothed comb, but the overall structure looks good to me.

I'm not going to click merge right now in case anyone else wants to look closer, or you had any additional concerns or changes you were thinking of. If there isn't more progress by sometime next week, then I think it would be fine to just merge.

weihanglo:
@ehuss just check in and see if we can merge this. I'd also like to do a follow-up for <https://github.com/rust-lang/cargo/issues/13191>.

src/cargo/ops/vendor.rs
epage, line 1: Can we also move away from the mutually exclusive package cache lock?

weihanglo: I think yes. It should be downgraded to DownloadExclusive at least.

weihanglo: Fixed in b2fbd65a13820296bbb29142a6e5fa86cfc27552

weihanglo: It turns out that I chose [the wrong lockmode](https://github.com/rust-lang/cargo/pull/15514/commits/9eaa5a3c9dc65a08ea92ed3fbff975e6aaec391f#diff-4cf00ca0e6ab96b4a634ed67084038c409058a090534376b495bbde41ac52861R677). Thanks Ed for helping me discovery the bug.
Really fixed in the latest push now 3b89abb511277733c59f8b226b4ad2acebdf33dd

src/cargo/ops/vendor.rs
```diff
@@ -35,7 +40,7 @@ pub fn vendor(ws: &Workspace<'_>, opts: &VendorOptions<'_>) -> CargoResult<()> {
         extra_workspaces.push(ws);
     }
     let workspaces = extra_workspaces.iter().chain(Some(ws)).collect::<Vec<_>>();
-    let _lock = gctx.acquire_package_cache_lock(CacheLockMode::MutateExclusive)?;
+    let _lock = gctx.acquire_package_cache_lock(CacheLockMode::DownloadExclusive)?;
```
epage, line 42: Technically, this change belongs in 9fb4dbcb9da1e85278960474f5f6a914e11ec501 ("fix(vendor): remove workaround") since this commit is still making edits to the cache

weihanglo: Nice catch. Updated.

epage (approved):

epage:
This is early enough in the release that I'm going to go ahead and merge this now.

src/cargo/ops/vendor.rs
```diff
@@ -578,3 +561,25 @@ fn copy_and_checksum<T: Read>(
             .with_context(|| format!("failed to write to {:?}", dst_path))?;
     }
 }
+
+/// Filters files we want to vendor.
+///
+/// `relative` is a path relative to the package root.
+fn vendor_this(relative: &Path) -> bool {
```
epage, line 610: Should we filter out `.cargo_vcs_info.json` and explicitly bring it back in?

Also, we're closing #11000 without having a test for that case

weihanglo: Got it rearranged.

epage (approved):

PR #15522, opened 2025-05-13
fix: remove unnecessary workaround in standard_lib test

motorailgun:

 - As  issue https://github.com/rust-lang/rust/issues/125246 has already been fixed, there must be no need for commenting out `--sysroot` anymore in the tests.

 - The bug mentioned in `shared_std_dependency_rebuild()` looks to be already solved and does not reproduce. So I guess it's safe to un-comment those lines.

### DIFF
--- a/tests/testsuite/standard_lib.rs
+++ b/tests/testsuite/standard_lib.rs
@@ -87,7 +87,7 @@ fn setup() -> Setup {
     let p = ProjectBuilder::new(paths::root().join("rustc-wrapper"))
         .file(
             "src/main.rs",
-            r#"
+            &r#"
                 use std::process::Command;
                 use std::env;
                 fn main() {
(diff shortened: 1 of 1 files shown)


motorailgun:
Oops, fixing tests...

motorailgun:
I must admit that I'm kind of stuck here. The original issue, where ld needed to be present inside sysroot on `x86_64-unknown-linux-gnu`, must be solved and looks to be not happening anymore. However, for the test `standard_lib::shared_std_dependency_rebuild`, I could not find the cause of the E0460, or judge if this is an intentional behavior.

@weihanglo Would you mind looking into the failing test if possible...?

tests/testsuite/standard_lib.rs
```diff
@@ -104,14 +104,8 @@ fn setup() -> Setup {
...
                         // for more information on this.
-                        //
-                        // FIXME: this is broken on x86_64-unknown-linux-gnu
-                        // due to https://github.com/rust-lang/rust/pull/124129,
-                        // because it requires lld in the sysroot. See
-                        // https://github.com/rust-lang/rust/issues/125246 for
-                        // more information.
-                        // args.push("--sysroot".to_string());
-                        // args.push("/path/to/nowhere".to_string());
+                        args.push("--sysroot".to_string());
```
weihanglo, line 107: Because the hack predates the test `build_std_with_no_arg_for_core_only_target`, which really requires a target shared library from sysroot, for example `~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/aarch64-unknown-none/lib/`

weihanglo: One way to fix this is also ensure that the `--target` passed in is rustc host target.

motorailgun: Ah I see... well, ensuring that passed `--target` is the host target seems a good way to go.

motorailgun: (By the way, thank you for your advice)

weihanglo (approved):
Thanks!

PR #15525, opened 2025-05-14
feat: skip `publish=false` pkg when publishing entire workspace

### LINKED ISSUE #15006: `cargo publish` should optionally skip unpublishable packages
This is about the unstable `-Zpackage-workspace` feature. Discussion branched from https://github.com/rust-lang/cargo/issues/1169#issuecomment-2568067086.

### Problem

In a large workspace, there might be many packages to publish, and many packages that should not be published (for development only, or applications that are distributed by means other than Cargo). `cargo publish -Zpackage-workspace` will attempt to publish the workspace’s default members, whether or not they are marked unpublishable with `package.publish = false`, and error out if any of them are.

This seems to me a missing feature; it should be possible to choose to use `package.publish` metadata to control whether a package is selected for publishing.

### Proposed Solution

* Alternative 1: Add a new flag (with what name?) which selects packages, from members or default-members, as follows:

  * If a package has `package.publish = false`, it is not selected.
  * If a package has `package.publish = true`, it is selected.
  * If neither of the above applies, error out and do not publish anything.

  This flag would be compatible with `--workspace` but incompatible with `--package`.

  This version encourages workspace authors to explicitly think about whether each package should be published if they wish to use this feature, which seems to me the best outcome. However, it gives new semantics to `package.publish = true`, which was previously identical to the field being absent. I think this is a good thing (publishing should be more opt-in, since it is close to irrevocable), but it would be new.

* Alternative 2: Add a new flag `--exclude-unpublishable`.
  * if a package was selected due to being in `default-members` or due to `--workspace` being passed, and it has `package.publish = false`, then it is excluded from selection as if by `--exclude`.
  * If a package was selected by `--package` and it has `package.publish = false`, this is still an error.

* Alternative 3: Automatically skip `publish = false` packages without any specific flag.

  This might be too “DWIM” but it is the behavior I was implicitly expecting when I tried out `-Zpackage-workspace` for the first time.

### Notes

In the status quo, it would be possible to use the workspace’s `default-members` to control which packages are published. I think this is not sufficient, because `default-members` should be a package selection which is appropriate for *routine development,* which will usually be much more common than publishing. In particular, a package whose job is entirely tests should not be published, but should be run by default with `cargo test`.

@rustbot label +Z-package-workspace

weihanglo:

Fixes #15006

This changes how cargo-publish works with the unstable
`-Zpackage-workspace` feature.

Before this, when publishing the entire workspace,
like `cargo publish --workspace`,
if there is a package with `package.pulibsh=false,
it'll fail the entire publish process.

After this, when `--workspace` is passed,
or when publishing the virtual workspace,
the intent is more like “publish all publishable in this workspace”,
so skip `publish=false` packages and proceed to publish others.

The new overall behavior looks like this:

* `cargo publish` (inside a `package.publish = false` package): error
* `cargo publish -p publishable -p unpublishable`: error
* `cargo publish --workspace`: skips `package.publish = false

See https://github.com/rust-lang/cargo/issues/15006#issuecomment-2847660911

* `workspace_flag_with_unpublishable_packages` was added to ensure `--workspace` work with non-virtual workspace.
* `unpublishable_package_as_versioned_dev_dep` was added to ensure versioned dev-dependencies won't be skipped and still fail cargo-publish, as they are required to be published.

There is a new scenario that nothing is going to publish.
I went with a warning instead of hard error because missing publish for the entire worspace should be a fairly visible. However, this is open for future to configure via Cargo lint system.

### DIFF
--- a/src/cargo/ops/registry/publish.rs
+++ b/src/cargo/ops/registry/publish.rs
@@ -117,6 +117,22 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
         );
     }
 
+    if pkgs.is_empty() {
+        if allow_unpublishable {
+            let n = unpublishable.len();
+            let plural = if n == 1 { "" } else { "s" };
+            ws.gctx().shell().warn(format_args!(
+                "nothing to publish, but found {n} unpublishable package{plural}"
+            ))?;
+            ws.gctx().shell().note(format_args!(
+                "to publish packages, set `package.publish` to `true` or a non-empty list"
+            ))?;
+            return Ok(());
+        } else {
+            unreachable!("must have at least one publishable package");
+        }
+    }
+
     let just_pkgs: Vec<_> = pkgs.iter().map(|p| p.0).collect();
     let reg_or_index = match opts.reg_or_index.clone() {
         Some(r) => {
--- a/src/cargo/ops/registry/publish.rs
+++ b/src/cargo/ops/registry/publish.rs
@@ -93,6 +93,20 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
         .filter(|(m, _)| specs.iter().any(|spec| spec.matches(m.package_id())))
         .collect();
 
+    let (unpublishable, pkgs): (Vec<_>, Vec<_>) = pkgs
+        .into_iter()
+        .partition(|(pkg, _)| pkg.publish() == &Some(vec![]));
+    if !unpublishable.is_empty() {
+        bail!(
+            "{} cannot be published.\n\
+            `package.publish` must be set to `true` or a non-empty list in Cargo.toml to publish.",
+            unpublishable
+                .iter()
+                .map(|(pkg, _)| format!("`{}`", pkg.name()))
+                .join(", "),
+        );
+    }
+
     let just_pkgs: Vec<_> = pkgs.iter().map(|p| p.0).collect();
     let reg_or_index = match opts.reg_or_index.clone() {
         Some(r) => {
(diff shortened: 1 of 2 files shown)


tests/testsuite/publish.rs
```diff
@@ -4103,10 +4103,183 @@ fn multiple_unpublishable_package() {
 
     p.cargo("publish -Zpackage-workspace")
         .masquerade_as_nightly_cargo(&["package-workspace"])
+        .replace_crates_io(registry.index_url())
+        .with_stderr_data(str![[r#"
+[WARNING] nothing to publish, but found 2 unpublishable packages
```
epage, line 4108: This test isn't serving is no longer serving its purpose. We should add a publishable package to it and add a new test for "all are unpublishable".

epage, line 4108: > There is a new scenario that nothing is going to publish.
I went with a warning instead of hard error because missing publish for the entire worspace should be a fairly visible. However, this is open for future to configure via Cargo lint system.

What happens if I do `cargo check --bins` and none are built due to `required-features`?

src/cargo/ops/registry/publish.rs
```diff
@@ -705,19 +745,6 @@ fn package_list(pkgs: impl IntoIterator<Item = PackageId>, final_sep: &str) -> S
 }
 
 fn validate_registry(pkgs: &[&Package], reg_or_index: Option<&RegistryOrIndex>) -> CargoResult<()> {
-    let unpublishable = pkgs
```
epage, line 708: Small nit, feel free to pass on. I'm mentioning this only since I had other comments.

Would make reviewing easier if ĵust moving this error check was a separate refactor commit.

epage:
Is there a reason this isn't marked as fixing #15006?

tests/testsuite/publish.rs
weihanglo:
```
warning: target filter `bins` specified, but no targets matched; this is a no-op
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.03s
```

weihanglo:
All review comments were addressed.

> Is there a reason this isn't marked as fixing https://github.com/rust-lang/cargo/issues/15006?

I forgot 😬.

epage (approved):

epage:
Thanks!

PR #15533, opened 2025-05-16
Add `cfg_version` support

### LINKED ISSUE #15531: Tracking Issue for _cfg_version_
### Summary

RFC: [#2523](https://github.com/rust-lang/rfcs/pull/2523) 
Original issue: N/A
Implementation:
- [ ] Initial implementation
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#cfg-version

Allow dependencies conditioned on the Rust version in a way that matches with `#[cfg(version)]`
```toml
[target.'cfg(not(version(1.70.0))`.dependencies]
is-terminal = "0.4.2"
```

See also 
- rust-lang/rust#64796
- https://github.com/rust-lang/rfcs/blob/master/text/2523-cfg-path-version.md

### Unresolved Issues

_No response_

### Future Extensions

_No response_

### About tracking issues

mo8it:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/pull/15533)*

#15531

### DIFF
--- a/crates/cargo-platform/src/cfg.rs
+++ b/crates/cargo-platform/src/cfg.rs
@@ -130,24 +130,24 @@ impl fmt::Display for Cfg {
 
 impl CfgExpr {
     /// Utility function to check if the key, "cfg(..)" matches the `target_cfg`
-    pub fn matches_key(key: &str, target_cfg: &[Cfg]) -> bool {
+    pub fn matches_key(key: &str, target_cfg: &[Cfg], rustc_version: &semver::Version) -> bool {
         if key.starts_with("cfg(") && key.ends_with(')') {
             let cfg = &key[4..key.len() - 1];
 
             CfgExpr::from_str(cfg)
                 .ok()
-                .map(|ce| ce.matches(target_cfg))
+                .map(|ce| ce.matches(target_cfg, rustc_version))
                 .unwrap_or(false)
         } else {
             false
         }
     }
 
-    pub fn matches(&self, cfg: &[Cfg]) -> bool {
+    pub fn matches(&self, cfg: &[Cfg], rustc_version: &semver::Version) -> bool {
         match *self {
-            CfgExpr::Not(ref e) => !e.matches(cfg),
-            CfgExpr::All(ref e) => e.iter().all(|e| e.matches(cfg)),
-            CfgExpr::Any(ref e) => e.iter().any(|e| e.matches(cfg)),
+            CfgExpr::Not(ref e) => !e.matches(cfg, rustc_version),
+            CfgExpr::All(ref e) => e.iter().all(|e| e.matches(cfg, rustc_version)),
+            CfgExpr::Any(ref e) => e.iter().any(|e| e.matches(cfg, rustc_version)),
             CfgExpr::Value(ref e) => cfg.contains(e),
             CfgExpr::True => true,
             CfgExpr::False => false,
--- a/crates/cargo-platform/src/cfg.rs
+++ b/crates/cargo-platform/src/cfg.rs
@@ -1,4 +1,5 @@
 use crate::error::{ParseError, ParseErrorKind::*};
+use std::cmp::Ordering;
 use std::fmt;
 use std::hash::{Hash, Hasher};
 use std::iter;
@@ -15,6 +16,50 @@ pub enum CfgExpr {
     False,
 }
 
+// Major version restricted to `1`.
+#[derive(Eq, PartialEq, Hash, Ord, PartialOrd, Clone, Debug)]
+pub struct CfgRustVersion {
+    pub minor: u64,
+    pub patch: Option<u64>,
+}
+
+impl CfgRustVersion {
+    pub fn parse(version: &str) -> Option<Self> {
+        let minor_patch = version.strip_prefix("1.")?;
+        let (minor, patch) = match minor_patch.split_once('.') {
+            Some((minor, patch)) => (minor.parse().ok()?, Some(patch.parse().ok()?)),
+            None => (minor_patch.parse().ok()?, None),
+        };
+        Some(Self { minor, patch })
+    }
+
+    pub fn matches(&self, rustc_version: &semver::Version) -> bool {
+        match self.minor.cmp(&rustc_version.minor) {
+            Ordering::Less => true,
+            Ordering::Equal => match self.patch {
+                Some(patch) => {
+                    if rustc_version.pre.as_str() == "nightly" {
+                        false
+                    } else {
+                        patch <= rustc_version.patch
+                    }
+                }
+                None => true,
+            },
+            Ordering::Greater => false,
+        }
+    }
+}
+
+impl fmt::Display for CfgRustVersion {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.patch {
+            Some(patch) => write!(f, "version(\"1.{}.{patch}\")", self.minor),
+            None => write!(f, "version(\"1.{}\")", self.minor),
+        }
+    }
+}
+
 /// A cfg value.
 #[derive(Eq, PartialEq, Hash, Ord, PartialOrd, Clone, Debug)]
 pub enum Cfg {
@@ -22,6 +67,8 @@ pub enum Cfg {
     Name(Ident),
     /// A key/value cfg pair, like `target_os = "linux"`.
     KeyPair(Ident, String),
+    /// A Rust version cfg value, like `version("1.23.4")` or `version("1.23")`.
+    Version(CfgRustVersion),
 }
 
 /// A identifier
@@ -124,6 +171,7 @@ impl fmt::Display for Cfg {
         match *self {
             Cfg::Name(ref s) => s.fmt(f),
             Cfg::KeyPair(ref k, ref v) => write!(f, "{} = \"{}\"", k, v),
+            Cfg::Version(ref cfg_rust_version) => cfg_rust_version.fmt(f),
         }
     }
 }
@@ -148,6 +196,9 @@ impl CfgExpr {
             CfgExpr::Not(ref e) => !e.matches(cfg, rustc_version),
             CfgExpr::All(ref e) => e.iter().all(|e| e.matches(cfg, rustc_version)),
             CfgExpr::Any(ref e) => e.iter().any(|e| e.matches(cfg, rustc_version)),
+            CfgExpr::Value(Cfg::Version(ref cfg_rust_version)) => {
+                cfg_rust_version.matches(rustc_version)
+            }
             CfgExpr::Value(ref e) => cfg.contains(e),
             CfgExpr::True => true,
             CfgExpr::False => false,
@@ -275,6 +326,25 @@ impl<'a> Parser<'a> {
                         },
                         val.to_string(),
                     )
+                } else if name == "version" {
+                    self.eat(&Token::LeftParen)?;
+                    let token = self
+                        .t
+                        .next()
+                        .ok_or_else(|| ParseError::new(self.t.orig, InvalidVersion))??;
+                    let Token::String(version_str) = token else {
+                        return Err(ParseError::new(
+                            self.t.orig,
+                            UnexpectedToken {
+                                expected: "a string",
+                                found: token.classify(),
+                            },
+                        ));
+                    };
+                    self.eat(&Token::RightParen)?;
+                    let version = CfgRustVersion::parse(version_str)
+                        .ok_or_else(|| ParseError::new(self.t.orig, InvalidVersion))?;
+                    Cfg::Version(version)
                 } else {
                     Cfg::Name(Ident {
                         name: name.to_string(),
--- a/crates/cargo-platform/src/lib.rs
+++ b/crates/cargo-platform/src/lib.rs
@@ -34,10 +34,10 @@ impl Platform {
     /// Returns whether the Platform matches the given target and cfg.
     ///
     /// The named target and cfg values should be obtained from `rustc`.
-    pub fn matches(&self, name: &str, cfg: &[Cfg]) -> bool {
+    pub fn matches(&self, name: &str, cfg: &[Cfg], rustc_version: &semver::Version) -> bool {
         match *self {
             Platform::Name(ref p) => p == name,
-            Platform::Cfg(ref p) => p.matches(cfg),
+            Platform::Cfg(ref p) => p.matches(cfg, rustc_version),
         }
     }
 
--- a/crates/cargo-platform/src/lib.rs
+++ b/crates/cargo-platform/src/lib.rs
@@ -18,7 +18,7 @@ mod cfg;
 mod error;
 
 use cfg::KEYWORDS;
-pub use cfg::{Cfg, CfgExpr, Ident};
+pub use cfg::{Cfg, CfgExpr, CfgRustVersion, Ident};
 pub use error::{ParseError, ParseErrorKind};
 
 /// Platform definition.
@@ -97,6 +97,7 @@ impl Platform {
                              https://doc.rust-lang.org/cargo/reference/features.html"
                         ))
                     },
+                    Cfg::Version(..) => {},
                 }
                 CfgExpr::True | CfgExpr::False => {},
             }
@@ -130,6 +131,7 @@ impl Platform {
                             ));
                         }
                     }
+                    Cfg::Version(..) => {}
                 },
             }
         }
--- a/crates/cargo-platform/tests/test_cfg.rs
+++ b/crates/cargo-platform/tests/test_cfg.rs
@@ -1,7 +1,8 @@
 use std::fmt;
 use std::str::FromStr;
 
-use cargo_platform::{Cfg, CfgExpr, Ident, Platform};
+use cargo_platform::{Cfg, CfgExpr, CfgRustVersion, Ident, Platform};
+use semver::{BuildMetadata, Prerelease};
 use snapbox::assert_data_eq;
 use snapbox::prelude::*;
 use snapbox::str;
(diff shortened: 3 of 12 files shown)


mo8it:
@epage I will of course take care of proper error handling, but I just want to get feedback on the initial direction.

`cfg(nightly)` can be already parsed as a `Name`.

weihanglo (commented):
I just tweaked the PR description a bit. We usually keep a tracking issue open until stabilization.

src/cargo/core/compiler/custom_build.rs
```diff
@@ -416,6 +416,7 @@ fn build_work(build_runner: &mut BuildRunner<'_, '_>, unit: &Unit) -> CargoResul
                 let values = cfg_map.entry(k.as_str()).or_default();
                 values.push(v.as_str());
             }
+            Cfg::Version(..) => todo!(),
```
mo8it, line 419: No idea what to do here

crates/cargo-platform/src/cfg.rs
```diff
@@ -235,6 +250,25 @@ impl<'a> Parser<'a> {
                 self.eat(&Token::RightParen)?;
                 Ok(CfgExpr::Not(Box::new(e)))
             }
+            Some(Ok(Token::Ident(false, "version"))) => {
+                self.t.next();
+                self.eat(&Token::LeftParen)?;
```
epage, line 255: Can you have `--cfg version=hello` and so need to support `#[cfg(version)1.30.0)]` and `#[cfg(version = "hello")]`?

crates/cargo-platform/src/cfg.rs
```diff
@@ -398,6 +432,43 @@ impl<'a> Iterator for Tokenizer<'a> {
...
                 }
+                // Reject a major version starting with 0
+                Some((start, '1'..='9')) => {
+                    while let Some(&(end, ch)) = self.s.peek() {
+                        if ch == ')' {
+                            let mut iter = self.orig[start..end].split('.');
+                            let major = iter.next().unwrap().parse().unwrap();
+                            if let Some(minor) = iter.next().map(|s| s.parse().unwrap()) {
+                                if let Some(patch) = iter.next().map(|s| s.parse().unwrap()) {
+                                    if iter.next().is_some() {
```
epage, line 443: This is some fairly deep nesting.  Is there an alternative?

src/cargo/core/compiler/custom_build.rs
epage: I think the following code will be relevant:
https://github.com/rust-lang/cargo/blob/47c911e9e6f6461f90ce19142031fe16876a3b95/src/cargo/ops/cargo_compile/mod.rs#L495-L511

crates/cargo-platform/src/cfg.rs
epage, line 1: As a stylistic thing, as you add tests, we prefer tests to be added in a commit before the change, with them passing.  Your change then includes updating the tests to make sure they pass.  How to apply this is dependent on the PR (and sometimes it just doesn't work).  In this case, I'd write the tests without `#[cfg(version)]` and then add it in the follow ups.

src/cargo/core/compiler/custom_build.rs
epage, line 419: Should we have this behind an unstable gate or insta-stabilize once the stabilization report for the rustc side is done?

epage:
Be sure to check out rust-lang/rust#141137 for a summary of what has changed from the RFC

crates/cargo-platform/src/cfg.rs
mo8it: Looking at the updated RFC, I assume the answer is no?

src/cargo/core/compiler/custom_build.rs
mo8it: The code you point to wasn't helpful. I still don't know in what format I should write the version cfg as an env var.

ehuss: Just to be clear here, we do not expect rustc to emit this in `--print`, correct? That means this code path should never be reached, right?

And, similar to the point made at https://github.com/rust-lang/rust/pull/141137#discussion_r2096067981, it doesn't really make sense to include this here since it would need to emit *every* version to be correct. My inclination is that the env var shouldn't be set for build scripts.

epage: Ah, yes, I missed what this code was doing. I was giving code to help when evaluating dependencies.  I agree with Eric that we shouldn't do anything here.  Maybe worth a `trace!`

crates/cargo-platform/src/cfg.rs
epage: I thought the stabilization report said that `--cfg version=hello` would still work

> #[cfg(version = "1.20.0")] is not supported, and there is a warning of the unexpected_cfgs lint (but no compilation error)

(not supported within the context of this feature, `unexpected_cfgs` firing because `--check-cfg` wasn't specified for `version`)

> #[cfg(version = "1.20.0")] acts as if the cfg was false due to the wrong syntax, even if the compiler version is above the specified version

(again, framed within this feature; it could be true if `--cfg version="1.20.0"` was specified)

mo8it: I pushed a change so that `version = …` is ignored by the resolver now. I also raise a warning on it.

In what context could `--cfg version="1.23.4"` be used? I never specified a `cfg` flag. Where is that parsed? How should I handle it?

crates/cargo-platform/src/cfg.rs
```diff
@@ -124,30 +145,40 @@ impl fmt::Display for Cfg {
...
+            CfgExpr::Not(ref e) => !e.matches(cfg, rustc_version),
+            CfgExpr::All(ref e) => e.iter().all(|e| e.matches(cfg, rustc_version)),
+            CfgExpr::Any(ref e) => e.iter().any(|e| e.matches(cfg, rustc_version)),
+            CfgExpr::Value(Cfg::Version(CfgRustVersion { minor, patch })) => {
+                match minor.cmp(&rustc_version.minor) {
+                    Ordering::Less => true,
+                    Ordering::Equal => patch <= rustc_version.patch,
+                    Ordering::Greater => false,
+                }
+            }
```
epage, line 181: I believe the tracking issue mentions specific behavior for nightlies
- `1.80` includes 1.80 nightlies
- `1.80.0` does not include 1.80 nightlies

crates/cargo-platform/src/cfg.rs
```diff
@@ -124,30 +145,40 @@ impl fmt::Display for Cfg {
         match *self {
             Cfg::Name(ref s) => s.fmt(f),
             Cfg::KeyPair(ref k, ref v) => write!(f, "{} = \"{}\"", k, v),
+            Cfg::Version(CfgRustVersion { minor, patch }) => {
+                write!(f, "version(\"1.{minor}.{patch}\")")
```
epage, line 149: Can we delegate the formatting of `CfgRustVersion` to `CfgRustVersion`?

crates/cargo-platform/src/cfg.rs
epage, line 181: Like we did with `package.rust-version`, could we delegate the matching logic to `CfgRustVersion`?

crates/cargo-platform/src/lib.rs
```diff
@@ -88,15 +88,26 @@ impl Platform {
...
+                                 https://doc.rust-lang.org/cargo/reference/features.html"
+                            ));
+                        },
+                        "version" => {
+                            warnings.push(String::from(
+                                "Found `version = ...` in `target.'cfg(...)'.dependencies`. \
+                                This format is ignored by Cargo. \
+                                Use the format `version(\"1.23.4\")` or `version(\"1.23\")` instead."
+                            ));
+                        },
```
epage, line 107: Based on https://github.com/rust-lang/cargo/pull/15533#discussion_r2093685172 I don't think this should be present.

crates/cargo-platform/src/cfg.rs
epage: > In what context could --cfg version="1.23.4" be used? I never specified a cfg flag. Where is that parsed? How should I handle it?

More so the point is that this change should be agnostic to `--cfg version="1.23.4"` and should not change the behavior of `cfg(version = "1.23.4")`.

crates/cargo-platform/src/cfg.rs
ehuss: I'm not quite following this. Can you say more about where that is being discussed? That doesn't seem to be the current behavior.

crates/cargo-platform/tests/test_cfg.rs
```diff
@@ -152,33 +172,34 @@ fn cfg_expr_bad() {
 
 #[test]
 fn cfg_matches() {
```
epage, line 174: Can we add tests for `version`?

tests/testsuite/cfg.rs
```diff
@@ -817,3 +817,145 @@ fn cfg_booleans_rustflags_no_effect() {
...
+                [target.'cfg(version("1.87.0"))'.dependencies]
+                b = { path = 'b' }
+            "#,
+        )
+        .file("src/lib.rs", "extern crate b;")
+        .file("b/Cargo.toml", &basic_manifest("b", "0.0.1"))
+        .file("b/src/lib.rs", "")
+        .build();
+    p.cargo("check -v").run();
+}
```
epage, line 840: I wish we had a way to override the rust version so we could test with `1.2345` (the version we always use for "far future" in tests)

crates/cargo-platform/src/cfg.rs
epage: Maybe I'm misinterpeting what was said in

> Ultimately, a compromise was agreed upon, in which nightly releases are treated as "complete" i.e. cfg(version(1.20)) evals to true on nightly builds with version 1.20.0, but there is a nightly flag -Z assume-incomplete-release to opt into the behaviour that doesn't do this assumption. This compromise was implemented in PR https://github.com/rust-lang/rust/pull/81468.

The actual rustc tests should be checked to confirm.  We probably should most or all of their tests.

crates/cargo-platform/src/cfg.rs
```diff
@@ -15,13 +16,33 @@ pub enum CfgExpr {
...
+            Some((minor, patch)) => (minor.parse().ok()?, patch.parse().ok()?),
+            None => (minor_patch.parse().ok()?, 0),
+        };
+        Some(Self { minor, patch })
+    }
+}
+
 /// A cfg value.
 #[derive(Eq, PartialEq, Hash, Ord, PartialOrd, Clone, Debug)]
 pub enum Cfg {
```
weihanglo, line 39: Should we make it non-exhaustive so we don't need to bump version in the future?

epage: Keeping it this way makes it easier to find places to update, so unless we have a strong reason for avoiding major bumps, I'd lean towards this (similar with `cargo-util-schemas`)

crates/cargo-platform/tests/test_cfg.rs
```diff
@@ -152,33 +172,34 @@ fn cfg_expr_bad() {
...
-    assert!(e!(all(foo, bar)).matches(&[c!(foo), c!(bar)]));
-    assert!(e!(not(foo)).matches(&[c!(bar)]));
-    assert!(e!(not(foo)).matches(&[]));
-    assert!(e!(any((not(foo)), (all(foo, bar)))).matches(&[c!(bar)]));
-    assert!(e!(any((not(foo)), (all(foo, bar)))).matches(&[c!(foo), c!(bar)]));
-    assert!(e!(foo).matches(&[c!(r # foo)]));
-    assert!(e!(r # foo).matches(&[c!(foo)]));
-    assert!(e!(r # foo).matches(&[c!(r # foo)]));
+    let rustc_version = semver::Version::new(0, 0, 0);
+    assert!(e!(foo).matches(&[c!(bar), c!(foo), c!(baz)], &rustc_version));
```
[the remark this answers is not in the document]
mo8it: Rewriting the Git history would be too tricky for me. It is my first real contribution to Cargo. I am learning to adapt the style and will apply it in future commits :)
Thanks for the feedback, much appreciated :D

crates/cargo-platform/src/cfg.rs
mo8it: Did I get the desired logic here?

https://github.com/rust-lang/cargo/blob/120bea75161530b7e4d3c26c1d5da32c95d740e3/crates/cargo-platform/tests/test_cfg.rs#L224-L234

https://github.com/rust-lang/cargo/blob/120bea75161530b7e4d3c26c1d5da32c95d740e3/crates/cargo-platform/src/cfg.rs#L36-L52

crates/cargo-platform/tests/test_cfg.rs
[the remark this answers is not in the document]
epage: I'm prepping a rewrite of the history for you which also includes switching `cargo-platform` over to snapshot testing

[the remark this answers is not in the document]
mo8it: All edits are welcome <3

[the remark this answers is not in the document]
epage: My commits are pushed.  Hope I didn't lose anything

mo8it:
@epage Thanks for rewriting the history. I found it a bit disappointing that I wasn't listed as an author anymore. Therefore, I updated the author information of the commits and hope that you are happy with the current state.

I also fixed the CI by removing `dbg!(input)`.

Finally, I resolved the merge conflicts.

Learning `jj` today helped a lot :D

If we agree on https://github.com/rust-lang/cargo/pull/15533#discussion_r2098745262, then the PR should be ready for merging.

epage:
> I found it a bit disappointing that I wasn't listed as an author anymore.

Sorry that I forgot to preserve that!

tests/testsuite/cfg.rs
epage: Could we use `RUSTC_FORCE_RUSTC_VERSION`?

tests/testsuite/cfg.rs
```diff
@@ -817,3 +817,144 @@ fn cfg_booleans_rustflags_no_effect() {
...
+[LOCKING] 1 package to latest compatible version
+[CHECKING] foo v0.0.0 ([ROOT]/foo)
+error[E0463]: can't find crate for `b`
+ --> src/lib.rs:1:1
+  |
+1 | extern crate b;
+  | ^^^^^^^^^^^^^^^ can't find crate
+
+For more information about this error, try `rustc --explain E0463`.
+[ERROR] could not compile `foo` (lib) due to 1 previous error
```
ehuss, line 956: We can't match on the exact output from rustc because if they change the output it would prevent this test from passing. I think something like the following should be close enough:

```suggestion
error[E0463]: can't find crate for `b`
...
[ERROR] could not compile `foo` (lib) due to 1 previous error
```

ehuss (changes requested):
Can you include some tests as mentioned from https://github.com/rust-lang/cargo/pull/15533#issuecomment-2895074734? Perhaps something like:

```rust
#[cargo_test]
fn cfg_version_from_index() {
    // Checks that cfg(version) works in the index.
    Package::new("only_newer", "1.0.0").publish();
    Package::new("bar", "1.0.0")
        .file("src/lib.rs", "extern crate only_newer;")
        .target_dep("only_newer", "1.0", "cfg(version(\"1.87\"))")
        .publish();

    let p = project()
        .file(
            "Cargo.toml",
            r#"
                [package]
                name = "foo"
                edition = "2024"

                [dependencies]
                bar = "1.0"
            "#,
        )
        .file("src/lib.rs", "")
        .build();

    p.cargo("check")
        .with_stderr_data(str![[r#"
[LOCKING] 2 packages to latest Rust [..] compatible versions
[DOWNLOADING] crates ...
[DOWNLOADED] only_newer v1.0.0 (registry `dummy-registry`)
[DOWNLOADED] bar v1.0.0 (registry `dummy-registry`)
[CHECKING] only_newer v1.0.0
[CHECKING] bar v1.0.0
[CHECKING] foo v0.0.0 ([ROOT]/foo)

"#]])
        .run();

    p.cargo("tree")
        .with_stdout_data(str![[r#"
foo v0.0.0 ([ROOT]/foo)
└── bar v1.0.0
    └── only_newer v1.0.0

"#]])
        .run();
}

#[cargo_test]
fn cfg_version_from_index_in_future() {
    // Checks that cfg(version) for a version from the future is ignored.
    Package::new("only_newer", "1.0.0").publish();
    Package::new("bar", "1.0.0")
        .file("src/lib.rs", "")
        .target_dep("only_newer", "1.0", "cfg(version(\"1.2345\"))")
        .publish();

    let p = project()
        .file(
            "Cargo.toml",
            r#"
                [package]
                name = "foo"
                edition = "2024"

                [dependencies]
                bar = "1.0"
            "#,
        )
        .file("src/lib.rs", "")
        .build();

    p.cargo("tree")
        .with_stdout_data(str![[r#"
foo v0.0.0 ([ROOT]/foo)
└── bar v1.0.0

"#]])
        .run();
}
```

crates/cargo-platform/src/cfg.rs
mo8it: @epage Any update on that?

I will have time on Tuesday or Wednesday to finish the PR after clarifying the desired behavior for nightly.

epage: what ehuss said is what we should go with: all unstable versions are considered the same as their stable version.  I tried pushing for explicit tests for this to explicitly specify the behavior but I'm getting push back.  It doesn't help that the nightly behavior is considered unstable ([comment](https://github.com/rust-lang/rust/pull/141766#issuecomment-2954324289)) though its important for us to sync up on it.

epage:
FYI a concern was raised on the new stabilization issue (https://github.com/rust-lang/rust/pull/141766#issuecomment-2946372974) about what syntax is used for this.

mo8it:
@epage @ehuss Unfortunately, I was not able to find time to finish this PR. I will also be away for about 3 weeks. So feel free to add your changes and then merge.

epage:
@mo8it don't worry, there is nothing to do atm while the design is being re-evaluated

ranger-ross:
Given that this has sat stale for over a year and the original issue is [`S-blocked-external`](https://github.com/rust-lang/cargo/issues?q=sort%3Aupdated-desc+state%3Aopen+label%3A%22S-blocked-external%22+) I am gonna go ahead and close this. 

Would be great to see this feature land in some form one day.
Thanks for your effort on this

PR #13696, opened 2024-04-03
Switch to using gitoxide by default for listing files

arlosi:
### What does this PR try to resolve?

Uses gitoxide by for listing the contents of a git repository by default. Fixes #10150

It's possible out-opt of this change with the environment variable `__CARGO_GITOXIDE_DISABLE_LIST_FILES=1`. This opt-out mechanism is temporary and will be removed before the next release.

### How should we test and review this PR?

The newly added test fails with the `git2` implementation.

PR #15534, opened 2025-05-16
Use `gix` for `cargo package`

Byron:
This should also help fixing these spurious "cannot package because some excluded file is untracked" issues.

### Tasks

* [ ] step-by-step conversion of `vcs.rs`
* [ ] use proper feature toggle
* [ ] cleanup


src/cargo/ops/cargo_package/vcs.rs
weihanglo, line 1: Thanks for kicking of this!

Regardless the approach taken here. I wonder if it is possible to avoid checking repo state completely. In `PathSource::list_files` we already walked through it. And maybe we can attach extra info in `PathEntry`. Of course it'll slow down `list_files` a bit, but with that we could perhaps resolve <https://github.com/rust-lang/cargo/issues/14941>?.

It is just a bit relevant to this PR though. We can still merge this when ready.

Byron: It's about time I get started with this - this setup here is super-hacky, but it's designed to allow a very step-wise conversion, hopefully allowing to work on it 15 minutes at a time.

Regarding #14941, I hope with #14955 I have found a representative issue. Right now I can't tell how that should be solved, so I'd probably go for having a `gitoxide` alternative for `cargo package` here and tackle the related issues another time. Ideally, the new implementation will be a bit faster and I hope I can test that on the `aws-sdk-rust` repository.

Byron: I couldn't resist to profile the `aws-sdk-rust` publishing step in Cargo and here is a picture of the run so far.

Screenshot 2025-05-17 at 12 54 31
6a-a1e5-e0dc7977a2f4" />

I don't know if this will ever finish, and I don't know how they can even get this published, maybe it just takes a very long time.

The culprit, maybe among other things, may be `repo.status_file()` calls for which there isn't an equivalent in `gitoxide` anyway. Thus I think this will look very differently when this PR is done and maybe that's an avenue for `aws-sdk-rust` to publish much faster.

I will definitely test that once it's working and share results :).

weihanglo: Oops I linked to the wrong issue. Was meant to say #14955

Byron: I just realized that `dirty_files_outside_pkg_root` is probably the reason for the individual `git_status_file` calls.
The `gitoxide` implementation can certainly be backported to the `git2` one if there is demand, so potential gains could be available sooner.

Byron: Also I realize that even with these improvements, and without `--no-vcs-info` or similar it will still do one git-status costing about 1s each for each of these 404 crates, wasting nearly 7 minutes. Nonetheless, it's probably acceptable compared to the over three ours that it took otherwise (which would be equivalent to ~10800 git status checks).

Byron: And I think #15416 is also properly fixed now, which also means the output is different in 2 tests when using the `gitoxide` implementation.

weihanglo: Nice! Really appreciate the effort here!

We are probably not going to backports, as the backporting process involves a bit more work, and usually reserved for some regression that has no way out. 

Also, given the cargo team has agreed that the dirtiness check is best effort and informational only, I think once the PR is ready we can merge it without FCP.

Thanks again, really a great job!

src/cargo/ops/cargo_package/vcs/gix.rs
```diff
@@ -0,0 +1,408 @@
...
+    if !manifest_exists {
+        gctx.shell().verbose(|shell| {
+            shell.warn(format_args!(
+                "Cargo.toml not found at `{}` in workdir `{}`",
+                path.display(),
+                workdir.display()
+            ))
+        })?;
+        // TODO: Either remove this whole block, or have a test.
+        //       It's hard to have no Cargo.toml here?
```
Byron, line 68: It's unclear to me how this could possibly happen - there is no test for it either. From looking at `git blame` it wasn't obvious what motivated this check.
Probably nothing wrong with keeping it.

Byron (commented):
I think this is ready for a first review. Only `gitoxide` needs a release which is planned for the 22nd, which is when this PR becomes formally mergeable as it can point to the new release instead of the Git repository.

Especially after RustConf where I was quite apologetic and feared `gitoxide` could be removed again for not making progress, I thought maybe we should instead consider stabilizing the well-working parts earlier, such as in:

* nightly with opt-in
* stable with opt-out
* stable, with `git2` codepath removed.

`cargo package` would be candidate for it as the new implementation is more correct and as fast as it can be.

What do you think?

src/cargo/ops/cargo_package/mod.rs
```diff
@@ -450,7 +450,11 @@ fn prepare_archive(
     let src_files = src.list_files(pkg)?;
 
     // Check (git) repository state, getting the current commit hash.
-    let vcs_info = vcs::check_repo_state(pkg, &src_files, ws, &opts)?;
+    let vcs_info = if ws.gctx().cli_unstable().gitoxide.is_some() {
```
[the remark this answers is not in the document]
weihanglo: Personally I would lean towards making it insta-stable, as I mentioned the check is a best-effort, though I'll put this in Cargo team meeting agenda next week!

src/cargo/ops/cargo_package/vcs/mod.rs
weihanglo, line 1: When merge, I'll recommend we split rename file into a single comment, so Git knows better to track it.

src/cargo/ops/cargo_package/vcs/gix.rs
weihanglo: Maybe only some parts of the package is under git? Like `src` is a git repo.

src/cargo/ops/cargo_package/vcs/gix.rs
```diff
@@ -0,0 +1,404 @@
...
+                repo.path().display()
+            )
+        })?;
+
+        let rel_path = gix::path::from_bstr(status.location());
+        let path = workdir.join(&rel_path);
+        if relative_package_root.is_some_and(|pkg_root| !rel_path.starts_with(pkg_root)) {
+            dirty_files_outside_of_package_root.push(path);
+            continue;
+        }
```
weihanglo, line 297: Out of curiosity. If we adopt the same optimization in git2 version, how the performance would look like?

src/cargo/ops/cargo_package/vcs/gix.rs
weihanglo, line 1: Looks like we are mostly doing the same thing except using gitoxide. Do I understand correctly?
Apart from the `dirty_files_outside_of_package_root` work, have you found any other stuff to optmize?

src/cargo/ops/cargo_package/mod.rs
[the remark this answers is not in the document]
weihanglo: To avoid wasting efforts, let's hold off this until we hear back from the team :)

src/cargo/ops/cargo_package/vcs/gix.rs
Byron: > Maybe only some parts of the package is under git? Like src is a git repo.

It seemed that what the code really was testing is if the file exists, and not if it was tracked by Git. Thus I made it an explicit 'file exists' check, which seemed strange as Cargo won't function without such a manifest. Removing the whole block doesn't fail CI either, so it's a bit unclear why it's there. There is still no tool I know that could do forensics and find everything that Git knows about it, that would certainly be interesting.

Byron: > Out of curiosity. If we adopt the same optimization in git2 version, how the performance would look like?

The 'architecture' changes here could all be applied to the `git2` implementation, with the same effects. Performance problems should be solved then.

src/cargo/ops/cargo_package/vcs/gix.rs
Byron: The driver of these changes is the lack of single-file APIs, and I noticed no other (previously unknown) optimisation opportunities. 
But now that I am thinking about it, there might be one: I a workspace like `aws-sdk-rust` I believe this code runs once for each crate, meaning there will be a full `git status` each time. It would certainly require more complex changes to be able to cache these between the publish calls. Each one takes a little less than a second there, so it's quite some time in aggregate.
But before that, I'd probably optimize the `fetch crates` stage away, it takes most of the time.

weihanglo:
While the PR is not ready to merge yet, I am kicking off the process to get a consensus on how we want to merge this.

@rfcbot fcp merge

Proposal: Switch to gix for `cargo package` VCS dirtiness check completely, _without_ any fallback to libgit2.

Reasons:

* This is mostly a performance fix. Not really something fundamentally changed.
* Unlike #13696, VCS dirtiness check informational only, as the team have agreed on.
* There is already `--allow-dirty` users can opt out the functionality completely.

src/cargo/ops/cargo_package/mod.rs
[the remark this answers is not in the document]
weihanglo: https://github.com/rust-lang/cargo/pull/15534#issuecomment-2895041842

Looks like people are okay with no fallback to libgit2. I think we can move forward just changing the implementation of `vcs::check_repo_state`.

Byron:
It's too bad this had to wait for so long but I decided to finally get it off my queue, and replaced the `git2` implementation with the `gix` one entirely as suggested.

Did I miss anything that was part of the review?

Further, the Cargo test-suite discovered a bug that makes submodule listings fail.

I will fix that in `gix`, but also feel like it could be caught here to not further delay merging and wait until `gix` is released on the 22nd or so. What do you think?

weihanglo:
If it is an existing bug, we can wait and fix it later. Regardless, do whatever makes sense and is easy to follow on both review and implementation side.

Byron:
I realised that the `gitoxide` issue was already fixed 2 months ago, and made a release instead.

Unfortunately, there still is a failure on Windows related to how `gix status` sees symlinks - apparently it doesn't detect it's a link and walks right through, somewhat rightfully considering the file in the submodule untracked.

For today I am out of time, but I'd hope to debug this more tomorrow.

Byron:
Without me doing anything, the checks have passed.

My last commit (`fix test that fails on CI`, now removed) added some debug output to gather more dat for today, and that just passed for no good reason.

Let's see how prone it now is to failure. The test-setup is simple as well and it is something that should be reproducible in `gitoxide` as well, it's just the question how much effort it takes to reproduce on a local Windows VM, or if it reproduces at all. In the worst case, flakiness on CI only, it would be much tougher to fix, particularly because the issue had a smell of "filesystem race condition".

I marked the PR as ready for review nonetheless, in case CI passes, and will be back to investigate this phantom otherwise.

tests/testsuite/package.rs
```diff
@@ -1470,13 +1470,18 @@ fn dirty_file_outside_pkg_root_inside_submodule() {
...
     git::add(&repo);
     git::commit(&repo);
-    // This dirtyness should be detected in the future.
+    // This dirtiness should be detected in the future.
     p.change_file("submodule/file.txt", "changed");
 
-    p.cargo("package --workspace --no-verify")
+    let mut execs = p.cargo("package --workspace --no-verify");
+    execs
+        .with_status(101)
```
weihanglo: nit

```suggestion
    p.cargo("package --workspace --no-verify")
        .with_status(101)
```

tests/testsuite/package.rs
```diff
@@ -1470,13 +1470,18 @@ fn dirty_file_outside_pkg_root_inside_submodule() {
     p.symlink("submodule/file.txt", "isengard/src/file.txt");
     git::add(&repo);
     git::commit(&repo);
-    // This dirtyness should be detected in the future.
+    // This dirtiness should be detected in the future.
```
weihanglo: This line can be deleted.

src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -206,14 +243,11 @@ fn git(
         .collect();
     let dirty = !dirty_src_files.is_empty();
     if !dirty || opts.allow_dirty {
-        // Must check whetherthe repo has no commit firstly, otherwise `revparse_single` would fail on bare commit repo.
+        // Must check whether the repo has no commit firstly; otherwise `revparse_single` would fail on bare commit repo.
         // Due to lacking the `sha1` field, it's better not record the `GitVcsInfo` for consistency.
```
weihanglo: The comment here is not relevant anymore I guess?

src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -228,6 +262,112 @@ fn git(
...
+/// Helper to collect dirty statuses for a single repo.
+/// `relative_package_root` is `Some` if the root is a sub-directory of the workdir.
+/// Returns the dirty files outside `relative_package_root`.
+#[must_use]
+fn collect_statuses(
+    repo: &gix::Repository,
+    workdir: &Path,
+    relative_package_root: Option<&Path>,
+    dirty_files: &mut Vec<PathBuf>,
+) -> CargoResult<Vec<PathBuf>> {
```
weihanglo: nit: May be just personal taste. I feel like it is a bit inconsistent that `dirty_files` is collected via an in-out parameter, while `dirty_files_outside_of_package_root` is a return value. Could we make them both a parameter, and have maybe one or two lines of comment to explain their difference?

weihanglo (commented):
Nice!! Just some nits and let me know what you think :)

src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -228,6 +262,112 @@ fn git(
...
+                // them is desirable here.
+                .emit_ignored(Some(EmissionMode::CollapseDirectory))
+                .emit_tracked(false)
+                .recurse_repositories(false)
+                .symlinks_to_directories_are_ignored_like_directories(true)
+                .emit_empty_directories(false)
+        })
+        .tree_index_track_renames(TrackRenames::Disabled)
+        .index_worktree_submodules(None)
+        .into_iter(None /* pathspec patterns */)
```
weihanglo: How is the recent benchmark? Just compared with 6833aa715d724437dc1247d0166afe314ab6854e against aws-sdk-rust and this PR slows down the process. Haven't checked what caused that. I assume it might be we collecting status for the entire repo for each workspace member, instead of using pathspec?

```
﻿﻿cargo package --allow-dirty --workspace --no-verify --no-metadata --offline
```

I ran this btw.

weihanglo: `CARGO_LOG_PROFILE=1 cargo package --allow-dirty -p aws-sigv4 --no-verify --no-metadata --offline`

### master branch: 6833aa715d724437dc1247d0166afe314ab6854e

[master-6833aa7.json](https://github.com/user-attachments/files/21263734/master-6833aa7.json)

master


### This PR d41d177506079f5b781e439d74a0c62769a082f8

gix-pr

[pr15534-d41d177.json](https://github.com/user-attachments/files/21263736/pr15534-d41d177.json)

Byron: *I didn't have time today as planned to look into this, but just had to leave a comment here nonetheless*.

This is a very interesting find! On the latest `main` of the `aws-sdk-rust` repository, the status time for the whole repository is ~600ms for me (hot cache). It's strange that this takes so long in your measurements and also, that the `git2` version that uses `status_file` on each file is fast(er) all of the sudden. It's not at all what I measured in my run.

So I will definitely look into this.
Edit: I can definitely reproduce the slowdown, awesome catch!

src/cargo/ops/cargo_package/vcs.rs
Byron: That's a great catch, it looks and feels much better when doing what's suggested here.

src/cargo/ops/cargo_package/vcs.rs
Byron: I started with `hyperfine`:

```
❯ hyperfine -w3 '/Users/byron/dev/github.com/rust-lang/cargo/target/release/cargo package --no-verify --no-metadata -p aws-smithy-runtime-api --offline' 'cargo package --no-verify --no-metadata -p aws-smithy-runtime-api --offline'
Benchmark 1: /Users/byron/dev/github.com/rust-lang/cargo/target/release/cargo package --no-verify --no-metadata -p aws-smithy-runtime-api --offline
  Time (mean ± σ):     967.5 ms ±  75.5 ms    [User: 718.7 ms, System: 3491.2 ms]
  Range (min … max):   923.2 ms … 1111.0 ms    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: cargo package --no-verify --no-metadata -p aws-smithy-runtime-api --offline
  Time (mean ± σ):     593.1 ms ±   7.2 ms    [User: 506.7 ms, System: 93.1 ms]
  Range (min … max):   585.3 ms … 610.4 ms    10 runs

Summary
  cargo package --no-verify --no-metadata -p aws-smithy-runtime-api --offline ran
    1.63 ± 0.13 times faster than /Users/byron/dev/github.com/rust-lang/cargo/target/release/cargo package --no-verify --no-metadata -p aws-smithy-runtime-api --offline
```

Even though the result is less pronounced, it's still 1.6x slower than it was before. The timings also fluctuate much more, probably a side-effect of spawning many threads for the full status check.

The profile run shows the same. Here we see how `gitoxide` splits up the status run into parallel operations, with the dirwalk to find untracked files being the slowest at ~700ms wallclock, the index check is parallelized and finishes in ~300ms wallclock, and finally there is the tree-index diff with ~80ms.

Screenshot 2025-07-18 at 05 17 20

Also, there is no other bottleneck when looking at the actual `package` function.

Screenshot 2025-07-18 at 05 24 59

This means in theory, if the root of the project would be provided here…

https://github.com/rust-lang/cargo/blob/a67e9ec8a6720217b8e12d2b897b829271cac4f5/src/cargo/ops/cargo_package/vcs.rs#L292

…one should be able to save a lot of time.

Initially I thought pathspecs weren't supported, but realised that `dirty_files_outside_package_root` is the reason it runs the status on the whole repository.

To me it feels inevitable to run the whole status, but the previous implementation certainly shows how it could be done. I also wonder where the performance problems are that I encountered.

My intuition here would be to eat the cost when publishing single repositories, but when multiple are done, it could do the full-repo status just once, while using a pathspec to do a per-project status instead.

I am looking at this loop:

https://github.com/rust-lang/cargo/blob/46ad7afcd034df840586e0e73ff46ec1d6a5d2e6/src/cargo/ops/cargo_package/mod.rs#L288-L294

If that was done, I'd think that a package operation per project could be about 200ms, down from 960ms or 600ms respectively.

For today I am out of time, but if you think this is acceptable (i.e. *possible* regression for single-packages, but speedup for multiple packages), then I'd implement it as described here. The last 'record' to package all of `aws-sdk-rust` is 12m (without  `--offline`), it could probably be 10m without that extra expense.

Kobzol: Btw, just a walkby comment: it seemed to me like you were comparing yourself to distributed (not locally built) cargo? If that's the case, I just wanted to note that we build Cargo on CI with `lto=thin`, which is otherwise not configured in the Cargo repo. The LTO config should extend also to the built C/C++ code, so to libgit.

weihanglo: What about we do the same logic in gitoxide first, and the optimize it in follow ups? Would the same logic as the current libgit2 one be faster in gitoxide?

Byron: > Btw, just a walkby comment: it seemed to me like you were comparing yourself to distributed (not locally built) cargo? If that's the case, I just wanted to note that we build Cargo on CI with `lto=thin`, which is otherwise not configured in the Cargo repo. The LTO config should extend also to the built C/C++ code, so to libgit.

Thanks for the hint, it's appreciated! 
In this case, I think the effect of this would be minor as the time is spent with IO, not in heavy computation that could be more optimised in official release builds.

> What about we do the same logic in gitoxide first, and the optimize it in follow ups? Would the same logic as the current libgit2 one be faster in gitoxide?

I took a stab at that and it seems to be up to 1.4x faster than the respective `git2` version, in the best case. It turned out that the candidates for dirty files outside of the package root is very small, maybe even no more than 1 or 2 files. So it definitely makes sense to avoid all that extra work.
Let's see of using pathspecs works as it should.

Byron: Alright, now that I am done with the change, back to the previous algorithm, we are looking at 'only' 25% faster packaging in our particular example.

```
❯ hyperfine -w2 '/Users/byron/dev/github.com/rust-lang/cargo/target/release/cargo package --no-verify --no-metadata -p aws-smithy-runtime-api --offline' 'cargo package --no-verify --no-metadata -p aws-smithy-runtime-api --offline'
Benchmark 1: /Users/byron/dev/github.com/rust-lang/cargo/target/release/cargo package --no-verify --no-metadata -p aws-smithy-runtime-api --offline
  Time (mean ± σ):     487.4 ms ±   8.1 ms    [User: 419.8 ms, System: 103.2 ms]
  Range (min … max):   479.3 ms … 508.9 ms    10 runs

Benchmark 2: cargo package --no-verify --no-metadata -p aws-smithy-runtime-api --offline
  Time (mean ± σ):     604.6 ms ±   4.6 ms    [User: 518.0 ms, System: 93.9 ms]
  Range (min … max):   597.9 ms … 613.1 ms    10 runs

Summary
  /Users/byron/dev/github.com/rust-lang/cargo/target/release/cargo package --no-verify --no-metadata -p aws-smithy-runtime-api --offline ran
    1.24 ± 0.02 times faster than cargo package --no-verify --no-metadata -p aws-smithy-runtime-api --offline
```

This should address the final issue :).

src/cargo/ops/cargo_package/vcs.rs
```diff
@@ -228,36 +260,119 @@ fn git(
...
 /// * `package.readme` and `package.license-file` pointing to paths outside package root
-/// * symlinks targets reside outside package root
+/// * symlinks targets residing outside package root
 /// * Any change in the root workspace manifest, regardless of what has changed.
 ///
 /// This is required because those paths may link to a file outside the
 /// current package root, but still under the git workdir, affecting the
 /// final packaged `.crate` file.
-fn dirty_files_outside_pkg_root(
+fn dirty_files_outside_pkg_root_orig(
```
weihanglo: Probably forgot to rename back?

```suggestion
fn dirty_files_outside_pkg_root(
```

Byron: Indeed 😅.

Byron:
Thanks again for the review and the additional performance testing, @weihanglo, it's much appreciated. It's good to see it's also faster on Linux, albeit by a smaller margin.

I have squashed the PR down into 3 commits, one for the gix update, then the main implementation, and a last one for the performance optimisation hoping that it might prevent someone from thinking the previous version could be faster or simpler.

Can't wait for this to finally land :).

weihanglo:
> It's good to see it's also faster on Linux, albeit by a smaller margin.

I was using c6a.24xlarge which is fairly not representative 😬

weihanglo (commented):
## 🚀

weihanglo (approved):

PR #15562, opened 2025-05-19
Allow configuring arbitrary codegen backends

bjorn3:
Rather than just builtin codegen backends. This is perma-unstable on the rustc side, but is useful to enable a codegen backend built separately from rustc to be used without overwriting RUSTFLAGS as well as to allow configuring the codegen backend on a per-package basis.

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -2465,25 +2465,15 @@ pub fn validate_profile(
     features: &Features,
     warnings: &mut Vec<String>,
 ) -> CargoResult<()> {
-    validate_profile_layer(root, name, cli_unstable, features)?;
+    validate_profile_layer(root, cli_unstable, features)?;
     if let Some(ref profile) = root.build_override {
         validate_profile_override(profile, "build-override")?;
-        validate_profile_layer(
-            profile,
-            &format!("{name}.build-override"),
-            cli_unstable,
-            features,
-        )?;
+        validate_profile_layer(profile, cli_unstable, features)?;
     }
     if let Some(ref packages) = root.package {
-        for (override_name, profile) in packages {
+        for profile in packages.values() {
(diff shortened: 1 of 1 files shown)


src/cargo/util/toml/mod.rs
```diff
@@ -2548,26 +2538,17 @@ pub fn validate_profile(
...
             _ => {}
         }
-
-        if codegen_backend.contains(|c: char| !c.is_ascii_alphanumeric() && c != '_') {
-            bail!(
-                "`profile.{}.codegen-backend` setting of `{}` is not a valid backend name.",
-                name,
-                codegen_backend,
-            );
-        }
```
bjorn3, line 2570: If preferred I can also keep this check on the cargo side when nightly features are disabled.

weihanglo: How arbitrary you'd like to have for codegen-backend name? What is the real use case here?

The change seems fine to me though, as a nightly feature.

bjorn3: Basically `-Zcodegen-backend` allows either of the following:

* The name of a builtin codegen backend. This will eventually be stabilized once one of the non-LLVM backends gets stabilized.
* The full path to a dylib implementing a codegen backend. **This will never be stable.** It exists to allow codegen backends not shipped as part of rustc to be used. For example cg_clif uses it when you locally build it or when using the version from its CI rather than the rustup distributed version (the CI version is newer). Same for cg_gcc when you locally build it.

weihanglo: Thanks for the info! It makes sense to relax it a bit for now.

When stabilization, Cargo should only allow a set of known backends. We should track it somewhere, though we dont really have a Cargo tracking issue for codegen-backend right now.

weihanglo (approved):

PR #15569, opened 2025-05-21
fix(vendor)!: vendor files with .rej/.orig suffix

### LINKED ISSUE #13191: Normalization of `Cargo.toml` can break crates with deps. that read `Cargo.toml`

weihanglo:

This is meant to fixes #13191

As git sources and registry sources are considered immutable.
I don't think there is any reason excluding those files.
There might be a little chance local Git repositories might have those,
though that is a rare use case.

Alternatively,
we could reject all `.rej`/`.orig` files but `Cargo.toml.orig`.

Test updates should be sufficient.

### Additional information

This is a follow-up of #15514

### DIFF
--- a/src/cargo/ops/vendor.rs
+++ b/src/cargo/ops/vendor.rs
@@ -618,10 +618,6 @@ fn vendor_this(relative: &Path) -> bool {
         // Temporary Cargo files
         Some(".cargo-ok") => false,
 
-        // Skip patch-style orig/rej files. Published crates on crates.io
-        // have `Cargo.toml.orig` which we don't want to use here and
-        // otherwise these are rarely used as part of the build process.
-        Some(p) if p.ends_with(".orig") || p.ends_with(".rej") => false,
         _ => true,
     }
 }
--- a/tests/testsuite/vendor.rs
+++ b/tests/testsuite/vendor.rs
@@ -1087,13 +1087,14 @@ fn ignore_files() {
     Package::new("url", "1.4.1")
         // These will be vendored
         .file(".cargo_vcs_info.json", "")
+        .file("Cargo.toml.orig", "")
+        .file("foo.orig", "")
+        .file("foo.rej", "")
         .file("src/lib.rs", "")
         // These will not be vendored
         .file(".cargo-ok", "")
         .file(".gitattributes", "")
         .file(".gitignore", "")
-        .file("foo.orig", "")
-        .file("foo.rej", "")
         .publish();
 
     p.cargo("vendor --respect-source-config").run();
@@ -1105,6 +1106,9 @@ fn ignore_files() {
   "files": {
     ".cargo_vcs_info.json": "[..]",
     "Cargo.toml": "[..]",
+    "Cargo.toml.orig": "[..]",
+    "foo.orig": "[..]",
+    "foo.rej": "[..]",
     "src/lib.rs": "[..]"
   },
   "package": "[..]"
--- a/tests/testsuite/vendor.rs
+++ b/tests/testsuite/vendor.rs
@@ -1094,11 +1094,19 @@ fn ignore_files() {
 
     p.cargo("vendor --respect-source-config").run();
     let csum = p.read_file("vendor/url/.cargo-checksum.json");
-    assert!(!csum.contains("foo.orig"));
-    assert!(!csum.contains(".gitignore"));
-    assert!(!csum.contains(".gitattributes"));
-    assert!(!csum.contains(".cargo-ok"));
-    assert!(!csum.contains("foo.rej"));
+    assert_e2e().eq(
+        csum,
+        str![[r#"
+{
+  "files": {
+    "Cargo.toml": "[..]",
+    "src/lib.rs": "[..]"
+  },
+  "package": "[..]"
+}
+"#]]
+        .is_json(),
+    );
 }
 
 #[cargo_test]
--- a/tests/testsuite/vendor.rs
+++ b/tests/testsuite/vendor.rs
@@ -1085,10 +1085,14 @@ fn ignore_files() {
         .build();
 
     Package::new("url", "1.4.1")
+        // These will be vendored
+        .file(".cargo_vcs_info.json", "")
         .file("src/lib.rs", "")
-        .file("foo.orig", "")
-        .file(".gitignore", "")
+        // These will not be vendored
+        .file(".cargo-ok", "")
         .file(".gitattributes", "")
+        .file(".gitignore", "")
+        .file("foo.orig", "")
         .file("foo.rej", "")
         .publish();
 
@@ -1099,6 +1103,7 @@ fn ignore_files() {
         str![[r#"
 {
   "files": {
+    ".cargo_vcs_info.json": "[..]",
     "Cargo.toml": "[..]",
     "src/lib.rs": "[..]"
   },


weihanglo:
Maybe I should do it in reversed order.

src/cargo/ops/vendor.rs
```diff
@@ -618,10 +618,6 @@ fn vendor_this(relative: &Path) -> bool {
         // Temporary Cargo files
         Some(".cargo-ok") => false,
 
-        // Skip patch-style orig/rej files. Published crates on crates.io
-        // have `Cargo.toml.orig` which we don't want to use here and
-        // otherwise these are rarely used as part of the build process.
-        Some(p) if p.ends_with(".orig") || p.ends_with(".rej") => false,
```
epage, line 624: What are `.rej` files?

src/cargo/ops/vendor.rs
```diff
@@ -618,10 +618,6 @@ fn vendor_this(relative: &Path) -> bool {
         // Temporary Cargo files
         Some(".cargo-ok") => false,
```
epage, line 619: Now that we are doing fresh extractions, is this still needed?

src/cargo/ops/vendor.rs
weihanglo: From GNU patch

> The names for reject files (files containing patches that patch could not find a place to apply) are normally the name of the output file with ‘.rej’ appended (or ‘#’ if using ‘.rej’ would make the backup file name too long). 

https://www.gnu.org/software/diffutils/manual/html_node/Reject-Names.html

epage: I feel like if the file is inside the `.crate`, we might as well expose it, so this seems fine.

src/cargo/ops/vendor.rs
weihanglo: For git checkouts we still have that, though there is no test exercising that code path. Do you want a test for it in this PR?

epage (approved):

src/cargo/ops/vendor.rs
epage: Let's go ahead and merge this

PR #15573, opened 2025-05-21
fix(embedded): Resolve multiple bugs in frontmatter parser

epage:

This pulls in the tests from rust-lang/rust#140035 and ensures we pass them, including switching our parser to be more like the one from rust-lang/rust#137193.

I've added comments specifying why our test results differ from rustcs.

This is part of #12207

### Additional information

We still have work to do to improve the quality of our error messages

### DIFF
--- a/src/cargo/util/toml/embedded.rs
+++ b/src/cargo/util/toml/embedded.rs
@@ -120,9 +120,6 @@ impl<'s> ScriptSource<'s> {
         if !info.is_empty() {
             source.info = Some(info);
         }
-        let rest = rest
-            .strip_prefix('\n')
-            .expect("earlier `found` + `split_at` left us here");
 
         // Ends with a line that starts with a matching number of `-` only followed by whitespace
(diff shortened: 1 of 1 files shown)


src/cargo/util/toml/embedded.rs
weihanglo, line 1: I don't quite follow the conversion in <https://github.com/rust-lang/rust/pull/137193> or the other merged PR. How much effort should Cargo provide to parse the frontmatter? Do we still want to deletect to rustc to some diagnostics? Or I just need to make sure this parser look correct on its own, without looking at rustc part?

epage: Any error we produce will fire before rustc ever gets a chance to report one, so we should aim for rustc-like error messages (yes, we are not there yet). However, we can't catch all cases and have to defer to rustc for those.

Alternatively, we could be really sloppy and ignore, rather than error. The downside is an error will only be produced if rustc is invoked. I'd prefer to not go this route.

weihanglo (approved):
Thanks!

PR #15574, opened 2025-05-22
Add autocomplete for nested examples

MrMinimal:

Not all examples show up in auto-completion.
If the `examples` directory has nested examples, only the top level ones will be shown with  `cargo build --example<TAB>`.
[Bevy examples](https://github.com/bevyengine/bevy/tree/latest/examples) only show `hello_world` instead of all available examples.

image

Old:
image

New:
image

Go into a Rust project with nested example directories (Bevy for example) and press `cargo build --example<TAB>`.
Observe all examples being listed now.

### DIFF
--- a/src/etc/cargo.bashcomp.sh
+++ b/src/etc/cargo.bashcomp.sh
@@ -246,7 +246,7 @@ _get_examples(){
 	local manifest=$(_locate_manifest)
 	[ -z "$manifest" ] && return 0
 
-	local files=("${manifest%/*}"/examples/*.rs)
+	local files=("${manifest%/*}"/examples/**/*.rs)
 	local names=("${files[@]##*/}")
 	local names=("${names[@]%.*}")
 	# "*" means no examples found


src/etc/cargo.bashcomp.sh
```diff
@@ -246,7 +246,7 @@ _get_examples(){
 	local manifest=$(_locate_manifest)
 	[ -z "$manifest" ] && return 0
 
-	local files=("${manifest%/*}"/examples/*.rs)
+	local files=("${manifest%/*}"/examples/**/*.rs)
```
epage, line 249: Should we do something similar for zsh?
https://github.com/rust-lang/cargo/blob/c2c636a6dbe98f0b1a6f14ff5eefe2aa66a9243e/src/etc/_cargo#L472-L477

epage, line 249: This could pick up both examples and submodules.  Unsure whether that is worth it or not

epage:
Are you aware of our experimental completion system ([docs](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#native-completions))?  It uses `cargo` to do things like look up all examples.

MrMinimal:
@epage I wasn't aware of the experimental system, just wanted to contribute back a quick fix I had.
If the new system replaces this one, I'll close the PR.

src/etc/cargo.bashcomp.sh
ehuss: I would be a little worried that the false-positive rate would be too high. In bevy for example, this would auto-complete a ton of names that aren't examples.

I think the only real solution here is to parse Cargo.toml, and I think native-completions is the only way to realistically do that. This bashcomp could do some json parsing of `cargo metadata`, but I wouldn't want to go there.

PR #15587, opened 2025-05-22
docs: add lock file section to the resolver docs

Fishrock123:
Adds some documentation about how lockfiles work to the resolver section, particularly regarding when lock files exist in library dependencies.

Effectively re-adds an updated section of docs that was lost here: https://github.com/rust-lang/cargo/pull/12382/files#diff-197a732275c32bdbdb079bdd92ac8a4ba585ee556ea978e9e661804eb76ce9eeL117-L121

Related to this Zulip thread https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/cargo.20lock.20in.20dependency/with/519679182

### DIFF
--- a/src/doc/src/reference/resolver.md
+++ b/src/doc/src/reference/resolver.md
@@ -12,6 +12,25 @@ resolver.
 [dependency specification]: specifying-dependencies.md
 [`cargo tree`]: ../commands/cargo-tree.md
 
+## `Cargo.lock` lock-files
+
+The `Cargo.lock` file provides deterministic builds at different times and on
+different systems, by ensuring that the exact same dependencies, versions, and
+sources are used as when the `Cargo.lock` file was last generated.
+Dependancy resolution is not run when cargo reads from `Cargo.lock`.
+
+Not all cargo commands use `Cargo.lock` by default. Examples include
+`cargo install` and `cargo update`.In these cases, `--locked` can usually be
+passed to force cargo to use the dependency graph from an existing `Cargo.lock`
+(if one exists).
+
+### Libraries with `Cargo.lock`
+
+Cargo treats `Cargo.lock` files differently when a crate is used as a library
+dependency where an upper level `Cargo.lock` would exist. In these cases cargo will
+not inspect the dependency's `Cargo.lock` even if it exists, using only the
+top-level `Cargo.lock`.
+
 ## Constraints and Heuristics
 
 In many cases there is no single "best" dependency resolution.


src/doc/src/reference/resolver.md
```diff
@@ -12,6 +12,25 @@ resolver.
 [dependency specification]: specifying-dependencies.md
 [`cargo tree`]: ../commands/cargo-tree.md
 
+## `Cargo.lock` lock-files
+
+The `Cargo.lock` file provides deterministic builds at different times and on
```
weihanglo, line 17: There are some other places the book mentions Cargo.lock

* https://doc.rust-lang.org/nightly/cargo/appendix/glossary.html
* https://doc.rust-lang.org/nightly/cargo/guide/cargo-toml-vs-cargo-lock.html#cargotoml-vs-cargolock
* https://doc.rust-lang.org/nightly/cargo/faq.html#why-have-cargolock-in-version-control
* The previous paragraph.

it is a bit duplicate everywhere. We might want some level of consolidation.

src/doc/src/reference/resolver.md
```diff
@@ -12,6 +12,25 @@ resolver.
...
+different systems, by ensuring that the exact same dependencies, versions, and
+sources are used as when the `Cargo.lock` file was last generated.
+Dependancy resolution is not run when cargo reads from `Cargo.lock`.
+
+Not all cargo commands use `Cargo.lock` by default. Examples include
+`cargo install` and `cargo update`.In these cases, `--locked` can usually be
```
weihanglo, line 23: This is a bit off. `cargo update` would always consult the existing Cargo.lock, especially when doing a partial update.

src/doc/src/reference/resolver.md
```diff
@@ -12,6 +12,25 @@ resolver.
 [dependency specification]: specifying-dependencies.md
 [`cargo tree`]: ../commands/cargo-tree.md
 
+## `Cargo.lock` lock-files
+
+The `Cargo.lock` file provides deterministic builds at different times and on
+different systems, by ensuring that the exact same dependencies, versions, and
+sources are used as when the `Cargo.lock` file was last generated.
+Dependancy resolution is not run when cargo reads from `Cargo.lock`.
```
weihanglo, line 20: This is incorrect. Cargo always performs dependency resolutions and prefers to versions tracked in Cargo.lock.

src/doc/src/reference/resolver.md
```diff
@@ -12,6 +12,25 @@ resolver.
...
+
+Not all cargo commands use `Cargo.lock` by default. Examples include
+`cargo install` and `cargo update`.In these cases, `--locked` can usually be
+passed to force cargo to use the dependency graph from an existing `Cargo.lock`
+(if one exists).
+
+### Libraries with `Cargo.lock`
+
+Cargo treats `Cargo.lock` files differently when a crate is used as a library
+dependency where an upper level `Cargo.lock` would exist. In these cases cargo will
```
weihanglo, line 30: Upper-level / top-level is pretty vague here. Cargo actually ignores any lockfile but the one at the workspace root.

src/doc/src/reference/resolver.md
weihanglo, line 1: Perhaps instead, we can just talk about the property of a single view of dependency resolution, so that we don't repeat ourselves and still provide what you originally wanted to know.

```markdown
### Single view of dependency resolution

In Cargo, when building a package or a workspace, all packages and their
dependencies share a single, unified view of dependency resolution (which
results in `Cargo.lock`). This maximizes the reuse of package versions
within the resolution process. Unlike some other package managers, Cargo
does not inspect a local resolution result from an individual dependency,
nor it reads their `Cargo.lock` files.

```

epage:
For myself, I would try to fit lockfiles into a new section under `Constraints and Heuristics`.  Unsure if I'm losing too much nuance but we can describe lockfiles as being relevant to `pick_next_version`, preferring the locked version over all else.

epage:
> For myself, I would try to fit lockfiles into a new section under `Constraints and Heuristics`. Unsure if I'm losing too much nuance but we can describe lockfiles as being relevant to `pick_next_version`, preferring the locked version over all else.

Did this in #15958.

ranger-ross:
@Fishrock123 do you still plan to pursue this PR?

#15958 added some documentation but about cargo.lock but I think there might still be some value in the "Libraries with `Cargo.lock`" section if corrected.

PR #15595, opened 2025-05-24
Add the future edition

ehuss:
This adds support for the "future" edition which was added to rustc in https://github.com/rust-lang/rust/pull/137606.

To enable support for unstable editions, this introduces a new `unstable-edition` cargo feature. The intent is that instead of having a new feature for each edition that we reuse this feature for all new editions. I don't see a particular reason we should have a separate one for each edition, and this helps a bit with scalability and simplifies some of the edition process.

This also includes a change to rework `supports_compat_lint` explained in the comment.

### DIFF
--- a/src/cargo/core/features.rs
+++ b/src/cargo/core/features.rs
@@ -151,11 +151,8 @@ pub type AllowFeatures = BTreeSet<String>;
 /// - Update the [`FromStr`] impl.
 /// - Update [`CLI_VALUES`] to include the new edition.
 /// - Set [`LATEST_UNSTABLE`] to Some with the new edition.
-/// - Add an unstable feature to the [`features!`] macro invocation below for the new edition.
-/// - Gate on that new feature in [`toml`].
 /// - Update the shell completion files.
 /// - Update any failing tests (hopefully there are very few).
-/// - Update unstable.md to add a new section for this new edition (see [this example]).
 ///
 /// ## Stabilization instructions
 ///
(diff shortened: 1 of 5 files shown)


src/cargo/core/features.rs
```diff
@@ -210,11 +209,15 @@ impl Edition {
...
         Self::Edition2024,
+        Self::EditionFuture,
     ];
     /// Possible values allowed for the `--edition` CLI flag.
     ///
     /// This requires a static value due to the way clap works, otherwise I
     /// would have built this dynamically.
+    ///
+    /// This does not include `future` since we don't need to create new
+    /// projects with it.
```
weihanglo, line 220: nit: we use packages 📦 right?

```suggestion
    /// packages with it.
```

src/cargo/core/features.rs
```diff
@@ -274,18 +282,23 @@ impl Edition {
...
+            EditionFuture => {
+                cmd.arg("--force-warn=edition_future_compatibility");
+            }
+            e => {
+                // Note that cargo always passes this even if the
+                // compatibility lint group does not exist. When a new edition
+                // is introduced, but there are no migration lints, rustc does
+                // not create the lint group. That's OK because rustc will
+                // just generate a warning about an unknown lint which will be
+                // suppressed due to cap-lints.
```
weihanglo, line 299: What if people deny unknown lints? Would that fail `cargo fix`?

ehuss: No, cap-lints takes precedence.

src/cargo/core/features.rs
```diff
@@ -519,6 +535,9 @@ features! {
 
     /// Allow paths that resolve relatively to a base specified in the config.
     (unstable, path_bases, "", "reference/unstable.html#path-bases"),
+
+    /// Allows use of editions that are not yet stable.
+    (unstable, unstable_edition, "", "reference/unstable.html#unstable-edition"),
```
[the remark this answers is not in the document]
ehuss: Sure! My thinking was that there is only ever one edition allowed in a manifest, but I'm fine either way.

weihanglo (approved):
Thanks!

PR #15596, opened 2025-05-24
Fix edition

ehuss:
This adds a new `cargo fix -Zfix-edition` permanently unstable flag to assist with edition migration testing, particularly with crater.

The commits and given documentation should explain how it works, but there will be forthcoming edition documentation (on the forge) once all the pieces with crater come together.

⚠ Draft as this is stacked on top of https://github.com/rust-lang/cargo/pull/15595


tests/testsuite/fix.rs
```diff
@@ -2963,3 +2963,104 @@ dep_df_false = { workspace = true, default-features = false }
...
+        .run();
+    assert_e2e().eq(
+        p.read_file("Cargo.toml"),
+        str![[r#"
+cargo-features = ["unstable-edition"]
+
+            [package]
+            name = "foo"
+edition = "future"
+
```
traviscross, line 3063:
```suggestion

```

I can't tell, is the trailing whitespace on this line intended?

ehuss: Yes, it's just an artifact of how the `str!` macro works (which handles automatic updates). There isn't a blank line here.

ehuss: Ah, I see. Yea, that's a little awkward also due to defining inline. I went ahead and removed it.

src/doc/src/reference/unstable.md
```diff
@@ -1915,6 +1916,19 @@ When new editions are introduced, the `unstable-editions` feature is required un
 
 The special "future" edition is a home for new features that are under development, and is permanently unstable. The "future" edition also has no new behavior by itself. Each change in the future edition requires an opt-in such as a `#![feature(...)]` attribute.
 
+## `fix-edition`
+
+`-Zfix-edition` is a permanently unstable flag to assist with testing edition migrations, particularly with the use of crater. It only works with the `cargo fix` subcommand. It takes two different forms:
```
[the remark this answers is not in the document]
traviscross: On the Reference and other documents we're just unwrapping now.

PR #15597, opened 2025-05-25
docs(README): fix the link to the changelog in the Cargo book

AudaciousAxiom:

When Cargo's changelog was moved into the book in  #15123, it seems the extension was left as `.md`, whereas the link should point to the rendered version, which ends with `.html`.

Following the new link in the README's rendered view and observing that it does indeed point to the changelog should be enough.


weihanglo (approved):
Thanks!

PR #15598, opened 2025-05-25
chore(toml): disable `toml`'s default features, unless necessary

AudaciousAxiom:

This reduces the build time of `cargo-util-schemas` by disabling the unused `display` and `parse` features of `toml` (which are enabled by default).

This is useful as `cargo-util-schemas` is intended to be reused outside of Cargo.

As I believe these features only *feature-gate* items on `toml` (they do not affect runtime behavior silently), the CI building successfully should be enough to make sure this does not break anything. ~If this does not build, I will try to tend to it today, otherwise feel free to push to this branch.~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,7 +104,7 @@ tar = { version = "0.4.43", default-features = false }
 tempfile = "3.20.0"
 thiserror = "2.0.11"
 time = { version = "0.3.37", features = ["parsing", "formatting", "serde"] }
-toml = "0.8.20"
+toml = { version = "0.8.20", default-features = false }
 toml_edit = { version = "0.22.23", features = ["serde"] }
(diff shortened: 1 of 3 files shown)


weihanglo (approved):
Nice. Thanks for figuring out required features for each crate!

Cargo.toml
weihanglo, line 1: There is one caveat: In Cargo's CI we currently have no job verifying each package compile separately. We might want to setup one just in case.

weihanglo: No really. The `msrv` job has done this for us already 👍🏾.

https://github.com/rust-lang/cargo/blob/68db37499f2de8acef704c73d9031be6fbcbaee4/.github/workflows/main.yml?plain=1#L283-L288

PR #15601, opened 2025-05-27
refactor: separate "global" mode from CompileMode

weihanglo:

This separates the concern of two different "mode".

- UserIntent: focus on the overall goal of the build
- CompileMode: the actual compile operation for each unit (I'd like to rename it to something else in the future, such as CompileAction)

This is a preparation of adding `-Zno-link`/`-Zlink-only` support,
which we'll have `CompileMode::Link` but that doesn't make sense to
show up in `UserIntent`.

It should have no functional change.

### Additional information


ehuss (approved):
Thanks!

src/cargo/core/compiler/build_config.rs
```diff
@@ -271,9 +251,64 @@ impl CompileMode {
...
+/// individual units, [`UserIntent`] represents the overall goal of the build
+/// process as specified by the user.
+///
+/// For example, when a user runs `cargo test`, the intent is [`UserIntent::Test`],
+/// but this might result in multiple [`CompileMode`]s for different units.
+#[derive(Clone, Copy, Debug)]
+pub enum UserIntent {
+    /// Build benchmark binaries, e.g., `cargo bench`
+    Bench,
+    /// Build binaries and libraray, e.g., `cargo run`, `cargo install`, `cargo build`.
```
bjorn3, line 274: *library

src/cargo/core/compiler/build_config.rs
```diff
@@ -167,37 +163,28 @@ pub enum MessageFormat {
...
-    /// `--test` should be passed to rustc) and by using `Test` instead it
-    /// allows some de-duping of Units to occur.
-    Bench,
-    /// A target that will be documented with `rustdoc`.
-
+    /// Document with `rustdoc`.
+    ///
     /// If `deps` is true, then it will also document all dependencies.
     /// if `json` is true, the documentation output is in json format.
     Doc { deps: bool, json: bool },
```
bjorn3, line 182: Why is the deps field on CompileMode too and not just on UserIntent?

src/cargo/core/compiler/build_config.rs
weihanglo: https://github.com/rust-lang/cargo/pull/15608

src/cargo/core/compiler/build_config.rs
weihanglo: I missed that. You're right.

weihanglo: Went ahead and removed the `deps` field as well. Thanks for the good catch!

PR #15608, opened 2025-05-29
refactor: cleanup for `CompileMode`

weihanglo:

Some more cleanup after #15601

See each commit message for details.

Help review whether it has really no behavior change.


ehuss (approved):
Thanks!

PR #15610, opened 2025-05-30
test(trim-paths): re-enable more symbols trimming check

weihanglo:

rust-lang/rust#117652 has been fixed, so re-enable those skipped.

NOT READY FOR REVIEW.

### DIFF
--- a/tests/testsuite/profile_trim_paths.rs
+++ b/tests/testsuite/profile_trim_paths.rs
@@ -592,24 +592,6 @@ fn object_works_helper(split_debuginfo: &str, run: impl Fn(&std::path::Path) -> 
             if memchr::memmem::find(line, b" OSO ").is_some() {
                 continue;
             }
-
-            // on macOS `SO` symbols are embedded in final binaries and should be trimmed.
-            // See rust-lang/rust#117652.
-            if memchr::memmem::find(line, b" SO ").is_some() {
-                continue;
-            }
-        }
-
-        #[cfg(target_os = "linux")]
-        {
-            // There is a bug in rustc `-Zremap-path-scope`.
-            // See rust-lang/rust/pull/118518
-            if memchr::memmem::find(line, b"DW_AT_comp_dir").is_some() {
-                continue;
-            }
-            if memchr::memmem::find(line, b"DW_AT_GNU_dwo_name").is_some() {
-                continue;
-            }
         }
 
         panic!(


tests/testsuite/profile_trim_paths.rs
```diff
@@ -592,24 +592,6 @@ fn object_works_helper(split_debuginfo: &str, run: impl Fn(&std::path::Path) ->
...
-        }
-
-        #[cfg(target_os = "linux")]
-        {
-            // There is a bug in rustc `-Zremap-path-scope`.
-            // See rust-lang/rust/pull/118518
-            if memchr::memmem::find(line, b"DW_AT_comp_dir").is_some() {
-                continue;
-            }
-            if memchr::memmem::find(line, b"DW_AT_GNU_dwo_name").is_some() {
```
weihanglo, line 610: When all the conditions are met:

* `split-debuginfo=packed|unpacked`, and
* `-Zremap-path-scope=object` is used,
* have at least one extern crate (dependency)

In the binary, `DW_AT_GNU_dwo_name` symbols from extern crates are not trimmed. I haven't traced down why yet.

cc @Urgau, let me know if you need a better repro.

Urgau: For the record: I tried reproducing the issue with `rustc` and failed to do. Maybe a Cargo specific thing.
```bash
$ rustc +nightly -C split-debuginfo=unpacked -g --remap-path-prefix=/tmp/split-rustc/=/remapped/path dep.rs --out-dir /tmp/split-rustc/target/ --crate-type lib -Zremap-path-scope=object
$ rustc +nightly -C split-debuginfo=unpacked -g --remap-path-prefix=/tmp/split-rustc/=/remapped/path main.rs --out-dir /tmp/split-rustc/target/ --extern=dep=/tmp/split-rustc/target/libdep.rlib -Zremap-path-scope=object
$ llvm-dwarfdump target/main | rg DW_AT_GNU_dwo_name
              DW_AT_GNU_dwo_name        ("/remapped/path/target/main.main.62a074bc4c4ee7d3-cgu.0.rcgu.dwo")
              DW_AT_GNU_dwo_name        ("/remapped/path/target/dep.dep.45866446ec9e29d2-cgu.0.rcgu.dwo")
```

Urgau: I see something interesting in the Cargo test regarding the dependency, I don't think you are remapping the `--out-dir`, which would explain why the path isn't remapped, as I assume that the `dwo`/`dwp` would be in the `target/` directory which isn't covered by those `--remap-path-prefix` (in the dependency `rustc` args).
```
    --remap-path-prefix=[ROOT]/home/.cargo/registry/src= \
    --remap-path-prefix=[..]/lib/rustlib/src/rust=/rustc/[..]
```
To me it seems like you are missing a `--remap-path-prefix` for the `build.build-dir`.

weihanglo: Really thank you very much Urgau! Excellent debugging skill!

weihanglo:
Close in favor of #15614

Issue #12137, opened 2023-05-13
Tracking Issue for trim-paths RFC 3127


ehuss:
### Summary

RFC: [#3127](https://github.com/rust-lang/rfcs/pull/3127)
Documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#profile-trim-paths-option
Issues: https://github.com/rust-lang/cargo/labels/Z-trim-paths
rustc tracking issue: https://github.com/rust-lang/rust/issues/111540

This adds a new profile setting (among other things) to control how paths are sanitized in the resulting binary.

### Unresolved Issues

* [ ] (non-blocking) Should make `--remap-path-prefix` work with `rustdoc` and doctest?
  * Yes as rustdoc supports both
    * https://github.com/rust-lang/rust/pull/107099
    * https://github.com/rust-lang/rust/pull/128736
  * [x] Integrate `--remap-path-prefix` in rustdoc invocations (`cargo doc`) https://github.com/rust-lang/cargo/pull/14389
  * [ ] Integrate `--remap-path-prefix` and `-Zremap-path-scope` in doctest (`cargo test --doc`)
* [x] Determine the remap rules: https://github.com/rust-lang/cargo/issues/13171
  * [x] A cross-platform stable hash for `CARGO_HOME` layout? https://github.com/rust-lang/cargo/pull/14917
    * [x] On Windows the local path contains prefix component e.g.`C:/`, which essentially results in a different hash result from what on other platforms. This is unlikely to be resolved.
    * [x] benchmark BLAKE3, BLAKE2s, and rustc-stable-hasher
       * See result in <https://github.com/rust-lang/cargo/pull/14116#issuecomment-2486293308>
       * rustc-stable-hash is on par with the original siphash, while blake3 is 2-3x slower. The total spent in stable hashing for a `cargo build` invocation may be unimportant. It took 500μs-1500μs for buidling cargo itself.
    * [x] This is the final remap rules: https://github.com/rust-lang/cargo/pull/17302
    * [x] To help user to unremap back, an unremap file is added as an companion files of binaries: https://github.com/rust-lang/cargo/pull/17303 
* [ ] Should workspace member paths in trimmed artifacts stay relative, or get a virtual prefix like dependencies?
  * Relative keeps the common case easy, for example, debuggers resolve workspace sources with zero configuration
[...]

PR #15614, opened 2025-05-31
fix(trim-paths): remap all paths to `build.build-dir`

weihanglo:

Remap all paths pointing to `build.build-dir`,
i.e., `[BUILD_DIR]/debug/deps/foo-[HASH].dwo` would be remapped to
`/cargo/build-dir/debug/deps/foo-[HASH].dwo`
(note the `/cargo/build-dir` prefix).

This covers scenarios like:

* Build script generated code. For example, a build script may call `file!`
  macros, and the associated crate uses `include!` to include the expanded
  `file!` macro in-place via the `OUT_DIR` environment.
* On Linux, `DW_AT_GNU_dwo_name` that contains paths to split debuginfo
  files (dwp and dwo).

Should be quite straightforward.

The open question is what we want to remap _to_, to help debugger to find the source files.
cc #12137 and #13171

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1393,6 +1393,7 @@ fn trim_paths_args_rustdoc(
     // Order of `--remap-path-prefix` flags is important for `-Zbuild-std`.
     // We want to show `/rustc/<hash>/library/std` instead of `std-0.0.0`.
     cmd.arg(package_remap(build_runner, unit));
+    cmd.arg(build_dir_remap(build_runner));
     cmd.arg(sysroot_remap(build_runner, unit));
 
     Ok(())
@@ -1420,6 +1421,7 @@ fn trim_paths_args(
     // Order of `--remap-path-prefix` flags is important for `-Zbuild-std`.
     // We want to show `/rustc/<hash>/library/std` instead of `std-0.0.0`.
     cmd.arg(package_remap(build_runner, unit));
+    cmd.arg(build_dir_remap(build_runner));
     cmd.arg(sysroot_remap(build_runner, unit));
 
     Ok(())
@@ -1490,6 +1492,26 @@ fn package_remap(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> OsString {
         remap.push("-");
         remap.push(unit.pkg.version().to_string());
     }
+    remap
+}
+
+/// Remap all paths pointing to `build.build-dir`,
+/// i.e., `[BUILD_DIR]/debug/deps/foo-[HASH].dwo` would be remapped to
+/// `/cargo/build-dir/debug/deps/foo-[HASH].dwo`
+/// (note the `/cargo/build-dir` prefix).
+///
+/// This covers scenarios like:
+///
+/// * Build script generated code. For example, a build script may call `file!`
+///   macros, and the associated crate uses [`include!`] to include the expanded
+///   [`file!`] macro in-place via the `OUT_DIR` environment.
+/// * On Linux, `DW_AT_GNU_dwo_name` that contains paths to split debuginfo
+///   files (dwp and dwo).
+fn build_dir_remap(build_runner: &BuildRunner<'_, '_>) -> OsString {
+    let build_dir = build_runner.bcx.ws.build_dir();
+    let mut remap = OsString::from("--remap-path-prefix=");
+    remap.push(build_dir.as_path_unlocked());
+    remap.push("=/cargo/build-dir");
     remap
 }
 
--- a/tests/testsuite/profile_trim_paths.rs
+++ b/tests/testsuite/profile_trim_paths.rs
@@ -247,6 +247,78 @@ fn registry_dependency() {
 }
 
 #[cargo_test(nightly, reason = "-Zremap-path-scope is unstable")]
+fn registry_dependency_with_build_script_codegen() {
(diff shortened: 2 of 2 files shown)


tests/testsuite/profile_trim_paths.rs
```diff
@@ -592,24 +664,6 @@ fn object_works_helper(split_debuginfo: &str, run: impl Fn(&std::path::Path) ->
...
-        }
-
-        #[cfg(target_os = "linux")]
-        {
-            // There is a bug in rustc `-Zremap-path-scope`.
-            // See rust-lang/rust/pull/118518
-            if memchr::memmem::find(line, b"DW_AT_comp_dir").is_some() {
-                continue;
-            }
-            if memchr::memmem::find(line, b"DW_AT_GNU_dwo_name").is_some() {
```
weihanglo, line 638: The root cause of the `DW_AT_GNU_dwo_name` case has been found in <https://github.com/rust-lang/cargo/pull/15610>, and should be fixed altogether by this PR via remapping `build.build-dir`.

weihanglo: Originally posted by @Urgau in <https://github.com/rust-lang/cargo/pull/15610#discussion_r2119070173> and <https://github.com/rust-lang/cargo/pull/15610#discussion_r2119070640>:

> For the record: I tried reproducing the issue with `rustc` and failed to do. Maybe a Cargo specific thing.
> ```bash
> $ rustc +nightly -C split-debuginfo=unpacked -g --remap-path-prefix=/tmp/split-rustc/=/remapped/path dep.rs --out-dir /tmp/split-rustc/target/ --crate-type lib -Zremap-path-scope=object
> $ rustc +nightly -C split-debuginfo=unpacked -g --remap-path-prefix=/tmp/split-rustc/=/remapped/path main.rs --out-dir /tmp/split-rustc/target/ --extern=dep=/tmp/split-rustc/target/libdep.rlib -Zremap-path-scope=object
> $ llvm-dwarfdump target/main | rg DW_AT_GNU_dwo_name
>               DW_AT_GNU_dwo_name        ("/remapped/path/target/main.main.62a074bc4c4ee7d3-cgu.0.rcgu.dwo")
>               DW_AT_GNU_dwo_name        ("/remapped/path/target/dep.dep.45866446ec9e29d2-cgu.0.rcgu.dwo")
> ```

> I see something interesting in the Cargo test regarding the dependency, I don't think you are remapping the `--out-dir`, which would explain why the path isn't remapped, as I assume that the `dwo`/`dwp` would be in the `target/` directory which isn't covered by those `--remap-path-prefix` (in the dependency `rustc` args).
> ```
>     --remap-path-prefix=[ROOT]/home/.cargo/registry/src= \
>     --remap-path-prefix=[..]/lib/rustlib/src/rust=/rustc/[..]
> ```
> To me it seems like you are missing a `--remap-path-prefix` for the `build.build-dir`.

src/cargo/core/compiler/mod.rs
```diff
@@ -1493,6 +1495,25 @@ fn package_remap(build_runner: &BuildRunner<'_, '_>, unit: &Unit) -> OsString {
     remap
 }
 
+/// Remap all paths pointing to `build.build-dir`,
+/// i.e., `[BUILD_DIR]/debug/deps/foo-[HASH].dwo` would be remapped to
+/// `debug/deps/foo-[HASH].dwo`.
```
epage: > The open question is what we want to remap to, to help debugger to find the source files.

If we could assume things were in `target` then we could just add that as a prefix and make this like the pkg root remapping.  People can match on that and remap it in their debugger.

As we move the build dir, this will be less likely to be a match.  This makes me wonder if we should be forward looking and name it something more contextually meaningful like `build-dir`.

weihanglo: What about a `/cargo/build-dir` prefix. We don't need to make the final call right now. This eventually needs to re-review before stabilization.

We might probably want to do that for other kinds of dependencies.

epage: Whatever seems to work best.  I mostly feel we need some kind of prefix

tests/testsuite/profile_trim_paths.rs
```diff
@@ -297,7 +297,7 @@ fn registry_dependency_with_build_script_codegen() {
         .masquerade_as_nightly_cargo(&["-Ztrim-paths"])
         // Macros should be sanitized
         .with_stdout_data(str![[r#"
-[ROOT]/foo/target/debug/build/bar-[HASH]/out/bindings.rs
+/cargo/build-dir/debug/build/bar-[HASH]/out/bindings.rs
```
epage, line 300: Regarding build reproducibility, is `bar-[HASH]` reproducible?

weihanglo: https://github.com/rust-lang/cargo/blob/b646f8393543aa9dfd2b8382b0de10961dad6bc5/src/cargo/core/compiler/build_runner/compilation_files.rs?plain=1#L237-L245

It is `pkg_dir`, and both target_short_hash and c_extra_filename use rustc-stable-hash. I believe it is reproducible.

weihanglo:
@teknalb this should fix the issue you mentioned in <https://github.com/rust-lang/rust/issues/111540#issuecomment-2524209170>. It is not yet available in nightly, though.

PR #15617, opened 2025-06-01
fix(fingerprint): explicit reason than "stale; unknown reason"

weihanglo:

This was discovered during playing with rmeta reuse between cargo-check and cargo-build.

An rmeta mtime failure is basically the same as "failed to read metadata".

While we don't report in tracing log, that should be fine because dirty reason will be reported in the other place.

### DIFF
--- a/src/cargo/core/compiler/fingerprint/mod.rs
+++ b/src/cargo/core/compiler/fingerprint/mod.rs
@@ -1167,15 +1167,12 @@ impl Fingerprint {
         // minimum mtime as it's the one we'll be comparing to inputs and
         // dependencies.
         for output in self.outputs.iter() {
-            let mtime = match paths::mtime(output) {
-                Ok(mtime) => mtime,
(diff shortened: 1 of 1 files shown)


epage (approved):

weihanglo:
https://github.com/rust-lang/cargo/actions/runs/15396140671/job/43317289386#step:13:4554

Network glitch?

```
 > [internal] load metadata for docker.io/library/alpine:3.22:
------
Dockerfile:1
--------------------
   1 | >>> FROM alpine:3.22
   2 |     
   3 |     RUN apk add --no-cache openssh git
--------------------
ERROR: failed to solve: alpine:3.22: failed to resolve source metadata for docker.io/library/alpine:3.22: failed to authorize: failed to fetch oauth token: unexpected status from POST request to https://auth.docker.io/token: 500 Internal Server Error
```

Going to re-queue

Issue #5042, opened 2018-02-15
Nested workspaces


nrc:
Allow multiple layers of workspaces, not just one. This would be useful for tools in the Rust repo, and I think more generally.

nrc:
A lightweight way to help here might be to just ignore nested virtual manifests. If we want to be strict, we could do this only if the outer manifest includes all the workspace members in the inner manifest. This might admit some errors where there is an ignored Cargo.toml somewhere in a tree of workspaces, but I don't think that is much of a hazard, especially if the outer manifest explicitly includes the inner members.

One question is forwards compatibility, if one day we want to allow true nested workspaces, then I'm not sure how well that co-exists with this simpler proposal. I think technically it would be OK, but it might be hard to understand for users.

cc @rust-lang/cargo - thoughts? If this sounds OK, then I'd quite like to implement so that we can make Rustfmt and maybe RLS use workspaces (and even Cargo!).

matklad:
This is already supported with the help of `exclude` key I think?

```
λ pwd
/home/matklad/trash/foo

~/trash/foo master*
λ tree .
.
├── bar
│   ├── Cargo.toml
│   └── src
│       └── lib.rs
├── Cargo.lock
├── Cargo.toml
└── src
    └── lib.rs

3 directories, 5 files

~/trash/foo master*
λ cat Cargo.toml
[package]
name = "foo"
version = "0.1.0"
authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]

[workspace]
exclude = [ "./bar" ]

[dependencies]
bar = { path = "./bar" }

~/trash/foo master*
λ cat bar/Cargo.toml
[package]
name = "bar"
version = "0.1.0"
authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]

[workspace]

[dependencies]

~/trash/foo master*
λ cargo build
   Compiling bar v0.1.0 (file:///home/matklad/trash/foo/bar)
   Compiling foo v0.1.0 (file:///home/matklad/trash/foo)
    Finished dev [unoptimized + debuginfo] target(s) in 0.55 secs

```

nrc:
I don't think the exclude trick quite works for the case I have in mind. I have:

```
.
├── Cargo.toml
├── bar
│   ├── Cargo.toml
│   └── baz
│       ├── Cargo.toml
│       └── src
│           └── lib.rs
└── foo
    ├── Cargo.toml
    └── src
        └── lib.rs
```

[...]

PR #15620, opened 2025-06-02
[Pre-RFC] Enable `.cargo/root`, `CARGO_ROOTS` + `--root` to limit config + manifest searches

### LINKED ISSUE #5418: override "current package believes it's in a workspace"
If you create a crate inside another one, sometimes you get the above quoted error when trying to build it.

As far as I can tell, there's no solution. You can never build the crate because it's not really in the workspace, yet you can't build it separately either. The only recourse is to move the whole thing to another directory.

I propose a new flag, called something like `--no-workspace` or `--believe-nothing--reality-is-fake--just-build-here`, that would ignore the surrounding workspace and build the crate as it would have done before workspaces were invented. Does this make sense?

### LINKED ISSUE #7871: Put Searching Parent Directories for `Cargo.toml` Behind Logical Gate
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/issues/7871)*

See also rust-lang/rfcs#3279

For `.cargo/config.toml` side of this, see #7621

---

In [`cargo-outdated`](https://github.com/kbknapp/cargo-outdated) we are seeing an [issue](https://github.com/kbknapp/cargo-outdated/issues/194) (we build the project under `/tmp`) where if there's a `cargo.toml` that was moved to `/tmp/` cargo build will fail.

Either a seperate `find_root()` ([link](https://github.com/rust-lang/cargo/blob/master/src/cargo/core/workspace.rs#L385))that we can trigger from `workspace::new()`, or a config option to stop the function from crawling upwards ([link](https://github.com/rust-lang/cargo/blob/master/src/cargo/core/workspace.rs#L417)) in the directory ancestors. 

**Notes**
This was mentioned in https://github.com/rust-lang/cargo/issues/4992#issuecomment-401129850 

I understand this is necessary for cargo and it's the intended function; however, using cargo as an API there should be a way to avoid this :)

rib:
_This is a draft proposal that I think could address a number of issues in Cargo (e.g. #5418, #7871) and could be an alternative (or complementary) approach to addressing the safety issues discussed in https://github.com/rust-lang/rfcs/pull/3279_

_The intention here is to prototype a practical solution that can then hopefully inform any follow up discussions._

_My current [motivation](https://github.com/rust-lang/cargo/issues/5418#issuecomment-2886497944) for looking at this comes from using Cargo as part of a [rustdoc-markdown](https://github.com/Dooyo-Labs/rustdoc-markdown) toolchain that should be able to build third-party crates without worrying about side effects from unrelated configs or workspace manifests in parent directories._

--

This adds the notion of root directories to Cargo, which act as limits to how far Cargo  will search ancestors looking for `.cargo` config files or `Cargo.toml` manifests (instead of walking to the root of the filesystem).

This has some similarity to Git's `GIT_CEILING_DIRECTORIES` mechanism.

There are three notable searches that root directories affect:
1. How far should Cargo search when loading `.cargo/Config.toml` files
2. How far should Cargo search from when loading a `Cargo.toml` package manifest
3. How far should Cargo search from a package manifest directory looking for a `Cargo.toml` workspace manifest.

Roots can be specified in three ways:

1. By setting the `CARGO_ROOTS` environment variable to a colon (semicolon on Windows) separated list of directories
4. By passing `--root` on the command line (can be passed multiple times)
   i. Suited to tool configurations where appending to environment variables may be limited (e.g. limited environment variable interpolation in vscode settings)
5. The existence of a `.cargo/root` file (discovered by checking parents up towards the root of the filesystem)
    i. Suited to sandboxing directories when the absolute path might vary, such in within monorepos that could be cloned to different locations or directories that may get mounted in Docker containers.

By default, if `CARGO_ROOTS` has not been set then the user's home directory is implicitly added as a root directory.
- This is for safer default behaviour that stops Cargo from loading configs from shared directories like `/tmp/.cargo/`, or `C:/.cargo` or searching for workspace manifests.

The default home-directory root also mitigates issues with triggering automounters on some systems.

Whenever Cargo needs to search ancestors for files it will resolve the closest root directory that is an ancestor of the starting directory and that root is the last directory Cargo will search.

If the starting directory is not within any root then Cargo will only search the starting directory. Cargo will also print a warning to indicate that it's loading of configs + workspace manifests is limited.

_As a safety / ergonomics (+ backwards compatibility) trade-off, there is a special case that allows Cargo to search up to the root of the filesystem when searching for the first `Cargo.toml` package manifest, IFF the starting directory is outside of all roots. This enables a package to be unpacked under a shared directory like `/tmp/my-package` and start a build from `/tmp/my-package/sub/dir` which will load `/tmp/my-package/Cargo.toml` but will _not_ then allow Cargo to attempt loading a workspace at `/tmp/Cargo.toml`._

Fixes: #5418
- For use cases where the workaround described in #6805 isn't appropriate (to add an empty workspace to nested packages)

Fixes: #7871
  - Addresses safety issues with building under shared directories like `/tmp`
  - Mitigates triggering home directory automounters

Addresses: https://github.com/rust-lang/rfcs/pull/3279

This gives more flexibility for tools that want to use cargo as an implementation detail. In particular this allows you to sandbox the build of nested third-party workspaces that may be (unknowingly) dynamically unpacked within an outer workspace, in situations where neither the workspace being built and the outer workspace are owned by the tool that is managing the build.

This does not attempt to support nesting workspaces with inheritence (as described in #5042).

For example a tool based on rustdoc-json should be able to fetch and build documentation for third-party crates under any user-specified build/target directory without having to worry about spooky action at a distance due to config files and workspaces in ancestor directories.

In my case, I have an runtime for coding with LLMs that is given a repo to work on and is expected to keep its artifacts contained to a `.ai/` directory. This runtime supports building markdown documentation for Rust crates, which involves using cargo to generate rustdoc-json data. That tool is expected to keep its artifacts contained within `.ai/docs/rust/build/` within the project being worked on. It's possible that the project itself is Rust based and could define a workspace or `.cargo/config.toml` but from the pov of this toolchain those have nothing to do with the crate whose documentation is being generated (which are packages downloaded from crates.io).

_**TODO**: write tests_

Root directories can be smoke tested by attempting to build any package that doesn't define a workspace, nested within a repo that does, E.g.:

```bash
mkdir cargo-root-test
cd cargo-root-test

# Clone anything that defines an outer workspace...
git clone https://github.com/serde-rs/serde.git

# Then clone a nested repo that doesn't have a workspace
cd serde
git clone https://github.com/meta-rust/rust-hello-world

# Attempting to build will lead to an error like "error: current package believes it's in a workspace when it's not"
cargo build

# But if we know that the outer workspaces is unrelated we can specify a root directory and isolate the build...
cargo --root . build
# Or
env CARGO_ROOTS="$PWD" cargo build
# Or
mkdir .cargo
touch .cargo/root
cargo build
```

### Known Issues

The current implementation makes heavy use of `std::fs::canonicalize` while the original expectation had been that the UNC prefixes for Windows wouldn't matter when just internally checking paths for equivalence.

As things have evolved I think the use of `canonicalize` should be replaced with a normalization that just maps root directories to absolute paths and uses Cargo's `normalize_path` utility. This way we would avoid resolving symlinks and rely on lexical path comparisons.

Ideally the normalized paths can even be kept as a hidden implementation detail when resolving a `SearchRoute` that would allow the following use of `path::ancestors` to iterate in accordance with any non-normalized starting directory.

### DIFF
--- a/benches/benchsuite/benches/global_cache_tracker.rs
+++ b/benches/benchsuite/benches/global_cache_tracker.rs
@@ -35,7 +35,7 @@ fn initialize_context() -> GlobalContext {
     let cwd = homedir.clone();
     let mut gctx = GlobalContext::new(shell, cwd, homedir);
     gctx.nightly_features_allowed = true;
-    gctx.set_search_stop_path(root());
+    gctx.set_config_search_root(root());
     gctx.configure(
         0,
         false,
--- a/crates/cargo-util/src/paths.rs
+++ b/crates/cargo-util/src/paths.rs
@@ -441,14 +441,14 @@ pub struct PathAncestors<'a> {
 
 impl<'a> PathAncestors<'a> {
     fn new(path: &'a Path, stop_root_at: Option<&Path>) -> PathAncestors<'a> {
-        let stop_at = env::var("__CARGO_TEST_ROOT")
-            .ok()
-            .map(PathBuf::from)
-            .or_else(|| stop_root_at.map(|p| p.to_path_buf()));
+        tracing::trace!(
+            "creating path ancestors iterator from `{}` to `{}`",
+            path.display(),
+            stop_root_at.map_or("<root>".to_string(), |p| p.display().to_string())
+        );
         PathAncestors {
             current: Some(path),
-            //HACK: avoid reading `~/.cargo/config` when testing Cargo itself.
-            stop_at,
+            stop_at: stop_root_at.map(|p| p.to_path_buf()),
         }
     }
 }
@@ -466,6 +466,7 @@ impl<'a> Iterator for PathAncestors<'a> {
                 }
             }
 
+            tracing::trace!("next path ancestor: `{}`", path.display());
             Some(path)
         } else {
             None
--- a/src/bin/cargo/cli.rs
+++ b/src/bin/cargo/cli.rs
@@ -2,6 +2,7 @@ use anyhow::{anyhow, Context as _};
 use cargo::core::{features, CliUnstable};
 use cargo::util::context::TermConfig;
 use cargo::{drop_print, drop_println, CargoResult};
+use cargo_util::paths;
 use clap::builder::UnknownArgumentValueParser;
 use itertools::Itertools;
 use std::collections::HashMap;
@@ -17,6 +18,34 @@ use crate::util::is_rustup;
 use cargo::core::shell::ColorChoice;
 use cargo::util::style;
 
+fn closest_valid_root<'a>(
+    cwd: &std::path::Path,
+    config_root: Option<&'a std::path::Path>,
+    env_root: Option<&'a std::path::Path>,
+    cli_root: Option<&'a std::path::Path>,
+) -> anyhow::Result<Option<&'a std::path::Path>> {
+    for (name, root) in [
+        (".cargo/root", config_root),
+        ("CARGO_ROOT", env_root),
+        ("--root", cli_root),
+    ] {
+        if let Some(root) = root {
+            if !cwd.starts_with(root) {
+                return Err(anyhow::format_err!(
+                    "the {} `{}` is not a parent of the current working directory `{}`",
+                    name,
+                    root.display(),
+                    cwd.display()
+                ));
+            }
+        }
+    }
+    Ok([config_root, env_root, cli_root]
+        .into_iter()
+        .flatten()
+        .max_by_key(|root| root.components().count()))
+}
+
 #[tracing::instrument(skip_all)]
 pub fn main(gctx: &mut GlobalContext) -> CliResult {
     // CAUTION: Be careful with using `config` until it is configured below.
@@ -25,6 +54,7 @@ pub fn main(gctx: &mut GlobalContext) -> CliResult {
 
     let args = cli(gctx).try_get_matches()?;
 
+    let mut need_reload = false;
     // Update the process-level notion of cwd
     if let Some(new_cwd) = args.get_one::<std::path::PathBuf>("directory") {
         // This is a temporary hack.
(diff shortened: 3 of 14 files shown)


epage:
> I guess something like this might need a [Pre-]RFC?, but as I'm https://github.com/rust-lang/cargo/issues/5418#issuecomment-2886497944 with nested workspaces while using Cargo as part of a [rustdoc-markdown](https://github.com/Dooyo-Labs/rustdoc-markdown) toolchain, I figured I'd go ahead and take a stab at drafting a solution that could at least work for my use case, to get some practical insight into the problem.

Personally, I do not look at PRs until a design is agreed on.  Having design discussions here splits the conversation and makes it harder to follow.  This is also feature with a lot of potential workflow considerations and I would personally want to see a more deliberate approach, like an RFC,. to make sure we understand what problems we are trying to solve, how this has been solved before, and how this will impact people.

FYI there is also discussion at #7871 and this is somewhat related to rust-lang/rfcs#3279

rib:
> > I guess something like this might need a [Pre-]RFC?, but as I'm [#5418 (comment)](https://github.com/rust-lang/cargo/issues/5418#issuecomment-2886497944) with nested workspaces while using Cargo as part of a [rustdoc-markdown](https://github.com/Dooyo-Labs/rustdoc-markdown) toolchain, I figured I'd go ahead and take a stab at drafting a solution that could at least work for my use case, to get some practical insight into the problem.
> 
> Personally, I do not look at PRs until a design is agreed on. Having design discussions here splits the conversation and makes it harder to follow. This is also feature with a lot of potential workflow considerations and I would personally want to see a more deliberate approach, like an RFC,. to make sure we understand what problems we are trying to solve, how this has been solved before, and how this will impact people.
> 
> FYI there is also discussion at #7871 and this is somewhat related to [rust-lang/rfcs#3279](https://github.com/rust-lang/rfcs/pull/3279)

Right, I at least like to write some code, where possible, as a proof of concept / point of reference before getting into design discussions to try and avoid getting too academic, but it makes sense to avoid fragmented design discussions here, beyond linking to existing discussions or perhaps giving pointers for framing a new RFC.

Thanks for the links  - I hadn't come across those discussions from searching myself.

tarekk22-afk (approved):

rib:
While I've been looking at some of the other related issues in #7871 (including issues with `/tmp/Config.toml` or `/tmp/.cargo/Config.toml` files or triggering home directory automounters) and the safety issues discussed in https://github.com/rust-lang/rfcs/pull/3279 I've taken another pass at this prototype and updated the description accordingly.

I'll try and find some time to follow up on the discussions for those separately, but I'm currently optimistic that something close to this draft could be a viable solution for several of these issues.

tarekk22-afk (approved):

tarekk22-afk (approved):

0xPoe:
Thank you for working on this!

Since this PR has been stale for about a year and has many conflicts, I would like to close it. If you want to rework this PR, as suggested in https://github.com/rust-lang/cargo/pull/15620#issuecomment-2932236162, we could restart from the discussions.

PR #15621, opened 2025-06-02
test(trim-paths): enable more tests for windows-msvc

weihanglo:

Before this some end-to-end tests didn't cover Windows platforms.

After this, we cover windows-msvc for 

* End-to-end debugger tests.
* Check path is trimmed with symbol viewers like `strings`.

windows-gnu isn't covered

There are things needing attentions:

* This adds a new CI job for window-msvc "nightly" toolchain.
* In 2f923b3ff25f847d we don't check if an executable's availability by running `<cmd> --version`. Instead, we check the file execute bit.
* Enabled windows-msvc tests rely on the software provided by [GitHub windows runner image](https://github.com/actions/runner-images/blob/e330e24b7e154207c9d76ce3b5472bfc6ceef55f/images/windows/Windows2022-Readme.md)
  * Windows SDK which provides cdb and other debugger tools
  * `strings` is provided by MinGW.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -155,6 +155,10 @@ jobs:
           os: windows-latest
           rust: stable-msvc
           other: i686-pc-windows-msvc
+        - name: Windows x86_64 MSVC nightly
+          os: windows-latest
+          rust: nightly-msvc
+          other: i686-pc-windows-msvc
         - name: Windows x86_64 gnu nightly # runs out of space while trying to link the test suite
           os: windows-latest
           rust: nightly-gnu
@@ -179,6 +183,10 @@ jobs:
     - run: sudo apt update -y && sudo apt install lldb gcc-multilib libsecret-1-0 libsecret-1-dev -y
       if: matrix.os == 'ubuntu-latest'
     - run: rustup component add rustfmt || echo "rustfmt not available"
+    - name: Add Windows debuggers bin to PATH
+      shell: pwsh
+      run: Add-Content $env:GITHUB_PATH "C:\Program Files (x86)\Windows Kits\10\Debuggers\x64"
+      if: matrix.os == 'windows-latest'
     - name: Configure extra test environment
       run: echo CARGO_CONTAINER_TESTS=1 >> $GITHUB_ENV
       if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -155,6 +155,10 @@ jobs:
           os: windows-latest
           rust: stable-msvc
           other: i686-pc-windows-msvc
+        - name: Windows x86_64 MSVC nightly
```
weihanglo, line 158: WARN: A new job is added

crates/cargo-test-macro/src/lib.rs
```diff
@@ -302,7 +302,37 @@ fn check_command(command_path: &Path, args: &[&str]) -> bool {
...
+fn is_executable<P: AsRef<Path>>(path: P) -> bool {
+    use std::os::unix::prelude::*;
+    std::fs::metadata(path)
+        .map(|metadata| metadata.is_file() && metadata.permissions().mode() & 0o111 != 0)
+        .unwrap_or(false)
+}
+
+#[cfg(windows)]
+fn is_executable<P: AsRef<Path>>(path: P) -> bool {
+    path.as_ref().is_file()
```
epage: `which` seems to do a bit more but this is likely good enough for us.

crates/cargo-test-macro/src/lib.rs
```diff
@@ -302,7 +302,37 @@ fn check_command(command_path: &Path, args: &[&str]) -> bool {
 }
 
 fn has_command(command: &str) -> bool {
-    check_command(Path::new(command), &["--version"])
+    use std::env::consts::EXE_EXTENSION;
+    let Some(paths) = std::env::var_os("PATH") else {
+        return false;
+    };
+    // ALLOWED: For testing cargo itself only.
+    #[allow(clippy::disallowed_methods)]
```
epage: Thought so, CI says this is on the wrong line.  It should be on `var_os`

crates/cargo-test-macro/src/lib.rs
weihanglo: That is true. See this thread <https://github.com/rust-lang/rustup/pull/4324#discussion_r2079649252> as a reference.

epage (approved):

crates/cargo-test-macro/src/lib.rs
weihanglo: 🤦🏾‍♂️

Fixed!

PR #15623, opened 2025-06-03
Add --offline for comp

A4-Tacks:

Bash and Zsh `cargo add` completion option `--offline`

### DIFF
--- a/src/etc/_cargo
+++ b/src/etc/_cargo
@@ -75,6 +75,7 @@ _cargo() {
                         {-F+,--features=}'[specify features to activate]:feature' \
                         "--default-features[enable the default features]" \
                         "--no-default-features[don't enable the default features]" \
+                        "--offline[run without accessing the network]" \
                         "--optional[mark the dependency as optional]" \
                         "--no-optional[mark the dependency as required]" \
                         "--dev[add as a dev dependency]" \
--- a/src/etc/_cargo
+++ b/src/etc/_cargo
@@ -12,6 +12,7 @@ _cargo() {
         '(-q --quiet)*'{-v,--verbose}'[use verbose output]'
         '(-q --quiet -v --verbose)'{-q,--quiet}'[no output printed to stdout]'
         '-Z+[pass unstable (nightly-only) flags to cargo]: :_cargo_unstable_flags'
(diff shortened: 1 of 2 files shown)


src/etc/cargo.bashcomp.sh
```diff
@@ -49,7 +49,7 @@ _cargo()
 	local opt_targets="--lib --bin --bins --example --examples --test --tests --bench --benches --all-targets"
 
 	local opt___nocmd="$opt_common -V --version --list --explain"
-	local opt__add="$opt_common -p --package --features --default-features --no-default-features $opt_mani --optional --no-optional --rename --dry-run --path --git --branch --tag --rev --registry --dev --build --target --ignore-rust-version"
+	local opt__add="$opt_common -p --package --features --default-features --no-default-features $opt_mani --offline --optional --no-optional --rename --dry-run --path --git --branch --tag --rev --registry --dev --build --target --ignore-rust-version"
```
ehuss: I think this should be opt_locked instead. Maybe something like:
```suggestion
	local opt__add="$opt_common $opt_lock -p --package --features --default-features --no-default-features $opt_mani --optional --no-optional --rename --dry-run --path --git --branch --tag --rev --registry --dev --build --target --ignore-rust-version"
```

src/etc/_cargo
```diff
@@ -75,6 +75,7 @@ _cargo() {
                         {-F+,--features=}'[specify features to activate]:feature' \
                         "--default-features[enable the default features]" \
                         "--no-default-features[don't enable the default features]" \
+                        "--offline[run without accessing the network]" \
```
ehuss: Shouldn't this be added to all commands (that use it)? Probably something similar to how bash does it.

src/etc/_cargo
weihanglo, line 1: BTW, are you aware of [native-completions](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#native-completions)? It aims to provide a better shell completion story, and doesn't require any extra shell completion framework.

src/etc/_cargo
A4-Tacks: ok

PR #15626, opened 2025-06-04
fix(package): Allow packaging of self-cycles with -Zpackage-workspace

epage:

Found this as I was preparing for stabilization.

As we create a graph of packages to ensure publish order, a self-cycle shouldn't matter, so we can skip tracking it.

### DIFF
--- a/src/cargo/ops/cargo_package/mod.rs
+++ b/src/cargo/ops/cargo_package/mod.rs
@@ -417,6 +417,11 @@ fn local_deps<T>(packages: impl Iterator<Item = (Package, T)>) -> LocalDependenc
                 continue;
             };
 
+            // We don't care about cycles
+            if dep.source_id() == pkg.package_id().source_id() {
+                continue;
+            }
+
             if let Some(dep_pkg) = source_to_pkg.get(&dep.source_id()) {
                 graph.link(pkg.package_id(), *dep_pkg);
             }
--- a/tests/testsuite/package.rs
+++ b/tests/testsuite/package.rs
@@ -7560,6 +7560,40 @@ fn unpublished_cyclic_dev_dependencies() {
     );
 }
 
+#[cargo_test]
+fn unpublished_cyclic_dev_dependencies_nightly() {
+    registry::init();
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                [package]
+                name = "foo"
+                version = "0.0.1"
+                edition = "2015"
+                authors = []
+                license = "MIT"
+                description = "foo"
+                documentation = "foo"
+
+                [dev-dependencies]
+                foo = { path = ".", version = "0.0.1" }
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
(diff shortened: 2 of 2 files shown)


tests/testsuite/package.rs
```diff
@@ -7560,6 +7560,40 @@ fn unpublished_cyclic_dev_dependencies() {
     );
 }
 
+#[cargo_test]
+fn unpublished_cyclic_dev_dependencies_nightly() {
```
weihanglo, line 7564: Do we expect to rename or remove this test once it is stabilized? Could we leave a comment for that?

tests/testsuite/package.rs
```diff
@@ -7560,6 +7560,40 @@ fn unpublished_cyclic_dev_dependencies() {
...
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("package --no-verify --exclude-lockfile -Zpackage-workspace")
+        .masquerade_as_nightly_cargo(&["package-workspace"])
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] crates-io is replaced with remote registry dummy-registry;
```
weihanglo: This error looks pretty weird. What is the actual error?

weihanglo: I cannot reproduce this error with `cargo 1.89.0-nightly (64a124607 2025-05-30)` outside cargo-test-support (not setting `CARGO_HOME`).

tests/testsuite/package.rs
epage: I have a stabilization branch locally. Also, I copied this from an existing set of tests I updated that I found made this easier.

tests/testsuite/package.rs
epage: You need to replace crates.io

weihanglo: Yeah I know, but is there a real world repro?

src/cargo/ops/cargo_package/mod.rs
```diff
@@ -417,6 +417,11 @@ fn local_deps<T>(packages: impl Iterator<Item = (Package, T)>) -> LocalDependenc
                 continue;
             };
 
+            // We don't care about cycles
+            if dep.source_id() == pkg.package_id().source_id() {
```
weihanglo, line 421: It is a bit unintuitive, as comparing source_id doesn't generally imply they are cyclic. However, given we filtered in local packages earlier (which source ID contains path to package root as the unique identifer), it makes sense.

weihanglo (approved):
There is no blocker and the fix seems reasonable. Thank you!

src/cargo/ops/cargo_package/mod.rs
epage: Yeah, in general this code assumes that the `source_id` is unique and looks up a package id by the source id.    I could move this into the following `if` so we can then compare package ids.  That won't make a practical differende but maybe it makes the intent clearer?

PR #15629, opened 2025-06-04
fix(package): Skip registry check if its not needed

epage:

As of #13947, `-Zpackage-workspace` made `cargo package` require extra arguments when it didn't need it before.  When a packaging operation will resolve dependencies, we need to know what registry packages would be published to in order to correctly generate the overlay to generate the right lockfile.

We skipped this if there were no dependencies, so no overlay was going to be used, to reduce the impact of this.

This change goes a step further and only runs the check if the resolver will run.  This should mean that `-Zpackage-workspace` should now only error when `cargo package` would have failed before.

To verify this, the existing failure tests were forked, removing `-Zpackage-workspace`, and then `--exclude-lockfile`, `--no-verify`, and `--exclude-lockfile --no-verify` variants were added to characterize when the check is behavior-neutral vs not needed and that the behavior is now the same.

### DIFF
--- a/src/cargo/ops/cargo_package/mod.rs
+++ b/src/cargo/ops/cargo_package/mod.rs
@@ -240,10 +240,13 @@ fn do_package<'a>(
     let just_pkgs: Vec<_> = pkgs.iter().map(|p| p.0).collect();
 
     let mut local_reg = if ws.gctx().cli_unstable().package_workspace {
-        // The publish registry doesn't matter unless there are local dependencies,
+        // The publish registry doesn't matter unless there are local dependencies that will be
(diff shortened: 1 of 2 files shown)


epage:
This does not reduce any registry updates.  This bypasses us creating an overlay registry when it will go unused.  The creation of an overlay registry, in some cases, requires the user to pass in an extra command line flag.  With us not needing that extra command line flag, stabilization of `-Zpackage-workspace` will mean that  a success case won't be turned into an error case.

weihanglo (approved):
Thanks. I misunderstood the PR.

Issue #14903, opened 2018-04-09
Tracking issue for RFC 2196, "metabuild: semantic build scripts for Cargo"


Centril:
This is a tracking issue for the RFC "metabuild: semantic build scripts for Cargo" (rust-lang/rfcs#2196).

**Steps:**

- [x] Implement the RFC (#5628)
- [ ] Implement Multiple Build Scripts
  - [x] Parsing #15630
  - [x] Running multiple build scripts #15704 
- [ ] Implement Build Script Parameter Passing
- [ ] Implement Build Script Delegation
- [ ] Adjust documentation ([see instructions on forge][doc-guide])
- [ ] Stabilization PR ([see instructions on forge][stabilization-guide])

[stabilization-guide]: https://forge.rust-lang.org/stabilization-guide.html
[doc-guide]: https://forge.rust-lang.org/stabilization-guide.html#updating-documentation

**Unresolved questions:**

- `OUT_DIR`: how to read `OUT_DIR` when there are multiple build scripts was deferred out of #15704
- Conflicts between build scripts: if two build scripts write the same env variable, etc, what should happen?  As of #15704, the last one alphabetically wins (except sometimes the first alphabetically wins).  We need this consistent and users might need control over this.  So long as its build scripts, the user can change names to control order. However, when we get to build script delegation, that won't work.

epage:
Since this RFC was created...

We now have bindeps / artifact deps and I feel that pull in metabuild dependencies as artifact deps, rather than rlibs, would be better as it allows a shared build script to be compiled once.  Yes, the rlib gets compiled once but we then still have to have the wrapper pull it in and that can be significant, depending on the profile settings.

You can now `use dep::main;` to define a `main` function.

We have a request for being able to run unit tests (#9942).  If build scripts were dedicated packages in a workspace, then they would get that "for free".  Granted, this then requires publishing the build script package.

There is more scrutiny of supply chains and auditing of code that gets run during builds (e.g. #5720 is talked about frequently in the community).  This allowing us to only audit a common set of shared build scripts and not even wrapper scripts in user programs would be a big help.

There is a continued emphasis on build times and I would be concerned about this needing a toml parser, serde, and enough of the manifest schema to make this work.  That could very impede this proposal being adopted in core crates.  Yes, we could do intermediate solutions like pulling out `package.metadata`, serializing it to json, and passing that into the build script via an env variable, but that still requires `serde`, `serde_json`.

epage:
[...]

PR #15630, opened 2025-06-04
feat(toml): Parse support for multiple build scripts

namanlp:
Hi Everyone!

This is PR for the manifest parsing of the first milestone of [GSoC Project : Build Script Delegation](https://summerofcode.withgoogle.com/programs/2025/projects/nUt4PdAA)

Currently, just a single build script is allowed for each package. This PR will allow users to create and use multiple build scripts, and is backward compatible with single script as well as boolean values. 

**Motivation :** This will help users to maintain separate smaller and cleaner build scripts instead of one large build script. This is also necessary for build script delegation.

**Open questions:**
- What should the build script target names be?
  - Currently they use the file stem of the build script which could run into conflicts

**Known Issues:**
- This is just parsing support, and currently, only the first build script of the array is actually executed.

```toml
```

**This PR is ready to be reviewed and merged**


epage:
Feel free to rebase on `master`, rather than merge it.

crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -165,6 +165,15 @@ pub struct InheritablePackage {
...
 }
 
+#[derive(Deserialize, Serialize, Clone, Debug)]
+#[serde(untagged, rename_all = "kebab-case")]
+pub enum TomlPackageBuild {
+    /// If there's just one script or boolean value
+    SingleScript(StringOrBool),
+    /// If multiple build script are there
+    MultipleScripts(Vec<String>),
+}
```
epage: Personally, I'd suggest
```suggestion
#[derive(Deserialize, Serialize, Clone, Debug)]
#[serde(untagged, rename_all = "kebab-case")]
pub enum TomlPackageBuild {
    Auto(bool)
    /// If there's just one script or boolean value
    SingleScript(String),
    /// If multiple build script are there
    MultipleScripts(Vec<String>),
}
```

epage: If doing this, split this into commits,
- Fork `StringOrBool`
- Add `MultipleScripts`

crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -257,9 +266,12 @@ impl TomlPackage {
     pub fn normalized_build(&self) -> Result<Option<&String>, UnresolvedError> {
         let readme = self.build.as_ref().ok_or(UnresolvedError)?;
         match readme {
```
epage: Mind, in a separate commit, renaming this from `readme` to` build`?  Fun copy paste issues

crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -165,6 +165,15 @@ pub struct InheritablePackage {
     pub rust_version: Option<RustVersion>,
 }
 
+#[derive(Deserialize, Serialize, Clone, Debug)]
+#[serde(untagged, rename_all = "kebab-case")]
```
epage: Don't forget to use the `UntaggedEnumVisitor`

crates/cargo-util-schemas/src/manifest/mod.rs
namanlp: I think it should be fine now?

epage: Yes, `TomlPackageBuild` now uses `Auto`, thanks!

Could you split adding `MultipleScripts` out into its own commit?

namanlp: I have just added and set to `unimplemented!`.

Should I remove it and then do it again?

epage: While doing experiments (which you pushing your draft pr will end up including) it is fine for the history to be dirty.  However, for complete chunks of code (rename `readme`, forking `StringOrBool`), it would be good to clean up the history for those.

A clean history is important for understanding a change.  The commits in a PR should tell a story with each part small enough that it is trivial to look at.

If you need help in cleaning up any of the history, ping me on Zulip on we can get on a call.

src/cargo/core/features.rs
```diff
@@ -1361,6 +1362,7 @@ impl CliUnstable {
             "msrv-policy" => self.msrv_policy = parse_empty(k, v)?,
             // can also be set in .cargo/config or with and ENV
             "mtime-on-use" => self.mtime_on_use = parse_empty(k, v)?,
+            "multiple-build-scripts" => self.multiple_build_scripts = parse_empty(k, v)?,
```
epage: This commit should also add a section for this to our unstable docs.  There isn't much to say but we should add more to it as features get implemented.

crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -260,6 +260,9 @@ impl TomlPackage {
             TomlPackageBuild::Auto(false) => Ok(None),
             TomlPackageBuild::Auto(true) => Err(UnresolvedError),
             TomlPackageBuild::SingleScript(value) => Ok(Some(value)),
+            TomlPackageBuild::MultipleScript(_scripts) => {
+                unimplemented!("Multiple Build Scripts feature is not implemented yet!")
+            }
         }
```
epage: In a commit before this one, we should update `normalized_build` to return `Result<Option<&Vec<String>>, UnresolvedError>` to avoid this `unimplemented!`

src/cargo/ops/vendor.rs
```diff
@@ -515,24 +515,34 @@ fn prepare_toml_for_vendor(
...
+            };
+            package.build = Some(build);
+        }
+        Some(TomlPackageBuild::MultipleScript(_scripts)) => {
+            if gctx.cli_unstable().multiple_build_scripts {
+                unimplemented!("Multiple build script is not implemented yet!");
+            } else {
+                bail!("`build = […]` requires `-Z multiple-build-scripts`");
+            }
+        }
```
epage: I'd recommend unifying this code so there is a single code path for single or multiple scripts.

Maybe something like https://doc.rust-lang.org/std/option/enum.Option.html#method.as_slice could help

src/cargo/util/toml/mod.rs
```diff
@@ -669,10 +669,16 @@ fn normalize_package_toml<'a>(
         .map(|value| field_inherit_with(value, "authors", || inherit()?.authors()))
         .transpose()?
         .map(manifest::InheritableField::Value);
+    // Check if multiple build scripts are allowed or not
+    let multiple_build_scripts = gctx.cli_unstable().multiple_build_scripts;
     let build = if is_embedded {
         Some(TomlPackageBuild::Auto(false))
     } else {
-        targets::normalize_build(original_package.build.as_ref(), package_root)
+        targets::normalize_build(
```
epage: "the hackiest hack is the best hack"
or
"when implementing short-lived or hacky code, be as surgical as possible"

I'd recommend we do the unstable check here, rather than updating `normalize_build`s API to take in a parameter related to this

src/cargo/core/features.rs
```diff
@@ -840,6 +840,7 @@ unstable_cli_options!(
     minimal_versions: bool = ("Resolve minimal dependency versions instead of maximum"),
     msrv_policy: bool = ("Enable rust-version aware policy within cargo"),
     mtime_on_use: bool = ("Configure Cargo to update the mtime of used files"),
+    multiple_build_scripts: bool = ("Enables the use of multiple build scripts"),
```
epage: I overlooked this before, this was implemented as a `-Z` when this needs to be a `cargo_features`.

src/cargo/util/toml/mod.rs
```diff
@@ -2886,24 +2892,35 @@ fn prepare_toml_for_publish(
...
-            TomlPackageBuild::SingleScript(path)
-        } else {
-            ws.gctx().shell().warn(format!(
-                "ignoring `package.build` as `{}` is not included in the published package",
-                path.display()
-            ))?;
-            TomlPackageBuild::Auto(false)
-        };
-        package.build = Some(build);
+    match &package.build {
```
epage: ditto what I said with vendor

src/cargo/util/toml/targets.rs
```diff
@@ -1078,31 +1078,39 @@ Cargo doesn't know which to use because multiple target files found at `{}` and
...
-            Some(TomlPackageBuild::SingleScript(build))
+            Ok(Some(TomlPackageBuild::SingleScript(build)))
         }
         Some(TomlPackageBuild::Auto(true)) => {
-            Some(TomlPackageBuild::SingleScript(BUILD_RS.to_owned()))
+            Ok(Some(TomlPackageBuild::SingleScript(BUILD_RS.to_owned())))
+        }
+        Some(TomlPackageBuild::MultipleScript(_scripts)) => {
+            if multiple_build_scripts {
+                unimplemented!("Multiple Build Scripts feature is not implemented yet!");
```
epage: Let's go ahead and do this in this commit

epage: Also, I'd like to see runtime errors, rather than panics for when we cut users off from this feature.

src/cargo/core/features.rs
namanlp: Ok, replaced it with `cargo_features`

epage: Thanks!

crates/cargo-util-schemas/src/manifest/mod.rs
epage: https://doc.rust-lang.org/std/slice/fn.from_ref.html should help

src/cargo/util/toml/mod.rs
namanlp: Should be done

src/cargo/util/toml/targets.rs
namanlp: Should be done

crates/cargo-util-schemas/src/manifest/mod.rs
namanlp: Should I do it after this commit?

src/cargo/util/toml/targets.rs
```diff
@@ -105,7 +105,7 @@ pub(super) fn to_targets(
         if metabuild.is_some() {
             anyhow::bail!("cannot specify both `metabuild` and `build`");
         }
-        let custom_build = Path::new(custom_build);
+        let custom_build = Path::new(&custom_build[0]);
```
epage: As there can never be more than one, there should at least be an assert within this commit (even if a later commit turns it into an error)

namanlp: It is when `MultipleScript` was not introduced...

src/cargo/util/toml/targets.rs
```diff
@@ -1079,30 +1082,33 @@ Cargo doesn't know which to use because multiple target files found at `{}` and
...
             );
-            Some(TomlPackageBuild::SingleScript(build))
+            Ok(Some(TomlPackageBuild::SingleScript(build)))
         }
         Some(TomlPackageBuild::Auto(true)) => {
-            Some(TomlPackageBuild::SingleScript(BUILD_RS.to_owned()))
+            Ok(Some(TomlPackageBuild::SingleScript(BUILD_RS.to_owned())))
+        }
+        Some(TomlPackageBuild::MultipleScript(_scripts)) => {
+            anyhow::bail!("multiple build scripts feature is not implemented yet!");
```
epage: I forgot to call this out as well if we decided to do `prepare_toml_for_vendor` and `prepare_toml_for_publish` in this commit

namanlp: I guess it would be better to deal with them in different commit, because it is not dependant? Or maybe we can do this in this one?

src/cargo/util/toml/targets.rs
```diff
@@ -105,7 +105,10 @@ pub(super) fn to_targets(
...
             anyhow::bail!("cannot specify both `metabuild` and `build`");
         }
-        let custom_build = Path::new(custom_build);
+        if custom_build.len() > 1 {
+            anyhow::bail!("multiple build scripts feature is not implemented yet! ")
+        }
+        let custom_build = Path::new(&custom_build[0]);
         let name = format!(
             "build-script-{}",
             custom_build
```
epage: 1. We'll need to loop over
2. Do we need to make build script names more unique?

src/doc/src/reference/unstable.md
```diff
@@ -331,6 +332,23 @@ extra-info = "qwerty"
...
+Add the paths of the build scripts as an array in `package.build`. For example:
+
+```toml
+cargo-features = ["multiple-build-scripts"]
+
+[package]
+name = "mypackage"
+version = "0.0.1"
+build = ["foo.rs", "bar.rs"]
+```
```
epage, line 352: Tests:
- [ ] `cargo metadata`
- [ ] `cargo check` that uses the feature
- [ ] `cargo check` with a single build script but is broken because of that.  *Once* the feature is done, it switches to the feature to show that the bug is fixed
- [ ] feature gate test: uses the feature without being enabled, showing it fails
- [ ] Somehow we need to verify each build script run is isolated

Open question: how do we deal with multiple OUT_DIRs or is there only one and the build scripts have to cooperate?

namanlp: Ok, so `cargo metadata` is not getting generated in the latest commit because 

> Caused by:
  multiple build scripts feature is not implemented yet!


So, I guess it would be best to just include the test with this output?

epage: I generally recommend the tests be inserted before any feature work is done, so thats fine.

epage: So that means I'd recommend moving the test commit to before `MultipleScripts` is added

tests/testsuite/build_scripts_multiple.rs
```diff
@@ -89,17 +87,42 @@ fn multiple_build_scripts_metadata() {
...
+        .find(|p| p["name"].as_str().unwrap() == "foo")
+        .unwrap()["targets"]
+        .as_array()
+        .unwrap()
+        .iter()
+        .find(|p| p["name"].as_str().unwrap() == "build-script-build2")
+        .unwrap()["src_path"]
+        .as_str()
+        .unwrap();
+    assert_eq!(build2, p.root().join("build2.rs").as_os_str());
```
epage: See `tests/testsuite/metadata.rs` for how we expect `metadata` to be validated (ie snapshot testing of the json, including as much as is relevant for the test)

src/cargo/util/toml/targets.rs
epage: So for testing publish
- [ ] We filter out build scripts that aren't included
- [ ] Normalize paths

src/cargo/util/toml/mod.rs
```diff
@@ -2890,25 +2890,33 @@ fn prepare_toml_for_publish(
...
+        }
+        package.build = Some(if included_scripts.is_empty() {
             TomlPackageBuild::Auto(false)
-        };
-        package.build = Some(build);
+        } else if included_scripts.len() == 1 {
+            TomlPackageBuild::SingleScript(included_scripts[0].clone())
+        } else {
+            TomlPackageBuild::MultipleScript(included_scripts)
+        });
```
epage: nit: I'd represent this as a `match included_scripts.len()` rather than a set of if-elses.  It makes the intent a bit clearer

src/cargo/util/toml/mod.rs
epage, line 1: > Refactor prepare_toml_for_publish for MultipleScript 

nit: a refactor has no user-facing changes while this is implementing publish support for multiple build scripts which is user facing

src/cargo/util/toml/mod.rs
```diff
@@ -2890,25 +2890,33 @@ fn prepare_toml_for_publish(
...
+            if included {
+                let path = path
+                    .into_os_string()
+                    .into_string()
+                    .map_err(|_err| anyhow::format_err!("non-UTF8 `package.build`"))?;
+                let path = normalize_path_string_sep(path);
+                included_scripts.push(path);
+            } else {
+                ws.gctx().shell().warn(format!(
+                    "ignoring `package.build` as `{}` is not included in the published package",
```
namanlp: Should we change this message because `package.build` isn't ignored if some of the scripts are not included? However, it does change message for other as well

epage: Would could change it to

```
ignoring `package.build` entry `{}` as it is ...
```

src/doc/src/reference/unstable.md
```diff
@@ -331,6 +332,23 @@ extra-info = "qwerty"
 Metabuild packages should have a public function called `metabuild` that
 performs the same actions as a regular `build.rs` script would perform.
 
+## Multiple Build Scripts
+* Original Pull Request: [#15630](https://github.com/rust-lang/cargo/pull/15630)
```
epage: Let's use #14903 as the tracking issue.  Please link to it from here

namanlp: Should I also mention about GSoC?

epage: Not needed.

epage:
Thanks!

PR #15633, opened 2025-06-05
fix(workspace): reload current manifest path member only

### LINKED ISSUE #15625: `cargo clippy --fix` builds too many packages in a workspace (since `nightly-2025-06-01`)
### Problem

Since `nightly-2025-06-01`, a command like `cargo clippy --manifest-path foo/Cargo.toml --fix` where "foo" is in a workspace, the `--fix` option will cause all workspace members to be built, not just the one package that was specified by the --manifest-path.

# Example

Using this repo https://github.com/devjgm/bug-cargo-clippy-fix-repro with two `foo` and `bar` members that were both created with `cargo new`, we see the following output.

```console
❯ cargo clippy --manifest-path foo/Cargo.toml --fix
    Checking foo v0.1.0 (/private/tmp/bug-cargo-clippy-fix-repro/foo)
    Checking bar v0.1.0 (/private/tmp/bug-cargo-clippy-fix-repro/bar)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.28s
```

Note that I've specified `--manifest-path foo/Cargo.toml`, so I do not expect `bar` to be built, but we see that `bar` is also built.

* The problem goes away you remove `--fix`
* The problems goes away if you roll back to `nightly-2025-05-31`
* If I specify `-p foo`, the problem also goes away. However, `-p` should likely not be required if I've already specified `--manifest-path`.

I believe the issue started in https://github.com/rust-lang/cargo/pull/15596/commits/778ff6885ac9538a73d8c6ecf692917dc2f09759. cc @ehuss 

### Steps

1. `cd /tmp && git clone https://github.com/devjgm/bug-cargo-clippy-fix-repro`
2. `cargo clippy --manifest-path foo/Cargo.toml --fix`

Problem: You see that `bar` is also built, but it should not be.

### Possible Solution(s)

I believe the problem was introduced in https://github.com/rust-lang/cargo/pull/15596/commits/778ff6885ac9538a73d8c6ecf692917dc2f09759

### Notes

_No response_

### Version

```text
❯ cargo version --verbose
cargo 1.89.0-nightly (64a124607 2025-05-30)
release: 1.89.0-nightly
commit-hash: 64a12460708cf146e16cc61f28aba5dc2463bbb4
commit-date: 2025-05-30
host: aarch64-apple-darwin
libgit2: 1.9.0 (sys:0.20.2 vendored)
libcurl: 8.7.1 (sys:0.4.80+curl-8.12.1 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 3.5.0 8 Apr 2025
os: Mac OS 15.4.1 [64-bit]
```

lemorage:

As of #15625, the manifest path argument in `cargo clippy --manifest-path foo/Cargo.toml --fix` will be ignored. All the workspace members will be built. The cause is due to the `reload` usage in `cargo::ops::fix`. We reload the `root_manifest` in the function, which contains all workspace members.

Will close #15625.

The first commit in the PR demonstrates the current problem, and the second commit corrects it. Use `cargo test --test testsuite workspaces::fix_only_check_manifest_path_member` to see the test results.

### DIFF
--- a/tests/testsuite/workspaces.rs
+++ b/tests/testsuite/workspaces.rs
@@ -2716,3 +2716,43 @@ fn nonexistence_package_together_with_workspace() {
 "#]])
         .run();
 }
+
+// A failing case from <https://github.com/rust-lang/cargo/issues/15625>
+#[cargo_test]
+fn clippy_fix_all_members_in_a_workspace() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+            [workspace]
+            members = ["foo", "bar"]
+            resolver = "3"
+            "#,
+        )
+        .file(
+            "foo/Cargo.toml",
+            r#"
+            [package]
+            name = "foo"
+            version = "0.1.0"
+            edition = "2021"
+            "#,
+        )
+        .file("foo/src/main.rs", "fn main() {}")
+        .file(
+            "bar/Cargo.toml",
+            r#"
+            [package]
+            name = "bar"
+            version = "0.1.0"
+            edition = "2021"
+            "#,
+        )
+        .file("bar/src/main.rs", "fn main() {}")
+        .build();
+
+    p.cargo("clippy --manifest-path foo/Cargo.toml --fix --allow-no-vcs")
+        .with_stderr_contains("[CHECKING] foo v0.1.0 ([ROOT]/foo/foo)")
+        .with_stderr_contains("[CHECKING] bar v0.1.0 ([ROOT]/foo/bar)")
+        .run();
+}
--- a/tests/testsuite/workspaces.rs
+++ b/tests/testsuite/workspaces.rs
@@ -2719,7 +2719,7 @@ fn nonexistence_package_together_with_workspace() {
 
 // A failing case from <https://github.com/rust-lang/cargo/issues/15625>
 #[cargo_test]
-fn clippy_fix_all_members_in_a_workspace() {
+fn fix_only_check_manifest_path_member() {
     let p = project()
         .file(
             "Cargo.toml",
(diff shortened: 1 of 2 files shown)


tests/testsuite/workspaces.rs
```diff
@@ -2716,3 +2716,43 @@ fn nonexistence_package_together_with_workspace() {
...
+            name = "bar"
+            version = "0.1.0"
+            edition = "2021"
+            "#,
+        )
+        .file("bar/src/main.rs", "fn main() {}")
+        .build();
+
+    p.cargo("clippy --manifest-path foo/Cargo.toml --fix --allow-no-vcs")
+        .with_stderr_contains("[CHECKING] foo v0.1.0 ([ROOT]/foo/foo)")
```
epage: Would it work to use `with_stderr`? If there is a determinism issue, the `unordered` variant can help.

`contains` doesn't allow snapshot updating and `does_not_contain` can easily test for the wrong thing as output changes. We mostly use these as a last resort.

lemorage: Ah, right, I'll correct that.

tests/testsuite/workspaces.rs
```diff
@@ -2716,3 +2716,46 @@ fn nonexistence_package_together_with_workspace() {
...
+            [package]
+            name = "bar"
+            version = "0.1.0"
+            edition = "2021"
+            "#,
+        )
+        .file("bar/src/main.rs", "fn main() {}")
+        .build();
+
+    p.cargo("clippy --manifest-path foo/Cargo.toml --fix --allow-no-vcs")
```
weihanglo: The bug is in cargo-fix. We can minimize it to that.

lemorage: Sure!

Issue #15622, opened 2025-06-02
When publishing, Cargo cannot unify a local and a registry dependency in the lockfile


epage:
### Problem

When you have a direct dependency on a workspace member *and* an indirect dependency on a copy from the registry, when publishing they aren't unified and the process fails.

A simplified version of when I ran into this
```
clap_complete_nushell (local) -> clap (local)
clap_complete_nushell (local) -> nu (dev)

nu -> clap
```
If the version requirement on `clap` is too high to use the registry version, then it fails to resolve because it is locked to the lower version.

The user has to manually publish one package, `cargo update`, and then publish the other package.

*Originally reported at https://github.com/rust-lang/cargo/issues/1169#issuecomment-2913503095*

### Steps

As a test case:
```rust

#[cargo_test]
fn unify_local_and_regitry_dependency() {
    let registry = RegistryBuilder::new().http_api().http_index().build();
    Package::new("dep", "1.0.0+registry").publish();
    Package::new("transitive", "1.0.0")
        .dep("dep", "1.0.0+registry")
        .publish();

    let p = project()
        .file(
            "Cargo.toml",
            r#"
            [workspace]
            members = ["parent", "dep"]
            "#,
        )
        .file(
            "parent/Cargo.toml",
            r#"
            [package]
            name = "parent"
            version = "1.0.0"
            edition = "2015"
            license = "MIT"
            description = "foo"
            repository = "foo"

            [dependencies]
            transitive = "1.0.0"
            dep = { version = "1.100", path = "../dep" }
        "#,
        )
        .file("parent/src/lib.rs", "")
        .file(
            "dep/Cargo.toml",
            r#"
            [package]
            name = "dep"
            version = "1.100.0+local"
            edition = "2015"
            license = "MIT"
            description = "foo"
            repository = "foo"
        "#,
        )
        .file("dep/src/lib.rs", "")
        .build();

    p.cargo("generate-lockfile").run();
    p.cargo("package --workspace -Zpackage-workspace")
        .masquerade_as_nightly_cargo(&["package-workspace"])
        .replace_crates_io(registry.index_url())
        .with_status(101)
        .with_stderr_data(str![[r#"
[PACKAGING] dep v1.100.0+local ([ROOT]/foo/dep)
[PACKAGED] 4 files, [FILE_SIZE]B ([FILE_SIZE]B compressed)
[PACKAGING] parent v1.0.0 ([ROOT]/foo/parent)
[UPDATING] crates.io index
[...]

Issue #14396, opened 2024-08-14
"no hash listed" error with `-Zpackage-workspace`


workingjubilee:
### Problem

In the [pgrx](https://github.com/pgcentralfoundation/pgrx) repo.

Command:
```bash
./update-versions.sh 0.0.999-rc.999 && \
    cargo +nightly package \
        --workspace -Zpackage-workspace \
        --allow-dirty \
        --features "pg14"
```

Result:
```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 50.13s
   Verifying pgrx v0.0.999-rc.999 (/home/jubilee/tcdi/pgrx/pgrx)
    Updating crates.io index
   Unpacking pgrx-pg-sys v0.0.999-rc.999 (registry `/home/jubilee/tcdi/pgrx/target/package/tmp-registry`)
error: failed to verify package tarball

Caused by:
  failed to download `pgrx-pg-sys v0.0.999-rc.999`

Caused by:
  unable to get packages from source

Caused by:
  no hash listed for pgrx-pg-sys v0.0.999-rc.999
note: this is an unexpected cargo internal error
note: we would appreciate a bug report: https://github.com/rust-lang/cargo/issues/
note: cargo 1.82.0-nightly (94977cb1f 2024-08-06)
```
baffling? no idea what I did beyond running this command to cause this. I just dropped everything to open a report first. I've run almost this precise command several times successfully...? Now every time I run this command in that repo, with that working state, I get the error:
```bash
cargo +nightly package --workspace -Zpackage-workspace --allow-dirty --features "pg14"
```

I saved my entire repo directory as a duplicate before I did anything else.

### Version

```text
cargo 1.82.0-nightly (94977cb1f 2024-08-06)
release: 1.82.0-nightly
commit-hash: 94977cb1fab003d45eb5bb108cb5e2fa0149672a
commit-date: 2024-08-06
host: x86_64-unknown-linux-gnu
libgit2: 1.8.1 (sys:0.19.0 vendored)
libcurl: 8.9.0-DEV (sys:0.4.74+curl-8.9.0 vendored ssl:OpenSSL/3.3.1)
ssl: OpenSSL 3.3.1 4 Jun 2024
os: Arch Linux Rolling Release [64-bit]
```

epage:
CC @jneem

jneem:
@workingjubilee can you help me obtain that commit? A fresh clone doesn't give me it...

[...]

PR #15636, opened 2025-06-05
feat(publish): Stabilize multi-package publishing

### LINKED ISSUE #1169: cargo publish --all
It would be nice to have a flag to `cargo publish` which publishes all local packages in a DAG fashion.

### LINKED ISSUE #10948: cargo package --workspace is not very useful

epage:

A user will now be able to use flags like `--workspace` with `cargo publish`.
`cargo package` will now also work with those flags without having to pass `--no-verify --exclude-lockfile`.

Many release tools have come out that solve this problem. They will still need a lot of the logic that went into that for other parts of the release process.
However, a cargo-native solution allows for:
- Verification during dry-run
- Better strategies for waiting for the publish timeout

`cargo publish` is non-atomic at this time.
If there is a server side error, network error, or rate limit during the publish, the workspace will be left in a partially published state. Verification is done before any publishing so that won't affect things. There are multiple strategies we can employ for improving this over time, including
- atomic publish
- `--idempotent` (#13397)
- leave this to release tools to manage

This includes support for `--dry-run` verification. As release tools didn't have a way to do this before, users may be surprised at how slow this is because a `cargo build` is done instead of a `cargo check`.  This is being tracked in #14941.

This adds to `cargo package` the `--registry` and `--index` flags to help with resolving dependencies when depending on a package being packaged at that moment.
These flags are only needed when a `cargo package --workspace` operation would have failed before due to inability to find a locally created dependency.

Regarding the publish timeout, `cargo publish --workspace` publishes packages in batches and we only timeout if nothing in the batch has finished being published within the timeout, deferring the rest to the next wait-for-publish. So for example, if you have packages `a`, `b`, `c` then we'll wait up to 60 seconds and if only `a` and `b` were ready in that time, we'll then wait another 60 seconds for `c`.

During testing, users ran into issues with `.crate` checksums:
- ~~#15647~~ Fixed for `cargo publish --dry-run` in #15711 
  - But `cargo package` still has the problem
- #14396 (not been able to reproduce)
- #15622 (reproducible with consecutive `cargo publish` calls)

Fixes #1169
Fixes #10948

By stabilizing this, Cargo's behavior becomes dependent on an overlay registry.
When generating a lockfile or verifying a package, we overlay the locally generated `.crate` files on top of the registry so the registry appears as it would and everything works.
If there is a conflict with a version, the local version wins which is important for the dry-run mode of release tools as they won't have bumped the version yet.
Our concern for the overlay registry is dependency confusion attacks. Considering this is not accessible for general user operations, this should be fine.

### DIFF
--- a/src/bin/cargo/commands/package.rs
+++ b/src/bin/cargo/commands/package.rs
@@ -7,8 +7,8 @@ use cargo::ops::PackageOpts;
 pub fn cli() -> Command {
     subcommand("package")
         .about("Assemble the local package into a distributable tarball")
-        .arg_index("Registry index URL to prepare the package for (unstable)")
-        .arg_registry("Registry to prepare the package for (unstable)")
+        .arg_index("Registry index URL to prepare the package for")
+        .arg_registry("Registry to prepare the package for")
         .arg(
             flag(
                 "list",
@@ -57,22 +57,6 @@ pub fn cli() -> Command {
 }
 
 pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
-    if args._value_of("registry").is_some() {
-        gctx.cli_unstable().fail_if_stable_opt_custom_z(
-            "--registry",
-            13947,
-            "package-workspace",
-            gctx.cli_unstable().package_workspace,
-        )?;
-    }
-    if args._value_of("index").is_some() {
-        gctx.cli_unstable().fail_if_stable_opt_custom_z(
-            "--index",
-            13947,
-            "package-workspace",
-            gctx.cli_unstable().package_workspace,
-        )?;
-    }
     let reg_or_index = args.registry_or_index(gctx)?;
     let ws = args.workspace(gctx)?;
     if ws.root_maybe().is_embedded() {
--- a/src/bin/cargo/commands/publish.rs
+++ b/src/bin/cargo/commands/publish.rs
@@ -20,8 +20,8 @@ pub fn cli() -> Command {
         .arg_silent_suggestion()
         .arg_package_spec_no_all(
             "Package(s) to publish",
-            "Publish all packages in the workspace (unstable)",
-            "Don't publish specified packages (unstable)",
+            "Publish all packages in the workspace",
+            "Don't publish specified packages",
         )
         .arg_features()
         .arg_parallel()
@@ -43,23 +43,6 @@ pub fn exec(gctx: &mut GlobalContext, args: &ArgMatches) -> CliResult {
             ws.root_manifest().display()
         )
         .into());
-    }
-
-    let unstable = gctx.cli_unstable();
-    let enabled = unstable.package_workspace;
-    if args.get_flag("workspace") {
-        unstable.fail_if_stable_opt_custom_z("--workspace", 10948, "package-workspace", enabled)?;
-    }
-    if args._value_of("exclude").is_some() {
-        unstable.fail_if_stable_opt_custom_z("--exclude", 10948, "package-workspace", enabled)?;
-    }
-    if args._values_of("package").len() > 1 {
-        unstable.fail_if_stable_opt_custom_z(
-            "--package (multiple occurrences)",
-            10948,
-            "package-workspace",
-            enabled,
-        )?;
     }
 
     ops::publish(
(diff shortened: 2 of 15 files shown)


epage:
@rfcbot fcp merge

See the PR description for details

weihanglo (commented):
Code change has no issues

epage:
I have gotten a reproduction case for 1 of the 2 checksum issues, see #15647.

epage:
As an update, #15647 is being fixed in #15711.  Both are scoped only for `cargo publish --dry-run` and `cargo package` continues to error.  Whether to error or not in `cargo package` is a workflow question.

Also, I forgot to note #15622 in the stabilization report.  While it can be reproduced with consecutive `cargo publish` calls, making it not directly related to this stabilization, it is tied to the workflows that are used with this feature so I figured it is worth noting.

weihanglo (approved):

epage:
@weihanglo CI is green again

MOZGIII:
Quite odd that this flag assumes all the packages have changed versions; I have a workspace where I control versions precisely and independently, and this flag doesn't work for me. I bump one crate and it aborts with `warning: crate not-the-one-i-bumped@1.2.3 already exists on crates.io index` (where `not-the-one-i-bumped@1.2.3` is not the crate I bumped - but one that I didn't touch at all).

I also mentioned this at https://github.com/rust-lang/cargo/issues/13397 - but it feels like this is would be a baseline use-case, not an extension

Sytten:
For others, I think I misunderstood what @epage said with:
> Verification is done before any publishing so that won't affect things. There are multiple strategies we can employ for improving this over time, including

This seems to mean that it will fail instead of skipping already published packages, which was what I expected to happen

epage:
See #13397 for skipping published packages

PR #15639, opened 2025-06-06
test(lints): update snapshot

weihanglo:
Haven't figured out why CI was green, but it failed on my machine.


weihanglo:
Perhaps the terminal detection is drifted between Cargo and anstyle. Given it is not an issue on our CI, let me turn this into an issue.

PR #15645, opened 2025-06-09
Use `Not::not` rather than a custom `is_false` function

joshtriplett:
The `is_false` function already exists in the standard library, as `Not::not`; use that in `skip_serializing_if` rather than defining an `is_false` function.


weihanglo (approved):
Thanks!

PR #15653, opened 2025-06-10
Allow using Cargo-as-a-library with gix's reqwest backend

dtolnay:

Unconditionally enabling "blocking-http-transport-curl" made the `cargo` library incompatible with crates that prefer reqwest. An example being the `rustsec` crate with git support:

- https://github.com/rustsec/rustsec/blob/rustsec/v0.30.2/rustsec/Cargo.toml#L45-L47
- https://github.com/rustsec/rustsec/blob/rustsec/v0.30.2/rustsec/Cargo.toml#L34
- https://github.com/EmbarkStudios/tame-index/blob/0.21.0/Cargo.toml#L14-L17

Having `cargo` and `rustsec` in the same dependency graph makes `gix-transport` fail to compile.

```toml
[dependencies]
cargo = "0.88.0"
rustsec = "0.30.2"
```

```console
error[E0428]: the name `Impl` is defined multiple times
   --> $CARGO_HOME/registry/src/index.crates.io-1949cf8c6b5b557f/gix-transport-0.45.0/src/client/blocking_io/http/mod.rs:220:1
    |
217 | pub type Impl = curl::Curl;
    | --------------------------- previous definition of the type `Impl` here
...
220 | pub type Impl = reqwest::Remote;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `Impl` redefined here
    |
    = note: `Impl` must be defined only once in the type namespace of this module

error: Cannot set both 'http-client-reqwest' and 'http-client-curl' features as they are mutually exclusive
  --> $CARGO_HOME/registry/src/index.crates.io-1949cf8c6b5b557f/gix-transport-0.45.0/src/client/blocking_io/http/mod.rs:26:1
   |
26 | compile_error!("Cannot set both 'http-client-reqwest' and 'http-client-curl' features as they are mutually exclusive");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

After this PR, dependency graphs that prefer reqwest can switch to Cargo's http-transport-reqwest feature.

```toml
[dependencies]
cargo = { default-features = false, features = ["http-transport-reqwest"] }
rustsec = "0.30.2"
```

Cargo will continue to have a direct dependency on `curl`, but HTTP operations performed through gix will use `reqwest`. This means both curl's HTTP implementation and reqwest's HTTP implementation will be linked. This is still much better than the only existing solution, which is that you must pick versions of `cargo` and other dependency (`rustsec`) which depend on semver-incompatible versions of gix, causing 2 entire versions of gix to be linked, in order to sidestep the mutually exclusive features being enabled on the same version of gix. Gix version numbers advance rapidly enough that this is often possible, but sometimes (like right now) you would be unable to use the most recent published release of `cargo`.

- `cargo check --lib`
- `cargo check --lib --no-default-features --features http-transport-reqwest`

Also tested by backporting this commit onto <https://github.com/rust-lang/cargo/pull/15391>'s base commit (i.e. when gix 0.70 was used) to ensure a conflict with rustsec's gix dependency, and successfully building the following project.

```toml
[package]
name = "repro"
version = "0.0.0"
edition = "2024"
publish = false

[dependencies]
cargo = { path = "../cargo", default-features = false, features = ["http-transport-reqwest"] }
rustsec = "0.30.2"
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ flate2 = { version = "1.1.1", default-features = false, features = ["zlib-rs"] }
 git2 = "0.20.0"
 git2-curl = "0.21.0"
 # When updating this, also see if `gix-transport` further down needs updating or some auth-related tests will fail.
-gix = { version = "0.72.0", default-features = false, features = ["blocking-http-transport-curl", "progress-tree", "parallel", "dirwalk"] }
+gix = { version = "0.72.0", default-features = false, features = ["progress-tree", "parallel", "dirwalk"] }
 glob = "0.3.2"
 # Pinned due to https://github.com/sunng87/handlebars-rust/issues/711
 handlebars = { version = "=6.3.1", features = ["dir_source"] }
@@ -271,10 +271,15 @@ test = false
 doc = false
 
 [features]
+default = ["http-transport-curl"]
 vendored-openssl = ["openssl/vendored"]
 vendored-libgit2 = ["libgit2-sys/vendored"]
 # This is primarily used by rust-lang/rust distributing cargo the executable.
 all-static = ['vendored-openssl', 'curl/static-curl', 'curl/force-system-lib-on-osx', 'vendored-libgit2']
+# Exactly one of 'http-transport-curl' or 'http-transport-reqwest' must be enabled
+# when using Cargo as a library. By default, it is 'http-transport-curl'.
+http-transport-curl = ["gix/blocking-http-transport-curl"]
+http-transport-reqwest = ["gix/blocking-http-transport-reqwest"]
 
 [lints]
 workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 4
 
 [[package]]
+name = "addr2line"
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -271,10 +271,15 @@ test = false
 doc = false
 
 [features]
+default = ["http-transport-curl"]
```
weihanglo, line 274: I don't have any objection to this. Let me bring it to the Cargo team's meeting.

dtolnay:
CI was trying to run `cargo semver-checks check-release --workspace` which runs (effectively) `cargo check --all-features --lib`, causing both gix backends to be enabled, which is an error. I added a workaround but not sure if it is the best one.

weihanglo:
Thank you! Would you mind cleaning up some intermediate and revert commits?

Issue #15524, opened 2025-05-13
`cargo add`: Misleading error message when feature isn't available in semver


ultrabear:
### Problem

When using `cargo add foo --features bar` when bar was added in `1.0.1` (and is locked to that version) but Cargo.toml already specifies `foo = 1.0`, cargo correctly errors because bar is not in 1.0, however, the error message does not very clearly state the version checked, nor does it attempt to parse the currently locked version and offer a hint to the developer that maybe they should change Cargo.toml to specify `1.0.1`.

### Steps

```sh
cargo new test_add
cd test_add
cargo add serde_json@1
#    Updating crates.io index
#      Adding serde_json v1 to dependencies
#             Features as of v1.0.0:
#             - linked-hash-map
#             - preserve_order
#    Updating crates.io index
#     Locking 10 packages to latest Rust 1.86.0 compatible versions

# above "Features as of v1.0.0" is noted

# but the following command was run at a later point in time such that it was forgotten

cargo add serde_json --features raw_value
#    Updating crates.io index
#      Adding serde_json v1 to dependencies
#error: unrecognized feature for crate serde_json: raw_value
#disabled features:
#    linked-hash-map, preserve_order

```

### Possible Solution(s)

there are two avenues of improvement, ideally both can be implemented:
- adding `Features as of xxx` to the second cargo add command to make it clearer that the feature is not available in the current semver
- give a hint to the programmer that in a later version of `serde_json` (and indeed the one that was in the lockfile) the feature is available, so they should update that dependencies semver specification

### Notes

This is clearly low priority, its a confused me once and never again type of issue, but it felt right to at least track an area of improvement.

### Version

```text
cargo 1.86.0 (adf9b6ad1 2025-02-28)
release: 1.86.0
commit-hash: adf9b6ad14cfa10ff680d5806741a144f7163698
commit-date: 2025-02-28
host: x86_64-unknown-linux-gnu
libgit2: 1.9.0 (sys:0.20.0 vendored)
libcurl: 8.12.0-DEV (sys:0.4.79+curl-8.12.0 vendored ssl:OpenSSL/1.1.1w)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Manjaro 25.0.1 (Zetar) [64-bit]
```

weihanglo:
[...]

PR #15658, opened 2025-06-13
feat: always show list of features on `cargo add`

motorailgun:
Partially solves #15524 

Currently, cargo shows list of features available for the package added only when features specified correctly. In turn, if the features specified by `cargo add --features` doesn't exist (at least for the version specified in Cargo.toml), it just bails with "unrecognized feature(s)" error.

This PR modifies the timing of `print_dep_table_msg()` so that it's always called.

The changes of UI can be observed by peeking the modified SVGs. I think those changes make sense, however there might be different opinions.


motorailgun:
Thank you for reviewing! Just squashed and pushed.

tests/testsuite/cargo_add/unknown_inherited_feature/stderr.term.svg
epage, line 1: #15438 added the "similar feature" messages and the lists of disabled and enabled features that are showcased in these error messages.  We even customized the displaying of those specifically for errors (collapsing enabled features while the regular display will collapse disabled features).

So this feels a bit redundant to show the feature list.  Is there a reason you feel we should do both?

epage (changes requested):

tests/testsuite/cargo_add/unknown_inherited_feature/stderr.term.svg
epage: FYI our [contrib docs](https://doc.crates.io/contrib/process/working-on-cargo.html) recommend opening issues first so we can discuss these things before implementing so we know what direction to implement and avoid you going in a direction we might not end up going with

motorailgun: Oops I somehow didn't notice that. I thought the "as of v x.x.x" part gives better insights in cases like the issue I mentioned, however it's undeniable that this feels too redundant.

And I'm sorry about discussion part - I hurried too much to do something; I should have talked on the issue beforehand.

Closing this PR, thanks for taking your time to review!

PR #15659, opened 2025-06-13
highlight the correct words

tshepang:


src/doc/contrib/src/index.md
```diff
@@ -5,7 +5,7 @@ overview of how to contribute to Cargo, how to dive into the code, and how the
 testing infrastructure works.
 
 There are many ways to contribute, such as [helping other users], [filing
-issues], [improving the documentation], [triaging issues], [fixing bugs], and
```
weihanglo, line 8: I think this is intentional. We highlight verbs as well for each.

tshepang: Those other ones link to docs regarding the highlighted text (like "helping other users" links to a site where you can help other users), while the one I changed just links to the docs in general, and not specifically docs on how to help improve the docs.

weihanglo: I am convinced. Let's do this until we have a dedicated chapter for documentation contribution.

weihanglo (approved):

PR #15661, opened 2025-06-13
fix(docs): align code block styles for shell snippets

motorailgun:

Current contributor document has inconsistent styles for shell-command snippets, like with or without leading `$`. This PR removes all leading `$` and specify bash for syntax highlighting, for the ease of copy-pasting commands.

### DIFF
--- a/src/doc/contrib/src/implementation/debugging.md
+++ b/src/doc/contrib/src/implementation/debugging.md
@@ -8,7 +8,7 @@ It also supports filtering for specific modules with comma-separated [directives
 Feel free to use [shorthand macros] to help with diagnosing problems.
 We're looking forward to making Cargo logging mechanism more structural!
 
-```sh
+```bash
 # Outputs all logs with levels debug and higher
 CARGO_LOG=debug cargo generate-lockfile
 
--- a/src/doc/contrib/src/tests/profiling.md
+++ b/src/doc/contrib/src/tests/profiling.md
@@ -13,9 +13,9 @@ At process exit, your trace will be in a file like `trace-1668480819035032.json`
 Open that file with [ui.perfetto.dev](https://ui.perfetto.dev) (or chrome://tracing) to browse your trace.
 
 Example:
-```console
-$ # Output first three levels of profiling info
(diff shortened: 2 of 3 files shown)


src/doc/contrib/src/implementation/debugging.md
```diff
@@ -8,7 +8,7 @@ It also supports filtering for specific modules with comma-separated [directives
 Feel free to use [shorthand macros] to help with diagnosing problems.
 We're looking forward to making Cargo logging mechanism more structural!
 
-```sh
+```bash
 # Outputs all logs with levels debug and higher
 CARGO_LOG=debug cargo generate-lockfile
```
epage, line 14: I tend to lean towards `console` blocks for commands

motorailgun: Fair point as it feels shell-independent. Will fix those!

motorailgun: Umm, I thought so, but seems like `bash` blocks works better (in terms of syntax highlighting) for multi-line commands with comments like this. How do you think about that...?

epage: I've tended to do `$ #`.  Not great but I feel its more semantically correct than describing these as bash scripts (which is what `bash` means).  Unsure if others have other opinions.

motorailgun: That sounds fair, however for especially one-liner like this, I'd prefer the ease of copy-pasting by removing leading `$`: 

https://github.com/rust-lang/cargo/blob/f6cea42b7988fd737409abc7521a9ade796c26cf/src/doc/contrib/src/tests/writing.md?plain=1#L224-L226

epage: I see that as a tooling limitation and I personally still lean towards `console`

epage:
@motorailgun this has not had any responses to the review feedback and has developed conflicts.  Do you still plan to drive this forward?

motorailgun:
@epage Ah, sorry for letting this be as is. I'd personally want to amend and make this complete. Is it desirable direction to align all shell snippets' code block styles to `console` ?

PR #15662, opened 2025-06-13
feat: Add custom completer for `cargo remove <TAB>`

### LINKED ISSUE #15656: Autocomplete `cargo remove` with installed dependencies
### Problem

When I type `cargo remove` and press tab twice in a workspace with dependencies, a dash appears.

If I press tab again, I see a lot of options, but none of them are a dependency installed in the current workspace.

### Proposed Solution

Pressing tab after typing `cargo remove` shows installed dependencies in the workspace that we can remove.

### Notes

_No response_

motorailgun:
fix #15656 

This PR implements custom TAB completer for cargo remove <package_name>. 

This patch works like this:

```console
$ cat Cargo.toml
[package]
name = "sandbox"
version = "0.1.0"
edition = "2024"

[dependencies]
rand = "0.9.1"
serde_json = "1"

$ ~/cargo/target/debug/cargo remove 
completing values
--build          -- Remove from build-dependencies
--color          -- Coloring
--config         -- Override a configuration value
--dev            -- Remove from dev-dependencies
--dry-run        -- Don't actually write the manifest
--frozen         -- Equivalent to specifying both --locked and --offline
--help           -- Print help
--locked         -- Assert that `Cargo.lock` will remain unchanged
--lockfile-path  -- Path to Cargo.lock (unstable)
--manifest-path  -- Path to Cargo.toml
--offline        -- Run without accessing the network
--package        -- Package to remove from
--quiet          -- Do not print cargo log messages
--target         -- Remove from target-dependencies
--verbose        -- Use verbose output (-vv very verbose/build.rs output)
-Z               -- Unstable (nightly-only) flags to Cargo, see 'cargo -Z help' for details
rand        serde_json

```

(on zsh)

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -2,6 +2,7 @@ use crate::core::compiler::{
     BuildConfig, CompileKind, MessageFormat, RustcTargetData, TimingOutput,
 };
 use crate::core::resolver::{CliFeatures, ForceAllTargets, HasDevUnits};
+use crate::core::Dependency;
 use crate::core::{profiles::Profiles, shell, Edition, Package, Target, TargetKind, Workspace};
 use crate::ops::lockfile::LOCKFILE_NAME;
 use crate::ops::registry::RegistryOrIndex;
@@ -23,8 +24,10 @@ use cargo_util_schemas::manifest::RegistryName;
 use cargo_util_schemas::manifest::StringOrVec;
 use clap::builder::UnknownArgumentValueParser;
 use home::cargo_home_with_cwd;
+use indexmap::IndexSet;
+use itertools::Itertools;
 use semver::Version;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::ffi::{OsStr, OsString};
 use std::path::Path;
 use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


src/cargo/util/command_prelude.rs
```diff
@@ -1404,6 +1405,30 @@ fn get_packages() -> CargoResult<Vec<Package>> {
...
+    let cwd = std::env::current_dir()?;
+    let gctx = GlobalContext::new(shell::Shell::new(), cwd.clone(), cargo_home_with_cwd(&cwd)?);
+    let ws = Workspace::new(&find_root_manifest_for_wd(&cwd)?, &gctx)?;
+
+    let dependencies = ws
+        .members()
+        .flat_map(|pkg| pkg.dependencies().into_iter().cloned())
+        .collect::<HashSet<_>>()
+        .into_iter()
+        .collect::<Vec<_>>();
```
epage: Since we put this through a `HashSet`, should we do a `dependencies.sort()`?

epage: We could also do a fancier sort, like put deps within the current package first and then put deps for all workspace members.  Thoughts?

motorailgun: Good catch, will do!

src/cargo/util/command_prelude.rs
```diff
@@ -1404,6 +1406,53 @@ fn get_packages() -> CargoResult<Vec<Package>> {
...
+        .map(|dep| dep.package_name().to_string().into())
+        .sorted()
+        .collect_vec();
+    let all_package_deps_package_names = all_package_deps
+        .into_iter()
+        .map(|dep| dep.package_name().to_string().into())
+        .sorted()
+        .collect_vec();
+
+    current_package_deps_package_names.extend(all_package_deps_package_names);
```
epage: This can end up with duplicates.

Should we remove the `HashSet` in get_dependencies_from_metadata` and store the content in this function in an `IndexSet` so we'll preserve the sorted order of the current package and everything else after that?

motorailgun: Oops! Nice to know `IndexSet`... I'm going to fix this.

epage (approved):

epage:
Thanks!

PR #11969, opened 2023-04-13
Support `per-pkg-target` for `-Zbuild-std`, take two

fee1-dead:
This is a simplified minimal change, which changes the build-std to iterate over the packages to collect a list of targets std should be built for. Does the iterating over packages part have issues with artifact dependencies? I am not very familiar with artifact deps to know so any help would be appreciated. But it should be fine for now since artifact dependencies don't work with build-std yet. #10444

Issue #9451, opened 2021-05-03
The new `per-package-target` feature does not work with `-Zbuild-std`


phil-opp:
**Problem**

I just tried to use the new `per-package-target` feature implemented in https://github.com/rust-lang/cargo/pull/9030 ([tracking issue](https://github.com/rust-lang/cargo/issues/9406)) together with `-Zbuild-std`, but it results in an `error: -Zbuild-std requires --target`.

**Steps**

1. Add a `default-target` or `forced-target` key in `Cargo.toml`.
2. Try to compile by running `cargo build -Zbuild-std=core`.

**Possible Solution(s)**

I _think_ the issue is that the target check already happens here: https://github.com/rust-lang/cargo/blob/db741ac733f3744c2d7ee593d5a01b18e6158b69/src/cargo/ops/cargo_compile.rs#L394-L399

But the `default-target`/`forced-target` fields are only considered later in `generate_targets`: https://github.com/rust-lang/cargo/blob/db741ac733f3744c2d7ee593d5a01b18e6158b69/src/cargo/ops/cargo_compile.rs#L925-L943

The `generate_targets` function is invoked here: https://github.com/rust-lang/cargo/blob/db741ac733f3744c2d7ee593d5a01b18e6158b69/src/cargo/ops/cargo_compile.rs#L470-L488
**Notes**

Output of `cargo version`:

```
cargo 1.53.0-nightly (4369396ce 2021-04-27)
release: 1.53.0
commit-hash: 4369396ce7d270972955d876eaa4954bea56bcd9
commit-date: 2021-04-27
```

cc @Ekleog

kennystrawnmusic:
Moving `let mut units` from line 475 up to line 387 should fix this.

kennystrawnmusic:
Submitted rust-lang/cargo:10084 as a fix for this.

kennystrawnmusic:
Submitted Pull Request #10084 as a fix for this.

kennystrawnmusic:
[...]

PR #15670, opened 2025-06-16
allow `build-std` to be configured per-target in `.cargo/config.toml`

fee1-dead:

`-Zbuild-std` has some issues with per-pkg-target and bindeps, as seen with #10444 and #9451.

At the same time, since these features essentially enable builds where crates with different targets can co-exist, some people may want to use build-std for one target (in my own project: `amdgcn-amd-amdhsa`, for OS dev people: `x86_64-unknown-none`) but not for another target where std is already shipped: #8733

This PR overhauls the way build-std interacts with targets and we're able to fix the issues described above by introducing an additional per-target config in `.cargo/config.toml` for `build-std`. Note that these configs co-exist with the global `-Zbuild-std` flag, and has different behavior whether or not the two are used together: if the global flag is enabled, then all targets will attempt to build their std crates, with per-target config as an override. If the global flag is not used, then only targets that are configured via per-target config will try to build std crates.

This supercedes my two previous attempts at resolving the same issue: #10330 and #11969

Use `.cargo/config.toml`:

```toml
[target.TARGET]
build-std = ['core']
build-std-features = ['default']
```

I've already included some essential tests in `tests/testsuite/standard_lib.rs` such as intersection with per-pkg-target and bindeps. Feel free to suggest additional test cases.

I haven't run the cross compile test locally as I don't know how to get an environment for it on NixOS, will (ab)use the CI for this I guess.

Some unresolved questions:

* Should we introduce and use `[target.TARGET.unstable]` instead?
* Is there a remaining use for the global flags or maybe we can consider removing the existing `-Zbuild-std` way of usage?
* How important is it that we gate `target.TARGET.build-std = 'xxx'` when `TARGET` isn't being built? (thus not parsed in the current state)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -256,6 +256,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - run: rustup update nightly && rustup default nightly
+    - run: rustup target add aarch64-unknown-none
     - run: rustup component add rust-src
     - run: cargo build
     - run: cargo test -p cargo --test build-std


fee1-dead:
cc @phil-opp if you want to do some testing/review :)

ehuss:
Thanks for the PR! However, I don't think this is something we want to move forward with at this time. We are [currently working on the build-std design](https://rust-lang.github.io/rust-project-goals/2025h1/build-std.html), and the design is in flux and may not go in this direction. We probably want to hold off on making design changes at this time. Also, generally we prefer to [have designs approved](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo) before moving to the implementation. I'm going to go ahead and close this for now. After we have made more progress on getting agreement on the direction for build-std, we may revisit exactly how this may work.

fee1-dead:
In that case, could you please keep this open in case this _is_ the direction to go in?

cc @adamgemmell and @davidtwco

fee1-dead:
Note that also it doesn't look like there was active design work in the area. Nor does it appear that there is a solution emerging from the previous discussions. I'm going to comment on the Zulip thread.

PR #15673, opened 2025-06-16
Add `[hints]` table in `Cargo.toml`, and a `hints.mostly-unused` hint

joshtriplett:
The `[hints]` table in a `Cargo.toml` manifest provides optional
information that Cargo can use for building the package, and will use
even when using the package as a dependency. All hints can be safely
ignored, and Cargo only warns about unknown hints, but does not error.
This allows packages to use hints without depending on new Cargo.

Add a `mostly-unused` hint, which allows a package to hint that most
users of the package will not use most of its items. This is useful for
improving the build performance of crates with large dependencies.

Crates can override this hint using `hint-mostly-unused = false` in
their profile for a dependency.

---

In the future, this same mechanism could be used for other hints, such as
`min-opt-level`.

This PR is built atop https://github.com/rust-lang/cargo/pull/15643 . I'd
suggest reviewing that PR separately, then just reviewing the new commits in
this PR.

The new "hints" testsuite module demonstrates the expected behavior of hints.

Like https://github.com/rust-lang/cargo/pull/15643 , the nightly-only tests
will only pass once https://github.com/rust-lang/rust/pull/135656 has been
merged into Rust.

### DIFF
--- a/crates/cargo-util-schemas/src/manifest/mod.rs
+++ b/crates/cargo-util-schemas/src/manifest/mod.rs
@@ -1650,7 +1650,11 @@ pub enum TomlLintLevel {
 #[serde(rename_all = "kebab-case")]
 #[cfg_attr(feature = "unstable-schema", derive(schemars::JsonSchema))]
 pub struct Hints {
-    pub mostly_unused: Option<bool>,
+    #[cfg_attr(
+        feature = "unstable-schema",
+        schemars(with = "Option<TomlValueWrapper>")
+    )]
+    pub mostly_unused: Option<toml::Value>,
 }
 
 #[derive(Copy, Clone, Debug)]
--- a/crates/cargo-util-schemas/src/manifest/mod.rs
+++ b/crates/cargo-util-schemas/src/manifest/mod.rs
@@ -56,6 +56,7 @@ pub struct TomlManifest {
     pub build_dependencies2: Option<BTreeMap<PackageName, InheritableDependency>>,
     pub target: Option<BTreeMap<String, TomlPlatform>>,
     pub lints: Option<InheritableLints>,
+    pub hints: Option<Hints>,
 
     pub workspace: Option<TomlWorkspace>,
     pub profile: Option<TomlProfiles>,
@@ -85,6 +86,7 @@ impl TomlManifest {
                 .map(|_| "build-dependencies"),
             self.target.as_ref().map(|_| "target"),
             self.lints.as_ref().map(|_| "lints"),
+            self.hints.as_ref().map(|_| "hints"),
         ]
         .into_iter()
         .flatten()
@@ -1642,6 +1644,13 @@ pub enum TomlLintLevel {
     Deny,
     Warn,
     Allow,
+}
+
+#[derive(Serialize, Deserialize, Debug, Default, Clone)]
+#[serde(rename_all = "kebab-case")]
+#[cfg_attr(feature = "unstable-schema", derive(schemars::JsonSchema))]
+pub struct Hints {
+    pub mostly_unused: Option<bool>,
 }
 
 #[derive(Copy, Clone, Debug)]
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1135,7 +1135,7 @@ fn build_base_args(
         strip,
         rustflags: profile_rustflags,
         trim_paths,
-        hint_mostly_unused,
+        hint_mostly_unused: profile_hint_mostly_unused,
         ..
     } = unit.profile.clone();
     let hints = unit.pkg.hints().cloned().unwrap_or_default();
@@ -1327,15 +1327,29 @@ fn build_base_args(
         opt(cmd, "-C", "incremental=", Some(dir));
     }
 
-    if hint_mostly_unused.or(hints.mostly_unused).unwrap_or(false) {
+    let pkg_hint_mostly_unused = match hints.mostly_unused {
+        None => None,
+        Some(toml::Value::Boolean(b)) => Some(b),
+        Some(v) => {
+            bcx.gctx.shell().warn(format!(
+                "ignoring unsupported value type ({}) for 'hints.mostly-unused', which expects a boolean",
+                v.type_str()
+            ))?;
+            None
+        }
+    };
+    if profile_hint_mostly_unused
+        .or(pkg_hint_mostly_unused)
+        .unwrap_or(false)
+    {
         if bcx.gctx.cli_unstable().profile_hint_mostly_unused {
             cmd.arg("-Zhint-mostly-unused");
         } else {
-            if hint_mostly_unused.is_some() {
+            if profile_hint_mostly_unused.is_some() {
                 bcx.gctx
                     .shell()
                     .warn("ignoring 'hint-mostly-unused' profile option, pass `-Zprofile-hint-mostly-unused` to enable it")?;
-            } else if hints.mostly_unused.is_some() {
+            } else if pkg_hint_mostly_unused.is_some() {
                 bcx.gctx
                     .shell()
                     .warn("ignoring 'hints.mostly-unused', pass `-Zprofile-hint-mostly-unused` to enable it")?;
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1148,6 +1148,13 @@ fn build_base_args(
             unit.pkg.package_id().version()
         ))
     };
+    let unit_capped_warn = |msg: &str| {
+        if unit.show_warnings(bcx.gctx) {
+            warn(msg)
+        } else {
+            Ok(())
+        }
+    };
 
     cmd.arg("--crate-name").arg(&unit.target.crate_name());
 
@@ -1339,7 +1346,7 @@ fn build_base_args(
         None => None,
         Some(toml::Value::Boolean(b)) => Some(b),
         Some(v) => {
-            warn(&format!(
+            unit_capped_warn(&format!(
                 "ignoring unsupported value type ({}) for 'hints.mostly-unused', which expects a boolean",
                 v.type_str()
             ))?;
@@ -1354,11 +1361,12 @@ fn build_base_args(
             cmd.arg("-Zhint-mostly-unused");
         } else {
             if profile_hint_mostly_unused.is_some() {
+                // Profiles come from the top-level unit, so we don't use `unit_capped_warn` here.
                 warn(
                     "ignoring 'hint-mostly-unused' profile option, pass `-Zprofile-hint-mostly-unused` to enable it",
                 )?;
             } else if pkg_hint_mostly_unused.is_some() {
-                warn(
+                unit_capped_warn(
                     "ignoring 'hints.mostly-unused', pass `-Zprofile-hint-mostly-unused` to enable it",
                 )?;
             }
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1141,6 +1141,14 @@ fn build_base_args(
     let hints = unit.pkg.hints().cloned().unwrap_or_default();
     let test = unit.mode.is_any_test();
 
+    let warn = |msg: &str| {
+        bcx.gctx.shell().warn(format!(
+            "{}@{}: {msg}",
+            unit.pkg.package_id().name(),
+            unit.pkg.package_id().version()
+        ))
+    };
+
     cmd.arg("--crate-name").arg(&unit.target.crate_name());
 
     let edition = unit.target.edition();
@@ -1331,7 +1339,7 @@ fn build_base_args(
         None => None,
         Some(toml::Value::Boolean(b)) => Some(b),
         Some(v) => {
-            bcx.gctx.shell().warn(format!(
+            warn(&format!(
                 "ignoring unsupported value type ({}) for 'hints.mostly-unused', which expects a boolean",
                 v.type_str()
             ))?;
@@ -1346,13 +1354,13 @@ fn build_base_args(
             cmd.arg("-Zhint-mostly-unused");
         } else {
             if profile_hint_mostly_unused.is_some() {
-                bcx.gctx
-                    .shell()
-                    .warn("ignoring 'hint-mostly-unused' profile option, pass `-Zprofile-hint-mostly-unused` to enable it")?;
+                warn(
+                    "ignoring 'hint-mostly-unused' profile option, pass `-Zprofile-hint-mostly-unused` to enable it",
+                )?;
             } else if pkg_hint_mostly_unused.is_some() {
-                bcx.gctx
-                    .shell()
-                    .warn("ignoring 'hints.mostly-unused', pass `-Zprofile-hint-mostly-unused` to enable it")?;
+                warn(
+                    "ignoring 'hints.mostly-unused', pass `-Zprofile-hint-mostly-unused` to enable it",
+                )?;
             }
         }
     }
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -1138,6 +1138,7 @@ fn build_base_args(
         hint_mostly_unused,
         ..
     } = unit.profile.clone();
+    let hints = unit.pkg.hints().cloned().unwrap_or_default();
     let test = unit.mode.is_any_test();
 
     cmd.arg("--crate-name").arg(&unit.target.crate_name());
@@ -1326,13 +1327,19 @@ fn build_base_args(
         opt(cmd, "-C", "incremental=", Some(dir));
     }
 
-    if hint_mostly_unused {
+    if hint_mostly_unused.or(hints.mostly_unused).unwrap_or(false) {
         if bcx.gctx.cli_unstable().profile_hint_mostly_unused {
             cmd.arg("-Zhint-mostly-unused");
         } else {
-            bcx.gctx
-                .shell()
-                .warn("ignoring 'hint-mostly-unused' profile option, pass `-Zprofile-hint-mostly-unused` to enable it")?;
+            if hint_mostly_unused.is_some() {
+                bcx.gctx
+                    .shell()
+                    .warn("ignoring 'hint-mostly-unused' profile option, pass `-Zprofile-hint-mostly-unused` to enable it")?;
+            } else if hints.mostly_unused.is_some() {
+                bcx.gctx
+                    .shell()
+                    .warn("ignoring 'hints.mostly-unused', pass `-Zprofile-hint-mostly-unused` to enable it")?;
+            }
         }
     }
 
--- a/src/cargo/core/profiles.rs
+++ b/src/cargo/core/profiles.rs
@@ -578,7 +578,7 @@ fn merge_profile(profile: &mut Profile, toml: &TomlProfile) {
         profile.trim_paths = Some(trim_paths.clone());
     }
     if let Some(hint_mostly_unused) = toml.hint_mostly_unused {
-        profile.hint_mostly_unused = hint_mostly_unused;
+        profile.hint_mostly_unused = Some(hint_mostly_unused);
     }
     profile.strip = match toml.strip {
         Some(StringOrBool::Bool(true)) => Strip::Resolved(StripInner::Named("symbols".into())),
@@ -629,8 +629,8 @@ pub struct Profile {
     // remove when `-Ztrim-paths` is stablized
     #[serde(skip_serializing_if = "Option::is_none")]
     pub trim_paths: Option<TomlTrimPaths>,
-    #[serde(skip_serializing_if = "std::ops::Not::not")]
-    pub hint_mostly_unused: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub hint_mostly_unused: Option<bool>,
 }
 
 impl Default for Profile {
@@ -652,7 +652,7 @@ impl Default for Profile {
             strip: Strip::Deferred(StripInner::None),
             rustflags: vec![],
             trim_paths: None,
-            hint_mostly_unused: false,
+            hint_mostly_unused: None,
         }
     }
 }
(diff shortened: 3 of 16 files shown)


jeffparsons:
Sorry if this is the wrong place for this, but I'd love to see how this affects a simple binary using the `aws-sdk-ec2` crate. I think if it is able to shave much time / memory consumption off those "worst offenders" then that would make a pretty compelling case for its value.

src/cargo/core/profiles.rs
```diff
@@ -577,9 +577,7 @@ fn merge_profile(profile: &mut Profile, toml: &TomlProfile) {
     if let Some(trim_paths) = &toml.trim_paths {
         profile.trim_paths = Some(trim_paths.clone());
     }
-    if let Some(hint_mostly_unused) = toml.hint_mostly_unused {
-        profile.hint_mostly_unused = hint_mostly_unused;
-    }
```
epage: Why this change?

tests/testsuite/hints.rs
```diff
@@ -0,0 +1,245 @@
...
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+            [package]
+            name = "foo"
+            version = "0.0.1"
+            edition = "2015"
+
+            [hints]
```
epage, line 18: On stable, shouldn't this warn about being unused?  Otherwise, this is in a way insta-stabilizing this field

tests/testsuite/hints.rs
epage, line 1: Mind moving the tests into a prior commit?  That way this commit will show what impact adding support for `[hints]` has on the tests.

src/doc/src/reference/unstable.md
```diff
@@ -924,6 +924,36 @@ profile-rustflags = true
 rustflags = [ "-C", "..." ]
 ```
 
+## Profile `hint-mostly-unused` option
```
ehuss: Be sure to also include a link to this from the index at the top of the file.

joshtriplett:
Note: should add a test for `[hints]` parsing to robustly handle a key it knows but syntax it doesn't know (because it might be future syntax).

src/cargo/core/profiles.rs
joshtriplett: @epage The previous code parsed the hint as an `Option<bool>`, but then the profile converted it to a boolean with a default of `false`. This change makes the profile preserve the `Option<bool>` type, with a default of `None`. This supports the logic where the profile overrides the hint, which requires distinguishing between `Some(false)` (which would override a hint of `true`) and `None`.

joshtriplett:
> Sorry if this is the wrong place for this, but I'd love to see how this affects a simple binary using the `aws-sdk-ec2` crate. I think if it is able to shave much time / memory consumption off those "worst offenders" then that would make a pretty compelling case for its value.

For a crate depending on aws-sdk-ec2, a release build went from 4m07s to 2m04s, and a non-release build went from 2m26s to 1m28s.

tests/testsuite/hints.rs
joshtriplett: @epage As discussed, that's the intention here: `[hints]` are always best-effort and not guaranteed, unknown hints produce a warning but no error, and known hints don't produce a warning. The individual hints, if specified, will have feature-gated behavior; for instance, there's a warning if you specify `hints.mostly-unused` but don't enable the feature gate. So, any given hint will typically go from "unknown key" to "known key, but specifying it gets ignored unless you specify the feature gate".

src/cargo/core/profiles.rs
epage: Except we call `merge_profile` multiple times, so one `profile` table not having it set will unset it, rather than leaving it to a different level to deal with it.

src/doc/src/reference/unstable.md
```diff
@@ -924,6 +924,36 @@ profile-rustflags = true
...
+crate's `Cargo.toml` without mandating the use of a newer Cargo to build it.
+
+A crate can also provide this hint automatically for crates that depend on it,
+using the `[hints]` table (which will likewise be ignored by older Cargo):
+
+```toml
+[hints]
+mostly-unused = true
+```
+
```
epage, line 973: If we're insta-stabilizing `[hints]`, we should add it to the stable docs

src/doc/src/reference/unstable.md
```diff
@@ -924,6 +924,36 @@ profile-rustflags = true
...
+will otherwise function without erroring. This allows using the hint in a
+crate's `Cargo.toml` without mandating the use of a newer Cargo to build it.
+
+A crate can also provide this hint automatically for crates that depend on it,
+using the `[hints]` table (which will likewise be ignored by older Cargo):
+
+```toml
+[hints]
+mostly-unused = true
+```
```
epage, line 972: Do we need to decide on workspace inheritance before insta-stabilizing?

epage, line 972: As a reminder for us to hold an FCP for this for being insta-stabilized

src/doc/src/reference/unstable.md
```diff
@@ -924,6 +924,36 @@ profile-rustflags = true
...
+introduction of this feature will give an "unused manifest key" warning, but
+will otherwise function without erroring. This allows using the hint in a
+crate's `Cargo.toml` without mandating the use of a newer Cargo to build it.
+
+A crate can also provide this hint automatically for crates that depend on it,
+using the `[hints]` table (which will likewise be ignored by older Cargo):
+
+```toml
+[hints]
+mostly-unused = true
```
epage, line 971: Please document the precedence of `mostly-unused`

epage: Hmm, will we run into problems with `min-opt-level` to carry this precedence design story forward?

src/doc/src/reference/unstable.md
joshtriplett: I don't *think* we do, as long as I fix the parsing to handle arbitrary Toml types (e.g. tables).

src/doc/src/reference/unstable.md
joshtriplett: Done.

src/doc/src/reference/unstable.md
joshtriplett: Precedence is documented.

I don't expect that `min-opt-level` will cause precedence issues. I would expect `hints.min-opt-level` to be overridden by an `opt-level` explicitly specified in a profile; I don't think that'll be particularly hard to implement.

src/doc/src/reference/unstable.md
joshtriplett: Which I've now done.

src/cargo/core/profiles.rs
joshtriplett: @epage Good catch. Fixed.

src/doc/src/reference/unstable.md
joshtriplett: Done.

tests/testsuite/hints.rs
joshtriplett: I *think* I figured out what you mean here. I've added a version of the hints tests prior to the introduction of `[hints]`, to show what prior versions of Cargo will do with them; the introduction of `[hints]` then modifies the tests to reflect the new behavior of Cargo.

tests/testsuite/hints.rs
```diff
@@ -49,11 +49,197 @@ fn unknown_hints_warn() {
...
 [CHECKING] foo v0.0.1 ([ROOT]/foo)
 [RUNNING] `rustc --crate-name foo [..]`
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 
 "#]])
         .run();
 }
+
+#[cargo_test]
+fn hints_mostly_unused_warn_without_gate() {
```
epage, line 139: I had meant for all of these tests to be added in the prior test commit

tests/testsuite/hints.rs
```diff
@@ -58,6 +58,56 @@ fn unknown_hints_warn() {
         .run();
 }
 
+#[cargo_test]
+fn hint_unknown_type_warn() {
```
epage, line 86: ditto about adding this in the test commit

tests/testsuite/hints.rs
joshtriplett: Alright, I've now pre-added versions of all the tests, and then evolved them to reflect the desired behavior in the commit adding that behavior.

tests/testsuite/hints.rs
joshtriplett: Alright, I've now pre-added versions of all the tests, and then evolved them to reflect the desired behavior in the commit adding that behavior.

src/cargo/core/compiler/mod.rs
```diff
@@ -1320,6 +1322,36 @@ fn build_base_args(
         opt(cmd, "-C", "incremental=", Some(dir));
     }
 
+    let pkg_hint_mostly_unused = match hints.mostly_unused {
+        None => None,
+        Some(toml::Value::Boolean(b)) => Some(b),
+        Some(v) => {
+            bcx.gctx.shell().warn(format!(
+                "ignoring unknown value type ({}) for 'hints.mostly-unused'",
```
epage:
```suggestion
                "ignoring unsupported value type ({}) for 'hints.mostly-unused' which expects a boolean",
```

joshtriplett: Good idea; fixed.

epage (approved):
Just waiting on #15643

epage:
@rfcbot fcp merge

This insta-stabilizes a new `[hints]` table to `Cargo.toml` that affects building this as a dependency.  As these are just hints, there are no guarantees about what will actually be done, including us removing the table completely.  As such, unstable fields do not error and wrongly-shaped fields do not error.

While this is blocked on rustc, I figured we could run our FCP in parallel.

src/cargo/core/compiler/mod.rs
```diff
@@ -1325,6 +1327,36 @@ fn build_base_args(
         opt(cmd, "-C", "incremental=", Some(dir));
     }
 
+    let pkg_hint_mostly_unused = match hints.mostly_unused {
+        None => None,
+        Some(toml::Value::Boolean(b)) => Some(b),
+        Some(v) => {
+            bcx.gctx.shell().warn(format!(
+                "ignoring unsupported value type ({}) for 'hints.mostly-unused', which expects a boolean",
```
ehuss, line 1350: Can this warning (and the ones below) include the package id? Otherwise I don't think the user would be able to know where this is coming from.

src/doc/src/reference/manifest.md
```diff
@@ -565,6 +566,21 @@ As for dependents, Cargo suppresses lints from non-path dependencies with featur
...
+The `[hints]` section allows specifying hints for compiling this crate, which
+crates depending on this one will use by default but may override. Hints are,
+by design, always safe for Cargo to ignore; if Cargo encounters a hint it
+doesn't understand, or a hint it understands but with a value it doesn't
+understand, it will warn, but not error.
+
+Individual hints may have an associated unstable feature gate that you need to
+pass in order to apply the configuration they specify, but if you don't specify
+that unstable feature gate, you will again get only a warning, not an error.
+
```
ehuss, line 582: Can this say something about "there are no stable hints at this time, see *link to unstable hint-mostly-unused*".

Otherwise this is telling the user that there is a mysterious table with nothing in it.

joshtriplett: Done.

src/cargo/core/compiler/mod.rs
joshtriplett: Done.

joshtriplett:
Another fun benchmark number for this, using `--timings`:

I built a crate that depended on the `windows` crate, with all of its "Graphics*" and "UI*" features enabled.

Without hint-mostly-unused, `windows` took 15.1s, of which 49% was codegen.

With hint-mostly-unused, `windows` took 7.5s, of which 1% was codegen.

weihanglo (approved):

src/doc/src/reference/manifest.md
```diff
@@ -565,6 +566,25 @@ As for dependents, Cargo suppresses lints from non-path dependencies with featur
 
 > **MSRV:** Respected as of 1.74
 
+## The `[hints]` section
+
+The `[hints]` section allows specifying hints for compiling this crate, which
+crates depending on this one will use by default but may override. Hints are,
```
kpreid: I believe that “crates depending on this one” is an ambiguous or misleading phrase here.

I assume that it actually means that *the workspace/package the `cargo` command was executed on* may override the effect of this hint on this package in its `[profile]`. However, I expect that many will read it to mean that it has some sort of effect that applies to *direct dependents*, which (I assume) it does not.

(Also, both occurrences of “crate” should be “package”, should they not?)

kpreid (dismissed):

epage:
@joshtriplett looks like something got messed up in a rebase?

joshtriplett:
@epage So it appears. Fixing.

src/cargo/core/compiler/mod.rs
```diff
@@ -1362,9 +1362,13 @@ fn build_base_args(
         } else {
             if profile_hint_mostly_unused.is_some() {
                 // Profiles come from the top-level unit, so we don't use `unit_capped_warn` here.
-                warn("ignoring 'hint-mostly-unused' profile option, pass `-Zprofile-hint-mostly-unused` to enable it")?;
+                warn(
```
epage, line 1365: nit: this should be squashed into the appropriate commit

tests/testsuite/hints.rs
```diff
@@ -32,6 +32,21 @@ fn empty_hints_no_warn() {
 
 #[cargo_test]
 fn unknown_hints_warn() {
+    Package::new("bar", "1.0.0")
```
epage, line 35: nit: test changes like this should be made before any of the feature work so that we can tell just by looking at the tests what the behavior change is.  By changing the test, it obscures that.

(granted, in some cases, changing the test is part of showcasing the feature kind of like the `rerun_untracks_other_files` test in a24abd6fc2347b3c0de0a96be40bf98df4ea030f

crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -1650,7 +1650,11 @@ pub enum TomlLintLevel {
 #[serde(rename_all = "kebab-case")]
 #[cfg_attr(feature = "unstable-schema", derive(schemars::JsonSchema))]
 pub struct Hints {
-    pub mostly_unused: Option<bool>,
+    #[cfg_attr(
+        feature = "unstable-schema",
+        schemars(with = "Option<TomlValueWrapper>")
+    )]
+    pub mostly_unused: Option<toml::Value>,
```
epage, line 1657: nit: you could do `schemars(with = "Option<bool>")` to communicate the effective schema even if the types take the long way around to get to it

src/doc/src/reference/manifest.md
```diff
@@ -565,6 +566,26 @@ As for dependents, Cargo suppresses lints from non-path dependencies with featur
...
+Individual hints may have an associated unstable feature gate that you need to
+pass in order to apply the configuration they specify, but if you don't specify
+that unstable feature gate, you will again get only a warning, not an error.
+
+There are no stable hints at this time. See the [hint-mostly-unused
+documentation](unstable.md#profile-hint-mostly-unused-option) for information
+on an unstable hint.
+
+> **MSRV:** Specifying hints does not impact MSRV. Older versions of Cargo will
+> ignore hints with a warning, but will not error.
```
epage: The way we normally word this is what version is the field "respected".  The lints table is worded that way.

crates/cargo-util-schemas/src/manifest/mod.rs
```diff
@@ -56,6 +56,7 @@ pub struct TomlManifest {
     pub build_dependencies2: Option<BTreeMap<PackageName, InheritableDependency>>,
     pub target: Option<BTreeMap<String, TomlPlatform>>,
     pub lints: Option<InheritableLints>,
+    pub hints: Option<Hints>,
```
epage, line 59: The job "CI / check-version-bump" is pointing out that we've not bumped the breaking version for `cargo-util-schemas` this release cycle and we need to do so because `TomlManifest` is exhaustive.

epage:
Sorry for the last minute finds

src/cargo/core/compiler/mod.rs
joshtriplett: Done.

tests/testsuite/hints.rs
joshtriplett: I don't think doing so provided much value in this case, but nonetheless, done.

crates/cargo-util-schemas/src/manifest/mod.rs
joshtriplett: If I do that, the schema then says it takes a boolean or null, and I think that's incorrect; anyone relying on that schema would become fragile and potentially break if we introduce new possible values for this field. I think the schema *should* say that it takes a TomlValue or null.

The schema shouldn't encourage people to be brittle, when cargo is going out of its way to be permissive and allow for future evolution without it being a breaking change.

src/doc/src/reference/manifest.md
joshtriplett: Done.

crates/cargo-util-schemas/src/manifest/mod.rs
joshtriplett: As far as I can tell, isn't that test comparing against the base of the PR, rather than the last major version bump?

I can add a major version bump in this PR, but if I'm understanding the workflow correctly, a future breaking change to the schema *in this same cycle* would still fail this CI test, right?

joshtriplett: I've added a cargo-util-schemas version bump to this PR.

crates/cargo-util-schemas/src/manifest/mod.rs
epage: I think its dependent on how the schema is being used, whether its prescriptively match our compatibility guarantees or to describe the behavior at this time.  For example, some editors support reading these schemas and applying completions off of it.

I lean towards the latter but not going to push it.  We can always adjust later.

src/doc/src/reference/manifest.md
```diff
@@ -565,6 +566,27 @@ As for dependents, Cargo suppresses lints from non-path dependencies with featur
...
+pass in order to apply the configuration they specify, but if you don't specify
+that unstable feature gate, you will again get only a warning, not an error.
+
+There are no stable hints at this time. See the [hint-mostly-unused
+documentation](unstable.md#profile-hint-mostly-unused-option) for information
+on an unstable hint.
+
+> **MSRV:** Respected as of 1.90. However, specifying hints in a crate does not
+> impact the MSRV of the crate. Older versions of Cargo will ignore hints with
+> a warning, but will not error.
```
epage: The `However` is describing what we mean by "respected".  This place doesn't feel any different than any other for us to go into detail on what we mean.

crates/cargo-util-schemas/src/manifest/mod.rs
epage: The job mostly compares against the beta branch.

tests/testsuite/hints.rs
```diff
@@ -32,6 +32,21 @@ fn empty_hints_no_warn() {
...
+            name = "bar"
+            version = "1.0.0"
+            edition = "2015"
+
+            [hints]
+            this-is-an-unknown-hint = true
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .publish();
```
epage, line 49: This commit us publishing a `bar` package but it seems to be unused by this test.

tests/testsuite/hints.rs
```diff
@@ -86,6 +101,9 @@ fn hint_unknown_type_warn() {
 
             [dependencies]
             bar = "1.0"
+
+            [hints]
+            mostly-unused = "string"
```
epage, line 115: nit: there still seem to be test changes in this commit

tests/testsuite/hints.rs
```diff
@@ -136,6 +155,9 @@ fn hints_mostly_unused_warn_without_gate() {
 
             [dependencies]
             bar = "1.0"
+
+            [hints]
+            mostly-unused = true
```
epage, line 168: here also

crates/cargo-util-schemas/src/manifest/mod.rs
joshtriplett: @epage I really don't know how the schemas are intended to be used. I don't object to this change, I just don't know how we expect people to handle compatibility with anything that gets generated based on this.

src/doc/src/reference/manifest.md
joshtriplett: I think that's a sign that we may want to update *other* MSRV tags like this. But in lieu of a systematic effort there, I've moved the relevant and non-duplicated portion to the body of this section.

tests/testsuite/hints.rs
joshtriplett: This commit is the one making changes to the test, prior to the subsequent change modifying what Cargo does.

tests/testsuite/hints.rs
joshtriplett: This commit is the one making changes to the test, prior to the subsequent change modifying what Cargo does.

tests/testsuite/hints.rs
joshtriplett: Good catch, fixed.

crates/cargo-util-schemas/src/manifest/mod.rs
joshtriplett: I'm not seeing where in `ci/validate-version-bump.sh` or in the rest of `.github` it's comparing against beta, rather than against the base of the PR.

epage: https://github.com/rust-lang/cargo/blob/eabb4cd923deb73e714f7ad3f5234d68ca284dbe/crates/xtask-bump-check/src/xtask.rs#L281-L306

tests/testsuite/hints.rs
epage: I think there is a misunderstanding of the spirit of what we are trying to accomplish with the "add tests before".

The goal is for commits like this one to show how behavior changed.  That isn't met when you opt-in to the new behavior at the same time as you change behavior.

epage (approved):

PR #15674, opened 2025-06-17
feat: Introduce perma unstable `--compile-time-deps` option for `cargo build`

### LINKED ISSUE #14434: Introduce perma-unstable `--compile-time-deps` flag
Excerpt taken (and slight adjusted) from https://github.com/rust-lang/rfcs/pull/3344:

Add a flag `--compile-time-deps` to `cargo build` that will cause cargo to only build proc-macros and build-scripts (and their required dependencies) of the package graph well as running the build-scripts.

In case of the ecosystem introducing more compile time (executable) artifacts that are required to fully build a project, these artifacts should be included by this flag accordingly.

- [Package Selection](https://doc.rust-lang.org/cargo/commands/cargo-build.html#package-selection) will limit the built artifacts to the selected packages and their dependencies.
- [Target Selection](https://doc.rust-lang.org/cargo/commands/cargo-build.html#target-selection) will limit the built artifacts to the selected targets and their dependencies. This only has a noticeable effect for proc-macro packages, where not selecting their lib target will skip building it.
- [Feature Selection](https://doc.rust-lang.org/cargo/commands/cargo-build.html#feature-selection) apply the selected features to the build as usual.
- [Compilation Options](https://doc.rust-lang.org/cargo/commands/cargo-build.html#compilation-options) have no effect on the compiled artifacts. This is the same as a plain `cargo build` where build scripts and proc-macros are always built for host platform.
- [Output Options](https://doc.rust-lang.org/cargo/commands/cargo-build.html#output-options) affect the locations of the generated artifacts as usual.
- [Display Options](https://doc.rust-lang.org/cargo/commands/cargo-build.html#display-options)  affect the build as usual.
- [Manifest Options](https://doc.rust-lang.org/cargo/commands/cargo-build.html#manifest-options) affect the build as usual.
- [Miscellaneous Options](https://doc.rust-lang.org/cargo/commands/cargo-build.html#miscellaneous-options) affect the build as usual.

Note "as usual" here refers to the behavior of a `cargo build` invocation without the proposed flag.

The main reasoning behind this is that rust-analyzer (and likely RustRover) want to get rid of their buggy, incorrect, and cargo breaking `RUSTC_WRAPPER` asap. Though as this flag requires a lot of use case collecting (outside of the IDE space) and general design iterations stabilizing it (especially in the proposed form) will take quite a long time either way so we mainly want to introduce this in some unstable form for these tools to consume for now. That does mean the IDE will likely have to employ some shenanigans to make use of this flag on stable (via `RUSTC_BOOTSTRAP=1` for example) and then keeping a `RUSTC_WRAPPER` that will strip that flag again before building (as to not break cursed build scripts probes).

Created this as per https://github.com/rust-lang/rfcs/pull/3344#issuecomment-2299220793.

ShoyuVanilla:
Resolves #14434

### DIFF
--- a/src/cargo/core/compiler/build_config.rs
+++ b/src/cargo/core/compiler/build_config.rs
@@ -50,6 +50,8 @@ pub struct BuildConfig {
     pub timing_outputs: Vec<TimingOutput>,
     /// Output SBOM precursor files.
     pub sbom: bool,
+    /// Build compile time dependencies only, e.g., build scripts and proc macros
+    pub compile_time_deps_only: bool,
 }
 
 fn default_parallelism() -> CargoResult<u32> {
@@ -129,6 +131,7 @@ impl BuildConfig {
             future_incompat_report: false,
             timing_outputs: Vec::new(),
             sbom,
+            compile_time_deps_only: false,
         })
     }
 
--- a/src/cargo/core/manifest.rs
+++ b/src/cargo/core/manifest.rs
@@ -1079,6 +1079,11 @@ impl Target {
         *self.kind() == TargetKind::CustomBuild
     }
 
+    /// Returns `true` if it is a compile time depencencies, e.g., build script or proc macro
+    pub fn is_compile_time_dependency(&self) -> bool {
+        self.is_custom_build() || self.proc_macro()
+    }
+
     /// Returns the arguments suitable for `--crate-type` to pass to rustc.
     pub fn rustc_crate_types(&self) -> Vec<CrateType> {
         self.kind().rustc_crate_types()
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -2190,3 +2190,17 @@ More information can be found in the [config chapter](config.md#cache).
 ## doctest-xcompile
 
 Doctest cross-compiling is now unconditionally enabled starting in Rust 1.89. Running doctests with `cargo test` will now honor the `--target` flag.
+
+## compile-time-deps
+
+This permanently-unstable flag to only build proc-macros and build scripts (and their required dependencies),
+as well as run the build scripts.
+
+It is intended for use by tools like rust-analyzer and will never be stabilized.
+
+Example:
+
+```console
+cargo +nightly build --compile-time-deps -Z unstable-options
+cargo +nightly check --compile-time-deps --all-targets -Z unstable-options
+```
--- a/tests/testsuite/compile_time_deps.rs
+++ b/tests/testsuite/compile_time_deps.rs
@@ -8,13 +8,11 @@ fn gated_by_unstable_opts() {
         .build();
 
     p.cargo("check --compile-time-deps")
-        .with_status(1)
+        .with_status(101)
         .with_stderr_data(str![[r#"
-[ERROR] unexpected argument '--compile-time-deps' found
-
-Usage: cargo check [OPTIONS]
-
-For more information, try '--help'.
+[ERROR] the `--compile-time-deps` flag is unstable, and only available on the nightly channel of Cargo, but this is the `stable` channel
+See https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.
+See https://github.com/rust-lang/cargo/issues/14434 for more information about the `--compile-time-deps` flag.
 
 "#]])
         .run();
@@ -55,11 +53,10 @@ fn non_comp_time_dep() {
         .file("bar/src/lib.rs", r#"pub fn bar() {}"#)
         .build();
 
-    p.cargo("check")
+    p.cargo("-Zunstable-options check --compile-time-deps")
+        .masquerade_as_nightly_cargo(&["compile-time-deps"])
         .with_stderr_data(str![[r#"
 [LOCKING] 1 package to latest compatible version
-[CHECKING] bar v0.0.1 ([ROOT]/foo/bar)
-[CHECKING] foo v0.0.1 ([ROOT]/foo)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 
 "#]])
@@ -151,11 +148,11 @@ fn proc_macro_dep() {
         .file("baz/src/lib.rs", r#"pub fn baz() {}"#)
         .build();
 
-    p.cargo("check --package foo")
+    p.cargo("-Zunstable-options check --package foo --compile-time-deps")
+        .masquerade_as_nightly_cargo(&["compile-time-deps"])
         .with_stderr_data(str![[r#"
 [COMPILING] baz v0.0.1 ([ROOT]/foo/baz)
 [COMPILING] bar v0.0.1 ([ROOT]/foo/bar)
-[CHECKING] foo v0.0.1 ([ROOT]/foo/foo)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 
 "#]])
@@ -163,10 +160,9 @@ fn proc_macro_dep() {
 
     p.cargo("clean").run();
 
-    p.cargo("check --package bar")
+    p.cargo("-Zunstable-options check --package bar --compile-time-deps")
+        .masquerade_as_nightly_cargo(&["compile-time-deps"])
         .with_stderr_data(str![[r#"
-[CHECKING] baz v0.0.1 ([ROOT]/foo/baz)
-[CHECKING] bar v0.0.1 ([ROOT]/foo/bar)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 
 "#]])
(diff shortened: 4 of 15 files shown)


tests/testsuite/compile_time_deps.rs
```diff
@@ -0,0 +1,381 @@
...
+                [package]
+                name = "baz"
+                version = "0.0.1"
+                edition = "2021"
+            "#,
+        )
+        .file("bar/baz/src/lib.rs", r#"pub fn baz() {}"#)
+        .build();
+
+    p.cargo("-Zunstable-options build --compile-time-deps")
```
epage: Would it work to make these `check`s instead? It speeds up tests and takes up less disk space

src/cargo/core/manifest.rs
```diff
@@ -1078,6 +1078,9 @@ impl Target {
     pub fn is_custom_build(&self) -> bool {
         *self.kind() == TargetKind::CustomBuild
     }
+    pub fn is_compile_time_dependency(&self) -> bool {
```
epage, line 1083: Why no blank line before?

src/doc/man/includes/options-compile-time-deps.md
```diff
@@ -0,0 +1,8 @@
+{{#option "`--compile-time-deps`"}}
```
epage: Why is this hidden in `--help` but shown in `man`?

tests/testsuite/compile_time_deps.rs
epage, line 1: We should have a test that the `-Z` is present

src/cargo/core/compiler/build_config.rs
```diff
@@ -183,6 +183,12 @@ pub enum CompileMode {
     Docscrape,
     /// Execute the binary built from the `build.rs` script.
     RunCustomBuild,
+    /// Build the given unit's build dependencies, proc macro dependencies and
+    /// their transitive dependencies only (skipping itself).
+    ///
+    /// If `test` is true, then it is also compiled with `--test` to include
+    /// dev-dependencies for tests.
+    CompileTimeDeps { test: bool },
```
epage: Why a whole new mode rather than filtering the build graph?

src/cargo/core/manifest.rs
ShoyuVanilla: Because there were no blank lines for other two methods above so I thought there would be some reasons. Will fix them 😅

tests/testsuite/compile_time_deps.rs
ShoyuVanilla: I added this new flag only to `cargo build`. Should this be added to `check` also?

src/doc/man/includes/options-compile-time-deps.md
ShoyuVanilla: I followed [this instruction](https://github.com/rust-lang/cargo/issues/14434#issuecomment-2301847538)

src/cargo/core/compiler/build_config.rs
ShoyuVanilla: Well, I though this would be simpler, because this new feature makes dependency graph somewhat sparse and disconnected, so it might be hard to correctly handle along with sane build graphs.
But if the build graph filtering is more desirable, I'll change the implementations

src/doc/man/includes/options-compile-time-deps.md
ShoyuVanilla: Oh, you've suggested it to be only documented on unstable.md. Sorry, I'll fix that 😅

tests/testsuite/compile_time_deps.rs
ShoyuVanilla: I might have misunderstood but do you mean adding test that triggering this new flag without -Zunstable-options and guarded by unstable error?

ShoyuVanilla:
I've pondered upon this more and decided not to introduce new `UserIntent/CompileMode` and I'll re-request you for reviews once I finished it. Sorry for bothering you

tests/testsuite/compile_time_deps.rs
```diff
@@ -186,6 +203,7 @@ fn build_dep() {
             r#"
                 pub fn bar() {
                     baz::baz();
+                    std::fs::write("build-script-output", "build script run").unwrap();
```
epage: Is there a reason this isn't in the commit that added the tests?

tests/testsuite/compile_time_deps.rs
epage: I guess in this case it doesn't really matter since anything that would be a `check` doesn't happen anyways.

tests/testsuite/compile_time_deps.rs
ShoyuVanilla: I forgot to check whether the build script is triggered in test and adjusted this in the current revision 😅

tests/testsuite/compile_time_deps.rs
epage: Yes, making sure we correctly error on stable rust, rather than accidentally insta-stabilize it

src/doc/src/reference/unstable.md
```diff
@@ -2190,3 +2190,17 @@ More information can be found in the [config chapter](config.md#cache).
 ## doctest-xcompile
 
 Doctest cross-compiling is now unconditionally enabled starting in Rust 1.89. Running doctests with `cargo test` will now honor the `--target` flag.
+
+## compile-time-deps
+
+This permanently-unstable flag only build proc-macros and build scripts (and their required dependencies),
+as well as run the build scripts.
```
epage:
```suggestion
This permanently-unstable flag to only build proc-macros and build scripts (and their required dependencies),
as well as run the build scripts.
```

src/doc/src/reference/unstable.md
```diff
@@ -2190,3 +2190,17 @@ More information can be found in the [config chapter](config.md#cache).
...
+
+It is intended for use by tools like rust-analyzer and will never be stabilized.
+
+Example:
+
+```sh
+cargo +nightly build --compile-time-deps -Z unstable-options
+cargo +nightly check --compile-time-deps --all-targets -Z unstable-options
+```
```
epage:
```suggestion
```console
$ cargo +nightly build --compile-time-deps -Z unstable-options
$ cargo +nightly check --compile-time-deps --all-targets -Z unstable-options
```
```

File: tests/testsuite/compile_time_deps.rs [hunk 8]
```diff
@@ -0,0 +1,381 @@
...
+use cargo_test_support::{project, str};
+
+#[should_panic]
+#[cargo_test]
+fn gated_by_unstable_opts() {
+    let p = project()
+        .file("src/main.rs", r#"fn main() { println!("Hello, World!") }"#)
+        .build();
+
+    p.cargo("check --compile-time-deps").run();
```
epage: This should test specifically for the error, not that an error occurs.  Otherwise, a different error could be making this test "pass" (like it does in the current commit)

ShoyuVanilla: Fixed this in my last commit. Should I also fix the commit introducing this test?

ShoyuVanilla:
Oh, test failure looks strange 🤔 Maybe something related to `Sized` changed in rustc nightly?

```
 test failed running `/Users/runner/work/cargo/cargo/target/debug/cargo build --lib --target custom-target.json -v`
error: process exited with code 101 (expected 0)
--- stdout

--- stderr
   Compiling foo v0.0.1 (/Users/runner/work/cargo/cargo/target/tmp/cit/t1384/foo)
     Running `rustc --crate-name foo --edition=2015 src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no -C debuginfo=2 --check-cfg 'cfg(docsrs,test)' --check-cfg 'cfg(feature, values())' -C metadata=d4fb0c790b31c5ca -C extra-filename=-cf2ab74374a2d944 --out-dir /Users/runner/work/cargo/cargo/target/tmp/cit/t1384/foo/target/custom-target/debug/deps --target /Users/runner/work/cargo/cargo/target/tmp/cit/t1384/foo/custom-target.json -L dependency=/Users/runner/work/cargo/cargo/target/tmp/cit/t1384/foo/target/custom-target/debug/deps -L dependency=/Users/runner/work/cargo/cargo/target/tmp/cit/t1384/foo/target/debug/deps`
error: requires `meta_sized` lang_item
  --> src/lib.rs:9:1
   |
9  | / pub trait Sized {
10 | |     // Empty.
11 | | }
   | |_^
```

ShoyuVanilla:
I filed a new issue addressing this CI failure #15678

tests/testsuite/compile_time_deps.rs
epage: Yes please so we see how the behavior evolves

ShoyuVanilla: Fixed it on the first commit too, though the CI will still fail due to the issue I've commented below 😂

epage (approved):

epage:
Thanks!

ShoyuVanilla:
Rebased on `master` to resolve the CI failure

epage (approved):

PR #15675, opened 2025-06-17
feat: Add JSONPath query support via `cargo metadata --query <query>`.

iddm:

The `metadata` subcommand of `cargo` is *very* useful, for many different reasons. However, it outputs too much uncompressed information, which is hard to understand at a glance without a parser. In almost all cases, `cargo metadata` is used to extract some information about specific things. These specific things require parsing the full JSON and traversing it to find it externally.
To ease the use of `cargo metadata` and to make it much more useful, this PR adds a `query` argument which expects a `JSONPath` string. This string is used to print out only the information requested out of everything available, filtering all the noise and the results which are of no interest to the user.

If in the root of this repository,
```sh
> $ cargo build
> $ cargo run -- metadata --query $.target_directory
>
> "/mnt/workspace/cargo/target"
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -339,6 +339,7 @@ dependencies = [
  "itertools 0.14.0",
  "jiff",
  "jobserver",
+ "jsonpath-rust",
  "lazycell",
  "libc",
  "libgit2-sys",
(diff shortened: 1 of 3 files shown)


iddm:
@rustbot r? @epage 

Since @weihanglo is on vacation.

epage:
Note that we generally ask for discussing problems and potential solutions in Issues, reserving PRs for when a solution is decided and for reviewing the implementation, see https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo.  This makes it easier to find past discussions, avoids splitting conversations if multiple PRs end up being made of the same problem being solved, focuses the conversation, and reduces throw away work.

Especially since this is a change that involves extending our compatibility guarantees, I'm going to mark this PR as a draft, pending problem/solution discussions in an Issue.

PR #15680, opened 2025-06-18
Document automatic override of `strip`

0x9fff00:
Related: #13638, #13677

Since #13257, Cargo may automatically set `strip = "debuginfo"`, but the documentation still states that the default is `"none"`. Although this is technically true (as pointed out in comments on the related PRs), it can be misleading. This PR retains the statements that `"none"` is the default, but adds details about the cases where it is overridden to `"debuginfo"`.

### DIFF
--- a/src/doc/src/reference/profiles.md
+++ b/src/doc/src/reference/profiles.md
@@ -119,7 +119,9 @@ strip = "debuginfo"
 ```
 
 Possible string values of `strip` are `"none"`, `"debuginfo"`, and `"symbols"`.
-The default is `"none"`.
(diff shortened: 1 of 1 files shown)


epage:
While we're not strict about this, I'm going to say that with this having been rejected twice, this should go through an Issue first per our [contributing guidelines](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo) to bring the conversation into one place, rather than continuing to fragment it.  If you decide to open an issue, I would ask to focus on the problem (how is this affecting you) and to summarize and then address past points.

PR #15684, opened 2025-06-19
implement package feature unification

mladedav:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/pull/15684)*

Implements another part of feature unification (#14774, [rfc](https://github.com/rust-lang/rfcs/blob/1c590ce05d676e72e2217845ee054758d3a6df34/text/3692-feature-unification.md)). The `workspace` option was implemented in #15157, this adds the `package` option.

The important change is changing `WorkspaceResolve` so it can contain multiple `ResolvedFeature`s. Along with that, it also needs to know which specs those features are resolved for. This was used in several other places:
- `cargo fix --edition` (from 2018 to 2021) - I think it should be ok to disallow using `cargo fix --edition` when someone already uses this feature.
- building std - it should be safe to assume std is not using this feature so I just unwrap there. I'm not sure if some attempt to later feature unification would be better.
- `cargo tree` - I just use the first feature set. This is definitely not ideal, but I'm not entirely sure what's the correct solution here. Printing multiple trees? Disallowing this, forcing users to select only one package?

Based on comments in #15157 I've added tests first with `selected` feature unification and then changed that after implementation. I'm not sure if that's how you expect the tests to be added first, if not, I can change the history.

I've expanded the test checking that this is ignored for `cargo install` although it should work the same way even if it is not ignored (`selected` and `package` are the same thing when just one package is selected).

### DIFF
--- a/src/cargo/ops/fix/mod.rs
+++ b/src/cargo/ops/fix/mod.rs
@@ -588,7 +588,15 @@ fn check_resolver_change<'gctx>(
             feature_opts,
         )?;
 
-        let diffs = v2_features.compare_legacy(&ws_resolve.resolved_features);
+        if ws_resolve.specs_and_features.len() != 1 {
+            bail!(r#"cannot fix edition when using `feature-unification = "package"`."#);
+        }
+        let resolved_features = &ws_resolve
+            .specs_and_features
+            .first()
+            .expect("We've already checked that there is exactly one.")
+            .resolved_features;
+        let diffs = v2_features.compare_legacy(resolved_features);
         Ok((ws_resolve, diffs))
     };
     let (_, without_dev_diffs) = resolve_differences(HasDevUnits::No)?;
--- a/src/cargo/ops/resolve.rs
+++ b/src/cargo/ops/resolve.rs
@@ -81,7 +81,9 @@ use crate::util::CanonicalUrl;
 use anyhow::Context as _;
 use cargo_util::paths;
 use cargo_util_schemas::core::PartialVersion;
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
+use std::rc::Rc;
 use tracing::{debug, trace};
 
 /// Filter for keep using Package ID from previous lockfile.
@@ -96,9 +98,18 @@ pub struct WorkspaceResolve<'gctx> {
     /// This may be `None` for things like `cargo install` and `-Zavoid-dev-deps`.
     /// This does not include `paths` overrides.
     pub workspace_resolve: Option<Resolve>,
-    /// The narrowed resolve, with the specific features enabled, and only the
-    /// given package specs requested.
+    /// The narrowed resolve, with the specific features enabled.
     pub targeted_resolve: Resolve,
+    /// Package specs requested for compilation along with specific features enabled. This usually
+    /// has the length of one but there may be more specs with different features when using the
+    /// `package` feature resolver.
+    pub specs_and_features: Vec<SpecsAndResolvedFeatures>,
+}
+
+/// Pair of package specs requested for compilation along with enabled features.
+pub struct SpecsAndResolvedFeatures {
+    /// Packages that are supposed to be built.
+    pub specs: Vec<PackageIdSpec>,
     /// The features activated per package.
     pub resolved_features: ResolvedFeatures,
 }
@@ -145,10 +156,20 @@ pub fn resolve_ws_with_opts<'gctx>(
     force_all_targets: ForceAllTargets,
     dry_run: bool,
 ) -> CargoResult<WorkspaceResolve<'gctx>> {
-    let specs = match ws.resolve_feature_unification() {
-        FeatureUnification::Selected => specs,
-        FeatureUnification::Workspace => &ops::Packages::All(Vec::new()).to_package_id_specs(ws)?,
+    let feature_unification = ws.resolve_feature_unification();
+    let individual_specs = match feature_unification {
+        FeatureUnification::Selected => vec![specs.to_owned()],
+        FeatureUnification::Workspace => {
+            vec![ops::Packages::All(Vec::new()).to_package_id_specs(ws)?]
+        }
     };
+    let specs: Vec<_> = individual_specs
+        .iter()
+        .map(|specs| specs.iter())
+        .flatten()
+        .cloned()
+        .collect();
+    let specs = &specs[..];
     let mut registry = ws.package_registry()?;
     let (resolve, resolved_with_overrides) = if ws.ignore_lock() {
         let add_patches = true;
@@ -229,9 +250,9 @@ pub fn resolve_ws_with_opts<'gctx>(
 
     let pkg_set = get_resolved_packages(&resolved_with_overrides, registry)?;
 
-    let member_ids = ws
-        .members_with_features(specs, cli_features)?
-        .into_iter()
+    let members_with_features = ws.members_with_features(specs, cli_features)?;
+    let member_ids = members_with_features
+        .iter()
         .map(|(p, _fts)| p.package_id())
         .collect::<Vec<_>>();
     pkg_set.download_accessible(
@@ -243,33 +264,49 @@ pub fn resolve_ws_with_opts<'gctx>(
         force_all_targets,
     )?;
 
-    let feature_opts = FeatureOpts::new(ws, has_dev_units, force_all_targets)?;
-    let resolved_features = FeatureResolver::resolve(
-        ws,
-        target_data,
-        &resolved_with_overrides,
-        &pkg_set,
-        cli_features,
-        specs,
-        requested_targets,
-        feature_opts,
-    )?;
+    let mut specs_and_features = Vec::new();
 
-    pkg_set.warn_no_lib_packages_and_artifact_libs_overlapping_deps(
-        ws,
-        &resolved_with_overrides,
-        &member_ids,
-        has_dev_units,
-        requested_targets,
-        target_data,
-        force_all_targets,
-    )?;
+    for specs in individual_specs {
+        let feature_opts = FeatureOpts::new(ws, has_dev_units, force_all_targets)?;
+
+        let narrowed_features = match feature_unification {
+            FeatureUnification::Selected | FeatureUnification::Workspace => {
+                Cow::Borrowed(cli_features)
+            }
+        };
+
+        let resolved_features = FeatureResolver::resolve(
+            ws,
+            target_data,
+            &resolved_with_overrides,
+            &pkg_set,
+            &*narrowed_features,
+            &specs,
+            requested_targets,
+            feature_opts,
+        )?;
+
+        pkg_set.warn_no_lib_packages_and_artifact_libs_overlapping_deps(
+            ws,
+            &resolved_with_overrides,
+            &member_ids,
+            has_dev_units,
+            requested_targets,
+            target_data,
+            force_all_targets,
+        )?;
+
+        specs_and_features.push(SpecsAndResolvedFeatures {
+            specs,
+            resolved_features,
+        });
+    }
 
     Ok(WorkspaceResolve {
         pkg_set,
         workspace_resolve: resolve,
         targeted_resolve: resolved_with_overrides,
-        resolved_features,
+        specs_and_features,
     })
 }
 
--- a/src/cargo/ops/resolve.rs
+++ b/src/cargo/ops/resolve.rs
@@ -162,6 +162,7 @@ pub fn resolve_ws_with_opts<'gctx>(
         FeatureUnification::Workspace => {
             vec![ops::Packages::All(Vec::new()).to_package_id_specs(ws)?]
         }
+        FeatureUnification::Package => specs.iter().map(|spec| vec![spec.clone()]).collect(),
     };
     let specs: Vec<_> = individual_specs
         .iter()
@@ -269,7 +270,28 @@ pub fn resolve_ws_with_opts<'gctx>(
     for specs in individual_specs {
         let feature_opts = FeatureOpts::new(ws, has_dev_units, force_all_targets)?;
 
+        // We want to narrow the features to the current specs so that stuff like `cargo check -p a
+        // -p b -F a/a,b/b` works and the resolver does not contain that `a` does not have feature
+        // `b` and vice-versa. However, resolver v1 needs to see even features of unselected
+        // packages turned on if it was because of working directory being inside the unselected
+        // package, because they might turn on a feature of a selected package.
         let narrowed_features = match feature_unification {
+            FeatureUnification::Package => {
+                let mut narrowed_features = cli_features.clone();
+                let enabled_features = members_with_features
+                    .iter()
+                    .filter_map(|(package, cli_features)| {
+                        specs
+                            .iter()
+                            .any(|spec| spec.matches(package.package_id()))
+                            .then_some(cli_features.features.iter())
+                    })
+                    .flatten()
+                    .cloned()
+                    .collect();
+                narrowed_features.features = Rc::new(enabled_features);
+                Cow::Owned(narrowed_features)
+            }
             FeatureUnification::Selected | FeatureUnification::Workspace => {
                 Cow::Borrowed(cli_features)
             }
--- a/src/cargo/ops/tree/mod.rs
+++ b/src/cargo/ops/tree/mod.rs
@@ -5,6 +5,7 @@ use crate::core::compiler::{CompileKind, RustcTargetData};
 use crate::core::dependency::DepKind;
 use crate::core::resolver::{features::CliFeatures, ForceAllTargets, HasDevUnits};
 use crate::core::{Package, PackageId, PackageIdSpec, PackageIdSpecQuery, Workspace};
+use crate::ops::resolve::SpecsAndResolvedFeatures;
 use crate::ops::{self, Packages};
 use crate::util::CargoResult;
 use crate::{drop_print, drop_println};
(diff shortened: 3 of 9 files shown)


src/doc/src/reference/unstable.md
```diff
@@ -1860,7 +1860,7 @@ Specify which packages participate in [feature unification](../reference/feature
 * `selected`: Merge dependency features from all packages specified for the current build.
 * `workspace`: Merge dependency features across all workspace members,
   regardless of which packages are specified for the current build.
-* `package` _(unimplemented)_: Dependency features are considered on a package-by-package basis,
+* `package`: Dependency features are considered on a package-by-package basis,
```
epage, line 1902: This PR implements it as `projects` instead of `package`.  What was the motivation for that?

tests/testsuite/feature_unification.rs
epage, line 1: Would be good to have a `cargo tree` test as well

epage: and `cargo fix`

src/cargo/ops/fix/mod.rs
```diff
@@ -588,7 +588,15 @@ fn check_resolver_change<'gctx>(
             feature_opts,
         )?;
 
-        let diffs = v2_features.compare_legacy(&ws_resolve.resolved_features);
+        if ws_resolve.specs_and_features.len() != 1 {
+            bail!(r#"Cannot fix edition when using `feature-unification = "projects"`."#);
```
epage: Cargo error messages should start with a lower case letter

epage: > cargo fix - from what I understand, the location is reached only with --edition. I think it should be ok to disallow using cargo fix --edition when someone already uses this feature.

This should just be limited to projects with `edition = "2018"` which should be fine

src/cargo/ops/resolve.rs
```diff
@@ -243,33 +261,43 @@ pub fn resolve_ws_with_opts<'gctx>(
...
-    pkg_set.warn_no_lib_packages_and_artifact_libs_overlapping_deps(
-        ws,
-        &resolved_with_overrides,
-        &member_ids,
-        has_dev_units,
-        requested_targets,
-        target_data,
-        force_all_targets,
-    )?;
+    for specs in individual_specs {
```
epage, line 270: So we're creating independent resolves for each package.  I take it we can then leave merging of these graphs to a future possibility to further optimize this?

src/cargo/ops/tree/mod.rs
```diff
@@ -179,10 +179,17 @@ pub fn build_and_print(ws: &Workspace<'_>, opts: &TreeOptions) -> CargoResult<()
         .map(|pkg| (pkg.package_id(), pkg))
         .collect();
 
+    // TODO build multiple graphs, one for each resolved features set.
+    let resolved_features = &ws_resolve
```
epage: > cargo tree - I just use the first feature set. This is definitely not ideal, but I'm not entirely sure what's the correct solution here. Printing multiple trees? Disallowing this, forcing users to select only one package?

I see this as no different than `cargo tree --workspace` where there will be multiple root packages.  `--edges features` (ro whatever its called) will be needed to fully understand it.

src/doc/src/reference/unstable.md
mladedav: It was a mistake but where I got the wrong name, I really can't say. Fixed everywhere, hopefully.

epage:

epage:
Feel free to edit the commits as needed

tests/testsuite/feature_unification.rs
mladedav: I've added a `cargo tree` test, I'll just have to fixup history so that it's added before the feature.

That said, I think it might need a bit of design work because using `package` and `selected` feature unification has almost identical outuput, but in the former each tree should be taken on its own while in the latter the user should be aware that the features are unified.

For example in one simplified case where a depends on `common/a` and b depends on `common/b`; `package` outputs:
```
common v0.1.0 ([ROOT]/foo/common)
a v0.1.0 ([ROOT]/foo/a)
└── common feature "a"
    └── common v0.1.0 ([ROOT]/foo/common)
b v0.1.0 ([ROOT]/foo/b)
└── common feature "b"
    └── common v0.1.0 ([ROOT]/foo/common)
```

while `selected` currently outputs:
```
a v0.1.0 ([ROOT]/foo/a)
└── common feature "a"
    └── common v0.1.0 ([ROOT]/foo/common)

b v0.1.0 ([ROOT]/foo/b)
└── common feature "b"
    └── common v0.1.0 ([ROOT]/foo/common)

common v0.1.0 ([ROOT]/foo/common)
```

and for completeness, `workspace` outputs:
```
a v0.1.0 ([ROOT]/foo/a)
└── common feature "a"
    └── common v0.1.0 ([ROOT]/foo/common)

b v0.1.0 ([ROOT]/foo/b)
└── common feature "b"
    └── common v0.1.0 ([ROOT]/foo/common)

common v0.1.0 ([ROOT]/foo/common)
```

I'm not sure what would be the ideal behavior here though, we probably don't want to have both features coming out of both packages in `selected` because people often use this to find out what package turned on a given feature. But I don't think we should keep it as is either. Maybe adding an edge that signifies a feature but somehow marks it as coming from unification elsewhere?

For example something like this?
```
a v0.1.0 ([ROOT]/foo/a)
└── common feature "a"
    └── common v0.1.0 ([ROOT]/foo/common)
└── common feature "b" (unified)
    └── common v0.1.0 ([ROOT]/foo/common)

b v0.1.0 ([ROOT]/foo/b)
└── common feature "b"
    └── common v0.1.0 ([ROOT]/foo/common)
└── common feature "a" (unified)
    └── common v0.1.0 ([ROOT]/foo/common)

common v0.1.0 ([ROOT]/foo/common)
```

---

Should there be some kind of proposal for these changes? And in that case should I just fixup history with the current behavior or wait for the outcome of that? Or is it fine to make the changes as part of this PR or a different one but without any larger proposal than just the PR?

epage: > Should there be some kind of proposal for these changes? And in that case should I just fixup history with the current behavior or wait for the outcome of that? Or is it fine to make the changes as part of this PR or a different one but without any larger proposal than just the PR?

We can clean this up and get it merged and leave it as an Open Question on the tracking issue to have the discussion run independent of this PR

tests/testsuite/feature_unification.rs
```diff
@@ -106,6 +107,442 @@ fn workspace_feature_unification() {
...
+    └── outside v0.1.0
+b v0.1.0 ([ROOT]/foo/b)
+├── common feature "b"
+│   └── common v0.1.0 ([ROOT]/foo/common)
+├── common feature "default"
+│   └── common v0.1.0 ([ROOT]/foo/common)
+├── outside feature "b"
+│   └── outside v0.1.0
+└── outside feature "default"
+    └── outside v0.1.0 ([ROOT]/foo/common)
```
mladedav: This is wrong, the path is incorrect. I will get back to this.

src/cargo/ops/fix/mod.rs
mladedav: Fixed

tests/testsuite/feature_unification.rs
mladedav: Ok, tree tests are added now as well as one fix test. I did not know what to test for though so I just took a random fix test from edition 2021 to 2024 since the relevant change should break just a specific upgrade from 2018. If you want to test for something different, let me know.

src/cargo/ops/resolve.rs
mladedav: Just independent `ResolvedFeatures`, it seems everything works with just a single `targeted_resolve`, though I did not expect that since the field is documeted with "The narrowed resolve, with the specific features enabled." Then from those independent unit graphs are constructed and those are then merged into a single build context.

I probably don't understand the whole process well enough, but I don't see a way to merge anything earlier?

src/cargo/ops/tree/mod.rs
mladedav: True, I got confused, this part is fine. I think that just signalling to the user whether features are unified between the trees or not is a new problem.

tests/testsuite/feature_unification.rs
mladedav: Seems I've just committed a wip broken version, even the CI seemed to reject this with the code in the commit it ran on. Oh well, this was just noise.

mladedav:
I've also just added a test with weak dependencies, which works. I tried to also add namespaced `dep:` features, but those seem to not be unified (#15694) so I'm ignoring that, but I guess that might also go towards the feature unification's unresolved questions.

tests/testsuite/feature_unification.rs
epage: Added a note to the tracking issue

We should have a test specifically for the broken `cargo fix --edition` case

src/cargo/ops/resolve.rs
epage: Maybe phrased differently, if you have the following
```
a -> common w/ features=foo
b -> common w/ features=foo
c -> common
```
Will we have 2 or 3 entries in the unit graph?

mladedav: It should have 2 entries for `common` (so 5 entries in total).

I've added a test where there are the same features enabled on a common dependency and it is compiled just once.

tests/testsuite/feature_unification.rs
mladedav: I wasn't able to trigger it. Looking at it now, I think that the branch is hit only if migrating a package without a workspace so in that case there will never be multiple packages to have features unified. But I'm not entirely sure.

src/cargo/ops/resolve.rs
epage: Thanks!

mladedav:
I had to also narrow down `CliFeature`s passed to the feature resolver because otherwise `cargo check -p a -p b -F a,b` complains that `a` does not have feature named `b`. The same code already ran with all selected packages and features so if there were unknown features, it would fail.

However, simply narrowing it down does not work for resolver v1 because there are tests like `other_member_from_current` which check edge cases regarding a feature of an unselected package inside cwd enabling a feature of a selected package. Since resolver v1 will not be using package feature unification, I just pass all cli features as before and only narrow it down for package feature unification.

I've added a test for this but added it just to the commit that's supposed to fix the tests. I think I've done three different things regarding splitting tests into commits by now since this is adding a feature so behavior of tests before and after the addition may be either behavior with `feature-unification = "selected"` or it could be that it just didn't compile and now it does so if you want me to fixup the test commits, feel free to tell me what's optimal.

epage:
Per usual, would prefer tests and test cases to be added in that initial commit.  Whatever it does (nothing, change behavior, start passing, start erroring), seeing that change of state through the diffs is very helpful for reviewing and demonstrating what the behavior is.

tests/testsuite/feature_unification.rs
epage: The following test hits it.  To get this to work, I took this from `tests/testsuite/fix.rs` and had to
- add a workspace member
- ensure the workspace manifest was "real" (had a package) (or else the "edition" is not controlling the resolver so it doesn't fire)
- used `--workspace`
```rust
#[cargo_test]
fn edition_v2_resolver_report() {
    // Show a report if the V2 resolver shows differences.
    Package::new("common", "1.0.0")
        .feature("f1", &[])
        .feature("dev-feat", &[])
        .add_dep(Dependency::new("opt_dep", "1.0").optional(true))
        .publish();
    Package::new("opt_dep", "1.0.0").publish();

    Package::new("bar", "1.0.0")
        .add_dep(
            Dependency::new("common", "1.0")
                .target("cfg(whatever)")
                .enable_features(&["f1"]),
        )
        .publish();

    let p = project()
        .file(
            "Cargo.toml",
            r#"
                [workspace]
                members = ["bar"]

                [package]
                name = "foo"
                version = "0.1.0"
                edition = "2018"

                [dependencies]
                common = "1.0"
                bar = "1.0"

                [build-dependencies]
                common = { version = "1.0", features = ["opt_dep"] }

                [dev-dependencies]
                common = { version="1.0", features=["dev-feat"] }
            "#,
        )
        .file("src/lib.rs", "")
        .file(
            "bar/Cargo.toml",
            r#"
                [package]
                name = "bar"
                version = "0.1.0"
                edition = "2018"
            "#,
        )
        .file("bar/src/lib.rs", "")
        .build();

    p.cargo("fix --edition --allow-no-vcs --workspace")
        .env("CARGO_RESOLVER_FEATURE_UNIFICATION", "package")
        .arg("-Zfeature-unification")
        .masquerade_as_nightly_cargo(&["feature-unification"])
        .with_status(101)
        .with_stderr_data(str![[r#"
[MIGRATING] Cargo.toml from 2018 edition to 2021
[DOWNLOADING] crates ...
[DOWNLOADED] common v1.0.0 (registry `dummy-registry`)
[DOWNLOADED] bar v1.0.0 (registry `dummy-registry`)
[DOWNLOADED] opt_dep v1.0.0 (registry `dummy-registry`)
[MIGRATING] bar/Cargo.toml from 2018 edition to 2021
[ERROR] cannot fix edition when using `feature-unification = "package"`.

"#]].unordered())
        .run();
}

```

mladedav: Thanks, I've added it now.

mladedav:
Ok, I've added all the tests with `feature-unification = "selected"` and change them later to `package`.

src/cargo/ops/resolve.rs
```diff
@@ -269,7 +270,28 @@ pub fn resolve_ws_with_opts<'gctx>(
...
+                            .iter()
+                            .any(|spec| spec.matches(package.package_id()))
+                            .then_some(cli_features.features.iter())
+                    })
+                    .flatten()
+                    .cloned()
+                    .collect();
+                narrowed_features.features = Rc::new(enabled_features);
+                Cow::Owned(narrowed_features)
+            }
```
epage: From the `--features` [docs](https://doc.rust-lang.org/cargo/commands/cargo-check.html)

> Space or comma separated list of features to activate. Features of workspace members may be enabled with package-name/feature-name syntax. This flag may be specified multiple times, which enables all specified features.

The way this is worded, it makes it sound like you can specify a feature of a dependency so long as its a workspace member.  Whether its intended or not, you can even specify features for transitive dependencies so long as all of the activated deps are already in your `Cargo.lock`

@ehuss wanted to double check my reading of those docs and if you had any additional ideas about the problem of `CARGO_RESOLVER_FEATURE_UNIFICATION=package cargo check -p a -F a/a -p b -F b/b` trying to activate the `b/b` feature when `b` isn't in the resolve.

epage:
Just need to double check on one question (https://github.com/rust-lang/cargo/pull/15684#discussion_r2164744359) but otherwise, this looks good.  Thanks for your work on this and patience through the review feedback!

src/cargo/ops/resolve.rs
mladedav: It does work that way, I've added a test case just now to the cli features tests having `cargo check -p a -F a/a,common/b` because `members_with_features` already contains the `a` feature and also a `DepFeature` for `common/b`, both for package `a`. If that's what you mean.

Arguably that reading could even allow `cargo check -p a -F b/b` even if `b` is not a dependency of `a` (the quoted docs don't mention dependencies at all) but that's rejected regardless of feature unification settings.

epage: So if I'm understanding correctly, the narrowed features are preserving that existing behavior

Didn't help that I read through this multiple times before it clicked what you are doing with this code block.  You had already divided up features by workspace member, so now you are getting the ones only relevant for the selected workspace members to apply to resolution.

What would this then do with `CARGO_RESOLVER_FEATURE_UNIFICATION=package cargo check -p a -F transitive/feature`?  I'm assuming it silently ignores `transitive/feature` as it wasn't in `members_with_features` and so doesn't get forwarded on to the feature resolver.

That case isn't officially supported but it would at least be good to call out if the behavior deviates from regular cargo behavior in the tracking issue.  We can then decide how we want to handle that case more generally.

mladedav: Do you mean something like this? It fails for both `package` and `selected` so I assume not, but then I don't understand what case do you mean.

<details><summary>Test case</summary>


```rust
#[cargo_test]
fn transitive_feature() {
    Package::new("child", "1.0.0")
        .add_dep(&Dependency::new("grand_child", "1.0"))
        .publish();
    Package::new("grand_child", "1.0.0")
        .feature("a", &[])
        .file(
            "lib.rs",
            r#"
            #[cfg(not(feature = "a"))]
            compile_error!("missing feature");
            "#,
        )
        .publish();

    let p = project()
        .file(
            "Cargo.toml",
            r#"
                [package]
                name = "foo"
                version = "0.1.0"
                edition = "2024"

                [dependencies]
                chile = "1.0"
            "#,
        )
        .file("src/lib.rs", "")
        .build();

    p.cargo("check -F grand_child/a")
        .arg("-Zfeature-unification")
        .masquerade_as_nightly_cargo(&["feature-unification"])
        .env("CARGO_RESOLVER_FEATURE_UNIFICATION", "selected")
        .with_status(101)
        .with_stderr_data(str![[r#"
[ERROR] the package 'foo' does not contain this feature: grand_child/a

"#]])
        .run();

    p.cargo("check -F grand_child/a")
        .arg("-Zfeature-unification")
        .masquerade_as_nightly_cargo(&["feature-unification"])
        .env("CARGO_RESOLVER_FEATURE_UNIFICATION", "package")
        .with_status(101)
        .with_stderr_data(str![[r#"
[ERROR] the package 'foo' does not contain this feature: grand_child/a

"#]])
        .run();
}
```

</details>

epage: Looks like I over-estimated what is supported.  The case I'm talking about is limited to direct dependencies:
```rust
#[cargo_test]
fn non_member_features() {
    Package::new("child", "1.0.0")
        .add_dep(&Dependency::new("grand_child", "1.0"))
        .feature("a", &[])
        .file(
            "src/lib.rs",
            r#"
            #[cfg(feature = "a")]
            compile_error!("`a` is active");
            "#,
        )
        .publish();
    Package::new("grand_child", "1.0.0")
        .feature("a", &[])
        .file(
            "src/lib.rs",
            r#"
            #[cfg(feature = "a")]
            compile_error!("`a` is active");
            "#,
        )
        .publish();

    let p = project()
        .file(
            "Cargo.toml",
            r#"
                [package]
                name = "foo"
                version = "0.1.0"
                edition = "2024"

                [dependencies]
                child = "1.0"
            "#,
        )
        .file("src/lib.rs", "")
        .build();

    p.cargo("check -F child/a")
        .with_status(101)
        .with_stderr_data(str![[r#"
[LOCKING] 2 packages to latest Rust 1.87.0 compatible versions
[DOWNLOADING] crates ...
[DOWNLOADED] grand_child v1.0.0 (registry `dummy-registry`)
[DOWNLOADED] child v1.0.0 (registry `dummy-registry`)
[CHECKING] grand_child v1.0.0
[CHECKING] child v1.0.0
[ERROR] `a` is active
 --> [ROOT]/home/.cargo/registry/src/-[HASH]/child-1.0.0/src/lib.rs:3:13
  |
3 |             compile_error!("`a` is active");
  |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[ERROR] could not compile `child` (lib) due to 1 previous error

"#]])
        .run();

    p.cargo("check -F grand_child/a")
        .with_status(101)
        .with_stderr_data(str![[r#"
[ERROR] the package 'foo' does not contain this feature: grand_child/a

"#]])
        .run();
}
```

mladedav: I've added a test at the end trying to cover as much of this as I could with all three cases of feature unification. It seems to work the same way regardless of the option, i.e. you can set features on direct dependencies but not on grand-dependencies.

This has a consequence that `Cargo.toml`s are more expressive than command line, e.g. you can set default features so that `a` builds with `common/a` and `b` builds with `common/b`. If there are no default features (or features on `a` and `b` turning on the respective features), one cannot just write something like `cargo check -p a -p b -F a/common/a -F b/common/b` to turn the featur on only for the given root package. I would say this is correct behavior, just something to point out, because until now I believe one could theoretically just not define any features and then work around that on the command line, which is no longer the case with `feature-unification = package`. I'm not sure if what I wrote is comprehensible, if not, I'll try to crate full examples for illustration.

mladedav:
> Thanks for your work on this and patience through the review feedback!

Honestly, this has been a blast. I wouldn't say I needed pretty much any patience with how fast you were reacting to each iteration. Thanks for the time and reviews!

tests/testsuite/feature_unification.rs
```diff
@@ -203,3 +1005,362 @@ fn unstable_config_on_stable() {
...
+        .file(
+            "Cargo.toml",
+            r#"
+                [workspace]
+                resolver = "2"
+                members = ["parent", "child", "grandchild"]
+            "#,
+        )
+        .file(
+            "grandchild/Cargo.toml",
```
epage: Whats the intention of this test?

mladedav: I tried to cover what you were talking about regarding turning on features of non-selected dependency crates inside a workspace. I thought it would be different in some cases based on `feature-unification` but it's consistent.

Do you want me to remove this?

epage: There are three cases for `foo` in `foo/bar`
1. A workspace member
2. A direct dep of a workspace member
3. An indirect dep of a workspace member

In https://github.com/rust-lang/cargo/pull/15684#discussion_r2164935746, I was talking about (2) while this test covers (1).

mladedav: Right, I tried to cover the first two cases in the original `package_feature_unification_cli_features`.

Here I tried specifically to get close to 3 but since that does not work in the general case, I tried to get closer to it by the indirect dependency being part of the workspace, but either not being selected or being excluded.

I was unsure about how some of the scenarios would (or wouldn't) work but arguably since all of the cases work the same way regardless of `feature-unification`, it might not belong here as it's more of a general features test. Or it might as it should be the same for all of them.

So I can either leave this as is, move it somewhere or just remove it completely.

And if you want the other cases covered more than they are in the other test, feel free to tell me.

epage: > Right, I tried to cover the first two cases in the original package_feature_unification_cli_features.

Ok, that looks to be the case `"check -p a -F a/a,outside/b"`.

> So I can either leave this as is, move it somewhere or just remove it completely.

Unsure but we should at least make the name more clear that this is focused on activating features for specific packages only within a workspace, contracted with `package_feature_unification_cli_features` which focused on a combination of workspace members and registry dependencies while having a similar name of `package_feature_unification_grandchild_cli_features`.

mladedav: I've renamed the second one to `feature_unification_of_cli_features_within_workspace`. I didn't come up with anything more descriptive yet reasonable than `package_feature_unification_cli_features` for the first one so I've left that as is.

epage:
@mladedav looks like there is a conflict?

mladedav:
Right, I've rebased this on top of master now

epage (approved):

epage:
Thanks!

PR #15686, opened 2025-06-20
Make `lldb_works_after_trimmed` test robust against lldb output

joshtriplett:
Some versions of lldb (including the version in CI) print "Process $PID
resuming" after `continue`, breaking the test. Add another wildcard to
allow for that, without breaking the test on versions of lldb that
*don't* print that message.

Fixes a CI test failure.

This PR aims to resolve a CI test failure, so it should pass CI.


weihanglo (commented):
If this test blocks other PR, maybe we can disable it at this moment. I can try fixing it when I am back.

joshtriplett:
@weihanglo Looking at the test failure on macOS, I think more is wrong with this test than this PR would fix, so yeah, I think disabling it is the right answer. I'll send a PR doing that.

joshtriplett:
@weihanglo https://github.com/rust-lang/cargo/pull/15687

PR #15689, opened 2025-06-21
Configure snippets renderer to match rustc/Cargo style

kornelski:
`Renderer::styled()` defaults to green labels for `note` and `help`, but Cargo and `rustc` use cyan.

### DIFF
--- a/src/cargo/core/shell.rs
+++ b/src/cargo/core/shell.rs
@@ -415,7 +415,11 @@ impl Shell {
         writeln!(
             self.err(),
             "{}",
-            Renderer::styled().term_width(term_width).render(message)
+            Renderer::styled()
+                .note(NOTE)
+                .help(HELP)
+                .term_width(term_width)
+                .render(message)
         )
     }
 }
--- a/src/cargo/util/style.rs
+++ b/src/cargo/util/style.rs
@@ -8,6 +8,7 @@ pub const PLACEHOLDER: Style = AnsiColor::Cyan.on_default();
 pub const ERROR: Style = AnsiColor::Red.on_default().effects(Effects::BOLD);
 pub const WARN: Style = AnsiColor::Yellow.on_default().effects(Effects::BOLD);
 pub const NOTE: Style = AnsiColor::Cyan.on_default().effects(Effects::BOLD);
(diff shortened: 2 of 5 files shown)


src/cargo/core/shell.rs
```diff
@@ -415,7 +415,11 @@ impl Shell {
         writeln!(
             self.err(),
             "{}",
-            Renderer::styled().term_width(term_width).render(message)
+            Renderer::styled()
+                .note(NOTE)
```
epage, line 419: this should be a pr to annotate snippets if there is a discrepancy

epage (changes requested):

src/cargo/core/shell.rs
kornelski: I don't understand.

`shell.render_message(Level::Note.title("foo"))` uses a wrong color. The snippets don't specify a color themselves by design, the color is imposed on the snippets by the stylesheet initialized by `styled()`.

kornelski: Oh, I see. I thought the CI test failure was a false positive:

  
  ```
       Summary semver requires new major version: 1 major and 0 minor checks failed
  --- failure constructible_struct_adds_field: externally-constructible struct adds field ---
  
  Description:
  A pub struct constructible with a struct literal has a new pub field. Existing struct literals must be updated to include the new field.
          ref: https://doc.rust-lang.org/reference/expressions/struct-expr.html
         impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.41.0/src/lints/constructible_struct_adds_field.ron
```

but there's also a UI test failing.

epage: The expectation is that `annotate-snippets` defaults *are* rustc's defaults.  If `annotate-snippets` is not matching rustc, then that is a bug with `annotate-snippers` that should be resolved.  We should not work around that bug in Cargo.

epage: Also, i would recommend being pretty sure about a resolution when marking things resolved, otherwise leave it to the original person.  While this is a smaller PR and so its less likely to be lost, in larger PRs that can easily make it so misunderstood feedback gets lost.

PR #15692, opened 2025-06-22
Rework `cargo-test-support` & `testsuite` to use `CARGO_BIN_EXE_*` for Cargo

ranger-ross:

This PR reworks `cargo-test-support` and `testsuite` to use Snapbox's [`cargo_bin!()`](https://docs.rs/snapbox/latest/snapbox/cmd/macro.cargo_bin.html) instead of [`cargo_bin()`](https://docs.rs/snapbox/latest/snapbox/cmd/fn.cargo_bin.html) which makes assumptions about the structure of Cargo's build directory. 
`cargo_bin!()` uses `CARGO_BIN_EXE_*` for locating the `cargo` binary which should be more resilient to directory/layout changes.

Linking a relevant Zulip discussion [here](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/cargo_bin_exe.20and.20tests/with/513638220)[#t-cargo > cargo_bin_exe and tests](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/cargo_bin_exe.20and.20tests/with/513638220)

### Implementation Notes 

`cargo_bin!()` only works when being called from the `testsuite` as it's only set when executing integration tests and `cargo-test-support` is a regular crate. 
To make this change, I introduced an extension trait `CargoProjectExt` in `testsuite` for running `.cargo()` and implemented it on `Project`. 

In `cargo-test-support` other functionality relies on `.cargo()` so these also needed to be moved to `testsuite`
* [`src/tools.rs`](https://github.com/rust-lang/cargo/blob/master/crates/cargo-test-support/src/tools.rs) 
* Parts [`src/cross_compile`](https://github.com/rust-lang/cargo/blob/master/crates/cargo-test-support/src/cross_compile.rs)
  * I had to split this up unfortunately, as `disabled()` requires running Cargo to check if we should disable cross compile tests.
  * Other fns in `cross_compile` are used in `cargo-test-support` so moving everything to `testsuite` would have ended up requiring moving many things to test suite.

I'd definitely recommend reviewing commit by commit.
There are a lot of diffs due to the nature of reorganizing things. 
I did my best to split things things into smaller PRs but they still contain a lot of `use` statement diffs.

r? @epage


crates/cargo-test-support/src/registry.rs
```diff
@@ -2,7 +2,7 @@
 //!
 //! # Example
 //!
-//! ```no_run
+//! ```ignore
```
epage: Would prefer for these not to be `ignore`

tests/testsuite/bad_config.rs
```diff
@@ -6,6 +6,8 @@ use cargo_test_support::registry::{self, Package};
 use cargo_test_support::str;
 use cargo_test_support::{basic_manifest, project, rustc_host};
 
+use crate::utils::ext::CargoProjectExt;
```
epage: Should we put this in a prelude and maybe re-export `cargo_test_support::prelude::*` in it?

epage: That could even be done as a refactor commit before any of the other changes

ranger-ross: sure, that was something I was considering but didn't end following through with.

I think its a good idea.

epage: Also, we should probably check how https://doc.crates.io/contrib/tests/writing.html needs to be updated

crates/cargo-test-support/src/registry.rs
ranger-ross: The issue is that these example use `.cargo()` which is no longer callable in `cargo-test-support`. So the example no longer compiles

Perhaps, we can keep it `no_run` if we comment (or remove)  the `.cargo()` calls from the examples.


Maybe something like this for `publish.rs`

```rs
//! let registry = RegistryBuilder::new().http_api().http_index().build();
//!
//! let p = project()
//!     .file(...)
//!     .file("src/main.rs", "fn main() {}")
//!     .build();
//!
//!  // p.cargo("publish --no-verify")
//!  //   .replace_crates_io(registry.index_url())
//!  //   .run();
//!
//! validate_upload(
//!     r#"...."#,
//!     "foo-0.0.1.crate",
//!     &["Cargo.lock", "Cargo.toml", "Cargo.toml.orig", "src/main.rs"],
//! );
```

epage: Either that or we could ust do:
```rust
//! validate_upload(
//!     r#"...."#,
//!     "foo-0.0.1.crate",
//!     &["Cargo.lock", "Cargo.toml", "Cargo.toml.orig", "src/main.rs"],
//! );
```

tests/testsuite/bench.rs
```diff
@@ -1,8 +1,9 @@
 //! Tests for the `cargo bench` command.
 
-use cargo_test_support::prelude::*;
+use crate::prelude::*;
 use cargo_test_support::{basic_bin_manifest, basic_lib_manifest, basic_manifest, project, str};
 
+
```
epage: whats this blank line that got added to a lot of these files?

tests/testsuite/main.rs
```diff
@@ -193,7 +193,12 @@ mod weak_dep_features;
 mod workspaces;
 mod yank;
 
-use cargo_test_support::prelude::*;
+use crate::prelude::*;
+
+pub mod prelude {
+    pub use crate::utils::ext::CargoProjectExt;
```
epage, line 200: Can you split out the creation of `CargoProjectExt` from the prelude, as in create the prelude in one commit and then `CargoProjectExt` after it?

This touches a lot of files and so being minimal with what is changed can make it a lot easier to review or come back later and see what was done

See also https://github.com/rust-lang/cargo/pull/15692#discussion_r2161860056

tests/testsuite/alt_registry.rs
```diff
@@ -2,14 +2,13 @@
...
-use cargo_test_support::compare::assert_e2e;
 use crate::prelude::*;
+use cargo_test_support::compare::assert_e2e;
 use cargo_test_support::publish::validate_alt_upload;
 use cargo_test_support::registry::{self, Package, RegistryBuilder};
 use cargo_test_support::str;
 use cargo_test_support::{basic_manifest, paths, project};
 
-
 #[cargo_test]
```
epage: This commit seems to have some unrelated changes in it

tests/testsuite/lints/error/mod.rs
```diff
@@ -1,4 +1,5 @@
 use crate::prelude::*;
+use crate::utils::ext::CargoCommandExt;
```
epage: Why is the `use` added when its in the `prelude`?

ranger-ross:
Okay sorry, I cleaned up the commits. (Not quiet sure what happened)

I split out the commits as you suggested in [this comment](https://github.com/rust-lang/cargo/pull/15692#discussion_r2164495253) and updated the testing documentation as well

PR #15693, opened 2025-06-22
test(msrv): use a far future "newer" version

cuviper:
Two tests were using `.rust_version("1.90.0")` for a "newer" Rust, but
this failed in rust-lang/rust#142792 for the start of the *real* 1.90.0.
Let's bump that to 1.999.0 so we won't have to deal with it for a while.

### DIFF
--- a/tests/testsuite/rust_version.rs
+++ b/tests/testsuite/rust_version.rs
@@ -518,7 +518,7 @@ higher v0.0.1 ([ROOT]/foo)
 #[cargo_test]
 fn resolve_edition2024() {
     Package::new("only-newer", "1.6.0")
-        .rust_version("1.90.0")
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks! I'm not 100% positive why it was done the way that it was, but this seems like it should fix the issue. I'll post a submodule update.

ehuss:
Hm, I would like to also get https://github.com/rust-lang/cargo/pull/15696 merged before the beta branch (otherwise we would need to immediately beta backport it). Would it be ok to wait until tomorrow (or whenever someone gets back online)?

cuviper:
Sure, I think it's ok if the 1.89/1.90 branching slips a little bit.

PR #15696, opened 2025-06-22
Bump cargo-util-schemas version

ehuss:
https://github.com/rust-lang/cargo/pull/15643 introduced a semver breaking change by adding a new pub field to `TomlProfile`.


weihanglo (approved):
Thanks!

PR #15704, opened 2025-06-25
feat: Implementation and tests for `multiple-build-scripts`

namanlp:
Hi Everyone!

This is PR for the implementation of the first milestone of [GSoC Project : Build Script Delegation](https://summerofcode.withgoogle.com/programs/2025/projects/nUt4PdAA)

This will provide actual implementation for #15630

Now, multiple build scripts are parsed, this PR aims to implement the functioning the feature. This PR will allow users to use multiple build scripts, and is backward compatible with single script as well as boolean values. 

**Motivation :** This will help users to maintain separate smaller and cleaner build scripts instead of one large build script. This is also necessary for build script delegation.

Deferred
- Accessing each build script's `OUT_DIR`:  This will be handled in a follow up PR.  For now, each build script writes to its own `OUT_DIR` and `OUT_DIR` for the regular build targets is set to the build script with the **lexicographically largest** name..
- User control over which build script wins in a conflict.  This will be handled in a follow up PR.  If two build scripts write to the same env variable, which gets applied to the binary?  Currently, its the build script with the **lexicographically largest** name.  This makes it deterministic.  With some futzing, users can control this for now.  However, with build script delegation, users won't be able to control this.  We likely want it based off of the order the user assigns into the build script array.
- Something about linking a C library is actually preferring **lexicographically smallest** name.  We should handle conflicts consistently.  We need to dig into what parts are doing it based on smallest and make sure that whatever priority scheme we use for env variables applies here as well.

```toml
```

### DIFF
--- a/src/cargo/core/compiler/build_runner/mod.rs
+++ b/src/cargo/core/compiler/build_runner/mod.rs
@@ -248,23 +248,25 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
                 args.extend(compiler::features_args(unit));
                 args.extend(compiler::check_cfg_args(unit));
 
-                let script_meta = self.find_build_script_metadata(unit);
-                if let Some(meta) = script_meta {
-                    if let Some(output) = self.build_script_outputs.lock().unwrap().get(meta) {
-                        for cfg in &output.cfgs {
-                            args.push("--cfg".into());
-                            args.push(cfg.into());
-                        }
+                let script_metas = self.find_build_script_metadatas(unit);
+                if let Some(meta_vec) = script_metas.clone() {
+                    for meta in meta_vec {
+                        if let Some(output) = self.build_script_outputs.lock().unwrap().get(meta) {
+                            for cfg in &output.cfgs {
+                                args.push("--cfg".into());
+                                args.push(cfg.into());
+                            }
 
-                        for check_cfg in &output.check_cfgs {
-                            args.push("--check-cfg".into());
-                            args.push(check_cfg.into());
-                        }
+                            for check_cfg in &output.check_cfgs {
+                                args.push("--check-cfg".into());
+                                args.push(check_cfg.into());
+                            }
 
-                        for (lt, arg) in &output.linker_args {
-                            if lt.applies_to(&unit.target, unit.mode) {
-                                args.push("-C".into());
-                                args.push(format!("link-arg={}", arg).into());
+                            for (lt, arg) in &output.linker_args {
+                                if lt.applies_to(&unit.target, unit.mode) {
+                                    args.push("-C".into());
+                                    args.push(format!("link-arg={}", arg).into());
+                                }
                             }
                         }
                     }
@@ -285,7 +287,7 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
                     args,
                     unstable_opts,
                     linker: self.compilation.target_linker(unit.kind).clone(),
-                    script_meta,
+                    script_metas,
                     env: artifact::get_env(&self, self.unit_deps(unit))?,
                 });
             }
@@ -420,29 +422,40 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
         &self.bcx.unit_graph[unit]
     }
 
-    /// Returns the `RunCustomBuild` Unit associated with the given Unit.
+    /// Returns the `RunCustomBuild` Units associated with the given Unit.
     ///
     /// If the package does not have a build script, this returns None.
-    pub fn find_build_script_unit(&self, unit: &Unit) -> Option<Unit> {
+    pub fn find_build_script_units(&self, unit: &Unit) -> Option<Vec<Unit>> {
         if unit.mode.is_run_custom_build() {
-            return Some(unit.clone());
+            return Some(vec![unit.clone()]);
         }
-        self.bcx.unit_graph[unit]
+
+        let build_script_units: Vec<Unit> = self.bcx.unit_graph[unit]
             .iter()
-            .find(|unit_dep| {
+            .filter(|unit_dep| {
                 unit_dep.unit.mode.is_run_custom_build()
                     && unit_dep.unit.pkg.package_id() == unit.pkg.package_id()
             })
             .map(|unit_dep| unit_dep.unit.clone())
+            .collect();
+        if build_script_units.is_empty() {
+            None
+        } else {
+            Some(build_script_units)
+        }
     }
 
     /// Returns the metadata hash for the `RunCustomBuild` Unit associated with
     /// the given unit.
     ///
     /// If the package does not have a build script, this returns None.
-    pub fn find_build_script_metadata(&self, unit: &Unit) -> Option<UnitHash> {
-        let script_unit = self.find_build_script_unit(unit)?;
-        Some(self.get_run_build_script_metadata(&script_unit))
+    pub fn find_build_script_metadatas(&self, unit: &Unit) -> Option<Vec<UnitHash>> {
+        self.find_build_script_units(unit).map(|units| {
+            units
+                .iter()
+                .map(|u| self.get_run_build_script_metadata(u))
+                .collect()
+        })
     }
 
     /// Returns the metadata hash for a `RunCustomBuild` unit.
@@ -480,11 +493,11 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
     /// Returns a [`UnitOutput`] which represents some information about the
     /// output of a unit.
     pub fn unit_output(&self, unit: &Unit, path: &Path) -> UnitOutput {
-        let script_meta = self.find_build_script_metadata(unit);
+        let script_metas = self.find_build_script_metadatas(unit);
         UnitOutput {
             unit: unit.clone(),
             path: path.to_path_buf(),
-            script_meta,
+            script_metas,
         }
     }
 
--- a/src/cargo/core/compiler/compilation.rs
+++ b/src/cargo/core/compiler/compilation.rs
@@ -45,7 +45,7 @@ pub struct Doctest {
     /// The script metadata, if this unit's package has a build script.
     ///
     /// This is used for indexing [`Compilation::extra_env`].
-    pub script_meta: Option<UnitHash>,
+    pub script_metas: Option<Vec<UnitHash>>,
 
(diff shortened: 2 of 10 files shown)


epage:
> This is PR for the manifest parsing of the first milestone of [GSoC Project : Build Script Delegation](https://summerofcode.withgoogle.com/programs/2025/projects/nUt4PdAA)

Bad copy/paste?

src/cargo/core/compiler/build_runner/mod.rs
```diff
@@ -420,29 +422,40 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
...
     }
 
     /// Returns the metadata hash for the `RunCustomBuild` Unit associated with
     /// the given unit.
     ///
     /// If the package does not have a build script, this returns None.
-    pub fn find_build_script_metadata(&self, unit: &Unit) -> Option<UnitHash> {
-        let script_unit = self.find_build_script_unit(unit)?;
-        Some(self.get_run_build_script_metadata(&script_unit))
+    pub fn find_build_script_metadata(&self, unit: &Unit) -> Option<Vec<UnitHash>> {
```
epage: `find_build_script_metadata` -> `find_build_script_metadatas` to express through the name that this is plural

src/cargo/core/compiler/compilation.rs
```diff
@@ -45,7 +45,7 @@ pub struct Doctest {
     /// The script metadata, if this unit's package has a build script.
     ///
     /// This is used for indexing [`Compilation::extra_env`].
-    pub script_meta: Option<UnitHash>,
+    pub scripts_meta: Option<Vec<UnitHash>>,
```
epage: Always a weird question on whether to pluralize the first part or second. I lean towards the second

src/cargo/core/compiler/compilation.rs
```diff
@@ -61,7 +61,7 @@ pub struct UnitOutput {
     /// The script metadata, if this unit's package has a build script.
     ///
     /// This is used for indexing [`Compilation::extra_env`].
-    pub script_meta: Option<UnitHash>,
+    pub script_meta: Option<Vec<UnitHash>>,
```
epage: Should we also pluralize this?

src/cargo/core/compiler/compilation.rs
```diff
@@ -190,14 +190,14 @@ impl<'gctx> Compilation<'gctx> {
...
-        self.fill_env(cmd, &unit.pkg, None, unit.kind, ToolKind::Rustc)
+        self.fill_env(cmd, &unit.pkg, &None, unit.kind, ToolKind::Rustc)
     }
 
     /// Returns a [`ProcessBuilder`] for running `rustdoc`.
     pub fn rustdoc_process(
         &self,
         unit: &Unit,
-        script_meta: Option<UnitHash>,
+        script_meta: &Option<Vec<UnitHash>>,
```
epage: This should be switched to `Option<&Vec<UnitHash>>,` (accessed via `Option::as_ref`)  or `Option<&[UnitHash]>,`  (accessed via `Option::as_deref`)

epage: True for all other `&Option`s

src/cargo/core/compiler/build_runner/mod.rs
```diff
@@ -420,29 +422,40 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
...
                     && unit_dep.unit.pkg.package_id() == unit.pkg.package_id()
             })
             .map(|unit_dep| unit_dep.unit.clone())
+            .collect();
+        if build_script_units.is_empty() {
+            None
+        } else {
+            Some(build_script_units)
+        }
     }
```
epage, line 446: What order is this returning units?  We should do it in a deterministic order, preferably one controlled by the user.

epage: If this commit is just about changing types, we can always add this in a later commit

tests/testsuite/build_scripts_multiple.rs
```diff
@@ -470,17 +469,18 @@ fn custom_build_script_second_index_script_failed() {
         .file("build2.rs", "fn main() { std::process::exit(101); }")
         .build();
 
-    p.cargo("build -v")
+    p.cargo("check -v")
```
epage, line 472: Mind moving this to its own commit.  If there are any other `build`s that you added, they should probably all be changed to `check`.  We try to avoid `build` as much as possible in CI because it is slow.  This is something I overlooked before.

src/cargo/core/compiler/unit_dependencies.rs
epage, line 1: > Add multiple build scripts to unit deps 

This seems like this belongs as a commit before " Use Vector for build script Units and UnitHash".  I'm even fine with it being its own PR.

tests/testsuite/build_scripts_multiple.rs
namanlp: Sorry, I did this by mistake 😅 I was checking whether `check` produces error or not

src/cargo/core/compiler/unit_dependencies.rs
namanlp: I think we can do this in same PR, but can reorder for sure

tests/testsuite/build_scripts_multiple.rs
epage: Except this is a change we should do.  `build` is a lot slower and uses up more disk space (a problem we have in CI).  We want as many tests as possible to use `check` instead of `build`.

epage: In another commit, could you fix this for https://github.com/rust-lang/cargo/blob/37ac0c28b83751f334f8f1b881db0b9d3f0bfb1f/tests/testsuite/build_scripts_multiple.rs#L413

src/cargo/core/compiler/unit_dependencies.rs
epage: Where is this at in finishing up the extraction of build script data?  One reason to split is the rest seems ready to go and it could be good to merge incrementally.

src/cargo/core/compiler/build_runner/mod.rs
namanlp: It will return the units in the same order as user is providing as far as I could trace the code. In `dep_build_script`, we are simply iterating and on targets. 

I verified it in my personal test repo. ( by debugging )

namanlp:
I am marking this PR Ready for Review. Please do let me know if I have missed anything

src/cargo/core/compiler/build_runner/mod.rs
epage, line 1: Now that we have multiple build scripts working, we need to update https://github.com/rust-lang/cargo/blob/4a137fa64dc1a7bdc619979eec93beff267e5280/tests/testsuite/build_scripts_multiple.rs#L413

src/cargo/core/compiler/compilation.rs
```diff
@@ -343,10 +343,12 @@ impl<'gctx> Compilation<'gctx> {
...
-            if let Some(env) = self.extra_env.get(&meta) {
-                for (k, v) in env {
-                    cmd.env(k, v);
+        if let Some(meta_vec) = script_metas {
+            for meta in meta_vec {
+                if let Some(env) = self.extra_env.get(meta) {
+                    for (k, v) in env {
+                        cmd.env(k, v);
+                    }
                 }
```
epage, line 352: What should happen when two build scripts output conflicting information?
- Last one specified wins (current behavior, see also https://github.com/rust-lang/cargo/pull/15704#discussion_r2172123815
- First one specified wins
- Error
- Unspecified

Whichever route we go, we should have a test for it.  In fact, environment variable tests are probably the best way to test multiple build scripts ran.

epage:
Also, another case we need to handle is `OUT_DIR`.  That could be a follow up PR.

src/cargo/core/compiler/compilation.rs
namanlp: While performing experiments, I don't know why, last one specified doesn't win. Rather, the lexicographically smallest build script wins.

For example, I have 2 build scripts, `build1.rs` and `build2.rs`, building files `a.cpp` and `b.cpp` (using `cc`), such that both `a.cpp` and `b.cpp` has functions `foo`, but `foo` is different in these files.

So, if we run it, no matter the order in `Cargo.toml`, `build1.rs` always wins, and hence, `a.cpp` always wins. Also, changing `a.cpp` to `d.cpp` doesn't change anything, but if we change `build1.rs` to `build3.rs`, `build2.rs` wins.

And it is deterministic, and not random, I have tested it like 50 times on my system ( I made a little bash script that `cargo clean` and  then `cargo run` ). I am not completely sure why though...

namanlp: For some reason, it is exactly opposite in environment variables. The lexicographically largest script, ie, `build3.rs` always wins.

epage: This is now documented in the PR description

tests/testsuite/build_scripts_multiple.rs
```diff
@@ -639,27 +650,45 @@ fn bar() {
...
 
-    p.change_file("assets/foo.txt", "foo updated");
     p.cargo("check -v")
+        .masquerade_as_nightly_cargo(&["multiple-build-scripts"])
         .with_stderr_data(str![[r#"
 [FRESH] foo v0.1.0 ([ROOT]/foo)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 
 "#]])
         .run();
```
epage: This no-op check run was added but we have one just before this.  How come?

tests/testsuite/build_scripts_multiple.rs
```diff
@@ -639,27 +650,45 @@ fn bar() {
...
 
-    // Editing bar.txt will recompile
+    // Editing foo.txt will also recompile now since they are separate build scripts
+    p.change_file("assets/foo.txt", "foo updated");
```
epage: Editing what is being tested should be in a prior commit.  This commit should only be about switching the test to use two build scripts and updating the output

tests/testsuite/build_scripts_multiple.rs
namanlp: Earlier, even if we updated `foo.txt`, it didn't trigger recompiling.

PR #15705, opened 2025-06-26
fix: Expand error messages around path dependency on `cargo package` and `cargo publish`

### LINKED ISSUE #15607: Confusing error when publishing a package with a path dependency without a version requirement
### Problem

I used to use a cargo plugin to publish an entire workspace to [crates.io](http://crates.io/), but now just tried the (unstable) built-in functionality with:

```
cargo +nightly publish -Z package-workspace --workspace
```

But I get this error reported:

```
error: all dependencies must have a version specified when publishing.
dependency `pigdef` does not specify a version
Note: The published dependency will use the version from crates.io,
the `path` specification will be removed from the dependency declaration.
```

The crate mentioned (pigdef) inherits the version from the workspace:

```
[package]
name = "pigdef"
description = "Core struct and const definitions for multiple crates in the pigg project"
version.workspace = true
edition.workspace = true
authors.workspace = true
license.workspace = true
documentation.workspace = true
repository.workspace = true
homepage.workspace = true
readme.workspace = true
exclude.workspace = true
rust-version.workspace = true
```

My repo is here: https://github.com/andrewdavidmackenzie/pigg

Workspace members inheriting the package version from the workspace Cargo.toml seems fairly common practice, and so it would be great to have it working with `cargo publish --workspace`

### Steps

- Have a workspace repo with one or more members
- Have the members inherit the package version from the workspace (`rust-version.workspace = true`)
- Try to publish the workspace by running
```
cargo +nightly publish -Z package-workspace --workspace
```
### Possible Solution(s)

_No response_

### Notes

_No response_

### Version
cargo version is 1.87.0
cargo nightly is...

```
cargo +nightly --version
cargo 1.89.0-nightly (68db37499 2025-05-22)
```
(so a week old, and seems to be the latest?)

motorailgun:
fix #15607.

Current error message about path/git dependency being inside manifest on `cargo package` and `cargo publish` lacks some information and it can be confusing. This PR adds some context on which manifest error originates from, so that users can understand the exact cause of the error cargo reports.


epage (approved):
Thanks!

PR #15707, opened 2025-06-27
Fix for #15703

### LINKED ISSUE #15703: Cargo bench can't find file where it does exist.
Valid Cargo.toml that includes:

```
[[bench]]
name = "benches.rs"
harness = false
```

cause `cargo bench` to say:

```
error: failed to parse manifest at `Cargo.toml`

Caused by:
  can't find `benches.rs` bench at `benches/benches.rs` or `benches/benches.rs/main.rs`. Please specify bench.path if you want to use a non-default path.
```

But:
```
$ ls benches/
benches.rs
```

Fix is to use `name = "benches"`, which would have been obvious if the error has used ".rs.rs", so better would probably have been:
```
error: failed to parse manifest at `Cargo.toml`

Caused by:
  can't find `benches.rs` bench at `benches/benches.rs.rs` or `benches/benches.rs/main.rs`. Please specify bench.path if you want to use a non-default path.
```

Tested on: cargo 1.90.0-nightly (84709f085 2025-06-22)

hkBst:

fixes #15703


tests/testsuite/bench.rs
epage, line 1: Looks like there are similar tests in `tests/testsuite/build.rs`, e.g. `non_existing_test`.  Would be good to group with those and to test this with a non-bench case so it runs on stable so its more likely it will be seen.

tests/testsuite/bench.rs
```diff
@@ -3,6 +3,36 @@
...
+#[cargo_test(nightly, reason = "bench")]
+fn cargo_bench_find_benches() {
+    let mut manifest = basic_lib_manifest("foo");
+    manifest = manifest.replace("2015", "2024");
+    manifest.push_str(
+        r#"
+        [[bench]]
+        name = "benches.rs"
+        "#,
+    );
```
epage: In a case like this, I'd recommend just writing out the manifest

tests/testsuite/bench.rs
hkBst: Thanks! It seems to work without nightly, maybe since no actual benchmarks need to be run...

hkBst: And perhaps it would be better to actually make this work, instead of writing a test to forbid it?

src/cargo/util/toml/targets.rs
```diff
@@ -752,7 +752,7 @@ fn infer_any(package_root: &Path, entry: &DirEntry) -> Option<(String, PathBuf)>
 
 fn infer_file(package_root: &Path, entry: &DirEntry) -> Option<(String, PathBuf)> {
     let path = entry.path();
-    let stem = path.file_stem()?.to_str()?.to_owned();
+    let stem = path.file_name()?.to_str()?.to_owned();
```
hkBst: @epage: This seems to be the root cause which strips the extension from the bench name. However with this change it produces: "[ERROR] invalid character '.' in crate name: `benches.rs`" and I can't find where that error is produced :(

epage: If you could put the fix in a commit after the test is added, that way the change in the test will show what the fix did

epage: This will change how we discover files and would need a lot of care and extra planning to make it work.

What I was suggesting is we just change https://github.com/rust-lang/cargo/blob/7388d86d28b8599968e6e6f58a6f4e5528b604a1/src/cargo/util/toml/targets.rs#L977-L981 to not call `set_ext` but instead append the extension.

hkBst: That's probably the smart thing to do, for which I was eyeballing [add_extension](https://doc.rust-lang.org/nightly/std/path/struct.PathBuf.html#method.add_extension). I don't suppose we can use this nightly API?

epage: No, cargo must work on stable.

tests/testsuite/bench.rs
epage: Not sure what you mean by that last comment.

The only reason nightly isn't needed is this errors before it gets to that point.  I would still lean towards using something besides benches.

Either way, this should be moved next to those other tests.

src/cargo/util/toml/targets.rs
```diff
@@ -972,15 +972,14 @@ fn target_path_not_found_error_message(
...
             let mut path = target_path.clone();
-            path.set_extension("rs");
+            path.push(format!("{name}.rs"));
             path
         };
         let target_path_subdir = {
-            target_path.push("main.rs");
+            target_path.extend([name, "main.rs"]);
             target_path
         };
```
epage, line 978: In #15703, I mentioned the possibility of going a step beyond and doing a typo check using [`closest_msg`](https://docs.rs/cargo/latest/cargo/util/edit_distance/fn.closest_msg.html).  Is that something you would be interested in doing?  If so, it should at least be in a different commit.  If you do it, let me know whether it will be a part of this PR so I know whether to wait for it or not.

tests/testsuite/bench.rs
hkBst: I'll look into the non-bench equivalents and moving this test. It's just that the build.rs test file is already very long, which makes it very hard to get a good sense of what is in there, and I don't quite see how this test fits in there. Or perhaps you're saying that it does not really matter that we're doing "cargo bench" since it errors in a path that is common to most cargo commands, namely build. That makes a lot of sense, except I wouldn't expect it to be called build. :D

src/cargo/util/toml/targets.rs
hkBst: I separated the test and fix in separate commits, with the test failing in the first commit. Is that what you meant? Or do you want the test passing immediately, with changes to the test in the commit with the fix?

src/cargo/util/toml/targets.rs
hkBst: I hadn't understood what you meant by that before, and I haven't looked into that function yet, but I think that makes a lot of sense, and I definitely want to have a look at it for this PR.

Your help so far has been great BTW, so thanks!

tests/testsuite/bench.rs
hkBst: I've stuck the tests in bad_config.rs. Let me know if you think this is appropiate.

src/cargo/util/toml/targets.rs
hkBst: Having looked into `closest_msg`, I don't really see how to use that here, so please consider this PR as is.

tests/testsuite/bench.rs
epage: I was hoping to have these tests right next to https://github.com/rust-lang/cargo/blob/f013ef54bb9945d8853d56cc9df145834e937ccc/tests/testsuite/build.rs#L2254-L2374

src/cargo/util/toml/targets.rs
epage: >  Or do you want the test passing immediately, with changes to the test in the commit with the fix?

This.  See the tip at https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request

src/cargo/util/toml/targets.rs
epage: I'm fine splitting that out but figured I'd at least take the time to explain it.

This is the caller to `target_path_not_found_error_message`:
https://github.com/rust-lang/cargo/blob/f013ef54bb9945d8853d56cc9df145834e937ccc/src/cargo/util/toml/targets.rs#L1026-L1058

This function receives an `inferred: &[(String, PathBuf)],`.  This is all of the build-target `name`s and `path`s we found that we then do lookups within.  If we then pass that to `target_path_not_found_error_message`, we can then call `closest_msg` , comparing the name the user specified with the names we found on disk.  If `closest_msg` didn't find a match, we could then show all of the names to the user, making it easier to debug what went wrong with them specifying one.

src/cargo/util/toml/targets.rs
hkBst: Alright, will adjust.

tests/testsuite/bench.rs
hkBst: I agree these tests probably belong together. What do you say about me moving these out of build.rs and changing "cargo build" to "cargo check" (I'm assuming check is enough here too)?

src/cargo/util/toml/targets.rs
hkBst: Ah yes, that seems to have unblocked me. Added in latest commit.

tests/testsuite/bench.rs
epage: In its own commit, feel free to move these over to `bad_config.rs` and in another commit swithc from `build` to `check`.

epage (approved):

epage:
Thanks!

hkBst:
@epage thanks for your guidance!

PR #15712, opened 2025-06-28
add CARGO_CRATES_DOWNLOAD_URL env var instead of config.json dl

abtzz:
add CARGO_CRATES_DOWNLOAD_URL env var Set the download URL, replace the dl in the
  registry index config.json, and download through a nexus proxy when in the intranet
  and without external network

### DIFF
--- a/src/cargo/sources/registry/download.rs
+++ b/src/cargo/sources/registry/download.rs
@@ -22,12 +22,14 @@ use std::fs::{self, File, OpenOptions};
 use std::io::prelude::*;
 use std::io::SeekFrom;
 use std::str;
+use std::env;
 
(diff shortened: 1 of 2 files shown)


epage:
Thanks for wanting to improve Cargo!

Could you create an issue, focusing on the problem you are having that led to this implementation? We generally [ask for Issues before PRs](https://doc.crates.io/contrib/process/working-on-cargo.html#before-hacking-on-cargo) so we can dig into requirements, potential solutions, and to ensure alignment. This avoid wasted effort on your side and help to keep the conversation in one place as there can end up being multiple PRs for the same need.

I'm going to go ahead and close this to focus the conversation on that issue once its created.

PR #15729, opened 2025-07-05
test: Switch config tests to use snapshots

epage:

This is prep for updating `toml` which will change some of these error messages


weihanglo (approved):
Thanks!

PR #15734, opened 2025-07-08
Mark cachelock tests that rely on interprocess blocking behaviour as unsupported on AIX.

uyoyo0:
This PR marks several tests within cache_lock.rs as unsupported on the AIX platform. The tests relies on flock() behaviour that is not supported on AIX.


tests/testsuite/cache_lock.rs
```diff
@@ -112,6 +112,7 @@ fn multiple_shared() {
     a_b_nested(CacheLockMode::Shared, CacheLockMode::Shared);
 }
 
+#[cfg_attr(target_os = "aix", ignore)]
```
weihanglo: Could we add a description to ignore, so others could know why without running `git blame` to figure out?

uyoyo0: Thanks for the suggestion, I just amended the commit with those changes.

weihanglo: Sorry it is still unclear to me. What kind of behavior is unsupported in AIX? Would you mind adding details and may AIX official docs to the PR description? Doesn't need to update the source code though.

weihanglo: Oh I see interprocess blocking. Never mind. Going to merge. Thank you

weihanglo (approved):

Issue #15649, opened 2025-06-09
Use another hashing algorithm? 250K calls to Siphash


blyxyas:
While profiling Clippy I came accross 250k calls to `<core::hash::sip::Hasher<core::hash::sip::Sip13Rounds> as core::hash::Hasher>::write`, and they turn out to be the most time-consuming by themselves (at least that's what callgrind says), all of them coming from Cargo.

These numbers in particular is from `bumpalo-3.16.0`

The 4 callers of this function are from hashing `&url::Url` (`hash_one` -> 26 296 times), hashing `&semver::Version` (`hash_one` -> 83 342x), hashing `package_id::PackageId` (`hash` -> 48 554x) and `dependency::Dependency` (`hash` -> 31 099x)

## To reproduce

```
cd bumpalo-3.16.0
CARGO_TARGET_DIR=/tmp/m1 valgrind --tool=callgrind --dump-instr=yes --collect-jumps=yes --trace-children=yes /home/meow/.rustup/toolchains/nightly-2025-05-31-x86_64-unknown-linux-gnu/bin/cargo clippy
```

Note that this is without incremental compilation, thus the `CARGO_TARGET_DIR` env var.

I'm not sure if there's a reason why Cargo uses the default hasher.

weihanglo:
Cargo leverages <https://github.com/rust-lang/rustc-stable-hash>, as we want path on file system as stable as possible, regardless endianness or platform. Tha said, not everything in this codebase has this requirement There are some previous attempts but got stale
https://github.com/rust-lang/cargo/pull/14665


Edit: Oh I see. You're talking about siphash from std not from rustc-stable-hash. The thing about stable hash may be irrelevant here.

weihanglo:
I wonder how bad it was in terms of total run time of the clippy command being benchmarked.

blyxyas:
On this crate, it seems to represent a 9.89% of the runtime, with 14.155.123 instructions.

epage:
If we do this, a question is which hash.  Currently, Cargo depends on `ahash` through `hashbrown@0.14`.  However, `hashbrown@0.15` has switched to `foldhash` ( rust-lang/hashbrown#563).  For toml, I'm following suit in toml-rs/toml#956

blyxyas:
> On this crate, it seems to represent a 9.89% of the runtime, with 14.155.123 instructions.

[...]

PR #15736, opened 2025-07-08
perf: Speed up TOML parsing by upgrading toml

epage:

For numbers, see https://epage.github.io/blog/2025/07/toml-09/

Further areas for improvement:
- Enable `fast_hash` (see #15649)
- Only track spans for local manifests, allowing us to skip the `make_owned` call for most packages

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -211,7 +211,7 @@ tar.workspace = true
 tempfile.workspace = true
 thiserror.workspace = true
 time.workspace = true
-toml = { workspace = true, features = ["std", "serde", "parse", "display"] }
+toml = { workspace = true, features = ["std", "serde", "parse", "display", "preserve_order"] }
 toml_edit.workspace = true
 tracing = { workspace = true, features = ["attributes"] }
 tracing-subscriber.workspace = true


tests/testsuite/alt_registry.rs
```diff
@@ -751,13 +751,11 @@ fn bad_registry_name() {
...
 
 
-  --> Cargo.toml:8:17
-   |
- 8 | /                 [dependencies.bar]
- 9 | |                 version = "0.0.1"
-10 | |                 registry = "bad name"
-   | |_____________________________________^
-   |
+ --> Cargo.toml:8:17
```
weihanglo, line 754: This seems like a diagnostic regression?

tests/testsuite/git.rs
```diff
@@ -2809,11 +2808,11 @@ fn invalid_git_dependency_manifest() {
         .with_status(101)
         .with_stderr_data(str![[r#"
 [UPDATING] git repository `[ROOTURL]/dep1`
-[ERROR] duplicate key `categories` in table `package`
```
weihanglo, line 2812: Diagnostic regression IMO, but not a big deal.

src/cargo/util/lints.rs
```diff
@@ -191,43 +190,33 @@ fn verify_feature_enabled(
...
-                let leaf_prefix_span = leaf_decor.prefix().and_then(|p| p.span());
-                let leaf_suffix_span = leaf_decor.suffix().and_then(|s| s.span());
-                if let (Some(leaf_prefix_span), Some(leaf_suffix_span)) =
-                    (leaf_prefix_span, leaf_suffix_span)
-                {
-                    Some(leaf_prefix_span.start..leaf_suffix_span.end)
-                } else {
-                    key.span()
-                }
+                Some(key.span())
```
weihanglo, line 206: So the span method is effectively the same as the previous implementation that includes decorations?
(Cannot tell unless I look into the implementation)

weihanglo (commented):
None of these review comments are blocking. Thanks for the efforts!

tests/testsuite/git.rs
epage: As this highlights the duplicate key, I figured that removing it from the message wasn't too big of a deal.

src/cargo/util/lints.rs
epage: We were getting the span of the entire header by grabbing the spans of the decor and making a span of what came between.  Looking at https://github.com/rust-lang/cargo/pull/15736/files/539a48452a03a68d18cd32481e364550ee42f329#diff-1c71ec9c445774ebb8b0c3dd4c35817dda698c78e15fb83d8a61d8b145c5dd34, it appears that that behavior is now the default for spans so this is built-in now.  Even if it wasn't, its likely not worth it to keep using `toml_edit`.

weihanglo (approved):
Thanks for the performance optimization and the good write-up!

blyxyas:
Is there anything blocking from adding the `fast_hash` feature for a further ~14% improvement? Seems stable enough to already be used in hashbrown, why not also use it on Cargo?

epage:
@blyxyas I was just wanting to leave that to work on a custom hash more generally (#15649) to avoid any questions blocking this PR on the hash used, what hashes are appropriate, etc.

tests/testsuite/alt_registry.rs
epage: There are two tiers for this
- Untagged enums and flattened structs in serde which mean we don't get precise field data, only the entire table
- its weird to know what spans are "right" for standard tables and implicit tables because their definition is split over multiple locations.

In this case, the error is for `registry`s value but we are highlighting the entire table.

In `toml_parse`, I kept things simple by only tracking spans for the table headers and not table bodies.  I'm open to changing that.

PR #15759, opened 2025-07-20
Remove unnecessary target-c-int-width from target specs

Noratrieb:

This is not necessary, as 32 is the default, and actually of the wrong type now since it's a number now.
When planning to make these type mismatches error in https://github.com/rust-lang/rust/pull/144218, cargo would fail here, so I just removed it.

This custom target test very much shows how Cargo should be a subtree, in this case it was fine because there's a compatible fix that I can push now, otherwise it would have been very annoying.

If the test suite passes, it works

### DIFF
--- a/tests/build-std/main.rs
+++ b/tests/build-std/main.rs
@@ -271,7 +271,6 @@ fn cross_custom() {
                 "arch": "x86_64",
                 "target-endian": "little",
                 "target-pointer-width": "64",
-                "target-c-int-width": "32",
                 "os": "none",
                 "linker-flavor": "ld.lld"
             }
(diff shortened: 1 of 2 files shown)


weihanglo (approved):

weihanglo:
How did you ever found this btw? Those tests shouldn't be run on rust-lang/rust by default.

weihanglo:
> This custom target test very much shows how Cargo should be a subtree, in this case it was fine because there's a compatible fix that I can push now, otherwise it would have been very annoying.

Usually it would be annoying to Cargo maintainers, and we'll see tests failing on nightly next day. But yes I agree this is another data point that subtree could help rustc contributors.

Noratrieb:
> How did you ever found this btw? Those tests shouldn't be run on rust-lang/rust by default.

Ah, I see. I just grepped through the tree to find references to `target-c-int-width` to fix them (since some rustc UI tests failed because of it), and realized I was in Cargo.

PR #15764, opened 2025-07-22
fix: `no-proc-macro` is overridden by subsequent edges

### LINKED ISSUE #15763: `cargo tree -e no-proc-macro` only takes effect when `no-proc-macro` is placed at the end
### Problem

When using the `cargo tree` command with the `-e` flag, the `no-proc-macro` option only takes effect when placed at the end, causing the output to differ depending on its position in the command.

### Steps

1. `cargo new deptest`
2. `cd deptest`
3. `cargo add argh@0.1.13`
4. The outputs vary depending on the order in which `no-proc-macro` is specified. The `no-proc-macro` option only takes effect when it is placed at the end.
    ```shell
    ❯ cargo tree -e no-build,no-proc-macro -d
    warning: nothing to print.
    
    To find dependencies that require specific target platforms, try to use option `--target all` first, and then narrow your search scope accordingly.
    ```

    ```shell
    ❯ cargo tree -e no-proc-macro,no-build -d
    argh_shared v0.1.13
    └── argh_derive v0.1.13 (proc-macro)
        └── argh v0.1.13
            └── deptest v0.1.0 (/Users/chenruiwei/Code/rust/deptest)
    
    argh_shared v0.1.13
    └── argh v0.1.13 (*)
    ```

### Possible Solution(s)

The issue is caused by the code below. While iterating through all edges, the `no_proc_macro` flag is overridden by subsequent edges.

https://github.com/rust-lang/cargo/blob/5b295b77ee4379dec14242ead7b75c84ec4e9f75/src/bin/cargo/commands/tree.rs#L251-L262

### Notes

_No response_

### Version

```text
cargo 1.90.0-nightly (930b4f62c 2025-06-28)
release: 1.90.0-nightly
commit-hash: 930b4f62cfcd1f0eabdb30a56d91bf6844b739bf
commit-date: 2025-06-28
host: aarch64-apple-darwin
libgit2: 1.9.0 (sys:0.20.2 vendored)
libcurl: 8.7.1 (sys:0.4.80+curl-8.12.1 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 3.5.0 8 Apr 2025
os: Mac OS 15.5.0 [64-bit]
```

ChenRuiwei:
To close #15763.

See steps in #15763.

### DIFF
--- a/tests/testsuite/cargo_tree/deps.rs
+++ b/tests/testsuite/cargo_tree/deps.rs
@@ -2357,12 +2357,11 @@ foo v0.1.0 ([ROOT]/foo)
 "#]])
         .run();
 
-    // change flag order, no-proc-macro takes no effect if followed by other flags
+    // change flag order, expecting the same output
     p.cargo("tree -e no-proc-macro,normal")
         .with_stdout_data(str![[r#"
 foo v0.1.0 ([ROOT]/foo)
-├── dep v1.0.0
-└── pm v1.0.0 (proc-macro)
+└── dep v1.0.0
 
 "#]])
         .run();
--- a/tests/testsuite/cargo_tree/deps.rs
+++ b/tests/testsuite/cargo_tree/deps.rs
@@ -2318,3 +2318,52 @@ foo v1.0.0 ([ROOT]/foo)
 "#]])
         .run();
 }
+
+#[cargo_test]
+fn no_proc_macro_order() {
+    Package::new("dep", "1.0.0").publish();
+    Package::new("pm", "1.0.0").proc_macro(true).publish();
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+            [package]
+            name = "foo"
+            version = "0.1.0"
+
+            [dependencies]
+            pm = "1.0"
+            dep = "1.0"
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+
+    p.cargo("tree")
+        .with_stdout_data(str![[r#"
+foo v0.1.0 ([ROOT]/foo)
+├── dep v1.0.0
+└── pm v1.0.0 (proc-macro)
+
+"#]])
+        .run();
+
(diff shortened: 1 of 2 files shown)


weihanglo (commented):
Thanks for the contribution! Would you like to add a test for this fix?

Generally, new tests would be added in a commit before behavior changes, and pass. This shows the current buggy behavior. The next commit would include the actual code fix and the test change. The diff with the fix would show how the behavior changed.

ChenRuiwei:
> Thanks for the contribution! Would you like to add a test for this fix?
> 
> Generally, new tests would be added in a commit before behavior changes, and pass. This shows the current buggy behavior. The next commit would include the actual code fix and the test change. The diff with the fix would show how the behavior changed.

Thanks for your review!

New test has been added.

tests/testsuite/cargo_tree/deps.rs
```diff
@@ -2318,3 +2318,51 @@ foo v1.0.0 ([ROOT]/foo)
...
+└── dep v1.0.0
+
+"#]])
+        .run();
+
+    // change flag order, expecting the same output
+    p.cargo("tree -e no-proc-macro,normal")
+        .with_stdout_data(str![[r#"
+foo v0.1.0 ([ROOT]/foo)
+└── dep v1.0.0
```
weihanglo, line 2364: nit: should be like this in this commit, so it can be served as a minimal reproducible example, and we love to see tests pass in every commit. After this, the next commit will have both the fix, and the test change to demonstrate the behavior change.


```suggestion
├── dep v1.0.0
└── pm v1.0.0 (proc-macro)
```

another nit: the commmit message should be scoped as "test" not "feat"

ChenRuiwei: All fixed by force push.

weihanglo:
Thanks for the contribution!

Issue #15666, opened 2025-06-15
Timings graph should fit the screen by default


Expurple:
### Problem

When I open an HTML file with timings, it always chooses "Scale: 20". Depending on the project, the graph is always too wide or too narrow with this default setting. I always manually adjust the scale to fit the screen, and that's very annoying.

Here's how wide it is for my proprietary project (note the horizontal scrollbar at the bottom, the graph is 2x wider than the window):

![Wide proprietary graph](https://github.com/user-attachments/assets/99fcedb6-08f6-42e4-b2a5-a0bb8bdbacfe)

And here's how narrow it is for SeaORM:

![Narrow SeaORM graph](https://github.com/user-attachments/assets/2e7dd451-9242-4a22-9e59-6de03a6e9980)

The graph width seems roughly proportional to the total compile time? (134 vs 14 seconds)

It's unclear what unit "Scale" is supposed to be.

### Steps

1. `git clone --single-branch --depth=1  https://github.com/SeaQL/sea-orm.git`
2. `cd sea-orm/`
3. `cargo +nightly build --timings`
4. `firefox target/cargo-timings/cargo-timing.html`

### Possible Solution(s)

Automatically choose the "Scale" setting that makes the graph fit the screen. At least, there should be a button that jumps to the optimal scale.

The proprietary graph with manually adjusted Scale (note the whitespace on the right. The horizontal scrollbar at the bottom is gone):

![Adjusted proprietary graph](https://github.com/user-attachments/assets/cfae5925-fa84-453f-8bcc-6983c7ed25a0)

SeaORM graph with manually adjusted Scale:

![Adjusted SeaORM graph](https://github.com/user-attachments/assets/71a81276-950b-4b40-9a72-fccb1f6ea38f)

I can't even manually adjust SeaORM graph to screen width, because of the max scale limit. But that's a separate issue and there could be legitimate reasons. I'd be OK if it at least auto-expanded to this max allowed scale.

### Notes

I use Firefox 136.0 on Linux. I've also checked Chromium 137.0.7151.103, the same thing happens.

Could be somehow related: https://github.com/rust-lang/cargo/issues/8850

### Version

```text
cargo 1.89.0-nightly (fc1518ef0 2025-06-06)
release: 1.89.0-nightly
commit-hash: fc1518ef02b77327d70d4026b95ea719dd9b8c51
commit-date: 2025-06-06
host: x86_64-unknown-linux-gnu
[...]

PR #15766, opened 2025-07-22
Make timings graphs scalable to user's window

Kobzol:

This PR changes the way the charts produced by `cargo build --timings` scale. It changes the scale slider so that its min/max values adapt to the duration of the build, to allow zooming in/out even for very short build durations. It also automatically initializes the scale value based on the client's window width.

The number of pixels per second per scale value has been changed from 1 to 8, to avoid having too many scale values for the given duration of supported chart widths, which I have determined in this PR to be `[200, 4096]` pixels.

https://github.com/user-attachments/assets/3e6e9f14-eabe-425a-a568-9fcb5c835145

Run `cargo build --timings` e.g. on https://github.com/BurntSushi/ripgrep. Then open the resulting page in a browser, and try to enlarge/ensmall the window (possibly using mobile emulation), and see how the charts react to window size.

Fixes: https://github.com/rust-lang/cargo/issues/15666

### DIFF
--- a/src/cargo/core/compiler/timings.js
+++ b/src/cargo/core/compiler/timings.js
@@ -23,6 +23,42 @@ let UNIT_COORDS = {};
 let REVERSE_UNIT_DEPS = {};
 let REVERSE_UNIT_RMETA_DEPS = {};
 
+const MIN_GRAPH_WIDTH = 200;
+const MAX_GRAPH_WIDTH = 4096;
+
+// How many pixels per second is added by each scale value
+const SCALE_PIXELS_PER_SEC = 8;
+
+function scale_to_graph_width(scale) {
+  // The scale corresponds to `SCALE_PIXELS_PER_SEC` pixels per seconds.
+  // We thus multiply it by that, and the total duration, to get the graph width.
+  const width = scale * SCALE_PIXELS_PER_SEC * DURATION;
+
+  // We then cap the size of the graph. It is hard to view if it is too large, and
+  // browsers may not render a large graph because it takes too much memory.
+  // 4096 is still ridiculously large, and probably won't render on mobile
+  // browsers, but should be ok for many desktop environments.
+  // Also use a minimum width of 200.
+  return Math.max(MIN_GRAPH_WIDTH, Math.min(MAX_GRAPH_WIDTH, width));
+}
+
+// This function performs the reverse of `scale_to_graph_width`.
+function width_to_graph_scale(width) {
+  const maxWidth = Math.min(MAX_GRAPH_WIDTH, width);
+  const minWidth = Math.max(MIN_GRAPH_WIDTH, width);
+
+  const trimmedWidth = Math.max(minWidth, Math.min(maxWidth, width));
+
+  const scale = Math.round(trimmedWidth / (DURATION * SCALE_PIXELS_PER_SEC));
+  return Math.max(1, scale);
+}
+
+// Init scale value and limits based on the client's window width and min/max graph width.
+const scaleElement = document.getElementById('scale');
+scaleElement.min = width_to_graph_scale(MIN_GRAPH_WIDTH);
+scaleElement.max = width_to_graph_scale(MAX_GRAPH_WIDTH);
+scaleElement.value = width_to_graph_scale(window.innerWidth * 0.75);
+
 // Colors from css
 const getCssColor = name => getComputedStyle(document.body).getPropertyValue(name);
 const TEXT_COLOR = getCssColor('--text');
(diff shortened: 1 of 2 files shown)


Kobzol:
As an alternative to filling the screen by defaut, we could set the scaler so that the graph takes e.g. 75% of screen width, might be a bit less "eye-popping".

epage:
What browsers have you tested on?

Kobzol:
Just Chrome, but that's essentially what most browsers use these days anyway :laughing: Also tested in Firefox now, looks fine.

src/cargo/core/compiler/timings.js
```diff
@@ -23,6 +23,36 @@ let UNIT_COORDS = {};
...
+    const width = window.innerWidth;
+    // Leave some margin around the chart, to account for axis labels
+    const max_available_space = Math.max(200, width - 150);
+
+    for (let scale = 1; scale < 100; scale++) {
+        const width = scale_to_graph_width(scale);
+
+        // If we ran out of space, return the previous scale
+        if (width > max_available_space) {
+            return Math.max(1, scale);
```
epage: Isn't this `max` redundant?

src/cargo/core/compiler/timings.js
```diff
@@ -23,6 +23,36 @@ let UNIT_COORDS = {};
...
+    return Math.min(200 + scale * 20, 4096);
+}
+
+// Guess a reasonable initial scale value to fit the graphs to the width of the user's window.
+function init_graph_scale() {
+    const width = window.innerWidth;
+    // Leave some margin around the chart, to account for axis labels
+    const max_available_space = Math.max(200, width - 150);
+
+    for (let scale = 1; scale < 100; scale++) {
```
epage: Can't we calculate this rather than do a linear search?

src/cargo/core/compiler/timings.js
Kobzol: Right, this should have actually been `scale - 1`, then the max makes sense. Thanks!

src/cargo/core/compiler/timings.js
Kobzol: Yeah, we can. But the reason I didn't do that is because we would have to implement the scaling logic twice, because it's not "reversible". We have one function that computes the width based on the scale, and in the other direction we'd need a function that computes the scale from the width. So we'd have to either duplicate the logic or create some shared constants on which those two functions would operate, but both risks the logic diverging. The for loop makes sure that we can't have a bug in the mapping between those two directions, because we only implement the direction scale -> width.

But if you don't like it, I can just reimplement the mapping logic from the other direction.

src/cargo/core/compiler/timings.js
epage: Ah, we exceeded, so pick the previous

src/cargo/core/compiler/timings.rs
```diff
@@ -780,7 +780,15 @@ static HTML_CANVAS: &str = r#"
   </tr>
   <tr>
     <td><input type="range" min="0" max="30" step="0.1" value="0" id="min-unit-time"></td>
-    <td><input type="range" min="1" max="50" value="20" id="scale"></td>
+    <!--
+        The scale has the following interpretation:
+        - Value=N corresponds to a graph width of 200 + N*20 px.
```
kpreid: This removes a useful characteristic of the current implementation: it is possible (though also the default) to set two different timings charts to identical scales. This is something I use a lot to evaluate attempted build time improvements.

Certainly there is value in making it a good size by default, but I would like to request that the scale _value_ continues to mean the same “length per time” unit for all charts, and not be relative to the overall build time, so that it is possible to set identical scales on two different charts.

weihanglo: See [#t-cargo > Making timings chart scalable](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Making.20timings.20chart.20scalable/with/530629537)

From @Kobzol:

> ok, I see. so we would automatically set the scale value to match the viewport width, and adjust the maximum scale value so that you can always scale the graph up to 4096 pixels, regardless of how long the build took

Kobzol:
Updated the implementation based on Zulip discussion. The PR description has been updated with the description of the current approach.

src/cargo/core/compiler/timings.js
```diff
@@ -23,6 +23,42 @@ let UNIT_COORDS = {};
...
+
+  const scale = Math.round(trimmedWidth / (DURATION * SCALE_PIXELS_PER_SEC));
+  return Math.max(1, scale);
+}
+
+// Init scale value and limits based on the client's window width and min/max graph width.
+const scaleElement = document.getElementById('scale');
+scaleElement.min = width_to_graph_scale(MIN_GRAPH_WIDTH);
+scaleElement.max = width_to_graph_scale(MAX_GRAPH_WIDTH);
+scaleElement.value = width_to_graph_scale(window.innerWidth * 0.75);
```
weihanglo, line 60: Where is the `0.75` from?

Regardless, feel good, and we can always tweak if people find a better default. Thank you for the PR!

Kobzol: That is an arbitrary constant, to fit ~75% of the screen :sweat_smile:

weihanglo: I figured that. Anyway, I feel 75% is pretty balanced and eye-friendly.

PR #15767, opened 2025-07-22
cargo-credential-libsecret: give FFI correctly-sized object

nabijaczleweli:

The type is
```c
typedef struct {
	const gchar *name;
	SecretSchemaFlags flags;
	SecretSchemaAttribute attributes[32];

	/* <private> */
	gint reserved;
	gpointer reserved1;
	gpointer reserved2;
	gpointer reserved3;
	gpointer reserved4;
	gpointer reserved5;
	gpointer reserved6;
	gpointer reserved7;
} SecretSchema;
```
so the current object we give it is 8 pointers too short

It's incredibly lucky that libsecret, at this time, only uses `reserved`, and not in any of the functions we call

Also, some obvious cleanups while I was there and comparing with [my implementation](https://github.com/nabijaczleweli/cargo-update/blob/v17.0.0/src/ops/mod.rs#L1443) from [cargo-update 17.0.0](https://github.com/nabijaczleweli/cargo-update/releases/v17.0.0).

Observe https://sources.debian.org/src/libsecret/0.20.5-3/libsecret/secret-schema.h/#L43 I suppose?

### DIFF
--- a/credential/cargo-credential-libsecret/src/lib.rs
+++ b/credential/cargo-credential-libsecret/src/lib.rs
@@ -168,7 +168,7 @@ mod linux {
 
             let index_url_c = CString::new(registry.index_url).unwrap();
             let mut error: *mut GError = null_mut();
-            let attr_url = CString::new("url").unwrap();
+            let attr_url = b"url\0".as_ptr() as *const gchar;
             let schema = schema();
             match action {
                 cargo_credential::Action::Get(_) => {
@@ -177,7 +177,7 @@ mod linux {
                             &schema,
                             null_mut(),
                             &mut error,
-                            attr_url.as_ptr(),
+                            attr_url,
                             index_url_c.as_ptr(),
                             null() as *const gchar,
                         );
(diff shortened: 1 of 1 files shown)


credential/cargo-credential-libsecret/src/lib.rs
```diff
@@ -145,51 +167,46 @@ mod linux {
             }
 
             let index_url_c = CString::new(registry.index_url).unwrap();
+            let mut error: *mut GError = null_mut();
+            let attr_url = b"url\0".as_ptr() as *const gchar;
```
weihanglo: cstr literal maybe?
Granted some other places may also need to update.

```suggestion
            let attr_url = c"url".as_ptr() as *const gchar;
```

nabijaczleweli: I was gonna, but nothing else seemed to use `c""` literals. Applied.

weihanglo (approved):
Thanks!

Issue #14834, opened 2024-11-18
Pull out the HTTP mock server as `cargo-dev-regsitry`


Rustin170506:
### Problem

In Cargo's unit tests, we used a mock HTTP server as our test infrastructure. But it is not a fully implemented registry. Some unit tests are just using custom handler to mock some registry API.

So as I discussed with @weihanglo and @epage, we both agree that having full featured dev registry in cargo would help the testing a lot. And maybe help us to catch some bugs that can only be caught in a real cargo registry.

At the same time, if we can have a standalone dev registry, it would help us a lot when testing or debugging the cargo itself. At the moment, it is difficult to test the real publish function, because setting up a real registry from scratch is difficult.

Furthermore, if we try to implement this dev registry, we can give other people a reference about how to implement the basic API for a cargo registry. We can also extend the common library for building a cargo registry. We believe this would not only help the cargo itself, but also help the community to implement a registry more easily.

### Proposed Solution

As discussed with @epage, he suggested that we can do the following things to pull out the mock HTTP server as our `cargo-dev-regsitry`.

- switch cargo's summary generation to use a normalized manifest, rather than a Package
- pull registry types like IndexSummary out into cargo-util-schemas
- pull out index/manifest/publish conversions into a cargo-util-registry
- switch our dev registry to using the above rather than reimplement
- pull our dev registry out into cargo-dev-regsitry with an optional cli for spinning it up

Rustin170506:
@arlosi If I remember correctly, I think we mentioned in one of our weekly meetings that we wanted a full-featured registry to test the Cargo, but I don't remember the context. Maybe it was related to authentication. Do you remember that requirement or use case?

epage:
For past references, see
- https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Downloading.20a.20crate.20and.20all.20of.20its.20dependencies/near/466097168
- https://hachyderm.io/@epage/112893814784038149

CC @shepmaster, @Turbo87 as this involves trying to share more code between Cargo and registries

> As discussed with @epage, he suggested that we can do the following things to pull out the mock HTTP server as our cargo-dev-regsitry.

To clarify, I expect `cargo-dev-regsitry` to be a first-party crate, not a built-in command.

> pull out index/manifest/publish conversions into a cargo-util-registry

[...]

PR #15770, opened 2025-07-23
feat(schema): Expose `IndexPackage`, the description of a package within a Registry Index

0xPoe:

ref https://github.com/rust-lang/cargo/issues/14834

In this pull request, I moved `IndexPackage` to `cargo-util-schemas` to enable third-party applications and `cargo-dev-registry` to utilize this common struct.

For instance, crates.io has its own `IndexPackage` located at https://github.com/rust-lang/crates.io/blob/005667f0416d03932a50309db6cc672945b033fe/crates/crates_io_index/data.rs#L5.

By moving it to `cargo-util-schemas`, we allow different applications to use this struct more easily.

It shouldn't break any tests; it's just a refactoring.


crates/cargo-util-schemas/src/core/index.rs
epage, line 1: Why is this in `core` instead of being top-level?

epage, line 1: Can you add a schema file?

0xPoe: Yes, I will add it and use the manifest schema as an example.

0xPoe: I believe it should be at the top level as the manifest. I will update it.

epage (approved):
Note for commit messages, the "refactor" is a "feature" commit for the schema

0xPoe:

PR #15776, opened 2025-07-26
Accessing each build script's `OUT_DIR` and in the correct order

namanlp:
Hi Everyone!

This PR is aimed to have some improvements over #15704

Now, multiple build scripts are built correctly. But there are some underlying issues, that this PR is targeting.

- Preserving the order of the build scripts: Earlier the build scripts were sorted by default, but now, the order will be preserved.

```toml
```

### DIFF
--- a/src/cargo/core/compiler/unit_dependencies.rs
+++ b/src/cargo/core/compiler/unit_dependencies.rs
@@ -30,7 +30,9 @@ use crate::core::dependency::{Artifact, ArtifactKind, ArtifactTarget, DepKind};
 use crate::core::profiles::{Profile, Profiles, UnitFor};
 use crate::core::resolver::Resolve;
 use crate::core::resolver::features::{FeaturesFor, ResolvedFeatures};
-use crate::core::{Dependency, Package, PackageId, PackageSet, Target, TargetKind, Workspace};
+use crate::core::{
+    Dependency, Feature, Package, PackageId, PackageSet, Target, TargetKind, Workspace,
+};
 use crate::ops::resolve_all_features;
 use crate::util::GlobalContext;
 use crate::util::interning::InternedString;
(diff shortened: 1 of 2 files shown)


src/cargo/core/compiler/unit_dependencies.rs
```diff
@@ -142,8 +144,28 @@ pub fn build_unit_dependencies<'a, 'gctx>(
...
+            list.sort_by_key(|unit_dep| {
+                if unit_dep.unit.target.is_custom_build() {
+                    (0, None)
+                } else {
+                    (1, Some(unit_dep.clone()))
+                }
+            });
+        } else {
+            list.sort();
+        }
```
epage: `for (unit, list) in &mut state.unit_dependencies` can avoid accessing the tuple fields

src/cargo/core/compiler/unit_dependencies.rs
```diff
@@ -142,8 +144,28 @@ pub fn build_unit_dependencies<'a, 'gctx>(
...
+            .manifest()
+            .unstable_features()
+            .require(Feature::multiple_build_scripts())
+            .is_ok();
+
+        let list = unit_dep.1;
+        if is_multiple_build_scripts_enabled {
+            list.sort_by_key(|unit_dep| {
+                if unit_dep.unit.target.is_custom_build() {
+                    (0, None)
```
epage, line 160: Comment that we are preserving the users order for build scripts

src/cargo/core/compiler/unit_dependencies.rs
```diff
@@ -142,8 +144,27 @@ pub fn build_unit_dependencies<'a, 'gctx>(
...
+            .unstable_features()
+            .require(Feature::multiple_build_scripts())
+            .is_ok();
+
+        // We do not sort build scripts to preserve the user-defined order.
+        // It ensures deterministic behaviour as well as user control over build scripts order.
+        if is_multiple_build_scripts_enabled {
+            list.sort_by_key(|unit_dep| {
+                if unit_dep.unit.target.is_custom_build() {
+                    (0, None)
```
epage: Code organization wise, a comment like this belongs inside of the `if is_custom_build()` branch as its explaining that branch.   The connotation for this location is its explaining the nightly check.  Inside of this `if` would be for explaining the overall logic.

```suggestion
        if is_multiple_build_scripts_enabled {
            list.sort_by_key(|unit_dep| {
                if unit_dep.unit.target.is_custom_build() {
                    // We do not sort build scripts to preserve the user-defined order.
                    //
                    // In terms of determinism, we are assuming nothing interferes with order from when the user set it in `Cargo.toml` to here
                    (0, None)
```
(also a small tweak to the determinism comment)

PR #15780, opened 2025-07-28
Add initial integration for `--json=timings`

Kobzol:

This PR adds initial support into Cargo for JSON timing sections, implemented in rustc in https://github.com/rust-lang/rust/pull/142123. This allows Cargo to read frontend/codegen/linking time from rustc, and thus reporting slightly more detailed data in the `cargo build --timings` output.

Currently this PR modifies Cargo to tell rustc to emit the section messages (`--json=...,timings`), and also renders the sections in the unit table of the unit timings table in the HTML output. As a follow-up, I also want to integrate the individual sections in the actual unit graph.

Note that the JSON timings are currently only supported on the nightly compiler (they are not stabilized). That means two things:
- Cargo has to somehow figure out if it should pass `--json=timings` to `rustc` or not. I don't think that it's worth it to add a new unstable Cargo flag just for this (ofc if you disagree, I'd be happy to implement that), instead I think that Cargo could just detect if the used rustc is nightly, and if yes, pass the flag always. I'm not sure how best to detect that though. So far I just used `gctx.nightly_features_allowed`, but I don't know if it's correct.
- The timings HTML page has to be prepared for the data to be both present and missing. I tried to do that with the `AggregatedSections` enum, and by modifying the table to skip the linking column when no data for it are present. A slight change from before is that both the Frontend and Codegen columns are now always present, while before the Frontend column value had to be "inferred" by the user (by computing `Total` - `Codegen`). I think that's an all-around improvement.

You can run e.g. this to generate the timing report with a nightly compiler:
```bash
export RUSTC=`rustup +nightly which rustc`
target/debug/cargo build --timings
```
on some crate, e.g. [ripgrep](https://github.com/BurntSushi/ripgrep).

I have also included example output with and without the section timings: [timings.zip](https://github.com/user-attachments/files/21472709/timings.zip).

### DIFF
--- a/src/cargo/core/compiler/mod.rs
+++ b/src/cargo/core/compiler/mod.rs
@@ -90,6 +90,7 @@ use self::output_depinfo::output_depinfo;
 use self::output_sbom::build_sbom;
 use self::unit_graph::UnitDep;
 use crate::core::compiler::future_incompat::FutureIncompatReport;
+use crate::core::compiler::timings::SectionTiming;
 pub use crate::core::compiler::unit::{Unit, UnitInterner};
 use crate::core::manifest::TargetSourcePath;
 use crate::core::profiles::{PanicStrategy, Profile, StripInner};
@@ -1095,6 +1096,12 @@ fn add_allow_features(build_runner: &BuildRunner<'_, '_>, cmd: &mut ProcessBuild
 ///
 /// [`--error-format`]: https://doc.rust-lang.org/nightly/rustc/command-line-arguments.html#--error-format-control-how-errors-are-produced
 fn add_error_format_and_color(build_runner: &BuildRunner<'_, '_>, cmd: &mut ProcessBuilder) {
+    let enable_timings = build_runner.bcx.gctx.nightly_features_allowed
+        && !build_runner.bcx.build_config.timing_outputs.is_empty();
+    if enable_timings {
+        cmd.arg("-Zunstable-options");
+    }
+
     cmd.arg("--error-format=json");
     let mut json = String::from("--json=diagnostic-rendered-ansi,artifacts,future-incompat");
 
@@ -1104,6 +1111,11 @@ fn add_error_format_and_color(build_runner: &BuildRunner<'_, '_>, cmd: &mut Proc
         }
         _ => {}
     }
+
+    if enable_timings {
+        json.push_str(",timings");
+    }
+
     cmd.arg(json);
 
     let gctx = build_runner.bcx.gctx;
@@ -1955,6 +1967,12 @@ fn on_stderr_line_inner(
         }
         state.future_incompat_report(report.future_incompat_report);
         return Ok(true);
+    }
+
+    let res = serde_json::from_str::<SectionTiming>(compiler_message.get());
+    if let Ok(timing_record) = res {
+        state.on_section_timing_emitted(timing_record);
+        return Ok(false);
     }
 
     // Depending on what we're emitting from Cargo itself, we figure out what to
--- a/src/cargo/core/compiler/timings.rs
+++ b/src/cargo/core/compiler/timings.rs
@@ -16,6 +16,7 @@ use std::collections::HashMap;
 use std::io::{BufWriter, Write};
 use std::thread::available_parallelism;
 use std::time::{Duration, Instant};
+use tracing::warn;
 
 /// Tracking information for the entire build.
 ///
@@ -63,6 +64,14 @@ pub struct Timings<'gctx> {
     cpu_usage: Vec<(f64, f64)>,
 }
 
+/// Section of compilation.
+struct TimingSection {
+    /// Start of the section, as an offset in seconds from `UnitTime::start`.
+    start: f64,
+    /// End of the section, as an offset in seconds from `UnitTime::start`.
+    end: Option<f64>,
+}
+
 /// Tracking information for an individual unit.
 struct UnitTime {
     unit: Unit,
@@ -79,6 +88,8 @@ struct UnitTime {
     unlocked_units: Vec<Unit>,
     /// Same as `unlocked_units`, but unlocked by rmeta.
     unlocked_rmeta_units: Vec<Unit>,
+    /// Individual compilation section durations, gathered from `--json=timings`.
+    sections: HashMap<String, TimingSection>,
 }
 
 /// Periodic concurrency tracking information.
@@ -181,6 +192,7 @@ impl<'gctx> Timings<'gctx> {
             rmeta_time: None,
             unlocked_units: Vec::new(),
             unlocked_rmeta_units: Vec::new(),
+            sections: Default::default(),
         };
         assert!(self.active.insert(id, unit_time).is_none());
     }
@@ -231,6 +243,26 @@ impl<'gctx> Timings<'gctx> {
             crate::drop_println!(self.gctx, "{}", msg);
         }
         self.unit_times.push(unit_time);
+    }
+
+    /// Handle the start/end of a compilation section.
+    pub fn unit_section_timing(&mut self, id: JobId, section_timing: &SectionTiming) {
+        if !self.enabled {
+            return;
+        }
+        let Some(unit_time) = self.active.get_mut(&id) else {
+            return;
+        };
+        let now = self.start.elapsed().as_secs_f64();
+
+        match section_timing.event {
+            SectionTimingEvent::Start => {
+                unit_time.start_section(&section_timing.name, now);
+            }
+            SectionTimingEvent::End => {
+                unit_time.end_section(&section_timing.name, now);
+            }
+        }
     }
 
     /// This is called periodically to mark the concurrency of internal structures.
@@ -578,6 +610,46 @@ impl UnitTime {
     fn name_ver(&self) -> String {
         format!("{} v{}", self.unit.pkg.name(), self.unit.pkg.version())
     }
+
+    fn start_section(&mut self, name: &str, now: f64) {
+        if self
+            .sections
+            .insert(
+                name.to_string(),
+                TimingSection {
+                    start: now - self.start,
+                    end: None,
+                },
+            )
+            .is_some()
+        {
+            warn!("Compilation section {name} started more than once");
+        }
+    }
+
+    fn end_section(&mut self, name: &str, now: f64) {
+        let Some(section) = self.sections.get_mut(name) else {
+            warn!("Compilation section {name} ended, but it has no start recorded");
+            return;
+        };
+        section.end = Some(now - self.start);
+    }
+}
+
+/// Start or end of a section timing.
+#[derive(serde::Deserialize, Debug)]
+#[serde(rename_all = "kebab-case")]
+pub enum SectionTimingEvent {
+    Start,
+    End,
+}
+
+/// Represents a certain section (phase) of rustc compilation.
+/// It is emitted by rustc when the `--json=timings` flag is used.
+#[derive(serde::Deserialize, Debug)]
+pub struct SectionTiming {
+    pub name: String,
+    pub event: SectionTimingEvent,
 }
 
 fn render_rustc_info(bcx: &BuildContext<'_, '_>) -> String {
--- a/src/cargo/core/compiler/timings.rs
+++ b/src/cargo/core/compiler/timings.rs
@@ -92,6 +92,50 @@ struct UnitTime {
     sections: HashMap<String, TimingSection>,
 }
 
+impl UnitTime {
+    fn aggregate_sections(&self) -> AggregatedSections {
+        let end = self.duration;
+
+        let codegen_section = self.sections.get("codegen");
+        let link_section = self.sections.get("link");
+
+        // Best case: we know all three sections
+        if let (Some(codegen_section), Some(link_section)) = (codegen_section, link_section) {
+            let link_start = link_section.start;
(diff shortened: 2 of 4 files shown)


src/cargo/core/compiler/timings.rs
```diff
@@ -578,6 +610,46 @@ impl UnitTime {
...
+            )
+            .is_some()
+        {
+            warn!("Compilation section {name} started more than once");
+        }
+    }
+
+    fn end_section(&mut self, name: &str, now: f64) {
+        let Some(section) = self.sections.get_mut(name) else {
+            warn!("Compilation section {name} ended, but it has no start recorded");
```
epage: nit: these are generally lower case

src/cargo/core/compiler/timings.rs
```diff
@@ -578,6 +610,46 @@ impl UnitTime {
...
+            .insert(
+                name.to_string(),
+                TimingSection {
+                    start: now - self.start,
+                    end: None,
+                },
+            )
+            .is_some()
+        {
+            warn!("Compilation section {name} started more than once");
```
epage: ditto

src/cargo/core/compiler/timings.rs
```diff
@@ -92,6 +92,50 @@ struct UnitTime {
...
+                    end: rmeta_time,
+                },
+                codegen: SectionData {
+                    start: rmeta_time,
+                    end,
+                },
+            }
+        } else {
+            AggregatedSections::OnlyTotalDuration
+        }
```
epage: Can we keep this generic over what sections are present?

src/cargo/core/compiler/mod.rs
```diff
@@ -1095,6 +1096,12 @@ fn add_allow_features(build_runner: &BuildRunner<'_, '_>, cmd: &mut ProcessBuild
 ///
 /// [`--error-format`]: https://doc.rust-lang.org/nightly/rustc/command-line-arguments.html#--error-format-control-how-errors-are-produced
 fn add_error_format_and_color(build_runner: &BuildRunner<'_, '_>, cmd: &mut ProcessBuilder) {
+    let enable_timings = build_runner.bcx.gctx.nightly_features_allowed
+        && !build_runner.bcx.build_config.timing_outputs.is_empty();
```
epage: > Cargo has to somehow figure out if it should pass --json=timings to rustc or not. I don't think that it's worth it to add a new unstable Cargo flag just for this (ofc if you disagree, I'd be happy to implement that), instead I think that Cargo could just detect if the used rustc is nightly, and if yes, pass the flag always. I'm not sure how best to detect that though. So far I just used gctx.nightly_features_allowed, but I don't know if it's correct.

src/cargo/core/compiler/timings.rs
```diff
@@ -549,6 +627,18 @@ impl<'gctx> Timings<'gctx> {
 
     /// Render the table of all units.
     fn write_unit_table(&self, f: &mut impl Write) -> CargoResult<()> {
```
epage, line 650: Should we include rustc's timing information in our unstable timings json output?

src/cargo/core/compiler/timings.rs
Kobzol: Well, that depends on in which direction the rustc feature evolves. In theory we could output anything from rustc and make it fully opaque for downstream tools (including cargo). But then we would probably have to use some hierarchical structure for the sections, which complicates things. And the current build timings output is not really prepared for visualizing such complex data, I think.

So for the first version, I want to only include the three known sections (frontend/backend/linker), as I think that provides the most bang for the buck (e.g. to hint to people that they might try a faster linker, or that the linker is not the bottleneck).

src/cargo/core/compiler/mod.rs
Kobzol: Did you mean to reply something here? :sweat_smile: It looks like CI/tests doesn't like using just `nightly_features_allowed`..

src/cargo/core/compiler/timings.rs
Kobzol: Good catch! Added it.

src/cargo/core/compiler/mod.rs
epage: I was trying to move your comment in the PR description to an issue on the PR so it has a proper place to be discussed and not be missed.

src/cargo/core/compiler/timings.rs
epage: Maybe I'm missing something but I feel like being opaque would be easier to stabilize. 

If we are wanting to be explicit about these names being a part of the format, should we deserialize into an `enum` instead of a `String`?  That also makes it so we don't bury string comparisons in the timings code.

src/cargo/core/compiler/timings.rs
```diff
@@ -64,8 +64,9 @@ pub struct Timings<'gctx> {
     cpu_usage: Vec<(f64, f64)>,
 }
 
-/// Section of compilation.
-struct TimingSection {
```
epage: Mind moving the rename into the commit that added it so this commit can be focused just on json output?

src/cargo/core/compiler/timings.rs
```diff
@@ -723,20 +725,20 @@ impl UnitTime {
...
         {
-            warn!("Compilation section {name} started more than once");
+            warn!("compilation section {name} started more than once");
         }
     }
 
     fn end_section(&mut self, name: &str, now: f64) {
         let Some(section) = self.sections.get_mut(name) else {
-            warn!("Compilation section {name} ended, but it has no start recorded");
+            warn!("compilation section {name} ended, but it has no start recorded");
```
epage, line 805: These also belong in the original commit

src/cargo/util/machine_message.rs
```diff
@@ -87,6 +88,12 @@ impl<'a> Message for BuildScript<'a> {
     }
 }
 
+#[derive(Serialize)]
+pub struct TimingInfoCompilationSection {
```
epage: Is this unused?

src/cargo/core/compiler/timings.rs
Kobzol: > Maybe I'm missing something but I feel like being opaque would be easier to stabilize.

That might very well be true, if we don't promise the specific contents, it might indeed be easier to stabilize this on the rustc side.

On the other hand, I'm not sure how well we can accommodate arbitrary sections in the timings page. Should we just gather whatever sections we find and then create a single column in the table for each section, in the same order as we saw them in the input? Before we kind of hardcoded frontend/backend from the rmeta time, so I was thinking that we would just extend that with the link time with the section timings.

epage: I'd be fine making the columns dynamic.  Its hard coded now because we have discrete, hard coded pieces of data that cargo collects.

Kobzol:
Ok, so based on feedback I'm fine with moving this to a fully dynamic approach.

Before I do that though, do you have any suggestions on how to detect whether the timings flag should be passed to Cargo? Is there any precedent for doing something in Cargo *only* on the nightly channel, or more precisely only with nightly rustc? Or would you prefer that we add an unstable Cargo flag for this?

epage:
> Before I do that though, do you have any suggestions on how to detect whether the timings flag should be passed to Cargo? Is there any precedent for doing something in Cargo only on the nightly channel, or more precisely only with nightly rustc? Or would you prefer that we add an unstable Cargo flag for this?

We discussed this in today's cargo team meeting.  We should not enable an unstable rustc feature implicitly but require an explicit opt-in.
- People using nightly for other functionality (e.g. miri or a sanitizer) shouldn't pick up unstable features they aren't using
- We'd want more than to assume that rustc outputs it in a compatible way / the flag exists and instead have feature detection which is difficult to get right.

Kobzol:
Fair enough, I'll add an explicit feature flag then. It's true that if it was implicit, if the rustc format changes in the future, it could break nightly cargo users of `cargo build --timings`. That would be solvable if Cargo was a subtree, but with a submodule it would be quite difficult to ensure that both rustc and cargo change at once.

Kobzol:
Ok, I rewrote the PR and updated the PR description. It is now gated behind an unstable `-Zsection-timings` flag, and it renders the sections into the HTML table in a fully dynamic manner.

src/cargo/core/compiler/timings.rs
```diff
@@ -88,8 +90,9 @@ struct UnitTime {
     unlocked_units: Vec<Unit>,
     /// Same as `unlocked_units`, but unlocked by rmeta.
     unlocked_rmeta_units: Vec<Unit>,
-    /// Individual compilation section durations, gathered from `--json=timings`.
-    sections: HashMap<String, TimingSection>,
+    /// IndexMap is used to keep original insertion order, we want to be able to tell which
+    /// sections were started in which order.
+    sections: IndexMap<String, CompilationSection>,
```
epage, line 98: Should we keep that original comment as well?

src/cargo/core/compiler/timings.rs
epage, line 1: This commit has the same name as the previous commit

src/cargo/core/compiler/timings.rs
```diff
@@ -90,11 +91,75 @@ struct UnitTime {
...
+
+impl UnitTime {
+    fn aggregate_sections(&self) -> AggregatedSections {
+        let end = self.duration;
+
+        if !self.sections.is_empty() {
+            // We have some detailed compilation section timings, so we postprocess them
+            // Since it is possible that we do not have an end timestamp for a given compilation
+            // section, we need to iterate them and if an end is missing, we assign the end of
+            // the section to the start of the following section.
```
epage, line 112: Are sections non-overlapping, allowing this to work?

src/cargo/core/compiler/timings.rs
```diff
@@ -480,6 +574,7 @@ impl<'gctx> Timings<'gctx> {
             .enumerate()
             .map(|(i, ut)| (ut.unit.clone(), i))
             .collect();
+
```
epage: Why is this newline here?

src/cargo/core/compiler/timings.rs
```diff
@@ -562,20 +709,53 @@ impl<'gctx> Timings<'gctx> {
       <th></th>
       <th>Unit</th>
       <th>Total</th>
-      <th>Codegen</th>
+      {headers}
```
epage, line 712: Would it work to split out a refactor commit that adjusts the existing Frontend/Backend logic to this and then add section timings on top?

src/cargo/core/compiler/timings.rs
Kobzol: What do you mean by original comment here? :thinking:

src/cargo/core/compiler/timings.rs
Kobzol: Currently, yes. If it changes in the future, we'd need to of course adapt to that.

src/cargo/core/compiler/timings.rs
Kobzol: Just stylistic I guess, to separate the variable from the item, it looked weird to me:
```rust
let unit_map: HashMap<Unit, usize> = self
    .unit_times
    .iter()
    .enumerate()
    .map(|(i, ut)| (ut.unit.clone(), i))
    .collect();
#[derive(serde::Serialize)]
struct UnitData {
```

src/cargo/core/compiler/timings.rs
Kobzol: Hmm, it's all very dependent on one another, but I tried to split it.

src/cargo/core/compiler/timings.rs
epage: 5a266b2009762b1635bb130ecdc09ca7d801943e changed the comment from
```rust
/// Individual compilation section durations, gathered from `--json=timings`.
```
to
```rust
    /// IndexMap is used to keep original insertion order, we want to be able to tell which
    /// sections were started in which order.
```

src/cargo/core/compiler/timings.rs
epage: Its unrelated to the rest of the commit though.  Please split stylistic changes out to separate commits

src/cargo/core/compiler/timings.rs
epage: Why was this marked as resolved?  I'm not seeing it resolved as requested nor an explanation for why it shouldn't happen

src/cargo/core/compiler/timings.rs
epage: Thanks, thats a big help

src/cargo/core/compiler/timings.rs
Kobzol: I fixed that in the latest push, the comment now contains both of these sentences.

src/cargo/core/compiler/timings.rs
Kobzol: I reverted the change.

src/cargo/core/compiler/timings.rs
epage: Could you please fix that within this commit?

Kobzol:
Restructured the commits

src/cargo/core/compiler/timings.rs
epage: Noted this in the tracking issue, #15817.

epage (approved):
Thanks!

PR #15781, opened 2025-07-28
More helpful error for invalid cargo-features = []

kornelski:
The error message for unknown values of `cargo-features` was unhelpful:

```text
Caused by:
  unknown cargo feature `build-dir`
```

I've made it more explanatory, and matching the style of the other errors:

```text
Caused by:
  feature `build-dir` is not a valid unstable feature for Cargo.toml

  This feature can be enabled via -Zbuild-dir or the `[unstable]` section in config.toml.
  See https://doc.rust-lang.org/cargo/reference/unstable.html for more information.
```

### DIFF
--- a/src/cargo/core/features.rs
+++ b/src/cargo/core/features.rs
@@ -620,7 +620,32 @@ impl Features {
     ) -> CargoResult<()> {
         let nightly_features_allowed = self.nightly_features_allowed;
         let Some((slot, feature)) = self.status(feature_name) else {
-            bail!("unknown cargo feature `{}`", feature_name)
+            let mut msg = format!("unknown Cargo.toml feature `{feature_name}`\n\n");
+            let mut append_see_docs = true;
+
+            if feature_name.contains('_') {
+                let _ = writeln!(msg, "Feature names must use '-' instead of '_'.");
+                append_see_docs = false;
+            } else {
+                let underscore_name = feature_name.replace('-', "_");
+                if CliUnstable::help()
+                    .iter()
+                    .any(|(option, _)| *option == underscore_name)
+                {
+                    let _ = writeln!(
+                        msg,
+                        "This feature can be enabled via -Z{feature_name} or the `[unstable]` section in config.toml."
+                    );
+                }
+            }
+
+            if append_see_docs {
+                let _ = writeln!(
+                    msg,
+                    "See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html for more information."
+                );
+            }
+            bail!(msg)
         };
 
         if *slot {
--- a/tests/testsuite/cargo_features.rs
+++ b/tests/testsuite/cargo_features.rs
@@ -173,7 +173,9 @@ fn unknown_feature() {
 [ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
-  unknown cargo feature `foo`
+  unknown Cargo.toml feature `foo`
+
+  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html for more information.
 
 "#]])
         .run();
@@ -202,7 +204,10 @@ fn wrong_kind_of_feature() {
 [ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
-  unknown cargo feature `build-dir`
+  unknown Cargo.toml feature `build-dir`
+
+  This feature can be enabled via -Zbuild-dir or the `[unstable]` section in config.toml.
+  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html for more information.
 
 "#]])
         .run();
@@ -231,7 +236,9 @@ fn feature_syntax() {
 [ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
-  unknown cargo feature `bad_feature`
+  unknown Cargo.toml feature `bad_feature`
+
+  Feature names must use '-' instead of '_'.
 
 "#]])
         .run();
--- a/tests/testsuite/cargo_features.rs
+++ b/tests/testsuite/cargo_features.rs
@@ -180,6 +180,64 @@ Caused by:
 }
 
 #[cargo_test]
+fn wrong_kind_of_feature() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                cargo-features = ["build-dir"]
+
+                [package]
+                name = "a"
+                version = "0.0.1"
+                edition = "2015"
+                authors = []
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+    p.cargo("check")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
+
+Caused by:
+  unknown cargo feature `build-dir`
+
+"#]])
+        .run();
+}
+
+#[cargo_test]
+fn feature_syntax() {
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+                cargo-features = ["bad_feature"]
+
+                [package]
+                name = "a"
+                version = "0.0.1"
+                edition = "2015"
+                authors = []
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .build();
+    p.cargo("check")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
+
+Caused by:
+  unknown cargo feature `bad_feature`
+
+"#]])
+        .run();
+}
+
+#[cargo_test]
 fn stable_feature_warns() {
     let p = project()
         .file(


src/cargo/core/features.rs
```diff
@@ -620,7 +620,30 @@ impl Features {
     ) -> CargoResult<()> {
         let nightly_features_allowed = self.nightly_features_allowed;
         let Some((slot, feature)) = self.status(feature_name) else {
-            bail!("unknown cargo feature `{}`", feature_name)
+            let mut msg = format!(
+                "feature `{feature_name}` is not a valid unstable feature for Cargo.toml\n\n"
```
epage: I prefer the old verbage (`"unknown cargo feature"`)
- Highlights cargo feature, rather than just feature
- It keeps the important information at the start, rather than burying it

src/cargo/core/features.rs
```diff
@@ -620,7 +620,30 @@ impl Features {
...
+                if CliUnstable::help()
+                    .iter()
+                    .any(|(option, _)| *option == underscore_name)
+                {
+                    let _ = writeln!(
+                        msg,
+                        "This feature can be enabled via -Z{feature_name} or the `[unstable]` section in config.toml."
+                    );
+                }
+            }
```
epage: No tests for these cases.  Please
- Make them end-to-end tests
- Add them in a commit before this one with them passing

src/cargo/core/features.rs
```diff
@@ -620,7 +620,30 @@ impl Features {
...
+
+            let _ = writeln!(
+                msg,
+                "See https://doc.rust-lang.org/cargo/reference/unstable.html for more information."
```
epage: Should we make this nightly?  That would be the most extensive version of this though we don't document when unstable features are introduced.  Anything other than `nightly` seems like a crap shoot.
```suggestion
                "See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html for more information."
```

epage:
Thanks for working to improve this!

src/cargo/core/features.rs
```diff
@@ -620,7 +620,30 @@ impl Features {
     ) -> CargoResult<()> {
         let nightly_features_allowed = self.nightly_features_allowed;
         let Some((slot, feature)) = self.status(feature_name) else {
-            bail!("unknown cargo feature `{}`", feature_name)
+            let mut msg = format!(
+                "feature `{feature_name}` is not a valid unstable feature for Cargo.toml\n\n"
+            );
+
+            if feature_name.contains('_') {
+                let _ = writeln!(msg, "Feature names must use '-' instead of '_'.");
```
epage, line 627: Should we take this a step further and
- See if the name with `-` exists?
- Use `closest_msg` to correct for typos? (this one might require more work to do)

src/cargo/core/features.rs
kornelski: I've changed it to "unknown Cargo.toml feature `foo`", because I think it's important to highlight that it's limited to a specific subset of features, not a general unstable feature mechanism.

src/cargo/core/features.rs
kornelski: I just want to improve the error message. I have low success rate with making Cargo changes, and such features would add even more work that may not get merged.

src/cargo/core/features.rs
kornelski: I've added tests.

epage: Could you please add them in the separate commit as asked

tests/testsuite/cargo_features.rs
```diff
@@ -173,7 +173,73 @@ fn unknown_feature() {
...
+    p.cargo("check")
+        .with_status(101)
+        .with_stderr_data(str![[r#"
+[ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
+
+Caused by:
+  unknown Cargo.toml feature `build-dir`
+
+  This feature can be enabled via -Zbuild-dir or the `[unstable]` section in config.toml.
+  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html for more information.
```
epage, line 210: Thanks for the tests as it makes it easier to do a UX review

In looking at this and the other case with additional information, I feel like the link is superfluous.  In this case, we told them exactly what they need, so they are unlikely to need the link.  I feel like its similar for the `-` / `_` case.  I guess I see the link is acting as a replacement for not enumerating every single feature.  I feel like we should make this 3 mutually exclusive cases (casing, feature name, link)

src/cargo/core/features.rs
epage: I'm mixed on this.  What we really need is to just highlight the section in the source code.  We're looking at doing that a lot more once we get the new annotate snippets version.  Likely good enough until then.

tests/testsuite/cargo_features.rs
kornelski: The link explains the difference between `cargo-features` and `[unstable]`. Providing the list of valid names is also useful.

This error easily fits on screen, so I don't see any downside to adding the line here.

src/cargo/core/features.rs
kornelski: I've implemented structured errors in Cargo recently, trying to clean up out-of-band calls to `shell.warn()` and `"help:"` tacked-on to `bail!`. 

However, I've tried doing that by reusing `annotate_snippets`, and it turned out to be a poor fit. The library doesn't seem to have enough flexibility to nicely format chains of "caused by" errors. It has hardcoded bold style for primary errors, which makes them jarringly different from Cargo's current style. In `rustc` this works, because errors are almost always mixed with code snippets, but few things in Cargo have any source code to point to. Types from `annotate_snippets` are [borrowing](https://docs.rs/annotate-snippets/latest/annotate_snippets/enum.Level.html#method.title), so due to self-referential types limitations they can't be used to store structured errors in the actual error type, and require creation of additional structs for constructing structured errors, which would either duplicate `annotate_snippets` API, or need a much more involved mechanism for semantic errors, like diagnostic derives in `rustc`.

epage: `annotate_snippets` is  not meant to carry diagnostics but to render them.

We have introduced it for some errors.  We'll likely continue the transition with using them for `warn`s and other manifest errors but I think its fully reasonable to work through the error chain issues, it just might take some time and experimentation.

tests/testsuite/cargo_features.rs
epage: That may work in this case but what about the `-` / `_` case?

As for whether things fit on the screen, that is a fairly low bar for UX.  Being focused and intentional is important for people being able to extract the relevant information.

kornelski: I agree with the substance of your feedback. However, it's unclear to me whether you treat it as a requirement for getting this PR merged.

If you're just enthusiastically looking for further improvements that's cool, and I appreciate your engagement and attention to detail. But if I have to make this change to get this merged, then it rubs me as "perfect is enemy of good" scenario. Sorry, I have trouble reading your intentions, it's not the first time, I don't know why.

I would prefer to get this merged as-is. Further improvements would be nice, but I think they approach diminishing returns, and the message is already better than what was there before.

I think I'd have more positive impact overall if I could make multiple incremental improvements quickly, rather than spend more of your and my time/energy on going back and forth on perfecting details of a single change. 

It's frustrating to me when the bar for landing an improvement is set much higher than the quality bar of the code being improved. There status quo wins by default, and it can feel unfair. If my version of the error message was already in Cargo, and somebody made a PR to delete it and replace it with `"unknown cargo feature"`, that would have been an obvious step backwards.

I see the irony in me complaining about this particular situation taking more time than applying the change you asked for, but I see it as an example of a pattern, and it hit me in my pre-existing frustration with contributing to Cargo. I feel like I'm capable of improving Cargo, and there are areas I think really need fixing, and which I believe I could fix (#15792), but the whole process around making these improvements feels very slow and inefficient to me, and it frustrates me.

src/cargo/core/features.rs
epage: Thank you for working to separate out the tests however, I think there was a misunderstanding.  We ask for each commit to be atomic which includes passing of tests.  Part of the intent of the test commit before the fix is to show what the old behavior is and make clear what the behavior change is.  This makes reviewing *a lot* easier and makes it easy to share these results with others.

If it helps, I'm hoping for what you did with #15124 to be applied here.

src/cargo/core/features.rs
epage: Fully understand doing things incrementally, isolating controversy, etc.  As I mentioned with at least the second, I wasn't even sure how trivial it was to do.  Even the first might not be trivial.  I brought this up in case it was of interest especially since I feel like we too easily forget to apply `closest_msg` to our error messages when it can be a big help to users.

tests/testsuite/cargo_features.rs
epage: Maybe I'm missing something but this reads more like a reply to https://github.com/rust-lang/cargo/pull/15781#discussion_r2237626228.  In that, I opened the question for further improvements.  It was not intending to be blocking.

As for this thread, I got confused at the intent / actionability of the new message which means users are more likely to run into that.  I fully support incremental improvements but if half part of it is a clear improvement and half doesn't feel like it, that makes it a less clear win than "merge and fix later".   One strategy we sometimes use for these cases is to just split up the PR for the part that is ready can be merged while the part that isn't can continue to be iterated on.

> I see the irony in me complaining about this particular situation taking more time than applying the change you asked for, but I see it as an example of a pattern, and it hit me in my pre-existing frustration with contributing to Cargo. 

I fully get it.  Working to address bigger picture problems is better overall for the long term efforts.  For example, this is why we give some of the feedback we do.  Some things we ask for aren't always needed (issues first, splitting out test commits, focused commits, etc) but encouraging people to use them, especially repeat contributors, improves the experience for everyone in the long run.  These are about making the process more efficient in the long term, not less, especially for time-limited reviewers.

Please also remember that some changes are reversible and I or another reviewer can merge, like this or many of the other error improvements you have made.  Some changes though, like the build script work you pointed to, are either not reversible or only partially reversible with a lot of work.  A contributor or reviewer is likely insufficient to have a full enough perspective (or lack of bias) to identify concerns and we need multiple perspectives on these to help vet them which can be as small as the team sign off with an FCP or wider like an Issue or Internals  Even then, it doesn't mean it has to be slow.  With a case like [solving build script directive extensibility](https://github.com/rust-lang/cargo/issues/11461), we quickly came to a decision and the path for implementation and stabilization was relatively smooth.

src/cargo/core/features.rs
kornelski: What confused me was that the old behavior was already fully covered by tests, since it was so basic. So the separate test commit ends up being just copypaste of the same test three times.

tests/testsuite/cargo_features.rs
kornelski: Thank you for your explanation and patience here. Your points about bigger changes and overall strategy make a lot of sense and I think we're on the same page there.

My comment was specifically about your **blocking** objection to the link after "Feature names must use '-' instead of '_'." We have a different value judgement here. While I agree that a tangentially-related link lowers the signal-to-noise ratio of this error, I don't think the confusion it may cause is nearly as bad as the confusion that the existing message causes.

I've made this PR, because I was literally confused by the "unknown cargo feature" error message. It prevented me from testing the `build-dir` feature properly (BTW, I'm very grateful for your handling of it, it's great). I've had an invalid assumption that all Cargo features are in the same namespace, so when `cargo-features` told me it's "unknown cargo feature", I interpreted it as saying that `build-dir` is not a Cargo feature *at all*, and it won't work in any configuration file, and therefore there's nothing else other than `-Zbuild-dir` CLI arg (because these are called options, not features). The lack of information in this error message has delayed my work on integrating it with `cargo-deb`, `cargo-xcode`, `cargo-show-asm`, and discovery of a bootstrap incompatibility.

So I just don't agree that the level of confusion an extra link may cause outweighs the damage that the current message caused, and from my perspective your objection was a nitpick threw away a genuine improvement.

tests/testsuite/cargo_features.rs
```diff
@@ -173,7 +173,67 @@ fn unknown_feature() {
 [ERROR] failed to parse manifest at `[ROOT]/foo/Cargo.toml`
 
 Caused by:
-  unknown cargo feature `foo`
+  unknown Cargo.toml feature `foo`
+
+  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html for more information.
```
epage, line 178: This looks like it belongs to the follow up commit

src/cargo/core/features.rs
epage: Yes, test commits on their own do not always provide value.  Its the diff between it and the follow up commit, along with the tests moving forward where the value is.

epage (approved):
Thanks!

PR #2912, opened 2016-07-24
Move `target_dir` to Workspace and fix #2848

matklad:
Target dir is now an API of Workspace. It is still initialized eagerly, just later. 

I also had to errors in the config file when retrieving `verbose` and `color` config. Seems fishy, but is probably OK.

PR #15783, opened 2025-07-29
perf(package): Always reuse the workspace's target-dir

ranger-ross:

While preparing to stabilize `build-dir`, [it was discovered](https://github.com/rust-lang/cargo/issues/14125#issuecomment-3127467689) that `cargo package` will sometimes use the build cache for dependency crates if `CARGO_TARGET_DIR` is explicitly set.

If target-dir is not set, the build will be preformed in an "inner target dir" at `target/package/{name}-{version}/target` and rebuild everything.

This PR changes the default behavior to always use the build cache,
matching the behavior of having `target-dir` set.

Added a dedicated test to verify the previous behavior (cargo recompiled non-workspace crates) in the first commit.
Then updated it to verify the new behavior.

### Related links/notes

* I did some Git archaeology and I _think_ this behavior was introduced in #2912
  * While searching through the commit/pr history was never able to find any explicit reason why the behavior is the way it is.
  * Even if it was intentional, it has been 9 years since and Cargo's fingerprinting/rebuild detection is much more mature now so I believe its worth revisiting if this is still the desired behavior.
* Would help mitigate https://github.com/rust-lang/cargo/issues/14941

### DIFF
--- a/src/cargo/ops/cargo_package/verify.rs
+++ b/src/cargo/ops/cargo_package/verify.rs
@@ -66,10 +66,13 @@ pub fn run_verify(
     let new_pkg = src.root_package()?;
     let pkg_fingerprint = hash_all(&dst)?;
 
+    // When packaging we use an ephemeral workspace but reuse the build cache to reduce
+    // verification time if the user has already compiled the dependencies and the fingerprint
+    // is unchanged.
     let target_dir = if gctx.cli_unstable().build_dir {
         Some(ws.build_dir())
     } else {
-        None
+        Some(ws.target_dir())
     };
 
     let mut ws = Workspace::ephemeral(new_pkg, gctx, target_dir, true)?;
--- a/tests/testsuite/package.rs
+++ b/tests/testsuite/package.rs
@@ -1828,7 +1828,7 @@ fn package_two_kinds_of_deps() {
 }
 
 #[cargo_test]
-fn package_should_not_use_build_cache() {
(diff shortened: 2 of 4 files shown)


ranger-ross:
Linking to the [Cargo meeting discussion](https://hackmd.io/@rust-cargo-team/H1iucap35#2025-07-29) about this PR.

It appears no consensus was made as many members were out that week.

ehuss (approved):
I think I'm fine with making this change. I don't recall any reasoning for keeping it separate, but I suspect if it was intentional that the intent was to avoid bad fingerprinting from properly validating the package. However, I'm comfortable now with proper fingerprinting from catching anything that needs to be rebuilt.

@epage or @weihanglo Do you have any other context on this? Otherwise I think we should merge.

epage:
Build scripts and proc macros can mess with things.

However, if the packaged packages (which is what we care about verifying) are getting a unique package id (which it should) it should then have a unique fingerprint, avoiding those problems.

Unless people are digging around outside of whats fingerprinted. I assume the current-dir is the package root and I'm assuming that is in the packaging directory, so that interferes with mucking around. I don't think we officially expose target-dir but people try to approximate it through what `cargo metadata` returns (dir for config lookup will be different). Unsure what problems people will have, if any, and if its worth holding this up.

epage:
We discussed this in today's meeting.  There were no objections and we can always revert this later.

PR #15788, opened 2025-07-30
fix(bindeps): do not propagate artifact dependency to proc macro or build deps

Lencerf:

This PR fixes a bug for the nightly feature `bindeps`. Basically, if

- a package `foo` has an artifact dependency `artifact` with a specified target TARGET_ARTIFACT different from host TARGET_HOST,
- `artifact` depends on proc macro `macro`,
- `macro` conditionally depends on `arch` on TARGET_HOST,

cargo build would panic on TARGET_HOST with the following error message:

```
did not find features for (PackageId { name: "arch", version: "0.0.1", source: "/Users/lencerf/Developer/cargo/target/tmp/cit/t0/foo/arch" }, ArtifactDep(CompileTarget { name: "x86_64-apple-darwin" })) within activated_features:
[
    (
        PackageId {
            name: "foo",
            version: "0.0.1",
            source: "/Users/lencerf/Developer/cargo/target/tmp/cit/t0/foo",
        },
        NormalOrDev,
    ),
    (
        PackageId {
            name: "macro",
            version: "0.0.1",
            source: "/Users/lencerf/Developer/cargo/target/tmp/cit/t0/foo/macro",
        },
        ArtifactDep(
            CompileTarget {
                name: "x86_64-apple-darwin",
            },
        ),
    ),
    (
        PackageId {
            name: "artifact",
            version: "0.0.1",
            source: "/Users/lencerf/Developer/cargo/target/tmp/cit/t0/foo/artifact",
        },
        ArtifactDep(
            CompileTarget {
                name: "x86_64-apple-darwin",
            },
        ),
    ),
]
```

Similar panic happens when

- a package `foo` has an artifact dependency `artifact` with a specified target TARGET_ARTIFACT different from host TARGET_HOST,
- `artifact` has a build dependency `builder`,
- `builder` conditionally depends on `arch` on TARGET_HOST.

From the above message, it is clear that proc macro `macro` was wrongly associated with `FeaturesFor::ArtifactDep` instead of `FeaturesFor::HostDep`. Package `arch` was later ignored because cargo thought `macro` should be built for TARGET_ARTIFACT (x86_64-apple-darwin), while `arch` was conditionally needed on TARGET_HOST (aarch64-apple-darwin).

Similar analyses apply to the other test case.

This commit fixes 2 paths:
- when resolving features, if we encounter build dependencies or proc  macros, always associate them with `FeaturesFor::HostDep`.
- when deriving UnitFor for dependencies, stop propagating artifact_target_for_features if the the dependency is a build dependency or a proc macro.

This PR contains 2 commits

- the first commit adds 2 test cases to reproduce the issue above. Since they cannot pass now, they are marked with `#[should_panic]`
- the second commit fixes the bug and removes the `#[should_panic]` from the 2 test cases.

`cargo test` can pass on each commit.

### DIFF
--- a/src/cargo/core/profiles.rs
+++ b/src/cargo/core/profiles.rs
@@ -1177,12 +1177,18 @@ impl UnitFor {
         } else {
             self.panic_setting
         };
+        let artifact_target_for_features =
+            if dep_target.proc_macro() || parent.target.is_custom_build() {
+                None
+            } else {
+                self.artifact_target_for_features
+            };
         UnitFor {
             host: self.host || dep_for_host,
             host_features,
             panic_setting,
             root_compile_kind,
-            artifact_target_for_features: self.artifact_target_for_features,
+            artifact_target_for_features,
         }
     }
 
--- a/src/cargo/core/resolver/features.rs
+++ b/src/cargo/core/resolver/features.rs
@@ -906,11 +906,11 @@ impl<'a, 'gctx> FeatureResolver<'a, 'gctx> {
                         // All this may result in a dependency being built multiple times
                         // for various targets which are either specified in the manifest
                         // or on the cargo command-line.
-                        let lib_fk = if fk == FeaturesFor::default() {
-                            (self.track_for_host
-                                && (dep.is_build() || self.has_proc_macro_lib(dep_id)))
-                            .then(|| FeaturesFor::HostDep)
-                            .unwrap_or_default()
+                        let lib_fk = if fk != FeaturesFor::HostDep
+                            && self.track_for_host
+                            && (dep.is_build() || self.has_proc_macro_lib(dep_id))
+                        {
+                            FeaturesFor::HostDep
                         } else {
                             fk
                         };
--- a/tests/testsuite/artifact_dep.rs
+++ b/tests/testsuite/artifact_dep.rs
@@ -3394,3 +3394,181 @@ staticlib present: true
 "#]],
     );
 }
+
+#[should_panic]
+#[cargo_test]
+fn artifact_dep_target_does_not_propagate_to_deps_of_build_script() {
+    if cross_compile_disabled() {
+        return;
+    }
+    let bindeps_target = cross_compile::alternate();
+    let native_target = cross_compile::native();
+
+    let p = project()
+        .file(
+            "Cargo.toml",
+            &r#"
+            [package]
+            name = "foo"
+            version = "0.0.1"
+            edition = "2015"
+            resolver = "2"
+
+            [dependencies.artifact]
+            path = "artifact"
+            artifact = "bin"
+            target = "$TARGET"
+        "#
+            .replace("$TARGET", bindeps_target),
+        )
+        .file(
+            "src/main.rs",
+            r#"
+            fn main() {
+                let _b = include_bytes!(env!("CARGO_BIN_FILE_ARTIFACT"));
+            }
+        "#,
+        )
+        .file(
+            "artifact/Cargo.toml",
+            r#"
+            [package]
+            name = "artifact"
+            version = "0.0.1"
+            edition = "2015"
+
+            [build-dependencies]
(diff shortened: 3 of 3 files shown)


Lencerf:
This PR should fix #12358 too.

Lencerf:
Hi @weihanglo , it has been a few months, do you have any feedbacks on this PR? Thank you!

weihanglo:
Oops. It completely slipped off my radar 😅. I'll find some time to review it this week. Sorry for the huge delay

src/cargo/core/resolver/features.rs
```diff
@@ -906,11 +906,11 @@ impl<'a, 'gctx> FeatureResolver<'a, 'gctx> {
                         // All this may result in a dependency being built multiple times
                         // for various targets which are either specified in the manifest
                         // or on the cargo command-line.
-                        let lib_fk = if fk == FeaturesFor::default() {
-                            (self.track_for_host
-                                && (dep.is_build() || self.has_proc_macro_lib(dep_id)))
-                            .then(|| FeaturesFor::HostDep)
-                            .unwrap_or_default()
+                        let lib_fk = if fk != FeaturesFor::HostDep
```
weihanglo, line 909: Nice catch!

src/cargo/core/profiles.rs
```diff
@@ -1177,12 +1177,18 @@ impl UnitFor {
         } else {
             self.panic_setting
         };
+        let artifact_target_for_features =
```
weihanglo, line 1183: Could we add a simple comment here explaining why we clearing the marker here? Like build.rs itself (not build-deps) and proc macros are always for host.

tests/testsuite/artifact_dep.rs
```diff
@@ -3394,3 +3394,179 @@ staticlib present: true
...
+            version = "0.0.1"
+            edition = "2015"
+            "#,
+        )
+        .file(
+            "arch/src/lib.rs",
+            r#"pub fn add(a: i32, b: i32) -> i32 { a + b }"#,
+        )
+        .build();
+    p.cargo("test -Z bindeps")
```
weihanglo, line 3487: Mind adding `with_stderr_data(str![[""]])` to ensure that we really build `arch`? Ditto for proc macro test.

weihanglo (commented):
Thanks. Looks pretty good to me. And sorry for the extremely long delay 🙇🏾‍♂️.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/cargo/15788/37eeab7c165aaccc7c70dfdf2bbd239cbd2b7269..b54633809b04ee5c4ce8bbf7cc880639f72dccd7)*

tests/testsuite/artifact_dep.rs
weihanglo: you'll need to set `env SNAPSHOTS=overwrite` to regenerate test snapshot locally and commit it btw

Lencerf: Done

src/cargo/core/profiles.rs
Lencerf: Done

tests/testsuite/artifact_dep.rs
weihanglo: Thanks, @Lencerf!

https://github.com/rust-lang/cargo/actions/runs/19524318318/job/55893849063?pr=15788#step:15:4772

I know it is annoying, but we also need to append `[EXE]` for binary paths for windows compatibility

```diff
- [RUNNING] unittests src/main.rs (target/debug/deps/foo-[HASH])
+ [RUNNING] unittests src/main.rs (target/debug/deps/foo-[HASH][EXE])
```

Lencerf: Thanks for the feedback! CI now passes!

PR #15765, opened 2025-07-22
Non actionable error message

ojuschugh1:
### What does this PR try to resolve?

This PR improves the error reporting experience when running `cargo publish` on a workspace. Previously, if publishing failed partway through (for example, due to rate limiting or a server error), Cargo would emit a non-actionable error message that did not indicate which package failed or which packages were still pending publication. This made it difficult for users to recover from errors or know what steps to take next.

This PR addresses [issue #15754](https://github.com/rust-lang/cargo/issues/15754) by:

- Clearly indicating **which specific package failed to publish**.
- Listing **which packages in the workspace have not yet been published**.
- Providing **actionable, user-friendly error messages** to help users understand what went wrong and what remains to be done.

### How to test and review this PR?

**Testing:**

- New and updated tests are included in `tests/testsuite/publish.rs`:
  - The tests simulate workspace publishing scenarios where errors occur (such as rate limiting or server errors).
  - The tests verify that error messages now include:
    - The name of the package that failed to publish.
    - A list of remaining packages that still need to be published.
    - Clear guidance on what happened.
- To run the tests:
  ```sh
  cargo test workspace_publish_error_reporting
  cargo test workspace_transmit_error_includes_remaining_packages
  cargo test transmit_error_includes_package_name
  ```
- All existing tests continue to pass, ensuring no regressions.

**Review Guidance:**

- Review the changes in `src/cargo/ops/registry/publish.rs` for error context enhancements and message formatting.
- Review the new and updated tests in `tests/testsuite/publish.rs` to see how error scenarios are simulated and validated.
- Try running `cargo publish` on a workspace with multiple packages and simulate a failure (e.g., by using a test registry or causing a network error) to see the improved error output in action.

**Example improved error message:**
```
error: failed to publish to registry at https://crates.io
Package: my-package

Caused by:
  the remote server responded with an error (status 429 Too Many Requests): You have published too many new crates in a short period of time. Please try again after Fri, 18 Jul 2025 20:00:34 GMT or email help@crates.io to have your limit increased.

Remaining packages to publish: other-package-1, other-package-2, other-package-3
```

---

**Summary:**  
This PR makes workspace publishing failures much more actionable and user-friendly, helping users quickly identify what went wrong and what steps to take next. All changes are covered by tests and follow Cargo’s contribution guidelines.

Issue #15754, opened 2025-07-18
Non-actionable error message when publish fails in the middle of a worskpace


epage:
### Problem

I was doing a first-time publish on a workspace and was rate limited.  The only output I got was
```
error: failed to publish to registry at https://crates.io

Caused by:
  the remote server responded with an error (status 429 Too Many Requests): You have published too many new crates in a short period of time. Please try again after Fri, 18 Jul 2025 20:00:34
 GMT or email help@crates.io to have your limit increased.
```
What I don't know is
- which package failed to publish
- what packages have not been published yet

### Steps

1. Create a large workspace with brand new packages
2. Publish all of them at once and hit the rate limit

### Possible Solution(s)

Report which package failed in the message and maybe what packages that have still not been processed.

### Version

```text
cargo 1.90.0-nightly (930b4f62c 2025-06-28)
release: 1.90.0-nightly
commit-hash: 930b4f62cfcd1f0eabdb30a56d91bf6844b739bf
commit-date: 2025-06-28
host: x86_64-unknown-linux-gnu
libgit2: 1.9.0 (sys:0.20.2 vendored)
libcurl: 8.12.1-DEV (sys:0.4.80+curl-8.12.1 vendored ssl:OpenSSL/3.5.0)
ssl: OpenSSL 3.5.0 8 Apr 2025
os: Pop!_OS 22.4.0 (jammy) [64-bit]
```

epage:
Fixed in #15879.

PR #15791, opened 2025-07-30
 Non actionable error message #15765 

ojuschugh1:

This PR fixes [issue #15754](https://github.com/rust-lang/cargo/issues/15754) by improving error reporting for workspace publish failures.  
Previously, when publishing a workspace, if one package failed (e.g., due to an existing version), Cargo did not clearly list which packages failed and which remained unpublished.  
This change ensures that the error message now includes both the failed package and any remaining unpublished packages, making it easier for users to understand what went wrong and what still needs to be published.

---

- A new test (`workspace_publish_failure_reports_remaining_packages`) was added to publish.rs to demonstrate the previous behavior and verify the fix.
- The test first fails with the old behavior, then passes after the fix is applied.
- All commits are split for clarity: the test is added first, then the implementation is updated.
- To review:
  1. Check that the new test covers the relevant error scenario.
  2. Review the implementation changes in publish.rs for error context propagation.
  3. Confirm that all tests pass (`SNAPSHOTS=overwrite cargo test`).
  4. Review commit-by-commit for a clear history.

---

**How to test locally:**
```bash
SNAPSHOTS=overwrite cargo test --test testsuite publish::workspace_publish_failure_reports_remaining_packages
```
or simply
```bash
SNAPSHOTS=overwrite cargo test
```
to run the full suite.

---


ojuschugh1:
Hi @epage , tagging you for your kind review :) 

Thanks !!

epage:
What is the difference between this PR and #15765?

And if you are looking for a review, why is it marked as a draft?

tests/testsuite/publish.rs
epage, line 1: You should be able to run `cargo test` on this first commit and it should pass.  This is a part of what it means to have atomic commits.

So the only change that should exist in this commit is the adding of `workspace_publish_failure_reports_remaining_packages` and it should report the existing output on `master`, before your changes

tests/testsuite/publish.rs
```diff
@@ -2309,7 +2315,10 @@ fn api_error_200() {
 [PACKAGING] foo v0.0.1 ([ROOT]/foo)
 [PACKAGED] 4 files, [FILE_SIZE]B ([FILE_SIZE]B compressed)
 [UPLOADING] foo v0.0.1 ([ROOT]/foo)
-[ERROR] failed to publish to registry at http://127.0.0.1:[..]/
+[ERROR] failed to publish `foo` v0.0.1
+
+Caused by:
+  failed to publish to registry at http://127.0.0.1:37651/
```
epage: Note that the original error message had the port number redacted with `[..]`.  This avoids per-run information being embedded in the test output

epage: Any of this should be squashed into the relevant commit

src/cargo/ops/registry/publish.rs
```diff
@@ -160,7 +160,27 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
...
+                            "failed to publish `{}` v{}; the following crates have not been published yet: {}",
+                            pkg.name(),
+                            pkg.version(),
+                            remaining.join(", ")
+                        )
+                    } else {
+                        format!("failed to publish `{}` v{}", pkg.name(), pkg.version())
+                    };
+                    e.context(message)
+                })?;
```
epage: Do we need to extend the verify step with this information?

Part of the problem is that we can end up in a partially-published state when doing `transmit`.  At this point in the code, nothing is published

tests/testsuite/publish.rs
```diff
@@ -2462,7 +2477,10 @@ fn api_other_error() {
 [PACKAGING] foo v0.0.1 ([ROOT]/foo)
 [PACKAGED] 4 files, [FILE_SIZE]B ([FILE_SIZE]B compressed)
 [UPLOADING] foo v0.0.1 ([ROOT]/foo)
-[ERROR] failed to publish to registry at http://127.0.0.1:[..]/
+[ERROR] failed to publish `foo` v0.0.1
+
+Caused by:
+  failed to publish to registry at http://127.0.0.1:46397/
```
epage: Why are we reporting two `failed to publish` messages?  Shouldn't we just put all of this information in the one `failed to publish` message?

PR #15799, opened 2025-08-01
Fix Non-actionable error messages

ojuschugh1:

This PR fixes [issue #15754](https://github.com/rust-lang/cargo/issues/15754) by improving error reporting for workspace publish failures.  
Previously, when publishing a workspace, if one package failed (e.g., due to an existing version), Cargo did not clearly list which packages failed and which remained unpublished.  
This change ensures that the error message now includes both the failed package and any remaining unpublished packages, making it easier for users to understand what went wrong and what still needs to be published.

---

- A new test (`workspace_publish_failure_reports_remaining_packages`) was added to publish.rs to demonstrate the previous behavior and verify the fix.
- The test first fails with the old behavior, then passes after the fix is applied.
- All commits are split for clarity: the test is added first, then the implementation is updated.
- To review:
  1. Check that the new test covers the relevant error scenario.
  2. Review the implementation changes in publish.rs for error context propagation.
  3. Confirm that all tests pass (`SNAPSHOTS=overwrite cargo test`).
  4. Review commit-by-commit for a clear history.

---

**How to test locally:**
```bash
SNAPSHOTS=overwrite cargo test --test testsuite publish::workspace_publish_failure_reports_remaining_packages
```
or simply
```bash
SNAPSHOTS=overwrite cargo test
```
or 
```bash
cargo test
```
to run the full suite.

---

### DIFF
--- a/src/cargo/ops/registry/publish.rs
+++ b/src/cargo/ops/registry/publish.rs
@@ -238,67 +238,47 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
                 )?));
             }
 
-            let transmit_result = if original_packages.len() > 1 {
-                // For workspace publishes, wrap transmit with enhanced error reporting
-                transmit(
-                    opts.gctx,
-                    ws,
-                    pkg,
-                    tarball.file(),
-                    &mut registry,
-                    source_ids.original,
-                    opts.dry_run,
-                ).map_err(|e| {
-                    // Collect remaining packages that have not been published yet
-                    let mut remaining: Vec<_> = original_packages
-                        .iter()
-                        .filter(|id| **id != pkg_id)
-                        .map(|id| {
-                            let pkg = &pkg_dep_graph.packages[&id].0;
-                            format!("{} v{}", pkg.name(), pkg.version())
-                        })
-                        .collect();
-                    // Also include any packages that are still waiting for confirmation
-                    for id in to_confirm.iter().filter(|id| **id != pkg_id) {
+            // Prepare workspace context for error message
+            let workspace_context = if original_packages.len() > 1 {
+                let mut remaining: Vec<_> = original_packages
+                    .iter()
+                    .filter(|id| **id != pkg_id)
+                    .map(|id| {
                         let pkg = &pkg_dep_graph.packages[&id].0;
-                        let entry = format!("{} v{}", pkg.name(), pkg.version());
-                        if !remaining.contains(&entry) {
-                            remaining.push(entry);
-                        }
+                        format!("{} v{}", pkg.name(), pkg.version())
+                    })
+                    .collect();
+                // Also include any packages that are still waiting for confirmation
+                for id in to_confirm.iter().filter(|id| **id != pkg_id) {
+                    let pkg = &pkg_dep_graph.packages[&id].0;
+                    let entry = format!("{} v{}", pkg.name(), pkg.version());
+                    if !remaining.contains(&entry) {
+                        remaining.push(entry);
                     }
+                }
 
-                    let message = if !remaining.is_empty() {
-                        format!(
-                            "failed to publish `{}` v{}; the following crates have not been published yet: {}",
-                            pkg.name(),
-                            pkg.version(),
-                            remaining.join(", ")
-                        )
-                    } else {
-                        format!("failed to publish `{}` v{}", pkg.name(), pkg.version())
-                    };
-
-                    e.context(message)
-                })
-            } else {
-                // For single package publishes, preserve original top-level error message with package name
-                transmit(
-                    opts.gctx,
-                    ws,
-                    pkg,
-                    tarball.file(),
-                    &mut registry,
-                    source_ids.original,
-                    opts.dry_run,
-                )
-                .map_err(|e| {
-                    e.context(format!(
-                        "failed to publish `{}` v{}",
-                        pkg.name(),
-                        pkg.version()
+                if !remaining.is_empty() {
+                    Some(format!(
+                        "the following crates have not been published yet: {}",
+                        remaining.join(", ")
                     ))
-                })
+                } else {
+                    None
+                }
+            } else {
+                None
             };
+
+            let transmit_result = transmit(
+                opts.gctx,
+                ws,
+                pkg,
+                tarball.file(),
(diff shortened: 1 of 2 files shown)


ojuschugh1:
> ### What does this PR try to resolve?
> 
> This PR fixes [issue #15754](https://github.com/rust-lang/cargo/issues/15754) by improving error reporting for workspace publish failures. Previously, when publishing a workspace, if one package failed (e.g., due to an existing version), Cargo did not clearly list which packages failed and which remained unpublished. This change ensures that the error message now includes both the failed package and any remaining unpublished packages, making it easier for users to understand what went wrong and what still needs to be published.
> ### How to test and review this PR?
> 
>     * A new test (`workspace_publish_failure_reports_remaining_packages`) was added to publish.rs to demonstrate the previous behavior and verify the fix.
> 
>     * The test first fails with the old behavior, then passes after the fix is applied.
> 
>     * All commits are split for clarity: the test is added first, then the implementation is updated.
> 
>     * To review:
>       
>       1. Check that the new test covers the relevant error scenario.
>       2. Review the implementation changes in publish.rs for error context propagation.
>       3. Confirm that all tests pass (`SNAPSHOTS=overwrite cargo test`).
>       4. Review commit-by-commit for a clear history.
> 
> 
> **How to test locally:**
> 
> ```shell
> SNAPSHOTS=overwrite cargo test --test testsuite publish::workspace_publish_failure_reports_remaining_packages
> ```
> 
> or simply
> 
> ```shell
> SNAPSHOTS=overwrite cargo test
> ```
> 
> or
> 
> ```shell
> cargo test
> ```
> 
> to run the full suite.

This PR also solves the feedback received on my old fix -: https://github.com/rust-lang/cargo/pull/15791

New Changes include-: 1) All commits are atomic and passed cargo test.
2) Solved the Port Number issue by falling back to  [..].
3) Removed verify step runs BEFORE any packages are published since nothing is published at that point so no need for it.

tests/testsuite/cargo_add/git_branch/mod.rs
```diff
@@ -21,7 +21,7 @@ fn case() {
             .file("src/lib.rs", "")
     });
     let branch = "dev";
-    let find_head = || (git_repo.head().unwrap().peel_to_commit().unwrap());
+    let find_head = || git_repo.head().unwrap().peel_to_commit().unwrap();
```
epage: The commit title doesn't seem related to the diff

tests/testsuite/publish.rs
epage, line 1: Same feedback as https://github.com/rust-lang/cargo/pull/15791#discussion_r2244056531

This commit should be able to pass as-is but it can't because its including test changes for the following commit

src/cargo/ops/registry/publish.rs
epage, line 1: > refactor: clean up publish error context and remove unused variables 

Note that a "refactor" means a change to the code with no user-visible impact. If you wish to use Conventional Commits, this would be better term a `fix`

Also, the commit title talks about cleanup and removing a variable but the commit is about adding extra context to error messages.

tests/testsuite/publish.rs
```diff
@@ -2357,7 +2363,10 @@ fn api_error_code() {
 [PACKAGING] foo v0.0.1 ([ROOT]/foo)
 [PACKAGED] 4 files, [FILE_SIZE]B ([FILE_SIZE]B compressed)
 [UPLOADING] foo v0.0.1 ([ROOT]/foo)
-[ERROR] failed to publish to registry at http://127.0.0.1:[..]/
+[ERROR] failed to publish `foo` v0.0.1
+
+Caused by:
+  failed to publish to registry at http://127.0.0.1:[..]/
```
epage: https://github.com/rust-lang/cargo/pull/15791#discussion_r2244064132 is not resolved

> Why are we reporting two `failed to publish` messages? Shouldn't we just put all of this information in the one `failed to publish` message?

tests/testsuite/publish.rs
```diff
@@ -2462,7 +2474,10 @@ fn api_other_error() {
...
 
 Caused by:
```
ojuschugh1, line 2467: Hi @epage  so regarding this https://github.com/rust-lang/cargo/pull/15799/files#r2252499850 , I found out that already in some places like here there are more than one failed to publish messages so i thought to keep like this.

tests/testsuite/publish.rs
epage: From @ojuschugh1 at https://github.com/rust-lang/cargo/pull/15799/files#r2252608902

> Hi @epage so regarding this https://github.com/rust-lang/cargo/pull/15799/files#r2252499850 , I found out that already in some places like here there are more than one failed to publish messages so i thought to keep like this.

epage: Could you clarify what you mean?

ojuschugh1: Hi , @epage  We show two lines because Cargo prints the error chain: the top-level is a workspace-specific summary (which crate failed + what’s left), and the “Caused by” line is the transport/server detail. This keeps it readable and preserves single-crate behavior. If you prefer one line, I can collapse it and update tests.

epage: Can't we extend the original error to include the context?  Seems like that would provide much better results.

ojuschugh1: Sure, I will try to do this.

weihanglo:
Could you edit the title to make it clearer what this is about at first glance?
(Yeah maintainer can surely do that though, just being nice if you want to edit your stuff)

ojuschugh1:
> Could you edit the title to make it clearer what this is about at first glance? (Yeah maintainer can surely do that though, just being nice if you want to edit your stuff)

Sure no issues

tests/testsuite/publish.rs
ojuschugh1: @epage , I have fixed this. Kindly please review and let me know, if anything else you want me to do.

src/cargo/ops/registry/publish.rs
ojuschugh1: I will fix this commit title , first i wanted to complete the implementation in this pr, I will open a new pr with proper final fixes and commits. So that the git history should be clean

ojuschugh1:
Thanks @epage  and @weihanglo  !!

PR #15806, opened 2025-08-05
Recommend `package.rust-version` in the Rust version section of `reference/semver.md`.

kpreid:
Adds a recommendation that, to mitigate “possibly breaking” Rust version requirement increases, they should be declared using `package.rust-version`.

My main goal in this change is to make it *possible to find* the Rust version section in this page by searching it, but the most appropriate way to fit it into this section seemed to me to be making this recommendation. Declaring (and maintaining) `package.rust-version` is additional work, but it is a much lesser burden on maintainers than other items in the same list such as "Provide a large window of support".

### DIFF
--- a/src/doc/src/reference/semver.md
+++ b/src/doc/src/reference/semver.md
@@ -2028,6 +2028,8 @@ Mitigation strategies:
   [`#[cfg(accessible(..))]`][cfg-accessible] features which provide an opt-in
   mechanism for new features. These are currently unstable and only available
   in the nightly channel.
+* Use [`package.rust-version`] to declare your package’s compatibility.
+  This allows `cargo update` to avoid introducing an incompatible update.
 
 ### Possibly-breaking: changing the platform and environment requirements {#env-change-requirements}
 
@@ -2296,6 +2298,7 @@ document what your commitments are.
 
 [`cfg` attribute]: ../../reference/conditional-compilation.md#the-cfg-attribute
 [`no_std`]: ../../reference/names/preludes.html#the-no_std-attribute
+[`package.rust-version`]: https://doc.rust-lang.org/cargo/reference/rust-version.html
 [`pub use`]: ../../reference/items/use-declarations.html
 [Cargo feature]: features.md
 [Cargo features]: features.md


src/doc/src/reference/semver.md
```diff
@@ -2028,6 +2028,8 @@ Mitigation strategies:
   [`#[cfg(accessible(..))]`][cfg-accessible] features which provide an opt-in
   mechanism for new features. These are currently unstable and only available
   in the nightly channel.
+* Use [`package.rust-version`] to declare your package’s compatibility.
```
weihanglo: * Instead of `cargo update`, should we say "Cargo's dependency resolution"?
* MSRV resolver at this moment is best-effort, it will still fall back to incompatible version if nothing found, if we want to be more precise.

weihanglo (commented):
Thanks for the contribution!

src/doc/src/reference/semver.md
```diff
@@ -2028,6 +2028,8 @@ Mitigation strategies:
...
+  This allows `cargo update` to avoid introducing an incompatible update.
```
epage: What do you think of:
```suggestion
* Specify your minimum-supported Rust version by setting [`package.rust-version`], allowing Cargo to [select older versions of your package] when needed rather than erroring.  Be sure to consider the [support expectations] when doing so.

[`package.rust-version`]: https://doc.rust-lang.org/cargo/reference/rust-version.html
[select older versions of your package]: https://doc.rust-lang.org/cargo/reference/resolver.html#rust-version
[support expectations]: https://doc.rust-lang.org/cargo/reference/rust-version.html#support-expectations
```
- Re-orders and rewords things so the emphasis is on your package's support, to avoid people setting and forgetting this and it being wrong
- Explicitly calling out support expectations to avoid people setting and forgetting this and it being wrong
- Applies feedback from weihanglo

Also, based on what we've written for the `package.rust-version` page, I think it would be fine to list this first, rather than last.

kpreid: Done, including putting it first.

src/doc/src/reference/semver.md
kpreid: * Done.
* Wasn’t sure how to work this in cleanly, but it now says that Cargo may “attempt to” select older versions.

epage:
Mind cleaning up the commits into logical, atomic units rather than the history of how this was implemented?

src/doc/src/reference/semver.md
```diff
@@ -2018,6 +2018,10 @@ previous releases). Just keep in mind that some large projects may not be able
 to update their Rust toolchain rapidly.
 
 Mitigation strategies:
+* Specify your package’s minimum-supported Rust version by setting
```
epage: If making changes to the history, one last tweak (I know, I recommended this wording):

```suggestion
* Document your package’s minimum-supported Rust version by setting
```

This changes the base assumption to be that this item is for if you have an MSRV and so you likely also verify it.  We have the note at the end about support considerations but people are likely to miss that when scanning through this.

I'm tempted to tweak this further to find a way to make it clear that an MSRV isn't always literally what compiles but not finding a great way, so leaving it.

kpreid:
> Mind cleaning up the commits into logical, atomic units rather than the history of how this was implemented?

I can never seem to get this right …
I think that initially I was trying to preserve the authorship but `Co-authored-by` can be preserved in a rebase.

Anyway, I have now rebased it into one commit.

weihanglo (approved):
Looks good. Thanks for doing this!

weihanglo:
<https://github.com/rust-lang/cargo/actions/runs/18222468102/job/51885374574>.

```
---- cache_lock::shared_then_download stdout ----

thread 'cache_lock::shared_then_download' panicked at C:\a\cargo\cargo\crates\cargo-test-support\src\lib.rs:1618:13:
test did not finish within 10 attempts (1.0038996s total)
```

requeuing

weihanglo:
At this point, we might want to temporarily disable `reserved_windows_name`. It has happened quite often recently.

```
 ---- package::reserved_windows_name stdout ----
running `D:\a\cargo\cargo\target\debug\cargo.exe package`

thread 'package::reserved_windows_name' (8764) panicked at tests\testsuite\package.rs:2997:10:

test failed running `D:\a\cargo\cargo\target\debug\cargo.exe package`
error: process exited with code 0 (expected 101)
--- stdout

--- stderr
warning: manifest has no documentation, homepage or repository.
   Packaging foo v0.0.1 (D:\a\cargo\cargo\target\tmp\cit\t2538\foo)
    Packaged 4 files, 1.4KiB (877B compressed)
   Verifying foo v0.0.1 (D:\a\cargo\cargo\target\tmp\cit\t2538\foo)
 Downloading crates ...
  Downloaded bar v1.0.0 (registry `dummy-registry`)
   Compiling bar v1.0.0
   Compiling foo v0.0.1 (D:\a\cargo\cargo\target\tmp\cit\t2538\foo\target\package\foo-0.0.1)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.59s
```

Issue #10937, opened 2022-08-04
`cargo:rustc-link-arg-tests` not available to unit/module tests


metasim:
### Problem

According to [The Cargo Book](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-arg-tests), the `build.rs` file can emit the `cargo:rustc-link-arg-tests=FLAG` setting to have it applied passed to `rustc` as a `-C link-arg=FLAG` but only when building a "tests target". From the documentation it's not immediately clear that this only works with integration-style tests in `tests/`. Furthermore, there's no indication of  how special link args might be specified for unit tests (this is particularly relevant for running tests for PyO3 extension modules on MacOS).

Also, if there are no tests `tests/`, the `cargo test` comands fails with the message:

    error: invalid instruction `cargo:rustc-link-arg-tests` from build script of ...
    The package xyz ... does not have a test target

### 1. Create a project with the following files:

`Cargo.toml`:
```toml
[package]
name = "build-script-link-args-for-tests"
version = "0.1.0"
edition = "2021"
[dependencies]
```

`build.rs`:
```rust
fn main() {
    println!("cargo:rustc-link-arg-tests=-Wl,-rpath,/usr/local/lib") // Arbitrary flag.
}
```

`src/lib.rs`:
```rust
pub struct Foo;

#[cfg(test)]
mod tests {
    #[test]
    fn can_haz_foo() {
        let _ = crate::Foo;
    }
}
```

### 2. Run tests

```shell
cargo test
```
(observe cargo "invalid instruction" error)

### 3. Add an integration test

`tests/test_foo.rs`:
```rust
use build_script_link_args_for_tests::Foo;

#[test]
fn i_haz_foo() {
    let _ = Foo;
}
```

### 4. Run tests
```shell
cargo test
```
(observe no cargo error)

### 5. Observe application of linker arguments

```shell
RUSTFLAGS='--print link-args' cargo test | grep '\-Wl,\-rpath,/usr/local/lib'
```

(observe linker arg only applied to `.../build-script-link-args-for-tests/target/debug/deps/test_foo-...`)

### Possible Solution(s)

a. Allow `cargo:rustc-link-arg-tests` to apply to unit (in-module) tests, or create a new flag specific to unit tests. 😄 
b. Update documentation to indicate unit tests aren't supported by this feature. 😞 

### Notes

May also apply to `#[bench]` functions.

### Version

```text
[...]

PR #15814, opened 2025-08-07
docs: clarify that `cargo:rustc-link-arg-tests` only applies to integration tests

t-moe:

`cargo:rustc-link-arg-tests` only applies to integration test targets, not to all test targets as originally intended.

This PR updates the doc, to at least reflect what `cargo:rustc-link-arg-tests` currently does, until #10937 is resolved.


epage:
In documenting this, I worry this could be viewed as being prescriptive and could influence whether we can fix #10937.

epage (changes requested):
In documenting this, I worry this could be viewed as being prescriptive and could influence whether we can fix #10937.

PR #15828, opened 2025-08-12
add is_inherited methods to InheritableDependency and InheritableField

2ndDerivative:
I was looking into adding cargo_util_schemas as a parsing backend for crates-io and this was a little method I found myself missing a lot.

### DIFF
--- a/crates/cargo-util-schemas/src/manifest/mod.rs
+++ b/crates/cargo-util-schemas/src/manifest/mod.rs
@@ -351,6 +351,10 @@ impl<T> InheritableField<T> {
             InheritableField::Value(defined) => Some(defined),
         }
     }
+
+    pub fn is_inherited(&self) -> bool {
+        matches!(self, Self::Inherit(_))
+    }
 }
 
 //. This already has a `Deserialize` impl from version_trim_whitespace


crates/cargo-util-schemas/src/manifest/mod.rs
weihanglo, line 1: > parsing backend for crates-io

I am fine we adding these helper functions. Just curious about the thing you're doing. Mind sharing a bit more?

weihanglo (approved):

crates/cargo-util-schemas/src/manifest/mod.rs
2ndDerivative: I am experimenting with replacing the toml parsing backend of crates.io with cargo_util_schemas since that's what cargo is using to produce the toml file in cargo package.

It's just an experiment, so I'd like to keep these either way (I also wanna use c_u_s for an own registry I'm implementing)

weihanglo: Replacing with cargo-util-schemas in the official crates.io project? You might want to look at <https://github.com/rust-lang/crates.io/discussions/7811>, and the other functionalities it is using that cargo-utuil-schemas lacks.

Anyway, thanks for the info. Merging!

PR #15833, opened 2025-08-13
Stabilize `build.build-dir`

### LINKED ISSUE #14125: Redefine `CARGO_TARGET_DIR` to be only an artifacts directory

ranger-ross:
This PR stabilizes `build.build-dir` which moves intermediate build artifacts out of `target-dir`.

* Tracking issue: #14125
  * Call for testing: https://github.com/rust-lang/cargo/issues/14125#issuecomment-3029073591
  * FCP: https://github.com/rust-lang/cargo/issues/14125#issuecomment-3029073904
* Implementation: https://github.com/rust-lang/cargo/pull/15104

Closes #14125

### DIFF
--- a/src/cargo/core/features.rs
+++ b/src/cargo/core/features.rs
@@ -844,7 +844,6 @@ unstable_cli_options!(
     avoid_dev_deps: bool = ("Avoid installing dev-dependencies if possible"),
     binary_dep_depinfo: bool = ("Track changes to dependency artifacts"),
     bindeps: bool = ("Allow Cargo packages to depend on bin, cdylib, and staticlib crates, and use the artifacts built by those crates"),
-    build_dir: bool = ("Enable the `build.build-dir` option in .cargo/config.toml file"),
     #[serde(deserialize_with = "deserialize_comma_separated_list")]
     build_std: Option<Vec<String>>  = ("Enable Cargo to compile the standard library itself as part of a crate graph compilation"),
     #[serde(deserialize_with = "deserialize_comma_separated_list")]
@@ -969,6 +968,8 @@ const STABILIZED_DOCTEST_XCOMPILE: &str = "Doctest cross-compiling is now always
 
 const STABILIZED_PACKAGE_WORKSPACE: &str =
     "Workspace packaging and publishing (a.k.a. `-Zpackage-workspace`) is now always enabled.";
+
+const STABILIZED_BUILD_DIR: &str = "build.build-dir is now always enabled.";
 
 fn deserialize_comma_separated_list<'de, D>(
     deserializer: D,
@@ -1352,6 +1353,7 @@ impl CliUnstable {
             "registry-auth" => stabilized_warn(k, "1.74", STABILIZED_REGISTRY_AUTH),
             "check-cfg" => stabilized_warn(k, "1.80", STABILIZED_CHECK_CFG),
             "package-workspace" => stabilized_warn(k, "1.89", STABILIZED_PACKAGE_WORKSPACE),
+            "build-dir" => stabilized_warn(k, "1.91", STABILIZED_BUILD_DIR),
 
             // Unstable features
             // Sorted alphabetically:
@@ -1360,7 +1362,6 @@ impl CliUnstable {
             "avoid-dev-deps" => self.avoid_dev_deps = parse_empty(k, v)?,
             "binary-dep-depinfo" => self.binary_dep_depinfo = parse_empty(k, v)?,
             "bindeps" => self.bindeps = parse_empty(k, v)?,
-            "build-dir" => self.build_dir = parse_empty(k, v)?,
             "build-std" => self.build_std = Some(parse_list(v)),
             "build-std-features" => self.build_std_features = Some(parse_list(v)),
             "cargo-lints" => self.cargo_lints = parse_empty(k, v)?,
--- a/src/doc/src/reference/build-cache.md
+++ b/src/doc/src/reference/build-cache.md
@@ -1,10 +1,21 @@
 # Build cache
 
-Cargo stores the output of a build into the "target" directory. By default,
-this is the directory named `target` in the root of your
-[*workspace*][def-workspace]. To change the location, you can set the
+Cargo stores the output of a build into the "target" and "build" directories. By default,
+both directories point to a directory named `target` in the root of your
+[*workspace*][def-workspace]. To change the location of the target-dir, you can set the
 `CARGO_TARGET_DIR` [environment variable], the [`build.target-dir`] config
-value, or the `--target-dir` command-line flag.
+value, or the `--target-dir` command-line flag. To change the location of the build-dir, you can set the
+`CARGO_BUILD_BUILD_DIR` [environment variable] or the [`build.build-dir`] config value.
+
+Artifacts are split in two catagories:
+* Final build artifacts
+  * Final build artifacts are output meant for end users of Cargo
+  * e.g. binaries for bin crates, output of `cargo doc`, Cargo `--timings` reports
+  * Stored in the target-dir
+* Intermediate build artifacts
+  * Intermediate build artifacts are internal to Cargo and the Rust compiler
+  * End users will generally not need to interact with intermediate build artifacts
+  * Stored in the Cargo build-dir
 
 The directory layout depends on whether or not you are using the `--target`
 flag to build for a specific platform. If `--target` is not specified, Cargo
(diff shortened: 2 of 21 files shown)


tests/testsuite/build_dir.rs
epage, line 1: This seems to be removing `-Zbuild-dir` flags.  We generally expect commits to be atomic which includes tests passing

src/cargo/core/features.rs
```diff
@@ -1360,7 +1359,6 @@ impl CliUnstable {
             "avoid-dev-deps" => self.avoid_dev_deps = parse_empty(k, v)?,
             "binary-dep-depinfo" => self.binary_dep_depinfo = parse_empty(k, v)?,
             "bindeps" => self.bindeps = parse_empty(k, v)?,
-            "build-dir" => self.build_dir = parse_empty(k, v)?,
```
epage, line 1363: See https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/features/index.html#stabilization

src/doc/src/reference/unstable.md
```diff
@@ -80,7 +80,6 @@ Each new feature described below should explain how to use it.
     * [feature-unification](#feature-unification) --- Enable new feature unification modes in workspaces
 * Output behavior
     * [artifact-dir](#artifact-dir) --- Adds a directory where artifacts are copied to.
-    * [build-dir](#build-dir) --- Adds a directory where intermediate build artifacts are stored.
```
epage, line 83: See https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/features/index.html#stabilization

src/doc/src/reference/config.md
epage, line 1: We should probably also update https://doc.rust-lang.org/cargo/reference/build-cache.html

src/doc/src/reference/config.md
```diff
@@ -482,6 +482,22 @@ is a directory named `target` located at the root of the workspace.
 
 Can be overridden with the `--target-dir` CLI option.
 
+#### `build.build-dir`
```
epage, line 488: We need to update the example config at the top of the document

epage, line 488: Should we add any nuance, cross-linking, or interaction documentation to `target-dir` above?

tests/testsuite/build_dir.rs
ranger-ross: Sorry, was trying to split up the commits to make the review easier but makes sense.
I squashed everything into a single stabilization commit

src/cargo/core/features.rs
ranger-ross: Thanks, was unaware of this process.
I updated the changes to follow that process

src/doc/src/reference/config.md
ranger-ross: I added a link to the build cache documentation to both target-dir and build-dir sections

src/doc/src/reference/config.md
```diff
@@ -70,7 +70,8 @@ rustc-wrapper = "…"           # run this wrapper instead of `rustc`
 rustc-workspace-wrapper = "…" # run this wrapper instead of `rustc` for workspace members
 rustdoc = "rustdoc"           # the doc generator tool
 target = "triple"             # build for the target triple (ignored by `cargo install`)
-target-dir = "target"         # path of where to place all generated artifacts
+target-dir = "target"         # path of where to place generated artifacts
+build-dir = "build"           # path of where to place intermediate build artifacts
```
epage: Should we assign this to `target` as well?

src/doc/src/reference/build-cache.md
```diff
@@ -53,15 +64,15 @@ Directory | Description
...
+Cargo also creates several other directories and files in the build-dir needed for the build
+process. The build-dir layout is considered internal to Cargo, and is subject to
 change. Some of these directories are:
 
 Directory | Description
 ----------|------------
-<code style="white-space: nowrap">target/debug/deps/</code> | Dependencies and other artifacts.
-<code style="white-space: nowrap">target/debug/incremental/</code> | `rustc` [incremental output], a cache used to speed up subsequent builds.
-<code style="white-space: nowrap">target/debug/build/</code> | Output from [build scripts].
+<code style="white-space: nowrap">build-dir/debug/deps/</code> | Dependencies and other artifacts.
```
weihanglo, line 73: Should we write `<build-dir>/debug/deps` to disambiguate that it is a placeholder?

src/doc/src/reference/build-cache.md
weihanglo, line 1: https://doc.rust-lang.org/nightly/cargo/appendix/glossary.html#target

Should we update this glossary page as well, or at least link to the "Build cache" page?

src/doc/src/reference/build-cache.md
ranger-ross: both suggestions seem reasonable, I will open a follow up docs PR  :)

Issue #13051, opened 2023-11-26
Add `target="host"` meaning the current host


kpreid:
### Problem

Cargo defaults to compiling for the host's target triple, which is appropriate for most cases, particularly because this is the only (default) way to build tests that can be run. However, once the `target` has been overridden [by configuration](https://doc.rust-lang.org/cargo/reference/config.html), there is no way to explicitly request on the command line that it be the host.

### Proposed Solution

Wherever a target triple is accepted, accept the special string `host` as an alias for the host platform's target triple.

### Notes

Use cases:

* Any build automation that wishes to work with different target triples can unconditionally pass a string without needing a special “absent” case.
* Library code that supports `no_std`, or is concerned with other platform limitations such as pointer width and atomics, could have an accompanying configuration which automatically builds for the host, for testing, *and* another target, as a compatibility check.
* This could be used as `cargo test --target=host` to run some platform-independent tests in a package otherwise devoted to being cross-compiled (such that its `config.toml` has a `target` specified to enable IDE support and better default behaviors).
* `--target=host` will not be identical to the behavior with no options specified, due to the behavior differences triggered by using `--target` at all. This feature will introduce a way to trigger that intentionally _without_ also needing to specify a particular target. This could be considered an alternative to [`target-applies-to-host = false`](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#target-applies-to-host), but specifically does not introduce any new semantics that are not currently obtainable by `--target=<whatever the host triple is>`.

bravely-beep:
I'm also running into this issue; specifically, the second bullet point use-case in the original post would be very helpful for my project.

zdivelbiss:
It seems relatively obvious that `target` supporting `host` is the way to go with this. Not only is it already mentioned in internal documentation as a potentially useful feature—there's good traction around this issue.

My specific use-case is for a project that supports multiple non-host platforms, and trying to get that to work nicely with Rust Analyzer, without breaking my `xtask` or the project's crates. Indeed, if I set a default target for Rust Analyzer to use (non-host for this project) that results in Rust Analyzer then miscompiling the `xtask` crate because it can't find an `openssl` version to facilitate needless cross-compilation to the non-host target. Being able to specify `target = "host"` in the Cargo config would solve this handily.

[...]

Issue #13423, opened 2024-02-08
Fail early when trying to build for an unavailable target


mrivnak:
### Problem

If trying to build for a target, wasm32-unknown-unknown, for example, and the target is not installed with rustup, cargo will still attempt to build and throw lots of errors which don't have an obvious cause. These are rustc mostly being unable to find items from the standard library. It would be helpful for cargo to check that the correct target toolchain is available and provide and appropriate error message.

### Notes

I imagine this isn't an issue for no_std crates, but since that is resolved at the crate level beforehand it should be feasible for cargo to make that determination.

mrivnak:
Also, specifying a target in the `rust-toolchain.toml` file will automatically download the target when you run `cargo build` so perhaps cargo could do the same when specifying a target via the command line.

weihanglo:
On a vacation so the reply may be quite shallow.

> Also, specifying a target in the `rust-toolchain.toml` file will automatically download the target when you run `cargo build` so perhaps cargo could do the same when specifying a target via the command line.

In this case, the download is done by rustup. `cargo` being called is actually a wrapper (aka [proxy](https://rust-lang.github.io/rustup/concepts/proxies.html) provided by rustup, and then rustup delegates to the real `cargo` binary. On this point, Cargo doesn't know whether a platform target is available or not. 

Even if a target is installed and available, it can still during the compilation when some system tools are missing. For example, rustc can build `wasm32-unknown-emscripten` very well until it needs `emcc` to link to other C libs. 

I don't have a good idea for now, but always welcome any improvement in error messages.

epage:
When you add a target, that just means `rust-std` is installed.  We still call into the same compiler.  So this likely being discovered late in the build process when it sees you using features and can't find what to link against.  I suspect that improving the error message would be something rustc would have to do, if it can even do it.

As for cargo vs rustup, something to add to the above is all of this needs to work without rustup as that is an optional package.  Even if we made it required there are restrictions on what we could do within certain environments the Rust project targets that we would have to carefully evaluate.

mrivnak:
Thank you for your responses. I think minimally it would be nice to have the rustup cargo proxy handle the `--target` option the same as it does for targets in `rust-toolchain.toml`. As for a warning in the build process yeah it seems like that would just be a rustc thing then, and a special case for rust-std.

PR #15838, opened 2025-08-14
Implement `host`-target substitution

zdivelbiss:

This PR resolves: https://github.com/rust-lang/cargo/issues/13051

Namely, it allows users to invoke cargo subcommands that accept a `--target` directive to specify the `host` target, which is later substituted in the command processing layer by the host's real target triple (for instance, on most Linux distributions, `cargo build --target host` would effectively run `cargo build --target x86_64-unknown-linux-gnu`).

This additionally applies to usage within `config.toml`, like so:

```toml
# .cargo/config.toml

[build]
target = "host"
```

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -670,7 +670,7 @@ pub trait ArgMatchesExt {
             bail!(
                 "\"--target\" takes a target architecture as an argument.
 
-Run `{cmd}` to see possible targets, or specify `host` for the host architecture."
+Run `{cmd}` to see possible targets."
             );
         }
         Ok(self._values_of("target"))
--- a/src/doc/man/includes/options-target-triple.md
+++ b/src/doc/man/includes/options-target-triple.md
@@ -1,18 +1,19 @@
 {{#option "`--target` _triple_"}}
-{{actionverb}} for the specified target {{~#if multitarget }} (may be specified multiple times) {{~/if}}
-
+{{actionverb}} for the specified target architecture. {{~#if multitarget }} Flag may be specified multiple times. {{~/if}}
 {{~#if target-default-to-all-arch}} The default is all architectures.
 {{~else}} The default is the host architecture.
 {{~/if}} The general format of the triple is
 `<arch><sub>-<vendor>-<sys>-<abi>`.
 
-You may specify the following kinds of targets:
-- Any supported target in `rustc --print target-list` (note: you have to install/add the target to use it).
-- `host`, which will internally be substituted by the host's target. This can be particularly useful if you're cross-compiling some crates, and don't want to specify your host's machine as a target (for instance, an `xtask` in a shared project that may be worked on by many hosts).
-- A path to a custom target specification (further reading [here](https://doc.rust-lang.org/rustc/targets/custom.html#custom-target-lookup-path)).
+Possible values:
+- Any supported target in `rustc --print target-list`.
+- `"host"`, which will internally be substituted by the host's target. This can be particularly useful if you're cross-compiling some crates, and don't want to specify your host's machine as a target (for instance, an `xtask` in a shared project that may be worked on by many hosts).
+- A path to a custom target specification. See [Custom Target Lookup Path](../../rustc/targets/custom.html#custom-target-lookup-path) for more information.
 
 
 This may also be specified with the `build.target` [config value](../reference/config.html).
 
-**Note**: Specifying this flag makes Cargo run in a different mode where the target artifacts are placed in a separate directory. See the [build cache](../reference/build-cache.html) documentation for more details.
+Note that specifying this flag makes Cargo run in a different mode where the
+target artifacts are placed in a separate directory. See the
+[build cache](../reference/build-cache.html) documentation for more details.
 {{/option}}
--- a/src/doc/src/reference/config.md
+++ b/src/doc/src/reference/config.md
@@ -459,16 +459,16 @@ Sets the executable to use for `rustdoc`.
 
 The default [target platform triples][target triple] to compile to.
 
-You may specify the following kinds of targets:
-- Any supported target in `rustc --print target-list` (note: you have to install/add the target to use it).
-- `host`, which will internally be substituted by the host's target. This can be particularly useful if you're cross-compiling some crates, and don't want to specify your host's machine as a target (for instance, an `xtask` in a shared project that may be worked on by many hosts).
-- A path to a custom target specification (further reading [here](https://doc.rust-lang.org/rustc/targets/custom.html#custom-target-lookup-path)).
+Possible values:
+- Any supported target in `rustc --print target-list`.
+- `"host"`, which will internally be substituted by the host's target. This can be particularly useful if you're cross-compiling some crates, and don't want to specify your host's machine as a target (for instance, an `xtask` in a shared project that may be worked on by many hosts).
+- A path to a custom target specification. See [Custom Target Lookup Path](../../rustc/targets/custom.html#custom-target-lookup-path) for more information.
 
 Can be overridden with the `--target` CLI option.
 
 ```toml
 [build]
-target = ["host", "x86_64-unknown-linux-gnu", "i686-unknown-linux-gnu"]
+target = ["x86_64-unknown-linux-gnu", "i686-unknown-linux-gnu"]
 ```
 
 #### `build.target-dir`
--- a/src/doc/man/generated_txt/cargo-bench.txt
+++ b/src/doc/man/generated_txt/cargo-bench.txt
@@ -220,30 +220,31 @@ OPTIONS
 
    Compilation Options
        --target triple
-           Benchmark for the specified target (may be specified multiple times)
-           The default is the host architecture. The general format of the
-           triple is <arch><sub>-<vendor>-<sys>-<abi>.
+           Benchmark for the specified target architecture. Flag may be
+           specified multiple times. The default is the host architecture. The
+           general format of the triple is <arch><sub>-<vendor>-<sys>-<abi>.
 
-           You may specify the following kinds of targets:
+           Possible values:
 
-           o  Any supported target in rustc --print target-list (note: you have
-              to install/add the target to use it).
+           o  Any supported target in rustc --print target-list.
 
-           o  host, which will internally be substituted by the host’s
+           o  "host", which will internally be substituted by the host’s
               target. This can be particularly useful if you’re
               cross-compiling some crates, and don’t want to specify your
               host’s machine as a target (for instance, an xtask in a shared
               project that may be worked on by many hosts).
 
-           o  A path to a custom target specification (further reading here
-              <https://doc.rust-lang.org/rustc/targets/custom.html#custom-target-lookup-path>).
+           o  A path to a custom target specification. See Custom Target Lookup
+              Path
+              <https://doc.rust-lang.org/rustc/targets/custom.html#custom-target-lookup-path>
+              for more information.
 
            This may also be specified with the build.target config value
            <https://doc.rust-lang.org/cargo/reference/config.html>.
 
-           Note: Specifying this flag makes Cargo run in a different mode where
-           the target artifacts are placed in a separate directory. See the
-           build cache
+           Note that specifying this flag makes Cargo run in a different mode
+           where the target artifacts are placed in a separate directory. See
+           the build cache
            <https://doc.rust-lang.org/cargo/reference/build-cache.html>
            documentation for more details.
 
--- a/src/doc/man/generated_txt/cargo-build.txt
+++ b/src/doc/man/generated_txt/cargo-build.txt
@@ -137,30 +137,31 @@ OPTIONS
 
    Compilation Options
        --target triple
-           Build for the specified target (may be specified multiple times) The
-           default is the host architecture. The general format of the triple
-           is <arch><sub>-<vendor>-<sys>-<abi>.
+           Build for the specified target architecture. Flag may be specified
+           multiple times. The default is the host architecture. The general
+           format of the triple is <arch><sub>-<vendor>-<sys>-<abi>.
 
-           You may specify the following kinds of targets:
+           Possible values:
 
-           o  Any supported target in rustc --print target-list (note: you have
-              to install/add the target to use it).
+           o  Any supported target in rustc --print target-list.
 
-           o  host, which will internally be substituted by the host’s
+           o  "host", which will internally be substituted by the host’s
               target. This can be particularly useful if you’re
               cross-compiling some crates, and don’t want to specify your
               host’s machine as a target (for instance, an xtask in a shared
               project that may be worked on by many hosts).
 
-           o  A path to a custom target specification (further reading here
-              <https://doc.rust-lang.org/rustc/targets/custom.html#custom-target-lookup-path>).
+           o  A path to a custom target specification. See Custom Target Lookup
+              Path
+              <https://doc.rust-lang.org/rustc/targets/custom.html#custom-target-lookup-path>
+              for more information.
 
            This may also be specified with the build.target config value
            <https://doc.rust-lang.org/cargo/reference/config.html>.
 
-           Note: Specifying this flag makes Cargo run in a different mode where
-           the target artifacts are placed in a separate directory. See the
-           build cache
+           Note that specifying this flag makes Cargo run in a different mode
+           where the target artifacts are placed in a separate directory. See
+           the build cache
            <https://doc.rust-lang.org/cargo/reference/build-cache.html>
            documentation for more details.
 
--- a/src/doc/man/generated_txt/cargo-check.txt
+++ b/src/doc/man/generated_txt/cargo-check.txt
@@ -134,30 +134,31 @@ OPTIONS
 
    Compilation Options
        --target triple
-           Check for the specified target (may be specified multiple times) The
-           default is the host architecture. The general format of the triple
-           is <arch><sub>-<vendor>-<sys>-<abi>.
+           Check for the specified target architecture. Flag may be specified
+           multiple times. The default is the host architecture. The general
+           format of the triple is <arch><sub>-<vendor>-<sys>-<abi>.
 
-           You may specify the following kinds of targets:
+           Possible values:
 
-           o  Any supported target in rustc --print target-list (note: you have
-              to install/add the target to use it).
+           o  Any supported target in rustc --print target-list.
 
-           o  host, which will internally be substituted by the host’s
+           o  "host", which will internally be substituted by the host’s
               target. This can be particularly useful if you’re
               cross-compiling some crates, and don’t want to specify your
               host’s machine as a target (for instance, an xtask in a shared
               project that may be worked on by many hosts).
 
-           o  A path to a custom target specification (further reading here
-              <https://doc.rust-lang.org/rustc/targets/custom.html#custom-target-lookup-path>).
+           o  A path to a custom target specification. See Custom Target Lookup
+              Path
+              <https://doc.rust-lang.org/rustc/targets/custom.html#custom-target-lookup-path>
+              for more information.
 
            This may also be specified with the build.target config value
            <https://doc.rust-lang.org/cargo/reference/config.html>.
 
-           Note: Specifying this flag makes Cargo run in a different mode where
-           the target artifacts are placed in a separate directory. See the
-           build cache
+           Note that specifying this flag makes Cargo run in a different mode
+           where the target artifacts are placed in a separate directory. See
+           the build cache
            <https://doc.rust-lang.org/cargo/reference/build-cache.html>
            documentation for more details.
 
--- a/src/doc/man/generated_txt/cargo-clean.txt
+++ b/src/doc/man/generated_txt/cargo-clean.txt
@@ -45,30 +45,31 @@ OPTIONS
            target in the root of the workspace.
 
        --target triple
-           Clean for the specified target (may be specified multiple times) The
-           default is the host architecture. The general format of the triple
-           is <arch><sub>-<vendor>-<sys>-<abi>.
+           Clean for the specified target architecture. Flag may be specified
+           multiple times. The default is the host architecture. The general
+           format of the triple is <arch><sub>-<vendor>-<sys>-<abi>.
 
-           You may specify the following kinds of targets:
+           Possible values:
 
-           o  Any supported target in rustc --print target-list (note: you have
-              to install/add the target to use it).
+           o  Any supported target in rustc --print target-list.
 
-           o  host, which will internally be substituted by the host’s
+           o  "host", which will internally be substituted by the host’s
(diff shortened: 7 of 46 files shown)


zdivelbiss:
In writing up this PR's explanation, I was unsure how to properly outline the following:

```
```

... due to my unfamiliarity with how testing the program that tests programs would exactly work. Perhaps, simply clone the PR and try to run `--target host`?

Thanks.

src/cargo/core/compiler/compile_kind.rs
weihanglo, line 1: Thanks for the contribution. We'll need some tests for it, as well as documentation update.

* For documenting, this is the place to go: <https://github.com/rust-lang/cargo/blob/master/src/doc/man/includes/options-target-triple.md>. Also see how to regenerate docs [here](https://github.com/rust-lang/cargo/blob/a2f54c794bfe6004fd3025cd24dcc5eb3103a6e2/src/doc/README.md?plain=1#L36-L53).
* For writing tests, see <https://doc.crates.io/contrib/tests/index.html>. I would assume it would be some functional tests, at least one for `--target host`, the other for `build.target = ["host"]`. Those tests may go into `tests/testsuite/build.rs` file. (multitarget.rs might be another candidate since `build.rs` has become too big a file).

src/cargo/core/compiler/compile_kind.rs
```diff
@@ -84,15 +84,27 @@ impl CompileKind {
         fallback: CompileKindFallback,
     ) -> CargoResult<Vec<CompileKind>> {
         let dedup = |targets: &[String]| {
-            Ok(targets
+            let deduplicated_targets = targets
```
weihanglo, line 87: It is fine, but not sure what the intent of creating a variable here is.

src/cargo/util/command_prelude.rs
```diff
@@ -1248,15 +1248,18 @@ fn get_target_triples() -> Vec<clap_complete::CompletionCandidate> {
...
 
     if let Ok(targets) = get_target_triples_from_rustup() {
-        candidates = targets;
+        candidates.extend(targets);
     }
 
     if candidates.is_empty() {
         if let Ok(targets) = get_target_triples_from_rustc() {
-            candidates = targets;
+            candidates.extend(targets);
```
weihanglo: why this change?

src/cargo/util/command_prelude.rs
```diff
@@ -1248,15 +1248,18 @@ fn get_target_triples() -> Vec<clap_complete::CompletionCandidate> {
...
         }
     }
 
+    // Allow tab-completion for `host` as the desired target.
+    candidates.push(clap_complete::CompletionCandidate::new("host"));
```
epage: Can you read `env!("RUST_HOST_TARGET")` and put that in the help string in some way, like `alias: {}`?

src/cargo/core/compiler/compile_kind.rs
epage: Cases we care about
- `--target host` and  `build.target = ["host"]` work and are treated as cross-compilation
- de-duplication of targets works
- that `"host"` in other contexts doesn't accidentally work

As a tip on test writing: you can add the tests in a commit before, doing a `"{SOMETHING}"..replace(cargo_test_support:rustc_host())`.  Then in this commit, you can replace that with `host"` and nothing else in the test should change, making it clear through the diff that `"host"` truly has the same behavior as `--target <something>`

src/cargo/util/command_prelude.rs
zdivelbiss: I believe I made that change incidentally (while debugging), and forgot to revert it. Thanks for catching it.

src/cargo/core/compiler/compile_kind.rs
zdivelbiss: Wrapping such a large block of code in an `Ok()` seems very unclear / noisy. I can revert if it's important.

src/cargo/core/compiler/compile_kind.rs
zdivelbiss: > Thanks for the contribution. We'll need some tests for it, **as well as documentation update.**
> 
>  * For documenting, this is the place to go: https://github.com/rust-lang/cargo/blob/master/src/doc/man/includes/options-target-triple.md.

Do you feel this is an acceptable addition to the documentation?

```
...
{{~/if}} The general format of the triple is
`<arch><sub>-<vendor>-<sys>-<abi>`. Run `rustc --print target-list` for a
list of supported targets.

my addition
v v v v v v v
In addition to the supported targets, you may specify `host` as the target. This has the same behavior as specifying a regular target, and produces a build for the host machine's architecture.

{{~#if multitarget }} This flag may be specified multiple times. {{~/if}}
...
```

src/cargo/util/command_prelude.rs
zdivelbiss: Definitely, but where exactly do these completions show up? And their `.help()` strings. I assumed it was tab-completion, but navigating to `target/debug/` and tab-completing `./cargo build --target` only provides one option (the only installed target, which is the host architecture).

epage: Do you have our [unstable completions configured](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#native-completions) and using your custom of cargo?

Whether the help string shows up is dependent on the shell.  Clap supports it through some shells native mechanisms.  There are ways to hack it in to some shells but haven't done that yet.

src/cargo/core/compiler/compile_kind.rs
weihanglo: Got it. I thought you wanted to do something more but forgot. Never mind. Fine with it :)

src/cargo/core/compiler/compile_kind.rs
weihanglo: > --target host and build.target = ["host"] work and are treated as cross-compilation

So probably put tests under `tests/testsuite/cross_compile.rs`. Sounds good.

> Do you feel this is an acceptable addition to the documentation?

Looks great, though is it possible to be a bit more concise?

[`--jobs` flag](https://doc.rust-lang.org/nightly/cargo/commands/cargo-build.html#miscellaneous-options) has this for the "default" string:

```
 -j N, --jobs N
     Number of parallel jobs to run. May also be specified with the
     ...
     If a string default is provided, it sets the value back to defaults.
     Should not be 0.
```

zdivelbiss: @weihanglo I opted to rewrite the documentation, as I felt the original was a bit incomplete (didn't mention custom target specifications, for instance). I understand if the change is too dramatic, please let me know.

```md
{{#option "`--target` _triple_"}}
{{actionverb}} for the specified target {{~#if multitarget }} (may be specified multiple times) {{~/if}}

{{~#if target-default-to-all-arch}} The default is all architectures.
{{~else}} The default is the host architecture.
{{~/if}} The general format of the triple is
`<arch><sub>-<vendor>-<sys>-<abi>`.

You may specify the following kinds of targets:
- Any supported target in `rustc --print target-list` (note: you have to add the target to use it).
- `host`, which will internally be substituted by the host's target. This can be particularly useful if you're cross-compiling some crates, and don't want to specify your host's machine as a target (for instance, an `xtask` in a shared project that may be worked on by many hosts).
- A path to a custom target specification (further reading [here](https://doc.rust-lang.org/rustc/targets/custom.html#custom-target-lookup-path)).


This may also be specified with the `build.target` [config value](../reference/config.html).

**Note**: Specifying this flag makes Cargo run in a different mode where the target artifacts are placed in a separate directory. See the [build cache](../reference/build-cache.html) documentation for more details.
{{/option}}
```

zdivelbiss:
I have additionally tested locally that the following `host` specifications do not compile:
- `[target.<triple>]` in `.cargo/config.toml`
- `[target.<triple>.dependencies]` in `Cargo.toml`

Should I somehow mirror this within the PR, rather than just ensuring locally that builds with the above conditions do not compile? I think I could write an always-fail test that does this, but I wanted to make sure that's the way we want it done.

src/cargo/core/compiler/compile_kind.rs
zdivelbiss: @epage Would adding specific tests for the `host` target specifier be acceptable? It seems like bit of a special case, whose implementation is somewhat easy to miss (being just a few lines inside a rather deeply nested call stack), so I think a couple tests to ensure the specifier works as expected into the future would be a good idea.

epage: I feel I'm missing something in the question because it sounds like what I said was needed (`--target host` tests)

zdivelbiss: @epage This bit seemed to suggest to me that you wanted an impermanent test (showing soundness via the diff, rather than a persistent test):

> As a tip on test writing: you can add the tests in a commit before, doing a "{SOMETHING}"..replace(cargo_test_support:rustc_host()). Then in this commit, you can replace that with host" and nothing else in the test should change, making it clear through the diff that "host" truly has the same behavior as --target <something>

... though I admittedly didn't really understand what you wrote. I'll go ahead and push the tests!

epage: Not an impermanent test but a suggestion on how to structure commits in the PR that help demonstrate what the PR does and help test the tests. See also https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request

weihanglo: Doc looks fine. You might also want to update its config doc: <https://doc.rust-lang.org/nightly/cargo/reference/config.html#buildtarget>.

src/doc/src/reference/config.md
```diff
@@ -459,17 +459,16 @@ Sets the executable to use for `rustdoc`.
...
 The default [target platform triples][target triple] to compile to.
 
-This allows passing either a string or an array of strings. Each string value
-is a target platform triple. The selected build targets will be built for each
-of the selected architectures.
-
-The string value may also be a relative path to a `.json` target spec file.
+You may specify the following kinds of targets:
+- Any supported target in `rustc --print target-list` (note: you have to install/add the target to use it).
+- `host`, which will internally be substituted by the host's target. This can be particularly useful if you're cross-compiling some crates, and don't want to specify your host's machine as a target (for instance, an `xtask` in a shared project that may be worked on by many hosts).
```
weihanglo:
```suggestion
- `"host"`, which will internally be substituted by the host's target. This can be particularly useful if you're cross-compiling some crates, and don't want to specify your host's machine as a target (for instance, an `xtask` in a shared project that may be worked on by many hosts).
```

src/doc/man/includes/options-target-triple.md
```diff
@@ -1,16 +1,18 @@
...
+
 
-This may also be specified with the `build.target`
-[config value](../reference/config.html).
+This may also be specified with the `build.target` [config value](../reference/config.html).
 
-Note that specifying this flag makes Cargo run in a different mode where the
-target artifacts are placed in a separate directory. See the
-[build cache](../reference/build-cache.html) documentation for more details.
+**Note**: Specifying this flag makes Cargo run in a different mode where the target artifacts are placed in a separate directory. See the [build cache](../reference/build-cache.html) documentation for more details.
```
weihanglo: Just realized it is not a new thing but a cosmetic change on existing stuff. Could we back these out?

src/cargo/util/command_prelude.rs
```diff
@@ -670,7 +670,7 @@ pub trait ArgMatchesExt {
             bail!(
                 "\"--target\" takes a target architecture as an argument.
 
-Run `{cmd}` to see possible targets."
+Run `{cmd}` to see possible targets, or specify `host` for the host architecture."
```
weihanglo: I feel like this might be emphasized too much? Specifying `host` doesn't seem to be a common case, and this may lead people to specify it which is usually unnecessary.

tests/testsuite/cross_compile.rs
```diff
@@ -125,6 +125,119 @@ fn simple_cross_config() {
...
+                cross_compile::native_arch()
+            ),
+        )
+        .build();
+
+    let target = rustc_host();
+    p.cargo("build -v --target host").run();
+    assert!(p.target_bin(target, "foo").is_file());
+
+    if cross_compile_can_run_on_host() {
```
weihanglo: This `if` seems unnecessary. We are compiling to host and can just run it.

```suggestion
```

tests/testsuite/cross_compile.rs
```diff
@@ -125,6 +125,119 @@ fn simple_cross_config() {
...
+                        assert_eq!(env::consts::ARCH, "{}");
+                    }}
+                "#,
+                cross_compile::native_arch()
+            ),
+        )
+        .build();
+
+    let target = rustc_host();
+    p.cargo("build -v --target host").run();
```
weihanglo: If we are not asserting verbose output, `-v` can be removed.

```suggestion
    p.cargo("build --target host").run();
```

However you might want to assert the verbose output `rustc [..] --target {rustc_host}` or alike, so we might be able to build a lib instead of a bin, which speeds up the test.

src/doc/man/includes/options-target-triple.md
```diff
@@ -1,16 +1,18 @@
 {{#option "`--target` _triple_"}}
-{{actionverb}} for the given architecture.
+{{actionverb}} for the specified target {{~#if multitarget }} (may be specified multiple times) {{~/if}}
```
weihanglo: [`target`](https://doc.rust-lang.org/nightly/cargo/appendix/glossary.html#target) is overloaded. I would recommend keeping the old way. Or "target architecture"?

weihanglo: for `multitarget` I prefer the old style, but fine with this.

src/doc/man/includes/options-target-triple.md
```diff
@@ -1,16 +1,18 @@
...
+
 {{~#if target-default-to-all-arch}} The default is all architectures.
 {{~else}} The default is the host architecture.
 {{~/if}} The general format of the triple is
-`<arch><sub>-<vendor>-<sys>-<abi>`. Run `rustc --print target-list` for a
-list of supported targets.
-{{~#if multitarget }} This flag may be specified multiple times. {{~/if}}
+`<arch><sub>-<vendor>-<sys>-<abi>`.
+
+You may specify the following kinds of targets:
```
weihanglo: What do you think a bit more concise version?

```suggestion
Possible values:
```

src/doc/man/includes/options-target-triple.md
```diff
@@ -1,16 +1,18 @@
...
+- Any supported target in `rustc --print target-list` (note: you have to install/add the target to use it).
```
weihanglo: > (note: you have to install/add the target to use it)

not every cargo user uses rustup, it might be hard to understand what install/add means here. I might defer this to #13423.

src/doc/src/reference/config.md
```diff
@@ -459,17 +459,16 @@ Sets the executable to use for `rustdoc`.
...
+- A path to a custom target specification (further reading [here](https://doc.rust-lang.org/rustc/targets/custom.html#custom-target-lookup-path)).
 
 Can be overridden with the `--target` CLI option.
 
 ```toml
 [build]
-target = ["x86_64-unknown-linux-gnu", "i686-unknown-linux-gnu"]
+target = ["host", "x86_64-unknown-linux-gnu", "i686-unknown-linux-gnu"]
```
weihanglo: Same. I am not sure if we want to emphasize `host` everywhere.

src/doc/src/reference/config.md
weihanglo: The writing style I feel like uses a bit more parenthesis than the existing style. For examaple,  for "(further reading [here])" we usually say `"See X for more information". I am not sure which is better, but personally the old style is a bit more pleasant to me.

tests/testsuite/cross_compile.rs
zdivelbiss: I am honestly not sure what the function does? I can certainly remove it.

src/doc/man/includes/options-target-triple.md
zdivelbiss: You mentioned not liking paren'd text so much, so maybe we can change it to
```
{{~#if multitarget }} Flag may be specified multiple times. {{~/if}}
```
?

tests/testsuite/cross_compile.rs
zdivelbiss: I changed it to the following:
```rust
    // instead of `src/main.rs` for ProjectBuilder
    .file("src/lib.rs", r#""#)

    p.cargo("build -v") // + `--target host` for the `--target`-specific test
        .with_stderr_contains("[RUNNING] `rustc [..] --target [HOST_TARGET] [..]`")
        .run();
```
That should make the test faster then?

src/cargo/util/command_prelude.rs
zdivelbiss: I'm not sure. I find it a particularly useful feature, so I am probably biased in how useful I think it will be for the average user haha. I'll elide it for now, and if we want more clarity later, it can always be added.

src/doc/man/includes/options-target-triple.md
zdivelbiss: Sounds good.

src/doc/src/reference/config.md
zdivelbiss: I've gone with:
```md
- A path to a custom target specification. See [Custom Target Lookup Path](https://doc.rust-lang.org/rustc/targets/custom.html#custom-target-lookup-path) for more information.
```

src/doc/man/includes/options-target-triple.md
zdivelbiss: I think "target architecture" feels correct. I've changed it.

src/doc/man/includes/options-target-triple.md
zdivelbiss: Looks good. I've changed it.

src/doc/src/reference/config.md
zdivelbiss: Sure thing. I've removed it.

src/doc/man/includes/options-target-triple.md
zdivelbiss: Sure. I've removed it.

src/doc/man/includes/options-target-triple.md
weihanglo: Looks good

src/cargo/util/command_prelude.rs
weihanglo: Thanks for the understanding!

src/doc/src/reference/config.md
weihanglo: Thanks! Let me know when this is ready for the next round of review.

src/doc/src/reference/config.md
```diff
@@ -459,11 +459,10 @@ Sets the executable to use for `rustdoc`.
...
 
-This allows passing either a string or an array of strings. Each string value
-is a target platform triple. The selected build targets will be built for each
-of the selected architectures.
-
-The string value may also be a relative path to a `.json` target spec file.
+Possible values:
+- Any supported target in `rustc --print target-list`.
+- `"host"`, which will internally be substituted by the host's target. This can be particularly useful if you're cross-compiling some crates, and don't want to specify your host's machine as a target (for instance, an `xtask` in a shared project that may be worked on by many hosts).
+- A path to a custom target specification. See [Custom Target Lookup Path](https://doc.rust-lang.org/rustc/targets/custom.html#custom-target-lookup-path) for more information.
```
weihanglo: We usually use relative paths to link to docs under the same toolchain versions.



```suggestion
- A path to a custom target specification. See [Custom Target Lookup Path](../../rustc/targets/custom.html#custom-target-lookup-path) for more information.
```

(Doesn't matter for man pages as they don't support relative paths beyond cargo book)

weihanglo (commented):
Thanks. I believe this is the last review round!

BTW, would you mind cleanup your commit history. There are some intermediate commits that may be good if we squash them for easy git diff tracing in the future.

zdivelbiss:
@weihanglo Yes, I plan to clean up the history some time today. It's been a while since I've done a rebase & squash, so I'll need to do some research to make sure I don't permanently damage something with Git.

src/doc/src/reference/config.md
zdivelbiss: I wondered about that! Thank you.

weihanglo:
>  It's been a while since I've done a rebase & squash, so I'll need to do some research to make sure I don't permanently damage something with Git.

See some resources Rust Compiler Development Guide: <https://rustc-dev-guide.rust-lang.org/git.html#rebasing-and-conflicts>. And I just [explained](https://github.com/rust-lang/cargo/pull/15849#discussion_r2280955775) why/how we maintain good Git history in Cargo in another PR 😆

weihanglo:
I can also do the rebase for you, if you don't mind no longer being the committer of those commits. You will still be the author of those and GitHub gives contribution credits to authors AFAIK.

zdivelbiss:
That's alright, I can do it. It's valuable knowledge, this git-fu. Thank you for the resources. I'm working on it now.

src/doc/man/includes/options-target-triple.md
```diff
@@ -1,14 +1,17 @@
 {{#option "`--target` _triple_"}}
-{{actionverb}} for the given architecture.
+{{actionverb}} for the specified target architecture {{~#if multitarget }} Flag may be specified multiple times. {{~/if}}
```
weihanglo: Oops. A period `.` is missing

```suggestion
{{actionverb}} for the specified target architecture. {{~#if multitarget }} Flag may be specified multiple times. {{~/if}}
```

zdivelbiss:
This commit history looks good to me. Any concerns?

src/doc/src/reference/config.md
weihanglo: I just realized manpage also support it. Nice. However, we forgot to update config.md (which is this file).

src/doc/man/includes/options-target-triple.md
zdivelbiss: Fixed.

zdivelbiss:
Thank you for catching the URL usage in the `config.md` changes. I didn't quite understand what the URI you suggested was relative to, but I see now that's is relative to `docs.rust-lang.org`.

It may be worth considering expanding the contribution guide to include more information on how documentation is written and contributed. There were many elements in this PR that were directly related to things that aren't covered in the contribution guide: syntax (e.g. that a special documentation preprocessor is used), relative pathing (confusion above), the need to run `cargo build-man`, general styling (may not be possible to document? thinking of my extensive paren usage, which is out-of-place in the docs). It may help cut down on documentation-based nits, though I'm obviously not sure how often this actually affects contributors' efficiency.

zdivelbiss:
@weihanglo No frustration at all; just thinking of how to make the process smoother for contributors in the future. Thanks for your help in getting this PR merged so quickly—it was great to receive such quick feedback.

PR #15841, opened 2025-08-15
Reorder `lto` options in profiles.md

smoelius:
Aside from `false`, the `lto` options seem to be ordered from "most optimizing" to "least optimizing".

If this interpretation is correct, then I think `false` should go between `thin` and `off`.

cc: @ehuss (who I think wrote that text)

### DIFF
--- a/src/doc/src/reference/profiles.md
+++ b/src/doc/src/reference/profiles.md
@@ -166,7 +166,7 @@ The `lto` setting controls `rustc`'s [`-C lto`], [`-C linker-plugin-lto`], and
 LTO can produce better optimized code, using whole-program analysis, at the cost
 of longer linking time.
 
-The valid options are:
+The valid options from most to least optimizing are:
 
 * `true` or `"fat"`: Performs "fat" LTO which attempts to perform
   optimizations across all crates within the dependency graph.


src/doc/src/reference/profiles.md
```diff
@@ -168,14 +168,14 @@ of longer linking time.
 
 The valid options are:
```
weihanglo: Just FYI, [rustc](https://doc.rust-lang.org/nightly/rustc/codegen-options/index.html#lto) doesn't sort them by optimization levels. And perhaps when ehuss wrote this was thinking of putting boolean options closer to each other. That also makes sense.

I think the reorder is fine, though should we mention the order as well? Like

```suggestion
The valid options from most to least optimizing are:
```

weihanglo (commented):
I gu

src/doc/src/reference/profiles.md
smoelius: Thanks for the heads up, @weihanglo. If this PR is accepted, I'll submit an analogous PR to rustc.

smoelius: Or would it be better for me to submit that PR now so that the two sources remain consistent?

weihanglo: Yeah that would be great.

I am going to merge this now, as it is not hard to revert if needed. Thank you

weihanglo (approved):

src/doc/src/reference/profiles.md
```diff
@@ -166,16 +166,16 @@ The `lto` setting controls `rustc`'s [`-C lto`], [`-C linker-plugin-lto`], and
 LTO can produce better optimized code, using whole-program analysis, at the cost
 of longer linking time.
 
-The valid options are:
+The valid options from most to least optimizing are:
```
bjorn3, line 169: As I understand it, this is not true in the general case. fat LTO can produce faster executables than ThinLTO and even if it doesn't, it is often very close. ThinLTO just generally produces a bit bigger executables.

weihanglo: So in terms of executable performance,

is it `fat >= ThinLTO` or `ThinLTO >= fat`, or hard to say?

bjorn3: As I understand it, it could be either way nowadays. There is probably still some bias for fat LTO producing faster executables than ThinLTO though.

weihanglo: I see. Thanks for pointing it out!

There is another similar ongoing in rust-lang/rust <https://github.com/rust-lang/rust/pull/145533>. Do you think we should `r-` that, and revert this PR as well?

bjorn3: I think the order makes sense. I just think we should not explicitly say "from most to least optimizing" here.

PR #15847, opened 2025-08-17
feat(lints): new lint invalid_spdx_license_expression

### LINKED ISSUE #2039: Lint `license` field (old: SPDX dual-license inconsistency)
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/cargo/issues/2039)*

According to the [current SPDX spec](https://spdx.org/sites/spdx/files/SPDX-2.0.pdf), dual-licensing should be done as: `(Apache-2.0 OR MIT)` (and there's also `AND`, `+` and `WITH`), but our [manifest format](http://doc.crates.io/manifest.html) says one should simply separate licences with `/` without being 100% clear on whether that means an "AND" or "OR".

This just seems to be an unnecessary inconsistency with the SPDX spec. Can we change the doc to encourage dual-licensing using the `(x OR y)` syntax instead?

weihanglo:

Fixes #2039

[rendered lint doc](https://github.com/weihanglo/cargo/blob/spdx-license-lints/src/doc/src/reference/lints.md#invalid_spdx_license_expression)

Add a new lint `invalid_spdx_license_expression`

New tests should already cover

* all valid/invalid examples in Lint docs
* EditionFuture (2027?) deny-by-default
* workspace inherited `package.license`

#### Should this lint workspace.package.license even nobody inherits from it?

I guess not, and that should be covered by an unused workspace inherit lint.

#### `spdx` crate version pinning

This adds a new dependency `spdx` but I didn't put exact pin (`=`) on the version. crates.io use (0.10.9). SPDX crates.io uses is still 2.3 but the latest spec version is v3. Anyway it seems that `spdx@0.10` is for SPDX spec v2.3, and `spdx@0.11` is for SPDX sped v3. I feel like it is fine we don't pin it at this moment.

https://github.com/rust-lang/crates.io/blob/bc421ae1ade7eed6d1c046456f52292ab9829095/src/licenses.rs

#### Name bikeshedding

1. `invalid_spdx_license_expression`
2. `invalid_spdx_license_expr`
3. `invalid_license_expr`
4. `non_spdx_license_expression`
5. `non_spdx_license_expr`
6. ???

--- 

BTW I can kinda feel that we need to pull out the lint machinery into a separate helper crate once we have three lints. I am repeating something there…

### DIFF
--- a/src/cargo/util/lints.rs
+++ b/src/cargo/util/lints.rs
@@ -476,6 +476,37 @@ const INVALID_SPDX_LICENSE_EXPRESSION: Lint = Lint {
     feature_gate: None,
     docs: Some(
         r#"
+### What it does
+
+Checks that the `license` field in `Cargo.toml` is a valid SPDX license expression.
+See the doc of [the `license` field] for the SPDX specification version Cargo currently supports.
+
+[the `license` field]: manifest.md#the-license-and-license-file-fields
+
+### Why it is bad
+
+Build tools, package registries, and compliance systems may fail to handle
+non-SPDX licenses, which can lead to build failures, rejected uploads,
+incorrect license reporting, or legal risks.
+
+### Examples
+
+```toml
+license = "MIT / Apache-2.0"       # Invalid: uses "/" instead of "OR"
+license = "GPL-3.0 with exception" # Invalid: uses lowercase "with" instead of "WITH"
+license = "GPL-3.0+"               # Invalid: uses the deprecated "+" operator instead of "GPL-3.0-or-later"
+license = "MIT OR (Apache-2.0"     # Invalid: unclosed parenthesis
+```
+
+Use instead:
+
+```toml
+license = "MIT OR Apache-2.0"
+license = "GPL-3.0 WITH exception"
+license = "GPL-3.0-or-later"
+license = "(MIT OR Apache-2.0) AND GPL-3.0-or-later WITH Classpath-exception-2.0"
+```
+
 "#,
     ),
 };
--- a/src/doc/src/reference/lints.md
+++ b/src/doc/src/reference/lints.md
@@ -11,6 +11,37 @@ These lints are all set to the 'warn' level by default.
 ## `invalid_spdx_license_expression`
 Set to `warn` by default
 
+### What it does
+
+Checks that the `license` field in `Cargo.toml` is a valid SPDX license expression.
+See the doc of [the `license` field] for the SPDX specification version Cargo currently supports.
+
+[the `license` field]: manifest.md#the-license-and-license-file-fields
+
+### Why it is bad
+
+Build tools, package registries, and compliance systems may fail to handle
+non-SPDX licenses, which can lead to build failures, rejected uploads,
+incorrect license reporting, or legal risks.
+
+### Examples
+
+```toml
+license = "MIT / Apache-2.0"       # Invalid: uses "/" instead of "OR"
+license = "GPL-3.0 with exception" # Invalid: uses lowercase "with" instead of "WITH"
+license = "GPL-3.0+"               # Invalid: uses the deprecated "+" operator instead of "GPL-3.0-or-later"
+license = "MIT OR (Apache-2.0"     # Invalid: unclosed parenthesis
+```
+
+Use instead:
+
+```toml
+license = "MIT OR Apache-2.0"
+license = "GPL-3.0 WITH exception"
+license = "GPL-3.0-or-later"
+license = "(MIT OR Apache-2.0) AND GPL-3.0-or-later WITH Classpath-exception-2.0"
+```
+
 
 
 ## `unknown_lints`


weihanglo:
r? Muscraft 

You might be interested in reviewing new lints.

tests/testsuite/lints/invalid_spdx_license_expression.rs
```diff
@@ -248,6 +280,20 @@ license.workspace = true
     p.cargo("check -Zcargo-lints")
         .masquerade_as_nightly_cargo(&["cargo-lints"])
         .with_stderr_data(str![[r#"
+[WARNING] invalid SPDX license expression: `MIT / Apache-2.0`
+ --> Cargo.toml:6:16
+  |
+6 | license = "MIT / Apache-2.0"
+  |                ------------ invalid character(s)
+  |
```
epage, line 288: Should this be "invalid characters" or "deprecated"?

tests/testsuite/lints/invalid_spdx_license_expression.rs
epage, line 1: Can you add a test for an invalid license identifier?

src/cargo/util/lints.rs
```diff
@@ -467,6 +467,141 @@ pub fn check_im_a_teapot(
...
+    }
+
+    let Some(license) = manifest.metadata().license.as_ref() else {
+        return Ok(());
+    };
+
+    // We must be a subset of crates.io's parse mode:
+    // https://github.com/rust-lang/crates.io/blob/bc421ae1/src/licenses.rs#L3-L8
+    // we don't want to allow something that would be rejected by crates.io
+    let Err(e) = spdx::Expression::parse_mode(license, spdx::ParseMode::STRICT) else {
```
epage, line 542: This isn't just checking the validity of the expression but whether the identifier is from a known list or not

https://github.com/EmbarkStudios/spdx/blob/e4eac6db2030a7deeca57cc7f5d0485598d8c9e5/src/lexer.rs#L253-L282

https://github.com/EmbarkStudios/spdx/blob/main/src/identifiers.rs#L18

I thought we only wanted to check syntax (and maybe have a separate lint for being a known identifier).  Unfortunately, it looks like there isn't a way to do that with `spdx`.

tests/testsuite/lints/invalid_spdx_license_expression.rs
weihanglo: The error message is from `spdx`, should we tweak them?

src/cargo/util/lints.rs
weihanglo: It would be good for checking the expression, and yeah spdx doesn't offer that 😞. It might be fine as normally people want known identifiers. The remaining questions are:

* If we have a new lint for checking known identifier, is it a breaking change if this one turns into a expression-only lint?
* Will we have per-line configuration so people can allowlist their identifiers?

tests/testsuite/lints/invalid_spdx_license_expression.rs
epage: If we want to treat `/` separately (which I thought was what was discussed), we'd likely have to do a parse fallback scheme (parse with strict, and if it fails, parse with `/` allowed and see if that fails).

src/cargo/util/lints.rs
epage: I think the main concern is people using a new identifier that cargo doesn't support yet.

It would be bad to have to turn off the syntax lint "temporarily" (if people remember to turn it back on).  These also seem like different levels of concern.  For example, we were talking about transitioning this to an error on an edition but we likely shouldn't do that for identifiers, leaving that as a warning or allow.

> If we have a new lint for checking known identifier, is it a breaking change if this one turns into a expression-only lint?

Lints are allowed to change, unsure how much though.

> Will we have per-line configuration so people can allowlist their identifiers?

Having a way to add new identifers through lint config might be a reasonable way to mitigate this but that also requires a change to `spdx` iiuc

It sounds like we'd need a change either way to `spdx` to stabilize this.

tests/testsuite/lints/invalid_spdx_license_expression.rs
weihanglo: I was keeping the first implementation simple.

We can certainly perform basic checks for common mistake like lowercase OR/AND/WITH, or slash.  If that is a requirement, can do that.

src/cargo/util/lints.rs
weihanglo: Changing `spdx` is definitely the right approach, totally agreed.

However, in practice,

* I doubt people will use new license identifier right after it gets into SPDX spec.
* Even if they do, it is likely pretty rare.
* Will crates.io fail an upload due to license parse error? Asked because if yes, then they will still fail during publish, and this lint must adhere with what crates.io is doing.

epage: > Will crates.io fail an upload due to license parse error? Asked because if yes, then they will still fail during publish, and this lint must adhere with what crates.io is doing.

We can be less strict than crates.io as we support multiple registries.

We also have different update requirements than crates.io (every 6 weeks, no updates to previous version vs crates.io effectively living at HEAD)

weihanglo: Just had a discussion with @Muscraft. Let's say we have two lints:

* invalid_spdx_license_expression
* invalid_spdx_license_identifier

And here we have dependency problem betweewn two lints dependency issue. If the expression is invalid, identifiers can't be meaningfully validated.

To fix this issue, we have some ideas:

* Have two lints. Run these lints in one function with two passes: first expression, then identifier.
  * Lint level: `expression > identifier` -> No issue. Just fire lints.
  * Lint level: `expression < identifier` -> We may need a special diagnostic saying "identifiers lint could not be checked due to invalid expression".
* Make just one lint but have its failure cases be configurable
  * For example, exposing [ParseMode](https://docs.rs/spdx/0.10.9/spdx/lexer/struct.ParseMode.html) but add config for invalid identifiers and else.
  * Clippy also has lints [with lots of configuration](https://rust-lang.github.io/rust-clippy/master/index.html#arbitrary_source_item_ordering)

epage: Personally, I'd go with the two lints in one function.

Might also be good to get someone's opinion who has dealt more with licensing, like @joshtriplett

weihanglo: See also the discussion in `spdx` crate with its author (also the author of `cargo-deny`): <https://github.com/EmbarkStudios/spdx/issues/82>.

joshtriplett: :+1: for the idea of separating "this is a valid SPDX expression" from "all the identifiers in this SPDX expression are known". I'd also love to see this allow for future extensibility, so that people can say "here's a license file, let this new term refer to it" and then write an expression that uses that term. (We should *probably* check with SPDX at some point about whether there's a reserved portion of the identifier namespace.)

weihanglo: Put some of my thoughts on the linked issue <https://github.com/EmbarkStudios/spdx/issues/82#issuecomment-3201758469>.

Added a bit more:

* It is not uncommon one allow a lint because it has false-positive / doesn't meet their their requirements.
* The more lints the better? Probably yes, but the more lints with similar functionalities, the more confusing (on both implementation and user side).
* Setting up lint rules is kinda showing a stance of the team. If we being strict on SPDX license, people follow, if we prove invalid identifiers that implies we think that should be a thing people want.
* Configuration are a bit better as people may want only OSI-approved or copyleft licenses.
* The initial idea of #2039 is for SPDX license syntax consistency. I feel like we are drifting to a broader licensing issue now 😆.

weihanglo: FWIW, the search result shows some repos already use `LicenseRef-` in their `package.license` field.

https://github.com/search?q=path%3ACargo.toml+%22LicenseRef-%22&type=code

weihanglo:
Close as our linting system is evolving and this couldn't catch up. Aslo the lint itself still need more refinement to decide the actual behavior, so not a good candidate for the first lint.

See further discussion in the original rust-lang/cargo#2039 and https://github.com/EmbarkStudios/spdx/issues/82

PR #15848, opened 2025-08-17
feat(unstable): Added `-Zbuild-dir-new-layout` unstable feature

ranger-ross:

Added `-Zbuild-dir-new-layout` unstable feature as part of #15010

I will open a follow up PR with an initial implementation as mentioned [here](https://github.com/rust-lang/cargo/issues/15010#issuecomment-3193725891) 

r? @weihanglo

### DIFF
--- a/src/doc/src/reference/unstable.md
+++ b/src/doc/src/reference/unstable.md
@@ -80,6 +80,7 @@ Each new feature described below should explain how to use it.
     * [feature-unification](#feature-unification) --- Enable new feature unification modes in workspaces
 * Output behavior
     * [artifact-dir](#artifact-dir) --- Adds a directory where artifacts are copied to.
+    * [build-dir-new-layout](#build-dir-new-layout) --- Enables the new build-dir filesystem layout
     * [Different binary name](#different-binary-name) --- Assign a name to the built binary that is separate from the crate name.
     * [root-dir](#root-dir) --- Controls the root directory relative to which paths are printed
 * Compile behavior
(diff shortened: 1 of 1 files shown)


ranger-ross:
oops, disregard the last push. I mistakenly pushed the wip layout implementation to the wrong branch.

Reverted that commit

src/doc/src/reference/unstable.md
```diff
@@ -1939,6 +1940,14 @@ The `-Zbuild-analysis` feature records and persists detailed build metrics
 enabled = true
 ```
 
+## build-dir-new-layout
+
+* Tracking Issue: [#15010](https://github.com/rust-lang/cargo/issues/15010)
+
+Enables the new build-dir filesystem layout.
```
weihanglo, line 1947: The new layout won't be that useful enough to stabilize on its own, right? Anyway the layout changing itself could lead to unexpected breakages to workflow making assumption of the layout, so agree it is better having its own unstable flag.

ranger-ross: > The new layout won't be that useful enough to stabilize on its own, right? Anyway the layout changing itself could lead to unexpected breakages to workflow making assumption of the layout, so agree it is better having its own unstable flag.

Yeah it seems unlikely that this alone would give users a tangible benefit, but I think there is value in allowing tool authors to test their cargo integration with the new layout.

---

Part of me is wondering if it would be a good idea to bundle the layout changes with moving the default `build-dir` location to inside of `CARGO_HOME` (or perhaps using XDG config https://github.com/rust-lang/cargo/issues/1734 ?)
My thought is that it might be easier to bundle both of these changes together to avoid churn in the default `build-dir` behavior.

weihanglo: Probably not? I feel like they are different kinds of churn. And XDG is always a hot topic that I am afraid of touching 😆.

epage: Agreed that these are independent changes.

My hope was to have fine grained locking be the feature to push this to being stabilized.

PR #15849, opened 2025-08-17
update tests to match lint message changes from rust-lang/rust#140794

karolzwolak:
Needed for [rust-lang/rust#140794](https://github.com/rust-lang/rust/pull/140794) (not merged yet).


tests/testsuite/warning_override.rs
```diff
@@ -75,7 +75,7 @@ fn rustc_caching_deny_first() {
 1 | fn main() { let x = 3; }
   |                 ^ [HELP] if this is intentional, prefix it with an underscore: `_x`
   |
-  = [NOTE] `#[warn(unused_variables)]` on by default
+  = [NOTE] `#[warn(unused_variables)]` (part of `#[warn(unused)]`) on by default
```
weihanglo: Use `[..]` wildcard so it matches both old and new diagnostics

```suggestion
  = [NOTE] `#[warn(unused_variables)]` [..]on by default
```

karolzwolak:
Update: 
* Use `SNAPSHOTS=overwrite` instead of manually blessing the outputs.
* Bump rustfix version
* Leave be [old lint message](https://github.com/karolzwolak/cargo/blob/ca2c81cd66151558cc95e4328fe614229a38f9d6/tests/testsuite/fix.rs#L2003) of [fake rustc](https://github.com/karolzwolak/cargo/blob/ca2c81cd66151558cc95e4328fe614229a38f9d6/tests/testsuite/fix.rs#L1950)

To be fair, the rustfix tests don't strcitly need the messages to be accurate to pass the tests, they tests the fix suggestions.
I thought the compiler output should be kept in sync in the tests, but I can revert it.

tests/testsuite/warning_override.rs
karolzwolak: That's interesting, if we were to add the wildcard everywhere, and revert the rustfix test output blessing, we could merge this without breaking tests.

We could merge the rustfix blessing after the rustc PR lands (or don't change it at all).

weihanglo: Yeah, this is the consequence of Cargo not being a subtree in rust-lang/rust: [#t-cargo > Turning cargo into a subtree](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Turning.20cargo.20into.20a.20subtree/with/527165966).

IMHO, I would update wherever needed to be compatible for CI to pass in the meanwhile. If rustfix snapshots don't really block CI, it is fine updating to the desired latest one now.

weihanglo: I still need to tweak some transitive dependency versions in Cargo's lockfile to unblock the submodule update though. It might take a while to land. Sorry for the quirky workflow

karolzwolak: Yeah, the workflow is a bit annoying, but no worries, not your fault. Cargo as a rust subtree would be nice here.

> IMHO, I would update wherever needed to be compatible for CI to pass in the meanwhile. If rustfix snapshots don't really block CI, it is fine updating to the desired latest one now.

I'm not sure I get what you mean. The rustfix doesn't block anything, its tests pass without any changes, it's the cargo tests that are failing.

weihanglo: No worries. Let's fix CI and merge this!

weihanglo (approved):
Thanks!

tests/testsuite/fix.rs
```diff
@@ -1863,9 +1863,13 @@ fn non_edition_lint_migration() {
...
-...
-  = [NOTE] `#[warn(unused_imports)]` on by default
-...
+  |                     ^^^^^^^^^^^^^^^^^^^
+  |
+  = [NOTE] `#[warn(unused_imports)]` [..]on by default
+
+[WARNING] `foo` (lib) generated 1 warning (run `cargo fix --lib -p foo` to apply 1 suggestion)
+[FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
+
```
weihanglo: Could you keep `...` so we dont excessively assert too many stuff?


```suggestion
...
  = [NOTE] `#[warn(unused_imports)]` [..]on by default
...
```

weihanglo: Also, would you mind cleaning up commit like squashing some of them, so it is less confusing when tracing back git history?

karolzwolak: Done. Btw why not just squash such PR into 1 commit either using github squash & merge or via some bot?

weihanglo: There are meanings in each commit. For example, we follow [a variant of the atomic commit pattern](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request) especially for bugfix. 

Take this <https://github.com/rust-lang/cargo/pull/14432/commits> as an example. The first commit added a test showing the current problematic behavior. It can be see as a minimal reproducible example to make sure you're fixing a really-existing bug. The next bugfix commit contains both the fix and the test change, so through the test diff you can clearly see the behavior changed.


I know it is a burden on contributors to clean up git history. I can actually do it by myself before merge if the PR allows maintainers to edit. However, sometimes PR author may feel offended like we took their contribution credits. Not sure what the right way is though.

weihanglo: Thanks for your help anyway!

karolzwolak: Yeah, I can see the appeal of atomic commits, but I'd argue that squashing PR into a single commits has its uses too. If we apply the 'atomic' principle to PRs themselves, the meaningful change is a single commit, and you don’t have to check which PR a commit belongs to. 

I just thought it’d be cool to combine both approaches, if commits from a single PR were grouped (like squashed), but you could still see the original commits that make up the PR - not just their names, but the actual revisions and changes too.

> I know it is a burden on contributors to clean up git history. I can actually do it by myself before merge if the PR allows maintainers to edit. However, sometimes PR author may feel offended like we took their contribution credits. Not sure what the right way is though.

I think leaving the authors to do the rebasing themselves is the way to go, since you won't really learn unless you do it yourself. Also github keeps the overwritten commits for some time, so it's not like someone can completely overwrite and lose their work.

Thanks too for the quick reviews.

PR #15851, opened 2025-08-17
chore: downgrade to libc@0.2.174

weihanglo:

There is incompatibility between rustix@1.0.8 and libc@0.2.175 that leads to compilation errors.

See

* https://github.com/bytecodealliance/rustix/issues/1496
* https://github.com/rust-lang/rust/pull/145478#issuecomment-3193494156

In rust-lang/rust, update the `src/tools/cargo` submodule accordingly, and run

```
cargo run --manifest-path src/ci/citool/Cargo.toml run-local dist-loongarch64-musl
```

With this downgrade it should compile.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ itertools = "0.14.0"
 jiff = { version = "0.2.15", default-features = false, features = [ "std" ] }
 jobserver = "0.1.33"
 lazycell = "1.3.0"
-libc = "0.2.175"
+libc = "0.2.174" # Please ensure in lockfile it stays as 0.2.174 until bytecodealliance/rustix#1496 resolved
 libgit2-sys = "0.18.2"
 libloading = "0.8.8"
 memchr = "2.7.5"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -64,7 +64,7 @@ itertools = "0.14.0"
 jiff = { version = "0.2.15", default-features = false, features = [ "std" ] }
 jobserver = "0.1.33"
 lazycell = "1.3.0"
-libc = "0.2.175"
+libc = "0.2.174" # Please ensure in lockfile it stays as 0.2.174 until bytecodealliance/rustix#1496 resolved
```
klensy, line 67: Why not pin version with `libc = "=0.2.174"` if it exactly what comment says?

weihanglo: because pinning may confuse the resolver and is not good for packages taking cargo as a dependency. Give the bug only happens on loongarch64 we should not block others build just for it.

weihanglo: I should have done this when seeing your comment <https://github.com/rust-lang/cargo/pull/15918> 😆

Issue #4247, opened 2017-07-04
Is there a way to skip some submodules?


BusyJay:
Currently, cargo will populate all git submodule automatically and recursively. But sometimes not all submodules are needed, populate those unneeded modules (specially some large git repos) are just a waste of time. So is there a way to tell cargo to skip those submodules?

alexcrichton:
Unfortunately we don't currently have a way of saying "exclude this submodule", but I could imagine interpreting the `exclude` key for this! (I don't think we do that today)

dtolnay:
I noticed this for `syn` – syn-0.11.11.crate is 62k but .cargo/git/checkouts/syn-6f4912dca572d3ab is 2.8 gigs.

BusyJay:
After #4837 is merged, it will be more slow to add large repository with a lot of optional submodules as dependency.

whereistejas:
Is using `exclude` still the recommended way to ignore git submodules?

dtolnay:
The recommended way is to not have submodules I believe. See https://github.com/dtolnay/syn/pull/183. Cargo fetches all submodules as part of checking out the repo, and reads the Cargo.toml *after* checking out the repo, so an `exclude` key in Cargo.toml wouldn't currently affect the submodule behavior.

https://github.com/rust-lang/cargo/issues/4247#issuecomment-312911391 proposes cloning the repo without submodules first, reading `exclude` from Cargo.toml, and fetching only submodules that are not excluded. As far as I know that is not implemented in Cargo which is why this issues remains open.

Muscraft:
Do you mean that it doesn't skip a submodule that is a submodule of a skipped submodule?

```
- git repo
  - submodule (skipped)
    - submodule (updated)
```

BusyJay:
No, I'm talking about skipping submodule that is a submodule of non-skipped submodule. For example, a native library may include tests framework as a submodule, so the layout becomes:
```
- rust binding repo
     - native library source
         - testing framework
```

Apparently, testing framework is not necessary for a rust binding library, but it will be cloned anyway.

Muscraft:
#10717 only makes sure that submodules aren't updated according to `.gitmodules`. It might be a good idea to open this back up since it is related to pulling in submodules (as you described), not updating them.

jsitnicki:
[...]

PR #15853, opened 2025-08-17
docs: mention how Cargo fetch git submodules

weihanglo:

Clarify how git submodules are handled in Cargo for `git` dependencies.

See https://github.com/rust-lang/cargo/issues/4247#issuecomment-3160707916

### DIFF
--- a/src/doc/src/reference/specifying-dependencies.md
+++ b/src/doc/src/reference/specifying-dependencies.md
@@ -301,6 +301,20 @@ See [Multiple locations](#multiple-locations) section below for detailed explana
 > locations](#multiple-locations) section for a fallback alternative for `git`
 > and `path` dependencies.
 
+### Git submodules
+
+When cloning a `git` dependency,
+Cargo automatically fetches its submodules recursively
+so that all required code is available for the build.
+
+To skip updates for a specific submodule,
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/specifying-dependencies.md
```diff
@@ -301,6 +301,20 @@ See [Multiple locations](#multiple-locations) section below for detailed explana
...
+
+When cloning a `git` dependency,
+Cargo automatically fetches its submodules recursively
+so that all required code is available for the build.
+
+To skip updates for a specific submodule,
+Cargo respects the [`submodule.<name>.update`] setting in `.gitmodules`.
+Setting it to `none` disables updates for that submodule if it isn't needed for the build.
+This is usually set in the repository itself,
+so changes require access to the dependency's repo.
```
epage, line 314: The first paragraph talks in terms of behavior (fetch) while the second paragraph talks in terms of a git submodule specific term (update) that I worry may confuse users.

What do you think of:
```suggestion
To skip fetching submodules unrelated to the build, you can set [`submodule.<name>.update = none`] in the dependency repo's `.gitmodules`.
This requires write access to the repo and will disable submodule updates more generally.
```

weihanglo: This seems a bit more concise and understandable!

PR #15855, opened 2025-08-18
docs(profile): revert wrong statement of lto options' optimization

weihanglo:
There is no clear winner between fat LTO and ThinLTO, so don't say "from most to least optimizing" here.

See https://github.com/rust-lang/cargo/pull/15841#discussion_r2281037386


ehuss:
Resolver error #6258, retrying.

PR #15874, opened 2025-08-21
Added better filesystem layout testing harness

ranger-ross:

The goal is to make filesystem layout tests easier to understand and review.
The `build-dir` test's have a [helper fns](https://github.com/rust-lang/cargo/blob/master/tests/testsuite/build_dir.rs#L720) that have imperative checks. While this works, its not easy to see what its validating at a glance.

My thought is to reuse the snapbox infrastructure we already have for the cargo ui tests. 
The prototype in this PR generates a file tree (like the unix tree command) and uses snapbox to compare against a snapshot. This makes the layout tests much easier to read and failures are much more obvious what what went wrong.

### Implementation Details

* The `LayoutTree` struct handles comparison logic (as opposed to using snapbox directly)
  * This avoids the issues initial implementation (ordering and platform specific files, see full details below)
  * If the comparison fails, 2 human readable tree's are generated and snapbox is used to display the diffs. 
    * Note: These diffs are in-memory thus, do not support `SNAPSHOTS=overwrite`. This is arguably good as often the overwrite would be incorrect as the snapshots are platform specific.
* Snapshots would support conditional files based on the target os / platform env
  * Files/directories may be suffixed with `[target_platform=<target-platform>]`
  * Multiple platforms may be specified separated by comma

<details><summary>initial implementation issues</summary>

However there are some problems with the current implementation that limit the usefulness of it.

1. Different platforms have different files which cause some tests to fail
   * Examples
      * lib prefix/suffixes based on platform
      * `dSYM` on macos
   * One idea I have would be to have a cfg suffix after file name in the tree like so
     * `└── foo-[HASH].dSYM [cfg(target_os = "macos")]`
     * This would also require rethinking the "tree lines" (`└`, `├`, `│`) to handle optional files.
2. When dealing with build scripts there are multiple `target/<profile>/build/pkg-[HASH]` directories. The hash changes the order of the directories when generating the tree making snapshots inconsistent.
   * We have redactions to handle replacing the hash with a placeholder `[HASH]`, but this does not help with the order.

</details> 

---

Fun fact: These changes were written and tested at [Rust Forge 2025](https://rustforgeconf.com/) in New Zealand. :new_zealand: :partying_face:

### DIFF
--- a/src/doc/contrib/src/tests/writing.md
+++ b/src/doc/contrib/src/tests/writing.md
@@ -64,6 +64,16 @@ test.
   - See [`support::compare`] for an explanation of the string pattern matching.
     Patterns are used to make it easier to match against the expected output.
 
+#### Filesystem layout testing
+
+Tests often to need to verify Cargo created/removed files.
+The `CargoPathExt` trait (implemented by `Path` and `PathBuf`) provides a `assert_dir_layout()` to verify the files in a directory (including nested directories).
+This takes a snapshot of file paths for the given directory and asserts that all files are present and no new files have been created.
+This function also takes a list of patterns to ignore from the snapshot to make working with platform specific files easier.
+
+Note: You will commonly need to call `unordered()` before passing your snapshot to deal with platform differences like binaries having `.exe` on Windows.
+`assert_build_dir_layout` is a more specialized version of `assert_dir_layout()` that is automatically unordered and ignores common platform specific files designed for the Cargo build cache.
+
 #### Testing Nightly Features
 
 If you are testing a Cargo feature that only works on "nightly" Cargo, then


weihanglo:
Haven't look at the change, but thanks a lot for improving the test infrastructure!

crates/cargo-test-support/src/compare.rs
```diff
@@ -229,6 +229,11 @@ fn add_regex_redactions(subs: &mut snapbox::Redactions) {
     .unwrap();
     subs.insert("[HASH]", regex!(r"/[a-z0-9\-_]+-(?<redacted>[0-9a-f]{16})"))
         .unwrap();
+    // Match multi-part hashes like 06/b451d0d6f88b1d used in directory paths
+    subs.insert("[HASH]", regex!(r"/(?<redacted>[a-f0-9]{2}\/[0-9a-f]{14})"))
+        .unwrap();
+    subs.insert("[HASH]", regex!(r"[a-z0-9]+-(?<redacted>[a-f0-9]{16})"))
```
epage, line 236: Whats this redaction covering?

ranger-ross: this is handling the hashes in file/dir names in the fingerprint/build dirs
```
└── debug
    ├── .cargo-lock
    ├── .fingerprint
    │   └── foo-[HASH]
    │       ├── bin-foo
    │       ├── bin-foo.json
    │       ├── dep-bin-foo
    │       └── invoked.timestamp
```

epage: Could you add a corresponding comment?

ranger-ross: although I think the first multi-part hash is incorrect now. (I added that when I was still experimenting with a flat file paths)

ranger-ross:
I updated the implementation to fix the flaws of the initial design.
See the PR description for the details

The build-dir tests now pass on all of the platforms tested in CI.
I think the changes are now in a state worthy of review.
Let me know what you think

crates/cargo-test-support/src/paths.rs
```diff
@@ -236,6 +241,20 @@ impl CargoPathExt for Path {
...
+    fn verify_file_layout(&self, expected: impl AsRef<str>) {
+        let actual_layout = LayoutTree::from_path(&self).unwrap();
+
+        let expected_layout = LayoutTree::parse(expected.as_ref());
+
+        if !actual_layout.matches_snapshot(&expected_layout) {
+            let actual_snapshot = actual_layout.to_string();
+            let expected_snapshot = expected_layout.to_string();
+
+            compare::assert_e2e().eq(actual_snapshot, expected_snapshot);
```
epage: > * The `LayoutTree` struct handles comparison logic (as opposed to using snapbox directly)
>       
>   * This avoids the issues initial implementation (ordering and platform specific files, see full details below)
>    * If the comparison fails, 2 human readable tree's are generated and snapbox is used to display the diffs.
>         
>     * Note: These diffs are in-memory thus, do not support `SNAPSHOTS=overwrite`. This is arguably good as often the overwrite would be incorrect as the snapshots are platform specific.
> 
> * Snapshots would support conditional files based on the target os / platform env
>       
>   * Files/directories may be suffixed with `[target_platform=<target-platform>]`
>   * Multiple platforms may be specified separated by comma

Our general aim should be to support snapshotting wherever possible.

Also, tracking of platform-specific details is rough for contributors
- Likely won't discover until the PR is posted
- Likely don't have access to a machine, so must do a slow iteration by pushing and seeing CI's result

Unsure f there is something better but felt this was at least worth specifically discussing

tests/testsuite/build_dir.rs
```diff
@@ -267,17 +476,53 @@ fn examples_should_output_to_build_dir_and_uplift_to_target_dir() {
...
+    │       ├── example-foo
+    │       ├── example-foo.json
+    │       └── invoked.timestamp
+    ├── build
+    ├── deps
+    ├── examples
+    │   ├── foo.d                               [target_platform=windows-msvc]
+    │   ├── foo.pdb                             [target_platform=windows-msvc]
+    │   ├── foo[EXE]                            [target_platform=windows-msvc]
+    │   ├── foo-[HASH].d                        [target_platform=macos,linux,windows-gnu]
```
epage: I feel like some of the benefit of `tree` output is lost with the conditionals breaking up the output

crates/cargo-test-support/src/paths.rs
ranger-ross: Yes I totally agree with you.

One idea I have to only check for files in the snapshot and ignore other files. Meaning if a new file was is created a test would not fail. So tests would be able to ignore many of the platform specific files that are not relevant to the test.
But this comes with the trade off of potentially not including a file that is important.

I think I lean toward the current implementation that explicitly matches the file tree structure

tests/testsuite/build_dir.rs
ranger-ross: Yeah, it was not very bad until I realized `windows-msvc` does not include hashes in many places that linux, macos, and winddows-gnu does. I felt this was still better than duplicating the tree with  `#[cfg(target_os = "linux")]`.

ranger-ross:
I rebased to resolve the conflict with #15910 

I'm waiting to see what we decide to do on this PR before writing tests for the build-dir layout changes.

If we feel this is not a meaningful improvement, I am happy to close this and continue with the existing approach in `testsuite/build-dir.rs`

crates/cargo-test-support/src/paths.rs
epage: Two alternative ideas:
- Have a function that enumerates files in a directory as relative-path-per-line
  - a snapshot assert can then be run against that with the expected results having a `...` inside it and `.unordered()` called on it
  - a parameter could take a glob to ignore
- Have a low level function that accepts a glob to ignore and enumerates files as a tree.  A higher level assert function could assume a specific policy for globs and compare against a snapshot

The problem comes in with
```
[target_platform=windows-msvc]
    │   ├── foo[EXE]
[target_platform=macos,linux,windows-gnu]
    │   ├── foo-[HASH][EXE]   
```

epage: The current tests handle this by using globs

epage: The problem with the existing tests doing `format!("build-dir/debug/deps/foo*{EXE_SUFFIX}")` and if we did `foo[..][EXE]` is that it over matches.  On Unix-like systems, this could match any file and not just the binary.

epage: Maybe we limp along for now with `foo[..][EXE]` as #15947 would unblock removing `-Cextra-filename` from more places, see  https://github.com/rust-lang/cargo/pull/15947#issuecomment-3303317501

As for why msrv doesn't include the hash, see
https://github.com/rust-lang/cargo/blob/e0e1a6e27f05102bd8590a96e703bdb49ff02b14/src/cargo/core/compiler/build_runner/compilation_files.rs#L835-L865

ranger-ross:
In the latest push I made some changes based on the feedback so far:

* Renamed `verify_file_layout()` to `assert_file_layout()`
* Added an `assert_file_layout_with_ignored_paths()` fn that accepts a list of paths that will be ignored
* `assert_file_layout()` adds a predefined list of paths (see: [`default_ignored_paths`](https://github.com/rust-lang/cargo/pull/15874/files#diff-aa8fbecc067fd53b21cfd6c545e92af433478017a604bd9cce844ef0e30c8dc5R339))
  * Currently ignoring platform specific debug symbols on windows and macos that generally get in the way during testing 
* Used wildcard matches (e.g. `foo[..][EXE]`) as mentioned in [this comment](https://github.com/rust-lang/cargo/pull/15874#discussion_r2355759807) 

With these changes the tests are quite a bit cleaner. :) 

I left the `[target_platform=...]` functionality for now as I think it could be potentially useful in the future and is only about 15 lines of code.

epage:
Part of my hope with the rework is we wouldn't need to parse; we generate and then assert. What is left that we still need parsing?
 
> I left the `[target_platform=...]` functionality for now as I think it could be potentially useful in the future and is only about 15 lines of code.

I'd rather stick to what we need than carry around a burden of unused features that exist for speculative purposes, especially if they over encourage the lack of snapshotting.

epage:
Also, while the tree layout works well for inspecting a test, how well does that work for reading the diff and (in rare cases) updating it by hand?

src/doc/contrib/src/tests/writing.md
```diff
@@ -64,6 +64,21 @@ test.
   - See [`support::compare`] for an explanation of the string pattern matching.
     Patterns are used to make it easier to match against the expected output.
 
+#### Filesystem layout testing
+
+Tests often to need to verify Cargo created/removed files.
+The `CargoPathExt` trait (implmented by `Path` and `PathBuf`) provides a `assert_file_layout()` to verify a file system tree against a snapshot. The snapshot format is very similar to the unix `tree` command output.
+
+Files vary across operating systems, for example `.pdb` files on Windows.
```
weihanglo: I wonder instead this platform specific string we should just have different snapshots for different platforms. 

Or maybe we do this kind of tree layout test assertions only on one platform, presumably Linux, which is the most available platforms. For example, I don't have a Windows physical machine. Some don't have macbook. But everyone can access a Linux container :)

Also, some tier 2 and 3 platform maintainers are running Cargo tests in their CI. While they already patched out some tests, we may want to be nice not to fail their CI more.

ranger-ross:
> Part of my hope with the rework is we wouldn't need to parse; we generate and then assert. What is left that we still need parsing?

The issue is the file hashes change the order of files/dirs, resulting tests having diff file orders across platforms.
```
    ├── foo-[HASH]
    │        ├── dep-test-integration-test-foo
    │        ├── invoked.timestamp
    │        ├── test-integration-test-foo
    │        └── test-integration-test-foo.json
    └── foo-[HASH]
            ├── bin-foo
            ├── bin-foo.json
            ├── dep-bin-foo
            └── invoked.timestamp
```
Using a tree layout order of lines matters but not the order of the dirs/files.
That is why the `LayoutTree` is used.


> I'd rather stick to what we need than carry around a burden of unused features that exist for speculative purposes, especially if they over encourage the lack of snapshotting.

Sure I can drop this functionality. Its pretty easy to add back if we need it in the future.

> Also, while the tree layout works well for inspecting a test, how well does that work for reading the diff and (in rare cases) updating it by hand?

I found the diffs when the test are wrong are pretty good. 
The 2 pain points I had were, 
* `[EXE]` being removed by mistake in the diff as I was running on linux. Though this is not related to the tree layout.
* At times it was a bit awkward to change a `├` to a `└` and vise versa. Though this was usually just a matter of copy/pasting from another test.

src/doc/contrib/src/tests/writing.md
ranger-ross: > I wonder instead this platform specific string we should just have different snapshots for different platforms.

I considered that, but it would end up bloating testing and I think it would be quite painful to write tests if you do not have access to all platforms. (and even if you did, it would probably still be tedious)

weihanglo: Yeah that is why another alternative is testing only on Linux. Anything we really want to test platform differences here?

ranger-ross: As someone that develops primarily on linux, I am okay with this though not sure about other contributors on Windows.
Though I suppose, its already a problem since they need to think about linux and macos in tests already.

ranger-ross: > Anything we really want to test platform differences here?

Not really. We are ignoring the most of the platform specific things like debug symbols because they clutter up the tests.
At a high level, the `build-dir.rs` tests are trying to verify the file layout implementation.
In a perfect world, we would verify all files in the tree, but in practice its difficult to create such a test that is easy to write and maintainable.

epage:
> The issue is the file hashes change the order of files/dirs, resulting tests having diff file orders across platforms.

At least for myself, this would have me lean towards having a flat list of files and directories.  I'm not too much of a fan of the complexity involved in having a bespoke format and then having to parse it.  This also loses out on several aspects of our snapshot tooling.

ranger-ross:
> At least for myself, this would have me lean towards having a flat list of files and directories. I'm not too much of a fan of the complexity involved in having a bespoke format and then having to parse it. This also loses out on several aspects of our snapshot tooling.

I think this is fair enough. 
Sorry, I think I was over complicating this. 
I was worried about the ordering of a flat list but learning about `.unordered()` I think that helps solve the issue. 
I updated the implementation to use a flat structure. Seems like this approach resolves many of the small issues with the tree structure and the implementation is significantly simpler.

Let me know what y'all think

ranger-ross:
After some further thought I decided to completely remove `assert_build_dir_layout` and `assert_artifact_dir_layout` in favor of the new `assert_file_layout`. I feel the are largely redundant and file tree snapshots feel like a better approach. 

(And removing them means I don't need to update these methods in https://github.com/rust-lang/cargo/pull/15947)

crates/cargo-test-support/src/paths.rs
```diff
@@ -236,6 +249,37 @@ impl CargoPathExt for Path {
...
+    }
+
+    #[track_caller]
+    fn assert_file_layout_with_ignored_paths(
+        &self,
+        expected: impl snapbox::IntoData,
+        ignored_paths: &[PathBuf],
+    ) {
+        let assert = assert_e2e();
+        let actual = WalkDir::new(self)
```
epage, line 267: Can we enable `sort_by_file_name()`?  Having a more consistent layout can improve the quality of the snapshots and make them easier to browse

crates/cargo-test-support/src/paths.rs
```diff
@@ -236,6 +249,37 @@ impl CargoPathExt for Path {
...
+                    let ignored = ignored.to_str().unwrap();
+                    if match_contains(&ignored, file, &assert.redactions()).is_ok() {
+                        return false;
+                    }
+                }
+                return true;
+            })
+            .join("\n");
+
+        assert.eq(format!("{actual}\n"), expected.unordered());
```
epage: shouldn't it be up to the caller to decide to do `unordered()`, especially if we sort?

crates/cargo-test-support/src/paths.rs
```diff
@@ -290,6 +349,20 @@ where
     }
 }
 
+/// The default paths to ignore when [`CargoPathExt::assert_file_layout`] is called
+fn default_ignored_paths() -> Vec<PathBuf> {
```
epage: Is there a reason this returns `PathBuf`s?  That  makes clear that these are path-like but they aren't ever paths but patterns and this requires the callers to do `unwrap` to undo this.

Maybe the name should be clarified to be about ignored patterns.

crates/cargo-test-support/src/paths.rs
ranger-ross: I was considering this. 
I ended up adding it inside of this method to reduce boilerplate since nearly all callers for the `build-dir` tests were going to need `.unordered()` due to the hashes.

Though I don't mind pulling this out if you think that would be better

crates/cargo-test-support/src/paths.rs
ranger-ross: seems reasonable, perhaps `fn default_ignored_path_patterns() -> Vec<String>`

crates/cargo-test-support/src/paths.rs
epage: For this low level of a function, I don't think its appropriate.

One option could be we document it and require the caller to set it.  We have other places where callers have to regularly set it.

I was already a little concerned with the names `assert_file_layout`, `assert_file_layout_with_ignored_paths`, and `default_ignored_paths`.  Another options is we specialize these more
- `default_ignored_paths` -> `build_dir_ignored_path_patterns`
- `assert_file_layout` -> `assert_build_dir_layout`
- `assert_file_layout_with_ignored_paths` -> `assert_dir_layout`

With `assert_build_dir_layout` adding `.unordered()` and `assert_dir_layout` having a comment that hashes in paths may cause non-deterministic ordering and `.unordered()` may be needed in those cases.

ranger-ross: sure I like the idea of specialization.
It keeps the tests clean and while removing any specialized build-dir logic from `assert_dir_layout`

crates/cargo-test-support/src/paths.rs
```diff
@@ -236,6 +245,37 @@ impl CargoPathExt for Path {
             });
         }
     }
+
+    #[track_caller]
+    fn assert_build_dir_layout(&self, expected: impl snapbox::IntoData) {
+        self.assert_dir_layout(expected.unordered(), &build_dir_ignored_path_patterns());
```
epage, line 257: This should have a comment about why `.unordered()` is needed for build dirs

epage (approved):
Thanks!

PR #15879, opened 2025-08-22
 Fix Non-actionable error messages

ojuschugh1:
…space publish rate limiting

### DIFF
--- a/tests/testsuite/publish.rs
+++ b/tests/testsuite/publish.rs
@@ -4402,3 +4402,124 @@ See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for
 "#]])
         .run();
 }
+
+#[cargo_test]
+fn workspace_publish_rate_limit_error() {
+    // This test demonstrates the current behavior when workspace publishing
+    // fails due to rate limiting. The error message doesn't indicate which
+    // package failed or what packages remain to be published.
+
+    let registry = registry::RegistryBuilder::new()
+        .http_api()
+        .http_index()
+        .add_responder("/api/v1/crates/new", |_req, _| {
+            // For simplicity, let's just return rate limit error for all requests
+            // This simulates hitting rate limit during workspace publish
+            Response {
+                code: 429,
+                headers: vec!["Retry-After: 3600".to_string()],
+                body: format!(
+                    "You have published too many new crates in a short period of time. Please try again after Fri, 18 Jul 2025 20:00:34 GMT or email help@crates.io to have your limit increased."
+                ).into_bytes(),
+            }
+        })
+        .build();
+
+    let p = project()
+        .file(
+            "Cargo.toml",
+            r#"
+            [workspace]
+            members = ["package_a", "package_b", "package_c"]
+
+            [package]
+            name = "workspace_root"
+            version = "0.0.0"
+            edition = "2015"
+            license = "MIT"
+            description = "workspace root"
+            repository = "https://github.com/test/workspace"
+            publish = false
+            "#,
+        )
+        .file("src/lib.rs", "")
+        .file(
+            "package_a/Cargo.toml",
+            r#"
+            [package]
+            name = "package_a"
+            version = "0.1.0"
+            edition = "2015"
+            license = "MIT"
+            description = "package a"
+            repository = "https://github.com/test/package_a"
+            "#,
+        )
+        .file("package_a/src/lib.rs", "")
+        .file(
+            "package_b/Cargo.toml",
+            r#"
+            [package]
+            name = "package_b"
+            version = "0.1.0"
+            edition = "2015"
+            license = "MIT"
+            description = "package b"
+            repository = "https://github.com/test/package_b"
+            "#,
+        )
+        .file("package_b/src/lib.rs", "")
+        .file(
+            "package_c/Cargo.toml",
+            r#"
+            [package]
+            name = "package_c"
+            version = "0.1.0"
+            edition = "2015"
+            license = "MIT"
+            description = "package c"
+            repository = "https://github.com/test/package_c"
+            "#,
+        )
+        .file("package_c/src/lib.rs", "")
+        .build();
+
+    // This demonstrates the current non-actionable error message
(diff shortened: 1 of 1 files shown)


tests/testsuite/publish.rs
```diff
@@ -2461,7 +2461,7 @@ fn api_other_error() {
 [PACKAGING] foo v0.0.1 ([ROOT]/foo)
 [PACKAGED] 4 files, [FILE_SIZE]B ([FILE_SIZE]B compressed)
 [UPLOADING] foo v0.0.1 ([ROOT]/foo)
-[ERROR] failed to publish to registry at http://127.0.0.1:[..]/
+[ERROR] failed to publish `foo` v0.0.1 to registry at http://127.0.0.1:[..]/
```
epage: Like in #15765, could you split adding the package name to the error in a commit before adding the unpublished packages?

src/cargo/ops/registry/publish.rs
```diff
@@ -641,9 +682,24 @@ fn transmit(
...
+                context
+            )
+        } else {
+            format!(
+                "failed to publish `{}` v{} to registry at {}",
+                pkg.name(),
+                pkg.version(),
+                registry.host()
+            )
+        }
```
epage: but we do for a lot of errors is include the separator for optional context in that context and use an empty string instead of `None`, so we don't need a conditional like this but always include the context.

tests/testsuite/publish.rs
```diff
@@ -4507,7 +4507,7 @@ fn workspace_publish_rate_limit_error() {
 [COMPILING] package_c v0.1.0 ([ROOT]/foo/target/package/package_c-0.1.0)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 [UPLOADING] package_a v0.1.0 ([ROOT]/foo/package_a)
-[ERROR] failed to publish to registry at http://127.0.0.1:[..]/
+[ERROR] failed to publish `package_a` v0.1.0 to registry at http://127.0.0.1:[..]/; the following crates have not been published yet: package_b v0.1.0, package_c v0.1.0
```
epage: In #15765, the additional packages were on a separate line
```
[ERROR] failed to publish package 'a' to registry at http://127.0.0.1:[..]/

Remaining packages to publish: b v0.1.0 and c v0.1.0
```
Likely what this should look like is
```
[ERROR] failed to publish `package_a` v0.1.0 to registry at http://127.0.0.1:[..]/

note: the following crates have not been published yet:
  package_b v0.1.0
  package_c v0.1.0
```

src/cargo/ops/registry/publish.rs
```diff
@@ -210,7 +212,7 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
         // `b`, and we uploaded `a` and `b` but only confirmed `a`, then on
         // the following pass through the outer loop nothing will be ready for
         // upload.
-        for pkg_id in plan.take_ready() {
+        for pkg_id in plan.take_ready().into_iter() {
```
epage: Why is this change needed?

src/cargo/ops/registry/publish.rs
```diff
@@ -236,15 +238,52 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
...
+                    })
+                    .collect();
+                // Also include any packages that are still waiting for confirmation
+                for id in to_confirm.iter().filter(|id| **id != pkg_id) {
+                    let pkg = &pkg_dep_graph.packages[&id].0;
+                    let entry = format!("{} v{}", pkg.name(), pkg.version());
+                    if !remaining.contains(&entry) {
+                        remaining.push(entry);
+                    }
+                }
```
epage: Why do we include these?  They have been published, we just don't know if the server has fully finished processing it yet or not.

As is, the current message is misleading.  It says a package hasn't been published yet.  As a user, that means I need to go in and publish it but then I'll get an error saying its published.

src/cargo/ops/registry/publish.rs
```diff
@@ -236,15 +238,52 @@ pub fn publish(ws: &Workspace<'_>, opts: &PublishOpts<'_>) -> CargoResult<()> {
...
+            // Prepare workspace context for error message
+            let workspace_context = if original_packages.len() > 1 {
+                let mut remaining: Vec<_> = original_packages
+                    .iter()
+                    .filter(|id| **id != pkg_id)
+                    .map(|id| {
+                        let pkg = &pkg_dep_graph.packages[&id].0;
+                        format!("{} v{}", pkg.name(), pkg.version())
+                    })
+                    .collect();
```
epage: If I'm reading this correctly, this is grabbing the list of all packages, including those that have already been published (except for the current package)

weihanglo:
I don't understand why we have a couple of similar PR open for the same issue. Shouldn't we consolidate all the discussions in a single place?

src/cargo/ops/registry/publish.rs
ojuschugh1: The change from `for pkg_id in plan.take_ready()` to `for pkg_id in plan.take_ready().into_iter()` was needed because of a Rust type inference issue.

The `take_ready()` method returns a `BTreeSet<PackageId>`, and Rust couldn't automatically determine which iterator implementation to use (there are multiple: `IntoIterator`, `Iter`, `IterMut`). By explicitly calling `.into_iter()`, we tell Rust to use the consuming iterator, which is what we want since `take_ready()` takes ownership of the set.

This makes the code more explicit and avoids ambiguity in type inference.

epage: I just did a checkout of your branch and deleted the `into_iter` and it built just fine.  In fact, [`for` is defined as calling `IntoIterator::into_iter`](https://doc.rust-lang.org/reference/expressions/loop-expr.html#r-expr.loop.for.intro)

ojuschugh1: I will revert this change

epage (approved):

PR #15880, opened 2025-08-23
 test: avoid hardcoded target spec json

weihanglo:
Read a real target spec JSON so we no longer need to hardcode
a target spec JSON here.
Cargo itself should not care about the spec schema.

Let's stop bothering compiler contributors.

* https://github.com/rust-lang/rust/pull/144443
* https://github.com/rust-lang/rust/pull/145764#issuecomment-3216104449
* [#t-compiler > Is it possible to add a commit to a submodule? @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Is.20it.20possible.20to.20add.20a.20commit.20to.20a.20submodule.3F/near/535865215)

### DIFF
--- a/crates/cargo-test-support/src/lib.rs
+++ b/crates/cargo-test-support/src/lib.rs
@@ -50,6 +50,7 @@ use std::fs;
 use std::os;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Output};
+use std::sync::LazyLock;
 use std::sync::OnceLock;
 use std::thread::JoinHandle;
 use std::time::{self, Duration};
@@ -1262,6 +1263,28 @@ pub fn basic_lib_manifest(name: &str) -> String {
     "#,
         name, name
     )
+}
+
+/// Gets a valid target spec JSON from rustc.
+///
+/// To avoid any hardcoded value, this fetches `x86_64-unknown-none` target
+/// spec JSON directly from `rustc`, as Cargo shouldn't know the JSON schema.
+pub fn target_spec_json() -> &'static str {
+    static TARGET_SPEC_JSON: LazyLock<String> = LazyLock::new(|| {
+        let json = std::process::Command::new("rustc")
+            .env("RUSTC_BOOTSTRAP", "1")
+            .arg("--print")
+            .arg("target-spec-json")
+            .arg("-Zunstable-options")
+            .arg("--target")
+            .arg("x86_64-unknown-none")
+            .output()
+            .expect("rustc --print target-spec-json")
+            .stdout;
+        String::from_utf8(json).expect("utf8 target spec json")
+    });
+
+    TARGET_SPEC_JSON.as_str()
 }
 
 struct RustcInfo {
--- a/tests/build-std/main.rs
+++ b/tests/build-std/main.rs
@@ -20,7 +20,13 @@
 
 #![allow(clippy::disallowed_methods)]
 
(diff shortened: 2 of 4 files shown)


bjorn3:
Instead of having a specific target json file would an option be to ask rustc for the target json file of an existing target and just remove the is-builtim flag before passing it back to cargo? That way it will remain correct even if the target json spec format changes in any way other than renaming the is-builtin field.

weihanglo:
@bjorn3 good idea, though where is the `is-builtin` flag you were talking about?

```console
$ rustc +nightly --print target-spec-json -Zunstable-options --target wasm32-unknown-unknown
{
  "arch": "wasm32",
  "binary-format": "wasm",
  "crt-objects-fallback": "true",
  "data-layout": "e-m:e-p:32:32-p10:8:8-p20:8:8-i64:64-i128:128-n32:64-S128-ni:1:10:20",
  "dll-prefix": "",
  "dll-suffix": ".wasm",
  "dynamic-linking": true,
  "eh-frame-header": false,
  "emit-debug-gdb-scripts": false,
  "exe-suffix": ".wasm",
  "generate-arange-section": false,
  "has-thread-local": true,
  "is-like-wasm": true,
  "limit-rdylib-exports": false,
  "linker": "rust-lld",
  "linker-flavor": "wasm-lld",
  "linker-is-gnu": false,
  "lld-flavor": "wasm",
  "llvm-target": "wasm32-unknown-unknown",
  "max-atomic-width": 64,
  "metadata": {
    "description": "WebAssembly",
    "host_tools": false,
    "std": true,
    "tier": 2
  },
  "only-cdylib": true,
  "os": "unknown",
  "panic-strategy": "abort",
  "pre-link-args": {
    "wasm-lld": [
      "-z",
      "stack-size=1048576",
      "--stack-first",
      "--allow-undefined",
      "--no-demangle",
      "--no-entry"
    ],
    "wasm-lld-cc": [
      "-Wl,-z",
      "-Wl,stack-size=1048576",
      "-Wl,--stack-first",
      "-Wl,--allow-undefined",
      "-Wl,--no-demangle",
      "--target=wasm32-unknown-unknown",
      "-Wl,--no-entry"
    ]
  },
  "relocation-model": "static",
  "singlethread": true,
  "target-family": [
    "wasm"
  ],
  "target-pointer-width": "32",
  "tls-model": "local-exec"
}
```

weihanglo:
I should go ahead and remove _all" hardcoded target spec jsons.

bjorn3:
I guess rustc itself nowadays already filters out the is-builtin field before printing.

weihanglo:
```
Run rustup update --no-self-update stable
info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
error: failed to download file error=Reqwest(reqwest::Error { kind: Request, url: "https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256", source: hyper_util::client::legacy::Error(Connect, Custom { kind: Other, error: Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" } }) })
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256' to '/home/runner/.rustup/tmp/y2jjoktwl9rzp757_file': error downloading file: error sending request for url (https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256): client error (Connect): Connection reset by peer (os error 104)
```

Retried an succeeded. Merging.

PR #15881, opened 2025-08-24
docs(package): expand the manifest nromalization

### LINKED ISSUE #15836: `cargo publish` stripping `hints` section from Cargo.toml
### Problem

We recently added `hints.mostly-unused = true` to several of the aws-sdk-rust crates to improve compilation time (ex: [aws-sdk-ec2](https://github.com/awslabs/aws-sdk-rust/blob/4fcc47b2babc2d3c78c9e7861d75328a6315fba5/sdk/ec2/Cargo.toml#L146-L147)). A [user reported](https://github.com/awslabs/aws-sdk-rust/issues/113#issuecomment-3182336182) that compiling with the new `-Zprofile-hint-mostly-unused` option did not improve their compilation times and that they did not see a `hints` section when looking at the crate in the Cargo cache. 

When looking at the [published source](https://docs.rs/crate/aws-sdk-ec2/1.158.0/source/Cargo.toml#149) on docs.rs (which I assume is the same as that on crates.io?) I also do not see a `hints` section. Running `cargo publish --dry-run` and looking at the generated package in `target/package` I can confirm that the `hints` section is removed from the package.

### Steps

```sh
git clone --depth 1 https://github.com/awslabs/aws-sdk-rust.git;
cd aws-sdk-rust/sdk/ec2;

# Finds hints
cat Cargo.toml | grep hints;
cargo publish --dry-run;

# Does not find hints
cat ../../target/package/aws-sdk-ec2-1.158.0/Cargo.toml | grep hints;
```

### Possible Solution(s)

I suspect that `cargo publish` is stripping out unrecognized sections in Cargo.toml, and we are not publishing with nightly and the `-Zprofile-hint-mostly-unused` option. This should likely be more permissive in allowing sections through. 

### Notes

_No response_

### Version

```text

```

weihanglo:

Resolves #15836

This also temporarily removes the mention of restriction in `git` and
`path` dependencies without a `version` key, and `dev-dependencies`.
Those are mentioned in the "Specifying Dependencies" chapter already.

(To be honest, I am not sure whether they should mention here or there)

Also added a note about the form of the normalized manifest is not "stable".

Check whether if we've exposed implementation details that users don't need to know, like `license-file`? Though that one Cargo already emitted a warning if happened.

### DIFF
--- a/src/doc/man/cargo-package.md
+++ b/src/doc/man/cargo-package.md
@@ -27,6 +27,8 @@ stored in the `target/package` directory. This performs the following steps:
     - Fields not recognized by the current Cargo version are dropped.
     - Certain paths, such as `package.readme` and `package.license-file`,
       are rewritten if they point outside the package root.
+    - **Note**: the normalized `Cargo.toml` is intended to be stable,
+      but its exact form may change between Cargo versions.
 
 2. Create the compressed `.crate` file.
     - The normalized `Cargo.toml` is included, and the original manifest is
--- a/src/doc/src/commands/cargo-package.md
+++ b/src/doc/src/commands/cargo-package.md
@@ -22,6 +22,8 @@ stored in the `target/package` directory. This performs the following steps:
     - Fields not recognized by the current Cargo version are dropped.
     - Certain paths, such as `package.readme` and `package.license-file`,
       are rewritten if they point outside the package root.
+    - **Note**: the normalized `Cargo.toml` is intended to be stable,
+      but its exact form may change between Cargo versions.
 
 2. Create the compressed `.crate` file.
     - The normalized `Cargo.toml` is included, and the original manifest is
--- a/src/doc/man/generated_txt/cargo-package.txt
+++ b/src/doc/man/generated_txt/cargo-package.txt
@@ -31,6 +31,9 @@ DESCRIPTION
           o  Certain paths, such as package.readme and package.license-file,
              are rewritten if they point outside the package root.
 
+          o  Note: the normalized Cargo.toml is intended to be stable, but its
+             exact form may change between Cargo versions.
+
        2. Create the compressed .crate file.
 
           o  The normalized Cargo.toml is included, and the original manifest
--- a/src/etc/man/cargo-package.1
+++ b/src/etc/man/cargo-package.1
@@ -43,6 +43,11 @@ stored in the \fBtarget/package\fR directory. This performs the following steps:
 \h'-04'\(bu\h'+03'Certain paths, such as \fBpackage.readme\fR and \fBpackage.license\-file\fR,
 are rewritten if they point outside the package root.
 .RE
+.sp
+.RS 4
+\h'-04'\(bu\h'+03'\fBNote\fR: the normalized \fBCargo.toml\fR is intended to be stable,
+but its exact form may change between Cargo versions.
+.RE
 .RE
 .sp
 .RS 4


src/doc/src/commands/cargo-package.md
```diff
@@ -37,9 +37,6 @@ stored in the `target/package` directory. This performs the following steps:
       this step.
 4. Check that build scripts did not modify any source files.
 
-The list of files included can be controlled with the `include` and `exclude`
-fields in the manifest.
```
epage, line 41: What is this considered duplicate with?

src/doc/man/cargo-package.md
```diff
@@ -18,14 +18,19 @@ This command will create a distributable, compressed `.crate` file with the
...
+1. Normalize the original `Cargo.toml`.
+    - The manifest is reformatted into a packaging-specific TOML layout.
+    - Workspace-inherited fields are resolved to their concrete values.
+    - Path and Git dependency specifications are removed.
+    - Development dependencies without a `version` key specified are removed.
+    - `[patch]`, `[replace]`, and `[workspace]` sections are removed.
+    - Fields not recognized by the current Cargo version are dropped.
+    - Certain paths, such as `package.readme` and `package.license-file`,
+      are rewritten if they point outside the package root.
+
```
epage: For me, the question of whether we should resolve this in docs isn't settled, see conversation at #15836

src/doc/man/cargo-package.md
```diff
@@ -27,6 +27,8 @@ stored in the `target/package` directory. This performs the following steps:
     - Fields not recognized by the current Cargo version are dropped.
     - Certain paths, such as `package.readme` and `package.license-file`,
       are rewritten if they point outside the package root.
+    - **Note**: the normalized `Cargo.toml` is intended to be stable,
+      but its exact form may change between Cargo versions.
```
epage, line 31: imo this sounds contradictory.  If we say this, we need to a bit clearer in what we mean by both

src/doc/src/commands/cargo-package.md
weihanglo: > Files and directories are included or excluded based on rules mentioned in
      [the `[include]` and `[exclude]` fields](../reference/manifest.html#the-exclude-and-include-fields).

src/doc/man/cargo-package.md
weihanglo: I think it is pretty much like the layout of `~/.cargo/registry/src/<url>-<hash>`. We'd strive to make it as stable as possible to avoid churns, but we don't commit to a forever stable format. Do you have an idea how to say this better in manpage like narrative?

epage: So the focus on this is that we aim to minimize changes in the textual output between Cargo versions?  Is that something that belongs here?  I'm trying to think to what purpose it is serving by being here.

src/doc/src/commands/cargo-package.md
epage: So the content within the packaging steps.

This is a case where there is potential benefit to duplication.  The steps mix in a lot of things that don't matter to almost everyone and things that matter a lot.  This buries information, so having it separate like this can help.

Speaking of burying information, I feel like the packaging steps should be broken out into its own section to better highlight lines like this one that is proposed for removing

weihanglo:
Close as this needs more discussions. See #15836

PR #15886, opened 2025-08-26
test(add): Cover some frontmatter corner cases

epage:

When working with the frontmatter parser, I identified some corner cases that we should cover.

I cannot think of a programmatic way to require escaping when going from no frontmatter to a frontmatter or to require more escaping than is already present.


weihanglo (approved):
Thanks!

PR #15889, opened 2025-08-27
feat: Don't stop at first error when emitting lints and warnings

Muscraft:
While emitting warnings, if `cargo` encounters a critical warning or an error while linting, it returns early, potentially before all packages could emit their warnings or be linted. This could make it so users have to run a `cargo` command more than once before seeing all of the relevant output. Beyond this, having an error caused by a lint be a "stop the world" event seems wrong, as it (currently) doesn't inhibit outputting other warnings or linting other packages.

To address this, I made it so that `cargo` waits until the end of `emit_warnings` to return an error if one was encountered. To keep the cause of the error the same as before, `cargo` saves off the first error it encounters, and returns it at the end.

### DIFF
--- a/src/cargo/core/workspace.rs
+++ b/src/cargo/core/workspace.rs
@@ -1205,10 +1205,15 @@ impl<'gctx> Workspace<'gctx> {
     }
 
     pub fn emit_warnings(&self) -> CargoResult<()> {
+        let mut first_emitted_error = None;
         for (path, maybe_pkg) in &self.packages.packages {
             if let MaybePackage::Package(pkg) = maybe_pkg {
                 if self.gctx.cli_unstable().cargo_lints {
-                    self.emit_lints(pkg, &path)?
+                    if let Err(e) = self.emit_lints(pkg, &path)
+                        && first_emitted_error.is_none()
+                    {
+                        first_emitted_error = Some(e);
+                    }
                 }
             }
             let warnings = match maybe_pkg {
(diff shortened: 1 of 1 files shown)


weihanglo:
Would you mind comparing with clippy and rustc whether they have the same behavior? It is not necessary Cargo needs to follow but good to know where we are

src/cargo/core/workspace.rs
```diff
@@ -1220,7 +1225,9 @@ impl<'gctx> Workspace<'gctx> {
                     let err = anyhow::format_err!("{}", warning.message);
                     let cx =
                         anyhow::format_err!("failed to parse manifest at `{}`", path.display());
-                    return Err(err.context(cx));
+                    if first_emitted_error.is_none() {
```
weihanglo, line 1228: Actually the unstable [`build.warnings`](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#buildwarnings) has the same behavior, so fine with that and also this is still unstable.

However, what if there is error critical enough and cargo should stop analyzing? Or at least stop the current package and continue with the next one?

weihanglo, line 1228: What is the reason we we show all warnings but only first error encountered? Should we show all errors as well?

epage: I see this as an incremental improvement. Showing all errors has a bit more policy to figure out (how to render, what error to report back up)

weihanglo: That makes sense!

weihanglo: However it still changed the behavior. Does it make sense to print all warning after critical errors? Will those become false positive or irrelevant?

weihanglo: But anyway we don't really distinguish those too much, so going to merge this.

weihanglo:
Ah. Forgot to merge this. Do it now.

PR #15891, opened 2025-08-28
[WIP] Accessing each build script's `OUT_DIR`

namanlp:
This PR is continuation of my work on [GSoC Project : Build Script Delegation](https://summerofcode.withgoogle.com/programs/2025/projects/nUt4PdAA)

Through this PR, I want the user to be able to access each build script's `OUT_DIR`.

```toml
```

Then, you can access the `OUT_DIR` of given build script by accessing it like a compile time environment variable, for example:

```rust
const BUILD1_OUT_DIR: &'static str = env!("BUILD1_OUT_DIR");
const BUILD2_OUT_DIR: &'static str = env!("BUILD2_OUT_DIR");

fn main() {
    println!("{}", BUILD1_OUT_DIR);
    println!("{}", BUILD2_OUT_DIR);
}
```


src/cargo/core/compiler/build_runner/mod.rs
```diff
@@ -210,7 +210,7 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
             plan.output_plan(self.bcx.gctx);
         }
 
-        // Add `OUT_DIR` to env vars if unit has a build script.
+        // Add `OUT_DIR` to runtime env vars if unit has a build script.
```
epage: Saying `runtime` here might be confusing as we are adding it for use at compile time of the package and not for `cargo run` or `cargo test` running binaries.

src/cargo/core/compiler/build_runner/mod.rs
```diff
@@ -227,11 +227,18 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
...
+
+                    let env = self
+                        .compilation
                         .extra_env
                         .entry(script_meta)
-                        .or_insert_with(Vec::new)
-                        .push(("OUT_DIR".to_string(), out_dir));
+                        .or_insert_with(Vec::new);
+                    env.push(("OUT_DIR".to_string(), out_dir.clone()));
+                    env.push((out_dir_name, out_dir));
```
epage: We'll need to update https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

src/cargo/core/compiler/build_runner/mod.rs
```diff
@@ -227,11 +227,18 @@ impl<'a, 'gctx> BuildRunner<'a, 'gctx> {
...
+                    out_dir_name.push_str("_OUT_DIR");
+
+                    let env = self
+                        .compilation
                         .extra_env
                         .entry(script_meta)
-                        .or_insert_with(Vec::new)
-                        .push(("OUT_DIR".to_string(), out_dir));
+                        .or_insert_with(Vec::new);
+                    env.push(("OUT_DIR".to_string(), out_dir.clone()));
```
epage: Can we make `OUT_DIR` an explicit choice for when we write it, rather than relying on it by incident?  That'll make it less likely for people to break this in the future.

src/cargo/core/compiler/build_runner/mod.rs
epage: Wait, this is us setting the specialized `OUT_DIR` on the build script itself?  We probably shouldn't be doing that.  Depending on how we do build script delegation, the name will be dependent on how the caller sees the build script which will make it undiscoverable for build scripts to use.

src/cargo/core/compiler/mod.rs
```diff
@@ -1706,10 +1706,11 @@ fn build_deps_args(
...
         if dep.unit.mode.is_run_custom_build() {
-            cmd.env(
-                "OUT_DIR",
-                &build_runner.files().build_script_out_dir(&dep.unit),
-            );
+            let out_dir = &build_runner.files().build_script_out_dir(&dep.unit);
+            let mut out_dir_name = dep.unit.target.name().to_owned();
+            out_dir_name.push_str("_OUT_DIR");
+            cmd.env("OUT_DIR", &out_dir);
+            cmd.env(&out_dir_name, &out_dir);
```
epage: don't forget to update the test from 6c8bc087b804b49ed8258de72447e4f7401885fc

src/cargo/core/compiler/build_runner/mod.rs
namanlp: So, should I remove it?

src/cargo/core/compiler/build_runner/mod.rs
namanlp: In a separate commit maybe? Or should I update the latest commit itself?

src/cargo/core/compiler/build_runner/mod.rs
epage: Yes.  While this could make it easier to match up build script to rust source, it runs into the problem mentioned.

src/cargo/core/compiler/build_runner/mod.rs
epage: Unless the commit is ginormous (e.g. adding all of `cargo add` into `cargo`), I would generally update docs as part of commit that added the feature, like tests.

src/cargo/core/compiler/build_runner/mod.rs
namanlp: I'll get back to this

src/cargo/core/compiler/mod.rs
namanlp: Done. I added an extra test for duplicated/conflicting declarations of same function

src/cargo/core/compiler/build_runner/mod.rs
namanlp: Not the earlier code, right?

epage: Not sure what you are asking about.  We should only set the new environment variable on the libs/bins/etc and not on the build script execution.

namanlp: I meant that there was a code doing this earlier. Should I remove it? Or maybe find a new place for it or something like that?

epage: Can you say that again without "this" or "it"?  I'm still not sure what you are referring to.

namanlp: Removed as per our discussion.

tests/testsuite/build_scripts_multiple.rs
```diff
@@ -835,8 +837,105 @@ fn multiple_out_dirs() {
...
         .with_stdout_data(str![[r#"
+Hello, from Build Script 1!
 Hello, from Build Script 2!
 
 "#]])
         .run();
 }
+
+#[cargo_test]
+fn build_script_with_fn_conflicts() {
```
epage: Please add this test in the previous commit

tests/testsuite/build_scripts_multiple.rs
```diff
@@ -835,8 +837,105 @@ fn multiple_out_dirs() {
...
+    // In this, multiple scripts have duplicated/conflicting declarations of same function
```
epage: Why do we need this?

tests/testsuite/build_scripts_multiple.rs
namanlp: Sure.

tests/testsuite/build_scripts_multiple.rs
namanlp: Because duplicated/conflicting functions must throw an error. Ohh, we might have similar test already? Should I remove this one?

epage: Conflicting functions is a compiler-level concern.

Are you thinking of conflicting `<name>_OUT_DIR`s?

namanlp: Ohh, okay. Should I add test for conflicting name of build scripts? Right now, it is throwing error:

```
thread '<unnamed>' panicked at src/cargo/core/compiler/custom_build.rs:1389:39:
build script output collision for multiple-build-scripts-test-repo v0.1.0
```

epage: We should have a test for it and as part of this work, we should be identifying this problem and reporting it as a proper error to the user

We have similar checks
- https://github.com/rust-lang/cargo/blob/7f8ed80cce4b4473f768fcc2147310dc85394c5b/src/cargo/util/toml/targets.rs#L889-L902
- https://github.com/rust-lang/cargo/blob/7f8ed80cce4b4473f768fcc2147310dc85394c5b/src/cargo/util/toml/mod.rs#L2462

tests/testsuite/build_scripts_multiple.rs
namanlp: I removed it.

tests/testsuite/build_scripts_multiple.rs
namanlp: Done adding code, adding tests

src/cargo/core/compiler/build_runner/mod.rs
namanlp: I think only docs are remaining. Everything else should be done.

src/cargo/core/compiler/mod.rs
```diff
@@ -1704,12 +1704,23 @@ fn build_deps_args(
...
-                "OUT_DIR",
-                &build_runner.files().build_script_out_dir(&dep.unit),
-            );
+            let out_dir = &build_runner.files().build_script_out_dir(&dep.unit);
+            let mut out_dir_name = dep.unit.target.name().to_owned();
+            out_dir_name = out_dir_name
+                .strip_prefix("build-script-")
+                .unwrap_or(&out_dir_name)
+                .to_string();
+            out_dir_name.push_str("_OUT_DIR");
```
epage: Can't this be simplified
```suggestion
            let target_name = dep.unit.target.name();
            out_dir_prefix = target_name
                .strip_prefix("build-script-")
                .unwrap_or(target_name);
            let out_dir_name = format!("{out_dir_prefix}_OUT_DIR");
```

tests/testsuite/build_scripts_multiple.rs
```diff
@@ -549,7 +549,7 @@ fn build_script_with_conflicting_out_dirs() {
                 build = ["build1.rs", "build2.rs"]
             "#,
         )
-        // OUT_DIR is set to the lexicographically largest build script's OUT_DIR by default
+        // By default, OUT_DIR is set to that of the last build script in the array
```
epage:
```suggestion
        // By default, OUT_DIR is set to that of the first build script in the array
```
Right?

src/cargo/util/toml/targets.rs
```diff
@@ -901,6 +902,22 @@ fn validate_unique_names(targets: &[TomlTarget], target_kind: &str) -> CargoResu
...
+    let mut seen = HashSet::new();
+    for script in scripts {
+        let stem = Path::new(script).file_stem().unwrap().to_str().unwrap();
+        if !seen.insert(stem) {
+            anyhow::bail!(
+                "found duplicate build script file stem {}, \
+                but all build scripts must have a unique file stem",
+                stem
+            );
+        }
```
epage: Can we gather this into a `HashMap<Name, Vec<Path>>` and then iterate through and if `Vec<Path>` has more than one element, we then error?  That would led us report all errors at once and report the full path for each conflicting build script so users can better understand what they need to change

tests/testsuite/build_scripts_multiple.rs
namanlp: Yes, thanks for fixing it 😅

src/cargo/util/toml/targets.rs
namanlp: Sure, I was just following how earlier `validate_unique_name` functions were written.

src/cargo/core/compiler/mod.rs
namanlp: Sure 👍🏼

src/cargo/core/compiler/build_runner/mod.rs
epage: Is this still the only thing left?

namanlp: It should be done

src/doc/src/reference/environment-variables.md
```diff
@@ -256,6 +256,9 @@ corresponding environment variable is set to the empty string, `""`.
 * `OUT_DIR` --- If the package has a build script, this is set to the folder
   where the build script should place its output. See below for more information.
   (Only set during compilation.)
+* `<script-name>_OUT_DIR` --- If the package has build script(s), this is set to the folder where the
+  build script should place its output. The `<script-name>` is the file-stem of the build script, exactly as-is.
+  For example, `bar_OUT_DIR` for script at `foo/bar.rs`. (Only set during compilation.)
```
epage: This should probably be on `unstable.md`

src/cargo/core/compiler/mod.rs
```diff
@@ -1704,12 +1704,22 @@ fn build_deps_args(
...
-                "OUT_DIR",
-                &build_runner.files().build_script_out_dir(&dep.unit),
-            );
+            let out_dir = &build_runner.files().build_script_out_dir(&dep.unit);
+            let target_name = dep.unit.target.name();
+            let out_dir_prefix = target_name
+                .strip_prefix("build-script-")
+                .unwrap_or(target_name);
+            let out_dir_name = format!("{out_dir_prefix}_OUT_DIR");
+            cmd.env(&out_dir_name, &out_dir);
```
epage: Doesn't is need a feature gate?

src/cargo/util/toml/targets.rs
```diff
@@ -901,6 +902,35 @@ fn validate_unique_names(targets: &[TomlTarget], target_kind: &str) -> CargoResu
...
+        "found build scripts with duplicate file stems, but all build scripts must have a unique file stem",
+    );
+    for (stem, paths) in seen {
+        if paths.len() > 1 {
+            conflict_file_stem = true;
+            err_msg += &format!(
+                "\nBuild Scripts : {} have the same file stem, that is {}",
+                paths.join(", "),
+                stem
+            );
```
epage:
```suggestion
            write!(&mut err_msg, "  for stem `{stem}`: {}", paths.join(", "))?;
```
- generally better to use `write!` to append formatted text
- duplicating the error message
- needs indentation to set apart from the rest of the message

tests/testsuite/build_scripts_multiple.rs
```diff
@@ -764,3 +764,118 @@ fn bar() {
 "#]])
         .run();
 }
+
+#[cargo_test]
+fn multiple_out_dirs() {
```
epage, line 769: Shouldn't this test be added before the validation?

epage: Sorry, I either I was on an outdated view or I was on the wrong view

PR #15896, opened 2025-08-29
fix(cli): Show the bad manifest path

epage:

In most cases, this will just be what you passed on the command line. Even when its not, that likely represents a programmer error.

I mostly did this to help see what was the cause of a test failure.

### DIFF
--- a/src/cargo/util/command_prelude.rs
+++ b/src/cargo/util/command_prelude.rs
@@ -1075,7 +1075,10 @@ pub fn root_manifest(manifest_path: Option<&Path>, gctx: &GlobalContext) -> Carg
         // but in this particular case we need it to fix #3586.
         let path = paths::normalize_path(&path);
         if !path.ends_with("Cargo.toml") && !crate::util::toml::is_embedded(&path) {
-            anyhow::bail!("the manifest-path must be a path to a Cargo.toml file")
+            anyhow::bail!(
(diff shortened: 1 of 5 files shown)


epage:
I can go either way on whether we merge this

src/cargo/util/command_prelude.rs
```diff
@@ -1075,7 +1075,10 @@ pub fn root_manifest(manifest_path: Option<&Path>, gctx: &GlobalContext) -> Carg
         // but in this particular case we need it to fix #3586.
         let path = paths::normalize_path(&path);
         if !path.ends_with("Cargo.toml") && !crate::util::toml::is_embedded(&path) {
-            anyhow::bail!("the manifest-path must be a path to a Cargo.toml file")
+            anyhow::bail!(
```
weihanglo, line 1078: Sometimes I feel like Cargo always presents absolute paths in the very front of the error message, and that makes the important information less outstanding, especially with long path.
Do we want to move it later, something like

```
the manifest-path must be a path to a Cargo.toml file, but got `<path>`
```

I am fine with merging as-is also

weihanglo (approved):

src/cargo/util/command_prelude.rs
epage: Sure.  I had wondered about doing that but hearing you also say it means I probably should

weihanglo (approved):

Issue #15887, opened 2025-08-26
Help new users understand the unintuitive crate name -> Rust namespace mapping for hyphenated names


Timmmm:
### Problem

An annoying papercut with Rust is crate names containing hyphens, e.g. [aho-carosick](https://crates.io/crates/aho-corasick).

It has been a number of years since I learned this, but at some point I knew that you accessed crate members based on the crate name:

```
use crate_name::foo;
```

but I did not know that crate names with hyphens were automatically converted to underscores. I know it sounds obvious if you already know it, but really - this is not obvious, nor is it easy to search for a solution.

Even worse if `lib.name` is set (see also #6827)

I think Cargo (and crates.io, but I guess that's out of scope here) could do more to help:

### Proposed Solution

1. ~~When you create a new crate via `cargo init` with a crate name containing hyphens, warn that it may be confusing to new users because the crate name doesn't match the crate name. This could be a one-time or few-time warning, so as not to annoy people. See also #8883~~
2. When you `cargo add aho-carosick`, note that you will need to use `aho_carosick` in code. Ditto on few-time warnings. Currently it actually warns in the opposite direction which is kind of silly: `warning: translating regex_syntax to regex-syntax`.
3. ~~crates.io should list the command as `cargo add aho_carosick` not `cargo add aho-carosick`. Both work but the former makes it more obvious how to use the code.~~

It's going to be *really* easy for seasoned Rust developers (including me, now) to dismiss this as trivial and obvious. Trust me it wasn't when I was learning, and I'm sure it isn't for many other people.

epage:
For (1) and (3), as mentioned at https://github.com/rust-lang/cargo/issues/2775#issuecomment-3225690028, these make the use of underscores the preferred solution for package names and the Cargo team is not ready at this time to weigh in on making `-` or `_` the preferred case.

Note that this problem isn't just about translating `-` to `_` but users can also override the namespace using `lib.name` (see also #6827).

Quoting from https://github.com/rust-lang/cargo/issues/2775#issuecomment-3224798363

> Another issue brought up in this thread that is indirectly related to the problem at hand which is translating dependency names to Rust namespaces. We now have docs.rs and if packages provide examples, that should help clarify things. **For further improvement, I'd recommend creating an issue against rust-lang/rust for trying to detect when `-` is used in a root namespace and suggest changing it to `_`**.

(emphasis added)

epage:
[...]

Issue #14948, opened 2024-12-17
Reduce the need for users to write build scripts


epage:
Like RUSTFLAGS, build scripts are an important escape hatch.  Like RUSTFLAGS (#12739), we should find replacements for common uses of build scripts so people don't have to reach to this escape hatch so often

Reducing build scripts would
- Improve build times
- Reduce risk of bugs
- Reduce the dependency review audit scope

Uses of build scripts
- [ ] Version detection
  - Stabilizing rust-lang/rust#64797
  - Stabilizing rust-lang/rust#64796
- [ ] Making `cfg` values available at runtime (e.g. in `--bugreport`, `--version`, or crash reports, or for tests building examples)
  - Possibly provide `cfg_value!`, `cfg_values!`, `try_cfg_value!`, `try_cfg_values!` (see also [mutually exclusive features](https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618))
- [ ] Making `TARGET` available at runtime (e.g. `--version`, building examples in tests)
- [ ] Codegen
  - [using tests](https://matklad.github.io/2022/03/26/self-modifying-code.html)
  - #12552
- [ ] [cfg_aliases](https://docs.rs/cfg_aliases/latest/cfg_aliases/)
  - see also https://github.com/rust-lang/cargo/issues/14948#issuecomment-2581426536)
  - see also https://github.com/rust-lang/rfcs/pull/3804
- [ ] Feature warnings (deprecations, a feature being "disabled" due to a platform or another feature, etc)
  - see also https://github.com/rust-lang/rfcs/pull/3486
- [ ] Catch-all: rust-lang/cargo#14903
  - Note: this would also allow consolidating the ecosystem which would help with baking-in more features to replace build scripts
  - Enable linker warnings on Windows
  - Embedded windows manifests (#16328)
  - Can we use this for `-sys` crates / FFI?

ChrisDenton:
I would draw a distinction between main binary build scripts (e.g. the executable or cdylib) and libraries. I think a build scripts for an executable is far more tolerable than for libraries. They're in direct control of the crate author and don't have any other users. Also a single top-level build script is going to have negligible impact on build times for all but the smallest projects.

Which is not to say they're not important too, just that reducing the need for library build scripts will likely have the far bigger impact.

epage:
[...]

PR #15908, opened 2025-09-02
Detect when the initial binary launch may be slow on macOS

madsmtm:
### Motivation

macOS has an antivirus system called XProtect, which does a signature-based check in applications and binaries upon first launch to see if they contain known malicious code. See [Apple's documentation](https://support.apple.com/en-gb/guide/security/sec469d47bd8/web#secffa0f9b55) for details.

This takes a relatively long time though, see the following table for the approximate process run time of the simplest `fn main() {}` on my M2 Macbook (gauged with `time target/debug/foo`):

| - | Cold cache<br>First launch | Cold cache<br>Subsequent launch | Warm cache<br>First launch | Warm cache<br>Subsequent launch |
| --- | --- | --- | --- | --- |
| With XProtect | ~180ms | 8ms | ~120ms | ~4ms |
| Without XProtect | ~9ms | ~7ms | ~3ms | ~2ms |

This is amplified if the binary links other libraries or frameworks, and even more so by the fact that **the `XprotectService` daemon runs in a single thread, so if you try to launch 10 new binaries at once, the slowdown will be more than a second**.

This PR tries to resolve this by querying the OS to try to detect whether XprotectService is likely to spend a lot of time verifying our binaries, and if we think it will, **emit a warning to the user instructing them how they can disable it in their terminal**.

This check and associated warning is gated behind the `-Zdetect-antivirus` flag, and it can be controlled by the configuration option `build.detect-antivirus`.

Implementing a check such as this one really only makes sense in Cargo in contrast to e.g. Rustup, since it may change if the user switches Terminal (thus we cannot do this as a one-time check). We could also do this in the compiler, but it's not really relevant there, the compiler produces binaries, it doesn't care about how they are invoked.

See [this Zulip thread](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/build.20scripts.20slow.20on.20macOS.3F) for further discussion.

### Possible concerns and considerations

- This adds a dependency on [the `objc2` ecosystem](https://github.com/madsmtm/objc2) (`objc2`, `block2` and `objc2-encode`), which I maintain. They are fairly widely used, and Mozilla [plans on vetting them soon-ish](https://github.com/gfx-rs/wgpu/pull/5641), though I acknowledge that this could still be seen as a supply-chain attack. Implementing the execution policy check without these crates is doable, though a lot more code.
- This is fairly complex, though luckily quite self-contained. A simpler solution would be to do a compile+link+run twice and compare the timings, though it will also likely be slower, and it'd be difficult to get the timing right (trying to measure timings suddenly becomes a statistical problem, so there will be a lot more false positives - though maybe possible if we stored the result and kept track over many Cargo invocations?).
- The wording in the user-facing warning needs careful review, we are recommending they disable their antivirus for all programs they run in their Terminal.
- What should the config option be named? I went with `build.detect-antivirus`, an alternative would be to put it under `[term]`. Or maybe `build.detect-slow-bin-first-run`, to focus on the intention instead of what it does?

### Testing

I guess there's two parts to this testing; testing that the code works as written, and testing that the detection logic itself is correct.

To test the code itself with SIP disabled, do:
1. Go to `System Preferences > Security & Privacy > Developer Tools`.
2. Uncheck and remove all items.
3. Run `cargo check -Zdetect-antivirus` on any project.
4. Verify that a warning is produced.
5. Go back to `System Preferences > Security & Privacy > Developer Tools`.
6. Verify that your terminal (usually Terminal.app, might be something else) is added, but disabled.
7. Enable it.
3. Run `cargo check -Zdetect-antivirus` on any project.
9. Verify that no warning is produced.

Testing with SIP enabled:
1. Go to `System Preferences > Security & Privacy > Developer Tools`.
2. Uncheck and remove all items.
3. Run `cargo check -Zdetect-antivirus` on any project.
4. Verify that no warning is produced.
5. Go back to `System Preferences > Security & Privacy > Developer Tools`.
6. Verify that the list is still empty was added.

If you want to change the SIP parts, use `csrutil disable` and/or `csrutil enable --without fs`. This is [possible in a VM with UTM](https://docs.getutm.app/advanced/recovery/), btw, if you don't want to touch your main system.

To check that the output of `cargo check -Zdetect-antivirus` is correct, I've been doing the following:
```zsh
# Create basic binary
rustc -o foo <(echo "fn main() {}")
# Bust XProtect's cache, and see the timings for first and second run
codesign --sign=- --force foo && time ./foo && time ./foo
```

<details><summary>Sample output when XProtect is enabled</summary>
<p>
```console
$ codesign --sign=- --force foo && time ./foo && time ./foo
foo: replacing existing signature
./foo  0,00s user 0,00s system 2% cpu 0,221 total
./foo  0,00s user 0,00s system 73% cpu 0,004 total
```
</p>
</details> 

<details><summary>Sample output when XProtect is disabled</summary>
<p>
```console
$ codesign --sign=- --force foo && time ./foo && time ./foo
foo: replacing existing signature
./foo  0,00s user 0,00s system 66% cpu 0,004 total
./foo  0,00s user 0,00s system 80% cpu 0,003 total
```
</p>
</details> 

Note that it seems like disabling Developer Tool permissions takes a little while to propagate to XProtect, so you might not see the slow timings immediately. A reboot solves this, closing your terminal and waiting a few minutes also does.

See also [this repository I made](https://github.com/madsmtm/check_execution_policy) for testing how this interacts with GitHub Actions (they disable SIP).


src/cargo/util/detect_antivirus/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+            \
+            We recommend this option by default, since it will make your \
+            iteration time lower, though please be aware of the security \
+            implications in doing this. Another option is to disable this \
+            warning by add `build.detect-antivirus = false` to your \
+            ~/.cargo/config.toml.\n\
+            \
+            See <https://support.apple.com/en-gb/guide/security/sec469d47bd8/web> \
+            for more information.",
+        )?;
```
madsmtm: This is an initial stab at the wording, but it definitely needs work!

Note in particular that we are recommending they disable their antivirus for all programs they run in their terminal.

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -556,6 +556,21 @@ where `<compatible-ver>` is the latest version supporting rustc {rustc_version}"
...
     }
 
+    if build_config.detect_antivirus {
+        // TODO(madsmtm): Maybe only do this when we have above a certain
+        // number of build scripts or test binaries to run?
+
+        // TODO(madsmtm): We probably don't want to do this check when doing
+        // `cargo install`?
+
+        // TODO(madsmtm): Consider only warning once every X days.
```
epage: Calling these out to make sure they get tracked

epage:
FYI we tend to prefer [small but atomic commits](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request).  Feel free to edit your commits through the review process.

src/cargo/ops/cargo_compile/mod.rs
madsmtm: There's a few more, I'll add all of it to the tracking issue once we create one (which I'm not sure I should yet?)

Cargo.toml
```diff
@@ -69,6 +70,7 @@ libgit2-sys = "0.18.2"
 libloading = "0.8.8"
 memchr = "2.7.5"
 miow = "0.6.0"
+objc2 = "0.6.2"
```
epage, line 73: > This adds a dependency on [the objc2 ecosystem](https://github.com/madsmtm/objc2) (objc2, block2 and objc2-encode), which I maintain. They are fairly widely used, and Mozilla https://github.com/gfx-rs/wgpu/pull/5641, though I acknowledge that this could still be seen as a supply-chain attack. Implementing the execution policy check without these crates is doable, though a lot more code.

src/doc/src/reference/unstable.md
```diff
@@ -1947,6 +1948,25 @@ enabled = true
...
+
+Currently only implemented for macOS' XProtect/Gatekeeper, but could be expanded to Windows Defender in the future.
+
+```toml
+# Example ~/.cargo/config.toml
+
+# Disable warning, e.g. if using a workplace-issued Mac that
+# doesn't allow granting Developer Tool permissions.
+[build]
+detect-antivirus = false
```
epage, line 1969: > What should the config option be named? I went with build.detect-antivirus, an alternative would be to put it under [term]. Or maybe build.detect-slow-bin-first-run, to focus on the intention instead of what it does?

src/doc/src/reference/unstable.md
```diff
@@ -1947,6 +1948,25 @@ enabled = true
...
+```
```
epage, line 1970: Can you document this field like we do in our [config reference](https://doc.rust-lang.org/cargo/reference/config.html)

src/cargo/util/detect_antivirus/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        // TODO(madsmtm): We could possibly also try to compile a simple
+        // binary here, run it twice, and compare the time it took?
+        // Just as a way to verify if our checks above were actually correct.
+
+        // TODO(madsmtm): Make a page in Cargo's docs that explains this in
+        // more depth.
+
+        // TODO(madsmtm): Iterate on the wording here.
```
epage: Calling these out so we track them

madsmtm:
> FYI we tend to prefer [small but atomic commits](https://doc.crates.io/contrib/process/working-on-cargo.html#submitting-a-pull-request). Feel free to edit your commits through the review process.

I can try to split it up further, maybe as:
1. Execution policy check.
2. SIP check.
3. Add `-Z` flag.
4. Add config option.

Or would you prefer a different order / more splitting?

src/cargo/util/detect_antivirus/mod.rs
epage: Should this be a `note` instead of a `warn`?  I feel like that would reduce some of the pressure on the user to feel they need to disable security measures

epage: Should we be making a recommendation or just offering two different suggested ways of resolving this?

epage: > "XProtect (macOS' antivirus software) is likely going to make it \
> slower for you to run build scripts and test binaries, since it \
> scans every new binary for signatures of known malware (and this \
> can take up to severals seconds per binary).\n\

Does this also affect proc-macros?

epage: > "XProtect (macOS' antivirus software) is likely going to make it \
> slower for you to run build scripts and test binaries, since it \
> scans every new binary for signatures of known malware (and this \
> can take up to severals seconds per binary).\n\

This just describes XProtect but doesn't say that it is enabled.

Maybe starting with:

> detected XProtect is enabled on this terminal which may slow down builds as it scans build scripts and test binaries before they are run.
>
> suggestion: if you trust your packages and their dependencies, you can disable XProtect by ...
>
> suggestion: otherwise, you can silence this message by ...

src/doc/src/reference/unstable.md
epage: Another option is we move forward with an `[advice]` table, mirroring git's.  It would be ideal if we could suggest a command to run to turn off the advice (rather than saying how to edit a file)

I know `[advice]` has at least come up at https://github.com/rust-lang/cargo/issues/15887#issuecomment-3228244230, unsure where else.

epage:
> I can try to split it up further, maybe as:

To clarify, I feel the current commits aren't too atomic
- The first has dead code (`if false`)
- The second says its about adding a feature flag but it also adds the config

You could either
- Merge the commits
- switch the `if false` to `if true` and split the config and `-Z` commits

src/cargo/ops/cargo_compile/mod.rs
epage: For myself, I would prefer all TODOs to be resolved before we merge, whether by making a decision and change within this PR, moving them to issues, or moving them to unresolved questions on a tracking issue.

TODOs act as a second backlog and their resolution isn't always resolving them but sometimes they instead as more of `HACK` comments.  I prefer just to not have TODOs present and either track everything in one backlog (our issues) or recognize them as `HACK` comments.

src/cargo/util/detect_antivirus/mod.rs
madsmtm: > Should this be a `note` instead of a `warn`? I feel like that would reduce some of the pressure on the user to feel they need to disable security measures

Ah, didn't know Cargo had `note`, let's definitely go with that!

> Should we be making a recommendation or just offering two different suggested ways of resolving this?

True, we don't have to make a recommendation here. I'll leave that up to you to decide - we're very focused on speed here, but there is a little value in protecting your system from malicious build scripts, and perhaps a lot of value in not leaving your terminal open like this in general.

> Does this also affect proc-macros?

Good question! I tried testing it, and I think the answer is no, `dlopen`-ing the proc macro like `rustc` does is apparently not affected the same way.

>  Maybe starting with:

Good suggestion, I'll update the text.

src/doc/src/reference/unstable.md
madsmtm: I've tried to do so now, though there isn't really precedent for this in `unstable.md`

madsmtm:
I've tried to reorder and split things in a way that maybe makes a bit more sense:
- Add -Z flag.
- Add config option.
- Add the actual logic for the check.

src/cargo/util/detect_antivirus/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+            under `System Preferences > Security & Privacy > Developer Tools`. \
+            (Cargo has made an entry for your current terminal appear there \
+            already, though you will need to go and manually enable it).\
+            \n\
+            Alternatively, you can disable this note by adding \
+            `build.detect-antivirus = false` to your ~/.cargo/config.toml.\
+            \n\
+            See <https://support.apple.com/en-gb/guide/security/sec469d47bd8/web> \
+            for more information.",
+        )?;
```
madsmtm: This needs further work on the wording!

Previously discussed in https://github.com/rust-lang/cargo/pull/15908#discussion_r2316584967.

src/cargo/util/detect_antivirus/mod.rs
madsmtm: I'll resolve this one, and move it to https://github.com/rust-lang/cargo/pull/15908#discussion_r2316733490, since the wording is now something else.

src/cargo/util/detect_antivirus/mod.rs
madsmtm: I went with "session" over "terminal", since this term also applies if you're connecting via ssh (or running Cargo directly as a launchd agent). WDYT?

(Technically, I think it's something like "process context" or "TCC entitlement inheritance state", but the other terms are probably clearer).

src/doc/src/reference/unstable.md
epage: Its a more recent thing I've encouraged and apparently everything that has done it has been stabilized, e.g. https://doc.rust-lang.org/cargo/reference/unstable.html#build-dir

madsmtm: Ah, makes sense

epage:
Thanks for your work on this!  At a high level, its looking good!

Note that I've not dug into the unsafe code and the cargo team needs to discuss this further.

src/cargo/util/detect_antivirus/execution_policy.rs
```diff
@@ -0,0 +1,185 @@
...
+impl Drop for ExecutionPolicyHandle {
+    fn drop(&mut self) {
+        // SAFETY: The handle is valid.
+        let _ = unsafe { libc::dlclose(self.0) };
+        // Ignore errors when closing. This is also what `libloading` does:
+        // https://docs.rs/libloading/0.8.6/src/libloading/os/unix/mod.rs.html#374
+    }
+}
+
+/// A  that interacts with the system via XPC.
```
nnethercote: missing word?

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -556,6 +561,44 @@ where `<compatible-ver>` is the latest version supporting rustc {rustc_version}"
...
+        // We also don't want to do this check when installing, since there
+        // might be `cargo install` users who are not necessarily developers
+        // (and so the note will be irrelevant to them).
+        if (10 < num_binaries && build_config.intent != UserIntent::Install)
+            || build_config.detect_antivirus == DetectAntivirus::Always
+        {
+            if let Err(err) = detect_antivirus::detect_and_report(gctx) {
+                // Errors in this detection are not fatal.
+                tracing::error!("failed detecting whether binaries may be slow to run: {err}");
+            }
```
madsmtm, line 594: I am a bit unsure whether the end of `create_bcx` is the right place for this check?

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -556,6 +561,44 @@ where `<compatible-ver>` is the latest version supporting rustc {rustc_version}"
...
+        // don't hit this.
+        //
+        // TODO(madsmtm): What should the threshold be?
+        //
+        // We also don't want to do this check when installing, since there
+        // might be `cargo install` users who are not necessarily developers
+        // (and so the note will be irrelevant to them).
+        if (10 < num_binaries && build_config.intent != UserIntent::Install)
+            || build_config.detect_antivirus == DetectAntivirus::Always
+        {
```
madsmtm: If we want this heuristic, what should the threshold be?

As a data point, a simple binary using `reqwest` (in blocking mode) currently requires 10 build scripts.

src/cargo/ops/cargo_compile/mod.rs
madsmtm: I've implemented the heuristics now, just so you could get a feel for what it'd look like. I'd be fine with not including those parts.

src/cargo/ops/cargo_compile/mod.rs
epage: Any particular reason for the concern?

Note that we have another type of check in here, the rustc compatibility one which you put this right after so that works from a consolidating environment checks perspective.

src/cargo/ops/cargo_compile/mod.rs
epage: One challenge with this heuristic is that tests, doctests, and build scripts are very different

For tests (in `cargo test`s model), the overhead of the test itself is likely to dwarf XProtect.  We don't have parallel test running yet (working on the foundation to allow it) which defers that problem.

For doctests, 2024 edition at least reduces the number of binaries.  However, there is no caching of the binaries that are built (yet) so there are no binaries on every run which is a problem if there is any memory to XProtect.

Both of the above may not be relevant if the user is doing a `cargo check` or even a `cargo build` which is what this location is for.

Regarding build scripts, I wouldn't expect those in a beginner project except through dependencies which can really blow up the numbers with deps like `reqwest` atm.   After #14948, beginnings should hopefully only deal with build scripts for `-sys` packages and I have a likely unrealistic dream that we can consolidate those down to 1.  Larger projects may also have build scripts for gathering commit information and adding a windows manifest.

For build scripts, the single threaded nature of XProtect becomes a problem for build scripts that are likely to run in parallel.

epage: My counter proposals
- no binaries check
- only check build scripts for now and note in #5609 that we should add a check like this to `cargo test`, and either
  - check if any build scripts (`0<num_binaries`)
  - check a small number (e.g. `4<num_binaries`)

src/cargo/ops/cargo_compile/mod.rs
madsmtm: I _think_ it's the right place, but I just wanted to call to attention that I'm not familiar with Cargo's internals, and that there might be a more suited place? But if you say it's correct, then it's correct.

src/cargo/ops/cargo_compile/mod.rs
madsmtm: > One challenge with this heuristic is that tests, doctests, and build scripts are very different

Just throwing an idea out there, we could make it a weighted sum instead?
```rust
let binary_weight = unit_graph.keys().filter(|unit| match unit.mode {
    UnitMode::RunCustomBuild => 1.0,
    UnitMode::Test => 0.5, // Or something
    UnitMode::Doctest => 0.7, // Or something
    _ => 0.0,
}).sum();
```

> For tests (in `cargo test`s model), the overhead of the test itself is likely to dwarf XProtect. We don't have parallel test running yet (working on the foundation to allow it) which defers that problem.

Not sure I agree with this assessment? If you have a lot of integration tests, it's going to take a long time to run all those if XProtect wants to scan them all first. Sure, there's also build and link time of those, but that can be done in parallel.

A quick test with 100 integration tests revealed ~2 seconds build+link time and ~15 seconds test run time on my machine. With XProtect disabled, the test run time was down to ~0.4 seconds (even though they were run serially by Cargo).

> For doctests, 2024 edition at least reduces the number of binaries. However, there is no caching of the binaries that are built (yet) so there are no binaries on every run which is a problem if there is any memory to XProtect.

I'm not sure I understand what you mean? If doctest's binaries aren't cached, then there's even more reason why XProtect is going to be problematic there?

> Both of the above may not be relevant if the user is doing a `cargo check` or even a `cargo build` which is what this location is for.

True. Not that a plain `cargo check`/`cargo build` will produce `UnitMode::Test`, but `cargo test --no-run` and `cargo build --tests` both do, and in those cases the heuristic is definitely wrong.

Maybe there's a way to know how many binaries we're _actually_ going to run? Or maybe we should do the check later?

> Regarding build scripts, I wouldn't expect those in a beginner project except through dependencies which can really blow up the numbers with deps like `reqwest` atm.

Agreed, and yes, we should also work on reducing that.

epage: > A quick test with 100 integration tests revealed ~2 seconds build+link time and ~15 seconds test run time on my machine. With XProtect disabled, the test run time was down to ~0.4 seconds (even though they were run serially by Cargo).

Uh, wow.  I had assumed that the biggest problem was that it runs serially so the overrhead for an already serial operation would be minimal.  This is strange.

> I'm not sure I understand what you mean? If doctest's binaries aren't cached, then there's even more reason why XProtect is going to be problematic there?

Yes, thats what I'm calling out.  On the surface level, Edition 2024 reduced the impact of this but the fact that they are rebuilt every run is a problem.

> True. Not that a plain cargo check/cargo build will produce UnitMode::Test, but cargo test --no-run and cargo build --tests both do, and in those cases the heuristic is definitely wrong.

That builds them but still doesn't run them and so therefore isn't running into the issue.  I feel weird providing the message speculatively like that which is why I was suggesting the test check be in `cargo test`.  Granted, I thought that wouldn't be a problem until #5609 but maybe it would be needed now.

src/cargo/core/features.rs
```diff
@@ -400,6 +400,36 @@ impl FromStr for FixEdition {
     }
 }
 
+/// The value for `-Zdetect-antivirus`.
+#[derive(Debug, Deserialize, Default, PartialEq, Eq, Hash, Copy, Clone)]
+#[serde(rename = "lowercase")]
```
ehuss: This isn't working correctly.

```suggestion
#[serde(rename_all = "lowercase")]
```

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -556,6 +561,44 @@ where `<compatible-ver>` is the latest version supporting rustc {rustc_version}"
...
+            if let Err(err) = detect_antivirus::detect_and_report(gctx) {
```
ehuss, line 591: One thing I wanted to note is that there is a performance cost to this. At least on my system, an error result took over 60ms, and a passing result took 14ms. We try to keep cargo's overhead relatively small (I like to keep the budget under 500ms total). 14ms is probably ok, but it's inching towards the territory where it is significant.

src/cargo/util/detect_antivirus/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+        gtcx.shell().note(
+            "detected that XProtect is enabled in this session, which may \
+            slow down builds as it scans build scripts and test binaries \
+            before they are run.\
+            \n\
+            If you trust your packages and their dependencies, then this \
+            overhead can be avoided by giving your session more permissions \
+            under `System Preferences > Security & Privacy > Developer Tools`. \
+            (Cargo has made an entry for your current session appear there \
+            already, though you will need to go and manually enable it).\
```
ehuss: Messaging-wise, this seems much more widespread than just trusting cargo packages and dependencies, since it is authorizing the terminal/editor/tool that cargo is running under. That means, for example, you have to trust everything that happens in the terminal or editor.

One question that came up is whether or not it is possible to trust just cargo/rustup, or does it have to be the App that it is running under? It wasn't really clear to us what the rules were here.

src/cargo/core/features.rs
```diff
@@ -400,6 +400,36 @@ impl FromStr for FixEdition {
...
+}
+
+impl FromStr for DetectAntivirus {
+    type Err = anyhow::Error;
+    fn from_str(s: &str) -> Result<Self, <Self as FromStr>::Err> {
+        Ok(match s {
+            "always" => Self::Always,
+            "auto" => Self::Auto,
+            "never" => Self::Never,
+            _ => bail!("invalid `-Zfix-edition, expected `always`, `auto` or `never`, got `{s}`"),
```
ehuss: The `-Zfix-edition` is wrong here. Probably best to just remove that part of the message.

src/cargo/util/detect_antivirus/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+            \n\
+            Alternatively, you can disable this note by adding \
+            `build.detect-antivirus = false` to your ~/.cargo/config.toml.\
```
ehuss, line 111: I see there may be some discussion about the exact configuration value, but I wanted to note that this currently isn't correct since `false` isn't a valid value. Just want to make sure that all gets updated.

madsmtm: I intended for:
- `-Zdetect-antivirus=auto`, `-Zdetect-antivirus=always` and `-Zdetect-antivirus=never`, because I thought that would be useful while testing.
- `[build] detect-antivirus = false` or `[build] detect-antivirus = true` in configuration file, I don't actually want users to be able to tweak the heuristic of when this fires.

What do you mean by `false` not being a valid value?

src/cargo/ops/cargo_compile/mod.rs
madsmtm: Hmm, you're right.

On my machine (to have comparable numbers to those I've given elsewhere), I get:
| Action | Time |
| --- | --- |
| `dlopen`-ing `ExecutionPolicy.framework` | 0.2-0.5ms |
| Create `EPDeveloperTool` | ~0.3ms |
| Get authorization status (fail) | 5-7ms |
| Get authorization status (success) | 4-6ms |
| Check SIP status (fail) | 2-4ms |
| Request access (fail) | 8-22ms |

This is of course unfortunate, and is more reason to not do the check unless the heuristic triggers (whatever the heuristic ends up being).

src/cargo/util/detect_antivirus/mod.rs
madsmtm: > One question that came up is whether or not it is possible to trust just cargo/rustup, or does it have to be the App that it is running under? It wasn't really clear to us what the rules were here.

If there is, I haven't found a way to do so. Adding just the `cargo` / `rustup` binary as a Developer Tool does not work.

I recall someone suggesting something about making a launchd daemon (or agent). I think that would allow `cargo` to XPC to `cargod` and request it to build and run build scripts, and only have the daemon have the elevated privileges - but there's wayyy too many issues with that idea.

src/cargo/util/detect_antivirus/mod.rs
madsmtm: Actually, never mind, "terminal" is probably clearer for the user, so instead I've used "session" at the beginning of the diagnostic, and "terminal" later.

Current wording:
```
note: detected that XProtect is enabled in this session, which may slow down builds as it scans build scripts and test binaries before they are run.
If you trust the software that you run in your terminal, then this overhead can be avoided by giving it more permissions under `System Preferences > Security & Privacy > Developer Tools`. (Cargo has made an entry for the current terminal appear there, though you will need to go and manually enable it).
Alternatively, you can disable this note by adding `build.detect-antivirus = false` to your ~/.cargo/config.toml.
See <https://doc.rust-lang.org/cargo/appendix/antivirus.html#xprotect> for more information.
```

src/cargo/util/detect_antivirus/mod.rs
madsmtm: > Messaging-wise, this seems much more widespread than just trusting cargo packages and dependencies, since it is authorizing the terminal/editor/tool that cargo is running under. That means, for example, you have to trust everything that happens in the terminal or editor.

I have tried to make the text a bit clearer in this regard, the current wording is:
```
note: detected that XProtect is enabled in this session, which may slow down builds as it scans build scripts and test binaries before they are run.
If you trust the software that you run in your terminal, then this overhead can be avoided by giving it more permissions under `System Preferences > Security & Privacy > Developer Tools`. (Cargo has made an entry for the current terminal appear there, though you will need to go and manually enable it).
Alternatively, you can disable this note by adding `build.detect-antivirus = false` to your ~/.cargo/config.toml.
See <https://doc.rust-lang.org/cargo/appendix/antivirus.html#xprotect> for more information.
```

See also `appendix/antivirus.md` that I just added to the docs.

src/doc/src/appendix/antivirus.md
```diff
@@ -0,0 +1,32 @@
...
+
+<!-- TODO: Talk about Cargo's `build.detect-antivirus = false` once stable -->
+
+### Security considerations
+
+Unfortunately, there doesn't seem to be a way to scope this to select binaries, e.g. adding Cargo directly as a developer tool has no effect, it has to be the "top-level" process.
+
+As such, **this disables your antivirus for all software that is launched via your terminal**. This is only one part of macOS' security protections that specifically checks for known malware signatures, and it is nowhere near as unsafe as disabling SIP or giving Full Disk Access would be - but depending on your security constraints, you might still consider leaving it enabled.
+
+Changing this setting has not been tested on systems in enterprise environments, it might affect more things there.
```
madsmtm: I added this appendix to the documentation, so that the note can link to it. Needs iteration on the wording.

src/cargo/ops/cargo_compile/mod.rs
madsmtm: I have removed the TODOs now.

src/cargo/util/detect_antivirus/mod.rs
madsmtm: I have added a page to Cargo's docs.

I decided to not add the verification step, it'd be quite a hazzle to get right, and I'd rather we just tell users some commands they can run when we do the call-for-testing of this feature.

src/cargo/ops/cargo_compile/mod.rs
```diff
@@ -556,6 +561,40 @@ where `<compatible-ver>` is the latest version supporting rustc {rustc_version}"
...
+
+        // Heuristic: Only do the check if we have to run more than a specific
+        // number of binaries. This makes it so that small beginner projects
+        // don't hit this.
+        //
+        // We also don't want to do this check when installing, since there
+        // might be `cargo install` users who are not necessarily developers
+        // (and so the note will be irrelevant to them).
+        if (10 < num_binaries && build_config.intent != UserIntent::Install)
+            || build_config.detect_antivirus == DetectAntivirus::Always
```
madsmtm, line 589: As an alternative, or additional thing, we could consider only doing the check once every X days, a bit similar to the automatic garbage collection.

This has the problem that it might take a long time for the user to notice the message, and worse, it makes Cargo's output non-deterministic, so I'm inclined not to do it.

3a6db5:
I'm not a Cargo contributor, but I wanted to share my thoughts on this.

I don't think `cargo` should suggest disabling the antivirus to speed up builds.
I'm fully in favor of documenting the behavior so that users who care about build‑time overhead can find the information, but I think it shouldn't be shown to most users through `cargo` invocations.

My arguments are:

1. It isn't clear that the build‑speed gain outweighs the loss of security.
2. Many readers (students learning their first language from the Rust book, or developers without a cybersecurity background) might be tempted to follow `cargo`'s advice without fully grasping the risks. Even for experienced individuals, I think it is very difficult to "trust the software you run in your terminal". Recent supply‑chain attacks on [npm packages](https://socket.dev/blog/tinycolor-supply-chain-attack-affects-40-packages) and [Cargo crates](https://blog.rust-lang.org/2025/09/24/crates.io-malicious-crates-fasterlog-and-asyncprintln/) show how real the threat is to everyone.
3. If someone disables their antivirus based on `cargo`'s recommendation and later gets hacked, Cargo could be perceived as bearing some responsibility.

A legitimate scenario where disabling the check _might_ make sense is in certain CI/CD environments.

Perhaps we could solicit input from Apple.

src/doc/src/reference/unstable.md
3a6db5: An entry in the `[advice]` table could also disable the note introduced in #13371.

src/cargo/util/detect_antivirus/mod.rs
jabedude: > One question that came up is whether or not it is possible to trust just cargo/rustup, or does it have to be the App that it is running under? It wasn't really clear to us what the rules were here.

Like TCC permissions, the developer tool exception applies to the "responsible process" which would usually be the terminal emulator application (iTerm/Terminal.app/etc). Targeting `cargo` only would require using this macOS API to set a given child process as responsible for itself (and all it's children):
```
int responsibility_spawnattrs_setdisclaim(posix_spawnattr_t attrs, int disclaim);
```

See https://www.qt.io/blog/the-curious-case-of-the-responsible-process for more info

madsmtm:
Fair points. I do agree that XProtect can be useful! To test it out a bit, I opened a fresh VM, downloaded a few of the binaries on [this page](https://objective-see.org/malware.html), removed the `com.apple.quarantine` attribute on them, disabled network access, and ran them via the Terminal. One of them gave this pop-up:

screenshot-malware-blocked

Retrying the same binary with "Developer Tools" permissions, the malware just ran.

---

So clearly, this _can_ help! That said, I struggle to see many real-world developer environment scenarios where it would? Remember, XProtect only scans binary launches (it doesn't get a chance to inspect malware in interpreted programs), and it is documented to use [YARA signatures](https://virustotal.github.io/yara/), which means that it is inherently flaky and won't necessarily catch a problem if the binary is recompiled.

I tried to make a table of different kinds of attacks that most people are vulnerable to in their terminal, and whether I think XProtect would help you (take it with a grain of salt, I'm not an expert in YARA signatures):

| Attack vector | XProtect helps? | Comment |
| --- | --- | --- |
| npm, Python etc. ecosystem | No | Node.js, Python etc. are interpreted |
| Recent crates.io attack | Maybe | Depends on whether XProtect would be able to detect the malicious code after it'd been inlined and optimized by `rustc`, I find it somewhat unlikely |
| crates.io build.rs | Maybe | Same as above |
| crates.io proc-macro | No | `dlopen`ed by the compiler, bypasses XProtect |
| Homebrew formula | No | Ruby is interpreted |
| Homebrew bottle (prebuilt formula) | Yes | Assuming the signature of the attack is added to Apple's signature lists, and that the formula doesn't bypass being prebuilt |
| VSCode extension | Probably not | I think these are launched via JavaScript, and hence interpreted? Also unsure whether they're launched in the Terminal.app context or the VSCode.app context? |
| A random `.app` from the internet | Yes | This is what XProtect is built for, and would still work as long as users don't open the app in their Terminal by launching the binary directly as `./Malware.app/Contents/MacOS/Malware` |

In summary, my hypothesis is that having XProtect enabled in your Terminal will only very marginally increase the security of your development machine (to the degree that it's almost worse if you think you can rely on it. You can't. You need proper sandboxing).

Or do you have evidence to / an idea of the contrary? Am I wrong, would the npm and crates.io attacks you mention have been prevented by XProtect?

> Perhaps we could solicit input from Apple.

I'd be fine with opening a Feedback Request with them, what would you suggest I ask about? Whether they think it's a good idea to disable their antivirus? Ofc they won't say yes to that. But maybe we should request that they optimize XProtect?

madsmtm:
I realized I may have not really responded to your arguments, lemme be more explicit:

> It isn't clear that the build‑speed gain outweighs the loss of security.

Well, that's probably subjective, and this something we could argue about for an eternity and probably never reach a conclusion. I have presented my view above that I believe the security loss is small, and I have found the speedup to be huge when working on larger projects like the compiler itself (test suite runs 3x faster!).

> Many readers (students learning their first language from the Rust book, or developers without a cybersecurity background) might be tempted to follow `cargo`'s advice without fully grasping the risks. Even for experienced individuals, I think it is very difficult to "trust the software you run in your terminal". Recent supply‑chain attacks on [npm packages](https://socket.dev/blog/tinycolor-supply-chain-attack-affects-40-packages) and [Cargo crates](https://blog.rust-lang.org/2025/09/24/crates.io-malicious-crates-fasterlog-and-asyncprintln/) show how real the threat is to everyone.

I agree that this is a problem, I just disagree that antivirus software is the solution, you have to trust the software in your terminal regardless of whether you have your antivirus enabled or not.

> If someone disables their antivirus based on `cargo`'s recommendation and later gets hacked, Cargo could be perceived as bearing some responsibility.

True. I doubt Cargo could get into legal trouble for doing this, but IANAL so we'd have to ask for legal help from the Rust Foundation. But there's definitely a social aspect, it could cause a fairly big backlash if there was some widespread attack that affects a lot of Rust users because they've disabled this. I perceive this risk to be small, but I could very well be wrong here.

src/cargo/util/detect_antivirus/mod.rs
madsmtm: Thanks for the link!

Notably, if we wanted the user to give Cargo those permissions, it would not be something that Cargo could easily do. It has to be the terminal itself that does that.

Unless maybe if Cargo were to spawn a child copy of itself? Or maybe we could get the Rustup wrappers to do it? That said, it'd probably only shift and not eliminate the attack vector, if we were to do this then Cargo could still be spawned by a malicious actor to obtain the heightened privileges.

3a6db5:
Thank you for reviewing my points, I really appreciate the nuance you added.

> In summary, my hypothesis is that having XProtect enabled in your Terminal will only very marginally increase the security of your development machine (to the degree that it's almost worse if you think you can rely on it. You can't. You need proper sandboxing).

While I agree that XProtect isn't perfect, it remains one layer of defense and disabling it would weaken overall protection. Though, I think the effectiveness of XProtect shouldn't determine whether we add the warning or not.

> Am I wrong, would the npm and crates.io attacks you mention have been prevented by XProtect?

My reference to the recent supply‑chain attacks wasn't meant to suggest that XProtect would have caught them, but to illustrate that crates can be infection vectors. I imagine a build script could download and run malware that XProtect already knows about.

> I'd be fine with opening a Feedback Request with them, what would you suggest I ask about? Whether they think it's a good idea to disable their antivirus? Ofc they won't say yes to that. But maybe we should request that they optimize XProtect?

We could ask Apple whether there’s an alternative workaround (such as suggested [in jabedude's comment](https://github.com/rust-lang/cargo/pull/15908#discussion_r2386087720)) and/or whether our recommendation to users is appropriate (though, as you pointed out, they will probably disagree).

> True. I doubt Cargo could get into legal trouble for doing this, but IANAL so we'd have to ask for legal help from the Rust Foundation. But there's definitely a social aspect, it could cause a fairly big backlash if there was some widespread attack that affects a lot of Rust users because they've disabled this. I perceive this risk to be small, but I could very well be wrong here.

My focus was more on the public perception rather than any legal implications.

PR #15910, opened 2025-09-02
fix(publish): Don't generate final artifacts

epage:

When splitting `build-dir` out of `target-dir` (see #14125), we kept `.crate` files as final artifacts of `cargo package`.  However, that also made them final artifacts of `cargo publish` when the side effect of that should be the publish operation.

This changes things so that we instead package within `build-dir` and `cargo package` then uplifts these to `target-dir`.

Behavior change: when running `cargo publish` without `build-dir` set, `target/package/*.crate` will be moved to `target/package/tmp-crate/*.crate`.  This should be fine as its an intermediate artifact but there is a risk that this can break someone who did not realize that.  Part of the reason for making the `build-dir` / `target-dir` split is to make our intermediate/final artifact intent clearer.  We did intend to be a transition period before changing things, like with #15010.  I'm assuming its fine that we don't have a transition period here.

### Notes

This was discussed at [#t-cargo > &#96;-Zbuild-dir&#96; and &#96;cargo publish&#96; / &#96;cargo package&#96; @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/.60-Zbuild-dir.60.20and.20.60cargo.20publish.60.20.2F.20.60cargo.20package.60/near/537336819)

### DIFF
--- a/src/cargo/ops/cargo_package/mod.rs
+++ b/src/cargo/ops/cargo_package/mod.rs
@@ -217,10 +217,12 @@ pub fn package(ws: &Workspace<'_>, opts: &PackageOpts<'_>) -> CargoResult<Vec<Fi
     // So we need filter
     pkgs.retain(|(pkg, _feats)| specs.iter().any(|spec| spec.matches(pkg.package_id())));
 
-    Ok(do_package(ws, opts, pkgs)?
-        .into_iter()
-        .map(|x| x.2)
-        .collect())
+    let packaged = do_package(ws, opts, pkgs)?;
+
+    let mut result = Vec::new();
+    result.extend(packaged.into_iter().map(|(_, _, src)| src));
+
+    Ok(result)
 }
 
 /// Packages an entire workspace.
--- a/src/cargo/ops/cargo_package/mod.rs
+++ b/src/cargo/ops/cargo_package/mod.rs
@@ -160,7 +160,7 @@ fn create_package(
     }
 
     let filename = pkg.package_id().tarball_name();
-    let dir = ws.target_dir().join("package");
+    let dir = ws.build_dir().join("package");
     let mut dst = {
         let tmp = format!(".{}", filename);
         dir.open_rw_exclusive_create(&tmp, gctx, "package scratch space")?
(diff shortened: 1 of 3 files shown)


src/cargo/ops/cargo_package/mod.rs
```diff
@@ -160,11 +160,8 @@ fn create_package(
     }
 
     let filename = pkg.package_id().tarball_name();
-    let dir = ws.target_dir().join("package");
-    let mut dst = {
-        let tmp = format!(".{}", filename);
-        dir.open_rw_exclusive_create(&tmp, gctx, "package scratch space")?
-    };
+    let dir = ws.build_dir().join("package").join("tmp-crate");
```
weihanglo: I am a bit unsure about adding this `tmp-crate`. From [the doc](https://doc.rust-lang.org/1.88.0/cargo/reference/build-cache.html)


> Contains the output of the [cargo package](https://doc.rust-lang.org/1.88.0/cargo/commands/cargo-package.html) and [cargo publish](https://doc.rust-lang.org/1.88.0/cargo/commands/cargo-publish.html) commands.

which means something is there, and it was not clear whether they were intermediate or final artifacts. To minimize the impact, is there a way to do this without the new `tmp-crate` directory?

weihanglo (commented):
We'll need to update the doc as well <https://doc.rust-lang.org/1.89.0/cargo/reference/build-cache.html>.

src/cargo/ops/cargo_package/mod.rs
epage: We already have `tmp-registry` and the decompressed directories in there.

Is the concern about mixing intermediate and final artifacts in similar locations?  We can split that but since its larger than just this, should we split this out?

weihanglo: The concern is people already relying on the package directory layout with `cargo publish`. While that is not intended, the existing doc putting package and publish together makes them looks like the same. Should we give people one more release to adjust?
We could include a note in the changelog saying this will change in the next release even you are not using `build.build-dir`. What do you think?

epage: I'm fine if we want more of a transition.  I've dropped the last commit from this PR which means that by default `cargo publish` will still output `.crate` files in the same spot and it will only change when `build-dir` is overridden.

weihanglo: Thanks!

weihanglo: Should probably track it somewhere so when a new nightly branches off we can add the commit back

epage: I'll post a draft PR once this is merged.

If we want a transition period, we likely will want more than one release.

PR #15915, opened 2025-09-03
fix(publish): Move `.crate` out of final artifact location

epage:
When `target_dir == build_dir`, ensure `cargo publish` doesn't put intermediate artifacts in the final artifact location of `cargo package`.

In #15910, users could identify that `.crate` files from `cargo publish` are not final artifacts by setting a custom `build-dir`.  This extends that to all users, ie when `build-dir = target-dir` (the default currently), making it clear that these files are internal.

This also cleans things up by consolidating all of the uplifting logic and avoids dealing with overlapping `target_dir` and `build_dir`.

### Notes

We could optimize this further by doing a `rename` and only doing a copy if that fails, effectively a `rename_or_copy` as opposed to our `hardlink_or_copy` we normally use for uplifting.  The difference is that we don't do change tracking for `.crate` files but fully re-generate, so we don't benefit from keeping the `.crate` around in the original location.

### DIFF
--- a/src/cargo/ops/cargo_package/mod.rs
+++ b/src/cargo/ops/cargo_package/mod.rs
@@ -163,11 +163,8 @@ fn create_package(
     let filename = pkg.package_id().tarball_name();
     let build_dir = ws.build_dir();
     paths::create_dir_all_excluded_from_backups_atomic(build_dir.as_path_unlocked())?;
-    let dir = build_dir.join("package");
-    let mut dst = {
-        let tmp = format!(".{}", filename);
-        dir.open_rw_exclusive_create(&tmp, gctx, "package scratch space")?
-    };
+    let dir = build_dir.join("package").join("tmp-crate");
+    let dst = dir.open_rw_exclusive_create(&filename, gctx, "package scratch space")?;
 
     // Package up and test a temporary tarball and only move it to the final
     // location if it actually passes all our tests. Any previously existing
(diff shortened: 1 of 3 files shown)


src/cargo/ops/cargo_package/mod.rs
```diff
@@ -160,11 +160,8 @@ fn create_package(
     }
 
     let filename = pkg.package_id().tarball_name();
-    let dir = ws.build_dir().join("package");
-    let mut dst = {
-        let tmp = format!(".{}", filename);
-        dir.open_rw_exclusive_create(&tmp, gctx, "package scratch space")?
-    };
+    let dir = ws.build_dir().join("package").join("tmp-crate");
+    let dst = dir.open_rw_exclusive_create(&filename, gctx, "package scratch space")?;
```
epage: As this will cause a behavior change, Ive marked this as a draft while we wait for #15910 to propagate to users so they can prepare.

src/cargo/ops/cargo_package/mod.rs
weihanglo, line 1: Ready for merge?

epage: For myself, if we want an intermediate period, I feel like more than one release is probably needed.

epage: Eh, maybe 2 is fine?  That is 3 months

weihanglo, line 1: @epage I am going to do a submodule update today before new nightly branch cutoff. Mainly because of <https://github.com/rust-lang/cargo/pull/16337>.

Do you want to include this in 1.93 or 1.94? I am fine with either.

epage: 2 releases would be 1.93 but there is no rush on this

PR #15916, opened 2025-09-03
Update note on default-members and root package

tacertain:
Clarify usage of default-members with root package in workspaces.

I found this section of the doc confusing. I have made clarifications that align with my understanding. If this understanding is incorrect, then presumably a different change would be in order.

Doc change only

### DIFF
--- a/src/doc/src/reference/workspaces.md
+++ b/src/doc/src/reference/workspaces.md
@@ -137,11 +137,12 @@ members = ["path/to/member1", "path/to/member2", "path/to/member3/*"]
 default-members = ["path/to/member2", "path/to/member3/foo"]
 ```
 
-> Note: when a [root package](#root-package) is present,
-> you can only operate on it using `--package` and `--workspace` flags.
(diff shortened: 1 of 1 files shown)


src/doc/src/reference/workspaces.md
```diff
@@ -137,11 +137,12 @@ members = ["path/to/member1", "path/to/member2", "path/to/member3/*"]
 default-members = ["path/to/member2", "path/to/member3/foo"]
 ```
 
-> Note: when a [root package](#root-package) is present,
-> you can only operate on it using `--package` and `--workspace` flags.
+> Note: If `default-members` is specified and a [root package](#root-package) is present,
+> you can only operate on the root package using `--package` and `--workspace` flags.
+> There is no way to specify the root package explicitly in the `default-members` field.
```
weihanglo, line 142: Actually you can do that through `default-members = ["."]`.

Maybe we should remove this instead?

tacertain:
I'm just going to close this since clearly nobody cares.

PR #15917, opened 2025-09-03
WIP: fix(shell): Switch to annotate snippets for warnings

epage:

This is part of #15944

### DIFF
--- a/src/cargo/core/shell.rs
+++ b/src/cargo/core/shell.rs
@@ -222,7 +222,10 @@ impl Shell {
 
     /// Prints an amber 'warning' message.
     pub fn warn<T: fmt::Display>(&mut self, message: T) -> CargoResult<()> {
-        self.print(&"warning", Some(&message), &WARN, false)
+        let report = &[annotate_snippets::Group::with_title(
+            annotate_snippets::Level::WARNING.secondary_title(message.to_string()),
+        )];
+        self.print_report(report, false)
     }
 
     /// Prints a cyan 'note' message.
--- a/tests/testsuite/bad_config.rs
+++ b/tests/testsuite/bad_config.rs
@@ -962,7 +962,7 @@ fn dev_dependencies2() {
         .build();
     p.cargo("check").with_stderr_data(str![[r#"
 [WARNING] `dev_dependencies` is deprecated in favor of `dev-dependencies` and will not work in the 2024 edition
-(in the `foo` package)
+         (in the `foo` package)
 [LOCKING] 1 package to latest compatible version
 [CHECKING] foo v0.1.0 ([ROOT]/foo)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
@@ -1077,7 +1077,7 @@ fn build_dependencies2() {
         .build();
     p.cargo("check").with_stderr_data(str![[r#"
 [WARNING] `build_dependencies` is deprecated in favor of `build-dependencies` and will not work in the 2024 edition
-(in the `foo` package)
+         (in the `foo` package)
 [LOCKING] 1 package to latest compatible version
 [CHECKING] foo v0.1.0 ([ROOT]/foo)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
@@ -1185,7 +1185,7 @@ fn lib_crate_type2() {
     p.cargo("check")
         .with_stderr_data(str![[r#"
 [WARNING] `crate_type` is deprecated in favor of `crate-type` and will not work in the 2024 edition
-(in the `foo` library target)
+         (in the `foo` library target)
 [CHECKING] foo v0.5.0 ([ROOT]/foo)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 
@@ -1276,7 +1276,7 @@ fn bin_crate_type2() {
     p.cargo("check")
         .with_stderr_data(str![[r#"
 [WARNING] `crate_type` is deprecated in favor of `crate-type` and will not work in the 2024 edition
-(in the `foo` binary target)
+         (in the `foo` binary target)
 [CHECKING] foo v0.5.0 ([ROOT]/foo)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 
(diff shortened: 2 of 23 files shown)


src/cargo/core/shell.rs
```diff
@@ -222,12 +222,18 @@ impl Shell {
 
     /// Prints an amber 'warning' message.
     pub fn warn<T: fmt::Display>(&mut self, message: T) -> CargoResult<()> {
-        self.print(&"warning", Some(&message), &WARN, false)
+        let report = &[annotate_snippets::Group::with_title(
+            annotate_snippets::Level::WARNING.secondary_title(message.to_string()),
```
epage, line 226: I used `secondary_title` (custom styling) as a short term hack because the callers may be passing in styled text to workaround the fact that we don't have annotate snippets.  As we migrate things, we'll be able to use `primary_title` (no custom styling) as we should.

weihanglo (commented):
Do you want to rebase and fix the CI failure?

epage:
> Do you want to rebase and fix the CI failure?

Might be good to give any feedback in parallel so I know its worth putting in time to fix things :)

tests/testsuite/build_dir.rs
```diff
@@ -479,8 +479,8 @@ fn template_should_error_for_invalid_variables() {
         .with_status(101)
         .with_stderr_data(str![[r#"
 [ERROR] unexpected variable `fake` in build.build-dir path `{fake}/build-dir`
-
-[HELP] available template variables are `{workspace-root}`, `{cargo-cache-home}`, `{workspace-path-hash}`
+       
+       [HELP] available template variables are `{workspace-root}`, `{cargo-cache-home}`, `{workspace-path-hash}`
```
weihanglo: This seems a bit off. I would expect `Help: ` start without leading spaces.

tests/testsuite/freshness_checksum.rs
```diff
@@ -21,7 +21,7 @@ fn non_nightly_fails() {
     p.cargo("build -Zchecksum-freshness")
         .with_stderr_data(str![[r#"
 [ERROR] the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `stable` channel
-See https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.
+       See https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.
```
weihanglo: This See stuff should start from its own line, or have a `note:`

src/cargo/core/shell.rs
weihanglo, line 1: I feel like we will need to rework on most of our messages. Not all the `help` and `note` should be under the same group.

weihanglo, line 1: Couldn't find a concrete example, but conceptually adding indentations to trailing lines would add efforts for people to remove those indentations when copy-pasting.

weihanglo (commented):
Generally no objection to this change, though I'd like to understand the plan and where we are heading towards.

tests/testsuite/cargo_add/add_toolchain/stderr.term.svg
epage, line 1: Let's merge #15928 first

tests/testsuite/build_dir.rs
epage: Yes, we'd need to update the call site to call `print_report` directly so that it can add a `help` message to this Report's Group.

tests/testsuite/freshness_checksum.rs
epage: Yes, now that we have the full capabilities of annotate-snippets, we'll have to re-evaluate our warn/notes to decide if and how we should re-format them.  The question is which would be blocking for this PR vs things we do over time.  My hope is one like this could be non-blocking.

epage:
> Generally no objection to this change, though I'd like to understand the plan and where we are heading towards.

My long term vision is
- Multiple errors report by cargo that matches rustc
- as many warnings as possible converted to lints
- as many errors, warnings, notes, and lints as possible show code snippets
- All of this match rustc in style

This gets our existing output to be more like rustc and paves a way for using more features.

src/cargo/core/shell.rs
```diff
@@ -407,16 +409,22 @@ impl Shell {
...
             .diagnostic_terminal_width()
             .unwrap_or(annotate_snippets::renderer::DEFAULT_TERM_WIDTH);
-        writeln!(
-            self.err(),
-            "{}",
-            Renderer::styled().term_width(term_width).render(report)
-        )
+        let rendered = Renderer::styled().term_width(term_width).render(report);
+        self.err().write_all(rendered.as_bytes())?;
+        self.err().write_all("\n".as_bytes())?;
```
ehuss: minor nit, take it if you will

```suggestion
        self.err().write_all(b"\n")?;
```

tests/testsuite/collisions.rs
```diff
@@ -577,18 +577,18 @@ fn collision_with_root() {
...
 [CHECKING] abc v1.0.0 ([ROOT]/foo/abc)
 [DOCUMENTING] foo-macro v1.0.0 ([ROOT]/foo/foo-macro)
 [DOCUMENTING] abc v1.0.0 ([ROOT]/foo/abc)
 [FINISHED] `dev` profile [unoptimized + debuginfo] target(s) in [ELAPSED]s
 [GENERATED] [ROOT]/foo/target/doc/abc/index.html and 1 other file
+         The lib target `foo_macro` in package `foo-macro v1.0.0` has the same output filename as the lib target `foo_macro` in package `foo-macro v1.0.0 ([ROOT]/foo/foo-macro)`.
+         Colliding filename is: [ROOT]/foo/target/doc/foo_macro/index.html
+         The targets should have unique names.
+         This is a known bug where multiple crates with the same name use
+         the same path; see <https://github.com/rust-lang/cargo/issues/6313>.
```
ehuss: Some of these warnings seem to have some ordering issues where the message is split? Like the other messages are showing up in the middle of the warning.

src/cargo/core/shell.rs
epage: Posted #15937 to take care of this since this was actually from a PR that has already been merged.

epage:
This is blocked until the multi-line warnings have been cleaned up, similar to how #15945 is being handled.

PR #15918, opened 2025-09-03
chore: downgrade to libc@0.2.174

weihanglo:

We have downgraded this in rust-lang/cargo#15851
We should also let renovate ignore it.

This also again downgrades libc to 0.2.174 again.

See <https://github.com/rust-lang/rust/pull/146176#issuecomment-3251098344>.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -8,6 +8,8 @@
   ignorePaths: [
     '**/tests/**',
   ],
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Sorry, forgot that we did this.

Renovate seems somewhat broken right now, so I'm not sure how much this will help. This also requires us to remember to remove it, which isn't great. I'll open some issues.

I'll merge just to unblock things.

PR #15922, opened 2025-09-04
fix(cli): Use Shell::print_report for error chains

epage:

This replaces `Shell::error`, etc calls (see also #15917) with `Shell::print_report` when error chains are involved in an effort to render things more like rustc


epage:
Unlike #15917 which can be touched up in the callers if we don't like it, this will need a more invasive change of creating a diagnostic-first error type if we're not satisfied with the results as-is, so looking to solicit feedback on the general idea and what all directions we may want to take this before going down those roads.

PR #15923, opened 2025-09-04
Render individual compilation sections in `--timings` pipeline graph

Kobzol:
This PR expands the support for `--json-timings` (added in https://github.com/rust-lang/cargo/pull/15780), by rendering the individual compilation sections in the pipeline graph of the `--timings` page.

Before, the linking section was only shown in the table at the bottom of the page, now it should also be clearly visible in the compilation graph, which should help more quickly understand how much time is spent in linking.

image

I also added a legend to the pipeline graph, to explain what do the colors mean.
image

One wart is that the linking time actually ends a bit before the unit ends, so there is some "vacuum" at the end where rustc does cleanup, persists files to disk, deallocates things, etc. That's why I marked the blue section "Frontend/rest" in the legend.

Same as for https://github.com/rust-lang/cargo/pull/15780, e.g.:
```bash
export RUSTC=`rustup +nightly which rustc`
target/debug/cargo build -Zsection-timings --timings
```
on some crate, e.g. [ripgrep](https://github.com/BurntSushi/ripgrep).

### DIFF
--- a/src/cargo/core/compiler/timings.js
+++ b/src/cargo/core/compiler/timings.js
@@ -66,7 +66,7 @@ const BG_COLOR = getCssColor('--background');
 const CANVAS_BG = getCssColor('--canvas-background');
 const AXES_COLOR = getCssColor('--canvas-axes');
 const GRID_COLOR = getCssColor('--canvas-grid');
-const BLOCK_COLOR = getCssColor('--canvas-block');
+const CODEGEN_COLOR = getCssColor('--canvas-codegen');
 const CUSTOM_BUILD_COLOR = getCssColor('--canvas-custom-build');
 const NOT_CUSTOM_BUILD_COLOR = getCssColor('--canvas-not-custom-build');
 const DEP_LINE_COLOR = getCssColor('--canvas-dep-line');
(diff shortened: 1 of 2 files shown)


epage:
> One wart is that the linking time actually ends a bit before the unit ends, so there is some "vacuum" at the end where rustc does cleanup, persists files to disk, deallocates things, etc. That's why I marked the blue section "Frontend/rest" in the legend.

Should we just have this as a separate "Other" category?

Kobzol:
Could be. We could create an artificial "Other" section, which would be from the end of the last known section to the end of the compilation of the unit (as long as this duration is non-empty, of course). Does that work?

src/cargo/core/compiler/timings.js
```diff
@@ -188,7 +201,9 @@ function render_pipeline_graph() {
 function get_section_color(name) {
     if (name === "codegen") {
         return CODEGEN_COLOR;
-    } else {
+    } else if (name === "link") {
+        return LINK_COLOR;
+    } {
```
epage: `else`?

epage:
> Could be. We could create an artificial "Other" section, which would be from the end of the last known section to the end of the compilation of the unit (as long as this duration is non-empty, of course). Does that work?

Works for me!

weihanglo (commented):
This is nice. Do you have a sense of when the rustc flag will be stabilized?

Kobzol:
I mean, I could propose stabilization immediately, but I'm not sure if it's a good idea yet. For example, @bjorn3 is currently doing some changes of how LTO works, and while doing that, we found out that you can have multiple codegen + linking sections during a single rustc invocation. That kind of breaks the assumption that `--timings` makes (even without taking sections into account).

Kobzol:
Fixed the missing `else` and added the "other" section. I can imagine improving this page further (e.g. adding a tooltip on hover on a crate that would show the names and durations of the individual sections), but it's kinda bothersome with the raw JavaScript and the canvas :/

epage (approved):
Thanks!

PR #15927, opened 2025-09-04
fix(manifest): Report script manifest errors for the right line number

epage:

Currently, if you have a manifest parse error for a cargo script, the line number will be relative to the start of the frontmatter and not the source file.  This changes it so we get the line number relative to the start of the source file.

### Notes

To do this, I added span tracking to our frontmatter parser.  To make this easier, I used some helpers from winnow which is an existing transitive dependency.

This span tracking will also come into use when I change the frontmatter parser to start reporting spans in errors so we get annotated snippets of source in the error messages to users.

### DIFF
--- a/src/cargo/util/frontmatter.rs
+++ b/src/cargo/util/frontmatter.rs
@@ -4,12 +4,20 @@ type Span = std::ops::Range<usize>;
 
 #[derive(Debug)]
 pub struct ScriptSource<'s> {
+    /// The full file
     raw: &'s str,
+    /// The `#!/usr/bin/env cargo` line, if present
     shebang: Option<Span>,
+    /// The code fence opener (`---`)
     open: Option<Span>,
+    /// Trailing text after `ScriptSource::open` that identifies the meaning of
+    /// `ScriptSource::frontmatter`
     info: Option<Span>,
+    /// The lines between `ScriptSource::open` and `ScriptSource::close`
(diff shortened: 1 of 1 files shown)


src/cargo/util/frontmatter.rs
```diff
@@ -1,41 +1,58 @@
 use crate::CargoResult;
 
+type Span = std::ops::Range<usize>;
+
 #[derive(Debug)]
 pub struct ScriptSource<'s> {
```
weihanglo, line 6: While it seems straight forwards atm, we may want to write down an example and teach which part is what, so people don't need to read the implementation?

tests/testsuite/script/cargo.rs
```diff
@@ -182,6 +182,41 @@ fn requires_z_flag() {
         .run();
 }
 
+#[cargo_test(nightly, reason = "-Zscript is unstable")]
+fn manifest_parse_error() {
+    let script = r#"#!/usr/bin/env cargo
```
weihanglo, line 188: Should we add a comment letting people know these blank lines are intentional for showing span is tracked?

epage: Added

src/cargo/util/frontmatter.rs
epage: Added comments.  Let me know if you were intending something else or more.

PR #15928, opened 2025-09-04
fix(cli): Adjust messages to match rustc 

### LINKED ISSUE #12740: Cargo and rustc are using different tones of red for errors on the terminal

epage:

Fixes #12740

This is also prep for #15917, #15922

Changes:
- `note` changes from Cyan to Bright Green
  - `cargo info` had to be changed because the header and context colors became the same
- Switch to Bright colors which is usually redundant with Bold but that is theme dependent
  - `warning`  only changes to Bright on Windows which is to work around a shell issue on some Windows versions
  - except for `PLACEHOLDER` to tell it apart from `LITERAL`
- the `:` in `note:` is no longer bolded to match rustc

### Notes

While annotate-snippets does not have a style for every one of our styles, I updated our manual styles to be similar to annotate snippets.

For annotate snippets color definitions, see https://github.com/rust-lang/annotate-snippets-rs/blob/d38b08b81d7d574369ecfd30124195a970b37fd3/src/renderer/mod.rs#L41-L78

### DIFF
--- a/tests/testsuite/cargo/help/stdout.term.svg
+++ b/tests/testsuite/cargo/help/stdout.term.svg
@@ -2,8 +2,9 @@
   <style>
     .fg { fill: #AAAAAA }
     .bg { background: #000000 }
+    .fg-bright-cyan { fill: #55FFFF }
+    .fg-bright-green { fill: #55FF55 }
     .fg-cyan { fill: #00AAAA }
-    .fg-green { fill: #00AA00 }
     .container {
       padding: 0 10px;
       line-height: 18px;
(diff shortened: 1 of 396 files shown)


tests/testsuite/cargo/help/stdout.term.svg
epage, line 1: Something is wrong with the help output here

epage: Forgot to update the `color_print::cstr` calls

tests/testsuite/cargo_info/specify_version_within_ws_and_match_with_lockfile/stderr.term.svg
epage, line 1: I am a little disappointed that our `note` color is the same as our headers.  However, unsure if we have the plethora of options to do much about it wholly on our end

tests/testsuite/cargo_add/default_features/stderr.term.svg
epage, line 1: Aside: strange that the deleted version had `Adding` in cyan despite files had it being green *and* the code isn't conditional

tests/testsuite/cargo_info/specify_version_within_ws_and_match_with_lockfile/stderr.term.svg
epage: Maybe we should make sure this is merged after beta branch so we have extra time to tweak things (e.g. turning some notes into helps)

tests/testsuite/cargo_add/default_features/stderr.term.svg
weihanglo: Really strange. It happened not only once, so probably some bug in term_svg in snapbox?

https://github.com/rust-lang/cargo/blob/3b379fcc541b39321a7758552d37e5e0cc4277b9/tests/testsuite/cargo_add/default_features/stderr.term.svg#L31

I have produced it, but not sure why. All `Adding` status with a `(available:)` extra messages seem to have the issue.
image

tests/testsuite/cargo_info/within_ws/stderr.term.svg
weihanglo, line 1: A bit prefer to the previous version. `note` is more outstanding.

weihanglo (approved):
Looks good for matching rustc style. Thank you!

PR #15929, opened 2025-09-05
test(credential): Switch more expected results to snapshots

epage:

Had to deal with manually updating these when changing notes.
This will avoid that.


Muscraft (approved):

weihanglo (approved):
Thanks!

PR #15930, opened 2025-09-05
fix: Switch from --nocapture to --no-capture

epage:

`--no-capture` was added in 1.88, with `--nocapture` being soft-deprecated.


ehuss (approved):
Thanks!

PR #15935, opened 2025-09-06
refactor: replace flock with std flock

weihanglo:

Replace flock with std flock.

Given we've supported Oracle since 2022-12 and don't want to break people,
this is blocked on std supporting flock emulation via fcntl rust-lang/rust#146269


weihanglo:
https://github.com/rust-lang/rust/pull/146269 merged.

ehuss (approved):
Thanks!

Issue #15882, opened 2025-08-24
Turning cargo into a git subtree in rust-lang/rust


weihanglo:
### Problem

This is an issue tracking how many issue we encountered because Cargo is not a Git subtree in rust-lang/rust repository. See also the discussion in [#t-cargo > Turning cargo into a subtree](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Turning.20cargo.20into.20a.20subtree/with/527165966).

Below is a summary last time the Cargo team has discussed this (originally posted in [#t-cargo > Turning cargo into a subtree @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Turning.20cargo.20into.20a.20subtree/near/493511854)).

---

The Cargo team has discussed subtree migration last week. We've got a couple of questions and would like to understand before proceeding:

* Motiviation of turning into a subtree.
* Burden of resolving conflicts
* Issue / PR managements
* Possibility to revert to submodule

### Motivation of turning into a subtree

It seems that it was brought up again because there were some recent pull requests blocked by Cargo's overly-fitted test assertions. Apart from that, are there any other reasons or benefits people wanting Cargo to be a subtree? How painful is it that people need it to happen soon otherwise it toils the development?

From 1.75 (2023-12-28) to 1.83 (as of 2025-01-07), there are around 11 pull requests opened in rust-lang/cargo because test failures in Cargo's test suite really blocked their pull request in rust-lang/rust.

See a list of pull requests

> There are  other cross-repo efforts like check-cfg and checksum-freshness. However I don't think it is better do it in subtree, as those changes are way larger and may require scrutiny from the Cargo team. See the full list of PRs in the end of this post

* https://github.com/rust-lang/cargo/pull/14990
* https://github.com/rust-lang/cargo/pull/14963
* https://github.com/rust-lang/cargo/pull/14602
* https://github.com/rust-lang/cargo/pull/14341
* https://github.com/rust-lang/cargo/pull/14135
* https://github.com/rust-lang/cargo/pull/13964
* https://github.com/rust-lang/cargo/pull/13415
* https://github.com/rust-lang/cargo/pull/13388
* https://github.com/rust-lang/cargo/pull/13382
* https://github.com/rust-lang/cargo/pull/13072
* https://github.com/rust-lang/cargo/pull/12951
* https://github.com/rust-lang/cargo/pull/12800

[...]

PR #15936, opened 2025-09-08
chore(ci): reject uncanonicalized issue link in commits

weihanglo:

This is needed if we want to turn cargo submodule into a subtree in
the rust-lang/rust repository (rust-lang/cargo#15882), because
commits are cherry-picked with their messages left as-is, and the magic keyword
like `fixes #99999` may unintentionally close issues in rust-lang/rust
while it was intended only for rust-lang/cargo issues.

Unlike `check-commits = false`, instead of rejecting any issue links
in commit messages, the `"uncanonicalized"` option still allows issues
in a canonicalized form, such as

* `https://github.com/rust-lang/cargo/issues/15882`
* `rust-lang/cargo#15882`

but the bare issue number like `#15882` will be warned by triagebot
in a PR comment.

### Drawbacks

It is bad because contributors will be forced to write commit messages
in a specific style. Worse if we decide not to be a subtree and then
this change puts unnecessary works on contributors.

### Is it the last step towards rust-lang/cargo#15882?

Personally I consider not. Because merge commit still contains
the PR number in the uncanonicalized form, as we use GitHub merge queue
and it hasn't yet support [merge commit template][1].

However, this change on its own makes issue link more obvious and
less ambiguous, especially when reading from outside GitHub web UI.
It also stops us from adding more uncanonicalized issue links in the
future, which we already have too many in the past commits [2]

There are some other concerns in <https://github.com/rust-lang/cargo/issues/15882#issuecomment-3224806645> and <https://github.com/rust-lang/cargo/issues/15882#issuecomment-3267730188>. Until we resolve them and feel this is the only thing left, we might not want to merge this.

For discussions, see

* https://forge.rust-lang.org/triagebot/issue-links.html
* https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Turning.20cargo.20into.20a.20subtree/near/537951273
* https://github.com/rust-lang/cargo/issues/15882

[1]: https://github.com/orgs/community/discussions/5955
[2]: https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Turning.20cargo.20into.20a.20subtree/near/537560661

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -61,6 +61,9 @@ review_labels = ["S-waiting-on-review"]
 remove_labels = ["S-waiting-on-author"]
 add_labels = ["S-waiting-on-review"]
 
+[issue-links]
+check-commits = "uncanonicalized"
+
 [autolabel."A-build-execution"]
 trigger_files = [
   "src/cargo/core/compiler/compilation.rs",


weihanglo:
cc @epage and @ehuss  as you are both in the review rotation:

https://github.com/rust-lang/cargo/blob/a9e120f95ea0a39060f260822163088749d895ac/triagebot.toml#L52-L53

epage:
The triagebot documentation doesn't make it clear why we are using this setting and the other two links have a lot of content in them for other topics.

Could you update the PR description with the motivation for this, summarizing the relevant parts of those conversations?

Also, how close are we for being ready for subtrees?  Should this be one of the later things we do so we don't take on costs when we aren't getting the benefits?

weihanglo:
Updated. Let me know if it is still unclear.

epage:
The part related to the motivation that remained unclear was I didn't know that subtrees cherry pick all commits.  That seems noisy for rustc but oh well, thats the choice they make.

Urgau (approved):

PR #15939, opened 2025-09-08
fix(manifest): Show error source to users

epage:

Ooch our way towards rustc's quality of error reporting to make up for the fact that users won't see most frontmatter rustc error messages.

Several parts of this are not ideal yet
- Frontmatter close should show the open and show the EOF, instead of
  pointing at the open
- Trailing content on close will usually have a newline
- Multiple frontmatters should show the original frontmatter
- Some content, like supported infostrings, should go in a help
- Ideally we try to recover on no closing and instead point out a
  mismatched open/close

But this is still an improvement over nothing!

### DIFF
--- a/src/cargo/util/toml/mod.rs
+++ b/src/cargo/util/toml/mod.rs
@@ -67,8 +67,7 @@ pub fn read_manifest(
     let mut errors = Default::default();
 
     let is_embedded = is_embedded(path);
-    let contents = read_toml_string(path, is_embedded, gctx)
-        .map_err(|err| ManifestError::new(err, path.into()))?;
+    let contents = read_toml_string(path, is_embedded, gctx)?;
     let document = parse_document(&contents)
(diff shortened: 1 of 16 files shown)


src/cargo/util/frontmatter.rs
```diff
@@ -120,8 +120,15 @@ impl<'s> ScriptSource<'s> {
 
         source.content = content_start..content_end;
 
-        let repeat = Self::parse(source.content())?;
-        if repeat.frontmatter.is_some() {
+        if let Some(nl_end) = strip_ws_lines(input.as_ref()) {
+            let _ = input.next_slice(nl_end);
+        }
+        let fence_length = input
```
weihanglo, line 132: How was it not precise? I didn't see any test snapshot changed

weihanglo: Resolved after review all commits. It was unclear when seeing the commit standalone.

